PR #882, opened 2020-12-17
BLD Enable ccache for x86_64

rth:
Closes https://github.com/iodide-project/pyodide/issues/838

Following discussion https://github.com/iodide-project/pyodide/issues/838#issuecomment-747054343 this would only help for numpy compilation time. It doesn't take that long, but this is also not too complicated, so it might be worth it.

Also in the future some packages will require scipy as a build dependency (e.g. latest versions of scikit-learn) and that takes takes much longer to build.

Also even if we update numpy to a version that has wheels for Python 3.8 (https://github.com/iodide-project/pyodide/issues/437) we could use them for build dependencies, but not for benchmarks. Since wheels use OpenBLAS making performance comparison with our reference BLAS implementation not practical.

### DIFF
--- a/pyodide_build/pywasmcross.py
+++ b/pyodide_build/pywasmcross.py
@@ -31,6 +31,7 @@ import os
 from pathlib import Path
 import re
 import subprocess
+import shutil
 import sys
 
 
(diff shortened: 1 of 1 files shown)


rth:
So here building packages takes 13min 32 s, while on master it's 14min18s, but it's hard to tell how much of that is just random fluctuation.

dalcde (approved):

PR #883, opened 2020-12-17
BLD Don't use emscripten ports in global LDFLAGS

rth:
Emscripten ports are specified in package LDFLAGS as needed. So there is no need to also specify them in the global LDFLAGS. This should make building the CPython + code pyodide package slightly faster.


rth:
Hmm, no it looks like those options still need to be passed globally. Which is strange as removing them earlier in https://github.com/iodide-project/pyodide/pull/658 worked fine. In any case the benefit is minimal and probably not worth the time investigating.

Issue #877, opened 2020-12-17
Python 3.8 warnings


casatir:
[Python issue #34850](https://bugs.python.org/issue34850) introduces a warning in Python 3.8 about missusing the `is` (or `is not`) statement.

In Pyodide, for some several 'old' modules we can get this warning. This occurs at least for `numpy` and `matplotlib`. This shows up in `console.html`with #875

```python
import numpy

/lib/python3.8/site-packages/numpy/ma/core.py:4467: SyntaxWarning: "is" with a literal. Did you mean "=="?
  if self.shape is ():

import matplotlib.pyplot

/lib/python3.8/site-packages/matplotlib/pyplot.py:2422: SyntaxWarning: "is" with a literal. Did you mean "=="?
  if fignum is False or fignum is 0:
/lib/python3.8/site-packages/matplotlib/contour.py:870: SyntaxWarning: "is not" with a literal. Did you mean "!="?
  if self.extend is not 'neither':
```

There may be other warnings introduced in 3.8. One should identify theme and start upgrading impacted modules though.

**Edit:** list of packages that needs updating
 - [x] numpy
 - [x] matplotlib
 - [x] mpmath
 - [ ] ...

rth:
yes, it will certainly become a major issue once we are able to see warnings in `console.html` :) 

We should be able update packages (except for scipy cf https://github.com/iodide-project/pyodide/issues/549) to a version post Oct. 14, 2019 (Python 3.8 release date).   Typically they should have a Python 3.8 compatible release by then. Unless some are blocked by out outdated scipy version (e.g. scikit-learn, scikit-image, statsmodels). This mostly concerns compiled packages with patches, all the rest should be already up to date.

rth:
PR for numpy done in https://github.com/iodide-project/pyodide/pull/887

rth:
There is also,
```
/lib/python3.8/site-packages/pandas/compat/__init__.py:117: UserWarning: Could not import the lzma module. Your installed Python is incompl
ete. Attempting to use lzma compression will result in a RuntimeError.
```
for `pandas` which is a bit strange, since lzma is a non mandatory build option.

rth:
I believe all of these were fixed in Pyodide 0.20.0.

The pandas warning was also fixed in pandas 1.4.0+ (which we ship) since https://github.com/pandas-dev/pandas/pull/43495

PR #863, opened 2020-12-15
Document some emsdk patches

dalcde:

PR #886, opened 2020-12-18
Upgrading matplotlib to current version (v3.3.3).

casatir:
Matplotlib is 2 years old and 6000+ commits late. Pyodide's version is `2.2.3` while upstream is `3.3.3`. Moreover, use of Python 3.8, shows warning as mentioned in #877. Needless to say that it's time to update!

This PR ports Pyodide's Matplotlib version to latest `3.3.3`. I put some comments here to explain how it is achieved for easiest review and as a memo for later updates. Key steps are:

- cleaning patches since most of them have been merged upstream
- updating `fontList.json` to `v330` format
- adding Pillow as dependency
- force the use of `dummy_threading` instead of threading (see [this comment](https://github.com/matplotlib/matplotlib/pull/11262#issuecomment-390322014) )
- fix two issues in `wasm_backend.py`

I am not sure if the patch that disable Macos support is necessary.

I have visually tested 2D/3D plots and tools from the toolbar (zoom, exports, ...) and launch `test_matplotlib.py`. Looks right.

You can [test it](https://devel.basthon.fr/notebook) but the version here can change soon.


dalcde:
Would you mind documenting the patches you add? It would make future upgrading efforts easier. You can simply add explanations to the beginning of a patch file as long as it doesn't accidentally look like an actual patch (see e.g. #863)

packages/matplotlib/patches/fix-threading.patch
```diff
@@ -0,0 +1,67 @@
+diff --git a/lib/matplotlib/backend_bases.py b/lib/matplotlib/backend_bases.py
+--- a/lib/matplotlib/backend_bases.py
++++ b/lib/matplotlib/backend_bases.py
+@@ -1993,6 +1993,7 @@ class FigureCanvasBase:
+ 
+     def draw(self, *args, **kwargs):
+         """Render the `.Figure`."""
++        pass
```
dalcde, line 8: Is this accidentally left over or is there some mysterious python shenanigans?

casatir: I think it is accidental but can't figure out why it doesn't throw error... I just fixed it passing by.

casatir: This is not fixed upstream and they have multiple use of `pass` in this file. I'll proposes a PR upstream.

casatir: Empty function with docstring seems [allowed by the Python's grammar](https://stackoverflow.com/questions/17735170/why-does-python-allow-an-empty-function-with-doc-string-body-without-a-pass).

casatir:
Sure, I will comment patches.

packages/matplotlib/patches/fix-threading.patch
casatir: Anyway, this is confusing in this PR. So, upstream PR + removed from this PR if you agree.

dalcde: Sounds good to me. Removing from this PR also means one less patch to worry about when we update later.

packages/matplotlib/patches/fix-freetype.patch
```diff
@@ -0,0 +1,16 @@
+diff --git a/setupext.py b/setupext.py
+--- a/setupext.py
++++ b/setupext.py
+@@ -522,6 +522,9 @@ class FreeType(SetupPackage):
+     name = "freetype"
+ 
+     def add_flags(self, ext):
++        if "PYODIDE_PACKAGE_ABI" in os.environ:
++            ext.libraries.append('freetype')
++            return
```
rth, line 10: Could you contribute this patch upstream?

packages/matplotlib/patches/disable_macosx_backend.patch
```diff
@@ -1,15 +1,13 @@
...
+--- a/setupext.py
++++ b/setupext.py
+@@ -665,6 +665,7 @@ class BackendMacOSX(OptionalPackage):
      name = 'macosx'
  
-     def check_requirements(self):
-+        raise CheckFailed("MacOSX backend disabled")
+     def check(self):
++        raise Skipped("MacOSX backend disabled")
          if sys.platform != 'darwin':
```
rth, line 9: Maybe 
```suggestion
         if sys.platform != 'darwin' or "PYODIDE_PACKAGE_ABI" in os.environ:
```
and we could also try to upstream this fix. Is there a need to rename this method?

packages/matplotlib/patches/fix-threading.patch
```diff
@@ -0,0 +1,67 @@
...
+         """
+diff --git a/lib/matplotlib/backends/backend_agg.py b/lib/matplotlib/backends/backend_agg.py
+--- a/lib/matplotlib/backends/backend_agg.py
++++ b/lib/matplotlib/backends/backend_agg.py
+@@ -22,6 +22,8 @@ TODO:
+ """
+ 
+ try:
++    # force use of dummy_threading
++    raise ImportError()
```
rth, line 20: What happens if you keep using the threading import? Recently in https://github.com/iodide-project/pyodide/pull/796 I have removed all `dummy_threading` patches, using it for actual threading will lead to deadlocks anyway. While `threading` works for locks etc in Python 3.8

packages/matplotlib/patches/hardcoded-font-cache.patch
```diff
@@ -0,0 +1,16 @@
+diff --git a/lib/matplotlib/font_manager.py b/lib/matplotlib/font_manager.py
+--- a/lib/matplotlib/font_manager.py
++++ b/lib/matplotlib/font_manager.py
+@@ -1394,8 +1394,7 @@ def is_opentype_cff_font(filename):
+         return False
+ 
+ 
+-_fmcache = os.path.join(
+-    mpl.get_cachedir(), 'fontlist-v{}.json'.format(FontManager.__version__))
++_fmcache = os.path.join(os.path.dirname(__file__), "fontlist.json")
```
rth, line 10: Do you understand why we need this? Couldn't we just use the version of FontManager that's used in matplotlib?

rth (commented):
Thanks for doing this work @casatir ! A few comments are below. 

Also there are couple test failures that would need investigating.

packages/matplotlib/patches/fix-freetype.patch
casatir: Sure.

packages/matplotlib/patches/disable_macosx_backend.patch
casatir: You mean the check method?

packages/matplotlib/patches/fix-threading.patch
casatir: I'll try. Thought it was broken in Pyodide.

packages/matplotlib/patches/hardcoded-font-cache.patch
casatir: You mean the `fontlist.json` file ? Removing it will lead to a full rebuild which may take some time. See below.

```python

def _rebuild():
    global fontManager
    _log.info("Generating new fontManager, this may take some time...")
    fontManager = FontManager()
    json_dump(fontManager, _fmcache)


try:
    fontManager = json_load(_fmcache)
except Exception:
    _rebuild()
else:
    if getattr(fontManager, '_version', object()) != FontManager.__version__:
        _rebuild()
    else:
        _log.debug("Using fontManager instance from %s", _fmcache)
```

I'll investigate to see if it really take some time.

casatir: Between about 1.5 second to build the font cache and save it to virtual FS. What do we do?

casatir: Doing this a see big differences between generated version and the one of this PR. If we choose to keep it on the Pyodide I should update it.

packages/matplotlib/patches/fix-threading.patch
casatir: It seems to work! There is a `threading.Timer` used in `font_manager.py` just to delay a warning (it was also broken with dummy threading). I'll remove it.

casatir:
Regarding the CI, two tests fails. One should be corrected with the last commit and the other comes from `yt` which is too old (it tries to import the internal matplotlib's PNG module which uses PIL now).

Will submit a separate PR to fix `yt`.

casatir:
CI output confirms that to get all green, we should merge #890.

casatir:
Tests are OK. It remains to decide if we ship `fontlist.json` with Pyodide or rebuild it at each import.
Any advice?

packages/matplotlib/patches/disable_macosx_backend.patch
rth: Yes, why rename it from `check_requirements` to `check`?

packages/matplotlib/patches/hardcoded-font-cache.patch
rth: Thanks for investigating, let's keep the vendored version in Pyodide and update it.

rth:
> if we ship fontlist.json with Pyodide

I think it might be indeed better to keep doing it, thanks for checking.

casatir:
Great! Just added a test to ensure vendored version is exactly the same as built one.

This as one side effect: it also tests that the build work. This ensure our patch about removing the use of `threading.Timer` in `font_manager.py` is OK.

packages/matplotlib/patches/disable_macosx_backend.patch
casatir: The rename comes [from upstream](https://github.com/matplotlib/matplotlib/commit/aceddbfa769d23479608be262680c436c4bb4dc3).

casatir:
Apart from upstream PR, what remains to be done?

rth (approved):
Thanks @casatir !

PR #887, opened 2020-12-18
Update to numpy 1.17.5

rth:
Closes https://github.com/iodide-project/pyodide/issues/437
Partially addresses https://github.com/iodide-project/pyodide/issues/877

Updates numpy to 1.17.5 , released 01/01/2020 which should be fully compatible with Python 3.8. Not trying the very last numpy, as I'm not sure it would work well with a very outdated scipy that we have.

The patches are migrated but there is still a build issue of duplicate symbols at link time that need to be fixed.


rth:
There is still a linking error about,
```
error: Linking globals named 'npy_sinl': symbol multiply defined!
```
which should be defined once [here](https://github.com/numpy/numpy/blob/648fe68f79dfbad21947db4b55873703db0675c4/numpy/core/include/numpy/npy_math.h#L168), but this header is imported multiple times. I won't have time to investigate this in more detail now.

rth:
So building numpy 1.17.5 should now work, however I'm still getting errors are runtime.

For instance in chrome,
**test_typed_arrays**
```
E           conftest.JavascriptException: Error: Pyodide has suffered a fatal error, refresh the page. Please report this to the Pyodide maintainers.
E               at Object.Module.public_api.<computed> [as loadPackage] (http://127.0.0.1:53665/pyodide.js:380:17)
E               at run (eval at executeAsyncScript (:457:5), <anonymous>:5:51)
E               at eval (eval at executeAsyncScript (:457:5), <anonymous>:8:40)
E               at eval (eval at executeAsyncScript (:457:5), <anonymous>:24:15)
E               at eval (eval at executeAsyncScript (:457:5), <anonymous>:25:12)
E               at executeAsyncScript (<anonymous>:457:26)
E               at <anonymous>:473:29
E               at callFunction (<anonymous>:361:33)
E               at <anonymous>:371:23
E               at <anonymous>:372:3
```
**test_runpythonasync_numpy**
```
>           raise JavascriptException(retval[1], retval[2])
E           conftest.JavascriptException: TypeError: Cannot read property 'apply' of undefined

E               at http://127.0.0.1:53665/pyodide.asm.wasm:wasm-function[2745]:0x2b87a3
```
<details>

```
E               at obj.<computed> (http://127.0.0.1:53665/pyodide.asm.js:9:189946)
E               at PyUFunc_GenericFunction (<anonymous>:wasm-function[3898]:0x11a18e)
E               at __static_1984 (<anonymous>:wasm-function[3945]:0x11eb7b)
E               at byn$fpcast-emu$__static_1984 (<anonymous>:wasm-function[8968]:0x19a901)
E               at _PyObject_MakeTpCall (http://127.0.0.1:53665/pyodide.asm.wasm:wasm-function[741]:0x1cc4bf)
E               at http://127.0.0.1:53665/pyodide.asm.wasm:wasm-function[2745]:0x2b8712
E               at _PyEval_EvalFrameDefault (http://127.0.0.1:53665/pyodide.asm.wasm:wasm-function[2738]:0x2b587d)
E               at byn$fpcast-emu$_PyEval_EvalFrameDefault (http://127.0.0.1:53665/pyodide.asm.wasm:wasm-function[17136]:0x6cb604)
E               at _PyEval_EvalCodeWithName (http://127.0.0.1:53665/pyodide.asm.wasm:wasm-function[2734]:0x2b05dc)
E               at PyEval_EvalCode (http://127.0.0.1:53665/pyodide.asm.wasm:wasm-function[2733]:0x2af7aa)
E               at __static_1147 (http://127.0.0.1:53665/pyodide.asm.wasm:wasm-function[2679]:0x2ad11d)
E               at byn$fpcast-emu$__static_1147 (http://127.0.0.1:53665/pyodide.asm.wasm:wasm-function[24226]:0x6e4a01)
E               at __static_686 (http://127.0.0.1:53665/pyodide.asm.wasm:wasm-function[1606]:0x2167c3)
E               at byn$fpcast-emu$__static_686 (http://127.0.0.1:53665/pyodide.asm.wasm:wasm-function[23765]:0x6e31cb)
E               at PyVectorcall_Call (http://127.0.0.1:53665/pyodide.asm.wasm:wasm-function[744]:0x1cc8d8)
E               at PyCFunction_Call (http://127.0.0.1:53665/pyodide.asm.wasm:wasm-function[753]:0x1cdddb)
E               at _PyEval_EvalFrameDefault (http://127.0.0.1:53665/pyodide.asm.wasm:wasm-function[2738]:0x2b5da5)
E               at byn$fpcast-emu$_PyEval_EvalFrameDefault (http://127.0.0.1:53665/pyodide.asm.wasm:wasm-function[17136]:0x6cb604)
E               at _PyEval_EvalCodeWithName (http://127.0.0.1:53665/pyodide.asm.wasm:wasm-function[2734]:0x2b05dc)
E               at _PyFunction_Vectorcall (http://127.0.0.1:53665/pyodide.asm.wasm:wasm-function[749]:0x1cd1d5)
E               at byn$fpcast-emu$_PyFunction_Vectorcall (http://127.0.0.1:53665/pyodide.asm.wasm:wasm-function[16373]:0x6c8b0d)
[...]
E               at PyVectorcall_Call (http://127.0.0.1:53665/pyodide.asm.wasm:wasm-function[744]:0x1cc8d8)
E               at PyObject_Call (http://127.0.0.1:53665/pyodide.asm.wasm:wasm-function[745]:0x1cc993)
E               at _pyproxy_apply (http://127.0.0.1:53665/pyodide.asm.wasm:wasm-function[335]:0x1a9ca8)
E               at Module.PyProxyClass.apply (http://127.0.0.1:53665/pyodide.asm.js:9:175880)
E               at Object.apply (http://127.0.0.1:53665/pyodide.asm.js:9:180266)
E               at Object.globalThis.languagePluginLoader.Module.runPython (http://127.0.0.1:53665/pyodide.js:421:48)
E               at Object.globalThis.languagePluginLoader.Module.runPythonAsync (http://127.0.0.1:53665/pyodide.js:499:19)
E               at async run (eval at executeAsyncScript (:457:5), <anonymous>:6:14)
E               at async eval (eval at executeAsyncScript (:457:5), <anonymous>:16:34)
```
</details>

Which is related to https://github.com/iodide-project/pyodide/issues/834

I have built with `Werror=implicit-function-declaration -Werror=mismatched-parameter-types -Werror=mismatched-return-types` as done in scipy, and with `-Oz` which should enable assertions. That was the fix for other similar cases https://github.com/emscripten-core/emscripten/issues/11722 but doesn't seem to help here.

@joemarshall  did you by chance encounter similar issues in your other PRs?


Also, the diff will be smaller once #1287 is merged.

joemarshall:
just noticed this - I did have similar problems in porting scipy to new emscripten, they were mostly because if something is multiply defined with slightly different signatures, even with all the stuff turned on, it outputs symbols in two versions, one is just the normal one, the other is with some kind of type-conflict label in front of the name, I can't remember exactly what it was, something like '$type-conflict$name or similar. If you call the conflicting signature one they are defined as null so they crash out. Even with all those linker and compiler flags scipy still generated them until I tracked down where they were coming from. Mostly it is an int conflicting with a void return type, but there's also fortran compiled with f2c creating conflicting types compared to fortran adapters generated with f2py, which those patches to f2py in numpy fixed. Did the f2py patches port across okay to new numpy?

rth:
Thanks for the pointers @joemarshall !

>  they were mostly because if something is multiply defined with slightly different signatures, even with all the stuff turned on, it outputs symbols in two versions, one is just the normal one, the other is with some kind of type-conflict label in front of the name, I can't remember exactly what it was, something like '$type-conflict$name or similar.

Do you remember where you saw those conflicting symbols exactly? So far I don't see any such symbols after running `llvm-nm` (or `llvm-objdump -t') on `.o` files. I guess it could happen when linking multiple .o into an .so, but symbols appear to be stripped there.

> Did the f2py patches port across okay to new numpy?

They do apply without modifications, but it's also possible that some of the additional function were added there meanwhile which would need adjusting.

rth:
>  but it's also possible that some of the additional function were added there

No as far as I can tell, the only changes in files patched by `numpy/patches/make-int-return-values.patch` between `v1.15.4...v1.17.5` are minor code refactoring so it should still apply without modification.

joemarshall:
You can see the multiple symbols in Chrome, if you open up the wasm instances loaded, you can go into them without source code mapping and see the raw wasm symbol tables for each wasm item. You have to work out which wasm thing relates to the function that is crashing, I think I mainly did that through stopping the debugger on the invoke call in JavaScript and double clicking on the function that it was trying to call.

rth:
Indeed. Thanks!

So with debug flags the stacktrace is,
```
TypeError: Cannot read property 'apply' of undefined
    at obj.<computed> (pyodide.asm.js:formatted:6814)
    at PyUFunc_GenericFunction (<anonymous>:wasm-function[6466]:0x63a721)
    at ufunc_generic_call (<anonymous>:wasm-function[6527]:0x64e457)
    at byn$fpcast-emu$ufunc_generic_call (<anonymous>:wasm-function[14350]:0x842702)
    at _PyObject_MakeTpCall (pyodide.asm.wasm:wasm-function[739]:0x1cc5ce)
    at pyodide.asm.wasm:wasm-function[2744]:0x2b8814
    at _PyEval_EvalFrameDefault (pyodide.asm.wasm:wasm-function[2737]:0x2b597f)
    at byn$fpcast-emu$_PyEval_EvalFrameDefault (pyodide.asm.wasm:wasm-function[17137]:0x6cb8e0)
    at _PyEval_EvalCodeWithName (pyodide.asm.wasm:wasm-function[2733]:0x2b06de)
    at PyEval_EvalCode (pyodide.asm.wasm:wasm-function[2732]:0x2af8ac)
```
as far as I can tell  some those wasm definition cannot be accessed in Chrome because the number of disassembled WASM lines is truncated to 1M with the message "text is truncated due to size".  Tried to patch and re-compile the chrome devtools as suggested [here](https://diverto.hr/en/blog/2020-08-15-WebAssembly-limit/). However in my case it was then not compatible with Chrome 88, probably some version mismatch.

Anyway I think the exception happens in https://github.com/emscripten-core/emscripten/blob/55ce3d9a2582324ab839af01f04de72b4f727345/src/library_dylink.js#L472 and we probably need better debugging information when `resolveSymbol` returns undefined there.

joemarshall:
Oh,

I downloaded the build artifacts from circleci and run a web server against them.

The problem is that npy_get_floatstatus_barrier and npy_clear_floatstatus_barrier aren't created on wasm build. 

There's a comment deep in the numpy code in:
https://github.com/numpy/numpy/blob/maintenance/1.17.x/numpy/core/src/npymath/ieee754.c.src
```
/*
 * Define floating point status functions. We must define
 * npy_get_floatstatus_barrier, npy_clear_floatstatus_barrier,
 * npy_set_floatstatus_{divbyzero, overflow, underflow, invalid}
 * for all supported platforms.
 */
```
which says that this needs defining for all platforms, of which wasm isn't one. It isn't defined so it breaks the build.

You can see this if you stick a breakpoint on the middle line of:
```
if (!resolved) {
                        resolved = moduleExports[sym]
                    }
```
in pyodide.asm.js, that makes it stop on the symbol name that isn't defined when you do import numpy. The problem symbol is `npy_clear_floatstatus_barrier`

apply in pyodide.asm.js should really error with a function name if the function is missing rather than just bailing out with an exception, maybe worth a patch to upstream emscripten for that.

For pyodide, it needs a patch to https://github.com/numpy/numpy/blob/maintenance/1.17.x/numpy/core/src/npymath/ieee754.c.src to make it have a wasm version of these functions. Personally I'd try just a function that looks at a volatile variable as it does on the sun implementation, that would force optimization to split things before and after the call.

joemarshall:
having said that, there is an else that says
```
/* General GCC code, should work on most platforms */
```
so I'm not sure why that doesn't either a)bomb out on compile or b)create an export in the build...

rth:
Aww, that probably means that my patch `packages/numpy/patches/fix-ieee754.patch` is not good. Normally with https://github.com/numpy/numpy/pull/12124 we should have used definitions from `fenv.h` in Musl however for some reason it failed to find them, and so I reverted to the version before that PR in the patch but probably missed something.

> apply in pyodide.asm.js should really error with a function name if the function is missing rather than just bailing out with an exception, maybe worth a patch to upstream emscripten for that.

Definitely.

joemarshall:
https://numpy.org/devdocs/release/1.15.0-notes.html says that 
npy_clear_floatstatus_barrier
and the other barrier function is new in 1.15.0, so I guess those two are missing from the patch

rth:
The CI now is green, it was indeed an issue with `npy_get_floatstatus_barrier` and `npy_clear_floatstatus_barrier`. The current version of the patch i.e. always return 0, same as in 1.15.4 means we will still be experiencing https://github.com/iodide-project/pyodide/issues/156

Issue #636, opened 2020-04-15
Cannot catch pyodide.open_url not found exceptions


mauricelam:
I am trying to use pyodide.open_url, and hit an issue when doing do. If the URL I am requesting is not found, the javascript side throws an exception, but I cannot catch it from the Python side. In fact, when that happens, the entire runPython call seems to exit on the javascript side.

```js
try {
pyodide.runPython(`
import pyodide
try:
  pyodide.open_url("http://doesnotexist.abcdef")
finally:
  print("This is not printed out")
print("This is not printed out either")
`)
} catch (e) {
  console.log('Javascript can catch the exception though', e);
}
```

![Screen Shot 2020-04-14 at 8 09 12 PM](https://user-images.githubusercontent.com/1264702/79294849-073a3d80-7e8c-11ea-8cf8-d0bdf47196e3.png)

More generally, I think any exceptions throw by code `import js` is not available on the Python side and leads to confusing behavior.

BTW this project is really cool! Thanks for all the hard work!

rth:
This is now fixed, thanks to @hoodmane's work on the FFI. The example,
```py
try:
  pyodide.open_url("http://doesnotexist.abcdef/")
except Exception as err:
  print('Exception raised')
  print(type(err).__name__, err)
finally:
  print("This is printed out")
...
Exception raised
JsException NetworkError: A network error occurred.
This is printed out
```
works as expected.  Closing. Thanks for the report.

PR #880, opened 2020-12-17
Await jsproxy

hoodmane:
Implemented `__await__` on `JsProxy`. 
This is a work in progress towards #245. In particular, this is related to a suggestion in @oeway's comment:
https://github.com/iodide-project/pyodide/issues/245#issuecomment-744072568

If this were in Python the definition would be:

```python
def __await__(self):
   x = yield self
   return x
```

The code is a WIP still: it does not correctly handle the case where you `await` a `JsProxy` that is not a `Promise`. I'm not sure what the best way to handle this case is. Probably the best thing to do would be to add `hiwire_is_promise` using the `Promise.resolve(x) === x` check and raise an error if `hiwire_is_promise` returns false.

PR #891, opened 2020-12-18
Raise Exceptions when js2python is applied to javascript errors

### LINKED ISSUE #888: JsProxy does not handle Error correctly
Example:
```python
from js import Error
e = Error.new("Hi")
raise e
```
Raises: `TypeError: exceptions must derive from BaseException`.
This came up in https://github.com/iodide-project/pyodide/pull/880#pullrequestreview-555341317. I will try to fix this.

hoodmane:
This is intended to resolve #888, necessary for PR #880 to work correctly.
Work in progress.

Currently there is a serious issue: if the Javascript `Error` was created without `new`, this will crash the Pyodide runtime. I have absolutely no idea why.


hoodmane:
I think this is ready now @oeway.

oeway (commented):
Looks good!

I have two questions related to this: 1) have you figured why it's crashing when using Error without new? 2) if we do a round trip from js to python then back to js, will we get the same error object?

src/type_conversion/jsproxy.c
```diff
@@ -544,8 +645,54 @@ JsProxy_AsJs(PyObject* x)
...
+
+  module = PyImport_ImportModule("asyncio");
+  if (module == NULL) {
+    goto fail;
+  }
+
+  asyncio_get_event_loop = PyObject_GetAttrString(module, "get_event_loop");
+  if (asyncio_get_event_loop == NULL) {
+    goto fail;
+  }
```
rth, line 679: That's unrelated to error conversions, isn't it?

Also some `_Py_IDENTIFIER` in the beginning of this file.

hoodmane: @oeway Yes, I figured that out and fixed it. I was incorrectly using a C api and it caused crazy undefined behavior. It should have always crashed, I got unlucky with this `Error` vs `new Error` thing.

hoodmane: Right, I'm not that great at git and this is bleedover from the `await_jsproxy` PR #880.

hoodmane:
>1) have you figured why it's crashing when using Error without new? 
Yes I was incorrectly using a C Python API causing undefined behavior. This particular symptom was bad luck. It is fixed.

> 2) if we do a round trip from js to python then back to js, will we get the same error object?
Yes. See test here:
https://github.com/iodide-project/pyodide/blob/ad90a07b5ea6003e2947f0486641292cb53ac42a/src/tests/test_pyodide.py#L41

oeway:
Great, thank you!

hoodmane:
@rth I think this is ready to be merged unless you have further concerns?

src/type_conversion/jsproxy.c
```diff
@@ -544,8 +638,42 @@ JsProxy_AsJs(PyObject* x)
...
+  _Exc_JsException.tp_base = (PyTypeObject*)PyExc_Exception;
+
+  PyObject* module;
+  PyObject* exc;
+
+  module = PyImport_ImportModule("pyodide");
+  if (module == NULL) {
+    goto fail;
+  }
+  if (PyObject_SetAttrString(module, "JsException", Exc_JsException)) {
```
rth, line 668: Could you please add a comment why we need to do this?

rth (commented):
Thanks for working on this @hoodmane and thank you for the review @oeway !

Two comments below, otherwise LGTM.

Please also add an entry to `docs/changelog.md`.

src/tests/test_pyodide.py
```diff
@@ -20,3 +23,39 @@ def test_find_imports():
         )
     )
     assert set(res) == {"numpy", "scipy", "six", "matplotlib"}
+
+
+def test_javascript_error(selenium):
```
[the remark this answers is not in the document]
hoodmane: Moved to `test_type_conversions.py`.

src/type_conversion/jsproxy.c
hoodmane: Yes. (This is in case someone wants to catch `JsException`.)

hoodmane:
I also added an entry for `JsException` in the api docs. To be clear, I'm envisioning code like:

```pyodide
from js import js_async_func
from pyodide import JsException
async def wrapper(arg):
   try:
      await js_async_func(arg)
   catch JsException:
      raise ValueError(f"Unknown thunk {arg}")
```

rth:
Could you please resolve conflicts following #909 ? Otherwise should be good to merge assuming CI passes.

I have a doubt, would this address https://github.com/iodide-project/pyodide/issues/636 or not?

hoodmane:
This does not fix #636. To do this we should:.
* Add `hiwire_setexc(id)`: this will call `PyExc_SetObject(JsProxy_new_error(id))`.
* Adjust `hiwire_call` and `hiwire_call_member` to use try + catch and call `hiwire_setexc(hiwire_new(err))` for any caught error.

Once this is merged I can implement that. 

Error conversion in the other direction is also not very good: currently if python code throws an error, it prints stuff to stdout directly, but it should instead be possible to catch the error in javascript. To get the errors printed to stdout, we should explicitly catch them and call some separate print function.

hoodmane:
I think the title "Raise Exceptions when js2python is applied to javascript errors" is misleading: `js2python` doesn't raise an error at all, it returns an exception that can be raised. These are very different things:

f raises an error:
```python
def f():
  raise Exception("hi")
```

f returns an error:
```python
def f():
  return Exception("hi")
```

rth:
I had to shorten it as it was too long for a commit message. Didn't mean to rename the issue only the commit. I agree that's not ideal.

hoodmane:
Thanks!

Issue #737, opened 2020-08-04
Importing dynamically written module gives ModuleNotFoundError


rumkeller:
In #419, @mdboom suggests the following way of importing a dynamically created file (without the two commented lines):
```python
import sys
sys.path.insert(0, '.')

# with open("bar.txt", "w") as fd:
#  fd.write("bar")

with open("module.py", "w") as fd:
   fd.write("foo = 42")

import module
module.foo
```

As is, this works fine.

However, with the two lines uncommented this yields `ModuleNotFoundError: No module named 'module'`. This is unexpected, as creating `bar.txt` seems unrelated to the rest of the code.

Adding `import os; print(os.listdir("."))` just before the import shows that both files were indeed created.

Also, with the order of the two `with` blocks swapped it does work as expected.

Tested with v0.15.0 on Firefox 78.0.2.

rth:
Thanks for the report. I can reproduce in Firefox, and it's indeed very puzzling.

casatir:
In [Basthon](https://basthon.fr), users need to import their own modules like this.

With this issue, Pyodide seems limited to import one file per directory. This suggests a workaround : put each module in a separate folder. This is the workaround strategy implemented in Basthon (it should be released soon). Maybe clearer with this code snipet:
```python
import os
import sys

loaded_modules_root = os.path.join("/", "loaded_modules")

for m in ('my_module.py', 'my_other_module.py'):
    root = os.path.join(loaded_modules_root, os.path.splitext(m)[0])
    os.makedirs(root, exist_ok=True)
    with open(os.path.join(root, m), 'w') as f:
        f.write("a = {}\n".format(len(m)))
    sys.path.insert(0, root)

import my_module
print(my_module.a)

import my_other_module
print(my_other_module.a)
```

@rth: does this helps to start investigations?

rth:
Thanks @casatir that's certainly an interesting and puzzling behavior. Wrote a test case in https://github.com/iodide-project/pyodide/commit/dbc0251ccb28ffe20df883c04162a200e497b8a7 (last assert fails there) . What's really puzzling is that `os.listdir`  produces the correct list of files.

It as though there is some per folder cache in the python import system that's not updated when new files are added. We'll need to dig deeper into the Python import system mechanism to understand what is happening.

[...]

PR #893, opened 2020-12-18
modifiy MEMFS timestamp handling to support better caching

nopid:
PR implementing https://github.com/iodide-project/pyodide/issues/737#issuecomment-747722765

Modify `emscripten` filesystem behavior to look more POSIXish: the timestamp of a directory is modified each time a file is created or deleted inside the directory.

Relates to upstream discussion https://github.com/emscripten-core/emscripten/issues/11454
A long time upstreamable solution would be to implement a filesystem handling three different kinds of timestamps.

Partially solves #737

### DIFF
--- a/emsdk/patches/posixish_memfs.patch
+++ b/emsdk/patches/posixish_memfs.patch
@@ -2,7 +2,7 @@ Modify MEMFS to update timestamp of parent directory on file creation and remova
 More POSIXish compliant for a one timestamp fs, behaves like a modification date.
 
 --- a/emsdk/fastcomp/emscripten/src/library_memfs.js    2019-08-19 18:11:40.000000000 +0200
-+++ b/emsdk/fastcomp/emscripten/src/library_memfs.js	2020-12-18 14:07:14.000000000 +0100
++++ b/emsdk/fastcomp/emscripten/src/library_memfs.js	2020-12-23 12:22:38.000000000 +0100
 @@ -88,6 +88,7 @@
        // add the new node to the parent
        if (parent) {
@@ -11,14 +11,15 @@ More POSIXish compliant for a one timestamp fs, behaves like a modification date
        }
        return node;
      },
-@@ -213,12 +214,15 @@
+@@ -213,12 +214,16 @@
          }
          // do the internal rewiring
          delete old_node.parent.contents[old_node.name];
-+        old_node.parent.timestamp = Date.now()
++        var now = Date.now();
++        old_node.parent.timestamp = now;
          old_node.name = new_name;
          new_dir.contents[new_name] = old_node;
-+        new_dir.timestamp = old_node.parent.timestamp;
++        new_dir.timestamp = now;
          old_node.parent = new_dir;
        },
        unlink: function(parent, name) {
@@ -27,7 +28,7 @@ More POSIXish compliant for a one timestamp fs, behaves like a modification date
        },
        rmdir: function(parent, name) {
          var node = FS.lookupNode(parent, name);
-@@ -226,6 +230,7 @@
+@@ -226,6 +231,7 @@
            throw new FS.ErrnoError({{{ cDefine('ENOTEMPTY') }}});
          }
          delete parent.contents[name];
--- a/emsdk/patches/posixish_memfs.patch
+++ b/emsdk/patches/posixish_memfs.patch
@@ -1,0 +1,37 @@
+Modify MEMFS to update timestamp of parent directory on file creation and removal.
+More POSIXish compliant for a one timestamp fs, behaves like a modification date.
+
+--- a/emsdk/fastcomp/emscripten/src/library_memfs.js    2019-08-19 18:11:40.000000000 +0200
++++ b/emsdk/fastcomp/emscripten/src/library_memfs.js	2020-12-18 14:07:14.000000000 +0100
+@@ -88,6 +88,7 @@
+       // add the new node to the parent
+       if (parent) {
+         parent.contents[name] = node;
++        parent.timestamp = node.timestamp;
+       }
+       return node;
+     },
+@@ -213,12 +214,15 @@
+         }
+         // do the internal rewiring
+         delete old_node.parent.contents[old_node.name];
++        old_node.parent.timestamp = Date.now()
+         old_node.name = new_name;
+         new_dir.contents[new_name] = old_node;
++        new_dir.timestamp = old_node.parent.timestamp;
+         old_node.parent = new_dir;
+       },
+       unlink: function(parent, name) {
+         delete parent.contents[name];
++        parent.timestamp = Date.now();
+       },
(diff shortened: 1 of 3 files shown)


casatir:
Great job @nopid! Your solution looks quite straightforward indeed. Maybe the breaking use case of [#737 (comment)](https://github.com/iodide-project/pyodide/issues/737#issue-673026287) should be tested.

nopid:
Thanks @casatir. I added a test (based on @rth test) from #737.

rth (commented):
Thanks for working on this @nopid ! I think it could be still worth a try to upstream this. As far as I understand it's a step toward better POSIX compliance without adding additional metadata per node. Would you mind opening a PR at emscripten to see how they react to it? 

If emscripten doesn't accept it, we have been trying hard to remove emscripten patches, but this is also very useful. We could vendor `emsdk/fastcomp/emscripten/src/library_memfs.js` which is not very long, but it looks like there isn't currently a way to use a custom file system in emscripten https://github.com/emscripten-core/emscripten/issues/9372#issuecomment-528124876

Also could you please check if we can now remove the `invalidate_caches` call from `src/pyodide.js`  and `packages/micropip/micropip/micropip.py`?

nopid:
There is still something I don't get with the caches...

This test passes:
```python
import sys
sys.path.append('.')

with open("bar.txt", "w") as fd:
   fd.write("bar")

with open("module.py", "w") as fd:
   fd.write("foo = 42")

import module
module.foo
```

This one always fails (insert path instead of append):
```python
import sys
sys.path.insert(0,'.')

with open("bar.txt", "w") as fd:
   fd.write("bar")

with open("module.py", "w") as fd:
   fd.write("foo = 42")

import module
module.foo
```
and this one fails in Firefox and passes in Chrome (adding some busy loop):
```python
import sys
sys.path.insert(0, '.')

with open("bar.txt", "w") as fd:
   fd.write("bar")

y=0
for i in range(1000000):
  y=i

with open("module.py", "w") as fd:
   fd.write("foo = 42")

import module
module.foo
```

nopid:
Thanks for the review @rth, once the second part of #737 (see there for a new comment and test case) is solved we will be able to discuss `invalidate_caches`. In the meantime, I will submit the `emscripten` filesystem patch upstream.

emsdk/patches/posixish_memfs.patch
```diff
@@ -0,0 +1,37 @@
...
+       return node;
+     },
+@@ -213,12 +214,15 @@
+         }
+         // do the internal rewiring
+         delete old_node.parent.contents[old_node.name];
++        old_node.parent.timestamp = Date.now()
+         old_node.name = new_name;
+         new_dir.contents[new_name] = old_node;
++        new_dir.timestamp = old_node.parent.timestamp;
```
dalcde, line 21: I think it would be clearer if this said `new_dir.timestamp = Date.now()` instead.

dalcde:
It would be nice to have a test that directly tests the patch by checking modification timestamps. These can be added to `emsdk/tests/` (you'll have to merge master to see this), and ideally are simple C code. This would make it easier/faster to ensure the patches aren't messed up when upgrading emscripten.

emsdk/patches/posixish_memfs.patch
nopid: I think it is better if both the old and the new dir have the same timestamp after rename.
Would you prefer if I use a new variable now = Date.now() and use it to update both old_node.parent and new_dir timestamps?

dalcde: That would work too

nopid:
Thanks @dalcde for the review. I introduced the variable as discussed and a C test in `emsdk/tests/` inspired by the already existing tests of the same directory.

emsdk/tests/test_memfsmtime.py
```diff
@@ -0,0 +1,78 @@
...
+                "-s",
+                "EMULATE_FUNCTION_POINTER_CASTS=1",
+            ],
+            check=True,
+            env=common.env,
+        )
+        out = subprocess.run(
+            ["node", "a.out.js"], capture_output=True, check=True, env=common.env
+        )
+        assert out.stdout == b"0 tmpdir\n1 tmpdir\n1 tmpdir/foo.py\n2 tmpdir\n"
```
dalcde, line 78: I don't have strong opinions either way but it might be cleaner to run the assert statements inside the C code.

emsdk/tests/test_memfsmtime.py
```diff
@@ -0,0 +1,78 @@
...
+            )
+
+        subprocess.run(
+            [
+                "emcc",
+                "-s",
+                "MAIN_MODULE=1",
+                "main.c",
+                "-s",
+                "EMULATE_FUNCTION_POINTER_CASTS=1",
```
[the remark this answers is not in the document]
nopid: done

emsdk/tests/test_memfsmtime.py
nopid: done

emsdk/tests/test_memfsmtime.py
```diff
@@ -0,0 +1,78 @@
...
+                "MAIN_MODULE=1",
+                "main.c",
+            ],
+            check=True,
+            env=common.env,
+        )
+        out = subprocess.run(
+            ["node", "a.out.js"], capture_output=True, check=False, env=common.env
+        )
+        assert out.returncode == 0
```
dalcde, line 78: Why this instead of simply doing `check=True`? I think using `check=True` gives better error messages when it fails.

nopid: With check = True, I get the following error message (when inserting a dummy assertion inside C code):
```
E               subprocess.CalledProcessError: Command '['node', 'a.out.js']' returned non-zero exit status 1.                                                                                               
                                                                                                                                                                                                             
/usr/local/lib/python3.8/subprocess.py:512: CalledProcessError
```

With check=False and the assertion on out.returncode I can see the assertion error that failed in the C code:
```
>           assert out.returncode == 0                                                                                                                                                                       
E           AssertionError: assert 1 == 0                                                                                                                                                                    
E            +  where 1 = CompletedProcess(args=['node', 'a.out.js'], returncode=1, stdout=b'Assertion failed: 1 == 2, at: main.c,54,main\n', st..._memfsmtime0/a.out.js:32475:5)\n    at runCaller (/tmp/pytest-of-root/pytest-12/test_memfsmtime0/a.out.js:32375:19)\n').returncode                                                                                                                                     
                                                                                                                                                                                                             
emsdk/tests/test_memfsmtime.py:80: AssertionError 
```

dalcde: Hmm I seem to recall having the opposite experience. lgtm then.

dalcde (approved):

rth:
Thank you!

PR #894, opened 2020-12-18
PKG Fix Python 3.8 warnings in mpmath

casatir:
Python 3.8 `mpmath` warnings are [corrected upstream](https://github.com/fredrik-johansson/mpmath/issues/547) but no release since 2018...

This PR patch mpmath to remove this warnings.

This partially solves #877.


packages/mpmath/patches/fix-warnings.patch
```diff
@@ -0,0 +1,21 @@
+diff --git a/mpmath/ctx_mp_python.py b/mpmath/ctx_mp_python.py
```
rth, line 1: Could you please add a comment about the fact that this should be removed with the next mpmath release? Otherwise LGTM.

rth (approved):
Thanks!

Issue #892, opened 2020-12-18
Can't properly get a JS object back to Python


casatir:
(This comes from [this comment](https://github.com/iodide-project/pyodide/pull/875#issuecomment-747369299))

When you get a Python object on the JS side (e.g. `runPython[Async]`'s result), it seems impossible to get it back into Python. As an illustration, compare the result of the two following code snippets:

```javascript
// returns "[object Object]"
pyodide.runPython("import js ; js.window.x = {1: 2} ; repr(js.window.x)")
// returns "{1: 2}"
pyodide.runPython("x = {1: 2} ; repr(x)")
```

Long story short,

``` javascript
pyodide.runPython("repr({1: 2})")
// is not equivalent to:
pyodide.pyimport("repr")(pyodide.runPython("{1: 2}"))
```

This is closely related to #780.

hoodmane:
The specific problem here is that the definition of `_python2js_dict` is completely wrong: it performs a copy rather than proxying the dict. This is not the expected behavior at all: if `x` is a dict then `y=x` stores a reference to the same dict and `y[2] = 3` changes `x`. I think it's a major violation of expectations that `js.window.x = x` stores a copy of x, so that `js.window.x[2]=5` leaves `x` unchanged.

I think we need more tests for this kind of stuff.

https://github.com/iodide-project/pyodide/blob/641ce219b8d41d4f708f610106797c91abe1d734/src/type_conversion/python2js.c#L202

hoodmane:
In `js2python` only immutable types get copied, whereas in `python2js` everything gets copied unless we can't figure out what it is. If we change `python2js` to only copy immutable types and proxy everything else then `js2python` and `python2js` will be inverses to each other like they ought to be and issues like this will be resolved.

It might be a good idea to expose an api to attempt to copy a jsproxy into a native python object and an api to copy a pyproxy into a native js object but these should require specific invocation.

rth:
This now works as expected and produces identical results (thanks to @hoodmane's efforts),
```js
pyodide.runPython("repr({1: 2})")
pyodide.globals.get("repr")(pyodide.runPython("{1: 2}"))
```

Issue #780, opened 2020-10-30
js->py type conversions for arrays/lists not using native types


rth:
A round trip of a Python list (for instance of strings) to JS and back produces a JsProxy object instead of the original list,

**Example**
In Python,
```py
x = ['a', 'b']
```

then in JS,
```js
var x = pyodide.pyimport('x')
```
and finally in Python,
```py
import js
type(js.x)  # produces a JsProxy, instead of a list.
```

oeway:
I am a bit confused, do you mean:
```
import js
type(js.x)  # <---------type(js.x) instead of type(js)?
```

rth:
Yes, indeed it was a typo.

hoodmane:
This was resolved by #1167. I just added tests for this in #1188.

PR #898, opened 2020-12-19
Reorganized type conversions tests into separate files

hoodmane:
There are currently several issues (#780, #892) about the type conversions not working correctly. As far as I can tell there are no tests of the type conversions? The type conversions are a core part of Pyodide which deserve careful testing. I added a few tests.

Currently `test_list_from_py` and `test_dict_from_js` are failing which correspond to the examples respectively in issue #780 and in issue #892. I will open an issue with a proposal for a modification to the type conversions to fix these issues.


rth:
Type conversion tests are generally in https://github.com/iodide-project/pyodide/blob/master/src/tests/test_python.py and those for numpy arrays are in packages/numpy/test_numpy.py.

Could you please check that the added tests are not redundant with those in `test_python.py` and we could indeed move conversion specific tests to a dedicated test file.

hoodmane:
Ah of course, that is way more comprehensive, I just didn't know where to look for it.

> we could indeed move type conversion specific tests to a dedicated test file.

I will remove my redundant tests and copy the tests from `test_python` into `test_typeconversions`.

hoodmane:
I removed the redundant tests for literals.

I added two new files: specifically for testing pyproxy and jsproxy, since these correspond one-to-one with the logic implemented in the c source files of the same name.

I also removed the new failing tests, since they contradict existing tests. Once this gets merged I can add a new PR changing the test suite to expect the behavior I described in proposal #900.

rth:
> Currently [..] test_dict_from_js are failing

It seems to be passing it CI? Do you understand why?

Otherwise LGTM, thanks!

hoodmane:
> It seems to be passing it CI?

Yeah I changed that:

> I also removed the new failing tests, since they contradict existing tests. Once this gets merged I can add a new PR changing the test suite to expect the behavior I described in proposal #900.

rth:
Thanks @hoodmane!

PR #901, opened 2020-12-20
BLD Clean up CLAPACK build

dalcde:
The biggest change is that most of `0001-hacks-for-WebAssembly-build.patch` is eliminated, and the remaining is split up into individual single-purpose patches with corresponding documentation.

Most of `0001-hacks-for-WebAssembly-build.patch` is in fact unnecessary. We can continue to produce `.a` files instead of `.bc` files, and all we have to do to make it build is to set `$(ARCH) = emar` and `$(RANLIB) = emranlib`. In fact, we don't have to set them to anything, since the point of emmake is to set these for us (except emmake uses `$(AR)` so we set `$(ARCH) = $(AR)`).

There are places where `ar` and `cc` etc. are used directly instead of the variables; we use `sed` to fix this in the parent `Makefile` instead of patching.

In addition, there are also the following changes:
 - Change the build system to build CLAPACK-WA/INSTALL as well. We should always have built it, but we managed to get away with not doing so due to a bug in emscripten that is fixed in 1.38.43. See https://github.com/iodide-project/pyodide/pull/843#issuecomment-744428471
 - Generate `F2CLIB/libf2c/arith.h` properly instead of setting it to an empty file.
 - Remove a seemingly unnecessary patch (https://github.com/iodide-project/pyodide/pull/901/commits/945f06c46e4c88c2a12bffd5dd7252751ccc617e)


packages/CLAPACK/patches/0001-hacks-for-WebAssembly-build.patch
```diff
@@ -75,114 +50,22 @@ index 57eff0d..d9a42da 100644
  	getenv_.o signal_.o s_stop.o s_paus.o system_.o cabs.o ctype.o\
```
dalcde, line 50: @rth what is the purpose of patching `MISC` in the line above? Is it another duplicate symbol issue?

dalcde:
I can't reproduce the pandas build failure locally, and I don't see how my changes could have affected it

dalcde:
Many of the patches ought to be upstreamed, but I haven't managed to figure out how to do so. The email at https://www.netlib.org/clapack/ is unreachable.

packages/CLAPACK/Makefile
```diff
@@ -19,17 +19,25 @@ $(SRC)/lapack_WA.bc:  $(SRC)/Makefile
...
-$(SRC)/Makefile:
+clapack.tgz:
 	wget -q -O clapack.tgz http://www.netlib.org/clapack/clapack.tgz
 	md5sum --quiet --check checksums || (rm -f clapack.tgz; false)
+
+$(SRC)/Makefile: clapack.tgz
 	tar xzf clapack.tgz
 	rm -rf $(SRC)
 	mv CLAPACK-$(CLAPACK_VERSION) $(SRC)
+	chmod -R o+rx $(SRC)
```
rth, line 32: Why do we need this?

rth:
> Many of the patches ought to be upstreamed, but I haven't managed to figure out how to do so.

I don't think CLAPACK is still an active project. People usually have a Fortran compiler and use an optimized BLAS (https://github.com/iodide-project/pyodide/issues/227) with [LAPACK](https://github.com/Reference-LAPACK/lapack) and I think we should try move into that direction in the future as well. Right now the performance of numpy related code is also not great because we use the reference BLAS.

rth (approved):
Generally LGTM (side for the `chmod` use), thanks for doing this work!

packages/CLAPACK/Makefile
dalcde: Well we don't *need* it, but it makes life much easier if you build in docker, since it would not be you who owns the content and it defaults to `550`.

rth: Right, fair enough. Maybe add a comment about it? Otherwise feel free to merge.

packages/CLAPACK/patches/0001-hacks-for-WebAssembly-build.patch
rth: That's likely yes.

dalcde: Well it seems to work without the patch, so I'll leave it out

Issue #676, opened 2020-05-29
Interrupting execution


danzvara:
Hi all! 

I have a question, is there a possibility of interrupting running execution in Pyodide? I saw there's nothing like that in the API, i'm mostly curious about whether it's even possible to implement

Thanks!

rth:
On a regular system you would interrupt Python execution by sending a `SIGINT` signal via `Ctrl+C`. That's not supported in emscripten and that signal [would be ignored](https://github.com/emscripten-core/emscripten/blob/eeacee1bb8f66561a08a8d3fa6703b1d326dd640/src/library_signals.js#L17) (also because pyodide is single threaded). 

One solution is to run pyodide in a webworker (cf docs) then [terminate the worker](https://developer.mozilla.org/en-US/docs/Web/API/Web_Workers_API/Using_web_workers#Terminating_a_worker) when needed. Though webworkers support is still experimental in pyodide. In particular, https://github.com/iodide-project/pyodide/issues/441 is a known issue.

stefnotch:
Would threading (https://github.com/iodide-project/pyodide/issues/237) allow for interrupting execution?

rth:
> Would threading (#237) allow for interrupting execution?

It should allow to, though I haven't investigated it much. There are still some blockers for that issue.

hoodmane:
You can do this without any changes to Pyodide itself using `sys.settrace` as long as your browser supports `SharedArrayBuffer`. Maybe there is another approach to this using `postMessage` / the message handlers, but I don't know how to do that.

My current implementation seems to degrade performance by a factor of about 10. I am hoping that implementing the trace function in C, I can improve on that.

My setup:
(1) Pyodide runs on a webworker
(2) When the main thread wants the worker to execute a command, it sends a 4 byte SharedArrayBuffer as a cancellation token to the worker.
(3) The worker uses sys.settrace() to install a trace function. This trace function checks the value in the SharedArrayBuffer every 10,000 ticks. If the value is nonzero, then it can throw a KeyboardInterrupt (or do something else depending on a message encoded in the value).

Some code snippets: the command for the main thread to ask the pyodide worker to execute something:
```javascript
    execute(code){
        const interrupt_buffer = new Int32Array(new SharedArrayBuffer(4));
        const uuid = uuid4();
        const execution = new Execution(interrupt_buffer);
        this.executions[uuid] = execution;
        this._postMessage("execute", uuid, {code, interrupt_buffer});
        return execution;
    }
```
[...]

Issue #831, opened 2020-12-06
Release v0.16.0


rth:
I think we could make the v0.16.0 release  in the next few days.

Remaining thing that we would need to merge IMO are,
 - [x] Smaller docker image (https://github.com/iodide-project/pyodide/pull/815)
 - [x] make sure `console.html` downloads packages from jsdelivr when deployed inside CI (https://github.com/iodide-project/pyodide/pull/855)
 - [ ] update packages to avoid warnings with Python 3.8 (https://github.com/iodide-project/pyodide/issues/877)
 - [x] update release notes (https://github.com/iodide-project/pyodide/pull/813)

Please comment if you think something else should be included as well.

dalcde:
I would be interested in streamlining the process of building
third-party C-based libraries for pyodide. I believe this involves
publishing our patched `wasm-opt` binary and python headers, as well as
some helper scripts to glue things together.

rth:
> I believe this involves
publishing our patched `wasm-opt` binary and python headers, as well as
some helper scripts to glue things together.

A pre-built docker image https://github.com/iodide-project/pyodide/issues/656 wouldn't be sufficient for that use case?

dalcde:
A pre-built docker image contains much more than that, and in particular
is much larger.

rth:
Responded in more detail in https://github.com/iodide-project/pyodide/issues/814

We can certainly wait a bit more, but I think there have been a lot of changes since the last release in May and we should make a release soon. There is nothing that prevents us of doing another release not too long after if need be.

rth:
The plan for now is to make a release tomorrow evening. I'll take a week off from contributing after that.

ehfd:
@rth could you enable discussions in the repository?

rth:
@ehfd I have experimentally enabled Github Discussions. Let's see if it could help discussion with the community. For me the separation between Github Issues, Discussion posts and StackOverflow questions is still not very clear in the case of pyodide.

PR #902, opened 2020-12-20
Improved keyboard interrupt handling backend

hoodmane:
This is a replacement for pyodide-interrupts.
It fixes two quirks in pyodide-interrupts: It interacts correctly with sys.settrace, and it uses signals correctly.
As a result, if you use `signal.signal(SIGINT, some_handler)` then `some_handler` will be called when an interrupt is triggered.

In order to interact correctly with `sys.settrace`, I have to monkey patch `sys.settrace`. In particular, I copied the entire definition of `sys.settrace` into my code and added one extra line to call my extra interrupt handling. This has the disadvantage that it pins a specific version of `sys.settrace` from Python 3.8.2.

If people prefer I could keep this as a separate package rather than integrating it into Pyodide core.


hoodmane:
The monkey patch of `sys.settrace` was really bugging me. When I was looking into the inner workings of the `signal` package, I found `Py_AddPendingCall` which sounded like what I wanted. And indeed, `Py_AddPendingCall` can be used for this purpose avoiding all monkey patching. Currently one of my tests is failing but I'm confident this approach can be used and it's a significant improvement.

hoodmane:
Looks like the failing test passes on CI, I think it's something screwed up about the state of my local copy and doing a fresh build will make it go away.

src/type_conversion/interrupts.c
```diff
@@ -0,0 +1,228 @@
...
+  //   goto fail;
+  // }
+
+  // func = PyCFunction_NewEx(&settrace_methoddef, (PyObject*)module, name);
+  // if (func == NULL) {
+  //   goto fail;
+  // }
+  // if (PyObject_SetAttrString(module, "settrace", func)) {
+  //   goto fail;
+  // }
```
rth, line 181: Commented code should be removed.

src/pyodide.py
```diff
@@ -7,6 +7,31 @@
...
+def interrupt_buffer(callback):
+    """
+    Periodically polls ``callback``. If ``callback`` returns a nonzero value, triggers a ``SIGINT`` signal.
+    By default, the signal handler for ``SIGINT`` raises a ``KeyboardException``, but using the ``signals`` package this can be changed.
+
+    If ``callback`` returns a value that cannot be interpreted as an integer or if an exception is triggered inside of ``callback``
+    then the exception is allowed to propogate but the interrupt buffer is set to ``None``.
+
+    Args:
+        callback -- a zero argument function which returns an int. If it returns a nonzero value, triggers ``SIGINT``.
```
rth, line 27: We use numpy docstring format https://numpydoc.readthedocs.io/en/latest/format.html#docstring-standard 

So it needs a 1 line summary, and then something like
```
Parameters
----------
callback : callable
    a zero ...
```

src/type_conversion/interrupts.c
```diff
@@ -0,0 +1,228 @@
...
+             "can be changed."
+             "If ``callback`` returns a value that cannot be interpreted as an "
+             "integer or if an exception is triggered inside of ``callback``"
+             "then the exception is allowed to propagate but the interrupt "
+             "buffer is set to ``None``."
+             "If called with ``None``, interrupt polling is turned off."
+             "\n"
+             "Args:\n"
+             "   callback -- a zero argument function which returns an int. If "
+             "it returns a nonzero value, triggers ``SIGINT``.\n");
```
rth, line 75: Same comment about docstring format as above.

rth (commented):
Thanks @hoodmane ! You would need to fix merge conflicts following #909 . Maybe we could put it under `src/pyodide-py/pyodide/interrupt.py`

> If people prefer I could keep this as a separate package rather than integrating it into Pyodide core.

I'm not opposed to merging this into pyodide code. However I won't have the bandwidth to review this in detail before the release #831 . So unless @dalcde looks into it, it might be easier to update it as a separate package for now.

Also please update the changelog.

dalcde:
Two comments:
1. Can we benchmark the performance impact of using this? I recall that the previous `settrace` method has essentially undetectable performance impact. Is the same true for `Py_AddPendingCall`?
2. Can we add interrupt functionality to `webworker.js`? Hopefully we can integration test the interrupt functionality this way (e.g. with a sleep loop). This will let us close #676

hoodmane:
> Can we benchmark the performance impact of using this?

We should do that before merging this.

`Py_AddPendingCall` acquires a lock (always open when there is only one thread), pushes the callback onto a stack, and then does a relaxed atomic store into a signal that there are pending callbacks so that is effectively free.

Oh uhh and then on each step through the `ceval` main loop it calls my callback 32 times because `NPENDINGCALLS` is `32` and the callback recursively sets itself. For most of these 32 calls my callback just decrements a counter and returns, the cost of this is probably (?) still negligible. Annoyingly there is no clear way to defer the next call to after the next python instruction is executed.

Currently I know of four possible entrypoints to add code to the main loop:
`PyEval_SetTrace`, `PyEval_SetProfile`, setting a custom `PyInterpreterState.eval_frame` and `Py_AddPendingCall`. 

Of these four options, the only one that mixes well with all possible other code is `Py_AddPendingCall`. However, `eval_frame` is a good option because it is not exposed to the Python user directly like `sys.settrace` and `sys.setprofile` are. Custom `eval_frame` is currently only used for JIT, and I don't expect us to support JITs for Pyodide any time soon...

hoodmane:
Oh `PyInterpreterState.eval_frame` doesn't work because it only runs once per stack frame which isn't enough: we want keyboard interrupts to work even if the interpreter is in an infinite loop with no function calls in it, such as `while True: pass`

hoodmane:
> I'm not opposed to merging this into pyodide code. However I won't have the bandwidth to review this in detail before the release #831.

This is more of a fine tuning details PR, the current `pyodide-interrupts` code works for practical purposes so there is no rush on including it until we get it right.

hoodmane:
I am now thinking that the best approach to this might be to add a patch to `ceval.c` to expose an extra callback. @dalcde @rth Would you guys object to that? You have any thoughts?

hoodmane:
Closed in favor of #1148.

PR #905, opened 2020-12-21
Test emsdk patches

dalcde:


dalcde:
I am unable to produce a test for `dynCall_so.patch`. It appears that mdboom had a similar difficulty when trying to upstream the patch.

PR #907, opened 2020-12-21
Fixed error message when attempting to pyimport a nonexistent value

hoodmane:
Before this PR, the error message thrown by `pyodide.pyimport("missing_name")` was `Uncaught Error: No exception type or value`. Switching out `PyDict_GetItem` for `PyDict_GetItemWithError` fixes the problem. Now the error message is `KeyError: 'missing_name'`.

### DIFF
--- a/src/tests/test_python.py
+++ b/src/tests/test_python.py
@@ -1,5 +1,5 @@
+from selenium.common.exceptions import WebDriverException
 import time
-
 import pytest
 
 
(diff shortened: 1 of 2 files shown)


hoodmane:
Apparently this didn't work. I guess it'd be good to explicitly specify an error message there anyways.

rth:
Thanks for investigating this! I'll mark it as a Draft PR meanwhile.

hoodmane:
I think I will close this in favor of #923 which reimplements `pyimport` in terms of `pyodide.globals`. See also #926.

PR #908, opened 2020-12-21
Add tests for dyncall_so patch

dalcde:


dalcde:
It would be nice if there were an easy way to make sure the tests fail *without* the patch. It is slightly tricky because the dyncall_so test would fail for a different reason if the `emulate_pointer.patch` is not applied, so if we simply don't patch at all we won't actually test `dynCall_so.patch`.

dalcde:
Another issue with this test is that without the patch, the error message we get is not the same as the one we get from `test_brentq`, but they seem similar in spirit.

Issue #900, opened 2020-12-19
Proposal: Changes to type conversions


hoodmane:
Currently there are a couple of issues (#780, #892) with type conversions. I have an idea for how to fix this but it would be a pretty significant change so we should discuss first.

The source of both #780 and #892 is that rather than proxying Python dicts and lists into javascript, `python2js` copies them into js objects and Arrays. On the other hand, `js2python` proxies most mutable types. This means that converting javascript ==> python ==> javascript gives back the original object more often but converting python ==> javascript ==> python doesn't give the original type for lists and dicts.

I think the current problematic copy conversions are:
* `dict` ==> javascript
* `list` ==> javascript
* `typedarray` ==> python

Also, we convert python `bytes` object to a `Uint8ClampedArray` which when imported back to python turns into a `memoryview`. We should add a flag to the `Uint8ClampedArray` to indicate that it came from a python `bytes` object and convert it to `bytes` on return to python rather than a `memoryview`.

I propose that we change the apis so that all copies are explicit. This is easier to do for `dict` and `list` than for `typedarray`. I think for `dict` and `list` we can just remove lines [255-258 in python2.js.c](https://github.com/iodide-project/pyodide/blob/master/src/type_conversion/python2js.c#L255) and add `deep_copy_into_native_js` and `shallow_copy_into_native_js` methods to `PyProxy`.

I think for `typedarray` we should import it as `WrappedTypedArray` and provide methods like `copy_into_memoryview`, `copy_from_memoryview` and `clone_as_memoryview`. This way direct modification of the imported object actually modifies the Javascript `TypedArray`, it's clear from the api that the `memoryview` is a copy and so modifying it won't change the original `TypedArray` and there is an api for writing the result of a computation back into the original `TypedArray`.

Because tuples are immutable, we could consider adding back a shallow copy to convert tuples to javascript arrays. We could add a marker to the returned array to indicate that it came from a python tuple. This would mean that the round trip python ==> js ==> python would return a different tuple with identical contents, but because tuples are immutable, tuples with identical contents are exchangeable.

[The docs say:](https://pyodide.readthedocs.io/en/latest/type_conversions.html#class-instances)
> The most notable limitation is that while Python has distinct ways of accessing attributes and items (x.foo and x[foo]), Javascript conflates these two concepts.

I think here the correct model is the Javascript object `Map`.  So: 
* `dict["x"]` <==> `map.get("x")` 
* `dict["x"] = 2` <==> `map.set("x", 2)`
[...]

PR #912, opened 2020-12-22
proxy dict and list into javascript and add deep_to_js api to them.

hoodmane:
This is a WIP towards #900. I've already changed too much stuff to really belong in one pull request though. This proxies `dict` and `list` into javascript and adds a `deep_to_js` api. The `deep_to_js` api is exactly the same code as the old `py2js` translation.
This whole PR is on py2js conversion, I've left js2py alone.

### Miscellaneous changes

* Split `py2js` into `py2js_copy` and `py2js_nocopy`. All but one of the original `py2js` calls I changed to the nocopy version. TODO: The remaining `copy` call really should be a shallow copy.
* If `py2js_copy` would do any copying on a `PyProxy` I add `deep_to_js`.
* In `conftest.py` use `deep_to_js` on the result whenever it is available. This prevents a large number of test breakages when results are communicated back to selenium. Todo: a better solution is to implement `toJSON` on `PyProxy` (it should call `json.dumps`).
* Refactored hiwire (see PR #920 which is only the changes to hiwire)
*  I switched the `objects` map for `hiwire` and for `pyproxy` from an `object` to a `Map` because maps ["perform better in scenarios involving frequent additions and removals of key-value pairs"](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map#Objects_vs._Maps).

### Changes to `PyProxy`

I completely rewrote `PyProxy`, to separate it into three layers: the innermost layer is written in C and wraps the Python abstract protocols that I use with hiwire. This is boilerplate. The middle layer communicates between javascript and hiwire, also boilerplate. The innermost layer defines a new set of javascript protocols. Then I do reflection on the `PyObject`, and based on which abstract protocols it implements / which concrete types it extends, I add appropriate javascript protocols in batches.
(This strategy is inspired by the implementation of `PyType_Ready`.)

If the object is a "sequence" (i.e., it implements `__getitem__` and `__getitem__(some_string)` raises a `TypeError`, not a `KeyError`) then I allow the elements of the object to be accessed like `a[4]`. For other `__getitem__` implementers, the items must be accessed through methods imitating the javascript `Map` api. 

I think the old test suite still passes, so currently the behavior of `PyProxy` is pretty similar to the old behavior. I think this design will allow significant ease of use improvements in the future.

### TODOs
py2js:
* `shallow_to_js`
* Conversion to js should handle a few more standard conversions, like python set ==> js Set.
* Fix py2js memorybuffers. What used to happen was correct, I just need to patch it back in. Presumably some numpy tests will be failing right now.
* Implement `toJSON` on `PyProxy`
* Use [FinalizerRegistry](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/FinalizationRegistry) to register destructors.
 
js2py:
* `deep_to_py` and `shallow_to_py`.
* Wrapped `TypedArray`s.

Also, I need to rewrite the docs about type conversions.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -7,7 +7,7 @@ defaults: &defaults
   environment:
     - EMSDK_NUM_CORES: 4
       EMCC_CORES: 4
-      PYODIDE_BASE_URL: https://cdn.jsdelivr.net/pyodide/dev/full/
+      PYODIDE_BASE_URL: https://cdn.jsdelivr.net/pyodide/v0.16.0/full/
 
 jobs:
   lint:
@@ -84,10 +84,6 @@ jobs:
           no_output_timeout: 1200
           command: |
             ccache -z
-            # Mark CLAPACK as not built as it's currently incorrectly marked as built
-            # when running make with PYODIDE_PACKAGES
-            # TODO: this should be resolved with #713
-            make -C packages/CLAPACK clean
             make
             ccache -s
 
@@ -148,6 +144,16 @@ jobs:
           name: test
           command: |
             tools/pytest_wrapper.py packages/test* packages/*/test* -v -k chrome -n 2
+
+  test-emsdk:
+    <<: *defaults
+    steps:
+      - attach_workspace:
+          at: .
+      - run:
+          name: test
+          command: |
+            pytest emsdk/tests -v
 
   test-python:
     <<: *defaults
@@ -256,6 +262,9 @@ workflows:
           filters:
             tags:
               only: /.*/
+      - test-emsdk:
+          requires:
+            - build-core
       - test-python:
           filters:
             tags:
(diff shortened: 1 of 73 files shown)


rth:
Thanks for doing this work @hoodmane ! I'll try to review beginning of January, but it's true that splitting this into smaller PRs would be easier to review.

hoodmane:
I will try to split it up a bit. I think for instance my hiwire changes should be part of a separate PR.

What's the best way to deal with dependencies between PRs though? I'm not an expert in this version control stuff. What if I just redo these same changes but make a bunch of marked intermediate commits that pass CI?

rth:
> What's the best way to deal with dependencies between PRs though? 

The easiest might be to keep this PR with everything, then make one or two small independent PRs, and merging master into this branch when they are merged. Then making a few more independent PRs, etc. That you won't have to deal with dependencies too much.

PR #920, opened 2020-12-22
Some internal improvements to hiwire

hoodmane:
This is split off from PR #912 to try to make reviewing easier. This is purely maintenance / internal code improvement, it should have no observable effect to the end user.

## Hiwire API changes
I exposed the hiwire constants as `Module.hiwire.ERROR`, etc. I replaced all magic hiwire numbers in javascript code by references to `Module.hiwire.SOME_CONSTANT`. I added a convenience method `hiwire_bool` because there seems to be several places where the code says `x ? hiwire_true() : hiwire_false()` and `hiwire_bool(x)` seems simpler.

Renamed:
* `Module.hiwire_new_value` ==> `Module.hiwire.new_value`
* `Module.hiwire_get_value` ==> `Module.hiwire.get_value`
* `Module.hiwire_dec_ref` ==> `Module.hiwire.dec_ref`

Most of the diff for this is from these renames.

## Internal changes to hiwire.c
 I used a Map for hiwire objects internally instead of a js object because maps ["perform better in scenarios involving frequent additions and removals of key-value pairs"](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map#Objects_vs._Maps).
I tried to improve the clarity of the distinction between private and public methods.

### DIFF
--- a/src/type_conversion/hiwire.c
+++ b/src/type_conversion/hiwire.c
@@ -2,132 +2,11 @@
 
 #include "hiwire.h"
 
-EM_JS(void, hiwire_setup, (), {
-  // These ids must match the constants above, but we can't use them from JS
-  var hiwire = { objects : {}, counter : 1 };
-  hiwire.objects[-2] = undefined;
-  hiwire.objects[-3] = true;
-  hiwire.objects[-4] = false;
-  hiwire.objects[-5] = null;
-
-  Module.hiwire_new_value = function(jsval)
-  {
-    var objects = hiwire.objects;
-    while (hiwire.counter in objects) {
-      hiwire.counter = (hiwire.counter + 1) & 0x7fffffff;
-    }
-    var idval = hiwire.counter;
-    objects[idval] = jsval;
-    hiwire.counter = (hiwire.counter + 1) & 0x7fffffff;
-    return idval;
-  };
-
-  Module.hiwire_get_value = function(idval) { return hiwire.objects[idval]; };
-
-  Module.hiwire_decref = function(idval)
-  {
-    if (idval < 0) {
-      return;
-    }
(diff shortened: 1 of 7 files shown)


src/type_conversion/hiwire.c
```diff
@@ -2,51 +2,115 @@
...
+    // }
+    // if(jsval === null){
+    //   return Module.hiwire.NULL;
+    // }
+    // if(jsval === true){
+    //   return Module.hiwire.TRUE;
+    // }
+    // if(jsval === false){
+    //   return Module.hiwire.FALSE;
+    // }
```
rth, line 71: Please remove commented code, unless it's part of the comment?

src/type_conversion/hiwire.c
```diff
@@ -2,51 +2,115 @@
...
+    _hiwire.objects.set(idval, jsval);
+    _hiwire.counter = (_hiwire.counter + 1) & 0x7fffffff;
     return idval;
   };
 
-  Module.hiwire_get_value = function(idval) { return hiwire.objects[idval]; };
+  Module.hiwire.get_value = function(idval)
+  {
+    if (!idval) {
+      throw new Error("Argument to hiwire_get_value is undefined");
```
rth, line 84:
```suggestion
      throw new Error("Argument to hiwire.get_value is undefined");
```

rth (commented):
Thanks @hoodmane, it's certainly more readable this way. Two minor comments otherwise LGTM.

src/type_conversion/hiwire.c
hoodmane: Okay will remove.

PR #875, opened 2020-12-16
Improving console.html

casatir:
`console.html`is a great tool. It's a showcase for Pyodide and [the devel version](https://pyodide-cdn2.iodide.io/dev/full/console.html) allows developers to track issues and make quick tests.

Unfortunately, IMHO, it lacks some important features like:

1. sync behavior (do not set next prompt until current command finished, e.g. with imports)
2. full stdout/stderr support (e.g. import warnings are not visible)
3. correct result representation with `repr` (type `"1"` and it shows `1`, type `globals()` and it shows `[object Object]`)

While 1. and 2. are addressed in this PR, 3. seems tricky, at least with the current setup (JQuery's terminal + `copy.py` interactive console + `runPythonAscyn`) but not impossible (see [Basthon](https://basthon.fr)).

Getting correct representations for basic types (like strings) are addressed in this PR. Difficulty to get correct `repr` for complex objects comes from the fact that, when you get a Python object on the JS side (e.g. runPython[Async]'s result), it seems impossible to get it back into Python. As an illustration, compare the result of the two following code snippets:

```javascript
// returns "[object Object]"
pyodide.runPython("import js ; js.window.x = {1: 2} ; repr(js.window.x)")
// returns "{1: 2}"
pyodide.runPython("x = {1: 2} ; repr(x)")
```
Long story short, 

```javascript
pyodide.runPython("repr({1: 2})")
// is not equivalent to:
pyodide.pyimport("repr")(pyodide.runPython("{1: 2}"))
```

Any idea to fully address 3.with the current setup?

Should I edit changelog ?

Closes https://github.com/iodide-project/pyodide/issues/897

PR #923, opened 2020-12-23
Rewrote runpython

hoodmane:
`runpython.c` used it's own ad-hoc wrappers for `pyodide_py.eval_code` and `pyodide_py.find_imports` duplicating the code in `pyproxy` for calling python functions. Also, it was confusingly organized and leaked a reference to the `pyodide` module.

I rewrote it to use the pyproxy wrappers which eliminated about half the contents of `runpython.c`.

In conversation with @casatir on #875 we've talked about how we want different things out of a repl. Before this PR it's absolutely unclear how you could implement a different version of `runPython` or `runPythonAsync` to customize your repl. Not even monkeypatching `eval_code` works because `runpython.c` stored the values of `eval_code` and `find_imports` at startup and doesn't care about your monkeypatch. Now monkeypatching `eval_code` behaves as expected.

Better yet, though, would be to allow a user to define their own version of `runPython` and `runPythonAsync`. To this end, I factored `runPythonAsync` into a new public API `loadPackagesForCode` and then a call to `runPython`. So now it's trivial for anyone to implement their own similar method.

```javascript
Module.runPythonAsync = async function(code, messageCallback, errorCallback){
    await Module.loadPackagesForCode(code, messageCallback, errorCallback);
    return Module.py_pyodide.eval_code(code, Module.globals);
};
```
The definition of `runPython` is also a one-liner now:
```javascript
Module.runPython = function(code){
   return Module.py_pyodide.eval_code(code, Module.globals);
};
```
This resolves both:
* "What if I want my own version of `eval_code`?" and
* "What if I want a different namespace than `pyodide.globals`?"


src/pyodide.js
```diff
@@ -330,6 +331,37 @@ var languagePluginLoader = new Promise((resolve, reject) => {
...
+        if (name in packageNames) {
+          packages.add(name);
+        }
+      }
+      if (packages.size) {
+        await loadPackage(Array.from(packages.keys()), messageCallback,
+                          errorCallback);
+      }
+    }
+  }
```
dalcde, line 355: Putting a semicolon here would most likely significantly improve the formatting.

hoodmane: We should really try to switch to a js linter for javascript code rather than using clang-format.

src/pyodide.js
```diff
@@ -331,31 +331,35 @@ var languagePluginLoader = new Promise((resolve, reject) => {
   Module.preloadedWasm = {};
   let isFirefox = navigator.userAgent.toLowerCase().indexOf('firefox') > -1;
 
-  Module.runPython = function(
-      code) { return Module.py_pyodide.eval_code(code, Module.globals); };
+  // clang-format off
```
dalcde, line 334: Do you ever turn this back on?

hoodmane: Nope, I forgot about that. I kinda think we should just disable it on the whole file and get a damn js linter.

dalcde:
I find that clang-format formats reasonably as long as you don't forget
your semi-colons.

hoodmane:
And also to disable / reenable around `===` and `!==`.

hoodmane:
Actually, it would probably be good to keep a `debugRunPython` like the original `runPython` in case someone is working on `pyproxy` and breaks `pyproxy.call`.

dalcde:
Would you consider the formatting changes as in https://github.com/dalcde/pyodide/commit/0b042b920b07dbb915561d5836fb6be1c42e7f3e ?

docs/js-api/pyodide_pypyodide.md
```diff
@@ -0,0 +1,3 @@
+(js_api_pyodide_pypyodide)=
+# pyodide.pypyodide
```
dalcde, line 2: This should be py_pyodide

src/pyodide.js
```diff
@@ -330,6 +331,41 @@ var languagePluginLoader = new Promise((resolve, reject) => {
   Module.preloadedWasm = {};
   let isFirefox = navigator.userAgent.toLowerCase().indexOf('firefox') > -1;
 
+  // clang-format off
+
+  Module.runPython = function(code) { return Module.py_pyodide.eval_code(code, Module.globals); };
+
+  Module.loadPackagesForCode = async function(code, messageCallback, errorCallback) {
+    let imports = Module.py_pyodide.find_imports(code);
+    if (imports.length) {
```
dalcde, line 340: No a huge deal but this zero check isn't really necessary imo

src/pyodide.js
```diff
@@ -330,6 +331,41 @@ var languagePluginLoader = new Promise((resolve, reject) => {
...
+    }
+  };
+
+  Module.pyimport = function(name){
+    return Module.globals[name];
+  };
+
+  Module.runPythonAsync = async function(code, messageCallback, errorCallback) {
+    await Module.loadPackagesForCode(code, messageCallback, errorCallback);
+    return Module.py_pyodide.eval_code(code, Module.globals);
```
dalcde, line 364: Call `runPython` here instead?

hoodmane:
Thanks for the review @dalcde!

hoodmane:
> Would you consider the formatting changes as in dalcde@0b042b9 ?

I pulled that.

dalcde:
If the goal of runPythonDebug is to facilitate debugging, it might make sense to go one step further, and simply return a string representation of the returned value, so that we also don't have to go through python2js.

hoodmane:
How about `console.log("runPythonDebug result repr:\n", repr(result))` and then do the `py2js` and return the result?

Failure in `js2py` before running the code is annoying for debugging. Failure in conversion back afterwards is useful.

hoodmane:
Okay I think that is better.

src/type_conversion/runpython.c
```diff
@@ -1,26 +1,35 @@
...
 
+int
 _runPythonDebug(char* code)
 {
   PyObject* py_code;
   py_code = PyUnicode_FromString(code);
   if (py_code == NULL) {
-    return pythonexc2js();
+    fprintf(stderr, "runPythonDebug -- error occurred converting argument:\n");
+    PyErr_Print();
```
dalcde, line 20: Do you not want to return early here?

dalcde (commented):
Some tests for `runPythonDebug` would be nice, though I don't think it would catch the problem below

src/type_conversion/runpython.c
hoodmane: I'm confused. Locally I see:
```C
  if (result == NULL) {
    fprintf(stderr, "runPythonDebug -- error occurred\n");
    PyErr_Print();
    return HW_UNDEFINED;
  }
```
and:
```
Switched to branch 'runpython-fix'
Your branch is up to date with 'origin/runpython-fix'.
```

hoodmane:
I sorta figure with a debug entrypoint that it isn't particularly important to test it, but it is probably a good idea.

hoodmane:
@rth Could you review this? I think this one should merge next among my PRs. I am still planning to add a couple of tests to ensure that `runPythonDebug` doesn't always crash.

dalcde:
I'm happy with everything here apart from tests for runPythonDebug.

src/pyodide.js
```diff
@@ -330,6 +331,41 @@ var languagePluginLoader = new Promise((resolve, reject) => {
   Module.preloadedWasm = {};
   let isFirefox = navigator.userAgent.toLowerCase().indexOf('firefox') > -1;
 
+  Module.runPython = code => Module.py_pyodide.eval_code(code, Module.globals);
+
+  // clang-format off
+  Module.loadPackagesForCode = async function(code, messageCallback, errorCallback) {
```
rth, line 337: Maybe,
```suggestion
  Module.loadPackagesFromImports = async function(code, messageCallback, errorCallback) {
```
to be more consistent with `pyodide.find_imports`? Also it's not really loading packages used in the code, only imports matter.

rth (commented):
`docs/custom_runPython.md` is an empty file and should be removed?

Minor comments below, also let's indeed add at least one test for `_runPythonDebug`. Otherwise LGTM. Thanks @hoodmane , it's a nice improvement!

hoodmane:
We should do an audit of added APIs before the next release and think about naming / call style / appropriateness.

hoodmane:
I think I fixed the problem with `runPythonDebug`, the issue was sloppy reference counting in `runpython_init`. It was a good idea to add `runPythonDebug` though because my wrong reference counting probably would trigger problems with `runPython` under some circumstances.

I will rearrange `runpython_init` to use the try / finally / fail pattern which makes it easier to reference count correctly.

src/type_conversion/runpython.c
```diff
@@ -1,165 +1,132 @@
...
-  return runpython_init_js() || runpython_init_py() || runpython_finalize_js();
+fail:
+  if (PyErr_Occurred()) {
+    PyErr_Print();
+  }
+  Py_CLEAR(pyodide_py);
+  Py_CLEAR(globals);
+  hiwire_decref(pyodide_py_proxy);
+  hiwire_decref(globals_proxy);
+  return -1;
```
rth, line 131: Wouldn't +1 be a more classical exit status for errors?

src/type_conversion/runpython.c
```diff
@@ -1,165 +1,132 @@
...
-{
-  return runpython_init_js() || runpython_init_py() || runpython_finalize_js();
+fail:
+  if (PyErr_Occurred()) {
+    PyErr_Print();
+  }
+  Py_CLEAR(pyodide_py);
+  Py_CLEAR(globals);
+  hiwire_decref(pyodide_py_proxy);
+  hiwire_decref(globals_proxy);
```
rth, line 130: If we fail to load, do we actually have to care about memory leaks?

rth (commented):
Minor comment otherwise LGTM.

Also as a side comment, better not to use mentions in commits (`@user`) as it creates unnecessary notifications, and also it will ping the user if the commit is pushed to any branch on Github in the future. Just `user` is fine.

src/type_conversion/runpython.c
```diff
@@ -1,165 +1,132 @@
...
 }
 
-EM_JS(int, runpython_init_js, (), {
-  Module._runPythonInternal = function(pycode)
+EM_JS(int, runpython_init_js, (JsRef pyodide_py_proxy, JsRef globals_proxy), {
+  Module.pyodide_py = Module.hiwire.get_value(pyodide_py_proxy);
+  Module.globals = Module.hiwire.get_value(globals_proxy);
+
+  // Use this to test python code separate from pyproxy.apply.
+  Module.runPythonDebug = function(code)
```
dalcde, line 47: Would it make more sense to move this to pyodide.js alongside the definition of `runPython`?

src/type_conversion/runpython.c
hoodmane: I'm trying to be consistent with the Python conventions as much as possible. The majority of Python functions that return ints have the convention that they `Return 0 on success or -1 on failure.` The one annoying exception is that the argument parsing functions return 1 on success and 0 on fail. There are also special cases for functions for which -1 is a valid return value, in that case you have to check `PyErr_Occurred()`.

hoodmane: In particular, python module initialization functions have this calling convention.

hoodmane:
> better not to use mentions in commits ... Just `user` is fine.

Thanks for the tip rth!

src/type_conversion/runpython.c
hoodmane: No, but I think it's best to be as consistent as possible. For one thing, this conveys extra information about my beliefs about the reference counting. If it's right, then it's one more example to help someone reading the code to understand how to do things. If it's wrong, it's an extra chance to notice that I have a misconception which might cause problems in more important code paths.

src/type_conversion/runpython.c
hoodmane: I prefer to keep it here: the definition of `runPython` doesn't do explicit memory management, it leaves that to `PyProxy`, whereas `runPythonDebug` just does memory management / js <==> C conversion and then calls the C function.

src/type_conversion/runpython.c
hoodmane: Right I guess the point is that if the function returns a boolean it's `1` = true, `0` = false, `-1` = failed. And there are a lot of functions that naturally return a nonnegative integer, like the length of a list.

rth:
Thanks for doing this work @hoodmane !

hoodmane:
I'm having a blast, Pyodide is outrageously cool.

PR #925, opened 2020-12-23
Add apply-lints make recipe

hoodmane:
Added "apply-lints" recipe to Makefile that modifies files in place to apply lints.
I've been finding it quite annoying to track down the reason that "make lint" is failing: clang-format does not say which files cause the problem.


rth (approved):
Thanks!

Issue #713, opened 2020-07-07
Build C libraries with the Python build system


rth:
Currently numerous C libraries (e.g. lz4, libxml) and also some packages (e.g. six, parso)  are packaged with Makefiles in the root folder of pyodide. I this can be confusing for new contributors and is not sustainable long term, as it makes it hard to distinguish what is pyodide code and what are packages.

Related to https://github.com/iodide-project/pyodide/issues/38

I think we should,
 - [x] move all of those libraires under `packages/` folder (while keeping the use of Makefiles). This would be just a path change. Done in https://github.com/iodide-project/pyodide/pull/714
 - [ ] migrate these libraries to be defined with a `meta.yaml` and build them with the Python build system. This would allow managing dependencies better (e.g. build LAPACK only when scipy is built etc).

rth:
To give a more concrete example for step 1, consider [`packages/zlib/Makefile`](https://github.com/iodide-project/pyodide/blob/master/packages/zlib/Makefile), we would want to replace it with something along the lines of,

**packages/zlib/meta.yaml**
```yaml
package:
  name: zlib
  version: 1.2.11
source:
  sha256: c3e5e9fdd5004dcb542feda5ee4f0ff0744628baf8ed2dd5d66f8ca1197cb1a1
  url: https://zlib.net/zlib-1.2.11.tar.gz
build:
  script:
    - emconfigure ./configure  --prefix=$(SRC)
    - emmake make install -j ${PYODIDE_JOBS:-3} 
```

and then,
 1. adjust [pyodide_build/buildpkg.py](https://github.com/iodide-project/pyodide/blob/master/pyodide_build/buildpkg.py#L230) to these build script commands if they are available and fall back to the `python setup.py` command otherwise as it's done now.
 2. figure out how to pass environment variables such as `PYODIDE_JOBS` and also some paths variables (here `$SRC` which is the current dir). There is probably some standard approaches of doing this with conda `meta.yaml`. It would require reading the `meta.yaml` spec and maybe also browsing a few examples on conda-forge.
 3. Add `zlib` as a requirements to any package that needs it, here lxml.
 4. Remove the corresponding command form the [main Makefile](https://github.com/iodide-project/pyodide/blob/bc3ffaae427e35b9ecd9a6216fadad031a363a47/Makefile#L268).

**Note:** zlib is required by libxml, so it would likely need to be ported at the same time.  Or one could start with a different package, for instance lz4.

Issue #832, opened 2020-12-07
Use correct python cross-compilation terminology


dalcde:
What we call `host` and `target` is what python calls `build` and `host`. In particular, `host` has the completely opposite meaning.

rth:
I would say it's an issue with CPython terminology.

Host/target is fairly standard,

> In GCC world, every host/target combination ...

in Rust, https://rust-lang.github.io/rustup/cross-compilation.html

> When you first install a toolchain, rustup installs only the standard library for your *host* platform - that is, the architecture and operating system you are presently running. To compile to other platforms you must install other *target* platforms.

I also feel it's much easier to understand that build and host as well. Closing.

PR #927, opened 2020-12-23
Build C libraries via buildall.py

dalcde:
This addresses part of #713, by listing and building C libraries as dependencies. This means we automatically build CLAPACK when needed, and `lxml`'s C dependencies are not built if lxml is not. In particular, building "core" should now be faster.

The building itself is still performed by Makefile.

### DIFF
--- a/docs/new_packages.md
+++ b/docs/new_packages.md
@@ -162,12 +162,14 @@ This directory should contain (at least) two files:
   included in the `pyodide` repository.
 
 - `meta.yaml` that specifies metadata about the package. For C libraries, only
-  two options are supported:
+  three options are supported:
 
-  - `library/name`: The name of the library, which must equal the directory
+  - `package/name`: The name of the library, which must equal the directory
     name.
   - `requirements/run`: The dependencies of the library, which can include both
     C libraries and python packages.
+  - `build/library`: This must be set to `true` to indicate that this is a
+    library and not an ordinary package.
 
 After packaging a C library, it can be added as a dependency of a python
 package like a normal dependency. See `lxml` and `libxml` for an example (and
--- a/docs/new_packages.md
+++ b/docs/new_packages.md
@@ -149,6 +149,34 @@ A list of required packages.
 
 (Unlike conda, this only supports package names, not versions).
 
+## C library dependencies
+Some python packages depend on certain C libraries, e.g. `lxml` depends on
+`libxml`.
+
+To package a C library, create a directory in `packages/` for the C library.
+This directory should contain (at least) two files:
+
+- `Makefile` that specifies how the library should be be built. Note that the
+  build system will call `make`, not `emmake make`. The convention is that the
+  source for the library is downloaded by the Makefile, as opposed to being
+  included in the `pyodide` repository.
+
+- `meta.yaml` that specifies metadata about the package. For C libraries, only
+  two options are supported:
+
+  - `library/name`: The name of the library, which must equal the directory
+    name.
+  - `requirements/run`: The dependencies of the library, which can include both
+    C libraries and python packages.
+
+After packaging a C library, it can be added as a dependency of a python
+package like a normal dependency. See `lxml` and `libxml` for an example (and
+also `scipy` and `CLAPACK`).
+
+*Remark:* Certain C libraries come as emscripten ports, and do not have to be
+built manually. They can be used by adding e.g. `-s USE_ZLIB` in the `cflags`
+of the python package. See e.g. `matplotlib` for an example.
+
 ## Manual creation of a Pyodide package (advanced)
 The previous sections describes how to add a python package to the pyodide
 build.
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -84,10 +84,6 @@ jobs:
           no_output_timeout: 1200
           command: |
             ccache -z
-            # Mark CLAPACK as not built as it's currently incorrectly marked as built
-            # when running make with PYODIDE_PACKAGES
-            # TODO: this should be resolved with #713
-            make -C packages/CLAPACK clean
             make
             ccache -s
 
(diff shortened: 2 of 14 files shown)


dalcde:
Comparing to recent `master` builds, this should save ~3 minutes on compile time (since `build-packages` blocks on a few large packages, its build time should not be affected by this commit)

hoodmane:
You mean it will reduce build time by three minutes with `PYODIDE_PACKAGES="micropip"`? That sounds amazing!

dalcde:
It will. It should also reduce the whole CI time by 3 minutes. (the timing is based on CI times with ccache enabled; local build times may differ)

hoodmane:
Very much looking forward to that, I spend so much time waiting for it to compile.

dalcde:
I think at this point most of the build time is spent `./configure`ing CPython

hoodmane:
Are you talking about only fresh builds? I mostly am not doing fresh builds.

dalcde:
Well then it depends on which part you are rebuilding, obviously :P. I think this part of the code is unlikely to be rebuilt

hoodmane:
Right, I'm changing files in `src/typeconversions`.

hoodmane:
Still will be nice to get faster responses from CI.

rth:
Thanks @dalcde !

>   `library/name`: The name of the library, which must equal the directory name.

The classical way of doing this is to keep `package/name` but differentiate between build and run dependencies https://conda-forge.org/docs/maintainer/adding_pkgs.html#build-host-and-run

So if possibly I would rather we added those dependencies to a `requirements/build` section. These could also include in time other Python build dependencies (such as numpy when building scipy). Which means that we won't need to install all runtime dependencies when building  Python packages in the future.

**Note:** in the conda-forge vocabulary the proper section would be `requirements/host` but that's a bit confusing in the cross-compilation context, so I think `requirements/build` is fine.

It might take a bit more work, so if you don't want to look into it here, as a temporary solution adding a `build/library = True` flag would also be fine IMO.

In follow up PRs we could download, verify checksum and explicitly write the [build/script command](https://conda-forge.org/docs/maintainer/adding_pkgs.html#optional-bld-bat-and-or-build-sh) in meta.yaml (making Makefiles mostly unnecessary), but this is already a  very nice improvement.

dalcde:
On Wed, Dec 23, 2020 at 12:58:21AM -0800, Roman Yurchak wrote:
>Thanks @dalcde !
>
>>   `library/name`: The name of the library, which must equal the directory name.
>
>The classical way of doing this is to keep `package/name` but differentiate between build and run dependencies https://conda-forge.org/docs/maintainer/adding_pkgs.html#build-host-and-run

The differentiation I'm trying to make is that between something that is
genuinely a python package (something one can pip install), and a C
library. The difference between run and build dependencies is a separate
matter imo.

>So if possibly I would rather we added those dependencies to a `requirements/build` section. These could also include in time other Python build dependencies (such as numpy when building scipy). Which means that we won't need to install all runtime dependencies when building  Python packages in the future.

I considered doing this but I thought it would make more sense to do the
separation when we do it for python packages too.

>**Note:** in the conda-forge vocabulary the proper section would be `requirements/host` but that's a bit confusing in the cross-compilation context, so I think `requirements/build` is fine.

I think conda's terminology is consistent with cross-compilation
terminology, c.f. #832

>In follow up PRs we could download, verify checksum and explicitly write the [build/script command](https://conda-forge.org/docs/maintainer/adding_pkgs.html#optional-bld-bat-and-or-build-sh) in meta.yaml (making Makefiles mostly unnecessary), but this is already a  very nice improvement.

I agree

rth:
> The differentiation I'm trying to make is that between something that is
genuinely a python package (something one can pip install), and a C
library. The difference between run and build dependencies is a separate
matter imo.

Yes, you are right it's an additional differentiation. Do we really want to differentiate between Python and non Python packages in the build system in the long term? Conda doesn't.

For instance, for BLAS it could be a separate pyodide package, dynamically linked from multiple python packages (assuming we manage to make that work), and loaded independently via the regular dependency resolution mechanism with loadPackage. Otherwise it would mean we would need to vendor it in each package that uses it. And this case, one would add BLAS both to build and run dependencies. One could say that run = add to the dependency graph packages.json and build = don't. 

For now though, adding a binary flag `build/library` would be also fine I think.

> I think conda's terminology is consistent with cross-compilation
terminology, c.f. #832

OK, good to know.

dalcde:
Fair enough. I think there are enough planned changes in the pipeline
that what exactly we do now doesn't matter too much. It would be nice to
get more proper dynamic linking going on, e.g. we also have several
copies of zlib lying around.

rth:
Please add a changelog entry as well..

docs/new_packages.md
```diff
@@ -149,6 +149,34 @@ A list of required packages.
...
+- `Makefile` that specifies how the library should be be built. Note that the
+  build system will call `make`, not `emmake make`. The convention is that the
+  source for the library is downloaded by the Makefile, as opposed to being
+  included in the `pyodide` repository.
+
+- `meta.yaml` that specifies metadata about the package. For C libraries, only
+  two options are supported:
+
+  - `library/name`: The name of the library, which must equal the directory
+    name.
```
rth, line 168: This also needs updating..

rth:
> I think at this point most of the build time is spent ./configureing CPython

One way to improve that situation could be to say that CPython is just another dependency that is built with the python build system (as it's done in conda) and then at least some of the CPython dependencies (zlib, bzip2, sqlite) could have their configure steps run in parallel. + libxml etc can be built in parallel for non minimal builds.

Though that would likely require a significant amount of work.

dalcde:
On Wed, Dec 23, 2020 at 02:34:30AM -0800, Roman Yurchak wrote:
>> I think at this point most of the build time is spent ./configureing CPython
>
>One way to improve that situation could be to say that CPython is just another dependency that is built with the python build system (as it's done in conda) and then at least some of the CPython dependencies (zlib, bzip2, sqlite) could have their configure steps run in parallel. + libxml etc can be built in parallel for non minimal builds.

I think as it stands, the CPython dependencies will be built and
configured in parallel if we call `make -C cpython -j4`. The logs will
become fairly mangled though.

PR #928, opened 2020-12-23
Added minimal make recipe

hoodmane:
Added rule so that `make minimal` does `PYODIDE_PACKAGES="micropip" make`.


dalcde:
See also #658

hoodmane:
Alternatively, maybe this rule could be called "nopackages".
Relatedly: Is there a way to get docker to remember by repl history between sessions?

Makefile
```diff
@@ -305,3 +305,6 @@ FORCE:
 
 check:
 	./tools/dependency-check.sh
+
+minimal :
+	PYODIDE_PACKAGES="micropip" make
```
rth, line 310: LGTM, but there is something in your editor configuration that removes newlines at the end of files.  https://unix.stackexchange.com/a/18789

You can [use pre-commit](https://pyodide.readthedocs.io/en/latest/contributing.html#development-workflow) which [would fix it](https://github.com/iodide-project/pyodide/blob/41592637e6a5231d1fa74235ec5f46919c18e25d/.pre-commit-config.yaml#L7).

hoodmane: I installed pre-commit, but it always failed so I uninstalled it. Probably should open an issue.

hoodmane: Wow I love shit like this: "A file that is not empty and does not end with a newline is therefore not a text file." Very interesting.

hoodmane:
@dalcde Could you merge this? It would save me a fair amount of typing and @rth already signed off on it.

dalcde:
Can you add back the final newline?

hoodmane:
I guess this is another issue with `make lint` that it does not reject missing final end of line.

PR #929, opened 2020-12-23
Added fatal error messages to main

hoodmane:
If one of the init functions called in main fails, print a fatal error message. This should make it much easier to debug issues.
Example: If I add a name error to pyodide_py it now prints:

```
FATAL ERROR: Failed to initialize module runpython.
Error was triggered by Python exception:
  File "/lib/python3.8/site-packages/pyodide/__init__.py", line 3, in <module> 
    test pyodide.asm.js:8:57150
NameError: name 'test' is not defined
Python exception: 
SystemError: null argument to internal routine
```

### DIFF
--- a/src/main.c
+++ b/src/main.c
@@ -10,6 +10,24 @@
 #include "python2js.h"
 #include "runpython.h"
 
+#define FATAL_ERROR(args...)                                                   \
+  do {                                                                         \
+    printf("FATAL ERROR: ");                                                   \
+    printf(args);                                                              \
+    if (PyErr_Occurred()) {                                                    \
+      printf("Error was triggered by Python exception:\n");                    \
+      PyErr_Print();                                                           \
+    }                                                                          \
+  } while (0)
+
+#define TRY_INIT(mod)                                                          \
+  do {                                                                         \
+    if (mod##_init()) {                                                        \
+      FATAL_ERROR("Failed to initialize module %s.\n", #mod);                  \
+      return 1;                                                                \
+    }                                                                          \
+  } while (0)
+
 int
 main(int argc, char** argv)
 {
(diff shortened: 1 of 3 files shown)


rth (approved):
Thank you! More explicit error messages on failure will be helpful indeed!

src/main.c
```diff
@@ -10,6 +10,24 @@
...
+    }                                                                          \
+  } while (0)
+
+#define TRY_INIT(mod)                                                          \
+  do {                                                                         \
+    if (mod##_init()) {                                                        \
+      FATAL_ERROR("Failed to initialize module %s.\n", #mod);                  \
+      return 1;                                                                \
+    }                                                                          \
+  } while (0)
```
rth, line 29: Do we need the `while (0)` statements here and above?

rth:
Also note that's preferable not to force push in PR as it makes review harder. When new commits are added only those can be checked, while after a force push everything has to be re-checked. If you need to sync with master
```
git merge upstream/master
```
also works, it's a bit more explicit about what's going on.

hoodmane:
> preferable not to force push in PRs (particularly after a review) as it makes review harder

Okay thanks for the info. This is my first time contributing to a project with code reviews so any tips like this are very much appreciated!

What are my force pushes:
* Repeated lint fails: I figured having 20 commits "lint", "try lint again", "try lint again", "try lint again" isn't very helpful. This should be better after PR #925. Sometimes the issue is a missing `\\ clang-format-off` and I get into a feedback loop where the linter fails, then I run the lints and it breaks the build and then I fix the compile error and the lint fails again.

* I accidentally `git commit --amend` a bunch of stuff. You don't really get a good opportunity to cancel `git commit --amend` unless you delete the entire commit log, and I find this difficult to undo.

* I think there is a bug right now with `build-packages` that causes occasional failures. I've been doing an empty `git commit --amend; git push -f` to make it rerun. I guess instead `git commit --empty -m "rerun CI" ; git push` would do.

src/main.c
hoodmane: `do { stuff } while (0)` is the standard macro boilerplate to allow C macros to look like function calls:
https://gcc.gnu.org/onlinedocs/cpp/Swallowing-the-Semicolon.html#Swallowing-the-Semicolon

rth:
Yes, pre-commit should address a number of those. Don't hesitate to open issue if something is not working.

PR #932, opened 2020-12-23
Use pip3 to install stuff

hoodmane:
precommit doesn't work if you use python2

### DIFF
--- a/docs/contributing.md
+++ b/docs/contributing.md
@@ -8,7 +8,7 @@ See {ref}`building_from_sources` and {ref}`testing` documentation.
 
 For code-style the use of [pre-commit](https://pre-commit.com/) is also recommended,
 ```
-pip install pre-commit
+pip3 install pre-commit
 pre-commit install
 ```
 This will run a set of linters at each commit. Currently it runs yaml syntax validation and is removing trailing whitespaces.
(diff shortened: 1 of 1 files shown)


rth:
I'm not sure about this. Most projects I know indicate `pip install` in the documentation and it does imply Python 3 (e.g. https://numpy.org/install/). That's also the command given by PyPi e.g. https://pypi.org/project/black/ for Python 3 only projects. I would say it's up to the user/contributor to pick the right version of python, for instance by using a virtual environments or by using `python -m pip install` to be sure.

hoodmane:
Well `pip3` is defined in all of the python3 virtualenvs. Various linux utilities are written in python2 and need `pip` to refer to python2. With `pip3` either:
(1) they are in a python3 virtualenv or otherwise in a situation where pip points to python3. In that case pip3 also points to python3 and it works, or
(2) they are in a situation where pip points to python2 and pip3 points to python3, and it works.

With pip, situation (2) doesn't work. The goal of docs should be to make things less error prone when possible. Putting `pip3` reduces the chance of wasting time by accidentally installing something on python2 and then having it fail in a confusing way.

rth:
From https://packaging.python.org/tutorials/installing-packages/#installing-from-pypi

Which uses `pip install` everywhere,

> Due to the way most Linux distributions are handling the Python 3 migration, Linux users using the system Python without creating a virtual environment first should replace the python command in this tutorial with python3 and the pip command with pip3 --user. Do not run any of the commands in this tutorial with sudo: if you get a permissions error, come back to the section on creating virtual environments, set one up, and then continue with the tutorial as written.

So `pip3 install pre-commit` is also not correct as when executed literally it would try to install packages in the system python (probably with sudo), which should be avoided. So that's why all OSS projects that I'm aware of use `pip install` with the assumption that the user works in some kind of virtual environment. I need to add documentation on how to build in a conda env.

hoodmane:
> when executed literally it would try to install packages in the system python (probably with sudo), which should be avoided
That doesn't seem correct. I just tested it and inside a python3 virtualenv `pip3` and `pip` do the same thing:
```sh
>>> which pip
/usr/bin/pip
>>> which pip3
/usr/bin/pip3
>>> source virtualenv/bin/activate
>>> which pip
/home/hood/virtualenv/bin/pip
>>> which pip3
/home/hood/virtualenv/bin/pip3
```

PR #934, opened 2020-12-23
Py2js variants

hoodmane:
More work split off from #912 and fleshed out. 
I implemented variants of `python2js`: `python2js_once` which does a shallow copy, and `python2js_minimal` which only copies python objects that are immutable. The original version I renamed to `python2js_deep`. Currently none of these new functions are used in public APIs and so this PR should have no observable effect on the behavior of public APIs.

The difference between the variants is in how they handle dict and list. `python2js_minimal` does not copy dict or list, but proxies them. `python2js_shallow` copies the object if it is a list or dictionary, doing `python2js_minimal` to all of the fields of the object. 

Also, I did some infrastructure work to allow testing of C functions which is in the separate PR #937.


hoodmane:
@rth There is exactly one test failing here, it's a numpy test:
```python
selenium.run("x = np.array([['string1', 'string2'], ['string3', 'string4']])")
assert selenium.run_js("return pyodide.pyimport('x').length") == 2
```
and it fails at `Internal error: Invalid type to convert to array buffer.` inside of `js2python_buffer`: 
https://github.com/hoodmane/pyodide/blob/6c6476eb79cf5144833a2488475f8537ecc46f9e/src/core/python2js_buffer.c#L474

Any clue what's going on? I'm completely unfamiliar with numpy, though I should probably try harder to diagnose this.
(A bit awkward that it throws a type error with message saying "Internal Error:", that isn't the usual way of Python error messages.)

hoodmane:
Superceeded by #1128.

PR #937, opened 2020-12-24
Maint: Add a feature gate for C tests, C test miniframework

hoodmane:
This is some infrastructure work to allow testing of C functions, split off from #934.
The idea is that any private apis that need to be exposed for testing will be added to `pyodide.TestEntrypoints`. In many cases there will need to be extra code to interface between pytest and the C code that is only needed for testing, so I added a `TEST` feature gate. Currently TEST is always set and I only expose ``pyodide.TestEntrypoints` if `window.TEST_PYODIDE` is truthy.

In the future, it would be good to have the CLI build a version with `TEST` and a version without `TEST`. Then we can adjust the test suite to do integration testing of public APIs using the build without the TEST symbols and unit tests of private APIs using the build with the TEST symbols. When we update CLI to distribute a version built with `TEST` unset, then we should remove the `window.TEST_PYODIDE` gate.

This is part of my ongoing attempt to improve the C code in `src/typeconversions`.


hoodmane:
@dalcde Do you understand where / how `EM_JS` exposes the functions it defines? I looked at [the explanation of EM_JS here](
https://github.com/emscripten-core/emscripten/blob/9bb322f8a7ee89d6ac67e828b9c7a7022ddf8de2/system/include/emscripten/em_js.h), but I don't really understand what that means. I can find the `__em_js__` prefixed functions on `pyodide._module` but these are not usable directly. I was expecting there to be functions there like `pyodide._module._test_c_tests_blah`.

I want to automate this:
```javascript
 Module.Tests.c_tests_expect_success_success =
    _test_c_tests_expect_success_success;
  Module.Tests.c_tests_expect_fail_fail = _test_c_tests_expect_fail_fail;
  ...
```

dalcde:
In my understanding the primary purpose of `EM_JS` is to write inline JavaScript in C files (similar to inline asm in "normal" C). While you can re-export the functions to JavaScript, it seems to make more sense to just write a normal JavaScript file in the first place.

If you really want to re-export, something like this works:
```c
// test.c
#include <emscripten.h>

EM_JS(int, ret3_js, (), {
  return 3;
});

int ret3() {
  return ret3_js();
}
```
```shell
$ emcc test.c -s 'EXPORTED_FUNCTIONS=["_ret3"]'
```

hoodmane:
My goal is to generate tests with C macros. The reason I want to generate them with C macros instead of in Javascript is that the C macro knows the line of code and the source file which is useful info to include in error messages if the tests fail. It's not possible to get this information from javascript since it doesn't have macros and doesn't have any native notion of a "source file". I can make the tests work this way, but I can't figure out how to automatically expose them.

I will try to think if it's possible to work inside an "init_tests()" C function. Inside an init function I can directly modify `Module` but I can't declare C functions, so I think that also doesn't work.

hoodmane:
I could easily use the macro to generate a wrapper like
```C
int ret3() {
  return ret3_js();
}
```
Then could grep the source for instances of `DEFINE_TEST(test_name,` and set `EXPORTED_FUNCTIONS` based on the located test names. This seems a bit elaborate though...

hoodmane:
@dalcde How do you feel about grepping in the makefile to locate tests to add to `EXPORTED_FUNCTIONS`? Is that acceptable or too weird?

hoodmane:
Another option: I could write a Pythons script that outputs a file `hiwire_init_tests.c` which defines an `EM_JS` function `int hiwire_init_tests();` and then run `hiwire_init_tests()` in `main`. That's also not great, but it seems okay?

Or maybe my whole approach is completely wrongheaded and I should be doing something else.

dalcde:
It appears to `-s LINKABLE=1 -s EXPORT_ALL=1` suffices, which we have already set in the Makefile, so in fact we do not need to manually list the (wrapped) functions.

hoodmane:
Oh that's great! Thanks for the wrapping tip, I don't think that would have occurred to me to do.

What the heck is `["___cxa_guard_acquire", "__ZNSt3__28ios_base4initEPv"]`? Does this do anything when `-s LINKABLE=1 -s EXPORT_ALL=1` is set?

dalcde:
I was about to investigate

Makefile
```diff
@@ -161,7 +162,7 @@ clean-all: clean
 	rm -fr cpython/build
 
 %.bc: %.c $(CPYTHONLIB) $(wildcard src/**/*.h)
-	$(CC) -o $@ -c $< $(CFLAGS) -Isrc/type_conversion/
+	$(CC) -o $@ -c $< $(CFLAGS) $(EXTRA_CFLAGS) -Isrc/type_conversion/
```
rth, line 165: `EXTRA_CFLAGS` is never defined I think?

src/pyodide.js
```diff
@@ -299,18 +299,14 @@ var languagePluginLoader = new Promise((resolve, reject) => {
...
-    'loadedPackages',
-    'pyimport',
-    'repr',
-    'runPython',
-    'runPythonAsync',
-    'checkABI',
-    'version',
-    'autocomplete',
+    'globals', 'loadPackage', 'loadedPackages', 'pyimport', 'repr', 'runPython',
+    'runPythonAsync', 'checkABI', 'version', 'autocomplete'
```
rth, line 303: Are those changes due to clang-format, maybe revert them as they are unrelated?

src/type_conversion/testing.c
```diff
@@ -0,0 +1,124 @@
...
+    if (name.startsWith("_test_")) {
+      Module.Tests[name.slice("_test_".length)] = function()
+      {
+        return c_string_to_js_string(Module[name]());
+      };
+    }
+  }
+
+  // TODO: figure out how to avoid doing this?
+  // Module.Tests.c_tests_expect_success_success =
```
rth, line 90: Are these comments still relevant?

src/tests/test_testing.py
```diff
@@ -1,7 +1,52 @@
...
+
+
+def test_C_tests_succeed1(selenium):
+    assert_C_test(
+        selenium.run_js("return pyodide.Tests.c_tests_expect_success_success();")
+    )
+
+
+def test_C_tests_succeed2(selenium):
+    assert_C_test(selenium.run_js("return pyodide.Tests.c_tests_expect_fail_fail();"))
```
rth, line 28: The somewhat non ideal thing about it is that one has to first write a test in `src/type_conversion/testing.c`, and then some boilerplate code in this file. It would be better if pytest would just look at `pyodide.Tests` methods, and run those that have the `test` substring.Excluding tests that are expected to fail -- those are only to test the framework, right?

For instance, something like,
```py
@pytest.mark.parametrize(
    'name',
    ['c_tests_expect_success_success',
     'c_tests_expect_fail_fail',
     # add future tests here
     ]
)
def test_core_c_func(name, selenium):
     assert_C_test(selenium.run_js(f"return pyodide.Tests.{name}();"))
```
would be a start. There might be some more automated way of listing  all methods in `pyodide.Tests` maybe by using [parametrized fixtures](https://docs.pytest.org/en/stable/fixture.html#fixture-parametrize) but I don't see how to easily do that while also using the selenium fixture to generate the list of tests.

src/type_conversion/testing.h
```diff
@@ -0,0 +1,90 @@
...
+/**
+ *
+ * We define two macros to make tests:
+ * DEFINE_TEST(name, body):
+ *   Defines a test with given name and body. Test should use ASSERT macro
+ *   above.
+ *
+ * DEFINE_TEST_EXPECT_FAIL(name, match, body):
+ *    This variant expects the test to fail with a message that matches js regex
+ *    "match".
```
rth, line 20: Could you add a section to the contributing documentation with these two macros?

src/type_conversion/testing.c
hoodmane: No.

src/pyodide.js
hoodmane: Yeah it's a clang-format thing.

Makefile
hoodmane: This is probably leakage from a different branch.

src/tests/test_testing.py
hoodmane: I agree, I just don't understand pytest well enough to know how to do this. I can set it up as you show above. I was thinking we could add a make job to grep for `DEFINE_TEST_EXPECT_FAIL(` and `DEFINE_TEST(` and stick them into an autogenerated test file.

src/pyodide.js
rth: OK, but that would mean that the lint is not working in CI? None of these lines was changed in this PR..

src/tests/test_testing.py
rth: yes, you can probably also just run regexp in Python the .c file and then auto-generate tests with the [`pytest_generate_tests` hook](https://docs.pytest.org/en/stable/parametrize.html#basic-pytest-generate-tests-example) as done [here](https://github.com/iodide-project/pyodide/blob/115507754122d77c81d987f2f2c1210ab205d98e/src/tests/test_core_python.py#L50)

src/pyodide.js
hoodmane: I don't think it's the CI's fault, I think it was something local on my computer. Not sure.

src/tests/test_testing.py
hoodmane: Thanks!

src/type_conversion/testing.h
hoodmane: I will put it in #1048.

src/tests/test_testing.py
hoodmane: So I guess what I will do is manually wrap the test tests, because some of the tests have to fail in order to cover the failure cases in the testing code. In the autogen we can just skip grepping `testing.c`

hoodmane:
@rth @dalcde I think this is probably unnecessary because there are other better strategies for testing. I guess I might as well close it -- I can come back to it if it turns out a need for this approach.

PR #940, opened 2020-12-24
Release v0.16.0

rth:
Closes https://github.com/iodide-project/pyodide/issues/831

Final changes for the v0.16.0 release. The checklist from https://github.com/iodide-project/pyodide/issues/831 is mostly done. It's unfortunate that https://github.com/iodide-project/pyodide/issues/877 (and in particular https://github.com/iodide-project/pyodide/pull/887) wasn't fully resolved but that would require more investigation.

I know we have many more things in the pipeline, but this release needs to happen at some point, we can't postpone it indefinitely and I won't be available for one week starting from this evening. We can always switch to a shorter release cycle in the future.

The release notes are in https://pyodide.readthedocs.io/en/latest/changelog.html

@dalcde @hoodmane Please let me know if you have any comments.

**Edit:** also added a list of contributors to the release with,
```bash
git shortlog -s 0.15.0.. | cut -f2- | sort --ignore-case | tr '\n' ';' | sed 's/;/, /g;s/, $//' | fold -s
```
(adapted from https://scikit-learn.org/stable/developers/maintainer.html#making-a-release)

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -7,7 +7,7 @@ defaults: &defaults
   environment:
     - EMSDK_NUM_CORES: 4
       EMCC_CORES: 4
-      PYODIDE_BASE_URL: https://cdn.jsdelivr.net/pyodide/dev/full/
+      PYODIDE_BASE_URL: https://cdn.jsdelivr.net/pyodide/v0.16.0/full/
 
 jobs:
   lint:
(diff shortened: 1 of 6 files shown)


hoodmane:
I just submitted a quick patch #941 to allow monkey patching of eval_code. I think we should include it in the release.

@dalcde already put in all of the stuff that was screwing up our build system downstream. I'm happy to get release 0.16 out so we can focus on the next version, most of the stuff I'm working on now will require some effort to get right and doesn't .

hoodmane:
By the way, thanks for maintaining this project @rth! It's a super cool project and it's been a pleasure working with you (well, over like the past week since I started contributing).

rth:
Thanks for your contributions as well @hoodmane !

PR #876, opened 2020-12-16
Improvements to eval_code

hoodmane:
If code ends in semicolon suppress return value (return None).
If code ends in expression return the value of the expression (this is as before).
If code ends in await expression, return the value of awaited expression.
~~If code ends in an assignment, we return the assigned lvalue.~~
~~This works with arbitrary destructuring assignments.~~

One tiny drawback of destructuring assignment is that `l[x] = 5`
gets turned into
```python
   l[x] = 5
   l[x]
```
If `l` has bad definitions for getitem and setitem, it is possible that
the second `l[x]` yields an error. This is in any case a bug in the definition
of `l`, but it makes that bug more confusing.

I also updated it to compile with `ast.PyCF_ALLOW_TOP_LEVEL_AWAIT`.
This only works with python 3.8 and up, but I think we are intending to only
support these versions anyways. For backwards compatibility we could check
if the flag exists first.

The await handling only really will do anything once we have async/await
support, see #245.

PR #941, opened 2020-12-24
Enable monkey patching of eval_code and find_imports

hoodmane:
runpython.c uses a nonidiomatic strategy to index the `pyodide-py` module (which is defined primary in Python). This is a significant usability concern (see conversation with Casatir in #876). This is a quick patch to use the correct CPython API to call a pure python package from C and enable monkey patching.


rth:
Thanks @hoodmane !

Issue #791, opened 2020-11-08
META: Migrate to a conda compatible packages.json


rth:
Currently pyodide has a very simple [packages.json](https://pyodide-cdn2.iodide.io/v0.15.0/full/packages.json), and it would be helpful to add additional useful information (package versions, checksums, file sizes, python and emscripten versions, date of the build, etc).

Conda has a fairly well defined [`repodata.json`](https://docs.conda.io/projects/conda-build/en/latest/concepts/generating-index.html#repodata-json)  (see e.g. another example [here](https://repo.anaconda.com/pkgs/pro/osx-64/repodata.json)) and I would propose to just use that format as much as possible.

This has the advantage of an already existing standard. Also when one day conda package managers (e.g. [mamba](https://github.com/mamba-org/mamba/)) are available in WebAssembly it might be possible to adapt them for dependency resolution of pyodide packages. cc @wolfv @SylvainCorlay

Currently the differences I see with the way conda packages work are,
 - pyodide packages are not necessarily compatible between releases, so for now there is 1 `packages.json` per pyoidide release (which also explains why the very simple dependency resolution that ignores versions is mostly sufficient)
 - there are 2 files (.js, .data) per package instead of 1, so we would probably need to use `sha256_data` and `sha256_js` keys

Also users can install [pure python wheels from PyPi](https://pyodide.readthedocs.io/en/latest/loading_packages.html#micropip) which poses its own challenges for dependency resolution.  

This would a small step in the direction of a better integration with conda-forge (https://github.com/iodide-project/pyodide/issues/38)

wolfv:
Hi @rth!

This is a super-exciting topic for me. I would love to experiment with micromamba compiled to webassembly, and also with conda-webassembly packages. I think we could have a cross-compiling toolchain on conda-forge, and eventually produce webassembly outputs from conda-forge.

Do you guys have an idea what it would take? I guess the first step would be to make emscripten available on conda-forge?

wolfv:
it looks like binaryen (albeit an older version) is already on conda-forge:

- https://github.com/conda-forge/binaryen-feedstock
- https://anaconda.org/conda-forge/binaryen/

wolfv:
@jakirkham ping on this, since you were interested in the conda-side as well.

rth:
Hi @wolfv, good to hear from you! I opened https://github.com/iodide-project/pyodide/issues/795 for a more general discussion of this topic, to keep this issue more specifically about the `packages.json` format..

[...]

PR #945, opened 2020-12-24
Make a copy of `console.html` at `builld/index.html`

hoodmane:
The standard for websites is to make `index.html` be the main file. Since our main file is `console.html`, I propose we should make a copy at `index.html`.

I also added a title and some meta tags.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -60,6 +60,7 @@ all: check \
 	build/pyodide.asm.data \
 	build/pyodide.js \
 	build/console.html \
+	build/index.html \
 	build/renderedhtml.css \
 	build/test.data \
 	build/packages.json \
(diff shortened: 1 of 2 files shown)


rth:
I was thinking that in the future we should put the list of packages as the index.html. But that would probably require doing #791 first (I have a partial solution, need to make a PR).

hoodmane:
I think it's good to put something there, we can always change it later. I can just fetch packages.json and iterate over it in js...

hoodmane:
I think we should start with a link to the console and then to the docs. Then the list of packages would be good. If possible it would be nice to insert links to the package docs / pypi.

dalcde:
Do you want to close this PR and turn it into an issue?

hoodmane:
Sounds good to me.

PR #947, opened 2020-12-25
BLD Make package build output less verbose

dalcde:
1. Only print build logs when a build fails.
2. Raise a custom RuntimeError when build fails instead of verbose CalledProcessError --- we already print build logs

My main concern is that there might be cases where the package build does not error but produces incorrect output files, and one might want to inspect the build process for that, but in that case one can still build locally.

### DIFF
--- a/pyodide_build/buildall.py
+++ b/pyodide_build/buildall.py
@@ -72,10 +72,15 @@ class Package:
                     stderr=subprocess.STDOUT,
                 )
 
-        with open(self.pkgdir / "build.log", "r") as f:
-            shutil.copyfileobj(f, sys.stdout)
+        try:
+            p.check_returncode()
+        except subprocess.CalledProcessError:
+            print(f"Error building {self.name}. Printing build logs.")
 
-        p.check_returncode()
+            with open(self.pkgdir / "build.log", "r") as f:
+                shutil.copyfileobj(f, sys.stdout)
+
+            raise
 
         if not self.library:
             shutil.copyfile(
--- a/packages/distlib/meta.yaml
+++ b/packages/distlib/meta.yaml
@@ -2,7 +2,7 @@ package:
   name: distlib
   version: 0.3.1
 source:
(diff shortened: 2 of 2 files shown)


dalcde:
A reasonable compromise to the problem is to upload the build logs as build artifacts, but I don't think circleci accepts wildcards as for artifact paths. One option would be to tar all the artifacts and upload that, but it would be much less convenient

pyodide_build/buildall.py
```diff
@@ -72,10 +72,15 @@ def build(self, outputdir: Path, args) -> None:
...
+        try:
+            p.check_returncode()
+        except subprocess.CalledProcessError:
+            print(f"Error building {self.name}. Printing build logs.")
 
-        p.check_returncode()
+            with open(self.pkgdir / "build.log", "r") as f:
+                shutil.copyfileobj(f, sys.stdout)
+
+            raise
```
rth, line 83: Do we actually need to print the full traceback here, as opposed to just printing the logs +  error message and doing a `sys.exit(1)`. The nested exceptions with no relevant information can be confusing to users.

dalcde:
>Do we actually need to print the full traceback here, as opposed to just printing the logs +  error message and doing a `sys.exit(1)`. The nested exceptions with no relevant information can be confusing to users.

I was hoping for each function to individually make sense, and forcefully sys.exit'ing in the middle of a function is pretty rude; any error ought to be catchable.

The nested exceptions can probably be suppressed in the parent caller though.

rth:
> I was hoping for each function to individually make sense, and forcefully sys.exit'ing in the middle of a function is pretty rude; any error ought to be catchable.
> The nested exceptions can probably be suppressed in the parent caller though.

Fair enough, we can look into it in some other PR.

dalcde:
Welcome back, by the way!

rth:
Thanks for the work you all have been doing lately :)

PR #950, opened 2020-12-25
Release 0.16.1

rth:
Addresses https://github.com/iodide-project/pyodide/issues/944  following the CI fix in https://github.com/iodide-project/pyodide/pull/946

Makes a new release as 0.16.1, I'll remove the failed 0.16.0 release altogether. 

We could have re-uploaded the 0.16.0 release but those files (and in particular packages.json) are served by S3 with a 1 year cache headers, and there is no automated way to invalidate jwDelivr cache, so it's probably better to just make a new release. 0.16.0post1 would have been the correct version, but it's not a very nice name in the URL so 0.16.1 will do.

Once it's deployed, we would also need to build and upload the pre-build docker image. I don't have access to a machine where I can do it this week. @dalcde you should have permissions to push to https://hub.docker.com/repository/docker/iodide/pyodide (and pyodide-env) now. Otherwise I'll upload it in January.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -7,7 +7,7 @@ defaults: &defaults
   environment:
     - EMSDK_NUM_CORES: 4
       EMCC_CORES: 4
-      PYODIDE_BASE_URL: https://cdn.jsdelivr.net/pyodide/v0.16.0/full/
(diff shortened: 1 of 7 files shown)


Lyoko-Jeremie:
thank you very much for your work~~

i'm online wait to the fixed deploy to release page ~ XD

PR #953, opened 2020-12-26
Restore PYODIDE_BASE_URL

dalcde:
A more permanent solution ought to be found, but CircleCI does not seem to make it easy. This would make dev builds work for now.


rth:
Indeed. Thanks!

PR #141, opened 2018-09-05
Avoid using UTF8ToString

mdboom:
The only places we were using it was for strings that are ASCII anyway,
so this is both more efficient and avoids this Chrome bug:

https://bugs.chromium.org/p/chromium/issues/detail?id=868404

PR #956, opened 2020-12-26
Streamline hiwire_get_dtype

hoodmane:
Implementation of `hiwire_get_dtype` is split across two files. The reason seems to be that `get_dtype` returns two values: format and size. I changed it to take pointers to write the results into. This makes the code shorter and easier to understand (someone reading the files would have to search the source to find out that those constants are only used in two spots).

### DIFF
--- a/src/type_conversion/hiwire.c
+++ b/src/type_conversion/hiwire.c
@@ -332,9 +332,8 @@ EM_JS(JsRef, hiwire_typeof, (JsRef idobj), {
   return Module.hiwire.new_value(typeof Module.hiwire.get_value(idobj));
 });
 
-EM_JS(int, hiwire_constructor_name, (int idobj), {
-  return Module.hiwire.new_value(
-    Module.hiwire.get_value(idobj).constructor.name);
+EM_JS(char*, hiwire_constructor_name, (JsRef idobj), {
+  return stringToNewUTF8(Module.hiwire.get_value(idobj).constructor.name);
 });
 
 #define MAKE_OPERATOR(name, op)                                                \
--- a/src/type_conversion/hiwire.c
+++ b/src/type_conversion/hiwire.c
@@ -428,14 +428,14 @@ EM_JS(void, hiwire_get_dtype, (int idobj, int format_ptr, int size_ptr), {
     for (let[key, [ format, size ]] of entries) {
       let format_utf8 =
         allocate(intArrayFromString(format), "i8", ALLOC_NORMAL);
-      Module.hiwire.dtype_map.set(key, [ format, size ]);
+      Module.hiwire.dtype_map.set(key, [ format_utf8, size ]);
     }
   }
   let jsobj = Module.hiwire.get_value(idobj);
-  let[format, size] =
+  let[format_utf8, size] =
     Module.hiwire.dtype_map.get(jsobj.constructor.name) || [ 0, 0 ];
   // Store results into arguments
-  setValue(format_ptr, format, "i8*");
+  setValue(format_ptr, format_utf8, "i8*");
   setValue(size_ptr, size, "i32");
 });
 
--- a/src/type_conversion/hiwire.c
+++ b/src/type_conversion/hiwire.c
@@ -417,34 +417,37 @@ EM_JS(void, hiwire_copy_to_ptr, (JsRef idobj, void* ptr), {
   Module.HEAPU8.set(new Uint8Array(buffer), ptr);
 });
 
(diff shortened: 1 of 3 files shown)


hoodmane:
@dalcde Can you take a look at the errors here? In test-core-chrome I see a lot of errors failing to import snowballstemmer which is probably a build system / CI problem.
There's also some errors that are problems with my code mixed in.

dalcde:
I would guess it is due to a broken micropip, e.g. see the `stdout` of `test_install_custom_url[chrome]`

hoodmane:
@dalcde What about merging this one? It's a fairly straightforward improvement I think.

hoodmane:
I guess first I'd better finish merging it.

src/type_conversion/hiwire.h
```diff
@@ -459,8 +460,8 @@ hiwire_typeof(JsRef idobj);
  *
  * Returns: New reference to Javascript string
  */
-int
-hiwire_constructor_name(int idobj);
+char*
+hiwire_constructor_name(JsRef idobj);
```
dalcde, line 464: This ought to be HwRef

src/type_conversion/hiwire.c
```diff
@@ -404,53 +408,45 @@ EM_JS(int, hiwire_get_byteLength, (JsRef idobj), {
...
-      dtype = 3; // UINT8CLAMPED_TYPE;
-      break;
-  }
-  return dtype;
-});
+EM_JS(void,
+      hiwire_get_dtype,
+      (JsRef idobj, char** format_ptr, Py_ssize_t* size_ptr),
+      {
+        if (!Module.hiwire.dtype_map) {
```
dalcde, line 423: Not complaining; just trying to document some reasons. Why is this created lazily instead of at `hiwire_init`? Because most users are not very likely to run into this code path?

hoodmane: Two thoughts:
1. don't pay for it if you aren't using it, and
2. keeps the instantiation of the map closer in the code to where it is used.

The second thing is the stronger motivating factor for me though.

src/type_conversion/hiwire.c
```diff
@@ -404,53 +408,45 @@ EM_JS(int, hiwire_get_byteLength, (JsRef idobj), {
...
+            'Float32Array' : [ "f", 4 ],
+            'Float64Array' : [ "d", 8 ],
+            'ArrayBuffer' : [ 'B', 1 ], // Default to Uint8;
+          });
+          Module.hiwire.dtype_map = new Map();
+          for (let[key, [ format, size ]] of entries) {
+            let format_utf8 =
+              allocate(intArrayFromString(format), "i8", ALLOC_NORMAL);
+            Module.hiwire.dtype_map.set(key, [ format_utf8, size ]);
+          }
```
dalcde, line 441: Arguably cleaner way of doing this:
```suggestion
          let alloc = c => allocate(intArrayFromString(c), "i8", ALLOC_NORMAL);
          Module.hiwire.dtype_map = new Map([
            ['Int8Array', [ alloc('b'), 1 ]],
            ['Uint8Array', [ alloc('B'), 1 ]],
            ['Uint8ClampedArray', [ alloc('B'), 1 ]],
            ['Int16Array', [ alloc('h'), 2 ]],
            ['Uint16Array', [ alloc('H'), 2 ]],
            ['Int32Array', [ alloc('i'), 4 ]],
            ['Uint32Array', [ alloc('I'), 4 ]],
            ['Float32Array', [ alloc('f'), 4 ]],
            ['Float64Array', [ alloc('d'), 8 ]],
            ['ArrayBuffer', [ alloc('B'), 1 ]], // Default to Uint8];
          ]);
```

src/type_conversion/hiwire.c
```diff
@@ -404,53 +408,45 @@ EM_JS(int, hiwire_get_byteLength, (JsRef idobj), {
...
+        }
+        let jsobj = Module.hiwire.get_value(idobj);
+        let[format_utf8, size] =
```
dalcde, line 444:
```suggestion
        let [format_utf8, size] =
```

src/type_conversion/hiwire.c
hoodmane: I found the API `stringToNewUTF8` which I think should be preferred over `allocate(intArrayFromString(string), "i8", ALLOC_NORMAL);`. Apparently there was some old chrome bug that meant we couldn't use this #141.

dalcde: I remember seeing that API but then failed to find it anywhere. Glad to know I was not crazy.

Issue #237, opened 2018-10-30
Threading support


rth:
Just an issue to track the status on the threading support.

Emscripten allows Pthreads support via `SharedArrayBuffer` capability in browsers. The later got disabled due to Spectre etc vulnerabilities, however,
 - a few days ago there was an announcement that threshing should work with WASM in Chrome 70 (https://developers.google.com/web/updates/2018/10/wasm-threads)
 - `SharedArrayBuffer` is available in Firefox since 57 via a config option, but that is disabled by default.  The switch to enabled by default can be tracked in https://bugzilla.mozilla.org/show_bug.cgi?id=1477743

There also has been some [activity in emscripten](https://github.com/kripken/emscripten/pulls?utf8=%E2%9C%93&q=is%3Apr+sort%3Aupdated-desc+label%3Amultithreading+) on this subject lately.

Because of,
> If code is compiled with -s USE_PTHREADS=1 and the current browser does not support multithreading, then an exception will be thrown at page load time. It is not possible to build one binary that would be able to leverage multithreading when available and fall back to single threaded when not. For such backwards compatibility, two separate builds must be done, one with -s USE_PTHREADS=1 and the other with -s USE_PTHREADS=0.
(taken from [emscripten docs](https://kripken.github.io/emscripten-site/docs/porting/pthreads.html#special-considerations))

I imagine we won't be able to do much until both Firefox and Chrome have it enabled by default in a stable release. Still it might be worth trying to make it work in a WIP PR to see what are the difficulties that could arise. Opening this issue for future reference.

This would also very partially address #144

pbazant:
Dear pyodide devs, I'd like to start experimenting with the threading support -- our target browser is Chrome anyway (we need web-bluetooth support). Before I delve into the pyodide codebase: How hard should I expect enabling the threading to be?

rth:
@pbazant I don't have a good understand of much work it would be, but it would be certainly worth trying to build with threading support following emscripten docs and reporting here if you run into any issues :)

Though we likely won't be able to merge it in pyodide until support has been enabled in firefox by default, and it doesn't look like it happened yet.

gatesn:
Looks like this fails in that Emscripten won't allow us to use SIDE_MODULES and PTHREADS in the same build: https://github.com/emscripten-core/emscripten/issues/3494

pbazant:
[...]

PR #958, opened 2020-12-26
Add web event loop for supporting asyncio

oeway:
This PR provide an asyncio event loop implementation based on the browser event loop. Fixes https://github.com/iodide-project/pyodide/issues/245

We can simplify the test when https://github.com/iodide-project/pyodide/pull/880 get merged.


oeway:
Not sure if I understand correctly, but I placed the webloop module under `pyodide-py/pyodide`. Maybe that's the wrong place? I am getting the `ModuleNotFoundError: No module named 'js'` error.

Meanwhile, I try to run my test locally via `pytest src/tests/test_webloop.py -k firefox` but it complains it cannot find the module.

oeway:
@pbazant It seems you have already implemented a selector based event loop (https://github.com/iodide-project/pyodide/issues/237#issuecomment-615722268), would you mind to share it? The issue with this one is that we won't be able to block the execution of `loop.run_until_complete`, some library requires that to convert async operations to sync (e.g. [syncer](https://github.com/miyakogi/syncer)).

hoodmane:
I think this implementation needs a fair amount of work. In particular, the browser already has an event loop in it and as much as possible we should rely on the browser event loop. This can be used to simplify the implementation a lot. 

For instance, here is a (prototype of a) better implementation of `call_later`:
```python
def call_later(self, delay, callback):
   if delay < 0:
      raise Exception("Can't schedule in the past")
   self.call_soon(call_later_helper(delay, callback))

async def call_later_helper(delay, callback):
   await sleep(delay)
   self.call_soon(callback)
```

src/pyodide-py/pyodide/webloop.py
```diff
@@ -0,0 +1,425 @@
...
+        """
+        return time.time()
+
+    def run_forever(self):
+        """
+        Run the event loop until stop() is called.
+
+        Note that this function is different from the standard asyncio loop implementation in two ways:
+         1) It won't block the execution
+         2) It returns a Promise object
```
hoodmane, line 84: Please explain here under what conditions the promise will return, under what conditions it will raise, what values will it return, what will it raise.

src/pyodide-py/pyodide/webloop.py
```diff
@@ -0,0 +1,425 @@
...
+        self._scheduled = []
+        self._futures = []
+        self._next_handle = None
+        self._debug = debug
+        self._stop = False
+        self._interval = interval
+        self._timeout_promise = js.eval(
+            "self._timeoutPromise = function(time){return new Promise((resolve)=>{setTimeout(resolve, time);});}"
+        )
+        self._exception_handler = None
```
hoodmane, line 54:
```suggestion
        self._exception_handler = self.default_exception_handler
```

src/pyodide-py/pyodide/webloop.py
```diff
@@ -0,0 +1,425 @@
...
+        Parameters
+        ----------
+        handler
+          If handler is None, the default exception handler will be set.
+        Otherwise, handler must be a callable with the signature matching (loop, context),
+        where loop is a reference to the active event loop, and context is a dict object
+        containing the details of the exception (see call_exception_handler() documentation
+        for details about context).
+        """
+        self._exception_handler = handler
```
hoodmane, line 269:
```suggestion
        self._exception_handler = handler or self.default_exception_handler
```

src/pyodide-py/pyodide/webloop.py
```diff
@@ -0,0 +1,425 @@
...
+            ‘future’ (optional): asyncio.Future instance;
+            ‘handle’ (optional): asyncio.Handle instance;
+            ‘protocol’ (optional): Protocol instance;
+            ‘transport’ (optional): Transport instance;
+            ‘socket’ (optional): socket.socket instance.
+        """
+        if self._exception_handler:
+            self._exception_handler(self, context)
+        else:
+            self.default_exception_handler(context)
```
hoodmane, line 255:
```suggestion
        self._exception_handler(self, context)
```

src/pyodide-py/pyodide/webloop.py
```diff
@@ -0,0 +1,425 @@
...
+        Parameters
+        ----------
+        future
+        A future or coroutine object
+
+
+        Returns
+        -------
+        A Promise object with the returned result or exception
+        """
```
hoodmane, line 116: A simpler implementation could be to wrap the future in a promise, so get the browser event loop to schedule it.
My PR #880 adds conversion promise ==> future. We should add conversion future ==> promise. This will again defer work to the browser event loop.

hoodmane (commented):
I think this implementation needs a fair amount of work. In particular, the browser already has an event loop in it and as much as possible we should rely on the browser event loop. This can be used to simplify the implementation a lot.

For instance, here is a (prototype of a) better implementation of call_later:
```python
def call_later(self, delay, callback):
   if delay < 0:
      raise Exception("Can't schedule in the past")
   self.call_soon(call_later_helper(delay, callback))

async def call_later_helper(delay, callback):
   await sleep(delay)
   self.call_soon(callback)
```
This can be used to get rid of self._scheduled and all the associated logic. 
A simpler implementation  of `run_until_complete` could be to wrap the future in a promise, so get the browser event loop to schedule it.

Once we rely on the browser event loop for everything, we can way reduce the webloop implementation: `_do_tasks` can get removed entirely, `is_running` always returns true, `stop` does nothing, `run_forever` does nothing, etc.

I haven't thought about this approach as much as you have @oeway, so maybe it has drawbacks that haven't occurred to me. We should discuss.

src/pyodide-py/pyodide/webloop.py
oeway: I implemented as you suggested in my last version but then I noticed that it was wrong according to the docs. The issue is a custom exception handler takes two arguments(loop, context) while the default handler takes one argument (context), so we cannot really treat them the same.

hoodmane: Add a comment please?

oeway:
I like the idea of relying on browser event loop, and we need to definitely think and discuss more about how to implement this. My current solution with setTimeout is for sure not ideal. 

However, we need to also stick with the asyncio interface, because these functions will be called by asyncio itself. For example the alternative implementation you propose might not work, because the call_later function should return an asyncio.TimerHandle object (such that one can cancel the task). We can perhaps build our own asyncio.TimerHandle with setTimeout. Let's think a bit more.

The downside for using the browser event loop is that we cannot block the execution, which can limit some applications as I mentioned an comment above. The maximum compatibility can be achieved by implementing the 'select'.

hoodmane:
[https://esdiscuss.org/topic/cancel-promise-pattern-no-cancellable-promises](For cancellation you can use `Promise.race`: make a second promise and resolve that when you want to cancel.)

oeway:
> [For cancellation you can use `Promise.race`:](https://esdiscuss.org/topic/cancel-promise-pattern-no-cancellable-promises) make a second promise and resolve that when you want to cancel.

I understand that you can cancel a setTimeout by using `promise.race`, however, for `call_soon`, we will need a way to cancel promise which is not supported ([here](https://github.com/tc39/proposal-cancelable-promises) is a withdrew proposal). There are [different ways](https://medium.com/@benlesh/promise-cancellation-is-dead-long-live-promise-cancellation-c6601f1f5082) to implement our own cancelation mechanism, however, they are either hacky or require additional library, not sure if that will be better than our own schedule logic. Further idea for promise cancellation are welcome!

src/pyodide-py/pyodide/webloop.py
oeway: Interesting idea, but since run_until_complete is linked with other functions such as `create_task` and `call_soon` (e.g. in the coroutine one can call create task and we need wait also for the newly created tasks), it is likely we need to fix the promise cancellation issue.

src/pyodide-py/pyodide/webloop.py
oeway: Added a comment to note the difference.

src/pyodide-py/pyodide/webloop.py
oeway: Let's continue the discussion below.

hoodmane:
How about this: we have a future, the promise just is making the browser repeatedly check whether future.send(None) raises `StopIteration`. The future can be cancelled. Then the future can just raise `CancelledError` into the wrapping promise.

oeway:
> In the case that we awaited a native future that came from a python library, the promise just is making the browser repeatedly check whether future.send(None) raises StopIteration. The future can be cancelled. Then the future can just raise CancelledError into the wrapping promise.

I feel like I am already doing it in a similar way, I used a setTimeoutPromise to "make the browser repeatedly check and run tasks". Except we don't need to call `future.send(None)` etc. it's all handled by asyncio itself.

> My takeway from this is that cancellation works or doesn't based on whether the behavior originated in a browser promise or a python future. The stuff originating on the python side (which is only compositions of stuff coming from the browser) is easily cancellable regardless of whether we use our own scheduler or a Promise. On the other hand, the futures that wrap promises are impossible to cancel no matter what.

This also means we cannot simply wrap a future into a promise and let the browser event loop handle it. There should be a queue maintained in Python. What's why I have the heapq (in fact inherit from the original EventSimulator) to maintain the tasks.

Another issue is, if we do a dead loop only for python tasks (see [here] for the original [EventSimulator](https://gist.github.com/damonjw/35aac361ca5d313ee9bf79e00261f4ea#file-eventsim-py-L30-L43) implementation), the js promise don't get cpu time to execute. That's why I have to do a recursive loop with setTimeout to switch between python and js.

With all these, do you see an easier way to implement it?

hoodmane:
> With all these, do you see an easier way to implement it?

I still think I see an easier way to implement it but I am still catching up with your thought processes for doing it this way.

> This also means we cannot simply wrap a future into a promise and let the browser event loop handle it. There should be a queue maintained in Python. That's why I have the heapq (in fact inherit from the original EventSimulator) to maintain the tasks.

I don't understand this paragraph, can you tell me more about what the purpose of this queue is?

> if we do a dead loop only for python tasks, the js promise don't get cpu time to be executed.

Yes, this one I had in mind.

hoodmane:
I think we should put a lot of explanatory comments into this code since it is clearly very subtle.

oeway:
> I don't understand this paragraph, can you tell me more about what the purpose of this queue is?

Consider we are doing `run_forever` and there are many tasks scheduled, we need a queue to 1) store these tasks 2) sort them by their priority (i.e. time)

> I still think I see an easier way to implement it but I am still catching up with your thought processes for doing it this way.

OK, then I would say, go ahead and work on this PR or do a complete rewrite.

hoodmane:
> Consider we are doing run_forever and there are many tasks scheduled, we need a queue to 1) store these tasks 2) sort them by their priority (i.e. time)

Only if we can't use the browser event loop for both of those things. So far, you've raised two concerns about using the browser event loop:
1. you can't cancel promises
2. you can't block

I argued before that problem 1. you can't cancel promises is not improved in any way by doing our own hand management. If we use the browser event loop, we can't cancel promises but we can cancel python futures. If we use our own queue, we can't cancel promises but we can cancel python futures.

The fix for problem 2 is to use [wasm threads](https://developers.google.com/web/updates/2018/10/wasm-threads). I believe that wasm threads has a POSIX compatible pthreads and so we can use select.

I propose we make a simple loop using my approach and the more complicated loop open as a PR blocking on making pyodide work with wasm threads.

hoodmane:
Wasm threads will presumably also allow keyboard interrupts to work the normal way rather than using my weird hack?

oeway:
I got it roughly, not totally clear to me how you would schedule the immediate and delayed tasks to the browser event loop. It would help if you have a prototype that I can play with.

oeway:
BTW, have you tried to compile it with the pthread flag? If we can have select enabled then we might not need this loop. For example, I would like to use the [fsspec](https://filesystem-spec.readthedocs.io/en/latest/) library, but it runs a loop in a separate thread.

hoodmane:
No I have not compiled with pthread flag. You can try, or we could beg @dalcde to look into it.

oeway:
Would like to try but I have some issue with building the core locally via docker, and I don't know exactly where to add the flag. Indeed, it would great if you can try that or provide some guidance @dalcde

hoodmane:
Right now I'm busy fixing the typing of hiwire. I thought `EM_JS` forced you to only use the `int` type because they only use `int` types. If it's a hiwire key or a `PyObject*` or a `char*` or a `bool`, they all only use int.

rth:
> BTW, have you tried to compile it with the pthread flag? If we can have select enabled then we might not need this loop.

We won't be able to enable threads (even if it works) for some time yet as browser support is very new #237

oeway:
> We won't be able to enable threads (even if it works) for some time yet as browser support is very new #237

Maybe it's already the time to investigate? It's going to be very useful, imagine with pthread enabled, we can potentially have a web loop closer to native asyncio, allowing more powerful data analysis with multi-threading, the subsequent packages won't need to patch the threading related parts etc. Besides that, the wasm thread features is supported already in Firefox 79 (released 5 month ago) and Chrome 70 (2+ years ago), given the current status of pyodide, most users are early adopters that can afford to upgrade to the latest browser version.

rth:
@oeway replied in https://github.com/iodide-project/pyodide/issues/237#issuecomment-752949285

rth:
Note that we want to be able to import the `pyodide-py` package outside of pyodide to be able to test parts of it in python.
```
pytest src/
```
with python 3.8 should not raise exceptions. That's the reason the CI is failing now. So it's better to `import js` inside the 4 methods that need it. Then you should be able to monkeypach js, and test at least some part of it with python outside of pyodide.

Also, following methods are never called in tests,
 - `get_debug`, `set_debug`
 - `is_running`, `is_closed`
 - `stop`
 - `get_exception_handler`
 - `call_later`
 - `create_task`
 - `set_task_factory`, `get_task_factory`

I suppose they might be used by asyncio internally, but it would still be good to test them explicitly.

Following method is never used, and might not be necessary since it is private?
 - `_timer_handle_cancelled`

src/pyodide-py/pyodide/webloop.py
```diff
@@ -0,0 +1,429 @@
...
+
+class WebLoop(asyncio.AbstractEventLoop):
+    """A custom event loop for running asyncio in Pyodide
+
+    It works by utilizing the browser event loop via the setTimeout function
+    """
+
+    def __init__(self, debug: Optional[bool] = False, interval: Optional[int] = 10):
+        """
+        Instatiate the web loop
```
rth, line 41: The `__init__` docstring should be removed and merged with class docstring. We also need to document different parameters in the docstring.
See https://numpydoc.readthedocs.io/en/latest/format.html#docstring-standard and other examples in the code base.

src/pyodide-py/pyodide/webloop.py
```diff
@@ -0,0 +1,429 @@
...
+        Return the current time, as a float value, according to the time module (time.time()).
+        """
+        return time.time()
+
+    def run_forever(self):
+        """
+        Run the event loop until stop() is called.
+
+        Note that this function is different from the standard asyncio loop implementation in two ways:
+         1) It won't block the execution
```
rth, line 83: Could you elaborate on why that's not the issue when using it and the motivation behind it.

src/pyodide-py/pyodide/webloop.py
```diff
@@ -0,0 +1,429 @@
...
+            self._do_tasks(forever=True)
+
+        return js.Promise.new(run)
+
+    def run_until_complete(self, future: Awaitable):
+        """
+        Run until the future (an instance of Future) has completed.
+
+        Note that this function is different from the standard asyncio loop implementation in two ways:
+         1) It won't block the execution
```
rth, line 104: Same comment as above

src/pyodide-py/pyodide/webloop.py
```diff
@@ -0,0 +1,429 @@
...
+
+    It works by utilizing the browser event loop via the setTimeout function
+    """
+
+    def __init__(self, debug: Optional[bool] = False, interval: Optional[int] = 10):
+        """
+        Instatiate the web loop
+        """
+        self._running = False
+        self._immediate = []
```
rth, line 44: Maybe use a `collections.deque` since the only operations on it are append and pop. That would also avoid taking slices to pop the first element..

src/pyodide-py/pyodide/webloop.py
```diff
@@ -0,0 +1,429 @@
...
+        """
+        self._exception = None
+        self._result = None
+        self._running = True
+        if self._stop:
+            self._quit_running()
+            return
+        while len(self._immediate) > 0:
+            h = self._immediate[0]
+            self._immediate = self._immediate[1:]
```
rth, line 146:
```suggestion
            h = self._immediate.popleft()
```
once it's a deque

src/pyodide-py/pyodide/webloop.py
```diff
@@ -0,0 +1,429 @@
...
+
+    def call_later(
+        self,
+        delay: float,
+        callback: Callable,
+        *args,
+        context: contextvars.Context = None
+    ):
+        """
+        Schedule callback to be called after the given delay number of seconds (can be either an int or a float).
```
rth, line 307: Need documentation of parameters.

src/pyodide-py/pyodide/webloop.py
```diff
@@ -0,0 +1,429 @@
...
+
+class WebLoopPolicy(asyncio.DefaultEventLoopPolicy):
+    """
+    A simple event loop policy for managing WebLoop based event loops.
+    """
+
+    def __init__(self):
+        """
+        Instantiate the web loop policy
+        """
```
rth, line 395: Docstring should be removed and merged with the class docstring.

rth:
Generally we would need to document all method parameters in docstrings.

oeway:
@rth Thanks for reviewing it, I will get back when I get more time.

src/pyodide-py/pyodide/webloop.py
hoodmane: This is a fun one. It's actually impossible for it to block: the majority of async calls that need to be done require the help of the browser, in particular they run on the browser event loop. If our duplicate event loop blocks, the browser event loop never gets a chance to run and the IO we're waiting for never happens. (Also in native python we would go over to a different thread to schedule tasks because the one thread is blocked. This is hard without threading.)

hoodmane:
> So it's better to import js inside the 4 methods that need it.

Actually, this is even important just for running in the browser: pyodide_py is bootstrapped together out of `core` and the python sources. Currently, the boot strapping order JUST BARELY allows this to work:
```C
  TRY_INIT(js2python);
  TRY_INIT(JsImport); // first becomes possible to import from js after this
  TRY_INIT(JsProxy); // Currently we first import pyodide here
  TRY_INIT(pyproxy); // it first becomes possible to call a jsfunc after this
  TRY_INIT(python2js); // This is actually a noop
```
So inside of pyodide you can say `from js import Promise` at top level but if you say `Promise.new(some_func)` at top level then it will trigger an error something like `Uncaught TypeError: Module.PyProxies is undefined`. I guess I should open a separate issue about handling this bootstrapping in a more understandable manner, or at least documenting it.

But I think the ideal thing is to avoid using `js` in this file entirely. `self.timeout_promise` should be replaced with `from pyodide.util import sleep`, we can define `sleep` as a testing shim when testing natively (how though?) and replace it in `pyodide.js` or `core`. It's a bit weird though to test this because this event loop is only supposed to run when there is another event loop running.
The other thing used here is `Promise`. In the `await-jsproxy` PR we add the ability to convert `promise` ==> `future`. We could convert the other way, or add `pyodide.util.Promise` which can be bootstrapped as appropriate.

hoodmane:
The bootstrapping issues I described in my previous comment were fixed by #1090 so they no longer apply.

src/pyodide-py/pyodide/webloop.py
itszn: So does this mean that this is essentially a busy loop? (ie constantly running a tight loop until a something changes)

hoodmane: @itszn I think it sort of means the opposite of that. With a busy loop, we would starve out the browser event loop. Normal native Python event loops use `select` to wait on other threads. We basically just tell the browser event loop to do the work and then wait on it.

By the way, in my opinion this PR is superseded by #1158 where I took the ideas here and simplified the implementation as much as possible.

itszn: (This is just for my own understanding)
As far as I was aware, in browsers a WASM export can't block (This is why sleep is a busy loop in emscriptn
https://emscripten.org/docs/porting/pthreads.html#blocking-on-the-main-browser-thread). How does the python event loop wait for the js resolve without returning out of wasm? I know one option is asyncify but that isn't supported yet right?

hoodmane: This is all pretty confusing, it took me a good deal of careful contemplation to make sense of it.

> in browsers a WASM export can't block

That's correct. 

> How does the python event loop wait for the js resolve without returning out of wasm?

We do return out of wasm, this is necessary because if we don't return from wasm we block the browser event loop and nothing gets resolved. We use setTimeout. The main loop is:
```python
def _do_tasks(self, ...):
    # do various stuff
    setTimeout(self._interval, self._do_tasks)
```

Similarly, in PR #1158 I define `call_later` to directly do setTimout:
https://github.com/iodide-project/pyodide/blob/a1c30a9329d8fd6265ecbf2b915b0aa175b79e81/src/pyodide-py/pyodide/simple_web_loop.py#L95

> I know one option is asyncify but that isn't supported yet right?

asyncify is to allow normal synchronous C functions to do async calls. I think the primary point of this is that in native C you can do IO by blocking the current thread until the OS gets back to you with the result. This can't work in browser because there is only one thread and if that thread blocks no IO ever happens.
I believe the way this works is that emscripten saves the current stack and yields to the browser event loop, then when the asynchronous task is finished emscripten "rewinds" the stack and then continues as if nothing happened.

In our case we don't need anything like asyncify because the python APIs we're using are already asynchronous: an async coroutine is a type of generator, when you await a future, it yields the future to the event loop. Then when the future is resolved, the event loop calls `coroutine.send(result_of_future)`, and the execution continues from the `await` keyword with `result_of_future` as the result of the `await`.

hoodmane: I'd suggest you try experimenting with `send` directly. Consider the following, which you can do in 
https://pyodide-cdn2.iodide.io/dev/full/console.html

```python
import asyncio
class DumbLoop(asyncio.AbstractEventLoop):    
   def create_future(self):
      return asyncio.Future(loop=self)
   def get_debug(self):
      return False
asyncio.set_event_loop(DumbLoop())
from js import fetch
async def test():
   resp = await fetch("packages.json")
   json = await resp.json()
   return json
c = test()
```
Now `c` is a coroutine. If you run `c.send(None)` that starts the coroutine running and runs it up to the first `await`.
```python
fut1 = c.send(None)
```
Now `fut1` is a `Future` which wraps the `Promise` returned by `fetch`. It will resolve when the `fetch` finishes. Wait a few seconds, then:
```python
fut1.result() # ==> [object Response]
```
The response object looks like:
```javascript
Response { type: "basic", url: "https://pyodide-cdn2.iodide.io/dev/full/packages.json", redirected: false, status: 200, ok: true, statusText: "OK", headers: Headers, body: ReadableStream, bodyUsed: false }
```
anyways now that we have the result, the coroutine is expecting us to send it in:
```python
fut2 = c.send(fut1.result())
```
the coroutine stores the response object `fut1.result()` into `resp` and then runs until `await resp.json()`. The return value `fut2` is a future that resolves when `resp.json()` is done.
Now `fut2.result()` will be the actual json object, deserialized into javascript. Finally, we're supposed to send this back into the coroutine. The coroutine stores it into `json` and then `return json` has the effect of `raise StopIteration(json)`:
```python
try:
   c.send(fut2.result())
except StopIteration as retval:
   result = retval
```
Now `result` is the json object.

oeway:
Close this now since there is a simpler implementation in https://github.com/iodide-project/pyodide/pull/1158 by @hoodmane

PR #96, opened 2018-07-26
Fix #93 by avoiding use of TextDecoder

mdboom:
Any use of `TextDecoder` on a part of the Wasm HEAP seems to cause crashes
when using `WebAssembly.instantiate` later on (in Chrome).

The fix is to:

- Avoid use of `TextDecoder` in emscripten-generated code by setting
  `-s TEXTDECODER=0`
- Do the Python-to-JS string conversion a different way:

  - Handle the native format of a Python Unicode string (which is either UCS1,
    UCS 2 or UCS 4) directly.  This has the added advantage of being
    computationally simpler than encoding/decoding to/from UTF8.

PR #962, opened 2020-12-27
BLD Remove TEXTDECODE=0 build flag

dalcde:
This was introduced in #96 to bypass a Chrome bug introduced in Chrome 69 and fixed in Chrome 70. Chrome 69 supposedly has a market share of 0.09%, so I think it is safe to drop support.


rth (approved):
Thanks!

PR #965, opened 2020-12-27
DOC Build dependency update

mgreminger:
Cython is now needed to compile SciPy so I added that to the dependency list for Linux. Also added the Cython dependancy for Mac and looked up the brew command for Cython. I also added a note that Pyodide can be built on windows using WSL2 since this is how I'm currently building on Windows. Actually works pretty well. WSL2 made a lot of improvements on the file systems side compared to the original WSL, which makes building large projects possible.


dalcde:
I believe SciPy requires Tempita to build as well.  Can you add that to the documentation?

mgreminger:
I was able to compile without Tempita. Looks like Cython includes a version of tempita that [SciPy pulls in](https://github.com/scipy/scipy/blob/5f4c4d802e5a56708d86909af6e5685cd95e6e66/tools/cythonize.py#L112) if it's not installed.

dalcde:
Good catch. We should stop installing it in our docker image then.

dalcde:
Thanks!

PR #967, opened 2020-12-28
DOCK Don't install tempita

dalcde:
Cython ships with its own copy of Tempita, so there is no need to install it manually.

The first two commits in the PR uninstalls it during the build process, and the last undoes that change and modifies the Dockerfile instead.


rth (approved):
Thanks!

PR #968, opened 2020-12-28
Remove SWAPPABLE_ASM_MODULE setting

dalcde:
This is removed in emscripten 1.39.16, and for the wasm backend, was broken since 1.39.6. Eliminating the setting will make it easier to upgrade in the future.

This setting was added in b2599e31 without much explanation; removing it seems to work.


rth (approved):
I'm also not sure about the initial motivation for adding it. Great that everything is working without this flag.

PR #970, opened 2020-12-28
Remove EMULATED_FUNCTION_POINTERS setting

dalcde:
According to @pmp-p it is no longer needed, and this setting is
unavailable when using the upstream backend.

https://github.com/iodide-project/pyodide/issues/476#issuecomment-533285690

The theme of this series of PRs (this, #968, #966, #962) is to remove some unneeded build flags. There are various reasons:
 - There is no point in specifying unneeded options; as far as I can tell, these options do not result in "better' code.
 - Some of these options are no longer supported in later versions of emscripten. Removing them earlier on makes upgrading simpler.
 - Using default settings brings us closer to well-tested code paths in emscripten, lowering the likelihood of encountering emscripten bugs when upgrading (or just building things in general).

The branch https://github.com/dalcde/pyodide/tree/all merges all the changes together to ensure simultaneously removing all these settings still works.


pmp-p:
Well all I can say it's C-Python 3.8 / 3.9 releases are fixed for that, but it could be entirely possible that some external modules outside stdlib require patching too ( maybe case of ctypes callbacks or alike ) i did not investigate.

dalcde:
In any case, EMULATED_FUNCTION_POINTERS will not be available in the future, so if that is the case, we ought to start patching those external modules.

dalcde:
If we could remove the EMULATE_FUNCTION_POINTER_CASTS flag, that would eliminate a lot of our troubles. I'm trying that on a branch

dalcde:
We get lots of errors if we remove EMULATE_FUNCTION_POINTER_CASTS.

rth:
Thanks! 

> We get lots of errors if we remove EMULATE_FUNCTION_POINTER_CASTS.

We can probably open a separate issue about it.

PR #971, opened 2020-12-28
Use the C type system in hiwire

hoodmane:
Before this PR, hiwire effectively opts out of the C type system entirely by declaring everything as int. A `char*`? Cast it to `int`. What about a `PyObject*`? Better cast it to `int` too. This makes the calling code more verbose (at least if the calling code uses the type system) and is not good for maintainability.

I made a NewType called `HwObject` for hiwire keys. If you google how to make a NewType in C, people recommend a single field struct. However, `EM_JS` does not convert that compatibly: it converts any struct to the address of the struct, but an integer to the value of the integer. This created trouble. I decided instead to make a dummy empty struct and work with pointers to that struct. To ensure that converting to and from the `NewType` always requires an explicit cast, I added `-Werror=int-conversion -Werror=incompatible-pointer-types` to the compiler options. (I like this setting better in any case.) 

I added asserts to `main.c` checking that `HwObject` has the same alignment and size as `int`. Hopefully this will prevent confusing bugs if that ever changes.

Unfortunately this is going to have major merge conflicts with all my other open PRs but I think this is a very important improvement.

### DIFF
--- a/src/main.c
+++ b/src/main.c
@@ -33,11 +33,11 @@
 int
 main(int argc, char** argv)
 {
-  if (alignof(HwRef) != alignof(int)) {
-    FATAL_ERROR("HwRef doesn't have the same alignment as int.");
+  if (alignof(JsRef) != alignof(int)) {
+    FATAL_ERROR("JsRef doesn't have the same alignment as int.");
   }
-  if (sizeof(HwRef) != sizeof(int)) {
-    FATAL_ERROR("HwRef doesn't have the same size as int.");
+  if (sizeof(JsRef) != sizeof(int)) {
+    FATAL_ERROR("JsRef doesn't have the same size as int.");
   }
   TRY_INIT(hiwire);
 
--- a/src/main.c
+++ b/src/main.c
@@ -10,6 +10,8 @@
 #include "python2js.h"
 #include "runpython.h"
 
+#include <stdalign.h>
+
 #define FATAL_ERROR(args...)                                                   \
   do {                                                                         \
     printf("FATAL ERROR: ");                                                   \
@@ -31,6 +33,11 @@
 int
 main(int argc, char** argv)
 {
+  printf("alignof(HwObject): %zu, alignof(int): %zu\n",
+         alignof(HwObject),
+         alignof(int));
+  printf(
+    "sizeof(HwObject): %zu, sizeof(int): %zu\n", sizeof(HwObject), sizeof(int));
   hiwire_setup();
 
   setenv("PYTHONHOME", "/", 0);
--- a/src/main.c
+++ b/src/main.c
@@ -33,11 +33,11 @@
 int
 main(int argc, char** argv)
 {
-  if (alignof(HwObject) != alignof(int)) {
-    FATAL_ERROR("HwObject doesn't have the same alignment as int.");
+  if (alignof(HwRef) != alignof(int)) {
+    FATAL_ERROR("HwRef doesn't have the same alignment as int.");
   }
-  if (sizeof(HwObject) != sizeof(int)) {
-    FATAL_ERROR("HwObject doesn't have the same size as int.");
+  if (sizeof(HwRef) != sizeof(int)) {
+    FATAL_ERROR("HwRef doesn't have the same size as int.");
   }
   TRY_INIT(hiwire);
 
--- a/src/main.c
+++ b/src/main.c
@@ -1,5 +1,7 @@
 #include <Python.h>
+#include <assert.h>
 #include <emscripten.h>
+#include <stdalign.h>
 
 #include "hiwire.h"
 #include "js2python.h"
@@ -9,8 +11,6 @@
 #include "pyproxy.h"
 #include "python2js.h"
 #include "runpython.h"
-
-#include <stdalign.h>
 
 #define FATAL_ERROR(args...)                                                   \
   do {                                                                         \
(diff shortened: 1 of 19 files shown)


hoodmane:
@dalcde Again I think these timeout errors and stuff related to micropip aren't due to the changes I made?

src/main.c
```diff
@@ -31,8 +33,12 @@
 int
 main(int argc, char** argv)
 {
-  hiwire_setup();
```
dalcde, line 34: I blame this for your test failures

src/type_conversion/types.h
```diff
@@ -0,0 +1,49 @@
+#ifndef MY_TYPES_H
+#define MY_TYPES_H
```
dalcde, line 2: Any reason not to use `stdint.h` instead?

src/type_conversion/python2js.c
```diff
@@ -313,17 +311,20 @@ _python2js_remove_from_cache(PyObject* map, PyObject* pyparent)
   int result = PyDict_DelItem(map, pyparentid);
   Py_DECREF(pyparentid);
 
-  return result ? HW_ERROR : 0;
+  return result;
 }
 
-int
+struct test
```
dalcde, line 317: You probably meant to delete this?

dalcde:
Some final bikeshedding; feel free to ignore
1. Would it make more sense to rename `HwObject` to `HwRef`, or something like that? A HwObject is a pointer/reference to an object, instead of the object itself.
2. The newtype hack makes me feel uncomfortable. I think in general it is a bad idea to treat pointers as if they are just integers, because pointers have provenance. I'd imagine it doesn't matter if you never dereference it (or construct references in the first place), but it is disconcerting nonetheless.

src/main.c
hoodmane: Yes that looks like a problem. I wonder why it worked locally.

src/type_conversion/types.h
hoodmane: Because `uint32_t` is kind of a mouthful, especially the underscore. Since Rust and Linux both use `u32` I feel like there is a precedent. Maybe I should include stdint and then `typedef uint32_t u32;`

hoodmane:
> Would it make more sense to rename `HwObject` to `HwRef`?

Yes I like that. 

> I think in general it is a bad idea to treat pointers as if they are just integers

They are just integers though. In this case, our program effectively has two allocators, the normal heap and hiwire. The C language is pretty limited and I cannot declare a second variety of pointer / reference. I would do the single field struct approach, but `EM_JS` will usually but not always convert that into a memory address rather than an int.

src/type_conversion/types.h
dalcde: `typedef uint32_t u32` is definitely an improvement imo. I think strictly speaking, the C standard doesn't specify what e.g. `int` must be, so `uint32_t` would be more explicit and more obviously correct.

wasm uses these type names too so I would definitely consider them standard

hoodmane: Great, I'll do that.

dalcde: On the other hand, there is also advantage in using more standard definitions in terms consistency, especially since these types are not used a lot. I don't have strong opinions either way.

hoodmane: > strictly speaking, the C standard doesn't specify what e.g. `int` must be

Absolutely, this is an annoying deficiency of the C language (it's even worse in OpenGL though).

dalcde:
This needs some formatting fix

hoodmane:
Thanks.

src/type_conversion/hiwire.h
```diff
@@ -397,117 +410,117 @@ hiwire_call_member(int idobj, int ptrname, int idargs);
...
+hiwire_new(HwRef idobj, HwRef idargs);
 
 /**
  * Returns the value of the `length` member on a Javascript object.
  *
  * Returns: C int
  */
-int
-hiwire_get_length(int idobj);
+bool
```
dalcde, line 421:
```suggestion
int
```

src/type_conversion/hiwire.h
```diff
@@ -14,48 +16,63 @@
...
+#define HW_ERROR ((HwRef)(-1))
+#define HW_UNDEFINED ((HwRef)(-2))
+#define HW_TRUE ((HwRef)(-3))
+#define HW_FALSE ((HwRef)(-4))
+#define HW_NULL ((HwRef)(-5))
 
 /**
  * Initialize the variables and functions required for hiwire.
  */
 void
```
dalcde, line 45:
```suggestion
int
```

dalcde:
My plan is to merge this (my) tomorrow morning if it is green, so if this goes through you can start updating your other PRs.

hoodmane:
Thanks =D. They turned you into a maintainer really quick. Very convenient.
I think I'll change `HwRef` ==> `JsRef`. I think that is more readable.

hoodmane:
Maybe the constants `HW_NULL` should be `JS_NULL` too.

hoodmane:
How would you feel about renaming everything in hiwire like this? e.g., `hiwire_throw_error` ==> `Js_throw_error`

hoodmane:
And then rename the file `jsref.c` and `jsref.h` and `Module.jsref.incref()`, `Module.jsref.decref()`?
I like the name hiwire a lot aesthetically, but it isn't very descriptive.

dalcde:
I wouldn't be comfortable merging the huge renaming without the opinions of the original maintainers. Are you okay with postponing it to a separate PR?

It would be a simple sed and we can do it when things are quieter to avoid conflicts

hoodmane:
> postponing it to a separate PR
Okay.

hoodmane:
I moved the rename to #980.

Issue #788, opened 2020-11-03
Nested attribute access in JS->Python type conversion


rth:
Currently the following code fails,
```js
>>> from js import window
>>> window.URL.createObjectURL
Error: Traceback (most recent call last):
  File "/lib/python3.7/site-packages/pyodide.py", line 45, in eval_code
    return eval(compile(expr, '<eval>', mode='eval'), ns, ns)
  File "<eval>", line 1, in <module>
AttributeError: 'JsBoundMethod' object has no attribute 'createObjectURL'
```
(while `window.URL.createObjectURL` is a valid JS object) because nested attributes (i.e. attribute of an attribute) don't seem to be supported. It would have been nice to make it work, though I have not looked at how difficult that would be.

hoodmane:
Ah this is another issue with the typeconversions code! I'll try to fix it.

hoodmane:
The issue here is a bad interaction between `pyproxy.c` and `jsproxy.c`. In particular, if you index a `jsproxy` and the value of the field  is a `pyproxy` then instead of returning the wrapped python object `jsproxy` returns a `JsBoundMethod`. It does this because the `pyproxy` is a function. Why is it a function? Well it might implement `__call__` and javascript throws an error if you try to call something that is not a function. So the `PyProxy` is a function.
https://github.com/iodide-project/pyodide/blob/41592637e6a5231d1fa74235ec5f46919c18e25d/src/type_conversion/pyproxy.c#L144

Then `JsProxy` should check first whether the function is a `PyProxy` or not before just blindly returning a `JsBoundMethod`.
https://github.com/iodide-project/pyodide/blob/master/src/type_conversion/jsproxy.c#L70

I can submit a PR to fix this, though this is code that I plan to rewrite much more drastically soon.

Issue #922, opened 2020-12-22
Funny bug


hoodmane:
In a fresh copy of `console.html`:
```python
>>> pyodide.version()
0.15.0
>>> pyodide.version()
Error: Traceback (most recent call last):
  File "/lib/python3.8/site-packages/pyodide/_base.py", line 65, in eval_code
    return eval(compile(expr, "<eval>", mode="eval"), ns, ns)
  File "<eval>", line 1, in <module>
AttributeError: module 'pyodide' has no attribute 'version'
```
Hilarious. 

Not sure the best way to deal with `jspyodide` vs `pypyodide` for naming purposes.

rth:
Yes, that's certainly confusing,
```
>>> pyodide
[object Object]
>>> pyodide.version()
0.15.0
>>> pyodide
<module 'pyodide' from '/lib/python3.8/site-packages/pyodide/__init__.py'>
```

> Not sure the best way to deal with jspyodide vs pypyodide for naming purposes.

If we change one or both names, that would also break backward compatibility, wouldn't it? For the package names so far we went with `pyodide-js` and `pyodide-py` but I'm not sure about imports. 

This is only an issue with REPL isn't it, maybe we can just makes sure the JS module is not accessible in the Python REPL?

hoodmane:
Yeah somewhere in the C code it sets `globals["pyodide"] = Module`. This is probably a bad decision.

dalcde:
Do you want to open an issue for the underlying problem?

hoodmane:
Sounds like a good idea.

hoodmane:
Oh I already did that: #1000.

Issue #918, opened 2020-12-22
Three.js example does not work in Python


mdamien:
I translated this example to Python: https://threejs.org/docs/#manual/en/introduction/Creating-a-scene

```python
from js import THREE, window

scene = THREE.Scene()
camera = THREE.PerspectiveCamera(75, window.innerWidth / window.innerHeight, 0.1, 1000)

renderer = THREE.WebGLRenderer()
renderer.setSize(window.innerWidth, window.innerHeight)
document.body.appendChild(renderer.domElement)
```

But I get `can't redefine non-configurable property "id"`

hoodmane:
My guess is that the error occurs in the last line: `document.body.appendChild(renderer.domElement)`.
Could you try commenting it out and see if the script works up to that point? If the bug is in the last line, maybe we can find an example of the issue that doesn't require the `THREE` dependence.

Also, how are you getting `THREE`? Is there a cdn?

mdamien:
The minimal example to reproduce the error:

```python
from js import THREE, window

scene = THREE.Scene()
camera = THREE.PerspectiveCamera(75, window.innerWidth / window.innerHeight, 0.1, 1000)
```

I'm getting Three via a CDN yes

rth:
Note that in JS  example we have,
```js
const scene = new THREE.Scene();
```
so with pyodide to call the new operator it would be,
```py
scene = THREE.Scene.new();
```
cf https://pyodide.readthedocs.io/en/latest/type_conversions.html#javascript-from-python

However it's likely this will also not work due to the current limitation of the JS > Python interface https://github.com/iodide-project/pyodide/issues/788

hoodmane:
I think this is closed by #1124 and #1126 and maybe some stuff. Please open a new issue if this still doesn't work.

akaszynski:
If anyone is still interested, here's a few more pointers:

## `three_example.py`

```py
def display_scene():
    from pyodide import create_proxy

    from js import THREE, window

    scene = THREE.Scene.new();
    camera = THREE.PerspectiveCamera.new(75, window.innerWidth / window.innerHeight, 0.1, 1000)

    renderer = THREE.WebGLRenderer.new()
    renderer.setSize(window.innerWidth, window.innerHeight)

    geometry = THREE.BoxGeometry.new()
    material = THREE.MeshBasicMaterial.new(color="#00ff00")
    cube = THREE.Mesh.new( geometry, material )
    scene.add( cube )

    camera.position.z = 5
[...]

PR #972, opened 2020-12-29
Flexible jsimport

### LINKED ISSUE #768: TypeError: Illegal invocation
I'm just curious why I can't call some js functions like `alert` or `fetch`.
Every time I call `alert` I get `TypeError: Illegal invocation` without any further Information and call stack.

```python
import js

js.alert('Hi!')
```

If I assign `window` to the `pyodide.globals` object from js, `window.alert` will work form python
```js
pyodide.globals.window = window
```

Any ideas?

### LINKED ISSUE #960: Configurable "import js"
Add a `pyodide.mountPackage("name", object)` api that mounts `object` at `name`, so that `import "name"` imports `object`, and `from name import field` imports `object.field` as `field`.

For example, the current `js` import would then be installed by saying `pyodide.mountPackage("js", globalThis)` though we would keep the current behavior as the default. We could get rid of `import js` by saying `pyodide.mountPackage("js", undefined)`.

hoodmane:
**Edit**: Removed resolution of #768 since it ran into #788 and fixing both will require a dedicated PR.

This PR implements the feature proposed in #960 and ~resolves/ #768~ (and probably some other open issues with the same cause).
The fix for #768 is to wrap the `js` object we are importing from (by default `globalThis`) in a `JsProxy`. Then `dir` and `getattr` have already been implemented on `JsProxy`. In particular, if the resulting object is a function, `JsProxy.__getattr__` will wrap it in a `JsBoundMethod`. By using `JsProxy` internally to the `JsImports` modules, we get consistent behavior. I also implemented a method `jsproxy` that returns the wrapped `jsproxy`.

I added two new APIS to the `pyodide-js` module: `pyodide.mountPackage(name, object)` and `pyodide.dismountPackage(name)`. `mountPackage(name, object)` makes `object` a new Python module with name `name`. `dismountPackage` removes the python module with name `name` only if it was created with `mountPackage`.

You can of course `mountPackage` an object on top of an existing python package that hasn't been imported yet. This is probably not a good idea but I see no good way to check for that. If you later `dismountPackage(name)` or `del sys.modules[name]` then you can import the real package.

Also, to handle the problem of trash in the object namespace, if `object` is a javascript `Map` then `getattr` will use `Map.get`. I am going to encourage users to do it this way in the docs because I think it is strictly better: with the `Map` the namespace has exactly the things you put into it yourself and nothing more.
(resolves #960)


hoodmane:
Well this #768 but makes #788 appear more frequently, hence the failing tests. Arguably this is an improvement: the bug in #768 partially hid the bug in #788. 

I think the correct move now is to merge `JsBoundMethod` into `JsProxy`, which will partially fix #788. There's still the bug that I described in #788, which will then cause a `JsProxy` that wraps a `PyProxy`, but I just have to modify the logic as suggested in #788 to fix that.

dalcde:
Am I right that you have completely repurposed `jsimport.c` from being a python module into something supplying the mountPackage function?

hoodmane:
No, it still makes python modules, it just dynamically allocates them. Calling `mountPackage("js", globalThis)` creates a dynamic module named "js" that pulls objects off of `globalThis`. Originally `js` was a static module (as in not on the heap) but otherwise the behavior should be pretty similar.

dalcde:
I see. It just no longer creates *the* js module.

src/type_conversion/jsimport.c
```diff
@@ -1,77 +1,350 @@
...
+
+static int
+JsImportDir_init(PyObject* o, PyObject* args, PyObject* kwargs)
+{
+  JsImportDir* self = (JsImportDir*)o;
+  PyObject* module;
+  if (!PyArg_UnpackTuple(args, "__init__", 1, 1, &module)) {
+    return -1;
   }
-  int idval = hiwire_get_global((int)c);
```
dalcde, line 17: These should be the only uses of hiwire_get_global. You can delete that function together with this PR.

hoodmane:
Do you think I should revert the last three commits on this and move them to a different PR? They are sort of unrelated to the original PR I was just trying to make tests pass by fixing unrelated bugs. This PR just exposes more deficiencies of `JsProxy` by using it more consistently.

dalcde:
You can move them to a different PR but I don't think you have to revert them. I'd assume you need them for the tests to have a chance to pass, and if the other PR gets merged first git should sort things out for you

hoodmane:
Well I got tired of trying to hack the current `JsProxy` into working, so I'm restructuring the whole thing into a pile of mixins (which mirrors the way the Python type instantiation system handles types). `PyProxy` also needs to be turned into a pile of mixins like I started to do in #912.

All of my PRs are going to have humongous merge conflicts with each other =(

dalcde:
> All of my PRs are going to have humongous merge conflicts with each other =(
Merge state

That's more reason to split things up into easy-to-merge bit-sizes.

hoodmane:
I'm getting better at that, but it's hard. I suppose a good approach is to do a bunch of stuff on a scratch branch and then look at the results and split it up into different branches. But for instance #971 touches every file but is logically indecomposable.

src/type_conversion/jsimport.c
hoodmane: Well I reverted some changes. Removing this and `JsBoundMethod` and fixing the bugs will have to wait for another PR.

hoodmane:
@rth @dalcde I think this is ready now.

hoodmane:
Okay well now I've got the reference counting wrong again.

hoodmane:
Should be fixed now. I was accidentally freeing `sys.__dict__` causing mayhem and destruction.

Makefile
```diff
@@ -36,6 +36,7 @@ LDFLAGS=\
 	-lstdc++ \
 	--memory-init-file 0 \
 	-s "BINARYEN_TRAP_MODE='clamp'" \
+	-s ASSERTIONS=2 \
```
dalcde, line 39: This change should be removed

Makefile
```diff
@@ -72,7 +73,7 @@ build/pyodide.asm.js: src/main.bc src/type_conversion/jsimport.bc \
 	date +"[%F %T] Building pyodide.asm.js..."
 	[ -d build ] || mkdir build
 	$(CXX) -s EXPORT_NAME="'pyodide'" -o build/pyodide.asm.js $(filter %.bc,$^) \
-		$(LDFLAGS) -s FORCE_FILESYSTEM=1 --preload-file root/lib@lib
+		$(LDFLAGS) $(EXTRA_LDFLAGS) -s FORCE_FILESYSTEM=1 --preload-file root/lib@lib
```
dalcde, line 76: EXTRA_LDFLAGS doesn't appear to be defined

src/tests/test_pyodide.py
```diff
@@ -118,3 +120,147 @@ def eval_code(code, ns):
...
+        pyodide.mountPackage("a", a);
+        pyodide.mountPackage("b", b);
+        return pyodide.runPython(`
+            from a import x
+            from b import x as x2
+            result = [x(), x2()]
+            import a
+            import b
+            result += [a.s, dir(a), dir(b)]
+            # clean up to prevent this test
```
dalcde, line 148: Is this comment incomplete?

src/type_conversion/jsimport.c
```diff
@@ -1,77 +1,416 @@
...
+          // TODO: Do we need to free name or does the module take ownership of it?
+          let name_utf8 = stringToNewUTF8(name);
+          if(_JsImport_mount(name_utf8, obj_id)){
+            _pythonexc2js();
+          }
+          _free(name_utf8);
+        };
+
+        Module.dismountPackage = function(name){
+          let name_utf8 = stringToNewUTF8(name);
```
dalcde, line 400: Would you want to remove the object from hiwire as well?

Makefile
hoodmane: These two are yet again in the "Hood isn't all that great at git category": I didn't want to commit local changes to Makefile, but I said `git commit -a`.

src/tests/test_pyodide.py
hoodmane: I think this stuff is no longer necessary.

src/type_conversion/jsimport.c
hoodmane: [The hiwire object gets removed automatically when the `JsProxy` is freed.](https://github.com/iodide-project/pyodide/blob/2d33d5ead916146d62c1f30a479db685028a976a/src/type_conversion/jsproxy.c#L33)
However, `JsProxy_cNew` does not steal the reference passed into it, so in `Module.mountPackage` I am leaking `obj_id`. Thanks for checking.

hoodmane: These refcounts are really hard work to keep track of.

hoodmane:
Thanks @dalcde!

hoodmane:
@rth I think this one is ready if the CI accepts the merge of this with the "runPython" branch.

rth:
What's the use case for mountPackage / dismountPackage being part of the public API? For the js module we could do it even if it's private..

hoodmane:
Since currently this PR adds to the public API it's probably good to explicitly document why we think this is a good idea.

The original code is a case of "smells like a too weak API": we are hoarding a special capability to ourselves and using it for one specific purpose, but it's not clear why users _wouldn't_ be expected to have other uses for this. Some examples:

1. We should default to mounting `pyodide_js` at that name (cf #922, #992). (Could be done privately.)
2. We should add a `js_utils` package with things like `sleep` and other standard js functions. (Could be done privately.)
3. Suppose someone is using `THREE.js` (#918) or [`tensorflow.js`](https://github.com/iodide-project/pyodide/issues/960#issuecomment-751476350). You can always do `from js import THREE` and then `some_api = THREE.some_api`, but being able to offer `from three_js import some_api` is an improvement.

As a future improvement we could allow an API to mark certain entries as submodules, so to enable `from three_js.submodule import some_api`. I hope you'll find this convincing. Of course there are ways to get around this but are hacks. I think users will be fond of this API.

hoodmane:
@rth Thoughts on this? I think both @dalcde and @oeway have indicated approval of the new public API so there's a mini consensus in favor.

rth:
This particular API sound useful indeed. We should probably start by merging https://github.com/iodide-project/pyodide/pull/1017 

> The original code is a case of "smells like a too weak API": we are hoarding a special capability to ourselves and using it for one specific purpose, but it's not clear why users wouldn't be expected to have other uses for this.

Note that there is a cost of expanding the public API though: we are expected keep backward compatibility as much as possible, since breaking it also not ideal for users, which would constrain all future development. For this reason it's preferable to keep the public API to a minimal, and if we make something public it needs a clear motivating use case.


For the method names, maybe `registerJsPackage` / `unregisterJsPackage`? The mount/unmount has a file-system related ring to it, and I would say what we call loadPackage is actually "a load then mount package". Also with `Js` in the method name, to make it clear that one cannot register natively Python packages?

hoodmane:
> there is a cost of expanding the public API though

Absolutely. I didn't try to evaluate this cost in my previous post but my goal was to argue that this feature is useful and important enough (even though it is just adding syntactic sugar) that it will be worth it to maintain it because in the future when we are maintaining it I think it will still be a good API that adds significant value to users. But yeah we'd better double check the details and make sure they make sense.

> one cannot register natively Python packages?

Uh-oh, what DOES happen if I say:
```javascript
pyodide.runPython("import somepkg");
pyodide.mountPackage("blah", pyodide.globals.somepkg);
```
?
Does it:
1. Work fine now `somepkg` can be imported at `blah` (unlikely),
2. raise an exception (unlikely but would be good),
3. immediately crash,
4. crash as soon as you try to use `somepkg`, or
5. some other wacky thing?

I have no idea. Better find out!

hoodmane:
> For the method names, maybe `registerJsPackage` / `unregisterJsPackage`?

Maybe even `registerJsModule` and `unregisterJsModule` would be better: technically the resulting object I create is a `Module` not a `Package`. 

It could be helpful to add a special field `__submodules__` for the `jsPackage` we are mounting. which holds a list of keys which should be turned into submodules. Then we could support the `from jspkg.submodule import somename` notation.

hoodmane:
> make it clear that one cannot register natively Python packages?

Uh-oh, what DOES happen if I say:
```javascript
pyodide.runPython("import somepkg");
pyodide.mountPackage("blah", pyodide.globals.somepkg);
```
?
Does it:
1. Work fine now `somepkg` can be imported at `blah` (unlikely),
2. raise an exception (unlikely but would be good),
3. immediately crash,
4. crash as soon as you try to use `somepkg`, or
5. some other wacky thing?

I have no idea. Better find out! This needs to be added to the test suite.

hoodmane:
Another improvement I should probably add is to test the `name` argument of `registerJsModule` for whether or not it's a valid python identifier and raise if not.

hoodmane:
@rth I would appreciate it if you could continue to play the skeptic against any APIs I try to add, though I should think you are likely to do this either way whether I appreciate it or not =)

hoodmane:
> what DOES happen if I say: "registerJsPackage("somename", pymodule)

Answer is:

> 5. some other wacky thing?

It doesn't crash but the mounted module is glitched and throws exceptions when you try to do anything with it.

hoodmane:
Well it turns out that my current implementation of this is not right. One should implement this by adding a new entry to `sys.meta_path`. This will have the advantage of working better with native Python mechanisms and will prevent shadowing of known packages.

rth:
> It doesn't crash but the mounted module is glitched and throws exceptions when you try to do anything with it. This is because the jsproxy field ends up not containing a jsproxy but the ordinary Python module.

Yeah, but it's also somewhat expected that it wouldn't work.

> I should probably add is to test the name argument of registerJsModule for whether or not it's a valid python identifier and raise if not.

+1

> Maybe even registerJsModule and unregisterJsModule would be better: technically the resulting object I create is a Module not a Package.

+1

> I should think you are likely to do this either way

Likely :)

> Well it turns out that my current implementation of this is not right. One should implement this by adding a new entry to sys.meta_path

You could try it. Though last time I tried to do something with `sys.meta_path` I found it quite complex to use and despite it being the recommended way now, creating python modules the old way significantly easier. I'm OK with the approach in the current PR even if it allows shadowing existing packages.

hoodmane:
Yeah I've been reading about `sys.meta_path` and it's extremely complicated. Incredibly, in the relevant PEPs the fact that this system is incomprehensible to mortal humans is never raised as a potential downside.

hoodmane:
> Yeah, but it's also somewhat expected that it wouldn't work.

Expected that it won't work currently, but it should work. In this case "working" means throwing an error when you try to mount it.

hoodmane:
@dalcde I rewrote a large amount of this to use `sys.meta_path`, would you mind reviewing again?

hoodmane:
Closed in favor of #1146.

Issue #911, opened 2020-12-21
"use strict" as much as possible (particularly in EM_JS)


hoodmane:
The javascript generated by Emscripten is by default not in strict mode. Strict mode javascript generates errors earlier and makes bugs easier to find and fix. We should `"use strict"` as much as possible.

rth:
+1. https://github.com/iodide-project/pyodide/pull/792 might help, at least for `pyodide.js`file

hoodmane:
Done in EM_JS quite a while ago. For the remaining cases I'm closing this in favor of #1477.

PR #792, opened 2020-11-09
Add pyodide-js library (alternative to pyodide.js and webworker.js)

gzuidhof:
I took the liberty to modernize the `src/pyodide.js` script, I was hoping to contribute this back. This addresses the following pain points I had integrating Pyodide into [Starboard](https://starboard.gg/) [Notebook](https://github.com/gzuidhof/starboard-notebook):

* There is no way to embed `pyodide.js` into your bundle without copy pasting it into your own project and editing it yourself. There are a few independent projects which each have taken pyodide.js and altered it slightly: their update path will be painful if any meaningful changes are made.
* There was no (clean) way to import `pyodide.js` without immediately also loading the pyodide runtime.
* It could use some love: a lot of `var` could be replaced with `const` and `let`, and many promises could be replaced with async functions using `await`.
* There was no way to turn off `console.log` or `console.error` when loading packages, even when specifying your own callback.
* The CDN URL was string-replaced at build time.
* `pyodide.js` was a very large script which was hard to follow. I didn't go too far with large refactorings - it can definetely have another pass.
* There was a fair bit of variable shadowing (`resolve`).

This PR introduces a new `pyodide-js` package which can be published to NPM. 

Changes:
* There are now 4 entrypoints: 
  * `index.mjs` and `index.cjs` for if one wants to bundle pyodide themselves (ES Module & CommonJS)
  *  `webworker.js` and `browser.js` which should be functionallity equivalent to the existing scripts, we can replace `pyodide.js` by `browser.js` and nothing should change from a user's perspective. Here's the [built version](https://gist.github.com/gzuidhof/c29bec0ae55ea52a6780399b0d526cc9) of `browser.js` as the time of creating this PR.
  * Type definitions are also bundled.
* Rollup is used to create the bundles, and type annotations were added with Typescript.
* There is now a `PyodideLoader` class that keeps some of the state, it can be used directly by those choosing to bundle Pyodide-JS themselves.
* The naming of variables was normalized (there was a mixture of `PascalCase`, `camelCase` and `snake_case` before).
* `async` functions and `await` are used throughout, simplifying the code a lot (especially `preloadWasm()`).
* Split helper functions into a separate file `util.ts`.
* Any `iodide`-specific code was removed.
* `pyodideArtifactsUrl` is now preferred to `languagePluginUrl`, `languagePluginUrl` still works as a fallback (the name doesn't make sense outside of iodide's context).
* If you pass in a message or error callback to `loadPackage`, it no longer logs to console by itself.
[...]

PR #979, opened 2020-12-30
Replace var ==> let and added semicolons to the ends of some js lines.

hoodmane:
Automatically replaced all instances of `var` with `let` via:
```sh
find src -type f -name "*.js" -or -name "*.html" -or -name "*.c"  -or -name "*.h" -or -name "*.py" -or -name *.md | xargs sed -E -i 's/\bvar\b/let/g'
```
and then again to the docs folder. Probably should wait to merge this until after some of the other PRs since it's easy to merge them into this.


hoodmane:
This is working toward using modern javascript, related to #911. Another good thing would be to use async/await in `pyodide.js`. Unfortunately `EM_JS` seems to not support modern javascript though.

dalcde:
See also #792. I think overall we should look into moving large EM_JS chunks to independent javascript files, e.g. `pyproxy_init`, giving us access to better tooling etc.

rth:
Have you managed to identify  why the CI is failing?  There are some conflicts as well.

hoodmane:
> Have you managed to identify why the CI is failing?

The issue is here:
```javascript
  let package_name_regexp = '[a-z0-9_][a-z0-9_\-]*'
  let package_uri_regexp =
      new RegExp('^https?://.*?(' + package_name_regexp + ').js$', 'i');
  let package_name_regexp = new RegExp('^' + package_name_regexp + '$', 'i');
```
==> error `package_name_regexp` redeclared

rth:
> The issue is pyodide.js lines 17--20:

That would be easy to fix, wouldn't it? Otherwise this should be good to merge once the CI passes.

hoodmane:
Well now the 6 webworker tests fail and the others pass.

hoodmane:
@dalcde This failure is still pretty useless.

hoodmane:
Furthermore I think this error is the fault of `conftest.py`. Running one specific test
`pytest -k "fire and destroy" src/tests/test_pyproxy.py`
locally gives the same error:
`conftest.JavascriptException: ReferenceError: languagePluginLoader is not defined`

Repeating the test manually in `console.html` by executing the following in browser console:
```javascript
pyodide.runPython(`
        class Foo:
          bar = 42
          def get_value(self, value):
            return value * 64
        f = Foo()
`)
let f = pyodide.pyimport('f');
console.assert(f.get_value(1) === 64);
f.destroy();
f.get_value();
```
Test works as expected:
```javascript
pyodide.asm.js:8 Uncaught Error: Object has already been destroyed
    at Object.getPtr (pyodide.asm.js:8)
    at Object.get (pyodide.asm.js:8)
    at <anonymous>:11:3
```

hoodmane:
Oh interesting, every single firefox test failed but only six chrome ones failed (probably due to real problems).

hoodmane:
In chrome all the webworker tests failed with nondiagnostic timeout errors (but webworker-specific test failures never give diagnostic error messages).

dalcde:
Were we previously genuinely relying on the semantics of var here? What happens if you revert back to var languageLoaderPlugin = ... ?

hoodmane:
I think `var` automatically defines it on the global object but `let` scopes it and so you need `window.languagePluginLoader = ...`

hoodmane:
Confirmed that converting back to `var` makes it so that firefox and chrome fail the same set of tests. Will try `globalThis.languageLoaderPlugin` next.

hoodmane:
As expected, `globalThis.languageLoaderPlugin` works. Thanks for your help @dalcde!

hoodmane:
So now exactly the webworker tests fail. I'm not really sure yet what the best way to test this manually is. 

I think we need to make a version of `console.html` that uses the webworker to enable manual testing of webworkers.

hoodmane:
I'm really surprised how hard this is and how easy #1080 was.

hoodmane:
Well @dalcde's refactor of `pyodide.js` removed most of the `var`s so hopefully now this will work?

rth (approved):

PR #980, opened 2020-12-30
Rename hiwire to jsref

hoodmane:
A natural continuation of #971 in my opinion.


dalcde:
What is the etymology of hiwire anyway?

hoodmane:
I looked on git blame, there is no explanation at all.

rth:
> What is the etymology of hiwire anyway?

I also don't know.

I agree that renaming `Module.hiwire` to `Module.jsref` would be more explicit, but once ones knows that

> hiwire: A super-simple framework for converting values between C and Javascript.

`hiwire_incref` is IMO more specific than `js_incref` because nothing in this name indicates that it uses this framework (or what it does exactly). So generally I have mixed feeling about this change.

hoodmane:
> hiwire_incref is IMO more specific than js_incref because nothing in this name indicates that it uses this framework (or what it does exactly). So generally I have mixed feeling about this change.

What about `jsref_incref` and `jsref_decref`? The repetition of ref slightly bugs me but I think it's better than `js_incref` because as you say that's a bit ambiguous.

I think we should add a "structure of the core C code" that explains the main types, the purposes of the main files, the init conventions, etc. This would very prominently discuss `JsRef`.

This happens all over the code, it's not exactly obscure. We get a `JsRef` from `jsref_array` and remove it via `jsref_decref`.

In a random code example, it seems pretty transparent that the `jsref` stuff is responsible for managing the objects of the type `JsRef`:
```C
static JsRef
_python2js_sequence(PyObject* x, PyObject* map)
{
  JsRef jsarray = jsref_array();
  if (_python2js_add_to_cache(map, x, jsarray)) {
    jsref_decref(jsarray);
    return Js_ERROR;
  }
  size_t length = PySequence_Size(x);
  for (size_t i = 0; i < length; ++i) {
    PyObject* pyitem = PySequence_GetItem(x, i);
    if (pyitem == NULL) {
      jsref_decref(jsarray);
      PyErr_Clear();
      Py_INCREF(x);
      return pyproxy_new(x);
    }
    JsRef jsitem = _python2js_cache(pyitem, map);
    // ... snip
    Py_DECREF(pyitem);
    jsref_push_array(jsarray, jsitem);
    jsref_decref(jsitem);
  }
    // ... snip
  return jsarray;
}
```

rth:
> What is the etymology of hiwire anyway?

Pure speculation on my part, but maybe somehow related to [high-wire](https://www.merriam-webster.com/dictionary/high-wire) and/or [haywire](https://www.thefreedictionary.com/haywire), as in this type conversion framework is risky and/or not always functioning as it should :)

Overall maybe having a more original name is not such a bad thing, instead of `jsref`. `jsref_push_array` etc might be fine, but in the end `hiwire_push_array` would also be OK once there is a documentation about hiwire somewhere.

@dalcde WDYT?

hoodmane:
I do rather like the name hiwire, it has a nice ring to it.

hoodmane:
> as in this type conversion framework is risky and/or not always functioning as it should

Oh that's a cool idea. Ironic though because hiwire is the most solid part of the type conversions system. Honestly the type conversions work darn well considering how different python and javascript are.

dalcde:
I don't have strong opinions on this. I think a more descriptive name is helpful for new contributors if they work on adjacent code that uses hiwire, but it's not too much of a barrier if we explain things clearly.

hoodmane:
I think the consensus of "ehhh who cares" doesn't merit merging this so I will close.

Issue #904, opened 2020-12-20
CI should build and deploy docs


hoodmane:
We don't currently build the docs in the CI. I think we should.

rth:
Yes, we should. I think there is a possible integration service with readthedocs that shows status in CI, but getting RDT access permissions would likely take some time (mdboom is the only maintainer for the RTD project at the moment). So for now building the docs in a separate CircleCI job might be enough.

hoodmane:
Yeah, it'd be very convenient to automatically deploy the current development version docs, but it'd already be a big improvement just to know that they build correctly.

rth:
Fixed with Readthedocs integration.

PR #983, opened 2020-12-31
Build docs in CI

dalcde:


hoodmane:
Based on it's name I'm assuming this will resolve #904.

dalcde:
I think it is reasonable to repurpose #904 to deploy docs in addition to building them

hoodmane:
I changed the title. @rth was already thinking in terms of deploying the docs in CLI in his comments on #904 anyways.

dalcde:
I've gone for the python:alpine image here, but I'm also happy to use the usual image. The slight pain point is that `make` isn't installed, but sphinx's `Makefile` is just a thin wrapper around the `sphinx` command, so I used that directly.

If we eventually intend to deploy the docs, we would probably need to pick a more specialized image anyway.

(I killed CI to save the world some energy)

.circleci/config.yml
```diff
@@ -19,6 +19,23 @@ jobs:
...
+    steps:
+      - checkout
+
+      - run:
+          name: Install prerequisites
+          command: pip install -r docs/requirements-doc.txt
+
+      - run:
+          name: Build docs
+          command: python3 -m sphinx -M html docs docs/_build
```
rth, line 37: Could you store `docs/_build` as artifacts, so we can actually check the output in PR?

rth (commented):
Thanks @dalcde ! Minor comments otherwise LGTM.

In the build logs I see,
> build succeeded, 1 warning.

I think with the default `make html` command warnings are shown (which could be things like incorrect cross-linking etc). Could we also enable it here?

> rth was already thinking in terms of deploying the docs in CLI in his comments on #904 anyways.

If we keep using readthedocs, I don't think we would need to do the deployment ourselves, it would be just one extra CI provider pre-configured by readtheodocs.

dalcde:
On Thu, Dec 31, 2020 at 01:56:54AM -0800, Roman Yurchak wrote:
>I think with the default `make html` command warnings are shown (which could be things like incorrect cross-linking etc). Could we also enable it here?

My build command should be identical to what `make html` calls, unless
sphinx does some magic to detect if it is invoked by make.

>Could you store `docs/_build` as artifacts, so we can actually check the output in PR?

Done

rth:
Great! it's working https://436-316638554-gh.circle-artifacts.com/0/root/repo/docs/_build/html/index.html In the future we could enable https://github.com/larsoner/circleci-artifacts-redirector to have this link auto-generated in the CI status (we could have also done this with https://github.com/iodide-project/pyodide/issues/648 as well but it needs a CORS proxy unfortunately).

PR #987, opened 2020-12-31
Don't include jedi and parso in the core pyodide package

rth:
This makes jedi and parso regular packages, instead of vendoring them in the core package, which in turn will reduce the minimal download size for pyodide by 1MB or so since jedi is quite large.

Also updates to jedi 0.18

Closes #658
Helps toward https://github.com/iodide-project/pyodide/issues/646

Related to https://github.com/iodide-project/pyodide/issues/984


docs/changelog.md
```diff
@@ -16,14 +16,20 @@
 
   As part of the change, Module.checkABI is no longer present.
   [#991](https://github.com/iodide-project/pyodide/pull/991)
-- six is no longer vendored in the main pyodide package, and needs to be loaded
-  explicitly [#1010](https://github.com/iodide-project/pyodide/pull/1010)
+- six, jedi and parso is no longer vendored in the main pyodide package, and needs to be loaded
```
dalcde, line 19: *are* no longer, *need* to be

dalcde (commented):
I suspect this breaks `Module.autocomplete`, since `jedi` is no longer present by default. Perhaps as a stopgap, before we remove `Module.autocomplete`, automatically load jedi anyway during startup?

dalcde:
You can also remove the jedi/parso lines from `.gitignore`.

rth:
> Perhaps as a stopgap, before we remove Module.autocomplete, automatically load jedi anyway during startup?

I added to the changelog that jedi needs to be loaded to use `pyodide.autocomplete`. I would rather not auto-load jedi, as the point of this PR is also to reduce the initial download size.

dalcde:
Fair enough, especially if we intend to remove pyodide.autocomplete all together before the next release

dalcde (approved):

dalcde:
I'm heading off so go ahead and merge when CI passes

Issue #990, opened 2020-12-31
Make pyodide package files keep track of target pyodide version


dalcde:
In the past we attempted to use PYODIDE_PACKAGE_ABI to ensure people don't accidentally load packages compiled for a different ABI. However, we never really updated the number, so in practice it is not effective. This required a custom version of `file_packager.py`, so it was removed in #991.

The solve the original issue, we can consider requiring the js files to include the intended pyodide version, e.g. `numpy-pyodide0.16.1.js`. This would mean that the "abi version" would automatically update with releases

rth:
+1 to do it, the version in the package name sounds like a good solution as well.

hoodmane:
We can do this as a followup to #2027.

rth:
Right now emscripten version is used to track ABI version in wheel, but we may indeed switch to the pyodide version in the future. In any case, currently there is no action needed on this so closing this issue.

PR #991, opened 2020-12-31
Use upstream file packager instead of fork

dalcde:
This removes PYODIDE_PACKAGE_ABI and the corresponding ABI checks. We may introduce new ways to ensure packages are compiled for the right pyodide version in the future.

There is documentation yet to be updated. Addresses #990


rth (commented):
Thanks @dalcde. One comment otherwise LGTM. 

> Addresses #990

You need to use key words "closes" or "fixes" for Github to auto-close the issue (full list in https://docs.github.com/en/free-pro-team@latest/github/managing-your-work-on-github/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword).

Makefile.envs
```diff
@@ -17,6 +17,3 @@ export PYTHONINCLUDE=$(PYODIDE_ROOT)/cpython/installs/python-$(PYVERSION)/includ
 
 # Use env variable if defined, otherwise fallback to './'
 export PYODIDE_BASE_URL?=./
-# This env variable is used to detect pyodide at build time,
-# do not rename it.
-export PYODIDE_PACKAGE_ABI=1
```
rth, line 22: (My comment didn't seem to be saved, so re-writing it).

The issue with removing this env variable, is that we have contributed fixes with it to upstream packages (at least scikit-image and scikit-learn) as a way to detect building it pyodide. Maybe we could keep it here for now, with a comment and open a separate issue on how to properly detect building in Pyodide (also taking into account that we might use cross-compilation in the future). Using `PYODIDE_BASE_URL` would be OK for this PR, but I'm not sure it's a good solution long term.

dalcde:
>> Addresses #990
>
>You need to use key works "closes" or "fixes" for Github to auto-close the issue (full list in https://docs.github.com/en/free-pro-team@latest/github/managing-your-work-on-github/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword).

I'm trying to avoid that so that I can repurpose #990 to tell us to
re-introduce versioning somehow

Makefile.envs
dalcde: We could introduce a `PYODIDE` environment variable specifically for this purpose

Makefile.envs
```diff
@@ -17,6 +17,9 @@ export PYTHONINCLUDE=$(PYODIDE_ROOT)/cpython/installs/python-$(PYVERSION)/includ
 
 # Use env variable if defined, otherwise fallback to './'
 export PYODIDE_BASE_URL?=./
-# This env variable is used to detect pyodide at build time,
-# do not rename it.
+
+# This environment variable is used for packages to detect if they are built
+# for pyodide during build time
+export PYODIDE=1
```
rth, line 23: OK, sounds good. Thanks!

PR #995, opened 2020-12-31
Detect pyodide.js failure in conftest.py

hoodmane:
Before this PR, if pyodide.js throws an error, it got lost. The webdriver would timeout after 40 seconds and give an error message that did not reflect the failure. Now it captures the failure and prints a message.


hoodmane:
We should also refactor pyodide.js to convert all promises to async functions. Async functions have better error handling.

hoodmane:
@dalcde @rth I would appreciate if this could merge quickly, this problem has been making it hard for me to debug other branches.

hoodmane:
Todo in a future PR: pass fatal errors in the C initialization out to Javascript so they can be reported through the same mechanism.

src/templates/test.html
```diff
@@ -1,27 +1,36 @@
...
-            window.languagePluginUrl = './';
-        </script>
-        <script src="./pyodide.js"></script>
-    </head>
-    <body></body>
+  <head>
+  <meta charset="utf-8"/>
+  <script type="text/javascript">
+    window.logs = [];
+    console.log = function(message) {
```
dalcde, line 7: If python is printing to stdout/err instead of calling `console.log` via `js` directly, I think the correct thing to do is to override `Module.print` and `Module.printErr`

dalcde (commented):
Is it possible to add a test that this indeed works?

src/templates/test.html
hoodmane: Maybe we should do both?

dalcde: I would prefer not to override console.log, in general, as that could have surprising consequences down the road.

hoodmane: Just to be clear, though, that was what was in there *before* this change.

hoodmane:
> Is it possible to add a test that this indeed works?

I'm having trouble. I think I need to add another fixture, but I don't really understand how fixtures work. I can add
```javascript
    if(new URLSearchParams(location.search).has("fatal")){
      globalThis.TEST_FATAL = true;
    }
```
to test.html and 
```
  if(globalThis.TEST_FATAL){
    throw new Error("Test fatal error");
  }
```
to pyodide.js but how do I get "?fatal" added to the end of the url?

src/templates/test.html
dalcde: Oh right you simply re-formatted it. We can change it in a separate PR then

dalcde:
Experimenting with this locally, everything seems to work as expected, except if I immediately `reject` in `languagePluginLoader`'s promise function:
```js
var languagePluginLoader = new Promise((resolve, reject) => {
  reject("Foo");
```
In this case the environment doesn't seem to catch the error and merrily goes on (and then I get `pyodide.runPython is not a function`). Is my javascript understanding lacking here?

conftest.py
```diff
@@ -48,18 +48,23 @@ def pytest_addoption(parser):
...
+    return driver.execute_script(
+        "return window.PYODIDE_ERROR && window.PYODIDE_ERROR.message;"
+    )
 
 
-class PackageLoaded:
-    def __call__(self, driver):
-        inited = driver.execute_script("return window.done")
-        return bool(inited)
+def PackageLoaded(driver):
```
rth, line 65: Since these are functions they should be words separated by underscores,
```suggestion
def package_loaded(driver):
```
and same for the other two.

conftest.py
```diff
@@ -93,10 +98,15 @@ def __init__(
             )
         driver.get(f"http://{server_hostname}:{server_port}/test.html")
         try:
-            wait.until(PyodideInited())
+            wait.until(PyodideInited)
         except TimeoutException:
             _display_driver_logs(self.browser, driver)
             raise TimeoutException()
+        # raise Exception("Pyodide initialization failed:")
```
rth, line 105: Please remove commented code.

rth:
Could you sync with upstream master and address remaining comments? I think we could merge even with the known limitation described in https://github.com/iodide-project/pyodide/pull/995#issuecomment-753249628

hoodmane:
Yup will do. This is somewhere on the scale from partial mitigation to a fix, but I believe that it is an improvement.

hoodmane:
> In this case the environment doesn't seem to catch the error and merrily goes on (and then I get pyodide.runPython is not a function). 

Really weird.

> Is my javascript understanding lacking here?

I'm asking the same question.

hoodmane:
Okay I think this is all set now.

hoodmane:
Nope, lots of test failures. I forgot to replace `PackageLoaded` ==> `package_loaded` in a bunch of places.

rth:
Thanks! It's a nice improvement.

PR #998, opened 2021-01-01
pyodide.js refactor

hoodmane:
I refactored `pyodide.js` to make it easier to read and added lots of comments. I tried to keep the behavior the same, I guess we will find out if the CI thinks the behavior is the same. The error branches of this code may not have good test coverage though? I'm not sure.

My main goals for refactor are to increase code legibility and move to more idiomatic javascript: reduce depth of nested functions / blocks, reduce length of nested functions / blocks, use async/await instead of promises (except when directly wrapping callback style code). I also tried to make it clearer how we are interacting with the Module-global callbacks `dependencyCallbacks` and `locateFile` which I found particularly hard to understand in the original code.

This should make future maintenance easier.


hoodmane:
One weird looking bit of code is: `window.addEventListener('error', windowErrorHandler);` inside of `loadPackage`.
The `windowErrorHandler` then cancels the load if the callback is run. I tried throwing an error in the console while `loadPackage` but it doesn't do anything. Is it possible to hit this code path? Can we remove it?

Also, in a future PR, I would make `loadPackage` slightly less noisy. No matter what, it always writes info-level messages to the console.

dalcde:
See also #792

hoodmane:
I merged test-fail-better because without that branch the test failures are not very informative.

rth:
Note that the `pyodide.js` will be completely removed once https://github.com/iodide-project/pyodide/pull/792 is merged (which I would like to do soon). So I'm not sure what's the best way to deal with it, whether we merge this and then try to merge it with changes in https://github.com/iodide-project/pyodide/pull/792, or directly integrate some of these changes into that PR. Also your review on that PR would be very welcome.

hoodmane:
> directly integrate some of these changes into that PR

I think that the best approach is probably for me (or someone else) to manually do that. I think it would be good for the git history to merge this before merging #792 even if #792 then deletes this. I was looking at #792 and I saw the large diff and I thought it would be good to first make a reference version that modified the file in place.

> your review on that PR would be very welcome

Okay will do.

hoodmane:
Also, I think this PR shows some unfortunate behavior of selenium: I cannot replicate these test failures outside of selenium and the diagnostic tools available to me in pytest are few (console.log doesn't seem to produce any info in the test failure that I can read off). Perhaps if I started an interactive selenium session I could get some insight. #995 improves the situation, but it's still pretty janky.

rth:
> I was looking at #792 and I saw the large diff and I thought it would be good to first make a reference version that modified the file in place.

Yes, the diff there is quite large which is not ideal, and I think we can certainly merge improvements like #979 before. For larger changes if we spend time on refactoring/reviewing `pyodide.js` only then to remove it because https://github.com/iodide-project/pyodide/pull/792 implements it differently it feels that that time could be better spent..

> I think this PR shows some unfortunate behavior of selenium: I cannot replicate these test failures outside of selenium

Yes, selenium certainly has some issues and is a big hard to debug in the current setup.

hoodmane:
> only then to remove it because #792 implements it differently it feels that that time could be better spent

Right, this is part of why I tried to maintain the same behavior here as much as possible (in addition to the fact that this is already a fairly large diff).

src/pyodide.js
```diff
@@ -1,105 +1,224 @@
...
-  var package_uri_regexp =
-      new RegExp('^https?://.*?(' + package_name_regexp + ').js$', 'i');
-  var package_name_regexp = new RegExp('^' + package_name_regexp + '$', 'i');
+  // Loading Pyodide
+  let Module = {
+    noImageDecoding : true,
+    noAudioDecoding : true,
+    noWasmDecoding : true,
+    preloadedWasm : {},
+    checkABI : function(ABI_number) {
```
rth, line 39: This can now be removed following #991

src/templates/test.html
```diff
@@ -1,27 +1,36 @@
...
+    window.languagePluginUrl = './';
+  </script>
+  <script src="./pyodide.js"></script>
+  <script type="text/javascript">
+    async function main(){
+      try {
+        await globalThis.languagePluginLoader;
+        window.PYODIDE_READY = true;
+      } catch(e) {
+        window.PYODIDE_ERROR = e;
```
rth, line 29: These are unrelated changes from https://github.com/iodide-project/pyodide/pull/995, right?

rth (commented):
Did a quick review, it's a bit difficult to review because functions are shuffled in the file so the diff shows that everything changed, and one has to search differences function by function manually.  Is there a way to preserve the initial order a bit more so that the diff would be easier to read?

A few additional comments below,

src/templates/test.html
hoodmane: I merged #995 into this in hopes that it would help me understand the CI failure.

hoodmane:
Most of the what I did was reorder stuff in the file, so I don't know that the diff can be too much better. A big part of the problem with the original version is that control flow jumped around a lot in weird ways. 

I moved publicAPI to the top since I think it's the most important thing. I also moved the Module definition to the top because it is the first code that actually is executed and it's confusing when reading the file that all the functions use `Module` a bunch but we don't first see what it is until all the way down on [line 323](https://github.com/iodide-project/pyodide/blob/master/src/pyodide.js#L323).

`loadPackage` used to stretch all the way from line 96 to 261, roughly 45% of the file was this one function. Most of the changes were to this one function. 

I pulled out `chaseDependencies` (was [lines 112--160](https://github.com/iodide-project/pyodide/blob/master/src/pyodide.js#L112),  `fetchPackages` (was [lines 227--251](https://github.com/iodide-project/pyodide/blob/master/src/pyodide.js#L227)), and `installPackages` (was [lines 188--214](https://github.com/iodide-project/pyodide/blob/master/src/pyodide.js#L188) but happened after `fetchPackages` due to weird callback control flow). These steps are: figure out which packages we need to get, download them and let emscripten get them going, and then record the metadata for them.

I also pulled out the management of the Module-global callbacks, which was the most confusing aspect of the original function. (Originally, the assignment to `Module.runDependencies` looked like an API, but then when it was later deleted and updated that was a big surprise.)

So [lines 162 -- 172](https://github.com/iodide-project/pyodide/blob/master/src/pyodide.js#L162) got replaced by `Module.locateFile = getFileLocator(toFetch);` and some of [lines 183--214](https://github.com/iodide-project/pyodide/blob/master/src/pyodide.js#L183) went into `ScriptLoadPromiseManager` which doles out a promise for me to signal when I'm ready to run `installPackages`. [Lines 218--224](https://github.com/iodide-project/pyodide/blob/master/src/pyodide.js#L218) define a suspicious-looking global error event handler, I pulled those out into a separate function.

Other than these factorizations, I also did a bunch of small local changes to use async. These are fairly mechanical / algebraic in nature and are designed to allow code to flow in the normal way, using normal control flow: normal function calls, try/catch, if, and loops. To do async stuff we need two new control flow types: `Promise.all` and `Promise.race`. Define helper async functions for the body of the collected promises.

Before:
```javascript
      for (let pkg in toLoad) {
        let scriptSrc;
        let package_uri = toLoad[pkg];
        if (package_uri == 'default channel') {
          scriptSrc = `${baseURL}${pkg}.js`;
        } else {
          scriptSrc = `${package_uri}`;
        }
        _messageCallback(`Loading ${pkg} from ${scriptSrc}`)
        loadScript(scriptSrc, () => {}, () => {
          // If the package_uri fails to load, call monitorRunDependencies twice
          // (so packageCounter will still hit 0 and finish loading), and remove
          // the package from toLoad so we don't mark it as loaded, and remove
          // the package from packageList so we don't say that it was loaded.
          _errorCallback(`Couldn't load package from URL ${scriptSrc}`);
          delete toLoad[pkg];
          let packageListIndex = packageList.indexOf(pkg);
          if (packageListIndex !== -1) {
            packageList.splice(packageListIndex, 1);
          }
          for (let i = 0; i < 2; i++) {
            self.pyodide._module.monitorRunDependencies();
          }
        });
      }
```
After:
```javascript
  async function fetchPackages(toFetch, messageCallback, errorCallback) {
    let promises = [];
    let fetched = {};
    async function fetchPackageHelper(pkg, package_uri) {
      let scriptSrc;
      try {
        if (package_uri === 'default channel') {
          scriptSrc = `${baseURL}${pkg}.js`;
        } else {
          scriptSrc = `${package_uri}`;
        }
        messageCallback(`Loading ${pkg} from ${scriptSrc}`);
        await loadScript(scriptSrc);
        fetched[pkg] = package_uri;
      } catch {
        errorCallback(`Couldn't load package from URL ${scriptSrc}`);
        ScriptLoadPromiseManager.scriptLoadFailed();
      }
    }

    for (let [pkg, package_uri] of Object.entries(toFetch)) {
      promises.push(fetchPackageHelper(pkg, package_uri));
    }
    await Promise.all(promises);
    return fetched;
  }
```

dalcde:
I think both this and #792 can benefit from a more incremental approach to make it reviewable, e.g. starting with a pass that simply uses modern JS syntax (async/await, let/const) and leaves everything else in place

hoodmane:
Okay, will try again on another branch.

dalcde:
I would not write more code until we decide a way forward

hoodmane:
I was writing the same code incrementally. Do you mean decide on a way forward with #792? I think in any case we should clean up `pyodide.js` first for git history reasons. Looking at #792 it should not be hard to either copy the updated pyodide.js code into that or apply a similar set of changes to the updated code.

dalcde:
Yeah, but for example, if we want to use ES6 modules, the function wrapper wouldn't be so necessary

hoodmane:
Isn't that an easy thing to swap out later though?

hoodmane:
Annoying though that clang-format forces the indentation.

PR #999, opened 2021-01-01
Explicitly tell sqlite popen is not available

dalcde:
sqlite isn't able to detect that, and this becomes a hard error in future versions of emscripten.


rth:
`test_python2js_numpy_scalar`  failed twice with firefox, but I don't see how it could be related to these changes, so LGTM. Thanks.

PR #1002, opened 2021-01-01
Install uglifyjs and lessc via npm

dalcde:
The installation is very fast, and this avoids having two copies of node around. In particular, this avoids the need to monkey-patch uglify-js to use the system node.

This can be further streamlined when #792 is merged


rth:
You can probably remove it as a dependency when building from sources in the docs as well. Otherwise looks good. Thanks.

dalcde:
Should I push the image to docker and update the files when this passes?

rth:
Yes, feel free to push an updated image..

PR #637, opened 2020-04-17
Update to emscripten with upstream LLVM v2

rth:
Another attempt at this. For now mostly https://github.com/iodide-project/pyodide/pull/531 with the latest emscripten version.

PR #531, opened 2019-10-10
WIP: Porting to LLVM upstream

mdboom:
Ref: https://github.com/iodide-project/pyodide/issues/476

This *doesn't* work yet due to some bugs in emscripten.  But hopefully this will be useful to others tackling the same problem.

This is based on work started by @rth, but I had to rearrange the history and his original commits were lost.  I intend to come back and give him authorship credit on those commits before this is finally merged.

PR #1008, opened 2021-01-02
Build with -fPIC

dalcde:
This is needed for the upstream backend, and I believe is a noop under the fastcomp backend (and some packages already build with `-fPIC`, but it doesn't hurt to specify twice)

This is part of the changes in #637, #531, and joemarshall's emsripten 2.0 branch

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -16,7 +16,7 @@ SHELL := /bin/bash
 CC=emcc
 CXX=em++
 OPTFLAGS=-O2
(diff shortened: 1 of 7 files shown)


rth:
Thanks LGTM. I'll add mdboom and @joemarshall as co-authors to the commit once merged. If you adapt some of the other patches, don't hesitate to add a `Co-authored-by: User <email>` in the commit message.

Issue #795, opened 2020-11-09
Better integration with conda/conda-forge for building packages


rth:
The idea to rely on conda-forge for building Python packages to WebAssembly has been mentioned for a while now (https://github.com/iodide-project/pyodide/issues/38#issuecomment-410270067, https://github.com/conda/conda/issues/7619, https://github.com/regro/cf-scripts/issues/1052#issuecomment-651995625), and in this issue I wanted start a discussion about current situation and existing challenges to move in that direction from the perspective of pyodide, as I understand it (please correct me if needed).

First the main motivation is that the present way of building all the packages in one repo is not sustainable with the increase of the number of packages and the associated increase in CI time. To resolve this we would need significant development resources, which we don't have. Even if we did, it would amount to doing many things (including a community) that already exist and work great at conda-forge, which wouldn't make sense.

Now as to challenges (it's a long post),

### 1. Updating emscripten

With a single repo it's relatively fast to rebuild all packages with a different version of the emsdk toolchain (emscripten, binaryen, ..) or different options. We currently still have a couple of patches applied to emscripten, and we also ideally need to update emscripten frequently to benefit from improvements and fixes (currently 1.5 years late with respect to the latest release, unfortunately).  In conda-forge rebuilding all the packages with a new compiler would take longer (though this got better recently https://github.com/regro/cf-scripts/issues/1052#issuecomment-652003811). Also the use-case where a) we update emscripten version b) some package fails to build c) we have to go back and change some global emscripten settings would really be unpractical I think.

This would hopefully become less of an issue with time as emscripten becomes more and more stable, but it's still an issue now (see e.g. https://github.com/iodide-project/pyodide/pull/480#issuecomment-635823062)
     
### 2. Build approach

The cross-compilation of scientific Python packages (based on distutils) is difficult (https://github.com/scipy/scipy/issues/8571, https://github.com/numpy/numpy/issues/17620) as far as I understand, even on Linux between different architectures.

I'm not sure if this was the reason, but pyodide doesn't do cross-compilation in the classical sense. Instead it compiles the package with the host compilers, stores a log of all executed compilation commands and re-run those commands with the emscripten compiler.

### 3. Shared package specifications

Package specifications where chosen as close as possible to the `meta.yaml` in conda, and hopefully soon the package index will also use the same format (https://github.com/iodide-project/pyodide/issues/791)

### 4. Artifacts format

[...]

PR #1009, opened 2021-01-02
Remove unneeded test/imports in package meta.yaml

dalcde:
In particular, in the cases of e.g. scipy, the scipy.\* ones are not needed because when we find imports, we strip everything after the first dot.

### DIFF
--- a/packages/asciitree/meta.yaml
+++ b/packages/asciitree/meta.yaml
@@ -4,6 +4,3 @@ package:
 source:
   sha256: 4aa4b9b649f85e3fcb343363d97564aa1fb62e249677f2e18a96765145cc0f6e
   url: https://files.pythonhosted.org/packages/2d/6a/885bc91484e1aa8f618f6f0228d76d0e67000b0fdd6090673b777e311913/asciitree-0.3.3.tar.gz
-test:
-  imports:
-  - asciitree
--- a/packages/astropy/meta.yaml
+++ b/packages/astropy/meta.yaml
@@ -9,7 +9,3 @@ source:
 requirements:
   run:
     - numpy
-
-test:
-  imports:
-    - astropy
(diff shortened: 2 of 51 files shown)


rth:
> the scipy.* ones are not needed because when we find imports, we strip everything after the first dot.

For `import_name_to_package_name` in [packages.json](https://pyodide-cdn2.iodide.io/dev/full/packages.json), I agree  that we should be stripping everything after the first `.`, but we are not currently doing this. For [import tests](https://github.com/iodide-project/pyodide/blob/a8a0b73642279e78d9d14232b4fd7ef40642b07f/packages/test_common.py#L85) I think we are checking all included imports, not just the first one. That's particularly important in the case of scipy, where `import scipy` imports mostly none of the compiled modules and one has to import submodules explicitly to check.

Also I think it might be better to explicitly keep the import name in the `meta.yaml`, it doesn't cost us much and it makes it clearer for contributors what needs to be done when they use one of those files as an exemple.

dalcde:
> For [import tests](https://github.com/iodide-project/pyodide/blob/a8a0b73642279e78d9d14232b4fd7ef40642b07f/packages/test_common.py#L85) I think we are checking all included imports, not just the first one. That's particularly important in the case of scipy, where `import scipy` imports mostly none of the compiled modules and one has to import submodules explicitly to check.

Fair enough. That explains why it is in the test section.

>Also I think it might be better to explicitly keep the import name in the `meta.yaml`, it doesn't cost us much and it makes it clearer for contributors what needs to be done when they use one of those files as an exemple.

On the contrary, I think it would be a pain to have to exhaustively list all such imports when adding a new package.

rth:
> On the contrary, I think it would be a pain to have to exhaustively list all such imports when adding a new package.

Usually you don't need to list all such import, just the top level import per package is enough, so it's 2 lines per meta.yaml.  I agree that networkx went a bit overboard listing them all. However for some packages (e.g. scipy) you need to tests imports for submodules, as I explained above; removed those imports for would not be correct.

The point is one can't consistently determine the import name from the package name. Yes, `.replace('-', '_')` would work in 75% of cases (here you did replace it in 49 packages / 65) but it won't for the remaining 25% leading to harder to debug errors. Also some packages can include multiple imports, for instance `setuptools` installs `setuptools` and `pkg_ressources`, and such cases are easier to miss if this mapping is done in a way that's hidden from the user.

Finding a better heuristic for this mapping could be interesting when using `find_imports` with PyPi packages, but it's also a more difficult problem.

BTW, this functionality is actually part of the conda `meta.yaml` spec https://docs.conda.io/projects/conda-build/en/latest/resources/define-metadata.html#python-imports it would make sense to avoid diverging from it unless there is a very good reason ref #795

PR #1010, opened 2021-01-02
Build six using usual build system

dalcde:


dalcde:
The list of packages requiring six are obtained by
```
grep -r six packages/*/build/*/{setup.py,setup.cfg,*requirement*}
```

dalcde:
After this PR is merged, we can get rid of some six dependencies by upgrading the packages

rth (approved):
Thanks! LGTM, assuming CI passes.

rth:
We also need to update the docs,
```bash
$ rg -A1 six docs/
docs/loading_packages.md
4:Only the Python standard library and [six](https://pypi.org/project/six/) are 
5-available after importing Pyodide.
```

PR #1012, opened 2021-01-02
Only build wasm-opt in binaryen

dalcde:


dalcde:
This makes building emsdk marginally faster, and paves way for #1013

PR #1013, opened 2021-01-02
Create script to set emscripten env vars

dalcde:
This uses emsdk's emsdk_env.sh, so we no longer have to hardcode the node version and emscripten backend. This also makes debugging more convenient.

This is built on top of the changes of #1012. This is not exactly dependent of it, but #1012 streamlines this commit a bit.


dalcde:
It would be nice to be able to get rid of the hardcoded paths in `emsdk/tests/common.py` as well, but that seems more tricky

set_env.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/bash
```
rth, line 1: Maybe we could call this `pyodide_env.sh` to be more specific.

Makefile.envs
```diff
@@ -1,12 +1,8 @@
...
 export BINARYEN_VERSION = version_86
 
-export PATH := $(PYODIDE_ROOT)/ccache:$(PYODIDE_ROOT)/emsdk/emsdk:$(PYODIDE_ROOT)/emsdk/emsdk/node/12.18.1_64bit/bin:$(PYODIDE_ROOT)/emsdk/emsdk/fastcomp/emscripten/:$(PYODIDE_ROOT)/node_modules/.bin/:$(PATH)
-
-export EMSDK = $(PYODIDE_ROOT)/emsdk/emsdk
-export EM_CONFIG = $(PYODIDE_ROOT)/emsdk/emsdk/.emscripten
-export EM_CACHE = $(PYODIDE_ROOT)/emsdk/emsdk/.emscripten_cache
-export BINARYEN_ROOT = $(PYODIDE_ROOT)/emsdk/emsdk/binaryen
+export BASH_ENV := $(PYODIDE_ROOT)/set_env.sh
+SHELL := /bin/bash
```
rth, line 5: It would be helpful to have a reminder of what BASH_ENV does and indicate the we need to ensure we use bash because of it (and not some other reason).

rth (commented):
I was looking forward to this, thanks! LGTM, aside for minor comments.

In the future it would have been great to be able to do something like,
```
source activate ./pyodide_env.sh
pyodide_build buildpkg numpy
```
although currently it would probably still require `make -C emsdk` first, also it's unclear to me how to better to build the core package/cpython in that context and how to best load cflags etc for `buildpkg`

dalcde:
For pillow and matplotlib, it would be cleaner to use $EM_CACHE in the ldflags and let pywasmcross expand it, but these lines will no longer be necessary when we upgrade to latest emscripten, so I'll leave it

PR #1014, opened 2021-01-02
Enable ccache for emsdk

### LINKED ISSUE #1007: CPython build does not use ccache

dalcde:
It seems difficult to get emconfigure to use ccache, so I abandoned the cpython part. Hopefully with latest emscripten EM_COMPILER_WRAPPER will do the job

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -61,7 +61,7 @@ jobs:
           no_output_timeout: 1200
           command: |
             ccache -z
-            make -C cpython
+            make cpython
             ccache -s
 
       - run:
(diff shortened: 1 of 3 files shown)


dalcde:
Hmm cpython still seems to be built without ccache

dalcde:
So this seems to save ~50s, not impressive but it is also a one-line change

rth:
Are you sure that adding ccache to the path is enough? Currently on master compiling binaryen takes around 10-15min, and it doesn't look like ccache gets used. From the report at the end of the "build emsdk" step,
```
cache hit (direct)                     0
cache hit (preprocessed)               0
cache miss                             0
cache hit rate                      0.00 %
compile failed                         1
cleanups performed                     0
files in cache                      5929
```

PR #1016, opened 2021-01-02
Use Uint32 counter in hiwire

hoodmane:
This is a few small changes to hiwire. By making `counter` into a `Uint32` we get wrapping behavior for free, reducing chances of writing the wrong bitmask or forgetting to bitmask (of course `counter` is only ever touched in a single 13 line method so there's not much risk). 

It also has performance benefits (though this is probably not measurable). Because `counter` isn't local to the function, I think in the previous code the JIT had to convert `counter` from a float to an int and back a lot. Now, we directly advise that `counter` is an int.

In order to take advantage of the native wrapping / overflow behavior of Uint32, I reserve the least significant bit to indicate whether the `JsRef` is a singleton (instead of the most significant bit). I want to reserve `NULL` for error values, so I treat even integers as singletons and odd integers as normal references. 

I also replaced all `JsERROR` instances with `NULL`. This should make error handling more uniform: the Python APIs all use `NULL` as an error code, so we might as well use it too.


src/type_conversion/hiwire.c
```diff
@@ -39,16 +33,20 @@ hiwire_bool(bool boolean)
 }
 
 EM_JS(int, hiwire_init, (), {
-  let _hiwire = { objects : new Map(), counter : 1 };
+  let _hiwire = {
+    objects : new Map(),
+    // Use a native uint32 for our counter. Counter starts at 1.
+    counter : new Uint32Array([1])
```
rth, line 39: This needs more extensive comments, including on overflow behavior and what odd and even values mean.

src/type_conversion/hiwire.c
```diff
@@ -39,16 +33,20 @@ hiwire_bool(bool boolean)
...
+  };
+  // Special values are even. 0 == C NULL is an error code.
```
rth, line 41: Should this be appended to the comment above?

rth (commented):
Thanks @hoodmane !

>  By making counter into a Uint32 we get wrapping behavior for free,

That's a neat trick.

Overall LGTM, but this needs more inline comments with explanations (cf below).

Also if you want to resolve conflicts, and if usual method doesn't work, you might want to do
```
git mv src/type_conversions src/core
git commit -a
```
manually before merging in master.

hoodmane:
> needs more inline comments with explanations

Never hurts! This codebase has a lot of short methods full of really confusing stuff.

> you might want to do [snip] manually before merging in master

I swear git is harder to learn than C. I'm thinking there should be a "Pinball Wizard" parody: "The github wizard has such a subtle wriiiiiist!" Too bad I'm not much good at writing meter.

src/type_conversion/hiwire.c
```diff
@@ -73,14 +71,18 @@ EM_JS(int, hiwire_init, (), {
...
+      // clang-format on
     }
     return _hiwire.objects.get(idval);
   };
 
   Module.hiwire.decref = function(idval)
   {
-    if (idval < 0) {
+    // clang-format off
+    if (idval % 2 === 0) {
```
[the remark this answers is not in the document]
hoodmane: I'm also going to change `% 2` to `& 1` because the latter more literally means "get the least significant bit".

hoodmane:
@rth Can this merge now?

Issue #834, opened 2020-12-07
pyodide scipy.optimize.minimize


11mhg:
I've been trying to get the scipy optimize minimize to work but it seems that no matter the method I use, I get a `apply is undefined` error.

Code to reproduce 
```
%% py
from builtins import range
import random
import autograd.numpy as np
from autograd import value_and_grad

from scipy.optimize import minimize

# Fluid simulation code based on
# "Real-Time Fluid Dynamics for Games" by Jos Stam
# and taken from examples of autograd
# http://www.intpowertechcorp.com/GDC03.pdf




def project(vx, vy):
    """Project the velocity field to be approximately mass-conserving,
       using a few iterations of Gauss-Seidel."""
    p = np.zeros(vx.shape)
    h = 1.0/vx.shape[0]
    div = -0.5 * h * (np.roll(vx, -1, axis=0) - np.roll(vx, 1, axis=0)
                    + np.roll(vy, -1, axis=1) - np.roll(vy, 1, axis=1))

    for k in range(10):
        p = (div + np.roll(p, 1, axis=0) + np.roll(p, -1, axis=0)
                 + np.roll(p, 1, axis=1) + np.roll(p, -1, axis=1))/4.0

    vx -= 0.5*(np.roll(p, -1, axis=0) - np.roll(p, 1, axis=0))/h
    vy -= 0.5*(np.roll(p, -1, axis=1) - np.roll(p, 1, axis=1))/h
    return vx, vy  

def advect(f, vx, vy):
    """Move field f according to x and y velocities (u and v)
       using an implicit Euler integrator."""
    rows, cols = f.shape
    cell_ys, cell_xs = np.meshgrid(np.arange(rows), np.arange(cols))
    center_xs = (cell_xs - vx).ravel()
    center_ys = (cell_ys - vy).ravel()

    # Compute indices of source cells.
    left_ix = np.floor(center_xs).astype(int)
    top_ix  = np.floor(center_ys).astype(int)
    rw = center_xs - left_ix              # Relative weight of right-hand cells.
    bw = center_ys - top_ix               # Relative weight of bottom cells.
    left_ix  = np.mod(left_ix,     rows)  # Wrap around edges of simulation.
    right_ix = np.mod(left_ix + 1, rows)
    top_ix   = np.mod(top_ix,      cols)
    bot_ix   = np.mod(top_ix  + 1, cols)

    # A linearly-weighted sum of the 4 surrounding cells.
[...]

Issue #522, opened 2019-09-11
scipy.optimize.fmin lacks the optional argument "initial_simplex"


tttom:
Hi all,

It appears that `scipy.optimize.fmin` doesn't recognize the initial_simplex argument. This optional input argument is documented [here](https://docs.scipy.org/doc/scipy/reference/generated/scipy.optimize.fmin.html).

Specifically, I call the fmin function as:
`## py`
`import scipy.optimize`
`scipy.optimize.fmin(lambda x: x**2, 0.0, initial_simplex=[[0.0], [1.0]])`
on [https://alpha.iodide.io/tryit](https://alpha.iodide.io/tryit)

I am a bit surprised that something this specific is not implemented while nearly everything else runs perfectly fine with pyodine. Am I missing something? If not, does anybody know where the issue lies?

Thanks,

Tom

rth:
We are unfortunately using a somewhat outdated version of scipy 0.17, so the right documentation for that function would be https://docs.scipy.org/doc/scipy-0.17.0/reference/generated/scipy.optimize.fmin.html

Updating scipy would be good, but it would require some effort and working around Fortran issues https://github.com/iodide-project/pyodide/issues/184

tttom:
Thanks for the quick answer!

rth:
This should be resolved in the next release with https://github.com/pyodide/pyodide/pull/2065

PR #1019, opened 2021-01-02
BLD Add cxxflags to build packages using cpp

mgreminger:
In an effort to add the [Nlopt](https://nlopt.readthedocs.io/en/latest/) package, a `cxxflags` option needs to be added to the meta.yaml format. nlopt now builds and will be a separate pull request if this one is merged. The inclusion of nlopt will address issues #834 and #522 where the current SciPy minimize library is not functional due to Fortran dependencies. Additionally, nlopt is several orders of magnitude smaller than Scipy (~300 KB compared to 166MB), so those who just need the optimization capability will have a much smaller total size and less memory usage issues. The `cxxflags` option is needed since nlopt requires C++11 compatibility and em++ will only compile C++11 features with the -std=c++11 compiler flag. If emcc recieves the -std=c++11 compiler flag for .c files, it errors out. Therefore, a mechanism is needed to provide separate additional flags for C++. gcc will happily compile C++11 features without the flag so setup.py doesn't need to handle this issue to compile nlopt successfully.

The following changes where made:
- `cxxflags` option added and applied in addition to the `cflags` option for C++ source files only
- Now .cc files, in addition to .cpp files, are detected as C++
- Documentation of the new feature added
- Test coverage of the new feature added


dalcde:
Thanks!

mgreminger:
You're welcome, I'll work on getting the NLopt package pull request submitted in the next day or two.

PR #1020, opened 2021-01-03
Rename type_conversion ==> core

hoodmane:
I think the name `type_conversion` is a little bit overly specific. Of the 9 files, only 3 of them are primarily concerned with type conversions. The name `core` is a bit overly general since `pyodide-py` and `pyodide-js` are also part of the core. I also moved `main.c` into `core`, I think it belongs in the same folder as the rest of the C code.


rth (approved):
Yes, I agree it's a bit strange to have to `runpython.c` in `type_conversions`. LGTM.

PR #1021, opened 2021-01-03
Update bleach and packaging

dalcde:


rth:
Note that you can update all pure python packages with,
```
make -C packages update-all
```
Though the results need to be double checked manually afterwards.

PR #1022, opened 2021-01-03
Simplify micropip

dalcde:


packages/micropip/micropip/micropip.py
```diff
@@ -103,40 +103,36 @@ def _parse_wheel_url(url: str) -> Tuple[str, Dict[str, Any], str]:
...
-        if fileinfo.get("digests") is None:
-            # No checksums available, e.g. because installing
-            # from a different location than PyPi.
-            return
-        sha256 = fileinfo["digests"]["sha256"]
-        m = hashlib.sha256()
-        m.update(data.getvalue())
-        if m.hexdigest() != sha256:
-            raise ValueError("Contents don't match hash")
+def extract_wheel(fd):
```
rth, line 106: We probably should call all of these with a leading `_`, e.g. `_extract_wheel` to make it explicit that they are not part of the public API.

PR #1025, opened 2021-01-03
Fixes for new backend

joemarshall:
This contains 4 fixes which are needed before we can make upstream backend work (I'll push the upstream backend fixes as a separate pull request once these are in if that's okay)

I think they don't break the fastcomp backend, but it was one hell of a job getting them merged in with recent changes so there's a chance I broke something

### DIFF
--- a/packages/scipy/patches/make_f2c_subroutines_void.patch
+++ b/packages/scipy/patches/make_f2c_subroutines_void.patch
@@ -1,0 +1,374 @@
+diff --git a/scipy/sparse/linalg/dsolve/SuperLU/SRC/.cgsrfs.c.swp b/scipy/sparse/linalg/dsolve/SuperLU/SRC/.cgsrfs.c.swp
+new file mode 100644
+index 0000000..1acfb3d
+Binary files /dev/null and b/scipy/sparse/linalg/dsolve/SuperLU/SRC/.cgsrfs.c.swp differ
+diff --git a/scipy/sparse/linalg/dsolve/SuperLU/SRC/cgsrfs.c b/scipy/sparse/linalg/dsolve/SuperLU/SRC/cgsrfs.c
+index e7c2bc7..c871681 100644
+--- a/scipy/sparse/linalg/dsolve/SuperLU/SRC/cgsrfs.c
++++ b/scipy/sparse/linalg/dsolve/SuperLU/SRC/cgsrfs.c
+@@ -158,13 +158,13 @@ cgsrfs(trans_t trans, SuperMatrix *A, SuperMatrix *L, SuperMatrix *U,
+     float   *rwork;
+     int      *iwork;
+ 
+-    extern int clacon_(int *, complex *, complex *, float *, int *);
++    extern void clacon_(int *, complex *, complex *, float *, int *);
+ #ifdef _CRAY
+     extern int CCOPY(int *, complex *, int *, complex *, int *);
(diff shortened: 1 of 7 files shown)


rth:
Thanks for doing this work @joemarshall ! Would you mind opening a separate PR for each of these 4 commits? It would be easier to address comments there, they will be faster to merge, and make for a clearer history since we squash commits in PRs?

Why is `python_minifier` directly necessary for the new backend?
Also for `minify_python` and ideally `--ignore-libs` we would need to add unit tests in `pyodide_build/tests/`.
It would be also better to put code in ` pyodide_build/fix_f2c.py` into some function and write a minimal test for it in `pyodide_build/tests/test_fix_f2c.py`

pyodide_build/pywasmcross.py
```diff
@@ -262,6 +268,13 @@ def handle_command(line, args, dryrun=False):
         # Don't include any system directories
         if arg.startswith("-L/usr"):
             continue
+        skip_lib = False
```
dalcde, line 271: I've looked into this more, and I noticed there are three problems arising from using emscripten ports (`-s USE_FOO=1`)
1. Before emscripten 1.39.5, the linker doesn't know where to look for these ports, and so specifying `-lfoo` gives an error. This results in what's at line 323. This can be removed after upgrading emscripten.
2. `-lfoo` looks for `.a` files but we have `.bc` files for some (but not all) ports. This explains why we don't have to supply extra ldflags for `freetype` and `zlib`, but we need it for `libpng`. With the upstream backend, all ports are supplied in the form of `.a`, so we are good here.
3. In matplotlib, it supplies `-lpng16`, but we only have `libpng.a`, not `libpng16.a`. In this case the right course of action is to *replace* `-lpng16` with `-lpng`, not ignore it all together. This problem is common to both 1.38.44+fastcomp and 2.0.x+upstream

So I think, ultimately, there are no libraries that we ought to skip, but instead we should supply replacements

packages/scipy/patches/make_f2c_subroutines_void.patch
```diff
@@ -0,0 +1,374 @@
+diff --git a/scipy/sparse/linalg/dsolve/SuperLU/SRC/.cgsrfs.c.swp b/scipy/sparse/linalg/dsolve/SuperLU/SRC/.cgsrfs.c.swp
```
dalcde, line 1: Can you add some documentation to this patch? Are these changes something that can be done automatically?

dalcde:
Just to echo @rth's comment, splitting these up would also make things easier to debug

pyodide_build/pywasmcross.py
joemarshall: aha, so we could instead of having ignore libraries, have replace-libraries as a flag? So one could replace-libraries png16=png;someotherlibrary.3.2=someotherlibrary or to ignore a library, could just do replace-libraries: gfortran=;

On fastcomp I guess it ought to do something in pywasmcross to fix the bc file problem so -lpng works?

I'll put the different things in different patches in a mo.

dalcde: I think before we fully commit to upstream backend, just having the replace-library flag would do; we have "hacks" in place for the other problems

rth:
As far as I can tell all these changes were merged as part of https://github.com/iodide-project/pyodide/pull/1029 https://github.com/iodide-project/pyodide/pull/1028 and https://github.com/iodide-project/pyodide/pull/1076 so we can close this PR?

PR #1027, opened 2021-01-03
Build C libraries with the Python build system

### LINKED ISSUE #713: Build C libraries with the Python build system

dalcde:
Fixes #713


packages/CLAPACK/meta.yaml
```diff
@@ -1,5 +1,38 @@
...
+    # In CLAPACK's Makefiles, some commands are mistakenly (?) hardcoded
+    # instead of using the right variables
+    sed -i 's/^	-ranlib /^	$(RANLIB)/' **/Makefile
+    sed -i 's/^	ar /^	$(ARCH)/' **/Makefile
+    sed -i 's/^	ld /^	$(LD)/' **/Makefile
+
+    # We build BLAS/LAPACK only for target.
+    # On host we include -LCLAPACK-3.2.1 path which has no effect on host.
+    # On target it gets rewritten by pywasmcross to the full patch of
+    # blas_WA.bc, lapack_WA.bc which are linked statically in scipy
```
rth, line 36:
```suggestion
    # blas_WA.a, lapack_WA.a which are linked statically in scipy
```
with latest commits

pyodide_build/buildpkg.py
```diff
@@ -92,7 +92,11 @@ def download_and_extract(
                 tarballname = tarballname[: -len(extension)]
                 break
 
-        return buildpath / tarballname
+        if "extract_dir" in pkg["source"]:
+            return buildpath / pkg["source"]["extract_dir"]
+        else:
+            return buildpath / tarballname
```
rth, line 98:
```suggestion
        return buildpath / pkg.get("source", {}).get("extract_dir", tarballname)
```

rth (commented):
The changes generally look good, however there is a CI issue. I'm not sure if it's related to this PR or not.

dalcde:
There was both a CI issue and a genuine problem. CI seems to be a lot more flaky lately, which may or may not be due to changes on our part

rth (approved):

rth:
@dalcde Can you add me as a contributor to your fork so that I can restart CI? Failures in scipy on Chrome are suspicious, but they also happened https://github.com/iodide-project/pyodide/pull/987, so it might be unrelated, and CI on master was green..

PR #1028, opened 2021-01-03
build .o files rather than .bc files. 

joemarshall:
asm.js fastcomp build doesn't care about the file extension. upstream (wasm direct) build needs it to be .o

### DIFF
--- a/packages/CLAPACK/Makefile
+++ b/packages/CLAPACK/Makefile
@@ -8,16 +8,16 @@ ROOT=$(abspath .)
 SRC=$(ROOT)/CLAPACK-WA
 
 
-all: $(SRC)/lapack_WA.bc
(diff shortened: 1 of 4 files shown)


packages/CLAPACK/Makefile
```diff
@@ -8,16 +8,16 @@ ROOT=$(abspath .)
 SRC=$(ROOT)/CLAPACK-WA
 
 
-all: $(SRC)/lapack_WA.bc
+all: $(SRC)/lapack_WA.a
```
dalcde, line 11: I think we are already building .a files here; this is a typo in the Makefile. It is good to correct it either way.

joemarshall: yeah it is

dalcde:
Thanks!

PR #1029, opened 2021-01-03
Replace libs

joemarshall:
This is a patch to allow replacing libraries by adding to the yaml
e.g. 
build:
  replace-libs:
    - png16=png

### DIFF
--- a/pyodide_build/pywasmcross.py
+++ b/pyodide_build/pywasmcross.py
@@ -262,6 +262,12 @@ def handle_command(line, args, dryrun=False):
         # Don't include any system directories
         if arg.startswith("-L/usr"):
             continue
+        for l in args.replace_libs.split(";"):
+            if len(l)>0:
+                (from_lib,to_lib) = l.split("=")
+                if len(from_lib)>0 and arg.startswith("-l"+from_lib):
+                    arg= arg.replace("-l"+from_lib,"-l"+to_lib)
+
         # threading is disabled for now
         if arg == "-pthread":
             continue
@@ -451,6 +457,13 @@ def make_parser(parser):
                 "default. Set to 'skip' to skip installation. Installation is "
                 "needed if you want to build other packages that depend on this one."
             ),
+        )
+        parser.add_argument(
+            "--ignore-libs",
+            type=str,
+            nargs="?",
+            default="",
+            help="Libraries to ignore in final link",
         )
     return parser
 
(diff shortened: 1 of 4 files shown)


pyodide_build/pywasmcross.py
```diff
@@ -262,6 +262,12 @@ def handle_command(line, args, dryrun=False):
         # Don't include any system directories
         if arg.startswith("-L/usr"):
             continue
+        for l in args.replace_libs.split(";"):
+            if len(l)>0:
+                (from_lib,to_lib) = l.split("=")
+                if len(from_lib)>0 and arg.startswith("-l"+from_lib):
```
dalcde, line 268: I'm not sure it is a good idea to use startswith instead of equals. For example, `libz` and `libzstd` are very different libraries.

pyodide_build/pywasmcross.py
```diff
@@ -262,6 +262,12 @@ def handle_command(line, args, dryrun=False):
         # Don't include any system directories
         if arg.startswith("-L/usr"):
             continue
+        for l in args.replace_libs.split(";"):
```
dalcde, line 265: Would you want to do this splitting at the beginning of the function instead of repeating it every iteration?

pyodide_build/pywasmcross.py
```diff
@@ -452,6 +458,13 @@ def make_parser(parser):
                 "needed if you want to build other packages that depend on this one."
             ),
         )
+        parser.add_argument(
+            "--ignore-libs",
```
dalcde, line 462: This needs to be updated

pyodide_build/pywasmcross.py
joemarshall: doh, fixing now

pyodide_build/pywasmcross.py
joemarshall: fair enough will do

pyodide_build/pywasmcross.py
joemarshall: just fixed, sorry

dalcde:
In preparation for future changes, we can make matplotlib use this functionality and replace `-lpng16` with `-lpng` at https://github.com/iodide-project/pyodide/pull/1029/files#diff-419e54aa9a46f5a00fc126d21c02d3346048ca9ef2e6f63ca00e6c828a1c481eL325

joemarshall:
sorry for the multiple commits. I've learnt how to use pytest locally now, and tests seem to pass (just waiting for core build)

PR #1033, opened 2021-01-03
Fix serious soundness error in JsProxy_Call kwargs handling

### LINKED ISSUE #974: Bug: JsProxy.__call__ calls the js function with an extra argument
Javascript:
```javascript
window.a = function(){return arguments.length;}
a(1) // ==> 1
a(1,2,3) // ==> 3
```
Python:
```python
from js import a
a(1) # ==> 2
a(1,2,3) # ==> 4
```

hoodmane:
This resolves #974. There is a serious bug in [`JsProxy_Call`](https://github.com/iodide-project/pyodide/blob/d5a7ada6ffe9820b020f1138222ff48b2b0e737d/src/core/jsproxy.c#L126): 
```C
  if (PyDict_Size(kwargs)) {
    JsRef idkwargs = python2js(kwargs);
    hiwire_push_array(idargs, idkwargs);
    hiwire_decref(idkwargs);
  }
```
["If there are no kwargs, the argument kwargs can be `NULL` or an empty dictionary."](https://docs.python.org/3/c-api/call.html?highlight=tp_call#the-tp-call-protocol) This code assumes it will be an empty dictionary but in practice, it's usually `NULL` (to avoid the overhead for making an empty dict). 
[The definition of PyDict_Size is:](https://github.com/python/cpython/blob/3.8/Objects/dictobject.c#L2699)
```C
Py_ssize_t
PyDict_Size(PyObject *mp)
{
    if (mp == NULL || !PyDict_Check(mp)) {
        PyErr_BadInternalCall();
        return -1;
    }
    return ((PyDictObject *)mp)->ma_used;
}
```
So we see that what usually happens is that it sets `BadInternalCall` and returns -1 indicating an error. As usual for this codebase, the calling code doesn't check for errors. What happens next is confusing to me. I think the next python API that is called is `js2python` ==> `JsProxy_cnew` ==> `JsProxyType.tp_alloc` (neither `js2python` nor `JsProxy_cnew` attempt any validation on their arguments). Chasing through `tp_alloc` to find where error handling happens is confusing, but on the master build, the error gets cleared. In various of my branches, I have added better error handling which causes the `BadInternalCall` to be correctly thrown.

The result of all of this is that most of the time javascript functions get called with an extra glitched argument.
```javascript
function f(x){
    window.x = x;
}
pyodide.runPython("from js import f; f()");
x.toString() // ==>  "<NULL>"  (this is PyObject_Print(NULL, stdout, 0);)
x() // ==> Uncaught Error: TypeError: '�' object is not callable
```
Just typing `x.` into the browser console, the browser attempts to lookup the fields on `x` for intellisense and calls `dir` on the null pointer, so as I'm typing these examples the console is belching errors like this:
```python
Python exception: pyodide.asm.js:8:504291
SystemError: frame does not exist
```


hoodmane:
@rth This is ready for review.

rth:
I don't have much review bandwidth this week so I may take a while to respond/review.

hoodmane:
Okay, thanks for letting me know.

src/core/hiwire.c
```diff
@@ -342,6 +342,11 @@ EM_JS(bool, hiwire_is_function, (JsRef idobj), {
   // clang-format on
 });
 
+EM_JS(bool, hiwire_function_supports_kwargs, (JsRef idfunc), {
+  let funcstr = Module.hiwire.get_value(idfunc).toString();
```
dalcde, line 346: You ought to handle the case where the function definition is not available, e.g. it is a wasm function. I think in this case you can assume it does not support kwargs.

hoodmane: That case works automatically actually:
```javascript
>>> pyodide._module.asm.__PyAST_Optimize.toString()
"function 2738() { [native code] }"
```
Maybe I should add comment?

dalcde: That makes sense

hoodmane:
@dalcde @rth I would much appreciate a review of this if either of you have time.

hoodmane:
Just for fun, I posted `supports_kwargs` as a codegolf question:
https://codegolf.stackexchange.com/questions/217511/does-this-javascript-function-support-keyword-arguments

src/core/jsproxy.c
```diff
@@ -128,45 +130,67 @@ JsProxy_SetAttr(PyObject* self, PyObject* attr, PyObject* pyvalue)
...
-  hiwire_CLEAR(idkwargs);
-  hiwire_CLEAR(idresult);
-  if (!success) {
-    Py_CLEAR(pyresult);
-  }
+  JsRef idresult = hiwire_call(JsProxy_JSREF(self), idargs);
+  hiwire_decref(idargs);
+  PyObject* pyresult = js2python(idresult);
+  hiwire_decref(idresult);
+  Py_LeaveRecursiveCall(/* " in JsProxy_Vectorcall" */);
```
rth, line 193: We are no longer doing any error checking in this function?

src/pyodide.js
```diff
@@ -361,6 +361,95 @@ var languagePluginLoader = new Promise((resolve, reject) => {
     return Module.runPython(code);
   };
 
+  Module.function_supports_kwargs = function(funcstr) {
```
rth, line 364: Do we have to re-implement this ourselves? Is no code we could just vendor here, just so that we don't have to extensively check all corner cases?

src/core/jsproxy.c
hoodmane: This is more or less a merge / sequencing mistake on my part. I wrote this while `error_handling.h` was under work and I meant to throw in a comment saying to come back and add the error handling when that branch was merged, I apparently forgot. Thanks for pointing that out, I will fix it!

src/pyodide.js
hoodmane: Good question. I will look on npm. I crowdsourced this a bit by posting it on codegolf:
https://codegolf.stackexchange.com/questions/217511/does-this-javascript-function-support-keyword-arguments
See in particular the one javascript answer:
https://codegolf.stackexchange.com/a/217518/77634
They use `Reflect.parse` which is a `SpiderMonkey` API absent from web standard.

Once we tranform `pyodide.js` into a npm package I think we should use this:
https://www.npmjs.com/package/ast-types

hoodmane: The codegolf folks found exactly one edge case that defeated my implementation: that is if someone defines a function like this:
`function f(...[x, y, { option1, option2}]){}`. This function technically "supports keyword arguments" but my function gives a false negative. However, nobody defined functions like that.

hoodmane: I think given that I have a large number of tests for this and both my test suite and my implementation (minus string support) was vetted by the five people who answered the question on codegolf, it would be reasonable to add this for now.

What do you think?

hoodmane: @dalcde Do you have an opinion?

rth: I'm not opposed to include it if there are no alternatives.

What about performance? I see you cache it, but I imagine there are use cases where one would continuously create new PyProxy objects in a loop and call them? What would be the performance impact of this check?

hoodmane: > What would be the performance impact of this check?

I don't think the performance impact is very high:

1. The check never happens unless you try to call a javascript function with a keyword argument
2. The performance cost: first it copies the string representation of the function, then it scans the string representation doing a small number of comparisons per character. If you are willing to copy the string a second time, I could implement the same algorithm in C, but this sort of code likely JITs extremely well so it's probably better to leave it alone.

rth (approved):

rth:
Please add a changelog entry and maybe update the title to be more specific and mention kwargs.

PR #1035, opened 2021-01-03
Make all fortran subroutines return void

joemarshall:
This fixes bugs in scipy/numpy/CLAPACK where void/int are used interchangeably
which works in normal C and asm.js, but fails in WASM build for upstream emscripten


joemarshall:
I made fix_f2c a function and added a test for it btw

pyodide_build/tests/test_pywasmcross.py
```diff
@@ -71,14 +71,40 @@ def test_handle_command():
...
+      END 
+
+"""
+    )
+    f2c(["gfortran", str(tmp_f)])
+    tmp_c = tmp_path / "temp.c"
+    c_code = tmp_c.read_text()
+    print(c_code)
+    # check that the subroutine returns void not int
+    assert c_code.find("void sayhello") != -1
```
dalcde, line 107: Can we compare all of c_code to a fixed string instead? We would want to make sure fix_f2c doesn't mess up anything else.

dalcde:
Just to understand the situation better, at the fortran-c interface, the return value of a function is dummy and essentially always 0 or void (the actual return value is written to a pointer). f2c opts to make it always return 0, but scipy sometimes expects void and sometimes expect 0.

In this patch you try to make everything return void, which requires patching both the parts of scipy that expects 0, and the output of f2c.

Would it not be simpler to change the parts of scipy that expect void to expect an int instead? A quick grep suggests there are fewer things to change on the scipy side this way too.

joemarshall:
hang on, the changes to package build have broken this slightly

joemarshall:
I went with void because numpy expects void pretty consistently, so it removed the need to patch that.  It also exposes implicit declarations of various externals (we should probably set implicit declarations to error when we switch to upstream build, but right now it could be a massive pain to fix throughout the build) implicit declarations can break at arbitrary points in wasm when they are called with slightly different types (i.e. something that actually is declared as taking int, but it is called with char).

I'm getting failed build on pillow right now, I can't see what I've broken, something to do with freetype ft2build.h

joemarshall:
void is also conceptually nicer - subroutines don't return anything, so they should return void

dalcde:
I see. That makes a lot of sense; I thought numpy didn't use fortran because it didn't use CLAPACK, so I didn't check. Ultimately, we have to deal with the issue that half the packages would expect void and the other half int (and some indecisive). It would be nice if there were a better solution, but I guess this would do for now

joemarshall:
I think once fortran compiler support happens for LLVM, then subroutines surely should return void anyway, so long term this is just a bit of a patch over something that should resolve itself anyway. I'm assuming this just because numpy seems to have swung that way with it's lapack imports, but I may be wrong.

joemarshall:
Also, any idea why pillow isn't building on this branch? I can't see what I've changed that would break it, but it happens both on my machine and in circleci, so I don't think I've just left something hanging in my build here.

dalcde:
I know approximately zero about fortran, but according to this document: https://northstar-www.dartmouth.edu/doc/solaris-forte/manuals/fortran/prog_guide/11_cfort.html the return value can sometimes be meaningful (although it is "implementation dependent and its use should be avoided"). If C code somehow decides to use the return value, us turning it into void will break the code significantly. On the other hand, if a function originally returns void and we make it return int, there is less possibility of breakage. So making everything return int seems safer imo

joemarshall:
Oh god, I'd forgotten about alternate returns in F77, I think it is 22 years since I last got paid to write any fortran. Those are disgusting and I hope nobody uses them in practice in any modern code, but given F77 supports them I guess return int is correct then and numpy is wrong. Will make this patch again but so it works the other way then. I guess it saves having to fix f2c. Incidentally, building clapack as .so which should work fine in upstream also reduces the pain of this kind of problem, whilst we still use FPcast emulation.

I'll close this pr and make another one from the base rather than piling up commits.

dalcde:
I think your branch actually breaks the building of all C-based libraries, but fails silently for other packages.

PR #1039, opened 2021-01-04
Fix commented trailing semicolon in eval_code.

casatir:
Semicolon at the end of a comment does not return last expression's result.

``` python
eval_code("1+1 # comment;")
```
returns `None` instead of `2`.

This should be fixed by this PR.

### DIFF
--- a/src/pyodide-py/pyodide/_base.py
+++ b/src/pyodide-py/pyodide/_base.py
@@ -9,6 +9,7 @@ from asyncio import iscoroutine
 from io import StringIO
 from textwrap import dedent
 from typing import Dict, List, Any, Tuple
+import tokenize
 
 
 class JsException(Exception):
@@ -40,6 +41,31 @@ def open_url(url: str) -> StringIO:
     req.open("GET", url, False)
     req.send(None)
     return StringIO(req.response)
+
+
+def quiet_code(code: str) -> bool:
+    """
+    Does the last nonwhitespace character of code is a semicolon?
+
+    This can be overridden to customize the way eval_code is silenced.
+    """
+    # largely inspired from IPython:
+    # https://github.com/ipython/ipython/blob/86d24741188b0cedd78ab080d498e775ed0e5272/IPython/core/displayhook.py#L84
+
+    codeio = StringIO(code)
+    tokens = list(tokenize.generate_tokens(codeio.readline))
+
+    for token in reversed(tokens):
+        if token.type in (
+            tokenize.ENDMARKER,
+            tokenize.NL,
+            tokenize.NEWLINE,
+            tokenize.COMMENT,
+        ):
+            continue
+        return (token.type == tokenize.OP) and (token.string == ";")
+
+    return False
 
 
 def eval_code(code: str, ns: Dict[str, Any]) -> None:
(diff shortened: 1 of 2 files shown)


hoodmane:
Right, `code.trim().endswith(";")` seems a little suspicious but I didn't think of comments. Good catch!

src/pyodide-py/pyodide/_base.py
```diff
@@ -42,6 +43,31 @@ def open_url(url: str) -> StringIO:
...
+def quiet_code(code: str) -> bool:
+    """
+    Does the last nonwhitespace character of code is a semicolon?
+
+    This can be overridden to customize the way eval_code is silenced.
+    """
+    # largely inspired from IPython:
+    # https://github.com/ipython/ipython/blob/86d24741188b0cedd78ab080d498e775ed0e5272/IPython/core/displayhook.py#L84
+
+    codeio = StringIO(code)
```
hoodmane, line 55:
```suggestion
    # We need to wrap tokens in a buffer because:
    # "Tokenize requires one argument, readline, which must be a callable object which provides the same interface as the io.IOBase.readline() method of file objects"
    codeio = StringIO(code)
```

src/pyodide-py/pyodide/_base.py
```diff
@@ -42,6 +43,31 @@ def open_url(url: str) -> StringIO:
...
+    tokens = list(tokenize.generate_tokens(codeio.readline))
+
+    for token in reversed(tokens):
+        if token.type in (
+            tokenize.ENDMARKER,
+            tokenize.NL,
+            tokenize.NEWLINE,
```
hoodmane, line 62: It looks like `NL` starts each line and `NEWLINE` ends the line? Do you need to skip both in case the line is completely empty?

hoodmane (approved):

src/pyodide-py/pyodide/_base.py
hoodmane: Would have been friendly of the `tokenize` package to have an api that accepts a normal iterable.

src/pyodide-py/pyodide/_base.py
casatir: It depends on what we want. To me, evaluating `"1+1;\n\n"` should return `None` and `1+1\n\n` should return `2`. In that case, `NL` should be skipped. WDYT?

src/pyodide-py/pyodide/_base.py
casatir: You're right!

src/pyodide-py/pyodide/_base.py
casatir: I'll add tests for that.

hoodmane: > To me, evaluating `"1+1;\n\n"` should return `None` and `1+1\n\n` should return `2`. In that case, `NL` should be skipped.

hoodmane:
Looks great. (These random timeout exceptions showing up in the CI all the time are driving me crazy though.)

src/pyodide-py/pyodide/_base.py
```diff
@@ -42,6 +43,31 @@ def open_url(url: str) -> StringIO:
     return StringIO(req.response)
 
 
+def quiet_code(code: str) -> bool:
+    """
+    Does the last nonwhitespace character of code is a semicolon?
+
+    This can be overridden to customize the way eval_code is silenced.
+    """
```
rth, line 51: Could you add a docstring example here,

```
Examples
--------
>>> quiet_code('x + 1')
False
>>> quiet_code('x +1 ;')
True
>>> quiet_code('x +1 # comment ;')
False
```
which would help understanding what it does also serve as a doctest

rth (approved):
Thanks @casatir  (and @hoodmane for the review)!

> random timeout exceptions showing up in the CI all the time

They seem to be happening more often lately. Sometimes I feel like studying the statistical distributions of those CI failures would be an interesting side project.

PR #1040, opened 2021-01-04
Added a monkey patch disabling _path_importer_cache

hoodmane:
I'm not sure if this is a good choice, but it seems like an option to consider.

We've had various troubles with the import cache and the EmscriptenFS date modified times. This completely disables the cache which should prevent any need for `importlib.invalidate_cache` and any bugs related to import cache from ever coming up again.

I'm not sure what the performance implications of this are. We could also implement a different caching strategy which is more conservative than the normal one but less conservative than always refetching from scratch.

### DIFF
--- a/src/pyodide.js
+++ b/src/pyodide.js
@@ -249,12 +249,6 @@ var languagePluginLoader = new Promise((resolve, reject) => {
           }
         });
       }
-
-      // We have to invalidate Python's import caches, or it won't
-      // see the new files. This is done here so it happens in parallel
-      // with the fetching over the network.
-      self.pyodide.runPython('import importlib as _importlib\n' +
-                             '_importlib.invalidate_caches()\n');
     });
 
     return promise;
--- a/src/pyodide-py/pyodide/__init__.py
+++ b/src/pyodide-py/pyodide/__init__.py
@@ -1,5 +1,10 @@
 from ._base import open_url, eval_code, find_imports, as_nested_list, JsException
+from . import _importhooks
 from .console import get_completions
+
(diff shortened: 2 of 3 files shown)


src/pyodide.js
```diff
@@ -249,12 +249,6 @@ var languagePluginLoader = new Promise((resolve, reject) => {
           }
         });
       }
-
-      // We have to invalidate Python's import caches, or it won't
-      // see the new files. This is done here so it happens in parallel
-      // with the fetching over the network.
-      self.pyodide.runPython('import importlib as _importlib\n' +
-                             '_importlib.invalidate_caches()\n');
```
hoodmane, line 257: Maybe we should keep this?

hoodmane:
@rth I think this PR is a bad idea if the existing solutions for cache problems work (what's the status of that). However, I think it would be good to keep this around as documentation if cahce problems ever come up again.

rth:
> I'm not sure what the performance implications of this are.

You could check  `benchmarks.json` in the artifacts tab of the `benchmarks` CircleCI job and compare it to its value on master. Those benchmarks results are fairly noisy though.

rth:
> what's the status of that

Do you mean https://github.com/iodide-project/pyodide/pull/893 ? That was merged.

> We've had various troubles with the import cache and the EmscriptenFS date modified times.

Could it be due to the above PR?

hoodmane:
Ah no, I was referring to #737, I haven't seen this problem myself. #893 says it "Partially solves #737". What is partial about that solution?

hoodmane:
Okay, since #737 works, this is unnecessary.

Issue #245, opened 2018-11-07
Asyncio support


oeway:
Hi, following a discussion here: https://github.com/iodide-project/pyodide/issues/228 I think asyncio is not working right now. 

Would be great to have that supported to call JavaScript functions with async/await in Python.

mdboom:
That's correct.  The default implementation of `asyncio` in Python uses `select`, which isn't available.  This would probably require writing a main loop based on Javascript primitives instead.

In the meantime, you can already use Promises to do most of the things that require async in Javascript, just without the nice syntax.

amirouche:
@mdboom what do you mean by "use Promises" exactly?

AFAIU, One must write an `asyncio` event loop based on [`requestAnimationFrame`](https://developer.mozilla.org/en-US/docs/Web/API/window/requestAnimationFrame), is that correct?

mdboom:
By "use Promises", I mean that the Web APIs that return promises (e.g. `fetch`) work just fine from Python, so you can still do async things without the nice Python async syntax.

I'm not sure what primitives one would build an event loop on -- I had thought `setTimeout`.  But that's the right idea, I think.

oeway:
@mdboom coming back to this question, you mentioned that `select` is not available. Would it be possible to have `select` in pyodide and then `asyncio` will be supported?

mdboom:
One would have to write a select implementation in Javascript built on Web API primitives.  (This is how all the other system calls are implemented in emscripten).  I haven't spent any time looking into how difficult that would be.  Theoretically, if one could write on that's POSIX compliant (presumably only for local "virtual" sockets since remote ones wouldn't work) then Python's asyncio would work.

oeway:
Thanks for the clarification, giving how different(better) it is to use async/await compare to callback style, I would really like to see this feature being supported in the near future. I may also have a look at what you mentioned above, but I am not familiar with these low level libraries.

AKKlostermann:
My idea was not to implement "select" and such, but rather to write a coroutine-based replacement which uses Javascript promises under the hood. This would not support the asyncio libraries, most of which make no sense in a browser anyway.

Instead it would support things like 'sleep'(via a setTimeout wrapper), fetch, websocket messages and so on. A lot of Web APIs can be converted to promises and that would be the easiest way to go.

mdboom:
@AKKlostermann : Sure.  I'd love to see a proof-of-concept of that if you get around to it.

oeway:
[...]

PR #1041, opened 2021-01-04
Simplifying eval_code and avoid injecting dummy variable.

casatir:
#876 brings interesting features to `eval_code`. However, I am not an `AST` expert but the resulting code looks overly complicated to me:

- is the introduction of ̀`<EXEC-LAST-EXPRESSION>` to assign last expression considered good practice?
- is it really safe?
- the redundancy between `eval_code` and `_eval_code_async` doesn't follow the DRY principle
- the previous `mod.body` splitting was nice and clear and can be kept
- `_adjust_ast_to_store_result` and `_adjust_ast_to_store_result_helper` can be avoided (no need to test them!)

This PR aims at recovering the clearness of `eval_code` while supporting the semicolon feature and factorizing  `eval_code` and `_eval_code_async`.

A review by @hoodmane will be nice as he probably wrote it like this for some reason that eludes me.

Lastly, I should mention that this splitting method is what is [used in IPython ](https://github.com/ipython/ipython/blob/3587f5bb6c8570e7bbb06cf5f7e3bc9b9467355a/IPython/core/interactiveshell.py#L3229) apart from the fact that last expression is evaluated in `single`mode (the result is then caught by `sys.displayhook`).

### DIFF
--- a/src/pyodide-py/pyodide/_base.py
+++ b/src/pyodide-py/pyodide/_base.py
@@ -43,7 +43,7 @@ def open_url(url: str) -> StringIO:
     return StringIO(req.response)
 
 
-class CodeRunner(object):
+class CodeRunner:
     """
     A code runner to serve eval_code and eval_code_async.
 
@@ -61,9 +61,9 @@ class CodeRunner(object):
         Parameters
         ----------
         code
-        the Python code to run.
+           the Python code to run.
         ns
-        `locals()` or `globals()` context where to execute code.
+           `locals()` or `globals()` context where to execute code.
         """
         # handle mis-indented input from multi-line strings
         code = dedent(code)
@@ -93,11 +93,11 @@ class CodeRunner(object):
 
         Examples
         --------
-        >>> CodeRunner('x + 1', {}).quiet()
+        >>> CodeRunner('1 + 1', {}).quiet()
         False
-        >>> CodeRunner('x + 1 ;', {}).quiet()
+        >>> CodeRunner('1 + 1 ;', {}).quiet()
         True
-        >>> CodeRunner('x + 1 # comment ;', {}).quiet()
+        >>> CodeRunner('1 + 1 # comment ;', {}).quiet()
         False
         """
         # largely inspired from IPython:
--- a/src/pyodide-py/pyodide/_base.py
+++ b/src/pyodide-py/pyodide/_base.py
@@ -65,8 +65,6 @@ class CodeRunner:
     """
 
     def __init__(self, ns: Dict[str, Any] = None):
-        # handle mis-indented input from multi-line strings
-
         self.ns = ns if ns is not None else {}
         self.filename = "<exec>"
 
@@ -119,6 +117,7 @@ class CodeRunner:
         code object
             last expression's code object (or None)
         """
+        # handle mis-indented input from multi-line strings
         code = dedent(code)
 
         mod = ast.parse(code)
--- a/src/pyodide-py/pyodide/_base.py
+++ b/src/pyodide-py/pyodide/_base.py
@@ -43,41 +43,137 @@ def open_url(url: str) -> StringIO:
     return StringIO(req.response)
 
 
-def quiet_code(code: str) -> bool:
+class CodeRunner(object):
     """
-    Does the last nonwhitespace character of code is a semicolon?
-
-    This can be overridden to customize the way eval_code is silenced.
+    A code runner to serve eval_code and eval_code_async.
 
     Examples
     --------
-    >>> quiet_code('x + 1')
-    False
-    >>> quiet_code('x + 1 ;')
-    True
-    >>> quiet_code('x + 1 # comment ;')
-    False
+    >>> CodeRunner("1+1", {}).run()
+    2
+    >>> CodeRunner("1+1;", {}).run()
     """
-    # largely inspired from IPython:
-    # https://github.com/ipython/ipython/blob/86d24741188b0cedd78ab080d498e775ed0e5272/IPython/core/displayhook.py#L84
 
-    codeio = StringIO(code)
-    tokens = list(tokenize.generate_tokens(codeio.readline))
+    def __init__(self, code: str, ns: Dict[str, Any]):
+        """
+        Constructor.
 
-    for token in reversed(tokens):
-        if token.type in (
-            tokenize.ENDMARKER,
-            tokenize.NL,  # ignoring empty lines (\n\n)
-            tokenize.NEWLINE,
-            tokenize.COMMENT,
-        ):
-            continue
-        return (token.type == tokenize.OP) and (token.string == ";")
+        Parameters
+        ----------
+        code
+        the Python code to run.
+        ns
+        `locals()` or `globals()` context where to execute code.
+        """
+        # handle mis-indented input from multi-line strings
+        code = dedent(code)
 
-    return False
+        self.code = code
+        self.ns = ns
+        self.filename = "<exec>"
+        self.mod = None
+        self.last_expr = None
+
+        mod = ast.parse(code)
+        if len(mod.body) == 0:
+            return
+
+        # we extract last expression
+        last_expr = None
+        if isinstance(mod.body[-1], (ast.Expr, ast.Await)) and not self.quiet():
+            last_expr = ast.Expression(mod.body.pop().value)  # type: ignore
+        self.mod = mod
+        self.last_expr = last_expr
+
+    def quiet(self) -> bool:
+        """
+        Does the last nonwhitespace character of code is a semicolon?
+
+        This can be overridden to customize the way run() is silenced.
+
+        Examples
+        --------
+        >>> CodeRunner('x + 1', {}).quiet()
+        False
+        >>> CodeRunner('x + 1 ;', {}).quiet()
+        True
+        >>> CodeRunner('x + 1 # comment ;', {}).quiet()
+        False
+        """
+        # largely inspired from IPython:
+        # https://github.com/ipython/ipython/blob/86d24741188b0cedd78ab080d498e775ed0e5272/IPython/core/displayhook.py#L84
+
+        # We need to wrap tokens in a buffer because:
+        # "Tokenize requires one argument, readline, which must be
+        # a callable object which provides the same interface as the
+        # io.IOBase.readline() method of file objects"
+        codeio = StringIO(self.code)
+        tokens = list(tokenize.generate_tokens(codeio.readline))
+
+        for token in reversed(tokens):
+            if token.type in (
+                tokenize.ENDMARKER,
+                tokenize.NL,  # ignoring empty lines (\n\n)
+                tokenize.NEWLINE,
+                tokenize.COMMENT,
+            ):
+                continue
+            return (token.type == tokenize.OP) and (token.string == ";")
+
+        return False
+
+    def _compile_mod(self, flags: int = 0x0):
+        """
+        Compile first part of the code (everything but last expression)
+        """
+        return compile(self.mod, self.filename, "exec", flags=flags)  # type: ignore
+
+    def _compile_last_expr(self, flags: int = 0x0):
+        """
+        Compile last part of the code (last expression)
+        """
+        return compile(self.last_expr, self.filename, "eval", flags=flags)  # type: ignore
+
+    def run(self) -> Any:
+        """
+        Returns
+        -------
+        If the last nonwhitespace character of code is a semicolon,
+        return `None`.
+        If the last statement is an expression, return the
+        result of the expression.
+        """
+        # running first part
+        if self.mod is not None:
+            exec(self._compile_mod(), self.ns, self.ns)
+
+        # evaluating last expression
+        if self.last_expr is not None:
+            return eval(self._compile_last_expr(), self.ns, self.ns)
+
+    async def run_async(self) -> Any:
+        """ //!\\ WARNING //!\\
+        This is not working yet. For use once we add an EventLoop.
+
+        Note: see `_eval_code_async`.
+        """
+        raise NotImplementedError("Async is not yet supported in Pyodide.")
+        flags = ast.PyCF_ALLOW_TOP_LEVEL_AWAIT
+        # running first part
+        if self.mod is not None:
+            coro = eval(self._compile_mod(flags), self.ns, self.ns)
+            if iscoroutine(coro):
+                await coro
+
+        # evaluating last expression
+        if self.last_expr is not None:
+            res = eval(self._compile_last_expr(flags), self.ns, self.ns)
+            if iscoroutine(res):
+                res = await res
+            return res
 
 
-def eval_code(code: str, ns: Dict[str, Any], compile_flags: int = 0) -> Any:
+def eval_code(code: str, ns: Dict[str, Any]) -> Any:
     """Runs a code string.
 
     Parameters
@@ -86,8 +182,6 @@ def eval_code(code: str, ns: Dict[str, Any], compile_flags: int = 0) -> Any:
        the Python code to run.
     ns
        `locals()` or `globals()` context where to execute code.
-    compile_flags
-       AST compile flags.
 
     Returns
     -------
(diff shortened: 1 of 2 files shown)


hoodmane:
Thanks @casatir. My code was first-thing-that-works code that I have repurposed several times and so it has vestigial crap from supporting several different sets of features that are absent from this code and also things in it that were never necessary. Most of what you've done I think is an improvement, however the definition of `eval_code_async` you have does not work. Correctly implementing `eval_code_async` took me a lot of time, I first had a whole series of subtly and less-subtly broken variants.

hoodmane:
>this splitting method is what is used in IPython

I think several different methods are used in IPython:
1. splitting
2. wrapper function with `co_newlocals` set `False` (makes local variables in the function body write directly into enclosing frame)
3. storage into temporary result

IPython's support for `async` is rather inelegant because it doesn't fit very well with past features.

hoodmane:
You cannot use `exec` with `async` because `exec` returns `void`. If you `exec` code compiled with `PyCF_ALLOW_TOP_LEVEL_AWAIT` and an await occurs in the code, it generates a coroutine and then drops it, which means nothing actually happens and the code generates `Runtime Warning: coroutine never awaited.`

You could use the splitting method like:
```python
if len(mod.body):
    coroutine_or_garbage = eval(compile(mod, "<exec>", "exec", flags=compile_flags), ns, ns)
    if iscoroutine(coroutine_or_garbage):
       await coroutine_or_garbage
if last_expr is not None:
    result = eval(compile(last_expr, "<exec>", "eval", 
    if iscoroutine(result):
       result = await result
    return result
```

hoodmane:
So I think the only design constraint from the original code that you are missing is that in the `async` version you must only use `eval` and every time you `eval` you must store the result and `await` it if it is a coroutine. With that in mind I decide it was better to store into a temporary global.

> is `<EXEC-LAST-EXPRESSION>` really safe?

Only way you could run into trouble is if someone is trying to use `globals()["<EXEC-LAST-EXPRESSION>"]` for something else?

hoodmane:
Lastly, we should talk about what the eventual API this code should support will look like when Pyodide is run on a WebWorker. The API I have in mind looks like:
```rust
Executor.run_python(code : str) -> Execution

Execution.set_stdout_callback(callback : (str) -> void ) -> void
Execution.set_stderr_callback(callback : (str) -> void ) -> void
async Executor.syntax_check() -> Option<SyntaxError>
async Executor.result() -> Result<any, any>
```
In the lifecycle of an execution there are four relevant types of events: stdout writes, stderr writes, whether the syntax check passes, and when the code has run to completion. stdout writes and stderr writes each happen 0 or more times, the syntax check and getting the result (either a value or an error). This API makes it quite easy for the UI to do special handling for syntax errors which is usually desirable.

casatir:
Thanks @hoodmane for the review.

> however the definition of `eval_code_async` you have does not work.

That's what I thought, I wrote it blindly: how do you test it? By the way, is your version really working? 

``` python
if iscoroutine(res):
    await res
```

shouldn't it be:

``` python
if iscoroutine(res):
    return await res
```

In one hand, it is important to prepare `async` support (thanks for your work in that way) but in the other hand, there is no value in having tricky code in main places like `eval_code` only to implement `eval_code_async` that is not tested and not supported yet. WDYT?

> 3. storage into temporary result

Can't see this in IPython's code, at least not the way you do it. Maybe you talk about [this](https://github.com/ipython/ipython/blob/3587f5bb6c8570e7bbb06cf5f7e3bc9b9467355a/IPython/core/interactiveshell.py#L3414)? The temporary variable is in a separate namespace.

> So I think the only design constraint from the original code that you are missing is that in the async version you must only use eval and every time you eval you must store the result and await it if it is a coroutine. With that in mind I decided it was better to store into a temporary global. Using the temporary global approach, you can in fact unify eval_async and eval.

Your right, I miss this point. Wouldn't it be cleaner to wrap it in an `async` function? ala IPython?

> >    is <EXEC-LAST-EXPRESSION> really safe?

> Only way you could run into trouble is if someone is trying to use globals()["<EXEC-LAST-EXPRESSION>"] for something else?

Imagine a custom namespace (`dict`) that counts how many named objects are created during execution. It gets +1 with your approach and the user can't explain it.

Your WebWorker API sounds good but does it has an impact on the way `eval_code` should be designed? Only `eval_code_async` though.

casatir:
As far as I understand, the common part in `eval_code` and `eval_code_async` is mainly the splitting. What about something like this?

```python

def eval_code(...):
    runner = CodeRunner(...)
    return runner.run()

async def eval_code_async(...):
    runner = CodeRunner(...)
    return await runner.run_async()
```

with the class `CodeRunner` managing the splitting and other common stuff.

hoodmane:
> By the way, is your version really working?

Yup, both in a native repl and in pyodide. The pyodide version you may try out here:
https://spectralsequences.com/repl/
```python
from js_wrappers.async_js import wrap_promise
resp = await wrap_promise((await wrap_promise(fetch("api/charts/does-this-exist"))).json())
resp.code # ==> 'get-chart::failed::not-found'
```
Since the version I dumped in here is not tested it is best assumed to be broken by the "if it isn't tested it is broken" principle, but it is very similar to well-tested stuff in my private code so I have confidence.

> shouldn't it be:
```python
return await res
```
For some annoying reason `return await` is a syntax error. I think this is planned to fix in v3.10? But no, the return value for mine is always `None`. In fact there is no way to return anything the normal way without the wrapper function.

> Your right, I miss this point. Wouldn't it be cleaner to wrap it in an async function? ala IPython?

I found that their wrapper function approach causes bugs with self assignment. I think this is what happened:

```python
>>> x = 5
>>> x = x + 1 # ==> Unbound Local Error: Local variable x referred to before assignment
```
Whereas
```
>>> x = 5
... x = x + 1
```
works fine. This is weird, I assume that their version of the code must work, but whenever I try it I get subtle bugs. It's worth pointing out that in `3.8` there is an API like `new_code_object = CodeObject.replace(co_obj, co_new_locals = False)` so if you go this way you don't have to repeat all of the fields of the code object like they do in IPython.

hoodmane:
> there is no value in having tricky code in main places like eval_code only to implement eval_code_async that is not tested and not supported yet. 

There is absolutely value in this. In this case, in order to support async/await there is at least one significant PR and several minor fixes that need to be made to core, another PR to add a web loop, and eval_code_async needs to work. Once we have all the pieces, we will presumably need an extra PR to work out kinks.
If we need to make changes ABCDE in order to achieve some new functionality, but by itself each of those changes introduces unnecessary complexity to the code and adds no user-facing features, then your logic would prevent ever implementing the new functionality.

hoodmane:
> Your WebWorker API sounds good but does it has an impact on the way eval_code should be designed? Only eval_code_async though.

You either need to parse the code twice or you need to add some extra logic just after `mod = ast.parse(code)` to send a message to the `Executor` to tell it that the syntax check passed. Of course you can factor out everything after `mod = ast.parse(code)` into a separate method. This is how I got to having `_adjust_ast_to_store_result` factored out in the first place.

hoodmane:
> how do you test it?

Use a dummy webloop (stolen from someone in the #245 discussion):
```python
class WebLoop(AbstractEventLoop):
    def call_soon(self, coro, arg=None):
        try:
            x = coro.send(arg)
            x = x.then(partial(self.call_soon, coro))
            x.catch(partial(self.fail,coro))
        except StopIteration as result:
            pass

class WrappedPromise:
    def __init__(self, promise):
        self.promise = promise
    def __await__(self):
        x = yield self.promise
        return x


def wrap_promise(promise):
    return WrappedPromise(promise)
```

hoodmane:
If you like I can add some tests for `eval_code_async` along these lines. 

I've been more focused on fixing `core` since it is full of weird behaviors and disasters waiting to happen (see for example #1033).

casatir:
> it is best assumed to be broken by the "if it isn't tested it is broken" principle

I don't know where does this comes from but it sounds difficult with multiple contributors. Everyone should look at tests to know what is broken? I would prefer the "if it's broken it should be mentioned (and maybe raise an error)" principle.

> For some annoying reason `return await` is a syntax error.

Are you sure about this?

``` python
import asyncio

async def foo() -> str:
    return 'bar'

async def bar() -> str:
    return await foo()

asyncio.run(bar())
```

gives me `'bar'` without any syntax error on Python 3.8.1.

>> there is no value in having tricky code in main places like eval_code only to implement eval_code_async that is not tested and not supported yet.

> There is absolutely value in this. In this case, in order to support async/await there is at least one significant PR and several minor fixes that need to be made to core, another PR to add a web loop, and eval_code_async needs to work. Once we have all the pieces, we will presumably need an extra PR to work out kinks.

Your answer does not explain why it needs to put tricky code in `eval_code`. I maintain.

> If we need to make changes ABCDE in order to achieve some new functionality, but by itself each of those changes introduces unnecessary complexity to the code and adds no user-facing features, then your logic would prevent ever implementing the new functionality.

This is not what I am saying. I think that we should minimize the "pollution" of **existing** code with features that require several PR. This does not ease contributions. Atomicity may be the goal. If this can't be achieved, it should be developed in a separate branch before being merged into master (as async support), but that's a personal point of view.

hoodmane:
> Are you sure about this?

I was sure about it, but now I am sure that it is wrong. =) I guess `return yield x` is a syntax error, and `await` and `yeild` are practically identical constructs so I got confused.

hoodmane:
>> it is best assumed to be broken by the "if it isn't tested it is broken" principle
> I don't know where does this comes from but it sounds difficult with multiple contributors. Everyone should look at tests to know what is broken? I would prefer the "if it's broken it should be mentioned."

Sorry, I think you misunderstood what I was saying. I am absolutely against adding broken or untested code without a warning. We want our code not to broken and "if it isn't tested we assume it is broken" so that means we must test. (Though it's important to remember that testing code does not prevent it from being broken: the converse that "if it is tested it works" is completely wrong.)

> I think that we should minimize the "pollution" of existing code with features that require several PR.

The options are duplication or pollution.

> If this can't be achieved, it should be developed in a separate branch before being merged into master (as async support), but that's a personal point of view.

Maintaining a separate branch is a lot of work, I've been spending a lot of time maintaining pending PRs as it is. I think as long as this project is maintained by unpaid volunteers, there is a limit to what you can expect.

hoodmane:
I think the main problem here is that my previous PR was highly unpolished and it was merged before I addressed these significant and reasonable complaints you have about it.

hoodmane:
> Your answer does not explain why it needs to put tricky code in `eval_code`. I maintain.

I think this is sort of fair, this `eval_code` code is rather simple compared to the webloop and modifications to `JsProxy`. It does feel to me like you are rather selectively policing this one file though.

casatir:
> I am absolutely against adding broken or untested code without a warning.

That's what you did. Maybe I missed the warning.

> The options are duplication or pollution.

You're right, branching is a kind of duplication. I prefer duplication.

> I've been spending a lot of time maintaining pending PRs as it is

Understanding overly complicated code also takes time.

> I think the main problem here is that my previous PR was highly unpolished and it was merged before I addressed these significant and reasonable complaints you have about it.

But you seemed in a hurry: https://github.com/iodide-project/pyodide/pull/876#issuecomment-750024392

> I think this is sort of fair, this eval_code code is rather simple compared to the webloop and modifications to JsProxy.

You mean since Pyodide has tricky parts, we should not polish elsewhere?

> It does feel to me like you are rather selectively policing this one file though.

Replace "policing" by "polishing" and you exactly get the purpose of this PR, yes.

I have a lot of respect for your implication in Pyodide and all I want is to make it better, at my own pace.

src/pyodide-py/pyodide/_base.py
```diff
@@ -100,61 +102,25 @@ def eval_code(code: str, ns: Dict[str, Any]) -> None:
     if len(mod.body) == 0:
         return None
 
-    target_name = "<EXEC-LAST-EXPRESSION>"
-    mod = _adjust_ast_to_store_result(target_name, mod, code)
-    eval(compile(mod, "<exec>", mode="exec"), ns, ns)
-    return ns.pop(target_name)
+    # we extract last expression
+    last_expr = None
+    if isinstance(mod.body[-1], (ast.Expr, ast.Await)) and not quiet_code(code):
```
hoodmane, line 107: You should combine 
`isinstance(mod.body[-1], (ast.Expr, ast.Await))` into `quiet_code`, probably rename `quiet_code` to `should_code_return_value` or something like that. Then we can keep my tests but apply them to `should_code_return_value`.

rth:
I agree that it would be good to simplify that code, and add more tests. But asyncio features are indeed very likely be included once https://github.com/iodide-project/pyodide/pull/880 , https://github.com/iodide-project/pyodide/pull/958, and possibly additional PRs are accepted. We can probably make a META PR with everything to make sure the combination is working as expected, but in the end it will be better to merge them separately as they are now, for clearer history and attribution. So how do we move forward from here?

> If you update this to remove eval_code_async entirely, I'm fine with merging this and then I will make a second PR adding a tested version of eval_code_async.

Could we avoid that code churn and just add more tests in a separate PR? And improve  `eval_code` in this one? 

@casatir don't hesitate to comment on/review PRs in general if you have any concerns.

hoodmane:
> how do we move forward from here?

I could open another PR on top of this branch with my opinion if you like.

casatir:
Here is a proposition to save `_eval_code_async` with warning and error but that should not be broken (provided that one remove exceptions).

> @casatir don't hesitate to comment on/review PRs in general if you have any concerns.

Thanks, time is lacking.

hoodmane:
@rth I'm fine to merge this, I can delete the spurious `NotImplementedError`s and add tests for `eval_code_async` in a subsequent PR.

hoodmane:
I really don't like the `raise NotImplementedError`s, they are a nuisance and also incorrect because they sit above a complete implementation. I think they need to be replaced with warnings e.g.,
```python
import warnings
warnings.warn("asyncio is an experimental and untested feature, things may go wrong!")
```

rth:
>  really don't like the raise NotImplementedError

It doesn't matter much either way, I think. We will fix that once the other pieces of the asyncio support are integrated (hopefully soon).

hoodmane:
Right, well `eval_async` can and should be tested independently of the `EventLoop` and other async support components.

Speaking of which, `runPythonAsync` is kind of occupying the prime space for an entrypoint to `eval_async`...

src/pyodide-py/pyodide/_base.py
```diff
@@ -43,41 +43,137 @@ def open_url(url: str) -> StringIO:
...
 
     Examples
     --------
-    >>> quiet_code('x + 1')
-    False
-    >>> quiet_code('x + 1 ;')
-    True
-    >>> quiet_code('x + 1 # comment ;')
-    False
+    >>> CodeRunner("1+1", {}).run()
```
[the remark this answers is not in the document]
casatir: Great idea! Just pushed it.

src/pyodide-py/pyodide/_base.py
```diff
@@ -49,55 +49,38 @@ class CodeRunner:
...
     --------
-    >>> CodeRunner("1+1", {}).run()
+    >>> CodeRunner().run("1+1")
     2
-    >>> CodeRunner("1+1;", {}).run()
+    >>> CodeRunner().run("1+1;")
     """
 
-    def __init__(self, code: str, ns: Dict[str, Any]):
+    def __init__(self, ns: Dict[str, Any] = {}):
```
rth, line 57: We should better make it None by default, then set to `{}` if it's None, to avoid the issue of https://stackoverflow.com/a/5592432/179127

src/pyodide-py/pyodide/_base.py
```diff
@@ -49,55 +49,38 @@ class CodeRunner:
...
         """
         Constructor.
 
         Parameters
         ----------
-        code
-           the Python code to run.
         ns
            `locals()` or `globals()` context where to execute code.
         """
```
rth, line 65: This docstring should be merged with class docstring. `__init__` doesn't need one.

Also we should add a note that after the first call to `run`, the ns will be modified inplace (as far as I can tell) which will impact subsequent code evaluation.

src/pyodide-py/pyodide/_base.py
```diff
@@ -49,55 +49,38 @@ class CodeRunner:
...
-
-        # we extract last expression
-        last_expr = None
-        if isinstance(mod.body[-1], (ast.Expr, ast.Await)) and not self.quiet():
-            last_expr = ast.Expression(mod.body.pop().value)  # type: ignore
-        self.mod = mod
-        self.last_expr = last_expr
 
-    def quiet(self) -> bool:
+    def quiet(self, code: str) -> bool:
```
rth, line 71: Can probably even be a `@staticmethod` ?

src/pyodide-py/pyodide/_base.py
casatir: Of course! My mistake! Never put a mutable as default argument!

src/pyodide-py/pyodide/_base.py
casatir: In the future, the `quiet` behavior can be impacted by options from the constructor (see #1056).

src/pyodide-py/pyodide/_base.py
casatir: Right. Is `Parameters` allowed in class docstring?

rth: yes, https://numpydoc.readthedocs.io/en/latest/format.html#documenting-classes

src/pyodide-py/pyodide/_base.py
```diff
@@ -43,41 +43,149 @@ def open_url(url: str) -> StringIO:
...
-            tokenize.COMMENT,
-        ):
-            continue
-        return (token.type == tokenize.OP) and (token.string == ";")
 
-    return False
-
-
-def eval_code(code: str, ns: Dict[str, Any]) -> None:
+    def __init__(self, ns: Dict[str, Any] = None):
```
rth, line 67: Type annotations need an `Optional` here, but this can be adressed in a follow up PR. mypy in CI says that everything is OK which is very suspicious.

PR #1044, opened 2021-01-05
Actually generate a sysconfigdata.py

### LINKED ISSUE #1042: Generate sysconfigdata properly
Currently, our sysconfigdata is generated by copying from the host system, which generally produces incorrect values, since our host system is most likely 64-bit but wasm32 is, well, 32-bits.

dalcde:
Previously, we simply copied from the host, which has wildly incorrect values. Fixes #1042


phorward (approved):
This looks good to me, thanks for the change.

Issue #1031, opened 2021-01-03
Problems with conftest.py


hoodmane:
While working on #998 I ran into some significant bugs in `conftest.py` which caused a lot of spurious test failures. These bugs have been making development in `core` hard too on occasion but the effects on `pyodide.js` are much worse.
The main problem is the predicate in `PyodideInited`:
```python
        inited = driver.execute_script(
            "return window.pyodide && window.pyodide.runPython && window.pyodide.pyodide_py"
        )
```
This predicate triggers both false negatives and false positives, and both types cause useless test failures.

I tried in #995 to improve this predicate by attaching it to `languagePluginLoader` but I'm not sure whether it worked. It seems to me that those changes are an improvement but the system is still extremely brittle. Perhaps the problem is that I don't understand the execution model of `<script>` tags in html well enough.

Another problem that seriously harms the utility of the test output is that every test failure produces **enormous** amounts of the same stuff. If somehow we could reduce the amount of junk that gets output, it would make diagnosing failed tests much easier and more pleasant.

I think this is a high priority to fix. I think it's a blocking issue for doing anything like #792 or #998 and a modest but manageable impediment to the work I've been doing on core.

hoodmane:
@dalcde Could you take a stab at this?

dalcde:
Not directly addressing the concern, but I think we can simplify a lot of the current code by defining a `selenium.run_js_async` function that awaits on a promise and returns/throws appropriately. We seem to do this by hand in a lot of different places.

dalcde:
driver.execute_async_script does the job neatly. I'm submitting a PR soon

hoodmane:
I think this resolved by @dalcde's work on conftest.py.

PR #1047, opened 2021-01-05
Experimental new test system

dalcde:
This introduces a new way of writing tests via the decorator @pytest.run_code. This runs the entire decorated function via selenium.run, and passes on any error.

Nothing in here is intended to be final. For example, there ought to be a better way to expose run_code to the tests. Comments welcome.

This also produces more legible errors, e.g. if I change the code to simply assert False, I get
```
__________________________________________________ test_bz2[firefox] __________________________________________________
Message: Error: Traceback (most recent call last):
  File "/lib/python3.8/site-packages/pyodide/_base.py", line 105, in eval_code
    eval(compile(mod, "<exec>", mode="exec"), ns, ns)
  File "<exec>", line 5, in <module>
  File "<exec>", line 3, in test_bz2
AssertionError

------------------------------------------------ Captured stdout setup ------------------------------------------------
Spawning webserver at http://127.0.0.1:49829 (see logs in /tmp/tmp26ej3u19/http-server.log)
---------------------------------------------- Captured stdout teardown -----------------------------------------------
Python exception:
  File "/lib/python3.8/site-packages/pyodide/_base.py", line 105, in eval_code
    eval(compile(mod, "<exec>", mode="exec"), ns, ns)
  File "<exec>", line 5, in <module>
  File "<exec>", line 3, in test_bz2
AssertionError

_hiwire_throw_error@http://127.0.0.1:49829/pyodide.asm.js:8:178853
@http://127.0.0.1:49829/pyodide.asm.wasm:wasm-function[378]:0x11316d
@http://127.0.0.1:49829/pyodide.asm.wasm:wasm-function[376]:0x112e70
pyodide/Module.__pyproxy_apply@http://127.0.0.1:49829/pyodide.asm.js:8:2397731
apply@http://127.0.0.1:49829/pyodide.asm.js:8:182091
languagePluginLoader</Module.runPython@http://127.0.0.1:49829/pyodide.js:333:48
@http://127.0.0.1:49829/test.html:4:34
@http://127.0.0.1:49829/test.html:6:8
```

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -45,7 +45,7 @@ try:
 
 
 except ImportError:
-    pytest = None
+    pytest = None  # type: ignore
 
 
 class JavascriptException(Exception):
@@ -55,7 +55,7 @@ class JavascriptException(Exception):
 
     def __str__(self):
         if self.stack:
-            return self.msg + "\n" + self.stack
+            return self.msg + "\n\n" + self.stack
         else:
             return self.msg
 
--- a/docs/testing.md
+++ b/docs/testing.md
@@ -71,3 +71,36 @@ To lint the code, run:
 ```bash
 make lint
 ```
+
+## Testing framework
+
+### run_in_pyodide
+Many tests simply involve running a chunk of code in pyodide and ensuring it
+doesn't error. In this case, one can use the `run_in_pyodide` decorate from
+`tools/testing.py`, e.g.
+
+```python
+from tools.testing import run_in_pyodide
+
+@run_in_pyodide
+def test_add():
+    assert 1 + 1 == 2
+```
+In this case, the body of the function will automatically be run in pyodide.
+The decorator can also be called with arguments. It has two configuration
+options --- standalone and packages.
+
+Setting `standalone = True` starts a standalone browser session to run the test
+(the session is shared between tests by default). This is useful for testing
+things like package loading.
+
+The `packages` option lists packages to load before running the test. For
+example,
+```python
+from tools.testing import run_in_pyodide
+
+@run_in_pyodide(standalone = True, packages = ["regex"])
+def test_regex():
+    import regex
+    assert regex.search("o", "foo").end() == 2
+```
--- a/docs/testing.md
+++ b/docs/testing.md
@@ -77,10 +77,10 @@ make lint
 ### run_in_pyodide
 Many tests simply involve running a chunk of code in pyodide and ensuring it
 doesn't error. In this case, one can use the `run_in_pyodide` decorate from
-`tools/testing.py`, e.g.
+`pyodide_build/testing.py`, e.g.
 
 ```python
-from tools.testing import run_in_pyodide
+from pyodide_build.testing import run_in_pyodide
 
 @run_in_pyodide
 def test_add():
@@ -97,7 +97,7 @@ things like package loading.
 The `packages` option lists packages to load before running the test. For
 example,
 ```python
-from tools.testing import run_in_pyodide
+from pyodide_build.testing import run_in_pyodide
 
 @run_in_pyodide(standalone = True, packages = ["regex"])
 def test_regex():
--- a/packages/jedi/test_jedi.py
+++ b/packages/jedi/test_jedi.py
@@ -1,11 +1,10 @@
-def test_jedi(selenium_standalone):
-    selenium_standalone.load_package("jedi")
-    result = selenium_standalone.run(
-        """
-        import jedi
-        script = jedi.Script("import json\\njson.lo", path='example.py')
-        completions = script.complete(2, len('json.lo'))
-        [el.name for el in completions]
-        """
-    )
-    assert result == ["load", "loads"]
+from tools.testing import run_in_pyodide
+
+
+@run_in_pyodide(standalone=True, packages=["jedi"])
+def test_jedi():
+    import jedi
+
+    script = jedi.Script("import json\njson.lo", path="example.py")
+    completions = script.complete(2, len("json.lo"))
+    assert [el.name for el in completions] == ["load", "loads"]
--- a/packages/jedi/test_jedi.py
+++ b/packages/jedi/test_jedi.py
@@ -1,4 +1,4 @@
-from tools.testing import run_in_pyodide
+from pyodide_build.testing import run_in_pyodide
 
 
 @run_in_pyodide(standalone=True, packages=["jedi"])
--- a/tools/testing.py
+++ b/tools/testing.py
@@ -1,0 +1,66 @@
+import pytest
+import inspect
+
+
+def run_in_pyodide(_function=None, standalone=False, packages=[]):
+    """
+    This decorator can be called in two ways --- with arguments and without
+    arguments. If it is called without arguments, then the `_function` kwarg
+    catches the function the decorator is applied to. Otherewise, standalone
+    and packages are the actual arguments to the decorator.
+
+    See docs/testing.md for details on how to use this
+    """
+
+    def decorator(f):
+        def inner(selenium):
+            if len(packages) > 0:
+                selenium.load_package(packages)
+            lines, start_line = inspect.getsourcelines(f)
+            # Remove first line, which is the decorator. Then pad with empty lines to fix line number.
+            lines = ["\n"] * start_line + lines[1:]
+            source = "".join(lines)
+
+            err = None
+            try:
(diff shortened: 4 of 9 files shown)


dalcde:
mypy is failing due to the horrible hack to expose run_code

dalcde:
New error log:
```
__________________________________________________ test_bz2[firefox] __________________________________________________
Error running function in pydodie

  File "/src/src/tests/test_bz2.py", line 3, in test_bz2
AssertionError
------------------------------------------------ Captured stdout setup ------------------------------------------------
Spawning webserver at http://127.0.0.1:59799 (see logs in /tmp/tmpfjec_mjy/http-server.log)
---------------------------------------------- Captured stdout teardown -----------------------------------------------
Python exception:
  File "/lib/python3.8/site-packages/pyodide/_base.py", line 105, in eval_code
    eval(compile(mod, "<exec>", mode="exec"), ns, ns)
  File "<exec>", line 12, in <module>
  File "<exec>", line 3, in test_bz2
AssertionError

_hiwire_throw_error@http://127.0.0.1:59799/pyodide.asm.js:8:178853
@http://127.0.0.1:59799/pyodide.asm.wasm:wasm-function[378]:0x11316d
@http://127.0.0.1:59799/pyodide.asm.wasm:wasm-function[376]:0x112e70
pyodide/Module.__pyproxy_apply@http://127.0.0.1:59799/pyodide.asm.js:8:2397731
apply@http://127.0.0.1:59799/pyodide.asm.js:8:182091
languagePluginLoader</Module.runPython@http://127.0.0.1:59799/pyodide.js:333:48
@http://127.0.0.1:59799/test.html:4:34
@http://127.0.0.1:59799/test.html:6:8
```
The line number and file name are now accurate

dalcde:
This partially addresses #1031

hoodmane:
Thanks Dexter! This looks fantastic! 
If tests are written like this, it also would fix a couple of other issues: bad editor syntax highlighting for the important part of the code (the Python part), and using lots of short calls to `selenium.run` and selenium.run_js` makes it hard to run the test on `console.html`. This way I should be able to open the browser terminal, type 
```javascript
pyodide.runPython(`
<ctrl+v>
`)
```
and run the test without having to completely rearrange it.

Maybe we should add a `pyodide_py.run_js` API for the javascript bits. Or we could use my `@jsfunc` decorator.

conftest.py
```diff
@@ -455,6 +458,57 @@ def service_actions():
         httpd.serve_forever()
 
 
+def run_in_pyodide(*args):
```
rth, line 461: not sure `*args` is the best options. `packages=['package1']` might be better. Also we need to be able to switch between `selenium` and  `selenium_standalone` fixtures.

In terms of API maybe something like,
```py
@run_in_pyodide(selenium, packages=['numpy')
def test_...
```

would be more explicit.


Actually I don't understand how the selenium fixture is getting in there, since it only the argument to `wrapped`, not a closure.

conftest.py
```diff
@@ -455,6 +458,57 @@ def service_actions():
...
+    # This can be called as @run_in_pyodide or @run_in_pyodide("first_dep",
+    # "second_dep"). In the former case, run_in_pyodide acts as the decorator.
```
rth, line 463: If it can only be called with the second variant and it allow to simplify that's also fine.

conftest.py
```diff
@@ -455,6 +458,57 @@ def service_actions():
...
+                )
+
+        return wrapped
+
+    if direct_call:
+        return decorator(args[0])
+    else:
+        return decorator
+
+builtins.run_in_pyodide = run_in_pyodide
```
rth, line 510: Assuming there is no closure on the selenium fixture object, we could just put this in some normal python module. Maybe `pyodide_build/testing.py` that way it's a bit less magical?

rth (commented):
It would certainly be great to be able to do this!

conftest.py
dalcde: What is the difference between selenium and selenium_standalone? selenium shares an instance within each file and selenium_standalone is independent? How do we decide which to use?

conftest.py
dalcde: That would work too

conftest.py
rth: Yes, all tests with the `selenium` fixture share the same session and those with `selenium_standalone` don't. Which also means they take much longer because pyodide needs to be reloaded.

Because of the performance cost of `selenium_standalone` the idea is to always `selenium` except when one needs a new sessions. For instance, it's problematic to test `loadPackage` with `selenium` because the package might have already been loaded in some earlier test. That's one of the the reason package tests take so long BTW.

dalcde: Can we reuse the browser instance with a different "tab"?

rth: Yeah, there are probably ways to improve it.

dalcde:
I have squashed and force-pushed because I rewrote a huge chunk of it and moved it to a different file, so the diffs won't be very useful either way.

packages/jedi/test_jedi.py
```diff
@@ -1,11 +1,10 @@
...
-    result = selenium_standalone.run(
-        """
-        import jedi
-        script = jedi.Script("import json\\njson.lo", path='example.py')
-        completions = script.complete(2, len('json.lo'))
-        [el.name for el in completions]
-        """
-    )
-    assert result == ["load", "loads"]
+from tools.testing import run_in_pyodide
```
rth, line 1: Tools is not really a place for python modules. I would rather we put it in `pyodide_build`. I mean it's called build, but in practice it's anything that shouldn't be included in WASM VM (build, testing, etc). So can also rename it to be more general.

Among other things mypy and pytest are not currently run on the `tools` folder.

tools/testing.py
```diff
@@ -0,0 +1,66 @@
...
+
+
+def run_in_pyodide(_function=None, standalone=False, packages=[]):
+    """
+    This decorator can be called in two ways --- with arguments and without
+    arguments. If it is called without arguments, then the `_function` kwarg
+    catches the function the decorator is applied to. Otherewise, standalone
+    and packages are the actual arguments to the decorator.
+
+    See docs/testing.md for details on how to use this
```
rth, line 12: Please document parameters in the docstring still and ideally add type annotations.

rth (commented):
Thanks mostly looks good. Minor comments below. 

Also,
 - I'm fairly surprised that using a decorator defining an internal function with a pytest fixture works. Could we find some reference on whether it's expected to work, or if this is relying on some internal implementation detail of pytest which can be broken at any time.
 - just curious would adding `pytest.mark.parametrize` work (or could be made to work -- though not in in this PR)?

dalcde:
I would expect fixtures to work by querying the parameter names, e.g. via inspect. This should work with dynamically generated functions.

docs/testing.md
```diff
@@ -71,3 +71,36 @@ To lint the code, run:
...
+
+## Testing framework
+
+### run_in_pyodide
+Many tests simply involve running a chunk of code in pyodide and ensuring it
+doesn't error. In this case, one can use the `run_in_pyodide` decorate from
+`tools/testing.py`, e.g.
+
+```python
+from tools.testing import run_in_pyodide
```
rth, line 83:
```suggestion
from pyodide_build.testing import run_in_pyodide
```

docs/testing.md
```diff
@@ -71,3 +71,36 @@ To lint the code, run:
...
+options --- standalone and packages.
+
+Setting `standalone = True` starts a standalone browser session to run the test
+(the session is shared between tests by default). This is useful for testing
+things like package loading.
+
+The `packages` option lists packages to load before running the test. For
+example,
+```python
+from tools.testing import run_in_pyodide
```
rth, line 100:
```suggestion
from pyodide_build.testing import run_in_pyodide
```

docs/testing.md
```diff
@@ -71,3 +71,36 @@ To lint the code, run:
 ```bash
 make lint
 ```
+
+## Testing framework
+
+### run_in_pyodide
+Many tests simply involve running a chunk of code in pyodide and ensuring it
+doesn't error. In this case, one can use the `run_in_pyodide` decorate from
+`tools/testing.py`, e.g.
```
rth, line 80:
```suggestion
`pyodide_build/testing.py`, e.g.
```

rth (approved):
Thanks!

PR #1048, opened 2021-01-05
DOCS: document core

hoodmane:
Added a new file to docs for contributors trying to modify core C code explaining the purpose of the files in core, relevant design decisions, how to do error handling / propagation and reference counting.


hoodmane:
@rth @dalcde I would appreciate your thoughts: everything I just put in is strictly opinion.

docs/core.md
```diff
@@ -0,0 +1,122 @@
...
+## CPython APIs
+
+### Conventions for indicating errors
+The two main ways to indicate errors:
+1. If the function returns a pointer, (usually PyObject*) then to indicate an error set an exception and return NULL.
+2. If the function returns int and the output is nonnegative, to indicate an error set an exception and return NULL.
+
+Certain functions have "successful errors" like ``PyIter_Next`` (successful error is "StopIteration") and ``PyDict_GetItemWithError`` (successful error is "KeyError"). These functions will return NULL without setting an exception to indicate the "successful error" occurred. Check what happened with ``PyErr_Occurred``. Also, functions that return ``int`` for which ``-1`` is a valid return value will return ``-1`` with no error set to indicate that the result is ``-1`` and ``-1`` with an error set if an error did occur. The simplest way to handle this is to always check ``PyErr_Occurred``.
+
+Lastly, the argument parsing functions ``PyArg_ParseTuple``, ``PyArg_Parse``, etc behave weirdly. These return ``true`` on success and return ``false`` and set an error on failure.
```
rth, line 98: "behave weirdly" is not OK wording wise for the documentation.

rth (commented):
Thanks @hoodmane ! Generally looks good, minor comments below. However this more about how to contribute to the core package rather than what is the core package. 
So maybe we could make `docs/contributing` a folder, and then add it as a subsection there `docs/contributing/core.md`?

For the core documentation I would expect more a discussion on the type conversion (we have a section for that already) or on  the other API endpoints defined in core.

hoodmane:
> Thanks @hoodmane ! Generally looks good, minor comments below. However this more about how to contribute to the core package rather than what is the core package.

I think the core package is a private implementation detail: it isn't a functionally related collection of stuff it is "everything that we are currently implementing in C". This was sort of the point of changing the name: `type_conversions` suggests that the code relates to type conversions, whereas `core` is not really descriptive at all which gets closer to the truth. The user-facing docs are subdivided into logical units of things users care about, it is relevant to users where an API is exposed and whether it is API is Python-facing or javascript-facing, but not really which of the three languages we implemented it in.

> So maybe we could make `docs/contributing` a folder, and then add it as a subsection there `docs/contributing/core.md`?

Sounds good to me.

docs/core.md
```diff
@@ -0,0 +1,122 @@
+# Core
+
+## Structure of functions
+
+It takes special care to correctly and cleanly handle both reference counting and exceptions.
+To do this, we divide any function that produces more than a couple of owned ``PyObject*``s or ``JsRef``s into several "segments".
```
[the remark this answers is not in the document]
hoodmane: I rewrote and expanded significantly on my reasoning behind these decisions. When you get a chance, I'd appreciate it if you could look over it again.

hoodmane:
Maybe it would be fun to link this picture for highwire:
https://en.wiktionary.org/wiki/highwire_walker#/media/File:Wonderland_Walker_5.jpg
I think if we make the picture a link to the `wiktionary/media` file above it satisfies [the license](https://creativecommons.or/licenses/by/2.0/). That's what [the wiktionary page](https://en.wiktionary.org/wiki/highwire_walker) does.

docs/core.md
hoodmane: Changed to "are edge cases"

hoodmane:
@rth Can we merge this now?

rth (commented):
Thanks for this writ up. A few minor comments otherwise LGTM.

hoodmane:
Having deleted `pyimport` and `jsimport`, I'm starting to feel like `type_conversions` was actually a pretty good name for `core`, since `runpython` is the only file that is not about type conversions and it doesn't really do much anymore.

hoodmane:
Okay, I think I addressed your comments.

rth:
> I'm starting to feel like type_conversions was actually a pretty good name for core, since runpython is the only file that is not about type conversions and it doesn't really do much anymore.

We can still come back on that decision, ideally before the next release.

Issue #1049, opened 2021-01-05
Bug: pyproxies need to be refcounted


hoodmane:
Thanks to @mgreminger (see his original report in #1006).
```javascript
pyodide.runPython('a=lambda x: 2*x');
ref1 = pyodide.runPython('a');
ref2 = pyodide.runPython('a');
console.log('first try:', ref2(10)); // prints 20
ref1.destroy();
console.log('second try:', ref2(20)); // results in Error: Object has already been destroyed
```

mgreminger:
Thanks, @hoodmane. After thinking about this more, there's potentially another issue going on in addition to reference counting. I realized that `a` is still in scope on the Python side in my example so `a` will have a ref count of at least 1 (ideally 2 after ref1 is destroyed). So the second potential issue is what `.destroy()` is doing with objects that still have a ref count > 0 on the Python side. My understanding was that `.destroy()` was just decrementing the ref count on the Python side to allow the Python GC to do it's work, but there's something more going on here. To truly test what I was intending, I would of had to call `del a` from the Python side before trying to access `ref2` for the second time.

hoodmane:
@mgreminger I think you are missing the point here. Yes `destroy` does `Py_DECREF` the wrapped `PyObject*`. But the more important thing it does is disable the pyproxy. The pyproxy is disabled regardless of what the new Python refcount is after `Py_DECREF`. If we just stopped caching the pyproxies and generated a new pyproxy every time, this bug would be resolved. I don't even think we would leak much more memory. The caching of the pyproxies is wrong, because it behaves as if the pyproxy were immutable but destroy mutates it.

As you can see from the fact that the error occurs without `del a` or invoking the Python garbage collector, python GC is unrelated to this bug.

mgreminger:
@hoodmane, thanks for the clarification. Yes, totally agree that there are two different issues here.

PR #1050, opened 2021-01-05
Combine pyproxy_new and pyproxy_use into pyproxy_new

hoodmane:
The difference between pyproxy_new and py_proxy_use is a pyproxy implementation detail. Combined them into get_pyproxy.


hoodmane:
No effort yet to address #1049.

rth:
I'm not sure this factorization is worth it. You do remove `pyproxy_new` and `pyproxy_use` from the header, but then add a new function, which also doesn't follow the implied `pyproxy_<action>` naming convention. Also `pyproxy_get` already exist and can be confused with `get_pyproxy`. 

Maybe meging `pyproxy_new` and `pyproxy_use` into a single function (possibly still `pyproxy_use`) would be better? They have 7 LoC among them.

Or keeping the code as it is is also fine IMO.

hoodmane:
> I'm not sure this factorization is worth it

I think these two functions need to be merged. Consider the following code from the current `py2js`:
```C
    ret = pyproxy_use(x);
    return get_pyproxy(x);
    if (ret != NULL) {
      return ret;
    }

    // Reference counter is increased only once when a PyProxy is created.
    Py_INCREF(x);
    return pyproxy_new(x);
```
This is what you are supposed to do every time you need a `PyProxy`. This pollutes the call site with an implementation detail from `pyproxy`.

hoodmane:
> Maybe merging `pyproxy_new` and `pyproxy_use` into a single function (possibly still `pyproxy_use`) would be better? They have 7 LoC among them.

That sounds like the right choice to me.

hoodmane:
We currently get a `JsProxy` by calling `JsProxy_cnew`. Why don't we call the function `pyproxy_new`? Also if it's going to be `JsProxy` shouldn't it be `PyProxy`?

rth:
>  Also if it's going to be JsProxy shouldn't it be PyProxy

`pyproxy_new` sounds good. If you change it to PyProxy_new you would need to also change all other functions for consistency. Maybe let's avoid it for now, to much code churn also make using git blame hader..

hoodmane:
If I get around to doing a `pyproxy` refactor I can rename it then.

hoodmane:
@dalcde @rth Still a lot of improvement to be had in conftest:
See [this test failure](https://app.circleci.com/pipelines/github/iodide-project/pyodide/1165/workflows/86ec6230-b4e8-4bc4-b1a5-0b141a22f71e/jobs/13928): no information is conveyed about the nature of the error. If I go to `console.html` locally I see the following perfectly diagnostic message in the browser console.
```javascript
pyodide.asm.js:8 exception thrown: ReferenceError: x is not defined,ReferenceError: x is not defined
    at _pyproxy_new (http://localhost:8002/pyodide.asm.js:8:188778)
    at __python2js (http://localhost:8002/pyodide.asm.wasm:wasm-function[387]:0x108a51)
    at _python2js (http://localhost:8002/pyodide.asm.wasm:wasm-function[389]:0x108e72)
    at _runpython_init (http://localhost:8002/pyodide.asm.wasm:wasm-function[410]:0x109822)
    at _main (http://localhost:8002/pyodide.asm.wasm:wasm-function[321]:0x10690a)
    at Object.Module._main (http://localhost:8002/pyodide.asm.js:8:2822531)
    at callMain (http://localhost:8002/pyodide.asm.js:8:3472842)
    at doRun (http://localhost:8002/pyodide.asm.js:8:3473421)
    at run (http://localhost:8002/pyodide.asm.js:8:3473593)
    at runCaller (http://localhost:8002/pyodide.asm.js:8:3472493)
```

PR #1051, opened 2021-01-05
EM_JS wrappers to convert calls to use Python error conventions

### LINKED ISSUE #1005: Detecting crashes of the Pyodide WASM environment
Copied from:
https://github.com/iodide-project/iodide/issues/994

> It would be very useful if Iodide could detect when Pyodide WASM environment crashes (e.g. after a segfault), and warn the user.
> 
> For instance, running the simple code snippet from iodide-project/pyodide#200 in the demo notebook https://iodide.io/pyodide-demo/python.html produces a segfault which crashes Pyodide, meaning that any subsequent command will also fail.
> 
> This situation is very confusing for users, because Iodide still works in this scenario. So the user gets some error in one notebook cell, and then thinking that maybe something was wrong in the provided code, they would try to change the input code, but irrespective of the input the cell would still fail. More frustrating, even other cells that used to work now will also fail. The only solution is to restart the page.
> 
> Even if though segfaults shouldn't happen, it's still possible to run into one, particularly when people build some custom packages. I'm not sure if running out of memory can also produce a somewhat similar outcome.

hoodmane:
The goal of this PR is to make error handling uniform between `EM_JS`-defined functions and Python APIs. Most Python APIs return `NULL` and use `PyErr_SetString` or one of its friends. If we call javascript code from C and the javascript code throws an error that is not caught, the browser stack will unwind past the C code until it hits a javascript stack frame again, because C has no "try/catch". Thus, if we want to be able handle errors generated in the javascript code we better catch javascript errors generated and set C errors. 

These wrapper can also be modified to add "use strict"; to every method at some point, but then we'll have to fix the extra errors that result.

Currently I rethrow the error, dummying out the javascript error to C error conversion. This is because right now the C code does not check whether `EM_JS` functions return errors (nor whether Python APIs return errors for the most part).

After this PR I am going to go through systematically and add error checking everywhere. Then I can enable this code.

This will eventually lead to a fix for #1005 and allow for cooperative stack unwinding a few PRs down the road.

### DIFF
--- a/src/core/error_handling.h
+++ b/src/core/error_handling.h
@@ -28,11 +28,11 @@ error_handling_init();
 #define EM_JS_REF(ret, func_name, args, body...)                               \
   EM_JS(ret, func_name, args, {                                                \
     /* "use strict";  TODO: enable this. */                                    \
-    try    /* intentionally no braces. */                                      \
+    try    /* intentionally no braces, body already has them */                \
       body /* <== body of func */                                              \
     catch (e) {                                                                \
-        throw e;                                                               \
         /* Dummied out until calling code is ready to catch these errors */    \
+        throw e;                                                               \
         Module.handle_js_error(e);                                             \
         return -1;                                                             \
     }                                                                          \
@@ -41,11 +41,11 @@ error_handling_init();
 #define EM_JS_NUM(ret, func_name, args, body...)                               \
   EM_JS(ret, func_name, args, {                                                \
     /* "use strict";  TODO: enable this. */                                    \
-    try    /* intentionally no braces. */                                      \
+    try    /* intentionally no braces, body already has them */                \
       body /* <== body of func */                                              \
     catch (e) {                                                                \
-        throw e;                                                               \
         /* Dummied out until calling code is ready to catch these errors */    \
+        throw e;                                                               \
         Module.handle_js_error(e);                                             \
         return -1;                                                             \
     }                                                                          \
--- a/src/core/error_handling.h
+++ b/src/core/error_handling.h
@@ -12,37 +12,34 @@ typedef int errcode;
 
 // Hiwire wants to import us for errcode, so import hiwire after typedef.
 #include "hiwire.h"
-void
-PyodideErr_SetJsError(JsRef err);
+#include <emscripten.h>
 
+// clang-format off
 #define EM_JS_REF(ret, func_name, args, body...)                               \
   EM_JS(ret, func_name, args, {                                                \
     /* "use strict";  TODO: enable this. */                                    \
-    try {                                                                      \
-      body                                                                     \
-    } catch (e) {                                                              \
-      /* Dummied out until calling code is ready to catch these errors */      \
-      throw e;                                                                 \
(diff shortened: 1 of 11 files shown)


hoodmane:
@rth @dalcde Review requested. =) I think this is a really important PR, it's going to make a huge difference to the stability of Pyodide, the quality of the error messages we can show to users, etc.

The main weakness is that C macros are not actually composable without changing their behavior, and I think there's no way to do `EM_JS_REF` / `EM_JS_NUM` without incurring macro expansion in the body of the javascript function. The only damage this causes is to break `true` and `false` in javascript.
(C macro programming has me missing latex programming which has much stronger expansion control with `\expandafter` and `\edef` and `\noexpand`, `\romannumeral`, etc. C macros give you very poor control.)

Anyways, my fix for this is to redefine `true` to be `(!!1)` and `false` to be `(!!0)` which are correct in both C and javascript.

src/core/error_handling.h
```diff
@@ -0,0 +1,55 @@
...
+// In types.h we provide nonstandard definitions:
+// false ==> (!!0)
+// true ==> (!!1)
+// These work as expected in both C and javascript.
+
+// clang-format off
+#define EM_JS_REF(ret, func_name, args, body...)                               \
+  EM_JS(ret, func_name, args, {                                                \
+    /* "use strict";  TODO: enable this. */                                    \
+    try    /* intentionally no braces. */                                      \
```
rth, line 31: Because the body already has braces, right? It might be good to write it more explicitly.

src/core/error_handling.h
```diff
@@ -0,0 +1,55 @@
...
+      body /* <== body of func */                                              \
+    catch (e) {                                                                \
+        throw e;                                                               \
```
rth, line 34: Wait so why do you need to throw the error? I'm probably missing something.

rth (commented):
It's a nice improvement, thanks @hoodmane !

> C macro programming has me missing latex programming

Let's not add LaTeX to pyodide ^^. There is https://github.com/vadimkantorov/busytex if you really do.

hoodmane:
> Let's not add LaTeX to pyodide

Oh god no. LaTeX barely manages to run in the native OS.

src/core/error_handling.h
hoodmane: That temporarily disables the wrapper functionality, because it breaks tests. I will fix the test cases and remove the guard in a second PR. Unless you want me to do both at once, but I figured this was already a significant diff.

hoodmane: I guess I should move the comment up just a little bit.

src/core/error_handling.h
hoodmane: Exactly, will do.

rth (approved):
LGTM, once the comments are updated.

> I figured this was already a significant diff.

Yes, mostly because clang-format can produce a 10 line diff for a 3 char change, which is not great.

hoodmane:
Pretty exciting, after this merges all of the little issues are straightened out and we can finally implement real error handling.

hoodmane:
Oh one more mistake.

PR #1055, opened 2021-01-06
Refactor conftest.py

dalcde:
This eliminates a lot of duplicate code using selenium's driver.execute_async_script.

Error handling is manual, by returning a discriminator. The "proper" solution should be to use selenium's JavascriptException type, but it doesn't seem to support Chrome, and we get a generic WebDriverException that is less useful.

This ends up simpliflying the error output, but there is still work to be done on that front.

Partially addresses #1031

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -48,23 +48,6 @@ except ImportError:
     pytest = None
 
 
-def package_loaded(driver):
-    inited = driver.execute_script("return window.done;")
-    return bool(inited)
-
-
-def _display_driver_logs(browser, driver):
-    if browser == "chrome":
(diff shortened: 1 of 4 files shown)


rth (approved):
Thanks! This looks indeed like a more elegant/simpler solution than what we had before, and it's a good idea to use `driver.execute_async_script`. I have not tested what error output it produced on error, but I guess we'll see soon.

hoodmane:
> I have not tested what error output it produced on error, but I guess we'll see soon.

HOLY SHIT IT IS SO MUCH BETTER!!

PR #1056, opened 2021-01-06
Customizable eval_code behavior

casatir:
This is built on top of #1041.

#876 brings really interesting features. Some of them were not accepted, like returning last assignment value (by my fault), others does not reflect the CPython interpret behavior (like silencing by trailing semicolon).

I am totally agree with the current behavior but what if the user want something else?

This PR adds three customization to `eval_code` (and `eval_code_async`) without modifying the current behavior:

 - compilation filename (default is `"<exec>"`)
 - trailing semicolon behavior (quiet or not to switch between CPython and IPython behavior)
 - returned value:
   - last expression,
   - last expression and (named) assignment
   - nothing

WDYT?

I have added some tests.

### DIFF
--- a/src/pyodide-py/pyodide/_base.py
+++ b/src/pyodide-py/pyodide/_base.py
@@ -52,6 +52,14 @@ class CodeRunner:
     ns
         `locals()` or `globals()` context where to execute code.
         This namespace is updated by the subsequent calls to `run()`.
+    mode
+        'last_expr' , 'last_expr_or_assign' or 'none',
+        specifying what should be evaluated and what should be executed.
+        'last_expr' will return the last expression
+        'last_expr_or_assign' will return the last expression
+        or the last (named) assignment.
+        'none' will always return `None`.
+        Other values will be interpreted as 'none'.
     filename:
         file from which the code was read.
 
@@ -66,9 +74,15 @@ class CodeRunner:
     6
     """
 
-    def __init__(self, ns: Dict[str, Any] = None, filename: str = "<exec>"):
+    def __init__(
+        self,
+        ns: Dict[str, Any] = None,
+        mode: str = "last_expr",
+        filename: str = "<exec>",
+    ):
         self.ns = ns if ns is not None else {}
         self.filename = filename
+        self.mode = mode
 
     def quiet(self, code: str) -> bool:
         """
@@ -107,6 +121,35 @@ class CodeRunner:
 
         return False
 
+    def _last_assign_to_expr(self, mod: ast.Module):
+        """
+        Implementation of 'last_expr_or_assign' mode.
+        It modify the supplyied AST module so that the last
+        statement's value can be returned in 'last_expr' mode.
+
+        Largely inspired from IPython:
+        https://github.com/ipython/ipython/blob/3587f5bb6c8570e7bbb06cf5f7e3bc9b9467355a/IPython/core/interactiveshell.py#L3229
+        """
+        assign_nodes = (ast.AugAssign, ast.AnnAssign, ast.Assign)
+        single_targets_nodes = (ast.AugAssign, ast.AnnAssign)
+
+        last_node = mod.body[-1]
+
+        if not isinstance(last_node, assign_nodes):
+            return
+
+        target: Any
+        if isinstance(last_node, ast.Assign) and len(last_node.targets) == 1:
+            target = last_node.targets[0]
+        elif isinstance(last_node, single_targets_nodes):
+            target = last_node.target
+        else:
+            target = None
+        if isinstance(target, ast.Name):
+            last_node = ast.Expr(ast.Name(target.id, ast.Load()))
+            mod.body.append(last_node)
+            ast.fix_missing_locations(mod)
+
     def _split_and_compile(self, code: str, flags: int = 0x0) -> Tuple[Any, Any]:
         """
         Split code in two parts, everything but last expression and
@@ -126,9 +169,18 @@ class CodeRunner:
         if not mod.body:
             return None, None
 
+        mode = self.mode
+        if mode == "last_expr_or_assign":
+            self._last_assign_to_expr(mod)
+            mode = "last_expr"
+
         # we extract last expression
         last_expr = None
-        if isinstance(mod.body[-1], (ast.Expr, ast.Await)) and not self.quiet(code):
+        if (
+            mode == "last_expr"
+            and isinstance(mod.body[-1], (ast.Expr, ast.Await))
+            and not self.quiet(code)
+        ):
             last_expr = ast.Expression(mod.body.pop().value)  # type: ignore
 
         # we compile
@@ -187,7 +239,9 @@ class CodeRunner:
             return res
 
 
-def eval_code(code: str, ns: Dict[str, Any], filename: str = "<exec>") -> Any:
+def eval_code(
+    code: str, ns: Dict[str, Any], mode: str = "last_expr", filename: str = "<exec>"
+) -> Any:
     """Runs a code string.
 
     Parameters
@@ -196,6 +250,14 @@ def eval_code(code: str, ns: Dict[str, Any], filename: str = "<exec>") -> Any:
        the Python code to run.
     ns
        `locals()` or `globals()` context where to execute code.
+    mode
+       'last_expr' , 'last_expr_or_assign' or 'none',
+       specifying what should be evaluated and what should be executed.
+       'last_expr' will return the last expression
+       'last_expr_or_assign' will return the last expression
+       or the last (named) assignment.
+       'none' will always return `None`.
+           Other values will be interpreted as 'none'.
     filename:
        file from which the code was read.
 
@@ -205,11 +267,11 @@ def eval_code(code: str, ns: Dict[str, Any], filename: str = "<exec>") -> Any:
     If the last statement is an expression, return the
     result of the expression.
     """
-    return CodeRunner(ns, filename).run(code)
+    return CodeRunner(ns, mode, filename).run(code)
 
 
 async def _eval_code_async(
-    code: str, ns: Dict[str, Any], filename: str = "<exec>"
+    code: str, ns: Dict[str, Any], mode: str = "last_expr", filename: str = "<exec>"
 ) -> Any:
     """ //!\\ WARNING //!\\
     This is not working yet. For use once we add an EventLoop.
@@ -221,7 +283,7 @@ async def _eval_code_async(
       - add tests
     """
     raise NotImplementedError("Async is not yet supported in Pyodide.")
-    return await CodeRunner(ns, filename).run_async(code)
+    return await CodeRunner(ns, mode, filename).run_async(code)
 
 
 def find_imports(code: str) -> List[str]:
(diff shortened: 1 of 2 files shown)


hoodmane:
Looks like good work @casatir! Of course since #941 merged, anyone can sub out their own version of `eval_code` and since #923 merged it's possible to write your own variant of `runPython` using your own variant of `eval_code` without changing the system version. With this in mind, I don't think it is necessary to implement every possible behavior. However, I like the look of this.

I think we really need at this point to make a simple version of `runPython` to use internally. I think by default the system should just do a direct `exec` without any of this fancy stuff, and to get the result it should look up a variable in the namespace directly. This should also make developing `eval_code` changes easier because mistakes won't cause fatal errors.

hoodmane:
Do you know of any circumstance where an `Assign` node has `targets` field with a list of length greater than 1? One would guess that:
```python
a, b = 1, 2
```
would lead to that, but in this case `targets` is a single tuple.

src/pyodide-py/pyodide/_base.py
```diff
@@ -43,41 +43,206 @@ def open_url(url: str) -> StringIO:
...
+        if isinstance(last_node, ast.Assign) and len(last_node.targets) == 1:
+            target = last_node.targets[0]
+        elif isinstance(last_node, single_targets_nodes):
+            target = last_node.target
+        else:
+            target = None
+        if isinstance(target, ast.Name):
+            last_node = ast.Expr(ast.Name(target.id, ast.Load()))
+            mod.body.append(last_node)
+            ast.fix_missing_locations(mod)
```
hoodmane, line 143: What about also supporting destructuring assignments, item assignment, etc? See comment.

hoodmane:
On L140 you could do:
```python
        # If node is already an Assign, deep copy the lvalue of the Assign and store that structure
        # into our result.
        # This has the consequence that "[a, b] = (1,2)" returns "[1, 2]", while "a, b = (1,2)" returns "(1,2)".
        # This could be mildly unexpected behavior but it seems entirely harmless.
        # Also in case of l[5] = 7 evaluates l[5] at the end. Python lvalues can be pretty complicated.
        import copy
        expr = copy.deepcopy(target)
        # The deep copied expression was an lvalue but we are trying to use it as an rvalue.
        # Need to replace all the "Store" lvalue context markers with "Load" rvalue context markers.
        for x in ast.walk(expr):
            if hasattr(x, "ctx"):
                x.ctx = ast.Load()
        mod.body.append(ast.Expr(expr))
```

hoodmane:
Also, I still disagree with the `NotImplementedError`s. Not implemented means not implemented. It is implemented, you just haven't tested it. Could you instead just emit a warning "this is an experimental feature"? This wouldn't prevent testing / experimentation.

casatir:
Thanks @hoodmane for the review.

I have updated with the new version of #1041.

> Of course since #941 merged, anyone can sub out their own version of eval_code and since #923 merged it's possible to write your own variant of runPython using your own variant of eval_code without changing the system version. With this in mind, I don't think it is necessary to implement every possible behavior.

You're right but that hold only for advanced users to get the behavior implemented in this PR no?

> I think we really need at this point to make a simple version of runPython to use internally.

There is no obligation to trace this back to `runPython`.

> to get the result it should look up a variable in the namespace directly

Can't get this point.

> This should also make developing eval_code changes easier because mistakes won't cause fatal errors.

I agree that this weighs down `eval_code` a bit. This is a proposition. If this is useless or not appropriate, I am fine with it.

> Do you know of any circumstance where an Assign node has targets field with a list of length greater than 1?

Just do `a = b = 1` and you will get an `Assign` with a `targets` of length 2.

> What about also supporting destructuring assignments, item assignment, etc? See comment.

Seems too complicated to me at this point (maybe I'm lazy...). Not easy to think that it can imply deep copy for the user though.

hoodmane:
> You're right but that hold only for advanced users to get the behavior implemented in this PR no?

Agreed, I think what you've done here is good.

> I think we really need at this point to make a simple version of runPython to use internally. I think by default the system should just do a direct exec without any of this fancy stuff, and to get the result it should look up a variable in the namespace directly. This should also make developing eval_code changes easier because mistakes won't cause fatal errors.

This paragraph ^^ was not intended as action items for this PR or anything, just speculation about the future. Sorry, I'll try to better label feedback vs ramble in the future.

hoodmane:
`Just do a = b = 1 and you will get an Assign with a targets of length 2.`

I didn't even realize that was syntactically valid!

hoodmane:
> Not easy to think that it can imply deep copy for the user though.

More complicated stuff already happens inside the parser when it creates the AST. The depth of the deep copy is usually very shallow (like depth 2 or something), it would only be a big copy if they did something like

[[[[[[[[[x]]]]]]]]] = [[[[[[[[[1]]]]]]]]]

casatir:
> I didn't even realize that was syntactically valid! In that case can't you just take target 0 and return 1?

Not sure it's the only case! But we can try.

casatir:
Just rebased to `master` since #1041 is merged.

rth (commented):
Thanks @casatir ! Overall this LGTM, it's indeed nice to have those features configurable. 

For future PRs note that in [docstrings with nympy style](https://numpydoc.readthedocs.io/en/latest/format.html#docstring-standard), there is a 1 line summary, then an empty line then the rest of the docstring. The 1 line summary is then what's used in summary tables by sphinx https://pyodide.readthedocs.io/en/latest/api_reference.html#python-api . It matters less for private methods.

I have not reviewed `_last_assign_to_expr` function it detail. @hoodmane could you please approve the PR if you are OK with it.

src/pyodide-py/pyodide/_base.py
```diff
@@ -124,9 +180,18 @@ def _split_and_compile(self, code: str, flags: int = 0x0) -> Tuple[Any, Any]:
         if not mod.body:
             return None, None
 
+        mode = self.mode
+        if mode == "last_expr_or_assign":
+            self._last_assign_to_expr(mod)
+            mode = "last_expr"
```
hoodmane, line 186:
```suggestion
        if self.mode == "last_expr_or_assign":
            # If the last statement is an assignment, add an extra expression
            # to the end with just the L-value so that we can handle it with 
            # the last_expr code.
            self._last_assign_to_expr(mod)
```

src/pyodide-py/pyodide/_base.py
```diff
@@ -124,9 +180,18 @@ def _split_and_compile(self, code: str, flags: int = 0x0) -> Tuple[Any, Any]:
...
+
         # we extract last expression
         last_expr = None
-        if isinstance(mod.body[-1], (ast.Expr, ast.Await)) and not self.quiet(code):
+        if (
+            mode == "last_expr"
```
hoodmane, line 191:
```suggestion
            self.mode == "last_expr" or self.mode == "last_expr_or_assign"
```

src/pyodide-py/pyodide/_base.py
```diff
@@ -105,6 +132,35 @@ def quiet(self, code: str) -> bool:
...
+        """
+        Implementation of 'last_expr_or_assign' mode.
+        It modify the supplyied AST module so that the last
+        statement's value can be returned in 'last_expr' mode.
+
+        Largely inspired from IPython:
+        https://github.com/ipython/ipython/blob/3587f5bb6c8570e7bbb06cf5f7e3bc9b9467355a/IPython/core/interactiveshell.py#L3229
+        """
+        assign_nodes = (ast.AugAssign, ast.AnnAssign, ast.Assign)
+        single_targets_nodes = (ast.AugAssign, ast.AnnAssign)
```
hoodmane, line 145:
```suggestion
```

src/pyodide-py/pyodide/_base.py
```diff
@@ -105,6 +132,35 @@ def quiet(self, code: str) -> bool:
...
+
+        last_node = mod.body[-1]
+
+        if not isinstance(last_node, assign_nodes):
+            return
```
hoodmane, line 150:
```suggestion
```

src/pyodide-py/pyodide/_base.py
```diff
@@ -105,6 +132,35 @@ def quiet(self, code: str) -> bool:
...
+
+        target: Any
+        if isinstance(last_node, ast.Assign):
+            target = last_node.targets[0]
```
hoodmane, line 154:
```suggestion
        if isinstance(last_node, ast.Assign):
            # In this case there can be multiple targets as in `a = b = 1`.
            # We just take the first one.
            target = last_node.targets[0]
```

src/pyodide-py/pyodide/_base.py
```diff
@@ -105,6 +132,35 @@ def quiet(self, code: str) -> bool:
...
+        elif isinstance(last_node, single_targets_nodes):
+            target = last_node.target
```
hoodmane, line 156:
```suggestion
        elif isinstance(last_node, (ast.AugAssign, ast.AnnAssign)):
            target = last_node.target
```

src/pyodide-py/pyodide/_base.py
```diff
@@ -105,6 +132,35 @@ def quiet(self, code: str) -> bool:
...
+        else:
+            target = None
+        if isinstance(target, ast.Name):
+            last_node = ast.Expr(ast.Name(target.id, ast.Load()))
+            mod.body.append(last_node)
+            ast.fix_missing_locations(mod)
```
hoodmane, line 162:
```suggestion
            # Update the line numbers shown in error messages.
            ast.fix_missing_locations(mod)
```

hoodmane (approved):
It looks good to me. I suggested adding a few comments and making the logic clearer in a couple of places.

hoodmane:
Could you add some tests with multiple nonempty lines like
```
x=2
x
```
and
```
def f(x):
   return x*x+1
[f(x) for x in range(10)]
```
These don't work in console.html going back to version 0.15.0, which is a separate issue.

src/pyodide-py/pyodide/_base.py
```diff
@@ -105,6 +132,35 @@ def quiet(self, code: str) -> bool:
...
+        if not isinstance(last_node, assign_nodes):
+            return
+
+        target: Any
+        if isinstance(last_node, ast.Assign):
+            target = last_node.targets[0]
+        elif isinstance(last_node, single_targets_nodes):
+            target = last_node.target
+        else:
+            target = None
```
hoodmane, line 158:
```suggestion
            return
```

hoodmane:
Could you add some tests with multiple nonempty lines like
```
x=2
x
```
and
```
def f(x):
   return x*x+1
[f(x) for x in range(10)]
```
These don't work in console.html going back to version 0.15.0, which is a separate issue.

hoodmane:
Some more thoughts for the future (not necessarily for this PR):
1. `ns` needs to be replaced with separate `locals` and `globals` arguments. Then in `pyodide.js` we can use `eval_code` with a throwaway temporary dictionary to avoid polluting global namespace.
2. runPython could always use the same instance of `CodeRunner`

casatir:
Thanks @rth and @hoodmane.

> For future PRs note that in docstrings with nympy style, there is a 1 line summary, then an empty line then the rest of the docstring. The 1 line summary is then what's used in summary tables by sphinx https://pyodide.readthedocs.io/en/latest/api_reference.html#python-api . It matters less for private methods.

Thanks, I should improve my skills on this. 

I added the changes from @hoodmane review.

hoodmane:
Looks good to me, I think this can merge.

rth:
Thanks @casatir and @hoodmane !

PR #595, opened 2020-01-06
Show matplotlib figures in documents other than iodide notebooks

basvandertol:
The wasm_backend for matplotlib creates a div element by a iodide.output.element('div') call,
if that fails a plain div is created by a document.createElement('div') call. In the latter case the newly created div is never appended to the document and remains therefore invisible. 

This PR fixes that by adding a call to document.body.appendChild(div) after the creation of the div.

PR #1057, opened 2021-01-06
Remove `from js import iodide` from matplotlib wasm_backend

hoodmane:


dalcde:
Related: #595

rth:
Yes, assuming some iodide development continues one day, we should make sure it's able to run with the public API without requiring us to vendor iodide specific code.

PR #1060, opened 2021-01-07
Fix build error for zlib

mgreminger:
#1027 introduced build errors on my system. The \`cwd\` call in the script portion of zlib's meta.yaml is an undefined call (cwd is not a standard bash command) and results in the next line of the script attempting to install zlib to the root file system. This probably doesn't cause a CI error because of differences in the permissions on the CI system. On my system, this results in write permission errors. I replaced the \`cwd\` with a . since that's the intent. I also updated the bash calls in buildpkg.py to use the -e option in addition to the -c option. The -e option forces bash to exit on the first line that has an error. Without the -e, we only see the return status of the last command regardless of whether there were errors on lines before that. Incidentally, the -e wouldn't have caught the problem with \`cwd\` since bash does not bubble up back-tick command errors by default. With these changes, pyodide and all packages build cleanly on my system.

### DIFF
--- a/packages/zlib/meta.yaml
+++ b/packages/zlib/meta.yaml
@@ -9,5 +9,5 @@ source:
 build:
   library: true
   script: |
-     CFLAGS="-fPIC" emconfigure ./configure --prefix=.
+     CFLAGS="-fPIC" emconfigure ./configure --prefix=$PWD
      emmake make install -j ${PYODIDE_JOBS:-3}
--- a/packages/zlib/meta.yaml
+++ b/packages/zlib/meta.yaml
@@ -9,5 +9,5 @@ source:
 build:
   library: true
   script: |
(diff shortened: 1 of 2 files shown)


dalcde:
Would `pwd` work better? I think it is better to use an absolute path instead of a relative one.

mgreminger:
. works fine but if absolute path is desired, $PWD would be better than \`pwd\` since we shouldn't encourage backtick commands because any errors they generate are ignored by bash.

dalcde:
Sounds good. I'm evidently not a bash guru

mgreminger:
Updated to use $PWD instead of .

mgreminger:
Oops, accidentally closed the pull request. I think it's reopened now.

dalcde:
Thanks!

PR #1061, opened 2021-01-07
More careful impl of hiwire_bool, added test cases

hoodmane:
For a jsproxy, `bool(val)` is defined as `(val && (val.length === undefined || val.length))`. This seems to encode the assumption that anything with a length field is some sort of container. However, there are many things that are not a container that have a length field. [`window.length` is the number of <frame> elements in the window](https://developer.mozilla.org/en-US/docs/Web/API/Window/length), typically zero, so it considers `window` to be falsey. Functions also have a length field which counts their arguments, zero argument javascript functions are currently falsey.

Also, this implementation doesn't know about `Set` or `Map`, empty javascript sets and maps are truthy. I replaced it with a smarter implementation.

I also replaced the `==` and `!=` in `hiwire_eq` resp `hiwire_neq` with `===` and `!==`. On the code paths that use `hiwire_eq` and `hiwire_neq`, these do the same thing, but `===` and `!==` are less confusing.

There is one more `!=` operator:
https://github.com/iodide-project/pyodide/blob/1cb32c304d3901a57e95226e6b528501047cff01/src/core/hiwire.c#L428

However, the one place that `!=` operator is used is in `JsProxy_IterNext`:
```C
  JsRef iddone = hiwire_get_member_string(idresult, "done");
  int done = hiwire_nonzero(iddone);
  hiwire_decref(iddone);
```
Here `iddone` is either `true` or `false` and `false !== 0` is `true` but `false != 0` is `false`. So if we corrected `hiwire_nonzero` to be sensible it would break `JsProxy_IterNext`.

### DIFF
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -35,6 +35,7 @@
 
 ### Fixed
 - getattr and dir on JsProxy now report consistent results and include all names defined on the Python dictionary backing JsProxy. [#1017](https://github.com/iodide-project/pyodide/pull/1017)
+- `JsProxy.__bool__` now produces better results: it used to be `bool(window)` and `bool(zero-arg-callback)` were `False`, and `bool(empty_js_set)` and `bool(empty_js_map)` were `True`. All four of these outputs have been negated. [#1061](https://github.com/iodide-project/pyodide/pull/1061)
 
 ## Version 0.16.1
 *December 25, 2020*
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -35,7 +35,7 @@
 
 ### Fixed
 - getattr and dir on JsProxy now report consistent results and include all names defined on the Python dictionary backing JsProxy. [#1017](https://github.com/iodide-project/pyodide/pull/1017)
-- `JsProxy.__bool__` now produces better results: it used to be `bool(window)` and `bool(zero-arg-callback)` were `False`, and `bool(empty_js_set)` and `bool(empty_js_map)` were `True`. All four of these outputs have been negated. [#1061](https://github.com/iodide-project/pyodide/pull/1061)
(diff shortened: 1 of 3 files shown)


dalcde:
I must be missing something obvious; what is wrong with `!!val` ? When
the object is a PyProxy?

hoodmane:
They wanted empty containers to be falsey. `!![] === true` in javascript. Question is whether we are trying to be like javascript or like python in our answers.

src/core/hiwire.c
```diff
@@ -346,7 +346,17 @@ EM_JS_NUM(int, hiwire_get_length, (JsRef idobj), {
 EM_JS_NUM(bool, hiwire_get_bool, (JsRef idobj), {
   var val = Module.hiwire.get_value(idobj);
   // clang-format off
-  return (val && (val.length === undefined || val.length)) ? 1 : 0;
+  if(!val){
```
rth, line 349: That was not formatted by clang-format, was it? And CI passes, so something is broken there.

rth (commented):
Thanks! Please add a changelog entry.

src/core/hiwire.c
hoodmane: See `// clang-format off` just above. It's the same problem as usual: we need to protect the javascript `===` and `!==` ops from clang-format so js code with comparisons doesn't get formatted consistently.

docs/changelog.md
```diff
@@ -35,6 +35,7 @@
 
 ### Fixed
 - getattr and dir on JsProxy now report consistent results and include all names defined on the Python dictionary backing JsProxy. [#1017](https://github.com/iodide-project/pyodide/pull/1017)
+- `JsProxy.__bool__` now produces better results: it used to be `bool(window)` and `bool(zero-arg-callback)` were `False`, and `bool(empty_js_set)` and `bool(empty_js_map)` were `True`. All four of these outputs have been negated. [#1061](https://github.com/iodide-project/pyodide/pull/1061)
```
rth, line 38: Nit,

maybe we could find a more specific synonym to "better results": "more consistent" etc
"were `False`, and" > "were `False`, while"

"All four of these outputs have been negated." -> "now produce False."

hoodmane: I changed the language, it's clearer now but still not perfect.

hoodmane:
Thanks!

PR #1062, opened 2021-01-07
Fail macros

hoodmane:
I unified the QUIT macros into `error_handling.h` and renamed them `FAIL`. There are four of them:

1. `FAIL()` -- unconditional failure
2. `FAIL_IF_NULL(ref)` -- fail if `ref == NULL` 
3. `FAIL_IF_MINUS_ONE(num)` -- fail if `num == -1` 
4. `FAIL_IF_ERR_OCCURRED(num)` -- fail if `PyErr_Occurred()`.

Additionally if the feature flag `DEBUG_F` is defined `FAIL` prints a message like:
`Raised exception on line 231 in func JsImport_GetAttr, file jsimport.c` with a javascript / wasm stack trace.

I also added a new macro `hiwire_CLEAR` which is like `Py_CLEAR`. The point is that unlike `hiwire_decref`, `hiwire_CLEAR` is idempotent so I don't have to worry about doing it twice. This eliminates error handling case work and reduces the pain of handling errors in C.

I also went through the beginning of `JsProxy` and applied new error handling approach to a few functions there.

### DIFF
--- a/src/core/hiwire.c
+++ b/src/core/hiwire.c
@@ -118,6 +118,11 @@ EM_JS_NUM(errcode, hiwire_decref, (JsRef idval), {
   Module.hiwire.decref(idval);
 });
 
+EM_JS_NUM(errcode, hiwire_log_error, (char* msg), {
+  let jsmsg = UTF8ToString(msg);
+  console.error(jsmsg);
+});
+
 EM_JS_REF(JsRef, hiwire_int, (int val), {
   return Module.hiwire.new_value(val);
 });
--- a/src/core/hiwire.c
+++ b/src/core/hiwire.c
@@ -118,11 +118,6 @@ EM_JS_NUM(errcode, hiwire_decref, (JsRef idval), {
   Module.hiwire.decref(idval);
 });
 
-EM_JS_NUM(errcode, hiwire_log_error, (char* msg), {
-  let jsmsg = UTF8ToString(msg);
-  console.error(jsmsg);
-});
-
 EM_JS_REF(JsRef, hiwire_int, (int val), {
   return Module.hiwire.new_value(val);
 });
--- a/src/core/error_handling.c
+++ b/src/core/error_handling.c
@@ -4,6 +4,11 @@
 #include "jsproxy.h"
 #include <emscripten.h>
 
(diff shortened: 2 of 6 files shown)


src/core/hiwire.c
```diff
@@ -118,6 +118,11 @@ EM_JS_NUM(errcode, hiwire_decref, (JsRef idval), {
   Module.hiwire.decref(idval);
 });
 
+EM_JS_NUM(errcode, hiwire_log_error, (char* msg), {
```
dalcde, line 121: Does this belong to the hiwire namespace?

hoodmane: I don't know. What is the purview of the hiwire namespace? This is a short `EM_JS` function. I suppose your reasoning must be that it takes a `char* msg` and takes care of alloc and dealloc whereas hiwire mostly would take a `JsRef` and not explicitly allocate? It would be a better api if it took a `JsRef` though.

I could see moving it to `error_handling.c` in either case. `hiwire` is a bit of a weird file anyways. Most of the stuff in there is helpers for `JsRef` with helpers for random other stuff mixed in.

dalcde: error_handling would make more sense to me.

hoodmane: Okay, will do that then. Thanks Dexter!

rth (approved):
It's nice to see consistent error handling. LGTM unless @dalcde has other comments.

hoodmane:
@dalcde Can you merge this after tests pass? They are pretty important for any other code I'm going to write.

hoodmane:
Actually I take it back, I need to double check that the tracing feature flags that I added work as expected.

hoodmane:
Okay I messed around a bit with some adding another tracing macro but I think it wasn't worth it so I reverted back to "Moved log_error". I checked the existing `DEBUG_F` feature flag and it works as intended. 
Please merge this.

dalcde:
If you simply force pushed back to the original commit it wouldn't have to pass through CI again

hoodmane:
Okay. I was asked not to force push things that have been reviewed, I don't know what the proper exceptions are yet. Thanks for the tip.

PR #1065, opened 2021-01-07
Add `version` field to pyodide in main.c

### LINKED ISSUE #922: Funny bug

hoodmane:
I made a couple of API changes:
1. `pyodide.version` is now a string constant not a function and no longer imports the `pyodide` python package into the main user namespace.
2. I added `pyodide.pythonVersion` which is the same as `sys.version`.

This fixes #922, but without really addressing it's root cause.


hoodmane:
It looks like for some reason `selenium.run_js("return pyodide.pythonVersion;")` has an extra newline at the beginning. Could this be a `conftest` bug?

hoodmane:
Nope, the test is functioning correctly, `pyodide.pythonVersion` is the empty string.

rth:
+1 to make `version` a string constant. 

I'm less sure about adding `pythonVersion`. The python version is only one thing among  others that describe the pyodide contents, and we don't want to expose them all manually in JS. `pyodide.runPython("import sys; sys.version")` is fairly short and very versatile.

hoodmane:
> I'm less sure about adding `pythonVersion`

Sounds reasonable, will remove it.

Issue #984, opened 2020-12-31
Remove `pyodide.get_completions` and `pyodide.repr`


hoodmane:
`pyodide.get_completions` should be removed, it is a bad reexport of the Jedi API. A couple of days ago I think I was talking to someone who was confused by it. I also got confused when I tried to use `get_completions` and finally went to the Jedi docs and used Jedi directly. I don't understand what value `get_completions` adds compared to telling people that Jedi is available and suggesting they use it themselves. We could update the docs to prominently link to Jedi, though we should add a warning about the recursion errors #821.

`pyodide.repr` is also a bad API. It is not idiomatic javascript. We should get rid of it in favor of `PyProxy.toString()` or perhaps add `PyProxy.repr()`.

See also my other issues about public APIs #926 and #870.

rth:
I think the motivation for `get_completion` was that different iodide language plugins needed to have this method to make competition work. +1 to remove it and make jeudi a regular package (i.e. remove it from the core pyodide package) which should reduce the initial download size by 1MB or so.

>  We should get rid of it in favor of PyProxy.toString() or perhaps add PyProxy.repr()

We could indeed.

We would need to think how to handle breaking changes. In the python world one would first deprecate and then remove in a version or two, but I'm not sure that would work for us. So maybe making the change, and make it very prominent in the change log (maybe with badges as in https://scikit-learn.org/stable/whats_new/v0.24.html) would be fine.

dalcde:
>We would need to think how to handle breaking changes. In the python world one would first deprecate and then remove in a version or two, but I'm not sure that would work for us. So maybe making the change, and make it very prominent in the change log (maybe with badges as in https://scikit-learn.org/stable/whats_new/v0.24.html) would be fine.

 From the semver point of view, I would consider upgrading from 0.x._ to 0.y._ to be a "major" release (similar to 1._._ to 2._._). For python it's within the same major release, so they "shouldn't" remove anything at all.

rth:
> From the semver point of view, I would consider upgrading from 0.x._ to 0.y._ to be a "major" release

[...]

PR #1066, opened 2021-01-07
API change: Removed completions api

hoodmane:
As I suggested in #984, I don't think we should re export the completions API. The exported API is very modestly simpler, has many fewer options than the Jedi API, and is less well documented. We should send users to the Jedi docs and have them use it directly. If we think that Jedi should offer a simpler "very basic API" we should suggest a change to Jedi.


rth:
I agree, I think the initial motivation was compatibility with iodide, but in practice it's indeed better to use jedi directly.  And if iodide development continues in the future, it could be re-defined as part of a iodide specific plugin, but not great to maintain it in the public API.

It's a breaking change so definitely we need a changelog entry.

@casatir @gzuidhof Please let us if this change would be problematic for you.

hoodmane:
I don't see why Iodide shouldn't define this themselves if they need it. There's no C code present, it doesn't rely on internals, and in fact there's nothing about "get completions from jedi in iodide" that is different from "get completions from jedi in native" (other than all the recursion errors, but the `get_completions` API did nothing to mitigate them).

gzuidhof:
Starboard doesn't make use of it, so no objections from me :). Thanks for asking!

casatir:
This is fine for me. It is not used in Basthon and I plan to use Jedi directly.

hoodmane:
> I plan to use Jedi directly.

Presumably because the reexposed `get_completions` is too weak for most practical purposes.

casatir:
As a side note, `src/pyodide-py/pyodide/console.py` will be re-created by #875 (or its children). I am working on it: everything is setup, just polishing the splitting into several PR.

PR #1067, opened 2021-01-07
API Change: Remove repr API

hoodmane:
The repr API is unnecessary (instead use `PyProxy.toString` or import `repr` from python yourself). Furthermore it has a bug that is like the reverse of the bug in #922: 

```javascript
>>> pyodide.repr.toString()
 TypeError: pyodide.repr is undefined
>>> pyodide.globals.toString()
"<module 'builtins' (built-in)>"
>>> pyodide.repr.toString()
"<built-in function repr>"
```
This was done in v0.15.0. The output is different now but the same bug is still here. I think the fact that nobody has reported this bug suggests that nobody was using this redundant API in the first place.

See also #1066 and #984.


rth (commented):
Thanks! This needs a changelog entry as well.

PR #1072, opened 2021-01-08
Remove sitecustomize

dalcde:
It is supposed to help tests pass, but tests seem to pass without it

### DIFF
--- a/src/sitecustomize.py
+++ b/src/sitecustomize.py
@@ -1,9 +1,0 @@
-"""
-Pyodide-specific startup code that's always run at Python startup.
-"""
-
-# We weren't invoked at the commandline, but set some commandline arguments as
-# if we were anyway. Helps many of the CPython tests pass.
-import sys
-
-sys.argv = ["pyodide"]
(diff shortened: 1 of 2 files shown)


src/sitecustomize.py
```diff
@@ -1,9 +0,0 @@
-"""
-Pyodide-specific startup code that's always run at Python startup.
-"""
-
-# We weren't invoked at the commandline, but set some commandline arguments as
-# if we were anyway. Helps many of the CPython tests pass.
-import sys
-
-sys.argv = ["pyodide"]
```
rth, line 9: Yes, we can remove these two lines, however I wonder it it would make sense to keep the file. It's a standard mechanism to execute install specific code at startup, which might be better than doing it at import. For instance, if we want to monkeypatch some stdlib modules to have different behavior this would be the place for it IMO.

dalcde:
I think we can add back the file if we need it, but the way we currently
do it appears to be to simply modify the modules directly.

PR #1073, opened 2021-01-08
Stop hardcoding file_packager.py directory

dalcde:
This makes our build system more flexible, especially if we start using unpatched emscripten in the future.

### DIFF
--- a/pyodide_build/common.py
+++ b/pyodide_build/common.py
@@ -4,16 +4,8 @@ import shutil
 
 ROOTDIR = Path(__file__).parents[1].resolve()
 TOOLSDIR = ROOTDIR / "tools"
-
-# Use emcc.py because emcc may be a ccache symlink
-_EMCC_PATH = shutil.which("emcc.py")
-if _EMCC_PATH is None:
-    print("emcc.py not found. Setting file_packager.py path to /dev/null")
-    PACKAGERPATH = Path("/dev/null")
-else:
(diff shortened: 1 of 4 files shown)


pyodide_build/common.py
```diff
@@ -1,9 +1,18 @@
...
 
 ROOTDIR = Path(__file__).parents[1].resolve()
 TOOLSDIR = ROOTDIR / "tools"
-PACKAGERDIR = ROOTDIR / "emsdk" / "emsdk" / "fastcomp" / "emscripten" / "tools"
+
+# Use emcc.py because emcc may be a ccache symlink
+_EMCC_PATH = shutil.which("emcc.py")
+if _EMCC_PATH is None:
+    print("emcc.py not found. Setting file_packager.py path to /dev/null")
+    PACKAGERPATH = Path("/dev/null")
```
rth, line 12: Shouldn't we rather error here?

dalcde: That would break all pyodide_build tests

dalcde: I could make it a function instead

rth: Otherwise maybe keep to to None here, and then in the other place where we need it, since that's not used in test, check it if's None and error. 

I'm just not very keen on the `Path(/dev/null)` because if it's actually used somewhere it will be confusing to debug.

PR #1077, opened 2021-01-08
Refcount support macros, more macro documentation 

hoodmane:
I added four new macros to "error_handling.h". Without this, it's a bit tedious to ensure that the variables cleared at the end of the function are exactly the same as those declared at the beginning. With this, it's a quick copy+paste, replace `DECLARE` with `CLEAR`. 
I applied this to one example in jsproxy for illustration.

I also added thorough explanations of the intended usage of the macros in "error_handling.h". This documentation should eventually also be added to contribution docs, though once I apply it consistently it should also be easy for new contributors to just imitate the way things are already done.

### DIFF
--- a/src/core/error_handling.h
+++ b/src/core/error_handling.h
@@ -1,16 +1,10 @@
 #ifndef ERROR_HANDLING_H
 #define ERROR_HANDLING_H
 
-/**  Wrap EM_JS so that it produces functions that follow the Python return
- *  conventions. We catch javascript errors and proxy them and use
- *  `PyErr_SetObject` to hand them off to python. We need two variants, one
- *  for functions that return pointers / references (return 0)
- *  the other for functions that return numbers (return -1).
- */
-
 typedef int errcode;
 
-// Hiwire wants to import us for errcode, so import hiwire after typedef.
+// Hiwire wants to import us for errcode, so import hiwire after "typedef int
+// errcode;".
 #include "hiwire.h"
 #include <emscripten.h>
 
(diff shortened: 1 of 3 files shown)


hoodmane:
I think the license on `map.h` is appropriate for us: it says:

>  * Permission is hereby granted, free of charge, to any person
>  * obtaining a copy of this software and associated documentation
>  * files (the "Software"), to deal in the Software without
>  * restriction, including without limitation the rights to use, copy,
>  * modify, merge, publish, distribute, sublicense, and/or sell copies
>  * of the Software, and to permit persons to whom the Software is
>  * furnished to do so, subject to the following conditions:

And then the conditions are 1. keep this copyright and permissions notice, 2. no warranty and no liability, 3. don't use the names of the authors without permission.

rth:
Sorry I'm -1 for this change. Previously we had,
```c
  Py_CLEAR(object__dir__);
  Py_CLEAR(keys);
  Py_CLEAR(result_set);
  hiwire_decref(iddir);
  Py_CLEAR(pydir);
  Py_CLEAR(null_or_pynone);
```
which although repetitive is fairly easy to find documentation on since it's part of the C Python API, and a function that we define.

Now with this change we would have,
```c
  CLEAR_PyObjects(object__dir__, keys, result_set, pydir, null_or_pynone, );
  CLEAR_JsRefs(iddir);
```
which is indeed shorter, however to understand what it really does, a contributor would have to go to the `CLEAR_PyObjects` macro then to the `_eh_MAP_ALLOW_TRAILING_COMMA` then to the definition of `MAP_UD` and `_eh_DO_F_If_NONEMPTY` which branches out to `_eh_IF_EMPTY`, `_eh_TEST`, `_eh_IS_EMPTY`, `_eh_HAS_COMMA`, `_eh_TRIGGER_PARENTHESIS_`, `_eh_HAS_COMMA_HELPER`, `MAP1_UD`, etc before getting completely lost in all those macros. All that to call  `Py_CLEAR` 5 time. So I think in this case keeping the original version is much simpler and readable even it means some repetition. Code repetition is sometime preferable to the alternatives.

hoodmane:
@rth Not offended, I have mixed feelings about it too. The problem is that I got nerd sniped by the macro preprocessor so I had to do it.

hoodmane:
The only real benefit is that I _have_ been spending a fair amount of time manually converting `PyObject* x = NULL;` to `Py_CLEAR(x);`, but maybe this is a job for sed or something?

hoodmane:
I can open a separate PR with just the docs update.

rth:
I suspect you are still trying to add some LaTeX spirit in there, but I'm not sure it's the right use case :) Thanks for experimenting with this in any case !

hoodmane:
I have a major weakness for macros.

hoodmane:
> a contributor would have to go to the `CLEAR_PyObjects` macro then to the `_eh_MAP_ALLOW_TRAILING_COMMA` then to the definition of `MAP_UD` and `_eh_DO_F_If_NONEMPTY` which branches out to `_eh_IF_EMPTY`, `_eh_TEST`, `_eh_IS_EMPTY`, `_eh_HAS_COMMA`, `_eh_TRIGGER_PARENTHESIS_`, `_eh_HAS_COMMA_HELPER`, `MAP1_UD`, etc before getting completely lost in all those macros. 

Seems to be a fact about many macro languages that it's impossible to implement nontrivial logic without a forest of helper macros.

PR #1080, opened 2021-01-09
Enable emjs error handling

hoodmane:
Enable the error handling set up in #1051.

### DIFF
--- a/src/core/error_handling.c
+++ b/src/core/error_handling.c
@@ -22,7 +22,7 @@ error_handling_init()
     Module.handle_js_error = function(e)
     {
       let err = Module.hiwire.new_value(e);
-      PyodideErr_SetJsError(err);
+      _PyodideErr_SetJsError(err);
       Module.hiwire.decref(err);
     };
   });
--- a/Makefile
+++ b/Makefile
@@ -16,8 +16,7 @@ CC=emcc
 CXX=em++
 OPTFLAGS=-O2
 CFLAGS=$(OPTFLAGS) -g -I$(PYTHONINCLUDE) -fPIC \
-	-Wno-warn-absolute-paths -Werror=int-conversion -Werror=incompatible-pointer-types \
-	-D DEBUG_F
+	-Wno-warn-absolute-paths -Werror=int-conversion -Werror=incompatible-pointer-types
 
 LDFLAGS=\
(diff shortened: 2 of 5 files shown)


hoodmane:
@dalcde This most recent failure gave a useful error message where it wouldn't have prior to your conftest work.

hoodmane:
WOOHOO Tests pass! =D

hoodmane:
I guess these test failures are real: this doesn't work in a webworker in firefox. It does work in chrome or not in a webworker.

src/core/error_handling.c
```diff
@@ -22,7 +22,7 @@ error_handling_init()
     Module.handle_js_error = function(e)
     {
       let err = Module.hiwire.new_value(e);
-      PyodideErr_SetJsError(err);
+      _PyodideErr_SetJsError(err);
```
rth, line 25: There is no such function on master?

hoodmane: AHA. But there is, defined just above.
https://github.com/iodide-project/pyodide/blob/0c52f180f6c2b7960a9146fcdd733f10e999b6b8/src/core/error_handling.c#L12

That's right, for calling C functions from EM_JS, the functions are automatically prefixed with an underscore. Why? Heck if I know, it's confusing. I suppose it's to prevent C shadowing? When I get back to writing C contributor docs I might try to make a "unpleasant surprises you may experience starting out" section with this fact in it (I have to remember what other things I was confused about, there must be too many to list...).

hoodmane: Oh according to [the C spec](https://en.cppreference.com/w/c/language/identifier#Reserved_identifiers):
"All identifiers that begin with an underscore followed by a capital letter" "are reserved and may not be declared in a program".

So if you declared your own `_PyodideErr_SetJsError` in C that was shadowed by this, you would in fact be violating the C spec.

rth: Right, by my question is why it needs changing since it was working fine without an `_` before?

hoodmane: Before if that code path ever occurred at runtime it lead to a `ReferenceError`.

hoodmane:
@rth I think this can be accepted unless you have any further reservations?

PR #1083, opened 2021-01-09
Separate ns variable in eval_code into globals and locals

hoodmane:
This will allow `pyodide.js` to avoid pollution of global namespace by using `eval_code` with new empty dict.

### DIFF
--- a/src/pyodide-py/pyodide/_base.py
+++ b/src/pyodide-py/pyodide/_base.py
@@ -50,19 +50,24 @@ class CodeRunner:
     Parameters
     ----------
     globals
-        The global scope in which to execute code. This is used as the `exec` `globals` parameter.
-        See [the exec documentation](https://docs.python.org/3/library/functions.html#exec) for more info.
+        The global scope in which to execute code. This is used as the `exec`
+        `globals` parameter. See
+        [the exec documentation](https://docs.python.org/3/library/functions.html#exec)
+        for more info.
     locals
-        The local scope in which to execute code. This is used as the `eval` `globals` parameter.
-        See [the exec documentation](https://docs.python.org/3/library/functions.html#exec) for more info.
-    mode
-        'last_expr' , 'last_expr_or_assign' or 'none',
-        specifying what should be evaluated and what should be executed.
-        'last_expr' will return the last expression
-        'last_expr_or_assign' will return the last expression
-        or the last (named) assignment.
-        'none' will always return `None`.
-        Other values will be interpreted as 'none'.
+        The local scope in which to execute code. This is used as the `exec`
+        `locals` parameter. As with `exec`, if `locals` is absent, it is set equal
+        to `globals`. See
+        [the exec documentation](https://docs.python.org/3/library/functions.html#exec)
+        for more info.
+    return_mode
+        Specifies what should be returned, must be one of 'last_expr',
+        'last_expr_or_assign' or `None`. On other values an exception is raised.
+
+        'last_expr' -- return the last expression
+        'last_expr_or_assign' -- return the last expression or the last
+        (named) assignment.
+        `None` -- always return `None`.
     quiet_trailing_semicolon
         wether a trailing semicolon should 'quiet' the result or not.
         Setting this to `True` (default) mimic the CPython's interpret
@@ -86,7 +91,7 @@ class CodeRunner:
         self,
         globals: Optional[Dict[str, Any]] = None,
         locals: Optional[Dict[str, Any]] = None,
-        mode: str = "last_expr",
+        return_mode: str = "last_expr",
         quiet_trailing_semicolon: bool = True,
         filename: str = "<exec>",
     ):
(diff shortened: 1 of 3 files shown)


casatir:
+1 for this. As a side effect, it replaces the `ns` parameter which I found not really explicit.

src/pyodide-py/pyodide/_base.py
```diff
@@ -287,12 +290,19 @@ def eval_code(
...
+        globals=globals,
+        locals=locals,
+        mode=mode,
+        quiet_trailing_semicolon=quiet_trailing_semicolon,
+        filename=filename,
+    ).run(code)
 
 
-async def _eval_code_async(
+async def eval_code_async(
```
casatir, line 302: One micro step towards official async support?

casatir (approved):

casatir:
> avoid pollution of global namespace

Can you add some tests that demonstrate this please?

src/pyodide-py/pyodide/_base.py
hoodmane: Yeah.

hoodmane:
> Can you add some tests that demonstrate this please?

Of course =)

hoodmane:
@casatir How's that look for a test?

src/pyodide-py/pyodide/_base.py
```diff
@@ -81,12 +81,14 @@ class CodeRunner:
 
     def __init__(
         self,
-        ns: Optional[Dict[str, Any]] = None,
+        globals: Optional[Dict[str, Any]] = None,
+        locals: Optional[Dict[str, Any]] = None,
```
rth, line 85: You are shadowing the `global` / `locals` objects here, which is not great. Maybe `ns_globals` and `ns_locals`?
You also need to update the docstring above and in `eval_*`

rth (commented):
I agree it very useful. Please add a changelog entry.

src/pyodide-py/pyodide/_base.py
hoodmane: Yeah, it's not great. I've seen this done in enough other peoples' code that I've gotten used to it. `ns_globals` seems fine to me. @casatir?

hoodmane: @rth Actually, look at the built in `eval` and `exec` functions:
https://docs.python.org/3/library/functions.html#eval
https://docs.python.org/3/library/functions.html#exec
They also call these arguments `globals` and `locals`. I think it's a good idea to name them `globals` and `locals` here too.

src/pyodide-py/pyodide/_base.py
```diff
@@ -237,21 +242,22 @@ async def run_async(self, code: str) -> Any:
...
             if iscoroutine(res):
                 res = await res
             return res
 
 
 def eval_code(
     code: str,
-    ns: Dict[str, Any],
+    globals: Dict[str, Any] = None,
+    locals: Dict[str, Any] = None,
```
rth, line 260:
```suggestion
    globals: Optional[Dict[str, Any]] = None,
    locals: Optional[Dict[str, Any]] = None,
```

src/pyodide-py/pyodide/_base.py
```diff
@@ -262,8 +268,12 @@ def eval_code(
...
     code
        the Python code to run.
-    ns
-       `locals()` or `globals()` context where to execute code.
+    globals
+        The global scope in which to execute code. This is used as the `exec` `globals` parameter.
+        See [the exec documentation](https://docs.python.org/3/library/functions.html#exec) for more info.
+    locals
+        The local scope in which to execute code. This is used as the `eval` `globals` parameter.
+        See [the exec documentation](https://docs.python.org/3/library/functions.html#exec) for more info.
```
rth, line 276: Please wrap docstrings to 88 char as well.

src/pyodide-py/pyodide/_base.py
```diff
@@ -49,9 +49,12 @@ class CodeRunner:
...
     ----------
-    ns
-        `locals()` or `globals()` context where to execute code.
-        This namespace is updated by the subsequent calls to `run()`.
+    globals
+        The global scope in which to execute code. This is used as the `exec` `globals` parameter.
+        See [the exec documentation](https://docs.python.org/3/library/functions.html#exec) for more info.
+    locals
+        The local scope in which to execute code. This is used as the `eval` `globals` parameter.
+        See [the exec documentation](https://docs.python.org/3/library/functions.html#exec) for more info.
```
rth, line 57: Same

src/pyodide-py/pyodide/_base.py
```diff
@@ -49,9 +49,12 @@ class CodeRunner:
...
     Parameters
     ----------
-    ns
-        `locals()` or `globals()` context where to execute code.
-        This namespace is updated by the subsequent calls to `run()`.
+    globals
+        The global scope in which to execute code. This is used as the `exec` `globals` parameter.
+        See [the exec documentation](https://docs.python.org/3/library/functions.html#exec) for more info.
+    locals
+        The local scope in which to execute code. This is used as the `eval` `globals` parameter.
```
rth, line 56: `eval` `locals` parameter. ?

Also indicate that when not provided `globals` parameter is used.

rth (commented):
A few remaining comments, otherwise LGTM. Thanks.

Also please add a changelog entry.

hoodmane:
I also changed `mode` ==> `return_mode` (more explicit) and changed it so that it will raise an error if passed an unexpected mode. Otherwise it's way too easy for someone to pass i.e., `"last-expr-or-assign"` or `"last_exp_or_assign"` and be confused when it does something different than expected. Could be friendly to offer:
```
class ReturnMode:
    LAST_EXPR = "last_expr"
    LAST_EXPR_OR_ASSIGN = "last_expr_or_assign"
    NONE = None
```

PR #1084, opened 2021-01-09
Refactor loadPackage

dalcde:
From the API point of view, this removes the messageCallback and errorCallback arguments in loadPackage, runPythonAsync etc. These are used for logging. Previously, any message will be written to the console, and then {message,error}Callback is called with the message as an argument. In most of our use cases, these callbacks do nothing.

If we were to reintroduce this feature, I think we should introduce a logger interface, and users can supply custom loggers. This lets them choose to not output to stdout/stderr as well.

Under the hood, the original motivation of this patch was to get rid of using packageCounter to determine whether all packages have been loaded. This is prone to error if we change the way we load packages (which I attempted in a different branch until I encountered an emscripten bug).

One way or another, I ended up rewriting a lot of the code into something with a clearer control flow imo.


src/pyodide.js
```diff
@@ -171,102 +161,127 @@ var languagePluginLoader = new Promise((resolve, reject) => {
...
-      let packageList = Array.from(Object.keys(toLoad));
-      _messageCallback(`Loading ${packageList.join(', ')}`)
+    // Add a promise that rejects when any exceptions is thrown in the process
+    // of loading a script. The promise never resolves, and we combine it
+    // with other promises via Promise.race.
+    let windowErrorHandler;
+    let windowErrorPromise = new Promise((_res, rej) => {
+      windowErrorHandler = e => rej(e.message);
+      self.addEventListener('error', windowErrorHandler);
+    });
```
hoodmane, line 221: Do you know what this `windowErrorHandler` is supposed to handle? I think unless we can find a case where this gets triggered we should remove it. It seems dangerous to me: it could swallow error messages and respond to errors that are not actually related to loading the package. I would suggest removing it and if that causes trouble trying resolving the resulting problems in a less dangerous way.

src/pyodide.js
```diff
@@ -171,102 +161,127 @@ var languagePluginLoader = new Promise((resolve, reject) => {
...
-        }
-      };
+    // Wait for all scripts to be loaded first. Then all the run dependencies
+    // have been added. Then wait for the dependencies to be removed.
+    let successPromise = Promise.all(scriptPromises).then(() => runDepPromise);
+    try {
+      await Promise.race([ successPromise, windowErrorPromise ]);
+    } finally {
+      delete self.pyodide._module.monitorRunDependencies;
+      self.removeEventListener('error', windowErrorHandler);
```
hoodmane, line 249: Both of these lines make me really unhappy (I know this is just stuff you moved around).

src/pyodide.js
```diff
@@ -171,102 +161,127 @@ var languagePluginLoader = new Promise((resolve, reject) => {
...
-            self.pyodide._module.monitorRunDependencies();
-          }
-        });
-      }
+    let resolveMsg = `Loaded `;
+    if (packageList.length > 0) {
+      resolveMsg += packageList.join(', ');
+    } else {
+      resolveMsg += 'no packages'
+    }
```
hoodmane, line 263:
```suggestion
    let resolveMsg;
    if (packageList.length > 0) {
      let package_names = packageList.join(', ');
      resolveMsg =  `Loaded ${packageList}`;
    } else {
      resolveMsg = 'No packages loaded';
    }
```

src/pyodide.js
```diff
@@ -171,102 +161,127 @@ var languagePluginLoader = new Promise((resolve, reject) => {
...
-      // We have to invalidate Python's import caches, or it won't
-      // see the new files. This is done here so it happens in parallel
-      // with the fetching over the network.
-      self.pyodide.runPython('import importlib as _importlib\n' +
-                             '_importlib.invalidate_caches()\n');
-    });
+    if (!isFirefox) {
+      await preloadWasm();
+    }
+    console.log(resolveMsg);
```
hoodmane, line 268: Another place we unconditionally write `info` level messages to the console.

hoodmane (commented):
Thanks for taking a stab at this. I still have lots of issues with this code but this is a clear and manageable improvement. I would have trouble leaving enough stuff alone to maintain a manageable diff (cf #998).
Maybe change `toLoad` from an object to a `Map`? I think that'd be better.

src/pyodide.js
dalcde: We load a package by loading an external javascript file via inserting a `<script>` tag. There is no reasonable way to catch errors from that file without blanket catching all errors. One way we can trigger this is by removing the `.data` file, but that's unpleasant to do in a test environment

src/pyodide.js
dalcde: As I mentioned in the OP, I think the right way out is to let user supply loggers. Before that, I'm maintaining the original behaviour...

src/pyodide.js
```diff
@@ -171,102 +161,127 @@ var languagePluginLoader = new Promise((resolve, reject) => {
...
+    };
 
-      let packageList = Array.from(Object.keys(toLoad));
-      _messageCallback(`Loading ${packageList.join(', ')}`)
+    // Add a promise that rejects when any exceptions is thrown in the process
+    // of loading a script. The promise never resolves, and we combine it
+    // with other promises via Promise.race.
+    let windowErrorHandler;
+    let windowErrorPromise = new Promise((_res, rej) => {
+      windowErrorHandler = e => rej(e.message);
```
hoodmane, line 219:
```suggestion
      windowErrorHandler = e => {
         console.error("Unhandled error. We don't know what it is or whether it is related to 'loadPackage' but out of an abundance of caution we will assume that loading failed.");
         console.error(e);
         rej(e.message);
      };
```
We're logging info level stuff all over the place, but the one time we grab any and all unhandled errors regardless of whether they have anything to do with us and we don't even log it!

docs/js-api/pyodide_loadPackagesFromImports.md
```diff
@@ -20,11 +20,9 @@ See also {ref}`js_api_pyodide_runPythonAsync`.
...
 | *code*            | String   | code to inspect for packages to load. |
-| *messageCallback* | function | A callback, called with progress messages. (optional) |
-| *errorCallback*   | function | A callback, called with error/warning messages. (optional) |
 
 *Returns*
 
 | name       | type    | description                              |
 |------------|---------|------------------------------------------|
-| *result*   | Promise | Resolves to undefined on success. |
+| *result*   | Promise | Resolves to undefined on success. Rejects if a fatal error occurs when loading imports. |
```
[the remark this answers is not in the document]
dalcde: Well yes. I think if a package's url exists but loading it fails, the error is non-fatal. Here "fatal" is defined to be "causes the original implementation to return early".

dalcde:
Just for the record, the very original motivation was to get rid of our custom preloadWasm code and use the one that comes with emscripten, but turns out that's not compatible with `-s LZ4`

hoodmane:
See [window.onerror](https://developer.mozilla.org/en-US/docs/Web/API/GlobalEventHandlers/onerror):

* When a JavaScript runtime error (including syntax errors and exceptions thrown within handlers) occurs, an error event using interface `ErrorEvent` is fired at window and `window.onerror()` is invoked (as well as handlers attached by `window.addEventListener` (not only capturing)).
* When a resource (such as an `` or `<script>`) fails to load, an error event using interface `Event` is fired at the element that initiated the load, and the `onerror()` handler on the element is invoked. These error events do not bubble up to window, but can be handled with a `window.addEventListener` configured with `useCapture` set to true.

I think we're trying to catch the second type of errors? So the correct way to do it is to attach an `onError` element to the `script` tag? What else is going to generate uncaught errors?

dalcde:
I think I am concerned mostly about the first kind of error within the loaded script. The second kind is captured by the `loadScript` function already.

hoodmane:
It seems sad to be unable to catch the error any closer to the source, but if we add my noisy and explicit `console.error` message, I think it's okay. Can you make a test that hits that execution path and cross reference against the test in the code though? I think it's natural to wonder why the error can't be isolated any better.

Incidentally, it's sort of remarkable that there isn't a less error prone and hacky way to load a script. Why doesn't the `loadScript` api exist on main? Because it's blocking? Why is it blocking?! Browsers are so weird >_<

dalcde:
Well, we can fetch the source code and eval it ....

hoodmane:
> fetch the source code and eval

Do we lose anything by doing that? Maybe weaker CORS restrictions?

hoodmane:
Maybe it's forbidden by "Content Security Policy" or something?

docs/js-api/pyodide_loadPackagesFromImports.md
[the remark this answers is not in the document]
hoodmane: Might be helpful to expand on this a bit in the body text before these tables.

hoodmane:
Okay so if I understand correctly for the global window handler: 
1. this is necessary and can happen 
2. the other ways to do this they have their own drawbacks and it's unclear if they are any better, and 
3. you know how to cause the code to trigger but it's hard to test

In that case I guess the most reasonable thing is just to put a comment explaining the situation and saying that it is technically necessary but unfortunate.

src/pyodide.js
```diff
@@ -171,102 +161,127 @@ var languagePluginLoader = new Promise((resolve, reject) => {
...
+    // so we must be careful to only await for this after we know all
+    // runDependencies have been added.
+    let runDepPromise = Promise.resolve();
+    // runDepResolve is the resolve function of runDepPromise if it is
+    // unresolved, undefined otherwise.
+    let runDepResolve = undefined;
+
+    // We must start monitoring before we load any script, as this function is
+    // triggered by the script loading.
+    self.pyodide._module.monitorRunDependencies = (n) => {
```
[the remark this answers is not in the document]
dalcde: I think making it a class is a bit overkill if we don't use it elsewhere.

docs/js-api/pyodide_loadPackagesFromImports.md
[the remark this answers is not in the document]
dalcde: I've reverted these changes. "Rejects if a fatal error occurs" is somewhat of a tautology anyway, and I don't think my rejection strategy is particularly exotic now.

src/pyodide.js
dalcde: I'm open to suggestions, but otherwise I'm satisfied that I have left the code in a slightly nicer state.

src/pyodide.js
```diff
@@ -120,26 +106,24 @@ var languagePluginLoader = new Promise((resolve, reject) => {
...
-          _errorCallback(`URI mismatch, attempting to load package ` +
-                         `${pkg} from ${package_uri} while it is already ` +
-                         `loaded from ${loadedPackages[pkg]}!`);
-          return;
+          err(`URI mismatch, attempting to load package ` +
+              `${pkg} from ${package_uri} while it is already ` +
+              `loaded from ${loadedPackages[pkg]}!`);
         } else {
-          _messageCallback(`${pkg} already loaded from ${loadedPackages[pkg]}`)
+          console.log(`${pkg} already loaded from ${loadedPackages[pkg]}`)
```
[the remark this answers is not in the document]
dalcde: See discussion below

rth:
> Previously, any message will be written to the console, and then {message,error}Callback is called with the message as an argument. In most of our use cases, these callbacks do nothing.

In our uses cases yes, but it would allow users to customize where errors/messages are shown.

> If we were to reintroduce this feature, I think we should introduce a logger interface, and users can supply custom loggers.

In my understanding this feature is critical for downstream applications (e.g. Basthon, Starboard Notebook). Granted our error handling there can certainly be improved, but if we remove it we can't really make a release until a new API is includes. I think it would be better to,
 - have a separate PR with as much of refactoring from this PR as possible
 - open an issue and discuss how we want our error handling API to look. A logger might indeed be nice, but I also don't see it as fundamentally different from error callbacks. For instance #792 still has callbacks.
 - make a PR that replaces the logging API if necessary. Just to avoid a situation where we decide after removing them that callback where actually fine.

Also for all changes that break backward compatibility and change the Public API, it would be better to have the changed method/arguments names in the title, as opposed to just "refactoring".  Here this have been, `API Remove messageCallback and errorCallback arguments from loadPackage`

dalcde:
>In my understanding this feature is critical for downstream applications (e.g. Basthon, Starboard Notebook). Granted our error handling there can certainly be improved, but if we remove it we can't really make a release until a new API is includes. I think it would be better to,
> - have a separate PR with as much of refactoring from this PR as possible

I'll re-introduce the callbacks into this PR

> - open an issue and discuss how we want our error handling API to look. A logger might indeed be nice, but I also don't see it as fundamentally different from error callbacks. For instance #792 still has callbacks.

In my mind a logger interface is an object with ".log" and ".error"
functions, so basically combining the two into one. Perhaps I could just
do that and pass in console by default. But I'm thinking of using it
more pervasively, e.g. by default stdout/stderr should go there.

>Also for all changes that break backward compatibility and change the Public API, it would be better to have the changed method/arguments names in the title, as opposed to just "refactoring".  Here this have been, `API Remove messageCallback and errorCallback arguments from loadPackage`

Yeah this PR diverted from its original mission by *quite* a bit.

dalcde:
But I think I'll change the behaviour so that the default callbacks are console.log and console.error, and supplying a custom callback would cause it to stop logging to console.

rth:
> In my mind a logger interface is an object with ".log" and ".error"
functions, so basically combining the two into one.

Yeah, that sound good. But let's still open an issue to discuss it before implementing it? I would like to also get feedback from downstream package maintainers.

hoodmane:
Yeah we should discuss with downstream whether they actually need an error callback. I think the right thing to do is to be able to request a logging level ("error", "warn", "info", "debug"). If they want to execute code when there is an error, what about `try` + `catch`?

src/pyodide.js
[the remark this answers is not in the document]
dalcde: My latest commit at https://github.com/iodide-project/pyodide/pull/1084/commits/2f94fe176c90027a41fea1a58805f5dfcb0a1f12 should make it more pleasant. Would be nice if upstream handled this better though

rth (approved):
Thanks! Should be good to merge unless @hoodmane has more comments.

src/pyodide.js
```diff
@@ -76,198 +70,201 @@ var languagePluginLoader = new Promise((resolve, reject) => {
   }
   // clang-format on
 
-  function loadScript(url, onload, onerror) {
-    if (self.document) { // browser
+  let loadScript = () => {};
```
hoodmane, line 73:
```suggestion
  let loadScript;
```

src/pyodide.js
```diff
@@ -76,198 +70,201 @@ var languagePluginLoader = new Promise((resolve, reject) => {
...
-        // Set up a new Promise chain, since this one failed
-        loadPackagePromise = new Promise((resolve) => resolve());
-        reject(err.message);
+    // Try to catch errors thrown when running a script. Since the script is
+    // added via a script tag, there is no good way to capture errors from
+    // the script only, so try to capture all errors them.
+    //
+    // windowErrorPromise rejects when any exceptions is thrown in the process
+    // of loading a script. The promise never resolves, and we combine it
+    // with other promises via Promise.race.
```
hoodmane, line 158: Can we change it so that we only install the `windowErrorHandler` if we're on the main thread? If we're in a webworker it seems that we don't need it.

hoodmane (approved):
Looks really nice!

hoodmane:
Okay this looks good to me, I think it should be accepted.

rth:
Thanks @dalcde and @hoodmane !

PR #1085, opened 2021-01-09
Var to let in core

hoodmane:
PR #979 replaces `var` ==> `let`. Turns out this is really hard, so I reverted the EM_JS in that PR to using `var`. It still doesn't pass the tests on a webworker though. This PR does var ==> let on the EM_JS. Hopefully it will be a bit easier?

The surprisingly subtle and frustrating bugs that are making this difficult is further proof that `var` needs to be replaced.

This was done with:
`find src -type f -name "*.c"  -or -name "*.h" | xargs sed -E -i 's/\bvar\b/let/g'`

### DIFF
--- a/src/core/error_handling.h
+++ b/src/core/error_handling.h
@@ -32,46 +32,27 @@ log_error(char* msg);
  */
 
 // clang-format off
-#ifdef DEBUG_F
-// Yes, the "do {} while(0)" trick solves the same problem in the same way in
-// javascript!
-#define LOG_EM_JS_ERROR(__funcname__, err)                                     \
-  do {                                                                         \
-    console.error(                                                             \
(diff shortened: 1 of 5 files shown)


src/core/error_handling.h
```diff
@@ -32,27 +32,46 @@ log_error(char* msg);
  */
 
 // clang-format off
+#ifdef DEBUG_F
```
rth, line 35: There are unrelated code changes here which should be removed. Otherwise yes, splitting that PR in smaller chunks in the right approach and this LGTM.

hoodmane: See: "Check out error_handling.h and Makefile from var-to-let" this was intended to improve the error messages generated.

PR #1087, opened 2021-01-09
Document f2c patches

dalcde:
@joemarshall can you have a quick look to see if the documentation is accurate?


joemarshall:
Yep that looks fine

PR #1089, opened 2021-01-09
Use strict in emjs

hoodmane:


hoodmane:
With the help of `"use strict";` I just fixed an interesting peculiarity of the code. The `destroy` function used to globally write to `ptrobj`. Here's the behavior:
```javascript
    destroy : function() {
      ptrobj = Module.PyProxy.getPtr(this);
      // this ^^ is a global assignment
      // because ptrobj is undeclared and sloppy mode semantics.
      __pyproxy_destroy(ptrobj);
      this['$$']['ptr'] = null;
    },
```
Then 
```C
void
_pyproxy_destroy(PyObject* ptrobj)
{
  Py_DECREF(ptrobj);
  // EM_ASM does not see C variables. the ptrobj 
  // below is actually reading the global value of ptrobj.
  // Unfortunately, we also just leaked the value of ptrobj to the global scope,
  // so this always behaves as expected despite being completely wrong.
  EM_ASM(delete Module.PyProxies[ptrobj];);
}
```

dalcde:
At the end, is the "move stuff around in pyproxy" change necessary for tests to pass?

hoodmane:
No, though that change is an improvement. I suppose it should be split into a separate PR.

hoodmane:
Though the diff here isn't super big and splitting it off will take more effort.

hoodmane:
I was just hunting for the change on line 155 which was hard to spot.

dalcde:
I think it is better to keep to one change per PR. Having a change totally unrelated to the PR topic can be very confusing for future people

Issue #1079, opened 2021-01-08
It's hard to figure out where things are defined


hoodmane:
Currently any function on `pyodide_js` can be defined either in `pyodide.js` or in any of the `*_init` functions in `core`, or it could be a python function wrapped in a pyproxy and copied in either of the above places onto `pyodide`.

Similarly, functions on `pyodide_py` can be defined either in `pyodide` or in `core` or can be wrappers of `js` functions.

I don't think this is difficult to fix.

rth:
More work could always be done on this, but I think you improved the situation considerably.

PR #1090, opened 2021-01-09
Add _pyodide_core C extension module

hoodmane:
This is meant to address two issues: 
1. hard to tell which things are defined in core #1079,
2. the bootstrap order is confusing -- `pyodide_py` needs a most of the stuff from `core` in order to work, but `core` imports `pyodide_py` fairly early on in the initialization. Currently it works fine because `pyodide_py` only needs `core` to run it's methods not for initialization, but it's easy to make innocuous-seeming changes that mess it up.

Instead of directly adding stuff to `pyodide_py`, we make a C module called `_core` and add things to this instead. I also added `_core.py` which imports names from `_core` if we are running in the browser and spits out empty shims otherwise. If needed for testing reasons, these shims can be modified. In `_core.py` we can specify which C source file added each object imported from `_core`.


src/tests/test_pyproxy.py
```diff
@@ -62,6 +62,49 @@ def get_value(self, value):
     assert selenium.run_js("return pyodide.pyimport('f').toString()").startswith("<Foo")
 
 
+def test_pyproxy2(selenium):
```
rth, line 65: The test name needs to be meaningful, here and below

src/pyodide-py/pyodide/_core.py
```diff
@@ -0,0 +1,26 @@
+# type: ignore
+import platform
+
+if platform.system() == "Emscripten":
+    from _core import JsProxy, JsBoundMethod, JsException
```
rth, line 5: It should have pyodide in its import name. Otherwise it's not clear at all that `_core` has anything to do with pyodide. Maybe `_pyodide_core` ?

src/pyodide-py/pyodide/_core.py
```diff
@@ -0,0 +1,26 @@
...
+else:
+    # Can add shims here if we are so inclined.
+    class JsException(Exception):
```
rth, line 8: If it only contains exception maybe put it in `pyodide/_exceptions.py` ?

src/core/jsproxy.c
```diff
@@ -696,27 +696,39 @@ JsException_AsJs(PyObject* err)
   return hiwire_incref(js_error->js);
 }
 
+// Copied from Python 3.9
```
rth, line 699:
```suggestion
// Copied from Python 3.9
// TODO: remove once we update to Python 3.9
```
If I understand correctly?

src/tests/test_pyproxy.py
```diff
@@ -62,6 +62,49 @@ def get_value(self, value):
...
+        """
+        pyodide.runPython(`
+            class Foo:
+                pass
+            f = Foo()
+        `);
+        return pyodide.pyimport('f').toString();
+        """
+    )
+    print(result)
```
rth, line 76:
```suggestion

```

src/tests/test_pyproxy.py
```diff
@@ -62,6 +62,49 @@ def get_value(self, value):
...
+    assert result.startswith("<Foo")
+
+
+def test_pyproxy3(selenium):
+    selenium.run_js(
+        """
+        window.jspython = pyodide.pyimport("open");
+        pyodide.runPython(`
+            from js import jspython
+            jspython is open
```
rth, line 86:
```suggestion
            assert jspython is open
```
?
Is this related to this PR though? Maybe it would be better to add theses tests in a separate PR?

src/tests/test_pyproxy.py
```diff
@@ -62,6 +62,49 @@ def get_value(self, value):
...
+        """
+        window.jspython = pyodide.pyimport("open");
+        console.log("jspython:", window.jspython)
+        """
+    )
+    assert selenium.run(
+        """
+        from js import jspython
+        print(jspython, open)
+        jspython is open
```
rth, line 103:
```suggestion
        assert jspython is open
```

src/core/jsproxy.c
hoodmane: Absolutely, it'll lead to a linker error otherwise.

src/pyodide-py/pyodide/_core.py
hoodmane: `JsProxy` and `JsBoundMethod` aren't exceptions? Also I'm planning to automatically add `console` into this as well. `console.error` produces incredibly more helpful messages than `print`, even when just applied to python objects.

src/tests/test_pyproxy.py
hoodmane: Yes of course =) I swear I meant to rename them. When I first name them, I'm thinking more about whether `pytest -k whatever` will get them and the test I already had.

This test is actually just a redundant minimization of an existing test, so either this should be removed or the other test should be broken up into smaller pieces.

src/tests/test_pyproxy.py
hoodmane: Again this is a minimization of an existing test that was breaking and should be removed in any case (it did lead to PR #1098).

src/pyodide-py/pyodide/_core.py
hoodmane: Sounds good.

rth (approved):
Thanks !

PR #1093, opened 2021-01-10
PKG Add optlang package 

rudolfspetrovs:
[Optlang](https://optlang.readthedocs.io/en/latest/) is a package implementing a modeling language for solving mathematical optimization problems, i.e. maximizing or minimizing an objective function over a set of variables subject to a number of constraints.

This pull request also adds required swiglpk Python package with C library [GLPK](https://www.gnu.org/software/glpk/) (GNU Linear Programming Kit), intended for solving large-scale linear programming (LP), mixed integer programming (MIP), and other related problems.

This package is needed for COBRApy, which a package for constraint-based modeling of metabolic networks - important technique in computational biology.

### DIFF
--- a/packages/glpk/meta.yaml
+++ b/packages/glpk/meta.yaml
@@ -9,5 +9,5 @@ source:
 build:
   library: true
   script: |
-     CFLAGS="-fPIC" emconfigure ./configure  --prefix=`cwd`
+     CFLAGS="-fPIC" emconfigure ./configure  --prefix=$PWD
      emmake make -j ${PYODIDE_JOBS:-3}
--- a/packages/glpk/meta.yaml
+++ b/packages/glpk/meta.yaml
@@ -3,8 +3,8 @@ package:
   version: 4.65
 
 source:
+  sha256: 4281e29b628864dfe48d393a7bedd781e5b475387c20d8b0158f329994721a10
   url: https://ftp.gnu.org/gnu/glpk/glpk-4.65.tar.gz
-  sha256sum: 4281e29b628864dfe48d393a7bedd781e5b475387c20d8b0158f329994721a10
 
 build:
   library: true
--- a/packages/glpk/meta.yaml
+++ b/packages/glpk/meta.yaml
@@ -1,0 +1,13 @@
+package:
+  name: glpk
+  version: 4.65
+
+source:
+  url: https://ftp.gnu.org/gnu/glpk/glpk-4.65.tar.gz
+  sha256sum: 4281e29b628864dfe48d393a7bedd781e5b475387c20d8b0158f329994721a10
+
+build:
+  library: true
+  script: |
+     CFLAGS="-fPIC" emconfigure ./configure  --prefix=`cwd`
+     emmake make -j ${PYODIDE_JOBS:-3}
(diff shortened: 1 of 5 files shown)


packages/glpk/meta.yaml
```diff
@@ -0,0 +1,13 @@
...
+  version: 4.65
+
+source:
+  sha256: 4281e29b628864dfe48d393a7bedd781e5b475387c20d8b0158f329994721a10
+  url: https://ftp.gnu.org/gnu/glpk/glpk-4.65.tar.gz
+
+build:
+  library: true
+  script: |
+     CFLAGS="-fPIC" emconfigure ./configure  --prefix=`cwd`
```
dalcde, line 12:
```suggestion
     CFLAGS="-fPIC" emconfigure ./configure  --prefix=$PWD
```

packages/optlang/test_optlang.py
```diff
@@ -0,0 +1,37 @@
...
+model.objective = obj
+model.add([c1, c2, c3])
+
+status = model.optimize()
+"""
+    )
+    result = selenium.run("""model.status""")
+    assert result == "optimal"
+    result = selenium.run("""'%.4f' % model.objective.value""")
+    assert result == "733.3333"
```
rth, line 31:
```suggestion
    result = selenium.run("model.objective.value")
    assert result == pytest.approx(733.3333, abs=1e-4)
```
etc. We don't need to format as strings to assert float equality.

packages/swiglpk/test_swiglpk.py
```diff
@@ -0,0 +1,48 @@
...
+glp_simplex(lp, None);
+Z = glp_get_obj_val(lp);
+x1 = glp_get_col_prim(lp, 1);
+x2 = glp_get_col_prim(lp, 2);
+x3 = glp_get_col_prim(lp, 3);
+glp_delete_prob(lp);
+"""
+    )
+    result = selenium.run("""'Z = %g; x1 = %g; x2 = %g; x3 = %g' % (Z, x1, x2, x3)""")
+    assert result == "Z = 733.333; x1 = 33.3333; x2 = 66.6667; x3 = 0"
```
rth, line 48: This is really a usecase for `numpy.testing.assert_allclose` but numpy is not guaranteed to be installed, and we can't use it, so this is also fine.

packages/optlang/test_optlang.py
```diff
@@ -0,0 +1,37 @@
+def test_optlang(selenium):
+    selenium.load_package("optlang")
+    selenium.run(
+        """
+from optlang import Model, Variable, Constraint, Objective
```
rth, line 5: You can keep indent all this block by 8 spaces, it will still run fine and will be more readable.

packages/swiglpk/test_swiglpk.py
rudolfspetrovs: Well, I just used an example from swiglpk docs, which I modified a bit.
So I will leave it as it is

packages/optlang/test_optlang.py
rudolfspetrovs: Done in the latest commit

rth:
Thanks @rudolfspetrovs !

PR #1094, opened 2021-01-10
Follow Python include advice

hoodmane:
It says [here](https://docs.python.org/3/c-api/intro.html#include-files) that:

1. you _must_ include `Python.h` before any standard headers are included, and
2. it is recommended to always define `PY_SSIZE_T_CLEAN` before including `Python.h`.

We aren't doing either of these things. It doesn't seem to be causing any harm anywhere, but I figure it's good to follow recommendations. I adjusted the files so that `Python.h` is always the first include and before including it we always `#define PY_SSIZE_T_CLEAN`.

### DIFF
--- a/src/core/error_handling.c
+++ b/src/core/error_handling.c
@@ -1,6 +1,6 @@
 #define PY_SSIZE_T_CLEAN
-#include "Python.h"
 #include "error_handling.h"
+#include "Python.h"
 #include "jsproxy.h"
 #include <emscripten.h>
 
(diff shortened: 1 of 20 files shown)


dalcde:
I think it would be clearer to always include `Python.h` first, in case, for whatever absurd reason, e.g. `hiwire.h` stopped importing `Python.h`.

hoodmane:
Yeah, that sounds like the right idea to me.

hoodmane:
It seems that the linter prefers for `<>` includes to occur after `""` includes.

Issue #1091, opened 2021-01-10
Reexamine build dependencies


dalcde:
As far as I can tell, we no longer need build dependencies of Python, since we no longer recompile Python, but we might need some to build numpy. It would be good to trim down the list to make it easier for development

rth:
Generally recipes on conda-forge is a good way to check build dependencies. For instance, the only [build dependencies for numpy](https://github.com/conda-forge/numpy-feedstock/blob/master/recipe/meta.yaml) should be python, cython and a working C/C++ compiler.

PR #1095, opened 2021-01-10
Use python:3.8.2-slim-buster Docker base image

dalcde:
This doesn't come with any build dependencies by default, so this clarifies what really is needed (and as you can see from the diff, not much).

See also #1091


rth (approved):
Sound good thanks. We can merge after the updated image in uploaded and used in CI.

Issue #1069, opened 2021-01-07
Makefile: stale versions of pyodide_py


hoodmane:
The make recipe for `root/.built` depends on `src/pyodide-py/`, I think based on the idea that it will be rebuilt if any file in `pyodide-py` is new, then `pyodide-py` will also show that it has been modified. I thought maybe there was something wrong with my computer but I tested on two other machines I have ssh access to:
```sh
>>> mkdir test
>>> touch test/test
>>> stat test/test | grep Modify | sed "s/.*:\(.*\):.*/\1/"
26 # test/test minute modified
>>> stat test | grep Modify | sed "s/.*:\(.*\):.*/\1/" 
26 # test minute modified

>>> touch test/test
>>> stat test/test | grep Modify | sed "s/.*:\(.*\):.*/\1/"
27 # test/test minute modified changed
>>> stat test | grep Modify
26 # test minute modified is unchanged
```
This is quite a nuisance since I keep compiling with stale versions of pyodide-py.

dalcde:
I think in general we should try to avoid using /.* files in the
Makefile

hoodmane:
What does that mean? Are you saying that the best practice is to manually `touch src/pyodide-py` whenever I update `pyodide-py`? I thought the makefile and OS were supposed to cooperate to work out which files were stale.

dalcde:
Nah I'm saying generating root/.built is not a good way to let Makefile
work out which files are stale, which is not exactly relevant to this
issue.

hoodmane:
Okay, I don't know what that means. Thanks for the fix!

Issue #646, opened 2020-05-09
Optimize the size of minimal pyodide build


rth:
I have been reading https://pythontips.com/2019/05/22/running-python-in-the-browser/ which does a nice comparison on tools to run python in the browser. The main downside mentioned for pyodide there was the size of downloads. While optimizing the size of numpy, pandas etc is probably more challenging, I think we can reduce the size of downloads in the case of a minimal pyodide build with just cpython, pyodide tools and micropip.

I would propose to introduce en optional env variable `PYODIDE_MINIMAL=true` that together with partial builds in https://github.com/iodide-project/pyodide/pull/645 could be used as follows,
```
PYODIDE_MINIMAL=true PYODIDE_PACKAGES="micropip" make
```

This would,
 - [ ] build with [`USE_FREETYPE=0`](https://github.com/iodide-project/pyodide/blob/ea74555209dab0f31ae36751f8ec2978caa839ea/Makefile#L40) (it is statically linked as far as I understand? `libfreetype.a` is 1.1MB), `USE_PNG=0` (libpng.a is 350k).
 - [ ] remove some of the left over test folders in stdlib,
   ```
   Lib/distutils/tests
   Lib/sqlite3/test
   Lib/unittest/test
   ```
   that are now up to 800 kB in size. The main `Lib/test` (30MB) is already [packaged as a separate python module](https://github.com/iodide-project/pyodide/blob/ea74555209dab0f31ae36751f8ec2978caa839ea/Makefile#L161). for these I think we can probably just remove them when `PYODIDE_MINIMAL=true` and keep running these tests in CI otherwise.
 - [ ] re-evaluate if some of the other larger and exotic stdlib modules can be removed with [remove_modules.txt](https://github.com/iodide-project/pyodide/blob/ea74555209dab0f31ae36751f8ec2978caa839ea/remove_modules.txt). For instance [pydoc](https://docs.python.org/3/library/pydoc.html) and pydoc_data are around 1MB in size. There are some .exe files in `Lib/distutils/command/` for ~2MB (not sure if they are actually included in pyodide).
 - [ ] re-evaluate if some of the modules build in the main Makefile and included by default are still useful in this use-case. For instance, jeudi + parso is 6MB.
 - [ ] on the countrary I would have included `micropip` by default instead of a making it a standalone package.

I'm not sure how these sizes would map to the size of compressed pyodide package, but I'm hoping we can improve the current situation which is,
```
[...]

PR #1096, opened 2021-01-10
Don't copy artifacts to root

dalcde:
Load them directly with --preload-file. Avoids issues such as #1069

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -19,8 +19,6 @@ host
 installs
 ccache
 
-/root/
-
 /emsdk/emsdk
 
 *.egg-info/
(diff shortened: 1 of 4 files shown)


rth:
If you check the `check-size` setup in the build-core CI job,
```
-rw-r--r-- 1 root root  51M Jan 10 03:37 pyodide.asm.data
```
this is currently clearly including some folders that it shouldn't. Before it was,
```
-rw-r--r-- 1 root root 5.5M Jan 10 09:22 pyodide.asm.data
```

Otherwise the idea sounds good.

dalcde:
If I didn't mess this up this should remove e.g. sqlite tests as well (#646)

dalcde:
It's now 5.2 MB

rth:
> It's now 5.2 MB

Great!

rth:
Actually this might remove tests that we don't properly extract in the test package. See second point from https://github.com/iodide-project/pyodide/issues/646 but as far as I can tell we are not running them anyway (the number of tests run here and on master is the same), so this change should be OK.

PR #1097, opened 2021-01-10
List failed tests

dalcde:


dalcde:
Flaky tests exactly when I need them

rth (commented):
You could add it to the default pytest option in `setup.cfg`. We already have `-r sxX` there. In particular I don't think we should hardcode it in tools/pytest_wrapper.py.

dalcde:
Ah I see. We could add -v there too. Do we want to print skipped and xfailed/xpassed ones? I don't think it is very useful.

rth:
> Do we want to print skipped and xfailed/xpassed ones? I don't think it is very useful.

Yes, we can remove them. The idea was to detect CPython tests that used to fail but now pass, however, we need to run pytest with `--run-xfail` (our custom option) to actually execute xfail tests to the end anyway (otherwise they pytest.xfail is raised and the rest of the test is not executed).

PR #1098, opened 2021-01-10
Make builtins behavior consistent with native python interpreter

hoodmane:
In native Python if you `del` a builtin it `NameError`s. In Pyodide, if you `del` a builtin, it doesn't name error the first time, but `NameError`s the second time:
javascript:
```javascript
let open = pyodide.pyimport("open");  // works
```
python:
```python
del open # this raises no error
```
javascript:
```javascript
let open2 = pyodide.pyimport("open");  // returns undefined now
```
python:
```python
open # still no name error, returns the same object as before we del'd it
del open # raises NameError the second time we del it.
```
This PR makes the behavior consistent with native python: `del open` raises a `NameError` the first time, and `pyimport("open")` works fine no matter how many times you try `del open`.


src/tests/test_python.py
```diff
@@ -169,6 +169,19 @@ def test_unknown_attribute(selenium):
     )
 
 
+def test_del_builtin(selenium):
+    msg = "NameError"
+    with pytest.raises(selenium_standalone.JavascriptException, match=msg):
+        # can't del a builtin
+        selenium.run(
+            """
+            del globals
```
dalcde, line 178:
```suggestion
            del open
```

hoodmane: Thanks.

hoodmane:
There's a really weird bug here:
```
pyodide.runPython(`
   import numpy as np
   x = np.array([['string1', 'string2'], ['string3', 'string4']])
`);
pyodide.globals.x // Works
pyodide.globals.x // Invalid type to convert to array buffer.
pyodide.globals.x // Works
pyodide.globals.x // Invalid type to convert to array buffer.
pyodide.globals.x // Works
pyodide.globals.x // Invalid type to convert to array buffer.
...
```
this alternates forever. I think this is a latent bug somewhere else in the code since I've gotten this same failure due to several unrelated changes.

hoodmane:
This is again running into the #768 / #788 bug complex.

hoodmane:
@dalcde The test suite shows:
```
response = {'status': 500, 'value': '{
   "value": { 
      "error" : "unknown error",
      "message" : "Error: TypeError: attribute name must be string...e.get_value@http://127.0.0.1:54147/pyodide.asm.js:8:169831\\nget@http://127.0.0.1:54147/pyodide.asm.js:8:179927\\n"
   }
}'}`
```
The `message` field has a truncated stack trace. If I could see this stack trace I would most likely be able to fix the test failure. Any advice?

dalcde:
Looking at the `run_js` code, the only place where the exception will not be caught is in the final catch block, so I suspect something might be going wrong in the `e.toString()` call (or `e.stack`). I would try wrapping that in a further catch block and/or printing `e.stack` to stdout before computing `e.toString()` and returning.

hoodmane:
I'm pretty sure the error is in `e.toString()`.

hoodmane:
So I added `print(retval)`:
```python
retval = self.driver.execute_script(wrapper % code)
print(retval) # <--- here
```
and it prints "[0, None]`. I think the error is happening when it tries to serialize `threading.locals()` to send back to the server. It fails to serialize so somehow that fails the test but outputs `None`.

hoodmane:
@dalcde You have any idea what's going wrong with the benchmarks here?

dalcde:
The following snippet works on master (and normal python) but not on your branch:
```python
def foo():
    pass
from __main__ import foo
```

hoodmane:
Right, okay, I'm supposed to be mimicking `PyRun_SimpleStringFlags`, where they do:
```C
    m = PyImport_AddModule("__main__");
    if (m == NULL)
        return -1;
    d = PyModule_GetDict(m);
    v = PyRun_StringFlags(command, Py_file_input, d, d, flags);
```
where they use `__main__.__dict__` as both the globals and the locals.

hoodmane:
@dalcde @rth @phorward Anyone willing to review this?

PR #1099, opened 2021-01-10
Fix SHA256 hash checking

rudolfspetrovs:
An incorrect (old?) key was used in meta.yaml, they do not seem to fail on the incorrect hash.

### DIFF
--- a/packages/CLAPACK/meta.yaml
+++ b/packages/CLAPACK/meta.yaml
@@ -3,9 +3,9 @@ package:
   version: 3.2.1
 
 source:
+  sha256: 6dc4c382164beec8aaed8fd2acc36ad24232c406eda6db462bd4c41d5e455fac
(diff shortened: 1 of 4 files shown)


dalcde:
Which key was incorrect?

rudolfspetrovs:
sha256sum was used instead of sha256

dalcde:
Oh I thought you meant the checksum was incorrect

stupendous-wave:
No, I discovered that if you use "sha256sum" in meta.yaml, then build (hash check) does not fail on incorrect hash
I guess it checks for "sha256" key and does not check hash if it is not present. 
Documentation **does** say to use "sha256" key, so I decided to check if it is synonymous.

dalcde:
It's definitely my fault for using the wrong key. Perhaps we should check for unrecognized keys.

dalcde:
Thanks!

PR #1102, opened 2021-01-10
Emscripten 2.0.12 Upstream backend

### LINKED ISSUE #476: Migrate to upstream LLVM for emscripten backend
It looks like the now recommended compiler backend for emscripten is the upstream LLVM instead of fastcomp + asm2wasm (https://github.com/WebAssembly/binaryen/issues/2177#issuecomment-504269479). 

This might also help with scipy dynamic linking, as the section of code that errors now in https://github.com/iodide-project/pyodide/pull/473 will no longer be run.

~~There is no tagged emsdk release with it yet, I think, but hopefully, one be released soon.~~

I don't have a very good understanding of how much work this would involve. I'm hoping existing emsdk patches would still apply. Maybe it would be necessary to update to the latest emsdk release first.

joemarshall:
What it says. Emscripten 2.0.9 upstream (non-fastcomp, compiles straight to wasm), plus patches for binaryen and emscripten so that dynamic linking and passing of function pointers between side modules and main modules works.

I'm part way through pull requests for binaryen and emscripten so that the patches there aren't needed.

All other patches to emscripten and binaryen have been dropped - I'm not sure if they were all things that are fixed in the new backend - none of them worked with new backend anyway

Closes #531
Closes #476


joemarshall:
I need to fix the tests, will do it later if I have a chance

Makefile
```diff
@@ -31,12 +31,12 @@ LDFLAGS=\
 	-s WASM=1 \
 	-s USE_FREETYPE=1 \
 	-s USE_LIBPNG=1 \
+    -s LZ4=1 \
```
rth, line 34: It's already present below.

Makefile.envs
```diff
@@ -1,5 +1,5 @@
-export EMSCRIPTEN_VERSION = 1.38.44
-export BINARYEN_VERSION = version_89
+export EMSCRIPTEN_VERSION = 2.0.9
+export BINARYEN_VERSION = a8ded16f56afd880a9a6459fe5ce55a8667d9b3e
```
rth, line 2: Is this master? Should we use the latest release instead (version_99) ?

emsdk/patches/fix_emulate_pointers_dlsym.patch
```diff
@@ -0,0 +1,62 @@
+diff --git a/src/passes/FuncCastEmulation.cpp b/src/passes/FuncCastEmulation.cpp
+index 9d232c9..35d6d68 100644
+--- a/emsdk/binaryen/src/passes/FuncCastEmulation.cpp
++++ b/emsdk/binaryen/src/passes/FuncCastEmulation.cpp
+@@ -167,13 +167,14 @@ private:
+ struct FuncCastEmulation : public Pass {
+   void run(PassRunner* runner, Module* module) override {
+     Index numParams =
+-      std::stoul(runner->options.getArgumentOrDefault("max-func-params", "16"));
++      std::stoul(runner->options.getArgumentOrDefault("max-func-params", "61"));
```
rth, line 10: It should be possible to pass this as a param, since https://github.com/WebAssembly/binaryen/pull/3411 included in version_99

pyodide_build/pywasmcross.py
```diff
@@ -44,6 +44,13 @@
 symlinks = set(["cc", "c++", "ld", "ar", "gcc", "gfortran"])
 
 
+class EnvironmentRewritingArgument(argparse.Action):
+    def __call__(self, parser, namespace, values, option_string=None):
+        for e_name, e_value in os.environ.items():
+            values = values.replace(f"$({e_name})", e_value)
+        setattr(namespace, self.dest, values)
```
rth, line 51: Could you please add some tests (outside of pyodide) for this class? in pyodide_built/tests/test_pywasmcross.py

Makefile
dalcde, line 34: We already have this below

pyodide_build/pywasmcross.py
```diff
@@ -44,6 +44,13 @@
 symlinks = set(["cc", "c++", "ld", "ar", "gcc", "gfortran"])
 
 
+class EnvironmentRewritingArgument(argparse.Action):
+    def __call__(self, parser, namespace, values, option_string=None):
```
rth, line 48: Some type annotations would help readability.

pyodide_build/pywasmcross.py
```diff
@@ -270,9 +279,19 @@ def handle_command(line, args, dryrun=False):
...
+            # skip second one
+            if arg in used_libs:
+                continue
+            used_libs[arg] = 1
+        if arg.startswith("-l"):
+            for lib_name in replace_libs.keys():
+                # this enables glob style **/* matching
+                if PurePosixPath(arg[2:]).match(lib_name):
+                    if len(replace_libs[lib_name]) > 0:
+                        arg = "-l" + replace_libs[lib_name]
```
rth, line 292: Could you please add tests for this.

Makefile.envs
dalcde, line 2: emscripten 2.0.12 and binaryen version_99 have recently been released, which covers the changes we need. In particular, it incorporates the packager patch, which you removed

pyodide_build/pywasmcross.py
```diff
@@ -332,6 +355,10 @@ def handle_command(line, args, dryrun=False):
         # See https://github.com/emscripten-core/emscripten/issues/8650
         if arg in ["-lfreetype", "-lz", "-lpng", "-lgfortran"]:
             continue
+        # don't use -shared, SIDE_MODULE is already used
+        # and -shared breaks it
+        if arg in ["-shared"]:
+            continue
```
rth, line 361: Adding a test for it also wouldn't hurt.

emsdk/Makefile
```diff
@@ -8,13 +8,14 @@ all: emsdk/.complete
 emsdk/.complete: ../Makefile.envs $(wildcard patches/*.patch)
 	if [ -d emsdk ]; then rm -rf emsdk; fi
 	git clone --depth 1 https://github.com/emscripten-core/emsdk.git
-	cd emsdk && ./emsdk install --build=Release $(EMSCRIPTEN_VERSION)-fastcomp
-	git clone --branch $(BINARYEN_VERSION) --depth 1 https://github.com/WebAssembly/binaryen.git emsdk/binaryen
+	cd emsdk && ./emsdk install --build=Release $(EMSCRIPTEN_VERSION)
+	git clone --depth 1 https://github.com/WebAssembly/binaryen.git emsdk/binaryen
+	cd emsdk/binaryen && git fetch origin $(BINARYEN_VERSION) && git checkout $(BINARYEN_VERSION)
```
dalcde, line 13: Can we not keep the original `git clone --branch $(BINARYEN_VERSION)`?

src/pyodide.js
```diff
@@ -377,9 +377,12 @@ var languagePluginLoader = new Promise((resolve, reject) => {
...
     // filesystem to install itself. Once that's complete, it will be replaced
     // by the call to `makePublicAPI` with a more limited public API.
-    self.pyodide = pyodide(Module);
-    self.pyodide.loadedPackages = {};
-    self.pyodide.loadPackage = loadPackage;
+    pyodide(Module).then((x) => {
+      self.pyodide = x;
+      self.pyodide.loadedPackages = {};
+      self.pyodide.loadPackage = loadPackage;
+      self.LZ4 = Module['LZ4'];
```
rth, line 384: Please add an inline comment why it's needed.

emsdk/patches/fix_emscripten_dylink.patch
```diff
@@ -0,0 +1,75 @@
+diff --git  a/emsdk/upstream/emscripten/src/library_dylink.js b/emsdk/upstream/emscripten/src/library_dylink.js
+index f8ddec2..a5e8682 100644
+--- a/emsdk/upstream/emscripten/src/library_dylink.js
++++ b/emsdk/upstream/emscripten/src/library_dylink.js
+@@ -57,6 +57,7 @@ var LibraryDylink = {
+         err("new GOT entry: " + symName);
+ #endif
+       }
++
```
dalcde, line 9: This diff can be removed

rth (commented):
Thank you @joemarshall, really looking forward to this!

emsdk/patches/fix_emulate_pointers_dlsym.patch
dalcde, line 10: This can be achieved by passing `-s BINARYEN_EXTRA_PASSES="--pass-arg=max-func-params@61"` as an ldflag

src/pyodide.js
dalcde, line 384: This line should not be needed with the latest emscripten

pyodide_build/pywasmcross.py
```diff
@@ -270,9 +279,19 @@ def handle_command(line, args, dryrun=False):
...
-
+            # WASM link doesn't like libraries being included twice
+            # skip second one
+            if arg in used_libs:
+                continue
+            used_libs[arg] = 1
+        if arg.startswith("-l"):
+            for lib_name in replace_libs.keys():
+                # this enables glob style **/* matching
+                if PurePosixPath(arg[2:]).match(lib_name):
```
dalcde, line 290: This is a useful change but I don't think it is required for the emscripten upgrade?

pyodide_build/pywasmcross.py
```diff
@@ -255,6 +262,8 @@ def handle_command(line, args, dryrun=False):
 
     lapack_dir = None
 
+    used_libs = {}
```
dalcde, line 265: I think it would be more semantically correct to use a Set instead of an object

dalcde (commented):
Thanks a lot! I'm really excited about this. I've left some minor comments

pyodide_build/pywasmcross.py
dalcde: This is already tested for (and the tests currently fail because they haven't been updated)

dalcde:
We should try not reviewing simultaneously :). Glad that we mostly caught distinct things though.

joemarshall:
Current status on this - I've commited some of the changes suggested above. In doing so I had to move to emscripten 2.0.12 to get that packager fix in. It needed some more fixes to emscripten dynamic linking to make the tests there work (plus some modifying of the fpcast test so that it actually tried calling something with the wrong signature!). Right now there's a bunch of tests in src that I need to work through to see what is broken. Updating seems to have broken python output (at least from print in console.html), I suspect this is causing a bunch of the fails.

joemarshall:
Fixed a bunch of tests now, but there's some stuff in scipy that I think is to do with yet more fortran return types being inconsistent which causes some test fails when sub modules don't import. This also breaks astropy and some things that use scipy submodules.

Also, wasm in new emscripten breaks sometimes on access to uninitialised variables I discover, whereas it appears asm.js initializes to zero. There was one in the jsproxy that caused all sort of weird intermittent errors.

rth:
> yet more fortran return types being inconsistent which causes some test fails when sub modules don't import

BTW, apparently there are a number of known issues around it https://github.com/iodide-project/pyodide/issues/72#issuecomment-504728763

>  There was one in the jsproxy that caused all sort of weird intermittent errors.

Was or is? As there are lots of core (and package) test are still failing with `RuntimeError: unreachable executed`. It would be great to make those work first. Maybe @hoodmane would have some insights on debugging jsproxy related issues..

Could you also please merge upstream master in to resolve conflicts?

joemarshall:
Oops, didn't push the unreachable thing, which was that uninitialized variable in jsproxy. 

There's still weirdness with function parameters to lapack code in scipy which I am debugging by building it with function parameter and implicit function definition warnings set to error

hoodmane:
Looking into this, at a first glance these all seem to be assertion errors fairly deep inside of CPython. I took a look at the core-chrome tests.
 
Most failures (380) are coming from `pyproxy_apply` ==> `PyObject_VectorCall` ==> "function signature mismatch". These could be issues related to `pyproxy`, but I would suspect something about the vectorcall protocol might be amiss since we see so many of these failures. `VectorCall` looks up a function pointer based on an offset stored in the PyType struct. Does "function signature mismatch" mean it actually found a function pointer, or is it possible that it didn't even get a correct function pointer at all. 

In 5 cases it fails inside of the `ceval.c` method `import_all_from` which is the lowest layer of the import mechanism. In one case it fails at 
```
from js import TEST
from js import a
import micropip (x3)
```
I think this is happening at `_PyObject_LookupAttrId(v, &PyId___dict__, &dict)` on line 5269 of `ceval.c`. There are a couple of cases while executing the `LOAD_ATTR` opcode (2966 of ceval).

3 times it fails in `___sys_wait4` with "abort(cannot wait on child processes)", that may need a patch?

hoodmane:
Anyways, there's a clear first thing to look into since `VectorCall` failures account for the vast majority of problems. I don't think most of these tests are run in `selenium_standalone` so it is plausible that the other failures are all caused by state corruption that occurred in the `VectorCall` panics.

joemarshall:
I'll try to reproduce one of those errors in browser, there's probably something similar to the proxy error that I pushed a patch for, some kind of assumption that function types don't matter because they don't in C, or that things are initialized to zero...

hoodmane:
In the firefox version, it gives unreadable stack traces (no function name symbols) so you can't tell that it's inside of `VectorCall` but the errors read `RuntimeError: indirect call to null`, so that suggests that whatever the vectorcall function pointers are actually all `NULL` somehow? Or is it possible that `indirect call to null` is just what Firefox says when any bad function call happens.

joemarshall:
Aha. I think this is a binaryen/emscripten thing - what's going on is that something somewhere is making a function pointer that doesn't point to the binaryen version (with the 61 arguments), whereas binaryen rewrites all indirect calls to call with that signature. I suspect what is happening here is that while I rewrite the function table in emscripten for things that are imported into the main module, it also needs the same rewriting for the main module itself. It only exhibits it when python modules are imported that are already loaded when the main module starts I think. 

Good news is, by copying python tests into the root folder, I have this reproducing in a browser window with debugging, which makes things a whole lot easier. I think it will hopefully just be a matter of rewriting the main module table on load in the JS.

Weirdest thing about this is how hard it was to reproduce - I guess most python module calls of this type are to things that are included as modules themselves so go through the binaryen logic properly already... Thank goodness for  regression tests.

hoodmane:
> a browser window with debugging

What's your setup for that? You have any debugging tips? (I always just use print statements.)

joemarshall:
The browser window with debugging, you need to build with -g4 (you can do export EMMAKEN_CFLAGS=-g4 rather than messing with the Makefile), and then it makes source maps on compile. 

The source maps are usually broken in that they have different base folders depending on how you built them. You can either fix this by doing some hacking of build parameters, or use a custom python web server that just fixes everything up for you like this very minor modification of the example pyodide_serve python script:

```
alternative_bases=["cpython/build/Python-3.8.2/","src/"]

class Handler(SimpleHTTPRequestHandler):
    def do_GET(self):
        # to make source maps work, if we look for a file and
        # are about to return 404, then check if
        # it exists on alternative base paths
        if not os.path.exists(self.path):
            if len(self.path)==0:
                self.path="/"
            for base in alternative_bases:
                if os.path.exists(os.path.join(base,self.path[1:])):
                    print(f"Redirect: {self.path} -> {base}")
                    self.path=os.path.join(base,self.path[1:])
                    break
        super().do_GET()


    def end_headers(self):
        # Enable Cross-Origin Resource Sharing (CORS)
        self.send_header('Access-Control-Allow-Origin', '*')
        super().end_headers()
```

hoodmane:
Thanks! I'm excited to try this.

hoodmane:
> Also, wasm in new emscripten breaks sometimes on access to uninitialised variables I discover, whereas it appears asm.js initializes to zero. There was one in the jsproxy that caused all sort of weird intermittent errors.

Sorry I didn't read this part of the thread earlier. I've been trying to fix these things systematically (initialize all variables to `NULL` on declaration!) but if you know specifically where this bug is I'd be happy to hear about it.

joemarshall:
Ah, the big bad thing was that in the tests it calls os.waitpid. That calls through to a syscall wait4 which is not implemented in emscripten and is set to abort the runtime. Aborting the runtime leaves it in a funny state which is why that call goes bad.

I tried telling python build that we don't have waitpid, but that causes problems with a bunch of things which expect it to exist if we're on anything posix like. So right now I've made pyodide.js override waitpid and return (0,0) from it, along with a console warning. This seems to fix tests on my machine at least.

I still don't 100% understand why the tests work on old emscripten? Maybe there is something I'm missing in the cpython build for asmjs that means this doesn't die? Maybe the library in fastcomp is different and doesn't abort on that syscall?

dalcde:
I would compare the ./configure outputs and see if they did anything different

joemarshall:
Aha, 

emscripten has changed so it doesn't pull things from the javascript library*.js files if they are in the libc library. waitpid is implemented in the libc library. But that calls through to syscall wait4, which is defined as abort. I've put a patch in to make it do the same as the javascript override for waitpid (return the exception 'no child processes'.) which appears to fix those test fails.

There's still a few wrong function prototype things in scipy to fix, but core python I think is worky.

joemarshall:
Fixed astropy build and scipy I think has consistent signatures now.

Couple of tests etc. left to fix still.

cpython/Makefile
```diff
@@ -109,16 +109,16 @@ $(BUILD)/Makefile: $(BUILD)/.patched $(ZLIBBUILD)/.configured $(SQLITEBUILD)/lib
 	cp config.site $(BUILD)/
 	( \
 		cd $(BUILD); \
-		EMCC_CFLAGS="-s WASM=0" CONFIG_SITE=./config.site READELF=true LD_RUN_PATH="$(SQLITEBUILD):$(BZIP2BUILD)" emconfigure \
+		 CONFIG_SITE=./config.site READELF=true LD_RUN_PATH="$(SQLITEBUILD):$(BZIP2BUILD)" emconfigure \
 		  ./configure \
-			  CFLAGS="-fPIC" \
+			  CFLAGS="-fPIC -sEMULATE_FUNCTION_POINTER_CASTS=1" \
```
dalcde, line 114: I don't think you need `EMULATE_FUNCTION_POINTER_CASTS` in `CFLAGS`.

hoodmane:
The one core test that's still failing now is the test's fault.

cpython/Makefile
dalcde: (but see also #1137)

src/pyodide.js
```diff
@@ -461,6 +461,13 @@ globalThis.languagePluginLoader = new Promise((resolve, reject) => {
     let response = await fetch(`${baseURL}packages.json`);
     let json = await response.json();
     fixRecursionLimit(self.pyodide);
+    // lots of things in tests try to wait for processes to end using
+    // os.waitpid. This breaks because emscripten doesn't support processes (and
+    // in recent emscripten this syscall actually calls abort on the wasm
+    // runtime so most tests break.)
+    pyodide.runPython(
+        "import os\ndef _wait_pid_nop(x,y):\n print('WARNING: wait_pid is a NOP in emscripten')\n return (0,0)\nos.waitpid=_wait_pid_nop\n");
+
```
dalcde, line 470: Can this be removed given the emscripten patch?

src/templates/console.html
```diff
@@ -1,4 +1,4 @@
-<!doctype html>
+    <!doctype html>
```
hoodmane, line 1: Actually this is why the test fails.

dalcde:
You accidentally included some `.swp` files into scipy's make-int-return-values patch. The patch itself should already be in master.

joemarshall:
Wow. Scipy.linalg now imports. That took quite a lot of messing with the f2py wrappers. Basically scipy is built to assume that everything is cool with symbols being defined multiple times in multiple places, and makes some quite specific assumptions about how fortran functions are called from c, which don't hold for clapack. I changed the f2py wrappers to call clapack directly, rather than wrapping things in weird fortran calling conventions.

On the plus side, I have learnt a lot more about debugging webassembly! One big thing to know is that if symbols are multiply defined or there's an extern def with different arguments but the same name somewhere, wasm will just silently add an empty function export called $signature_mismatch:NAME, and bad things will happen, as some calls will just fail, some function pointers will be null etc... Fortunately you can easily spot them in chrome devtools. I suspect there's a need for some tests which call wasm-objdump somehow on packages to spot problems like this automatically.

joemarshall:
darn, release build vs debug build problems... Needed sorting out the clapack patches again, what I commited now should work in release as well as debug.

rth:
Thanks so much for working on this @joemarshall ! I'm very much looking forward to having this in master :)

It certainly doesn't look simple on the scipy / CLAPACK interface..

packages/scipy/patches/make-int-return-values.patch
```diff
@@ -1,37 +1,71 @@
...
-the lapack source files, apply _cython_signature_generators.py to get some .txt
-files, then apply _cython_wrapper_generators.py to generate the header files
-from the .txt files.
-
-To avoid having to fetch the original Fortran lapack source files, we only
-patch _cython_wrapper_generators.py and tell it to replace every instance of
-void with int. We then regenerate header files from the vendored (and
-incorrect) .txt files.
-
-Of course, there is still non-generated code that needs to be patched by hand.
```
rth, line 20: We would still need some description to this patch.

packages/scipy/patches/make-int-return-values.patch
```diff
@@ -268,3 +1499,50 @@ Of course, there is still non-generated code that needs to be patched by hand.
...
++int dgetrf_(int *m, int *n, double *a, int *lda, int *ipiv, int *info);
++int dgecon_(char *norm, int *n, double *a, int *lda, double *anorm, double *rcond, double *work, int *iwork, int *info);
++
+ #define qh_dgetrf F_FUNC(dgetrf,DGETRF)
+ #define qh_dgecon F_FUNC(dgecon,DGECON)
+ #define qh_dgetrs F_FUNC(dgetrs,DGETRS)
+diff --git a/scipy/special/amos/.master.gz b/before/scipy/special/amos/.master.gz
+deleted file mode 100644
+index 8cd34d4..0000000
+--- a/scipy/special/amos/.master.gz
```
rth, line 1520: Is it really necessary to delete it?

rth (commented):
If you look at the "Files changed" tab on Github, there are still a few review comments without a response.

Also could you please merge master in again? Thanks!

pyodide_build/pywasmcross.py
joemarshall: Found out about monkeypatch and done this now. Learn a new thing every day!

emsdk/tests/test_emulate.py
```diff
@@ -9,7 +9,9 @@ def test_emulate_function(tmpdir):
                 """\
 #include <stdio.h>
 
-void foo() {
+// emulate function pointer casts - this has the
+// wrong arguments and return type
+int foo(int extra_args) {
   puts("hello from library");
 }"""
```
dalcde, line 16:
```suggestion
int foo(int extra_args) {
  puts("hello from library");
  return 0;
}"""
```

emsdk/tests/test_emulate.py
```diff
@@ -21,6 +21,8 @@ def test_emulate_function(tmpdir):
         subprocess.run(
             [
                 "emcc",
+                "-s"
+                'BINARYEN_EXTRA_PASSES="--pass-arg=max-func-params@61"',
```
dalcde, line 26: Why are these necessary for the test to pass? I think it should do fine without it

emsdk/tests/test_emulate.py
```diff
@@ -22,8 +22,7 @@ def test_emulate_function(tmpdir):
         subprocess.run(
             [
                 "emcc",
-                "-s"
-                'BINARYEN_EXTRA_PASSES="--pass-arg=max-func-params@61"',
+                "-s" 'BINARYEN_EXTRA_PASSES="--pass-arg=max-func-params@61"',
```
dalcde, line 25: You would want a comma between them; This currently gives string concatenation (but as I mentioned previously I think the entire line can be omitted; otherwise something is pretty wrong with the emsdk patch)

emsdk/tests/test_emulate.py
joemarshall: Yeah, just left in from earlier testing. Will remove.

emsdk/tests/test_dyncall.py
```diff
@@ -20,19 +20,10 @@ def test_dyncall(tmpdir):
...
   type_iifid fp = &indirect_function;
-
-  jmp_buf buf;
-  int i = setjmp(buf);
-
+  int i=5;
   printf("%d\\n", i);
   assert(fp(i, 0, 0, 0) == i);
-
-  if (i == 0) longjmp(buf, 1);
```
dalcde, line 35: Can you leave the original test in some form? This was originally included to fix a problem that is triggered specifically with setjmp/longjmp. I don't think the bug is still present, but it manifest itself in one of the last `scipy` tests, so it would be nice to be able to catch it earlier on.

joemarshall: darn, I wondered why that was in there. The test works if the main.c has setjmp and longjmp, otherwise setjmp doesn't get linked in with new wasm thanks to the pretty hardcore dead-code elimination. I'll mess with the common / main.c so it works then.

rth:
Thanks for your persistence and your work on this @joemarshall ! I see that aside for matploltib all tests are now passing! :tada:

I think we could start to merge chunks of this in separate PRs that could be done independently, to reduce the diff here somewhat and make it easier to review. I can help with that unless someone else wants to do it (while preserving your commit attribution).

For increasing the `TOTAL_MEMORY` it's indeed a good idea and as side effect it might fix https://github.com/iodide-project/pyodide/issues/721#issuecomment-764516491

Once it is merged it would be great to upstream some the fixes for linking with EMULATE_FUNCTION_POINTER_CASTS.

joemarshall:
The emulate function pointer casts stuff, I'm talking to binaryen people about how best to do that  Currently, there's a bunch of ways to get function pointers, and they return different things depending on what you call, and whether you call them from a side module or the main module. There's a fair bit of fiddling needed in binaryen and emscripten to catch all the edge cases.

Invoke calls are a pain particularly because they currently expect an unmodified call signature, and for static methods there's no way to map from fpcast to unmodified calls. 

I think the ideal would be if everything consistently used the fpcast versions of things. That would be more easy if pyodide built with bigint support, as then JavaScript can call fpcast signature functions, which have int64 on, but that does mean losing support for older browsers.

It also occurs to me that since I patched various packages to have consistent signatures, I don't actually know whether the whole thing would build without fpcast now... Would be worth a try building the whole thing with warning=error on function signature checks. Scipy and lapack build like that now anyway.

hoodmane:
> That would be more easy if pyodide built with bigint support, as then JavaScript can call fpcast signature functions, which have int64 on, but that does mean losing support for older browsers.

Apparently 86.7% of internet users have bigint support, and I suspect that there are likely other issues with pyodide for many of the remaining 13%.
https://caniuse.com/bigint

> I don't actually know whether the whole thing would build without fpcast now... Would be worth a try building the whole thing with warning=error on function signature checks

PR #1169 removes the pointer casts that appear in our `core` code: as explained there, a `METH_NOARGS` function should have signature `PyObject* f(PyObject*, PyObject*)` but the second argument is always `NULL`. Without that PR, we declare the `METH_NOARGS` functions as `PyObject* f(JsProxy*)` and cast them to `PyObject* f(PyObject*, PyObject*)`.

dalcde:
Glad to see a build with full CI passing! Great work!

dalcde:
I'll drop the observation that at times, the chrome tests run way slower than firefox. Not sure if it is consistent.

On the other hand, build times seem to have significantly improved, apart from emsdk taking longer in general. I'm not too concerned about the binaryen build, as the changes can (hopefully) be upstreamed and we would not have to build it ourselves anymore. In any case, there is not much we can do about it anyway...

joemarshall:
It is weird the tests running slow in chrome, it seems pretty sporadic. 

I've just removed some xfails for packages that need tests to run for chrome now (scipy, scikit-image etc.). It appears to work now, although the chrome tests sometimes time out and I can't work out why. 

I think it is pretty much ready for splitting up into multiple pull requests for the various different things. 

I'd suggest spitting by:
1) everything in scipy, numpy & Clapack - these should merge in without breaking old pyodide, and are quite big. They can't separate down any more than that as they're all interdependent.
2) other package fixes - same, they shouldn't break things
3) makefile / pyodide.js / emsdk / pyodide_build changes - I think any of these will break everything else if they are merged separately)

I'm fine with someone else doing the split, I am pretty sure you will be better at hacking git than me. Alternatively I don't mind doing it but I'd probably need a load of hand holding, and I think I'm unlikely to get onto it until end of the week, I've got a bunch of teaching to prep (some of it using pyodide which is fun...).

rth:
Thanks that sounds good. We merged some of the changes in the above linked PRs, which reduces the diff by 1kLoC here. I just merged master in to sync. Will look into clapack and f2c related changes next.

emsdk/tests/test_dyncall.py
```diff
@@ -42,8 +43,10 @@ def test_dyncall(tmpdir):
         subprocess.run(
             [
                 "emcc",
+                "-g4",
                 "-s",
                 "SIDE_MODULE=1",
+                "-sWASM_BIGINT",
```
dalcde, line 49: Do we want to keep these flag changes?

packages/pillow/meta.yaml
```diff
@@ -12,8 +12,8 @@ source:
       - src/setup.cfg
       - ./setup.cfg
 build:
-  cflags: -s USE_ZLIB=1 -s USE_LIBJPEG=1 -s USE_FREETYPE=1
-  ldflags: -L ../../../../emsdk/emsdk/fastcomp/emscripten/cache/asmjs/ ../../../../emsdk/emsdk/fastcomp/emscripten/cache/asmjs/libjpeg.bc
+  cflags: -s USE_ZLIB=1 -s USE_LIBJPEG=1 -s USE_FREETYPE=1 -fPIC
```
dalcde, line 15: This is already in the default fclags
```suggestion
  cflags: -s USE_ZLIB=1 -s USE_LIBJPEG=1 -s USE_FREETYPE=1
```

packages/test_common.py
```diff
@@ -68,34 +68,13 @@ def test_import(name, selenium_standalone):
...
-        ))
-        """
-        )
-        == baseline_pyc
-    )
-
+    # instead of loading packages using pyodide.loadPackage, this test now
+    # tests loading packages just through calling import, as that should
+    # call loadPackage internally, and is more representative of how
+    # python actually does it
```
dalcde, line 74: This explains the diff rather than the code
```suggestion
```

emsdk/tests/test_dyncall.py
rth: More [debugging flags](https://emscripten.org/docs/porting/Debugging.html#debug-information) for tests shouldn't hurt. Not too sure about `WASM_BIGINT` but it might be good to build everything with it anyway in the future.

src/pyodide.js
```diff
@@ -372,7 +373,7 @@ globalThis.languagePluginLoader = new Promise((resolve, reject) => {
     let packages = new Set();
     for (let name of imports) {
       if (name in packageNames) {
-        packages.add(name);
+        packages.add(packageNames[name]);
```
rth, line 376: Looks like we didn't have enough tests for this

rth (approved):
Generally LGTM. I think we can merge the rest in this PR. Thanks again for all you work @joemarshall !

rth:
The only thing is that pillow on chrome seems to frequently timeout (or there are timeouts in other packages as well). Though I have also seen time outs on master recently, so I would be OK merging this as is, and try to iron out unreliable tests in follow up PRs.

emsdk/tests/test_dyncall.py
joemarshall: Oops, those are left in there, I think bigint shouldn't be in at least until wasm bigint is enabled on everything

rth:
I pushed a few commits to fix review comments as I'm very much looking forward to see this merged :)

rth:
BTW it looks like with the upstream backend in this PR pure Python code in pyodide code runs 25 to 30% faster on average (ref https://github.com/iodide-project/pyodide/issues/1120)

Below are benchmark results for Firefox aggregated over 3 CI runs in this PR and on master. Chrome is comparable. Benchmarks on the top are mostly pure python code, and at the bottom code mostly using numpy,



image

rth:
The only concern is that Chrome does seem to time out more often here and we removed the custom timeout in `conftest.py` which was a few min, while the current default timeout of 10 or 20 min is too long.

dalcde:
The current timeout should be 30 seconds, aiui.

rth:
Also actually timeouts for packages in Chrome might have something to do with the fact that this removes xfail on scipy dependent packages. Previously it's not that they weren't working, but had a non negligible change of failing (or I guess now time out) hence the xfail.

rth:
Thanks again for your work on this @joemarshall and @dalcde for improving the build setup earlier to make it possible! This fixed a number of scipy related issues (linked above) and reduced the scipy size significantly https://github.com/iodide-project/pyodide/issues/240#issuecomment-775384365, which was a blocker for a long time.

PR #1103, opened 2021-01-10
Don't print python traceback in pytest_wrapper

dalcde:


dalcde:
(the test failures are due to flakiness; I find it handy to leave it in this state to see the effects of the change)

rth (approved):
Sounds good, thanks!

PR #1105, opened 2021-01-10
Add  `meta.yaml` validation

rth:
This adds a function to validate the parsed `meta.yaml`. In particular to check that,
 - there are no unknown keys (to avoid issues like https://github.com/iodide-project/pyodide/pull/1099)
 - that the keys have the expected dtype. For instance version `0.16` (without quotes) was previously incorrectly parsed as float. Now it would error during validation instead.

It also helps to have a single place where all valid `meta.yaml` keys are programmatically enforced, to check that we are not forgetting anything in the documentation.

In general `meta.yaml` validation is now run as part of `packages/test_common.py::test_parse_package` before the build, so it should provide fast feedback about errors during packaging.

Currently type validation is fairly basic. For instance we check for `list` and not `List[str]`. I'm not if there is a simple way to improve that without adding too much dependencies (i.e. without using something like typegard or pydantic). In any case that could be investigated in a follow up PR.

### DIFF
--- a/docs/new_packages.md
+++ b/docs/new_packages.md
@@ -161,6 +161,12 @@ Shell commands to run after building the library. These are run inside of
 
 (This key is not in the Conda spec).
 
+#### `build/replace-libs`
+
+A list of strings of the form `<old_name>=<new_name>`, to rename libraries when linking. This in particular
+might be necessary when using emscripten ports.
+For instance, `png16=png` is currently used in matplotlib.
+
 ### `requirements`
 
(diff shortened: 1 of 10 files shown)


dalcde:
How about checking it during build time? This would be useful for people
who want to build third party packages.

rth:
We are checking it each time the `meta.yaml` is loaded, so also at build time. I meant that it's additionally checked as part of the CI.

dalcde (approved):

PR #1106, opened 2021-01-10
ENH: runPython variant with function-like scoping instead of module-like scoping

### LINKED ISSUE #1000: Use separate namespace for runPython doing internal stuff
pyodide needs to import packages for various internal purposes. Currently these packages pollute the global namespace. #922 demonstrates a mixture of two problems: one is a confusion between `pyodide_py` and `pyodide_js` (see #992). The other problem is this namespace pollution.

Pending merge of #923, we can use custom namespaces. I think we should make a separate namespace for internal use and a separate entrypoint `runPythonInternal` which does `eval_code(code, internal_namespace)` and then use this in initialization and for stuff like `invalidate_caches`.

hoodmane:
Building on #1083, I added a new pyodide function called `runPythonWithLocals` and used it for the two calls to `runPython` in `pyodide.js` to prevent namespace leakage. Basically, `runPython` runs code at module scope, as if it was just written into the top level of the file whereas `runPythonWithLocals` runs code as if it were the body of a function call.

I think `runPythonWithLocals` should be made a public API, it is easy to imagine reasons users might want it. I would be interested in a discussion of what the best name would be.

Resolves #1000.


hoodmane:
I created a function `pyodide_py._eval_code_with_locals` which is a workaround to use until #900 is implemented. In particular the problem is:
Python:
```python
def f(a):
   return str(type(a))
```
Javascript:
```javascript
console.log(pyodide.globals.f(pyodide.globals.dict())) // jsproxy
```
What happened is we created an empty dict, `python2js` converted it to an empty javascript object, and then `js2python` converted it to a JsProxy. 

Actually all that's actually required of `locals` is that it implement `PyMappingProtocol`. `JsProxy` _does_ implement `PyMappingProtocol` so we _could_ use it. The results would be really weird though:
```javascript
pyodide.runPythonWithLocals(`
    a = {}
    a[(1,)] = 7
    a[1] # Yup, this is 7!
`);
```

dalcde:
In #1047 I thought it would be useful to be able to use the same local namespace for various calls. I haven't checked after #1083, but before that, the namespace has to be a pyproxy of a python dict, which is slightly annoying to procure

hoodmane:
> the namespace has to be a pyproxy of a python dict, which is slightly annoying to procure

Still true, still annoying. I'm getting closer to being able to implement #900, but if you want in the meantime I could add a C function that does roughly:
```C
PyObject* dict = PyDict_New();
return pyproxy_new(dict);
```
(with some extra stuff for error handling and ref counting).

hoodmane:
Probably every test should have a separate local namespace and some of the uses of `selenium_standalone` should be removed. If many tests in a row define the same global variable, it's hard to be sure that a later test isn't succeeding only because of pollution from a different test.

dalcde:
On Mon, Jan 11, 2021 at 05:51:40AM -0800, Hood Chatham wrote:
>> the namespace has to be a pyproxy of a python dict, which is slightly annoying to procure
>
>Still true, still annoying. I'm getting closer to being able to implement #900, but if you want in the meantime I could add a C function that does roughly:

I suspect we can achieve the same with 

let ns = runPython("{}")

hoodmane:
> I suspect we can achieve the same with `let ns = runPython("{}")`

NOPE! `ns` will be an empty argument. You need to short circuit `python2js` somehow, which you can't do through the normal entrypoints.

PR #1108, opened 2021-01-10
Added "make debug" which builds with DEBUG_F flag

hoodmane:
DEBUG_F provides extra logging info: whenever `FAIL` is invoked, a messages is printed with a javascript stack trace, and the source line function and file where the failure happened. When #1080 is merged, it will add debug statements whenever an error occurs in `EM_JS`.

I also added a missing semicolon in the `DEBUG_F` version of `FAIL`.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -208,4 +208,6 @@ minimal :
 	PYODIDE_PACKAGES="micropip" make
 
 debug :
-	EXTRA_C_FLAGS="-D DEBUG_F" PYODIDE_PACKAGES="micropip" make
+	EXTRA_C_FLAGS="-D DEBUG_F -s ASSERTIONS=2" \
+	PYODIDE_PACKAGES="micropip,pyparsing,pytz,packaging,kiwisolver" \
+	make
--- a/Makefile
+++ b/Makefile
@@ -15,7 +15,10 @@ PYODIDE_CXX=$(PYODIDE_ROOT)/ccache/em++
 CC=emcc
 CXX=em++
 OPTFLAGS=-O2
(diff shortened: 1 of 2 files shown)


dalcde:
Should debug enable `-s ASSERTIONS=2` as well?

hoodmane:
Let me check how much difference that makes to the linking time.

hoodmane:
Okay they take the same amount of time so I'll add it.

hoodmane:
Btw, with `-s ASSERTIONS=2` there's a bunch of weird warnings like:
`warning: unexpected number of arguments 0 in call to 'hiwire_array', should be 1`

Also, no matter what it always says:
`emcc:INFO: Enabling --no-heap-copy because -s ALLOW_MEMORY_GROWTH=1 is being used with file_packager.py (pass --no-heap-copy to suppress this notification)`

hoodmane:
Also instead of `PYODIDE_PACKAGES="micropip"` I should probably make it do all the packages for the core tests.

hoodmane:
Oh this now has the problem that it's annoying to get an extra package built because the recipe writes over `PYODIDE_PACKAGES`. Maybe `PYODIDE_PACKAGES+=...`?

hoodmane:
@dalcde @rth I think this is a simple enough addition, would be helpful to merge (I will switch from using `make minimal` to `make debug` as my default).

Makefile
```diff
@@ -203,3 +205,8 @@ check:
 
 minimal :
 	PYODIDE_PACKAGES="micropip" make
+
+debug :
+	EXTRA_CFLAGS="-D DEBUG_F -s ASSERTIONS=2" \
```
dalcde, line 210: I think `ASSERTIONS=2` should be an ldflag. It affects the js code gen

hoodmane: Funny, I just noticed that as I was working on #1126.

PR #1111, opened 2021-01-11
BUGFIX: If python2js_buffer fails, python2js leaves the error flag in an invalid state

### LINKED ISSUE #1110: Soundness error in python2js_buffer
This bug goes back at least to version 0.15.0. It's been messing up my error handling code.
First:
```python
import numpy as np
x = np.array([['string1', 'string2'], ['string3', 'string4']])
```
Second: type `x<enter>1<enter>x<enter>1<enter>`.

hoodmane:
I fixed #1110. If `python2js_buffer` fails, we need to clear the Python error flag before trying a backup approach.

### DIFF
--- a/src/tests/test_typeconversions.py
+++ b/src/tests/test_typeconversions.py
@@ -339,7 +339,7 @@ def test_memoryview_conversion(selenium):
     selenium.run_js(
         """
         pyodide.globals.a
-        if(pyodide._module._PyErr_Occurred){
+        if(pyodide._module._PyErr_Occurred()){
             pyodide._module._pythonexc2js();
         }
         """
--- a/src/tests/test_typeconversions.py
+++ b/src/tests/test_typeconversions.py
@@ -327,3 +327,20 @@ def test_javascript_error_back_to_js(selenium):
         return pyodide.globals["py_err"] === err
         """
     )
+
+
+def test_memoryview_conversion(selenium):
+    selenium.run(
+        """
+        import array
+        a = array.array("Q", [1,2,3])
+        """
+    )
+    selenium.run_js(
+        """
+        pyodide.globals.a
+        if(pyodide._module._PyErr_Occurred){
+            pyodide._module._pythonexc2js();
+        }
+        """
+    )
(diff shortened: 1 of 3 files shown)


dalcde:
Can you test this without using numpy? This ought to be a type conversion test.

hoodmane:
I don't know, will investigate. I'm going to update `conftest` so that it checks whether `PyErr_Occurred()` at the end of every test too, that would have caught this bug.

hoodmane:
I think we should also try converting the numpy tests from `selenium_standalone` to `selenium`. I am suspicious that the reason they were changed to `selenium_standalone` in the first place may have been because of this bug.

dalcde:
Also, what's the difficulty in turning the original bug reproduction
method into an actual test case?

hoodmane:
Original bug reproduction involved the code in `console.html` in a weird way.

dalcde:
You will have to load the numpy package first.

hoodmane:
Indeed I should.

src/tests/test_typeconversions.py
```diff
@@ -327,3 +327,20 @@ def test_javascript_error_back_to_js(selenium):
...
+    selenium.run(
+        """
+        import array
+        a = array.array("Q", [1,2,3])
+        """
+    )
+    selenium.run_js(
+        """
+        pyodide.globals.a
+        if(pyodide._module._PyErr_Occurred()){
```
dalcde, line 342: This check would be made redundant by #1113, right?

dalcde: If so, I think the correct course of action is to merge this and then remove this line in #1113

hoodmane: Either we should remove this here and keep the line in #1113 or keep the line here and close #1113.

dalcde: I meant merge this test, and then in #1113 remove this particular line in this test

hoodmane: Oh yeah. I agree.

PR #1115, opened 2021-01-11
Remove dead code

dalcde:


hoodmane:
This should read 
```
  if (PyDict_Check(pyobj)) {
    result = PyObject_SetItem(pyobj, pykey, pyval);
  } else {
    result = PyObject_SetAttr(pyobj, pykey, pyval);
  }
```
[PyObject_SetItem](https://docs.python.org/3/c-api/object.html#c.PyObject_SetItem) and [PyObjectSetAttr](https://docs.python.org/3/c-api/object.html?highlight=pyobject_setattr#c.PyObject_SetAttr) return error codes which need to be checked.

dalcde:
I see. Do you want to open a PR to fix?

PR #1118, opened 2021-01-11
MAINT: Use PyConfig to configure python interpreter

hoodmane:
Prior to Python 3.8, configuring Python was an ad-hoc process. [PEP 587](https://www.python.org/dev/peps/pep-0587/) added a systematic way to configure Python. Using this leads to simpler, clearer code with compile time checking for the config fields, better error handling and more discoverable documentation.

For example, in `Py_InitializeEx(0)` the `0` means "don't install signal handlers". This is opaque and requires looking `Py_InitializeEx` in the docs to interpret. (Of course a comment could fix that.) Instead we now say:
```C
config.install_signal_handlers = false;
```

### DIFF
--- a/src/core/main.c
+++ b/src/core/main.c
@@ -24,6 +24,11 @@
     }                                                                          \
   } while (0)
 
+#define FAIL_IF_STATUS_EXCEPTION(status)                                       \
+  if (PyStatus_Exception(status)) {                                            \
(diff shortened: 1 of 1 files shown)


rth (approved):
This looks nicer indeed! Thanks!

PR #1121, opened 2021-01-11
Fix mistake in merging #1047 and #1083

hoodmane:
#1083 changed the signature of `eval_code` and #1047 cares about the method signature of `eval_code`. Due to a merging race condition these both got merged into master even though they don't like each other. This adjusts #1047 to reflect the change in #1083.


rth (approved):

hoodmane:
It's nice also that nobody even made a mistake here -- this sort of problem is more or less unavoidable because "commits change different files" does not imply "commits are orthogonal."

PR #1124, opened 2021-01-11
FIX: Make JsBoundMethod a subclass of JsProxy

### LINKED ISSUE #768: TypeError: Illegal invocation

### LINKED ISSUE #788: Nested attribute access in JS->Python type conversion

### LINKED ISSUE #1123: `from js import window` gives a badly glitched object
We have two types of `window`: 
Type 1: Equal to all windows, can be accessed to retrieve a variable but if variable is a function instead it gives a Type 2 window.
Type 2: Equal to windows of Type 1 but not to other windows of Type 2 (so equality is not transitive). Cause AttrributeError when accessed.

Type 1 windows result from converting `window` from javascript. Both Type 1 and Type 2 windows are converted to the ordinary window object if converted to javascript.

This following code works in both v0.16.1 and in the most recent master build.
```javascript
pyodide.runPython(`
      import js
      from js import Array
      from js import window
      assert js.Array == Array # True, good!
      assert window.Array != Array # what?
      window2a = window.Array
      window2b = window.Array
      # equality is not transitive:
      assert window2a != window2b
      assert window2a == window 
      assert window2b == window
      # window2's cannot be accessed:
      try:
          window2a.Array # AttributeError: 'JsBoundMethod' object has no attribute 'Array'
      except AttributeError as e:
          print(e)
      else:
          assert False
`)
pyodide.globals.window2a === pyodide.globals.window2b // true
window1 = pyodide.globals.window2a
pyodide.runPython(`
      from js import window1
      assert window1 == window2a
      assert window1 == window2b
      assert window1 == window
      try:
          window1.Array
      except AttributeError as e:
          assert False
`);
```

hoodmane:
Fixes #788 (tested in `test_nested_attribute_access`) and fixes #1123 (tested in `test_window_isnt_super_weird_anymore`). This also resolves a problem that blocks any fix for #768.


rth:
This is a very useful, thanks!

If this the fix for https://github.com/iodide-project/pyodide/issues/788 why is test_nested_attribute_access passing in https://github.com/iodide-project/pyodide/pull/1126 ? Generally there is some diff overlap with that PR: which one should be merged first?

hoodmane:
Ugh yes, there's like three overlapping bugs here. I merged this one into #1126 in order to make the #1126 tests pass (otherwise a bunch of tests fail). 

[Commit 709118c](https://github.com/iodide-project/pyodide/pull/1126/commits/709118c0291f66d5bf736e885e37d75ab209df2c) on that branch has everything needed to fix the bug, after that is only test fixes.

So merge this one first and then I can merge master on that one and then merge that one? 

Also I have some tests that will only pass once this, #1126, and #1033 all are merged. I guess I can add them to the last one to merge (if I can remember where I put them...). 

I think between this one and #1033 what should happen is that the Vectorcall implementation in #1033 should be moved to JsBoundMethod, the implementation of `JsProxy_Vectorcall ` should be removed so that `JsProxy` is no longer callable (I don't think it's possible to get a function anymore that _isn't_ a `JsBoundMethod`) and then rename `JsBoundMethod` to `JsProxy_Callable`.

hoodmane:
Oh yeah #1098 is also blocked on this -- after #1124 and possibly #1126 are merged, the test suite for #1098 should start passing.

hoodmane:
Also finishing #972 is blocked on this.

rth (approved):
Thanks! Generally LGTM, but I'm not the most relevant reviewer for C code. 

Please add a changelog entry.

PR #1125, opened 2021-01-11
Interactive console in pyodide-py

casatir:
This is the 1/3 part of the #875 splitting. It addresses:

 - sync behavior
 - full stdout/stderr support
 - clean namespace
 - tests
 - bigger font (in `console.html`)
 
In a separate PR, I will address:

- correct result representation
- more tests

And in a last PR, I will propose a completion system.

### DIFF
--- a/src/pyodide-py/pyodide/console.py
+++ b/src/pyodide-py/pyodide/console.py
@@ -17,8 +17,13 @@ class _StdStream(io.TextIOWrapper):
         Function to call at each flush.
     """
 
-    def __init__(self, flush_callback: Callable[[str], None]):
-        buffer = io.BufferedWriter(_CallbackBuffer(flush_callback))
+    def __init__(
+        self, flush_callback: Callable[[str], None], name: Optional[str] = None
+    ):
+        # we just need to set internal buffer's name as
+        # it will automatically buble up to each buffer
+        internal_buffer = _CallbackBuffer(flush_callback, name=name)
+        buffer = io.BufferedWriter(internal_buffer)
         super().__init__(buffer, line_buffering=True)  # type: ignore
 
 
@@ -32,8 +37,11 @@ class _CallbackBuffer(io.RawIOBase):
         Function to call at each flush.
     """
 
-    def __init__(self, flush_callback: Callable[[str], None]):
+    def __init__(
+        self, flush_callback: Callable[[str], None], name: Optional[str] = None
+    ):
         self._flush_callback = flush_callback
+        self.name = name
 
     def writable(self):
         return True
@@ -97,8 +105,11 @@ class InteractiveConsole(code.InteractiveConsole):
         self._old_stdout = sys.stdout
         self._old_stderr = sys.stderr
 
-        sys.stdout = _StdStream(meta_stdout_callback)
-        sys.stderr = _StdStream(meta_stderr_callback)
+        # it would be more robust to use sys.stdout.name and sys.stderr.name
+        # but testing system oveload them. Anyway it should be pretty stable
+        # upstream.
+        sys.stdout = _StdStream(meta_stdout_callback, name="<stdout>")
+        sys.stderr = _StdStream(meta_stderr_callback, name="<stderr>")
 
     def restore_stdstreams(self):
         """Restore stdout/stderr to the value it was before
--- a/src/pyodide-py/pyodide/console.py
+++ b/src/pyodide-py/pyodide/console.py
@@ -61,17 +61,21 @@ class InteractiveConsole(code.InteractiveConsole):
     """Interactive Pyodide console
 
     Base implementation for an interactive console that manages
-    stdout/stderr. Their value will be restored at destruction or
-    at `restore_stdstreams` call.
+    stdout/stderr redirection.
 
     `self.stdout_callback` and `self.stderr_callback` can be overloaded.
 
     Parameters
     ----------
+    locals
+        Namespace to evaluate code.
     stdout_callback
         Function to call at each `sys.stdout` flush.
     stderr_callback
         Function to call at each `sys.stderr` flush.
+    persistent_stream_redirection
+        Wether or not the std redirection should be kept between calls to
+        `runcode`.
     """
 
     def __init__(
@@ -79,37 +83,49 @@ class InteractiveConsole(code.InteractiveConsole):
         locals: Optional[dict] = None,
         stdout_callback: Optional[Callable[[str], None]] = None,
         stderr_callback: Optional[Callable[[str], None]] = None,
+        persistent_stream_redirection: bool = False,
     ):
         super().__init__(locals)
+        self._stdout = None
+        self._stderr = None
         self.stdout_callback = stdout_callback
         self.stderr_callback = stderr_callback
-        self.redirect_stdstreams()
+        self._persistent_stream_redirection = persistent_stream_redirection
+        if self._persistent_stream_redirection:
+            self.redirect_stdstreams()
 
     def redirect_stdstreams(self):
         """ Toggle stdout/stderr redirections. """
 
-        # we use meta callbacks to allow self.std{out,err}_callback
-        # overloading.
-        # we check callback against None at each call since it can be
-        # changed dynamically.
+        if self._stdout is None:
+            # we use meta callbacks to allow self.std{out,err}_callback
+            # overloading.
+            # we check callback against None at each call since it can be
+            # changed dynamically.
+            def meta_stdout_callback(*args):
+                if self.stdout_callback is not None:
+                    return self.stdout_callback(*args)
 
-        def meta_stdout_callback(*args):
-            if self.stdout_callback is not None:
-                return self.stdout_callback(*args)
+            # for later restore
+            self._old_stdout = sys.stdout
 
-        def meta_stderr_callback(*args):
-            if self.stderr_callback is not None:
-                return self.stderr_callback(*args)
+            # it would be more robust to use sys.stdout.name but testing
+            # system oveload them. Anyway it should be pretty stable
+            # upstream.
+            self._stdout = _StdStream(meta_stdout_callback, name="<stdout>")
 
-        # for later restore
-        self._old_stdout = sys.stdout
-        self._old_stderr = sys.stderr
+        if self._stderr is None:
 
-        # it would be more robust to use sys.stdout.name and sys.stderr.name
-        # but testing system oveload them. Anyway it should be pretty stable
-        # upstream.
-        sys.stdout = _StdStream(meta_stdout_callback, name="<stdout>")
-        sys.stderr = _StdStream(meta_stderr_callback, name="<stderr>")
+            def meta_stderr_callback(*args):
+                if self.stderr_callback is not None:
+                    return self.stderr_callback(*args)
+
+            self._old_stderr = sys.stderr
+            self._stderr = _StdStream(meta_stderr_callback, name="<stderr>")
+
+        # actual redirection
+        sys.stdout = self._stdout
+        sys.stderr = self._stderr
 
     def restore_stdstreams(self):
         """Restore stdout/stderr to the value it was before
@@ -117,8 +133,17 @@ class InteractiveConsole(code.InteractiveConsole):
         sys.stdout = self._old_stdout
         sys.stderr = self._old_stderr
 
+    def runcode(self, code):
+        if self._persistent_stream_redirection:
+            super().runcode(code)
+        else:
+            self.redirect_stdstreams()
+            super().runcode(code)
+            self.restore_stdstreams()
+
     def __del__(self):
-        self.restore_stdstreams()
+        if self._persistent_stream_redirection:
+            self.restore_stdstreams()
 
     def banner(self):
         """ A banner similar to the one printed by the real Python interpreter. """
--- a/src/pyodide-py/pyodide/console.py
+++ b/src/pyodide-py/pyodide/console.py
@@ -119,3 +119,9 @@ class InteractiveConsole(code.InteractiveConsole):
 
     def __del__(self):
         self.restore_stdstreams()
+
+    def banner(self):
+        """ A banner similar to the one printed by the real Python interpreter. """
+        # copyied from https://github.com/python/cpython/blob/799f8489d418b7f9207d333eac38214931bd7dcc/Lib/code.py#L214
+        cprt = 'Type "help", "copyright", "credits" or "license" for more information.'
+        return f"Python {sys.version} on {sys.platform}\n{cprt}"
--- a/src/pyodide-py/pyodide/console.py
+++ b/src/pyodide-py/pyodide/console.py
@@ -2,6 +2,7 @@ from typing import Optional, Callable
 import code
 import io
 import sys
+import platform
 
 
 __all__ = ["InteractiveConsole"]
@@ -149,4 +150,4 @@ class InteractiveConsole(code.InteractiveConsole):
         """ A banner similar to the one printed by the real Python interpreter. """
         # copyied from https://github.com/python/cpython/blob/799f8489d418b7f9207d333eac38214931bd7dcc/Lib/code.py#L214
         cprt = 'Type "help", "copyright", "credits" or "license" for more information.'
-        return f"Python {sys.version} on {sys.platform}\n{cprt}"
+        return f"Python {platform.python_version()} {platform.python_build()} on WebAssembly VM\n{cprt}"
--- a/src/pyodide-py/pyodide/console.py
+++ b/src/pyodide-py/pyodide/console.py
@@ -1,0 +1,22 @@
+from typing import Optional
+import code
+import io
+import sys
+
+
+class InteractiveConsole(code.InteractiveConsole):
+    """
+    Base implementation for an interactive console that manages
+    stdout/stderr. It does not actually run any code.
+
+    A subclass may overload runcode and access std streams via
+    sys.stdout.getvalue() and sys.stderr.getvalue(). Remember to
+    start calling parent's runcode.
+    """
+
+    def __init__(self, locals: Optional[dict] = None):
+        super().__init__(locals)
+
+    def runcode(self, code):
+        sys.stdout = io.StringIO()
+        sys.stderr = io.StringIO()
--- a/src/pyodide-py/pyodide/console.py
+++ b/src/pyodide-py/pyodide/console.py
@@ -1,22 +1,110 @@
-from typing import Optional
+from typing import Optional, Callable
 import code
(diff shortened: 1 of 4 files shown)


hoodmane:
Can you temporarily make this into a `console2.html` and leave the original `console.html` as is so I can easily compare the two side by side?

Stream redirection should be handled with a context manager e.g.,:
```python
from contextlib import redirect_stdout, redirect_stderr
import pyodide
def run_code(code):
   with  redirect_stdout(mystdoutstream), redirect_stderr(WriteStream(mystderr):
      pyodide.eval_code(code)
```
Exactly what you don't want is for the existence of a terminal object somewhere to funnel all writes to stdout into your terminal.

casatir:
Thanks @hoodmane for looking at this.

> Can you temporarily make this into a console2.html and leave the original console.html as is so I can easily compare the two side by side?

Can't understand why you need me to do this.

> Stream redirection should be handled with a context manager e.g.

At first, I planed to do it like this. But ...

> Exactly what you don't want is for the existence of a terminal object somewhere to funnel all writes to stdout into your terminal.

I think that it could be exactly what you want! You can stop the redirection with `restore_stdstreams` if you need. There can be async stuff that continue to print to stdout that you precisely want to catch.

hoodmane:
I am constantly opening up the browser console and typing `pyodide.runPython(stuff)` into the browser console. I don't like it when messages caused by this get vacuumed up by the terminal. I think commands invoked through the terminal should log into the terminal, commands invoked elsewhere should be logged into the browser console.

Effectively there are two separate use cases: normal people and developers. I am okay if you make this stream change to `console.html` as long as I can have a `devconsole.html` that doesn't do it. Which gets back to my request to keep the current `console.html` around in some form.

> There can be async stuff that continue to print to stdout that you precisely want to catch.

I would try to solve this in another way before going to the "alter global state" approach which I think is a last resort.

hoodmane:
One other noteworthy thing: normally if you write to `stderr`, when the message is printed in the console it gets a javascript stack trace attached. If you redirect that to the terminal and swallow the console message, the javascript stack trace is lost.

casatir:
Thanks @hoodmane for looking at this.

> I am constantly opening up the browser console and typing pyodide.runPython(stuff) into the browser console. I don't like it when messages caused by this get vacuumed up by the terminal. I think commands invoked through the terminal should log into the terminal, commands invoked elsewhere should be logged into the browser console.

Can't get your point: (after the first command is run,) stdout were already constantly redirected to terminal. I am not changing the behavior here. If you want an html page that loads Pyodide so that you can do some `pyodide.runPython(stuff)` without interference then `console.html` may not be the right tool.

>> There can be async stuff that continue to print to stdout that you precisely want to catch.

> I would try to solve this in another way before going to the "alter global state" approach which I think is a last resort. Also, if the async stuff is happening on the javascript side, this won't catch it anyways.

It could. When implementing p5.js support in Basthon, I can have a `draw` function, written in Python and called from P5 that can do some async print. This is the JS side that calls some Python.

> One other noteworthy thing: normally if you write to stderr, when the message is printed in the browser console it gets a javascript stack trace attached. If you redirect that to the terminal and swallow the browser console message, the javascript stack trace is lost.

Can't get any trace from `pyodide.runPython("print('foo', file=sys.stderr)")`

hoodmane:
> Can't get any trace from `pyodide.runPython("print('foo', file=sys.stderr)")`

Sure can:
[](url)
console-print

hoodmane:
> If you want an html page that loads Pyodide so that you can do some `pyodide.runPython(stuff)` without interference then `console.html` may not be the right tool.

The current `console.html` is far from perfect but it gets the job done. I would be fine with copying it to `devconsole.html` and applying this stdout vacuuming semantics in `console.html`.

hoodmane:
> clean namespace

What do you mean by this? I am having trouble finding where in this code this is taken care of.

hoodmane:
> Can't get any trace from `pyodide.runPython("print('foo', file=sys.stderr)")`

Actually now that I'm looking into it, the behavior is kind of funny: printing to stderr works correctly until you execute the first command in the repl. After the first command in the repl is executed, all writes to stdout and stderr are swallowed. I can fix it to use context manager in `devconsole.html` so that it will work a bit more consistently.

casatir:
> What do you mean by this? I am having trouble finding where in this code this is taken care of.

Just do something like `Console()` instead of `Console(locals=globals())` to let `code.InteractiveConsole` manage it's own namespace.

> Actually now that I'm looking into it, the behavior is kind of funny: printing to stderr works correctly until you execute the first command in the repl. After the first command in the repl is executed, all writes to stdout and stderr are swallowed.

This is what I was saying by:

> (after the first command is run,) stdout were already constantly redirected to terminal. I am not changing the behavior here.

This is because `sys.stdout/stderr` are set at `runcode`.

casatir:
The point is that I want `console.html` to be a showcase and you want it to be a dev tool...

hoodmane:
> The point is that I want console.html to be a showcase and you want it to be a dev tool...

@casatir The point is that you don't have great reading comprehension:

>  I am okay if you make this stream change to `console.html` as long as I can have a `devconsole.html` that doesn't do it.

>  I would be fine with copying it to `devconsole.html` and applying this stdout vacuuming semantics in `console.html`.

> I can fix it to use context manager in `devconsole.html` so that it will work a bit more consistently.

casatir:
> @casatir The point is that you don't have great reading comprehension. I said three times I wanted a tool called devconsole.html which is similar to console.html but designed to be a devtool:

Ouch! I was just saying that to go your way and support the fact that we need both. You start the review with a dev tool in mind but I was advertising for a showcase in #875:

> It's a showcase for Pyodide

This was just to make it clear that, at first, we have different purpose in mind.

Moreover, please be kind. With this kind of answer, you will ruin the motivation of volunteers that takes on their (work and family) time for working on Pyodide and that are not fluent in english.

rth:
Thanks for this improvement @casatir and for the review @hoodmane !

> Moreover, please be kind. 

Yes, absolutely. It's OK to have a technical disagreement, and a PR author can refuse changes suggested in a PR, while still having a pleasant conversation. As far as I understand the implied meaning was "I won't change it, unless you have better arguments" rather than literately "I don't understand". I know it's easy to misunderstand each other.

In this case, I would also prefer to keep a single `console.html` and address @hoodmane 's concern,
 
 > Stdout were already constantly redirected to terminal. I am not changing the behavior here.

likely in a different PR, unless the changes in this PR prevent that?

hoodmane:
After my repeated attempts at compromise, @casatir's tone comes across as accusatory and lacking of acknowledgement of the compromises.

> This was just to make it clear that, at first, we have different purpose in mind.

"at first" is a qualification that you didn't give in your original message. It also makes your comment rather stale information, since I suggested other positions based on the stuff you've already said. It is of course true that the user I am thinking of first is myself, but I am looking for some option that can make us both happy (of course making us both happy is different than doing the best thing or even from doing a reasonable thing, but hopefully it is at least vaguely related). 

> The point is that I want console.html to be a showcase and you want it to be a dev tool...

This is an accusation that I don't care about end users? I'm thinking: He isn't happy with the compromise options I offered? He just doesn't want my feedback? Or did he just not read my posts very carefully and didn't notice that I changed my position at all? It sounds like a hard bargain to me.

hoodmane:
> in a different PR

Yes, I am now convinced by that. The current behavior is weird, this preserves the weird current behavior while making unrelated improvements.

hoodmane:
The main source of my confusion in this discussion is that I've been taking the ground truth state of the system to be what happens in a fresh copy of `console.html`. Because the fresh copy of `console.html` has different behaviors from a `console.html` that has been used once, I had incorrect beliefs about how `console.html` normally works.

casatir:
Thanks @rth.

> It's OK to have a technical disagreement

Totally agree but it's not OK to have personal attack like this:

> you don't have great reading comprehension

> After my repeated attempts at compromise, @casatir's tone comes across as accusatory and lacking of acknowledgement of the compromises.

Wait wait wait. 
1. You never suggest implementing this compromise in this PR.
2. My PR doesn't affect the current behavior (at least after the first command is run) on the point we are interested in.
3. Can't find where my tone comes across as accusatory but if I hurt you I apology. No acknowledgment from you for this work neither.
4. Please, check your tone.

> I am looking for some option that can make us both happy (of course making us both happy is different than doing the best thing or even from doing a reasonable thing, but hopefully it is at least vaguely related).

Does this suggest I want to do wrong or unreasonable things?

>This is an accusation that I don't care about end users? I'm thinking: He isn't happy with the compromise options I offered? He just doesn't want my feedback? Or did he just not read my posts very carefully and didn't notice that I changed my position at all? It sounds like a hard bargain to me.

There is no accusation here, I promise. I have no doubt you are really concern by end users. I had just not understand you proposes this compromise for this PR.

> The main source of my confusion in this discussion is that I've been taking the ground truth state of the system to be what happens in a fresh copy of console.html. Because the fresh copy of console.html has different behaviors from a console.html that has been used once, I had incorrect beliefs about how console.html normally works.

Totally agree that this stupid bug makes a lot of confusion in our discussion.

Let's forget all this. I have a proposition that will certainly makes all of us happy. I am not really interested by the `console.html` behavior but more on the underlying `pyodide.console.InteractiveConsole` one. What I propose is something like that

``` python
class InteractiveConsole(code.InteractiveConsole):
    def __init__(
        self,
        locals: Optional[dict] = None,
        stdout_callback: Optional[Callable[[str], None]] = None,
        stderr_callback: Optional[Callable[[str], None]] = None,
        persistent_stream_redirection: bool = False,
    ):
``` 

The `persistent_stream_redirection` (maybe there is better name for it) when set to `True` will act as it is in this PR while it will act as you (and @rth) want when set to `False`. Default value will be `False` and it will be set to `False` in `console.html`.

WDYT?

If you are OK, I'll add this to this PR.

hoodmane:
> You never suggest implementing this compromise in this PR.

Right, I meant to imply that you should do nothing in this PR and I would do something to deal with it in a future PR.

> I have a proposition that will certainly makes all of us happy.
> ...

Sounds perfect!

> Default value will be `False` and it will be set to `False` in `console.html`.

I don't have a super strong opinion about what the default value should be, I still think it could be a good idea to have a `devconsole.html` with slightly different config than `console.html`. But if we can only have one official entrypoint for both end users and devs, then it would be nice to have one with this set to `False.

hoodmane:
@casatir Sorry for the misunderstanding, I really appreciate your work! I will try not to read your tone as so confrontational in the future. This has been a really rough year and I am not exactly a pillar of emotional stability. 

> No acknowledgment from you for this work neither.

That's definitely my bad!

> Please, check your tone.

I will try.

> Does this suggest I want to do wrong or unreasonable things?

No sorry don't mean to suggest that. I mean that theoretically "contributors are happy" is independent from "a good decision was made", you can have either one without the other, both, or neither.

casatir:
> Sounds perfect!

Great!

> I don't have a super strong opinion about what the default value should be,

Me neither.

> I still think it could be a good idea to have a devconsole.html with slightly different config than console.html. But if we can only have one official entrypoint for both end users and devs, then it would be nice to have one with this set to False.

Could be nice to have @rth though on this.

And what about an URL parameter like `?dev=true/false` or something like that. It's easy to parse with the JS URL parser in `console.html` and it can be hidden to standard users while advertised to devs. Then we can manage the way the console is created.

hoodmane:
> And what about an URL parameter like `?dev=true/false` or something like that. It's easy to parse with the JS URL parser in `console.html` and it can be hidden to standard users while advertised to devs. Then we can manage the way the console is created.

I think this would be a good way to go as long as the config options are simple.

src/pyodide-py/pyodide/console.py
```diff
@@ -0,0 +1,127 @@
...
+        sys.stderr = self._old_stderr
+
+    def __del__(self):
+        self.restore_stdstreams()
+
+    def banner(self):
+        """ A banner similar to the one printed by the real Python interpreter. """
+        # copyied from https://github.com/python/cpython/blob/799f8489d418b7f9207d333eac38214931bd7dcc/Lib/code.py#L214
+        cprt = 'Type "help", "copyright", "credits" or "license" for more information.'
+        return f"Python {sys.version} on {sys.platform}\n{cprt}"
```
rth, line 127: This would show "Python 3.8 on emscripten". Technically it's true but it might be confusing to users. I'm not sure what else we could put there, WebAssembly VM, pyodide, browser? Also keeping the snake emoji we had before would be a nice touch.

src/pyodide-py/pyodide/console.py
```diff
@@ -0,0 +1,127 @@
...
+
+    Base implementation for an interactive console that manages
+    stdout/stderr. Their value will be restored at destruction or
+    at `restore_stdstreams` call.
+
+    `self.stdout_callback` and `self.stderr_callback` can be overloaded.
+
+    Parameters
+    ----------
+    stdout_callback
```
rth, line 71: locals also need a docstring entry

src/tests/test_console.py
```diff
@@ -0,0 +1,61 @@
...
+    assert sys.stderr.name == "<stderr>"
+
+    # std redirections
+    print("foo")
+    assert my_stdout == "foo\n"
+    print("bar", file=sys.stderr)
+    assert my_stderr == "bar\n"
+
+    # redirections disabled at destruction
+    shell.restore_stdstreams()
```
rth, line 52: If this test fails before this line does it mean that we get monkeypatched stdout/stderr globally? 

Maybe it would be good add a pytest fixure that stores and restores the stdout just in case. Something like,
```py
@pytest.fixture
def restore_stdstreams():
    # save streams to local variables
    yield
    # restore streams after the test function exits

def test_interactive_console_streams(restore_stdstreams):
    ...

```

BTW, it's interesting that pytest also captures stderr/stdout but does it at the level of file system descriptors https://docs.pytest.org/en/stable/capture.html#setting-capturing-methods-or-disabling-capturing

rth:
Minor comments above but otherwise LGTM. Thanks!

> And what about an URL parameter like ?dev=true/false or something like that. 

Also +1 for it.

rth:
Also please add a changelog entry.

src/pyodide-py/pyodide/console.py
casatir: It shows
```
Python 3.8.2 (default, Jan  4 2021, 12:17:49)
[Clang 6.0.1 (/b/s/w/ir/cache/git/chromium.googlesource.com-external-github.com on emscripten
>>> 
```
which is not nice, you are right... 

I see something like

```
Python 3.8.2 (default, Jan  4 2021, 12:17:49) on WebAssembly VM
```

for the default banner and

```
Python 3.8.2 (default, Jan  4 2021, 12:17:49) on WebAssembly VM
```

for the one customized in `console.html`. WDYT?

src/pyodide-py/pyodide/console.py
casatir: Right. Corrected.

src/pyodide-py/pyodide/console.py
casatir: Or maybe:

```
Python 3.8.2 (default, Jan  4 2021, 12:17:49) on WebAssembly VM
```

src/tests/test_console.py
casatir: Right, it could be a bad issue!

casatir:
Sorry for the force push but there were conflicts with #979 so I rebased on master.

src/templates/console.html
```diff
@@ -66,6 +71,8 @@
         };
 
         term.handlePythonResult = function(result) {
+          c.restore_stdstreams();
+          term.resume();
```
casatir, line 75: `console.html` is not really clean with this `restore_stdstreams()` in the JS code but they will disappear in the part 2/3 of the console improvement, I promise.

src/pyodide-py/pyodide/console.py
rth: I personally prefer the last version.

rth (approved):
Thanks! Please merge upstream master into this PR to resolve new conflicts.

**Edit:** Actually I'll do it.

rth:
Actually I fixed conflicts via UI. Will merge on green CI.

src/templates/console.html
```diff
@@ -11,51 +11,56 @@
...
+
+        let c = pyodide.pyimport('_c');
+
         function pushCode(line) {
-          handleResult(c.push(line))
+          handleResult(c.push(line));
         }
 
-        let term = $('body').terminal(
+        var term = $('body').terminal(
```
rth, line 49:
```suggestion
        let term = $('body').terminal(
```
unless it was intentional change?

src/pyodide-py/pyodide/console.py
casatir: Me too, this is the one I commited.

src/templates/console.html
casatir: No, probably forgotten when merging with #979.

casatir:
Conflicts should be fixed. `var/let` too. Is there a way to rebase on master without destroying the review history?

dalcde:
Merging master should be fine

casatir:
Thanks @dalcde. You mean `git merge upstream/master`? Right, i will do this. I am a fond of clean linear history so I rebase everywhere. But merging is fine since commits will be squashed at PR merge.

rth:
It's working very nicely. But now we definitely need to fix #877

src/pyodide-py/pyodide/console.py
```diff
@@ -0,0 +1,153 @@
...
+
+    def __del__(self):
+        if self._persistent_stream_redirection:
+            self.restore_stdstreams()
+
+    def banner(self):
+        """ A banner similar to the one printed by the real Python interpreter. """
+        # copyied from https://github.com/python/cpython/blob/799f8489d418b7f9207d333eac38214931bd7dcc/Lib/code.py#L214
+        cprt = 'Type "help", "copyright", "credits" or "license" for more information.'
+        return f"Python {platform.python_version()} {platform.python_build()} on WebAssembly VM\n{cprt}"
```
rth, line 153: This currently prints,
```
Python 3.8.2 ('default', 'Jan 15 2021 09:54:23') on WebAssembly VM
```
while the classical Python REPL displays,
```
Python 3.8.6 (default, Sep 25 2020, 09:36:53)
```
It would be nice to remove the quotes.

casatir: Arf, sorry for this. Good catch! It will be corrected in #1141. Is that OK?

rth: Yes, sure, thanks!

Issue #261, opened 2018-11-22
support named arguments/keywords in `JsProxy_Call`


oeway:
Python function calls support named arguments, but in JS there is no native named arguments. In ECMAScript 6, functions can take an object with default values:
```javascript
 function js_range({ a = 0, b = 10 }) {
        console.log(`${a}-${b}`)
 };

execute `js_range({})` in the JS console will get `0-10` as output.
execute `js_range({a: 99, b: 999})`  in the JS console will get `99-999` as output.
```

Right now we can do this in pyodide:
```python
from js import js_range

js_range({'a': 99, 'b': 999})

```

However, I think it would be better to call functions defined in JS (e.g.: `js_range` defined above) with named arguments in Python. The following code is not working now, but would be really great to have this feature:
```python
from js import js_range

js_range(a=99, b=999)
```

As to my understanding so far, this type of conversion should be relatively simple.  I found a function `JsProxy_Call` in https://github.com/iodide-project/pyodide/blob/master/src/jsproxy.c#L99 . 

`static PyObject* JsProxy_Call(PyObject* o, PyObject* args, PyObject* kwargs)`

In order to support named argument style, we just need to check if `kwargs` is empty or not, if not, pass kwargs as a JS object and append to the argument list.

mdboom:
This is a cool idea.  Is there a way to tell if the Javascript function is of this form?  I'm trying to see around corners if there might be negative consequences to doing this if the function isn't expecting this convention.

PR #1126, opened 2021-01-12
BUGFIX: Bind functions to window on `from js import func` (#461)

### LINKED ISSUE #461: from js import fetch treats fetch as a free function
`fetch` is a member function of `window`.
However, using `from js import fetch` doesn't realize that and leads to the error:

`TypeError: 'fetch' called on an object that does not implement interface Window.`

For Reproducing the Error:
```
%%py

from js import document, Request, fetch, URL
img_tag = document.createElement('img')
req = Request.new('https://i.ibb.co/3f4yJQS/face4.jpg')

def func(response):
    return response.blob()

def func2(blob):
    objURL = URL.createObjectURL(blob)
    img_tag.src = objURL

fetch(req).then(func).then(func2)

document.body.appendChild(img_tag)
```

hoodmane:
Instead of making ad-hoc duplicate definitions of `getattr` and `dir` in `jsimport.c`, make a `JsProxy` wrapping `window` and then define `JsImport_GetAttr` and `JsImport_Dir` to be transparent passthroughs to `JsProxy_GetAttr` and `JsProxy_Dir`. (This current implementation of `JsImport_Dir` leaves out some fields that are on the actual `PyModule` object, but the previous implementation also leaves those out. 

I plan to rewrite this whole file anyways in #972, this is just the minimal change that fixes the bug.
Resolves #461.


dalcde:
Can you add a test for this?

hoodmane:
About to, just checking if they actually work first.

hoodmane:
I have a bunch of tests that only work with both this and #1124, so I have to sort through carefully to find the ones that test just this bug separately.

hoodmane:
Currently this always crashes on initialization.

hoodmane:
Test failures are expected: this won't work without #1124 being applied first. I guess I'll merge that here now so that the tests can pass.

hoodmane:
@rth @dalcde I added `pytest.mark.xfail` to the test to show that #261 has been resolved because this bug complex also interacts with #974 which I have a fix for in #1033 so I figure we can re enable the key word arguments test when we merge #1033.

dalcde (approved):

dalcde:
@hoodmane `test_jsproxy_call_kwargs` is currently failing

hoodmane:
Right this is because `JsBoundMethod_Call` does not implement the keyword arguments handling but this change makes it so that all functions are always wrapped in `JsBoundMethod` and never in `JsProxy`. Correct solution is to move `JsProxy_Vectorcall` and `JsProxy_New` to `JsBoundMethod` and also perhaps to rename `JsBoundMethod` to `JsMethod` or something.

hoodmane:
@dalcde correctly fixing `test_jsproxy_call_kwargs` is moderately complicated so I think the best thing is to `xfail` it here and open a second PR that fixes it.

hoodmane:
Funnily enough, the last attempt at this #461 ran into exactly the same problem that it broke `test_jsproxy_kwargs`. I guess it's not surprising.

dalcde:
>@dalcde correctly fixing `test_jsproxy_call_kwargs` is moderately complicated so I think the best thing is to `xfail` it here and open a second PR that fixes it.

It *was* working before this PR though

hoodmane:
> It *was* working before this PR though

Yes, do you prefer: 

* a moderate-complexity set of changes added to this PR, or
* same changes in a separate PR?

I guess I can just `git checkout -b` and open a separate PR with more changes on this same branch and then you can take your pick.

dalcde:
I would prefer a fix in a separate PR, assuming the fix doesn't depend on this. Then this can be merged after the fix is merged.

hoodmane:
The fix will break ~15 tests until this one is merged.

dalcde:
Then perhaps, as you said, a separate PR on top of this, and then I can merge them consecutively? I would very much prefer master to not regress, but it would be nice to logically separate the changes in history.

hoodmane:
Yeah, this is a pretty tricky tangle of changes. Will do that.

PR #1128, opened 2021-01-12
BACKEND: Python2js converter with depth param

hoodmane:
I added an extra `python2js_with_depth` function. When `depth=-1` (as in `depth=∞`), this behaves the same as the old `python2js` function. When `depth=0`, only "immutable" Python objects are converted, all others are proxied (we are slightly loose about what we consider to be "immutable" for practical reasons). When `depth=1`, one layer of conversion is performed unconditionally, further conversions are only applied if they would be applied with `depth=0`.

Currently, this function is only used with `depth=∞` so this PR should not change the user-observable behavior of pyodide.

This is work towards #900.


src/core/python2js.c
```diff
@@ -231,8 +231,17 @@ _python2js_dict(PyObject* x, PyObject* map)
   return jsdict;
 }
 
+#define RETURN_IF_SUCCEEDS(x)                                                  \
+  do {                                                                         \
+    JsRef fresh1278134_result = x;                                             \
```
phorward, line 236: Hi @hoodmane, this pull request looks good to me so far, except this part.
Can you please shortly explain why you used this strange variable name?
If this is for not colliding with other local variables, I would propose to use a Python-style variable name like __fresh_result here. What do you think?

phorward (changes requested):
Please clarify my comment shortly, for better understanding. Thanks!

src/core/python2js.c
hoodmane: I wasn't sure what to do here. Since `C` has no macro hygiene, there must be some standard thing to do. [The C standard 7.1.3](https://en.cppreference.com/w/c/language/identifier#Reserved_identifiers) forbids using names that start with `__` so I don't think `__fresh_result` is okay:

> The following identifiers are reserved and may not be declared in a program (doing so invokes undefined behavior): 
> ...
> 2) All external identifiers that begin with an underscore.
> 3) All identifiers that begin with an underscore followed by a capital letter or by another underscore (these reserved identifiers allow the library to use numerous behind-the-scenes non-external macros and functions).

(We currently are breaking rule 2 above because things like `_python2js_cache` are not declared static. I guess I can open a PR making them static.)

I think it's okay even without the fresh variable, the `do {}` block starts a scope and I think it works to say:
```C
int x = 7;
{
   int x = x;
}
```
And of course I only use this macro three times so it doesn't have to be absolutely nuclear bomb proof.

hoodmane: Actually the CPython source code doesn't care about that section of the standard so maybe it's not worth worrying about.

hoodmane: Okay I used `_fresh_result` (single leading underscore) since that is okay by the C standard.

hoodmane:
@rth @dalcde Would appreciate review on this one. Once this is merged, I can finally get back to implementing the most important part of #900.

phorward (approved):

hoodmane:
Or if you are willing to merge this @phorward that'd be great.

Thanks again for the reviews!

PR #1131, opened 2021-01-13
Simplify pythonexc2js and add more careful error handling

hoodmane:
I simplified the logic in `pythonexc2js` and implemented better error handling. I'm pretty sure that the original version actually always leaks the exception state because `PyErr_Fetch` takes ownership of `type`, `value`, and `traceback` but it never decref'd them. 

Another issue with the original code is that it calls `PyErr_Print()` when the error indicator is unset. The docs say that `PyErr_Print()` may cause a fatal system error if called when the error indicator is unset:
https://docs.python.org/3/c-api/exceptions.html?highlight=pyerr_print#c.PyErr_PrintEx
This is not apparently true, looking at the source code of `PyErr_Print`, there is a guard check whether the error indicator is set, if not just does nothing. (This is really what it should do, not like it costs much to be reasonable.) Anyways, because the documented behavior of `PyErr_Print` says it triggers a fatal error in this case, I figure we ought to try to avoid calling it in that condition.

In future PRs I am hoping to update the error formatting to alternate between Python and javascript stack traces as the code crosses between languages and to implement fatal error trapping.

### DIFF
--- a/src/core/python2js.c
+++ b/src/core/python2js.c
@@ -73,16 +73,11 @@ finally__skip_print_tb:
   Py_CLEAR(pylines);
   Py_CLEAR(empty);
   Py_CLEAR(pystr);
-  if (excval != NULL) {
-    // this throws an error making it pretty difficult to decref excval.
-    // hiwire_throw_error will decref it for us (so it steals a reference to its
-    // argument).
-    hiwire_throw_error(excval);
-  } else {
-    // In this case there will be an error in the calling code, calling code
-    // currently expects us to throw no matter what.
-    PySys_WriteStderr("Internal error: failed to generate exception!\n");
-  }
+  // hiwire_string_ascii never fails so excval is guaranteed not to be null at
+  // this point. This throws an error making it pretty difficult to decref
+  // excval, so hiwire_throw_error will decref it for us (in other words
+  // hiwire_throw_error steals a reference to its argument).
+  hiwire_throw_error(excval);
 }
 
 int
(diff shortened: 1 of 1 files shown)


src/core/python2js.c
```diff
@@ -11,83 +11,77 @@
...
+  FAIL_IF_NULL(pylines);
+  empty = PyUnicode_New(0, 0);
+  FAIL_IF_NULL(empty);
+  pystr = PyUnicode_Join(empty, pylines);
+  FAIL_IF_NULL(pystr);
+  const char* pystr_utf8 = PyUnicode_AsUTF8(pystr);
+  FAIL_IF_NULL(pystr_utf8);
+  PySys_WriteStderr("Python exception:\n");
+  PySys_WriteStderr("%s\n", pystr_utf8);
+  excval = _python2js_unicode(pystr);
```
dalcde, line 57: If I understand this correctly, if this line fails, then the error indicator will be set, but no error is thrown. This does not seem ideal to me. Is there any reason this is not a `FAIL_IF_NULL(excval)` and let the `!success` logic deal with it?

hoodmane: Good catch, I think that you do understand correctly and this is a logic error. Will fix!

src/core/python2js.c
```diff
@@ -11,83 +11,78 @@
...
-  if (format_exception == NULL) {
-    excval = hiwire_string_ascii("Could not get format_exception function");
+finally__skip_print_tb:
+  Py_CLEAR(type);
+  Py_CLEAR(value);
+  Py_CLEAR(traceback);
+  Py_CLEAR(pylines);
+  Py_CLEAR(empty);
+  Py_CLEAR(pystr);
+  if (excval != NULL) {
```
dalcde, line 76: I think excval is now always non-null? `hiwire_string_ascii` always succeeds

hoodmane: That's great since this function is supposed to have a "never returns" signature.

dalcde (approved):

PR #1132, opened 2021-01-13
Convert some package tests to use new test system

dalcde:
This is not exhaustive; I mostly picked the ones that can be converted reasonably programmatically.


rth (approved):
Thanks, it's much more readable indeed!

PR #463, opened 2019-06-12
Fix #461: Handle member functions in `from js import ...`

mdboom:

PR #1138, opened 2021-01-15
Move call and new to JsMethod

hoodmane:
The goal of this PR is to ensure that `JsProxy` is no longer callable, now only `JsMethod` will be callable. Many bugs were caused by differences in behavior between the `JsProxy_Call` and `JsBoundMethod_Call`. If we build a proxy of a callable javascript object, we will use the subclass `JsMethod`. This way, there is only one `Call` method and no bugs caused by divergent behaviors.

This is on top of #1126 and is required for #1126 to pass the `test_jsproxy_call_kwargs` test (the problem in #1126 is that only `JsProxy_call` supports key word arguments but the object being called is an instance of `JsBoundMethod_call`.

(Note that #463 makes similar changes to this PR and PR #1126 for similar reasons. Once this is merged, #461 and #463 can be closed.)


dalcde:
Should I merge this?

hoodmane:
Reviewing the diff just now I caught one minor mistake which I updated. If the diff looks okay to you, I think I'm happy with this. Presumably the plan is still to merge #1126 just before this?

dalcde:
yep

PR #1141, opened 2021-01-15
Correct repr in InteractiveConsole

casatir:
This should close #875.

Result representation is not satisfactory in `pyodide.console.InteractiveConsole` (and so in `console.html`). See below for an illustration.

```python
>>> 2 == 1+1  # this should be 'True' and not 'true'
true
>>> "foo"  # there should be quotes around foo
foo
>>> globals()
[object Object]
>>> {1: 2}
[object Object]
>>> 
```

Moreover, the traceback is not satisfactory as it shows the underlying evaluation pipeline.

```python
>>> 1 / 0
Python exception:
  File "/lib/python3.8/site-packages/pyodide/_base.py", line 303, in eval_code
    return CodeRunner(
  File "/lib/python3.8/site-packages/pyodide/_base.py", line 229, in run
    return eval(last_expr, self.globals, self.locals)
```

Note that #1125 accidentally introduces two bugs: errors are shown twice and syntax errors are not shown.

All this should be corrected with this PR. To achieve nice result representation, the interactive console is fully implemented in Python. This has a major drawback: we can't use `eval_code` alone, since packages should be loaded. The `InteractiveConsole`'s `runcode` method then returns a promise.

Warning: it actually does not use `eval_code` but the CPython's `code.InteractiveConsole` evaluation system. This mean the console no more reflects the `eval_code` behavior but the CPython's one. For sure we should discuss this since you may disagree. It is of course possible to internally use `eval_code` but it will require more work.

Unfortunately, I was not able to use the (great) new test system with `run_in_pyodide` since I needed to access the selenium driver to wait for async eval behavior implied by package load.


src/templates/console.html
```diff
@@ -19,74 +19,79 @@
...
+            import builtins
             import js
+            from reprlib import Repr
+            from pyodide.console import InteractiveConsole
 
 
-            class Console(console.InteractiveConsole):
+            # limiting output size to 1000 to avoid console hang
+            _repr = Repr()
+            _repr.maxdict = _repr.maxlist = _repr.maxtuple = _repr.maxset = _repr.maxfrozenset = _repr.maxdeque = _repr.maxarray = _repr.maxlong = _repr.maxstring = _repr.maxother = 10 ** 3
```
rth, line 31: I think 1000 elements might be too much, particularly for arrays represented as lists. It will still be very slow in the console.

Why not use values closer to [replib](https://docs.python.org/3/library/reprlib.html) defaults?

src/templates/console.html
```diff
@@ -19,74 +19,79 @@
...
+                        text = bytes.decode(sys.stdout.encoding, 'strict')
+                        sys.stdout.write(text)
+                sys.stdout.write("\\n")
+                builtins._ = value
+
+            sys.displayhook = displayhook
+
+
+            class PyConsole(InteractiveConsole):
+                def __init__(self):
```
rth, line 55: I think we could still put this files in `pyodide-py/pyodide/console.py` to benefit from code formatting, etc. Just making them private with a leading `_` and indicate that they are used in this file should be enough?

src/pyodide-py/pyodide/console.py
```diff
@@ -134,14 +145,34 @@ def restore_stdstreams(self):
...
+        if not self._persistent_stream_redirection:
             self.restore_stdstreams()
 
+    def runsource(self, *args, **kwargs):
+        # syntax errors are not catched at runcode level but at runsource
+        with self.stdstreams_redirections():
+            return super().runsource(*args, **kwargs)
+
+    def runcode(self, code):
+        """ Load imported packages then run code, async. """
```
rth, line 163: This and and the above method would need a more detailed description in the docstring with some motivation for their existence (that you wrote in the issue description)

src/templates/console.html
hoodmane: Another option is to compute the entire repr and test the length of the output, then only take the beginning and end of the output, I've seen this done in some places. On my screen `console.html` is 34 lines high when it's full screen and at normal zoom. If we assume 34 * 80 characters wide, then we can fit 2720 characters per screen. Maybe we are targeting around 70% of that, so limit the output to 2000 characters? 

`reprlib` can both undershoot and overshoot this. For instance the following doesn't hang the console but it does generate 42000 characters out output, so ~20 screens worth of text:
```python
r = []
for i in range(5):
   r = [r,]*8
reprlib.repr(r)
```
On the other hand, by default `reprlib` only lists the first five elements of a 1d list, which is unacceptably stingy IMO (What if you want to see the first 10 elements to find a pattern? Sure you can explicitly ask for more to be printed, but I think this would be annoying.).

Also, the current console does not limit the width of the output region to 80 characters wide. On my screen in firefox with 100% zoom the console wraps at 140 characters.

src/templates/console.html
hoodmane: Only disadvantage of that is that increases the pyodide download size even for people who don't use the code. Probably not by enough to worry about though.

hoodmane:
> Note that #1125 accidentally introduces two bugs: errors are shown twice and syntax errors are not shown.

I do not think that "errors are shown twice" should exactly be considered to be a bug introduced by changes in #1125, at least it doesn't happen because of any mistakes with #1125. In `pythonexc2js`, the current behavior is that it throws a javascript error with a Python stack trace *and* writes the error to stderr. I think this is an incorrect behavior in `pythonexc2js` that your changes in #1125 merely exposed.

src/templates/console.html
rth: Well, right now the minimal download size for pyodide is 6.5MB, compressed, this code snippet is 0.8kB compressed. So I think there lots of other things to optimize before we need to worry about it..

hoodmane: Absolutely, I realize that was a bit of a silly remark.

src/templates/console.html
rth: > Another option is to compute the entire repr and test the length of the output, then only take the beginning and end of the output

Maybe that would be easier indeed. As right if you have a list of a list of a dict, it's indeed hard to tell how long the final displayed output will be..

> Also, the current console does not limit the width of the output region to 80 characters wide. On my screen in firefox with 100% zoom the console wraps at 140 characters.

As far as I can tell the the behavior is identical to the classical Python REPL, in that the width limit is given by the window size? The only difference I see is that say when you print a very long list of str, Python REPL will not break strings in the middle, while this console will. But maybe we could look into it in a separate issue..

hoodmane: >> does not limit width
>
> maybe we could look into it in a separate issue..

Agreed.

casatir: Great, I'll do

```python
# limiting output size to REPR_LIMIT to avoid console hang
REPR_LIMIT = 1000
REPR_SPLIT = REPR_LIMIT // 4
def custom_repr(value):
    text = repr(value)
    if len(text) > REPR_LIMIT:
        text = f"{text[:REPR_SPLIT]}...{text[-REPR_SPLIT:]}"
    return text
```

This will avoid `repr` longer than 1000 and replace them by 503 character-long representation. Fill free to comment on the value I choose for `REPR_LIMIT` and `REPR_SPLIT`.

casatir:
Recent commits addresses:
 - banner issue from [this comment](https://github.com/iodide-project/pyodide/pull/1125#discussion_r558493515)
 - Docstrings from @rth review
 - support for multi-line pasting
 - size-limiting system integrated to `console.py`

̀`console.html` seems sufficiently small to be kept here and illustrate the simplicity of integrating a Pyodide console in an html page if users look at the code. WDYT?

I'll investigate the testing fail... it passes on my machine when run inside docker and after a complete rebuilt with up to date upstream base...

hoodmane:
> I'll investigate the testing fail... it passes on my machine when run inside docker and after a complete rebuilt with up to date upstream base...

I think it could help to make a promise for the repl to be completed / in "ready state" and await on the promise between `shell.push` and looking at the result. This would help to distinguish between the website crashing in the test for some reason, or other sorts of timeouts versus just making a mistake in the result of the test.

hoodmane:
Also, I think it would be good to add some multiline tests like:
```python
def f(x):
   return x*x + 1
[f(x) for x in range(10)]
```

src/templates/console.html
hoodmane: Those look like reasonable values, though I think I'd set `REPR_SPLIT = REPR_LIMIT//2`? That way if it is 1010 characters it gets reduced to 1000 ish characters, it seems a little weird that if the output is 1000 characters the output is 1000 characters long but with 1001 characters it gets reduced to 500 ish characters.

It might also be good to do something like:

`text = f"{text[:REPR_SPLIT]}\n<<<<Long repl output truncated>>>>\n{text[-REPR_SPLIT:]}"` or something like that to make it a little to tell that the output has been truncated. Otherwise, the `...` will just appear in some random spot in the screen and probably not be all that noticeable. (If it's possible to make `<<<<Long repl output truncated>>>>` or whatever exactly it says a different color I think that'd be even better.)

casatir:
> I think it could help to make a promise for the repl to be completed / in "ready state" and await on the promise between shell.push and looking at the result. This would help to distinguish between the website crashing in the test for some reason, or other sorts of timeouts versus just making a mistake in the result of the test.

This is exactly what I am looking for. The promise already exists, it's `shell.run_complete`, but how do you await for a JS promise in the outer Python testing context?

> Also, I think it would be good to add some multiline tests

I'll add some.

src/templates/console.html
casatir: > it seems a little weird that if the output is 1000 characters the output is 1000 characters long but with 1001 characters it gets reduced to 500 ish characters.

That's what `numpy` do. I get full output for `np.array(range(1000))` and

```python
>>> np.array(range(1001))
array([   0,    1,    2, ...,  998,  999, 1000])
```

but I'm OK with what you say.

> It might also be good to do something like:

I'll try to add something like that.

casatir:
I think I have something for the test failure. Packages are not available at the time of core tests right? This is why it fails when importing `numpy`... Any idea for a workaround to test non-core package load?

dalcde:
On Fri, Jan 15, 2021 at 05:26:13PM -0800, casatir wrote:
>I think I have something for the test failure. Packages are not available at the time of core tests right? This is why it fails when importing `numpy`... Any idea for a workaround to test non-core package load?

Some packages are available at the time of core tests, explicitly for
the purposes of such testing. These are micropip, distlib, pyparsing,
pytz, packaging and kiwisolver.

casatir:
Great! Thanks @dalcde!

hoodmane:
@dalcde I'm pretty sure you also know the answer to another question @casatir asked:

>  how do you await for a JS promise in the outer Python testing context?

dalcde:
selenium has an `execute_async_script` function, which runs a javascript snippet and blocks until a callback is called. For example, to evaluate a promise, you can do
```
selenium.driver.execute_async_script("""
let cb = arguments[arguments.length - 1]; // This is where you find the callback function
promise.then(cb)
""")
```

casatir:
Another big thank to @dalcde. Recent commits addresses:

 - long result splitting suggestions by @hoodmane 
 - test failure
 - more tests
 - merge into master

src/pyodide-py/pyodide/console.py
```diff
@@ -1,8 +1,40 @@
...
+# this import can fail when we are outside a browser (e.g. for tests)
+try:
+    import js
+
+    dummy_promise = js.Promise.resolve()
+    load_packages_from_imports = js.pyodide.loadPackagesFromImports
+
+except ImportError:
+
+    class FakePromise:
```
rth, line 18: This doesn't need to be part of Public API? If so let's rename it,
```suggestion
    class _FakePromise:
```

src/pyodide-py/pyodide/console.py
```diff
@@ -1,8 +1,40 @@
...
+
+        def __init__(self, args=None):
+            self.args = (args,) if args is not None else ()
+
+        def then(self, func, *args):
+            return FakePromise(func(*self.args))
+
+    dummy_promise = FakePromise()
+
+    def load_packages_from_imports(*args):
```
rth, line 36: Same I'm not sure if this needs to be public for now, otherwise,
```suggestion
    def _load_packages_from_imports(*args):
```

src/pyodide-py/pyodide/console.py
```diff
@@ -150,4 +218,49 @@ def banner(self):
...
+    """A displayhook with custom `repr` function.
+
+    It is intendend to overload `sys.displayhook`. Note that monkeypatch
+    `builtins.repr` does not work in `sys.displayhook`. The pointer to
+    `repr` seems hardcoded in default `sys.displayhook` version
+    (which is written in C)."""
+    # from https://docs.python.org/3/library/sys.html#sys.displayhook
+    if value is None:
+        return
+    builtins._ = None  # type: ignore
```
rth, line 254: This needs an inline comment on why we need it.

src/pyodide-py/pyodide/console.py
```diff
@@ -150,4 +218,49 @@ def banner(self):
...
+    text = repr(value)
+    try:
+        sys.stdout.write(text)
+    except UnicodeEncodeError:
+        bytes = text.encode(sys.stdout.encoding, "backslashreplace")
+        if hasattr(sys.stdout, "buffer"):
```
rth, line 260: Maybe we could add an inline comment in which cases `sys.stdout` would have or not have a `buffer`.

src/pyodide-py/pyodide/console.py
```diff
@@ -150,4 +218,49 @@ def banner(self):
         """ A banner similar to the one printed by the real Python interpreter. """
         # copyied from https://github.com/python/cpython/blob/799f8489d418b7f9207d333eac38214931bd7dcc/Lib/code.py#L214
         cprt = 'Type "help", "copyright", "credits" or "license" for more information.'
-        return f"Python {platform.python_version()} {platform.python_build()} on WebAssembly VM\n{cprt}"
+        version = platform.python_version()
+        build = f"({', '.join(platform.python_build())})"
+        return f"Python {version} {build} on WebAssembly VM\n{cprt}"
+
+
+def repr_abreviate(
```
rth, line 226: I think `abreviate`might not be the best term. Maybe either `repr_shorten` (there is a https://docs.python.org/3.4/library/textwrap.html#textwrap.shorten which is very similar) or `repr_truncate` ?

rth (commented):
Thanks! Generally LGTM, but I think we need to be more careful about the separation of public and private API for Python classes/functions (also in other unrelated PRs).
 - public functions (without leading `_` in the name), should have a docstring (even if it's an import from `js`), be added to the documentation list in  https://github.com/iodide-project/pyodide/blob/master/docs/api_reference.md#python-api, and be in `__all__`. For any non backward change we would have to add a 'breaking change' changelog entry.
So that's why by default it might be better to have objects private (with a leading `_`) unless the use case for the users is very strong and we make some commitments that we won't change the function in a backward compatible way in the near future. Users can still use private functions, it just means that "this may change in the future version without warnings".

In the case of the console module, I think it might be good to have
 - Private: `_dummy_promise`, `_FakePromise`, `_load_packages_from_imports` (we might make it public at some point, but it would need docstrings and maybe let's not do it in this PR), `_StdStream`,  `_CallbackBuffer`
 -  Public: `InteractiveConsole, `repr_abreviate` (or the other name we give it cf my comment below), `displayhook` (it would be useful for it to be public, right?)

src/tests/test_console.py
```diff
@@ -94,3 +99,84 @@ def stderr_callback(string):
...
+
+
+def test_repr(safe_sys_redirections):
+    assert len(console.repr_abreviate("x" * 10 ** 5, 100)) <= 110
+
+    sys.stdout = io.StringIO()
+    console.displayhook([0] * 100, lambda v: console.repr_abreviate(v, 100))
+    assert len(sys.stdout.getvalue()) <= 110
+
+
```
rth, line 121: It would be good to add tests not using pyodide for `repr_abreviate` (or the other name we give it). In particular, with a small `limit` (say around 10) checking the output is correct for odd and even combination of input lenght and limit values.

src/pyodide-py/pyodide/console.py
casatir: You're right. I'll make it private.

src/pyodide-py/pyodide/console.py
casatir: OK.

src/pyodide-py/pyodide/console.py
casatir: `repr_shorten` looks good to me.

src/pyodide-py/pyodide/console.py
casatir: You mean a comment on why we need displayhook? In the docstring?

casatir:
Thanks @rth for the review. I agree with your comments and added modifications you suggested. Can you precise the comment you are looking for in `displayhook`?

casatir:
>Warning: it actually does not use eval_code but the CPython's code.InteractiveConsole evaluation system. This mean the console no more reflects the eval_code behavior but the CPython's one. For sure we should discuss this since you may disagree. It is of course possible to internally use eval_code but it will require more work.

What about this? Is it something you want to be addressed in another PR? In this one? Or is this behavior OK for you guys?

hoodmane:
> Warning: it actually does not use eval_code but the CPython's code.InteractiveConsole evaluation system. This mean the console no more reflects the eval_code behavior but the CPython's one. For sure we should discuss this since you may disagree. It is of course possible to internally use eval_code but it will require more work.

In this case the behavior is more important than the implementation. If you think that the `InteractiveConsole` behavior is satisfactory, I think it's probably fine to use it. No reason you have to use the same entrypoint for the console as we defined for general purpose use.

casatir:
> In this case the behavior is more important than the implementation. If you think that the InteractiveConsole behavior is satisfactory, I think it's probably fine to use it. No reason you have to use the same entrypoint for the console as we defined for general purpose use.

Totally agree with this. This is the way I see it. But if you want to use `console.html` for testing (as a dev tool, as we already discuss), it might disturb if the REPL engine is not our general purpose entry point.

`InteractiveConsole` behavior is satisfactory to me as it is really close to the CPython's REPL.

It still lacks completion (another PR) and I am fighting with trailing newlines in JQuery terminal but can't get satisfaction with `echo_newline.js` as advertised [here](https://terminal.jcubic.pl/api_reference.php#echo) and [there](https://stackoverflow.com/questions/57289511/jquery-terminal-how-to-echo-without-line-break/57715395#57715395).

hoodmane:
For debugging, the main important thing is to be able to read logging messages, so I care mainly about stdout / stderr redirection. If I don't like the execution behavior, I can just use `runPython` from the console.

src/pyodide-py/pyodide/console.py
rth: No I mean why we need to manually change `builtins._` here and the end of the function. Adding an inline comment in the code would help the reader...

casatir: OK. I'll so this.

src/pyodide-py/pyodide/console.py
casatir: It should not happen with our `_StdStream` but the user can overload them.

casatir:
`console.html` now have correct `sys.stdout` behavior regarding the newline issue added by JQuery terminal. The issue is still present on `sys.stderr` but I think it's not really an issue. This kind of examples now works correctly.

```python
>>> for c in 'foo':
...     print(c, end='', flush=True)
... 
foo>>> 
 ```

**Comment by casatir** [comment]
In CPython's REPL,

```python
>>> for c in 'foo':
...     print(c, end='')
... 
foo>>> 
```

In our `console.html`:

```python
>>> for c in 'foo':
...     print(c, end='')
... 
>>> 
```

(look at the 'foo' not displayed) My investigations shows that the flush is triggered by the `input` prompt. This is why it is not triggered in our case since we don't use `input`. I corrected this by forcing flush after each computation.

rth (approved):

PR #1142, opened 2021-01-16
Fine tuning of error_handling debug tools

hoodmane:
I fixed a typo in `EM_JS_REF` and `EM_JS_NUM`: namely `LOG_EM_JS_ERROR(func_name, err);` ==> `LOG_EM_JS_ERROR(func_name, e);`

Also added an extra function `log_error_obj` which takes a `JsRef` and does `console.error(jsobj)`, useful for debugging.

### DIFF
--- a/src/core/error_handling.h
+++ b/src/core/error_handling.h
@@ -15,6 +15,8 @@ error_handling_init();
 errcode
 log_error(char* msg);
 
+// Right now this is dead code (probably), please don't remove it.
+// Intended for debugging purposes.
 errcode
 log_error_obj(JsRef obj);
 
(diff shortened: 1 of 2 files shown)


src/core/error_handling.h
```diff
@@ -7,13 +7,17 @@
...
 typedef int errcode;
+#include "hiwire.h"
 
 int
 error_handling_init();
 
 errcode
 log_error(char* msg);
 
+errcode
```
dalcde, line 18: Can you add a comment indicating that this is supposed to be not used in any existing code?

hoodmane: You mean like something like "This is a debug tool, it's probably dead code but please don't remove it"?

dalcde: Yeah. Explain that it is intentionally dead, not accidentally dea.

hoodmane: Okay, will do.

dalcde (approved):

PR #1143, opened 2021-01-16
TST: Combine run_js_async and run_js

dalcde:
run_js_async works just as well for synchronous code, and the code is mostly duplicated.

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -81,7 +81,7 @@ class SeleniumWrapper:
             )
         self.driver.get(f"http://{server_hostname}:{server_port}/test.html")
         self.run_js("Error.stackTraceLimit = Infinity")
-        self.run_js_async("await languagePluginLoader")
+        self.run_js("await languagePluginLoader")
 
     @property
     def logs(self):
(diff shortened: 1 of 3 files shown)


dalcde:
@hoodmane do you mind giving the failed test in the first commit (and the fix in the second) a quick look? I think it is a selenium problem rather than a jsproxy problem

hoodmane:
I looked at the test failure, don't really know what to make of it since I haven't see that test ever come up as the simplest failure. It's failing to import `window.ImageData` even though in other tests it successfully imported all sorts of stuff `from js` elsewhere, that particular behavior has lots of test coverage. So probably as you say it's some sporadic failure in selenium?

Note that here again there is a stack trace that looks useful but it got truncated:
`"message":"Error: Accessing TypedArray data over Xrays is ...se@chrome://marionette/content/listener.js:697:26\\ndispatch/</<@chrome://marionette/content/listener.js:533:31\\n"`

hoodmane:
Actually, "Xrays" seems like an unusual term in that stack trace. Googling "Accessing TypedArray data over Xrays" shows that the full message is:

"Error: Accessing TypedArray data over Xrays is slow, and forbidden in order to encourage performant code. To copy TypedArrays across origin boundaries, consider using Components.utils.cloneInto()."

dalcde:
On Fri, Jan 15, 2021 at 06:46:47PM -0800, Hood Chatham wrote:
>I looked at the test failure, don't really know what to make of it since I haven't see that test ever come up as the simplest failure. It's failing to import `window.ImageData` even though in other tests it successfully imported all sorts of stuff `from js` elsewhere, that particular behavior has lots of test coverage. So probably as you say it's some sporadic failure in selenium?
>
>Note that here again there is a stack trace that looks useful but it got truncated:
>`"message":"Error: Accessing TypedArray data over Xrays is ...se@chrome://marionette/content/listener.js:697:26\\ndispatch/</<@chrome://marionette/content/listener.js:533:31\\n"`

I'm fairly sure it is not sporadic. I'd reckon it's got to do with
passing the ImageData object from pyodide to javascript back to
selenium's python.

hoodmane:
Yeah issues like that cause a lot of issues in the tests. Maybe it would be a good idea to update `selenium.run` to defensively throw away things that look like they won't serialize correctly? That's been creating problems for me on #1098, maybe this is a related bug.

hoodmane:
Certainly the tests failing on #1098 would be fixed by throwing away the result.

rth (approved):
Thanks!

PR #1144, opened 2021-01-16
Don't print duplicate error message in chrome

dalcde:
In chrome, the error message is contained in the stack. Avoid printing it twice.

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -52,12 +52,12 @@ class JavascriptException(Exception):
     def __init__(self, msg, stack):
         self.msg = msg
         self.stack = stack
+        # In chrome the stack contains the message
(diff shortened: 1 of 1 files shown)


phorward (approved):
This looks okay. Did you also test it in Firefox?

dalcde:
On Sat, Jan 16, 2021 at 02:36:16AM -0800, Jan Max Meyer wrote:
>@phorward approved this pull request.
>
>This looks okay. Did you also test it in Firefox?

It works well in firefox, where the error message is not part of the
stack trace. The code is written so that we would never lose information
if the formats change, since we only hide `self.msg` if it is a
substring of `self.stack`.

PR #1145, opened 2021-01-16
Lower selenium timeout to 10s

dalcde:
It used to be 30s, and can cause long delays when some tests hang.

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -70,7 +70,7 @@ class SeleniumWrapper:
             build_dir = BUILD_PATH
 
         self.driver = self.get_driver()
-        self.driver.set_script_timeout(10)
+        self.driver.set_script_timeout(15)
         self.server_port = server_port
         self.server_hostname = server_hostname
         self.server_log = server_log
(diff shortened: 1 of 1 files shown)


dalcde:
I wonder if I'm going about this the wrong way; I should *increase* the timeout so that tests don't get killed

phorward:
> I wonder if I'm going about this the wrong way; I should _increase_ the timeout so that tests don't get killed

I'm of the same opinion. Maybe this 10sec limit can be used in combination with a retry option, so that flacky tests failing due timeout are simple restarted again. This might also reduce the number of (often) wrongly indicated, failed tests.

phorward:
@dalcde maybe do some more tests with higher timeout to get rid of the flacky builds?

dalcde:
On Sat, Jan 16, 2021 at 06:52:58AM -0800, Jan Max Meyer wrote:
>@dalcde maybe do some more tests with higher timeout to get rid of the flacky builds?

Currently we retry all failed tests, but it's hard to test if something
does get rid of flaky builds by its very nature.

I've considered keep retrying tests until the number of failed tests
becomes constant, which should keep most flakiness out.

PR #1146, opened 2021-01-16
Flexible jsimport take two

### LINKED ISSUE #960: Configurable "import js"

hoodmane:
Now that the tricky #461 / #788 bug complex has been resolved, there is a much easier implementation for "flexible js import". This PR resolves #960.

I added two new APIS to the `pyodide-js` module: `pyodide.registerPackage(name, object)` and `pyodide.unregisterPackage(name)`. `pyodide.registerPackage` adds a new javascript object to the import path, while `pyodide.unregisterPackage` will remove the javascript object from the import path (though if the object has already been imported, it must also be removed from `sys.modules` as well to actually remove it).


src/pyodide-py/pyodide/_importhooks.py
```diff
@@ -0,0 +1,80 @@
...
+    @classmethod
+    def find_module(cls, fullname, path):
+        raise NotImplementedError(
+            "find_module() is deprecated in favor of find_spec() since Python 3.4"
+        )
+
+    @staticmethod
+    def invalidate_caches() -> None:
+        pass
+
```
dalcde, line 55: I believe both of these can be left out; they have default implementations: https://github.com/python/cpython/blob/3.8/Lib/importlib/abc.py
```suggestion
```

src/pyodide-py/pyodide/_importhooks.py
```diff
@@ -0,0 +1,80 @@
...
+    # that short circuits the file system access stuff.
+    # create_module_inner defined in jsimport.c
+    def create_module(self, spec):
+        return self.jsproxy
+
+    # use ExtensionFileLoader.exec_module (no override)
+
+    # used by importlib.util.spec_from_loader
+    def is_package(self, fullname):
+        return True  # ???
```
dalcde, line 67: What is the `???` issue here?

src/tests/test_jsproxy.py
```diff
@@ -471,3 +471,123 @@ def test_window_isnt_super_weird_anymore():
...
+        try {
+            pyodide.unregisterJsModule("doesnotexist");
+            throw new Error("unregisterJsModule should have thrown an error.");
+        } catch(e){
+            if(!e.message.includes("Cannot unregister 'doesnotexist': no javacript module with that name is registered")){
+                throw e;
+            }
+        }
+        """
+    )
```
dalcde, line 581: Split this into a separate test?

dalcde (commented):
I'm a huge fan of a python-based implementation. I've left a few minor comments.

src/pyodide-py/pyodide/_importhooks.py
hoodmane: I think it's right to say that these things are a `package`, I am just a little bit unclear on what a package is. I think it is just an issue of whether the syntax:
```python
from a.b import c`
```
is supported? If `a` is not a package, then this will raise an error? In that case I think `return True` is the correct behavior.

dalcde: That is my understanding too.

src/pyodide-py/pyodide/_importhooks.py
```diff
@@ -0,0 +1,80 @@
...
+        )
+
+    @staticmethod
+    def invalidate_caches() -> None:
+        pass
+
+    ### Loader methods
+    # Overwrite ExtensionFileLoader.create_module with our own mechanism
+    # that short circuits the file system access stuff.
+    # create_module_inner defined in jsimport.c
```
dalcde, line 59: I suspect this comment is outdated

hoodmane: Yup. A lot of stuff was outdated, I just cleaned up a bit.

dalcde:
I think this is good to go after some documentation and changelog. (also, your description in the OP refers to the old names of the new API)

hoodmane:
> I think this is good to go after some documentation and changelog.

So for docs `registerJsModule` / `unregisterJsModule` need to go in Js API, `register_js_module` and `unregister_js_module` need to go in Python API. Do you think I should add anything about these in the "FAQ" or other sections?

> your description in the OP refers to the old names of the new API

I edited the description, thanks.

dalcde:
An FAQ entry would be nice, I think. "How do I create custom python packages from javascript"

hoodmane:
Okay, I added some docs. Let me know if you have any complaints.
Before the next release it would probably be good to do some maintenance on the docs.

PR #1148, opened 2021-01-17
Patch ceval to add callback to use for KeyboardInterrupt handling

hoodmane:
This is a new attempt to integrate keyboard interrupt handling into Pyodide. I investigated the three existing CPython ceval hooks that I found: the `SetTrace` / `SetProfile` hooks, the `PendingCall` hook, or the JIT Frame evaluator hook. The JIT Frame evaluator hook doesn't work at all, the `PendingCall` hook is slow (`PendingCall`s are considered to be rare events), and `SetTrace` / `SetProfile` either screw up `sys.settrace` / `sys.setprofile` or require duplicating the current Python version of `sys.settrace` / `sys.setprofile`. I think they are probably also slower than this approach (they certainly enable more per-instruction conditionals).

Anyways, this patches in a new callback called `pyodide_callback`, which is intended to handle keyboard interrupts. The idea is that we will use this with pyodide running in a webworker to check a one-byte `SharedArrayBuffer` to see whether the main thread has requested a `KeyboardInterrupt` by setting a value into the `SharedArrayBuffer`.


hoodmane:
Here is the explanation I put in the patch file:

This patch adds a callback called `pyodide_callback` with signature 
`int callback(void)` to the main loop in `ceval.c`. This function gets called once 
per opcode except after opcodes `SETUP_FINALLY`, `SETUP_WITH`, `BEFORE_ASYNC_WITH`,
and `YIELD_FROM`. The main loop normally prevents signal handling, etc from happening
after these instructions, so I figured this should apply to my callback too.
(There is an extra layer of protection here though because in the callback we use 
`PyErr_SetInterrupt` which simulates a SIGINT signal and triggers the KeyboardInterrupt
via the standard mechanism.)

Note that we call the callback outside of the normal 
`if (_Py_atomic_load_relaxed(eval_breaker))` 
block where most of the "periodic things" happen. This is because normally when a 
"periodic thing" is queued, the `eval_breaker` flag is set signalling the need for
handling. The whole point of this patch though is that we need to be able to set an
interrupt from a remote thread and the `eval_breaker` flag lives on the WASM heap.
Unless the whole WASM heap is made into a `SharedArrayBuffer` and shared between 
webworkers, we have no way to set the `eval_breaker` flag. We still want to skip
the callback after the sensitive opcodes `SETUP_FINALLY`, `SETUP_WITH`, 
`BEFORE_ASYNC_WITH`, and `YIELD_FROM`, so we hoisted the check for that condition
out of the `eval_breaker` conditional.

We also patched the `threadstate` struct to include a `pyodide_callback` field, patch
`new_threadstate` to initialize the `pyodide_callback` field to `NULL`, and add a new
API `PyPyodide_SetPyodideCallback` to set `pyodide_callback`. This makes 
`PyPyodide_SetPyodideCallback` behave much like `PyEval_SetTrace`, except lighter 
weight.

phorward (approved):
Hello this looks good to me so far. I fixed and sorted the Makefile a little bit to avoid further conflicts in future. I'm not sure why the document tests failed, maybe the problem is temporary.

hoodmane:
Thanks @phorward. 

I suspect the test failure is temporary since nothing you did should have affected the docs build. I think you can rerun the test if you go to the circleci webpage and hit the "rerun" button. I don't have permission to do this though, I can only rerun the full suite.

rth:
Thanks @phorward ! It's also failing on master, so I don't think it's temporary. Probably some doc related dependency released a new version with breaking changes..

phorward:
> Thanks @phorward ! It's also failing on master, so I don't think it's temporary. Probably some doc related dependency released a new version with breaking changes..

Thanks @rth, I've seen this also in the latest builds and rebuilt also failed. I'm not familiar with Sphinx issues, so someone else should take a look on it, or I need some time.

hoodmane:
myst-parser had a new release 11 hours ago (version 0.13.2) and this is what's causing the breakage:

https://pypi.org/project/myst-parser/#history

rth:
Yes, should be fixed in https://github.com/iodide-project/pyodide/pull/1157

rth:
This would need a changelog entry, could you add it please @hoodmane ?

hoodmane:
Should I add it as a separate PR? Also, the pyodide-interrupts package could be removed now since this does the same thing in a more polished way.

rth:
yes, please.

hoodmane:
It's worth noting that at this point I haven't exported `setInterruptBuffer` as a public function so this PR didn't change the publicly exported behavior it just adds an unexported backend feature. We could decide to not export `setInterruptBuffer` at all and only include it as part of a higher level webworker API. I guess this would be a bad choice though, probably a high-level webworker API should preferably be built entirely on lower level public abstractions in case someone doesn't like the API we expose and wants to make their own.

I guess I will add `setInterruptBuffer` to the `PUBLIC_API` and update the changelog accordingly.

hoodmane:
Okay I did this in #1173.

PR #1149, opened 2021-01-17
DOCS: Move docs into three subfolders based on top level sections

hoodmane:
I think the current organization of the docs is a bit cluttered: there are 15 .md files in the top level of the docs directory. I made three new folders: "usage", "development", and "project" reflecting the top level subdivisions in the docs and moved the files constituting each of these top level sections into the appropriate folder.


rth:
I was also thinking we should do it. The only thing is it will break existing links to the documentation. It's probably still worth doing, but we need to be sure we won't change it again soon.

Maybe `user-guide` instead of `usage`? Also why we are at it, maybe rename `_` to `-` in paths which is a more common separator for spaces in URLs?

hoodmane:
> The only thing is it will break existing links to the documentation.

Good point hadn't thought of that.

> Maybe `user-guide` instead of `usage`? Also why we are at it, maybe rename `_` to `-` in paths which is a more common separator for spaces in URLs?

Both changes sound fine to me. Do you want to also change the section title to `user-guide`?

rth:
> Do you want to also change the section title to User Guide?

Sounds good. 

Also while we are at it, I'm not very fond of the URL `/using_pyodide_from_javascript.html` maybe we could call it `user-guide/quickstart.html` or something more general? Since it's about how to use pyodide in general. And then for webworkers, `user-guide/webworker.html` . I don't think that `using_pyodide_from_*` belongs in those URLs since there is already pyodide in the subdomain name. But then maybe `user-guide/webworker.html` is not specific enough? Not sure. Naming is hard.  

WDYT?

hoodmane:
I don't like `using_pyodide_from_javascript.html` either. I think `quickstart.html` and `webworker.html` are good: what else would `webworker.html` tell you about than using pyodide in a webworker?

> Naming is hard.

Amazingly so. They sometimes call it "bikeshedding" when people spend too much energy naming things, but names are important engineering decisions.

hoodmane:
So in `usage/python-api`, the generated files have names `pyodide.<API name>.rst`. In `usage/js-api`, we hand write files with names `pyodide_<API name>.md`. Maybe switch `pyodide_` for `pyodide.` in our handwritten files for consistency?

rth:
> Maybe switch pyodide_ for pyodide. in our handwritten files for consistency?

Sounds good.

hoodmane:
Another question: we have been speaking of `pyodide_py` and `pyodide_js` to distinguish between the python and javascript pyodide modules. We could potentially use these in the docs. I'm not sure whether it is a good or not. Any thoughts?

rth (approved):
Thanks, looks good.

>  we have been speaking of pyodide_py and pyodide_js to distinguish between the python and javascript pyodide modules. We could potentially use these in the docs. I'm not sure whether it is a good or not. Any thoughts?

Potentially we could, it's shorter but maybe Python / JS pyodide package is less confusing, as `pyodide_py` is still used as  `pyodide` imports etc. I think PR this has enough changes to be merged separately in any case.

PR #1150, opened 2021-01-18
Add more noreturn markers

hoodmane:
A couple commits back I added a `_Py_NO_RETURN` marker to `pythonexc2js`, but I didn't add `_Py_NO_RETURN` to `hiwire_throw_error`, which made the compiler emit a warning. This adds a `_Py_NO_RETURN` marker to `hiwire_throw_error` suppressing that warning.

### DIFF
--- a/src/core/hiwire.c
+++ b/src/core/hiwire.c
@@ -222,7 +222,7 @@ EM_JS_REF(JsRef, hiwire_float64array, (f64 * ptr, int len), {
   return Module.hiwire.new_value(array);
 })
 
-EM_JS(void, hiwire_throw_error, (JsRef idmsg), {
+EM_JS(void _Py_NO_RETURN, hiwire_throw_error, (JsRef idmsg), {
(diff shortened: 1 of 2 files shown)


rth:
I saw https://github.com/python/cpython/blob/b3b98082c5431e77c64cab2c85525a804436b505/Include/pyport.h#L857 but could please you explain (or link to your previous explanation) why we need it? Was it for the performance impact?

hoodmane:
It does enable extra performance optimizations by the compiler but I don't imagine they'd make a measurable difference. These functions have as a part of their contract that they don't return to their calling C function though, instead triggering Javascript stack unwinding out to the nearest enclosing javascript `try` block. The main reasons I added these markers are to make the code easier to read by explicitly indicating this `noreturn` contract and to trigger compiler warnings if the function is later modified to return.

rth (approved):
Sounds good, thanks for the explanations!

Issue #951, opened 2020-12-26
Failed to Capture Recursion Error


oeway:
Related to https://github.com/iodide-project/pyodide/issues/821, I noticed that we cannot capture the recursion error.

Here is what I got in https://pyodide-cdn2.iodide.io/v0.16.1/full/console.html:
```py
>>> def f(): return 1 + f()
... 
>>> try: f()
... except: print('error')
... 
RangeError: Maximum call stack size exceeded  <---------this is a js error
>>>
``` 
The even more wired thing is, after hitting this error, I cannot run `import micropip`:
```py
>>> print('134')
134
>>> import micropip
Error: Traceback (most recent call last):
  File "/lib/python3.8/site-packages/pyodide/_base.py", line 70, in eval_code
    eval(compile(mod, "<exec>", mode="exec"), ns, ns)
  File "<exec>", line 1, in <module>
  File "/lib/python3.8/site-packages/micropip.py", line 18, in <module>
    import hashlib
  File "/lib/python3.8/hashlib.py", line 167, in <module>
    import _hashlib
  File "<frozen importlib._bootstrap>", line 991, in _find_and_load
  File "<frozen importlib._bootstrap>", line 971, in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 914, in _find_spec
  File "<frozen importlib._bootstrap_external>", line 1342, in find_spec
  File "<frozen importlib._bootstrap_external>", line 1309, in _get_spec
RecursionError: maximum recursion depth exceeded in __instancecheck__
```

In a native python3, I can get the correct output:
```py
>>> def f(): return 1 + f()
... 
>>> try: f()
... except: print('error')
... 
error
```

Go back to the pyodide console, we can capture the error in JS, for example, run the following code and I can see the error captured by JS. However, it still makes python stuck into a state that no additional import can be made.
```py
>>> import js
>>> js.eval('try{pyodide.runPython("def f(): return 1 + f()\\nf()")}catch(e){console.error("error!")}')
>>> import micropip
Error: Traceback (most recent call last):
  File "/lib/python3.8/site-packages/pyodide/_base.py", line 70, in eval_code
    eval(compile(mod, "<exec>", mode="exec"), ns, ns)
  File "<exec>", line 1, in <module>
  File "/lib/python3.8/site-packages/micropip.py", line 18, in <module>
    import hashlib
  File "/lib/python3.8/hashlib.py", line 167, in <module>
    import _hashlib
[...]

Issue #821, opened 2020-12-04
Low level stack overflows using Jedi in Chrome


hoodmane:
If I type into the pyodide console here: https://pyodide-cdn2.iodide.io/v0.15.0/full/console.html
`import jedi; import numpy; jedi.Interpreter("numpy.split().", [globals()]).completions()`
the result is "Range Error: Maximum Call Stack Exceeded". The pyodide instance is not completely borked by this but it is pretty screwed up. For instance, there is an error left in `sys.exc_info()`:
```python
>>> import sys; sys.exc_info()
<class 'KeyError'>,KeyError(((), frozenset())),<traceback object at 0x4976ab0>
```
Trying to set a very small recursion limit will tell you that a 334 calls worth of glitched junk is left over on the Python stack:
```python
>>> sys.setrecursionlimit(10)
<... many lines of stack trace omitted>
RecursionError: cannot set the recursion limit to 10 at the recursion depth 334: the limit is too low
```
In a fresh repl this would say `cannot set the recursion limit to 10 at the recursion depth at the recursion depth 9`.

If I start out with `sys.setrecursionlimit(100)` before doing the Jedi call, then it will return a proper Python RecursionError. The initial value of `sys.getrecursionlimit()` is 315.

I did all of this with Chrome v. 86.0.4240.198 under Windows 10.

hoodmane:
Note that the same code works completely fine in Firefox.
Even with a sufficiently low `recursionlimit` so that no low level stack overflows occur, there is still the problem that Jedi seems to only work in Chrome with extremely simple cases because in all other cases there is a stack overflow.

I saw somewhere mention of enlarging the Emscripten [TOTAL_STACK](https://github.com/emscripten-core/emscripten/blob/49029d44cbef0f535fcd48445300808c7efbdb10/src/settings.js#L113) constant, does that have any relationship with this issue? It seems like it might be an issue to raise with the chromium team, but I don't know if there is any chance of convincing them to enlarge the stack.

rth:
Thanks for the report @hoodmane ! Interesting I can't reproduce with Chrome 80.0.3987 and Windows 10, it works fine there with 0.15.0.

Could you please try with the latest development version https://pyodide-cdn2.iodide.io/dev/full/console.html  (which also has a later version of jedi) ?

> enlarging the Emscripten TOTAL_STACK constant, does that have any relationship with this issue?

Yes, it could be worth a try, related to https://github.com/iodide-project/pyodide/issues/716

[...]

Issue #200, opened 2018-10-01
Segfault when listing files in /proc/self/fd 


rth:
Running,
```py
import os
os.listdir('/proc/self/fd')
```
produces the following error `abort({}). Build with -s ASSERTIONS=1 for more info.`, which is effectively a segfault. Running any subsequent command, even `1+1`, will produce the same error message.

After re-compiling with `ASSERTIONS=2`, a more detailed traceback is,
```
E       stackTrace@http://127.0.0.1:33465/pyodide.asm.js:4:22873
E       abort@http://127.0.0.1:33465/pyodide.asm.js:4:11184512
E       ___syscall5@http://127.0.0.1:33465/pyodide.asm.js:4:196773
```
which corresponds to the `open` system call.

It should be noted that, the following works as expected,
```py
>>> os.listdir('/proc/self')
['fd']
```

hoodmane:
This is still an issue:
```
pyodide.asm.js:8 Uncaught abort(TypeError: Cannot read property 'parent' of undefined) at Error
    at jsStackTrace (http://localhost:8001/pyodide.asm.js:8:506153)
    at stackTrace (http://localhost:8001/pyodide.asm.js:8:506360)
    at abort (http://localhost:8001/pyodide.asm.js:8:7435951)
    at ___syscall5 (http://localhost:8001/pyodide.asm.js:8:648188)
    at _open (http://localhost:8001/pyodide.asm.wasm:wasm-function[9730]:0x86a11b)
    at _opendir (http://localhost:8001/pyodide.asm.wasm:wasm-function[9777]:0x86c929)
    at _os_listdir (http://localhost:8001/pyodide.asm.wasm:wasm-function[5721]:0x4078fb)
    at http://localhost:8001/pyodide.asm.wasm:wasm-function[20626]:0x9c3e27
    at _cfunction_vectorcall_FASTCALL_KEYWORDS (http://localhost:8001/pyodide.asm.wasm:wasm-function[1727]:0x18afc9)
    at http://localhost:8001/pyodide.asm.wasm:wasm-function[23488]:0x9cde26
```

hoodmane:
@dalcde Is this an emscripten bug?

dalcde:
I think so. The following program fails to run when compiled with emcc:
```
#include <dirent.h>
#include <sys/types.h>

int main() {
  DIR *dp = opendir("/proc/self/fd");
  return 0;
}

dalcde:
[...]

PR #1151, opened 2021-01-18
Handle fatal runtime errors

### LINKED ISSUE #954: Should raise fatal error if javascript recursion error occurs
If I say:
`import sys; sys.setrecursionlimit(100000); def f(): return 1 + f(); f()`
on native python, the result is `Segmentation fault` and the OS kills Python. This is the correct behavior. Pyodide should also raise a fatal error and refuse to evaluate further code. This should prevent people from getting confused (see discussion in #917, #951).

### LINKED ISSUE #1005: Detecting crashes of the Pyodide WASM environment

hoodmane:
This resolves #954 and resolves #1005.

### DIFF
--- a/src/pyodide.js
+++ b/src/pyodide.js
@@ -319,6 +319,7 @@ globalThis.languagePluginLoader = new Promise((resolve, reject) => {
 
   function makePublicAPI(module, public_api) {
     let namespace = {_module : module};
+    module.public_api = namespace;
     for (let name of public_api) {
       namespace[name] = module[name];
     }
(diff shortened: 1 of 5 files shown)


src/pyodide.js
```diff
@@ -336,6 +337,27 @@ globalThis.languagePluginLoader = new Promise((resolve, reject) => {
...
+      }
+      // Have to do this case first because typeof(some_pyproxy) === "function".
+      if (Module.PyProxy.isPyProxy(value)) {
+        value.destroy();
+        continue;
+      }
+      if (typeof (value) === "function") {
+        Module.public_api[key] = function() {
+          throw Error("Pyodide has suffered a fatal error, refresh the page. " +
+                      "Please report this to the Pyodide maintainers.");
```
phorward, line 354: This will mean: A lot of new tickets ;-)

phorward (approved):
I think this looks good and cleans up the code in some places.

hoodmane:
Thanks for the review @phorward.

> This will mean: A lot of new tickets ;-)

Yeah I'm really curious to learn any new ways to trigger this. I currently know exactly two ways: one is issue #200 which is an emscripten bug:
```python
import os
os.listdir('/proc/self/fd')
```
The other is with recursion:
```python
def f(): f()
f()
```
which happens because `setRecursionLimit` isn't conservative enough when it sets the Python recursion limit (issue #821 / #951).

PR #1152, opened 2021-01-18
Don't automatically copy python objects into javascript

hoodmane:
This implements (the most important change suggested in) #900. This should also make the custom globals / locals feature of `eval_code` implemented in #1106 work significantly better.


packages/matplotlib/test_matplotlib.py
```diff
@@ -2,15 +2,15 @@ def test_matplotlib(selenium_standalone):
     selenium = selenium_standalone
     selenium.load_package("matplotlib")
     selenium.run("from matplotlib import pyplot as plt")
-    selenium.run("plt.figure()")
-    selenium.run("plt.plot([1,2,3])")
+    selenium.run("plt.figure(); pass")
+    selenium.run("x = plt.plot([1,2,3])")
```
phorward, line 6: @hoodmane can you please shortly explain this change?

phorward (approved):
Hello @hoodmane, thanks for the Pull Request. From the code base this looks fine to me. Can you please shortly explain the changes in the test suite?

dalcde:
Since selenium.run uses runPython under the hood, wouldn't the semicolon itself suffice?

hoodmane:
Yeah it would work, I opted for `; pass` because I think that is a little more conspicuous.

packages/matplotlib/test_matplotlib.py
hoodmane: Well I'm not exactly sure why this is needed, but the tests now fail without it. My belief is that it is a problem caused when `selenium` tries to serialize a `PyProxy`, it somehow is raising an exception. I'm not really sure why it worked before because it seems to me unlikely that the return value of `plt.figure()` was a dictionary before?

Anyways the point of this test was to check that `plt.figure()` succeeded and not whether `selenium` knew how to serialize it. Probably it would be better to fix the serialization, but that seems like an independent change.

hoodmane: Saying `x = plt.plot([1,2,3])` also blocks the serialization of the plot. I blocked it by assigning the plot to `x` because most of the other similar tests also say `x = plt.plot([1,2,3])`.

phorward: @hoodmane Ok, thanks for the explanation.

phorward (approved):

phorward:
@hoodmane: It seems that we now have a regression with a test_console.py and your change, can you please take a look at it? I'm don't know how to fix it quickly.
```
___________________________ test_completion[firefox] ___________________________
[gw2] linux -- Python 3.8.2 /usr/local/bin/python

selenium = <conftest.FirefoxWrapper object at 0x7f6529803a00>
safe_selenium_sys_redirections = None

    def test_completion(selenium, safe_selenium_sys_redirections):
        selenium.run(
            """
        from pyodide import console
    
        shell = console.InteractiveConsole()
        """
        )
    
>       assert selenium.run("shell.complete('a')") == [
            [
                "and ",
                "as ",
                "assert ",
                "async ",
                "await ",
                "abs(",
                "all(",
                "any(",
                "ascii(",
            ],
            0,
        ]
E       AssertionError: assert [None, 0] == [['and ', 'as...bs(', ...], 0]
E         At index 0 diff: None != ['and ', 'as ', 'assert ', 'async ', 'await ', 'abs(', 'all(', 'any(', 'ascii(']
E         Full diff:
E           [
E         +  None,
E         -  ['and ',
E         -   'as ',
E         -   'assert ',
E         -   'async ',
E         -   'await ',
E         -   'abs(',
E         -   'all(',
E         -   'any(',
E         -   'ascii('],
E            0,
E           ]

src/tests/test_console.py:201: AssertionError
```
Next time it would be better to first merge master into this branch, and re-test it first.

phorward:
@hoodmane can you please investigate? I think the problem is in the conversion of Tuples, which is the case here.

```js
pyodide.runPython("from pyodide import console; shell = console.InteractiveConsole(); shell.complete('a')")
```
returns `Array [ Proxy, 0 ]`, which seems to be invalid, as it should be `Array [ Array ['and ', 'as ', 'assert ', 'async ', 'await ', 'abs(', 'all(', 'any(', 'ascii('], 0 ]`.

phorward:
@hoodmane I'm sorry I reverted your Pull Request because after a 3 hour examination of the problem I wasn't able to get into a solution. Maybe you should take a look on it. Can you then please re-open the Pull Request again, as I don't have a possibility to do it here.

hoodmane:
This PR currently is a bit bad for usability, it needs some supporting changes to pyproxy to make it a bit easier to use. I was thinking the right approach would be to merge this as-is and then preferably fix the ergonomics problems before the next release, but I could see adding more to this PR.
The tests can be fixed by turning:
```python
        shell.complete('a = 0 ; print.__g')
```
into
```python
        [completions, start] = shell.complete('a = 0 ; print.__g')
        [tuple(completions), start]
```

hoodmane:
Oh also the docs on type conversions and the changelog should be updated.

hoodmane:
Presumably implementing `toJSON` on `pyproxy` would make the tests pass more painlessly.

PR #1153, opened 2021-01-18
Make JsBuffer into a subclass of JsProxy

hoodmane:


phorward (changes requested):
Can you please shortly explain why this change is useful?

hoodmane:
The point is to support duck typing. I think the idea of duck typing is to be able to implement functions like:
```python
def isaduck(obj):
   return hasattr(obj, "quack") and iscallable(obj,quack)
```
If we implement quack on every JsProxy and just throw an error if we can't figure out how to quack:
```python
class JsProxy:
   def quack(self):
      if not hasattr(self.jsobject, "quack"):
         raise TypeError("Underlying Js object is not a duck")
      return self.jsobject.quack()
```
Now `isaduck(jsproxy)` _always_ returns true, but the `jsproxy` may be planning to raise a type error every time you call `jsproxy.quack()`. If we know that `quack()` has no side effects, then you can try:
```python
def isaduck(obj):
   try:
      obj.quack()
      return True
   except TypeError, NotImplementedError:
      return False
```
but this is a dangerous assumption. Thus I view this as a problematic antipattern. The point of this PR is to gradually move away from this antipattern towards conditionally implementing exactly the set of capabilities that the underlying javascript object can support.

phorward (approved):

phorward:
Thank you, that sounds useful!

hoodmane:
I suppose I should open an issue with the explanation I just gave and then further PRs in this direction I can cross reference against the issue.

hoodmane:
Thanks @phorward!

PR #1154, opened 2021-01-19
Add mixed js/python traceback formatting

hoodmane:
This PR adds support for traceback formatting of mixed Python / javascript stack tracebacks. I don't know what the ideal way to signal when we switch from js to python or from python to js is. Currently there is some inconsistency in presentation between the javascript frames and the Python ones.

@rth @dalcde I'm interested in feedback on how the mixed tracebacks should be formatted.

#### Raise `ValueError` into console

```python
raise ValueError("hi")
```
We see two layers of javascript stack trace (starting with `at`) and three layers of python stack trace (starting with `File`):
```python
    at Object.Module.runPythonAsync (http://172.27.140.60:8002/pyodide.js:387:19)
    at Object.Module.runPython (http://172.27.140.60:8002/pyodide.js:340:48)
  File "/lib/python3.8/site-packages/pyodide/_base.py", line 302, in eval_code
    return CodeRunner(
  File "/lib/python3.8/site-packages/pyodide/_base.py", line 225, in run
    exec(mod, self.globals, self.locals)
ValueError: hi
```

#### `__context__` when an error is thrown in a javascript function in a catch block
```javascript
function f(){
    g();
}
function g(){
    pyodide.globals.h();
}
function m(){
    throw new Error("oops");
}
pyodide.runPython(`
    from js import f, m
    def h():
        k()

    def k():
        try:
            raise ValueError("hi")
        except Exception as e:
            m()
`);
try {
    f()
} catch(e) {
    pyodide.formatError(e);
}
```
currently results in the following traceback:
```python
  File "<exec>", line 8, in k
ValueError: hi

During handling of the above exception, another exception occurred:

    at <anonymous>:22:5
    at f (<anonymous>:2:5)
    at g (<anonymous>:5:21)
  File "<exec>", line 4, in h
  File "<exec>", line 10, in k
    at m (<anonymous>:8:11)
Error: oops
```

### DIFF
--- a/src/core/hiwire.c
+++ b/src/core/hiwire.c
@@ -227,7 +227,9 @@ EM_JS_REF(JsRef, hiwire_float64array, (f64 * ptr, int len), {
 EM_JS(void _Py_NO_RETURN, hiwire_throw_error, (JsRef iderr), {
   let jserr = Module.hiwire.get_value(iderr);
   Module.hiwire.decref(iderr);
-  if (typeof(jserr) == = "string") {
+  // clang-format off
+  if (typeof(jserr) === "string") {
+  // clang-format on
     jserr = new Error(jserr);
   }
   throw jserr;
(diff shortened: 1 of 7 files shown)


dalcde:
I don't have time to look into this yet, but note that javascript stack
traces are not standardized and specifically, Chrome and Firefox display
them differently.

hoodmane:
I guess that might make testing this annoying.

PR #1155, opened 2021-01-19
Completion support in InteractiveConsole

casatir:
This is a proposition to add optional Jedi's completion support in `pyodide.console.InteractiveConsole` and so in `console.html`.

### DIFF
--- a/src/pyodide-py/pyodide/console.py
+++ b/src/pyodide-py/pyodide/console.py
@@ -1,4 +1,4 @@
-from typing import Optional, Callable, Any
+from typing import Optional, Callable, Any, List
 import code
 import io
 import sys
@@ -111,6 +111,10 @@ class InteractiveConsole(code.InteractiveConsole):
     persistent_stream_redirection
         Wether or not the std redirection should be kept between calls to
         `runcode`.
+    completion
+        Should completion be used? This preloads the `jedi` module to
+        later be used in `complete`. The underlying promise is set to
+        `self.preloads_complete`.
     """
 
     def __init__(
@@ -119,6 +123,7 @@ class InteractiveConsole(code.InteractiveConsole):
         stdout_callback: Optional[Callable[[str], None]] = None,
         stderr_callback: Optional[Callable[[str], None]] = None,
         persistent_stream_redirection: bool = False,
+        completion=True,
     ):
         super().__init__(locals)
         self._stdout = None
@@ -129,6 +134,8 @@ class InteractiveConsole(code.InteractiveConsole):
         if persistent_stream_redirection:
             self.redirect_stdstreams()
         self.run_complete = _dummy_promise
+        if completion:
+            self.preloads_complete = _load_packages_from_imports("import jedi")
 
     def redirect_stdstreams(self):
         """ Toggle stdout/stderr redirections. """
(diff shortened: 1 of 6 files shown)


casatir:
Unfortunately, Jedi is not available at the time of core tests. Should I put this test in `packages/jedi` or can we have Jedi (and Parso) available for core tests ?

rth:
Thanks! Yes, you can add jeudi to https://github.com/iodide-project/pyodide/blob/e8a8a107ffa15e0d53d2dae87a86e872a849fdfd/.circleci/config.yml#L73 have it available in core tests.

src/pyodide-py/pyodide/console.py
```diff
@@ -119,6 +123,7 @@ def __init__(
         stdout_callback: Optional[Callable[[str], None]] = None,
         stderr_callback: Optional[Callable[[str], None]] = None,
         persistent_stream_redirection: bool = False,
+        completion=True,
```
rth, line 126: Maybe let's make it False by default since jedi is not guaranteed to be installed.

src/pyodide-py/pyodide/console.py
```diff
@@ -241,6 +248,19 @@ def banner(self):
...
+    def complete(self, source: str) -> List[Any]:
+        """Use jedi to complete a source from local namespace.
+
+        Returns
+        -------
+        A list of Jedi's Completion objects, sorted by name.
+        """
+        try:
+            import jedi
+        except ImportError:
```
rth, line 260: I think it would be better not to catch the exception and error explicitly. Otherwise when the returned list of exceptions is empty it's a bit hard to debug

casatir:
Great! Thanks, I'll do that.

src/pyodide-py/pyodide/console.py
casatir: This will not fail actually.

rth: I know but it will return an empty completion list, which is also unexpected..

casatir: OK, I've seen your comment below.

src/pyodide-py/pyodide/console.py
casatir: I agree with this. Just to explain why I designed it like this: if you try to get completion before jedi is available (without awaiting for `preloads_complete`), it will not fail with an error but returns an empty list. I think we can get both actually. I'll try.

rth:
> Just to explain why I designed it like this: if you try to get completion before jedi is available (without awaiting for preloads_complete), it will not fail with an error but returns an empty list. I think we can get both actually

I see what you mean.

Generally looks good. Could you please add InteractiveConsole and other public functions in console to the list in https://github.com/iodide-project/pyodide/blame/master/docs/usage/api-reference.md#L14 Thanks!

rth (approved):

rth:
Do I have to anything specific to make it work? For instance,
```py
>>> import numpy as np
>>> import jedi
>>> 
>>> np.arra + <tab>
```
doesn't seem to work on the dev console. Jedi is correctly loaded and there are no errors. Am I missing something?

At the same time,
```py
>>> su + <tab>
sum     super
```
appears to work fine.

casatir:
Thanks @rth. I think I get it. This is JQuery terminal that drops name if prefix does not match.

You can see that completion from the `InteractiveConsole` is working fine:

```python
>>> import js
>>> import numpy as np
>>> [c.name for c in js.pyconsole.complete("np.arr")]
['array', 'array2string', 'array_equal', 'array_equiv', 'array_repr', 'array_split', 'array_str']
```

To better see such drops:
```python
>>> p <tab> <tab>
pass        pow     print       property
>>> [c.name for c in js.pyconsole.complete("p")]
['pass', 'PendingDeprecationWarning', 'PermissionError', 'pow', 'print', 'ProcessLookupError', 'property', 'pyconsole', 'pyodide']
```

Uppercase 'P' are dropped.

I'll continue to investigate to see if it's possible to avoid such drops.

Please note that this is not necessary to import ̀`jedi` (as you did in your examples, maybe for debugging purpose).

casatir:
OK, I've fixed it! Doing so, I've discovered the great `rlcompleter` core CPython's completer which is faster than Jedi. I'll submit a PR with the fix + use of `rlcompleter` (it actually simplify the fix!).

PR #1156, opened 2021-01-19
MAINT Remove rendered.html

casatir:
This comes from [this discussion](https://github.com/iodide-project/pyodide/commit/e8a8a107ffa15e0d53d2dae87a86e872a849fdfd#commitcomment-46098873) where `renderedhtml` in `console.html` seems useless since #1141 is merged.


dalcde:
We can remove the installation of less as well

rth (approved):
Thanks @casatir ! It doesn't look like it has an impact indeed since we now print frames to str. I also couldn't see any difference after commenting out those CSS rules.

PR #1158, opened 2021-01-21
Add simple web loop

hoodmane:
This is a simple webloop aimed at enabling asyncio support as simply and painlessly as possible. It defers all work to the browser using `setTimeout`. Because the browser event loop has no lifecycle (well no lifecycle visible from javascript), there is no reason for us to have a lifecycle either, and all lifecycle methods are dummied out. I didn't implement `call_exception_handler` or it's friends at all because they are normally invoked during the teardown of an event loop, but we never do teardown.
Also, like in #958, `run_forever` and `run_until_complete` cannot block because we only have one thread.

I think this event loop will have better performance than the one in #958. It may well have worse compatibility, but I stole @oeway's test suite so any difference in compatibility is not caught by the tests. I'm not really sure what the tradeoffs involved are, there might be reason to have a simple event loop with better performance for normal purposes and a more complicated event loop with worse performance but with better compatibility.

Anyways I think this is a good starting point for now and if there is need for something more elaborate we can find out and deal with that down the road. Many thanks to @oeway for his work on this, I used a lot of his ideas here.

After we add an event loop, the only thing left that we need for asyncio support is to implement `then` on pyproxies for awaitable python objects, so that a `pyproxy` of a python awaitable is a javascript awaitable.


rth:
>  I'm not really sure what the tradeoffs involved are, there might be reason to have a simple event loop with better performance for normal purposes and a more complicated event loop with worse performance but with better compatibility.

I also don't have a good understanding of the tradeoffs, but I would tend to go with this simpler solution first, if it's sufficient for most use cases.

oeway (commented):
Looks good, I also like simpler solution!

src/pyodide-py/pyodide/simple_web_loop.py
```diff
@@ -0,0 +1,203 @@
...
+
+    def is_closed(self) -> bool:
+        return False
+
+    def _check_closed(self):
+        """ Used in create_task. Would raise an error if self.is_closed(), but we are skipping all lifecycle stuff. """
+        pass
+
+    def run_forever(self):
+        """We cannot block like a normal event loop would
```
rth, line 49: Also need a 1L summary.
```suggestion
        """Run the event loop forever. Does nothing in this implementation.
        
        We cannot block like a normal event loop would
```

src/pyodide-py/pyodide/simple_web_loop.py
```diff
@@ -0,0 +1,203 @@
...
+        """ Since we cannot block, we just ensure that the future is scheduled. """
+        return asyncio.ensure_future(future)
+
+    #
+    # Scheduling methods: use browser.setTimeout to schedule tasks on the browser event loop.
+    #
+
+    def call_soon(self, callback: Callable, *args, context: contextvars.Context = None):
+        """
+        Schedule the callback callback to be called with args arguments at the next iteration of the event loop.
```
rth, line 65: Need 1L summary + wrap documentation lines at 80 or 88 char.

src/pyodide-py/pyodide/simple_web_loop.py
```diff
@@ -0,0 +1,203 @@
...
+        *args,
+        context: contextvars.Context = None
+    ):
+        """
+        Schedule callback to be called after the given delay number of seconds (can be either an int or a float).
+        """
+        from js import setTimeout
+
+        if delay < 0:
+            raise Exception("Can't schedule in the past")
```
rth, line 93:
```suggestion
            raise ValueError("Can't schedule in the past")
```

src/pyodide-py/pyodide/simple_web_loop.py
```diff
@@ -0,0 +1,203 @@
...
+        Copied from BaseEventLoop.create_future
+        """
+        return futures.Future(loop=self)
+
+    def create_task(self, coro, *, name=None):
+        """Schedule a coroutine object.
+
+        Return a task object.
+
+        Copied from BaseEventLoop.create_task
```
rth, line 139: > BaseEventLoop should not be subclassed by third-party code; the internal interface is not stable.

It's a shame that we have to re-define everything.

src/pyodide-py/pyodide/simple_web_loop.py
```diff
@@ -0,0 +1,203 @@
...
+        """
+        self._check_closed()
+        if self._task_factory is None:
+            task = tasks.Task(coro, loop=self, name=name)
+            if task._source_traceback:
+                del task._source_traceback[-1]
```
rth, line 145: Could you add a inline commment about this?

rth (commented):
Generally looks reasonable (for someone who hasn't worked much with asyncio). A few minor mostly docstring related comments below. Also please add a changelog entry.

rth:
FYI, you can add all review suggestions in one commit/batch if you do it from the "Files changed" tab.

hoodmane:
Thanks for the tip.

src/pyodide-py/pyodide/simple_web_loop.py
hoodmane: Yeah I thought about subclassing `BaseEventLoop` but it has a _lot_ of logic in it and most of it should never be invoked by our event loop, and also they say not to do it. So it's better to just copy these five methods, they're all really short anyways.

hoodmane:
I also copied more docstrings from `BaseEventLoop`, as much as the original docstrings are applicable.

src/pyodide-py/pyodide/simple_web_loop.py
hoodmane: I don't actually understand why this is here either, which is sort of covered by the comment:

> Copied from BaseEventLoop.create_task

I'll go look at CPython git blame and see if I can figure out what it means.

hoodmane: This code was added in this commit:

https://github.com/python/cpython/commit/80f53aa9a0b2af28c9d8052c46b452cccb8e0b41

Specifically:

> The traceback is truncated to hide internal calls in asyncio, show only the traceback from user code.

The conditional `if task._source_traceback:` will be true if and only if the event loop returns `True` from `get_debug()`, which we currently never do.

rth:
Also could you please add the added classes to https://github.com/iodide-project/pyodide/blob/master/docs/usage/api-reference.md and check that that the corresponding documentation is rendering fine?

hoodmane:
> Also could you please add the added classes to https://github.com/iodide-project/pyodide/blob/master/docs/usage/api-reference.md and check that that the corresponding documentation is rendering fine?

I guess it's rendering fine, but it doesn't look great. The autodoc adds a lot of methods from `AbstractEventLoop` that we don't implement. In general, the autodoc config could use a fair amount of work. I don't think there's anything appropriate to do about it in this PR though.

rth (approved):
We still need a changelog entry

hoodmane:
Oops, will add it.

hoodmane:
I also added a separate entry for #880. So after this #1170 is the last step for full asyncio support.

docs/project/changelog.md
```diff
@@ -58,13 +58,24 @@
...
   [#1148](https://github.com/iodide-project/pyodide/pull/1148) and
   [#1173](https://github.com/iodide-project/pyodide/pull/1173)
+<<<<<<< HEAD
+- A `JsProxy` of a Javascript `Promise` or other awaitable object is now a 
+  Python awaitable.
+  [#880](https://github.com/iodide-project/pyodide/pull/880)
+- Added a Python event loop to support asyncio by scheduling coroutines to run 
+  as jobs on the browser event loop. This event loop is available by default and 
+  automatically enabled by any relevant asyncio API, so for instance 
+  `asyncio.ensure_future` works without any configuration.
```
rth, line 68: Should we use create_task here and in tests instead? It is recommended in the docs https://docs.python.org/3/library/asyncio-task.html#asyncio.create_task

Also merge conflicts in this file.

hoodmane: I have a bad habit of opening a merge tool in a separate console window and then forgetting to save before I commit.

hoodmane: On `create_task`, sounds fine.

hoodmane:
Seems like `create_task` behaves differently than `ensure_future`? I thought that when the argument to `ensure_future` was a coroutine that it would just call `create_task`...

hoodmane:
So `asyncio.create_task` uses `get_running_loop()`:
```python
loop = events.get_running_loop() 
task = loop.create_task(coro)
```
whereas `asyncio.ensure_future` uses `get_event_loop()`:
```python
if coroutines.iscoroutine(coro_or_future):
    if loop is None:
        loop = events.get_event_loop()
    task = loop.create_task(coro_or_future)
```
The difference in behaviors occurs only when there is no currently running event loop. In that case, `ensure_future` uses the `EventLoopPolicy` to start an event loop automatically, whereas `create_task` throws an error. I suppose I should just automatically start the event loop during pyodide initialization and that'll make everything simpler.

hoodmane:
Well automatically starting the event loop causes trouble with one CPython test. I guess I'll revert the `create_task` commits.

rth:
OK, thanks for investigating.

mfripp:
I'm not sure if I should be commenting here or opening a new issue. But I think my question is pretty closely related to this, so I'll put it here.

I'm building a web page where students can edit Python code, then click "Run" to run it in pyodide. I started with all of this running in the main browser thread, which worked pretty well. As part of this, I added some glue code so that `input()` in Python became a `prompt()` in the browser. 

However, when pyodide runs in the main thread, my code cannot update the display (e.g., show output), and controls are locked, so I can't have a "Stop" button to kill Python code that runs out of control. So I am working on a version where pyodide runs as a web worker. I've got that working for most of what I want, except for getting input back from the user.

Specifically, I send messages to from the worker to the main browser thread to launch the prompt, then send a message back from the main thread to the worker with the result. The problem is, I can't see any way to get the Python code to wait for the return message. I think that may be possible in general, if I write all-async code. The problem is, the incoming code is not written for an async environment (e.g., the students will use `input()`, not `await input()`). So even if I patch `input()`, my new version has to be a sync function to give a result back to the main code. So then my patched `input()` can't `await` an async function, e.g., a JavaScript promise that clears when the message comes back from the main browser thread.

I tried using `asyncio.get_running_loop().run_until_complete()` to await the promise, but this doesn't do anything, as noted above. I've investigated using emscripten's [Asyncify framework](https://emscripten.org/docs/porting/asyncify.html), and I'm pretty sure I can get that to create a synchronous C function that will await a JavaScript promise. But I'm not sure if I can call the C function from Python in a way that is compatible with the Asyncify system. And even if I can, it will be pretty kludgy.

I'm now thinking about using shared memory between the browser thread and the worker, then running a busy loop in the worker that polls for a response via the shared memory (since I [don't think sleep is possible](https://github.com/iodide-project/pyodide/issues/97)). But I'd rather not use such a narrow solution and peg the processor that way.

So I just wanted to check, is there currently any path to get synchronous Python code to wait for a JavaScript promise or async function?

hoodmane:
@mfripp I moved this over to #1219.

PR #1161, opened 2021-01-21
Minor Makefile improvements

phorward:
This is just some minor code formatting in the Makefile, to make changes easier to track.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -15,9 +15,7 @@ PYODIDE_CXX=$(PYODIDE_ROOT)/ccache/em++
 
 CC=emcc
 CXX=em++
-
 OPTFLAGS=-O2
-
 CFLAGS=\
 	$(OPTFLAGS) \
 	-g \
@@ -27,7 +25,6 @@ CFLAGS=\
 	-Werror=int-conversion \
 	-Werror=incompatible-pointer-types \
 	$(EXTRA_CFLAGS)
-
 LDFLAGS=\
 	$(OPTFLAGS) \
 	-s MODULARIZE=1 \
@@ -59,8 +56,7 @@ all: check \
 	build/packages.json \
 	build/test.html \
 	build/webworker.js \
-	build/webworker_dev.js \
-
+	build/webworker_dev.js
 	echo -e "\nSUCCESS!"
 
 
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -13,13 +15,21 @@ PYODIDE_CXX=$(PYODIDE_ROOT)/ccache/em++
 
 CC=emcc
 CXX=em++
+
 OPTFLAGS=-O2
-CFLAGS=$(OPTFLAGS) -g -I$(PYTHONINCLUDE) -fPIC \
-	-Wno-warn-absolute-paths -Werror=int-conversion -Werror=incompatible-pointer-types \
+
```
rth, line 20: Not sure about the extra newlines here and above. Do they really help readability?

Makefile
```diff
@@ -49,27 +59,28 @@ all: check \
...
+	src/core/pyproxy.o \
+	src/core/python2js_buffer.o \
+	src/core/python2js.o \
+	src/core/runpython.o \
+	src/pystone.py \
+	src/_testcapi.py \
+	src/webbrowser.py \
+	$(wildcard src/pyodide-py/pyodide/*.py) \
+	$(CPYTHONLIB) \
+
```
rth, line 83: I agree we should be more consistent in the makefile. I have mixed feeling on using an extra indentation vs a newline. WDYT @dalcde ?

rth (commented):
Thanks @phorward !

Makefile
phorward: yeah, it helps to better keep track when options are changed or added. I currently do some experiments with the compiler flags, and as the Makefile is maintained by hand and not generated, it makes sense to improve readability and tracking of changes whenever possible.

Makefile
phorward: the double indentation you can see in the current master came in from the GitHub web-based editor when I solved a merge conflict some days ago. I don't know why it did that.

Makefile
rth: For putting one line per option, I agree that it's certainly better for the diff.
I mean the extra empty lines added before and after variables, particularly those that are on 1 line (such as empty lines around `OPTFLAGS=-O2`): as far as I understand that should shouldn't matter for the diff. And if we start doing it everywhere consistently e.g. here https://github.com/iodide-project/pyodide/blob/6a4423356283a5d644c0e144f59570be0d1ba74c/Makefile#L5-L12 I find it a bit strange in terms of formatting.

phorward:
@rth I've removed the empty lines now. You're right, this was a little bit strange. I hope its better now.

rth (approved):
Thanks @phorward !

PR #1166, opened 2021-01-22
Fixed test_console on no-auto-copy branch

hoodmane:


hoodmane:
The ergonomics of this no-auto-copy approach will need to be improved.

PR #1169, opened 2021-01-23
Fix METH_NOARGS function signatures

hoodmane:
According to the python documentation [METH_NOARGS](https://docs.python.org/3/c-api/structures.html#METH_NOARGS) functions should be of type [PyCFunction](https://docs.python.org/3/c-api/structures.html#c.PyCFunction). This means they should have function signature:
```C
PyObject *PyCFunction(PyObject *self, PyObject *args);
```
The second argument `args` will always be `NULL`. I think the gcc calling convention makes it okay to cast a one argument function to a two argument function and then call it as if it had two arguments and I assume that we are using the gcc calling convention, but I think it's in any case preferable not to do that.

### DIFF
--- a/src/core/jsproxy.c
+++ b/src/core/jsproxy.c
@@ -178,7 +178,7 @@ JsProxy_RichCompare(PyObject* a, PyObject* b, int op)
 }
 
 static PyObject*
-JsProxy_GetIter(PyObject* o)
+JsProxy_GetIter(PyObject* o, PyObject* _args)
 {
   JsProxy* self = (JsProxy*)o;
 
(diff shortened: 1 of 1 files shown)


hoodmane:
I wonder why this has no CI.

rth (approved):
Thanks @hoodmane !

rth:
> I wonder why this has no CI.

Ahh, yes, I forgot about the part where CI failed to trigger. It looks like the build failed on master, so I'll have to revert for now, unless you can think of a simple fix.

hoodmane:
Kind of unfortunate, if the build failed you see the red x and it stands out very well. If the CI didn't happen at all for mysterious reasons, the absence of the green checkmark doesn't stand out in the same way.

hoodmane:
Error is a simple fix, I'll submit another PR.

hoodmane:
Okay opened #1176. Sorry about the confusion.

PR #1170, opened 2021-01-23
Make a pyproxy of an awaitable python object an awaitable javascript object

hoodmane:
This PR together with #1158 together finish async/await support. I merged #1158 here even though the change sets are independent because some sort of event loop is necessary in order to test this PR.

This implements `await` on `pyproxy` whenever the `pyproxy` is awaitable. It uses `_PyCoro_GetAwaitableIter` to check whether the object is awaitable just like in [the `GET_AWAITABLE` opcode](https://github.com/python/cpython/blob/v3.8.2/Python/ceval.c#L2012). Normally in implementing the `await` keyword in Python, Python would next use the `YIELD_FROM` command, but this assumes that Python is yielding into an enclosing coroutine. Instead we use `asyncio.ensure_future` to schedule the awaitable if necessary and wrap in a future, then we create a promise and connect the promise handles to the future. This could be implemented rather more concisely in Python rather than C but the current bootstrapping scheme makes using Python code to implement PyProxy a bit annoying.

This mirrors the semantics of the underlying python awaitable, so if it is a coroutine then awaiting it multiple times will raise an error, if it is a future-like object then awaiting multiple times is fine.

The error formatting is a bit limited, these `PythonError` wrappers don't print particularly useful messages in Javascript by default. To improve this requires better javascript / python error integration, something like #1154.


hoodmane:
@rth @dalcde @phorward Would appreciate review on this one. It'd be nice to get this merged to finish the async/io support. After merging this, I think we should add a section to the usage docs explaining how asyncio works because discoverability is currently poor.

rth (commented):
Thanks @hoodmane ! A few minor comments below.

Also
 - all tests currently rely on `eval_code_async`, it might be good to add a test that manually generates an awaitable object with run in Python, and passes it to JS.
 - could you please test that in JS the `then` method is defined only if the Python object is awaitable.

src/core/pyproxy.c
```diff
@@ -223,6 +229,137 @@ _pyproxy_destroy(PyObject* ptrobj)
...
+FutureDoneCallback_dealloc(FutureDoneCallback* self)
+{
+  hiwire_CLEAR(self->resolve_handle);
+  hiwire_CLEAR(self->reject_handle);
+  Py_TYPE(self)->tp_free((PyObject*)self);
+}
+
+int
+FutureDoneCallback_call_resolve(FutureDoneCallback* self, PyObject* result)
+{
```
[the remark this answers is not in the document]
hoodmane: Sounds good. I'm intending to do more documentation and cleanup like this soon.

hoodmane:
@rth I think I addressed all the comments in your review.

src/core/pyproxy.c
```diff
@@ -223,6 +229,166 @@ _pyproxy_destroy(PyObject* ptrobj)
   EM_ASM({ delete Module.PyProxies[$0]; }, ptrobj);
 }
 
+/**
+ * Test if a PyObject is awaitable.
+ * Uses _PyCoro_GetAwaitableIter like in the implementation of the GET_AWAITABLE
+ * opcode (see ceval.c). Unfortunately this is not a public API (see issue
+ * https://bugs.python.org/issue24510) so it could be a source of instability.
```
rth, line 236: FWIW that function hasn't been modified in 6 years since it was added and it's still there in 3.10

src/core/pyproxy.c
```diff
@@ -223,6 +229,166 @@ _pyproxy_destroy(PyObject* ptrobj)
...
+static PyTypeObject FutureDoneCallbackType = {
+    .tp_name = "FutureDoneCallback",
+    .tp_doc = "Callback for internal use to allow awaiting a future from javascript",
+    .tp_basicsize = sizeof(FutureDoneCallback),
+    .tp_itemsize = 0,
+    .tp_flags = Py_TPFLAGS_DEFAULT,
+    .tp_dealloc = (destructor) FutureDoneCallback_dealloc,
+    .tp_call = (ternaryfunc) FutureDoneCallback_call,
+};
+
```
rth, line 354:
```suggestion
// clang-format on

```

rth (approved):
A minor formatting comment otherwise looks good. Thanks!

src/core/pyproxy.c
hoodmane: Thanks for checking! This isn't the only semi-public API we're using (included in the public headers but not documented, and starts with `_`), we also have:
`_PyType_Name` [added in version 3.7](https://bugs.python.org/issue31497)
`_PySet_Update`
`_PyList_Extend` 
`_Py_IDENTIFIER`
`_PyObject_CallMethodId`
`_PyObject_CallMethodIdObjArgs`
`_PyObject_GetAttrId`
`_PyObjectCall_NoArg` (stabilized to `PyObject_CallNoArgs` in v3.9)
`_Py_TPFLAGS_HAVE_VECTORCALL` (stabilized to `Py_TPFLAGS_HAVE_VECTORCALL` in v3.9)
`_PyErr_FormatFromCause`
`_PyGen_Send` ([going to be removed in v3.10](https://github.com/python/cpython/pull/22663))

hoodmane: Probably should copy this info to somewhere more useful.

hoodmane: Okay moved to #1251.

Issue #1168, opened 2021-01-23
Use javascript ndarray package for Buffer conversions


hoodmane:
The `ndarray` npm package seems to be the clear favorite for ndarrays in javascript:
https://www.npmjs.com/package/ndarray
They don't have an analogue for the `suboffsets` field of the Python Buffer Protocol, but as long as `suboffsets` is `null`, it should be possible to do a fairly direct conversion. This should be much better than the current array copying approach, which doesn't really work correctly (see #1134).

rth:
+1 to some standard container for arrays, however I think we should study this question in more details. By now I was hoping that there would be some standard implementations if N dimensional arrays for WASM, maybe written in Rust, that we could use. But after after a quick look it doesn't seem to be the case (e.g. https://github.com/rust-ndarray/ndarray/issues/845).

https://github.com/scijs/ndarray is indeed a lightweight implementation (350 LoC), but as far as I understand it's based on 1D JS arrays and I wonder how the performance is if you try to do basic operations with it. For instance, say I have a 2D array and I want to sum it along an axis. It looks like this would need to be done manually with for loops. So as a container it works, but not so much as an ndarray library (at least coming from python) since the only implemented feature seems to be indexing.

The is also the question of [Appache Arrow](https://arrow.apache.org/) which is getting fairly popular for tabular data. There is a JS implementation in https://www.npmjs.com/package/apache-arrow That would likely make sense for tabular data, and possibly 1D array but I'm not sure they support ND arrays. For more details about this topic see #802

hoodmane:
> By now I was hoping that there would be some standard implementations if N dimensional arrays for WASM, maybe written in Rust, that we could use.

That would be great, and I agree that it seems like this ought to exist.

> it's based on 1D JS arrays 

To be fair, "1D JS arrays" are just Javascript's memory buffers. In some sense all efficient multidimensional arrays are abstractions built on top of memory which is a 1D array.

> the only implemented feature seems to be indexing

It looks to me like there is a significant amount of functionality supported in small side modules. If you search for `ndarray` on npmjs, most packages found there add functionality to this ndarray package. It's common in packages intended for the web to minimize download size by splitting into lots of small pieces so you only pay for what you need. 

https://github.com/scijs/ndarray/wiki/ndarray-module-list

hoodmane:
[...]

PR #1176, opened 2021-01-29
Fix compile error from methnoargs-signatures branch

hoodmane:
We were using `JsProxy_GetIter` as `.tp_iter` (signature is `unaryfunc` i.e., `PyObject* f(PyObject*)`) and to define a `METH_NOARGS`  `__iter__` method (signature should be `PyCFunction` i.e., `PyObject* f(PyObject*, PyObject*)`, second argument always `NULL`. In the standard library, iterables only define `tp_iter` and don't explicitly define an `__iter__` method so I just removed the `__iter__` method.

There is a similar issue with `.tp_as_async->am_await` which is a `unaryfunc` and `__await__` which is a `PyCFunction`. Again, it's nonstandard for builtin C types to define an `__await__` method so I removed it.

### DIFF
--- a/src/core/jsproxy.c
+++ b/src/core/jsproxy.c
@@ -178,7 +178,7 @@ JsProxy_RichCompare(PyObject* a, PyObject* b, int op)
 }
 
 static PyObject*
-JsProxy_GetIter(PyObject* o, PyObject* _args)
+JsProxy_GetIter(PyObject* o)
 {
   JsProxy* self = (JsProxy*)o;
 
@@ -389,10 +389,6 @@ static PyNumberMethods JsProxy_NumberMethods = {
 };
 
 static PyMethodDef JsProxy_Methods[] = {
-  { "__iter__",
-    (PyCFunction)JsProxy_GetIter,
-    METH_NOARGS,
-    "Get an iterator over the object" },
   { "__await__", (PyCFunction)JsProxy_Await, METH_NOARGS, ""},
   { "__dir__",
     (PyCFunction)JsProxy_Dir,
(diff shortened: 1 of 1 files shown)


hoodmane:
This time the CI is running =)

hoodmane:
It's weird that there is no `tp_dir` actually. Indeed, looking at the standard library, extension modules with a `__dir__` function just define a `METH_NOARGS` method called `__dir__` like we do.

hoodmane:
@rth Looks like a sporadic failure.

rth:
Thanks for the explanation and the fix @hoodmane !

rth:
It again failed at build time on master with,
```
src/core/jsproxy.c:419:14: error: incompatible pointer types initializing 'getiterfunc' (aka 'struct _object *(*)(struct _object *)') with an expression of type 'PyObject *(PyObject *, PyObject *)' (aka 'struct _object *(struct _object *, struct _object *)') [-Werror,-Wincompatible-pointer-types]
  .tp_iter = JsProxy_GetIter,
             ^~~~~~~~~~~~~~~
1 error generated.
```
while CI was fine here. Not really sure what's happening. Maybe resetting the CircleCI cache key would help.

hoodmane:
Looks to me like it still has the old version -- could it be a problem
because I used the same branch for this pr as for the old version somehow
it didn't fetch the new commits correctly?

hoodmane:
Looks like a problem merging #1176 with the partially merged version from
#1168. In particular, line 181 of jsproxy.c is wrong in master but it's
right at the tip of my branch. Will submit a third PR when I get back to a
computer and hopefully get it right on the third try.

hoodmane:
Okay now #1177 should hopefully fix this problem actually for real this time...

PR #1178, opened 2021-01-30
Run benchmarks in alphabetical order

dalcde:
This makes it easier to compare.

### DIFF
--- a/benchmark/benchmark.py
+++ b/benchmark/benchmark.py
@@ -72,7 +72,9 @@ def parse_numpy_benchmark(filename):
 
 def get_numpy_benchmarks():
     root = Path(__file__).resolve().parent / "benchmarks"
-    for filename in root.iterdir():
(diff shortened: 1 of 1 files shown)


benchmark/benchmark.py
```diff
@@ -72,7 +72,9 @@ def parse_numpy_benchmark(filename):
 
 def get_numpy_benchmarks():
     root = Path(__file__).resolve().parent / "benchmarks"
-    for filename in root.iterdir():
+    filenames = list(root.iterdir());
+    filenames.sort()
+    for filename in filenames:
```
rth, line 77:
```suggestion
    for filename in sorted(root.iterdir()):
```

rth (approved):
Yes, I also just noticed it when looking into performance results  in https://github.com/iodide-project/pyodide/pull/1102

Issue #973, opened 2020-12-29
jsfunc decorator?


hoodmane:
Here's a funny idea:

```python
import inspect
from js import Function
from functools import wraps
def jsfunc(func):
    argspec = inspect.getargspec(func)
    code = inspect.getdoc(func)
    return Function.new(argspec.args, code)

@jsfunc
def test(a, b, c):
   """
   try {
      throw Error("I am in Javascript!");
   } catch(e){
      console.log(e);
   }
   return new Array(1, 2, 3, ...arguments);
   """
test(7, 0 ,2)
```
Returns `[1, 2, 3, 7, 0, 2, Proxy]`. Not sure how the proxy creeps in there, maybe our implementation of `JsProxy.__call__` is messed up. 

It does not work as a closure of course. As is you can't reference any of the variables from the enclosing python scope. You could do something like wrap it in a javascript closure and pass in a function `$` and then do name lookup like `$("a")` or `$.a`.

dalcde:
I quite like the idea, as long as you get the extra arguments out of
the way...

hoodmane:
One question is how we should do actual docstrings for this. Maybe if it starts with a js comment we can use that?
Annoyingly it doesn't seem possible to get the ast for dynamically defined functions in any reasonable way. I can raise a runtime warning if the disassembly doesn't look like the disassembly for `def f(*args, **kwargs): return None`
```asm
  3           0 LOAD_CONST               1 (None)
              2 RETURN_VALUE
```

hoodmane:
I opened #974 for the argument issue.

oeway:
This is a very cool feature to have, especially when the user can define their own type conversions as discussed in https://github.com/iodide-project/pyodide/issues/964.

hoodmane:
See also #510.

rth:
Do we still want this @hoodmane ? Also https://github.com/pyodide/pyodide/pull/2426 looks to be going in this direction.

ryanking13:
Closing as we decided to use FFI Instead (#1179)

PR #1179, opened 2021-01-30
Jsfunc decorator (WIP interface, interested in feedback)

hoodmane:
This is a WIP toward #973.
This defines a method decorator `jsfunc`. If a function is decorated with `jsfunc`, then its docstring is treated as the source for a javascript function, the actual method body is thrown away. For example:
```python
@jsfunc
def is_promise(obj):
    """
    return Object.prototype.toString.call(obj) === "[object Promise]";
    """
```
In the case where the function uses no closure and no global variables everything works pretty nicely. The global variables will work correctly once #1167 is merged. To get closures to work correctly is a bit more tricky. The setup for the closure is produced at compile time. If the body of the function is just a string, then the compiler will have no way to know that a closure is desired. Currently the following is needed:
```python
from pyodide._base import jsfunc
def f():
    x = 2
    @jsfunc
    def g():
        """return x;"""
        x # have to mention x in body of function
    return g
f()() # ==> 2
```
Another option which is perhaps better would be:
```python
def f():
    # blah blah blah stuff
    @jsfunc
    def g():
         closure_vars(x, y, z) # closure_vars doesn't do anything and doesn't exist, just helps read code
         body("""return x + y + z;""") # body doesn't exist but it prevents dead code elimination
```
It's a bit error prone and annoying having to explicitly list all the variables in the closure, and if `x` is a local variable and also has a global binding either in the python `globals` or `builtins` or on the javascript `globalThis`, then forgetting to `closure_vars` it won't even cause a `ReferenceError`.

In order to avoid the need to do this, we would need to use a procedural macro, which of course do not exist in Python. People have been requesting macros to be added to Python for at least ten years and there is a draft proposal from September PEP 638 to add macros to Python, but it seems likely to be slated for Python 4.0 if it is accepted at all. The discussion about PEP 638 on various programming forms is quite contentious.

We have the option to patch the Python compiler to achieve the same effect of course. I'm not sure if this is desirable (too much complexity probably?), but it sounds fun.

### DIFF
--- a/src/pyodide-py/pyodide/_base.py
+++ b/src/pyodide-py/pyodide/_base.py
@@ -394,17 +394,31 @@ def as_nested_list(obj) -> List:
     except TypeError:
         return obj
 
+
 def jsfunc(func):
     from pyodide_js._module import jsFuncWrapper
     import inspect
-    from functools import wraps
     sig = inspect.signature(func)
-    arg_names = list(sig.parameters.keys())    
+    arg_names = list(sig.parameters.keys())
     code = inspect.getdoc(func)
-    inner = jsFuncWrapper(func.__name__, arg_names, code,  func.__code__.co_freevars, func.__closure__)
+    builtins = func.__globals__["__builtins__"]
+    if type(builtins) != dict:
+        # builtins might be a module sometimes...
+        builtins = builtins.__dict__
+    inner = jsFuncWrapper(
+        func.__name__,
+        arg_names,
+        code,
+        func.__globals__,
+        builtins,
+        func.__code__.co_freevars,
+        func.__closure__,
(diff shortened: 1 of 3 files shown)


phorward:
Hello @hoodmane,
I think this feature can be useful. There should be a hint in the documentation about what's possible and where problems can occur (as you mentioned, with closures). For me, personally, I see some use-cases, where currently `js.eval` is used. What are the others thinking?

rth:
Do we still want this @hoodmane ? I have a bit mixed feelings about this proposal, as compared to just using the JS objects via the FFI.

hoodmane:
I think just using the ffi is better.

PR #1181, opened 2021-01-31
Add packages: PyYAML and its dependency libyaml

daoxian:
Closes https://github.com/iodide-project/pyodide/issues/1171

Following https://github.com/iodide-project/pyodide/pull/1172


packages/pyyaml/patches/Makefile.patch
```diff
@@ -0,0 +1,11 @@
+--- downloads/pyyaml-5.3.1/Makefile	2020-03-18 21:09:19.000000000 +0000
++++ patches/Makefile	2020-11-20 06:33:24.117490628 +0000
+@@ -1,7 +1,7 @@
+ 
+ .PHONY: default build buildext force forceext install installext test testext dist clean
+ 
+-PYTHON=/usr/bin/python
++PYTHON=/usr/local/bin/python3
```
rth, line 8: Any way to use,
```
/bin/env python
```
to autodetect the python used?

rth (commented):
Thanks @daoxian ! Generally LGTM, however I would rather merge https://github.com/iodide-project/pyodide/pull/1102 first which includes major changes in the build. That should happen over the next few days, after which we should be able to include this PR.

packages/pyyaml/patches/Makefile.patch
daoxian: > Thanks @daoxian ! Generally LGTM, however I would rather merge #1102 first which includes some major changes in the build. That should happen over the next few days, after which we should be able to include this PR.

That's great and thank you.

daoxian:
Oops,  this time it seems something wrong when testing...

E       selenium.common.exceptions.TimeoutException: Message: Timed out after 30000 ms

/usr/local/lib/python3.8/site-packages/selenium/webdriver/remote/errorhandler.py:242: TimeoutException

So, how to trigger CircleCI to retry?

dalcde:
I've triggered a rebuild for you

hoodmane:
> Oops, this time it seems something wrong when testing...

These inconsistent test failures come up a lot. When I get them, I like to merge master to trigger the retest if possible, if not `git commit --allow-empty -m Retest ; git push` will trigger a redo of the entire test suite. Or you can save power and ask @dalcde to rerun the particular test that failed.

packages/libyaml/meta.yaml
```diff
@@ -0,0 +1,14 @@
...
+source:
+  url: https://github.com/yaml/libyaml/archive/0.2.1.zip
+
+build:
+  library: true
+  script: |
+    emcmake cmake -DCMAKE_INSTALL_PREFIX=.. .
+    emmake make -j ${PYODIDE_JOBS:-3}
+    ln -s libyaml-0.2.1/libyaml_static.a ../libyaml.a
+    ln -s libyaml-0.2.1/include ..
```
rth, line 14: Why do we need to create these symlinks, couldn't we just include / link to the right directory in pyyaml?

You can also use the `replace-libs` config option to rename a library when linking. See https://github.com/iodide-project/pyodide/blob/0755f191b76f2ef767f4ed3a798846ddd72717e9/packages/matplotlib/meta.yaml#L26-L27

rth (commented):
Could you please please merge master in to make sure it build fine with the latest emscripten?

packages/libyaml/meta.yaml
rth: Actually never mind, the current approach is probably indeed better.

rth:
Thanks @daoxian !

PR #1183, opened 2021-02-01
Fix uninitialized variables in PyProxy

rth:
A fix taken from https://github.com/iodide-project/pyodide/pull/1102


hoodmane:
Oops, thanks for catching my mistakes @joemarshall

PR #1184, opened 2021-02-01
pywasmcross improvements to work with the upstream backend

rth:
Taken from https://github.com/iodide-project/pyodide/pull/1102


dalcde:
I think it would make pywasmcross much easier to read if the list of functions to fix were in a separate file, one way or another.

rth:
> I think it would make pywasmcross much easier to read if the list of functions to fix were in a separate file, one way or another.

Done. 

I think for now we have to keep output as `.wasm` files instead of `.so` as otherwise I get for instance,
```
emcc -O2 -Werror -s EMULATE_FUNCTION_POINTER_CASTS=1 -s SIDE_MODULE=1 -s WASM=1 -s
BINARYEN_TRAP_MODE='clamp' --memory-init-file 0 -s LINKABLE=1 -s EXPORT_ALL=1
build/temp.linux-x86_64-3.8/regex_3/_regex.o build/temp.linux-x86_64-3.8/regex_3/_regex_unicode.o
-o build/lib.linux-x86_64-3.8/regex/_regex.cpython-38-x86_64-linux-gnu.so

shared:ERROR: SIDE_MODULE must only be used when compiling to an executable shared library, and not
when emitting an object file.  That is, you should be emitting a .wasm file (for wasm) or a .js file
(for asm.js). Note that when compiling to a typical native suffix for a shared library (.so, .dylib,
.dll; which many build systems do) then Emscripten emits an object file, which you should then compile
to .wasm or .js with SIDE_MODULE.
```

rth:
In the end I had to revert some of the changes, as linking was failing, and it was not worth the time investigating given that the remaining changes will be done in https://github.com/iodide-project/pyodide/pull/1102 in any case.

dalcde (approved):
I haven't looked into the f2c/clapack changes much, but my opinion is that we ought to merge first and try to simplify it later. It would be tremendously helpful to be able to track upstream emscripten.

rth:
> my opinion is that we ought to merge first and try to simplify it later. It would be tremendously helpful to be able to track upstream emscripten.

I agree. We should merge it asap.

PR #1187, opened 2021-02-03
Rewrite JsProxy_IterNext

hoodmane:
I fixed a couple of minor things in the implementation of `JsProxy_IterNext`:
1. When `done` is true, we are supposed to raise `StopIteration` with the value given by the iterator. This is optional, but it is the correct way to handle generator functions that use both `yield` and `return`.
2. To test the boolean value `done`, it used a function called `hiwire_nonzero` which does the check using `x != 0`. This is weird: `false != 0` is `false`, but `false !== 0` is `true`. It's generally bad to use `!=`. 
3. Unnecessary allocations and conversions.


rth (approved):
Looks good, thanks!

PR #1188, opened 2021-02-03
Test against #780 (and the same for dicts)

### LINKED ISSUE #780: js->py type conversions for arrays/lists not using native types

hoodmane:
#780 was resolved by #1167. This adds tests to show that #780 is resolved.

Closes #780


rth (approved):
Thanks !

PR #1189, opened 2021-02-03
Change isPromise to accept nonstandard javascript awaitables

hoodmane:
Any object that implements a `then` method is considered awaitable by the javascript language. For instance, #1170 makes certain `PyProxy` into nonstandard javascript awaitables. We should implement `__await__` on a `JsProxy` of any object with a `then` method.


rth (approved):
Yes, this is likely better than checking the repr in any case. Thanks!

PR #1193, opened 2021-02-04
F2C, CLAPACK, scipy and numpy fixes for the upstream backend

rth:
Taken from https://github.com/iodide-project/pyodide/pull/1102


rth:
I think this is OK. The patches related to the change of function return type can probably also be done with a conversion script with some regexp and a list of paths. Still as discussed  in https://github.com/iodide-project/pyodide/pull/1184#pullrequestreview-583284387 we might as well do the migration to the upstream backend to unblock a number of points and then try to simplify the different patches.

@dalcde Let me know if you have any other comments. Thanks!

dalcde:
I don't have any comments

packages/scipy/patches/make-int-return-values.patch
```diff
@@ -3,209 +3,2563 @@ is treated as returning void, when it really should return int (values are
...
+-    extern int izmax1_(int *, doublecomplex *, int *);
+-    extern double dzsum1_(int *, doublecomplex *, int *);
+ 
+     safmin = dlamch_("Safe minimum");
+     if ( *kase == 0 ) {
+diff --git a/scipy/sparse/linalg/dsolve/SuperLU/scipychanges.patch b/before/scipy/sparse/linalg/dsolve/SuperLU/scipychanges.patch
+deleted file mode 100644
+index 7a0a3ba..0000000
+--- a/scipy/sparse/linalg/dsolve/SuperLU/scipychanges.patch
++++ /dev/null
```
dalcde, line 2162: @joemarshall why is this patch removed?

rth: I reverted the removal for now. Let's see if it will an issue in https://github.com/iodide-project/pyodide/pull/1102 Thanks for the feedback.

joemarshall: Weird, I can't for the life of me remember why I killed that. I think it was possibly accidental when I was going through superlu and fixing the missing externs. I think that patch may not be ever called, I think it just is there to show the diff between scipy and base superlu...

PR #1195, opened 2021-02-05
Simplify setup code

hoodmane:
Now that #1167 is merged, it's possible to produce the `globals` object from Python. I've been thinking about deleting `runpython.c` for a while now, and I think this is the right time. I made a new function `runPythonSimple` which basically just calls `_PyRun_SimpleString`. Then the setup code in `runpython.c` can be done much simpler in python using `runPythonSimple`.

I think it also makes it easier to understand the C code now that the `pyodide` module isn't imported until `main()` is finished.


dalcde:
Can you resolve conflicts again?

src/pyodide.js
```diff
@@ -382,6 +382,15 @@ globalThis.languagePluginLoader = new Promise((resolve, reject) => {
     throw e;
   };
 
+  Module.runPythonSimple = function(code) {
+    let code_c_string = Module.stringToNewUTF8(code);
+    Module._PyRun_SimpleString(code_c_string);
+    Module._free(code_c_string);
+    if (Module._PyErr_Occurred()) {
+      pythonexc2js();
```
dalcde, line 390: Do you need `Module._` here? In any case, can you add tests for this code path?

dalcde:
From the user's point of view, what is the difference between `raw_globals` and `globals`?

src/pyodide.js
hoodmane: Yeah I need `Module._PyErr_Occurred` here. In `EM_JS` (or `EM_ASM`) they offer the shortened version e.g., `_PyErrOccurred` rather than `Module._PyErrOccurred`, but outside of `EM_JS` I need the full path.

I can add a test, sure, though because this is a private entrypoint, this branch reflects a coding error which should be fixed. It's wrong now though, it needs to be `Module._pythonexc2js();`, currently any error in the Python code will probably cause a `ReferenceError`.

hoodmane:
`raw_globals` is a private implementation detail of the bootstrapping system, so only `globals` should exist from the user POV,

dalcde:
But why can't we just rename raw_globals to globals and let users use that?

hoodmane:
Great question. There are two reasons: builtins lookup, and the possibility of providing a special wrapper around `globals` to support access like `pyodide.globals.some_name` rather than `pyodide.globals.get("some_name")` (see discussion with @phorward on #1175).

If we can just delete the `pyodide.globals.some_name` access in favor of `pyodide.globals.get("some_name")` I think that's worth considering. 

However, I think it is a good thing that e.g., `pyodide.globals.get("dict")` returns `builtins["dict"]` because most users won't make a strong distinction between global variables and builtins (they are slightly different but very similar in behavior). Currently `builtins` access through `pyodide.globals` is supported by saying `globals.update(builtins.__dict__)`. I think this should be removed and we should make `globals` a javascript object that first attempts lookup in `raw_globals` (i.e., the normal globals) and then falls back to lookup on `builtins`. This closely mirrors the normal Python name lookup process.

src/pyodide.js
hoodmane: Okay the error branch works fine as-is. So maybe it'd work to say `if(PyErr_Occurred())`? I'm pretty confused about why it works...

hoodmane: Definitely does not work with `if(PyErr_Occurred())`...

hoodmane: Oh okay `PyRun_SimpleString` has its own error handling, so that branch never happens.

src/core/main.c
```diff
@@ -73,6 +72,8 @@ static struct PyModuleDef core_module_def = {
   .m_size = -1,
 };
 
+int py_exec = Py_file_input;
+
```
dalcde, line 76: Is this still needed?

hoodmane: Yes it is needed, I'll rearrange the code to make it more understandable.

hoodmane: Okay I rearranged things a bit. Let me know if it looks better now.

PR #1197, opened 2021-02-06
Use sphinx-js for pyodide_js documentation

hoodmane:
[sphinx-js](https://github.com/mozilla/sphinx-js) is like autodoc but for javascript. This PR moves the javascript API docs into JsDoc style doc comments in `pyodide.js`. Unfortunately it doesn't seem to work with markdown, but it's possible to use raw sphinx markup in the documentation comments. This yields higher quality docs and keeps the documentation closer to the code.


hoodmane:
@dalcde @rth The docs build is failing on the `npm` command. sphinx-js has a dependence on the npm `jsdoc` package. From the sphinx-js docs:

>Install JSDoc. The tool must be on your $PATH, so you might want to install it globally:
> ```sh
> npm install -g jsdoc
> ```

What should  `.circleci/config.yml` look like?

rth:
Great idea! I was looking for something like it.

> The makefile and various tests use npm / node so it should be available?

I would prefer not to have to build emsdk to compile docs, or tie to to the Makefile. So you could just pick a Docker image that has both Python and Node (or install the missing package). We also need to find out how to install node in ReadThedocs CI https://github.com/iodide-project/pyodide/blob/master/.readthedocs.yml

docs/usage/loading-packages.md
```diff
@@ -3,18 +3,17 @@
...
- - {ref}`pyodide.loadPackage <js_api_pyodide_loadPackage>` for packages built
-   with pyodide, or 
+ - {any}`pyodide.loadPackage` for packages built with pyodide, or 
  - `micropip.install` for pure Python packages with wheels available on PyPi or
    from other URLs.
 
 ```{note}
 `micropip` can also be used to load packages built in pyodide (in
-which case it relies on {ref}`pyodide.loadPackage <js_api_pyodide_loadPackage>`).
+which case it relies on {any}`pyodide.loadPackage`).
```
rth, line 12: Wouldn't we have a name collision if a Python function has the same name?

rth (commented):
Please also update https://github.com/iodide-project/pyodide/blob/master/docs/development/contributing.md#building-the-docs

docs/usage/loading-packages.md
hoodmane: It might be a good idea to call these things `pyodide_js` and `pyodide_py`. Also, we should not name Python functions the same things as the Javascript functions, this will be confusing. Any name with multiple words should be fine since `snake_case` and `snakeCase` don't clash and it's obvious which API is which.

hoodmane:
Okay now the CI works. Conveniently circleCI has a "python 3.8.2 and node" image.

> We also need to find out how to install node in ReadThedocs CI

It claims in the readme that what I did is sufficient:
https://github.com/mozilla/sphinx-js#example

I'm a bit skeptical of this claim, but maybe ReadTheDocs always has JsDoc installed to begin with? I guess that wouldn't be outrageous. How do we test the ReadTheDocs config?

rth:
> How do we test the ReadTheDocs config?

At the moment  mdboom is the only maintainer at https://readthedocs.org/projects/pyodide/ we need to ask for access. But meanwhile merging into master and checking the build log on the above link could be a workaround :)

hoodmane:
Another minor point: this PR moved the javascript API docs directly into `api-reference.md` rather than putting a table there and the actual docs on a separate page. I'm planning to reorganize the API docs a bit soon in a separate PR, I'm not sure what the best organization is but I don't think it's great to have a separate page for each API.

rth:
>  I'm not sure what the best organization is but I don't think it's great to have a separate page for each API.

That's what's often done in the Python ecosystem. A table (with sphinx autosummary) that links to the documentation of different functions / classes. For larger projects I find that having all API on one page make it a bit harder to find stuff particularly if there are many functions/methods, it's also not as well indexed by search engines. For instance if you search for "pandas groupby" you end up [here](https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.DataFrame.groupby.html) if it was part of a a single page you would never end up at the correct location with a search engine, and generally having the documentation of all the other methods there can be confusing.. Of course for 5-10 methods as we have now, single page is likely fine, for 100 methods, probably better not to put them in a single page. I'm not sure where the limit is.

This is an improvement in any case, it doesn't look like there is a way is an easy way to keep the previous behavior short of manually creating those files? The other issue is that it will break existing documentation links from the outside (e.g. in from Github). We can change it if we decide to keep this behavior in the future, but probably better decide now and not change it back and forth.

hoodmane:
> That's what's often done in the Python ecosystem. A table (with sphinx autosummary) that links to the documentation of different functions / classes. For larger projects I find that having all API on one page make it a bit harder to find stuff particularly if there are many functions/methods, it's also not as well indexed by search engines. For instance if you search for "pandas groupby" you end up here if it was part of a a single page you would never end up at the correct location with a search engine, and generally having the documentation of all the other methods there can be confusing.. Of course for 5-10 methods as we have now, single page is likely fine, for 100 methods, probably better not to put them in a single page. I'm not sure where the limit is.

Sounds like a bit of a judgement call: both methods are used to good effect in the Python ecosystem. The official Python docs use one giant file and I'm pretty okay with it:
https://docs.python.org/3/c-api/init_config.html#c.PyConfig_SetString

You are correct though that it basically always requires (1) google (2) ctrl+F. I think our APIs are hopefully going to stay small enough that it shouldn't be a problem. 

Another example that does it all in one page is Jedi:
https://jedi.readthedocs.io/en/latest/docs/api-classes.html#completion

Pandas has a HUGE api and fairly detailed docs for the individuals functions, so it's natural they would go with separate files. Our current setup is bad because you can't see in the sidebar the nearby APIs. If I go to [`pandas.DataFrame.size`](https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.DataFrame.size.html) I can easily see `pandas.DataFrame.shape` in the sidebar. On the other hand, from [`pyodide.register_js_module`](https://pyodide.readthedocs.io/en/latest/usage/python-api/pyodide.register_js_module.html#pyodide.register_js_module) I can't see the matching `unregister_js_module` API. 

So in either case it would be good to have a better sidebar listing.

> it doesn't look like there is a way is an easy way to keep the previous behavior short of manually creating those files?

I think I could do some sphinx magic to get this to work if we decide we want it.

hoodmane:
Though the `pandas` docs have a deficiency in this manner: when the page loads the sidebar should be scrolled so that the current API is in view. They also have "next API" and "previous API" buttons at the bottom of the page which help.

rth:
Assuming our API stays small which is certainly a goal, we could indeed put it on a single page. Maybe we could at least have a separate page for the JS API though not to mix it with the Python API?

rth:
> That's what's often done in the Python ecosystem.

I meant more scientific python packages, which tend to have very large API surface indeed.

hoodmane:
> Maybe we could at least have a separate page for the JS API though not to mix it with the Python API?

I think this is a good idea. You want this rearrangement on this PR?

rth (approved):
A follow up PR might be indeed better. Thanks! 

CI failures unfortunately due to https://github.com/iodide-project/pyodide/issues/1204 but enough tests passed to indicate that the changes in pyodide.js should work.

hoodmane:
I'm also going to try to get some API docs for PyProxy and JsProxy (or at least for PyProxy) going before doing the reorganization since it'd be good to take those into account too.

I'm not really sure how to do JsProxy docs. What's the standard way to use sphinx with C extensions? Presumably this must be well documented?

hoodmane:
This is hard to search for: it's hard enough as it is finding documentation on documentation systems, but sphinx also has it's own extensions, and can be used to document C code.

hoodmane:
@rth Another question: do you know what API docs usually do for documenting operator overloads?

hoodmane:
I guess Numpy does this here:
https://numpy.org/doc/stable/reference/arrays.ndarray.html#arithmetic-matrix-multiplication-and-comparison-operations

PR #1199, opened 2021-02-06
Fix broken links in README.md

rdb:
However, you may want to also set up redirects for these pages in ReadTheDocs.


rth:
Thanks for catching this!

> However, you may want to also set up redirects for these pages in ReadTheDocs.

Yes, it would be good.

Issue #1202, opened 2021-02-06
Performance issues with buffer conversions from python to javascript


hoodmane:
Hello @hoodmane, thank you for the magnificent APIs (deep/shallowCopyToJavaScript) that would eliminate the heavy cost of big array conversion from python to Javascript.  I think the shallowCopy version is designed for memory address reference instead of bulks of content copy. But I find the perfomance not so good as I've imagined, which is displayed as the following:
![image](https://user-images.githubusercontent.com/34263430/107121264-74600900-68cc-11eb-8e92-da3b8b09b5f4.png)

It takes about 1~7 seconds for shallowCopyToJavascript() to complete the memory address reference and maybe some necessary meta data copy I guess. However, it's not  adequate for a realtime computation.  Any suggestions for better conversion performance?

_Originally posted by @daoxian in https://github.com/iodide-project/pyodide/issues/1167#issuecomment-774488338_

hoodmane:
Hi @daoxian. Thanks for reporting this bad behavior. 

The handling of memory buffers is currently not good, this is an issue we hope to fix at some point. I think the current behavior is to make a javascript `Array` of 1080 javascript `Array`s of 3-byte `Uint8Array`. The innermost layer is sliced out of the buffer memory so saying `arr[700][500][0] = 255` will adjust the original data, but populating those large javascript `Array`s is very inefficient at all (remember that an `Array` is a hashmap). 

My guess is it would do much better if you transposed the array to have shape `(3, 1080, 1920)`, though of course this isn't very ergonomic if for instance you are intending to use this data to back an image.

Would you look at #1168? That issue is closely related to the poor performance here. If you have an opinion about that discussion it would be helpful. One thing that limits my ability to fix the Buffer handling code is that I don't really know what the use cases look like, so I really appreciate any input you have. Of course if you want to take a stab at improving the buffer conversion code, let me know and I can tell you what I think you'll need to know.

daoxian:
Indeed, it's much faster after the image transposed to (3, 1080, 1920).

daoxian:
see the actual case from #1168  and codes in #1203

daoxian:
@hoodmane 
The `shallowCopyToJavascript` method seems only available in ES6 JS, not in nodejs.
In Node.js, either of the follows will trigger a runtime error:
>  img = img.shallowCopyToJavascript(depth=1);
>  img = img.shallowCopyToJavascript(1);

` JsException: ReferenceError: depth is not defined`


[...]

Issue #1192, opened 2021-02-04
API for casting PyProxy  / JsProxy to native types


rth:
Following  https://github.com/iodide-project/pyodide/pull/1167#issuecomment-772912630 on improving `PyProxy.deepCopyToJavascript` and  `PyProxy.shallowCopyToJavascript ` name, I think it might be better to continue that discussion in a separate issue rather than a PR.

rth:
>  If you have a preference I can open a PR. Maybe even copyToJS, copy2Js, deepCopy2Js. Is it better to keep the word deep in there? What about shallowCopyToJs and copyToJs? I guess deepCopyToJs is my favorite?

Thinking more about this I'm not sure "copy" is the right term. Copy is 
> make a similar or identical version of; reproduce

while we are not copying, we are more converting or casting the from language to another. There is also the question of,
 - if you have a nested object and you can't convert one of these objects, you would likely still return a PyProxy (or fail I guess, I haven't checked what is done, or it could be an option). So the conversion to JS is not always possible, while the word "copy"  somewhat implies that it would never fail.
 - another question is whether we would eventually need a similar casting for JsProxy, in which case it would be good to think about the API for it (Maybe there is one already I haven't checked). There maybe something like `to_py(deep=False)` would make sense.

In terms of naming maybe we could  also take some inspiration from [PyO3](https://github.com/PyO3/pyo3) which has an API for converting between Python and Rust. For instance, the [into_py](https://docs.rs/pyo3/0.13.1/pyo3/conversion/trait.IntoPy.html) trait so maybe `JsProxy.into_py(deep=False)` would indeed be not so bad. And then I imagine similarly `PyProxy.intoJs` though I'm not sure practical it would be to pass `deep` as an argument?

phorward:
> In terms of naming maybe we could also take some inspiration from [PyO3](https://github.com/PyO3/pyo3) which has an API for converting between Python and Rust. For instance, the [into_py](https://docs.rs/pyo3/0.13.1/pyo3/conversion/trait.IntoPy.html) trait so maybe `JsProxy.into_py(deep=False)` would indeed be not so bad. And then I imagine similarly `PyProxy.intoJs` though I'm not sure practical it would be to pass `deep` as an argument?

Hi Roman, this sounds good, and you're right, the function names are not well decided for now. As a Rust-newbie, I'd also written several into_-functions and this is a good paradigm to borrow from. 

What about 

[...]

PR #1208, opened 2021-02-07
Minor fix: add parameter for shallowCopyToJavascript 

daoxian:
Follow #1202

### DIFF
--- a/src/core/pyproxy.c
+++ b/src/core/pyproxy.c
@@ -306,7 +306,7 @@ EM_JS(int, pyproxy_init, (), {
       }
       return Module.hiwire.pop_value(idresult);
     },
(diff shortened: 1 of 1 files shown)


hoodmane:
Thanks @daoxian. I'm thinking about removing this API entirely since this makes it no different than `deepCopyToJavascript` but with a different default.

phorward:
> I'm thinking about removing this API entirely since this makes it the same as `deepCopyToJavascript` but with a different default.

Will this be part of #1192, with [`Pyproxy.intoJs()` with a deep-parameter](https://github.com/iodide-project/pyodide/issues/1192#issuecomment-773536247)?

hoodmane:
@phorward Exactly, yes.

Issue #1204, opened 2021-02-06
CI issues after the migration to the upstream emscripten backend


rth:
After the migration to the upstream emscripten backend was merged https://github.com/iodide-project/pyodide/pull/1102, there have been quite a few CI failure on master (cf below).  The CI in the PR was green aside for one timeout in chrome packages.

Currently following issues have been observed in the CI on master (even after restarting the CI once),
 - [x] firefox core:  failures in webworker tests,
      ```
      =========================== short test summary info ============================
      FAILED src/tests/test_webworker.py::test_runwebworker_different_package_name[firefox]
      FAILED src/tests/test_webworker.py::test_runwebworker_no_imports[firefox] - s...
      FAILED src/tests/test_webworker.py::test_runwebworker_missing_import[firefox]
      FAILED src/tests/test_webworker.py::test_runwebworker_exception[firefox] - se...
      FAILED src/tests/test_webworker.py::test_runwebworker_exception_after_import[firefox]
      FAILED src/tests/test_webworker.py::test_runwebworker_micropip[firefox] - sel...
      ===== 6 failed, 1 passed, 1 skipped, 1136 deselected in 313.01s (0:05:13) ======
      ```
 - [x] chrome and firefox core: occasional failure in test_keyboard_interrupt https://github.com/iodide-project/pyodide/issues/1200
 - [x] chrome core: occasional timeouts
 - [x] firefox and firefox packages: sometimes,
       ```
       packages/numpy/test_numpy.py::test_runwebworker_numpy[firefox] FAILED    [ 66%]
       packages/pandas/test_pandas.py::test_load_largish_file[firefox] FAILED   [100%]
       ```
       due to timeout other times other times pillow #1201. In chrome sometimes almost all package import fail (after 2hours because of a 30s timeout per test)

At the same time I think this is a CI issue because **locally code from master builds and tests fine** in the same Docker image (aside from test_keyboard_interrupt and test_pillow for me). Maybe we have to re-evaluate running some tests in parallel, and revert xfail scipy related tests as we did before https://github.com/iodide-project/pyodide/pull/1102 There are also other things we could explore.

Because updating to latest emscripten is critical, I think it might be good to try to fix it in the current state. Rather than revert and try to fix in a separate PR. Though this means that master will sometimes show a broken CI which will be an inconvinience for other PRs untill we fix it. If we can't find workarounds soon then we can revert (and re-open a new PR with fixes on top). WDYT?

[...]

PR #1209, opened 2021-02-07
MAINT Update to Firefox 87 in tests

rth:
Following https://github.com/iodide-project/pyodide/issues/1204#issuecomment-774547684, updates to,
 - Firefox 87 (March 2021)
 - geckodriver 0.29.1 (April 2021)
 - ~~chromedriver 87.0 (Nov 2020) to match the chrome version we are using~~ Done in #1462 

~~Also installs selenium 4.0.0.a7 (Nov 2020).~~ Will be done in https://github.com/pyodide/pyodide/pull/1422


src/pyodide-py/pyodide/webloop.py
```diff
@@ -65,7 +65,7 @@ def run_forever(self):
         """
         pass
 
-    def run_until_complete(self, future: Awaitable):
+    def run_until_complete(self, future: Awaitable):  # type: ignore
```
rth, line 68: mypy errors saying that the signature doesn't match the Abstract class, but I don't see any issues there https://github.com/python/cpython/blob/920bf6a3a656e329c2bcbb761eb8c13c46c8cd05/Lib/asyncio/events.py#L214

src/pyodide-py/pyodide/webloop.py
```diff
@@ -223,7 +223,7 @@ def get_task_factory(self):
         return self._task_factory
 
 
-class WebLoopPolicy(asyncio.DefaultEventLoopPolicy):
+class WebLoopPolicy(asyncio.DefaultEventLoopPolicy):  # type: ignore
```
rth, line 226: Similarly mypy errors with a message that DefaultEventLoopPolicy is not a valid type.

rth:
In summary, this something that it would be add to do at some point. And it works locally, however it doesn't address the underlying CI issues from #1204. New failures on Firefox would also require more investigation.

hoodmane:
@rth Now that the CI issues are fixed, maybe you should merge master on this one and see if the tests pass?

rth:
Trying to update Firefox only in this PR. We need this if we are to show benchmarks for the 0.17 release, can't really show those with a Firefox released 1.5 years ago (particularly if Chrome is up to date).

rth:
I think I applied workarounds for all failures. So hopefully Firefox CI would pass now, and we would be finally using up to date browsers for tests.

hoodmane:
CI looks good!

rth:
I additionally marked as xfail `test_runwebworker_numpy` for Firefox which wasn't too stable before and seems to be failing more frequently now. We can always investigate later.

PR #1212, opened 2021-02-07
Remove debug statements from fix_emulate_pointers

hoodmane:
Two `printf` statements in `fix_emulate_pointers_dlsym.patch` print a lot of junk to the console during compilation. I suspect they were added for debugging purposes. This PR removes them.

### DIFF
--- a/emsdk/patches/fix_emscripten_dylink.patch
+++ b/emsdk/patches/fix_emscripten_dylink.patch
@@ -61,9 +61,6 @@ index c8af59d..7084ff7 100644
 +    // only ever called from javascript
 +    var dyncallInvokeMap=new Map();
 +#endif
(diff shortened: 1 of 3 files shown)


rth:
Could you also please remove the accidental patch of `tools/__pycache__/shared.cpython-38.pyc` in `emsdk/patches/fix_emscripten_dylink.patch`? 

Currently the modified patch fails to apply (cf CI).

hoodmane:
Can you explain further what you mean? I have no clue what needs to be done.

hoodmane:
Oh I see I was looking in the wrong patch file. Thanks!

rth:
Hmm, I don't understand why patching fails in CI with,
```
patching file emsdk/binaryen/src/passes/FuncCastEmulation.cpp
patch: **** malformed patch at line 305: diff --git a/upstream/emscripten/src/library.js b/upstream/emscripten/src/library.js
```

hoodmane:
Maybe it's best to apply the patches locally, edit the patched files, and then generate a new patch file?

PR #1213, opened 2021-02-07
Simplify hiwire constants

hoodmane:
We have macros e.g., `Js_UNDEFINED` that represent the constant hiwire keys for interpreter constants. We had functions like `hiwire_undefined` that just return `Js_UNDEFINED`. I removed the functions `hiwire_undefined` and replaced their uses with `Js_UNDEFINED`. In order to get the values into javascript, instead of defining functions that just return constants, I define global variables e.g., `Js_undefined`.

### DIFF
--- a/src/core/hiwire.c
+++ b/src/core/hiwire.c
@@ -6,6 +6,11 @@
 
 #include "hiwire.h"
 
+const JsRef Js_undefined = ((JsRef)(2));
+const JsRef Js_true = ((JsRef)(4));
+const JsRef Js_false = ((JsRef)(6));
+const JsRef Js_null = ((JsRef)(8));
+
 JsRef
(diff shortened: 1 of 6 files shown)


dalcde:
This raises the obvious question - why have the macros at all? I agree it is more idiomatic to use macros, but if we are also defining the global variables, we might as well use the variables everywhere

hoodmane:
Well if you use a global variable the compiled code is different. In particular, the macro tells the compiler that the value will never change whereas with a global variable the emitted code will look up the current value of the global variable. There's no chance this could make a measurable performance difference though.

hoodmane:
I think I get the best of both worlds with `const JsRef Js_undefined = 2;`

hoodmane:
Thanks @dalcde

PR #1214, opened 2021-02-08
Re-enable ccache for binaryen build

dalcde:
Binaryen now invokes cc instead of gcc, and the cc symlink isn't in /usr/lib/ccache. This replaces the implementation in #1014.

This has the disadvantage that build is now broken if ccache is not available.


rth (approved):
Thanks!

Issue #1006, opened 2021-01-01
Use new Javascript GC apis to leak less memory


hoodmane:
@phorward has discussed here #693 the fact that we leak lots of memory. With `FinalizationRegistry` and `WeakRef` we can do better. We must still leak general Python ==> javascript ==> Python reference loops (I believe I can prove that it is theoretically impossible to detect these with the available APIs unless we are allowed to only use javascript objects produced by pyodide), but there are two cases that we can handle:
1. the case when no reference loops cross the javascript / python border, and
2. the case when the js objects involved in the reference loop are only referenced in ways that the Python GC knows about (so the javascript objects are owned by Python).

Case 1 is simple, case 2 is complicated (we'd have to implement a mark and sweep algorithm) and definitely not worth doing except for fun.
One issue that would come up with testing these things is that Javascript gives no way to ask for the garbage collector to run. So there's no way to get a deterministic expected behavior -- how do you test "Javascript will eventually do this"?

mgreminger:
Yes, utilizing `FinalizationRegistry` would be good to automate reference count decrementing of Python created objects when they go out of scope on the JS side. Right now, there is the manual approach to do this using the [`.destroy()`](https://pyodide.readthedocs.io/en/latest/type_conversions.html?highlight=.destroy()#python-from-javascript) method from the JS side, which was necessitated before the availability of `FinalizationRegistry` . Of course, the manual `.destroy()` method doesn't solve the issue raised in #693, but using `FinalizationRegistry` should allow us to fix this. However, to make this work correctly, we'll need to get the reference counting correct for Python objects used on the javsacript side. If the same python is sent to javascript multiple times, its reference count in python needs to be increased each time. This is currently not handled correctly (I initially thought this was due to #708, but the same thing happens in pyodide v15, which was before #708). The following example has an error when the first reference to the Python object is destroyed while the second reference is still in scope. This will need to be fixed for `FinalizationRegistry` to work properly otherwise Python may garbage collect objects that are still in scope on the JS side like in the example below.


```html
<!DOCTYPE html>
<html>
  <head>
      <script type="text/javascript">
          // set the pyodide files URL (packages.json, pyodide.asm.data etc)
          window.languagePluginUrl = 'https://cdn.jsdelivr.net/pyodide/v0.16.1/full/';
      </script>
      <script src="https://cdn.jsdelivr.net/pyodide/v0.16.1/full/pyodide.js"></script>
  </head>
  <body>
    Pyodide test page <br>
[...]

PR #1215, opened 2021-02-08
Add PyProxyBufferMethods

### LINKED ISSUE #1202: Performance issues with buffer conversions from python to javascript

hoodmane:
This resolves #1202. I added an API `getRawBuffer(type)` to all PyProxies of Python objects that support the Python Buffer protocol. The return value of `getRawBuffer` is an object with fields `buffer`, `shape`, `strides`, `offset`, and `release`. The `shape` and `strides` are as in the PyBuffer protocol but `strides` has been divided by an appropriate alignment factor. The equivalent of `x[idx0, idx1, ..., idxn]` in python is `buffer[ offset + strides[0] * idx0 + ... + strides[n] * idxn]`.
The arguments `buffer`, `shape`, `strides`, `offset` are suitable to be used as the arguments to the constructor for ndarray:
https://scijs.net/packages/#scijs/ndarray
e.g.,
```javascript
let {buffer, shape, strides, offset, release} = proxy.getRawBuffer("i32");
let mat = new ndarray(buffer, shape, strides, offset);
```
`release` is a function to call to free the buffer memory (you'll have to destroy the proxy you got it from too though).

I think I might add a `readonly` flag to indicate whether the buffer had the `readonly` flag set. This will hopefully fail on any buffer with `suboffsets` (unless they decide to copy themselves in `bf_getbuffer`...)

@daoxian please let me know what you think.


daoxian:
Great work! The ability to construct common JS ndarray of python's raw Buffer without copying data is enough for my cases.  As for the `release` method,  since python will auto-release the memory of a local variable if the execution goes out of the variable's definition scope, maybe the method is not necessary?

daoxian:
And also I suffer from some memory leaks, such as :
python calls 
> js.g_tracker.resize_img(img_bytes, ...)

while in JS resize_img, I think img_bytes is a Ref to the python bytes memory,   should I release the memory in JS by set img_bytes=null ?

hoodmane:
@daoxian You have to `destroy` the PyProxy and `release` the memory.

> So should I release the memory in JS by setting `img_bytes=null`? 

Currently you need to say `img_bytes.destroy()`, unfortunately setting `img_bytes = null` leaks the memory. See issues #1006 and #1049. I have a fairly good idea of what the garbage collection code should like, but in cases where you have allocated large amounts of memory it'll be best to be safe and explicitly destroy the proxy even after I implement automated garbage collection (and browsers like Safari can't support it).

Note that pyodide is currently rather full of memory leaks, you have to take special care to avoid them.

> If so , what's the necessarity of release ?

`getRawBuffer` makes a call to [`PyObject_GetBuffer`](https://docs.python.org/3/c-api/buffer.html#c.PyObject_GetBuffer). This has to be paired with a call to [`PyBuffer_Release`](https://docs.python.org/3/c-api/buffer.html#c.PyBuffer_Release) when you are done with the memory.

Note that no matter what `getRawBuffer` must have some sort of release method. Well, I suppose one could tie the lifetime of the `rawBuffer` to the lifetime of the `PyProxy`. But currently a `PyProxy` can be unexpectedly destroyed (#1049) and then you could trigger use after free errors. I think it's good to make the caller of `getRawBuffer` responsible for releasing the buffer.

hoodmane:
Here's an example:
```javascript
function resize_img(src_data, dst_data) {
   let src_buffer = src_data.getRawBuffer();
   let dst_buffer = dst_data.getRawBuffer();
   src_data.destroy();
   dst_data.destroy();
   const src_img=cv.matFromArray(src_buffer.shape[0], src_buffer.shape[1], cv.CV_8UC3, src_buffer.buffer); 
   const dst_img=cv.matFromArray(dst_buffer.shape[0], dst_buffer.shape[1], cv.CV_8UC3, dst_buffer.buffer);
   // not really sure what this call should look like
   cv.resize(src_img, dst_img);
   src_buffer.release();
   dst_buffer.release();
}
```

daoxian:
I see. Are both APIs (destroy & release) suitable for ver 0.16.x ?   
Btw, I think a automatic garbage-collection mechanism is necessary for pyodide.   Manually releasing memory is not friendly to beginners.

hoodmane:
> I see. Are both APIs (destroy & release) suitable for ver 0.16.x ?

destroy is present in ver 0.16.1 (and much older versions as well). Release is not.

> I think a automatic garbage-collection mechanism is necessary for pyodide. Manually releasing memory is not friendly to beginners.

I agree that it is a high priority to improve the GC situation. Unfortunately, browsers have a really poor APIs for destructors. Until July 2020 it was technically impossible to do anything better. With `FinalizationRegistry` it is possible to do automatic reference counted garbage collection, but it is still technically impossible to automatically collect reference cycles between Python and Javascript.

The debugging infrastructure is also very limited: there is no way to request for the browser to perform a collection, which is generally important for debugging destructors.
https://caniuse.com/?search=finalizationregistry

Using `FinalizationRegistry` to implement partial automatic GC is high on my priority list. See #1006 though.

daoxian:
> > I see. Are both APIs (destroy & release) suitable for ver 0.16.x ?
> 
> destroy is present in ver 0.16.1 (and much older versions as well). Release is not.
> 
> > I think a automatic garbage-collection mechanism is necessary for pyodide. Manually releasing memory is not friendly to beginners.
> 
> I agree that it is a high priority to improve the GC situation. Unfortunately, browsers have a really poor APIs for destructors. Until July 2020 it was technically impossible to do anything better. With `FinalizationRegistry` it is possible to do automatic reference counted garbage collection, but it is still technically impossible to automatically collect reference cycles between Python and Javascript.
> 
> The debugging infrastructure is also very limited: there is no way to request for the browser to perform a collection, which is generally important for debugging destructors.
> https://caniuse.com/?search=finalizationregistry
> 
> Using `FinalizationRegistry` to implement partial automatic GC is high on my priority list. See #1006 though.

Sorry for the modification. What about this question: "bytes in Python will be converted to a Uin8ClampedArray array. Then the array is used in JS, so should I release the array after using it?"

hoodmane:
> bytes in Python will be converted to a Uin8ClampedArray array. Then the array is used in JS, so should I release the array after using it?

No that uses the previous implicit conversion method. It has lifetime tied to the Python object that it came from. In particular, I think uses this `Uint8ClampedArray` opens you up to use-after-free errors, make sure you have a python reference to the `bytes` for the entire period of time for which you use that array.

daoxian:
From issue https://github.com/iodide-project/pyodide/issues/1221#issuecomment-782889804

I found a bug in your proxy-buffer-procs branch:
src/core/pyproxy.c, line 667:

> if(window.BigInt64Array)

will break the web worker's rule (cannot access window in web worker) and throws exception when in web workers. You could fix it before PR merge.

> if(globalThis.BigInt64Array)

is just fine to run.

src/core/pyproxy.c
```diff
@@ -618,6 +618,105 @@ _pyproxy_ensure_future(PyObject* pyobject,
...
+  }
+
+  bool success = false;
+  if (view.ndim == 0) {
+    FAIL();
+  }
+  buffer_struct result = { 0 };
+  result.shape = array_to_js(view.shape, view.ndim);
+
+  result.start_ptr = result.smallest_ptr = result.largest_ptr = view.buf;
```
rth, line 678: This could use some comments particularly about what what `smallest_ptr`, `largest_ptr` etc do

src/core/pyproxy.c
```diff
@@ -1172,8 +1310,47 @@ EM_JS_NUM(int, pyproxy_init_js, (), {
...
+      if(start % alignment !== 0 || smallest % alignment !== 0 || largest % alignment !== 0){
+        _PyBuffer_Release(view_ptr);
+        _PyMem_Free(view_ptr);
+        throw new Error(`Buffer does not have valid alignment for type ${type}`);
+      }
+
+      let length = (largest - smallest) / alignment;
+      let offset = (start - smallest) / alignment;
+
+      let buffer = new ArrayType(HEAP8.buffer, smallest, length);
```
rth, line 1346: I find these variable names a bit confusing. For instance in the Float32Array signature, `byteOffset` is significantly more explicit that `smallest`. And then in an ndarray what `length` could correspond to (size of the buffer, etc) is also unclear.

rth (commented):
This is nice very work! Did you base this on the [Python buffer protocol reference](https://docs.python.org/3/c-api/buffer.html). Another useful reference would be [Numpy's Array Interface](https://numpy.org/doc/stable/reference/arrays.interface.html) document. 

For the `PyBuffer` fields:
 - I'm wondering if we should rename "buffer" field to "data" to match the numpy definition.
 - should we add a "type" field (represented as a string)?
 - In numpy an important information would be C or Fortran contiguity. In Python buffer protocol that would be specified by PyBUF_C_CONTIGUOUS, and PyBUF_F_CONTIGUOUS. I'm not sure where that information should be contained. I doesn't need to be implemented in this PR, but we should take it into account.

For the method name `getBuffer` should be OK I think. We don't have non raw buffers anyway.

For Python objects that implement [the buffer protocol](https://docs.python.org/3/c-api/buffer.html#c.Py_buffer.format) we know the type/format of single element in the buffer, so we shouldn't have to pass the type as a parameter `getRawBuffer` as far as I understand.

Also please add a changelog entry and a bit more detailed docstrings.

I think it would also be interesting to write more tests for different flavors of nummy arrays (C, F ordered, non contiguous, read-only), just to see what happens even if those are not supported in this PR. I can help with this part if necessary.

hoodmane:
Thanks for the review! I'd especially appreciate any input you have since I don't have that much experience using this sort of low level buffer API.

> Did you base this on the Python buffer protocol reference? 

Exactly, I read that very very carefully.

> Another useful reference would be Numpy's Array Interface document.

Thanks, I'll take a look. Of course, it's easier for us to wrap `PyBufferProtocol` since the Python buffer objects already support that API.

> we should rename "buffer" field to "data" 

Sounds good. "buffer" is bad because it is reusing the word buffer since the object is already an instance of the `PyBuffer` class. FWIW, the `Py_buffer` API does call that field `buf`.

> should we add a "type" field (represented as a string)?

Or maybe "format" since that's what the Py_buffer API calls it.

> Contiguity

This is measured by the C API `PyBuffer_IsContiguous`. The Python `memoryview` API has `c_contiguous`, `f_contiguous`, and `contiguous` (which means either) as fields, it'd be easy enough to mimic that.

>  write more tests for different flavors of numpy arrays

Yeah it'd be great if you could help with that. I'm not that familiar with the relevant APIs.

src/core/pyproxy.c
hoodmane: I renamed these `minByteOffset`, `maxByteOffset`, `numBytes` and `numEntries`. Is it clear enough now?

hoodmane:
> we know the type/format of single element in the buffer, so we shouldn't have to pass the type as a parameter

Well in principle that should be the case, but it seems like there are a couple of confusing points: What do we do with wrong-endianness? So for instance with the format `>H`? With `H`, we would return a `Uint16Array`, but since `>H` is big endian, we probably return a `Uint8Array`? Or maybe a `DataView` since that has getters and setters that can handle big endian data?

What about if it's a struct `HH`? Then presumably we return a `Uint16Array`. But what about `BBH`? Should the result be a `DataView`? Or maybe a `Uint8Array`?

Maybe take an optional type argument and in any ambiguous case throw an error if the type is missing?

src/core/pyproxy.c
```diff
@@ -676,13 +686,17 @@ _pyproxy_get_buffer(PyObject* ptrobj)
...
-    PyErr_Clear();
-    return NULL;
+    // Buffer cannot be represented without suboffsets. The bf_getbuffer method
+    // should have set a PyExc_BufferError saying something to this effect.
+    pythonexc2js();
   }
 
   bool success = false;
   if (view.ndim == 0) {
+    // TODO: what should we do here?
```
rth, line 699: ndim=0 works with numpy,
```py
>>> import numpy as np
>>> x = np.array(0)
>>> x
array(0)
>>> x.ndim
0
>>> x.shape
()
```
and [is defined in the buffer protocol](https://docs.python.org/3/c-api/buffer.html#c.Py_buffer.ndim) so could we return that ?

rth:
> So for instance with the format >H? With H, we would return a Uint16Array, but since >H is big endian, we probably return a Uint8Array?

Is big endian relevant for us though? WASM, amd64 and modern ARM processors are all little endian. It might be fine to just error on big endian format, and ask the user to re-allocate as little endian in Python if they ever get there.

> What about if it's a struct HH? Then presumably we return a Uint16Array. But what about BBH? Should the result be a DataView? Or maybe a Uint8Array?

Well maybe we could have a default for types that map directly to JS arrays, and error for everything else, saying that types should be manually provided?

hoodmane:
> Is big endian relevant for us though?

Not super, networking protocols are big endian but hopefully we are protected from that.

> Well maybe we could have a default for types that map directly to JS arrays, and error for everything else, saying that types should be manually provided?

Yeah, I'm thinking if endianness is normal and every entry has the same size then we should automatically return an array with entries of that size if we don't receive the type argument, and in other cases require a type argument.

src/core/pyproxy.c
hoodmane: So you think we should return it as a single entry TypedArray?

What about the case when the shape has a 0 in it? You have an opinion about that one?

rth: If prod(shapes) == 0, it means there is no data, so I think one would return an buffer with an empty array.

hoodmane: Okay I think I fixed it to correctly handle these cases.

hoodmane:
Okay I think I implemented all the desired features and the docs are decent, but we need more test coverage.

rth:
Does the numpy test passes for you?

I'm getting,
```
E           conftest.JavascriptException: ReferenceError: alignment is not defined
E               at Proxy.getBuffer (http://127.0.0.1:39253/pyodide.asm.js:9:205636)
```
with the latest version of this PR

hoodmane:
Oops, one minute let me fix that.

hoodmane:
I think it should be better now?

hoodmane:
Nope still not working...

hoodmane:
Okay, well `getBuffer` succeeds without throwing any errors now but it still seems a bit buggy.

rth:
Thanks! At least the numpy test passes. I'll try to add a few more.

hoodmane:
> it still seems a bit buggy

Nevermind, that specific problem was just me misunderstanding the numpy API.

packages/numpy/test_numpy.py
```diff
@@ -191,3 +194,115 @@ def test_runwebworker_numpy(selenium_standalone):
...
+    )
+
+
+@pytest.mark.parametrize(
+    "arg",
+    [
+        "np.arange(6).reshape((2, -1))",
+        pytest.param(
+            "np.arange(12).reshape((3, -1))[::2, ::2]", marks=pytest.mark.xfail
+        ),
```
rth, line 234: The buffer size (and contents) are different in this case. It's likely expected given this implementation where start and endoffsets are recomputed but I wonder if it is an issue.

packages/numpy/test_numpy.py
```diff
@@ -191,3 +194,115 @@ def test_runwebworker_numpy(selenium_standalone):
...
+
+            assert_equal(x_js_buf.ndim, x.ndim)
+            # The following check fails
+            # assert_equal(x_js_buf.shape.to_py(), list(x.shape))
+            assert_equal(x_js_buf.strides.to_py(), list(x.data.strides))
+            assert_equal(x_js_buf.format, x.data.format)
+            assert_equal(len(x_js_buf.data), np.prod(x.shape))
+            assert_equal(
+                x_js_buf.data.tobytes(order='A').hex(),
+                x.data.tobytes(order='A').hex()
```
rth, line 272: Converting to hexadecimal so it's easier to compare the diff

packages/numpy/test_numpy.py
```diff
@@ -191,3 +194,115 @@ def test_runwebworker_numpy(selenium_standalone):
...
+        window.x_js_buf = pyodide.pyimport("x").getBuffer();
+        """
+    )
+
+    selenium.run_js(
+        """
+        pyodide.runPython(`
+            from unittest import TestCase
+            from js import x_js_buf
+            assert_equal = TestCase().assertEqual
```
rth, line 262: I'm not sure if we have converged to a better way for asserting things from within runPython with a reasonable error messages?

packages/numpy/test_numpy.py
```diff
@@ -191,3 +194,115 @@ def test_runwebworker_numpy(selenium_standalone):
...
+        """
+        await pyodide.runPythonAsync(`
+            x = np.ones(2, dtype=np.float32).byteswap()
+        `);
+        window.x_js_buf = pyodide.pyimport("x").getBuffer();
+        pyodide.runPython(`
+            from unittest import TestCase
+            from js import x_js_buf
+            assert_equal = TestCase().assertEqual
+            assert_equal(x_js_buf.data[0], 1.0)
```
rth, line 305: This runs but produces incorrect result. Maybe we should just error on big endian for now?

rth (commented):
I added a couple more numpy tests

packages/numpy/test_numpy.py
```diff
@@ -191,3 +194,115 @@ def test_runwebworker_numpy(selenium_standalone):
...
+        """
+        pyodide.runPython(`
+            from unittest import TestCase
+            from js import x_js_buf
+            assert_equal = TestCase().assertEqual
+
+            assert_equal(x_js_buf.ndim, x.ndim)
+            assert_equal(x_js_buf.shape.to_py(), list(x.shape))
+            # The following check fails
+            # assert_equal(x_js_buf.strides.to_py(), list(x.data.strides))
```
rth, line 267: This check also fails, and I'm not sure why. 

You can run xfailed test (excluding this one) with the `--runxfail` pytest flag.

packages/numpy/test_numpy.py
hoodmane: Yeah this is the expected behavior: if the buffer isn't contiguous the `data` field will contain extra entries that are not part of the array.
What does the `-1` in the shape mean?

packages/numpy/test_numpy.py
hoodmane: It doesn't seem like `byteswap` changes the format of the array to mark it as having changed endianness. If the test is changed to `assert_equal(x_js_buf.data[0], x[0])` it will pass.

hoodmane: Okay I think that these wrong endian buffers are basically impossible to produce and nothing seems to support them. In fact, I think that all of this convoluted code I wrote was unnecessary and the formats can only be single letters. Would be nice if the docs were a little bit easier to understand.

packages/numpy/test_numpy.py
hoodmane: It seems to pass alright for me.

packages/numpy/test_numpy.py
hoodmane: Okay, so I removed this test and any mention of endianness from the code.

packages/numpy/test_numpy.py
hoodmane: I looked into it and I managed to parse code into ast, do pytest assertion rewrites on the ast, pickle the ast, and then send the pickled ast to pyodide unpickle and compile the ast and run it. It generates nice looking pytest error messages from pyodide. Likely more effort than it's worth to do this in our test suite though.

packages/numpy/test_numpy.py
rth: -1 in reshape means the remaining shape. So for instance `np.arange(6).reshape((2, -1))` is identical to `np.arange(6).reshape((2, 3))`. It will error if the remaining size isn't an int.

rth: OK, maybe we could keep this line with a comment but remove the xfail

packages/numpy/test_numpy.py
rth: They should be possible to create. Maybe we are doing something wrong. Anyway, as a initial version this is likely fine, let's see if there are user reports about it.

docs/usage/type-conversions.md
```diff
@@ -368,15 +368,30 @@ numpy_array = np.asarray(array)
 
 ### Converting Python Buffer objects to Javascript
 
-Python `bytes` and `buffer` objects are translated to Javascript as
-`TypedArray`s without any memory copy at all. This conversion is thus very
```
rth, line 372: For bytes we still support converting it to TypedArray with toJs, right? Might be worth mentioning it.

src/core/pyproxy.js
```diff
@@ -662,7 +662,287 @@ TEMP_EMJS_HELPER(() => {0, /* Magic, see comment */
...
+        }
+      }
+    }
+  };
+
+  /**
+   * A class to allow access to a Python data buffers from Javascript. These are
+   * produced by :any:`PyProxy.getBuffer` and cannot be constructed directly.
+   * When you are done, release it with the :any:`release <PyBuffer.release>`
+   * method.
```
rth, line 825: See 
```suggestion
   * method.  See 
   `Python buffer protocol documentation <https://docs.python.org/3/c-api/buffer.html>`_
   for more information.
 ```

File: src/core/pyproxy.js [hunk 11]
```diff
@@ -662,7 +662,287 @@ TEMP_EMJS_HELPER(() => {0, /* Magic, see comment */
...
+        /**
+         * The shape of the buffer, that is how long it is in each dimension.
+         * The length will be equal to ``ndim``. For instance, a 2x3x4 array
+         * would have shape ``[2, 3, 4]``.
+         * @type {number[]}
+         */
+        this.shape;
+
+        /**
+         * @type {number[]}
```
rth, line 907:
```suggestion
         * An array of of length ndim giving the number of elements to skip to get
         * to a new element in each dimension.
         * @type {number[]}
```

rth (approved):

packages/numpy/test_numpy.py
hoodmane: I will try to modify the test to assert the right thing for noncontiguous buffers.

docs/usage/type-conversions.md
hoodmane: Well I think we should discuss whether we should get rid of that, I think it's bad. The `TypedArray` resulting allows the user to modify the `bytes` object even though it's read only and might be shared with a bunch of other places. It also leads to potential use after free errors, and can violate expectations because in every other case `toJs` either copies stuff or returns a `PyProxy` (the other biggest problem with `toJs` is that it creates an unpredictable number of `PyProxy` objects that must be located and freed in order to avoid a leak).

Perhaps we should convert `toJs` to exclusively copy the buffer into a Javascript array, so that `toJs` is slow but safe and returns easy to use buffers whereas `getBuffer` is fast and unsafe but only does the one thing so anyone who reads the docs will see the warnings about modifying readonly buffers.

rth: Could you please copy the above comment to a new issue? Unless there is one already about it. I'll respond there..

hoodmane: #1221 is on the same topic, but I think it's better to start a fresh issue.

hoodmane: Opened #1374.

Makefile
```diff
@@ -114,7 +114,7 @@ clean-all: clean
 	make -C cpython clean
 	rm -fr cpython/build
 
-%.o: %.c $(CPYTHONLIB) $(wildcard src/**/*.h)
+%.o: %.c $(CPYTHONLIB) $(wildcard src/**/*.h src/**/*.js)
```
hoodmane, line 117: @rth How do I make it so that `pyproxy.o` depends on `pyproxy.js` but not make it so the other `*.o` files depend on `pyproxy.js`?

hoodmane: After this I think this PR is ready to merge.

packages/numpy/test_numpy.py
hoodmane: Yeah if you can find code that generates a numpy array with `array.data.format == '>I'` or even with `len(array.data.format) > 1` please lmk.

Makefile
rth: I'm not aware of a way to do that. Short of writing the compilation command for each *.o separately. Is it a significant issue though? Compilation of those is fast particularly with ccache..

hoodmane: No it's not important, just a compulsive thing. In that case I guess I'll merge this.

PR #1222, opened 2021-02-09
TST xfail the pillow test that fails on Chrome 

rth:
This ~~skips part of~~ xfails the pillow test that leads to timeouts on Chrome: https://github.com/iodide-project/pyodide/issues/1201
It takes some work to detect browser from within pyodide so for now I'm skipping it for both browsers.

Also reverts https://github.com/iodide-project/pyodide/pull/1205 since it should no longer be necessary given that scipy size is now much smaller.

Closes https://github.com/iodide-project/pyodide/issues/1204


rth:
So the total CI pipeline takes 34min here! ccache (and avoiding time outs in tests) really helps.

PR #1223, opened 2021-02-10
Update to latest (unreleased) emscripten version

dalcde:
This PR will perpetually be a draft, and when new emscripten releases are made,
a separate PR will do the update and incorporate necessary changes.

The goal is to always keep on top of upstream changes, so that it is easier to
pinpoint breakages. In particular, if a genuine emscripten regression is found,
we can report it and hopefully get it fixed/reverted before it lands in a
release.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -8,6 +8,8 @@ defaults: &defaults
     - EMSDK_NUM_CORES: 4
       EMCC_CORES: 4
       PYODIDE_BASE_URL: https://cdn.jsdelivr.net/pyodide/dev/full/
+      PYODIDE_EMSCRIPTEN_VERSION: tot
+      PYODIDE_BINARYEN_VERSION: HEAD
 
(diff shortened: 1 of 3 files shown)


dalcde:
I believe the latest breakage is due to https://github.com/llvm/llvm-project/commit/3a1513c142f4a1ddb97d882a12c89f90cb2529ac :
> This patch adds the flag FlangOnlyOption to the existing options `-ffixed-form`, `-ffree-form` and `-ffree-line-length-` in Options.td.

Previously clang just blindly accepts `-ffixed-form` regardless of whether it applies. Now it complains if it is not used in a fortran context

rth:
Good idea! We could potentially also run it as a [cron job](https://circleci.com/docs/2.0/workflows/#nightly-example)

rth:
emscripten 2.0.14 is out, I'll let you do the PR since you know which patches to remove :)

PR #1224, opened 2021-02-10
Replace some scipy patches with sed

dalcde:


rth (approved):
Thanks ! For the other replacements e.g. `void f_dgemm(` to `int f_dgemm(` I'm not sure if it would be better to do them in a python script with regex or keep them as they are currently.

dalcde:
Yeah that would be more complex. I thought to get the low-hanging fruit first.

Another thing we should look into is why this worked previously; we were compiling to wasm previously too so it cannot be a fundamental restriction of wasm

rth:
> Another thing we should look into is why this worked previously

You mean with previous function signatures? Well it wasn't fully working. A number of scipy functions produced errors at runtime previously e.g. https://github.com/iodide-project/pyodide/issues/492 https://github.com/iodide-project/pyodide/issues/615

dalcde:
Fair enough

hoodmane:
Really cool how many of these old issues we've fixed recently!

phorward (approved):

PR #1225, opened 2021-02-10
Skip -ffixed-form flag

dalcde:
This is only effective when compiling fortran code, which is not what we are doing since we have applied f2c. Future versions of clang forbid specifying the flag when not compiling fortran.


rth (approved):

rth:
The build failure in scipy due to https://github.com/iodide-project/pyodide/issues/936 happens quite frequently lately.

PR #1226, opened 2021-02-10
Tell cpython we don't have waitpid

dalcde:
Instead of patching emscripten


rth (approved):
Thanks, indeed!

phorward (approved):

dalcde:
This is the (real) error message:
```
  File "/lib/python3.8/asyncio/__init__.py", line 8, in <module>
    from .base_events import *
  File "/lib/python3.8/asyncio/base_events.py", line 25, in <module>
    import subprocess
  File "/lib/python3.8/subprocess.py", line 684, in <module>
    class Popen(object):
  File "/lib/python3.8/subprocess.py", line 1724, in Popen
    def _internal_poll(self, _deadstate=None, _waitpid=os.waitpid,
AttributeError: module 'os' has no attribute 'waitpid'
```
triggered when `JsProxy` initialization imports asyncio

dalcde:
As noticed in https://github.com/iodide-project/pyodide/pull/1102#issuecomment-760200742 this breaks a *lot* of things. Perhaps we should instead upstream the patch.

Issue #1219, opened 2021-02-09
IO in a webworker without asyncio?


hoodmane:
I'm not sure if I should be commenting here or opening a new issue. But I think my question is pretty closely related to this, so I'll put it here.

I'm building a web page where students can edit Python code, then click "Run" to run it in pyodide. I started with all of this running in the main browser thread, which worked pretty well. As part of this, I added some glue code so that `input()` in Python became a `prompt()` in the browser. 

However, when pyodide runs in the main thread, my code cannot update the display (e.g., show output), and controls are locked, so I can't have a "Stop" button to kill Python code that runs out of control. So I am working on a version where pyodide runs as a web worker. I've got that working for most of what I want, except for getting input back from the user.

Specifically, I send messages to from the worker to the main browser thread to launch the prompt, then send a message back from the main thread to the worker with the result. The problem is, I can't see any way to get the Python code to wait for the return message. I think that may be possible in general, if I write all-async code. The problem is, the incoming code is not written for an async environment (e.g., the students will use `input()`, not `await input()`). So even if I patch `input()`, my new version has to be a sync function to give a result back to the main code. So then my patched `input()` can't `await` an async function, e.g., a JavaScript promise that clears when the message comes back from the main browser thread.

I tried using `asyncio.get_running_loop().run_until_complete()` to await the promise, but this doesn't do anything, as noted above. I've investigated using emscripten's [Asyncify framework](https://emscripten.org/docs/porting/asyncify.html), and I'm pretty sure I can get that to create a synchronous C function that will await a JavaScript promise. But I'm not sure if I can call the C function from Python in a way that is compatible with the Asyncify system. And even if I can, it will be pretty kludgy.

I'm now thinking about using shared memory between the browser thread and the worker, then running a busy loop in the worker that polls for a response via the shared memory (since I [don't think sleep is possible](https://github.com/iodide-project/pyodide/issues/97)). But I'd rather not use such a narrow solution and peg the processor that way.

So I just wanted to check, is there currently any path to get synchronous Python code to wait for a JavaScript promise or async function?

_Originally posted by @mfripp in https://github.com/iodide-project/pyodide/issues/1158#issuecomment-775632060_

hoodmane:
Your understanding of everything is pretty much accurate. 

[...]

Issue #140, opened 2018-09-05
Write http.client in terms of Web APIs


mdboom:
This might not even be possible given blocking issues.

However, if we could write [http.client](https://docs.python.org/3/library/http.client.html) in terms of Web APIs, we might be able to get things like `pip` partially working.  As it stands, Python libraries built on top of raw sockets don't (can't) work.

anshuldutt21:
Hi, I would like to contribute to this issue.

rth:
If I'm not mistaken, @kikocorreoso mentioned that another possibility could to look at the work done in [Brython](https://github.com/brython-dev/brython) where some of these standard library modules might be re-implemented in JavaScript.

I quickly checked, and for instance [http.client](https://github.com/brython-dev/brython/blob/master/www/src/Lib/http/client.py) according to git commit messages, is identical to upstream, while [_socket.py](https://github.com/brython-dev/brython/blob/master/www/src/Lib/_socket.py) doesn't do anything, but maybe I'm missing something.  Also https://github.com/brython-dev/brython/issues/1032#issuecomment-460934098 suggests that there are no replacements for low level network connectivity.  Also couldn't find anything about it in [Skulpt](https://github.com/skulpt/skulpt).

Anyway it would indeed be a good idea to read earlier discussion on this subject in the issue tracker of these projects.

kikocorreoso:
@rth my comment was more in the vein of using js libs instead of py libs when it makes sense. Some of them are maybe implemented in Brython, batavia, skulpt,..., and coul be reused in some way.

For instance, `re` was very slow as it was implemented in pure Python in Brython so @PierreQuentel reimplemented the functionality in JS.
https://github.com/brython-dev/brython/issues/1519

I suppose `re` is in WASM in PyOdide so maybe this example it is not very useful. I was thinking more in pure PY libs that have been rewritten in JS to adapt some behaviour to the browser/ for performance reasons, etc.

I don't know if this could help in terms of "DoNotReinventTheWheel", performance,...

rth:
> this could help in terms of "DoNotReinventTheWheel"

Yes, absolutely. Thanks for your comment! We should definitely look at what could be used/adapted in JS before implementing stuff :)

hoodmane:
The main issue we have to deal with is the fact that http.client is a
synchronous api and the relevant web APIs are asynchronous. I implemented a
small piece of the http.client api on one of my comlink/syncifiers branches
as a proof of concept and I think it works quite well.

[...]

PR #1228, opened 2021-02-10
Remove `pyodide_py.open_url`

hoodmane:
This function is pretty simple and really has nothing to do with core pyodide purposes. We aren't wrapping the entire javascript standard library, so why do we specifically need a wrapper for `XMLHttpRequest`? Also this makes a blocking request which is [deprecated on the main thread](https://developer.mozilla.org/en-US/docs/Web/API/XMLHttpRequest/Synchronous_and_Asynchronous_Requests#synchronous_request).

I think we should add an faq entry like the following:

## How can I load a url?
Pretty much the same way as you would in javascript. Use `fetch` for simple cases. For more complicated uses, use `XMLHttpRequest`.
```javascript
await pyodide.eval_code_async(`
   from js import fetch
   text = await (await fetch("some_url")).text()
`);
```
### What if I don't want to use `async`?
You can make a blocking call to `XMLHttpRequest`, for instance:
```python
def open_url(url):
    from js import XMLHttpRequest
    from io import StringIO
    req = XMLHttpRequest.new()
    asynchronous = False # make the request synchronous
    req.open("GET", url, asynchronous)
    req.send(None)
    return StringIO(req.response)
```
Note that blocking `XMLHttpRequest` calls are [deprecated on the main browser thread](https://developer.mozilla.org/en-US/docs/Web/API/XMLHttpRequest/Synchronous_and_Asynchronous_Requests#synchronous_request), so if you are going to do this you should run pyodide on a webworker.

### DIFF
--- a/docs/usage/api-reference.md
+++ b/docs/usage/api-reference.md
@@ -14,7 +14,6 @@ Backward compatibility of the API is not guaranteed at this point.
    pyodide.as_nested_list
    pyodide.eval_code
    pyodide.find_imports
-   pyodide.open_url
    pyodide.JsException
    pyodide.register_js_module
    pyodide.unregister_js_module
--- a/packages/pandas/test_pandas.py
+++ b/packages/pandas/test_pandas.py
@@ -62,11 +62,16 @@ def test_load_largish_file(selenium_standalone, request, httpserver):
 
     selenium.run(
         f"""
-        import pyodide
         import matplotlib.pyplot as plt
         import pandas as pd
+        def open_url(url):
+            from js import XMLHttpRequest
+            req = XMLHttpRequest.new()
+            req.open("GET", url, False)
+            req.send(None)
+            return StringIO(req.response)
 
-        df = pd.read_json(pyodide.open_url('{request_url}'))
+        df = pd.read_json(open_url('{request_url}'))
         assert df.shape == ({n_rows}, 8)
-    """
+        """
     )
--- a/packages/pandas/test_pandas.py
+++ b/packages/pandas/test_pandas.py
@@ -66,6 +66,7 @@ def test_load_largish_file(selenium_standalone, request, httpserver):
         import pandas as pd
         def open_url(url):
             from js import XMLHttpRequest
+            from io import StringIO
             req = XMLHttpRequest.new()
             req.open("GET", url, False)
             req.send(None)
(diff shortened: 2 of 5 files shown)


rth:
> Use fetch for simple cases.

I see you point. The thing is most simple cases in Python are sync, and asking users to copy paste a snippet of code every time they want to download a files is not ideal. The ideal case would be some partial inplace replacement of existing http.client https://github.com/iodide-project/pyodide/issues/140 . The `pyodide.open_url` has limitations, and in particular it doesn't work with binary files https://github.com/iodide-project/pyodide/issues/400 however it might be better than ask users to copy paste 8 LoC from the FAQ?

hoodmane:
> The thing is most simple cases in Python are sync

I think we should try to encourage people to use asnychronous IO as much as possible. Because we are currently constrained to be single threaded, blocking IO doesn't make a whole lot of sense. I guess it prevents support for existing python libraries.

> however it might be better than ask users to copy paste 8 LoC from the FAQ?

Unfortunately I think you are right that it is better to keep this.
In particular it's important to consider education-related use cases. Requiring async may be considered beginner-unfriendly (e.g., #1219 the issue is that for education reasons he wants to avoid async).

> The ideal case would be some partial inplace replacement of existing http.client #140 

One promising direction is that it appears that with [`Atomics.wait`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Atomics/wait) we can get back blocking IO in a webworker. This will require us to set up a fair amount of infrastructure to get it to work though and like most good things is unavailable on safari.

dalcde:
I agree that a synchronous version of fetch is useful, especially since
blocking on async javascript function seems difficult at the moment.

hoodmane:
Yeah I think I'll close this. Perhaps I should open a discussion about using `Atomics.wait` to allow more synchronous IO on a webworker, for instance for use cases like #1219. That will have to wait until we get a basic webworker API set up first.

rth:
> Atomics.wait we can get back blocking IO in a webworker. This will require us to set up a fair amount of infrastructure to get it to work though and like many good things is unavailable on safari.

Support was merged to WebKit I think https://github.com/WebKit/WebKit/commit/0587157751f7e3bc724eb41dbab986c978d1fd02 so it should be available there soon too I guess.

hoodmane:
That's good to hear!

PR #1231, opened 2021-02-11
Rename deepCopyToJavascript -> toJs & rm shallowCopyToJavascript

hoodmane:
As discussed in #1192. Closes #1208.

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -98,8 +98,8 @@ class SeleniumWrapper:
         return self.run_js(
             f"""
             let result = pyodide.runPython({code!r});
-            if(result && result.deepCopyToJavascript){{
-                let converted_result = result.deepCopyToJavascript();
+            if(result && result.toJs){{
+                let converted_result = result.toJs();
                 result.destroy();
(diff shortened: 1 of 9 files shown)


phorward (approved):
Hello @hoodmane,
Thanks for the renaming, I'm fine with this.
Is it maybe useful and possible to provide tests with deep-parameter setting > 1 ?

hoodmane:
Thanks for the review @phorward!

> Is it maybe useful and possible to provide tests with deep-parameter setting > 1 ?

Yes, both useful and possible. These tests already exist in disguise, in the form of `test_python2js_with_depth` in `test_typeconversions`. I think I will convert that one to use `toJs` and remove the private test entrypoint in `python2js` now that there is a public way to access this API.

hoodmane:
I am going to need to update `test_python2js_with_depth` in #1234 anyways, so I think this PR should be merged as-is and I can do the cleanup in #1234.

rth:
Thanks! Sounds good to me as well.

Issue #1232, opened 2021-02-11
Language question: nomenclature for "type conversion"


hoodmane:
In #1231 based on discussion in #1192 we introduced the word "conversion" to describe the behavior of `toJs` of converting a native object from one language into a native copy in the other language. We use the term "proxy" to describe producing an object in one language that redirects requests into the other language. The `python2js` and `js2python` functions will most immutable objects and proxy most everything else. The current section of the documentation named "Type Conversions" is describing the behavior of `python2js` and `js2python`. If we also use the word "conversion" to describe the behavior of  `toJs`, then we're using the same word "conversion" to mean two different things.

Anyone have an idea of how to resolve this name clash?

daoxian:
The word "replicate" means to reproduce something, to construct a copy of something, to make a facsimile".  The objects or instances in Python are likely but not mandatorily totally the same as the equivalents that in JS. We expect a similar behavior between them. So I think "replicate" or "replication/replica" may be a candidate.

casatir:
Replicate sounds good. I also thought about "mirror".

hoodmane:
"mirror" is similar to "proxy", "reflect". "replicate" suggests "duplicate" / "copy". The nice thing about "convert" is that you normally "convert" from one thing to another "I converted the image from a jpg to a png" or "I converted the object from python to javascript". 

How about type translations, since we are translating from one language to another?

hoodmane:
So in particular, the idea is that `python2js` and `js2python` will do "type translation". Depending on the object, to translate we either "proxy" or "convert" it as appropriate. We choose to proxy things as much as possible when in doubt. If we "proxy" the object, it *may* be possible to explicitly cause it to be "converted" with the `toJs` and `toPy` APIs.

I think this sounds okay?

casatir:
Crystal clear!

hoodmane:
Okay I am going with "type translation" for now then. Thanks @daoxian @casatir for the feedback!

Issue #1221, opened 2021-02-09
Soundness issues in buffer conversions code


hoodmane:
Currently if one converts a 1d ndarray from Python ==> javascript, `_python2js_shareable_buffer_recursive` produces a `TypedArray` object that shares memory with the `ndarray` but without associating the lifetime of the `TypedArray` with the backing memory in any way, so the memory is borrowed. If the Python object is later garbage collected, then the memory backing this `TypedArray` will be freed, but without any warning to someone who stored the `TypedArray`. The memory may then be reallocated to something else, potentially leading to memory corruption. If the `TypedArray` is reimported into Python it yields a Python `memoryview` which also points to the same borrowed memory.

I think the only reason this doesn't happen more often is that we leak a lot of memory and people don't use the buffer conversions that much.

I think in the case that the user attempts to import a `TypedArray` on the Wasm heap into python, we should actually throw an error because that memory has been malloc'd and it may eventually be freed but we have no way of knowing when it is freed. I think we need some extra Javascript type which is a `TypedArray` referencing owned memory with a `release` method to let go of the memory when it is done. If the memory has a static lifetime then the `release` method can be a no-op. Then this pair `(TypedArray, release)` can be safely imported into Python as a Buffer with a `bf_release` implementation that calls the `release` function. 

This obviously will require a more carefully designed API, but the current system simply does not work.

See PR #1215 which partially addresses this. @daoxian

hoodmane:
Example:
```python
pyodide.runPython("x = b'123'");
a = pyodide.pyimport("x"); // a ==> Uint8ClampedArray(3) [49, 50, 51]
pyodide.runPython("del x");
a // ==> Uint8ClampedArray(3) [64, 0, 0]
```

daoxian:
Roger that and thanks a lot.

daoxian:
Some extra comments: In most cases this API (pyodide.getRawBuffer) works well.  When an numpy ndarray is NOT contingious after some complicated ops, the raw buffer may possibly become  unrecognizable by `cv.matFromArray()`, which will cause a "array is too long" exception.  The resolution is to make the numpy ndarray contingious again by `numpy.ascontingiousarray()`  before passing it as a argument to that opencv method.  And another not-so-efficient way is to convert it into bytes before passing into JS.  However, the issue has nothing to do with the  API.

hoodmane:
For instance if you say: `myslice = myarray[::2]`, does `cv.matFromArray` have a problem with it already?

[...]

Issue #1203, opened 2021-02-06
Performance issues in buffer conversions from javascript to python 


hoodmane:
since in my case from time to time in python I have to aquire arrays returned from JS methods,  i.e.,  a huge image array returned by some method of opencv.js which is now lacked in python packages.  The cost of automatic copying the array data from JavaScript to Python is too much, and can't be beared in some realtime computation cases.

_Originally posted by @daoxian in https://github.com/iodide-project/pyodide/issues/1192#issuecomment-774496411_

hoodmane:
@daoxian If you wouldn't you mind posting a small code example, it would be helpful.

daoxian:
> @daoxian If you wouldn't you mind posting a small code example, it would be helpful.

Sure. 

In python, I want to resize a image in opencv. Currently no opencv present in pyodide package , so I have to call opencv.js which residents in JS:

>             print(im_patch.size, im_patch.shape)  # m_patch is an numpy ndarray stands for an image
>             im_patch_bytes = im_patch.tobytes()   # conver to bytes in order to avoid mem copy when converting to JS arg of resize_img()
>             scaled_img_raw = js.g_tracker.resize_img(im_patch_bytes, (im_patch.shape[0], im_patch.shape[1]), (model_sz, model_sz))
>             im_patch = np.asarray(scaled_img_raw, dtype=np.uint8).reshape(model_sz, model_sz, 3)


In  JavaScript class method:
>   resize_img (img_bytes /* auto converted to Uint8ClampedArray already */, from_shape, to_shape) {
>     const img=cv.matFromArray(from_shape[0], from_shape[1], cv.CV_8UC3, img_bytes);    
>     const scaledImg = new cv.Mat();
>     cv.resize(img, scaledImg, new cv.Size(to_shape[1], to_shape[0]), 0, 0, cv.INTER_NEAREST);
>     return new Uint8ClampedArray(scaledImg.data); 
>   }

As to why I don't use Pillow to resize an image is due to its poor performance.

daoxian:
The case is that I'll migrate an `Object Tracking Algorithm` written in python to JavaScript so as to run it in the browser. It's a realtime tracking algorithm. With the input of a streaming of images, it should process every image and run as fast as possible.

hoodmane:
Thanks! This is a very interesting example.

In your particular use case there is a way to fix the performance. I think roughly the following should work:
```javascript
function alloc_matrix_3d(shape0, shape1, shape2){
[...]

PR #1233, opened 2021-02-11
Update documentation for type translations

hoodmane:
This is a draft with my work on updating the section  of the documentation formerly called "type conversions". In #1232 @casatir and I have both voted to change the terminology to "type translations" and reserve the word "conversion" for producing a native object in the target language (e.g., for the behavior of the `JsProxy.to_py` and `PyProxy.toJs` APIs that we've discussed in #1192). I've included this nomenclature change here.

The documentation here includes behaviors in some of my unmerged PRs and also some changes that I haven't opened a PR for yet. Before merging we should try to ensure that specified behavior and the actual behavior agree.


docs/usage/type-translations.md
```diff
@@ -0,0 +1,252 @@
+(type-translations)=
```
rth, line 1: Could we for now keep this file as `type-conversions`, so that Github show the diff properly? Then later change the file name.

hoodmane: Sure, will do that. I think the diff is pretty large either way though.

hoodmane:
Okay this got screwed up somehow, for some reason a bunch of files are deleted from the commit...

docs/usage/type-translations.md
hoodmane: Does github generally not understand `git mv`? It's funny because `git diff` gives a completely clear diff for these things.

docs/usage/type-conversions.md
```diff
@@ -1,114 +1,186 @@
...
-they will leak. To do this, call `.destroy()` on the object, after which Javascript will no longer have access to the object.
+### Proxying from Javascript into Python
+
+When most Javascript objects are translated into Python a `JsProxy` is returned.
+The following operations are currently supported on a `JsProxy`. (More should be
+possible in the future -- work is ongoing to make this more complete):
+
+| Python                    | Javascript             |
+|---------------------------|------------------------|
+| `repr(proxy)`             | `x.toString()`         |
```
rth, line 95: More usual would be
```suggestion
| `str(proxy)`             | `x.toString()`         |
```
It would call `__repr__` anyway I think?

docs/usage/type-conversions.md
```diff
@@ -1,114 +1,186 @@
...
+| `proxy.get(foo)`           | `x[foo]`                 |
+| `proxy.set(foo, bar)`      | `x[foo] = bar`           |
+| `proxy.delete(foo)`        | `del x[foo]`             |
+| `x.type`                   | `type(x)`                |
+| `x[Symbol.iterator]()`     | `iter(x)`                |
+| `x.next()`                 | `next(x)`                |
+| `await x`                  | `await x`                |
+| `Object.entries(x)`        |  `repr(x)`               |
+
+An additional limitation is that when proxying a Python object into Javascript,
```
rth, line 177: This could be a warning or a note see [admonitions](https://myst-parser.readthedocs.io/en/latest/using/syntax-optional.html?highlight=admonition#code-fences-using-colons) in myst parser. 

Generally we should maybe use those more in the docs.

rth (commented):
Thanks for thus nice write-up! 

A few comments / questions below.

docs/usage/type-conversions.md
```diff
@@ -1,114 +1,186 @@
...
+between the two languages. Depending on the type of the object we either
+translate it by implicitly "converting" it or by "proxying" it. By "converting"
+an object we mean producing a new object in the target language which is the
+equivalent of the object from the source language, for example converting a
+Python string to the equivalent a Javascript string. By "proxying" an object we
+mean producing a special object in the target language that forwards requests to
+the source language. When we proxy a Javascript object into Python, the result
+is a `JsProxy` object. When we proxy a Python object into Javascript, the result
+is a `PyProxy` object. A proxied object can be explicitly converted using the
+explicit conversion methods `JsProxy.to_py` and `PyProxy.toJs`.
```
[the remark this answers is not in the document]
hoodmane: I have a draft implementation on a local branch, will be a PR soon.

docs/usage/type-conversions.md
hoodmane: Confirmed that it calls `__repr__`. (Incidentally, maybe we should try to make `__repr__` return a better quality result for standard objects at some point. `toString` is shockingly low quality.)

docs/usage/type-conversions.md
```diff
@@ -1,114 +1,186 @@
...
+
+Python to Javascript translations occur:
+
+- when returning the final expression from a {any}`pyodide.runPython` call,
+- when using {any}`pyodide.pyimport`,
+- when passing arguments to a Javascript function called from Python,
+- when returning the results of a Python function called from Javascript,
+- when indexing a `PyProxy`
```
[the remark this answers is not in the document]
hoodmane: How about "when accessing fields of a `PyProxy`"? I mean `proxy.x` or `proxy["x"]`.

docs/usage/type-conversions.md
```diff
@@ -1,114 +1,186 @@
...
-
-The following basic types are implicitly converted between Javascript and
-Python. The values are copied and any connection to the original object is lost.
+- passing arguments to a Python function called from Javascript
+- returning the result of a Javascript function called from Python
+- when indexing a `JsProxy`
+
+## Round trip conversions 
+Translating an object from Python to Javascript and then from Javascript back to
+Python is guaranteed to give an object that is equal to the original object
```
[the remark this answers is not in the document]
hoodmane: Once #1234 is merged this will be true.

docs/usage/type-conversions.md
```diff
@@ -1,114 +1,186 @@
...
+(with the exception of `nan` because `nan != nan`). Furthermore, if the object
+is proxied into javascript, then translation back unwraps the proxy, and the
+result of the round trip conversion `is` the original object (in the sense that
+they live at the same memory address).
+
+Translating an object from Javascript to Python and then from Python back to
+Javascript gives an object that is `===` to the original object (with the
+exception of `NaN` because `NaN !== NaN`, and of `null` which after a round trip
```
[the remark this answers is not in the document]
hoodmane: This is not quite true yet, there is one additional exception for TypedArrays:
```javascript
a = new Uint8Array(10);
b = pyodide.runPython(`
	from js import a
	a
`)
```
`b` is a `PyProxy` of a `memoryview` and the `memoryview` is backed by a copy of the data from `a` not the original data. Planning to fix this as part of the other buffer adjustments to address #1202, #1203, #1221.

docs/usage/type-conversions.md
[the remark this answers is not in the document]
hoodmane: #1244

rth:
I think we can merge this soon?

hoodmane:
I think this needs to wait on merging #1186 and #1264 because they implement some of the logic described here. I should also add tests checking the assertions about round trip translations giving back the object we started with. I think once #1186 and #1264 are merged, I can add tests for the assertions made in the type-translations docs and then we can merge this. This will leave inaccuracies about buffers to be cleaned up in the future.

Other than that, some additional issues that should be addressed in future work are:

1. Name collisions between proxy mixin methods and proxied values (need to explain what happens in this case and make sure the logic actually works the way we claim it does).
2. Buffer conversions: buffer conversion from Python to Js should give a `PyProxy` with a `getRawBuffer` API #1215. Buffer conversion from Javascript to Python should give a `JsProxy` with APIs to copy data into or out of the `JsProxy` into native Python buffers. Furthermore if the Javascript buffer was on the Wasm heap, we should throw an error. If people want to make unsafe raw writes to the Python heap from Python, they should have to be explicit about wanting that.
3. Automatic garbage collection of `PyProxy` when the GC apis are available. Without automatic GC, the current `toJs` method is a bit of an unpleasant API because it creates memory leaks that are impractical to clean up. In general, the `PyProxy` is very leaky.
4. Converting bigint into Python? Currently bigints are converted to `JsProxy` which has a `__repr__` that looks like a normal integer but there isn't any practical way to use it at all. Well I guess you can do `int(repr(x))`.

rth:
> I think this needs to wait on merging #1186 and #1264 because they implement some of the logic described here.

Those are now done, and I think we should merge this. Could you please resolve conflicts? Generally it would probably be better to go back to a situation where type conversion PRs also update the documentation otherwise it's difficult to make sure it's in sync.

> Other than that, some additional issues that should be addressed in future work are:

Could you copy that checklist to a new (or an existing) issue?

hoodmane:
Okay so the current state of this is:
1. It makes general claims that are incorrect when applied to buffers (round trip conversions, etc). My PRs about buffers will make these true.
2. I added a bunch of stuff about memory leaks. I think this could be improved by moving it all into a different section. #1306 will improve things a lot, since we can say "if the browser is not Safari, probably everything will get collected eventually"

hoodmane:
I also changed the example of the `destroy` api. The previous example had a leak:

```js
let foo = pyodide.pyimport('foo');
foo.call_method(); // creates a temporary PyProxy for call_method and leaks it
foo.destroy();
foo.call_method();
```

rth (approved):
Thanks! I think we can adjust the remaining points together with the associated PRs.

PR #1234, opened 2021-02-11
Don't implicitly convert tuple and bytes objects into Javascript

hoodmane:
`tuple` and `bytes` objects are immutable Python objects with no corresponding immutable type in Javascript. For this reason, it is impossible to do a round trip conversion of `tuple` and `bytes` through Javascript and have it come out as an equivalent Python object. I think it's better to preserve as much as possible the invariant that a round trip translation gives back the original object. People who want the idiomatic Javascript object can then use the `toJs` API.

The current implicit conversions for buffers has various limitations (#1221, #1202). I'm hoping to replace it with a new explicit `PyBufferView` API see #1215.

### DIFF
--- a/src/core/python2js.c
+++ b/src/core/python2js.c
@@ -230,7 +230,7 @@ _python2js_dict(PyObject* x, PyObject* map, int depth)
   } while (0)
 
 static JsRef
-_python2js_immutable(PyObject* x, PyObject* map, int depth)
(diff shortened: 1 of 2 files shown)


rth (approved):
Sounds good to me, thanks! Again we should add a changelog entry once the PR with the new behavior is merged.

PR #1237, opened 2021-02-12
Add PyProxy.type

hoodmane:
I added a field `PyProxy.type` that returns the fully qualified name of the type. For example `list`, `dict`, `array.array`, `_io.StringIO`. If you say `repr(type(x))` the answer will be `"<class 'something'>"` and then `xproxy.type` is `"something"`. (This is usually equal to `type(x).__module__ + "." + type(x).__name__` except for builtins when it is equal to `type(x).__name__`.)

This is the partner to the `JsProxy.typeof` api.

I also added tests (and did some minor formatting on some of the existing type conversions tests).

### DIFF
--- a/src/core/pyproxy.c
+++ b/src/core/pyproxy.c
@@ -217,7 +217,8 @@ _pyproxy_iter_fetch_stopiteration()
 }
 
 JsRef
-_pyproxy_type(PyObject* ptrobj){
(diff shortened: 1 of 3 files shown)


rth (approved):
Thanks @hoodmane ! I would be good to add a changelog entry but we can do it in a follow up PR for changelog refactoring.

PR #1239, opened 2021-02-12
Use async functions for languageLoaderPlugin

dalcde:
This allows us to use more conventional "modern" javascript.

In the future, I hope to make languageLoaderPlugin a function instead of a promise (which would be very easy given this commit), so that users can supply a custom Module object. One useful application would be to supply stdout/stderr handlers. This would be a breaking change so I'll leave it for antoher time.

### DIFF
--- a/src/pyodide.js
+++ b/src/pyodide.js
@@ -2,7 +2,7 @@
  * The main bootstrap script for loading pyodide.
  */
 
-globalThis.languagePluginLoader = new Promise((resolve, reject) => {
+globalThis.languagePluginLoader = (async () => {
   let Module = {};
   // Note: PYODIDE_BASE_URL is an environement variable replaced in
   // in this template in the Makefile. It's recommended to always set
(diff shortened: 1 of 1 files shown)


src/pyodide.js
```diff
@@ -619,9 +619,24 @@ globalThis.languagePluginLoader = new Promise((resolve, reject) => {
...
+  let moduleLoaded = new Promise(r => Module.postRun = r);
+
+  const scriptSrc = `${baseURL}pyodide.asm.js`;
+
+  await loadScript(scriptSrc);
+
+  // The emscripten module needs to be at this location for the core
+  // filesystem to install itself. Once that's complete, it will be replaced
+  // by the call to `makePublicAPI` with a more limited public API.
+  self.pyodide = await pyodide(Module);
```
hoodmane, line 632: This is not really related to this PR, but I think it would be good to use a different export name:
```sh
-s EXPORT_NAME="'pyodide'"
```
Maybe `make_pyodide_module` or something like that. It's explained that there is a reason that we first put `Module` at `pyodide` and then move it to `pyodide._module`, but there isn't any important reason that we have to reuse the name `pyodide` for three different things?

hoodmane (approved):
Thanks @dalcde, I think this small change makes a huge improvement to readability.

hoodmane:
See also PR #277.

rth (approved):
Thanks!

PR #1246, opened 2021-02-12
DOC Update project subtitle to be more general

rth:
Closes https://github.com/iodide-project/pyodide/issues/553

Changes the project subtitle to be more general.

From 
> The Python scientific stack, compiled to WebAssembly.

to

> Python and the scientific stack, compiled to WebAssembly.

Or maybe 

> Python, and the scientific Python stack, compiled to WebAssembly.

(not sure if we need a comma after Python).

The vast majority of users don't use the scientific stack and just use Python with Pyodide (e.g. for teaching). Also a number of included packages are not scientific packages, so maybe we should go with a more general

> Python, compiled to WebAssembly

as suggested in https://github.com/iodide-project/pyodide/issues/553

WDYT?

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,11 +8,11 @@
 [![Build Status](https://circleci.com/gh/iodide-project/pyodide.png)](https://circleci.com/gh/iodide-project/pyodide)
 [![Documentation Status](https://readthedocs.org/projects/pyodide/badge/?version=latest)](https://pyodide.readthedocs.io/?badge=latest)
 
-The Python scientific stack, compiled to WebAssembly.
+Python and the scientific stack, compiled to WebAssembly.
 
 ## What is Pyodide?
 
-**Pyodide** brings the Python 3.8 runtime to the browser via WebAssembly, along with the Python scientific stack including NumPy, Pandas, Matplotlib, parts of SciPy, and NetworkX. The [`packages` directory](https://github.com/iodide-project/pyodide/tree/master/packages) lists over 35 packages which are currently available.
+**Pyodide** brings the Python 3.8 runtime to the browser via WebAssembly, along with the Python scientific stack including NumPy, Pandas, Matplotlib, SciPy, and scikit-learn. The [`packages` directory](https://github.com/iodide-project/pyodide/tree/master/packages) lists over 75 packages which are currently available. In addition it's possible to install pure Python wheels from PyPi.
 
 **Pyodide** provides transparent conversion of objects between Javascript and Python.
 When used inside a browser, Python has full access to the Web APIs.
@@ -23,8 +23,7 @@ browser**.
 
 ## Try Pyodide (no installation needed)
 
-Try the [iodide demo notebook](https://alpha.iodide.io/notebooks/300/) or fire
-up a [Python REPL](https://pyodide-cdn2.iodide.io/v0.16.1/full/console.html) directly in your
+Try pyodide in [Python REPL](https://pyodide-cdn2.iodide.io/v0.16.1/full/console.html) directly in your
 browser.
 
 For further information, look through the [documentation](https://pyodide.readthedocs.io/).
--- a/README.md
+++ b/README.md
@@ -8,7 +8,7 @@
 [![Build Status](https://circleci.com/gh/iodide-project/pyodide.png)](https://circleci.com/gh/iodide-project/pyodide)
 [![Documentation Status](https://readthedocs.org/projects/pyodide/badge/?version=latest)](https://pyodide.readthedocs.io/?badge=latest)
 
-Python and the scientific stack, compiled to WebAssembly.
+Python with the scientific stack, compiled to WebAssembly.
 
 ## What is Pyodide?
 
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -6,16 +6,15 @@
 Pyodide
 =======
 
-The Python scientific stack, compiled to WebAssembly.
+Python and the scientific stack, compiled to WebAssembly.
 
 .. note::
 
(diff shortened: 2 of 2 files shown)


README.md
```diff
@@ -8,11 +8,11 @@
 [![Build Status](https://circleci.com/gh/iodide-project/pyodide.png)](https://circleci.com/gh/iodide-project/pyodide)
 [![Documentation Status](https://readthedocs.org/projects/pyodide/badge/?version=latest)](https://pyodide.readthedocs.io/?badge=latest)
 
-The Python scientific stack, compiled to WebAssembly.
+Python and the scientific stack, compiled to WebAssembly.
 
 ## What is Pyodide?
 
-**Pyodide** brings the Python 3.8 runtime to the browser via WebAssembly, along with the Python scientific stack including NumPy, Pandas, Matplotlib, parts of SciPy, and NetworkX. The [`packages` directory](https://github.com/iodide-project/pyodide/tree/master/packages) lists over 35 packages which are currently available.
+**Pyodide** brings the Python 3.8 runtime to the browser via WebAssembly, along with the Python scientific stack including NumPy, Pandas, Matplotlib, SciPy, and scikit-learn. The [`packages` directory](https://github.com/iodide-project/pyodide/tree/master/packages) lists over 75 packages which are currently available. In addition it's possible to install pure Python wheels from PyPi.
```
rth, line 15: scikit-learn probably has a larger user base than networkx, and my personal goal is to make scikit-learn work well :)

docs/index.rst
```diff
@@ -24,8 +23,8 @@ Using Pyodide
 =============
 
 Pyodide may be used in several ways: directly from JavaScript, or to execute
-Python scripts asynchronously in a web worker. Although still experimental,
-additional packages may be installed from PyPI to be used with Pyodide.
+Python scripts asynchronously in a web worker. Additional packages may be
```
rth, line 26: micropip is as experimental as everything else in pyodiode, I'm not sure we need to highlight it specifically.

hoodmane:
Like the commenter in #593, I mostly have no interest in most of the Python scientific stack. (I care about mathematical computing, but I don't do anything with floating point arithmetic, instead most of what I do is arithmetic in finite fields.)
However, when evaluating various web python implementations, what I was looking for was maximum compatibility with the real python ecosystem. So reading "Python scientific stack" I think (1) it has a bunch of stuff I don't use (2) they can use Python packages written in C. The latter is a big selling point for me, I don't think any of the alternatives offer anything like it. 

From a pedagogical perspective, the high degree of compatibility with native CPython is desirable because if you teach students in Pyodide, any difference between the behavior of Pyodide and the behavior of CPython is likely to end up confusing them. The fact that our primary goal is maximizing the compatibility (because you want a large number of packages to work) ends up making Pyodide a relatively good tool for teaching.

Anyways "Python and the scientific stack, compiled to WebAssembly" sounds good to me.

hoodmane (approved):

README.md
```diff
@@ -8,11 +8,11 @@
 [![Build Status](https://circleci.com/gh/iodide-project/pyodide.png)](https://circleci.com/gh/iodide-project/pyodide)
 [![Documentation Status](https://readthedocs.org/projects/pyodide/badge/?version=latest)](https://pyodide.readthedocs.io/?badge=latest)
 
-The Python scientific stack, compiled to WebAssembly.
+Python and the scientific stack, compiled to WebAssembly.
```
phorward, line 11: What about

> Python, including the scientific stack, compiled to WebAssembly.

? Or 

> Python in WebAssembly - And no compromises!

okay latter one is a little joke maybe. For me, your original draft is also just fine.

docs/index.rst
phorward: This version sounds much better. Maybe we could also name it

> Additional pure Python packages may be installed from PyPI to be used with Pyodide.

rth:
> Python, including the scientific stack, compiled to WebAssembly.

Including is a bit too long IMO in the title. How about,

> Python with the scientific stack, compiled to WebAssembly

It might be a bit better than "and"

docs/index.rst
phorward: > Including might be a bit too long in the title, also I'm not sure what's the correct name for this grammatical construction but the two commas make it a bit harder to read IMO. How about,
> 
> > Python with the scientific stack, compiled to WebAssembly
> 
> It might be a bit better than "and"

That is nice. What about [this comment](https://github.com/iodide-project/pyodide/pull/1246#discussion_r575882456)?

phorward:
> Including might be a bit too long in the title, also I'm not sure what's the correct name for this grammatical construction but the two commas make it a bit harder to read IMO. How about,
> 
> > Python with the scientific stack, compiled to WebAssembly
> 
> It might be a bit better than "and"

That is nice. What about [this comment](https://github.com/iodide-project/pyodide/pull/1246#discussion_r575882774)?

rth:
> That is nice. What about this comment?

Thanks done!

docs/index.rst
```diff
@@ -23,8 +23,8 @@ Using Pyodide
 =============
 
 Pyodide may be used in several ways: directly from JavaScript, or to execute
-Python scripts asynchronously in a web worker. Additional packages may be
-installed from PyPI to be used with Pyodide.
+Python scripts asynchronously in a web worker. Additional Pure python packages
```
hoodmane, line 26: probably should be "pure" with a lower case p and "Python" with an upper case P.

rth: Yes, probably. I'll have to do another PR to improve the documentation now...

Issue #1245, opened 2021-02-12
toJs conversion of dict keys


hoodmane:
Dictionaries allow keys not to be a string, but in an `Object` all keys are strings. This may cause unexpected data loss:
```javascript
>>> pyodide.runPython(`{ 1 : 7, "1" : 9}`).toJs()
{1: 9}
>>> pyodide.runPython(`{ "1" : 9, 1 : 7}`).toJs()
{1: 7}
```
We could use a `Map` instead of an `Object` to fix this problem. 

Currently using a tuple key doesn't lead to the best result either:
```javascript
>>> pyodide.runPython(`{ (2,3) : 6}`).toJs()
{ "2,3" : 6 }
```
If the key is a `bytes object this is weird too:
```javascript
>>> pyodide.runPython(`{ b"abc" : 6}`).toJs()
{"97,98,99": 6}
```
One thing to consider is at least to use `str` on the key rather than doing `key.toJs().toString()` like we are doing now. Then at least we'd convert `{ (2,3) : 6}` to `{ "(2, 3)" : 6}`. The bytes example would give `{ "b'abc'" : 6}` which isn't great but is probably better than the current behavior.

One annoying thing about `Map` is that it compares on memory address:
```javascript
>>> let m = new Map();
>>> m.set([2,3], 7);
>>> m.get([2,3]); 
undefined // second [2,3] is not at the same memory address as the first
```
For this reason it might still be best to use `str` on the keys even when returning a `Map` except when integers are used as keys.

rth:
> For this reason it might still be best to use str on the keys even when returning a Map except when integers are used as keys.

Thanks for the analysis! That sounds very reasonable to me given the `Map` constraints. This should be a warning in the doc somewhere though.

hoodmane:
The other most reasonable option is to throw an error, maybe that would be better actually.

rth:
+1 for error in that case.

hoodmane:
Okay, I'll open a PR for this.

hoodmane:
Closed by #1248.

PR #1248, opened 2021-02-14
Fixed various minor issues in python2js

hoodmane:
This fixes several issues in python2js:

1. If a list has several copies of the same list, they were not converted to the same list. E.g.,
```javascript
let x = pyodide.runPython(`
a = []
b = [a,a,a]
`).toJs();
x[0] === x[1] // ==> False
```
([Wikipedia page about this bug](https://en.wikipedia.org/wiki/Billion_laughs_attack))

2. It attempted to convert any object satisfying `PySequence_Check` to a list but [`PySequence_Check` returns `1` for Python classes with a `__getitem__` method unless they are dict subclasses](https://docs.python.org/3/c-api/sequence.html?highlight=pysequence_chec#c.PySequence_Check) so this causes too many false positives to be a good choice IMO.
3. It converted `dict` to an `Object` but a `dict` can have any type of key whereas an `Object` can only admit strings. This converts instead to `Map` which can have any type of key. (Issue #1245)

I refactored to use error handling macros and to be generally more pedantic about error handling than before. I added a new error type `ConversionError` and raise this when problems arise in conversion.


packages/numpy/test_numpy.py
```diff
@@ -107,11 +107,6 @@ def assert_equal():
 
     assert selenium.run("np.array([True, False])") == [True, False]
 
-    selenium.run("x = np.array([['string1', 'string2'], ['string3', 'string4']])")
```
rth, line 110: What's the situation with numpy arrays containing strings after this PR?

src/core/python2js.c
```diff
@@ -190,89 +189,128 @@ _python2js_bytes(PyObject* x)
...
-    JsRef jsitem = _python2js_cache(pyitem, map, depth);
-    if (jsitem == NULL) {
-      _python2js_remove_from_cache(map, x);
-      Py_DECREF(pyitem);
-      hiwire_decref(jsarray);
-      return NULL;
-    }
-    Py_DECREF(pyitem);
+    FAIL_IF_NULL(pyitem);
+    FAIL_IF_NULL(jsitem = _python2js_cache(pyitem, map, depth));
```
rth, line 211: I know it works in C, but still it's maybe more readable with,

```suggestion
    jsitem = _python2js_cache(pyitem, map, depth);
    FAIL_IF_NULL(jsitem);
```

src/core/python2js.c
```diff
@@ -190,89 +189,128 @@ _python2js_bytes(PyObject* x)
   return hiwire_bytes(x_buff, length);
 }
 
+/** WARNING: This function is not suitable for fallbacks. If this function
+ * returns NULL, we must assume that the cache has been corrupted and bail out.
+ */
```
rth, line 194: So the cache is corrupted because we are never calling `_python2js_remove_from_cache` on failure unlike before, as far as I understand?

src/core/python2js.c
```diff
@@ -372,58 +445,65 @@ _python2js_cache(PyObject* x, PyObject* map, int depth)
...
+    }
+  } else {
+    PyErr_SetString(internal_error,
+                    "Internal error occurred in python2js_with_depth");
+  }
+  return NULL;
 }
 
 JsRef
 python2js_with_depth(PyObject* x, int depth)
```
rth, line 475: Could we add some docstrings for these functions while we are at it?

rth (commented):
Thanks @hoodmane ! A few minor comments / questions below. Otherwise looks good! 

Could you add a changelog entry? The title could also probably be updated to be more explicit than "fix minor issues"

src/core/python2js.c
hoodmane: Okay, will keep with that convention, was just experimenting a bit here.

src/core/python2js.c
hoodmane: That's correct, but the code was wrong before too. The minimal counterexample is a bit complicated though. Let me cook one up.

hoodmane: Actually now I think it was actually right before too. Confusing.

hoodmane: But yeah you understand correctly.

src/core/python2js.c
hoodmane: Okay, I added some docstrings and tried to improve names of things a bit.

packages/numpy/test_numpy.py
hoodmane: I'm confused. I must have thought it was correct for this test to fail because I deleted it, but I don't remember why.
The behavior is really weird and inconsistent though, definitely should be improved. I'm not sure what it should do though. Some examples:

```javascript
ajs = pyodide.runPython(`
   a = np.array([0, 1])
   a
`).toJs()
```
`ajs` is an `Int32Array` pointing to the same memory as `a` and suffering from #1221.

```javacript
bjs = pyodide.runPython(`
   b = np.array([True, False])
   b
`).toJs()
```
`bjs` is an Array containing `[true, false]`! Even though `b[0]` is of type `numpy.bool_`. `b[0]` actually makes a `PyProxy` such that when you call `toJs()` on it it turns to `true`. No idea how that happens, I will probably have to insert some print statements into the conversion code and trace through what the heck is going on.

Lastly 
```javacript
cjs = pyodide.runPython(`
   c = np.array(["abc", "def"])
   c
`).toJs()
```
is still a proxy.

rth: OK, let's address this in a separate issue.

rth (approved):
Could you resolve the merge conflict? Thanks!

phorward (approved):

PR #1250, opened 2021-02-14
DOC update to sphinx book theme

rth:
This started as looking into a way for adding badges on the what's new entry (as in [this example](https://scikit-learn.org/stable/whats_new/v0.24.html#legend-for-changelogs)). However because readthedocs theme does not use bootstrap, it didn't seem very easy.

So I ended up switching to the [Sphinx book theme](https://sphinx-book-theme.readthedocs.io/en/latest/index.html) which feels a bit more modern and is based on Boostrap 4. By the same people who made the MyST parser that we use. This,
 - resolves the issue on how to add API Changed/FIX/Feature badges in the changelog, while re-organizing it by submodule as was done in the previous release. 
 - also the rendering on wide screen seems a bit better
 - There are also interesting developments in this theme to interactively run code as part of the documentation (https://sphinx-book-theme.readthedocs.io/en/latest/launch.html#live-code-cells-with-thebe) with [Thebe](https://thebe.readthedocs.io/en/latest/) (the latter project doesn't seem to be developped anymore) but we can probably still hack it to make it work with pyodide instead of jupyter in the backend.

The downside is that the integration with readthedocs and in particular version management might be slightly worse.

## Current theme
(on a monitor with a 2560 × 1440 resolution)

image

## Updated theme

image

Rendered version [here](https://21273-122663163-gh.circle-artifacts.com/0/home/circleci/repo/docs/_build/html/index.html)

A theme using Bootstrap 4 also means that we can probably more easily can make a customized home page as part of the documentation https://github.com/iodide-project/pyodide/issues/1243

I'll make a follow up PR re-organizing the what's new entry by pyodide submodules once this one is merged.


hoodmane:
Looks nice. One thing about the readthedocs theme I don't particularly care for is the red color of the `inline code blocks`. I think the pink color looks a lot better. Also the orange color they use for comments is more of a highlight which is good.

If someone wants to waste a lot of time mucking around with parser rules, another interesting thing would be to try to hack the syntax highlighter so that it syntax highlights the stuff inside of `runPython` with a multiline uptick string as Python code.

phorward (approved):
Hi Roman, the new theme looks much better! Thanks for cleaning this up.
> If someone wants to waste a lot of time mucking around with parser rules, another interesting thing would be to try to hack the syntax highlighter so that it syntax highlights the stuff inside of `runPython` with a multiline uptick string as Python code.

I really like this idea. From IntelliJ-based IDEs, I know about a feature that a string can be annotated using `#language = JavaScript` or `//language = Python` for example right in front of, to mark it as some embedded language, but unluckily I think this isn't supported by myst yet.

Maybe open a separate issue for this?

hoodmane:
> Maybe open a separate issue for this?

Opened PR #1252.

PR #1252, opened 2021-02-15
Highlight embedded python code in docs

hoodmane:
I made new `Pyodide` and `HtmlPyodide` lexers that syntax highlight Python code inside of javascript and used them in the docs. Turns out there are disappointingly few instances of `runPython` in the docs though =(.


rth (approved):
Interesting, thanks!

PR #1253, opened 2021-02-15
Update faq on custom namespaces

hoodmane:
Since #1167 merged, there is a simpler way to use a custom namespace. This updates the FAQ with the new way. (I also added `pyodide_py` to the public API, I hadn't noticed it wasn't public.)


src/pyodide.js
```diff
@@ -329,17 +329,9 @@ globalThis.languagePluginLoader = (async () => {
...
-    'pyimport',
-    'runPython',
-    'runPythonAsync',
-    'version',
-    'registerJsModule',
-    'unregisterJsModule',
-    'setInterruptBuffer',
+    'globals', 'loadPackage', 'loadPackagesFromImports', 'loadedPackages',
+    'pyimport', 'runPython', 'runPythonAsync', 'version', 'registerJsModule',
+    'unregisterJsModule', 'setInterruptBuffer', 'pyodide_py'
```
rth, line 334: This linting change by clang-format doesn't make sense (and as far as I remember it's not the first time). Should we just apply clang format off, and keep one entry per line manually for the `PUBLIC_API`?

rth (approved):
Thanks!

src/pyodide.js
hoodmane: Yeah I was thinking about that. It seems to be a bit unstable about which it prefers here for some reason.

PR #1255, opened 2021-02-15
Re-organize changelog by submodule

rth:
The changelog for this version is getting long, and it should make it easier to find things.

Rendered version: https://21664-122663163-gh.circle-artifacts.com/0/home/circleci/repo/docs/_build/html/project/changelog.html

### DIFF
--- a/docs/_static/css/pyodide.css
+++ b/docs/_static/css/pyodide.css
@@ -1,0 +1,4 @@
+#release-notes .badge {
+    margin-right: 5px;
+    opacity: 0.7;
+}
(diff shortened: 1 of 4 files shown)


rth:
Hmm, it's a lot of colors, maybe we should at least make the badges to use more pastel colors, not sure.


image

hoodmane:
It'd be nice to increase the right margin on the badges if possible. As is the text around the badges looks pretty crowded.

rth:
Added some right margin, and added opacity so that badges are a bit less bright. Rendered verion [here](https://21771-122663163-gh.circle-artifacts.com/0/home/circleci/repo/docs/_build/html/project/changelog.html). What do you think?

hoodmane:
Looks good to me. Definitely better with the opacity.

Issue #1258, opened 2021-02-16
Not all tests pass in master branch


albertas:
I have executed all master df3861e branch  tests in the pre-built docker container and got 5 tests failing:
- FAILED src/tests/test_asyncio.py::test_await_pyproxy_eval_async[firefox]
- FAILED src/tests/test_asyncio.py::test_await_pyproxy_async_def[firefox] 
- FAILED src/tests/test_asyncio.py::test_await_pyproxy_eval_async[chrome] 
- FAILED src/tests/test_asyncio.py::test_await_pyproxy_async_def[chrome]
- FAILED src/tests/test_core_python.py::test_cpython_core[test_compile-chrome]

The first four failed with `NotImplementedError` and the last one failed with:

```
Traceback (most recent call last):
  File "/lib/python3.8/test/libregrtest/main.py", line 634, in main
    self._main(tests, kwargs)
  File "/lib/python3.8/test/libregrtest/main.py", line 704, in _main
    sys.exit(2)
SystemExit: 2
```

I suggest marking these 5 tests to be skipped to make all master tests pass. I managed to do so by:
1. Adding `@pytest.mark.skip` decorator for [test_await_pyproxy_async_def](https://github.com/iodide-project/pyodide/blob/master/src/tests/test_asyncio.py#L339) and [test_await_pyproxy_eval_async](https://github.com/iodide-project/pyodide/blob/master/src/tests/test_asyncio.py#L266).
2. Commenting out [this line]( https://github.com/iodide-project/pyodide/blob/master/src/tests/python_tests.txt#L129). This also removed one passing test.

rth:
> the pre-built docker container

Please try the regular (non-prebuilt) pyodide Docker container, after running `make clean-all` then `git clean -xdf` (to remove all untracked files in git).  There have been a lot of changes in the code base since the release, and the pre-built container would have build artifacts from a month ago. I'm not sure what happens if you mix those with the latest version of code from master -- probably nothing good. The prebuilt image is meant to be run with the last release not master.

albertas:
I have tried:

```
make clean-all
git clean -xdf
./run_docker
make
pytest src/ pyodide_build/ packages/*/test_*
```
And the same 5 tests still failed.

rth:
[...]

PR #1259, opened 2021-02-16
Fix test_asyncio by removing DumbLoop

hoodmane:
This fixes many of the test failures reported by @albertas in #1258. The problem was that the older tests in `test_asyncio` globally set up a `DumbLoop` which is a very feature poor event loop meant for low level asyncio testing, but they didn't restore the event loop when they were finished.

The newer tests are more integration tests of the asyncio testing features (they test the code added in all 3-4 of the major asyncio PRs mixed together) and they expect to use the typical `WebLoop`, not the `DumbLoop`.

~Anyways I switched to using a `dumb_loop` fixture that restores the normal event loop on exit.~ 
Edit: I removed `DumbLoop` entirely.
I also adjusted the fixtures in `test_console` to use the standard
```python
try:
   yield
finally:
   # cleanup
```
setup.

### DIFF
--- a/src/tests/test_asyncio.py
+++ b/src/tests/test_asyncio.py
@@ -8,6 +8,7 @@ import time
 from pyodide._base import eval_code_async
 import asyncio
 
+
 def test_await_jsproxy(selenium):
     selenium.run(
         """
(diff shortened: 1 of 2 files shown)


hoodmane:
On second thought I should probably just remove `DumbLoop` altogether...

rth (approved):
Thanks @hoodmane ! It's indeed a good idea to test with the default event loop directly.

It would be nice some mechanism to skip package building when only core tests are changed (or the doc).

PR #1261, opened 2021-02-17
Fix console.html completions

hoodmane:
I think #1167 broke `console.html` completions, this fixes it. I tried to figure out how to add a test against this, but it seemed annoying. Maybe we should make a `test_console` that actually runs in `console.html`?


phorward (approved):

rth:
It would have been nice to have some tests for console, but it doesn't look very straightforward indeed.

PR #1262, opened 2021-02-17
Conditionally skip unnecessary CI steps

rth:
This is an attempt to save CI resources by not not running build steps when they are not needed. Current rules are,
 1. do not build core (and packages) if only documentation or readme is changed.
 2. do not build packages if files if only `src/tests/*` or `src/pyodide-py/*` are changed

I'm still not sure whether it's better to,
 - perform build if any of the include file patterns match (currently implemented) 
 - skip build if all changed files are in the exclude file patterns


rth:
Generally this works as expected, however for now I haven't found a way on how cancel dependent jobs from a given CI jobs. So this script has to be re-run in each CI job. In addition it marks the job as completed successfully, when a job is skipped, which can be confusing if a job that should have run is skipped. The only way to detect such errors is currently to look at the logs. 

I still need to investigate why monkeypatching doesn't work in test_git_get_files_changed in CI.

hoodmane:
Presumably it still tests the packages if only `src/pyodide-py` changed? Is it necessary to rebuild packages if `src/core` changes? It's certainly necessary to test them, but I don't see that it's necessary to rebuild them (if it's possible to used a cached build of packages at all). 

Should there be some way to force the tests to run in a commit (for instance when preparing to deploy)? I guess one can always make a tiny change, but presumably this looks for actual changes in the files and won't work with `touch`? The checks will give the same value every time, does the CI API give an option to check whether the run is automatic or someone has manually requested to rerun the CI? In the manual case it'd be helpful to actually do everything.

rth:
Thanks for the feedback @hoodmane !

> Presumably it still tests the packages if only src/pyodide-py changed? Is it necessary to rebuild packages if src/core changes? It's certainly necessary to test them, but I don't see that it's necessary to rebuild them (if it's possible to used a cached build of packages at all).

Right now caching packages would require some work, we are already caching compilation with ccache, so it's probably simpler to rebuild, though I agree this can be improved.

> Should there be some way to force the tests to run in a commit (for instance when preparing to deploy)?

Yes, I need to add the ability to re-run everything based on some commit message. Maybe a `[build-all]` prefix in the commit message.

rth:
This won't be useful until we switch to a separate repo for packages IMO which would make this problem somewhat simpler. Closing.

Issue #693, opened 2020-06-25
Python-object is not garbage collected when used with EventListener


phorward:
Hi!

I have a special case here regarding garbage collection in combination with JavaScript's addEventListener and removeEventListener functions.

1. Creating an instance of a class, associating it with a DOM element and removing it again works - the object is entirely freed.
2. Creating an instance of a class, associating it with a DOM element, attaching an event to it and detaching it immediately after still holds a reference to it the object, even if it is explicitly removed with `del`

I guess, that the problem here is in JavaScript, which still holds a reference and this is normal with JavaScript. But I'm quite unsure, and would like to ask for further investigation, eventually its really a bug.

Test case is this Python script. Attached is [bug.txt](https://github.com/iodide-project/pyodide/files/4834270/bug.txt)
 to be renamed bug.html, which can be copied into the `build/` folder and directly be executed.

```python
from js import document
import weakref, gc

print("start")

class TestWrapper:
    def __init__(self):
        super().__init__()
        self.div = document.createElement("div")

    def testCallback(self, event):
        print("Got Event")

    def attach(self):
        print("attach")
        self.div.addEventListener("click", self.testCallback)

    def detach(self):
        print("detach")
        self.div.removeEventListener("click", self.testCallback)

    def __del__(self):
        print("I am removed")

print("---111---")

div = TestWrapper()
rw = weakref.ref(div)
gc.collect()
print("1 Weakref resolves to: %s" % rw())
del div
gc.collect()
print("1 Weakref after dropped reference resolves to: %s" % rw())  # << this is None, as expected!

print("---222---")

div = TestWrapper()
rw = weakref.ref(div)
div.attach()
del div
gc.collect()
print("2 Weakref to Wrapper-Class with active event listener: %s" % rw())
rw().detach()
gc.collect()
print("2 Weakref to Wrapper-Class with removed event listener: %s" % rw())  # << At least this should be None
```

The logging output in the JavaScript console is this:
```
Python initialization complete
start
---111---
1 Weakref resolves to: <TestWrapper object at 0x9a8a60>
I am removed
1 Weakref after dropped reference resolves to: None
---222---
attach
2 Weakref to Wrapper-Class with active event listener: <TestWrapper object at 0x9a8a60>
detach
2 Weakref to Wrapper-Class with removed event listener: <TestWrapper object at 0x9a8a60>
```

It can be reproduced both in Firefox 77.0.1 and Chromium Version 83.0.4103.97

rth:
[...]

PR #1264, opened 2021-02-18
Pyproxy mixins

hoodmane:
This PR improves several things about `PyProxy`:

1. Reduce stray properties to the absolute minimum. As much as possible attribute access should go to the Python object.
2. Extra methods defined on Proxies of particular types of Python objects should overridden by properties on the python object (as discussed in #1175)
3. Share prototypes between PyProxies of the same type to reduce memory usage.
4. Organization, explanations, more consistent behavior.

Prior to this PR, fields called "name", "length", or "prototype" all showed weird behavior ("prototype" would be shadowed entirely, "name" and "length" show inconsistent results with the "in" keyword). Likewise "caller" and "arguments" were in all cases shadowed. Now the "name" and "length" inconsistencies are fixed, "prototype" is no longer shadowed (and setting and deleting the "prototype" field works as expected) but "prototype" shows some minor (and unremoveable) inconsistency with the "in" keyword. "caller" and "arguments" are now only shadowed if the `PyProxy` wraps a callable Python object, in other cases they behave normally.

I also added a lot of comments and organized and simplified the logic for testing which abstract object protocols the Python object supports.

This is on top of #1175.


src/core/pyproxy.c
```diff
@@ -13,6 +13,136 @@ _Py_IDENTIFIER(add_done_callback);
...
+//
+// Object protocol wrappers
+//
+// This section defines wrappers for Python Object protocol API calls that we
+// are planning to offer on the PyProxy. Much of this could be written in
+// Javascript instead. Some reasons to do it in C:
+//  1. Some CPython APIs are actually secretly macros which cannot be used from
+//     Javascript.
+//  2. The code is a bit more concise in C.
+//  3. It may be preferable to minimize the number of times we cross between
```
phorward, line 132: IMHO we should avoid talking about "not being sure" about something in the code. I think the solution as it is now has its right to exist, therefore it is fine.

phorward (commented):
Hello @hoodmane ,
thank for this pull request, it looks very impressive. From my first view I don't  see any problems, the tests look nice on the mixins and the documentation is very well. The only "little" thing which bothers me are the comments about when you're not sure about something. I think these kind of thoughts must not be part of the source code, but that's only my opinion.

Separating the Javascript-parts from pyproxy.c into a new pyproxy.js could also be done within this pull request, but IMHO having all related C-JS-stuff in one file is okay.

phorward:
Addresses #1049, #1006, #693, I'll do a test soon on my original issue :+1:

hoodmane:
@phorward Thanks for the review.

I don't think this PR addresses the GC issues. I'll probably do that after this is merged.

phorward:
> I don't think this PR addresses the GC issues. I'll probably do that after this is merged.

Yes, I found this already out, and I'm very excited about that change.

src/core/pyproxy.c
phorward: @hoodmane what do you think about this?

hoodmane: I'm happy to change the comment. Should I just remove "I'm not sure"?

phorward:
Thank you @hoodmane that is ok for me. Let's wait for CI, and can you please merge the current master into your branch?

src/core/pyproxy.c
hoodmane: In general I think it is fine to have comments stating things like "I'm not sure". For reference, CPython 3.8.2 source has 27 comments saying variations of "I'm not sure", the emscripten source has 11.

hoodmane:
So what's the status on this? Is this okay to merge @phorward?

PR #1266, opened 2021-02-19
Use globals() for console.html

hoodmane:
This updates `console.html` to use the `globals()` scope directly like it used to. This improves debugging convenience: it's useful if variables defined in the console are available in the browser console (via `pyodide.globals` or `pyodide.runPython`) and vice versa.


casatir:
Hi @hoodmane.

As discussed in #1125, I think we should have two versions of `console.html`, showcase and dev. Non dev users does not really want to be bother with dirty `globals()` I guess, while dev could be really interested with improvements you propose.

In https://github.com/iodide-project/pyodide/pull/1125#issuecomment-758716874, we discussed about an URL parameter. With `dev=true`, we could init console as you do and let it as it was otherwise.

WDYT?

I can submit a PR implementing the `dev` URL flag in `console.html` if you want.

rth:
Thanks! I was going to open an issue about it.

>  think we should have two versions of console.html, showcase and dev. Non dev users does not really want to be bother with dirty globals() I guess

@casatir Could you remind me the motivation for keeping a separate namespace there? I reread but I'm still not fully sure about (aside from the fact that's it's just better to have a separate namespace).
One of the selling points of Python is to be able to mix Python and JS, and the REPL is IMO the easiest way to experiment with it. I would expect this to be of interest to users not just devs. 

Or alternatively is there any simple way to access Python objects from the JS console without changing the namespace? I guess having a separate namespace would make things like https://github.com/iodide-project/pyodide/issues/703 easier?

hoodmane:
> non dev users does not really want to be bother with dirty `globals()`

I agree, so we should make sure that `__main__.__dict__` namespace is not dirty. In the native repl the following code works:
```python
x = 2
import __main__
assert __main__.x == 2
```
so for the most consistency with the native repl, we should use the `__main__` namespace for the repl and figure out how to keep the `__main__` namespace clean of setup junk. I think this will lead to better less buggy code anyways.

In this direction, I have already fixed it so that system functions no longer leak variables or imported modules into the global namespace. I guess the code in `console.html` does leak into the `__main__` namespace though. Perhaps we should make a `pyodide.createModule`:

```javascript
function createModule(name, code){
    let new_module = pyodide.pyodide_py.eval_code(`
        import sys
        import types

        new_module = types.ModuleType('${name}')
        sys.modules['${name}'] = new_module
        new_module
    `);
    let new_module_dict = new_module.__dict__;
    pyodide.pyodide_py.eval_code(code, new_module_dict);
    new_module_dict.destroy();
    return new_module;
}
```
Then instead of `pyodide.runPython` in `console.html`, you could use ```pyodide.createModule("pyconsole", `...`);``` and that would leave the `PyConsole` class accessible by `from pyconsole import PyConsole` without leaking it into the `__main__` name space.

The other weird thing about `__main__.__dict__` still is that we do `__main__.__dict__.update(builtins.__dict__)`. I have plans to get rid of that soon (by adding extra logic in the namespace proxy instead).

> is there any simple way to access Python objects from the JS console without changing the namespace?

Probably you should make the namespace accessible on the `PyConsole` object, by adding after
https://github.com/iodide-project/pyodide/blob/9bedba0d06a90a56035ab55632bebe35c0a6b707/src/pyodide-py/pyodide/console.py#L124
`self.namespace = locals`.
Then if you wanted to access the variables from `PyConsole` you could do
```javascript
pyodide.runPython(`
   from js import pyconsole
   print(pyconsole.namespace["x"] + pyconsole.namespace["y"])
`);
```
or
```javascript
pyodide.pyodide_py.eval_code(`
   x + y
`, pyconsole.namespace);
```

rth:
> so for the most consistency with the native repl, we should use the __main__ namespace for the repl and figure out how to keep the __main__ namespace clean of setup junk.

We could. Though I think it's also good to keep the REPL here as simple as possible, and I'm not sure the additional complexity involved is worth it. Though I can be convinced otherwise. In a more immediate future I think we just need to be able to use the JS console in combination with this REPL in some way.

hoodmane:
@rth Does the most recent commit look like a simple enough solution to be worth it?

rth (approved):
yes, this looks nice, thanks!

rth:
@casatir Do you have any comments about the current version of this PR?

casatir:
Really sorry guys, I was off this days.

> @casatir Could you remind me the motivation for keeping a separate namespace there?

The main motivation was to have it clean at startup.

> so for the most consistency with the native repl, we should use the __main__ namespace for the repl and figure out how to keep the __main__ namespace clean of setup junk

Good point @hoodman, totally agree.

> Probably you should make the namespace accessible on the PyConsole object, by adding self.namespace = locals

This is already accessible using `self.locals`:

``` python
>>> from js import pyconsole
>>> pycosole.locals
```
should do what you want.

> @rth Does the most recent commit look like a simple enough solution to be worth it?

Really nice, thanks @hoodmane!

casatir (approved):

rth:
CI failures look genuine: I have restarted the CI to check and it's still failing.

hoodmane:
@rth Thanks, I'll look into it.

hoodmane:
@rth Okay, fixed it.

PR #1269, opened 2021-02-20
Use eval_code_async in runPythonAsync

hoodmane:
This updates `runPythonAsync` to use `pyodide_py.eval_code_async` internally. This has the effect that Python code run with `runPythonAsync` is executed with `PyCF_ALLOW_TOP_LEVEL_AWAIT`, which is what I and I think many other people expected `runPythonAsync` to do.

If this change to the code is approved, I'll add more appropriate updates to the docs.

It still makes the call to `pyodide.loadPackagesFromImports`. I think ideally we would remove the call to `loadPackagesFromImports` and have people do that separately if they want it. I believe that this would make `runPythonAsync` function more in the expected way.

### DIFF
--- a/docs/project/changelog.md
+++ b/docs/project/changelog.md
@@ -87,6 +87,8 @@ substitutions:
   `pyodide.loadPackage`, `pyodide.runPythonAsync` and
   `pyodide.loadPackagesFromImport`, then the messages are no longer
   automatically logged to the console.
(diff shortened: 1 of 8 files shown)


rth (approved):
Thanks @hoodmane ! LGTM. Do we need a what's new entry?

hoodmane:
I believe so, yes.

PR #1274, opened 2021-02-23
Use jquery v3.5.1 instead of v1.11.1

hoodmane:
Oddly enough `https://code.jquery.com/jquery-latest.js` does not point to the latest version of jquery, it points to version 1.11.1 (released in 2014). The jsdelivr link `https://cdn.jsdelivr.net/npm/jquery` points to 3.5.1. The terminal seems to work better with the up to date version of jquery, despite the fact that terminal docs suggested using the jquery-latest link.

Closes https://github.com/iodide-project/pyodide/issues/1276

### DIFF
--- a/src/templates/console.html
+++ b/src/templates/console.html
@@ -70,7 +70,7 @@
           }
         );
 
-        pyconsole.stdout_callback = (s) => term.error(s, {newline : false});
+        pyconsole.stdout_callback = (s) => term.echo(s, {newline : false});
         pyconsole.stderr_callback = (s) => term.error(s, {newline : false});
       });
     </script>
--- a/src/templates/console.html
+++ b/src/templates/console.html
@@ -4,6 +4,7 @@
     <meta charset="UTF-8">
     <script src="https://cdn.jsdelivr.net/npm/jquery"></script>
     <script src="https://cdn.jsdelivr.net/npm/jquery.terminal/js/jquery.terminal.min.js"></script>
+    <script src="https://cdn.jsdelivr.net/npm/jquery.terminal/js/echo_newline.js"></script>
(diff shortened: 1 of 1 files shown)


src/templates/console.html
```diff
@@ -69,28 +70,8 @@
...
-          lines[0] = this._prompt_tail + lines[0];
-          this._prompt_tail = lines.pop();
-          this.set_prompt(this._prompt_tail + prompt);
-          if( lines.length )
-            this.echo(lines.join('\n'));
-        }).bind(term);
-
-        pyconsole.stdout_callback = term.echo_no_newline;
-        pyconsole.stderr_callback = (s) => term.error(s.replace(/\n+$/, ""));
+        pyconsole.stdout_callback = (s) => term.error(s, {newline : false});
```
rth, line 73: Is it OK to use the error method for the stdout though?

hoodmane: Oops, typo.

rth (approved):
Great!

casatir:
Thanks @hoodman, I was not able to make this `echo_newline.js` work. Probably because of the JQuery version.

Unfortunately, errors are not always displayed in red:

``` python
Python 3.8.2 (default, Feb 24 2021 21:10:02) on WebAssembly VM
>>> 1/0
ZeroDivisionError: division by zero   <<< displayed white
>>> 1+1  <<< hitting enter here makes previous error displayed in red!
2
```

Upstream bug?

hoodmane:
> because of the JQuery version.

Yes, there was definitely a version problem.

> Upstream bug?

Yes, jterm-newline uses a hack where it puts the temporary stuff into the prompt, but jterm prompt has a bug where it doesn't seem to display `terminal-error` formatted prompts as red. We could do `term.echo(`[[;red;]${"abc"}]`, {newline : false});` instead.

PR #1281, opened 2021-02-24
Set sys.last_type, sys.last_value, sys.last_traceback on exception

hoodmane:
Normally repls store the most recent unhandled exception in `sys.last_value`, etc. This PR updates pyodide to do this too.


rth:
Thanks! Could you add a test?

PR #1287, opened 2021-02-25
BLD Fix overwriting package optflags via clfags etc in meta.yaml

rth:
Previously if a Python package was compiled with some optflags, for instance,
```
  gcc -O3 test.c -o test.o
```
and some cflags were provided in `meta.yaml` with a different optimization level such as `cflags='-Oz'`, then they would be injected by `handle_command` at the beginning of the command as,
```
  emcc -Oz -O3 test.c -o test.o
```
Then because when multiple optimization options are passed, clang [would take the last one](https://stackoverflow.com/a/61831017/1791279) thus ignoring the optimization option from cflags.

This fixes this issue by only taking the optflag from cflags/cxxflags/ldflags when they are conflicting with earlier optflags present in the initial command.

A simpler way of solving this could have been to append clfags, etc instead of prepending them. However I feel this would result in harder to read logs with potentially conflicting information and it doesn't take that much work to fix it, as done in this PR

### DIFF
--- a/pyodide_build/pywasmcross.py
+++ b/pyodide_build/pywasmcross.py
@@ -264,12 +264,26 @@ def handle_command(line, args, dryrun=False):
     elif new_args[0] == "em++":
         new_args.extend(args.cflags.split() + args.cxxflags.split())
 
+    optflags_valid = [f"-O{tok}" for tok in "01234sz"]
+    optflag = None
+    # Identify the optflag (e.g. -O3) in cflags/cxxflags/ldflags. Last one has
+    # priority.
+    for arg in new_args[::-1]:
+        if arg in optflags_valid:
+            optflag = arg
+            break
+
     lapack_dir = None
 
     used_libs = set()
 
     # Go through and adjust arguments
     for arg in line[1:]:
+        if arg in optflags_valid and optflag is not None and arg != optflag:
+            # There are multiple contradictory optflags provided, use the one
+            # from cflags/cxxflags/ldflags
+            continue
+
         if arg.startswith("-I"):
             if (
                 str(Path(arg[2:]).resolve()).startswith(sys.prefix + "/include/python")
--- a/pyodide_build/tests/test_pywasmcross.py
+++ b/pyodide_build/tests/test_pywasmcross.py
@@ -14,7 +14,7 @@ from pyodide_build.pywasmcross import make_parser
 
 @dataclass
 class BuildArgs:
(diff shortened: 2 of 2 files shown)


rth:
@hoodmane Could you have a look at this PR?

hoodmane:
Presumably the line `gcc -O3 test.c -o test.o` comes from the `Makefile` and the `-Oz` added in `meta.yaml` might be for debugging purposes? It sounds useful to have the `-Oz` flag actually take effect, though if just for debugging one could also just modify the `Makefile`. The logic you described sounds reasonable and like it's an improvement. The implementation looks good.

My understanding of the build system is still extremely limited though, so my opinions about this should probably be taken with a grain of salt.

rth:
Thanks for the review!

> Presumably the line gcc -O3 test.c -o test.o comes from the Makefile and the -Oz added in meta.yaml might be for debugging purposes?

Not exactly. The issue is that you compile an arbitrary Python C extension on x86_64, in my experience, it will be compiled with `-O3`. I'm not exactly sure where it's coming from, we specify [O2 in the Makefile](https://github.com/iodide-project/pyodide/blob/a1241301dcd28d01c855de0d109f8a55ee7b4ea0/Makefile#L15), maybe something to do with `python3-config --cflags` . 

And because we take those exact commands and change the compiler to build with emcc, the final modules are going to be built with `-O3` even if we didn't ask for it. That's might not be ideal because `-O3` results in noticeably [larger output sizes](https://discuss.python.org/t/why-does-cpython-use-o3/1915/13) with [relatively small performance improvement](https://discuss.python.org/t/why-does-cpython-use-o3/1915/9). So in our case, it might well be that `Os` (i.e. `O2` + size optimization) or `Oz` ( `O2` + more size optimizations) might be preferable. Of course we could set those globally for all packages, but it's also useful to be able to set them per package (including for debugging).

hoodmane:
Interesting! From the gcc docs:

> -O2 is the recommended level of optimization unless the system has special needs.
> ...
> Compiling with -O3 is not a guaranteed way to improve performance, and in fact, in many cases, can slow down a system due to larger binaries and increased memory usage. -O3 is also known to break several packages. Using -O3 is not recommended.

hoodmane:
I guess the benchmarks in the thread you linked show a ~5% speedup (eyeballing an average) for `-O3` over `-O2` and the binary is 20% larger with `-O3` than with `-O2`.

PR #1288, opened 2021-02-26
Update to scipy 1.0.1

rth:
This aims to update to scipy 1.0.1 (released March 2018). 
Toward https://github.com/iodide-project/pyodide/issues/549

**Edit:** No nevermind, I though I had migrated some of the patches but it doesn't seem to be the case. Probably better to simplify patches before trying to migrate them. I'll come back to this later.

PR #1291, opened 2021-02-26
MAINT Simplify scipy patches in make-int-return-values.patch

rth:
Replace some of the patches in `make-int-return-values.patch` with regex using `sed`.

The context for changes can be harder to understand with regex than with patches, but this replaces 720 lines of patches with 7 lines of `sed` so it's probably still worth it. It's also fairly straightforward to recover the initial patches by running these sed commands on the scipy repo.

Helps for #1288


rth:
Note that those sed commands won't work on Mac OS unless GNU sed is installed, but it was already the case before this PR.

PR #1292, opened 2021-02-27
match host and target in case of multiple pythons

pmp-p:
so cpython make install step can proceed in case of multiple python3.x in path


rth (approved):
Thanks @pmp-p !

PR #1293, opened 2021-03-01
WIP Update to scipy 1.6.1

rth:
A WIP attempt to update scipy from 0.17.1 to  1.6.1 (latest).

The main difficulty is that a number of BLAS/LAPACK related signature fixes done by @joemarshall in https://github.com/iodide-project/pyodide/pull/1198 need to be adapted/extended, because the way scipy wraps BLAS/LAPACK changed somewhat. For instance `scipy/linalg/_cython_wrapper_generators.py` no longer exists.

There is still work necessary here, there are for instance still some function signatures mismatches. The upside is that at least we can contribute some of the fixes upstream, once it's working.

I'll add back comments on patches once patches are finalized.

Closes https://github.com/iodide-project/pyodide/issues/549


hoodmane:
> I'll add back comments on patches once patches are finalized.

I noticed migrating patch comments is pretty annoying.

dalcde:
And so is migrating uncommented patches :)

hoodmane:
@rth What's the status on this? This seems to be necessary to be able to support statsmodels in Python 3.9.

rth:
Well the status is that I think this will require a large amount of work because this is scipy version newer by 6 years, there have been lots of changes since 0.17.1 and we also have lots of patches https://github.com/pyodide/pyodide/pull/1193 some of which might also need applyed to new parts of the code base etc. Currently this fails with some function signature errors. I'll try to get back to it later this summer.

What is the blocker for statsmodels exactly? You need to update to newer statsmodels version for Python 3.9 which requires newer scipy?

hoodmane:
I started looking into automating some of the more complicated patches here into llvm passes:
https://github.com/hoodmane/pyodide-llvm#voidtointreturn

hoodmane:
Okay so further research shows:
1. My `VoidToIntReturn` llvm pass _does_ fix the `void` vs `int` return value linker errors.
2. It is possible to use the custom llvm passes plugins with emscripten, currently at a cost of increasing the docker image size by ~400mb. Of this, 96MB is libLLVM and another 98MB is libclang + libclangc++. So the minimum possible cost of adding our own plugins is around 200MB for libLLVM + libclang. There's another ~38MB of llvm related stuff which might not be necessary and ~170MB of ??? dependencies that get loaded as part of installing llvm. Probably there are a few large things in there we could delete, I bet best case scenario is a cost of 300mb into the docker image.
3. Adding an empty llvm plugin to clang causes segfaults on a small percentage of input files: https://github.com/llvm/llvm-project/issues/52788 conveniently the segfaults aren't on any of the files that need patching. I just catch them and retry without the plugin and it works fine.

I think we would be better off patching the linker though. In particular, we could add a linker mode where instead of emitting a linker error when there's a function signature mismatch, it instead creates an adapter function and links that instead. I think this is the ideal solution. We could run it once and collect the list of functions that need an adapters, then on subsequent runs load that list and crash if it needs to shim a function that is not in the list.

hoodmane:
Oh I guess since `wasm-ld` is part of LLVM, patching it would actually force us to build at least part of llvm. Probably best to avoid that.

hoodmane:
I opened a new PR #2065 where I have the build working but it fails at link time with a missing export.

rth:
Superseded by https://github.com/pyodide/pyodide/pull/2065

PR #1294, opened 2021-03-01
ENH Support PyErr_CheckSignals interrupts

hoodmane:
Normally native C Python extensions use `PyErr_CheckSignals` to support keyboard interrupts. My previous patch for the interrupt system didn't patch `PyErr_CheckSignals` so it wouldn't work with native C extensions (they won't notice the keyboard interrupt). This patches `PyErr_CheckSignals` too. This also adds a function `pyodide.checkInterrupt()` that allows long running Javascript functions to support keyboard interrupts. For instance, a blocking implementation of `time.sleep` in Javascript looks like this:

```js
let blockingSleepBuffer = new Int32Array(new SharedArrayBuffer(4));
function sleep(t){
    for(let i = 0; i < t * 20; i++){
        // last argument to Atomics.wait is a timeout in milliseconds.
        // We want to wait a total of 1000 * t
        Atomics.wait(blockingSleepBuffer, 0, 0, 50);
        // handle keyboard interrupts at a time resolution of 50 ms.
        pyodide.checkInterrupt();
    }
}
```

I've tested this locally and it works. I'll try to figure out how to add a test for it. I think it would be nice to make `time.sleep` do this whenever `SharedArrayBuffer` is available and we're on a webworker (when [[CanBlock]] is true).

One concern is what the performance impact of this is. Natively `PyErr_CheckSignals` is guaranteed to report a keyboard interrupt the first time it is called after the signal is received by the python process. This implementation does that too, but it might be too expensive. `PyErr_CheckSignals` is a very hot function. If this is too slow, I can update it to only actually check for the keyboard interrupt every nth call for an appropriate value of n.


hoodmane:
Based on the benchmarks, I think this has an unmeasurable performance cost (I have labeled "with ib" to mean the tests running with the interrupt buffer checks).

rth:
@joemarshall maybe you would be interested in reviewing this PR, given other related topics you have worked on?

rth:
Is this still something we want @hoodmane for 0.18? If so please add a changelog entry and the file `src/core/pyproxy2.js` should likely be removed.

hoodmane:
Okay, I will try to bring this up to date soon. It's not a must-have feature for 0.18 since I don't think hardly anyone is using keyboard interrupts yet anyways. If we have time, it would be good to get it included since this has been sitting around for quite a while.

hoodmane:
This shows a fairly noticeable ~25% drop in performance in the pystone benchmark when the interrupt system is on. None of the other benchmarks show this performance cost, and strangely it doesn't seem to matter how much I space out the checks (when I checked it once every ~1000 python opcodes, once ~100 or once every ~50, the benchmark shows similar numbers). I wonder why Pystone is the only one affected.

joemarshall:
Looking at pystone, I suspect that it it could be simply the effect of an extra function call per opcode or line or whatever frequency that the pyerr_checksignals call is made. pystone uses a lot of lines to do not very much. You could verify this just by putting a do-nothing function in as the pyodide callback?

Given pyodide is single-threaded, there is a truly disgusting hack that could be done which is to check a flag in the sharedarraybuffer directly in the check signals so no function call would be needed, but I don't know whether the performance improvement would be worth it...

hoodmane:
> there is a truly disgusting hack that could be done which is to check a flag in the sharedarraybuffer directly in the check signals so no function call would be needed

Could you share this? I am always interested to learn new disgusting hacks.

PR #1299, opened 2021-03-03
[docs] rename files and fix broken links in quickstart 

mireille-raad:
I was reading the quickstart on https://pyodide.readthedocs.io/en/latest/usage/quickstart.html 

It says:
> See the following section on [serving pyodide files](https://pyodide.readthedocs.io/en/latest/usage/quickstart.html#serving-pyodide-files) for more details.

That link goes nowhere. So I looked around the document and it seems that [Serving pyodide packages](https://pyodide.readthedocs.io/en/latest/usage/serving-pyodide-packages.html) is actually about "Serving pyodide files"

This PR:
- renames the file from `serving-pyodide-packages.md` to `serving-pyodide-files.md` and updates the title/reference.
- fixes the broken link. 
- updates the index.

I figured that this is the more "honest" way to fix the docs, but if renaming the files is a hassle or might cause some breaking changes for people with existing PRs... then you can reject this and I can just fix the broken link without file changes.

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -32,7 +32,7 @@ may be installed from PyPI to be used with Pyodide.
 
    usage/quickstart.md
    usage/webworker.md
-   usage/serving-pyodide-packages.md
+   usage/serving-pyodide-files.md
    usage/loading-packages.md
    usage/type-conversions.md
    usage/api-reference.md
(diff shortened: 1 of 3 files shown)


rth:
>  So I looked around the document and it seems that Serving pyodide packages is actually about "Serving pyodide files"

Thanks for the PR @mireille-raad ! I'm not sure the term "file" would be better there. For instance `numpy.data` is not just file it's a container of numpy package files that are to be unpacked in the WebAssembly virtual file system. Same for pyodide data file which includes the CPython stdlib. So "package" might be more accurate, and just fixing the link would be enough.

Don't hesitate to open other PRs if you see things to improve though.

PR #1303, opened 2021-03-04
DOC fixes to quickstart >> "Accessing Python scope from Javascript"

mireille-raad:
This PR fixes issues in the section **Accessing Python scope from Javascript** in the quickstart on [this link](https://pyodide.org/en/latest/usage/quickstart.html).

`pyodide.globals.get('x')` and `pyodide.globals.set()` do not seem to work. I figured that `pyodide.pyimport()` does... and I looked around the issues and saw that the correct syntax is `pyodide.globals.x`  (see screenshot at end).

So, this PR mentions both options, there seems to be no consensus on which is better ( see issue https://github.com/iodide-project/pyodide/issues/926 )

I also fixed things with the examples that uses numpy, so it all runs. For example, `let x` was creating a re-assignment issue, numpy needs to be imported and so on (also see screenshots)

## screenshot / before
Screen Shot 2021-03-04 at 10 20 20 AM

Screen Shot 2021-03-04 at 10 43 51 AM

### DIFF
--- a/docs/usage/quickstart.md
+++ b/docs/usage/quickstart.md
@@ -124,33 +124,39 @@ Create and save a test `index.html` page with the following contents:
 
 ## Accessing Python scope from Javascript
 
-You can also access from Javascript all functions and variables defined in Python using the {any}`pyodide.globals` object.
+You can access from Javascript all functions and variables defined in Python by using the {any}`pyodide.globals` object or {any}`pyodide.pyimport` .
 
-For example, if you initialize the variable `x = numpy.ones([3,3])` in Python, you can access it from Javascript in your browser's developer console as follows: `pyodide.globals.get("x")`. The same goes for functions and imports. See {ref}`type_conversions` for more details.
+For example, if you initialize the variable `x = numpy.ones([3,3])` in Python, you can access it from Javascript in your browser's developer console by using either `pyodide.globals.x` or `pyodide.pyimport('x')`. The same goes for functions and imports. See {ref}`type_conversions` for more details.
 
 You can try it yourself in the browser console:
 ```js
+pyodide.runPython(`import numpy`);
 pyodide.runPython(`x=numpy.ones([3, 3])`);
-pyodide.globals.get("x");
+pyodide.globals.x;
+pyodide.pyimport('x');
 // >>> [Float64Array(3), Float64Array(3), Float64Array(3)]
 
-// create the same 3x3 ndarray from js
-let x = pyodide.globals.get("numpy").ones(new Int32Array([3, 3]));
-// x >>> [Float64Array(3), Float64Array(3), Float64Array(3)]
+// update the ndarray from js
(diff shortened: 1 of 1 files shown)


hoodmane:
Thanks for your docs mireille! 

It seems to me that you don't have a current build of pyodide. Can you do your tests on the dev console here:
https://pyodide-cdn2.iodide.io/dev/full/console.html
That link always goes to a recent build of the master branch, and so for the most part is appropriate for establishing the "ground truth" of the behavior (in addition to the test suite).

We are switching from using `pyodide.globals.x` to `pyodide.globals.get("x")`. The reason for this is that we want to make `globals` into a normal proxy of a Python dictionary that is accessed in the same way. The plan is that `proxy.x` should do `getattr(obj, "x")` and `proxy.get("x")` should do `obj["x"]`.  So in the current dev build if you do:

```js
>>> let d = pyodide.runPython("{ 'x' : 7}");
... d.get("x");
7
```
so if `globals` is the globals dictionary, we should say the same:
```js
>>> pyodide.runPython("x=2");
... pyodide.globals.get("x");
2
```
Some other dictionary can be used as a second argument to `runPython` to run in a different global scope:
```js
>>> pyodide.runPython("y = 2*x", d);
... d.get('y')
14
```

Here's what I see when I do these things on the most recent build https://pyodide-cdn2.iodide.io/dev/full/console.html:

image

mireille-raad:
ahh 🤦‍♀️  ... well thank you for the pointer :) and for taking the time to explain. I will make sure to double check things there and fix my build.

hoodmane:
The rest of the edits here look like improvements if you revert the `globals.x` changes.

rth:
> The rest of the edits here look like improvements if you revert the globals.x changes.

+1

Also for the next release we should probably make https://pyodide.org/en/stable the default instead of https://pyodide.org/en/latest to avoid the confusion where the docs do not correspond to the last release.

mireille-raad:
I will do the edits for numpy and figure out why I was having a different version of pyodide. I copied the example that was using the cdn link, not a local build.

rth:
In the CND URL you can replace the `v0.16.0` by `dev` to get the version from master.

hoodmane:
Good idea. Ideally we should set it up to suggest the `dev` in https://pyodide.org/en/lastest and the stable url in https://pyodide.org/en/stable/. Maybe it'd be fine to change it to a hard coded `dev` url in the docs for now? In a separate PR maybe we could set up some flag for whether we are building release docs or a dev copy of the docs and determine the url based on that? 
@rth you have any thoughts?

hoodmane:
Maybe we could use a conf-tag:
https://www.sphinx-doc.org/en/master/usage/configuration.html#conf-tags
with the `only` directive:
https://www.sphinx-doc.org/en/master/usage/restructuredtext/directives.html#including-content-based-on-tags

PR #1305, opened 2021-03-05
Actually use pyodide.globals in the console

hoodmane:
`globals()` refers to the temporary namespace that we execute the setup code in. It needs to be `__main__.__dict__` or `js.pyodide.globals`.

### DIFF
--- a/src/templates/console.html
+++ b/src/templates/console.html
@@ -23,6 +23,7 @@
             import sys
             import js
             from pyodide import console
+            import __main__
 
 
             def displayhook(value):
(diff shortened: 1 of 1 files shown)


rth (approved):
Thanks!

We really should find an easier way to make https://github.com/iodide-project/pyodide/issues/648 work so we could test PRs in console directly without having to build or do some other file manipulations locally.

hoodmane:
Yeah it's really hard to update `console.html` without tests. Even though it's only 60 lines of js, seems like practically every change I make introduces a regression.

Issue #382, opened 2019-04-17
importing the same python object several times should return the same js object


lovasoa:
Currently, the following notebook does not work as intented

```python
%% md
# Click me!
%% py
from js import document

def f(*args):
    document.querySelector("h1").innerHTML += "(>.<)"

document.body.addEventListener('click', f)
document.body.removeEventListener('click', f)

%%js
pyodide.pyimport("f") == pyodide.pyimport("f")
```

pyodide creates a new js wrapper around the same python function each time it needs one, which makes using some web APIs such as `removeEventListener` impossible.

mdboom:
Thanks for finding this.  Indeed, you make a good point, and it would be good to fix this if possible.

PR #1306, opened 2021-03-05
Use FinalizationRegistry for garbage collection when available

### LINKED ISSUE #1006: Use new Javascript GC apis to leak less memory

hoodmane:
This addresses #1049 and #1006. Hopefully it also addresses #693, maybe we should add a test for that.
If the `FinalizationRegistry` API is missing (Safari sucks), it gracefully falls back to leaking objects like crazy. 

Limitations:
1. The browser may not run the finalizers in a timely manner at all. In particular, the WASM heap is one giant allocation for the browser GC and browser GC does not care whether the WASM heap is experiencing memory pressure. To the browser GC, each `PyProxy` is tiny. It does seem that if you make enough small javascript objects it is possible to see the finalizer run without explicitly requesting garbage collection. In tests it seems like making around 1000 empty objects is enough to make the garbage collector run. So probably in long running programs it will eventually happen.
2. Still impossible to collect reference cycles between Python and Javascript. The Python GC API is good enough to allow it, but the Javascript GC API is not. We would need a hook that is run whenever the garbage collector visits an object.

The test only works in chrome. In Firefox apparently the way to trigger the GC in Selenium is 
1. Open a new tab and navigate to "about:memory"
2. Trigger a click to the GC button there.
I've done the test manually in firefox in this way, but I didn't find any actual selenium code examples where someone had done this and preliminary messing around seemed to suggest it would be a hassle. If people strongly prefer I can try to get this to work again...

Closes https://github.com/pyodide/pyodide/issues/1049
Closes #1006


hoodmane:
#382 and #1049 are nearly contradictory requests...

hoodmane:
In particular there are a few options I see:

1. Current state, #382 works as expected and #1049 doesn't work
2. Break #382 and fix #1049 
3. Remove the destroy API and fix #1049 by making it impossible to do
4. Make variants of `addEventListener` / `removeEventListener` with an extra `Map<int, WeakRef<PyProxy>>` that work as expected by adding special handling to `JsProxy`, thereby fixing the particular example in #382, but leaving any other add/remove API that compares on object identity broken. People using `add/removeEventListener` directly from Javascript need to store the particular `PyProxy` they used it with.

hoodmane:
I'm leaning toward implementing 2 above with an option of adding 4 in the future.
@phorward What do you think?

rth:
Could you resolve merge conflicts, or are you waiting for https://github.com/pyodide/pyodide/pull/1322 to get in?

hoodmane:
Just hadn't noticed them, will do.

hoodmane:
Well I hope I got that merge right, the interaction between this and #1446 is a bit tricky.

src/core/pyproxy.js
```diff
@@ -390,28 +437,19 @@ JS_FILE(pyproxy_init_js, () => {0,0; /* Magic, see include_js_file.h */
      *
      * @returns {Iterator} An iterator for the proxied Python object.
      */
-    [Symbol.iterator] : function*() {
+    [Symbol.iterator] : function() {
+      let ptr = _getPtr(this);
```
hoodmane, line 441:
```suggestion
      let ptrobj = _getPtr(this);
```

src/core/pyproxy.js
```diff
@@ -390,28 +437,19 @@ JS_FILE(pyproxy_init_js, () => {0,0; /* Magic, see include_js_file.h */
...
-        if (_PyErr_Occurred()) {
-          throw new TempError();
-        }
-      } catch (e) {
-        if (e instanceof TempError) {
-          _pythonexc2js();
-        } else {
-          Module.fatal_error(e);
-        }
+        let iterptr = _PyObject_GetIter();
```
hoodmane, line 444:
```suggestion
        let iterptr = _PyObject_GetIter(ptrobj);
```

docs/usage/faq.md
```diff
@@ -167,3 +167,31 @@ Unpickling data is similar to `eval`. On any public-facing server it is a really
...
+Note that the most straight forward way of doing this will not work:
+```py
+from js import document
+def f(*args):
+    document.querySelector("h1").innerHTML += "(>.<)"
+
+# Leaks f and doesn't remove the event listener
+document.body.addEventListener('click', f)
+document.body.removeEventListener('click', f)
+```
```
rth, line 182: Maybe expand what you mean "won't work" in text instead of the comment,

"""
since this will leak f and ...
"""

docs/usage/faq.md
```diff
@@ -167,3 +167,31 @@ Unpickling data is similar to `eval`. On any public-facing server it is a really
...
+
+To do this correctly use the following:
```
rth, line 184: Nit, maybe

```suggestion
To do this correctly use :func:`pyodide.create_proxy` as follows:
```
Or some other way to link to https://pyodide.org/en/latest/usage/api/python-api.html#pyodide.create_proxy as it's not that obvious what that function does.

rth (approved):
Thanks for this PR @hoodmane. Generally LGTM.

> I'm leaning toward implementing 2 above with an option of adding 4 in the future.

Fine with me.

hoodmane:
>> I'm leaning toward implementing 2 above with an option of adding 4 in the future.
> Fine with me.

That is out of date, 2 + `create_proxy` resolves the issue as described in the faq.

hoodmane:
This also requires some updates to the docs.

hoodmane:
Well something is screwed up here and has been for the last few commits. I think I may have messed up the merge in f721f0b.

PR #1311, opened 2021-03-09
Print fatal error message when the fatal error occurs

hoodmane:
Currently when there's a fatal error, it updates the various pyodide apis to throw an error saying that pyodide has fatally failed, and then it reraises the original error. However, it's a bit confusing because it first says there's been a fatal error the _next_ time a pyodide API is used. This adds a `console.error` to print the fatal error message before rethrowing the original error which should make it easier to locate where the problem actually occurred.


rth (approved):
Thanks @hoodmane !

Issue #301, opened 2019-01-21
Moving large objects from JavaScript to python is super linearly slow


jstafford:
I'm playing around with pyodide to see if it would be suitable for running user defined scripts (written in python) against datasets that are in memory in JavaScript. 

What I'm finding is that the ```from js import data``` construct becomes incredibly slow when ```data``` is an array of strings with 10000 or more entires. Looking at js2python.c, jsproxy.c, and hiwire.c I don't see anything obvious that would cause this to be the case, but I'm not 100% sure I'm looking at the right code, and I don't know how to debug it if I am in the right spot.

Is this a known problem, and if so, is it something that is likely to be fixed anytime soon?

My simplistic test code for this is as follows: 
```
  window.languagePluginLoader.then(() => {
    // pyodide is now ready to use...
    console.log(window.pyodide.runPython('import sys\nsys.version'));
    window.pyodide.loadPackage('numpy').then(() => {
      console.log('numpy loaded')
      const len = 30000
      window.colA = new Array( len ).fill('2.1')
      window.colB = new Array( len ).fill('3.2')
      const start = performance.now()
      console.log(window.pyodide.runPython(
        'import numpy\n' +
        'from decimal import Decimal\n' +
        'from js import colA\n' +
        'from js import colB\n' +
        'decimal_a = [Decimal(x) for x in colA]\n' +
        'decimal_b = [Decimal(x) for x in colB]\n' +
        'vec_a = numpy.array(decimal_a)\n' +
        'vec_b = numpy.array(decimal_b)\n' +
        'prod_ab = vec_a * vec_b\n' +
        'str_prod_ab = [str(x) for x in prod_ab]\n' +
        'str_prod_ab'
      ));
      const stop = performance.now()
      console.log('duration', stop - start)

    })
  });
```

On my machine this takes about 90 seconds to run. When I move the initialization of colA and colB from JavaScript into the python code it goes down to about 2 seconds. To see this replace the ```from js import``` lines with the following lines:
```
        `colA = ['2.1' for i in range(${len})]\n` +
        `colB = ['3.2' for i in range(${len})]\n` +
```

I've also tested it with replacing the numpy usage with a simpler: 
```
        'prod_ab = [a*b for a,b in zip(decimal_a,decimal_b)]\n' +
```
but it runs about the same.

I'm using strings and Decimal because preserving significant digits is important in my use case.

jstafford:
[...]

PR #1315, opened 2021-03-10
Some tweaks to internal error handling

hoodmane:
This improves the error messages generated when `hiwire.get_value` is called with a `NULL` argument (if error indicator set, explicitly tie the Python error that presumably caused the problem to the warning, otherwise explicitly state that the error indicator was not set).

I implemented the hiwire reference count check suggested in #301, behind the DEBUG_F flag. (While manually testing this feature, I think I discovered a memory leak...)

I moved the error handling code in `python2js` to `error_handling` since it seems like it more logically belongs there.

### DIFF
--- a/src/core/error_handling.c
+++ b/src/core/error_handling.c
@@ -21,7 +21,7 @@ EM_JS_NUM(errcode, log_error, (char* msg), {
 
 // Right now this is dead code (probably), please don't remove it.
 // Intended for debugging purposes.
-EM_JS_NUM(errcode, log_error_obj, (JsRef obj), {
+EM_JS_NUM(errcode, console_error_obj, (JsRef obj), {
   console.error(Module.hiwire.get_value(obj));
 });
 
@@ -114,14 +114,7 @@ wrap_exception(bool attach_python_error)
   success = true;
 finally:
   // Log an appropriate warning.
-  if (success) {
-    EM_ASM(
-      {
-        let msg = Module.hiwire.get_value($0).message;
-        console.warn("Python exception:\n" + msg + "\n");
-      },
-      jserror);
-  } else {
+  if (!success) {
     PySys_WriteStderr("Error occurred while formatting traceback:\n");
     PyErr_Print();
     if (type != NULL) {
@@ -146,11 +139,19 @@ finally:
   return jserror;
 }
 
+EM_JS_NUM(errcode, log_python_error, (JsRef jserror), {
+  let msg = Module.hiwire.get_value(jserror).message;
+  console.warn("Python exception:\n" + msg + "\n");
+  return 0;
+});
+
 void _Py_NO_RETURN
 pythonexc2js()
 {
   JsRef jserror = wrap_exception(false);
-  if (jserror == NULL) {
+  if (jserror != NULL) {
+    log_python_error(jserror);
+  } else {
     jserror =
       new_error("Error occurred while formatting traceback", Js_undefined);
   }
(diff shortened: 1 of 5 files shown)


rth:
> I moved the error handling code in python2js to error_handling since it seems like it more logically belongs there.

Could you do that in a separate PR to be merged before this one? As otherwise it's hard to identify what's a code copy and what's a modification when reviewing. Thanks!

> While manually testing this feature, I think I discovered a memory leak...

Would periodically building/testing core with sanitizers https://emscripten.org/docs/debugging/Sanitizers.html help?

hoodmane:
> do that in a separate PR 

Sure.

> Would testing core with sanitizers help?

It couldn't hurt, but I think that's a bit orthogonal to the problem here: the issue is that Javascript objects are getting added to the hiwire map that are never removed. I think from the point of view of emscripten, these leaked Javascript objects don't exist at all. It should be pretty simple for us to write test code to try to detect this, but currently we don't have much in the way of assertions that we don't leak like crazy.

On the other hand, I think the sanitizers would likely notice that we leak a lot of PyProxies too...

hoodmane:
Okay I did the move in #1320 and rebased this PR on top of it. The second commit here has all the interesting changes, the first commit is #1320.

src/core/hiwire.c
```diff
@@ -60,6 +65,14 @@ EM_JS_NUM(int, hiwire_init, (), {
     let idval = _hiwire.counter[0];
     _hiwire.objects.set(idval, jsval);
     _hiwire.counter[0] += 2;
+#ifdef DEBUG_F
```
rth, line 68: Could you add a section about building core with debug flags in https://pyodide.org/en/latest/development/building-from-sources.html as currently it's not really documented..

src/core/error_handling.c
```diff
@@ -146,11 +139,19 @@ wrap_exception(bool attach_python_error)
   return jserror;
 }
 
+EM_JS_NUM(errcode, log_python_error, (JsRef jserror), {
+  let msg = Module.hiwire.get_value(jserror).message;
+  console.warn("Python exception:\n" + msg + "\n");
```
rth, line 144: In the future we should find a way to support custom error callbacks as we do in pyodide.js https://github.com/iodide-project/pyodide/blob/a31f0c4e39599ead2523c72b6ed19d20e68780ca/src/pyodide.js#L521

rth (approved):
One comment otherwise LGTM. Thanks!

src/core/error_handling.c
hoodmane: What about using `sys.excepthook`?

src/core/hiwire.c
hoodmane: I think I should also change `make debug` not to set `-s ASSERTIONS=2`, I find that it slows compilation and runtime down a lot and makes a lot of noise which isn't very useful. On the other hand, `DEBUG_F` usually provides really helpful diagnostics. I've been compiling with `EXTRA_CFLAGS="-D DEBUG_F" make minimal` to avoid the `ASSERTIONS`.

hoodmane:
Okay I added the info to `building_from_sources`. I only changed the text in the `Add some extra info...` commit, the other commits just redid the textwrap in `building_from_sources`.

src/core/error_handling.c
rth: You mean user is expected to monkeypatch it?  Could be as long as it's documented.

PR #1321, opened 2021-03-12
Filter integers out of dir(js_array)

hoodmane:
Javascript arrays list a bunch of integers among their ownKeys. I think that including these in `dir(js_array)` is generally undesirable: the point of `dir` is supposed to be for finding out which methods are available, if you want to know what keys an array contains it's more common to inspect the `length` field. This strips them out.

### DIFF
--- a/src/tests/test_jsproxy.py
+++ b/src/tests/test_jsproxy.py
@@ -41,15 +41,22 @@ def test_jsproxy_dir(selenium):
     selenium.run_js(
         """
         window.a = [0,1,2,3,4,5,6,7,8,9];
+        a[27] = 0;
         a[":"] = 0;
         a["/"] = 0;
+        a.abcd = 0;
+        a.α = 0;
+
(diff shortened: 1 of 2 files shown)


src/tests/test_jsproxy.py
```diff
@@ -38,6 +38,21 @@ def test_jsproxy_dir(selenium):
     assert set1.issuperset(jsproxy_items)
     assert set1.issuperset(callable_items)
     assert set1.isdisjoint(a_items)
+    selenium.run_js(
+        """
+        window.a = [0,1,2,3,4,5,6,7,8,9];
+        a[":"] = 0;
+        a["/"] = 0;
```
rth, line 45: Could you also add an example with a 2 char integer that is filtered and a multi character key that's not?

Maybe also a test with a non ascii key?

rth (commented):
LGTM otherwise, thanks!

PR #1324, opened 2021-03-13
micropip improvements: fetch instead of XMLHttpRequest, async functions, remove memory leaks

hoodmane:
Micropip used a synchronous `XMLHttpRequest` to fetch the json header data from PyPi. Synchronous http requests are deprecated on the main thread and it generates a deprecation warning. Furthermore, the `XMLHttpRequest` API itself is pretty lousy and the `fetch` API should be used instead. Using the `fetch` API allows much simpler code and also avoids leaking the callback that we need to attach to the `XMLHttpRequest`. Avoiding the synchronous requests means that we can now fetch all the header data for all the packages we are loading simultaneously, rather than waiting for one request to finish before the next starts.

Micropip was designed with a fair amount of callback hell before. Now that we have full support for asyncio, we can just `await` the `fetch` API directly, at the cost of requiring the `_get_url()` function to be marked as `async`. This also requires the entire tree of functions calling `_get_url` to be marked `async` and allows us to get rid of all of the `resolve` and `reject` and `callback` callback hell parameters. Also, rather than manually counting how many requests have been resolved, I switched to using `asyncio.gather`.

The one issue here is that `asyncio.gather` schedules coroutines on the event loop, but for testing purposes we'd rather avoid using an event loop at all. For this I made a mock version of `gather` that just runs the coroutines in order.

Finally, rather than returning a promise, I switched to returning a future. This has several benefits: shorter code, more idiomatic python, it works the same natively (so now with a trivial mock event loop one could test the main entry point) and it avoids leaking the promise executor.

Similar changes probably should be made to `console.py` which leaks a bunch of python closures in the same way.

**API Change:** Now returns a `Future` instead of a `Promise`. When using from Javascript, this isn't be noticeable since `Future`s get wrapped in a `Promise`. When using from Python, the API is different but in trade we avoid memory leaks.

### DIFF
--- a/docs/project/changelog.md
+++ b/docs/project/changelog.md
@@ -113,6 +113,7 @@ substitutions:
 
 - {{ Feature }} `micropip` now supports installing wheels from relative urls.
   [#872](https://github.com/iodide-project/pyodide/pull/872)
+- {{ Change }} `micropip.install` now returns a Python `Future` instead of a Javascript `Promise`.
 
 ### Build system
 
--- a/docs/project/changelog.md
+++ b/docs/project/changelog.md
@@ -113,7 +113,7 @@ substitutions:
 
 - {{ Feature }} `micropip` now supports installing wheels from relative urls.
   [#872](https://github.com/iodide-project/pyodide/pull/872)
-- {{ Change }} `micropip.install` now returns a Python `Future` instead of a Javascript `Promise`.
+- {{ API }} `micropip.install` now returns a Python `Future` instead of a Javascript `Promise`.
 
 ### Build system
 
--- a/packages/micropip/micropip/micropip.py
+++ b/packages/micropip/micropip/micropip.py
@@ -11,6 +11,7 @@ from distlib import markers, util, version
 
 import sys
 
+# Provide stubs for testing in native python
 IN_BROWSER = "js" in sys.modules
 
 if IN_BROWSER:
@@ -19,7 +20,6 @@ if IN_BROWSER:
 else:
     WHEEL_BASE = Path(".") / "wheels"
 
-# Implementations of HTTP fetching for in-browser and out-of-browser
 if IN_BROWSER:
     from js import fetch
 
--- a/packages/micropip/micropip/micropip.py
+++ b/packages/micropip/micropip/micropip.py
@@ -9,18 +9,37 @@ from typing import Dict, Any, Union, List, Tuple
 
 from distlib import markers, util, version
 
+import sys
+IN_BROWSER = "js" in sys.modules
 
-try:
-    from js import XMLHttpRequest
+if IN_BROWSER:
+    # In practice, this is the `site-packages` directory.
+    WHEEL_BASE = Path(__file__).parent
+else:
+    WHEEL_BASE = Path(".") / "wheels"
+
+# Implementations of HTTP fetching for in-browser and out-of-browser
+if IN_BROWSER:
+    from js import fetch
+    async def _get_url(url):
+        resp = await fetch(url)
+        if not resp.ok:
+            raise OSError(f"Request for {url} failed with status {resp.status}: {resp.statusText}")
+        return io.BytesIO(await resp.arrayBuffer())
+else:
+    from urllib.request import urlopen
+    async def _get_url(url):
+        with urlopen(url) as fd:
+            content = fd.read()
+        return io.BytesIO(content)
+
+if IN_BROWSER:
     from js import pyodide as js_pyodide
-except ImportError:
-    XMLHttpRequest = None
-
+else:
     class js_pyodide:  # type: ignore
         """A mock object to allow import of this package outside pyodide
         Report that all dependencies are empty.
         """
-
         class _module:
             class packages:
                 class dependencies:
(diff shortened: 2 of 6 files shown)


packages/micropip/test_micropip.py
```diff
@@ -61,24 +59,34 @@ def test_parse_wheel_url():
 
 def test_install_custom_url(selenium_standalone, web_server_tst_data):
     server_hostname, server_port, server_log = web_server_tst_data
-    selenium_standalone.load_package("micropip")
-    selenium_standalone.run("import micropip")
     base_url = f"http://{server_hostname}:{server_port}/"
     url = base_url + "snowballstemmer-2.0.0-py2.py3-none-any.whl"
-    selenium_standalone.run(f"micropip.install('{url}')")
-    # wait untill micropip is loaded
-    time.sleep(1)
```
rth, line 70: Great! We can get rid of all the `time.sleep` after using micropip.
There is another one in `src/tests/test_python.py`

packages/micropip/test_micropip.py
```diff
@@ -61,24 +59,34 @@ def test_parse_wheel_url():
...
-    selenium_standalone.run("import snowballstemmer")
+    selenium_standalone.run_js(
+        f"""
+        await pyodide.runPythonAsync(`
```
rth, line 66: We have `selenium_standalone.run_async` for this don't we?

packages/micropip/test_micropip.py
```diff
@@ -8,26 +8,24 @@
...
-
-    selenium_standalone.run("import snowballstemmer")
-    selenium_standalone.run("stemmer = snowballstemmer.stemmer('english')")
-    assert selenium_standalone.run(
-        "stemmer.stemWords('go going goes gone'.split())"
-    ) == ["go", "go", "goe", "gone"]
+    assert (
+        selenium_standalone.run_js(
+            """
+        let result = await pyodide.runPythonAsync(`
```
rth, line 14: Also selenium_standalone.run_async

packages/micropip/micropip/micropip.py
```diff
@@ -148,53 +138,47 @@ def __init__(self):
...
             self.installed_packages.update(dict((k, None) for k in pyodide_packages))
-            js_pyodide.loadPackage(list(pyodide_packages)).then(do_resolve)
+            wheel_promises.append(js_pyodide.loadPackage(list(pyodide_packages)))
 
         # Now install PyPI packages
         for name, wheel, ver in transaction["wheels"]:
-            _install_wheel(name, wheel, do_resolve, reject)
+            wheel_promises.append(_install_wheel(name, wheel))
             self.installed_packages[name] = ver
+        await gather(*wheel_promises)
```
rth, line 178: Could we double check that nothing can go wrong when installing wheels concurrently? 
As far as I understand, we have a guarantee that wheels names are already de-duplicated by design?

packages/micropip/test_micropip.py
```diff
@@ -61,24 +59,34 @@ def test_parse_wheel_url():
...
     pytest.importorskip("distlib")
     import micropip
 
     transaction = {"wheels": []}
-    micropip.PACKAGE_MANAGER.add_requirement(
+    coroutine = micropip.PACKAGE_MANAGER.add_requirement(
         "./snowballstemmer-2.0.0-py2.py3-none-any.whl", {}, transaction
     )
+    try:
+        coroutine.send(None)
```
rth, line 84: I can find very little documentation on this use case. Could you add a comment that it's a way to run that coroutine from a sync function?

rth (commented):
Nice, it's much more readable!

> The one issue here is that asyncio.gather schedules coroutines on the event loop, but for testing purposes we'd rather avoid using an event loop at all.

If feels to me that we are going to have some amount of async code that we might want to test outside of pyodide, so we might as well add [pytest-asyncio](https://pypi.org/project/pytest-asyncio/) to test dependencies and directly use an event loop.

> API Change: Now returns a Future instead of a Promise

Please add a changelog entry.

packages/micropip/test_micropip.py
hoodmane: Yeah I agree this is hardly described anywhere. It's kind of surprising to me that there isn't more explanation for this, because all of the `EventLoop` and asyncio in Python is abstractions on top of this: a coroutine is just a generator, whenever it wants to do io, etc it yields a request and the event loop schedules the request and when it gets the result `send`s it back to the coroutine, when it's done it raises `StopIteration`. If the coroutine does no async operations, then it just runs to completion on the first try. I can add a comment. Using `pytest-asyncio` would probably be easier to understand though.

packages/micropip/test_micropip.py
hoodmane: Yeah, but I prefer the tests to be styled so that they can be copied into the browser console when they fail. If you use `selenium_standalone.run_async` then when copying it you have to add an extra `pyodide.runPythonAsync()` around everything.

I've noticed that it's more common for errors to be generated in the `selenium.run` wrapper code than in the `selenium.run_js` wrapper code too (though gradually improving `run` is also probably a good thing), but again, such wrapper code errors are annoying because they are harder to reproduce interactively.

The bigger tradeoff is in consolidating calls to `selenium.run` or not consolidating them. If you consolidate the calls, then the selenium error message is less specific. I find that this is more than made up for when directly copying the test into console.html demonstrates the bug as well, because then I can easily see what step the error happened at and investigate the state just before and after.

packages/micropip/micropip/micropip.py
hoodmane: I'm pretty sure it was installing them concurrently before using callbacks, the part that was serial was fetching the json header data from pypi (and I did confirm that those calls commute with each other).

hoodmane:
> It feels to me that we are going to have some amount of async code that we might want to test outside of pyodide, so we might as well add pytest-asyncio to test dependencies and directly use an event loop.

I am in favor of this. Maybe you could do that in a separate PR and then we could switch to the simpler approach? (This PR and that one could be merged in either order IMO.)

> Update changelog

Done.

docs/project/changelog.md
```diff
@@ -113,6 +113,7 @@ substitutions:
 
 - {{ Feature }} `micropip` now supports installing wheels from relative urls.
   [#872](https://github.com/iodide-project/pyodide/pull/872)
+- {{ Change }} `micropip.install` now returns a Python `Future` instead of a Javascript `Promise`.
```
rth, line 116:
```suggestion
- {{ API }} `micropip.install` now returns a Python `Future` instead of a Javascript `Promise`.
```

rth:
>  Maybe you could do that in a separate PR and then we could switch to the simpler approach?

Sounds good. Thanks!

wojsam:
I'm not sure if it's worth mentioning, but while using micropip in pyodide.runPython I ran into "out-of-browser" mode. I switched to pyodide.runPythonAsync and it's fine.

PR #1328, opened 2021-03-13
Fix bug and memory leak in python2js_buffer

### LINKED ISSUE #1326: Potential memory leak in hiwire.c
I am profiling memory for my application, which sends an updated nested dictionary to JS every 100 ms. I call deepCopyToJavascript and destroy the proxy before handing it off. I am seeing continual memory growth, and during profiling, one of the culprits appears to be converting numpy uint8 arrays to JS arrays. The report has the array allocation here:

https://github.com/iodide-project/pyodide/blob/ab249a0a50e401783dda7a270b5de8d26d0c3f33/src/core/hiwire.c#L185

Looking through the hiwire code, I can't figure out what new_value is supposed to accomplish. From what I can tell, it places it in a JS dict and never deletes it. I hand off the converted dictionary to Vuex and never touch it again. It should be destroyed when the data is replaced 100 ms later, but it is still being retained.

hoodmane:
Resolves #1326 (at least the part of it that still exists on master). Thanks to @johnwason for the report.

### DIFF
--- a/src/core/python2js_buffer.c
+++ b/src/core/python2js_buffer.c
@@ -335,33 +335,34 @@ _python2js_buffer_to_typed_array(Py_buffer* buff)
     }
   }
 
+  Py_ssize_t len = buff->len / buff->itemsize;
   switch (format) {
     case 'c':
     case 'b':
-      return hiwire_int8array((i8*)buff->buf, buff->len);
+      return hiwire_int8array((i8*)buff->buf, len);
     case 'B':
-      return hiwire_uint8array((u8*)buff->buf, buff->len);
+      return hiwire_uint8array((u8*)buff->buf, len);
     case '?':
       return NULL;
     case 'h':
-      return hiwire_int16array((i16*)buff->buf, buff->len);
+      return hiwire_int16array((i16*)buff->buf, len);
     case 'H':
-      return hiwire_uint16array((u16*)buff->buf, buff->len);
+      return hiwire_uint16array((u16*)buff->buf, len);
     case 'i':
     case 'l':
     case 'n':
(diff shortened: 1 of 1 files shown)


src/core/python2js_buffer.c
```diff
@@ -335,33 +335,34 @@ _python2js_buffer_to_typed_array(Py_buffer* buff)
     }
   }
 
+  Py_ssize_t len = buff->len / buff->itemsize;
   switch (format) {
     case 'c':
     case 'b':
-      return hiwire_int8array((i8*)buff->buf, buff->len);
+      return hiwire_int8array((i8*)buff->buf, len);
```
rth, line 342: Shouldn't we have been able to detect this in test, via the array length and if so could we add  a regression test.

rth (commented):
Thanks!

Please add a changelog if applicable.

src/core/python2js_buffer.c
hoodmane: No, it's actually pretty weird, it first makes a too-big array in `_python2js_buffer_to_typed_array`:
https://github.com/iodide-project/pyodide/blob/69e28cbe0ef9e0292235575df1bc23a96cbb7815/src/core/python2js_buffer.c#L473
and then it slices the result out of it in `_python2js_shareable_buffer_recursive` 
https://github.com/iodide-project/pyodide/blob/69e28cbe0ef9e0292235575df1bc23a96cbb7815/src/core/python2js_buffer.c#L482
The result it makes has the right size. Then it leaks the too-big array. Only reason I noticed this internal miscalculation is that it leaked the object.
I think I should add a regression test for the memory leak though.

rth: OK, thanks for the clarification!

PR #1329, opened 2021-03-13
Use Python Futures and async instead of Promises in console.py

hoodmane:
This is similar to #1324. The current `console.py` leaks the `load_packages_and_run` and `run` closures by using them with the `promise.then` api. Instead, using a `Future` avoids the leak.


rth:
Maybe @casatir could have a look..

hoodmane:
I think a different approach might be better here, will experiment a bit.

hoodmane:
Closed by #1334.

PR #1330, opened 2021-03-13
Fix two memory leaks in jsproxy

hoodmane:


src/tests/test_jsproxy.py
```diff
@@ -736,3 +736,19 @@ def raises(exc, match=None):
         `);
         """
     )
+
+
+def test_memory_leaks(selenium):
```
rth, line 741: Maybe

```suggestion
def test_jsproxy_get_iter_memory_leak(selenium):
```
?

rth (approved):

src/tests/test_jsproxy.py
hoodmane: Here I tested both `iter` and `repr`. I think it could be reasonably practical to update the test to include a bunch of other commands (we can just call various APIs and then assert that nothing leaked at the end). Anyways, I figure the test name and contents can be updated as we increase leak test coverage. Maybe `test_jsproxy_memory_leaks` would be good (though this is already in a file called `test_jsproxy`).

rth:
Actually `test_memory_leaks` fails when run as part of the test suite, but passes when re-run later independently. See for instance CI in this PR.

hoodmane:
Right these sorts of tests are very sensitive to the environment.

We should make it so that it fails the CI if the test only passes on the second run? I often run tests locally only individually and then if the CI passes I don't look at the results. The rerun-failed-tests thing is a really helpful diagnostic tool but it shouldn't be a relaxation of the test suite.

hoodmane:
I opened #1350 for this.

PR #1334, opened 2021-03-13
create_proxy and create_once_proxy APIs (for reducing memory leaks)

### LINKED ISSUE #1325: Potential memory leak in WebLoop.py
WebLoop.py uses `js.setTimeout` to implement call_* functions. The code is:

```
h = asyncio.Handle(callback, args, self, context=context)
setTimeout(h._run, delay * 1000)
return h
```

https://github.com/iodide-project/pyodide/blob/de7c3420103fabbf01c589fcf2a429d157a6bd5e/src/pyodide-py/pyodide/webloop.py#L139

I think this may result in a memory leak, since the `h._run` JS proxy is never released.

hoodmane:
This PR is intended to help resolve memory leaks like the one in #1325, and the promise related leaks in `console.py` and `micropip` and in `jsproxy.c`. It also should resolve the problem I discussed in #1306 with the difficulty between #382 and #1049: to remove an event listener you should make a proxy explicitly so you can delete it later:

```py
def callback(event):
   # handle event
   
proxy = pyodide.create_proxy(callback)
from js import document
document.body.addEventListener("click", proxy)

# Later:
document.body.removeEventListener("click", proxy)
proxy.destroy()
```

Resolves #1325.

There was a very tough to locate fatal error due to undefined behavior in JsProxy call, I made a separate PR to fix that #1335.


src/core/pyproxy.c
```diff
@@ -1225,9 +1232,69 @@ EM_JS_NUM(int, pyproxy_init_js, (), {
 });
 // clang-format on
 
+JsRef
+create_once_proxy(PyObject* obj)
+{
+  Py_INCREF(obj);
+  return (JsRef)EM_ASM_INT(
+    {
+      let o = { $$ : { ptr : $0 } };
```
rth, line 1241: Remind me what's "$$" in this context?

src/core/pyproxy.c
```diff
@@ -1225,9 +1232,69 @@ EM_JS_NUM(int, pyproxy_init_js, (), {
 });
 // clang-format on
 
+JsRef
+create_once_proxy(PyObject* obj)
```
rth, line 1236: Could you add a docstring, as it's not clear from the name that the resulting proxy can only be called once.

src/pyodide-py/pyodide/__init__.py
```diff
@@ -26,4 +26,6 @@
     "JsException",
     "register_js_module",
     "unregister_js_module",
+    "create_once_proxy",
+    "create_proxy",
```
rth, line 30: Please also add these to https://github.com/iodide-project/pyodide/blob/master/docs/usage/api-reference.md#python-api and make sure that the generated signature / documentation looks OK?

src/core/pyproxy.c
```diff
@@ -1225,9 +1232,69 @@ EM_JS_NUM(int, pyproxy_init_js, (), {
...
+  return result;
+}
+
+static PyMethodDef pyproxy_methods[] = {
+  {
+    "create_once_proxy",
+    create_once_proxy_py,
+    METH_O,
+    PyDoc_STR("Create a wrapper around a Python function that can be called "
+              "once from Javascript"),
```
rth, line 1283: I think the first line of the docstring should be a signature line here and below?
https://www.python.org/dev/peps/pep-0007/#documentation-strings

rth (commented):
Thanks for working on reducing memory leaks! A few comments below otherwise looks good to me.

rth:
Also please add a changelog entry

src/core/pyproxy.c
hoodmane: This is a hack, I've fixed it in #1340, I suppose I will copy the code from that branch back to this one.

src/pyodide-py/pyodide/__init__.py
hoodmane: It's a bit annoying that we have to duplicate the doc strings in order for sphinx to find them.

rth: I don't think it's specific to sphinx. It's also necessary to make help(create_proxy) show the correct signature.

hoodmane:
I think we need a third common variant of this to use with `Promise.then`.

src/pyodide-py/pyodide/__init__.py
hoodmane: It isn't specific to Sphinx I agree. The issue I was referring to is the fact that our C code can't be built for native Python so any C extension docstrings currently need to be duplicated into `_core.py` so that they are present in the stubs.

src/core/hiwire.c
```diff
@@ -388,17 +388,37 @@ EM_JS_REF(JsRef, hiwire_dir, (JsRef idobj), {
...
   return Module.hiwire.new_value(jsfunc(... jsargs));
 });
 
-EM_JS_REF(JsRef, hiwire_call_OneArg, (JsRef idfunc, JsRef idarg), {
-  let jsfunc = Module.hiwire.get_value(idfunc);
-  let jsarg = Module.hiwire.get_value(idarg);
-  return Module.hiwire.new_value(jsfunc(jsarg));
-});
+JsRef
+hiwire_call_va(JsRef idobj, ...)
```
rth, line 413: What does "va" stand for in the name? Could we use something more explicit?

OK I see http://www.cplusplus.com/reference/cstdarg/
Maybe then at least have the term "Variable arguments" appear somewhere in the docstring.

src/core/jsproxy.c
```diff
@@ -644,26 +646,111 @@ JsProxy_Await(JsProxy* self, PyObject* _args)
...
   Py_CLEAR(set_exception);
-  Py_DECREF(fut);
+  hiwire_CLEAR(promise_id);
+  hiwire_CLEAR(promise_handles);
+  hiwire_CLEAR(promise_result);
+  return result;
+}
+
+/**
+ * Overload for `then` for JsProxies with a `then` method.
```
rth, line 669: What was the issue with the previous implementation? Memory leaks?

src/core/pyproxy.c
```diff
@@ -1220,14 +1230,136 @@ EM_JS_NUM(int, pyproxy_init_js, (), {
...
+      "of the Callable. The javascript function also has a `destroy` API that "
+      "can be used to release the proxy without calling it."),
+  },
+  {
+    "create_proxy",
+    create_proxy,
+    METH_O,
+    PyDoc_STR("create_proxy(obj : Any) -> JsProxy"
+              "\n\n"
+              "Create a PyProxy"),
```
rth, line 1354:
```suggestion
              "Create a JsProxy from a Python object."),
```

src/core/pyproxy.c
```diff
@@ -1220,14 +1230,136 @@ EM_JS_NUM(int, pyproxy_init_js, (), {
...
+  return result;
+}
+
+static PyMethodDef pyproxy_methods[] = {
+  {
+    "create_once_callback",
+    create_once_callback_py,
+    METH_O,
+    PyDoc_STR(
+      "create_once_callback(obj : Callable) -> JsProxy"
```
rth, line 1341: Maybe,
```suggestion
      "create_proxy_once(obj : Callable) -> JsProxy"
```
 to be more consistent with the one below? For me a callback is something that is meant to be called repeatedly, and even if we were in pure Python it's strange to have a function that creates a callback object. There is nothing that distinguishes a callback function from a non callback function generally (or in this case a Promise like object), unless I'm missing something? 

Also the docstring should mention that then and finally methods?

src/pyodide-py/pyodide/_core.py
```diff
@@ -16,5 +22,19 @@ class JsProxy:  # type: ignore
 
         # Defined in jsproxy.c
 
+        # Defined in jsproxy.c
+
+    def create_once_callback(obj: Callable) -> JsProxy:
+        """Wrap a Python callable in a Javascript function that can be called
+        once. After being called the proxy will decrement the reference count
+        of the Callable. The javascript function also has a `destroy` API that
+        can be used to release the proxy without calling it.
```
rth, line 31: I think you can remove the docstring short of a 1 line description. Otherwise we would have to keep it in sync. This function doesn't do this anyway.

src/core/hiwire.c
hoodmane: > Could we use something more explicit?

Perfectly happy to if you have a suggestion. For what it's worth, the Python equivalents are `PyObject_CallFunctionObjArgs` and `PyObject_CallMethodObjArgs`. (I also think we would do well to move to mixed snake / camel case naming in hiwire like what Python does.)

src/pyodide-py/pyodide/_core.py
hoodmane: Well this is the version that will appear in the api docs.

src/core/jsproxy.c
hoodmane: Yeah, before we just allowed it to go through `JsProxy_Vectorcall`. This leaked the `onFulfilled` and `onResolved` handles, and you can't fix it with `create_once_proxy` because exactly one of `onFulfilled` and `onResolved` gets called. Actually the implementation of `catch` is still wrong though.

src/core/pyproxy.c
hoodmane: I guess I meant that the argument should be a function of some sort, and the resulting object can only be used to call the function as opposed to `create_proxy` which makes sense for any Python argument and creates a proxy that can be used in other ways. I'm not sure what the best way to indicate this in the names though. `create_once_proxy` sounds better to me for word order than `create_proxy_once` though.

What about `create_once_callable`?

rth: > What about create_once_callable?

Sounds good.

src/pyodide-py/pyodide/_core.py
rth: Right, forgot about that.

src/core/hiwire.c
rth: `hiwire_CallFunctionObjArgs` ?  I guess keeping the CPython C naming conventions would make sense.

But no strong feeling about it, we can also keep it as is and improve names in some other PR.

src/pyodide-py/pyodide/_core.py
```diff
@@ -16,5 +22,23 @@ class JsProxy:  # type: ignore
 
         # Defined in jsproxy.c
 
+        # Defined in jsproxy.c
+
+    def create_once_callable(obj: Callable) -> JsProxy:
+        """Wrap a Python callable in a Javascript function that can be called
+        once. After being called the proxy will decrement the reference count
```
rth, line 29: FYI, we use the numpydoc convention and [docstring need a 1 line summary](https://numpydoc.readthedocs.io/en/latest/format.html#sections), followed by an empty line, and then the rest of the docstring.

Though it seems to be rendering fine in [sphinx table summary](https://24620-122663163-gh.circle-artifacts.com/0/home/circleci/repo/docs/_build/html/usage/api-reference.html) so it's probably for to keep it as is.

rth:
Thanks @hoodmane !

PR #1335, opened 2021-03-14
Critical bugfix: Fix UB in JsProxy_Call

hoodmane:
The C compiler hates me. This was introduced in #1033.


rth:
Well the issue is that, particularly after doing a bit of Rust, that the compiler is not very helpful there. Is there a way we could enable more static analysis, or use third party tools to do that? 

I think at the very least we should compile core with `-Wall` (possibly `-Wextra`) and maybe enable a few more warnings as errors.

rth:
Yes, this should have been caught as `-Wsometimes-uninitialized` warning / error by the compiler. Cf #1336

PR #1336, opened 2021-03-14
MAINT Compile core with -Wall

rth:
Related to https://github.com/iodide-project/pyodide/pull/1335 adds `-Wall` to CFLAGS for core.

Fixes a few unused variables. The function `_python2js_bytes` is also never used currently, I'm not sure if it's an error or if we should remove it.

There are a few remaining `-Wsometimes-uninitialized` warnings, such as,
```
src/core/jsproxy.c:847:3: warning: variable 'idkwargs' is used uninitialized whenever 'if' conditio
n is true [-Wsometimes-uninitialized]                                                              
  FAIL_IF_NONZERO(Py_EnterRecursiveCall(" in JsProxy_Vectorcall"));                                
  ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~                                 
src/core/error_handling.h:166:9: note: expanded from macro 'FAIL_IF_NONZERO'                       
    if ((num) != 0) {                                                          \                   
        ^~~~~~~~~~                                                                                 
src/core/jsproxy.c:889:16: note: uninitialized use occurs here                                     
  hiwire_CLEAR(idkwargs);                                                                          
               ^~~~~~~~                                                                            
src/core/hiwire.h:49:19: note: expanded from macro 'hiwire_CLEAR'                                  
    hiwire_decref(x);                                                          \                   
                  ^                                                                                
src/core/jsproxy.c:847:3: note: remove the 'if' if its condition is always false                   
  FAIL_IF_NONZERO(Py_EnterRecursiveCall(" in JsProxy_Vectorcall"));                                
  ^
src/core/error_handling.h:166:5: note: expanded from macro 'FAIL_IF_NONZERO'
    if ((num) != 0) {                                                          \
    ^
src/core/jsproxy.c:849:3: note: variable 'idkwargs' is declared here
  JsRef idkwargs = NULL; 
```
cc @hoodmane 
we probably should add `-Werror=sometimes-uninitialized` to the compilation flags.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -137,7 +137,7 @@ benchmark: all
 
 clean:
 	rm -fr build/*
-	rm -fr src/*.o
+	rm -fr src/*/*.o
 	rm -fr node_modules
 	make -C packages clean
 	echo "The Emsdk, CPython are not cleaned. cd into those directories to do so."
--- a/Makefile
+++ b/Makefile
@@ -18,7 +18,9 @@ CFLAGS=\
 	-g \
 	-I$(PYTHONINCLUDE) \
 	-fPIC \
+	-Wall \
 	-Wno-warn-absolute-paths \
+	-Werror=unused-variable \
 	-Werror=int-conversion \
 	-Werror=incompatible-pointer-types \
 	$(EXTRA_CFLAGS)
--- a/Makefile
+++ b/Makefile
@@ -21,6 +21,7 @@ CFLAGS=\
 	-Wall \
 	-Wno-warn-absolute-paths \
 	-Werror=unused-variable \
+	-Werror=sometimes-uninitialized \
 	-Werror=int-conversion \
 	-Werror=incompatible-pointer-types \
 	$(EXTRA_CFLAGS)
--- a/src/core/jsproxy.c
+++ b/src/core/jsproxy.c
@@ -199,8 +199,6 @@ finally:
 static PyObject*
 JsProxy_RichCompare(PyObject* a, PyObject* b, int op)
 {
-  JsProxy* aproxy = (JsProxy*)a;
-
   if (!JsProxy_Check(b)) {
     switch (op) {
       case Py_EQ:
@@ -818,6 +816,7 @@ JsMethod_Vectorcall(PyObject* self,
                     PyObject* kwnames)
 {
   bool kwargs = false;
+  bool success = false;
   if (kwnames != NULL) {
     // There were kwargs? But maybe kwnames is the empty tuple?
     PyObject* kwname = PyTuple_GetItem(kwnames, 0); /* borrowed!*/
@@ -846,7 +845,6 @@ JsMethod_Vectorcall(PyObject* self,
 
   // Recursion error?
   FAIL_IF_NONZERO(Py_EnterRecursiveCall(" in JsProxy_Vectorcall"));
-  bool success = false;
   JsRef idargs = NULL;
   JsRef idkwargs = NULL;
   JsRef idarg = NULL;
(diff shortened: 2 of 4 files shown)


src/core/jsproxy.c
```diff
@@ -846,7 +845,6 @@ JsMethod_Vectorcall(PyObject* self,
 
   // Recursion error?
   FAIL_IF_NONZERO(Py_EnterRecursiveCall(" in JsProxy_Vectorcall"));
-  bool success = false;
```
rth, line 849: Here it was used uninitialized in the finally block if the `FAIL_IF_NONZERO` above failed.

hoodmane: It's weird though: all the following declarations are ALSO used uninitialized when there's a recursion error, but somehow `-Werror=sometimes-uninitialized` only caught the first one?

hoodmane: Funny thing though, the following declarations are also used uninitialized in that case but the compiler only warns about the first...

src/core/jsproxy.c
```diff
@@ -818,6 +816,7 @@ JsMethod_Vectorcall(PyObject* self,
                     PyObject* kwnames)
 {
   bool kwargs = false;
+  bool success = false;
```
hoodmane, line 819:
```suggestion
  bool success = false;
  bool success = false;
  JsRef idargs = NULL;
  JsRef idkwargs = NULL;
  JsRef idarg = NULL;
  JsRef idresult = NULL;
  PyObject* pyresult = NULL;
```

src/core/jsproxy.c
```diff
@@ -846,7 +845,6 @@ JsMethod_Vectorcall(PyObject* self,
 
   // Recursion error?
   FAIL_IF_NONZERO(Py_EnterRecursiveCall(" in JsProxy_Vectorcall"));
-  bool success = false;
   JsRef idargs = NULL;
   JsRef idkwargs = NULL;
   JsRef idarg = NULL;
```
hoodmane, line 850:
```suggestion
  JsRef idargs = NULL;
  JsRef idkwargs = NULL;
  JsRef idarg = NULL;
```
```suggestion
```

hoodmane: Plus the next two lines won't let me suggest it.

hoodmane (commented):
Would be good also to fix this one:
https://github.com/iodide-project/pyodide/blob/7db54856a61d4b988950c93311457346c821e022/src/core/error_handling.c#L97
should be `PyObject* pystr = NULL`.

hoodmane:
As far as I can tell this also fails to warn about the UB in #1335. I feel a little bit relieved about that (it took me all afternoon yesterday to track down this bug, trigger was completely unrelated), but it's also unfortunate...

hoodmane:
> The function _python2js_bytes is also never used currently

I can remove it on #1215, I think the `getRawBuffer` API does the same thing better.

Makefile
```diff
@@ -137,7 +137,7 @@ benchmark: all
 
 clean:
 	rm -fr build/*
-	rm -fr src/*.o
+	rm -fr src/*/*.o
```
hoodmane, line 140: Oh this has been causing trouble for me for a long time. Funny.

rth:
> As far as I can tell this also fails to warn about the UB in #1335. I feel a little bit relieved about that (it took me all afternoon yesterday to track down this bug, trigger was completely unrelated), but it's also unfortunate...

I think we should invest more into tooling for C code. I'm not sure what other projects with lots of Python C API code do, in addition to code review, but it could be worth investigating. I was thinking that we could enable https://lgtm.com/ after we migrate to a separate org, but I'm not sure how useful it is for C. For Python it can be useful but it also has a fair amount of false positives.

rth:
I think I addressed all comments. @hoodmane feel free to merge if you are OK with it.

Issue #1337, opened 2021-03-14
Use hypothesis in the core test suite


rth:
I think it would make sense to use [hypothesis](https://hypothesis.readthedocs.io/en/latest/index.html) for some tests in the core test suite, rather than writing test parametrization explicitly with pytest. Particularly things like, 
- building arbitrary nested objects for type conversion
- arbitrary string generators including edge cases with unicode etc

It's particularly helpful for checking some conservation properties: e.g. any nested object using some subset of types should be preserved after a py -> js -> py conversion. Writing a hypothesis generator for such types / objects would likely be useful to find falsifying examples.

The downside of hypothesis is that it can generate a large amount of test cases, so it should only be used with the `selenium` fixture (not `selenium_standalone`)

cc @hoodmane

hoodmane:
Looks pretty cool!

hoodmane:
> it can generate a large amount of test cases, so it should only be used with the `selenium` fixture 

The hypothesis docs say that the fixture is run only once per batch of generated tests:

> each fixture will run once for the whole function, not once per example

https://hypothesis.readthedocs.io/en/latest/supported.html?highlight=fixture#testing-frameworks

rth:
We are already doing this, closing as fixed.

PR #1339, opened 2021-03-16
Use hypothesis to test string conversions between js and python

hoodmane:
This PR adds appropriate modifications for hypothesis to the various configuration and as an example uses hypothesis to test our string conversion code, as recommended in #1337.

Impressively, the string conversion code correctly handles all of the weird stuff the hypothesis tester throws at it: null bytes, unprintable characters and control bytes, any combination of quote marks, and arbitrarily weird unicode.


.circleci/config.yml
```diff
@@ -3,7 +3,7 @@ version: 2
 defaults: &defaults
   working_directory: ~/repo
   docker:
-    - image: iodide/pyodide-env:11
+    - image: iodide/pyodide-env:12
```
rth, line 6: You need to first push to some temporary path to test that it's working. And then once the CI passes push that image to this location and update the path.

Could you send me your Docker Hub username, so I can give you write permission to that Docker org?

rth:
> Impressively, the string conversion code correctly handles all of the weird stuff the hypothesis tester throws at it: null bytes, unprintable characters and control bytes, any combination of quote marks, and arbitrarily weird unicode.

Great! It is pretty good at checking edge cases.

hoodmane:
> It is pretty good at checking edge cases.

Yeah it kept breaking my test code until I figured out to transfer it into the browser / pyodide as bytes and then decode. I tried three other string escaping approaches and none were sufficient.

conftest.py
```diff
@@ -243,37 +242,35 @@ def selenium_standalone(request, web_server_main):
...
-        selenium = cls(
-            build_dir=request.config.option.build_dir,
-            server_port=server_port,
-            server_hostname=server_hostname,
-            server_log=server_log,
-        )
+    @pytest.fixture(params=["firefox", "chrome"], scope="function")
+    def selenium_standalone(request, web_server_main):
         try:
+            selenium = next(selenium_common(request, web_server_main))
```
hoodmane, line 250: @rth Is this a correct way to compose generators? I'm not sure that the finally block code from `selenium_common` ever gets run...

hoodmane: I think it doesn't.

rth: Something like,
```py
with selenium_common(..) as selenium:
    yield selenium
```
doesn't work?

hoodmane: Okay, it does. Probably what happens in the generator destructor looks something like:
```py
if self.gi_running:
    self.throw(StopIteration(None))
```

hoodmane:
```py
with selenium_common(..) as selenium: 
   yield selenium
```
would work, but then I have to decorate `selenium_common` with `@contextlib.contextmanager`. Maybe that's the best approach.

rth: >  but only if we decorate selenium_common with @contextlib.contextmanager

+1

hoodmane: Yeah okay what I said above about generator destructors is correct except instead of `StopIteration` it throws `GeneratorExit`.

hoodmane: Doing
```py
with selenium_common(..) as selenium: 
   yield selenium
```
runs finalizers from inside out (which is probably what you expect), doing what I did runs the finalizers from outside in.

.circleci/config.yml
hoodmane: I have "hoodmane" at Docker Hub too.

> You need to first push to some temporary path to test that it's working. And then once the CI passes push that image to this location and update the path.

I don't understand this.

rth: > I have "hoodmane" at Docker Hub too.

Hmm, looks like we can have 3 people max in https://hub.docker.com/orgs/iodide presumably under the free plan. We probably need to request OSS support for https://hub.docker.com/orgs/pyodide but   that can take some time.

> I don't understand this.

You can,
1. Build the image locally as `hoodmane/pyodide-env:12` (or some other version) and push it there.
2. Update the path in this PR to `hoodmane/pyodide-env:12`, make sure the CI works
3. Now push that image to `iodide/pyodide-env:12` (in this case I'll probably have to do it, while we figure out Docker hub org size issue)
4. Update the path in the PR back to `iodide/pyodide-env:12`

It's a bit tedious but I haven't found a better solution so far.

hoodmane: Okay I did steps 1+2.

conftest.py
```diff
@@ -243,37 +243,36 @@ def selenium_standalone(request, web_server_main):
...
+            yield selenium_module_scope
         finally:
-            print(_selenium_cached.logs)
+            print(selenium_module_scope.logs)
+
+    selenium = pytest.fixture(selenium_per_function)
+    # Hypothesis is unhappy with function scope fixtures. Instead, use the
+    # module scope fixture `selenium_module_scope` and use:
+    # `with selenium_context_manager(selenium_module_scope) as selenium`
+    selenium_context_manager = contextlib.contextmanager(selenium_per_function)
```
rth, line 275: I'm somewhat surprised that you are able to use this as a normal context manager even though it depends on a pytest fixture. Though if it works it should be fine.

hoodmane: This works because I pass the fixture in myself:
```py
@given(s=text())
def test_string_conversion(selenium_module_scope, s):
    with selenium_context_manager(selenium_module_scope) as selenium:
```

Issue #703, opened 2020-06-28
Restart pyodide and clear user defined variables


therealadityashankar:
Hey !

I'm currently using pyodide to build a project and realized that pyodide uses only one interpreter throughout its lifetime

a really neat option would be to be able to do `pyodide.reset()` that clears any new variables, the implementation would be really simple as all you'd have to do is clear globals()

rth:
Ipython has something [similar](https://ipython.readthedocs.io/en/stable/interactive/magics.html#magic-reset). We could try to implement it, but I'm not sure it's as simple as clearning globals. Global namespace does start with some objects so we need to preserve those.

For instance, Pyodide CLI at init starts with,
```
>>> print(list(globals()))
['__name__', '__doc__', '__package__', '__loader__', '__spec__', '__annotations__', '__builtins__', '__build_class__', '__import__', 'abs', 'all', 'any', 'ascii', 'bin', 'brea
kpoint', 'callable', 'chr', 'compile', 'delattr', 'dir', 'divmod', 'eval', 'exec', 'format', 'getattr', 'globals', 'hasattr', 'hash', 'hex', 'id', 'input', 'isinstance', 'issu
bclass', 'iter', 'len', 'locals', 'max', 'min', 'next', 'oct', 'ord', 'pow', 'print', 'repr', 'round', 'setattr', 'sorted', 'sum', 'vars', 'None', 'Ellipsis', 'NotImplemented'
, 'False', 'True', 'bool', 'memoryview', 'bytearray', 'bytes', 'classmethod', 'complex', 'dict', 'enumerate', 'filter', 'float', 'frozenset', 'property', 'int', 'list', 'map',
 'object', 'range', 'reversed', 'set', 'slice', 'staticmethod', 'str', 'super', 'tuple', 'type', 'zip', '__debug__', 'BaseException', 'Exception', 'TypeError', 'StopAsyncItera
tion', 'StopIteration', 'GeneratorExit', 'SystemExit', 'KeyboardInterrupt', 'ImportError', 'ModuleNotFoundError', 'OSError', 'EnvironmentError', 'IOError', 'EOFError', 'Runtim
eError', 'RecursionError', 'NotImplementedError', 'NameError', 'UnboundLocalError', 'AttributeError', 'SyntaxError', 'IndentationError', 'TabError', 'LookupError', 'IndexError
[...]

PR #1340, opened 2021-03-17
Universally check for hiwire memory leaks in test suite

hoodmane:
I updated `conftest.py` to check that the number of hiwire keys at the end of the test is the same as the number of hiwire keys at the end of the test. This more or less requires that we never leak any hiwire keys. It'll require a fair amount of work to make the test suite pass like this.


hoodmane:
I think that most of the remaining test failures represent actual memory leaks and not issues with the tests...

hoodmane:
Merging #1334 will probably fix most of the asyncio-related tests, and then there seems to be another buffer memory leak, and some assortment of random other issues.

rth:
This is quite interesting!

I wonder if we should implement some object to take a snapshot the keys in `globals()` and window, and then automatically delete additional keys at the end of the test in the fixture, to avoid having to do that manually. Related to https://github.com/iodide-project/pyodide/issues/703 That way created objects would be automatically cleaned at the end of the test (and you can do these reference counting checks afterwards)

hoodmane:
Yeah it sure does suggest doing that. I think there is a good way to implement #703 now aside from the fact that we would leak the old `globals` dict. I think we can just call `old_globals.clear()` and then we're only leaking an empty dict, which is better than the alternative of forcefully freeing it (if someone *actually* has an old reference lying around, in one case they see in empty dict, in the other case we crash).

hoodmane:
Actually all we probably have to do is call `pyodide.globals.clear()` at the end of every test.

hoodmane:
It would be helpful if the "short test summary info" used some standard sort order for easier comparisons.

hoodmane:
@rth Would appreciate review on this one, I think it's ready to merge. The important stuff is in `conftest.py`, the rest is just a lot of cleanup and minor fixes in the test suite, or already got merged in one of the separate PRs.

conftest.py
```diff
@@ -242,6 +252,42 @@ def get_driver(self):
...
+
+    Pytest issue #5044:
+    https://github.com/pytest-dev/pytest/issues/5044
+    """
+    selenium = None
+    if "selenium" in item._fixtureinfo.argnames:
+        selenium = item.funcargs["selenium"]
+    if "selenium_standalone" in item._fixtureinfo.argnames:
+        selenium = item.funcargs["selenium_standalone"]
+    if selenium and pytest.mark.norefs.mark not in item.own_markers:
```
rth, line 273:
```suggestion
    if selenium and pytest.mark.skip_refcount_check.mark not in item.own_markers:
```
to be more explicit.

packages/matplotlib/test_matplotlib.py
```diff
@@ -1,12 +1,21 @@
+import pytest
+
+
+@pytest.mark.norefs
```
rth, line 4:
```suggestion
@pytest.mark.skip_refcount_check
```

here and everywhere to be more explicit

pytest.ini
```diff
@@ -0,0 +1,3 @@
+[pytest]
+markers =
+    norefs: Dont run refcount checks
```
rth, line 3: This should go into setup.cfg which already has a pytest config section

src/tests/test_jsproxy.py
```diff
@@ -389,65 +392,36 @@ def test_mount_object(selenium):
...
+        return result;
         """
     )
     assert result[:3] == ["x1", "x2", 3]
     assert set([x for x in result[3] if len(x) == 1]) == set(["x", "y", "s", "t"])
     assert set([x for x in result[4] if len(x) == 1]) == set(["x", "y", "u", "t"])
-
-
-@pytest.mark.xfail
-def test_mount_map(selenium):
```
rth, line 409: It's redundant with existing tests?

src/tests/test_jsproxy.py
```diff
@@ -739,16 +723,13 @@ def raises(exc, match=None):
 
 
 def test_memory_leaks(selenium):
-    assert 1 == selenium.run_js(
+    selenium.run_js(
```
rth, line 726: Maybe add a comment that refcounts are checked elsewhere

rth (commented):
Thanks! A few comments, otherwise LGTM.

src/tests/test_jsproxy.py
hoodmane: It tests for behavior that was never implemented and that I'm not really planning to implement. Note the `xfail` marker.

PR #1341, opened 2021-03-19
Minor improvements to build handling of PYODIDE_PACKAGES

hoodmane:
Allow spaces, leading / trailing comma in `PYODIDE_PACKAGES`. This makes `PYODIDE_PACKAGES="somepkg" make debug` work as expected (builds `somepkg` in addition to the standard list). 

Currently would result in `PYODIDE_PACKAGES` that looks like `somepkgmicropip,...` which leads to an error because no package named `somepkgmicropip` exists. Currently you can say `PYODIDE_PACKAGES="somepkg," make debug` but `PYODIDE_PACKAGES="somepkg," make` doesn't work because of the trailing comma.

### DIFF
--- a/pyodide_build/buildall.py
+++ b/pyodide_build/buildall.py
@@ -129,8 +129,6 @@ def generate_dependency_graph(
 
     while packages:
         pkgname = packages.pop()
-        if not pkgname:
-            continue
 
         pkg = Package(packages_dir / pkgname)
         pkg_map[pkg.name] = pkg
--- a/pyodide_build/buildall.py
+++ b/pyodide_build/buildall.py
@@ -302,7 +302,7 @@ def make_parser(parser):
         nargs="?",
         default=None,
         help=(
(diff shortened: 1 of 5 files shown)


pyodide_build/buildall.py
```diff
@@ -129,6 +129,8 @@ def generate_dependency_graph(
 
     while packages:
         pkgname = packages.pop()
+        if not pkgname:
+            continue
```
rth, line 133: I think we should rather get rid of those in `_parse_package_subset`. Here `pkgname` is str, so the fact that it can be an empty string is worrisome (or otherwise it means that type annotations are not correct)

hoodmane: It's filtering empty string. If `PYODIDE_PACKAGES="micropip, " then `packages=["micropip", ""]`.

hoodmane: Updated.

PR #1349, opened 2021-03-19
Add pyodide._module.resetState (from #1340)

hoodmane:
Add a command to reset the state of pyodide as much as we can manage. This is intended for testing purposes, but could also be exposed as a public API for #703.

### DIFF
--- a/src/pyodide-py/pyodide/_state.py
+++ b/src/pyodide-py/pyodide/_state.py
@@ -1,0 +1,40 @@
+import __main__
+import sys
+import gc
+
+from ._core import JsProxy
+from ._importhooks import jsfinder
+
+
+def save_state():
+    loaded_js_modules = {}
+    for [key, value] in sys.modules.items():
+        if isinstance(value, JsProxy):
+            loaded_js_modules[key] = value
+
+    return dict(
+        globals=dict(__main__.__dict__),
+        js_modules=dict(jsfinder.jsproxies),
+        loaded_js_modules=loaded_js_modules,
+    )
+
+
+def restore_state(state):
+    __main__.__dict__.clear()
+    __main__.__dict__.update(state["globals"])
+
+    jsfinder.jsproxies = state["js_modules"]
+    loaded_js_modules = state["loaded_js_modules"]
+    for [key, value] in list(sys.modules.items()):
+        if isinstance(value, JsProxy) and key not in loaded_js_modules:
+            del sys.modules[key]
+    sys.modules.update(loaded_js_modules)
+
+    sys.last_type = None
+    sys.last_value = None
+    sys.last_traceback = None
+
+    return gc.collect(2)
+
+
+__all__ = ["save_state", "restore_state"]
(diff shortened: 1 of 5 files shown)


src/pyodide.js
```diff
@@ -679,11 +679,43 @@ def temp(Module):
...
+  saved_globals.update(globals)
+
   Module.version = pyodide.__version__
   Module.globals = globals
   Module.builtins = builtins.__dict__
   Module.pyodide_py = pyodide
 `);
+  Module.resetState = function() {
+    pyodide.globals.clear();
+    Module.runPythonSimple(`
```
rth, line 693: I think we need to add documentation on runPythonSimple and when it should be used instead of runPython. Right now that's not clear to me.

src/pyodide.js
```diff
@@ -679,11 +679,43 @@ def temp(Module):
   globals = __main__.__dict__
   globals.update(builtins.__dict__)
 
+  global saved_globals
+  saved_globals = {}
+  saved_globals.update(globals)
```
rth, line 684: It's surprising to always make a copy of globals even when the user hasn't asked for it.

Couldn't we more explicitly do,
```py
state = Module.saveState()

# some time later

Module.resetState(state)
```
?

src/pyodide.js
hoodmane: The primary point of `runPythonSimple` is as a bootstrap method. It doesn't depend on any of the other code in `core` so it will work as soon as Python is initialized even if everything else goes really wrong.

The differences are: 
1. `runPythonSimple` doesn't return anything (and so won't leak PyProxies)
2. `runPythonSimple` doesn't require access to any state on the `pyodide_js` module. 
3. `runPython` uses `pyodide.eval_code`, whereas `runPythonSimple` uses `PyRun_String` which is the C API for `eval` / `exec`.
4. `runPythonSimple` runs with `globals` a separate dict which I've called `init_dict`
5. `runPythonSimple` doesn't dedent the argument

src/pyodide.js
hoodmane: Sure.

src/pyodide.js
rth: It would be good to add that to the docstring of `runPythonSimple` (not in this PR)

src/pyodide-py/pyodide/_state.py
```diff
@@ -0,0 +1,40 @@
...
+            loaded_js_modules[key] = value
+
+    return dict(
+        globals=dict(__main__.__dict__),
+        js_modules=dict(jsfinder.jsproxies),
+        loaded_js_modules=loaded_js_modules,
+    )
+
+
+def restore_state(state):
```
rth, line 22: Please add type annotations and at least a 1 line summary docstring if they are private.

rth (approved):
Otherwise LGTM.

pauleveritt:
Did this later get removed?

hoodmane:
No, but it's been moved around. Currently these functions are here:
https://github.com/pyodide/pyodide/blob/main/src/py/pyodide/_state.py

PR #1350, opened 2021-03-19
Always fail CI if we have to rerun failed tests

hoodmane:


rth:
Right but then we will still get broken CI on master occasionally which we wanted to avoid with that wrapper script.

hoodmane:
Hmm okay. Would be nice to have the more lenient behavior only on master and the more strict behavior on PRs, but I'm not sure how to do that. Maybe have the wrapper return different return codes 0 if passed first try, 1 if passed on retest, 2 if failed, then make the CI script check if we're on master and accept a 1 return code in that case. But probably it's fine to leave it as is...

rth:
It could also be worth checking the failure stats. If the core tests rarely fail (which seem to be the case aside for the `test_memory_leaks` issue which needs to be fixed) the first time they are run with the current version of the code, we could as well use pytest directly without the wrapper for them. And so return an exit status of 1 on first failure. While using the wrapper for the package tests, which appear to be more likely to fail.

hoodmane:
`test_memory_leaks` is fixed in #1340  which should be ready to merge as soon as I update it to reflect #1349.

PR #1352, opened 2021-03-21
Automatically generatate C method->ml_doc docstrings from Python

hoodmane:
This generates the docstrings on publicly exposed builtin methods from the stubs in `_core.py`. This has the advantage that the sphinx docstrings and the live-docstrings come from the same place. Also, there are very specific invariants expected of the signature line on the C docstrings (the parser is very simple and when it fails it does so quietly). This approach ensures that the signature line is exactly right.

I put the `_core` stubs in an inner python package called `_pyodide` that is loaded from core while initialization is in progress. This should make it possible to move more code from C into Python in the future, the main obstacle to this in the past has been bootstrapping issues, which are resolved by having an inner and outer package. (This is also the approach used in CPython, they have a bunch of files with names like `_bootstrap_inner.py`, `_bootstrap.py`, `_bootstrap_outer.py` which are safe to load at varying points.)


src/pyodide-py/pyodide/_core.py
```diff
@@ -1,44 +1,10 @@
...
-        create_once_callable,
-    )
-else:
-    # Can add shims here if we are so inclined.
-    class JsException(Exception):  # type: ignore
-        """
-        A wrapper around a Javascript Error to allow the Error to be thrown in Python.
-        """
+if "_pyodide_core" not in sys.modules:
+    from _pyodide import _core as _pyodide_core
```
rth, line 4: Maybe,
```suggestion
    from _pyodide_inner import _core as _pyodide_core
```
to make it clear that it's really not a module the user would be interested in. The difference between pyodide and _pyodide is not that clear.

src/pyodide-py/_pyodide/_core.py
```diff
@@ -0,0 +1,76 @@
...
+
+from typing import Any, Callable
+
+# All docstrings for public `core` APIs should be extracted from here. We use
+# the utilities in `docstring.py` and `docstring.c` to format them
+# appropriately.
+
+# Sphinx uses __name__ to determine the paths and such. It looks better for it
+# to refer to e.g., `pyodide.JsProxy` than `_pyodide._core.JsProxy`.
+__name__ = "pyodide"
```
rth, line 11: As long as it doesn't break the import mechanism otherwise..

src/pyodide-py/_pyodide/_core.py
```diff
@@ -0,0 +1,76 @@
...
+
+# From jsproxy.c
+class JsException(Exception):
+    """
+    A wrapper around a Javascript ``Error`` to allow the ``Error`` to be thrown in Python.
+    """
+
+
+class JsProxy:
+    """A proxy to make a Javascript object behave like a Python object"""
```
rth, line 21: Great! Good to have them here, and it would certainly simplify sphinx generation.

src/pyodide-py/_pyodide/docstring.py
```diff
@@ -0,0 +1,31 @@
...
+    return docstring
+
+
+def get_cmeth_docstring(func):
+    """Get the value to use for the PyMethodDef.ml_doc attribute for a builtin
+    function. This is used in docstring.c.
+
+    The ml_doc should start with a signature which cannot have any type
+    annotations. The signature must end with the exact characters ")\n--\n\n".
+    For example: "funcname(arg1, arg2)\n--\n\n"
```
rth, line 18: Maybe we could do a doctest here for some builtin function that's unlikely to change such as "sum" ?

src/pyodide-py/_pyodide/docstring.py
```diff
@@ -0,0 +1,31 @@
+from textwrap import dedent
+
+
+def dedent_docstring(docstring):
```
rth, line 4: It might help to have a doctest to have an example what this function does.

src/core/jsproxy.c
```diff
@@ -1533,8 +1578,30 @@ JsProxy_init(PyObject* core_module)
...
+  _Py_IDENTIFIER(JsProxy);
+  jsproxy_mock =
+    _PyObject_CallMethodIdObjArgs(_pyodide_core, &PyId_JsProxy, NULL);
+  FAIL_IF_NULL(jsproxy_mock);
+
+#define SET_DOCSTRING(x)                                                       \
+  FAIL_IF_MINUS_ONE(set_method_docstring(&x, jsproxy_mock))
+  SET_DOCSTRING(JsProxy_object_entries_MethodDef);
+  SET_DOCSTRING(JsProxy_object_keys_MethodDef);
+  SET_DOCSTRING(JsProxy_object_values_MethodDef);
```
rth, line 1596: Could you please add a comment about the effect of these `SET_DOCSTRING` macros?

rth (commented):
Thanks for doing this work! Looks very useful indeed.

Could we then add JsProxy to API reference?

src/pyodide-py/pyodide/_core.py
hoodmane: I think it's a standard thing for Python packages to have private versions with a leading underscore. For instance pytest does this:
https://github.com/pytest-dev/pytest/tree/6.2.2/src

src/pyodide-py/_pyodide/_core.py
hoodmane: Maybe I should restore it at the bottom of the file.

src/pyodide-py/_pyodide/docstring.py
hoodmane: Okay I added a bit of explanation and a doctest.

src/pyodide-py/_pyodide/_core.py
```diff
@@ -0,0 +1,78 @@
...
+        def new(self, *args, **kwargs) -> "JsProxy":
+            """Construct a new instance of the Javascript object"""
+
+        def to_py(self) -> Any:
+            """Convert the :class:`JsProxy` to a native Python object as best as possible"""
+            pass
+
+        def then(self, onfulfilled: Callable, onrejected: Callable) -> "Promise":
+            """The ``Promise.then`` api, wrapped to manage the lifetimes of the handlers.
+            Pyodide will automatically release the references to the handlers when the promise resolves.
```
rth, line 44:
```suggestion
            """The ``Promise.then`` api, wrapped to manage the lifetimes of the handlers.
            
            Pyodide will automatically release the references to the handlers when the
            promise resolves.
```

src/pyodide-py/_pyodide/_core.py
```diff
@@ -0,0 +1,78 @@
...
+            """
+
+        def catch(self, onrejected: Callable) -> "Promise":
+            """The ``Promise.catch`` api, wrapped to manage the lifetimes of the handler.
+            Pyodide will automatically release the references to the handler when the promise resolves.
+            """
+
+        def finally_(self, onfinally: Callable) -> "Promise":
+            """The ``Promise.finally`` api, wrapped to manage the lifetimes of the handler.
+            Pyodide will automatically release the references to the handler when the promise resolves.
```
rth, line 54:
```suggestion
            """The ``Promise.finally`` api, wrapped to manage the lifetimes of the handler.
            
            Pyodide will automatically release the references to the handler when the promise
            resolves.
```

src/pyodide-py/_pyodide/_core.py
```diff
@@ -0,0 +1,78 @@
...
+            pass
+
+        def then(self, onfulfilled: Callable, onrejected: Callable) -> "Promise":
+            """The ``Promise.then`` api, wrapped to manage the lifetimes of the handlers.
+            Pyodide will automatically release the references to the handlers when the promise resolves.
+            """
+
+        def catch(self, onrejected: Callable) -> "Promise":
+            """The ``Promise.catch`` api, wrapped to manage the lifetimes of the handler.
+            Pyodide will automatically release the references to the handler when the promise resolves.
```
rth, line 49:
```suggestion
            """The ``Promise.catch`` api, wrapped to manage the lifetimes of the handler.
            
            Pyodide will automatically release the references to the handler when the promise
            resolves.
```

src/pyodide-py/_pyodide/_core.py
```diff
@@ -0,0 +1,78 @@
...
+__name__ = "pyodide"
+try:
+    # From jsproxy.c
+    class JsException(Exception):
+        """
+        A wrapper around a Javascript ``Error`` to allow the ``Error`` to be thrown in Python.
+        """
+
+    class JsProxy:
+        """A proxy to make a Javascript object behave like a Python object"""
```
rth, line 21:
```suggestion
        """A proxy to make a Javascript object behave like a Python object
        
        For more information see :ref:`type-translations` documentation.
        """
```

rth (approved):
Thanks a few more comments otherwise LGTM. Great to finally have documentation for PyProxy! 

For the `then`, `finally_`, and `catch` methods I think we should indicate in their docstring that they are only defined if the underlying object implements them.

hoodmane:
> For the `then`, `finally_`, and `catch` methods I think we should indicate in their docstring that they are only defined if the underlying object implements them.

In fact we only need the underlying object to have a `then` method (a "thenable"). At some point it'd be nice to figure out how to organize the API docs for the proxies into sections based on collections of gated features. I'm not sure what the best layout / approach for that is though.

rth:
> At some point it'd be nice to figure out how to organize the API docs for the proxies into sections based on collections of gated features. I'm not sure what the best layout / approach for that is though.

As long as the docstring states that it's conditional I think it's fine. Here is another example of a [conditional method in another project](https://scikit-learn.org/stable/modules/generated/sklearn.model_selection.GridSearchCV.html#sklearn.model_selection.GridSearchCV.predict_log_proba) so it's a fairly standard practice.

PR #1353, opened 2021-03-21
DOC Fix formatting in some docstrings

hoodmane:


packages/micropip/micropip/micropip.py
```diff
@@ -239,22 +239,21 @@ def install(requirements: Union[str, List[str]]):
...
+    This only works for packages that are either pure Python or for packages
+    with C extensions that are built in Pyodide. If a pure Python package is not
+    found in the Pyodide repository it will be loaded from PyPi.
 
     Parameters
     ----------
-    requirements
-       A requirement or list of requirements to install.
-       Each requirement is a string.
+    requirements : ``str | List[str]``
```
rth, line 248: Normally you shouldn't have to escape types here. Was the rendering not OK without it?

Of course it would have been better if the information was taken from the type annotations directly (https://github.com/numpy/numpydoc/issues/196) but I'm now aware of a tool that does that reliably yet.

hoodmane: Well I'm not sure how we want the type rendered, but yeah it seems to put render the type annotations as plain not-monospaced text. Maybe there is some way to configure it differently?

PR #1354, opened 2021-03-21
DOC Implement autosummary for js autodoc, use autodocsumm for py autodoc

hoodmane:
This makes it so that both the Javascript and Python docs are automatically generated (both the summary and the main contents are auto generated, and also the list of items that should be documented is chosen fairly automatically). It also ensures a high level of stylistic consistency between the Javascript and Python docs.


rth (commented):
The output looks nice! 

I think we should add one or two sentences of description to [api-reference.html](https://25069-122663163-gh.circle-artifacts.com/0/home/circleci/repo/docs/_build/html/usage/api-reference.html).

I'm a bit concerned how we are going to maintain this code without tests. Maybe we should try to move it to a separate module in doc/ and write at least some unit tests.

hoodmane:
> we should add one or two sentences of description to api-reference.html.

I was looking into whether it would be possible to put the summary tables in `api-reference.md`. I can put the Javascript summary table there, but it would require more code to split the 

> we should try to move it to a separate module in doc/ and write at least some unit tests.

Sounds good to me.

hoodmane:
@rth Okay I got some basic tests working here.

rth (approved):
Thanks @hoodmane !

PR #1356, opened 2021-03-21
Release 0.17.0a2 (alpha 2)

rth:
Makes an alpha 2 release of the current master in preparation for the coming 0.17.0 (final) release. 

Just to have to a link that can be used for testing that's a bit more stable (and better cached by the CDN) than the dev release, and that has the full list of packages that can be downloaded as an archive from Github (https://github.com/iodide-project/pyodide/issues/1316)

These changes will have to be reverted once the CI passes on master.


hoodmane:
I see five extra instances of `0.16.1`, three are in the changelog and obviously mean "the version 0.16.1" and not "latest stable". Then there is the one in `setup.cfg` (don't know what that file does) and `PYODIDE_PREBUILT_IMAGE_TAG="0.16.1"` which probably doesn't need to change?

hoodmane (commented):
Looks fine to me.

rth:
> Then there is the one in setup.cfg (don't know what that file does)

Those are config settings for [bumpversion](https://github.com/peritus/bumpversion) I haven't used it here, so it can be ignored.

>  and PYODIDE_PREBUILT_IMAGE_TAG="0.16.1"

Yeah, it's probably better to upload the pre-built image as well. I'll update it.

PR #1363, opened 2021-03-23
Make "initializePyodide" function

hoodmane:
Closes #277. As discussed in #1239.

Instead of exposing a promise called `languagePluginLoader`, expose a function called `initializePyodide`. For backwards compatibility, if `languagePluginURL` is defined, then it automatically does `initializePyodide({ baseURL : languagePluginURL })`. In this case we also print a deprecation warning.

We need to add some documentation for this.

I also tried to reduce the number of places that need `pyodide` to be in global scope. We still need this to load packages, but I think it's possible to avoid that by using dynamic module imports.

### DIFF
--- a/src/pyodide.js
+++ b/src/pyodide.js
@@ -1,18 +1,12 @@
 /**
  * The main bootstrap script for loading pyodide.
  */
-
-/**
- * A promise that resolves to ``undefined`` when Pyodide is finished loading.
- *
- * @type Promise
- */
-globalThis.languagePluginLoader = (async () => {
+globalThis.initializePyodide = async function(config) {
   let Module = {};
-  // Note: PYODIDE_BASE_URL is an environement variable replaced in
+  // Note: PYODIDE_BASE_URL is an environment variable replaced in
   // in this template in the Makefile. It's recommended to always set
   // languagePluginUrl in any case.
-  let baseURL = self.languagePluginUrl || '{{ PYODIDE_BASE_URL }}';
+  let baseURL = config.baseURL || "{{ PYODIDE_BASE_URL }}";
   baseURL = baseURL.substr(0, baseURL.lastIndexOf('/')) + '/';
 
   ////////////////////////////////////////////////////////////
@@ -52,9 +46,9 @@ globalThis.languagePluginLoader = (async () => {
 
   function recursiveDependencies(names, _messageCallback, errorCallback,
                                  sharedLibsOnly) {
-    const packages = self.pyodide._module.packages.dependencies;
-    const loadedPackages = self.pyodide.loadedPackages;
-    const sharedLibraries = self.pyodide._module.packages.shared_library;
+    const packages = Module.packages.dependencies;
+    const loadedPackages = Module.loadedPackages;
+    const sharedLibraries = Module.packages.shared_library;
     const toLoad = new Map();
 
     const addPackage = (pkg) => {
@@ -153,7 +147,7 @@ globalThis.languagePluginLoader = (async () => {
     let scriptPromises = [];
 
     for (let [pkg, uri] of toLoad) {
-      let loaded = self.pyodide.loadedPackages[pkg];
+      let loaded = Module.loadedPackages[pkg];
       if (loaded !== undefined) {
         // If uri is from the DEFAULT_CHANNEL, we assume it was added as a
         // depedency, which was previously overridden.
@@ -210,7 +204,7 @@ globalThis.languagePluginLoader = (async () => {
 
     let packageList = [];
     for (let [pkg, uri] of toLoad) {
-      self.pyodide.loadedPackages[pkg] = uri;
+      Module.loadedPackages[pkg] = uri;
       packageList.push(pkg);
     }
 
@@ -383,7 +377,6 @@ globalThis.languagePluginLoader = (async () => {
 
   ////////////////////////////////////////////////////////////
   // Loading Pyodide
-  self.Module = Module;
 
   Module.noImageDecoding = true;
   Module.noAudioDecoding = true;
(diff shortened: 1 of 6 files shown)


rth:
Yes, this would be nice! We just need to be make sure the API isn't going to change again if we incorporate something like https://github.com/iodide-project/pyodide/pull/792 in the future. There a `loadPyodide` function was defined, as a name maybe it's a bit more consistent with loadPackage that we already have.

rth (commented):
We need to update all the documentation examples with this, also please add a changelog entry.

Following files also need updating:
 - `src/webworker.js`
 - `src/templates/test.html` and `conftest.py`

I think if the user called `loadPyodide()` without specifying the URL but `languagePluginUrl` was defined we should error instead of producing a warning because it would mean that we would fallback to the hardcoded default which can be different from what the user provided and they can miss the warnings. If a url was provided to `loadPyodide()` and  


> loadPyodide({ baseURL : <some_url>})"

How about rather the `packageIndexURL` or something more explicit than `baseURL` to indicate that's where packages are going to be loaded from.

In the parameter name should also take into account that in the future we may want to extend this to support multiple channels (e.g. use default packages form jsDelivr + a few custom packages from some other location).

src/pyodide.js
```diff
@@ -770,13 +772,42 @@ def temp(Module):
...
+  Module.registerJsModule("js", globalThis);
+  Module.registerJsModule("pyodide_js", pyodide);
+  globalThis.pyodide = pyodide;
+  return pyodide;
+};
+
+if (globalThis.languagePluginUrl) {
+  console.warn(
+      "languagePluginUrl is deprecated and will be removed in version 0.18.0," +
+      "instead use loadPyodide({ baseURL : <some_url>})");
```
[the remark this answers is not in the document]
hoodmane: Currently I've set it up so that it *does* have an effect. These changes are backwards compatible except for the old case when people had left `languagePluginUrl` unset:
https://github.com/iodide-project/pyodide/blob/782a596d6f573f6a8d311b4afb2db5783e942701/src/pyodide.js#L812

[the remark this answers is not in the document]
rth: I meant that it doesn't when calling `loadPyodide` and setting languagePluginUrl. But since it's used for backward compatibility in languagePluginLoader you are right that we can't warn here.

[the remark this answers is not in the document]
hoodmane: Okay I made `loadPyodide` throw an error if called twice and added a special warning if `languagePluginURL` is defined.

hoodmane:
> We need to update all the documentation examples with this, also please add a changelog entry.

I think I did that.

> Following files also need updating:
>
>    src/webworker.js
>    src/templates/test.html and conftest.py

Done.

> I think if the user called `loadPyodide()` without specifying the URL but `languagePluginUrl` was defined we should error instead of producing a warning 

I made it throw an error if `loadPyodide` is called twice, does that cover this?

> How about rather the `packageIndexURL` or something more explicit than `baseURL` to indicate that's where packages are going to be loaded from.

Done.

> In the parameter name should also take into account that in the future we may want to extend this to support multiple channels (e.g. use default packages form jsDelivr + a few custom packages from some other location).

I think no action is needed on this? That sounds like a good idea, but I think it would be fine to add a new named parameter if/when we add support for multiple channels.

rth (commented):
@hoodmane Can you confirm that core tests (and in particular webworker tests) pass for you locally with this PR?

I'm getting some failures locally (when not using the pytest wrapper) including on master related to #1389, so I'm not sure what we can trust anymore )


Otherwise LGTM.

docs/usage/quickstart.md
```diff
@@ -1,50 +1,51 @@
...
-scope called {js:mod}`pyodide`.
+The `pyodide.js` file defines a single async function called
+{any}`loadPyodide <globalThis.loadPyodide>` which sets up the Python
+environment. When the `loadPyodide` function finishes, Pyodide installs global
+namespace called {js:mod}`pyodide`.
 
 ```pyodide
-languagePluginLoader.then(() => {
+async function main() {
+  await loadPyodide({ packageIndexURL : "https://cdn.jsdelivr.net/pyodide/v0.17.0a2/full/" });
```
rth, line 23: Maybe we could ask for another opinion on this parameter name from someone?  We won't be able to change it later. I know I suggested it, but I can't say I'm particularly happy with the current name either. Maybe I should look what names similar parameters have in other libraries.

hoodmane: `packageIndexURL` is at least better than `baseURL`. We can change it later by adding a new named option which is synonymous to the current one and changing the docs to reflect the new name while keeping the old option as a synonym for backwards compatibility. No doubt that it's preferable to pick a good name the first time though.

hoodmane:
> Can you confirm that core tests (and in particular webworker tests) pass for you locally with this PR?

When I ran the entire test suite, 3 of the 6 webworker tests failed with `TimeoutException`s, and every test in `test_typeconversions` failed with `SessionNotCreatedException`. When I reran just the two files `test_webworker` and `test_typeconversions`, all tests passed.

docs/usage/quickstart.md
rth: So,
 - pip has a `index-url` setting https://pip.pypa.io/en/stable/user_guide/#config-file
 - conda uses the term "channel_url" https://conda.io/projects/conda/en/latest/user-guide/configuration/use-condarc.html#show-channel-urls-show-channel-urls
 - nmp uses the "registry url" term https://docs.npmjs.com/configuring-your-registry-settings-as-an-npm-enterprise-user#setting-your-companys-npm-enterprise-registry-as-your-default-registry
 - cargo uses both registry and index terms https://doc.rust-lang.org/cargo/reference/registries.html#using-an-alternate-registry

Overall how about going with a `indexURL` parameter? Which is more or less what pip and cargo have. It's a bit shorter/ more readable than `packageIndexURL`.

hoodmane: I'm happy with `indexURL`.

rth: OK, I renamed the parameter as I would like to merge this PR (since I have another one that depends on it)

PR #1364, opened 2021-03-23
Use "import" to import packages to avoid global variables

hoodmane:
This PR is on top of #1363 and it makes it possible to use `pyodide` without requiring there to be a `pyodide` global variable. Ideally it would be good to add some tests to try to check this behavior, but for now I've just avoided making a `pyodide` global variable in console.html and checked that it's possible to load and import `micropip` and use it to install and import other packages.

This also allows a simplification to `loadPackage` by removing the global error handler.


hoodmane:
Guess this doesn't work in webworkers in Firefox.

PR #1367, opened 2021-03-23
Deprecate pyodide.pyimport

rth:
Closes https://github.com/iodide-project/pyodide/issues/926

Deprecates `pyodide.pyimport` in favor of using `pyodide.globals.get`

### DIFF
--- a/src/pyodide.js
+++ b/src/pyodide.js
@@ -542,6 +542,10 @@ globalThis.languagePluginLoader = (async () => {
 
   /**
    * Access a Python object in the global namespace from Javascript.
+   *
+   * Note: this function is deprecated and will be removed in version 0.18.0.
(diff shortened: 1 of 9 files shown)


src/pyodide.js
```diff
@@ -503,23 +502,23 @@ globalThis.languagePluginLoader = (async () => {
 
   // clang-format off
   /**
-   * Inspect a Python code chunk and use :js:func:`pyodide.loadPackage` to load any known 
-   * packages that the code chunk imports. Uses
```
rth, line 507: Looks like pre-commit wasn't applied consistently in previous PRs...

hoodmane (approved):
test_numpy.py also uses pyimport, we should replace it there too.

PR #1368, opened 2021-03-24
Split PyProxy method definitions into separate js file

hoodmane:
This moves the PyProxy Javascript definitions into a separate `pyproxy.js` file which we #include into `pyproxy.c`. The point of this is so that JsDoc can generate API docs for pyproxy methods. I'll need to modify the sphinx extension code a bit to get the generated API docs to look reasonable, but I figured it would be good to move the files around as a separate PR.

Note that JsDoc needs the entire file to parse as valid Javascript. In order to ensure this, we use a hack with macros, see the comment at the top of pyproxy.js.


src/core/pyproxy.js
```diff
@@ -0,0 +1,565 @@
...
+// look like a function call, which we do with `()=>{`. However, `()=>{` is an
+// invalid C string so the macro needs to remove it. We put `()=>{0,`,
+// TEMP_EMJS_HELPER removes everything up to the comma and replace it with a
+// single open brace.
+// 
+// See definition of TEMP_EMJS_HELPER:
+// #define TEMP_EMJS_HELPER(a, args...)                                           \
+//   EM_JS(int, pyproxy_init, (), UNPAIRED_OPEN_BRACE { args return 0; })
+
+// clang-format off
```
rth, line 17: So we disable clang-format for all of the file? Was the formatting really not good?

rth (approved):
Generally LGTM, once the CI passes. Thanks!

src/core/pyproxy.js
hoodmane: It seems that we had disabled clang-format for all ~600 lines of `pyproxy_init_js` in `pyproxy.c` before this PR:
https://github.com/iodide-project/pyodide/blob/f2d6137673e0040e69b05b69a8fd0a0f8c98bd7f/src/core/pyproxy.c#L690

I just checked what happens with the linter turned on and it's... not bad at all. Thanks for pointing that out.

PR #1369, opened 2021-03-24
Move all configuration variables for compile into Makefile.envs

joemarshall:
Because if you're fiddling with the build at all it is a pain that settings for building are spread between:
/pyodide_build/common.py
/cpython/Makefile,
/Makefile
/Makefile.envs

I don't think this is urgent, so probably can hold off until after the coming release if it needs to, but it just makes it so much easier to check changes to build flags and make sure optimisation and compiler options are consistent where they need to be.

### DIFF
--- a/Makefile.envs
+++ b/Makefile.envs
@@ -7,6 +7,7 @@ export PYODIDE_BINARYEN_VERSION ?= version_100
 export BASH_ENV := $(PYODIDE_ROOT)/pyodide_env.sh
 SHELL := /bin/bash
 
+export TOOLSDIR=$(PYODIDE_ROOT)/tools
 export PYVERSION=3.8.2
 export PYMINOR=$(basename $(PYVERSION))
 export HOSTPYTHONROOT=$(shell python3.8 -c "import sys; print(sys.prefix)")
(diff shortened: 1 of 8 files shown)


rth (commented):
Thanks @joemarshall ! I agree that it would be nice to put all those scattered options in one place.

In `pyodide_build/core.py` I think it might be better to define a function that returns the compilation flags as dict, and call it when necessary, instead of calling it at import time and use globals. Maybe we could use a cache there as well.

```py
import functools

@functools.cache
def get_compile_flags() -> Dict[str, str]:
    ...
    return {'DEFAULTLDFLAGS': ... }
```
(or alternatively return a nameduple or a dataclasse)

It would also be good to a test in `pyodide_build/tests/test_common.py` to make sure that the parsing is done correctly.

Makefile.envs
```diff
@@ -23,3 +23,51 @@ export PYODIDE=1
...
+export MAIN_MODULE_CFLAGS= $(CFLAGS_BASE) \
+    	-Wall \
+    	-Wno-warn-absolute-paths \
+    	-Werror=unused-variable \
+    	-Werror=sometimes-uninitialized \
+    	-Werror=int-conversion \
+    	-Werror=incompatible-pointer-types \
+		-I$(PYTHONINCLUDE) 
+
+export PYTHON_CFLAGS=$(CFLAGS_BASE)
```
rth, line 70: Maybe keep this in `cpython/Makefile` similarly to how we keep package specific flags in their respective `meta.yaml` files?

rth (commented):
You would also need to run black on the python files to fix linter CI.

Makefile.envs
joemarshall: I guess so. I think it is unlikely that we'll want different cflags for python anyway. Will do.

joemarshall:
Okay, it's in a cached function now, and there's tests in test_common.py

rth:
Thanks!

PR #1376, opened 2021-03-26
Rework python2js_buffer

### LINKED ISSUE #749: problem with sending bytes to javascript
Is there any restriction on sending bytes from python to javascript?

```javascript
pyodide.runPython('bytes([10])')
> Uint8ClampedArray [10]
pyodide.runPython('bytes([10, 11])')
> Uint8ClampedArray(2) [24, 0]
pyodide.runPython('bytes([10, 11, 12])')
> Uint8ClampedArray(3) [24, 0, 0]
pyodide.runPython('bytes([10, 11, 12, 13])')
> Uint8ClampedArray(4) [10, 11, 12, 13]

pyodide.runPython('b"A"')
> Uint8ClampedArray [65]
pyodide.runPython('b"AB"')
> Uint8ClampedArray(2) [24, 0]
pyodide.runPython('b"ABC"')
> Uint8ClampedArray(3) [24, 0, 0]
pyodide.runPython('b"ABCD"')
> Uint8ClampedArray(4) [65, 66, 67, 68]
```

What's `[24, 0]`?

It's different for different lengths:

```javascript
pyodide.runPython("b'0' * 4")
Uint8ClampedArray(4) [48, 48, 48, 48]
pyodide.runPython("b'0' * 100")
Uint8ClampedArray(100) [48, 48, 48, 48, 48, 48, 48, 48, 48, 48, ...
pyodide.runPython("b'0' * 200")
Uint8ClampedArray(200) [48, 48, 48, 48, 48, 48, 48, 48, 48, 48, ...
pyodide.runPython("b'0' * 300")
Uint8ClampedArray(300) [88, 51, 157, 0, 0, 0, 0, 0, 48, 48, 48, ...
```

It's `[88, 51, 157, 0, 0, 0, 0, 0,` this time.

Seems that there is no problem from javascript to python:

```javascript
f = pyodide.runPython(`
def f(x):
    print([x, x.tobytes()])
    return x
f
`)
> Proxy {$$: {…}, length: 0, name: "target", arguments: null, caller: null, …}
f(new Uint8Array([0]))
> pyodide.asm.js:8 [<memory at 0xa7d308>, b'\x00']
> Uint8Array [0]
f(new Uint8Array([0, 1]))
> pyodide.asm.js:8 [<memory at 0xa7d308>, b'\x00\x01']
> Uint8Array(2) [0, 1]
f(new Uint8Array([0, 1, 2]))
> pyodide.asm.js:8 [<memory at 0xa7d308>, b'\x00\x01\x02']
> Uint8Array(3) [0, 1, 2]
f(new Uint8Array([0, 1, 2, 3]))
> pyodide.asm.js:8 [<memory at 0xa7d308>, b'\x00\x01\x02\x03']
> Uint8Array(4) [0, 1, 2, 3]

```

I tested on this version:

```
pyodide.version(): "0.15.0"
pyodide.runPython('sys.version'): "3.7.4 (default, May 20 2020, 19:25:33) [Clang 6.0.1 ]"
Chrome 83
Firefox 76
```

### LINKED ISSUE #1374: Remove or change toJs behavior for buffers?
If we create a PyProxy of a bytes object and then run the `toJs` method on the `PyProxy`, the resulting `TypedArray` allows the user to modify the `bytes` object even though it's read only and might be shared with a bunch of other places. It also leads to potential use after free errors, and can violate expectations because in every other case `toJs` either copies stuff or returns a `PyProxy` (the other biggest problem with `toJs` is that it creates an unpredictable number of `PyProxy` objects that must be located and freed in order to avoid a leak).

Perhaps we should convert `toJs` to exclusively copy the buffer into a Javascript array, so that `toJs` is slow but safe and returns easy to use buffers whereas `getBuffer` is fast and unsafe but only does the one thing so anyone who reads the docs will see the warnings about modifying readonly buffers.

hoodmane:
As discussed in #1374, this simplifies python2js_buffer and removes various soundness issues (use after free, sometimes shared sometimes copied memory). I also added support for suboffsets. Now in all cases the innermost levels will be TypedArrays, and all the other levels will be normal Arrays.

I rewrote the file `python2js_buffer.c` from scratch so the diff isn't super useful.

Resolves #1374 and resolves #749.


hoodmane:
@rth Okay the trick to the revered endianness is this `newbyteorder` method. The `byteswap` method swaps the bytes without changing the format field, whereas `newbyteorder` method changes the marked endianness but not the data.

rth (commented):
Some comments that I forgot to submit a few days ago. Not sure if they still apply.

src/core/python2js_buffer.c
```diff
@@ -12,487 +13,195 @@
...
+        return (buff) => new Int32Array(swapFunc(buff));
+        return "getInt32";
+      case 'I':
+      case 'L':
+      case 'N':
+        return (buff) => new Uint32Array(swapFunc(buff));
+      case 'q':
+        return (buff) => new BigInt64Array(swapFunc(buff));
+      case 'Q':
+        return (buff) => new BigUint64Array(swapFunc(buff));
```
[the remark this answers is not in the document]
hoodmane: Well it will already throw a `ReferenceError` if `BigInt64Array` is undefined =P

src/core/python2js_buffer.js
```diff
@@ -0,0 +1,214 @@
...
+      if (suboffset >= 0) {
+        curptr = HEAP32[curptr / 4] + suboffset;
+      }
+      result.push(
+          Module._python2js_buffer_recursive(curPtr, curdim + 1, bufferData));
+    }
+    return result;
+  };
+
+  Module.get_converter = function(format, itemsize) {
```
rth, line 160: Could you please add a docstring here and for all top level functions in this file?

src/core/python2js_buffer.js
```diff
@@ -0,0 +1,214 @@
...
+        curptr = HEAP32[curptr / 4] + suboffset;
+      }
+      buffer.set(HEAP8.subarray(curptr, curptr + itemsize), i * itemsize);
+    }
+    return converter(buffer.buffer);
+  };
+
+  Module._python2js_buffer_recursive = function(ptr, curdim, bufferData) {
+    "use strict";
+    let n = HEAP32[bufferData.shape / 4 + curdim];
```
rth, line 129: Please add an inline comment why we need the "/4" (for adaptation from HEAP8 to HEAP32 I imagine?)

rth (approved):
It's a nice code refactoring and simplification!

> Now in all cases the innermost levels will be TypedArrays, and all the other levels will be normal Arrays.

Please add a changelog entry

hoodmane:
@rth Here there are a couple of tests failing with `NameError: name 'run_in_pyodide' is not defined`. Any idea if that's related to your recent improvements to the test suite or what?
https://app.circleci.com/pipelines/github/hoodmane/pyodide/111/workflows/450447e8-b1cf-4caf-86f1-4c9900c5e6c1/jobs/884

hoodmane:
> @rth Here there are a couple of tests failing with `NameError: name 'run_in_pyodide' is not defined`.

Oh I see this is what #1444 was about.

PR #1377, opened 2021-03-26
Support PyProxy get and set with a list as index for ndarrays

hoodmane:


docs/project/changelog.md
```diff
@@ -56,6 +56,8 @@ substitutions:
 - {{ API }} Added `PyProxy.getBuffer` API to allow direct access to Python
   buffers as Javascript TypedArrays.
   [1215](https://github.com/iodide-project/pyodide/pull/1215)
+- {{ API }} Added support for multiindexes with `PyProxy.get`, `PyProxy.set`, etc.
```
rth, line 59: multiindexing can have a bit different meaning in scientific python https://pandas.pydata.org/pandas-docs/stable/user_guide/advanced.html 

```suggestion
- {{ API }} Added support indexing with tuples using `PyProxy.get`, `PyProxy.set`, etc.
  For instance, `X.get(`
```

rth (commented):
Note that for numpy (and other packages) indexing by tuples or list is not equivalent,
```py
>>> x = np.arange(12).reshape(3, 4)
>>> x[[0, 0]]
array([[0, 1, 2, 3],
       [0, 1, 2, 3]])
>>> x[(0, 0)]
0
>>> x[0, 0]
0
```
so `pyodide.globals.get("x").get([0, 0])` should produce a slice of first row twice, while `pyodide.globals.get("x").get((0, 0))` should produce the first element I think.

`x[(a, b)]` is equivalent to `x[a, b]` but that's not the case with lists.

cf https://numpy.org/doc/stable/reference/arrays.indexing.html

hoodmane:
> pyodide.globals.get("x").get((0, 0))

In Javascript, if you say `f((0,0))` then `f` actually receives a single `0` as an argument so that isn't available as a syntax, but it'd be possible to use `get(0,0)`. Does that sound reasonable to you to convert `x.get(0,0)` to `x[0,0]` and `x.get([0,0])` to x[[0,0]]`?

It might be better not to do anything though, since we have a pretty limited ability to emulate Python syntax.

rth:
> Does that sound reasonable to you to convert x.get(0,0) to x[0,0] and x.get([0,0]) to x[[0,0]]`?

Yes, sounds great.

PR #1378, opened 2021-03-27
Implement slicing for JsProxy of Array

hoodmane:
Adds slice get, slice assign, slice delete for JsProxy of an array


hoodmane:
It seems like it would be easier and better to just implement `sq_item` and `sq_ass_item` instead of `mp_subscript` and `mp_ass_subscript`.

PR #1385, opened 2021-03-28
MAINT Improve error handling

hoodmane:
I added a separate reserved hiwire value `Js_novalue` to use as a `None` result when the return value would be `Option<JsRef>` (as distinct from `NULL` which indicates an error). 

Mostly it's bad to say:
```py
try:
    # something
except: # catch anything!
    # handling code
```
and uses of `PyErr_Clear()` that aren't guarded by a conditional that checks if `PyErr_ExceptionMatches` a certain type of error are equivalent to this. We mostly don't do this already, but I fixed a few cases where we did. Also, I moved the buffer conversion behind a check for permission with `PyObject_CheckBuffer` so that we can throw an error when the buffer conversion fails rather than falling back to returning a `PyProxy`. This should make it easier to discover programming errors in our buffer conversion code.

### DIFF
--- a/src/core/hiwire.c
+++ b/src/core/hiwire.c
@@ -14,6 +14,9 @@ const JsRef Js_true = ((JsRef)(4));
 const JsRef Js_false = ((JsRef)(6));
 const JsRef Js_null = ((JsRef)(8));
 
+// For when the return value would be Option<JsRef>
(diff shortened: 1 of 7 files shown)


hoodmane:
This has the now-typical set of test failures but no more. I'm going to merge it.

PR #1387, opened 2021-03-29
MAINT Fix test failures due to negative refcounts by using selenium_standalone

hoodmane:


rth (commented):
OK, all core tests are good except for `string_conversion`. Could you please increase [the deadline setting](https://hypothesis.readthedocs.io/en/latest/settings.html#hypothesis.settings.deadline) (say to `deadline=600`) to make it pass?

rth (approved):
Thanks! The other failures are still there but these tests were indeed fixed.

PR #1391, opened 2021-03-30
MNT Minor cleanup after #1385

hoodmane:
In one spot we should return `Js_novalue` rather than `NULL`.

### DIFF
--- a/src/core/python2js.c
+++ b/src/core/python2js.c
@@ -260,13 +260,6 @@ _python2js_immutable(PyObject* x)
     return _python2js_long(x);
   } else if (PyFloat_Check(x)) {
     return _python2js_float(x);
-  } else if (PyNumber_Check(x)) {
-    JsRef result = _python2js_float(x);
-    if (result == NULL) {
-      PyErr_Clear();
-      return Js_novalue;
-    }
-    return result;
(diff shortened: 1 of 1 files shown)


hoodmane:
It's a bit surprising to me that the test suite didn't catch this, though it may have been masked by other problems.

src/core/python2js.c
```diff
@@ -260,6 +260,13 @@ _python2js_immutable(PyObject* x)
     return _python2js_long(x);
   } else if (PyFloat_Check(x)) {
     return _python2js_float(x);
+  } else if (PyNumber_Check(x)) {
```
rth, line 263: I think we need a regression test for at least one type included in `PyNumber_Check` https://github.com/python/cpython/blob/f0a6fde8827d5d4f7a1c741ab1a8b206b66ffd57/Objects/abstract.c#L756 but not selected in earlier `else if` branches.

hoodmane: I think the right approach is to add a test with a class with an `__int__` method that returns a constant and one that throws an error.

hoodmane: Actually I think this change is misguided so I'll just remove it.

PR #1398, opened 2021-03-31
Apply lints suggested by lgtm.com

hoodmane:
It seems like lgtm won't be able to lint the C files without some effort (it currently crashes on `Could not find PyYAML.  Please install that before continuing.`).

### DIFF
--- a/src/core/pyproxy.js
+++ b/src/core/pyproxy.js
@@ -716,7 +716,7 @@ TEMP_EMJS_HELPER(() => {0, /* Magic, see comment */
      *    "f32", or "f64,
      * @returns PyBuffer
      */
-    getBuffer : function(type = "u8") {
+    getBuffer : function(type) {
       let ArrayType = undefined;
       if (type) {
         ArrayType = type_to_array_map.get(type);
--- a/src/core/pyproxy.js
+++ b/src/core/pyproxy.js
@@ -719,7 +719,7 @@ TEMP_EMJS_HELPER(() => {0, /* Magic, see comment */
     getBuffer : function(type = "u8") {
       let ArrayType = undefined;
       if (type) {
-        let ArrayType = type_to_array_map.get(type);
+        ArrayType = type_to_array_map.get(type);
         if (ArrayType === undefined) {
           throw new Error(`Unknown type ${type}`);
         }
--- a/src/core/pyproxy.js
+++ b/src/core/pyproxy.js
@@ -716,10 +716,10 @@ TEMP_EMJS_HELPER(() => {0, /* Magic, see comment */
      *    "f32", or "f64,
      * @returns PyBuffer
      */
-    getBuffer : function(type) {
+    getBuffer : function(type = "u8") {
       let ArrayType = undefined;
       if (type) {
-        ArrayType = type_to_array_map.get(type);
+        let ArrayType = type_to_array_map.get(type);
         if (ArrayType === undefined) {
           throw new Error(`Unknown type ${type}`);
         }
--- a/conftest.py
+++ b/conftest.py
@@ -409,8 +409,6 @@ def run_web_server(q, log_filepath, build_dir):
     sys.stdout = log_fh
     sys.stderr = log_fh
 
-    test_prefix = "/src/tests/"
-
     class Handler(http.server.SimpleHTTPRequestHandler):
         def log_message(self, format_, *args):
             print(
--- a/lgtm.yml
+++ b/lgtm.yml
@@ -1,0 +1,8 @@
+queries:
+  - exclude: py/import-and-import-from
+
+extraction:
+  cpp:
+    prepare:
+      packages:
+        - python3-yaml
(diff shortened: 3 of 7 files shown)


src/core/pyproxy.js
```diff
@@ -719,7 +719,7 @@ TEMP_EMJS_HELPER(() => {0, /* Magic, see comment */
     getBuffer : function(type = "u8") {
       let ArrayType = undefined;
       if (type) {
-        let ArrayType = type_to_array_map.get(type);
+        ArrayType = type_to_array_map.get(type);
```
hoodmane, line 722: This is the one major bug fixed in this PR.

rth:
Nice! Should we try to enable it on this repo?

>  (it currently crashes on Could not find PyYAML. Please install that before continuing.).

I guess we could always disable C analysis for the time being via a config setting.

src/core/pyproxy.js
hoodmane: Probably we should add some tests of `getBuffer` that actually use the `type` argument. Actually this will probably break the tests because there are a pair of bugs here: 
1. I meant for `type` to default to `undefined` but 
2. the type argument is always ignored.

rth:
LGTM should be installed now. Could you try pushing another commit?

hoodmane:
What's going on with the CI here?

hoodmane:
I think you can install the `lgtm` github app here:
https://github.com/apps/lgtm-com/installations/new

rth:
Latest analysis in https://lgtm.com/projects/g/pyodide/pyodide/?mode=list

> Module 'asyncio' is imported with both 'import' and 'import from'

That's a not useful error from LGTM that we'll need to find a way to ignore ideally.

So the C setup fails with,
```
[2021-03-31 16:11:48] [build-stdout] [2021-03-31 16:11:48] [build-stderr] Could not find PyYAML.  Please install that before continuing.
[2021-03-31 16:11:48] [build-stdout] [2021-03-31 16:11:48] [build-stderr] make: *** [Makefile:156: check] Error 1
[2021-03-31 16:11:48] [build-stdout] [2021-03-31 16:11:48] [build-stderr] + '[' -f build.ninja ']'
[2021-03-31 16:11:48] [build-stdout] [2021-03-31 16:11:48] [build-stderr] + '[' -d ../_lgtm_build_dir ']'
[2021-03-31 16:11:48] [build-stdout] [2021-03-31 16:11:48] [build-stderr] + for f in build build.sh
[2021-03-31 16:11:48] [build-stdout] [2021-03-31 16:11:48] [build-stdout] Semmle autobuild: no supported build system detected.
[2021-03-31 16:11:48] [build-stdout] [2021-03-31 16:11:48] [build-stderr] + '[' -x build ']'
[2021-03-31 16:11:48] [build-stdout] [2021-03-31 16:11:48] [build-stderr] + for f in build build.sh
[2021-03-31 16:11:48] [build-stdout] [2021-03-31 16:11:48] [build-stderr] + '[' -x build.sh ']'
[2021-03-31 16:11:48] [build-stdout] [2021-03-31 16:11:48] [build-stderr] + '[' -f setup.py ']'
[2021-03-31 16:11:48] [build-stdout] [2021-03-31 16:11:48] [build-stderr] + echo 'Semmle autobuild: no supported build system detected.'
[2021-03-31 16:11:48] [build-stdout] [2021-03-31 16:11:48] [build-stderr] + exit 1
```
if it's trying to compile everything with make, good luck to it. Related https://blog.semmle.com/cplusplus-projects-lgtm/

We could try to fix the pyyaml issue with something like, the following in `lgtm.yml` I imagine
```yml
extraction:
  cpp:
    prepare:
      packages:
        - python3-yaml
```
https://lgtm.com/help/lgtm/lgtm.yml-configuration-file

hoodmane:
> > Module 'asyncio' is imported with both 'import' and 'import from'
> That's a not useful error from LGTM that we'll need to find a way to ignore ideally.

I think it can be silenced with:
```yml
queries:
  - exclude: py/import-and-import-from
```
in lgtm.yml

src/core/pyproxy.js
rth: Might be good to in a different PR

rth (approved):
Thanks, LGTM! I also disabled comments in their UI, as a comment per PR does add to the notification noise. That information is in the CI status anyway.

rth:
> What's going on with the CI here?

The unfortunate side effect of being more limited in concurrent CircleCI jobs is that if all available runners (4 currently) are used in other PRs, the PR will have  green CI status (or no CI) instead of pending CI. Which means we have to check how many CI jobs were actually run before merging.

hoodmane:
Are the Mozilla people planning to give the new Pyodide org access to the same custom CI plan?

rth:
Unclear for now, we'll probably know more by the 0.17 release. If not we'll need to look into other possible workarounds.

Issue #1389, opened 2021-03-29
CI issues after GH org migration


rth:
For a discussion of CI related issues following the migration to the pyodide GH org, previously discussed in https://github.com/pyodide/pyodide/pull/1383 and in particular https://github.com/pyodide/pyodide/pull/1383#issuecomment-809187564

rth:
Right now we only have 4 concurent jobs which leads to PRs pending CI with more than a sumultaneously updated branches/PRs.

dalcde:
If contributors set up CircleCI on their own branches then it could help with the queuing

hoodmane:
Is it possible you could use this page:
https://app.circleci.com/settings/plan/github/pyodide/share
to share or transfer the iodide-project plan with pyodide?

rth:
No I don't have permissions to access that page on the previous org.

Anyway, I suspect iodide was on a paid plan, and Mozilla ends up receiving invoices for it. So we need explicit approval.

I also think queuing will get better once we fix CI issues, which at least in webworkers were symptomatic of actual issues/deadlocks as far as I can tell. Currently CI takes long, in part due to waiting for tests that time out after 30s. https://github.com/pyodide/pyodide/pull/1404 should help there.

But setting it up for our forks can indeed also help. Currently we are burning through CI credits fast,
![image](https://user-images.githubusercontent.com/630936/113346146-c30daa80-9333-11eb-8a9b-fa34361820f7.png)
(with a total of 400k credits per month). I don't even understand how we could have spent 32 CPU hours for CI in 1 day. Probably timeouts and retrying build which take forever without ccache on a new fork.

hoodmane:
I set it up on my fork earlier today but I've already used "10,186 of 2,500 credits". It seems to be running anyways, I don't know what their policy is when you go over.

rth:
You normally get [400k per month](https://circleci.com/open-source/) that counter is not accurate. Though I'm not sure if they sum up credits for forks somehow..

rth:
To give an update on this, unfortunately we won't be able to get Mozilla support for CI. 

In the near term, I think it should be manageable with the 3 of us running Circle CI on forks. Currently for latest commit on master, there were still 2 test failures for Firefox packages and 4 test failures for Chrome packages. And Chrome core still occasionally fails to initialize selenium. Generally though the situation improved since this issue opened.

[...]

PR #1404, opened 2021-03-31
MAINT Faster and more reliable WebWorker tests

rth:
Refactors webworker test mechanism to hopefully be faster and more reliable (related to #1389 ).

 - previously in webworker tests Pyodide was loaded both on the main page and in the webworker. Now it is only loaded in the webworker. This results in approximately 25% speed up for `tests/test_webworker.py` for Chrome.
 - previously if there was an error while loading pyodide in the webworker, this lead to a deadlock, as far as I can tell, and we timed out after 30 sec. Thanks to https://github.com/pyodide/pyodide/pull/1363 we can now explicitly call `loadPyodide` when the first message is received and handle errors appropriately.
 - added `selenium_webworker_standalone` that is identical to `selenium_standalone` but doesn't call `loadPyodide` on the test.html page.
 - I also reduced the timeout from 30s to 20s for `driver.execute_async_script` so that CI takes less time when there are timeouts.

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -102,10 +102,10 @@ class SeleniumWrapper:
                 f"{(build_dir / 'test.html').resolve()} " f"does not exist!"
             )
         self.driver.get(f"http://{server_hostname}:{server_port}/test.html")
+        self.run_js("Error.stackTraceLimit = Infinity;", pyodide_checks=False)
         if load_pyodide:
             self.run_js("await loadPyodide({ indexURL : './'});")
             self.save_state()
-        self.run_js("Error.stackTraceLimit = Infinity;", pyodide_checks=False)
         self.driver.set_script_timeout(20)
 
     @property
--- a/conftest.py
+++ b/conftest.py
@@ -81,7 +81,12 @@ class SeleniumWrapper:
     JavascriptException = JavascriptException
 
     def __init__(
-        self, server_port, server_hostname="127.0.0.1", server_log=None, build_dir=None
+        self,
+        server_port,
+        server_hostname="127.0.0.1",
+        server_log=None,
+        build_dir=None,
+        load_pyodide=True,
     ):
         if build_dir is None:
             build_dir = BUILD_PATH
@@ -97,9 +102,11 @@ class SeleniumWrapper:
                 f"{(build_dir / 'test.html').resolve()} " f"does not exist!"
             )
         self.driver.get(f"http://{server_hostname}:{server_port}/test.html")
-        self.run_js("Error.stackTraceLimit = Infinity;")
-        self.run_js("await loadPyodide({ indexURL : './'});")
-        self.save_state()
+        if load_pyodide:
+            self.run_js("await loadPyodide({ packageIndexURL : './'});")
+            self.save_state()
+        self.run_js("Error.stackTraceLimit = Infinity;", pyodide_checks=False)
+        self.driver.set_script_timeout(20)
 
     @property
     def logs(self):
@@ -144,17 +151,14 @@ class SeleniumWrapper:
             """
         )
 
-    def run_js(self, code):
+    def run_js(self, code, pyodide_checks=True):
+        """Run JavaScript code and check for pyodide errors"""
         if isinstance(code, str) and code.startswith("\n"):
             # we have a multiline string, fix indentation
             code = textwrap.dedent(code)
 
-        wrapper = """
-            let cb = arguments[arguments.length - 1];
-            let run = async () => { %s }
-            (async () => {
-                try {
-                    let result = await run();
+        if pyodide_checks:
+            check_code = """
                     if(globalThis.pyodide && pyodide._module && pyodide._module._PyErr_Occurred()){
                         try {
                             pyodide._module._pythonexc2js();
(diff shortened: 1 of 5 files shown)


conftest.py
```diff
@@ -165,14 +169,25 @@ def run_js(self, code):
                             throw new Error(`Python exited with error flag set!`);
                         }
                     }
+           """
+        else:
+            check_code = ""
```
rth, line 174: It's true that the above code snippet is already not run if pyodide is not loaded. However I still find it useful to be able to run some JS code with all interaction with pyodide disabled.

rth:
@hoodmane Any ideas what could lead to,
```
JavascriptException: _createPyodideModule is not defined
```
? I don't see how present changes could have an impact on it anyway. Of course locally this PR works just fine for me.  The lack of build/CI reproducibility is getting tiring.

conftest.py
hoodmane: Yeah I agree that it is useful. For instance I was running into some issues with this stuff when testing the fatal errors because after the fatal error nothing works. We might consider adding a `mark` to disable all checks.

src/webworker.js
```diff
@@ -1,17 +1,19 @@
...
+      }
     }
-  }
 
-  try {
+    if (typeof self.pyodide !== "undefined") {
+      await loadPyodide({packageIndexURL : '{{ PYODIDE_BASE_URL }}'});
+    }
     self.postMessage(
         {results : await self.pyodide.runPythonAsync(data.python)});
```
hoodmane, line 18: This looks weird to me. If `self.pyodide === "undefined"` then you skip `loadPyodide` but then `self.pyodide.runPythonAsync` is definitely going to fail too. Maybe we should just throw an error if it's undefined? When would we expect to run into that condition?

conftest.py
```diff
@@ -196,8 +211,7 @@ def run_webworker(self, code):
         return self.run_js(
             """
             let worker = new Worker( '{}' );
```
hoodmane, line 213: Would we perhaps want to make `selenium_webworker_standalone` start a webworker and then have all `run_webworker` uses run in the same webworker? It's a little surprising to me that each call to `run_webworker` creates it's own `Worker` instance.

src/webworker.js
rth: Indeed, that was meant to be a `===` not `!==`.

The idea was to initialize pyodide on first recieved message, and avoid initializing it on the following messages, in order to move in the direction of your other comment in https://github.com/pyodide/pyodide/pull/1404#discussion_r605753753 

The fact that it didn't fail for me locally, means I was likely not testing this actual PR. I'll try again, maybe I'll now be able to reproduce other failures in this PR.

conftest.py
rth: Yep, I also found the current behavior surprising and +1 to move in that direction.

Though for now, I'm only looking into fixing CI and this refactoring can be done separately. Currently we are always calling `run_webworker` only once per test, so it should be fine.

src/webworker.js
hoodmane: Oh right, that makes a lot more sense!

conftest.py
hoodmane: > Any ideas what could lead to, `JavascriptException: _createPyodideModule is not defined`

No clue! Maybe a problem loading `pyodide.asm.js` which is where that is defined...

hoodmane:
> Any ideas what could lead to, `JavascriptException: _createPyodideModule is not defined`

No clue! Maybe a problem loading `pyodide.asm.js` which is where that is defined...

rth:
> No clue! Maybe a problem loading pyodide.asm.js which is where that is defined...

For the record the issues was a few missing `packageIndexURL -> indexURL` replacement as I started with an earlier version of your PR. As a result indexURL was then not defined, fell back to `PYODIDE_BASE_URL` where,
 - locally the default is "./" so it was still working
 - in CircleCI it defaults to dev version on JsDelivr which didn't have `_createPyodideModule`

hence the error. So everything makes sense now.

src/webworker.js
```diff
@@ -1,22 +1,24 @@
...
-        {results : await self.pyodide.runPythonAsync(data.python)});
+    if (typeof self.__pyodideLoading === "undefined") {
+      await loadPyodide({indexURL : '{{ PYODIDE_BASE_URL }}'});
+    }
+    let res = await self.pyodide.runPythonAsync(data.python);
+    self.postMessage({results : res});
   } catch (e) {
     // if you prefer messages with the error
-    self.postMessage({error : e.message});
+    self.postMessage({error : e.message + '\n' + e.stack});
```
rth, line 21: I'm not sure what's best to do here. `e.message` by itself is not enough as it's hard to debug without the stack trace.

Now for a good fraction of Python exceptions, the `e.stack` also appears to contain the message, meaning that it would be included twice with the current version.
However if I remove `e.message` I'm concerned that we then won't get the actual error message for JS exceptions.

So I guess repeating the message twice is better then potentially missing it. I'm not if there is a better way of managing it.

rth:
So it doesn't  appear fix all webworker issues. There were still a few timeouts the first time core webworker tests run on Firefox. Upon re-running they passed making core firefox CI green (one commit before last).

On Chrome, webworker tests were already passing and errors are unrelated.

Overall it's some improvement. Let me know if you have other comments @hoodmane

src/webworker.js
hoodmane: > repeating the message twice is better then potentially missing it

I think repeating the message is fine and normal.

hoodmane:
Looks good to me! 

As we said above it'd be great in the future to share most webworker tests between a single worker. Dunno if that would improve stability at all, but it'd probably help speed up the tests.

PR #1407, opened 2021-04-01
Handling for BigInt, more type conversion tests

hoodmane:
I added more hypothesis tests for round trip conversions of Python objects from Python to Javascript back to Python. In particular, I told `hypothesis` to generate examples of absolutely anything subject to the constraints that
1. it is pickleable, and
2. it is == to itself.

The only counterexamples it found were integers larger than `MAX_SAFE_INTEGER`. I added BigInt support to the conversions to fix that case. I also added a few other number related tests.


rth (commented):
It might be good to have some fallback if BigInt is not available to avoid breaking things for Safari v13 users: https://caniuse.com/bigint

src/core/hiwire.c
```diff
@@ -274,6 +280,22 @@ EM_JS_REF(JsRef, hiwire_int, (int val), {
   return Module.hiwire.new_value(val);
 });
 
+EM_JS_REF(JsRef, hiwire_int_from_hex, (const char* s), {
+  let result;
+  // clang-format off
+  if (HEAP8[s] === 45) {
```
rth, line 286: Could you add a comment why 45?

src/tests/test_typeconversions.py
```diff
@@ -26,6 +26,182 @@ def test_string_conversion(selenium_module_scope, s):
...
+@given(obj=from_type(type).flatmap(from_type))
+@settings(deadline=600)
+def test_hyp_py2js2py(selenium_module_scope, obj):
+    with selenium_context_manager(selenium_module_scope) as selenium:
+        import pickle
+
+        try:
+            assume(obj == obj)  # make sure no NaNs.
+        except:
+            assume(False)
```
rth, line 131: That's not equivalent to
```
assume(not isnan(x))
```
from the example in https://hypothesis.readthedocs.io/en/latest/details.html#hypothesis.assume, or I guess it fails on arbitrary objects ?

docs/usage/type-conversions.md
```diff
@@ -65,26 +67,33 @@ the same type as the original object, we proxy `tuple` and `bytes` objects.
...
+| Python          | Javascript            |
+|-----------------|-----------------------|
+| `int`           | `Number` or `BigInt`* |
+| `float`         | `Number`              |
+| `str`           | `String`              |
+| `bool`          | `Boolean`             |
+| `None`          | `undefined`           |
+
+* An `int` is converted to a `Number` if the `int` is between -2^{53} and 2^{53}
+  inclusive, otherwise it is converted to a `BigInt`.
```
rth, line 79: Maybe add,
"if the browser supports BigInt" and indicate what would happen otherwise.

rth (commented):
Minor comments otherwise LGTM. Thanks!

src/core/hiwire.c
rth: My question was more how would one independently determine that 45 corresponds to the - sign.

src/tests/test_typeconversions.py
hoodmane: I think I tried `assume(not isnan(x))` first and it didn't work, but I can't remember. Equality with `nan` is really weird:
```py
from math import nan
a = [nan]
b = [float('nan')]
a == a # true
b == b # true
a == [nan] # true
a == b # false
b == [nan] # false
b == [float('nan')] # false
```

docs/usage/type-conversions.md
hoodmane: Currently it converts it to a number. Maybe we should throw an error though in interest of the principle of least surprise.

hoodmane: Since only 5% of browsers don't support bigint it's probably not worth worrying this case too much.

src/tests/test_typeconversions.py
hoodmane: Oh well the current version also supports the case where `x == x` raises an error, which `isnan` I think won't handle.

hoodmane: In summary there are three options for `x==x`:
1. True
2. False
3. raises an exception
In cases 2 & 3, we don't want to run the test.

rth: Sounds good!

PR #1413, opened 2021-04-01
Another problem in console.html created in #1363

hoodmane:
initializePyodide ==> loadPyodide

We really ought to have some sort of regression test for console.html.

### DIFF
--- a/src/templates/console.html
+++ b/src/templates/console.html
@@ -14,7 +14,7 @@
   <body>
     <script>
       async function main() {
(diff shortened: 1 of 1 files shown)


rth:
Actually, can't we just load console.html in selenium using
https://github.com/pyodide/pyodide/blob/455f302d7dc174af292f97713f601c1db05a2004/conftest.py#L99
? At least to check that it loads without JS errors.

hoodmane:
Yeah that would have caught both of these mistakes.

PR #1414, opened 2021-04-01
Fix matplotlib backend

hoodmane:
In #1234 I broke the matplotlib backend: when Matplotlib was introduced in commit https://github.com/pyodide/pyodide/commit/918ccc7300ed9a092b50c4087a7d1950414798bc the type conversions code was altered so that a `bytes` object would be converted to a `Uint8ClampedArray`. This persisted until I removed that conversion in #1234. The reason to convert it to a `Uint8ClampedArray`, rather than a `Uint8Array` which would seem more normal, was that the `ImageData` constructor demands that its argument be a `Uint8ClampedArray` (this constraint is a bit silly IMO).

Anyways my recent changes to buffer type conversions make it possible, if a bit clumsy, to manually ensure that the bytes object gets converted to a `Uint8ClampedArray`.

Most of the changes here come from ~#1376~ #1322 and #1399. The only changes that are specific to this PR are the ones to `wasm_backend.py`.


rth:
Thanks! We'll have to merge https://github.com/pyodide/pyodide/pull/1376 first then.

hoodmane:
I'm sorry, this actually needs #1322 not #1376.

rth (approved):
The diff is clearly smaller now :)

LGTM, assuming CI passes.

rth:
Hmm, matplotlib Chrome tests fail with `RuntimeError: memory access out of bounds` I imagine related to https://github.com/pyodide/pyodide/issues/1473 Should we merge despite it, as it's likely unrelated to this PR?

hoodmane:
Ideally it'd be good if we had a regression test for this, but I don't feel like making one right now so I'm going to merge this.

hoodmane:
> Hmm, matplotlib Chrome tests fail with RuntimeError: memory access out of bounds I imagine related to #1473 Should we merge despite it, as it's likely unrelated to this PR?

Oops only just saw this. Yeah it's not related. This one occurs in `PyObject_GenericGetAttr` which is in main Python so I think that rules out my ideas on #1473 about there being a relationship between the fatal errors and extension modules.

Issue #1380, opened 2021-03-27
Update the REPL URL to not use iodide.io


rth:
Currently the REPL is accessible at https://pyodide-cdn2.iodide.io/v0.17.0a2/full/console.html

As part of #1217 , we should also move away from the iodide.io domain, as that  adds to the confusion. cc @teonbrooks 

Fist we need to decide what subdomain to host it on. Maybe `repl.pyodide.org/v0.17.0a2/full/console.html` or possibly something more generic as `packages.pyodide.org` since other HTML files could be served from there in the future https://github.com/iodide-project/pyodide/issues/977 ? 

Second we need to decide how to technically implement it.
 1. We can upload those files separately to Github pages or Netlify. However this has to be done for each for each version and kept in sync with the packages deployed to S3/JsDelivr.
 2. Alternatively we could find a way to serve those files from S3 since they are already there (which was done for pyodide-cdn2.iodide.io). However the challenge there that we need to only serve .html files and not all the other assets (which should go through JsDelivr exclusively). I haven't seen of a simple solution to do that yet. I guess it could be done with some AWS Lambda functions but that would require some work/maintenance effort.

So overall maybe we should go with GH pages.  Any suggestions for the domain name?

teonbrooks:
`repl.pyodide.org` sounds cool. it would be nice if its index.html resolves to the latest version, and that you can access previous builds with the more verbose url. just my 2c

teonbrooks:
for me, i would think that if I were to go to `packages.pyodide.org` i would be getting an index page of all the available packages

hoodmane:
I agree that repl.pyodide.org should point to the latest stable repl.

rth:
In the end in https://github.com/pyodide/pyodide/pull/1445 I went with deploying console.html as part of the docs, as it already has versioned URLs and we already have a deployment mechanism there, so it was easiest to setup.

If that's merged, after the release the console would be available as https://pyodide.org/en/stable/console.html

PR #1420, opened 2021-04-02
DOCS Update readme, index and about page, edits to usage docs

hoodmane:
The main change here is that I copied the first two paragraphs of the readme into index.rst replacing the title page language in index.rst. The language in the readme is much tighter and better edited so I think this is an improvement. I like "Python **and** the scientific stack" very slightly better than "Python **with** the scientific stack".

I also made various improvements to the text and line wrapped the readme.


README.md
```diff
@@ -8,38 +8,44 @@
 [![Build Status](https://circleci.com/gh/pyodide/pyodide.png)](https://circleci.com/gh/pyodide/pyodide)
 [![Documentation Status](https://readthedocs.org/projects/pyodide/badge/?version=latest)](https://pyodide.readthedocs.io/?badge=latest)
 
-Python with the scientific stack, compiled to WebAssembly.
+Python and the scientific stack, compiled to WebAssembly.
 
 ## What is Pyodide?
 
-**Pyodide** brings the Python 3.8 runtime to the browser via WebAssembly, along with the Python scientific stack including NumPy, Pandas, Matplotlib, SciPy, and scikit-learn. The [`packages` directory](https://github.com/pyodide/pyodide/tree/master/packages) lists over 75 packages which are currently available. In addition it's possible to install pure Python wheels from PyPi.
+**Pyodide** brings the Python 3.8 runtime to the browser via WebAssembly, along
```
hoodmane, line 15: I find the bold text here a bit distracting. What do you think about removing it?

README.md
```diff
@@ -8,38 +8,44 @@
 [![Build Status](https://circleci.com/gh/pyodide/pyodide.png)](https://circleci.com/gh/pyodide/pyodide)
 [![Documentation Status](https://readthedocs.org/projects/pyodide/badge/?version=latest)](https://pyodide.readthedocs.io/?badge=latest)
 
-Python with the scientific stack, compiled to WebAssembly.
+Python and the scientific stack, compiled to WebAssembly.
```
rth, line 11: The issue with "and" is that then less clear that it's Python scientific stack we are talking about not, the scientific stack in general.

README.md
rth: Fine with me.

rth (commented):
Thanks! LGTM, aside for the "with" -> "and" change in the title.

hoodmane:
@rth What do you think about my changes to the README? 

I think the old text was a bit odd because it defined Pyodide in contrast to Iodide. That would make sense if we thought people were coming here from Iodide or were going to be confused between them or something. I moved the "Pyodide may be used in any context where you want to run Python inside a web browser." Also, if we are going to mention Iodide it seems like we say that Iodide is no longer maintained and suggest an alternative so I added a link to Starboard.

hoodmane:
Also, the about page has very similar content:
https://pyodide--1420.org.readthedocs.build/en/1420/project/about.html
What about merging the about page into the README and just including a link to the README there?

hoodmane:
I guess I'll just try something and see.

rth:
> What about merging the about page into the README and just including a link to the README there?

They are currently similar but I was thinking we could add more thing in the future to about such as a list of current maintainers, other sponsors of there are any etc see example in https://scikit-learn.org/dev/about.html

I agree there is some overlap, but for me the expected audience is a bit different:
 - Readme: new or potential users -> mostly interested in what the project does, how to use it + links to documentation for additional information. They will probably spend seconds before deciding if they are interested in project or not, so anything not related to new users particularly at the top should be avoided.
 - About: for people interested in more detail how the project is and how it is organized.  

so I would rather we kept them separate.

> I think the old text was a bit odd because it defined Pyodide in contrast to Iodide.

+1 to change it.

README.md
```diff
@@ -1,6 +1,6 @@
 <div align="center">
   <a href="https://github.com/pyodide/pyodide">
-  <img src="./docs/_static/img/pyodide-logo-readme.png" alt="Pyodide">
+  <img src="https://github.com/pyodide/pyodide/raw/master/docs/_static/img/pyodide-logo-readme.png" alt="Pyodide">
```
rth, line 3: This is to include it into the sphinx docs as it? It's not great to use Github raw as a CDN and also it would break readme for released tags/branches if that path changed for any reason on master in the future.

README.md
```diff
@@ -12,48 +12,73 @@ Python with the scientific stack, compiled to WebAssembly.
...
+## History
+Pyodide was created in 2018 by [Michael Droettboom](https://github.com/mdboom)
+at Mozilla as part of the [iodide
+project](https://github.com/iodide-project/iodide). Iodide is an experimental
+web-based notebook environment for literate scientific computing and
+communication.
+
+Iodide is no longer maintained. To use Pyodide in an interactive client-side
+notebook, you may also be interested in [Starboard
+Notebook](https://starboard.gg/).
```
rth, line 62: Maybe rather link to https://pyodide.org/en/latest/project/related-projects.html#notebook-environements-ides-repls to be more neutral and leave the choice  up to user? From what I understand there might be other such systems in development so that list could get longer in the future.

README.md
hoodmane: I think I'm going to get rid of this anyways to put back the separate about page. Unless you think we should include both the `about.md` and `README.md` into the docs?

rth: +1 Separate about.md and README.md is easier IMO even if it means keeping some semi-duplicate text in sync.

hoodmane: I am going to keep the history and acknowledgements in the README too because I think they fit well in there. So for now about.md is going to just be a subset of the info in the readme. Hopefully we'll fill it out more in the future as you suggested above:

> we could add more thing in the future to about such as a list of current maintainers, other sponsors of there are any etc

rth: +1 for keeping history there, particularly given the possible confusion pyodide/iodide, but maybe we could put it after contributing and  before license?

I would rather we removed acknowledgements for the readme though,
 - I put RDT there to be fair, but technically they are already running paid ads on the docs to make it sustainable. As of now we still don't have Mozilla CI support.
 - it would create the expectation that other potential sponsors would also be listed the readme, which I think is not ideal.

hoodmane: > I would rather we removed acknowledgements for the readme 

Sounds good. Ideally it would be nice to have an acknowledgments section in the README that acknowledged specific *people* rather than sponsors. But given that we are acknowledging companies I see your point.

We got an acknowledgement of sorts for @mdboom into the README by adding the history section which calls him out as the creator, that seems like the most important one anyways. You would also deserve an acknowledgment but as long as you are the main maintainer it'd be weird. We should probably add a maintainers list, like with in a paper the authors don't acknowledge themselves but they are listed as authors...

hoodmane:
Okay, let me know how it looks to you now.

README.md
```diff
@@ -12,48 +12,64 @@ Python with the scientific stack, compiled to WebAssembly.
...
 
-**Pyodide** brings the Python 3.8 runtime to the browser via WebAssembly, along with the Python scientific stack including NumPy, Pandas, Matplotlib, SciPy, and scikit-learn. The [`packages` directory](https://github.com/pyodide/pyodide/tree/master/packages) lists over 75 packages which are currently available. In addition it's possible to install pure Python wheels from PyPi.
+Pyodide may be used in any context where you want to run Python inside a web
+browser.
 
-**Pyodide** provides transparent conversion of objects between Javascript and Python.
-When used inside a browser, Python has full access to the Web APIs.
+Pyodide brings the Python 3.8 runtime to the browser via WebAssembly, along with
+the Python scientific stack including NumPy, Pandas, Matplotlib, SciPy, and
+scikit-learn. The [packages directory](packages) lists over 75 packages which
```
hoodmane, line 20: At some point it'd be good to have a packages index page as suggested in #1380 and then we could link that here.

rth:
Thanks!

PR #1421, opened 2021-04-02
Allow relative URLs in loadPackage

hoodmane:
I updated `package_uri_regexp` to not demand that the URL begins with `https?://` so that we can load relative URLs. Now a string is considered a URL if it ends in ".js". I guess if we are worried about python packages named `something.js` we could change it so that to be considered a URL it also needs to have a slash somewhere (so then you would have to say `pyodide.loadPackage("./pytz.js")` rather than `pyodide.loadPackage("pytz.js")`. I also updated the API docs to be slightly more specific about what a URL should look like.

### DIFF
--- a/src/pyodide.js
+++ b/src/pyodide.js
@@ -45,17 +45,11 @@ globalThis.loadPyodide = async function(config = {}) {
   const DEFAULT_CHANNEL = "default channel";
 
   // Regexp for validating package name and URI
-  const package_uri_regexp =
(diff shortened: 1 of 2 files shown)


rth (approved):
Thanks! Looks good assuming the added test passes in CI.

hoodmane:
It passed locally and in chrome. The firefox-core tests almost all errored including this one.

PR #1422, opened 2021-04-02
MAINT Update to selenium 4.0.0.b3

rth:
Updates to latest selenium 4.0.0.b2  ~~and mypy 0.812~~. The version we have been using was from 2018 and there were no stable releases since.

We can now finally stop monkeypatching selenium with https://github.com/SeleniumHQ/selenium/pull/6480

Initially included in #1209 but it's probably easier the identify the source of potential issues in a separate PR.


rth:
It looks like this does require updating browsers first https://github.com/pyodide/pyodide/pull/1209

PR #263, opened 2018-11-28
Expose Port 8000 for easy testing

madhur-tandon:
Exposing the Port 8000 enables us to start a server within the docker environment. This is helpful in testing the ``pyodide`` builds running within the docker environment using external browser programs of the same machine.

PR #1423, opened 2021-04-03
run_docker: Add option to run docker without binding to any port

ondrej-stanek-ozobot:
Useful for building in multi-user environment.

Context:
`./run_docker` bash script configures docker to bind to a TCP/IP port (8000 by default). This feature was added in PR https://github.com/pyodide/pyodide/pull/263 and is used for accessing a server launched arbitrarily within the docker image. This feature is not needed for the actual `pyodide` build.

Binding to a fixed port causes a problem on multi-user systems, as ports are a shared resource, and users experience port collisions. Users could reconfigure the ports themselves (with the `--port` option), but this is not a systematic solution.

We propose to add an `--no-port` option that will disable the port binding feature.

As an alternative, we propose no to bind to any port unless the `--port` parameter is specified. This change, however, would break backward compatibility. Therefore, our PR addresses this issue in a backward-compatible manner, yet less systematic from design perspective.

### DIFF
--- a/run_docker
+++ b/run_docker
@@ -14,6 +14,8 @@ Usage:
                                 This is ignored if the env var PYODIDE_DOCKER_IMAGE is set
     -p, --port <port>           System port to which to forward.
                                 This is ignored if the env var PYODIDE_SYSTEM_PORT is set
+                                If set to 'none', docker instance will not bind to any port
+                                (Useful for running the docker in multi-user environment)
 
   Prerequisites:
     Docker has to be set up on your system
@@ -69,8 +71,22 @@ PYODIDE_DOCKER_PORT=${PYODIDE_DOCKER_PORT:-"8000"}
 PYODIDE_SYSTEM_PORT=${PYODIDE_SYSTEM_PORT:-${DEFAULT_PYODIDE_SYSTEM_PORT}}
 PYODIDE_DOCKER_IMAGE=${PYODIDE_DOCKER_IMAGE:-${DEFAULT_PYODIDE_DOCKER_IMAGE}}
 
+# in case the port is not a number, do not bind the port
+case $DEFAULT_PYODIDE_SYSTEM_PORT in
+  none)
+  PORT_CONFIGURATION_LINE=""
+  ;;
+  ''|*[!0-9]*) # contains a non-digit character, therefore it is not a number
+  echo "WARNING: Invalid port argument '$DEFAULT_PYODIDE_SYSTEM_PORT'. Port binding disabled."
+  PORT_CONFIGURATION_LINE=""
+  ;;
+  *)
+  PORT_CONFIGURATION_LINE="-p $PYODIDE_SYSTEM_PORT:$PYODIDE_DOCKER_PORT"
+  ;;
+esac
+
 exec docker run \
-    -p "$PYODIDE_SYSTEM_PORT":"$PYODIDE_DOCKER_PORT" \
+    $PORT_CONFIGURATION_LINE \
     -it --rm \
     -v $PWD:/src \
     --user root -e NB_UID=$UID -e NB_GID=$GID \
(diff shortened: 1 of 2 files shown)


rth:
Thanks for proposing this! How about `--port none` or `--port false`, that way we could add this feature without breaking backward compatibility and adding a new parameter?

run_docker
```diff
@@ -69,8 +71,19 @@ PYODIDE_DOCKER_PORT=${PYODIDE_DOCKER_PORT:-"8000"}
 PYODIDE_SYSTEM_PORT=${PYODIDE_SYSTEM_PORT:-${DEFAULT_PYODIDE_SYSTEM_PORT}}
 PYODIDE_DOCKER_IMAGE=${PYODIDE_DOCKER_IMAGE:-${DEFAULT_PYODIDE_DOCKER_IMAGE}}
 
+# in case the port is not a number, do not bind the port
+case $DEFAULT_PYODIDE_SYSTEM_PORT in
+  ''|*[!0-9]*) # contains a non-digit character, therefore it is not a number
+  echo "WARNING: port binding disabled (received '$DEFAULT_PYODIDE_SYSTEM_PORT')"
```
hoodmane, line 77: Probably would be good to have a separate "none" case which doesn't display this message.

ondrej-stanek-ozobot: Makes sense. Fixed.

run_docker
```diff
@@ -69,8 +71,22 @@ PYODIDE_DOCKER_PORT=${PYODIDE_DOCKER_PORT:-"8000"}
 PYODIDE_SYSTEM_PORT=${PYODIDE_SYSTEM_PORT:-${DEFAULT_PYODIDE_SYSTEM_PORT}}
 PYODIDE_DOCKER_IMAGE=${PYODIDE_DOCKER_IMAGE:-${DEFAULT_PYODIDE_DOCKER_IMAGE}}
 
+# in case the port is not a number, do not bind the port
+case $DEFAULT_PYODIDE_SYSTEM_PORT in
+  none)
+  PORT_CONFIGURATION_LINE=""
+  ;;
+  ''|*[!0-9]*) # contains a non-digit character, therefore it is not a number
+  echo "WARNING: port binding disabled (got '$DEFAULT_PYODIDE_SYSTEM_PORT')"
```
hoodmane, line 80: How about:
```suggestion
  echo "WARNING: Invalid port argument '$DEFAULT_PYODIDE_SYSTEM_PORT'. Port binding disabled."
```

run_docker
```diff
@@ -69,8 +71,22 @@ PYODIDE_DOCKER_PORT=${PYODIDE_DOCKER_PORT:-"8000"}
 PYODIDE_SYSTEM_PORT=${PYODIDE_SYSTEM_PORT:-${DEFAULT_PYODIDE_SYSTEM_PORT}}
 PYODIDE_DOCKER_IMAGE=${PYODIDE_DOCKER_IMAGE:-${DEFAULT_PYODIDE_DOCKER_IMAGE}}
 
+# in case the port is not a number, do not bind the port
+case $DEFAULT_PYODIDE_SYSTEM_PORT in
+  none)
+  PORT_CONFIGURATION_LINE=""
+  ;;
+  ''|*[!0-9]*) # contains a non-digit character, therefore it is not a number
```
ondrej-stanek-ozobot, line 79: Maybe the additional check for a number could be dropped, because the error from `docker` would be descriptive enough:
```
pyodide$ ./run_docker --port false
docker: invalid publish opts format (should be name=value but got 'false:8000').
See 'docker run --help'.
```
But this will cause a hard fail.

hoodmane: I think this is good as-is.

hoodmane:
@rth Something has gone awry with the readthedocs build since yesterday. It worked fine on the PRs opened yesterday but is failing on the ones opened today.

hoodmane:
Thanks @ondrej-stanek-ozobot!

PR #1427, opened 2021-04-03
DOCS Edits to Javascript API docs

hoodmane:
I edited the API docs for consistency and completeness. I added a `async` label to the async functions, I think this makes it clearer what's going on rather than saying "it returns a promise that resolves when it's done". I set the font size of code to be the same as the size of normal text, this may be a slightly debatable change but the visual inconsistency in the API docs due to repeated text size changes was bothering me.

I also reordered a few functions in `pyodide.js` to put related functions together. This reorders the functions in the API docs as well and I think mildly improves readability of the file.


docs/_static/css/pyodide.css
```diff
@@ -2,3 +2,7 @@
     margin-right: 5px;
     opacity: 0.7;
 }
+
+code {
+    font-size: 100% !important;
```
rth, line 7: Which code blocks does this refer to. Might be preferable to use a more specific selector rather than the `!important` tag.

src/pyodide.js
```diff
@@ -523,64 +566,6 @@ globalThis.loadPyodide = async function(config = {}) {
     return Module.pyodide_py.eval_code(code, globals);
   };
 
-  // clang-format off
```
rth, line 526: What was the motivation for moving this code block around? It would make using git blame more difficult in the future.

rth (commented):
A few comments otherwise LGTM. Thanks!

docs/_static/css/pyodide.css
hoodmane: It's overriding another selector that says
```css
code {
    font-size: 87.5% !important;
}
```
so I think there is no other choice here.

src/pyodide.js
rth: > but I think it's helpful for them to be next to each other in the source too.

I'm OK with the change here if you find it useful. In future PRs though it is really helpful to go back in git history with git blame to understand changes, and moving functions around kind of breaks that workflow. It's still possible, just that the resulting improvement need to be worth that cost.

> It's also possible to sort API docs based on the order in PUBLIC_API

Maybe we could sort them alphabetically as autodoc would do by default https://www.sphinx-doc.org/en/master/usage/extensions/autodoc.html#confval-autodoc_member_order ?

hoodmane: In this case I think the small benefit is almost exactly matched by the small cost... \*shrug\*

So there are three options for sort: sort as they appear in `pyodide.js`, sort alphabetically, or sort by `PUBLIC_API`. The reason I didn't sort alphabetically is that sometimes it sorts closely related methods far away from each other (e.g., `register` / `unregister`). It's a pretty minor issue, I'm happy to just pick whatever and stick with it.

hoodmane: Okay I reverted this, will sort js API docs alphabetically.

rth (approved):
Thanks!

PR #1433, opened 2021-04-04
Adjustments to JsMethod kwargs handling

hoodmane:
This removes the `supports_kwargs` test and shares the argument handling between the `new` method and the `__call__` method. This allows key word arguments in constructor calls.

I removed the `supports_kwargs` test because it generates too many false negatives and doesn't work at all on constructors. This sort of question is really unanswerable in Javascript since there is no good way to get a function signature. For instance `fetch` takes a large number of keyword arguments but `supports_kwargs` would say it doesn't. With this PR, we can do:
```py
import json
async def post_data(url, data):
    response = await fetch(url,
        method='POST',
        mode='cors',
        # ... many more options possible
        body=json.dumps(data)
    )
   js_json = await response.json()
   return js_json.to_py()
```
whereas before it would raise an error saying that `fetch` doesn't accept kwargs.


rth (approved):
Thanks! Generally this sounds good to. I'm happy this allows to remove "function_supports_kwargs"

PR #1438, opened 2021-04-07
Add run_in_executor shim to webloop

hoodmane:
This provides a webloop implementation of `run_in_executor`. The point of `run_in_executor` is for threading, so without threads around it is pretty pointless. We just run the code in this thread and then wrap the result in a future, ignoring the executor. Looking at the executor funny might cause it to try to start a thread and throw an error.

This should make some code that tries to mix asyncio with threading work by skipping the threading.

### DIFF
--- a/src/pyodide-py/pyodide/webloop.py
+++ b/src/pyodide-py/pyodide/webloop.py
@@ -156,6 +156,20 @@ class WebLoop(asyncio.AbstractEventLoop):
         delay = when - cur_time
         return self.call_later(delay, callback, *args, context=context)
 
+    def run_in_executor(self, executor, func, *args):
+        """This is normally supposed to run func(args) in a separate thread and
+        signal back to our event loop when it is done. It's possible to make the
+        executor, but if we actually try to submit any functions to it, it will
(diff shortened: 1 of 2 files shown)


src/pyodide-py/pyodide/webloop.py
```diff
@@ -156,6 +156,20 @@ def call_at(
         delay = when - cur_time
         return self.call_later(delay, callback, *args, context=context)
 
+    def run_in_executor(self, executor, func, *args):
+        """This is normally supposed to run func(args) in a separate thread and
```
rth, line 160: Could we have a 1 line summary, then an empty line? :)

It's not just numpydoc conversion also general Python convention for docstrings,

https://www.python.org/dev/peps/pep-0257/#multi-line-docstrings

> Multi-line docstrings consist of a summary line just like a one-line docstring, followed by a blank line, followed by a more elaborate description. The summary line may be used by automatic indexing tools; it is important that it fits on one line and is separated from the rest of the docstring by a blank line.

rth (commented):
Could you add a test for it? Thanks! Otherwise LGTM.

rth (approved):

PR #1441, opened 2021-04-07
TST/CI test dependent selenium WebDriver timeout

rth:
Adds a mechanism to specify the selenium WebDriver timeout per test. The default (20s) is reasonable for core tests, however packages that depend on scipy for instance can take longer. At the same time having a larger value globally means that in case of genuine failure tests can take very long to complete.

With this PR,
 - we keep the default timeout of 20s everywhere
 - it's possible to set the timeout for individual tests with,
   ```
   @pytest.mark.driver_timeout(<value in seconds>)
   ```
 - in CI, pytest tests exit if more than ~30 to 40 tests fail (`pytest --maxfail`). Which limits the maxium run time in case of failure where all tests timeout (this still seems to happen arbitrarily occasionally) to ~15min. 

This should fix some of the package tests that currently timeout, 

Ref https://github.com/pyodide/pyodide/issues/1389

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -22,15 +22,21 @@ BUILD_PATH = ROOT_PATH / "build"
 sys.path.append(str(ROOT_PATH))
 
 from pyodide_build._fixes import _selenium_is_connectable  # noqa: E402
+from pyodide_build.testing import set_webdriver_script_timeout, parse_driver_timeout
 
-try:
-    import selenium.webdriver.common.utils  # noqa: E402
 
-    # XXX: Temporary fix for ConnectionError in selenium
+def _monkeypatch_selenium():
+    try:
+        import selenium.webdriver.common.utils  # noqa: E402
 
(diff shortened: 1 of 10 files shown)


rth:
> Which limits the maxium run time in case of failure where all tests timeout (this still seems to happen arbitrarily occasionally) to ~15min.

For comparison currently Chrome core occasionally fails with,
```
E       selenium.common.exceptions.SessionNotCreatedException: Message: session not created exception
E       from timeout: Timed out receiving message from renderer: 600.000
```
for all tests, which results in a 35 min run time with all tests that failed.

hoodmane (approved):

rth:
So it doesn't solve all issues but it's an improvement. For instance on master 15 package tests failed on Chrome, here there are 4.

> In CI, pytest tests exit if more than ~30 to 40 tests fail (pytest --maxfail)

Unfortunately apparently this works if the tests failed, not if the fixture failed to initialize (which is what currently happens with Chrome core tests occasionally)

PR #1445, opened 2021-04-09
Deploy console.html as part of the documentation

rth:
Closes https://github.com/pyodide/pyodide/issues/1380

In the end deploying console.html as part of the docs is probably the easiest solution. We'll just have to remember to update the `PYODIDE_BASE_URL` (in `docs/conf.py`) for releases, same as we do in `.circleci/config.yml`.

Unfortunately it doesn't look like it's possible to pass env variables to a Readthedocs build (https://github.com/readthedocs/readthedocs.org/issues/6311) so we have to hard-code it in  `docs/conf.py` for now.

We will only be able to update the REPL link in the readme once there a new release is made. And of course more inter-linking could be done with the documentation, but I'll leave that for a next PR.

Rendered version: https://pyodide--1445.org.readthedocs.build/en/1445/console.html


hoodmane (approved):
Looks nice. Once we merge this we should probably add some links to it in the docs, otherwise no one will ever find it.

rth:
> Once we merge this we should probably add some links to it in the docs, otherwise no one will ever find it.

Yes, definitely. A few icons with links from the console to the doc/GH would also have been nice.

hoodmane:
> icons with links from the console to the doc/GH would also have been nice.

That's a great idea.

PR #1447, opened 2021-04-09
Merge reentrant Python stack traces

hoodmane:
The point of this PR is to recover the original exception if a Python exception bubbles out into Javascript and then back to Python. This should result in more legible tracebacks and more reasonably behaved error handling. I also refactored `error_handling.c` a bit.

As an example, suppose we run the following code:
```js
function a(){
   pyodide.globals.get("pyfunc")();
}
pyodide.runPython(`
def pyfunc():
   raise KeyboardInterrupt
from js import a
try:
   a()
except Exception as e:
   print(e)
print("finished")
`);
```
Currently the `KeyboardInterrupt` turns into a `PythonError` which turns into a `JsException` which inherits from `Exception` and gets caught. With this PR, the `KeyboardInterrupt` comes out of the `a()` function call as a `KeyboardInterrupt` and does not get caught.

I add an extra line to the traceback that says:
`File "???.js", line -1, in <javascript frames>`
In the future I think it would be nice to use [stacktracejs](https://www.stacktracejs.com/) to replace this with more detailed stack trace info.


src/core/error_handling.c
```diff
@@ -88,55 +149,55 @@ format_exception_traceback(PyObject* type, PyObject* value, PyObject* traceback)
...
-wrap_exception(bool attach_python_error)
+wrap_exception()
 {
   bool success = false;
-  PyObject* type = NULL;
-  PyObject* value = NULL;
-  PyObject* traceback = NULL;
+  PyObject* type;
+  PyObject* value;
+  PyObject* traceback;
```
rth, line 172: Wouldn't it be be better to explicitly initialize it to `NULL` as before, before passing it to functions?

rth (approved):
Thanks @hoodmane ! Could you please merge main to make sure it still passes, and add a changelog entry?

src/core/error_handling.c
hoodmane: I suppose so. The functions unconditionally write to these by using them as arguments to [`PyErr_Fetch`](https://docs.python.org/3/c-api/exceptions.html?highlight=pyerr_fetch#c.PyErr_Fetch), but it is harmless to initialize to `NULL`.

Issue #1384, opened 2021-03-28
Memory access out of bounds in Chrome


dmondev:
I'm getting the following intermittent error running the develop version in Chrome, running from a web worker.

Chrome Version 89.0.4389.90 (Official Build) (64-bit)
Windows 10 1909 build 18363.1379

pyodide.js:406 The cause of the fatal error was:
 RuntimeError: memory access out of bounds
    at PyArray_IntpFromIndexSequence (<anonymous>:wasm-function[991]:0x737a1)
    at PyArray_IntpConverter (<anonymous>:wasm-function[990]:0x73702)
    at byn$fpcast-emu$PyArray_IntpConverter (<anonymous>:wasm-function[3415]:0x15bfeb)

leopsidom:
I can confirm this behavior for 0.17.0a2. I'm getting quite a few crashes when trying to load pandas. It seems like the dev version consumes more memory than the previous version ?

hoodmane:
I've seen this a couple of times too, but not enough times to have any idea of a pattern for when it happens. If anyone can figure out a way to reproduce it, that would be super helpful.

hoodmane:
Also it would be useful to know whether it is browser dependent, so if anyone has seen a similar crash in firefox that'd be interesting info.

dmondev:
Weirdly enough it happens less when I have dev tools open. I'll try to reproduce with a minimal setup/MRE.

hoodmane:
Thanks!

leopsidom:
+1 on open dev tool makes this happen less frequently. Another pattern I observe is that this seems to happen more frequently on Mac (chrome) than Windows (chrome).

dmondev:
I can now say that this happens constantly when Chrome DEV tools is **NOT** open. 

The issue doesn't happen in Firefox.

I'm able to reproduce with the following codepen: https://codepen.io/dmondev/pen/eYgdBmm

This is as close as possible to how I'm using pyodide (in a worker, with custom namespaces).

hoodmane:
Note that the problem goes away here if you import numpy _before_ pandas. Have you tried that in your actual use case? Presumably the error just appears further down the road?

hoodmane:
Also I'd appreciate if you could modify the codepen to add some indication when the worker responds to the message: because the problem goes away when the devtools is open it's not all that convenient that the only indication that it finished goes into the browser console.

dmondev:
Done. updated the codepen with output to the document itself.

[...]

PR #1449, opened 2021-04-11
Fix numpy in chromium v89 

hoodmane:
There seems to be a bug in the v8 wasm runtime. The failure occurs in the numpy
definition of PyArray_IntpFromIndexSequence in the following code:
```C
vals[i] = PyArray_PyIntAsIntp(op);    // line 958
Py_DECREF(op);                        // line 959
if(vals[i] == -1) {                   // line 960, fails here "RuntimeError: memory access out of bounds"

}
```
This is compiled to the following:
```wat
;; First calculate &vals[i]. vals is an array of i32, so this is vals + 4*i.
;; line 958: vals[i] = PyArray_PyIntAsIntp(op);
local.get $var1 ;; vals
local.get $var3 ;; i
i32.const 2 
i32.shl ;; i << 2
i32.add ;; &vals[i] == vals + 4*i
local.tee $var7 ;; v7 = &vals[i], leave &vals[i] on stack
(; Snipped out argument setup and inlined PyArray_PyIntAsIntp ;)
call $PyArray_PyIntAsIntp
i32.store ;; vals[i] = PyArray_PyIntAsIntp(op) ;; Store to vals[i]
;; line 959: Py_DECREF(op);
(; Snipped out Py_DECREF(op) ;)
block $label1
    ;; line 960: if(vals[i] == -1) {
    local.get $var7 ;; &vals[i]
    i32.load ;; Load vals[i] back <== Fails here "RuntimeError: memory access out of bounds"
    i32.const -1 
    i32.ne ;; vals[i] != -1
    br_if $label1
```

It doesn't make a whole lot of sense that it fails here, we just succesfully
stored to vals[i] so why does it fail when we load back? Unforunately the bug
disappears when devtools are open making it very difficult to assess what state
the wasm VM is in when the bug is triggered.

The goal here was to refactor the source to remove the offending load
instruction. There are several options for how to change it, I chose the
following:
```C
int x = PyArray_PyIntAsIntp(op);    // line 958
Py_DECREF(op);                      // line 959
if(x == -1) {                       // line 960

}
vals[i] = x;                        // line 971
```
Updated WAT looks like what you'd expect:
```wat
    ;; line 958: int x = PyArray_PyIntAsIntp(op);
    (; Snipped out argument setup and inlined PyArray_PyIntAsIntp ;)
    call $PyArray_PyIntAsIntp
    ;; Now instead of storing result into vals[i], we store it in a local variable
    local.set $var6 
    ;; line 959: Py_DECREF(op)
    (; Snipped out Py_DECREF(op) ;)
    ;; line 960: if(x == -1)
    block $label1
        local.get $var6 ;; now check value of local variable
        i32.const -1
        i32.ne
        br_if $label1
        (; Snipped out body of conditional ;)
    end $label1
    ;; line 971: vals[i] = x
    local.get $var1 ;; vals
    local.get $var3 ;; i
    i32.const 2   
    i32.shl         ;; 4*i
    i32.add         ;; vals + 4*i
    local.get $var6 ;; x
    i32.store
```
Indeed the offending `i32.load` is gone and so is the crash.

### DIFF
--- a/packages/numpy/meta.yaml
+++ b/packages/numpy/meta.yaml
@@ -7,6 +7,7 @@ source:
   sha256: 16507ba6617f62ae3c6ab1725ae6f550331025d4d9a369b83f6d5a470446c342
 
   patches:
+    - patches/fix-i32-load-memory-access-oob.patch
     - patches/add-emscripten-cpu.patch
     - patches/disable-maybe-uninitialized.patch
     - patches/dont-include-execinfo.patch
--- a/packages/numpy/patches/fix-i32-load-memory-access-oob.patch
+++ b/packages/numpy/patches/fix-i32-load-memory-access-oob.patch
@@ -1,0 +1,115 @@
+From 6cf164b6a1a7cc151aa52ab5bb9e20a95a6a5e67 Mon Sep 17 00:00:00 2001
+From: Hood <hood@mit.edu>
+Date: Fri, 2 Apr 2021 14:24:24 -0700
+Subject: [PATCH] Fix "load" memory access out of bounds on Chrome v89 #1384
+
+Resolves https://github.com/pyodide/pyodide/issues/1384.
+
+There seems to be a bug in the v8 wasm runtime. The failure occurs in the numpy
+definition of PyArray_IntpFromIndexSequence in the following code:
+```C
+vals[i] = PyArray_PyIntAsIntp(op);    // line 958
+Py_DECREF(op);                        // line 959
+if(vals[i] == -1) {                   // line 960, fails here "RuntimeError: memory access out of bounds"
+
+}
+```
+This is compiled to the following:
+```wat
+;; First calculate &vals[i]. vals is an array of i32, so this is vals + 4*i.
+;; line 958: vals[i] = PyArray_PyIntAsIntp(op);
+local.get $var1 ;; vals
+local.get $var3 ;; i
+i32.const 2 
+i32.shl ;; i << 2
+i32.add ;; &vals[i] == vals + 4*i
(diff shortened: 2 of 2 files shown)


hoodmane:
Update: this doesn't seem to work. When I run the copy of Pyodide that I compiled locally, importing numpy works fine. However, the version here:
https://1061-322074228-gh.circle-artifacts.com/0/root/repo/build/console.html
does not work...

hoodmane:
Further investigation shows the generated code on the CI build is the same as the original code without my patch:
```wat
            br_if $label0
            local.get $var1
            local.get $var3
            i32.const 2
            i32.shl
            i32.add
            local.tee $var7
            (; elided  PyArray_PyIntAsIntp(op); ;)
            (; elided Py_DECREF(op); ;)
            block $label1
              local.get $var7
              i32.load ;; crash here
              i32.const -1
              i32.ne
              br_if $label1
```

hoodmane:
@rth could this be a problem with CCache or something like that? I saw a similar issue locally where it took several builds before the generated WASM reflected the patch.

hoodmane:
So it turns out that
https://1061-322074228-gh.circle-artifacts.com/0/root/repo/build/console.html
was loading Pyodide not from `circle-artifacts.com` but from `jsdeliver` so that was the problem.

hoodmane:
Okay @jmsmdy confirmed on #1384 that it works for him too, so I think this patch actually fixes the problem.

rth (approved):
I can confirm the fix also works for me with local overrides in Chrome!

Great works @hoodmane !

PR #1451, opened 2021-04-12
FAQ pickle to send objects between server and client

### LINKED ISSUE #581: What is the best way to pass a Python object from backend to pyodide, or is it even possible?
Say I want to serve a frontend page that's running pyodide, while the backend is in Python. Is it possible to pass a Python object from backend to the pyodide runtime?

hoodmane:
This resolves #581.

### DIFF
--- a/docs/usage/faq.md
+++ b/docs/usage/faq.md
@@ -101,7 +101,7 @@ if platform.system() == 'Emscripten':
 This however will not work at build time (i.e. in a `setup.py`) due to the way
 the Pyodide build system works. It first compiles packages with the host compiler
 (e.g. gcc) and then re-runs the compilation commands with emsdk. So the `setup.py` is
-never run inside the Pyodide environement.
+never run inside the Pyodide environment.
 
 To detect Pyodide, **at build time** use,
 ```python
--- a/docs/usage/faq.md
+++ b/docs/usage/faq.md
@@ -144,7 +144,7 @@ assert my_js_module.f(7) == 50
 assert h(9) == 80
 assert c == 2
 ```
-## How can I send a Python object from my server to Pyodide (or vice-versa)?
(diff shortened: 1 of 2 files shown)


docs/usage/faq.md
```diff
@@ -144,3 +144,15 @@ assert my_js_module.f(7) == 50
...
+The best way to do this is with pickle. We do this in our test suite on a wide
+variety of objects. If the version of Python used in the server exactly matches
+the version of Python used in the client, then in our experience, every object
+that can be successfully pickled can be sent to the client and unpickled in
+Pyodide. If the versions of Python are different then for instance sending AST
+is unlikely to work since there are breaking changes to Python AST in most
+Python minor versions.
+
+We don't have any test coverage for pickling objects in Pyodide and unpickling
+on the server, but that probably works quite well too.
```
rth, line 158:
```suggestion
The best way to do this is with pickle. If the version of Python used in the
server exactly matches the version of Python used in the client, then objects
that can be successfully pickled can be sent to the client and unpickled in
Pyodide. If the versions of Python are different then for instance sending AST
is unlikely to work since there are breaking changes to Python AST in most
Python minor versions.

Similarly when pickling Python objects defined in a Python package, the package
 version needs to match exactly between the server and pyodide. 

Generally, pickles are portable between architectures (here amd64 and wasm32).
The rare cases when they are not portable, for instance currently tree based
 models in scikit-learn, can be considered as a bug in the upstream library.  
```

rth (approved):
Thanks!

docs/usage/faq.md
rth: Maybe we should also mention some of the security implications of using pickles. For instance from https://scikit-learn.org/stable/modules/model_persistence.html#security-maintainability-limitations

Issue #648, opened 2020-05-09
Use pyodide dev from CircleCI artifacts


rth:
So each commit on master builds and stores the packages in CircleCI artifacts, e.g.

https://circleci.com/gh/iodide-project/pyodide/5623#artifacts/containers/0
which produces for instance,
https://5623-122663163-gh.circle-artifacts.com/0/home/circleci/repo/build/test.html

If we could use that, this means that we would get nightly builds for free as anyone could use the latest pyodide version from master.

Currently loading that `test.html` files produces,
```
CompileError: wasm validation error: at offset 0: failed to match magic number 
pyodide_dev.js line 337 > WebAssembly.compile:337
```
here,
https://github.com/iodide-project/pyodide/blob/c9fd3e1d572354113ca57351a09c09c872ba0f56/src/pyodide.js#L337

This doesn't seem to a MIME problem, which was the previous issue https://github.com/iodide-project/pyodide/issues/174 and was fixed in https://github.com/iodide-project/pyodide/pull/628. This might need more investigation to understand what's wrong exactly.

cc @jcaesar

rth:
So essentially, `languagePluginUrl` needed to be set properly and then one has to use a CORS proxy to resolve cross-origin restrictions.

For instance, the following should work to get the latest pyodide version from master,
```html
<html>
<head>
  <meta charset="utf-8">
</head>
<body>
  <script type="text/javascript">
       window.languagePluginUrl = 'https://yacdn.org/serve/https://5623-122663163-gh.circle-artifacts.com/0/home/circleci/repo/build/';

  </script>
  <script type="text/javascript" src="https://5623-122663163-gh.circle-artifacts.com/0/home/circleci/repo/build/pyodide_dev.js"></script>
  <script type="text/javascript">
    languagePluginLoader.then(function () {

        pythonCode = `
	    import sys
	    print(sys.version)
        `
        languagePluginLoader.then(() => {
          pyodide.runPython(pythonCode);
        })
  
    });
  </script>
</body>
</html>
```

where 
 - `5623-122663163-gh.circle-artifacts.com` is the hostname that one can find by looking in the "Artifacts" CircleCI tab of a particular commit.
 - `https://yacdn.org/serve/` is a CORS proxy prefix (see https://ovsoinc.github.io/yacdn.org/ for more details). For some reason I haven't managed to make `https://cors-anywhere.herokuapp.com/` that we use in micropip work. Of course using a third party proxy can have security concerns.

[...]

PR #1453, opened 2021-04-12
Correct PYODIDE_BASE_URL for  console.html in CircleCI artifacts

rth:
Closes https://github.com/pyodide/pyodide/issues/648

By adjusting PYODIDE_BASE_URL before artifacts are stored in CircleCI but after the workspace has been saved (which is used for S3 deployment), I think we should be able to use console.html from CircleCI artifacts without using local overrides in Chrome https://github.com/pyodide/pyodide/issues/1384#issuecomment-817411787

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -131,7 +131,7 @@ jobs:
             - ./build
       - run:
           name: Prepare CircleCI artifacts
-          command: PYODIDE_BASE_URL="$CIRCLE_BUILD_URL/artifacts/0/root/repo/build/" make build/console.html
+          command: PYODIDE_BASE_URL="./" make build/console.html
 
       - store_artifacts:
           path: /root/repo/build/
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -129,6 +129,9 @@ jobs:
           paths:
             - ./packages/.artifacts
             - ./build
+      - run:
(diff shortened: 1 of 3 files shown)


rth:
No this still fails due to CORS when loading packages in `console.html`. I need to change the `PYODIDE_BASE_URL` to take into account the first redirect that happens when downloading packages.

.circleci/config.yml
```diff
@@ -129,6 +129,9 @@ jobs:
           paths:
             - ./packages/.artifacts
             - ./build
+      - run:
+          name: Prepare CircleCI artifacts
+          command: PYODIDE_BASE_URL="$CIRCLE_BUILD_URL/artifacts/0/root/repo/build/" make build/console.html
```
hoodmane, line 134: I think it will work if you do:
```suggestion
          command: PYODIDE_BASE_URL="./" make build/console.html
```

rth: Yes of course that's the most straightforward fix. Thanks!

hoodmane:
Could you restart the `build-packages` CI so we can test this?

rth:
OK, this this now work https://217-140441556-gh.circle-artifacts.com/0/root/repo/build/console.html !

hoodmane:
Now that this works we should probably add a suggestion somewhere to the dev docs.

PR #1457, opened 2021-04-12
FIX Tell loadPackage about packages installed with micropip

### LINKED ISSUE #725: Version selection for packages availble both on PyPi and in Pyodide
For packages not built in pyodide, version selection works as expected. For instance,
```py
>>> import micropip
>>> micropip.install('idna==2.9')   # version before last on PyPi, package not in pyodide
Installed idna
>>> import idna
>>> idna.__version__
2.9
```

However, when one specifies the version for a package available in the pyodide distribution, it is ignored and the version from pyodide is installed regardless if PyPi includes the requested version,
```py
>>> import micropip
>>> micropip.install('pytz==2020.1')
Installed pytz
>>> import pytz
>>> pytz.__version__
2019.3
```

hoodmane:
This resolves #725. However, there are still some issues. The following code fails with an IO error
```js
import micropip
await micropip.install("https://files.pythonhosted.org/packages/70/94/784178ca5dd892a98f113cdd923372024dc04b8d40abe77ca76b5fb90ca6/pytz-2021.1-py2.py3-none-any.whl")
import pytz
`)
```
because both `loadPackagesFromImports` and `micropip.install` try to install `pytz`.

The situation would be solved if we removed the `loadPackagesFromImports` call from `runPythonAsync`. 
@rth What do you think about this?

### DIFF
--- a/docs/project/changelog.md
+++ b/docs/project/changelog.md
@@ -55,16 +55,16 @@ substitutions:
   `pyodide.globals.get('key')`. [#1367](https://github.com/pyodide/pyodide/pull/1367)
 - {{ API }} Added `PyProxy.getBuffer` API to allow direct access to Python
   buffers as Javascript TypedArrays.
-  [1215](https://github.com/pyodide/pyodide/pull/1215)
+  [#1215](https://github.com/pyodide/pyodide/pull/1215)
 - {{ API }} The innermost level of a buffer converted to Javascript used to be a
   TypedArray if the buffer was contiguous and otherwise an Array. Now the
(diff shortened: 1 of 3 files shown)


hoodmane:
There seems to be a second bug that breaks `micropip.install("pytz==2021.1")`. This is a regression since v0.16.1, probably I introduced it in one of my micropip refactors.

rth (approved):
Thanks! Could you please add a test or extend and existing one?

> There seems to be a second bug that breaks micropip.install("pytz==2021.1")

We could look into it in a separate PR.

hoodmane:
Okay I think it passed the test.

PR #1459, opened 2021-04-13
Support top level await in console.html

hoodmane:
One console test is still broken.


rth (commented):
This looks like a nice improvement! Could you merge master in so that one can run the REPL from build-core artifacts? Also please add a changelog entry.

rth:
I'm not sure if it's this PR or something else we merged very recently, but in this version if there is an exception in console, it will be printed for any following commands essentially making REPL unusable.

For instance with,
```py
>>> from js import Promise
>>> Promise()
```

hoodmane:
I think it's an issue with this PR.

hoodmane:
Okay that should be fixed.

hoodmane:
Okay this works now. We still need to fix the syntax errors problem. My PR on jquery terminal probably won't be merged for a long time still jcubic/jquery.terminal#624 so our options are to use a custom version of jquery terminal built from my branch or to revert to @casatir's workaround. In interest of minimizing effort let's revert to @casatir's workaround for 0.17.0.

hoodmane:
Also, we really should switch to pytest-asyncio, but it's a bit of a pain to update the docker image.

Issue #936, opened 2020-12-24
Flaky scipy build


dalcde:
Recently we are sometimes getting the following error when building scipy
```
Traceback (most recent call last):
  File "setup.py", line 265, in <module>
    setup_package()
  File "setup.py", line 253, in setup_package
    from numpy.distutils.core import setup
ModuleNotFoundError: No module named 'numpy'
```
There have been similar issues with firefox tests too.

rth:
From https://github.com/pyodide/pyodide/issues/1508

> We observe that non-interactive builds do not work any longer on 0.17.0 preview

@ondrej-stanek-ozobot This also occasionally happens in CI. Likely a glitch in the multi-threaded build mechanism. It's likely (but not 100% certain) that setting `PYODIDE_JOBS=1` env variable would resolve it (while making the build slower). 

We would need to investigate in more detail. I have re-checked the code recently and haven't seen any obvious parts that could cause this. https://github.com/pyodide/pyodide/issues/1476 would likely be a long term solution.

rth:
Another reason for https://github.com/pyodide/pyodide/issues/1476 is that currently it's hard to reproduce and take action on that exception since it happens inside the setup.py installation code not our code.

hoodmane:
I thought it would be a quick fix to install numpy with pip on the docker, copy over the host numpy config as we do, then get rid of `skip_host: False`.

Building numpy takes up a significant percentage of the build-packages CI job, both because it is slow itself and because a lot of other slow-building packages have to wait on it. If it works to skip host build that might improve build time quite a lot.

rth:
> I thought it would be a quick fix to install numpy with pip on the docker, copy over the host numpy config as we do,

For Docker likely, but that's harder to implement for people installing outside of docker (which we also support) since we then have less control over paths. Also I'm not sure if copying that config would be enough. It would be worth a try.

> Building numpy takes up a significant percentage of the build-packages CI job,

Hmm, so on my laptop installing numpy 1.17.5 from source takes 2min (without ccache). In our build system it should be 2x slower (and then a bit more slower due to CI). Still it shouldn't be that much.

hoodmane:
From watching the build-packages job I think building numpy is >50% of the total time. We should add timestamps to the "building pkgname" / "built pkgname" logs so it would be easier to see.

rth:
[...]

PR #822, opened 2020-12-05
PKG Downgrade parso to 0.7.1

dalcde:
This is what jedi 0.17.2 wants.

This addresses the crashes observed in the comments of #821 (but not the original issue)

PR #1460, opened 2021-04-13
Used pyodide_build auto package update system

hoodmane:
It doesn't look like all of the resulting `meta.yml` files are even valid. Probably some of these packages should be updated?


packages/joblib/meta.yaml
```diff
@@ -1,15 +1,12 @@
 package:
   name: joblib
-  version: "0.11"
-
+  version: 1.0.1
```
rth, line 3: Probably better to exclude this one for now, it would likely need to update patches.

Actually the update script should not have applied on a package with patches (and the yaml is not correct)

packages/cytoolz/meta.yaml
```diff
@@ -1,13 +1,13 @@
...
-  sha256: 82f5bba81d73a5a6b06f2a3553ff9003d865952fcb32e1df192378dd944d8a5c
-  url: https://files.pythonhosted.org/packages/62/b1/7f16703fe4a497879b1b457adf1e472fad2d4f030477698b16d2febf38bb/cytoolz-0.10.1.tar.gz
+  version: 0.11.0
 requirements:
   run:
   - nose
   - toolz
+source:
+  sha256: c64f3590c3eb40e1548f0d3c6b2ccde70493d0b8dc6cc7f9f3fec0bb3dcd4222
+  url: https://files.pythonhosted.org/packages/58/67/1c60da8ba831bfefedb64c78b9f6820bdf58972797c95644ee3191daf27a/cytoolz-0.11.0.tar.gz
```
rth, line 10: The fact that order of sections seems to be arbitrary is unfortunate. I have not seen this issue before when using this updater.

packages/scikit-image/meta.yaml
```diff
@@ -1,22 +1,22 @@
...
-  run:
-    - numpy
-    - scipy
-    - matplotlib
-    - networkx
-    - pillow
-    - imageio
-    - pywavelets
+  sha256: fbb618ca911867bce45574c1639618cdfb5d94e207432b19bc19563d80d2f171
+  url: https://files.pythonhosted.org/packages/6e/be/a8ccf9d949a55023cf02438310e068c263ce3dd6bbf31f9229d3db6e551a/scikit-image-0.18.1.tar.gz
```
rth, line 19: Similarly let's address this in a separate PR. And scikit-image currently cannot be updated as it requires a newer version of scipy.

rth:
We really need to fix https://github.com/pyodide/pyodide/issues/936, I have a feeling it's happening 1 time out of 2 now.

hoodmane:
> it's happening 1 time out of 2 now

Yeah sounds about right.

hoodmane:
Closed in favor of #1456 where I improved the update-all script.

hoodmane:
Actually I guess I'll repurpose this.

hoodmane:
Now this has the result of running `make -C packages update-all` on #1465.

rth (approved):
Thanks, looks good assuming Firefox CI passes.

In the future we should probably run proper dependency resolution, say with conda for the include packages to makes sure all versions are compatible, but for now this should be fine.

hoodmane:
Yeah independently picking the most recent version of every package seems a bit worrisome to me. Oh in particular, we should keep parso 0.8.1 because parso 0.8.2 is incompatible with the current version of jedi.

hoodmane:
Actually I guess we should be using parso-0.7.1
https://github.com/pyodide/pyodide/pull/822

hoodmane:
@rth beautifulsoup4 and soupsieve depend on each other but our dependency resolution can't support it. I think I'll hack this in for now.

rth:
Thanks!
> beautifulsoup4 and soupsieve depend on each other but our dependency resolution can't support it.

yes, that circular dependency is pretty strange. I think the current workaround is fine.

PR #1461, opened 2021-04-13
Store artifacts in build-core CI job

rth:
Following #1453 also store artifacts from the core CI job so we can test a PR in a REPL without waiting for all packages to build.


rth:
This also works, the REPL is then accessible only with core packages https://217-140441556-gh.circle-artifacts.com/0/root/repo/build/console.html

PR #1465, opened 2021-04-13
mkpkg update improvements

hoodmane:
This fixes some issues with mkpg update discovered in #1460. Now update preserves yaml key order. Also, update-all will not update a package if it has patches. I also added some colors to the terminal output of update-all to make it easier to see what it did. I set it up to catch a few common errors and remove the stack traces.

Updating a package still normalizes indentation in meta.yml but I think this is harmless. I applied the new update-all so it is clear what it does.

This added a dependency on `ruamel.yaml` which is a yaml library that avoids reordering keys. I did not add it to any build dependencies, I'm not sure where they go.


pyodide_build/io.py
```diff
@@ -127,10 +127,12 @@ def parse_package_config(path: Path, check: bool = True) -> Dict[str, Any]:
     """
     # Import yaml here because pywasmcross needs to run in the built native
     # Python, which won't have PyYAML
```
rth, line 129: Comment needs updating

pyodide_build/mkpkg.py
```diff
@@ -83,37 +121,45 @@ def update_package(package: str):
         print(f"Skipping: {package} is a local package!")
         sys.exit(0)
 
-    if set(yaml_content.keys()).difference(
-        ("package", "source", "test", "requirements")
-    ):
```
rth, line 88: Yes, this wasn't too reliable

pyodide_build/mkpkg.py
```diff
@@ -123,6 +169,11 @@ def make_parser(parser):
 complex things.""".strip()
     parser.add_argument("package", type=str, nargs=1, help="The package name on PyPI")
     parser.add_argument("--update", action="store_true", help="Update existing package")
+    parser.add_argument(
+        "--update-if-not-patched",
```
rth, line 173: Maybe rather `--update-if-patched` (i.e. don't update patched packages by default) which is a safer default choice and matches the current defaults.

pyodide_build/mkpkg.py
```diff
@@ -134,11 +185,17 @@ def make_parser(parser):
...
+    try:
+        package = args.package[0]
+        if args.update:
+            update_package(package, update_patched=True)
+            return
+        if args.update_if_not_patched:
+            update_package(package, update_patched=False)
+            return
+        make_package(package, args.version)
+    except MkpkgFailedException as e:
```
rth, line 197: Not sure about this. If something fails, I would rather get the actual exception rather than a simple error message saying that it failed.

rth (commented):
Thanks! A few comments otherwise LGTM

pyodide_build/mkpkg.py
hoodmane: Well I kept this, I just moved it down because I want it to report whether the package is out of date or not even if it can't be updated.

pyodide_build/mkpkg.py
hoodmane: I only caught two types of errors this way:
1. When the request fails, it produces messages like:
Failed to load metadata for libxslt from https://pypi.org/pypi/libxslt/json: HTTP Error 404: Not Found

2. If there is no sdist it prints an error message like:
No sdist URL found for package swiglpk (https://pypi.org/project/swiglpk/)

All unexpected exceptions are not a `MkpkgFailedException` and so raise a different exception. I just want to avoid cluttering the output with tracebacks from common failures.

rth: OK, fair enough. Thanks for the additional information.

hoodmane: I'll add this as a comment.

pyodide_build/mkpkg.py
hoodmane: I think when using `mkpkg` to update just one package it's better if it just updates and prints the warning? If you do `./bin/pyodide mkpkg update whatever_package` and then can't figure out how to migrate the patches you can always `git checkout packages/whatever_package` to restore it. It's really just when updating in bulk that it's a nuisance if it updates the patched ones.

pyodide_build/io.py
hoodmane: I don't really understand this comment, I guess I'll just delete it...

We could definitely improve the situation with build dependencies. Probably we should at least add a suggestion to
https://pyodide.org/en/0.17.0a2/development/building-from-sources.html#build-using-make
to look in the Dockerfile for the guaranteed up to date build dependencies?

pyodide_build/mkpkg.py
rth: Agreed, I haven't done that in a while and forgot about that use-case. Lately I was mostly using `update-all`.

rth:
> This added a dependency on ruamel.yaml which is a yaml library that avoids reordering keys

Hmm, I'm not sure if would be a good idea to switch from pyyaml to this library completly. Since it's a pure python package the performance doesn't look great https://sourceforge.net/p/ruamel-yaml/tickets/159/ (not that we have that many files to parse) also the fact that the issue tracker is on Sourceforge doesn't help. The other alternative is use it for mkpkg and still rely on pyyaml for parsing files in the build. 

If we go with this second option, we could be fine to mention it in the CLI command description and fail if it's not installed, use `pytest.importorskip('ruamel')` in the tests when necessary, and manually install it in the test-python CI job. Then have something like a `backend` parameter in `parse_package_config` with pyaml being the default.

hoodmane:
Yeah I think it would be logical to only use `ruamel.yaml` for the purpose of modifying existing yml files.

rth:
Should we try to get it in for 0.18?

hoodmane:
I think I addressed the remaining issues:
* I restored `parse_package_config` to use the normal yaml library and only use `ruamel.yaml` for `update_package`. 
* I set up the ci config to install ruamel.yaml for test-python
* I added `pytest.importorskip` to the `update_package` test

pyodide-build/pyodide_build/mkpkg.py
```diff
@@ -73,8 +84,35 @@ def make_package(package: str, version: Optional[str] = None):
...
+    UNDERLINE = "\033[4m"
+
+
+def abort(msg):
+    print(bcolors.FAIL + msg + bcolors.ENDC)
+    sys.exit(1)
+
+
+def warn(msg):
+    warnings.warn(bcolors.WARNING + msg + bcolors.ENDC)
```
rth, line 105: This is reinventing a logging library a bit. It doesn't necessarily make sense to define these utils functions here and not somewhere where they could be used by other scripts. As then one need to check that this renders OK on various terminals which is not certain.

If we want to go there, maybe we should consider either [loguru](https://github.com/Delgan/loguru#pretty-logging-with-colors) or switch this CLI to typer (example of using colors [here](https://typer.tiangolo.com/tutorial/printing/), or some other appropriate library. Something that could be discussed in a separate PR.

Though for this PR it's fine.

rth (approved):
Thanks @hoodmane !

PR #1467, opened 2021-04-13
Remove windows executables from distutils and distlib

### LINKED ISSUE #653: Remove bundled windows executables in distlib
Distlib includes 4 windows executables https://bitbucket.org/pypa/distlib/src/master/distlib/ for around 350kB.  

Removing them would reduce the size of downloads when using micropip (which has distlib as a dependency).

We should likely come up with a way of removing files via a setting in `meta.yaml` without having to apply patches, e.g.,
```yaml
remove:
  - "distlib/*exe"
```

hoodmane:
This resolves #653. I removed another copy of those same executables from the standard library too.

### DIFF
--- a/src/core/main.c
+++ b/src/core/main.c
@@ -51,6 +51,7 @@ initialize_python()
   FAIL_IF_STATUS_EXCEPTION(status);
   config.write_bytecode = false;
   config.install_signal_handlers = false;
+  config.verbose = true;
   status = Py_InitializeFromConfig(&config);
   FAIL_IF_STATUS_EXCEPTION(status);
 
--- a/src/core/main.c
+++ b/src/core/main.c
@@ -51,7 +51,6 @@ initialize_python()
   FAIL_IF_STATUS_EXCEPTION(status);
   config.write_bytecode = false;
   config.install_signal_handlers = false;
-  config.verbose = true;
   status = Py_InitializeFromConfig(&config);
   FAIL_IF_STATUS_EXCEPTION(status);
 
--- a/Makefile
+++ b/Makefile
@@ -53,8 +53,7 @@ build/pyodide.asm.js: \
 		--preload-file src/pyodide-py/_pyodide@/lib/python$(PYMINOR)/site-packages/_pyodide \
 		--exclude-file "*__pycache__*" \
 		--exclude-file "*/test/*"		\
-		--exclude-file "*/tests/*"		\
-		--exclude-file "*.exe"
+		--exclude-file "*/tests/*"
 	date +"[%F %T] done building pyodide.asm.js."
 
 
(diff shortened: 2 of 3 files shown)


src/core/main.c
```diff
@@ -51,6 +51,7 @@ initialize_python()
   FAIL_IF_STATUS_EXCEPTION(status);
   config.write_bytecode = false;
   config.install_signal_handlers = false;
+  config.verbose = true;
```
rth, line 54: Should we remove this before merging?

rth (approved):

src/core/main.c
hoodmane: Yes we really don't want that.

hoodmane:
I looked around for more stuff we could remove from stdlib. One idea would be to remove some of the text encodings, the stdlib defines ~250 text encodings which take up 1.7mb of disk space. If we reduce it to ~20 encodings then that saves 1.5mb, which seems to be around 600kb compressed. Other than that, it's a bit hard to know what we could take out.

rth:
> I removed another copy of those same executables from the standard library too.

Hmm they were included in pyodide artifacts,
```
>>> os.listdir('/lib/python3.8/site-packages/distlib')
['_backport', '__init__.py', 'resources.py', 'scripts.py', 'locators.py', 'index.py', 'database.py', 'version.py', 'wheel.py', 'metadata.py', 'markers.py', 'util.py', 'compat.py', 'manife st.py', 't32.exe', 't64.exe', 'w32.exe', 'w64.exe']
```
but the size decrease of `pyodide.asm.data` is too small to account for it.

Master
```
-rw-r--r-- 1 root root 580K Apr 13 21:22 distlib.data
-rw-r--r-- 1 root root  13K Apr 13 21:22 distlib.js
[..]
-rw-r--r-- 1 root root 5.2M Apr 13 21:21 pyodide.asm.data
-rw-r--r-- 1 root root 3.6M Apr 13 21:22 pyodide.asm.js
-rwxr-xr-x 1 root root  11M Apr 13 21:22 pyodide.asm.wasm
-rw-r--r-- 1 root root  30K Apr 13 21:22 pyodide.js
```

This PR,

```
-rw-r--r-- 1 root root 294K Apr 14 06:01 distlib.data
-rw-r--r-- 1 root root 9.5K Apr 14 06:01 distlib.js
[...]
-rw-r--r-- 1 root root 5.1M Apr 14 06:00 pyodide.asm.data
-rw-r--r-- 1 root root 3.6M Apr 14 06:01 pyodide.asm.js
-rwxr-xr-x 1 root root  11M Apr 14 06:01 pyodide.asm.wasm
```

>  If we reduce it to ~20 encodings then that saves 1.5mb, which seems to be around 600kb compressed.

Nice! As long as it's not difficult to determine which ones to keep.

> Other than that, it's a bit hard to know what we could take out.

I think there is lots of things we can do in that direction, but I would rather leave it to the next release. #646

hoodmane:
> leave it for the next release

Indeed, I'm getting distracted here.

hoodmane:
> Edit I can't see those in
> ...
> Maybe they were already excluded in some other way?

I think I was confused about this. I now believe that the `.exe` filter is not needed, I'll remove it.

PR #1469, opened 2021-04-14
Use packaging instead of distlib in micropip

hoodmane:
This is on top of #1463. Over there we discussed switching to `packaging` which seems to be recommended over `distlib`. Since `packaging` together with it's dependency `pyparsing` are together smaller than `distlib`, even when we remove the executables from `distlib`, and the code using `packaging` is somewhat simpler this seems to be a bit helpful in several ways.

### DIFF
--- a/packages/micropip/test_micropip.py
+++ b/packages/micropip/test_micropip.py
@@ -186,7 +186,29 @@ def test_install_different_version(selenium_standalone_micropip):
     )
     selenium.run_js(
         """
-        pyodide.runPython(`
+        await pyodide.runPythonAsync(`
+            import pytz
+            assert pytz.__version__ == "2020.5"
+        `);
+        """
+    )
+
+
+def test_install_different_version2(selenium_standalone_micropip):
+    selenium = selenium_standalone_micropip
+    selenium.run_js(
+        """
+        await pyodide.runPythonAsync(`
+            import micropip
+            await micropip.install(
+                "pytz == 2020.5"
+            );
+        `);
+        """
+    )
+    selenium.run_js(
+        """
+        await pyodide.runPythonAsync(`
             import pytz
             assert pytz.__version__ == "2020.5"
         `);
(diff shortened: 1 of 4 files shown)


rth (approved):
Great thanks!

Could you also please add a changelog entry for this and the other micropip fixed you made?

packages/micropip/test_micropip.py
```diff
@@ -140,20 +140,10 @@ def test_install_custom_relative_url(selenium_standalone_micropip):
 def test_last_version_from_pypi():
     pytest.importorskip("distlib")
     import micropip
-
-    class Namespace:
-        def __init__(self, **entries):
-            self.__dict__.update(entries)
+    from packaging.requirements import Requirement
 
     # requirement as returned by distlib.util.parse_requirement
```
rth, line 145: This probably need updating

hoodmane:
So just to explain where I'd like to go with this: I think we should update `loadPackage` to require that if it loads a file from `http://some_website/pyodide_packages/blah.js` or `./blah.js` or whatever, there must be a `packages.json` in the same directory. Then we can load the version info (and dependencies) from the `packages.json`. This will allow us to fix the remaining package loading bugs. (If you load a package with `loadPackage`, then `micropip` won't know about it and may install a different version over it in the future, `micropip` isn't aware of the `loadPackage` dependency resolution.)

Then at that point, I think we can add extra logic into `loadPackage` so that it can load and call into micropip if need be. Then we can support installing wheels and packages from `pypi` using `loadPackage` and remove references to micropip from the docs, encouraging people to use `loadPackage` instead. For use from Python we can add `pyodide.load_package` which as just an alias to `loadPackage`.

I think that removing the mental overhead of figuring out the difference between `loadPackage` and `micropip` will help beginners out a lot.

rth:
Could you open an issue with the above comment (or add it to an existing issue)?

Issue #854, opened 2020-12-13
CLI documentation


rth:
It would be good to generate CLI documentation for the `pyodide_build` scripts.

It might also be a good occasion to migrate them to a more modern CLI library such as [click](https://click.palletsprojects.com/en/7.x/) or [typer](https://github.com/tiangolo/typer). Though one issue there might be with pywasmcross.py. As far as I remember when trying to migrate to click two years ago, conditional argument passing in (https://github.com/iodide-project/pyodide/blob/master/pyodide_build/pywasmcross.py#L440-L443) was problematic (and more specifically support of arbitrary arguments there).

Issue #604, opened 2020-01-23
Create a PyPi package for pyodide_build


gatesn:
Similar to https://github.com/iodide-project/pyodide/issues/328

Would make it easier to generate additional packages for Pyodide without building the entire toolchain.

rth:
> Would make it easier to generate additional packages for Pyodide without building the entire toolchain.

At this stage, I'm not sure that having pyodide_build on PyPi would help much. One would still need to build emsdk and cpython, and the main Makefile is still building some of the C dependencies required by packages (though https://github.com/iodide-project/pyodide/issues/713 might help with that).

PR #1471, opened 2021-04-14
DOC Add pyodide_build CLI documentation

rth:
Adds CLI documentation for pyodide_build.  Closes https://github.com/pyodide/pyodide/issues/854

Typically it would be called as `bin/pyodide <subcommand>`. The other, maybe clearer alternative, is `python -m pyodide_build <subcommand>` but because pyodide_build is not yet an installable package (doesn't have a `setup.py`) that only works from the root dir which is confusing.  

Done with https://github.com/gaborbernat/sphinx-argparse-cli, the other alternative was https://github.com/alex-rudakov/sphinx-argparse but it's not longer maintained and the rendering looks better with sphinx-argparse-cli.

- still haven't figured out how to cross-reference sub-commands https://github.com/gaborbernat/sphinx-argparse-cli/issues/7
- the description of sub-commands doesn't seem to be taken into account, will open an issue about it

Rendered docs: https://pyodide--1471.org.readthedocs.build/en/1471/usage/api/pyodide-build-cli.html

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -45,6 +45,8 @@ extensions = [
     "sphinx_js",
     "autodocsumm",
     "sphinx_pyodide",
+    #    "sphinxarg.ext",
+    "sphinx_argparse_cli",
 ]
 
 myst_enable_extensions = ["substitution"]
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -45,7 +45,6 @@ extensions = [
     "sphinx_js",
     "autodocsumm",
     "sphinx_pyodide",
-    #    "sphinxarg.ext",
     "sphinx_argparse_cli",
 ]
 
--- a/docs/development/new-packages.md
+++ b/docs/development/new-packages.md
@@ -28,7 +28,7 @@ load a package's dependencies automatically.
 ## mkpkg
 
 If you wish to create a new package for Pyodide, the easiest place to start is
-with the {ref}`mkpkg tool <pyodide-build-cli>`. If your package is on PyPI, just run:
+with the {ref}`mkpkg tool <pyodide-mkpkg>`. If your package is on PyPI, just run:
 
 `bin/pyodide mkpkg $PACKAGE_NAME`
 
--- a/docs/development/new-packages.md
+++ b/docs/development/new-packages.md
@@ -28,7 +28,7 @@ load a package's dependencies automatically.
 ## mkpkg
 
 If you wish to create a new package for Pyodide, the easiest place to start is
-with the `mkpkg` tool. If your package is on PyPI, just run:
+with the {ref}`mkpkg tool <pyodide-build-cli>`. If your package is on PyPI, just run:
 
 `bin/pyodide mkpkg $PACKAGE_NAME`
 
--- a/docs/requirements-doc.txt
+++ b/docs/requirements-doc.txt
@@ -6,3 +6,4 @@ distlib   # required by micropip
 sphinx-js==3.1
 autodocsumm
 docutils==0.16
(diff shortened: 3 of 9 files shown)


hoodmane:
I think we should recommend for people not to use `buildpkg` or `buildall`, instead use `PYODIDE_PACKAGES="mypkg" make -C packages`. It is difficult to make `buildpkg` work directly. I think the same thing is true for `pywasmcross` -- is there a use case for invoking `pywasmcross` directly? In the future maybe we could make these commands "private" and expose a different set that are more user friendly. I would prefer the syntax `./bin/pyodide buildpkg mypkg` to `PYODIDE_PACKAGES="mypkg" make -C packages` if it worked.

rth:
Absolutely agree for both points.

For the first one, currently https://github.com/gaborbernat/sphinx-argparse-cli/issues/8 is preventing doing that for the docs since description for subcommands is not shown. We could also put it somewhere under development/ folder to mark it more explicitly as not user facing.

hoodmane:
I think for now the simplest fix would be to move `wasmcross`, `buildpkg`, and `buildall` to a `bin/_pyodide` executable. Then `mkpkg` and `serve` are easier to identify as commands that can actually be used to do things (part of the issue is that after learning that `buildpkg` can't be used, people might expect the same to be true of `mkpkg`). 

Later we could get rid of `bin/_pyodide` and switch to just calling Python functions directly.

But maybe while we're at it it would be nice to make a new `bin/pyodide buildpkg mypkg` that does 
```sh
rm -r packages/mypkg/build # or: touch packages/mypkg/meta.yml
make -C packages
```
I guess I'll open a PR with this.

rth:
Generally I agree, however the situation is a bit more complex than that. `bin/pyodide` is a hack to call `python -m pyodide_build` with the correct PYTHON_PATH. So if we create `bin/_pyodide` we would also need to create yet another entrypoint in pyodide_build.  I think the right long term solution, is to make `pyodide_build` an installable Python package and stop manually tinkering around with PYTHON_PATH https://github.com/pyodide/pyodide/issues/604. Then we can remove `bin/pyodide` altogether, and we could then also more easily define private and public CLI endpoints (and maybe use something better than argparse). 

For now since we have more critical things for 0.17.0, I have just added a note to those command that they are private (shown in the description when `bin/pyodide mkpkg --help` is called for instance) and removed them from the CLI documentation.

docs/conf.py
```diff
@@ -45,6 +45,8 @@
     "sphinx_js",
     "autodocsumm",
     "sphinx_pyodide",
+    #    "sphinxarg.ext",
```
rth, line 48: I don't know where this comment is coming from. I don't have it in my local files.

rth: Nevermind. I was checking the wrong file locally. Twice.

hoodmane:
> however the situation is a bit more complex than that

I'm glad you understand what it's doing because I certainly don't. Environment variables are really pretty terrible. Someone pointed out that you wouldn't write an API like
```py
def g():
   global input, output
   # do some work with input
   output = result
```
and then call it like
```py
input = 4
g()
# do stuff with output
```
but that's what we're doing a lot with environment variables.

hoodmane:
> we have more critical things for 0.17.0

Very true.

rth:
Yes, build code is complex and we clearly need to improve it.

Let me know if you have other comment, if not I'm going to merge this. Thanks!

hoodmane:
I'm sure this could be way better, but since we have to prioritize other stuff right now I'm happy to merge as is.

Issue #729, opened 2020-07-20
Intermittent "memory access out of bounds" error when importing sklearn.datasets


6palace:
We are in chrome 83 running pyodide in web workers. Running any command that includes `from sklearn.datasets import load_iris` (with or without pre-loading using `loadPackage`) will intermittently crash with "memory access out of bounds" errors. 

Whether the error occurs is consistent until a browser refresh, meaning if running the command throws this error once, running again will always throw the error, even though it is run in a newly initialized web worker/pyodide environment. this can also be reproduced with any of the sklearn.dataset objects. Can provide more debugging info if needed.

Is this known behavior that can be mitigated?

Attached some stacktraces:
```
python error RuntimeError: memory access out of bounds
    at _free (https://pyodide-cdn2.iodide.io/v0.15.0/full/pyodide.asm.wasm:wasm-function[12336]:0x872828)
    at __PyMem_RawFree (https://pyodide-cdn2.iodide.io/v0.15.0/full/pyodide.asm.wasm:wasm-function[1754]:0x1ade3b)
    at https://pyodide-cdn2.iodide.io/v0.15.0/full/pyodide.asm.wasm:wasm-function[15550]:0x975d11
    at _freechildren (https://pyodide-cdn2.iodide.io/v0.15.0/full/pyodide.asm.wasm:wasm-function[461]:0xff835)
    at _freechildren (https://pyodide-cdn2.iodide.io/v0.15.0/full/pyodide.asm.wasm:wasm-function[461]:0xff789)
    at _freechildren (https://pyodide-cdn2.iodide.io/v0.15.0/full/pyodide.asm.wasm:wasm-function[461]:0xff789)
    at _freechildren (https://pyodide-cdn2.iodide.io/v0.15.0/full/pyodide.asm.wasm:wasm-function[461]:0xff789)
    at _freechildren (https://pyodide-cdn2.iodide.io/v0.15.0/full/pyodide.asm.wasm:wasm-function[461]:0xff789)
    at _freechildren (https://pyodide-cdn2.iodide.io/v0.15.0/full/pyodide.asm.wasm:wasm-function[461]:0xff789)
    at _freechildren (https://pyodide-cdn2.iodide.io/v0.15.0/full/pyodide.asm.wasm:wasm-function[461]:0xff789)
```

```
python error RuntimeError: memory access out of bounds
[...]

Issue #69, opened 2018-07-06
Running package test suites


rth:
In the case of numpy, for instance, [`test_numpy.py`](https://github.com/iodide-project/pyodide/blob/master/test/test_numpy.py) looks very succinct. Is there any   information about how much of the numpy internal test suite passes? (Even if it can't be run at each build due to run times..)

Actually, what would be necessary to run the test suite (assuming it uses Pytest) of a Python package from pyodide?

mdboom:
The Python core is already being tested in this way (using stdlib `unittest`, not anything fancier).  That certainly could (and should) be adapted to run the package test suites as well.

The biggest annoyance with the core test suite was the number of tests that fire off subprocesses to do their work (which obviously can't be done in WebAssembly).

rth:
Looking at `test_python.py:pytest_generate_tests` as far as I understand currently a list of test modules is created with `make_test_list.py`, then each is executed with `selenium.run("test.libregrtest.main([<module_path>])")`.

For projects that use pytest, such approach would be problematic because we would need to parse the per package `conftest.py` etc. I wonder if the simplest approach might be to run pytest collection & evaluation inside selenium. For instance, to run numpy tests, one could do,
```py
def test_numpy_suite(selenium):
   selenium.load_package('pytest')
   selenium.run("import pytest;"
                 "pytest.main(['--pyargs', 'numpy'])")
```

I'm hoping this might also be faster than calling `selenium.run` for each included test as well.

Opened a PR to package pytest in https://github.com/iodide-project/pyodide/pull/95

I'm not sure yet, whether this nested use of pytest might be problematic.

mdboom:
It would be great to run the whole test suite like this, but it would be nice to somehow turn each test in the internal testing to a test at the pyodide test level.  I don't know if that's possible.  At a minimum, we'll want it to display progress to the console as it goes or tools like Circle-CI will assume it's dead and kill it.

rth:
> It would be great to run the whole test suite like this, but it would be nice to somehow turn each test in the internal testing to a test at the pyodide test level.

I wonder if it's possible to use or adapt pytest-xdist for this. For instance, one could imaging running an pytest inside WASM environment, that would communicate with pytest run from the host system: unless I'm mistaken may not be that different from remote/distributed testing.

mdboom:
That seems like a promising approach, but I know nothing about any of the details... :)

rth:
The full output of the numpy 1.15.1 test suite can be found below,
[...]

PR #1474, opened 2021-04-15
Fix memory access out of bounds in numpy + chrome 89

### LINKED ISSUE #1473: Memory access out of bounds numpy + chrome 89
This is a continuation of #1384. The new problem is in `PyArray_Broadcast`. The crash specifically happens on line 1177 here.
https://github.com/numpy/numpy/blob/648fe68f79dfbad21947db4b55873703db0675c4/numpy/core/src/multiarray/iterators.c#L1177

Relevant WAT:
```wat
local.get $var0 ;; mit
local.get $var3 ;; j
i32.const 2
i32.shl
i32.add ;; &mit->it[j]
i32.load offset=152 ;; crash here 24 + 32 * 4 = 152
```
See also the definition of PyArrayMultiIterObject:
https://github.com/numpy/numpy/blob/648fe68f79dfbad21947db4b55873703db0675c4/numpy/core/include/numpy/ndarraytypes.h#L1231

I've been triggering it with 
```py
import numpy as np
np.meshgrid(range(10), range(10))
```
which isn't completely consistent but works frequently.

Luckily the CI catches this: the benchmark evolve has been failing to this bug ever since @rth updated the CI to use chrome 89 
#1462.
https://app.circleci.com/pipelines/github/pyodide/pyodide/2673/workflows/a9aec892-261d-4eb9-bd4c-3fee0f8bd7a5/jobs/27749

Here's a typical stack trace:
```
pyodide.js:433 Pyodide has suffered a fatal error, refresh the page. Please report this to the Pyodide maintainers.
pyodide.js:434 The cause of the fatal error was:
pyodide.js:435 RuntimeError: memory access out of bounds
    at PyArray_Broadcast (<anonymous>:wasm-function[1536]:0xcdab3)
    at <anonymous>:wasm-function[1538]:0xce06f
    at __static_809 (<anonymous>:wasm-function[1542]:0xce5f0)
    at byn$fpcast-emu$__static_809 (<anonymous>:wasm-function[6879]:0x219772)
    at __static_861 (https://cdn.jsdelivr.net/pyodide/dev/full/pyodide.asm.wasm:wasm-function[1971]:0x22cb00)
    at byn$fpcast-emu$__static_861 (https://cdn.jsdelivr.net/pyodide/dev/full/pyodide.asm.wasm:wasm-function[22300]:0x7be366)
    at PyObject_Call (https://cdn.jsdelivr.net/pyodide/dev/full/pyodide.asm.wasm:wasm-function[762]:0x1caeef)
    at _PyEval_EvalFrameDefault (https://cdn.jsdelivr.net/pyodide/dev/full/pyodide.asm.wasm:wasm-function[2757]:0x2a9d0c)
    at byn$fpcast-emu$_PyEval_EvalFrameDefault (https://cdn.jsdelivr.net/pyodide/dev/full/pyodide.asm.wasm:wasm-function[15505]:0x7a5f10)
    at _PyEval_EvalCodeWithName (https://cdn.jsdelivr.net/pyodide/dev/full/pyodide.asm.wasm:wasm-function[2753]:0x2a41e4)
File "/lib/python3.8/site-packages/numpy/lib/stride_tricks.py", line 191 in _broadcast_shape
File "/lib/python3.8/site-packages/numpy/lib/stride_tricks.py", line 264 in broadcast_arrays
File "<__array_function__ internals>", line 5 in broadcast_arrays
File "/lib/python3.8/site-packages/numpy/lib/function_base.py", line 4213 in meshgrid
File "<__array_function__ internals>", line 5 in meshgrid
File "<exec>", line 1 in <module>
File "/lib/python3.8/site-packages/pyodide/_base.py", line 276 in run_async
File "/lib/python3.8/site-packages/pyodide/_base.py", line 419 in eval_code_async
File "/lib/python3.8/site-packages/pyodide/console.py", line 233 in load_packages_and_run
```

hoodmane:
This is an attempt to fix #1473.


hoodmane:
@rth There's a new intermittent bug in build-core =(

```
error: can't create or remove files in install directory

The following error occurred while trying to add or remove files in the
installation directory:

    [Errno 17] File exists: '/root/repo/packages/.artifacts/lib/python'

The installation directory you specified (via --install-dir, --prefix, or
the distutils default setting) was:

    /root/repo/packages/.artifacts/lib/python/
```

rth:
Thanks for trying to fix this!

> error: can't create or remove files in install directory

Never saw this one before so far.

hoodmane:
I think I've identified a general (but tedious) procedure to fix these:
1. Locate failing instruction, it's either a load or a store
2. Locate line of source failing instruction occurred inside of
3. Extract offending load or store into a separate `noinline` function

hoodmane:
Actually I think I can add:
```
__attribute__((noinline))
NPY_NO_EXPORT
void
assign_to_ptr(void *ptr, void* val){
    *ptr = val;
}

__attribute__((noinline))
NPY_NO_EXPORT
void *
read_from_ptr(void *ptr){
    return *ptr;
}
```
and replace most memory accesses in `PyArray_Broadcast` with these two functions.

hoodmane:
@rth I suspect fixing 0.17.0 + chrome 89 + numpy may be a lost cause. Question is, why are these random memory access failures only happening in numpy? Is it loaded in an unusual way? Or maybe these load/store issues aren't just in numpy but also in any C extension in a side module? How much test code do we have that uses C extension modules? 

I think we have good reason to believe that this chrome 89 bug does not affect the core Pyodide since that code has a lot of coverage. I think it's possible that not just numpy but all extension modules are broken. If it really is only numpy and not all extension modules that are affected, then we would need to get to the bottom of why numpy specifically.

One thing that would be useful would be to go through the benchmarks one at a time and see which ones succeed / fail. That probably would provide valuable clues.

rth:
> Question is, why are these random memory access failures only happening in numpy? [...] I think it's possible that not just numpy but all extension modules are broken.

Hmm, well we had some of those before as well e.g. in scipy https://github.com/pyodide/pyodide/issues/729#issue-661500906 even before the emscripten migration. So it's not a completely new occurence, just new in numpy.

> I suspect fixing 0.17.0 + chrome 89 + numpy may be a lost cause.

Yeah, if it's widespread patching is likely not going to be enough. How about emscripten or LLVM options for instance to prevent inlining https://github.com/emscripten-core/emscripten/blob/e91b2274b84cafcad05382f77e4c4b6e1ddc354d/src/settings.js#L292. Those can be added for numpy specifically (in meta.yaml) or for all packages.

> How much test code do we have that uses C extension modules?

We can run package test suites inside pyodide in REPL. For instance,
```
>>> import pytest
>>> import regex
>>> pytest.main(['--pyargs', 'regex', '--continue-on-collection-errors'])
```
to run tests for the `regex` package (which has C extensions) and produces in Chrome
```
[31m[1m===================== 1 failed, 100 passed in 2.32 seconds =====================[0m
```
So it doesn't look like all C extensions are broken.

For numpy better to run those tests submodule by submodule, as there is no output until all tests are run, and it takes a while. Previously for numpy there weren't that many test failures https://github.com/pyodide/pyodide/issues/69#issuecomment-417360053  We should probably start running those tests more systematically (if not for each commit).  It's also possible to use `-k` flags to filter specific tests as with normal pytest run, so I think that should help us identify parts of the numpy test suite that fail with this error at least.

hoodmane:
> it doesn't look like all C extensions are broken.

Well that's good =)

> We should probably start running those tests more systematically (if not for each commit).

Ideally it'd be nice to run occasional tests on other browsers too. It'd be great to have a larger test suite that would run say every two weeks but only if there's been a new commit. Sounds annoying to set up though.

hoodmane:
> `numpy.core` tests 

I think the core test suite should likely catch these bugs we've been seeing.

hoodmane:
One thing that is more worrisome about this than say #729 is that the errors in #729 are at `free` and `malloc`. These numpy problems occur at normal wasm `load` and `store` instructions, so in principle they could occur at any memory operation. Particularly in `PyArray_Broadcast` the failures seem to be willing to move between different memory operations without any clear pattern. 

If the troubles are only in `PyArray_Broadcast` then I'm pretty confident my current methods will be sufficient with enough patience.

hoodmane:
> How about emscripten or LLVM options for instance to prevent inlining?

Unfortunately these errors are not problems with inlining. The crashes are happening with stuff like:
```C
for(i = 0; i < len; i++){
    max = MAX(array[i], max);
}
for(i = 0; i < len; i++){
    x = array[i]; // crash here
}
```

hoodmane:
@rth The docs say:

> build/skip_host¶
> 
> Skip building C extensions for the host environment. Default: True.
> 
> Setting this to False will result in ~2x slower builds for packages that include C extensions. It should only be needed when a package is a build time dependency for other packages. For instance, numpy is imported during installation of matplotlib, importing numpy also imports included C extensions, therefore it is built both for host and target.

Is there a reason we couldn't use pip to install numpy in the Dockerfile and then just use that for the host numpy rather than building it? I think this could speed up build packages by a lot and also fix all those `import numpy` crashes during the scipy and matplotlib builds.

rth:
> Is there a reason we couldn't use pip to install numpy in the Dockerfile and then just use that for the host numpy rather than building it?

It might work, but I think we still need to patch the configuration https://github.com/pyodide/pyodide/blob/4c315c7c2fb24cdd5c57107094b95190091b08d9/packages/numpy/meta.yaml#L30 

In any case, https://github.com/pyodide/pyodide/issues/1476 would be a solution to that build issue I think.

hoodmane:
@rth Okay this fixes it! (Benchmarks passed in test suite.)

hoodmane:
Looks like there is one more segfault in the chrome packages test suite, this one in `PyObject_GenericGetAttr`.
```C
E           conftest.JavascriptException: RuntimeError: memory access out of bounds
E               at __static_1 (<anonymous>:wasm-function[174]:0x17748)
E               at byn$fpcast-emu$__static_1 (<anonymous>:wasm-function[506]:0x87748)
E               at __static_2 (<anonymous>:wasm-function[175]:0x17eca)
E               at __static_199 (<anonymous>:wasm-function[339]:0x538d5)
E               at byn$fpcast-emu$__static_199 (<anonymous>:wasm-function[704]:0x881f7)
E               at __static_254 (http://127.0.0.1:35323/pyodide.asm.wasm:wasm-function[900]:0x1d3c92)
E               at byn$fpcast-emu$__static_254 (http://127.0.0.1:35323/pyodide.asm.wasm:wasm-function[21695]:0x7bc48e)
E               at _PyObject_GenericGetAttrWithDict (http://127.0.0.1:35323/pyodide.asm.wasm:wasm-function[1712]:0x2183aa)
E               at PyObject_GenericGetAttr (http://127.0.0.1:35323/pyodide.asm.wasm:wasm-function[1711]:0x21815f)
```

rth:
> Okay this fixes it! (Benchmarks passed in test suite.)

Wonderful , great work @hoodmane  !

> Looks like there is one more segfault in the chrome packages test suite

I'll try to run the tests numpy test suite later and will report how it goes. It might be worth adding a comment on how you compute those offsets.

hoodmane:
> It might be worth adding a comment on how you compute those offsets.

Working on cleaning things up now.

hoodmane:
Okay I think this is ready to merge.

hoodmane:
`JsException: ReferenceError: it_dims_k is not defined`

rth:
The good news is as far as I can tell this issue (or at least the original one) is not happening with Chrome 91 (and possibly 90) and with browser auto-update, most users actually [seem to use the latest Chrome version](https://gs.statcounter.com/browser-version-market-share/desktop/worldwide/). So 6 month after 91 release we could very well set it as minimum requirement and remove these patches.

Spend some more time digging through the Chrome issue tracker, and opened a new issue https://bugs.chromium.org/p/chromium/issues/detail?id=1200031 to make sure it doesn't go unreported (if it wasn't already fixed)

rth:
The code looks fine to me, thanks for the documentation effort.

Let me just run a few more numpy tests before merging.

hoodmane:
> opened a new issue https://bugs.chromium.org/p/chromium/issues/detail?id=1200031 to make sure it doesn't go unreported

Thanks, that clearly needed to be done. I added some info there.

hoodmane:
@rth We good to merge this now?

rth:
Run a few numpy broadcasting tests, to check the soundness of the re-implementation,
```py
>>> import numpy
>>> import pytest
>>> pytest.main(['--pyargs', 'numpy', '--continue-on-collection-errors', '-k', 'broadcast'])
```

 - on master (in Firefox 87),
   `128 passed, 8727 deselected, 3 error in 267.95 seconds`
 - in this PR (in Firefox 87),
   `128 passed, 8727 deselected, 3 error in 264.91 seconds`
 - in this PR (in Chrome 89) 
   `128 passed, 8727 deselected, 3 error in 215.32 seconds`

The 3 failures are due to ctypes. To measurable impact on the overall performance either.

Great work @hoodmane !

daoxian:
Still crashed in Chrome  90.0.4430.72.
pyodide.js:432 Pyodide has suffered a fatal error, refresh the page. Please report this to the Pyodide maintainers.
Module.fatal_error @ pyodide.js:432
pyodide.js:433 The cause of the fatal error was:
Module.fatal_error @ pyodide.js:433
pyodide.js:434 RuntimeError: memory access out of bounds
    at PyArray_Broadcast (<anonymous>:wasm-function[1536]:0xcdab3)
    at <anonymous>:wasm-function[1538]:0xce06f
    at __static_809 (<anonymous>:wasm-function[1542]:0xce5f0)
    at byn$fpcast-emu$__static_809 (<anonymous>:wasm-function[6879]:0x219772)
    at __static_861 (https://【my-domain-name】/wasm.master/pyodide.asm.wasm:wasm-function[1972]:0x22cb48)
    at byn$fpcast-emu$__static_861 (https://【my-domain-name】/wasm.master/pyodide.asm.wasm:wasm-function[22302]:0x7be3ba)
    at PyObject_Call (https://【my-domain-name】/wasm.master/pyodide.asm.wasm:wasm-function[763]:0x1caf37)
    at _PyEval_EvalFrameDefault (https://【my-domain-name】/wasm.master/pyodide.asm.wasm:wasm-function[2758]:0x2a9d54)
    at byn$fpcast-emu$_PyEval_EvalFrameDefault (https://【my-domain-name】/wasm.master/pyodide.asm.wasm:wasm-function[15507]:0x7a5f64)
    at _PyEval_EvalCodeWithName (https://【my-domain-name】/wasm.master/pyodide.asm.wasm:wasm-function[2754]:0x2a422c)
Module.fatal_error @ pyodide.js:434
pyodide.asm.js:9 Stack (most recent call first):
pyodide.asm.js:9   File "/lib/python3.8/site-packages/numpy/lib/stride_tricks.py", line 191 in _broadcast_shape
pyodide.asm.js:9   File "/lib/python3.8/site-packages/numpy/lib/stride_tricks.py", line 264 in broadcast_arrays
pyodide.asm.js:9   File "<__array_function__ internals>", line 5 in broadcast_arrays
pyodide.asm.js:9   File "/lib/python3.8/site-packages/numpy/lib/function_base.py", line 4213 in meshgrid
pyodide.asm.js:9   File "<__array_function__ internals>", line 5 in meshgrid
pyodide.asm.js:9   File "/lib/python3.8/site-packages/pysot/tracker/siamrpn_tracker.py", line 85 in generate_anchor
pyodide.asm.js:9   File "/lib/python3.8/site-packages/pysot/tracker/siamrpn_tracker.py", line 59 in __init__
pyodide.asm.js:9   File "<exec>", line 9 in <module>
pyodide.asm.js:9   File "/lib/python3.8/site-packages/pyodide/_base.py", line 270 in run_async
pyodide.asm.js:9   File "/lib/python3.8/site-packages/pyodide/_base.py", line 419 in eval_code_async
pyodide.asm.js:9   File "/lib/python3.8/asyncio/tasks.py", line 280 in __step
pyodide.asm.js:9   File "/lib/python3.8/asyncio/events.py", line 81 in _run
pyodide.js:456 Uncaught RuntimeError: memory access out of bounds
    at PyArray_Broadcast (<anonymous>:wasm-function[1536]:0xcdab3)
    at <anonymous>:wasm-function[1538]:0xce06f
    at __static_809 (<anonymous>:wasm-function[1542]:0xce5f0)
    at byn$fpcast-emu$__static_809 (<anonymous>:wasm-function[6879]:0x219772)
    at __static_861 (https://【my-domain-name】/wasm.master/pyodide.asm.wasm:wasm-function[1972]:0x22cb48)
    at byn$fpcast-emu$__static_861 (https://【my-domain-name】/wasm.master/pyodide.asm.wasm:wasm-function[22302]:0x7be3ba)
    at PyObject_Call (https://【my-domain-name】/wasm.master/pyodide.asm.wasm:wasm-function[763]:0x1caf37)
    at _PyEval_EvalFrameDefault (https://【my-domain-name】/wasm.master/pyodide.asm.wasm:wasm-function[2758]:0x2a9d54)
    at byn$fpcast-emu$_PyEval_EvalFrameDefault (https://【my-domain-name】/wasm.master/pyodide.asm.wasm:wasm-function[15507]:0x7a5f64)
    at _PyEval_EvalCodeWithName (https://【my-domain-name】/wasm.master/pyodide.asm.wasm:wasm-function[2754]:0x2a422c)

hoodmane:
Using `import numpy as np; np.meshgrid(range(10), range(10))` on `https://pyodide-cdn2.iodide.io/dev/full/console.html`? I can't reproduce.

daoxian:
> Using `import numpy as np; np.meshgrid(range(10), range(10))` on https://pyodide-cdn2.iodide.io/dev/full/console.html? I can't reproduce.

OK， you're right. I was using the old version code.

Now, after I've pulled the newest code from git, no crash appears any more.

rth:
Great! Thanks for checking @daoxian !

hoodmane:
> I was using the old version code.

What a relief!

PR #1478, opened 2021-04-17
MAINT Improved Module.fatal_error

hoodmane:
The current `Module.fatal_error` has a few drawbacks:
1. `pyodide._module.runPython` won't work anymore
2. After it fails some actions cause `Error: Object already destroyed` which is unhelpful in the CI.

I also added `assertThrows` to `conftest.py` and added `test_testing` tests which test `assertThrows` and `assert`.

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -120,6 +120,38 @@ class SeleniumWrapper:
                     throw new Error(`Assertion failed: ${cb.toString().slice(6)}${message}`);
                 }
             };
+            window.assertThrows = function assert(cb, errname, pattern){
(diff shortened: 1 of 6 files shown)


hoodmane:
Can see on the CI that the old `Object has already been destroyed` error is replaced with `Pyodide already fatally failed and can no longer be used` which is much more explicit.

rth:
> Can see on the CI that the old Object has already been destroyed error is replaced with Pyodide already fatally failed and can no longer be used which is much more explicit.

Much better, thanks!

Issue #1416, opened 2021-04-02
loadPyodide is not defined


daoxian:
According to the manual: https://pyodide.org/en/latest/usage/quickstart.html

```
<!DOCTYPE html>
<html>
  <head>
      <script src="https://cdn.jsdelivr.net/pyodide/v0.17.0a2/full/pyodide.js"></script>
  </head>
  <body>
    Pyodide test page <br>
    Open your browser console to see Pyodide output
    <script type="text/javascript">
      async function main(){
        await loadPyodide({
          indexURL : "https://cdn.jsdelivr.net/pyodide/v0.17.0a2/full/"
        });
        console.log(pyodide.runPython(`
            import sys
            sys.version
        `));
        console.log(pyodide.runPython(`print(1 + 2)`));
      }
      main();
    </script>
  </body>
</html>
```

It doesn't work.
![image](https://user-images.githubusercontent.com/34263430/113382939-55a25e00-93b5-11eb-8a12-2293a68b118d.png)
What's the problem? @rth @hoodmane

hoodmane:
`loadPyodide` doesn't exist yet in version `0.17.0a2`, to get it use the master version https://cdn.jsdelivr.net/pyodide/dev/full/.

hoodmane:
@rth We really ought to make a more prominent link to the latest stable docs.

hoodmane:
I guess it'd be nice if https://pyodide.org/en/0.16.1/ and https://pyodide.org/en/0.17.0a2/ were valid links, and maybe pyodide.org should better redirect to the docs for v0.16.1.

daoxian:
> `loadPyodide` doesn't exist yet in version `0.17.0a2`, to get it use the master version https://cdn.jsdelivr.net/pyodide/dev/full/.

No, it doesn't work either.
```
<html>
  <head>
      <script src="https://cdn.jsdelivr.net/pyodide/dev/full/pyodide.js"></script>
  </head>
  <body>
    Pyodide test page <br>
    Open your browser console to see Pyodide output
    <script type="text/javascript">
      async function main(){
        await loadPyodide({
          indexURL : "https://cdn.jsdelivr.net/pyodide/dev/full/"
        });
        console.log(pyodide.runPython(`
            import sys
            sys.version
        `));
        console.log(pyodide.runPython(`print(1 + 2)`));
      }
      main();
    </script>
  </body>
</html>

```
[...]

PR #1479, opened 2021-04-17
DOC Make dev documentation consistent + version warnings

rth:
Closes https://github.com/pyodide/pyodide/issues/1416

This aims to make the documentation versions more consistent,
 
- **stable version**. Currently "v0.17.0a2"  https://pyodide.org/en/0.17.0a2/. Will be https://pyodide.org/en/stable/ for next release. Should already be consistent
- Unversionned path: https://pyodide.org now redirects to "v0.17.0a2" in the future will redirect to "stable"  (this can be changed as "Default version" in https://readthedocs.org/dashboard/pyodide/advanced/)
- **dev/latest version**. Uses dev URLs from the CDN, and now includes a banner about it being a development version.
- **older versions**. In the future, will show a banner about them being an outdated version thanks to https://github.com/humitos/sphinx-version-warning

Fixed README to only point to the "v0.17.0a2" (previously there was 1 link to dev out there)

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -28,7 +28,7 @@ import pyodide
 import micropip  # noqa
 
 # The full version, including alpha/beta/rc tags.
-release = version = "0.15.0"  # pyodide.__version__
+release = version = pyodide.__version__
 
 
 # -- General configuration ---------------------------------------------------
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -57,8 +57,8 @@ versionwarning_messages = {
     "latest": (
         "This is the development version of the documentation. ",
         'See <a href="https://pyodide.org/">here</a> for latest stable '
-        "documentation. In particular, do not use Pyodide with non "
-        "versionned (`dev`) URLs from the CDN for deployed applications.",
+        "documentation. Please do not use Pyodide with non "
+        "versioned (`dev`) URLs from the CDN for deployed applications!",
     )
 }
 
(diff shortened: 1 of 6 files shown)


rth:
So overall this should work, however I'm not sure how to test it short of deploying the docs (https://github.com/humitos/sphinx-version-warning/issues/33). Another thing I'm not sure about if for older versions it would link to latest or stable. This needs to be verified before the release (and worse case scenario we could always patch that plugin in a GH fork).

BTW, @hoodmane  it looks like a way to not run build on documentation PRs is to add `[ci skip]` in the commit message. That skips circle CI, but not RTD CI.

docs/conf.py
```diff
@@ -46,12 +46,22 @@
...
 myst_enable_extensions = ["substitution"]
 js_source_path = ["../src/", "../src/core"]
 root_for_relative_js_paths = "../src/"
 
+versionwarning_messages = {
+    "latest": (
+        "This is the development version of the documentation. ",
+        'See <a href="https://pyodide.org/">here</a> for latest stable '
+        "documentation. In particular, do not use Pyodide with non "
+        "versionned (`dev`) URLs from the CDN for deployed applications.",
```
hoodmane, line 61:
```suggestion
        "versioned (`dev`) URLs from the CDN for deployed applications.",
```

hoodmane, line 61:
```suggestion
        "documentation. Please do not use Pyodide with non "
        "versioned (`dev`) URLs from the CDN for deployed applications!",
```

hoodmane (approved):

hoodmane:
Thanks!

PR #1480, opened 2021-04-17
Improve console error handling, add console.html tests

hoodmane:
There are three types of errors that can occur:
Normal error
Syntax error
Fatal Error

Before this PR, console.html correctly handles normal errors but not syntax errors or fatal errors. 
@casatir originally made it handle syntax errors correctly too but then I made simplifications to the code that screwed it up. I opened a PR on the jquery terminal to fix the bug that caused the problem but it isn't going to be merged in time for the release.

I also added tests for console.html.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -7,7 +7,6 @@ defaults: &defaults
   environment:
     - EMSDK_NUM_CORES: 4
       EMCC_CORES: 4
-      PYODIDE_BASE_URL: https://cdn.jsdelivr.net/pyodide/dev/full/
 
 jobs:
   lint:
@@ -95,9 +94,6 @@ jobs:
           paths:
             - .
 
-      - run:
-          name: Prepare CircleCI artifacts
-          command: PYODIDE_BASE_URL="./" make build/console.html
 
       - store_artifacts:
           path: /root/repo/build/
@@ -136,9 +132,6 @@ jobs:
           paths:
             - ./packages/.artifacts
             - ./build
-      - run:
-          name: Prepare CircleCI artifacts
-          command: PYODIDE_BASE_URL="./" make build/console.html
 
       - store_artifacts:
           path: /root/repo/build/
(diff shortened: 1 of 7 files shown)


src/templates/console.html
```diff
@@ -66,11 +90,45 @@
...
+        pyodide._module.on_fatal = async (e) => {
+          term.error("Pyodide has suffered a fatal error. Please report this to the Pyodide maintainers.");
+          term.error("The cause of the fatal error was:");
+          term.error(e);
+          term.error("Look in the browser console for more details.");
+          await term.ready;
+          term.pause();
+          await sleep(15);
+          term.pause();
+        };
```
hoodmane, line 129: I'm not sure what the deal is with the fatal errors, they should be caught on line 67 and handled there but in neither Chrome nor Firefox is it actually caught. So I added this as a temporary hack.

hoodmane:
Our interaction with code.InteractiveConsole is currently pretty awkward. It'd be best to clean that up in the future but it'd take some thought to work out the best way.

hoodmane:
@rth This is weird, this passes in selenium locally and in the `console.html` CircleCI artifact, but consistently fails in the CircleCI test suite. Traceback from the failure shows an out of date `console.py`.

hoodmane:
Oh it must be related to the `indexURL` in `console.html`. Your CI modification to make the artifact version of `console.html` use the built Pyodide doesn't go into the tested version?

.circleci/config.yml
```diff
@@ -239,6 +232,9 @@ jobs:
             tar cjf pyodide-build-${CIRCLE_TAG}.tar.bz2  pyodide/
```
hoodmane, line 232: I think it's better to leave "./" in the github release. People who download the release presumably want to serve their own version, and I think they are better off seeing a 404 than unexpectedly hitting the cdn.

hoodmane:
@rth Could you take a look at the CI config change I made here? I think it's an improvement.

rth:
> Could you take a look at the CI config change I made here? I think it's an improvement.

+1 for those changes.

> I also added tests for console.html.

Looking forward to those!

hoodmane:
Oh wow my jquery terminal PR just landed and they made a new release with it.

Makefile
```diff
@@ -95,6 +95,11 @@ build/webworker_dev.js: src/webworker.js
 	cp $< $@
 	sed -i -e 's#{{ PYODIDE_BASE_URL }}#./#g' $@
 
+update_base_url: \
+	build/console.html \
+	build/pyodide.js \
+	build/webworker.js \
+	docs/_build/html/console.html
```
rth, line 102: We should remove the last line. 

Currently S3 deployment fails with,
```
cp: cannot create regular file 'docs/_build/html/console.html': No such file or directory
make: *** [Makefile:83: docs/_build/html/console.html] Error 1
```

hoodmane: Maybe add a `mkdir -p`?

hoodmane: Makes sense. Will open a PR.

hoodmane: Actually I really don't want to think too hard about when this needs or doesn't need to be updated so I'm gonna `mkdir -p`

rth: I can do the PR. I think the console.html deployed via docs for releases, might also not be correct with the current setup.

hoodmane: Okay, I'll close mine then.

hoodmane: Or you can take it over sure.

PR #1484, opened 2021-04-18
DOCS memory leak avoidance tips

hoodmane:


rth (approved):
Thanks for these explanation in the docs!

docs/usage/faq.md
```diff
@@ -196,3 +196,35 @@ document.body.removeEventListener('click', proxy_f)
...
+  , "headers": { "Content-Type": "application/json" }
+})
+```
+this leaks the dictionary and the `fetch` api ignores the options that we
+attempted to provide. There are two correct ways to do this:
+```py
+from pyodide import to_js
+resp = await js.fetch('example.com/some_api',
+  method= "POST",
+  body= '{ "some" : "json" }',
```
teonbrooks, line 216: does this dictionary also need to be passed to `to_js`?

hoodmane: No, because it's a string. I will update the example though because every time I look at it I think the same thing and have to notice the quote marks. Thanks!

teonbrooks: ah! yeah, it totally got me :)

Issue #1331, opened 2021-03-13
v0.17.0 release checklist


rth:
We should probably make a v0.17.0 release soon, there have been lots of major changes since v0.16.

Please don't hesitate to update the https://github.com/iodide-project/pyodide/milestone/4 tag.

In technical blockers I see,
 - [x] to get a consistent version for type conversions and update the documentation https://github.com/iodide-project/pyodide/pull/1233 What it's left there in your opinion @hoodmane ?
 - [x] make the migration to a separate org https://github.com/iodide-project/pyodide/issues/1217
 - [x] make sure PyProxy and JsProxy methods are documented.

It would have also been nice to move forward on numpy (https://github.com/iodide-project/pyodide/pull/887) / scipy (https://github.com/iodide-project/pyodide/pull/1293) version updates but it is time consuming and I'm not sure it would be done.

hoodmane:
The type conversions stuff is in good shape, when the buffers stuff gets merged I think there's mostly only polish and documentation left to do. I think the polish can be left to future releases. The docs on #1233 are accurate aside from the buffers. One question is what to do with the current `toJs` code for buffers. I would sort of like to get rid of it in favor of the `getRawBuffer` API which is faster and safer, but the `getRawBuffer` API is less user friendly. `getRawBuffer` is really intended to be used with some sort of `ndarray` library (I have confirmation from @daoxian that it works well for this).

I'd like to merge most of my open (non draft) PRs. Other than that, more work needs to be done on docs and it could be nice to hunt down some more memory leaks.

hoodmane:
Oh another concern: Method resolution order for proxies.
Currently Python attributes override PyProxy method overloads always. We haven't explained this in the docs. Also if I make a Python class with a "destroy" field / method, the `PyProxy` destroy method becomes inaccessible, which might not be the greatest thing.

On the other hand, `JsProxy` methods override fields of the same name on the javascript object. It'd probably be good to change this to match the behavior of `PyProxy`?

rth:
Please tag issues/PRs you would like to include in the release so we have an overall view.

> Oh another concern: Method resolution order for proxies.

Yeah, I guess documentation on that would be a start.

hoodmane:
The current master branch is our release candidate for version 0.17.0, scheduled for release this upcoming Thursday. If any of you could try it out before then and give us feedback, it'd be much appreciated!

[...]

PR #1485, opened 2021-04-18
Version 0.17.0 release notes

hoodmane:
WIP


hoodmane:
@rth I started writing some more substantial release notes here but I guess there is a question about precedent, maybe we won't want to put too much effort into release notes in the future.

rth:
I think it's a good idea, even if we don't keep it in the future. It's a bit more pleasant to read than the changelog. We don't necessarily need to expand to much on what's new since that will be in the announcement post. However some description and in particularly for API changes/migration guide, would be helpful.

In terms of page organization what could work, we still name the changelog as "Release notes" (or "What's new"), then from the beginning of the changelog section for 0.17, in the description link to `release-notes/v0.17.0.md` with this content (and at the end of that page link back to the changelog). That way for a future release if don't won't to release notes we won't make them, and there won't be a page where their absence will be problematic. 

As if we have two parallel pages changelog.md and release-notes.md in the side menu, it can be a bit confusing to users which ones they should check, if they want to know what's new in a given release.

hoodmane:
Okay sounds good. I will leave it in a separate file for now, we can rearrange it as you say before we merge.

hoodmane:
One thing I would like to expand on more here is the maintenance because IMO the most important stuff that happened in this release was a huge amount of maintenance. It just doesn't make for the most exciting reading.

Anyways I will mess around writing whatever I feel like and if you think it's not appropriate I will not be offended if we cut it.

hoodmane:
Also, IMO think we're all ready code wise. There's nothing else I want included into the release other than some last few docs updates.

rth:
Sounds good.

> Also, I think we're all ready code wise. There's nothing else I want included into the release other than some last few docs updates

Yeah, we are really close. I wonder if we should tag some active contributors and users in https://github.com/pyodide/pyodide/issues/1331 and ask them to beta test the dev version before the release.

hoodmane:
> tag some active contributors and users in #1331 and ask them to beta test the dev version before the release.

+1 for this.

hoodmane:
@rth Could you skim over what I have here so far? I think the length is pretty much on target. It might be fun to add whatever benchmarks chart(s) you make here too.

docs/_static/css/pyodide.css
```diff
@@ -6,3 +6,14 @@
 code {
     font-size: 100% !important;
 }
+
+
+h3 {
```
rth, line 11: Maybe let's add another sector specific to the page content. Never know where h3 / h4 could be used.

docs/project/release-notes.md
```diff
@@ -0,0 +1,219 @@
...
+```
+(Added in PR {pr}`1146`)
+
+### New Conversion APIs
+We added several new conversion APIs to give more explicit control over the
+foreign function interface. In particular, the goal was to make it easier for
+users to avoid leaking memory.
+
+For the basic use cases, we have {any}`PyProxy.toJs` and {any}`JsProxy.to_py`.
+We also added also "wrong-way" conversion functions {any}`pyodide.to_js` and
```
rth, line 91: It might be helpful to add slightly more explanation about these function as it might not be obvious for the reader.

docs/project/release-notes.md
```diff
@@ -0,0 +1,219 @@
...
+documentation, though gaps remain.
+
+See issue {issue}`900` for some of the discussion.
+
+(Mostly implemented in PRs {pr}`1152` and {pr}`1167`, see also {pr}`1186` which )
+
+### Changes to buffer translations
+
+The old buffer translation code was inflexible, leaky, and had serious bugs
+including use after free ({issue}`749`) and buffer overflow errors.
```
rth, line 134:
```suggestion
The buffer translation code in previous versions was less flexible, leaked memory,
and had serious bugs including use after free ({issue}`749`) and buffer overflow errors.
```
to be a bit more neutral

docs/project/release-notes.md
```diff
@@ -0,0 +1,219 @@
...
+We finally completed the migration to the latest version of Emscripten
+(https://emscripten.org/) compiler toolchain which uses the upstream LLVM
+backend. This allows us to take advantage of  recent improvements to the
+toolchain reducing package size and execution time.
+
+For instance, the scipy package  shrank dramatically from 92 MB to 15 MB. Scipy
+is now automatically cached in browsers, greatly improving the usability of
+scientific Python packages that depend on scipy, such as scikit-image and
+scikit-learn. The size of the base Pyodide environment with only the CPython
+standard library shrank from 8.1 MB to 6.4 MB.
```
rth, line 162: Well to be precise this is more due to not vendoring jedi than to Emscripten update, but I wasn't sure how and where to phrase in the current narrative.

docs/project/release-notes.md
```diff
@@ -0,0 +1,219 @@
...
+The core C code base was improved with new macros that streamline the most
+common tasks and help with consistency, and in addition to the extensive
+additions to the test suite we performed a manual audit of the entire C codebase
+to locate leaks and logic errors.
+
+We fixed many of the long standing bugs caused by inconsistencies in the
+behavior of {any}`JsProxy`. There used to be two different code paths for
+producing a {any}`JsProxy` and two different code paths for calling a
+{any}`JsProxy` leading to four different behaviors, all of which were buggy in
+some cases. This fixed numerous bugs, including issues {issue}`461`,
```
rth, line 183:
```suggestion
{any}`JsProxy` leading to four different behaviors, all of which were presented
errors in some cases. This fixed numerous bugs, including issues {issue}`461`,
```

rth (commented):
It's a nice read. A few minor notes below.

I think we can merge after changing the path `docs/project/release-notes.md` to `docs/project/release-notes/v0.17.0.md` (and the page title accordingly) and linking to it from the changelog.

docs/project/release-notes.md
hoodmane: > more due to not vendoring jedi

That's a bit disappointing.

rth: Yeah, I mean it does reduce the size of the generated wasm but in base pyodide by <0.5MB or so, but fair amount of size is due to `pyodide.asm.data` anyway.

docs/_static/css/pyodide.css
hoodmane: Well here again I looked at the selector that was determining the size of the headers and it's just an `h3` by itself. But it seems that I can do `.section > h3` which is probably better.

PR #1487, opened 2021-04-19
Use sphinx issues to link prs / issues

hoodmane:
A few links went to the wrong place because of the repeated number, or were mixed up whether they were /pull/### or /issues/###. This declutters the changelog and makes it less likely to make these sorts of errors.


rth:
Very nice improvement! Particularly for the changelog.

PR #1492, opened 2021-04-19
MAINT mkdir -p docs/_build/html before trying to copy there

hoodmane:
@rth this okay?


rth:
From https://github.com/pyodide/pyodide/pull/1480#discussion_r615945190

> I think the console.html deployed via docs for releases, might also not be correct with the current setup, so I need to fix that.

I'll push a few fixes here.

PR #1494, opened 2021-04-19
MAINT xfail remaining Chrome 89 fatal errors

hoodmane:
See #1493 and #1495.

### DIFF
--- a/packages/pandas/test_pandas.py
+++ b/packages/pandas/test_pandas.py
@@ -51,6 +51,10 @@ def test_extra_import(selenium, request):
 @pytest.mark.skip_refcount_check
 def test_load_largish_file(selenium_standalone, request, httpserver):
     selenium = selenium_standalone
+    if selenium.browser == "chrome":
+        pytest.xfail(
+            "test_load_largish_file triggers a fatal runtime error in Chrome 89 see #1495"
+        )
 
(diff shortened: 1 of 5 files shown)


packages/pandas/test_pandas.py
```diff
@@ -51,6 +51,10 @@ def test_extra_import(selenium, request):
 @pytest.mark.skip_refcount_check
 def test_load_largish_file(selenium_standalone, request, httpserver):
     selenium = selenium_standalone
+    if selenium.browser == "chrome":
+        pytest.xfail(
+            "test_load_largish_file triggers a fatal runtime error in Chrome 89 see #1495"
```
rth, line 56: FWIW, this also used to fail a long time ago in Chrome.

rth (approved):
Thanks!

packages/pandas/test_pandas.py
hoodmane: Really weird stuff, see #1495.

hoodmane:
`test_numcodecs` also is a persistent failure due to recursion errors, I think if I xfail that too then packages-chrome can pass. It makes sense that chrome would have more recursion errors since it's stack is 2.5x smaller than the firefox stack.

PR #1500, opened 2021-04-19
Reduce pyodide.asm.js by factor of 2 with sed

hoodmane:
I stripped out C++ names with:
`sed -i 's/var __ZN[1KS][^;]*;//g' build/pyodide.asm.js`

Turns out that reduces pyodide.asm.js from 3.6Mb to 1.9Mb.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -55,7 +55,14 @@ build/pyodide.asm.js: \
 		--exclude-file "*__pycache__*" \
 		--exclude-file "*/test/*"		\
 		--exclude-file "*/tests/*"
-	sed -i 's/var __ZN[1KS][^;]*;//g' build/pyodide.asm.js
+	# Strip out C++ symbols which all start __Z.
+	# There are 4821 of these and they have VERY VERY long names.
+	# Reduces size of pyodide.asm.js by a factor of 2.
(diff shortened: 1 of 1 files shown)


rth:
Cool! Note however that `pyodide.asm.js` is only 361 KB once compressed (see the network tab when loading the REPL from the CDN). So large files locally are not necessarily large to download. It would still take time to process them I imagine though.

rth:
Could you add a comment how you deteremined which symbols to strip please as well?

hoodmane:
@rth How do I tell how big something will be from the CDN?

hoodmane:
240K now brotli compressed using defaults: `brotli build/pyodide.asm.js`. 120kb isn't as exciting as 1.8MB but it's still something.

rth:
It's the "Transferred" column if you click on "Disable cache",

image

and yes `brotli` applied to that file.

rth:
> 120kb isn't as exciting as 1.8MB but it's still something.

Absolutely, I may not be download time so much, but I imagine parsing 1.8MB of JS still takes time. So it's good to reduce it.

hoodmane:
In any case it's much lower hanging fruit to cut down `pyodide.asm.js` because it's a text file.

PR #1502, opened 2021-04-20
DOC Add JupyterLite to related projects

rth:
cc @jtpio How experimental is it? :)


jtpio:
Still an early prototype for now, but actively iterating on it.

Thanks for adding it!

PR #1505, opened 2021-04-20
Add "Try it online" section to the docs

jtpio:
There is already a link to the interactive console in the README.md to try Pyodide in the browser without installing anything:

https://github.com/pyodide/pyodide#try-pyodide-no-installation-needed

It would be nice to also have one in the docs, for folks landing on https://pyodide.org directly.

And with the latest docs, the console seems to be reachable with: https://pyodide.org/en/latest/console.html

### DIFF
--- a/docs/usage/quickstart.md
+++ b/docs/usage/quickstart.md
@@ -2,6 +2,10 @@
 
 # Getting started
 
+## Try it online
+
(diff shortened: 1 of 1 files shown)


hoodmane:
Much appreciated @jtpio!

PR #1509, opened 2021-04-20
BLD Display package build time in logs

rth:
Ref https://github.com/pyodide/pyodide/issues/936#issuecomment-823521369


rth:
In the end it's not clear that there is any reduction in build time with this PR, but it's probably a good idea to reduce CPU over-subscription anyway.

PR #1510, opened 2021-04-20
DOC Disable sphinx versionwarning extension

rth:
Temporarily disables sphinx versionwarning introduced in https://github.com/pyodide/pyodide/pull/1479

It should have shown a banner on the latest version but it didn't. Similarly I would like to make sure it doesn't mark 0.17.0 as outdated with respect to latest. Since it's not very straightforward to test locally and I didn't get an answer in https://github.com/humitos/sphinx-version-warning/issues/33 , better to exclude it from the release.

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -46,7 +46,8 @@ extensions = [
     "autodocsumm",
     "sphinx_pyodide",
     "sphinx_argparse_cli",
-    "versionwarning.extension",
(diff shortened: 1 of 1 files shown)


rth:
> It should have shown a banner on the latest version but it didn't

Actually the latest doc was probably just not deployed, since it required to update RTD settings following the renaming of the `master` branch.

PR #1511, opened 2021-04-20
DOC Add in missing api docs by monkey patching autodoc

hoodmane:
I monkey patched `get_module_members` to find apis in submodules. Not necessarily the ideal approach, but it was the first thing I found that worked.


rth (approved):
Thanks! As as quickfix that works. We should probably find something not involving monkeypatching in the long run.

hoodmane:
It's really hard to avoid monkey patching with sphinx though: the APIs provided are extremely inflexible.

PR #1515, opened 2021-04-21
Release 0.17.0

rth:
Final changes for the 0.17.0 release. We really should automate version replacements in the docs.

Depends on https://github.com/pyodide/pyodide/pull/1514 fixing the Firefox CI.

Closes https://github.com/pyodide/pyodide/issues/1331


hoodmane:
I'm going to open a PR about #1516 and then another PR putting a link to that PR in the `InteractiveConsole` docs so people will know that any code that touches it will break in v0.17.1.

hoodmane:
Guess I'll grep to see if I can catch any suspicious looking version data anywhere.

rth:
Yes, I think it should be fine, but if you could double check it would be good.

hoodmane:
One minute.

daoxian:
Great work! It's a big step forward.

PR #1519, opened 2021-04-21
API Rework _base.py and console.py

### LINKED ISSUE #1516: InteractiveConsole API is not ready for public use
Trying to subclass the `InteractiveConsole` from `pyodide.console`, it looks like the signature of the `runcode` method might be misleading.

With `0.17.0a2`, the parameter `code` would be pass to the parent for execution and be taken into account:

https://github.com/pyodide/pyodide/blob/afb522ed59a08dc7edd0ae7935ec2999d55ed544/src/pyodide-py/pyodide/console.py#L228

With the latest (fetching from https://pyodide-cdn2.iodide.io/dev/full/pyodide.js), it doesn't seem to be the case anymore with the `code` parameter being ignored:

https://github.com/pyodide/pyodide/blob/60100c8fc66d3acffdca79e6ff3f1d4dd2200a09/src/pyodide-py/pyodide/console.py#L212-L225

Maybe the  `runcode` method could simply be replaced by the following?

```python
def runcode(self, code):
    self.run_complete = ensure_future(
        self.load_packages_and_run(self.run_complete, code)
    )
```

hoodmane:
Fixes #1516.
WIP.
Opening this now so I can put a link into v0.17.0 docs.

### DIFF
--- a/packages/micropip/micropip/micropip.py
+++ b/packages/micropip/micropip/micropip.py
@@ -11,12 +11,14 @@ from packaging.requirements import Requirement
 from packaging.version import Version
 from packaging.markers import default_environment
 
-from pyodide._core import IN_BROWSER
-from pyodide import to_js
-
 # Provide stubs for testing in native python
-if IN_BROWSER:
+try:
     import pyodide_js
+    from pyodide import to_js
+
+    IN_BROWSER = True
+except ImportError:
+    IN_BROWSER = False
 
 if IN_BROWSER:
     # In practice, this is the `site-packages` directory.
--- a/packages/micropip/micropip/micropip.py
+++ b/packages/micropip/micropip/micropip.py
@@ -11,14 +11,12 @@ from packaging.requirements import Requirement
 from packaging.version import Version
 from packaging.markers import default_environment
 
-# Provide stubs for testing in native python
-try:
-    import pyodide_js
-    from pyodide import to_js
+from pyodide._core import IN_BROWSER
+from pyodide import to_js
 
-    IN_BROWSER = True
-except ImportError:
-    IN_BROWSER = False
+# Provide stubs for testing in native python
+if IN_BROWSER:
+    import pyodide_js
 
 if IN_BROWSER:
     # In practice, this is the `site-packages` directory.
--- a/packages/pillow/test_pillow.py
+++ b/packages/pillow/test_pillow.py
@@ -1,9 +1,9 @@
-import pytest
 from pyodide_build.testing import run_in_pyodide
 
 
-@pytest.mark.xfail
-@run_in_pyodide(packages=["pillow"])
+@run_in_pyodide(
+    packages=["pillow"], xfail_browsers={"firefox": "timeout", "chrome": ""}
+)
 def test_pillow():
     from PIL import Image, ImageDraw, ImageOps
     import io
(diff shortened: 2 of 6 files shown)


rth:
Thanks for this works. 

That's a fairly large code diff and a number of backward compatible API changes. What's the motivation behind it aside from fixing https://github.com/pyodide/pyodide/issues/1516 ? For that issue we could very well say that `runcode` is not exposed and users have to use the `push` method otherwise.

hoodmane:
This is just scratch work, usually I would put things back / select a subset of the changes that are actually good before opening a PR, but I want to link it in the docs.

hoodmane:
But my motivation is to simplify the code and add features. In particular, I would like to be able to use the `_base` AST manipulation with `code.InteractiveConsole`, but without parsing the code twice and compiling it twice. I think the fact that we have to recompute a bunch of stuff is a sign of a poor design.

hoodmane:
To avoid repeated compilation, we need something like the`parse_and_compile` method I've added here.

hoodmane:
Two of the problems are:
1. `code.InteractiveConsole` doesn't pass through enough data
2. `eval_code` and `eval_code_async` are not very well designed

Together, 1 & 2 force pervasive use of class-scoped variables. Note that only instance of `InteractiveConsole` is expected to exist at most times, since `InteractiveConsole` is effectively a singleton so all of these class-scoped variables are actually global variables in disguise.

hoodmane:
Another issue is that `InteractiveConsole` is too opinionated about output. When it gets a result or error, it should pass it directly to the user rather than automatically formatting a traceback or printing it. We should move the more opinionated part to a subclass.

rth:
For functions that were added in this release, if you have concerns, it might be better to make them private before the release. That would make it more explicit that changes are likely in the future.

hoodmane:
Working on that. `CodeRunner` and `InteractiveConsole` are not ready, I'm going to strip the docs and add a message.

`eval_code` and `eval_code_async` won't change their API, only implementation.

rth:
> I'm going to strip the docs and add a message.

Making them private (e.g. `_InteractiveConsole`) and removing them from `__all__` so they are not part of the documentation might be better. In 0.16 we had only [1 function in console.py](https://github.com/pyodide/pyodide/blob/0d10cfaded5586ffc21ed38c7b2d392490871c6e/src/pyodide-py/pyodide/console.py)

hoodmane:
Okay. (Actually docs unconditionally ignore things with a leading semicolon due to my monkeypatch.)

rth:
Sounds good. Let me know if you need help there. I'd like to release so we can change the documentation links in the announcement post.

hoodmane:
> Let me know if you need help there.

Yeah if you did it that would be good. As I said, I think it's just `InteractiveConsole` and `CodeRunner` that need to be private.

hoodmane:
Wow, the way the Python REPL works is pretty darn janky:
<details>
<summary>Details</summary>

> Approach:
> 
> First, check if the source consists entirely of blank lines and
> comments; if so, replace it with 'pass', because the built-in
> parser doesn't always do the right thing for these.
> 
> Compile three times: as is, with \n, and with \n\n appended.  If it
> compiles as is, it's complete.  If it compiles with one \n appended,
> we expect more.  If it doesn't compile either way, we compare the
> error we get when compiling with \n or \n\n appended.  If the errors
> are the same, the code is broken.  But if the errors are different, we
> expect more.  Not intuitive; not even guaranteed to hold in future
> releases; but this matches the compiler's behavior from Python 1.4
> through 2.2, at least.

</details>

PR #1520, opened 2021-04-21
API loadPackage handling for Python package list

### LINKED ISSUE #1517: Can't call loadPackage from Python
```python
pyodide.runPython("import js ; js.pyodide.loadPackage(['numpy'])")
```
gives an error:

`Skipping unknown package '['numpy']'`

This is because `loadPackage` does not check for `PyProxy` input.

Doing

``` javascript
window.loadPackage = async function (packages) {
    if( pyodide.isPyProxy(packages) ) packages = packages.toJs();
    return await pyodide.loadPackage(packages);
};
```

then calling this `js.loadPackage` from Python seems to fix this issue. (I did not take care for extra parameters in `loadPackage` in this example)

I can submit a PR if needed.

casatir:
This is an attempt to fix #1517.

### DIFF
--- a/src/pyodide.js
+++ b/src/pyodide.js
@@ -287,6 +287,16 @@ globalThis.loadPyodide = async function(config = {}) {
    * @async
    */
   Module.loadPackage = async function(names, messageCallback, errorCallback) {
+    if( Module.isPyProxy(names) ) {
+      let temp;
+      try {
+        temp = names.toJs();
+      } finally {
+        names.destroy();
+      }
+      names = temp;
+    }
+
     if (!Array.isArray(names)) {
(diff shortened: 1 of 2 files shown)


hoodmane (approved):
Thank you Casatir! This is definitely an improvement.

Going forward, perhaps we should have a standard way to mark APIs like `JsProxy.then` and now `loadPackage` that handle conversion / reference counting, and also make a consistent effort to increase the number of APIs that do.

hoodmane:
@casatir Please fix lints. When CI finishes I'll merge it (unless something weird happens).

hoodmane:
Thanks!

PR #1523, opened 2021-04-21
Bump main to 0.18.0dev0

rth:
We really should automate some of these replacements in the docs. The issue last time I tried was that sphinx replacements didn't work inside code blocks.

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -51,8 +51,7 @@ extensions = [
     "autodocsumm",
     "sphinx_pyodide",
     "sphinx_argparse_cli",
-    #  TODO: Temporary disabling for the 0.17.0 release, needs more investigation
-    #     "versionwarning.extension",
+    "versionwarning.extension",
     "sphinx_issues",
 ]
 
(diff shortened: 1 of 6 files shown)


hoodmane:
I can write a shell script if you like?

rth:
> I can write a shell script if you like?

I think https://pypi.org/project/bumpversion/ should work for this. We already have a partial config in `setup.cfg`, we just need to start using it. And maybe pick a maintained fork since it's no longer maintained.

hoodmane:
Darn that would have been fun.

rth:
OK, version warning failed building docs on main with,
```
  File "/home/docs/checkouts/readthedocs.org/user_builds/pyodide/envs/latest/lib/python3.8/site-packages/sphinx/events.py", line 111, in emit
    results.append(listener.handler(self.app, *args))
  File "/home/docs/checkouts/readthedocs.org/user_builds/pyodide/envs/latest/lib/python3.8/site-packages/versionwarning/signals.py", line 40, in generate_versionwarning_data_json
    message=message.format(
AttributeError: 'tuple' object has no attribute 'format'
```
and I can't reproduce locally so far.

PR #1525, opened 2021-04-22
DOCS Add faq on redirecting staandard streams

hoodmane:
I meant for this to go into 0.17.0 but it got lost.


rth:
Thanks ! You can manually cherry-pick this commit to `stable` upstream branch if you want it to be included in the stable documentation branch.

hoodmane:
Okay I will do that.

hoodmane:
I guess then if we make an 0.17.0 examples folder we could cherry-pick that to stable too.

hoodmane:
Btw I think the reason I couldn't commit to other peoples' PRs may have been because my upstream remote was using https rather than ssh.

PR #1531, opened 2021-04-22
MAINT hide jsdoc_dump.json from grep results

rth:
Lately `docs/sphinx_pyodide/tests/jsdoc_dump.json` is frequently returned by grep, and because it has few line breaks it also returns large chunks of that json. The same would apply for search on Github.

This gzip compress it, so it would be considered a binary file by search tools and therefore ignored.

I run tests locally. Skipping CI as it's a bit wasteful to run 2h of CI (1 in PR and 1 on main) for such changes. I really need to continue https://github.com/pyodide/pyodide/pull/1262


hoodmane:
Thanks, it is really annoying.

PR #1532, opened 2021-04-22
Convert eval_code helper methods to pure functions

hoodmane:
This is a first step towards doing the stuff in #1519.

I tried to change as little as possible, the main changes are converting `self.some_field` accesses to function arguments. I also lightly refactored to reduce the number of fields that `should_quiet` and `_split_and_compile` need access to.

Most of the diff comes from reordering the run and run_async methods.

### DIFF
--- a/src/pyodide-py/pyodide/_base.py
+++ b/src/pyodide-py/pyodide/_base.py
@@ -161,113 +161,113 @@ class CodeRunner:
         )
 
 
-if True:
+def should_quiet(source: str) -> bool:
+    """
+    Should we suppress output?
 
-    def should_quiet(source: str) -> bool:
-        """
-        Should we suppress output?
+    Returns ``True`` if ``quiet_trailing_semicolon`` is set to ``True`` and
+    the last nonwhitespace character of ``code`` is a semicolon.
 
-        Returns ``True`` if ``quiet_trailing_semicolon`` is set to ``True`` and
-        the last nonwhitespace character of ``code`` is a semicolon.
+    Examples
+    --------
+    >>> should_quiet('1 + 1')
+    False
+    >>> should_quiet('1 + 1 ;')
+    True
+    >>> should_quiet('1 + 1 # comment ;')
+    False
+    """
+    # largely inspired from IPython:
+    # https://github.com/ipython/ipython/blob/86d24741188b0cedd78ab080d498e775ed0e5272/IPython/core/displayhook.py#L84
 
-        Examples
-        --------
-        >>> should_quiet('1 + 1')
-        False
-        >>> should_quiet('1 + 1 ;')
-        True
-        >>> should_quiet('1 + 1 # comment ;')
-        False
-        """
-        # largely inspired from IPython:
-        # https://github.com/ipython/ipython/blob/86d24741188b0cedd78ab080d498e775ed0e5272/IPython/core/displayhook.py#L84
+    # We need to wrap tokens in a buffer because:
+    # "Tokenize requires one argument, readline, which must be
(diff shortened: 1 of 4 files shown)


rth:
Could you please expand a bit more on the motivation for these changes?

In the end whether the class has the implementation is called inside the function or the opposite should be more or less equivalent? 
Personally I'm all for using more functions and less classes, but this is somewhat opposite to the changes we did in https://github.com/pyodide/pyodide/pull/1041 and it would be good to avoid code churn. 

Code wise this looks good  (I think you can remove the "if True" even if it results in a larger diff due to indentation) but we need to all agree it's the way forward for those functions.

cc @casatir

hoodmane:
I don't think these particular changes do all that much or are all that important. I just find the code easier to work with in this form.

I think `should_quiet(source)` is a much better function than `quiet(self, source)` which returns 
`not self.quiet_trailing_semis and should_quiet(source)`. Similarly, pulling the `return_mode` out of `_split_and_compile` is an improvement. These functions each do fewer things: inner functions unconditionally do one thing and then outer functions check options before 

I wanted these changes in #1041 but I was having a hard time communicating with Casatir and I figured I could try to redo it later.

hoodmane:
There are a few issues which aren't resolved in this PR, but are resolved in my branch here
https://github.com/hoodmane/pyodide/tree/base-refactor
I am planning to come back and make a second PR fixing these after.

1. No way for caller to do additional ast transformations without duplicating our code.
2. No way for caller to know when parsing is finished and execution is starting.
3. Currently
`eval_code_async("x = async_func()", globals())`
vs
`x = eval_code_async("async_func()")`
assign different values to x. In the first case x is a coroutine, in the second case it's the result of doing the coroutine.

rth:
OK, sounds good to me thanks. Could you please fix lint and test-python CI and remove the extra indentation due to `if True`?

hoodmane:
While we're at it, could we remove CodeRunner entirely? I don't really see that it's necessary.

rth:
> While we're at it, could we remove CodeRunner entirely?

We are indeed not using it, and it is private but Basthon for instance uses it a lot. Let's open a separate issue about it. In the current state I agree, it doesn't add much value, what I'm not sure whether it's still the best abstraction in the long term if there are more features. Didn't we adapt it from IPython or some such project? We just definitely want avoid a situation where we remove it now to re-add it later and I think it should be a separate discussion from this PR.

hoodmane:
> a separate discussion

Sounds good.

Though IMO if anyone needs this it's a completely trivial addition:

```py
import functools
from pyodide import eval_code, eval_code_async
class CodeRunner:
	def __init__(self, **kwds):
       self.run = functools.partial(eval_code, **kwds)
       self.run_async = functools.partial(eval_code, **kwds)
```

hoodmane:
I guess for instance the `InteractiveConsole` wants to persist the future flags:
`from __future__ import some_feature` does roughly `co_flags |= PyCf_SOME_FEATURE`. In particular, unlike other imports this doesn't affect `globals()`, it directly adjusts the code object created. This has an unintuitive result that
```py
ns = {}
exec("import some_pkg", ns)
exec("<<use some_pkg here>>", ns)
```
works fine but
```py
ns = {}
exec("from __future__ import some_feature", ns)
exec("<<use some_feature here>>", ns)
```
doesn't.

I think it might be a good idea to adjust `eval_code` to load the future flags from the `globals()` when it starts and write back any extra future flags when it's done to minimize chance of confusion. Maybe store it to `__active_future_flags` on the global namespace or something.

rth:
Let's move the above discussion to a separate issue :)

PR #1534, opened 2021-04-22
Use same image when deploying dev as when deploying release

hoodmane:
For both v0.16.0 and v0.17.0 we had issues with the deploy-release recipe. Problem is that we don't run it very often, so we don't know if it works except when we really care whether it works.

In lieu of getting better test coverage for the deploy-release recipe, we can just make the deploy-dev recipe really similar. So we should use the same docker images, install the same requirements, and as far as practically possible execute the same commands.

I added a `ghr -draft` to the deploy-dev recipe. I have no idea if this is actually a good idea, but it seemed like it would be appropriate for the use case of making sure that the rest of the command would succeed.

https://github.com/tcnksm/ghr/#github-api-token
https://docs.github.com/en/github/administering-a-repository/managing-releases-in-a-repository

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -348,7 +348,7 @@ workflows:
               ignore: /.*/
             tags:
               only: /^\d+\.\d+\.\w+$/
-      - deploy-s3:
+      - deploy-dev:
           requires:
             - lint
             - test-python
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -265,8 +265,7 @@ jobs:
       - run:
           name: Install requirements
           command: |
-            sudo apt-get update
-            sudo apt-get install -y groff
+            apk add --no-cache --update python3 make
             python3 -m pip install awscli
       - run:
           name: Deploy to pyodide-cdn2.iodide.io
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -211,11 +211,11 @@ jobs:
           path: /root/repo/build/benchmarks.json
 
   deploy-release:
+    # To reduce chance of deployment issues, try to keep the steps here as
+    # similar as possible to the steps in deploy-release!
     resource_class: small
 
     docker:
-      # To reduce chance of deployment issues, make sure to use same image for
-      # deploy-dev and deploy-release
       - image: cibuilds/github:0.13
 
     steps:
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -245,25 +247,34 @@ jobs:
             # update 301 redirect for the /latest/* route.
             aws s3api put-bucket-website --cli-input-json file://.circleci/s3-website-config.json
 
-  deploy-s3:
+  deploy-dev:
```
rth, line 250: If you rename it here, you need to rename it in the `workflows` as well.

.circleci/config.yml
```diff
@@ -211,6 +211,8 @@ jobs:
           path: /root/repo/build/benchmarks.json
 
   deploy-release:
+    # To reduce chance of deployment issues, try to keep the steps here as
+    # similar as possible to the steps in deploy-release!
```
rth, line 215:
```suggestion
    # similar as possible to the steps in deploy-dev!
```

.circleci/config.yml
```diff
@@ -245,25 +247,34 @@ jobs:
...
+            apk add --no-cache --update python3 make
             python3 -m pip install awscli
+      - run:
+          name: Deploy Draft Github Release
+          command: |
+            cp -r build /tmp/pyodide
+            cd /tmp/
+            export CIRCLE_TAG="dev"
+            tar cjf pyodide-build-${CIRCLE_TAG}.tar.bz2  pyodide/
+            ghr -draft -t "${GITHUB_TOKEN}" -u "${CIRCLE_PROJECT_USERNAME}" -r "${CIRCLE_PROJECT_REPONAME}" -c "${CIRCLE_SHA1}" -delete  "${CIRCLE_TAG}" ./pyodide-build-${CIRCLE_TAG}.tar.bz2
```
rth, line 274: My earlier comment about this wasn't saved apparently.

So I think this wouldn't work because release (including draft) needs to be associated with a tag and in this case "${CIRCLE_TAG}" would be empty. Maybe we could run `ghr --help` or `ghr --version` (not sure which one would be correct according to their doc).

hoodmane: Well I did attempt to account for that: see line 272 above:
`export CIRCLE_TAG="dev"`

rth: Yeah, but the tag wouldn't exist in git history.

rth: When trying that in https://github.com/pyodide/pyodide/releases/new

Says "Excellent! This tag will be created from the target when you publish this release. "

I don't think we want to create tags for each commit either. So if we release a bit frequently we will be fine, this is probably not worth it.

hoodmane: So delete these lines?

rth: yes, let's do that.

Issue #1504, opened 2021-04-20
Improved Webworker ease of use


hoodmane:
Currently using Pyodide in a webworker involves writing a bunch of messy message passing code and is rather error prone and confusing.

We should figure out best practices and document them. Furthermore, if the best practices are too annoying, we should figure out how to add features that make them less annoying.

jtpio:
Another thing about using Pyodide running in a web worker is not being to access the DOM and create elements.

There is already a "Caveats" section in the docs mentioning some of these limitations: 

https://pyodide.org/en/latest/usage/webworker.html#caveats

A concrete example would be when using matplotlib with:

```python
import numpy as np
import matplotlib.pyplot as plt

x = np.linspace(0, 10, 1000)
plt.plot(x, np.sin(x));

plt.show()
```

Which gives the following error as it tries to create new elements with the wasm backend:

```
Traceback (most recent call last):
  File "<console>", line 2, in <module>
  File "/lib/python3.8/site-packages/matplotlib/pyplot.py", line 2336, in <module>
    switch_backend(rcParams["backend"])
  File "/lib/python3.8/site-packages/matplotlib/pyplot.py", line 276, in switch_backend
    class backend_mod(matplotlib.backend_bases._Backend):
  File "/lib/python3.8/site-packages/matplotlib/pyplot.py", line 277, in backend_mod
    locals().update(vars(importlib.import_module(backend_name)))
  File "/lib/python3.8/importlib/__init__.py", line 127, in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
  File "/lib/python3.8/site-packages/matplotlib/backends/wasm_backend.py", line 23, in <module>
    from js import document
ImportError: cannot import name 'document' from 'js' (unknown location)
```

Maybe the best practices could document these differences based on the environment where Pyodide is loaded. In the case of matplotlib, maybe it's a matter of having another backend that would work in a web worker (or document how to make it work if that's possible).

hoodmane:
> maybe it's a matter of having another backend that would work in a web worker

This is a TODO. Would you mind opening an issue for it?

jtpio:
Sure, just opened https://github.com/pyodide/pyodide/issues/1518.

stefnotch:
I've used Pyodide in a webworker and here are my two cents:
- It should still be possible to use it in a SharedWorker (I've been abusing those to prevent having to reload Pyodide all the time during development)
[...]

PR #1535, opened 2021-04-23
Comlink support

hoodmane:
This enables very basic Comlink support which can give fairly direct DOM access to Pyodide in a webworker. A bit experimental. CF pyodide/pyodide#1504, pyodide/matplotlib-pyodide#6.

I am serving this branch here:
https://spectralsequences.com/worker-pyodide-console/
demo code is here:
https://github.com/hoodmane/worker-pyodide-console/

You can use the following code to register an event listener on the main thread:
```py
from js import callProxy, windowProxy
def f(e):
    print("click!!", e.to_py())
await callProxy(windowProxy.document.addEventListener, "click", f)
```
then if you click it will log into the console:
`click!! {'target': {'id': '', 'classList': [], 'clientX': 795.2000122070312, 'clientY': 256.20001220703125}}`


src/pyodide.js
```diff
@@ -432,6 +432,15 @@ globalThis.loadPyodide = async function(config = {}) {
       false; // we preload wasm using the built in plugin now
   Module.preloadedWasm = {};
 
+  /**
+   * Tell Pyodide about an available copy of Comlink.
+   * Necessary to enable use of Comlink Proxies from Python.
+   * @private
+   */
+  Module.registerComlink = function(Comlink){
```
hoodmane, line 440: It seems like we could use some sort of optional features system: there are three different packages that can lead to special interop when present:
* Comlink
* ndarray https://github.com/scijs/ndarray -- if present we could automatically fill in shape and strides when converting to Python rather than forcing user to use `np.as_array` and then do a manual cast.
* Stack trace js https://www.stacktracejs.com/ -- if present, when combined with #1447 we can produce more informative tracebacks.

src/pyodide.js
```diff
@@ -432,6 +432,15 @@ globalThis.loadPyodide = async function(config = {}) {
...
+    Module.Comlink = Comlink;
```
rth, line 441: Something like
```suggestion
    Module._Comlink = Comlink;
```
might be a bit more explicit about the fact that this object is not part of the public API.

src/core/pyproxy.js
```diff
@@ -639,8 +642,7 @@ JS_FILE(pyproxy_init_js, () => {0,0; /* Magic, see include_js_file.h */
       // error if user tries to write over any key of type 1. things we have to
       // return to avoid making Javascript angry
       if (typeof (jskey) === "symbol") {
-        throw new TypeError(
-            `Cannot set read only field '${jskey.description}'`);
+        return Reflect.set(jsobj, jskey, jsval);
```
rth, line 645: The above comment would need updating

rth (commented):
>  very basic Comlink support which can give fairly direct DOM access to Pyodide in a webworker.

Very nice demo!

Do we have any way to test it as part of our test suite?

hoodmane:
Thanks for the review! I think this PR is going to be a good addition, but I think I should check that it works well with the other stuff I'm working with before I get back to it so I'll mark it as a draft for now.

hoodmane:
> Do we have any way to test it as part of our test suite?

Yeah I think there's a fine way to do this: I think we need to add new files (comlink_worker.html / comlink_worker.js) then have some tests that do stuff like my demo.

PR #1538, opened 2021-04-23
Remove loadPackagesFromImports from runPythonAsync

### LINKED ISSUE #1496: Remove loadPackagesFromImports from runPythonAsync?
This would be a backwards incompatible API change but I think it's for the best.

hoodmane:
Resolves #1496.

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -181,7 +181,8 @@ class SeleniumWrapper:
     def run_async(self, code):
         return self.run_js(
             f"""
-            let result = await pyodide.loadPackagesAndRunPythonAsync({code!r});
+            await pyodide.loadPackagesFromImports({code!r})
+            let result = await pyodide.runPythonAsync({code!r});
             if(result && result.toJs){{
                 let converted_result = result.toJs();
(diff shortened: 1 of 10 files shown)


rth:
One one side I understand the motivation for removing it https://github.com/pyodide/pyodide/issues/1496#issuecomment-826902735 on the other function `loadPackagesAndRunPythonAsync` is not really readable IMO and it does look like we are using a fair amount of times (and aside for examples are never using this new version of `runPythonAsync`). 

Maybe addition an optional parameter to runPythonAsync would be more readable?

hoodmane:
> it does look like we are using a fair amount of times

I just converted everything to the old version to save energy on converting them.
I will go through and replace with:
```
await pyodide.loadPackagesFromImports(code); 
await pyodide.runPythonAsync(code)
```
I think this is much less surprising. I think it's better to just have separate functions `do_a()` and `do_b()` than to have `do_b(do_a_also=True)`.

rth:
OK, then maybe we could remove `loadPackagesAndRunPythonAsync` as well? As you said calling it explicitly is just one extra line.

PR #1539, opened 2021-04-25
API Add PyProxy.callKwargs

### LINKED ISSUE #1122: Support for kwargs between python and js
We should support using kwargs to call python functions from javascript and vice-versa. This would have made #1121 obvious.

hoodmane:
Add a method to call a PyProxy of a callable with keyword arguments. Resolves #1122.

### DIFF
--- a/docs/project/changelog.md
+++ b/docs/project/changelog.md
@@ -12,6 +12,10 @@ substitutions:
 
 ## [Unreleased]
 
+- {{ Enhancement }} Added the {any}`PyProxy.callKwargs` method to allow using
+  Python functions with keyword arguments from Python.
+  {pr}`1539`
+
 ## Version 0.17.0
 *April 21, 2020*
 
--- a/docs/project/changelog.md
+++ b/docs/project/changelog.md
@@ -13,7 +13,7 @@ substitutions:
 ## [Unreleased]
 
 - {{ Enhancement }} Added the {any}`PyProxy.callKwargs` method to allow using
-  Python functions with keyword arguments from Python.
+  Python functions with keyword arguments from Javascript.
   {pr}`1539`
 
 ## Version 0.17.0
--- a/src/core/pyproxy.c
+++ b/src/core/pyproxy.c
@@ -368,11 +368,39 @@ finally:
   return iddir;
 }
 
+/**
+ * This sets up a call to _PyObject_Vectorcall. It's a helper fucntion for
+ * callPyObjectKwargs. This is the primary entrypoint from Javascript into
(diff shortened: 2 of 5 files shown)


hoodmane:
Perhaps some other name like "withKwargs" would be more appropriate?

phorward (approved):
Hello @hoodmane, thanks for the feature. I'm fine with it, also with the name `callKwargs`. Let's wait for other reviewers to comment on this.

docs/project/changelog.md
```diff
@@ -12,6 +12,10 @@ substitutions:
 
 ## [Unreleased]
 
+- {{ Enhancement }} Added the {any}`PyProxy.callKwargs` method to allow using
+  Python functions with keyword arguments from Python.
```
rth, line 16: Isn't it from JS?

rth (approved):
Thanks @hoodmane and @phorward ! 

This LGTM as well.

> also with the name callKwargs

`callWithKwargs` might be a bit more readable but fairly long so the current name sounds good.

So if I understand correctly the motivation is that if we do,
```js
f(8, { y : 4 })
```
we can't auto-determine if the last argument is a kwarg or a positional argument that happened to be a object? Determining that by looking at the python function signature wouldn't work?

hoodmane:
> we can't auto-determine if the last argument is a kwarg or a positional argument that happened to be a object? Determining that by looking at the python function signature wouldn't work?

It won't work in full generality though it will work well in a lot of cases. I am definitely interested in implementing such a scheme but I think being smart like that requires a lot of care to ensure that it doesn't cause problems. With this API, we can make the clever system conservative and have people fall back to the explicit way when it doesn't work out.

docs/project/changelog.md
hoodmane: Yes.

src/core/pyproxy.c
```diff
@@ -753,24 +815,15 @@ _pyproxy_get_buffer(PyObject* ptrobj)
...
-    return result_heap;
-  } else {
-    hiwire_CLEAR(result.shape);
-    hiwire_CLEAR(result.strides);
-    PyBuffer_Release(&view);
-    return NULL;
-  }
+  // The result.view memory will be freed when (if?) the user calls
+  // Py_Buffer.release().
+  result.view = (Py_buffer*)PyMem_Malloc(sizeof(Py_buffer));
```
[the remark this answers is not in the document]
hoodmane: Yeah I was thinking about that too. But this is an unrelated change, I should move it to a different PR.

Issue #1477, opened 2021-04-16
ES6 modules/import support


josephrocca:
Is there any desire from the core team to move to ES6 modules? This would make it easier to work with for modern JS development and it would also make it compatible with [Deno](https://deno.land) out-of-the-box, since Deno is highly compatible with the browser JS environment (though not to the point of supporting script element creation as in the snippet below, hence the current lack of Deno support). 

https://github.com/pyodide/pyodide/blob/093c0dd18ddcefa4cc1bce5f404f4cec2444ef9d/src/pyodide.js#L58-L73

Just now I "wrapped" `ftfy` into [a JS module](https://github.com/josephrocca/ftfy-pyodide) and I am **amazed** at how easy/seamless it was. The Python module is only 4x faster than this Pyodide-based JS one, which is significantly better than I was expecting. I'd love to be able to bring it to the server-side, since there isn't any JS module like this as far as I'm aware.

It's kind of mind blowing to me that, thanks to Pyodide, the JS ecosystem can suddenly benefit from the many, many developer-years that have gone into the Python package ecosystem.

There are quite a few Python packages like `ftfy` where no comparable JS package exists, and so I'd love to be able to bring them the both the browser and the server. Using ES6 modules would make that possible.

I would be happy to do the work required to move to ES6 modules - I'd just need to some guidance on what changes would need to be made.

rth:
Hi @josephrocca,

yes it would be great for 0.18.0. 

Right now we are focused on the 0.17.0 release which will happen by the end of next week. After that we should make a plan on how to incrementally make this transition (see related issues https://github.com/pyodide/pyodide/issues/14 https://github.com/pyodide/pyodide/issues/328 https://github.com/pyodide/pyodide/pull/792 ), and your contribution would certainly be very welcome!

josephrocca:
@rth Okay, great! Let me know when you've got some work for me to do 👍

hoodmane:
@josephrocca Thanks for this suggestion.

> Is there any desire from the core team to move to ES6 modules?

Yes so much desire for this (from me anyways). With this we could eliminate the rest of the global variables from the Pyodide runtime, and also our current loading scheme is really terrible and module loading is 
 See #1364 for my most recent attempt.

Currently we are blocked on Firefox/Safari support for using ES6 Modules in webworkers. It seems that this feature landed in Firefox 8 days ago and in Safari in February (two months ago):
https://bugzilla.mozilla.org/show_bug.cgi?id=1247687
[...]

PR #1540, opened 2021-04-26
MAINT Change --export-name to globalThis.pyodide._module

hoodmane:
This is one of the steps @josephrocca suggested in #1477. Hopefully CI passes, I haven't built locally with this.


hoodmane:
Okay this works fine.

Issue #1542, opened 2021-04-27
Reduce the size of the stdlib


rth:
Part of https://github.com/pyodide/pyodide/issues/646#issuecomment-827735529

Also related to https://discuss.python.org/t/minifying-the-stdlib-in-pyodide/8414

Largest modules are (size is in kB),
```
$ du -a . | sort -n -r | head -n 20
11060   .
1732    ./encodings
746    ./distutils
664     ./pydoc_data
660     ./pydoc_data/topics.py
504     ./asyncio
436     ./email
336     ./xml
300     ./distutils/command
256     ./unittest
224     ./_pydecimal.py
208     ./http
176     ./importlib
172     ./logging
164     ./urllib
148     ./xml/dom
116     ./inspect.py
108     ./pydoc.py
```

**Encodings**

From https://github.com/pyodide/pyodide/issues/646#issuecomment-827524917 by @hoodmane 
> I still think it would be a good idea to figure out which text encodings are actually used. They take up a lot of space in stdlib. I took ~80% of them out on my build and have been using it this way for a while. I've only gotten one missing codec error so far.

Reading https://docs.python.org/3/library/codecs.html#standard-encodings maybe the ideal case would be by default to only include UTF-8, possibly latin1 (maybe some Python files still use it?) and possibly Python specific encodings. Then put all the rest into an independent optional package. 

**distutils**

Per [PEP 632](https://www.python.org/dev/peps/pep-0632/) is going to be deprecated in Python 3.10 and removed in 3.12, so we might as well unpackage it as a standalone package earlier. I'm not sure why we would need it at runtime.

**pydoc + pydoc_data**

Potentially we could make it standalone, but it's necessary to make "help" work and also used in a few places in stdlib.

**Other**

Then there is all the work by @phorward in https://github.com/pyodide/pyodide/issues/646#issuecomment-823657002 and suggestions from https://discuss.python.org/t/minifying-the-stdlib-in-pyodide/8414

hoodmane:
> the ideal case would be by default to only include UTF-8, possibly latin1 (maybe some Python files still use it?) and possibly Python specific encodings. 

This is a good starting point, but I found that Python uses weird encodings more often than you would hope, so I suspect we'll need to throw a few odd ones back in. latin1 is definitely necessary.

hoodmane:
> Potentially we could make it standalone, but it's necessary to make "help" work and also used in a few places in stdlib. So that's probably less obvious.

[...]

PR #1543, opened 2021-04-27
Unvendor distutils

rth:
Packages the distutils stdlib module as a standalone package,
```
377K distutils.data
221K distutils.data.br
19K distutils.js
```
to reduce the minimal pyodide size.

Part of #1542 

This also,
 - adds documentation on what's supported or not in the stdlib
 - makes a few changes to `pyodide_build/buildall.py` to be able to add such stdlib packages as a dependency in other packages (which didn't work before). For instance, until there is a release with https://github.com/pypa/packaging/pull/396 micropip has a dependency on distutils.


Makefile
```diff
@@ -156,6 +160,16 @@ build/test.data: $(CPYTHONLIB) $(UGLIFYJS)
...
+build/distutils.data: $(CPYTHONLIB) $(UGLIFYJS)
+	( \
+		cd $(CPYTHONLIB)/distutils; \
+		find . -type d -name __pycache__ -prune -exec rm -rf {} \; \
+	)
+	( \
+		cd build; \
+		python $(FILEPACKAGER) distutils.data --lz4 --preload ../$(CPYTHONLIB)/distutils@/lib/python$(PYMINOR)/distutils --js-output=distutils.js --export-name=pyodide._module --exclude __pycache__ \
+	)
+	$(UGLIFYJS) build/distutils.js -o build/distutils.js
```
rth, line 172: Not particularly happy about having to repeat this for each unvendored package, but I'm not sure what else we can do.

I considered creating `meta.yaml` for such stdlib package and treating them as regular package. But that doesn't work as we need to be sure that CPython build artifacts are available (which is not part of the current build / dependency resolution setup)

docs/index.rst
```diff
@@ -25,6 +25,7 @@ Using Pyodide
    usage/serving-pyodide-packages.md
    usage/loading-packages.md
    usage/type-conversions.md
+   usage/wasm-constraints.md
```
rth, line 28: I'm open to better names

pyodide_build/buildall.py
```diff
@@ -131,7 +159,11 @@ def generate_dependency_graph(
     while packages:
         pkgname = packages.pop()
 
-        pkg = Package(packages_dir / pkgname)
+        pkg: BasePackage
+        if pkgname in UNVENDORED_STDLIB_MODULES:
+            pkg = StdLibPackage(packages_dir / pkgname)
+        else:
+            pkg = Package(packages_dir / pkgname)
```
rth, line 166: Most of the code refactoring in this file is to be able to do the above.

docs/usage/wasm-constraints.md
```diff
@@ -0,0 +1,44 @@
...
+
+Most of the Python standard library is functional, except for the modules
+listed in the sections below. A large part of the CPython test suite passes except for
+tests skipped in
+[`src/tests/python_tests.txt`](https://github.com/pyodide/pyodide/blob/main/src/tests/python_tests.txt)
+or via [patches](https://github.com/pyodide/pyodide/tree/main/cpython/patches).
+
+### Removed modules
+
+Following modules are removed from the standard library to reduce download size and
```
hoodmane, line 14:
```suggestion
The following modules are removed from the standard library to reduce download size and
```
and similarly in the other sections.

rth:
So distutils ended up being used more frequently than I though in packages. I fixed it for the packages we include though that would still be an issue for micropip installed packages. We may want to have an option in `loadPyodide` to load the full stdlib in the future.

hoodmane:
> So distutils ended up being used more frequently than I though in packages

An unfortunate aspect of standard libraries is that packages use them in a lot of circumstances where they wouldn't add a dependency.

> We may want to have an option in `loadPyodide` to load the full stdlib in the future.

For stuff like this it can be nice to add it as an option with the old way of doing things as the default. I like this approach a lot when it makes sense because it gives an extra opportunity to collect info about the change before committing to it. Changing from opt-in to opt-out and then later removing the option entirely makes two smaller breaking changes out of one bigger one.

rth:
Yes, I agree it's better to keep backward compatibility on stdlib with default options of `loadPyodide`, will look into it.

rth:
OK, it should be now loading the the full stdlib by default. I'll extend this to a few other large and infrequently used modules once this PR is merged.

docs/project/changelog.md
```diff
@@ -12,6 +12,17 @@ substitutions:
 
 ## [Unreleased]
 
+## Standard library
+
+- Following standard library modules are now available as standalone packages
```
hoodmane, line 17:
```suggestion
- The following standard library modules are now available as standalone packages
```

docs/usage/wasm-constraints.md
```diff
@@ -0,0 +1,46 @@
...
+ - lib2to3
+ - tkinter
+ - turtle.py
+ - turtledemo
+ - venv
+ - webbrowser.py
+
+### Included but not working modules
+
+The following modules can be imported, but are not functional due to the limitations of the WebAssembly VM,
```
hoodmane, line 41:
```suggestion
The following modules can be imported, but are not functional due to the limitations of the WebAssembly VM:
```

src/pyodide.js
```diff
@@ -17,6 +17,9 @@ globalThis.pyodide = {};
  * Pyodide module as a global object called ``pyodide``.
  * @param {string} config.indexURL - The URL from which Pyodide will load
  * packages
+ * @param {bool} config.fullStdLib - Load the full Python standard library.
+ * Setting this parameter to false would explude following modules: distutils.
```
hoodmane, line 21:
```suggestion
 * Setting this to false excludes following modules: distutils.
```

hoodmane (approved):
Looks good. Hopefully this process will be simplified as we unvendor more things.

PR #1544, opened 2021-04-27
DOC Add communication channels links

rth:
List communication channels more prominently. 

In particular, thanks to @msapiro we now have a public mailing list `pyodide@python.org` for announcements (https://mail.python.org/mailman3/lists/pyodide.python.org/ to subscribe).

### DIFF
--- a/README.md
+++ b/README.md
@@ -69,6 +69,13 @@ independent and community-driven open-source project. The decision making
 process is outlined in the [Project
 governance](https://pyodide.org/en/stable/project/governance.html).
 
+## Communication
+
+- Mailing list: [mail.python.org/mailman3/lists/pyodide.python.org/](https://mail.python.org/mailman3/lists/pyodide.python.org/)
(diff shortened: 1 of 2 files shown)


hoodmane (approved):
How do I get post access to the mailing list? We should perhaps send out a couple of test mails to see that it works.

rth:
You need to click on the URL subscribe, then you will be able to send emails to that email address. You can send emails without subscribing, but then posts are moderated and someone needs to approve them.

rth:
> We should perhaps send out a couple of test mails to see that it works.

I have used this system enough to be reasonably confident that it works without test emails :) The main issue is that only 2 of us subscribed at present.. I made you admin so you should be able to see the list of subscribers as well.

hoodmane:
I'll make an issue and @ the same people from the 0.17.0 release.

rth:
Any objections to merging this so I can link to it in the readme?.

hoodmane:
No objections.

PR #1549, opened 2021-04-29
Add PyProxy.clone

hoodmane:
Allows easier control over Python references from Javascript. Prerequisite for "borrowed PyProxies" plan.

### DIFF
--- a/src/core/pyproxy.js
+++ b/src/core/pyproxy.js
@@ -98,7 +98,8 @@ JS_FILE(pyproxy_init_js, () => {0,0; /* Magic, see include_js_file.h */
   function _getPtr(jsobj) {
     let ptr = jsobj.$$.ptr;
     if (ptr === null) {
-      throw new Error("Object has already been destroyed");
+      throw new Error(jsobj.$$.destroyed_msg ||
+                      "Object has already been destroyed");
     }
     return ptr;
   }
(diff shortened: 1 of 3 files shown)


src/core/pyproxy.js
```diff
@@ -227,20 +228,33 @@ JS_FILE(pyproxy_init_js, () => {0,0; /* Magic, see include_js_file.h */
...
+      this.$$.destroyed_msg = destroyed_msg;
       try {
         _Py_DecRef(ptrobj);
       } catch (e) {
         Module.fatal_error(e);
       }
     }
+    /**
+     * Make a new PyProxy pointing to the same Python object.
+     * Useful if someone else is destroying your PyProxy.
```
rth, line 252: "someone" makes mes wonder what other person might have access to those PyProxy.

Maybe "if your PyProxy gets destroyed elsewhere in the code" or something similar?

rth (approved):

src/core/pyproxy.js
hoodmane: Sounds good.

rth:
@hoodmane  About this, I'm wondering, maybe this should have been a `copy` instead of `clone`? That would be more familiar to Python users. We can still change the name since it's not yet released.

Unless we plan to have both `copy` and `clone` as in Rust for proxies?

hoodmane:
I think we are only going to have one. I don't care whether it is named `copy` or `clone`. In fact both names are a bit weird because when you "clone" a proxy you get a new proxy that points to the same Python object. So it's neither a deep copy nor a shallow copy. It just creates a new reference...

But yeah I'm happy to rename to copy if you think it's better.

rth:
> In fact both names are a bit weird because when you "clone" a proxy you get a new proxy that points to the same Python object.

That sounds similar to a shallow copy to me. Don't we have the same thing with,
```py
a = {'x': [23]}
```
where `a.copy()` will produce a new object that points to the same list object? 

Anyway made the PR in https://github.com/pyodide/pyodide/pull/1630

PR #1550, opened 2021-04-30
MAINT "use strict" in pyodide.js

hoodmane:
Last time I tried to do this it didn't work but apparently we've cleaned up pyodide.js enough that it works now.


rth (approved):
Great! 

On unrelated subject, I think we may have introduced an issue with webworker tests that also happen on master https://github.com/pyodide/pyodide/issues/1389#issuecomment-829881445

PR #1552, opened 2021-04-30
MAINT Refactor package lock

hoodmane:
Hide the promise-chain hijinks in a named function to clean up application logic.

This changes the logic slightly by not releasing the lock between the two calls to `_loadPackage`. I think that the old version was wrong, but because the first `_loadPackage` call is currently only ever used to load LAPACK, we didn't ever notice.


hoodmane:
Do we have a unit test that makes a bunch of calls to `loadPackage` at once to cause contention on this lock?

PR #1554, opened 2021-04-30
MAINT Use prettier to format javascript

hoodmane:


src/pyodide.js
```diff
@@ -53,27 +54,35 @@ globalThis.loadPyodide = async function(config = {}) {
...
   }
 
-  function recursiveDependencies(names, _messageCallback, errorCallback,
-                                 sharedLibsOnly) {
+  function recursiveDependencies(
+    names,
+    _messageCallback,
+    errorCallback,
+    sharedLibsOnly
+  ) {
```
rth, line 85: yeah, it looks a bit more consistent with black.

Makefile
```diff
@@ -161,6 +162,9 @@ $(UGLIFYJS): emsdk/emsdk/.complete
 	npm i --no-save uglify-js
 	touch -h $(UGLIFYJS)
 
+$(PRETTIER): emsdk/emsdk/.complete
+	npm i --no-save prettier
```
rth, line 166:
```suggestion
	npm install --no-save prettier
```

rth (approved):
Sounds good once CI passes. We would also need to update contributing docs I think.

hoodmane:
I'm just really excited about no more `// clang-format off` every time I compare two things.

Dockerfile
```diff
@@ -36,5 +22,39 @@ RUN wget --quiet https://chromedriver.storage.googleapis.com/89.0.4389.23/chrome
   && mv $PWD/chromedriver /usr/local/bin \
   && rm -f chromedriver_linux64.zip
 
+# Copy node install from node-image
+COPY --from=node-image /usr/local/bin/node /usr/local/bin/
+COPY --from=node-image /usr/local/lib/node_modules /usr/local/lib/node_modules
+RUN ln -s ../lib/node_modules/npm/bin/npm-cli.js /usr/local/bin/npm
+RUN ln -s ../lib/node_modules/npm/bin/npx-cli.js /usr/local/bin/npx
+
+RUN rm -rf /var/lib/apt/lists/*
```
rth, line 31: You could make this a single `RUN` step, so it wouldn't create too many docker layers..

Dockerfile
```diff
@@ -36,5 +22,39 @@ RUN wget --quiet https://chromedriver.storage.googleapis.com/89.0.4389.23/chrome
...
+  pytest-instafail \
+  pytest-rerunfailures \
+  pytest-xdist \
+  "selenium==4.0.0.b3"
+
+RUN pip3 --no-cache-dir install \
+  black \
+  "cython<3.0" \
+  packaging \
+  flake8 \
```
rth, line 46: Same here, better to keep this a single step.

Dockerfile
hoodmane: Hmm:

> Where possible, only copy the artifacts you need into the final image. This allows you to include extra tools and/or to debug information in your intermediate build stages, without increasing the size of the final image.

Ideally it'd be nice to have a version with tons of layers for working on the image and then completely flatten it for the actual version. Maybe this is too much trouble though.

hoodmane: Seems like --squash does this flattening.

rth: Yeah, but that means that you can't reuse existing layer (e.g. from base image). So better to just avoid creating too many layer in the Docker image. Also because this might be built and used by other people.

hoodmane: Unfortunately it requires an experimental docker-daemon flag to enable it, so perhaps we can wait until they release it as a normal feature. For now, will merge these steps as you suggested.

hoodmane: The Dockerfile API is pretty clumsy.

hoodmane: Is there a way to merge these COPYs:
```
COPY --from=node-image /usr/local/bin/node /usr/local/bin/
COPY --from=node-image /usr/local/lib/node_modules /usr/local/lib/node_modules
```

hoodmane:
@rth I am getting:
`Version '89.*' for 'chromium' was not found`

hoodmane:
Funnilly enough, this newest image is smaller than pyodide-env:15 even though I installed a bunch more stuff in it (node, some node packages, more python packages).

rth:
It doesn't look like copying chrome files from the previous image is working. The issue is that chromium has lots of dependencies that we would be missing here. 



> Strangely, it seems to want to install version 88 not version 89:

The following works fine for me with Chrome 90,

<details>

```diff
diff --git a/Dockerfile b/Dockerfile
index c3ef05b..bf64934 100644
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,5 +1,4 @@
 FROM node:14.16.1-buster-slim AS node-image
-FROM pyodide/pyodide-env:15 AS prev-image
 FROM python:3.8.2-slim-buster
 
 RUN apt-get update \
@@ -8,7 +7,7 @@ RUN apt-get update \
                   bzip2 ccache clang-format-6.0 cmake f2c g++ gfortran git make \
                   patch pkg-config swig unzip wget xz-utils \
                   # testing packages: libgconf-2-4 is necessary for running chromium
-                  libgconf-2-4 \
+                  libgconf-2-4 "chromium=90.*" \
   && rm -rf /var/lib/apt/lists/*
 
 RUN pip3 --no-cache-dir install \
@@ -45,7 +44,7 @@ RUN wget -qO- https://ftp.mozilla.org/pub/firefox/releases/87.0/linux-x86_64/en-
   && wget -qO- https://github.com/mozilla/geckodriver/releases/download/v0.29.1/geckodriver-v0.29.1-linux64.tar.gz | tar zxC /usr/local/bin/
 
 # Get recent version of chromedriver
-RUN wget --quiet https://chromedriver.storage.googleapis.com/89.0.4389.23/chromedriver_linux64.zip \
+RUN wget --quiet https://chromedriver.storage.googleapis.com/90.0.4430.24/chromedriver_linux64.zip \
   && unzip chromedriver_linux64.zip \
   && mv $PWD/chromedriver /usr/local/bin \
   && rm -f chromedriver_linux64.zip
@@ -62,11 +61,5 @@ RUN npm install -g \
   rollup \
   rollup-plugin-terser
 
-COPY --from=prev-image /usr/bin/chromium /usr/bin/
-COPY --from=prev-image /usr/lib/chromium /usr/lib/
-COPY --from=prev-image /etc/chromium.d /etc/
-COPY --from=prev-image /etc/chromium /etc/
-COPY --from=prev-image /usr/share/chromium /usr/share/
-
 CMD ["/bin/sh"]
 WORKDIR /src
```
</details>

I agree we should find a way to pin the version of chromium installed, but maybe let's not do that here? It does sound non that obvious.

Dockerfile
```diff
@@ -36,5 +50,23 @@ RUN wget --quiet https://chromedriver.storage.googleapis.com/89.0.4389.23/chrome
...
+COPY --from=node-image /usr/local/lib/node_modules /usr/local/lib/node_modules
+RUN ln -s ../lib/node_modules/npm/bin/npm-cli.js /usr/local/bin/npm \
+    && ln -s ../lib/node_modules/npm/bin/npx-cli.js /usr/local/bin/npx
+
+RUN npm install -g \
+  jsdoc \
+  uglify-js \
+  prettier \
+  rollup \
+  rollup-plugin-terser
```
rth, line 63: I think we should also replace uglify-js with terser, which is more modern (supports ES6+), better maintained, and used in rollup. We can do it in a separate PR, just need to make sure it's installed here, which should be the case I think?

hoodmane: See #1559.

rth:
BTW, not sure if I told you already, but you should now be able to push images to `pyodide/pyodide` on Docker Hub.

hoodmane:
Push to pyodide/pyodide-env is failing with:

`denied: requested access to the resource is denied`

rth:
Could you try again? I think there was an issue with permissions that got reset for some reason.

hoodmane:
Yes, it works now.

rth (approved):

PR #1555, opened 2021-04-30
Don't disable right click globally for matplotlib figures

gzuidhof:
Let's disable it only on the figure.

This disables the need for [this hack](https://github.com/gzuidhof/starboard-python/blob/master/src/global.ts#L51-L57):

```javascript
/** Naughty matplotlib WASM backend captures and disables contextmenu globally.. hack to prevent that */
window.addEventListener("contextmenu", function (event) {
    if (event.target instanceof HTMLElement && event.target.id.startsWith("matplotlib_") && event.target.tagName === "CANVAS") {
        return false;
    }
    event.stopPropagation();
}, true);
```


hoodmane (approved):
Looks good! (It'd be nice if we could have some test coverage...)

hoodmane:
If we get around to cleaning up the matplotlib backend we could add tests that render the image and then trigger mouse interaction. Probably would take a fair bit of work.

rth (approved):
This would be non easy to test, and a I also think it should be OK to merge. 

Thanks @gzuidhof !

PR #1556, opened 2021-04-30
Include necessary matplotlib icons, don't scroll into view

gzuidhof:
Another small step towards iodide independence: bundle the necessary icons instead of having an implicit requirement on fontawesome.

Another small change I included: don't scroll the window to the latest matplotlib figure. This should be done in the surrounding application - it's not for the matplotlib backend to decide.

Note: I currently don't have a working build pipeline to verify this change, please give it a sanity check that it works!

### DIFF
--- a/packages/matplotlib/src/wasm_backend.py
+++ b/packages/matplotlib/src/wasm_backend.py
@@ -11,7 +11,6 @@ this primarily is just overriding methods in the base class.
 """
 
 # TODO: Figure resizing support
-
 import base64
 import io
 import math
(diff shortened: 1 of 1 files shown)


hoodmane:
One tip: Make a symlink from `build/wasm_backend.py` to the source file and then do:
```js
let wasm_backend_src = await fetch("./wasm_backend.py");
pyodide._module.FS.unlink("/lib/python3.8/site-packages/matplotlib/backends/wasm_backend.py");
pyodide._module.FS.writeFile("/lib/python3.8/site-packages/matplotlib/backends/wasm_backend.py", new Uint8Array(await wasm_backend_src.arrayBuffer()));
```
This way you can try out changes to `wasm_backend` without rebuilding anything.

ryanking13:
matplotlib now lives under [matplotlib-pyodide](https://github.com/pyodide/matplotlib-pyodide), and things have changed a lot since this PR opened. So, closing for now.

PR #1559, opened 2021-05-01
Split pyodide.js into several files with rollup

hoodmane:
I split pyodide.js into four files. I did a small amount of rearrangement to make a logical split into files and to make all of the name errors go away. `loadPyodide` is now only 75 lines long.

This is on top of prettier #1554, I'll keep it as a draft until that is merged. I think it will take some work to fix the Javascript API docs.

The files:

`module.js` : this creates the module and writes a few settings to it. This needs to be accessible everywhere, so I think it's natural to make it in a separate file and import it in each other file (in particular this avoids any circular imports).

`load_package.js`: This contains the first big chunk of `loadPyodide` from before. I added a new function `initializePackageIndex` that loads `packages.json`, which used to happen at the end of `loadPyodide`, but I felt was more naturally moved here. I export `loadScript`, `loadedPackages`, and `loadPackages`. Nothing else changed.

`api.js`: This contains all of the public APIs from before, `PUBLIC_API`, and `makePublicNamespace`. A few things had to change from being `Module.name` to `name` or from `name` to `Module.name`. I also turned `PUBLIC_API` into an object in order to accommodate the differences in how things work with rollup or ES6 modules (in particular I think there is no way to directly access the file-level scope which is kind of weird...)

`pyodide.js`: I moved the definitions of `fixRecursionLimit`, `fatal_error`, and `runPythonSimple` into the top level of the file because with rollup the top level of the file is a separate scope from the global scope. I didn't really like the fact that `__pyodideLoading` was an extra global variable, so I added a flag `loadPyodide.inProgress` that can be used to measure whether `loadPyodide` has been invoked yet (if you would prefer to ask for permission before calling rather than asking for forgiveness). `loadPyodide` now calls `initializePackageIndex` and `makePublicNamespace` imported from `load_package.js` and `api.js`.


rth:
Should this be closed in favor of https://github.com/pyodide/pyodide/pull/1580 etc?

Issue #1562, opened 2021-05-01
Remove CodeRunner


hoodmane:
In my opinion, `CodeRunner` should be removed. We don't ever use it, it is redundant with `functools.partial` and it can be trivially added by anyone who needs it.

Issue #1426, opened 2021-04-03
what is a proper way to reuse intermediate object of previous execution


grimmer0125:
Hi, thanks for this good tool. 

In a normal python program, we can use 
1.  class instantiation way 
```
# in main program 
a = demoClass()
a.doSomething1()
# after a while
a.doSomething2() # this can use some internal object
```
2. or we can use some `static` or `global` variable 
---

I realize that every python allocated object will keep in memory and the 2nd execution of the python script can access them. However, the current implementation (`pyodide.runPython`) looks like it is to execute a one-time python script for the 1st time, 2nd time, 3rd time...

The only way I think of is to use like 
```
if 'myVar' in globals():
    # it means 2nd time and do a different thing, e.g. use existing object
```
A little tricky, or is it possible to add one parameter for `runPython` to specify different behavior for the same python code in the future? Or using `pyodide.eval_code_async` (instead of `runPythonAsync`) to adjust the parameters passed into the 2nd time execution and use it to do conditional statement?

grimmer0125:
One more question is
For a pure Python package, will it be faster after building it to a `pyodide` package? Comparing with installing the original python package with `micropip`. Also, will its size become larger?

hoodmane:
@mgreminger I don't understand your question really. But I think this works basically the same way it would in native Python. Code run using `runPython` runs in global scope, you can write variables into global scope, use `"x" in globals()` to check if a global variable named `x` exists. You can create other scopes using the usual means.

Perhaps you are interested in the `globals` argument to `runPython`?
https://pyodide.org/en/latest/usage/api/js-api.html#pyodide.runPython

grimmer0125:
@hoodmane 
My goal is to reuse object variables and also try to execute different business logic for the 2nd time execution of the script or even each time execution has different business logic. The normal way is just to pass an argument to indicate which function/block will be called. 
Using `"x" in globals()` can achieve the first goal but how can I achieve the second goal?  `runPython` does not allow pass an argument.

hoodmane:
How about this:
```js
pyodide.runPython(`
    def some_func(*args):
        print("some_func was called with args:", args)
        return 2
`);
let some_func = pyodide.globals.get("some_func");
some_func(1,2,3);
```

grimmer0125:
Cool. It looks like a good way. Thank you !!

hoodmane:
Okay, I'll close as answered. Please don't hesitate to open another issue if you have any more questions.

[...]

PR #1563, opened 2021-05-01
ENH Improvements to eval_code

hoodmane:
This fixes a bug in the previous version where if the code does not use top level await but returns a generator or coroutine, the code would try to await it. This sometimes causes strange errors or other unexpected behaviors.

Instead, I switched to a new method of returning by raising an exception. This makes it so the coroutines produced by top level await are returned via a separate channel from return values from the code so we won't confuse them anymore.

I added a new class Executor. On creation, the Executor is given the source code and config options and produces an AST. The caller can then manipulate the AST as they see fit before calling `compile` and then `run` or `run_async`. This passes to the user the most powerful possible interface, and should make it so that Executor is helpful even in complicated use cases.

I put the main parse/compile logic into a separate generator to avoid storing any of the config options on the Executor class. This helps separate out control logic from config options and limit the scope of variables as far as possible.


rth:
Thanks! The changes of parameter names from `code` to `change` in public functions are technically breaking changes. I agree that it might be slightly better, but I'm not convinced that improvement justifies asking all users to migrate their code.

> On creation, the Executor is given the source code and config options and produces an AST. The caller can then manipulate the AST as they see fit before calling compile and then run or run_async. This passes to the user the most powerful possible interface, and should make it so that Executor is helpful even in complicated use cases.

In terms of API I wonder if we should suggest using something like,
```py
code_ast = Executor(**params).compile(source)
code_ast.run()
```
where `compile` returns `self`, instead of
```py
executor = Executor(**params)
executor.compile(source)
executor.run()
```
which poses the risk of calling those in the wrong order.

Technically compile could have returned a specific class, but it would be a bit more complex.
With option A might also make sense to raise an error if `run` is executed twice in a row, which could suggest that a call to `compile` was forgotten.

Also maybe `CodeExecutor` to be more explicit?
But this means we are back to an idea of a class for running code i.e. a `CodeRunner`. How about instead of removing it in #1562 , change it's API to do what the current `Executor` does (and still call it `CodeRunner`)?

hoodmane:
Thanks for the review!

> change of parameter names from `code` to `source` ... technically breaking changes

Okay, I guess I will revert that. Might be good to have a different name for the actually compiled code to reduce confusion.

hoodmane:
> Something like:
> ```py
> code_ast = Executor(**params).compile(source)
> code_ast.run()
> ```

I think it would be better named "Execution" instead of "Executor" since this class actually only records the life cycle of a single execution, it's not reusable.

It's important to remember that there are three steps: 
1. Parse code into AST
2. compile AST into bytecode
3. run bytecode

We want to allow people to inspect or modify the AST between steps 1 & 2, and to allow them to inspect or modify the bytecode between steps 2 & 3. So I don't think `code_ast = Executor(**params).compile(source)` works because parsing the source requires having the source. Currently I do step 1 on construction, then do step 2 when you run compile, and then do step 3 when you run `run_async` or `run`.

Anyways I'm open to a different API as long as it gives the opportunity to transform the AST before compilation and to transform the bytecode before running it.

hoodmane:
I think for one thing, if the class only is for a single execution, then it could perhaps be better called `Execution` or something? (`CodeExecution`?)

hoodmane:
Note also that it is possible to try to call the commands in the wrong order, but it will raise a runtime error so I think it's okay?

rth:
> So I don't think code_ast = Executor(**params).compile(source) works because parsing the source requires having the source

Yes, sorry I meant, `code_ast = Executor(source, **params).compile()` (didn't read the signatures carefully enough). My comment was bascially to say that if `compile` returns self it would make it easier to chain those calls.

> I think for one thing, if the class only is for a single execution, then it could perhaps be better called Execution or something? 

Maybe we should check how similar object are called in other REPL related packages. Naming wise, I think I still prefer, `CodeRunner(source).compile().run()` to  `CodeExecution(source).compile().run()` even if it is executed only once (which is encoded in the API already by passing source to `__init__`), but I'm open to other ideas.

> Note also that it is possible to try to call the commands in the wrong order, but doing so will raise a runtime error so I think it's okay?

As long as it does raise an error, sure.

hoodmane:
> I prefer `CodeRunner(source).compile().run()`'

Sounds good to me, let's go with that.

src/pyodide-py/pyodide/_base.py
```diff
@@ -224,244 +101,289 @@ def _last_assign_to_expr(mod: ast.Module):
...
-    Uses
-    `PyCF_ALLOW_TOP_LEVEL_AWAIT <https://docs.python.org/3/library/ast.html#ast.PyCF_ALLOW_TOP_LEVEL_AWAIT>`_
-    to compile to code.
+    """Runs a string as Python source code.
 
     Parameters
     ----------
-    code : ``str``
+    source
+        the Python source code to run.
```
rth, line 294: We still want the documentation of parameters here I think.

src/pyodide-py/pyodide/_base.py
```diff
@@ -224,244 +101,289 @@ def _last_assign_to_expr(mod: ast.Module):
...
-    filename : ``str``
+    Uses
+    [PyCF_ALLOW_TOP_LEVEL_AWAIT](https://docs.python.org/3/library/ast.html#ast.PyCF_ALLOW_TOP_LEVEL_AWAIT)
+    to compile the code.
 
-        The file name to use in error messages and stack traces
+    Parameters
+    ----------
+    source
+        the Python source code to run.
```
rth, line 337: Same comment.

rth (commented):
Thanks! Could you please document parameters for all added classes in docstrings?

rth:
> Instead, I switched to a new method of returning by raising an exception.

Hmm, I not sure about this. Exception handling does have a measurable overhead: doing exception handing (and also passing return values as part of the exception) each time one runs `eval_code` does not seem ideal in terms of design.

Any other approach we could try to address the initial issue?

hoodmane:
> Exception handling does have a measurable overhead: doing exception handing (and also passing return values as part of the exception) each time one runs eval_code does not seem ideal in terms of design.

Recall though that
```py
for x in l:
   # do stuff
```
desugars to:
```py
it = iter(l)
try:
   while True:
      x = next(it)
      # do stuff
except StopIteration:
   pass
```
so you are already paying this overhead every time you use a for loop.

hoodmane:
The problem with other methods is that there are two options for running code: `exec` and `eval`. `exec` never returns anything, but if the code you ran used top level await then the coroutine is lost the code is never executed and you get a "coroutine never awaited" runtime warning.

With `eval` you have a multichannel problem where it's unclear whether returned objects come from top level await or are just a random returned generator or coroutine.

I guess there is another approach: we should be able to inspect the code for a `YIELD_FROM` instruction and only treat it as a coroutine if it contains this. That's probably a better approach.

hoodmane:
This is like the fifth approach I've tried to solve this simple-seeming problem...

hoodmane:
Well actually I still think that raising the result is the best possible approach, combining the virtues of simplicity and correctness. I think the alternative I just suggested is more complicated in a few different ways (plus I'm not that excited about rearranging everything again).

I think that if raising an exception is a good enough way to terminate a for loop, it should also be a good enough way to return from `exec`?

hoodmane:
> users could very well eval_code in a loop

People also use loops inside loops. I suspect that the per-use expense of `eval_code` is mostly in parsing and compiling the code in any case. 

As a side note, we should encourage people to use `runPython` to define functions and then import them with `let my_func = pyodide.globals.get("my_func")` and then use `my_func` directly. This is faster and significantly more convenient. However, people do not find this particularly discoverable (see #1426, also my personal experience). It would be good to gradually make more of the examples in the docs use this pattern.

rth (approved):
Could you please add a changelog entry? Otherwise as mentionned in https://github.com/pyodide/pyodide/pull/1563#issuecomment-836956379 LGTM. Thanks!

PR #1566, opened 2021-05-02
Create an installable pyodide-build package

rth:
Closes https://github.com/pyodide/pyodide/issues/604

This creates an installable `pyodide-build` Python package that we can also upload to PyPi under https://pypi.org/project/pyodide-build/ (currently uploaded the dev version to reserve the name).

There is also the identically named `pyodide-build` CLI command which replaces `bin/pyodide`.

The short term goals is to use fewer hacks with `PATH` and `PYTHONPATH`. Long term goal make it a user facing interface for building packages instead of the Makefile.


pyodide-build/pyodide_build/pywasmcross.py
```diff
@@ -59,6 +58,7 @@ def collect_args(basename):
     It writes the arguments to the build.log, and then delegates to the real
     native compiler or linker.
     """
+    TOOLSDIR = Path(common.get_make_flag("TOOLSDIR"))
```
rth, line 61: This loads `Makefile.env` which wouldn't be available if `pyodide-build` is installed from PyPi so we avoid to run that command at import time.

pyodide-build/pyodide_build/tests/test_buildall.py
```diff
@@ -6,7 +6,7 @@
 from pyodide_build import buildall
 import pytest
 
-PACKAGES_DIR = (Path(__file__) / ".." / ".." / ".." / "packages").resolve()
+PACKAGES_DIR = (Path(__file__).parents[3] / "packages").resolve()
```
hoodmane, line 9: Maybe:
```suggestion
PACKAGES_DIR = (Path(__file__).absolute().parents[3] / "packages").resolve()
```
I think we should do this everywhere, if you try to run the file individually it won't work otherwise.

.gitignore
```diff
@@ -24,3 +24,4 @@ geckodriver.log
 node_modules/
 packages/.artifacts
 packages/*/build.log
+dist/
```
hoodmane, line 27: Not entirely related to this PR:

I think we should also switch to building the main pyodide files into a folder named `dist` rather than building into `build`. I found this a bit confusing -- in particular I think newcomers may not find it intuitive what the difference between `pyodide-build` and `build` is, whereas `dist` is clearer.

hoodmane (approved):
Looks good I think.

.gitignore
rth: > I think we should also switch to building the main pyodide files into a folder named dist rather than building into build

Yeah, definitely.

pyodide-build/pyodide_build/tests/test_buildall.py
rth: It's actually already done with `resolve` at the end, absolute is not really public API https://bugs.python.org/issue29688

PR #1570, opened 2021-05-03
Use PYMAJOR/PYMINOR according to Python devguide

pmp-p:
also fixes hostpython paths so that pyodide can be built everywhere a clang based hostpython is available


Makefile
```diff
@@ -5,21 +5,20 @@ include Makefile.envs
 .PHONY=check
 
 FILEPACKAGER=$$EM_DIR/tools/file_packager.py
-UGLIFYJS=npx uglifyjs
-PRETTIER=npx prettier
```
rth, line 9: Thanks a lot @pmp-p ! Could you please merge upstream/main in as I think this was added just a few commits ago on main.

pmp-p:
@rth indeed there were changes, looks ok now i think my edited makefile was from before renaming the main branch !

pyodide_build/buildpkg.py
```diff
@@ -171,7 +171,8 @@ def compile(path: Path, srcpath: Path, pkg: Dict[str, Any], args):
 
     post = pkg.get("build", {}).get("post")
     if post is not None:
-        site_packages_dir = srcpath / "install" / "lib" / "python3.8" / "site-packages"
+        pyfolder = ''.join(["python", os.environ.get("PYMAJOR","3"), ".", os.environ.get("PYMINOR","8")])
+        site_packages_dir = os.path.sep.join(srcpath, "install", "lib", pyfolder, "site-packages" )
```
rth, line 175: I think you would need
```suggestion
        site_packages_dir = os.path.join(srcpath, "install", "lib", pyfolder, "site-packages" )
```
or pass a list of paths

rth (approved):
Thanks a lot @pmp-p ! Could you please run `black` on `pyodide_build/buildpkg.py` and apply the fix below, to hopefully make CI pass? Otherwise LGTM.

pyodide_build/buildpkg.py
```diff
@@ -171,7 +171,8 @@ def compile(path: Path, srcpath: Path, pkg: Dict[str, Any], args):
 
     post = pkg.get("build", {}).get("post")
     if post is not None:
-        site_packages_dir = srcpath / "install" / "lib" / "python3.8" / "site-packages"
+        pyfolder = ''.join(["python", os.environ.get("PYMAJOR","3"), ".", os.environ.get("PYMINOR","8")])
```
rth, line 174: Maybe we could add

```suggestion
        # use Python, 3.8 by default
        pyfolder = ''.join(["python", os.environ.get("PYMAJOR","3"), ".", os.environ.get("PYMINOR","8")])
```
so we can find it by grep the day we change the default python version.

pmp-p:
@rth please do

hoodmane:
@rth What's the status on this? Can we merge it?

rth:
I still need to fix CI. Something is not quite right. Looking into it.

Issue #1529, opened 2021-04-22
Passing python dict into a js function


oeway:
Hi, I just noticed that we now (v0.17.0) convert python dict to a `Map` in js, and it break my previous code.

I have 3 questions:
1. What would be the easiest way to send a dictionary in Python to Js so it will become a regular object. 
2. What is the reason to choose converting `dict` to `Map`? I find it's bit inconvenient if we get a Map instead of an object by default. I mean, `Object`/`{}` is obviously more commonly used than `Map` in JS, wouldn't it be a better default type for converting `dict` in Python?
3. Previously, if we import js function (e.g. `postMessage`) into Python, then send a python dictionary( `js.postMessage({'data': 123})`, it automatically convert the arguments into js object, but now it complains that `JsException: DataCloneError: Failed to execute 'postMessage' on 'DedicatedWorkerGlobalScope': [object Object] could not be cloned.`.

hoodmane:
Hi @oeway, thanks for the question. We didn't have time to vet this as well as I would have liked and so there are some usability issues. Ideally it would be good to have a careful discussion about each detail like this, but since we have such limited developer time available on this project, there's inevitably going to be things that we don't consider carefully enough.

3. You can do `js.postMessage(**{'data': 123})` or `js.postMessage(data= 123)`.

1. If the object is the last argument:
```py
jsfunc(arg1, arg2, **my_dict)
```
If the object is not the last argument:
```py
from js import Object
from pyodide import to_js
jsfunc(Object.from_entries(to_js(my_dict)), arg2)
```

2. There are several reasons pointing in each direction. The main reason I chose `Map` is that Objects only can have strings as keys, whereas `Map` can have anything. (This isn't as beneficial as I thought though, because only strings or numbers key on values, everything else keys on object identity.) Going from only Strings to Strings or Numbers isn't necessarily enough of a benefit to make it worth the annoyance.

Also, on a PyProxy of a dict you would look up a key with `proxy.get(key)`. If we convert it to a map, then `proxy.toJs().get(key)` does the same thing. It seemed nice to keep the `getattr`/`getitem` distinction if possible.

Also, `pyodide.toPy(my_dict.toJs())` will give back a dict equal to the original one this way, whereas with an object, `{ 1 : 2}` will turn into `{"1" : 2}`.

But at the moment I am thinking it might have been a better idea to keep the `Object`.

hoodmane:
[...]

PR #1573, opened 2021-05-03
ENH Destroy proxies created in JsProxy.__call__ when done with call

hoodmane:
This changes the calling convention for calling a Javascript function from Python so that any proxied arguments are automatically destroyed at the end of the function call. This is a breaking change. It should make it much more convenient to use the foreign function interface. 

The docs need to be updated to explain the new calling convention before this can be merged. We should put a link to the appropriate section in the docs into the error message emitted when a borrowed proxy is used after it is free so that people have an easy time learning the model by trial and error.


rth:
> The docs need to be updated to explain the new calling convention before this can be merged.

Thanks! Even for reviewing, updating examples in the docs would make it more clear the implications of this change.

hoodmane:
> Even for reviewing

Even for designing...

The example that is giving me trouble right now is:
```js
async function test(x){
	return x;
}
pyodide.runPython(`
from js import test
test({})
`);	
```

hoodmane:
The example that is giving me trouble right now is:
```js
async function test(x){
	return x;
}
pyodide.runPythonAsync(`
from js import test
test({})
`);
```
It seems I don't actually understand what was happening when you run this code even before this change.

hoodmane:
Okay I am pretty convinced that asynchronously returning the argument when it's a PyProxy has to be either an error or a memory leak. @rth are you okay with making it an error? On the one hand, it's a confusing behavior, on the other hand, the code is actually a mistake.

One can always do:
```py
proxy = create_proxy(x)
y = js_return_argument(proxy)
assert x is y
proxy.destroy()
```
if it's necessary for some reason.

hoodmane:
In general, there isn't really a good way to return Python objects from Javascript right now. E.g., suppose I want to make a function that takes a pair of PyProxies and makes them into a Python list:
```
function f(x, y){
	return pyodide.toPy([x, y]);
}
```
this leaks the list. We need something like:
```
function f(x, y){
	return pyodide.toPy([x, y]).destroy_after_conversion();
}
```

hoodmane:
@rth okay I added some explanation. The stuff about memory management is getting a bit disorganized though.

docs/usage/type-conversions.md
```diff
@@ -589,23 +697,24 @@ console.log(result); // [1, 6, 7, 14]
 ```
 
 ### Calling Javascript functions from Python
```
rth, line 699: In terms of sections I find it a bit confusing to have this title here while we had,

 - "PyProxy memory management calling Javascript functions from Python"

just before. Generally maybe it would be better to create a separate page for more advanced memory management  topics that it linked from within the type translations (same as is done for making a release from under contributing)? As otherwise this page is getting both long and dense.

rth (commented):
I have not yet reviewed the implementation in detail, but in term of API it looks nice and it does look like it would solve usability issues. 

What I'm wondering about is,
 a) how often would one have exiting methods (or functions) that store input arguments and that would break with this change.
 b) if there functions/methods in external packages that currently work and that would need to be re-implemented with this change.

Generally I think it would be good to send a mail to the mailing list with this proposed change (and https://github.com/pyodide/pyodide/issues/1529#issuecomment-825542386 illustrated on a few examples) to get more feedback/

docs/usage/type-conversions.md
hoodmane: > create a separate page for more advanced memory management

This is probably a good idea. I agree that the organization is getting a bit messed up. I have another PR planned to reduce the memory management burden in another way which would also require changes to the docs generally reducing the complexity of memory management in typical use cases.

Anyways my thought was that if/when these PRs are accepted we can go back afterwards and do a global edit on the docs to make it flow reasonably again. At least I don't want to try too hard to make the docs read well if we haven't decided for sure to make the change.

hoodmane:
> send a mail to the mailing list with this proposed change ... illustrated on a few examples

Sounds good, I will do that when I get around to it.

> how often would existing methods (or functions) ... break with this change.

I think a significant number of methods will break but it should be possible to make the process of fixing most of them quick and systematic -- either `create_proxy` is needed at the Python call site or the Javascript function needs to use `PyProxy.clone`. One thing that could help is to put the Python stack trace of the Javascript call that produced the problem into the error message. We wouldn't want to do this in general I don't think, we would probably want to have a debug-mode. This should make it easier to track down the source of the offending issues.

rth:
> I think a significant number of methods will break but it should be possible to make the process of fixing most of them quick and systematic

The question is also with external packages. Not an accurate example, but just to give an idea. Say,
```py
x = numpy.asarray(x)
```
worked before with JsProxy and returned it unchanged. Now it will break because it returns a borrowed object. We can't realistically ask users to re-implement numpy functions (or other functions from popular libraries). 

This particular example is not valid, but I wonder how often something like this would be happening.

hoodmane:
Well for instance you could consider a function like this:
```js
function appendToList(l, x){
   l.append(x);
   return l;
}
```
and then use it like
```py
from js import appendToList
l = [1, 2]
l = appendToList(l, 3)
l = appendToList(l, 4)
```
Previously this would leak a bunch of l proxies, but now it's fine. Well actually it's still not fine because we leak `l.append`. My hope is to fix this in a future PR by "borrowing" `l.append` from `l` and destroying it automatically with `l`. (Of course there is absolutely no reason you would ever write this code but for example.)

On the other hand, the following function:
```js
function appendToList(l, x){
   let k = l.copy();
   k.append(x);
   return k;
}
```
is going to leak a copy of k (in addition to input l.copy and output k.append). Currently it's actually impossible to write this so that it doesn't leak, we need something like
```js
function appendToList(l, x){
   let l_copy = l.copy; 
   let k = l_copy();
   let k_append = k.append;
   try {
	  k_append(x);
      return k.destroy_after_return(); // This doesn't exist...
   } finally {
     l.destroy(); // needed without this PR but not with it
     l_copy.destroy(); // I intend to remove the need for these in a future PR
     k_append.destroy();
   }
}
```
To make it more like the CPython behavior, we could always destroy a returned proxy. If you want to keep the return value, you should clone it:
```js
function getX(){
   return window.X.clone();
}
```
```py
import js
js.X = [1,2,3]
a = getX()
b = getX()
assert a is b
```
With all three proposed changes, this function would be not leaky:
```js
function appendToList(l, x){
   let k = l.copy();
   k.append(x);
   return k;
}
```
However if you made a second intermediate copy say:
```js
function appendToList(l, x){
   let k = l.copy();
   k.append(x);
   let j = k.copy();
   j.append(x);
   k.destroy(); // destroy k
   return j;
}
```
In this case we need to destroy `k` because it isn't an argument or a return value.

Issue #1548, opened 2021-04-29
Remove PYODIDE_BASE_URL


rth:
I'm wondering if we should remove `PYODIDE_BASE_URL` altogether (and keep it to `"./"` by default).

The upside is that it would make our build / deployment simpler and less error prone. Also, if we ever start using a JS bundler, having to re-run it twice once for test and once for deployment is not ideal.

However, that would mean,
 - that users would always would always have to  pass `indexURL` to `loadPyodide`. It's somewhat longer, but I think this might be preferable anyway to avoid surprising behavior and be explicit about the location where the packages are loaded from. that's also what we currently suggest in the documentation.
 - with webworkers, we would also have to pass `indexURL` in the first `postMessage` so it can be used with `loadPyodide`

WDYT?

rth:
This env variable is only used in console.html now, which is probably fine.

PR #1575, opened 2021-05-04
Use rollup

hoodmane:
This introduces a minimal set of changes to begin using rollup to bundle `pyodide.js`. Once this is merged, I will introduce more PRs to gradually split `pyodide.js` into several files as in #1559.


rth:
I think maybe we should create `src/pyodide-js` already to avoid having package.lock under src/.

To make it work I think you also need to pass `indexURL` to the webworker, so it can use it when loading Pyodide. Otherwise all webworker tests will fail. And if you do that, that would close https://github.com/pyodide/pyodide/issues/1548 and we would need a changelog entry and you can remove all `PYODIDE_BASE_URL` . Unless you see a way not to do that change in this PR?

rth:
Otherwise looking forward to merging this PR!

rth:
> Otherwise all webworker tests will fail.

Or at least as soon as you start using rollup for `webworker.js`, so maybe we could postpone the discussed changes to a follow PR indeed.

src/js/pyodide.js
```diff
@@ -38,7 +38,10 @@ globalThis.loadPyodide = async function (config = {}) {
   // Note: PYODIDE_BASE_URL is an environment variable replaced in
   // in this template in the Makefile. It's recommended to always set
   // indexURL in any case.
-  let baseURL = config.indexURL || "{{ PYODIDE_BASE_URL }}";
+  if(!config.indexURL){
+    throw new Error("Undefined indexURL");
```
rth, line 42: Better to have messages that indicate what the user needs to do to fix it. Maybe something along the lines, 
```suggestion
    throw new Error("Please provide indexURL parameter to loadPyodide");
```

hoodmane: Yup I was being lazy.

hoodmane:
> I think maybe we should create `src/pyodide-js` already to avoid having `package.lock` under `src/`.

I moved `pyodide.js` into `src/js`. I would actually prefer to move the `pyodide-py` folder to `py` because `pyodide-py` and `pyodide-js` and `pyodide.js` share long common prefixes which is annoying for tab completion and similar. Also, everything here is obviously part of `pyodide`...

Currently `package-lock.json` and `package.json` are in the root directory. If you like I can move under `src/js`?

rth:
> Currently package-lock.json and package.json are in the root directory. If you like I can move them under src/js?

Well if we would like to have pyodide-js to be a NPM installable package one day, then we would need a readme for it in the same folder as `package-lock.json` I think. Putting that readme in `src/` wouldn't work because there are multiple associated packages there. So I think it would make sense to put `package-lock.json` directly in the folder where it's going to be in the end i.e. `src/js/`

> I moved pyodide.js into src/js. I would actually prefer to move the pyodide-py folder to py because pyodide-py and pyodide-js and pyodide.js share long common prefixes which is annoying for tab completion and similar. 

OK, sounds good. It's more or less what https://github.com/apache/arrow is doing as well apparently. Though adding a brief README.md to `src/py/` and `src/js/` wouldn't hurt then to make it more obvious what those contain.

hoodmane:
I will add stub readmes for now, we can fill them in better as we go?

rollup.config.js
```diff
@@ -0,0 +1,31 @@
...
+      name: "loadPyodide",
+      file: `${dir}/${input}${minifierSuffix}.${ext}`,
+      format,
+      sourcemap: true,
+    },
+    plugins: [
+      minify
+        ? terser({
+            compress: true,
+            mangle: true,
```
rth, line 19: Probably no need to mangle names, to save <1kB given the size of Pyodide?

hoodmane: Okay.

src/tests/test_pyodide.py
```diff
@@ -3,7 +3,7 @@
 import sys
 from textwrap import dedent
 
-sys.path.append(str(Path(__file__).parents[2] / "src" / "pyodide-py"))
+sys.path.append(str(Path(__file__).absolute().parents[2] / "src" / "py"))
```
rth, line 6: Path.absolute is not a documented/public method in https://docs.python.org/3/library/pathlib.html 

```suggestion
sys.path.append(str(Path(__file__).resolve().parents[2] / "src" / "py"))
```
same for `src/tests/test_console.py`

hoodmane: Huh I never noticed.
```
>>> print(pathlib.Path.absolute.__doc__)
Return an absolute version of this path.  This function works even if the path doesn't point to anything.

No normalization is done, i.e. all '.' and '..' will be kept along. Use resolve() to get the canonical path to a file.
```

rth: https://discuss.python.org/t/pathlib-absolute-vs-resolve/2573

hoodmane:
@rth Perhaps in a separate PR you could add appropriate info about the build system into these various readmes (we could also use a `pyodide-build/README.md`). I'm a bit biased towards explaining what happens at run time over what happens at compile time.

rth:
> Perhaps in a separate PR you could add appropriate info about the build system into these various readmes

At the same time it might be easier to have consistent docs about the build system (or any other submodule) in the documentation which could be linked to from the readme. IMO it's mostly a question if that readme is going to end up in some package repository in which case it would need to have some that information duplicated unfortunately.

rth:
But generally yes, I agree that we need more docs for pyodide-build.

hoodmane:
I think it's good to have the readmes in the folders. Perhaps they could just include a link to the docs. Problem with the docs is that they are all in the `docs` folder. The READMEs have the advantage that they are close to the files they describe.

rth:
> Perhaps they could just include a link to the docs.

Yeah, that's what I meant.

what-is-all-this.md
```diff
@@ -0,0 +1,23 @@
...
+5. The js/pyodide package which defines the Javascript public API, sets up the
+   process of loading the core/pyodide emscripten application + CPython
+   interpreter, and then completes the bootstrap by injecting the js/pyodide
+   API into the Python `sys.modules`.
+6. The packages directory, which contains a large number of CPython packages
+   built to run in Pyodide.
+
+One of our long-term organizational goals is to carefully organize core/pyodide,
+py/_pyodide, py/pyodide, and js/pyodide to clarify which functionality is not part of the
+tangled web of runtime dependencies that define Pyodide's core behavior.
```
rth, line 23:
```suggestion
py/_pyodide, py/pyodide, and js/pyodide to clarify which functionality is not part of
runtime dependencies that define Pyodide's core behavior.
```

what-is-all-this.md
```diff
@@ -0,0 +1,23 @@
+The Pyodide runtime consists of the following components, sorted in terms of
```
rth, line 1: Maybe `repository-structure.md` or something similar?

rth (approved):
Tanks minor comments otherwise LGTM.

what-is-all-this.md
hoodmane: Not a very accurate name yet but maybe if we name it that we can add content to eventually make it a more complete description of the repository structure. I rather like `what-is-all-this.md` but I understand it might not quite be the tone you're aiming for in technical docs.

Issue #1518, opened 2021-04-21
Matplotlib backend in a web worker


jtpio:
At the moment using `matplotlib` when Pyodide is loaded in a web worker with the following code snippet:

```python
import numpy as np
import matplotlib.pyplot as plt

x = np.linspace(0, 10, 1000)
plt.plot(x, np.sin(x));

plt.show()
```

Gives the following error as it tries to create new elements with the wasm backend:

```
Traceback (most recent call last):
  File "<console>", line 2, in <module>
  File "/lib/python3.8/site-packages/matplotlib/pyplot.py", line 2336, in <module>
    switch_backend(rcParams["backend"])
  File "/lib/python3.8/site-packages/matplotlib/pyplot.py", line 276, in switch_backend
    class backend_mod(matplotlib.backend_bases._Backend):
  File "/lib/python3.8/site-packages/matplotlib/pyplot.py", line 277, in backend_mod
    locals().update(vars(importlib.import_module(backend_name)))
  File "/lib/python3.8/importlib/__init__.py", line 127, in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
  File "/lib/python3.8/site-packages/matplotlib/backends/wasm_backend.py", line 23, in <module>
    from js import document
ImportError: cannot import name 'document' from 'js' (unknown location)
```

There is already a "Caveats" section in the docs mentioning limitations when using Pyodide in a web worker: https://pyodide.org/en/latest/usage/webworker.html#caveats

Maybe there could be another backend for matplotlib that would work in a web worker. Or document a workaround (if that's possible).

hoodmane:
This is indeed possible. I have a plan involving [comlink](https://github.com/GoogleChromeLabs/comlink). If anyone is interested in implementing this, I can explain what I'm thinking.

rth:
I think we should also open an upstream issue at matplotlib, see if there is any interest there and try to get them involved. 

@madhur-tandon Might also have some perspective on this.

hoodmane:
I think the point here is to use the existing implementation as-is but after wrapping dom calls in Comlink proxy. Though it occurs to me that we should just wrap the entire DOM with Comlink and then code on a webworker can transparently use all of the DOM calls, with the caveat that they all become async.

hoodmane:
Once we implement #1503, we can remove the async caveat. 
(I've already started working on #1503, @joemarshall did the hard part, we just have to figure out how to integrate it properly.)

rth:
[...]

PR #508, opened 2019-08-15
Fix Stroke Bug for 0 linewidth

madhur-tandon:

PR #1579, opened 2021-05-05
Revive html5 <canvas> based renderer for matplotlib

madhur-tandon:
This is in accordance with discussion here: https://github.com/pyodide/pyodide/issues/1518

Summary:
The `wasm_backend` is refactored into 3 files. The `browser_backend.py` file contains code that is common to both `wasm_backend` and `html5_canvas_backend`. Both these backends inherit from the common functionality present in `browser_backend.py` and implement only the required differences.

Testing Infrastructure is also added for this new `html5_canvas_backend`.


hoodmane:
Looks nice.

How does this relate to @gzuidhoff's concerns #1555 #1556? Does this address either of them or are the changes orthogonal?

SylvainCorlay:
Congrats on this @madhur-tandon! I hope this goes through.

hoodmane:
My hope is that we'll be able to merge this promptly, but review bandwith is quite limited.

@madhur-tandon Could you give a small example I could run with this locally to see what it looks like?

madhur-tandon:
Sorry @hoodmane, I just want to make sure the CI passes here first.
Btw, it does pass on my local machine. 
`matplotlib` (after it's built) should have a `build` directory (from which we can extract `fonts`)
But it seems like the CI fails precisely here with the error:

`cp: cannot stat 'packages/matplotlib/build/matplotlib-3.3.3/lib/matplotlib/mpl-data/fonts/ttf/*': No such file or directory`

It used to pass before, two years back.
Do you have any idea why it happens currently?

madhur-tandon:
> review bandwith is quite limited.

All this stuff has already been reviewed by Mike (@mdboom) two years back. Only minor changes are made to accommodate the current version of Pyodide.

> @madhur-tandon Could you give a small example?

How do you currently use `matplotlib`? Do you have some sort of frontend (for eg: the `iodide` notebook was one before it got discontinued) where you can see the images?

Otherwise, if you simply run the tests, images should be generated in the `reference-images` folder under `matplotlib`. Btw, the images will be generated but the tests will fail. This is expected. It's only after building matplotlib using `make` that these reference images get copied to pyodide's `build` directory (from which the testing infrastructure can access them).

madhur-tandon:
> How does this relate to @gzuidhof's concerns #1555 #1556? Does this address either of them or are the changes orthogonal?

This work is completely independent of these two issues. But of course we can integrate their solutions to this renderer as well. There should be no problems in doing that.

casatir:
Hi there !

> Do you have any idea why it happens currently?

Packages are not build when `build-core` is executed so you can't add your rule there in the Makefile. Could it be done in `meta.yaml` ?

packages/matplotlib/src/html5_canvas_backend.py
```diff
@@ -103,30 +103,40 @@ def _get_url_async(url, threshold):
 
             def callback(e):
                 if req.readyState == 4:
-                    ref_data = _png.read_png_int(io.BytesIO(req.response))
+                    ref_data = np.asarray(Image.open(io.BytesIO(req.response.to_py())))
                     mean_deviation = np.mean(np.abs(canvas_data - ref_data))
                     window.deviation = mean_deviation
-                    window.result = mean_deviation <= threshold
+                    window.result = bool(mean_deviation <= threshold)
```
hoodmane, line 108: Does this `bool` do anything?

hoodmane:
> All this stuff has already been reviewed by Mike (@mdboom) two years back. Only minor changes are made to accommodate the current version of Pyodide.

I see what you mean: there are a lot of lines affected by your most recent commit but mostly it's because of black. Could you provide a link to the discussion where @mdboom reviewed it in the past?

packages/matplotlib/src/html5_canvas_backend.py
madhur-tandon: Yeah, it converts a `numpy._bool` type to `bool`. If we don't do this, the result would be `[]` instead of `True`. This conversion from `numpy._bool` to `[]` is done automatically by `pyodide` so to avoid that, we explicitly convert it to a `bool`.

madhur-tandon:
> Could you provide a link to the discussion where @mdboom reviewed it in the past?

There is no such discussion per se. But all the commits were reviewed and merged by him only. You can check the `gsoc` branch of this repository (and not my fork's since that is this PR itself). Alternatively, as a final submission after completing this project, I made a Github Gist (https://gist.github.com/madhur-tandon/33d5b3e80e969a199494826e13486e5e) which was approved by him. You could also check each of the Pull Requests associated with the commits of this PR. 

For eg: the Pull request https://github.com/pyodide/pyodide/pull/508 is associated with the commit `Fix Stroke Bug for 0 linewidth`

I think all related PRs and their discussions are available here: https://github.com/pyodide/pyodide/issues?q=label%3A%22GSoC+2019%22+is%3Aclosed

packages/matplotlib/src/html5_canvas_backend.py
hoodmane: > convert `numpy._bool` to `bool`

Would you add an inline comment saying that?

> the result would be `[]`

Actually it seems to make a `PyProxy`, though if you call `to_js` on that it turns it into a length 0 array. This conversion behavior seems to be wrong, there's no way a length 0 array is the correct conversion for a numpy scalar...

madhur-tandon:
Btw, the recent changes (to make the renderer work with the current version of pyodide) only consist of https://github.com/pyodide/pyodide/pull/1579/commits/2d4a785f86992e0adca42f07dcb36360ea9a50f2 commit. This also includes using `Pillow` instead of `libpng` because of update to matplotlib `3.3.3`

Otherwise, all else is old work. The commits after this are linting stuff only.

Btw, we can anyway switch to `wasm_backend` too. By specifying which one to use in the `setup.cfg` file.
So we haven't removed the `wasm_backend`, we can choose which one we want.

packages/matplotlib/src/html5_canvas_backend.py
madhur-tandon: Yeah, that's why I cast it to `bool` first. I will add the comment. Just waiting for CI to finish first.

hoodmane:
> You can check the `gsoc` branch of this repository

Okay great. Then I think as you say, the main thing that needs to be reviewed is your most recent commit which looks fine.

> we can anyway switch to `wasm_backend` too

What in your opinion are the pros and cons of the html5 backend vs the wasm backend? We should add some documentation about this if possible (perhaps in a separate PR).

> How do you currently use matplotlib? Do you have some sort of frontend (for eg: the iodide notebook was one before it got discontinued) where you can see the images?

I don't currently use matplotlib, and I'm not really sure how it is meant to be used. Again would be good if we could get something about how to use it into the docs because I think that the features here are very cool but not particularly discoverable.

madhur-tandon:
> Could it be done in `meta.yaml` ?

Thanks, it worked!

hoodmane:
Firefox tests failing again due to recursion depth problems unrelated to this PR.

Makefile
```diff
@@ -118,9 +120,16 @@ update_base_url: \
 test: all
 	pytest src emsdk/tests packages/*/test* pyodide-build -v
 
+copy-fonts:
+	mkdir -p build/fonts
+
+copy-reference-images:
+	mkdir -p build/matplotlib-test
+	cp -n packages/matplotlib/reference-images/*.png build/matplotlib-test/
```
hoodmane, line 128: We don't want to copy the test images into the `build` folder because then they will be distributed as part of Pyodide releases. You should make a symlink as part of the test and then remove it when the test is done. 

See here:
https://github.com/pyodide/pyodide/blob/8d7659b1e11ef602a1492b637edc557587c67e00/packages/micropip/test_micropip.py#L149

madhur-tandon:
I was waiting for CI to finish. It took quite some time. But the tests associated with this PR pass. So I will be making required changes later today.

Makefile
madhur-tandon: Thank you!

packages/matplotlib/src/html5_canvas_backend.py
madhur-tandon: Done!

madhur-tandon:
@hoodmane `build-packages` fails with not being able to build `scipy`. The error seems to be 
```
from numpy.distutils.core import setup
ModuleNotFoundError: No module named 'numpy'
```

while we do get `Thread 1 built numpy in 859.6 s` before `Thread 3 building scipy` so I am not sure where is the error.

Anyway, we are not able to run the CI because of this. Can you please have a look?

hoodmane:
> ModuleNotFoundError: No module named 'numpy'

This is a sporadic CI failure unrelated to the changes in this PR. There are a significant number of such sporadic failures. I reran the CI, probably it won't happen again this time.

Makefile
```diff
@@ -117,9 +118,12 @@ update_base_url: \
 test: all
 	pytest src emsdk/tests packages/*/test* pyodide-build -v
 
+copy-fonts:
+	mkdir -p build/fonts
```
hoodmane, line 122: This should also be moved to `matplotlib/meta.yaml` post section, above line 40.

packages/matplotlib/meta.yaml
```diff
@@ -31,6 +37,7 @@ build:
     rm -rf $SITEPACKAGES/matplotlib/backends/web_backend
     rm -rf $SITEPACKAGES/sphinxext
     cp $PKGDIR/src/fontlist.json $SITEPACKAGES/matplotlib
+    cp $SITEPACKAGES/matplotlib/mpl-data/fonts/ttf/* $PKGDIR/../../build/fonts/
```
hoodmane, line 40: `$PKGDIR/../../build/fonts/`

Isn't there an environment variable like `$BUILDDIR` that would be more appropriate here? @rth?

Makefile
madhur-tandon: Done!

packages/matplotlib/meta.yaml
madhur-tandon: Yeah, but this isn't available to us according to https://pyodide.org/en/stable/development/new-packages.html#build-post

hoodmane: In that case I guess this is fine for now even though not ideal.

hoodmane (approved):
I am happy with everything here. @rth Do you want to review this?

packages/matplotlib/meta.yaml
madhur-tandon: Maybe we can make it available in a separate PR?

hoodmane: Yeah, @rth will have to weigh in on this. In any case I don't think this PR needs to be held up on this point.

madhur-tandon:
Thanks @hoodmane!
There are some tests that fail on `chrome`, some of them are unrelated to this PR.

For the 2 of them that are related to this PR and fail, the reason behind their failure is
`RecursionError: maximum recursion depth exceeded during compilation` and
`RecursionError: maximum recursion depth exceeded` respectively.

hoodmane:
Those last two do look a bit concerning. I'll rerun it and see if the results are the same.

madhur-tandon:
@hoodmane Have you had a chance to re-run the tests on chrome? Btw, these two tests didn't fail on one of my previous commits. Perhaps it happens because the recursion limit was recently reduced or something?

rth:
Thanks a lot @madhur-tandon for this work! I'll try to review soon.

hoodmane:
> Perhaps it happens because the recursion limit was recently reduced?

Perhaps. One issue is that the low level stack frames are slightly variable in size but the Python stack check is only based on total call depth. Thus, a low level stack overflow can be triggered with 150 big Python function calls, but won't necessarily be triggered with 150 small Python function calls. The Python native solution is to set the stack size conservatively so that Python won't segfault even if the stack is full of large function calls. (Unfortunately for us, Python also allows itself an extra 50 frames beyond the stack depth to give itself space to handle the recursion error).

Anyways this all makes sense in the native situation but it's a bit troublesome in our case when stack space is so limited (especially in chrome).

But why does mathtext parser need 100+ stack frames to parse some simple latex?

hoodmane:
On native matplotlib v3.3.3, the latex parser doesn't seem to need hardly any recursion:
```py
mathtext_demos = { # 8 latex snippets copied from your tests }
from matplotlib.mathtext import MathTextParser
mathtext_parser = MathTextParser("bitmap")
import sys
sys.setrecursionlimit(10)
[mathtext_parser.parse(x) for x in mathtext_demos.values()]
```

Makefile
```diff
@@ -119,7 +119,7 @@ test: all
 
 lint:
 	# check for unused imports, the rest is done by black
-	flake8 --select=F401 src tools pyodide-build benchmark conftest.py docs
+	flake8 --select=F401 src tools pyodide-build benchmark conftest.py docs packages/matplotlib/src/
```
hoodmane, line 122: @rth maybe we should be applying flake8 to micropip too?

madhur-tandon:
Hmm, the test passes locally for me. Let's see if it passes on CI. Else, should we consider reverting the change that reduced the limit?

hoodmane:
I reduced the limit because there was a segfault in
`test_core_python.py::cpython_core[test_json.test_recursion-firefox]`
That test looks like this:
```py
def test_endless_recursion(self):
    # See #12051
    class EndlessJSONEncoder(self.json.JSONEncoder):
        def default(self, o):
            """If check_circular is False, this will keep adding another list."""
            return [o]

    with self.assertRaises(RecursionError):
        EndlessJSONEncoder(check_circular=False).encode(5j)
```
(Presumably this is the test for a recursion error with the fattest stack frames.)
I don't think we can raise the recursion limit unless this test passes.

In any case, the matplotlib latex parser isn't even particularly recursive normally (it runs fine with a very low recursion limit in native) so I think this is actually a problem with the browser port of matplotlib.

madhur-tandon:
> I think this is actually a problem with the browser port of matplotlib.

Hmm, but both these tests pass inside the docker container on my local machine. And the other test that fails i.e. `draw_text_rotated` doesn't even use the latex parser!

The fact that both of them passed on a previous commit perhaps showcases us that it is indeed related to reducing the limit? The correctness of the port shouldn't be the issue because they don't fail on Firefox.

Anyway, what would you suggest we could do about this?

hoodmane:
xfail?

madhur-tandon:
Ok, I shall mark these as xfail but only for `chrome`. Hopefully that should suffice. Will be pushing a commit in a while.
Anyway, what about the other tests that fail on `chrome`? i.e. the tests unrelated to this PR?
There seem to be too many unrelated ones failing too.

hoodmane:
Just for fun could you try the following: 
1. Add `sys.setrecursionlimit(10)` on line 351 of `test_matplotlib.py` and run test locally and check for recursion error
2. If there was a recursion error, raise the recursion limit until you see the test not recursion error, and let me know roughly how big the recursion limit had to be for it to pass

hoodmane:
Don't worry about the other failing tests, the CI doesn't have to pass for us to merge.

hoodmane:
Interesting! Not all recent chrome installations can support a recursion limit as high as 250 so we will need to xfail this.

madhur-tandon:
@hoodmane Can you re-trigger `build-packages`? Thanks!

madhur-tandon:
@hoodmane This should be ready. Maybe you could also re-trigger `benchmark`.

RossSmyth:
Hello. I'm not sure if this is related or not, but would this improve the frame rate when interacting with plots? For a project I am working on, when one pans/zooms the frame rate dumps to less than 1 FPS. Then once it's done loading/calculating it goes back to 60F PS nominally.

madhur-tandon:
I have forced pushed after removing the recent conflicts introduced by the `main` branch.

madhur-tandon:
@hoodmane Can you re-trigger `build-core`?
I just force pushed to resolve conflicts with `Makefile`

hoodmane:
@madhur-tandon Can you use merge for this instead of rebase + force push in the future?

hoodmane:
But we should merge this so you don't keep having to resolve conflicts. I was hoping for @rth to weigh in on it first...

Makefile
rth: yes, but we can do that in a separate PR.

rth (commented):
Thanks a lot for this PR and your work on it @madhur-tandon !

My main question is can we still keep the previous backend as the default, make this one as an opt in for now? Then once there are more users that tried it, after we could switch to it as the default, in a release or two.   That would be a safer migration plan, unless there are technical issues that would prevent us from supporting both?

Also, 
 - for subclassed methods could you please copy the docstrings from matplotlib  (or add a docstring when necessary)
 - related to the previous question, for tests where it makes sense it would be nice to support both matplotlib backends via `pytest.mark.parametrize` (if possible)
 - please add a [changelog entry](https://github.com/pyodide/pyodide/blob/main/docs/project/changelog.md)

packages/matplotlib/src/setup.cfg
```diff
@@ -1,5 +1,5 @@
 [rc_options]
-backend=module://matplotlib.backends.wasm_backend
+backend=module://matplotlib.backends.html5_canvas_backend
```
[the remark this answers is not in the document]
madhur-tandon: Yes, this can be done. Both of them are supported. I will switch it back to `wasm_backend`.

packages/matplotlib/src/browser_backend.py
```diff
@@ -0,0 +1,502 @@
...
+    def onmouseup(self, event):
+        x, y, button = self._convert_mouse_event(event)
+        self.button_release_event(x, y, button, guiEvent=event)
+
+    def onmousedown(self, event):
+        x, y, button = self._convert_mouse_event(event)
+        self.button_press_event(x, y, button, guiEvent=event)
+
+    def onmouseenter(self, event):
+        return
```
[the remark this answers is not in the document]
madhur-tandon: I think you are viewing an outdated file here. The correct one doesn't include the `return` statement. I am going to mark this as resolved.

rth:
A last point is that I think we should generate docs for it with sphinx. We can probably start by generating docs for the current matplotlib backend in a separate PR, maybe as a new subsection under  https://pyodide.org/en/stable/usage/api-reference.html  Then sync this PR with master when that PR is merged. That would also make methods that need docstrings more obvious.

hoodmane:
> we should generate docs for it with sphinx

Yeah some people don't seem to know that matplotlib support exists at all. It'd probably help if there were API docs.

Makefile
madhur-tandon: Resolving since it is to be done in a separate PR.

packages/matplotlib/src/html5_canvas_backend.py
```diff
@@ -0,0 +1,471 @@
...
+
+_capstyle_d = {"projecting": "square", "butt": "butt", "round": "round"}
+
+# The URLs of fonts that have already been loaded into the browser
+_font_set = set()
+
+if hasattr(window, "testing"):
+    _base_fonts_url = "/fonts/"
+else:
+    _base_fonts_url = "/pyodide/fonts/"
```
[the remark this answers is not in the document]
madhur-tandon: This is from the earlier days of Iodide. Basically, if you download all the files related to Pyodide i.e. the `.data` and `.js` files and place them in a directory. Then, if you serve the directory, the path of `matplotlib.js` file would be `https://localhost:8000/matplotlib.js` and for fonts it would be `https://localhost:8000/fonts/cmb10.ttf`. This is what is used in tests as well, that's we the path is only `/fonts/` if we are using the tests.

However, back in the case of Iodide, we used to have all the `.data` and `.js` files in a directory known as `pyodide`. Suppose this `pyodide` directory is inside another directory (call it `A`) and now if we serve directory `A`, the paths for the files would become `https://localhost:8000/pyodide/matplotlib.js` and `https://localhost:8000/pyodide/fonts/cmb10.ttf`.

So this is assuming that when one deploys `pyodide`, instead of serving at the level where all `.data` and `.js` files exist, we are instead serving from one level up where the folder containing all the `.data` and `.js` files are inside a directory called `pyodide`.

But yes, we can change it. Since Iodide is deprecated now. Please let me know what should we choose here.

packages/matplotlib/src/browser_backend.py
```diff
@@ -0,0 +1,502 @@
...
+        190: ".",
+        191: "/",
+        192: "`",
+        219: "[",
+        220: "\\",
+        221: "]",
+        222: "'",
+    }
+
+    def _convert_key_event(self, event):
```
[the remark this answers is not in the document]
madhur-tandon: Can this be done in a separate PR please?

madhur-tandon:
> we should generate docs for it with sphinx.

I think I can do it in a separate PR once this is merged.

packages/matplotlib/src/setup.cfg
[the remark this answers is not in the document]
madhur-tandon: Done!

madhur-tandon:
New updates pushed via merge and not rebase (as requested by @hoodmane)
- wasm backend by default
- changelog entry
- tests written for html5 canvas backend are skipped for wasm backend i.e. test suite is dynamic w.r.t which backend is chosen

Things left:
- decision about the `/fonts/` and `/pyodide/fonts` thingie
- docstrings for matplotlib (can we do this in a separate PR?)
- sphinx docs about matplotlib (again, can we address this in a separate PR?)

madhur-tandon:
@hoodmane @rth Can I have a final pass over the remaining stuff I mentioned about `/fonts` and `/pyodide/fonts` thingie? Apart from this, I have incorporated all changes. I can do docstrings + docs in another PR. Please let me know!
Thanks.

hoodmane:
Okay I'm just gonna go ahead and merge this.

madhur-tandon:
Thanks a lot @hoodmane!
I will add docs / docstrings on the weekend.

rth:
Thanks for your work on this new backend @madhur-tandon !

winniederidder:
Is this renderer operational and useable within a WebWorker context? Trying to import matplotlib still yields an ImportError on `from js import document`. In the original [issue](https://github.com/pyodide/pyodide/issues/1518), @hoodmane mentions wrapping document calls in Comlink.proxy which could solve the document issue, but I can't seem to find more information on this being completed yet.

ryanking13:
> Is this renderer operational

You can use it with

```py
 matplotlib.use("module://matplotlib.backends.html5_canvas_backend")
```

> useable within a WebWorker context

No, it has same restriction as original wasm_backend.

> wrapping document calls in Comlink.proxy which could solve the document issue, but I can't seem to find more information on this being completed yet.

AFAIK, No core dev of Pyodide is working on this topic.

PR #1580, opened 2021-05-05
Split more functionality out of loadPyodide into other files

hoodmane:
This migrates the rest of the stuff out of the the `loadPyodide` function. The definitions of public API functions moved into `api.js`, while the private functions `fixRecursionLimit`, `runPythonSimple`, and `fatal_error` moved to the top of the file. I also added docstrings for these three functions.

### DIFF
--- a/src/js/api.js
+++ b/src/js/api.js
@@ -1,0 +1,297 @@
+import { Module } from "./module";
+
+/**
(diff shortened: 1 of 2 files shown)


rth (approved):
Great to remove those nested function definitions.

There are merge conflicts to resolve though.

PR #1590, opened 2021-05-13
bokeh and dependencies

akloster:
The libyaml package is installed to /usr/local. This works with the docker toolchain, but may break in other setups, and I don't know how to change this to something proper.


packages/bokeh/meta.yaml
```diff
@@ -0,0 +1,18 @@
...
+  version: 2.3.1
+requirements:
+  run:
+    - distutils
+    - numpy
+    - Jinja2
+    - pillow
+    - python-dateutil
+    - six
+    - typing-extensions
```
hoodmane, line 12:
```suggestion
    - typing_extensions
```

packages/typing_extensions/meta.yaml
```diff
@@ -0,0 +1,9 @@
+package:
+  name: typing_extensions
```
rth, line 2: The package name is 
```suggestion
  name: typing-extensions
```
https://pypi.org/project/typing-extensions/

rth (commented):
> The libyaml package is installed to /usr/local. This works with the docker toolchain, but may break in other setups

Thanks @akloster! You don't necessarily need to run `make install`. running `make` and then pointing to the built objects should be enough (or otherwise you can always change the install dir to a local folder). That would also avoid using OS dependent system dirs (putting libraries compiled for the wasm target in `/usr/local` on `x89_64` is not something we can leave, given that users could also build it outside of Docker).

See how it's done for libxml for instance https://github.com/pyodide/pyodide/blob/main/packages/libxml/meta.yaml

gabelerner:
@akloster @rth is there a chance the comments are incorporated soon and this PR gets merged? i have a use case that i'd like to try this for. if we don't expect movement on this PR, i'll try to open my own taking the comments into account

rth:
I addressed the few remaining comments and merged.

akloster:
Thank you, I didn't have time to finish it up...

Issue #328, opened 2019-02-27
npm package for pyodide


Kreijstal:
I've seen there has been a lot of progress with this project, any ideas about making an npm package about it, soon?

mdboom:
That's a great idea.  I don't know if I will have the bandwidth soon, but I certainly would be open to pull requests.

rth:
There is a in implementation in https://github.com/iodide-project/pyodide/pull/792 though it would require some work to integrate it.

rth:
So we are progressively moving toward a npm package for pyodide.js thanks to @hoodmane  (https://github.com/pyodide/pyodide/pull/1575, https://github.com/pyodide/pyodide/pull/1578). 

@jtpio @bollwyvl If you have any thoughts on this or design constraints given your work in jupyterlight please let us know.

bollwyvl:
epic work, :clap: @hoodmane :cake: 

aww, thanks for thinking about us. @jtpio may have more thoughts, but here are some off the top of my head, without much filter, so watch your toes:

I don't know if we have a lot of _constraints_ other than "works in browser," which i assume is table stakes.

With webpack 5, there are a lot fewer worries about things, as it's much easier to _copy a file into place_ and have it named correctly and predictably, without webpack even looking at it. See OOM errors, etc.

The biggest thing that would help us, specifically, would be if it shipped with first-party `"types": "lib/index.d.ts"`, so that we could more confidently upgrade inside of an already-complex monstrosity, but that may be out-of-scope for an initial release, as types that aren't tested are probably worse than no types at all. The lightest touch for this would be jsdoc annotations which can [generate the required file](https://www.typescriptlang.org/docs/handbook/declaration-files/dts-from-js.html).

I don't know if the `.data` files get served correctly by all web hosts as wasm, but I actually haven't seen a lot of problems with it. I don't know how packaging can help, more of a hosting problem, but then `application/wasm` is _still_ provisional, i think...

Sooner rather than later, `jupyterlite` is gonna need (optional, but strongly recommended) ServiceWorker support, and not just because of pyodide wasm loading/caching. This might be something that pyodide could offer (or detect) natively, as an alternate build in the same dist. I haven't dug into it much, but this seems like the biggest win for cutting down both time-to-first interaction and memory footprint in "production" deployments, especially ones with multiple tabs. The are rather complicated to make work in all hosting settings, so maybe best to leave it to downstreams, but maybe have some docs.

[...]

PR #1592, opened 2021-05-14
Some basic types

hoodmane:
In the discussion on #328 the consensus seems to be that people like type definitions. This moves toward generating types using jsdoc as suggested by @bollwyvl in that thread.

I don't have much Typescript experience so I would very much appreciate it if people could test them out. The types are available as CircleCI artifacts here:
https://3914-322074228-gh.circle-artifacts.com/0/root/repo/build/pyodide.d.ts

@bollwyvl said that

> types that aren't tested are probably worse than no types at all

If anyone could contribute tests, that would be very helpful (I don't know what they should look like). 

The main limitation I ran into with this approach is that Typescript supports Typescript types in jsdoc annotations but jsdoc does not. This means that if we want to automatically generate docs, we have to stick to the much more limited syntax that jsdoc supports. For example, `isPyProxy` should be typed `obj is PyProxy` but this approach does not allow that as far as I can tell.


docs/sphinx_pyodide/sphinx_pyodide/jsdoc.py
```diff
@@ -78,29 +78,70 @@ def get_object_from_json(self, json):
         obj.kind = kind
         return obj
 
+    def flatten_suffix_tree(self):
+        """Flatten self._doclets_by_path into a dictionary self.doclets.
```
rth, line 82: If possible it would have good nice to have a test (or possibly even a doctest) just for this function

rth (approved):
Very nice, thanks for working on it @hoodmane !

LGTM, feel free to merge.

bollwyvl:
Hooray!

PR #1593, opened 2021-05-16
MAINT Use rollup instead of EM_JS to include pyproxy.js 

hoodmane:
This is partial work towards including Typescript types for pyproxy.js. Typescript does not understand assignments to `Module`, it wants information to be communicated via `import` and `export`. Therefore, the public functions from `pyproxy.js` that we want type annotations for need to be exported using the `export` keyword. This necessitates switching to importing pyproxy.js via rollup.

There are two commits here, in the first commit I made the manual changes needed preserving indentation in pyproxy.js, and in the second commit I applied the linter to pyproxy.js. Almost every line in pyproxy.js has a change in indentation. These indentation changes are my main motivation for making this into a separate PR.

Currently I apply the macro preprocessor to pyproxy.js with 
```sh
cat $< | $(CC) -E -C -P -imacros src/core/pyproxy.c -o $@ $(MAIN_MODULE_CFLAGS) -
```
As far as I can tell the effect of this is close to identical to what was happening before. In the future, we may consider rearranging things to avoid the need for preprocessing `pyproxy.js` -- it is unusual and has the potential to confuse people.


src/js/pyodide.js
```diff
@@ -5,6 +5,7 @@
 import { Module } from "./module";
 import { loadScript, initializePackageIndex } from "./load-pyodide";
 import { PUBLIC_API, makePublicAPI } from "./api";
+import {} from "./pyproxy";
```
rth, line 8: It this just to check that `./pyproxy.js` exists or does it do something else?

hoodmane:
> "Hide whitespace changes" github option in the review helps there

Wow what a life saver!

hoodmane:
> give `src/core/pyodide.js` a file name that more explicitly says that it's an auto-generated files

I was going to change the file extension to `.jso` but it seems that the Javascript tooling really cares that the file extension should be `.js`. Subject to this constraint, do you have an opinion what the generated name should be?

> I can really see contributors (or even myself a few months from now :)) editing the wrong file by accident otherwise.

To be honest, I have already repeatedly edited the wrong file.

hoodmane:
> So if we wanted to upload an npm package that's not dependent on src/core, does it mean that we would have to run the preprocessor to generate src/js/pyproxy.js before uploading it?

Yes that will work. In this case we'd also upload `pyodide.asm.js` and `pyodide.asm.wasm` I'd assume? Perhaps we could name the generated file `src/js/pyproxy.asm.js` so then the three files that are generated from `src/core` would have a consistent naming style?

src/js/package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "pyodide",
   "version": "0.18.0dev0",
-  "dependencies": {
+  "devDependencies": {
     "prettier": "^2.2.1",
     "rollup-plugin-terser": "^7.0.2"
```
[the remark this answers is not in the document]
hoodmane: So I think the `devDependencies` vs `dependencies` difference really only applies well to node applications: in that case the `dependencies` are the things you need to run the node application and the `devDependencies` are the things you need just to develop or build it. For a browser application, no dependencies are needed to run it so I think everything belongs in `devDependencies`.

src/js/pyodide.js
hoodmane: I am importing `pyproxy.js` for its side effects (namely adding a bunch of stuff to `Module`). [There is a special syntax for importing things for their side effects](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/import#import_a_module_for_its_side_effects_only) so we should probably use that:
```suggestion
`import "./pyproxy";`
```

rth:
> Subject to this constraint, do you have an opinion what the generated name should be?

How about `pyproxy.gen.js` (inspired by https://github.com/golang/go/issues/13560#issuecomment-255619359) or `pyproxy_gen.js` if the first doesn't work?

> I am importing pyproxy.js for its side effects (namely adding a bunch of stuff to Module)
> [ ... ] 
> In this case we'd also have to upload pyodide.asm.js and pyodide.asm.wasm

OK, but we only need to add stuff to Module for typing? I would have though that the pyodide-js npm module should basically only contain the JS code needed to load packages (or some other util JS code) but none of the WASM modules. Wouldn't there be a way to keep `pyodide.js` completely independent from `js/core` as it currently is?

hoodmane:
> Wouldn't there be a way to keep pyodide.js completely independent from core as it currently is?

I would not say that `pyodide.js` is completely independent from `core` right now -- the contact surface area between the two is too large, so I wouldn't say that `core` has an internal interface. We should discuss what the goals of such a separation would be. If we can establish goals then we can work towards achieving them, but I think that the current state of things is that there is not a separation.

The tooling imposes complicated constraints -- emcc, typescript, and jsdoc all have strong opinions and it takes care to reconcile them. Being clear about what sort of separation we are looking for is particularly important due to the already constrained nature of making emcc, typescript, and jsdoc get along.

hoodmane:
Okay, `.gen.js` works fine.

rth (approved):
OK, I played with this a bit locally and I agree with you that it is a useful improvement. And given that there was already indeed a strong coupling between pyodide.js and core, this probably doesn't change the situation much in that respect.  +1 for merging (once the conflicts are resolved). Thanks!

PR #1595, opened 2021-05-19
DOCS Fix year of v0.17.0 release in changelog (#1595) [skip ci]

fuyutarow:


hoodmane:
Oops... @rth I suppose we should cherry-pick this to stable?

hoodmane:
Thanks!

rth:
Thanks! Looks like I was convinced we were still in 2020 when I wrote that :) Yes, we should cherry-pick it.

PR #1596, opened 2021-05-20
Enable all emscripten file systems: IDBFS, NODEFS, PROXYFS and WORKERFS

bollwyvl:
## Elevator Pitch

Enable file system persistence with IndexedDB (`IDBFS`), Node.js (`NODEFS`), blobs (`WORKERFS`), proxy (`PROXYFS`).

Screenshot from 2021-05-19 23-34-01

## Motivation

Continuing the discussion from #328, this would allow a pyodide instance to work more naturally with persistent files. 

Some external use cases:
- persisting files between user sessions
- sharing files between concurrent browser tabs

Some internal use cases:
- caching downloaded wheels and API responses?

## Changes

- [x] add ` -l*fs.js` to `Makefile` to restore all backends
  - ~~considered [other backends](https://emscripten.org/docs/api_reference/Filesystem-API.html#persistent-data), but as this has some impact on build time and size without a clear use case, I have not included them for now, but an eventual API should probably plan for supporting them explicitly~~
    - [x] #1689 enabled node, and i've already though about uses cases for WORKERFS
- [x] #1692 ~~hoist `FS.mount`, `FS.mkdir` and `FS.sync` and `IDBFS` up to the `pyodide` API?~~
  - rather than doing something fancy, this has the highest likelihood of working with other language runtimes
  - as discussed below, the preferred approach is to hoist `FS` to the `pyodide` namespace, pending a better name...
- [x] ~~add a convenience API in e.g. `loadPyodide({mount: "/home/user1"})`?~~
  - a lot of use cases would be covered by "simple" persistence
  - could also add it to `sys.prefix`
  - moved to https://github.com/pyodide/pyodide/issues/1715
- [x] tests?
  - haven't looked into what this will take...
- [x] docs?
  - depends on what features we end up landing...
  - focus  on documenting notional API https://github.com/pyodide/pyodide/issues/1715

## Findings

- you can make as many mount points as you want
- they all get synced into the `MEMFS` (which must exist) with `FS.sync(true, (err)=>{...})` and likewise back out with `FS.sync(false, (err) => {...})`
  - the example guidance suggests sync-in at the start of the process and sync-out before exiting, but this is pretty limiting!

## Dead-ends

The image above was gotten by hacking `main.c`, emulating a naive/broken approach for polling. A better path is being investigated!
```c

  // Create and mount userfs immediately.
	EM_ASM({
		FS.mkdir('/home/jo');
		FS.mount(IDBFS, {}, '/home/jo');
    FS.syncfs(true, function (err) {
      err && console.error('error syncing FROM IndexedDB', err);
    });
    setInterval(function() {
        FS.syncfs(false, function (err) {
          err && console.error('error syncing TO IndexedDB', err);
      });
    }, 5000);
	});

```


bollwyvl:
With the ba6915e API in place, it works from the console:

Screenshot from 2021-05-20 23-05-18

FF doesn't expose `IndexedDB` in private browsing, but  otherwise seems pretty robust.

rth:
Thanks for working on this @bollwyvl: very interesting!

One comment: although persistence is definitely something that would be good in some use cases, in others the lack of persistence with MEMFS might be preferable. Or at least more work would be necessary to manage file system state otherwise. For instance, for teaching Python it's appealing to know that users can re-start the tab and restart from from a clean env if necessary. Same for some [small demos](https://pycqa.github.io/isort/docs/quick_start/0.-try/) where there is no need for persistence. 

So it might be good to support passing `IDBFS`  as a `file_system` parameter (with a default to MEMFS) to `loadPyodide`. Also for Node we would need a different FS anyway.

Another thing I'm curious about is performance. I would imagine that MEMFS would be faster since no actual disk operations need to happen, though it probably also depends on how efficient both implementations are.

hoodmane:
@rth I think the idea here is that you use MEMFS for everything except some particular directory which is mounted as a persistent directory. I believe that the model supported by Emscripten is: no matter what, the main file system is MEMFS. Inside of MEMFS it is possible to set up directories that sync to other file systems.

bollwyvl:
Right, as this PR currently stands, it's an entirely opt-in API, and layers on top of the existing, required `MEMFS`. It's not even possible to mount to any other `*FS` to `/`... perhaps we could provide a more helpful preflight validation error than is presently presented (don't have one handy, but had to go read about it). Similarly, the directory must already exist. But given this is presently more for downstreams, rather than _users_... even though it is available from the JS console (maybe) and within a pyodide session with `import pyodide; pyodide.syncfs`, I don't think anyone will be impacted much... aside from the slightly larger builds (haven't measured). If this is an issue, as mentioned in #328, having a second build would be an option.

From a performance perspective: I don't think it's a big driver in the interactive state, as the sync in-and-out is not automagical in any way. A simple usage might choose to only pay the `IndexedDB` toll (_also_ haven't measured) on startup and shutdown. Of course, on jupyterlite, we're looking to get real crazy with it, so we'd be able to let you know!

rth:
OK, then great! I missed your example. The size impact seem negligible. 

> hoist FS.mount, FS.mkdir and FS.sync and IDBFS up to the pyodide API? 

Could you elaborate on the motivation there? If we can avoid re-exporting and documenting a significant part of Emscripten's File system API (and use it directly instead) it would be preferable IMO.

bollwyvl:
>
>
Again, this is my first time out with the emscripten plumbing. I don't have
a solid feel for what would be available at run time to a downstream or
user if it's not declared within the thing being compiled, but only that
the various *FS have to be compiled in.

If there is a pyodide.Emscripten hanging out someplace I don't know about,
then wrapping these three of the tens of FS members is silly, but for
example, making syncfs into an awaitable makes it seem more like the rest
of the pyodide public API.

>

hoodmane:
We currently haven't exposed the emscripten filesystem as a part of our public API. There is clear desire downstream to be able to directly adjust the filesystem, so I think we should expose it. I think it would make sense to add `FS` to the public API, possibly under a different name. This probably should be done as a separate PR before merging this one.

bollwyvl:
> add FS to the public API, possibly under a clearer name

Sure, that sounds good! I had a look at the typings for emscripten, and they seem widely-used, but pretty out-of-date, but perhaps the `@see` is the best we can do!

> pyodide.Emscripten

Oh, and hiding down at the bottom of the API, i did find `pyodide._module` :blush: might have to play with that a little downstream to get an idea of what we might be dealing with... though of course `IDBFS` is not available there... in the meantime, I'll roll back the API changes here, so it's _just_ that one line to the `Makefile`.

rth:
@bollwyvl Can this be marked as "Ready for review" so it can be merged? Thanks!

bollwyvl:
I'll also work up a PR just exposing the as-documented `FS` API, not sure if that's blocking for this...

bollwyvl:
Ar the fails actually indicating anything? 

Also, regarding the notional `FS` PR: I've apparently lost the thread on how to compile the project on a fairly stock ubuntu 20.04... seeing some errors like:

<details>
<pre>
configure: ./configure --host=wasm32-unknown-linux --prefix=~/pyodide/cpython/build/libffi/target --enable-static --disable-shared --disable-dependency-tracking --disable-builddir --disable-multi-os-directory --disable-raw-api
checking build system type... x86_64-pc-linux-gnu
checking host system type... wasm32-unknown-linux-gnu
checking target system type... wasm32-unknown-linux-gnu
checking for gsed... sed
checking for a BSD-compatible install... /usr/bin/install -c
checking whether build environment is sane... yes
checking for wasm32-unknown-linux-strip... no
checking for strip... strip
checking for a thread-safe mkdir -p... /usr/bin/mkdir -p
checking for gawk... gawk
checking whether make sets $(MAKE)... yes
checking whether make supports nested variables... yes
checking for wasm32-unknown-linux-gcc... ~/pyodide/emsdk/emsdk/upstream/emscripten/emcc
checking whether the C compiler works... no
configure: error: in `~/pyodide/cpython/build/libffi':
configure: error: C compiler cannot create executables
See `config.log' for more details
make[1]: *** [Makefile:120: ~/pyodide/cpython/installs/python-3.9.5/lib/libffi.a] Error 77
make[1]: Leaving directory '~/pyodide/cpython'
make: *** [Makefile:205: cpython/installs/python-3.9.5/lib/python3.9] Error 2
</pre>

where `config.log` just says `configure: exit 77` without anything too out of the ordinary
</details>


I don't know enough about the tea leaves to unstick, so would not hold my breath on it for me.

On the up-side: the Emscripten typings were [recently updated](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/emscripten/index.d.ts)...

bollwyvl:
Ah, more digging:
```
configure:3862: ~/pyodide/emsdk/emsdk/upstream/emscripten/emcc -O3 -fPIC -DPYODIDE_FPCAST  -L~/pyodide/cpython/build/libffi/target/lib -O3 -s EXPORTED_FUNCTIONS=_main,_malloc,_free -s ALLOW_TABLE_GROWTH conftest.c  >&5
emcc: error: setting `EXPORTED_FUNCTIONS` expects `<class 'list'>` but got `<class 'str'>`
```

hoodmane:
The CI failures so far are just the normal ones. CircleCi is a bit flakey for us, the webworker tests in particular fail randomly all the time.

Is there a reason you don't want to use the docker image to do the local install? It should prevent these issues. Though it's good for people to try to build natively occasionally so that we can update the instructions.

hoodmane:
What does `~/pyodide/emsdk/emsdk/upstream/emscripten/emcc --version` say?

hoodmane:
The problem comes from here:
https://github.com/hoodmane/libffi-emscripten/blob/3615097cf819519445cc42bb2fa5c452cddff829/build.sh#L39

bollwyvl:
haha, would not have found that. This is getting me... further...


```diff
diff --git a/cpython/Makefile b/cpython/Makefile
index 4d0399a..2741c4e 100644
--- a/cpython/Makefile
+++ b/cpython/Makefile
@@ -117,7 +117,9 @@ $(INSTALL)/lib/libffi.a :
        rm -rf $(FFIBUILD)
        git clone $(LIBFFIREPO) $(FFIBUILD) --shallow-exclude upstream-base
        source $(PYODIDE_ROOT)/emsdk/emsdk/emsdk_env.sh
-       cd $(FFIBUILD) && git checkout $(LIBFFI_COMMIT) && ./build.sh --pyodide-fpcast && make install
+       cd $(FFIBUILD) && git checkout $(LIBFFI_COMMIT)
+       cd $(FFIBUILD) && sed -i 's/_main,_malloc,_free/["_main","_malloc","_free"]/' build.sh
+       cd $(FFIBUILD) && ./build.sh --pyodide-fpcast && make install
        cp $(FFIBUILD)/target/include/*.h $(BUILD)/Include/
        mkdir -p $(INSTALL)/lib
        cp $(FFIBUILD)/target/lib/libffi.a $(INSTALL)/lib/

```

bollwyvl:
#1692 is up, with the strawman API name of `fileSystem`

bollwyvl:
> Is there a reason you don't want to use the docker image to do the local install? It should prevent these issues. 

Yeah, I have to deal with _all kinds_ of container junk at `$DAY_JOB`, and try to stay pretty clear of it until I need:

- something very custom in CI
- helm

> Though it's good for people to try to build natively occasionally so that we can update the instructions.

Welp, anytime PRs comes from me, they'll at least have _sorta_ built locally, once! I generally try to make the `conda-forge` stack work unless I get very deep into weird things, and indeed, reckon CF and condadide/mambadide will eventually provide immense value...

bollwyvl:
So this is probably as far as this PR would need to go, aside from tests: while it feels like `mocha` tests would likely be the _right_ place to land them, I can't seem to get them to run locally, or on CI... is there someplace else it would make sense to add them?

hoodmane:
Yeah currently the `test-js`is failing on main, I haven't bothered to investigate. Maybe you could write a selenium test? I think something like this could work (not tested):
```py
def test_idbfs(selenium_standalone):
    selenium.run_js(
        """
        pyodide.fileSystem.mkdir('/lib/python3.9/site-packages/test_idbfs');
        pyodide.fileSystem.mount(pyodide._module.IDBFS, { root: '.' }, "/lib/python3.9/site-packages/test_idbfs")
        pyodide.runPython(`
            import pathlib
            p = pathlib.Path('/lib/python3.9/site-packages/test_idbfs/__init__.py')
            p.write_text("def test(): return 7")
            from importlib import invalidate_caches
            invalidate_caches()
            from test_idbfs import test
            assert test() == 7 
        `);
        pyodide.fileSystem.syncfs(false, (res) => assert(() => res === null)); // Save file system to idb
        """
    )
    selenium.driver.refresh()
    selenium.javascript_setup()
    selenium.run_js(
        """
        window.pyodide = await loadPyodide({ indexURL : './', fullStdLib: false });
        pyodide.fileSystem.mkdir('/lib/python3.9/site-packages/test_idbfs');
        pyodide.fileSystem.mount(pyodide._module.IDBFS, { root: '.' }, "/lib/python3.9/site-packages/test_idbfs");
        pyodide.fileSystem.syncfs(true, (res) => assert(() => res === null)); // restore from idb
        pyodide.runPython(`
            from importlib import invalidate_caches
            invalidate_caches()
            from test_idbfs import test
            assert test() == 7 
        `);
        """
    )
```

hoodmane:
Maybe we want to copy `IDBFS`, etc to `pyodide.fileSystem`? I don't think we should export them as `pyodide.IDBFS`, I think it makes more sense to keep them under the `fileSystem` namespace.

What do you think about renaming `fileSystem` ==> `filesystem`? I think it looks a bit better without the capital S.

Could you add a section to the docs explaining how to use this? Many of the parameters are pretty obscure. It would help if you could write a couple of worked examples.

bollwyvl:
> write a selenium test

Thanks, those were _precisely_ the pointers i needed. I had to do a tiny bit more fixturing, and add some `await`s, but the test comes up clean and is semi-instructive.

> What do you think about renaming fileSystem ==> filesystem

I don't feel too strongly on the actual name... I was just trying to ape the current API members, which are all camel-cased. I personally like `FS` (or `fs`), as it is both short, and most like the upstream docs, which, sparse as they are, are _someone else's problem_ :blush:. My issue with `filesystem` as then there's a `js.pyodide.filesystem.filesystems`, which is starting to smell like weird java stuff.

> section to the docs explaining how to use this ... worked examples

Well, that's the rub, i guess... docs are good, but if not tested, not so good. And anyway, I'm obviously learning 95% of this stuff as I go along :blush: .

Once there are nightlies that have this stuff, we'll be able to start banging _really hard_ on it over on jupyterlite, and will be able to report back a lot more gotchas, and would be able to offer a more targeted tutorial.

There's probably also a few places that are actually executed that might be good candidates for first-party use of the public APIs (outside the scope of this PR):
- a `micropip` wheel/wasm cache
- some form of persistence in the console demo, e.g. try to mount an idbfs, start the interpreter in that folder, sync it back out.

hoodmane:
> personally like `FS` (or `fs`), as it is both short, and most like the upstream docs

I agree `FS` could be good for consistency with upstream, though `filesystem` is a bit more descriptive.

> sparse as they are

They are *really* sparse.

> Once there are nightlies that have this stuff, we'll be able to start banging really hard on it over on jupyterlite, and will be able to report back a lot more gotchas, and would be able to offer a more targeted tutorial.

Sounds like a good plan to me.

bollwyvl:
Semi-related: i don't know how deep the `pyinotify` rabbit hole goes, or how useful it would be for the jupyterlite case, but specifically for a "file browser" where folk can upload stuff and start working with it, write new files from python, see them _almost_ instantly, and then easily get them back out, would be pretty awesome, and much less "magical" than the deep `FS.filesystem` stuff that i am contemplating having to do.

bollwyvl:
circle says: 
```
[gw2] [ 72%] PASSED test_filesystem.py::idbfs_persist_code[firefox] 
```
:tada:

rth:
> I agree FS could be good for consistency with upstream, though filesystem is a bit more descriptive.

+1 for `FS` particularly that we are just re-exporting emscripten's API, so we might as well keep their conventions. Also it's more convenient to search in the emscripten repo for `FS.syncfs` rather than knowing that one need to map `filesystem.syncfs` to `FS.syncfs`

bollwyvl:
Set it to FS. As a light indicator, the ipdfs test reads a bit more nicely.

bollwyvl:
Down to just `core-firefox` flaking... guess I'd call that a win!

hoodmane:
Yup CI is doing good enough to merge now if it passes review. You have only 4/6 checks marked, you want this reviewed? A convenience API could probably be added later, might be nice to have something in the docs about this stuff though.

bollwyvl:
>  A convenience API could probably be added later,

Sure, closed that out and opened a ticket.

bollwyvl:
>  have something in the docs about this stuff though.

As mentioned... I haven't been able to _play_ with the API much, but _will_... once there are nightlies with persistence. It'll be a lot easier for me to start rapping on examples once I've fought some of our (self-made) problems on jupyterlite. As it is, the API should have fairly good SEO with `FS`... i had to learn it all by digging around in Other People's Code, anyhow!

To that end, is there a "fast path" for docs-only PRs that wouldn't invoke the entire CI pipeline?

hoodmane:
Putting [skip CI] in the commit won't invoke the pipeline.

hoodmane:
Right updating the docs later makes sense to me. Not much harm in having a feature that nobody knows about for a little while. I guess the fear is that if people get the features they want added in they may never actually come back to update the docs. But it seems fine to me.

bollwyvl:
> people get the features they want added in they may never actually come back 

don't i know it, friend! I added a note to that effect: https://github.com/pyodide/pyodide/issues/1715. Basically, it would be my preference if we could get some idea of what people might _do_ with it (after _doing_ some of it) the Hard Way, make a _better_ API (especially for people actually using pyodide-flavored python) and _land it here_, rather than trying to do it our darned selves on `lite`. Semi-related: we might have to do some `%pip` stuff (to maintain parity with our _other_ upstream), but the hope would be to rapidly turn them there, get feedback from our deployers, then take the lessons learned and propose them to `micropip`... so we don't have to maintain them.

bollwyvl:
So i guess i've argued myself out of the last two bullets, so this is probably ready for review!

hoodmane:
Okay I'll take a look when I have time.

hoodmane (approved):

hoodmane:
If you resolve the merge conflict I will merge this.

bollwyvl:
I'm still having some issues with the node-based tests locally, but everything else appears fine, locally :grin:. Here's hoping!

bollwyvl:
ha, I'll take it:

```
warning: no blob constructor, cannot create blobs with mimetypes
warning: no BlobBuilder
Python initialization complete
Loading distutils
Loading distutils from ../../build/distutils.js
Loaded distutils
  FS
    ✔ no dir (96ms)
    ✔ mkdir
    ✔ made dir (81ms)

  Module
    noWasmDecoding
      ✔ should be false 

  Pyodide
    ✔ runPython
Loading micropip, pyparsing, packaging, distutils
Loading micropip from ../../build/micropip.js
Loading pyparsing from ../../build/pyparsing.js
Loading packaging from ../../build/packaging.js
distutils already loaded from default channel
Loaded micropip, pyparsing, packaging, distutils
    ✔ loadPackage (205ms)
    ✔ micropip.install (1664ms)


  7 passing (16s)


```

bollwyvl:
It looks like `test_idbfs_persist_code` passed on both browsers:

- https://app.circleci.com/pipelines/github/pyodide/pyodide/2923/workflows/60105a3d-7d66-4471-922f-e7a2b145549d/jobs/30224/steps?invite=true#step-102-1060
- https://app.circleci.com/pipelines/github/pyodide/pyodide/2923/workflows/60105a3d-7d66-4471-922f-e7a2b145549d/jobs/30225/steps?invite=true#step-102-882

bollwyvl:
Thanks for working through this with me! Looking forward to exploring how
we can use it to do fun stuff!

hoodmane:
Thanks for contributing!

sylvain-ri:
Guys I can just say that you are doing some magic here 😁
Thanks for developing pyodide, i wished i understand enough to contribute.

PR #1601, opened 2021-05-23
ENH More complete typescript types

hoodmane:
This PR finishes adding types (partial types were added in #1592, #1593 did some prerequisite rearrangement).

To generate the types correctly in `pyproxy.js`, I changed the protocols like `PyProxyLengthMethods` into ES6 classes (they used to just be objects). This was needed because otherwise Typescript makes the protocol into a namespace but I want it to be a type to indicate that an object has certain methods. To compensate for this change, I had to modify `getPyProxyClass` accordingly. This change required a number of syntactical changes.

I also added a bunch of predicate methods to the base `PyProxy` class to test whether the proxy supports certain extra methods. These are needed to use as Typescript type guards, but they probably are also useful by themselves when code needs to do reflection on the proxies.

I don't know whether the exposed types or the type guard predicates are well named, I would be happy for recommendations of improvements to the names -- both the type names and the predicates are public and deserve to be named with care.

JsDoc does not recognize typescript type guards or typescript type definitions, so I added a JsDoc plugin which strips out type definitions and transforms typeguards `x is y` ==> `boolean`. There's one JsDoc typescript parse error that still happens but it doesn't cause any harm to the output.

I also changed the way that we collect javascript API docs to be a bit more robust. In particular, it used to be organized by file. I changed it as follows: 
1. any API part of `globalThis` or marked with `@memberof globalThis` is treated as a global, 
2. any API of a class whose name starts with `PyProxy` is treated as a part of the `PyProxy` API,
3. anything else is assumed to be part of the pyodide API

As before, things marked `@private` are left out.

I also added tests for the type declarations using `tpd`. They are a new CI step.


src/core/pyproxy.js
```diff
@@ -290,40 +321,93 @@ Module.PyProxyClass = class {
     }
     return Module.hiwire.pop_value(idresult);
   }
-  apply(jsthis, jsargs) {
-    return Module.callPyObject(_getPtr(this), ...jsargs);
+  /**
+   * Test whether the :any:`PyProxy.length` getter is available on this PyProxy. A
+   * Typescript type guard.
```
rth, line 326:
```suggestion
   * Check whether the :any:`PyProxy.length` getter is available on this PyProxy. A
   * Typescript type guard.
```

I understand it's used as a type guard, but this also changes [the API PyProxy by adding these methods](https://pyodide--1601.org.readthedocs.build/en/1601/usage/api/js-api.html#pyproxy), doesn't it? Is this something we want and are willing to maintain in the future?  If so we should add an extra changelog entry.

Same for all such methods.

rth (commented):
Very nice! Thanks for working on this @hoodmane ! One comment below otherwise LGTM.

BTW you might be interested to know that emscripten is also moving toward [a better compatibility with bundlers](https://twitter.com/RReverser/status/1398018576062332929)

src/core/pyproxy.js
hoodmane: I think these methods are a nice improvement even without Typescript. The nature of the javascript `Proxy` object makes duck typing hard: the alternative would be to check with 
```javascript
x.has && !pyodide.isPyProxy(x.has)
```
which currently leaks memory, is not obvious, and is much slower than the `hasHas` (not that speed is important for this).

hoodmane: We could have better names for `hasX`. MyPy seems to use `supportsHas`, do you think that is better?

rth: Not just MyPy but also [typing apparently](https://docs.python.org/3/library/typing.html#typing.SupportsBytes). No strong opinions but maybe +1 for `supportsX`

rth (approved):
+1  for merging once methods are renamed to `supportsX`. Thanks!

PR #1604, opened 2021-05-23
ENH Use leading $ to distinguish shadowed PyProxy methods

### LINKED ISSUE #1589: Use leading $ to distinguish shadowed PyProxy methods
I am thinking again of the problem of a Python class with a method that gets shadowed by a PyProxy method. For example:
```js
let t = pyodide.runPython(`
class Test:
   def destroy(self):
      print("destroy!")
Test()
`);
t.destroy(); // should this destroy the PyProxy or call `Test.destroy`?
```
`$` is legal in Javascript identifiers but not in Python identifiers. So I think we should use a trailing `$` for one of these and no trailing `$` for the other:
```js
t.destroy$(); // prints "destroy!"
t.destroy();  // destroys proxy
```
Or maybe the reverse? Or maybe:
```js
t.destroy$py(); // prints "destroy!"
t.destroy$js();  // destroys proxy
```
not sure what the best option here is, but I think using `$` is a sound approach.

hoodmane:
This resolves #1589. It updates the MRO for PyProxy: now `proxy.destroy` always refers to the method PyProxy class method that destroys the proxy, `proxy.$destroy` refers to an attribute or method on the proxied Python object. Similarly, `proxy.get` refers to the `PyProxy.get` method if `type(proxy)` has a `__getitem__` method. Otherwise, it refers to an attribute or method on the proxied object. To unambigiously look up `get` on the proxied object, use `proxy.$get`.

The docs should be updated to explain this.

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -109,7 +109,7 @@ class SeleniumWrapper:
         self.run_js("Error.stackTraceLimit = Infinity;", pyodide_checks=False)
         self.run_js(
             """
-            window.assert = function assert(cb, message=""){
+            window.assert = function(cb, message=""){
                 if(message !== ""){
                     message = "\\n" + message;
                 }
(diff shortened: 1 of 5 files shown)


rth (approved):
Thanks a lot @hoodmane !

PR #1606, opened 2021-05-23
Add support for pyproxy allocation tracing

hoodmane:
A while back we added tests to make sure we weren't leaking hiwire keys. As a special case of this, we also were testing that we weren't leaking JsProxies. In the future, I would like to add tests to show that we aren't leaking PyProxies either. Currently we leak quite a lot of them.

Before we actually add any tests to show that we aren't leaking proxies, I'd like to rework the model of the PyProxy lifetime. The planned changes should make it far easier to actually avoid these leaks while still using a reasonably natural coding style.

This change has no user-facing effect: the tracing is disabled by default and the API to enable it is private and always will be.

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -249,6 +249,15 @@ class SeleniumWrapper:
     def restore_state(self):
         self.run_js("pyodide._module.restoreState(self.__savedState)")
 
+    def get_num_proxies(self):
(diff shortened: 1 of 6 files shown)


rth (approved):
Thanks LGTM. Minor comment: It might help to mention the `trace_pyproxies` option somewhere under contributing docs. Otherwise +1 for merging.

PR #1613, opened 2021-05-29
Rearrange InteractiveConsole stream redirections

hoodmane:
This simplifies the way that `InteractiveConsole` does stream redirections and adds stream redirection for `stdin`. I also moved `banner` from a class method of `InteractiveConsole` to a module level constant `BANNER`.


src/py/pyodide/console.py
```diff
@@ -34,57 +41,50 @@ async def _load_packages_from_imports(*args):
...
+        self.write_handler(text)
+
+    def flush(self):
+        pass
+
 
-    def seekable(self):
-        return False
+class ReadStream:
+    def __init__(self, read_handler, name=None):
```
rth, line 79: Could you please add a docstring? Do you we want them to be public classes? Maybe private would be better?

rth (approved):
Very nice, thanks @hoodmane ! I wasn't aware that contextlib could be used form stdout/stderr redirections.

src/py/pyodide/console.py
hoodmane: Yeah let's make them private.

PR #1615, opened 2021-05-30
Handle mixed case package names

### LINKED ISSUE #1614: micropip.install of builtin packages should be case insensitive
PyPI distribution names are case insensitive per [PEP-0426](https://www.python.org/dev/peps/pep-0426/#name). However micropip seems to be case sensitive when it comes to distributions in builtin_packages. This is problematic when installing a distribution that depends on one of the builtin_packages but refers to it with different casing, since micropip will instead try to install the dependency from PyPI.

To workaround you can set the name with the casing used, e.g.:
```python
import micropip
micropip.PACKAGE_MANAGER.builtin_packages['jinja2'] = micropip.PACKAGE_MANAGER.builtin_packages['Jinja2']
```

I came across this wanting to install the altair charting library in jupyterlite. While display is not working yet I was able to install altair using:
```python
import micropip
await micropip.install('Jinja2')
micropip.PACKAGE_MANAGER.builtin_packages['jinja2'] = micropip.PACKAGE_MANAGER.builtin_packages['Jinja2']
# Last jsonschema version before pyrsistent dependency added (native, not wheel)
await micropip.install("https://files.pythonhosted.org/packages/77/de/47e35a97b2b05c2fadbec67d44cfcdcd09b8086951b331d82de90d2912da/jsonschema-2.6.0-py2.py3-none-any.whl")
await micropip.install("altair")
import sys
sys.setrecursionlimit(255)
```

hoodmane:
I think this fixes #1614. It could use tests though.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -66,7 +66,7 @@ jobs:
           command: |
             ccache -z
             # The following packages are currently used in the main pyodide test suite
-            PYODIDE_PACKAGES="micropip,pyparsing,pytz,packaging,kiwisolver" make
+            PYODIDE_PACKAGES="micropip,pyparsing,pytz,packaging,kiwisolver,Jinja2" make
             ccache -s
 
       - run:
--- a/packages/micropip/test_micropip.py
+++ b/packages/micropip/test_micropip.py
@@ -228,3 +228,18 @@ def test_install_different_version2(selenium_standalone_micropip):
         `);
         """
     )
+
+
+@pytest.mark.parametrize("jinja2", ["jinja2", "Jinja2"])
+def test_install_mixed_case2(selenium_standalone_micropip, jinja2):
+    selenium = selenium_standalone_micropip
+    selenium.run_js(
+        """
+        await pyodide.loadPackage("micropip");
+        await pyodide.runPythonAsync(`
+            import micropip
+            await micropip.install("{jinja2}")
+            import jinja2
+        `)
+        """
+    )
--- a/packages/micropip/test_micropip.py
+++ b/packages/micropip/test_micropip.py
@@ -234,7 +234,7 @@ def test_install_different_version2(selenium_standalone_micropip):
 def test_install_mixed_case2(selenium_standalone_micropip, jinja2):
     selenium = selenium_standalone_micropip
     selenium.run_js(
-        """
+        f"""
         await pyodide.loadPackage("micropip");
         await pyodide.runPythonAsync(`
             import micropip
--- a/pyodide-build/pyodide_build/buildall.py
+++ b/pyodide-build/pyodide_build/buildall.py
@@ -253,6 +253,7 @@ def build_packages(packages_dir: Path, outputdir: Path, args) -> None:
         "import_name_to_package_name": {},
         "shared_library": {},
         "versions": {},
+        "orig_case": {},
     }
 
     libraries = [pkg.name for pkg in pkg_map.values() if pkg.library]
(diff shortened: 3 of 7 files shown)


.circleci/config.yml
```diff
@@ -66,7 +66,7 @@ jobs:
           command: |
             ccache -z
             # The following packages are currently used in the main pyodide test suite
-            PYODIDE_PACKAGES="micropip,pyparsing,pytz,packaging,kiwisolver" make
+            PYODIDE_PACKAGES="micropip,pyparsing,pytz,packaging,kiwisolver,Jinja2" make
```
rth, line 69: I think it would be better to use an existing package for tests, for instance pytz, without adding Jinja2 here.

pyodide-build/pyodide_build/buildall.py
```diff
@@ -261,13 +262,14 @@ def build_packages(packages_dir: Path, outputdir: Path, args) -> None:
...
+            package_data["shared_library"][name.lower()] = True
+        package_data["dependencies"][name.lower()] = [
+            x.lower() for x in pkg.dependencies if x not in libraries
         ]
-        package_data["versions"][name] = pkg.version
+        package_data["versions"][name.lower()] = pkg.version
         for imp in pkg.meta.get("test", {}).get("imports", [name]):
-            package_data["import_name_to_package_name"][imp] = name
+            package_data["import_name_to_package_name"][imp] = name.lower()
+        package_data["orig_case"][name.lower()] = name
```
rth, line 272: Wait why do we care about the original case, here? Using all lowercase would be enough, wouldn't it?

packages/micropip/test_micropip.py
```diff
@@ -228,3 +228,18 @@ def test_install_different_version2(selenium_standalone_micropip):
         `);
         """
     )
+
+
+@pytest.mark.parametrize("jinja2", ["jinja2", "Jinja2"])
+def test_install_mixed_case2(selenium_standalone_micropip, jinja2):
```
rth, line 234: Maybe let's only test the capitalized version (`PyTz`) to make sure it's clearly not the right capitalization, as `selenium_standalone` tests are very slow so we should avoid adding that for cases that are already covered (e.g. `pytz` above)

.circleci/config.yml
hoodmane: pytz's name is all lower case so it doesn't work for this though.

pyodide-build/pyodide_build/buildall.py
hoodmane: We need the original case to figure out what the right file path is for the http request.

packages/micropip/test_micropip.py
hoodmane: It's important that the package name has a capital letter in it though. We can't make requests to `build/jinja2.js` because that isn't a real file, the file is called `build/Jinja2.js`.

rth:
OK, I see the issue, makes sense. thanks! I was thinking to test that `PyTz` correctly loads `pytz` but you are right it's not exactly the same edge case.

LGTM. The `orig_case` name would go await once https://github.com/pyodide/pyodide/issues/791 is implemented where the metadata could look more naturally as,
```py
{
   ...,
   'jinja2': {
       'name': 'Jinja2',
       'version': ...
       ...
    }
}
```
which would get rid of `orig_case` and those individual key / value mappings in general. I'll open a PR soon.

rth (approved):
Assuming CI passes. Thanks!

Please add a changelog entry.

hoodmane:
> individual key / value mappings in general

Yeah transposing the structure would make more sense.

PR #1623, opened 2021-06-04
Include 'audioop' in standard library.

### LINKED ISSUE #801: Standard library 'import audioop' does not work when calling from Javascript
I am new to this project and it looks awesome, but I can't import the 'audioop' library, which is part of the python standard library. As a result, the following code line fails: 
`import audioop`

with error: 
"ModuleNotFoundError: No module named 'audioop'"

Here's a minimal modified HTML example showing the problem that can be pasted into a Code Pen and run. Select a file to run the code. Python successfully loads and runs, and micropip is imported correctly, but an attempt to import the standard audioop library results in a traceback

```
<!DOCTYPE html>
<html>

<head>
        window.languagePluginUrl = 'https://pyodide-cdn2.iodide.io/v0.15.0/full/';
    </script>
    <script src="https://pyodide-cdn2.iodide.io/v0.15.0/full/pyodide.js"></script>
</head>

<body>

 
  <!-- The `multiple` attribute lets users select multiple files. -->
<input type="file" id="file-selector" multiple>
    function stringifyFiles(jsFileList) {
        let myArray = [];

        // manually create a new file obj for each File in the FileList
        for (var i = 0; i < jsFileList.length; i++) {

            let file = {
                'lastMod': jsFileList[i].lastModified,
                'lastModDate': jsFileList[i].lastModifiedDate,
                'name': jsFileList[i].name,
                'size': jsFileList[i].size,
                'type': jsFileList[i].type,
            }

            //add the file obj to your array
            myArray.push(file)
        }
        return JSON.stringify(myArray);
    }

  const fileSelector = document.getElementById('file-selector');
  fileSelector.addEventListener('change', (event) => {
    const fileList = event.target.files;
    console.log(fileList);
     const data = stringifyFiles(fileList)
      languagePluginLoader.then(async function () {
          await pyodide.loadPackage(['pandas', 'micropip'])
            .then(() => {
              pyodide.runPython(`
                
                import micropip
                import audioop
                
                
               `)
              
          });
      });
  });
</script>

</body>
```

P.S. The only documentation I can find on Pyodide implies this is a bug and that the whole Python standard library should be supported. If this is not true and I am asking a newbie question or don't understand the limitations of Pyodide, can someone please point me at the list of which standard libraries are and aren't supported?

ijc8:
Fixes #801.

Effect on build size:
- pyodide.asm.data: 4888450 (-985)
- pyodide.asm.js: 1841020 (+165)
- pyodide.asm.wasm: 11380901 (+30084)

### DIFF
--- a/src/tests/python_tests.txt
+++ b/src/tests/python_tests.txt
@@ -43,6 +43,7 @@ test__osx_support   platform-specific
 test__xxsubinterpreters
 test_abc
 test_abstract_numbers
+test_aifc
 test_argparse  crash-chrome
 test_array
 test_asdl_parser
@@ -74,6 +75,7 @@ test_asyncio.test_windows_events
 test_asyncio.test_windows_utils
 test_asyncore bad ioctl syscall async
 test_atexit
+test_audioop
 test_audit   subprocess
 test_augassign
 test_base64
@@ -426,6 +428,7 @@ test_slice
 test_smtpd
 test_smtplib  bad ioctl syscall 21537
 test_smtpnet
+test_sndhdr
 test_socket
 test_socketserver
 test_sort
@@ -447,6 +450,7 @@ test_structmembers
 test_structseq
 test_subclassinit
 test_subprocess
+test_sunau
 test_sundry
 test_super
 test_support  multiprocessing
@@ -526,6 +530,7 @@ test_uuid subprocess
 test_venv nonsense
 test_wait3  threading
 test_wait4  threading
+test_wave
 test_weakref  threading
 test_weakset
 test_webbrowser replaced
--- a/src/tests/python_tests.txt
+++ b/src/tests/python_tests.txt
@@ -12,7 +12,6 @@
 # - platform-specific: This is testing something about a particular platform
 #   that isn't relevant here
 # - async: relies on async
-# - audioop: Requires the audioop module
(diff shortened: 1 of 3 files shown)


rth:
Thanks @ijc8! 

Could you please also,
 - remove the "audiloop" tag in the list of tests  in  https://github.com/pyodide/pyodide/blob/b401350a2a972ad48a80d8def635c310fa9acbb0/src/tests/python_tests.txt#L47 that currently are skipped as a result, to see it they would pass.
 - add a [changelog entry](https://github.com/pyodide/pyodide/blob/main/docs/project/changelog.md)

src/tests/python_tests.txt
```diff
@@ -44,7 +43,6 @@ test__osx_support   platform-specific
 test__xxsubinterpreters
 test_abc
 test_abstract_numbers
-test_aifc   audioop
```
rth, line 47: No I meant that we still need the `test_aifc`, just without the tag. Otherwise it wouldn't be run. Let me fit it.

ijc8: Got it, I had the mistaken impression that this was a list of exceptions.
Anyway, thanks for merging! Now I can discard my hack for importing `wave`. :smile:

PR #1628, opened 2021-06-05
Adding PyRTL package

mdko:
This PR adds the PyRTL package, a library for hardware design in Python.

I tested it by building and running the added `test_pyrtl.py` file:
Screen Shot 2021-06-05 at 4 15 44 AM

I also ran a local Pyodide console to verify I can import and use it properly:
Screen Shot 2021-06-05 at 4 18 12 AM

### DIFF
--- a/packages/pyrtl/patches/ctypes_and_renderer.patch
+++ b/packages/pyrtl/patches/ctypes_and_renderer.patch
@@ -1,11 +1,11 @@
-commit c5df13161c176b73c8071d1ae3bef51972ec0d97
+commit 30333d25771050779daac12c7779dd4b3836ab1f
 Author: Michael Christensen <chmdko@gmail.com>
 Date:   Sat Jun 5 00:25:52 2021 -0700
 
     Move ctype import for Pyodide
 
 diff --git a/pyrtl/compilesim.py b/pyrtl/compilesim.py
-index 4ce0d45..403efc2 100644
+index 4ce0d45..5b82962 100644
 --- a/pyrtl/compilesim.py
 +++ b/pyrtl/compilesim.py
 @@ -1,6 +1,5 @@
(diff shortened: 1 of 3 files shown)


packages/pyrtl/test_pyrtl.py
```diff
@@ -0,0 +1,39 @@
...
+            x.next |= a
+            y.next |= b
+        with x > y:
+            x.next |= x - y
+        with y > x:
+            y.next |= y - x
+        with pyrtl.otherwise:
+            done |= True
+
+    gcd <<= x
```
rth, line 25: That would be a syntax error unless I'm missing something?

packages/pyrtl/patches/ctypes_and_renderer.patch
```diff
@@ -0,0 +1,95 @@
...
+ from .core import working_block
+ from .wire import Input, Output, Const, WireVector, Register
+@@ -24,6 +22,11 @@ class DllMemInspector(collections.Mapping):
+     """ Dictionary-like access to a hashmap in a CompiledSimulation. """
+ 
+     def __init__(self, sim, mem):
++        try:
++            import ctypes
++        except ImportError:
++            raise PyrtlError('need ctypes installed')
```
rth, line 33: Do you really need to catch the error? Failing with the ImportError might be enough?

rth (commented):
Thanks @mdko! LGTM aside from the comments below, assuming the test passes.

Do you think it would it would be possible to add this patch upstream, at least the part that moves the ctypes imports inside the functions?

packages/pyrtl/patches/ctypes_and_renderer.patch
mdko: Yes, I think it makes sense to fail with the ImportError in most of the cases except when trying to initialize the CompiledSimulation object; I think it's helpful to have at least one message giving the user an idea what do (i.e. use a different type of simulation).

I've updated the patch to just `import ctypes` where needed like you suggest, and now only display the additional PyrtlError in the object initialization.

packages/pyrtl/test_pyrtl.py
mdko: This isn't a syntax error. At that point, `done` and `gcd` are WireVectors, which have `__ilshift__` and `__ior__` (. (among many other methods) defined to do the behind-the-scenes magic of creating netlists behind the scenes

mdko:
First, thank you for your quick feedback! I've updated the patch (see comments above) and ran black on the test.

If you're okay with it, I'd prefer to leave the upstream code as is with respect to the `ctypes` import, since `ctypes` is part of the Python standard library and can normally be expected to be available. I feel like patching is the appropriate way to fit this into the framework.

rth (approved):
Thanks @mdko LGTM.

Issue #1617, opened 2021-05-31
RFC Improve lifetime management for PyProxy attributes


hoodmane:
## Proposal

When you look up an attribute on a PyProxy, if the value of the attribute is another PyProxy then the lifetime of the attribute should be bound to the lifetime of the original object. When all PyProxies that are clones of the original PyProxy are destroyed, all such "borrowed" proxies should be destroyed as well. 

An implementation is present here: #1636

## Justification

It is currently quite annoying to avoid leaking PyProxies when making nested attribute access or method call. If we make a call like: 
```js
let result = proxy.field.method();
```
this will generate a temporary `PyProxy` for `pyproxy.field` and another temporary for `proxy.field.method` then leak both of these. To avoid the leaks, we currently need to say:
```js
let proxy_field = proxy.field;
let proxy_field_method = proxy_field.method;
proxy_field.destroy();
let result;
try {
   result = proxy_field_method();
} finally {
   proxy_field_method.destroy();
}
```
which converts the original single line into 9 lines. The point of this proposal is to make the short code work without leaking memory. As an added advantage, we can avoid creating temporary Python bound methods and improve speed for repeated lookups.

## Dictionary lookups are unchanged

To be clear, this proposal has no effect on dictionary lookups or more generally on the return values from method calls. Methods still return proxies that are owned by the caller. So 
```js
let d = pyodide.runPython({1 : { 1 : 2}});
let e = d.get(1);
d.destroy();
// lifetime of e is not bound to d
let x = e.get(1); // 2
e.destroy();
```
will work as before. The following will still be leaky:
```js
d.get(1).get(1)
```

## Backwards Incompatibility

If before you used code like
```js
let a = proxy.a;
proxy.destroy();
// use a
```
now `a` is borrowed from `proxy` and is destroyed when `proxy` is destroyed. With this proposal you would have to use `clone()`:
```js
let a = proxy.a.clone();
proxy.destroy();
```
An indicative error message can be chosen for when people do this incorrectly.

rth:
Thanks for the proposal! Yes, looking at https://github.com/pyodide/pyodide/pull/1616 the current approach is clearly not ideal.

> now a is borrowed from proxy and is destroyed when proxy is destroyed.

I agree that that this makes more sense overall. The fact that if you destroy the proxy you need to explicitly copy objects you want to continue using is fairly logical. So +1 for this proposal.

rth:
[...]

PR #1636, opened 2021-06-13
Improve lifetime management for PyProxy attributes

hoodmane:
This implements the features proposed in #1617.


src/js/pyodide.js
```diff
@@ -235,11 +235,13 @@ def temp(Module):
...
-  Module.restoreState = (state) =>
-    Module.pyodide_py._state.restore_state(state);
-
   Module.init_dict.get("temp")(Module);
+
+  let _state = Module.pyodide_py._state;
+  let save_state = _state.save_state;
+  let restore_state = _state.restore_state;
+  Module.saveState = () => save_state();
+  Module.restoreState = (state) => restore_state(state);
```
rth, line 244: We now have those in api.js I think? Syncing with main would probably show conflicts here.

src/core/pyproxy.js
```diff
@@ -95,7 +96,7 @@ Module.disable_pyproxy_allocation_tracing();
  * still make "prototype in proxy" be true though.
  * @private
  */
-Module.pyproxy_new = function (ptrobj) {
+Module.pyproxy_new = function (ptrobj, cache) {
```
rth, line 99: Could you please mention cache in the above docstring.

src/core/pyproxy.js
```diff
@@ -188,6 +194,46 @@ Module.getPyProxyClass = function (flags) {
...
 Module.PyProxy_getPtr = _getPtr;
+Module.pyproxy_mark_borrowed = function (proxy) {
+  proxy.$$.borrowed = true;
+};
+
+const pyproxy_cache_destroyed_msg =
+  "This borrowed attribute proxy was automatically destroyed in the " +
+  "process of destroying the proxy it was borrowed from. Try using the 'copy' method.";
+
+function pyproxy_destroy_cache(cache) {
```
rth, line 205: Maybe we could add a doc string with a sentence explaining what this function does. It's non obvious that despite its name it does not always destroy the cache (depending on the refcnt value). Maybe a better name would be `pyproxy_decref_cache` ?

rth (approved):
Thanks @hoodmane , aside for the following comments LGTM.

src/js/pyodide.js
hoodmane: Yeah they are better in a different file.

src/core/pyproxy.js
hoodmane: Yeah I think I named it before I figured out that the cache would need to be refcounted.

src/core/pyproxy.js
hoodmane: Also the existing docstring has a lot of unimportant information about cleaning up callable `PyProxy` objects which is duplicated in comments in the function body, so I'll edit that down.

hoodmane:
Merging master caused a bunch of hiwire leak checks to fail again. Will have to investigate why.

hoodmane:
I think now the only failing tests will be normal flakiness. Will merge if this CI run looks good.

Issue #978, opened 2020-12-30
Update to python 3.9.1


hoodmane:
Python 3.9 is out. It would be good to upgrade. (In particular I have my eye on several convenient extensions to the C API.)
https://docs.python.org/release/3.9.1/whatsnew/3.9.html#porting-to-python-3-9

rth:
It would be good to update, but we need to be sure that all packages we use made a Python 3.9 compatible release. They should be mostly compatible, but it would still be good to avoid issues such as #877 . Also Cython is not forward compatible for packages that include the Cythonized sources in .tar.gz

hoodmane:
Can you make another checklist for this? Probably Python 3.9 is still too newly released. It was released in October, so maybe give six months for packages to add support for it and try to update in April?

rth:
Something we can look into for 0.18.0. The constraint is not so much that packages won't support 3.9 or 3.10 (they usually would mostly without changes) but that for packages released some time ago there might not be binary wheels on PyPi for those python versions.
Which is important if you want to share objects as pickles between native python and Pyodide. So we should probably make an automated summary of packages that have wheels for the the python version we ship.

hoodmane:
It would be good to add a list of instructions for migrating between versions of Python to the development docs when we do this, in interest of explicitly documenting maintenance procedures.

rth:
Yes, maybe as a subsection of `development/updating-packages.html` where we should also move https://pyodide.org/en/latest/development/contributing.html#migrating-patches

henryiii:
FYI, here's a quick n'dirty snippet to see what the current status is, using pipx & Python 3.10:

```python
from pathlib import Path
import subprocess

packages = [str(p.parent) for p in Path(".").glob("*/meta.yaml")]


def check_pkg(package: str) -> dict[str, bool]:
    res = subprocess.run(
        ["pipx", "run", "pypi-command-line", "wheels", package],
        capture_output=True,
        text=True,
    ).stdout
    return {
        "found": "Package Type:" in res,
        "universal": "none-any.whl" in res,
        "py39": "cp39" in res,
        "py310": "cp310" in res,
    }

ans = {package: check_pkg(package) for package in packages}

for name, pkg in ans.items():
    match pkg:
        case {"found": False}:
            print(name, "not found")
for name, pkg in ans.items():
    match pkg:
        case {"found": True, "py39": True, "py310": False}:
            print(name, "only has Python 3.9 wheels")
for name, pkg in ans.items():
[...]

Issue #1629, opened 2021-06-07
TST RecursionError maximum recursion in test_zarr


rth:
`zarr/test_zarr.py` is currently consistently failing consistently in Chrome with a recursion error. 

It might be a good occasion to compile Pyodide with `-Os` for smaller stack frame size https://github.com/pyodide/pyodide/issues/1541 and increasing the recursion limit a bit.

mgreminger:
@rth, I've found that the [heuristics](https://github.com/pyodide/pyodide/blob/b401350a2a972ad48a80d8def635c310fa9acbb0/src/js/pyodide.js#L140) that are currently used to set the recursion limit are overly conservative, especially in web workers in Firefox (maybe this is the case for the test_zarr failure as well). For the app I'm working on, I need to load SymPy. I found that SymPy needs a recursion limit of at least 131 to load successfully. If I let Pyodide set the recursion limit,  about 10% of the time, SymPy will fail to load because of too low a recursion limit being set by Pyodide (when using Firefox in a web worker). I now manually set the recursion limit to 200 and I've never had the app fail in Firefox or Chrome due to exceeding the stack limit. With the improved error handling in version 0.17 that raises a fatal error if the call stack is exceeded, I think it would be better to make the default recursion limit a fixed number (200 or even 400, I think 1000 is the default for cpython). We would probably want to change the fatal error message from the [current message](https://github.com/pyodide/pyodide/blob/b401350a2a972ad48a80d8def635c310fa9acbb0/src/js/pyodide.js#L64) to one that includes the possibility that the call stack was exceeded to make debugging a call stack error easier. I'm currently using a try-catch around my python calls that looks for RangeError (chrome) or InternalError (firefox) so that I know if the call stack is exceeded, but it has never exceeded the call stack in my app under quite a bit of testing.

rth:
I agree with you that the current situation need to be improved. We have [many issues](https://github.com/pyodide/pyodide/issues?q=is%3Aissue+is%3Aopen+sort%3Aupdated-desc+recursion)  on this topic.

With the current heuristic on 0.17.0 I get,

| Browser                                | sys.getrecusionlimit() |
|----------------------------------------|------------------------|
| Firefox 89 (Linux)                     | 637                    |
| Firefox Mobile 89                      | 267                    |
| Chrome 89 (Linux), Chrome 91 (Windows) | 279                    |
| Safari 13.1 (Mac OS Catalina)          | 750                    |


[...]

PR #1637, opened 2021-06-14
Update to Python 3.9.5

hoodmane:
Closes https://github.com/pyodide/pyodide/issues/978


rth:
Thanks for working on this @hoodmane !

hoodmane:
A lot of packages are failing to build with the error:
```
[wasm-validator error in module] unexpected true: Imported global cannot be mutable,
```
This comes from wasm-opt. If I remove `--mvp-features` and add `--enable-mutable-globals` the error goes away. I guess I'll try patching emscripten with this...

cpython/patches/0005-add-emscripten-host.patch
```diff
@@ -10,25 +18,25 @@
...
  	| avr-* | avr32-* \
  	| ba-* \
  	| be32-* | be64-* \
-@@ -527,6 +529,9 @@
+@@ -527,6 +529,9 @@ case $basic_machine in
  	asmjs)
  		basic_machine=asmjs-unknown
  		;;
 +	wasm32)
 +		basic_machine=wasm32-unknown
```
rth, line 34: We should really contribute this upstream. Except that I just looked into it and it looks like this file had some major changes in CPython `main`, so I would probably have to try to build Pyodide there just to make sure the patch is correct.

src/js/pyodide.js
```diff
@@ -147,10 +147,7 @@ function fixRecursionLimit() {
     recurse();
   } catch (err) {}
 
-  let recursionLimit = depth / 60;
-  if (recursionLimit > 1000) {
-    recursionLimit = 1000;
-  }
+  let recursionLimit = Math.min(depth / 50, 400);
```
rth, line 150: We should do this in a separate PR for traceability. Related to the discussion in #1629

src/tests/make_test_list.py
```diff
@@ -6,7 +6,64 @@
...
+# - locale: Fails due to limitations in the included locale implementation.
+# - multiprocessing: Fails due to no multiprocessing implementation.
+# - fs: Fails due to virtual filesystem issues.
+# - nonsense: This functionality doesn't make sense in this context. Includes
+#   things like `pip`, `distutils`
+# - crash: The Python interpreter just stopped without a traceback. Will require
+#   further investigation. This usually seems to be caused by calling into a
+#   system function that doesn't behave as one would expect.
+# - crash-chrome: Same as crash but only affecting Chrome
+# - crash-firefox: Same as crash but only affecting Firefox
```
rth, line 50: I think we should have this list either in that file or here, not both. Otherwise it's going to be annoying to maintain in sync.

src/tests/make_test_list.py
```diff
@@ -6,7 +6,64 @@
...
+
+"""
+
+
+def collect_old_error_flags():
+    old_error_flags = {}
+    with open(Path(__file__).parent / "python_tests.txt") as fp:
```
rth, line 57: That's for debugging right?

src/tests/test_pyodide.py
```diff
@@ -502,8 +506,8 @@ def h():
               File "<exec>", line 6 in g
               File "<exec>", line 4 in f
               File "<exec>", line 9 in <module>
-              File "/lib/python3.8/site-packages/pyodide/_base.py", line 242 in run
-              File "/lib/python3.8/site-packages/pyodide/_base.py", line 344 in eval_code
+              File "/lib/pythonxxx/site-packages/pyodide/_base.py", line 242 in run
+              File "/lib/pythonxxx/site-packages/pyodide/_base.py", line 344 in eval_code
```
rth, line 510: I don't understand this change. Shouldn't the path be `/lib/python3.9`?

hoodmane: It's better not to have to change the number in too many places every time we update, so I changed `strip_stack_trace` to remove the Python version number from the stack trace.

src/tests/make_test_list.py
hoodmane: This copies over the error markings from the old `python_tests.txt` to the new one (so if there is a new test with the same name as the old test, the new test will xfail for the same reason). My thought is that `make_test_list.py` should be idempotent. I guess it should be changed to just return an empty dict if the file doesn't exist.

src/tests/make_test_list.py
hoodmane: I made it so that `make_test_list.py` writes this header into `python_tests.txt`. `python_tests.txt` is a weird mix of a generated file and a hand modified file...

src/tests/test_pyodide.py
rth: I agree changing all those paths is a bit annoying but `strip_stack_trace` is also fairly surprising. People expect to be able to open the files that are shown in tracebacks, while with this option those files don't exist. So in terms of usability it's not ideal.

hoodmane: It only happens in this one test though?

hoodmane: Plus it refers to a file in the emscripten virtual file system, so it can't necessarily be found in this location on the host operating system...

rth: Aww, for this test sure. I haven't looked at the diff. Nevermind.

src/js/pyodide.js
hoodmane: Opened #1638.

emsdk/patches/fix-wasm-validator-error.patch
```diff
@@ -0,0 +1,34 @@
...
+-  if '--detect-features' not in cmd:
++  if '--detect-features' not in cmd and tool != "wasm-opt":
+     cmd += get_binaryen_feature_flags()
+   # if we are emitting a source map, every time we load and save the wasm
+   # we must tell binaryen to update it
+@@ -1622,6 +1622,7 @@ def run_binaryen_command(tool, infile, outfile=None, args=[], debug=False, stdou
+ 
+ 
+ def run_wasm_opt(*args, **kwargs):
++  kwargs["args"] += ["--enable-mutable-globals"]
```
hoodmane, line 28: I hope this will fix the package builds. We should ask upstream about these errors:
```
[wasm-validator error in module] unexpected true: Imported global cannot be mutable
```

hoodmane:
Scipy build is consistently failing with:
```python
   from numpy.distutils.core import setup
ModuleNotFoundError: No module named 'numpy'
```

hoodmane:
Pandas build also consistently fails with 
```
ModuleNotFoundError: No module named 'numpy'
```

hoodmane:
`numpy` tests pass. Among packages that don't depend on `numpy` all tests pass except for `nltk` which apparently causes a stack overflow on import in both chrome and firefox and `setuptools` which needs to be patched to prevent it from importing ctypes on import.

Packages depending on `numpy` don't build because of a problem with paths or something: they can't find numpy at `packages/.artifacts/lib/python/numpy-1.17.5-py3.9-linux-x86_64.egg`. If I copy `numpy` out of the egg into the folder above, they seem to build correctly. I guess I'll try adding that as a step...

hoodmane:
Scipy build failing with:
```sh
'/root/repo/emsdk/emsdk/upstream/bin/wasm-ld -o build/lib.linux-x86_64-3.9/scipy/spatial/_distance_wrap.cpython-39-x86_64-linux-gnu.so --whole-archive -L/root/repo/packages/numpy/build/numpy-1.17.5/install/lib/python3.8/site-packages/numpy/core/lib/ --fatal-warnings --strip-all build/temp.linux-x86_64-3.9/scipy/spatial/src/distance_wrap.o -Lbuild/temp.linux-x86_64-3.9 -lnpymath -L/root/repo/emsdk/emsdk/upstream/emscripten/cache/sysroot/lib/wasm32-emscripten/pic --no-whole-archive -mllvm -combiner-global-alias-analysis=false -mllvm -enable-emscripten-sjlj -mllvm -disable-lsr --lto-legacy-pass-manager --allow-undefined --import-memory --strip-debug --export-all --no-gc-sections --experimental-pic -shared' failed (1)
```

hoodmane:
The `No module named 'numpy'` error is appearing also in #1639 so it appears to be caused by the setuptools update.

hoodmane:
All packages are building correctly now except scipy and dependencies thereof.

hoodmane:
The problem with `nltk` is caused because `import sqlite3` fatally fails. It expects a C function called `sqlite3_libversion` to exist but it doesn't.

rth:
We could patch nltk to work around that sqlite3 import, but it's likely it would be necessary to investigate the import failure anyway to avoid a regression with this update.

hoodmane:
I don't understand how `src/tests/test_sqlite3` is passing...

hoodmane:
`sqlite3` works correctly after `build-core` but after `build-packages` the symbols go missing.

hoodmane:
So after `build-core`, `pyodide.asm.js` has the expected lines to export the sqlite stuff:
```js
var _sqlite3_libversion = Module["_sqlite3_libversion"] 
= function(){return(_sqlite3_libversion=Module["_sqlite3_libversion"]
= Module["asm"]["sqlite3_libversion"]).apply(null,arguments)};
```
after `build-packages` it has instead:
```js
function _sqlite3_libversion(){return Module["_sqlite3_libversion"].apply(null,arguments)}
```

hoodmane:
Okay so replacing `make` with `make -C packages` in the `build-packages` CI step did make the problem with `sqlite3` go away. No idea what is going on with that though...

hoodmane:
Okay so the biggest remaining issue is fixing the scipy build, in particularly helping it find `-lnpymath`.


Perhaps a better solution should be found for the copy `numpy` out of the egg problem? It'd probably be good to install all of the packages we build locally as we used to do. Or something?

The updates to flags to fix the wasm-validator problem are a bit ad hoc and I don't know what caused the problem in the first place but they make the build work and the tests pass so I think it's okay.


The other changes all make sense IMO.

hoodmane:
I asked about the "imported mutable global" error on emscripten: #14460

hoodmane:
Okay fixed the problem with scipy, of course it was a path issue.

hoodmane:
Statsmodels added support for Python 3.9 in version 0.12 but that requires scipy >= 1.2, so to support statsmodels + Python 3.9 we need to upgrade scipy.

hoodmane:
The problematic code in statsmodel is generated by Cython, so in principle we could regenerate it ourselves by running the build tool they provide to generate Cython files (`statsmodel/tools/cythonize.py`). However, Cythonize requires scipy to run so we'd probably have to add `skip_host: False` to scipy which would add a lot of time to the build.

Probably better just to update scipy.

hoodmane:
`scikit-image` has a similar cython problem.

hoodmane:
With `scikit-image` and `statsmodel` removed for Cython `tp_print` troubles, the remaining packages are all building and tests pass.

hoodmane:
The hypothesis tests are failing in core but they just have to be told to ignore `ZoneInfo` objects.

rth:
Two comments about statsmodels (and scikit-learn),
 - scipy should be [a build dependency](    - scipy >=1.2), so I'm actually not sure why the build wasn't failing so far.
 - The problem with cythonized sources only working for some Python is frequent with PyPi. If you download the same release from a Git tag the build should re-cythonize everything it needs automatically (at least for sure for scikit-learn)

hoodmane:
> If you download the same release from a GitHub tag the build should re-cythonize everything it needs automatically (at least for sure for scikit-learn)

Certainly the github does not include the Cython files so it will try. There are some issues with the path / cython install directory, maybe these wouldn't come up.

Rather than setting `skip-host : False` I am going to try just installing the host scipy with pip.

hoodmane:
@rth Would you be willing to set up the the proper cython stuff? I don't have any experience with these packages so you'd probably have an easier time of it.

rth:
yes, I can have a look later today

rth:
Were the issues at build time or at run time?

If I revert scikit-learn, scikit-image, statsmodels setup to the version on main, and then change the URL to a GitHub tag https://github.com/pyodide/pyodide/commit/3e28f4a89717933887c080e187526da36857de2b they all build fine for me locally as part of this PR.
 - scikit-image tests also pass
 - scikit-learn and statsmodels fail to import which need more investigation.

hoodmane:
At build time. I haven't tested whether they build locally until just recently, I was just looking at the CI.

hoodmane:
> If I revert scikit-learn, scikit-image, statsmodels setup to the version on main, and then change the URL to a GitHub tag 3e28f4a they all build fine for me locally as part of this PR.

Could you push this here?

hoodmane:
On CI just now the Firefox statsmodels import test passes. Did you try to import it in Chrome or Firefox? We marked that test xfail in Chrome.

rth:
Tests for statmodels and scikit-image also passed when installed from GH with no other changes, but only after restarting the CI. So I wonder if it means there is a function signature mismatch somewhere and it fails half the times.

hoodmane:
> there is a function signature mismatch somewhere and it fails half the times.

Is that a characteristic symptom of function signature mismatch? It does seem particularly flaky here.

rth:
Well one of possible causes https://github.com/pyodide/pyodide/blob/f9a68998e54174bb5505ed55bff6d2270d2cb37e/packages/scipy/meta.yaml#L29-L31

Actually maybe we should apply those flags globally to all packages.

hoodmane:
> Actually maybe we should apply those flags globally to all packages.

Sounds logical to me.

hoodmane:
> a randomly selected 50% of calls will fail

Is the use of the word "random" here accurate? Or does it really mean that some consistent fraction of the calls will fail? I don't really understand the mechanism that would cause the randomness here. Seems to me that there is a chance we are just hitting on typical CI flakiness here.

hoodmane:
Okay all tests passed. I think we should proceed with merging this.

src/tests/make_test_list.py
```diff
@@ -33,7 +93,13 @@ def collect_tests(base_dir):
...
+    old_error_flags = collect_old_error_flags()
     tests = collect_tests(TEST_DIR)
     with open("python_tests.txt", "w") as fp:
+        fp.write(explanation)
         for test in tests:
-            fp.write(test + "\n")
+            error_flags = " ".join(old_error_flags.get(test, []))
+            line = test
+            if error_flags:
+                line += "\t" + error_flags
```
rth, line 104: Could we use 4 space instead? Or some number of spaces anyway?
```
```suggestion
                line += "    " + error_flags
```

pyodide-build/pyodide_build/buildpkg.py
```diff
@@ -137,6 +137,13 @@ def compile(path: Path, srcpath: Path, pkg: Dict[str, Any], args):
     if (srcpath / ".built").is_file():
         return
 
+    pre = pkg.get("build", {}).get("pre")
+    if pre is not None:
+        site_packages_dir = srcpath / "install" / "lib" / "python3.9" / "site-packages"
+        pkgdir = path.parent.resolve()
+        env = {"SITEPACKAGES": str(site_packages_dir), "PKGDIR": str(pkgdir)}
+        subprocess.run(["bash", "-ce", pre], env=env, check=True)
```
rth, line 145: In the end this PR doesn't need it, I think? Could you please remove it from this PR and add as a separate PR ideally with a test and updated documentation https://pyodide.org/en/stable/development/new-packages.html#the-meta-yaml-file

rth (approved):
Thanks for doing this work @hoodmane ! 

A few comments otherwise LGTM!

Please also,
 - add a changelog entry
 - don't forget to push the docker image and update `.circleci/config.yml`

rth:
> Is the use of the word "random" here accurate? Or does it really mean that some consistent fraction of the calls will fail? I don't really understand the mechanism that would cause the randomness here. Seems to me that there is a chance we are just hitting on typical CI flakiness here.

I think such issues might explain in part the CI flakiness. It does look like under some conditions emcripten compiled code fails non deterministically, there is some number of such issues in their issue tracker. Also apparently it's a documented behavior for [`va_arg`](https://linux.die.net/man/3/va_arg),
```
If there is no next argument, or if type is not compatible with the type of the actual
next argument (as promoted according to the default argument promotions), random errors will occur. 
```

hoodmane:
> under some conditions emscripten compiled code fails non deterministically

This is certainly good to know. Sad though...

> random errors will occur

Right... when I read these things I usually plug my ears and scream "unpredictable but repeatable". But one thing UB is allowed to do is fail non repeatably. In the `va_arg` example, presumably it depends what is in the memory on the stack next to the last argument, which will vary from run to run based on who-knows-what.

src/tests/make_test_list.py
hoodmane: Before, the file had spaces to ensure vertical alignment and putting tabs approximates visual alignment without having to count spaces. But that isn't all that important of course.

pyodide-build/pyodide_build/buildpkg.py
hoodmane: This might be redundant with `build/script`:

> The script section is required for a library package (build/library set to true). For a Python package this section is optional. If it is specified for a Python package, the script section will be run before the build system runs setup.py. This script is run by bash in the directory where the tarball was extracted.

Though it might be better to use `script` only for libraries and `pre` in other cases. Also we should probably improve documentation about what the working directory is in the various build steps.

PR #1639, opened 2021-06-15
Package updates from Python 3.9 branch

hoodmane:
Part of https://github.com/pyodide/pyodide/pull/1637

### DIFF
--- a/packages/numpy/meta.yaml
+++ b/packages/numpy/meta.yaml
@@ -29,6 +29,7 @@ build:
     # copy the correct numpy config into the build artifacts. Otherwise scipy will try to build with the config
     # from the build computer and bad things will happen
     cp config/* $PKGDIR/../.artifacts/lib/python/numpy-1.17.5-py3.8-linux-x86_64.egg/numpy/core/include/numpy
+    cp -r $PKGDIR/../.artifacts/lib/python/numpy-1.17.5-py3.8-linux-x86_64.egg/numpy $PKGDIR/../.artifacts/lib/python/numpy
 
 test:
   imports:
--- a/packages/Jinja2/meta.yaml
+++ b/packages/Jinja2/meta.yaml
@@ -1,12 +1,12 @@
 package:
   name: Jinja2
-  version: 2.11.3
+  version: 3.0.1
 requirements:
   run:
   - MarkupSafe
 source:
-  sha256: a6d58433de0ae800347cab1fa3043cebbabe8baa9d29e668f1c768cb87a333c6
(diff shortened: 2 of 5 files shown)


packages/numpy/meta.yaml
```diff
@@ -29,6 +29,7 @@ build:
     # copy the correct numpy config into the build artifacts. Otherwise scipy will try to build with the config
     # from the build computer and bad things will happen
     cp config/* $PKGDIR/../.artifacts/lib/python/numpy-1.17.5-py3.8-linux-x86_64.egg/numpy/core/include/numpy
+    cp -r $PKGDIR/../.artifacts/lib/python/numpy-1.17.5-py3.8-linux-x86_64.egg/numpy $PKGDIR/../.artifacts/lib/python/numpy
```
rth, line 32: All those should be py3.9, no?

rth:
> > Packages depending on numpy don't build because of a problem with paths or something: they can't find numpy at packages/.artifacts/lib/python/numpy-1.17.5-py3.9-linux-x86_64.egg. If I copy numpy out of the egg into the folder above, they seem to build correctly. 

Hmm, it's likely due to some change in setuptools.

Normally when we run [`setup.py install --home`](https://github.com/pyodide/pyodide/blob/d1a2778088231dd859443eff53a2adfeef893c13/pyodide-build/pyodide_build/pywasmcross.py#L143) it should have copied it into that folder I think.

I reviewed the [setuptools changelog](https://setuptools.readthedocs.io/en/latest/history.html). Following entries might be relevant but not fully sure about it. There are quite a few items in the changelog.

> v50.0.0
> 
> 20 Aug 2020
> Breaking Changes
>    #2232: Once again, Setuptools overrides the stdlib distutils on import. For environments or invocations where this behavior is undesirable, users are provided with a temporary escape hatch. If the environment variable SETUPTOOLS_USE_DISTUTILS is set to stdlib, Setuptools will fall back to the legacy behavior. Use of this escape hatch is discouraged, but it is provided to ease the transition while proper fixes for edge cases can be addressed.

It vendors distutils from Python 3.9 so we shouldn't need it.

> v49.0.0
> 03 Jul 2020
> Breaking Changes
> #2165: Setuptools no longer installs a site.py file during easy_install or develop installs. As a result, .eggs on PYTHONPATH will no longer take precedence over other packages on sys.path. If this issue affects your production environment, please reach out to the maintainers at #2165.

rth (approved):
Thanks @hoodmane ! Yes, we could try to unvendor distutils from setuptools later.

PR #1642, opened 2021-06-16
Allow Comlink proxies to be imported

hoodmane:
This patches the feature detection code to fix handling of Comlink proxies when the `Comlink` namespace has been registered with `pyodide` via `pyodide.registerComlink`. This issue came up recently in discussion with @aadereiko in gitter.


rth (approved):
Thanks! Looks good. Would you like to add a changelog entry?

PR #1645, opened 2021-06-18
Update to emscripten 2.0.16

hoodmane:
There is an emscripten 2.0.24 but since before we had issues with 2.0.16 we should make sure update to 2.0.16 works first.


hoodmane:
See emscripten-core/emscripten#14488 where I asked how to fix pillow compilation.

rth (approved):
I restarted the CI, to see if failures are deterministic. Otherwise LGTM. Thanks for investigating that issue!

hoodmane:
Yeah I think the failures are all just flaky tests.

Issue #643, opened 2020-05-08
Stop including builds for pure python packages in favor of micropip


rth:
I think it might make sense to stop including builds for pure python packages in this repo, in favor of using micropip to install them from PyPi. This would have the effect that,
 - users get latest updated version as when it's published on PyPi, without requiring additional maintenance effort for pyodide (the handling of versions in micropip might need improvement)
 - CI is slightly faster
 - package maintainers are motivated to publish universal wheels (at least if they want their package to work in pyodide)
 - generally it seems more sustainable long term.

I would also be in favor of reviewing currently listed packages and remove those that are pure python and are not required by an existing dependency with C extensions.

teonbrooks:
@rth, is there a way to easily build a package for local deployment? the last time I tried it wasn't successful...my current use case, I have mne-python included in the supported packages and I am happy to remove it to save on CI build, however, I need the assets built from CI for a project I'm working on. Is it straightforward for local build?

rth:
Hi @teonbrooks , for MNE you are right! For some reason I though that it used C-extensions. Since it already has a pure python wheel on PyPi (https://pypi.org/project/mne/#files), if we remove it, you will be able to just run,
```
micropip.install('mne')
```
and it should work by downloading from PyPi. The build doesn't take long,
```
rg -i building /tmp/build.txt | grep mne
[2020-07-11 12:22:51] Building package mne...
[2020-07-11 12:23:05] done building package mne in 13.9 s.
```
but it would save some work for you as you then won't need to update the version in pyodide (for instance now it's 2 versions behind). Also I noticed that occasionally the import of mne fails (likely due to the size of scipy in pyodide), so not having those false positives in the CI help.

To test that it's working as expected, you could try to pass the full URL already now,
```
micropip.install('https://files.pythonhosted.org/packages/92/ab/9b79f927b599da515335afb4b666a7bb336930a6d8345e7b483a9980a9c1/mne-0.20.7-py3-none-any.whl')
```

rth:
@teonbrooks  After more investigation, there is a bug that prevents installing a package from PyPi if it's included in pyodide https://github.com/iodide-project/pyodide/issues/725 which makes it a bit harder to demo this.

Locally, installing mne from pypi works for mne>0.19.0 as for earlier versions there aren't pure python wheels on PyPi. However there is an issue at import time I think due to the fact that scipy version needs updating,

<details>

```py
>>> import micropip
>>> micropip.install('mne==0.19.0')
[...]

PR #1650, opened 2021-06-22
Stop shadowing plotly package

jmsmdy:
As of the latest release, `plotly==5.0.0`, plotly can be loaded directly from pypi, so there is no need for the pyodide project to maintain its own version. You can verify this here (uses the latest dev version of pyodide): https://codepen.io/jmsmdy/pen/MWpdjVZ

In the interest of keeping the number of packages that are maintained by the pyodide project to a minimum (see  https://github.com/pyodide/pyodide/issues/643), this pull request removes the version of plotly in pyodide/packages so that `micropip.install('plotly')` will pull the latest version from pypi by default.

@nicolaskruchten is a maintainer for plotly who is interested in this.

I am keeping the custom build of retrying, in case it is used elsewhere. Retrying is no longer a dependency of plotly, and I know retrying itself is not being maintained, but it may still be being used by another package I am not aware of. There are not going to be any new versions of retrying coming out, so it doesn't cause any additional work to keep it in the repo without maintaining it (at least until something breaks it).


nicolaskruchten:
Thanks for the long-running interest in this @jmsmdy :)

And to folks on the Pyodide team: we at Plotly are quite interested in what you're doing, please reach out if our packages could be structured better to facilitate your work <3

PR #1656, opened 2021-06-23
ENH Add Ctypes support

### LINKED ISSUE #728: ctypes support
Is there a plan to enable the `ctypes` library in pyodide?  I'd like to use it to call from my pyodide python code into a WebAssembly.  Alternatively, is there some other mechanism I can use to call into a WebAssembly?

Here's why I ask.  I am trying to get shapely, a popular Python package for computational geometry, running in the pyodide environment.  It is a wrapper around the C GEOS library.  I'm able to compile GEOS to a WebAssembly, and to load the shapely package into pyodide.  However, shapely uses the ctypes library when it calls the GEOS functions, and pyodide doesn't seem to support ctypes, so I get runtime errors.  

E.g., if I give pyodide the line `from ctypes import pointer, c_size_t, c_char_p, c_void_p`, I get a `ModuleNotFoundError: No module named 'ctypes'`

The specific pieces of ctypes that shapely uses are byref, cast, c_byte, c_double, c_int, c_uint, c_char_p, c_size_t, c_void_p, pointer, py_object, CDLL, cdll, CFUNCTYPE, POINTER, find_library.  So, most of it!

Any guidance, or pointers to other packages that have done similar things, would be appreciated!

hoodmane:
Resolves #728.

I took [the libffi port](https://brionv.com/log/2018/05/06/emscripten-fun-porting-libffi-to-webassembly-part-1/) started by @brion and added support for structs and closures (and fixed some issues with `long double`). I have that code here:
https://github.com/hoodmane/libffi-emscripten

It currently passes 56/56 of the "call" tests and 84/94 of the "closures" tests. It would be good to set up continuous integration for the library and to upstream it into the emscripten patches.


hoodmane:
I think CircleCI is having an outage or something.

Makefile.envs
```diff
@@ -59,6 +62,7 @@ export MAIN_MODULE_LDFLAGS= $(LDFLAGS_BASE) -s MAIN_MODULE=1 \
 	$(wildcard $(CPYTHONROOT)/build/bzip2*/libbz2.a) \
 	-lstdc++ \
 	--memory-init-file 0 \
+	-s WASM_BIGINT \
```
rth, line 65: The only blocker for bigint was Safari 13.X as far as I can tell, and according to https://caniuse.com/bigint the market share of Safari 13.X is now 1.7% as compared to 12%+ for Safari 14.X, so it might be indeed reasonable to start to start requiring it. We might still want explicit feature checks with explicit errors in `pyodide.js` if it's not available. Maybe we can do that in a separate PR though ?

rth (commented):
Very excited about the progress you have been making on this @hoodmane !

Makefile.envs
hoodmane: Oh yeah I meant to remove that, I added it to see if it would fix the build at some point but it's irrelevant so I removed it but then I reverted the commit that removed it.

hoodmane: I'm not really sure about the ABI differences between using `WASM_BIGINT` and not using it, I guess the libffi port was already handling that difference alright before I got to it. But I'm not even sure which way that I was running it because my system for running the libffi tests is a bit ad hoc. I really need to set up CI for that.

hoodmane:
It built!! Finally!

hoodmane:
Uh-oh now it's failing with `ModuleNotFoundError: No module named '_socket'`.

hoodmane:
Indeed for some reason it didn't build a bunch of built in modules.

hoodmane:
Okay the build in 27df231 actually worked and a basic test of ctypes works:
```py
>>> import ctypes
>>> tuple_size = ctypes.CFUNCTYPE(ctypes.c_long, ctypes.py_object).in_dll(ctypes.pythonapi, "PyTuple_Size")
>>> tuple_size((1,2,3))
3
```
https://6515-322074228-gh.circle-artifacts.com/0/root/repo/build/console.html

hoodmane:
The `CoExtra` test fails because `PyCode_SetExtra` doesn't seem to work at all. Using it corrupts memory. Not ctype's fault. `cpython_core[test_ctypes]` passes on both Firefox and Chrome. Still have to investigate why CLAPACK build is failing...

hoodmane:
Okay so `cpython_core[test_unicode-chrome]` is failing for some reason and CLAPACK build is failing for some reason but other packages build and pass tests, even with ctypes patches removed. So we're in pretty good shape here.

hoodmane:
Right so `test_unicode` uses ctypes to do a bunch of tests for C APIs when ctypes is available. Probably it's again an issue with one of the C string APIs just like the code test was failing because `PyCode_SetExtra` is bad.

hoodmane:
I think the problem with `test_unicode` is that `_PyUnicode_FromFormat` is variadic. I'm not sure how to make call variadic functions from Javascript.

hoodmane:
@rth Any clue what could be causing the CLAPACK build failure on 3f2baf4? The error message is pretty not helpful. I guess CLAPACK doesn't link Python and it doesn't take that long to build so it should be easy to bisect.

hoodmane:
Okay it broke between 66400d0 and 426f5b2. I guess that emsdk patch causes trouble -- those optimization passes I dummied out are probably necessary for the CLAPACK build to succeed. Luckily I can remove the patch because I found out it is caused by the `123n` syntax for a `BigInt` so I can remove uses of that from the libffi implementation and we should be good to go.
emscripten-core/emscripten/issues/14525

hoodmane:
Well removing that patch definitely fixes the CLAPACK build. But removing the bigint `n`'s didn't fix the linking.

rth:
Looks like you found the solution, so I'm not going to be of much help :)

hoodmane:
Scipy imports are failing. I guess I'll restore the patch.

hoodmane:
Okay on 281c721 all tests passed except for two typical `chrome-packages` timeouts.

docs/project/changelog.md
```diff
@@ -56,6 +56,9 @@ substitutions:
 - The standard library module `audioop` is now included, making the `wave`,
   `sndhdr`, `aifc`, and `sunau` modules usable. {pr}`1623`
 
+- Added support for `ctypes`.
```
rth, line 59:
```suggestion
- Added support for the [`ctypes`](https://docs.python.org/3/library/ctypes.html)
  stdlib module.
```

rth (approved):
Great work @hoodmane !

PR #1666, opened 2021-06-25
Use prettier to format html files

hoodmane:


rth (approved):
Very nice! I wasn't aware it also worked on HTML.

Is there any simple way to also run it in pre-commit (and in CI)?

hoodmane:
I think this PR already makes it run in CI and precommit. I updated `make lint` which is used by the CI and `tools/apply-lint.sh` which is used by precommit.

rth:
For CI yes, for pre-commit, it uses `./tools/clang-format-precommit.sh` which is not run on HTML  unless I'm missing something. Actually `./tools/clang-format-precommit.sh` and `tools/apply-lint.sh` have a fair amount of overlap (and risk having slightly different behavior as now); maybe we could remove  `tools/apply-lint.sh` in favor of pre-commit?

PR #1670, opened 2021-06-26
Add command compiler and tests

hoodmane:
More code split off from #1635. This is a version of the `codeop` package that uses `CodeRunner` instead of the built in `compile` function. Since `code.InteractiveConsole` needs to use `codeop`, this is needed in order to swap out the standard compiler for our `CodeRunner`.


rth:
Thanks you @hoodmane !

PR #1673, opened 2021-06-29
Various improvements in core

hoodmane:
I introduced a new set of macros for dereferencing pointers from Javascript and used them everywhere which should make the code clearer. I made it so `BigInt` conversion no longer converts the number to a string and back.

Also, we were setting `checked_ptr` incorrectly in `get_buffer_datatype` due to a typo (I guess there's no test coverage for unchecked buffer assignments).

### DIFF
--- a/cpython/Makefile
+++ b/cpython/Makefile
@@ -27,7 +27,7 @@ BZIP2URL=https://sourceware.org/pub/bzip2/bzip2-1.0.2.tar.gz
 
 FFIBUILD=$(ROOT)/build/libffi
 LIBFFIREPO=https://github.com/hoodmane/libffi-emscripten
-LIBFFI_COMMIT=dd8318b543b92d038da4914fac42d34056a80c7d
+LIBFFI_COMMIT=f671fe3521fb2f383d9b74bd64254b1d3760a31e
 
 all: $(INSTALL)/lib/$(LIB) $(INSTALL)/lib/libffi.a
 
@@ -117,7 +117,7 @@ $(INSTALL)/lib/libffi.a :
 	rm -rf $(FFIBUILD)
 	git clone $(LIBFFIREPO) $(FFIBUILD) --shallow-exclude upstream-base
 	source $(PYODIDE_ROOT)/emsdk/emsdk/emsdk_env.sh
-	cd  $(FFIBUILD) && git checkout $(LIBFFI_COMMIT) && bash ./build.sh ; make install
+	cd $(FFIBUILD) && git checkout $(LIBFFI_COMMIT) && ./build.sh --pyodide-fpcast && make install
 	cp $(FFIBUILD)/target/include/*.h $(BUILD)/Include/
 	mkdir -p $(INSTALL)/lib
 	cp $(FFIBUILD)/target/lib/libffi.a $(INSTALL)/lib/
--- a/cpython/Makefile
+++ b/cpython/Makefile
@@ -133,6 +133,7 @@ $(BUILD)/Makefile: $(BUILD)/.patched $(ZLIBBUILD)/.configured $(INSTALL)/lib/lib
 			  --without-pymalloc \
 			  --disable-shared \
 			  --disable-ipv6 \
+			  --enable-big-digits=30 \
 			  --enable-optimizations \
 			  --host=wasm32-unknown-emscripten\
 			  --build=$(shell $(BUILD)/config.guess) \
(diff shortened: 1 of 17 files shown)


cpython/Makefile
```diff
@@ -133,6 +133,7 @@ $(BUILD)/Makefile: $(BUILD)/.patched $(ZLIBBUILD)/.configured $(INSTALL)/lib/lib
 			  --without-pymalloc \
 			  --disable-shared \
 			  --disable-ipv6 \
+			  --enable-big-digits=30 \
```
rth, line 136: Would be worth adding a comment why we want this.

src/core/python2js_buffer.js
```diff
@@ -181,12 +181,12 @@ JS_FILE(python2js_buffer_init, () => {
...
-    // When indexing HEAP32 we need to divide the pointer by 4
-    let n = HEAP32[bufferData.shape / 4 + curdim];
-    let stride = HEAP32[bufferData.strides / 4 + curdim];
+    // Stride and suboffset are signed, n is unsigned.
+    let n = DEREF_U32(bufferData.shape, curdim);
+    let stride = DEREF_I32(bufferData.strides, curdim);
     let suboffset = -1;
     if (bufferData.suboffsets !== 0) {
-      suboffset = HEAP32[bufferData.suboffsets / 4 + curdim];
+      suboffset = DEREF_I32(bufferData.suboffsets, curdim);
```
rth, line 189: Yes, it is more clear.

src/core/jsmemops.h
```diff
@@ -0,0 +1,49 @@
...
+
+// No BigUint64Array, have to manually split / join lower and upper byte
+//
+#define BIGINT_LOWER(x) (Number((x)&BigInt(0xffffffff)) | 0)
+#define BIGINT_UPPER(x) (Number((x) >> BigInt(32)) | 0)
+#define UBIGINT_FROM_PAIR(lower, upper)                                        \
+  (BigInt(lower) | (BigInt(upper) << BigInt(32)))
+
+#define IBIGINT_FROM_PAIR(lower, upper)                                        \
+  (BigInt(lower) | (BigInt(upper - 2 * (upper & 0x80000000)) << BigInt(32)))
```
rth, line 36: Might be good to remind that in this case BigInt is likely `Number`, right?

src/core/js2python.c
```diff
@@ -115,9 +116,24 @@ EM_JS_NUM(errcode, js2python_init, (), {
...
   Module.__js2python_bigint = function(value)
   {
-    let ptr = stringToNewUTF8(value.toString(16));
-    let result = _PyLong_FromString(ptr, 0, 16);
-    _free(ptr);
+    let value_orig = value;
+    let length = 0;
+    while (value && ~value) {
+      length++;
+      value >>= BigInt(32);
```
rth, line 123: Wouldn't it be faster to compute the length with `Math.log2(value)` instead of byte shifts?

rth (approved):
Thanks generally LGTM, a few comments below.

src/core/js2python.c
hoodmane: It doesn't seem to work:
```js
>>> Math.log2(1471057108571074017501n)
TypeError: can't convert BigInt to number
```

cpython/Makefile
hoodmane: Sounds good. 

The reason is as follows: there are two options for this, 15 and 30. 30 is faster as long as the system has good support for 64 bit arithmetic, otherwise 15 is faster. By default Python uses 15 if the system has 32 bit pointers and 30 if the system has 64 bit pointers. The majority of users are running 64 bit systems and even most systems with 32 bit pointers have efficient 64 bit arithmetic so I think for most machines arithmetic will be faster with this set to 30.

src/core/jsmemops.h
hoodmane: Well `WASM_BIGINT` is not about whether the browser supports `BigInt` (that may be assumed in all cases), it's about whether it supports `BigInt64Array`.

cpython/Makefile
rth: Yeah, but the WASM VM is 32 bit, so I'm not sure we would really see the impact of 64bit systems?

hoodmane: It has 32 bit pointers but it has native 64 bit arithmetic.

src/core/js2python.c
rth: It might be necessary to convert to double
```js
>>> Math.log2(Number(1471057108571074017501n))
```
but `log` is also expensive, so maybe the current version is better.

cpython/Makefile
hoodmane: For instance the following
```C
long long test(long long a, long long b){
    return a * b;
} 
```
compile to native 64 bit arithmetic instructions:
```wat
(func $test (type 0) (param i64 i64) (result i64)
    (local i32 i32 i32 i64 i64 i64)
    ;; omitted preamble
    i64.load offset=8
    local.set 5
    local.get 4
    i64.load
    local.set 6
    local.get 5
    local.get 6
    i64.mul
    ;; omitted postamble
    return
)
```

rth: Yes, indeed, makes sense.

hoodmane:
There seems to have been a significant recent increase in CI flakiness.

Issue #1577, opened 2021-05-05
fpcast emulation tracing and removing


joemarshall:
We currently build with function pointer cast emulation. It is really hard to remove this because it is hard to know where bad function calls are being made (i.e. where an indirect call is made where the call arguments don't match the target function arguments).

I would suggest that what is needed is some kind of fpcast_emulation with tracing build, then run the full test suite on that, hopefully it would hit the majority of the function pointer casting shenanigans. In theory this was fixed in python in 3.8, so we should be good with core python, but we do bundle some other stuff for images and fonts etc. that might cause issues. The patches to scipy/numpy removed most issues there too, but there may be some that I missed. 

I suspect the easiest way to do this is to make a build which in emscripten replaces indirect calls to a call to a python indirect call trampoline, so that rather than call to fpcast versions, it calls to something in imports which is javascript and does the actual call. I think that side of things would need a change to visitCallIndirect in the binaryen pass to just make it call the javascript trampoline with the signature of the indirect call as a parameter. Then the fpcast trampoline could check that against the signature of what it was calling (using the same , and chuck out a log message if they didn't match. This would be way more efficient if webassembly type reflection was enabled otherwise some manual type reflection cunningness would be needed to get the type of the target. Type reflection is behind a flag in chromium (enable experimental webassembly features). 

I imagine it would be slow as a slow thing, but for testing it would do the job I imagine.

I would suggest it isn't much work to update binaryen to call the trampoline, then creating a build that works in chromium with webassembly reflection enabled. It might be that if one can get pyodide itself loading in chromium and run the test suite, that is enough to exercise the code enough to find out what is the cause of the naughty pointer calls and sort it.

I won't have time to do anything till teaching and marking ends (i.e. june/july at earliest), but if anyone wants to do it, I'm happy to input.

rth:
Thanks for opening this issue with a detailed plan @joemarshall !

Personally I also won't have much availability to look into this in the near future, but I very much look forward to getting rid of fpcast emulation. It should also help for performance as far as I understand.

joemarshall:
I had a tiny fiddle with this - interestingly pyodide itself now loads, I can run code in the JS console before the python console.html loads, but I can't run anything in the console without hitting a failure somewhere in jsproxy.

Latest chrome canary plus webassembly debugging however means that
a) you get the exception about null pointer or wrong pointer type
b) you actually can get the code lines and stuff that is failing in the C code now. 

[...]

PR #1677, opened 2021-06-30
Fpcast removal

joemarshall:
This is a first attempt at removing fpcast from the build. Quite a lot of tests pass now, including numpy, scipy.

Closes https://github.com/pyodide/pyodide/issues/1577


joemarshall:
CI seemed to fail on lots of things here, despite working fine on my build machine. Lots of timeouts...

hoodmane:
Well at least it passed on Firefox (the `run_webworker` and `reentrant_error` tests regularly fail on main). It must be crashing chrome. I don't have any good way of understanding what is going on when code works on one browser and totally breaks the other though.

One recent problem I hit is that chrome is more delicate about pointer alignment than firefox. The following code crashes chrome consistently (though it is also noticeably slow on firefox):
```js
let addr = _malloc(8);
let byte_ptr = ptr1 + 3;
let ptr_ptr = addr + 4;
HEAP8[byte_ptr] = 7;
HEAPU32[ptr_ptr/4] = byte_ptr;
_print_byte_ptrptr(ptr_ptr);
```
C:
```C
void print_byte_ptrptr(**char p){
	printf("byte is %x\n", **p);
}
```

hoodmane:
Oops didn't mean to push here, sorry.

rth:
I synced with main after patches were merged in https://github.com/pyodide/pyodide/pull/1708. There were still a few failures even with the latest CI setup that look genuine to me though.

hoodmane:
Well one problem is that ctypes needs to be updated.

hoodmane:
`buffer.to_bytes()` has a problematic fpcast, I probably forgot an unused argument.

hoodmane:
Seems matplotlib has some function call mismatches. In particular, there are a bunch of functions that are declared as `METH_VARARGS` but have signatures as if they are `METH_VARARGS|METH_KEYWORDS`. So the functions take three arguments but ask to be called with only two. Then they ignore the third argument.

ryanking13:
I don't know much about function pointer cast logic in Emscripten, will merging this PR break functions that have mismatching return types?

hoodmane:
>  will merging this PR break functions that have mismatching return types?

I don't think so. The primary issue is when a function is defined to take a certain number of arguments and is cast to a different function type that takes a different number of arguments. If the wasm types don't match, there will also be breakages but the wasm type system is much coarser than the C type system.

hoodmane:
There's still the mystery here of why all the tests time out on chrome. Especially given that they work fine on node...

hoodmane:
Okay so I think I've gotten the problems down to:
1. Almost all chrome tests time out
2. `test_import[scipy]` import times out but all tests that import scipy and do something with it pass
3. `test_emsdk` is failing for some reason

hoodmane:
Oh no! So many `indirect call signature mismatch`!

hoodmane:
Okay, it's because upgrading numpy introduced a bunch of new messed up function signatures. *sigh*

hoodmane:
This last problem seems to already be fixed upstream on main.

rth:
> Okay, it's because upgrading numpy introduced a bunch of new messed

We need to see if they could add some way to check those signature mismatches with extra compilation flags in CI, otherwise it will keep happening.

hoodmane:
Unfortunately I think it's not possible with compilation flags. The problem is, people tend to use a subtype for the first argument:
```C
PyObject* some_method(SomeTypeObject* self, PyObject* args){
	// body
}
```
but then the struct they are putting `some_method` into expects `PyObject* (*f)(PyObject* self, PyObject* args)`. So generally they explicitly cast the function pointer. This prevents `assignment from incompatible pointer type` warnings from appearing even if the function has the wrong number of arguments (which is what causes us trouble).

hoodmane:
After doing some research, it seems like you could build a tool to find these bad casts using a clang / llvm ast walker.

In particular, in a clang ast dump the problematic casts look as follows:
```C
|-CStyleCastExpr 0x5587f3527f50 <line:343:3, col:16> 'PyCFunction':'PyObject *(*)(PyObject *, PyObject *)' <BitCast>
  `-ImplicitCastExpr 0x5587f3527f38 <col:16> 'PyObject *(*)(PyObject *)' <FunctionToPointerDecay>
```
We see we cast a  `PyObject *(*)(PyObject *)` to a `PyObject *(*)(PyObject *, PyObject *)`.

hoodmane:
numpy v1.21.4 seems to have a very large number of function pointer problems that are fixed in numpy tip of tree.

joemarshall:
> After doing some research, it seems like you could build a tool to find these bad casts using a clang / llvm ast walker. In particular, in a clang ast dump the problematic casts look as follows:
> 
> ```c
> |-CStyleCastExpr 0x5587f3527f50 <line:343:3, col:16> 'PyCFunction':'PyObject *(*)(PyObject *, PyObject *)' <BitCast>
>   `-ImplicitCastExpr 0x5587f3527f38 <col:16> 'PyObject *(*)(PyObject *)' <FunctionToPointerDecay>
>         `-DeclRefExpr 0x5587f3527ed8 <col:16> 'PyObject *(PyObject *)' Function 0x5587f3527650 'some_function' 'PyObject *(PyObject *)'
> ```
> 
> We see we cast `some_function` which was defined as a `PyObject *(*)(PyObject *)` to a `PyObject *(*)(PyObject *, PyObject *)`.

Good idea, I only got as far as working out why things didn't work, but didn't work out how to do compiler based checking.

The only pain is 
PyMethodDef in python extensions. The function is always stored in that as pycfunction even if it is fastcall or other calling convention which uses a different signature. 

The ideal would be to make PyMethodDef contain a union of all possible call types and make python use the right Union element when it does the call.  That would mean the only place the seemingly broken casts would be left would be in non-updated module code. But it's a patch to python itself and I don't know how big a one it is.

hoodmane:
> The ideal would be to make PyMethodDef contain a union of all possible call types and make python use the right Union element when it does the call. That would mean the only place the seemingly broken casts would be left would be in non-updated module code. But it's a patch to python itself and I don't know how big a one it is.

I think if they did that, they would make a new `PyMethodDef2` and make new APIs that use the new one, then gradually deprecate the old one. It would take a lot of effort to migrate.

Unfortunately, it also doesn't fix the problem because people like to declare their methods like:
```C
PyObject *some_func(SomeTypeObject *self, ...)
```
so even if it has the right "shape", a cast is still needed.

I am thinking instead that we should patch the compiler to do an ast transform that fixes up these definitions. We can tell they are wrong at compile time by looking at the AST of the static module definition records.

hoodmane:
I have work towards automating these patches here:
https://github.com/hoodmane/clang-python-fpcast-plugin

hoodmane:
So as an update, the clang-plugin route _works_, the C++ code for the plugin is about 50% shorter than the Python code I am using here, and it involves less text parsing. However, building the clang plugin is problematic. In particular, emscripten ships with a version of clang that doesn't come with any build tools, so it's impossible to build against the emscripten clang. I can use the llvm installer script to install llvm, clang, and libclang-dev and then I can build against this version of clang. However, doing this increases the size of the docker image by about a gigabyte! 

I figured it wasn't worth taking that much of a space hit, so instead what I do is patch `emcc.py` to do the following:
1. For each source file, grep it to check if it contains the string `PyMethodDef` or `PyGetSetDef` anywhere.
2. If the source file contains these things, invoke the clang parser with `clang -cc1 -ast-dump` on the file and put the result into `source.c.ast`.
3. Read the text ast dump from `source.c.ast` and parse it into a tree with Python, then search this tree for problematic casts.
4. Patch the source file to fix the problematic casts, and write the result back over the source file.
5. Invoke original compiler call.

This seems like it won't matter that it is slower than the clang-plugin, since we don't have to look at all that much of the ast. I could improve it if it turns out to be slow (or even rewrite into some other language if for some reason it's a horrible problem) but on real input files it seems plenty fast enough.

One disadvantage with this approach is that it's a bit brittle to changes in the text representation of the clang ast, but on the other hand, at least it's not brittle to tiny changes in the clang ABI...

rth:
Thanks for working on it  @hoodmane !

> In particular, emscripten ships with a version of clang that doesn't come with any build tools, so it's impossible to build against the emscripten clang.

I agree that the second approach of modifying emcc would be much easier to maintain indeed.
Though IMO it would be better to put that code into `pyodide_build.function_casts` (or some similar module) and then import it in emcc if need (which would allow testing that code properly).

An even better solution would be to add this to pywasmcorss.py if possible. We are already using compiler aliases,
```bash
$ ls -l tools/
lrwxrwxrwx 1 rth 1000   47 Nov 26 15:53 cc -> /src/pyodide-build/pyodide_build/pywasmcross.py
[...]
```
so that cc calls pywasmcross.py which [calls emcc with some fixes](https://github.com/pyodide/pyodide/blob/a255d54f3f8be00c58c6f1e58e2befd4e8764357/pyodide-build/pyodide_build/pywasmcross.py#L53). Unless there is a reason why that wouldn't work with your current approach?

hoodmane:
> I agree that the second approach of modifying emcc would be much easier to maintain indeed.

It's a little bit of a lose lose situation because it's hard to read stuff like location information out of the code. Currently to get the locations I have:

<details>
<summary>Details</summary>

```python
location_info_re = re.compile(
    "<([a-zA-Z._/]*):([0-9]*)(:[0-9]*)?(, ([a-z]*):([0-9]*)(:[0-9]*)?)?>( ([a-z]*):([0-9]*)(:[0-9]*)?)?"
)
# ....
    if parent.children:
        last_sibling = parent.children[-1]
        last_loc = last_sibling.end_loc
    else:
        last_loc = parent.next_loc
    m = location_info_re.search(info)
    start_line = 0
    start_col = 0
    end_line = 0
    end_col = 0
    next_line = 0
    next_col = 0
    file = None
    if m:
        gps = m.groups()
        if gps[0] == "col":
            start_line = last_loc[0]
            start_col = int(gps[1])
        else:
            start_line = int(gps[1])
            start_col = int(gps[2][1:])
            if gps[0] != "line":
                file = gps[0]

        if gps[4] == "col":
            end_line = start_line
            end_col = int(gps[5])
        elif gps[4]:
            end_line = int(gps[5])
            end_col = int(gps[6][1:])
        else:
            end_line = start_line
            end_col = start_col

        if gps[8] == "col":
            next_line = start_line
            next_col = int(gps[9])
        elif gps[8]:
            next_line = int(gps[9])
            next_col = int(gps[10][1:])
        else:
            next_line = start_line
            next_col = start_col
    if not file:
        file = parent.file

    tree.start_loc = (start_line, start_col)
    tree.end_loc = (end_line, end_col)
    tree.next_loc = (next_line, next_col)
    tree.file = file
```
</details>
It's really a shame that [libclang](https://pypi.org/project/libclang/) isn't better but when I tried to use it, there seemed to be no way to access a lot of information that was necessary. At least the llvm people seem to be interested in improving this stuff.

> An even better solution would be to make this call in pywasmcross.py if possible.

Yes that sounds like a better solution. I am pretty sure that will work and it will be easier to develop (updating the patch is annoying). o make this call in pywasmcross.py if possible.

hoodmane:
Actually I think we do need to patch emcc for this -- we need all of the flags it produces to figure out which system libraries to have includes. Another problem is we don't use `pywasmcross` everywhere, we only use it to build the packages. It's probably better to also use the wrapper when building CPython core and Pyodide core too which this achieves.

hoodmane:
Just realized that `ast-dump=json` makes it produce a json ast... Luckily the json seems to come with better location info.

hoodmane:
@rth On circleci I keep getting:

```sh
make[1]: *** [Makefile:2020: Modules/unicodedata.o] Killed
```
each time in the same place, at around 11 minutes and 20 seconds +/- a bit. Sometimes it doesn't happen. Related to #2013?

hoodmane:
@rth Also, is there a way to manually flush all ccaches after merging this? I changed the compiler, so all caches need to be invalidated, but I think the cache invalidation mechanism only looks for changes in the source, not for changes in the compiler.

hoodmane:
Oh I think the problem might be that it uses up too much ram loading the AST for `unicodedata.o`. The AST json dump `unicodedata.c.ast` is a gigabyte! Maybe I should check whether the ast is super big and if so skip the patching step...

hoodmane:
Another option would be to use a compiled language to parse the ast or something. Sounds annoying though.

hoodmane:
`unicodedata.c` has 8000 lines of static data. The ast really blows that up by recording a bunch of location information for each 0.

rth:
> Also, is there a way to manually flush all ccaches when merging this?

Normally it hashes Makefile.env for the cache key, so you should be OK. Otherwise you can bump the date in https://github.com/pyodide/pyodide/blob/d2c6ecad1d2647ab2a6cf184fc8ec561382597d4/.circleci/config.yml#L47 and similar lines manually.

> The AST json dump unicodedata.c.ast is a gigabyte! Maybe I should check whether the ast is super big and if so skip the patching step...

Any way to process it with a generator without loading it all in memory?

hoodmane:
> Any way to process it with a generator without loading it all in memory?

Ironically processing the human-readable AST made this a lot easier. But the best approach is probably to use `jq`.

hoodmane:
https://github.com/stedolan/jq/wiki/FAQ#streaming-json-parser

hoodmane:
I think this is working now, but am having trouble with build timeouts.
But I just thought of a different approach to this which might be easier. Basically, we should just patch call trampolines into the Python interpreter in a few places rather than trying to fix the package source code.

hoodmane:
Closed in favor of #2019.

PR #1679, opened 2021-06-30
Simplify js2python string conversion code a bit

hoodmane:


rth (approved):
Assuming CI is OK, thanks!

hoodmane:
There are problems in Firefox in the CI. I do not yet understand why.

hoodmane:
Problem was just a sporadic failure.

hoodmane:
We might want to revert this, it seems to be inconsistently breaking up firefox-core tests on CI (cannot reproduce test failures locally though).

rth:
Yes, it does look like there are more timeouts after the last few commits. Thoughg it could be also due to CircleCI having larger loads and being slower as well.

PR #1680, opened 2021-06-30
Use new in 3.9 OneArg and NoArgs call methods instead of ObjArgs

hoodmane:
The `ObjArgs` methods have a `NULL` terminated argument list which can be a bit error prone. The new `OneArg` and `NoArgs` methods are a bit clearer and easier to use.


rth (approved):
Assuming the CI passes.

PR #1681, opened 2021-06-30
Better bootstrap

hoodmane:
The goal of this PR is to make the bootstrap process a bit cleaner which should make it easier to work on code in the `pyodide` Python package. 

Currently we have five Python packages `_pyodide`, `pyodide`, `_pyodide_core`, `js`, and `pyodide_js`. Of these, `_pyodide` and `pyodide` are Python modules, `_pyodide_core` is a C module, and `js` and `pyodide_js` are js modules. 

The goal is to make it so that the packages `_pyodide_core`, `js`, and `pyodide_js` are all available at import time for `pyodide`, so it is possible to make top level imports from these packages inside of `pyodide`. Not all functions from the `pyodide_js` package are ready to call at `pyodide_py` import time, but the only functions in `pyodide_js` that are not ready are ones that just call Python APIs and would never be needed from Python anyways. So for practical purposes both imports and runtime behavior from these packages are ready for `pyodide` to use at import time.

Of course, when running tests in a native Python none of these packages are really available. I added an `IN_BROWSER` flag defined in `pyodide/_core.py` that we can use to test before attempting an import. It's important to ask permission and not catch an `ImportError` because an `ImportError` may be generated in the browser because something else is screwed up.

I also want to move all code that doesn't depend on `_pyodide_core`, `js`, or `pyodide_js` into `_pyodide`.

The bootstrap process now works as follows. In `main.c`, `_pyodide` is imported very early on. Then `main.c` initializes the C code and creates and installs `_core_module`, this can all rely on code from `_pyodide`. At the end of `loadPyodide` after `main.c` has run to completion we use `runPythonSimple` to do the following steps:

1. Call `_pyodide._importhook.register_js_importer`. This requires imports from `_core_module`, but it's okay to do now because `main.c` set that up. This adds the `JsImporter` to `sys.metapath`.
2. Add `js` and `pyodide_js` to the list of js modules.
3. Import `pyodide` now that `js` and `pyodide_js` are ready.
4. Add the PyProxies for the Python `pyodide` package and the `__main__.__dict__` globals to `pyodide_js`.

At this point `runPython` can be used and the bootstrap process is essentially complete.


rth (approved):
Thanks for the analysis and +1 for anything that makes the bootstrap process simpler.


**Edit**: Restarted the CI to see if would help reduce the number of errors.

PR #1684, opened 2021-07-03
Setup a JavaScript test suite

rth:
This sets up a JS test suite using [Mocha](https://mochajs.org/). 

The goal is to be able to test individual functions in `src/js` without running the build system. 

The other goal is to start playing with Node.js in the future, and have tests for it. Though then we would need to build Pyodide core first.

As to the choice of the test package. It looks like the popular ones are [Mocha](https://mochajs.org/) and [Jest](https://jestjs.io/). I'm not familiar with neither. I went with the first one as it seems more lightweight and also more of a community project. But we can change that if anyone has strong opinions about it.

### DIFF
--- a/docs/development/testing.md
+++ b/docs/development/testing.md
@@ -15,7 +15,7 @@ Install [geckodriver](https://github.com/mozilla/geckodriver/releases) and
 [chromedriver](https://sites.google.com/a/chromium.org/chromedriver/downloads)
 and check that they are in your `PATH`.
 
-### Running the test suite
+### Running the Python test suite
 
 To run the pytest suite of tests, type on the command line:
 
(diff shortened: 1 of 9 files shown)


src/js/pyodide.js
```diff
@@ -1,16 +1,16 @@
 /**
  * The main bootstrap code for loading pyodide.
  */
-import { Module } from "./module";
+import { Module } from "./module.js";
```
rth, line 4: Otherwise I was getting errors like,
```
Error [ERR_MODULE_NOT_FOUND]: Cannot find module '/home/rth/src/pyodide/src/js/module' imported from
 /home/rth/src/pyodide/src/js/api.js                                                              
Did you mean to import ../module.js?
```

hoodmane (approved):

docs/development/testing.md
```diff
@@ -15,20 +15,28 @@ Install [geckodriver](https://github.com/mozilla/geckodriver/releases) and
...
 - `src/tests/`: general Pyodide tests and tests running the CPython test suite
 - `pyodide-build/pyodide_build/tests/`: tests related to Pyodide build system
   (do not require selenium to run)
 - `packages/*/test_*`: package specific tests.
 
+### Running the JavaScript test suite
+
+To run tests on the JavaScript Pyodide package using Mocha, run the following commands,
+```
+cd src/js
```
hoodmane, line 36: Perhaps could mention the type tests too here?

rth: Done.

PR #1685, opened 2021-07-03
Try to fix timeouts occurring starting with #1679

hoodmane:
Changes to string conversions in #1679 seem to cause extra test timeouts, particularly on firefox. Maybe this will be better?


hoodmane:
@rth I'm starting to think that the recent increase in timeouts might be unrelated to code changes. I guess I can make a branch that reverts the last few commits and see what happens with CI on that.

rth:
I think I was an issue with CircleCI, now the CI should pass without timeouts. Closing.

PR #1688, opened 2021-07-03
Try reverting some recent commits to see what CI does

hoodmane:
This reverts commit 9ec832be0191acba16f304beaf048cfb6ef386e4.

Trying to diagnose recent timeout errors on CI.


rth:
I also restarted CI on 2ab427e287da694ecca537600f75b94839bf5e74 before the last 3 commits, to see what happens. It was previously green.

hoodmane:
Okay I think the failures of these
```
test_string_conversion[firefox]
test_bigint_conversions[firefox]
test_hyp_py2js2py[firefox]
```
on the rerun of 2ab427e demonstrates that my recent changes to string and bigint conversion aren't actually implicated in the timeouts...

rth:
Yes, as much as I like CircleCI, I think it might be time to evaluate an alternative CI provider to see if we also get such issues there. I'll try to setup Azure pipelines which have 10 parallel builds and unlimited CI minutes (though I'm not very fond of their UI).

rth:
It might still be worth investigating Github Actions in #1690 in the future but for now CI appears to be green again.

rth:
Or at least I did see green CI in https://github.com/pyodide/pyodide/pull/1684 which included all of these commits.

PR #1689, opened 2021-07-04
Azure pipelines CI

rth:
Ref: https://github.com/pyodide/pyodide/pull/1688#issuecomment-873565477


rth:
OK, apparently now because of cryptocurrency mining, they explicitly validate open-source projects. Sent a request, indicated processing time is 2 to 3 days.

rth:
The Circle CI seems to be passing without timeout now, so I'm going to close this for now. Particularly that there have been more progress with Github Actions https://github.com/pyodide/pyodide/pull/1690. Still we could come back to this if ever we need more CI resources.

PR #1690, opened 2021-07-04
CI with Github Actions

rth:


rth:
Tests in Chrome generally work, in Firefox there still some issue with selenium setup that needs fixing.

rth:
To give a status update on this. Most of the CI jobs should now work. The behavior is fairly similar from what I can tell to CircleCI, there are still occasional failures (and more so when using pytest-xdist). It's maybe slightly more stable because GH Actions have runners with 7GB RAM instead of 4GB for CircleCI. We also get 20 parallels runs (and unlimited total runtime) which is great. 

The limitations are,
 - it's currently impossible to restart a single job (https://github.com/actions/runner/issues/432, [[2]](https://github.community/t/ability-to-rerun-just-a-single-job-in-a-workflow/17234)). Which because jobs still occasionally fail, I think, is a deal breaker for us.
 - more minor point, it's also impossible to run it on forks only on the upstream repo or on PRs.

So I would say until https://github.com/actions/runner/issues/432 is resolved we can't use it for everything. However maybe we could at least start using it for a few CI jobs such as linting and test-python, to reduce the load on CircleCi a bit. The disadvantage is that we then won't be able to depend on those jobs in CircleCI workflow but I would say it has never been an issue so far.

Another thing is that we got approved for Azure Pipelines (also 2CPU, 7GB RAM, 10 parallel jobs). Currently that would be more or less equivalent to CirleCI it's also likely we could get more parallel jobs if we ask for it. Though that would require re-writing this CI setup yet another time.

ehfd:
If building docker containers are also within the scope of CI, will it be possible to add the containers to GitHub Packages Container Registry (ghcr.io) in addition to Docker Hub? The rate limiting issues for Docker Hub is notorious, especially in multi-user environments.

rth:
@ehfd Could you please open a separate issue about that?

rth:
Following the comment in https://github.com/pyodide/pyodide/pull/1770#issuecomment-896709566 and https://github.com/pyodide/pyodide/issues/1850#issuecomment-924164863 my plan now is to adapt this PR to have a minimal build + maybe numpy & pandas outside of Docker for Linux and eventually Mac. It's a bit redundant with existing CI but we really need to make it easier to build Pyodide without docker as sometimes that's not an option.

Issue #14, opened 2018-02-28
Run in node.js


mdboom:
I don't really understand the implications of this or whether it's possible, but it might be handy for testing.

gabrielfreire:
Hi @mdboom , i made an attempt to load `pyodide.asm.wasm` on `nodejs` but no success so far, i installed a nice npm package for the boilerplate `npm install webassembly` 

the code looks like this
```javascript
const path = require("path");
const wa = require("webassembly");
const wasmPath = path.join(__dirname, '/wasm/pyodide.asm.wasm');
const memory = new WebAssembly.Memory({ initial: 8192, maximum: 8192 });
const table = new WebAssembly.Table({ initial: 16384, element: "anyfunc" });
const importObj = { memory: memory, table: table };
wa.load(wasmPath, { imports: importObj }).then((module) => {
    console.log(module.exports);
}).catch((err) => {
    console.log(`some error: ${err}`);
});
```
i'm getting this error
```
LinkError: Import #4 module="env" function="DYNAMICTOP_PTR" error: global import must be a number
```

can't find a solution for this, not sure what's causing it, maybe you have some idea?

mdboom:
I wouldn't expect this one to be easy, but it would be really cool if we can find a solution ;)

There are a bunch of things beyond just the Memory and Table that need to be set up to import the WebAssembly.  In general, I wouldn't expect to be able to run emscripten-generated code directly like this.  These variables are set up in `pyodide.asm.js`, and get passed to us in the `Module.instantiateWasm` callback in `pyodide.js`.

In general, I would look at the code in `pyodide.js` to see how all this gets bootstrapped and try to emulate that in the node environment.  

Even then, I think there are probably some browser-specific things that emscripten is generating here that may break things in node.  You may have to recompile pyodide with different settings.  I would look at the emscripten docs for info on porting wasm from browser to node in general.

gabrielfreire:
Would you have the `pyodide.asm.js` prettified version
i started adding stuff manually but i think it's too much stuff
```javascript
const importObj = {
    global: {
        NaN: NaN,
        Infinity: Infinity
    },
    env: {
        memory: memory, 
        table: table, 
        tableBase: 0, 
        memoryBase: 0,
        DYNAMICTOP_PTR: 2042576,
        tempDoublePtr: 2041040,
        STACKTOP: 2042592,
        STACK_MAX: 7285472,
        gb: 1024,
        fb: 0,
        abort: function(cb) {},
[...]

PR #1691, opened 2021-07-06
Minimal Node.js support

rth:
Partially addresses https://github.com/pyodide/pyodide/issues/14

This adds the few last pieces to run Pyodide in Node.js with local files. Also adds a test that checks that we can,
 - Run Pyodide in Node.js
 - Load Pyodide packages (currently from disk)
 - Load a package from PyPi with micropip using node-fetch. I'm not at all sure this is the best approach, I'll comment in the parent issue about alternatives. 

Next step would be to figure out how to add a Node.js target to our current test suite.

For adding documentation, I think it would be better to do this once we publish an alpha package on npm, as currently one would need to clone the repo and cd into `src/js` to use it.

### DIFF
--- a/src/js/test/pyodide.test.js
+++ b/src/js/test/pyodide.test.js
@@ -1,0 +1,30 @@
+import assert from "assert";
+import { loadPyodide } from "../pyodide.js";
+
+import fetch from "node-fetch";
+
+describe("Pyodide", () => {
+  let pyodide;
+  it("loadPyodide", async () => {
+    pyodide = await loadPyodide({ indexURL: "../../build/" });
+  });
+  it("runPython", async () => {
+    let res = pyodide.runPython("1+1");
+    assert.equal(res, 2);
+  });
+  it("loadPackage", async () => {
+    await pyodide.loadPackage(["micropip"]);
+  });
+  it("micropip.install", async () => {
+    // TODO: micropip currently requires a globally defined fetch function
+    global.fetch = fetch;
+    await pyodide.runPythonAsync(
+      'import micropip; await micropip.install("snowballstemmer")'
+    );
+    let res = pyodide.runPython(`
+    import snowballstemmer
+    len(snowballstemmer.stemmer('english').stemWords(['A', 'node', 'test']))
+    `);
+    assert.equal(res, 3);
+  });
+});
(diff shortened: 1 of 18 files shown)


src/js/test/pyodide.test.js
```diff
@@ -0,0 +1,30 @@
...
+  it("runPython", async () => {
+    let res = pyodide.runPython("1+1");
+    assert.equal(res, 2);
+  });
+  it("loadPackage", async () => {
+    await pyodide.loadPackage(["micropip"]);
+  });
+  it("micropip.install", async () => {
+    // TODO: micropip currently requires a globally defined fetch function
+    global.fetch = fetch;
```
hoodmane, line 20: I think we could best fix this as follows: define a js package `fetch_js` via 
```
let fetch_js_mod;
if(we are in node){
	fetch_js_mod = {
		fetch : await import("fetch")
	};
} else {
	fetch_js_mod = { fetch : globalThis.fetch };
}
pyodide.registerJsModule("fetch_js", fetch_js_mod);
```
then `from fetch_js import fetch` in `micropip` rather than from `js`.

rth: Yes, this is clearly for test/demo purposes only. I commented in https://github.com/pyodide/pyodide/issues/14#issuecomment-875116949 to discuss what should be the ideal long term solution. Ideally I would still love to keep this npm package as independent as possible from the rest of the Pyodide code that requires the build setup. Though last time we talked about it was difficult as far as I remeber.

rth:
Any objections on merging this @hoodmane ? I'd rather work on better caching system for package downloads in Node in a follow up PR (and it will likely not use node-fetch).

hoodmane (approved):

hoodmane:
No objections.

karthiknadig:
@rth I had made similar changes to pyodide Js files to get Jedi working locally for some experiment that I was doing. I had replaced the fetch with node JS File system API. and got Jedi working. but I had to make several changes to pyodide.asm.js to get everything running. I am not really familiar with how pyodide is built so I just modified the JS files until it started working.

These are the sub-set of changes that I had to make to get it to work in node. I am not sure if this is of any use to pyodide team, but just wanted to let you know. 
https://github.com/karthiknadig/node-pyodide/compare/node-support

PR #1692, opened 2021-07-07
Expose Module.FS

bollwyvl:
- [x] reduces the scope of (and maybe blocks?) https://github.com/pyodide/pyodide/pull/1596#issuecomment-846470213
- [x] hoist `Module.FS` to `fileSystem` on the public API
- [x] add `@types/emscripten`
- [x] centralize typescript options in `tsconfig.json`
- [x] add a test of `fileSystem.mkdir` 

I didn't add any examples or tests, but the [docs](https://pyodide--1692.org.readthedocs.build/en/1692/usage/api/js-api.html?highlight=filesystem#pyodide.fileSystem) seemed to update properly.

Screenshot from 2021-07-06 21-35-46


bollwyvl:
Given I'm having to duplicate the `--lib` in two places, how do we feel about biting the bullet and just having a `src/js/tsconfig.json`? There are probably some other static analysis advantages to be gained with more compiler flags that would start to get pretty lengthy at the CLI (though nothing on some of the these `emcc` calls :laughing: )

hoodmane:
Fine with me, I just set up the flags in the way that was fastest for me at the time.

bollwyvl:
Welp, i wouldn't consider the `tsconfig.json` a blocker for at least review... otherwise are any of the fails indicators of things i broke? seem like mostly timesout or stuff that flakes on the #1596, too.

hoodmane (approved):

hoodmane:
Yeah you got some slightly unusual failures there but the failures couldn't possibly be related to this change set. It's just normal CI flakiness.

hoodmane:
We need a changelog entry for this and it would be good if you could add a simple test. Plus there are conflicts with `package.json` now.

bollwyvl:
I've done the `tsconfig.json` change, and added a test... but it appears to fail the same way as the other tests on `main`.

hoodmane:
If you update the changelog I will merge it.

bollwyvl:
> update the changelog

Gah, literally the first thing you wrote! Incoming.

hoodmane:
Thanks! A lot of people will need to use this.

bollwyvl:
You bet! We're looking forward to trying it out.

I'm away from keys, but will follow up with the IDBFS one... and add all
the backend flags, which should complement the new node stuff.

PR #1695, opened 2021-07-09
Fix run_docker to work on Mac (#1695) [skip ci]

ondrej-stanek-ozobot:
The long options to `id` commands are not supported on Mac,
use compatible short variants instead.

### DIFF
--- a/run_docker
+++ b/run_docker
@@ -13,11 +13,11 @@ USER_HOME="/src/.docker_home"
 USER_NAME="$(id -u -n)"
 USER_PASS="x"
 USER_ID="$(id -u)"
-USER_GID=0
+USER_GID="$(id -g)"
 USER_COMMENT_FIELD="${USER_NAME} pyodide user alias"
 USER_INTERPRETER="/sbin/nologin"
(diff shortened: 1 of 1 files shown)


run_docker
```diff
@@ -17,7 +17,7 @@ USER_GID=0
 USER_COMMENT_FIELD="${USER_NAME} pyodide user alias"
 USER_INTERPRETER="/sbin/nologin"
 USER_ACCOUNT_INFO="${USER_NAME}:${USER_PASS}:${USER_ID}:${USER_GID}:${USER_COMMENT_FIELD}:${USER_HOME}:${USER_INTERPRETER}"
-USER_FLAG="--user $(id --user):$(id --group)"
+USER_FLAG="--user $USER_ID:$(id -g)"
```
ondrej-stanek-ozobot, line 20: @hoodmane , please advise if there should be `$(id -g)` (backward compatible) or `$USER_GID` (evaluates to `0`). I am unsure about this, so I kept the variant that doesn't break anything.

hoodmane: I think it should say `$USER_GID`. But is there a good reason that we set `USER_GID=0` rather than `USER_GIT=$(id -g)`?

ondrej-stanek-ozobot: Yes, `USER_GID=$(id -g)` makes sense to me. I am unsure why there was 0 in the first place.

hoodmane (approved):

PR #1696, opened 2021-07-09
PyProxy attribute improvements

hoodmane:
Two small improvements to PyProxy:

1. We added the `$` prefix system to get properties of a Python object shadowed by Javascript. I updated the rules to also handle a `$` prefix in the setter and deleter for consistency.
2. Allow setting and deleting symbols on the PyProxy. This allows the use of `PyProxy` with `Comlink.proxy`.


rth (approved):
Does this need a changelog entry? Otherwise looks good. Thanks!

PR #1698, opened 2021-07-09
Add stack usage benchmark

### LINKED ISSUE #1541: CPython stack frame size estimation
The JavaScript stack capacity is limited in web browsers. In turn, this imposes a limit on the maximum CPython stack size and maximum recursion depth in a python program.

Some python applications hit the stack limit. One example of such python module is `jedi` (an autocompletion, static analysis and refactoring library).

Although we cannot do anything about the JavaScript stack limit in web browsers, we noticed that the stack frame size of CPython can be optimized. By reducing the stack frame size, it is possible to achieve higher recursion depth.

We observe that the stack frame is approx 18% larger in pyodide version 0.17.0 (as of rev. c327405), compared to pyodide versions 0.14.1 and 0.15.0.
However, we see an improvement in this aspect when comparing version 0.17.0 to version 0.16.1.

We use this script to approximate the CPython stack frame size:
```javascript
loadPyodide()
measure_available_js_stack_depth = () => {
    let depth = 0;
    function recurse() { depth += 1; recurse(); }
    try { recurse(); } catch (err) { }
    return depth;
}; pyodide.runPython("import js\ndef recurse(n):\n    return js.measure_available_js_stack_depth() if n==0 else recurse(n-1)\n(recurse(0)-recurse(100))/100")
```

The code snippet above is to be pasted to the JavaScript console in the `build/test.html` page. The script outputs a single number, which is the relative stack frame size of CPython. The smaller the better.

The pyodide team [has shown an interest](https://github.com/pyodide/pyodide/issues/1331#issuecomment-823483383) to integrate stack frame size estimation into CI to hunt down regressions.

This issue was created to facilitate any discussion on that topic.

hoodmane:
Resolves #1541

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -186,6 +186,16 @@ jobs:
       - store_test_results:
           path: test-results
 
+  benchmark-stack-size:
+    <<: *defaults
+    steps:
+      - attach_workspace:
+          at: .
+      - run:
+          name: stack-size
+          command: |
+            pytest -s benchmark/stack_usage.py  | sed -n 's/## //pg'
+
   test-emsdk:
     <<: *defaults
     resource_class: small
@@ -329,6 +339,9 @@ workflows:
           requires:
             - build-core
       - test-js:
+          requires:
+            - build-core
+      - benchmark-stack-size:
           requires:
             - build-core
       - benchmark:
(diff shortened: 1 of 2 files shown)


hoodmane:
On firefox the `py_usage` metric does not really work, so two of the columns just give garbage on firefox. Luckily, the metrics we care about are not impacted.

PR #1699, opened 2021-07-09
Remove fpcast calls from CALL_FUNCTION opcode

hoodmane:
In order to try to reduce stack usage.

If we do a basic infinite recurse of Python functions we see the following:
```js
    at pyodide.asm.wasm:0x1f08e1
    at _PyFunction_Vectorcall (pyodide.asm.wasm:0x1f06b2)
    at byn$fpcast-emu$_PyFunction_Vectorcall (pyodide.asm.wasm:0x7eec18)
    at pyodide.asm.wasm:0x2d9f0d
    at _PyEval_EvalFrameDefault (pyodide.asm.wasm:0x2d7383)
    at byn$fpcast-emu$_PyEval_EvalFrameDefault
```
This should cut out the bottom two frames I think. In particular, the `fpcast-emu` calls take up a very large amount of stack space so if we can remove them that will help. 

Another idea would be to make `fpcast-emu` functions take a struct. This would move their call stack space onto the argument stack. I don't know what the speed implications of that would be.


hoodmane:
Based on the benchmark-stack-size result, this is super effective.

On main:
```
browser  js_depth  py_depth   py_usage
firefox     39052      1056         36
chrome      13951       628         22
```
On this branch:
```
browser  js_depth  py_depth  py_usage
firefox     39052      1594        24
chrome      13951       923        15
```
Stack usage per call frame is reduced by a factor of 2/3 and the maximum recursion depth is increased by a factor of 3/2. We would probably see a similar reduction in stack usage if we could remove the other pointer cast.

hoodmane:
Seems to have improved the speed a bit too, ratio of benchmarks after/before:
```
                               chrom fire	
selenium init                  1.148 0.850
load numpy                     1.281 0.584
pystone                        0.915 0.895
allpairs_distances             1.009 0.866
allpairs_distances_loops       0.924 0.910
arc_distance                   0.899 0.914
check_mask                     0.901 0.899
create_grid                    0.870 0.860
cronbach                       0.881 0.863
diffusion                      0.911 0.926
evolve                         0.886 0.887
fdtd                           0.905 0.900
grayscott                      0.895 0.905
grouping                       0.940 0.933
growcut                        0.925 0.913
harris                         0.899 0.918
hasting                        0.959 1.687
julia                          0.913 0.911
l2norm                         1.245 1.011
large_decimal_list             0.917 0.872
local_maxima                   0.928 0.908
log_likelihood                 0.970 0.948
lstsqr                         1.004 1.108
mandel                         0.945 1.011
multiple_sum                   0.888 0.922
pairwise_loop                  0.891 0.921
periodic_dist                  0.908 0.925
repeating                      0.980 1.018
reverse_cumsum                 0.964 0.948
rosen                          1.024 1.005
slowparts                      0.927 0.965
specialconvolve                1.001 0.946
vibr_energy                    0.912 0.904
wave                           1.260 0.917
```
The geometric mean of these is 0.944, so this might take as much as 5% off the run time.

hoodmane:
With the second fpcast call removed:
```
browser  js_depth  py_depth  py_usage
firefox     39052      1996        19
chrome      13951      1708         8
```
Seems that this fixes our recursion limit woes.

hoodmane:
Well I think this shows pretty convincingly that #1677 will be an enormous improvement when it is finished.

hoodmane:
The benchmarks show that the second change slowed things down a bit (not that surprising I suppose). Overall there is a 1% slowdown and a 2.75x reduction in stack space used for a normal function call.

rth (approved):
Very nice! There is still some number of CI failure, restarted CI to see  if it's just flakiness or if there some underling issues.

> Well I think this shows pretty convincingly that #1677 will be an enormous improvement when it is finished.

Yeah, performance wise it should be huge improvement.

Issue #977, opened 2020-12-30
Make a build/index.html


hoodmane:
I think index.html should start with links to the console, the docs, and github. Then the list of packages would be good. If possible it would be nice to insert links to the package docs / pypi.

LinasNas:
Hi, 

I recently asked a similar question on issue #1672 and before doing any automation, I just wanted to see if this is close to what you have imagined:

1. [Here])(https://gist.github.com/LinasNas/0d896259a809683da948eeb3f28e7c86#file-get_package_descriptions-py) is a Python script that can fetch the description and useful URLs for each package in the Pyodide [package folder](https://github.com/pyodide/pyodide/tree/main/packages). 

2. [Here](https://gist.github.com/LinasNas/3b3a793730efa816a059a1529bbe8ad1#file-write_to_meta-py) is a script that would update the ```meta.yaml``` file for each package, by adding a short summary about the package, and useful URLs like links to official documentation, PyPi, or Github. 

3. [This](https://gist.github.com/LinasNas/441bd8472179ae5c7692ae49cf5b0e05) is an example of how the ```meta.yaml``` file in the [pandas folder](https://github.com/pyodide/pyodide/tree/main/packages/pandas) would look like, if the first two steps were executed 

A few comments:
- I fetched the package information for Step 1 from PyPi, and because several packages were not available there, their descriptions would be left empty.
- In addition to a short summary about each package, it's also possible to fetch a much longer description. I didn't include that part, because the ```meta.yaml``` files become quite long and less comprehensible. [Here's](https://gist.github.com/LinasNas/05918af6d60a5ed910cf868bbf5dd2da#file-pandas_long_meta-yaml) an example of how the ```meta.yaml``` file for pandas would look like with the long description.

Would love to hear your feedback about the current stage of work, and also the next steps regarding the automation of these updates (e.g. by using Github actions.)

rth:
Thanks for looking into this @LinasNas !

To add more context to my comment in https://github.com/pyodide/pyodide/issues/1672#issuecomment-870950384 we really have two choices,
 a) fetch this information at build time.
 b) fetch it during the creation of meta.yaml then permanently store the output in git.

a) would result in lighter `meta.yaml` but at the cost of more possibilities for build times failure if something changes in unexpected way on PyPi for instance. That's why, I think, b) would be preferable.

> Here is a Python script that can fetch the description and useful URLs for each package in the Pyodide package folder.

[...]

PR #1700, opened 2021-07-09
Toward a conda compatible packages.json

rth:
Partially addresses https://github.com/pyodide/pyodide/issues/791

This changes the packages.json have the following format,
```
{
  "info": { "arch": "wasm32", "platform": "Emscripten-1.0" },
  "packages": {
    "test": { "name": "test", "version": "1.0", "depends": [] },
    "micropip": {
      "name": "micropip",  // the name uses the original case
      "version": "0.1",
      "depends": ["pyparsing", "packaging", "distutils"],
      "imports": ["micropip"]
    },
    ...
}
```
instead of currently,
```
{
  "dependencies": {
    "micropip": ["pyparsing", "packaging", "distutils"],
    ...
  },
  "import_name_to_package_name": {
    "micropip": "micropip",
     ...
  },
  "shared_library": {},
  "versions": {
    "micropip": "0.1",
     ...
  },
  "orig_case": {
    "micropip": "micropip",
    ...
  }
}
```

which has the advantage of being more straightforward to use, particularly if we add
 - other per package fields (sha256 etc).
 - other general fields (e.g. pyodide version or build date)
 
Also it has better compatibility with conda's repodata.json. 

**TODO**
 - [ ] fix a few remaining tests


bollwyvl:
Cool prgoress! The devil's always in the metadata! I'm a big proponent of both:
- moving closer to `conda` compatibility
- JSON schema for package-manager stuff

Unfortunately, I know less about `repodata.json` than any of the other conda artifacts, so can't be of much help... but could certainly take a crack at the JSON schema piece if that was of interest.

rth:
It would be nice to have a better validation of the JSON Schema. Currently it is mostly used on the examples I found (such as this [one](https://repo.anaconda.com/pkgs/pro/osx-64/repodata.json)) with a few adapted fields
 - the "packages" keys are package names not file names (since those make currently less sense for Pyodide)
 - extra "imports" field that we require.

If you see any ways to improve it @bollwyvl feedback would be very welcome.

bollwyvl:
> package names not file names 

I _could_ see this becoming relevant if one package ends up needing to be built for different runtime environments, e.g. `nodejs` but shipped in the same `packages.json`.

>  better validation of the JSON Schema

Well, there is at least one [semi-official schema](https://github.com/conda/schemas/blob/master/repodata-1.schema.json) I know of, as well as [a narrative spec](https://github.com/mamba-org/conda-specs/blob/master/specs/repodata.md) which is used to implement mamba's handling. No doubt they disagree on some points.

I think whatever it is, having a formal schema around as `.json` and checking the built asset against it is a swell idea, even if it isn't 1-for-1 compatible with the upstream... though micropip would not need access to it (or `jsonschema`, though that would be nice for other things :blush: ) especially if its types were _generated_ from the schema, a la [quicktype](https://github.com/quicktype/quicktype).

Also as `pygments` vs `Pygments` has been a bit of a bummer at times, i'd explicitly add that to the test cases.

rth:
> Well, there is at least one semi-official schema I know of, as well as a narrative spec which is used to implement mamba's handling. 

Thanks for the references!

> I think whatever it is, having a formal schema around as .json and checking the built asset against it is a swell idea, even if it isn't 1-for-1 compatible with the upstream... though micropip would not need access to it (or jsonschema, though that would be nice for other things blush ) especially if its types were generated from the schema, a la quicktype.

+1 for that. Maybe in a follow up PR to this one. Here I explicitly left alone a few fields such as checksums (since we need to decide how to deal with them when there are two files per package). We can probably add those incrementally. Also related to https://github.com/pyodide/pyodide/issues/977#issuecomment-877471582

bollwyvl:
> checksums

The only potentially useful thing i can think of for this would be to provide [SSRI-compatible hashes](https://developer.mozilla.org/en-US/docs/Web/Security/Subresource_Integrity)... only semi-useful as i don't know if it works directly with `fetch`.

> two files per package

Hm, these probably _could_ be unified into a single js with e.g. rollup and transformer tricks. Half the requests (to flake) is never bad, and would remove the requirement for a `.wasm`-ready host... but it looks like [existing techniques](https://guido.io/posts/embedding-webassembly-in-javascript/) unsurprisingly use `base64` encoding which would increase the request size significantly by _33%_, and require the browser to parse the titanic strings, which is kind of heavy...

rth:
>  SSRI-compatible hashes... only semi-useful as i don't know if it works directly with fetch.

It should work apparently https://javascript.info/fetch-api#integrity  


>  these probably could be unified into a single js [..] require the browser to parse the titanic strings,

Yeah, I'm not sure that would be a good idea to send MB of data via in a JS file. The opposite could also be done. Defining a binary file with standardized fields and having the logic to manipulate it shared in pyodide.js, but that would likely require lots of work (including rewriting `file_packager.py` ) and I'm not convinced there is enough reasons to do it. Aggregating many small JS files into a single bundle certainly makes sense. I'm less sure about files in the 1MB-10MB range. Leaving them distinct for parallel downloads might actually be better. And while the browser is downloading those large files, a few small extra .js might not matter.

More pragmatically we could keep checksums of JS file in `packages.json`, then include the checksum of the `.data` inside the JS file and let it check it when loading.

bollwyvl:
> It should work apparently https://javascript.info/fetch-api#integrity

That's great, as sleazy ISPs (or worse) like injecting junk, apparently.

> inside the JS file and let it check it when loading.

So long as it can rely on standard browser crypto facilities, and not have to bring in a JS one :rofl: that sounds pretty good, aside from the entropy now increasing even further (hash sums in things that also get hash-summed).

Alternately, and at the price of _another_, tiny request-and-parse-per-package, the `packages.json` could point at a dumb-as-rocks `package.json` (gah, can't call it that... maybe `buildinfo.json`?) file that pointed at those _n_ files (again with `sha384` sums)... and would be a place to put the `exports` field, which seems a little odd to include in top-level package metadata (haven't looked at where it's used), but maybe that's my conda-is-not-just-for-python side showing. 

By default, a conda package also includes what was used to build it, etc. in `recipe/`... as I don't imagine this conda-_like_ system is self-hosting, there's likely little expectation that given the built package, one would be able to rebuild the package... but at least knowing _which_ patched compiler, etc. it was built with, etc. would be a step towards long-term reproducibility and auditability. Of course, `recipe/` (as built) is _also_ gets hashed, which is just one of it's of its reproducibility (big R) shortcomings.

bollwyvl:
Semi-related, this was recently announced:

https://packaging-con.org

rth:
@hoodmane I restarted CI to get the few last failures but it should be OK I think. Last build was [this one](https://app.circleci.com/pipelines/github/rth/pyodide/163/workflows/30f017bf-5569-46a1-b764-dcd2fba0a6c4) and the failure does not look related to me. I'll push a commit fixing linting before merging with no CI to avoid rebuilding everything because of black.

pyodide-build/pyodide_build/buildall.py
```diff
@@ -263,39 +263,54 @@ def builder(n):
...
         ]
-        package_data["versions"][name.lower()] = pkg.version
-        for imp in pkg.meta.get("test", {}).get("imports", [name]):
-            package_data["import_name_to_package_name"][imp] = name.lower()
-        package_data["orig_case"][name.lower()] = name
+        pkg_entry["imports"] = pkg.meta.get("test", {}).get("imports", [name])
+
+        package_data["packages"][name.lower()] = pkg_entry
 
     # Hack for 0.17.0 release
```
hoodmane, line 299: We could maybe update this comment.

pyodide-build/pyodide_build/buildall.py
```diff
@@ -263,39 +263,54 @@ def builder(n):
...
+
+    return package_data
+
+
+def build_packages(packages_dir: Path, outputdir: Path, args) -> None:
+    pkg_map = generate_dependency_graph(packages_dir, args.only)
+
+    build_from_graph(pkg_map, outputdir, args)
+
+    package_data = generate_packages_json(pkg_map)
```
hoodmane, line 312: Factoring this out did improve the readability a bit.

src/tests/test_asyncio.py
```diff
@@ -412,9 +412,9 @@ def test_await_pyproxy_eval_async(selenium):
             from js import fetch
             await (await fetch('packages.json')).json()
         `);
-        let packages = await c;
+        let result = await c;
```
hoodmane, line 415: We should move to fetching some other test JSON data by default in tests. I don't know how often this will happen but I think it's not ideal to have a bunch of tests asserting stuff about the structure of `packages.json`.

src/js/load-pyodide.js
```diff
@@ -177,7 +182,7 @@ async function _loadPackage(names, messageCallback, errorCallback) {
         continue;
       }
     }
-    let pkgname = Module.packages.orig_case[pkg] || pkg;
+    let pkgname = Module.packages[pkg]?.name || pkg;
```
hoodmane, line 185: This `?.` works correctly? I've had it cause trouble for me in the past. Maybe that was just an issue with the emscripten code transformation tools that don't understand newer syntaxes.

src/js/load-pyodide.js
```diff
@@ -122,8 +126,9 @@ function recursiveDependencies(
   if (sharedLibsOnly) {
     let onlySharedLibs = new Map();
     for (let c of toLoad) {
-      if (c[0] in sharedLibraries) {
-        onlySharedLibs.set(c[0], toLoad.get(c[0]));
+      name = c[0];
+      if (!!Module.packages[name].shared_library) {
```
hoodmane, line 130: `if(!!x)` is the same as `if(x)`.
```suggestion
      if (Module.packages[name].shared_library) {
```

src/js/api.js
```diff
@@ -137,19 +137,25 @@ export async function loadPackagesFromImports(
   if (imports.length === 0) {
     return;
   }
-  let packageNames = Module.packages.import_name_to_package_name;
+  if (typeof Module._import_name_to_package_name === "undefined") {
```
hoodmane, line 140: I think it would be better to do this in `initializePackageIndex`.

src/js/load-pyodide.js
```diff
@@ -14,11 +14,17 @@ export async function initializePackageIndex(indexURL) {
     fs.readFile(`${indexURL}packages.json`, (err, data) => {
       if (err) throw err;
       let response = JSON.parse(data);
-      Module.packages = response;
+      Module.packages = response["packages"];
     });
   } else {
     let response = await fetch(`${indexURL}packages.json`);
-    Module.packages = await response.json();
+    Module.packages = (await response.json())?.packages;
```
hoodmane, line 21: The logic here seems a bit messy to me (though it isn't really the fault of this PR). I think the following would be better:
```
let package_json;
if(in_node){
    const fsPromises = await import("fs/promises");
	const package_string = await fsPromises.readFile(`${indexURL}packages.json`);
	package_json = JSON.parse(package_string);
} else {
    let response = await fetch(`${indexURL}packages.json`);
    package_json = await response.json();
}
if (!package_json.packages) {
  throw new Error(
    "Loaded packages.json does not contain the expected key 'packages'."
  );
}
Module.packages = package_json.packages;
Module._import_name_to_package_name = new Map();
for (let name of Object.keys(Module.packages)) {
  for (let import_name of Module.packages[name].imports) {
    Module._import_name_to_package_name.set(import_name, name);
  }
}
```

hoodmane: Also, ideally we could define `IN_NODE` somewhere so that we could say `if(IN_NODE)` which would be more readable.

hoodmane (approved):

src/js/load-pyodide.js
rth: Yes, you are right it's maybe a bit too recent. https://caniuse.com/?search=%3F. So probably better to avoid it.

rth:
Thanks for the review! 

> Semi-related, this was recently announced: packaging-con.org

Interesting thanks! Would be good to participate. We have lots of open topics for packaging :)

PR #1704, opened 2021-07-11
Ran make -C packages update-all

hoodmane:


packages/pytest/meta.yaml
```diff
@@ -1,24 +1,20 @@
 package:
   name: pytest
-  version: 3.6.3
-
+  version: 6.2.4
```
rth, line 3: I'm not sure it would work without many more patches. It might be better to try updating pytest in a separate PR.

hoodmane: Maybe it would be a good idea to mark certain files not to auto update? Or to deal separately with files with patches? Also it would be nice if it could auto update C packages that load the source from PyPi... Hmm.

PR #1706, opened 2021-07-11
Make build scripts share environment and update lxml package

hoodmane:
The newer versions of lxml need to be told the location of xml2 and xslt via environment variables. It is a bit of a nuisance that the environment from the `source` script is not shared by the `compile` and `post` steps, so I made a new class `SubprocessWithSharedEnvironment` that runs bash code with shared env. It only runs bash code because there isn't a general way of getting the final values of environment variables from a process.

With this, environment variables defined in `script` are available in the later steps, so we can define `WITH_XML2_CONFIG` and `WITH_XSLT_CONFIG` for the compile step of lxml.


pyodide-build/pyodide_build/buildpkg.py
```diff
@@ -6,21 +6,62 @@
...
+class BashRunnerWithSharedEnvironment:
+    """Run multiple bash scripts with persisent environment.
+
+    Environment is stored to "env" member between runs. This can be updated
+    directly to adjust the environment, or read to get variables.
+    """
+
+    def __init__(self, env=None):
+        if env is None:
+            env = dict(os.environ)
```
rth, line 34: I'm not sure about this default. I think it might be better to explicitly pass the env variables needed. We have so many env variables set in various makefiles that it's a bit hard to follow. Normally none of those variables should be needed to build Python packages anyway I think?

rth (approved):
This is indeed very useful! Could you add a changelog entry please? 

Looks aside for the comment below (and assuming there are no major issues in CI).

pyodide-build/pyodide_build/buildpkg.py
hoodmane: This keeps the logic in the file the same though. It was passing all of `os.environ` to all three stages (`build.script`, `compile` and `build.post`). I can move it so that it is passed as an argument?

hoodmane: I figure because the whole environment was exposed before, if you want to change that it would be most appropriate to do so in a separate PR.

rth: OK, I though before we weren't passing the full env. Then never mind.

hoodmane:
There's a CI problem that needs to be fixed, `build-packages` keeps timing out.

PR #1708, opened 2021-07-12
Patches from joemarshall/fpcast

hoodmane:
The fpcast branch #1677 crashes chrome in CI but not when testing locally. These patches work fine by themselves. Maybe reducing the diff might help with debugging that.


rth (approved):
The CI looks OK now. I think we can include it, but we would first need to add a bit more documentation of these patches.

hoodmane:
I think if we just put `remove-fpcasts` or something in the name of all the patches, they are otherwise pretty self explanatory. I guess it could be helpful to track down why the fixed signatures are the correct ones.

rth:
I added a few more comments to patches, otherwise good to merge.

@joemarshall did a great job and fixed most of these upstream, so there isn't much left to do for us here.

PR #1709, opened 2021-07-12
Fix node tests

hoodmane:
I think this will make the node tests pass.

It would be nice if we could work towards running the rest of our test suite in node side by side with our selenium tests. I don't think this would be too hard -- we need to redo the selenium type conversions but as far as I can tell those basically only cover things that can be serialized to/from JSON.


rth:
So the issue is that my PR for Node was was missing a few latest commits from master, I guess in particular the one about boostrapping? The CI still doesn't seem to be passing. 

> It would be nice if we could work towards running the rest of our test suite in node side by side with our selenium tests.

+1

hoodmane:
> The CI still doesn't seem to be passing.

It seems like we just need to increase the timeout now.

> So the issue is that my PR for Node was missing ... bootstrapping PR?

Yeah I guess that was the main issue, though I also ran into the issue that the different tests are run in the same context and they tried to load Pyodide multiple times, so I extracted loading Pyodide into a "fixture".

rth:
The issue was that `mocha.opts` was supported only for mocha <=7, put the config into `package.json` directly now.

rth:
BTW, @hoodmane I enabled the auto-cancel of CI jobs in CircleCI on your fork (same as on upstream/main) so that when a new commit is pushed earlier CI runs in that PR are cancelled.

PR #1711, opened 2021-07-12
Remove CMake -S and -B options

ryanking13:
`-S` and `-B` options in CMake are [not (officially) supported](https://cgold.readthedocs.io/en/latest/glossary/-H.html) in Cmake < 3.13.

The latest version of Cmake in [Ubuntu 18.04 package archive is 3.10](https://packages.ubuntu.com/bionic/cmake), so I think it is good to avoid using those options.


rth (approved):
Thanks @ryanking13 ! Looks good assuming build CI passes.

PR #1716, opened 2021-07-16
remove unnecessary 'else' after return

slowy07:
removed unnecessary else after return
example 
```python
def test(x, y, z):
    if x:
        return y
    else:  # This is unnecessary here.
        return z
```
to
```python
def test(x, y, z):
    if x:
        return y
    return z
```


rth (approved):
Thanks @slowy07 !

slowy07:
> Thanks @slowy07 !

thanks @rth !!!

PR #1718, opened 2021-07-16
CI Disable pytest-xdist

rth:
An attempt to improve CI stability by avoiding to run tests in parallel. 

I generally the issues we have have to do with CPU oversubscription as mentioned in https://github.com/pyodide/pyodide/issues/1389 . Even when only one test is run at a time, multiple threads are created by the browser. Also for each test failure we have a 30s penalty until the timeout it reached. So running tests serially might not be that much slower, and should, hopefully, be more reliable. 

Alternative to #1690

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -153,7 +153,7 @@ jobs:
             mkdir test-results
             tools/pytest_wrapper.py \
               --junitxml=test-results/junit.xml \
-              --verbose --numprocesses 3 \
+              --verbose \
               << parameters.test-params >>
       - store_test_results:
           path: test-results


hoodmane:
So far so good, the core tests actually succeeded which has been pretty rare. I really hope this fixes the problems, the CI issues are a huge pain.

rth:
yes, I think I would still prefer slow CI to unreliable CI. And I have a feeling things got worse since https://github.com/pyodide/pyodide/commit/49321abcf2279e8d6bf2cdb5d5479aae9a0b926e which increased the n_jobs to 3 everywhere while before it was at 2 or 1 depending on the job. At the very least we should probably revert to passing this value via `test-params` parametrization.

hoodmane:
Slow is definitely much better than unreliable.

hoodmane:
Okay I vote we merge this.

rth:
Well it looks better, but hard to be sure without some stats. The general issue with this is that, I think, it depends on the CircleCI load or resource allocations, on some days our setup works on others it would time out.

I'll try to get to the bottom of https://github.com/pyodide/pyodide/pull/1690 as well. Last I tried core tests seemed more stable with Github Actions, even with multiprocessing set to 2.

hoodmane:
These have been failing so consistently lately that it would be surprising if it worked this well by chance.

rth:
Good to finally see a green CI.

Apparently there whole papers on CI flakiness with Selenium https://par.nsf.gov/servlets/purl/10100316 :)

> [..] while Chrome works best on faster systems or where configuration
can be optimized. We demonstrated that hardware has a
significant impact upon the runtime and reliability of a test
suite and that a faster CI environment makes a meaningful
difference.

Which is more or less our experience as well.

PR #1721, opened 2021-07-18
Change converter depth arguments to be named arguments

hoodmane:
I think this is clearer (especially because I don't expect the `depth` argument to be used all that much) and I want to add other optional named arguments.


rth (approved):
+1 for this.

rth:
`ci/circleci: deploy-dev` triggered because you created it from your main branch probably by accident.

hoodmane:
> probably by accident

Definitely by accident.

PR #1726, opened 2021-07-20
 Add pyproxies named argument to toJs

hoodmane:
This adds an argument called `pyproxies` to `toJs`. If passed a Javascript list, we collect all the `PyProxies` created by `toJs` into the list. This makes it easy to destroy them all when done (without needing to recursively walk the structure). 

I also added an optional argument `create_pyproxies` which is a boolean that defaults to true. If set to `False` we raise a `ConversionError` rather than creating a PyProxy. 

I also added these arguments to `to_js`.

### DIFF
--- a/src/core/pyproxy.js
+++ b/src/core/pyproxy.js
@@ -373,7 +373,6 @@ class PyProxyClass {
       proxies_id = Module.hiwire.new_value(pyproxies);
     } else {
       proxies_id = Module.hiwire.new_value([]);
-      s;
     }
     try {
       idresult = Module._python2js_with_depth(ptrobj, depth, proxies_id);
--- a/src/core/pyproxy.js
+++ b/src/core/pyproxy.js
@@ -348,26 +348,38 @@ class PyProxyClass {
   }
   /**
    * Converts the ``PyProxy`` into a Javascript object as best as possible. By
-   * default does a deep conversion, if a shallow conversion is desired, you
-   * can use ``proxy.toJs({depth : 1})``.
-   * See :ref:`Explicit Conversion of PyProxy
+   * default does a deep conversion, if a shallow conversion is desired, you can
+   * use ``proxy.toJs({depth : 1})``. See :ref:`Explicit Conversion of PyProxy
    * <type-translations-pyproxy-to-js>` for more info.
    *
    * @param {object} options
-   * @param {number} options.depth How many layers deep to perform the conversion.
-   * Defaults to infinite.
+   * @param {number} options.depth How many layers deep to perform the
+   * conversion. Defaults to infinite.
+   * @param {array} options.pyproxies If a list, ``toJs`` will store all
+   * PyProxies created in this list. This allows you to easily destroy all the
+   * PyProxies by iterating the list without having to recurse over the
+   * generated structure. If the value ``null`` is passed, then `toJs` will not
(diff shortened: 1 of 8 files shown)


src/core/pyproxy.js
```diff
@@ -348,26 +348,38 @@ class PyProxyClass {
...
+   * use ``proxy.toJs({depth : 1})``. See :ref:`Explicit Conversion of PyProxy
    * <type-translations-pyproxy-to-js>` for more info.
    *
    * @param {object} options
-   * @param {number} options.depth How many layers deep to perform the conversion.
-   * Defaults to infinite.
+   * @param {number} [options.depth] How many layers deep to perform the
+   * conversion. Defaults to infinite.
+   * @param {array} [options.pyproxies] If provided, ``toJs`` will store all
+   * PyProxies created in this list. This allows you to easily destroy all the
```
rth, line 359: It would be also worth mentioning that's it's expected to pass an empty array.

src/core/python2js.c
```diff
@@ -481,10 +481,20 @@ to_js(PyObject* self,
...
-  static const char* const _keywords[] = { "depth", 0 };
-  static struct _PyArg_Parser _parser = { "O|$i:to_js", _keywords, 0 };
-  if (kwnames != NULL && !_PyArg_ParseStackAndKeywords(
-                           args, nargs, kwnames, &_parser, &obj, &depth)) {
+  PyObject* pyproxies = NULL;
+  bool create_proxies = true;
+  static const char* const _keywords[] = {
+    "", "depth", "pyproxies", "create_pyproxies", 0
+  };
+  static struct _PyArg_Parser _parser = { "O|$iOp:to_js", _keywords, 0 };
```
rth, line 489: I know it was there before, but could you please add a link in a comment explaining how to decode that `O|$iOp` value ?

src/py/_pyodide/_core_docs.py
```diff
@@ -122,7 +125,13 @@ def create_proxy(obj: Any) -> JsProxy:
...
     # from python2js
 
-    def to_js(obj: Any, depth: int = -1) -> JsProxy:
+    def to_js(
+        obj: Any,
+        *,
+        depth: int = -1,
+        pyproxies: JsProxy = None,
+        create_pyproxies: bool = True
+    ) -> JsProxy:
```
rth, line 134: Generally we would need to add documentation for parameters,
```
Parameters
----------
obj : 
     Object to convert
depth : int, default=-1
     ...
```

and ideally a docstring example of how to use this. 

In `pyproxies` it would be good to mention that's it's a JsProxy of a JS Array.

src/core/pyproxy.js
hoodmane: To be clear there is a perfectly valid use case for passing nonempty arrays. Namely you may want to batch deallocate:
```js
let a = [];
let x1 = px1.toJs({ pyproxies : a});
let x2 = px2.toJs({ pyproxies : a});
let x3 = px3.toJs({ pyproxies : a});
// Do stuff
// ...
// Reclaim all proxies
for(let p of a){
   p.destroy();
}
``
I will add extra explanation. It would be useful to add a similar option to `runPython`, but it may require more careful consideration first.

rth (approved):
Thanks!

Issue #758, opened 2020-09-20
Prompt of `input(prompt="..")` is not shown in the browser input dialog


albertas:
**Expected behaviour**
Browser input prompt with message provided in `input(prompt="..")` is shown.

**Actual behaviour**
Prompt with message "Input" appears, instead of `prompt` value.

**Versions**
Firefox 79.0
pyodide 0.15.0

albertas:
Managed to fix this behavior by adding this code:

```
<script>
function input_fixed(text) {
    var value = prompt(text);
    return value.toString();
};
window.input_fixed = input_fixed;

languagePluginLoader.then(() => {
  pyodide.runPythonAsync(`
from js import input_fixed

input = input_fixed
__builtins__.input = input_fixed
`);
});
</script>
```

ChristopherHButler:
Has this been fixed in pyodide? I tested using input with this `choice = input("Enter choice(1/2/3/4): ")` and when I click the `ok` button the dialog just keeps prompting and no idea what actually happens to the value I gave as input

ChristopherHButler:
Ok I used @albertas trick and it works nicely :)

hoodmane:
Indeed has not been fixed. Hopefully we'll add support for synchronous IO in v0.18.

albertas:
I will just mention here, that `breakpoint()` or `import pdb; pdb.set_trace()` are also not working in Pyodide. And the reason is probably the same as with `input()`, since input field keeps reappearing.

hoodmane:
By the way, I have a demo with a synchronous IO implementation here. I think input, pdb, etc work with it.
https://spectralsequences.com/worker-pyodide-console/

hoodmane:
Actually I'm not sure it is working right now, there seems to be a CORS problem or something.

hoodmane:
Nevermind, it works it just takes forever to download b/c I haven't set up the server to zip the files.

stefnotch:
I mentioned this on Gitter, but I suppose noting it down here would be useful. I took a stab at this, including synchronous I/O and think I figured out how to do so properly

First I edited `pyodide.js` (now called `load-pyodide.js`) to support redirecting stdin. While I was at it, I also decided to support redirecting the print and printErr functions
```js
// Inject custom stdin/stdout/stderr
if(config.stdin) {
  Module.preRun = [function() {
    Module.FS.init(config.stdin, null, null);
  }];
}
if(config.print) {
  Module.print = config.print;
}
[...]

PR #1728, opened 2021-07-21
Allow overriding input

stefnotch:
Partially addresses https://github.com/pyodide/pyodide/issues/758#issuecomment-792253221

This lets the user override the input and output functions as follows

```js
self.loadPyodide({
  indexURL: ...,
  stdin: () => prompt(),
  stdout: (text) => console.log(text),
  stderr: (text) => console.error(text),
});
```

I hope this compiles, but I have no idea, since getting Pyodide to compile on my machine is hard.


src/js/pyodide.js
```diff
@@ -162,12 +161,18 @@ function fixRecursionLimit() {
  * (This can be fixed once `Firefox adopts support for ES6 modules in webworkers
  * <https://bugzilla.mozilla.org/show_bug.cgi?id=1247687>`_.)
  *
- * @param {{ indexURL : string, fullStdLib? : boolean = true }} config
+ * @param {{ indexURL : string, fullStdLib? : boolean = true, input?: () => string, print?: (text: string) => void, printErr?: (text: string) => void }} config
```
stefnotch, line 164: I wonder if someday it would be worthwhile to use Typescript. Currently, this doesn't get recognized by Visual Studio Code.

rth: Indeed. Opened https://github.com/pyodide/pyodide/issues/1729

src/js/pyodide.js
```diff
@@ -198,6 +203,35 @@ export async function loadPyodide(config) {
   Module.indexURL = baseURL;
   let packageIndexReady = initializePackageIndex(baseURL);
 
+  if(config.input) {
```
hoodmane, line 206: Could you factor this out into a function in `module.js` called `setStandardStreams(stdin, stdout, stderr)`, import it into this file and here just invoke `setStandardStreams(config.input, config.print, config.printErr)`? I think that's the best place to put this.

src/js/pyodide.js
```diff
@@ -162,12 +161,18 @@ function fixRecursionLimit() {
...
  * @param {string} config.indexURL - The URL from which Pyodide will load
  * packages
  * @param {boolean} config.fullStdLib - Load the full Python standard library.
  * Setting this to false excludes following modules: distutils.
  * Default: true
+ * @param {undefined | (() => string)} config.input - Override the standard input callback. Should ask the user for one line of input.
+ * Default: undefined
+ * @param {undefined | ((text: string) => void)} config.print - Override the standard output callback.
+ * Default: undefined
+ * @param {undefined | ((text: string) => void)} config.printErr - Override the standard error output callback.
```
hoodmane, line 174: How about calling these options `stdin`, `stdout`, and `stderr`?

src/js/pyodide.js
```diff
@@ -198,6 +203,35 @@ export async function loadPyodide(config) {
...
+
+      if (inputIndex < input.length) {
+        let character = input[inputIndex];
+        inputIndex++;
+        return character;
+      } else {
+        inputIndex = -1;
+        return null;
+      }
+    }
```
hoodmane, line 223: Is this function right? It always returns `null` once after every complete string, and it doesn't seem to allow `config.input` to respond that it doesn't have any data. I would think something more like this:
```js
    let input = [];
    let inputIndex = 0;
    let encoder = new TextEncoder();
    function stdin() {
      if(inputIndex < input.length){
        let character = input[inputIndex];
	    inputIndex++;
	    return character;
      }
      let newInput = config.input();
      if(newInput === undefined){
        return null;
      } 
      if(typeof newInput !== "string"){
      	throw new TypeError(`stdin callback returned invalid output type ${typeof newInput}, expected a string or undefined.`);
      }
      input = TextEncoder.encode(newInput + "\n");
      inputIndex = 1;
      return input[0];
    }
```

hoodmane (commented):
Might be good to add a test, though that'll be tricky if you can't get it working locally.

src/js/pyodide.js
stefnotch: Sure, I can do that. I initially didn't do it, since what they're exposing is already a bit of a wrapper around Emscripten's `stdin`, `stdout`, `stderr`

src/js/pyodide.js
stefnotch: I wrote the function with the typical behavior of `input` in mind.

So, when something from stdin is requested, the `config.input()` function gets called exactly once for every line.
To tell Emscripten that the line/user input is done, I have to return `null` after it. Otherwise a single `input()` call would result in multiple prompts until the user hits the cancel button.

src/js/pyodide.js
stefnotch: Of course! Should it be possible to call those functions at any point during runtime or only before Pyodide gets loaded?

hoodmane: I was thinking the `setStandardStreams` function should be private, only invoked in `loadPyodide`.

src/js/pyodide.js
hoodmane: Great that makes sense. I assume it doesn't work if you return 0? If returning 0 would work, we could just say `TextEncoder.encode(newInput + "\n\x00");` in my code above?

src/js/module.js
```diff
@@ -8,3 +8,43 @@ Module.noImageDecoding = true;
...
+ * @param {undefined|((text: string) => void)} stdout
+ * @param {undefined|((text: string) => void)} stderr
+ */
+export function setStandardStreams(stdin, stdout, stderr) {
+  if(stdin) {
+    let input = [];
+    let inputIndex = -1; // -1 means that we just returned null
+    function stdinWrapper() {
+      if (inputIndex === -1) {
+        input = Module.intArrayFromString((stdin() || "") + "\n", true, 0);
```
hoodmane, line 24: In any case it would be better to use `TextEncoder` here than `intArrayFromString`.
Also, what about the case where `stin()` returns `undefined` or `null`? Do we really want that to be the same as if it returned an empty string?

src/js/pyodide.js
stefnotch: @hoodmane I'll admit, I haven't tried it out. It's possible that it would work, but I'd prefer to rely on the little bit of emscripten that's documented if possible

> It should return an ASCII character code when data is available, or `null` when it isn’t.
> -- https://emscripten.org/docs/api_reference/Filesystem-API.html#setting-up-standard-i-o-devices

src/js/module.js
```diff
@@ -8,3 +8,43 @@ Module.noImageDecoding = true;
...
 Module.noWasmDecoding = false; // we preload wasm using the built in plugin now
 Module.preloadedWasm = {};
+
+/**
+ *
+ * @param {undefined|(() => string)} stdin
+ * @param {undefined|((text: string) => void)} stdout
+ * @param {undefined|((text: string) => void)} stderr
+ */
+export function setStandardStreams(stdin, stdout, stderr) {
```
hoodmane, line 18: How about this for the function body:
```
    let input = [];
    let inputIndex = 0;
    let encoder = new TextEncoder();
    function stdin() {
      if(inputIndex < input.length){
        let character = input[inputIndex];
	    inputIndex++;
	    return character;
      }
      let newInput = config.input();
      if(newInput === undefined){
        return null;
      } 
      if(typeof newInput !== "string"){
      	throw new TypeError(`stdin callback returned invalid output type ${typeof newInput}, expected a string or undefined.`);
      }
      input = encoder.encode(newInput + "\n\x00");
      inputIndex = 1;
      return input[0] || null; // we need to return a `null` to terminate the string.
    }
```

src/js/pyodide.js
stefnotch: Sounds good

src/js/module.js
stefnotch: Okay, I agree. It's probably better to use a proper web standard instead of emscripten internals.

I'm not sure how to best handle the case where `stdin` returns `undefined` or `null`. It would correspond to someone hitting the `cancel` button on a prompt...

hoodmane: I guess our options are (1) throw or (2) return `null`?

src/js/module.js
stefnotch: Seems nice! 
One minor issue that I have with it is that this would make inputting zero bytes into one line of `stdin` impossible. Sure, nobody is ever going to do that, but it feels weird using zero bytes as terminators for strings.

hoodmane: I guess this could hypothetically be harmful if the string had actual null bytes in it, but I suspect such null bytes will be interpreted by C to terminate the string anyways.

hoodmane: I will look at the internals and see if they pass the length on to the C APIs.

src/js/module.js
stefnotch: Okay, I took another look at this

Code
```py
print(input())
```

Input: Empty string (just hit ok on the prompt)
Output: `""` (also an empty string)

Input: `null` (hit cancel on the prompt)
Output: 
```
  File "/lib/python3.8/asyncio/futures.py", line 178, in result
    raise self._exception
  File "/lib/python3.8/asyncio/tasks.py", line 280, in __step
    result = coro.send(None)
  File "/lib/python3.8/site-packages/pyodide/_base.py", line 419, in eval_code_async
    return await CodeRunner(
  File "/lib/python3.8/site-packages/pyodide/_base.py", line 276, in run_async
    res = eval(last_expr, self.globals, self.locals)
EOFError: EOF when reading a line
 ```

File: src/js/module.js (hunk 6 above)
Comment (stefnotch) [reply src/js/module.js:18]: I decided to just test it out by hardcoding the user input to `text = "cat\0dog";`

Apparently the null bytes really do get preserved. I'm pretty surprised at the fact that emscripten actually handles that
image

Comment (hoodmane) [reply src/js/module.js:18]: Hmm it does look like without this null bytes would be handled correctly since `read` returns how many bytes were written to the buffer. So better to keep the approach that handles null bytes just in case (they are allowed to appear both in js strings and in Python strings, so no reason we should convert them wrong).

File: src/js/module.js (hunk 5 above)
Comment (stefnotch) [reply src/js/module.js]: I guess it's your call as to whether you'd prefer an input of `null` (and also `undefined) to result in an empty user input or an exception.

Comment (hoodmane) [reply src/js/module.js]: I think cancel is a bit like `Ctrl+C`. In the native console, that would raise `KeyboardInterrupt`. So I think raising `EOF` is good in that case.

File: src/js/module.js [hunk 7]
```diff
@@ -8,3 +8,44 @@ Module.noImageDecoding = true;
...
+  if(stdin) {
+    const encoder = new TextEncoder();
+    let input = new Uint8Array(0);
+    let inputIndex = -1; // -1 means that we just returned null
+    function stdinWrapper() {
+      if (inputIndex === -1) {
+        input = encoder.encode(text + (text.endsWith("\n") ? "" : "\n"));
```
hoodmane, line 25:
```suggestion
		let text = stdin();
		if(text === undefined || text === null){
			return null;
		}
		if(typeof text !== "string"){
            throw new TypeError(); // emscripten will catch this and set an IOError
		}
		if(!text.endsWith("\n")){
			text += "\n";
		}
        input = encoder.encode(text);
```

stefnotch: Oh thank you very much! I was about to write this code

src/js/module.js
stefnotch: Sounds good then!

src/js/module.js
```diff
@@ -8,3 +8,54 @@ Module.noImageDecoding = true;
...
+ * @param {undefined|(() => string)} stdin
+ * @param {undefined|((text: string) => void)} stdout
+ * @param {undefined|((text: string) => void)} stderr
+ */
+export function setStandardStreams(stdin, stdout, stderr) {
+  if(stdin) {
+    const encoder = new TextEncoder();
+    let input = new Uint8Array(0);
+    let inputIndex = -1; // -1 means that we just returned null
+    function stdinWrapper() {
```
hoodmane, line 23: Could we add a comment here explaining what we are supposed to be doing (return one the string data one byte one at a time, then return a `null` to indicate that end of input)?

stefnotch: Good point, this sort of stuff deserves a comment explaining it/

hoodmane:
We also need a test, something like:
```py
def test_custom_stdin_stdout(selenium_webworker_standalone):
	selenium.run_js(
		"""
		function* stdinStrings(){
			yield "hello world";
			yield "hello world\n";
			yield "This has a \x00 null byte in the middle...";
			yield "pyodidé";
			yield "碘化物";
			yield "🐍";
		}
		let stdinStringsGen = stdinStrings();
		function stdin(){
			return stdinStringsGen.next();		
		}
		self.stdoutStrings = [];
		self.stderrStrings = [];
		function stdout(s){
			stdoutStrings.push(s);
		}
		function stderr(s){
			stderrStrings.push(s);
		}
			indexURL : './',
			fullStdLib: false,
			jsglobals : self,
			stdin,
			stdout,
			stderr,			
		});
        self.pyodide = pyodide;
        globalThis.pyodide = pyodide;
		"""
	)
	selenium.save_state()
	selenium.run_js(
		"""
		pyodide.runPython(`
			assert input() == "hello world"
			# ... test more stuff
		`);
		"""
	)
```

hoodmane:
Except we can't use `selenium_worker_standalone` because that skips the tests on `node`, so we need to update `conftest.py` with a new fixture, say `selenium_standalone_noload` which is exactly the same as `selenium_worker_standalone` except also running in `node` and then use that.

Test should also probably have:
```py
```

hoodmane:
Also, we need a changelog entry.

hoodmane:
And you'll need to run the formatters to fix lint errors.

hoodmane:
Also do we want to change the default behavior since the current default stdin behavior is a bit broken?

stefnotch:
I was hoping that someone else could take care of the tests and formatter, since getting Pyodide to compile and run on my machine would involve a significant number of hours of my time. Getting Docker to work nicely on Windows is a pain.

I added a changelog entry.

Sure, we can change the default behavior. I'd suggest `stdin ?? prompt`

hoodmane:
> someone else could take care of the tests and formatter

Can do, lemme see if I can manage to push here...

hoodmane:
> I'd suggest stdin ?? prompt

Sounds good to me.

stefnotch:
> 
> 
> > someone else could take care of the tests and formatter
> 
> Can do, lemme see if I can manage to push here...

I think you should be able to push to this branch, if it doesn't work, I'm happy to add you to my fork.
Also, thank you very much for all your work and awesome feedback on this! 

> 
> 
> > I'd suggest stdin ?? prompt
> 
> Sounds good to me.

Alright, I committed that change

hoodmane:
Could you enable circleci on your fork @stefnotch?

stefnotch:
@hoodmane Sure thing, how do I do that?

Also, it looks like ci/circleci is already running? But I cannot see the details, since I don't have a circleci account.

image

hoodmane:
Follow these instructions: 
https://circleci.com/docs/2.0/getting-started/?section=getting-started#setting-up-circleci

stefnotch:
> 
> 
> Follow these instructions:
> https://circleci.com/docs/2.0/getting-started/?section=getting-started#setting-up-circleci

Thank you, it should be up and running now.

hoodmane:
Indeed, looks good.

stefnotch:
Interesting, some of the test now fail, complaining that `prompt is not defined`. Should I first check if it's defined and only then set it as a default?

src/js/module.js
```diff
@@ -8,3 +8,72 @@ Module.noImageDecoding = true;
...
+  }
+
+  if (stderr) {
+    Module.printErr = stderr;
+  }
+
+  if (stdin) {
+    Module.preRun = [
+      function () {
+        Module.FS.init(createStdinWrapper(stdin), null, null);
```
hoodmane, line 30: Could we have a comment explaining why we set `stdin` with `FS.init` but set `stdout` and `stderr` with `Module.print` and `Module.printErr`?

stefnotch: Sure, I added a comment. It's basically because that's how the Emscripten API is.

hoodmane: Perfect.

hoodmane:
@stefnotch If you have no other concerns I will merge this when CI passes.

stefnotch:
@hoodmane I don't think I have anything left to add to this PR. Looking forward to seeing this being merged 🚀

stefnotch:
All of the CI tests passed, except for `ci/circleci: test-packages-chrome`, `ci/circleci: test-packages-firefox` and `ci/circleci: test-packages-node`. I'm not entirely sure what's up with those failures.

hoodmane:
> I'm not entirely sure what's up with those failures.

They are sporadic failures. Not related to this PR.

tomitrescak:
Hello, does this solve #758 ? I get no value in stdin()

stefnotch:
It doesn't solve #758 , it only addresses the part about the `prompt()`s being called multiple times. It also makes them overrideable.

I didn't fix the actual issue, because I'm not sure what the best way of doing it is. One option would be to take the last thing that was written to stdout (without a newline) and display that in the prompt.

The other option would be to listen for an auditing event.
> Raises an auditing event `builtins.input` with argument `prompt` before reading input
> -- https://docs.python.org/3/library/functions.html#input

tomitrescak:
Thank you very much for your response. The multiple calls are already quite helpful!

PR #1731, opened 2021-07-21
Update mkpkg.py

LinasNas:
This PR is for issue https://github.com/pyodide/pyodide/issues/977

It modifies the `make_package` function in `pyodide-build/pyodide_build/mkpkg.py`

During the creation of metadata.yaml file, it adds four fields of general information about the package - link to homepage, link to PyPi, summary, and license. 

Example of information which would be added at the end of [metadata.yaml file in pandas folder](https://github.com/mainsail-org/pyodide/blob/main/packages/pandas/meta.yaml):

```yaml
about:
  home: https://pandas.pydata.org
  PyPi: https://pypi.org/project/pandas
  summary: Powerful data structures for data analysis, time series, and statistics
  license: BSD-3-Clause
```

### Questions: 
- Are the field names okay? I used conventions from [conda](https://docs.conda.io/projects/conda-build/en/latest/resources/define-metadata.html#about-section)
- In order to preserve the order of how keys are dumped to yaml file, I used ```ruamel.yaml``` package, instead of ```yaml```. I've noticed it was used already in ```update_package``` function, and it was suggested in the issue. For this reason, I also deleted the argument ```default_flow_style=False``` which doesn't exist in ruamel. Is that alright? 
- Are there any other fields that would be useful to add? For example, a link to the json endpoint on PyPi's API. Here's an example link for pandas: https://pypi.org/pypi/pandas/json

### DIFF
--- a/pyodide-build/pyodide_build/io.py
+++ b/pyodide-build/pyodide_build/io.py
@@ -35,6 +35,12 @@ PACKAGE_CONFIG_SPEC: Dict[str, Dict[str, Any]] = {
     "test": {
         "imports": list,  # List[str]
     },
+    "about": {
+        "home": str,
+        "PyPi": str,
+        "summary": str,
+        "license": str,
+    },
 }
 
 
--- a/pyodide-build/pyodide_build/mkpkg.py
+++ b/pyodide-build/pyodide_build/mkpkg.py
@@ -63,7 +63,8 @@ def make_package(package: str, version: Optional[str] = None):
     Creates a template that will work for most pure Python packages,
     but will have to be edited for more complex things.
     """
-    import yaml
+    from ruamel.yaml import YAML
+    yaml = YAML()
 
     pypi_metadata = _get_metadata(package, version)
     sdist_metadata = _extract_sdist(pypi_metadata)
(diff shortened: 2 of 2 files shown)


hoodmane:
>  Are the field names okay? I used conventions from conda

Looks good to me.

> I used ruamel.yaml ... Is that alright?

+1 sounds good

> Are there any other fields that would be useful to add? For example, a link to the json endpoint on PyPi's API. Here's an example link for pandas: pypi.org/pypi/pandas/json

Well that particular info seems redundant, since it's just pypi.org/pypi/{pkgname}/json.

hoodmane (approved):

hoodmane:
Please run the linter and update `PACKAGE_CONFIG_SPEC` in `io.py` to fix the failing test `test_mkpkg`.

LinasNas:
Hi,
I've just committed changes to ```pyodide-build/pyodide_build/io.py```:

1. In ```PACKAGE_CONFIG_SPEC```, I added "about" key and specified the type for each of its values
2. I then ran black on  ```io.py``` to format the file. 

Is that what you had in mind?

hoodmane:
Yes, but lint check is still failing.

rth:
For future reference, it's easier to run black on `.` to make sure all changed files are correctly formatted.

PR #1732, opened 2021-07-22
MAINT: Pull python2js conversion state into a struct

hoodmane:
This puts the conversion state (depth, cache, and proxy list) into a struct, improving code readability and making it easier to add new stuff to the conversion state without changing a bunch of function signatures.

Partial work towards adding parametrization requested in #1529.


hoodmane:
Maybe it would be better named `ConversionContext`?

PR #1733, opened 2021-07-22
Increase recursion limit

hoodmane:
#1699 reduced the stack usage by a lot. `benchmark-stack-size` shows `js_depth/py_depth` is 19 in Firefox and 8 in Chrome, so dividing `stack_depth` by 20 to get the Python recursion depth seems reasonable.

I guess we'll find out if certain call patterns lead to callstacks with lots of "fat frames" that lead to segfaults. Most of the stack overflows we were seeing before were from pure Python code so I'm not that worried about this. (I'm pretty sure all Python ==> Python function calls take up exactly the same amount of stack space.)

### DIFF
--- a/src/core/main.c
+++ b/src/core/main.c
@@ -94,6 +94,13 @@ run_python_simple_inner(char* code)
 int
 numpy_patch_init();
 
+int
+get_python_stack_depth()
+{
+  PyThreadState* tstate = PyThreadState_GET();
+  return tstate->recursion_depth;
+}
+
 /**
  * Bootstrap steps here:
  *  1. Initialize init_dict so that runPythonSimple will work.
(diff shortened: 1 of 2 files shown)


rth:
It does fail in Firefox currently `InternalError: too much recursion`. Looks like we would need to increase the divisor a bit.

hoodmane:
It seems like we really need to have a different formula in Firefox.

hoodmane:
Interestingly, it seems like since my patches to remove the fpcasts, v8 probably supports more recursion than Firefox. More research is probably needed into how this works, it is still pretty mysterious to me. `benchmark-stack-size` is really helpful though.

Currently, the formula leads to a recursion limit of 500 across the board (see benchmark-stack-size).

Anyways I'm going to merge this, I think this new formula is better than the old one even if it is far from perfect.

hoodmane:
Oops, didn't realize `packages` tests hadn't finished. Oh well.

PR #1742, opened 2021-07-23
Add dict_converter parameter to toJs

### LINKED ISSUE #1529: Passing python dict into a js function

hoodmane:
This resolves #1529 by allowing `dict_converter=Object.fromEntries`.

Perhaps we should rename the options for `toJs` to follow the Javascript naming conventions (so then `to_js` would have `dict_converter` and `toJs` would have `dictConverter`.

### DIFF
--- a/src/core/python2js.c
+++ b/src/core/python2js.c
@@ -20,11 +20,20 @@ _python2js_immutable(PyObject* x);
 int
 _python2js_add_to_cache(PyObject* cache, PyObject* pyparent, JsRef jsparent);
 
-typedef struct
+struct ConversionContext_s;
+
+typedef struct ConversionContext_s
 {
   PyObject* cache;
   int depth;
   JsRef proxies;
+  JsRef jscontext;
+  JsRef (*dict_new)(struct ConversionContext_s context);
+  int (*dict_add_keyvalue)(JsRef target,
+                           JsRef key,
+                           JsRef value,
+                           struct ConversionContext_s context);
+  JsRef (*dict_post)(JsRef dict, struct ConversionContext_s context);
 } ConversionContext;
 
 JsRef
@@ -156,7 +165,8 @@ _python2js_dict(PyObject* x, ConversionContext context)
   // result:
   JsRef jsdict = NULL;
 
-  jsdict = JsMap_New();
+  jsdict = context.dict_new(context);
+  FAIL_IF_NULL(jsdict);
   FAIL_IF_MINUS_ONE(_python2js_add_to_cache(context.cache, x, jsdict));
   PyObject *pykey, *pyval;
   Py_ssize_t pos = 0;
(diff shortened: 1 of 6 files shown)


src/py/_pyodide/_core_docs.py
```diff
@@ -144,22 +145,30 @@ def to_js(
...
-        pyproxies: JsProxy, default = None
-            Should be a Javascript ``Array``. If provided, any ``PyProxies`` generated
-            will be stored here. You can later use :any:`destroy_proxies` if you want
-            to destroy the proxies from Python (or from Javascript you can just iterate
-            over the ``Array`` and destroy the proxies).
-
-        create_pyproxies: bool, default=True
-            If you set this to False, :any:`to_js` will raise an error
-
+        obj : Any The Python object to convert
```
rth, line 148: Even, if this seems to currently render fine, there is really a spec for Python docstrings so it should be,
```
x : type
    Description of parameter `x`.
```
with optionally a default value, as it was before. Indentation value for the Description doesn't matter. https://numpydoc.readthedocs.io/en/latest/format.html#parameters

src/core/python2js.c
```diff
@@ -486,6 +511,85 @@ python2js_with_depth(PyObject* x, int depth, JsRef proxies)
...
+}
+
+JsRef
+python2js_custom_dict_converter(PyObject* x,
+                                int depth,
+                                JsRef proxies,
+                                JsRef dict_converter)
+{
+  if (dict_converter == NULL) {
+    return python2js_with_depth(x, depth, proxies);
```
rth, line 572: Maybe add a comment that by default it converts to a Map?

rth (commented):
Thanks! Generally LGTM, except that the formatting of Python docstrings in to_js would need to be reverted. (could `[skip ci]` in commits) (cf comment below)

src/py/_pyodide/_core_docs.py
hoodmane: Oops, my text editor textwrap tool did this.

hoodmane: This is why I had the blank lines in there before.

PR #1748, opened 2021-07-25
Improve documentation around runPythonAsync (#1748) [skip ci]

wlach:
* Fix some formatting issues that caused the example to render incorrectly
* Add a warning that this function will no longer load any imports for you


hoodmane:
Thanks! The example code is also wrong...

PR #1749, opened 2021-07-26
MAINT Add Js_IDENTIFIER, new hiwire_CallMethod commands

hoodmane:


danainacore (commented):
- -

rth:
Could you add more description about the goal of this PR?

hoodmane:
We have two ways of making Javascript method calls from C:
1. via a helper `EM_JS` function (or using `EM_ASM` but that has various drawbacks).
2. via one of the `hiwire_call` functions

Having `EM_JS` helper functions is not great for code readability because they tend to contain chunks of logic that would not otherwise be factored out as a separate function if it weren't for the ffi constraints. Prior to these changes we also have the problem that most of these helper functions do the obvious thing, but some have surprising extra logic (e.g., `hiwire_call_get_method`). So for these reasons it's preferable to avoid helper methods when possible. When helper functions are necessary it would be good to name them in a standard way derived from the function they are helpers for and to put the definitions of the helper functions next to the function they are helpers for, rather than putting them in `hiwire.c` even if they perform a pretty specific task only to be used in one place.

On the other hand, `hiwire_call_method` has the issue that it does repeated conversions of C strings to Javascript every time the method is called. So we want to avoid helper methods, but we want to have static Javascript strings from C. CPython has the same issue of needing a convenient way to declare and use static Python strings from C, which uses this `Py_IDENTIFIER` macro. I copied the definition of `Py_IDENTIFIER` to make a `Js_IDENTIFIER` macro which serves the same purpose but for static Javascript strings.

With this in place, I replaced all the `hiwire_has_XXX_method` calls in `jsproxy.c` with a single `hiwire_HasMethodId` call that can be used for all of them, and similarly almost all of the `hiwire_call_XXX_method` calls can be replaced by `hiwire_CallMethodId` variants. 

The one surprising one `hiwire_call_get_method` which has extra logic to try to distinguish "value missing" from "value present but is undefined". This got moved out of hiwire into a helper function defined right next to the function body, making it clearer that there is custom logic.

src/core/hiwire.h
```diff
@@ -47,6 +47,20 @@ extern const JsRef Js_null;
...
+  JsRef object;
+} Js_Identifier;
+
+#define Js_static_string_init(value)                                           \
+  {                                                                            \
+    .string = value, .object = NULL                                            \
+  }
+#define Js_static_string(varname, value)                                       \
+  static Js_Identifier varname = Js_static_string_init(value)
+#define Js_IDENTIFIER(varname) Js_static_string(JsId_##varname, #varname)
```
rth, line 62: Maybe you could add a comment to indicate that this is defined by analogy with https://github.com/python/cpython/blob/24da544014f78e6f1440d5ce5c2d14794a020340/Include/cpython/object.h#L17 to give a bit more context about why this is done for anyone reading it?

src/core/jsproxy.c
```diff
@@ -479,6 +486,24 @@ JsProxy_ass_subscript_array(PyObject* o, PyObject* item, PyObject* pyvalue)
   return success ? 0 : -1;
 }
 
+EM_JS_REF(JsRef, JsProxy_subscript_js, (JsRef idobj, JsRef idkey), {
```
rth, line 489: A 1 line (or a few line) description of what this does would be useful.

rth (approved):
Thanks for this PR. It's good to be able to remove those specialized `hiwire_has_XXX_method` methods. LGTM aside from a couple of minor comments.

PR #1750, opened 2021-07-26
Don't bind a port in run_docker by default

hoodmane:


rth (approved):
Sounds good, but would probably deserve a changelog entry, as this can be unexpected for users.

PR #1753, opened 2021-07-27
Use lazy_wheel in micropip

### LINKED ISSUE #1619: Strange behavior with importing seaborn !
Hello! 
Someone ran into issues importing seaborn through a starboard notebook and, at first it seemed like a simple issue of wheel availability and versions. 
But in trying to find a specific version that maybe works, something just doesn't seem right. 

Ultimately I think that 
micropip.install('seaborn==0.9.0') worked in pyodide 0.16 but it doesn't seem to work in 0.17, and I can't quite figure out why! 

The post + my original response is here, for further context: 
https://community.starboard.gg/t/issue-with-micropip-installing-packages/100

### LINKED ISSUE #1745: Fix micropip dependency resolution for old package versions
Micropip does not correctly handle dependency resolution for old package versions #1619. It is difficult to fix because collecting the dependencies for old versions of a package from pypi is not that easy. pip uses a complicated scheme involving range requests, but pypi does not set CORS headers on responses to range requests pypa/conveyor#11 so we cannot use the same approach.

I opened another issue about this pypa/warehouse#9835 suggesting that they should add some simpler method of getting requirements of old package versions. This would allow improvements to pip too.

I think in the meantime since #1596 has been merged we should consider moving to a slower but correct approach and caching dependencies in IDBFS if possible. Perhaps other methods of caching package dependencies could also be explored.

hoodmane:
This fixes #1619 and #1745. We still use a greedy algorithm to resolve dependencies, but we use the correct dependencies for old versions of packages and for packages hosted at custom URLs.

Most of the new code is copied directly from pip, with some dependencies removed and the requests in `lazy_wheel` replaced with web apis.

Large portions of `pkg_resources.py` can probably be deleted. I copied it from https://github.com/pypa/pip/blob/main/src/pip/_vendor/pkg_resources/__init__.py and then deleted just the things that caused trouble.

Closes #1619
Closes #1745


rth:
If we adapt code from pip with a few modifications, I think it would be better to put it inder micropip/externals/pip/ folder (ideally preserving more or less their tree structure) so it's a bit clearer what's our code what is the code from external sources. It would make it more obvious to sync with upstream changes if necessary in the future for instance.  It also makes it possible to assume that that code is tested and we don't need to write tests for it (the folder can be excluded from code coverage).

> Large portions of pkg_resources.py can probably be deleted.

Yes, running https://pypi.org/project/vulture/ on that code might help.

hoodmane:
> running pypi.org/project/vulture on that code might help

It was indeed useful. Reduced file to 500 lines from 3000.

hoodmane:
> it would be better to put it under micropip/externals/pip/ folder (ideally preserving more or less their tree structure)

Done.

packages/micropip/src/micropip/externals/pip/_internal/network/lazy_wheel.py
```diff
@@ -0,0 +1,257 @@
+"""Adapted from
+https://github.com/pypa/pip/blob/main/src/pip/_internal/network/lazy_wheel.py"""
```
rth, line 2: Could you indicate the version used (or commit hash)? Maybe in `packages/micropip/src/micropip/externals/pip/__init__.py` and then this header is no longer necessary. The link to master is going to be non accurate in a few months/years anyway.

packages/micropip/src/micropip/micropip.py
```diff
@@ -21,10 +22,11 @@
...
 
 if IN_BROWSER:
-    # In practice, this is the `site-packages` directory.
-    WHEEL_BASE = Path(__file__).parent
+    import site
+
+    WHEEL_BASE = site.getsitepackages()[0]
 else:
-    WHEEL_BASE = Path(".") / "wheels"
+    WHEEL_BASE = str(Path(".") / "wheels")
```
rth, line 29: Probably better to keep paths as `Path` (i.e. make the one above a Path as well). It's used in `Zipfile.extractall` which should support it.

src/py/_pyodide/console.py
```diff
@@ -97,8 +98,13 @@ def __init__(
...
     def __call__(self, source, filename, symbol) -> CodeRunner:  # type: ignore
         return_mode = self.return_mode
-        if self.quiet_trailing_semicolon and should_quiet(source):
-            return_mode = None
+        try:
+            if self.quiet_trailing_semicolon and should_quiet(source):
+                return_mode = None
+        except TokenError:
+            # Invalid code, let the Python parser throw the error later.
+            pass
```
rth, line 106: Is this related to this PR?

rth (commented):
Thanks! Overall looks good to me assuming CI passes. So this would make one more API call to PyPi per package, right?

src/py/_pyodide/console.py
hoodmane: Nope, just saw this issue in the process of working on this PR. I will move it into a separate PR.

rth:
Could you add a changelog entry as well?

hoodmane:
> assuming CI passes

It doesn't work quite yet, I need to add an extra hack to get around the fact that `fetch` is asynchronous but zip expects reading from a file descriptor to be synchronous.

> one more API call to PyPi per package, right?

I think it will most commonly it will make two, but this depends on implementation details of `zipfile.py`. I will write a detailed explanation of how it works in a few minutes.

hoodmane:
It seems possible to make it most often take one request, sometimes two, and very rarely more than two. If the wheel isn't from pypi, then it needs one more request to find out how big the wheel is.

hoodmane:
There is a problem loading the url in `test_install_custom_url`, I'm not sure what it is though.

rth:
Are you sure our test server supports range requests?

hoodmane:
Well it's not purely a range request issue because `test_install_custom_relative_url` passes but `test_install_custom_url` fails. Confusing...

hoodmane:
Okay so specifically what happens here:
1. If the wheel is from a custom URL, we need to make an api call to find out the size of the wheel. If it's from pypi, the metadata includes the size and we can skip this.
2. We make a temporary file of the same length as the zip file. We need to load enough data to get the zip header data. This is at the end of the zip file and has variable length depending on how many files are in the zip archive. We load 10kb at the end of the file to start, it certainly will include the `end of central directory` which has some variable length components but these seem they seem to be short or absent in wheels. For all but very big wheels it will also include all of the "central directory".
3. We try to unzip the temp file and record where reads occur.
4. If any reads occur outside of already loaded regions, we catch any error that occurs and load a chunk of the file around the read. Currently I am making 10kb requests, this is tuned towards preferring fewer requests over minimizing data loaded. Then we try to unzip again. When unzipping and loading the METADATA file succeeds, we don't need to load any more of the zip.

If we are lucky and the METADATA file is near the end, we might make only one request and get all the METADATA. This seems to happen a decent percentage of the time. If we are unlucky, we will probably have to make a second request to get the METADATA file. We generally shouldn't have to make more than two requests (plus one extra to get the size of the wheel from custom url).

I do sort of wonder whether this is actually faster than just loading the whole wheel.

hoodmane:
It does seem to be failing at the range request though.

rth:
Wow, I haven't realized that business with hacking the zip extraction files in pip. Though it makes sense to use the head information from the zip.

> I do sort of wonder whether this is actually faster than just loading the whole wheel.

Yes, so they do that because they want to be able to [support backtracking](https://pip.pypa.io/en/stable/topics/dependency-resolution/#backtracking) I assume without downloading downloading lots of wheels for a package. Also wheels can be huge (e.g. https://pypi.org/project/tensorflow/ is 450MB) but even allocating an empty file with it would still be somewhat of a problem.   

In our case though we don't support it yet. So once we try to install of version of a package A (A) that requires packages B (B) with version that's incompatible to the currently installed version B (X), we are not going to try a different version of A, right? So that means that each wheel will be downloaded only once anyway? 

Unless we do plan to support backtracking in the future (by adapting more of pip code) in which case supporting this would make sense.

hoodmane:
> Wow, I haven't realized that business with hacking the zip extraction files in pip. 

It's a cool hack but I think it's crazy that they do this.

> each wheel will be downloaded only once anyway?

I believe so. The thing is, in the ideal case we could do dependency resolution first and once dependency resolution is complete load all full wheels in parallel. Many of these LazyWheel instances only require a single 10kb request, though others require two 10kb requests again depending on layout of the wheel which is kind of random. If many wheels require only one request, then I think this LazyWheel thing could be a speed improvement even with a greedy resolver. But I agree it would be more helpful with a backtracking resolver.

I think if we ever do a backtracking resolver, we should put a strict limit on how much it will backtrack to achieve a good bound on runtime. A bit of backtracking can add convenience without wrecking the runtime. Guaranteeing to find a solution if it exists is a bad idea since dependency resolution is an NP Hard problem and hard cases appear in practice. It is better for everyone if we fail after at most a limited amount of backtracking so people can do the resolution manually and preload the right versions of dependencies. Hard coding solutions is extra work but it is worth it if it helps to avoid large and unpredictable runtime costs.

hoodmane:
Closed in favor of #1756.

PR #1755, opened 2021-07-29
Update webworker.md (#1755) [skip ci]

teonbrooks:
If I understand this correctly, loadPyodide now returns a pyodide object and doesn't attach to the global namespace. I think this should now refer to the `self.pyodide` instead of `pyodide`


hoodmane:
I think this is a neutral edit since `self` is the `WebworkerGlobalScope` so the only difference between `self.name` and `name` is what happens if `name` wasn't declared. Probably it's more obvious to a reader to say `self.pyodide`.

Thanks @teonbrooks!

PR #1756, opened 2021-07-29
Simpler dependency handling for micropip

### LINKED ISSUE #1619: Strange behavior with importing seaborn !

### LINKED ISSUE #1745: Fix micropip dependency resolution for old package versions

hoodmane:
This is a simpler alternative to #1753 that skips the whole `LazyWheel` thing and just downloads the whole wheel directly. Hopefully we can merge this for v0.18 and then assess the merits of `LazyWheel` / #1753 separately after the v0.18 release.

Closes #1619
Closes #1745


hoodmane:
This is around 2x faster than `LazyWheel` (installing sphinx takes about 10 seconds with `LazyWheel` and 5 seconds this way and on master). I wonder if it is even worth using `LazyWheel` for doing dependency resolution, probably it's better to make fewer requests and not worry about downloading unneeded data.

hoodmane:
Yet again `install-custom-url` is failing.

packages/micropip/src/micropip/micropip.py
```diff
@@ -222,13 +239,20 @@ async def add_requirement(self, requirement: str, ctx, transaction):
...
+    async def add_wheel(self, name, wheel, version, extras, ctx, transaction):
+        transaction["locked"][name] = version
+        response = await fetch(wheel["url"])
+        wheel_bytes = (await response.arrayBuffer()).to_py()
+        wheel["wheel_bytes"] = wheel_bytes
 
-        recurs_reqs = metadata.get("info", {}).get("requires_dist") or []
-        for recurs_req in recurs_reqs:
+        zip_file = ZipFile(io.BytesIO(wheel_bytes))  # type: ignore
+        dist = pkg_resources_distribution_for_wheel(zip_file, name, "???")
```
rth, line 251:
```suggestion
        with ZipFile(io.BytesIO(wheel_bytes)) as zip_file:  # type: ignore
            dist = pkg_resources_distribution_for_wheel(zip_file, name, "???")
```

rth (commented):
Generally LGTM. +1 to merge this simpler implementation. 

> Yet again install-custom-url is failing.

Hmm, looks like something is different between PyPi and hosting ourselves. Don't have much ideas about it short of spending time debugging..

hoodmane:
Okay it's definitely getting a 404.

hoodmane:
Okay the problem is really simple, I changed the url in the test and for some reason I didn't look at the diff with the test on `main` until just now.

hoodmane:
Should be fixed.

PR #1758, opened 2021-07-29
Avoid circular references when runsource raises SyntaxError

hoodmane:
Futures create risk of generating circular references because exceptions have a reference to the call frames that were unwound when they threw. If local variables in those frames carry a reference to the future, then this is a circular reference: future ==> exception ==> frame ==> future. These circular references can take a long time to be reclaimed and keep local variables around for much longer than necessary. This removes references to the Future from the call frame in a finally block and also strips the local variables from the call frame of SyntaxErrors (these frames would never be interesting anyways).


rth:
Looks like `test_console*` tests fail?

hoodmane:
Yeah I'm confused what is going on with that: setting a couple of variables equal to `None` in a finally block should be pretty harmless? This isn't super high priority though, I'll try to fix it when I get around to it.

PR #1760, opened 2021-07-30
Update libffi to tip of tree

hoodmane:
Libffi merged my changes to their test suite a while back so diff from upstream libffi shrank a lot.


rth (approved):
Thanks for contributing it upstream!

PR #1761, opened 2021-07-30
Add cffi package

hoodmane:


hoodmane:
@rth I think this works now (tests pass locally).

pyodide-build/pyodide_build/buildpkg.py
```diff
@@ -99,7 +99,7 @@ def download_and_extract(
         srcdir = Path(pkg["source"]["path"])
 
         if not srcdir.is_dir():
-            raise ValueError("'path' must point to a path")
+            raise ValueError("'path' must point to a directory that exists")
```
rth, line 102: Maybe,
```suggestion
            raise ValueError(f"path={srcdir} must point to a directory that exists")
```
while we are at it,

packages/cffi_example/meta.yaml
```diff
@@ -0,0 +1,11 @@
...
+  name: cffi_example
+  version: "0.1"
+requirements:
+  run:
+    - cffi
+source:
+  path: python-cffi-example
+test:
+  imports:
+    - cffi_example
```
rth, line 11: Could you add about/home link pointing to github as well and the license? Actually can't we install it from a Github release? Or did you make lots of changes to the example?

rth (approved):
Great! A few minor comments otherwise LGTM.

packages/cffi_example/meta.yaml
hoodmane: I did not make many any changes to the example (just deleted some unneeded files). I don't think there is support in `pyodide-build` for loading a github repo, and they don't have seem to have made any releases. But I agree we should fix it to download it from github in some way.

rth: Even without a release you should be able to download it as https://github.com/wolever/python-cffi-example/archive/refs/heads/master.zip  Or with the appropriate commit in the URL. It's the "Download Zip" button,
image

Issue #1725, opened 2021-07-19
Add docs for using Pyodide from Node


rth:
We would need to add some docs on how to use Pyodide with Node. For now probably the usage would involve,
 - npm installing Pyodide
 - downloading the archive of packages
 - pointing to the folder with packages when loading Pyodide

Then we can see if some caching https://github.com/pyodide/pyodide/issues/14#issuecomment-875116949 could making it into this release.

@hoodmane Can we decide under which name we are going to upload the npm package https://github.com/pyodide/pyodide/discussions/1403#discussioncomment-972302 ?

keverw:
Yeah, I was here looking how to use with Node! That would be nice! I'm taking a class for college where we use Python in the fall while I've already have experience with programming and a huge fan of TypeScript/Node ecosystem so being able to use Python code in my Node/React projects be nice! I know Python has a huge module ecosystem too, so sounds like the best of both worlds!

hoodmane:
I'm good with putting it at `pyodide`. It seems that a lot of packages have the main package at `some_name` and side modules at `@some_name/submodule`.

rth:
Let's do that then.

PR #1762, opened 2021-07-30
Release src/js as a npm package

rth:
Changes to be able to release of src/js as a standalone NPM package  (https://www.npmjs.com/package/pyodide).

Closes https://github.com/pyodide/pyodide/issues/328
xref: https://github.com/pyodide/pyodide/issues/14

Will add more documentation for using Pyodide in Node (https://github.com/pyodide/pyodide/issues/1725) in a follow up PR.


src/core/pyproxy.js
```diff
@@ -17,7 +17,7 @@
  * See Makefile recipe for src/js/pyproxy.js
  */
 
-import { Module } from "../js/module.js";
+import { Module } from "./module.js";
```
hoodmane, line 20: This will mess up the PyProxy API docs.

hoodmane: I guess we can make a symlink `pyproxy.js` ==> `../core/pyproxy.gen.js`.

hoodmane: Actually the output it doesn't seem to be messed up so I guess this change is fine by itself.

src/js/README.md
```diff
@@ -1,12 +1,40 @@
...
 
-The Javascript code in this folder is responsible for the following tasks:
+<a href="https://www.npmjs.com/package/pyodide"><img src="https://img.shields.io/npm/v/pyodide" alt="npm"></a>
 
-1. Defines the public Javascript API
+## Usage
+
+Download and extract Pyodide packages from [Github
+releases](https://github.com/pyodide/pyodide/releases)
+(`pyodide-build-*.tar.bz2`). The version of the release needs to match exactly the version of this package.
```
hoodmane, line 9: We should try to simplify this by copying the `build` directory into `src/js` in the `prepublish` script.

rth: I agree that we need to simplify, but as discussed in https://github.com/pyodide/pyodide/issues/14#issuecomment-875116949 I'm not convinced that bundling all packages is the solution. We need a solution that would also work for micropip installed packages anyway. 

So I'm planning to add some local caching next and support remote CDN in node so there is no difference between browser and node usage when starting Pyodide.

rth:
Merging. We can improve this in follow up PRs.

PR #590, opened 2019-12-31
Mac catalina build

aseering:
Closes https://github.com/iodide-project/pyodide/issues/213

I ran into some trouble building Pyodide on my Mac running OS X Catalina (10.15).  Looks like this is a known issue, tracked by https://github.com/iodide-project/pyodide/issues/572 .  I hacked things into working and put together a PR.

PR #1770, opened 2021-08-01
Changes for Mac compatibility and JS build tweaks

gzuidhof:
Changes so far:
* Use `python3` explicitly everywhere (on Mac one will have `python` and `pip` point towards Python 2 😖 
* Replace `sed` commands to be cross-compatible, see comments. I wish it was prettier.
* Don't rely on npx (I personally see it as dangerous even, if someone pushes a malignant version of typescript we will just run it without noticing), instead we save the tools we need as dependencies.
* CD into `/emsdk` and then source the `emsdk_env.sh` file.. For some reason that does work on Mac.
* Add missing exports to make the Typescript compiler happy.

I'm close to getting everything to build on a Mac. The only thing that seems to be broken still is trying to convince the python package builder to use `glibtool` instead of the Mac version of that, I'm not that familiar with `configure` and how to set a flag to make it do that (tips are welcome!)

I'm trying stuff like `make "CFLAGS='-fPIC'" "LIBTOOL='glibtool -static'"`, but I don't think that makes it through the sh file into the py file into make..


Makefile
```diff
@@ -106,7 +110,7 @@ src/js/pyproxy.gen.js : src/core/pyproxy.* src/core/*.h
 	echo "// This file is generated by applying the C preprocessor to core/pyproxy.js" >> $@
 	echo "// It uses the macros defined in core/pyproxy.c" >> $@
 	echo "// Do not edit it directly!" >> $@
-	cat $< | $(CC) -E -C -P -imacros src/core/pyproxy.c $(MAIN_MODULE_CFLAGS) - >> $@
+	cat src/core/pyproxy.js | $(CC) -E -C -P -imacros src/core/pyproxy.c $(MAIN_MODULE_CFLAGS) - >> $@
```
gzuidhof, line 113: `$<` would end up with `pyproxy.c` being passed into this, don't ask me why.

hoodmane:
Does this superceed #590?

> Use python3 explicitly everywhere

I think it is important to use a virtual environment. We really need Python 3.9.x and on most machines python3 won't point there by default.

gzuidhof:
Ah I wasn't aware of that PR. It's very possible a bunch more tweaks are needed to build the individual packages.. this PR contains some first steps towards cross compilability.

rth (commented):
Generally looks good, however yes better to revert the python3 commands and I'll update the documentation to indicate that a virtualenv with the same python version needs to be used.

Also CI currently fails with `ReferenceError: loadPyodide is not defined` so some change here does break the build on linux.

hoodmane:
It would probably be helpful to install gnu-sed for mac builds:
https://stackoverflow.com/a/27834828/1942152

Since we don't test the mac build, it might be hard for us to ensure that all uses of sed continue to be compatible with mac sed strangeness.

src/core/pyproxy.js
```diff
@@ -274,7 +274,7 @@ Module.callPyObject = function (ptrobj, ...jsargs) {
  * @typedef {(PyProxyClass & {[x : string] : Py2JsResult})} PyProxy
  * @typedef { PyProxy | number | bigint | string | boolean | undefined } Py2JsResult
  */
-class PyProxyClass {
+export class PyProxyClass {
```
hoodmane, line 277: Are these `export` keywords relevant?

gzuidhof: They were for me: the reason being that these get referenced in the types, so a package importing something else from this file might return this type indirectly. Typescript would complain without it

wlach:
> Since we don't test the mac build, it might be hard for us to ensure that all uses of sed continue to be compatible with mac sed strangeness. It seems pretty likely that the CI build failures are due to the sed changes because there aren't that many other changes.

GitHub actions has pretty good support for Mac. It might be worth merging a modified version of #1690 just for that.

rth:
Thanks for this work. Closing as it should now work (we have MacOS in CI). If there are any remaining fixes it would be easier to open a new PR than to resolve conflicts probably.

PR #1773, opened 2021-08-02
Declare name in recursiveDependencies

hoodmane:
We forgot to declare `name` here.


rth (approved):
Thanks! I misread it being already defined above in the same scope. Too many braces..

hoodmane:
We might consider not reusing the same identifier so much, it's a bit confusing.

PR #1774, opened 2021-08-03
Strip definitions of _orig$_Z<very very long> from pyodide.asm.js

hoodmane:
This removes about 52kb from `pyodide.asm.js`.


rth:
Sounds good. Though the issue is that that file is 2MB uncompressed now so 50kB is not going to make much difference.  We can probably investigate how to reduce the size further for 0.19.

rth:
The errors in the CI do seem related to this change (such as `TypeError: Cannot read property 'apply' of undefined` typically for missing symbols)?

Issue #1715, opened 2021-07-14
Discuss desired features/syntax for persistence convenience js/py API


bollwyvl:
As a follow-on to #1596, which exposes the emscripten [FS](https://emscripten.org/docs/api_reference/Filesystem-API.html#persistent-data) directly, more useful wrapper could be added to the persistence API.

Some ideas:

- on the JS side, add additional options to `loadPyodide` to auto-mount a backend
  - the tokens (e.g. `pyodide.FS.filesystems.IDBFS`) would not yet be avaialble, so it might have to be by name
  - it might be unclear how to persist data back out, but a naive polling solution would work
- on the python side, offer a some context managers, etc. that would wrap up the lifecycle of:
  - `mount`, `syncfs(true)`, _do some stuff_, `syncfs(false)`

bollwyvl:
For example, rather than node-style callbacks, this (reverted) commit offered a [promise wrapper](https://github.com/pyodide/pyodide/pull/1596/commits/ddc8421bb55b2c9be794942643bb11b569ba5cce
). This sorta lives on in the [tests](https://github.com/pyodide/pyodide/pull/1596/commits/86f6856cc88951a79101d440275ba7afd548cd7e).

hoodmane:
Promise wrappers would be great.

xhlulu:
I'm not familiar with emscripten, but I saw the tweet about https://github.com/pyodide/pyodide/pull/1596 and I found the elevator pitch and motivation to be quite interesting. Came here to ask: will the new 0.18 release enable an "easy" way to statically served files into the virtual python FS? Currently [I'm using base64 encoding/decoding](https://github.com/xhlulu/pyodide-multi-module-template) (see https://github.com/pyodide/pyodide/issues/419#issuecomment-881760931), but curious if this might be a cleaner alternative.

hoodmane:
Yeah this should be a better alternative. In particular to make a file from some url you can do something like:
```js
let data = await (await fetch(some_url)).text();
pyodide.FS.writeFile(some_path, data);
```

bollwyvl:
I've found that using `.text` can sometimes be troublesome in different browsers... i stole the `arrayBuffer` trick from [micropip](https://github.com/pyodide/pyodide/blob/0.18.0/packages/micropip/src/micropip/micropip.py#L60) for some large-ish stuff.

[...]

Issue #1386, opened 2021-03-29
Docker image for linux/arm64/v8


mdtanrikulu:

rth:
Indeed it could have been nice to investigate if building on arm64/v8 works.

A first step would be to build [the docker image we use](https://github.com/pyodide/pyodide/blob/master/Dockerfile) locally on ARM machine. The [pre-built binaries for browsers](https://github.com/pyodide/pyodide/blob/fd88a18ca373b6edad2be0ceb54e1760e092220e/Dockerfile#L29) would likely need to be installed from different URLs for ARM. 
```
docker build -t iodide/pyodide-env:12 .
```
where the tag needs to match one used in [run_docker](https://github.com/pyodide/pyodide/blob/master/run_docker).

Then [build packages in this docker image](https://pyodide.org/en/latest/development/building-from-sources.html#using-docker).
I suspect at least a few architecture dependent fixes would be necessary (e.g. [here](https://github.com/pyodide/pyodide/blob/fd88a18ca373b6edad2be0ceb54e1760e092220e/packages/numpy/meta.yaml#L29).


Though after all that work, I'm not sure how much faster native arm64 images would be. It would still run [a VM with qemu](https://docs.docker.com/docker-for-mac/apple-m1/) on Mac M1 as far as I understand.

mdtanrikulu:
Oh I should update docker to the rc3 then. Not sure if my version is using qemu, but seems they made it default on newest version. Thanks for the information in general. For future this issue might be beneficial as a reminder about M1 native support if docker also completes it's own experimental virtualization option.

grimmer0125:
https://docs.docker.com/desktop/mac/apple-silicon/ is in GA stage and I already used it for a while and feels good.

thadk:
When trying to use an `arm64`-built version of pyodide-env 20, I had to comment out the official Chrome download in the Dockerfile because you'd have to use Chromium-arm64 instead I think, but I got as far as this on `arm64`. It does seem much snappier. The issue below seems like could be in `emsdk`?: 
```
➜ ./run_docker
thadk@33367d8d87ea:/src$ PYODIDE_PACKAGES="kglab" make
./tools/dependency-check.sh
date +"[%F %T] Building emsdk..."
[2022-07-01 08:05:09] Building emsdk...
make -C emsdk
make[1]: Entering directory '/src/emsdk'
if [ -d emsdk ]; then rm -rf emsdk; fi
[...]

PR #1776, opened 2021-08-03
Local cache for downloads with Node.js

rth:
The general idea is to,
 - expose as private API `pyodide_js._fetch_cached` which would transparently call fetch API in the browser and some fetch like function with local caching in Node.
 - Use this function everywhere we currently use fetch, including in micropip. This in particular avoids having to expose a global `fetch` function in Node.

The main motivation is to make sure we are not repeatedly re-downloading the same packages from micropip and JsDelivr CDN in Node.

TODO:
 - [ ] pick a good fetch library in library with caching. So far I tried,
   - node-fetch: this works fine, however caching is not supported and won't be https://github.com/node-fetch/node-fetch/issues/68
   - [node-fetch-cache](https://github.com/mistval/node-fetch-cache) fairly new, not very popular, maintained -> fails when loading node-fetch as a ES6 module in the dependencies. Probably needs reporting upstream.
   - [make-fetch-happen](https://github.com/npm/make-fetch-happen) -> currently used, I still need to debug some issues.
   - Maybe in the long term it would make sense to [axios](https://github.com/axios/axios) but not in this PR.
  
**Secondary goals:**
 - ideally this should also allow to use,
   ```
   let pyodide = await pyodide_pkg.loadPyodide({
     indexURL: "<remote JsDelivr URL>", 
   })
   ```
   in Node. Technically most code parts should be implemented for it to work, however I'm currently getting an error about `require not defined` in pyodide.asm.js because apparently it considers that it's a ES6 module (https://github.com/nodejs/node/issues/33741#issuecomment-639463213). I'm not sure what's different in that aspect between loading local or remote file, it needs more debugging.

  The mocha test suite can also be run with a remote URL by setting the `PYODIDE_BASE_URL` env variable.

### DIFF
--- a/packages/micropip/src/micropip/micropip.py
+++ b/packages/micropip/src/micropip/micropip.py
@@ -31,7 +31,7 @@ else:
     WHEEL_BASE = Path(".") / "wheels"
 
 if IN_BROWSER:
-    from js import fetch
+    fetch = pyodide_js._fetch_cached
 
 else:
     from urllib.request import urlopen, Request
--- a/src/js/api.js
+++ b/src/js/api.js
@@ -1,5 +1,5 @@
 import { Module } from "./module.js";
-import { loadPackage, loadedPackages } from "./load-pyodide.js";
+import { loadPackage, loadedPackages, fetch_cached } from "./load-pyodide.js";
 import { isPyProxy, PyBuffer } from "./pyproxy.gen.js";
 export { loadPackage, loadedPackages, isPyProxy };
 
@@ -356,6 +356,7 @@ export function makePublicAPI() {
   };
 
   namespace._module = Module; // @private
+  namespace._fetch_cached = fetch_cached; // @private
   Module.public_api = namespace;
   return namespace;
 }
--- a/src/js/load-pyodide.js
+++ b/src/js/load-pyodide.js
@@ -1,7 +1,12 @@
 import { Module } from "./module.js";
 
-const IN_NODE =
+export const IN_NODE =
   typeof process !== "undefined" && process.release.name !== "undefined";
+
+let fetchPromise;
+if (IN_NODE) {
+  fetchPromise = import("node-fetch").then((M) => M.default);
+}
 
 /**
  * Check whether the provided path is a URL or a local path
@@ -18,6 +23,23 @@ export function pathIsUrl(path) {
   } else {
     return false;
   }
+}
+
+/**
+ * Fetch and cache results for the browser and Node.js. Caching is only
+ * explicitly done for Node.js, since browsers will cache fetch requests by
+ * default.
+ * @param {string} input path
+ * @returns {boolean} Path to the cache dir.
+ * @private
+ */
+export async function fetch_cached(url) {
+  let _fetch = typeof fetch == "undefined" ? undefined : fetch;
+
+  if (IN_NODE) {
+    _fetch = await fetchPromise;
+  }
+  return await _fetch(url);
 }
 
 /**
@@ -54,7 +76,7 @@ export async function initializePackageIndex(indexURL) {
     );
     package_json = JSON.parse(package_string);
   } else {
-    let response = await fetch(`${indexURL}packages.json`);
+    let response = await _fetch_cached(`${indexURL}packages.json`);
     package_json = await response.json();
   }
   if (!package_json.packages) {
@@ -104,14 +126,12 @@ if (globalThis.document) {
   };
 } else if (IN_NODE) {
   const pathPromise = import("path").then((M) => M.default);
-  const fetchPromise = import("node-fetch").then((M) => M.default);
   const vmPromise = import("vm").then((M) => M.default);
   loadScript = async (url) => {
     if (pathIsUrl(url)) {
       // If it's a url, have to load it with fetch and then eval it.
-      const fetch = await fetchPromise;
       const vm = await vmPromise;
-      vm.runInThisContext(await (await fetch(url)).text());
+      vm.runInThisContext(await (await fetch_cached(url)).text());
     } else {
       // Otherwise, hopefully it is a relative path we can load from the file
       // system.
--- a/src/js/load-pyodide.js
+++ b/src/js/load-pyodide.js
@@ -5,7 +5,7 @@ export const IN_NODE =
 
 let fetchPromise;
 if (IN_NODE) {
-  fetchPromise = import("node-fetch").then((M) => M.default);
+  fetchPromise = import("node-fetch-cache");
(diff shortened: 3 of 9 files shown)


grimmer0125:
Hi @rth, I just want to submit the same thing in discussions and found out this PR. From my perspective, 
- Pyodide packages: The ideal situation is to install them via npm packages for both browser and Node.js usages. For example, like `@pyodie/core`, `@pyodide/numpy`, if it is doable.  
- Users' Python scripts: as proposed in https://github.com/pyodide/pyodide/discussions/1876, already some actions: #1917 and #1935
- Third-party Python packages on PyPI: It may be a normal Python package or even somebody wants to publish a Python package used in Pyodide environment. For PyPI packages, I think of the same re-downloading issue, it feels it can be improved although browsers sometimes will use built-in cache. Some cases. 
    1.  Browser + no self-hosted Pyodide: I'm not sure if this #1776 is also for browser caching or not and I think some caching mechanism is good, either some JS implementation (e.g. using IndexedDB directly) or via #1715 pyodide File system API. 
    2. Browser + self-hosted Pydoide: It can be a Browser extension or mobile environment. People may adopt the same 1. approach or prefer to bundle everything inside. For the latter case, e.g. my Chrome extension, I manually download some Python wheel package and host it. If there is a tool that is able to automate it, it will be great. 
    3. Use Pyodide in Node.js: Just like this PR, it will be great. Besides HTTP caching, I'm not sure using the 2. approach + some `micropip_load_local_file` is a good idea or not. I don know much about the current design/flow of Pyodide Node.js support.

I saw `require not define` error in my tests in #1949. It is apparent that browsers do not support `CommonJS` require syntax and only support `ES6 import` to load a module. There are three difficulties
- How Pyodide JS import these third-party packages. Some cases:
    1. ES6 import to use http fetch packages, not local file protocol. 
    2. CommonJS `require` (not work in browser)
    3. Bundle them (traditional way) into Pyodide JS. 
- These third-party packages may use other packages, either Node.js built-in packages or other npm packages. Usually, they will not bundle these `node_moduels` so that people (e.g. Pyodide JS) need to install their dependencies.  But this comes to a problem. Some cases:
    1. Pyodide JS does not bundle external dependencies and their dependencies, it can not work directly in browsers since browsers can not natively install npm package (<-CDN way). It may only work if Pyodide users use some bundler (e.g. parcel/webpack) and have a proper setting. 
    2. Pyodide JS bundles these packages and dependencies. In this way, there is no `require not defined` issue. 
    3. *Pyodide only bundle these packages but exclude dependencies <- I think it probably will not happen in any bundler settings.
- Node.js only API does not work in browsers. Need to find out polyfill package/usage.
 
 If adding npm dependency in Pyoide npm package or PyodideJS use some Node.js only API, some of these issues may happen in browsers. I encounter these issues when publishing my @d4c/numjs package. So I just share the thoughts which you may already know these.

rth:
Thanks for your analysis @grimmer0125 !

> Pyodide packages: The ideal situation is to install them via npm

I'll open an issue about it.

> Browser + no self-hosted Pyodide: I'm not sure if this Local cache for downloads with Node.js is also for browser caching or not

I admit I have not tested all browsers, but in my tests `fetch` was fully cached by the browser (or at least it respects the `Cache-Control` headers in the response it gets). So this means that micropip downloads and Pyodide packages should be cached. I'm not sure what happens in the browser extensions, worth checking.

> Browser + self-hosted Pydoide

The above should apply, provided that you serve files with appropriate `Cache-Control` headers. For instance, in Pyodide deployed packages, it's done [here](https://github.com/pyodide/pyodide/blob/main/.circleci/config.yml#L290) (`--cache-control 'max-age=30758400, immutable, public'`)

Thanks for the rest of your comment. I have to think about it. I'm not really using JS much in my everyday work, and your analysis is definitely appreciated. Also, if you are interested in continuing this PR let me know (otherwise I'll get to it eventually).

grimmer0125:
> The above should apply, provided that you serve files with appropriate `Cache-Control` headers. For instance, in Pyodide deployed packages, it's done [here](https://github.com/pyodide/pyodide/blob/main/.circleci/config.yml#L290) (`--cache-control 'max-age=30758400, immutable, public'`)

Originally I guess `an error about require not defined in pyodide.asm.js` is talking about the fetch_cache npm package you tried not working in the browser. So this PR is also for the browser and already got initial working in the browser. The error about `pyodide.asm.js` is a different thing? 

> Thanks for the rest of your comment. I have to think about it. I'm not really using JS much in my everyday work, and your analysis is definitely appreciated. Also, if you are interested in continuing this PR let me know (otherwise I'll get to it eventually).

I would like to help but I can not guarantee how much time I can work on it. I start https://github.com/pyodide/pyodide-webpack-example/pull/4 and I think I can finish that first? Or if you think this local cache has a higher priority and we can discuss. 

BTW, it seems that this PR does not need a local Pyodide building environment (either building core or using emcc  preprocess to generate `pyproxy.gen.js`), isn't it? I can not build Pyodide on my mac M1.

rth:
> I start pyodide/pyodide-webpack-example#4 and I think I can finish that first?

Yes, let's finish that first.

> BTW, it seems that this PR does not need a local Pyodide building environment (either building core or using emcc preprocess to generate pyproxy.gen.js), isn't it?

Likely not, but you might still need to re-build the pyproxy.gen.js, the first time, not sure.

>  I can not build Pyodide on my mac M1.

We need to improve the situation there. I'm a bit stuck on https://github.com/pyodide/pyodide/pull/1912 right now, but it should be fixable. Once that works one should be able to get the x86_64 conda for Mac, and rely on x86_64 emulation on M1 which is pretty good. I guess one could do that with homebrew as well, but because it doesn't have virtual env for system packages, I'm less certain about recommending that as a general solution.

grimmer0125:
> homebrew as well, but because it doesn't have virtual env for system packages
Is the virtual env referring to Python virtual environment? I do not use `conda` and usually, I use homebrew to install some non-python packages and `pyenv` for global interpreters. Then use poetry which uses Python built-in `venv`. This is my habit and it is acceptable to use conda for Pyodide development for me. Either #1386 (arm64 docker) or native build on Mac (x86 or emulation) should be good enough.

ryanking13:
Just for an update: Node.js is started to [experimentally support fetch API](https://github.com/nodejs/node/commit/6ec225392675c92b102d3caad02ee3a157c9d1b7), though it looks like there is no local cache support (yet?).

rth:
This is actually already implemented. When loading packages with the node module, they will be put to node_modules/pyodide/ the first time they are downloaded, then reused. Closing.

PR #1777, opened 2021-08-04
DOC Post-release documentation improvements

rth:
A few documentation improvements,
 - add the list of the contributors and a date for the 0.18.0 release
 - add instructions on how to publish pyodide-build to PyPi and src/js to npm.
 

cc @hoodmane


rth:
Also it might be better to avoid merging master into stable. Since we cherry pick commits between master and stable/release branches, merging them includes those commits twice making it really hard to know what's in the release branch. So I think a better approach is the one outlined in https://pyodide.org/en/stable/development/maintainers.html create a release branch e.g. 0.17.X from stable (for archival), then delete stable locally, re-create it from the 0.18.0 tag and force push. That's what I did for this release.

Issue #1720, opened 2021-07-17
time.sleep is not respected


leopsidom:
Not sure if this is a known issue, but I tried to run the following code:

```
 pyodide.runPython(`
import time
for i in range(3):
    time.sleep(20)
    print(i)
  `)
```

It prints immediately without actually sleeping for 20 seconds between each print

hoodmane:
This is a special case of #1503.
AFIAK there are two ways to implement synchronous I/O: `Atomics.wait` or evil `unthrow` stack rewinding magic. There is a third option for sleeping which is to busy loop.

If you are running Pyodide on the main thread then you aren't allowed to block at all so unthrow is the only option (or busy loop and hang the thread).

Off the main thread you can sleep with `Atomics.wait`. But here we run into trouble with our nemesis Safari.
https://caniuse.com/?search=Atomics

So we are very interested in fixing this but it is complicated.

anyoung-mc:
Is this still being blocked by Safari? Based on the link you provided, It looks like Safari now supports Atomics.

hoodmane:
I think Safari 15.2 released in December 2021 added atomics support yes.

ryanking13:
time.sleep now works as expected (Emscripten implemented it using busy wait)

PR #1780, opened 2021-08-06
Fix the repr of lists and dicts in console.html

### LINKED ISSUE #1779: dictionary and list display improvement in pyodide console [0.18.0]
Dictionary / list doesn't seem to display optimally in pyodide console. Current behavior in REPL: https://pyodide.org/en/stable/console.html:

```
>>>{'a': 1}
[object Map]

>>>[1,2,3]
1,2,3
```

while in a normal terminal, this would display:

```
>>> {'a': 1}
{'a': 1}

>>> [1,2,3]
[1, 2, 3]
```

hoodmane:
Resolves #1779.

### DIFF
--- a/src/templates/console.html
+++ b/src/templates/console.html
@@ -37,7 +37,7 @@
             BANNER = "Welcome to the Pyodide terminal emulator 🐍\\n" + BANNER
             js.pyconsole = PyodideConsole(__main__.__dict__)
             async def await_fut(fut):
-              return to_js([await fut]);
+              return to_js([await fut], depth=1)
         `,
           namespace
         );
--- a/src/tests/test_console.py
+++ b/src/tests/test_console.py
@@ -340,19 +340,30 @@ def test_console_html(console_html_fixture):
         return get_result()
 
     welcome_msg = "Welcome to the Pyodide terminal emulator 🐍"
-    assert selenium.run_js("return term.get_output()")[:len(welcome_msg)] == welcome_msg
+    assert (
+        selenium.run_js("return term.get_output()")[: len(welcome_msg)] == welcome_msg
+    )
 
     assert exec_and_get_result("1+1") == ">>> 1+1\n2"
     assert exec_and_get_result("1 +1") == ">>> 1 +1\n2"
     assert exec_and_get_result("1+ 1") == ">>> 1+ 1\n2"
     assert exec_and_get_result("[1,2,3]") == ">>> &#91;1,2,3&#93;\n[1, 2, 3]"
-    assert exec_and_get_result("{'a' : 1, 'b' : 2, 'c' : 3}") == ">>> {'a' : 1, 'b' : 2, 'c' : 3}\n{'a': 1, 'b': 2, 'c': 3}"
-    assert exec_and_get_result("{'a': {'b': 1}}") == ">>> {'a': {'b': 1}}\n{'a': {'b': 1}}"
-    assert exec_and_get_result("[x*x+1 for x in range(5)]") == '>>> &#91;x*x+1 for x in range(5)&#93;\n[1, 2, 5, 10, 17]'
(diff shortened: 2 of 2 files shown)


src/templates/console.html
```diff
@@ -37,7 +37,7 @@
             BANNER = "Welcome to the Pyodide terminal emulator 🐍\\n" + BANNER
             js.pyconsole = PyodideConsole(__main__.__dict__)
             async def await_fut(fut):
-              return to_js([await fut]);
+              return to_js([await fut], depth=1)
```
leopsidom, line 40: Will this also convert `{'a': {'b': 1}}` correctly ? Also I'm wondering what is the significance of this method (await_fut) ? Is there a way to make it so client doesn't have to rely on this ?

hoodmane: Yeah it converts that correctly. I just want to convert the outermost list wrapper to Javascript so that we can directly de-structure it. Setting `depth=1` does a shallow conversion, so it makes the outermost list into a length 1 list containing a PyProxy with the result of `await fut`.

The point of this `await_fut` method is as follows: suppose that `f` is an async function. We wish to distinguish between `await f()` and `f()`. So e.g., after `async def f(): return 7`, `f()` should return `<coroutine object f at 0x??????>` and `await f()` should return 7.

I think the issue is that if we await an awaitable from Javascript and the result is also awaitable, then it will await the result too. Thus, there is no way in Javascript for the result of one Promise to be a second distinct Promise without collapsing the inner promise into the outer promise. If we allowed this to happen, then `f()` would return `7` but e.g., `x=f()` would set `x` to a coroutine. This is weird. If `f()` returns `7` then `x=f()` should result in `x` equal to `7`.

So to fix this, I have the hack of putting the result into a one entry list. This list is surely not awaitable. Then we destructure it into Javascript and can so avoid accidentally folding it into the outer promise.

Not sure if there is a better solution.

hoodmane: At the very least we should have a comment explaining this in the source code.

leopsidom: I'm not very familiar with coroutine related concepts in python. Is it possible to wait for `await f()` in python and return the result to Javascript instead of having to await in javascript ? Or maybe this is related to the `time.sleep` issue here: https://github.com/pyodide/pyodide/issues/1720

hoodmane: Asynchronous Python calls are scheduled on the browser event loop so ultimately it is necessary to `await` any asynchronous Python code from Javascript if you care about the result. (The function `runPythonAsync` returns a normal promise.) So yeah this is just a small difference between Python and Javascript: In Python:
```py
async def f():
   # do stuff
   return 7

async def g():
   return f()
c = await g() # c is now a coroutine
```
In Javascript:
```js
async function f(){
   // do stuff
   return 7;
}

async function g() {
   return f();
}
let c = await g(); // c is now 7
```
Python's behavior here is slightly better suited to our needs...

leopsidom: _ultimately it is necessary to await any asynchronous Python code from Javascript_. Hmm, how about `x = await g()` ? It doesn't seem like it's awaited in javascript or it is at the backend ? I'm just wondering if `x = await g()` can be evaluated in python VM completely we might also be able to do the same for `await g()` ?

hoodmane: Well if you run `x = await g()` or anything with top level await, you have to run it with `runPythonAsync`. So it would look like `await pyodide.runPythonAsync("x = await g()")`. Stated in yet another way, the problem is that 
```
await pyodide.runPythonAsync("xpy = g()");
``` 
results in `xpy` equal to a coroutine whereas 
```
let xjs = await pyodide.runPythonAsync("g()")
```
results in `xjs` equal to `7` because the promise returned by `runPythonAsync` merges in the coroutine `g()` and so awaiting it resolves both.

leopsidom: Okay got it. That makes sense. So it's because when python returns a coroutine, it's converted to Promise which will get automatically awaited since `runPythonAsync` result is always `awaited`.

rth (approved):
Thanks!

rth:
There is still a failure in `test_console_html` ..

Issue #982, opened 2020-12-31
Specify supported browsers


dalcde:
Having a clear list of supported browser gives a concrete policy for what javascript/webassembly features we can use.

hoodmane:
The de facto state of things is we only support the browsers that we run selenium on.

rth:
There is some hope for WebKit based browsers as well (https://github.com/iodide-project/pyodide/issues/989), for instance, Safari. I think package loading was working fine there before https://github.com/iodide-project/pyodide/pull/600

PR #1789, opened 2021-08-12
DOC Document minimal supported browser versions

rth:
Closes https://github.com/pyodide/pyodide/issues/982

Minimal supported browser versions are determined by bisection on BrowserStack. 

For Edge, versions earlier than 80 probably work but they are not available there so I couldn't test them.

### DIFF
--- a/docs/usage/index.md
+++ b/docs/usage/index.md
@@ -42,6 +42,31 @@ To avoid confusion, note that:
  - `cdn.jsdelivr.net/npm/pyodide@0.18.0/` is a mirror of the Pyodide NPM package, which includes none of the WASM files
 ```
 
+### Supported browsers
+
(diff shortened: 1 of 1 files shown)


docs/usage/index.md
```diff
@@ -42,6 +42,33 @@ To avoid confusion, note that:
  - `cdn.jsdelivr.net/npm/pyodide@0.18.0/` is a mirror of the Pyodide NPM package, which includes none of the WASM files
 ```
 
+### Supported browsers
+
+Pyodide works in any modern web browser with WebAssembly support.
+
+**Tier 1** browsers are tested as part of the test suite with continuous integration,
```
rth, line 49: The wording is largely inspired by https://doc.rust-lang.org/nightly/rustc/platform-support.html

PR #1790, opened 2021-08-12
Fix terminal Ctrl+C handling

### LINKED ISSUE #1787: Ctrl+C support in REPL
Right now if one enters a statement that is syntactically invalid, for instance,
```py
def func():
    """A function""
    return
``` 
with unbalanced quotes, the REPL will continue asking for new input on Enter, and assumes that the statement is incomplete. Except that there is no way to fix that statement since we currently don't allow editing previous lines. I have not found a way to interrupt it short of restarting the REPL.

Adding support for Ctrl+C to stop asking for new REPL input would be good.

In the standard Python REPL,
```
>>> a # + Ctrl+C
KeyboardInterrupt
```

hoodmane:
This resolves #1787 and should bring repl behavior a bit closer to what happens in native console. I also made the text a bit whiter because I think the text contrast is slightly low.

### DIFF
--- a/docs/project/changelog.md
+++ b/docs/project/changelog.md
@@ -11,6 +11,15 @@ substitutions:
 
 # Change Log
 
+## Unreleased
+
+### Console
+
+- {{Fix}} Ctrl+C handling in console now works correctly with multiline input.
+  New behavior more closely approximates the behavior of the native Python
+  console.
+  {pr}`1790`
+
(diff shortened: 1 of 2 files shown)


rth (approved):
Great, thanks!

rth:
Could you add a changelog entry (under a similar subsection to the one used in 0.18 release)? Thanks!

PR #1794, opened 2021-08-14
DOCS Fix typo in webworker example

etnnth:
When copy pasting the code in the examples the results where always null because of a typo in the webworker.js example.


hoodmane:
Thanks!

PR #1797, opened 2021-08-15
Fix exposure of setInterruptBuffer

### LINKED ISSUE #1795: setInterruptBuffer is undefined [0.18.0]
The `setInterruptBuffer` API seems to be undefined in 0.18.0.

Reproduce step:
1. go to REPL: https://pyodide.org/en/stable/console.html
2. open browser console, and type `pyodide.setInterruptBuffer`

It returns undefined.

The API is available on 0.17.0: https://pyodide.org/en/0.17.0/console.html

```
pyodide.setInterruptBuffer
ƒ (buffer){Module.interrupt_buffer=buffer;if(buffer){_PyPyodide_SetPyodideCallback($0)}else{_PyPyodide_SetPyodideCallback(0)}}
```

hoodmane:
During the rollup rearrangement we messed up the exposure of `setInterruptBuffer` because of initialization order issues. This fixes it. Also, I updated the test to use the public `setInterruptBuffer` entrypoint so we'd see a test failure if a similar problem arises in the future. Resolves #1795.

### DIFF
--- a/src/js/api.js
+++ b/src/js/api.js
@@ -312,11 +312,7 @@ Module.restoreState = (state) => Module.pyodide_py._state.restore_state(state);
  */
 export function setInterruptBuffer(interrupt_buffer) {
   Module.interrupt_buffer = interrupt_buffer;
-  if (interrupt_buffer) {
-    Module._PyPyodide_SetPyodideCallback(Module._pyodide_callback);
-  } else {
-    Module._PyPyodide_SetPyodideCallback(0);
-  }
+  Module._set_pyodide_callback(!!interrupt_buffer);
 }
 
 export function makePublicAPI() {
--- a/src/js/api.js
+++ b/src/js/api.js
@@ -310,9 +310,14 @@ Module.restoreState = (state) => Module.pyodide_py._state.restore_state(state);
 /**
  * @param {TypedArray} interrupt_buffer
  */
-function setInterruptBuffer(interrupt_buffer) {}
-setInterruptBuffer = Module.setInterruptBuffer;
-export { setInterruptBuffer };
+export function setInterruptBuffer(interrupt_buffer) {
+  Module.interrupt_buffer = interrupt_buffer;
+  if (interrupt_buffer) {
+    Module._PyPyodide_SetPyodideCallback(Module._pyodide_callback);
+  } else {
+    Module._PyPyodide_SetPyodideCallback(0);
+  }
+}
 
 export function makePublicAPI() {
   /**
@@ -333,7 +338,6 @@ export function makePublicAPI() {
    * @type {FS} The Emscripten File System API.
    */
   const FS = Module.FS;
-
   let namespace = {
     globals,
     FS,
--- a/src/tests/test_pyodide.py
+++ b/src/tests/test_pyodide.py
@@ -239,9 +239,7 @@ def test_keyboard_interrupt(selenium):
         try {
             pyodide.runPython(`
                 from js import triggerKeyboardInterrupt
-                x = 0
-                while True:
-                    x += 1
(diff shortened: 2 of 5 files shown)


src/js/api.js
```diff
@@ -310,9 +310,14 @@ Module.restoreState = (state) => Module.pyodide_py._state.restore_state(state);
 /**
  * @param {TypedArray} interrupt_buffer
  */
-function setInterruptBuffer(interrupt_buffer) {}
-setInterruptBuffer = Module.setInterruptBuffer;
-export { setInterruptBuffer };
+export function setInterruptBuffer(interrupt_buffer) {
```
leopsidom, line 313: Would there be a difference moving the function from `core` to `api` level ?

src/tests/test_pyodide.py
```diff
@@ -246,7 +246,8 @@ def test_keyboard_interrupt(selenium):
                         triggerKeyboardInterrupt()
             `)
         } catch(e){}
-        return pyodide.runPython('x')
+        pyodide.setInterruptBuffer(undefined);
```
leopsidom, line 249: Is it necessary to set the buffer back to `undefined` ?

src/js/api.js
hoodmane: It means that the function can't be called as easily from core code, but it wasn't being called from other core code anyways. The main reason to move it is that it makes it more consistent with the other public APIs. It also is simpler for the api documentation system, for the type definitions, etc. When we split `pyodide.js` into multiple files with rollup, we moved all the other public apis into `api.js` but for whatever reason we left `setInterruptBuffer` defined in this really awkward way. The main downside of this move is that it moves the logic further away from the function definitions in `keyboard_interrupt.c`.

src/tests/test_pyodide.py
hoodmane: I think it's a good idea because `setInterruptBuffer` changes global state so if we don't do this and having an interrupt buffer changes the behavior of other tests it could cause confusion.

src/js/api.js
hoodmane: Then again, tests aren't passing so maybe I am not understanding something correctly.

hoodmane: Okay, taking function pointers from Javascript seems to be very error prone.

rth (commented):
It looks like Chrome tests timeout, possibly as a result of this PR? I restarted them once.

src/js/api.js
hoodmane: The way Python handles sigint is that the OS level interrupt handler sets a flag saying an interrupt was received. Nothing else can be done right away because it is unclear what is going on. The flag is a special interpreter variable that is not guarded by the GIL, unlike the exception state. The next time `PyErr_CheckSignals` is called by a C extension (which requires the GIL to be held) or at the beginning of every nth interpreter opcode (excluding opcodes that are FAST_DISPATCH'd and a few others), the interpreter will take note that this flag is set and run the interrupt handler that is set. This handler usually sets a `KeyboardInterrupt` but using the `signal` module, you can set the handler to be any Python function.

After the `KeyboardInterrupt` is set, the Python stack is unwound in the usual way for when an exception is set.

The `PyErr_SetInterrupt` just sets the flag to indicate that an interrupt was received. The interrupt handler will be run as normal when Python gets around to it. Emscripten has no OS level signal handlers so we use this to artificially trigger the usual process of handling interrupts. Note however that `PyErr_CheckInterrupt` won't work as expected, so currently keyboard interrupts won't work inside of C modules (see #1294).

I don't understand why repeated keyboard interrupts would cause a stack overflow, please open an issue with info about this.

PR #1799, opened 2021-08-18
ENH: add hdf5 library

bmaranville:
I haven't been able to build h5py, but I was able to build the hdf5 library to wasm with hints from https://github.com/attilaolah/wasm and https://github.com/aertslab/webhdf5

The resulting library files are build/hdf5-hdf5-1_12_1/build/hdf5/lib/libhdf5.a and build/hdf5-hdf5-1_12_1/build/hdf5/lib/libhdf5_hl.a, and the includes are in build/hdf5-hdf5-1_12_1/build/hdf5/include


ax3l:
That's cool - I would use that if it were finalized & merged :)

hoodmane:
Well it looks like this works, I'm not sure why no one ever reviewed this. We should merge main and add a change log entry and if the build still works we can merge it.

ax3l:
Sounds great, @bmaranville can you update or can someone push on this PR? :)

rth:
I updated with main, and added a changelog entry. 

Are we sure this is working through: it builds, but someone should probably try to use it to confirm?

hoodmane:
Yeah now that you point it out @rth this should probably just wait for h5py. Unless this library is also used by other Python packages?

ax3l:
I would use it for #2169 :)

hoodmane:
@ax3l Why don't you merge this on #2169 and then we can test hdf5 as part of that PR? When you get that PR working, we can merge this one first.

packages/libhdf5/meta.yaml
```diff
@@ -0,0 +1,32 @@
...
+source:
+  sha256: e6dde173c2d243551922d23a0387a79961205b018502e6a742acb30b61bc2d5f
+  url: https://github.com/HDFGroup/hdf5/archive/refs/tags/hdf5-1_12_1.tar.gz
+
+requirements:
+  run:
+    - zlib
+
+build:
+  library: true
```
ax3l, line 14: Not sure what that means.

This is a C++ library with Python package. The primary need here is to use the Python package.

packages/libhdf5/meta.yaml
```diff
@@ -0,0 +1,32 @@
...
+  script: |
+    mkdir -p build;
+    cd build \
+        && LDFLAGS="-s NODERAWFS=1" emcmake cmake ../ \
+        -DCMAKE_CROSSCOMPILING_EMULATOR="$PYODIDE_ROOT/emsdk/emsdk/node/14.18.2_64bit/bin/node" \
```
ax3l, line 19: Would it be easier to maintain if we wrote this in a version agnostic way? :)
Maybe with `$(which node)` or a variable? :)

ax3l:
We can also merge a `h5py` package on this branch.
I think it should read like this:
```yaml
package:
  name: h5py
  version: 3.6.0

source:
  url: https://github.com/h5py/h5py/archive/3.6.0.tar.gz
  sha256: 840fca0787bca8ad2729b9c20e06a16f219e7a24ccc59fdf5177a05ec8472b2

requirements:
  run:
    - numpy
    - libhdf5

test:
  imports:
    - h5py

build:
  script: |
    export HDF5_MPI=OFF
    export H5PY_SETUP_REQUIRES="0"

    pip install . --no-deps --ignore-installed --no-cache-dir -vv
```

packages/libhdf5/meta.yaml
bmaranville: Of course - that would be great!

hoodmane: `$(which node)` isn't necessarily a good idea though because you want to use the emscripten node, and I'm concerned that it would find a different node in a lot of cases.

hoodmane:
Weird, it makes an archive which should really be called something like `libopenpmd.a` but it calls it a `.so` file.

packages/libhdf5/meta.yaml
ax3l: If email/ only contains one version at a time, we could also rely on a bash expansion and use * for it

ax3l:
For `h5py`, it might be easier if we build HDF5 as shared library, instead of a static one (see above). This would also reduce code duplication of the static lib in futher packages.

packages/libhdf5/meta.yaml
```diff
@@ -0,0 +1,32 @@
...
+        -DCMAKE_INSTALL_PREFIX=hdf5 \
+        -DH5_HAVE_GETPWUID=0 \
+        -DH5_HAVE_SIGNAL=0 \
+        -DBUILD_SHARED_LIBS=0 \
+        -DBUILD_STATIC_LIBS=1 \
```
ax3l, line 24: Let's give this a try?
```suggestion
        -DBUILD_SHARED_LIBS=1 \
        -DBUILD_STATIC_LIBS=0 \
```

rth:
@ax3l Feel free to add h5py, on top in a separate MR. Then if that's working we could merge this, then the MR with h5py.

> (But I am not very familiar with the details of shared libraries in webassembly yet

It should work, we are doing that for CLAPACK (loaded from scipy) for instance.

> wasm-ld: error: unknown argument: --enable-new-dtags

You would need to skip that argument in https://github.com/pyodide/pyodide/blob/main/pyodide-build/pyodide_build/pywasmcross.py#L251

ax3l:
Started in #2180 :)

From a local build, I can see that `libhdf5` is not yet building `.so` libs, but still builds statically.

I think `libhdf5` will need the same injection as I do in #2169:
https://github.com/emscripten-core/emscripten/issues/15276#issuecomment-1039349267

```bash
# CLI
... -DCMAKE_PROJECT_INCLUDE=overwriteProp.cmake
```
```cmake
# overwriteProp.cmake
set_property(GLOBAL PROPERTY TARGET_SUPPORTS_SHARED_LIBS TRUE)
set(CMAKE_SHARED_LIBRARY_CREATE_C_FLAGS "-s SIDE_MODULE=1")
set(CMAKE_SHARED_LIBRARY_CREATE_CXX_FLAGS "-s SIDE_MODULE=1")
set(CMAKE_STRIP FALSE)  # used by default in pybind11 on .so modules
```

rth:
Resolved in https://github.com/pyodide/pyodide/pull/2411

mobileskyfi:
Is this working still and/or did it make it into a packaged/release build?  

Regardless if I use an `import h5py` or `micropip.install('h5py'), I get an error that it's not found.  This happens when with REPL emulator from the GH project page:
```
Python 3.10.2 (main, Sep 15 2022 23:28:12) on WebAssembly/Emscripten
>>> import h5py
ModuleNotFoundError: No module named 'h5py'
>>> import micropip
>>> await micropip.install("h5py")
  File "/lib/python3.10/site-packages/micropip/_micropip.py", line 548, in install
    await transaction.gather_requirements(requirements)
  File "/lib/python3.10/site-packages/micropip/_micropip.py", line 305, in gather_requirements
  File "/lib/python3.10/asyncio/futures.py", line 284, in __await__
    yield self  # This tells Task to wait for completion.
  File "/lib/python3.10/asyncio/tasks.py", line 304, in __wakeup
    future.result()
    raise self._exception
    result = coro.send(None)
  File "/lib/python3.10/site-packages/micropip/_micropip.py", line 312, in add_requirement
    return await self.add_requirement_inner(Requirement(req))
  File "/lib/python3.10/site-packages/micropip/_micropip.py", line 407, in add_requirement_inner
    wheel = find_wheel(metadata, req)
  File "/lib/python3.10/site-packages/micropip/_micropip.py", line 275, in find_wheel
    raise ValueError(
ValueError: Can't find a pure Python 3 wheel for 'h5py'.
See: https://pyodide.org/en/stable/usage/faq.html#micropip-can-t-find-a-pure-python-wheel
You can use `micropip.install(..., keep_going=True)`to get a list of all packages with missing wheels.
>>> await micropip.install("h5py", keep_going=True)
  File "/lib/python3.10/site-packages/micropip/_micropip.py", line 552, in install
    raise ValueError(
ValueError: Can't find a pure Python 3 wheel for: 'h5py'
See: https://pyodide.org/en/stable/usage/faq.html#micropip-can-t-find-a-pure-python-wheel

```

ax3l:
I am on a vacation the next two weeks - happy if anyone wants to take it over and merge it.

Axel

ryanking13:
@mobileskyfi We disabled `h5py` for a while due to Emscripten issues, but it is enabled now. You can use it in our dev (main) branch or in dev console (https://pyodide.org/en/latest/console.html).

mobileskyfi:
@ryanking13 - yup that worked, both in REPL and in observablehq. Thanks!

_I should have thought to try the [dev](https://pyodide.org/en/stable/usage/downloading-and-deploying.html) .wasm_

Issue #1295, opened 2021-03-02
Build only micropip with default PYODIDE_PACKAGES


rth:
Currently when a user or contributor downloads pyodide and runs make they end up with the full build which can take a while and with packages they might not need.  This situation is going to get worse as more packages are added.

I propose to,
 -  by default set `PYODIDE_PACKAGES='micropip'` so only it is built, forcing users to explicitly specify packages they need in addition.
 - add an option `PYODIDE_PACKAGES='*'` to build everything.

WDYT?

phorward:
@rth thanks for this proposal, I totally agree with you. For most cases I'm using `make minimal` as build target.

Generally, a fellow and I are currently focusing on building a *nano*-build command for Pyodide to end up with the tiniest working Pyodide which is possible (without support for any native-code packages) but with the full JavaScript integration. We're currently at 4,4 MB for a version which is suitable for our needs. This also heavily relates to #646, and I will update there when we have a fine and solid solution.

rth:
> We're currently at 4,4 MB for a version which is suitable for our needs. We're currently at 4,4 MB for a version which is suitable for our needs.

Great, looking forward to it :)

hoodmane:
Once in a while when I forget to `export PYODIDE_PACKAGES="micropip"` or `make minimal` or `make debug` I come back and it's building all the packages and I have to quit and `make minimal` to get it to output `packages.json`. Would be nice to have `make minimal` by default.

I think instead of just micropip it might be a good idea to by default include the other packages used in the core test suite: pyparsing, pytz, packaging, and kiwisolver.

rth:
I agree that the core test suite should pass with the default build. Another way to get there though is to reduce number of packages it needs, by removing the dependency on at least one or two of the current requirements. As far as I remember the only reason we need so many is to load packages that depend on one another. Maybe we could mock or do isolated unit tests for some of those.

rth:
>  Maybe we could mock or do isolated unit tests for some of those.

Just looked into it and it doesn't look simple.

So maybe we should go with that full list as the default.

dmondev:
Would it make sense to even split the repo, moving the packages outside of the core project, similar to what projects with plugins/extensions do ?

This could be done as well as separate npm packages, with their only lifecycle, and potentially maintainers.

```
@pyodide/numpy
@pyodide/pandas
@someotherorg/my-super-duper-package
```

rth:
[...]

PR #1804, opened 2021-08-25
ENH Change to minimal build by default

rth:
With this PR `make` only builds the packages needed for the core test suite (previously equivalent to `make minimal`), 
 - `make minimal` is removed as redundant
 - To build all packages `PYODIDE_PACKAGES='*'` can be used.
 - This supports meta packages `core` and `min-scipy-stack` (cf PR docs for a description)

The end goal in a follow up PR is to switch to only building packages in `min-scipy-stack`, instead of all, to make CI faster.

Closes https://github.com/pyodide/pyodide/issues/1295

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -64,8 +64,7 @@ jobs:
           no_output_timeout: 1200
           command: |
             ccache -z
-            # The following packages are currently used in the main pyodide test suite
-            PYODIDE_PACKAGES="micropip,pyparsing,pytz,packaging,Jinja2" make
+            PYODIDE_PACKAGES="core" make
             ccache -s
 
       - run:
@@ -108,7 +107,7 @@ jobs:
           no_output_timeout: 1800
           command: |
             ccache -z
-            make -C packages
+            PYODIDE_PACKAGES='*' make -C packages
             ccache -s
 
       - run:
--- a/packages/test_packages_common.py
+++ b/packages/test_packages_common.py
@@ -22,7 +22,10 @@ def registered_packages() -> List[str]:
 
 @functools.cache
 def built_packages() -> List[str]:
-    """Returns a list of built package names"""
+    """Returns a list of built package names.
+
+    This functions lists the names of the .data files in the build/ directory.
+    """
     if not BUILD_DIR.exists():
         return []
     registered_packages_ = registered_packages()
--- a/packages/test_packages_common.py
+++ b/packages/test_packages_common.py
@@ -1,12 +1,16 @@
 import pytest
 import os
 from pathlib import Path
+import functools
+
 from pyodide_build.common import _parse_package_subset
 from pyodide_build.io import parse_package_config
 
 PKG_DIR = Path(__file__).parent
+BUILD_DIR = PKG_DIR.parent / "build"
 
 
+@functools.cache
 def registered_packages():
     """Returns a list of registered package names"""
     packages = []
--- a/packages/test_packages_common.py
+++ b/packages/test_packages_common.py
@@ -1,9 +1,9 @@
 import pytest
 import os
 from pathlib import Path
+from typing import List
 import functools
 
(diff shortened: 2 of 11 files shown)


packages/test_packages_common.py
```diff
@@ -59,9 +79,7 @@ def test_import(name, selenium_standalone):
             )
         )
 
-    built_packages = _parse_package_subset(os.environ.get("PYODIDE_PACKAGES"))
-    # only a subset of packages were built
-    if built_packages is not None and name not in built_packages:
+    if name not in built_packages():
```
rth, line 82: It's more robust to actually check the artifacts present on disk instead of relying on the env variable, as we often pass it only to the make script and not to pytest.

.circleci/config.yml
```diff
@@ -64,8 +64,7 @@ jobs:
           no_output_timeout: 1200
           command: |
             ccache -z
-            # The following packages are currently used in the main pyodide test suite
-            PYODIDE_PACKAGES="micropip,pyparsing,pytz,packaging,Jinja2" make
+            PYODIDE_PACKAGES="core" make
```
rth, line 67: This is equivalent to just `make`. Added the env variable here for clarity.

hoodmane (approved):

pyodide-build/pyodide_build/common.py
```diff
@@ -12,13 +12,46 @@ def _parse_package_subset(query: Optional[str]) -> Optional[Set[str]]:
...
+       if query is None.
+     - 'min-scipy-stack': includes the "core" meta-package as well as some of the
+       core packages from the scientific python stack and their dependencies:
+       {"numpy", "scipy", "pandas", "matplotlib", "scikit-learn", "joblib", "pytest"}.
+       This option is non exaustive and is mainly intended to make build faster
+       while testing a diverse set of scientific packages.
+     - '*': corresponds to all packages (returns None)
+
     Returns:
       a set of package names to build or None.
```
hoodmane, line 27: Would be good to say separately that `None` as an argument results from `PYODIDE_PACKAGES` being unset and leads to only `core` packages being built, and that `None` as a return value means all packages.

It might be better in the * case to return the set of all packages, which would eliminate this special case.

rth: > It might be better in the * case to return the set of all packages, which would eliminate this special case.

This function is only responsible for parsing CLI string input, not  loading the list of available packages from disk. So we don't have that list at this point. In the case of  all packages (`*`) we need some special marker. 

FWIW before this PR the output of None also meant to build all packages, but I agree that the fact that,
 - before: "None" input -> "None": output
 - now: "None: input -> 'core' meta package, "*" input -> "None" output

might be a bit confusing. I'll improve the docs.

packages/test_packages_common.py
```diff
@@ -16,6 +20,22 @@ def registered_packages():
     return packages
 
 
+@functools.cache
+def built_packages() -> List[str]:
+    """Returns a list of built package names"""
```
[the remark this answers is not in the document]
rth: Comment added. It's not consistent but also not exactly the same thing,
 1. when building we need to know if build artifacts in packages/*/ folders were created (with `.packaged`) and so the package doesn't need to be rebuilt
 2. here we want to know if the final build artifacts exist

For instance, it could happen that the build/ folder was removed (meaning tests cannot be run) but the `pacakges/**/.packaged` are still there and packages don't need to be rebuilt from 0.

rth:
Hmm, I still don't understand why Chrome consistently times out in this PR after I synced with master (while CI was green before)

hoodmane:
> Chrome consistently times out in this PR

Chrome is timing out a lot in general of late, I think it is probably unrelated to this PR?

rth:
Most likely, but it's hard to say anything for sure with our current CI.

PR #1805, opened 2021-08-25
BLD Use outer ccache wrapper

rth:
This should be a better way to use ccache with emsdk according to https://github.com/emscripten-core/emscripten/pull/13498

Investigating into whether this would fix https://github.com/pyodide/pyodide/issues/1007


rth:
The status on this is that ccache is correctly being called now during compilation. However for some reason it doesn't recognize the files to compile as being identical between different CI runs and so we consistently get cache misses. Need to investigate that next.

.circleci/config.yml
```diff
@@ -8,6 +8,9 @@ defaults: &defaults
     - EMSDK_NUM_CORES: 3
       EMCC_CORES: 3
       PYODIDE_JOBS: 3
+      # Make sure the ccache dir is consistent between core and package builds
+      # (it's not the case otherwise)
```
rth, line 12: Maybe something to do with emsdk activate, by it in any case we were previously caching the wrong path for packages

.circleci/config.yml
```diff
@@ -73,8 +88,8 @@ jobs:
 
       - save_cache:
           paths:
-            - ~/.ccache
-          key: -{{ checksum "Makefile.envs" }}-v20201205-{{ .BuildNum }}
+            - /root/.ccache
+          key: -core-{{ checksum "Makefile.envs" }}-v20210910-
```
rth, line 92: No longer store the cache for each build, as then we are not sure which one gets used, and it tend to grows with each build if there are cache misses

.circleci/config.yml
```diff
@@ -100,12 +115,17 @@ jobs:
...
+            - -pkg-{{ checksum "Makefile.envs" }}-v20210911-
 
       - run:
           name: build packages
           no_output_timeout: 1800
           command: |
+            source pyodide_env.sh
+
+            # Set mtime for EM_CONFIG to avoid ccache cache misses
+            touch -m -d '1 Jan 2021 12:00' emsdk/emsdk/.emscripten
```
rth, line 127: Source: https://github.com/juj/ccache/blob/d69e1d05d68b06a256f00a86a38de644ff23b66d/src/emcc.cpp#L627

This can probably be switched to hashing the file, but mtime+size is the default ccache strategy for compilers apparently (see `CCACHE_COMPILERCHECK`)

rth (commented):
It's finally working:

| Build time (min) | build emsdk+cpython+core | build packages |
|------------------|--------------------------|----------------|
| Without ccache   | 20                       | 80 ± 10        |
| ccache (main)    | 17 ± 2                   | 80 ± 10        |
| ccache (this PR) | 14 ± 2                   | 44             |

.circleci/config.yml
```diff
@@ -41,20 +44,29 @@ jobs:
 
       - restore_cache:
           keys:
-            - -{{ checksum "Makefile.envs" }}-v20201205-
```
hoodmane, line 44: Why do we want to keep the `core` and `packages` ccache separate? The `packages` ccache should include caches from building various packages? Perhaps we could use the same cache for both but only `save_cache` on `build-packages`? Perhaps keeping them separate is simpler.

.circleci/config.yml
hoodmane: So is the assumption here that the ccache will only be updated if `Makefile.envs` changes? Maybe it would be better to key also on `Makefile` and `cpython/Makefile`?

hoodmane (approved):

.circleci/config.yml
rth: Yes, it's a balance minimizing cache resets and doing that if something major changes. Currently `Makefile.envs` includes all the global flags. 
If I look at git history,
 - `Makefile` most of the changes there are either about js packaging, linting or something other about build that doesn't affect compilation
 - `cpython/Makefile`: most changes also also rather local (even libffi addition). Also I'd like to move most of those under the main build system as standalone packages eventually if possible.

The cost of rebuilding without cache is roughly 40min, and the issue is that the cache for each fork is independent as far as I can tell, so it's multiplied by the number of contributors. So I think even if a few compilation options change for a subset of files in a package (including CPython), and we get some cache misses overall it's probably still fine to keep the same cache. With the assumption that we update emsdk periodically which resets the cache anyway.

.circleci/config.yml
rth: > The packages ccache should include caches from building various packages?

Yes.

> Perhaps we could use the same cache for both but only save_cache on build-packages? Perhaps keeping them separate is simpler.

I think would I rather prefer to have them separate,
 - first, there are surprisingly many things that could go wrong with ccache (particularly with this little documented emsdk fork we are using), and than you have to debug by, setting,
   ```
   export CCACHE_DEBUG=1
   export CCACHE_DEBUGDIR="/root/ccache-debug/"
   ```
   storing that folder as artifacts, than comparing `*.ccache-input-text` and `*.ccache-log` files for subsequent rebuilds to see why there are cache misses. It's easier to do that if there are no inter-dependencies for cache between jobs.
 - for instance on main, ccache is broken for packages because cache loaded contains only files from core. It's hard to detect that because you see "33MB of ccache loaded" and so you think well maybe that's what's supposed to happen. If caches were seprate you would see "0MB of ccache loaded" and also 0 files in `ccache -s` reports making it more obvious that something is wrong.
 - And more generally, it practical to be able to manually reset only the core cache (which is fast to rebuild) and keep the packages cache unchanged if needed. Which also provides a solution for https://github.com/pyodide/pyodide/pull/1805#discussion_r706508747

.circleci/config.yml
rth: Actually, yes for the core build we can certainly include `cpython/Makefile` in the hash since the core/package caches are separate.

.circleci/config.yml
hoodmane: Oh right, given that `build-core` and `build-packages` don't overlap at all in what they build, there isn't even any benefit to sharing the caches.

PR #1806, opened 2021-08-26
Fix string conversion from javascript to python

### LINKED ISSUE #1801: `open_url` call results in infinite recursion in 0.18.0
Hi team,

Pyodide 0.18.0 seems to have an issue with the Python API for opening URLs, 'pyodide.open_url`, which results in an infinite recursion when returning. The following code snippet reproduces the error in the [Pyodide Console](https://pyodide.org/en/latest/console.html):

```python
>>> import pyodide; pyodide.open_url('https://plotly.github.io/datasets/country_indicators.csv')
  File "/lib/python3.9/site-packages/pyodide/_util.py", line 27, in open_url
    return StringIO(req.response)
JsException: RangeError: Maximum call stack size exceeded
```

I can confirm this is working in 0.17.0.

~The root cause seems to be the `_pythonexc2js` function, but I am not entirely sure.~ I'm happy to take a look if someone could point me in the right direction!

Many thanks 😄

hoodmane:
I introduced a bug in v0.18.0 that breaks conversion of long strings, thanks to @ibdafna and @ryanking13 for finding it. Hopefully this should fix the bug. Resolves #1801.

### DIFF
--- a/docs/project/changelog.md
+++ b/docs/project/changelog.md
@@ -20,6 +20,9 @@ substitutions:
   console.
   {pr}`1790`
 
+- {{Fix}} Conversion of very large strings from Javascript to Python works
+  again. {pr}`1806`
+
 ## Version 0.18.0
 
 _August 3rd, 2021_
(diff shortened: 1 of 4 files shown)


hoodmane:
Actually that OOM error must have been unrelated because 1) 400,000 isn't _that_ big I don't think and 2) package tests are run with `selenium_standalone`.

Chrome tests seem to be timing out a lot recently...

PR #1812, opened 2021-09-02
MAINT Simplify error handling in js2python.c

hoodmane:
This simplifies error handling in js2python a bit. In the old version, we propagate errors manually in the Javascript code: when an error branch is reached, the function returns 0, then each calling function checks this and returns 0. The code also does an odd thing where it throws `TempError` and then usually recatches it in the same function, effectively as a `goto fail`. I think this is because much of the code was rather directly converted from C to Javascript.

In this case what we want to do is if we call any CPython API that returns an error, we want to unwind all Javascript calls until we hit `EM_JS` and then return an error to C without adjusting the Python error flag. I added a special mechanism for this to `error_handling.c`: in this case we throw a `Module._PropagatePythonError`.

I also made internal Javascript helpers for `error_handling.c` private to the file and improved function naming a bit.

### DIFF
--- a/src/core/error_handling.c
+++ b/src/core/error_handling.c
@@ -225,6 +225,12 @@ char* error__js_filename_string = "???.js";
 EM_JS_NUM(errcode, error_handling_init_js, (), {
   Module.handle_js_error = function(e)
   {
+    if (e instanceof Module._PropagatePythonError) {
+      // Python error indicator is already set in this case. If this branch is
+      // not taken, Python error indicator should be unset, and we have to set
+      // it. I think in this case we don't want to tamper with the traceback?
+      return;
+    }
     let restored_error = false;
     if (e instanceof Module.PythonError) {
       // Try to restore the original Python exception.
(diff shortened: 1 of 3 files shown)


src/core/error_handling.c
```diff
@@ -225,6 +225,12 @@ char* error__js_filename_string = "???.js";
 EM_JS_NUM(errcode, error_handling_init_js, (), {
   Module.handle_js_error = function(e)
   {
+    if (e instanceof Module._PropagatePythonError) {
+      // Python error indicator is already set in this case. If this branch is
+      // not taken, Python error indicator should be unset, and we have to set
+      // it. I think in this case we don't want to tamper with the traceback?
```
rth, line 231:
```suggestion
      // it. In this case we don't want to tamper with the traceback.
```
otherwise it's not clear what the reader is supposed to do with questions in comments.

src/core/js2python.c
```diff
@@ -171,18 +190,13 @@ EM_JS_NUM(errcode, js2python_init, (), {
       cache.set(obj, list);
       for (let i = 0; i < obj.length; i++) {
         entryid = Module.hiwire.new_value(obj[i]);
-        item = Module.__js2python_convert(entryid, cache, depth);
-        // clang-format off
-        if (item === 0) {
```
rth, line 176: If I understand correctly this moves throwing of the error into js2python_convertOther inside js2python_convert?

src/core/js2python.c
```diff
@@ -359,12 +348,12 @@ EM_JS_NUM(errcode, js2python_init, (), {
     return _JsProxy_create(id);
   };
 
-  Module.__js2python_convert = function(id, cache, depth)
+  Module.js2python_convert = function(id, cache, depth)
```
rth, line 351: Could you please add a short docstring for this function?

rth (approved):
Thanks @hoodmane !

src/core/js2python.c
hoodmane: That is correct.

PR #1813, opened 2021-09-02
Updating PyRTL version

mdko:
This updates the PyRTL version to the latest. Was able to build successfully:
Screen Shot 2021-09-02 at 3 49 19 PM

Among the several fixes since the last PyRTL version is one that improves the Ascii renderer used during simulation (which is used when PyRTL is used in the Pyodide console).

For example, before the outputs weren't aligned:
Screen Shot 2021-09-02 at 4 04 46 PM

Now the output is aligned:
Screen Shot 2021-09-02 at 4 03 46 PM


rth:
Thanks @mdko ! Actually I think micropip installation from PyPi should work,
```py
>>> import micropip
>>> await micropip.install('pyrtl==0.10.1')
```
and the patch is no longer necessary. So I would be in favor of removing it from Pyodide packages, so that we don't have to update it, and one can install the latest version as soon as it's on PyPi (https://github.com/pyodide/pyodide/issues/643).

mdko:
> Thanks @mdko ! Actually I think micropip installation from PyPi should work,
> 
> ```python
> >>> import micropip
> >>> await micropip.install('pyrtl==0.10.1')
> ```
> 
> and the patch is no longer necessary. So I would be in favor of removing this package from Pyodide packages, so that we don't have to update it, and one can install the latest version as soon as it's on PyPi (#643).

Awesome. I've made a new commit removing all of PyRTL package entirely.

hoodmane:
Thanks @mdko!

PR #1816, opened 2021-09-06
Fix refcount error in restore_sys_last_exception

hoodmane:
According to the docs, [`PyErr_Restore`](https://docs.python.org/3/c-api/exceptions.html?highlight=repr#c.PyErr_Restore) steals a reference to each of its arguments (this also makes sense because it makes the common `PyErr_Fetch` / `PyErr_Restore` usage work nicely).
Anyways, I did not correctly take this into account in the definition of `restore_sys_last_exception`. As a result, bad things happen every time this code path is hit. This might be why `test_reentrant_error` fails sporadically. I found another test that consistently causes use after free errors (which doesn't lead to a fatal error because the problem occurs inside of the error handling code).

### DIFF
--- a/docs/project/changelog.md
+++ b/docs/project/changelog.md
@@ -40,7 +40,7 @@ substitutions:
   it was before.
   {pr}`1797`
 
-- {{Fix}} Fixed a serious but rare bug in the error handling code.
+- {{Fix}} Fixed a use after free bug in the error handling code.
   {pr}`1816`
 
 ## Version 0.18.0
--- a/docs/project/changelog.md
+++ b/docs/project/changelog.md
@@ -20,6 +20,9 @@ substitutions:
   console.
   {pr}`1790`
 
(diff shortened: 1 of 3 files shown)


docs/project/changelog.md
```diff
@@ -20,6 +20,9 @@ substitutions:
   console.
   {pr}`1790`
 
+- {{Fix}} Fixed a serious but rare bug in the error handling code.
```
rth, line 23: It would help to be more specific.

rth (approved):
Thanks for catching this!

docs/project/changelog.md
hoodmane: It would probably be better to say memory error / use after free bug? I guess I could also say that it is triggered when an error propagates Python ==> js ==> Python or something like that. I have to update the changelog anyways because of the conflict.

rth: I think "use after free bug" would be enough . Of course feel free to add more details is you prefer.

PR #1819, opened 2021-09-07
Handle syntax error in find_imports

### LINKED ISSUE #616: RunPythonAsync catch not working "Uncaught (in promise) Error"
Hi, I'm currently trying the Pyodide library inside a WebWorker, in a React project.

So my worker, will run the python code (which the user enter) and show the output even if there's an error.
Here is my code : 
```js
// myWorker.js
const data = {
  python: 'impt numpy as np'
}
languagePluginLoader.then(() => {
    pyodide
      .runPythonAsync(data.python, () => {})
      .then(results => {
        console.log(results);
        self.postMessage({ output: results });
      })
      .catch(err => {
        console.log(err);
        self.postMessage({ error: err.message });
    });
});
```

The problem here is the catch of `runPythonAsync(...).catch(err => {})` won't catch this kind of error thrown by `pyodide.asm.js` file, we'll only see it in the console..

image
```
Uncaught (in promise) Error: Traceback (most recent call last):
  File "/lib/python3.7/site-packages/pyodide.py", line 58, in find_imports
    mod = ast.parse(code)
  File "/lib/python3.7/ast.py", line 35, in parse
    return compile(source, filename, mode, PyCF_ONLY_AST)
  File "<unknown>", line 1
    impt numpy as np
             ^
SyntaxError: invalid syntax

    at _hiwire_throw_error (http://localhost:3000/static/modules/pyodide.asm.js:8:38727)
    at wasm-function[358]:0xfc6c0
    at wasm-function[386]:0xfe474
    at Object.Module.__findImports (http://localhost:3000/static/modules/pyodide.asm.js:8:1989949)
    at Object.Module.runPythonAsync (http://localhost:3000/static/modules/pyodide.asm.js:8:43347)
    at http://localhost:3000/443e11adcc88e32bff24.worker.js:106:8
```
So how I can handle this type of Python error ? Cause right now the catch function able me to catch error when something goes wrong in the `then(()=> {})` function.. 
Thanks.

hoodmane:
> it would still be best if `find_imports` caught SyntaxError and returned an empty list.

_Said @alexmojaki in https://github.com/pyodide/pyodide/issues/616#issuecomment-913799555_ 
I agree. This resolves #616.

### DIFF
--- a/docs/project/changelog.md
+++ b/docs/project/changelog.md
@@ -43,6 +43,10 @@ substitutions:
 - {{Fix}} Fixed a use after free bug in the error handling code.
   {pr}`1816`
 
+- {{Enhancement}} If `find_imports` is used on code that contains a syntax
+  error, it will return an empty list instead of raising a `SyntaxError`.
+  {pr}`1819`
+
 ## Version 0.18.0
 
 _August 3rd, 2021_
--- a/src/py/_pyodide/_base.py
+++ b/src/py/_pyodide/_base.py
@@ -515,7 +515,8 @@ def find_imports(source: str) -> List[str]:
     Returns
     -------
     ``List[str]``
-        A list of module names that are imported in ``source``.
+        A list of module names that are imported in ``source``. If ``source`` is not
+        syntactically correct Python code (after dedenting), returns an empty list.
 
     Examples
     --------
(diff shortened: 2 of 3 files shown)


alexmojaki:
:fireworks:

rth (approved):
LGTM, thanks! It would need a changelog entry. Also I don't think it fully resolves that issue https://github.com/pyodide/pyodide/issues/616#issuecomment-914330202

PR #1823, opened 2021-09-09
FIX Patch emscripten to fix problems with dup

### LINKED ISSUE #1791: How should I test my own package?
Is there any official pattern for testing a 3rd party package on Pyodide, that does not intend to have their package become part of Pyodide?

I'm trying to run my own package, [dit_cli](https://github.com/ditabase/dit-cli), in Pyodide. This was actually quite easy to set up, you can see a very rough demo [here](https://dev.ditabase.io/demo.html). This is a 100% pure Python package, I'm not using the science stack, I just want to run Python/WASM efficiently in the browser.

Now I would like to set up testing, following the Pyodide pattern using Selenium. I've got the basics working, but a lot of what `conftest.py` does is very unclear, and apparently not intended for testing 3rd party packages.

- Should I package Pyodide with my project, just for testing?
- Should I just copy `conftest.py` and the rest of the testing system? What else do I need to know if I do that?

hoodmane:
This should fix two problems with the `dup` system calls:
1. Pipes cannot be duplicated (https://github.com/emscripten-core/emscripten/issues/14640)
2. `TemporaryFiles` cannot be duplicated (https://github.com/emscripten-core/emscripten/issues/15012)

Both of these issues cause trouble with pytest. There is an upstream pull request that would fix this problem:
https://github.com/emscripten-core/emscripten/pull/9396/

This patch only partially resolves the problems with `dup` (it doesn't fully duplicate the changes in the emscripten PR) but I think it will be good enough to fix pytest.

This resolves #1791.

### DIFF
--- a/docs/project/changelog.md
+++ b/docs/project/changelog.md
@@ -69,6 +69,13 @@ substitutions:
 
 - {{Fix}} pillow now correctly encodes/decodes JPEG image format. {pr}`1818`
 
+### Micellaneous
+
+- {{Fix}} Patched emscripten to make the system calls to duplicate file
+  descriptors closer to posix-compliant. In particular, this fixes the use of
+  `dup` on pipes and temporary files, as needed by `pytest`.
+  {pr}`1823`
+
 ## Version 0.18.0
 
 _August 3rd, 2021_
--- a/emsdk/patches/fix-dup.patch
+++ b/emsdk/patches/fix-dup.patch
@@ -70,13 +70,6 @@ index 96d2ec0c3..0001624ec 100644
    },
    __sys_pipe2__nothrow: true,
    __sys_pipe2__proxy: false,
-diff --git a/tests/third_party/posixtestsuite b/tests/third_party/posixtestsuite
-index 2e3f88374..a5e7ff5f9 160000
---- a/tests/third_party/posixtestsuite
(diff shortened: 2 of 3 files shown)


rth (approved):
Sounds good, though I might have been better if this was exactly that PR, and the emscripten tests passed there (which is not the case currently). As it is I don't have a very high confidence that this won't have some other unintended side effects. Still our test suite looks good.

Could you also add a test with dup2 for TemporaryFile and pipe?

hoodmane:
> I don't have a very high confidence that this won't have some other unintended side effects

I've thought about this fairly carefully. The affected code that I changed was quite wrong and causes a bunch of different bugs. (I got one PR accepted to emscripten that improved the situation with pipes a bit.) I am pretty confident that the code in this PR is better than the unpatched code but worse than a correct PR.

In particular, the original code has two bugs:
1. it thinks every file descriptor should point to a file, and
2. seek position and flags are not shared between dup'd descriptors

The upstream PR fixes 1+2, this PR just fixes 1. This is basically a subset of the upstream changes. When an upstream PR lands, we can replace this patch with the upstream one.

rth:
Sounds good. Thanks!

hoodmane:
Further investigation shows that `dup` is indeed still quite buggy. At least pytest doesn't crash now though.

PR #1825, opened 2021-09-10
Ensure that dup'd file descriptors share position and flags

hoodmane:
This is a continuation of #1823. This should make pytest work better by ensuring that a duplicate file descriptor shares position with the original one. pytest does something like:
```
stdout_redirect_file = TemporaryFile()
saved_stdout_fd = os.dup(sys.stdout.fileno())

def redirect_stdout():
	os.dup2(stdout_redirect_file.fileno(), sys.stdout.fileno())

def restore_stdout():
    os.dup2(saved_stdout_fd, sys.stdout.fileno())
```
The problem comes up when `stdout` is redirected, restored, and redirected again. Instead of adding new writes to `stdout` to the end of the `stdout_redirect_file`, they write over `stdout_redirect_file` from the beginning. The intended behavior is that they would be added to the end. This is because `stdout_redirect_file` is always positioned at the beginning of the file so every time `redirect_stdout` is called, `stdout` is moved to point to the beginning of the file. 

On the other hand, if `dup2` behaved the way the posix spec says, then the position of `stdout_redirect_file` is updated whenever `stdout` is written, and so after repeated calls to `redirect_stdout`, writes to stdout will still be appended to the end of the file.

### DIFF
--- a/emsdk/patches/fix-dup.patch
+++ b/emsdk/patches/fix-dup.patch
@@ -1,22 +1,50 @@
-From 52c0b163bad96cde2081c734680b9ec485fea908 Mon Sep 17 00:00:00 2001
+From 05458d136277cf801c22f7d8bdac99d6b157209d Mon Sep 17 00:00:00 2001
 From: Hood <hood@mit.edu>
 Date: Wed, 8 Sep 2021 17:49:15 -0700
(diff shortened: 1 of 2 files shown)


rth (approved):
Thanks!

PR #1838, opened 2021-09-15
MAINT Remove deprecated functions for 0.18.0

rth:


hoodmane:
While we're at it I think it would be good to remove the numpy segfault patches (the bad versions 89+90 of chrome make up about 0.5% of browser usage now).

rth:
Indeed, but then we would need to update Chrome used for tests to at least 91, and since it's currently coming from the system install it's not very straightforward. So probably better to do it in a separate PR.

hoodmane:
Okay. It's a bit surprising to me that upgrading the selenium browser to the latest version of Chrome / Firefox is so tricky -- would think everyone wants to do this.
I wonder whether we could copy the code here:
https://github.com/SeleniumHQ/docker-selenium/blob/trunk/NodeChrome/Dockerfile

rth:
I'm still getting a test failure in  test-docs because I haven't managed to update `docs/sphinx_pyodide/tests/jsdoc_dump.json.gz`. When I run,
```
jsdoc -c ./docs/jsdoc_conf.json -X ./src/js/ ./src/core/ | gzip > docs/sphinx_pyodide/tests/jsdoc_dump.json.gz
```
from the docker image, I get errors about,
```
ERROR: Unable to parse a tag's type expression for source file /src/src/js/module.js in line 
12 with tag title "param" and text "{undefined|(() => string)} stdin": Invalid type expression
 "undefined|(() => string)": Expected "!", "$", "'", "*", ".", "...", "0", "?", "@", 
"Function", "\"", "\\", "_", "break", "case", "catch", "class", "const", "continue", 
"debugger", "default", "delete", "do", "else", "enum", "export", "extends", "false", 
"finally", "for", "function", "if", "implements", "import", "in", "instanceof", "interface", 
"let", "new", "null", "package", "private", "protected", "public", "return", "static",
 "super", "switch", "this", "throw", "true", "try", "typeof", "undefined", "var", "void", 
"while", "with", "yield", "{", Unicode letter number, Unicode lowercase letter, Unicode 
modifier letter, Unicode other letter, Unicode titlecase letter, Unicode uppercase letter, or 
[1-9] but "(" found.
```
though the file does seem to be generated.
@hoodmane Did you get those also the first time you run them, or does it mean that some of those signatures changed in between?

hoodmane:
@rth These errors appear every time. They show up since #1728 because the typescript function types are not parsed by jsdoc. They seem to be okay in `@typedef`'s, but this is probably because jsdoc just ignores the typedef tag.

This problem would best be solved by migrating to typescript. However, I expect doing so to be a bit over-constrained and frustrating since getting the typescript type generator, jsdoc, sphinx_js, emscripten, and rollup to all agree to accept the same files was quite hard. Another complication is that we have two different ways of loading js files, via rollup and via `EM_JS` and `#include` and we will need to account for that.

rth:
OK, I still don't understand why `test_pyodide_analyzer` fails even after regenerating `jsdoc_dump.json.gz`,
```
def test_pyodide_analyzer():
        function_names = {x.name for x in pyodide_analyzer.js_docs["pyodide"]["function"]}
        attribute_names = {x.name for x in pyodide_analyzer.js_docs["pyodide"]["attribute"]}
        ...
E       AssertionError: assert {'checkInterr...sModule', ...} == {'isPyProxy',...onAsync', ...}
E         Extra items in the left set:
E         'setStandardStreams'
E         'registerComlink'
E         'checkInterrupt'
E         Use -v to get the full diff
```

@hoodmane Would you be able to have a look, please?

PR #1840, opened 2021-09-16
Update docs following #1573 [skip ci]

hoodmane:
I added some updates to the docs following #1573. I deleted the old "avoiding memory leaks in ffi calls" section and wrote a new section describing more specifically what goes on when calling cross-language functions. New text is a bit shorter and better organized.

There are also some unrelated updates to the FAQ.

### DIFF
--- a/docs/usage/faq.md
+++ b/docs/usage/faq.md
@@ -193,7 +193,7 @@ def f(*args):
 document.body.addEventListener('click', f)
 ```
 
-Now every time you click, an error will be raised (see {ref}`type-translations-proxy-arguments`).
+Now every time you click, an error will be raised (see {ref}`call-js-from-py`).
 
 To do this correctly use {func}`pyodide.create_proxy` as follows:
 
--- a/docs/usage/faq.md
+++ b/docs/usage/faq.md
@@ -255,13 +255,29 @@ resp = await js.fetch('example.com/some_api', to_js({
 
 ## How can I control the behavior of stdin / stdout / stderr?
 
-This works much the same as it does in native Python: you can overwrite
-`sys.stdin`, `sys.stdout`, and `sys.stderr` respectively. If you want to do it
-temporarily, it's recommended to use
+If you wish to override `stdin`, `stdout` or `stderr` for the entire Pyodide
+runtime, you can pass options to {any}`loadPyodide <globalThis.loadPyodide>`: If
+you say
+
+```
+loadPyodide({
+  ..., stdin: stdin_func, stdout: stdout_func, stderr: stderr_func
+});
+```
+
+then every time a line is written to `stdout` (resp. `stderr`), `stdout_func`
+(resp `stderr_func`) will be called on the line. Everytime `stdin` is read,
+`stdin_func` will be called with zero arguments. It is expected to return a
+string which is interpreted as a line of text.
+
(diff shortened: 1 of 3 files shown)


docs/usage/faq.md
```diff
@@ -1,18 +1,24 @@
...
   `python setup.py bdist_wheel` and
   {ref}`load it with micropip <micropip-installing-from-arbitrary-urls>`.
-- fetch the Python code as a string and evaluate it in Python,
+- fetch Python code as a string and evaluate it:
   ```js
   pyodide.runPython(await (await fetch("https://some_url/...")).text());
   ```
+- Use `pyodide.FS.writeFile(path, data)` to directly create a file inside the
+  file system. Note that the directory must already exist, if not use
+  `pyodide.FS.mkdir` first. The file system API is documented
```
rth, line 16: That exists, by why would we be recommending that as opposed to python `open`, `os.mkdir` etc? Unless one would like to do it on the JS side. 

Also it's not really an alternative solution to fetch above, I would say, it still would need to be fetched somehow.

docs/usage/type-conversions.md
```diff
@@ -326,6 +329,115 @@ Javascript map can use `true` and `1` as distinct keys but a Python `dict`
...
+    test
+`);
+let result_py = test([1,2,3,4]);
+// result_py is a PyProxy of a list.
+let result_js = result_py.toJs();
+// result_js is the array [1, 4, 9, 16]
+result_py.destroy();
+```
+
+If a function is indended to be used from Javascript, you can use {any}`to_js <pyodide.to_js>` on the return value. This is particularly important for
```
rth, line 356: Maybe it would be good to repeat why one would use `to_js` there

docs/usage/type-conversions.md
```diff
@@ -326,6 +329,115 @@ Javascript map can use `true` and `1` as distinct keys but a Python `dict`
...
+`{a : 2, b : 3}`.
+
+When a Javascript function is called and it returns anything but a promise, if
+the result is a `PyProxy` it is destroyed. Also, any arguments that are
+PyProxies that were created in the process of argument conversion are also
+destroyed. If the `PyProxy` was created in Python using
+{any}`pyodide.create_proxy` it is not destroyed.
+
+When the Javascript function returns a `Promise` (for example, if the function
+is an `async` function), the behavior is more complicated. In this case, the
```
rth, line 407:
```suggestion
is an `async` function), the behavior is more complicated to minimize memory leaks. 
In this case, the
```

rth (approved):
A few comments, otherwise LGTM. Thanks!

docs/usage/type-conversions.md
hoodmane: Well it's sort of the opposite: we destroy stuff to avoid memory leaks. If a `Promise` is returned, there is a high chance the arguments are still needed for whatever work the `Promise` is going to do and so that destroying them will cause trouble. So when you return a Promise the lifetime of the proxies is extended from "for the duration of this function call" to "until this `Promise` resolves".

hoodmane:
@rth I think it might be a good idea to release an 0.19.0a1 soon: #1573 is a pretty significant change by itself which may break peoples' code.

PR #1847, opened 2021-09-17
Fix bug when function returns destroyed proxy

### LINKED ISSUE #1841: Low level crash due to destroyed proxy
The following code causes a low level failure:
```js
globalThis.f = function(x){ return x };
pyodide.runPython(`
   from pyodide import create_proxy
   from js import f
   p = create_proxy([])
   p.destroy()
   f(p) 
`);
```
This is a regression caused by #1573. Discovered while writing #1840.

hoodmane:
This resolves #1841.

The root cause is that we have a bunch of `EM_JS_NUM` functions with boolean return types. These return `-1` if there is an error, but they are used like:
```C
if(js_conditional(args)){
	// True case
} else {
	// False case
}
```
If an error is thrown inside of `js_conditional` then `EM_JS_NUM` returns `-1` to indicate an error occurred. So then we take the true case of the conditional and erroneously leave the error flag set. If we wish to use these functions like that then they cannot raise an exception, they can only succeed (and return true or false) or fatally fail. So I switched all of these boolean return value functions to use `EM_JS`. This way, if errors are raised inside of them they will directly produce fatal errors which are much easier to debug. Otherwise, bad stuff happens anyways but it looks more confusing.

Once this problem masking the bug is removed, the issue is that if an error is thrown inside of `hiwire_is_promise` we need to catch it and return false.

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -226,6 +226,13 @@ class SeleniumWrapper:
     def get_num_hiwire_keys(self):
         return self.run_js("return pyodide._module.hiwire.num_keys();")
 
+    @property
+    def force_test_fail(self) -> bool:
+        return self.run_js("return !!pyodide._module.fail_test;")
+
+    def clear_force_test_fail(self):
+        self.run_js("pyodide._module.fail_test = false;")
+
     def save_state(self):
         self.run_js("self.__savedState = pyodide._module.saveState();")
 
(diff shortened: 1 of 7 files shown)


hoodmane:
Sure enough, switching those `EM_JS_NUM` to `EM_JS` turned up another spot where code was doing strange things but in a way that didn't cause any obvious problem.

hoodmane:
Actually it was printing an `Internal Error` message to the console in that case so the problem wasn't completely swallowed. But CI passed.

hoodmane:
Oh the other problem was also introduced by #1573.

rth (approved):
Thanks! This would need a changelog entry?

Issue #1839, opened 2021-09-15
Release Python code as pypi package


hoodmane:
In [this discussion](https://github.com/pyodide/pyodide/discussions/1833#discussioncomment-1335038) @grimmer0125 identified the issue that `JsProxy` cannot be used as a type label: static type analysis tools won't be able to suggest autocompletions and mypy etc can't typecheck the file. We already went to the trouble to make it possible to import pyodide in native Python, but we need some way to install it for the type analysis software. I think the best solution is to the Python code in `src/py` as a pypi package.

grimmer0125:
I just recently noticed this, `stub` https://www.python.org/dev/peps/pep-0484/#stub-files and not sure if it is another solution or not. I have no experience with it and just pass this info. [Mypy's stub section](https://mypy.readthedocs.io/en/stable/stubs.html)

rth:
Yes, this would also be necessary to pass host micropip tests in https://github.com/pyodide/micropip/pull/3.

How should we call this package on PyPi: `pyodide-core-py` ?

rth:
Actually another important reason we need this package on PyPi, and for it to be runnable in host as much as possible is notebooks. If someone uses any functionality  (e.g. pyfetch) from this package in say JupyterLite, then download the notebook and try to run it locally, it will fail unless they are also able to pip install this package.

> How should we call this package on PyPi: pyodide-core-py ?

Any objections?

hoodmane:
Well, it's not a great name, but I guess it gets the point across.

rth:
I share the sentiment. It's just that `pyodide-core` is technically also the FFI, and the JS package. 

Maybe actually `pyodide-py` would be better?

hoodmane:
Yeah let's go with `pyodide-py` maybe. That also isn't fantastic but I think it has a nice enough ring to it. 

At some point we could try to change the docs to call it `pyodide-py` more often, since we still have the confusion between `pyodide` the Python module accessible from JavaScript as `pyodide.pyodide_py` and `pyodide` the Javascript object importable from Python as `pyodide_js`.

rth:
I uploaded [`pyodide-py`](https://pypi.org/project/pyodide-py/) to PyPi some month ago, as we needed it in micropip. We still need to auto-deploy it in CI.

PR #1848, opened 2021-09-18
Add package config for pyodide package

hoodmane:
I added configuration files for the Pyodide package to allow `pip install -e src/py`. This is intended to help with #1839. Also I removed a few hacks that confuse type checking software.


Makefile
```diff
@@ -156,10 +156,11 @@ lint: node_modules/.installed
 	mypy --ignore-missing-imports    \
 		pyodide-build/pyodide_build/ \
 		src/ 					     \
-		packages/micropip/src/ 		 \
 		packages/*/test* 			 \
 		conftest.py 				 \
 		docs
+	mypy --ignore-missing-imports    \
+		packages/micropip/src/
```
rth, line 163: What was the issue there? Passing everything should give it more type information I imagine.
If you don't want to debug that here, maybe add a TODO in a comment?

hoodmane: mypy gets angry about there being multiple files called `setup.py`:
https://github.com/python/mypy/issues/10428
I tried a bunch of different things to fix it but this was the only thing that seemed to work well. Will add a comment, hopefully mypy will improve this eventually.

hoodmane:
> Can we not keep this either part of pyodide or _pyodide. It's not great to have to install 3 packages.

> Why can't we set `IN_BROWSER` in pure python as before?

The reason is that type checkers that do static analysis rather than running the code do not understand the following code:
```py
sys.modules["some_name"] = some_module
from some_name import SomeType
```
What happens is that `SomeType` gets type `Any` because `some_name` is an unknown package. In particular the vscode analysis software has this problem. So issues 1 and 3 from @grimmer0125's list here:
https://github.com/pyodide/pyodide/discussions/1833#discussioncomment-1335038 will still come up.

But perhaps there is a better way to achieve this.

hoodmane:
Again, we're in a similar situation as with the typescript, jsdoc, sphinx-js, emscripten, and rollup all have to understand the same js types, we want need native Python and Pyodide as well as mypy, pylance, and whatever other type checkers to understand the same Python code. This means we can't do anything that makes any of these tools confused or unhappy...

I think that the following probably works:
```
IN_BROWSER = True
try:
   from _core_module import blah1, blah2, blah3
except:
   from _pyodide._core_docs import blah1, blah2, blah3
   IN_BROWSER = False	
```
but I like the approach I have currently in this PR better. In the browser there are three packages: `_core_module`, `_pyodide`, and `pyodide`. I think it makes sense to also have three packages when using the code natively in order to maximize consistency.

hoodmane:
But actually either approach is probably fine. If you prefer the other way, I will switch it.

rth:
>  In the browser there are three packages: _core_module, _pyodide, and pyodide. I think it makes sense to also have three packages when using the code natively in order to maximize consistency.

Right, but while I understand the need for `_pyodide` during pyodide init, `_pyodide_core` looks more like a hack for typing. Also there are circular imports between `_pyodide` and `_pyodide_core` which is not so good from a Python point of view.

Mypy can sometimes be fastidious but normally with [`typing.TYPE_CHECKING`](https://mypy.readthedocs.io/en/stable/runtime_troubles.html#typing-type-checking) I think we can customize/rewrite a lot of what is happening when types are checked. 

I'm just concerned that this going to be confusing to contributors. So I would rather we check if there are some way to avoid having this extra module and setting from C code.
```py
IN_BROWSER = True
...
try:
   ...
except ImportError:
   IN_BROWSER = False
```
is just so much easier to understand IMO than,
```py
IN_BROWSER = True
```
and False is set somewhere in C code, that gets loaded from the JS module, than you can only obtain if you build Pyodide from sources and oh you need Docker do to that :)

But then I have not worked as much on this part of the code as you, so I might miss some of the constraints.

hoodmane:
There is no circular import between `_pyodide` and `_pyodide_core`. In both browser and shim version they are topologically sorted as follows:
* `_pyodide`
* `_pyodide_core`
* `pyodide`
Packages depend on ones above them but not below.

That aside, I think you are right. I will try to do it the other way.

> normally with `typing.TYPE_CHECKING` I think we can customize/rewrite a lot of what is happening when types are checked. 

I have only passing familiarity with this, but I will keep it in mind if I run into trouble.

hoodmane:
> There is no circular import between `_pyodide` and `_pyodide_core`.

I'm sorry I was not correct. `_pyodide._importhook.register_js_finder` needs an import from `_pyodide_core`. That's why `register_js_finder` is a function and happens at initialization time (in browser, in native Python it is never called).

hoodmane:
In any case, no issues came up in the process of rearranging things and the way you suggested is definitely better.

rth (approved):
Thanks @hoodmane !

PR #1849, opened 2021-09-19
Fix `IN_NODE`: avoid error with unrelated global.

### LINKED ISSUE #1846: Error calling `loadPyodide` if a global exists named `process`
Reproducing example:
```html
<script>
function process(buffer) {
    // Some interesting processing occurs here...
}
</script>
<script src="https://cdn.jsdelivr.net/pyodide/v0.18.1/full/pyodide.js"></script>
<script>
async function main() {
    let pyodide = await loadPyodide({ indexURL : "https://cdn.jsdelivr.net/pyodide/v0.18.1/full/" })
    console.log("Done loading.")
}
main()
</script>
```
This can be worked around by making sure pyodide is pulled in first. However, the same fix does not work in a web worker:
```js
importScripts("https://cdn.jsdelivr.net/pyodide/v0.18.1/full/pyodide.js")

function process(buffer) {
    // Some interesting processing occurs here...
}

async function main() {
  await loadPyodide({ indexURL: "https://cdn.jsdelivr.net/pyodide/v0.18.1/full/" })
  console.log("Done loading.")
}

main()
```

Both of these examples fail with an error like `TypeError: process.release is undefined` or `TypeError: Cannot read property 'name' of undefined`.

I suspect this is an unintended consequence of the check to see if Pyodide is running in Node.js [here](https://github.com/pyodide/pyodide/blob/f03db870a25039bd70ac8822e0c3fe8be8ddb3fa/src/js/load-pyodide.js#L4). I think the check should instead look something like
```js
const IN_NODE = typeof process !== "undefined" && typeof process.release?.name !== "undefined"
```
(See also: the comments on this [SO answer](https://stackoverflow.com/a/35813135/13204291).)

ijc8:
Fixes #1846.

### DIFF
--- a/docs/project/changelog.md
+++ b/docs/project/changelog.md
@@ -19,6 +19,12 @@ substitutions:
   error, it will return an empty list instead of raising a `SyntaxError`.
   {pr}`1819`
 
+### Javascript package
+
+- {{Fix}} {any}`loadPyodide <globalThis.loadPyodide>` no longer fails in the
+  presence of a user-defined global named `process`.
+  {pr}`1849`
+
 ### Python / JavaScript type conversions
 
 - {{Enhancement}} Updated the calling convention when a Javascript function is
--- a/src/js/load-pyodide.js
+++ b/src/js/load-pyodide.js
@@ -1,7 +1,7 @@
 import { Module } from "./module.js";
 
 const IN_NODE =
-  typeof process !== "undefined" && process.release.name !== "undefined";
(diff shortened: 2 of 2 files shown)


hoodmane (approved):
Thanks! Will merge when CI finishes.

rth:
Thanks ! We should probably reuse the `IN_NODE` variable in https://github.com/pyodide/pyodide/blob/main/src/js/load-pyodide.js#L71 as well, as otherwise that will raise the same error.

PR #1854, opened 2021-09-21
Allow manually destroying borrowed attribute proxies

### LINKED ISSUE #1853: memory leak when accessing a.b. a is a Python instance & PyProxy in JS, b is either np.ndarray/bytes and is PyProxyBuffer in JS
Test environment: 
- OS: macOS 11.6
- Chrome: 93.0.4577.82
- Hardware: MacBook Air (M1, 2020)
- Pyodide: 0.18.0

User scenario: 
I'm trying to make a medical image viewer which needs frequent (**10~30 fps**) rerendering in some cases.  I define a Python class and plan to reuse it on JS side, this idea is from #1426 and a part of the code of the flow is like the part in https://github.com/pyodide/pyodide/discussions/1833#discussioncomment-1329114. 

Python class (in an external Python file): 

```
@dataclass
Class Dicom:
    ndarray_data: Optional[np.ndarray] = None # PyProxyBuffer in JS
    pixel_bytes: Optional[bytes] = None. #  # PyProxyBuffer in JS
```

When the mouse button is pressed and a mouse move event is triggered, JS will 
- execute `dicomObj.rerender()`.  dicomObj is a PyProxy. 
    - some `ndarray`/`bytes` internally are re-calculated and renewed to `ndarray_data/pixel_bytes` attributes on Python side.
- trying to access the resulted ndarray_data or pixel_bytes. 

The accessing result is 
- ndarray_data
    - as long as `dicomObj.rgba_1d_ndarray` is accessed on JS. The memory leak is over **100 MB/s** (fps is over 30fps), regardless I tried `rgba_1d_ndarray.destroy()` and commenting the following `getBuffer` usage.
    - I tried `@property` to access Python internal ndarray_data but got the same memory leak.  
    - Use a method to access internal `ndarray` data and the memory **will not leak**. Use `dicomObj.get_ndarray_data()` to access. On python:
    ```
    ## workaround solution
    def get_ndarray_data(self):
        return self.ndarray_data
    ```
- pixel_bytes: 
    - it will have memory leak but very slow (in similar flow), **1 MB/s**  (fps is over 30fps). And direct accessing the attribute, using property, and even using a method will all leak. Comparing to `ndarray_data`, this part is not very harmful.

hoodmane:
This should fix #1853. A JsProxy created by an attribute lookup `let p_attr = p.attr` should be destroyed when either:
1. `p.destroy()` is called, or
2. `p_attr.destroy()` is called.

This way manual cleanup of attributes is not needed as long as the root object is destroyed. This gives a way to  deal with cases like #1853 where the root object should stick around but its attributes need to be destroyed.

### DIFF
--- a/docs/project/changelog.md
+++ b/docs/project/changelog.md
@@ -32,6 +32,10 @@ substitutions:
   arguments and return values are automatically destroyed when the function is
   finished. {pr}`1573`
 
+- {{Fix}} It is now possible to destroy borrowed attribute `PyProxy` of a
+  `PyProxy` (as introduced by {pr}`1636`) before destroying the root `PyProxy`.
+  {pr}`1854`
+
 ### pyodide-build
 
 - {{API}} By default only a minimal set of packages is built. To build all
--- a/pyodide-build/pyodide_build/testing.py
+++ b/pyodide-build/pyodide_build/testing.py
@@ -80,17 +80,13 @@ def run_in_pyodide(
                     selenium.run_js(
                         f"""
                         let eval_code = pyodide.pyodide_py.eval_code;
-                        try {{
-                            eval_code.callKwargs(
-                                {{
-                                    source : atob({encoded}.join("")),
-                                    globals : pyodide._module.globals,
-                                    filename : {filename!r}
-                                }}
-                            )
-                        }} finally {{
-                            eval_code.destroy();
-                        }}
+                        eval_code.callKwargs(
+                            {{
+                                source : atob({encoded}.join("")),
+                                globals : pyodide._module.globals,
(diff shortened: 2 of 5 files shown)


rth (approved):
Thanks! Could you add a non regression test and a changelog entry? Otherwise LGTM.

hoodmane:
One concern about this is that it revives a version of #1049:
```js
let a1 = proxy.a;
let a2 = proxy.a;
a1 === a2 // true so if I destroy one it destroys both.
a1.destroy(); // Now both a1 and a2 are destroyed
let a3 = proxy.a; // a3 is a new JsProxy, not equal to the first two.
```

Issue #1683, opened 2021-07-02
`ModuleNotFoundError` after using `micropip.install` on custom wheel url


hoodmane:
I'm facing a similar issue of running 

```py
import micropip
await micropip.install('./wheels/pyahocorasick-1.4.2-cp38-cp38-linux_x86_64.whl')
import ahocorasick
```

where `/wheels/pyahocorasick-1.4.2-cp38-cp38-linux_x86_64.whl` is a manually compiled wheel of `pyahocorasick` using `pip wheel`, but I'm getting `ModuleNotFoundError`. Any ideas on fixing this?

_Originally posted by @kevinlu1248 in https://github.com/pyodide/pyodide/issues/500#issuecomment-871803889_

NishiPhalke:
Hi @hoodmane,

NishiPhalke:
Hi @hoodmane,

NishiPhalke:
Hi @hoodmane ,

I am trying install whl using micropip install URL(https://storage.googleapis.com/gcp.wenglab.org/bitarray-2.3.4-cp38-cp38-linux_x86_64.whl) created out of bitarray (https://github.com/ilanschnell/bitarray). But I am getting  this error when try to import bitarray.
Traceback (most recent call last):
  File "<console>", line 1, in <module>
  File "/lib/python3.9/site-packages/bitarray/__init__.py", line 14, in <module>
    from bitarray._bitarray import (bitarray, decodetree, _sysinfo,
ModuleNotFoundError: No module named 'bitarray._bitarray'

NishiPhalke:
@rth Thanks for your detailed response!

ryanking13:
This issue is resolved by #1859. Closing. If this behavior remains, please reopen this issue or make a new one.

PR #1859, opened 2021-09-25
Raise error when trying to install a non-pure python wheel

ryanking13:
Related Issue: #1683

This PR makes micropip raise an error when trying to install a non-pure (platform-dependent) python wheel.

### DIFF
--- a/docs/project/changelog.md
+++ b/docs/project/changelog.md
@@ -48,6 +48,11 @@ substitutions:
 - {{ Enhancement }} Better support for ccache when building Pyodide
   {pr}`1805`
 
+### micropip
+
+- {{Fix}} micropip now raises error when installing non-pure python wheel directly from url.
+  {pr}`1859`
+
 ### packages
 
 - {{ Enhancement }} Unit tests are now unvendored from Python packages and
--- a/packages/micropip/src/micropip/micropip.py
+++ b/packages/micropip/src/micropip/micropip.py
@@ -92,6 +92,10 @@ async def _get_pypi_json(pkgname):
     return json.load(fd)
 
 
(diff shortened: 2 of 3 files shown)


hoodmane:
Thanks @ryanking13! Looks like a decent solution to me.

I was looking at [PEP 425](https://www.python.org/dev/peps/pep-0425/) and the [packaging](https://packaging.pypa.io/en/latest/tags.html#packaging.tags.sys_tags) apis. It seems that technically we can accept a lot more target triples, since we do implement the cpython ABI, so technically a wheel labeled e.g., `cp39-cp39-any` would work fine. I don't know if there are any wheels actually labeled like this in the wild. Of course a package could be written in pure Python but use some CPython implementation details. Likely such a wheel in the wild would just be labeled `py3-none-any`. If these are indeed vanishingly rare, then it's not worth worrying about.

But we could consider only requiring that the platform be `any` instead of the whole triple being `py3-none-any`.

In the future we could consider distributing Pyodide extensions as wheels with triple`cp39-cp39-emscripten_wasm32`.

hoodmane (approved):

ryanking13:
> But we could consider only requiring that the platform be any instead of the whole triple being py3-none-any.

According to PEP425, there can be a pypy/jython targeted package (I have never seen such cases in the wild though).

Maybe we can try something like:

```python
(
    python_tag.startswith(["py2.py3", "py3", "cp3"]) and \
    abi_tag.startswith(["none", "abi3", "cp3"]) and \
    platform_tag == "any"
)
```

hoodmane:
> we can update the label checking logic later if needed?

Yes.

hoodmane:
Thanks again!

PR #1860, opened 2021-09-26
Replace "Javascript" with "JavaScript" in documents and comments

grimmer0125:
I saw more GitHub repositories use "JavaScript" in their documentation part. So I make this PR.


hoodmane:
Thanks. I think JavaScript looks kind of ugly though... \*\*shrug\*\*

ryanking13:
I vote for [JavaScript](https://stackoverflow.com/questions/3989731/is-it-javascript-or-javascript) 😂

rth:
Thanks! It's true that it's the official name.

PR #1864, opened 2021-09-28
Added TypedArray methods tostring, tobytes, and tobytearray to help minimize copies

hoodmane:
Currently we convert Javascript TypedArrays to memoryviews. This works well if we are planning to use them with numpy or similar. However, if we are planning to convert to a type like `string`, `bytes`, or `bytearray` it forces us to make an extra copy of the data. This adds direct conversions so that we can convert to `bytes` and `bytesarray` with one copy instead of two and to string with two copies instead of three.

### DIFF
--- a/src/py/_pyodide/_core_docs.py
+++ b/src/py/_pyodide/_core_docs.py
@@ -108,25 +108,25 @@ class JsProxy:
         an ArrayBuffer view.
         """
 
-    def tomemoryview(self):
+    def tomemoryview(self) -> memoryview:
         """Convert the buffer to a memoryview.
 
         Copies the data once. This currently has the same effect as :any:`to_py`.
         """
 
-    def tobytes(self):
+    def tobytes(self) -> bytes:
         """Convert the buffer to a bytes object.
 
         Copies the data once.
         """
 
-    def tobytearray(self):
+    def tobytearray(self) -> bytearray:
         """Convert the buffer to a bytearray object.
 
         Copies the data once.
         """
 
-    def tostring(self, encoding=None):
+    def tostring(self, encoding=None) -> str:
         """Convert the buffer to a string object.
 
         Copies the data twice.
--- a/src/py/_pyodide/_core_docs.py
+++ b/src/py/_pyodide/_core_docs.py
@@ -108,6 +108,29 @@ class JsProxy:
         an ArrayBuffer view.
         """
 
+    def tobytes(self):
+        """Convert the buffer to a bytes object.
+
+        Copies the data once.
+        """
+
+    def tobytearray(self):
+        """Convert the buffer to a bytearray object.
+
+        Copies the data once.
+        """
+
+    def tostring(self, encoding=None):
+        """Convert the buffer to a string object.
+
+        Copies the data twice.
+
+        The encoding argument will be passed to the Javascript ``TextDecoder``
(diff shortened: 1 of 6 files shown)


src/py/_pyodide/_core_docs.py
```diff
@@ -108,6 +108,36 @@ def assign_to(self, to: Any):
         an ArrayBuffer view.
         """
 
+    def tomemoryview(self) -> memoryview:
+        """Convert the buffer to a memoryview.
+
+        Copies the data once. This currently has the same effect as :any:`to_py`.
```
rth, line 114: Could you add a note that this and below methods are only defined for buffers? As right not it's not explicit.

rth (commented):
Thanks for this PR! A few comments,

- To be consistent with `JsProxy.to_py` it might be better to call them `.to_string`, `.to_bytes`, etc I think. Also people used to pandas would find this more familiar (e.g. https://pandas.pydata.org/docs/search.html?q=to_)
- Another thing that bothers me a bit is that I would expect `obj.to_string` to be equivalent to `str(obj)` and to be the equivalent of `PyProxy.toString`. However that is not the case here, is it? What happens if one calls `str(obj)` on a buffer, it calls the repr, right? I'm not sure what's the best action there, we might need to think about this consistency issue more. 

- Is `tobytesarray` really useful? If it's not used frequently, it might be better to remove it in favor of to_bytes + conversion to bytesarray even if it means 1 memory copy. Unless there are usecases where one would want to use bytesarray?
- It would be good to add a (or update an existing) test that those methods don't exist for non buffer objects.
- Also could be worth testing what the error is raised as expected for a buffer of bytes that can't be converted to a string.

hoodmane:
Thanks for the review!

> - To be consistent with `JsProxy.to_py` it might be better to call them `.to_string`, `.to_bytes`, etc I think. Also people used to pandas would find this [more familiar](https://pandas.pydata.org/docs/search.html?q=to_).

Done.

> - Another thing that bothers me a bit is that I would expect `obj.to_string` to be equivalent to `str(obj)`, and also similar to `PyProxy.toString`. However that is not the case here, is it? What happens if one calls `str(obj)` on a buffer, it calls the repr, right? I'm not sure what's the best action there, we might need to think about this consistency issue more. 

Yes. I think it should stay this way because making `str` not always call `toString` is weirder than having a `to_string` method that does something smarter some of the time. But I am not sure, can discuss more in the future.

- Is `tobytesarray` really useful? If it's not used frequently, it might be better to remove it in favor of to_bytes + conversion to bytesarray even if it means 1 memory copy. Unless there are usecases where one would want to use bytesarray?

Removed it, we can add it again if it turns out to be useful.

- It would be good to add a (or update an existing) test that those methods don't exist for non buffer objects.

Okay, can do this. I don't think we perform these tests at all currently. If somehow these methods did get added to nonbuffer objects they would access uninitialized data and be very crashy.

- Also could be worth testing what the error is raised as expected for a buffer of bytes that can't be converted to a string. 

Without [the `fatal` option to `TextDecoder`](https://developer.mozilla.org/en-US/docs/Web/API/TextDecoder/TextDecoder#parameters), `to_string` will never throw an error. Instead it will insert glitch characters at the site of the encoding problems. Maybe we should use this `fatal` option and test for encoding errors?

PR #1865, opened 2021-09-28
Add fetch_string and fetch_buffer

### LINKED ISSUE #1811: Add async open_url for binary files
Related to the discussion in https://github.com/pyodide/pyodide/issues/400#issuecomment-882101286 it would be really useful to have an async equivalent of open_url that would work for binary files.

The steps outlined in https://github.com/pyodide/pyodide/issues/400#issuecomment-882101102 are fairly complex, and it would help to put that into some helper function while ensuring that there are no memory leaks.

> Even better would be if we could find existing APIs that we can re-implement, but unfortunately most Python APIs are synchronous.

Yes, as far as I can tell one popular async library is [aiohttp](https://github.com/aio-libs/aiohttp) however it has a fairly complex API due to reasons outlined [here](https://docs.aiohttp.org/en/latest/http_request_lifecycle.html#why-is-aiohttp-client-api-that-way). We could try monkeypatching it to work in Pyodide, however another issue is that it has a bunch of dependencies and it would be better if it was possible to download/upload binary files with minimal dependencies in Pyodide. Since it's a basic operation.  

There was also https://github.com/spyoungtech/grequests but it looks like it's being phased out. 

I think just a thin wrapper around fetch would already be very useful,
```py
response = await pyodide_fetch(<url>)
stream = await response.body # returns bytes (or .content to take requests conventions)
```

with a name TBD. Or maybe even just a helper function for binary files,
```py
from js import fetch
from pyodide.utils import fetch_response_to_bytes

response = fetch(<url>)
stream = fetch_response_to_bytes(response)
```
since the other methods .json and .text should work with minimal effort as is.

hoodmane:
This resolves #1811 by adding a bunch of fetch wrappers that return various data types.


hoodmane:
How's it look to you now @rth?

hoodmane:
Currently the `json` api returns a Javascript json object. Maybe it would be better to use the Python JSON decoder to get a Python object instead?

rth:
No worries. Thanks for all you are doing for Pyodide! Good luck for the class: 350 students is massive. I hope you don't have to grade them manually :) Also we should really try to find more contributors.

PR #1866, opened 2021-09-29
Fix download path of f2c in buildf2c

owenthereal:
Fix download path of f2c in buildf2c


owenthereal:
Unsure about the cicleci failure. It looks unrelated to this PR

hoodmane:
Yeah failure is caused by #1860.

hoodmane:
Thanks!

PR #1868, opened 2021-09-30
Correct the wording about from JavaScript to Python, and vice versa

grimmer0125:
I think the original sentences probably are happened to be inversed. To speed up the flow, I open this PR directly. If my understanding is wrong, feel free to close this PR.


hoodmane:
Huh, thanks!

PR #1873, opened 2021-10-03
DOCS Fix the URL about debugging WebAssembly in Chrome [skip ci]

grimmer0125:
The original URL is broken or a typo.


rth (approved):
Thanks @grimmer0125 !

Issue #1572, opened 2021-05-03
Default compilation flags


rth:
To continue a discussion on Gitter started by @pmp-p ,

> did you test benefits of -O2 -g for real, because i rarely saw a speed improvement vs -Os -g0 (which give a serious size reduction and better jit in v8 )

In my tests some time ago `-O3` was faster by 10% and made the overall download size by 10% larger or so as compared to `-Os`, if I remember correctly. 

However there is more an more evidence that `-Os` would be preferable  https://github.com/pyodide/pyodide/issues/1541 so I agree that we should probably switch to it.

Regarding `-g` vs `-g0` I have not tried it, it would make stack traces less readable right? We should investigate how much size it would save in Pyodide.

carrycooldude:
Love to work on this @rth

hoodmane:
I think this issue is a bit out of date...

ryanking13:
Well, I think we still don't have a good benchmark comparing the effect of compilation flags (or at least it is not documented anywhere), so contributions are welcome.

PR #1882, opened 2021-10-09
Size optimization compilation options

rth:
Partially addresses https://github.com/pyodide/pyodide/issues/1572

Switches from -O2 to -Os as the default compilation option.

| Size (MB)        | main (-O2 -g) | -Os -g | -Os -g0 | -Oz -g0 |
|------------------|---------------|--------|---------|---------|
| pyodide.asm.wasm | 12            | 9.4    | 9.4     | 8.3     |
| pyodide.asm.data | 5.2           | 5.2    | 5.2     | 5.2     |
| numpy.data       | 6.8           | 6.2    | 6.2     |         |

Performance impact needs to be checked. 
I'm a bit surprised that -g0 option has no effect, it's possible that it's not correctly taken into account. Also, it would necessary to check the logs more carefully, in some places even the -O option might not be passed. In particular, I find it suspicious that the size of pyodide.asm.data is not changed.

### DIFF
--- a/Makefile.envs
+++ b/Makefile.envs
@@ -74,7 +74,6 @@ export LDFLAGS_BASE=\
 	-s EXPORT_ALL=1 \
 	-s WASM=1 \
 	-std=c++14 \
-	-s ASSERTIONS=1 \
 	-s LZ4=1 \
 	-L $(CPYTHONROOT)/installs/python-$(PYVERSION)/lib/ \
 	$(EXTRA_LDFLAGS)
--- a/src/core/python2js.c
+++ b/src/core/python2js.c
@@ -600,7 +600,6 @@ python2js_custom(PyObject* x,
                  JsRef dict_converter,
                  JsRef default_converter)
 {
-
   PyObject* cache = PyDict_New();
   if (cache == NULL) {
     return NULL;
@@ -699,12 +698,6 @@ to_js(PyObject* self,
                                     &pyproxies,
                                     &py_dict_converter,
                                     &py_default_converter)) {
-    return NULL;
-  }
-
-  if (obj == NULL) {
-    PyErr_SetString(PyExc_TypeError,
-                    "Call to _PyArg_ParseStackAndKeywords failed");
     return NULL;
   }
 


rth:
So now I get  a fatal error in `JsArray_Push_unchecked` when running the CPython test suite, and in particular `test_reentrant_error` in Firefox, GH Actions. Related to the earlier discussion in #1950 that it's unlikely to fail :) I'll need to rebuild with assertions to see what's going on.
Main branch is still green though
```
E           conftest.JavascriptException: PythonError: Traceback (most recent call last):
E             File "???.js", line -1, in <javascript frames>
E           JsException: RuntimeError: abort(Assertion failed: callable != NULL, at: /home/runner/work/pyodide/pyodide/cpython/installs/python-3.9.5/include/python3.9/cpython/abstract.h,72,PyVectorcall_Function). Build with -s ASSERTIONS=1 for more info.
E           
E           
E           PythonError@http://127.0.0.1:55517/pyodide.asm.js:14:220802
E           new_error@http://127.0.0.1:55517/pyodide.asm.js:14:249825
E           @http://127.0.0.1:55517/pyodide.asm.wasm:wasm-function[308]:0x1bc84e
E           _createPyodideModule/Module._wrap_exception@http://127.0.0.1:55517/pyodide.asm.js:14:945198
E           hiwire_init/Module.hiwire.get_value@http://127.0.0.1:55517/pyodide.asm.js:14:235841
E           JsArray_Push_unchecked@http://127.0.0.1:55517/pyodide.asm.js:14:211397
[..]

conftest.py:224: JavascriptException
--------------------------- Captured stdout teardown ---------------------------
Assertion failed: callable != NULL, at: /home/runner/work/pyodide/pyodide/cpython/installs/python-3.9.5/include/python3.9/cpython/abstract.h,72,PyVectorcall_Function
Internal error: Argument '0' to hiwire.get_value is falsy. This was probably because the Python error indicator was set when get_value was called. The Python error that caused this was:
JsException: RuntimeError: abort(Assertion failed: callable != NULL, at: /home/runner/work/pyodide/pyodide/cpython/installs/python-3.9.5/include/python3.9/cpython/abstract.h,72,PyVectorcall_Function). Build with -s ASSERTIONS=1 for more info.
````

cc @hoodmane

**Comment by hoodmane** [comment]
Something is definitely going wrong here with the error handling, but it doesn't really have anything to do with `JsArray_Push`. Something went wrong inside of the conversion code, and a null pointer was handed to `_python2js_deep`. It did a bunch of stuff with it and didn't notice it had a null pointer, and decided to wrap it in a `pyproxy` so called `pyproxy_new`. `pyproxy_new` starts by doing type inspection on the object with `pyproxy_getflags`. `pyproxy_getflags` also doesn't check if the proxy is null and passes the null pointer to all sorts of different methods that are supposed to panic if the argument is NULL but none of them do until it gets to `_PyVectorcall_Function` which has an assert that its argument is not NULL.

So a couple of changes that could be made here:
1. If `hiwire_get_value` is called on a null pointer, we should panic instead of raising a nonfatal error.
2. Probably more of our functions should make defensive checks that their argument is not NULL.

Not sure where the null pointer is coming from to begin with though.

**Comment by hoodmane** [comment]
Another annoying thing is that when there is a fatal error, we call `Py_DumpStack` and this produces super super helpful diagnostic info, but for whatever reason selenium doesn't pick that information up at all.

**Comment by rth** [comment]
Looking into this more, what appears to be happening is that `_PyArg_ParseStackAndKeywords` (https://github.com/pyodide/pyodide/blob/main/src/core/python2js.c#L691) didn't set `obj` and so it's kept as NULL (the initial value), which is then passed to `python2js_custom` then calls `_python2js` and `_python2js_deep` which never check for NULL.

For instance, with changes in this PR it would fail earlier.

The `_PyArg_ParseStackAndKeywords` looks a bit obscure to me, we are checking for the exit status, same as in CPython code as far as I can tell, but the exit status indicates success. I'm not sure under which conditions it wouldn't set the `obj` object (or set it to `NULL`). Any ideas @hoodmane ? 
BTW, we couldn't use public CPython API there?

The above happens when we are making the 
https://github.com/pyodide/pyodide/blob/66fe7ede188856c03d7fc5d4160f5529c5a442fb/src/js/load-package.ts#L176

call with a buffer of 983040 bytes for filename='distutils.tar'.

**Comment by hoodmane** [comment]
Note that they want to stabilize some version of `_PyArg_ParseStack` but there are various design considerations and they haven't gotten around to it yet:
https://bugs.python.org/issue44214

**Comment by hoodmane** [comment]
> BTW, we couldn't use public CPython API there?

Perhaps the easiest fix would be to switch `to_js` to `METH_VARARGS | METH_KEYWORDS`. This is slower but `to_js` isn't a super hot function or anything and probably in any case the actual behavior of `to_js` should dwarf the argument parsing. Though Python's call protocols are a remarkable snarl with layers and layers of indirection and `METH_FASTCALL` reduces the number of layers of indirection (this may also be better in Python v3.10). But in any case, it would fix this crash.

> I'm not sure under which conditions it wouldn't set the obj object (or set it to NULL). Any ideas @hoodmane ?

It shouldn't be allowed to do that. `obj` is a mandatory argument so the contract of the function is it either sets `obj` to not `NULL` or it returns an error.

**Comment by hoodmane** [comment]
`_PyArg_ParseStackAndKeywords` is never called:
```C
if (kwnames != NULL && !_PyArg_ParseStackAndKeywords(args,
```

hoodmane:
It is also fine to call `_PyArg_ParseStackAndKeywords` with `kwnames` null:
https://github.com/python/cpython/blob/main/Python/getargs.c#L2063
```C

    if (kwargs != NULL) {
        nkwargs = PyDict_GET_SIZE(kwargs);
    }
    else if (kwnames != NULL) {
        nkwargs = PyTuple_GET_SIZE(kwnames);
        kwstack = args + nargs;
    }
    else {
        nkwargs = 0;
    }
```

rth:
> Note that they want to stabilize some version of _PyArg_ParseStack

Good to know.

> This is slower but to_js isn't a super hot function or anything and probably in any case the actual behavior of to_js should dwarf the argument parsing. Though Python's call protocols are a remarkable snarl with layers and layers of indirection and METH_FASTCALL reduces the number of layers of indirection (this may also be better in Python v3.10).

Yeah, well we can probably keep it as is, now that the issue is resolved, though their API is certainly not very user friendly for devs.

> _PyArg_ParseStackAndKeywords is never called:

So your fix did work, but actually thinking more about it, I don't understand how this could have been the issue here. Indeed if `kwnames == NULL` then `_PyArg_ParseStackAndKeywords` is never called. However in that case we are returning `NULL` in that if condition (and presumably erroring), and so it shouldn't have been able to error in the code located after that return statement.

Anyway, I does work now.

There are still a few test failures mostly around signals/interrupt not working as expected, with the following messages,
```
 +  'sigaction: signal type not supported: this is a no-op.',
  +  'Calling stub instead of siginterrupt()',
```
looking into that. It's probably not a bad idea to compile with different optimization options occasionally to discover latent bugs.

hoodmane:
> if kwnames == NULL then _PyArg_ParseStackAndKeywords is never called. However in that case we are returning NULL in that if condition (and presumably erroring), and so it shouldn't have been able to error in the code located after that return statement.

It's hard to read the code here because it is nonsense but look carefully:

```C
if (kwnames != NULL && !_PyArg_ParseStackAndKeywords(args,...)){
    return NULL;
}
```

rth:
Ahh, yes, of course :)

rth:
This very simple change keeps eluding me. When built with -Os interrupt test fails. So as discussed in https://github.com/pyodide/pyodide/pull/2405#issuecomment-1101580124 the idea was to build `Python/ceval.c` and `Modules/signalmodule.c` with `-O2` since it's more performance sensitive.

I tried setting `-O2`
 - in CPython PY_STDMODULE_CFLAGS (which is supposed to be specific to building the interpreter object files)
 - in CPython CFLAGS in general
 - in MAIN_MODULE_CFLAGS, and MAIN_MODULE_LDFLAGS and it still fails.

I'm probably missing something obvious. I'll try to get back to it later.

PR #1883, opened 2021-10-10
Drop Python built-in pwd module support

### LINKED ISSUE #1857: Error while writing downloaded *.tar.gz file
Hi, In below code, I am trying to download a file and write it. but I get TypeError: a bytes-like object is required, not 'pyodide.JsProxy' . Can you help me with writing contents of downloaded file ?

from io import BytesIO
from js import XMLHttpRequest

url = 'URL to Tar gz (*.tar.gz) File'
req = XMLHttpRequest.new()
req.open("GET", url, True)
req.responseType = "arraybuffer"

def callback(e):
    if req.readyState == 4:
        with open('/weights.tar.gz', 'wb') as fh:
            **fh.write(BytesIO(req.response).getvalue())**
        print('done writing')
        file_size = os.path.getsize('/weights.tar.gz')
        print("File Size is :", file_size, "bytes")
req.onreadystatechange = callback
req.send(None)

ryanking13:
This PR drops Python built-in pwd module support. 
The pwd module is a [Unix specific module](https://docs.python.org/3/library/unix.html) which provides access to unix user account and passwd database.

Emscripten [does not support pwd related syscalls](https://github.com/emscripten-core/emscripten/blob/aab919ddfe92e88c8f1d67159d41e2291af7a1dd/src/library.js#L2546) (there is no user account in browser  environment),
so currently calling pwd methods raises an error:

image

As CPython supports non-Unix environments (e.g. Windows), I believe dropping pwd module will cause no bad side effects.
I looked for some cases where CPython import pwd module. It handles error when pwd module is not available:
- [pathlib](https://github.com/python/cpython/blob/56c1f6d7edad454f382d3ecb8cdcff24ac898a50/Lib/pathlib.py#L342-L345)
- [shutil](https://github.com/python/cpython/blob/b7eac52b466f697d3e89f47508e0df0196a98970/Lib/shutil.py#L868-L871)
- [http](https://github.com/python/cpython/blob/17b16e13bb444001534ed6fccb459084596c8bcf/Lib/http/server.py#L945-L948)

Note that other Unix specific modules like `grp`, `spwd`, `syslog` are already not included in pyodide.

This resolves #1857, so I think this is a fix.

### DIFF
--- a/docs/usage/wasm-constraints.md
+++ b/docs/usage/wasm-constraints.md
@@ -32,6 +32,7 @@ since they currently wouldn't work in the WebAssembly VM,
 - turtle.py
 - turtledemo
 - venv
+- pwd
 
 ### Included but not working modules
 


hoodmane:
Thanks @ryanking13! 

I was wondering if you could use `Setup.local` to do this rather than patching `Setup` but I found the following text in the docs:

> you'll still have to edit Setup if you want to disable modules that are enabled by default, or if such modules need non-standard options.

Not sure if this is current info but I think it is and it means this is the right way to get rid of `pwd`.

ryanking13:
@hoodmane Thanks for the suggestion!

>I was wondering if you could use Setup.local to do this rather than patching Setup

Maybe it is possible with a custom Setup.local file with `disabled` option.

I'll try if it is possible.

```sh
# https://github.com/python/cpython/blob/3d1ca867ed0e3ae343166806f8ddd9739e568ab4/Modules/Setup#L19-L21
# Modules are not built at all, not by the Makefile,
# nor by the setup.py script, when they are preceded by a "*disabled*" tag
# line.
```

Issue #1875, opened 2021-10-04
./node_modules/pyodide/load-pyodide.js Module not found: Can't resolve 'fs/promises'., pyodide npm 0.18.2


grimmer0125:
I create a [create react app](https://github.com/facebook/create-react-app) project, and `yarn add pyodide` to test `pyodide` import function. It will throw the error message when using `yarn start` (as long as I add `await import` that line) , the result is `Failed to compile.` The below is the added code on the bootstrap project.  

node: v15.14.0
yarn: 1.22.10
ref repo: https://github.com/grimmer0125/test_pyodide_npm/blob/master/src/App.js

```
async function test() {
  let pyodide_pkg = await import("pyodide/pyodide.js"); 
}

function App() {
  useEffect(() => {
    async function init() {
      await test()
    }
    init();
  }, []);
```

grimmer0125:
Duplicated with #1863

rth:
@grimmer0125 Should we close this issue now?

grimmer0125:
Yes, I'm going to close this issue. I have tested #1900 on another famous front-end framework, [angular](https://angular.io/) and it works too.

TatyanaVl:
I get this issue as well and its not clear how the issue was solved.

PR #1900, opened 2021-10-24
Fix webpack compatibility issue

grimmer0125:
This is to solve 
1) #1875 for Creat React App environment and **2.** environment
2) Fix webpack building error when using `npx webpack` with its default and minimal setting for your front-end code. Error: import("path") and import("vm") & **1.**
3) after fixing **1 & 2**, `import` of loadScript` in the browser will throw `module not found` error since the `import` function is already polluted by `webpack`. It is a runtime issue. 

Besides this PR, `Pydodie npm` also needs #1849 to make it work.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -224,6 +224,10 @@ jobs:
             npx tsd
             npm install
             npm test
+            git clone -b test_circle_meant_error --single-branch https://github.com/grimmer0125/webpack-pyodide-demo.git
+            cd webpack-pyodide-demo
+            npm install 
+            npx webpack
 
   benchmark:
     <<: *defaults
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -224,6 +224,9 @@ jobs:
             npx tsd
             npm install
             npm test
+      - run:
+          name: check if webpack cli works well with load-pyodide.js
+          command: | 
             git clone https://github.com/grimmer0125/webpack-pyodide-demo.git
             cd webpack-pyodide-demo
             git checkout test_circle_meant_error
@@ -231,6 +234,8 @@ jobs:
             cp ../load-pyodide.js node_modules/pyodide/load-pyodide.js                  
             head -20 node_modules/pyodide/load-pyodide.js
             npx webpack
+            cd ..
+            rm -rf webpack-pyodide-demo          
 
   benchmark:
     <<: *defaults
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -224,9 +224,12 @@ jobs:
             npx tsd
             npm install
             npm test
-            git clone -b test_circle_meant_error --single-branch https://github.com/grimmer0125/webpack-pyodide-demo.git
+            git clone https://github.com/grimmer0125/webpack-pyodide-demo.git
             cd webpack-pyodide-demo
-            npm install 
+            git checkout test_circle_meant_error
+            npm install     
+            cp ../load-pyodide.js node_modules/pyodide/load-pyodide.js                  
+            head -20 node_modules/pyodide/load-pyodide.js
             npx webpack
 
   benchmark:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -227,15 +227,14 @@ jobs:
       - run:     
           name: check if webpack cli works well with load-pyodide.js
           command: | 
-            git clone https://github.com/grimmer0125/webpack-pyodide-demo.git
-            cd webpack-pyodide-demo
-            git checkout test_circle_meant_error
+            git clone https://github.com/pyodide/pyodide-webpack-example.git
+            cd pyodide-webpack-example
             npm install     
             cp ../src/js/load-pyodide.js node_modules/pyodide/load-pyodide.js                  
             head -20 node_modules/pyodide/load-pyodide.js
             npx webpack
             cd ..
-            rm -rf webpack-pyodide-demo          
+            rm -rf pyodide-webpack-example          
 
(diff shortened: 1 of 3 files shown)


grimmer0125:
> Could you clone your repo and build it with webpack in as [a new run step in test-js](https://github.com/pyodide/pyodide/blob/main/.circleci/config.yml#L226) ?

I have tried to add the failure case (before the fix) and switched to the successful case (after the fix). Currently, I just copy the fixed `load-script.js` file from this pyodide repo into the cloned repo. It can change to use symbolic link the whole folder or new published pyodide package. 

I think the following steps are
1. organize the commits of my repo 
2. move it in Pyodide organization 
3. Change the used GitHub URL in the `config.yml` in this PR 
4. Merge the PR. 
5. Release a new npm 
6. (optional) change to use new pyodide npm package instead 

@rth does this make sense? 

The test on the fixed runtime issue (`loadScript = async (url) => await import(/* webpackIgnore: true */ url);` could be another PR if need.

grimmer0125:
> @grimmer0125 I think we prefer if you use merge rather than rebase because the force pushes confuse the github review tools.

Got it. I forget I can use other branches to do the circleci experienemcne.

hoodmane:
> I forget I can use other branches to do the circleci experiments.

Well if there is a good reason force pushes are okay. Just try to avoid if practical.

.circleci/config.yml
```diff
@@ -224,6 +224,18 @@ jobs:
...
+          command: | 
+            git clone https://github.com/grimmer0125/webpack-pyodide-demo.git
+            cd webpack-pyodide-demo
+            git checkout test_circle_meant_error
+            npm install     
+            cp ../src/js/load-pyodide.js node_modules/pyodide/load-pyodide.js                  
+            head -20 node_modules/pyodide/load-pyodide.js
+            npx webpack
+            cd ..
+            rm -rf webpack-pyodide-demo
```
rth, line 238:
```suggestion         
```
No need for this, the docker instance will be destroyed anyway.

rth:
Thanks a lot for doing this @grimmer0125 !

> organize the commits of my repo

I wouldn't worry too much about that, people don't look at commits history that much, but as you prefer. What we would also need however is to add a README explaining what that this repo is and if it's only for testing or if users can use it as an example. Also please add a LICENSE. The readme should also explain why there is a `webpack-pyodide-demo/node_modules/pyodide/load-pyodide.js` file and I assume that it would be removed after the next Pyodide release? Maybe let's add those via a PR so we can discuss it there?

> move it in Pyodide organization

Could you please add me as as admin to that repo? I'll move it? You will still keep your permissions on it.

grimmer0125:
> I wouldn't worry too much about that, people don't look at commits history that much, but as you prefer. What we would also need however is to add a README explaining what that this repo is and if it's only for testing or if users can use it as an example. Also please add a LICENSE. The readme should also explain why there is a webpack-pyodide-demo/node_modules/pyodide/load-pyodide.js file and I assume that it would be removed after the next Pyodide release? Maybe let's add those via a PR so we can discuss it there?

I've added README, LICENSE file, and note it is a minimal example project. To avoid things complicated, I removed the committed `load-pyodide.js` and it should be installed from npm package. So now this project does not work byt itself. And the CircleCI will copy `load-pyodide.js` to the downloaded [pyodide-webpack-example] (https://github.com/grimmer0125/pyodide-webpack-example) to make the test sucessfully. If you think it is better to keep `pyodide-webpack-example` running successfully by default now and we can add the fixed `load-pyodide.js` back.

>  Could you please add me as as admin to that repo? I'll move it? You will still keep your permissions on it.
I've added you as a collaborator of my [pyodide-webpack-example project](https://github.com/grimmer0125/pyodide-webpack-example).

rth:
> I've added you as a collaborator of my pyodide-webpack-example project.

I received the invite, but that only gives write permission, not admin ones so I'm not able to move it. Other maybe could you transfer it to me and I'll transfer it there? 

>  After a new Pyodide npm package is released, pyodide-webpack-example and Pyodide CircleCI can just use npm version of Pyodide to run well. 

Yes, let's do that. We can very well release a 0.18.3 to npm just with those fixes, without waiting for a complete Pyodide release. Though then the question how to keep a changelog in those conditions.

grimmer0125:
> I received the invite, but that only gives write permission, not admin ones so I'm not able to move it. Other maybe could you transfer it to me and I'll transfer it there?

I did not find out any admin role setting in a personal repo and it seems to only exist in a GitHub organization. I've started to transfer it to you and please check.

rth:
@grimmer0125 the repository was moved.

rth:
Please also add a changelog entry (preferable in a new commit without force-pushing :) )  . Otherwise LGTM. Thanks!

.circleci/config.yml
```diff
@@ -224,6 +224,17 @@ jobs:
...
+          name: check if webpack cli works well with load-pyodide.js
+          command: | 
+            git clone https://github.com/pyodide/pyodide-webpack-example.git
+            cd pyodide-webpack-example
+            npm install     
+            cp ../src/js/load-pyodide.js node_modules/pyodide/load-pyodide.js                  
+            head -20 node_modules/pyodide/load-pyodide.js
+            npx webpack
+            cd ..
+            rm -rf pyodide-webpack-example
```
rth, line 237:
```suggestion
   
```

grimmer0125:
I have added a changlog.

hoodmane (approved):

hoodmane:
Thanks @grimmer0125!

Issue #537, opened 2019-10-15
statsmodels crashes on import


mdboom:
```
import statsmodels.api
```

```
Error: Traceback (most recent call last):
  File "/lib/python3.7/site-packages/pyodide.py", line 42, in eval_code
    exec(compile(mod, '<exec>', mode='exec'), ns, ns)
  File "<exec>", line 1, in <module>
  File "/lib/python3.7/site-packages/statsmodels/api.py", line 7, in <module>
    from .regression.recursive_ls import RecursiveLS
  File "/lib/python3.7/site-packages/statsmodels/regression/recursive_ls.py", line 16, in <module>
    from statsmodels.tsa.statespace.mlemodel import (
  File "/lib/python3.7/site-packages/statsmodels/tsa/statespace/mlemodel.py", line 18, in <module>
    from .simulation_smoother import SimulationSmoother
  File "/lib/python3.7/site-packages/statsmodels/tsa/statespace/simulation_smoother.py", line 10, in <module>
    from .kalman_smoother import KalmanSmoother
  File "/lib/python3.7/site-packages/statsmodels/tsa/statespace/kalman_smoother.py", line 11, in <module>
    from statsmodels.tsa.statespace.representation import OptionWrapper
  File "/lib/python3.7/site-packages/statsmodels/tsa/statespace/representation.py", line 10, in <module>
    from .tools import (
  File "/lib/python3.7/site-packages/statsmodels/tsa/statespace/tools.py", line 207, in <module>
    set_mode(compatibility=None)
  File "/lib/python3.7/site-packages/statsmodels/tsa/statespace/tools.py", line 59, in set_mode
    from . import (_representation, _kalman_filter, _kalman_smoother,
ImportError: cannot import name '_representation' from 'statsmodels.tsa.statespace' (/lib/python3.7/site-packages/statsmodels/tsa/statespace/__init__.py)
```

marcpabst:
I can reproduce this with the latest release. Statsmodels would be really nice...

machow:
Hey, I think it's related to how statsmodels v0.9 checks for / builds against scipy on install.

Basically, `setup.py` tries `from scipy.linalg import cython_blas`, and if that fails then excludes certain files. This is normally okay, since on import time, statsmodels tries the same scipy import again. If this fails, then it enters "compatibility" mode and uses pure python implementations.

[...]

PR #1908, opened 2021-10-25
Patch statsmodels to enforce compatibility mode

marcpabst:
As discussed in #537, I added a small patch that forces `statsmodels` into compatibility mode to avoid it crashing upon import. 

However, I'm experiencing some problems building pyodide at the moment (even with the docker image) so I'd appreciate if anyone other than me could give it a try.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -28,3 +28,4 @@ packages/.artifacts
 packages/*/build.log*
 packages/build-logs
 dist/
+.DS_Store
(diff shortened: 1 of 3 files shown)


marcpabst:
No problem, I'll just create a new patch!

gitkwr:
Checking for updates

hoodmane:
I think this can be closed now because #2073 resolved #537. Thanks for your help @marcpabst!

PR #1912, opened 2021-10-27
CI Build for MacOS

rth:
Follow up on https://github.com/pyodide/pyodide/pull/1690 with the addition of CI for MacOS 

Among other things this aims to deal with small incompatibilities in `patch` and `sed` on MacOS.

Should integrate changes from https://github.com/pyodide/pyodide/pull/1770 in the future (or add this setup there).


rth:
Continued in https://github.com/pyodide/pyodide/pull/3602

Issue #1899, opened 2021-10-24
micropip import error


eyalk11:
After building pyodide, I tried the following code : 
```
            pyodide.loadPackage("micropip").then(() => {
                pyodide.runPythonAsync(`
                    import micropip
                `);
            });
```
```
pyodide.asm.js:14 Uncaught (in promise) PythonError: Traceback (most recent call last):
  File "/lib/python3.9/asyncio/futures.py", line 201, in result
    raise self._exception
  File "/lib/python3.9/asyncio/tasks.py", line 256, in __step
    result = coro.send(None)
  File "/lib/python3.9/site-packages/_pyodide/_base.py", line 494, in eval_code_async
    await CodeRunner(
  File "/lib/python3.9/site-packages/_pyodide/_base.py", line 345, in run_async
    coroutine = eval(self.code, globals, locals)
  File "<exec>", line 2, in <module>
ModuleNotFoundError: No module named 'micropip'

    at new_error (pyodide.asm.js:14)
    at pyodide.asm.wasm:0x1b9399
    at pyodide.asm.wasm:0x1bd7d6
    at pyodide.asm.wasm:0x853699
    at pyodide.asm.wasm:0x1f73c1
    at pyodide.asm.wasm:0x2f2f19
    at pyodide.asm.wasm:0x87250d
    at pyodide.asm.wasm:0x23ed39
    at pyodide.asm.wasm:0x8708c8
    at pyodide.asm.wasm:0x1f7b76
```

I built just some of packages, because of some errors. But `./packages/micropip/build/micropip-0.1/micropip` exists.

ryanking13:
It seems like multiple Python versions (Python 3.9 and Python 3.8) are installed in your system, and Python 3.8 is used when building Python packages instead of Python 3.9, which is a bug.

I think we can prevent this by strictly checking that `pyodide-build` is using the same Python version of what was used to build Pyodide.

https://github.com/pyodide/pyodide/blob/4042db26158c5c78cc8be6313927c29e5bba9181/pyodide-build/pyodide_build/__main__.py#L1

ryanking13:
I believe this issue is resolved with #1924.
@eyalk11 I would be very grateful if you could help us verify this.

eyalk11:
Hi, I tried this, and it didn't solve it.

ryanking13:
> Hi, I tried this, and it didn't solve it. My previous mitigation worked.

Hmm... I can't reproduce it.

Could you try 

```
[...]

PR #1924, opened 2021-11-03
FIX Explicitly use host Python version when building packages

ryanking13:
This change prevents the use of unintended Python versions when building packages.

Related: #1899

### DIFF
--- a/packages/Makefile
+++ b/packages/Makefile
@@ -1,6 +1,7 @@
 export PYODIDE_ROOT=$(abspath ..)
 PYODIDE_LIBRARIES=$(abspath ./.artifacts)
 CPYTHONROOT=$(PYODIDE_ROOT)/cpython
+HOSTPYTHON=$${HOSTPYTHON:-python3.9}
 include ../Makefile.envs
 
 export NUMPY_LIB=$(PYODIDE_ROOT)/packages/numpy/build/numpy-1.17.5/install/lib/python$(PYMAJOR).$(PYMINOR)/site-packages/numpy/core/lib/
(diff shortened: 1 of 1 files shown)


packages/Makefile
```diff
@@ -1,6 +1,7 @@
 export PYODIDE_ROOT=$(abspath ..)
 PYODIDE_LIBRARIES=$(abspath ./.artifacts)
 CPYTHONROOT=$(PYODIDE_ROOT)/cpython
+HOSTPYTHON=$${HOSTPYTHON:-python3.9}
```
rth, line 4: It's already defined in Makefile.envs I believe

https://github.com/pyodide/pyodide/blob/b205f8e6bd176b02e701407099843aa8900fe123/Makefile.envs#L19

Why would we need to re-define it here, particularly before including the Makefile.envs ?

rth (commented):
Minor comment, otherwise LGTM. Thanks @ryanking13 !

ryanking13:
> Why would we need to re-define it here, particularly before including the Makefile.envs ?

Oh, thanks for pointing that out. It was for a test but I forgot to remove it before the commit. Removed it.

ryanking13:
Hmm... It looks like the numpy build is failing in Github actions.

rth:
Due to setuptools I think: https://github.com/pyodide/pyodide/issues/1930

ryanking13:
> Due to setuptools I think: #1930

Indeed. https://github.com/numpy/numpy/pull/19173#issuecomment-960595766

rth:
Could you also add me and Hood as collaborators to your fork CI https://app.circleci.com/pipelines/github/ryanking13/pyodide/ so we can restart jobs? In particular, it's strange that there are so many firefox package test failures and I wanted to restart it.

ryanking13:
> Could you also add me and Hood as collaborators to your fork CI https://app.circleci.com/pipelines/github/ryanking13/pyodide/ so we can restart jobs? In particular, it's strange that there are so many firefox package test failures and I wanted to restart it.

I am having trouble finding a way to add a collaborator to my fork CI. Do you know how?
Anyway, I've restarted CI.

https://app.circleci.com/pipelines/github/ryanking13/pyodide?branch=packaging%2Fpython-version

rth:
> I guess giving write permission to my fork repository would give access to fork CI as well?

Yes, I think that was the solution. Anyway it was just a though it's not critical, it should get better with larger CircleCI runners.

PR #1929, opened 2021-11-03
CI Use larger ressource classes for CircleCI

### LINKED ISSUE #1928: Use larger CircleCI resource classes for chrome tests?
It looks like CircleCI has made Medium+ and Large resource classes available on free plans (their docs are currently inconsistent on this though). Maybe if we increased the size of the chrome tests then the tests would fail less frequently.

rth:
Closes #1928

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -39,7 +39,6 @@ jobs:
 
   build-core:
     <<: *defaults
-    resource_class: medium+
     steps:
       - checkout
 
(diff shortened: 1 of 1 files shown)


rth:
Due to the way CI is currently setup, I can only change the ressource type for all tests, so I updated it medium+. Since the bottleneck is wasm compilation and it's multi-threaded, it shouldn't hurt for Firefox as well. Here the test CI was green twice in a row. Let's try this, we should have enough CircleCI credits anyway.

Issue #1827, opened 2021-09-11
Upgrade numpy version to 1.20+


leopsidom:
Currently pyodide has numpy at version 1.17.5 compared to latest numpy release @ 1.21.2. Pandas at version 1.0.5 while the latest release is 1.3.2. Should we bump the versions for these packages ? 

PS: I tried to have a build on the latest numpy. It seems we are able to remove 2 patches with the upgrade due to upstream fix:

```
patches/add-emscripten-cpu.patch
patches/rm-duplicate-symbols-link.patch
```

I was able to fix some patches, and got it start building. But still getting some issues: 

> emcc: error: Passing any of -msse, -msse2, -msse3, -mssse3, -msse4.1, -msse4.2, -mavx, -mfpu=neon flags also requires passing -msimd128!

hoodmane:
Thanks for looking into this @leopsidom! 

> Should we bump the versions for these packages?

Yeah, if you do manage to update any of these a PR is quite welcome. See also:
https://github.com/pyodide/pyodide/issues/1464
It would maybe be helpful to update that issue.

leopsidom:
I'd be glad to take a look. But I have minimum knowledge of numpy source code and emscrypten, so not sure how I should estimate the effort.

leopsidom:
Just to add some update, I did  some bisect analysis on different numpy versions. The latest numpy that can build is 1.18.5. Since 1.19.0, we had some issues. For 1.20+ build, I'm currently getting  this issue:

> numpy/core/src/common/simd/avx512/avx512.h:54:18: error: unknown type name '__m512'; did you mean '__m128'?
typedef struct { __m512  val[3]; } npyv_f32x3;

I think it has something to do with numpy adding support for  avx512  architecture since 1.19.0: https://github.com/numpy/numpy/commit/5562a8c93fe18f0a51d9051f0c25c7cf525312fe

I cut an issue to numpy here: https://github.com/numpy/numpy/issues/19892

dynamicwebpaige:
Adding a +1 for updating the `pandas` version in Pyodide! 

There are several nice new features in v1.2 and v1.3 (including [table visualizations](https://pandas.pydata.org/pandas-docs/stable/user_guide/style.html) and exporting to LaTeX).

rth:
Pandas was updated in https://github.com/pyodide/pyodide/pull/1922: just bumping the version worked .

[...]

PR #1934, opened 2021-11-07
Upgrade numpy to 1.21.4

leopsidom:
### Description

Currently pyodide has numpy @1.17.5 but the newest numpy version is 1.21.4. This PR is to keep the numpy package in pyodide up to date.  Some changes made in this PR include:

1. patches removed: `add-emscripten-cpu.patch` and `rm-duplicate-symbols-link.patch`. These patches got removed because the underlying issue has been fixed in the numpy package.
2. patches added: 
     1. `fix-invalid-asm-instruction.patch` to fix invalid asm instructions
     2. `limit-cpu-baseline.patch` this is to limit cpu baseline to `AVX` so it doesn't try to build SIMD feature AVX512 that emscripten doesn't support;
     3. `not-build-lapack-lite-as-64-bit.patch` this is to prevent building lapack-lite as 64 bit integer mode which emscripten doesn't support;
3. patches modified (both are needed because the original files got changed and the original patches can no longer apply):
    1.  `fix-static-init-of-nditer-pywrap.patch`
    2. `use-local-blas-lapack.patch`

Notes: in numpy build file, I'm currently setting:
```
self.cpu_baseline = "avx"
self.cpu_dispatch = "avx"
```
I think this means there are still basic SIMD optimization built with this which is supported by emscripten. We might want to check if we want to enable it yet.

Issue: https://github.com/pyodide/pyodide/issues/1827

### Checklists

- [ ] Add a [CHANGELOG](https://github.com/pyodide/pyodide/blob/main/docs/project/changelog.md) entry or explain why an entry is not needed
- [ ] Add / update tests


packages/numpy/patches/disable_optimization.patch
```diff
@@ -0,0 +1,13 @@
+diff --git a/numpy/distutils/command/build.py b/numpy/distutils/command/build.py
```
leopsidom: Unintentionally included this file. This patch is not needed.

ryanking13:
This must've been a hard job. I appreciate your effort!

A minor comments:

1.

> * patches added:
>   
>   1. `fix-invalid-asm-instruction.patch` to fix invalid asm instructions
>   2. `limit-cpu-baseline.patch` this is to limit cpu baseline to `AVX` so it doesn't try to build SIMD feature AVX512 that emscripten doesn't support;
>   3. `not-build-lapack-lite-as-64-bit.patch` this is to prevent building lapack-lite as 64 bit integer mode which emscripten doesn't support;

I think adding these descriptions into the patch file would help other people who later see patch files.

2. It would be great if we add some tests for newly available features of numpy 1.21 (and if possible, 1.18, 1.19, 1.20 also). I'm not sure which are the core new functionalities of these numpy versions, but we could find out from their [release notes](https://numpy.org/devdocs/release.html).

rth:
Thanks a lot for doing this work @leopsidom !

> I think this means there are still basic SIMD optimization built with this which is supported by emscripten. We might want to check if we want to enable it yet.

Right now the CI fails with,
```
emcc: error: Passing any of -msse, -msse2, -msse3, -mssse3, -msse4.1, -msse4.2, -mavx, -mfpu=neon flags also requires passing -msimd128!
```
which seems to be expected as discussed in https://github.com/emscripten-core/emscripten/issues/12714. I'm very in favor of investigating what we need to use SIMD a separate issue (and in particular check that it doesn't break Safari), for now maybe the easiest would be,
```
self.cpu_baseline = "min"
self.cpu_dispatch = "min"
``` 
to disable simd ?

>> Add / update tests
>
> It would be great if we add some tests for newly available features of numpy 1.21 (and if possible, 1.18, 1.19, 1.20 also). 

This is another subject we should improve on in Pyodide. Historically we have not run much tests for packages, because while have the ability to run the package test suites in pyodide https://github.com/pyodide/pyodide/issues/69 running them would be very slow though, and there are also some known failure (cf related open issues about numpy). So we have just written some sanity tests and that's it. However now we have much more CI resources than before and I think we could run them but maybe triggered manually and not for each commit.  Improving the package test situation is also somewhat a blocker for https://github.com/pyodide/pyodide/issues/1577

So there is a lot work that can be done there, for this PR for instance running the numpy test suite and report new issues would probably be sufficient. We can also do this after merging of this PR. The only thing maybe we should add a test for, is one of the operations that uses SIMD to make sure that calling it doesn't error after all the workaround that were done here.

leopsidom:
> I think adding these descriptions into the patch file would help other people who later see patch files.

Good idea. I have added the description to the corresponding patch files.

> It would be great if we add some tests for newly available features of numpy 1.21

I think numpy has it's own test suite. Ideally we might want to see how we  can run the tests in pyodide as commented  by @rth.

leopsidom:
> Right now the CI fails with,
>   `emcc: error: Passing any of -msse, -msse2, -msse3, -mssse3, -msse4.1, -msse4.2, -mavx, -mfpu=neon flags also requires passing -msimd128!`

Yeah, that means we have to pass an extra cflag `-msimd128` to emscripten when compiling SIMD features. It looks like an explicit restriction that emscripten tries to call out about what it actually supports.


> for now maybe the easiest would be
> self.cpu_baseline = "min"
> self.cpu_dispatch = "min"

Sounds good. I've switched both to `min`. Though I don't know if this completely disables SIMD. We might have to use `self.disable_optimization = True` to completely disable it.

hoodmane:
Scipy build is failing because it can't link `npymath`.

leopsidom:
Hmm yeah, will check a little later this week.

leopsidom:
The `npymath` library issue can be fixed by updating `NUMPY_LIB` here: https://github.com/pyodide/pyodide/blob/main/packages/Makefile#L6. But it seems there are some more compatibility issue between numpy and scipy. 

> ModuleNotFoundError: No module named 'numpy.testing.decorators'

I think `numpy.testing.decorators`  has been removed in the newer version of numpy.

leopsidom:
I think this will depend on success of this upgrade: https://github.com/pyodide/pyodide/pull/1293

hoodmane:
Well it's still available as 
`from numpy.testing._private.decorators import setastest`
Maybe we can patch numpy to copy `numpy.testing._private.decorators` to `numpy.testing.decorators`.

leopsidom:
Ahh, good point. Let me try that.

pyodide-build/pyodide_build/pywasmcross.py
```diff
@@ -329,6 +329,8 @@ def handle_command(line, args, dryrun=False):
             continue
         if arg == "-lffi":
             continue
+        if arg == "-mpopcnt":
```
rth, line 332: Maybe let's add a comment that this is a SSE 4.2 instruction, and we currently build without SIMD.

Just so it's easier to know what to re-enable once we want to build with SIMD.

packages/numpy/patches/add-back-decorators-module.patch
```diff
@@ -0,0 +1,340 @@
...
++function object need to use
++::
++
++  nose.tools.make_decorator(original_function)(decorator)
++
++in returning the decorator, in order to preserve meta-data such as
++function name, setup and teardown functions and so on - see
++``nose.tools`` for more information.
++
++"""
```
rth, line 24: You could do 
```
from numpy.testing._private.decorators import *
```
in this patch maybe?

rth (commented):
Thanks @leopsidom, great I see the CI passes!

A few minor comments below otherwise looks good. Could you also please merge upstream/main to sync? I merged https://github.com/pyodide/pyodide/pull/1708 recently which adds a few more patches to numpy, but you should be able to remove the `fix-random-double-fill.patch` with this update.

leopsidom:
It seems the `fix-getters.patch` is also not needed: https://github.com/numpy/numpy/blob/main/numpy/core/src/multiarray/getset.c#L29 ?

rth:
> It seems the fix-getters.patch is also not needed

It's possible, I wasn't 100% sure in what release it got included.

rth:
No actually my initial comment was correct. That patch is [not included in 1.21.4](https://github.com/numpy/numpy/blob/c0b003e9c787ccab27f6fe57c154d7b881da5795/numpy/core/src/multiarray/getset.c#L29), so it's still necessary.

leopsidom:
> That patch is not included in 1.21.4, so it's still necessary.

Yeah, I was looking at a newer commit @1.22.0.dev0

leopsidom:
Also I'm seeing two patches not applied in `meta.yaml`: `0008-fix-dict.patch` and `make-int-return-values.patch`, wonder why we included them in the patches ?

pyodide-build/pyodide_build/pywasmcross.py
leopsidom: Added the comment -

hoodmane:
> Also I'm seeing two patches not applied in meta.yaml: 0008-fix-dict.patch and make-int-return-values.patch, wonder why we included them in the patches ?

Maybe they aren't needed any more, but we just forgot to delete them?

leopsidom:
Okay, makes sense. Let me just remove them -

leopsidom:
Oh wait I think they are correct. The first one is Cython patch not a numpy patch. The 2nd does get included in meta.yaml.

packages/numpy/patches/add-back-decorators-module.patch
leopsidom: Great idea. I have updated the patch

rth:
Thanks @leopsidom !

hoodmane:
Thanks @leopsidom! This is very helpful. Though I think we should have a changelog entry for this?

leopsidom:
No problems. Thanks for accepting the PR @rth. @hoodmane sure, I can take a look and add a changelog for it a little later.

leopsidom:
> Thanks @leopsidom! This is very helpful. Though I think we should have a changelog entry for this?

Opened a [PR](https://github.com/pyodide/pyodide/pull/2054) to update the change log.

PR #1935, opened 2021-11-08
ENH Add FetchResponse.unpack_archive

hoodmane:
### Description

Add an `unpack_archive` method to `FetchResponse`. Usage:
```
resp = await pyfetch("some_archive.tar.gz")
await resp.unpack_archive("target/directory")
```

This is aimed at improving installation of files into the Pyodide file system. See discussion here:
https://github.com/pyodide/pyodide/discussions/1876

I talked about adding a `pyodide.installFiles` api. I think for starters the `unpack_archive` method I added here would be helpful. In particular, I am imagining the following workflow for development:

1. Have a code directory on local machine with Python code to run under Pyodide
2. Set up a server has an endpoint that will zip the Pyodide code directory and send it
3. Use this API to unpack the zip archive into a spot on your path

### Checklists

- [x] Add / update tests


hoodmane:
Okay so I verified that my idea of a dev workflow based on this works. Here is a simple http server that zips and sends over a folder if you make a request to `folder/zipped.zip`.
```
import http.server
import socketserver
import shutil
import os

PORT = 8084


class ZippingHTTPRequestHandler(http.server.SimpleHTTPRequestHandler):
    def do_GET(self):
        """Serve a GET request."""
        if self.path.endswith("/zipped.zip") and not os.path.exists(self.path):
            self.send_directory_zip()
            return
        super().do_GET()
    
    def send_directory_zip(self):
        orig_path = self.translate_path(self.path)
        path = self.translate_path(self.path.removesuffix("/zipped.zip"))
        shutil.make_archive(orig_path.removesuffix('.zip'), 'zip', path)
        try:
            super().do_GET()
        finally:
            os.remove(orig_path)


Handler = ZippingHTTPRequestHandler

with socketserver.TCPServer(("", PORT), Handler) as httpd:
    print("serving at port", PORT)
    httpd.serve_forever()
```
Then this works, modifications to `my_package` will be reflected on refreshing the page.
```
  from pyodide.http import pyfetch
  resp = await pyfetch("my_package/zipped.zip")
  await resp.unpack_archive("/my_packages/my_package")
  import sys
  sys.path.append("/my_packages")
  import my_package	
`);
```

src/py/pyodide/http.py
```diff
@@ -148,8 +150,34 @@ async def bytes(self) -> bytes:
...
+            The archive format: one of “zip”, “tar”, “gztar”, “bztar”.
+            Or any other format registered with ``shutil.register_unpack_format()``. If not
+            provided, ``unpack_archive()`` will use the archive file name extension
+            and see if an unpacker was registered for that extension. In case
+            none is found, a ``ValueError`` is raised.
+        """
+        filename = self._url.rsplit("/", -1)[-1]
+        f = NamedTemporaryFile(suffix=filename)
+        f.write(await self.bytes())
+        shutil.unpack_archive(f.name, extract_dir, format)
```
rth, line 176: It's a shame there is no way to do this without tempfiles, but it looks indeed to be the case.

Related https://bugs.python.org/issue39846, if that was possible with that feature we could have used it in micropip.

rth (commented):
I quite like this, thanks @hoodmane !

> Set up a server has an endpoint that will zip the Pyodide code directory and send it

BTW, Github can do that, 
image
Though I'm not sure they are happy to be used as a CDN.

Looking at your example it's very tempting to add a `add_to_path=False` parameter to unpack_archive (usually the zip has the folder with the module, so it only need to be added to the unpacked parent dir). Though this would blur somewhat the goal of this method.

src/py/pyodide/http.py
```diff
@@ -148,8 +150,34 @@ async def bytes(self) -> bytes:
         self._raise_if_failed()
         return (await self.buffer()).to_bytes()
 
+    async def unpack_archive(self, extract_dir=None, format=None):
+        """Treat the data as an archive and unpack it into target directory.
+
+        Assumes that the file is an archive in a format that shutil has an
+        unpacker for. The arguments extract_dir and format are passed directly
+        on to ``shutil.unpack_archive``.
```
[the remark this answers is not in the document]
hoodmane: In a future PR?

ryanking13:
> Looking at your example it's very tempting to add a `add_to_path=False` parameter to unpack_archive (usually the zip has the folder with the module, so it only need to be added to the unpacked parent dir).

I think modifying the path needs to be handled by the wrapper of this method (e.g. `pyodide.installFiles`). If `unpack_archive` method modifies path, it's not a good design, and not everyone will use this method to download python modules.

rth (commented):
It still needs tests, otherwise LGTM.

grimmer0125:
> Though I'm not sure they are happy to be used as a CDN.

GitHub has CORS issue and will block http requests whose sources are not GitHub.

src/py/pyodide/http.py
```diff
@@ -148,8 +151,64 @@ async def bytes(self) -> bytes:
...
+    async def create_file(self, path: str):
+        """Uses the data to back a new file without copying it.
+
+        This method avoids copying the data when creating a new file. If you
+        want to write the data into an existing file, use
+
+        .. code-block:: python
+
+            buf = await resp.buffer()
+            buf.to_file(file)
```
rth, line 172: The method name is not consistent in this example with the one above I think?

Also I think `to_file` rather than `create_file` would be better, to be constent with the buffer method name

hoodmane: Well this creates a new file, whereas the buffer method appends to an existing file. I think they need to be named differently since they do different things.

rth: Usually, that kind of difference would be passed via the `mode="w"` or `mode='a'` parameter and I find it somewhat confusing to have different names for those. Also, a method that fails when a file exists (instead of overwriting it) is pretty much the opposite of what Python usually does by default?

For instance https://docs.rs/serde_any/0.1.2/serde_any/fn.to_file.html there is no other method for append/create as far as I can tell.
I understand there are implementation constraints but from a usability perspective, I find these different names fairly confusing in the public API.

Could we benchmark how much time we save as opposed to `buf.to_file` with an extra copy? As if the HTTP request is 100ms and the extra copy is 100ns maybe making API more complex for it it's not worth it.

hoodmane: > Also, a method that fails when a file exists (instead of overwriting it) is pretty much the opposite of what Python usually does by default?

I am happy to make it overwrite by default. I guess I am not that familiar with the typical Python behaviors.

>  maybe making API more complex for it it's not worth it.

Probably yeah. Seems to take about 1ms to copy a 300kb buffer, about 3ms to copy a 1mb buffer.

rth:
> Could we benchmark how much time we save as opposed to buf.to_file with an extra copy?

For instance, maybe I'm doing something wrong, but I'm not seeing much difference between between the last two
```py
>>> import timeit
>>> import numpy as np
>>> def timeit_std(code, setup, number=1000, repeat=5):
...     res = timeit.repeat(code, setup=setup, number=number, repeat=repeat)
...     return f'{np.mean(res):.4f}+-{np.std(res):.4f}'
... 
>>> setup = 'from js import Uint8Array; x = Uint8Array.new(range(1000)); from tempfile import TemporaryFile'
>>> kwargs = dict(setup=setup, number=10000, repeat=5
>>> timeit_std('with TemporaryFile() as f: pass', **kwargs)
'1.6424+-0.0389'
>>> timeit_std('with TemporaryFile() as f: f.write(x.to_bytes())', **kwargs)
'1.7394+-0.0197'
>>> timeit_std('with TemporaryFile() as f: x.to_file(f)', **kwargs)
'1.6922+-0.0208'
>>> timeit_std("with TemporaryFile() as f: x._into_file(f)", **kwargs)
'1.6904+-0.0157'
```
though most of the time is spent on `TemporaryFile` presumably to create temp folders.

hoodmane:
And `f.write(x.to_bytes())` is about twice the error margin away. I think you are right that copying the data is probably not the slow part. It is just the only part under my direct control. So yeah, adding complexity to the API is probably not worth it.

Of course, there are some people using Pyodide for real time data processing with very large amounts of data. I think if the data gets large enough it would be more worthwhile to avoid copying. But computers are very good at copying.

hoodmane:
Maybe it would be better to pull the file creation out of the benchmark loop though.

rth:
Moving the test folder setup outside,
```py
>> setup = 'from js import Uint8Array; x = Uint8Array.new(range(100000)); from tempfile import TemporaryFile; f = TemporaryFile()'
>>> kwargs = dict(setup=setup, number=1000, repeat=5)
>>> timeit_std('f.write(x.to_bytes())', **kwargs)
'0.9002+-0.0534'
>>> timeit_std('x.to_file(f)', **kwargs)
'0.8692+-0.0428'
>>> timeit_std("x._into_file(f)", **kwargs)
'0.0016+-0.0008'
```
which does show significant improvement for _into_file. 

If those numbers are correct it means that the first two approaches have a bandwidth around 100MB/s, which is not that bad, but also not great.

Anyway, let's maybe merge this with a public `unpack_archive` method, and a private `_create_file`, and and take some more time to think about a consistent API there? It's not a blocker for this PR in any case. If those numbers are valid, it would mean that users might also want to have a public _into_file method, in which case we might need to re-think the `to_file` / `_into_file` names.

Also we should probably do more I/O and system calls benchmarks.

Issue #1917, opened 2021-10-31
[Proposal] `pyodide-importer` package for external Python file import support


ryanking13:
- This proposal is initiated from the discussion: #1876
- Other related issues: #1808, #1784

## Problem statement

Currently, Pyodide doesn't support importing external Python packages (.py) except for installing a pure Python package wheel (.whl) or Pyodide Python packages(.js, .data) through `micropip.install` or `pyodide.loadPackage`.

This is because the external Python files need to be accessed through HTTP(s) (browser) or File System API (node) while the import statements try to find them inside the virtual file system.

```js
pyodide.runPython(await (await fetch("https://some_url/...")).text());
```

So for now, if someone wants to use external python files, [they need to fetch and execute them manually](https://pyodide.org/en/stable/usage/faq.html#how-can-i-load-external-python-files-in-pyodide), which is not very straightforward.

## Proposal

This issue proposes the `pyodide-importer` which is a Python package that supports seamless import statement support for Pyodide applications.

### Idea

When `import` statement is called, Python tries to find the import candidates through `PathFinder` registered in `sys.meta_path`.

`pyodide-importer` defines a new Python PathFinder (e.g. PyodidePathFinder), which is appended to the `sys.meta_path`, that searches and downloads import candidates through HTTP(s) and File System API.

0. import statement is called and built-in PathFinders failed to locate target packages.
1. PyodidePathFinder searches packages through HTTP(s) or File System API, searching order follows [PEP-420](https://www.python.org/dev/peps/pep-0420/).
2. If PyodidePathFinder successfully finds the target package (which is checked via HTTP status code or fs.stat), the target package is saved to the predefined path inside the virtual file system.
3. Rest of the import process is handled by built-in PathFinder.

### API design (draft)

```python3
"""
pyodide_importer.register_hook()
- appends PyodidePathFinder to sys.meta_path
- also adds virtual file system path to sys.path
- `base_url`: URLs (broswer) or File System Paths (Node) where Python packages are located. The default value is pyodide.indexURL.
- `download_path`: path inside virtual  file system where downloaded Python packages will be located.
- `modules`: names of the modules/packages that can be imported from `base_url`.
"""
def register_hook(
    base_url=pyodide.indexURL,
    download_path="",
    modules=None,
)

"""
pyodide_importer.unregister_hook()
- removes PyodidePathFinder from sys.meta_path
"""
def unregister_hook()
```

## Limitations
[...]

PR #1936, opened 2021-11-08
Add custom home directory support

ryanking13:
### Summary

Add `homeDirectory` parameter to `pyodide.loadPyodide()`, which sets the home directory of the Pyodide file system.

```js
    indexURL : "https://cdn.jsdelivr.net/pyodide/dev/full/",
    homeDirectory: "/home/pyodide",
});
```

### Description

Since we are planning to utilize the Pyodide virtual file system for external file installation,
I think having a proper home directory would improve the user experience. (https://github.com/pyodide/pyodide/issues/1917#issuecomment-962490416)

For now, root directory `/` is used for a home directory (cwd). But there exist some default subdirectories like `/tmp`, `/dev`, `/proc`, which users don't need to see and might have conflict with other newly created files.

image

This PR does:

- Add `homeDirectory` parameter to `pyodide.loadPyodide()`.
  - make `homeDirectory` inside the Pyodide virtual file system, then chdir to it.
  - default value is set to `/home/pyodide`

### Checklists

- [x] Add / update tests


src/js/module.js
```diff
@@ -84,3 +83,23 @@ function createStdinWrapper(stdin) {
...
   return stdinWrapper;
 }
+
+/**
+ *
+ * @param {string} path
+ */
+export function setHomeDirectory(path) {
+  Module.preRun.push(function () {
+    fallbackPath = "/";
```
hoodmane, line 93:
```suggestion
    const fallbackPath = "/";
```

hoodmane:
Thanks @ryanking13!

rth:
Thanks for this proposal @ryanking13 !

Overall sounds good, however,
 - we should more explicitly mention it in the changelog that this is a backward incompatible change, and to get previous behavior users need to "..".
 - If possible IMO we shouldn't expose `setHomeDirectory` to the public API (as it currently it), unless there is a clear value to users. Otherwise we would have to keep backward compatibility on it as much as possible, which adds more constraints for future development.

ryanking13:
@hoodmane Thanks. Applied reviews!

@rth

> * we should more explicitly mention it in the changelog that this is a backward incompatible change, and to get previous behavior users need to "..".

True. I'll update the changelog to be more explicit.

> * If possible IMO we shouldn't expose `setHomeDirectory` to the public API (as it currently it), unless there is a clear value to users. Otherwise we would have to keep backward compatibility on it as much as possible, which adds more constraints for future development.

Does this mean just making `setHomeDirectory` API private and keep `homeDirectory` param as is?
or does it mean we better not give `homeDirectory` option of changing home directory to users?

hoodmane:
Another thought is that we might want to add the home directory to `sys.path`. Not sure.

ryanking13:
> Another thought is that we might want to add the home directory to `sys.path`. Not sure.

I think it's a good idea as [Python actually does so](https://docs.python.org/3/library/sys.html#sys.path).

> As initialized upon program startup, the first item of this list, path[0], is the directory containing the script that was used to invoke the Python interpreter.

rth:
> Do you mean just making setHomeDirectory API private

I means keeping the function private, and the homeDirectory parameter public, if possible.

hoodmane:
> If possible IMO we shouldn't expose setHomeDirectory to the public API (as it currently it)

I don't think `setHomeDirectory` is actually "exposed". It's exported from `module.js` but I don't think we intend for exports from `module.js` to be used by other people. I think it's currently fine the way it is. We need to organize things a bit better to make it clear what is for use by other people and what it not -- it's currently not that clear. I don't think we follow the usual standards for js packages. I think usually you have a `index.js` file and only stuff exported from that counts as being exposed. 

But actually I'm really not sure. I'm having trouble finding info about how exports are supposed to be structured for Javascript packages.

hoodmane:
https://nodejs.org/api/packages.html#packages_subpath_exports

rth:
> I don't think we intend for exports from module.js to be used by other people. 

Yeah, it's also not very clear to me how to manage exports visibility. But right now it's in [the generated docs](https://pyodide--1936.org.readthedocs.build/en/1936/usage/api/js-api.html#pyodide.setHomeDirectory) which I interpret as being part of the public API.

hoodmane:
Definitely should keep it out of the docs. To do that, add `@private` to doc strings.

ryanking13:
> But right now it's in [the generated docs](https://pyodide--1936.org.readthedocs.build/en/1936/usage/api/js-api.html#pyodide.setHomeDirectory) which I interpret as being part of the public API.

Yes. I'll make the function private.
I think [setStandardStreams](https://pyodide.org/en/latest/usage/api/js-api.html#pyodide.setStandardStreams) also needs to be private?

src/js/module.js
```diff
@@ -84,3 +83,23 @@ function createStdinWrapper(stdin) {
   }
   return stdinWrapper;
 }
+
+/**
+ *
```
hoodmane, line 88:
```suggestion
 * @private
```

hoodmane:
> I think `setStandardStreams` also needs to be changed to private?

Yeah.

hoodmane:
I think in a future PR we should add something like:
```json
"exports" : { "." : "pyodide.js" }
```
to `package.json`. Also, we should move the actual source files into a `src` directory, rather than having source code mixed with package metadata.

ryanking13:
- Updated changelog of breaking changes (Added a new label for breaking change)
- Made `setHomeDirectory` and `setHomeDirectory` private

> Another thought is that we might want to add the home directory to sys.path. Not sure.

hoodmane:
> which will be the best place to update sys.path at initialization.

@ryanking13 Maybe wait until #1939 is merged, then add it here:

https://github.com/pyodide/pyodide/blob/a6ed599d921a5097604208660445d9d4e351babd/src/js/pyodide.js#L172
```
  Module.sys = Module.runPythonInternal("import sys; sys");
  Module.sys.setrecursionlimit(calculateRecursionLimit());
  Module.sys.path.append(config.homeDirectory);
```

hoodmane (approved):

hoodmane:
Actually now I think we should be setting `config.home` here:
https://github.com/pyodide/pyodide/blob/fdd1aa1c75ae1ca2c762ec04f93e4cfdfe0fb212/src/core/main.c#L50

hoodmane:
Maybe set `Module.arguments = [config.homeDirectory];` and then call `initialize_python(argsv[1]);` and then `PyConfig_SetBytesString(&config, &config.home, home_directory);`

ryanking13:
> Maybe set `Module.arguments = [config.homeDirectory];` and then call `initialize_python(argsv[1]);` and then `PyConfig_SetBytesString(&config, &config.home, home_directory);`

Yeah, the Python initialization step will be a more appropriate place for sys.path setup.

But It looks like setting config.home affects not only the working directory [but also affects the location of the standard Python libraries](https://docs.python.org/3/c-api/init.html#c.Py_SetPythonHome).
I will check which config variable is the right one to modify.

hoodmane:
> affects not only the working directory but also affects the location of the standard Python libraries.

Oh that doesn't sound like what we want. Also, to get the value into C, perhaps it would be easier to say `Module.homeDirectory = Module.stringToNewUTF8(config.homeDirectory);` and then use `EM_ASM_INT`:
```C
char* home_directory = EM_ASM_INT(Module.homeDirectory);
```

hoodmane:
Ah but see the note here:
https://docs.python.org/3/c-api/init.html#c.PySys_SetArgvEx

> It is recommended that applications embedding the Python interpreter for purposes other than executing a single script pass 0 as updatepath, and update sys.path themselves if desired.

So what you've already done is the right recommended way.

ryanking13:
> Ah but see the note here: https://docs.python.org/3/c-api/init.html#c.PySys_SetArgvEx
> 
> > It is recommended that applications embedding the Python interpreter for purposes other than executing a single script pass 0 as updatepath, and update sys.path themselves if desired.
> 
> So what you've already done is the right recommended way.

Indeed. It looks like in native Python, sys.path[0] is updated at [pymain_run_python](https://github.com/python/cpython/blob/9178f533ff5ea7462a2ca22cfa67afd78dad433b/Modules/main.c#L535), but we are not invoking that function so we have to update sys.path[0] ourselves.

ryanking13:
I think I can try updating sys.path[0] inside initialize_python() C code, by invoking `pymain_sys_path_add_path0()` directly. But I'm not sure it's necessary.

hoodmane:
I think we should keep it as is.

docs/project/changelog.md
```diff
@@ -13,6 +14,13 @@ substitutions:
 
 ## Unreleased
 
+### Backward incompatible changes
+
+- {{Breaking}} Default working directory(home directory) inside Pyodide virtual
+  file system has been changed from `/` to `/home/pyodide`. You can `os.chdir("/")`
+  to change working directory if needed.
```
rth, line 21:
```suggestion
  file system has been changed from `/` to `/home/pyodide`. To get previous behavior, you can
  - call `os.chdir("/")`  to change working directory
  - or call {any}`loadPyodide <globalThis.loadPyodide>` with the `homeDirectory="./"` argument
```

docs/project/changelog.md
```diff
@@ -33,7 +41,11 @@ substitutions:
 
 - {{Fix}} Webpack building compatibility issues and a {any}`loadPyodide <globalThis.loadPyodide>`
   runtime issue due to webpack are solved. 
-  {pr}`1900`  
+  {pr}`1900`
+
+- {{API}} {any}`loadPyodide <globalThis.loadPyodide>` now accepts `homeDirectory`
```
rth, line 46: I wonder if should name this argument `home` (name of UNIX [$HOME](https://en.wikipedia.org/wiki/Home_directory) also in [pathlib.Path.home()](https://docs.python.org/3/library/pathlib.html#pathlib.Path.home)) or maybe better `homedir` as in [Node.js](https://nodejs.org/api/os.html#os_os_homedir). `homeDirectory` is a bit long.

src/js/module.js
```diff
@@ -84,3 +84,24 @@ function createStdinWrapper(stdin) {
...
+ * @param {string} path
```
rth, line 90: Please add a brief function description.

rth (changes requested):
Marking as "changes requested" so we don't forget to add tests. Otherwise LGTM, aside from minor comments below.

ryanking13:
Applied comments!

- homeDirectory --> homedir
- description for setHomeDirectory()

rth:
Thanks @ryanking13 !  

I think we should come up with some standard way of indicating that a PR is ready to be merged. Here I'm still not merging it because all elements in the checklist in the PR description are not checked. So I think it would be good to either resolve each element or mark it as not applicable (but still check it). Here for instance we could 
 - add a test `src/tests/test_pyodide.py` to check that `os.getcwd()` returns the expected (default) value.
 - either check " Add new / update outdated documentation" or remove it altogether, as I think no other changes are needed to the docs (unless I missed something)?

Althernatively we could use the draft PR status -- it's just helpful to be able to tell quickly if a PR can be merged after an approval without re-reading all the  discussion (aside from the description and maybe the last few comments to see if something was forgotten) 

If you were planning to do this later, no problem at all. Just checking so we are on the same page expectation wise, so you don't wait on me to merge this PR while I wait for the checklist be finished :)

ryanking13:
> I think we should come up with some standard way of indicating that a PR is ready to be merged. Here I'm still not merging it because all elements in the checklist in the PR description are not checked.

Yeah, I think we need to add more description on PR template that checklists should be all checked or removed.

I'll update tests and ping you again, I don't think additional documentation is needed.

> Althernatively we could use the draft PR status -- it's just helpful to be able to tell quickly if a PR can be merged after an approval without re-reading all the discussion (aside from the description and maybe the last few comments to see if something was forgotten)

I don't think draft PR status is good for this purpose. Draft status usually means "It's WIP and not ready for review".
Maybe we can use "Ready for review" or "Ready to Merge" label instead.

ryanking13:
@rth Added tests and I think it's ready to be merged

rth (approved):
Thanks @ryanking13 !

PR #1937, opened 2021-11-09
MAINT Bring lib files together under src/py/lib directory

hoodmane:
This is supposed to simplify the build a bit and make it easier to understand what is happening with the files `pystone.py`, `_testcapi.py`, `_testinternalcapi.py` and `webbrowser.py` (namely they are all getting added to the Python standard library, not being used at build time or placed somewhere else).


hoodmane:
Also, I don't think `src/_headers` is used anywhere? I deleted it.

ryanking13:
I'm just curious, do we actually need those files built with Pyodide?

rth:
> Also, I don't think src/_headers is used anywhere? I deleted it.

Sounds good.

> and make benchmark.py directly execute the script instead of importing it?

we could indeed in a follow up PR.

> will CPython test break if these files do not exist?

I imagine it would just make the CPython test suite fail.

hoodmane:
> _testcapi, _testinternalcapi: will CPython test break if these files do not exist? We already don't have other test extensions I guess

Yes CPython tests absolutely break if these don't exist. We could just build them normally, I'm not sure what the size impact would be but it would increase test coverage. Though some of these internal apis have problems, like for instance calling `PyCode_SetExtra` causes memory corruption and eventually segfaults. So it would take some effort to remove tests that call broken APIs. (`PyCode_SetExtra` is only used for JIT which we have also disabled by throwing out the `PyFrame_Eval` hook so it's no big deal.)

We get some extra coverage out of the `ctypes` test suite because it redoes some of these api tests but in the context of testing ctypes (since the apis themselves are normally tested elsewhere).

Makefile
```diff
@@ -39,10 +39,7 @@ build/pyodide.asm.js: \
 	src/core/pyproxy.o \
 	src/core/python2js_buffer.o \
 	src/core/python2js.o \
-	src/pystone.py \
-	src/_testcapi.py \
-	src/_testinternalcapi.py \
-	src/webbrowser.py \
+	$(wildcard src/lib/**/*) \
```
rth, line 42: Maybe,
```suggestion
	$(wildcard src/lib/py/*.py) \
```
to be a bit more specific?

rth (approved):

Makefile
hoodmane: Oops what I had was just wrong.

PR #1938, opened 2021-11-09
MAINT Rearrange emsdk Makefile to do less stuff before applying patches

hoodmane:
It takes a while to build `ccache`. Avoid doing this before applying patches so that it takes less time to get to patch fail. I also changed it so that we clone only the branch we need of binaryen with depth 1.

### DIFF
--- a/emsdk/Makefile
+++ b/emsdk/Makefile
@@ -6,10 +6,10 @@ all: emsdk/.complete
 emsdk/.complete: ../Makefile.envs $(wildcard patches/*.patch)
 	if [ -d emsdk ]; then rm -rf emsdk; fi
 	git clone --depth 1 https://github.com/emscripten-core/emsdk.git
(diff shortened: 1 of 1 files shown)


rth:
Thanks LGTM.

> It takes a while to build ccache

BTW we are using another ccache to build ccache (and now binaryen) :) It shouldn't be that that long if the first one is working.

hoodmane:
It doesn't take *that* long, but it does take significantly longer than everything else that happens before patching. It's also much slower for me because of WSL, I should switch to real linux.

hoodmane:
Interesting, black had a sporadic failure here.

hoodmane:
Worrisome looking oob memory error in `test_cffi_asprintf` in test-packages chrome. I think it's an actual regression?

hoodmane:
Okay saw the same oob memory error on #1935 which modifies completely unrelated stuff. I think that means this PR is okay.

PR #1939, opened 2021-11-09
MAINT simpler bootstrap

hoodmane:
I got rid of `runPythonSimple` and instead of wrapping a `dict` as the first `PyProxy`, I wrap `_pyodide`. We can't import `pyodide` until after we hook up the Python and Javascript globals. But with `_pyodide` we can call `register_js_finder`, `register_js_module`, and `eval_code`. Calling `eval_code` directly allows us to return results from Python right away which makes the rest of the boostrap setup simpler.

### DIFF
--- a/src/core/main.c
+++ b/src/core/main.c
@@ -76,20 +76,6 @@ static struct PyModuleDef core_module_def = {
   .m_size = -1,
 };
 
-PyObject* init_dict;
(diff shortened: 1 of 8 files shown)


rth (approved):
This looks simpler indeed. Thanks!

PR #1942, opened 2021-11-10
ENH Add tofile and fromfile to buffer objects

hoodmane:
Javascript buffers require extra C support to write to / read from files without first copying the data. #1935 copies the archive an extra time, but with these APIs it won't need to.

- [x] Add / update tests


rth:
As names `from_file` and `to_file` I think would be clear enough (and analogous to [what is done for memoryviews](https://docs.python.org/3/c-api/memoryview.html)) . I find  reading from file into an existing PyProxy objects somewhat unclear, I would have expected `read_from_file` to be a class method. What bothers me a bit here is we would be inventing a custom serialization format, even if it's just the bytes stream. 

It would also be good to look more in details what API  for buffer serialization with minimal copy is used in Python. For instance there is [PEP 574](https://www.python.org/dev/peps/pep-0574/). Generally Python objects do not have a to_file or from_file methods but rather use pickle for that (related #1946 )

hoodmane:
> As names from_file and to_file I think would be clear enough

I'm not sure. `to_file` sounds like it is making a new file, rather than writing into an existing one?

Ah here is a method named exactly like that:
https://docs.python.org/3/library/array.html?highlight=tofile#array.array.tofile

hoodmane:
Okay let's name them `tofile` and `fromfile` since this matches with `array` from stdlib:
[docs.python.org/3/library/array.html?highlight=tofile#array.array.tofile](tofile)
[https://docs.python.org/3/library/array.html#array.array.fromfile](fromfile)

hoodmane:
Also, we can mount an existing buffer as a file without any copying using the `canOwn` argument to the emscripten `write` API. This would reduce us to the actual minimum number of copies when unpacking a zip archive.

hoodmane:
I guess I'm not sure about whether or not we should have the underscore: `to_file` or `tofile`, `to_bytes` or `tobytes`, etc. Array has them without underscores:
https://docs.python.org/3/library/array.html
but I assume other interfaces have them with underscores.

rth:
I would be +1 for the underscored version as more readable, also we already have `JsProxy.to_py`.

rth:
My impression is that  some of those method names in CPython are there for historical reasons and cannot be changed now, but I feel like newer API would have used the underscored version.

For instance,
 - historical API: `os.path.isfile`
 - new API: `pathlib.Path.is_file()`

and other examples in [pathlib](https://docs.python.org/3/library/pathlib.html#correspondence-to-tools-in-the-os-module)

But maybe I'm using pandas too much which has lots of `to_*` and `from_*` methods.

I'll respond to your earlier comments later.

hoodmane:
>  some of those method names in CPython are there for historical reasons and cannot be changed now

Seems right to me. In any case, since stdlib has examples of both I think we are in good shape as long as we stick to one convention. I think including the underscore makes more sense overall.

ryanking13:
+1 to including the underscore

src/py/_pyodide/_core_docs.py
```diff
@@ -124,6 +124,36 @@ def to_bytes(self) -> bytes:
...
         """
 
+    def to_file(self, fd: int):
+        """Writes the entire buffer to a file descriptor.
+
+        Will write the entire contents of the buffer to the current position of
+        the file descriptor.
+
+        Present only if the wrapped Javascript object is an ArrayBuffer or an
+        ArrayBuffer view.
```
rth, line 134: I think it would be help to have an example (adapted from your unit test) and an equivalent code that makes more copies. For instance,
```suggestion
ArrayBuffer view.

Example
------------
>>> from js import Uint8Array
>>> x = Uint8Array.new(range(10))
>>> with open('file.bin', 'w') as fh:
...    x.to_file(fh.fileno())

which is equivalent to,

>>> with open('file.bin', 'w') as fh:
...    fh.write(x.to_py().tobytes())

but the latter is less efficient and makes more memory copies.
```


(might need to add `# doctest: +SKIP`, or skip it in `conftest.py`) as there is probably no way to make it work from Python for doctests without building this for WASM.

src/py/_pyodide/_core_docs.py
```diff
@@ -124,6 +124,36 @@ def to_bytes(self) -> bytes:
...
+        """
+
+    def from_file(self, fd: int):
+        """Reads from a file descriptor into the buffer.
+
+        Will try to read a chunk of data of size the length of the buffer from
+        the current position of the file descriptor.
+
+        Present only if the wrapped Javascript object is an ArrayBuffer or an
+        ArrayBuffer view.
```
rth, line 144: Also an example would be good to really understand how to use this, something like,
```suggestion
        ArrayBuffer view.


Example
------------
>>> from js import Uint8Array
>>> # the JsProxy need to be pre-allocated
>>> x = Uint8Array.new(range(10))
>>> with open('file.bin', 'w') as fh:
...    x.read_file(fh.fileno())
which is equivalent to,
>>> with open('file.bin', 'w') as fh:
...    chunk = fh.read(size=x.byteLength)
...    # not sure how to assign this result to x buffer
but the latter is less efficient and makes more memory copies.
```

src/py/_pyodide/_core_docs.py
```diff
@@ -124,6 +124,36 @@ def to_bytes(self) -> bytes:
...
+        """
+
+    def _into_file(self, fd: int):
+        """Will write the entire contents of the buffer to the current position
+        of the file descriptor using ``canOwn : true``. If the file is in the
```
rth, line 149: I had to search emscripten source code to understand what this does. More explanations of what it means won't hurt.

Also couldn't this be a `tofile(fh, can_own=True)`?

rth (commented):
I understand this can be useful, but I think it need more examples to illustrate how to use these.

For `from_file` I understand the use case, but I still find it still unexpected that one needs to pre-allocate the JsProxy buffer with the correct size. If feels very low level. Are you aware of any Python modules that do similar things? For instance as far as I know array.fromfile doesn't need to know the size in advance.  https://docs.python.org/3/library/struct.html#struct.Struct.unpack does it, I guess. I'm not opposed to adding it, assuming there are more docstring examples.


> The point of this method is to avoid copying the data from the original buffer into the wasm heap just to write it into the file system.

It would be good to benchmark how much difference that copy would make?

src/py/_pyodide/_core_docs.py
```diff
@@ -124,6 +124,36 @@ def to_bytes(self) -> bytes:
         an ArrayBuffer view.
         """
 
+    def to_file(self, fd: int):
```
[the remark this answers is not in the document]
hoodmane: > Couldn't we take the file object as input instead?

Yes we could if you prefer. I was concerned that some IO objects don't have file descriptors, but we can just let the `OSError` be raised then. In my debugging, I more often just have a file descriptor so taking a file descriptor is useful to me. But I can just use `os.fdopen` to wrap my file descriptor in a file.

src/py/_pyodide/_core_docs.py
hoodmane: > couldn't this be a `tofile(fh, can_own=True)`?

It could be, but I would prefer to keep this one as a private API for now because it's a bit "unsafe". In particular, after mounting this `ArrayBuffer` into the file system, the file system "owns" the buffer so file operations on the file could be cached in the wasm memory. Changing the buffer directly can cause data corruption.

hoodmane: This `_into_file` command is intended primarily as an optimization for `unpack_archive`. It eliminates another copy compared to `to_file`

#### original code (3 copies)
```py
resp = await pyfetch("archive.zip")
bytes = await resp.bytes() # copy data into wasm heap
f.write(bytes) # copy data into memfs
shutil.unpack_archive(f.name) # copy data from memfs into wasm heap
```

#### to_file (2 copy)
```py
resp = await pyfetch("archive.zip")
buf = await resp.buffer() # no copy
buf.to_file(f) # copy data into memfs
shutil.unpack_archive(f.name) # copy data from memfs into wasm heap
```
vs
#### into_file (1 copy)
```py
resp = await pyfetch("archive.zip")
buf = await resp.buffer() # no copy
buf._into_file(f) # no copy, give memfs ownership of buf
shutil.unpack_archive(f.name) # copy data from memfs into wasm heap
```

hoodmane: How's it look now?

src/py/_pyodide/_core_docs.py
```diff
@@ -124,6 +125,81 @@ def to_bytes(self) -> bytes:
...
+
+        Will write the entire contents of the buffer to the current position of
+        the file.
+
+        Present only if the wrapped Javascript object is an ArrayBuffer or an
+        ArrayBuffer view.
+
+        Example
+        ------------
+        >>> import pytest; pytest.skip()
```
rth, line 139: We should find a better way of doing this but for now it's fine.

docs/project/changelog.md
```diff
@@ -46,6 +46,10 @@ substitutions:
   {any}`JsProxy.to_memoryview` to allow for conversion of `TypedArray` to
   standard Python types without unneeded copies. {pr}`1864`
 
+- {{Enhancement}} Added {any}`JsProxy.to_file` and {any}`JsProxy.from_file` to
+  allow reading and writing Javascript buffers to files without unneeded copies.
```
rth, line 50: Maybe,

```suggestion
  allow reading and writing Javascript buffers to files as a byte stream without unneeded copies.
```
as otherwise the question is what format we use for serialization.

src/py/_pyodide/_core_docs.py
```diff
@@ -124,6 +125,81 @@ def to_bytes(self) -> bytes:
...
+        ...    data = x.to_bytes()
+        ...    fh.write(data)
+        but the latter copies the data twice whereas the former only copies the
+        data once.
+        """
+
+    def from_file(self, file: IOBase):
+        """Reads from a file into the buffer.
+
+        Will try to read a chunk of data of size the length of the buffer from
```
rth, line 155: It's slightly hard to parse with so many `of`

Maybe,
```suggestion
        Will try to read a chunk of data corresponding to the length of the buffer from
```

rth (approved):
Thanks! Already having the file objects instead of the file descriptors feels much more user friendly..

LGTM.

PR #1648, opened 2021-06-19
Update to emscripten 2.0.24

hoodmane:

PR #1943, opened 2021-11-10
Update emscripten to 20.0.17

hoodmane:
Work in progress. Currently results in some out of bounds access errors in chrome. I wonder if there is any relation with the memory errors I see in #1938.

@rth I wonder if these errors have anything to do with the chrome bug in #1449? We should update the chrome driver.


rth:
> We should update the chrome driver.

Trying to update in https://github.com/pyodide/pyodide/pull/1952

ryanking13:
I found that we were trying to update emscripten to 20.0.24 in #1648, is this PR a stepping stone for #1648?

rth:
Yes, updating in smaller steps is often easier (when the large update fails has some issues). Make it easier to find the change in emsctipen that could responsible for the failure.

Issue #1940, opened 2021-11-09
New install and import APIs


hoodmane:
## 🚀 Feature

We should add an API like `pyodide.import("package_name");` that imports a package and returns it as a `PyProxy`. Possibly including a variant like `pyodide.import("package_name", ["name1", "name2"]);` or `pyodide.import("package_name", "*");`.

We should also add an API like `pyodide.installFiles` that wraps the new API in #1935.

### Motivation

We should work towards encouraging people to define Python functions and then call Python functions from Javascript, rather than using templating. `pyodide.runPython` is like `eval`, in general it's considered an antipattern to eval a templated string. But I think our current status makes it unclear to new users how to avoid it.

We have existing PRs to add Python home directory #1936 and add good APIs to directly install files into the Python home directory #1935. 

### Pitch

We should move towards thinking that `pyodide.runPython` is not the default way to run Python code. Instead, we should encourage that the users to write python source files (with all the tooling benefits that entails, like code highlighters, formatters, type checkers, intellisense). Then they should use an API to install their code into the file system (future work to add this to the js api). Then they should use this import API to import their package, and then directly call the APIs from Javascript.

So they would:
1. Call `pyodide.install("url to archive or single source file");` which would install the file in their home directory.
2. Say `let my_package = pyodide.import("my_package");`
3. Use their package directly e.g., `my_package.some_api(37, "some_arg");`

This instead of `pyodide.runPython` should be the new best practice for using `pyodide`.

ryanking13:
> We should work towards encouraging people to define Python functions and then call Python functions from Javascript, rather than using templating.

Yeah, definitely agree.

After APIs like `pyodide.install` are added, we could also encourage people to define Python packages that do the main job and use `pyodide.runPython` just for an entrypoint.

1. `pyodide.install("main_package")`
2. `pyodide.runPython("import main_package; minimal_bootstrap(); main_package.do_main_stuffs()")`

hoodmane:
I'm concerned that there is currently a slippery slope from 
`pyodide.runPython("main_package.do_main_stuffs()")`
to
```pyodide.runPython(`main_package.do_main_stuffs({"a" : ${some_value}, "b" : ${some_other_value}})`)```
when it would be better would be:
```js
let main_package = pyodide.runPython("import main_package; main_package");
main_package.do_main_stuffs(pyodide.toPy({a : some_value, b : some_other_value}));
```
[...]

PR #1944, opened 2021-11-10
ENH Add pyimport api that imports a Python package and returns it as a pyproxy

hoodmane:
See discussion in #1940. It is not a great idea to name it `import` since that is a Javascript reserve keyword. I am happy to consider other names.

This is on top of #1939 because it is simpler that way.

### Checklists

- [x] Add / update tests


ryanking13:
Maybe there can be a better name than `importPythonModule`, but I can't think of one. So the name LGTM

docs/project/changelog.md
```diff
@@ -32,8 +32,14 @@ substitutions:
   {pr}`1849`
 
 - {{Fix}} Webpack building compatibility issues and a {any}`loadPyodide <globalThis.loadPyodide>`
-  runtime issue due to webpack are solved. 
-  {pr}`1900`  
+  runtime issue due to webpack are solved.
+  {pr}`1900`
+
+- {{Enhancement}} Added a {any}`pyodide.pyimport` api to import a Python module
```
rth, line 38:
```suggestion
- {{Enhancement}} Added a {any}`pyodide.pyimport` API to import a Python module
```

docs/project/changelog.md
```diff
@@ -32,8 +32,14 @@ substitutions:
...
+  and return it as a `PyProxy`. Note that this does a different thing than the
+  original `pyimport` api: it imports a package and returns it without adding
```
rth, line 40:
```suggestion
  original `pyimport` API: it imports a package and returns it without adding
```

docs/project/changelog.md
```diff
@@ -32,8 +32,14 @@ substitutions:
...
+  the package to the global scope.
```
rth, line 41: But  it would still be in the sys.modules, if you access it from Python right? I'm just curious, no need to add it to the changelog.

src/js/api.js
```diff
@@ -288,6 +288,15 @@ export function toPy(obj, { depth = -1 } = {}) {
   return Module.hiwire.pop_value(result);
 }
 
+/**
+ *
```
rth, line 292: Please add a docstring description, and mention here that the behavior is completely different from the one we had before.

Also it could be good to have an example in the docstring of how to use it, and also mention this is equivalent to `pyodide.runPython('import module; module')` but without adding the module to globals.

rth (commented):
Thanks, looks nice! A few comments below.

rth:
Also similarly to https://github.com/pyodide/pyodide/pull/1936#issuecomment-968060924 I didn't mention tests in the review as I assume that you were planning to add them given that it's the checklist. 

This auto-generation of the checklist can probably be confusing with new contributors if they don't pay too much attention to it  :) Or rather it's sometimes OK to skip steps in a checklist, and rely on the review to catch it, but the way it's currently formulated is more like a todo list and then it can be confusing if that was intentional or just the template.

docs/project/changelog.md
hoodmane: Yes

PR #1948, opened 2021-11-11
MAINT Improve hiwire reference counting

hoodmane:
This reduces churn in the hiwire map. Instead of just adding a new unrelated key to `incref`, we store a refcount and increment this. Then for `decref`, instead of deleting the key we decrement the refcount, then check if the new refcount is zero. If it is, then we delete the key from the map.

This is a pretty minor change but I think it's a slight improvement.

### DIFF
--- a/src/core/hiwire.c
+++ b/src/core/hiwire.c
@@ -157,6 +157,8 @@ EM_JS_NUM(int, hiwire_init, (), {
     // clang-format on
   };
 
+  Module.hiwire.incref = function(idval) { _hiwire.objects.get(idval)[1]++; };
+
   Module.hiwire.pop_value = function(idval)
   {
     let result = Module.hiwire.get_value(idval);
@@ -261,7 +263,7 @@ EM_JS(JsRef, hiwire_incref, (JsRef idval), {
   if (idval & 1) {
     // least significant bit unset ==> idval is a singleton.
     // We don't reference count singletons.
-    Module.hiwire.get_value(idval)[1]++;
+    Module.hiwire.incref(idval);
   }
   return idval;
 });
--- a/src/core/hiwire.c
+++ b/src/core/hiwire.c
@@ -49,10 +49,10 @@ EM_JS_NUM(int, hiwire_init, (), {
   Module.hiwire.TRUE = DEREF_U8(_Js_true, 0);
   Module.hiwire.FALSE = DEREF_U8(_Js_false, 0);
 
-  _hiwire.objects.set(Module.hiwire.UNDEFINED, undefined);
-  _hiwire.objects.set(Module.hiwire.JSNULL, null);
-  _hiwire.objects.set(Module.hiwire.TRUE, true);
-  _hiwire.objects.set(Module.hiwire.FALSE, false);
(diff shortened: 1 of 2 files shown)


hoodmane:
One slightly alarming thing here was that `hiwire_decref` had a return value of `errcode` that wasn't ever checked. This is not good, errors need to be fatal or checked. I think we have several functions that have return values of `errcode` that are never checked. I think this is a sign that the error should be fatal.

hoodmane:
Another weird thing is that we don't check for error fatalness in `EM_JS` wrappers so it seems like it's possible that Python calls Javascript calls Python, the inner Python triggers a fatal error, the outer Python catches it, wraps it as a Python exception and rethrows it. Desired behavior would be to just let it pass...

hoodmane:
We should add `warn_unused_result` to these functions.

src/core/hiwire.c
```diff
@@ -254,20 +257,16 @@ JsString_FromId(Js_Identifier* id)
...
+EM_JS(JsRef, hiwire_incref, (JsRef idval), {
+  if (idval & 1) {
     // least significant bit unset ==> idval is a singleton.
     // We don't reference count singletons.
-    // clang-format on
-    return;
+    Module.hiwire.get_value(idval)[1]++;
   }
-  return Module.hiwire.new_value(Module.hiwire.get_value(idval));
+  return idval;
```
rth, line 266: This is clearly nicer than before!

rth (commented):
Thanks @hoodmane ! This is indeed a more elegant solution for reference counting.

Could we come up with some micro-benchmark to time the creation of lots JsProxy (and `_hiwire.objects`) and run it before and after this change? For now I find not obvious to tell if storing and creating an extra array for each object could impact the performance or not at all.

Also CI is still failing.

hoodmane:
I ran some benchmarks like

<details>
<summary>Make 40000 objects then decref them</summary>

```js
performance.mark("startTest");
let Module = __pyodide_module;
let x = Module.hiwire.new_value({});
let y;
for(let i = 0; i < 40000; i++){
    y = Module.hiwire.new_value({});
}
for(let i = x; i <= y; i+=2){
    Module.hiwire.decref(i);
}
performance.mark("endTest");
performance.measure("test", "startTest", "endTest");
```

</details>

and

<details>
<summary>Make 40,000 objects, incref them all, then decref them twice</summary>

```js
let hiwire_incref;
if(Module.hiwire.incref){
  hiwire_incref = function(id){
  		return Module.hiwire.incref(id);
  }
} else {
  hiwire_incref = function(id){
  		return Module.hiwire.new_value(Module.hiwire.get_value(id));
  }
}
performance.mark("startTest");
let Module = __pyodide_module;
let x = Module.hiwire.new_value({});
let y;
let ids = [];
for(let i = 0; i < 40000; i++){
    y = Module.hiwire.new_value({});
}
for(let i = x; i <= y; i+=2){
    ids.push(hiwire_incref(i));
}
for(let id of ids){
    Module.hiwire.decref(id);
}
for(let i = x; i <= y; i+=2){
    Module.hiwire.decref(i);
}
performance.mark("endTest");
performance.measure("test", "startTest", "endTest");
```

</details>

rth:
Thanks!

PR #1950, opened 2021-11-12
MAINT Be more careful about checking error code return values

hoodmane:
If we use a function defined with `EM_JS_NUM(errcode, ...)` we need to actually check the error code. If we don't ever check the error code, then we can make all errors fatal by defining it with `EM_JS(void, ...)` instead. I added `WARN_UNUSED` to some functions to help detect these mistakes in the future.

### DIFF
--- a/src/core/hiwire.h
+++ b/src/core/hiwire.h
@@ -4,6 +4,7 @@
 #include "Python.h"
 #include "stdalign.h"
 #include "types.h"
+#define WARN_UNUSED __attribute__((warn_unused_result))
 
 /**
  * hiwire: A super-simple framework for converting values between C and
@@ -97,7 +98,7 @@ hiwire_incref(JsRef idval);
 /**
  * Decrease the reference count on an object.
  */
-errcode
+void
 hiwire_decref(JsRef idval);
 
 /**
@@ -191,8 +192,14 @@ JsArray_New();
 /**
  * Push a value to the end of a JavaScript array.
  */
-errcode
+errcode WARN_UNUSED
 JsArray_Push(JsRef idobj, JsRef idval);
+
+/**
+ * Same as JsArray_Push but panics on failure
+ */
+void
+JsArray_Push_unchecked(JsRef idobj, JsRef idval);
 
 /**
  * Create a new JavaScript object.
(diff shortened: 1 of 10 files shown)


src/core/hiwire.h
```diff
@@ -191,9 +192,15 @@ JsArray_New();
 /**
  * Push a value to the end of a JavaScript array.
  */
-errcode
+errcode WARN_UNUSED
 JsArray_Push(JsRef idobj, JsRef idval);
 
+/**
+ * Same as JsArray_Push but panics on failure
```
rth, line 199: So if I understand correctly, with `JsArray_Push` we catch the JS exception, return the error code check it and then panic on failure.

While this panics directly?

rth:
Actually what's not very clear to me what's the general rule for choosing between,
 - catching (or not catching) the JS error and panicking with a fatal error
 - catching the JS error and re-throwing it in Python
 
I get that during init, we don't have a choice, and we would panic. But otherwise wouldn't it be generally better to go for option 2. Unless the failure is a sign of an unsound implementation, in which case we panic?

src/core/hiwire.h
hoodmane: Exactly.

hoodmane:
@rth Yeah.

> I get that during init, we don't have a choice, and we would panic. But otherwise wouldn't it be generally better to go for option 2. Unless the failure is a sign of an unsound implementation, in which case we panic?

Most internal commands that return booleans should not do error handling -- instead, they should panic on error and then be designed not to panic when used properly. 

This is consistent with the way that CPython works -- most CPython APIs that return booleans panic when used incorrectly and do no error handling. The reason for this is simple: it is very inconvenient to use these commands if they do error handling. Compare:
```C
if(PyDict_Check(o) || PyList_Check(o)){
    // Do stuff
}
```
with:
```C
int flag = PyDict_Check(o);
FAIL_IF_MINUS_ONE(flag);
if(!flag){
   flag = PyDict_Check(o);
}
FAIL_IF_MINUS_ONE(flag);
if(flag) {
   // Do stuff
}
```
If the command does error handling, then it tests as true when there is an error:
```C
if(get_boolean_with_error_handling(o)){
   // This code will run if get_boolean_with_error_handling raised an error or returned true
   // If an error happened, we leave error flag set which may cause weird stuff later.
} else {
   // false case
}
```
And indeed I consistently forgot to check for errors with our handled boolean-returning hiwire functions. That went okay because they were also designed with it in mind that they needed to be infallible.

hoodmane:
I could see an argument that we should get rid of `JsArray_Push_unchecked` and just always check the return value. But I think `JsArray_Push` is itself very very unlikely to fail in any case.

rth:
Thanks for the explanations! Makes sense. 

Also I think the fact that we are seeing very little reports about fatal failures is an indication that this approach works.

PR #1951, opened 2021-11-12
MAINT Avoid catching fatal errors in EM_JS wrappers

hoodmane:
If we raise a fatal error, we don't want to catch it later.


rth (commented):
It would be good to add a test for this, possibly with a function that raises a fatal error in a EM_JS, similar to how we have `_pyodide_core.trigger_fatal_error`.

Otherwise LGTM, thanks!

rth:
Also it's so much nicer to work with PRs when CI doesn't randomly fail (or at least not as often)!

Issue #1963, opened 2021-11-15
compile error, wasm-ld complain unknown argument


hzhangxyz:
## 🐛 Bug

Compile error: 
```
wasm-ld: error: unknown argument: --sort-common
wasm-ld: error: unknown argument: --as-needed
wasm-ld: error: unknown argument: --sort-common
wasm-ld: error: unknown argument: --as-needed
```

### To Reproduce

make the project

### Expected behavior

make succesful

### Environment

- Commit hash of Pyodide git repository: (f58d9c1)
- Build environment: arch linux

I compile with command:
```
PYODIDE_JOBS=5 PYODIDE_PACKAGES="micropip,numpy" PATH=$PATH:$PWD/../f2c/src make
```

### Additional context

make call many things I don't know about, but the last, the error is raised by `wasm-ldd`,
`wasm-ldd` is called by `emcc` which arguments contains:
```
-Wl,-O1,--sort-common,--as-needed,-z,relro,-z,now
 -fno-semantic-interposition -Wl,-O1,--sort-common,--as-needed,-z,relro,-z,now
```

It seems when pyodide run setup.py, it import my system's config, while in my system
```
/usr/lib/python3.9/_sysconfigdata__linux_x86_64-linux-gnu.py
```
have a line:
```
 'LDSHARED': 'gcc -pthread -shared '
             '-Wl,-O1,--sort-common,--as-needed,-z,relro,-z,now '
             '-fno-semantic-interposition '
             '-Wl,-O1,--sort-common,--as-needed,-z,relro,-z,now',
```

Is there any former mistake happened? the `emcc` argument should not contains parameter from my system's config?
Or I need to change my system's config to be compatible with wasm tool chain?

hzhangxyz:
Maybe `pywasmcross.py:handle_command` need to dealing with these wasm unknown argument?

rth:
On Linux if you don't want to deal with this, the easiest is to use Docker (cf [build instructions](https://pyodide.org/en/latest/development/building-from-sources.html) updated today).

> Maybe pywasmcross.py:handle_command need to dealing with these wasm unknown argument?

Otherwise I'm not sure why Arch config  adds those, but yes we can ignore them in `handle_command`. A PR would be welcome. Also if you find build instructions need improvement, please let us know.

ryanking13:
> Otherwise I'm not sure why Arch config adds those, but yes we can ignore them in `handle_command`.

[...]

PR #1965, opened 2021-11-16
Fix wasm-ld unknown arguments error (#1963)

hzhangxyz:
### Description

In ArchLinux, `--sort-common` and `--as-needed` was added
into `LDSHARED` of sysconfigdata, which cannot be recognized
by wasm-ld, so it is needed to remove them manually.

the only change is `pywasmcross.py:handle_command`,
I add two check for argument started with "-Wl", which
cannot be dealed with by wasm-ld

### Checklists

### DIFF
--- a/pyodide-build/pyodide_build/pywasmcross.py
+++ b/pyodide-build/pyodide_build/pywasmcross.py
@@ -313,11 +313,19 @@ def handle_command(line, args, dryrun=False):
         # some gcc flags that clang does not support actually
         if arg == "-Bsymbolic-functions":
             continue
-        if arg == "-Wl,-Bsymbolic-functions":
-            continue
-        # breaks emscripten see https://github.com/emscripten-core/emscripten/issues/14460
-        if arg == "-Wl,--strip-all":
-            continue
+        # ignore some link flags
+        # it should not check if `arg == "-Wl,-xxx"` and ignore directly here,
+        # because arg may be something like "-Wl,-xxx,-yyy" where we only want
+        # to ignore "-xxx" but not "-yyy".
+        if arg.startswith("-Wl"):
+            arg = arg.replace(",-Bsymbolic-functions", "")
+            # breaks emscripten see https://github.com/emscripten-core/emscripten/issues/14460
+            arg = arg.replace(",--strip-all", "")
+            # wasm-ld does not regconize some link flags
+            arg = arg.replace(",--sort-common", "")
+            arg = arg.replace(",--as-needed", "")
+            if arg == "-Wl":
+                continue
         # threading is disabled for now
         if arg == "-pthread":
             continue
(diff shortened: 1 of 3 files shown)


pyodide-build/pyodide_build/pywasmcross.py
```diff
@@ -313,11 +313,15 @@ def handle_command(line, args, dryrun=False):
...
-            continue
+        if arg.startswith("-Wl"):
+            arg = arg.replace(",-Bsymbolic-functions", "")
+            # breaks emscripten see https://github.com/emscripten-core/emscripten/issues/14460
+            arg = arg.replace(",--strip-all", "")
+            # wasm-ld does not regconize some link flags
+            arg = arg.replace(",--sort-common", "")
+            arg = arg.replace(",--as-needed", "")
+            if arg == "-Wl":
+                continue
```
hoodmane, line 324: I think this is a bit confusing. Maybe instead:
```py
    ignored_flags = [
        "-Bsymbolic-functions",
        "-Wl,-Bsymbolic-functions",
        "-Wl,--sort-common",
        "-Wl,--as-needed",
        # breaks emscripten see https://github.com/emscripten-core/emscripten/issues/14460
        "-Wl,--strip-all",
        # threading is disabled for now
        "-pthread",
        # On Mac, we need to omit some darwin-specific arguments
        "-bundle",
        "-undefined",
        "dynamic_lookup",
        # only the main module should link libffi, side modules are able to
        # access libffi functions because it is linked to main module.
        "-lffi",
        # this only applies to compiling fortran code, but we already f2c'd
        "-ffixed-form",
        # This flag is needed to build numpy with SIMD optimization but it
        # causes problems if SIMD isn't enabled
        "-mpopcnt",
    ]


    # Go through and adjust arguments
    for arg in line[1:]:
        if arg in optflags_valid and optflag is not None and arg != optflag:
            # There are multiple contradictory optflags provided, use the one
            # from cflags/cxxflags/ldflags
            continue

        if arg in ignored_flags:
            continue
```

hzhangxyz: I think -Wl need special handle, since user may write `-Wl,-xxx,-yyy,-zzz`, where we want to ignore xxx and yyy but not zzz

hoodmane: Got it.

hoodmane: Could you add a comment saying that?

hzhangxyz: OK, added

hzhangxyz:
> Thanks! Could you please add a test to `pyodide_build/tests/test_pywasmcross.py` (or extend an existing one there)?

Is this ok? I put the test into a new function.

rth (approved):
Yes, this works. Thanks!

rth:
For some reason CircleCI is not triggering anymore on new contributor's PRs. I couldn't see anything related in project settings. Here linting was actually failing but we couldn't see it. Maybe we should move more of those tests to GH Actions.

Fixed linting in https://github.com/pyodide/pyodide/commit/075116857f68e06259da39cae4ee10742ea8123c

PR #1969, opened 2021-11-16
ENH Makefile targets for cpython partial and full clean

phorward:
### Description

This PR is a preliminary for several other PRs I plan regarding Pyodide binary size reduction. It introduces further targets to partially or fully clean cpython.

- `make clean-python` cleans the cpython builds (as before, but also all sources)
- `make clean-all` cleans the cpython builds, cpython downloads and emsdk

The previous `make -C cpython clean` did only clean the Python build and install, but not the patched sources. Another run of `make -C cpython` then failed because of already patched sources in `$(ROOT)/build/Python-$(PYVERSION)`.

### Checklists

I think there are no further documentation and checks required on this, if you think different please let me know.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -16,9 +16,9 @@ all: check \
 	build/pyodide.asm.js \
 	build/pyodide.js \
 	build/console.html \
-	build/test.data \
 	build/distutils.data \
 	build/packages.json \
+	build/test.data \
 	build/test.html \
 	build/webworker.js \
 	build/webworker_dev.js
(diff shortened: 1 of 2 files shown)


rth (approved):
Could you restart CI? Otherwise LGTM, thanks @phorward !

Yeah, our 'clean' commands are very partial, and in the end I often use `git clean -xdf` to be sure there are no leftover build artifacts. Maybe we should document this (not in this PR).

rth:
> The previous make -C cpython clean did only clean the Python build and install, but not the patched sources.

I think this maybe have had something to do with the fact that we previously had a host and target builds of CPython, which is no longer relevant.

phorward:
> Could you restart CI? Otherwise LGTM, thanks @phorward !

Well, I tried, but I'm always getting `error cloning repository: ssh: handshake failed: knownhosts: key mismatch`? Is it known how to solve it?

rth:
Hmm, not sure. Anyway it should be fine. Merging thanks!

Issue #1932, opened 2021-11-06
Use long lived branches for specialized features


rth:
I wonder if using long lived branches for specialized features would make sense for Pyodide.

For instance, we can't use threading  https://github.com/pyodide/pyodide/issues/237 until all major browsers support it. At the same some people might want to experiment with it already now. So ideally we could have a build with threading and a build without. Doing both in our CI would be complex (and resource intensive) while I think having a separate upstream `threading` branch that would deploy to `cdn.jsdelivr.net/pyodide/dev/threading/` should be more manageable.  What's a bit non obvious there is now to sync releases from different branches, if we make them (maybe tag as v0.19.0-threading for instance?) but this could be solved once we get there. 

Other things that I wouldn't mind deploying on a separate branch is builds without fpcast (https://github.com/pyodide/pyodide/pull/1677). If people start using this, without it being the main deployment, I think we would feel more confident with integrating it into main.

Also maybe another feature branch for a very minimal Pyodide build, with a smallest size imprint we can manage. Ideally size optimizations should be opt-in in the main branch, bit I'm not sure all could be. cc @phorward

Of course having too many of those, would require more work on syncing them and resolving conflicts. But maybe we could find some middle ground.

ryanking13:
I like the idea. And I think Synchronous I/O (#1503) feature can be also developed in a separate branch.

phorward:
Hello @rth, thanks for starting the discussion. I like this idea as well, and maybe there will be some time to merge the different branches to a major release finally (as with the threading-feature).

> Also maybe another feature branch for a very minimal Pyodide build, with a smallest size imprint we can manage. Ideally size optimizations should be opt-in in the main branch, bit I'm not sure all could be. cc @phorward

Indeed. I still have this pull request outstanding, but I need to gather and collect some more information first about cost and usage first. Reducing size also means reducing features. The pull request has been created in a way that it can co-exists in the main branch and just is another Makefile target.

PR #1970, opened 2021-11-16
Use python-minifier to minify the Python stdlib for saving about 2.4 MB

phorward:
### Description

This pull request uses [python-minifier](https://github.com/dflook/python-minifier) to minify the Python standard library that is shipped with Pyodide by just removing literal statements. These are any literal values that stand on their own and are not used, including doc-strings or code that is "multi-line commented" as a string somewhere in the code.

To try it out using the Docker build, run these commands first:

```bash
$ pip install python-minifier
$ export PATH=/src/.docker_home/.local/bin:$PATH
```

The Pyodide docker image should be rebuilt with python-minifier pre-installed in case this pull request is being accepted.

#### make

This is the result of a current, ordinary main branch `make` run.
```
5.2M  pyodide.asm.data
 12M  pyodide.asm.wasm
```

#### PYODIDE_MINIFY=1 make

Running `PYODIDE_MINIFY=1 make` with this pull request invokes the standard library minimization using python-minifier (pyminify-command), and saves about 2.4 MB in total.

```
2.8M  pyodide.asm.data
 12M  pyodide.asm.wasm
```

#### --with-doc-strings / --without-doc-strings

The pull request contains also flagging `--without-doc-strings` as option to CPython's configure-script. The result is nearly the same, but the `--without-doc-strings`-version is 323343 bytes smaller in total, which is significantly much.

##### --with-doc-strings (default)
```
 2905177  pyodide.asm.data
12180003  pyodide.asm.wasm
 1995476  pyodide.asm.js
```

##### --without-doc-strings
```
 2904974  pyodide.asm.data
11856660  pyodide.asm.wasm
 1995473  pyodide.asm.js
```

#### Conclusion

This feature should just be an option to save some download bandwidth, especially for cases where no active Python development is wanted, but just a Python interpreter is required running in the browser. The flag `PYODIDE_MINIFY=1` is __not__ intended to be a standard setting.

What's your opinion on this?

### Checklists

- [ ] Run full test-suite with `PYODIDE_MINIFY=1 make`
- [ ] Update [pyodide/pyodide-env](https://hub.docker.com/r/pyodide/pyodide-env) to preinstalled `python-minifier`
- [ ] Add new documentation

### DIFF
--- a/cpython/Makefile
+++ b/cpython/Makefile
@@ -47,6 +47,14 @@ $(INSTALL)/lib/$(LIB): $(BUILD)/$(LIB) remove_modules.txt
 	rmdir `cat pybuilddir.txt`
 	rm pybuilddir.txt
 	cd $(INSTALL)/lib/python$(PYMAJOR).$(PYMINOR)/ && rm -rf `cat $(ROOT)/remove_modules.txt`
+	\
+	# Minify the Python standard library using pyminify
+	if [ "$(PYODIDE_MINIFY)" == "1" ]; then \
+		for py in `find $(INSTALL)/lib/python$(PYMAJOR).$(PYMINOR) -name "*.py"`; \
+			do echo "Minifying $${py}..."; \
+			pyminify --remove-literal-statements $${py} > $${py}.min; mv $${py}.min $${py}; \
+		done \
+	fi
 
 
 clean:
@@ -140,6 +148,7 @@ $(BUILD)/Makefile: $(BUILD)/.patched $(ZLIBBUILD)/.configured $(INSTALL)/lib/lib
 		  ./configure \
 			  CFLAGS="${PYTHON_CFLAGS}" \
 			  CPPFLAGS="-I$(SQLITEBUILD) -I$(BZIP2BUILD) -I$(ZLIBBUILD)" \
+			  $(if $(PYODIDE_MINIFY) == "1", --without-doc-strings, ) \
 			  --without-pymalloc \
 			  --disable-shared \
 			  --disable-ipv6 \
--- a/requirements.txt
+++ b/requirements.txt
@@ -18,3 +18,4 @@
   pytest-rerunfailures
   pytest-xdist
   selenium==4.0.0.b3
+  python-minifier


ryanking13:
> This feature should just be an option to save some download bandwidth, especially for cases where no active Python development is wanted, but just a Python interpreter is required running in the browser. The flag `PYODIDE_MINIFY=1` is **not** intended to be a standard setting.

Interesting. 2.4M size reduction is huge. Do you think many people can use minimized version for production just like js libraries? And will there be possible downsides?

rth:
Thanks for working on this @phorward !

There are a couple of test failures in the test suite that need investigating, but overall it looks quite promising!
+1 for the proposed approach. 

I think we would need to also figure out how to deploy it under a different path, for instance `/pyodide/dev/min/` as a start. As I think if's not deployed people will not use it. Setting up parallel CI for this sounds complicated, and we don't want to build it for each PR anyway. What we could do is to have this as an opt-in version on main, then have a separate long-lived branch (#1932 ) where it's enabled and deployed under a different URL that we sync with master manually from time to time. It could also be tagged as say `v0.19.0-min` for the release.

It would also be nice to run the modification on packages, but maybe in a follow-up PR.

> And will there be possible downsides of using minified version?

If you don't need docstrings, I think the main limitation is that tracebacks become a bit harder to work with: you lose the original variables names, and line positions. But it's still manageable.

phorward:
> There are a couple of test failures in the test suite that need investigating, but overall it looks quite promising! +1 for the proposed approach.

Yes, I'll take a look at this next. I'm currently unable to run the test suite locally on Docker using `make test`, even with a standard main branch checkout, are there any known reasons for this?
 
> I think we would need to also figure out how to deploy it under a different path, for instance `/pyodide/dev/min/` as a start. As I think if's not deployed people will not use it. Setting up parallel CI for this sounds complicated, and we don't want to build it for each PR anyway. 

Well this is a point I didn't keep in mind, yet. But you're right, people won't be using it when there is no deployed version of it. Maybe such a deployment should only be done for a release later on? 

> What we could do is to have this as an opt-in version on main, then have a separate long-lived branch (#1932 ) where it's enabled and deployed under a different URL that we sync with master manually from time to time. It could also be tagged as say `v0.19.0-min` for the release.

Sounds good. The origin of this pull request are experiments from [a fork we call *pyodide-nano*](https://github.com/phorward/pyodide). This is some kind of a "long-lived" branch, which is continuously updated from the upstream. pyodide-nano builds a Pyodide without mysql, bzip2 and zlib, a minified and reduced standard library and some compiler-flags (including `-Os`like in #1882)  that results in an amazing total size shown below.
```
1.3M  pyodide.asm.data
1.2M  pyodide.asm.js
4.6M  pyodide.asm.wasm
```
Due to its reduced standard library, it is not capable to run the Python scientific stack anymore, nor other packages making use of them. I want to merge some of these experimental results into this minified version of Pyodide starting in this PR now bit by bit, and possibly without loss of features or standard library modules.

> It would also be nice to run the modification on packages, but maybe in a follow-up PR.

We had the same idea here. But first, let's try to get this PR stable.

> > And will there be possible downsides of using minified version?
> 
> If you don't need docstrings, I think the main limitation is that tracebacks become a bit harder to work with: you lose the original variables names, and line positions. But it's still manageable.

That's correct. The main reason is to provide a Python interpreter running end-user applications, which we do in our company successfully with the pyodide-nano fork mentioned above. I'm also trying to improve python-minifier itself, as some doc-strings aren't removed due usage of them in the module (this is the case with ssl.py or ast.py in cpython/Lib).

ryanking13:
> Maybe such a deployment should only be done for a release later on?

I like this approach.

phorward:
> > Maybe such a deployment should only be done for a release later on?
> 
> I like this approach.

Sure, but you need some kind of test-suite to keep working builds up. Mhh...

PR #1974, opened 2021-11-18
MAINT remove `--memory-init-file 0` and consolidate link flags

### LINKED ISSUE #1972: Why `--memory-init-file 0`?
We've had `--memory-init-file 0` since the third commit:
https://github.com/pyodide/pyodide/blob/191ca2c7bd4ba6a05ba87f5427e89ea3ef5bdddc/Makefile#L14

[The emscripten docs about this flag](https://emscripten.org/docs/tools_reference/emcc.html#emcc-memory-init-file) say "Emit a separate memory initialization file in binary format. This is more efficient than storing it as text inside JavaScript but does mean you have another file to publish. The binary file will also be loaded asynchronously, which means main() will not be called until the file is downloaded and applied; you cannot call any C functions until it arrives."

We already have to wait on the file system data in `pyodide.asm.data` so it seems like it might be harmless to also have to download a memory init file.

hoodmane:
Remove `--memory-init-file 0` and consolidate all linker flags into Makefile.envs `MAIN_MODULE_LDFLAGS`.
Resolves #1972.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -46,19 +46,8 @@ build/pyodide.asm.js: \
 	$(CPYTHONLIB)
 	date +"[%F %T] Building pyodide.asm.js..."
 	[ -d build ] || mkdir build
-	$(CXX) -s EXPORT_NAME="'_createPyodideModule'" -o build/pyodide.asm.js $(filter %.o,$^) \
-		$(MAIN_MODULE_LDFLAGS) -s FORCE_FILESYSTEM=1 \
-		-lidbfs.js \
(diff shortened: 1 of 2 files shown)


ryanking13:
I thought removing `--memory-init-file` flag would generate a new `.mem` file as an output. But I don't see it in the build output.
Am I understanding something wrong?

rth (approved):
Thanks! The `test-js` also fails on main now which is unrelated.

PR #1975, opened 2021-11-18
MAINT Simplify shared library preload plugin

hoodmane:
I factored out `addPackage` from `recursiveDependencies` so that instead of closing over `toLoad`, it takes it explicitly as an argument. I think this improves readability.

I also factored out the "sharedLibraryPreloadPlugin" from `loadPackage` into helper functions `useSharedLibraryWasmPlugin` and `restoreOrigWasmPlugin`. These are idempotent, so we can also call `restoreOrigWasmPlugin` in the finally block just in case. Also, I made the `sharedLibraryWasmPlugin` a normal object instead of a `Proxy`. I think the original `Proxy` looked intimidating. It wasn't clear to me that the plugin only had two methods `canHandle` and `handle`.

### DIFF
--- a/src/js/load-pyodide.js
+++ b/src/js/load-pyodide.js
@@ -95,6 +95,23 @@ if (globalThis.document) {
   throw new Error("Cannot determine runtime environment");
 }
 
+function addPackageToLoad(name, toLoad) {
+  name = name.toLowerCase();
+  if (toLoad.has(name)) {
+    return;
+  }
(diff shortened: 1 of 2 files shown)


rth:
Thanks ! Avoiding those nested definitions is indeed much clearer. +1 to merge when CI passes, as I think the test-js CI is also broken in other PRs now.

hoodmane:
I haven't run this locally so it's possible it still doesn't work.

Issue #1966, opened 2021-11-16
micropip should report all dependent packages without wheels in one go


yuvipanda:
## 🚀 Feature

When installing packages via `micropip`, if there are dependents that don't have pure python wheels, it errors out on the first such dependent. However, it would be much nicer if it could instead collect *all* the dependents that don't have wheels, and print them out in one go before exiting

### Motivation

I was trying to get the [nbconvert](https://github.com/jupyter/nbconvert/) package installed on pyiodide. It failed with pyristant not having pure python wheels. So I [made some](https://github.com/yuvipanda/pyiodide-nbconvert/releases/tag/pyrsistent-0.18.0) and was excited to try it again. But this time it now errors with [pyzmq](https://github.com/zeromq/pyzmq) not having any wheels. And now I'm not sure which other packages will error out!

### Pitch

When I run `micropip.install('nbconvert')`, I want it to tell me exactly which dependent packages are causing installation to fail.

P.S. AMAZING PROJECT, MUCH <3 TO EVERYONE INVOLVED

rth:
Thanks for the feedback! I agree that this would be much more user friendly.

We could have a look at how to implement it. One thing to keep in mind is that until PyPi has reliable JSON API for dependencies, we have to parse the dependencies from the wheel. So if there is no pure Python wheel for a package, we can't get its dependencies.

Otherwise this change would likely mean to keep going with the installation even if a wheel is missing, then report all the errors at the end.

ryanking13:
> Otherwise this change would likely mean to keep going with the installation even if a wheel is missing, then report all the errors at the end.

If we want to keep the behavior of `micropip.install` failing fast, we may design a new API such as `micropip.installable` which checks missing wheels, not sure which one is better.

rth:
> we may design a new API such as micropip.installable

I think it might be better to keep the API surface to a minimum for future maintenance (and to avoid confusion). However indeed I'm not sure what would be best behavior on missing wheel between,
 - keep going and install the rest
 - keep going and download the rest of the dependencies (to get their respective dependencies) but don't install them on disk.
 
before finally reporting all the errors. Also it could be an option, between keep going or fail on first error.

yuvipanda:
> keep going and download the rest of the dependencies (to get their respective dependencies) but don't install them on in the file system.

As a user, I like this option! It might mean you're downloading things that you then can't install, but that seems ok.

bollwyvl:
[...]

PR #1976, opened 2021-11-18
ENH Add `keep_going` parameter to `micropip.install()`

ryanking13:
### Description

Add `keep_going` parameter to `micropip.install()` which make micropip report all (identifiable) dependent non-pure Python packages when installation failed.

Closes: #1966

### Checklists

- [x] Add / update tests

### DIFF
--- a/docs/project/changelog.md
+++ b/docs/project/changelog.md
@@ -108,8 +108,9 @@ substitutions:
 - {{Fix}} micropip now raises error when installing non-pure python wheel directly from url.
   {pr}`1859`
 
-- {{Enhancement}} When installation failed, micropip now shows all identifiable
-  dependents that don't have pure python wheels.
+- {{Enhancement}} micropip.install() now accepts `keep_going` parameter that, if set to True,
+  make micropip report all identifiable dependencies that don't have pure Python wheels when
+  installation has failed.
   {pr}`1976`
 
 ### packages
--- a/docs/project/changelog.md
+++ b/docs/project/changelog.md
@@ -108,9 +108,9 @@ substitutions:
 - {{Fix}} micropip now raises error when installing non-pure python wheel directly from url.
   {pr}`1859`
 
-- {{Enhancement}} micropip.install() now accepts `keep_going` parameter that, if set to True,
-  make micropip report all identifiable dependencies that don't have pure Python wheels when
-  installation has failed.
+- {{Enhancement}} {func}`micropip.install` now accepts a `keep_going` parameter. If set to True,
+  micropip reports all identifiable dependencies that don't have pure Python wheels, instead of
+  failing after processing the fist one.
   {pr}`1976`
 
 ### packages
--- a/docs/project/changelog.md
+++ b/docs/project/changelog.md
@@ -110,7 +110,7 @@ substitutions:
 
 - {{Enhancement}} {func}`micropip.install` now accepts a `keep_going` parameter. If set to True,
   micropip reports all identifiable dependencies that don't have pure Python wheels, instead of
-  failing after processing the fist one.
+  failing after processing the first one.
   {pr}`1976`
 
 ### packages
--- a/docs/project/changelog.md
+++ b/docs/project/changelog.md
@@ -108,6 +108,10 @@ substitutions:
 - {{Fix}} micropip now raises error when installing non-pure python wheel directly from url.
   {pr}`1859`
 
+- {{Enhancement}} When installation failed, micropip now shows all identifiable
+  dependents that don't have pure python wheels.
+  {pr}`1976`
+
 ### packages
 
 - {{ Enhancement }} Unit tests are now unvendored from Python packages and
--- a/packages/micropip/src/micropip/micropip.py
+++ b/packages/micropip/src/micropip/micropip.py
@@ -152,6 +152,7 @@ class _PackageManager:
             "wheels": [],
             "pyodide_packages": [],
             "locked": dict(self.installed_packages),
+            "failed": [],
         }
         requirement_promises = []
         for requirement in requirements:
@@ -164,6 +165,15 @@ class _PackageManager:
 
     async def install(self, requirements: Union[str, List[str]], ctx=None):
         transaction = await self.gather_requirements(requirements, ctx)
+
+        if transaction["failed"]:
+            failed_requirements = ", ".join(
+                [f"'{req}'" for req in transaction["failed"]]
+            )
+            raise ValueError(
+                f"Couldn't find a pure Python 3 wheel for: {failed_requirements}"
+            )
+
         wheel_promises = []
         # Install built-in packages
         pyodide_packages = transaction["pyodide_packages"]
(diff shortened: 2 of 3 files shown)


packages/micropip/test_micropip.py
```diff
@@ -236,3 +236,22 @@ def test_install_mixed_case2(selenium_standalone_micropip, jinja2):
...
     )
+
+
+def test_report_all_failed_dependencies():
+    pytest.importorskip("packaging")
+    from micropip import micropip
+
+    pkg = "nonpure-dummy"  # dummy package which requires both tensorflow and torch
+
+    msg = "tensorflow"
```
rth, line 247: You can use regex here, to check both and avoid making the call twice.

packages/micropip/test_micropip.py
```diff
@@ -236,3 +236,22 @@ def test_install_mixed_case2(selenium_standalone_micropip, jinja2):
         `);
         """
     )
+
+
+def test_report_all_failed_dependencies():
+    pytest.importorskip("packaging")
+    from micropip import micropip
+
```
rth, line 244: Could you try [monkeypatching](https://docs.pytest.org/en/6.2.x/monkeypatch.html) the http request to return the expected JSON response for this test? Unless it ends up being too complicated. You don't need to return the full JSON, just the fields that are used the dependency resolution. It doesn't need to be pytest/tensorflow could be an arbitrary fake package.

I know we have a few tests that make HTTP calls to PyPi, but it's really not ideal to do this systematically.

packages/micropip/src/micropip/micropip.py
```diff
@@ -263,7 +276,7 @@ def find_wheel(self, metadata, req: Requirement):
                 if _is_pure_python_wheel(fileinfo["filename"]):
                     return fileinfo, ver
 
-        raise ValueError(f"Couldn't find a pure Python 3 wheel for '{req}'")
+        return None, None
```
rth, line 279: Could you please also add type annotations for the output, and a docstring explaining in particular when you would get the `None, None` output? We use the [numpy docstring style](https://numpydoc.readthedocs.io/en/latest/format.html)

Not sure why those methods were not documented, but it's probably a good idea to add some docs at least for this one.

rth (commented):
Nice! A few comments are below. Please add a changelog as well.

packages/micropip/test_micropip.py
rth: It could look approximately like the following,
```py
async def _mock_fetch_string(url, **kwargs):
    if url.endswith("package1"):
        return "JSON encoded as a string"
    elif url.endswith("package1")
       return "another JSON"

def test_report_all_failed_dependencies(monkeypatch):
    pytest.importorskip("packaging")
    monkeypatch.setattr('micropip.micropip.fetch_string', _mock_fetch_string)
    ...
```

The returned  JSON can be similar aside from the few relevant fields that matter for this test

ryanking13:
Thanks for the review! Applied:

- type annotation / docstring for `find_wheel()` method
- mocking PyPI request
  - I mocked requests on dependencies, but the one initial request on the wheel file (`nonpure-dummy`) is still happening.
- use regex to match error

rth:
> I mocked requests on dependencies, but the one initial request on the wheel file (nonpure-dummy) is still happening.

thanks! Would it be difficult to also monkeypatch that first call? In particular, so we don't have to re-upload something to PyPi if we want to change something in this test.

ryanking13:
> Would it be difficult to also monkeypatch that first call?

I will try that (serving wheel files) with something like:

```py
def _mock_fetch_wheel():
    wheel_bytes = b'<raw_bytes_of_wheel_file>'
    return base64.b64decode(wheel_bytes)
```

rth:
Yeah, you can probably create the wheel at runtime by adding the required files zip archive written to a `BytesIO` object. You can unzip the wheel to see the files inside. Normally you only need one or two of those config files (and only the relevant fields of those).

rth:
Actually, might be easier to monkeypatch `pkg_resources_distribution_for_wheel` with the expected to avoid dealing with encoding/decoding etc of the wheel.  You can add a `breakpoint()` in there to see the current value in the debugger when running tests, and monkeypatch to that.

ryanking13:
Updated test and no requests to the PyPI are happening now.

Maybe mocking PyPI requests can be separated into a fixture for general usage later. But I didn't do it in this PR, since there is only one test using it.

ryanking13:
@rth https://github.com/pyodide/pyodide/issues/1966#issuecomment-975602211

> I actually feel like the current fail-fast is a sensible default, with a (non-tracebackful) message saying how to change the invocation keep_going=True (or fail_fast=False). One or two, sure, maybe you can get around it, but a heavy ML library or something is going to pop 100 tightly-pinned, incompatible packages after downloading 200mb.

How do you think of having a `failfast` parameter (and its default value)?

hoodmane:
I agree that the default should be the current behavior. No need to download extra stuff just to report how much it won't work.

rth:
+1 for `keep_going=False` by default. Open to other names. Maybe we should check what this parameter is called in other package managers. emerge (Gentoo) does have a `--keep-going` for instance.

ryanking13:
Added `keep_going` parameter with default value `False`. So the default behavior of micropip.install() is not changed unless `keep_going=True` is explicitly given.

docs/project/changelog.md
```diff
@@ -108,6 +108,11 @@ substitutions:
 - {{Fix}} micropip now raises error when installing non-pure python wheel directly from url.
   {pr}`1859`
 
+- {{Enhancement}} micropip.install() now accepts `keep_going` parameter that, if set to True,
+  make micropip report all identifiable dependencies that don't have pure Python wheels when
+  installation has failed.
```
rth, line 113: Minor improvement,

```suggestion
- {{Enhancement}} {func}`micropip.install` now accepts a `keep_going` parameter. If set to True,
  micropip reports all identifiable dependencies that don't have pure Python wheels, instead of
  failing after processing the fist one.
```

packages/micropip/src/micropip/micropip.py
```diff
@@ -142,7 +142,9 @@ class _PackageManager:
     def __init__(self):
         self.installed_packages = {}
 
-    async def gather_requirements(self, requirements: Union[str, List[str]], ctx=None):
+    async def gather_requirements(
+        self, requirements: Union[str, List[str]], ctx=None, keep_going=False
```
rth, line 146:
```suggestion
        self, requirements: Union[str, List[str]], ctx=None, keep_going: bool = False
```

packages/micropip/src/micropip/micropip.py
```diff
@@ -162,8 +166,19 @@ async def gather_requirements(self, requirements: Union[str, List[str]], ctx=Non
         await gather(*requirement_promises)
         return transaction
 
-    async def install(self, requirements: Union[str, List[str]], ctx=None):
-        transaction = await self.gather_requirements(requirements, ctx)
+    async def install(
+        self, requirements: Union[str, List[str]], ctx=None, keep_going=False
```
rth, line 170:
```suggestion
        self, requirements: Union[str, List[str]], ctx=None, keep_going: bool =False
```

rth (commented):
Thanks @ryanking13 ! A few minor comments otherwise LGTM.

ryanking13:
@rth Applied suggestions, thanks for the review!
Made micropip mock functions global, and made them little more flexible to use.

rth:
Thanks! There are some CI failures still.  

To fix `test-js` with https://github.com/pyodide/pyodide/pull/1980 please merge main in.

rth (approved):
There is no need for newlines before the parameter description in docstrings. Otherwise LGTM.

Issue #1979, opened 2021-11-19
`make test` in docker always fails due broken emsdk/tests


phorward:
## 🐛 Bug

Hey, is the emsdk/tests-suite broken or sort of? When I want to run the test suite locally in a docker container, it always fails.

`make test` produces this immediately on current main branch:
```
pytest src emsdk/tests packages/*/test* pyodide-build -v
============================= test session starts ==============================
platform linux -- Python 3.9.5, pytest-6.2.4, py-1.10.0, pluggy-0.13.1 -- /usr/local/bin/python
cachedir: .pytest_cache
hypothesis profile 'default' -> database=DirectoryBasedExampleDatabase('/src/.hypothesis/examples')
rootdir: /src, configfile: setup.cfg
plugins: hypothesis-6.14.2, asyncio-0.15.1, cov-2.12.1, forked-1.3.0, httpserver-1.0.0, instafail-0.4.2, rerunfailures-10.1, xdist-2.3.0
collecting ... collected 2653 items / 5 errors / 2 skipped / 2646 selected

==================================== ERRORS ====================================
____________________ ERROR collecting emsdk/tests/common.py ____________________
/usr/local/lib/python3.9/importlib/__init__.py:127: in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
<frozen importlib._bootstrap>:1030: in _gcd_import
    ???
<frozen importlib._bootstrap>:1007: in _find_and_load
    ???
<frozen importlib._bootstrap>:984: in _find_and_load_unlocked
    ???
E   ModuleNotFoundError: No module named 'tests.common'
_________________ ERROR collecting emsdk/tests/test_dyncall.py _________________
/usr/local/lib/python3.9/importlib/__init__.py:127: in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
<frozen importlib._bootstrap>:1030: in _gcd_import
    ???
<frozen importlib._bootstrap>:1007: in _find_and_load
    ???
<frozen importlib._bootstrap>:984: in _find_and_load_unlocked
    ???
E   ModuleNotFoundError: No module named 'tests.test_dyncall'
_________________ ERROR collecting emsdk/tests/test_dyncall.py _________________
ImportError while importing test module '/src/emsdk/tests/test_dyncall.py'.
Hint: make sure your test modules/packages have valid Python names.
Traceback:
/usr/local/lib/python3.9/importlib/__init__.py:127: in import_module
[...]

PR #1981, opened 2021-11-20
Better test discovery

rth:
This improves the pytest test discovery with the following points,
 - Running `pytest` with no arguments (or `pytest .`) now correctly collects all available tests (except for emsdk ones which require the compiler setup and have to be run separately with `make -C emsdk test`)
 - Running `pytest` on packages now skips package tests that are not built. This is done at the selenium fixture level and so will also apply to new tests.
 - `packages/test_packages_common.py::test_import` now will skip the test before loading the selenium_standalone fixture which is slow. Previously it was first loading the slow fixture, and then skipping.
 - removes `make test` as I think pytest should not be used via make, as then one loses access  to all useful flags (e.g. `--pdb`, `-n auto` (for parallelism), `-s` for printing stdout, etc)

Closes https://github.com/pyodide/pyodide/issues/1979

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -1,8 +1,9 @@
 """
 Various common utilities for testing.
 """
-
+import re
 import contextlib
+import functools
 import json
 import multiprocessing
 import textwrap
@@ -59,6 +60,11 @@ def pytest_configure(config):
         return result
 
     config.cwd_relative_nodeid = cwd_relative_nodeid
+
+
+@functools.cache
+def _package_is_built(package_name: str) -> bool:
+    return (BUILD_PATH / f"{package_name}.data").exists()
 
 
 class JavascriptException(Exception):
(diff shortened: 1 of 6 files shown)


conftest.py
```diff
@@ -463,8 +478,57 @@ def extra_checks_test_wrapper(selenium, trace_hiwire_refs, trace_pyproxies):
...
+    match = re.match(
+        r".*/packages/(?P<name>[\w\-]+)/test_[\w\-]+\.py", str(item.parent.fspath)
+    )
+    if match:
+        package_name = match.group("name")
+        if not _package_is_built(package_name):
+            skip_msg = f"package '{package_name}' is not built."
+
+    # Common package import test. Skip it if the package is not built.
+    if str(item.fspath).endswith("test_packages_common.py") and item.name.startswith(
```
phorward, line 497: Is possibly a test like this missing?
```suggestion
    if not skip_msg and str(item.fspath).endswith("test_packages_common.py") and item.name.startswith(
```

phorward (approved):
Hello @rth thanks for cleaning this up! I'm fine with this PR, lets wait for CircleCI to complete.

conftest.py
rth: Indeed it wouldn't hurt to check that.

phorward (approved):

PR #1991, opened 2021-11-23
Fix package rebuild check to detect micropip src change

ryanking13:
### Description

`needs_rebuild()` function which decides whether the package needs to be rebuilt is currently not working for packages that use source.path option in meta.yaml (for now, micropip is the only package that uses source.path).

### Checklists
- [x] Add / update tests


ryanking13:
It looks like the package building step is taking a longer time than usual, which is not related to this PR.

rth:
Could you please add a test for `source_files` which appears not to exist? Otherwise LGTM.

> It looks like the package building step is taking a longer time than usual, which is not related to this PR?

Yes, it timeouts occasionally.

ryanking13:
Added test for needs_rebuild() function!

PR #1992, opened 2021-11-23
Use cli command tsc to compile TypeScript files

grimmer0125:
### Description

Actually, I can not try `npx typescript --project src/js` successfully and I'm even not sure whether `typescript --project` is a valid CLI or not (can not find out on Google). 

But `tsc` is sure to be an official TypeScript CLI command to invoke compilation. 
- https://www.typescriptlang.org/docs/handbook/compiler-options.html
- https://code.visualstudio.com/docs/typescript/typescript-compiling 
- https://www.typescriptlang.org/tsconfig
- https://docs.microsoft.com/en-us/visualstudio/javascript/compile-typescript-code-npm?view=vs-2022

### Checklists

- [ ] Add / update tests

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -73,7 +73,7 @@ node_modules/.installed : src/js/package.json
 	touch node_modules/.installed
 
 build/pyodide.js: src/js/*.js src/js/pyproxy.gen.js node_modules/.installed
-	npx typescript --project src/js
+	npx tsc --project src/js
 	npx rollup -c src/js/rollup.config.js
 
 src/js/pyproxy.gen.js : src/core/pyproxy.* src/core/*.h


rth:
When I try,
```
$ npx tsc --help
```
I get,
```
Need to install the following packages:
  tsc
Ok to proceed? (y) n
```
however https://www.npmjs.com/package/tsc  as far as I can tell is not the typescript package?

rth:
Maybe we should install JS dependencies explicitly, without npx as suggested in https://github.com/pyodide/pyodide/pull/1770

grimmer0125:
`tsc` is in the CLI executable file/bin in the TypeScript package. If you executes `npx tsx` in a project which already install TypeScript and has `tsconfig.json` and it should work.

grimmer0125:
I do not use `npx` too much and I agree that the comment about #1770, for an existing project, usually people rely on npm `scripts` field, not `npx`.

rth:
>  If you execute npx tsx in a project which already installs TypeScript and has tsconfig.json and it should work.

OK, but I still find it suspicious that the installed package will depend on some configuration files. 

Let's rather move to properly install typescript and other JS dependencies before using them (i.e. not rely on npx)

grimmer0125:
> Let's rather move to properly install typescript and other JS dependencies before using them (i.e. not rely on npx)?

I add one more commit to use npm script: `npm run build`, the common way to execute `tsx`.  Is it OK?

bollwyvl:
Right, for reproducibility purposes, `npx` is kind of a nightmare.
 
Perhaps consider going even further: though I prefer `yarn` and `yarn.lock` (on jupyterlab, we've gone as far as to check in and white-label a `yarn` "binary"), having a checked-in `package-lock.json` is a superb entropy key for e.g. caching `node_modules`.

ryanking13:
I don't think this is needed anymore. Thanks for your work @grimmerk. Closing.

PR #1998, opened 2021-11-25
DOC Fix web worker api doc with concurrent Promises [skip ci]

### LINKED ISSUE #1996: Bug in web worker example with concurrent Promises
## 📚 Documentation

[The worker API](https://pyodide.org/en/stable/usage/webworker.html#the-worker-api) has a bug when `asyncRun` is called concurrently. Function `run` modifies `pyodideWorker.onmessage`, so only the last `Promise` could be resolved and wrongly get the first finished result.
```html
<html>
<head>
  <script src="https://cdn.jsdelivr.net/pyodide/v0.18.1/full/pyodide.js"></script>
</head>
<body>
  <script type="module">
    import { asyncRun } from "/py-worker.js";

    async function main() {
      const promise1 = asyncRun('1', {});
      const promise2 = asyncRun('2', {});
      const r2 = await promise2;
      console.log(r2);
      const r1 = await promise1;
      console.log(r1);
    }

    main();
  </script>
</body>
</html>
```
The output should be `Object { results: 2 }` and `Object { results: 1 }`
However in my Firefox, the output is `Object { results: 1 }`.

eagleoflqj:
### Description

Closes #1996

### DIFF
--- a/docs/usage/webworker.md
+++ b/docs/usage/webworker.md
@@ -119,7 +119,7 @@ self.onmessage = async (event) => {
   // make sure loading is done
   await pyodideReadyPromise;
   // Don't bother yet with this line, suppose our API is built in such a way:
-  const { python, ...context } = event.data;
+  const { id, python, ...context } = event.data;
   // The worker copies the context in its own "memory" (an object mapping name to values)
   for (const key of Object.keys(context)) {
     self[key] = context[key];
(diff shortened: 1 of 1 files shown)


rth:
Thanks @eagleoflqj ! We should probably move some of this code to the NPM package, in time, so it's easier to reuse.

PR #1999, opened 2021-11-26
MAINT use npm ci to make sure the consistent dependencies installed

grimmer0125:
### Description

This is to make sure the installed npm packages which are the same as `package-lock.json`. `npm ci` is included as of npm 5.7.1 (4 years ago) and is the recommended way for CI. `npm install` may install different versions from `package-lock.json`

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -68,7 +68,7 @@ env:
 
 
 node_modules/.installed : src/js/package.json
-	cd src/js && npm install --save-dev
+	cd src/js && npm ci
 	ln -sfn src/js/node_modules/ node_modules
 	touch node_modules/.installed
 
--- a/Makefile
+++ b/Makefile
@@ -67,7 +67,7 @@ env:
 	env
 
 
-node_modules/.installed : src/js/package.json
+node_modules/.installed : src/js/package.json package-lock.json
 	cd src/js && npm ci
 	ln -sfn src/js/node_modules/ node_modules
 	touch node_modules/.installed
--- a/Makefile
+++ b/Makefile
@@ -67,7 +67,7 @@ env:
 	env
 
 
-node_modules/.installed : src/js/package.json package-lock.json
+node_modules/.installed : src/js/package.json src/js/package-lock.json
 	cd src/js && npm ci
 	ln -sfn src/js/node_modules/ node_modules
 	touch node_modules/.installed
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -227,14 +227,14 @@ jobs:
           command: |
             cd src/js
             npx tsd
-            npm install
+            npm ci
             npm test
       - run:     
           name: check if webpack cli works well with load-pyodide.js
           command: | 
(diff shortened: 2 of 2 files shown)


rth:
Thanks @grimmer0125 LGTM !

FYI, for maintenance changes, we use the `MAINT` tag in the title. Also please edit the description to remove the non-relevant checklist items, as non-checked items mean you still plan to do them (and I think the current change is sufficient).

Makefile
```diff
@@ -68,7 +68,7 @@ env:
 
 
 node_modules/.installed : src/js/package.json
-	cd src/js && npm install --save-dev
```
rth, line 71: Maybe keep this as an install as it's not only going to be run from CI?

grimmer0125: I tried `npm install --save-dev` (without specifying any npm package) and its effect is the same as `npm install` which install dependencies and devDependencies. I think only `npm install --save-dev` somePackage` only makes the difference.

grimmer0125: I think the question becomes, among `npm install` or `npm ci`, which one is better for local dev installation? I did some search, and some people recommend using `npm ci` to install locked dependencies (after `git clone`) and only use `npm install`  to add or update some packages in the local environment. Not everyone says so but I have not found someone against this.

rth (approved):
Thanks!

> I think the question becomes, among npm install or npm ci, which one is better for local dev installation? I did some search, and some people recommend using npm ci to install locked dependencies (after git clone) and only use npm install to add or update some packages in the local environment. 

I have no objections, but let's wait for a second opinion from @hoodmane or @ryanking13

Makefile
```diff
@@ -68,7 +68,7 @@ env:
 
 
 node_modules/.installed : src/js/package.json
```
ryanking13, line 70: Maybe add package-lock.json as a dependency here for `npm ci`?

ryanking13 (commented):
Installing locked dependencies are mostly what we want, so I think using `npm ci` would be fine.

Makefile
grimmer0125: Sounds reasonable. I can add it.

hoodmane:
Yeah I'm also +1 for this.

grimmer0125:
> Thanks @grimmer0125 LGTM !
> 
> FYI, for maintenance changes, we use the `MAINT` tag in the title. Also please edit the description to remove the non-relevant checklist items, as non-checked items mean you still plan to do them (and I think the current change is sufficient).

Is "MAINT" only for PR title or also for a git commit?

Makefile
grimmer0125: Done.

rth:
Both, cf list of acronyms in https://numpy.org/doc/stable/dev/development_workflow.html#writing-the-commit-message but we can of course come up with other ones.

grimmer0125:
FYI: The commit message type (`chore`) I use is from https://www.conventionalcommits.org/en/v1.0.0/ is used by a part of the Python community and JS community.

Issue #2000, opened 2021-11-26
Review CPython patches / tests and contribute upstream


rth:
There is currently an active effort for adding wasm as the build target to CPython in https://bugs.python.org/issue40280 by Christian Heimes. So it would be an ideal timing to review again our CPython patches and failing tests and re-evaluate what could be contributed upstream. 

(Other things were already proposed upstream when possible).

tiran:
Hi,

I took some inspiration from your patch set. :)

* The patch files ``0005-add-emscripten-host.patch`` and ``0006-fix-Py_Sigset_Converter.patch`` should no longer be needed. I did not apply your patches directly, but I gave you credit. 
* The patch ``remove-duplicate-symbols-from-cfield.c.patch`` is causing issues on Windows, https://bugs.python.org/issue45898 .
* Patch ``0001-disable-set_inheritable.patch`` does not explain why the change is necessary. I would like to understand the reason for the patch first before we land it in upstream.
* I created a [CONFIG_SITE override](https://gist.github.com/tiran/5ccffa28723d3e4739db848451bd9efa), which is IMHO a cleaner approach than your ``pyconfig.undefs.h``.
* Eventually we will be able to replace the custom ``Setup.local`` with https://github.com/python/cpython/blob/main/Modules/Setup.stdlib.in and full dependency detection in configure. There are still some modules missing from the file, though.

ethanhs:
I've created a repo with my build scripts to build CPython (for now for emscripten but I hope to add work on WASI eventually...): https://github.com/ethanhs/python-wasm (based on @tiran's config.site and my own attempts).

I am happy to say I've successfully gotten CPython master running in the browser with emscripten!

The main trick up my sleeve is `emcc -o python.html Programs/python.o libpython3.11d.a -ldl -lm --preload-file .`, which loads the local build tree and the JS Emscripten system shims.

<img width="266" alt="Python master running in the browser" src="https://user-images.githubusercontent.com/9504279/143620240-7cf5245c-85cc-4e72-b2d6-c9919509e863.png">

Lots of things don't work though, like `help()` :P

brettcannon:
My hope is we eventually centralize all of this WASM build knowledge somewhere so that Pyodide can focus more on the "build the scientific stack" bit and the "build CPython for WASM/WASI" becomes a shared problem.

But obviously getting this all working upstream is the first step. 😅

rth:
Sounds great @tiran ! Thanks for writing this summary.

[...]

PR #2003, opened 2021-11-26
Remove disable-set-inheritable CPython patch

rth:
This patch was added in https://github.com/pyodide/pyodide/commit/f2c08119afdc52f069c6dc941d32701f66051a09 in 2018, but as far as I can tell it builds fine without it.

Related to https://github.com/pyodide/pyodide/issues/2000


hoodmane (approved):
Always great to remove a patch =)

Issue #1464, opened 2021-04-13
List of out of date packages


hoodmane:
The following pure Python packages with patches are out of date
* astropy v3.2.3 ==> 4.2.1
* joblib 0.11 ==> 1.0.1
* mpmath 1.1.0 ==> 1.2.1 
* numcodecs 0.7.2 ==> 0.7.3 
* ~pytest 3.6.3 ==> 6.2.3~ Updated by #1719.
* python-sat 0.1.6.dev6 ==> 0.1.7.dev1
* scikit-image 0.15.0 ==> 0.18.1   : *blocked by scipy*
* setuptools 40.0.0 ==> 56.0.0
* zarr 2.6.1 ==> 2.7.0

Following mixed packages are out of date:
* numpy 1.17.5 ==> 1.20.2
* libxml 4.4.1 ==> 4.6.3
* pillow 8.0.1 ==> 8.2.0
* pandas 1.0.5 ==> 1.2.4
* matplotlib 3.3.3 ==> 3.4.1
* scipy 0.17.1 ==> 1.6.2  : #1293
* scikit-learn 0.22.2 ==> 0.24.1   : *blocked by scipy*

teonbrooks:
`mne` is out of date as well but I've been thinking about whether it's necessary to keep it in the package list since it is a pure python package. 
I'm using pyodide in an Electron app and it's been convenient having `mne` bundled with the zip but I was thinking it might be useful to download the pypi wheel, and install it via micropip (I need the app to be able to run without internet so I can't load `mne` on the fly). It would help reduce the number of extraneous packages. 
just to confirm, can you load wheels locally? if so, would all i need to do is load pyodide in JS and do `micropip.install('./path/to/mne/mne_something.whl')`?

hoodmane:
mne will be updated to 0.22.1 in #1460.

> just to confirm, can you load wheels locally?

I believe that this works and we intend to support it. But if it turns out that it doesn't work for some reason, please open an issue. =)

teonbrooks:
ok cool :) I think I will wait til after the release to propose removing it. at the moment, it's harmless. just curious, is there a guiding principle to what gets added/removed to pyodide internal package manager? I assume requested non-pure py packages. this all might change if we worked to get conda(-forge) to make a wasm build target :)

hoodmane:
> is there a guiding principle to what gets added/removed to pyodide internal package manager? I assume requested non-pure py packages

[...]

PR #2005, opened 2021-11-28
Upgrade scikit-image to v0.19.1

oeway:
### Description

This PR upgrade the scikit-image package to v0.18.3.

### Checklists


rth:
The blocker on updating scikit-learn was the outdated version of scipy we have https://github.com/pyodide/pyodide/issues/1464 

What we could try, for instance, is to install scipy 0.17.1 locally with this version of scikit-image, see what test fails (`pytest --pyargs skimage`) and add backports for those if possible.

rth:
I synced with main which should now have Scipy 1.7.3 thanks to Hood. Let's see if it builds now.

rth:
```
skimage/_shared/geometry.c:645:10: fatal error: 'omp.h' file not found
#include <omp.h>
         ^~~~~~~
1 warning and 1 error generated
```
looks like some patches (or possibly env variables) to avoid building with OpenMP are still needed.

pyodide-build/pyodide_build/buildpkg.py
```diff
@@ -83,13 +83,17 @@ def close(self):
...
-        "NUMPY_LIB": os.environ["NUMPY_LIB"],
-        "PYODIDE": "1",
+        key: os.environ[key]
+        for key in [
+            "PATH",
+            "PYTHONPATH",
+            "PYODIDE_ROOT",
+            "PYTHONINCLUDE",
+            "NUMPY_LIB",
+            "PYODIDE_PACKAGE_ABI",
```
rth, line 93: It turns out the failure was because the fix I contributed upstream to scikit-image checked for the PYODIDE_PACKAGE_ABI variable, however, we deprecated it in favor of the `PYODIDE`. Re-adding that variable should fix the build.

hoodmane: It's nice that #2053 made it more explicit which environment variables we are passing down, but it does lead to some pain working out what is going on.

rth: Agreed, I like that it's explicit now.

docs/project/changelog.md
```diff
@@ -172,6 +171,9 @@ async function runPythonAsync(code, globals) {
 
 ### packages
 
+- {{ Update }} Upgrade `scikit-image` to 0.19.1
```
ryanking13, line 174:
```suggestion
- {{ Enhancement }} Upgrade `scikit-image` to 0.19.1
```

It looks like we are using `enhancement` badge for package update

hoodmane: Maybe it would make more sense to change them all to `Update`?

hoodmane:
Build failed with:
```C
skimage/feature/brief_pythran.cpp:1:10: fatal error: 'pythonic/core.hpp' file not found
#include <pythonic/core.hpp>
```

docs/project/changelog.md
ryanking13: I agree. And I think we better rename `Update` tag to `Upgrade` or `Version Up`?

rth:
Build should be OK now.

An issue is that since 2019 scikit-image has a [hard dependency on tifffile](https://github.com/scikit-image/scikit-image/blame/0f67fa74f9b7152aa109e9b8df9a63ca6b3af68f/skimage/io/collection.py#L24) which itself depends on imagecodecs which has around [25 codecs dependencies](https://github.com/conda-forge/imagecodecs-feedstock/blob/master/recipe/meta.yaml#L30). Currently patching to avoid tifffile import, which means that TIFF format won't work. 

Another issue is that Chrome tests for scikit-image timeout though I cannot reproduce this locally.

docs/project/changelog.md
rth: I agree we should indeed improve the changelog, I think it's simpler to just have a list, rather than an entry per package. Of course, major or long-awaited ones such as scipy still deserve their own changelog. For the rest I would rather spend time on some UI to list packages https://github.com/pyodide/pyodide/issues/977

rth:
As far as I can tell it times out because `scikit-image` is quite large (23MB) and takes a while to load in CI. Locally it doesn't happen because of more CPU resources. This is in part due to it bundling image examples for skimage/data/ for around 7.6MB which will also be incompressible. So for now I'm going to skip Chrome as it was done before. A medium-term solution should be to add some mechanism to unvendor data files as a separate package (as we are doing for tests), possibly with some regex in `meta.yaml`.

oeway:
@rth Great! thanks for taking care of this!

PR #2008, opened 2021-11-29
Add pyristent to main pyodide distribution

wlach:
pyrsistent is a dependency of (modern) jsonschema which is used by
altair, a popular python visualization library. Modern versions of it
use a C extension for speed, so let's make this part of the pyodide
distribution to ease installation.

As things stand, we can only use altair by resorting to a hack to
install an older version of jsonschema. Examples of this in the wild:

https://irydium.dev/repl?id=1a551ce8-f88c-11eb-866c-0a4e9a1089db
https://github.com/jtpio/jupyterlite/blob/main/examples/pyolite/altair.ipynb

### DIFF
--- a/packages/pyrsistent/meta.yaml
+++ b/packages/pyrsistent/meta.yaml
@@ -1,0 +1,14 @@
+package:
+  name: pyrsistent
+  version: 0.18.0
+source:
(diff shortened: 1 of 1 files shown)


rth (approved):
Thanks @wlach !

PR #2009, opened 2021-11-30
MAINT Store zipped build folder as circleci artifact

hoodmane:
The goal is to make it easier to run the circleci build locally. With this we should be able to download a zip archive of the full build directory then serve it to localhost.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -97,6 +97,11 @@ jobs:
             rm -rf cpython/{build,downloads}
             rm -rf emsdk/emsdk/binaryen
 
+      - run:
+          name: Zip build directory
+          command: |
+            tar cjf pyodide-build.tar.bz2  build
+
       - persist_to_workspace:
           root: .
           paths:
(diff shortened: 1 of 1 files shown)


rth:
At the point where we are, they will probably not mind if we store more things ). 

image

But we should probably try to optimize CircleCI more, as all of those numbers look quite high to me.

.circleci/config.yml
```diff
@@ -97,6 +97,11 @@ jobs:
             rm -rf cpython/{build,downloads}
             rm -rf emsdk/emsdk/binaryen
 
+      - run:
+          name: Zip build directory
+          command: |
+            tar cjf pyodide-build.tar.bz2  build
```
rth, line 103: If I remember correctly bz2 takes a long time to compress, maybe better to use gz instead? Let's see what CI says.

hoodmane: 50 seconds on the build-packages job.

hoodmane: Indeed only 30 seconds to gzip. gzipped archive is exactly the same size as the bzipeped one.

PR #2010, opened 2021-11-30
ENH Add a check for the browser shim of process when determining if the runtime is Node

bartbroere:
### Description

There's a browser shim for process, in which case we don't want to set `IN_NODE` to `true`.

From: https://www.npmjs.com/package/process
> It also exposes a "browser" member (i.e. process.browser) which is true in this implementation but undefined in node. This can be used in isomorphic code that adjusts it's behavior depending on which environment it's running in.

This fixes that packages like `starboard-notebook` have to patch `pyodide` at install time.

### DIFF
--- a/src/js/load-pyodide.js
+++ b/src/js/load-pyodide.js
@@ -4,7 +4,8 @@ const IN_NODE =
   typeof process !== "undefined" &&
   process.release &&
   process.release.name === "node" &&
-  typeof process.browser === "undefined";
+  typeof process.browser ===
+    "undefined"; /* This last condition checks if we run the browser shim of process */
 
 /** @typedef {import('./pyproxy.js').PyProxy} PyProxy */
 /** @private */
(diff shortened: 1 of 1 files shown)


rth:
Thanks! Please remove the checklist entries in your issue description if you don't plan to add them. I think this is probably fine without a test.

Generally LGTM.

bartbroere:
> Thanks! Please remove the checklist entries in your issue description if you don't plan to add them. I think this is probably fine without a test.
> 
> Generally LGTM.

Cool! Thanks for the reply; I removed the checklist.

bartbroere:
Could this be merged perhaps, despite the timed out and therefore failed build? It would help the starboard-notebook project quite a bit if it doesn't have to patch the pyodide package after installation. I'm hoping this fix gets in the next pyodide release.

rth (approved):
Sure, that works, thanks! The CI failure is due to https://github.com/pyodide/pyodide/issues/2013

Issue #1967, opened 2021-11-16
Add micropip.list


rth:
It would be nice to add `micropip.list` function to list all installed packages and their versions, by analogy with the `pip list` command.

The output could be a list like object of dictionaries (or data class objects) with the following fields,
```py
    name: str
    version : str
    path : Path
```
and a `__repr__`  method to display this nicely (except for maybe the path). Actually, tt could be either,
 1. a custom list like object with a `__repr__` containing standard dict objects
 2. a standard list with dataclasses that each has a custom `__repr__` method

or some mixture of the two. Option 2 would likely be easiest to use/implement IMO.

ryanking13:
Should the output of this API include js packages that are listed in `packages.json`? I think so if we are going to merge `micropip` and `pyodide.loadPackage`

rth:
Yes, I was thinking we could support both, and indicate the source in a separate column.
```
# Name        Version                   Source
package_1     0.2.2                     pypi
package_2     1.3.1                     pyodide
package_3     0.1.4                     <some-custom-url.whl>
```
This is what `conda list` is doing for instance by listing both pip and conda install packages.

Not 100% sure yet what would be the best output type to both have this repr and a usable container object with the results though.

rth:
> I think so if we are going to merge micropip and pyodide.loadPackage

Answered in https://github.com/pyodide/pyodide/issues/1470#issuecomment-970079464

Issue #1470, opened 2021-04-14
Plans for loadPackage / micropip.install


hoodmane:
I think we should update `loadPackage` to require that if it loads a file from `http://some_website/pyodide_packages/blah.js` or `./blah.js` or whatever, there must be a `packages.json` in the same directory. Then we can load the version info (and dependencies) from the `packages.json`. This will allow us to fix the remaining package loading bugs. (If you load a package with `loadPackage`, then `micropip` won't know about it and may install a different version over it in the future, `micropip` isn't aware of the `loadPackage` dependency resolution, `loadPackage` can't do dependency resolution unless the package is loaded from `indexURL`.)

Then at that point, I think we can add extra logic into `loadPackage` so that it can load and call into micropip if need be. Then we can support installing wheels and packages from pypi using `loadPackage` and remove references to micropip from the docs, encouraging people to use `loadPackage` instead. For use from Python we can add `pyodide.load_package` which as just an alias to `loadPackage`.

I think that removing the mental overhead of figuring out the difference between `loadPackage` and `micropip` will help beginners out a lot.

_Originally posted by @hoodmane in https://github.com/pyodide/pyodide/issues/1469#issuecomment-819286731_

oeway:
I find it confusing that we cannot do `micropip.install("numpy")` in the console: https://pyodide-cdn2.iodide.io/v0.17.0a2/full/console.html, is that related to the plan here?

hoodmane:
This bug was also fixed in #1457 / #1463.

But my goal is to get rid of all confusions about the differences between `micropip` vs `loadPackage` by merging them completely.

teonbrooks:
I like this proposal. I have definitely run into the problem of loading different versions of the same package, one listed in pyodide, the other being a pure python that can be loaded with micropip. it would enforce that pyodide listed package that version precedence. It would also clarify the process too: load package(s) first, import them after

rth:
> But my goal is to get rid of all confusions about the differences between micropip vs loadPackage by merging them completely.

I agree moving in that direction would make sense, and making it more user friendly would be good. I'm not sure however that we can completely smooth the difference between these two sources.

Major differences are,
 - In loadPackage, we can do full dependency resolution for available packages. With micropip we can't. Metadata from the PyPi JSON API is not fully reliable and sometimes you have download the package to discover its dependencies.
[...]

PR #2012, opened 2021-12-01
ENH Add micropip.list()

ryanking13:
### Summary

Closes: #1967

Adds a new API `micropip.list()` which returns a list of installed packages.

### Detail

- `PackageMetadata`: a dataclass object that contains (name, version, source) of a package
- `PackageList`: a dict-like object that maps package name to PackageMetadata
  - has custom `__repr__()` to print list of packages in tabularized format
- `micropip.list()` returns `PackageList` instance which contains list of installed packages
- open to better names and better container design 😀

```
>>> import micropip
>>> await micropip.install("black")
>>> await micropip.install("https://files.pythonhosted.org/packages/89/06/2c2d3034b4d6bf22f2a4ae546d16925898658a33b4400cfb7e2c1e2871a3/pytz-2020.5-py2.py3-none-any.whl")
>>> pkgs = micropip.list()
>>> print(pkgs)
| Name              | Version  | Source                                                                                                                                      |
| ----------------- | -------- | ------------------------------------------------------------------------------------------------------------------------------------------- |
| regex             | 2021.7.6 | pyodide                                                                                                                                     |
| click             | 8.0.3    | pypi                                                                                                                                        |
| typing-extensions | 4.0.1    | pypi                                                                                                                                        |
| platformdirs      | 2.4.0    | pypi                                                                                                                                        |
| pathspec          | 0.9.0    | pypi                                                                                                                                        |
| tomli             | 1.2.2    | pypi                                                                                                                                        |
| mypy-extensions   | 0.4.3    | pypi                                                                                                                                        |
| black             | 21.11b1  | pypi                                                                                                                                        |
| pytz              | 2020.5   | https://files.pythonhosted.org/packages/89/06/2c2d3034b4d6bf22f2a4ae546d16925898658a33b4400cfb7e2c1e2871a3/pytz-2020.5-py2.py3-none-any.whl |
>>> "regex" in pkgs
True
>>> "pytz" in pkgs
True
>>> "numpy" in pkgs
False
```

### Checklists

- [ ] Add / update tests

### DIFF
--- a/packages/micropip/src/micropip/micropip.py
+++ b/packages/micropip/src/micropip/micropip.py
@@ -1,4 +1,5 @@
 import asyncio
+import copy
 import hashlib
 import importlib
 import io
@@ -13,6 +14,7 @@ from typing import Dict, Any, Union, List, Tuple, Optional
 from zipfile import ZipFile
 
 from .externals.pip._internal.utils.wheel import pkg_resources_distribution_for_wheel
+from .package import PackageList, PackageMetadata
 
 from pyodide import IN_BROWSER, to_js
 
@@ -140,10 +142,13 @@ async def _install_wheel(name, fileinfo):
 
 class _PackageManager:
     def __init__(self):
-        self.installed_packages = {}
+        self.installed_packages = PackageList()
 
     async def gather_requirements(
-        self, requirements: Union[str, List[str]], ctx=None, keep_going: bool = False
+        self,
+        requirements: Union[str, List[str]],
+        ctx=None,
+        keep_going: bool = False,
     ):
         ctx = ctx or default_environment()
         ctx.setdefault("extra", None)
@@ -153,7 +158,7 @@ class _PackageManager:
         transaction: Dict[str, Any] = {
             "wheels": [],
             "pyodide_packages": [],
-            "locked": dict(self.installed_packages),
+            "locked": copy.deepcopy(self.installed_packages),
             "failed": [],
             "keep_going": keep_going,
         }
@@ -185,11 +190,11 @@ class _PackageManager:
         if len(pyodide_packages):
             # Note: branch never happens in out-of-browser testing because in
             # that case BUILTIN_PACKAGES is empty.
-            self.installed_packages.update(pyodide_packages)
+            self.installed_packages.update({pkg.name: pkg for pkg in pyodide_packages})
             wheel_promises.append(
                 asyncio.ensure_future(
                     pyodide_js.loadPackage(
-                        to_js([name for [name, _] in pyodide_packages])
+                        to_js([name for [name, _, _] in pyodide_packages])
                     )
                 )
             )
@@ -197,7 +202,13 @@ class _PackageManager:
         # Now install PyPI packages
         for name, wheel, ver in transaction["wheels"]:
             wheel_promises.append(_install_wheel(name, wheel))
-            self.installed_packages[name] = ver
+
+            # detect whether the wheel metadata is from PyPI or from custom location
+            # wheel metadata from PyPI has SHA256 checksum digest.
+            wheel_source = "pypi" if wheel["digests"] is not None else wheel["url"]
+            self.installed_packages[name] = PackageMetadata(
+                name=name, version=str(ver), source=wheel_source
+            )
         await gather(*wheel_promises)
 
     async def add_requirement(
@@ -230,7 +241,9 @@ class _PackageManager:
             and BUILTIN_PACKAGES[req.name]["version"] in req.specifier
         ):
             version = BUILTIN_PACKAGES[req.name]["version"]
-            transaction["pyodide_packages"].append((req.name, version))
+            transaction["pyodide_packages"].append(
+                PackageMetadata(name=req.name, version=version, source="pyodide")
+            )
             return
 
         if req.marker:
@@ -241,7 +254,7 @@ class _PackageManager:
 
         # Is some version of this package is already installed?
         if req.name in transaction["locked"]:
-            ver = transaction["locked"][req.name]
+            ver = transaction["locked"][req.name].version
             if ver in req.specifier:
                 # installed version matches, nothing to do
                 return
@@ -264,7 +277,7 @@ class _PackageManager:
             await self.add_wheel(req.name, wheel, ver, req.extras, ctx, transaction)
 
     async def add_wheel(self, name, wheel, version, extras, ctx, transaction):
-        transaction["locked"][name] = version
+        transaction["locked"][name] = PackageMetadata(name=name, version=version)
         wheel_bytes = await fetch_bytes(wheel["url"])
         wheel["wheel_bytes"] = wheel_bytes
 
@@ -306,6 +319,9 @@ class _PackageManager:
                     return fileinfo, ver
 
         return None, None
+
+    def list(self) -> PackageList:
+        return copy.deepcopy(self.installed_packages)
 
 
 # Make PACKAGE_MANAGER singleton
@@ -365,7 +381,30 @@ def install(requirements: Union[str, List[str]], keep_going: bool = False):
     )
 
 
-__all__ = ["install"]
+def list():
+    """Get the list of installed packages through micropip.
+
+    Returns
+    -------
+    ``PackageMetadataList``
+        A list of installed packages.
+
+        >>> package_list = micropip.list()
+        >>> print(package_list)
+        | Name              | Version  | Source  |
+        | ----------------- | -------- | ------- |
+        | regex             | 2021.7.6 | pyodide |
+        | tomli             | 1.2.2    | pypi    |
+        | mypy-extensions   | 0.4.3    | pypi    |
+        ...
+        >>> "regex" in package_list
+        True
+    """
+    return PACKAGE_MANAGER.list()
+
+
+# prevent sphinx autodoc from generating duplicated docs
+__all__: List[str] = []
 
 
 if __name__ == "__main__":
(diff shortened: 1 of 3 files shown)


packages/micropip/src/micropip/micropip.py
```diff
@@ -185,19 +190,25 @@ async def install(
...
             )
 
         # Now install PyPI packages
         for name, wheel, ver in transaction["wheels"]:
             wheel_promises.append(_install_wheel(name, wheel))
-            self.installed_packages[name] = ver
+
+            # detect whether the wheel metadata is from PyPI or from custom location
+            # wheel metadata from PyPI has SHA256 checksum digest.
+            wheel_source = "pypi" if wheel["digests"] is not None else wheel["url"]
```
ryanking13, line 208: This (checking digest) is a hacky way to detect whether the wheel is from PyPI or from a custom URL.
Another way can be adding a key that explicitly indicates the wheel is from a custom URL.

packages/micropip/src/micropip/package.py
```diff
@@ -0,0 +1,71 @@
...
+        return self.packages[key]
+
+    @property
+    def package_names(self):
+        return [pkg.name for pkg in self]
+
+    def update(self, pkg: Dict[str, PackageMetadata]):
+        self.packages.update(pkg)
+
+    def tabularize(self):
```
rth, line 68: This can probably be the `__repr__` method directly or a `_tabularize` method to keep it private.

packages/micropip/src/micropip/package.py
```diff
@@ -0,0 +1,71 @@
...
+
+    def __iter__(self):
+        return iter(astuple(self))
+
+    @staticmethod
+    def keys():
+        return [key.capitalize() for key in PackageMetadata.__dataclass_fields__.keys()]
+
+
+class PackageList(Collection):
```
rth, line 39: Since the expected bahavior is that of a custom dict, I wonder if it wouldn't be simpler to inherit from 
`collections.UserDict` which will make it unnecessary to implement most of the following `__` methods 

Also because it's returned by the list API, we might need to add it to sphinx documentation. Which means it would need a short docstring.

packages/micropip/src/micropip/micropip.py
```diff
@@ -365,7 +381,30 @@ def install(requirements: Union[str, List[str]], keep_going: bool = False):
...
+    ``PackageMetadataList``
+        A list of installed packages.
+
+        >>> package_list = micropip.list()
+        >>> print(package_list)
+        | Name              | Version  | Source  |
+        | ----------------- | -------- | ------- |
+        | regex             | 2021.7.6 | pyodide |
+        | tomli             | 1.2.2    | pypi    |
+        | mypy-extensions   | 0.4.3    | pypi    |
```
rth, line 398: Looks nice. How about,
```suggestion
        Name              | Version  | Source
        ----------------- | -------- | -------
        regex             | 2021.7.6 | pyodide
        tomli             | 1.2.2    | pypi   
        mypy-extensions   | 0.4.3    | pypi 
```
as something a bit lighter? Unless the goal was to make markdown compatible tables?

packages/micropip/src/micropip/micropip.py
```diff
@@ -185,19 +190,25 @@ async def install(
...
 
         # Now install PyPI packages
         for name, wheel, ver in transaction["wheels"]:
             wheel_promises.append(_install_wheel(name, wheel))
-            self.installed_packages[name] = ver
+
+            # detect whether the wheel metadata is from PyPI or from custom location
+            # wheel metadata from PyPI has SHA256 checksum digest.
+            wheel_source = "pypi" if wheel["digests"] is not None else wheel["url"]
+            self.installed_packages[name] = PackageMetadata(
```
rth, line 209: Not introduced in this PR; but it could possible that the above _install_wheel would fail, and yet we would still add this to the list of installed packages? Maybe better to do it via some callback to `_install_wheel ` after the installation succeded there ?

rth (commented):
Thanks @ryanking13 , I was hoping you would be interesting in doing this :)

PackageMetadata / PackageList could use some tests, or at least a few doctests.

packages/micropip/src/micropip/micropip.py
```diff
@@ -365,7 +381,30 @@ def install(requirements: Union[str, List[str]], keep_going: bool = False):
     )
 
 
-__all__ = ["install"]
+def list():
```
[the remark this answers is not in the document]
ryanking13: Yeah, I was also thinking that using the name `list` could be problematic.

packages/micropip/src/micropip/micropip.py
```diff
@@ -365,7 +381,30 @@ def install(requirements: Union[str, List[str]], keep_going: bool = False):
...
+        | mypy-extensions   | 0.4.3    | pypi    |
+        ...
+        >>> "regex" in package_list
+        True
+    """
+    return PACKAGE_MANAGER.list()
+
+
+# prevent sphinx autodoc from generating duplicated docs
+__all__: List[str] = []
```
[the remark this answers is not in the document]
ryanking13: Actually, I find out that just making micropip installation private (`_micropip.py`) that you suggested would fix this duplication docs problem :)

ryanking13:
Applied suggestions:

- Make micropip implementation private `_micropip.py` (This also prevent sphinx from generating duplicated docs)
- Rename `list()` to `_list()` and use aliasing when importing
- Lighter table formatting
- Use `collections.UserDict`, and rename `PackageList` to `PackageDict`
  - Now `PackageDict` is a really simple dict-like object with custom `__repr__` method.
- Update installed_packages after installation succeeds.
- Add tests for micropip.list()

packages/micropip/src/micropip/_micropip.py
```diff
@@ -307,6 +336,9 @@ def find_wheel(
 
         return None, None
 
+    def list(self) -> PackageDict:
+        return copy.deepcopy(self.installed_packages)
```
rth, line 340: Not sure why we need this. We could call this code directly in _list no?

packages/micropip/src/micropip/_micropip.py
```diff
@@ -365,7 +397,25 @@ def install(requirements: Union[str, List[str]], keep_going: bool = False):
     )
 
 
-__all__ = ["install"]
+def _list():
+    """Get the dictionary of installed packages through micropip.
+
+    Returns
+    -------
+    ``PackageDict``
```
rth, line 405: Maybe to link to the correct section int the docs
```suggestion
    packages : {class}`~micropip.package.PackageDict``
```

packages/micropip/src/micropip/_micropip.py
```diff
@@ -365,7 +397,25 @@ def install(requirements: Union[str, List[str]], keep_going: bool = False):
     )
 
 
-__all__ = ["install"]
+def _list():
+    """Get the dictionary of installed packages through micropip.
```
rth, line 401:
```suggestion
    """Get the dictionary of installed packages.
```
If we also support Pyodide packages here.

rth (commented):
Thanks! A few more minor comments below.

A more major point, is that if we want `micropip` to be the main API for users #1470, micropip.list should also list packages loaded via `loadPackage`. In particular, for a new page load in console it would currently show that no packages are installed, as far as I understand, while the following were loaded,
```py
>>> micropip.micropip.loadedPackages.to_py()
{'distutils': 'default channel', 'micropip': 'default channel', 'pyparsing': 'default channel', 'packaging': 'default channel'}
```
So I think we can fix this by merging both sources with something like,
```py
def _list():
    from importlib.metadata import version as get_version
    packages = copy.deepcopy(self.installed_packages)
    for name, _source in micropip.micropip.loadedPackages.to_py().items():
         if name in packages:
             continue    

          version = get_version(name)
          packages[name] = PackageMetadata(name=name, version=version, source='pyodide')
    return packages
```

Though we might want to add a check in `test_import` [here](https://github.com/pyodide/pyodide/blob/fd5f01f685fff75b61f95f6019a2d224441fb5cd/packages/test_packages_common.py#L114) that getting the version with importlib doesn't fail. 

WDYT?

rth:
Also, let's add a mention to https://pyodide.org/en/stable/usage/api/js-api.html#pyodide.loadedPackages docs that this only concerns packages installed with loadPackage and that for a more general solution it is recommended to use 
```
{func}`micropip.list`
```

packages/micropip/src/micropip/_micropip.py
```diff
@@ -28,7 +32,7 @@
 
     WHEEL_BASE = Path(getsitepackages()[0])
 else:
-    WHEEL_BASE = Path(".") / "wheels"
+    WHEEL_BASE = Path(tempfile.mkdtemp())
```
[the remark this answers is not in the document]
ryanking13: For now, testing micropip (== micropip is imported in a new process (outside of Pyodide)) generates `wheels` folder in the working directory.

How about renaming it to `.wheels` and adding it to `.gitignore`?

ryanking13:
> A more major point, is that if we want `micropip` to be the main API for users #1470, micropip.list should also list packages loaded via `loadPackage`.

Totally agreed, I'll apply that, thanks!

ryanking13:
> Also, let's add a mention to https://pyodide.org/en/stable/usage/api/js-api.html#pyodide.loadedPackages docs that this only concerns packages installed with loadPackage and that for a more general solution it is recommended to use
> 
> ```
> {func}`micropip.list`
> ```

Actually, `pyodide.loadedPackages` is updated when `micropip.install` is called. So there are some duplications now...

https://github.com/pyodide/pyodide/blob/337847ac91b6e3b50b3882d653a1aedb1d328dc8/packages/micropip/src/micropip/_micropip.py#L137-L142

rth (approved):
Thanks LGTM.

ryanking13:
Thanks! I'll look into the `pyodide.loadedPackages` later. Maybe we could make it private after #1470.

PR #2016, opened 2021-12-03
fpcast2

hoodmane:
A second, much simpler approach to handling the bad fpcasts.
There are only a few problematic call sites, so just patch call trampolines into the interpreter at these locations.


ryanking13:
I'm very looking forward to seeing this PR merged. Thank you for your efforts! 👍

hoodmane:
There is still the problem of it breaking Chrome CI to deal with...

hoodmane:
Okay, I think this is in good shape now:
1. All tests pass on node and firefox
2. All chrome tests time out. I still have no clue why. I tested locally on Chrome 90.0.4430.24, the version we use in CI. It works fine.

hoodmane:
Closed in favor of #2019.

Issue #2013, opened 2021-12-01
CI build-packages timeouts without ccache


rth:
It looks like CircleCI introduced a 1 hour timeout, or maybe it's due to us switching to a `Large` CI instance for the `build-packages`  job. 

In any case, this job would now timeout with an empty ccache cache, since it takes longer, including for new contributors e.g. https://github.com/pyodide/pyodide/pull/2010 .

I think the solution is to reduce the number of packages we build for each commit (which would be good for CI times in general), only keeping the most difficult to build &  popular ones . Tests should work well in that case after https://github.com/pyodide/pyodide/pull/1981

Then for,
 - tags
 -  if the commit message contains some pattern, for instance `[build all]`
 
build all packages.

ryanking13:
For a short term solution, maybe we could rely more on Github Actions that looks like having less limitations on public repositories.

rth:
The issue is that would mean re-designing all our deploy process to GH Actions which is a non-negligible amount of work. Another issue with GH actions is that we still can't restart individual jobs if there is a non-deterministic failure.

ryanking13:
> The issue is that would mean re-designing all our deploy process to GH Actions which is a non-negligible amount of work.

Yes. Rewriting a CI process would not be a easy process. I may said it too lightly.

PR #2017, opened 2021-12-05
MAINT Split CI build-packages step in two

hoodmane:
Split build-packages into two stages. Hopefully this will fix the CI timeouts when the cache is not around.

Closes https://github.com/pyodide/pyodide/issues/2013

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -105,6 +105,56 @@ jobs:
       - store_artifacts:
           path: /root/repo/build/
 
+  build-packages-no-numpy-dependents:
+    <<: *defaults
+    resource_class: large
+
+    steps:
+      - checkout
+
+      - attach_workspace:
+          at: .
+
+      - restore_cache:
+          keys:
+            - -pkg-{{ checksum "Makefile.envs" }}-v20210911-
+
+      - run:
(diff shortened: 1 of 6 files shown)


rth (commented):
Thanks ! This would resolve https://github.com/pyodide/pyodide/issues/2013 . The ideal case would have been to run those two jobs in parallel then merge the results. But those looks not very straightforward so we can leave it for later.

There are a few test failures in test-python otherwise LGTM.

hoodmane:
It seems to be building a bunch of the packages twice, just the second time the ccache is already warmed up. Why doesn't it think the packages are already built the second time?

hoodmane:
> The ideal case would have been to run those two jobs in parallel then merge the results. But that looks not very straightforward so we can leave it for later.

I don't think it would make any difference to do that. The first stage is 100% dependent on time to build numpy (it finishes last even if it starts first), and all of the packages in the second stage are dependencies of numpy.

hoodmane:
It would be pretty natural to split it again after building scipy, so that it would be:
1. numpy + packages that don't depend on numpy
2. scipy + packages that don't depend on scipy
3. scipy dependencies

PR #2018, opened 2021-12-05
ENH Add install archive api

hoodmane:
### Description
Adds an API that accepts a buffer and a string format one of 'bztar', 'gztar', 'tar', and 'zip', and an optional target directory. It treats the buffer as an archive of the indicated type and unpacks it into either the given target directory or the working directory.

See discussion in #1940.

Feedback is welcome, particularly on naming things. @alexmojaki @grimmer0125

### Checklists

- [x] Add / update tests

### DIFF
--- a/src/js/api.js
+++ b/src/js/api.js
@@ -318,7 +318,6 @@ export function pyimport(mod_name) {
   return Module.importlib.import_module(mod_name);
 }
 
-let _util_module;
 /**
  * Unpack an archive into a target directory.
  *
@@ -328,10 +327,10 @@ let _util_module;
  * @param {string=} extract_dir The directory to unpack the archive into. Defaults to the working directory.
  */
 export function unpackArchive(buffer, format, extract_dir) {
-  if (!_util_module) {
-    _util_module = Module.pyimport("pyodide._util");
+  if (!Module._util_module) {
+    Module._util_module = pyimport("pyodide._util");
   }
-  _util_module.unpack_buffer_archive(buffer, format, extract_dir);
+  Module._util_module.unpack_buffer_archive(buffer, format, extract_dir);
 }
 
 /**
--- a/src/js/api.js
+++ b/src/js/api.js
@@ -323,7 +323,9 @@ export function pyimport(mod_name) {
  *
  * @param {ArrayBuffer} buffer The archive as an ArrayBuffer (it's also fine to pass a TypedArray).
  * @param {string} format The format of the archive. Should be one of the formats recognized by `shutil.unpack_archive`.
- *   By default the options are 'bztar', 'gztar', 'tar', and 'zip'.
+ * By default the options are 'bztar', 'gztar', 'tar', 'zip', and 'wheel'. Several synonyms are accepted for each format, e.g.,
+ * for 'gztar' any of '.gztar', '.tar.gz', '.tgz', 'tar.gz' or 'tgz' are considered to be synonyms.
+ *
  * @param {string=} extract_dir The directory to unpack the archive into. Defaults to the working directory.
  */
 export function unpackArchive(buffer, format, extract_dir) {
--- a/src/js/api.js
+++ b/src/js/api.js
@@ -84,7 +84,7 @@ export let version = ""; // actually defined in loadPyodide (see pyodide.js)
  * is returned.
  *
  * @param {string} code Python code to evaluate
- * @param {PyProxy} globals An optional Python dictionary to use as the globals.
+ * @param {PyProxy=} globals An optional Python dictionary to use as the globals.
  *        Defaults to :any:`pyodide.globals`. Uses the Python API
  *        :any:`pyodide.eval_code` to evaluate the code.
  * @returns {Py2JsResult} The result of the Python code translated to JavaScript. See the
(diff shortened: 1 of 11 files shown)


alexmojaki:
I think there must be some files you forgot to push?

hoodmane:
@alexmojaki oh yeah. I always forget to add new files...

alexmojaki:
I really liked @grimmer0125's last suggestion to make this part of `loadPyodide`. That would make this really easy to use while giving almost complete parallelism at the same time. Here's a suggested API:

```js
loadPyodide({indexURL, extraFiles: [{url: "my_python.tar", format: "tar", destination: "/path/to/extract_dir"}]})
```

- `extraFiles` is an array in case people want to download multiple archives. They should all download in parallel.
- `url` would be required, but maybe in the future we could allow other arguments in its place, such as a `Promise<ArrayBuffer>`.
- `format` would be optional and would be the extension of `url` if omitted.
- `destination` would be optional of course.

`unpackArchive` can still be included as well for power users with very specific requirements, but it adds very little overall and still requires a decent amount of work from the user, especially for parallel downloading. The convenient API should satisfy almost everyone's needs, it certainly satisfies all of mine.

One way I can theoretically imagine this being used is passing several URLs that each directly download a wheel from PyPI. You could install multiple dependencies in parallel to each other and to pyodide itself without even needing micropip.

hoodmane:
> One way I can theoretically imagine this being used is passing several URLs that each directly download a wheel from PyPI. You could install multiple dependencies in parallel to each other and to pyodide itself without even needing micropip.

Maybe we could generate the necessary incantation with `micropip.freeze`...

src/py/pyodide/_util.py
```diff
@@ -0,0 +1,19 @@
+from tempfile import NamedTemporaryFile
+import shutil
+
+
+def get_format(format):
```
alexmojaki, line 5: I suggest checking for `.whl` and variations and returning `'zip'` for that.

hoodmane: Let's just add them to `shutil` formats on initialization.

hoodmane: > `.whl` and variations

Is there a place I can find the list of standard variations?

alexmojaki: There is `register_unpack_format` but honestly it looks easier to just do it manually in your function.

No idea about standards, I was just thinking about `whl` and `wheel` with an optional `.` prefix.

hoodmane: I can just take the code from here:
https://bugs.python.org/issue39846

ryanking13:
> I really liked @grimmer0125's last suggestion to make this part of `loadPyodide`. That would make this really easy to use while giving almost complete parallelism at the same time. Here's a suggested API:
> 
> ```js
> loadPyodide({indexURL, extraFiles: [{url: "my_python.tar", format: "tar", destination: "/path/to/extract_dir"}]})
> ```
> 
> * `extraFiles` is an array in case people want to download multiple archives. They should all download in parallel.
> * `url` would be required, but maybe in the future we could allow other arguments in its place, such as a `Promise<ArrayBuffer>`.
> * `format` would be optional and would be the extension of `url` if omitted.
> * `destination` would be optional of course.

Though I don't like `pyodide.loadPyodide` getting too much funtionalities, this would be really convenient for users.

hoodmane:
@ryanking @rth Could one of you review this?

src/js/api.js
```diff
@@ -318,6 +318,23 @@ export function pyimport(mod_name) {
...
+ * Unpack an archive into a target directory.
+ *
+ * @param {ArrayBuffer} buffer The archive as an ArrayBuffer (it's also fine to pass a TypedArray).
+ * @param {string} format The format of the archive. Should be one of the formats recognized by `shutil.unpack_archive`.
+ * By default the options are 'bztar', 'gztar', 'tar', 'zip', and 'wheel'. Several synonyms are accepted for each format, e.g.,
+ * for 'gztar' any of '.gztar', '.tar.gz', '.tgz', 'tar.gz' or 'tgz' are considered to be synonyms.
+ *
+ * @param {string=} extract_dir The directory to unpack the archive into. Defaults to the working directory.
+ */
+export function unpackArchive(buffer, format, extract_dir) {
```
ryanking13, line 331: How about matching parameter order of this with `FetchResponse.unpack_archive`?

https://github.com/pyodide/pyodide/blob/d6035abb888c3388106c08402627e637591f440e/src/py/pyodide/http.py#L185

And also maybe we could make `FetchResponse.unpack_archive` to use `unpack_buffer_archive`? Because there are some duplications in the unpacking code now.

ryanking13 (commented):
Thanks for your effort! I have some minor comments.

BTW, I think you referenced a wrong username (ryanking)?

hoodmane:
Thanks for the review!

> BTW, I think you referenced a wrong username (ryanking)?

Oops, sorry.

src/js/api.js
hoodmane: I added `*` into the argument list of `unpack_archive` and `unpack_buffer_archive` to alleviate argument order issues. I unified `unpack_archive` and `unpack_buffer_archive` as you suggested.

PR #2022, opened 2021-12-05
FIX Correctly handle cancelled handles in webloop

### LINKED ISSUE #2011: exception in asyncio.wait_for()
## 🐛 Bug

Waiting for an awaitable using `asyncio.wait_for()` prints exceptions to the browser's console:

```
pyodide.asm.js:14 Exception in callback None()
pyodide.asm.js:14 handle: <Handle cancelled>
```

This happens whether the awaitable finishes before the timeout or not. `asyncio.wait_for()` appears to still function as intended, but this makes applications that use `asyncio.wait_for()` frequently very spammy.

### To Reproduce

In [Pyodide console application](https://pyodide.org/en/stable/console.html), enter the following:

```
import asyncio
await asyncio.wait_for(asyncio.sleep(1), 2)
```

### Expected behavior

No exceptions are raised and no output is printed to the console.

### Environment

- Pyodide Version: 0.18.1 from https://cdn.jsdelivr.net/pyodide/v0.18.1/full/
- Browser version: Chrome 96.0.4664.55

hoodmane:
This resolves #2011.

### Checklists

- [x] Add / update tests


src/py/pyodide/webloop.py
```diff
@@ -143,7 +143,13 @@ def call_later(
         if delay < 0:
             raise ValueError("Can't schedule in the past")
         h = asyncio.Handle(callback, args, self, context=context)
-        setTimeout(create_once_callable(h._run), delay * 1000)
+
+        def run_handle():
+            if h._cancelled:
```
rth, line 148: Isn't the public metod `cancelled()` instead? https://docs.python.org/3/library/asyncio-eventloop.html#asyncio.Handle.cancelled

rth (approved):
Makes sense, thanks! LGTM, except that test_webloop_exception_handler fails.

src/py/pyodide/webloop.py
hoodmane: I guess so. I was copying from `base_events.py` which uses `_cancelled`, but this seems better.

PR #2023, opened 2021-12-09
MAINT Load main pyodide package as a tar file

hoodmane:
Bundle Pyodide into a separate tar file and unpack it into the `site-packages` directory on startup. We can eventually allow people to provide their own custom file name as an argument to `loadPyodide`. Since we use `shutil.unpack_archive` the file could be a zip, tar, gztar, etc. In particular, this should eventually enable contributions to the Pyodide Python code by windows users (they can zip the files) without building Python or any of the C files.

### DIFF
--- a/src/core/main.c
+++ b/src/core/main.c
@@ -122,7 +122,8 @@ main(int argc, char** argv)
 }
 
 int
-pyodide_init(void){
+pyodide_init(void)
+{
   PyObject* _pyodide = NULL;
   PyObject* core_module = NULL;
   JsRef _pyodide_proxy = NULL;
--- a/src/core/main.c
+++ b/src/core/main.c
@@ -100,6 +100,13 @@ get_python_stack_depth()
 int
 main(int argc, char** argv)
 {
+  EM_ASM({
+    // For some reason emscripten doesn't make UTF8ToString available on Module
+    // by default...
+    Module.UTF8ToString = UTF8ToString;
+    Module.wasmTable = wasmTable;
+  });
+
   // This exits and prints a message to stderr on failure,
   // no status code to check.
   initialize_python();
@@ -110,7 +117,12 @@ main(int argc, char** argv)
   if (sizeof(JsRef) != sizeof(int)) {
     FATAL_ERROR("JsRef doesn't have the same size as int.");
   }
+  emscripten_exit_with_live_runtime();
+  return 0;
+}
 
+int
+pyodide_init(void){
   PyObject* _pyodide = NULL;
   PyObject* core_module = NULL;
   JsRef _pyodide_proxy = NULL;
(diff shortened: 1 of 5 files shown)


hoodmane:
I can't figure out why lint is failing.

src/core/main.c
```diff
@@ -110,7 +117,12 @@ main(int argc, char** argv)
   if (sizeof(JsRef) != sizeof(int)) {
     FATAL_ERROR("JsRef doesn't have the same size as int.");
   }
+  emscripten_exit_with_live_runtime();
+  return 0;
+}
 
+int
+pyodide_init(void){
```
ryanking13, line 125: Looks like `clang-format` is complaining about this line?

```sh
clang-format-6.0 src/core/main.c | diff src/core/main.c -
```

hoodmane: Thanks!

rth:
I think we should agree on the container format for packages and go with it. For instance, if we go with wheels then it would make sense to also make this a zipped .whl file even if it's not strictly following the wheel standard, not sure. Also on Windows by default, there is no application to open .tar so it's probably not ideal in that sense. 

In any case, we need to whitelist package extensions with JSDelivr (I have to ask them by email), that's why I would rather only do it once when we are sure about the format we want.

hoodmane:
> Also on Windows by default, there is no application to open .tar so it's probably not ideal in that sense.

Yeah that's pretty significant.

How about .zip and .whl? They are the same file format, but I'm not sure it's a good idea to just stick the .whl extension on random zip files.

rth:
> I'm not sure it's a good idea to just stick the .whl extension on random zip files.

You are probably right.

> How about .zip and .whl?

But then maybe .tar and .whl is not so bad. It's not insurmountable to open it on Windows https://superuser.com/a/215630
And conceptually this would avoid double compression. If we store it as .tar, it would effectively be distributed as .tar.br or .tar.gz by the CDN due to browser compression. But then locally it would be stored as .tar and there would be no need to de-compress from the cache when reloading the page.

Which seems better than than brotli compressed zip files.  But then we would still have to do it for wheels anyway.

Though, the same could be applied to zip with no-compression. It not so much a concern for this particular case, but I wonder what are the best practices to avoid this double compression issue.

hoodmane:
> we need to whitelist package extensions

Could we go ahead and whitelist all three .tar, .whl, and .zip? Because we are unpacking with shutil, all it takes to change the format is fetching a different archive. We can use tar by default but windows users can zip things instead. This gives us plenty of flexibility if it eventually turns out that some particular choice is preferable.

src/js/load-pyodide.js
```diff
@@ -43,6 +43,17 @@ export async function initializePackageIndex(indexURL) {
   }
 }
 
+export async function fetchTarFile(indexURL, path) {
```
rth, line 46: There is nothing specific to .tar here? Maybe `_fetchBinaryFile` ? I would still like to indicate somehow that it is not expected to be used by users...

rth (approved):
Could you please add a changelog? Otherwise LGTM.

Putting .tar, .zip and .whl files should now work with the CDN thanks to https://github.com/MartinKolarik

hoodmane:
Thanks @rth!

PR #2027, opened 2021-12-10
Use wheels instead of file_packager

### LINKED ISSUE #655: Binary wheels for pyodide?
There was earlier discussion about this in https://github.com/iodide-project/pyodide/issues/38 and https://github.com/iodide-project/pyodide/issues/344 but I think it could be useful to have a separate issue to discuss what binary wheels for pyodide could look like.

Python wheels are basically just zip files with the package contents, see [PEP 427](https://www.python.org/dev/peps/pep-0427/). We are currently already supporting pure python wheels.

We could also consider trying to build wheels for packages that have C extensions. I'm not exactly sure what would be necessary on the technical side to make it work, and run the wasm binary through the compiler when the modules are loaded. That would require more investigation, and it could be a blocker, but I imagine it could be made to work. 

For instance a numpy wheel could be,
```
 numpy-1.4.0-py37-pyodide015-wasm.whl
```
where the latter two parts are the binary ABI and the platform.

**Advantages**
 - wheels have already a clear spec, deal with ABI compatibility and are natural for python users.
 - I think they could be distributed through PyPi (or any other 
warehouse / PyPi index). If the platform tag is not supported now, I wouldn't likely be too difficult to support it.
 - Because it's just a zip file, it should be relatively easy to repackage a set of wheels into one or strip unused parts for a given project (without dealing with emscripten side of things at all).
 - They might make package building a bit more sustainable, because we would have to deal less with it in pyodide, and users can do it and distribute results independently.
 - One file per package to download instead of two (`.js` + `.data`).
 
**Disadvantages**
 - it puts stronger constraints on ABI compatibility, meaning that we cannot make releases every other month that breaks compatibility (or in particular update emscripten). That's certainly not ideal, and we considered it a blocker in the past, https://github.com/iodide-project/pyodide/issues/344 but we do have to start to deal with it at some point in any case probably.
 - in particular, it likely means we cannot make releases too often. For the situation with PyPy wheels https://github.com/antocuni/pypy-wheels is a relevant reference.
 - related to the previous point, it makes re-generation of a full distribution more difficult, as one has lots of distinct packages.
 - I'm not sure that a zip format is ideal for the web, but at the same time we do need some container to distributes multiple files in a python package.

Again I'm not saying that we should migrate to using wheels, but it could be interesting to consider what would be necessary to make it work.

hoodmane:
Use wheels. Resolves #655.

In the process, I rewrote most of `load-pyodide`, eliminating all the remaining black magic and confusing control flow.

It is worth pointing out that it is perfectly possible now to make `package.json` point at pypi, and by doing this get automatic installation into the repl for a preselected collection of pypi packages. The only difference between `loadPackage` and `micropip` is the dependency resolution. `loadPackage` is designed for "static" dependency resolution, whereas `micropip` is designed for "dynamic" dependency resolution. For this reason, it might be good to encourage use of `loadPackage`. I think ideally what we could do is make `micropip.freeze` which dumps a new `packages.json` with the currently loaded set of packages added in. Then a user can get the setup they want and freeze it so they won't have to redo dependency resolution every time their app starts.


hoodmane:
In particular, I think `loadPackage` should be changed to *only* load packages in `package.json`. For loading anything else, we should use `micropip`. This will lead to a clear distinction in purposes between these two tools.

hoodmane:
Probably we need an RFC for this of some sort.

hoodmane:
Screenshot from 2021-12-10 08-54-24
Most of the packages are building correctly.

hoodmane:
I don't think the `needs_rebuild` command is working correctly. It seems to rebuild stuff a lot even when it isn't needed.

hoodmane:
Okay at least locally all 89 package builds are working.

hoodmane:
Now all everything builds correctly, but there is some problem with test collection. All package tests get SKIPPED, while the micropip tests have a collection failure. Not sure why this happens.

hoodmane:
@rth Could we mark url arguments to `pyodide.loadPackage` as deprecated in v0.19 in preparation for this?

hoodmane:
Interestingly core tests only seem messed up on circleci, not on github actions.

hoodmane:
Okay deleting the unpacked test and wheel directories fixed it.

hoodmane:
Okay the core problems are now down to:
1. test_cpython_core ast is mysteriously failing
2. test_package_loading load_failure_retry needs to be rewritten or removed
3. node tests has `fetch` is not defined name errors 

Packages:
1. CLAPACK raises a wasm compilation error "CompileError: wasm validation error: at offset 4008999: failed to start custom section" which breaks all dependent packages
2. a few packages try to import from deleted test files.

hoodmane:
Strangely, clapack works fine if I zip it, but it breaks if I tar it. It seems that the untar utility is buggy in some way...

hoodmane:
Strangely, clapack works fine if I zip it, but it breaks if I tar it. It seems that the untar utility is buggy in some way...

hoodmane:
Okay, indeed now that I am zipping CLAPACK, it works fine. Some day maybe we'll have to investigate what the heck is going wrong with untarring .so files...

hoodmane:
Okay I think this is all set now.

packages/Makefile
```diff
@@ -21,6 +21,9 @@ all: .artifacts/bin/pyodide-build
 	mkdir -p $(PYODIDE_LIBRARIES)
 	$(HOSTPYTHON) -m pip install -e ../pyodide-build --no-deps --prefix $(PYODIDE_LIBRARIES)
 
+.artifacts/bin/wheel:
+	git clone git@github.com:hoodmane/wheel.git --branch handle-files-with-commas --depth 1 .artifacts/bin/wheel
```
ryanking13, line 25: Maybe add some comment here to tell why we use custom fork of wheel and when it will be fixed?

hoodmane: Right. It looks like my PR will be merged soon probably:
https://github.com/pypa/wheel/pull/427

hoodmane: Okay I can actually just use `find -name '*,*' -delete`. Somehow I never think of the simple solution until after I've implemented a complicated solution =P

rth:
Great progress @hoodmane ! I'll try to review in the next few days.

hoodmane:
Okay now this is great: the node test that is failing passes if you redirect standard error.

Makefile
```diff
@@ -173,17 +173,11 @@ clean-all:
 # Stdlib modules that we repackage as standalone packages
 
 # TODO: also include test directories included in other stdlib modules
-build/test.data: $(CPYTHONLIB) node_modules/.installed
-	./tools/file_packager.sh build/test.data --js-output=build/test.js \
-		--preload $(CPYTHONLIB)/test@/lib/python$(PYMAJOR).$(PYMINOR)/test
-	npx terser build/test.js -o build/test.js
+build/test.tar: $(CPYTHONLIB) node_modules/.installed
+	cd $(CPYTHONLIB) && tar -cvf $(PYODIDE_ROOT)/build/test.tar test
```
rth, line 177: Should we exclude __pycache__ just in case?

packages/mne/patches/0001-Force-mne-to-accept-old-scipy.patch
```diff
@@ -0,0 +1,24 @@
...
+
+diff --git a/requirements.txt b/requirements.txt
+index af5d063e2..adcec3348 100644
+--- a/requirements.txt
++++ b/requirements.txt
+@@ -1,6 +1,6 @@
+ # requirements for full MNE-Python functionality (other than raw/epochs export)
+ numpy>=1.15.4
+-scipy>=1.1.0
++scipy>=0.17.1
```
rth, line 18: Can probably be removed now

packages/statsmodels/meta.yaml
```diff
@@ -9,6 +9,9 @@ source:
     - patches/fix-scipy-blas-cythonize.patch
 
 build:
+  script: |
+    # Get rid of files with commas in their name, it breaks the wheel package.
+    find -name '*,*' -delete
```
rth, line 14: I image those are not important files?

pyodide-build/pyodide_build/buildpkg.py
```diff
@@ -39,6 +39,16 @@ def chdir(new_dir: Path):
 from .io import parse_package_config
 
 
+def make_whlfile(*args, owner=None, group=None, **kwargs):
```
rth, line 42:
```suggestion
def _make_whlfile(*args, owner=None, group=None, **kwargs):
```
to keep it private

pyodide-build/pyodide_build/buildpkg.py
```diff
@@ -410,23 +415,29 @@ def compile(
...
+        nmoved = unvendor_tests(wheel_dir, test_dir)
+    if nmoved:
+        with chdir(distdir):
+            shutil.make_archive(f"{pkg_name}-tests", "tar", test_dir)
+    pack_wheel(wheel_dir)
+    # wheel_dir causes pytest collection failures for in-tree packages like
+    # micropip. To prevent these, we get rid of wheel_dir after repacking the
+    # wheel.
+    shutil.rmtree(wheel_dir)
+    shutil.rmtree(test_dir, True)
```
rth, line 439:
```suggestion
    shutil.rmtree(test_dir, ignore_errors=True)
```

src/py/pyodide/package_loader.py
```diff
@@ -0,0 +1,39 @@
+import tarfile
```
rth, line 1: Functions need documentation type annotation, and maybe let's make this `_package_loader` if we don't intend to make it public?

src/tests/test_package_loading.py
```diff
@@ -1,86 +1,8 @@
 import pytest
-import shutil
 from pathlib import Path
+import shutil
 
-
-@pytest.mark.parametrize("active_server", ["main", "secondary"])
-def test_load_from_url(selenium_standalone, web_server_secondary, active_server):
```
rth, line 7: This should still work after the file format change to avoid breaking backward compatibility, unless there is a reason to remove it?

rth (commented):
Thanks for working on this! A first round of review is below. I reviewed everything except for src/js/load-pyodide.js where the diff is quite large and maybe let's discuss the general comments before going into details there.

> This introduces significant backwards incompatibility, so we should discuss the best way to migrate. 

Is this unavoidable in this PR? Can't we have only build & .js + (.data  -> .whl) format change and keep the dependency resolution mostly as it was? About static vs dynamic dependency resolution, I agree we need to move in that direction but it's a fairly complex design decision and I think we need to spend more time iterating on what usage should look like  (with examples) and get user feedback on that. So that would be better done in a separate issue, and it would have been ideal to switch to wheels separately from that discussion.

packages/mne/patches/0001-Force-mne-to-accept-old-scipy.patch
hoodmane: =)

packages/statsmodels/meta.yaml
hoodmane: No, they are test data files. But we actually fixed this twice over:
1. They are not in stats models 0.13.1
2. wheel package v0.37.1 contains my fix for the bug https://github.com/pypa/wheel/pull/427

ryanking13:
> Is this unavoidable in this PR? Can't we have only build & .js + (.data -> .whl) format change and keep the dependency resolution mostly as it was? About static vs dynamic dependency resolution, I agree we need to move in that direction but it's a fairly complex design decision and I think we need to spend more time iterating on what usage should look like (with examples) and get user feedback on that. So that would be better done in a separate issue, and it would have ideal to switch to wheels separately from that discussion.

I agree, this PR is trying to do too many things at once, so I'm afraid we might miss something important.

pyodide-build/pyodide_build/buildpkg.py
```diff
@@ -463,7 +373,6 @@ def build_package(path: Path, args):
             # i.e. they need package running, but not compile
             if not pkg.get("build", {}).get("sharedlibrary"):
                 compile(path, srcpath, pkg, args, bash_runner)
-            package_files(buildpath, srcpath, pkg, compress=args.compress_package)
```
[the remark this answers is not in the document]
hoodmane: Currently this is handled in buildall:
```py
        if self.shared_library:
            file_path = shutil.make_archive(
                f"{self.name}-{self.version}", "zip", self.pkgdir / "dist"
            )
            shutil.copy(file_path, outputdir)
            return
```
Maybe we should zip the file in buildpkg instead though? And of course we could use a tar file instead of a zip file if you prefer.

Makefile
hoodmane: Sure, done.

pyodide-build/pyodide_build/buildpkg.py
hoodmane: Okay, done.

hoodmane:
Why is the format check failing on `_micropip.py`?

ryanking13:
> Why is the format check failing on `_micropip.py`?

You should update your local docker image (or local black version)
See https://github.com/pyodide/pyodide/pull/2084

hoodmane:
> You should update your local docker image 

I'm confused about this: I see `pyodide/pyodide-env:19` in `config.yml` and `PYODIDE_IMAGE_TAG="19"` in `run_docker` and I am running `pyodide/pyodide-env:19` locally. So why is the version of `black` different?

ryanking13:
> > You should update your local docker image
> 
> I'm confused about this: I see `pyodide/pyodide-env:19` in `config.yml` and `PYODIDE_IMAGE_TAG="19"` in `run_docker` and I am running `pyodide/pyodide-env:19` locally. So why is the version of `black` different?

We updated the docker image with the same tag, you have to remove your local image (`docker rmi`) and pull it again.
Maybe we better use a different tag whenever we push a updated docker image.

src/tests/test_package_loading.py
hoodmane: Okay fixed.

hoodmane:
@rth Okay, I think I addressed the first round of comments. At least I fixed the dependency overrides at least. The dependency overrides aren't terrible, but at some later point it would be good to consider whether to change our approach to them.

> `src/js/load-pyodide.js` where the diff is quite large

I split some of those changes into #2106, I will try to move more over there.

hoodmane:
Okay, so about 50% of the changes to `load-pyodide.js` I pulled out into #2106 (which no longer is accurately named hmm).

hoodmane:
Apparently setuptools comes with 700kb of windows binaries.

packages/distlib/meta.yaml
```diff
@@ -1,11 +1,14 @@
 package:
   name: distlib
-  version: 0.3.1
+  version: 0.3.4
 source:
-  sha256: edf6116872c863e1aa9d5bb7cb5e05a022c519a4594dc703843343a9ddd9bff1
-  url: https://files.pythonhosted.org/packages/2f/83/1eba07997b8ba58d92b3e51445d5bf36f9fba9cb8166bcae99b9c3464841/distlib-0.3.1.zip
+  sha256: 6564fe0a8f51e734df6333d08b8b94d4ea8ee6b99b5ed50613f731fd4089f34b
+  url: https://files.pythonhosted.org/packages/ac/a3/8ee4f54d5f12e16eeeda6b7df3dfdbda24e6cc572c86ff959a4ce110391b/distlib-0.3.4-py2.py3-none-any.whl
+  finished_wheel: True
```
rth, line 7: Do we need this parameter? We could pretty much look at the extension and the file ending with `none-any` couldn't we ?

pyodide-build/pyodide_build/buildall.py
```diff
@@ -33,7 +33,9 @@ class BasePackage:
     dependencies: List[str]
     unbuilt_dependencies: Set[str]
     dependents: Set[str]
-    unvendored_tests: Optional[bool] = None
+    unvendored_tests: Optional[Path] = None
+    file_name: Optional[str] = None
+    install_dir: str = "site"
```
rth, line 38:
```suggestion
    install_dir: Path = Path("site")
```
?

pyodide-build/pyodide_build/buildall.py
```diff
@@ -476,11 +472,21 @@ def build_packages(packages_dir: Path, outputdir: Path, args) -> None:
...
 
     build_from_graph(pkg_map, outputdir, args)
+    for pkg in pkg_map.values():
+        if isinstance(pkg, StdLibPackage):
+            pkg.file_name = pkg.name + ".tar"
+            continue
+        if pkg.shared_library:
+            pkg.file_name = f"{pkg.name}-{pkg.version}.zip"
+            continue
+        for file in (pkg.pkgdir / "dist").glob("*.whl"):
```
rth, line 482: That's a bit wide. Maybe let's at least put the `pkg.name.replace('-', '_')` and the version in the glob and make a case insensitive search. Then check that there is only one result? Or alternatively validate the path with https://packaging.pypa.io/en/latest/utils.html#packaging.utils.parse_wheel_filename

pyodide-build/pyodide_build/buildpkg.py
```diff
@@ -308,29 +307,29 @@ def patch(pkg_root: Path, srcpath: Path, src_metadata: Dict[str, Any]):
         fd.write(b"\n")
 
 
+def unpack_wheel(path):
+    with chdir(path.parent):
+        subprocess.run([sys.executable, "-m", "wheel", "unpack", path.name], check=True)
```
rth, line 312: We could use it directly without subprocess (with import of wheel inside this function) https://github.com/pypa/wheel/blob/c716c87ffa528ac9a31de3ed3217058848dc2cd2/src/wheel/cli/unpack.py however it's unclear how stable their public API is. I would guess it shouldn't be too high for these two functions https://github.com/pypa/wheel/issues/262

pyodide-build/pyodide_build/buildpkg.py
```diff
@@ -409,26 +401,57 @@ def compile(
                 replay_from=replay_from,
             )
         install_for_distribution()
+    del bash_runner.env["_PYTHON_HOST_PLATFORM"]
+
+
+def package_wheel(pkg_name, pkg_root, srcpath, build_metadata, bash_runner):
+    """
+    Parameters
```
rth, line 409: Nit: docstrings need a 1 line description followed by an empty line )

```suggestion
    """Package a wheel
    
    Parameters
```

pyodide-build/pyodide_build/buildpkg.py
```diff
@@ -409,26 +401,57 @@ def compile(
                 replay_from=replay_from,
             )
         install_for_distribution()
+    del bash_runner.env["_PYTHON_HOST_PLATFORM"]
+
+
+def package_wheel(pkg_name, pkg_root, srcpath, build_metadata, bash_runner):
```
rth, line 407: Could you add type annotations, please ?

src/js/load-pyodide.js
```diff
@@ -398,68 +395,69 @@ export async function loadPackage(names, messageCallback, errorCallback) {
...
+      if (loadedPackages[name]) {
+        // Handle the race condition where the package was loaded between when
+        // we did dependency resolution and when we acquired the lock.
+        toLoad.delete(name);
+        continue;
+      }
+      if (channel === DEFAULT_CHANNEL) {
+        packagePromises[name] = downloadPackage(name);
+      } else {
+        packagePromises[name] = downloadPackage(channel);
```
rth, line 428: Very minor: Maybe better to define a variable that's going to be passed to downloadPackage, to avoid repeating this line twice?

src/py/pyodide/_package_loader.py
```diff
@@ -0,0 +1,84 @@
...
+STD_LIB = Path(sysconfig.get_path("stdlib"))  # type: ignore
+TARGETS = {"site": SITE_PACKAGES, "lib": STD_LIB}
+ZIP_TYPES = {".whl", ".zip"}
+TAR_TYPES = {".tar", ".gz", ".bz", ".gz", ".tgz", ".bz2", ".tbz2"}
+
+
+def unpack_buffer(filename: str, buffer: JsProxy, target: str = "site") -> JsProxy:
+    """
+    Used to install a package either into sitepackages or into the standard library.
+    This is a helper method called from ``loadPackage``.
```
rth, line 23:
```suggestion
    """Used to install a package either into sitepackages or into the standard library.
    
    This is a helper method called from ``loadPackage``.
```
Assuming the first line is not too long.

src/py/pyodide/_package_loader.py
```diff
@@ -0,0 +1,84 @@
...
+from zipfile import ZipFile
+
+SITE_PACKAGES = Path(getsitepackages()[0])
+STD_LIB = Path(sysconfig.get_path("stdlib"))  # type: ignore
+TARGETS = {"site": SITE_PACKAGES, "lib": STD_LIB}
+ZIP_TYPES = {".whl", ".zip"}
+TAR_TYPES = {".tar", ".gz", ".bz", ".gz", ".tgz", ".bz2", ".tbz2"}
+
+
+def unpack_buffer(filename: str, buffer: JsProxy, target: str = "site") -> JsProxy:
```
rth, line 20: filename as Path should also work I think?

src/py/pyodide/_package_loader.py
```diff
@@ -0,0 +1,84 @@
...
+    target_dir = TARGETS[target]
+    with NamedTemporaryFile(suffix=filename) as f:
+        buffer._into_file(f)
+        shutil.unpack_archive(f.name, target_dir)
+        return to_js(get_dynlibs(f, target_dir))
+
+
+def get_dynlibs(archive: IO[bytes], target_dir: Path) -> List[str]:
+    """
+    List out the paths to .so files in a zip or tar archive.
```
rth, line 54:
```suggestion
    """List out the paths to .so files in a zip or tar archive.
```

src/py/pyodide/_package_loader.py
```diff
@@ -0,0 +1,84 @@
...
+    elif suffix in TAR_TYPES:
+        dynlib_paths_iter = (tinfo.name for tinfo in tarfile.open(archive.name))
+    else:
+        raise ValueError(f"Unexpected suffix {suffix}")
+
+    return [
+        str((target_dir / path).resolve())
+        for path in dynlib_paths_iter
+        if path.endswith(".so")
+    ]
```
rth, line 84: Can we add a test for this (running in system python not Pyodide)?

rth (commented):
Thanks a lot for doing this @hoodmane ! It looks great and I'm very much looking forward to using it.

A few comments below but overall it looks good to me.
The file sizes seem to be much smaller,
```
-rw-r--r-- 1 root root 7.1M Jan 19 11:48 numpy.data
[...]
-rw-r--r-- 1 root root 3.4M Jan 14 00:52 numpy-1.21.4-cp39-cp39-emscripten_wasm32.whl
```
we will see what's the impact after brotli compression but I though that .data was initially gzip compressed as well? So it looks like the gzip compression is significantly higher?
In some later time, we might want to experiment whether it's better to create wheels with the default zip compression or with 0 compression and let the browser compression be used instead.

About load times, our benchmarks are not too reliable but, we have

```
 -  load numpy  (main)
    native: nan  firefox: 0.938628  chrome: 3.942906
 -  load numpy  (this PR)
    native: nan  firefox: 0.837011  chrome: 3.967529 
```
so overall it looks comparable.

pyodide-build/pyodide_build/buildall.py
hoodmane: Well this is a choice between "site" ==> `SITE_PACKAGES` and "lib" ==> `STD_LIB`. I look up the proper locations for `SITE_PACKAGES` from `site.getsitepackages()` and for `STD_LIB` from `sysconfig.get_path("stdlib")` at the top of the file. Not sure if using a string for the setting is standard or what.

pyodide-build/pyodide_build/buildall.py
hoodmane: How do I do case insensitive search?

rth: Maybe https://stackoverflow.com/a/8151371/1791279 but yeah it looks not obvious. Just do something to avoid the situation where multiple wheels end up being in that folder and we pick the wrong one.

src/js/load-pyodide.js
hoodmane: Yeah this is a bit sloppy. I rearranged the logic a bit, I think it's better now.

pyodide-build/pyodide_build/buildall.py
hoodmane: How about for now we add an assert that it only contains one `.whl` file? The options should either be a `.whl`, a `.whl` and a `.tar` with unvendored tests, or a `.tar` with a shared lib.

hoodmane: I guess maybe if you update the version of the package or something, we will end up with two wheel files in the `dist` directory. Probably we'd want to pick the right wheel in that case rather than crashing. Or we could empty the `dist` directory before putting the new stuff into it...

src/py/pyodide/_package_loader.py
hoodmane: Yeah it would work, though we are calling this from Javascript so in practice it's gonna be a string.

src/py/pyodide/_package_loader.py
hoodmane: Sure. What's the best way to make the test files? I guess we could generate them with Python code in the test, would probably be easier just to include a wheel and a tar file in some `test_data` directory though.

packages/distlib/meta.yaml
hoodmane: Yeah, that's definitely possible, perhaps that'd be better. After we've merged this PR we should change all the unpatched pure Python packages to use this method.

src/py/pyodide/_package_loader.py
hoodmane: Even with patches, we could potentially download wheel, unpack wheel, apply patch, repack wheel. Then again, if we do this to avoid handling the alternate build backends used by `typing_extensions` and `distlib`, maybe we'd be better off actually figuring out how to handle alternate build backends. It doesn't look that hard.

packages/distlib/meta.yaml
hoodmane: Even with patches, we could potentially download wheel, unpack wheel, apply patch, repack wheel. Then again, if we do this to avoid handling the alternate build backends used by `typing_extensions` and `distlib`, maybe we'd be better off actually figuring out how to handle alternate build backends.

hoodmane: Also, would be good to update `mkpkg` to prefer to take pure Python wheels over sdists. Maybe with a flag to indicate that we actually want an sdist. For updates, can default to keeping the same type.

pyodide-build/pyodide_build/buildall.py
hoodmane: Okay, now I explicitly assert that there is only one wheel in the dist directory. I think this makes it less error prone.

src/py/pyodide/_package_loader.py
rth: Generating them at test time in some tmp folder would be better IMO: easier to both know what's inside and change something if needed.

hoodmane:
@rth @ryanking13 Okay, tests pass again. Any further comments? How do you feel about merging this?

rth (approved):
The "src/tests/data/test-scipy.{whl,tar.bz2}" files could be removed if they are not used?

Otherwise LGTM. Thanks!

conftest.py
```diff
@@ -72,8 +72,20 @@ def pytest_collection_modifyitems(config, items):
         _maybe_skip_test(item, delayed=True)
 
 
+@functools.cache
+def built_packages():
+    """Returns a list of built package names.
+
+    This functions lists the names of the .data files in the build/ directory.
```
ryanking13, line 79:
```suggestion
    This function lists the names of the .whl files in the build/ directory.
```

packages/test_packages_common.py
```diff
@@ -26,17 +27,10 @@ def built_packages() -> List[str]:
...
-    registered_packages_ = registered_packages()
-    packages = []
-    for fpath in os.listdir(BUILD_DIR):
-        if not fpath.endswith(".data"):
-            continue
-        name = fpath.split(".")[0]
-        if name in registered_packages_:
-            packages.append(name)
-    return packages
+    return list(json.loads(packages_json_path.read_text())["packages"].keys())
```
ryanking13, line 33: Would it be hard to avoid having two same functions `built_packages` in both here and in `conftest.py`?
I'm not sure which would be the best place to put this function (maybe somewhere inside `pyodide-build`?)

packages/distlib/meta.yaml
```diff
@@ -1,11 +1,13 @@
 package:
   name: distlib
-  version: 0.3.1
+  version: 0.3.4
```
ryanking13, line 3: It needs a changelog for version update

ryanking13 (approved):
Thanks for your huge effort! I have some minor comments, otherwise LGTM. And please update the changelog for this.

ryanking13:
We should also update the documentation which refers `.data`, but I think we could do that in a separate PR since this PR is already big enough.

packages/test_packages_common.py
hoodmane: We can just import it directly from conftest actually.

rth:
I really like that one can now just call `less` to list files in a wheel or .tar file, and `zcat` to get wheels contents. It makes debugging so much easier. Thanks again @hoodmane !

henryiii:
Is there a documented process somewhere for a user to build wheels? If I have a simple C/C++ package, could I produce one of these wheels now? (And ideally, the `emscripten_wasm32` tag should be allowed in PyPI, is that being pursued?)

rth:
> Is there a documented process somewhere for a user to build wheels? 

For pure Python packages, it's just regular universal Python wheels.
For packages with C extensions, you can run inside the pyodide Docker image,
```
PYODIDE_PACKAGES="<package-name>" make
```
see the dev docs https://pyodide.org/en/latest/development/building-from-sources.html and https://pyodide.org/en/latest/development/new-packages.html.  We are trying to migrate to a better build setup & CLI. `latest` vs `stable` in the URL to get the dev docs, since this feature is not yet included in the release.
This assumes though that you have a `meta.yaml` package defining the package and `meta.yaml` for its dependencies.

> And ideally, the emscripten_wasm32 tag should be allowed in PyPI, is that being pursued?

Yes, see https://github.com/pypa/warehouse/issues/10416 . Though it's likely a multi-year project. The feedback we got from core devs is that first CPython needs to officially support WASM (which hopefully would happen for the next release), then there need to be a PEP and enough traction/usage for Python with WASM from the community.  So the short term goal is just to deploy wheels to some other wheel services (or keep our custom deployment). A bit similar to what is done for https://github.com/antocuni/pypy-wheels

Having the ability to build wheels for WASM as part of cibuildwheel would be indeed ideal. I'll comment more on that in https://github.com/pypa/cibuildwheel/issues/1002

henryiii:
I'd be quite fine with that - calling this still exploratory, keeping the ABI from gaining requirements quite yet. I think it still could be "standalone" via cibuildwheel & probably docker.

PR #2030, opened 2021-12-11
ENH Destroy awaitable PyProxy after awaiting it

hoodmane:
This makes the overwhelmingly common use case very simple, at the expense of making some rare use cases more complicated.

### Checklists

- [x] Add / update tests

### DIFF
--- a/docs/project/changelog.md
+++ b/docs/project/changelog.md
@@ -93,6 +93,32 @@ substitutions:
   that looks up the name on `builtins` if lookup on `globals` fails.
   {pr}`1905`
 
+- {{Enhancement}} Coroutines have their memory managed in a more convenient way.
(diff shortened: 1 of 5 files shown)


rth:
I'm not sure I follow all the implications of this change. In particular, does it mean that awaiting twice would no longer raise errors?

 Could you please elaborate?

hoodmane:
Currently:
```js
let mypkg = pyodide.pyimport("mypkg");
let coroutine = mypkg.async_entrypoint();
let result;
try {
   result = await coroutine
} finally {
   coroutine.destroy();
}
```
With this PR:
```js
let mypkg = pyodide.pyimport("mypkg");
let result = await mypkg.async_entrypoint();
```

> In particular, does it mean that awaiting twice would no longer raise errors?

Yes, this is a minor benefit we pick up on the side.

PR #2031, opened 2021-12-11
Rewrite intro text

hoodmane:


ryanking13:
I always thought that the current catchphrase of the Pyodide (`Python with the scientific stack, compiled to WebAssembly.`) is too narrow to describe the possibility of the Pyodide. So I'm +1 to changing it.

But it is used in many other places like below and it actually defines what Pyodide is, so I think we need to carefully decide a single sentence that describes what the Pyodide is.

- Github about page
- Intro of the documentation: https://github.com/pyodide/pyodide/blob/main/docs/index.rst
- Twitter profile: https://twitter.com/pyodide
- Opencollective profile: https://opencollective.com/pyodide

hoodmane:
Yeah I'm aware that these updates would need a significant amount of propagation if accepted. I wanted to discuss the wording and see if you and @rth and other folks like it before doing those updates. But I think saying we are a web browser port of Python is more accurate and easier to understand.

hoodmane:
We could look through and see which packages we support are downloaded the most on PyPi and consider mentioning a few of them.

README.md
```diff
@@ -9,22 +9,26 @@
...
-are currently available. In addition, it's possible to install pure Python wheels
+We also port over 80 common Python extensions written in C, C++, and fortran.
+Supported packages range from pyyaml and lxml, which use C extension to do fast
+yaml and xml processing, to standard scientific computing packages like numpy,
+pandas, scipy, and matplotlib. It is also possible to use pure Python wheels
 from PyPI.
 
-Pyodide provides transparent conversion of objects between JavaScript and
-Python. When used inside a browser, Python has full access to the Web APIs.
+Pyodide comes with a robust Javascript <==> Python foreign function interface so
```
rth, line 26:
```suggestion
Pyodide comes with a robust Javascript 🡘 Python foreign function interface so
```

README.md
```diff
@@ -9,22 +9,26 @@
...
+that you can freely mix these two languages in your code with absolutely minimal
```
rth, line 27:
```suggestion
that you can freely mix these two languages in your code with minimal
```

README.md
```diff
@@ -9,22 +9,26 @@
...
 
-Python with the scientific stack, compiled to WebAssembly.
+Pyodide is a port of the Python 3.9 interpreter and many common Python packages
+to the browser.
 
 ## What is Pyodide?
 
-Pyodide may be used in any context where you want to run Python inside a web
-browser.
+Pyodide is a port of Python to WebAssembly/[Emscripten](https://emscripten.org/)
```
rth, line 17:
```suggestion
Pyodide is a port of CPython to WebAssembly/[Emscripten](https://emscripten.org/)
```

README.md
```diff
@@ -9,22 +9,26 @@
...
+for use in the browser or in node.
```
rth, line 18:
```suggestion
for use in the browser or in Node.js.
```

README.md
```diff
@@ -9,22 +9,26 @@
 [![Build Status](https://circleci.com/gh/pyodide/pyodide.png)](https://circleci.com/gh/pyodide/pyodide)
 [![Documentation Status](https://readthedocs.org/projects/pyodide/badge/?version=stable)](https://pyodide.readthedocs.io/?badge=stable)
 
-Python with the scientific stack, compiled to WebAssembly.
+Pyodide is a port of the Python 3.9 interpreter and many common Python packages
+to the browser.
```
rth, line 13: Will comment about this seprately.

README.md
```diff
@@ -9,22 +9,26 @@
...
-Pyodide brings the Python 3.9 runtime to the browser via WebAssembly thanks to
-[Emscripten](https://emscripten.org/).
-It builds the Python scientific stack including NumPy, Pandas, Matplotlib, SciPy, and
-scikit-learn. The [packages directory](packages) lists over 75 packages which
-are currently available. In addition, it's possible to install pure Python wheels
+We also port over 80 common Python extensions written in C, C++, and fortran.
+Supported packages range from pyyaml and lxml, which use C extension to do fast
+yaml and xml processing, to standard scientific computing packages like numpy,
+pandas, scipy, and matplotlib. It is also possible to use pure Python wheels
 from PyPI.
```
rth, line 24: In our current description really focuses on those 80 packages, which raises the question of why so few. Installing from PyPi is added as an afterthought.

To be more general, maybe,
```suggestion
Pyodide allows to install and run Python packages in the browser with [micropip](https://pyodide.org/en/stable/usage/api/micropip-api.html),
 - pure Python packages can be installed as wheels from PyPi
 - many [common packages](https://github.com/pyodide/pyodide/tree/main/packages) with C extensions are built as part of Pyodide. These range from general-purpose packages such as regex, pyyaml, lxml to 
 scientific Python packages including numpy, pandas, scipy, matplotlib, and scikit-learn.
```

It's not fully accurate, because we also build pure Python packages, but it would illustrate the direction where we are going.

rth (commented):
Thanks it could use updating indeed.

> I always thought that the current catchphrase of the Pyodide (Python with the scientific stack, compiled to WebAssembly.) is too narrow to describe the possibility of the Pyodide. So I'm +1 to changing it to a broader sentence.

I'm +1 on updating the sentence describing the project, however, I think the proposed one,

> Pyodide is a port of the Python 3.9 interpreter and many common Python packages
to the browser.

is a bit long and not very specific. Reading this sentence someone should be able to tell what the project does and how is it different from other similar projects. There are many attempts to port Python to the browser. Also, 
- it might be better to put the version there, avoid changing it everywhere each time we update the CPython version. It should be in the description below though.
- I'm not sure to what extent we can about a port of Python, given that more of this stuff is likely going to be done upstream with time.

It's not really a criticism, more some though so we could iterate on it. It's hard to get it right.

I also wonder if the term "Python distribution" wouldn't be more appropriate (it's a shorter way of saying it allows to use Packages), but maybe it doesn't include all the JS related work we do.
How about,
```
Pyodide is a  Python distribution for the browser and Node.js based on WebAssembly
```
?

hoodmane:
> Pyodide is a Python distribution for the browser and Node.js based on WebAssembly

hoodmane:
@rth @ryanking13 How does it look now?

ryanking13:
It looks good to me!

I think there is some documentation that needs to be updated together:

- https://github.com/pyodide/pyodide/blob/main/docs/index.rst
- https://github.com/pyodide/pyodide/blob/main/docs/project/about.md

rth (approved):
Thanks, LGTM as well. We would indeed need to propagate these changes everywhere.

phorward:
I'm also fine with this +1

PR #2019, opened 2021-12-05
Remove function pointer cast emulation

hoodmane:
This is a third attempt at the fpcast removal. The good news is that chrome is working okay here.

I guess this one works.

Closes #2016. Closes #1677. Closes #1577.

PR #1818, opened 2021-09-06
Fix Pillow JPEG encode/decode error

ryanking13:
This PR fixes JPEG encode/decode errors in python pillow module.

See: #1817

Closes #1201

**Edit:** Related #1781, #1632 (I'll close them manually after checking that the issue is fixed there).

PR #2034, opened 2021-12-11
BLD Remove libjpeg from the main module

### LINKED ISSUE #1817: Investigating pillow JPEG error
Related Issues:
- #1781
- #1632
- #1198
- #1201
- #1657 (might be related)

I have been investigating an error in pillow JPEG encode/decode functionality,
and I found that something wrong is happening because libjpeg library is linked to pillow side module.

When I [modified the code](https://github.com/ryanking13/pyodide/commit/4eead177abc8b7fb6f1295e74e81fb3f1ef3243f) such as:

1. **Link libjpeg with main module**

image

2. **Remove linking ligjpeg in pillow side module**

image

It looks like the jpeg error is fixed.

---

Now, things I want to discuss are:

1. **Why linking a library to a side module generates an error?**

The [Emscripten docs](https://github.com/emscripten-core/emscripten/wiki/Linking) says:

```
1. Main modules, which have system libraries linked in.
2. Side modules, which do not have system libraries linked in.
```

As it says, Emscripten suggests not to link system libraries to a side module.

But [when I made a simple Emscripten code which links system libraries to a side module](https://gist.github.com/ryanking13/5e0e0d36a60ed947b94286e82b4a40d7), It builds and runs with no error.
I am very new to Emscripten, so I am not sure how linking system libraries to a side module can generates an error.

2. Is adding ligjpeg to the main module is the right solution?

Adding libjpeg to the main module results in the code size increase of pyodide.asm.{js,wasm}.

- pyodide.asm.js: ~32KB increased
- pyodide.asm.wasm: ~440KB increased

```sh
# code size in bytes (linking libjpeg to the pillow)
1723564 pillow.data
38899 pillow.js
1998551 pyodide.asm.js
11729609 pyodide.asm.wasm
```
```
#  code size in bytes (linking libjpeg to the main module)
782498 pillow.data
23395 pillow.js
2031046 pyodide.asm.js
12180594 pyodide.asm.wasm
```

I'm not sure this amount of code size increase is critical.
But if there is another solution not linking libjpeg to the main module, I believe that will be better.
(Maybe we can select whether to link libjpeg dynamically when building?)

ryanking13:
### Description

Closes #1817

Removes libjpeg from the main Pyodide module, and statically links it to the side module (Pillow package) that actually uses it, which is reverting the change in #1818.

In #1818, Pillow jpeg module didn't work when libjpeg was linked to it, so I fixed it by linking libjpeg to the main module. But I recently found out that the function pointer casting was related to the linking problem, and after it's removal at #2019, it looks like the problem is fixed.

Therefore, we can now revert the previous change, which makes the main module (`pyodide.asm.wasm`) a little lighter (~0.4MB).

(It's very interesting to see that function pointer casting was generating several errors.)


hoodmane:
It's worth pointing out that some of the emscripten devs were surprised that it worked for us to link things like this at all. They said this libraries should always be linked into the main module. But since it works it works.

hoodmane:
@ryanking13 Could you add an empty commit to trigger CI?

ryanking13:
> It's worth pointing out that some of the emscripten devs were surprised that it worked for us to link things like this at all. They said this libraries should always be linked into the main module. But since it works it works.

Yeah, there are many things that I can't fully understand "why it works" or "why it doesn't work". It just miraculously works 🤔

> @ryanking13 Could you add an empty commit to trigger CI?

Triggered CI!

PR #1952, opened 2021-11-12
MAINT Update Chrome, Firefox and Selenium

rth:
Updates to,
 - Chrome 94
 - Firefox 94
 - Selenium 4.0.0

PR #2035, opened 2021-12-11
MAINT Update emscripten to 2.0.24

hoodmane:


hoodmane:
Well this definitely looks more promising than last time we tried to upgrade. test-packages-chrome shows a bunch of timeouts though...

hoodmane:
Chrome CI is really creating a lot of trouble lately =( 
I guess I should try to push #1952 forward...

hoodmane:
Seems like improvements to dynamic linking require wading through a new round of scipy Fortran hell to fix more stuff. Probably these linker errors are caused by real call signature mismatch errors that existed before too though, so I think this is a win.

hoodmane:
Okay this is ready to go after #2289 is merged.

PR #2040, opened 2021-12-13
MAINT Remove fpcast flags from CLAPACK

hoodmane:
I cannot comprehend why these flags don't break clapack...


rth:
> I cannot comprehend why these flags don't break clapack...

It's indeed extremely strange. Maybe it doesn't do any fpcasting?

hoodmane:
Yeah, the only reasonable interpretation is that no function pointers are ever passed into or out of CLAPACK.

PR #2041, opened 2021-12-13
Add dlib package and setup flags to meta.yaml 

viktorChekhovoi:
### Description

Closes #2039


packages/dlib/meta.yaml
```diff
@@ -0,0 +1,18 @@
...
+  version: 19.22.1
+
+source:
+  url: https://files.pythonhosted.org/packages/f0/a2/ba6163c09fb427990180afd8d625bcecc5555af699c253193c35ffd48c4f/dlib-19.22.1.tar.gz
+  sha256: 77e943b40cc53e377d7fc683552d2337937f1111c46d7753105f594c3958163c
+
+build:
+  cxxflags: -std=c++11
+  cflags: -std=c++11
+  setupflags: |
```
rth, line 12: Maybe,

```suggestion
  setup_install_args: |
```
?

I find it fairly unreadable otherwise without spaces.

rth (commented):
Thanks !

pyodide-build/pyodide_build/pywasmcross.py
```diff
@@ -461,6 +462,14 @@ def make_parser(parser):
             "Note: this is a private endpoint that should not be used "
             "outside of the Pyodide Makefile."
         )
+        parser.add_argument(
+            "--setupflags",
+            type=str,
+            nargs="?",
+            default="",
+            help="Extra flags to pass to setup.py",
```
henryiii, line 470: Building via setup.py is very [heavily deprecated](https://blog.ganssle.io/articles/2021/10/setup-py-deprecated.html). I wouldn't recommend adding anything that depends on it. Instead, I'd pass things though by environment variables. Conda-forge uses CMAKE_ARGS, and scikit-build uses SKBILD_CONFIGURE_OPTIONS, though I'm working on possibly adding support for CMAKE_ARGS there.

hoodmane: > Building via setup.py is very [heavily deprecated](https://blog.ganssle.io/articles/2021/10/setup-py-deprecated.html).

Thanks, that is interesting. I have been hoping to get away from it to something that respects packages' chosen build backends. Directly downloading pure Python wheels has given us a bit of a reprieve for packages that don't use setuptools as their backend but we still have the extensions to manage, and more and more packages use other build backends. But the packaging ecosystem is pretty complicated so it's hard to know the right approach.

henryiii: I wonder if https://pypi.org/project/installer/ would be useful for the non-PEP 517 pure python libs (there seem to be quite a few in /packages, about halfish?). It's targeting making it easier for distributions to install other key PyPA packages without needing to care about the build backend. It doesn't help for building things, of course. It just had its first useful release a couple of days ago. We are hoping to start using flit for some packages, especially ones with chicken-and-egg problems with setuptools.

henryiii: For building, what would be required to make `pypa/build` work, I wonder - that should be the wheel-making tool, and it's very much designed to be easy to bootstrap.

hoodmane: > what would be required to make `pypa/build` work, I wonder 

I was looking into that in December. My guess is that it will only take elbow grease.

hoodmane: I think #2027 cleared our main blocker for this.

hoodmane: One potential issue is that I believe pypa/build has build isolation whereas our current system does not, and particularly the numpy/scipy dependant packages rely on this. This is also a difficulty we discussed in #2179. But I don't really understand the ramifications of build isolation, so maybe this won't be an issue?

henryiii: You can use `--no-isolation`.

ryanking13: > I wonder if https://pypi.org/project/installer/ would be useful for the non-PEP 517 pure python libs

I was looking into it recently. It looks interesting, but it doesn't seem very stable for now.
(I was about to say that it also doesn't have proper CLI yet, but... it [has been released](https://pypi.org/project/installer/0.5.0/#history) yesterday 👀)

rth: Opened https://github.com/pyodide/pyodide/issues/2189 to contunue this discussion

PR #2042, opened 2021-12-14
MAINT Add a test for shared libraries

hoodmane:
It's annoying that the only shared library is CLAPACK and that it is used by scipy which takes forever to build. scipy is also very complicated and can fail for reasons specific to scipy or for reasons general to other shared libraries. This adds a tiny shared library and a tiny C extension that uses it so we can get faster feedback about whether shared libraries are working correctly.


rth (approved):
Thanks for doing this!

PR #2050, opened 2021-12-17
MAINT Don't include debug symbols in distribution

hoodmane:
Should reduce built size a lot. But we used to be using `-g` so I don't know why there's a size issue now.


ryanking13:
```
-rwxrwxrwx 1 siha siha 5.1M Dec 17 18:27 pyodide.asm.data
-rwxrwxrwx 1 siha siha 2.0M Dec 17 18:27 pyodide.asm.js
-rwxrwxrwx 1 siha siha 9.1M Dec 17 18:27 pyodide.asm.wasm
```

Happy to see that the size of `pyodide.asm.wasm` is now < 10MB.

PR #2051, opened 2021-12-18
Fix reentrant error test

ryanking13:
### Description

`test_reentrant_error` is always failing on CI in the first place and successes in the retrial.
This is due to the duplicated function name used in the test.


hoodmane:
Thanks @ryanking13!

PR #2053, opened 2021-12-19
MAINT Clean up environment variables and pywasmcross

hoodmane:
This is more maintenance on `pyodide-build`. The goal is to make it so that directly invoking `pyodide-build` from the command line actually works. This would be a huge benefit to debugging package builds, where it's a pain to select out individual failing build commands and repeat them because of issues with environment variables.

In particular, this fixes 
`python -m pyodide_build buildall ../packages/ ../build --only="core"`
so that it runs without errors.

I also made some readability improvements to pywasmcross.


rth (commented):
Thanks for doing this work! LGTM, when CI passes.

It would indeed be so much better to be able to run that command without the makefile.

ryanking13:
It looks like `*.gnu.org` is (temporarily) down, so fetching `libiconv` tarball failing.

hoodmane:
And it works! Just have to fix test-python.

PR #2058, opened 2021-12-20
MAINT Update to emscripten 3.0.0

hoodmane:
Let's see what happens!


hoodmane:
Yeah I'd be very happy if we could get this updated. Last time I checked even one patch version up to 2.0.17 is broken. @ryanking13's recent work on the ci gives me some hope, but we have very large exposure to the emscripten interfaces so we get broken by a lot of patch updates.

PR #2059, opened 2021-12-20
MAINT More pywasmcross cleanup

hoodmane:


pyodide-build/pyodide_build/pywasmcross.py
```diff
@@ -382,33 +515,42 @@ def replay_command(line, args, dryrun=False):
...
+    if not build_log_path.is_file():
+        return
+
+    lines_str = (
+        subprocess.check_output(["wc", "-l", str(build_log_path)])
+        .decode()
+        .split(" ")[0]
+    )
+    num_lines = str(lines_str)
+    with open(build_log_path, "r") as fd:
```
rth, line 549:
```suggestion
    with open(build_log_path, "r") as fd:
        num_lines = sum(1 for _ in fd)
        fd.seek(0)
```

pyodide-build/pyodide_build/pywasmcross.py
```diff
@@ -197,55 +203,199 @@ def replay_f2c(args, dryrun=False):
...
+        and "site-packages" not in arg
+    ):
+        return arg.replace("-I" + sys.prefix, "-I" + target_install_dir)
+    # Don't include any system directories
+    if arg[2:].startswith("/usr"):
+        return None
+    return arg
+
+
+SYS_ROOT_RE = re.compile(",--sysroot=[^,]*")
```
rth, line 313: Are you sure the solution without the regex wasn't better? Also this will match anything that doesn't have "," including spaces and I don't think that's correct?

rth (commented):
Thanks! A few comments otherwise LGTM.

pyodide-build/pyodide_build/pywasmcross.py
hoodmane: Spaces shouldn't appear, unless it was something like `--sysroot="/a/path with spaces/"` in which case we should get rid of them. This does fail on `--sysroot="/a/path,with,commas/"`, but that should never happen I hope.

The previous solution without the regex was incorrect. However there is a better approach I can switch to.

PR #2062, opened 2021-12-21
DOC Fix type-conversions.md example 

wisehackermonkey:
### Description
 
  the reasoning for the change
     - broken examples when run in the browser
details of updated code
       1) missing `import pyodide` in example
       2) when run in the browser `let` prohibits the example from running changing it to `var` makes the example run
### Checklists

### DIFF
--- a/docs/usage/type-conversions.md
+++ b/docs/usage/type-conversions.md
@@ -619,9 +619,10 @@ directly (i.e., `jsfunc(obj)`) or as `jsfunc(create_proxy(obj))`.
 `create_proxy` is particularly helpful with `addEventListener`:
 
 ```py
+import pyodide
+
 def callback():
     print("clicked!")
-import pyodide
 proxy = pyodide.create_proxy(callback)
 from js import document
 document.body.addEventListener("click", proxy)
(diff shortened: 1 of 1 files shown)


hoodmane:
> when run in the browser let prohibits the example from running changing it to var makes the example run

What do you mean by that? It runs fine with `let`? Which browser are you testing this in?

docs/usage/type-conversions.md
```diff
@@ -621,6 +621,7 @@ directly (i.e., `jsfunc(obj)`) or as `jsfunc(create_proxy(obj))`.
 ```py
 def callback():
     print("clicked!")
+import pyodide
```
rth, line 624:
```suggestion
import pyodide

def callback():
    print("clicked!")
```

docs/usage/type-conversions.md
```diff
@@ -674,7 +675,7 @@ correct due to `PyProxy` usage constraints.
 
 ```pyodide
 function callback(arg){
-    let res = arg.result();
+    var res = arg.result();
```
rth, line 678: Reverting since we can't reproduce this issue,

```suggestion
    let res = arg.result();
```

rth (approved):
Thanks, @wisehackermonkey !

rth:
Actually, this PR should have been made to main not stable, but I can't easily change the target branch now without causing conflicts. Merging to stable, and I'll try to backport it. Thanks.

wisehackermonkey:
thanks!

PR #2063, opened 2021-12-22
MAINT Invoke pywasmcross via function calls

hoodmane:
I adjusted it so that instead of using `subprocess` to invoke `pywasmcross`, we just make function calls. 

There is a valid use case for directly invoking `buildpkg` from the command line so I think it is good that `buildall` uses `subprocess.call` and then it's possible to directly build a single package (since #2053). However, the arguments to `pywasmcross` frequently contain options with a bunch of newlines, making it impractical to invoke directly from the command line in any case. Instead, extra options should be added to `buildpkg` to control which steps are applied.

As a side effect of these changes, we get further clarity on what is happening with environment variables.

### DIFF
--- a/pyodide-build/pyodide_build/buildpkg.py
+++ b/pyodide-build/pyodide_build/buildpkg.py
@@ -19,9 +19,11 @@ from urllib import request
 import fnmatch
 from contextlib import contextmanager
 
+from . import pywasmcross
+
 
 @contextmanager
-def chdir(new_dir):
+def chdir(new_dir: str):
     orig_dir = Path.cwd()
     try:
         os.chdir(new_dir)
@@ -300,6 +302,27 @@ def patch(pkg_root: Path, srcpath: Path, src_metadata: Dict[str, Any]):
         fd.write(b"\n")
 
 
+def install_for_distribution():
+    commands = [
+        sys.executable,
+        "setup.py",
+        "install",
+        "--skip-build",
+        "--prefix=install",
+        "--old-and-unmanageable",
+    ]
+    try:
+        subprocess.check_call(commands)
+    except Exception:
+        print(
+            f'Warning: {" ".join(str(arg) for arg in commands)} failed '
+            f"with distutils, possibly due to the use of distutils "
+            f"that does not support the --old-and-unmanageable "
+            "argument. Re-trying the install without this argument."
+        )
+        subprocess.check_call(commands[:-1])
+
+
 def compile(
     pkg_root: Path,
     srcpath: Path,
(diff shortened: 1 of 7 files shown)


pyodide-build/pyodide_build/buildpkg.py
```diff
@@ -19,9 +19,11 @@
 import fnmatch
 from contextlib import contextmanager
 
+from . import pywasmcross
+
 
 @contextmanager
-def chdir(new_dir):
+def chdir(new_dir: "os.PathLike[str]"):
```
rth, line 26: What would be the difference between os.PathLike[str] and pathlib.Path? It's an abstract class and therefore more general?  Also I think it should work without the quotes?

pyodide-build/pyodide_build/pywasmcross.py
```diff
@@ -136,24 +141,25 @@ def capture_make_command_wrapper_symlinks(env: Dict[str, str]):
         env[var] = symlink
 
 
-def capture_compile(args: argparse.Namespace):
+def capture_compile(*, host_install_dir: str, skip_host: bool, env: Dict[str, str]):
     TOOLSDIR = Path(common.get_make_flag("TOOLSDIR"))
-    env = dict(os.environ)
-    capture_make_command_wrapper_symlinks(env)
+    env["PYODIDE"] = "1"
```
rth, line 146: I think it might be better to make a copy of the env object before making any changes to it, similarly to what was done before?

pyodide-build/pyodide_build/pywasmcross.py
```diff
@@ -404,13 +416,15 @@ def replay_command_generate_args(
...
+    elif line[0] == "c++" or line[0] == "g++":
         new_args = ["em++"]
-    else:
+    elif line[0] == "cc" or line[0] == "gcc":
         new_args = ["emcc"]
         # distutils doesn't use the c++ compiler when compiling c++ <sigh>
         if any(arg.endswith((".cpp", ".cc")) for arg in line):
             new_args = ["em++"]
+    else:
+        assert False, f"Unexpected command {line[0]}"
```
rth, line 427: raise AssertionError(...)

rth (commented):
A few minor comments, otherwise LGTM.

pyodide-build/pyodide_build/buildpkg.py
hoodmane: The point is that it should accept either a string or `pathlib.Path`. Without the quotes I think it passed mypy but failed when run normally.

pyodide-build/pyodide_build/pywasmcross.py
hoodmane: Here we are getting `bash_runner.env`, so changing the environment in place means it applies to the post script. I think it would be a good idea to move the `PYODIDE=1` out into `get_bash_runner` in buildpkg. The path modification should indeed happen to a copied version -- if people invoke the compiler in the `post` script, it would be an unwelcome surprise to be invoking the fake compiler.

pyodide-build/pyodide_build/buildpkg.py
hoodmane: It does look like we only ever apply it to `pathlib.Path` though so maybe using that type would be better.

hoodmane:
Thanks for the review! Will open a followup PR with these fixes (and probably some other stuff).

PR #2064, opened 2021-12-23
Make buildall print list of in progress packages

hoodmane:


ryanking13:
https://github.com/pyodide/pyodide/blob/e8dff87985f1b4ae3b70f4ff0bd419afe0873951/pyodide-build/pyodide_build/buildall.py#L298-L301

This is nice, I was always confused to see `~~ built A.` even when the build was failed.

hoodmane:
It would also be nice to have a third message for when it decided it had nothing to do because the package was already built.

rth:
Thanks!

PR #2065, opened 2021-12-23
Update to Scipy 1.7.3

### LINKED ISSUE #549: Update scipy version
Currently we are using scipy 0.17 from 2016 because [later versions could include f90](https://github.com/scipy/scipy/issues/2829#issuecomment-223764054) and we don't know how to compile it.

However, I quickly looked at scipy git log, and the mean edit frequency of for fortran files in scipy is 6 years, so even if f90 is now accepted now, there was no active modernization effort from f77. Which means that later scipy versions could potentially be built, and we could update it. Latest version would be ideal so that we could contribute fixes upstream if needed.

Though the problem is that until dynamic linking is not resolved https://github.com/iodide-project/pyodide/issues/240, if newer versions add additional modules that use LAPACK it will increase the binary size which is not desirable.

hoodmane:
Resolves #549.


hoodmane:
The problem seems to be that it expects LAPACK to have a function called `sgeqrfp` but our LAPACK doesn't (we have `sgeqrf` without the p...).
I wonder if it's possible to just delete sgeqrfp from the various wrapper generation files.

hoodmane:
Okay the problem here is related to incompatibilities between LAPACK versions. It is a bit hard to tell but it seems like `sgeqrfp` was added in lapack v3.3.0 http://www.netlib.org/lapack/lapack-3.3.0.html, whereas we are using CLAPACK 3.2.1
It seems like it might become an issue for us that CLAPACK is not keeping up with LAPACK releases. LAPACK is up to version 3.10 released a year ago, but CLAPACK is on 3.2 which was released in 2008.

hoodmane:
To find exports from `clapack_all.so`
```sh
wasm-objdump clapack_all.so -d | sed -n 's/.*func.*<\(.*\)>:/\1/p' | sort -u > clapack_exports.txt
```
To find what scipy tries to import from clapack:
```sh
~/wabt-1.0.24/bin/wasm-objdump _flapack.so -d | sed -E -n 's/.*global.get [0-9]* <([a-z]*_?)>/\1/p' | sort -u > symbols.txt
```
there are two unrelated symbols in this second list. Anyways, I take the difference of these two sets and find the missing LAPACK functions:
```
cgemqrt cgeqrfp cgeqrt cuncsd csyconv csyconvf csyconvf_rook ctpmqrt ctpqrt 
dgemqrt dgeqrfp dgeqrt dorcsd dsyconv dsyconvf dsyconvf_rook dtpmqrt dtpqrt 
sgemqrt sgeqrfp sgeqrt sorcsd ssyconv ssyconvf ssyconvf_rook stpmqrt stpqrt
zgemqrt zgeqrfp zgeqrt zuncsd zsyconv zsyconvf zsyconvf_rook ztpmqrt ztpqrt
```
These can be copied from the LAPACK tree into our source tree and then I append them to the end of `scipy/_build_utils/src/wrap_dummy_g77_abi.f`. However,  `cuncsd`, `dorcsd`, `sorcsd` and `zuncsd` are not fortran 77 compliant, so these need to be stubbed out with empty definitions. I was unable to figure out even how to stub them in fortran so I stub them in the generated C.

hoodmane:
Build succeeded and tests pass:
```
scipy/test_scipy.py::scipy_linalg[firefox] PASSED
scipy/test_scipy.py::brentq[firefox] PASSED
scipy/test_scipy.py::scipy_linalg[chrome] PASSED
scipy/test_scipy.py::brentq[chrome] PASSED
scipy/test_scipy.py::scipy_linalg[numpy] PASSED
scipy/test_scipy.py::brentq[numpy] PASSED
```
but
```
FAILED test_packages_common.py::import[firefox-scipy]
FAILED test_packages_common.py::import[chrome-scipy]
FAILED test_packages_common.py::import[node-scipy]
```

hoodmane:
The failing import is `scipy.integerate`:
```py
>>> import scipy.integrate
  File "/lib/python3.9/site-packages/scipy/integrate/__init__.py", line 93, in <module>
    from ._ode import *
  File "/lib/python3.9/site-packages/scipy/integrate/_ode.py", line 95, in <module>
    _dop_int_dtype = _dop.types.intvar.dtype
AttributeError: 'NoneType' object has no attribute 'types'
```

hoodmane:
I don't understand why scipy doesn't detect as installed anymore.

hoodmane:
I think this is ready to go.

rth:
Wow, it's a nice Christmas present @hoodmane  !  It's fantastic that you managed to update scipy. I'll review it on Monday.

docs/project/changelog.md
```diff
@@ -174,6 +174,9 @@ async function runPythonAsync(code, globals) {
 - {{ Enhancement }} upgraded numpy to 1.21.4
   {pr}`1934`
 
+- {{ Enhancement }} Upgraded scipy to 1.6.1
```
leopsidom, line 177: Great work! nit: I think you meant 1.7.3 here ?

hoodmane: Well I wrote the entry before I switched to 1.7.3. Thanks for pointing it out.

packages/scipy/patches/remove-cuncsd-dorcsd-sorcsd-zuncsd.patch
```diff
@@ -0,0 +1,35 @@
+From 7b56e1341c486422973dfa890158872200527b51 Mon Sep 17 00:00:00 2001
+From: Hood Chatham <roberthoodchatham@gmail.com>
+Date: Sat, 25 Dec 2021 20:31:40 -0800
+Subject: [PATCH] Remove cuncsd, dorcsd, sorcsd, zuncsd
+
+These functions don't work because their definitions involve code that isn't 
+fortran 77 compliant. In order to get _flapack.so to link correctly, we need
+to provide stub definitions for these functions, but they don't work. Remove
+them to make sure people get an `AttributeError` if they try to use them 
+rather than just having weird buggy code.
```
rth, line 10: Sounds good.

packages/scipy/patches/rename-_page_trend_test.patch
```diff
@@ -0,0 +1,34 @@
+From 6ffb169dbe331627d70da3f79f928057f1a4e163 Mon Sep 17 00:00:00 2001
+From: Hood Chatham <roberthoodchatham@gmail.com>
+Date: Sun, 26 Dec 2021 07:34:40 -0800
+Subject: [PATCH] Rename _page_trend_test.py to prevent test unvendoring
+
+unvendor_tests will unvendor any file that ends in _test.py.
```
rth, line 6: yes, that's unfortunate. Layer, we should probably come up with some way to add exceptions to that code..

pyodide-build/pyodide_build/_f2c_fixes.py
```diff
@@ -17,157 +20,219 @@ def fix_f2c_clapack_calls(f2c_output_name: str):
...
+
+    code = "".join(lines)
+    for cur_name in lapack_names:
+        code = re.sub(rf"\b{cur_name}\b", "w" + cur_name, code)
+
+    with open(f2c_output, "w") as f:
+        f.write(code)
+
+
+def remove_ftnlen_args(lines: List[str]) -> List[str]:
```
rth, line 125: Any chance you could add doctests for the following functions, which would serve as examples of what is expected to happen?

packages/scipy/info.md
```diff
@@ -0,0 +1,95 @@
...
+scripts. Of the 36 missing symbol, 32 of them are written in Fortran 77 and work
+fine. The remaining 4 need removing but I couldn't figure out how to take them
+out in any sensible way (I can't understand all the layers of codegen that
+happen in scipy), so I give them do-nothing definitions, and then in the build
+script remove lines containing them with `sed`.
+
+Another big problem with f2c is that it cannot handle implicit casts of function
+arguments, because it tries to guess the types of the arguments of the function
+being called based on the types of the arguments at the call site. There are two
+distinct versions of this:
```
rth, line 43: Thanks for the summary, I'll respond to some of these in https://github.com/scipy/scipy/issues/15290

packages/scipy/meta.yaml
```diff
@@ -1,60 +1,92 @@
...
+  # Scipy has a bunch of custom logic implemented in
+  # pyodide-build/pyodide_build/_f2c_fixes.py.
   script: |
-    # Some fixes were extracted from patches/make-int-return-values.patch as regex.
+    pip install scipy==1.7.3 pybind11[global] pythran
+    pip install --upgrade pybind11[global] pythran -t $PYODIDE_ROOT/packages/.artifacts/
+    # We get linker errors because the following 36 functions are missing
+    # Copying them from a more recent LAPACK seems to work fine.
+    wget https://github.com/Reference-LAPACK/lapack/archive/refs/tags/v3.10.0.tar.gz
+    tar -xnf v3.10.0.tar.gz
```
rth, line 59:
```suggestion
    tar xzf v3.10.0.tar.gz
```
? Why the `-n` (seekable) option?  No need to change, I'm just curious.

rth (commented):
Thanks for doing this, and for writing detailed documentation @hoodmane ! It's a huge amount of work and the big step for scipy (and scipy dependent packages) in Pyodide.

A few minor comments are below, and I'll also comment in https://github.com/scipy/scipy/issues/15290

One issue though is that the size appears to have roughly doubled, I'm not sure if it's because scipy grew (that seems still too much though) or if we are including something that we shouldn't have. 
In any case that would be a noticeable regression for users in the context of Pyodide (need to check the size after compression as well). If we can't resolve it easily I wonder if we shouldn't merge this just after the release and spend a bit more time on investigating those aspects.

Also chrome CI is still failing

.circleci/config.yml
```diff
@@ -202,6 +202,7 @@ jobs:
             # Set mtime for EM_CONFIG to avoid ccache cache misses
             touch -m -d '1 Jan 2021 12:00' emsdk/emsdk/.emscripten
 
+            mkdir -p packages/.artifacts/
```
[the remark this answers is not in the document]
hoodmane: No I don't think we need to create it.

packages/scipy/meta.yaml
hoodmane: I don't actually know what these things do, I just typed some random letters and it didn't visibly screw up so I moved on to something else.

.circleci/config.yml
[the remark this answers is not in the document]
hoodmane: Will remove this.

packages/scipy/meta.yaml
```diff
@@ -1,60 +1,92 @@
 package:
   name: scipy
-  version: 0.17.1
+  version: 1.7.3
```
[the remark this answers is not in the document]
hoodmane: It would also be nice to get rid of the second install. There is an issue that downstream packages that use cython don't see our modified cython headers. This is why I currently have to patch scikit-learn to cythonize the files with an include path pointing to our modified scipy. This worked fine before but now it just doesn't seem to install our modified scipy the right way. We could always set `skip_host=False` but that seems pretty expensive.

hoodmane:
Size breakdown by submodule:
```
              0.7      1.7
_build_utils  16K      44K
_lib          72K     292K
cluster      380K     652K
constants    148K     188K
fft            0K     900K
fftpack      920K     204K
integrate    508K    1700K
interpolate 2900K    3400K
io           540K     552K
linalg      6000K   11000K
misc        2100K    2200K
ndimage      568K     652K
odr          600K     588K
optimize	1800K    4700K
signal       764K    1500K
sparse      6000K    6800K
spatial     1300K    2500K
special     4800K   13000K
stats       1020K    7000K

total      30000K   56000K
```

hoodmane:
For the record, there was no size change in linux wheels from 0.17.1 to 1.7.3: the manylinux x86_64 wheel went from 39.3 megabytes to 39.8 megabytes.

hoodmane:
Looking just at `_flapack.so` which increased its size from 2.0M to 4.3M, we see that the data segment doubled, an increase of around 400kb and the debug segments increased by a total 1900kb. Probably we should figure out how to get rid of all of those debug segments.

| Type       | size       | info               | info2   |
| -------- | ------ | --------------- | --- |
| Custom   | 15     | "dylink"        |     |
| Type     | 479    | count:          | 36  |
| Import   | 7145   | count:          | 344 |
| Function | 292    | count:          | 290 |
| Global   | 49     | count:          | 8   |
| Export   | 495    | count:          | 24  |
| Start    | 1      | start:          | 100 |
| Elem     | 519    | count:          | 1   |
| Code     | 400164 | count:          | 290 |
| Data     | 337212 | count:          | 1   |
| Custom   | 12301  | "name"          |     |
| Custom   | 312630 | ".debug_info"   |     |
| Custom   | 497651 | ".debug_loc"    |     |
| Custom   | 6622   | ".debug_ranges" |     |
| Custom   | 2126   | ".debug_abbrev" |     |
| Custom   | 395905 | ".debug_line"   |     |
| Custom   | 32796  | ".debug_str"    |     |

| Type       | size       | info               | info2   |
| -------- | ------- | --------------- | --- |
| Custom   | 15      | "dylink"        |     |
| Type     | 577     | count:          | 40  |
| Import   | 12220   | count:          | 598 |
| Function | 626     | count:          | 624 |
| Global   | 49      | count:          | 8   |
| Export   | 568     | count:          | 27  |
| Start    | 1       | start:          | 105 |
| Elem     | 1197    | count:          | 1   |
| Code     | 863494  | count:          | 624 |
| Data     | 762720  | count:          | 1   |
| Custom   | 24697   | "name"          |     |
| Custom   | 719839  | ".debug_info"   |     |
| Custom   | 1124800 | ".debug_loc"    |     |
| Custom   | 9790    | ".debug_ranges" |     |
| Custom   | 2282    | ".debug_abbrev" |     |
| Custom   | 858820  | ".debug_line"   |     |
| Custom   | 54961   | ".debug_str"    |     |

hoodmane:
I think we were even building with debug symbols with scipy 0.17...

hoodmane:
Built with `-g0` scipy is down to 21 megabytes.

hoodmane:
Okay, now we are up 3 megabytes from 18 megabytes before to 21 megabytes after. That's much more acceptable. Though there is a question why we didn't always have this problem with `-g`.

rgommers:
In case it matters: the `numpy.distutils` build is quite difficult to change in terms of compile flags, and it includes debug symbols by default. So what we do for wheels is to strip them after building: https://github.com/matthew-brett/multibuild/blob/1792851/configure_build.sh#L37.

rth:
> Built with -g0 scipy is down to 21 megabytes.

That's great!

The only remaining issues are fairly consistent error in Chrome,
```
ImportError: Could not load dynamic lib: /lib/python3.9/site-packages/scipy/linalg/_flapack.so
RangeError: WebAssembly.Compile is disallowed on the main thread, if the buffer size is larger than 4KB. Use WebAssembly.compile, or compile on a worker thread.
```
Couldn't this be a sign that the pre-load plugin is not working as expected for Chrome?

hoodmane:
> The only remaining issues are fairly consistent error in Chrome,

Works fine running in chromium 90.0.4430.0 and in chrome 96.0.4664.93 on my local setup, also `pytest packages/scipy/test_scipy.py` works fine on my local setup. Maybe it's running out of memory on the CI server or something? I don't really know how to interpret failures like this.

hoodmane:
Okay, all tests pass now!

rth:
>> Fix dynamic linking errors with _flapack.so
>
> Okay, all tests pass now!

 Perfect, thanks again for working on this!  +1 for merging once CI pass. It's strange that it was only failing in Chrome.

hoodmane:
>  It's strange that it was only failing in Chrome.

Yeah it's very strange. And only in chrome in CI.  By rerunning with ssh, I was able to replicate it and read the selenium logs which gave clear diagnostic information. I was able to replicate the bug in Firefox on the CI server via ssh with just the right setup, but it was generally very sensitive to tiny changes.

hoodmane:
Also, it would be great if we could run the full test suite for scipy once in a while.

hoodmane:
There's a pretty significant number of fatal errors and test failures in the test suite. Currently it makes it about 20% through. I'm running it with the following code:
```js
pyodide._module.packages["scipy-tests"] = {"name":"scipy-tests","version":"0.17.1","depends":["scipy"],"imports":[]}
await pyodide.loadPackage(["scipy-tests", "nose","pytest"]);
pyodide.runPython(`
from scipy import test
test(extra_argv=["--continue-on-collection-errors", "-k", "not multithreading and not TestNQuad", '-vv'])
`);
```
Currently I am running into trouble that `scipy.linalg` has a large number of functions that expect `x, &offx` and are called like `x+offx`. There's also a noticeable fraction of functions that don't fail but return completely wrong answers.

Perhaps it would be a good idea to merge this and fix more of these problems in followup PRs.

rth:
> Currently it makes it about 20% through the test suite before hitting a fatal error. 

That's already better than on main, where I think there is a fatal error after 3% of the test suite. So what I had to do back in 2018  to work around it was to run submodules one by one, e.g.
```
import pytest
pytest.main(["--pyargs", "scipy.linalg", "--continue-on-collection-errors", "-k", "not multithreading and not TestNQuad", '-vv'])
```
though it's very time-consuming to exclude the few tests that produce a fatal error. An ideal situation would have been to have something like pytest-xdist that runs tests in a webworker https://github.com/pytest-dev/pytest-xdist/issues/336 so that if there is a fatal error in a worker, it can mark the test as crashed, spawn a new worker and continue on.  But well, that's another side project  )  

Something like this would be necessary to reliably run the test suite in CI until the end as well. https://github.com/pyodide/pyodide/issues/69

> Currently it makes it about 20% through the test suite before hitting a fatal error.

What was the test success rate roughly for those 20% ?

> There's also a noticeable fraction of functions that don't fail but return completely wrong answers.

That's worrisome. Do you have the pytest logs -- in which submodules does this happen? 

> Perhaps it would be a good idea to merge this and fix more of these problems in followup PRs.

I agree there are already many things in this PR and it would be better to work on the test suite in a follow-up PR. The question is should we include it in the 0.19 release or not. If we do, at least the release notes should mention that scipy still has a number of known issues.

hoodmane:
> What was the test success rate roughly for those 20%?

So far I fixed `integrate` and `interpolate` which used to have fatal errors. The current status of the test suite is:


|module | failed | passed | % |
|-----------|--------:|------------:|-----:|
cluster | 0 | 134 | 100%
constants| 0 | 10 | 100%
fft | 0 | 4961 | 100%
fftpack | 0 | 553  | 100%
integrate | 7 | 157  | 96%
interpolate |  2 | 605  |  99.6%
io | 20  |  301 | 93.8%
linalg |  fatal error | 1% in
misc | 0 | 8 | 100%
ndimage| 5 | 3563 | 99.9%
odr |  fatal error | 11% in
optimize | 26 | 1659 |  98.5%
signal |  fatal error | 11% in | 
spatial  |  fatal error | 2% in | 
special | fatal error | 29% in
stats | fatal error | 1% in

hoodmane:
> The question is should we include it in the 0.19 release or not. If we do, at least the release notes should mention that scipy still has a number of known issues.

I think it's good to include in the 0.19 release, especially if the old test suite had a similar number of failures. We could release fixes in 0.19.1 for instance.

rth (approved):
Sounds good, let's do that. Thanks @hoodmane !

docs/project/changelog.md
```diff
@@ -181,7 +181,16 @@ async function runPythonAsync(code, globals) {
 - {{ Enhancement }} upgraded numpy to 1.21.4
   {pr}`1934`
 
-- {{ Fix }} The built-in pwd module of Python, which provides Unix specific
+- {{ Enhancement }} Upgraded SciPy to 1.7.3
```
rth, line 184:
```suggestion
- {{ Enhancement }} Upgraded SciPy to 1.7.3. Note that there are still known issues with 
  some SciPy components. 
```

docs/project/changelog.md
```diff
@@ -181,7 +181,16 @@ async function runPythonAsync(code, globals) {
...
+  {pr}`2065`
+
+- {{ Enhancement }} Upgraded scikit-learn to version 1.0.2
+  {pr}`2065`
+
+- {{ Enhancement }} Added threadpoolctl (a dependency of scikit-learn)
```
rth, line 190: There is  a "New packages" list below where this should be added.

rth:
Merging, as I would like to see if we can update scikit-image with this https://github.com/pyodide/pyodide/pull/2005. For the changelog, we would need to review it in any case before the release. Thanks.

PR #2066, opened 2021-12-25
Update python-sat

alexeyignatiev:
Hi folks,

This PR bumps the version of the `python-sat` package to the currently most up-to-date one. All the major changes are meant to be brought with the vanilla `python-sat` archive. There are also some cosmetic changes in the patches involved here but these are made only to support the compilation of the latest version of the package.

Thanks,
Alexey

### DIFF
--- a/packages/python-sat/patches/force_malloc.patch
+++ b/packages/python-sat/patches/force_malloc.patch
@@ -1,5 +1,5 @@
 diff --git a/cardenc/pycard.cc b/cardenc/pycard.cc
-index 4ae27fe..5941d0a 100644
+index c4d2ead..a0a30db 100644
 --- a/cardenc/pycard.cc
 +++ b/cardenc/pycard.cc
 @@ -11,6 +11,7 @@
@@ -8,12 +8,12 @@ index 4ae27fe..5941d0a 100644
  #include <stdio.h>
 +#include <stdlib.h>
  #include <Python.h>
-
+ 
  #include "card.hh"
 @@ -122,8 +123,14 @@ static struct PyModuleDef module_def = {
  	NULL,              /* m_free */
  };
-
+ 
 +static volatile int *_dummy_malloc;
 +
  PyMODINIT_FUNC PyInit_pycard(void)
@@ -23,10 +23,10 @@ index 4ae27fe..5941d0a 100644
 +	free((void *)_dummy_malloc);
 +
  	PyObject *m = PyModule_Create(&module_def);
-
+ 
  	if (m == NULL)
 diff --git a/solvers/pysolvers.cc b/solvers/pysolvers.cc
-index 7066cb4..6b4d7de 100644
+index a198f43..d45639b 100644
 --- a/solvers/pysolvers.cc
 +++ b/solvers/pysolvers.cc
 @@ -15,6 +15,7 @@
@@ -35,12 +35,12 @@ index 7066cb4..6b4d7de 100644
  #include <stdio.h>
 +#include <stdlib.h>
  #include <vector>
-
+ 
  #ifdef WITH_CADICAL
-@@ -493,8 +494,14 @@ static struct PyModuleDef module_def = {
+@@ -643,8 +644,14 @@ static struct PyModuleDef module_def = {
  	NULL,              /* m_free */
  };
-
+ 
 +static volatile int *_dummy_malloc;
 +
  PyMODINIT_FUNC PyInit_pysolvers(void)
@@ -50,5 +50,5 @@ index 7066cb4..6b4d7de 100644
 +	free((void *)_dummy_malloc);
 +
  	PyObject *m = PyModule_Create(&module_def);
-
+ 
  	if (m == NULL)
--- a/packages/python-sat/patches/proper_build.patch
+++ b/packages/python-sat/patches/proper_build.patch
@@ -1,20 +1,20 @@
 diff --git a/setup.py b/setup.py
-index 4e810df..d191d6d 100644
+index 82dac9c..3fe102f 100644
 --- a/setup.py
 +++ b/setup.py
-@@ -68,8 +68,8 @@ Details can be found at `https://pysathq.github.io <https://pysathq.github.io>`_
- 
+@@ -69,8 +69,8 @@ Details can be found at `https://pysathq.github.io <https://pysathq.github.io>`_
  # solvers to install
  #==============================================================================
--to_install = ['cadical', 'glucose30', 'glucose41', 'lingeling', 'maplechrono',
--        'maplecm', 'maplesat', 'minicard', 'minisat22', 'minisatgh']
-+to_install = ['cadical', 'glucose30', 'glucose41', 'maplechrono', 'maplecm',
-+        'maplesat', 'minicard', 'minisat22', 'minisatgh']
- 
+ to_install = ['cadical', 'gluecard30', 'gluecard41', 'glucose30', 'glucose41',
+-        'lingeling', 'maplechrono', 'maplecm', 'maplesat', 'mergesat3',
+-        'minicard', 'minisat22', 'minisatgh']
++        'maplechrono', 'maplecm', 'maplesat', 'mergesat3', 'minicard',
++        'minisat22', 'minisatgh']
  
  # example scripts to install as standalone executables
-@@ -77,47 +77,9 @@ to_install = ['cadical', 'glucose30', 'glucose41', 'lingeling', 'maplechrono',
- scripts = ['fm', 'genhard', 'lbx', 'lsu', 'mcsls', 'models', 'musx', 'rc2']
+ #==============================================================================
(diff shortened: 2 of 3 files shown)


packages/python-sat/patches/force_malloc.patch
```diff
@@ -1,5 +1,5 @@
 diff --git a/cardenc/pycard.cc b/cardenc/pycard.cc
-index 4ae27fe..5941d0a 100644
+index c4d2ead..a0a30db 100644
```
hoodmane, line 2: Did anything change in this patch? It looks like only whitespace changes? Did the patch not apply cleanly without changing the whitespace?

packages/python-sat/patches/proper_build.patch
```diff
@@ -1,20 +1,20 @@
...
- 
+@@ -69,8 +69,8 @@ Details can be found at `https://pysathq.github.io <https://pysathq.github.io>`_
  # solvers to install
  #==============================================================================
--to_install = ['cadical', 'glucose30', 'glucose41', 'lingeling', 'maplechrono',
--        'maplecm', 'maplesat', 'minicard', 'minisat22', 'minisatgh']
-+to_install = ['cadical', 'glucose30', 'glucose41', 'maplechrono', 'maplecm',
-+        'maplesat', 'minicard', 'minisat22', 'minisatgh']
- 
+ to_install = ['cadical', 'gluecard30', 'gluecard41', 'glucose30', 'glucose41',
```
hoodmane, line 8: Could you add a comment here saying that we are removing the `lingeling` solver? It's a bit hard to tell from looking at the diff. If you know it would also be great to say why (if you don't know, it would be great if you could try putting it back in  and seeing what goes wrong).

hoodmane (approved):
Thanks so much @alexeyignatiev!

packages/python-sat/patches/proper_build.patch
alexeyignatiev: Well, it does not work because all the solvers are currently built with a C++ compiler and with the c++11 flag. Lingeling is a solver written in C and so the compiler fails. I can try to put it back but then the set of building scripts needs to be changed. Let me try.

packages/python-sat/patches/force_malloc.patch
alexeyignatiev: Actually, nothing changed in this patch. I guess you are right that it should not be updated then.

alexeyignatiev: Ah, now I remember why I updated this patch too. The reason is that line numbers in the affected files changed and I thought it would be safer to update the patch accordingly.

alexeyignatiev:
OK, here are some comments to summarise the changes made and the discussion.

1. The package does not support `lingeling`. This is the only solver excluded. In my view, this should be fine as (1) there are still quite a large number of other solvers supported (2) `lingeling` does not offer full incrementality anyway. More on this in point 3.
2. The patch `force_malloc.patch` is updated due to the changes in the line numbers. I guess the older version might still work just fine but I wanted to keep it safe.
3. The patch `proper_build.patch` disables the sophisticated building process for the `pysolvers` extension as of the vanilla version of `python-sat`. Here, instead of compiling each solver into a static library and then linking the extension will all those individual static libraries, the routine compiles each solver source file into an object with `em++ --std=c++11` and hence no static libraries are involved. This breaks compilation of `lingeling` because it is a pure C-written solver. It may help to understand if there are other solutions.

In general, the changes made in the vanilla version of `python-sat` since the previously integrated version `0.1.6.dev6` are the following (copied from the original changelog):

##### 04.12.2021 (*0.1.7.dev15*)

- Fixed a minor bug in stratified RC2 related to model enumeration for empty
  formulas.
- Updated Hitman so that it opts for plain RC2 if all weights are the same.
- Fixed cost MUS computation in OptUx.

##### 24.11.2021 (*0.1.7.dev14*)

- Another attempt to fix the bug in topw handling if unspecified.

##### 24.11.2021 (*0.1.7.dev13*)

- Fixed a minor issue in fm.py related to handling native AM1 constraints.
- Fixed a bug in WCNF parser, which caused incorrect top weight computation if
  not specified in the preamble of the file explicitly.

##### 09.11.2021 (*0.1.7.dev12*)

- Fixed a bug related to handling CNF formulas and improper updating IDPool
  for PB constraints in the case of no auxiliary variables.

##### 12.10.2021 (*0.1.7.dev11*)

- Fixed a bug related to meaningless bounds in cardinality constraints (both
  AtMost and AtLeast).

##### 17.08.2021 (*0.1.7.dev10*)

- Added cluster-based stratification to `RC2`.
- Made the selection of stratification strategy in `RC2` optional.

##### 02.08.2021 (*0.1.7.dev9*)

- Added proper handling of border cases when creating cardinality constraints.
  Done for all encodings, and both AtMostK and AtLeastK constraints.
- Added an ``UnsupportedBound`` exception to be raised whenever *pairwise*,
  *bitwise*, or *ladder* encodings are created with the bound in `(1, N - 1)`.

##### 30.07.2021 (*0.1.7.dev8*)

- Fixed a bug in `IDPool` introduced in `0.1.7.dev7`, related to
  ``None`` handling.

##### 28.07.2021 (*0.1.7.dev7*)

- Updated `IDPool` to return a new variable ID without an object.
- Updated `RC2` to support IDPool instead of `self.topv`.

##### 17.07.2021 (*0.1.7.dev6*)

- Modified the implemention of the am1 heuristic in `RC2`.

##### 26.06.2021 (*0.1.7.dev5*)

- Got rid of relaxation variables in `RC2` - minor issue, which might
  still have affected the performance.

##### 11.06.2021 (*0.1.7.dev4*)

- Fixed compilation issues occurring for cadical on macOS >= 11.3 with the
  most recent version of clang. Thanks to András Salamon.

##### 24.05.2021 (*0.1.7.dev3*)

- Added a warning in `enum_models()` regarding unsatisfiability of the formula
  after model enumeration is finished.
- Fixed a couple of bugs in `FM`.
- Similar changes in other example implementations.

##### 20.04.2021 (*0.1.7.dev2*)

- Improved the sequential counters encoding based on Donald Knuth's
  irredundant variant, following the suggestion of Alex Healy.

##### 31.03.2021 (*0.1.7.dev1*)

- An attempt to fix compilation issues of Mergesat.

##### 31.03.2021 (*0.1.7.dev0*)

- Added basic interface to `Mergesat3`.
- Ported `Minicard`'s native cardinality support to Glucose, which
  resulted in solvers `Gluecard3` and `Gluecard4`.
- Added translation of `Cadical`'s binary DRUP format to text-based.

##### 25.03.2021 (*0.1.6.dev16*)

- A few minor corrections in `RC2`, `RC2Stratified`, and `Hitman` addressing the issues related to empty formulas.

##### 23.03.2021 (*0.1.6.dev15*)

- A minor correction in `OptUx` related to MUS cost value.

##### 23.03.2021 (*0.1.6.dev14*)

- A minor correction in the ``setup.py`` script.

##### 23.03.2021 (*0.1.6.dev13*)

- Added a way to bootstrap `LBX` and `MCSls` for computing
  MCSes that include a specified subset of clauses.
- Updated `Hitman` to support *weighted* hitting set problems.
- Added `OptUx`, which is an smallest/optimal MUS extractor and
  enumerator that aims to replicate the performance of Forqes.

##### 13.02.2021 (*0.1.6.dev12*)

- Fixed formula dumping in other formats (see `to_alien`).

##### 24.11.2020 (*0.1.6.dev11*)

- Added formula dumping in SMT-LIB2 (see `to_alien`).

##### 19.11.2020 (*0.1.6.dev10*)

- Fixed the number of propagations ins MapleChrono.

##### 27.09.2020 (*0.1.6.dev9*)

- Fixed a bug in copying of `CNFPlus` and `WCNFPlus` objects.
- Added an attempt to write down OPB and LP formats (likely to be buggy!).

##### 26.09.2020 (*0.1.6.dev8*)

- Fixed a bug in `solvers.py` related to the `CNFPlus` iterator.

##### 25.09.2020 (*0.1.6.dev7*)

- Added an example for accum_stats().
- Fixed a minor bug in musx example.
- A few fixes in `pysat.formula`.
- Fixed an issue in `pysat.card` related to non-list literals.

hoodmane:
Did you close this by accident?

alexeyignatiev:
Yes, I did. :(

hoodmane:
Okay. Will merge when tests pass.

alexeyignatiev:
Thank you, @hoodmane!

hoodmane:
Would you add a comment to the patch saying that it patches out lingeling, that it's more difficult to compile because it's the only one in C, and it could be fixed in principal but it's more effort than it's worth?

alexeyignatiev:
I presume you meant a comment in the top section of the patch. I tried to add one now. Please check.

hoodmane:
Looks good thanks.

PR #2069, opened 2021-12-28
ENH improvements to pyodide-build

hoodmane:
### Description

More improvements to pyodide-build. This includes:
1. some tuneups to pywasmcross extracted from #2065
2. a `--continue` flag that allows skipping parts of the package build in `buildpkg` for debugging the build
3. better handling of skipping already built packages in buildall -- in particular, it will now rebuild a package if a package that it depends on is being rebuilt, and the messages printed are a bit clearer.
4. a `--force` flag to force a rebuild

### Checklists

- [x] Add / update tests
- [x] Update documentation

### DIFF
--- a/pyodide-build/pyodide_build/buildall.py
+++ b/pyodide-build/pyodide_build/buildall.py
@@ -252,6 +252,14 @@ def get_progress_line(package_set):
 
 
 def format_name_list(l):
+    """
+    >>> format_name_list(["regex"])
+    'regex'
+    >>> format_name_list(["regex", "parso"])
+    'regex and parso'
+    >>> format_name_list(["regex", "parso", "jedi"])
+    'regex, parso, and jedi'
+    """
     if len(l) == 1:
         return l
     most = l[:-1]
--- a/pyodide-build/pyodide_build/buildall.py
+++ b/pyodide-build/pyodide_build/buildall.py
@@ -251,7 +251,7 @@ def get_progress_line(package_set):
     return f"In progress: " + ", ".join(package_set.keys())
 
 
-def format_name_list(l : List[str]) -> str:
+def format_name_list(l: List[str]) -> str:
     """
     >>> format_name_list(["regex"])
     'regex'
--- a/pyodide-build/pyodide_build/buildall.py
+++ b/pyodide-build/pyodide_build/buildall.py
@@ -280,11 +280,12 @@ def mark_package_needs_build(
         if dep not in needs_build:
             mark_package_needs_build(pkg_map, pkg_map[dep], needs_build)
 
+
 def generate_needs_build_set(pkg_map):
     """
-    Generate the set of packages that need to be rebuilt. 
+    Generate the set of packages that need to be rebuilt.
 
-    This consists of: 
+    This consists of:
     1. packages whose source files have changed since they were last built
        according to needs_rebuild, and
     2. packages which depend on case 1 packages.
@@ -295,6 +296,7 @@ def generate_needs_build_set(pkg_map):
         if needs_rebuild(pkg.pkgdir, pkg.pkgdir / "build", pkg.meta):
             mark_package_needs_build(pkg_map, pkg, needs_build)
     return needs_build
+
 
 def build_from_graph(pkg_map: Dict[str, BasePackage], outputdir: Path, args) -> None:
     """
@@ -324,7 +326,6 @@ def build_from_graph(pkg_map: Dict[str, BasePackage], outputdir: Path, args) -> 
         already_built = set()
     else:
         needs_build = generate_needs_build_set(pkg_map)
-
 
     # We won't rebuild the complement of the packages that we will build.
     already_built = set(pkg_map.keys()).difference(needs_build)
--- a/pyodide-build/pyodide_build/buildall.py
+++ b/pyodide-build/pyodide_build/buildall.py
@@ -326,7 +326,7 @@ def build_from_graph(pkg_map: Dict[str, BasePackage], outputdir: Path, args) -> 
     t0 = perf_counter()
     for pkg_name in needs_build:
         pkg = pkg_map[pkg_name]
-        if len(pkg.dependencies) == 0:
+        if len(pkg.unbuilt_dependencies) == 0:
             build_queue.put((job_priority(pkg), pkg))
 
     built_queue: Queue = Queue()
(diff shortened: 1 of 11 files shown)


pyodide-build/pyodide_build/buildall.py
```diff
@@ -242,6 +251,23 @@ def get_progress_line(package_set):
...
+    return " ".join(most) + " and " + l[-1]
+
+
+def mark_package_needs_build(
+    pkg_map: Dict[str, BasePackage], pkg: BasePackage, needs_build: Set[str]
+):
+    needs_build.add(pkg.name)
+    for dep in pkg.dependents:
+        mark_package_needs_build(pkg_map, pkg_map[dep], needs_build)
+
```
rth, line 269: Maybe better to return the needs_build and aggregate the output explicitly, rather than having a recursive function with side effects on  input variables,

```suggestion
    pkg_map: Dict[str, BasePackage], pkg: BasePackage
) -> Set[str]:
    needs_build = set(pkg.name)
    for dep in pkg.dependents:
        needs_build.update(mark_package_needs_build(pkg_map, pkg_map[dep]))
    return needs_build

```

pyodide-build/pyodide_build/buildall.py
```diff
@@ -264,10 +290,43 @@ def build_from_graph(pkg_map: Dict[str, BasePackage], outputdir: Path, args) ->
...
+            # Otherwise, rebuild packages that have been updated and their dependents.
+            if needs_rebuild(pkg.pkgdir, pkg.pkgdir / "build", pkg.meta):
+                mark_package_needs_build(pkg_map, pkg, needs_build)
+
+        # We won't rebuild the complement of the packages that we will build.
+        already_built = set(pkg_map.keys()).difference(needs_build)
+        for pkg_name in already_built:
+            # Hacky way to prevent these from being rebuilt: add a dependency
+            # that won't ever be satisfied
+            pkg_map[pkg_name].unbuilt_dependencies.add("!don't build this")
```
rth, line 310: That's a bit ugly, can't we rather remove it from needs_build ?

pyodide-build/pyodide_build/buildall.py
```diff
@@ -470,6 +529,13 @@ def make_parser(parser):
         default=None,
         help=("Only build the specified packages, provided as a comma-separated list"),
     )
+    parser.add_argument(
+        "--force",
```
rth, line 533: Maybe,
```suggestion
        "--rebuild",
```

to be more explicit ? `buildall --force` is a bit more ambiguous.
Unless we already use it for a similar use case elsewhere?

pyodide-build/pyodide_build/buildpkg.py
```diff
@@ -636,23 +654,31 @@ def build_package(
...
-            os.makedirs(build_dir)
-
-        srcpath = prepare_source(pkg_root, build_dir, src_path, source_metadata)
-        if build_metadata.get("script"):
-            run_script(build_dir, srcpath, build_metadata, bash_runner)
-        if build_metadata.get("library"):
-            create_packaged_token(build_dir)
-            return
+        if not should_prepare_source:
+            assert srcpath.exists()
```
rth, line 666: Let's raise an AssertionError or a IOError, with a verbose error message that includes the path. Otherwise this will be hard to debug when it fails.

pyodide-build/pyodide_build/buildpkg.py
```diff
@@ -734,36 +794,62 @@ def make_parser(parser: argparse.ArgumentParser):
     return parser
 
 
+def parse_continue_arg(continue_from: Optional[str]) -> Dict[str, Any]:
+    assert (
+        continue_from is None
+        or continue_from == "capture"
+        or re.fullmatch("replay(:[0-9]*)?", continue_from)
+    )
```
rth, line 802: Similarly, we need explicit exceptions, otherwise, this will be undebugable when not run in a debugger.

pyodide-build/pyodide_build/buildpkg.py
```diff
@@ -734,36 +794,62 @@ def make_parser(parser: argparse.ArgumentParser):
...
 
     pkg_root = meta_file.parent
     pkg = parse_package_config(meta_file)
+
+    pkg["source"] = pkg.get("source", {})
+    pkg["build"] = pkg.get("build", {})
+    build_metadata = pkg["build"]
+    build_metadata["cflags"] = build_metadata.get("cflags", "")
+    build_metadata["cxxflags"] = build_metadata.get("cxxflags", "")
+    build_metadata["ldflags"] = build_metadata.get("ldflags", "")
```
rth, line 834: We should probably improve default parameters handling in config  (and config validation) with something like https://pydantic-docs.helpmanual.io/ but this is OK for now.

pyodide-build/pyodide_build/buildpkg.py
```diff
@@ -724,6 +753,37 @@ def make_parser(parser: argparse.ArgumentParser):
...
+                Possible arguments:
+
+                    'capture' : redo capture step and replay step (but don't prepare
+                                sources again, use existing source directory). Good for debug
+                                builds with hand-modified sources. `--continue` with no argument
+                                has the same effect.
+
+                    'replay' : Don't redo the capture step but redo the replay step.
+
+                    'replay:15' : replay the capture step starting with the 15th compile command (any integer works)
```
rth, line 782: Can you mention this in the docs https://pyodide.org/en/stable/development/new-packages.html ? Maybe under some separate section on debugging build failures..

rth (commented):
A few comments, otherwise LGTM. Thanks!

pyodide-build/pyodide_build/buildall.py
hoodmane: Or maybe `--force-rebuild`?

pyodide-build/pyodide_build/buildall.py
hoodmane: This is basically the same implementation as `addPackageToLoad`:
https://github.com/pyodide/pyodide/blob/42180e86816e4db176d4bb770f0a897c600d7438/src/js/load-pyodide.js#L111
I'll try adding a separate function which does the base case of the recursion.

pyodide-build/pyodide_build/buildall.py
hoodmane: Well they already aren't in `needs_build` since we just calculated `already_built` as the complement of `needs_build`. Let me see what happens if I remove this.

hoodmane: Yeah it works fine without this, which makes sense given the logic below.

PR #2073, opened 2022-01-03
Upgrade to statsmodels 0.13.1

rth:
This should be now possible with https://github.com/pyodide/pyodide/pull/2065

Closes https://github.com/pyodide/pyodide/issues/537


rth:
Looks like there is a signature mismatch, probably due to our patches, that needs fixing
```
TypeError: C function scipy.linalg.cython_blas.ccopy has wrong signature (
    expected void (int *, __pyx_t_float_complex *, int *, __pyx_t_float_complex *, int *), 
          got int (int *, __pyx_t_float_complex *, int *, __pyx_t_float_complex *, int *)
)
```
Interestingly this only failed in Firefox and not Chrome.

hoodmane:
> Interestingly this only failed in Firefox and not Chrome.

Yeah it's very confusing to me why so many things that seem like they should be broken always are only broken in one browser.

For some reason, cython doesn't see our patched sources. I fixed a similar problem in scikit-learn with the following patch:
https://github.com/pyodide/pyodide/blob/main/packages/scikit-learn/patches/0001-Cythonize-_cython_blas.pyx-with-include-path-pointin.patch
Probably something similar will work here?

rth:
Yes, makes sense. Tried that, but it's still pulling the wrong header files appartenly.

rth:
> should be broken always are only broken in one browser.

Actually, after looking more carefully here it was just xfailed in Chrome.

packages/test_packages_common.py
```diff
@@ -49,10 +49,10 @@ def registered_packages_meta():
     }
 
 
-UNSUPPORTED_PACKAGES: Dict[str, List[str]] = {
-    "chrome": ["statsmodels"],
+UNSUPPORTED_PACKAGES: dict = {
```
rth, line 52:
```suggestion
UNSUPPORTED_PACKAGES: Dict[str, List[str]] = {
```

should fix CI

hoodmane:
I think we can put back the xfail in chrome and merge this.
Also, line 57 of `test_packages_common` has a funny typo:
```py
if "CI" is os.environ:
```
should read `if "CI" in os.environ:` ...

packages/test_packages_common.py
```diff
@@ -50,9 +50,9 @@ def registered_packages_meta():
 
 
 UNSUPPORTED_PACKAGES: Dict[str, List[str]] = {
-    "chrome": ["statsmodels"],
+    "chrome": [],
     "firefox": [],
-    "node": ["statsmodels"],
+    "node": [],
 }
 if "CI" is os.environ:
```
rth, line 57:
```suggestion
if "CI" is in os.environ:
```

rth:
Sounds good, thanks for catching it.

hoodmane:
Oops forgot a changelog entry...

Issue #2074, opened 2022-01-03
Avoid log overlay when building packages


rth:
https://github.com/pyodide/pyodide/pull/2064 added a list of in-progress packages, which is nice. However this can lead to overlays in the log, cf below
```
[3/20] (thread 3) built patsy in 4.06 simageio, numcodecs
[6/20] (thread 3) building pandasmageio, numcodecs
[4/20] (thread 4) built imageio in 4.35 snumcodecs, pandas
[7/20] (thread 4) building astropymcodecs, pandas
In progress: scipy, matplotlib, numcodecs, pandas, astropy
```

We should probably pad those lines with empty spaces up to the length of the last "In progress" line. Unless it's an issue with CircleCI terminal output, I haven't tried it locally.

@hoodmane

hoodmane:
It looks fine locally on my system. The behavior on the circleci log is a bit inconsistent.
https://github.com/pyodide/pyodide/blob/5309d21371297ae805300275b0fac456b860d65e/pyodide-build/pyodide_build/buildall.py#L237

I try to overwrite any existing stuff with spaces as follows:
```py
twidth = os.get_terminal_size()[0]
print(" " * twidth, end="\r")
```
this *should* blank the line. My suspicion is that `os.get_terminal_size()` is not reporting an accurate size. I think we should set an environment variable to disable it on CI, unless it also doesn't work correctly on other peoples' local setups. Alternatively, we could try to investigate what `os.get_terminal_size()` is returning.

rth:
Makes sense. IMO the best might be do use some CLI library that can deal with it, for instance, https://github.com/willmcgugan/rich as otherwise, I suspect we will keep running into some edge cases depending on the terminal where it's run. Also, it would allow getting nicer formatting in general, particularly for CLI once we get there.

PR #2076, opened 2022-01-04
Use rich for status bar

hoodmane:
As suggested in #2074. Turns out it's very simple. The docker environment needs to be updated to include rich though.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -72,6 +72,11 @@ jobs:
             ccache -s
 
       - run:
+          name: Install rich
+          command: |
+            pip install rich
+
+      - run:
           name: build pyodide core
           no_output_timeout: 1200
           command: |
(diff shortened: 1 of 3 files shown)


hoodmane:
This works very nicely locally, in CI I think the original code without either #2064 or this works best.

rth:
Interesting. It's also more code than I thought it would be. But I suppose it's also more robust and has better UI at least putting aside the CI rendering issues (tough I think this should be a known problem with possible solutions). Let's look back at it after the release.

ryanking13:
Maybe let's add rich to the requirements first  in a seperate PR, instead of putting `pip install rich` everywhere?

ryanking13:
> Maybe let's add rich to the requirements first in a seperate PR, instead of putting `pip install rich` everywhere to make CI pass.

Actually, adding rich to the dependency of `pyodide-build` would be better.

PR #2078, opened 2022-01-05
Add a convenience method to pyfetch for the response headers

bartbroere:
### Description
Having the `pyfetch` convenience wrapper for Javascript's `fetch` is quite nice. I would like to start using `fetch` through `pyfetch` instead of `XMLHttpRequest` in my `requests` shim. Having the headers available using this convenience method would be nice.

One thing I have considered, but not yet implemented, is using a variant of a `CaseInsensitiveDict`. Any thoughts on that?

### Checklists
I don't think any of these changes would warrant a changelog, test or documentation update, but feel free to disagree and I'll add them.


rth:
Thanks @bartbroere ! It would be good to have a test for this though. 
You could add one to https://github.com/pyodide/pyodide/blob/ce2746b690b5f2e93d7e7b12054a5abf67cfdd1b/src/tests/test_pyodide_http.py by adapting the ` test_open_url` there and adding some custom headers (and using pyfetch instead of open_url).

To avoid linting errors the easiest is to use pre-commit.

bartbroere:
Sorry for not configuring the pre-commit actions yet, I think it shows by the manual fix commits. The test has been added though. I chose headers that hopefully will not change any time soon.

src/tests/test_pyodide_http.py
```diff
@@ -58,3 +58,15 @@ async def test_pyfetch_unpack_archive():
...
         "tests",
     ]
+
+
+@run_in_pyodide
+async def test_pyfetch_headers():
+    from pyodide.http import pyfetch
+
+    response = await pyfetch(
+        "http://files.pythonhosted.org/packages/93/8d/e1e98360dc899e533cb3dd857494f2571b129bdffcee76365009b2bb507c/example_pypi_package-0.1.0.tar.gz"
```
rth, line 68: Could you please rather mock this with the `httpserver` fixture as done in test_open_url to avoid HTTP calls to external servers?

bartbroere: Sorry, I forgot about this PR, but I just noticed it had some activity again. I could add a test and use the convenience method, as mentioned in the previous comments.

ryanking13: Hi @bartbroere, I needed this feature and found your PR :) I added a test and will merge after CI pass. Thanks for your work!

rth (approved):
Thanks for finishing it @ryanking13 !

Issue #1973, opened 2021-11-17
Building a Python package that uses Rust


josephrocca:
## 🚀 Feature
Tooling and/or docs to support building Python packages that contain Rust code.

### Motivation
I was in the process of (attempting to) create a JS/wasm build of [Hugging Face Tokenizers](https://github.com/huggingface/tokenizers/) when I realised that it might just be easier/better for my use case to load it with Pyodide (especially since I might want to use Python packages that depend upon Tokenizers. The [docs show](https://github.com/pyodide/pyodide/blob/main/docs/development/new-packages.md) how to build a Python package that uses C (with emscripten), but not Rust.

Rust has great, simple wasm build support via [`wasm-pack`](https://github.com/rustwasm/wasm-pack). After annotating the Rust functions/modules that you want to export (and installing wasm pack with `cargo install wasm-pack`) it's just a single `wasm-pack build --target=web --out-dir=wasm` to get the wasm and glue files. [Here's a hello world tutorial](https://developer.mozilla.org/en-US/docs/WebAssembly/Rust_to_wasm).

So I'm wondering: Is this feasible now with the right configuration, or is there some extra work on the Pyodide dev tooling to get this working? If it's not possible with the current tooling, what might be the steps required to make it possible?

Thanks!

rth:
Thanks for proposing it! There was an earlier analysis of feasibility in https://github.com/pyodide/pyodide/issues/1282#issuecomment-785459008 and it doesn't look very promising. 

Maybe a start could be trying to build a very simple PyO3 Python library. Related https://github.com/PyO3/pyo3/issues/1522

Even if the end we manage to compile that library say for RustPython using the wasm-pack, you end up with a Python extension built for a different Python interpreter and the compatibility story there is not very good.

hoodmane:
I wonder if it could be possible to convince Rust to emit llvm ir and then use emcc to do the backend compilation and linkage. Rust also has a wasm32-emscripten target but it's apparently pretty broken.

josephrocca:
@hoodmane @rth Thanks for your tips! Having looked into this, it seems to be a bit over my head at my current level of knowledge. That said, it would be great if I could make a dent in at least part of the problem.

[...]

PR #2081, opened 2022-01-06
Rust/PyO3 proof of concept

hoodmane:
This is intended as a proof of concept for PyO3. I compiled a fork of [pyo3_samples](https://github.com/saidvandeklundert/pyo3). 
I made no changes to `Cargo.toml` or the package configuration! (Well, other than changing the name to `pyo3_samples`). My fork just removes code that fails to compile due to dynamic linking errors.

Partial work toward #1973.

Limitations:
1. Dynamic linker errors if the code uses `println!` or other I/O stuff (I think `#[derive(Debug)]` by itself is a problem).
2. Code size is HUGE -- 5.4M. A C extension implementing comparable functionality would be somewhere between 100 and 1000 times smaller.
3. Catching panics and wrapping them in Python errors doesn't work due to missing `catch_unwind`-related symbols

I've had some luck fixing `println!` in local experiments. Related to the issues with `println!`, PyO3 seems to generate a lot of issues with `dynCallLegacy` trying to call invoke stubs that don't exist. It would be easier if we could use `-s WASM_BIGINT=1`. Failing that, it might be a good idea to patch `dynCallLegacy` so that side modules don't have to depend on invoke stubs that may or may not be generated in the main module.


rth:
Very cool. I'll have a look after the release.

rth:
I would be very happy to include this as long as Rust doesn't become a mandatory dependency. So even PYODIDE_PACKAGES="*" shouldn't build this package IMO (though that would make the logic a bit more convoluted). Our build setup is complex enough not to ask people to also install Rust.

Maybe we could build this package (and test it)  in a separate GH Action CI job? We have more CI resources there.

packages/pyo3_samples/meta.yaml
```diff
@@ -0,0 +1,28 @@
...
+  version: 0.1.0
+source:
+  # url: https://github.com/hoodmane/pyo3_samples/tarball/main
+  # sha256: 5a402fc5323365cd2fce3379a7a619e6dba28083bf7eed1ec554820c1551dce2
+  path: src
+build:
+  sharedlibrary: true
+  script: |
+    source $HOME/.cargo/env
+    rustup update nightly
```
rth, line 12: Which features do you need nightly for, last stable release doesn't work?

packages/pyo3_samples/pyo3_config.ini
```diff
@@ -0,0 +1,9 @@
+implementation=CPython
+version=3.9
+shared=true
+abi3=false
+lib_name=python3.9
+lib_dir=/home/hood/Documents/programming/pyodide/cpython/installs/python-3.9.5/lib/
```
rth, line 6: It's not used then if it works in CI?

packages/pyo3_samples/src/src/lib.rs
```diff
@@ -0,0 +1,232 @@
+extern crate serde;
```
rth, line 1: Couldn't we load it as a zip from some pinned commit in https://github.com/saidvandeklundert/pyo3 directly?

packages/pyo3_samples/src/src/lib.rs
```diff
@@ -0,0 +1,232 @@
...
+    info!("A log message from {}!", "Rust");
+}
+
+#[pyfunction]
+fn get_fibonacci(number: isize) -> PyResult<u128> {
+    if number == 1 {
+        return Ok(1);
+    } else if number == 2 {
+        return Ok(2);
+    }
```
rth, line 127: Ahh, I wish we did Rust more instead of C, but that's unlikely to happen..

packages/pyo3_samples/pyo3_config.ini
hoodmane: Apparently not. I am pretty unclear on which of these are used -- without this file it definitely doesn't work, but spme of the entries seem to have no effect.

packages/pyo3_samples/src/src/lib.rs
hoodmane: Yeah we can switch to that, though currently I've modified it to remove the broken functions so we'd maybe need a commit on my fork. For experimenting, it's more convenient for it to be in-tree though.

packages/pyo3_samples/meta.yaml
hoodmane: It's needed to build the standard library with `-Z build-std`. Ideally we could figure out how to build the standard library as a separate shared library. I think it's several megabytes big and we will eventually have more than one rust package if we get this working so it'd be better not to duplicate that anyways. But I don't know how to do this.

hoodmane:
I think we should probably not merge this until we have some resolution for https://github.com/rust-lang/rust/issues/93015 because all real world programs of a reasonable size are likely to at some point try to use the missing symbols from the broken object file `compiler_builtins-e79221a6cfe3d845.compiler_builtins.76620b52-cgu.1.rcgu.o` that I am deleting. Though I think currently this deletes all of `libcompiler_builtins-e79221a6cfe3d845.rlib` and if we just delete the one bad object file instead of the whole library, there will probably be fewer problems. I will have to experiment to see which features are broken.

I think all current crashes are due to missing symbols from `compiler_builtins`, so if we resolve https://github.com/rust-lang/rust/issues/93015 then pyo3 packages will function correctly.

hoodmane:
I think this might actually work now. The bad object files problem seems gone, presumably because of changes to Rust, to Emscripten, or both.

PR #2084, opened 2022-01-07
Apply black with updated docker image version

ryanking13:
Our `black` in updated docker image complain about formatting


rth:
Thanks @ryanking13 !

PR #2085, opened 2022-01-07
Run pyodide-build update-all

hoodmane:
I also ran prettier on all meta.yaml files.


hoodmane:
zarr is now the only pure Python package with patches.

.circleci/config.yml
```diff
@@ -284,7 +284,7 @@ jobs:
               --verbose \
               -k 'not (chrome or firefox or node)' \
               --cov=pyodide_build --cov=pyodide \
-              src pyodide-build packages/micropip/
+              src/tests/ pyodide-build packages/micropip/
```
rth, line 287: There might be docstests in `py` that we would want to run as well I think? Generally, it's not a bad idea to run pytest on src/py. If you want to exclude things, better to do it in https://github.com/pyodide/pyodide/blob/main/setup.cfg#L9

packages/typing-extensions/meta.yaml
```diff
@@ -1,9 +1,18 @@
 package:
   name: typing-extensions
-  version: 3.10.0.0
+  version: 4.0.1
+  # I needed some dumb hacks to get this to work. TODO: fix build system after
+  # 0.19 release.
```
rth, line 5: Maybe let's open an issue to remember it?

src/tests/python_tests.txt
```diff
@@ -351,10 +351,10 @@ test_module
 test_modulefinder
 test_msilib
 test_multibytecodec
-test_multiprocessing_fork
-test_multiprocessing_forkserver
-test_multiprocessing_main_handling
-test_multiprocessing_spawn
```
rth, line 357: Previously in those tests multiprocessing was correctly detected as missing at runtime and now it's no longer the case. Maybe a better solution is to monkeypatch the problematic method? In any case, it would be worth doing these multiprocessing changes in a separate PR.

rth (commented):
A few comments. Multiprocessing changes aside LGTM.

packages/typing-extensions/meta.yaml
hoodmane: Sure. My thought is that with #2027, we can switch to using unmodified wheels from PyPI for all of our unpatched pure python packages.

rth: Yes, that would be ideal. Though we still might want to unvendor tests which would require repacking of the wheel but shouldn't be too hard (and won't need to run setup.py).

.circleci/config.yml
hoodmane: Okay, thanks. I just want to ignore the `_multiprocessing.py`, which of course is another point of concern.

packages/typing-extensions/meta.yaml
hoodmane: Yeah, my setup in #2027 divides up the process of creating the package into:
1. create or acquire wheel
2. `python -m wheel unpack` the wheel
3. Apply post script and unvendor tests
4. `python -m wheel pack`

src/tests/python_tests.txt
hoodmane: Opened a separate PR #2094.

hoodmane:
Thanks for the review @rth!

Issue #1476, opened 2021-04-16
Separate build/host and run dependencies in meta.yaml


rth:
In `meta.yaml` I think it would make sense to differentiate between build/host and run dependencies, same as conda does https://conda-forge.org/docs/maintainer/adding_pkgs.html#build-host-and-run

This would,
 - improve the parallelism for building packages. Since for instance, a number of packages depend on scipy, by few actually need scipy to build, and right now all those packages are waiting for scipy to build.
 - allow us to explicitly check that build dependencies are available (and if necessary install them). Which should address issues like https://github.com/pyodide/pyodide/issues/936

Currently the problem for this migration is that we are using [`setup.py build`](https://github.com/pyodide/pyodide/blob/4c315c7c2fb24cdd5c57107094b95190091b08d9/pyodide_build/pywasmcross.py#L137-L139) which doesn't have a `--no-deps` version as far as I can tell, and so will fetch all dependencies at this step.

Another possible workaround I considered would be to monkeypatch `socket.socket` at install time to prevent internet access, but I think it's able to escape that as setuptools uses subprocess internally.

Also related to https://github.com/pyodide/pyodide/issues/795

PR #2086, opened 2022-01-07
Update astropy to version 5.0

hoodmane:
Also removed the patch =)


packages/extension-helpers/meta.yaml
```diff
@@ -0,0 +1,9 @@
+package:
```
hoodmane, line 1: `extension-helpers` is only a build-time dependency. I think it is better just to add this as `pip install extension-helpers` in the build script.

hoodmane: At least I believe this is the case...

rth: Indeed, thanks! We should really add build dependencies to `meta.yaml` #1476

PR #2087, opened 2022-01-08
CI Deploy pyodide-build to PyPi on releases

rth:
Aims to simplify the release process a bit by auto-deploying pyodide-build to PyPi on git tags.

(Done from an upstream branch to be able to test it and pass secrets)

### DIFF
--- a/docs/development/maintainers.md
+++ b/docs/development/maintainers.md
@@ -39,33 +39,22 @@ the latest release branch named `stable` (due to ReadTheDocs constraints).
    ```
 5. Create a tag `X.Y.Z` (without leading `v`) and push
    it to upstream,
+
    ```bash
    git tag X.Y.Z
    git push upstream X.Y.Z
    ```
+
    Create a new `stable` branch from this tag,
+
    ```bash
    git checkout -b stable
    git push upstream stable --force
    ```
+
    Wait for the CI to pass and create the release on GitHub.
-6. Release the `pyodide-build` package and `pyodide` package:
-   ```bash
-   pip install twine build
-   cd pyodide-build/
-   python -m build .
-   ls dist/   # check the produced files
-   twine check dist/*X.Y.Z*
-   twine upload dist/*X.Y.Z*
-   ```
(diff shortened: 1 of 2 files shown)


docs/development/maintainers.md
```diff
@@ -39,41 +39,30 @@ the latest release branch named `stable` (due to ReadTheDocs constraints).
...
-6. Release the `pyodide-build` package and `pyodide` package:
-   ```bash
-   pip install twine build
-   cd pyodide-build/
-   python -m build .
-   ls dist/   # check the produced files
-   twine check dist/*X.Y.Z*
-   twine upload dist/*X.Y.Z*
-   ```
-   And to release the `pyodide` package:
```
rth, line 61: @hoodmane  I missed this part during the review, but the initial idea was not to release the pyodide package on PyPi and to have fixed version  https://pypi.org/project/pyodide/0.0.2/ that fails if you pip install it and points to GH for more information.

 If we put it on PyPi, it's going to be one of the top hits in a search engine when searching for Pyodide, and something like https://pypi.org/project/pyodide/0.19.0a1/ can be fairly confusing to new users (as in "what do I do now with this"), particularly that `pip install pyodide` would work but fail at import I suppose outside of WASM. 

This addition was about making the core package more modular, as far as I remember. If we want to put it on PyPi maybe at least call something like pyodide-core (and still imported as pyodide) to be more specific?

rth (commented):
OK, I conform that upload to test PyPi worked with this setup.

docs/development/maintainers.md
hoodmane: > the initial idea was not to release the pyodide package on PyPi and to have fixed version pypi.org/project/pyodide/0.0.2 that fails if you pip install it and points to GH for more information

I never understood this, but it makes much more sense. I agree that `pyodide_core` or `pyodide` or whatever is not very useful as a PyPI package so it's not clear why anyone would want to install it. Having `pip install pyodide` just print some info and a link to github seems clearly better.

PR #2088, opened 2022-01-08
Update yt

hoodmane:


hoodmane:
We hit this issue: https://github.com/yt-project/yt/issues/3477
So I guess this has to wait on yt version 4.0.2.

Issue #2048, opened 2021-12-17
RFC Replacing our web testing framework Selenium with Playwright


ryanking13:
## Problem statement

Currently, we are using Selenium for our web testing framework. Selenium is a battle-proven framework which have long been used in many other projects for end-to-end testing. However, we are experiencing various issues in using Selenium, such as:

### Timeout
 
- Running test timeouts often, due to the mysterious hang while WebDriver session initialization and script execution.
- This happens both in local and CI environments, which [might confuse new contributors](https://github.com/pyodide/pyodide/blob/d6035abb888c3388106c08402627e637591f440e/.github/PULL_REQUEST_TEMPLATE.md?plain=1#L4-L7), [makes test results unreliable](https://github.com/pyodide/pyodide/pull/2035), and also wastes CI resources.
- This issue has been alleviated a little after the CI resource update (#1929) but we are still experiencing timeouts often.

### Test speed
- Running a WebDriver and initializing Pyodide (compiling WASM module) take much time (#1982), which increases the total test running time.
- This can be mitigated through browser caching (#1824) for some tests, but it may raise an isolation issue between tests that needs to be carefully investigated.
 
### Browser upgrade
- Upgrading the Selenium WebDrivers version to up-to-date is not very straightforward (#1843).
  
## Proposal

This proposal suggests replacing our web testing framework Selenium with [Playwright](https://playwright.dev/).

Playwright is a web testing framework maintained by teams at Microsoft, who also built the [puppeteer](https://github.com/puppeteer/puppeteer) at Google. The main difference between Selenium and Playwright is that Playwright uses [Devtools protocol](https://chromedevtools.github.io/devtools-protocol/) to communicate with browsers, while Selenium uses WebDriver to communicate with a browser (Selenium also supports Devtools protocol for Chrome, but Playwright supports Chrome, Firefox, and Webkit through the patched version of browser binaries).

I think Playwright can solve issues we are experiencing on using Selenium to some extent.

### Timeout 

Unlike WebDriver that works as a middle-man server between client and browser, Devtools Protocol directly talks to a browser. Therefore, I expect using Playwright will make communication with the browser more stable. Although there's no evidence without my empirical results.

Also, with Playwright, we only need to run the browser once at the startup, then we can isolate each test through [BrowserContext](https://playwright.dev/python/docs/browser-contexts) and [Page](https://playwright.dev/python/docs/pages), which would prevent `session not created` issues in instantiating WebDrivers for each standalone tests.

[...]

PR #2091, opened 2022-01-10
TST Add Playwright for an auxiliary testing framework

ryanking13:
### Description

_Updated:_

Adds auxiliary playwright runner for testing.

```
pytest --runner [playwright, selenium]
```

__

~~I proposed changing our testing framework from Selenium to Playwright (#2048). However, Playwright is not battle-tested in pyodide environment, and it has some performance issues in Firefox. Therefore, I think it is not ready to be the primary testing framework of pyodide for now.~~

~~Instead, this PR makes Playwright an auxiliary testing framework, so that it can be used together with Selenium.
I expect Playwright can help us especially when Selenium is unstable.~~

~~When `PLAYWRIGHT` env variable is set, pytest will use Playwright instead of Selenium.~~

- [x] Update documentation


rth:
Thanks for working on this @ryanking13 and sorry for slow feedback. 

I think it's an interesting idea. I haven't looked in the code in detail yet, however a few general comments,
 - I think having a second `conftest_playwright.oy`, with significant code duplication is not ideal and is going to be difficult to maintain. Actually, even our conftest.py is getting quite large by now. So it could be a good time to,
   - factorize what we can into some new module in `pyodide_build`. We already have `pyodide_build.testing` but it would probably be better to put it in a separate file, as we don't want selenium to be a mandatory run time dependency of pyodide_build (optional dependency would be OK). 
   - For instance, maybe we could use a base class with methods that are re-used between SeleniumWrapper / PlaywrightWrapper. 
  
   This can be done in a separate PR.
  
-  If we do the previous work than maybe the `conftest.py` would be short enough to put both selenium and playwright configs there.
- If we are going to add it, it needs to run in automatically triggered CI otherwise there is a high likelihood that we would break it with future changes. I don't know where it would be better to put it. We could use it on GH Actions for Chrome for instance, if the setup is not too complex.
- For now, maybe let's not add it to the requirements.txt since it's an optional feature (but install in CI when needed).
- For using this runner, I think it would be better to add [a pytest option](https://docs.pytest.org/en/6.2.x/example/markers.html#custom-marker-and-command-line-option-to-control-test-runs) so that we could run it with `pytest --runner=playwright` (with the default being `selenium`)

ryanking13:
> * I think having a second `conftest_playwright.oy`, with significant code duplication is not ideal and is going to be difficult to maintain. Actually, even our conftest.py is getting quite large by now. So it could be a good time to,

Yeah, I definitely agree. I didn't want to make `conftest.py` more complicated, but having an another configuration file looks worse.
I'll look into `pyodide-build` to see what we can do.


> * If we are going to add it, it needs to run in automatically triggered CI otherwise there is a high likelihood that we would break it with future changes. I don't know where it would be better to put it. We could use it on GH Actions for Chrome for instance, if the setup is not too complex.

If we are going to automatically trigger both Selenium and Playwright tests, I think GH Actions would be a good place to put it (Playwright only or maybe both). Acutally, I also have interests in migrating CircleCI jobs to GH Actions.

> * For now, maybe let's not add it to the requirements.txt since it's an optional feature (but install in CI when needed).

Agreed.

> * For using this runner, I think it would be better to add [a pytest option](https://docs.pytest.org/en/6.2.x/example/markers.html#custom-marker-and-command-line-option-to-control-test-runs) so that we could run it with `pytest --runner=playwright` (with the default being `selenium`)

Thank you for the suggestion! I didn't know that option.

ryanking13:
Sorry for the delay. Here is an update, applying suggestions from @rth.
(It looks like the SQLite URL is temporarily dead, I'll re-trigger CI when it's restored)

1. I restructured the test wrapper class, adding a base class `BrowserWrapper`:

```
 ┏ BrowserWrapper
 ┣  SeleniumWrapper 
 ┃  ┣ SeleniumFirefoxWrapper
 ┃  ┗ SeleniumChromeWrapper
 ┣  PlaywrightWrapper 
 ┃  ┣ PlaywrightFirefoxWrapper
 ┃  ┗ PlaywrightChromeWrapper
 ┗  NodeWrapper
```

The `conftest.py` got much bigger due to this change. So in a subsequent PR, I would like to move some general testing-related stuff in a new module, as @rth mentioned.

2. Use playwright for GH actions

My first thought in this PR was to use Playwright for a 2nd testing framework for CI, where we can use it optionally when Selenium is unstable. But as @rth said, there is a high likelihood that we would break it with future changes if it is not automatically triggered. Therefore, I updated to use playwright in GH actions, so that we can double-test in both frameworks.
But actually, I think this is a temporary way and can be changed at any time. Maybe we can run every test in both frameworks later if our CI resource becomes affluent.
(BTW, It feels like selenium tests are quite stable after browser version update yay :) )

.github/workflows/main.yml
```diff
@@ -113,21 +115,26 @@ jobs:
         shell: bash -l {0}
         run: |
           pip install -r requirements.txt
+          # FIXME: playwright>=1.18.0 fails in testing `test_isinstance` due to the
+          #        'Maximum call stack size exceeded' error
+          pip install "playwright<1.18.0" && python -m playwright install
```
ryanking13, line 120: Playwright >= 1.18.0 (which comes with chrome >= v99) fails on this recursion test. I'm not sure this is a Playwright issue or a Chrome issue.

https://github.com/python/cpython/blob/0a7dcbdb13f1f2ab6e76e1cff47e80fb263f5da0/Lib/test/test_isinstance.py#L274-L282

ryanking13:
Back in this PR, I think this is ready to be reviewed.

Compared to the time that this PR was originally suggested, our Selenium test became a lot stable. Still, I think adding support for playwright have some benefits such as:

- Webkit (Safari) support with no additional effort
- Easy to install for new comers

src/tests/test_pyproxy.py
```diff
@@ -369,10 +369,10 @@ def test_pyproxy_get_buffer_type_argument(selenium, array_type):
         if fmt.lower() == "q":
             assert result == [hex(x).replace("0x", "") for x in list(mv.cast(fmt))]
         elif fmt == "f" or fmt == "d":
-            from math import isclose
+            from math import isclose, isnan
 
             for a, b in zip(result, list(mv.cast(fmt))):
-                if a and b:
+                if a and b and not (isnan(a) or isnan(b)):
```
hoodmane, line 375: This test should probably be improved, the `f` and `d` cases are pretty sloppy...
If they are nan shouldn't both be nan?

hoodmane (approved):

src/tests/test_pyproxy.py
ryanking13: > If they are nan shouldn't both be nan?

Both should be NaN, but I am getting some strange results in selenium and node:

```py
def test_temp(selenium):
    selenium.run_js(
        """
        let a = pyodide.runPython("bytes(range(252, 256))")
        let buf = a.getBuffer("f32")
        console.log(buf.data[0])
        """
    )
```

This should print `NaN`, but I get `None` in selenium and node. The playwright prints `NaN` correctly.
It also works correctly in the browsers, so maybe there is a bug in our selenium code.

ryanking13:
Thanks for the reviews!

hoodmane:
Thanks @ryanking13!

Issue #2092, opened 2022-01-10
micropip.list() is not a thing that exists.


S10MC2015:
## 📚 Documentation

https://pyodide.org/en/latest/usage/api/micropip-api.html#micropip.list

![image](https://user-images.githubusercontent.com/30575036/148768275-c103b8c2-5f0f-48a1-94f5-86913165594d.png)

ryanking13:
Thank you for the report! Which version of the Pyodide are you using?

`micropip.list()` will be added at version `0.19.0`, and current stable version of Pyodide is `0.18.1`, so `micropip.list()` won't be exist if you are using the stable version.

rth:
To add more details to the above response, `latest` documentation corresponds to the `main` dev branch. While stable docs is https://pyodide.org/en/stable/usage/api/micropip-api.html and doesn't have that function.

> micropip.list() will be added at version 0.19.0

The release will happen in a couple of days, but you can already use it now with the 0.19.0a1 pre-release.

Closing as answered, but don't hesitate to comment further.

S10MC2015:
Ah, sorry. Thanks for the help.

PR #2093, opened 2022-01-10
DOC Add version alert on latest docs

ryanking13:
### Description

Rendered version: https://pyodide-ryanking13.readthedocs.io/en/latest/

---

Display version alert on `latest` readthedocs documentation that it's an unreleased version.

This would prevent some issues like: #2092

Note: Readthedocs has an option `Show version warning` which `Show warning banner in non-stable nor latest versions.`. But it doesn't seem to have an option to warn on the latest version.


docs/_static/js/version-alert.js
```diff
@@ -0,0 +1,30 @@
...
+  warning.querySelector('a').href = window.location.pathname.replace('/latest', '/stable');
+
+  var parent = document.getElementById("main-content").firstChild;
+  while(parent != null && parent.nodeType == 3){ // skip TextNodes
+    parent = parent.nextSibling;
+  }
+  parent.insertBefore(warning, parent.firstChild);
+}
+
+document.addEventListener('DOMContentLoaded', warnOnLatestVersion);
```
hoodmane, line 30: Add trailing newline

docs/_static/js/version-alert.js
```diff
@@ -0,0 +1,30 @@
...
+
+function warnOnLatestVersion() {
+
+  // The warning text and link is really specific to RTD hosting,
+  // so we can just check their global to determine version:
+  if (!window.READTHEDOCS_DATA || window.READTHEDOCS_DATA.version !== "latest") {
+    return;  // not latest, or not on RTD
+  }
+
+  var warning = document.createElement('div');
```
hoodmane, line 14: Maybe it would be simpler to add the warning to the html template and use a css selector or inline style with `display: none`?
Then this code could just replace the link and set an inline style to `display: block`.

docs/_static/js/version-alert.js
```diff
@@ -0,0 +1,30 @@
...
+  warning.setAttribute('class', 'admonition danger');
+  warning.innerHTML = "<p class='first admonition-title'>Note</p> " +
+    "<p class='last'> " +
+    "This is <strong>unreleased documentation for Pyodide next version</strong>. " +
+    "For up-to-date documentation, see <a href='/en/stable/'>the stable version</a>." +
```
hoodmane, line 19: Maybe:
```suggestion
    "<strong>This is documentation for the development version of Pyodide.</strong> " +
    "<a href='/en/stable/'>Click here</a> for documentation about the latest stable version of Pyodide."
```
Might also be nice to provide an instruction for what people should do if they want to try out development APIs. Could be worth looking for another project with a warning like this.

hoodmane (commented):
Thanks @ryanking13!

docs/_static/js/version-alert.js
ryanking13: Okay, I'll look into the sphinx documentation whether we can override the theme we are using.

hoodmane: This is fine if it's too confusing.

docs/_static/js/version-alert.js
ryanking13: Yeah, it could help people who want to try unreleased APIs.
Other projects that I referenced is: [docusaurus](https://docusaurus.io/docs/next), [playwright](https://playwright.dev/docs/next/intro),  [marshmallow](https://marshmallow.readthedocs.io/en/latest/). They don't have such instructions, so I think we should create our own one.

ryanking13:
Well, it turned out that we already have this functionally. But it was not enabled. 😂

https://github.com/pyodide/pyodide/blob/9d515cc2d2524658c2a6e72039acf7aed63da311/docs/conf.py#L66-L73

https://github.com/pyodide/pyodide/blob/9d515cc2d2524658c2a6e72039acf7aed63da311/docs/conf.py#L56

ryanking13:
It looks like `sphinx-issue`'s version update makes the build broken.

ryanking13:
https://pyodide-ryanking13.readthedocs.io/en/latest/

Okay, I pinned the `sphinx-issue` version to 2.0.0 and it works now.
#1567 was due to the typo in string concatenation.

hoodmane (approved):
@ryanking13 is this ready to merge?

ryanking13:
> @ryanking13 is this ready to merge?

@hoodmane Yes, it is ready to be merged.

I think we could add some older version warnings too when #2100 has been resolved.

hoodmane:
Sounds good to me.

PR #2095, opened 2022-01-10
MAINT Apply prettier to everything by default

hoodmane:
This switches to checking everything with prettier, except for files specifically excluded with `.prettierignore`.


.circleci/s3-website-config.json
```diff
@@ -1,18 +1,18 @@
...
-                    "KeyPrefixEquals": "latest/"
-                },
-                "Redirect": {
-                    "ReplaceKeyPrefixWith": "v0.14.3/"
-                }
-            }
-        ]
-    }
+        "Redirect": {
+          "ReplaceKeyPrefixWith": "v0.14.3/"
```
hoodmane, line 13: Does this file do anything? It's suspicious that we haven't replaced this version for quite so long...

rth: I completely forgot about it.

https://cdn.jsdelivr.net/pyodide/latest/full/packages.json does redirect to https://cdn.jsdelivr.net/pyodide/v0.14.3//full/packages.json. But anyway it's not a very good idea to rely on URLs that change in time for data packages, so let's just remove this "RoutingRules" sections. Thanks for catching it!

rth (approved):
Thanks!

Issue #2101, opened 2022-01-11
CI Pre-build docker image not deploying on tags


rth:
There was an issue with deploying the pre-build docker image during last release,
https://github.com/pyodide/pyodide/runs/4768642291
```
error: invalid tag "pyodide/pyodide:": invalid reference format
Error: buildx failed with: error: invalid tag "pyodide/pyodide:": invalid reference format
```
which is an indication that in
https://github.com/pyodide/pyodide/blob/main/.github/workflows/prebuilt_docker_image.yml#L36
`github.event.inputs.version` needs to be replaced by the tag value. I couldn't test this part, because it requires creating a tag in the repo (and I haven't managed to make GH Actions to work in a fork).

cc @ehfd

ryanking13:
Docker images can be updated often (whenever we update Dockerfile or requirements.txt), so I think using git tag value is not a good way.

I think we better figure out why `github.event.inputs.version` was empty. It should be a srting that be manually set when running the workflow.

rth:
So the `pyodide-env` docker image can indeed be updated any time and has a manual trigger (that part works).
For the `pyodide` docker image with pre-build packages we usually only update it on releases, as it gives a reference point for people who don't want to spend time building from sources. In which case building it only on tags makes sense IMO.  One of the ideas of this is to reduce the number of manual actions needed for the release. 

> I think we better figure out why github.event.inputs.version was empty

Well, I think because it's indeed the manual trigger input, while we are not using it for releases.

ryanking13:
> For the `pyodide` docker image with pre-build packages we usually only update it on releases, as it gives a reference point for people who don't want to spend time building from sources. In which case building it only on tags makes sense IMO. One of the ideas of this is to reduce the number of manual actions needed for the release.

Oops. Sorry, I misunderstood. I was thinking of `pyodide-env` docker image.

ehfd:
> which is an indication that in
> https://github.com/pyodide/pyodide/blob/main/.github/workflows/prebuilt_docker_image.yml#L36
> `github.event.inputs.version` needs to be replaced by the tag value. I couldn't test this part, because it requires creating a tag in the repo (and I haven't managed to make GH Actions to work in a fork).

It's a bit complicated than what I previously expected.
https://github.community/t/how-to-get-just-the-tag-name/16241
This is the solution.

rth:
Thanks @ehfd, if you can make a PR that would be perfect ! We will test it for the future 0.19.1 release :)

[...]

PR #1995, opened 2021-11-24
Update GitHub Workflows Image Build and Support ghcr.io

ehfd:
### Description

PR based on issue #1851. Implemented trigger to tags and upload to ghcr.io for pre-built docker image, and to the non-prebuilt images also.

Closes #1851.

### Checklists

- [x] Configure the credentials for GitHub Container Repository correctly (requires admin intervention)
- [x] Implement pyodide/pyodide-env
- [x] Q. Does this workflow require QEMU? In other words alternative architectures such as aarch64, s390x, or others.
- [x] @rth should test everything works on his own environment

The default checklists:

- [x] Add a [CHANGELOG](https://github.com/pyodide/pyodide/blob/main/docs/project/changelog.md) entry
- [x] Add new / update outdated documentation

PR #2105, opened 2022-01-12
Fix CI Pre-build docker image not deploying on tags

ehfd:
### Description

Fix bug issue #2101 introduced while implementing PR #1995.
NOTE: I am 99% sure this is the fix, but since `${{ github.event.release.tag_name }}` is available only precisely with the workflow triggered by a new release, real life testing is only the way to verify (https://github.community/t/how-to-get-just-the-tag-name/16241/28). I just need confirmation that this should be triggered on release, not just a tag.

### Checklists

- [ ] Add / update tests

### DIFF
--- a/.github/workflows/prebuilt_docker_image.yml
+++ b/.github/workflows/prebuilt_docker_image.yml
@@ -1,8 +1,7 @@
 name: prebuilt-docker-image
 on:
-  push:
-    tags:
-      - "*.*.*"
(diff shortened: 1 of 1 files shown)


hoodmane:
Thanks @ehfd! We really appreciate your help with this.

rth:
Thank you @ehfd !

PR #2106, opened 2022-01-12
MAINT Refactor of load-pyodide (reduces diff to wheels PR)

hoodmane:
This simplifies handling of node modules in load-pyodide a bit. Eventually we could split this off into a separate compatibility file or something.

This also fixes a significant bug, which is that `readFile` is allowed to return a `Buffer` whose backing data is shared with other buffers. When we convert to `Uint8Array`, we have to keep track of which part of the buffer we are supposed to be looking at.

I also rearranged the `loadPackage` internals to prepare for #2027. I changed the error handling behavior a bit, so in some circumstances where `loadPackage` would have logged an error, it now throws an error.

This is split off from #2027.

### DIFF
--- a/src/js/load-pyodide.js
+++ b/src/js/load-pyodide.js
@@ -350,24 +350,15 @@ export async function loadPackage(names, messageCallback, errorCallback) {
   if (Module.isPyProxy(names)) {
     names = names.toJs();
   }
-
   if (!Array.isArray(names)) {
     names = [names];
   }
 
-  const [toLoad, toLoadShared] = recursiveDependencies(
-    names,
-    messageCallback,
-    errorCallback
-  );
-  const toLoadAll = [...toLoad, ...toLoadShared];
+  const [toLoad, toLoadShared] = recursiveDependencies(names, errorCallback);
   if (toLoad.size === 0 && toLoadShared.size === 0) {
     messageCallback("No new packages to load");
     return;
   }
-
-  let packageNames = Array.from(toLoad.keys()).join(", ");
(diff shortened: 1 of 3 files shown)


src/js/load-pyodide.js
```diff
@@ -375,52 +362,133 @@ function restoreOrigWasmPlugin() {
...
-    // do nothing - let the main load throw any errors
+
+  const [toLoad, toLoadShared] = recursiveDependencies(names, errorCallback);
+  if (toLoad.size === 0 && toLoadShared.size === 0) {
+    messageCallback("No new packages to load");
+    return;
   }
 
   let releaseLock = await acquirePackageLock();
+  for (let [pkg, uri] of [...toLoad, ...toLoadShared]) {
```
rth, line 381: Doesn't it make sense to load the shared libraries before the other packages, though I guess it doesn't make a difference in this particular case aside from the logging order.

rth (approved):
It's hard to make the correspondence with the previous code version in the diff, but I read through the new version of the code, and overall it LGTM. Thanks!

src/js/load-pyodide.js
hoodmane: I believe that the shared libraries need to be loaded first to ensure that linking works correctly (of course there is still exactly one shared library CLAPACK aside from my test library).

hoodmane: But yeah, I don't think the logging order matters much?

hoodmane:
Also, worth noting that this *does* now change the error handling behavior of `loadPyodide` so it's not a pure maintenance PR anymore.

PR #2107, opened 2022-01-13
fix js syntax typo in switch statement

fcrozatier:
The `continue` keyword in js has no effect in switch statements


hoodmane:
But it's also inside a for loop...

fcrozatier:
My bad in this context it is equivalent, sorry !

hoodmane:
No they are not equivalent, they are different and it needs to be continue...

PR #2110, opened 2022-01-16
Fix numpy comparator functions

hoodmane:
This creates fatal errors when used with e.g., `numpy.unique`, see added test.

My upstream PR https://github.com/numpy/numpy/pull/20833 with this patch was accepted.


hoodmane:
With this, `scipy.stats` tests make it all the way through without a fatal error.

rth:
Thanks @hoodmane !

PR #2112, opened 2022-01-17
PKG added sqlalchemy

clstaudt:
### Description

Add package `sqlalchemy` to pyodide distribution.

### Note

I was not able to build and test the package yet due to an issue with the Docker image. As discussed with @hoodmane, I am sending the PR anyway. Thank you for integrating `sqlalchemy` quickly.


bartbroere:
I'm afraid running `prettier` on the yaml has removed everything from the file.

hoodmane:
We really should fix `mkpkg` so that it runs `prettier` automatically...

clstaudt:
@bartbroere @hoodmane  Sorry, I made it worse by trying to also make the linter happy. Do you need me to fix the PR?

rth:
Thanks @clstaudt ! Yes, we would need to fix this PR to merge it. The first commit looked good aside from the linting issue. To apply linters, the easiest is to use pre-commit https://pyodide.org/en/stable/development/contributing.html#development-workflow

rth (approved):
Thanks LGTM.

hoodmane:
@rth is this okay to add to the next patch release? Not sure what our policy regarding package additions / updates and patch releases is. Seems like updates could be problematic, but adding a package isn't likely to break much existing code?

rth:
Agreed. I don't think there is a reason not to include it.

clstaudt:
@rth @hoodmane  Thank you. When do you plan to release the next patch?

rth:
A few weeks to a month I think. But you can already use it now from the dev channel https://pyodide.org/en/stable/usage/downloading-and-deploying.html#cdn

Issue #2111, opened 2022-01-17
New Package (pysam, and modbampy)


rrazaghi:
## 🐍 Package Request 1

- Package Name and Version : pysam
- Package URL :  https://github.com/pysam-developers/pysam
- Package Dependencies that needs to be resolved first: htslib

## 🐍 Package Request 2

- Package Name and Version : modbampy
- Package URL :  https://github.com/epi2me-labs/modbam2bed  , https://pypi.org/project/modbampy/
- Package Dependencies that needs to be resolved first: htslib

Hi,
thank you for this great project. I have been trying to build these two packages however I can't make any progress at the build (python -m pyodide_build buildpkg meta.yaml)

I would really appreciate it if you could help me with this. We can start with modbampy as it is a much much smaller package than pysam. pysam is heavily used in bioinformatics.

Thank you!

rrazaghi:
for pysam I receive the following issue when compiling htslib: 
`htslib/hfile_libcurl.c:47:10: fatal error: 'curl/curl.h' file not found
#include <curl/curl.h>
         ^~~~~~~~~~~~~
1 error generated.`

rrazaghi:
for modbampy I have some success adding some patches but now I'm at this:
`emcc: error: Passing any of -msse, -msse2, -msse3, -mssse3, -msse4.1, -msse4.2, -mavx, -mfpu=neon flags also requires passing -msimd128!`

what's the best way to resolve this?

hoodmane:
It's likely it would be helpful to filter out these flags in pywasmcross.
If you return `None` from this function:
https://github.com/pyodide/pyodide/blob/039618988fe9a908b08ef4f4f649775e3fdcd988/pyodide-build/pyodide_build/pywasmcross.py#L354
it will filter out the flag.

It's probably easiest if you open a PR with your partial progress.

rrazaghi:
Thank you @hoodmane! This is all very new to me and I'd appreciate your help.

rrazaghi:
It would be really great to figure out htslib dependencies. Almost all of the tools written for bioinformatics use htslib. This would really expand pyodide’s use case you n the field of genomics!

rrazaghi:
Hi,

I have made some progress on pysam package (I have been able to build htslib separately) however I am stuck at this stage:

to build pysam without the libcurl dependencies we have to install this way according to documentation:

`HTSLIB_CONFIGURE_OPTIONS='--enable-plugins --disable-libcurl' python setup.py build`

[...]

PR #2113, opened 2022-01-17
modbampy

rrazaghi:
### Description

partial PR for adding modbampy to packages. see #2111


rrazaghi:
I'm not sure how helpful this is but I found this relatively old repo that compiles htslib with Emscripten:

https://github.com/HiDiHlabs/htslib-js

PR #2114, opened 2022-01-18
Fix pywasmcross when current directory is changed

hoodmane:
If the current directory is changed during the process of building the package, then all the build commands in the different working directory go into a different `build.log` which is then ignored. This updates `pywasmcross` to keep track of the directory.

- [ ] Add a test.


rth:
Yes, we need to be careful with all those current directory changes.

hoodmane:
Closed in favor of #2256.

PR #2116, opened 2022-01-19
Add basic test for sqlalchemy

ryanking13:
A PR that supplement #2112.


rth (approved):
Thanks @ryanking13 , it would be good to have a test for it indeed.

PR #2120, opened 2022-01-21
Drop ruamel.yaml dependency

ryanking13:
`ruamel.yaml` was introduced at https://github.com/pyodide/pyodide/pull/1465 to prevent `mkpkg` command from reordering keys in meta.yaml.
But it seems like `pyyaml.dump(..., sort_keys=False)` does the same thing we want. So I think it's fine to remove the dependency.

__Checklists__

- [x] Add / update tests

### DIFF
--- a/pyodide-build/pyodide_build/mkpkg.py
+++ b/pyodide-build/pyodide_build/mkpkg.py
@@ -10,6 +10,7 @@ import sys
 from pathlib import Path
 from typing import Dict, Any, Optional
 import warnings
+import yaml
 
 from .io import parse_package_config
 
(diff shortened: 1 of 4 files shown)


hoodmane:
`ruamel.yaml` supposedly can also preserve comments which would be useful.

ryanking13:
> `ruamel.yaml` supposedly can also preserve comments which would be useful.

Oh, right. I didn't know that. Preserving comments are important for us.

ryanking13:
It looks obvious that [PyYAML won't support loading/saving comments in the short term](https://github.com/yaml/pyyaml/issues/90), so I'll close this for now.

PR #2123, opened 2022-01-23
Use error-stack-parser to make better stack traces

hoodmane:
With `ErrorStackParser` we can add more detail to the error message we produce when a Javascript error is raised into Python.

For example, the following code:
```js
function f(){
 throw new Error("hello there?");
}
function g(){
    f();
}
pyodide.runPython(`
from js import g
g()
`);
```
now results in the traceback:
```py
  File "???", line 5, in g
  File "???", line 2, in f
JsException: Error: hello there?
```
instead of before:
```py
JsException: Error: hello there?
```

### Checklists

- [x] Add / update tests


src/js/rollup.config.js
```diff
@@ -12,7 +14,10 @@ function config({ input, format, minify, ext = "js" }) {
       format,
       sourcemap: true,
     },
+    external: ["path", "fs/promises", "node-fetch", "vm"],
     plugins: [
+      commonjs(),
+      nodeResolve(),
```
rth, line 20: Could you please add a comment on what this does? I'm still not sure after reading their readme.

rth (commented):
Thanks you! Could you please add a test for it? Otherwise looks good to me.

PR #2125, opened 2022-01-23
FIX REPL tab at beginning of line should indent not complete

eagleoflqj:
### Description

In pyodide REPL, if you type
```py
for i in range(3):  # with a '\n'
```
then press tab, you get completion suggestions. But in normal REPL, a '\t' should be entered.
This PR checks the command and determines whether entering '\t' or falling back to completion.
### Checklists

### DIFF
--- a/docs/project/changelog.md
+++ b/docs/project/changelog.md
@@ -30,6 +30,9 @@ substitutions:
   improved with some context information.
   {pr}`2121`
 
+- {{Enhancement}} Pressing TAB in REPL no longer triggers completion when input
+  is whitespace. {pr}`2125`
+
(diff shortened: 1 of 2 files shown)


hoodmane:
Thanks @eagleoflqj! Looks good to me. Currently, the thing that bothers me the most about our repl is how hard it is to copy out of it...

eagleoflqj:
CI for the first commit passed https://github.com/pyodide/pyodide/actions/runs/1737304833
The second changelog commit conflicts with main, so I rebased and force pushed (thus the CI result is gone)

hoodmane:
We squash so it's okay to merge rather than rebase.

PR #2126, opened 2022-01-24
Use wheels instead of sdists for pure Python packages

hoodmane:
After #2027, we can directly use pure Python wheels from PyPI, instead of downloading the source distribution and running `setup.py`. This switches to using wheels as much as possible.

I updated `mkpkg` so that:
1. Making a new package, by default the preference is to use a pure Python wheel. If there isn't a pure Python wheel, then use an sdist. This can be adjusted by using the `--sdist` flag which indicates to prefer to use an `sdist`.
2. Updating a package, by default the preference is to keep the current type. Failing that, it will check if the other type is available. With `--sdist`, we can demand an sdist, with `--wheel` we can demand a wheel.

I also fixed `mkpkg` to run prettier on `meta.yaml` because it's annoying when auto-generated files fail the lint checker.

### Checklists

- [x] Add / update tests


rth:
What happens when both `wheel` and `sdist` are provided or none are provided? It sounds like a mutually exclusive flag. Wouldn't just the `wheel` flag be enough, and then argparse would also support `--no-wheel`,
```
>>> import argparse
>>> parser = argparse.ArgumentParser()
>>> parser.add_argument('--foo', action=argparse.BooleanOptionalAction)
>>> parser.parse_args(['--no-foo'])
Namespace(foo=False)
```

For the `make_package` function maybe let's call the argument `wheel_source` (or something similar)? Because for a function called `make_package(..., wheel=True)` out of context it's not very clear if we are talking about wheel as input or as output.

Otherwise LGTM.

Another unrelated thing is that for packages with wheel input, we can assume there are no build dependencies, which is a special case of https://github.com/pyodide/pyodide/issues/1476

rth:
In terms of build runtime could you check how long the repackaging of the wheel takes, as compared to creating a wheel from sdist?

I'm a bit surprised that the `build-packages-no-numpy-dependents` job takes longer in this PR (7min) as compared to master (5min). Though it's not very critical.

hoodmane:
> What happens when both `wheel` and `sdist` are provided or none are provided? It sounds like a mutually exclusive flag.

Well currently there are four options: I only want a wheel (`--wheel`), I only want an sdist (`--sdist), I am happy with either preference keep same type (neither), I am happy with either preference wheel (both). But this is probably not a great design.

> could you check how long the repackaging of the wheel takes, as compared to creating a wheel from sdist?

To be clear we do that always. From sdist:
1. download sdist
2. run build script
3. run setup.py
4. unpack wheel
5. run post script and unvendor tests
6. repack wheel

From wheel:
1. download wheel
2. unpack wheel
3. run post script and unvendor tests
4. repack wheel

Maybe what's happening is the wheels are bigger so it takes longer to download them?

rth:
> Well currently there are four options: I only want a wheel (--wheel), I only want an sdist (`--sdist), I am happy with either preference keep same type (neither), I am happy with either preference wheel (both). But this is probably not a great design.

Ok, how about then,
```
source_fmt=Optional[Literal['wheel', 'sdist']]
```
in CLI it would be `--source-fmt` or  even `--source-format` (or --source) another name if you prefer.

In CLI when source_fmt=None (or not default value and not provided in CLI arg) whatever would make sense as a default: I imagine try find a wheel and if it doesn't exist use sdist)? I don't really understand the practical difference between "neither" and "both" in the previous comment.

rth:
> Maybe what's happening is the wheels are bigger so it takes longer to download them?

I would imagine pure python wheels should be smaller. But anyway it's a minor point, this is certainly the way to go, we can always tune performance later.

ryanking13:
> > Maybe what's happening is the wheels are bigger so it takes longer to download them?
> 
> I would imagine pure python wheels should be smaller. But anyway it's a minor point, this is certainly the way to go, we can always tune performance later.

I guess it's not due to this PR. Build time varies often and the most time-taking jobs are large C extension libraries like CLAPACK, libxml.

pyodide-build/pyodide_build/mkpkg.py
```diff
@@ -186,14 +217,26 @@ def update_package(package: str, update_patched: bool = True):
...
+        sdist = True
+        if not yaml_content["source"]["url"].endswith("whl"):
+            # If it is currently an sdist, prefer to update to an sdist
+            dist_metadata = _find_sdist(pypi_metadata)
+
+    if not dist_metadata:
+        dist_metadata = _find_dist(pypi_metadata, wheel, sdist)
 
-    yaml_content["source"]["url"] = sdist_metadata["url"]
+    assert dist_metadata
```
rth, line 231: maybe let's rather raise an ValueError with an explicit message?

pyodide-build/pyodide_build/mkpkg.py
```diff
@@ -140,17 +176,19 @@ def success(msg):
...
 
     meta_path = PACKAGES_ROOT / package / "meta.yaml"
-    try:
-        yaml_content = parse_package_config(meta_path)
-    except:
+    if not meta_path.exists():
+        print(f"{meta_path} does not exist")
         sys.exit(0)
+    with open(meta_path, "rb") as fd:
+        yaml_content = yaml.load(fd)
```
rth, line 191: Slightly better to use the previous function, no, in case we change exactly how we parse/validate the .yaml files.

rth (commented):
A few comments, otherwise LGTM, thanks! 

There are a couple of CI failures in core tests though.

pyodide-build/pyodide_build/mkpkg.py
hoodmane: The problem is `parse_package_config` doesn't use `ruamel.yaml` so if you use `parse_package_config` then `mkpkg update` will destroy comments.

pyodide-build/pyodide_build/mkpkg.py
hoodmane: `_find_dist` never returns `None` instead it raises a `MkpkgFailedException`, so it should be impossible for this assertion to fail. It is just here to convince mypy that `dist_metadata` is not `None`.

hoodmane:
@rth You also requested changes above:

> Ok, how about then,
> 
> ```
> source_fmt=Optional[Literal['wheel', 'sdist']]
> ```
> 
> in CLI it would be `--source-fmt` or even `--source-format` (or --source) another name if you prefer.
> 
> In CLI when source_fmt=None (or not default value and not provided in CLI arg) whatever would make sense as a default: I imagine try find a wheel and if it doesn't exist use sdist)?

I think that's probably a good idea, but I haven't done it yet.

ryanking13:
Thanks, @hoodmane. LGTM, when CI passes and changelog updated.

> ```
> source_fmt=Optional[Literal['wheel', 'sdist']]
> ```
> 
> in CLI it would be `--source-fmt` or  even `--source-format` (or --source) another name if you prefer.

I like this too.

BTW, how about adding ruamel.yaml to the `install_requires` of `pyodide-build`, and remove `_import_ruamel_yaml` stuff?
I don't think there is a reason to install it seperately.

hoodmane:
> BTW, how about adding ruamel.yaml to the install_requires of pyodide-build, and remove _import_ruamel_yaml stuff?
I don't think there is a reason to install it seperately.

+1 for this.

pyodide-build/pyodide_build/mkpkg.py
hoodmane: Okay, so the better solution here is to fix the return type of `_find_dist`.

PR #2129, opened 2022-01-24
Split off webworker / node compatibility into separate file

hoodmane:
I moved `IN_NODE`, `loadScript` and `_loadBinaryFile` which depend on whether we are in node / a webworker / browser main thread into a separate file. I also renamed `load-pyodide.js` to `load-package.js`, which is a more appropriate name.


rth (approved):
There is a failure in test-js  otherwise LGTM. Thanks!

PR #2130, opened 2022-01-24
Fix typescript documentation

hoodmane:
#2133 moved the Javascript code to typescript except for the stuff included via `EM_JS`, but at the same time messed up the api docs. This fixes the API docs.

The main difficulty here is that `sphinx_js` only partially works for typescript. I monkey patch `_convert_node` to fix two crashes and to expand destructured arguments so that a function:
```ts
/**
* @param options
*/
function f({x , y } : {
    /** The x value */
    x : number,
    /** The y value */
    y : string
}){ ... }
```
will have api docs that looks like:
```
options.x (number) The x value
options.y (number) The y value
```
I also monkey patched `_type_name`: for certain types the old `_type_name` just returns a "TODO". If the type name returned by `_type_name` contains a "TODO", then I have extra logic to fix the type name.

### Checklists


docs/conf.py
```diff
@@ -57,9 +61,27 @@
...
+def remove_pyproxy_gen_ts():
+    Path("../src/js/pyproxy.gen.ts").unlink(missing_ok=True)
+
+
+atexit.register(remove_pyproxy_gen_ts)
+
+os.environ["PATH"] += f':{str(Path("../src/js/node_modules/.bin").resolve())}'
+
+if not shutil.which("typescipt") or not shutil.which("typedoc"):
+    subprocess.run(["npm", "ci"], cwd="../src/js")
```
rth, line 78: We might want to wrap this into some `if` condition checking that we are indeed running in sphinx, and not say that this file was mistakenly imported during pytest discovery.

rth:
Thanks for doing this @hoodmane ! 

Is there any chance this could be later split in smaller PRs: right now `+1,350 −619` is not going to be easy to review (even if changes are fairly consistent). For instance, maybe,
 - 1 separate smaller PR to use typescript in the JS package, but without the docs generation (even if it temporarily breaks the docs on main)
 - then what's left in this PR to fix the docs? 
WDYT?

docs/conf.py
hoodmane: Is there a good way to detect being inside readthedocs? We probably want to fail rather than run `npm ci` if the user is running this. I'd prefer to run it in a separate script but readthedocs doesn't seem to support that. Another option might be to move some stuff into a separate `conf_readthedocs.py` which could read as follows:
```py
import subprocess
subprocess.run(["npm", "ci"], cwd="../src/js")
from conf import *
```

hoodmane: Someone here has a technique to detect sphinx:
https://github.com/sphinx-doc/sphinx/issues/9805

rth: For being inside readthedocs see https://github.com/pyodide/pyodide/blob/main/docs/conf.py#L130

Well contributors still might want to build the docs locally, and ideally, it shouldn't fail even if it doesn't build everything

hoodmane:
> Is there any chance this could be later split in smaller PRs: right now +1,350 −619 is not going to be easy to review (even if changes are fairly consistent).

Note also that a lot of changes are from `package-lock.json`:
```sh
$ git diff --stat main HEAD -- ':(exclude)*lock*' src/
11 files changed, 650 insertions(+), 525 deletions(-)
$ git diff --stat main HEAD -- docs/
 6 files changed, 273 insertions(+), 111 deletions(-)
```

hoodmane:
I also don't really feel like figuring out how to fix the webpack test. I think it will probably involve a bunch of configuration nonsense.

hoodmane:
> Though maybe could you add a sphinx_pyodide/README.md explaining the reasons that make it necessary to have our own jsdoc.py as opposed to using sphinx_js without changes? 

Sure, that's a good idea. I think there are two main things going on: 

1. controlling the organization of the docs, and
2. `sphinx_js` is somewhat incomplete so I filled in some missing features (particularly related to automatic doc generation but also it incorrectly handles certain situations).

On controlling the organization of the docs: `sphinx-js` is opinionated about how code should be structured and our code is not structured the way it expects. `core/pyproxy.ts` in particular is problematic because of the dynamic class generation, which the auto generators don't really understand. In the docs, we merge all of the mixins into the main `PyProxy` class which requires custom logic. 

Similarly, it isn't obvious to the documentation generation that `loadPyodide` is globally exposed and most of the functions in `api.ts` are `pyodide` APIs. We could probably improve on this by restructuring the code, but before I found the logic of how the documentation generation works to be rather opaque and had to use a lot of trial and error to figure out how to rearrange the code to get the docs to look right. Ideally in the long run we should arrange the code in a way that is transparent to the documentation generator as far as possible. But if this imposes unnatural constraints on the code, I think it's better to add hacks in the documentation generator to get the docs right rather than to add weird hacks into the code itself. 

Either way, it's a maintenance hazard though.

hoodmane:
Some of these changes should be upstreamed into sphinx-js, but sphinx-js is not actively developed (the last commit was May 6th and the three PRs opened since then have not been responded to) so I don't know how likely that is to work.

rth:
Yes, as far as I can tell, the main dev left Mozilla for Google and hasn't committed anything in the last 3 months, so currently integrating changes upstream there doesn't look very promising indeed.

hoodmane:
> Another small thing in the rendered table we should probably set some max width for the first column as otherwise, it uses almost all of the available space

Yeah I can fix this in a separate PR, which we should backport to `stable`.

PR #2133, opened 2022-01-25
Use typescript (split from #2130, doesn't fix docs)

hoodmane:
As requested by @rth, to reduce the amount of code needing review at once this is split from #2130. This PR breaks the documentation generation by design, #2130 contains code to fix everything except the webpack test. My thought is that this can be reviewed and merged, then #2130, then we can fix the webpack tests (maybe @grimmer0125 can help with that?).

Ignoring `package-lock.json` the diff stat is:
```
13 files changed, 659 insertions(+), 532 deletions(-)
```


src/core/pyproxy.ts
```diff
@@ -361,31 +393,39 @@ class PyProxyClass {
...
-   * ``ConversionError`` rather than producing a ``PyProxy``.
-   * @param {boolean} [options.dict_converter] A function to be called on an
-   * iterable of pairs ``[key, value]``. Convert this iterable of pairs to the
-   * desired output. For instance, ``Object.fromEntries`` would convert the dict
-   * to an object, ``Array.from`` converts it to an array of entries, and ``(it) =>
-   * new Map(it)`` converts it to a ``Map`` (which is the default behavior).
-   * @return {any} The JavaScript object resulting from the conversion.
+   * @param options
+   * @return The JavaScript object resulting from the conversion.
    */
```
rth, line 398: Would we still need the parameter description that was here previously?

rth (approved):
I haven't used typescript much, but this does look like an improvement and I find it very readable.

Minor comment otherwise LGTM.

src/core/pyproxy.ts
hoodmane: This was one of the changes made to try to improve the rendered docs.

bollwyvl:
Awesome!

hoodmane:
Yeah, this is a huge improvement for typings =)

Issue #2067, opened 2021-12-26
JsProxy.to_py() does not convert Date objects


oeway:
## 🐛 Bug
When calling `.to_py()` on a js Date object, it doesn't make it a python datetime object. 

### To Reproduce

```
>>> a = eval("new Date('05 October 2011 14:48 UTC')")
>>> a.to_py()
Wed Oct 05 2011 16:48:00 GMT+0200 (Central European Summer Time)
>>> type(a.to_py())
<class 'pyodide.JsProxy'>
>>> a.to_py().typeof
'object'
```
### Expected behavior
It should be a datetime object:
```
>>> a = eval("new Date('05 October 2011 14:48 UTC')")
>>> type(a.to_py())
<class 'datetime.datetime'>
```
### Environment
- Pyodide Version: the latest from https://pyodide-cdn2.iodide.io/dev/full/console.html

ryanking13:
I think this is not a bug. But we could discuss about it.

Personally, I'm not sure that we should treat Date object differently.

rth:
I would have also expected that to_py should ideally convert it to datetime.datetime. The logic is that we convert stdlib objects between JS to/from Python when there is a strictly equivalent object. We are already doing it for some fraction of  https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects and Date sound like a good candidate. Unless you see some drawback in doing it?

hoodmane:
I'm +1 for this as well, will go ahead and implement.

oeway:
Thank you! 

I think in general when one call .to_py on a js object, the intention is quite clear that we want the output to be a python object.

Maybe we should even think about whether we should throw an error instead of return the same js proxy object if the conversion is not possible. In my case, I am calling to_py on a nested object and I spent quite some time to trace down the source of issue. IMO, it would be much easier if an error can be throw. Maybe we can add an option to tell the to_py()/toJs() conversion to throw error if the conversion not possible?

hoodmane:
> Maybe we can add an option to tell the to_py()/toJs() conversion to throw error if the conversion not possible?

Yeah sounds good. The hardest part would be naming the option.

oeway:
Maybe something like raise_conversion_error?

hoodmane:
[...]

PR #2138, opened 2022-01-27
MAINT refactor js2python to allow future options

hoodmane:
I moved the `depth` and `cache` into a `context` object argument which can be extended with more stuff in the future (e.g., something like #2067).

I also removed the underscore prefixes from most methods. There are a few reasons for this:

1. In Javascript, methods are private by default. They must be explicitly exported to make them public. Thus, using an underscore to denote private in Javascript is slightly controversial.
2. When invoking C functions from Javascript, they get an underscore prefix. It's better to reserve the underscore to denote "this is a C function" because that's more relevant than public vs private. Especially because almost everything in `src/core` is intended to be private to the downstream user.
3. No other Javascript functions are defined in `src/core` with an underscore prefix so it's not consistent with our own usage.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -178,7 +178,7 @@ clean-all:
 	make -C cpython clean-all
 
 
-%.o: %.c $(CPYTHONLIB) $(wildcard src/core/*.h src/core/python2js_buffer.js)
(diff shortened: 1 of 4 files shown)


ryanking13 (approved):
Thanks, @hoodmane. LGTM.

hoodmane:

PR #2139, opened 2022-01-27
MAINT Reorder make lint to typecheck first before running formatters

hoodmane:
We should run mypy before black. Fixing mypy type errors can mess up formatting, but fixing formatting can't mess up type errors. The current process looks like:
1. format error, run formatter
2. type errors, fix type errors
3. fixing type errors introduced new format errors, run formatter again

If we check types first then the process looks instead like:
1. type errors, fix type errors
2. format errors, run formatter

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -138,11 +138,6 @@ update_base_url: \
 lint: node_modules/.installed
 	# check for unused imports, the rest is done by black
 	flake8 --select=F401 src tools pyodide-build benchmark conftest.py docs packages/matplotlib/src/
-	find src -type f -regex '.*\.\(c\|h\)' \
-		| xargs clang-format-6.0 -output-replacements-xml \
-		| (! grep '<replacement ')
-	npx prettier --check .
-	black --check .
 	mypy --ignore-missing-imports    \
 		pyodide-build/pyodide_build/ \
 		src/ 					     \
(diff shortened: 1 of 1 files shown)


rth:
I would be in favor of removing that Make command altogether in favor of using pre-commit, which we can also run in CI to check.

hoodmane:
It would probably also be good to type check the typescript. In pre-commit, we should probably only do that if `pyproxy.gen.ts` exists (or just skip it entirely), in CI we can `cp src/core/pyproxy.ts src/js/pyproxy.gen.ts` and then type check typescript.

Also, I've turned off pre-commit because mypy takes a very long time.

ryanking13:
I wasn't using `pre-commit` before (just because  I'm used to running `make lint`), but I think it is a good idea to make a command altogether.
Maybe, we could also use [pre-commit.ci](https://pre-commit.ci/) to automatically fix minor formatting errors, which could help contributors who forgot formatting before PR. Though I'm not sure whether it will work well with mypy.

hoodmane:
> Though I'm not sure whether it will work well with mypy.

Yeah, right this is really for formatters, you can't auto fix type errors. I think the biggest improvement would be if we could automatically fix trailing spaces and files that are missing newlines. We particularly have trouble when people use `[skip ci]` for documentation changes, which can lead to us merging formatting problems.

hoodmane:
Could we merge this for now and have someone else make a better solution later? I think this is an improvement even if it isn't a long term solution.

ryanking13 (approved):
I'm +1 with merging this.

hoodmane:
Okay, I'll go ahead and merge it then.

PR #2142, opened 2022-01-28
Handle KeyboardInterrupt in event loop

### LINKED ISSUE #2141: runPythonAsync never returns when interrupted during await.
## 🐛 Bug

If an interrupt buffer is used to interrupt pyodide while pthon is awaiting some function, an exception is thrown in the internals of WebLoop, and the awaiting code never returns. If runPythonAsync was used to run the python code, the promise it returns never resolves.

### To Reproduce

A more complete, runnable reproduction can be found here: https://jsfiddle.net/jhb4d8s1/50/

```js
let interruptBuffer = new Uint8Array(new SharedArrayBuffer(1));
pyodide.setInterruptBuffer(interruptBuffer);

// Run some async python code
const result = pyodide.runPythonAsync(`
  import asyncio
  
  # If an interrupt happens here, `result` never resolves.
  asyncio.sleep(10)
`);

//wait some amount of time, then call
interruptBuffer[0] = 2

await result;
console.log("This line is never run");
```

### Expected behavior

```python
asyncio.sleep(10)
```
Should throw a KeyboardInterrupt exception.

### Environment

- Pyodide Version: 0.19.0
- Browser version: 97.0.4692.99

### Additional context

Calling `checkInterrupt()` from JS periodically does not resolve the issue. It does throw the exception much earlier, but the result is the same.

The exception that is thrown is different though.

With `checkInterrupt()`:
```
Uncaught PythonError: Traceback (most recent call last):
  File \&quot;/lib/python3.9/site-packages/_pyodide/_base.py\&quot;, line 352, in eval_code
    def eval_code(
KeyboardInterrupt
```

Without `checkInterrupt()`:
```
pyodide.asm.js:14 Uncaught PythonError: Traceback (most recent call last):
  File "/lib/python3.9/site-packages/pyodide/webloop.py", line 148, in run_handle
    if h.cancelled():
  File "/lib/python3.9/asyncio/events.py", line 75, in cancelled
    def cancelled(self):
KeyboardInterrupt

    at new_error (pyodide.asm.js:14:247183)
    at pyodide.asm.wasm:0xe6eb0
    at pyodide.asm.wasm:0xe6fae
    at Object.Module._pythonexc2js (pyodide.asm.js:14:944230)
    at Object.Module.callPyObjectKwargs (pyproxy.gen.js:391:12)
    at Object.Module.callPyObject (pyproxy.gen.js:397:17)
    at wrapper (pyodide.asm.js:14:215278)
```

hoodmane:
Resolves #2141 by making a keyboard interrupt cancel *all* ongoing event loop stuff.

It is possible to add the option to mark tasks as not keyboard interruptible, if people want background tasks that won't be interrupted by user interaction.

### Checklists

- [ ] Add / update tests


hoodmane:
I think to make this less of a breaking change we should make it opt-in rather than opt-out...

src/core/keyboard_interrupt.c
```diff
@@ -1,29 +1,86 @@
...
 #include <emscripten.h>
 
 static int callback_clock = 50;
+_Py_IDENTIFIER(get_event_loop);
+_Py_IDENTIFIER(set_interrupt);
 
+/**
+ * Calls WebLoop.handle_interrupt.
+ * WebLoop.handle_interrupt raises KeyboardInterrupt into all Tasks that are
+ * marked as interruptable.
```
rth, line 15: What's the practical effect of this? Shouldn't the KeyboardInterrupt exception be raised only from the active task since only one exception can be reported to the user anyway?

rth (commented):
Thanks for working on this! 

A general question: 

> This PR adds a context variable is_interruptable such that
> whenever a keyboard interrupt occurs, all tasks with is_interruptable set true have a KeyboardInterrupt raised into them

What is fundamentally different between our event loop implementation and the default one in asyncio to need this mechanism? I mean base asyncio is able to handle KeyboardInterrupt: what makes our use case different?

hoodmane:
> What is fundamentally different between our event loop implementation and the default one in asyncio to need this mechanism? I mean base asyncio is able to handle KeyboardInterrupt: what makes our use case different?

Yeah this is certainly the right question. Consider the following code:
```py
import asyncio
async def test():
   await asyncio.gather(f(),g())
   await asyncio.sleep(1)
   print("done")

async def f():
   await asyncio.sleep(1)
   print("f!")

async def g():
   await asyncio.sleep(2)
   print("g!")

asyncio.get_event_loop().run_until_complete(test())
```
First consider the normal Python case. If you do a keyboard interrupt, with very high probability no task will be running and you will see a stack trace like this:
```
  File "<stdin>", line 1, in <module>
  File "/usr/lib/python3.9/asyncio/base_events.py", line 629, in run_until_complete
    self.run_forever()
  File "/usr/lib/python3.9/asyncio/base_events.py", line 596, in run_forever
    self._run_once()
  File "/usr/lib/python3.9/asyncio/base_events.py", line 1854, in _run_once
    event_list = self._selector.select(timeout)
  File "/usr/lib/python3.9/selectors.py", line 469, in select
    fd_event_list = self._selector.poll(timeout, max_ev)
KeyboardInterrupt
```
This cancels the running of the event loop, but it has the same set of tasks scheduled as before. If you resume the event loop, it will print `f!\ng!\ndone\n`.

If you time the keyboard interrupt very accurately, you might manage to hit one of the function bodies, and get a traceback like:
```
  File "<stdin>", line 1, in <module>
  File "/usr/lib/python3.9/asyncio/base_events.py", line 629, in run_until_complete
    self.run_forever()
  File "/usr/lib/python3.9/asyncio/base_events.py", line 596, in run_forever
    self._run_once()
  File "/usr/lib/python3.9/asyncio/base_events.py", line 1890, in _run_once
    handle._run()
  File "/usr/lib/python3.9/asyncio/events.py", line 80, in _run
    self._context.run(self._callback, *self._args)
  File "<stdin>", line 8, in f
KeyboardInterrupt
```
This will actually cancel a task, in this case it cancels f and test. If you resume the event loop, it will print `g!\n`.

This behavior isn't necessarily super helpful. You could always hold on to the Tasks and call `task.cancel()`

Now consider the case in Pyodide (but not with this PR). Our browser event loop is always running but it is implemented in Javascript so the `KeyboardInterrupt` can't ever interrupt it. Suppose we run the same code. What will happen when a keyboard interrupt is signaled? Well you are never going to get the first type of keyboard interrupt that interrupts the event loop itself. It will kill one of f, g, or test, depending on timing. If it kills `f` it will print `g!\n`, if it kills `g` it will print `f!\n` and if it kills `test` it will print `f!\ng!\n`.

brendan0powers:
Hello, I'm wondering if there's anything I can do to help with this PR? This is pretty important to my use case. Thanks!

hoodmane:
I think this is almost right, but I ran out of stream on it before I got around to fixing the issues.

* We need to get the tests passing. 
* IIRC, various tests by hand seemed to show that this doesn't work as well as I would hope. The response to an interrupt can be different in counterintuitive ways depending on precise timings. Either my model of the system or the implementation is missing some small detail.
* The up-streaming of the keyboard interrupt code adds new constraints #2332. This PR either requires a redesign to fit around the new constraints or a new CPython patch.

I think the reason this is so hard is that the interrupt state is global state, not task local state. But we are trying to implement task-local stuff on top of it. We need very careful control over when interrupts are handled so that the interrupt doesn't happen in the event loop code itself. The goal is to trap interrupts and then decide which tasks should be interrupted and then raise interrupts into each of the tasks.

It's particularly likely for the interrupt to trigger in the setup code just before entering the task itself and so interrupt the event loop methods rather than the task (and then the task is left dangling never completed or in some other weird state). There might be some mistakes where little pieces of the event loop code are not properly protected. We also need to be very careful not to leak the handle or other PyProxies that are created in the process of maintaining loop state. We use "atomic" uninterruptible C and Javascript code to set the state as noninterruptible, do whatever set up we need, then reenable interrupts before entering the task frame so the interrupt is triggered in the proper spot.

PR #2144, opened 2022-01-29
MAINT Separate some things out of Module

hoodmane:
On main, there are a total of 7999 different attributes on `Module`. This moves ~20 things off `Module` onto `API`. Hopefully in the future we can add a bit more organization to things.

I left things that are both defined in `core` and only used in `core` on `Module`, so now when `Module` is used in `src/js` it is actually calling Emscripten APIs or functions we defined in C.


rth (approved):
Much better to have a separate namespace, thanks!

PR #2146, opened 2022-01-30
Fix typo

paulftw:
### Description
I'm very new to pyodide but I think this is what the folder name should be. Folder `package/` doesn't exist when I follow the docker instructions

### Checklists


rth (approved):
Indeed, thanks @paulftw !

PR #2152, opened 2022-02-01
Update Dockerfile to install latest browser versions

ryanking13:
### Description

- Bump selenium version to 4.1.0
- Update Dockerfile to specify the version of browsers to be installed
  - default is to install the latest stable versions, which is for now:

```
- Google Chrome 97.0.4692.99
- Mozilla Firefox 96.0.3
- ChromeDriver 97.0.4692.71 (adefa7837d02a07a604c1e6eff0b3a09422ab88d-refs/branch-heads/4692@{#1247})
- geckodriver 0.30.0 (d372710b98a6 2021-09-16 10:29 +0300)
```

### DIFF
--- a/.github/workflows/docker_image.yml
+++ b/.github/workflows/docker_image.yml
@@ -17,8 +17,6 @@ on:
 env:
   GHCR_REGISTRY: ghcr.io
   IMAGE_NAME: pyodide/pyodide-env
-  CHROME_VERSION: ${{ github.event.inputs.chrome_version }}
-  FIREFOX_VERSION: ${{ github.event.inputs.firefox_version }}
 jobs:
   build_docker:
     runs-on: ubuntu-latest
@@ -44,6 +42,9 @@ jobs:
           push: true
           tags: ${{ env.IMAGE_NAME }}:${{ github.event.inputs.version }}
           labels: ${{ steps.meta.outputs.labels }}
+          build-args: |
+            CHROME_VERSION=${{ github.event.inputs.chrome_version }}
+            FIREFOX_VERSION=${{ github.event.inputs.firefox_version }}
       - name: Image digest
         run: echo ${{ steps.build.outputs.digest }}
   build_ghcr:
@@ -70,5 +71,8 @@ jobs:
           push: true
           tags: ${{ env.GHCR_REGISTRY }}/${{ env.IMAGE_NAME }}:${{ github.event.inputs.version }}
           labels: ${{ steps.meta.outputs.labels }}
+          build-args: |
+            CHROME_VERSION=${{ github.event.inputs.chrome_version }}
+            FIREFOX_VERSION=${{ github.event.inputs.firefox_version }}
       - name: Image digest
         run: echo ${{ steps.build.outputs.digest }}
--- a/.github/workflows/docker_image.yml
+++ b/.github/workflows/docker_image.yml
@@ -81,5 +81,10 @@ jobs:
           push: true
           tags: ${{ env.GHCR_REGISTRY }}/${{ env.IMAGE_NAME }}:${{ github.event.inputs.version }}
           labels: ${{ steps.meta.outputs.labels }}
+          build-args: |
+            CHROME_VERSION=${{ github.event.inputs.chrome_version }}
+            CHROME_DRIVER_VERSION=${{ github.event.inputs.chrome_driver_version }}
+            FIREFOX_VERSION=${{ github.event.inputs.firefox_version }}
+            GECKODRIVER_VERSION=${{ github.event.inputs.firefox_driver_version }}
       - name: Image digest
         run: echo ${{ steps.build.outputs.digest }}
--- a/.github/workflows/docker_image.yml
+++ b/.github/workflows/docker_image.yml
@@ -6,11 +6,11 @@ on:
         description: "Version of the docker image to build."
         required: true
       chrome_version:
-        description: "Version of the chrome browser (e.g. 98.0.4758.9-1)"
+        description: "Version of the chrome browser (e.g. 97)."
         required: false
         default: "latest"
       firefox_version:
-        description: "Version of the firefox browser. (e.g. 96.0)"
+        description: "Version of the firefox browser (e.g. 96.0)."
         required: false
         default: "latest"
 
--- a/.github/workflows/docker_image.yml
+++ b/.github/workflows/docker_image.yml
@@ -6,21 +6,13 @@ on:
         description: "Version of the docker image to build."
         required: true
       chrome_version:
-        description: "Version of the chrome browser."
-        required: false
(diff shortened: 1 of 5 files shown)


ryanking13:
Chrome `fater_error` test is failing with the message:

```
FAILED test_pyproxy.py::fatal_error[chrome] - selenium.common.exceptions.WebDriverException: Message: unknown error: session deleted because of page crash
ERROR test_pyproxy.py::fatal_error[chrome] - selenium.common.exceptions.InvalidSessionIdException: Message: invalid session id
```

I don't think it is due to the memory size, I can't understand how a fatal error can crash the page.

hoodmane:
> trying to raise memory access out of bounds error?

Exactly. Feel free to delete it if it's causing trouble!

ryanking13:
> > trying to raise memory access out of bounds error?
> 
> Exactly. Feel free to delete it if it's causing trouble!

Okay, I temporarily disabled the crashing test.

.github/workflows/docker_image.yml
```diff
@@ -5,6 +5,23 @@ on:
...
+        required: false
+        default: ""
+      firefox_version:
+        description: "Version of the firefox browser."
+        required: false
+        default: "latest"
+      firefox_driver_version:
+        description: "Version of the firefox geckodriver."
+        required: false
+        default: "0.30.0"
```
rth, line 23: I think it would be better hardcode this in the Dockerfile, because otherwise it's hard to know which combination was used to generate the docker image. Also there is normally a compatibility range between the browser and its driver which is a bit tricky.

Dockerfile
```diff
@@ -8,25 +8,87 @@ RUN apt-get update \
...
-# Get recent version of chromedriver
-RUN wget --quiet https://chromedriver.storage.googleapis.com/90.0.4430.24/chromedriver_linux64.zip \
-  && unzip chromedriver_linux64.zip \
-  && mv $PWD/chromedriver /usr/local/bin \
-  && rm -f chromedriver_linux64.zip
+ARG CHROME_VERSION="google-chrome-stable"
+ARG CHROME_DRIVER_VERSION
+ARG FIREFOX_VERSION="latest"
+# Note: default geckodriver version needs to be updated manually
+ARG GECKODRIVER_VERSION="0.30.0"
```
rth, line 25: Related to the previous comment, maybe let's move them to ENV ? That way they are still defined separately but users don't need to pass 4 args if they want to rebuild the docker image.

rth (commented):
Very nice! Could you re-start the failed CI? Thanks!

.github/workflows/docker_image.yml
ryanking13: Alright, I'll remove them from the yaml file.

ryanking13:
> Very nice! Could you re-start the failed CI? Thanks!

It looks like the sha256 checksum of statsmodels has been changed mysteriously 🤔

rth:
> I'm really excited to get away from Chrome v90.

BTW, I finally figured out how to visualize browser versions in Google Analytics for our documentation (let me know if you want access), which gives some idea of what browsers and versions people use. For the last month, we get the following distribution: 62%  Chrome, 17% Safari 10% Firefox and 6% Edge. Also in terms of browser versions, they are pretty recent (cf table below). Though this might be biased toward developers, not necessarily end-users who often would not be reading Pyodide documentation. 

We might want to test [our minimal requirements](https://pyodide.org/en/stable/usage/index.html#supported-browsers) (which need updating), and say Chrome latest, rather than the latest browser versions.

For instance, maybe @casatir  would have some browser version stats for Basthon users.

image

ryanking13:
Update:

- It looks like google chrome apt registry doesn't support installing older versions of Chrome. So I changed it to directly download the `.deb` file.
- Formatting is falling due to the `black` version update (`22.1.0`). Since we will only update Dockerfile (and related stuffs) in this PR, and update the actual Docker image used for CI in a separate PR, it should be okay.

ryanking13:
> BTW, I finally figured out how to visualize browser versions in Google Analytics for our documentation (let me know if you want access), which gives some idea of what browsers and versions people use. For the last month, we get the following distribution: 62% Chrome, 17% Safari 10% Firefox and 6% Edge. Also in terms of browser versions, they are pretty recent (cf table below). Though this might be biased toward developers, not necessarily end-users who often would not be reading Pyodide documentation.

Updated GH actions workflow to choose chrome and firefox browser versions when deploying. I think we should specify browser versions in the tag. For example, I published the docker image in this PR with tag [20-chrome97-firefox96](https://hub.docker.com/layers/sihadan/pyodide-env-test/20-chrome97-firefox96/images/sha256-7b8928e44d848778d8f0e7c7ece8bc18d8fa7ab2b9a7d59e9eb3bbe48284587d?context=explore).

image

> 
> We might want to test [our minimal requirements](https://pyodide.org/en/stable/usage/index.html#supported-browsers) (which need updating), and say Chrome latest, rather than the latest browser versions.

Yeah, we should decide which versions of browsers would be used for CI.

Dockerfile
ryanking13: I guess ARG with a proper default value would be fine. Users don't have to pass args if they don't want to specify the browser version they want to use.

.github/workflows/docker_image.yml
ryanking13: I removed chromedriver and geckodriver from it but kept chrome and firefox version, so that we can update browser versions in the docker image without updating the Dockerfile. It is possible that geckodriver might have a compatibility issue when firefox browser version updates, not sure how we can avoid it. (for chromedriver, I managed to automatically detect valid chrome-chromedriver version set.)

> because otherwise it's hard to know which combination was used to generate the docker image.

I think we could solve this issue by adding browser versions in the docker image tag.

ryanking13:
Okay, this is ready to be reviewed.

Deploying docker image [can be tested in my fork](https://github.com/ryanking13/pyodide/actions/workflows/docker_image.yml) (It will be published to [sihadan/pyodide-env-test](https://hub.docker.com/r/sihadan/pyodide-env-test/tags), note that I only setup docker hub token, so publishing to GHCR would fail in my fork)

rth:
There is just a linting issue, otherwise LGTM. Thanks for doing this work!

Let's add "Latest" for now, and we can generate another docker image with minimal supported versions (TBD) and add a core test for those under GH actions.

I'm not 100% sure how we should update docker images with our new CI setup. Should we merge this without bumping the image used in CI, then deploy the docker image and then finally have another PR updating the docker image?

ryanking13:
> I'm not 100% sure how we should update docker images with our new CI setup. Should we merge this without bumping the image used in CI, then deploy the docker image and then finally have another PR updating the docker image?

Exactly, I made a PR for that (#2160). The linting issue is related to the image update, so I'll handle it there.

PR #2153, opened 2022-02-01
Update the statsmodels URL

ryanking13:
It looks like the sha256 checksum of statsmodels has been changed for a mysterious reason.

### DIFF
--- a/packages/statsmodels/meta.yaml
+++ b/packages/statsmodels/meta.yaml
@@ -3,8 +3,8 @@ package:
   version: 0.13.1
 
 source:
-  url: https://github.com/statsmodels/statsmodels/archive/refs/tags/v0.13.1.tar.gz
-  sha256: 428a60ade157172dc53a9987b134acc721dc15dcdf26ffd93af128157c0f8055
+  url: https://files.pythonhosted.org/packages/e7/86/8c95a2f43d8d66837f52fc0a2d9b4ea491e564789ee94d28f642d9d47ebc/statsmodels-0.13.1.tar.gz
+  sha256: 006ec8d896d238873af8178d5475203844f2c391194ed8d42ddac37f5ff77a69
   patches:
     - patches/fix-scipy-blas-cythonize.patch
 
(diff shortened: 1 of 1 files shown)


ryanking13:
Well, now zlib is also failing due to the checksum.
Maybe github has changed something in generating archives from the release, resulting in the change of checksum.

rth:
That's pretty crazy that they don't maintain backward compatibility on this. Also apparently not the first time https://github.com/spack/spack/issues/5411. Maybe let's wait a bit more in case it's an unintentional bug on their end but yes otherwise this gives another reason not to rely on GH hosted releases.

packages/statsmodels/meta.yaml
```diff
@@ -4,7 +4,7 @@ package:
 
 source:
   url: https://github.com/statsmodels/statsmodels/archive/refs/tags/v0.13.1.tar.gz
-  sha256: 4bf899994549079aa91a6ee5e06f4c900b7eba22af6d2f27a26ab58dd4a7ccac
```
rth, line 7: Let's use PyPi for this which guarantees the checksum to be immutable.

https://pypi.org/project/statsmodels/0.13.1/#files

For emscripten ports, that's going to be more problematic.

hoodmane:
Can we temporarily disable shasum checks for github stuff then? I'd really like to get #2152 merged...

ryanking13:
> Also, @ryanking13 if you setup CircleCi on your fork, could you please add me and Hood as contributors there? Otherwise, we can't restart CircleCi on your PRs.

Done!

> Can we temporarily disable shasum checks for github stuff then? I'd really like to get #2152 merged...

github devs say that they rolled back the change, let's see what happens now.

ryanking13:
Updated statsmodels' URL to PyPI, so packages are now fine. (Emscripten ports are still using GitHub URLs though)

ryanking13:
Okay this is working fine, I'll go and merge. We could patch emscripten URLs later if it causes trouble again.

PR #2160, opened 2022-02-09
CI Update to newer docker image

### LINKED ISSUE #1843: Simplify browser updates in CI
As mentionned by @hoodmane in https://github.com/pyodide/pyodide/pull/1838#issuecomment-920118708

> It's a bit surprising to me that upgrading the selenium browser to the latest version of Chrome / Firefox is so tricky -- would think everyone wants to do this.
I wonder whether we could copy the code here:
https://github.com/SeleniumHQ/docker-selenium/blob/trunk/NodeChrome/Dockerfile

ryanking13:
Updates CI image, continuing from #2152.

Note that this PR updates Python formatting due to the black version update.
(A good news is that black is now in [a stable state](https://github.com/psf/black/releases/tag/22.1.0), so there will be fewer formatting issues in the future)

Resolve #1843


ryanking13:
> selenium.common.exceptions.WebDriverException: Message: unknown error: session deleted because of page crash

The sad news is that we are still facing this error even after the browser update...

ryanking13:
I'm using the tag [latest](https://hub.docker.com/r/pyodide/pyodide-env/tags), as discussed in #2152, open to better tag naming conventions.

rth (approved):
Thanks @ryanking13 !

hoodmane:
I am a bit concerned about using the latest tag because it means that old commits won't be fully reproducible. With the old system, if we go back to a commit that used e.g., `pyodide-env:17` then we can build with that commit. Whereas `pyodide-env:latest` may eventually change in ways that break older builds. For instance, we already have the example where we updated `black` in place on top of `pyodide-env:19` so if you rerun CI on older commits, their lint stage will fail. 

This is particularly concerning if we ever need to `git bisect` a regression: if old commits don't reliably build, then `git bisect` won't work.

ryanking13:
> 

Yes, I agree that we should keep the old docker image for cases when we want to build older versions of Pyodide (and for other users who use older versions of Pyodide).

I think we could,

1. keep the `latest` tag for CI.
2. release additional Docker image with other tag when we update the docker image, so that older versions (commits) of Pyodide can be built with that image. (We will need a proper tag naming convention for that)

rth:
Good point. +1 to have the explicit version in CI instead of latest. It's a minimal effort to update it when needed, and it's would allow us to track it more explicitly in CI. Even if we release a separate image with the explicit version (while using latest), it would still be difficult to know which version is used if it's not in version control.

ryanking13:
> Even if we release a separate image with the explicit version (while using latest), it would still be difficult to know which version is used if it's not in version control.

Yeah, that's true. I agree that we should go with the explicit version.

I think we need to specify 2 things for versioning docker images.

1. the date of releasing the image
1. browser versions _(+ node version?)_

For example, `pyodide-env:0.20.0-20220211-chrome96-firefox96`.

Or without pyodide version, `pyodide-env:20220211-chrome96-firefox96`.

PR #2168, opened 2022-02-12
CI Improve docker tagging rules

ryanking13:
From the discussion at #2160, this PR does two things:

- Update docker image tagging rules
  - If tag is not set, it will be in default: `<date>-chrome[version]-firefox[version]` (e.g. 20220212-chrome97-firefox96)
  - Open to other tag naming rules
 
- Introduce a new config file `.docker-env-version`, which specifies the Docker image used for:
  - `run_docker` script
  - `prebuilt_docker_image.yml` GH actions workflow

When we release a new docker image, both of `.circleci/config.yml` and `.docker-env-version` need to be updated.
I want to make it to one, but not sure how it can be done. Dynamically setting a docker image used for CI seems like nonsense.

__TODO__

- [x] Update docker image tag in config files


.github/workflows/prebuilt_docker_image.yml
```diff
@@ -52,14 +51,17 @@ jobs:
...
+        run: echo "::set-output name=tag::$(cat .docker-env-version | tr -d " \t\n\r")"
       - name: Build and push Docker image to ${{ env.GHCR_REGISTRY }}
         id: build
         uses: docker/build-push-action@v2
         with:
           file: ./Dockerfile-prebuilt
           push: true
           build-args: |
-            VERSION=${{ env.PYODIDE_ENV_VERSION }}
+            VERSION=${{ steps.pyodide_env.outputs.tag }}
```
rth, line 64: This line and the change above needs updating as well. Thanks!

ryanking13: Oops, thanks!

PR #2169, opened 2022-02-12
Pkg: openPMD-api

ax3l:
### Description

Add openPMD-api, a Python package for scientific/engineering data.

### Checklists
- [ ] Add / update tests


packages/openpmd-api/meta.yaml
```diff
@@ -0,0 +1,39 @@
...
+    - python
+    - numpy
+    - libhdf5
+
+test:
+  imports:
+    - openpmd_api
+
+build:
+  library: true
```
ax3l, line 20: Help wanted: Not sure what that means :)

This is a C++ library with Python bindings as a module. The primary need here is to use the Python module.

rth: See https://pyodide.org/en/latest/development/meta-yaml.html#build-library

If the package doesn't have a  `setup.py`  you would want to set it to True.

ax3l: Thanks, ok!
I do have a `setup.py` for pip-based installing, but it does not export all the extra CMake flags that we need for the webassembly cross-compile here :)

packages/openpmd-api/meta.yaml
```diff
@@ -0,0 +1,45 @@
...
+      -DCMAKE_CROSSCOMPILING_EMULATOR="$PYODIDE_ROOT/emsdk/emsdk/node/14.18.2_64bit/bin/node" \
+      -DCMAKE_BUILD_TYPE=Release \
+      -DopenPMD_BUILD_SHARED_LIBS=OFF \
+      -DopenPMD_USE_MPI=OFF      \
+      -DopenPMD_USE_HDF5=OFF     \
+      -DopenPMD_USE_ADIOS1=OFF   \
+      -DopenPMD_USE_ADIOS2=OFF   \
+      -DopenPMD_USE_PYTHON=ON    \
+      -DPython_EXECUTABLE:FILEPATH=$(which python3) \
+      -DPython_INCLUDE_DIR=$(python3 -c "from sysconfig import get_paths as gp; print(gp()['include'])") \
```
ax3l, line 39: I think I am not yet pointing to the right, cross-compiled Python yet:
```
In file included from /src/packages/.artifacts/pybind11/include/pybind11/detail/common.h:215:
In file included from /usr/local/include/python3.9/Python.h:63:
/usr/local/include/python3.9/pyport.h:741:2: error: "LONG_BIT definition appears wrong for platform (bad gcc/glibc config?)."
#error "LONG_BIT definition appears wrong for platform (bad gcc/glibc config?)."
 ^
1 error generated.
```

packages/openpmd-api/meta.yaml
hoodmane: Well to be clear, you are only supposed to set this to be true if your package is a static library intended to be linked by some particular other package. It isn't appropriate for a Python package.

hoodmane:
```suggestion
```

packages/openpmd-api/meta.yaml
hoodmane, line 39: I think this should be:
```suggestion
      -DPython_INCLUDE_DIR=$PYODIDE_ROOT/cpython/installs/python-3.9.5/include/python3.9/ \
```

packages/openpmd-api/meta.yaml
```diff
@@ -0,0 +1,45 @@
...
+      -Dpybind11_DIR=$PYODIDE_ROOT/packages/.artifacts/pybind11/share/cmake/pybind11 \
+      -DopenPMD_USE_INTERNAL_PYBIND11=OFF \
+      -DopenPMD_BUILD_TESTING=OFF         \
+      -DopenPMD_BUILD_EXAMPLES=OFF        \
+      -DCMAKE_INSTALL_LIBDIR=lib
+    emmake cmake --build build -j ${PYODIDE_JOBS:-3} --target install
```
hoodmane, line 45:
```suggestion
    emmake cmake --build build -j ${PYODIDE_JOBS:-3}
```

hoodmane:
Okay well I think I got this to build manually. You have to add `-DCMAKE_CXX_FLAGS=-mmutable-globals \`.
Then after running the script here, you can link the shared library with:
```sh
    emcc -s SIDE_MODULE=1 -o openpmd_api_cxx.so
      CMakeFiles/openPMD.py.dir/src/binding/python/openPMD.cpp.o
      CMakeFiles/openPMD.py.dir/src/binding/python/Access.cpp.o
      CMakeFiles/openPMD.py.dir/src/binding/python/Attributable.cpp.o
      CMakeFiles/openPMD.py.dir/src/binding/python/BaseRecord.cpp.o
      CMakeFiles/openPMD.py.dir/src/binding/python/BaseRecordComponent.cpp.o
      CMakeFiles/openPMD.py.dir/src/binding/python/ChunkInfo.cpp.o
      CMakeFiles/openPMD.py.dir/src/binding/python/Container.cpp.o
      CMakeFiles/openPMD.py.dir/src/binding/python/Dataset.cpp.o
      CMakeFiles/openPMD.py.dir/src/binding/python/Datatype.cpp.o
      CMakeFiles/openPMD.py.dir/src/binding/python/Helper.cpp.o
      CMakeFiles/openPMD.py.dir/src/binding/python/Iteration.cpp.o
      CMakeFiles/openPMD.py.dir/src/binding/python/IterationEncoding.cpp.o
      CMakeFiles/openPMD.py.dir/src/binding/python/Mesh.cpp.o
      CMakeFiles/openPMD.py.dir/src/binding/python/ParticlePatches.cpp.o
      CMakeFiles/openPMD.py.dir/src/binding/python/ParticleSpecies.cpp.o
      CMakeFiles/openPMD.py.dir/src/binding/python/PatchRecord.cpp.o
      CMakeFiles/openPMD.py.dir/src/binding/python/PatchRecordComponent.cpp.o
      CMakeFiles/openPMD.py.dir/src/binding/python/Record.cpp.o
      CMakeFiles/openPMD.py.dir/src/binding/python/RecordComponent.cpp.o
      CMakeFiles/openPMD.py.dir/src/binding/python/MeshRecordComponent.cpp.o
      CMakeFiles/openPMD.py.dir/src/binding/python/Series.cpp.o
      CMakeFiles/openPMD.py.dir/src/binding/python/UnitDimension.cpp.o
      lib/libopenPMD.a
```
then if you make a wheel with the `.py` files and `openpmd_api_cxx.so` it will successfully compile the wasm module `openpmd_api_cxx.so` and import is successful.

ax3l:
Oh awesome, thank you so much for your help!
Had to leave the computer, but will add once I am back :)

hoodmane:
The second time it's running it from setup.py.

ax3l:
Oh, can I skip that? Otherwise I would run it solely through that and patch in the options we need to pass to CMake.

hoodmane:
I think the best is to run it via setup.py because otherwise it's hard to get the rest of the wheel made. Also, it's important that the generated .so file shouldn't have a platform tag in the file name saying it's x86_64 Linux because I think python won't try to import .so files with incompatible platform tags so it either needs our wasm32_emscripten tag or no tag. I'm not sure how cmake decided to name the file though.

hoodmane:
I've confirmed at least that it works if you name the file just `openpmd_api_cxx.so` and that it didn't work with the Linux platform tag.

packages/openpmd-api/meta.yaml
```diff
@@ -0,0 +1,39 @@
...
+    - openpmd_api
+
+build:
+  script: |
+    pip install -U pipx pybind11[global]
+    python -m pipx install cmake
+    export PATH="$(python3 -c 'import site; print(site.USER_BASE)')/bin:${PATH}"
+    pip install --upgrade pybind11[global] -t $PYODIDE_ROOT/packages/.artifacts/
+
+    export CMAKE_PREFIX_PATH=$PYODIDE_ROOT/packages/libhdf5/build/libhdf5-1.12.1/build/hdf5:$PYODIDE_ROOT/packages/zlib/build/zlib-1.2.11:${CMAKE_PREFIX_PATH}
```
ax3l, line 28: Now the last thing left is to actually find the _install_ root path of HDF5.
It looks like we cannot pick HDF5 up from its build directory directly.

ax3l:
X-ref: https://gitlab.kitware.com/cmake/cmake/-/merge_requests/6983

ax3l:
Now it just seems to forget to pick up the `.so` file we build when packing the wheel.

ax3l:
@hoodmane how did you get the built object in the wheel? :)

Usually, it setuptools just picks it up if I build it in the right [locations](https://github.com/openPMD/openPMD-api/blob/0.14.4/setup.py#L39-L50).

hoodmane:
Yeah there's something weird about that. I wonder why it isn't showing up in the right place. But you should be able to do something like:
```
build:
   ...
   post:
        cp build/path/to/so/file dist/openpmd_api-version#/openpmd_api
```
It may require some experimentation to figure out the right paths.

packages/openpmd-api/meta.yaml
```diff
@@ -0,0 +1,63 @@
...
+    export PATH="$(python3 -c 'import site; print(site.USER_BASE)')/bin:${PATH}"
+    pip install --upgrade pybind11[global] -t $PYODIDE_ROOT/packages/.artifacts/
+
+    export HDF5_ROOT=$PYODIDE_ROOT/packages/libhdf5/build/libhdf5-1.12.1/build/hdf5
+    export ZLIB_ROOT=$PYODIDE_ROOT/packages/zlib/build/zlib-1.2.11
+    export CMAKE_PREFIX_PATH=$HDF5_ROOT:$ZLIB_ROOT:${CMAKE_PREFIX_PATH}
+
+    cp $PYODIDE_ROOT/packages/openpmd-api/FindHDF5.cmake $PYODIDE_ROOT/.docker_home/.local/pipx/venvs/cmake/lib/python3.9/site-packages/cmake/data/share/cmake-3.22/Modules/FindHDF5.cmake
+
+    export EMSCRIPTEN=/src/emsdk/emsdk/upstream/emscripten
```
ryanking13, line 34:
```suggestion
    export EMSCRIPTEN=$PYODIDE_ROOT/emsdk/emsdk/upstream/emscripten
```

ryanking13:
> @hoodmane how did you get the built object in the wheel? :)
> 
> Usually, it setuptools just picks it up if I build it in the right [locations](https://github.com/openPMD/openPMD-api/blob/0.14.4/setup.py#L39-L50).

Yeah, I also sometimes face that issue in building other wheels. Not sure why setuptools doesn't pick the right file. It needs some investigation.

ax3l:
Could this be the logic behind it?
https://github.com/pypa/setuptools/blob/v60.9.0/setuptools/command/build_ext.py#L84-L104

ax3l:
Just to document my [naive tags question](https://github.com/pyodide/pyodide/pull/2169#issuecomment-1039652302), what would be the right tag for WebAsm `.so` Python modules?

ax3l:
Oh, I think the phase `running bdist_wheel` does not see the same environment variables as `running build_ext` in the build logic here.

I set `SETUPTOOLS_EXT_SUFFIX` and it's initially seen but then during packaging it is not anymore.

ax3l:
Hm, I cannot see with these fix applied why it does still not find the libs... uff

packages/openpmd-api/meta.yaml
```diff
@@ -0,0 +1,60 @@
...
+    python -m pipx install "cmake==3.22"
+    export PATH="$(python3 -c 'import site; print(site.USER_BASE)')/bin:${PATH}"
+    pip install --upgrade pybind11[global] -t $PYODIDE_ROOT/packages/.artifacts/
+
+    export HDF5_ROOT=$PYODIDE_ROOT/packages/libhdf5/build/libhdf5-1.12.1/build/hdf5
+    export ZLIB_ROOT=$PYODIDE_ROOT/packages/zlib/build/zlib-1.2.11
+    export CMAKE_PREFIX_PATH=$HDF5_ROOT:$ZLIB_ROOT:${CMAKE_PREFIX_PATH}
+
+    cp $PYODIDE_ROOT/packages/openpmd-api/FindHDF5.cmake \
+      ${HOME:-/src/.docker_home}/.local/pipx/venvs/cmake/lib/python3.9/site-packages/cmake/data/share/cmake-3.22/Modules/FindHDF5.cmake
```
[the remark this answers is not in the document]
ax3l: Could use `-t` as above for the `pip` install to simplify.

packages/openpmd-api/meta.yaml
```diff
@@ -0,0 +1,58 @@
...
+    pip install --upgrade pybind11[global] cmake==3.22 -t $PYODIDE_ROOT/packages/.artifacts/
+    export PATH="$PYODIDE_ROOT/packages/.artifacts/cmake/data/bin:${PATH}"
+    cp $PYODIDE_ROOT/packages/openpmd-api/FindHDF5.cmake \
+       $PYODIDE_ROOT/packages/.artifacts/cmake/data/share/cmake-3.22/Modules/FindHDF5.cmake
+
+    export HDF5_ROOT=$PYODIDE_ROOT/packages/libhdf5/build/libhdf5-1.12.1/build/hdf5
+    export ZLIB_ROOT=$PYODIDE_ROOT/packages/zlib/build/zlib-1.2.11
+    export CMAKE_PREFIX_PATH=$HDF5_ROOT:$ZLIB_ROOT:${CMAKE_PREFIX_PATH}
+
+    export EMSCRIPTEN=$PYODIDE_ROOT/emsdk/emsdk/upstream/emscripten
```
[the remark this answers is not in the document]
henryiii: There's also a `-C <cachefile>` option.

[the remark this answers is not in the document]
ax3l: Emscripten also sets a regular `EMSCRIPTEN` CMake variable with its toolchain.

ryanking13:

PR #2170, opened 2022-02-12
Add default_converter argument to to_py

hoodmane:
See the discussion with @oeway here: #2067.

This adds a new `default_converter` argument to `JsProxy.to_py` and `pyodide.toPy` which can be used to control the conversion of objects with no standard conversion. It can be used for instance to ensure that no `JsProxy`s are left over in the converted object, by just raising an error:
```py
def default_converter(_value, _ignored1, _ignored2):
    raise Exception("Object was not completely converted")
some_jsproxy.to_py(default_converter=default_converter)
```
If `default_converter` returns `None`, then `value` will be proxied. If you want to convert `Date` to `datetime` that can be done as follows:
```py
from datetime import datetime
def default_converter(value, _ignored1, _ignored2):
    if value.constructor.name == "Date":
        return datetime.fromtimestamp(d.valueOf()/1000)
```
If we have a pair class:
```js
class Pair {
    constructor(first, second){
        this.first = first;
        this.second = second;
    }
}
```
we can convert that as follows:
```py
def default_converter(value, convert, cache):
    if value.constructor.name != "Pair":
        return None
    result = []
    cache(value, result);
    result.append(convert(value.first))
    result.append(convert(value.second))
    return result
```
Note that this will handle self referential pairs correctly:
```js
const p = new Pair(0,0);
p.first = p;
```
will be converted to a list like:
```py
l = [0,0]
l[0] = l
```

### Checklist

- [x] Add / update tests

### DIFF
--- a/docs/project/changelog.md
+++ b/docs/project/changelog.md
@@ -46,6 +46,11 @@ substitutions:
 - {{Enhancement}} The Javascript package was migrated to Typescript.
   {pr}`2130` and {pr}`2133`
 
+- {{Enhancement}} Added a `default_converter` argument to {any}`JsProxy.to_py`
+  and {any}`pyodide.toPy` which is used to process any object that doesn't have
+  a built-in conversion to Python.
+  {pr}`2170`
+
 ## Version 0.19.0
 
 _January 10, 2021_
--- a/src/core/js2python.c
+++ b/src/core/js2python.c
@@ -57,8 +57,9 @@ EM_JS_REF(PyObject*, js2python, (JsRef id), {
  * Convert a JavaScript object to Python to a given depth. This is the
  * implementation of `toJs`.
  */
-EM_JS_REF(PyObject*, js2python_convert, (JsRef id, int depth), {
-  return Module.js2python_convert(id, depth);
+EM_JS_REF(PyObject*, js2python_convert, (JsRef id, int depth, JsRef default_converter), {
+  let defaultConverter = default_converter ? Module.hiwire.get_value(default_converter) : undefined;
+  return Module.js2python_convert(id, {depth, defaultConverter});
 });
 
 #include "include_js_file.h"
--- a/src/core/js2python.c
+++ b/src/core/js2python.c
@@ -57,10 +57,14 @@ EM_JS_REF(PyObject*, js2python, (JsRef id), {
  * Convert a JavaScript object to Python to a given depth. This is the
  * implementation of `toJs`.
  */
+// clang-format off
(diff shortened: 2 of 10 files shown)


oeway:
Great feature!

Regarding "return None", maybe it's more intuitive to just return  the value? Then, if we return None, it means we want to encode the value as None. It seems the default JSONEncoder in python does this too. Are you try to make sure a converter without explicit return to work by default?

hoodmane:
> Regarding "return None", maybe it's more intuitive to just return the value? Then, if we return None, it means we want to encode the value as None. It seems the default JSONEncoder in python does this too.

Yeah, that is a sensible suggestion.

ryanking13:
Thanks, @hoodmane! Generally LGTM, except one thing to ask.

```python
def default_converter(_value, _ignored1, _ignored2):
    raise Exception("Object was not completely converted")
some_jsproxy.to_py(default_converter=default_converter)
```

The example you've provided above and the name `default_converter` and is somewhat misleading to me.

If it's a "default" converter, I think it should have a priority to default conversion rules like Array to list or Map to dict.
Otherwise, if the default conversion rules always precede, we probably need other name then a default_converter.

WDYT?

hoodmane:
Well I can see how that's a bit confusing of a name but it is consistent with how the argument to `json.dumps` is named. Also I think default often means a fallback after trying other stuff.

Though rather than mimicking the `json.dumps` API another approach could be to mimick [`JSON.stringify`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON/stringify#the_replacer_parameter).

ryanking13:
> it is consistent with how the argument to `json.dumps` is named. Also I think default often means a fallback after trying other stuff.

Oh yes, that seems reasonable. Thanks for the explanation.

ryanking13 (approved):

hoodmane:
> wouldn't it be better to have a global setting somewhere

Maybe. I'm good with including something like that if someone else wants to do it but I'm more interested in getting the basic feature set first. In the meantime people can define a helper function with the appropriate settings.

hoodmane:
> So in practice what's the difference between, `some_jsproxy.to_py(default_converter=default_converter)` and `default_converter(some_jsproxy.to_py())` ?

Well the second thing only works on the outermost layer. This works for deep conversions (say you have an object where one of the values is a list of some class). It also can correctly handle reference loops -- see the last example I gave of a self referential `Pair`.

rth:
OK, it's clear, thanks for the explanation. +1 then to add it as an argument first, and we can always some global settings to be able to modify the default value.

PR #2171, opened 2022-02-12
Remove numpy patches work around for chrome 89 bug

hoodmane:
Now that we've finally updated our test version of chrome, we can remove these patches.


ryanking13:
Would this mean we will drop the support of old chrome versions?

hoodmane:
Yeah this drops numpy support for chrome 89 and 90. Older and newer chromes will still work.

ryanking13:
I'm fine with merging this, but we probably need to ask other pyodide-based-application devs about what would affect if we drop the support of chrome 89, 90. Maybe let's discuss this in a separate issue.

rth (approved):
Thanks! Could we add a note about those Chrome versions to https://pyodide.org/en/stable/usage/index.html#supported-browsers though?

ryanking13:
> Browser usage statistics show that chrome v89 and chrome v90 collectively add up to 0.8% of users, so I think it's not worth worrying about.

That's great! Maybe we could use that browser usage statistics as our policy to select minimal tested browser versions.

PR #2173, opened 2022-02-13
MAINT Add Hiwire to top level namespace

hoodmane:
We use `Module.hiwire` a total of 232 times, this switches all those to just `Hiwire`.


ryanking13 (approved):
Thanks, @hoodmane! This looks much better.

PR #2174, opened 2022-02-14
Add boost-histogram

henryiii:
### Description

This adds boost-histogram, one of three key compiled libraries in the
Scikit-HEP ecosystem (and like all three, useful beyond HEP). It was featured
at SciPy 2020. The other two packages (iMinuit and Awkward) use setuptools CMake in the build
process, so might be more challenging to add.

I wasn't able to run the build again after updating to 1.3.0, since on updating
pyodide I now get:

```
Error building pyparsing. Printing build logs.
/src/.venv/bin/python: No module named wheel
```

when trying to build. ~~Pretty sure I didn't do that.~~ Actually, I did, as I made a venv to put pyodide-build in.

See #2167.

### Checklists

- [x] Add / update tests


packages/boost-histogram/meta.yaml
```diff
@@ -0,0 +1,20 @@
...
+  sha256: 62240dae889cbf05f8b4f2203f0b7d6c7a2e029502723a377fc33226e425ce50
+requirements:
+  run:
+    - numpy
+build:
+  script: |
+    pip install setuptools_scm[toml]
+test:
+  imports:
+    - boost-histogram
```
henryiii, line 15:
```suggestion
    - boost_histogram
```

Will test locally tomorrow.

hoodmane: Well `boost-histogram` definitely isn't right here because `import boost-histogram` is a syntax error.

henryiii: Is there a way to run the tests locally? `make` seems to just build.

packages/boost-histogram/meta.yaml
```diff
@@ -0,0 +1,20 @@
+package:
+  name: boost-histogram
+  version: 1.3.0
+source:
+  url: https://files.pythonhosted.org/packages/0a/9a/1ebde600886eee433a99434e0ddb462630a0b25ab1f58efb91ffc385a0ff/boost_histogram-1.3.0.tar.gz
+  sha256: 62240dae889cbf05f8b4f2203f0b7d6c7a2e029502723a377fc33226e425ce50
+requirements:
+  run:
+    - numpy
```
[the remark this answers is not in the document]
hoodmane: Could you put a comment? @rth is working on separating run time from build time requirements in #2132

packages/boost-histogram/meta.yaml
hoodmane: Maybe:
```sh
PYODIDE_PACKAGES="boost-histogram" make -C packages
pytest -k "firefox and histogram" packages/test_packages_common.py
```

packages/boost-histogram/meta.yaml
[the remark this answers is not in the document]
henryiii:
```suggestion
    - numpy  # runtime only
``?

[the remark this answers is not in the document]
Comment (hoodmane) [reply packages/boost-histogram/meta.yaml:9]: Perfect.

File: packages/boost-histogram/test_boost_histogram.py [hunk 3]
```diff
@@ -0,0 +1,42 @@
+from pyodide_build.testing import run_in_pyodide
+
+
+@run_in_pyodide(packages=["boost-histogram"])
+def test_boost_histogram():
+    import boost_histogram as bh
+    import unittest
+
+    class TestBasicHistogram(unittest.TestCase):
```
hoodmane, line 9: I put similar code into test_wrapt.py, but as I'm looking at this I realized I'm not sure whether this actually runs the test. What does inheriting from `TestCase` do? Don't we also have to say to run the test somewhere?

henryiii: I'd guess that depends on how this is wrapped? I copied this out into a tmp.py file and ran `python -m unittest` on it, so I don't know if this actually can fail. I can break it then try the suggestion above.

hoodmane: Yeah I think you need `unittest.main()` at the bottom or it won't do the test.

henryiii: I _think_ this fails if you intentionally break the tests, though I get a weird error (TypeError: can't assign to property "pyodide_fatal_error" on 14527384: not an object) locally.

hoodmane: That uhh sounds like a bug. It would be nice to get a reproduction on that so we can fix it.

hoodmane: It sounds like we somehow confused a char* or PyObject* with a JavaScript object.

henryiii: I ran:

```
pytest -k "firefox and histogram" packages/boost-histogram/test_boost_histogram.py
```

And got:

```
_______________________________________________________________________________________________________________________________ test_boost_histogram[firefox] _______________________________________________________________________________________________________________________________
Error running function in pyodide

TypeError: can't assign to property "pyodide_fatal_error" on 11549448: not an object

API.fatal_error@http://127.0.0.1:54639/pyodide.js:15:39523
Module.callPyObjectKwargs@http://127.0.0.1:54639/pyodide.js:15:7005
Module.callPyObject@http://127.0.0.1:54639/pyodide.js:15:7208
apply@http://127.0.0.1:54639/pyodide.js:15:14354
apply@http://127.0.0.1:54639/pyodide.js:15:13439
runPython@http://127.0.0.1:54639/pyodide.js:15:32222
run@http://127.0.0.1:54639/test.html:4:45
@http://127.0.0.1:54639/test.html:7:40
@http://127.0.0.1:54639/test.html:25:15
@http://127.0.0.1:54639/test.html:27:8
```

henryiii:
```
TypeError: Cannot create property 'pyodide_fatal_error' on number '28811048'
```

There's a reproducer. ;)

hoodmane:
> There's a reproducer. ;)

Thanks!

packages/boost-histogram/test_boost_histogram.py
```diff
@@ -0,0 +1,35 @@
...
+    h = bh.Histogram(bh.axis.Integer(0, 10))
+    h.fill([1, 1, 1, 14])
+    assert h[bh.underflow] == 0
+    assert h[bh.loc(0)] == 0
+    assert h[bh.loc(1)] == 3
+    assert h[bh.overflow] == 1
+
+    assert h.sum() == 3
+    assert h.sum(flow=True) == 4
+    assert h[sum] == 4
```
hoodmane, line 17: Okay, that is... creative? I don't think I would have thought of this. The resulting behavior is very strange too. The number that is being thrown seems to be a memory address that is aligned to 8 bytes.

hoodmane: It is *not* a `PyObject*`, it is *not* a string. It seems to point to the four bytes `[76, -81, 9, 1]`, which is `17411916`. This is itself divisible by 4 (but not 8) and a reasonable size for a pointer. Looking at memory address `17411916`, it contains  6444 which is most likely a function pointer to `std::runtime_error::~runtime_error()`.

hoodmane: If that's the case, the number being thrown is a pointer to a pointer to a function pointer...

hoodmane: Okay, this is the result of an uncaught C++ exception:
```js
  __cxa_throw: function(ptr, type, destructor) {
#if EXCEPTION_DEBUG
    err('Compiled code throwing an exception, ' + [ptr,type,destructor]);
#endif
    var info = new ExceptionInfo(ptr);
    // Initialize ExceptionInfo content after it was allocated in __cxa_allocate_exception.
    info.init(type, destructor);
    exceptionLast = ptr;
    uncaughtExceptionCount++;
    throw ptr;
  }
```

hoodmane: The exception is being thrown inside of a function whose mangled name is 1014 bytes long:
```C++
pybind11::cpp_function::initialize<register_histogram<boost::histogram::storage_adaptor<std::__2::vector<double, std::__2::allocator<double> > > >(pybind11::module_&, char const*, char const*)::{lambda(boost::histogram::histogram<std::__2::vector<boost::histogram::axis::variant<boost::histogram::axis::regular<double, boost::use_default, metadata_t, boost::use_default>, boost::histogram::axis::regular<double, boost::use_default, metadata_t, boost::histogram::axis::option::bitset<1u> >, boost::histogram::axis::regular<double, boost::use_default, metadata_t, boost::histogram::axis::option::bitset<2u> >, boost::histogram::axis::regular<double, boost::use_default, metadata_t, boost::histogram::axis::option::bitset<0u> >, boost::histogram::axis::regular<double, boost::use_default, metadata_t, boost::histogram::axis::option::bitset<11u> >, boost::histogram::axis::regular<double, boost::use_default, metadata_t, boost::histogram::axis::option::bitset<6u> >, boost::histogram::axis::regular<double, boost::histogram::axis::transform::pow, metadata_t, boost::use_default>, boost::histogram::axis::regular<double, func_transform, metadata_t, boost::use_default>, axis::regular_numpy, boost::histogram::axis::variable<double, metadata_t, boost::use_default, std::__2::allocator<double> >, boost::histogram::axis::variable<double, metadata_t, boost::histogram::axis::option::bitset<1u>, std::__2::allocator<double> >, boost::histogram::axis::variable<double, metadata_t, boost::histogram::axis::option::bitset<2u>, std::__2::allocator<double> >, boost::histogram::axis::variable<double, metadata_t, boost::histogram::axis::option::bitset<0u>, std::__2::allocator<double> >, boost::histogram::axis::variable<double, metadata_t, boost::histogram::axis::option::bitset<11u>, std::__2::allocator<double> >, boost::histogram::axis::variable<double, metadata_t, boost::histogram::axis::option::bitset<6u>, std::__2::allocator<double> >, boost::histogram::axis::integer<int, metadata_t, boost::use_default>, boost::histogram::axis::integer<int, metadata_t, boost::histogram::axis::option::bitset<1u> >, boost::histogram::axis::integer<int, metadata_t, boost::histogram::axis::option::bitset<2u> >, boost::histogram::axis::integer<int, metadata_t, boost::histogram::axis::option::bitset<0u> >, boost::histogram::axis::integer<int, metadata_t, boost::histogram::axis::option::bitset<8u> >, boost::histogram::axis::integer<int, metadata_t, boost::histogram::axis::option::bitset<4u> >, boost::histogram::axis::category<int, metadata_t, boost::use_default, std::__2::allocator<int> >, boost::histogram::axis::category<int, metadata_t, boost::histogram::axis::option::bitset<8u>, std::__2::allocator<int> >, boost::histogram::axis::category<std::__2::basic_string<char, std::__2::char_traits<char>, std::__2::allocator<char> >, metadata_t, boost::histogram::axis::option::bitset<2u>, std::__2::allocator<std::__2::basic_string<char, std::__2::char_traits<char>, std::__2::allocator<char> > > >, boost::histogram::axis::category<std::__2::basic_string<char, std::__2::char_traits<char>, std::__2::allocator<char> >, metadata_t, boost::histogram::axis::option::bitset<8u>, std::__2::allocator<std::__2::basic_string<char, std::__2::char_traits<char>, std::__2::allocator<char> > > >, axis::boolean>, std::__2::allocator<boost::histogram::axis::variant<boost::histogram::axis::regular<double, boost::use_default, metadata_t, boost::use_default>, boost::histogram::axis::regular<double, boost::use_default, metadata_t, boost::histogram::axis::option::bitset<1u> >, boost::histogram::axis::regular<double, boost::use_default, metadata_t, boost::histogram::axis::option::bitset<2u> >, boost::histogram::axis::regular<double, boost::use_default, metadata_t, boost::histogram::axis::option::bitset<0u> >, boost::histogram::axis::regular<double, boost::use_default, metadata_t, boost::histogram::axis::option::bitset<11u> >, boost::histogram::axis::regular<double, boost::use_default, metadata_t, boost::histogram::axis::option::bitset<6u> >, boost::histogram::axis::regular<double, boost::histogram::axis::transform::pow, metadata_t, boost::use_default>, boost::histogram::axis::regular<double, func_transform, metadata_t, boost::use_default>, axis::regular_numpy, boost::histogram::axis::variable<double, metadata_t, boost::use_default, std::__2::allocator<double> >, boost::histogram::axis::variable<double, metadata_t, boost::histogram::axis::option::bitset<1u>, std::__2::allocator<double> >, boost::histogram::axis::variable<double, metadata_t, boost::histogram::axis::option::bitset<2u>, std::__2::allocator<double> >, boost::histogram::axis::variable<double, metadata_t, boost::histogram::axis::option::bitset<0u>, std::__2::allocator<double> >, boost::histogram::axis::variable<double, metadata_t, boost::histogram::axis::option::bitset<11u>, std::__2::allocator<double> >, boost::histogram::axis::variable<double, metadata_t, boost::histogram::axis::option::bitset<6u>, std::__2::allocator<double> >, boost::histogram::axis::integer<int, metadata_t, boost::use_default>, boost::histogram::axis::integer<int, metadata_t, boost::histogram::axis::option::bitset<1u> >, boost::histogram::axis::integer<int, metadata_t, boost::histogram::axis::option::bitset<2u> >, boost::histogram::axis::integer<int, metadata_t, boost::histogram::axis::option::bitset<0u> >, boost::histogram::axis::integer<int, metadata_t, boost::histogram::axis::option::bitset<8u> >, boost::histogram::axis::integer<int, metadata_t, boost::histogram::axis::option::bitset<4u> >, boost::histogram::axis::category<int, metadata_t, boost::use_default, std::__2::allocator<int> >, boost::histogram::axis::category<int, metadata_t, boost::histogram::axis::option::bitset<8u>, std::__2::allocator<int> >, boost::histogram::axis::category<std::__2::basic_string<char, std::__2::char_traits<char>, std::__2::allocator<char> >, metadata_t, boost::histogram::axis::option::bitset<2u>, std::__2::allocator<std::__2::basic_string<char, std::__2::char_traits<char>, std::__2::allocator<char> > > >, boost::histogram::axis::category<std::__2::basic_string<char, std::__2::char_traits<char>, std::__2::allocator<char> >, metadata_t, boost::histogram::axis::option::bitset<8u>, std::__2::allocator<std::__2::basic_string<char, std::__2::char_traits<char>, std::__2::allocator<char> > > >, axis::boolean> > >, boost::histogram::storage_adaptor<std::__2::vector<double, std::__2::allocator<double> > > > const&, pybind11::args&)#1}, double, boost::histogram::histogram<std::__2::vector<boost::histogram::axis::variant<boost::histogram::axis::regular<double, boost::use_default, metadata_t, boost::use_default>, boost::histogram::axis::regular<double, boost::use_default, metadata_t, boost::histogram::axis::option::bitset<1u> >, boost::histogram::axis::regular<double, boost::use_default, metadata_t, boost::histogram::axis::option::bitset<2u> >, boost::histogram::axis::regular<double, boost::use_default, metadata_t, boost::histogram::axis::option::bitset<0u> >, boost::histogram::axis::regular<double, boost::use_default, metadata_t, boost::histogram::axis::option::bitset<11u> >, boost::histogram::axis::regular<double, boost::use_default, metadata_t, boost::histogram::axis::option::bitset<6u> >, boost::histogram::axis::regular<double, boost::histogram::axis::transform::pow, metadata_t, boost::use_default>, boost::histogram::axis::regular<double, func_transform, metadata_t, boost::use_default>, axis::regular_numpy, boost::histogram::axis::variable<double, metadata_t, boost::use_default, std::__2::allocator<double> >, boost::histogram::axis::variable<double, metadata_t, boost::histogram::axis::option::bitset<1u>, std::__2::allocator<double> >, boost::histogram::axis::variable<double, metadata_t, boost::histogram::axis::option::bitset<2u>, std::__2::allocator<double> >, boost::histogram::axis::variable<double, metadata_t, boost::histogram::axis::option::bitset<0u>, std::__2::allocator<double> >, boost::histogram::axis::variable<double, metadata_t, boost::histogram::axis::option::bitset<11u>, std::__2::allocator<double> >, boost::histogram::axis::variable<double, metadata_t, boost::histogram::axis::option::bitset<6u>, std::__2::allocator<double> >, boost::histogram::axis::integer<int, metadata_t, boost::use_default>, boost::histogram::axis::integer<int, metadata_t, boost::histogram::axis::option::bitset<1u> >, boost::histogram::axis::integer<int, metadata_t, boost::histogram::axis::option::bitset<2u> >, boost::histogram::axis::integer<int, metadata_t, boost::histogram::axis::option::bitset<0u> >, boost::histogram::axis::integer<int, metadata_t, boost::histogram::axis::option::bitset<8u> >, boost::histogram::axis::integer<int, metadata_t, boost::histogram::axis::option::bitset<4u> >, boost::histogram::axis::category<int, metadata_t, boost::use_default, std::__2::allocator<int> >, boost::histogram::axis::category<int, metadata_t, boost::histogram::axis::option::bitset<8u>, std::__2::allocator<int> >, boost::histogram::axis::category<std::__2::basic_string<char, std::__2::char_traits<char>, std::__2::allocator<char> >, metadata_t, boost::histogram::axis::option::bitset<2u>, std::__2::allocator<std::__2::basic_string<char, std::__2::char_traits<char>, std::__2::allocator<char> > > >, boost::histogram::axis::category<std::__2::basic_string<char, std::__2::char_traits<char>, std::__2::allocator<char> >, metadata_t, boost::histogram::axis::option::bitset<8u>, std::__2::allocator<std::__2::basic_string<char, std::__2::char_traits<char>, std::__2::allocator<char> > > >, axis::boolean>, std::__2::allocator<boost::histogram::axis::variant<boost::histogram::axis::regular<double, boost::use_default, metadata_t, boost::use_default>, boost::histogram::axis::regular<double, boost::use_default, metadata_t, boost::histogram::axis::option::bitset<1u> >, boost::histogram::axis::regular<double, boost::use_default, metadata_t, boost::histogram::axis::option::bitset<2u> >, boost::histogram::axis::regular<double, boost::use_default, metadata_t, boost::histogram::axis::option::bitset<0u> >, boost::histogram::axis::regular<double, boost::use_default, metadata_t, boost::histogram::axis::option::bitset<11u> >, boost::histogram::axis::regular<double, boost::use_default, metadata_t, boost::histogram::axis::option::bitset<6u> >, boost::histogram::axis::regular<double, boost::histogram::axis::transform::pow, metadata_t, boost::use_default>, boost::histogram::axis::regular<double, func_transform, metadata_t, boost::use_default>, axis::regular_numpy, boost::histogram::axis::variable<double, metadata_t, boost::use_default, std::__2::allocator<double> >, boost::histogram::axis::variable<double, metadata_t, boost::histogram::axis::option::bitset<1u>, std::__2::allocator<double> >, boost::histogram::axis::variable<double, metadata_t, boost::histogram::axis::option::bitset<2u>, std::__2::allocator<double> >, boost::histogram::axis::variable<double, metadata_t, boost::histogram::axis::option::bitset<0u>, std::__2::allocator<double> >, boost::histogram::axis::variable<double, metadata_t, boost::histogram::axis::option::bitset<11u>, std::__2::allocator<double> >, boost::histogram::axis::variable<double, metadata_t, boost::histogram::axis::option::bitset<6u>, std::__2::allocator<double> >, boost::histogram::axis::integer<int, metadata_t, boost::use_default>, boost::histogram::axis::integer<int, metadata_t, boost::histogram::axis::option::bitset<1u> >, boost::histogram::axis::integer<int, metadata_t, boost::histogram::axis::option::bitset<2u> >, boost::histogram::axis::integer<int, metadata_t, boost::histogram::axis::option::bitset<0u> >, boost::histogram::axis::integer<int, metadata_t, boost::histogram::axis::option::bitset<8u> >, boost::histogram::axis::integer<int, metadata_t, boost::histogram::axis::option::bitset<4u> >, boost::histogram::axis::category<int, metadata_t, boost::use_default, std::__2::allocator<int> >, boost::histogram::axis::category<int, metadata_t, boost::histogram::axis::option::bitset<8u>, std::__2::allocator<int> >, boost::histogram::axis::category<std::__2::basic_string<char, std::__2::char_traits<char>, std::__2::allocator<char> >, metadata_t, boost::histogram::axis::option::bitset<2u>, std::__2::allocator<std::__2::basic_string<char, std::__2::char_traits<char>, std::__2::allocator<char> > > >, boost::histogram::axis::category<std::__2::basic_string<char, std::__2::char_traits<char>, std::__2::allocator<char> >, metadata_t, boost::histogram::axis::option::bitset<8u>, std::__2::allocator<std::__2::basic_string<char, std::__2::char_traits<char>, std::__2::allocator<char> > > >, axis::boolean> > >, boost::histogram::storage_adaptor<std::__2::vector<double, std::__2::allocator<double> > > > const&, pybind11::args&, pybind11::name, pybind11::is_method, pybind11::sibling>(register_histogram<boost::histogram::storage_adaptor<std::__2::vector<double, std::__2::allocator<double> > > >(pybind11::module_&, char const*, char const*)::{lambda(boost::histogram::histogram<std::__2::vector<boost::histogram::axis::variant<boost::histogram::axis::regular<double, boost::use_default, metadata_t, boost::use_default>, boost::histogram::axis::regular<double, boost::use_default, metadata_t, boost::histogram::axis::option::bitset<1u> >, boost::histogram::axis::regular<double, boost::use_default, metadata_t, boost::histogram::axis::option::bitset<2u> >, boost::histogram::axis::regular<double, boost::use_default, metadata_t, boost::histogram::axis::option::bitset<0u> >, boost::histogram::axis::regular<double, boost::use_default, metadata_t, boost::histogram::axis::option::bitset<11u> >, boost::histogram::axis::regular<double, boost::use_default, metadata_t, boost::histogram::axis::option::bitset<6u> >, boost::histogram::axis::regular<double, boost::histogram::axis::transform::pow, metadata_t, boost::use_default>, boost::histogram::axis::regular<double, func_transform, metadata_t, boost::use_default>, axis::regular_numpy, boost::histogram::axis::variable<double, metadata_t, boost::use_default, std::__2::allocator<double> >, boost::histogram::axis::variable<double, metadata_t, boost::histogram::axis::option::bitset<1u>, std::__2::allocator<double> >, boost::histogram::axis::variable<double, metadata_t, boost::histogram::axis::option::bitset<2u>, std::__2::allocator<double> >, boost::histogram::axis::variable<double, metadata_t, boost::histogram::axis::option::bitset<0u>, std::__2::allocator<double> >, boost::histogram::axis::variable<double, metadata_t, boost::histogram::axis::option::bitset<11u>, std::__2::allocator<double> >, boost::histogram::axis::variable<double, metadata_t, boost::histogram::axis::option::bitset<6u>, std::__2::allocator<double> >, boost::histogram::axis::integer<int, metadata_t, boost::use_default>, boost::histogram::axis::integer<int, metadata_t, boost::histogram::axis::option::bitset<1u> >, boost::histogram::axis::integer<int, metadata_t, boost::histogram::axis::option::bitset<2u> >, boost::histogram::axis::integer<int, metadata_t, boost::histogram::axis::option::bitset<0u> >, boost::histogram::axis::integer<int, metadata_t, boost::histogram::axis::option::bitset<8u> >, boost::histogram::axis::integer<int, metadata_t, boost::histogram::axis::option::bitset<4u> >, boost::histogram::axis::category<int, metadata_t, boost::use_default, std::__2::allocator<int> >, boost::histogram::axis::category<int, metadata_t, boost::histogram::axis::option::bitset<8u>, std::__2::allocator<int> >, boost::histogram::axis::category<std::__2::basic_string<char, std::__2::char_traits<char>, std::__2::allocator<char> >, metadata_t, boost::histogram::axis::option::bitset<2u>, std::__2::allocator<std::__2::basic_string<char, std::__2::char_traits<char>, std::__2::allocator<char> > > >, boost::histogram::axis::category<std::__2::basic_string<char, std::__2::char_traits<char>, std::__2::allocator<char> >, metadata_t, boost::histogram::axis::option::bitset<8u>, std::__2::allocator<std::__2::basic_string<char, std::__2::char_traits<char>, std::__2::allocator<char> > > >, axis::boolean>, std::__2::allocator<boost::histogram::axis::variant<boost::histogram::axis::regular<double, boost::use_default, metadata_t, boost::use_default>, boost::histogram::axis::regular<double, boost::use_default, metadata_t, boost::histogram::axis::option::bitset<1u> >, boost::histogram::axis::regular<double, boost::use_default, metadata_t, boost::histogram::axis::option::bitset<2u> >, boost::histogram::axis::regular<double, boost::use_default, metadata_t, boost::histogram::axis::option::bitset<0u> >, boost::histogram::axis::regular<double, boost::use_default, metadata_t, boost::histogram::axis::option::bitset<11u> >, boost::histogram::axis::regular<double, boost::use_default, metadata_t, boost::histogram::axis::option::bitset<6u> >, boost::histogram::axis::regular<double, boost::histogram::axis::transform::pow, metadata_t, boost::use_default>, boost::histogram::axis::regular<double, func_transform, metadata_t, boost::use_default>, axis::regular_numpy, boost::histogram::axis::variable<double, metadata_t, boost::use_default, std::__2::allocator<double> >, boost::histogram::axis::variable<double, metadata_t, boost::histogram::axis::option::bitset<1u>, std::__2::allocator<double> >, boost::histogram::axis::variable<double, metadata_t, boost::histogram::axis::option::bitset<2u>, std::__2::allocator<double> >, boost::histogram::axis::variable<double, metadata_t, boost::histogram::axis::option::bitset<0u>, std::__2::allocator<double> >, boost::histogram::axis::variable<double, metadata_t, boost::histogram::axis::option::bitset<11u>, std::__2::allocator<double> >, boost::histogram::axis::variable<double, metadata_t, boost::histogram::axis::option::bitset<6u>, std::__2::allocator<double> >, boost::histogram::axis::integer<int, metadata_t, boost::use_default>, boost::histogram::axis::integer<int, metadata_t, boost::histogram::axis::option::bitset<1u> >, boost::histogram::axis::integer<int, metadata_t, boost::histogram::axis::option::bitset<2u> >, boost::histogram::axis::integer<int, metadata_t, boost::histogram::axis::option::bitset<0u> >, boost::histogram::axis::integer<int, metadata_t, boost::histogram::axis::option::bitset<8u> >, boost::histogram::axis::integer<int, metadata_t, boost::histogram::axis::option::bitset<4u> >, boost::histogram::axis::category<int, metadata_t, boost::use_default, std::__2::allocator<int> >, boost::histogram::axis::category<int, metadata_t, boost::histogram::axis::option::bitset<8u>, std::__2::allocator<int> >, boost::histogram::axis::category<std::__2::basic_string<char, std::__2::char_traits<char>, std::__2::allocator<char> >, metadata_t, boost::histogram::axis::option::bitset<2u>, std::__2::allocator<std::__2::basic_string<char, std::__2::char_traits<char>, std::__2::allocator<char> > > >, boost::histogram::axis::category<std::__2::basic_string<char, std::__2::char_traits<char>, std::__2::allocator<char> >, metadata_t, boost::histogram::axis::option::bitset<8u>, std::__2::allocator<std::__2::basic_string<char, std::__2::char_traits<char>, std::__2::allocator<char> > > >, axis::boolean> > >, boost::histogram::storage_adaptor<std::__2::vector<double, std::__2::allocator<double> > > > const&, pybind11::args&)#1}&&, double (*)(boost::histogram::histogram<std::__2::vector<boost::histogram::axis::variant<boost::histogram::axis::regular<double, boost::use_default, metadata_t, boost::use_default>, boost::histogram::axis::regular<double, boost::use_default, metadata_t, boost::histogram::axis::option::bitset<1u> >, boost::histogram::axis::regular<double, boost::use_default, metadata_t, boost::histogram::axis::option::bitset<2u> >, boost::histogram::axis::regular<double, boost::use_default, metadata_t, boost::histogram::axis::option::bitset<0u> >, boost::histogram::axis::regular<double, boost::use_default, metadata_t, boost::histogram::axis::option::bitset<11u> >, boost::histogram::axis::regular<double, boost::use_default, metadata_t, boost::histogram::axis::option::bitset<6u> >, boost::histogram::axis::regular<double, boost::histogram::axis::transform::pow, metadata_t, boost::use_default>, boost::histogram::axis::regular<double, func_transform, metadata_t, boost::use_default>, axis::regular_numpy, boost::histogram::axis::variable<double, metadata_t, boost::use_default, std::__2::allocator<double> >, boost::histogram::axis::variable<double, metadata_t, boost::histogram::axis::option::bitset<1u>, std::__2::allocator<double> >, boost::histogram::axis::variable<double, metadata_t, boost::histogram::axis::option::bitset<2u>, std::__2::allocator<double> >, boost::histogram::axis::variable<double, metadata_t, boost::histogram::axis::option::bitset<0u>, std::__2::allocator<double> >, boost::histogram::axis::variable<double, metadata_t, boost::histogram::axis::option::bitset<11u>, std::__2::allocator<double> >, boost::histogram::axis::variable<double, metadata_t, boost::histogram::axis::option::bitset<6u>, std::__2::allocator<double> >, boost::histogram::axis::integer<int, metadata_t, boost::use_default>, boost::histogram::axis::integer<int, metadata_t, boost::histogram::axis::option::bitset<1u> >, boost::histogram::axis::integer<int, metadata_t, boost::histogram::axis::option::bitset<2u> >, boost::histogram::axis::integer<int, metadata_t, boost::histogram::axis::option::bitset<0u> >, boost::histogram::axis::integer<int, metadata_t, boost::histogram::axis::option::bitset<8u> >, boost::histogram::axis::integer<int, metadata_t, boost::histogram::axis::option::bitset<4u> >, boost::histogram::axis::category<int, metadata_t, boost::use_default, std::__2::allocator<int> >, boost::histogram::axis::category<int, metadata_t, boost::histogram::axis::option::bitset<8u>, std::__2::allocator<int> >, boost::histogram::axis::category<std::__2::basic_string<char, std::__2::char_traits<char>, std::__2::allocator<char> >, metadata_t, boost::histogram::axis::option::bitset<2u>, std::__2::allocator<std::__2::basic_string<char, std::__2::char_traits<char>, std::__2::allocator<char> > > >, boost::histogram::axis::category<std::__2::basic_string<char, std::__2::char_traits<char>, std::__2::allocator<char> >, metadata_t, boost::histogram::axis::option::bitset<8u>, std::__2::allocator<std::__2::basic_string<char, std::__2::char_traits<char>, std::__2::allocator<char> > > >, axis::boolean>, std::__2::allocator<boost::histogram::axis::variant<boost::histogram::axis::regular<double, boost::use_default, metadata_t, boost::use_default>, boost::histogram::axis::regular<double, boost::use_default, metadata_t, boost::histogram::axis::option::bitset<1u> >, boost::histogram::axis::regular<double, boost::use_default, metadata_t, boost::histogram::axis::option::bitset<2u> >, boost::histogram::axis::regular<double, boost::use_default, metadata_t, boost::histogram::axis::option::bitset<0u> >, boost::histogram::axis::regular<double, boost::use_default, metadata_t, boost::histogram::axis::option::bitset<11u> >, boost::histogram::axis::regular<double, boost::use_default, metadata_t, boost::histogram::axis::option::bitset<6u> >, boost::histogram::axis::regular<double, boost::histogram::axis::transform::pow, metadata_t, boost::use_default>, boost::histogram::axis::regular<double, func_transform, metadata_t, boost::use_default>, axis::regular_numpy, boost::histogram::axis::variable<double, metadata_t, boost::use_default, std::__2::allocator<double> >, boost::histogram::axis::variable<double, metadata_t, boost::histogram::axis::option::bitset<1u>, std::__2::allocator<double> >, boost::histogram::axis::variable<double, metadata_t, boost::histogram::axis::option::bitset<2u>, std::__2::allocator<double> >, boost::histogram::axis::variable<double, metadata_t, boost::histogram::axis::option::bitset<0u>, std::__2::allocator<double> >, boost::histogram::axis::variable<double, metadata_t, boost::histogram::axis::option::bitset<11u>, std::__2::allocator<double> >, boost::histogram::axis::variable<double, metadata_t, boost::histogram::axis::option::bitset<6u>, std::__2::allocator<double> >, boost::histogram::axis::integer<int, metadata_t, boost::use_default>, boost::histogram::axis::integer<int, metadata_t, boost::histogram::axis::option::bitset<1u> >, boost::histogram::axis::integer<int, metadata_t, boost::histogram::axis::option::bitset<2u> >, boost::histogram::axis::integer<int, metadata_t, boost::histogram::axis::option::bitset<0u> >, boost::histogram::axis::integer<int, metadata_t, boost::histogram::axis::option::bitset<8u> >, boost::histogram::axis::integer<int, metadata_t, boost::histogram::axis::option::bitset<4u> >, boost::histogram::axis::category<int, metadata_t, boost::use_default, std::__2::allocator<int> >, boost::histogram::axis::category<int, metadata_t, boost::histogram::axis::option::bitset<8u>, std::__2::allocator<int> >, boost::histogram::axis::category<std::__2::basic_string<char, std::__2::char_traits<char>, std::__2::allocator<char> >, metadata_t, boost::histogram::axis::option::bitset<2u>, std::__2::allocator<std::__2::basic_string<char, std::__2::char_traits<char>, std::__2::allocator<char> > > >, boost::histogram::axis::category<std::__2::basic_string<char, std::__2::char_traits<char>, std::__2::allocator<char> >, metadata_t, boost::histogram::axis::option::bitset<8u>, std::__2::allocator<std::__2::basic_string<char, std::__2::char_traits<char>, std::__2::allocator<char> > > >, axis::boolean> > >, boost::histogram::storage_adaptor<std::__2::vector<double, std::__2::allocator<double> > > > const&, pybind11::args&), pybind11::name const&, pybind11::is_method const&, pybind11::sibling const&)::{lambda(pybind11::detail::function_call&)#1}::operator()(pybind11::detail::function_call) 
```

hoodmane: The exception is thrown as the last wasm instruction in:
https://github.com/pybind/pybind11/blob/a25d40c7bceb4db227aeb4a44fc74e28eb1015f1/include/pybind11/pybind11.h#L1276

hoodmane: The error message is:
```
Unable to cast Python instance to C++ type (compile in debug mode for details)
```

henryiii: I wonder if something is causing `sum` to leak through into a C++ call. `[sum]` should convert to `slice(None, None, sum)` in https://github.com/scikit-hep/boost-histogram/blob/724b0b2e59f04c625d6248103596f3fdda32f2d1/src/boost_histogram/_internal/hist.py#L623 then be converted into the correct Boost.Histogram command in https://github.com/scikit-hep/boost-histogram/blob/724b0b2e59f04c625d6248103596f3fdda32f2d1/src/boost_histogram/_internal/hist.py#L811.

If I can activate and see logging, there should be logging commands before the C++ calls here. I think I can get this locally too (slightly different message, but close).

hoodmane: Are these C++ exceptions supposed to be caught and raised as a `RuntimeError` in normal Python? Are they expected to occur ever or are they supposed to reflect programmer errors?

We probably can't get them to be non panicking without enabling exceptions which I think causes a performance hit (maybe an acceptable one?) but I think I can improve the resulting error message a lot.

henryiii: They are supposed to be seen from Python. If you pass Python objects into a pybind11 function that you don't have a conversion for, you get a RuntimeError.

In fact, turning off EH would break this: https://github.com/scikit-hep/boost-histogram/blob/724b0b2e59f04c625d6248103596f3fdda32f2d1/src/boost_histogram/_internal/hist.py#L765-L771

henryiii: pybind11 uses exceptions pretty heavily: https://pybind11.readthedocs.io/en/stable/advanced/exceptions.html

henryiii: That said, this one is pretty easy to work around in Python if needed, I can just verify this really is a tuple of integers rather than just trying it and catching the C++ exception.

hoodmane: Can you add a test that triggers a different one of these C++ exceptions?

hoodmane: Oh sorry I didn't read your message carefully:

> For boost-histogram, I think this is the only place this occurs

henryiii: That's the only place we catch one.

henryiii: The rest of them are "real" errors, that is, they should show up to the user, and are not handled.

hoodmane: Apparently wasm exception handling is pretty far along, it's been released in Chrome 96 in September and by Safari 15.2 in December. Firefox only has it on firefox nightly. According to the v8 people, enabling wasm exception handling has a negligible performance cost.

We could also just try enabling js exception handling and see how big the cost is.

hoodmane: Huh, well it seems like we do have js exception handling enabled because otherwise it would just say:

> 'DISABLE_EXCEPTION_THROWING was set (likely due to -fno-exceptions), which means no C++ exception throwing support code is linked in, but such support is required by symbol ' + name + '. Either do not set DISABLE_EXCEPTION_THROWING (if you do want exception throwing) or compile all source files with -fno-except (so that no exceptions support code is required); also make sure DISABLE_EXCEPTION_CATCHING is set to the right value - if you want exceptions, it should be off, and vice versa.'

henryiii:
Ouch, during the rollout, we just discovered a bad regression in 1.3.0. Looks like there will be a 1.3.1 very soon; if this patch works, I can include it, so we won't need a patch here.

If you are curious as to what happens in pyodide when you get a segfault, I can tell you how to segfault 1.3.0. 🤦

hoodmane:
> If you are curious as to what happens in pyodide when you get a segfault, I can tell you how to segfault 1.3.0. 🤦

I am interested if Pyodide doesn't give a fatal error message. wasm32_emscripten has no memory protection so if you try to dereference a bad pointer it won't instantly crash unless the pointer is larger than the total size of the memory. Likewise, dereferencing null works just fine. So we have to manually detect segfaults and report them otherwise the runtime will try to keep going after corrupting its state. 

There is some luck required, it is possible the bad memory access will never propagate into anything that causes a trap and will just lead to weird stuff. But in most cases an actual wasm trap occurs a bit later and we can notice it and kill the runtime. Any case like #2178 where we fail to do this is very interesting.

henryiii:
Correction, I forgot to check the console, the segfaulting behavior acts like throwing a `py::key_error`, which also "hangs", but there's this in the console (either one):
Screen Shot 2022-02-15 at 3 02 07 PM

hoodmane:
Hmm okay that looks interesting, I'll have to reproduce it locally and see if I can improve our handling of these errors. Thanks so much for all your help!

henryiii:
If it could turn into proper KeyError like in CPython (even for just the latest browsers), that would be ideal/amazing. :)

hoodmane:
Yeah, that'd be great. There's a similar issue with panic unwinds on #2081 which would also be nice to fix, though they may or may not be related. I will try to investigate this soon.

hoodmane:
Okay I think #2178 fixes all of the examples you've given of bad fatal errors. On that branch your segfault shows up as:
```
CppException: Unable to cast Python instance to C++ type (compile in debug mode for details)
```

hoodmane:
I figured out how to get the exception handling working. @henryiii should I push it here? 

It increases the size of the main Pyodide module by about 150kb or about 1%, which is probably acceptable. I am not sure about impact on benchmarks, will have to see if it causes a noticeable performance regression.

henryiii:
That sounds fantastic for pybind11 libraries (and probably others). Can be here or another PR, I can combine both and try it out.

hoodmane:
I think we can avoid any overhead in the main module other than a small size increase.

henryiii:
I've triggered the release of 1.3.1, will update here when the release happens (2-3 hours). The `mean["invalid"]` example will still trigger a Python exception from C++, but we are avoiding using exceptions for control flow (the patch here) now, as well as the GIL issue has been fixed.

hoodmane:
It will work correctly on all browsers -- I'm using the Javascript exceptions not the wasm exceptions because there isn't enough browser support for wasm exceptions still. I have not investigated how it relates to Rust.

henryiii:
Okay, great. 1.3.1 is just out, so after this builds, I'll bump to 1.3.1 and add a test expecting `mean["invalid"]` to raise an exception instead.

emsdk/patches/fix-side-module-exception-handling.patch
```diff
@@ -0,0 +1,32 @@
+From 6d668a693154374809707b598bc658b3162c79c4 Mon Sep 17 00:00:00 2001
+From: Hood Chatham <roberthoodchatham@gmail.com>
+Date: Tue, 15 Feb 2022 23:27:03 -0500
+Subject: [PATCH] Fix side module exception handling
+
```
hoodmane, line 5: See https://github.com/emscripten-core/emscripten/issues/16302

hoodmane:
I compared the benchmarks to the ones on the main branch and there is no apparent difference, so I think they support my claim that this change doesn't impose any cost on the main module.

henryiii:
Is there something I need to rebuild? I'm trying this locally and I get:

Screen Shot 2022-02-16 at 2 12 58 PM

From what I understand (and from passing tests above), this should actually produce the Python error now. I'm guessing I need to rebuild something other than `PYODIDE_PACKAGES="boost-histogram" make`? Do I need to remove `--pre-build` from `./run_docker`? Would hope to keep at least some caching.

hoodmane:
Do `touch src/core/main.c` and then `PYODIDE_PACKAGES="boost-histogram" make`.

henryiii:
Didn't work, so used make clean on the main directory, plus in emsdk just to be sure.
Screen Shot 2022-02-16 at 3 29 14 PM

Binary size grew a little, the wheel is now 939K. Very nice!!!

Is there a way I can access my test suite and/or a python file from the terminal emulator? Curious to try benchmarking.

henryiii:
(I also just realized I don't need to be in the docker container to run `python3 -m http.server` 🤦)

henryiii:
From a simple copy-paste performance test, boost-histogram in the browser is 3.0x slower than native (python 3.10). For a 1D regularly binned histogram, boost-histogram in a browser still beats native NumPy by a few percent. 🤣 Native NumPy 1D histograming is 3.2x slower in the browser. 2D histograms are 1.8x slower in boost-histogram and 1.3x slower in NumPy (but there boost-histogram has a 10x starting lead). The numbers from time are oddly rounded, maybe there's a clock accuracy limitation? Looks like it stops at milliseconds, perhaps.

hoodmane:
> The numbers from time are oddly rounded, maybe there's a clock accuracy limitation? Looks like it stops at milliseconds, perhaps.

See https://developer.mozilla.org/en-US/docs/Web/API/Performance/now:

> It's important to keep in mind that to mitigate potential security threats such as [Spectre](https://spectreattack.com/), browsers typically round the returned value by some amount in order to be less predictable. This inherently introduces a degree of inaccuracy by limiting the resolution or precision of the timer. For example, Firefox rounds the returned time to 1 millisecond increments.

hoodmane:
No, that's a sporadic failure, I restarted CI.

packages/boost-histogram/test_boost_histogram.py
```diff
@@ -0,0 +1,41 @@
+import pytest
```
henryiii, line 1: Does this have to be imported inside the function, then?

hoodmane: Yeah. We usually use `unittest.assertThrows` since we don't want to install `pytest`.

hoodmane: Or more accurately, `unittest.TestCase().assertRaises`.

hoodmane (approved):
Looks good to me, but ideally either @rth or @ryanking13 should review the changes I made to enable exceptions. 

Note that my emscripten patch has been accepted: https://github.com/emscripten-core/emscripten/pull/16309
In the corresponding issue the emcripten folks confirmed that this approach should not degrade performance in the main models.

hoodmane:
Okay, I think what we should do is move the exception-related changes to the Makefiles over to #2178 and merge that first, then merge this.

rth:
Thanks, merging.

rth:
Can we backport this to 0.19.1 in https://github.com/pyodide/pyodide/pull/2192 without including https://github.com/pyodide/pyodide/pull/2178 ?

hoodmane:
Yes, ideally such a backport would include also the patch in https://github.com/pyodide/pyodide/pull/2174/commits/f3dcdf6f88bb476a1557ed1f9af147d7bb1d04b4. But I think it may be fine to backport #2178 too.

hoodmane:
Actually I think the best option is to add `EXCEPTION_CATCHING_ALLOWED=['we only want to allow exception handling in side modules']` to ldflags and flags and add the emsdk patch in addition to this commit. That allows this package to do it's exception handling without fatally falling. The rest of most of #2178 is more to make it easier to develop C++ packages.

PR #2177, opened 2022-02-14
style: improve pre-commit

henryiii:
### Description

Working on using pre-commit properly. Will be a series of commits, and undrafted after they are complete.

Rough plan:

- [x] Commit 1: update pre-commit, prepare mypy to run correctly.
- [x] Next commit: apply pre-commit fixes
- [x] Next commit: use pre-commit instead of manual runs inside Makefile.
- [x] Somewhere in between commit: fix manual issues

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -9,7 +9,6 @@ repos:
         exclude: .clang-format
       - id: end-of-file-fixer
       - id: trailing-whitespace
-        exclude: \.py$
 
   - repo: https://github.com/psf/black
     rev: "22.1.0"
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -3,36 +3,58 @@ default_language_version:
   python: "3.9"
 repos:
   - repo: https://github.com/pre-commit/pre-commit-hooks
-    rev: v2.3.0
+    rev: "v4.1.0"
     hooks:
       - id: check-yaml
+        exclude: .clang-format
       - id: end-of-file-fixer
       - id: trailing-whitespace
+        exclude: \.py$
+
   - repo: https://github.com/psf/black
-    rev: 20.8b1
+    rev: "22.1.0"
     hooks:
       - id: black
+
   - repo: https://gitlab.com/pycqa/flake8
-    rev: 3.7.8
+    rev: 3.9.2
     hooks:
       - id: flake8
         types: [file, python]
         # only check for unused imports, as the rest is done by black
         args: [--select=F401]
+
   - repo: https://github.com/pre-commit/mirrors-mypy
-    rev: v0.812
+    rev: v0.931
     hooks:
       - id: mypy
         args:
+          - --show-error-codes
           - --ignore-missing-imports
-  - repo: local
+        files: ^(src/|packages/.*/test|conftest.py|docs/)
+        additional_dependencies: &mypy-deps
+          - packaging
+          - types-docutils
+          - types-pyyaml
+          - types-setuptools
+      - id: mypy
+        name: mypy-micropip
+        files: ^packages/micropip/src/
+        args: &mypy-args
+          - --show-error-codes
+        additional_dependencies: *mypy-deps
+      - id: mypy
+        name: mypy-pyodide-build
+        files: ^pyodide-build/pyodide_build/
+        args: *mypy-args
+        additional_dependencies: *mypy-deps
+
+  - repo: https://github.com/pre-commit/mirrors-clang-format
+    rev: "v13.0.0"
     hooks:
       - id: clang-format
-        name: clang-format-6.0
-        language: system
-        entry: ./tools/clang-format-precommit.sh
 
   - repo: https://github.com/pre-commit/mirrors-prettier
-    rev: "v2.3.2" # Use the sha / tag you want to point at
+    rev: "v2.5.1"
     hooks:
       - id: prettier
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,0 +1,7 @@
+[tool.mypy]
+warn_unused_configs = true
+python_version = "3.9"
+
+[[tool.mypy.overrides]]
+module = ["pyodide", "pyodide_js", "pyodide.http", "pytest", "ruamel.yaml", "docutils.parsers.rst"]
+ignore_missing_imports = true
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,7 +1,15 @@
 [tool.mypy]
 warn_unused_configs = true
 python_version = "3.9"
+mypy_path = "src/py"
 
 [[tool.mypy.overrides]]
-module = ["pyodide", "pyodide_js", "pyodide.http", "pytest", "ruamel.yaml", "docutils.parsers.rst"]
+module = [
+  "_pyodide_core",
+  "docutils.parsers.rst",
+  "js",
+  "pyodide_js",
+  "pytest",
+  "ruamel.yaml",
+]
 ignore_missing_imports = true
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -23,6 +23,11 @@ jobs:
       - checkout
 
       - run:
+          name: Install lint requirements
+          command: |
+            python -m pip install pre-commit
+
+      - run:
           name: lint
           command: make lint
 
--- a/.prettierignore
+++ b/.prettierignore
@@ -8,5 +8,5 @@ emsdk
 cpython
 .vscode
 .pytest_cache
-.clang-format 
+.clang-format
 packages/CLAPACK/make.inc
--- a/Makefile
+++ b/Makefile
@@ -137,7 +137,7 @@ update_base_url: \
 
 .PHONY: lint
 lint:
-	pre-commit run -a
+	pre-commit run -a --show-diff-on-failure
 
 benchmark: all
 	$(HOSTPYTHON) benchmark/benchmark.py $(HOSTPYTHON) build/benchmarks.json
--- a/Makefile
+++ b/Makefile
@@ -135,33 +135,9 @@ update_base_url: \
 
 
 
-lint: node_modules/.installed
-	# check for unused imports, the rest is done by black
-	flake8 --select=F401 src tools pyodide-build benchmark conftest.py docs packages/matplotlib/src/
-	mypy --ignore-missing-imports    \
-		pyodide-build/pyodide_build/ \
-		src/ 					     \
-		packages/*/test* 			 \
-		conftest.py 				 \
-		docs
-#mypy gets upset about there being both : src / py / setup.py and
-#packages / micropip / src / setup.py.There is no easy way to fix this right now
-#see python / mypy #10428. This will also cause trouble with pre - commit if you
-#modify both setup.py files in the same commit.
-	mypy --ignore-missing-imports    \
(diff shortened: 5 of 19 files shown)


.pre-commit-config.yaml
```diff
@@ -3,36 +3,57 @@ default_language_version:
...
+          - types-pyyaml
+          - types-setuptools
+      - id: mypy
+        name: mypy-micropip
+        files: ^packages/micropip/src
+        args: &mypy-args
+          - --show-error-codes
+        additional_dependencies: *mypy-deps
+      - id: mypy
+        name: mypy-pyodide-build
```
rth, line 46: Do you think it's better to separate these by package, as opposed to running them together?

.pre-commit-config.yaml
```diff
@@ -3,36 +3,57 @@ default_language_version:
...
         args: [--select=F401]
+
   - repo: https://github.com/pre-commit/mirrors-mypy
-    rev: v0.812
+    rev: v0.931
     hooks:
       - id: mypy
         args:
+          - --show-error-codes
           - --ignore-missing-imports
```
rth, line 32: These are already defined in mypy settings now, I think

pyproject.toml
```diff
@@ -0,0 +1,7 @@
+[tool.mypy]
+warn_unused_configs = true
+python_version = "3.9"
+
+[[tool.mypy.overrides]]
+module = ["pyodide", "pyodide_js", "pyodide.http", "pytest", "ruamel.yaml"]
+ignore_missing_imports = true
```
rth, line 7: Do you think it's better bo restrict this to those modules as opposed to everything?

pyproject.toml
```diff
@@ -0,0 +1,7 @@
+[tool.mypy]
```
rth, line 1: If we add a `pyproject.toml` in the long term we should probably move configuration from https://github.com/pyodide/pyodide/blob/main/setup.cfg here as well.

henryiii:
Weird, the lint run really should have found an error.

```
docs/sphinx_pyodide/sphinx_pyodide/jsdoc.py:2: error: Module "docutils.parsers.rst" has no attribute "directives"; maybe "Directive"?  [attr-defined]
Found 1 error in 1 file (checked 82 source files)
```

.pre-commit-config.yaml
henryiii: That's what I did. And that's how you solve the dual `setup.py` bug. Since the environments exactly match, pre-commit caches them together.

.pre-commit-config.yaml
henryiii: No. For the general check, missing imports is ignored, otherwise you'd have a massive list of missing imports. For the packages (Pyodide-build and whatever the other one was), missing imports come from the configuration defined list.

pyproject.toml
henryiii: Yes, for normal code, it helps find bugs - ignoring an import hides bugs. If you can get the import, that's much better.

If you are referring to `pyodide", "pyodide.http"` vs. `"pyodide.*"`, I just went with minimal over blanket, have no idea what's better. Or if we really should be ignoring these, ideally they are typed and can be found. Does pyodide have a `py.typed` file?

henryiii:
> Up to you but we are generally squashing commits when merging, so don't spend too much effort on having a clean git history.

Good. :) Though clean commit history makes it easier to review. Also, rebase-and-merge on style changes allows you to then add style-only commits to a file to be git rev ignored, if you want. There is a somewhat standard filename for that, even ,but not picked up by GitHub, sadly, though.

pyproject.toml
rth: Your proposed setup sounds good to me, I was just wondering aloud as have never used that setting before.

> Does pyodide have a py.typed file?

No, but we probably should.

henryiii: Possibly dumb question: where is `pyodide`? I see `/src`, but not `/src/pyodide`. Probably why mypy can't find it either.

If mypy does find a package, it uses it - ignore_missing_imports just means it won't tell you if it doesn't find it, and makes it Any instead.

henryiii:
Looks like https://sourceforge.net/p/docutils/code/HEAD/tree/trunk/docutils/docutils/parsers/rst/directives/ is missing form https://github.com/python/typeshed/tree/master/stubs/docutils/docutils/parsers/rst. Long term, this should be fixed; for now, we can add it to the ignores. Known issue here: https://github.com/python/typeshed/issues/5755.

henryiii:
Checking to see what this looks like while running the old lint job on the newly formatted code, so ordering is a little different than above.

.pre-commit-config.yaml
```diff
@@ -3,36 +3,57 @@ default_language_version:
   python: "3.9"
 repos:
   - repo: https://github.com/pre-commit/pre-commit-hooks
-    rev: v2.3.0
+    rev: "v4.1.0"
     hooks:
       - id: check-yaml
+        exclude: .clang-format
       - id: end-of-file-fixer
       - id: trailing-whitespace
```
henryiii, line 11: This is breaking a test that has trailing whitespace in the output. Given that black handles .py files anyway, I can just turn it off for Python. Though not producing trailing whitespace in the first place might be better. ;)

hoodmane: Which test? Is it in `test_console.py`?

pyproject.toml
hoodmane: It's at src/py/pyodide. We have the python code under src/py, the typescript code which isn't run through the C preprocessor in src/js and a mix of C as well as typescript and javascript which get run through the C preprocessor in src/core.

.pre-commit-config.yaml
henryiii: https://github.com/pyodide/pyodide/blob/525d1111b094fc0c551dee9ad5e4f4d4546bdb61/packages/micropip/src/micropip/package.py#L10-L20

There's trailing whitespace there and it's needed for the test.

henryiii:
What should be done with:

```
packages/joblib/test_joblib.py:1:1: F401 'pytest' imported but unused
packages/scipy/test_scipy.py:1:1: F401 'pytest' imported but unused
packages/scipy/test_scipy.py:2:1: F401 'conftest.selenium_context_manager' imported but unused
packages/scipy/test_scipy.py:16:5: F401 'scipy as sp' imported but unused
packages/micropip/src/micropip/package.py:2:1: F401 'dataclasses.field' imported but unused
packages/micropip/src/micropip/package.py:3:1: F401 'pathlib.Path' imported but unused
packages/micropip/src/micropip/package.py:4:1: F401 'typing.Dict' imported but unused
packages/msgpack/test_pack.py:6:5: F401 'msgpack.Unpacker' imported but unused
packages/msgpack/test_pack.py:6:5: F401 'msgpack.Packer' imported but unused
packages/msgpack/test_pack.py:6:5: F401 'msgpack.pack' imported but unused
packages/test_packages_common.py:3:1: F401 'pathlib.Path' imported but unused
packages/test_packages_common.py:6:1: F401 'json' imported but unused
```

pyproject.toml
henryiii: Perfect, fixed this locally. Will update in the next push (after I know what to do with the missing imports).

henryiii:
As a followup, I'd recommend working on the other checks suggested here: https://scikit-hep.org/developer/style - there are a lot of useful things to both standardize on style, and find potential bugs. Flake8 has a lot of good checks, especially flake8-bugbear; I've already seen some tests that actually aren't testing anything due to missing `assert`'s, and bugbear has a check for pointless comparison that protects against that. You can very easily get a black-compatible flake8 run with `extend-ignore = E203, E501`, though my ideal config would be to add bugbear and then use:

```ini
[flake8]
extend-ignore = E203, E501, E722, B950
select = C,E,F,W,B,B9
```

(and a few extra ignores/complexity settings as needed).

Similarly, mypy should be looking inside typed defs, pyupgrade is really nice, etc. I can try a few followups if they don't require a lot of work. ;) This PR is not trying to add anything new, just move the linter driver to pre-commit.

henryiii:
Something weird is happening on CircleCI. It takes me 12 seconds to run pre-commit, with flake8 & mypy taking the most time. It seems to be hanging on CircleCI. I'm not that familiar with CircleCI, any ideas? Would you be interested in trying GHA for this, or pre-commit.ci?

ryanking13:
Thanks for working on this @henryiii!

> Something weird is happening on CircleCI. It takes me 12 seconds to run pre-commit, with flake8 & mypy taking the most time. It seems to be hanging on CircleCI. I'm not that familiar with CircleCI, any ideas?

Yes prettier rarely hangs in CircleCI, now sure why it does. Anyway, I re-triggered it.

ryanking13:
> Would you be interested in trying GHA for this, or pre-commit.ci?

Actually, I've read the pre-commit.ci docs thorougly for the first time today. It's more configurable then I thought and it's autoupdate and autofix (which can be triggered manually by comments) looks quite helpful. I think we could try using it, which was one discussed in (https://github.com/pyodide/pyodide/pull/2139#issuecomment-1024878787) @rth, @hoodmane?

rth:
We could try pre-commit.ci but I'm a bit wary of giving write permissions to third-party services as if those get compromised (which e.g. happened with Codecov last year), package deployment could also get compromised in our case. Unless there is a way to give write permission excluding protected branches?

>  Would you be interested in trying GHA

Yes, sure, feel free to move the lining job to GHA.

.pre-commit-config.yaml
rth: I find `trailing-whitespace` quite useful, particularly for docstrings (are you sure black would handle docstings)?  
Otherwise, for the doctest we could just use `...` to match anything. It's already  enabled.

rth (approved):
Thanks!

.pre-commit-config.yaml
ryanking13: If that specific doctest is the problem, maybe we could simply remove it, since `_format_table` is  a private function used only for `micropip.list()` and we already have a [doctest for micopip.list()](https://pyodide.org/en/stable/usage/api/micropip-api.html#micropip.list).

ryanking13:
> We could try pre-commit.ci but I'm a bit wary of giving write permissions to third-party services as if those get compromised (which e.g. happened with Codecov last year), package deployment could also get compromised in our case. Unless there is a way to give write permission excluding protected branches?

Yeah, I didn't thought of that perspective. I'll see whether we could restrict permissions.

ryanking13 (approved):

.pre-commit-config.yaml
henryiii: Pretty sure black does strip it, given there's a fmt off/on around this block already, presumably for black not to strip this (otherwise it doesn't touch docstrings).

Can't `_format_table` simply strip trailing whitespace? It's really a bit messy to put it into the output?

ryanking13: > Can't `_format_table` simply strip trailing whitespace? It's really a bit messy to put it into the output?

Yeah, that could be better. I made a PR for that #2181 :)

ryanking13:
We've merged #2181, so I guess `trailing-whitespace` won't complain anymore. Could you merge main, and enable it?

rth:
Thanks @henryiii !

ryanking13:
Thanks again @henryiii! This change is indeed quite convenient :-)

PR #2178, opened 2022-02-15
Support for C++ exceptions in packages

hoodmane:
See the discussion on #2174. An uncaught C++ error on main causes an error like:
```
TypeError: Cannot create property 'pyodide_fatal_error' on number '28811048'
```
With this the same error looks like:
```
CppException: Unable to cast Python instance to C++ type (compile in debug mode for details)
```


src/core/error_handling.ts
```diff
@@ -1,6 +1,96 @@
...
+  }
+  throw e;
+};
+
+class CppException extends Error {}
+Object.defineProperty(CppException.prototype, 'name', {
+  value: CppException.name,
+});
+
+function convertCppException(ptr : number): CppException {
```
ryanking13, line 86: Could you add a short comment on how function works?
I can understand it approximately, but [it looks like `CatchInfo` is not documented](https://github.com/emscripten-core/emscripten/commit/7f41d4344b6af0a7c98dee4d129268a38e42a10a), so a brief explanation of how cpp exception is handled would be helpful.

ryanking13 (approved):
Thanks @hoodmane, I have a minor comment, otherwise LGTM. Also, please update a changelog.

src/core/error_handling.ts
hoodmane: I changed how I did this because I think this approach wasn't actually quite right. I will mark this as resolved, let me know if you have comments on the new version.

hoodmane:
I should also add a test.

.circleci/config.yml
```diff
@@ -430,7 +430,7 @@ workflows:
 
       - test-main:
           name: test-core-chrome
-          test-params: -k "chrome and not webworker" src packages/micropip  packages/fpcast-test packages/sharedlib-test-py/
+          test-params: -k "chrome and not webworker" src packages/micropip  packages/fpcast-test packages/sharedlib-test-py/ packages/test_cpp_exceptions/
```
ryanking13, line 433: Myabe let's use hyphen instead of underscore for package name for consistency.

src/core/error_handling.ts
```diff
@@ -1,6 +1,101 @@
...
+Object.defineProperty(CppException.prototype, "name", {
+  value: CppException.name,
+});
+
+function convertCppException(ptr: number): CppException {
+  let msg;
+  try {
+    const msgPtr = Module._exc_what(ptr);
+    msg = Module.UTF8ToString(msgPtr);
+  } catch (e) {
```
ryanking13, line 91: It might be a dumb question, but how can we catch this?
It found that [std::exception::what has a noexcept specifier](https://en.cppreference.com/w/cpp/error/exception/what).
Does Emscripten ignores noexcept specifier, or are there other reasons?

ryanking13 (commented):
Thanks again, it looks like the test is failing because the new package's name starts with "test".

src/core/error_handling.ts
hoodmane: We're catching a segfault...

ryanking13: Oh... I got it. when it doesn't point to an exception it doesn't have `what` 🤦‍♀️

packages/test_cpp_exceptions/test_cpp_exceptions.py
```diff
@@ -0,0 +1,27 @@
+def test_sharedlib(selenium):
```
ryanking13, line 1: I guess the test name needs to be updated?

hoodmane:
I am going to try to upstream this functionality:
https://github.com/emscripten-core/emscripten/issues/16326

hoodmane:
Okay, this is ready for review. It's changed enough since @ryanking13 reviewed it last that it would be good for somebody to review it again.

henryiii:
You might want to edit the title, not very accurate anymore, and if it gets squashed and merged, it might not get changed. (See https://github.com/pybind/pybind11/commit/6493f496e30c80f004772c906370c8f4db94b6ec 🤦 )

rth:
LGTM, as well. Thanks for upstreaming it!

PR #2194, opened 2022-02-19
chore: move pytest config to pyproject.toml

henryiii:
### Description

This moves (as mentioned elsewhere) and hardens the pytest config. I expect to
probably have to filter or fix warnings, so I don't really expect this to
immediately pass.

### Checklists

- [ ] Add / update tests

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -33,3 +33,36 @@ known_first_party = [
   "pyodide_build",
   "_pyodide",
 ]
+
+
+[tool.pytest.ini_options]
+minversion = "6.0"
+norecursedirs = [
+  "build",
+  "cpython",
+  "emsdk",
+  ".artifacts",
+]
+addopts = [
+  "--doctest-modules",
+  "--ignore='packages/matplotlib/src'",
+  "--strict-markers",
+  "--strict-config",
+]
+markers = [
+  "skip_refcount_check: Dont run refcount checks",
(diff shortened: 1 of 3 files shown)


henryiii:
What should I do with:

```
pytest.PytestRemovedIn8Warning: Raising unittest.SkipTest to skip tests during collection is deprecated. Use pytest.skip() instead.
```

?

src/py/pyodide/http.py
```diff
@@ -208,7 +208,7 @@ async def unpack_archive(self, *, extract_dir=None, format=None):
 
 
 async def pyfetch(url: str, **kwargs) -> FetchResponse:
-    """Fetch the url and return the response.
+    r"""Fetch the url and return the response.
```
henryiii, line 211: FYI, I couldn't run this due to `\*` later, which is not a valid Python escape sequence. I just added an r here, but might be worth checking to see what the normal way to write `**kwargs` in this style of docstring is.

henryiii: It looks like just `**kwargs` is fine and doesn't mess with rst, but could be wrong. https://stackoverflow.com/questions/62511086/how-to-document-kwargs-according-to-numpy-style-docstring

hoodmane: I suspect that I added the backslashes to fix some issue with the generated output but I have repeatedly been wrong about this sort of thing so there's a good chance that they are pointless.

PR #2195, opened 2022-02-20
chore: pyupgrade 3.9+

henryiii:
### Description

This enables pyupgrade and upgrades older syntax to 3.9+ syntax.

Normally I'd use isort's include injection to add `from __future__ import annotations`, which provides a small performance benefit and a large readabilty benefit for codebases targeting older Python version, but a) the includes are not sorted by isort (would you like that) so this would sort _and_ inject, and b) with 3.9+, you actually get some nice readablity on typing already with the default generics - the main missing one is `Union[A, B]` -> `A | B` (which would come if/when this goes to 3.10+ as well). So this just focus on pyupgrade.

I forgot to exclude the micropip pip vendor, should that be excluded?

Also added a manual pre-commit way to clean up unused imports, which saved some time.

### Checklists

- [ ] Add / update tests

### DIFF
--- a/pyodide-build/pyodide_build/buildpkg.py
+++ b/pyodide-build/pyodide_build/buildpkg.py
@@ -19,7 +19,7 @@ from contextlib import contextmanager
 from datetime import datetime
 from pathlib import Path
 from textwrap import dedent
-from typing import Any, Dict
+from typing import Any
 from urllib import request
 
 from . import pywasmcross
@@ -59,7 +59,7 @@ class BashRunnerWithSharedEnvironment:
     def __init__(self, env=None):
         if env is None:
             env = dict(os.environ)
-        self.env: Dict[str, str] = env
+        self.env: dict[str, str] = env
         self._fd_read, self._fd_write = os.pipe()
         self._reader = os.fdopen(self._fd_read, "r")
 
@@ -113,7 +113,7 @@ def get_bash_runner():
         b.close()
 
 
-def check_checksum(archive: Path, source_metadata: Dict[str, Any]):
+def check_checksum(archive: Path, source_metadata: dict[str, Any]):
     """
     Checks that an archive matches the checksum in the package metadata.
 
@@ -144,7 +144,7 @@ def check_checksum(archive: Path, source_metadata: Dict[str, Any]):
             if len(chunk) < CHUNK_SIZE:
                 break
     if h.hexdigest() != checksum:
-        raise ValueError("Invalid {} checksum".format(checksum_algorithm))
+        raise ValueError(f"Invalid {checksum_algorithm} checksum")
 
 
 def trim_archive_extension(tarballname):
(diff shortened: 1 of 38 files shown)


hoodmane:
> the includes are not sorted by isort (would you like that)

I personally would like that. Normalizing the include order removes one pointless thing to worry about.

henryiii:
Okay, can do that in a follow up, and I agree. Reduces potential merge conflicts too, the same addition is always in the same place.

src/tests/test_console.py
```diff
@@ -429,7 +429,7 @@ def __repr__(self):
 
             >>> Test()
             [[;;;terminal-error]Traceback (most recent call last):
-              File \"/lib/python3.9/site-packages/_pyodide/console.py\", line 478, in repr_shorten
+              File \"/lib/python3.9/site-packages/_pyodide/console.py\", line 475, in repr_shorten
                 text = repr(value)
```
hoodmane, line 433: We really ought to remove this number from the string before making the comparison...

ryanking13:
Thanks, @henryiii!

> I forgot to exclude the micropip pip vendor, should that be excluded?

Yes, I think these should better be excluded.

> Normally I'd use isort's include injection to add `from __future__ import annotations`

If this line only affects Python versions < 3.10, I would prefer not to add it right now. Because we (hopefully) will upgrade Pyodide's Python version before long, and we don't maintain older Python compatibility.
Apart from that, sorting import orders with `isort` would be great.

src/tests/test_console.py
ryanking13: Yeah...

pyodide-build/pyodide_build/buildpkg.py
```diff
@@ -633,7 +633,7 @@ def build_package(
         return
 
     if not should_prepare_source and not srcpath.exists():
-        raise IOError(
+        raise OSError(
```
hoodmane, line 636: Why did `IOError` get replaced with `OSError`? Is `IOError` deprecated?

hoodmane (approved):

pyodide-build/pyodide_build/buildpkg.py
hoodmane: Oh:

> Changed in version 3.3: [EnvironmentError](https://docs.python.org/3/library/exceptions.html#EnvironmentError), [IOError](https://docs.python.org/3/library/exceptions.html#IOError), [WindowsError](https://docs.python.org/3/library/exceptions.html#WindowsError), [socket.error](https://docs.python.org/3/library/socket.html#socket.error), [select.error](https://docs.python.org/3/library/select.html#select.error) and mmap.error have been merged into [OSError](https://docs.python.org/3/library/exceptions.html#OSError), and the constructor may return a subclass.

hoodmane:
```py
>>> IOError is OSError
True
```

hoodmane:
Thanks again @henryiii for your maintenance help. We appreciate it a lot.

PR #2176, opened 2022-02-14
fix(mkpkg): arguments were incorrect

henryiii:
### Description

mkpkg was broken after #2126.

### Checklists

- [ ] Add a [CHANGELOG](https://github.com/pyodide/pyodide/blob/main/docs/project/changelog.md) entry
- [ ] Add / update tests
- [ ] Add new / update outdated documentation

Ideally mypy should catch this, but it seems `pre-commit` is not being run/checked, quite a few things fail:

```console
$ pre-commit run -a
Check Yaml...............................................................Failed
- hook id: check-yaml
- exit code: 1

expected a single document in the stream
  in ".clang-format", line 2, column 1
but found another document
  in ".clang-format", line 110, column 1

Fix End of Files.........................................................Passed
Trim Trailing Whitespace.................................................Failed
- hook id: trailing-whitespace
- exit code: 1
- files were modified by this hook

Fixing packages/CLAPACK/make.inc
Fixing packages/micropip/src/micropip/package.py
Fixing src/py/_pyodide/console.py
Fixing .prettierignore

black....................................................................Failed
- hook id: black
- exit code: 123
- files were modified by this hook

reformatted /Users/henryschreiner/git/software/pyodide/benchmark/benchmarks/log_likelihood.py
reformatted /Users/henryschreiner/git/software/pyodide/benchmark/benchmarks/vibr_energy.py
reformatted /Users/henryschreiner/git/software/pyodide/benchmark/benchmarks/lstsqr.py
reformatted /Users/henryschreiner/git/software/pyodide/packages/cloudpickle/test_cloudpickle.py
reformatted /Users/henryschreiner/git/software/pyodide/packages/sympy/test_sympy.py
reformatted /Users/henryschreiner/git/software/pyodide/benchmark/benchmarks/periodic_dist.py
reformatted /Users/henryschreiner/git/software/pyodide/packages/nlopt/test_nlopt.py
reformatted /Users/henryschreiner/git/software/pyodide/benchmark/benchmarks/wave.py
error: cannot format /Users/henryschreiner/git/software/pyodide/pyodide-build/pyodide_build/_f2c_fixes.py: INTERNAL ERROR: Black produced code that is not equivalent to the source.  Please report a bug on https://github.com/psf/black/issues.  This diff might be helpful: /var/folders/_8/xtbws09n017fbzdx9dmgnyyr0000gn/T/blk_fewlf2sd.log
[...]

PR #2204, opened 2022-02-23
tests: enabling mypy on untyped defs

henryiii:
### Description

This enables mypy checking inside all untyped functions. This would have helped catch https://github.com/pyodide/pyodide/pull/2176 before it happened. Will place a few comments inline.

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -113,7 +113,7 @@ class SeleniumWrapper:
         self.server_hostname = server_hostname
         self.base_url = f"http://{self.server_hostname}:{self.server_port}"
         self.server_log = server_log
-        self.driver = self.get_driver()
+        self.driver = self.get_driver()  # type: ignore[attr-defined]
         self.set_script_timeout(script_timeout)
         self.script_timeout = script_timeout
         self.prepare_driver()
@@ -562,6 +562,7 @@ def selenium_common(request, web_server_main, load_pyodide=True):
     """
 
     server_hostname, server_port, server_log = web_server_main
+    cls: type[SeleniumWrapper]
     if request.param == "firefox":
         cls = FirefoxWrapper
     elif request.param == "chrome":
@@ -569,7 +570,7 @@ def selenium_common(request, web_server_main, load_pyodide=True):
     elif request.param == "node":
         cls = NodeWrapper
     else:
-        assert False
+        raise AssertionError(f"Unknown browser: {request.param}")
     selenium = cls(
         server_port=server_port,
         server_hostname=server_hostname,
@@ -687,7 +688,7 @@ def spawn_web_server(build_dir=None):
 
     tmp_dir = tempfile.mkdtemp()
     log_path = pathlib.Path(tmp_dir) / "http-server.log"
-    q = multiprocessing.Queue()
+    q: multiprocessing.Queue[str] = multiprocessing.Queue()
     p = multiprocessing.Process(target=run_web_server, args=(q, log_path, build_dir))
 
     try:
@@ -740,8 +741,8 @@ def run_web_server(q, log_filepath, build_dir):
     with socketserver.TCPServer(("", 0), Handler) as httpd:
         host, port = httpd.server_address
         print(f"Starting webserver at http://{host}:{port}")
-        httpd.server_name = "test-server"
-        httpd.server_port = port
+        httpd.server_name = "test-server"  # type: ignore[attr-defined]
+        httpd.server_port = port  # type: ignore[attr-defined]
         q.put(port)
 
         def service_actions():
@@ -752,7 +753,7 @@ def run_web_server(q, log_filepath, build_dir):
             except queue.Empty:
                 pass
 
-        httpd.service_actions = service_actions
+        httpd.service_actions = service_actions  # type: ignore[assignment]
         httpd.serve_forever()
 
 
--- a/docs/sphinx_pyodide/sphinx_pyodide/packages.py
+++ b/docs/sphinx_pyodide/sphinx_pyodide/packages.py
@@ -61,7 +61,7 @@ def get_packages_summary_directive(app):
             )
 
         def format_packages_table(
-            self, packages: dict[str, Any], columns: tuple[str]
+            self, packages: dict[str, Any], columns: tuple[str, ...]
         ) -> list[Any]:
             table_spec = addnodes.tabular_col_spec()
             table_spec["spec"] = r"\X{1}{2}\X{1}{2}"
--- a/packages/wrapt/test_wrapt.py
+++ b/packages/wrapt/test_wrapt.py
@@ -3,6 +3,7 @@ from pyodide_build.testing import run_in_pyodide
 
 @run_in_pyodide(packages=["wrapt"])
 def test_wrapt():
+    import inspect
     import unittest
 
     import wrapt
--- a/pyodide-build/pyodide_build/buildall.py
+++ b/pyodide-build/pyodide_build/buildall.py
@@ -371,7 +371,9 @@ def build_from_graph(pkg_map: dict[str, BasePackage], outputdir: Path, args) -> 
     built_queue: Queue = Queue()
     thread_lock = Lock()
     queue_idx = 1
-    package_set: dict[str, None] = {}  # TODO: shoudn't this be a set?
+    package_set: dict[
+        str, None
+    ] = {}  # using dict keys for insertion order preservation
 
     def builder(n):
         nonlocal queue_idx
--- a/pyodide-build/pyodide_build/buildall.py
+++ b/pyodide-build/pyodide_build/buildall.py
@@ -371,7 +371,7 @@ def build_from_graph(pkg_map: dict[str, BasePackage], outputdir: Path, args) -> 
     built_queue: Queue = Queue()
     thread_lock = Lock()
     queue_idx = 1
-    package_set = {}
+    package_set: dict[str, None] = {}  # TODO: shoudn't this be a set?
 
     def builder(n):
         nonlocal queue_idx
--- a/pyodide-build/pyodide_build/buildpkg.py
+++ b/pyodide-build/pyodide_build/buildpkg.py
@@ -18,7 +18,7 @@ from contextlib import contextmanager
 from datetime import datetime
 from pathlib import Path
 from textwrap import dedent
-from typing import Any
+from typing import Any, Optional
 from urllib import request
 
 from . import pywasmcross
@@ -55,6 +55,9 @@ class BashRunnerWithSharedEnvironment:
     directly to adjust the environment, or read to get variables.
     """
 
+    _fd_read: Optional[int]
+    _fd_write: Optional[int]
+
     def __init__(self, env=None):
         if env is None:
             env = dict(os.environ)
@@ -81,7 +84,7 @@ class BashRunnerWithSharedEnvironment:
 
     def close(self):
         """Free the file descriptors."""
-        if self._fd_read:
+        if self._fd_read and self._fd_write:
             os.close(self._fd_read)
             os.close(self._fd_write)
             self._fd_read = None
--- a/pyodide-build/pyodide_build/io.py
+++ b/pyodide-build/pyodide_build/io.py
@@ -1,5 +1,5 @@
 from pathlib import Path
-from typing import Any, Optional
+from typing import Any, Optional, Union
 
 # TODO: support more complex types for validation
 
@@ -46,7 +46,9 @@ PACKAGE_CONFIG_SPEC: dict[str, dict[str, Any]] = {
 
 
 def check_package_config(
-    config: dict[str, Any], raise_errors: bool = True, file_path: Optional[Path] = None
+    config: dict[str, Any],
+    raise_errors: bool = True,
+    file_path: Optional[Union[Path, str]] = None,
 ) -> list[str]:
     """Check the validity of a loaded meta.yaml file
 
@@ -118,7 +120,7 @@ def check_package_config(
     return errors_msg
 
 
-def parse_package_config(path: Path, check: bool = True) -> dict[str, Any]:
+def parse_package_config(path: Union[Path, str], check: bool = True) -> dict[str, Any]:
     """Load a meta.yaml file
 
     Parameters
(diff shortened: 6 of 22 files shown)


conftest.py
```diff
@@ -562,14 +562,15 @@ def selenium_common(request, web_server_main, load_pyodide=True):
...
+    cls: type[SeleniumWrapper]
     if request.param == "firefox":
         cls = FirefoxWrapper
     elif request.param == "chrome":
         cls = ChromeWrapper
     elif request.param == "node":
         cls = NodeWrapper
     else:
-        assert False
+        raise AssertionError(f"Unknown browser: {request.param}")
```
henryiii, line 573: You shouldn't depend on assert False to catch, assertions could be disabled. Assertions should never "happen", (and) be used to control mypy.

docs/sphinx_pyodide/sphinx_pyodide/packages.py
```diff
@@ -61,7 +61,7 @@ def get_package_metadata_list(
             )
 
         def format_packages_table(
-            self, packages: dict[str, Any], columns: tuple[str]
+            self, packages: dict[str, Any], columns: tuple[str, ...]
```
henryiii, line 64: This was previously incorrect, it's not always exactly one column.

pyodide-build/pyodide_build/buildall.py
```diff
@@ -371,7 +371,7 @@ def build_from_graph(pkg_map: dict[str, BasePackage], outputdir: Path, args) ->
     built_queue: Queue = Queue()
     thread_lock = Lock()
     queue_idx = 1
-    package_set = {}
+    package_set: dict[str, None] = {}  # TODO: shoudn't this be a set?
```
henryiii, line 374: Shouldn't this be a set? IT's just assigning Nones, and the name literally includes "set".

pyodide-build/pyodide_build/buildpkg.py
```diff
@@ -81,7 +84,7 @@ def run(self, cmd, **opts):
 
     def close(self):
         """Free the file descriptors."""
-        if self._fd_read:
+        if self._fd_read and self._fd_write:
```
henryiii, line 87: Both should always be true, and this protects the later part for mypy. This would be valid place for an assert, instead.

pyodide-build/pyodide_build/io.py
```diff
@@ -46,7 +46,9 @@
 
 
 def check_package_config(
-    config: dict[str, Any], raise_errors: bool = True, file_path: Optional[Path] = None
+    config: dict[str, Any],
+    raise_errors: bool = True,
+    file_path: Optional[Union[Path, str]] = None,
```
henryiii, line 51: Some of the Path arguments here also were being given strings.

pyodide-build/pyodide_build/testing.py
```diff
@@ -127,7 +127,9 @@ def wrapped(selenium):  # type: ignore
 
 
 @contextlib.contextmanager
-def set_webdriver_script_timeout(selenium, script_timeout: Optional[Union[int, float]]):
+def set_webdriver_script_timeout(
+    selenium, script_timeout: Optional[Union[int, float, str]]
```
henryiii, line 131: Hopefully strings are allowed, because it's in the types elsewhere!

pyodide-build/pyodide_build/tests/test_buildall.py
```diff
@@ -62,7 +62,7 @@ def build(self, outputdir: Path, args) -> None:
 
     pkg_map = buildall.generate_dependency_graph(PACKAGES_DIR, {"lxml", "micropip"})
 
-    Args = namedtuple("args", ["n_jobs", "force_rebuild"])
+    Args = namedtuple("Args", ["n_jobs", "force_rebuild"])
```
henryiii, line 65: These should always match.

pyodide-build/pyodide_build/tests/test_mkpkg.py
```diff
@@ -61,12 +62,12 @@ def test_mkpkg_update(tmpdir, monkeypatch, old_dist_type, new_dist_type):
     source_fmt = new_dist_type
     if new_dist_type == "same":
         source_fmt = None
-    pyodide_build.mkpkg.update_package("idna", None, source_fmt)
+    pyodide_build.mkpkg.update_package("idna", False, source_fmt)
```
henryiii, line 65: This is a bool, and checked as such.

src/py/_pyodide/console.py
```diff
@@ -259,6 +259,7 @@ def persistent_redirect_streams(self):
         if self._stream_generator:
             return
         self._stream_generator = self._stdstreams_redirections_inner()
+        assert self._stream_generator is not None
```
henryiii, line 262: Example of type narrowing using assert.

src/py/lib/pystone.py
```diff
@@ -137,7 +137,7 @@ def Proc0(loops=LOOPS):
                 EnumLoc = Proc6(Ident1)
             CharIndex = chr(ord(CharIndex) + 1)
         IntLoc3 = IntLoc2 * IntLoc1
-        IntLoc2 = IntLoc3 / IntLoc1
+        IntLoc2 = IntLoc3 // IntLoc1
```
henryiii, line 140: This should return an int, so you need `//` in Python 3.

src/py/lib/pystone.py
```diff
@@ -183,17 +184,17 @@ def Proc3(PtrParOut):
 
     if PtrGlb is not None:
         PtrParOut = PtrGlb.PtrComp
+        PtrGlb.IntComp = Proc7(10, IntGlob)
     else:
         IntGlob = 100
-    PtrGlb.IntComp = Proc7(10, IntGlob)
```
henryiii, line 188: PtrGlb must not be None if you are assigning an attribute to it.

src/tests/test_package_loading.py
```diff
@@ -270,8 +270,10 @@ def test_test_unvendoring(selenium_standalone):
 def test_install_archive(selenium):
     build_dir = Path(__file__).parents[2] / "build"
     test_dir = Path(__file__).parent
+    # TODO: first arguement actually works as a path due to implementation,
+    # maybe it can be proposed to typeshed?
     shutil.make_archive(
-        test_dir / "test_pkg", "gztar", root_dir=test_dir, base_dir="test_pkg"
+        str(test_dir / "test_pkg"), "gztar", root_dir=test_dir, base_dir="test_pkg"
```
henryiii, line 276: Worth checking with typeshed, as this is technically allowed due to the impl, but is not indicated in the Python docs.

src/tests/test_testing.py
```diff
@@ -35,7 +35,7 @@ def func(f):
         "chrome": "nlopt set_min_objective triggers a fatal runtime error in chrome 89 see #1493",
     },
 )
-def some_func():
+def some_func(f):
```
henryiii, line 38: f was undefined, hopefully not intentional?

pyodide-build/pyodide_build/buildall.py
hoodmane: I think iteration over dictionaries is in insertion order whereas iteration over sets is not. I believe this is why we use a dictionary here.

conftest.py
hoodmane: Assertions can be disabled?

pyodide-build/pyodide_build/testing.py
hoodmane: I'm guessing that strings are not allowed. Here's the documentation for the selenium function we are calling:
https://selenium-python.readthedocs.io/api.html#selenium.webdriver.remote.webdriver.WebDriver.set_script_timeout
It would be weird if it accepted strings.

conftest.py
henryiii: Yes, `python -O` turn off assertions. If you were running pylint, it would balk at this, telling you `assert False` should never be used, and you should raise an AssertionError instead.

src/tests/test_testing.py
hoodmane: Nope, definitely doesn't look intentional. Of course we never plan to call this function so it doesn't matter too much.

pyodide-build/pyodide_build/testing.py
```diff
@@ -49,7 +49,7 @@ def run_in_pyodide(
         Whether to use a standalone selenium instance to run the test or not
     packages : List[str]
         List of packages to load before running the test
-    driver_timeout : Optional[Union[str, int]]
+    driver_timeout : Optional[int]
```
henryiii, line 52: This apparently wasn't supposed to support strings, it was being passed to selenium.

hoodmane: Probably the best type here would be float or int. Thanks for catching it!

packages/wrapt/test_wrapt.py
```diff
@@ -3,6 +3,7 @@
 
 @run_in_pyodide(packages=["wrapt"])
 def test_wrapt():
+    import inspect
```
hoodmane, line 6: Is this needed?

pyodide-build/pyodide_build/testing.py
henryiii: Hmm, float would likely be better?

src/py/lib/pystone.py
hoodmane: This file is an odd one. We only use it in `benchmarks.py` I wonder if we could replace it with something else that's less like... whatever this file is like. I think of it as an external file, so I have avoided doing any sort of maintenance on it.

packages/wrapt/test_wrapt.py
henryiii: `inspect` is undefined otherwise. I believe this test never runs (see #2174).

src/py/lib/pystone.py
henryiii: We can always add files to the exclude list if you would rather, potentially for all of pre-commit checks (though mypy is a little harder, as if it's imported from a file we are checking, it has to also be handled there).

rth: Yeah, it should be added to the exclude list. It's a copy from https://github.com/blackberry/Python/blob/master/Python-3/Lib/test/pystone.py

src/tests/test_console.py
```diff
@@ -229,7 +229,7 @@ def test_nonpersistent_redirection(safe_sys_redirections):
     my_stdout = ""
     my_stderr = ""
 
-    def stdin_callback():
+    def stdin_callback(string):
```
hoodmane, line 232: This isn't right. `stdin_callback` should take no inputs and return a string. I guess the problem is that line 228 in `src/py/_pyodide/console.py` is wrong. It currently reads:
```py
stdin_callback: Optional[Callable[[str], None]] = None,
```
and should be changed to:
```py
stdin_callback: Optional[Callable[[], str]] = None
```

hoodmane: I definitely thought mypy was running on these files.

packages/wrapt/test_wrapt.py
hoodmane: Oh yeah, I didn't notice it is used below. I forgot about this, thanks for the reminder.

pyodide-build/pyodide_build/testing.py
hoodmane: If I have:
```py
def f(a : float): ...
```
and I call `f(5)` is that a type error? What about `n : int = 5; f(n)`?

packages/wrapt/test_wrapt.py
rth: We should open an issue about it, as CI says that test_wapt.py passes, it probably doest run as expected.

src/tests/test_console.py
henryiii: Ahh, yes, that's the correct fix. MyPy only runs on things that have types (until this PR, which runs it on untyped functions and classes)

packages/wrapt/test_wrapt.py
henryiii: I wasn't able to figure out how to get this style of test to run in boost-histogram, so I just rewrote it. Pretty sure you need to manually force the class to run it's tests.

hoodmane (approved):
LGTM once we fix the types for the `stdin_callback` argument to `Console.__init__` and for `driver_timeout`.

Thanks again @henryiii!

src/tests/test_console.py
hoodmane: Right hmm. Is there a flag to require types for everything? I suppose we could enable that in a subsequent PR.

pyodide-build/pyodide_build/testing.py
henryiii: From [PEP 0484](https://www.python.org/dev/peps/pep-0484/#the-numeric-tower):

> Rather than requiring that users write import numbers and then use `numbers.Float` etc., this PEP proposes a straightforward shortcut that is almost as effective: when an argument is annotated as having type `float`, an argument of type `int` is acceptable;

src/tests/test_console.py
henryiii: That's

```
check_untyped_defs = true
```

And it's the main feature of this PR. :)

src/py/lib/pystone.py
henryiii: Added this file and the extern directory to the universal exclude list.

packages/wrapt/test_wrapt.py
hoodmane: Can't we just add `unittest.main()` to the end?

hoodmane: The tests seem to run successfully and fail somewhere but then it results in a useless error message...

henryiii: Didn't seem to work for me with boost-histogram, but I might have had another issue, so should try it.

hoodmane:
I will merge this when CI passes.

hoodmane:
One issue with this is that our lint phase times out in CI now. Ideally we should cache the pre-commit environment.

src/tests/test_console.py
hoodmane: > And it's the main feature of this PR.

Oh... right. Thanks for your patience @henryiii :laughing:

PR #2205, opened 2022-02-23
Add ruamel.yaml to the mandatory dependency of pyodide-build

ryanking13:
### Description

Adds `ruamel.yaml` to the dependency of pyodide-build, so that one doesn't need to additionally install it to run `pyodide-build mkpkg`

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -9,6 +9,7 @@ import subprocess
 import sys
 from pathlib import Path
 from typing import Any
+from unittest import mock
 
 # -- Project information -----------------------------------------------------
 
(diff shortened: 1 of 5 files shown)


ryanking13:
I don't clearly understand these lines.

https://github.com/pyodide/pyodide/blob/66856a74273dfc6634809fecfc80412bea6c600a/packages/Makefile#L9-L18

`pyodide-build` is installed to the `HOSTINSTALLDIR` (.artifacts), ignoring dependencies. But the built binary is not used. Is this intended behavior?

ryanking13:
I'll open a new PR, along with more fixes on `pyodide-build`.

PR #2208, opened 2022-02-23
Add `default_converter` argument to `to_js` and `toJs`

hoodmane:
This is the mirror image of #2170.

### Checklists

- [x] Add / update tests


src/core/js2python.js
```diff
@@ -312,9 +312,6 @@ JS_FILE(js2python_init, () => {
         context.converter,
         context.cacheConversion
       );
-      if (result_js === undefined) {
```
hoodmane, line 315: This was accidentally left in #2170. I guess maybe I should add a test for it.

src/core/python2js.c
```diff
@@ -557,45 +525,134 @@ EM_JS_REF(JsRef, _JsArray_PostProcess_helper, (JsRef jscontext, JsRef array), {
     Hiwire.get_value(jscontext).dict_converter(Hiwire.get_value(array)));
 })
 
+// clang-format off
```
ryanking13, line 528: Is there some reason to disable formatter here?

ryanking13 (commented):
Thanks, @hoodmane. Please add tests and update changelog.  [docstring for `to_js`](https://github.com/pyodide/pyodide/blob/57a41af11811d94947fe3c7a78a0119c8ca52c19/src/py/_pyodide/_core_docs.py#L322) needs to be updated too.

src/core/python2js.c
hoodmane: Yeah, clang-format sometimes does annoying things to em_js. Try removing it and running clang-format and you'll see why I don't like it. The whole function body gets indented by 10 spaces or something like that.

hoodmane:
Thanks for the review @ryanking13. I have a test on my local copy but it fails because of a reference leak which I haven't tracked down yet.

src/core/jsproxy.c
```diff
@@ -719,10 +722,10 @@ JsProxy_toPy(PyObject* self,
   }
   PyObject* result =
     js2python_convert(GET_JSREF(self), depth, default_converter_js);
-  if (default_converter_js != NULL) {
+  if (pyproxy_Check(default_converter_js)) {
```
ryanking13, line 725: BTW, unrelated to this PR, why is this `pyproxy_Check` function named in a strange mixed-case...?

src/core/python2js.c
```diff
@@ -557,45 +525,135 @@ EM_JS_REF(JsRef, _JsArray_PostProcess_helper, (JsRef jscontext, JsRef array), {
...
+  if (dict_converter !== 0) {
+    jscontext.dict_converter = Hiwire.get_value(dict_converter);
+  }
+  if (default_converter !== 0) {
+    jscontext.default_converter = Hiwire.get_value(default_converter);
+    jscontext.cacheConversion = function (input, output) {
+      // input should be a PyProxy, output should be a Javascript
+      // object
+      if (!API.isPyProxy(input)) {
+        throw new Error("Oops!");
```
ryanking13, line 571: A little more concrete error message would be great.

src/core/pyproxy.ts
```diff
@@ -424,12 +425,22 @@ export class PyProxyClass {
...
+    dict_converter?: (array: Iterable<[key: string, value: any]>) => any;
+    /**
+     * Optional argument to convert objects with no default conversion. See the
+     * documentation of :any:`pyodide.to_js`.
+     */
+    default_converter?: (
+      obj: PyProxy,
+      convert: (obj: PyProxy) => any,
+      cacheConversion: (obj: PyProxy, result: any) => void
+    ) => any;
```
ryanking13, line 437: It looks like the [docstrings are not properly parsed](https://pyodide--2208.org.readthedocs.build/en/2208/usage/api/js-api.html#PyProxy.toJs). If this is an issue about compatibility between typescript and sphinx-js, we could fix this in a separate PR as there are some other functions where their docstrings are not properly parsed as well.

ryanking13 (approved):
Thanks again @hoodmane! Generally LGTM with few minor comments.

src/core/python2js.c
hoodmane: Oh yeah I meant to go back and fix this 😂

src/core/pyproxy.ts
hoodmane: I monkey patched in the handling for this so I probably made a mistake. It seems like sphinx-js is not Python 3.10 compatible (from ci in #2225) so we probably need to fork it soon.

hoodmane: Thanks for noticing this.

src/core/jsproxy.c
hoodmane: I think it should ideally be `PyProxy_Check`. Though it is confusing that a PyProxy isn't a Python type. I am trying to follow Python naming conventions they have eg PyType_Check and a bunch of other functions named similarly. But my naming is not all that consistent in general...

src/core/pyproxy.ts
ryanking13: > so we probably need to fork it soon

Well, that's sad news 😥

hoodmane: Actually, turns out I can monkey patch Python3.10 compatibility =)

src/py/_pyodide/_core_docs.py
```diff
@@ -365,6 +368,72 @@ def to_js(
...
+        referential pair:
+
+        .. code-block:: javascript
+
+            p = Pair(0, 0);
+            p.first = p;
+
+        Without ``cache(value, result);``, converting ``p`` would lead to an
+        infinite recurse. With it, we can successfully convert ``p`` to an Array
+        such that ``l[0] === l``.
```
rth, line 436: This might a bit long for a docstring. I wonder if we should move some of it to a user manual section in type translations about default conversions. Though it does seem very specific to this particular argument so not sure. Feel free not to do anything.

rth (commented):
API wise LGTM, I have not reviewed the code, but if @ryanking13 is happy, +1 to merge.

PR #2209, opened 2022-02-23
chore: update pre-commit hooks

henryiii:
- chore: update pre-commit hooks
- style: run pre-commit -a, javascript now included

### Description

This bumps the pre-commit hooks. This was to avoid running into https://github.com/pre-commit/mirrors-clang-format/pull/10, but I noticed that you include JavaScript in .clang-format, so I'm guessing that's supposed to be included? It was missed in the original pre-commit runs. Assuming JSON should not be included? Or would that count under JavaScript?

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -107,6 +107,4 @@ SpacesInSquareBrackets: false
 Standard:        Cpp11
 TabWidth:        8
 UseTab:          Never
----
-Language: JavaScript
 ...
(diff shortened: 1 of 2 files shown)


hoodmane:
Nowadays we format Javascript and JSON with prettier (as well as anything else prettier is willing to format), not with clang-format. The `clang-format` rules for Javascript predate this switch. We do still have some Javascript code which is formatted as C code by `clang-format` because it's in `.c` files. This is annoying, but it's probably too much to ask for a formatter that recognizes the inside of `EM_JS` and `EM_ASM` as Javascript. At least we only have a bit of this,

henryiii:
Okay, I'll rebase with `JavaScript` removed, and remove the three javascript lines from `clang-format` if you are okay with that too.

PR #2220, opened 2022-02-26
Use ESM import in module type webworker

leopsidom:
### Description
The PR is to avoid `importScripts` error when loading pyodide in a module type worker. Currently `loadScripts`  uses `importScripts` to import dependency  which throws an error when the worker is of type `module`.

### Checklists

- [x] Add / update tests

### DIFF
--- a/src/tests/test_python.py
+++ b/src/tests/test_python.py
@@ -2,9 +2,7 @@ import pytest
 
 
 def test_init(selenium_standalone):
-    assert (
-        "Python initialization complete" in selenium_standalone.logs.splitlines()
-    )
+    assert "Python initialization complete" in selenium_standalone.logs.splitlines()
 
 
 def test_webbrowser(selenium):


hoodmane:
Thanks @leopsidom. Maybe it would be better instead to test `typeof loadScript !== "undefined"` in the conditional.

hoodmane:
> the issue is that importScripts is still defined in both classic and module type web worker. It just throws an error when the web worker is of type module at run time.

That is unfortunate... Is there any other direct means one can use to distinguish between a module worker and a classic worker? I looked through [the whatwg worker spec](https://html.spec.whatwg.org/multipage/workers.html) and I think the answer is no. Hopefully this is the only reason that `importScripts` would raise a `TypeError` though.

hoodmane (approved):
Ideally we should add a test for importing Pyodide as a module, both in a module worker and on the main thread, or else it might get broken again.

leopsidom:
> Ideally we should add a test for importing Pyodide as a module, both in a module worker and on the main thread, or else it might get broken again.

Yeah, that makes sense. I'll take a look at adding tests for module import, though the tests won't work for module type worker in firefox yet as it's still catching up with the support at the moment.

hoodmane:
> I'll take a look at adding tests for module import

If you get stuck, let me know. You will probably want the `selenium_webworker_standalone` fixture, but it might be possible to use it with module scope instead of function scope which would decrease time to execute tests. Though it's good to just get something working for now and we can futz with that sort of thing later if we feel like it.

>  the tests won't work for module type worker in firefox yet

Sure. You can xfail it in firefox or something.

leopsidom:
Sure, just read a little about module and function scope, I'm not completely sure that I understands the difference. 

For the tests, I'm thinking we could add a type parameter to [the `SeleniumWrapper` class](https://github.com/pyodide/pyodide/blob/8d00451142d5ddb0237595065773b1f7bf290d33/conftest.py#L101), which  can be further passed to [get_driver](https://github.com/pyodide/pyodide/blob/8d00451142d5ddb0237595065773b1f7bf290d33/conftest.py#L357) and [run_webworker](https://github.com/pyodide/pyodide/blob/8d00451142d5ddb0237595065773b1f7bf290d33/conftest.py#L302)  to import pyodide in the classic and module manner. Then we can create a separate `selenium_module_webworker_standalone` and `selenium_module_standalone` as counter part of [`selenium_webworker_standalone`](https://github.com/pyodide/pyodide/blob/8d00451142d5ddb0237595065773b1f7bf290d33/conftest.py#L617) and [`selenium_standalone`](https://github.com/pyodide/pyodide/blob/8d00451142d5ddb0237595065773b1f7bf290d33/conftest.py#L587). With which we can write minimum enough tests to just make sure pyodide is loaded correctly in the module syntax. This won't reuse existing tests, so potentially not enough coverage but  avoid redundant tests with assumption that after imported,  the script should behave identically in both classic and module type.

hoodmane:
Sounds like a good plan to me. 

There's a tradeoff we don't want to run the entire test suite in every combination of browser and webworker configuration since it would take too long. Just testing that we can successfully load Pyodide would catch the problem you reported.

hoodmane:
The issue with fixture scope is essentially how often we reload the browser tab. Refreshing the browser tab provides better isolation between tests but at the cost of taking much longer to run them.

leopsidom:
Cool. Got  it. I think as long as we don't get issues on tests interacting with each other, we can definitely change the scope so it saves time on refreshing the context.

Makefile
```diff
@@ -104,6 +105,8 @@ src/js/pyproxy.gen.ts : src/core/pyproxy.* src/core/*.h
 build/test.html: src/templates/test.html
 	cp $< $@
 
+build/module_test.html: src/templates/module_test.html
```
hoodmane, line 108: At some point we probably should try to combine some of these templates build rules (no action needed in this PR).

src/templates/module_test.html
```diff
@@ -0,0 +1,28 @@
...
+    };
+    console.info = function (message) {
+      window.logs.push(message);
+    };
+    console.error = function (message) {
+      window.logs.push(message);
+    };
+  </script>
+  <script type="module">
+    import { loadPyodide } from './pyodide.mjs'
```
hoodmane, line 22: Won't you need to add this to the global scope? Otherwise it won't be possible to run `loadPyodide` from `selenium.run_js`.

leopsidom: I think when we `import`, `loadPyodide` is automatically added to `global`.  But let's see how the tests run.

src/tests/test_webworker_module.py
```diff
@@ -0,0 +1,70 @@
+import pytest
+
+
+def test_runwebworker_different_package_name(selenium_module_webworker_standalone):
```
hoodmane, line 4: Rather than duplicating `test_webworker.py` we should use `@pytest.mark.parametrize` to run the same test in both module and classic variants.

leopsidom: Good point! Let me see if I can combine them.

src/templates/module_test.html
hoodmane: I think the `<script type="module">` comes with it's own scope and everything in `<script type="module">` is limited to this scope unless otherwise stated.

leopsidom: Ahh I think you're correct. Hmm ~ Let me fix that

src/tests/test_webworker_module.py
leopsidom: I ended up creating another fixture `script_type` to create the combination of classic / module vs firefox / chrome. Tried to use `@pytest.mark.parameterize` on the test but it overrides the `request.param` that `selenium_webworker_standalone` fixture already has (firefox/chrome), in which case if we pass the parameters to the test function, we need to pass in all combinations -- classic-firfox / classic-chrome / module-firefox / module-chrome, but that also changes the structure of `request.param`. So it seems creating  another fixture and pass it to the `selenium_webworker_standalone` fixture and test functions is the most straightforward way at the moment. But open to other options if we prefer to modify the `request.params` type.

hoodmane: This way is fine with me.

leopsidom:
The `test_draw_math_text` in `test-packages-firefox` seems to fail pretty consistently in other PR:  like [this](https://app.circleci.com/pipelines/github/pyodide/pyodide/3602/workflows/6b63d0f9-d110-4748-a07e-04c966122407/jobs/39062/steps) from [the latest commit](https://github.com/pyodide/pyodide/commit/e3760e81105c590285a2f1908316a0951e0496c5). Even for the passed test, there [appears to be a retry](https://app.circleci.com/pipelines/github/ryanking13/pyodide/205/workflows/64669409-6e65-4d1c-a8df-d1a4c1f75559/jobs/3003).

leopsidom:
Similar for [this workflow](https://app.circleci.com/pipelines/github/pyodide/pyodide/3574/workflows/f37ab0ae-3c18-495d-aa16-ad27332cab11/jobs/38712) from [this commit](https://github.com/pyodide/pyodide/commit/8d00451142d5ddb0237595065773b1f7bf290d33).

hoodmane:
Yeah the test failure has nothing to do with this PR.

hoodmane:
In particular, that means we are willing to merge this even with the test failure.

leopsidom:
> In particular, that means the test failure won't prevent us from merging this.

Yeah it seems to be failing inherently due to an unknown issue. So I guess we shouldn't let it block us from merging new changes.

hoodmane:
xfailing chrome for the recursion limit must be out of date, we fixed that cf #2019

hoodmane:
But yeah, perhaps just turning that benchmark off would be a good start.

ryanking13:
> The `test_draw_math_text` in `test-packages-firefox` seems to fail pretty consistently in other PR: like [this](https://app.circleci.com/pipelines/github/pyodide/pyodide/3602/workflows/6b63d0f9-d110-4748-a07e-04c966122407/jobs/39062/steps) from [the latest commit](https://github.com/pyodide/pyodide/commit/e3760e81105c590285a2f1908316a0951e0496c5). Even for the passed test, there [is a retry after the initial time out failure](https://app.circleci.com/pipelines/github/ryanking13/pyodide/205/workflows/64669409-6e65-4d1c-a8df-d1a4c1f75559/jobs/3003).

I am looking into it! sorry for the confusion :) (#2226)

src/tests/test_webworker.py
```diff
@@ -12,8 +16,10 @@ def test_runwebworker_different_package_name(selenium_webworker_standalone):
     assert isinstance(output, str)
 
 
-def test_runwebworker_no_imports(selenium_webworker_standalone):
+def test_runwebworker_no_imports(selenium_webworker_standalone, script_type):
     selenium = selenium_webworker_standalone
+    if selenium.browser == "firefox" and script_type == "module":
```
hoodmane, line 21: Since you do this on every single test that uses `selenium_webworker_standalone`, wouldn't it be better to move this into `selenium_webworker_standalone`?

src/tests/test_python_module.py
```diff
@@ -0,0 +1,33 @@
+import pytest
+
+
+def test_init(selenium_module_standalone):
```
hoodmane, line 4: It would be preferable to use a module-scoped fixture for these so we don't reload it three times. These tests don't need to be isolated from each other. (The `test_init` in `test_python.py` really shouldn't use `selenium_standalone` either...)

leopsidom: It turns out `test_init` uses `selenium_standalone` because the module scoped `selenium` fixture [clean logs](https://github.com/pyodide/pyodide/blob/main/conftest.py#L647) which can cause the test to fail -

src/templates/module_test.html
```diff
@@ -0,0 +1,26 @@
...
+      };
+      console.warn = function (message) {
+        window.logs.push(message);
+      };
+      console.info = function (message) {
+        window.logs.push(message);
+      };
+      console.error = function (message) {
+        window.logs.push(message);
+      };
```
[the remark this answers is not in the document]
hoodmane: Though maybe the issue is we specifically don't want it in `node`? I can't remember. I feel like I tried to do this before and it didn't work. If we did move it into `tools/testsetup.js` all `window` would have to be replaced with `self`.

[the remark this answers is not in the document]
leopsidom: Oh sure, I think we could move the logic in a separate js file. I suspect `node` should not use the `test.html` file, though need to double check. I can try it at some point.

src/templates/module_test.html
```diff
@@ -0,0 +1,26 @@
...
+    </script>
+    <script type="module">
+      import { loadPyodide } from "./pyodide.mjs";
+      window.loadPyodide = loadPyodidea;
```
ryanking13, line 22: I guess this must be a typo?

hoodmane: huh I wonder why that wouldn't break the tests

leopsidom: Hmm, that's interesting. Let me double check ~

leopsidom: Okay I think the tests pass because we already attach `loadPyodide` to [`globalThis` here](https://github.com/pyodide/pyodide/blob/main/src/js/pyodide.ts#L271), while this typoed line errors out due to undefined 
 `loadPyodidea`, so the line is not doing anything.

ryanking13: Thanks for the investigation. In that case we don't need to bind `loadPyodide` to the global in the test.html file anymore, so we would better remove those.

hoodmane: Well ideally we should eventually update the module build to avoid the global entrypoint. The `(globalThis as any).loadPyodide = loadPyodide;` is really only there for the non-ES6-module build. What we should do is have an extra file which just does
```js
import {loadPyodide} from "./pyodide.ts";
(globalThis as any).loadPyodide = loadPyodide;
```
and then build `pyodide.js` from this file and build `pyodide.mjs` from `pyodide.ts`.

leopsidom: So I'm guessing we may not necessarily need the `globalThis` line in `pyodide.ts`, since rollup will automatically attach all exports to a global variable specified in [the rollup config](https://github.com/pyodide/pyodide/blob/main/src/js/rollup.config.js#L13). I made a very simple example on umd vs esm build with rollup [here](https://stackblitz.com/edit/node-51syxy?file=bundle.js). In current implementation, we have `loadPyodide`  as a global variable. If we remove [this line](https://github.com/pyodide/pyodide/blob/main/src/js/pyodide.ts#L271). We should still be able to access the function in the browser via `loadPyodide.loadPyodide`, though it's probably better to rename the rollup config name as pyodide so we can access via `pyodide.loadPyodide`. But that also means we are changing the API..

leopsidom: @hoodmane Sorry  just saw your comments. Yeah I think even for non es6 build, it's ok to remove the line and let rollup handle the global variable binding for us. But that also means we are changing what pyodide exposes as global variable.

leopsidom: @ryanking13 Yeah, if we keep current implementation, then there's no need for that line in `test.html` file.

hoodmane: Yeah I don't want to make an API change like that. Currently we expose a function `loadPyodide` which returns `pyodide`. This makes sense to me. `pyodide.loadPyodide` or `loadPyodide.loadPyodide` are not so great. But I think we can easily just use a different entrypoint for the non-module version.

Incidentally, what I would really like is to bundle all of our top level code into a function called `loadPyodide` and expose that. We don't need any global variables currently, so there is nothing preventing loading two copies of Pyodide next to each other. But currently we use a lot of top level variables and I don't understand how to restructure the logic to avoid that while still allowing `import { func} from "./file"` and the doc generation stuff etc to work. I don't *think* rollup has "bundle everything into one big function" as an option.

leopsidom: I'm not sure what kind of top level variables / doc generation you are referring to. But to clarify, rollup bundle everything into an object and exposes that as a global variable, if you check the minimum example I linked above: https://stackblitz.com/edit/node-51syxy?file=bundle.js. which basically simplifies to something like the following:

```
(function (global, factory) {
  factory(global.pyodide = {});
}(this, (function (exports) { 'use strict';
  function loadPyodide() {
    console.log('loading pyodide');
  }
  exports.loadPyodide = loadPyodide;
  Object.defineProperty(exports, '__esModule', { value: true });
})));
```

hoodmane: > I'm not sure what kind of top level variables / doc generation you are referring to.

Well for instance in `module.ts` we define `export let Module: any = {};` which gets the emscripten module loaded into it. So you can't use `loadPyodide` twice because they would end up sharing the same emscripten module.

hoodmane: So ideally I would want rollup to create a (non es6 module) bundle that looks something like:
```js
function loadPyodide(config) {
  let module_exports;
  function init_module() {
    if (module_exports) {
      return;
    }
    module_exports = {};
    // top level contents of module.ts
    let Module = {};
    module_exports.Module = Module;
  }
  let load_pyodide_exports;
  function init_load_pyodide() {
    if (load_pyodide_exports) {
      return;
    }
    load_pyodide_exports = {};
    init_module();
    let Module = module_exports.Module;
    //... top level contents of load-pyodide.ts
  }
  // ...
  let pyodide_exports;
  function init_pyodide() {
    if (pyodide_exports) {
      return;
    }
    pyodide_exports = {};
    init_load_pyodide();
    let initializePackageIndex = load_pyodide_exports.initializePackageIndex;
    let loadPackage = load_pyodide_exports.loadPackage;
    // ... top level contents of pyodide.ts
    function loadPyodideInner(config) {
      // ...
    }
    pyodide_exports.loadPyodideInner = loadPyodideInner;
  }
  init_pyodide();
  pyodide_exports.loadPyodideInner(config);
}
window.loadPyodide = loadPyodide
```

hoodmane: For now I opened #2249 as a followup to this.

leopsidom: I'm not sure I understand. But isn't this what it's doing now ? The exports in `module.ts` are not exposed since it's not the entry point to rollup. Only exports in the entry point will be exposed publicly.

leopsidom: Oh I think I understand what you mean, when we call `loadPyodide` multiple times, we are overwriting the same Module instance.

hoodmane: Yeah exactly.

leopsidom: I personally don't have much use case to load multiple pyodide instances yet. But I think it could definitely be something useful. Maybe we could wrap modules and `loadPyodide` in a class `Pyodide` and expose that class. Then every time we instantiate a new instance of `Pyodide`, we create a  new standalone module that does not interfere with other instances.

hoodmane: Yeah maybe that would work.

hoodmane: See #2255.

PR #2230, opened 2022-02-28
chore: some package updates

henryiii:
I've updated the following packages:

- ~~chore: update lxml to 4.8.0~~
- chore: update MarkupSafe to 2.1.0
- chore: update astropy to 5.0.1
- chore: update logbook to 1.5.3
- chore: update regex to 2022.1.18
- chore: update sqlalchemy to 1.4.31
- chore: updated cffi to 1.15.0
- chore: updated imageio to 2.16.1
- chore: updated setuptools to 60.9.3
- chore: updated pytest to 7.0.1
- chore: updated typing-extensions to 4.1.1
- chore: updated pyrsistent to 0.18.1
- chore: updated threadpoolctl to 3.1.0

I wasn't sure how caching was done (if I update numpy, etc. all at once will
that blow up the CI build time?), and I'm hoping to avoid failures, so I left
the big or patched ones alone. Was after lxml (see #2225), but thought I'd get a couple other ones while at it. Removed it, having an issue so will be bumped separately.

* matplotlib 3.3.3 ❌ -> 3.5.1
* nltk 3.6.7 ❌ -> 3.7
* numpy 1.21.4 ❌ -> 1.22.2
* pandas 1.3.5 ❌ -> 1.4.1
* pillow 9.0.0 ❌ -> 9.0.1
* pyrsistent 0.18.0 ❌ -> 0.18.1
* python-sat 0.1.7.dev15 ❌ -> 0.1.7.dev16
* scikit-image 0.19.1 ❌ -> 0.19.2
* scipy 1.7.3 ❌ -> 1.8.0
* statsmodels 0.13.1 ❌ -> 0.13.2
* swiglpk 5.0.3 ❌ -> 5.0.4
* yt 3.6.1 ❌ -> 4.0.2
* zarr 2.10.3 ❌ -> 2.11.0

### DIFF
--- a/packages/MarkupSafe/meta.yaml
+++ b/packages/MarkupSafe/meta.yaml
@@ -1,9 +1,9 @@
 package:
   name: MarkupSafe
-  version: 2.0.1
+  version: 2.1.0
 source:
-  sha256: 594c67807fb16238b30c44bdf74f36c02cdf22d1c8cda91ef8a0ed8dabf5620a
-  url: https://files.pythonhosted.org/packages/bf/10/ff66fea6d1788c458663a84d88787bae15d45daa16f6b3ef33322a51fc7e/MarkupSafe-2.0.1.tar.gz
+  sha256: 80beaf63ddfbc64a0452b841d8036ca0611e049650e20afcb882f5d3c266d65f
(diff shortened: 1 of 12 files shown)


hoodmane:
Probably we should split off lxml. It looks like it has another issue with --skip-build or something.

hoodmane:
Also we probably need to flush the shell output in some way because the messages on the console don't come out in strictly chronological order, which is confusing when the error message appears before other stuff. Maybe the issue is that we should flush stdout before printing to street or something like that.

hoodmane:
Thanks @henryiii!

PR #2231, opened 2022-03-01
upgrade to jquery.terminal 2.32.0

eagleoflqj:
### Description

Upgrade to jquery.terminal 2.32.0 to fix https://github.com/jcubic/jquery.terminal/pull/741
`echo_command` should be replaced by `enter`, see https://github.com/jcubic/jquery.terminal/blob/master/CHANGELOG.md#2310


hoodmane:
Thanks for the upstream fix @eagleoflqj! Looks good, will merge when CI passes. For whatever reason circleci wasn't triggered but the github actions tests cover `test_console` which is good enough.

PR #2236, opened 2022-03-02
Handle strings and other weird things that are thrown as errors

hoodmane:
Split off from #2225. Unfortunately it is legal Javascript to throw weird stuff. Also unfortunately,
emscripten likes to throw weird stuff. Our `fatal_error` handling routine used
to incorrectly assume that things thrown were errors. This fixes it.


rth:
Could you add a test for this?

PR #2237, opened 2022-03-02
Add patches to make astropy work more fully

jobovy:
### Description

This PR updates the `astropy` package in `pyodide` to:
1. Apply patches to avoid crippling import errors due to the lack of `import _ssl` and full multiprocessing support in `pyodide`
2. Add a missing dependency `pyyaml` (see https://github.com/astropy/astropy/blob/v5.0.1/setup.cfg#L30)
3. Add import tests for all `astropy` sub-packages.

Without these packages, much of `astropy` doesn't work in `pyodide` even though most of `astropy` doesn't require SSL or multiprocessing (but these are imported in utility modules used throughout `astropy`). As you can see, the patches are very small and simple (just moving the imports into the functions that use them, causing the error to occur there).

With these patches, everything I've tried to run in `astropy` that I think should be able to run in a browser works. But I haven't tried running a test suite more systematically.

I will try to get these patches (or some version of them) accepted upstream so they would hopefully not be necessary for the next `astropy` release.

### Checklists

- [x] Add / update tests

### DIFF
--- a/packages/astropy/meta.yaml
+++ b/packages/astropy/meta.yaml
@@ -4,6 +4,9 @@ package:
 source:
   url: https://files.pythonhosted.org/packages/82/f0/fc3f6f250bb9d5e5aaa890469fb27c4edee66cf0e2ea4f8b0b287260c6f8/astropy-5.0.tar.gz
   sha256: 70203e151e13292586a817b4069ce1aad4643567aff38b1d191c173bc54f3927
+  patches:
+    - patches/0001-Move-ssl-import-into-functions-to-avoid-import-error.patch
+    - patches/0002-Move-concurrent.futures-import-into-function-to-avoi.patch
 build:
   script: |
     pip install extension-helpers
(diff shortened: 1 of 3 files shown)


jobovy:
Not sure why the benchmark test failed, I think it just ran out of time?

hoodmane:
Yeah nothing to do with your pr.

packages/astropy/meta.yaml
```diff
@@ -16,6 +19,26 @@ requirements:
     - packaging
     - numpy
     - pyerfa
+    - pyyaml
 test:
   imports:
     - astropy
+    - astropy.config
```
hoodmane, line 26: So I take it that with these extra imports the import test fails without your patches?

hoodmane (approved):
Looks good to me.

packages/astropy/meta.yaml
jobovy: I couldn't figure out how to run the tests locally (they don't seem to run when you build the package, which I was able to do), but at least some of them would fail without the patches, in particular `astropy.coordinates` (e.g., do `import astropy.coordinates` in the [current REPL](https://pyodide.org/en/stable/console.html)). I just added all sub-packages to the tests to make sure they all import. I'll add the config one now as well...

hoodmane:
Thanks @jobovy!

> I will try to get these patches (or some version of them) accepted upstream

Yes that's great please do this.

jobovy:
Thanks for reviewing and merging @hoodmane!

packages/astropy/meta.yaml
hoodmane: Incidentally, the way you would run these tests is:
```sh
pytest -k "chrome and astropy" packages/test_packages_common.py
```

jobovy:
This patch has now also been merged upstream (https://github.com/astropy/astropy/pull/12911) and will be backported to the current release series (https://github.com/astropy/astropy/pull/12920), so these patches should be able to be dropped when switching to the next release of `astropy`.

pllim:
Yes, please be on lookout for astropy 5.0.2. Thanks, everyone!

PR #2238, opened 2022-03-02
Better package cross compilation

hoodmane:
Instead of trying to convince setup.py to do what we want, the wheel
package empowers us to unpack and directly manipulate the wheel.
In particular this is important because `--skip-build` really does
not work.

First we build a native wheel. If skip_host is True, then this
"native" wheel has defective empty files as binaries, if it's False,
we build a real native wheel. If skip_host is False, then we use
pip to install our native wheel into host directory.

Then we unpack the wheel, replace the platform tag with
wasm32_emscripten, remove the non-metadata files, and copy in the
new files with the cross compiled binaries.

Of course, this is all only necessary for binary extensions.

Hopefully this will help get us towards using alternative build
backends. The one concern is that this approach depends on
implementation details of where the build backend puts the files
into the directory tree before copying them into the wheel. If
the build backend moved the linked binaries after building them,
we would be in trouble. Or worse if the build backend deleted the
files after packing them into the wheel, we'd be in trouble.

Basically, it seems clear that we can only support alternative
build backends on a case-by-case basis.

@henryiii I'm curious what your thoughts are cross compilation + arbitrary build backends.


henryiii:
I'd have to understand the cross-compile process a little better. Where is the "cross-compiled" run being done? I'm guessing it's picking up the compile commands, then reprocessing them, or something like that?

hoodmane:
>  I'm guessing it's picking up the compile commands, then reprocessing them, or something like that?

Yes. On this branch, we run `python setup.py bdist_wheel` with symlinks installed so that attempts to invoke the C compiler toolchain are recorded into `$pkg_root/build/$src_dir/build.log`. If `skip_host` is not false, we just emit an empty output file rather than actually invoking the compiler. Then we delete these "native" artifacts.
This is done by the `capture_compile` command in `pywasmcross`. `capture_make_command_wrapper_symlinks` places the symlinks into the path.
 
Then we replay the compilation with `replay_compile` which iterates over `build.log` and calls `replay_command`, which does some rewrite process to remove flags that don't make sense or replace them with other flags, rearrange which libraries are loaded, remove or reroute system libraries or includes into emscripten ones etc. (This is quite similar to what happens in `emcc.py` which is after a lot of the same stuff.)

Then we unpack the wheel, copy the new binary files into the wheel, rewrite the platform tag, and repack.

pyodide-build/pyodide_build/buildpkg.py
```diff
@@ -451,8 +438,34 @@ def compile(
...
+
+def update_wheel_platform(build_dir: Path, wheel_dir: Path):
+    import distutils.dir_util
+
+    # for some reason the .so name and the wheel name swap order of architecture
+    # and OS?
+    set_host_platform(wheel_dir, "emscripten_wasm32")
+    pywasmcross.clean_out_native_artifacts(wheel_dir)
+    lib_dir = next(build_dir.glob("lib*"))
+    distutils.dir_util.copy_tree(str(lib_dir), str(wheel_dir))
```
ryanking13, line 463: Maybe use `shutil.copytree` instead? distutils is deprecated.

hoodmane: They don't do the same thing and we need exactly the distutils behavior. We are trying to copy what setuptools does, which is they use this function. But we can remove it with #2256.

hoodmane: If you look at the packages CI for [71a07ce](https://github.com/pyodide/pyodide/pull/2238/commits/71a07ce02a55e7c2df8ffb0dcd14fd4039fa0ad7) you can see the problems we get with `shutil.copytree`.

ryanking13: Thanks for the explanation!

ryanking13 (approved):
Thanks, @hoodmane! Generally LGTM.

It would be great if rth can review this too, as I haven't reviewed our build system very much. But he is doing more important things so I'm fine with merging this as there are other PRs waiting for this. (We can discuss our build system more in #2256 anyway)

hoodmane:

> I haven't reviewed our build system very much

It's very confusing and we have a steep learning curve, but thankfully we seem to be gradually increasing how much we understand the build system. Probably it would be good to devote some time to improving the dev docs or in line comments on this topic to reduce the confusingness.

> I'm fine with merging this as there are other PRs waiting for this

To be clear, my opinion is that this replaces the hack we already have with a new hack which is marginally simpler and more predictable. But this is enough to allow us to update to Python 3.10 =)

#2256 will dramatically improve predictability and get us away from being tied to setuptools. It also removes slightly more than 400 loc on net.

hoodmane:
Okay I guess I'll merge this then.

pyodide-build/pyodide_build/buildpkg.py
```diff
@@ -451,8 +438,34 @@ def compile(
...
+    lines = WHEEL.read_text().splitlines()
+    result_lines = []
+    for line in lines:
+        if line.startswith("Tag:"):
+            line = line.rpartition("-")[0] + "-" + platform
+        result_lines.append(line)
+    WHEEL.write_text("\n".join(result_lines))
+
+
+def update_wheel_platform(build_dir: Path, wheel_dir: Path):
```
rth, line 455: Would you mind adding a docstring explaing what exactly we are updating and where?

pyodide-build/pyodide_build/buildpkg.py
```diff
@@ -451,8 +438,34 @@ def compile(
                 replace_libs=replace_libs,
                 replay_from=replay_from,
             )
-        install_for_distribution()
-    del bash_runner.env["_PYTHON_HOST_PLATFORM"]
+
+
+def set_host_platform(wheel_dir: Path, platform: str):
```
rth, line 443: Same, a docstring with a sentence saying that this edits *.dist-info/WHEEL to replace the Tag value with the platform, and linking to the PEP or documentation listing valid values there wouldn't hurt.

pyodide-build/pyodide_build/pywasmcross.py
```diff
@@ -144,21 +144,31 @@ def capture_compile(*, host_install_dir: str, skip_host: bool, env: dict[str, st
...
+        subprocess.check_call([sys.executable, "setup.py", "bdist_wheel"], env=env)
+        if not skip_host:
+            assert host_install_dir, "Missing host_install_dir"
+            result_wheel = str(list(Path("dist").glob("*.whl"))[0])
+            subprocess.check_call(
+                [
+                    sys.executable,
+                    "-m",
+                    "pip",
+                    "install",
```
rth, line 160: The only thing is pip will often use build isolation to install build requirements from `pyproject.toml` and not whatever dependencies we indicated in our build system.
In most cases, that's probably fine, but I wonder if this won't be an issue if the build dependency is patched. Say you build a package that needs numpy at build time, and it will use a different (and unpatched) numpy version to install it on the host when skip_host=False. 
Again I'm not sure if it's really an issue, but we could at least mention it in a comment that in some cases it might be appropriate to disable build isolation here.

rth (commented):
Thanks @hoodmane !

I agree it's a good idea to get rid of `setup.py install` there, and the general direction sounds good to me. This does depend on some of the implementation details, as you mentionned, but we have a wide enough variety of packages to catch most common issues I think. So as long as it builds it should be OK.

A few minor comments below, mostly about docs (feel fee to aggregate them into a single PR, I'm going to have a look at some other recent PRs as well).

pyodide-build/pyodide_build/buildpkg.py
hoodmane: I think this function was removed in #2256.

pyodide-build/pyodide_build/buildpkg.py
hoodmane: Same, removed in #2256

pyodide-build/pyodide_build/pywasmcross.py
hoodmane: Oh yeah, this was also removed in #2256 (along with the entire `skip_host` key). Only `numpy` was setting `skip_host` to false too, so if there were problems on this code path we probably wouldn't know.

PR #2240, opened 2022-03-02
Add cache for lint CI to reduce timeouts

hoodmane:


.circleci/config.yml
```diff
@@ -31,6 +34,11 @@ jobs:
           name: lint
           command: make lint
 
+      - save_cache:
+          key: pre-commit-{{ checksum ".pre-commit-config.yaml" }}
+          paths:
+            - "/root/.cache"
```
ryanking13, line 40: Maybe,

```suggestion
            - "/root/.cache/pre-commit"
```

as there are other programs that use `.cache` for a cache directory (pip, playwright, firefox, ...).

hoodmane: I want to cache those things too (currently pip and black).

hoodmane:
@henryiii I think another thing to do would be to split prettier into multiple steps, maybe docs, packages, src, and everything else. Since it seems to timeout on prettier a lot, and if there is any output to stdout the timeout is reset.

henryiii:
Yeah, but given that pre-commit takes 14.8 seconds to run _everything_ on my machine, of which 2.5 seconds is prettier, I don't think the timeout has anything to do with a slow check. It's either an issue downloading something, or it's an issue with CircleCI. The former should be mostly fixed by caching.

hoodmane:
Well I've seen it timeout at both the download and the prettier stage. This should at least prevent download timeouts.

PR #2241, opened 2022-03-02
MAINT use variables for Python version rather than hardcoding it

hoodmane:
This should make Python updates easier.

### DIFF
--- a/pyodide-build/pyodide_build/testing.py
+++ b/pyodide-build/pyodide_build/testing.py
@@ -1,12 +1,15 @@
 import contextlib
 import inspect
 from base64 import b64encode
-from sys import version_info
 from typing import Callable, Collection, Optional
 
 import pytest
 
-PYVERSION = f"python{version_info.major}.{version_info.minor}"
+from .common import get_make_flag
+
+PYMAJOR = get_make_flag("PYMAJOR")
+PYMINOR = get_make_flag("PYMINOR")
+PYVERSION = f"python{PYMAJOR}.{PYMINOR}"
 
 
 def _run_in_pyodide_get_source(f):
--- a/pyodide-build/pyodide_build/testing.py
+++ b/pyodide-build/pyodide_build/testing.py
@@ -1,9 +1,12 @@
 import contextlib
 import inspect
 from base64 import b64encode
+from sys import version_info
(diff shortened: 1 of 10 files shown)


pyodide-build/pyodide_build/testing.py
```diff
@@ -1,10 +1,13 @@
 import contextlib
 import inspect
 from base64 import b64encode
+from sys import version_info
 from typing import Callable, Collection, Optional
 
 import pytest
 
+PYVERSION = f"python{version_info.major}.{version_info.minor}"
```
ryanking13, line 9: I think tests should use Pyodide Python version not local Python version, though currently our assumption is that those two are strictly same.

hoodmane: Yeah. This is easier to write though. What should I do to get the Pyodide version here?

ryanking13: Well I think tests that require knowing python version can just do `version = pyodide.runPython('from sys import version_info, ...')` each time like you did at the `pyodide.ts`.
If there are too many of those, maybe we could expose a version variable to js such as `pyodide.pythonVersion` (I'm not sure though)

hoodmane: We could just add it as a attribute to selenium e.g., `selenium.pyversion`

hoodmane: Okay, how about this?

ryanking13: Yeah, this looks better. In the long term, I think we should split out testing-related things from pyodide-build, but that could be discussed in other place.

ryanking13 (approved):
Thanks, @hoodmane. This would indeed make updating Python version easier.

Do you have plans to also update hardcoded Python version strings in packages directory in this PR? (e.g. [scipy](https://github.com/pyodide/pyodide/blob/a725c6ce0fbd145c5ebc66f50710197dbb0f216d/pac kages/scipy/meta.yaml#L92)).

hoodmane:
I currently have that on #2238.

hoodmane:
So I'm good with merging this as is and dealing with more version numbers soon in other prs.

PR #2234, opened 2022-03-01
chore: matplotlib 3.5.1, yt 4

henryiii:
### Description

Bumping MPL version - it turns out that Hist (which uses boost-histogram) depends on mplhep, which in turn requires matplotlib 3.4+. So I migrated the patches and bumped MPL.

I noticed micropip doesn't seem to respect extras, that is, `await micropip.install("hist[plot]")` doesn't seem to work. (will investigate later)

* Updated matplotlib to 3.5
    * Added fonttools
* Updated yt to 4
    * Added `cmyt` and `unyt`, which I am pretty sure are yt specific. At least by looking at the name. 😆 
    * Also `colorspacious`
    * Also `tdqm`
    * Also `tomli` and `tomli-w` (backported from the next yt release to avoid adding a dead dependency `toml`)

Closes #2088 too. Also 3.5 was the first version to officially support 3.10, I think.

PR #2242, opened 2022-03-02
fix: better error messages, fewer tracebacks

henryiii:
This reduces the text output when an error is hit, removing the tracebacks for known errors. Tracebacks are for unknown/unexpected errors - if you expect an error, it's better to print out an actual error message. I've also made the patch applying a bit more verbose in case it helps, as it sometimes does.

Example after the changes:

```
Building the following packages: yt
[1/1] (thread 1) building yt
Error building yt. Printing build logs.
Hmm...  Looks like a unified diff to me...
--------------------------
|diff --git a/setupext.py b/setupext.py
|index 92c456bab..ce2c5fbad 100644
|--- a/setupext.py
|+++ b/setupext.py
--------------------------
patching file setupext.py
Using Plan A...
Hunk #1 FAILED at 57.
1 out of 1 hunk FAILED -- saving rejects to file setupext.py.rej
done
[2022-03-02 17:35:58] Building package yt...
ERROR: Patch /src/packages/yt/patches/skip-openmp.patch failed
[2022-03-02 17:36:14] Succeeded building package yt in 16.5 seconds.
ERROR: cancelling buildall
[1/1] (thread 1) failed yt in 16.76 s
make[1]: *** [Makefile:12: all] Error 1
make[1]: Leaving directory '/src/packages'
make: *** [Makefile:187: build/packages.json] Error 2
```

### DIFF
--- a/pyodide-build/pyodide_build/buildpkg.py
+++ b/pyodide-build/pyodide_build/buildpkg.py
@@ -297,9 +297,13 @@ def patch(pkg_root: Path, srcpath: Path, src_metadata: dict[str, Any]):
     # Apply all the patches
     with chdir(srcpath):
         for patch in patches:
-            subprocess.run(
-                ["patch", "-p1", "--binary", "-i", pkg_root / patch], check=True
+            result = subprocess.run(
+                ["patch", "-p1", "--binary", "--verbose", "-i", pkg_root / patch],
+                check=False,
             )
+            if result.returncode != 0:
+                print(f"ERROR: Patch {pkg_root/patch} failed")
+                raise SystemExit(result.returncode)
 
     # Add any extra files
     for src, dst in extras:
--- a/pyodide-build/pyodide_build/buildpkg.py
+++ b/pyodide-build/pyodide_build/buildpkg.py
@@ -14,11 +14,13 @@ import re
 import shutil
 import subprocess
 import sys
+import textwrap
 from contextlib import contextmanager
 from datetime import datetime
 from pathlib import Path
 from textwrap import dedent
-from typing import Any, Optional
+from types import TracebackType
+from typing import Any, NoReturn, Optional, TextIO
 from urllib import request
 
 from . import pywasmcross
@@ -48,6 +50,16 @@ shutil.register_unpack_format(
 )
 
 
+def exit_with_stdio(result: subprocess.CompletedProcess) -> NoReturn:
+    if result.stdout:
+        print("  stdout:")
+        print(textwrap.indent(result.stdout, "    "))
+    if result.stderr:
+        print("  stderr:")
+        print(textwrap.indent(result.stderr, "    "))
+    raise SystemExit(result.returncode)
+
+
 class BashRunnerWithSharedEnvironment:
     """Run multiple bash scripts with persistent environment.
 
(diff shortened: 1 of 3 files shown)


hoodmane:
Awesome, thanks. Those cascading subprocess tracebacks create a lot of noise in the build logs, this will probably improve quality of life when debugging build system quite significantly.

hoodmane:
Would you mind making similar changes to the other subprocess calls in pyodide-build?

henryiii:
> improve quality of life when debugging build system

It was developed due to debugging #2234 :)

pyodide-build/pyodide_build/buildpkg.py
```diff
@@ -450,7 +470,10 @@ def package_wheel(
     post = build_metadata.get("post")
     if post:
         bash_runner.env.update({"PKGDIR": str(pkg_root)})
-        bash_runner.run(post, check=True)
+        result = bash_runner.run(post)
+        if result.returncode != 0:
+            print("ERROR: post ({post}) failed")
```
hoodmane, line 475: Is this supposed to be a format string here? Also, post is generally a multiline script so I don't know whether this would look good.

pyodide-build/pyodide_build/buildpkg.py
```diff
@@ -49,7 +49,7 @@ def _make_whlfile(*args, owner=None, group=None, **kwargs):
 
 
 class BashRunnerWithSharedEnvironment:
-    """Run multiple bash scripts with persisent environment.
+    """Run multiple bash scripts with persistent environment.
```
hoodmane, line 52: It would be nice if the bashrunner would print the script it tried to run if the script failed. Currently it appears with a traceback with the bash script collapsed like `cmd1\ncmd2\ncmd3` which is pretty hard to read.

pyodide-build/pyodide_build/buildpkg.py
henryiii: Oops, was removing those, missed that one.

pyodide-build/pyodide_build/buildpkg.py
henryiii: Okay, will look at this!

hoodmane: Not necessary in this PR if you don't feel like it, but it would be helpful.

hoodmane: We could also really use to fix the bashrunner so that it returns an error if the script has an error rather than hanging or returning 0.

henryiii: This also needs some rewriting already (see #2194), maybe could do both.

pyodide-build/pyodide_build/buildpkg.py
```diff
@@ -75,20 +93,32 @@ def run(self, cmd, **opts):
...
-        cmd += "\n" + write_env_shell_cmd
+        full_cmd = f"{cmd}\n{write_env_shell_cmd}"
         result = subprocess.run(
-            ["bash", "-ce", cmd], pass_fds=[self._fd_write], env=self.env, **opts
+            ["bash", "-ce", full_cmd], pass_fds=[self._fd_write], env=self.env, **opts
         )
+        if result.returncode != 0:
+            print("ERROR: bash commmand failed")
+            print(textwrap.indent(cmd, "    "))
+            exit_with_stdio(result)
```
hoodmane, line 103: Yes this looks much better.

henryiii:
Okay, this should avoid a hang if the command fails, as it no longer tries to read a line from a file descriptor that might be empty, but instead immediately fails (and still properly closes the file descriptors). I'll rebase #2194 on this after it goes in, this should resolve the unclosed file descriptor warning uncovered there too.

hoodmane (approved):

hoodmane:
Thanks @henryiii this is a nice improvement.

PR #2245, opened 2022-03-02
chore: use GHA for linting

henryiii:
This might reduce flakiness.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -17,7 +17,6 @@ jobs:
       - uses: pre-commit/action@v2.0.3
 
   build-core:
-    needs: [lint]
     runs-on: ubuntu-latest
     env:
       EMSDK_NUM_CORES: 3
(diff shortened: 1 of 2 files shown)


.github/workflows/main.yml
```diff
@@ -7,7 +7,17 @@ on:
...
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/setup-python@v3
+        with:
+          python-version: "3.9"
+      - uses: pre-commit/action@v2.0.3
+
   build-core:
+    needs: [lint]
```
ryanking13, line 20: I think build (and test) should run even if lint fails.

ryanking13 (commented):
Thanks @henryiii, I hope pre-commit won't hang on GHA 😂

.github/workflows/main.yml
henryiii: Okay.

I've never had it hang on GHA, but I don't think I've used prettier, either. I don't see why it would be a problem, though.

hoodmane: I think run tests on lint failure is a critical quality of life feature. But I don't really look at the gha tests so it doesn't matter as much.

ryanking13 (approved):

PR #2257, opened 2022-03-07
chore: more pre-commit checking

henryiii:
- chore: clean up flake8 and mypy
- chore: add spellcheck
- chore: add shellcheck

This cleans up the flake8 & mypy checks, adds a bit of spell checking, and a linter for shell scripts.

https://github.com/python/typeshed/pull/7453 will help, along with https://github.com/python/typeshed/issues/7452 (already fixed in typeshed).

### DIFF
--- a/docs/development/core.md
+++ b/docs/development/core.md
@@ -23,7 +23,7 @@ The primary purpose of `core` is to implement {ref}`type translations <type-tran
 
 ### Type conversion from Python to JavaScript
 
-- `python2js` -- Translates types from Python to JavaScript, implicitly converting basic types and creating pyproxies for others. It also implements explicitly conversion from Python to JavaScript (the `toJs` method).
+- `python2js` -- Translates types from Python to JavaScript, implicitly converting basic types and creating pyproxies for others. It also implements explicit conversion from Python to JavaScript (the `toJs` method).
 - `python2js_buffer` -- Attempts to convert Python objects that implement the Python [Buffer Protocol](https://docs.python.org/3/c-api/buffer.html). This includes `bytes` objects, `memoryview`s, `array.array` and a wide variety of types exposed by extension modules like `numpy`. If the data is a 1d array in a contiguous block it can be sliced directly out of the wasm heap to produce a JavaScript `TypedArray`, but JavaScript does not have native support for pointers, so higher dimensional arrays are more complicated.
 - `pyproxy` -- Defines a JavaScript `Proxy` object that passes calls through to a Python object. Another important core file, `PyProxy.apply` is the primary entrypoint into Python code. `pyproxy.c` is much simpler than `jsproxy.c` though.
 
--- a/docs/development/core.md
+++ b/docs/development/core.md
@@ -23,7 +23,7 @@ The primary purpose of `core` is to implement {ref}`type translations <type-tran
 
 ### Type conversion from Python to JavaScript
 
-- `python2js` -- Translates types from Python to JavaScript, implicitly converting basic types and creating pyproxies for others. It also implements explicity conversion from Python to JavaScript (the `toJs` method).
+- `python2js` -- Translates types from Python to JavaScript, implicitly converting basic types and creating pyproxies for others. It also implements explicitly conversion from Python to JavaScript (the `toJs` method).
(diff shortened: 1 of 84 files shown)


docs/development/core.md
```diff
@@ -23,7 +23,7 @@ The primary purpose of `core` is to implement {ref}`type translations <type-tran
 
 ### Type conversion from Python to JavaScript
 
-- `python2js` -- Translates types from Python to JavaScript, implicitly converting basic types and creating pyproxies for others. It also implements explicity conversion from Python to JavaScript (the `toJs` method).
+- `python2js` -- Translates types from Python to JavaScript, implicitly converting basic types and creating pyproxies for others. It also implements explicitly conversion from Python to JavaScript (the `toJs` method).
```
hoodmane, line 26:
```suggestion
- `python2js` -- Translates types from Python to JavaScript, implicitly converting basic types and creating pyproxies for others. It also implements explicit conversion from Python to JavaScript (the `toJs` method).
```

src/py/_pyodide/console.py
```diff
@@ -362,10 +372,7 @@ def formatsyntaxerror(self, e: Exception) -> str:
         sys.last_type = type(e)
         sys.last_value = e
         sys.last_traceback = None
-        try:
-            return "".join(traceback.format_exception_only(type(e), e))
-        finally:
-            e = None  # type: ignore
```
hoodmane, line 368: Ah I guess this isn't needed since e wasn't thrown down the stack from this function, so e's traceback won't contain a reference to this scope hence no reference loop to break.

tools/buildf2c
```diff
@@ -70,11 +70,7 @@ tar -xvf f2c.tar
 
 gunzip -rf f2c/*
 
-cd f2c
```
hoodmane, line 73: I'm not convinced we should have this file in here at all. Again, it looks like it's specifically for OSX. Is there really no better way to install it there than using this script? I guess having an untested and possibly slightly broken script is better than none?

hoodmane (approved):
Thanks @henryiii nice to see all these typos fixes and other beautiful maintenance work.

hoodmane:
We really need to get rid of the line number in `test_console_html` it's not acceptable for that test to break every time `_pyodide/console.py` is modified.

tools/buildf2c
henryiii: Maybe it could be added to homebrew? I see it in macPorts.

hoodmane: Well I certainly won't do that, but it would be great if someone else would. In the meantime leaving this unmaintained script here is less work than fixing the problem.

src/py/_pyodide/console.py
henryiii: If it does cause a problem, I think you want `del e` rather than `e = None`.

src/tests/test_console.py
```diff
@@ -404,33 +404,41 @@ async def f():
...
+            f"""\
             >>> class Test:
             ...     def __repr__(self):
             ...         raise TypeError(\"hi\")
             ... \
 
             >>> Test()
             [[;;;terminal-error]Traceback (most recent call last):
-              File \"/lib/{PYVERSION}/site-packages/_pyodide/console.py\", line 465, in repr_shorten
+              File \"/lib/{PYVERSION}/site-packages/_pyodide/console.py\", line """
```
hoodmane, line 429: I was thinking of something like `result = re.sub("line [0-9]*", "line xxx", result)` but I suppose this gets the job done too.

hoodmane: I think doing the regex substitution is preferable because `assert x == y` gives better pytest error messages than `assert x.endswith(y)`.

tools/buildf2c
henryiii: I'd do it if it wasn't so annoying. No versioned link to the files AFAICT, and brew's not going to like that. Barely even versions (dates, I think).

src/py/_pyodide/console.py
hoodmane: I copied this from the Python source code. Which doesn't mean that it is the right way to do things of course. It seems that in the CPython source code for v3.10.2 there are 32 instances of `finally: some_var = None` and 36 instances of `finally: del some_var`.

hoodmane: Thanks for the explanation!

PR #2258, opened 2022-03-07
Fix a source of buildall flakiness

hoodmane:
This spot causes CI failures a lot, I'm not sure why. Catch errors with log file manipulation and throw them away.

### DIFF
--- a/pyodide-build/pyodide_build/buildall.py
+++ b/pyodide-build/pyodide_build/buildall.py
@@ -157,17 +157,27 @@ class Package(BasePackage):
             except StopIteration:
                 rebuilt = False
 
-        if rebuilt:
-            shutil.move(self.pkgdir / "build.log.tmp", self.pkgdir / "build.log")  # type: ignore
-            if args.log_dir and (self.pkgdir / "build.log").exists():
-                shutil.copy(
-                    self.pkgdir / "build.log", Path(args.log_dir) / f"{self.name}.log"
-                )
-        else:
-            (self.pkgdir / "build.log.tmp").unlink()
+        log_trouble = False
+        try:
+            if rebuilt:
+                shutil.move(self.pkgdir / "build.log.tmp", self.pkgdir / "build.log")  # type: ignore
+                if args.log_dir and (self.pkgdir / "build.log").exists():
+                    shutil.copy(
+                        self.pkgdir / "build.log",
+                        Path(args.log_dir) / f"{self.name}.log",
+                    )
+            else:
+                (self.pkgdir / "build.log.tmp").unlink()
+        except FileNotFoundError:
+            # This is a common source of flakiness in CI for some reason.
+            log_trouble = True
 
         if p.returncode != 0:
             print(f"Error building {self.name}. Printing build logs.")
+            if log_trouble:
+                print(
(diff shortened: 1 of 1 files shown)


henryiii:
Do you have a link to some logs with this failure?

hoodmane:
Most recently on #2257:
https://app.circleci.com/pipelines/github/pyodide/pyodide/3645/workflows/1b22d19f-eb5c-468b-8c57-71a906ec4092/jobs/39711

But I've seen this sporadically a lot of times. If you have any idea what's actually going on, it would be good to have a more principled fix, but I think this will help.

ryanking13:
Yes, we are facing this error periodically, thanks for investigating this @hoodmane.
Actually, I think the reason of this is because we are changing cwd during build process using `os.chdir` which affects all threads.
(If thread 1 changed cwd, and simultaniously thread 2 writes to build-logs dir, it crashes)

Which means the simple solution would be just using absolute path for `log_dir`.

https://github.com/pyodide/pyodide/blob/8323987b399cd0bd8e91393377f1cc5f04519a81/packages/Makefile#L14

ryanking13:
> _Edit:_ On the second thought, changing cwd in one process should not affect others because we are running each package build in a subprocess. Anyway, I still think that setting an absolute path for `log_dir` would likely fix this.

Here is a new suspect. `shutil.make_archive`, which is called in the main process, changes cwd internally and [it's not thread-safe.](https://docs.python.org/3/library/shutil.html#shutil.make_archive)

https://github.com/pyodide/pyodide/blob/4f8b0a03447531a416944ee24beee63f6a8ebc4f/pyodide-build/pyodide_build/buildall.py#L180-L183

ryanking13:
> Anyway, before that, you could lock around make_archive & writing the logs?

Yeah, or we could find a thread-safe version of make_archive.

hoodmane:
Well but the problem is actually that it's completely wrong to change the working directory in one thread we want to change it in all threads. But aren't we inside a subprocess in buildpkg? Or is there a cwd in buildall?

ryanking13:
> Well but the problem is actually that it's completely wrong to change the working directory in one thread we want to change it in all threads. But aren't we inside a subprocess in buildpkg? Or is there a cwd in buildall?

yes, `shutil.make_archive` is used in buildall.

hoodmane:
Maybe for now move this stuff into buildpkg? It is useful to use buildpkg on it's own and kind of a nuisance that the log isn't stored and the dist directory isn't copied.

ryanking13:
> Maybe for now move this stuff into buildpkg? It is useful to use buildpkg on it's own and kind of a nuisance that the log isn't stored and the dist directory isn't copied.

IMO, we can move the logic of generating a zip file of the shared library to `buildpkg` and leave other things as is.

`buildpkg` should be responsible for generating build output of each package (wheel files, static and shared library outputs), while moving each package logs and build outputs into a central place can be done by `buildall`.

henryiii:
I think we should start by adding the lock, it's easy and should fix the problem for now. Then this might get rewritten anyway. :)

hoodmane:
@henryiii could you add the lock? I will close this.

henryiii:
Sure, about to do it.

hoodmane:
Thanks!

PR #2260, opened 2022-03-08
Fix word splitting on run_docker script

ryanking13:
#2257 added a double quote on every bash variable, but there was an error because we have some intended word-splitting stuff.


run_docker
```diff
@@ -119,9 +119,10 @@ mkdir -p .docker_home
 
 # Start a detached container as root, add the host uname and uid to /etc/passwd,
 # then run forever
+# shellcheck disable=2086 # Intended splitting
 CONTAINER=$(\
   docker run \
-    "$PORT_CONFIGURATION_LINE" \
+    $PORT_CONFIGURATION_LINE \
```
henryiii, line 125: You can also make it explicit by using an array, then you don't need to disable the shellcheck check.

henryiii (approved):

run_docker
ryanking13: Yes, that could be better for consistency. Thanks for the suggestion.

PR #2262, opened 2022-03-09
fix(build): lock around logging and archiving portions

henryiii:
### Description

Retry of #2258. Hoping to avoid flakiness by locking around log manipulation code.

### DIFF
--- a/pyodide-build/pyodide_build/buildall.py
+++ b/pyodide-build/pyodide_build/buildall.py
@@ -15,7 +15,7 @@ from pathlib import Path
 from queue import PriorityQueue, Queue
 from threading import Lock, Thread
 from time import perf_counter, sleep
-from typing import Any, Optional
+from typing import Any, ClassVar, Optional
 
 from . import common
 from .buildpkg import needs_rebuild
@@ -79,6 +79,8 @@ class StdLibPackage(BasePackage):
 
 @total_ordering
 class Package(BasePackage):
+    lock: ClassVar[Lock] = Lock()
+
     def __init__(self, pkgdir: Path):
         self.pkgdir = pkgdir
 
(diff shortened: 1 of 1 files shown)


pyodide-build/pyodide_build/buildall.py
```diff
@@ -148,28 +150,30 @@ def build(self, outputdir: Path, args) -> None:
...
-
-        if rebuilt:
-            shutil.move(self.pkgdir / "build.log.tmp", self.pkgdir / "build.log")
-            if args.log_dir and (self.pkgdir / "build.log").exists():
-                shutil.copy(
-                    self.pkgdir / "build.log", Path(args.log_dir) / f"{self.name}.log"
-                )
-        else:
-            (self.pkgdir / "build.log.tmp").unlink()
+        with self.__class__.lock:
```
henryiii, line 153: This could be written less explicitly as `self.lock` instead.

pyodide-build/pyodide_build/buildall.py
```diff
@@ -178,10 +182,11 @@ def build(self, outputdir: Path, args) -> None:
         if self.library:
             return
         if self.shared_library:
-            file_path = shutil.make_archive(
-                f"{self.name}-{self.version}", "zip", self.pkgdir / "dist"
-            )
-            shutil.copy(file_path, outputdir)
+            with self.__class__.lock:
```
henryiii, line 185: I wonder if this is the only one needed - the others really should be in separate directories and should be thread safe, I think.

hoodmane:
So if I understand correctly, the problem is that `shutil` commands use `os.chdir` which changes process-level state and this leads to race conditions when using threads?

ryanking13:
> So if I understand correctly, the problem is that `shutil` commands use `os.chdir` which changes process-level state and this leads to race conditions when using threads?

Yes, thats what I think for the reason of the flakiness.

pyodide-build/pyodide_build/buildall.py
ryanking13: If we use absolute path for args.log_dir (so that `Path(args.log_dir)` doesn't change even when cwd changed), then I think this would be the only lock needed.

henryiii: Okay, let's start with that.

henryiii: It looks like make_archive is pretty simple and pretty easy to avoid: https://stackoverflow.com/questions/1855095/how-to-create-a-zip-archive-of-a-directory/1855118#1855118 if we wanted to go the direction of avoiding cwd changes in the first place.

ryanking13: Thanks for the investigation! I think the two solutions are all temporary and the long term solution would be moving some stuff in `buildall` to `buildpkg`, so for now let's go with this locking-based one and revisit this later.

ryanking13 (approved):
Thanks, @henryiii!

pyodide-build/pyodide_build/buildall.py
henryiii: Yes, I think #2256 is the next priority. After that, I'll probably look into some refactoring of buildall/buildpkg related to #2246 - if this was async, it wouldn't matter, and if it was moved to buildpkg, it also wouldn't matter. I'm also interested in working on supporting CMake packages, but that also requires #2256 (or at least I'd rather learn the system after the rewrite and not before).

ryanking13: > Yes, I think #2256 is the next priority.

Yes, #2256 will be a big change in our build system, and I will be glad if you can help us review it.

PR #2265, opened 2022-03-10
Rework core tests

hoodmane:
This is on top of #2225. The goal is to be able to skip specific tests. Many core test sets have just one test that uses subprocess or threads. Others have a few tests that are failing due to upstream problems in emscripten. I went through and switched to skipping specific tests in as many cases as I could. In modules where most or all tests use threading or subprocess, I gave up and put a comment saying something like 35/50 tests fork and I didn't want to bother to salvage the remaining ones.

In order to accommodate using patterns to skip specific tests (and also extra comments) I switched the format of `python_tests` to a yaml file. This gives us significantly more flexibility in the structure of the file and allows us to use `ruamel.yaml` to update it. We gain a lot flexibility while reducing the length of `make_test_lists.py` modestly from 108 to 78 loc.

In the future, hopefully we should be able to make fixes by deleting one skip pattern and looking into why that particular test fails.


emsdk/patches/0007-Fix-lookupPath-when-applied-to-a-symlink-loop.patch
```diff
@@ -0,0 +1,31 @@
...
+diff --git a/src/library_fs.js b/src/library_fs.js
+index 519fab972..ef91655f2 100644
+--- a/src/library_fs.js
++++ b/src/library_fs.js
+@@ -169,7 +169,7 @@ FS.staticInit();` +
+             var link = FS.readlink(current_path);
+             current_path = PATH_FS.resolve(PATH.dirname(current_path), link);
+ 
+-            var lookup = FS.lookupPath(current_path, { recurse_count: opts.recurse_count });
++            var lookup = FS.lookupPath(current_path, { recurse_count: opts.recurse_count + 1 });
```
rth, line 25: Is there an upstream issue?

src/tests/python_tests.yaml
```diff
@@ -0,0 +1,845 @@
...
+- test_binascii
+- test_binhex
+- test_binop
+- test_bisect
+- test_bool
+- test_buffer
+- test_bufio
+- test_builtin:
+    skip:
+      - test_open_non_inheritable
```
rth, line 129: Cool. The extra granularity is certainly very welcome.

src/tests/test_core_python.py
```diff
@@ -1,70 +1,99 @@
...
+    ignore_tests = info.get("skip", [])
+    if not isinstance(ignore_tests, list):
+        raise Exception("Invalid python_tests.yaml entry: 'skip' should be a list")
 
     selenium.load_package(list(UNVENDORED_STDLIB_MODULES))
     try:
         selenium.run(
+            f"""
+import platform
+from test import libregrtest
```
rth, line 59: maybe keep the original indentation and add a textwrap.dedent around it, if necessary, for better readability?

src/tests/test_core_python.py
```diff
@@ -1,70 +1,99 @@
...
+        info = {}
+    return [name, info]
+
+
+@cache
+def get_tests():
+    with open(Path(__file__).parent / "python_tests.yaml") as file:
+        data = yaml.load(file, Loader)
+
+    return {name: info for test in data for [name, info] in [get_test_info(test)]}
```
rth, line 93: Maybe making this a generator that yields `name, info` would be a bit more readable and it should work with the rest of the code.

src/tests/test_core_python.py
```diff
@@ -1,70 +1,99 @@
...
+                del info[key]
+    return info
 
-    if any(flag.startswith("segfault") for flag in error_flags):
-        pytest.skip('known segfault: "{}"'.format(" ".join(error_flags)))
 
-    if error_flags:
+def possibly_skip_test(request, info: dict[str, Any]) -> dict[str, Any]:
+    for reason in (
+        reason for [flag, reason] in info.items() if flag.startswith("segfault")
```
rth, line 26: For tuple unpacking, it's usually a tuple not a list,
```suggestion
        reason for (flag, reason) in info.items() if flag.startswith("segfault")
```

src/tests/test_core_python.py
```diff
@@ -1,70 +1,99 @@
...
+    info = dict(info)
     # keep only flags related to the current browser
     flags_to_remove = ["firefox", "chrome", "node"]
-    flags_to_remove.remove(selenium.browser)
-    for flag in flags_to_remove:
-        if "crash-" + flag in error_flags:
-            error_flags.remove("crash-" + flag)
+    flags_to_remove.remove(browser)
+    for browser in flags_to_remove:
+        for key in list(info.keys()):
```
rth, line 18:
```suggestion
        for key in info:
```

src/tests/python_tests.yaml
```diff
@@ -0,0 +1,845 @@
...
+- test_rlcompleter
+- test_robotparser:
+    xfail: socket
+- test_runpy:
+    skip:
+      - test_pymain_run* # fork
+- test_sax
+- test_sched:
+    skip:
+      - "*concurrent" # threads
```
rth, line 617: Could you point me to where the wildcard is managed for skip (couldn't find it)?  Thanks!

rth (commented):
Thanks @hoodmane ! It certainly makes managing those tests easier. A few comments otherwise LGTM.

emsdk/patches/0007-Fix-lookupPath-when-applied-to-a-symlink-loop.patch
hoodmane: https://github.com/emscripten-core/emscripten/pull/16410

src/tests/test_core_python.py
hoodmane: Don't you get in place modification errors from this?

hoodmane: `RuntimeError: dictionary changed size during iteration`

PR #300, opened 2019-01-20
Make matplotlib output figures resizable

ashutoshbsathe:
Fixes pyodide/matplotlib-pyodide#15 

Here's a working example.
(Sorry for bad quality of recording, I don't have proper GPU on my machine)

I would like if someone tests this on their local machine first and then merge.
I've made sure that it works fine on all my devices, but still would make me confident if someone tries on their machine as well.

Oh, and I didn't remove the commented placeholder functions present [here](https://github.com/iodide-project/pyodide/blob/60977f0db8a2a6bf308851e6227c3d61b22f831d/packages/matplotlib/src/wasm_backend.py#L406). I thought I would wait till the maintainers give me green signal for the current design and then before final merge I will remove the placeholders

Please review

PR #2271, opened 2022-03-12
Refactor matplotlib tests

### LINKED ISSUE #2226: Matplotlib test is failing in CI
## 🐛 Bug

We are experiencing (occasional) failure in CI on [testing matplotlib](https://app.circleci.com/pipelines/github/hoodmane/pyodide/2186/workflows/e951d6f7-9949-41e0-ab7c-f7083de030a0/jobs/22549).

What I've found so far is that matplotlib test checks [whether the font is loaded](https://github.com/pyodide/pyodide/blob/8d00451142d5ddb0237595065773b1f7bf290d33/packages/matplotlib/test_matplotlib.py#L623-L625), but the [callback which updates the font_counter](https://github.com/pyodide/pyodide/blob/8d00451142d5ddb0237595065773b1f7bf290d33/packages/matplotlib/src/html5_canvas_backend.py#L414) is not executed properly. I am trying to figure this out but strangely I can't reproduce this behavior locally.

Note that matplotlib tests were disabled(xfailed) for a long time, and re-enabled (https://github.com/pyodide/pyodide/pull/2151) recently.

_Update:_ the failing test is xfailed now (#2243)

ryanking13:
### Description

Close #2226

Refactors matplotlib tests:

- Prevent dangling selenium webdriver after test
- For asynchronous tests, use promise-await instead of selenium WebDriverWait
- Enable xfailed tests


packages/matplotlib/meta.yaml
```diff
@@ -28,11 +28,11 @@ build:
     - png16=png
   post: |
     cd build/matplotlib-3.5.1/dist/matplotlib-3.5.1/
-    wget -O matplotlib/mpl-data/fonts/ttf/Humor-Sans.ttf http://antiyawn.com/uploads/Humor-Sans-1.0.ttf
```
ryanking13, line 31: We were downloading this font every time from someone's personal website.
It is [not mandatory for matplotlib installation](https://matplotlib.org/stable/api/_as_gen/matplotlib.pyplot.xkcd.html), but for backward compatibility, I embedded this font in our repository.

packages/matplotlib/src/html5_canvas_backend.py
```diff
@@ -21,7 +22,7 @@
 from PIL import Image
 from PIL.PngImagePlugin import PngInfo
 
-from js import FontFace, ImageData, XMLHttpRequest, document, window
+from js import FontFace, ImageData, document
```
hoodmane, line 25: Maybe would be good to produce a better error message if this fails:
```py
try:
	from js import FontFace, ImageData, document
except ImportError:
	raise ImportError("html5canvasbackend is only supported in the browser in the main thread")
```

packages/matplotlib/src/html5_canvas_backend.py
```diff
@@ -38,16 +39,13 @@ class FigureCanvasHTMLCanvas(FigureCanvasWasm):
...
-        window.font_counter = 0
-
     def create_root_element(self):
         root_element = document.createElement("div")
         document.body.appendChild(root_element)
         return root_element
 
     def get_dpi_ratio(self, context):
-        if hasattr(window, "testing"):
+        if os.environ.get("TESTING_MATPLOTLIB"):
```
hoodmane, line 48: is `TESTING_MATPLOTLIB` ever set?

hoodmane: Oh sorry, the diff for test_matplotlib was collapsed so when I searched for it I didn't see it.

packages/matplotlib/src/html5_canvas_backend.py
```diff
@@ -412,7 +404,12 @@ def _load_font_into_web(loaded_face):
         if font_face_arguments not in _font_set:
```
hoodmane, line 404: I think we should consider moving toward requiring people to explicitly install fonts and throwing an error otherwise, at least by default. If they can't figure out ahead of time what fonts to load, we could enable this approach as a compatibility mode fallback. I don't like this "guess what people want and do it on the fly", particularly as a default.

ryanking13: Yes, I also feel like the current font loading system is somewhat problematic. It requires a fixed fonts path and there are no explicit ways to install custom fonts (#2252). I am interested in enhancing matplotlib functionalities (#2270, #300, #2156), but I need some study of matplotlib internals.

packages/matplotlib/src/html5_canvas_backend.py
ryanking13: Thanks, applied!

packages/matplotlib/src/html5_canvas_backend.py
```diff
@@ -382,11 +366,24 @@ def _draw_math_text(self, gc, x, y, s, prop, angle):
...
+            if os.environ.get("TESTING_MATPLOTLIB"):
+                target_font = os.environ.get("TESTING_MATPLOTLIB_FONT")
+                if not target_font or target_font == fontface.family:
+                    try:
+                        from js import resolve
+
+                        resolve()
+                    except Exception as e:
+                        raise ValueError("unable to resolve") from e
+
```
hoodmane, line 386: What about calling here `self._test_helper(fontface)` and then:
```py
def _test_helper(self):
	pass
```
then monkey patch this logic into `_test_helper` in the test suite?

packages/matplotlib/src/html5_canvas_backend.py
```diff
@@ -38,16 +44,13 @@ class FigureCanvasHTMLCanvas(FigureCanvasWasm):
...
-        window.font_counter = 0
-
     def create_root_element(self):
         root_element = document.createElement("div")
         document.body.appendChild(root_element)
         return root_element
 
     def get_dpi_ratio(self, context):
-        if hasattr(window, "testing"):
+        if os.environ.get("TESTING_MATPLOTLIB"):
```
hoodmane, line 53: Can't we get rid of this and monkey patch `get_dpi_ratio` in the test?

ryanking13: That sounds reasonable, thanks for the suggestion!

packages/matplotlib/src/html5_canvas_backend.py
ryanking13: I'll try monkey-patching  `_load_font_into_web` directly.

hoodmane (approved):
Thanks @ryanking13!

PR #2272, opened 2022-03-13
Use pypa/build

### LINKED ISSUE #2189: Move away from setup.py
As mentioned by @henryiii is 

> Building via setup.py is very [heavily deprecated](https://blog.ganssle.io/articles/2021/10/setup-py-deprecated.html). I wouldn't recommend adding anything that depends on it.

With the other alternatives mentioned being [build](https://pypi.org/project/build/) and [installer](https://pypi.org/project/installer/).

I think build isolation would be a bit difficult to use in our case, as for instance when building scipy we need the patched numpy on the host and not the numpy version specified in pyproject.toml (which would be unpatched).  However trying to switch to `build`, for instance, without build isolation would certainly be worthwhile.

hoodmane:
This resolves #2189.

> build isolation would be a bit difficult to use in our case, as for instance when building scipy we need the patched numpy on the host and not the numpy version specified in pyproject.toml (which would be unpatched)

This is indeed the case, certain packages cannot be isolated. My strategy is to make a list of packages that shouldn't be isolated and add symlinks from the isolated build environment into the `.artifacts` directory to "unisolate" them. Then we remove the unisolated package requirements from the list of packages to install, in case pesky constraints aren't satisfied. In particular, packages that expect to be used with `pypa/build` often feel free to put very specific constraints on their build dependencies (often asking them to be == to a particular version). Specific version constraints is good for build reproducibility and with build isolation doesn't cost anything. So we just ignore the constraints. Hopefully nothing goes wrong.

In particular, any package that does stuff both at build time and at runtime and requires synchronization between the build time and run time environments needs the unisolation. This includes cffi with `_cffi_backend.so`, and of course numpy and scipy. pycparser needs to be unisolated because it is a dependency of cffi.

Currently I have also unisolated pythran and cython, though these are build time only tools and do not really need to be unisolated. Cython I unisolated specifically because numcodecs needs it but it isn't in the numcodecs build dependencies. Pythran I unisolated because of a problem with the scipy build which I don't fully understand (some problem with long double feature detection). I believe both of these are removable, particularly once #2132 is finished.


hoodmane:
@henryiii my local copy of pre-commit isort and the ci version don't agree with each other =(

henryiii:
You must have a local directory named `build`. We can specify that build is always third party to avoid it happening again. I need to update the url to GitHub for flake8 too, btw.

hoodmane:
Got it that makes sense. Thanks @henryiii. We will generally have a build directory so that would be good. 

Though I've been thinking it would be good to rename it dist because that's more accurate. Usually a build directory contains temporary build artifacts that aren't the resulting distribution.

pyodide-build/pyodide_build/pypabuild.py
```diff
@@ -0,0 +1,108 @@
...
+from typing import Mapping
+
+from build import BuildBackendException, ProjectBuilder  # type: ignore[import]
+from build.__main__ import (  # type: ignore[import]
+    _STYLES,
+    _error,
+    _handle_build_error,
+    _IsolatedEnvBuilder,
+    _ProjectBuilder,
+)
```
hoodmane, line 16: These are all pretty short (longest is `_handle_build_error` at 20 lines) and don't use any other private functions from `build` so we could copy most of this into this file. We should probably use some terminal color package instead of hand implementing colors since unlike `build` we aren't part of a bootstrap.

pyodide-build/pyodide_build/pypabuild.py
```diff
@@ -0,0 +1,108 @@
...
+
+from .common import get_hostsitepackages, get_pyversion
+
+UNISOLATED_PACKAGES = ["numpy", "scipy", "cffi", "pycparser", "pythran", "cython"]
+
+
+def symlink_unisolated_packages(env: IsolatedEnv):
+    pyversion = get_pyversion()
+    site_packages_path = f"lib/{pyversion}/site-packages"
+    env_site_packages = Path(env._path) / site_packages_path
```
hoodmane, line 28: We do need access to `env._path` which is private to `IsolatedEnv`. Maybe we should request that `pypa/build` should make `path` public, or otherwise add functionality like this.

hoodmane:
@henryiii would greatly appreciate your review on this, particularly on `pypabuild.py`.

pyodide-build/pyodide_build/pypabuild.py
henryiii: Rich. :D

pyodide-build/pyodide_build/pypabuild.py
henryiii: There's a rewrite for IsolatedEnv coming that the ideal person to review hasn't had time for. I might review it in light of this (I've also become a little more familiar with this part of build recently, so I might be able to review it too).

henryiii:
This is not exactly what I was thinking of, though it isn't bad - it actually does replicate the isolated environments, which is nice. What I was thinking was that we could run build with both verification and isolation disabled, and manually install/verify installation of the contents of `build-system.requires` (minus version pinning/limits). If you don't support manually specifying versions, isolation is _less_ important, but still nice; build tools could still have conflicting requirements, etc. I'll check with the rest of the build team to see what we'd be willing to make public.

pyodide-build/pyodide_build/pypabuild.py
henryiii: I feel like a lot of this is just to get the same output as build. If you used rich instead, you could probably get a really nice output too. Some reimplementation, but not too bad.

pyodide-build/pyodide_build/pypabuild.py
```diff
@@ -0,0 +1,108 @@
...
+    _handle_build_error,
+    _IsolatedEnvBuilder,
+    _ProjectBuilder,
+)
+from build.env import IsolatedEnv  # type: ignore[import]
+from packaging.requirements import Requirement
+
+from .common import get_hostsitepackages, get_pyversion
+
+UNISOLATED_PACKAGES = ["numpy", "scipy", "cffi", "pycparser", "pythran", "cython"]
```
ryanking13, line 22: So, if I am understanding correctly, one should add a package here if he/she wants to make the package unisolated? I'm not sure whether this list would be updated often, but if so, I think we should provide more easy and non-hardcoded way to update the list (for now it's fine).

hoodmane:
There are some packages that request `setuptools < 60.0` and others that request `setuptools >= 60.0`, so I felt that having build isolation is helpful. We seem to have been giving everything setuptools v57.0.0 so apparently the packages that request a very new setuptools in build_requires don't really mean it.

ryanking13 (commented):
Thanks, @hoodmane! I have some minor comments, and I'll review this again after studying about `build` a bit.

packages/lxml/meta.yaml
```diff
@@ -6,8 +6,8 @@ source:
   url: https://files.pythonhosted.org/packages/3b/94/e2b1b3bad91d15526c7e38918795883cee18b93f6785ea8ecf13f8ffa01e/lxml-4.8.0.tar.gz
 build:
   script: |
-    export WITH_XML2_CONFIG=$PYODIDE_ROOT/packages/libxml/build/libxml-2.9.10/xml2-config
-    export WITH_XSLT_CONFIG=$PYODIDE_ROOT/packages/libxslt/build/libxslt-1.1.33/xslt-config
+    export XML2_CONFIG=$PYODIDE_ROOT/packages/libxml/build/libxml-2.9.10/xml2-config
+    export XSLT_CONFIG=$PYODIDE_ROOT/packages/libxslt/build/libxslt-1.1.33/xslt-config
```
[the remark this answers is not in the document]
hoodmane: Yeah I should be able to revert this, I tried this when debugging other problems with environment variables.

packages/numcodecs/meta.yaml
```diff
@@ -8,6 +8,10 @@ source:
     - patches/fixblosc.patch
     - patches/fixsetup.patch
     - patches/fixzlib.patch
+build:
+  script: |
+    rm numcodecs/blosc.c
```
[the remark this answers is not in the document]
hoodmane: Yeah. `blosc.c` is the result of applying cython to `blosc.pyx`. We patch `blosc.pyx` in `patches/fixblosc.patch`. If Cython is installed, it will notice that `blosc.pyx` is new and recythonize it with the patched source, replacing `blosc.c`. If Cython is not installed, it will build with the old `blosc.c`. Removing `blosc.c` forces the build to fail when Cython is missing, which makes it less confusing.

On a related note, I am super confused about why the `fixblosc.patch` is needed at all anymore. Originally we needed it because `multiprocessing.Lock()` raises `ModuleNotFoundError: _multiprocessing not found`, whereas the `.pyx` code only caught `OSError`. (On a side side note, we should consider making `Lock` available as a no-op: since there is no multiprocessing, implementing locking by doing nothing is correct.)

Anyways numcodecs has since been patched to also catch `ImportError`:
https://github.com/zarr-developers/numcodecs/blame/v0.9.1/numcodecs/blosc.pyx#L82
I believe that `ModuleNotFoundError` is a subtype of `ImportError` so it should be covered. Indeed on native Python, the following import succeeds with no error:
```py
import multiprocessing
def patch_Lock():
	raise ModuleNotFoundError()
multiprocessing.Lock = patch_Lock
import numcodecs.blosc
```
However, the same code with the unpatched version of numcodecs in Pyodide fails with `ModuleNotFoundError`. I simply cannot comprehend why this happens. I spent a significant amount of time trying to locate the problem in the debugger before giving up.

pyodide-build/pyodide_build/pywasmcross.py
```diff
@@ -124,6 +132,9 @@ def compile(*, mypy__Single_overload_definition_multiple_required: int):
 
 
 def compile(env, **kwargs):
+    args = environment_substitute_args(kwargs, env)
+    args["builddir"] = str(Path(".").absolute())
+
     args = environment_substitute_args(kwargs, env)
```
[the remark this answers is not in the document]
hoodmane: I suspect this is the result of a sloppy merge.

pyodide-build/pyodide_build/pypabuild.py
```diff
@@ -0,0 +1,108 @@
...
+    os.environ.clear()
+    os.environ.update(build_env)
+    try:
+        yield
+    finally:
+        os.environ.clear()
+        os.environ.update(old_environ)
+
+
+None
```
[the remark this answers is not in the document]
hoodmane: No.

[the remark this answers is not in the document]
henryiii: You mean None. :)

pyodide-build/pyodide_build/pypabuild.py
hoodmane: How about we keep it like this for this PR and then make a second PR to switch to use rich to make nice terminal output and remove the dependence on these private `build` names?

pyodide-build/pyodide_build/pypabuild.py
hoodmane: > one should add a package here if he/she wants to make the package unisolated?

Yes.

> I'm not sure whether this list would be updated often

I hope not.

> think we should provide more easy and non-hardcoded way to update the list

I was thinking of adding a key to `meta.yaml`, but this requires some extra logic to collect the `UNISOLATED_PACKAGES`, so I think we can do that later. Even if we don't change it much, I think that having our build logic care about certain packages by name (like scipy) is a bit surprising. If nothing else, it would probably be a good idea to put a really obvious warning in the `meta.yaml` file.

henryiii: This would be a list of packages with C extensions that show up in build requirements, which is pretty small, but could grow. For something like `cmake`, which is only needed at build time (it's not linked in or anything), I would assume that would not be added here?

pyodide-build/pyodide_build/pypabuild.py
henryiii: You pin to an exact build version, so it's safe, sure.

pyodide-build/pyodide_build/pypabuild.py
hoodmane: Anything build time only should be isolated. This includes pythran, which I have unisolated as a hack rather than actually figuring out what is going wrong with scipy's build. For some reason, isolated pythran feature detects that there is no 128 bit float support even though emscripten has perfectly fine 128 bit float support. This breaks the boost stats module build.

packages/numcodecs/meta.yaml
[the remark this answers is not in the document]
ryanking13: Thanks for the detailed explanation!

> On a side side note, we should consider making `Lock` available as a no-op: since there is no multiprocessing, implementing locking by doing nothing is correct.

Yes, we could consider monkey-patching some of multiprocessing functionalities.

ryanking13 (approved):
Thanks again, @hoodmane! Generally LGTM. Could you please update documentation for build-unisolation?

hoodmane:
I opened a new PR for the docs #2297. I'll merge this now, let's work on updating the docs over there.

PR #2273, opened 2022-03-13
First steps in adding suitesparse and pyspqr for sparse QR deocmposition in scipy

gabrielfougeron:
Hi,

This pull request contains my first attempts in porting suitesparse and pyspqr to Pyodide. 
The code compiles, but there are problems in the linking process.

After talking to Hood Chatham about it, I realized I was in way over my head with this project as my knowledge of C/C++, make and other utilities is extremely limited.

This pull request was suggested by Hood Chatham, in order to allow more experience and knowledgeable people to help me in the porting process.

Note that I changed CLAPACK to build into separate .so files for blas and lapack. It was probably a bad move, do change it back if necessary.


packages/CLAPACK/meta.yaml
```diff
@@ -33,4 +33,6 @@ build:
 
     emmake make -j ${PYODIDE_JOBS:-3} blaslib lapacklib
     mkdir -p dist/lib
-    emcc blas_WA.a lapack_WA.a F2CLIBS/libf2c.a -sSIDE_MODULE -o dist/lib/clapack_all.so
+    emcc blas_WA.a lapack_WA.a F2CLIBS/libf2c.a -sSIDE_MODULE=1 -o dist/lib/clapack_all.so
+    emcc blas_WA.a -sSIDE_MODULE=1 -o dist/lib/libblas.so
+    emcc lapack_WA.a -sSIDE_MODULE=1 -o dist/lib/liblapack.so
```
hoodmane, line 38: This is going to create trouble because whenever CLAPACK is loaded all of the lapack and blas symbols will be linked twice. I'm not sure what the emscripten dynamic linker will do with all these duplicate symbols, but it seems worrisome. If suitesparse doesn't directly use dlopen then it should be okay to put these into e.g. `build/lib` and pass `-L$(PACKAGE_ROOT)/CLAPACK/build/lib` to suitesparse. At runtime, the symbols will come from `clapack_all.so` but they are the same symbols.

packages/suitesparse/meta.yaml
```diff
@@ -0,0 +1,43 @@
...
+    - CLAPACK
+
+build:
+  sharedlibrary: true
+  # set linker and C flags to error on anything to do with function declarations being wrong.
+  # In webassembly, any conflicts mean that a randomly selected 50% of calls to the function
+  # will fail. Better to fail at compile or link time.
+  ldflags: |
+    -Wl,--fatal-warnings
+    -sSIDE_MODULE=1
```
hoodmane, line 23: The `ldflags` key does nothing when `library` or `sharedlibrary` is `true`
```suggestion
```

packages/suitesparse/meta.yaml
```diff
@@ -0,0 +1,43 @@
+package:
+  name: suitesparse
+  version: 1.0.0
+
+source:
+  sha256: 1480acb174081599604f5e894bbe8a1a551de5904eef6ea67f33b25f884cd74f
+  url: https://gabrielfougeron.github.io/files/suitesparse.zip
```
hoodmane, line 7: Please change this to point to 
1. preferably the upstream release: https://github.com/DrTimothyAldenDavis/SuiteSparse/archive/refs/tags/v5.11.0.zip
2. failing that, a fork of SuiteSparse: https://github.com/gabrielfougeron/SuiteSparse
it's a bit difficult to work with it if it isn't in a github repo. Eventually we want to have a release version + some patch files, but while working on the PR it is fine to point at a personal fork.

packages/sparseqr/meta.yaml
```diff
@@ -0,0 +1,24 @@
+package:
+  name: sparseqr
+  version: 1.1.2
+
+source:
+  sha256: 10090126e6d12ed80e548541c78f4e681f5bf0f344d9d1e89cc15b499f0b5bf4     
+  url: https://gabrielfougeron.github.io/files/sparseqr.zip
```
hoodmane, line 7: Same comment as about suitesparse URL.

packages/suitesparse/meta.yaml
```diff
@@ -0,0 +1,31 @@
+package:
+  name: suitesparse
+  version: 1.0.0
+
+source:
+  url: https://github.com/gabrielfougeron/SuiteSparse
```
hoodmane, line 6:
```suggestion
  url: https://github.com/gabrielfougeron/SuiteSparse/archive/refs/heads/master.zip
```

gabrielfougeron:
> To get this to work, I think all you need to do now is to remove `add_subdirectory("programs")` https://github.com/gabrielfougeron/SuiteSparse/blob/8a934c3d437359097fe51b5bc2a29f94e26a6705/metis-5.1.0/CMakeLists.txt#L28

==> Done !

hoodmane:
Okay, now you need to:
1. rebase your changes to suitesparse on the upstream release branch v5.11.0. Ideally minimize the diff from v5.11.0 (so for instance don't remove irrelevant files). Then format a patch file with your changes. (The best would be to not change things at all, but to set environment variables and make targets that makes it build just the stuff you want. Our goal is always to minimize patching). Update the version field for suitesparse to be accurate too.
2. sparseqr also has an inaccurate version number (upstream seems to be v1.0.0 not v1.1.2). You should reduce the patch you made to `sparseqr_gen.py` and generate a patch file for it. Ideally, you should work with the upstream to remove the need for the patch.

For generating patches: 
```sh
git checkout -b pyodide-patches
git rebase master --onto=v1.0.0 # assuming you're currently on master branch, use whatever release tag is appropriate
git format-patch v1.0.0 -o /path/to/pyodide/packages/some_pkg/patches
```

hoodmane:
>   ==> Done!

Did you forget to push?

gabrielfougeron:
Argh sorry I've got tons of problems with github connection stability today. I'll take a second look tommorow.

gabrielfougeron:
Wait... Looks to me like I did push: https://github.com/gabrielfougeron/pyodide

Is there something I'm missing? I'm confused.

hoodmane:
> Is there something I'm missing?

For some reason github didn't update the PR with the new commit. It probably has to do with the recent github outage.

PR #2274, opened 2022-03-14
chore: isort avoid local build confusion, bump pre-commit

henryiii:
### Description

Quick update of pre-commit files, fixing URL of flake8, and making sure `build` doesn't trigger as a local directory in isort (#2272).


hoodmane:
Thanks @henryiii!

PR #2275, opened 2022-03-14
Implement more accurate wheel finding using the packaging tags module

hoodmane:
### Description

This adds a `find_matching_wheels` function that correctly determines which wheels the Pyodide interpreter supports according to the platform compatibility tags spec. This should fix problems with extra host platform wheels lying around or wheels built for different versions of Python (e.g., cp39 vs cp310).

### Checklists

- [x] Add / update tests

### DIFF
--- a/pyodide-build/pyodide_build/common.py
+++ b/pyodide-build/pyodide_build/common.py
@@ -1,7 +1,34 @@
 import functools
 import subprocess
 from pathlib import Path
-from typing import Optional
+from typing import Iterable, Iterator, Optional
+
+from packaging.tags import Tag, compatible_tags, cpython_tags, parse_tag
+
+PLATFORM = "emscripten_wasm32"
+
+
+def pyodide_tags() -> Iterator[Tag]:
+    """
+    Returns the sequence of tag triples for the running interpreter.
+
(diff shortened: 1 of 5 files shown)


pyodide-build/pyodide_build/common.py
```diff
@@ -1,7 +1,45 @@
...
+    ----------
+    wheel_paths
+        A list of paths to wheels
+
+    Returns
+    -------
+    The subset of wheel_paths that have tags that match the Pyodide interpreter.
+    """
+    wheel_tags_list: list[tuple[Path, frozenset[Tag]]] = [
+        (wheel, parse_tag(wheel.stem.split("-", 2)[-1])) for wheel in wheel_paths
```
ryanking13, line 36: can't we use [parse_wheel_filename](https://packaging.pypa.io/en/latest/utils.html#packaging.utils.parse_wheel_filename) instead? (for here and other places)

ryanking13 (approved):
Thanks, @hoodmane! LGTM with a minor comment.

pyodide-build/pyodide_build/common.py
hoodmane: Good point.

PR #2276, opened 2022-03-15
Fix emscripten patch paths

hoodmane:
Having to add `emsdk/upstream/emscripten` to all the emscripten patches is a nuisance. If at some point we want to patch a file in a different folder, we can separate the patches into multiple folders. In particular this should make it easy to migrate patches using `git am`, `git rebase`, and `git format-patch`.


ryanking13:
Do we still need `file_packager` patch after #2027?

hoodmane:
Yeah I think so. We still use `file_packager` for the standard library. Our approach for everything else is to package it in zip files, but we want to use Python to access the zip files so in order to bootstrap this we need a different way to load the standard library. Towards #646, we might consider trying to reduce the amount of stuff in the standard library to just the things needed to import shutil, and then load everything else from a zip file. But I think we will probably always use file_packager to get the first few Python files in.

PR #2277, opened 2022-03-16
Remove Py2JsResult

hoodmane:
The Py2JsResult type is incorrect, it should actually be `any`. Strictly
speaking, `any` except it won't be `null`. But I am not sure
that "anything except for null" provides extra value of "any" .

For backwards compatibility, we still `export type Py2JsResult = any;`

### DIFF
--- a/docs/project/changelog.md
+++ b/docs/project/changelog.md
@@ -63,6 +63,11 @@ substitutions:
 - {{Breaking}} Removed the `skip-host` key from the `meta.yaml` format. If
   needed, install a host copy of the package with pip instead. {pr}`2256`
 
+- {{Fix}} The type `Py2JsResult` has been replaced with `any` which is more
+  accurate. For backwards compatibility, we still export `Py2JsResult` as an
+  alias for `any`.
+  {pr}`2277`
+
 _February 19, 2022_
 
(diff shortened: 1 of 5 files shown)


ryanking13 (approved):
Thanks, @hoodmane!

How about explictly deprecating `Py2JsResult` then removing it in 0.21 or 0.22? (though it looks like `Py2JsResult` not documented anywhere)

hoodmane:
> How about explicitly deprecating Py2JsResult then removing it in 0.21 or 0.22?

Sounds good, though I'm not really sure how to deprecate a type. I think our types are still pretty unstable in general, in particular we haven't given any guidance for how to use them. Ideally we would generate a single `pyodide.d.ts` file and say that the correct way to use the types is to import them from `pyodide.d.ts` and not from the individual source files. Ideally the files under `src/js` would be considered completely private. But ideally we could discuss this with people more experienced with typescript...

ryanking13:
Hmm, actually it looks like we didn't include our js=>ts replacement in `0.19.1` release, and before that `Py2JsResult` was just a type in a jsdoc docstring. So... I think we can remove it without considering the backward compatibility?

PR #2279, opened 2022-03-16
Move making archive of shared libraries to buildpkg

ryanking13:
### Description

Moves the logic of generating zip archive of a shared library from `buildall` to `buildpkg` so that thread-unsafe `shutil.make_archive` won't bother us anymore.


hoodmane:
Thanks @ryanking13!

PR #2292, opened 2022-03-19
ENH Calculate indexURL from error stack trace

### LINKED ISSUE #2290: Allow specifying the version of Pyodide just once
## 🚀 Feature

It seems that as things stand, to use Pyodide, you have to include the right version of the script (e.g., `<script src="https://cdn.jsdelivr.net/pyodide/v0.19.1/full/pyodide.js"></script>`) and request the right directory for `loadPyodide` (e.g., `loadPyodide({indexURL : "https://cdn.jsdelivr.net/pyodide/v0.19.1/full/"})`). The script should probably provide a function or variable that tells you the version (if it's not possible for it to recapitulate the entirety of its own URL) so that the version doesn't have to be repeated for `loadPyodide`.

### Motivation

Having to specify the same version twice makes it easy to accidentally desynchronize the versions.

hoodmane:
If no indexURL is provided, we throw and catch an error and
then use ErrorStackParser to calculate where pyodide.js has
been loaded from. Resolves #2290.

### DIFF
--- a/src/js/pyodide.ts
+++ b/src/js/pyodide.ts
@@ -209,7 +209,11 @@ type ConfigType = {
 export async function loadPyodide(
   options: {
     /**
-     * The URL from which Pyodide will load packages
+     * The URL from which Pyodide will load the main Pyodide runtime and
+     * packages. Defaults to the url that pyodide is loaded from with the file
+     * name (pyodide.js or pyodide.mjs) removed. It is recommended that you
+     * leave this undefined, providing an incorrect value can cause broken
+     * behavior.
      */
     indexURL?: string;
 
--- a/src/js/pyodide.ts
+++ b/src/js/pyodide.ts
@@ -151,6 +151,35 @@ function finalizeBootstrap(config: ConfigType) {
 declare function _createPyodideModule(Module: any): Promise<void>;
 
 /**
+ *  If indexURL isn't provided, throw an error and catch it and then parse our
+ *  file name out from the stack trace.
+ *
+ *  Question: But getting the URL from error stack trace is well... really
+ *  hacky. Can't we use
+ *  [`document.currentScript`](https://developer.mozilla.org/en-US/docs/Web/API/Document/currentScript)
+ *  or
+ *  [`import.meta.url`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/import.meta)
+ *  instead?
+ *
+ *  Answer: `document.currentScript` works for the browser main thread.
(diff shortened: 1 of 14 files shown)


hoodmane:
@Kodiologist how's this?

Kodiologist:
Evil. I like it.

src/js/pyodide.ts
```diff
@@ -211,8 +212,15 @@ export async function loadPyodide(config: {
...
-  if (!config.indexURL) {
-    throw new Error("Please provide indexURL parameter to loadPyodide");
+  if (!options.indexURL) {
+    let err;
+    try {
+      throw new Error();
+    } catch (e) {
+      err = e;
+    }
+    const fileName = ErrorStackParser.parse(err)[0].fileName!;
```
ryanking13, line 222: This is well... really hacky. Can't [`document.currentScript`](https://developer.mozilla.org/en-US/docs/Web/API/Document/currentScript) or [`import.meta.url`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/import.meta) be an option?

ryanking13 (commented):
Thanks, @hoodmane. It would be great if one can just call `loadPyodide()` without passing any argument for general usage.

But the idea of getting URL from error stack trace is well... really hacky. Can't we use [`document.currentScript`](https://developer.mozilla.org/en-US/docs/Web/API/Document/currentScript) or [`import.meta.url`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/import.meta) instead? If they don't fit, then I'm fine with this solution.

Please update the changelog and documentation for the changed behavior.

hoodmane:
> Can't we use [document.currentScript](https://developer.mozilla.org/en-US/docs/Web/API/Document/currentScript) or [import.meta.url](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/import.meta) instead?

`document.currentScript` works for the browser main thread. `import.meta` works for es6 modules. In a classic webworker, there is no approach that works. Also there needs to be some other approach for node when using `require`. On the other hand, this stack trace approach works for every case without any feature detection code. Arguably the classic webworker case is the most important situation.

ryanking13:
> `document.currentScript` works for the browser main thread. `import.meta` works for es6 modules. In a classic webworker, there is no approach that works. Also there needs to be some other approach for node when using `require`. On the other hand, this stack trace approach works for every case without any feature detection code. Arguably the classic webworker case is the most important situation.

Thanks for the explanation! It's really hard to support various JS use cases... :(
Could you add this explanation to the comments for the people who later see that code?

PR #2293, opened 2022-03-19
MAINT Use tee to pipe stdout to build.log from buildpkg

hoodmane:
This makes it so that buildpkg generates build.log on its own
even when it isn't run from buildall. This is helpful because
when working on a specific package it is nicer to run buildpkg
directly so you can see the build on the console. However, it
is frustrating when you want to grep the build log for instance
that there isn't a build log.

This tees stdout into a build log from insidde buildpkg. It also
fixes the issue with overwriting build.log when nothing needs to
be done (though that issue I think was also fixed in December by
calling needs_build from buildall). Now buildall can just redirect
the subprocess output to /dev/null.

### DIFF
--- a/pyodide-build/pyodide_build/buildpkg.py
+++ b/pyodide-build/pyodide_build/buildpkg.py
@@ -693,7 +693,7 @@ def build_package(
         if not continue_:
             prepare_source(pkg_root, build_dir, srcpath, source_metadata)
 
-            run_script(build_dir, srcpath, build_metadata, bash_runner)
+        run_script(build_dir, srcpath, build_metadata, bash_runner)
 
         if build_metadata.get("library"):
             create_packaged_token(build_dir)
--- a/pyodide-build/pyodide_build/buildpkg.py
+++ b/pyodide-build/pyodide_build/buildpkg.py
@@ -678,12 +678,22 @@ def build_package(
             f"directory at the path {srcpath}, but that path does not exist."
         )
 
+    import os
+    import subprocess
+    import sys
+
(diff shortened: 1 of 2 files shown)


pyodide-build/pyodide_build/buildpkg.py
```diff
@@ -678,12 +678,22 @@ def build_package(
...
+    os.dup2(tee.stdin.fileno(), sys.stdout.fileno())  # type: ignore[union-attr]
+    os.dup2(tee.stdin.fileno(), sys.stderr.fileno())  # type: ignore[union-attr]
+
     with chdir(pkg_root), get_bash_runner() as bash_runner:
         bash_runner.env["PKG_VERSION"] = version
         if not continue_:
             prepare_source(pkg_root, build_dir, srcpath, source_metadata)
 
-        run_script(build_dir, srcpath, build_metadata, bash_runner)
+            run_script(build_dir, srcpath, build_metadata, bash_runner)
```
ryanking13, line 696: Shouldn't we run `script` even if the build continues from the middle? We often set environment variables that are needed for the build.

ryanking13 (approved):
Thanks, @hoodmane!

pyodide-build/pyodide_build/buildpkg.py
hoodmane: Well it depends, in scipy there is a command that concatenates a bunch of stuff into a file which we only want to do once. We should go back to being able to choose whether to run the script or not which used to be an option before #2256.

hoodmane:

PR #2297, opened 2022-03-22
Update new package docs and add more detailed meta.yaml validation

hoodmane:
These have gotten a bit out of date.

### DIFF
--- a/docs/development/new-packages.md
+++ b/docs/development/new-packages.md
@@ -99,8 +99,11 @@ If the package has a git repository, the easiest way to make a patch is usually:
 1. Clone the git repository of the package. You might want to use the options `git clone --depth 1 --branch <version>`. Find the appropriate tag given the version
    of the package you are trying to modify.
 2. Make a new branch with `git checkout -b pyodide-version` (e.g., `pyodide-1.21.4`).
-3. Make whatever changes you want. Commit them. Write a detailed commit
-   message!!
+3. Make whatever changes you want. Commit them. Please split your changes up
+   into focused commits. Write detailed commit messages! People will read them
+   in the future, particularly when migrating patches or trying to decide if
+   they are no longer needed. The first line of each commit message will also
+   be used in the patch file name.
 4. Use `git format-patch <version> -o <pyodide-root>/packages/<package-name>/patches/`
    to generate a patch file for your changes and store it directly into the
    patches folder.
--- a/docs/development/new-packages.md
+++ b/docs/development/new-packages.md
@@ -16,8 +16,7 @@ you should start this process with package dependencies.
 
 Most pure Python packages can be installed directly from PyPI with
 {func}`micropip.install` if they have a pure Python wheel. Check if this is the
-case by going to the `pypi.org/project/<package-name>` URL and checking if the
-"Download files" tab contains a file that ends with `*py3-none-any.whl`.
+case by trying `micropip.install("package-name")`.
 
 If the wheel is not on PyPI, but nevertheless you believe there is nothing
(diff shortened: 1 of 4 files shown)


docs/development/new-packages.md
```diff
@@ -96,43 +96,72 @@ then restart the build.
...
+5. Rebase the patches with `git rebase old-version --onto new-version`
+   (e.g., `git rebase pyodide-1.21.4 --onto pyodide-1.22.0`). Resolve any rebase
+   conflicts. If a patch has been upstreamed, you can drop it with `git rebase --skip`.
+6. Remove old patches with `rm <pyodide-root>/packages/<package-name>/patches/*`.
+7. Use `git format-patch <version-tag> -o <pyodide-root>/packages/<package-name>/patches/`
+   to generate new patch files.
+
+#### Upstream your patches!
+
+Please create PRs or issues to discuss with the package to try to find ways to
```
rth, line 129:
```suggestion
Please create PRs or issues to discuss with the package maintainers to try to find ways to
```

docs/development/new-packages.md
```diff
@@ -96,43 +96,72 @@ then restart the build.
...
+- If source is a url (not in-tree):
+  - Download a source archive or a pure python wheel (usually from PyPI)
+  - Confirm integrity of the package by comparing it to a checksum
+  - If building from source (not from a wheel):
+    - Apply patches, if any, to the source distribution
+    - Add extra files, if any, to the source distribution
+- If the source is not a wheel (building from a source archive or an in-tree
+  source):
+  - Run `build/script` if present
+  - Install wrappers for `gfortran`, `gcc`, `g++`, `ar`, and `ld` to intercept
```
rth, line 148: Maybe "Use wrappers" or "Create symlinks for wrappers of", as "install" sounds like it's going to install gfortran etc which is not the case.

docs/development/new-packages.md
```diff
@@ -96,43 +96,72 @@ then restart the build.
...
+    the calls, rewrite the arguments, and pass them to the appropriate
+    emscripten tools.
+  - Using `pypa/build`:
```
rth, line 151: Nice!

pyodide-build/pyodide_build/io.py
```diff
@@ -103,20 +80,123 @@ def check_package_config(
...
             if not isinstance(value, expected_type):
-                errors_msg.append(
+                yield (
                     f"Wrong type for '{section_key}/{subsection_key}': "
                     f"expected {expected_type.__name__}, got {type(value).__name__}."
                 )
 
-    if raise_errors and errors_msg:
+
+def _cpc_source(config: dict[str, Any]) -> Iterator[str]:
```
rth, line 89: What's "cpc"? maybve "check" or "validate"?

**Edit:** ah, I see check_package_config. Maybe `_check_config_source` to be more explicit

rth (commented):
Thanks a few comments otherwise LGTM.

Btw in the future, we could switch to using [Pydantic](https://pydantic-docs.helpmanual.io/) which is a better way of validating config files, and also allows to auto-generate a json schema from it (which could be adapted for auto-generating docs).

docs/development/new-packages.md
hoodmane: Incidentally, we could just check the symlinks in `tools/symlinks` into the repository. I decided not to do this because I think it's clearer to keep the "source of truth" for that in pywasmcross. Though we should ideally update pywasmcross to delete all the symlinks in `tools/symlinks` before creating new ones...

PR #2301, opened 2022-03-22
Handle signals other than SIGINT with interrupt_buffer

hoodmane:
If the user writes any nonzero value into `interrupt_buffer`, call
`PyErr_SetInterruptEx` on it. This silently ignores out of range
signals.

`PyErr_SetInterruptEx` is new in Python 3.10.


ryanking13 (approved):
Thanks, @hoodmane!

PR #2296, opened 2022-03-20
Add user to sudoers in run_docker

hoodmane:

PR #2302, opened 2022-03-23
Wait until docker container initialization in run_docker script

ryanking13:
This should fix https://github.com/pyodide/pyodide/pull/2296#issuecomment-1075519173.

A possibly better way would be using docker's HEALTHCHECK command, but this is more simple and we don't need to update docker image.


hoodmane:
Thanks @ryanking13!

PR #2305, opened 2022-03-23
PKG Add opencv-python and ffmpeg

### LINKED ISSUE #381: Port Open CV
Hello,
It was not clear in the readme or docs where to see all the packages that have been ported, but I would love to see [Open CV](https://pypi.org/project/opencv-python/)
ported.
Open CV is already a C based distribution and there are [efforts to port it to the browser.](https://docs.opencv.org/3.4/d4/da1/tutorial_js_setup.html)
Along with [Tensorflow](https://www.tensorflow.org/js) and the already ported NumPy package, these will allow for much of the real-time image-based machine-learning projects.

### LINKED ISSUE #1627: install opencv-python with micropip error
I am recently working on a project that use ML and this library is truly a game changer for project in js... Any luck to have opencv-python included in the list of the python libraries?  thank you in advance

ryanking13:
### Description

Adds opencv-python and ffmpeg.

### Status

opencv-python is modularized with a bunch of modules, I'll keep updating the status.

- [x] Core library
- [x] JPEG
- [x] PNG
- [x] WEBP
- [x] Image processing
- [x] video processing (FFMPEG)
- [x] objectdetect
- [x] dnn
- [x] ml
- [x] calib3d
- [x] feature2d
- [x] highgui (built, but nonsense)
- [x] photo
- [x] stitching
- [x] flann
- contrib algorithms?
- extra modules?

### Checklists

- [x] Add / update tests

Resolve #381
Resolve #1627


hoodmane:
Thanks for working on this @ryanking13 looks really cool.

We aren't supposed to use the Lenna test image anymore because it is considered to have a disparate impact on women. See for instance:
https://github.com/google/knusperli/issues/4

ryanking13:
Oh, didn't know that. I'll replace it with other images. Thanks for noting that out!

ryanking13:
BTW, I have only added half of core modules in opencv yet, but the build time already takes more than scipy 🤯

hoodmane:
At least it has no dependents I suppose. Maybe we can have a completely separate CI job for opencv.

ryanking13:
> At least it has no dependents I suppose. Maybe we can have a completely separate CI job for opencv.

It is only dependent on numpy.

pyodide-build/pyodide_build/pywasmcross.py
```diff
@@ -516,6 +517,19 @@ def handle_command(
     if args.pkgname == "scipy":
         scipy_fixes(new_args)
 
+    # FIXME: For some unknown reason,
+    #        opencv-python tries to link a same library (libopencv_world.a) multiple times,
+    #        which leads to 'duplicated symbols' error.
+    if args.pkgname == "opencv-python":
+        duplicated_lib = "libopencv_world.a"
```
ryanking13, line 524: At the final link time, it tries to link the same library multiple times. I couldn't find a way to prevent this from happening.

packages/opencv-python/meta.yaml
```diff
@@ -0,0 +1,144 @@
...
+    -DBUILD_opencv_highgui=ON \
+    -DBUILD_opencv_features2d=ON \
+    -DBUILD_opencv_flann=ON \
+    -DBUILD_opencv_calib3d=ON \
+    -DBUILD_opencv_dnn=ON \
+    -DBUILD_opencv_ml=ON \
+    -DBUILD_opencv_objdetect=ON \
+    -DWITH_OPENCL=OFF \
+    -DOPENCV_DNN_OPENCL=OFF \
+    -DWITH_PROTOBUF=ON \
```
ryanking13, line 85: The protobuf library is currently built with opencv. As it is used in many deep learning packages, I'll try to make it a standalone package in a seperate PR (for now, opencv is the only one which uses protobuf).

ryanking13 (commented):
Okay, I think this is ready to be reviewed.

I added all core modules in opencv-python. [There are also many extra modules](https://docs.opencv.org/4.x/), I think we can add them one by one if there is a request because not all extra packages are used widely.
This PR also adds FFmpeg library which is needed for video processing functionalities.

Both opencv-python and FFmpeg are quite big libraries. In our CI, opencv-python takes around 2300s to build (~ 2x of scipy build time) and FFmpeg takes around 450s to build (~ numpy build time).

Skylion007 (approved):

packages/opencv-python/meta.yaml
ryanking13: I was able to build libprotobuf in a separate package, but linking it with opencv generates a mysterious conflict between protobuf and ffmpeg. It looks like opencv builds [not all modules of protobuf](https://github.com/opencv/opencv/blob/66f3c2673cd424a9c282fb0c8622a4ae43d3e81d/3rdparty/protobuf/CMakeLists.txt#L56), so I'll let it build it's custom protobuf for now.

ryanking13:
After tons of trial-and-error, I think I finally made CI work...

I made a separate CI job for building opencv-python as hoodmane suggested. Now we have 4 CI jobs for building packages:

- `build-packages-no-numpy-dependents`: same as before
- `build-packages-numpy-dependents`: builds numpy dependent packages except opencv-python
- `build-packages-opencv-python`: builds opencv-python
- `build-packages`: actually doesn't build any package, but creates a packages.json from all packages built from other jobs.

`build-packages-numpy-dependents` and `build-packages-opencv-python` runs in parallel. Both jobs take around 35~40min to run, so I think the additional time overhead in CI of adding opencv is quite small. (though it will use more Circle CI quota, which is a tradeoff)

I also split out `libwebp` which supports encoding and decoding `webp` image format. After this PR, I'll link `libwebp` to other image processing packages such as Pillow.

packages/opencv-python/meta.yaml
```diff
@@ -0,0 +1,152 @@
...
+    sed -i "s/cmake_install_dir=cmake_install_reldir/_cmake_install_dir=cmake_install_reldir/" setup.py
+
+    # export VERBOSE=1
+
+    export NUMPY_INCLUDE_DIR="$HOSTINSTALLDIR/lib/python$PYMAJOR.$PYMINOR/site-packages/numpy/core/include/"
+    export EMSCRIPTEN="$PYODIDE_ROOT/emsdk/emsdk/upstream/emscripten/"
+    export FFMPEG_ROOT="$PYODIDE_ROOT/packages/ffmpeg/build/ffmpeg-4.4.1/lib"
+    export LIBWEBP_ROOT="$PYODIDE_ROOT/packages/libwebp/build/libwebp-1.2.2/build/lib"
+
+    export CMAKE_ARGS=" \
```
hoodmane, line 61: Maybe move this into a separate file and source it, it's pretty long?

.circleci/config.yml
```diff
@@ -150,12 +150,69 @@ jobs:
             - /root/.ccache
           key: -pkg1-v20220303-{{ checksum "Makefile.envs" }}
 
+      - run:
+          name: Zip build directory
+          command: |
+            tar cjf pyodide-build.tar.gz  dist
```
hoodmane, line 156: No action needed here, but we should rename these to `pyodide-dist`.

.circleci/config.yml
```diff
@@ -165,7 +222,71 @@ jobs:
...
+      - store_artifacts:
+          path: /root/repo/pyodide-build.tar.gz
+
+      - store_artifacts:
+          path: /root/repo/packages/build-logs
+
+      - store_artifacts:
+          path: /root/repo/build-logs.tar.gz
+
+  build-packages-opencv-python:
```
hoodmane, line 240: There is getting to be a lot of repetition in these package build steps, maybe as a followup we could try to extract some of it into custom-defined steps or something.

.circleci/config.yml
```diff
@@ -427,13 +548,28 @@ workflows:
...
+          requires:
+            - build-packages-no-numpy-dependents
+          filters:
+            tags:
+              only: /.*/
+
+      - build-packages:
+          requires:
+            - build-packages-numpy-dependents
+            - build-packages-opencv-python
```
hoodmane, line 568: It might be good to eventually split the package tests up too, so that we can do `no-numpy-dependents` tests before scipy and opencv finish building. This would make working on simple packages significantly more ergonomic -- it takes a long time for package tests to come back currently.

ryanking13: Yeah, that would be really great. We can run `no-numpy-dependents` test before building large numpy dependent packages, but we will need a minor fix to not run the same tests twice.

hoodmane: >  need a minor fix to not run the same tests twice

Yeah that's why I haven't done this...

.circleci/config.yml
ryanking13: Done :)
(Duplicated `post_steps` looks little ugly, but I can't find a way to pass a list parameter)

packages/opencv-python/patches/0001-Enable-file-system.patch
```diff
@@ -0,0 +1,75 @@
+From c7e9f892204ce1e47774fe21790195e2cbd7f2b3 Mon Sep 17 00:00:00 2001
+From: ryanking13 <def6488@gmail.com>
```
hoodmane, line 2: Did you make an upstream PR for this?

ryanking13: Not yet, but I am planning to do it.

hoodmane (approved):
Really great work @ryanking13! Feel free to merge if you are done with it.

ryanking13:
Thank you for the review @hoodmane!

PR #2306, opened 2022-03-23
Change extract_dir argument of unpackArchive to be a keyword argument

hoodmane:
Old usage is still accepted but causes a deprecation warning, saying
we will remove it in v0.21. Similar to #2300. I am planning to add
an option to do wasm compilation in a subsequent PR.

I also did some cleanup from #2300 and added tests for the deprecation 
warnings.


ryanking13 (approved):
Thanks, @hoodmane. LGTM. Please update the changelog and the deprecation timeline.

PR #2307, opened 2022-03-23
Merge together unpack_buffer and unpack_buffer_archive

hoodmane:
When I added unpack_buffer_archive, in code review people said
it was redundant with unpack_buffer and they should be merged.
I said merging was too annoying. They were right. This merges
the functions together into a function with a 38 line docstring
and a 18 line implementation.


src/py/pyodide/_package_loader.py
```diff
@@ -16,38 +16,97 @@
...
+        If calculate_dynlibs is True, a Javascript Array of dynamic libraries.
+        Otherwise, return None.
+
     """
-    target_dir = TARGETS[target]
+    if format:
+        format = get_format(format)
+    if target and extract_dir:
+        raise ValueError("Cannot provide both 'target' and 'extract_dir'")
+    if not filename and not format:
```
rth, line 95:
```suggestion
    if not filename and format is None:
```
technically format="" would be provided.

src/py/pyodide/_package_loader.py
```diff
@@ -16,38 +16,97 @@
...
         The name of the file we are extracting. We only care about it to figure
-        out whether the buffer represents a tar file or a zip file.
+        out whether the buffer represents a tar file or a zip file. Ignored if
+        format argument is present.
 
-    buffer
-        A Javascript ``Uint8Array`` with the binary data for the archive.
+    format
+        Controls the format that we assume the archive has. Overrides the file
+        extension of filename.
```
rth, line 67: It's not very clear what happens when format=None. Maybe we could say it more explicitly.

src/py/pyodide/_package_loader.py
```diff
@@ -16,38 +16,97 @@
...
-        Controls which directory the package is installed into. Either "site"
-        which installs the package into the sitepackages directory or "lib"
-        which installs the package into the standard library.
+        Controls which directory the file is unpacked into. Either "site" which
+        unpacked the file into the sitepackages directory or "lib" which
+        unpacked the file into the standard library. Mutually exclusive with
+        extract_dir.
+
+    calculate_dynlibs
+        If true, will return a Javascript Array of paths to dynamic libraries
```
rth, line 80: Maybe return a list rather, and then call to_js where we need it?

rth (approved):
Thanks!

src/py/pyodide/_package_loader.py
hoodmane: So the options here are:
1. If format is present, we use that.
2. If file name is present, it should have an extension, like `a.zip`, `a.tar`, etc. Then we use that.
3. If neither is present or the file name has no extension, `shutil.unpack_archive` throws an unknown format error.

src/py/pyodide/_package_loader.py
hoodmane: Well if someone says `format=""` we will throw an error further down anyways, but I guess it's an invalid format error rather than "you didn't give us a format"

src/py/pyodide/_package_loader.py
hoodmane: I guess we could do that. This is going to be more convenient since otherwise in Javascript we have to do:
```js
const dynlibspy = unpack_archive(...);
const dynlibs = dynlibspy.toJs();
dynlibspy.destroy();
```
which adds two extra lines of code. Also, note that if not `IN_BROWSER`, `to_js` is defined to be the identity function so this doesn't cause trouble for testing.

src/py/pyodide/_package_loader.py
rth: I imagine. I meant that option 2 (which is the default) could be documented as well.

src/py/pyodide/http.py
```diff
@@ -202,9 +202,7 @@ async def unpack_archive(self, *, extract_dir=None, format=None):
         """
         buf = await self.buffer()
         filename = self._url.rsplit("/", -1)[-1]
-        unpack_buffer_archive(
-            buf, filename=filename, format=format, extract_dir=extract_dir
-        )
+        unpack_buffer(buf, filename=filename, format=format, extract_dir=extract_dir)
```
ryanking13, line 205: If we are going to use `unpack_buffer` here too. I think `unpack_buffer` should be more general, than having specific features of handling Python packages.

hoodmane: Well it's a private function so I think it can do what it wants. I actually think all of our public apis for unpacking archives should be modified to have an option to compile wasm they encounter. Then they will all have good reason to use this.

My reasoning is that I want to advertise that `pyfetch` or `pyodide.unpackArchive` can be used to install packages with no dependency management. But they can't install packages with any binary extensions because they don't compile wasm. So I think we should make it an option for them to compile wasm (and maybe make it the default option).

ryanking13: > Well it's a private function so I think it can do what it wants.

My tiny concern is that we may update `unpack_buffer` and it results in unexpected side effect on `FetchResponse.unpack_archive`.

> I actually think all of our public apis for unpacking archives should be modified to have an option to compile wasm they encounter. Then they will all have good reason to use this.

> My reasoning is that I want to advertise that `pyfetch` or `pyodide.unpackArchive` can be used to install packages with no dependency management. But they can't install packages with binary extensions because they don't compile wasm. So I think we should make it an option for them to compile wasm (and maybe make it the default option).

Yes, I agree that we want and should make our unpacking APIs to be used for installing packages.
I think the problem is that we don't have `urllib` or `requests` support yet, so people should also use them for a general perpose unpacking method.

ryanking13 (approved):

src/py/pyodide/http.py
hoodmane: > so people should also use them for a general purpose unpacking method.

Yeah they can use that too. I think we should add a `compile_wasm` boolean argument that controls the behavior. But most `.so` files that people download are probably actual wasm libraries, so I think it would be okay to default true and having people turn it off if they have `.so` files that aren't wasm libraries (or if they are wasm libraries but they don't want to compile them until later for some reason).

src/py/pyodide/_package_loader.py
hoodmane: I am going to leave this as-is for now, we could think about changing it later if you have a strong opinion or a good reason arises.

rth: Since it's a private function, OK to keep as is.

PR #2309, opened 2022-03-24
Remove pyodide-interrupts

hoodmane:
There is really no reason to use this instead of `setInterruptBuffer`. 
We've been thinking about removing it since v0.17.


docs/project/changelog.md
```diff
@@ -28,6 +28,10 @@ substitutions:
   suite, but none seem to be simple linker errors.
   {pr}`2289`
 
+- Removed pyodide-interrupts. If you were using this for some reason, use
+  `setInterruptBuffer` instead.
```
rth, line 32: Could you link to this function in docs?

rth (approved):

PR #2314, opened 2022-03-25
DOC Add deprecation timeline

### LINKED ISSUE #2303: [Discussion] Deprecation policy
Recently, we are updating our JS/Python APIs often, which results in deprecating several APIs. So I think it would be great to have some explicit deprecation policy.

Since we are in 0-based versioning, we have some free space to make breaking changes without considering backward compatibility too much. Still, having an explicit deprecation policy can help Pyodide users but also Pyodide devs, as we don't have to say "when should this API be removed?" every time.

Regarding our release cycle of around 3~4 months, I think (unless it's a critical change) **two minor version updates** between deprecation and removal would be reasonable so that we can notify users about the removal at least 6 months ago.

WDYT?

ryanking13:
### Description

Adds a deprecation timeline/policy page in docs.

Close #2303

### DIFF
--- a/docs/project/changelog.md
+++ b/docs/project/changelog.md
@@ -991,3 +991,10 @@ _Mar 21, 2019_
   have been removed.
 
 - The `run_docker` script can now be configured with environment variables.
+
+```{eval-rst}
+.. toctree::
+   :hidden:
+
+   deprecation-timeline.md
+```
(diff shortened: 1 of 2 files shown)


hoodmane (approved):
Thanks @ryanking13!

ryanking13:
The test failure is unrelated to this PR, so I'll merge.

PR #2319, opened 2022-03-26
ENH Automatically schedule coroutines

hoodmane:
Without this PR,
```js
let f = pyodide.globals.get("some_async_function");
setTimeout(f, 100);
```
doesn't work because `setTimeout` calls `f` which returns a coroutine which is left unscheduled and so the actual work in `f` is never executed. This is surprising to people, see for instance https://github.com/pyodide/pyodide/discussions/2229. This changes the behavior to automatically schedule all coroutines created from async functions called from Javascript so that async functions can be used as Javascript event handlers.

### DIFF
--- a/docs/project/changelog.md
+++ b/docs/project/changelog.md
@@ -101,6 +101,11 @@ substitutions:
 - {{ Enhancement }} Updated to Emscripten 2.0.27.
   {pr}`1102`
 
+- {{ Enhancement }} Async Python functions called from Javascript now have the
+  resulting coroutine automatically scheduled. For instance, this makes it
+  possible to use an async Python function as a Javascript event handler.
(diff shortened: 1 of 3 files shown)


ryanking13 (approved):
Thanks, @hoodmane! This looks very interesting. LGTM.

If I am understanding correctly, when we pass async Python function JS event handler, we will now always get `RuntimeWarning: coroutine 'f' was never awaited`, as it is scheduled but not awaited?

hoodmane:
> when we pass async Python function JS event handler, we will now always get `RuntimeWarning: coroutine 'f' was never awaited`, as it is scheduled but not explicitly awaited?

Exactly. I'm not sure that behavior is useful to anyone, hence this PR.

PR #2326, opened 2022-03-28
ENH Dependency management for .so loader code

hoodmane:
This implements more accurate dependency management for `.so` files. Some `.so` files say in their header that they depend on other `.so` files. If Emscripten sees this and judges that those other `.so` files haven't already been loaded, it will try to load them automatically by making http requests, etc. Its attempts to do this will fail, so we don't want it to do this. So we look up the dependencies of the `.so` file ourselves and wait until they are satisfied before proceeding. If we can't seem to satisfy the dependencies, then we can raise our own more sensible error.

The strategy to wait for the dependencies .so files is as follows:
1. If the dependency is already loaded, we're good
2. If not, check if there is a .so file of the right name in the current package we are loading. If so, wait for that one to load and we are good.
3. If not, pick a random package that the current package depends on and wait for it to load, then check again if our .so dependency is loaded. Continue until we run out of packages.
4. If the dependency .so is still not loaded we throw an error.

If somehow there were a circular dependency between two .so files in the same package this will just hang. But I think it would require intentional mischief to produce a circular dependency. The old approach wouldn't work in general -- it was very sensitive to the division of `.so` files into packages. This new approach should work with practically any set of .so files. For instance this fixes the case where one shared library depends on another, and the case where one .so depends on another in the same package.

### DIFF
--- a/Makefile.envs
+++ b/Makefile.envs
@@ -1,6 +1,8 @@
 export PYVERSION ?= 3.10.2
 export PYODIDE_EMSCRIPTEN_VERSION ?= 2.0.27
 
+export PLATFORM_TRIPLET=wasm32-emscripten
+
 # BASH_ENV tells bash to run pyodide_env.sh on startup, which sets various
 # environment variables. The next line instructs make to use bash to run each
 # command.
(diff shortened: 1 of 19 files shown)


rth:
Sorry for the lack of review. So, in the end, you prefer to choose a different solution, or come back to this in the future?

hoodmane:
I think @ryanking13 fixed this recently in #2686. The exploration I did in this PR was good enough that I could tell him how to fix it when he ran into the problem this is was supposed to prevent when loading sparseqr.

rth:
OK, great! A saw that issue in passing, but couldn't find it again before commenting here. Thanks!

PR #2327, opened 2022-03-28
MAINT Bump zlib version

hoodmane:
Apparently zlib v1.2.11 had an out of bounds memory access bug so
they yanked it. This bumps the version and gets the build working
again.


rth:
zlib build still fails, but otherwise LGTM.

hoodmane:
I'm still surprised we don't get duplicate symbol errors from linking a copy of zlib into the main module and also into side modules...

ryanking13 (approved):

PR #2328, opened 2022-03-29
chore: fix black breakage from click 8.1

henryiii:
See psf/black#2964 for details.


rth (approved):
Thanks @henryiii ! This broke the CI everywhere )

PR #2329, opened 2022-03-30
MAINT merge console scripts into one

ryanking13:
### Description

We have two Python console scripts: `_pyodide/console.py` and `pyodide/console.py`.
I think there is no clear reason for them to be separated, and Sphinx autodoc often complains about duplicate symbols. This PR merges them into `pyodide/console.py`.


hoodmane:
Right, I guess there is no chance that we would need to use `Console` in the Pyodide bootstrap. It's important that we have `eval_code` before we get `_pyodide_core` working, but we won't need this.

Thanks @ryanking13!

ryanking13:

PR #2331, opened 2022-03-30
Remove bad ctypes patch

### LINKED ISSUE #2330: ctypes patch breaks ctypes's in_dll() method
## 🐛 Bug

I think that the ctypes patch https://github.com/pyodide/pyodide/blob/main/cpython/patches/0006-Don-t-dereference-function-pointer.patch is incorrect and unnecessary. It breaks the ``in_dll`` method of ctypes primitives. The ``CDataType_in_dll()`` function does not return a function pointer. It returns a global variable.

### To Reproduce

The problem can be reproduced easily in the interactive REPL at https://pyodide.org/en/stable/console.html and ``PyAPI_DATA(int) Py_OptimizeFlag;``.

```
Python 3.9.5 (default, Feb 22 2022 14:12:02) on WebAssembly VM
>>> import ctypes, sys
>>> sys.flags.optimize
0
>>> ctypes.c_int.in_dll(ctypes.pythonapi, "Py_OptimizeFlag")
c_long(3223484)
```
### Expected behavior

The correct output is ``c_long(0)``.

```
Python 3.10.2 (main, Jan 17 2022, 00:00:00) [GCC 11.2.1 20211203 (Red Hat 11.2.1-7)] on linux
>>> import ctypes, sys
>>> sys.flags.optimize
0
>>> ctypes.c_int.in_dll(ctypes.pythonapi, "Py_OptimizeFlag")
c_int(0)
```

### Environment

- Pyodide Version: Python 3.9.5 (default, Feb 22 2022 14:12:02) on WebAssembly VM
- Browser version: Firefox
### Additional context

I can also reproduce the issue with upstream CPython 3.11 (latest head):

without patch:

```
$ node --experimental-wasm-threads --experimental-wasm-bulk-memory cpython/builddir/emscripten-node/python.js -c 'import ctypes; print(ctypes.c_int.in_dll(ctypes.pythonapi, "Py_OptimizeFlag"))'
c_long(0)
$ node ... python.js -O -c ...
c_long(1)
$ node ... python.js -OO -c ...
c_long(2)
```

With the ctypes patch:

```
$ node --experimental-wasm-threads --experimental-wasm-bulk-memory cpython/builddir/emscripten-node/python.js -c 'import ctypes; print(ctypes.c_int.in_dll(ctypes.pythonapi, "Py_OptimizeFlag"))'
c_long(3762604)
```

hoodmane:
This resolves #2330.

As tiran helpfully points out, the `Dont-dereference-function-pointer`
patch is wrong. Most likely it was always wrong and I added it thinking
that I was fixing a problem that never existed or that was fixed by some
other change (possibly to the libffi port) that I didn't test separately.
Removing this patch makes many more of the ctypes tests pass.


hoodmane:
Remaining failing tests:
- `test_callback_too_many_args`: requires a different (simpler but slower) JS trampoline approach in libffi. Might be worth switching because not having fatal errors when the signature is wrong is probably more important than being as fast as possible.
- `test_callback_large_struct`: there is an upstream `large_struct` test in `libffi` which also fails. Can't remember if it's due to some fundamental wasm limitation (or if I ever knew)
- `CFunctions`
- `test_struct_return_2H`
- `FunctionTestCase`

And also we have the problem that `PyCode_SetExtra` causes memory corruption, which is not related to ctypes but causes failures in the ctypes test suite.

rth (approved):
Great news!

hoodmane:
I think there are four or five failing libffi tests. It would be good to compare the remaining failing tests to the failing libffi tests and see if they match. If any of the failures here are due to bugs in my libffi port that are not covered upstream, we probably should add more tests to the libffi test suite (the libffi folks have appreciated my previous contributions to their test suite).

hoodmane:
And a few extra tests fail on node for some reason.

PR #2332, opened 2022-03-31
New keyboard interrupt patch from upstream

hoodmane:
This replaces the keyboard interrupt patch with one that I will try to upstream.


rth:
Thanks! Let's wait for the review of your upstream MR.

hoodmane:
Yeah I really made this to reference from that PR to give some indication that it works.

hoodmane:
Okay upstream patch was merged. The upstream patch isn't identical to this one, but backporting the final patch is nontrivial and they have the same interface (so if we accept this PR then when upgrading to v3.11 we can just delete the patch without changing any other code).

hoodmane:
I guess I should update this patch with some of the changes that went in before merging upstream:
* `Py_EMSCRIPTEN_SIGNAL_HANDLING` got marked with `EMSCRIPTEN_KEEPALIVE`,
* `_Py_CheckEmscriptenSignals_Helper` got adjusted to be a bit more robust,
* the signal check interval 50 got turned into a separate define

hoodmane:
@ryanking13 For some reason github actions is applying the patch that I removed [cpython/patches/0007-Patch-in-keyboard-interrupt-handling.patch](https://github.com/pyodide/pyodide/pull/2332/files#diff-466e5bc31cab30d4b671fff94cb0fbf8eee4be9b0894c58e436cc84cf9efa250) in addition to the patch that I added. CircleCI gets it right. Any idea why this is happening?

ryanking13:
> @ryanking13 For some reason github actions is applying the patch that I removed [cpython/patches/0007-Patch-in-keyboard-interrupt-handling.patch](https://github.com/pyodide/pyodide/pull/2332/files#diff-466e5bc31cab30d4b671fff94cb0fbf8eee4be9b0894c58e436cc84cf9efa250) in addition to the patch that I added. CircleCI gets it right. Any idea why this is happening?

Can you try merging main, re-check that the patch file has been removed, then re-trigger CI?

#2350 renamed `0007-Patch-in-keyboard-interrupt-handling.patch` to `0006-Patch-in-keyboard-interrupt-handling.patch`. So, after merge,  `0006-Patch-in-keyboard-interrupt-handling.patch` will remain. Maybe that can be the reason?

ryanking13:
Related link: https://github.com/actions/checkout/issues/15

hoodmane:
It's strange that gha would do that even when there is a merge conflict...

hoodmane:
Thanks for the explanation!

ryanking13:
> It's strange that gha would do that even when there is a merge conflict...

No, GHA triggered by PR doesn't run when there is a merge conflict (https://github.community/t/run-actions-on-pull-requests-with-merge-conflicts/17104/2)

hoodmane:
Any objections to merging this? I've triple-checked that the patch is a pretty reasonable backport, and the same changes in the Pyodide code work when built against Python tip of tree here:
https://app.circleci.com/pipelines/github/hoodmane/pyodide/2670/workflows/20d3f1cf-4a65-4591-8146-c26744067458/jobs/28989

rth (approved):
No objections, thanks for doing and upstreaming this!

PR #2334, opened 2022-03-31
Update openssl to 1.1.1n

hoodmane:
Update says: 'Security Advisory: one high severity fix' so probably
a good idea to include it.

### DIFF
--- a/packages/ssl/meta.yaml
+++ b/packages/ssl/meta.yaml
@@ -10,7 +10,7 @@ build:
     export DISTDIR=$(pwd)/dist
     cd $CPYTHONBUILD
     emcc $STDLIB_MODULE_CFLAGS -c Modules/socketmodule.c -o Modules/socketmodule.o
-    emcc $STDLIB_MODULE_CFLAGS -c Modules/_ssl.c -o Modules/_ssl.o -I$(OPEN_SSL_ROOT)/include/ \
+    emcc $STDLIB_MODULE_CFLAGS -c Modules/_ssl.c -o Modules/_ssl.o -I$OPEN_SSL_ROOT/include/ \
       -DOPENSSL_THREADS # This declares that OPENSSL is threadsafe. We are single threaded so everything is threadsafe.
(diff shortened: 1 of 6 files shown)


packages/ssl/meta.yaml
```diff
@@ -10,7 +10,7 @@ build:
     export DISTDIR=$(pwd)/dist
     cd $CPYTHONBUILD
     emcc $STDLIB_MODULE_CFLAGS -c Modules/socketmodule.c -o Modules/socketmodule.o
-    emcc $STDLIB_MODULE_CFLAGS -c Modules/_ssl.c -o Modules/_ssl.o -I$PYODIDE_ROOT/packages/openssl/build/openssl-1.1.1m/include/ \
+    emcc $STDLIB_MODULE_CFLAGS -c Modules/_ssl.c -o Modules/_ssl.o -I$(OPEN_SSL_ROOT)/include/ \
```
ryanking13, line 13:
```suggestion
    emcc $STDLIB_MODULE_CFLAGS -c Modules/_ssl.c -o Modules/_ssl.o -I$OPEN_SSL_ROOT/include/ \
```

ryanking13 (approved):
Thanks, @hoodmane!

packages/ssl/meta.yaml
hoodmane: I was totally confused by this! It's so hard to keep track of all the different ways to substitute variables in shell and makefile and autotools etc.

ryanking13: The problem is that we are using `$(VAR)` syntax in substituting env variables in `cflags, cxxflags, ldflags` which means 'execute command' in a shell. We need to change this in a way compatible to shell behavior ($VAR or ${VAR})

https://github.com/pyodide/pyodide/blob/3c2ae15052fc391668d4d2bbba5ff3c3d3e910ad/pyodide-build/pyodide_build/pywasmcross.py#L526-L537

PR #2335, opened 2022-03-31
Fix matplotlib event listeners

hoodmane:
This fixes https://github.com/pyodide/pyodide/discussions/2270, see also:

> on mouse hover (for all versions) in the console I get,
> ```
> Error: This borrowed proxy was automatically destroyed at the end of a function call. Try using create_proxy or create_once_callable.
> The object was of type \&quot;method\&quot; and had repr \&quot;&lt;bound method FigureCanvasWasm.onmousemove of &lt;matplotlib.backends.html5_canvas_backend.FigureCanvasHTMLCanvas object at 0xab2640&gt;&gt;\&quot;"
> ```
> (thanks @hoodmane  for improved error messages!)
> 
> so it sounds like this would need a fix in the MPL canvas backend.

_Originally posted by @rth in https://github.com/pyodide/pyodide-blog/pull/12#pullrequestreview-928200198_

It would be good if we had tests that actually hit these event listeners.

TODO for Pyodide v0.21: Add similar wrappers for `setTimeout`, `setInterval`, `addEventListener`, and their inverses to the Pyodide Python API.

Could someone else test this @rth @madhur-tandon? I don't ever use matplotlib so I am not sure how to test it. Ideally we could get tests that actually trigger these events in browser. I think it shouldn't be too hard to add selenium tests for `WasmTimer` at least. Presumably we could also mock mouse movement...


packages/matplotlib/src/browser_backend.py
```diff
@@ -1,8 +1,67 @@
 import math
+from types import Any, Callable
```
ryanking13, line 2:
```suggestion
from typing import Any, Callable
```

ryanking13 (approved):
Thanks, @hoodmane!

We really need to test matplotlib more. It's very easy to break, and for now we are not testing our default AGG backend. Maybe we can start deprecating AGG backend and start using canvas backend more, but I am not very experienced in matplotlib to do that decision.

hoodmane:
Since the AGG backend supposedly has better performance but larger code size, it is not clear to me that deprecating/removing it is a good deal. And both are missing test coverage. In any case, matplotlib isn't a super high priority for me in the short term: in some sense, it's the furthest down the stack of anything in Pyodide since it gets to specific UI rendering stuff. And when we do get to these more expansive needs, I would prioritize like getting synchronous IO working.

It is useful to have matplotlib to figure out what downstream might be concerned about: for instance it was helpful when designing the `create_proxy` and `getBuffer` APIs, and it gave the idea that we might include wrappers for `setTimeout` and `setInterval`.

madhur-tandon:
@hoodmane seems like `Callable` is not used correctly.
The correct usage is `Callable[[arg, ...], result]`

rth:
Yes, I think we would want to support both backend.

henryiii:
https://github.com/pyodide/pyodide/blob/389f8ff1d2cd4ae4bdb8d4669528483e175bd0b7/.pre-commit-config.yaml#L54 Perhaps?

henryiii:
Ahh, the problem was `Callable[[]]` which is - not really sure what that is. Hmmm, that does seem like something it should catch.

hoodmane:
Hmm, is it possible to list out which missing imports we want ignored? We really just need it to ignore `js`, `pyodide_js`, and maybe a couple others. If we can't do that, maybe we could add stub files to tell it what we expect to be in `js` and `pyodide_js`

henryiii:
FYI, the backend is not considered a GUI backend, so produces an warning or error when you call `.show()`. Been meaning to look into it but haven't had a chance recently, finishing up an unrelated project.

Sure, we can list exactly what missing imports we want (and that's what I'd always recommend). Probably should tighten up the mypy testing, the more strict it is, the more it helps.

henryiii:
I'm the same way. Though mypy will very much let stuff leak through if you don't tighten it up as much as possible. I've started moving past "strict=True" with three more optional strictness checks that catch even more things.

I think what happened here is you imported something that didn't exist, mypy ignored it as a missing import, so it was "Any" and `Any[[]]` is fine. Seems a little weird it would let you import something that didn't exist from an existing location. It's not letting me do that locally why I try it. Hmm, might need to make sure these are not getting skipped...

packages/matplotlib/src/browser_backend.py
```diff
@@ -1,8 +1,67 @@
...
+try:
+    from js import devicePixelRatio as DEVICE_PIXEL_RATIO
+except ImportError:
+    DEVICE_PIXEL_RATIO = 1
+
+
+EVENT_LISTENERS: dict[tuple[JsProxy, str, JsProxy], JsProxy] = {}
+
+
+def add_event_listener(elt: JsProxy, event: str, listener: Callable[[Any]]):
```
rth, line 18: Can you make all of these private, so we don't have to care about backward compatibility? `_add_event_listener` etc.

packages/matplotlib/src/browser_backend.py
```diff
@@ -1,8 +1,67 @@
...
+
+
+TIMEOUTS: dict[int, Destroyable] = {}
+
+
+def set_timeout(callback: Callable[[]], timeout: int) -> int:
+    id = None
+
+    def wrapper():
+        nonlocal id
```
rth, line 42: Are you sure it's necessary? I would have guessed it should work without this.

packages/matplotlib/src/browser_backend.py
```diff
@@ -1,8 +1,67 @@
...
+        callback()
+        del TIMEOUTS[id]
+
+    id = setTimeout(create_once_callable(wrapper), timeout)
+    return id
+
+
+DUMMY_DESTROYABLE = Destroyable()
```
rth, line 50: Could you add a comment explaining the logic with this, please?

rth (approved):
Thanks a lot! Needs a changelog entry as well.

packages/matplotlib/src/browser_backend.py
hoodmane: It's just an object with a no-op destroy method so I can do `TIMEOUTS.pop(id, DUMMY_DESTROYABLE).destroy()` and either it gets a real object and calls the real `destroy` method or it gets the fake which does nothing. This is to handle the case where `clear_timeout` is called after the timeout executes.

ryanking13 (approved):

PR #2339, opened 2022-04-01
fix: better mypy coverage

henryiii:
- fix(types): better mypy running
- fix(types): all src checked for imports
- chore(types): unreachable
- chore(types): enable a few easy settings

### Description

See #2335.


packages/test_packages_common.py
```diff
@@ -24,7 +24,7 @@ def registered_packages_meta():
     """Returns a dictionary with the contents of `meta.yaml`
     for each registered package
     """
-    packages = registered_packages
+    packages = registered_packages()
```
henryiii, line 27: Pretty sure this is a fix. You can't iterate a raw function...

pyproject.toml
```diff
@@ -1,8 +1,25 @@
 [tool.mypy]
-warn_unused_configs = true
 python_version = "3.10"
-mypy_path = "src/py"
+mypy_path = ["src/py", "pyodide-build"]
+show_error_codes = true
+warn_unreachable = true
+enable_error_code = ["ignore-without-code", "redundant-expr", "truthy-bool"]
+
+# Strict checks
```
henryiii, line 8: Ideally, all of these should eventually be true (and then you could just replace them with `strict = true`), but that would take some work. I tried to enable a few easy ones.

.pre-commit-config.yaml
```diff
@@ -49,25 +49,25 @@ repos:
     rev: "v0.942"
     hooks:
       - id: mypy
-        args:
-          - --show-error-codes
-          - --ignore-missing-imports
-        files: ^(src/|packages/.*/test|conftest.py|docs/)
+        args: [--ignore-missing-imports]
```
henryiii, line 52: This is really hard to disable, since every test file will have different imports based on what it's testing. But I did move the `/src` checking (after enabling it) over to the next check where this is not enabled.

Thinking about it, maybe `/src` can move there too, and thsi could just be the packages's non-src files and that's all.

henryiii:
List of unchecked files:

```
benchmark/benchmark.py
benchmark/benchmarks/matplotlib_benchmarks/canvas_custom_font.py
benchmark/benchmarks/matplotlib_benchmarks/canvas_image.py
benchmark/benchmarks/matplotlib_benchmarks/canvas_image_affine.py
benchmark/benchmarks/matplotlib_benchmarks/canvas_rendering.py
benchmark/benchmarks/matplotlib_benchmarks/canvas_text_rotated.py
benchmark/benchmarks/matplotlib_benchmarks/wasm_custom_font.py
benchmark/benchmarks/matplotlib_benchmarks/wasm_image.py
benchmark/benchmarks/matplotlib_benchmarks/wasm_image_affine.py
benchmark/benchmarks/matplotlib_benchmarks/wasm_rendering.py
benchmark/benchmarks/matplotlib_benchmarks/wasm_text_rotated.py
benchmark/benchmarks/numpy_benchmarks/allpairs_distances.py
benchmark/benchmarks/numpy_benchmarks/allpairs_distances_loops.py
benchmark/benchmarks/numpy_benchmarks/arc_distance.py
benchmark/benchmarks/numpy_benchmarks/check_mask.py
benchmark/benchmarks/numpy_benchmarks/create_grid.py
benchmark/benchmarks/numpy_benchmarks/cronbach.py
benchmark/benchmarks/numpy_benchmarks/diffusion.py
benchmark/benchmarks/numpy_benchmarks/evolve.py
benchmark/benchmarks/numpy_benchmarks/fdtd.py
benchmark/benchmarks/numpy_benchmarks/fft.py
benchmark/benchmarks/numpy_benchmarks/grayscott.py
benchmark/benchmarks/numpy_benchmarks/grouping.py
benchmark/benchmarks/numpy_benchmarks/growcut.py
benchmark/benchmarks/numpy_benchmarks/harris.py
benchmark/benchmarks/numpy_benchmarks/hasting.py
benchmark/benchmarks/numpy_benchmarks/hyantes.py
benchmark/benchmarks/numpy_benchmarks/julia.py
benchmark/benchmarks/numpy_benchmarks/l2norm.py
benchmark/benchmarks/numpy_benchmarks/large_decimal_list.py
benchmark/benchmarks/numpy_benchmarks/local_maxima.py
benchmark/benchmarks/numpy_benchmarks/log_likelihood.py
benchmark/benchmarks/numpy_benchmarks/lstsqr.py
benchmark/benchmarks/numpy_benchmarks/mandel.py
benchmark/benchmarks/numpy_benchmarks/multiple_sum.py
benchmark/benchmarks/numpy_benchmarks/pairwise_loop.py
benchmark/benchmarks/numpy_benchmarks/periodic_dist.py
benchmark/benchmarks/numpy_benchmarks/repeating.py
benchmark/benchmarks/numpy_benchmarks/reverse_cumsum.py
benchmark/benchmarks/numpy_benchmarks/rosen.py
benchmark/benchmarks/numpy_benchmarks/slowparts.py
benchmark/benchmarks/numpy_benchmarks/smoothing.py
benchmark/benchmarks/numpy_benchmarks/specialconvolve.py
benchmark/benchmarks/numpy_benchmarks/vibr_energy.py
benchmark/benchmarks/numpy_benchmarks/wave.py
benchmark/benchmarks/pystone_benchmarks/pystone.py
benchmark/plot_benchmark.py
benchmark/stack_usage.py
packages/fpcast-test/fpcast-test/setup.py
packages/micropip/src/micropip/externals/__init__.py
packages/micropip/src/micropip/externals/pip/__init__.py
packages/micropip/src/micropip/externals/pip/_internal/__init__.py
packages/micropip/src/micropip/externals/pip/_internal/utils/__init__.py
packages/micropip/src/micropip/externals/pip/_internal/utils/pkg_resources.py
packages/micropip/src/micropip/externals/pip/_internal/utils/wheel.py
packages/micropip/src/micropip/externals/pip/_vendor/__init__.py
packages/micropip/src/micropip/externals/pip/_vendor/pkg_resources.py
packages/micropip/src/setup.py
packages/nlopt/src/setup.py
packages/sharedlib-test-py/src/setup.py
pyodide-build/setup.py
src/py/setup.py
tools/pytest_wrapper.py
```

packages/micropip/src/micropip/_micropip.py
```diff
@@ -280,8 +280,8 @@ async def add_requirement(
                     f"but {req.name}=={ver} is already installed"
                 )
         metadata = await _get_pypi_json(req.name)
-        wheel, ver = self.find_wheel(metadata, req)
-        if wheel is None and ver is None:
+        maybe_wheel, maybe_ver = self.find_wheel(metadata, req)
+        if maybe_wheel is None or maybe_ver is None:
```
henryiii, line 284: This is a minor change (and to or) to make typing happy. I think they were always in sync.

packages/test_packages_common.py
hoodmane: Maybe this is dead code then?

henryiii: Probably?

henryiii:
Any idea why build-core failed? Doesn't immediately look related.

packages/test_packages_common.py
rth: Yes, it's never called you can remove it.

src/py/pyodide/_package_loader.py
```diff
@@ -54,9 +54,9 @@ def unpack_buffer(
     buffer: JsProxy,
     *,
     filename: str = "",
-    format: str = None,
-    target: Literal["site", "lib", None] = None,
-    extract_dir: str = None,
+    format: str | None = None,
```
rth, line 57: Actually the implicit optional is maybe not so bad.

pyproject.toml
```diff
@@ -1,8 +1,25 @@
...
+disallow_untyped_calls = false
+disallow_untyped_defs = false
+disallow_incomplete_defs = false
+disallow_untyped_decorators = false
+no_implicit_optional = true
+warn_redundant_casts = true
+warn_unused_ignores = true
+warn_return_any = false
+no_implicit_reexport = false
+strict_equality = false
```
rth, line 22: I'm not really sure about maintaining a custom set of mypy flags or figuring out what each of these do.  Using default flags, with maybe a strict mode would be preferable IMO.

I agree that mypy doesn't catch everything, but it will also not catch everything with a custom configuration.

Also it doesn't look like this required much changes in the code aside from the  `no_implicit_optional`

rth (commented):
A few comments otherwise LGTM.

rth:
Also would you mind removing the "Checklists" in the issue description if not relevant (as they are here)?

src/py/pyodide/_package_loader.py
henryiii: It's better to be explicit (plus it's required if we want to eventually go to strict mode).

pyproject.toml
henryiii: This is copied from strict=yrue; the idea is we can slowly work on making all these true. Once they are all true, we can replace all of them with strict=true.

By default, everything is false, so even just a single true in a custom configuration is better than nothing.

I intentionally only enabled things that were easy to enable in a first step (that is, didn't require lots of adding or fixing types).

henryiii: For better or worse, this is the design of mypy; everything off by default, and then turning one what you'd like to be strict about. `strict=true / --strict` was a concession, which is why it wasn't even available in the config mode for a long time (and still doesn't get handled properly in config mode if you try to override anything).

henryiii: If you prefer, I can replace the `= false` items with commented out `= true` items, which is the same thing, really.

src/py/pyodide/_package_loader.py
henryiii: In this case, for example, format can be None - either now or later. The type is `str | None` in either case - mypy is just smart enough to inject the `| None` for you. But other tools might not, a reader might not notice the `= None` part when reading the type, simply changing the default value changes the type, the type is incorrect if you try to copy it to a `.pyi` file (since the `= None` becomes `= ...`), etc. That's why implicit optionals were [removed from PEP 484](https://peps.python.org/pep-0484/#union-types). (There's a long discussion on it somewhere which I didn't bother to find the link for).

pyproject.toml
rth: > This is copied from strict=true; the idea is we can slowly work on making all these true.

OK, sounds good then, thanks!

src/py/pyodide/_package_loader.py
rth: OK, you convinced me that this is better :)

rth (approved):

Issue #2338, opened 2022-04-01
0.20.0a1: import scipy fails 


rth:
As reported by @mateuszcshark in 0.20.0a1 importing scipy fails,

```
>>> import scipy
Traceback (most recent call last):
  File "<console>", line 1, in <module>
  File "/lib/python3.10/site-packages/scipy/__init__.py", line 164, in <module>
    loadDynamicLibrary('/lib/python3.10/site-packages/scipy/linalg/_flapack.so')
JsException: Error: /lib/python3.10/site-packages/scipy/linalg/_flapack.so: file not found, and synchronous loading of external 
files is not available
```
I can also confirm with Firefox on main.

Likely related to latest changes in pre-loading so files cc @hoodmane . Though I don't understand why it worked in tests then.

ryanking13:
I can reproduce it with Chrome.

Interestingly, I have the same error when I test it in our Pyodide console. But it doesn't happen when I do the same thing in the browser devtools console.

```py
# Pyodide console
>>> import pyodide_js
>>> await pyodide_js.loadPackage("scipy")
>>> import scipy
# Exception
```

```js
// Browser devtools console
await pyodide.loadPackage("scipy")
pyodide.runPython("import scipy")
// works
```

rth:
Resolved in https://github.com/pyodide/pyodide/pull/2348

PR #2348, opened 2022-04-02
Fix scipy import failure #2338

hoodmane:
Trying to understand this bug made me realize that I am very deeply confused about a lot of stuff. But the actual cause of it is very simple: I patched this into `__init__.py`:
```py
if "pyodide_js" in sys.modules:
   from pyodide_js._module import loadDynamicLibrary
   loadDynamicLibrary('/lib/python3.10/site-packages/scipy/linalg/_flapack.so')
```
I uh... I have no idea what the conditional is supposed to mean. As far as I can tell, `if "pyodide_js" in sys.modules:` is just total nonsense. But it does explain the strange behavior of the bug. Anyways, I added a new patch with a detailed explanation.

I still don't really understand why importing scipy ever works. But this is okay, if the tests pass then who cares why it works right??

### DIFF
--- a/packages/scipy/patches/0001-loadDynamicLibrary-flapack.patch
+++ b/packages/scipy/patches/0001-loadDynamicLibrary-flapack.patch
@@ -1,6 +1,6 @@
-From 2b1c8ba0c5ee3acc2f9f221d517114c28b2c8bf9 Mon Sep 17 00:00:00 2001
+From 6e87b998a3c0073c507626a7c88963d9bda27c24 Mon Sep 17 00:00:00 2001
 From: Hood Chatham <roberthoodchatham@gmail.com>
-Date: Sat, 2 Apr 2022 11:13:37 -0700
+Date: Sat, 2 Apr 2022 14:20:25 -0700
 Subject: [PATCH] loadDynamicLibrary flapack
 
 We are using CLAPACK for our LAPACK, but CLAPACK only supports
@@ -26,14 +26,14 @@ look them up with `dlsym`). Scipy is not loaded this way. But we
 need the `lapack_extras.f` symbols to have this global visibility.
 So we have to call loadDynamicLibrary on it, hence this patch.
 ---
- scipy/__init__.py | 7 +++++++
- 1 file changed, 7 insertions(+)
+ scipy/__init__.py | 8 ++++++++
+ 1 file changed, 8 insertions(+)
 
 diff --git a/scipy/__init__.py b/scipy/__init__.py
-index 8c23069f9..4fa852612 100644
+index 8c23069f9..f408fc0b6 100644
 --- a/scipy/__init__.py
 +++ b/scipy/__init__.py
-@@ -158,3 +158,10 @@ else:
+@@ -158,3 +158,11 @@ else:
  
      # This makes "from scipy import fft" return scipy.fft, not np.fft
      del fft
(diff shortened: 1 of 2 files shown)


rth:
Thanks for the patch!

I don't understand why we need to load it explicitly as opposed to during import. I get there is some Fortran in there, but what makes it different from all the other .so files which as loaded with the global mechanism? Also normally `_flapack.so` is only loaded when importing `scipy.linlalg.lapack` and not the root scipy import.

hoodmane:
> Also normally `_flapack.so` is only loaded when importing `scipy.linlalg.lapack` and not the root scipy import.

Fair enough I moved the patch there.

hoodmane:
I believe there are three ways that symbols are loaded from shared libraries (I think this is true in both linux and emscripten):

1. `blah.so` is always loaded before `some.so`, and the loader looks up the symbols in `blah.so` at load time and writes the memory addresses of the `some.so` symbols directly into `blah.so`. On Wasm this is a bit different, but the loader locates those symbols in `blah.so` and puts them into the imports for the module as part of linking.
2. `some.so` file and `blah.so` are loaded in any order, `some.so` lazily loads the symbols it needs from the Global Object Table (GOT).
3. `some.so` uses `dlopen('blah.so')` and `dlsym`.

#2 is the only option that breaks cycles. Generally #1 and #2 are done automatically by the compiler, #3 requires breaking cycles. Different `.so` files can have symbols with the same name as long as they don't go into the GOT, only one symbol of each name can go in the GOT. 

Python loads packages using dlsym to load `PyInit_package_name` and then calls `PyInit_package_name` which inserts function pointers into a bunch of tables (effectively it using a fourth unrelated loaded system custom designed in Python).


Anyways, `loadDynamicLibrary` is somewhat poorly named. It doesn't load the library. What it does is it dumps the symbols from an already loaded library into the GOT. Scipy tries to load symbols from LAPACK via the GOT so we need LAPACK symbols to be in there. We have stuffed extra symbols from future versions of LAPACK into `_flapack.so`, and other Scipy libs try to look them up in the GOT so we need to get them in there.

ryanking13:

> `loadDynamicLibrary` is somewhat poorly named. It doesn't load the library. What it does is it dumps the symbols from an already loaded library into the GOT.

Huh, this is interesting...

packages/scipy/patches/0001-loadDynamicLibrary-flapack.patch
```diff
@@ -0,0 +1,54 @@
...
++++ b/scipy/linalg/lapack.py
+@@ -809,6 +809,15 @@ All functions
+    ilaver
+ 
+ """
++
++import sysconfig
++import pyodide_js
++ext_suffix = sysconfig.get_config_var("EXT_SUFFIX")
++pyodide_js._module.loadDynamicLibrary(f'/lib/python3.10/site-packages/scipy/linalg/_flapack{ext_suffix}')
```
ryanking13, line 44: For the future update:

```suggestion
+pyodide_js._module.loadDynamicLibrary(f'/lib/python{sys.version_info.major}.{sys.version_info.minor}/site-packages/scipy/linalg/_flapack{ext_suffix}')
```

hoodmane: Or maybe we can use:
```py
from site import getsitepackages
f"{getsitepackages}[0]}/scipy/linalg/_flapack{ext_suffix}"
```

ryanking13: Yes, that looks better.

PR #2350, opened 2022-04-03
Update libffi commit

hoodmane:
All libffi tests pass now. The only failing ctypes test is `test_callback_too_many_args` which doesn't segfault anymore, it only soft fails. Planning to submit a PR to cpython that fixes `test_callback_too_many_args` cf https://bugs.python.org/issue47208

- [x] Add / update tests


ryanking13 (approved):
Thanks, @hoodmane!

Any reason of pushing empty `cpython/make` file?

hoodmane:
> Any reason of pushing empty cpython/make file?

Yeah: `touch cpython/make<tab>` didn't autocomplete to Makefile because of lowercase `m`. Then I used `git commit -a`.

Issue #1977, opened 2021-11-18
RFC A better CLI for Pyodide


rth:
Following discussions in https://github.com/pyodide/pyodide/issues/655#issuecomment-972204766 and later comments, I would like to propose a new CLI API for Pyodide with a better user experience. Most Python and Javascript developers would probably rather avoid having anything to do with makefiles. 

The current proposal consists of the following steps.

## Create a new Pyodide CLI

It will be exposed as a new entry point under the `pyodide build package` with the following subcommand,

- **pyodide build**: would currently make the appropriate make call. It would work from any folder, assuming it's able to find the pyodide root directory (e.g. when passed via a CLI option)
    Examples:
    - `pyodide build numpy pandas` 
    - `pyodide build cpython`
    - `pyodide build emsdk`
 
- **pyodide clean**: cleaning the corresponding build artifacts
   Examples
   - `pyodide clean --all`
   - `pyodide clean cpython`
   - `pyodide clean numpy`: cleans only numpy artifacts
- **pyodide package new <package-name>**: create a new meta.yaml for a package. A replacement for `mkpkg` with I find not very expressive. Maybe there is a better name `pyodide recipe new` etc. Suggestions welcome.
- **pyodide package update <package-name>**:  update meta for one or multiple packages.  A replacement for `mkpkg --update`

This could be later extended with other subcommands, say for minifying pure python wheels, or downloading Pyodide packages locally.

This new CLI would be implemented with a modern library, for instance, [typer](https://typer.tiangolo.com/)

## Deprecate the current pyodide-build and make it private
 
The current [pyodide-build CLI](https://pyodide.org/en/stable/usage/api/pyodide-build-cli.html) would be deprecated, aside from the private commands we need (e.g. `pyodide-build buildpkg`), there isn't much useful there aside from,
 - `pyodide-build mkpkg`: which I imagine is not used that often (and certainly not in any production code), and it wouldn't be hard to switch
 - `pyodide-build server`: also a development-only tool. In most cases `pyodide -m http.server` should be sufficient.  Though maybe a command that starts a server with CORS enabled could be useful.

Alternatively, we could also try to keep better backward compatibility, by preserving the `pyodide-build` command name, but I think that `pyodide-build build`, `pyodide-build package new`, or even `pyodide-build download` (say for downloading packages locally) is rather confusing. 

## Implications

In the short term, this would mean harder to debug build failures since we would be wrapping make calls into another subprocess.

[...]

Issue #2020, opened 2021-12-05
Can't install pyodide-build using the instructions 


viktorChekhovoi:
## 🐛 Bug

The [pyodide documentation](https://pyodide.org/en/latest/development/new-packages.html#creating-the-meta-yaml-file) has instruction on how to install the tool pyodide-build. When I run the pip command, I get an error: [Errno 13] Permission denied: '/usr/local/lib/python3.9/site-packages/test-easy-install-24.write-test'

### To Reproduce

Clone the repository, run ./run_docker, then, in the docker, run the command 'pip install -e pyodide-build'

### Expected behavior

Pip should 'developer install' pyodide-build

### Environment

- Commit hash of Pyodide git repository: Commit 418813de331e8222e17b35f124043289bd6b0936
- Build environment

### Additional context

Not sure why this is happening, probably not my environment since I'm using Docker.

rth:
You don't necessarily need docker for `pyodide-build mkpkg`. I think using a virtualenv should work,
```
python -m venv venv
source venv/bin/activate
pip install -e requirements.txt
pip install -e pyodide-build
```

Otherwise yes, in docker the issue is that you are running it as non-root user and it wouldn't allow installing under `/usr/local`. So the solution is to either,
 - become root in the  container, with `./run_docker --root`, but then the produced artifacts will also be owned by root
 - or possibly install it as a user `pip install --user -e pyodide-build`

Neither of these is very satisfactory but it should work.

victorChekhovoy:
Thank you, I'll try running the command with the --root option.

ryanking13:
It seems like there is an issue on `editable install + user mode + pyproject.toml` (https://github.com/pypa/pip/issues/7953). I'll look into this in my possible times.

bmaranville:
It seems difficult to use a virtualenv to build after the change in #2351 - installing pyodide_build into the PYTHONUSERBASE directory means that the virtualenv can't find it.  If I remove the --prefix flag completely from the pip install invocation, it works fine, though I realize this won't necessarily work if you are using a system-installed python instead of a virtualenv.

ryanking13:
Hmm, I think the fundamental fix would be not using editable install because it keeps causing trouble. But that requires to do some non-trivial modification in our current logic of calling `pywasmcross.py` as a symlink during package build.

For a quick fix, I think simply checking env variable $VIRTUAL_ENV, then conditionally set --prefix would work.

ryanking13:
I think it is resolved now.

PR #2351, opened 2022-04-04
ENH Remove hard coded paths in `pyodide_build`

ryanking13:
### Description

We are pushing `pyodide_build` to PyPI as a Python package, but for now, installing `pyodide_build` from PyPI (i.e. pip install pyodide_build`) is almost useless because:

- there are bunch of hard-coded paths (e.g. `Path(__file__).parents[2]`),
- its dependencies are not specified in `setup.cfg`.

This PR is for mitigating this situation by removing hard-coded paths and adding tests, and is also a preparation for our new CLI (#1977).

### Checklists

- [x] Add / update tests


pyodide-build/pyodide_build/__main__.py
```diff
@@ -35,15 +35,12 @@ def make_parser() -> argparse.ArgumentParser:
 
 def main():
     if not os.environ.get("__LOADED_PYODIDE_ENV"):
-        from .common import get_hostsitepackages
-
-        PYODIDE_ROOT = str(pathlib.Path(__file__).parents[2].resolve())
-        os.environ["PYODIDE_ROOT"] = PYODIDE_ROOT
+        if "PYODIDE_ROOT" not in os.environ:
+            os.environ["PYODIDE_ROOT"] = str(pathlib.Path(os.getcwd()))
```
hoodmane, line 39: I think putting `Path(__file__).parents[2]` here would be better. Or throw an error here and require people to `source pyodide_env.sh` before running this command.

Otherwise this would break `pyodide-build buildpkg meta.yaml` in `packages/some_pkg` would break because `pyodide-build` would decide that `packages/some_pkg` is the root directory.

pyodide-build/pyodide_build/common.py
```diff
@@ -194,10 +152,10 @@ def get_make_environment_vars():
 
     This allows us to set all build vars in one place"""
     # TODO: make this not rely on paths outside of pyodide-build
-    rootdir = Path(__file__).parents[2].resolve()
+    PYODIDE_ROOT = Path(os.environ.get("PYODIDE_ROOT", os.getcwd()))
```
hoodmane, line 155: Similar, defaulting to cwd is no good.

pyodide-build/pyodide_build/mkpkg.py
```diff
@@ -89,23 +89,12 @@ def _get_metadata(package: str, version: str | None = None) -> dict:
...
-        ) from err
-    return YAML
-
-
 def run_prettier(meta_path):
     subprocess.run(["npx", "prettier", "-w", meta_path])
 
 
 def make_package(
+    root: Path,
```
hoodmane, line 97: Maybe better to pass in `packages_dir` here.

pyodide-build/pyodide_build/__main__.py
ryanking13: > Otherwise this would break `pyodide-build buildpkg meta.yaml` in `packages/some_pkg` would break because `pyodide-build` would decide that `packages/some_pkg` is the root directory.

Okay, I'll go with raising an error when `PYODIDE_ROOT` is not explicitly set for now.

My intention for the future is that we introduce a new config file similar to `packages.json` in node or `Cargo.toml` in rust which replaces the current `Makefile.env` and `pyodide_env.sh`, then our new CLI detects that file to ensure that we are in the right directory.

pyodide-build/pyodide_build/mkpkg.py
ryanking13: Yeah, that sounds more appropriate. Thanks!

pyodide-build/pyodide_build/__main__.py
hoodmane: Right. Until we get there I would prefer if try not to break the ways I am currently using these tools. Though of course it's fine if it is too hard to avoid.

packages/Makefile
```diff
@@ -6,20 +8,18 @@ else
...
-	PYTHONPATH="$(PYODIDE_ROOT)/pyodide-build/"\
-		python -m pyodide_build buildall . ../build \
-		$(ONLY_PACKAGES) --n-jobs $${PYODIDE_JOBS:-4} \
-		--log-dir=./build-logs
-
-.artifacts/bin/pyodide-build: ../pyodide-build/pyodide_build/**
+pyodide-build: ../pyodide-build/pyodide_build/**
 	mkdir -p $(HOSTINSTALLDIR)
-	$(HOSTPYTHON) -m pip install -e ../pyodide-build --no-deps --prefix $(HOSTINSTALLDIR)
+	$(HOSTPYTHON) -m pip install -q --user --upgrade --force-reinstall ../pyodide-build
```
ryanking13, line 18: We were using `pyodide-build` in a strange way during package build. We install it in `HOSTINSTALLDIR` but we don't use it. Instead, we just inject `pyodide-build` path to `PYTHONPATH`.

ryanking13:
Okay, this is ready to be reviewed.

After this PR, `pyodide-build` will less rely on its installation path. One can easily install it from PyPI, and run commands like `mkpkg` or run test. Note that it still has to read configs from `Makefile.envs` though.

The downside is that `PYODIDE_ROOT` needs to be always set to call `python -m pyodide_build`. This situation is not ideal. But currently, we don't have a way to detect that we are in a root directory of Pyodide. We can check the existence of `Makefile.env` or `pyodide_env.sh`, but that is just a temporary measure. This situation can be solved by introducing new CLI.

rth:
> The downside is that PYODIDE_ROOT needs to be always set to call python -m pyodide_build. This situation is not ideal. But currently, we don't have a way to detect that we are in a root directory of Pyodide. We can check the existence of Makefile.env or pyodide_env.sh, but that is just a temporary measure. This situation can be solved by introducing new CLI.

Yeah, I think when called from inside the Makefile setting PYODIDE_ROOT  is fine, however when used by end-users that's not ideal (also for running pytest, as a dev I would rather not have set it). So could we maybe do something like,
 - if PYODIDE_ROOT   is set use it.
 - otherwise, check that a `pyproject.toml` file exists in the current folder, and contains a line `[tool.pyproject]` (we can reasonably keep this in the future). If that's not found, error.  If there is a risk that it won't be in the current folder, maybe walk the folders up 2-3 levels or until `/` checking for the ``pyproject.toml` file. That's more or less what git does when trying to determine if a given folder belongs to a git repo.

WDYT?

packages/Makefile
rth: I think it would still be preferable to use editable install.

Also the `--user` option should almost always be avoided in favor of using a virtualenv, so IMO we should't silently install stuff un userdir.

pyodide-build/.gitignore
```diff
@@ -0,0 +1,11 @@
+*.pyc
```
rth, line 1: We already have a top level .gitignore which should take care of these?

pyodide-build/pyodide_build/pywasmcross.py
```diff
@@ -61,6 +59,7 @@ def make_command_wrapper_symlinks(env: MutableMapping[str, str]):
     exist.
     """
     exec_path = Path(__file__).resolve()
+    exec_path.chmod(0o755)
```
rth, line 62: It's weird to change the permission of the current file in place. It doesn't work without this? It was working before and fundamentally we are not changing anything with respect to calling this file via symlinks.

pyodide-build/pyodide_build/tests/_test_packages/beautifulsoup4/meta.yaml
```diff
@@ -0,0 +1,12 @@
+package:
```
rth, line 1: OK, yes it's probably cleaner to have standalone files for tests.

pyodide-build/requirements.txt
```diff
@@ -0,0 +1,9 @@
+build==0.7.0
```
rth, line 1: We don't need to have a requirements.txt file if they are already specified in setup.cfg IMO.
I'm not sure we really need to pin requirements at the moment. At the very least we could use semver compatible versions e.g. `build~=0.7.0` that would install 0.7.x but not 0.8.0.

rth (commented):
Thanks for this work @ryanking13 , it was a long-needed change!

Generally LGTM, aside from my previous comment about requiring `PYODIDE_ROOT` and a few minor comments below

pyodide-build/requirements.txt
hoodmane: Well we pretty aggressively use private features of `build`, basically in order to minimize the complexity of PR #2272. We probably should be pinning this exact version until we fix that.

rth: For build -- fine. I meant all the other packages there as well.

packages/Makefile
ryanking13: Alright, I'll find a way to use editable install and use it directly. Note that there is an issue in `pyproject.toml + user mode + editable install` https://github.com/pyodide/pyodide/issues/2020#issuecomment-1049430878

pyodide-build/pyodide_build/pywasmcross.py
ryanking13: This was added for non-editable installation. We don't need this if we will keep editable install.

pyodide-build/.gitignore
ryanking13: Right, good point.

pyodide-build/requirements.txt
ryanking13: I added this for testing in CI, but yeah, we probably don't need to pin versions (except for `build`).

ryanking13:
> * check that a `pyproject.toml` file exists in the current folder, and contains a line `[tool.pyodide]` (we can reasonably keep this in the future). If that's not found, error.  If there is a risk that it won't be in the current folder, maybe walk the folders up 2-3 levels or until `/` checking for the `pyproject.toml` file. That's more or less what git does when trying to determine if a given folder belongs to a git repo.

Yes, that sounds reasonable. Thanks for the suggestion!

rth (approved):
LGTM, thanks!

pyodide-build/pyodide_build/tests/test_common.py
```diff
@@ -113,3 +114,22 @@ def test_wheel_paths():
         "py3-none-any",
         "py2.py3-none-any",
     ]
+
+
+def test_search_pyodide_root():
```
rth, line 119: you can use the `tmp_path` fixture which directly returns a Path and auto-cleans at the end.

pyodide-build/pyodide_build/tests/test_common.py
```diff
@@ -113,3 +114,22 @@ def test_wheel_paths():
...
+        pyproject_file.write_text("[tool.pyodide]")
+        assert search_pyodide_root(temp_dir, 0) == temp_dir
+        assert search_pyodide_root(temp_dir / "subdir", 0) == temp_dir
+        assert search_pyodide_root(temp_dir / "subdir" / "subdir", 0) == temp_dir
+
+        pyproject_file.unlink()
+        try:
+            search_pyodide_root(temp_dir, 0)
+        except Exception as e:
+            assert isinstance(e, FileNotFoundError)
```
rth, line 135:
```suggestion
        with pytest.raises(FileNotFoundError):
            search_pyodide_root(temp_dir, 0)
```

pyodide-build/pyodide_build/common.py
```diff
@@ -209,3 +171,30 @@ def get_make_environment_vars():
...
+    by looking for the pyproject.toml file in the parent directories
+    which contains [tool.pyodide] section.
+    """
+
+    if depth >= max_depth:
+        raise FileNotFoundError(
+            "Could not find Pyodide root directory. If you are not in the Pyodide directory, set `PYODIDE_ROOT=<pyodide-root-directory>`."
+        )
+
+    base = Path(base)
```
rth, line 188: Wait you need to use depth somewhere. And also check that we are not walking higher than '/'. Something like,
```py
    # at the top
    base = Path(base)
    max_depth = min(max_depth, len(base.parents))
	if depth >= max_depth:
    ...
    
    current_dir = base.parents[depth]
    pyproject_file = current_dir / "pyproject.toml"
```
?

pyodide-build/pyodide_build/common.py
```diff
@@ -193,11 +152,14 @@ def get_make_environment_vars():
     """Load environment variables from Makefile.envs
 
     This allows us to set all build vars in one place"""
-    # TODO: make this not rely on paths outside of pyodide-build
-    rootdir = Path(__file__).parents[2].resolve()
+
+    PYODIDE_ROOT = os.environ.get("PYODIDE_ROOT")
+    if PYODIDE_ROOT is None:
+        PYODIDE_ROOT = str(search_pyodide_root(os.getcwd(), 0))
```
hoodmane, line 158:
```suggestion
        PYODIDE_ROOT = search_pyodide_root(os.getcwd(), 0)
```

pyodide-build/pyodide_build/common.py
```diff
@@ -193,11 +152,14 @@ def get_make_environment_vars():
...
+
     environment = {}
     result = subprocess.run(
-        ["make", "-f", str(rootdir / "Makefile.envs"), ".output_vars"],
+        ["make", "-f", str(Path(PYODIDE_ROOT) / "Makefile.envs"), ".output_vars"],
```
hoodmane, line 162:
```suggestion
        ["make", "-f", str(PYODIDE_ROOT / "Makefile.envs"), ".output_vars"],
```

pyodide-build/pyodide_build/common.py
```diff
@@ -209,3 +171,30 @@ def get_make_environment_vars():
             value = value.strip("'").strip()
             environment[varname] = value
     return environment
+
+
+def search_pyodide_root(base: str | Path, depth: int, max_depth: int = 5) -> Path:
```
hoodmane, line 176:
```suggestion
def search_pyodide_root(base: str | Path, depth: int = 0, max_depth: int = 5) -> Path:
```

pyodide-build/pyodide_build/common.py
hoodmane: I guess `Path("/").parent == Path("/")` so not checking that case isn't **wrong** persay it's just wasteful.

pyodide-build/pyodide_build/common.py
hoodmane: Probably for the type system it is convenient to use a helper function for the recursion...

pyodide-build/pyodide_build/common.py
```diff
@@ -209,3 +171,30 @@ def get_make_environment_vars():
...
+    """
+    Recursively search for the root of the Pyodide repository,
+    by looking for the pyproject.toml file in the parent directories
+    which contains [tool.pyodide] section.
+    """
+
+    if depth >= max_depth:
```
hoodmane, line 183:
```suggestion
    if depth >= max_depth or base.is_mount():
```

hoodmane:
I'm just going to make these final changes since I want to make the 0.20 release today.

hoodmane:
Thanks @ryanking13!

ryanking13:
Thank you for the reviews and changes!

PR #2355, opened 2022-04-04
Run make -C packages update-all

hoodmane:


ryanking13 (approved):
Thanks, @hoodmane!

BTW, we are not tracking package updates in our changelog recently. Should we check and record them before the release? Or we could simply not record package updates in our changelog from now on as we have [the list of packages in our documentation](https://pyodide.org/en/stable/usage/packages-in-pyodide.html).

hoodmane:
> BTW, we are not tracking package updates in our changelog recently. Should we check and record them before the release? 

I don't really want to do it. If we are going to try, maybe we ought to make a tool...

PR #2356, opened 2022-04-05
CI Disable log clipping in building docker images on Github Actions

ryanking13:
Disable log clipping behavior of docker buildkit.
This will help us analyze the log when `prebuilt_docker_image` action fails.


rth (approved):
Thanks a lot @ryanking13 !

PR #2358, opened 2022-04-07
DOC Remove dummy packages from docs

ryanking13:
Drop "cpp-exceptions-test", "fpcast-test", "sharedlib-test", "sharedlib-test-py" [from docs](https://pyodide.org/en/latest/usage/packages-in-pyodide.html).


docs/sphinx_pyodide/sphinx_pyodide/packages.py
```diff
@@ -11,6 +11,13 @@
 
 from pyodide_build.io import parse_package_config
 
+DUMMY_PACKAGES = (
+    "cpp-exceptions-test",
```
hoodmane, line 15: Maybe better to add a setting to `meta.yaml` to exclude a package from the docs?

ryanking13: That's one other option, but I don't like adding a non-build-related section to `meta.yaml`.

hoodmane: Maybe just under `package`? We could even use the version field:
```yaml
package:
  name: test-blah
  version: pyodide-test
```
We might also consider putting `version: python-stdlib` and `version: python-test` for the stdlib packages / test packages since none of these have real versions... Or better, leave the version out and put these as some sort of marker instead of a version...

packages/cpp-exceptions-test/meta.yaml
```diff
@@ -1,6 +1,6 @@
 package:
   name: cpp-exceptions-test
-  version: "0.1"
+  version: pyodide.test
```
hoodmane, line 3: Yeah this looks nice to me -- even if it is a slight bit odd I think it makes sense.

ryanking13: Yeah, thanks for the suggestion. The best way would be removing all those testing packages, but there is no trivial way to do that I guess?

hoodmane: > The best way would be removing all those testing packages, but there is no trivial way to do that I guess?

Well the package loading logic is quite subtle, and many of the behaviors covered by these tests cover bugs that normally only occur in scipy. Also, note for instance the value of `test_cpp_exceptions` in this bug: https://github.com/emscripten-core/emscripten/issues/16538. On the whole, I think they are one of the most important value parts of the test suite.

hoodmane: It's sort of similar to the reason that the Python interpreter comes with ~6 testing only extension modules.

ryanking13: Thanks for the explanation. Yes I also think those tests are really valuable.

rth: I see the intention, but I think it would be better to add a separate key for this, maybe,
```
package:
    tag: pyodide.test
```
rather than putting this information into the version which is supposed to verify PEP-386 version conventions https://docs.conda.io/projects/conda-build/en/latest/resources/define-metadata.html#package-version We are not checking that now, but we could very well do it in the future.

ryanking13: Alright, then I'll go with adding a separate key. What do you think of keeping the new key private, so that we can modify or remove it easily when we find a better way to specify those special packages?

ryanking13:
I added a new `tag` entry at the `meta.yaml` as @rth suggested, instead of using the `version` string.
I didn't add the `tag` entry into our meta.yaml specification as, for now, it is only related to building docs and has nothing to do with building packages.

hoodmane:
Perhaps we could call it `_tag` to indicate privateness?

ryanking13:
> Perhaps we could call it `_tag` to indicate privateness?

Done!

hoodmane:
Thanks @ryanking13! And sorry about all the bike shedding here =P

ryanking13:
> Thanks @ryanking13! And sorry about all the bike shedding here =P

Naming is very important :)

PR #2361, opened 2022-04-07
Update scipy to v1.8.0

hoodmane:
I also fixed mvnun and mvnun_weighted with a bit of a hack: f2c doesn't support variable length arrays as local variables, but we can substitute in some characteristic constant for the array length, and then find it in the f2c'd output and replace it back. We also have to remove the static storage classifier since it doesn't make sense to try to declare a static but variable length array. This same treatment could be applied to get cuncsd, duncsd, sorcsd, and zorcsd working but I haven't bothered here.

The PROPACK library also has a shared `timing` struct which needs to be marked `extern`.

I opened two scipy PRs to fix some problems I ran into here:
https://github.com/scipy/scipy/pull/15954
https://github.com/scipy/scipy/pull/15955


hoodmane:
So it turns out that the problem with `cuncsd` et al was just that `f2c` can't handle the `RECURSIVE` keyword. It is completely unbothered by functions that make recursive calls, but marking them as `RECURSIVE` causes all sorts of error messages.

hoodmane:
So it turns out that the problem with `cuncsd` et al was just that `f2c` can't handle the `RECURSIVE` keyword. It is completely unbothered by functions that make recursive calls, but marking them as `RECURSIVE` causes all sorts of error messages.

hoodmane:
Could someone review this so we can merge it? Then we're up to date on all packages! (Excepting the cryptography rust caveat.)

packages/scipy/patches/0002-Fix-getbreak-dlamch.patch
```diff
@@ -0,0 +1,39 @@
...
+diff --git a/scipy/sparse/linalg/_isolve/iterative/GMRESREVCOM.f.src b/scipy/sparse/linalg/_isolve/iterative/GMRESREVCOM.f.src
+index da4bde8ec..5f47afde2 100644
+--- a/scipy/sparse/linalg/_isolve/iterative/GMRESREVCOM.f.src
++++ b/scipy/sparse/linalg/_isolve/iterative/GMRESREVCOM.f.src
+@@ -154,7 +154,7 @@
+       INFO = 0
+       MAXIT = ITER
+       BRKDWN = .FALSE.
+-      EPS = <sdsd>LAMCH('EPS')
++      EPS = <sdsd>LAMCH(69)
```
rth, line 20: This looks strange to me. Why 69? LAMCH('EPS') is supposed to return machine precision if I understand correctly:
 - http://netlib.org/lapack/explore-html-3.3.0-2011-01-25/dlamch_8f_source.html
 - http://www.icl.utk.edu/~luszczek/lapack/lamch.html

pyodide-build/pyodide_build/_f2c_fixes.py
```diff
@@ -226,6 +229,45 @@ def fix_f2c_output(f2c_output_path: str):
             for line in lines
         ]
 
+    # Substitute back the dummy fixed array sizes. We also have to remove the
+    # "static" storage specifier since variable sized arrays can't have static
+    # storage.
```
rth, line 234: I can't wait for lfortran to have a good compiler with WASM support !

Thanks for finding this workaround!

pyodide-build/pyodide_build/pywasmcross.py
```diff
@@ -147,13 +147,24 @@ def replay_f2c(args: list[str], dryrun: bool = False) -> list[str] | None:
...
-            filename = os.path.abspath(arg)
+        if arg.endswith(".f") or arg.endswith(".F"):
+            filepath = Path(arg).resolve()
             if not dryrun:
                 fix_f2c_input(arg)
-                subprocess.check_call(
-                    ["f2c", os.path.basename(filename)], cwd=os.path.dirname(filename)
-                )
+                if arg.endswith(".F"):
+                    subprocess.check_call(
```
rth, line 155: So the standard way of not actually running the command doesn't work for .f. Might be worth adding a comment.

rth (commented):
Thanks for all your work @hoodmane !

packages/scipy/patches/0002-Fix-getbreak-dlamch.patch
hoodmane: 69 is the ascii code for "E".

pyodide-build/pyodide_build/pywasmcross.py
hoodmane: Yeah apparently `.F` commands expect first to be run through the C preprocessor (they have #ifdef's in them)

packages/scipy/patches/0002-Fix-getbreak-dlamch.patch
rth: OK, that makes sense thanks. Maybe worth adding in the patch comment.

hoodmane: But I think this patch isn't needed it should get handled by `char1_args_to_int`.

PR #2368, opened 2022-04-08
TST Enable some skipped or xfailed tests

ryanking13:
### Description

- Pillow - mistakenly skipped after #2278
- Jjinja2 - skipped due to wrong logic in `_package_is_built`.
- statsmodels - xfailed due to browser issue in older version of Chrome


rth (approved):
Thanks for catching this @ryanking13 !

PR #2372, opened 2022-04-08
Update cryptography to v3.4.8

hoodmane:
As recommended here: https://github.com/pyca/cryptography/issues/7051
we can build v3.4.8 with `export CRYPTOGRAPHY_DONT_BUILD_RUST=1`.

### DIFF
--- a/packages/cryptography/meta.yaml
+++ b/packages/cryptography/meta.yaml
@@ -1,16 +1,18 @@
 package:
   name: cryptography
-  version: 3.3.2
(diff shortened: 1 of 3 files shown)


rth:
Thanks! It looks like 3.4.8 is the [latest version for which this would work](https://github.com/pyca/cryptography/blob/7659fe4fd65e0aba35e69e71ab87c6c6eb3ece66/CHANGELOG.rst#3500---2021-09-29) unfortunately

PR #2377, opened 2022-04-09
xfail statsmodels in chrome

hoodmane:


ryanking13:
Thanks, @hoodmane. I thought this was an issue in older versions of Chrome, but it still happens. I should've double-checked...

hoodmane:
Yeah it's fine these configuration things require a certain amount of trial and error.

PR #2378, opened 2022-04-09
Cryptography v36.0.2 w/ Rust

### LINKED ISSUE #1973: Building a Python package that uses Rust

hoodmane:
Resolves #1973.


hoodmane:
@henryiii  the code spell is trying to correct `crate ==> create` again

hoodmane:
This is currently missing a fix for https://github.com/chronotope/chrono/issues/519, but once we patch `chrono` it will work.

henryiii:
Again? We can add `crate` to the wordlist.

henryiii:
https://github.com/hoodmane/pyodide/blob/4e476ff33bd20ee00fbf011a9c73ae3bfcf0538e/.pre-commit-config.yaml#L102

Needs to be added here.

ryanking13:
Thanks for working on this @hoodmane! It looks really cool that rust modules are working. Maybe we should discuss adding rust for the mandatory build dependency. I guess @rth was negative on that (https://github.com/pyodide/pyodide/pull/2081#issuecomment-1016407193).

hoodmane:
Well @rth *might* change his mind for this -- a fair number of people have requested cryptography support so I think it's a more compelling use case than a test package.

src/js/pyodide.ts
```diff
@@ -280,6 +280,8 @@ export async function loadPyodide(
   Module.locateFile = (path: string) => {
     throw new Error("Didn't expect to load any more file_packager files!");
   };
+  Module["__emscripten_get_now"] = Module["_emscripten_get_now"];
```
rth, line 283: Could you add a comment about these lines? I imagine something from C calls this JS function? What's going on with the `sig` though?

emsdk/patches/0001-Fix-dyncall-and-dynamic-linking.patch
```diff
@@ -0,0 +1,49 @@
+From 7c7a8f37b79e357cc28cb44f99258c9841d7385a Mon Sep 17 00:00:00 2001
+From: Hood Chatham <roberthoodchatham@gmail.com>
+Date: Fri, 8 Apr 2022 15:54:49 -0700
+Subject: [PATCH] Fix dyncall and dynamic linking
+
+We don't compile with WASM_BIGINT so if there is a 64 bit integer
+in the dynamic call, Emscripten tries to call a legalizer wrapper.
+These legalizer wrappers are only generated in the main module so
+sometimes a side module tries to call a wrapper that doesn't exist.
+This adds an extra handler in that case.
```
rth, line 10: Please link to https://github.com/emscripten-core/emscripten/pull/16693 

Thanks for contributing this to Emscripten! It might be good to wait a bit until they are happy with the PR there before merging this.

packages/cryptography/meta.yaml
```diff
@@ -1,18 +1,32 @@
...
+    export CARGO_BUILD_TARGET=wasm32-unknown-emscripten
+    export PYO3_CONFIG_FILE=$PKG_DIR/pyo3_config.ini
+    CARGO_TARGET_WASM32_UNKNOWN_EMSCRIPTEN_LINKER="emcc"
+    export EMSCRIPTEN_LINKING_RUST=1
+    export RUSTFLAGS=" \
+      -C relocation-model=pic \
+      -C target-feature=+mutable-globals \
+      -C link-arg=-sSIDE_MODULE=1 \
+    "
+    git clone --depth 1 https://github.com/hoodmane/chrono.git --branch pyodide
```
rth, line 24: Maybe we could add a TODO comment to remove this one https://github.com/chronotope/chrono/pull/593 is released.

packages/cryptography/meta.yaml
```diff
@@ -1,18 +1,32 @@
...
-  sha256: 94cc5ed4ceaefcbe5bf38c8fba6a21fc1d365bb8fb826ea1688e3370b2e24a1c
+  url: https://files.pythonhosted.org/packages/10/a7/51953e73828deef2b58ba1604de9167843ee9cd4185d8aaffcb45dd1932d/cryptography-36.0.2.tar.gz
+  sha256: 70f8f4f7bb2ac9f340655cbac89d68c527af5bb4387522a8413e841e3e6628c9
+  patches:
+    - patches/0001-Use-patched-chrono.patch
 build:
+  script: |
+    source $HOME/.cargo/env
+    rustup update nightly
+    rustup component add rust-src --toolchain nightly-x86_64-unknown-linux-gnu
```
rth, line 14: Maybe the above two lines should go after Rust installation.

The rest of flags might be better in Makefile.env or some global config specific to the build system.

packages/cryptography/meta.yaml
```diff
@@ -1,18 +1,32 @@
...
-  url: https://files.pythonhosted.org/packages/cc/98/8a258ab4787e6f835d350639792527d2eb7946ff9fc0caca9c3f4cf5dcfe/cryptography-3.4.8.tar.gz
-  sha256: 94cc5ed4ceaefcbe5bf38c8fba6a21fc1d365bb8fb826ea1688e3370b2e24a1c
+  url: https://files.pythonhosted.org/packages/10/a7/51953e73828deef2b58ba1604de9167843ee9cd4185d8aaffcb45dd1932d/cryptography-36.0.2.tar.gz
+  sha256: 70f8f4f7bb2ac9f340655cbac89d68c527af5bb4387522a8413e841e3e6628c9
+  patches:
+    - patches/0001-Use-patched-chrono.patch
 build:
+  script: |
+    source $HOME/.cargo/env
+    rustup update nightly
```
rth, line 13: Could you add an inline comment which feature require nightly?

packages/cryptography/test_cryptography.py
```diff
@@ -18,149 +18,6 @@ def test_cryptography():
     assert f1.decrypt(f.encrypt(b"abc")) == b"abc"
 
 
-@run_in_pyodide(packages=["cryptography", "pytest"])
```
rth, line 21: These are no longer relevant or do they fail?

pyodide-build/pyodide_build/pywasmcross.py
```diff
@@ -528,9 +534,17 @@ def handle_command(
...
+
     if returncode != 0:
         sys.exit(returncode)
 
+    # Rust gives output files a `.wasm` suffix, but we need them to have a `.so`
+    # suffix.
+    if line[0:2] == ["cargo", "rustc"]:
+        p = Path(args.builddir)
+        for x in p.glob("**/*.wasm"):
+            shutil.copy(x, x.with_suffix(".so"))
```
rth, line 546: Copy or rename? Is there a reason to keep the original ones? Also so they are not included in the final wheel.

rth (commented):
It's great that you managed to make this work!

By Rust  not being a hard dependency, I meant that people shouldn't have to install Rust to build an arbitrary Python package with Pyodide. Of course, if they want to build a Rust+PyO3 package (and there is some demand for it) they will need it.

So I think the current situation where CircleCI builds with Rust installed (we could actually add it to the docker image) while GHA builds the core packages without it, sounds very reasonable to me.

Generally LGTM, we would probably need, 
 - changelog entry + some documentation about ability to build Rust+PyO3 packages
 - maybe pin the nightly to a given date, to have the environment a bit more controlled

packages/cryptography/test_cryptography.py
hoodmane: I think these test features that were removed between 3.4.8 and 36.0.2, so they fail with name errors. Maybe we should ask the cryptography maintainers for more info on what to test.

pyodide-build/pyodide_build/pywasmcross.py
hoodmane: Yeah clearly move is the right approach.

src/js/pyodide.ts
hoodmane: "v" means that the function takes 0 arguments and returns void (first letter is return type, subsequent letters are argument types). The signature is consumed here:

https://github.com/emscripten-core/emscripten/blob/7e538a419c3649f3a540a57beab347aa8f6c6271/src/runtime_functions.js#L118

Any emscripten function to be called from C needs one. Probably this is an upstream bug.

emsdk/patches/0001-Fix-dyncall-and-dynamic-linking.patch
hoodmane: They suggested we should use `-sWASM_BIGINT` and shim `BigInt64Array` and `BigUint64Array` for Safari versions that don't have them. I think we can start out by using this patch and work on switching? Switching to `-sWASM_BIGINT` would be generally pretty good for us but it might take some effort.

hoodmane: Opened #2643 to try to switch to the strategy recommended upstream.

src/js/pyodide.ts
hoodmane: Hmm this doesn't seem to be necessary anymore?

emsdk/patches/0001-Fix-dyncall-and-dynamic-linking.patch
hoodmane: Hmm this patch also seems like it might not be necessary -- it was needed for `pyo3_samples` to work but it is unclear that it is needed for cryptography.

src/js/pyodide.ts
hoodmane: It does still seem necessary.

hoodmane: Opened an issue here:
https://github.com/emscripten-core/emscripten/issues/17106

hoodmane: Okay but "v" is also the wrong signature: the function takes zero arguments and returns a double. So if this function were ever actually called, we would get a crash.

.circleci/config.yml
```diff
@@ -129,6 +129,20 @@ jobs:
...
+            wget https://sh.rustup.rs -O /rustup.sh
+            sh /rustup.sh -y
+            source $HOME/.cargo/env
+            # We need nightly because we need to build the standard library.
+            # -Zbuild-std only works on nightly:
+            # https://doc.rust-lang.org/cargo/reference/unstable.html#requirements
+
+            # Make sure nightly-2022-05-28 matches the version in Makefile.envs
+            rustup toolchain install nightly-2022-05-28
+            rustup component add rust-src --toolchain nightly-2022-05-28
```
ryanking13, line 144: I think we will have to add this rust installation step in our Makeflie (e.g. `make rust`) so that people can build the cryptography package locally.

hoodmane:
Okay I think this PR is in good shape now:

1. I removed the need for `-Zbuild-std` so now we can build with rust stable
2. Ideally `RUSTFLAGS` would use `SIDE_MODULE_LDFLAGS` but I can't get it to work
3. The `--whole-archive` problem is recognized upstream, seems like it's just a matter of time before we work out an appropriate upstream fix. https://github.com/emscripten-core/emscripten/issues/17109
4. The `emscripten_get_now__sig` patch is merged upstream  https://github.com/emscripten-core/emscripten/pull/17123/files
5. The `chrono` bug is fixed upstream, but we need them to make a release and cryptography to update the version before we can use it.
6. The `instant` library looks unmaintained but I opened a PR https://github.com/sebcrozet/instant/pull/47
7. It would be fantastic to make rust emit `.so` files rather than emitting `.wasm` files and having to move them in pywasmcross, but the `wasm32-unknown-emscripten` target has `"dll-suffix": ".wasm"` and as far as I can tell this is not configurable without creating a custom target which would require us to maintain complete target metadata and would probably force us back to nightly rust.

hoodmane:
So there is still a lot that could be improved but given the state of the `wasm32-emscripten` rust target, I would say we are doing pretty good here.

hoodmane:
Probably we should add Rust to the docker image?

ryanking13:
> Probably we should add Rust to the docker image?

Yes I am +1 for it.

Dockerfile
```diff
@@ -94,5 +94,23 @@ RUN npm install -g \
...
+    RUST_VERSION=1.61.0
+
+RUN rustArch='x86_64-unknown-linux-gnu'; rustupSha256='3dc5ef50861ee18657f9db2eeb7392f9c2a6c95c90ab41e45ab4ca71476b4338' ; \
+    url="https://static.rust-lang.org/rustup/archive/1.24.3/${rustArch}/rustup-init"; \
+    wget "$url"; \
+    echo "${rustupSha256} *rustup-init" | sha256sum -c -; \
+    chmod +x rustup-init; \
+    ./rustup-init -y --no-modify-path --profile minimal --default-toolchain $RUST_VERSION --default-host ${rustArch}; \
+    rm rustup-init; \
+    chmod -R a+w $RUSTUP_HOME $CARGO_HOME; \
```
rth, line 109: Can't we use rustup directly which does most of the above? Also the advantage is that then we can use the same command in the docs.

rth (commented):
It would be good to update the documentation, otherwise LGTM as well.

hoodmane:
I can't get the docker image to work correctly so I reverted that stuff. I think we can deal with that as a followup.

ryanking13:
Sure, feel free to merge it when you think it is ready. Thanks for your hard work!

PR #2381, opened 2022-04-09
Changes for v0.20 release

hoodmane:


docs/project/roadmap.md
```diff
@@ -63,17 +63,6 @@ resolution.
 
 See issues {issue}`2045` and {issue}`1100`.
 
-## Switch to using wheels for Python packages
```
rth, line 66: I think we could keep these two, maybe just adjust the wording. 

Wheels work for us, but there is still an effort needed to get them accepted by PyPi, and there is likely standardization effort needed (including for backward compatibility policy)

For Rust, I see you are working on building cryptography, I imagine there are always things that could be improved size wise etc?

rth (approved):
Thanks!

docs/project/roadmap.md
hoodmane: The size is actually fantastic: it's about 3x smaller than the corresponding package for native platform. And it really works: I assume Cryptography uses a wide range of features and their test suite passes thousands of checks up until chrome crashes.

hoodmane: So I could see modifying the wheels one to discuss some new issues. For Rust, we could update it to say address some specific issues. But if we find out issues with Rust, we could always put it back in. I figure it's good to update the roadmap frequently or else it doesn't mean much.

rth: Great results for Cryptography!

OK, up to you -- we can always edit this later if needed.

hoodmane: Yeah I'll just leave it out for now. Ideally I think we should focus a lot on the docs and examples for the next few months, hopefully as a part of that we can update the roadmap a bit.

PR #2383, opened 2022-04-09
Replace runtime assertions about JsRef size and alignment with compile time _Static_asserts

hoodmane:


rth (approved):
Sounds good, thanks!

PR #2388, opened 2022-04-11
Update docker image

ryanking13:
This applies `requirements.txt` file update in the root directory to the docker image from #2351.


hoodmane:
Do we still install the stuff in `install_requires` in the docker image? It could be a bit inconvenient for people using `./run_docker` if it doesn't come with these installed. Though `pip` installs into `.docker_home` so maybe we can just tell people they need to run some command the first time they use the docker image? If so we should update the docs.

ryanking13:
Yes, I also think the docker image should better not contain Python packages.

Actually, our current `requirements.txt` file in the root directory contains packages only for testing and linting. All package build dependencies are moved into `pyodide-build` through #2351 and now they are installed during `make`. So those build dependencies are going to be removed from the docker image in this PR.

I am planning to gradually move the testing dependencies into the new CLI (or into the `pyodide-build`).

PR #2391, opened 2022-04-11
Allow multiple loading of Pyodide instances

### LINKED ISSUE #2255: Unable to retry loading pyodide, get error "Pyodide is already loading"
## 🐛 Bug

If `loadPyodide` fails, I'd like to call it again in case it's a transient network error. But trying this throws an error `Pyodide is already loading.`

### To Reproduce

I'm roughly doing this:

```js
import pRetry from 'p-retry';

pRetry(() => loadPyodide({indexURL: 'https://cdn.jsdelivr.net/pyodide/v0.19.0/full/'}), {retries: 3})
```

When I kill the network in the middle of loading, I get:

```
pyodide.js:243 Uncaught (in promise) Error: Pyodide is already loading.
    at loadPyodide (pyodide.js:243:1)
    at loadPyodideOnly (Worker.js:25:1)
    at RetryOperation._fn (index.js:50:1)
    at retry_operation.js:85:1
```

### Expected behavior

When `loadPyodide` fails, clean up and delete anything created during loading such as `globalThis.__pyodide_module`. This seems important for other reasons like memory usage. Hopefully it should allow `loadPyodide` to be called again. Or is there some deeper emscripten state problem with trying to load pyodide twice?

Adding some internal retrying logic to the various network requests would also be great, since I'm sure retrying all errors on my side isn't ideal. But I can understand if that raises tricky questions.

hoodmane:
This resolves #2255
@leopsidom this is much simpler than what I was considering before =)


hoodmane:
Yeah a few benefits:
1. If a network request fails, can retry `loadPyodide` (there are other ways to do this though like wrapping network requests in some retrying system)
2. After a fatal error, can reload a new Pyodide instance
3. Could be useful in testing
4. ??? other things?

It's also nice from an aestethic perspective, and I don't think this changeset introduces too much additional complexity.

src/js/rollup.config.js
```diff
@@ -36,8 +33,23 @@ function config({ input, format, minify, ext }) {
...
+  {
+    input: "pyodide.umd",
+    output: "dist/pyodide.js",
+    format: "umd",
+    name: "loadPyodide",
+    minify: true,
+  },
+  {
+    input: "api",
+    output: "src/js/_pyodide.out.js",
```
ryanking13, line 51: Coud you explain the necessity of creating a API as a seperate js file then loading it with `--pre-js`?

ryanking13 (commented):
Thanks, @hoodmane. It would be great if we can load multiple instances of Pyodide.
For example, on an online IDE, one can run multiple Python commands in multiple shells.

rth:
Thanks for making this possible! I agree it could be useful, and probably better design-wise anyway. Code wise LGTM, as well, but CI is still failing.

> on an online IDE, one can run multiple Python commands in multiple shells.

Right but in that use case, as long as we are single-threaded, I think a single instance with separate global dicts would be faster ?
So maybe we should mention that feature in this release note so that people don't start creating new pyodide instances each time they want to reset the globals dict :)

hoodmane:
> people don't start creating new pyodide instance each time they want to reset the globals dict

"We won't stop you from starting a new instance for each namespace you need, but it isn't a good idea. Only use if you need to reset the entire global state of Python, for instance in case of a fatal error."

src/js/rollup.config.js
hoodmane: Yeah, the point is that we have top-level state in these files that we want to be scoped to particular Pyodide instances rather than global. The trick is that stuff that we include in `--pre-js` has top level state instantiated once every time we call `_createPyodideModule`, whereas if we load the rolled up file directly then the state is instantiated once period (hence is global).

alexmojaki:
Thanks for solving this! Looking forward to being able to use it to retry loading.

> After a fatal error, can reload a new Pyodide instance

I have some questions about how this would work in practice.

- Would it be possible for Pyodide to automatically reload itself after a fatal error?
- If not, is there an easy way to know when a fatal error has happened and request a reload?
- Would the reload be faster than loading from scratch?
- Does the broken Pyodide instance continue using memory? Would it be better to just terminate the worker and start again from a clean slate?

hoodmane:
> Would it be possible for Pyodide to automatically reload itself after a fatal error?

Maybe, but I think it would be better for us not to do this so that downstream can decide what to do.

> If not, is there an easy way to know when a fatal error has happened and request a reload?

Well we could set a flag on the `pyodide` object and/or also add an `onFatal` handler. We use an internal `onFatal` handler for testing.

> Would the reload be faster than loading from scratch?

Probably because of browser caching? Someone will have to test to know for sure. I also wonder if you load another copy of the same Javascript/wasm code into the same worker whether there are benefits from the jit already being warmed up.

> Does the broken Pyodide instance continue using memory? Would it be better to just terminate the worker and start again from a clean slate?

Someone will have to test to know for sure. I think if you throw away all references to it, it should be garbage collected. Maybe we can throw away the references to `pyodide._module` and `pyodide._api`. Also every `PyProxy` carries a reference into the Pyodide instance, though if we use [`Proxy.revocable`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Proxy/revocable) and track the proxies, we could revoke all of them when the fatal error occurs. If this all succeeds I think everything could be gc'd.

Also, currently if you try to import old PyProxies into the new Pyodide runtime you will probably cause segfaults. So more work is needed here for sure. We would certainly appreciate it if you investigate any of these things or provide any further feedback on limitations!

src/js/rollup.config.js
rth: This entry raises a warning BTW,
```
(!) If you do not supply "output.name", you may not be able to access the exports of an IIFE bundle.
```
not sure if it's relevant

Issue #2393, opened 2022-04-11
npm package should not include .ts files


mneil:
## 🐛 Bug

I just did an `npm i pyodide` wanting to use this as [part of a projec](https://github.com/3p3r/cdk-web) for packaging AWS Lambda packages to support the Python Lambda Functions. I've been testing out pyodide and brython to see which one makes more sense. I am really liking pyodide. Up to now I've been including the file from a cdn and it's gone fine. However, the package on npm includes all of the typescript files.

An npm package should include type definitions (.d.ts) files for typescript and javascript users for type definitions. The rest of the code, however, should all be javascript. When you publish .ts files I have to now configure whatever build system I'm using (webpack) to support typescript and to explicitly hand transpiling the code. This is not the common way to distribute a package on npm.

In addition there is no main file defined in package.json. I am unsure what the main file is that I should include in order to get the full source. 
 

### To Reproduce

`npm i pyodide`

### Expected behavior

When I installed this I expected:

 - Javascript file or files
 - A single file defined as the main so that my requires or imports know what file to use when I request pyodide
 - The asm.js file for executing the pyodide application compiled with emscripten
 - The compiled python wasm that will be executed
 
 I believe all of these files should be included. This will make integration into other packages much easier.

### Environment

- Pyodide Version 0.20.0

### Additional context

I am happy to work on this if you agree. Hopefully I didn't misinterpret the point of the package on npm.

I have recently been using esbuild which is written in go and distributes a wasm for bundling. That code is successfully on the develop branch of cdk-web https://github.com/3p3r/cdk-web/tree/develop. I am comparing my experience dealing with the package distributed on npm for esbuild-wasm  vs. my experience with pyodide.

hoodmane:
Yeah, I agree that we are publishing the wrong stuff. It should be a simple fix: instead of publishing our Javascript source directory which is useless, we should publish the built files (especially because the directory `src/js` gets processed both by emscripten and rollup). Nobody here knows how to use node all that much so we haven't fixed it.

hoodmane:
For the record, to get something that actually works right now:
1. Download and unpack https://github.com/pyodide/pyodide/releases/download/0.20.0/pyodide-build-0.20.0.tar.bz2
2. you can `import  {loadPyodide} from "pyodide.mjs"` or `let {loadPyodide} = require("pyodide.js")`. The former is probably better, the latter defines `loadPyodide` and `_createPyodideModule` as global variables.

hoodmane:
[...]

Issue #2250, opened 2022-03-03
I want to mirror the numpy in my own cdn.


yxdragon:
I want to mirror the depend lib on my own cdn (to speed up the loading). for example, numpy.
how can I build a numpy.whl and use "micropip.install('xxxx.whl')"
I look for the numpy package source : https://github.com/pyodide/pyodide/blob/main/packages/numpy/meta.yaml
there is a zip url, and some patches. How to make the whl?

hoodmane:
From root directory run `PYTHONPATH=pyodide-build python -m pyodide_build buildpkg packages/numpy/meta.yaml`. The wheel will be in `packages/numpy/dist`.

Also, is there a reason you can't just download the wheel from our cdn and place it into yours?

yxdragon:
Do I need compiler to build the whl? where can I download your whl? I just find https://files.pythonhosted.org/packages/fb/48/b0708ebd7718a8933f0d3937513ef8ef2f4f04529f1f66ca86d873043921/numpy-1.21.4.zip in the yaml.

hoodmane:
The wheels are only on the development branch, #2027 will be included in pyodide v0.20. In v0.19 you need a pair of files, `numpy.js` and `numpy.data` available here:
https://cdn.jsdelivr.net/pyodide/v0.19.1/full/numpy.js
https://cdn.jsdelivr.net/pyodide/v0.19.1/full/numpy.data

yxdragon:
if you could look into this and let us know that'd be great =)

I do the test:
```html
 async function main(params) {
                log('Start!')
                pyodide = await loadPyodide({
                    indexURL: "https://cdn.jsdelivr.net/pyodide/v0.19.1/full/",
                });

                log('Install micropip!')
                await pyodide.loadPackage("micropip");

                await pyodide.runPythonAsync(`
                    import micropip
                    from js import log
                    log('Install numpy!')
                    await micropip.install(['numpy', 'planer', 'planer_zoo'])

                    log('Import numpy!')
                    from pyodide import to_js
                    import numpy as np
                    from time import time
                    import os, zipfile
                    import planer
                    from planer_zoo.ocr import paddleocr
                    
                    log('Download model!')
```

the result is:
load Pyodide: 3s
load Micropip: 1s
Install Numpy: 10s
import Numpy: 0

what does micro pip do? just download? or do some other thing such as unzip?
I think move numpy.whl on native cdn would works. and I would have a try.

[...]

PR #2394, opened 2022-04-12
Fix npm package

hoodmane:
When we load pyodide from node_modules, the calculated indexURL
starts with `file://` which causes trouble. Truncate it off.

Work towards fixing #2393.


src/js/pyodide.ts
```diff
@@ -175,7 +180,10 @@ function calculateIndexURL(): string {
   } catch (e) {
     err = e;
   }
-  const fileName = ErrorStackParser.parse(err)[0].fileName!;
+  let fileName = ErrorStackParser.parse(err)[0].fileName!;
+  if (IN_NODE && fileName.startsWith("file://")) {
+    fileName = fileName.slice("file://".length);
```
rth, line 185: This could use an explanatory comment about why we are doing this.

hoodmane: Well we get `file://path` but node expects paths not to start with `file://`

rth: OK, and I assume this would also work when importing pyodide js from local files? Could you please add a comment that this is necessary when determining indexURL with remote URLs in node?

hoodmane: > I assume this would also work when importing pyodide js from local files?

Well yeah. The path shouldn't ever start with `file://` in Node. If it starts with `file://` we remove it, if not we leave it alone. I think a sort of related issue is that in `compat.ts` in node we check if the path starts with `https://` in which case we make a web request otherwise we make a fs operation. But the fs operations get mad about paths starting with `file://`. Maybe a better solution would be to remove this in `compat.ts`.

src/js/package.json
```diff
@@ -1,6 +1,6 @@
 {
-  "name": "pyodide",
-  "version": "0.20.0",
+  "name": "--123hoodmane-pyodide",
```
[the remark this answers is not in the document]
hoodmane: Oops did not mean to commit this.

src/js/load-package.ts
```diff
@@ -160,7 +171,18 @@ async function downloadPackage(
...
+    return await _loadBinaryFile(baseURL, file_name);
+  } catch (e) {
+    if (!IN_NODE) {
+      throw e;
+    }
+  }
+  // If we are IN_NODE, download the package from the cdn, then stash it into
+  // the node_modules directory for future use.
+  let binary = await _loadBinaryFile(cdnURL, file_name);
+  await nodeFsPromisesMod.writeFile(`${baseURL}${file_name}`, binary);
```
ryanking13, line 184: I'm not very sure of this. IMO this kind of behavior should better be more explicit. For example, `puppeteer` downloads binaries during installation (though that is not what we want because Pyodide is not only for server-side usage), or `playwright` provides a command `playwright install` to download binaries after installation.

hoodmane: Maybe take a config variable? I guess my main concern is that `packages.json` is still going to include all of these packages but the request will fail when it tries to open the file which doesn't exist.

ryanking13: > My main concern is that `packages.json` is still going to include all of these packages but the request will fail when it tries to open the file which doesn't exist.

Yes, I also think it would be great if people don't need to download packages manually. So providing a way of automatically downloading packages would be nice for the user experience.

My opinion is that we should let Pyodide users know what is going on. They need to know that `whl` files are not included in the npm packages and those will be downloaded after installation.

If we are going to download packages in runtime as in this PR, I think there needs a proper log telling that it is downloading and stashing packages.

> Maybe take a config variable

Yeah, a config variable for deciding whether to download missing packages in runtime (or die with error) would be nice IMO. It would also be nice if CDN URLs can be configured at the top level as there [were some requests](https://github.com/pyodide/pyodide/issues/2250).
Still, I think it would be great to provide an explicit way to install packages before runtime (maybe `$ npx pyodide install` or something?) and recommend that way. The advantages of downloading them before runtime will be:

- If there is a problem in the CDN, it can be checked beforehand.
- Running time is stabilized (we don't need to download a large package at the first run)
- We can verify checksums of packages beforehand.

src/js/pyodide.ts
```diff
@@ -287,7 +287,11 @@ export async function loadPyodide(
 
   const pyodide = finalizeBootstrap(API, config);
   // API.runPython works starting here.
-
+  if (!pyodide.version.includes("dev")) {
+    // Currently only used in Node to download packages the first time they are
+    // loaded. But in other cases it's harmless.
+    API.setCdnUrl(`https://pyodide-cdn2.iodide.io/v${pyodide.version}/full/`);
```
ryanking13, line 293: Isn't `iodide.io` url deprecated BTW?

hoodmane: Probably. I still primarily use it...

src/js/load-package.ts
hoodmane: I am not likely to work on this myself, but I am happy if someone else would do it. In the meantime, I'm not sure what we should do in this PR. If you have an opinion, I would be happy for you to put in code that you like (you can push it directly to this branch). I think I agree with you about the long term solution, but I'm not what the best intermediate solutions are.

> I think there needs a proper log telling that it is downloading and stashing packages.

Perhaps adding logging to this for now would be okay? Or we could remove it from this PR and throw an error if a missing package is loaded?

ryanking13: > Perhaps adding logging to this for now would be okay?

Yeah I am okay with that for this PR. Perhaps let's not document this behavior for now. Thanks for working on this!

> I am not likely to work on this myself too soon, but I am happy if someone else would do it.

I'll open a issue so we can revisit this later.

src/js/pyodide.ts
ryanking13: Perhaps let's use `https://cdn.jsdelivr.net` as in our documentation.

ryanking13 (approved):
Thanks for working on node issues @hoodmane!

hoodmane:
Thanks for the reviews @rth and @ryanking13!

src/js/pyodide.ts
rth: FYI, we should definitely not use it, it doesn't cache well. I'm thinking we should fully discontinue it soon.

PR #2396, opened 2022-04-14
Fix type declarations

### LINKED ISSUE #2395: In pyodide 0.20.0, pyodide.d.ts only exports `loadPyodide`
With pyodide 0.19.0 (installed from the release tarball), the pyodide.d.ts file exported many types, including `PyProxy`, `PyProxyCallable`, and `loadPyodide`. As of 0.20.0, that file only exports `loadPyodide`. Is that intentional? Some of those types were useful for some TypeScript code I've been working on. 

In a similar vein, there previously were a few other .d.ts files included in the distrubution, like api.d.ts and pyproxy.gen.d.ts, but those files are no longer there. It wasn't clear to me exactly what the structure/purpose of these .d.ts files was, but I did make use of some of them.

hoodmane:
Hopefully resolves #2395.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -80,7 +80,7 @@ dist/pyodide.js: src/js/*.ts src/js/pyproxy.gen.ts src/js/error_handling.gen.ts 
 	npx rollup -c src/js/rollup.config.js
 
 dist/pyodide.d.ts: src/js/*.ts src/js/pyproxy.gen.ts src/js/error_handling.gen.ts
-	npx dts-bundle-generator -o my.d.ts pyodide.ts --export-referenced-types false
+	npx dts-bundle-generator -o pyodide.d.ts pyodide.ts --export-referenced-types false
 
 src/js/error_handling.gen.ts : src/core/error_handling.ts
 	cp $< $@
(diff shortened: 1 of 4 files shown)


Makefile
```diff
@@ -79,6 +79,9 @@ node_modules/.installed : src/js/package.json src/js/package-lock.json
 dist/pyodide.js: src/js/*.ts src/js/pyproxy.gen.ts src/js/error_handling.gen.ts node_modules/.installed
 	npx rollup -c src/js/rollup.config.js
 
+dist/pyodide.d.ts: src/js/*.ts src/js/pyproxy.gen.ts src/js/error_handling.gen.ts
+	npx dts-bundle-generator -o my.d.ts pyodide.ts --export-referenced-types false
```
ryanking13, line 83: I guess this is a typo?

```suggestion
	npx dts-bundle-generator -o pyodide.d.ts pyodide.ts --export-referenced-types false
```

src/js/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "pyodide",
-  "version": "0.20.0",
+  "version": "0.20.0-dev.0",
```
ryanking13, line 3:
```suggestion
  "version": "0.21.0-dev.0",
```

Makefile
hoodmane: Yup thanks.

rth (approved):
Thanks for fixing it!

rth:
For the changelog we can probably do a single entry about fixing the JS package (or otherwise feel free to add it later). Merging, thanks!

PR #2401, opened 2022-04-18
Fix garbage collection of once_callable

### LINKED ISSUE #2398: create_once_callable leak/throw if never called
## 🐛 Bug

There seems to be an issue with create_once_callable if the callback is never called.
If the object is released from the javascript side, an exception (Uncaught TypeError: Cannot set properties of undefined (setting 'leaked') (Chrome/Edge) or Uncaught TypeError: cache is undefined (Firefox)) occurs.
To reproduce, run this code, then force a garbage collection from devtools.

As far as i can tell, either

- create_once_callable must not be added to the FinalizationRegistry, or
- They must not be added with void 0 (undefined) as heldValue, or
- The cleanup function must not assume that heldValue is always an object with a leaked property

### To Reproduce

import js, pyodide

def test(): print("timeout")

timeoutID = js.window.setTimeout(pyodide.create_once_callable(test), 1000)
js.window.clearTimeout(timeoutID)

### Environment

- Pyodide Version: v0.20.0
- Browser version: Any

hoodmane:
This resolves #2398.


hoodmane:
Ideally we should add a test...

rth (approved):
Yes, a test would be good. Thanks!

PR #2405, opened 2022-04-18
Fix interrupt handler patch

### LINKED ISSUE #2400: Interrupt checking has become less reliable in 0.20.0
## 🐛 Bug

`setInterruptBuffer` *generally* works but some kinds of Python code no longer get interrupted in 0.20.0.

### To Reproduce

In 0.19.1, this Python code could be interrupted and a KeyboardInterrupt would be raised:

```python
while True:
    pass
```

In 0.20.0, that code no longer gets interrupted and just runs forever. This code *does* get interrupted:

```python
while True:
    str(3)
```

An example of more 'realistic' code that doesn't get interrupted:

```python
for i in range(100000000):
    i *= 2
    i += 9 + int(i < 3)
    str()
```

Here is where I'm actually testing this:

https://github.com/alexmojaki/pyodide-worker-runner/blob/5d7a7ba4652e23fb1f0483876793c3756cd688ef/test/index.ts#L197-L213

Upgrading that package to use 0.20.0 broke that test, replacing `pass` with `str(3)` as above fixed it (https://github.com/alexmojaki/pyodide-worker-runner/pull/9/commits/aee33fbead9c9afd22c3facc7c37226cb4138cc2)

### Expected behavior

Any pure-python synchronous code should eventually be interrupted, even something contrived like `while True: pass`, because sometimes people intentionally try stupid things to test what will happen, and sometimes people make dumb mistakes.

### Environment

- Pyodide Version: 0.20.0
- Browser version: 6 combinations of Chrome, Firefox, and Edge on Windows, OSX, and Ubuntu between myself and Browserstack.

hoodmane:
When Python is compiled with `USE_COMPUTED_GOTOS` (which seems to be the case in emscripten) `DISPATCH` does a calculated `goto` directly from one opcode to the next and never executes the top of the main loop. Instructions then opt into the eval breaker check by calling `CHECK_EVAL_BREAKER`. Thus the `CHECK_EVAL_BREAKER` macro needs to call `_Py_CHECK_EMSCRIPTEN_SIGNALS_PERIODICALLY` or else Emscripten signal handling won't work. The upstream patch included this call but when I backported the patch to Python v3.10 I accidentally left it out.

Resolves #2400.

### DIFF
--- a/cpython/patches/0001-Interrupt-handling.patch
+++ b/cpython/patches/0001-Interrupt-handling.patch
@@ -1,4 +1,4 @@
-From 743d9b5eff9a05f8187141a44403968f9bb457e3 Mon Sep 17 00:00:00 2001
+From e2e91c8940302f3c431d81c6ea2af2ea0571165d Mon Sep 17 00:00:00 2001
 From: Hood Chatham <roberthoodchatham@gmail.com>
 Date: Wed, 30 Mar 2022 14:50:30 -0700
 Subject: [PATCH] Interrupt handling
@@ -63,7 +63,7 @@ index 96881d4a49..1926f650b6 100644
          return 0;
      }
 diff --git a/Python/ceval.c b/Python/ceval.c
-index ab10b4166d..b06a453a97 100644
+index ab10b4166d..98bc690fae 100644
 --- a/Python/ceval.c
 +++ b/Python/ceval.c
 @@ -1152,6 +1152,30 @@ PyEval_EvalFrameEx(PyFrameObject *f, int throwflag)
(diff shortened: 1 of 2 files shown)


rth (approved):
Sounds good thanks! 

On a related note, in https://github.com/pyodide/pyodide/pull/1882 interrupt tests are the only tests that fail with the `-Os` flag (even with this patch). So there is still likely not fully stable in terms of compiler optimized code.

hoodmane:
There are a lot of `#ifdef`s in `ceval.c`. It might be worth investigating whether any of these compiler time branches go differently when compiling with `-Os` or what. Also, ideally it would be nice to fix `ceval.c` when compiled with `-Os` but probably even for size-optimized builds it would be a good idea to compile this file with `-O2` seeing as it's the hottest single file.

rth:
> but probably even for size-optimized builds it would be a good idea to compile this file with -O2 seeing as it's the hottest single file.

Indeed, I'll try that. Thanks!

Issue #1982, opened 2021-11-20
Optimize micropip test runtime 


rth:
Micropip tests are currently fairly slow, and we should improve their runtime if possible,
```
pytest packages/micropip --durations 10
```
Take 77.00s (0:01:17) to run only 13 tests on two browsers. 

Most of the time appears to be spent in the setup of the `selenium_standalone_micropip` fixture, 
```
========================================================================================= slowest 10 durations ==========================================================================================
4.64s setup    packages/micropip/test_micropip.py::test_install_custom_url[firefox-'.']
4.56s setup    packages/micropip/test_micropip.py::test_install_mixed_case2[firefox-jinja2]
4.55s setup    packages/micropip/test_micropip.py::test_install_custom_url[firefox-'{base_url}']
4.54s setup    packages/micropip/test_micropip.py::test_install_different_version[firefox]
4.53s setup    packages/micropip/test_micropip.py::test_install_mixed_case2[firefox-Jinja2]
4.52s setup    packages/micropip/test_micropip.py::test_install_different_version2[firefox]
4.07s setup    packages/micropip/test_micropip.py::test_install_simple[firefox]
2.55s setup    packages/micropip/test_micropip.py::test_install_simple[chrome]
2.53s setup    packages/micropip/test_micropip.py::test_install_custom_url[chrome-'.']
2.53s setup    packages/micropip/test_micropip.py::test_install_different_version2[chrome]
```
so this is mostly about improving pyodide load time in the `selenium_standalone` fixture,

Currently,

| Runtime (s)                                            | firefox | chrome | node |
|--------------------------------------------------------|---------|--------|------|
| selenium_standalone (load Pyodide)                     | 3.4     | 2.0    | 1.6  |
| selenium_standalone_micropip (load Pyodide + micropip) | 4.2     | 2.5    | 2.2  |

ryanking13:
I'm not sure whether there is room for optimizing setup time.
Maybe we can compare the performance of selenium with other products like playwright and consider migration? ([speed comparison](https://blog.checklyhq.com/puppeteer-vs-selenium-vs-playwright-speed-comparison/))

rth:
We should profile the load time, for instance, it's possible that similarly to packages (see https://github.com/pyodide/pyodide/issues/347#issuecomment-933834657) the bottleneck is WASM compilation.

> Maybe we can compare the performance of selenium with other products like playwright?

Could be interesting indeed.

PR #2406, opened 2022-04-19
Reuse selenium instance for package import test

ryanking13:
The purpose of this PR is to reduce CI running time.

- Reuse selenium instance for package import tests.
- Reduce the number of repetitions of benchmarks.

### DIFF
--- a/packages/test_packages_common.py
+++ b/packages/test_packages_common.py
@@ -41,7 +41,7 @@ def test_parse_package(name):
 
 
 @pytest.mark.skip_refcount_check
-@pytest.mark.driver_timeout(40)
+@pytest.mark.driver_timeout(60)
 @pytest.mark.parametrize("name", registered_packages())
 def test_import(name, selenium):
     if not _package_is_built(name):
--- a/packages/test_packages_common.py
+++ b/packages/test_packages_common.py
@@ -43,7 +43,7 @@ def test_parse_package(name):
 @pytest.mark.skip_refcount_check
 @pytest.mark.driver_timeout(40)
 @pytest.mark.parametrize("name", registered_packages())
-def test_import(name, selenium_standalone):
+def test_import(name, selenium):
     if not _package_is_built(name):
         raise AssertionError(
             "Implementation error. Test for an unbuilt package "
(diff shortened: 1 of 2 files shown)


packages/test_packages_common.py
```diff
@@ -41,9 +41,9 @@ def test_parse_package(name):
 
 
 @pytest.mark.skip_refcount_check
-@pytest.mark.driver_timeout(40)
+@pytest.mark.driver_timeout(60)
 @pytest.mark.parametrize("name", registered_packages())
-def test_import(name, selenium_standalone):
+def test_import(name, selenium):
```
rth, line 46: Well, this in particular aims to check that run dependency are correct. if we reuse the selenium instance, we could have a case, where dependencies are incorrect but the package was loaded in a previous test and so the current test passes even when it shouldn't. 

Though yes, it's slow and a bit unfortunate.

hoodmane: Maybe try reloading Pyodide rather than refreshing the page? My guess is this'll lead to trouble but it's worth trying just to see.

rth: Yes, I think we would eventually run into memory issues if we do that. Particularly with little RAM available in CI.

An ideal solution would to re-use the same browser profile and browser cache (including WASM compilation cache for Chrome), but I have never managed to make it work reliably when I tried in the past . By default, selenium creates a new browser profile for each new test run, with is really not great for our use case.

ryanking13: > Well, this in particular aims to check that run dependency are correct. if we reuse the selenium instance, we could have a case, where dependencies are incorrect but the package was loaded in a previous test and so the current test passes even when it shouldn't.

Yeah, that could be problematic. I'll seek for a better solution.

ryanking13:
Okay, I'm trying to reload Pyodide for each test run. Thanks, @hoodmane for making this available. Let's see what happens.

hoodmane:
No surprise:

> unknown error: session deleted because of page crash

ryanking13:
Well, at least it died after running ~50 tests, which is positive than I thought :)

We probably can optimize some of other tests with this "reload" idea (e.g. #1982)

ryanking13:
I'll let this PR draft until I find a better way to manage this. I'll apply benchmark and timeout updates separately in #2409.

hoodmane:
50 tests is surprisingly many. If we could group them into batches of 20 that reuse the same page with loadPyodide and then refresh that'd probably be an improvement. This might be hard though. I tried before to refresh the page on fatal errors but I couldn't get it to work. Maybe we could also automatically reload Pyodide on fatal error now...

But as a first step we can probably get rid of most of the other uses of `selenium_standalone`.

PR #2411, opened 2022-04-20
Pkg h5py

bmaranville:
### Checklists

- [ ] Add / update tests


hoodmane:
It will try to build a new package automatically if the file `packages/my_package/meta.yaml` exists, so I think what you've done is enough. Unless I am not understanding your question.

Makefile.envs
```diff
@@ -1,5 +1,5 @@
 export PYVERSION ?= 3.10.2
-export PYODIDE_EMSCRIPTEN_VERSION ?= 2.0.27
+export PYODIDE_EMSCRIPTEN_VERSION ?= 2.0.28
```
hoodmane, line 2:
```suggestion
export PYODIDE_EMSCRIPTEN_VERSION ?= 2.0.27
```
There is a reason we are using Emscripten v2.0.27:
https://github.com/emscripten-core/emscripten/issues/16538

bmaranville: I was worried this might cause problems - but there is some other clang or LLVM issue that causes a build error for version 2.0.27 and libhdf5.  It doesn't seem to happen for any other version of Emscripten.

hoodmane: That is lucky! These LLVM issues are pretty worrisome. Hopefully I can get some time to work on this issue soon. I think it's the only blocker for us to get up to the most recent Emscripten release which would be great for us.

packages/h5py/patches/configure.patch
```diff
@@ -0,0 +1,57 @@
...
+     def _load_hdf5_lib(self, libdirs):
+         """
+@@ -248,20 +249,21 @@ class HDF5LibWrapper:
+ 
+         Raises an exception if anything goes wrong.
+         """
+-        import ctypes
+-        from ctypes import byref
++        # import ctypes
++        # from ctypes import byref
```
hoodmane, line 22: Just delete stuff rather than commenting it out. It makes the patch shorter and the deleted code is in the patch in any case.
```suggestion
```

hoodmane (approved):
Looks good to me if CI passes, other than my minor nitpick about the patches.

docs/project/changelog.md
```diff
@@ -399,7 +399,7 @@ async function runPythonAsync(code, globals) {
   {pr}`2028`
 
 - New packages: logbook {pr}`1920`, pyb2d {pr}`1968`, and threadpoolctl (a
-  dependency of scikit-learn) {pr}`2065`
+  dependency of scikit-learn) {pr}`2065`, libhdf5 {pr}`1799`, h5py {pr}`2180`
```
hoodmane, line 402: Please move this up to the "Unreleased" section (it's ended up in the v0.20.0 changelog which isn't correct anymore).

rth:
OK, great. For better attribution, let's merge libhdf5 in https://github.com/pyodide/pyodide/pull/1799 then sync with main and merge the rest in this PR. I'll make changes to that PR accordingly.

bmaranville:
The final version of meta.yaml for libhdf5 that should be used is the one from this PR, not the original one.  I'll update it if you like.

rth:
OK, I realized that the other PR is also by you. So yes let's merge this :) Can you close the other one please?

rth:
Thanks a lot @bmaranville and @ax3l !

Issue #2397, opened 2022-04-14
More checks for the JS package before release


rth:
To avoid issues like https://github.com/pyodide/pyodide/issues/2393 I think it would be good to add more checks before the release of the JS package. 

In particular, we could,
 - [ ] re-enable the check in https://github.com/pyodide/pyodide/blob/main/.circleci/config.yml#L312
 - [ ] add a CI script that checks that a simple JS script with node using pyodide installed from the built JS package works
 - [ ] BTW, I thought we had the auto-deployment of the JS package to NPM, but it looks like it's not the case (or at least I couldn't find it). It would be good to add it.

hoodmane:
If we copy `package.json` into `dist` then we can use `npm link` inside `dist` and then `npm link pyodide` inside a test directory. We should then use this in the node selenium tests and the typing tests.

hoodmane:
It seems to me like the mocha tests aren't running at all.

hoodmane:
We now have the problem that we "roll up" the files and the `/* webpackIgnore: true */` comments get compiled out.

PR #2413, opened 2022-04-20
MAINT Fix being able to run javascript tests

mneil:
### Description

Looking at https://github.com/pyodide/pyodide/issues/2397 tests are not even running at all for the javascript package. I pulled and confirmed. I want to make changes to this package specifically on the JS side to support the browser better. Before I make too many changes though it would be good if the existing tests could run. This change fixes the js directory so that tests will actually run. All files were changed to typescript and ts-mocha now executes the tests. There is no need to make specific source files for esm, mjs, etc... These should be handled by the build process to output the correct module definitions like umd, etc...

Tests are failing. This is because the pyodide.asm.js file is not in the default directory. I will open a separate PR to change actual source code and resolve the tests so they work as well as add tests specifically for the browser via puppeteer.


src/js/module.ts
```diff
@@ -1,3 +1,15 @@
+export interface Module {
+  noImageDecoding: boolean;
+  noAudioDecoding: boolean;
+  noWasmDecoding: boolean;
+  preRun: { (): void }[];
+  print: (a: string) => void;
+  printErr: (a: string) => void;
+  ENV: { [key: string]: string };
+  preloadedWasm: any;
```
hoodmane, line 9:
```suggestion
  preloadedWasm: { [key: string]: any };
```

hoodmane:
>  I want to make changes to this package specifically on the JS side to support the browser better.

Fantastic! Thanks @mneil!

> Tests are failing. This is because the pyodide.asm.js file is not in the default directory. I will open a separate PR to change actual source code and resolve the tests so they work as well as add tests specifically for the browser via puppeteer.

So the tests still disabled in this PR? What further change is needed to enable them?

mneil:
Yes, tests are still disabled because they are failing. But now they run at least :). IDK what all is wrong but the first issue is the filepath to pyodide.asm.js. That file is built and in the dist but the test loads it assuming a root of cwd or the browser window.location.

mneil:
Would you like me to fix the tests in this PR as well or open it in another one. I fixed the loading by passing an indexURL of the dist directory from the tests https://github.com/mneil/pyodide/blob/fix-tests/src/js/test/conftest.ts#L13 Most tests are now passing. Micropip is failing to install now so not all tests are good yet.

mneil:
Ok, actually, fixed the tests so I went ahead and added them to this. All tests run and are passing. I haven't change anything with respect to them running in CI. I don't see an obvious place to change that at the moment. Feel free to reenable them or point me out where to me and I'll make them run again. Thank you!

hoodmane:
Well the CI tries to run the js tests here:
https://github.com/pyodide/pyodide/blob/main/.circleci/config.yml#L233

hoodmane:
This change set looks good to me but readthedocs is failing. The error looks a bit mysterious, I'm not sure what it has to do with your changes. Anyways I'm happy to merge this if you fix the readthedocs failure.

ryanking13:
> This change set looks good to me but readthedocs is failing. The error looks a bit mysterious, I'm not sure what it has to do with your changes. Anyways I'm happy to merge this if you fix the readthedocs failure.

https://github.com/pyodide/pyodide/blob/a7bd3cf63c099216ef7ecee6e90e0864ed28a091/src/js/module.ts#L1-L11

Maybe not exporting this interface would fix?

hoodmane:
Seems important to export the interface for the Module if we are going to bother to write it down at all. It's not like hardly anything happens in `module.ts` itself.

ryanking13:
#2417 will fix the doc build issue, please merge it re run CI after it is merged to main.

rth:
Thanks a lot for doing this @mneil !

mneil:
This was closed and not merged. Would you like me to open another one to fix these tests? Is there something more you'd like me to change?

hoodmane:
Probably was just a mistake.

src/js/module.ts
```diff
@@ -1,3 +1,15 @@
+export interface Module {
```
hoodmane, line 1:
```suggestion
/** @private */
export interface Module {
```

src/js/test/pyodide.test.ts
```diff
@@ -13,7 +13,7 @@ describe("Pyodide", () => {
   });
   it("micropip.install", async () => {
     // TODO: micropip currently requires a globally defined fetch function
-    global.fetch = fetch;
+    // global.fetch = fetch;
```
hoodmane, line 16:
```suggestion
```

ryanking13:
Oops, sorry. This was automatically closed by Github via #2417. Thanks for your work!

PR #2414, opened 2022-04-21
Fix micropip package name resolution

### LINKED ISSUE #2408: micropip does not resolve locally available versions
I have three wheels:

* travertino is v0.1.3
* toga-core is v0.3.0.dev33, and declares travertino>=0.1.3 as a requirement
* toga-web is v0.3.0.dev33, and declares toga-core==0.3.0.dev33 as a requirement

All three exist as `.whl` files in my static files folder. Travertino 0.1.3 has been published to PyPI; toga-core and toga-web dev32 have been published to PyPI, but dev33 are local development builds. 

On my Pyodide page, I run:
```
import micropip
micropip.install([
    './static/travertino-0.1.3-py3-none-any.whl',
    './static/toga_core-0.3.0.dev33-py3-none-any.whl',
    './static/toga_web-0.3.0.dev33-py3-none-any.whl',
])
```
This raises the following exception:
```
    raise self._exception
  File "/lib/python3.10/asyncio/tasks.py", line 234, in __step
    result = coro.throw(exc)
  File "/lib/python3.10/site-packages/micropip/_micropip.py", line 183, in install
    transaction = await self.gather_requirements(requirements, ctx, keep_going)
  File "/lib/python3.10/site-packages/micropip/_micropip.py", line 173, in gather_requirements
  File "/lib/python3.10/asyncio/futures.py", line 284, in __await__
    yield self  # This tells Task to wait for completion.
  File "/lib/python3.10/asyncio/tasks.py", line 304, in __wakeup
    future.result()
    raise self._exception
    result = coro.send(None)
  File "/lib/python3.10/site-packages/micropip/_micropip.py", line 245, in add_requirement
    await self.add_wheel(name, wheel, version, (), ctx, transaction)
  File "/lib/python3.10/site-packages/micropip/_micropip.py", line 316, in add_wheel
    await self.add_requirement(recurs_req, ctx, transaction)
  File "/lib/python3.10/site-packages/micropip/_micropip.py", line 286, in add_requirement
    raise ValueError(
ValueError: Couldn't find a pure Python 3 wheel for 'toga-core==0.3.0.dev33'. You can use `micropip.install(..., keep_going=True)` to get a list of all packages with missing wheels.
```

If I relax the `install_requires` on toga-web to remove the declared dependency on toga-core, the import succeeds, and the code runs successfully.

I've tried splitting the install into two passes:
```
await micropip.install([
    './static/travertino-0.1.3-py3-none-any.whl',
    './static/toga_core-0.3.0.dev33-py3-none-any.whl',
])
await micropip.install([
    './static/toga_web-0.3.0.dev33-py3-none-any.whl',
])
```
thinking that if toga-core's installation was completed, toga-web would be able to find it; that makes no difference. 

I've also tried producing the packages without the .dev33 version suffix, thinking that it might be "prerelease" versions that are the issue, but that made no difference either.

However, if I revert the version number on toga-core and toga-web to 0.3.0.dev32 (the last versions published to PyPI), the import *succeeds* - but the old version (published to PyPI) version of the toga-core code is used. 

Looking at the network requests that the page has made indicates that the wheels were all loaded - and then micropip requested the toga-core PyPI metadata, and toga_core-0.3.0.dev32 (the most recent published version) was downloaded from PyPI. Interestingly, *no* request was made to load Travertino 0.1.3 from PyPI as a result of the dependency declared in toga-core.

ryanking13:
### Description

Hopefully resolve #2408

### Checklists

- [x] Add / update tests


packages/micropip/src/micropip/_micropip.py
```diff
@@ -312,10 +315,43 @@ async def add_wheel(self, name, wheel, version, extras, ctx, transaction):
...
+        if real_name == "UNKNOWN":
+            real_name = name
+        try:
+            real_version = dist.version
+        except ValueError:
+            real_version = version
+
+        if real_name != name or real_version != version:
+            transaction["locked"].pop(name)
+            transaction["locked"][real_name] = PackageMetadata(
```
rth, line 333: I'm not sure that's the best approach. 

How about rather in the list of our locked and installed packages use `name.replace('-', '_')` as key? Then updating the actual name in the PackageMetadata.name.
That would be better than erroring if multiple variants are installed. Unless there are cases when a package name with a '-' vs '_' are different packages, but I think that would very ill-advised anyway.

hoodmane: Yeah +1 for this. We should be normalizing the keys. Maybe packaging has a method to do that?

ryanking13: Yes, I think keeping both the normalized name and actual project name will work without this update thing. I'll try with that. Thanks for the suggestion.

packages/micropip/src/micropip/_micropip.py
```diff
@@ -337,6 +337,20 @@ async def add_wheel(self, name, wheel, version, extras, ctx, transaction):
...
+            if real_name != _name:
+                continue
+
+            if real_version == _version:
+                return
+
+            raise ValueError(
+                f"Dependency resolution failed:"
+                f"Trying to install multiple versions of {real_name} ({real_version} <==> {_version})."
+            )
```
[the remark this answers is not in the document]
ryanking13: I added this to prevent from multiple versions of packages being queued. But I think we won't need this if we normalize names beforehand.

ryanking13:
Okay, keys for `installed_packages` are now normalized.

Note that as @rth mentioned, when there exist packages with the same normalized names (e.g. `blah-blah` and `blah_blah`) it will cause some trouble. But I hope that won't happen...

rth (approved):
Thanks a lot for fixing this @ryanking13 !

ryanking13:
Thanks for the review!

packages/micropip/src/micropip/package.py
```diff
@@ -31,6 +32,10 @@ def format_row(values, widths, filler=""):
     return "\n".join(rows)
 
 
+def normalize_package_name(pkgname: str):
+    return re.sub(r"[^\w\d.]+", "_", pkgname, re.UNICODE)
```
bollwyvl, line 36: perhaps also apply `.lower()`?

rth: yes, indeed, that would definitely make sense. Thanks @bollwyvl !

bollwyvl: e.g. https://github.com/pyodide/pyodide/pull/2445

PR #2416, opened 2022-04-21
Fix pyodide-build installation when in Python virtual environment

ryanking13:
### Description

A quick fix for https://github.com/pyodide/pyodide/issues/2020#issuecomment-1104065555.

@bmaranville could you confirm you are able to build with this fix?

### DIFF
--- a/packages/Makefile
+++ b/packages/Makefile
@@ -15,7 +15,12 @@ all: pyodide-build
 	--log-dir=./build-logs
 
 pyodide-build: ../pyodide-build/pyodide_build/**
-	$(HOSTPYTHON) -m pip install --prefix=$(shell $(HOSTPYTHON) -m site --user-base) -e ../pyodide-build
+ifdef VIRTUAL_ENV
+	export PREFIX=""
+else
+	export PREFIX="--prefix=$(shell $(HOSTPYTHON) -m site --user-base)"
+endif
+	$(HOSTPYTHON) -m pip install $(PREFIX) -e ../pyodide-build
 
 update-all:
 	for pkg in $$(find . -maxdepth 1 ! -name ".*" -type d -exec basename {} \; | tail -n +2); do \
--- a/packages/Makefile
+++ b/packages/Makefile
@@ -15,12 +15,7 @@ all: pyodide-build
 	--log-dir=./build-logs
 
 pyodide-build: ../pyodide-build/pyodide_build/**
-ifdef VIRTUAL_ENV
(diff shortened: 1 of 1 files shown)


packages/Makefile
```diff
@@ -15,7 +15,12 @@ all: pyodide-build
 	--log-dir=./build-logs
 
 pyodide-build: ../pyodide-build/pyodide_build/**
-	$(HOSTPYTHON) -m pip install --prefix=$(shell $(HOSTPYTHON) -m site --user-base) -e ../pyodide-build
+ifdef VIRTUAL_ENV
+	export PREFIX=""
+else
+	export PREFIX="--prefix=$(shell $(HOSTPYTHON) -m site --user-base)"
+endif
+	$(HOSTPYTHON) -m pip install $(PREFIX) -e ../pyodide-build
```
rth, line 23: Actually, I wonder if we should just ask users to pip install it before running make and get rid of the prefix altogether (users should use a virtualenv or a  conda env).

ryanking13: Yes, I think that is a good idea.

However, even if we ask users to install `pyodide-build` themselves before running `make`, this issue (editable install not working with pyproject.toml) will keep happening unless we stop using editable install.

bmaranville:
I can confirm that this works in an environment created with the python venv module.  Thanks!

packages/Makefile
rth: > this issue (editable install not working with pyproject.toml) will keep happening unless we stop using editable install.

I know there was some trouble with it, but I think they disabled build isolation with editable installs now ? In any case, I have no issues with using editable installs with pyproject.toml in other regular Python packages.

ryanking13: Hmm, yes I can also build `pyodide-build` in editable install mode without setting `--prefix` now.

I am testing with multiple Python / pip / setuptools versions, and it looks like Python >= 3.7 successes while 3.6 is failing to install in editable mode. I'm not sure which change made it available... I can't find an explicit changelog for this.

So maybe we can just remove `--prefix` and tell people to upgrade `pip` and `setuptools` if it fails? WDYT?

rth: Yes, let's do that. The minimal requirements for setuptools [is already in pyproject.toml](https://github.com/pyodide/pyodide/blob/main/pyodide-build/pyproject.toml#L2) but yes, it won't work with old pip versions. It's up to the user to have a reasonably recent pip.

PR #2418, opened 2022-04-21
MAINT Split conftest.py into modules

ryanking13:
### Description

Modularize fixtures / contexts / functions that are used in the test, which were previous all stacked in a long `conftest.py` file. I hope this change will help us adding new test features in the future (e.g. #2091)

In a long term, we probably can make this into a standalone pytest plugin (e.g. `pytest-pyodide`) so that Pyodide users can easily use them in other projects.

Note that I also moved helper functions in `pyodide-build.testing` into `pyodide_test_runner`. 

### Checklists

- [x] Add / update tests


conftest.py
```diff
@@ -1,32 +1,22 @@
...
 import sys
-import tempfile
-import textwrap
-import time
 
-import pexpect
 import pytest
 
+from test_fixture.decorator import *  # noqa: F401, F403
+from test_fixture.fixture import *  # noqa: F401, F403
```
rth, line 10: I'm not very keen on this. I think it would be better to a) avoid * imports, b) import things from the module where they are defined rather than conftest.py which is supposed to be a pytest-specific module not something we import from. Though we had started doing that a lot.

test_fixture/__init__.py
```diff
@@ -0,0 +1,16 @@
+from .browser import ChromeWrapper, FirefoxWrapper, NodeWrapper, SeleniumWrapper
```
rth, line 1: Maybe let's call this pyodide_pytest or pyodode-test-runner (if we want to be able to use this without pytest, not sure) even if it's not yet a proper pytest plugin?

test_fixture makes me think of a folder where tests for fixtures are rather than a folder that defines fixtures.

rth (commented):
Thanks for working on this @ryanking13 ! A few comments below. Me or someone else would need to review it in more detail, but I think it's going in the right direction. Our conftest.py is getting a bit out of control, particularly if we want to add playwright there one day, so it makes sens to put more of this logic into a separate module. 

The question how do we intend to distribute it? As `pyodide-pytest` or `pyodide-test-runner` on PyPy that depends on a specific version of Pyodide? That could work.

ryanking13:
> The question how do we intend to distribute it? As `pyodide-pytest` or `pyodide-test-runner` on PyPy that depends on a specific version of Pyodide? That could work.

I think the test runner does not need to depend on a specific version of Pyodide, because I think the API that the test runner depends on does not change frequently. But until we can assure that Pyodide API is stable enough, yes we can use the same version with Pyodide.

conftest.py
ryanking13: I made decorators directly imported from `pyodide-test-runner`. Note that it is not a proper package yet, instead I injected import path in `conftest.py`.

rth (approved):
Thanks @ryanking13 , the current version LGTM, and this is indeed much better than what we have now. The code itself wasn't changed much, just moved around right? I have only reviewed code organization not code details.

ryanking13:
> Thanks @ryanking13 , the current version LGTM, and this is indeed much better than what we have now. The code itself wasn't changed much, just moved around right? I have only reviewed code organization not code details.

Yes, there are some minor tweaks to make this work (e.g. injecting `--dist-dir` to benchmark), except that it's 99% just moving around codes.

PR #2425, opened 2022-04-29
Javascript tests run in browser and node against build pyodide.js

mneil:
## Description

Change javascript tests so that they run in both the browser (chrome via puppeteer) and in nodejs. 

I had previously converted all tests to `.ts` https://github.com/pyodide/pyodide/pull/2413. I had to revert the tests to use JS instead because typescript complicated the browser test too much when trying to do dynamic execution of defined code in the tests.

### Major changes

- Tests are now broken into 2 types. Unit and integration
- Unit tests are in typescript. They test individual working units of the typescript source. These are most useful for checking that changes to the source have no regressions or side effects
- Integration tests are written in Javascript. They follow a convention that allows the same test to be ran in both nodejs or the browser (Chrome). These tests run against the built javascript, not the source, and ensure completeness to the user interface exposed in the built package

### Notes

1. I did not add any additional tests. I only rearchitected the existing tests to add support for automated browser testing
2. Tests still do not run in CI. Looking at the setup circle ci is currently running tests in the `src/test-js` folder. IDK what purpose those tests serve other than type checking.

Based on point 2 I propose that the test-js folder is removed. Type assertions can be moved into the unit test folder of the `js` source. Circle CI task can be updated to run the tests in the js folder instead.

#### Missing / Not Doing

There are things I opted not to do in this PR in order to keep it smaller so we can discuss just the merit of this change.

- CI is broken. I don't think CI working impacts whether or not these changes should be accepted. Is this type of parallel testing something you want to include here? Does the format look ok? What other questions are there?
- I setup coverage but am not running it / enforcing it. Collecting coverage against browser / node is difficult. Coverage should be handled in another PR. Coverage did not previously exist.


hoodmane (approved):
I think this looks good to me, it would be good if @rth or @ryanking13 would also review this.

rth:
Thanks for doing this work @mneil !

> Based on point 2 I propose that the test-js folder is removed. Type assertions can be moved into the unit test folder of the js source.

Sounds good to me. It would be indeed good to finally run those tests in CI. Looking forward to your MR )

rth (approved):

hoodmane:
Thanks @mneil! We really appreciate your help.

PR #2426, opened 2022-05-01
Add run_js to pyodide Python package

hoodmane:
### Description

Add a `run_js` API that evals JavaScript code and returns the result as a `JsProxy`.

### Checklists

- [x] Add / update tests


ryanking13:
Thanks, @hoodmane. I'm just curious, what would be the difference between using `from js import eval` or `pyodide.run_js`?

hoodmane:
Good question. The only difference is in the error handling. `eval(not_a_string)` returns `not_a_string` whereas `run_js` raises `ValueError`. But I kind of forgot we could just do `from js import eval` when I wrote this PR.

I think the better thing would be to provide `run_js` as a Python function rather than doing this stuff in C. I think raising a `ValueError` is better behavior when getting a weird argument than silently returning it. But then we could do:
```
def run_js(code, /):
	from js import eval

	if not isinstance(code, str):
		raise ValueError(f"...")
	eval(code)
```

src/py/pyodide/_run_js.py
```diff
@@ -0,0 +1,8 @@
+def run_js(code, /):
```
rth, line 1: This needs a docstring.

rth (approved):
Indeed, it's a small improvement over JS eval, but still could be worth it,
 - to avoid overwriting python eval
 - to have better error checking
 - for better discoverability as part of the API docs

A minor comment otherwise LGTM.

PR #2427, opened 2022-05-01
MAINT Mark position-only params in _core_docs

hoodmane:
Most C functions have a position-only first parameter. This marks them as position-only in pypy and in the docs.

### DIFF
--- a/src/py/_pyodide/_core_docs.py
+++ b/src/py/_pyodide/_core_docs.py
@@ -144,7 +144,7 @@ class JsProxy:
         when the promise resolves.
         """
 
-    def catch(self, onrejected: Callable) -> "Promise":
+    def catch(self, onrejected: Callable, /) -> "Promise":
         """The ``Promise.catch`` API, wrapped to manage the lifetimes of the
         handler.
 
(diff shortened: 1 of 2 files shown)


ryanking13 (approved):
Thanks, @hoodmane! There is a failing test, except that LGTM.

src/py/_pyodide/_core_docs.py
```diff
@@ -280,7 +280,7 @@ def _into_file(self, file: IOBase):
         data.
         """
 
-    def to_string(self, encoding=None) -> str:
```
rth, line 283: This is one endpoint I would have maybe expected to work with kwarg, but yes, OK if that's what we have now.

rth (approved):
Thanks!

src/py/_pyodide/_core_docs.py
hoodmane: Actually we *do* support encoding with keyword args:
```C
  static const char* const _keywords[] = { "encoding", 0 };
  static struct _PyArg_Parser _parser = { "|s:to_string", _keywords, 0 };
```
so this should be reverted.

Issue #2403, opened 2022-04-18
Release alpha version with node fixes


hoodmane:
It would be good to release an alpha version with #2394 to help the node people test whether #2393 and related issues are fixed. We also have problems with webpack which probably aren't fixed, maybe we should try to improve the webpack situation first too.

I'm not sure how to version this stuff. I wouldn't normally want to include #2387 and #2391 in a patch release, but we could backport everything else. Or we could throw them into `0.20.1` they aren't humongous changes. Or we could plan to do a very small `v0.21.0` release.

 @alexmojaki @mnell @rth @ryanking13

ryanking13:
I think `v0.20.1` would be good because I recently found that we didn't version `pyodide-build` `v0.20.0` correctly in PyPI (https://pypi.org/project/pyodide-build/#history), so we can bump the version together.

hoodmane:
I think it would be good to release an alpha first e.g., `0.20.1a1` so that if there are more dumb mistakes with the node/webpage stuff we can iron those out first.

https://github.com/pyodide/pyodide/pull/2405 would also be good to include.

ryanking13:
> I think it would be good to release an alpha first e.g., `0.20.1a1` so that if there are more dumb mistakes with the node/webpage stuff we can iron those out first.
> 
> #2405 would also be good to include.

I'm +1 with it

alexmojaki:
Seems like this release could be made now? I'm keen to try it out.

ryanking13:
I made an alpha release `v0.20.1a1`, though we didn't fix webpack situation yet... #2387 and #2391 are included as they have conflict with recent changes and I didn't want to mess up. Excluded commits are #2411, #2407, #2389, #2305.

@rth @hoodmane could you publish JS package to npm registry? I don't have access to our npm registry.

rth:
Thanks! No worries, as long as it worked. I send you the invite.

ryanking13:
I find out that I published the wrong stuff (TypeScript files, not JavaScript) in npm `pyodide@0.20.1-alpha.1`. 

It looks like npm does not allow removing and republishing the same version. So I deprecated `pyodide@0.20.1-alpha.1` for now. We should publish again in another version name (Maybe, `pyodide@0.20.1-alpha.2`?).

Sorry about this. FYI, this happened because I called `npm publish` in `src/js` not in `dist`, which was changed after #2394.

rth:
[...]

PR #2443, opened 2022-05-02
Export PyodideInterface

hoodmane:
We should add type tests for this @mneil but I'm worried that doing it here might create conflicts with #2425 so maybe I'll hold off a bit.

_Requested by @alexmojaki in https://github.com/pyodide/pyodide/issues/2403#issuecomment-1114731607_


ryanking13 (approved):
Thanks, @hoodmane! LGTM, just in case let's merge main to this branch before merging.

PR #2445, opened 2022-05-02
also lowercase when normalizing micropip package names

bollwyvl:
### Checklists

- [x] Add / update tests

This extends #2414 to include uppercase characters.


hoodmane (approved):
Thanks @bollwyvl! Looks good aside from linter error.

rth:
Or you could just sync with main and pre-commit.ci should take care of it.

bollwyvl:
Thanks for the quick review, all: looks like the linter has been appeased!

ryanking13:
Thanks for catching this, @bollwyvl!

Issue #2442, opened 2022-05-02
Allow first argument of unpackArchive to be an ArrayBuffer


hoodmane:
## 🚀 Feature

Allow first argument of `unpackArchive` to be an `ArrayBuffer` as opposed to just a TypedArray. 
_Originally requested by @alexmojaki in https://github.com/pyodide/pyodide/issues/2403#issuecomment-1114731607_

hoodmane:
Closed by #2451.

PR #2451, opened 2022-05-02
Fix 1st argument of unpackArchive to allow for ArrayBuffer

vickumar1981:
### Description

Attempt at addressing: https://github.com/pyodide/pyodide/issues/2442

### Checklists

Added a test to assert that a TypeError with the right message is raised on exception.

- [x] Add / update tests


src/js/api.ts
```diff
@@ -378,6 +378,15 @@ export function unpackArchive(
     }
     options = { extractDir: options };
   }
+  if (ArrayBuffer.isView(buffer)) {
+    buffer = buffer.buffer;
+  }
```
rth, line 383: If I understand correctly if it's a TypedArray it's converted to ArrayBuffer here , and then on the following line back to TypedArray.

So that a roundtrip for TypedArray input. Can't we only convert ArrayBuffer to  TypedArray and keep the TypedArray unchanged?

rth (commented):
Thank you @vickumar1981 ! Just one comment below. 

Also please add a changelog entry to https://github.com/pyodide/pyodide/blob/main/docs/project/changelog.md

src/js/api.ts
hoodmane: Ideally we always want to pass `Uint8Array`. I'm concerned that the Emscripten FS APIs misbehave otherwise.

PR #2454, opened 2022-05-02
update: added missing build==0.7.0 dependency in requirements.txt

Zyell:
While building Pyodide and running tests, I came across a missing dependency in the requirements.txt file, so I added the requirement build==0.7.0.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -2,6 +2,7 @@
   # lint
   pre-commit
   # testing
+  build==0.7.0
   hypothesis
   pexpect
   pytest


rth (approved):
Thanks @Zyell !

ryanking13:
Thanks, @Zyell! I'm not sure why `build` isn't installed during the build [because is a dependency of `pyodide-build`](https://github.com/pyodide/pyodide/blob/6e4a63fb894bdba0c0072c34bfbbabb128d93821/pyodide-build/setup.cfg#L29).

Issue #2431, opened 2022-05-01
Verify checksums of Pyodide packages


rth:
A subset of https://github.com/pyodide/pyodide/issues/1881. 

To improve security it would be good to verify checksums of packages built by Pyodide when we are installing them.

This would involve,

1. [x] including a `sha256` key to package.json e.g.
        ```json
        packages": {
            "asciitree": {
                  "name": "asciitree", 
                  "version": "0.3.3", 
                  "file_name": "asciitree-0.3.3-py3-none-any.whl",
                  "sha256": "<sha256>"
                  ...
        ```
        which is generated in  [generate_packages_json](https://github.com/pyodide/pyodide/blob/3f9ae2ed1e1aee688251af86a4dc23d3d867740c/pyodide-build/pyodide_build/buildall.py#L433)
        
3. [ ] verify this checksum when installing it in [pyodide.loadPackage](https://github.com/pyodide/pyodide/blob/263309102d8158d82d051d0f3d4bbf7a4505c52a/src/pyodide.js#L288)

mger:
I will take a stab at this for the sprint.

rth:
Thanks for your work @mger! For the second part, actually, [fetch](https://developer.mozilla.org/en-US/docs/Web/API/fetch#credentials) already has `integrity` parameter which can be used to validate sha256. So it would be just a matter of passing this parameter with the right value when fetching the file.

amitsaha:
@mger are you planning to continue the work on this? Else i will jump in.

mger:
@amitsaha Feel free to jump in. I had to fly back from PyCon today, and I am not sure when I'll be able to get around to the second part.

amitsaha:
Sounds good, I will resume the work then.

amitsaha:
Couple of things:

- Should we add a checksum verification for `pyodide.js`?
- Also, as mentioned in the PR, the node-fetch module doesn't support integrity, but 17.5 has added `fetch()` as experimental. I guess i should create a separate issue for this.

rth:
Closing as resolved. Thanks again @amitsaha !

PR #2455, opened 2022-05-02
Add SHA-256 hash to packages.json entries

mger:
### Description

This PR adds a SHA-256 hash to the entries in the `packages.json` (the JSON is built in `generate_packages_json()` in `buildall.py`) which is the first step for #2431. This PR also modifies the test for `generate_packages_json()`. For the test, we create dummy package files to test the SHA-256 generation.

### Checklists

- [x] Add / update tests

### DIFF
--- a/pyodide-build/pyodide_build/buildall.py
+++ b/pyodide-build/pyodide_build/buildall.py
@@ -434,8 +434,8 @@ def build_from_graph(pkg_map: dict[str, BasePackage], outputdir: Path, args) -> 
 def _generate_package_hash(full_path: Path) -> str:
     sha256_hash = hashlib.sha256()
     with open(full_path, "rb") as f:
-        for byte_block in iter(lambda: f.read(4096), b""):
-            sha256_hash.update(byte_block)
+        while chunk := f.read(4096):
+            sha256_hash.update(chunk)
     return sha256_hash.hexdigest()
 
 
--- a/pyodide-build/pyodide_build/buildall.py
+++ b/pyodide-build/pyodide_build/buildall.py
@@ -5,6 +5,7 @@ Build all of the packages in a given directory.
 """
 
 import argparse
+import hashlib
 import json
 import os
 import shutil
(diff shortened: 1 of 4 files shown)


pyodide-build/pyodide_build/buildall.py
```diff
@@ -430,7 +431,15 @@ def builder(n):
     )
 
 
-def generate_packages_json(pkg_map: dict[str, BasePackage]) -> dict:
+def generate_package_hash(full_path: Path) -> str:
```
rth, line 434: Let's make this private,

```suggestion
def _generate_package_hash(full_path: Path) -> str:
```

pyodide-build/pyodide_build/buildall.py
```diff
@@ -430,7 +431,15 @@ def builder(n):
...
+    sha256_hash = hashlib.sha256()
+    with open(full_path, "rb") as f:
+        for byte_block in iter(lambda: f.read(4096), b""):
```
rth, line 437: Maybe the first version in https://stackoverflow.com/a/59056837/1791279 is slightly more readable

pyodide-build/pyodide_build/tests/test_buildall.py
```diff
@@ -22,12 +23,18 @@ def test_generate_dependency_graph():
     assert pkg_map["beautifulsoup4"].dependents == set()
 
 
-def test_generate_packages_json():
+def test_generate_packages_json(tmp_path):
+    # Set up directory to store dummy package files for SHA-256 hash verification
+    if not os.path.exists(tmp_path):
+        os.makedirs(tmp_path)
```
rth, line 29:
```suggestion
    """Set up directory to store dummy package files for SHA-256 hash verification"""
```
I think it's already created by default, so you don't need to do this.
Otherwise, you since it's a pathlib object you can use `tmp_path.mkdir(exist_ok=True)`

pyodide-build/pyodide_build/tests/test_buildall.py
```diff
@@ -22,12 +23,18 @@ def test_generate_dependency_graph():
...
     pkg_map = buildall.generate_dependency_graph(PACKAGES_DIR, {"pkg_1", "pkg_2"})
     for pkg in pkg_map.values():
         pkg.file_name = pkg.file_name or pkg.name + ".file"
+        # Write dummy package file for SHA-256 hash verification
+        with open(os.path.join(tmp_path, pkg.file_name), "w") as f:
```
rth, line 34:
```suggestion
        with open(tmp_path / pkg.file_name, "w") as f:
```

mger: Did not know this syntactic sugar exists! Sweet.

pyodide-build/pyodide_build/tests/test_buildall.py
mger: You're right, the directory is already created.

pyodide-build/pyodide_build/buildall.py
mger: Agreed - I prefer that too.

rth (approved):
Could you please add a changelog entry to https://github.com/pyodide/pyodide/blob/main/docs/project/changelog.md . Otherwise LGTM. Thanks a lot for doing this work @mger !

Issue #2430, opened 2022-05-01
Python wrappers for some browser APIs


hoodmane:
## Problem

In order to use `setTimeout` from JavaScript, we need to use `create_once_callable`:
Wont work:
```py
from js import setTimeout
async def test():
	await sleep(0.1)
	print("hi!")
setTimeout(test, 100)
```

Works:
```py
from js import setTimeout
async def test():
	await sleep(0.1)
	print("hi!")
setTimeout(create_once_callable(test), 100)
```

## 🚀 Feature

Python wrappers for `setTimeout`, `cancelTimeout`, `setInterval`, `cancelInterval`, `requestAnimationFrame`, `cancelAnimationFrame`
Possibly also `addEventListener` and `removeEventListener`.
These should automatically manage PyProxy lifetimes.

See:
https://github.com/pyodide/pyodide/blob/3f9ae2ed1e1aee688251af86a4dc23d3d867740c/packages/matplotlib/src/browser_backend.py#L18

lesterfan:
I'm going to try working on this!

hoodmane:
Closed by #2456.

PR #2456, opened 2022-05-02
Add browser API wrappers

lesterfan:
### Description

This adds Python wrappers for `setTimeout`, `clearTimeout`, `setInterval`, `clearInterval`, `addEventListener` and `removeEventListener` described in https://github.com/pyodide/pyodide/issues/2430.

### Checklists

- [x] Add / update tests


hoodmane:
@lesterfan I fixed the tests in chrome and firefox, they should run locally for you. Node still can't find `clearInterval`, etc.

lesterfan:
@hoodmane I just made the change to implement these wrappers in Python using `elt.js_id` for `add_event_listener`/`remove_event_listener` and fixed the resource leaks in my tests. As discussed, I'll leave fixing the imports in node to you. Let me know if there's something else I should do; I'm happy to help however I can to get this merged!

hoodmane:
@lesterfan I fixed the node tests. Turns out in node, `setTimeout` returns an object with some information rather than an int. So there needs to be some special handing for this.

lesterfan:
Ah, I see. I just pushed a change which (I think) will handle this for node. Locally, the node tests time out at `test_clear_timeout_destroyable_noop` and I'm not sure why, especially since the first two tests using `set_timeout` and `clear_timeout` pass.

src/py/pyodide/_browser_apis.py
```diff
@@ -53,7 +54,8 @@ def wrapper():
 DUMMY_DESTROYABLE = Destroyable()
 
 
-def clear_timeout(id: int):
+def clear_timeout(timeout_retval: int | JsProxy):
+    id = timeout_retval if isinstance(timeout_retval, int) else timeout_retval.js_id
     clearTimeout(id)
```
hoodmane, line 59: This needs to be `clearTimeout(timeout_retval)`.

src/py/pyodide/_browser_apis.py
```diff
@@ -0,0 +1,79 @@
...
+
+def set_interval(callback: Callable[[], None], interval: int) -> int:
+    id = -1
+
+    def wrapper():
+        nonlocal id
+        callback()
+        TIMEOUTS.pop(id, None)
+
+    callable = create_once_callable(wrapper)
```
hoodmane, line 71: I'm pretty sure you will need the same change for `setInterval` and `clearInterval`.

lesterfan: Ah, good point. I also realized I forgot to add tests for those two functions as well.

lesterfan:
@hoodmane When adding tests for `set_interval` I ran into this error when testing: ```
pyodide.asm.js:10 Uncaught Error: OnceProxy can only be called once
    at wrapper (pyodide.asm.js:10:246569)
```. Any ideas on how to go about fixing this?

hoodmane:
> Any ideas on how to go about fixing this?

We need to use `create_proxy` instead of `create_once_callable` in `set_interval`.

hoodmane:
Actually, `set_interval` and `clear_interval` should be like `add_event_listener` and `remove_event_listener` not like `set_timeout` and `clear_timeout`.

lesterfan:
Sounds good; I just made the change to `set_interval` and `clear_interval`.

hoodmane:
Okay, code looks good and CI passes. Could you add a changelog entry and docstrings for the new functions? Then we will merge this.

hoodmane:
Thanks @lesterfan!

Issue #347, opened 2019-03-18
Loading pandas is quite slow


wlach:
I think we discussed this a few months back but I can't remember what the resolution was.

pandas seems very slow to load. I don't think it's the network connection (at least here in Toronto, the pyodide assets download almost instantly, even when not cached, but rather something that's bound on the CPU. I noticed that this problem was much less pronounced on jaredkerim's new macbook on Friday (whereas on my machine it takes 20+ seconds to initialize pandas, and also pops up the slow script dialog). 

I feel like we should spend some time resolving this (if possible), as I think pandas is probably *the* library many people would like to use inside an iodide notebook. I am ok with stopgap solutions that don't solve the fundamental issue but move the ball forward.

vadimkantorov:
Maybe it's pandas fault: https://alpha.iodide.io/notebooks/222/ running hangs at: 
`Loading pandas, pytz, python-dateutil, numpy, matplotlib, pyparsing, kiwisolver, cycler`

rth:
@vadimkantorov Which browser (and versions) are you using? As far as I know, pandas works erratically with Chrome (https://github.com/iodide-project/pyodide/issues/128) and the corresponding tests are currently [marked as a known failure](https://github.com/iodide-project/pyodide/blob/875df76b2fb0d1dde9da26a421fd7fcbbbe3a3e5/test/packages/test_pandas.py#L7). Though maybe the situation has evolved since.

vadimkantorov:
Ah, it's unfortunate the hanging notebook is suggested as a demo in [README](https://github.com/iodide-project/pyodide/#pyodide).

I use a Chrome for Windows, help says: `Google Chrome is up to date. Version 73.0.3683.103 (Official Build) (64-bit)`.

vadimkantorov:
Though running `import pandas` after executing https://alpha.iodide.io/notebooks/300/ works fine for me. So it can be something else causes the hang

vadimkantorov:
When I run `import pandas` from a code cell, everything hangs. When I run it from the REPL, it hangs a bit an typing is unresponsive for a few seconds, but then it works fine.

vadimkantorov:
(filed a separate issue https://github.com/iodide-project/pyodide/issues/378)

wlach:
Some improvements: on Firefox nightly, pandas now only takes about 6 seconds to load:

https://definitely-staging.iodide.io/notebooks/44/

Still a little slow for my liking.

rth:
[...]

PR #2457, opened 2022-05-02
Use emscripten's -Oz option for pandas

wlach:
We found in #347 that one of the main culprits of slow load times
for pandas was WASM compilation times. Let's try using emscripten's
[-Oz optimization](https://emscripten.org/docs/tools_reference/emcc.html#emcc-oz) just for this package, which seems to reduce
the size of the wheel from 5.0M -> 4.4M on my machine.

### DIFF
--- a/docs/project/changelog.md
+++ b/docs/project/changelog.md
@@ -47,6 +47,9 @@ substitutions:
 
 - {{ Enhancement }} Pillow now supports WEBP image format {pr}`2407`.
 
+- Pandas is now compiled with `-Oz`, which significantly speeds up loading the library
+  on Chrome {pr}`2457`
+
 - New packages: opencv-python v4.5.5.64 {pr}`2305`, ffmpeg {pr}`2305`, libwebp {pr}`2305`,
   h5py, pkgconfig and libhdf5 {pr}`2411`, bitarray {pr}`2459`
 
--- a/packages/pandas/meta.yaml
+++ b/packages/pandas/meta.yaml
@@ -9,8 +9,8 @@ source:
 build:
   cflags:
     -Werror=implicit-function-declaration -Werror=mismatched-parameter-types
-    -Werror=mismatched-return-types
-  ldflags: --Wl,--fatal-warnings
+    -Werror=mismatched-return-types -Oz
+  ldflags: --Wl,--fatal-warnings -Oz
 requirements:
   run:
     - distutils


rth:
Thanks @wlach ! So we should get the benchmarks results (in the `benchmark` CI job) once the CI complete.

wlach:
Using this benchmark, I'm seeing a ~3% improvement 

```py
import time
def bench():
   x = time.time()
   import pandas
   print(time.time() - x)
 
bench()
```

Four runs with -O2: 2.483, 2.356, 2.644, 2.496
Four runs with -Oz: 2.388, 2.332, 2.382, 2.569

Definitely not mind-blowing but may still be worthwhile.

wlach:
Results in CI:

* Before: `native: nan  firefox: 6.036662  chrome: 17.138263`
* After: `native: nan  firefox: 5.218940  chrome: 9.989785`

We don't have any pandas benchmarks right now, but I feel like this is almost certainly worth applying. That's a huge speedup for Chrome (I forgot to mention that my earlier tests were for Firefox on a fast machine)

rth (commented):
Thanks! Could you time some simple pandas function before/after (maybe read_csv?), just to make sure there is no major performance regression. 

Otherwise LGTM, thanks. We can probably extend this to all packages, by passing the corresponding flag at build time in the future.

rth:
Might be worth adding a changelog entry to `docs/project/changelog.md` as well

wlach:
> Thanks! Could you time some simple pandas function before/after (maybe read_csv?), just to make sure there is no major performance regression.

Yup, in progress! I'll do this as a seperate PR so we can make a comparison.
 
> Otherwise LGTM, thanks. We can probably extend this to all packages, by passing the corresponding flag at build time in the future.

Yeah makes sense, we'll want to compare benchmarks before enabling this for numpy I imagine.

ryanking13:
Thanks for working on this @wlach!

> Results in CI:
> 
> * Before: `native: nan  firefox: 6.036662  chrome: 17.138263`
> * After: `native: nan  firefox: 5.218940  chrome: 9.989785`
> 
> We don't have any pandas benchmarks right now, but I feel like this is almost certainly worth applying. That's a huge speedup for Chrome (I forgot to mention that my earlier tests were for Firefox on a fast machine)

It looks very promising, though the most time-taking part of `pyodide.loadPackage()` is the download time which varies a lot for every run. We probably need a more precise benchmark for download / extract / import.

wlach:
> It looks very promising, though the most time-taking part of `pyodide.loadPackage()` is the download time which varies a lot for every run. We probably need a more precise benchmark for download / extract / import.

Aren't the selenium tests loading pyodide from the local network? In that case I'd expect the download times to be pretty small and deterministic.

rth:
Yes, I agree. I think I tried checking if the http.server is a bottleneck in local benchmarking/testing and it isn't.

ryanking13:
My experience on the `loadPackage` benchmark was quite unstable, but I can be wrong. I'll recheck the point of the bottleneck.

wlach:
Rebased on main, will check pandas benchmarks when CI done.

wlach:
Before:

```
 -  load pandas
    native: nan  firefox: 5.915973  chrome: 14.927018  
...
 -  join_df
    native: 0.125844  firefox: 0.042333  firefox(w/ ib): 0.041000  chrome: 0.042100  chrome(w/ ib): 0.041967  
 -  read_csv
    native: 0.013923  firefox: 0.021333  firefox(w/ ib): 0.021667  chrome: 0.019367  chrome(w/ ib): 0.018900  
```

After:

```
 -  load pandas
    native: nan  firefox: 5.372690  chrome: 10.965915  
...
 -  join_df
    native: 0.119750  firefox: 0.039000  firefox(w/ ib): 0.040000  chrome: 0.038567  chrome(w/ ib): 0.037700  
 -  read_csv
    native: 0.012331  firefox: 0.021667  firefox(w/ ib): 0.022000  chrome: 0.019267  chrome(w/ ib): 0.019133  
```

I don't actually trust that this genuinely improved performance on the pandas benchmarks (as `join_df` might imply) but I think we can at least be sure that it didn't degrade it. :smile: 

@rth anything more that needs to happen before we merge?

rth (approved):
Thankls @wlach ! Yeah, actually maybe we want to compile everything with Oz, assuming benchmarks are OK.

Issue #1317, opened 2021-03-10
Micropip doesn't add cookie in the GET request


124bit:
Like "withCredentials=true" in axios. However, maybe it is not a bug but a feature request.

124bit:
It happened during CORSE

rth:
Could you provide more details on your use case and what you are trying to achieve? 

We are using  [XMLHttpRequest for making get requests](https://github.com/iodide-project/pyodide/blob/958724101a9cd6a1181feadcec811f26153ff828/packages/micropip/micropip/micropip.py#L38) so I imagine setting there [`withCredentials` to `true`](https://developer.mozilla.org/en-US/docs/Web/API/XMLHttpRequest/withCredentials) could be a solution?

We could add a `with_credentials=False` parameter to `micropip.install`. A PR would be welcome.

124bit:
The usecase is when you try to get whl from other domain/port and it is available just in case you are authenticated, therefore you need to pass cookies with the request.
I think the proposed solution is good.
Sorry, I can’t do PR(

amitsaha:
I will take a stab at this

rth:
Resolved by @amitsaha in #2458, thanks!

PR #2458, opened 2022-05-03
ENH Add credentials kwarg to `micropip.install`

amitsaha:
### Description

Fix for #1317 

Add a new optional `credentials` argument to `micropip.install()` function. The value passed will be directly passed through to the underlying `fetch()` JS function call.

### Checklists

- [x] Add / update tests

### Manual testing

Build `pyodide` and other packages and serve it from `http://192.168.1.105:9000`.

Setup a HTTP server on `10.0.254.2:8000` to simulate a PyPi server:

```
import http.cookies
from http.server import BaseHTTPRequestHandler, HTTPServer


class MyRequestHandler(BaseHTTPRequestHandler):
        def do_GET(self):
            print(self.headers)
            body = b"hello world!"
            self.send_response(200)
            self.send_header("Content-type", "text/plain; charset=utf-8")
            self.send_header("Content-Length", str(len(body)))
            self.send_header("Access-Control-Allow-Origin", "http://192.168.1.105:9000")
            self.send_header("Access-Control-Allow-Credentials", "true")

            cookie = http.cookies.SimpleCookie()
            cookie['a_cookie'] = "Cookie_Value; SameSite=None"
        
            for morsel in cookie.values():
                self.send_header("Set-Cookie", morsel.OutputString())
            self.end_headers()
            self.wfile.write(body)


def run(server_class=HTTPServer, handler_class=MyRequestHandler):
    server_address = ('0.0.0.0', 8000)
    httpd = server_class(server_address, handler_class)
    httpd.serve_forever()

run()
```

Make a manual request to `http://10.0.254.2:8000` to set a cookie. 

Test HTML file to load `pyodide` from the above URL, `192.168.1.105:9000`:

```<html>
  <head>
    <meta charset="utf-8" />
  </head>
  <body>
    <script
      type="text/javascript"
      src="http://192.168.1.105:9000/pyodide.js"
    ></script>
      async function main() {
        import micropip
        await micropip.install('http://10.254.0.2:8000/sympy-1.10.1-py3-none-any.whl', credentials='include')
	`);
	
        from sympy import Symbol, factor, expand
        x = Symbol('x')
        y = Symbol('y')
        expr = x**2 - y**2
        f = factor(expr)
        print(f)
        print(expand(f))
      `);
      }
      main();
    </script>
  </body>
</html>
```

When `credentials='include'` is specified, the cookie set by `10.254.0.2` is sent with the request (manually verified)
When not specified, the cookie is not sent.


hoodmane:
It looks like some of the lint errors are unrelated to this patch?

hoodmane:
I think we need to turn off `pre-commit.ci` for now and revert the commit it made here. It doesn't seem to be working.

ryanking13:
> I think we need to turn off `pre-commit.ci` for now and revert the commit it made here. It doesn't seem to be working.

Yes, we can reconfigure it after the sprint ends.

henryiii:
The problem is that the pyproject.toml was deleted here (by mistake I hope). That must be reverted, it’s not pre-commit’s fault, it’s just a side effect of not having configuration anymore.

amitsaha:
> The problem is that the pyproject.toml was deleted here (by mistake I hope). That must be reverted, it’s not pre-commit’s fault, it’s just a side effect of not having configuration anymore.

That was a mistake, yes :( I have fixed that.

henryiii:
427d6d67c9bbd16a58bfd7911d2b011ad4ac18d4 Would have been made automatically by pre-commit.ci

packages/micropip/src/micropip/_micropip.py
```diff
@@ -81,9 +81,12 @@ async def gather(*coroutines):  # type: ignore[no-redef]
...
 
 
-async def _get_pypi_json(pkgname):
+async def _get_pypi_json(pkgname, with_credentials):
     url = f"https://pypi.org/pypi/{pkgname}/json"
-    return json.loads(await fetch_string(url))
+    extra_kwargs = dict()
+    if with_credentials:
+        extra_kwargs["credentials"] = "include"
+    return json.loads(await fetch_string(url, **extra_kwargs))
```
rth, line 89: Maybe just,
```
async def _get_pypi_json(pkgname, **kwargs):
    url = f"https://pypi.org/pypi/{pkgname}/json"
    return json.loads(await fetch_string(url, **kwargs))
```
we probably don't want to overwrite the value passed by the user, and there could still be a difference between same-origin and ommit?

packages/micropip/src/micropip/_micropip.py
```diff
@@ -371,7 +402,11 @@ def find_wheel(
 del _PackageManager
 
 
-def install(requirements: str | list[str], keep_going: bool = False):
+def install(
+    requirements: str | list[str],
+    keep_going: bool = False,
+    with_credentials: bool = False,
```
rth, line 408: Maybe still have,
```
credentials : Optional[str] = None
```
and by default not pass them at all to fetch, otherwise pass the value given by the user?

Also please update the docstring accordingly below

packages/micropip/test_micropip.py
```diff
@@ -458,3 +458,34 @@ def test_list_loaded_from_js(selenium_standalone_micropip):
         `);
         """
     )
+
+
+def test_install_with_credentials(selenium_standalone_micropip):
+    selenium = selenium_standalone_micropip
+    selenium.run_js(
```
rth, line 465: I think you should be able to use run_in_pyodide to reduce the nesting
```suggestion
@run_in_pyodide(standalone=True)
def test_install_with_credentials():
    import micropip
    ...
```

rth (commented):
Thanks a lot @amitsaha ! Could you please also add a changelog entry to docs/project/changelog.md

rth (approved):
Thanks @amitsaha !

Issue #2438, opened 2022-05-02
CI Automate release process more


ryanking13:
We have automated most of the release process:

- Publishing `pyodide_build` to PyPI.
- Deploying build results to CDN.
- Push prebuilt docker image.

But there exist a few more things that need to be automated to prevent human errors.

- Replacing version names (CDN links) in the documentation: we can probably use [sphinx substitution syntax](https://sublime-and-sphinx-guide.readthedocs.io/en/latest/reuse.html).
- Publishing Pyodide JS package to NPM: we should add this to our CI jobs.

hoodmane:
Also replacing version numbers in other places (`pyproject.toml`, `__init__.py`, ...)

rth:
Yes, it would be great.

> Publishing pyodide_build to PyPI.

BTW, it looks like https://pypi.org/project/pyodide-build/ wasn't deployed for 0.20.0

ryanking13:
> BTW, it looks like https://pypi.org/project/pyodide-build/ wasn't deployed for 0.20.0

Yeah right, we forgot to rename the pyodide_build version and published `0.20.0.dev0`.

andreysmelter:
I am going to try to take this first issue first part during PyCon sprints: 
* Replacing version names (CDN links) in the documentation: we can probably use [sphinx substitution syntax](https://sublime-and-sphinx-guide.readthedocs.io/en/latest/reuse.html).

ryanking13:
Closing as resolved via #2460, #2587 and #2776.

PR #2460, opened 2022-05-03
Add bump2version

andreysmelter:
WIP PR: 

Partially addresses [issue 2438](https://github.com/pyodide/pyodide/issues/2438)
by bumping up source code files.

Next steps:
- automate version bumpup in docs/*.md files

### DIFF
--- a/docs/development/maintainers.md
+++ b/docs/development/maintainers.md
@@ -24,17 +24,13 @@ the latest release branch named `stable` (due to ReadTheDocs constraints).
    - `docs/project/about.md` (the Zenodo citation),
    - `docs/development/building-from-sources.md`,
    - `docs/usage/downloading-and-deploying.md`,
-   - Bump version in source code files by running `bump2version` command, for example:
+   - Bump version in source code files by running `bump2version` command, for example,
 
      ```bash
-     bump2version --current-version 0.20.0 minor --dry-run --verbose
+     bump2version minor
      ```
 
-     and
-
-     ```bash
-     bump2version --current-version 0.20.0 minor
-     ```
+     check that the diff is correct with `git diff` before committing.
 
    After this, try using `ripgrep` to make sure there are no extra old versions
    lying around e.g., `rg -F "0.18"`, `rg -F dev0`, `rg -F dev.0`.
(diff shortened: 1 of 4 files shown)


ryanking13:
Thanks, @andreysmelter! `bump2version` looks like very cool stuff.

I have a question. For alpha/dev versions, Python packages and JS packages have different version schemes. For example, we use `0.20.0a1` for Python package while `0.20.0-alpha.1` for JS package.

Can we deal with this in a single command? For example, it would be great if we can do this in a config file:

```cfg
# inside .bumpversion.cfg  / setup.cfg
[bumpversion]
current_version = 0.20.1
new_version_python = 0.20.1a1
new_version_js = 0.20.1-alpha.1
```

hoodmane:
It might be better to use `setuptools-scm` since it has a Python interface and so we could do whatever complicated stuff we need to in a Python script rather than having to have a config file.

rth:
> It might be better to use setuptools-scm since it has a Python interface and so we could do whatever complicated stuff we need to in a Python script rather than having to have a config file.

Generally, I like bump2version for Python packages, but I'm not sure that's the best option here. setuptools-scm  is not really going to work on non Python files (e.g. run_docker) without significant development  where you provide it with some list of lines to replace and how to do that, which is what bump2version does.
Also e.g. the format `{next_version}.dev{distance}+{scm letter}{revision hash}` of setuptools-scm is not really appropriate for
 - js package
 - docker image
 - for the docs when pushing on the stable branch fixes after the release

So I think keeping things simple with a simple setuptools-scm config is a start (and is already way better than what we had before). Though yes, this means that the alpha versions for the JS package need manual adjustment. If we want to go further later, we can always create a custom Python script calling  bump2version in more custom ways. 
BTW, it's also already parsing the latest git tag https://github.com/c4urself/bump2version/blob/bc95374d0dff73e610bed520846f88859815e532/bumpversion/cli.py#L231 though I haven't looked in detail what it does with it.

rth:
Edited the above comment to avoid confusion between the two tools :)

andreysmelter:
@rth I can take a look at this next week and continue iterating on this PR. Thanks for the comments and ideas!

rth (approved):
OK, I can confirm this works as expected. I pushed a few extra minor improvements.  But yeah anyway one has to look at the git diff and often make manual adjustments before committing/tagging. Still, it's much better than the system we have now.

Merging this as a partial solution, and we can always improve it in follow up MRs (in particular to also do this in the docs with sphinx). Thanks, @andreysmelter !

PR #2462, opened 2022-05-03
allow running mkpkg without npx installed

megies:
### Description

Allow `mkpkg` to run (what is the proposed command for generating a yaml skeleton for a new package) when `npx` is not installed.

Seems like a non vital step, since the expected yaml is output without it) and I think it should not hold up people that don't have it installed or make them look into how to install `npx` or how vital it is.

### Checklists

- [ ] Add / update tests

None of the checklist items apply, I think.

### DIFF
--- a/pyodide-build/pyodide_build/mkpkg.py
+++ b/pyodide-build/pyodide_build/mkpkg.py
@@ -145,7 +145,11 @@ def make_package(
         raise MkpkgFailedException(f"The package {package} already exists")
 
     yaml.dump(yaml_content, meta_path)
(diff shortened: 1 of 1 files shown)


rth:
Thanks!

Yes, I think we should overall remove all `npx` and `pip install` calls for base build dependencies from the makefile, and instead explicitly install needed dev tools beforehand (maybe excluding dependencies when building packages).

rth (approved):
> None of the checklist items apply, I think.

BTW, in the future don't hesitate to remove them if they don't apply.

PR #2468, opened 2022-05-03
Add browser polyfills for webpack compilation.

jasongrout:
### Description

This adds the necessary polyfills, according to the spec at https://github.com/defunctzombie/package-browser-field-spec to enable webpack to ignore any imports of these modules when packing files for the browser.

To be clear, the "polyfills" added here are to just ignore the imports, since the imports shouldn't happen in the browser.

With this change, we can successfully webpack a simple js file that imports pyodide:
```js
import { loadPyodide } from "pyodide";
```
### Checklists

### DIFF
--- a/docs/project/changelog.md
+++ b/docs/project/changelog.md
@@ -14,6 +14,8 @@ substitutions:
 
 ## Unreleased
 
+- {{ Fix }} We now tell packagers (e.g., Webpack) to ignore npm-specific imports when packing files for the browser. {pr}`2468`
+
 - {{ Fix }} micropip now correctly handles package names that include dashes
   {pr}`2414`
 
--- a/src/js/compat.ts
+++ b/src/js/compat.ts
@@ -29,12 +29,12 @@ export async function initNodeModules() {
     return;
   }
   // @ts-ignore
-  nodePathMod = (await import(/* webpackIgnore: true */ "path")).default;
-  nodeFsPromisesMod = await import(/* webpackIgnore: true */ "fs/promises");
+  nodePathMod = (await import("path")).default;
+  nodeFsPromisesMod = await import("fs/promises");
   // @ts-ignore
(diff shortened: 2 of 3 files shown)


rth (commented):
That's great, thanks a lot for doing this @jasongrout !

Can we then remove the webpack related comments in https://github.com/pyodide/pyodide/blob/fda600e519414109b4b1ebb93b09723e08e29dea/src/js/compat.ts#L32 (and similar lines in that file)?

jasongrout:
> Can we then remove the webpack related comments in

Great question! Yes, I think so. I'll try it.

rth:
Hmm, looks like the webpack example in CI fails,
```
ERROR in ./node_modules/pyodide/load-pyodide.js 74:20-32
Module not found: Error: Can't resolve 'vm' in '/root/repo/pyodide-webpack-example/node_modules/pyodide'
```

.circleci/config.yml
```diff
@@ -232,14 +232,14 @@ jobs:
...
-      #       npm ci
-      #       cp ../src/js/*.js node_modules/pyodide/
-      #       npx webpack
+      - run:
+          name: check if webpack cli works well with load-pyodide.js
+          command: |
+            git clone https://github.com/pyodide/pyodide-webpack-example.git
+            cd pyodide-webpack-example
+            npm ci
+            cp ../src/js/*.js node_modules/pyodide/
```
rth, line 241: So I think the issue was that that package specifies pyodide 0.18.0 https://github.com/pyodide/pyodide-webpack-example/blob/main/package.json#L18
while we want to use it with latest pyodide. So this was a workaround for that, but after migrating to typescript, I don't think that's correct anymore?

rth:
It's also possible to merge this without running that test, and try to fix the test in a separate PR, as you prefer.

jasongrout:
> It's also possible to merge this without running that test, and try to fix the test in a separate PR, as you prefer.

I went ahead and reverted back to having that test commented out, then.

PR #2471, opened 2022-05-03
Update typescript target to es2017

jasongrout:
### Description

This will generate code that uses native async/await. Safari 14.1, Chrome, and Firefox appear to have [support for es2021](https://caniuse.com/?feats=mdn-javascript_builtins_string_replaceall,mdn-javascript_builtins_promise_any,mdn-javascript_builtins_weakref,mdn-javascript_operators_logical_or_assignment,mdn-javascript_operators_logical_and_assignment,mdn-javascript_operators_logical_nullish_assignment,mdn-javascript_grammar_numeric_separators): 

### Checklists

### DIFF
--- a/docs/project/changelog.md
+++ b/docs/project/changelog.md
@@ -14,6 +14,8 @@ substitutions:
 
 ## Unreleased
 
+- {{ Enhancement }} Update Typescript target to ES2017 to generate more modern Javascript code. {pr}`2471`
+
 - {{ Fix }} micropip now correctly handles package names that include dashes
   {pr}`2414`
 
--- a/src/js/package.json
+++ b/src/js/package.json
@@ -105,7 +105,7 @@
   "tsd": {
     "compilerOptions": {
       "lib": [
-        "ES2018",
+        "ES2017",
         "DOM"
       ]
     }
--- a/src/js/tsconfig.json
+++ b/src/js/tsconfig.json
@@ -1,10 +1,9 @@
 {
   "$schema": "http://json.schemastore.org/tsconfig",
   "compilerOptions": {
-    "lib": ["ES2021", "DOM"],
     "outDir": "../../dist",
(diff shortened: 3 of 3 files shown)


rth:
Thanks @jasongrout ! So what's the practical implication for users to switch to ES2021 from ES6 ? Because right now Pyodide works to some extent on Safari 13 and as far as I understand this would stop version for early 13.X versions.

Also https://pyodide.org/en/stable/usage/index.html#supported-browsers are probably not very accurate, but we would need to update them if we do this. 
So there is potentially some compatibility cost (even Chrome 85 released on August 25, 2020 is kind of recent, and can probably still be found in some corporate environment), and unless there are very clear advantages for users, maybe it would be worth waiting a bit more before updating? WDYT?

jasongrout:
I think updating to at least es2017 gives you much cleaner generated code. Since your lib was es2021, I updated to es2021, but I'll push an update moving that back to a much more conservative es2017 target. That's what we use in JupyterLab, for example.

jasongrout:
> Also https://pyodide.org/en/stable/usage/index.html#supported-browsers are probably not very accurate, but we would need to update them if we do this.

It looks like Safari 11 [supports es2017](https://caniuse.com/async-functions,object-values,object-entries,mdn-javascript_builtins_object_getownpropertydescriptors,pad-start-end,mdn-javascript_grammar_trailing_commas_trailing_commas_in_functions), so I think we're good.

rth:
es2017 sounds very reasonable indeed! Thanks!

rth (approved):
Just one last thing @jasongrout, could you please add a changelog entry to https://github.com/pyodide/pyodide/blob/main/docs/project/changelog.md ideally in a commit  with a commit message containing the ` [skip ci] ` string to avoid running the build. Thanks again!

PR #2475, opened 2022-05-03
Remove pandas docstrings

wlach:
Local patch of https://github.com/pandas-dev/pandas/pull/46934

I don't expect this to make a user-noticeable improvement, but it seems
worth doing regardless. Locally I see a reduction from 4546988 for the
whole wheel to 4545234 (so just under 1k) although some of the .so's
inside are a few kB smaller.


rth:
Thanks! I agree it's  very worthwile investigating cython options to reduce code size. I'm not sure about this one though as it's hard to be sure those are really private and not exposed somewhere (https://github.com/pandas-dev/pandas/pull/46934#issuecomment-1116760330) and the gain is relatively small.

ryanking13:
Thanks for the investigation @wlach!

PR #2476, opened 2022-05-04
Experimental removal of pandas c-extension autoloading

wlach:
It's not certain that this will improve things, but it can't hurt to
try. See #347.

### DIFF
--- a/packages/pandas/meta.yaml
+++ b/packages/pandas/meta.yaml
@@ -6,6 +6,7 @@ source:
   sha256: 92bc1fc585f1463ca827b45535957815b7deb218c549b7c18402c322c7549a12
   patches:
     - patches/fix_json_signature.patch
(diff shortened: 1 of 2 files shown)


wlach:
I'm not seeing much difference here in local testing, but I figure it can't hurt to give it a try in CI (which seems to be much lower powered).

hoodmane:
I don't think this sort of thing would help much because we have to do all the work of loading the binary module up front anyways. I think it is impossible to avoid this with synclink or similar because it is impossible to serialize a Wasm Module into an ArrayBuffer.

wlach:
> I don't think this sort of thing would help much because we have to do all the work of loading the binary module up front anyways. I think it is impossible to avoid this with synclink or similar because it is impossible to serialize a Wasm Module into an ArrayBuffer.

Yup, let's close this.

PR #2478, opened 2022-05-04
CI Don't cache dist/node_modules

rth:
For some reason, there is a `dist/node_modules` now after the build. This takes a long time to be cached by the CI (in particular because it includes chromium). 

So I remove it as a quick fix to improve CI runtimes, but we should definitely figure out why it's there in the first place and get rid of it.

PR #2480, opened 2022-05-04
Remove npm-link step

hoodmane:
It's causing trouble.


rth:
Yeah, that's probably better by https://github.com/pyodide/pyodide/pull/2478

rth (approved):

PR #2483, opened 2022-05-04
explore adding obspy

megies:
### Description

This PR explores whether [obspy](https://github.com/obspy/obspy) can be added to pyodide/pyscript. **It is mostly for information / documentation of WIP / as a draft right now.**

With a minimal recipe stub, after building the compiled wheel locally, the current state is..

 - C extensions get built
 - tried one of our really simple C extension (inside the `console.html` in Firefox) and it worked
 - tried one of our complex C extension and it crashed, saying something about GIL
 - quite a lot of obspy functionality can't be used at this state because of `requests`, either..
   - because `requests` is imported at the top of one of our submodules, or..
   - because a call goes through our entry point / plugin setup with `setuptools`, and as soon as we hit that, `setuptools` complains about `requests` missing which is listed as a hard dependency in our `setup.py`

So besides being unsure how much more work it would be to get all C extensions to work, the main caveat seems `requests`. So for now this is mostly "toying around" at the moment and I'm not sure how much more work it would be to get it done properly, both here in the recipe (for C extensions etc) and upstream/in patches (to workaround `requests` which we use quite heavily) which likely would take some major workarounds.

#### Side comments

It took me some time to figure out how to go forward with testing the locally built compiled wheel. I was using [this docs page](https://pyodide.org/en/stable/development/new-packages.html#creating-a-pyodide-package) mostly, but I think what it is lacking is how to test the locally built package. Maybe the following could be outlined in that page?

 - download the current build of pyodide and unpack locally
 - put the compiled wheel in there
 - add it manually in the `packages.json`
 - serve the folder locally with `python -m hhtp.server`
 - navigate browser to `http://localhost:8000/console.html` (or whatever port is stated on CL)
 - import package, play around etc..

What I still didn't figure out is how to try the locally built wheel in `pyscript`, I tried replacing in a downloaded `pyscript.js` the part that loads in `pyodide` with `http://localhost:8000/pyodide.js` but I didn't get it to work (some cross origin errors I believe).

### Checklists

- [ ] Add / update tests

pyodide_obspy

### DIFF
--- a/packages/obspy/meta.yaml
+++ b/packages/obspy/meta.yaml
@@ -1,0 +1,17 @@
+package:
+  name: obspy
+  version: 1.3.0
+
+source:
+  url: https://files.pythonhosted.org/packages/b5/40/cebd39066edb6d91b7fddeb676a788ac89c59d3e31e820c2b64983b368d8/obspy-1.3.0.tar.gz
+  sha256: 932eb8af44542d605184a07f518f4839c7f42c22939fc848a9ddb9d8f7125077
+
+test:
+  imports:
+    - obspy
+
+about:
+  home: https://www.obspy.org
+  PyPI: https://pypi.org/project/obspy
+  summary: ObsPy - a Python framework for seismological observatories.
+  license: GNU Lesser General Public License, Version 3 (LGPLv3)


hoodmane:
> which apparently proabably never can be in pyodide)

This is not quite true. The `quickstart` requests API should not be too hard to shim with synclink or something similar. Some of the advanced apis are impossible.

Ideally your application would separate IO from logic which would allow different IO in weird platforms.

> tried one of our complex C extension and it crashed, saying something about GIL

Can you give me a reproduction on this? I can look into weird package build problems for you (if I get to it...)

megies:
@hoodmane i can add all the run requirements again, i had them in the meta.yaml before, and IIRC they were in there when I built the wheel. I accidently deleted the more advanced `meta.yaml` I had when trying to figure out what to do with the built wheel and because at some point I ended up with a source wheel getting built and I was poking around `make`ing other stuff in the Makefile and did a `git clean -fdx` to clean up the whole checkout.. :grimacing:

megies:
@hoodmane compare https://github.com/conda-forge/obspy-feedstock/blob/main/recipe/meta.yaml

```
requirements:
  run:
    - python
    - setuptools
    - numpy
    - scipy
    - matplotlib
    - lxml
    - sqlalchemy
    #- requests
    - decorator
    # see https://github.com/obspy/obspy/issues/2984#issuecomment-1063161646
    #- libffi <3.4  # [linux64 and py==38]
```

PR #2488, opened 2022-05-04
Add xgboost

hoodmane:
As requested in #2486. Blocked on needs more recent CMake.

### Checklists

- [ ] Add / update tests


hoodmane:
@henryiii any insight into the CMake error here?

henryiii:
I think you need to run `emcmake`. You can capture the various flags emscripten injects to the cmake call - that's all the wrapper does. That's what I was doing previously, at least.

hoodmane:
So maybe what we need to do is add another symlink to pywasmcross and turn cmake into emcmake automagically.

henryiii:
As long as it’s not recursive…

hoodmane:
I think we already have careful path manipulation in pywasmcross to prevent that.

hoodmane:
It seems that xgboost needs a host copy of `libxgboost.so` and the `xgboost` binary. This means it will be really hard to crossbuild because it is using the xgboost c library to do inspection about the configuration that the c library was installed with, so we would need to create a cross build xgboost c library that gives inspection information about the target instead. Maybe one could shim it with a Python script or something but in any case it will be hard.

hoodmane:
Actually, I just need to say `emcmake cmake`. 
But now it's stuck on:
```
Could NOT find Backtrace (missing: Backtrace_LIBRARY Backtrace_INCLUDE_DIR)
```

ryanking13:
You probably don't want to symlink `emcmake`. It breaks our `symlink` strategy.

What `emcmake` does is adding an Emscripten Toolchain file.

https://github.com/emscripten-core/emscripten/blob/c7b74318316d4553687e177f34e8bb352040b8e6/emcmake.py#L34

Inside the toolchain file, it sets compilers to `emcc`, `em++`, ...
So it directly calls `emcc` instead of calling `pywasmcross` symlink of gcc.

```cmake
# inside Emscripten.cmake
# Specify the compilers to use for C and C++
if ("${CMAKE_C_COMPILER}" STREQUAL "")
  set(CMAKE_C_COMPILER "${EMSCRIPTEN_ROOT_PATH}/emcc${EMCC_SUFFIX}" CACHE FILEPATH "Emscripten emcc")
endif()
if ("${CMAKE_CXX_COMPILER}" STREQUAL "")
  set(CMAKE_CXX_COMPILER "${EMSCRIPTEN_ROOT_PATH}/em++${EMCC_SUFFIX}" CACHE FILEPATH "Emscripten em++")
endif()

if ("${CMAKE_AR}" STREQUAL "")
  set(CMAKE_AR "${EMSCRIPTEN_ROOT_PATH}/emar${EMCC_SUFFIX}" CACHE FILEPATH "Emscripten ar")
endif()
```

So when I was building OpenCV, I [copied some parts of Emscripten Toolchain file](https://github.com/pyodide/pyodide/blob/main/packages/opencv-python/cmake/Config.cmake) then provided it through `-DCMAKE_TOOLCHAIN_FILE`.

hoodmane:
Okay @ryanking13 maybe you could take over this PR if you have time? I am pretty sure my CMake wrangling skills are not up to the task here.

ryanking13:
> Okay @ryanking13 maybe you could take over this PR if you have time? I am pretty sure my CMake wrangling skills are not up to the task here.

Sure, I'll look into this in my possible time.

PR #2489, opened 2022-05-04
refactor: use cmake 3.22 from pip

henryiii:
### Description

See https://github.com/pyodide/pyodide/discussions/2486.


hoodmane:
Thanks @henryiii! Could you rebuild the docker image and push it to a dockerhub that you control, then update the image tags? Or I can do that and push to this branch if you don't want to.

henryiii:
Probably best if you do it, I'll be doing a few quick things, but have to prepare for a bunch of stuff for next week. If you can't get to it in the next few days, I can do it mid next week.

henryiii:
The compressed size of the current image is `docker manifest inspect pyodide/pyodide-env:20220411-chrome99-firefox98 | jq '.layers[].size' | jq -s 'add' | numfmt --to iec --suffix B` => `587MB`. New image is `560MB`. So it's noticeable, at least. :)

hoodmane:
Thanks @henryiii!

PR #2491, opened 2022-05-04
[pre-commit-ci-update-config] chore: remove fixed type ignores

henryiii:
This PR targets #2450 with MyPy updates due to fixes added to typeshed.


rth (approved):
Thanks a lot @henryiii !

PR #2492, opened 2022-05-04
Add caterva

aleixalcacer:
### Description

Make `caterva` package available in pyodide.

### DIFF
--- a/packages/caterva/meta.yaml
+++ b/packages/caterva/meta.yaml
@@ -19,3 +19,4 @@ about:
 build:
   script: |
     export SKBUILD_CONFIGURE_OPTIONS='-DCMAKE_BUILD_WITH_INSTALL_RPATH=ON'
+    export CMAKE_ARGS="-DCMAKE_TOOLCHAIN_FILE=$PYODIDE_ROOT/packages/caterva/cmake/Config.cmake"
--- a/packages/caterva/meta.yaml
+++ b/packages/caterva/meta.yaml
@@ -6,6 +6,11 @@ source:
   url: https://files.pythonhosted.org/packages/37/3e/0c2131e98ba327630f47d2db70e11c2416821fd0f44c4ee248c1f6a18951/caterva-0.7.2.tar.gz
   sha256: 109a744715e99e1771575871b9eb6b54d3fcc6de6f3ca56d52289d12c2cb9849
 
+requirements:
+  run:
+    - numpy
+    - ndindex
+
 test:
   imports:
     - caterva
@@ -18,5 +23,6 @@ about:
 
 build:
   script: |
+    export EMSCRIPTEN="$PYODIDE_ROOT/emsdk/emsdk/upstream/emscripten/"
     export SKBUILD_CONFIGURE_OPTIONS='-DCMAKE_BUILD_WITH_INSTALL_RPATH=ON'
     export CMAKE_ARGS="-DCMAKE_TOOLCHAIN_FILE=$PYODIDE_ROOT/packages/caterva/cmake/Config.cmake"
--- a/packages/caterva/meta.yaml
+++ b/packages/caterva/meta.yaml
@@ -1,14 +1,21 @@
 package:
   name: caterva
   version: 0.7.2
+
 source:
   url: https://files.pythonhosted.org/packages/37/3e/0c2131e98ba327630f47d2db70e11c2416821fd0f44c4ee248c1f6a18951/caterva-0.7.2.tar.gz
   sha256: 109a744715e99e1771575871b9eb6b54d3fcc6de6f3ca56d52289d12c2cb9849
+
 test:
   imports:
     - caterva
+
 about:
   home: https://github.com/Blosc/python-caterva
   PyPI: https://pypi.org/project/caterva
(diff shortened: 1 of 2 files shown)


aleixalcacer:
I followed the steps in https://pyodide.org/en/stable/development/new-packages.html but the CI failed. Please, could someone help me to continue? :)

hoodmane:
Error is:
```
CMake Error at caterva/CMakeLists.txt:13 (add_library):
  The install of the caterva_ext target requires changing an RPATH from the
  build tree, but this is not supported with the Ninja generator unless on an
  ELF-based or XCOFF-based platform.  The CMAKE_BUILD_WITH_INSTALL_RPATH
  variable may be set to avoid this relinking step.
```
maybe @ryanking13 or @henryiii can tell us what this means. I don't understand cmake so I am not of much use here.

henryiii:
Nice, I think this is the first scikit-build package to be included. You aren't trying to run from the build directory, so `CMAKE_BUILD_WITH_INSTALL_RPATH` should be safe enough to use. `SKBUILD_CONFIGURE_OPTIONS='-DCMAKE_BUILD_WITH_INSTALL_RPATH=ON'` should do it. At least I'd probably try that first. This is using the built-in PythonExtensions module of scikit-build, which I'm not a fan of, it might be setting up something strangely.

Since this is using CMake, it might be best to use the emcmake trick, too? I can teach scikit-build about emcmake if that helps.

hoodmane:
@ryanking13 was saying that we shouldn't use `emcmake` because it will route around our other toolchain wrappers.

henryiii:
Ahh, yes. So then the toolchain file should be provided.

hoodmane:
Maybe we can symlink cmake and automatically inject the toolchain file?

hoodmane:
@henryiii would you mind adding the needed environment info somewhere around here:
https://github.com/pyodide/pyodide/blob/6fc28c5669dfeb22c5e9544144a35a1f8b655365/pyodide-build/pyodide_build/pywasmcross.py#L111

I don't really understand what needs to be added.

aleixalcacer:
The RPATH error is already fixed. Thanks @henryiii :)

> @henryiii would you mind adding the needed environment info somewhere around here:
> 
> https://github.com/pyodide/pyodide/blob/6fc28c5669dfeb22c5e9544144a35a1f8b655365/pyodide-build/pyodide_build/pywasmcross.py#L111
> 
> I don't really understand what needs to be added.

@hoodmane  I think I should follow what @ryanking13 is saying in https://github.com/pyodide/pyodide/pull/2488#issuecomment-1118021145 in order to set the Emscripten toolchain file. I guess it's a caterva specific issue (not pyodide), but I'm not sure.

aleixalcacer:
> Shouldn't the toolchain be added globally, and then the environment variable be set by default in pywasmcross.py? All CMake packages will need this, I believe. `-DCMAKE_TOOLCHAIN_FILE` could be removed from OpenCV then too.

True, all packages that use CMake are going to need it. It would not be very difficult to implement it, right?

packages/caterva/meta.yaml
```diff
@@ -0,0 +1,28 @@
...
+  home: https://github.com/Blosc/python-caterva
+  PyPI: https://pypi.org/project/caterva
+  summary: Caterva for Python (multidimensional compressed data containers).
+  license: https://opensource.org/licenses/BSD-3-Clause
+
+build:
+  script: |
+    export EMSCRIPTEN="$PYODIDE_ROOT/emsdk/emsdk/upstream/emscripten/"
+    export SKBUILD_CONFIGURE_OPTIONS='-DCMAKE_BUILD_WITH_INSTALL_RPATH=ON'
+    export CMAKE_ARGS="-DCMAKE_TOOLCHAIN_FILE=$PYODIDE_ROOT/packages/caterva/cmake/Config.cmake"
```
henryiii, line 28: `CMAKE_ARGS` will be ignored if you have `SKBUILD_CONFIGURE_OPTIONS`. You can use either one, but the latter overrides the former.

ryanking13:
> @ryanking13 was saying that we shouldn't use `emcmake` because it will route around our other toolchain wrappers.

> Shouldn't the toolchain be added globally, and then the environment variable be set by default in pywasmcross.py? All CMake packages will need this, I believe. `-DCMAKE_TOOLCHAIN_FILE` could be removed from OpenCV then too.

Actually, I think we can use `emcmake` if we properly set `CMAKE_C_COMPILER`, `CMAKE_CXX_COMPILER`, ... to point to our `pywasmcross` symlink. Or we could also make our own toolchain file for Pyodide, which is "a portion" of Emscripten toolchain file.

BTW @henryiii, do you know how to override CMake variables in the command line through e.g. env variable? We often need to override some variables but I don't know the best way to do it. In OpenCV, I just put [everything in the toolchain file](https://github.com/pyodide/pyodide/blob/main/packages/opencv-python/cmake/Config.cmake#L24-L31).

ryanking13:
To fix build error, I think you need to add `ndindex` package through `python -m pyodide_build mkpkg ndindex`.

ryanking13:

Issue #2503, opened 2022-05-05
pyodide_build buildpkg does not install Cython as a build dependency when it is spell with a lower case c


ocefpaf:
## 🐛 Bug

When trying to build [cftime](https://github.com/Unidata/cftime) the isolated env does not install cython.

### To Reproduce

`python -m pyodide_build buildpkg packages/cftime/meta.yaml` on [this meta.yaml](https://gist.github.com/ocefpaf/8b9a90bfa40d7dc27c63e3bf22ef335a)

### Expected behavior

Successful build :smile: 

### Environment

- Pyodide Version:
- Browser version:
- Any other relevant information:

### Additional context

A patch to rename `cython` to `Cython` in the cftime pyproject.toml fixed it but we should not be case sensitive with PyPI names.

xref.: https://github.com/pyodide/pyodide/pull/2504

hoodmane:
Thanks for the report!

I think this line needs to change:
https://github.com/pyodide/pyodide/blob/d74bc74c9d76ce1d1579fd0af0a77845428cae1b/pyodide-build/pyodide_build/pypabuild.py#L42
to 
```py
            if avoid_name in req.name.lower():
```

ocefpaf:
Should I send a PR or do you prefer to send it?

hoodmane:
I'm not completely sure if this is a correct fix, could you try it and see? I will also test it out. I'm actually a bit confused. It would be great if you would make the PR, but only if this is the right fix.

ocefpaf:
Working on it...

hoodmane:
I don't think this fixes it...

hoodmane:
So it does work to remove `cython` from `UNISOLATED_PACKAGES`. I can't remember why I needed to put it in there right now, so we could try removing it. I suspect the build will fail somewhere but maybe we can find a different fix for that. The comments on #2272 may give some clues.

hoodmane:
> Cython I unisolated specifically because numcodecs needs it but it isn't in the numcodecs build dependencies.

Okay so this is a hack to fix numcodecs. In that case we should really patch numcodes.

hoodmane:
I think the problem is that some packages don't declare that they need cython and their plan is not to recythonize if it isn't present. So we want to include some version of it always.

ocefpaf:
[...]

PR #2504, opened 2022-05-05
add cftime

ocefpaf:
### Description

I'm not sure if this is the right way to add packages to pyodide. Feel free to close this PR if this is not OK.

### Checklists

- [ ] Add / update tests

### DIFF
--- a/packages/cftime/meta.yaml
+++ b/packages/cftime/meta.yaml
@@ -1,0 +1,23 @@
+package:
+  name: cftime
+  version: 1.6.0
+
+source:
+  url: https://files.pythonhosted.org/packages/20/3f/607a69f1253e5c8835400641da91a6b54cc49e3e9f6ac5848a745638f9ea/cftime-1.6.0.tar.gz
+  sha256: 13103e6650beea6552316bd5825d6aa3b7e98f5b8115026df4826798dff9f741
+  patches:
+    - patches/upper_case_cython.patch
+
+requirements:
+  run:
+    - numpy
+
+test:
+  imports:
+  - cftime
+
+about:
+  home: https://github.com/Unidata/cftime/actions'
+  PyPI: https://pypi.org/project/cftime
+  summary: Time-handling functionality from netcdf4-python
+  license: 'License :: OSI Approved :: MIT License'
--- a/packages/cftime/meta.yaml
+++ b/packages/cftime/meta.yaml
@@ -15,7 +15,7 @@ test:
     - cftime
 
 about:
-  home: https://github.com/Unidata/cftime/actions'
+  home: https://github.com/Unidata/cftime'
   PyPI: https://pypi.org/project/cftime
   summary: Time-handling functionality from netcdf4-python
   license: "License :: OSI Approved :: MIT License"
--- a/packages/cftime/meta.yaml
+++ b/packages/cftime/meta.yaml
@@ -14,10 +14,10 @@ requirements:
 
 test:
   imports:
-  - cftime
+    - cftime
 
 about:
   home: https://github.com/Unidata/cftime/actions'
   PyPI: https://pypi.org/project/cftime
   summary: Time-handling functionality from netcdf4-python
-  license: 'License :: OSI Approved :: MIT License'
+  license: "License :: OSI Approved :: MIT License"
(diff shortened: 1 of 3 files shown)


packages/cftime/patches/upper_case_cython.patch
```diff
@@ -0,0 +1,8 @@
+diff --color -Naur cftime-1.6.0.orig/pyproject.toml cftime-1.6.0/pyproject.toml
+--- cftime-1.6.0.orig/pyproject.toml	2022-03-03 10:59:04.000000000 -0300
++++ cftime-1.6.0/pyproject.toml	2022-05-05 10:11:01.523853642 -0300
+@@ -1,3 +1,3 @@
+ [build-system]
+-requires = ["setuptools>=41.2", "cython", "wheel", "oldest-supported-numpy"]
++requires = ["setuptools>=41.2", "Cython", "wheel", "oldest-supported-numpy"]
```
henryiii, line 7: Hmm, package names should probably be normalized in Pyodide.

ocefpaf: Looks like @hoodmane already have a solution for it. See https://github.com/pyodide/pyodide/issues/2503

hoodmane: Well I didn't solve the capitalization problem. Some part of the pypabuild unisolation introduces case sensitivity I guess, but I don't understand how. The fix was not to unisolate cython which is an improvement because what we actually want is to always include cython. But we can always include cython by always including it and this is better than unisolating it.

rth:
Thanks @ocefpaf ! I guess the patch will no longer be necessary once https://github.com/pyodide/pyodide/pull/2507 is merged.

ocefpaf:
> Thanks @ocefpaf ! I guess the patch will no longer be necessary once #2507 is merged.

Waiting on https://github.com/pyodide/pyodide/pull/2507 to remove it here.

PS: is this the right way to add packages? I want to work on a few more. Hopefully even get to libnetcdf, but I'll probably need some expert help for that one. B/c you already have hdf5 it should not be too difficult for someone who knows that they are doing. (Which I'm not!)

hoodmane:
Yup it's exactly right. 

Only a few of us understand the Pyodide build system at all right now so getting more people using it is really important. Then we can fix common issues, work on making it behave more in line with other build systems people are familiar with, and improve the docs. Lots of work to be done here...

rth (approved):
Thanks !

dopplershift:
> PS: is this the right way to add packages? I want to work on a few more. Hopefully even get to libnetcdf, but I'll probably need some expert help for that one.

@ocefpaf Just wanted to let you know that I'm also starting to poke around in this space so we might want to sync up... 😉

ocefpaf:
> @ocefpaf Just wanted to let you know that I'm also starting to poke around in this space so we might want to sync up...

I'm pretty sure our goal is the same: libnetcdf &rarr; netCDF4&rarr; World Wide Web domination.

Let's try to discuss this in our next conda-forge meeting, for now I'll play with simpler recipes like some more python+cython stuff.

packages/cftime/meta.yaml
```diff
@@ -0,0 +1,21 @@
...
+requirements:
+  run:
+    - numpy
+
+test:
+  imports:
+    - cftime
+
+about:
+  home: https://github.com/Unidata/cftime/actions'
```
rth, line 18: Probably a typo,

```suggestion
  home: https://github.com/Unidata/cftime'
```

Otherwise LGTM, thanks!

dopplershift: Isn't the trailing `'` a typo?

packages/cftime/meta.yaml
```diff
@@ -0,0 +1,21 @@
...
+requirements:
+  run:
+    - numpy
+
+test:
+  imports:
+    - cftime
+
+about:
+  home: https://github.com/Unidata/cftime'
```
henryiii, line 18:
```suggestion
  home: https://github.com/Unidata/cftime
```

packages/cftime/meta.yaml
rth: Yes, there were two typos thanks!

PR #2507, opened 2022-05-05
Fix Cython handling in pypa/build

### LINKED ISSUE #2503: pyodide_build buildpkg does not install Cython as a build dependency when it is spell with a lower case c

hoodmane:
Before we were unisolating it in order to insure it is always present.
But to ensure it is always present we should just always install it.
This fixes #2503. I still don't understand exactly what causes #2503
but it clearly is problematic to unisolate cython.

### DIFF
--- a/pyodide-build/pyodide_build/pypabuild.py
+++ b/pyodide-build/pyodide_build/pypabuild.py
@@ -19,7 +19,7 @@ from packaging.requirements import Requirement
 
 from .common import get_hostsitepackages, get_pyversion
 
-UNISOLATED_PACKAGES = ["numpy", "scipy", "cffi", "pycparser", "pythran", "cython"]
+UNISOLATED_PACKAGES = ["numpy", "scipy", "cffi", "pycparser", "pythran"]
 
 
 def symlink_unisolated_packages(env: IsolatedEnv):
@@ -39,7 +39,7 @@ def remove_unisolated_requirements(requires: set[str]) -> set[str]:
     for reqstr in list(requires):
         req = Requirement(reqstr)
         for avoid_name in UNISOLATED_PACKAGES:
-            if avoid_name in req.name:
+            if avoid_name in req.name.lower():
                 requires.remove(reqstr)
     return requires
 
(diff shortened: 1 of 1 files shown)


henryiii:
https://github.com/pypa/packaging/blob/ba124324e866e518ebfe0378a25b6ba4816e5880/packaging/utils.py#L32 should always be used on names.

pyodide-build/pyodide_build/pypabuild.py
```diff
@@ -58,6 +58,11 @@ def replace_env(build_env: Mapping[str, str]):
 
 def install_reqs(env: IsolatedEnv, reqs: set[str]):
     env.install(remove_unisolated_requirements(reqs))
+    # Some packages (numcodecs) don't declare cython as a build dependency and
+    # only recythonize if it is present. We need them to always recythonize so
+    # we always install cython. If the reqs included some cython version already
+    # then this won't do anything.
+    env.install(["cython"])
```
rth, line 65: OK, but this is also a temporary solution. The right solution is to have build dependencies in meta.yaml which I need to finish one day. Installing cython for each package build "just in case" is not optimal in the long term.

Maybe it would be worth adding a TODO: note in the comments.

rth (approved):
Thanks ! LGTM.

@ocefpaf could you please confirm this fixes your original issue?

pyodide-build/pyodide_build/pypabuild.py
ocefpaf: Coming from conda-forge I am obviously biased here but I agree. If we could specify build/host dependencies it will be better.

rth: I was pretty close in https://github.com/pyodide/pyodide/pull/2132 to getting it to work (btw if you have any feedback it would be very welcome), just need to figure out the remaining CI issues. Will try to get back to it in the near future. Meanwhile, this sounds like a reasonable temporary solution.

hoodmane: This all makes sense. There's a lot of work to be done on our build system, but it's improving pretty rapidly. We would really appreciate your help though. I am personally not experienced with build systems so I have to make stuff up as I go a lot, though our other maintainers can tell me when I go too far astray =)

rth: Absolutely would be happy to chat. Will send you an email later.

Issue #2505, opened 2022-05-05
Online console output is wrong when using `input()`


dragoncoder047:
## 🐛 Bug with online console

Stdout when using `input()` in online console is wrong.

### To Reproduce

Go to https://pyodide.org/en/stable/console.html and enter `input('foo bar: ')`. After entering `barbaz` in the prompt box:

```pycon
>>> input('foo bar: ')
foo bar: 'barbaz'
>>>
```

### Expected behavior

No prompt box, and:

```pycon
>>> input('foo bar: ')
foo bar: barbaz
'barbaz'
>>>
```

### Environment

- Pyodide Version: `Python 3.10.2 (main, Apr  9 2022 20:52:01)`. Latest release was `0.20.1a1`.
- Browser version: Chrome 96.0.4664.111 64 bit (on chromebook) *I don't think this is relevant, though.*

This issue may be related to #1898 at least with the prompt box. However, the stdin callback (in `loadPyodide`) could be fixed to echo the input back to the terminal.

hoodmane:
For now, this being broken is kind of the expected behavior. Getting `input()` to work requires nontrivial setup, for an example repl with working input see here:
https://github.com/hoodmane/worker-pyodide-console

Perhaps we should add a section of the docs giving limitations of the repl. Ideally someone downstream would make a really good repl... 
@rth what do you think about this?

dragoncoder047:
I'll leave the fixing the prompt box to you guys since I know nothing about Emscripten and webassembly. But the solution I was thinking of was something like this: (of the Javascript on the console page)

```diff
 async function main() {
-  globalThis.pyodide = await loadPyodide();
+  globalThis.pyodide = await loadPyodide({
+    stdin: x => {
+      term.echo(x, {newline: false});
+      return prompt(x);
+    }
+  });
   let namespace = pyodide.globals.get("dict")();
```

This will fix *half* of it. As for the other half, I'll leave it to you guys...

hoodmane:
You want to open a PR with that @dragoncoder047?

ryanking13:
I'll close this for now, as I think there are not many things we can do unless we solve the blocking issue in the main thread.
Thanks, @dragoncoder047 for the improvement in the REPL.

PR #2509, opened 2022-05-05
fix for half of pyodide/pyodide#2505

dragoncoder047:
applied patch here: https://github.com/pyodide/pyodide/issues/2505#issuecomment-1118591356

### DIFF
--- a/docs/project/changelog.md
+++ b/docs/project/changelog.md
@@ -14,7 +14,7 @@ substitutions:
 
 ## Unreleased
 
-- {{ Fix }} Fix output bug with using `input()` 
+- {{ Fix }} Fix output bug with using `input()` on online console
   {pr}`2509`
 
 - {{ Enhancement }} Update sqlite version to latest stable release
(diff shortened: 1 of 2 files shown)


rth:
Thanks @dragoncoder047 ! Could you please post a screenshot or copy-paste of the REPL result when using input with this change?

rth:
You can copy `src/templates/console.html` in a different folder. Update the Podide download link to `https://cdn.jsdelivr.net/pyodide/dev/full/pyodide.js`, then start a local HTTP server with `python -m http.server` and open the console.html in your browser.

ryanking13:
Here is the REPL result.

https://user-images.githubusercontent.com/24893111/167232548-1d05d5d6-7ba5-4888-b6de-041d45fa634c.mp4

I'm not sure if this is what we want. WDYT @rth?

hoodmane:
I think it's better than the current behavior, I'm +1 to merge it.

hoodmane:
Thanks for testing it @ryanking13!

rth (approved):
> I think it's better than the current behavior, I'm +1 to merge it.

Agreed. It is indeed closer to the standard REPL behavior. Thanks @dragoncoder047 and @ryanking13 !

PR #2510, opened 2022-05-05
Pytest rewrites for run_in_pyodide

hoodmane:
This updates `run_in_pyodide` to do pytest rewrites. We use private pytest rewrite apis to parse the ast of the test modules and store the appropriate transformed ast. Then we pickle the ast, base 64 encode it, and send it to the browser where we base 64 decode, unpickle, compile, and exec. Then we grab the function we defined and call it.

In the future, we should add support for pytest decorators, most importantly xfail and parametrize.

### Checklists

- [x] Add / update tests


hoodmane:
@henryiii I'm having trouble with the mypy pre-commit.ci job processing `conftest.py` even though it is supposed to be ignored:
```yaml
exclude: (setup.py|^src/tests|conftest.py)
```

henryiii:
mypy has to be double excluded. You need to make sure the file is not passed directly to mypy (via .pre-commit-config.yaml), and you have to make sure it is not transitivity picked up via another file that is included (via pyproject.toml `tool.mypy.exclude`).

henryiii:
Why is it listed directly in `files:` still, though?

henryiii:
Ahh, there are two mypy jobs, and if you don't want this checked, you'll need to remove it from the one where it's manually included.

hoodmane:
> you'll need to remove it from the one where it's manually included.

But we want it to be processed by `mypy-test`, that one thinks it typechecks fine. The problem is that `mypy` is also processing it and finds a bunch of errors. Fixing the errors that `mypy` generates causes errors in `mypy-test`.

ryanking13:
Thanks for working on this @hoodmane. I haven't read the code thoroughly yet, but it looks like a good improvement in our testing system.

I think #2418, which re-organizes the testing codes, will have some code conflicts with this, so would you mind reviewing that PR first? Of course it is also possible that I pause #2418 until you finish this one.

rth (commented):
Very nice! I think conceptually pickling the test ast, sending it to the runner, and pickling back the error makes very much sense.  A few comments otherwise looks good to me.

I imagine this idea could also be extended one day to package test suites where one could collect tests on the runner, send the list back to host, then generate individual pytests for those and finally run them one by one on the runner. Though it's more complex that this use case. I was considering some of it back in https://github.com/pytest-dev/pytest-xdist/issues/336

pyodide-build/pyodide_build/testing.py
```diff
@@ -1,29 +1,79 @@
...
-from pprint import pformat
+def _encode_ast(module_ast, funcname):
+
+    nodes: list[Any] = []
+    for node in module_ast.body:
+        # We need to include the magic imports that pytest inserts
+        if (
+            isinstance(node, ast.Import)
+            and node.names[0].asname
+            and node.names[0].asname.startswith("@")
```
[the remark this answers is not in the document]
hoodmane: Actual imports starting with an @. The @ is included to ensure that they aren't valid identifiers so that they are difficult to use without an ast transform. This reduces the chance they will clash with the identifiers used in the module.

They are injected into the ast here:
https://github.com/pytest-dev/pytest/blob/5f9d68c8d93b792116fc474cb4ec3072b7baa6f1/src/_pytest/assertion/rewrite.py#L716

hoodmane:
> this idea could also be extended one day to package test suites where one could collect tests on the runner, send the list back to host, then generate individual pytests for those and finally run them one by one on the runner. 

Yes this could be done though I still think it would be better to just have the test suite on the host machine and do the test collection in the host. This will significantly simplify the implementation and also I think it will be easier to use.

hoodmane:
#2418 fixed the precommit problems on this PR. Thanks @ryanking13!

pyodide-test-runner/pyodide_test_runner/decorator.py
```diff
@@ -16,112 +91,83 @@ def run_in_pyodide(
...
+        selenium driver timeout (in seconds). If missing, use the default
+        timeout.
 
+    pytest_assert_rewrites : bool, default = True
+        If True, use pytest assertion rewrites. This gives better error messages
+        when an assertion fails, but requires us to load pytest.
+    """
     xfail_browsers_local = xfail_browsers or {}
 
+    from conftest import ORIGINAL_MODULE_ASTS, REWRITTEN_MODULE_ASTS
```
ryanking13, line 125: Ideally in the future, we will have to remove circular imports between `conftest.py` and `pyodide-test-runner`. (no action required for this PR)

ryanking13 (approved):
Thanks, @hoodmane! Generally LGTM. Happy to see that test codes are getting better.

pyodide-test-runner/pyodide_test_runner/decorator.py
hoodmane: Probably the thing would be to define the pytest hooks in `pyodide_test_runner/pytest_hooks.py` and then import them from there into `conftest.py`.

hoodmane:

PR #2513, opened 2022-05-05
ENH Verifying checksum when loading packages in browser

amitsaha:
### Description

This PR does the following to complete https://github.com/pyodide/pyodide/issues/2431:

- Updates the change made in https://github.com/pyodide/pyodide/pull/2455 to generate Subresource integrity hash as explained [here](https://developer.mozilla.org/en-US/docs/Web/Security/Subresource_Integrity). As the document explains, it is a base64 encoded string of the "binary" hash, not the data returned by `hexdigest()` method. Hence, this change is made as well so that `package.json` contains the subresource integrity hash directly
- Implement the verification of the checksum when loading packages

This ignores the verification for NodeJS environment as the `node-fetch` module doesn't support the `integrity` option. Node 17.5.0 has added the fetch API as experimental, it would be prudent to come back and fix this when we are ready to use that version.

### Checklists

### DIFF
--- a/pyodide-build/pyodide_build/buildall.py
+++ b/pyodide-build/pyodide_build/buildall.py
@@ -5,7 +5,6 @@ Build all of the packages in a given directory.
 """
 
 import argparse
-import base64
 import hashlib
 import json
 import os
@@ -432,12 +431,12 @@ def build_from_graph(pkg_map: dict[str, BasePackage], outputdir: Path, args) -> 
     )
 
 
-def _generate_package_sub_resource_hash(full_path: Path) -> str:
+def _generate_package_hash(full_path: Path) -> str:
     sha256_hash = hashlib.sha256()
     with open(full_path, "rb") as f:
         while chunk := f.read(4096):
             sha256_hash.update(chunk)
-    return "sha256-" + base64.b64encode(sha256_hash.digest()).decode()
+    return sha256_hash.hexdigest()
 
 
 def generate_packages_json(output_dir: Path, pkg_map: dict[str, BasePackage]) -> dict:
@@ -458,9 +457,7 @@ def generate_packages_json(output_dir: Path, pkg_map: dict[str, BasePackage]) ->
             "version": pkg.version,
             "file_name": pkg.file_name,
             "install_dir": pkg.install_dir,
-            "sub_resource_hash": _generate_package_sub_resource_hash(
-                Path(output_dir, pkg.file_name)
-            ),
+            "sha_256": _generate_package_hash(Path(output_dir, pkg.file_name)),
         }
         if pkg.shared_library:
             pkg_entry["shared_library"] = True
--- a/pyodide-build/pyodide_build/buildall.py
+++ b/pyodide-build/pyodide_build/buildall.py
@@ -457,7 +457,7 @@ def generate_packages_json(output_dir: Path, pkg_map: dict[str, BasePackage]) ->
             "version": pkg.version,
             "file_name": pkg.file_name,
             "install_dir": pkg.install_dir,
-            "sha_256": _generate_package_hash(Path(output_dir, pkg.file_name)),
+            "sha256": _generate_package_hash(Path(output_dir, pkg.file_name)),
         }
         if pkg.shared_library:
             pkg_entry["shared_library"] = True
--- a/pyodide-build/pyodide_build/buildall.py
+++ b/pyodide-build/pyodide_build/buildall.py
@@ -5,6 +5,7 @@ Build all of the packages in a given directory.
 """
 
 import argparse
+import base64
 import hashlib
 import json
 import os
@@ -431,12 +432,12 @@ def build_from_graph(pkg_map: dict[str, BasePackage], outputdir: Path, args) -> 
     )
 
 
-def _generate_package_hash(full_path: Path) -> str:
+def _generate_package_sub_resource_hash(full_path: Path) -> str:
     sha256_hash = hashlib.sha256()
     with open(full_path, "rb") as f:
         while chunk := f.read(4096):
             sha256_hash.update(chunk)
-    return sha256_hash.hexdigest()
+    return "sha256-" + base64.b64encode(sha256_hash.digest()).decode()
 
 
 def generate_packages_json(output_dir: Path, pkg_map: dict[str, BasePackage]) -> dict:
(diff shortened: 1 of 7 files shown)


rth:
Also, don't hesitate to comment once it's ready for a review. So far the code looks good to me.

amitsaha:
@rth To test this manually, i created this HTML page.

```
<head>
    <meta charset="utf-8" />
  </head>
  <body>
    <script
      type="text/javascript"
      src="http://127.0.0.1:9000/pyodide.js"
    ></script>
      async function main() {
        import micropip
        await micropip.install('sympy')
	`);
	
        from sympy import Symbol, factor, expand
        x = Symbol('x')
        y = Symbol('y')
        expr = x**2 - y**2
        f = factor(expr)
        print(f)
        print(expand(f))
      `);
      }
      main();
    </script>
  </body>
</html>```

If I tamper with the hash in the `packages.json` file, an error like follows will be shown to the user:

```
Failed to find a valid digest in the 'integrity' attribute for resource 'http://127.0.0.1:9000/micropip-0.1-py3-none-any.whl' with computed SHA-256 integrity 'yT3ZgjhIbyKpFinBJq/YgkQvKZ3gyzCqVAWYRQ5vBOI='. The resource has been blocked.
```

**Comment by amitsaha** [comment]
> Also, don't hesitate to comment once it's ready for a review. So far the code looks good to me.

Thanks, ready for review now!

**Comment by rth** [comment]
Thanks @amitsaha ! So I think it might be better to still keep the hexdigest in the packages.json  because then one can use `sha256sum` to validate it locally. And in the JS package use Python functions to convert that to something that would work with fetch (see e.g. https://stackoverflow.com/q/7856196/1791279) at runtime.
I think it should be possible to use something,
```
let base64 = pyodide.pyimport('base64');
let binascii = pyodide.pyimport('binascii');
let hash_out = base64.b64encode(binascii.unhexlify(hash_in)).decode();
```
except using `API.importlib.import_module` instead of `pyodide.pyimport` (which is what it calls internally) inside the `src/js/load-package.ts` functions.

But maybe let's wait for a second opinion before making any changes.

**Comment by amitsaha** [comment]
@rth sounds like a fair call. 

The other alternative i am inclined to propose is have both the sha256 and the SRI hash, so i will add a new entry in `package.json` and use that directly and then avoid the calculation at runtime - that is consistent with typical hash verification where the correct hash is generated at build time and verified as it is instead of calculating the it at runtime.

**Comment by hoodmane** [comment]
> I think it should be possible to use something, similar to what you did at build time, ... 

I think instead of using `pyodide.pyimport` you should define an extra function `get_hash` or something in `src/py/pyodide/_package_loader.py` and then call it like `API.package_loader.get_hash`.

**Comment by rth** [comment]
I would say that let's still go with Hood's suggestion above, define a function to do this change and call it from JS, if that works for you.

On arguments against including both, is that I think it's better to have a single source of truth and not wonder if those two keys actually contain the same information (also avoiding extra stuff to download even if it's a little). Also, we are somewhat trying to be compatible with [conda's repodata.json](https://docs.conda.io/projects/conda-build/en/latest/concepts/generating-index.html#repodata-json) which has the sha256 key only. We are not 100% compatible but in this case, it doesn't cost much to use the same key.

**Comment by amitsaha** [comment]
@hoodmane Just a clarification on your comment:
> I think instead of using pyodide.pyimport you should define an extra function get_hash or something in src/py/pyodide/_package_loader.py and then call it like API.package_loader.get_hash.

Would this `get_hash()` function return the exact resource identifier i.e use the sha256 sum in the `package.json` and then derive the sub resource integrity hash and return it?

**Comment by rth** [comment]
Yes, I think so. Essentially we can take code from https://github.com/pyodide/pyodide/pull/2513#issuecomment-1120125016 but put it into a function.

**Comment by amitsaha** [comment]
@rth this PR is now ready for review. I think the test failures that happened are not related, but would be great to have your input when you get a chance.

File: src/js/load-package.ts [hunk 1]
```diff
@@ -162,17 +162,21 @@ async function downloadPackage(
...
 ): Promise<Uint8Array> {
-  let file_name;
+  let file_name, file_sub_resource_hash;
   if (channel === DEFAULT_CHANNEL) {
     if (!(name in API.packages)) {
       throw new Error(`Internal error: no entry for package named ${name}`);
     }
     file_name = API.packages[name].file_name;
+    file_sub_resource_hash = API.package_loader.sub_resource_hash(
+      API.packages[name].sha_256
```
rth, line 172: Unrelated to this PR, but it should be `sha256` without the `_` in the JSON to be consistent with https://docs.conda.io/projects/conda-build/en/latest/concepts/generating-index.html#repodata-json Sorry I missed it in your first PR.

src/py/pyodide/_package_loader.py
```diff
@@ -179,3 +181,19 @@ def get_dynlibs(archive: IO[bytes], target_dir: Path) -> list[str]:
...
+
+    Parameters
+    ----------
+    sha_256
+        A hexdigest of the SHA-256 sum.
+
+    Returns
+    -------
+        The sub resource integrity hash corresponding to the sum.
+    """
```
rth, line 197: Could you please add a doctest to illustrate how to use this (and it would serve a test)

rth (commented):
Thanks @amitsaha ! A few comments otherwise LGTM.

Unfortunately, the CI failure is related to this PR, it means that we are not deallocating a PyProxy object somewhere, and leaking memory (and the tests are catching that). I had a look, I can't see where it could happen: the python function take and returns a str which should be auto-converted between languages (and so it's not a PyProxy we need to destroy). Maybe @hoodmane  would have an idea.

amitsaha:
> Thanks @amitsaha ! A few comments otherwise LGTM.
> 
> Unfortunately, the CI failure is related to this PR, it means that we are not deallocating a PyProxy object somewhere, and leaking memory (and the tests are catching that). I had a look, I can't see where it could happen: the python function take and returns a str which should be auto-converted between languages (and so it's not a PyProxy we need to destroy). Maybe @hoodmane would have an idea.

Ah, all good, I will dig into it. Thanks.

amitsaha:
@rth I have a current test failure due to the `regex-tests` package not having a sha256 generated for it:

```
 "regex-tests": {
            "name": "regex-tests",
            "version": "2022.3.15",
            "depends": [
                "regex"
            ],
            "imports": [],
            "file_name": "regex-tests.tar",
            "install_dir": "site"
        },
```

Not sure what the right fix here is? Thanks.

pyodide-build/pyodide_build/buildall.py
```diff
@@ -479,6 +479,9 @@ def generate_packages_json(output_dir: Path, pkg_map: dict[str, BasePackage]) ->
                 "imports": [],
                 "file_name": pkg.unvendored_tests.name,
                 "install_dir": pkg.install_dir,
+                "sha256": _generate_package_hash(
```
amitsaha, line 482: Explained in https://github.com/pyodide/pyodide/pull/2513#issuecomment-1126651792

amitsaha:
@rth gentle reminder that this PR is read for final review.

hoodmane (approved):
It looks good to me. Thanks @amitsaha!

rth:
Thanks a lot @amitsaha !

PR #2515, opened 2022-05-06
Add js_id attribute to JsProxy for use as a dictionary key

hoodmane:
### Description

It's useful to be able to hash on object identity. For instance in Python if we want to store objects in a dictionary using object identity to store/access `obj` we use `id(obj)` as the index. This adds `js_id` as an attribute to `JsProxy` so that `jsobj.js_id` can be used as a dictionary key to key on identity of the JavaScript object.

### Checklists

- [x] Add / update tests

### DIFF
--- a/src/core/hiwire.c
+++ b/src/core/hiwire.c
@@ -68,7 +68,7 @@ EM_JS_NUM(int, hiwire_init, (), {
 
   Hiwire.new_value = function(jsval)
   {
-    let idval = _hiwire.obj_to_key.has(jsval);
+    let idval = _hiwire.obj_to_key.get(jsval);
     if(idval !== undefined){
       _hiwire.objects.get(idval)[1]++;
       return idval;
--- a/src/core/hiwire.c
+++ b/src/core/hiwire.c
@@ -16,7 +16,7 @@ const JsRef Js_false = ((JsRef)(6));
 const JsRef Js_null = ((JsRef)(8));
 
 // For when the return value would be Option<JsRef>
-const JsRef Js_novalue = ((JsRef)(1000));
+const JsRef Js_novalue = ((JsRef)(10));
 
 JsRef
 hiwire_from_bool(bool boolean)
@@ -59,7 +59,7 @@ EM_JS_NUM(int, hiwire_init, (), {
   HIWIRE_INIT_CONST(_Js_null, JSNULL, null);
   HIWIRE_INIT_CONST(_Js_true, TRUE, true);
   HIWIRE_INIT_CONST(_Js_false, FALSE, false);
-  let hiwire_next_permanent = Hiwire.FALSE + 2;
+  let hiwire_next_permanent = HEAPU8[_Js_novalue] + 2;
 
 #ifdef DEBUG_F
   Hiwire._hiwire = _hiwire;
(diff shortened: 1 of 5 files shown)


src/tests/test_jsproxy.py
```diff
@@ -1155,3 +1155,13 @@ def test_memory_leaks(selenium):
...
         """
     )
+
+
+@run_in_pyodide
+def test_js_id():
+    from js import eval as run_js
+
+    [x, y, z] = run_js("let a = {}; let b = {}; [a, a, b]")
+    assert x.js_id == y.js_id
```
rth, line 1165: BTW, we are currently not implementing `__eq__` for JsProxy, right?
Currently `x ==  y` evaluates to False as far as I can tell.   But that's a separate issue.

src/core/hiwire.c
```diff
@@ -67,17 +68,21 @@ EM_JS_NUM(int, hiwire_init, (), {
 
   Hiwire.new_value = function(jsval)
   {
-    // Should we guard against duplicating standard values?
-    // Probably not worth it for performance: it's harmless to occasionally
-    // duplicate. Maybe in test builds we could raise if jsval is a standard
-    // value?
```
rth, line 73: Could you add an inline comment explaining how it's handled now?

src/core/jsproxy.c
```diff
@@ -136,6 +136,20 @@ JsProxy_typeof(PyObject* self, void* _unused)
   return result;
 }
 
+static PyObject*
+JsProxy_js_id(PyObject* self, void* _unused)
+{
+  PyObject* result = NULL;
+
+  JsRef idval = JsProxy_REF(self);
+  int x[2] = { (int)Py_TYPE(self), (int)idval };
```
rth, line 145: So we add the type and hash because python hash is an identity for integers otherwise, as you said offline. But why it's a problem actually? Is there a situation where an object has the same idval but different types? 

In any case it would be good to add an inline comment explaining why this is done.

rth (commented):
Thanks ! A few comments otherwise, LGTM. Could you include it in the API docs and explain how this value is calculated in the docstring there?

src/tests/test_jsproxy.py
hoodmane: Right, we should probably add equality checks in a separate PR.

src/core/jsproxy.c
hoodmane: Well imagine you have the following code:

```py
def store_js_object(x):
	if instanceof(x, JsProxy):
		my_set.add(x.js_id)
	else:
		my_set.add(x)
```
so some types get converted others get proxied. This is perfectly reasonable code. However, it wouldn't work very well without this becuase you would have `JsProxies` with `js_id` things like 5, 7, 11, 17 but you might also have these numbers appearing. If we ensure that `js_id` is a uniformly distributed 32 bit integer, it's much less likely for there to be a collision.

hoodmane: Though it would be convenient for debugging internals to add a separate `_js_id` function to `_pyodide_core` that takes a `JsProxy` as argument and returns the hiwire key.

src/core/hiwire.c
hoodmane: Added a couple of comments explaining how we handle it now.

src/core/jsproxy.c
hoodmane: Added `hiwire_id` as a private function to `_pyodide_core` to return the hiwire id of the `JsProxy`.

Issue #2449, opened 2022-05-02
Race condition in build creating _sysconfigdata__emscripten_wasm32-emscripten


mdboom:
## 🐛 Bug

When building pyodide, it doesn't seem to create `_sysconfigdata__emscripten_wasm32-emscripten` in all cases.

### To Reproduce

Run `make` on Debian bullseye.

Running the console example has the following traceback in the browser console:

```
  File "/lib/python3.10/sysconfig.py", line 218, in _expand_vars
    _extend_dict(vars, get_config_vars())
  File "/lib/python3.10/sysconfig.py", line 615, in get_config_vars
    _init_posix(_CONFIG_VARS)
  File "/lib/python3.10/sysconfig.py", line 477, in _init_posix
    _temp = __import__(name, globals(), locals(), ['build_time_vars'], 0)
ModuleNotFoundError: No module named '_sysconfigdata__emscripten_wasm32-emscripten'
```

### Expected behavior

console example inits successfully.

### Environment

Debian bullseye on WSL, with all packages installed to match pyodide's `Dockerfile`.

### Additional context

Doing 

```
make
touch cpython/remove_modules.txt
make
```

forces a rerun of this [make instruction](https://github.com/pyodide/pyodide/blob/main/cpython/Makefile#L35).  This is probably a clue to the solution, but obviously not "the" solution.

aaaandrzej:
Trying to run `make` locally on MacOS (with M1, Python 3.10.2) fails at the same step. I think that `_PYTHON_SYSCONFIGDATA_NAME=$(SYSCONFIG_NAME) _PYTHON_PROJECT_BASE=$(BUILD) $(HOSTPYTHON) -m sysconfig --generate-posix-vars` in `cpython/Makefile` is failing to generate sysconfigdata there

rkaminsk:
The whole Makefile is just not written correctly. It completely fails if the `-j` option is passed to build with multiple processes.

hypervanse:
- __phello__
- _collections_abc
- _compat_pickle
- _compression
- _markupbase
- _py_abc
- _pydatetime
- _pyio
- _pylong
- _pyodide
- _sitebuiltins
- _sqlite3
- _strptime
- _sysconfigdata__emscripten_wasm32-emscripten
- _threading_local
- _weakrefset
- abc
- aifc
- antigravity
- argparse
- ast
- asyncio
- base64
- bdb
- bisect
- bz2
- cProfile
- calendar
- cgi
- cgitb
[...]

PR #2516, opened 2022-05-07
MAINT Use target sysconfigdata rather than host sysconfigdata when cross compiling

hoodmane:
### Description

`sysconfig.py` uses the environment variable `_PYTHON_SYSCONFIGDATA_NAME` to decide where to look for the sysconfig data file with relevant info about the compile target. We also need to separately insure that our sysconfig data file is on the path. We don't want the rest of our target stdlib on the path, so I made an extra sysconfigdata folder, copied the sysconfig data into it, and put it on the path.

I think this will improve things. Currently we have some hacky work-arounds for sysconfigdata but I can't remember where they are.

### Checklists


hoodmane:
We see another "it skips the compiler wrappers" problem here: the Emscripten sysconfigdata has the absolute paths for `emcc`, `em++`, etc. On the other hand, the information is *much* more accurate than in the host sysconfigdata so it is possible that if we appropriately shape the sysconfigdata we could simplify pywasmcross a lot. Ideally we could figure out how to fill the sysconfigdata in with paths to our compiler wrappers. Using `emconfigure` makes it hard to avoid the absolute paths to the compiler. 

I suppose good one option would be to write a python script to manually update the sysconfigdata at the end of the CPython Makefile rather than having to mess around with `emconfigure`. Idealy we could eventually get things like pkgconfig to work so that packages could correctly determine the paths for libraries to link rather than us needing to fix them manually. Probably there will always be some need for rewriting compiler args here.

ryanking13:
> Idealy we could eventually get things like pkgconfig to work so that packages could correctly determine the paths for libraries to link rather than us needing to fix them manually.

I recently started to work on installing libraries in a single path (#2516). It would be great if pkg-config or other similar things work so that we don't have to patch paths.

> On the other hand, the information is _much_ more accurate than in the host sysconfigdata so it is possible that if we appropriately shape the sysconfigdata we could simplify pywasmcross a lot

That sounds really great. pywasmcross is our bottleneck of package build, so it would be great if we can simplify (or in the future remove) it.

ryanking13:
Thanks, @hoodmane!

> We also need to separately insure that our sysconfig data file is on the path. We don't want the rest of our target stdlib on the path

Could you explain a little more about why it has to be placed in a separate path and how it is related to #2449?

hoodmane:
I think #2449 is unrelated: this probably won't make it any better or any worse.

We need a separate path for the `sysconfigdata` because we don't want to put an entire second copy of the Python stdlib on `sys.path`.  I think this could cause trouble. We are just trying to use a different `sysconfig`. The sysconfig data needs to be on the path, so this requires copying the sysconfig data to its own separate folder.

In any case I think this PR is only a very minor improvement. #2550 is significantly more important.

rth:
Thanks! The approach sounds good. I wonder if we could re-use part of https://github.com/benfogle/crossenv for cross-compilation (which also fixes the [target sysconfig](https://github.com/benfogle/crossenv/blob/master/docs/intro.rst#how-it-works)) that is in particular used by conda-forge. Otherwise, it feels like we are re-inventing some of these things though maybe our use case is different enough.

rth (approved):
Thanks, LGTM.

PR #2517, opened 2022-05-07
Escape double bracket in REPL

### LINKED ISSUE #2399: REPL give wrong output when it contains double bracket `[[ ]]`
## 🐛 Bug

I think because of this [feature](https://github.com/jcubic/jquery.terminal/issues/508#issue-463182053). The Pyodide [REPL](https://pyodide.org/en/stable/console.html) can not work correctly with input and output contains raw bracket string `[[ ]]`.

I solved the output part by replacing `]]` with `]&#93;` in the output of [repr_shorten](https://github.com/pyodide/pyodide/blob/93dc922f46e38f98bd30c3850c7bd094b1933163/src/py/pyodide/console.py#L477) but not sure how to handle the input case. Maybe we need upstream `jquery.terminal` to provide an option to turn off the feature.

### To Reproduce

Try input command `print('[[x]]')` or `print('[[1]]')`.

### Expected behavior

The output should be `[[x]]` but an `NameError: name 'x' is not defined` triggered.

### Environment

- Pyodide Version: Any
- Browser version: Any

### LINKED ISSUE #2485: SyntaxError: unmatched ']' in numpy.random.rand
### To Reproduce
Try to use the function in the REPL [console](https://pyodide.org/en/stable/console.html).
```
import numpy as np
np.random.rand(3, 2)
```

### Expected behavior
Return a numpy array to the console.

### Environment

- Pyodide Version: stable
- Browser version: Chromium 99.0.4844.84 

ryanking13:
### Description

Fix #2399
Fix #2485

We don't want to invoke text formatting when using JQuery Terminal. So escaping the double bracket.

I'm not sure this handles all edge cases, but it works for the most common case: printing a 2D array.

### Checklists

### DIFF
--- a/src/templates/console.html
+++ b/src/templates/console.html
@@ -51,6 +51,7 @@
         let await_fut = namespace.get("await_fut");
         let pyconsole = namespace.get("pyconsole");
         let clear_console = namespace.get("clear_console");
+        const echo = (msg) => term.echo(msg.replace("]]", "&rsqb;&rsqb;"));
         namespace.destroy();
 
         let ps1 = ">>> ",
@@ -91,7 +92,7 @@
             try {
               let [value] = await wrapped;
               if (value !== undefined) {
-                term.echo(
+                echo(
                   repr_shorten.callKwargs(value, {
                     separator: "\n[[;orange;]<long output truncated>]\n",
                   })
@@ -128,7 +129,7 @@
             "CTRL+C": async function (event, original) {
               clear_console();
               term.enter();
-              term.echo("KeyboardInterrupt");
+              echo("KeyboardInterrupt");
               term.set_command("");
               term.set_prompt(ps1);
             },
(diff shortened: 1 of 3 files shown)


src/templates/console.html
```diff
@@ -58,6 +58,8 @@
         let await_fut = namespace.get("await_fut");
         let pyconsole = namespace.get("pyconsole");
         let clear_console = namespace.get("clear_console");
+        const echo = (msg, ...opts) =>
+          term.echo(msg.replace("]]", "&rsqb;&rsqb;"), ...opts);
```
rth, line 62: Thanks a lot! Yes, clearly something needed escaping but I wasn't sure what.

So the opening double bracket doesn't need escaping?


Thanks for fixing the vars -> let as well.

rth (approved):
Thanks! 

Could you add a changelog entry?

src/templates/console.html
ryanking13: > So the opening double bracket doesn't need escaping?

Yes, that would be safer. If we escape opening brackets too, we need to stop using templates in our console.html.

https://github.com/pyodide/pyodide/blob/379388c20c8cb43e6d0a9c796f7247c4879dce5b/src/templates/console.html#L102-L104

We are using it to print a colored truncation message.

Issue #2519, opened 2022-05-07
Pyodide cannot use `CFUNCTYPE`


donno2048:
## 🐛 Bug

Pyodide cannot use `CFUNCTYPE`

### To Reproduce

This py-script example

```html
<!DOCTYPE html>
<html lang="en">
    <head>
        <meta charset="utf-8">
        <meta name="viewport" content="width=device-width, initial-scale=1.0">
        <link rel="stylesheet" href="https://pyscript.net/alpha/pyscript.css" />
        <script defer src="https://pyscript.net/alpha/pyscript.js"></script>
        <py-script>
async def main():
    from micropip import install
    await install('pyas')
    from pyas import function
    add_one = function(
        '8b c7'  # mov eax, edi
        '83 c0 01'  # add eax, 1
        'c3'  # ret
    )
    print(add_one(10))
main()
        </py-script>
    </head>
</html>
```

The `function` function implementation (roughly) 

```py
from ctypes import CFUNCTYPE, addressof, c_void_p as void
from binascii import unhexlify
from mmap import mmap
buffer = []
def function(asm: str):
    global buffer
    asm_parsed = unhexlify(str().join(chunk for l in asm.splitlines() for chunk in l.split()))
    buffer += [mmap(-1, length=max(len(asm_parsed) // 2, 8), prot=7)]
    for byte in asm_parsed: buffer[-1].write_byte(byte)
    return CFUNCTYPE(void)(addressof(void.from_buffer(buffer[-1])))
```
### Expected behavior

To print `11`

hoodmane:
Thanks for the report! I'm investigating.

hoodmane:
![Screenshot from 2022-05-07 12-47-13](https://user-images.githubusercontent.com/8739626/167269738-0908ff10-92e3-4bff-8ffb-74640f71dfc2.png)

hoodmane:
You are writing x86 assembly into memory and then trying to call it as a function pointer. Wasm is not x86 and this just can't work. It would be entertaining to have a library that could execute custom wasm strings, this is possible. To execute x86 you would need an x86 emulator.

donno2048:
Wasm is not x86, got it, thanks. I so dumb some times...

hoodmane:
Here's a full example that hard codes a wasm binary that defines an `add_one` function, loads it, and calls it with ctypes.

```py
[...]

PR #2520, opened 2022-05-07
FIX Fix feature detection when `obj.constructor` is undefined

hoodmane:
Currently the following code fails:
```py
from js import eval
eval("Object.create(null)")
```
with:
```py
JsException: TypeError: Cannot read properties of undefined (reading 'name')
```
This fixes it.

- [x] Update tests
- [x] Update changelog


ryanking13 (commented):
Wow thanks, @hoodmane. Nice catch.

```python
>>> from js import eval
>>> nullobj = eval("Object.create(null)")
>>> nullobj
  File "/lib/python3.10/site-packages/pyodide/console.py", line 489, in repr_shorten
    text = repr(value)
  File "<omitted>", line 102, in S.fn.init.interpreter
JsException: TypeError: Hiwire.get_value(...).toString is not a function
>>> repr(nullobj)
TypeError: Cannot read properties of undefined (reading 'fileName')
```

I'm getting this error when I try to print null object.

hoodmane:
null objects are problematic in a lot of ways... I ran into it in https://github.com/pyodide/pyodide/issues/2519#issuecomment-1120285958

src/core/jsproxy.c
```diff
@@ -119,6 +119,12 @@ static PyObject*
 JsProxy_Repr(PyObject* self)
 {
   JsRef idrepr = hiwire_to_string(JsProxy_REF(self));
+  if (idrepr == NULL) {
+    PyErr_Format(PyExc_TypeError,
+                 "Pyodide cannot generate a repr for this Javascript object "
+                 "because it has no 'toString' method");
+    return NULL;
```
ryanking13, line 126: Yes, I think this is reasonable as `alert(Object.create(null))` also raises TypeError in the browser.

ryanking13 (approved):
Thanks for work @hoodmane!

hoodmane:

PR #2521, opened 2022-05-08
Load the standard library from a tar file instead of emscripten file packager

hoodmane:
Turns out unpacking the tar requires a pretty small amount of code.

### Checklists


src/core/jsmemops.h
```diff
@@ -13,6 +13,9 @@
 #define DEREF_F64(addr, offset) HEAPF64[(addr >> 3) + offset]
 
 #define ASSIGN_U32(addr, offset, value) DEREF_U32(addr, offset) = value
+#define ALIGN_ADDRESS(addr, align) (addr &= (~((align)-1)))
+#define ALIGN_ADDRESS_UP(addr, align) (addr = ((addr + (align - 1)) & (-align)))
```
rth, line 17: Can you add a comment about these?

src/core/tar.js
```diff
@@ -0,0 +1,60 @@
...
+      SKIP_OCTAL(lastModified, 12);
+      SKIP_OCTAL(checksum, 8);
+      SET_STRING(type, 1);
+      SKIP_STRING(linkName, 100);
+      ALIGN_ADDRESS_UP(offset, 0x200);
+
+      if (type !== "" && type !== "0") {
+        throw new Error("Link support not implemented");
+      }
+      // GNU TAR ensures that the file name of a directory ends with a /.
```
rth, line 46:
```suggestion
      // GNU TAR ensures that the file name of a directory ends with a /
```
to avoid confusion

rth (commented):
Thanks for doing this! The implementation is indeed pretty short, did you loosely adapt an existing implementation, or wrote one from scratch based on the spec?

hoodmane:
> did you loosely adapt an existing implementation, or wrote one from scratch based on the spec?

I started based on the spec but when I got stuck I looked at an actual tar file in a hex editor and a few different existing implementations.

hoodmane:
I wonder why chrome-webworker fails...

rth:
OK, and we cannot use one of the existing C implementations due to code size? It would be better if we didn't have to worry about there being bugs that can lead to page crash in the tar implementation, particularly once people start loading their custom stdlib, which is different from what we test this code on in CI.

hoodmane:
I just had a harder time getting the existing implementations to work: they tend to have flexible APIs for a range of different use cases and it takes time to understand how to use the API correctly. It takes several calls to consume the C APIs and these are also quite easy to get wrong. I think this is the easiest approach, compared to figuring out how to pull in and build a C library and then figuring out how to use it.

It doesn't matter much, but this also is the fastest possible implementation. Working in JavaScript completely avoids memory copies: my code mounts the files directly into the file system as views of the tar buffer. To use C, the entire tar file would have to first be copied into wasm memory and then each file would be copied again to mount it into the file system. We are only talking about copying 12mb though.

ryanking13 (commented):
> I wonder why chrome-webworker fails...

When I tested locally, it sometimes success and sometimes fail (timeout). It is interesting that only "classic" type chrome webworkers are failing, while "module" type workers success...

rth:
Thanks a lot for working on this!

Taking a step back, as much as I personally want this feature, and despite you doing a great job reverse-engineering the tar format, I don't think it would be reasonable for us to ship a re-implementation of tar. Too many people rely on Pyodide to work right now, and the risk of this new implementation not working as expected in some edge cases is too large IMO. Particularly, once people start modifying the files they include in the stdlib. And also we have already lots of things to maintain if for any reason you are not here and there is a bug in this implementation, I don't think we should spend time debugging the tar parser.

> When I tested locally, it sometimes success and sometimes fail (timeout). It is interesting that only "classic" type chrome webworkers are failing, while "module" type workers success...

Which goes kind of in the direction of my previous comment.

So personally I would rather continue looking for C or JS library we could include or vendor to make this possible, and that we wouldn't have to maintain.

hoodmane:
> I would rather continue looking for C or JS library we could include or vendor to make this possible, and that we wouldn't have to maintain.

Okay, sounds good to me.

PR #2533, opened 2022-05-09
Fix python2js cache mechanism when used with dict_converter

### LINKED ISSUE #2530: toJS wrong when a dict contains multiple same dict
### To Reproduce

```js
d = pyodide.runPython('d={0: {1: 2}}; d[1]=d[0]; d').toJs({dict_converter: Object.fromEntries}) 
```
### Expected behavior

`d[0] === d[1]`, and changing `d[0]` will also change `d[1]`.
### Actual behavior
`d[0]` is an Object but `d[1]` becomes an Array.
### Environment

- Pyodide Version:0.20.0
- Browser version:Firefox 100.0, Chromium 101.0.4951.54

### Additional context

It's not serious for my usecase. I can just write `d[1]=d[0].copy()`. It's just confusing to get an Array before I realize I should copy the dict.

hoodmane:
This fixes #2530. Because we postprocess dictionaries, we can't fill them in until we've computed the final value. So store `Js_novalue` into the cache to mark that we will come back for it, then when we've actually computed the final value put that in the cache. If the `Js_novalue` is encountered, place the current object and key into a list to come back to later, and at the final step we fill in these objects with the appropriate final values.

### Checklists

- [x] Add / update tests


hoodmane:
@rth @ryanking13 Could one of you review this?

src/core/python2js.c
```diff
@@ -34,12 +34,40 @@ typedef struct ConversionContext_s
...
+  const cache = Hiwire.get_value(idcache);
+  for (const[parent, key, value] of list) {
+    let out_value = Hiwire.get_value(cache.get(value));
+    // clang-format off
+    if(parent.constructor.name === "Map"){
+      parent.set(key, out_value)
+    } else {
+      // This is unfortunately a bit of a hack, if user does something weird
+      // enough in dict_converter then it won't work.
+      parent[key] = out_value;
```
ryanking13, line 65: It is very tricky to deal with various edge cases... but yeah we can handle them when it happens.

ryanking13 (approved):
Thanks for working on this @hoodmane. The idea of postprocessing things looks decent. Generally LGTM.

PR #2536, opened 2022-05-09
meta.yaml keys for cross-build environment

hoodmane:
This moves unisolation into a package key `cross-build-env: true` means the package is part of the cross build environment and should be unisolated. `cross-build-files` gives a list of files that should be copied from the built copy of the package into the host copy of the package.

This will allow us to construct a cross build environment automatically as part of building packages. If we have these files and the Python `include` directory, this is sufficient for cross-building packages binary packages.

### DIFF
--- a/pyodide-build/pyodide_build/buildpkg.py
+++ b/pyodide-build/pyodide_build/buildpkg.py
@@ -136,6 +136,8 @@ def get_bash_runner():
             "NUMPY_LIB",
             "PYODIDE_PACKAGE_ABI",
             "HOSTINSTALLDIR",
+            "TARGETINSTALLDIR",
+            "SYSCONFIG_NAME",
             "HOSTSITEPACKAGES",
             "PYMAJOR",
             "PYMINOR",
@@ -145,6 +147,7 @@ def get_bash_runner():
             "SIDE_MODULE_LDFLAGS",
             "STDLIB_MODULE_CFLAGS",
             "OPEN_SSL_ROOT",
+            "UNISOLATED_PACKGES",
         ]
     } | {"PYODIDE": "1"}
     if "PYODIDE_JOBS" in os.environ:
@@ -454,6 +457,7 @@ def package_wheel(
     srcpath: Path,
     build_metadata: dict[str, Any],
     bash_runner: BashRunnerWithSharedEnvironment,
+    host_install_dir,
 ):
     """Package a wheel
 
@@ -507,6 +511,17 @@ def package_wheel(
         if result.returncode != 0:
             print("ERROR: post failed")
             exit_with_stdio(result)
+
+    python_dir = f"python{sys.version_info.major}.{sys.version_info.minor}"
+    host_site_packages = Path(host_install_dir) / f"lib/{python_dir}/site-packages"
+    if build_metadata.get("cross-build-env"):
+        subprocess.check_call(["pip", "install", "-t", str(host_site_packages),  f"{name}=={ver}"])
+    
+    cross_build_files : list[str] = build_metadata.get("cross-build-files")
+    if cross_build_files:
+        for file in cross_build_files:
+            shutil.copy((wheel_dir / file), host_site_packages/file)
+
 
     test_dir = distdir / "tests"
     nmoved = 0
@@ -758,6 +773,7 @@ def build_package(
                 srcpath,
                 build_metadata,
                 bash_runner,
+                host_install_dir
             )
 
         shutil.rmtree(pkg_root / "dist", ignore_errors=True)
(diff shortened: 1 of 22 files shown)


pyodide-build/pyodide_build/buildpkg.py
```diff
@@ -145,6 +146,7 @@ def get_bash_runner():
             "SIDE_MODULE_LDFLAGS",
             "STDLIB_MODULE_CFLAGS",
             "OPEN_SSL_ROOT",
+            "UNISOLATED_PACKGES",
```
ryanking13, line 149:
```suggestion
            "UNISOLATED_PACKAGES",
```

pyodide-build/pyodide_build/common.py
```diff
@@ -203,3 +202,46 @@ def search_pyodide_root(curdir: str | Path, *, max_depth: int = 5) -> Path:
...
+def get_pyodide_root():
+    init_environment()
+    return Path(os.environ["PYODIDE_ROOT"])
+
+
+@functools.cache
+def get_unisolated_packages():
+    import json
+
+    if "UNISOLATED_PACKGES" in os.environ:
```
ryanking13, line 238:
```suggestion
    if "UNISOLATED_PACKAGES" in os.environ:
```

pyodide-build/pyodide_build/common.py
```diff
@@ -203,3 +202,46 @@ def search_pyodide_root(curdir: str | Path, *, max_depth: int = 5) -> Path:
...
+        return json.loads(os.environ["UNISOLATED_PACKGES"])
```
ryanking13, line 239:
```suggestion
        return json.loads(os.environ["UNISOLATED_PACKAGES"])
```

pyodide-build/pyodide_build/common.py
```diff
@@ -203,3 +202,46 @@ def search_pyodide_root(curdir: str | Path, *, max_depth: int = 5) -> Path:
...
+    PYODIDE_ROOT = get_pyodide_root()
+    unisolated_packages = []
+    for pkg in (PYODIDE_ROOT / "packages").glob("**/*.meta"):
+        config = parse_package_config(pkg, False)
+        if config.get("build", {}).get("cross-build-env", False):
+            unisolated_packages.append(config["package"]["name"])
+    os.environ["UNISOLATED_PACKGES"] = json.dumps(unisolated_packages)
```
ryanking13, line 246:
```suggestion
    os.environ["UNISOLATED_PACKAGES"] = json.dumps(unisolated_packages)
```

pyodide-build/pyodide_build/common.py
```diff
@@ -203,3 +202,46 @@ def search_pyodide_root(curdir: str | Path, *, max_depth: int = 5) -> Path:
...
+
+@functools.cache
+def get_unisolated_packages():
+    import json
+
+    if "UNISOLATED_PACKGES" in os.environ:
+        return json.loads(os.environ["UNISOLATED_PACKGES"])
+    PYODIDE_ROOT = get_pyodide_root()
+    unisolated_packages = []
+    for pkg in (PYODIDE_ROOT / "packages").glob("**/*.meta"):
```
ryanking13, line 242: Shouldn't this be:

```suggestion
    for pkg in (PYODIDE_ROOT / "packages").glob("**/meta.yaml"):
```

It's interesting why it works...?

pyodide-build/pyodide_build/pypabuild.py
```diff
@@ -62,7 +60,7 @@ def install_reqs(env: IsolatedEnv, reqs: set[str]):
     # only recythonize if it is present. We need them to always recythonize so
     # we always install cython. If the reqs included some cython version already
     # then this won't do anything.
-    env.install(["cython"])
+    env.install(["cython", "pythran"])
```
ryanking13, line 63: If pythran is installed here, do we still need to install it in scipy?

ryanking13 (commented):
Thanks, @hoodmane! I have some minor comments.

pyodide-build/pyodide_build/pypabuild.py
hoodmane: This change should be reverted for now because it doesn't actually work.

hoodmane:

pyodide-build/pyodide_build/pypabuild.py
hoodmane: Currently the issue is that we are using the `host_site_packages` one for the C include path but importing from the pypa/build virtual environment. The problem is that we have no obvious way of pointing the C include path into the pypa/build virtual environment.

ryanking13 (approved):
Thanks, @hoodmane!

hoodmane:
Thanks for the review @ryanking13! We are very close to getting out of tree builds to work.

PR #2525, opened 2022-05-08
Add pygmt

weiji14:
### Description

Placeholder for adding [PyGMT](https://www.pygmt.org) (a library for processing geospatial data and making maps) to Pyodide.

TODO, include 2 runtime C dependencies, xref https://github.com/conda-forge/pygmt-feedstock/blob/b7ecf50df60e8e15cbf5d47122a7beb4ce349d9c/recipe/meta.yaml#L24-L31:
- [ ] gmt (https://github.com/GenericMappingTools/gmt)
- [ ] netCDF4, xref #1582

Other Python dependencies:
- [x] numpy >=1.19
- [x] packaging
- [x] pandas
- [x] xarray (https://github.com/pyodide/pyodide/pull/2538)

### Notes on testing this branch

1. Follow https://pyodide.org/en/latest/development/new-packages.html, get `pyodide` development environment set-up
2. `git clone` this branch
3. `cd packages/pygmt/`
4. `python -m pyodide_build buildpkg meta.yaml`
5. `cd ../../`
6. `PYODIDE_PACKAGES="pygmt" make`
  - May need to do `sudo apt install gfortran f2c ccache build-essential` and so on, ref https://pyodide.org/en/latest/development/building-from-sources.html#partial-builds
7. ...

### Checklists

- [ ] Add a [CHANGELOG](https://github.com/pyodide/pyodide/blob/main/docs/project/changelog.md) entry
- [ ] Add / update tests
- [ ] Add new / update outdated documentation

PR #2538, opened 2022-05-10
Add xarray

weiji14:
### Description

Package for working with N-D labeled arrays and datasets in Python! Repo is at https://github.com/pydata/xarray

Crossref conda-forge recipe file: https://github.com/conda-forge/xarray-feedstock/blob/51aa230d50427cd3a93a8708fad8920e92ce3d62/recipe/meta.yaml

### Checklists

- [ ] Add / update tests

### DIFF
--- a/packages/xarray/meta.yaml
+++ b/packages/xarray/meta.yaml
@@ -1,0 +1,14 @@
+package:
+  name: xarray
+  version: 2022.3.0
+source:
+  url: https://files.pythonhosted.org/packages/4e/6e/2d2430e021fcce47771dcb168599b3d71f26f63a89bd3ff78f56759f1701/xarray-2022.3.0-py3-none-any.whl
+  sha256: 560f36eaabe7a989d5583d37ec753dd737357aa6a6453e55c80bb4f92291a69e
+test:
+  imports:
+    - xarray
+about:
+  home: https://github.com/pydata/xarray
+  PyPI: https://pypi.org/project/xarray
+  summary: N-D labeled arrays and datasets in Python
+  license: Apache
--- a/packages/xarray/meta.yaml
+++ b/packages/xarray/meta.yaml
@@ -10,7 +10,7 @@ requirements:
     - numpy
     - packaging
     - pandas
-    - typing_extensions
+    # - typing_extensions
 test:
   imports:
     - xarray
--- a/packages/xarray/meta.yaml
+++ b/packages/xarray/meta.yaml
@@ -4,6 +4,13 @@ package:
 source:
   url: https://files.pythonhosted.org/packages/4e/6e/2d2430e021fcce47771dcb168599b3d71f26f63a89bd3ff78f56759f1701/xarray-2022.3.0-py3-none-any.whl
   sha256: 560f36eaabe7a989d5583d37ec753dd737357aa6a6453e55c80bb4f92291a69e
+requirements:
+  run:
+    # - importlib_metadata
+    - numpy
+    - packaging
+    - pandas
+    - typing_extensions
 test:
   imports:
     - xarray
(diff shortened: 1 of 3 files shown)


hoodmane:
You need to require `typing-extensions` with a dash instead of an underscore.

packages/xarray/meta.yaml
```diff
@@ -0,0 +1,21 @@
...
+source:
+  url: https://files.pythonhosted.org/packages/4e/6e/2d2430e021fcce47771dcb168599b3d71f26f63a89bd3ff78f56759f1701/xarray-2022.3.0-py3-none-any.whl
+  sha256: 560f36eaabe7a989d5583d37ec753dd737357aa6a6453e55c80bb4f92291a69e
+requirements:
+  run:
+    # - importlib_metadata
+    - numpy
+    - packaging
+    - pandas
+    # - typing_extensions
```
weiji14, line 13: > You need to require `typing-extensions` with a dash instead of an underscore.

Thanks for the pointer @hoodmane. I'm actually not 100% sure if `typing-extensions` is needed at runtime. According to https://github.com/pydata/xarray/blob/9b4d0b29c319f4b68f89328b1bf558711f339504/setup.cfg#L77-L80, it seems like only `numpy`, `pandas` and `packaging` are required? It might just be that `xarray`'s conda-forge `recipe/meta.yaml` file hasn't been updated for a while.

dopplershift:
General question: Does adding a pure Python package like this add some big benefit over just installing with micropip?

weiji14:
> General question: Does adding a pure Python package like this add some big benefit over just installing with micropip?

Not sure, I was just trying to add a package that depends on `xarray` at https://github.com/pyodide/pyodide/pull/2525/commits/dde47d75eb19c72b5c8d59beb61c7068b70af3c5 and there was this error in the CI https://app.circleci.com/pipelines/github/pyodide/pyodide/4066/workflows/1a49c5b5-9111-425d-a055-c4ef255e2091/jobs/44836?invite=true#step-104-57:

```
ValueError: Directory /root/repo/packages/xarray does not contain meta.yaml
```

so I thought to add it in here. Is there an alternative way to add dependencies installable with `micropip`? I'm still very new to pyodide actually.

packages/xarray/meta.yaml
weiji14: So it seems to work without `typing-extensions`? `test_packages_common.py::import[firefox-xarray] PASSED                   [ 43%]`, see https://app.circleci.com/pipelines/github/pyodide/pyodide/4083/workflows/39debe58-bf65-4fe9-a1dc-4b96350cf193/jobs/45126?invite=true#step-102-157

rth:
Yes, generally we are trying to avoid including pure Python packages in favor of using micropip https://github.com/pyodide/pyodide/issues/643 except when another package with C extensions needs it. Though even that would be resolved once #2132 is merged (unless it's a real build dependency)

The only advantage of including it is unvendoring of tests as a separate package. In the case of xarray, that's probably non-neglible (cf `build-packages-numpy-dependents` CI job, check_size step)
```
-rw-r--r-- 1 root root 507K May 10 18:13 xarray-2022.3.0-py3-none-any.whl
-rw-r--r-- 1 root root 2.0M May 10 18:13 xarray-tests.tar
```
it would be less after compression (the xarray wheel on Pypi is 870 kB), but still a not that small.

So for now we would probably need to package it indeed.

rth (approved):
Thanks LGTM. 
Could you add a changelog entry please?

packages/xarray/meta.yaml
rth: Yes, it's an optional dependency for py <3.10 https://github.com/pydata/xarray/blob/1941695ade6a3e4e46a07e6604f29bda5176de26/xarray/core/utils.py#L300 so it should be fine not to include it.

hoodmane:
Could you add a `test_xarray` with a few basic tests?

hoodmane (approved):
Will merge if CI passes.

hoodmane:
Thanks @weiji14!

weiji14:
Cheers, thanks everyone!

PR #2541, opened 2022-05-11
Allow using other decorators with `@run_in_pyodide`

hoodmane:
Allow using `pytest.mark.xfail`, `pytest.mark.parametrize`, hypothesis, and other decorators in `run_in_pyodide` tests

### Checklists

- [x] Add / update tests


pyodide-test-runner/pyodide_test_runner/decorator.py
```diff
@@ -1,68 +1,264 @@
 import ast
 import pickle
 import sys
-import traceback
+from ast import (
```
rth, line 4: Maybe we could still use them as `ast.<name>`?  it's a really short module name anyway. 

For instance, I find things like `Tuple` alone a bit confusing as it clashes with  `typing.Tuple` which people are more used seeing.

rth (commented):
Thanks, that's a nice feature!

Overall sounds good, but I haven't reviewed it in detail and it's a long PR.
I'll try to review soon unless Gyeongjae reviews it before.

pyodide-test-runner/pyodide_test_runner/decorator.py
hoodmane: I went this way because I was copying the output from `pprintast` which does not include the `ast.` prefix which requires pasting `ast.` into a LOT of places in complicated ast chunks. But now that it works, the other way probably does make more sense.

I may also look into using ast templating like here:
https://github.com/pyodide/pyodide/blob/8cfd9911af7754fd9a0f568de5f291aafe5ccf9a/src/py/_pyodide/_base.py#L104
 to simplify some of the ast construction.

hoodmane:
This could also be simplified a bit if we required `@run_in_pyodide` to be the innermost decorators.

hoodmane:
@rth @ryanking13 I think I reduced the diff enough that this should be reasonable to review now. It would be helpful if someone could review it, once we get this in almost all of the test suite can be rewritten using `run_in_pyodide` which should be a pleasant improvement.

pyodide-test-runner/pyodide_test_runner/decorator.py
```diff
@@ -44,17 +113,15 @@ def __new__(cls, function: Callable | None = None, /, **kwargs):
 
     def __init__(
         self,
-        *,
-        standalone: bool = False,
-        module_scope: bool = False,
+        selenium_fixture_name: str = "selenium",
```
ryanking13, line 116: Please update the docstring of `__init__` function according to updated arguments

rth (approved):
LGTM, thanks!

ryanking13 (approved):
Thanks a lot @hoodmane!

PR #2548, opened 2022-05-11
PKG Add msprime and deps including libgsl

benjeffery:
Add the msprime package, a wieldy used genetic simulation tool.

- [X] Add / update tests

### DIFF
--- a/packages/msprime/meta.yaml
+++ b/packages/msprime/meta.yaml
@@ -1,0 +1,31 @@
+package:
+  name: msprime
+  version: 1.1.1
+source:
+  url: https://files.pythonhosted.org/packages/30/ac/22b9d2dc6e3233cc35e613b714d52cf87e4af00064a53d226824401629cb/msprime-1.1.1.tar.gz
+  sha256: e840bed2ba37ae572b234f0b55af006dc408881593f184304481ee6058ea6f30
+build:
+  script: |
+    pip install -t $HOSTSITEPACKAGES numpy
+  cflags: |
+    -I$(PYODIDE_ROOT)/packages/libgsl/build/libgsl-2.7
+  ldflags: |
+    -L$(PYODIDE_ROOT)/packages/libgsl/build/libgsl-2.7/.libs
+    -L$(PYODIDE_ROOT)/packages/libgsl/build/libgsl-2.7/cblas/.libs
+requirements:
+  run:
+    - numpy
+    - newick
+    - tskit
+    - demes
+    - libgsl
+test:
+  imports:
+    - msprime
+about:
+  home: https://tskit.dev/msprime
+  PyPI: https://pypi.org/project/msprime
+  summary:
+    Simulate genealogical trees and genomic sequence data using population
+    genetic models
+  license: GNU GPLv3+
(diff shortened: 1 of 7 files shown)


benjeffery:
This PR is now ready for review/merge. Thanks again for the great docs and tooling, I was able to get `libgsl` compiled and added as a package quite easily. Quite a few other libaraies use GSL, so hopefully that saves some work for other people.

packages/msprime/test_msprime.py
```diff
@@ -0,0 +1,26 @@
...
+    packages=[
+        "numpy",
+        "svgwrite",
+        "attrs",
+        "pyrsistent",
+        "jsonschema",
+        "tskit",
+        "demes",
+        "msprime",
+        "newick",
```
rth, line 14: Normally you don't need to specify packages that are already dependencies of msprime. It's probably a good idea to omit those to check that dependencies are correctly installed.

packages/msprime/meta.yaml
```diff
@@ -0,0 +1,31 @@
+package:
+  name: msprime
+  version: 1.1.1
+source:
+  url: https://files.pythonhosted.org/packages/30/ac/22b9d2dc6e3233cc35e613b714d52cf87e4af00064a53d226824401629cb/msprime-1.1.1.tar.gz
+  sha256: e840bed2ba37ae572b234f0b55af006dc408881593f184304481ee6058ea6f30
+build:
+  script: |
+    pip install -t $HOSTSITEPACKAGES numpy
```
rth, line 9: So numpy is a build dependency as far as I understand? 

It should already be installed on when  building this package. So either it should build without this line altogether, or if it doesn't work you might need to adjust cflags/ldfags to include the right folders see for example https://github.com/pyodide/pyodide/blob/dc29ae65416e624e1fc2bc3b48c65f71a08ce99a/packages/scipy/meta.yaml#L42

rth (commented):
Thanks for doing this work @benjeffery ! A few comments otherwise LGTM.

packages/msprime/meta.yaml
benjeffery: Due to the way msprime is currently built it's `setup.py` imports `numpy`. This is planned to be fixed soon, at which point I will remove this. I've opened https://github.com/tskit-dev/msprime/issues/2055 to track.

rth: Right, so numpy is a build time dependency -- we have many such packages packaged (e.g. scipy, pandas), it's not an issue as far as Pyodide is concerned. My previous suggestions should still apply.

Also for the future if you could avoid rebasing and add commits instead -- it's easier to see what changed since last review, and we squash commits when merging anyway. Thanks!

hoodmane: This shouldn't be necessary in any case because it is already done by building the `numpy` package. So having it in the `run` requirements below covers it. Hopefully we will add `build` requirements and then `numpy` could be moved there.

hoodmane:
This will also need to be fixed for #2551 but maybe we can merge this first and make @ryanking13 deal with it.

hoodmane:
Thanks @benjeffery!

ryanking13:
> This will also need to be fixed for #2551 but maybe we can merge this first and make @ryanking13 deal with it.

Sure, I can deal with it. Thanks @hoodmane and @benjeffery!

PR #2550, opened 2022-05-12
Add pyodide_build create_xbuildenv and install_xbuildenv

hoodmane:
This adds `pyodide_build` command `create_xbuildenv` which creates a crossbuild environment (from a copy of Pyodide where scipy has been build) and `install_xbuildenv` which installs the cross build environment into a fresh copy of Pyodide.

I successfully installed the xbuild environment into a fresh checkout of Pyodide then built statsmodels and scikit-learn in isolation, without building the Python interpreter, numpy, or scipy. I dumped the generated wheels into a copy of Pyodide downloaded from CI, and was able to import and use them as normal.

The size of the xbuild environment is 1.5 megabytes, of which 1.2 megabytes is Python headers.

In a subsequent PR, we can update the CI to automatically upload these to aws s3 and then install the environment from there.

### DIFF
--- a/pyodide-build/pyodide_build/create_xbuildenv.py
+++ b/pyodide-build/pyodide_build/create_xbuildenv.py
@@ -18,7 +18,7 @@ def make_parser(parser: argparse.ArgumentParser):
 def copy_xbuild_files(xbuildenv_path):
     PYODIDE_ROOT = get_pyodide_root()
     site_packages = Path(get_make_flag("HOSTSITEPACKAGES"))
-    xbuild_site_packages = xbuildenv_path / "site_packages"
+    xbuild_site_packages = xbuildenv_path / "site-packages"
     for pkg in (PYODIDE_ROOT / "packages").glob("**/meta.yaml"):
         config = parse_package_config(pkg, check=False)
         xbuild_files = config.get("build", {}).get("cross-build-files", [])
--- a/pyodide-build/pyodide_build/create_xbuildenv.py
+++ b/pyodide-build/pyodide_build/create_xbuildenv.py
@@ -1,0 +1,42 @@
+import argparse
+import shutil
+import subprocess
+from pathlib import Path
+
+from .common import get_make_flag, get_pyodide_root, parse_package_config
+
+
+def make_parser(parser: argparse.ArgumentParser):
+    parser.description = (
+        "Create xbuild env.\n\n"
+        "Note: this is a private endpoint that should not be used "
+        "outside of the Pyodide Makefile."
+    )
+    return parser
+
+
+def copy_xbuild_files(xbuildenv_path):
+    PYODIDE_ROOT = get_pyodide_root()
+    site_packages = Path(get_make_flag("HOSTSITEPACKAGES"))
+    xbuild_site_packages = xbuildenv_path / "site_packages"
+    for pkg in (PYODIDE_ROOT / "packages").glob("**/meta.yaml"):
+        config = parse_package_config(pkg, False)
+        xbuild_files = config.get("build", {}).get("cross-build-files", [])
+        for path in xbuild_files:
+            target = xbuild_site_packages / path
+            target.parent.mkdir(parents=True, exist_ok=True)
+            shutil.copy(site_packages / path, target)
+
+
+def main(args):
+    pyodide_root = get_pyodide_root()
+    xbuildenv_path = pyodide_root / "xbuildenv"
+    shutil.rmtree(xbuildenv_path, ignore_errors=True)
+    xbuildenv_path.mkdir()
+    shutil.copytree(get_make_flag("PYTHONINCLUDE"), xbuildenv_path / "python-include")
+    copy_xbuild_files(xbuildenv_path)
+    res = subprocess.run(
+        ["pip", "freeze", "--path", get_make_flag("HOSTSITEPACKAGES")],
+        stdout=subprocess.PIPE,
+    )
+    (xbuildenv_path / "requirements.txt").write_bytes(res.stdout)
--- a/pyodide-build/pyodide_build/create_xbuildenv.py
+++ b/pyodide-build/pyodide_build/create_xbuildenv.py
@@ -20,7 +20,7 @@ def copy_xbuild_files(xbuildenv_path):
     site_packages = Path(get_make_flag("HOSTSITEPACKAGES"))
     xbuild_site_packages = xbuildenv_path / "site_packages"
     for pkg in (PYODIDE_ROOT / "packages").glob("**/meta.yaml"):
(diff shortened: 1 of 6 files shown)


pyodide-build/pyodide_build/create_xbuildenv.py
```diff
@@ -0,0 +1,42 @@
...
+    parser.description = (
+        "Create xbuild env.\n\n"
+        "Note: this is a private endpoint that should not be used "
+        "outside of the Pyodide Makefile."
+    )
+    return parser
+
+
+def copy_xbuild_files(xbuildenv_path):
+    PYODIDE_ROOT = get_pyodide_root()
```
hoodmane, line 19: We should also check that all the relevant packages have been built.

rth:
Thanks for this work! 

> cross build environment

In terms of vocabulary, I find it a bit confusing that we are not always using this environment to cross compile: maybe some of the confusion is between cross-build which would be done only with this command and cross-compile which we always do. And we talked about this for some time, so I can only imagine that other contributors might be lost here.  We might need to find a better way to name those.

> which creates a crossbuild environment (from a copy of Pyodide where scipy has been build) and install_xbuildenv which installs the cross build environment into a fresh copy of Pyodide.

Another thing that is not very clear to me, is that imagine we cache these files. So then there is a command (`install_xbuildenv`  ) to copy these files into a new Pyodide source tree. However, the user would never actually use `create_xbuildenv` (or do anything with the created env): it would be purely an internal command that would be used to create a folder/archive of files to cache right?

pyodide-build/pyodide_build/install_xbuildenv.py
```diff
@@ -0,0 +1,44 @@
...
+        "Note: this is a private endpoint that should not be used "
+        "outside of the Pyodide Makefile."
+    )
+    return parser
+
+
+def main(args):
+    pyodide_root = get_pyodide_root()
+    host_site_packages = Path(get_make_flag("HOSTSITEPACKAGES"))
+    xbuildenv_path = pyodide_root / "xbuildenv"
```
rth, line 22: Maybe take this as CLI argument, for the case where we download it from somewhere.

pyodide-build/pyodide_build/install_xbuildenv.py
```diff
@@ -0,0 +1,44 @@
...
+            "install",
+            "-t",
+            host_site_packages,
+            "-r",
+            xbuildenv_path / "requirements.txt",
+        ]
+    )
+    shutil.copytree(
+        xbuildenv_path / "site_packages", host_site_packages, dirs_exist_ok=True
+    )
```
rth, line 44: Hmm, but so NumPy would be in the `requirements.txt`, for instance, right? I'm not sure if installing the unpatched numpy version and then overwriting site_packages with the host_site_packages is that reliable and there isn't a chance we would break something.

rth (commented):
A few more comments on the code, and reading the code actually answers some of my previous questions :)
So what exactly are we storing in the cross-build env? I imagine those files are not going to be portable between Linux and MacOS for instance?

rth:
> If we install this environment, we get a setup which is the same as if we'd built Python, numpy, and scipy from source.

Yeah, as long as it's a private API fine with me. And maybe the description in the code could state that overall it's a faster way to get to the identical build configuration as with the regular build. Not some other env used for the build.

pyodide-build/pyodide_build/install_xbuildenv.py
hoodmane: I had in mind to leave this part incomplete in this PR and fix it up in a subsequent PR. My thought was that by default could calculate the URL from the current commit hash and get it from s3.

hoodmane:
> So what exactly are we storing in the cross-build env? I imagine those files are not going to be portable between Linux and MacOS for instance?

We are storing custom C and Cython headers. It should be platform independent because everything is about the capabilities of the target wasm-emscripten abstract machine, not about the build machine.

hoodmane:
Another point that might be worth making is that this is not currently useful with `buildall` only with `buildpkg` because `buildall` will try to rebuild the dependencies. We should ideally get #2132 finished and then figure out a way for `install_xbuildenv` to mark build dependencies as satisfied. 

The idea is that eventually people will be able to download the release `dist` directory, build a single extra binary package in the `pyodide-env` docker image, then put the single extra wheel and a new `packages.json` into the `dist` directory, and be able to use it.

pyodide-build/pyodide_build/install_xbuildenv.py
hoodmane: It's a bit sketchy, but I think it's correct and reproducible. We've locked the two numpy versions to be the same. When we update numpy, we'll update the lock too.

This logic is essentially unchanged here: since #2256, when we build numpy we've been downloading a copy of numpy from pip and writing over `_npyconfig.h`. Before that, we built a fresh native copy of numpy and then wrote over `_npyconfig.h`. The difference here is that we are planning to download both `numpy` and `_npyconfig.h` and paste them together.

pyodide-build/pyodide_build/create_xbuildenv.py
```diff
@@ -0,0 +1,42 @@
...
+        xbuild_files = config.get("build", {}).get("cross-build-files", [])
+        for path in xbuild_files:
+            target = xbuild_site_packages / path
+            target.parent.mkdir(parents=True, exist_ok=True)
+            shutil.copy(site_packages / path, target)
+
+
+def main(args):
+    pyodide_root = get_pyodide_root()
+    xbuildenv_path = pyodide_root / "xbuildenv"
```
ryanking13, line 33: For now, maybe add this path to `.gitignore`?

ryanking13 (approved):
Thanks, @hoodmane. The idea of installing cross-build env (or whatever it will be called later) looks good to me. As you mentioned, this PR seems incomplete on purpose, so I'll review it more thoroughly in a subsequent PR. I am OK with merging, but maybe Roman has some more ideas.

pyodide-build/pyodide_build/install_xbuildenv.py
```diff
@@ -0,0 +1,44 @@
...
+            "pip",
+            "install",
+            "-t",
+            host_site_packages,
+            "-r",
+            xbuildenv_path / "requirements.txt",
+        ]
+    )
+    shutil.copytree(
+        xbuildenv_path / "site_packages", host_site_packages, dirs_exist_ok=True
```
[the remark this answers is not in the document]
hoodmane: This makes sense. We also have to fix the symmetric mistake on line 21 of `create_xbuildenv`.

hoodmane:

PR #2551, opened 2022-05-12
BLD Install libraries to centralized directory

### LINKED ISSUE #2436: BLD Provide a way to easily specify library paths in `meta.yaml`
## Problem

When building Python packages that are dependent on some C/C++libraries,
we often hard-code paths of library headers (.h) / static binaries (.a).

This is annoying because when we update a library, all hard-coded paths need to be updated.

For example, in `lxml` package, there are a bunch of hard-coded paths.

https://github.com/pyodide/pyodide/blob/3f9ae2ed1e1aee688251af86a4dc23d3d867740c/packages/lxml/meta.yaml#L8-L22 
## Possible solutions

1. Put every library headers / binaries into a specific build directory.

e.g. put headers to `e.g. ${PYODIDE_ROOT}/build/pyodide/include` and static binaries to `${PYODIDE_ROOT}/build/pyodide/lib`.

Then inside `meta.yaml`, those can be easily specified such as

```
-I$(PYODIDE_INCLUDE_PATH)
-L$(PYODIDE_LIBRARY_PATH)
-lxml
```

2. Add a variable for each library.

For example,

```
-I$(LIBRARY_libxml)/include
```

instead of,

```
-I$(PYODIDE_ROOT)/packages/libxml/build/libxml-2.9.10/include
```

ryanking13:
### Description

Resolve #2436

Install libraries to `packages/.libs` and use `pkg-config` if possible. After this patch, we don't have to hardcode paths for each library.

In the future, maybe we can try setting [`EM_PKG_CONFIG_PATH`](https://github.com/emscripten-core/emscripten/blob/592f67ff77fea404f07f901b37ec4ff5eb017a25/tools/building.py#L153) variable to make emscripten to automatically detect library paths.

### DIFF
--- a/packages/pyyaml/meta.yaml
+++ b/packages/pyyaml/meta.yaml
@@ -7,7 +7,6 @@ source:
 build:
   cflags: |
     -I$(PYTHONINCLUDE)
-    -I$(PYODIDE_ROOT)/emsdk/emsdk/fastcomp/emscripten/system/include/libc
     -I$(WASM_LIBRARY_DIR)/include
   ldflags: |
     -L$(WASM_LIBRARY_DIR)/lib
--- a/packages/pyyaml/meta.yaml
+++ b/packages/pyyaml/meta.yaml
@@ -8,6 +8,9 @@ build:
   cflags: |
     -I$(PYTHONINCLUDE)
     -I$(PYODIDE_ROOT)/emsdk/emsdk/fastcomp/emscripten/system/include/libc
+    -I$(WASM_LIBRARY_DIR)/libyaml/include
+  ldflags: |
+    -L$(WASM_LIBRARY_DIR)/libyaml/lib
 requirements:
   run:
     - libyaml
--- a/packages/pyyaml/meta.yaml
+++ b/packages/pyyaml/meta.yaml
@@ -8,9 +8,9 @@ build:
   cflags: |
     -I$(PYTHONINCLUDE)
     -I$(PYODIDE_ROOT)/emsdk/emsdk/fastcomp/emscripten/system/include/libc
-    -I$(WASM_LIBRARY_DIR)/libyaml/include
+    -I$(WASM_LIBRARY_DIR)/include
   ldflags: |
-    -L$(WASM_LIBRARY_DIR)/libyaml/lib
+    -L$(WASM_LIBRARY_DIR)/lib
 requirements:
   run:
     - libyaml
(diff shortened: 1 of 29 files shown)


ryanking13:
This PR is ready to be reviewed.

packages/pyyaml/meta.yaml
```diff
@@ -5,10 +5,12 @@ source:
   url: https://files.pythonhosted.org/packages/36/2b/61d51a2c4f25ef062ae3f74576b01638bebad5e045f747ff12643df63844/PyYAML-6.0.tar.gz
   sha256: 68fb519c14306fec9720a2a5b45bc9f0c8d1b9c72adf45c37baedfcd949c35a2
 build:
-  cflags:
-    "-I$(PYTHONINCLUDE)\n-I$(PYODIDE_ROOT)/emsdk/emsdk/fastcomp/emscripten/system/include/libc\n\
-    -I$(PYODIDE_ROOT)/packages/libyaml/build/include\n"
-  ldflags: "-L$(PYODIDE_ROOT)/packages/libyaml/build/\n"
+  cflags: |
+    -I$(PYTHONINCLUDE)
+    -I$(PYODIDE_ROOT)/emsdk/emsdk/fastcomp/emscripten/system/include/libc
```
hoodmane, line 10:
```suggestion
```
This directory hasn't existed since #1102 was merged in January 2021, but I guess we didn't notice.

packages/ssl/meta.yaml
```diff
@@ -8,11 +8,14 @@ build:
   script: |
     mkdir dist
     export DISTDIR=$(pwd)/dist
+    export PKG_CONFIG_PATH=${WASM_PKG_CONFIG_PATH}:${PKG_CONFIG_PATH}
```
hoodmane, line 11: We set this in a large number of packages. Can we set `PKG_CONFIG_PATH` in `buildpkg`?

hoodmane (approved):
Thanks for working on this @ryanking13!

packages/ssl/meta.yaml
ryanking13: Yeah, I'll try that. Let's see if it doesn't conflict with other packages.

hoodmane:
I think this is ready to merge if CI passes.

ryanking13:
Thanks for the review @hoodmane! I'll merge when CI pass.

Issue #1610, opened 2021-05-25
os.name returns `posix`


karthiknadig:
I think this should return something other than `posix` or `nt`.

/cc @int19h

rth:
True, but it's more an issue for CPython. We are just following the docs which says,
```
The following names have currently been registered: 'posix', 'nt', 'java'.

See also

sys.platform has a finer granularity. os.uname() gives system-dependent version information.

The platform module provides detailed checks for the system’s identity.
```
and in the case of Pyodide `sys.platform` returns `"emscripten"`.

rth:
Actually, this is likely the expected behavior. Emscipten does aim to emulate the POSIX API even if the emulation is not perfect.

In any case, we should leave this decision to the CPython team, and currently in the CPython 3.11a7 WASM build https://cheimes.fedorapeople.org/python-wasm/ the behavior is the same.  See related discussion in https://discuss.python.org/t/expected-behavior-for-unsupported-stdlib-modules-in-the-browser/15689/8

So I'm going to close this issue. Thanks for the report.

PR #2552, opened 2022-05-12
DOC update FAQ instruction for detecting Emscripten runtime

rth:
As suggested by @tiran in https://discuss.python.org/t/expected-behavior-for-unsupported-stdlib-modules-in-the-browser/15689/2

I agree that using `sys` is better than `platform`.

### DIFF
--- a/docs/usage/faq.md
+++ b/docs/usage/faq.md
@@ -110,7 +110,7 @@ my_namespace.get("z"); // ==> 4
 
 ## How to detect that code is run with Pyodide?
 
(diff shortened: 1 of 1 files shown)


ryanking13 (approved):
Thanks, @rth Probably we should add a note about #1610 too?

rth:
Well, there is no action needed on that issue I think (I just closed it). If you want feel free to push a comment about it, but `os.name` is not going to help detect running in Emscripten.

PR #2554, opened 2022-05-12
Fix building CPython on macos (Fixes #2360)

### LINKED ISSUE #2360: fail to make libffi.a on MacOS
when I run 
pyodide % sudo make

It looks fine until it gets to this error
source pyodide/emsdk/emsdk/emsdk_env.sh
make[1]: source: No such file or directory
make[1]: *** [pyodide/cpython/installs/python-3.10.2/lib/libffi.a] Error 1
make: *** [pyodide/cpython/installs/python-3.10.2/lib/python3.10] Error 2

when I check I can see that there is in fact no file called libffi.a in 
pyodide/cpython/installs/python-3.10.2/lib/

Any help would be greatly apprecaited!

dopplershift:
### Description

Updates the `Makefile` to avoid separating the `source` command from the build.

### Checklists

- [ ] Add / update tests

### DIFF
--- a/cpython/Makefile
+++ b/cpython/Makefile
@@ -119,8 +119,7 @@ $(INSTALL)/lib/libbz2.a: $(BZIP2TARBALL)
 $(INSTALL)/lib/libffi.a :
 	rm -rf $(FFIBUILD)
 	git clone $(LIBFFIREPO) $(FFIBUILD) --shallow-exclude upstream-base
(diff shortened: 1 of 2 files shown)


hoodmane:
Thanks @dopplershift!

rth:
Thanks for fixing it @dopplershift !

PR #2555, opened 2022-05-13
Add PyProj to packages

dopplershift:
### Description

This adds PyProj to the packages, as well as its dependency PROJ (libproj). This required pointing directly to the sqlite3.wasm that's built with Python (or else the build refused to detect the dependency as present).

I also tried to build as a shared library, but it said that wasn't supported. 🤷‍♂️ 

### Checklists

- [ ] Add / update tests

### DIFF
--- a/packages/certifi/meta.yaml
+++ b/packages/certifi/meta.yaml
@@ -1,0 +1,17 @@
+package:
+  name: certifi
+  version: 2022.6.15
+
+source:
+  url: https://files.pythonhosted.org/packages/e9/06/d3d367b7af6305b16f0d28ae2aaeb86154fa91f144f036c2d5002a5a202b/certifi-2022.6.15-py3-none-any.whl
+  sha256: fe86415d55e84719d75f8b69414f6438ac3547d2078ab91b67e779ef69378412
+
+test:
+  imports:
+    - certifi
+
+about:
+  home: https://github.com/certifi/python-certifi
+  PyPI: https://pypi.org/project/certifi
+  summary: Certifi provides Mozilla’s carefully curated collection of Root Certificates for validating the trustworthiness of SSL certificates while verifying the identity of TLS hosts.
+  license: "License :: OSI Approved :: Mozilla Public License 2.0 (MPL 2.0)"
--- a/packages/libproj/meta.yaml
+++ b/packages/libproj/meta.yaml
@@ -17,7 +17,7 @@ build:
     mkdir -p build;
     cd build \
         && LDFLAGS="-s NODERAWFS=1 -sUSE_ZLIB=1 -sFORCE_FILESYSTEM=1" emcmake cmake ../ \
-        -DCMAKE_INSTALL_PREFIX=../../proj \
+        -DCMAKE_INSTALL_PREFIX=$WASM_LIBRARY_DIR \
         -DCMAKE_BUILD_TYPE=Release \
         -DENABLE_CURL=OFF \
         -DBUILD_APPS=OFF \
--- a/packages/libproj/meta.yaml
+++ b/packages/libproj/meta.yaml
@@ -9,6 +9,7 @@ source:
 requirements:
   run:
     - sqlite3
+    - libtiff
 
 build:
   library: true
@@ -19,10 +20,11 @@ build:
         -DCMAKE_INSTALL_PREFIX=../../proj \
         -DCMAKE_BUILD_TYPE=Release \
         -DENABLE_CURL=OFF \
-        -DENABLE_TIFF=OFF \
         -DBUILD_APPS=OFF \
         -DBUILD_SHARED_LIBS=OFF \
         -DBUILD_TESTING=OFF \
+        -DTIFF_INCLUDE_DIR=$WASM_LIBRARY_DIR/include \
+        -DTIFF_LIBRARY=$WASM_LIBRARY_DIR/lib/libtiff.a \
         -DSQLITE3_LIBRARY=${WASM_LIBRARY_DIR}/lib/libsqlite3.a \
         -DSQLITE3_INCLUDE_DIR=${WASM_LIBRARY_DIR}/include \
         -DCMAKE_C_FLAGS="-fPIC" \
--- a/packages/libproj/meta.yaml
+++ b/packages/libproj/meta.yaml
@@ -1,0 +1,30 @@
+package:
+  name: libproj
+  version: 8.2.1
+
+source:
+  sha256: 76ed3d0c3a348a6693dfae535e5658bbfd47f71cb7ff7eb96d9f12f7e068b1cf
+  url: https://download.osgeo.org/proj/proj-8.2.1.tar.gz
+
+requirements:
+  run:
+    - sqlite3
+
+build:
+  library: true
+  script: |
+    mkdir -p build;
+    cd build \
+        && LDFLAGS="-s NODERAWFS=1 -sUSE_ZLIB=1 -sFORCE_FILESYSTEM=1" emcmake cmake ../ \
+        -DCMAKE_INSTALL_PREFIX=../../proj \
+        -DCMAKE_BUILD_TYPE=Release \
+        -DENABLE_CURL=OFF \
+        -DENABLE_TIFF=OFF \
+        -DBUILD_APPS=OFF \
+        -DBUILD_SHARED_LIBS=OFF \
+        -DBUILD_TESTING=OFF \
+        -DSQLITE3_LIBRARY=${WASM_LIBRARY_DIR}/lib/libsqlite3.a \
+        -DSQLITE3_INCLUDE_DIR=${WASM_LIBRARY_DIR}/include \
+        -DCMAKE_C_FLAGS="-fPIC" \
+        -DCMAKE_CXX_FLAGS="-fPIC";
+    emmake make -j ${PYODIDE_JOBS:-3} install;
--- a/packages/pyproj/meta.yaml
+++ b/packages/pyproj/meta.yaml
@@ -19,7 +19,9 @@ build:
     export PROJ_DIR=${WASM_LIBRARY_DIR}
     export PROJ_INCDIR=${WASM_LIBRARY_DIR}/include
     export PROJ_LIBDIR=${WASM_LIBRARY_DIR}/lib
-
+    export PROJ_WHEEL=1
+    mkdir -p pyproj/proj_dir
+    cp -r ${WASM_LIBRARY_DIR}/share pyproj/proj_dir
 test:
   imports:
     - pyproj
--- a/packages/pyproj/meta.yaml
+++ b/packages/pyproj/meta.yaml
@@ -5,6 +5,8 @@ package:
 source:
   url: https://files.pythonhosted.org/packages/e3/4d/348402c2fb0d8a8e85a88b8babc6f4efaae9692b7524aedce5fddbef3baf/pyproj-3.3.1.tar.gz
   sha256: b3d8e14d91cc95fb3dbc03a9d0588ac58326803eefa5bbb0978d109de3304fbe
+  patches:
+    - patches/fix-build.patch
 
 requirements:
   run:
--- a/packages/pyproj/meta.yaml
+++ b/packages/pyproj/meta.yaml
@@ -11,6 +11,7 @@ source:
 requirements:
   run:
     - libproj
+    - certifi
 
 build:
   script: |
--- a/packages/pyproj/meta.yaml
+++ b/packages/pyproj/meta.yaml
@@ -1,0 +1,28 @@
+package:
+  name: pyproj
+  version: 3.3.1
+
+source:
+  url: https://files.pythonhosted.org/packages/e3/4d/348402c2fb0d8a8e85a88b8babc6f4efaae9692b7524aedce5fddbef3baf/pyproj-3.3.1.tar.gz
+  sha256: b3d8e14d91cc95fb3dbc03a9d0588ac58326803eefa5bbb0978d109de3304fbe
+
+requirements:
+  run:
+    - libproj
+
+build:
+  script: |
+    export PROJ_VERSION=8.2.1
+    export PROJ_DIR=${WASM_LIBRARY_DIR}
+    export PROJ_INCDIR=${WASM_LIBRARY_DIR}/include
+    export PROJ_LIBDIR=${WASM_LIBRARY_DIR}/lib
+
+test:
+  imports:
+    - pyproj
+
+about:
+  home: https://github.com/pyproj4/pyproj
+  PyPI: https://pypi.org/project/pyproj/
+  summary: Python interface to PROJ (cartographic projections and coordinate transformations library)
+  license: "License :: OSI Approved :: MIT License"
--- a/packages/pyproj/patches/fix-build.patch
+++ b/packages/pyproj/patches/fix-build.patch
@@ -1,0 +1,12 @@
+diff --git a/setup.py b/setup.py
+index d8c1872a..ad396e02 100644
+--- a/setup.py
++++ b/setup.py
+@@ -167,7 +167,6 @@ def get_extension_modules():
+     ext_options = {
+         "include_dirs": include_dirs,
+         "library_dirs": library_dirs,
+-        "runtime_library_dirs": library_dirs if os.name != "nt" else None,
+         "libraries": get_libraries(library_dirs),
+     }
+     # setup cythonized modules
--- a/docs/project/changelog.md
+++ b/docs/project/changelog.md
@@ -208,7 +208,7 @@ substitutions:
   svgwrite, jsonschema, tskit {pr}`2506`, xarray {pr}`2538`, demes, libgsl, newick,
   ruamel, msprime {pr}`2548`, gmpy2 {pr}`2665`, xgboost {pr}`2537`, galpy {pr}`2676`,
   shapely, geos {pr}`2725`, suitesparse, sparseqr {pr}`2685`, libtiff {pr}`2762`,
-  pytest-benchmark {pr}`2799`, termcolor {pr}`2809`
+  pytest-benchmark {pr}`2799`, termcolor {pr}`2809`, sqlite3, libproj, pyproj, certifi {pr}`2555`
 
 ### Miscellaneous
 
--- a/packages/pyproj/test_pyproj.py
+++ b/packages/pyproj/test_pyproj.py
@@ -1,0 +1,19 @@
+from pyodide_test_runner import run_in_pyodide
+
+
+@run_in_pyodide(packages=["pyproj"])
+def test_pyproj(selenium):
+    from pyproj import CRS, Transformer
+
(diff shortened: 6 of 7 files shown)


dopplershift:
The build is failing with:
```
-- Configuring incomplete, errors occurred!
See also "/root/repo/packages/libproj/build/libproj-8.2.1/build/CMakeFiles/CMakeOutput.log".
emcmake: error: 'cmake ../ -DCMAKE_INSTALL_PREFIX=../../proj -DCMAKE_BUILD_TYPE=Release -DENABLE_CURL=OFF -DENABLE_TIFF=OFF -DBUILD_APPS=OFF -DBUILD_SHARED_LIBS=OFF -DBUILD_TESTING=OFF -DSQLITE3_LIBRARY=/root/repo/cpython/build/sqlite-autoconf-3380500/sqlite3.wasm -DSQLITE3_INCLUDE_DIR=/root/repo/cpython/build/sqlite-autoconf-3380500/ -DCMAKE_C_FLAGS=-fPIC -DCMAKE_CXX_FLAGS=-fPIC -DCMAKE_TOOLCHAIN_FILE=/root/repo/emsdk/emsdk/upstream/emscripten/cmake/Modules/Platform/Emscripten.cmake -DCMAKE_CROSSCOMPILING_EMULATOR=/root/repo/emsdk/emsdk/node/14.18.2_64bit/bin/node' failed (returned 1)
[2022-05-13 17:12:51] Building package libproj...
ERROR: bash command failed
    mkdir -p build;
    cd build \
        && LDFLAGS="-s NODERAWFS=1 -sUSE_ZLIB=1 -sFORCE_FILESYSTEM=1" emcmake cmake ../ \
        -DCMAKE_INSTALL_PREFIX=../../proj \
        -DCMAKE_BUILD_TYPE=Release \
        -DENABLE_CURL=OFF \
        -DENABLE_TIFF=OFF \
        -DBUILD_APPS=OFF \
        -DBUILD_SHARED_LIBS=OFF \
        -DBUILD_TESTING=OFF \
        -DSQLITE3_LIBRARY=$PYODIDE_ROOT/cpython/build/sqlite-autoconf-3380500/sqlite3.wasm \
        -DSQLITE3_INCLUDE_DIR=$PYODIDE_ROOT/cpython/build/sqlite-autoconf-3380500/ \
        -DCMAKE_C_FLAGS="-fPIC" \
        -DCMAKE_CXX_FLAGS="-fPIC";
    emmake make -j ${PYODIDE_JOBS:-3} install;

[2022-05-13 17:13:03] Succeeded building package libproj in 13.0 seconds.
ERROR: cancelling buildall
Fatal Python error: _enter_buffered_busy: could not acquire lock for <_io.BufferedWriter name='<stdout>'> at interpreter shutdown, possibly due to daemon threads
Python runtime state: finalizing (tstate=0x0000559dad6cdc10)

Current thread 0x00007efc1ca5e740 (most recent call first):
  <no Python frame>

Extension modules: _ruamel_yaml, yaml._yaml (total: 2)
make: *** [Makefile:13: all] Aborted (core dumped)
make: Leaving directory '/root/repo/packages'

Exited with code exit status 
```
and I have no idea what that would mean...

ryanking13:
That error message is a general error message for the build failure.

I guess you'll need to look at `build.log` of `libproj` or `CMakeOutput.log` as the error message says.

hoodmane:
You should change the CI script to add an extra `store_artifacts` step that stores `packages/libproj` as an artifact:
https://github.com/pyodide/pyodide/blob/dc29ae65416e624e1fc2bc3b48c65f71a08ce99a/.circleci/config.yml#L166

ryanking13:
Actually, `build.log` is printed to the stdout if the build fails, so you can see the CI log instead.

It looks like there is an issue with sqlite3

```
CMake Error at CMakeLists.txt:181 (message):
  sqlite3 binary not found!


CMake Error at cmake/FindSqlite3.cmake:59 (file):
  file STRINGS file
  "/root/repo/cpython/build/sqlite-autoconf-3380500/sqlite3.h" cannot be
  read.
Call Stack (most recent call first):
  CMakeLists.txt:184 (find_package)


-- Found Sqlite3: /root/repo/cpython/build/sqlite-autoconf-3380500/sqlite3.wasm
-- Sqlite3 version: 
CMake Error at CMakeLists.txt:192 (message):
  sqlite3 >= 3.11 required!
```

packages/libproj/meta.yaml
```diff
@@ -0,0 +1,26 @@
...
+    cd build \
+        && LDFLAGS="-s NODERAWFS=1 -sUSE_ZLIB=1 -sFORCE_FILESYSTEM=1" emcmake cmake ../ \
+        -DCMAKE_INSTALL_PREFIX=../../proj \
+        -DCMAKE_BUILD_TYPE=Release \
+        -DENABLE_CURL=OFF \
+        -DENABLE_TIFF=OFF \
+        -DBUILD_APPS=OFF \
+        -DBUILD_SHARED_LIBS=OFF \
+        -DBUILD_TESTING=OFF \
+        -DSQLITE3_LIBRARY=$PYODIDE_ROOT/cpython/build/sqlite-autoconf-3380500/sqlite3.wasm \
```
ryanking13, line 22: I think this should point to a `.a` file.

```suggestion
        -DSQLITE3_LIBRARY=$PYODIDE_ROOT/cpython/build/sqlite-autoconf-3380500/.libs/libsqlite3.a \
```

ryanking13:
BTW, @hoodmane, if some packages require libsqlite3 as a dependency do you think we should move it to packages similar to `openssl` and `ssl`?

hoodmane:
Well it depends on whether we think `sqlite` should always be loaded. But I think ideally we would build `sqlite` as a shared library, only load it on request via `pyodide.loadPackage`, and link it dynamically into other dependent packages.

packages/libproj/meta.yaml
dopplershift: Ah, I didn't think to check `.libs`.

dopplershift:
Ok, I'm not understanding:
```
CMake Error at cmake/FindSqlite3.cmake:59 (file):
  file STRINGS file
  "/root/repo/cpython/build/sqlite-autoconf-3380500/sqlite3.h" cannot be
  read.
Call Stack (most recent call first):
  CMakeLists.txt:184 (find_package)
```
the file is there on my system? How does it have `.libs` but not the include?

ryanking13:
It looks like we remove CPython build artifacts after the build in CI so you can't use sqlite3 library (abd also its headers) when building packages.

https://github.com/pyodide/pyodide/blob/eb4f7d9fd9b5d870ac72783334666a99cd2afe4e/.circleci/config.yml#L90-L94

I think you'll need to add sqlite3 in `packages` just like other libraries.

dopplershift:
Ok, added sqlite3 as a separate package. That seems to be found.

PROJ needs an sqlite3 executable in order to build its internal database. I tried adding this to the `Dockerfile`, but it doesn't seem to have worked. Also, not sure if that's how y'all want to see that accomplished.

ryanking13:
> Ok, added sqlite3 as a separate package. That seems to be found.
> 
> PROJ needs an sqlite3 executable in order to build its internal database. I tried adding this to the `Dockerfile`, but it doesn't seem to have worked. Also, not sure if that's how y'all want to see that accomplished.

Can you try downloading sqlite3 binary during the build (adding a build script section in meta.yaml)? I think we prefer that way than installing it globally.

for example, you can do something like (it is how I did when we need protoc binary during the build):

```yaml
build:
  script: |
    # Note: We need protoc binary during the build
    case "$(uname -s)" in
        Linux*) export MACHINE=linux;;
        Darwin*) export MACHINE=osx;;
        *) export MACHINE="UNKNOWN"
    esac
    
    export PROTOC_VERSION="3.13.0"
    export PROTOC_ZIP="protoc-${PROTOC_VERSION}-${MACHINE}-x86_64.zip"
    export PROTOC_URL="https://github.com/protocolbuffers/protobuf/releases/download/v${PROTOC_VERSION}/${PROTOC_ZIP}"
    
    wget ${PROTOC_URL}
    unzip ${PROTOC_ZIP} -d protoc
    export PATH=${PATH}:$(pwd)/protoc/bin/
```

ryanking13:
> > Ok, added sqlite3 as a separate package. That seems to be found.
> > PROJ needs an sqlite3 executable in order to build its internal database. I tried adding this to the `Dockerfile`, but it doesn't seem to have worked. Also, not sure if that's how y'all want to see that accomplished.
> 
> Can you try downloading sqlite3 binary during the build (adding a build script section in meta.yaml)? I think we prefer that way than installing it globally.
> 
> for example, you can do something like (it is how I did when we need protoc binary during the build):
> 
> ```yaml
> build:
>   script: |
>     # Note: We need protoc binary during the build
>     case "$(uname -s)" in
>         Linux*) export MACHINE=linux;;
>         Darwin*) export MACHINE=osx;;
>         *) export MACHINE="UNKNOWN"
>     esac
>     
>     export PROTOC_VERSION="3.13.0"
>     export PROTOC_ZIP="protoc-${PROTOC_VERSION}-${MACHINE}-x86_64.zip"
>     export PROTOC_URL="https://github.com/protocolbuffers/protobuf/releases/download/v${PROTOC_VERSION}/${PROTOC_ZIP}"
>     
>     wget ${PROTOC_URL}
>     unzip ${PROTOC_ZIP} -d protoc
>     export PATH=${PATH}:$(pwd)/protoc/bin/
> ```

packages/libproj/meta.yaml
```diff
@@ -0,0 +1,44 @@
...
+  url: https://download.osgeo.org/proj/proj-8.2.1.tar.gz
+
+requirements:
+  run:
+    - sqlite3
+
+build:
+  library: true
+  script: |
+    # Note: We need protoc binary during the build
```
ryanking13, line 16:
```suggestion
```

packages/libproj/meta.yaml
```diff
@@ -0,0 +1,44 @@
...
+        Darwin*) export MACHINE=osx;;
+        *) export MACHINE="UNKNOWN"
+    esac
+
+    export SQLITE_VERSION="3380500"
+    export SQLITE_ZIP="sqlite-tools-${MACHINE}-x86-${SQLITE_VERSION}.zip"
+    export SQLITE_URL="https://www.sqlite.org/2022/${SQLITE_ZIP}"
+    wget ${SQLITE_URL}
+    unzip ${SQLITE_ZIP} -d sqlite
+    export PATH=${PATH}:$(pwd)/sqlite/bin/
```
ryanking13, line 28: It looks like there is a subdirectory inside a sqlite zip file (maybe this is wrong, please check by yourself):

```suggestion
    export PATH=${PATH}:$(pwd)/sqlite/sqlite-tools-${MACHINE}-x86-${SQLITE_VERSION}
```

packages/libproj/meta.yaml
```diff
@@ -0,0 +1,45 @@
...
+  name: libproj
+  version: 8.2.1
+
+source:
+  sha256: 76ed3d0c3a348a6693dfae535e5658bbfd47f71cb7ff7eb96d9f12f7e068b1cf
+  url: https://download.osgeo.org/proj/proj-8.2.1.tar.gz
+
+requirements:
+  run:
+    - sqlite3
```
ryanking13, line 11: I think you need to add libtiff here:

```suggestion
    - sqlite3
    - libtiff
```

dopplershift: Thanks, that might fix libtiff, but I still need to figure out why the pre-built sqlite3 is failing when the build tries to use it.

ryanking13: > Thanks, that might fix libtiff, but I still need to figure out why the pre-built sqlite3 is failing when the build tries to use it.

https://stackoverflow.com/questions/14082152/sqlite3-no-such-file-or-directory

Hmm it seems like pre-built sqlite3 binary doesn't work for 64bit systems. I'm not sure why they don't provide x86-64 binaries. I think we will have to build sqlite3 for host os or add sqlite3 binary in our docker image as you mentioned before...

dopplershift: Including a `sqlite3` binary like you would a compiler (using a distro package) seems like the easiest path here...

ryanking13: Alright, let's add sqlite3.

dopplershift: Can I just add it to the `Dockerfile`? Or does it need a separate PR?

ryanking13: I'll open a PR. One of core devs need to trigger the Docker deploy workflow.

ryanking13:
I opened a PR adding sqlite3 in our docker image (#2807). You can try it with: `pyodide/pyodide-env:20220629-py310-chrome102-firefox100`.

dopplershift:
🤞

ryanking13:
It looks like now it wants `PROJ_DIR` env variable to be set?

dopplershift:
Has anyone tried making a `certifi` package for inclusion in the pyodide runtime?

packages/pyproj/patches/fix-build.patch
```diff
@@ -0,0 +1,12 @@
+diff --git a/setup.py b/setup.py
+index d8c1872a..ad396e02 100644
+--- a/setup.py
++++ b/setup.py
+@@ -167,7 +167,6 @@ def get_extension_modules():
+     ext_options = {
+         "include_dirs": include_dirs,
+         "library_dirs": library_dirs,
+-        "runtime_library_dirs": library_dirs if os.name != "nt" else None,
```
ryanking13, line 9: Perhaps we should update `pywasmcross.py` to ignore `-R` flags in `wasm-ld` (no action required for this PR) @hoodmane.

packages/certifi/meta.yaml
```diff
@@ -0,0 +1,17 @@
+package:
+  name: certifi
+  version: 2022.6.15
+
+source:
+  url: https://files.pythonhosted.org/packages/cc/85/319a8a684e8ac6d87a1193090e06b6bbb302717496380e225ee10487c888/certifi-2022.6.15.tar.gz
```
ryanking13, line 6: certifi has a pure python wheel. Could you replace the source from `.tar.gz` with `.whl`?

ryanking13 (commented):
Thanks, @dopplershift! It looks very cool and I hope you had some fun working on this. Now you'll need to:

1. Add some basic tests for pyproj
2. Update changelog
3. Remove sqlite tar file that I think you accidentally pushed
4. Wait until #2807 gets merged and merge main to this branch

packages/certifi/meta.yaml
dopplershift: Done.

packages/pyproj/patches/fix-build.patch
dopplershift: This would save me upstreaming the patch.

dopplershift:
> Thanks, @dopplershift! It looks very cool and I hope you had some fun working on this. Now you'll need to:
> 
> 1. Add some basic tests for pyproj

Can you clarify what it would look like? I checked h5py, but didn't see anything.

> 2. Update changelog

Done.

> 3. Remove sqlite tar file that I think you accidentally pushed

Done.

> 4. Wait until #2807 gets merged and merge main to this branch

Ok.

ryanking13:
> > 1. Add some basic tests for pyproj
> 
> Can you clarify what it would look like? I checked h5py, but didn't see anything.

You need to add `test_pyproj.py` inside pyproj directory.

See `pyyaml` example for a reference: https://github.com/pyodide/pyodide/blob/main/packages/pyyaml/test_pyyaml.py

dopplershift:
Added the test and rebased on main with #2807.

dopplershift:
Good to have the test because it caught:
```
/lib/python3.10/site-packages/pyproj/__init__.py:91: UserWarning: Valid PROJ data directory not found. Either set the path using the environmental variable PROJ_LIB or with `pyproj.datadir.set_data_dir`.
  warnings.warn(str(err))
CppException osgeo::proj::io::FactoryException: Cannot find proj.db
```
Where are the data files supposed to be in our environment? What's the best way to point to them? Environment variable? Edit the package files?

ryanking13:
> Apparently it can get included in the wheel for `pyproj` if you set `PROJ_WHEEL` and have a copy of PROJ in a specific internal directory. So I guess to be more clear, the file `proj.db` and a few other data files should be part of the PROJ C library package (that I added here as `libproj`).

I added a script that copies proj.db and data files into pyproj directory, following the setup.py of pyproj. 

https://github.com/pyproj4/pyproj/blob/54ad87816a239ad7f42dd626beb18341987b01f0/setup.py#L199-L214

Probably in the future it would be good to ask the pyproj maintainer to add a feature to modify the search path by using env variables.

packages/pyproj/meta.yaml
```diff
@@ -19,7 +19,9 @@ build:
     export PROJ_DIR=${WASM_LIBRARY_DIR}
     export PROJ_INCDIR=${WASM_LIBRARY_DIR}/include
     export PROJ_LIBDIR=${WASM_LIBRARY_DIR}/lib
-
+    export PROJ_WHEEL=1
+    mkdir -p pyproj/proj_dir
+    cp -r ${WASM_LIBRARY_DIR}/share pyproj/proj_dir
```
ocefpaf, line 24: I'm guessing this is the best option for now but that would duplicate all these files in projects that depend both on proj and pyproj, right?

ryanking13: Probably, yes. I don't know much about libproj, but it seems like packages that use libproj all embeds `pyproj.db` inside its wheel. For example, inside [fiona](https://pypi.org/project/Fiona/), I can see `pyproj.db`.

```
$ unzip -l Fiona-1.8.21-cp310-cp310-manylinux_2_17_x86_64.manylinux2014_x86_64.whl  | grep proj.db
  7921664  2022-02-07 15:46   fiona/proj_data/proj.db
```

Since `proj.db` is a big file in terms of the browser (~7.7MB), it would be great if we can share it between multiple packages. Maybe someone can ask package maintainers to do such optimization.

ryanking13 (approved):
I think we can merge this now, thanks @dopplershift for your hard work!

dopplershift:
🎉 

@ryanking13 So just to be clear, it gets bundled in wheels because they need to bundle all of libproj because wheels/PyPI don't have a concept of non-Python dependencies. On the Conda side, because it just handles generic packages, the libproj package has `proj.db` that everything else (like fiona and pyproj) can "just use".

It's unclear to me which approach packaging within pyodide is taking.

packages/pyproj/meta.yaml
ocefpaf: It is a design problem with the wheel. They a need to bundle proj and its data and we ended up with multiple copies of it. Package manager that have a more "ecosystem" logic can share data and c-lib dependencies in a more efficient way.

I'm not a pyodide expert but maybe we can use the same approach here as we do with conda?

ryanking13: I think we will keep using wheels for compatibility, but that sound very interesting, thanks.

PR #2560, opened 2022-05-14
Canonicalize requirement names in micropip

hoodmane:
Use canonicalize_name rather than our own custom version.
Use canonicalize_name also on requirements.

- [ ] Add / update tests

### DIFF
--- a/packages/micropip/src/micropip/_micropip.py
+++ b/packages/micropip/src/micropip/_micropip.py
@@ -12,12 +12,13 @@ from zipfile import ZipFile
 
 from packaging.markers import default_environment
 from packaging.requirements import Requirement
+from packaging.utils import canonicalize_name
 from packaging.version import Version
 
 from pyodide import IN_BROWSER, to_js
 
 from .externals.pip._internal.utils.wheel import pkg_resources_distribution_for_wheel
-from .package import PackageDict, PackageMetadata, normalize_package_name
+from .package import PackageDict, PackageMetadata
 
 # Provide stubs for testing in native python
 if IN_BROWSER:
@@ -222,10 +223,7 @@ class _PackageManager:
                     ),
                     functools.partial(
                         self.installed_packages.update,
-                        {
-                            normalize_package_name(pkg.name): pkg
-                            for pkg in pyodide_packages
-                        },
+                        {canonicalize_name(pkg.name): pkg for pkg in pyodide_packages},
                     ),
                 )
             )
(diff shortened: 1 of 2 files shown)


rth (approved):
Yes, makes sense, thanks!

ryanking13:
Thanks, @hoodmane. I read the original bug report in the Gitter, and I can confirm that there is a bug in processing the package name.

However, I don't think this change is what we intended. What `normalize_package_name` was doing is to normalize package names to wheel filename format following [PEP 427](https://peps.python.org/pep-0427/#escaping-and-unicode). For example, it converts `keras-Preprocessing` to `keras_preprocessing`. On the other hand, `canonicalize_name` does the opposite: converts underscore to dash (`keras_preprocessing` to `keras-preprossing`).

I think the correct place to be fixed is here:

https://github.com/pyodide/pyodide/blob/663459e26b71c7bbc6808b955c91398009de6803/packages/micropip/src/micropip/_micropip.py#L359-L363

These lines were meant to retrieve the correct project name from the wheel metadata file. It is needed when micropip installs a package from a wheel file directly. But I think it is not working as intended somehow.

hoodmane:
Well I guess we need better test coverage. But pip itself uses `canonicalize_name` so if we're going based on what pip does it would make sense. Does it matter which way we go as long as we're consistent?

rth:
OK, I should have waited before merging. 

Yes, though as long as the projection allows to disambiguate common variations in package names, it shouldn't matter if `Keras-Preprocessing` is converted to `keras_preprocessing` or `keras-preprocessing`, should it? It's an internal key anyway.

ryanking13:
I read how PyPI normalizes package names in [PEP 503](https://peps.python.org/pep-0503/) and I think Hood was right. Using caconicalized name seems more suitable for this case.

> It's an internal key anyway.

We still need to correct way to retrieve and save "original" name of the package from the wheel file for `micropip.list()` though, since we don't want to show canonicalized name to users.

hoodmane:
Do we show them the name they asked for? What if it showed up because of a transitive requirement? Showing the canonicalized name in micropip.list doesn't seem terrible to me but I guess it's not ideal if the user specifically asked for a different name.

ryanking13:
> Showing the canonicalized name in micropip.list doesn't seem terrible to me but I guess it's not ideal if the user specifically asked for a different name.

Yes, it's not terrible but it can be improved.

```py
>>> import micropip
>>> await micropip.install("weeby.py")
>>> micropip.list()
Name               | Version   | Source
------------------ | --------- | -------
...
weeby-py           | 0.1.6     | pypi
...
>>> "weeby.py" in micropip.list()
True
>>> "weeby-py" in micropip.list()
True
>>> "weeby_py" in micropip.list()
True
>>> "WEEBY_PY" in micropip.list()
True
```

hoodmane:
I think the one with `weeby.py` should be false all others true. This would be consistent with the behavior of importlib.metadata.

PR #2561, opened 2022-05-14
Better typechecking for micropip

hoodmane:
#### Better typechecking for micropip

I turned wheel and transaction into dataclasses so we can typecheck
their fields. I also went through and ensured that all functions have
typed arguments.

### DIFF
--- a/packages/micropip/test_micropip.py
+++ b/packages/micropip/test_micropip.py
@@ -57,7 +57,6 @@ def mock_get_pypi_json(pkg_map):
                 }
             ]
 
-        print("pkgname", pkgname, json_data)
         return json_data
 
     return _mock_get_pypi_json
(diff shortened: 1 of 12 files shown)


packages/micropip/test_micropip.py
```diff
@@ -50,13 +50,14 @@ async def _mock_get_pypi_json(pkgname, **kwargs):
...
                     "filename": file,
-                    "url": "",
+                    "url": file,
                     "digests": {
                         "sha256": Wildcard(),
                     },
                 }
             ]
 
+        print("pkgname", pkgname, json_data)
```
ryanking13, line 60: If this is added for debugging, please remove this before merging.

ryanking13 (approved):
It looks great to me. Thanks, @hoomane!

packages/micropip/test_micropip.py
hoodmane: Oops, thanks for catching it.

hoodmane:

rth:
Yes, thanks a lot @ryanking13 ! I'm happy that the project became a bit more sustainable with more active maintainers!

PR #2562, opened 2022-05-16
Improve micropip error message when no package is found

ryanking13:
### Description

Show better error message when fetching metadata from PyPI fails. Previously it just showed "Failed to Fetch"


rth (approved):
Thanks! Do you want to add a changelog entry?

PR #2563, opened 2022-05-16
Fix ordering in micropip add_requirement

hoodmane:
I think the logic here is slightly wrong. Steps for `add_requirement`:

1. Convert `req` from a string to a Requirement if necessary
2. If `req.marker` is present, check whether `req.marker` matches `ctx`, if not we don't need it.
3. Check if `req` is already installed
4. Check if `req` is available from `packages.json`
5. Check if `req` is available from pypi

For some reason we had step 4 occurring before steps 2 and 3. This puts them back in the correct order.


ryanking13 (approved):
Sounds reasonable, thanks @hoodmane!

PR #2569, opened 2022-05-17
micropip Fix async race condition

hoodmane:
It's possible that while we were awaiting on `_get_pypi_json`, some other branch
of the resolver already installed a different version of the same wheel. We need
to check a second time for whether the constraint is satisfied between when we
get the `pypi` metadata and when we put the information into `transaction.locked`
or else we might accidentally install the same wheel twice.

### DIFF
--- a/packages/micropip/src/micropip/_micropip.py
+++ b/packages/micropip/src/micropip/_micropip.py
@@ -222,6 +222,22 @@ class Transaction:
 
         await self.add_wheel(wheel, extras=set())
 
+    def check_version_satisfied(self, req: Requirement):
+        ver = None
+        if req.name in self.locked:
+            ver = self.locked[req.name].version
+
+        if not ver:
+            return False
+
+        if req.specifier.contains(ver, prereleases=True):
+            # installed version matches, nothing to do
+            return True
+
+        raise ValueError(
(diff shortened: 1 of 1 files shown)


hoodmane:
In the long term I guess a better solution would be to somehow lock around `_get_pypi_json`. If two branches decide to install the same package, they will both try to make the same request and we can easily capture that and handle it somehow. One could even intersect the constraints before going on to `find_wheel`.

There are also issues with `micropip.install` not having a lock around the base. It also doesn't care about `pyodide.loadPackage`'s lock, though this only matters if we install a wheel from PyPI that was also available via `pyodide.loadPackage`. In this case, we can get crashy packages where we overwrote some of the files from one version of the package with a different version.

I already have a branch on top of this one that gets rid of `INSTALLED_PACKAGES` in `micropip` in favor of using `importlib.metadata`. I think we should also stop using `loadedPackages` to check if a package is loaded in `load-package` in favor of also using `importlib.metadata` so that there will be a single source of truth which is shared between the Python stdlib, `micropip`, and `loadPackage`.

hoodmane:
I guess it's not possible to rely 100% on wheel metadata because some things loaded by `loadPackages` are not wheels (shared libraries, stdlib modules, and tests). But these won't be available from pypi or other sources so it's okay for `loadPackages` to track them in its own private data structure.

PR #2572, opened 2022-05-17
ENH Use dist-info + `importlib.metadata` to store and retrieve package metadata

hoodmane:
This switches to using the file system to store the information about packages
and using `importlib.metadata` to retrieve it.
This has two related benefits:

1. We don't have to separately maintain our own separate state about what
we've installed.

2. We are guaranteed to agree with the Python stdlib about whether or not a
package is installed, and if so which version is installed. Since the state is 
maintained in only one place, there is no chance of it diverging.

According to the packaging specs, if the package is from a url we should put 
that into a file here:
https://packaging.python.org/en/latest/specifications/direct-url/#direct-url
Other than that, we should set the `INSTALLER` to `pyodide.loadPackage` 
if the package was loaded with `pyodide.loadPackage` either directly or 
indirectly via `micropip`. Otherwise set `INSTALLER` to `micropip`. That
way we can maintain the current behavior of `micropip.list`:
1. if `direct_url.json` is present, then `source` is the url in `direct_url.json`
2. otherwise, set `source` to `pyodide` if `INSTALLER` is `pyodide.loadPackage`
and set it to `micropip` if `INSTALLER` is `micropip`.

Oddly enough, the packaging specs suggest no place to put the source
index, so it seems that according to the specs if a package was installed
by name from some custom index, this info should not be recorded.

### Checklists

- [x] Add / update tests

### DIFF
--- a/packages/micropip/src/micropip/_micropip.py
+++ b/packages/micropip/src/micropip/_micropip.py
@@ -3,6 +3,7 @@ import hashlib
 import importlib
 import json
 from dataclasses import dataclass, field
+from importlib.metadata import PackageNotFoundError
 from importlib.metadata import distributions as importlib_distributions
 from importlib.metadata import version as importlib_version
 from io import BytesIO
@@ -235,7 +236,7 @@ class Transaction:
         ver = None
         try:
             ver = importlib_version(req.name)
-        except ModuleNotFoundError:
+        except PackageNotFoundError:
             pass
         if req.name in self.locked:
             ver = self.locked[req.name].version
--- a/packages/micropip/src/micropip/_micropip.py
+++ b/packages/micropip/src/micropip/_micropip.py
@@ -17,7 +17,7 @@ from packaging.utils import canonicalize_name
 from packaging.version import Version
 
 from pyodide import to_js
-from pyodide._package_loader import set_wheel_installer
+from pyodide._package_loader import parse_wheel_name, set_wheel_installer
 
 from ._compat import (
     BUILTIN_PACKAGES,
@@ -72,12 +72,7 @@ class WheelInfo:
         file_name = Path(url).name
         # also strip '.whl' extension.
         wheel_name = Path(url).stem
-        tokens = wheel_name.split("-")
-        # TODO: support optional build tags in the filename (cf PEP 427)
-        if len(tokens) < 5:
-            raise ValueError(f"{file_name} is not a valid wheel file name.")
-        version, python_tag, abi_tag, platform = tokens[-4:]
-        name = "-".join(tokens[:-4])
+        name, version, python_tag, abi_tag, platform = parse_wheel_name(wheel_name)
         return WheelInfo(
             name=name,
             version=Version(version),
--- a/packages/micropip/src/micropip/_micropip.py
+++ b/packages/micropip/src/micropip/_micropip.py
@@ -141,7 +141,9 @@ class WheelInfo:
     def set_installer(self):
         assert self.data
         wheel_source = "pypi" if self.digests is not None else self.url
-        set_wheel_installer(self.data, WHEEL_BASE, "micropip", wheel_source)
+        set_wheel_installer(
+            self.filename, self.data, WHEEL_BASE, "micropip", wheel_source
+        )
 
     async def install(self):
         url = self.url
--- a/packages/micropip/src/micropip/_micropip.py
+++ b/packages/micropip/src/micropip/_micropip.py
@@ -16,6 +16,7 @@ from packaging.utils import canonicalize_name
 from packaging.version import Version
 
 from pyodide import to_js
+from pyodide._package_loader import set_installer
 
 from ._compat import (
     BUILTIN_PACKAGES,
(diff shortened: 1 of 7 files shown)


hoodmane:
Opened a question on discuss.python.org:
https://discuss.python.org/t/pypa-spec-installers-record-package-index-source-into-dist-info-like-direct-url-json/15817

hoodmane:
Okay apparently according to the pypa spec installers are allowed to do whatever they want as long as they include the METADATA file in the `dist-info` directory:
https://packaging.python.org/en/latest/specifications/recording-installed-packages/#the-dist-info-directory

> The METADATA file is mandatory. All other files may be omitted at the installing tool’s discretion. Additional installer-specific files may be present.

So we are fine to include an extra `PYODIDE_SOURCE` file if we want.

ryanking13 (approved):
Thanks a lot @hoodmane! This would significantly improve our packaging system.

packages/micropip/src/micropip/_micropip.py
```diff
@@ -483,17 +462,29 @@ def _list():
         >>> "regex" in package_list # doctest: +SKIP
         True
     """
-    packages = copy.deepcopy(INSTALLED_PACKAGES)
 
     # Add packages that are loaded through pyodide.loadPackage
+    packages = PackageDict()
+    for dist in importlib_distributions():
+        name = dist.name
```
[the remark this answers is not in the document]
hoodmane: I think this already works on the main branch?

src/py/pyodide/_package_loader.py
```diff
@@ -149,6 +155,25 @@ def should_load_dynlib(path: str):
...
+    archive: IO[bytes],
+    target_dir: Path,
+    installer: str | None,
+    source: str | None,
+):
+    z = ZipFile(archive)
+    name, version = filename.split("-", 2)[:-1]
+    dist_info_name = f"{name}-{version}.dist-info"
+    if not (zipfile.Path(z) / dist_info_name).is_dir():
+        raise Exception("archive is not a valid wheel")
```
rth, line 169: Maybe it would be good to print the name of the archive here, otherwise, it might be hard to figure out from the traceback which package failed

src/py/pyodide/_package_loader.py
```diff
@@ -149,6 +155,25 @@ def should_load_dynlib(path: str):
...
+    dist_info = target_dir / dist_info_name
+    if installer:
+        (dist_info / "INSTALLER").write_text(installer)
```
rth, line 172: Could you add an inline comment linking to https://packaging.python.org/en/latest/specifications/recording-installed-packages/#the-dist-info-directory to indicate that this is not an arbitrary name.

src/py/pyodide/_package_loader.py
```diff
@@ -149,6 +155,25 @@ def should_load_dynlib(path: str):
     return not PLATFORM_TAG_REGEX.match(tag)
 
 
+def set_wheel_installer(
+    filename: str,
+    archive: IO[bytes],
+    target_dir: Path,
+    installer: str | None,
+    source: str | None,
+):
```
rth, line 164: Could you please add a docstring explaining the arguments.

src/py/pyodide/_package_loader.py
```diff
@@ -149,6 +155,25 @@ def should_load_dynlib(path: str):
...
+    z = ZipFile(archive)
+    name, version = filename.split("-", 2)[:-1]
+    dist_info_name = f"{name}-{version}.dist-info"
+    if not (zipfile.Path(z) / dist_info_name).is_dir():
+        raise Exception("archive is not a valid wheel")
+    dist_info = target_dir / dist_info_name
+    if installer:
+        (dist_info / "INSTALLER").write_text(installer)
+    if source:
+        (dist_info / "PYODIDE_SOURCE").write_text(source)
```
rth, line 174: I can't decide if we should be using [`direct_url.json`](https://packaging.python.org/en/latest/specifications/direct-url/#direct-url) for this if it's a separate use case.

packages/micropip/src/micropip/_micropip.py
```diff
@@ -483,17 +468,29 @@ def _list():
...
+        version = dist.version
+        source = dist.read_text("PYODIDE_SOURCE")
+        if not source:
```
rth, line 478:
```suggestion
        if source is None:
```
Maybe worth an inline comment that `source` is None if `PYODIDE_SOURCE` does not exis . It should never be empty.

rth (commented):
Great, thanks for doing this! 

src/py/pyodide/_package_loader.py
hoodmane: Ideally I think we should add `direct_url.json` according to the spec. This won't be sufficient for our use case, because if the package was NOT a `direct_url` then we want to know what index we downloaded it from. But we can do both.

rth: Well we could probably add custom fields if needed to that json

hoodmane: There's also the `REQUESTED` file. I think in a subsequent PR we could add `direct_url` and `REQUESTED`.
It's kind of weird that the spec has nothing about recording which index the package came from.

hoodmane: > Well we could probably add custom fields if needed to that json

We can add custom fields but according to the spec:

> This file MUST NOT be created when installing a distribution from an other type of requirement (i.e. name plus version specifier).

So that rules out using it to specify which index a package came from.

PR #2576, opened 2022-05-18
MAINT Remove outdated commands in Makefile

ryanking13:
- `PYODIDE_EMCC`, `PYODIDE_CXX`: removed in #1218
- `PYODIDE_BASE_URL` in webworker.js: removed in #2292

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -137,17 +137,14 @@ docs/_build/html/console.html: src/templates/console.html
 .PHONY: dist/webworker.js
 dist/webworker.js: src/templates/webworker.js
 	cp $< $@
-	sed -i -e 's#{{ PYODIDE_BASE_URL }}#$(PYODIDE_BASE_URL)#g' $@
 
 .PHONY: dist/module_webworker_dev.js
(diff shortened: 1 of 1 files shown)


hoodmane:
Thanks @ryanking13! What about the `update_base_url` recipe and the ci steps in the deploy jobs?

hoodmane:
Ah I guess `console.html` is the one place left with `PYODIDE_BASE_URL`.  So `update_base_url` can be changed to rebuild just `dist/console.html`

rth:
> PYODIDE_BASE_URL in webworker.js: removed

That's very good news! So we could actually move `webworker.js` from the templates and into the js package. I wanted to do that some time ago, but was blocked by the necessity to use PYODIDE_BASE_URL

ryanking13:
Thanks for the reviews!

> That's very good news! So we could actually move `webworker.js` from the templates and into the js package. I wanted to do that some time ago, but was blocked by the necessity to use PYODIDE_BASE_URL

I'm not sure of that. I think it is used for test-only? We can probably discuss this in a separate issue.

Issue #989, opened 2020-12-31
Run tests in a WebKit based browser (or Safari) in CI


rth:
It's would a bit difficult to run tests in Safari, but I think having tests with some other WebKit based browser available on Linux (e.g. Midori) would already be quite useful. 

ref: https://github.com/iodide-project/pyodide/issues/721

hoodmane:
Do we have to support Safari? It seems like an outlier on caniuse, there's a whole pile of features that were implemented in Chromium and Firefox several years ago that Safari still doesn't support. Pyodide intrinsically relies on a bunch of quite recent features.

hoodmane:
For instance, in order to get keyboard interrupts to work in any form we will need `SharedArrayBuffer`, but Safari doesn't support it:
https://caniuse.com/sharedarraybuffer

hoodmane:
Another example is WebGL2:
https://caniuse.com/?search=webgl2
I think it's fairly typical for Safari to be 3 years behind Mozilla / Chrome in feature support.

oeway:
Also see here: https://webassembly.org/roadmap/ too many red crosses for Safari ;)

rth:
> Do we have to support Safari? 

Well we don't necessarily have to, but the thing is that there is a number of MacOS and iOS users that keep reporting issue when using pyodide https://github.com/iodide-project/pyodide/issues/533#issuecomment-736928032. So if we could make it work (even with some features disabled) and without too much efforts on our side, it would be worth considering.

I opened this issue mostly to experiment, and see what actually breaks when we run the test suite.

joemarshall:
I want this too - for remote teaching, as we still have some students with Macs. 

I was going to build an alternative conftest.py that routed things through playwright instead of selenium. In theory shouldn't be too hard, and playwright means recent WebKit as well as Firefox and Chrome. I think selenium can only run WebKit easily on Mac and I don't have access to any during lockdown. I will have a play and see how hard it is maybe.

Incidentally, a lot of WebKit features got fixed on webkit 14, and Mac users surely should have recent Safari, so I think the basis for running in WebKit might well be there now, at least better than on that link above.

rth:
> I was going to build an alternative conftest.py that routed things through playwright instead of selenium.

[...]

PR #2578, opened 2022-05-18
Run core tests in Safari

rth:
Partially addresses https://github.com/pyodide/pyodide/issues/989

It runs, but there are several test failures, I'll open separate issues about those.


rth:
One blocker for this is that Safari apparently only supports 1 session at a time https://github.com/SeleniumHQ/selenium/issues/5057, (from https://developer.apple.com/documentation/webkit/about_webdriver_for_safari)
> One Session at a Time, to Mimic User Interaction
>
> Only one Safari browser instance can be active at any given time, and only one WebDriver session at a time can be attached to the browser instance. These constraints ensure that the simulated behavior (mouse, keyboard, touch, and so forth) accurately reflects what a user can do in a macOS windowing environment and prevents tests from competing with each other for window and keyboard focus.

so when we initialize the selenium module fixtures and then use a `selenium_standalone` fixture somewhere, we end up with 2 sessions and it fails with,
```
selenium.common.exceptions.SessionNotCreatedException: Message: Could not create a session: 
     The Safari instance is already paired with another WebDriver session.
```
I'm not sure what we can do about that. I suppose we could keep a global list of sessions, and close previous sessions when a new session is opened, but I don't think it would easily work with pytest fixture caching. 

Maybe @ryanking13 would have some ideas. I guess with playwright it would be the same issue.

ryanking13:
> I'm not sure what we can do about that. I suppose we could keep a global list of sessions,

Maybe we can adjust the test running order (I'm not sure pytest has that feature) so that tests that use `selenium_standalone` can be executed in advance?

One other option can be instantiating multiple Pyodide instances inside a single selenium session.
I've once tried that in #2406 but it was not very successful then...

> Maybe @ryanking13 would have some ideas. I guess with playwright it would be the same issue.

Playwright does not have that issue. Probably because it is not based on web drivers.

Speaking of playwright, I also did some experiments [on safari (webkit) tests with playwright in May](https://github.com/ryanking13/pyodide/tree/playwright-webkit). It is blocked because playwright sometimes hangs and I wasn't able to find a way to fix it (I believe that it is partially related to #2660).

rth:
> Maybe we can adjust the test running order (I'm not sure pytest has that feature) so that tests that use selenium_standalone can be executed in advance?

Thanks, @ryanking13 , very good point and that's actually rather straightforward to do. Though are are still a bunch of hard to debug issues with,
```
../miniforge3/envs/pyodide/lib/python3.10/site-packages/selenium/webdriver/remote/errorhandler.py:247: in check_response
    raise exception_class(message, screen, stacktrace)
E   selenium.common.exceptions.InvalidSessionIdException: Message:
```

rth:
OK, so I got it down to 9 failing core tests but it still needs some work.

ryanking13:
I synced with upstream main / pytest-pyodide and fixed some failing tests.

rth:
Aww, great so you manage to make tests pass! You have more persistance than me :) 

Code wise LGTM, so feel free to merge if you are OK with it. I'm just going to move the changelog entry to the next release.

ryanking13:
> Aww, great so you managed to make tests pass! You have more persistence than me :)

I only did the last 10% and you did the rest 90% of jobs :)

PR #2581, opened 2022-05-19
ENH Add `micropip.freeze`

hoodmane:
This adds a function `micropip.freeze()` which creates a `packages.json`
lock file which includes all packages that were loaded by micropip in the
current Pyodide session. This can be used in subsequent sessions to load
the same set of packages with `pyodide.loadPackage`.

For example in our repl:

```py
from js.console import log
import micropip
await micropip.install("sphinx-version-warning") # Installs 19 wheels from pypi
log(micropip.freeze())
```
Then opening the browser console, we can copy the JSON and make a new `packages.json`
file out of it. (Our repl will just say "<Long output truncated>" but the browser
console truncates it and provides tools to make it easy to copy.) Reloading the page
with the new `packages.json`, `versionwarning` will autoload:
```py
import versionwarning # Automatically loads 19 wheels from PyPI
```
We detected that `versionwarning` is an export provided by the `sphinx-version-warning`
package via the `top_level.txt` file. Distressingly this file is part of the `egg-info`
spec but isn't mentioned anywhere in the `dist-info` spec. But wheels seem to include
it nonetheless.
### Checklists

- [x] Add / update tests


packages/micropip/src/micropip/_micropip.py
```diff
@@ -445,6 +468,54 @@ async def install(
...
+    for dist in importlib_distributions():
+        name = dist.name
+        version = dist.version
+        url = dist.read_text("PYODIDE_URL")
+        if url is None:
+            continue
+
+        sha256 = dist.read_text("PYODIDE_SHA256")
+        assert sha256
+        imports = (dist.read_text("top_level.txt") or "").split()
```
ryanking13, line 492: Probably we should discuss the expected usage of `imports` key in packages.json later. It `meta.yaml` spec, we explained that [it is for testing](https://pyodide.org/en/stable/development/meta-yaml.html#test-imports) and we actually use it in [our test](https://github.com/pyodide/pyodide/blob/60b9344b5622958e9ce7b173b1ec48816a8df424/packages/test_packages_common.py#L72). But it is also used in [loadPackagesFromImports](https://github.com/pyodide/pyodide/blob/60b9344b5622958e9ce7b173b1ec48816a8df424/src/js/api.ts#L127), so it is not only for test, which is a little confusing. (no action required for this PR) (c.c @rth)

ryanking13 (commented):
Thanks @hoodmane! Please update the changelog and add some tests.

packages/micropip/src/micropip/_micropip.py
hoodmane: Maybe we should switch to using `top_level.txt` for `loadPackagesFromImports`.

PR #2583, opened 2022-05-19
refactor: use .path & type build

henryiii:
### Description

Replace `._path` with `.path`, and include build in static typing.

I'm seeing if we can add `.path` to the visible types here: https://github.com/pypa/build/pull/475


hoodmane (approved):
Much appreciated!

rth:
Thanks!

PR #2591, opened 2022-05-21
ENH Add support in micropip for installing binary wheels

hoodmane:
This PR is on top of #2589.

We use `packaging.tags.sys_tags` to get the list of supported tags
then use `packaging.utils.parse_wheel_filename` to get the set of
tags the current wheel implement then check if one of the wheel's
tags is a supported tag. This is a fully accurate check method and
could also catch things like abi3 wheels that are compatible with
multiple Python versions.

This allows micropip to be used to install binary emscripten wheels.
In order to manage the ABI issues, I also changed the wheel platform
tag to `emscripten_2_0_27_wasm32` so we can tell which version of
emscripten the wheel was built with.

The normal way that the platform tag is generated is by substituting
in the `uname` OS name and version info. Emscripten has dummy
information in `uname` instead of any actual version, so I patched
`uname` to report the Emscripten version. See upstream issue:
https://github.com/emscripten-core/emscripten/issues/16977

### Checklists

- [x] Add / update tests


rth:
Thanks for working on this!

> This allows micropip to be used to install binary emscripten wheels.
In order to manage the ABI issues, I also changed the wheel platform
tag to emscripten_2_0_27_wasm32 so we can tell which version of
emscripten the wheel was built with.


As far as I understand in [PEP 425](https://peps.python.org/pep-0425/) the filename is 
```
{distribution}-{version}(-{build tag})?-{python tag}-{abi tag}-{platform tag}.whl
```
so the platform tag would be `emscripten_wasm32` and then we could add an ABI tag indeed. I'm not sure what's best for binary tag, PEP 425 suggests going with the short git checksum for unstable ABIs. Using the emscripten version sounds better indeed. Maybe use `200027` as the abi tag (i.e. zero pad the minor version to 2 and patch version to 3 digits) though I imagine it's less reliable?  

Would you mind opening an issue about it at https://discuss.python.org/ so we don't discover it would have been better to use a different wheel file name format after the release?

rth:
Though the CPython version already appears in the ABI tag. So we probably should include both the Python version and the emscripten version there. So instead of,
```
numpy-1.22.3-cp310-cp310-emscripten_2_0_27_wasm32.whl
```
we could have
```
numpy-1.22.3-cp310-cp310_em200027-emscripten_wasm32.whl
```
In any case worth talking with CPython people about this.

rth:
But then looking at,
```
numpy-1.22.4-cp310-cp310-macosx_11_0_arm64.whl 
numpy-1.22.4-cp310-cp310-manylinux_2_17_aarch64.manylinux2014_aarch64.whl
```
maybe indeed you are right, and including the emscription version in the platform tag is the way to go. In that case, though, I don't really see the point of the ABI tag as it's almost always equal to the python tag.

hoodmane:
@ryanking13 any guesses why this would crash firefox in gha but work fine in circleci?

ryanking13:
> @ryanking13 any guesses why this would crash firefox in gha but work fine in circleci?

Maybe because the Firefox version installed in GHA environment is old? I didn't read other PR's split off from this yet.

hoodmane:
It's getting to the point that there isn't much left outside of `packages/micropip` I haven't split off. But surely changes to micropip shouldn't crash tests that don't load micropip?

rth:
Thanks for doing this! Though I think it would be better to return to a workflow where user-facing code is reviewed before merging. I know we have less review bandwidth at the moment, and you are opening a lot of MR. I agree that some of those related to maintenance changes can indeed be merged without review, but there has been a tendency lately to merge more things without review and I'm not sure it's good. For this particular PR everything LGTM. Don't hesitate to ping, if a PR remains without review for too long.

hoodmane:
> I think it would be better to return to a workflow where user-facing code is reviewed before merging. 

Right, agreed. I am a bit dissatisfied with github's PRs. If I have tasks A, B, C that depend on each other but should be reviewed separately, the options I am aware of are:

1. Do task A, open a PR, wait ~10 days or so for it to be reviewed and merged. A few days pass and then I remember that task B is still a thing I meant to do, so then I do it, wait for it, etc. This also destroys context for reviewers because they can't see the PR for B when A is available.
2. Combine task A and B into one bigger PR. This is more annoying to review and produces a worse commit log. But it's easier to work on.
3. Just merge PR A without review. This has major drawbacks in the resulting code quality and in how much other people know about what is going on. It does generate small commits and a helpful commit log.
4. Make a separate branch for B and create a PR on top of A. The diff for PR B then includes all of the stuff from PR A which really annoys me. When PR A gets reviewed and gets commits, I can either merge them into branch B or rebase branch B.
4a. Merge changes to A into B. Merging works well and easily until PR A is squash merged. Then I have to merge main into B and I get a huge number of merge conflicts because this history confuses git quite badly. I can resolve all of them by accepting B's version of events because B already includes A. It doesn't involve much brain power but it is time consuming.
4b. Rebase B onto A and force push. This works quite well but it requires that you plan ahead: when you create branch B, you have to do ```sh
git checkout branch-a
git checkout -b branch-b-base
git checkout -b branch-b
```
then whenever you want to integrate new changes from A into B:
```sh
git checkout branch-b
git rebase branch-b-base --onto branch-b
git branch -f branch-b-base branch-a
git push -f
```
when A is merged:
```sh
git checkout main
git pull upstream main
git checkout branch-b
git rebase branch-b-base --onto main
git push -f
git branch -D branch-b-base
```
Problem is if you screw this up it can take a large amount of mental power to figure out what happened with your branches. Carefully inspecting the reflog usually helps. Basically, the merge technique is hard to screw up but requires large amounts of menial labor, the rebase technique is efficient if you get it right but it involves a lot of steps and it gets extremely confusing if you mess them up.

Option 4 should be the obvious right answer but github is just really really bad at it. I think when the rebase is clean, github should be able to automate 4. If we merge merged rather than squash merging, 4a would work without all the pain. Then the problem is that github PRs get screwed up by the force pushes when you rebase the branch.

hoodmane:
Of course the most reasonable answer to all of that is to take a break.

rth:
> I am a bit dissatisfied with github's PRs.

I know, it's not perfect. For me having a large PR and then splitting things out, merging those smaller PRs, then updating the main PR doesn't work too badly.

>   the most reasonable answer to all of that is to take a break.

Yeah, on this topic, I keep occasionally remembering [Naomi Ceder's PyCon keynote](https://www.youtube.com/watch?v=sFmwGQu0cQU) where she mentions that open-source maintenance is a long term effort and that maintainers need to take care not to get overloaded with work; it is a good talk.  

Otherwise, if you see things that can be improved in our review workflow don't hesitate to make suggestions :)

Issue #2466, opened 2022-05-03
Performance visualization page


rth:
@wlach mentioned on gitter,
> I'm looking at the benchmarking results for pandas right now. I'm curious, has there been any thought about collecting these results on the main branch into a single data store? I'm thinking of a simple github pages site as an MVP

I was also thinking that having an HTML page which displays historical data on performance and code sizes would be great.

It could be as simple as,
 - an HTML page at https://github.com/pyodide/pyodide-blog which does these plots using pyodide, and data is loaded from some json that we update say for each release
 - or indeed a specific application, that maybe monitors performance for each commit. Though then it's a bit more complex, and maybe it would make more sense to re-use https://asv.readthedocs.io/en/stable/using.html

Note that we are also planning to completely change the benchmark suite, to get closer to the used by CPython and or scientific python projects (e.g. pandas, numpy, and scikit-learn) already have test suites, and we could take a few representative cases from there.

wlach:
asv looks fantastic! It looks like it has setup methods which could be used to initialize selenium or set up tests: 

https://asv.readthedocs.io/en/stable/writing_benchmarks.html#writing-benchmarks

mdboom:
Yeah, `asv` could be useful here, but I'm biased.

For some inspiration (though I don't know whether it makes sense to use directly), see speed.python.org, which is based on [codespeed](https://github.com/tobami/codespeed).  These haven't seen much maintenance in a while, but may get revived.

wlach:
One thing asv seems pretty hardcore about specifying is the machine on which the benchmarks are run, which brings up an issue with any benchmarking *we* might want to do: performance on a VPS host is notoriously variable (even taking aside the fact that they might randomly upgrade the hardware on us). Although it looks like the [scipy people had some success with using GitHub actions to benchmark their stuff](https://labs.quansight.org/blog/2021/08/github-actions-benchmarks/), so maybe it's possible.

rth:
>  the CPython performance team is still using the "machine sitting under someone's desk" model.

Yeah, AFAIK https://pandas.pydata.org/speed/ which uses ASV also had that setup in the past. I have a desktop under my desk that could be used for such purposes, but unfortunately my cat sometimes walks on the power button, and then it switches off. So that's not ideal ^^ 

> Maybe bare metal cloud options could work, but that's expensive.

[...]

PR #2592, opened 2022-05-22
FIX: compile all .so files concurrently rather than sequentially

hoodmane:
@wlach @rth I haven't profiled it but there is a chance this could be a significant performance improvement for loading wheels with many .so files. Because of poor coding, we have been loading so files one at a time.

### DIFF
--- a/src/js/load-package.ts
+++ b/src/js/load-package.ts
@@ -226,7 +226,9 @@ async function installPackage(
     installer: "pyodide.loadPackage",
     source: channel === DEFAULT_CHANNEL ? "pyodide" : channel,
   });
-  await Promise.all(dynlibs.map((dynlib : string) => loadDynlib(dynlib, pkg.shared_library)))
+  await Promise.all(
+    dynlibs.map((dynlib: string) => loadDynlib(dynlib, pkg.shared_library))
+  );
   loadedPackages[name] = pkg;
 }
 
--- a/src/js/load-package.ts
+++ b/src/js/load-package.ts
@@ -226,9 +226,7 @@ async function installPackage(
     installer: "pyodide.loadPackage",
     source: channel === DEFAULT_CHANNEL ? "pyodide" : channel,
   });
-  for (const dynlib of dynlibs) {
-    await loadDynlib(dynlib, pkg.shared_library);
-  }
+  await Promise.all(dynlibs.map((dynlib : string) => loadDynlib(dynlib, pkg.shared_library)))
   loadedPackages[name] = pkg;
 }
 


wlach:
Hmm, interesting! Based on CI, it seems rather inconclusive whether this improves things -- the "load pandas" benchmarks improved but not outside what I'm eyeballing as the normal standard deviation for these tests. This kind of speaks to the need for #2466.

On branch:
 -  selenium init
    native: nan  firefox: 3.835918  chrome: 6.050650  
 -  load numpy
    native: nan  firefox: 0.984476  chrome: 4.926071  
 -  load pandas
    native: nan  firefox: 5.201257  chrome: 10.597220  
 -  load matplotlib
    native: nan  firefox: 2.436375  chrome: 7.550147  

On main:

 -  selenium init
    native: nan  firefox: 3.347592  chrome: 5.768081  
 -  load numpy
    native: nan  firefox: 0.850075  chrome: 4.685861  
 -  load pandas
    native: nan  firefox: 5.456115  chrome: 10.208408  
 -  load matplotlib
    native: nan  firefox: 2.188029  chrome: 7.521396

hoodmane:
Yeah that looks like no difference. Maybe the browser engine itself isn't willing to concurrently compile things, or maybe wasm compilation isn't as big of a bottleneck as we think? If we think `loadDynlib` is the bottleneck, we should make a version of `loadDynlib` that prints (so file, file size, start time, end time, difference) for each file we compile.

rth:

As far as I know,
 - WASM compilation are already running multi-threaded
 - our package installation (including the `loadDynlib` call) are already called concurrently with a `Promise.all` [here](https://github.com/pyodide/pyodide/blob/f2d80e084332028491829e606b77f6226c61d543/src/js/load-package.ts#L431)
 
so I wonder if there is much to be gained by another concurrency level. As far as I understand, overall we are still running this in a single thread (putting aside WASM compilation threads) and the task sounds mostly CPU bound, so I'm not sure that context concurrency helps (or in any case it wouldn't be a massive improvement)

What I have observed though is that if you take say pandas, measure the load time, then bundle it into a single archive with a smaller subset of .so used (what I have been doing in https://github.com/rth/pyodide-pack example)  the load time is going to be similar or larger than for the original application. My guess so far is that it's due to package installation (and therefore loadDynlib) being done concurrently in the first case but not the second (though I have no benchmarks to corroborate it for now).

hoodmane:
> overall we are still running this in a single thread

I believe that this is not true: I think `WebAssembly.instantiate` and the other asynchronous apis in the `WebAssembly` namespace run in a different thread than the main runtime. This is to allow it to happen in the background without blocking the main execution thread. However, whether a single JavaScript thread has access to multiple wasm compilation threads is not clear, especially since this PR yields no improvement.

This change at least gives the possibility of browsers compiling more than one .so file at the same time. It shouldn't help much if we are primarily bound by the download time or if the browser doesn't give us more than one wasm compilation background thread at once. Without this patch we request for each `.so` file to be compiled sequentially whereas this requests them all at once and then leaves it up to the browser to decide what happens. Though if we wanted to do a thorough analysis we would also need to look carefully at what happens inside of Emscripten. Unfortunately the control flow there is callback heavy and so it's quite hard to read.

rth:
> I believe that this is not true: I think WebAssembly.instantiate and the other asynchronous apis in the WebAssembly namespace run in a different thread than the main runtime.

Yes, indeed for instance in https://github.com/pyodide/pyodide/issues/347#issuecomment-933834657 profiling pandas used 8 threads (though not fully loaded). Though it's indeed unclear if it would create new threads, or if it has a fixed thread pool that would be used. In the first case, there is a potential issue of CPU oversubscription. Say in pandas + numpy there are ~60 dynamic libraries with a total dynamic library initialization time of 6s or about 100ms per .so file.  So if we run all of this sequentially that could create 60 x (n_inner_threads) threads which, if that happens would lead to CPU oversubscription and likely degraded performance. In that case, limiting concurrency with e.g. batching https://stackoverflow.com/a/58686835/1791279 would likely be better.  Anyway for now this is all pure speculation, we need to profile to see what's happening.  I found the  Firefox profiling results quite useful. 

> Though if we wanted to do a thorough analysis we would also need to look carefully at what happens inside of Emscripten.

and likely also in-browser implementation of WebAssembly.instantiate etc.

PR #2584, opened 2022-05-19
Fix support for extras in micropip

amitsaha:
### Description

Fix for https://github.com/pyodide/pyodide/issues/2535

### Checklists

- [x] Add a [CHANGELOG](https://github.com/pyodide/pyodide/blob/main/docs/project/changelog.md) entry
- [x] Add / update tests

Issue #2535, opened 2022-05-09
micropip: support extras


rth:
As reported by @Zac-HD in https://github.com/pyscript/pyscript/issues/105 , currently micropip ignores extras in requirements for instance,
```py
>>> import micropip
>>> await micropip.install("hypothesis[cli]")
```
doesn't install black while it's in the `cli` extras.

Related: 
> [pip.pypa.io/en/stable/cli/pip_install/#requirement-specifiers](https://pip.pypa.io/en/stable/cli/pip_install/#requirement-specifiers) and [peps.python.org/pep-0508/#extras](https://peps.python.org/pep-0508/#extras)

amitsaha:
I will have a go at this one.

hoodmane:
We already try to support extras but apparently not successfully:
https://github.com/pyodide/pyodide/blob/b0f7feb210060b21da35b2d6bf047d362d1790b5/packages/micropip/src/micropip/_micropip.py#L392

ryanking13:
> We already try to support extras but apparently not successfully:
> 
> https://github.com/pyodide/pyodide/blob/b0f7feb210060b21da35b2d6bf047d362d1790b5/packages/micropip/src/micropip/_micropip.py#L392

Yeah... it is apparently a bug. Thanks for working on this @amitsaha!

amitsaha:
The bug is I believe here: https://github.com/pyodide/pyodide/blob/b0f7feb210060b21da35b2d6bf047d362d1790b5/packages/micropip/src/micropip/_micropip.py#L324

This condition here, evaluates to `False` since the `extra` "env" is set to an empty string by default. I am not exactly sure what the right fix is here. One guess and a "fix" I am contemplating is this change:

```
 if req.marker:
+            if str(req.marker._markers[0][0]) == "extra":
+                self.ctx["extra"] = str(req.marker._markers[0][2])
```
The above change will result in setting the value of `extra` to whatever the value of the `extra` requires the package requirement describes. I have to read up more on the spec and perhaps the implementation of `pip` to better understand whether this is the right fix or not. Any thoughts, please share.

amitsaha:
Thanks @ryanking13 - appreciate the insights. For now, based on the `pip` snippet, i have this code: https://github.com/pyodide/pyodide/commit/d51eaeeb6ab28e029efcc0eead7176b5fe76f5a2 

Some notes:

[...]

PR #2593, opened 2022-05-22
MAINT Add support for testing extras to `mock_fetch` fixture

hoodmane:
This PR adds support for extras to the `mock_fetch` fixture. A simple example would be as follows:
```python
@pytest.mark.asyncio
async def test_extras(mock_fetch: mock_fetch_cls, mock_importlib):
    mock_fetch.add_pkg_version("reportlab")
    mock_fetch.add_pkg_version("beagle-vote", extras={"pdf" : ["reportlab"]})

    await micropip.install("beagle-vote[pdf]")

    pkg_list = micropip.list()
    assert "reportlab" in pkg_list
```
This test does not pass #2535 but hopefully it will pass on branch #2584.

### DIFF
--- a/packages/micropip/test_micropip.py
+++ b/packages/micropip/test_micropip.py
@@ -74,32 +74,44 @@ class mock_fetch_cls:
         self.releases_map = {}
         self.metadata_map = {}
 
-    def add_pkg(
+    def add_pkg_version(
         self,
         name: str,
-        requirements: dict[str, list[str]],
+        version: str = "1.0.0",
+        *,
+        requirements: list[str] | None = None,
+        extras: dict[str, list[str]] | None = None,
         platform: str = "generic",
     ):
-        releases = {}
-        for version, reqs in requirements.items():
-            filename = make_wheel_filename(name, version, platform)
-            releases[version] = [
-                {
-                    "filename": filename,
-                    "url": filename,
-                    "digests": {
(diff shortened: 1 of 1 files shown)


packages/micropip/test_micropip.py
```diff
@@ -74,32 +74,44 @@ def __init__(self):
...
+                "digests": {
+                    "sha256": Wildcard(),
+                },
+            }
+        ]
+        metadata = [("Name", name), ("Version", version)] + [
+            ("Requires-Dist", req) for req in requirements
+        ]
+        for extra, reqs in extras.items():
+            metadata += [("Provides-Extra", extra)] + [
```
amitsaha, line 107: @hoodmane I am just catching up on this, and to learn more. Where can I find a description of this metadata specification which explains what `Provides-Extra` is? I understand what it is practically, but would like to read a definitive source on it.

amitsaha: > @hoodmane I am just catching up on this, and to learn more. Where can I find a description of this metadata specification which explains what `Provides-Extra` is? I understand what it is practically, but would like to read a definitive source on it.

Never mind i think i found some reference in the pip source code, i guess i have a good starting point there. thanks.

hoodmane: The other thing I recommend is looking at the actual `METADATA` files for installed wheels in your system.

PR #2595, opened 2022-05-23
fix(types): removing implicit generics

henryiii:

### DIFF
--- a/pyodide-build/pyodide_build/buildall.py
+++ b/pyodide-build/pyodide_build/buildall.py
@@ -5,6 +5,7 @@ Build all of the packages in a given directory.
 """
 
 import argparse
+import dataclasses
 import hashlib
 import json
 import os
@@ -30,11 +31,13 @@ class BuildError(Exception):
         super().__init__()
 
 
+@total_ordering
+@dataclasses.dataclass(eq=False, repr=False)
 class BasePackage:
     pkgdir: Path
     name: str
     version: str
-    meta: dict
+    meta: dict[str, Any]
     library: bool
     shared_library: bool
     dependencies: list[str]
@@ -43,6 +46,7 @@ class BasePackage:
     unvendored_tests: Path | None = None
     file_name: str | None = None
     install_dir: str = "site"
+    _queue_idx: int | None = None
 
     # We use this in the priority queue, which pops off the smallest element.
     # So we want the smallest element to have the largest number of dependents
@@ -60,8 +64,11 @@ class BasePackage:
             self.pkgdir, self.pkgdir / "build", self.meta.get("source", {})
         )
 
+    def build(self, outputdir: Path, args: Any) -> None:
+        raise NotImplementedError()
 
-@total_ordering
+
+@dataclasses.dataclass
 class StdLibPackage(BasePackage):
     def __init__(self, pkgdir: Path):
         self.pkgdir = pkgdir
@@ -75,7 +82,7 @@ class StdLibPackage(BasePackage):
         self.dependents = set()
         self.install_dir = "lib"
 
-    def build(self, outputdir: Path, args) -> None:
+    def build(self, outputdir: Path, args: Any) -> None:
         # All build / packaging steps are already done in the main Makefile
         return
 
@@ -86,7 +93,7 @@ class StdLibPackage(BasePackage):
         return None
 
 
-@total_ordering
+@dataclasses.dataclass
 class Package(BasePackage):
     def __init__(self, pkgdir: Path):
         self.pkgdir = pkgdir
@@ -340,7 +347,7 @@ def build_from_graph(pkg_map: dict[str, BasePackage], outputdir: Path, args) -> 
 
     # Insert packages into build_queue. We *must* do this after counting
     # dependents, because the ordering ought not to change after insertion.
-    build_queue: PriorityQueue = PriorityQueue()
+    build_queue: PriorityQueue[tuple[int, BasePackage]] = PriorityQueue()
 
     if args.force_rebuild:
         # If "force_rebuild" is set, just rebuild everything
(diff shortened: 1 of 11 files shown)


pyodide-build/pyodide_build/buildall.py
```diff
@@ -30,11 +31,13 @@ def __init__(self, returncode):
         super().__init__()
 
 
+@total_ordering
+@dataclasses.dataclass(eq=False, repr=False)
```
henryiii, line 35: It is not valid to set attributes directly on a class, that's very, very likely to not work as expected. These are actually data classes, which transforms the class attributes into instance attributes.

pyodide-build/pyodide_build/buildall.py
```diff
@@ -43,6 +46,7 @@ class BasePackage:
     unvendored_tests: Path | None = None
     file_name: str | None = None
     install_dir: str = "site"
+    _queue_idx: int | None = None
```
henryiii, line 49: Best to always declare all attributes of a class (same as with `.build` below).

pyodide-build/pyodide_build/buildall.py
```diff
@@ -412,11 +419,15 @@ def builder(n):
...
-        if isinstance(pkg, Exception):
-            raise pkg
+        match built_queue.get():
+            case BuildError() as err:
+                raise SystemExit(err.returncode)
+            case Exception() as err:
+                raise err
+            case a_package:
+                # MyPy should understand that this is a BasePackage
+                assert not isinstance(a_package, Exception)
```
henryiii, line 429: This shouldn't be necessary, will probably raise an issue.

pyodide-build/pyodide_build/buildpkg.py
```diff
@@ -352,6 +352,7 @@ def patch(pkg_root: Path, srcpath: Path, src_metadata: dict[str, Any]):
             result = subprocess.run(
                 ["patch", "-p1", "--binary", "--verbose", "-i", pkg_root / patch],
                 check=False,
+                encoding="utf-8",
```
henryiii, line 355: These three paths were actually buggy, the error printer assumed strings and not bytes.

pyodide-build/pyodide_build/mkpkg.py
```diff
@@ -10,11 +10,37 @@
 import urllib.request
 import warnings
 from pathlib import Path
-from typing import Any, Literal
+from typing import Any, Literal, TypedDict
 
 from ruamel.yaml import YAML
 
 
+class URLDict(TypedDict):
```
henryiii, line 18: You don't need all arguments (total=False is the default), only the ones you access, but I went ahead and added them.

hoodmane: Wow neat. We could use this in a bunch of places.

henryiii: Yes, you can fully type all your dict interfaces, which is really nice. 3.11 added Required/NotRequired so you can specify individual keys rather than making a complete or minimal TypedDict with total=True/False.

If you are building your own data, though, dataclass is the right way to go. Dicts should be used like this, for reading external APIs and files.

henryiii: (I could have also handled nested items here, but this was fine to check most of the usage)

hoodmane: > dataclass is the right way to go. Dicts should be used like this, for reading external APIs and files.

Thanks for the tip. I switched micropip to using dataclasses for `Transaction` and `Wheel` and it was indeed a much better approach because then it became apparent that I could reorganize a lot of the functions into methods on these classes and the code clarity improved a lot.

hoodmane: Ideally it would be nice to share more logic between `pyodide_build` and `micropip` since there is a significant amount of overlap.

src/py/_pyodide/_core_docs.py
```diff
@@ -154,7 +154,9 @@ def default_converter(value, convert, cache):
         """
         pass
 
-    def then(self, onfulfilled: Callable, onrejected: Callable) -> "Promise":
+    def then(
+        self, onfulfilled: Callable[..., Any], onrejected: Callable[..., Any]
```
hoodmane, line 158: I think all of these should be `Callable[Any, Any]` except `create_once_callable` which can be anything.

Eventually it would be nice to offer improved typing support for downstream. Most `JsProxy` objects have different subsets of the methods in this file, along with other methods from JavaScript. Of course the point of this file is primarily to generate docstrings and API docs, the use with mypy is very limited.

hoodmane (approved):
Thanks @henryiii!

pyodide-build/pyodide_build/mkpkg.py
rth: > Dicts should be used like this, for reading external APIs and files.

At least for meta.yaml we can probably switch to Pydantic to validate the spec, rather than doing our own custom dict validations.

rth (approved):
Thanks, LGTM as well.

henryiii:
Don't think I broke this?

```
======================================================================
[1576](https://github.com/pyodide/pyodide/runs/6582857449?check_suite_focus=true#step:6:1577)
FAIL: test_unencodeable (test.test_imp.NullImporterTests)
[1577](https://github.com/pyodide/pyodide/runs/6582857449?check_suite_focus=true#step:6:1578)
----------------------------------------------------------------------
[1578](https://github.com/pyodide/pyodide/runs/6582857449?check_suite_focus=true#step:6:1579)
[1579](https://github.com/pyodide/pyodide/runs/6582857449?check_suite_focus=true#step:6:1580)
  File "/lib/python3.10/test/test_imp.py", line 454, in test_unencodeable
[1580](https://github.com/pyodide/pyodide/runs/6582857449?check_suite_focus=true#step:6:1581)
    self.assertRaises(ImportError, imp.NullImporter, name)
[1581](https://github.com/pyodide/pyodide/runs/6582857449?check_suite_focus=true#step:6:1582)
AssertionError: ImportError not raised by NullImporter
```

hoodmane:
No `cpython_core[test_imp]` has sporadic failures.

hoodmane:
Thanks @henryiii!

PR #2596, opened 2022-05-23
fix(types): explicit reexport & strict equality

henryiii:
### Description

Enabling two more strictness flags.


src/py/pyodide/__init__.py
```diff
@@ -53,27 +53,31 @@
...
 
 __all__ = [
-    "open_url",
-    "eval_code",
-    "eval_code_async",
     "CodeRunner",
-    "find_imports",
-    "JsProxy",
+    "ConversionError",
+    "IN_BROWSER",
```
henryiii, line 58: This is the only added item, I also sorted the list.

src/py/pyodide/__init__.py
```diff
@@ -53,27 +53,31 @@
...
+    "run_js",
+    "set_interval",
+    "set_timeout",
+    "should_quiet",
+    "to_js",
+    "unregister_js_module",
 ]
+
+def __dir__() -> list[str]:
+    return __all__
```
henryiii: This ensures tab completion only shows the public items and is a handy thing to have on modules with an `__all__`.

src/tests/test_bz2.py
```diff
@@ -7,6 +7,6 @@ def test_bz2():
 
     text = "Hello test test test test this is a test test test"
     some_compressed_bytes = bz2.compress(text.encode("utf-8"))
-    assert some_compressed_bytes != text
+    assert some_compressed_bytes != text.encode("utf-8")
```
henryiii, line 10: This is likely a mistake, as otherwise it's comparing str to bytes, which is always not true.

src/py/pyodide/__init__.py
hoodmane: I don't think we want `IN_BROWSER` as part of the public API since it should always be `True` except when testing. Also the name is misleading because it is `True` in Node.

Skylion007 (approved):

src/py/pyodide/__init__.py
hoodmane: I guess `micropip` should be updated to import it from `pyodide._core`

henryiii: Done

hoodmane (approved):
Thanks @henryiii! This is very helpful.

PR #2598, opened 2022-05-23
Fix pyfetch when there is a cors error

hoodmane:
As pointed out by @ryanking13 in https://github.com/pyodide/pyodide/pull/2589#discussion_r878950933:
If `pyfetch` generates a cors error, it is thrown as a `JsException` rather than as an `OSError`.

### Checklists

- [x] Add / update tests


hoodmane:
@ryanking13 I suppose I should use `raise from None`

rth (approved):
LGTM, thanks!

PR #2603, opened 2022-05-24
MAINT Remove outdated chrome 89 xfails

hoodmane:


rth (approved):
Thanks! Sounds good assuming CI passes.

hoodmane:
It would be good to go through and check which other xfails are still needed. These are the only ones I could find that explicitly mentioned the Chrome v89 bug.

PR #2612, opened 2022-05-25
chore(types): fill out remaining partial types

henryiii:
### Description

Filling out remaining partially typed functions. Had to guess on a few, maybe some Any's could be avoided.


pyodide-build/pyodide_build/buildpkg.py
```diff
@@ -294,7 +296,7 @@ def prepare_source(
 
     Returns
     -------
-        The location where the source ended up.
+        The location where the source ended up. TODO: None, actually?
```
henryiii, line 299: Pretty sure this is wrong, the function returns None...

pyodide-build/pyodide_build/pywasmcross.py
```diff
@@ -87,16 +87,18 @@ def compile(
...
     replace_libs: str,
-):
+) -> None:
     ...
 
 
 @overload
-def compile(*, mypy__Single_overload_definition_multiple_required: int):
+def compile(
+    env: dict[str, str], *, mypy__Single_overload_definition_multiple_required: str
```
henryiii, line 96: This is only added because mypy requires two overloads - but it needs to be compatible with the implementation.

src/py/_pyodide/_core_docs.py
```diff
@@ -200,7 +200,7 @@ def assign(self, rhs: Any, /):
 
     # Argument should be a buffer.
     # See https://github.com/python/typing/issues/593
-    def assign_to(self, to: Any, /):
+    def assign_to(self, to: Any, /) -> None:
```
henryiii, line 203:
```suggestion
    def assign_to(self, to: Any, /) -> Any:
```

Oops, meant to make these all Any. Wasn't sure if they could be made more specific (or if some of them are None, etc). They are in docs so no implementation was handy.

src/py/_pyodide/_core_docs.py
```diff
@@ -223,7 +223,7 @@ def to_bytes(self) -> bytes:
         an ArrayBuffer view.
         """
 
-    def to_file(self, file: IOBase, /):
+    def to_file(self, file: IOBase, /) -> Any:
```
hoodmane, line 226:
```suggestion
    def to_file(self, file: IOBase, /) -> None:
```
https://github.com/pyodide/pyodide/blob/af972267c7969c6f023da709851463ec99e7e4ac/src/core/jsproxy.c#L1724

src/py/_pyodide/_core_docs.py
```diff
@@ -247,7 +247,7 @@ def to_file(self, file: IOBase, /):
         data once.
         """
 
-    def from_file(self, file: IOBase, /):
+    def from_file(self, file: IOBase, /) -> Any:
```
hoodmane, line 250: https://github.com/pyodide/pyodide/blob/af972267c7969c6f023da709851463ec99e7e4ac/src/core/jsproxy.c#L1743
```suggestion
    def from_file(self, file: IOBase, /) -> None:
```

src/py/_pyodide/_core_docs.py
```diff
@@ -273,7 +273,7 @@ def from_file(self, file: IOBase, /):
         data once.
         """
 
-    def _into_file(self, file: IOBase, /):
+    def _into_file(self, file: IOBase, /) -> Any:
```
hoodmane, line 276: https://github.com/pyodide/pyodide/blob/af972267c7969c6f023da709851463ec99e7e4ac/src/core/jsproxy.c#L1762
```suggestion
    def _into_file(self, file: IOBase, /) -> None:
```

src/py/_pyodide/_core_docs.py
```diff
@@ -466,7 +466,7 @@ class Promise(JsProxy):
     pass
 
 
-def destroy_proxies(pyproxies: JsProxy, /):
+def destroy_proxies(pyproxies: JsProxy, /) -> Any:
```
hoodmane, line 469:
```suggestion
def destroy_proxies(pyproxies: JsProxy, /) -> None:
```
https://github.com/pyodide/pyodide/blob/af972267c7969c6f023da709851463ec99e7e4ac/src/core/python2js.c#L838

pyodide-build/pyodide_build/pywasmcross.py
hoodmane: Would be nice if we could get mypy to fix this. Giving typings for `**kwargs` functions is useful. Between:
```py
def f(*, arg1 : Type1, arg2 : Type2, arg3 : Type3):
   d = dict(arg1=arg1, arg2=arg2, arg3=arg3)
```
and 
```py
@overload
def f(*, arg1 : Type1, arg2 : Type2, arg3 : Type3): ...

def f(**kwargs):
   d = kwargs
```
The second requires significantly less repetition.

henryiii: The former works at runtime too, though, you can inspect the signature. The later is a rather ugly hack.

hoodmane: Well if you think so I am fine with replacing it with the explicit way.

henryiii: FYI, I've used another ugly hack before:

```python
def f(*, arg1 : Type1, arg2 : Type2, arg3 : Type3):
   d = dict(**locals())
```

🤣 - It's not like this gets rewritten every day, or changes too much (I think), so the explicit, static way is likely fine.

packages/micropip/test_micropip.py
```diff
@@ -508,7 +511,9 @@ async def test_list_wheel_package(
 
 
 @pytest.mark.asyncio
-async def test_list_wheel_name_mismatch(mock_fetch: mock_fetch_cls, mock_importlib):
+async def test_list_wheel_name_mismatch(
+    mock_fetch: mock_fetch_cls, mock_importlib: Any
```
hoodmane, line 515:
```suggestion
    mock_fetch: mock_fetch_cls, mock_importlib: None
```

pyodide-build/pyodide_build/buildall.py
```diff
@@ -271,7 +271,7 @@ def print_with_progress_line(str, progress_line) -> None:
         print(progress_line, end="\r")
 
 
-def get_progress_line(package_set) -> str | None:
+def get_progress_line(package_set: dict[str, Any]) -> str | None:
```
hoodmane, line 274:
```suggestion
def get_progress_line(package_set: dict[str, None]) -> str | None:
```
`package_set` is a sorted set, so we use a dictionary.

pyodide-build/pyodide_build/buildall.py
```diff
@@ -327,7 +327,9 @@ def generate_needs_build_set(pkg_map: dict[str, BasePackage]) -> set[str]:
     return needs_build
 
 
-def build_from_graph(pkg_map: dict[str, BasePackage], outputdir: Path, args) -> None:
+def build_from_graph(
+    pkg_map: dict[str, BasePackage], outputdir: Path, args: Any
```
hoodmane, line 331:
```suggestion
    pkg_map: dict[str, BasePackage], outputdir: Path, args: argparse.Namespace
```

src/py/_pyodide/_core_docs.py
```diff
@@ -200,7 +200,7 @@ def assign(self, rhs: Any, /):
 
     # Argument should be a buffer.
     # See https://github.com/python/typing/issues/593
-    def assign_to(self, to: Any, /):
+    def assign_to(self, to: Any, /) -> Any:
```
hoodmane, line 203:
```suggestion
    def assign_to(self, to: Any, /) -> None:
```

pyodide-build/pyodide_build/buildall.py
henryiii: Pretty sure this breaks tests. But we’ll see.

packages/micropip/test_micropip.py
```diff
@@ -508,7 +511,9 @@ async def test_list_wheel_package(
 
 
 @pytest.mark.asyncio
-async def test_list_wheel_name_mismatch(mock_fetch: mock_fetch_cls, mock_importlib):
+async def test_list_wheel_name_mismatch(
+    mock_fetch: mock_fetch_cls, mock_importlib: None
+) -> None:
```
rth, line 516: I'm a bit dubious about the usefulness of typing the return type in all test functions, but I imagine without it mypy would complain with the strict typing options?

rth (approved):

packages/micropip/test_micropip.py
henryiii: I have mostly not added typing to anything yet. I’m only finishing off the typing for things that were partially typed. I would probably avoid requiring tests to be fully typed. The cost to benefit ratio is low.

henryiii: https://twitter.com/henryschreiner3/status/1527091252709183492?s=21&t=yeEMFASJ1TZml-hCG2CoWA

hoodmane: It's useful to put `mock_fetch_cls` in because then language servers know what attributes `mock_fetch` will have.

henryiii: Types are not harmful in tests, just not really worth the effort to add. They still improve readability (especially for fixtures), and IDE's benefit from them. But if you add one type, it's good to go ahead and add all of them, otherwise they are just implicitly Any anyway.

PR #2614, opened 2022-05-25
Mkpkg improvements

joemarshall:
I recently had to make a load of meta.yaml files, and I got annoyed by a couple of things so I did some code to fix them.

1) mkpkg doesn't build meta files for recursive dependencies.
2) the import name is often incorrect in meta files.

This PR adds two options to mkpkg.py
1) --make-dependencies which makes meta files for the target package and all dependencies.
2)--find-correct-imports which downloads the wheel / sdist and tries to find correct import names.


pyodide-build/pyodide_build/mkpkg.py
```diff
@@ -119,11 +166,106 @@ def run_prettier(meta_path):
...
 
+def _make_project_compare_name(project_name):
+    project_name_cleaned = project_name.replace("_", "").replace("-", "").lower()
+    if project_name_cleaned.startswith("py"):
+        project_name_cleaned = project_name[2:]
+    return project_name_cleaned
+
+
+# download a package from pypi and guess what modules are provided by it
+def _find_imports_from_package(url, project_name):
```
hoodmane, line 177: Couldn't you just read `top_level.txt` from the wheel?

hoodmane:
Have you tried out `micropip.freeze()` added in #2581? I think it might be helpful in the case where you want to add a bunch of pure python packages to your lock file. In any case, it would be good to share code between `mkpkg` and `micropip`, though we haven't figured out a good approach to that yet.

joemarshall:
It doesn't use top-level.txt because that's a setuptools file. Other packagers don't always use it. Poetry doesn't for example.

It currently list all distributions in test, not just the top level one. E.g bob.jones will be listed as well as the top level distribution bob. Basically anything that's a folder with an init Py in it. Is that right, or should it only list top level ones in meta.yaml

Micropip freeze looks cool. Wouldn't work for my stuff for two reasons I think, 1) I want all the wheels locally downloaded, which would mean scripting messing around with packages.json anyway. 2) micropip doesn't try with anything that has c extensions, whereas meta.yaml plus source can actually work for simple packages (and my code nicely follows all the dependencies through and gives you a good starting point for porting work)

Incidentally, I really want a synchronous import statement to save having to load all package dependencies beforehand. I did implement it with unthrow but I'm not so happy about using unthrow on anything where I don't have complete control of the code running. I guess I could do it with serviceworker, but then it would break in incognito windows, or with shared memory but then we're breaking GitHub pages...
________________________________

pyodide-build/pyodide_build/mkpkg.py
```diff
@@ -100,6 +105,48 @@ def _find_dist(
...
+                            if spec.contains(str(v)):
+                                chosen_version = str(v)
+                                break
+                    except packaging.specifiers.InvalidSpecifier as e:
+                        raise MkpkgFailedException(
+                            f"Bad specifier for  {package}{version} from "
+                            f"https://pypi.org/pypi/{package}{version}/json: {e}"
+                        )
+
+        except urllib.error.HTTPError as e:
```
rth, line 136: Could you please move this just after the `main_metadata = json.load(fd)` line and dedent everything after it, since none of the following lines need `fd`

pyodide-build/pyodide_build/mkpkg.py
```diff
@@ -100,6 +105,48 @@ def _find_dist(
...
+                all_versions = sorted(map(pkg_resources.parse_version, all_versions))
+                this_ver = pkg_resources.parse_version(version)
+                if this_ver in all_versions:
+                    chosen_version = str(this_ver)
+                else:
+                    try:
+                        spec = packaging.specifiers.SpecifierSet(version)
+                        chosen_version = None
+                        for v in reversed(all_versions):
+                            if spec.contains(str(v)):
```
rth, line 127: Maybe use `spec.filter(all_versions)` and then re-sort the results?

https://packaging.pypa.io/en/latest/specifiers.html#packaging.specifiers.SpecifierSet.filter 

It would be shorter.

pyodide-build/pyodide_build/mkpkg.py
rth: I imagine the issue is that it wouldn't work with .tar.gz / .zip files.

Though we could use `top_level.txt` if it exists and otherwise fallback to  [`setuptools.find_packages`](https://setuptools.pypa.io/en/latest/userguide/package_discovery.html#finding-simple-packages). It's already a build dependency. This would require unzipping files into some temporary folder so it would be slower, but a least we wouldn't have to maintain that code.

Overall it could be good to split this function into,
 - a function that downloads the package to some path
 - a function that finds imports from either an archive or a folder (if we go with `find_packages`).

pyodide-build/pyodide_build/mkpkg.py
```diff
@@ -119,11 +166,106 @@ def run_prettier(meta_path):
...
+                return []
+            main_prefix = None
+            main_prefix_level = None
+            # check for first bit of the path that looks like the package name
+            # without any _ or -
+            for mod in all_modules:
+                level = len(mod.parts)
+                path_part = mod.parts[-1]
+                cleaned = _make_project_compare_name(path_part)
+                if cleaned == project_name_cleaned:
```
rth, line 225: I dont think this would be very reliable e.g.

 - scikit-learn that import as pandas
 - pyarrow that import as arrow
 - etc.

Maybe just listing all the modules with an `__init__.py` (e.g. found by `find_packages`) and letting the user remove the extra ones would be more reliable.

pyodide-build/pyodide_build/mkpkg.py
```diff
@@ -151,10 +293,131 @@ def make_package(
...
+    dist = YamlDistribution(package_dir, project_name=package, version=version)
+    requires_packages = dist.requires(extras=our_extras)
+
+    env = EnvironmentHelper()
+
+    if make_dependencies:
+        ws = pkg_resources.WorkingSet([])
+        ws.resolve(requires_packages, env=env, extras=our_extras)
+
+    for r in requires_packages:
```
rth, line 398: I think it's a bit problematic to try to re-compute dependencies, because it's a hard problem it's never going to be right exactly, and it feels a bit redundant to what we do in micropip. It's just a lot of code to maintain.

If we wanted to resolve dependencies, maybe, we could run
```
pip download <package-name>
```
in some temporary folder?
Which will do correct dependency resolution by pip, and then check the wheels it downloaded into the temp folder? Bonus point is that most of those would be wheels so they would contain the `top_level.txt` file with imports. Though one would still need to fetch their json metadata and in the case of a binary wheel take the .tar.gz.

rth (commented):
Thanks a lot for proposing these improvements!

hoodmane:
> It currently list all distributions in test, not just the top level one. E.g bob.jones will be listed as well as the top level distribution bob. Basically anything that's a folder with an init Py in it. Is that right, or should it only list top level ones in meta.yaml?

It is sufficient to list top level imports, the other imports only make a difference to tests.

hoodmane:
See also importlib's function to determine top level imports:
https://github.com/python/cpython/blob/main/Lib/importlib/metadata/__init__.py#L1038

joemarshall:
So I updated it to use setuptools, with one special case exception which is single python file modules - these don't work with find_packages. Seems to work fine.

The dependency resolution, I've looked at pip, and pip-compile, and I don't think either of those are really flexible enough for this. I've left it in for now. 

One possibility is to add the ability to run micropip in the build process and use that to find dependencies. Or alternatively a way for micropip to make a zip file of meta.yaml files for packages not found in the current pyodide would do the job here also.  That would be nice for things like jupyterlite, where you could run micropip in the actual environment with whatever jupyterlite loads up, and then get a nice zipfile of meta.yaml files somehow (I can't quite work out how one downloads it...)

joemarshall:
I don't quite understand why but it pre-commit is failing a check on types. It looks like it has an old version of  pkg_resources.Environment in the signatures file that mypy uses. The pre-commit hooks run fine on my local install. Error shown below:


```
pyodide-build/pyodide_build/mkpkg.py:340: error: Signature of "best_match" incompatible with supertype "Environment"  [override]
pyodide-build/pyodide_build/mkpkg.py:340: note:      Superclass:
pyodide-build/pyodide_build/mkpkg.py:340: note:          @overload
pyodide-build/pyodide_build/mkpkg.py:340: note:          def best_match(self, req: Requirement, working_set: WorkingSet) -> Distribution
pyodide-build/pyodide_build/mkpkg.py:340: note:          @overload
pyodide-build/pyodide_build/mkpkg.py:340: note:          def [_T] best_match(self, req: Requirement, working_set: WorkingSet, installer: Callable[[Requirement], _T] = ...) -> _T
pyodide-build/pyodide_build/mkpkg.py:340: note:      Subclass:
pyodide-build/pyodide_build/mkpkg.py:340: note:          @overload
pyodide-build/pyodide_build/mkpkg.py:340: note:          def best_match(self, req: Requirement, working_set: WorkingSet, *, replace_conflicting: bool = ...) -> Distribution
pyodide-build/pyodide_build/mkpkg.py:340: note:          @overload
pyodide-build/pyodide_build/mkpkg.py:340: note:          def [_T] best_match(self, req: Requirement, working_set: WorkingSet, installer: Callable[[Requirement], _T], replace_conflicting: bool = ...) -> _T
Found 1 error in 1 file (checked 43 source files)

```

joemarshall:
I had more of a dig into micropip. It needs the fix to imports for things without top_level.txt adding, but other than that it should be pretty trivial to make it a)download wheels and add their links to the pyodide build as meta.yaml files, b) optionally make meta.yaml files for the sdist if no built wheel exists (every so often these do just work unpatched, but if not they're still a good starting point for a port).

rth:
> One possibility is to add the ability to run micropip in the build process and use that to find dependencies

Yes, using micropip for dependency resolution would also be a good solution. At least we wouldn't have to maintain two versions of dependency resolution that way. Thanks!

ryanking13:
pyodide-build has been moved to pyodide/pyodide-build. Please open the PR there if you are still interested in working on this.

PR #2615, opened 2022-05-25
MAINT Remove `driver_timeout` argument from `run_in_pyodide` in favor of `pytest.mark.driver_timeout`

hoodmane:


pyodide-test-runner/pyodide_test_runner/decorator.py
```diff
@@ -134,10 +131,6 @@ def __init__(
         packages : List[str]
             List of packages to load before running the test
 
-        driver_timeout : Optional[float]
-            selenium driver timeout (in seconds). If missing, use the default
-            timeout.
```
rth, line 139: Could we document the marks somewhere, maybe in https://pyodide.org/en/latest/development/testing.html ?

As with https://github.com/pyodide/pyodide/blob/main/setup.cfg#L13 even if one find the right file where they are listed, it doesn't really say how to use it.

rth (approved):
Thanks! Minor comment otherwise LGTM.

pyodide-test-runner/pyodide_test_runner/decorator.py
hoodmane: I will try to update the testing docs in a separate PR. Currently I have adjusted the docs in #2617 so there will be a merge conflict if I update it here too.

packages/scipy/test_scipy.py
```diff
@@ -44,6 +46,7 @@ def test_binom_ppf():
     assert binom.ppf(0.9, 1000, 0.1) == 112
 
 
+@pytest.mark.driver_timeout(40)
```
hoodmane, line 49: I'm a bit confused about why this change caused `test_scipy_pytest` to start timing out.

ryanking13: It seems like the driver_timeout update in other tests was affecting the timeout of `test_scipy_pytest` before this PR (I don't understand why :/). `test_scipy_pytest` always took more than 20 seconds.

hoodmane: Yeah this must be correct, but I can't figure out why it was happening either.

PR #2616, opened 2022-05-25
MAINT Add pytest.mark.xfail_browsers and use it to reduce boilerplate

hoodmane:
Use a decorator to indicate which browsers we want to xfail test. This works for `run_in_pyodide` as well as normal tests.


rth:
Thanks! Looks like an improvement indeed. I'm trying to keep up review-wise with all the work you are doing but with mixed success so far.

PR #2617, opened 2022-05-25
MAINT run_in_pyodide function should take selenium fixture as an explicit argument

hoodmane:
This updates `run_in_pyodide` to assume that the first argument of the function will be a `selenium` fixture.


rth (approved):
This feels more natural indeed than passing it to run_in_pyodide. Thanks!

PR #2619, opened 2022-05-26
Use tblib to pickle errors in `run_in_pyodide`

hoodmane:


requirements.txt
```diff
@@ -15,5 +15,6 @@
   pytest-rerunfailures
   pytest-xdist
   selenium==4.1.0
+  tblib
```
ryanking13, line 18: Do we need tblib installed in the host?

ryanking13 (commented):
Thanks, @hoodmane. It is not directly related to this PR, but I recently found that it is a little non-intuitive that one should build pytest to run a test that uses `run_in_pyodide` (which is because `pytest_assert_rewrites` is true in default).

For example,

```sh
$ PYODIDE_PACKAGES="some_pkg" make
$ pytest packages/some_pkg
```

will complain about pytest not being installed, so one has to do

```sh
$ PYODIDE_PACKAGES="some_pkg,core" make
# or
$ PYODIDE_PACKAGES="some_pkg,pytest" make
```

to run a test.

This PR also adds a dependency to run a test, so we probably update some documentation, or find a better way to deal with inner(?) test dependencies.

requirements.txt
hoodmane: Yeah it's needed in the host to unpickle the tracebacks.

hoodmane:
I should fix it so that if pytest isn't installed it just falls back to not using it. Same is possible for tblib. They just generate better tracebacks when a test fails, so on failure we can append "for a better traceback, build pytest and tblib".

PR #2627, opened 2022-05-27
FIX Handle large lengths and negative lengths in hiwire_get_length

hoodmane:
While investigating this bug:

> ```
> Pyodide has suffered a fatal error. Please report this to the Pyodide maintainers.
> The cause of the fatal error was:
> PythonError: OverflowError: Python int too large to convert to C ssize_t
> ```
> 
> _Originally posted by @amitsaha in https://github.com/pyodide/pyodide/issues/2584#issuecomment-1139530294_

I found a different one. Turns out handling lengths correctly is quite hard.


src/core/hiwire.c
```diff
@@ -639,19 +639,84 @@ EM_JS_BOOL(bool, hiwire_has_length, (JsRef idobj), {
...
+  // * -1: Either `val.size` or `val.length` was a getter which managed to raise
+  //    an error. Rude. (Also we don't defend against this in hiwire_has_length)
+  // * -2: Doesn't have a length or size, or they aren't of type "number".
+  //   But `hiwire_has_length` returned true? So it must have changed somehow.
+  // * -3: Length was >= 2^{31}
+  // * -4: Length was negative
+  if (result == -2) {
+    PyErr_SetString(PyExc_TypeError, "object does not have a valid length");
+  }
+  if (result >= -2) {
```
ryanking13, line 694: Since it is just two cases, so maybe to be more explicit: 

```suggestion
  if (result == -1 || result == -2) {
```

ryanking13 (approved):
Thanks @hoodmane! LGTM.

PR #2628, opened 2022-05-27
MAINT Rather than fatally failing, throw away errors in check methods

hoodmane:
I think this will fix the following error:

> ```
> Pyodide has suffered a fatal error. Please report this to the Pyodide maintainers.
> The cause of the fatal error was:
> PythonError: OverflowError: Python int too large to convert to C ssize_t
> ```
> 
> _Originally posted by @amitsaha in https://github.com/pyodide/pyodide/issues/2584#issuecomment-1139530294_

I am not sure though because I can't reproduce the error.

### DIFF
--- a/src/core/error_handling.h
+++ b/src/core/error_handling.h
@@ -121,6 +121,19 @@ console_error_obj(JsRef obj);
     }                                                                          \
     return 0;  /* some of these were void */                                   \
   })
+
+// If there is a Js error, catch it and return false.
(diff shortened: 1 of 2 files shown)


hoodmane:
Okay I found a reproduction and this indeed fixes it:
```py
from pyodide import create_proxy; create_proxy(range(1<<31))
```
Thanks again to hypothesis for finding weird edge cases.

Issue #1946, opened 2021-11-10
Make JsProxy picklable


rth:
It would be good if PyProxy were picklable, which is a standard way of serializing objects in Python.

### Example
```py
>>> from js import Uint8Array
>>> import pickle
>>> a = Uint8Array.new(range(10))

>>> with open('a.pkl', 'w') as fh:
...     pickle.dump(a, fh)
... 
Traceback (most recent call last):
  File "<console>", line 2, in <module>
TypeError: cannot pickle 'pyodide.JsProxy' object
```

related https://github.com/pyodide/pyodide/issues/1620, https://github.com/pyodide/pyodide/pull/1942

hoodmane:
This is a good idea. The internal browser spec defines various serialization and deserialization methods. We would want `StructuredSerializeForStorage` and `StructuredDeserialize`. Unfortunately, these are internal browser APIs that are not exposed to Javascript. It's possible to trigger `StructuredSerializeForStorage` synchronously using the history API, but it is not clear that this is useful. I also don't think the history API is available in workers.

https://html.spec.whatwg.org/multipage/structured-data.html#structuredserialize

`structuredClone` is available now as a synchronous method in firefox:
https://github.com/whatwg/html/pull/3414

I don't see any proposal to expose structured serialize, or any discussion of serializing a Javascript object into an `ArrayBuffer`. Probably there are security issues: if I could serialize a js object into an `ArrayBuffer`, I could change some of its bytes arbitrarily and maybe de-serializing it would allow me to attack the browser engine in some way.

On the other hand, `TypedArray` objects should be easy to serialize: you just have to say "I am a JsProxy of a Uint8Array with this data". Should be a very similar task to serializing a Python buffer.

hoodmane:
On the other hand, `TypedArray` objects should be easy to serialize: you just have to say "I am a JsProxy of a Uint8Array with this data". Should be a very similar task to serializing a Python buffer.

hoodmane:
Unfortunately, it's not possible to pickle most JavaScript objects. But we should probably add support for pickling anything on which `JSON.stringify` succeeds.

hoodmane:
Actually, `JSON.stringify` is not very suitable. Maybe we should try using yabson:
https://github.com/gildas-lormeau/yabson/

rth:
So what we need library that is able to serialize base JS types but also custom instances (including functions and methods) right? 
[...]

PR #2631, opened 2022-05-28
ENH Add support for pickling JsException

hoodmane:
This makes run_in_pyoidde work a bit better when a JsException
is thrown.


ryanking13 (approved):
Thanks @hoodmane! Please update the changelog.

hoodmane:
Okay. I'm not really sure I want to advertise this change by itself, since it's really just for the testing system: currently if you try to unpickle it in Pyodide, it raises a `TypeError` because `js_error` is a string. But it unpickles fine in the Pyodide test shim.

ryanking13:
That sound reasonable. Thanks for the explanation!

hoodmane:
CF #1946.

PR #2632, opened 2022-05-28
MAINT Use new run_in_pyodide features for some type conversions tests

hoodmane:
This should improve the tests by quite a lot:
1. assertions get formatted automatically
2. many fewer doubly / triply nested eval strings
3. IDE support for code formatting, syntax highlighting (the IDEs are pretty confused about other stuff but still...)

Also, messing around with these type conversions tests is a really good way to find lots of bugs, cf my other PRs from this afternoon.


ryanking13:
Thanks a lot @hoodmane! It looks like there were lots of improvements during the weekend :)

PR #2635, opened 2022-05-28
Add spinner + better attribution in REPL

rth:
Several changes to the REPL,
 - better attribution of Emscripten (we talked about it but didn't do the change): the gif below doesn't include this change.
 - Attribute for the Jquery Terminal: I added a clickable grey link on the right top of the REPL. On mobile, I think it should be OK, since we don't really scale text currently, so there is still enough space to display it. There was an issue that this was the only text that was displayed while loading.  
 - So I ended up adding a spinner for loading as well. Potentially we could re-use it for indicating that a package is being loaded as it's continuously running in the background, we just need to put it in the foreground with `z-index: 2` and hide it again after the package is loaded.
 
Peek 2022-05-28 12-05


eagleoflqj:
IMHO it would be better if background color for spin is consistent with terminal it self. Now there's a white to black transition.
https://sympy-beta.vercel.app/terminal for reference.

hoodmane (approved):
LGTM, though @eagleoflqj's suggestion that we should use a consistent background color between the loading screen and the repl.

Also, in the future we should might try to detect when loading crashes, remove the spinner and print a message. The point of the spinner is to reassure people that the page is doing something. If it stays up even when the page has actually crashed it feels a bit disingenuous.

hoodmane:
I guess all we would need to fix the background color issue is to set 
```css
body {
    background-color: black;
}
```

rth:
> it would be better if background color for spin is consistent with terminal it self.

Thanks for the suggestion! Good idea, I made the change. The simpy REPL is quite nice.

> Also, in the future we should try to detect when loading crashes, remove the spinner and print a message.

Yeah, also ideally it would be to display the terminal, and then load in the background with some spinner/progress indicator (as the sympy or the Christian's REPL does for instance). By the time the person moves the mouse to click on it and reads the displayed terminal prompt banner, one gains a few seconds in perception as to how fast it loads.

PR #2643, opened 2022-05-30
Enable Wasm BigInt

hoodmane:
This enables WASM_BIGINT while maintaining (hypothetical) Safari 14 support
by shimming BigInt64Array and BigUint64Array if they are missing. I think the
last time we tried to enable WASM_BIGINT was before #2019 so our chances
are significantly better this time.

This will fix dynamic linking bugs, reduce code size, and improve speed.

See also my upstream PR:
https://github.com/emscripten-core/emscripten/pull/17103

### DIFF
--- a/src/core/jsmemops.h
+++ b/src/core/jsmemops.h
@@ -34,7 +34,7 @@
   (BigInt(lower) | (BigInt(upper) << BigInt(32)))
 
 #define IBIGINT_FROM_PAIR(lower, upper)                                        \
-  (BigInt(lower) | (BigInt(upper - 2 * (upper & 0x80000000)) << BigInt(32)))
+  (BigInt(lower) | (BigInt(upper + 2 * (upper & 0x80000000)) << BigInt(32)))
 
 #define LOAD_U64(addr, offset)                                                 \
   UBIGINT_FROM_PAIR(DEREF_U32(addr, offset * 2),                               \
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -104,8 +104,10 @@ jobs:
 
     needs: [build-core]
     strategy:
+      fail-fast: false
       matrix:
-        selenium_runner: [firefox]
+        runner: [playwright]
+        browser: [firefox, chrome]
 
     steps:
       - uses: actions/checkout@v2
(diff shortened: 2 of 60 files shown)


hoodmane:
Well scipy is not happy about this.

hoodmane:
Okay, looks like it works now =)

hoodmane:
This change reduces the size of `pyodide.asm.js`, `pyodide.asm.wasm`, and clapack each by about 100kb. No measured effect on the benchmarks.

src/core/jsmemops.h
```diff
@@ -34,7 +34,7 @@
   (BigInt(lower) | (BigInt(upper) << BigInt(32)))
 
 #define IBIGINT_FROM_PAIR(lower, upper)                                        \
-  (BigInt(lower) | (BigInt(upper - 2 * (upper & 0x80000000)) << BigInt(32)))
```
rth, line 37: Could you add an explanatory comment for this ?

rth (commented):
Great that you managed to make it work! Let's maybe wait for feedback to your emscripten PR, other than that LGTM (assuming the debug-related lines are reverted).

src/core/jsmemops.h
hoodmane: Yeah I will add a comment. To quote wikipedia:
https://en.wikipedia.org/wiki/Two%27s_complement#Converting_from_two's_complement_representation

> The value w of an N-bit twos-complement integer ${\displaystyle a_{N-1}a_{N-2}\dots a_{0}} a_{N-1} a_{N-2} \cdots a_0$ is given by the following formula:
>
> $w=-a_{N-1}2^{N-1}+\sum a_{i}2^{i}$

So if we have $v = a_{N-1}2^{N-1}+\sum a_{i}2^{i}$ and we want $w = -a_{N-1}2^{N-1}+\sum a_{i}2^{i}$ we have to subtract $2\cdot a_{N-1}2^{N-1}$. Here $N = 32$. Confusingly, in JavaScript if you take a positive number between $+2^{31}$ and $+2^{32}$ and compute `x & 0x80000000`, the result is $-2^{31}$ so we need to *add* `2*(x & 0x80000000)`.

I don't really understand how bitwise operators work in JavaScript, I always have to use a lot of trial and error. Everything is actually floats so to do bitwise ops it converts from floats to... I'm not sure what size integers and then does the bit ops and then converts everything back to floats. Since there isn't really a leading bit, I don't understand what determines when the result of a bitwise operation is positive or negative. Maybe I should read the ECMAScript spec. The rules for BigInts are equally confusing and also different.

hoodmane:
I split off two PRs from this one: #2650 and #2648.

hoodmane:
According to caniuse this is dropping support for ~0.7% of users.

rth:
> According to caniuse this is dropping support for ~0.7% of users.

Is the limit on the JS bigint support in general, or are there specific WASM bigint features that need to be supported?

hoodmane:
Just js features, 64 bit integer support is part of the wasm mvp. But they are awkward if JavaScript doesn't support them because so many things touch the wasm/js interface.

src/core/jsmemops.h
hoodmane: Okay further experimentation gives the answer: when you do any bitwise operation, the floats are cast to i32 and then the bitwise operation occurs and then they are cast back to float.  So `(2**32) || 0` is 0, `(2**32 - 1) || 0` is `-1`. JavaScript is weird.

rth:
https://caniuse.com/bigint in terms of supported browser versions looks very reasonable. We don't support Safari 13 anway.

Looks like upstream is happy with your patch :) Thanks for doing it. LGTM as well once the patch is updated to match your upstream PR.

src/core/jsmemops.h
rth: yes, it sounds more complicated than I would have expected. 

Should we add some tests for IBIGINT_FROM_PAIR  in JS?

hoodmane:
I already backported the most up-to-date version of the patch. It's not a 100% clean backport because some of the compiler guards changed between 2.0.27 and tip of tree.

hoodmane:
Thanks for the review @rth!

PR #2645, opened 2022-05-30
ENH Allow specifying a custom lock file

hoodmane:
This option is intended to be used with `micropip.freeze`. A user can
save the lockfile generated by `micropip.freeze` and load that lock
file while using the rest of the files from the CDN.

### Checklists

- [x] Add / update tests


rth (commented):
This is great, thanks @hoodmane !

Yes, I would be to add a test + changelog for it, otherwise LGTM.

PR #2646, opened 2022-05-30
Convert more typeconversions tests to use run_in_pyodide

hoodmane:


ryanking13 (approved):
Thanks @hoodmane! Generally LGTM.

PR #2647, opened 2022-05-30
MAINT Convert some jsproxy tests to using run_in_pyodide

hoodmane:


ryanking13 (approved):
Thanks for this work @hoodmane!

PR #2653, opened 2022-05-31
Fix run_in_pyodide to return values and update tests

hoodmane:
Adding `return` here allows `run_in_pyodide` functions to return values. Also, add a test specifying what should happen if a `run_in_pyodide` function tries to access a closure variable (it should `NameError`)


rth (approved):
Thanks!

PR #2655, opened 2022-05-31
MAINT Update the packages.json info to include emscripten version and pyodide version

hoodmane:
Split from #2645

### DIFF
--- a/pyodide-build/pyodide_build/buildall.py
+++ b/pyodide-build/pyodide_build/buildall.py
@@ -457,9 +457,16 @@ def generate_packages_json(
     output_dir: Path, pkg_map: dict[str, BasePackage]
 ) -> dict[str, dict[str, Any]]:
     """Generate the package.json file"""
+
+    import sys
+
+    sys.path.append(str(common.get_pyodide_root() / "src/py"))
+    from pyodide import __version__
+
     # Build package.json data.
+    [platform, _, arch] = common.platform().rpartition("_")
     package_data: dict[str, dict[str, Any]] = {
-        "info": {"arch": "wasm32", "platform": "Emscripten-1.0"},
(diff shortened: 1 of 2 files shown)


pyodide-build/pyodide_build/buildall.py
```diff
@@ -457,9 +457,16 @@ def generate_packages_json(
     output_dir: Path, pkg_map: dict[str, BasePackage]
 ) -> dict[str, dict[str, Any]]:
     """Generate the package.json file"""
+
+    import sys
+
+    sys.path.append(str(common.get_pyodide_root() / "src/py"))
+    from pyodide import __version__
```
rth, line 464: I'm not sure that this package would always be importable on the host in the future, but we can deal with that issue when it happens.

rth (approved):
Thanks!

pyodide-build/pyodide_build/buildall.py
```diff
@@ -457,9 +457,16 @@ def generate_packages_json(
     output_dir: Path, pkg_map: dict[str, BasePackage]
 ) -> dict[str, dict[str, Any]]:
     """Generate the package.json file"""
```
ryanking13, line 459: For example, even maintainers often make a typo just like here.

ryanking13 (commented):
Thanks @hoodmane! Probably we should consider renaming `packages.json` as it is starting to contain more than just package information and sometimes it is confusing as its name is similar to `package.json` of Node.js.

ryanking13 (approved):
Thanks, @hoodmane! There is a failing test, otherwise LGTM.

hoodmane:
Okay I'll open a followup to change `packages.json` to `repodata.json`.

PR #2658, opened 2022-06-02
Remove tzdata from the build step

ryanking13:
### Description

`tzdata` is a first-party package primarily used in a standard library `zoneinfo` as a fall-back source of data for platforms that don't ship the IANA database.
(See: https://github.com/python/cpython/blob/877ad7b3b2778a305d3989d58ebd68cb01baf26e/Doc/whatsnew/3.9.rst#new-modules)

Currently, we install `tzdata` during the build step in default (added in #1637). I guess we added it because there is no system time zone data available in browser environments (cc @hoodmane).

However, I think we don't have to include it in the Pyodide distribution by default.
For example, in Windows (which also doesn't have system time zone data), `tzdata` is not included by default.

WDYT?

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -25,9 +25,6 @@ all: check \
 	dist/module_webworker_dev.js
 	echo -e "\nSUCCESS!"
 
-$(CPYTHONLIB)/tzdata :
-	pip install tzdata --target=$(CPYTHONLIB)
-
 dist/pyodide_py.tar: $(wildcard src/py/pyodide/*.py)  $(wildcard src/py/_pyodide/*.py)
 	cd src/py && tar --exclude '*__pycache__*' -cf ../../dist/pyodide_py.tar pyodide _pyodide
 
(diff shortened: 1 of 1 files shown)


hoodmane:
I added it to fix a build failure I think -- probably have to look back at the upgrade to python 3.9 pr. If the build succeeds without it I am okay to only load it conditionally.

hoodmane:
Shouldn't we add `tzdata` to the packages? Maybe include `zoneinfo` as an import name.

ryanking13:
> I was confusing it with pytz which we already ship. There are two ways to look at it,
> 
> 1. on one side you are right that it could be useful to package it so it's auto-loaded on import as it's used in the stdlib
> 2. on the other it's a pure Python package, most users use micropip anyway, and packaging it will not solve the problem that it would not be loaded when imported in zoneinfo
> 
> so I'm not sure, no strong opinion either way.

I think one can just do `micropip.install("tzdata")`.

ryanking13:
Thanks for the reviews!

PR #2668, opened 2022-06-07
Update emscripten to v3.1.13

hoodmane:
This drops support for firefox 95 -- 98. According to https://caniuse.com/usage-table, these versions total to 0.06% of users' browsers.


hoodmane:
Closed in favor of #2679.

PR #2671, opened 2022-06-07
ENH Add a new "disabled" key to meta.yaml

hoodmane:
Frequently when updating Python or Emscripten, a handlful of packages
stop working. Because of the way that the CI works, these prevent
other packages from being built or tested. However, packages with
many dependents are currently annoying to disable. This adds a key
to meta.yaml "disable: true" that turns off a package.

I also fixed the "!package" and no-numpy-dependents to be transitive.

We use graphlib to sort the packages topologically, then we traverse
the packages once in build order to locate all the packages which
transitively depend on disabled packages. Then we traverse in reverse
build order to locate all packages that are dependencies of non-disabled
requested packages.

### DIFF
--- a/packages/scikit-learn/meta.yaml
+++ b/packages/scikit-learn/meta.yaml
@@ -15,7 +15,6 @@ build:
 
 requirements:
   run:
-    - numpy
     - scipy
     - joblib
     - threadpoolctl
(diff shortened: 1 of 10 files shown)


ryanking13 (approved):
Thanks, @hoodmane! It looks very convenient.

If you think documenting `disabled` key will be useful for other users. Please add it to the docs. Personally I think it is okay to be used privately (in that case maybe rename it to `_disabled`).

PR #2673, opened 2022-06-08
MAINT Switch to using a tag to version libffi port

hoodmane:
I think it's better to tag the libffi commit we're going to use with the date and then use the tag in our Makefile rather than using a raw commit hash.


hoodmane:
> Is there a chance some of it could be accepted upstream?

I don't see why not.

PR #2675, opened 2022-06-08
Fix groupadd/useradd in `run_docker` so $HOME is set correctly

### LINKED ISSUE #2674: Permissions issue when running in the Docker image
## 🐛 Bug

I'm trying to build `pyodide` and some packages using the Docker image, but get tons of 
```
ccache: error: Failed to create directory //.ccache/tmp: Permission denied
```
errors when not running `make` as root. Running as root does work. According to the docs, this should work I think and I can't find any other issue about this.

### To Reproduce

```
./run_docker
```
and then
```
make
```
inside the container.

### Expected behavior

No permissions issues and a working build!

### Environment

This is using the current `main` branch. I'm using a new docker installation on CentOS 7.

- Commit hash: 587a73fa14e1ffceec5380d884214f9d36d45be1
- Build environment: CentOS Linux release 7.9.2009 (Core)

### Additional context

Running
```
./run_docker --root
```
works fine.

jobovy:
Fixes #2674

### DIFF
--- a/run_docker
+++ b/run_docker
@@ -127,9 +127,9 @@ CONTAINER=$(\
     --shm-size 2g \
     "${PYODIDE_DOCKER_IMAGE}" \
     /bin/bash -c " \
-      groupadd '$USER_GID'; \
+      groupadd --gid '$USER_GID' '$USER_NAME'; \
       useradd \
-        --home-dir '$USER_HOME' \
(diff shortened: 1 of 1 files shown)


hoodmane (approved):
Seems to work fine.

PR #2676, opened 2022-06-08
PKG Add galpy

jobovy:
### Description

Add [`galpy`](https://github.com/jobovy/galpy), a widely-used package in astrophysics for Galactic dynamics. `galpy` has been used [more than 450 times](https://docs.galpy.org/en/latest/#papers-using-galpy) in the astrophysical literature and is commonly used in undergraduate and graduate courses. It's also used in [my online textbook on galaxies](https://galaxiesbook.org/), for which I would eventually like to build a `pyodide`-backed interactive version (currently provided by Binder). So it would be great to add this to `pyodide`!

There are a few minor installation patches required that I will upstream for the next `galpy` release.

Thanks @[benjeffery](https://github.com/benjeffery) in #2548 for adding the GSL which made this easy!

### Checklists

- [ ] Add a [CHANGELOG](https://github.com/pyodide/pyodide/blob/main/docs/project/changelog.md) entry (will add once I know the PR number)
- [x] Add / update tests

### DIFF
--- a/packages/galpy/meta.yaml
+++ b/packages/galpy/meta.yaml
@@ -1,0 +1,36 @@
+package:
+  name: galpy
+  version: 1.7.2
+source:
+  url: https://files.pythonhosted.org/packages/67/7f/ae16e887e78d42c8db0406ca7eec651502ebf3e1c9782a00f59e506a8db3/galpy-1.7.2.tar.gz
+  sha256: 4011b57b427d791addd8201754989a8f08d46925ee81cd571473d414ee38c04a
+  patches:
+    - patches/0001-Force-GSL-version-to-2.7-for-pyodide-build.patch
+    - patches/0002-Don-t-use-OpenMP-for-pyodide-build.patch
+build:
+  cflags: |
+    -I$(PYODIDE_ROOT)/packages/libgsl/build/libgsl-2.7
+  ldflags: |
+    -L$(PYODIDE_ROOT)/packages/libgsl/build/libgsl-2.7/.libs
+    -L$(PYODIDE_ROOT)/packages/libgsl/build/libgsl-2.7/cblas/.libs
+requirements:
+  run:
+    - numpy
+    - scipy
+    - matplotlib
+    - astropy
+    - libgsl
+    - future
+    - setuptools
+test:
+  imports:
+    - galpy
+    - galpy.potential
+    - galpy.orbit
+    - galpy.actionAngle
+    - galpy.df
+about:
+  home: http://github.com/jobovy/galpy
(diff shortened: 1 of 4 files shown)


packages/galpy/meta.yaml
```diff
@@ -0,0 +1,36 @@
...
+  sha256: 4011b57b427d791addd8201754989a8f08d46925ee81cd571473d414ee38c04a
+  patches:
+    - patches/0001-Force-GSL-version-to-2.7-for-pyodide-build.patch
+    - patches/0002-Don-t-use-OpenMP-for-pyodide-build.patch
+build:
+  cflags: |
+    -I$(PYODIDE_ROOT)/packages/libgsl/build/libgsl-2.7
+  ldflags: |
+    -L$(PYODIDE_ROOT)/packages/libgsl/build/libgsl-2.7/.libs
+    -L$(PYODIDE_ROOT)/packages/libgsl/build/libgsl-2.7/cblas/.libs
```
hoodmane, line 15: Please use `pkg-config` for this. (See packages/msprime/meta.yaml.)

jobovy: Ah, will do! I was looking at the `msprime` PR, didn't notice that there had been updates...

hoodmane:
Also, please add some tests.

jobovy:
Okay, I fixed the `pkg-config` use and added some basic tests. I had to exclude `galpy` from the node tests and I had to skip the check of the number of 'hiwire' keys. Both of these appear necessary because `galpy` loads `matplotlib` and it seems these checks don't work with `matplotlib` (the node one because there is no `matplotlib` backend for node, the hiwire one I don't understand, but it's skipped in `test_matplotlib.py`, so I just copied that code).

hoodmane (approved):

hoodmane:
The refcount check issue is that matplotlib adds a bunch of global JavaScript objects that are never going to be removed. The fix would be to load and import micropip in a fixture which would count the statically allocated objects as part of the baseline. But we've been too lazy to fix it.

packages/galpy/meta.yaml
```diff
@@ -27,10 +27,6 @@ requirements:
 test:
   imports:
     - galpy
-    - galpy.potential
```
hoodmane: Put these back and add galpy to the list of `UNSUPPORTED_PACKAGES` for `node` on line 25 of `packages/test_packages_common.py`

jobovy:
Thanks @hoodmane and @rth! Yes, I'll apply these patches upstream so they shouldn't be necessary with the next release.

PR #2677, opened 2022-06-08
FIX Don't build a shared library for libgsl

hoodmane:
It is slow, we aren't using it, and it is incorrect. Newer versions
of Emscripten notice that there are problems and error out.


benjeffery:
@hoodmane Thanks for spotting that, sorry I didn't notice it when I added GSL!

hoodmane:
Ugh, hypothesis has discovered that TextEncoder.encode and TextDecoder.decode aren't inverses for each other =(

rth (approved):

PR #2678, opened 2022-06-08
FIX TextDecoder is not a left inverse for TextEncoder

hoodmane:
Strings on which decode(encode(s)) !== s will fail these tests
due to no fault of Pyodide. Took Hypothesis a long time to find this
though.

Counter example: '\ufeff'
https://app.circleci.com/pipelines/github/hoodmane/pyodide/3238/workflows/ca2c0d13-db82-43b3-b8e4-d4aad7c1cd3b/jobs/37480

### DIFF
--- a/src/core/jsproxy.c
+++ b/src/core/jsproxy.c
@@ -1678,7 +1678,7 @@ JsBuffer_DecodeString_js,
   if (encoding) {
     encoding_js = UTF8ToString(encoding);
   }
-  let decoder = new TextDecoder(encoding_js, {fatal : true});
+  let decoder = new TextDecoder(encoding_js, {fatal : true, ignoreBOM: true});
   let res;
   try {
     res = decoder.decode(buffer);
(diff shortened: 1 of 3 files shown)


hoodmane:
I added `ignoreBOM` in two places in the code too because I think it's the more expected behavior (it's good for decoding to be injective when possible).

ryanking13 (approved):

ryanking13:
Thanks, @hoodmane!

rth:
Shouldn't it be `ignoreBOM = false` though, if we want to decode exactly the same string (with BOM)?

hoodmane:
Apparently not -- try it.

rth:
Ahh, yes, indeed, 

> [if True]
> The parser will ignore that BOM mark, meaning it will treat it as a normal character sequence. 

that's rather confusing.

hoodmane:
Yeah it's confusing.

PR #2682, opened 2022-06-09
MAINT Fix pyproxy_init declaration

hoodmane:


hoodmane:
We could remove these init declarations because they are declared automatically by `TRY_INIT` and `TRY_INIT_WITH_CORE_MODULE`

PR #2683, opened 2022-06-10
Fix cryptography build after emscripten update

hoodmane:


ryanking13 (approved):
Thanks, @hoodmane. I think it would be great if we can merge `rust_emcc_wrapper.py` to `pywasmcross.py` (no action required for this PR).

rth:
Thanks! +1 to add this wrapper to pyodide build if possible in the future.

PR #2684, opened 2022-06-10
Update Emscripten patches again

hoodmane:
This drops the Rust patch and "Throw away errors in minify wasm_js" which I suspect was fixed by an emscripten update (I'm not sure how long ago). It replaces the two remaining emscripten patches with new versions that match the upstream PRs.


ryanking13 (approved):
Thanks, @hoodmane! LGTM, but I think we can wait until your upstream PR (https://github.com/emscripten-core/emscripten/pull/17184) gets merged. 

It seems like the remaining two emscripten patches can also be removed when we update the emscripten version later. Probably we can simplify our build setups a lot when all emscripten patches are removed, though often we add new patches...

hoodmane:
Yeah I think we should leave the code in to apply the patches but just leave the patch directory empty. We will probably have more in the future. But it is pretty exciting to be getting rid of them =)

PR #2637, opened 2022-05-28
Link with suitesparse

gabrielfougeron:
Hi,

The Pyodide team helped me a few month ago to setup a new package : suitesparse (PR #2273)
After a few iterations, everything was working fine.

I tried to update to the current state of Pyodide (now running Python 3.10 and not 3.9), and the code compiles just fine (!), but the following error occurs when trying to load the package in the REPL:

Failed to load suitesparse
pyodide.asm.js:10 The following error occurred while loading suitesparse:
pyodide.asm.js:10 RuntimeError: abort(Assertion failed: need the dylink section to be first). Build with -s ASSERTIONS=1 for more info.
    at abort (pyodide.asm.js:10:234403)
    at assert (pyodide.asm.js:10:224857)
    at getDylinkMetadata (pyodide.asm.js:10:309965)
    at Object.loadWebAssemblyModule (pyodide.asm.js:10:314162)
    at loadDynlib (pyodide.asm.js:10:195833)
    at async installPackage (pyodide.asm.js:10:195212)
    at async pyodide.asm.js:10:198223
    at async Promise.all (index 1)
    at async loadPackage (pyodide.asm.js:10:198363)
    at async Object.loadPackagesFromImports (pyodide.asm.js:10:213298)

After a Google search on the error, it seems to me that the problem might happen at link time (see https://github.com/WebAssembly/binaryen/issues/1582 ). I was not able to go further on my own.

Thanks for the help,

Best regards,

PR #2685, opened 2022-06-11
PKG Add sparseqr and suitesparse

ryanking13:
### Description

Add sparseqr and suitesparse, continuing from #2637.

This is on top of #2685

### Checklists

- [x] Add / update tests


ryanking13:
@gabrielfougeron Sorry I think I did something wrong in your PR and lost my permission to push, so I opened a new one.

I removed patches that seem not required anymore. If I mistakenly removed some patches that are actually needed. Please tell me. Could you tell us what the `Changed a few files` patch in `suitesparse` does? It's quite long and its name is ambiguous. We want to minimize the patch if possible.

Also, I removed `choreo` for now as you mentioned that it's not ready to be released. Feel free open a PR that adds `choreo` in a follow up PR.

packages/sparseqr/meta.yaml
```diff
@@ -0,0 +1,29 @@
...
+    - numpy
+    - scipy
+    - suitesparse
+    - distutils
+
+build:
+  script: |
+    # FIXME: We are not dealing with poetry build system correctly.
+    #        So for now remove `pyproject.toml` to force pypabuild to use setup.py
+    rm -f pyproject.toml
```
ryanking13, line 23: I don't understand this actually. We are using `pypa/build` so I thought that it would correctly handle packages that build with poetry. But the resulting `.whl` file was empty.

gabrielfougeron:
Indeed, this needs explaining! 
The poorly named `Changed a few files` removes anything in `suitesparse` that is not needed for `sparseqr`. This includes anything GPU related for instance, but also other sparse linear algebra libraries that I struggled to build on my PC, and that would not be used by sparseqr anyway.

gabrielfougeron:
It seems to me that you managed to have it work! Amazing :-) !
Congratulations and many thanks!

Is there anything I need to do?

ryanking13:
> It seems to me that you managed to have it work! Amazing :-) ! Congratulations and many thanks!
> 
> Is there anything I need to do?

I would like to reduce the size of the patch file. But I think I can handle that :)

I think this PR will take a couple of days to merge, as #2686 should be reviewed and merged first. Until then, it would be great if you can check if `sparseqr` is working correctly.
You can download the Pyodide build artifacts from the CI:

https://output.circle-artifacts.com/output/job/1eac3d01-414d-4c31-8ee1-1b57611c7c3e/artifacts/0/root/repo/pyodide-build.tar.gz

gabrielfougeron:
I've tested the files you sent me in the pyodide console and as far as I can tell, it is working perfectly.
Thanks very much !

ryanking13:
Alright, thanks for the experiments @gabrielfougeron. #2685 is merged so this is ready to be reviewed now.

ryanking13:
@hoodmane @rth could one of you review this?

packages/suitesparse/meta.yaml
```diff
@@ -0,0 +1,41 @@
...
+
+    # Fix return type to avoid linker error
+    grep -rl "void LAPACK" . | xargs sed -i 's/void LAPACK/int LAPACK/g'
+    grep -rl "void BLAS" . | xargs sed -i 's/void BLAS/int BLAS/g'
+
+    emmake make -j ${PYODIDE_JOBS:-3} install \
+      LDFLAGS="-sSIDE_MODULE=1 -L${WASM_LIBRARY_DIR}/lib " \
+      BLAS="$PYODIDE_ROOT/packages/CLAPACK/dist/clapack_all.so" \
+      LAPACK="$PYODIDE_ROOT/packages/CLAPACK/dist/clapack_all.so" \
+      INSTALL=${WASM_LIBRARY_DIR}
```
hoodmane, line 30: Is there a reason you install these into `WASM_LIBRARY_DIR`? Isn't that for static libraries?

packages/suitesparse/patches/0001-Remove-non-essential-build-targets.patch
```diff
@@ -0,0 +1,102 @@
...
+diff --git a/metis-5.1.0/CMakeLists.txt b/metis-5.1.0/CMakeLists.txt
+index 1a72ed0d..55c6313a 100644
+--- a/metis-5.1.0/CMakeLists.txt
++++ b/metis-5.1.0/CMakeLists.txt
+@@ -25,4 +25,4 @@ include_directories(include)
+ # Recursively look for CMakeLists.txt in subdirs.
+ add_subdirectory("include")
+ add_subdirectory("libmetis")
+-add_subdirectory("programs")
++#add_subdirectory("programs")
```
hoodmane, line 99: Delete lines rather than commenting them.

hoodmane (approved):
Suitesparse could use a better system to configure what subprojects to include. It's probably more effort than it's worth to try to contribute one.

packages/suitesparse/meta.yaml
ryanking13: No, it is for both static and shared libraries. For example, `cryptography` uses `pkg-config` to find openssl headers and object files, which requires `openssl` to be installed in a `WASM_LIBRARY_DIR`.

But yes, it is quite annoying that we have to manually copy `.so` files into the `dist` folder inside the source directory again after the build. Maybe we should find a way to automate that stuff...

ryanking13:
Thanks for the review!

PR #2686, opened 2022-06-11
Fix shared library loading order

ryanking13:
### Description

Currently, we load shared libraries in an archive in random order. More precisely, we load them in the same order as their entries in the actual ZIP file. However, some libraries like `suitesparse` contain multiple `.so` files that depend on each other. So we have to load them in a precise order.

This change makes shared library load order deterministic by sorting the path in a zipfile. I don't actually like this solution very much as one who writes a meta.yaml file needs to [know how `.so` files depend to each other](https://github.com/pyodide/pyodide/blob/11786c3f09481c82a2ce0e691aed02e3c8e82117/packages/suitesparse/meta.yaml#L33-L46). But works for a temporary solution.


hoodmane:
I had a potential fix for this a while back but @rth complained I had no examples. Would you mind making a simple test case for this in a branch?

ryanking13:
> Would you mind making a simple test case for this in a branch?

I updated `shared-library-test` package to test this case (thanks for adding this package BTW). Note that this test case is not very robust as if we update CPython version later, the algorithm that CPython zipfile module optimizes the file order in a zip archive may change.

> I had a potential fix for this a while back

Sounds great! I wasn't able to think of an easy fix for this. Apparently, my fix is pretty bad... so it would be great to have a more proper fix.

hoodmane:
I think the correct fix is to pass an fs argument to `loadDynamicLibrary`
https://github.com/emscripten-core/emscripten/blob/b1dbe135e722d8e25e769472d397b9171e5cd9eb/src/library_dylink.js#L728
here:
https://github.com/pyodide/pyodide/blob/dc33b25a6f3e62a9844fb2caa1480b726b56b809/src/js/load-package.ts#L294
then `loadDynamicLibrary` can look up what other libraries are needed in the `dylink` section of the wasm module and load those first if they aren't already loaded.

ryanking13:
Alright I'll try that.

ryanking13:
Okay it looks like adding `flags.fs` argument works correctly (applied also in #2685).

src/js/load-package.ts
```diff
@@ -282,11 +282,24 @@ async function loadDynlib(lib: string, shared: boolean) {
     byteArray = Module.FS.readFile(lib);
   }
   const releaseDynlibLock = await acquireDynlibLock();
+
+  // This is a fake FS-like object to make emscripten
+  // load shared libraries from the file system.
+  const libraryFS = {
+    _rootPath: lib.split("/").slice(0, -1).join("/"),
```
hoodmane, line 289: Could you add a comment here explaining why this is the root path?

ryanking13: Okay, I added a comment and renamed `_rootPath` to `_ldLibraryPath`, which seems more intuitive.
We are currently installing shared libraries into site-packages directory, probably in a subsequent PR, we should change it to install all shared libraries into `/usr/local/lib` or `/usr/lib`.

hoodmane: Sounds good.

src/js/load-package.ts
```diff
@@ -297,7 +312,7 @@ async function loadDynlib(lib: string, shared: boolean) {
       });
     }
   } catch (e: any) {
-    if (e.message.includes("need to see wasm magic number")) {
+    if (e && e.message && e.message.includes("need to see wasm magic number")) {
       console.warn(
         `Failed to load dynlib ${lib}. We probably just tried to load a linux .so file or something.`
```
hoodmane, line 317: Not related to this PR: another possibility that we might want to mention in this error message is that `-sSIDE_MODULE` went missing and the file `blah.so` actually contains JavaScript.

ryanking13: Yes, indeed. It is a pretty common reason of the failure.

hoodmane (approved):

ryanking13:
Thanks for the review!

PR #2690, opened 2022-06-11
Add support for backend flags to buildpkg

hoodmane:
And use it to disable numpy optimization by passing the `--disable-optimization` flag instead of patching numpy.


ryanking13:
Thanks @hoodmane. It looks like the test failure in `playwright` is related to the recent hypothesis update. Probably we should pin the hypothesis version of wait until their fix (https://github.com/HypothesisWorks/hypothesis/commit/dcb41212837c98ecf5759d9cbb15c95fe7453241) is released.

ryanking13 (approved):
Aside from the hypothesis issue, codewise LGTM. Please update meta.yaml specification docs.

PR #2692, opened 2022-06-11
Update fenv patch

hoodmane:
This isn't different in effect than the old patch but it can probably be upstreamed. Also, this new patch has the benefit that if wasm/emscripten ever does gain fenv support we will switch to using it automatically. 
Upstream PR: https://github.com/numpy/numpy/pull/21732


ryanking13 (approved):
Thanks, @hoodmane! LGTM and it looks like the patch is going to be accepted to the upstream soon.

hoodmane:
Okay patch was accepted with some changes to comments.

PR #2695, opened 2022-06-11
Update patch for numpy to determine pointer sizes correctly for cross build

hoodmane:
Upstream PR: https://github.com/numpy/numpy/pull/21733/files


rth (approved):
Let's wait for the fix to be merged in numpy, otherwise LGTM. Thanks!

hoodmane:
Upstream patch was accepted: https://github.com/numpy/numpy/pull/21735
It is slightly different in order to ensure windows support, but I don't think the difference matters for us. I think we should just merge with this patch and when we update to the not-yet-released next version of numpy we can drop it.

ryanking13 (approved):
> Upstream patch was accepted. It is slightly different in order to ensure windows support, but the difference doesn't matter for us. I think we should just merge with this patch and when we update to the not-yet-released next version of numpy we can drop it.

Yes, that sounds OK to me. Thanks for the maintenance work!

Issue #227, opened 2018-10-24
Using a high performance BLAS


rth:
Performance of numerical calculations is going to be significantly impacted by the BLAS used. 

https://github.com/iodide-project/pyodide/pull/211 aims to include CLAPACK with the reference BLAS which should have a baseline performance.

Once that is done (hopefully soon), e.g. [BLIS](https://github.com/flame/blis) could be a possible candidates for a better BLAS. That is probably more achievable than OpenBLAS. BLIS was already previously compiled for emscritpen in https://github.com/Maratyszcza/blis-bench (though targeting Asm.js). The relevant discussion at numpy is https://github.com/numpy/numpy/issues/7372.

rth:
Some benchmarks for matrix multiplication using the currently included reference BLAS can be found in https://gist.github.com/rth/c71fe792eb56fb271317e35e08576c7a

rth:
Although as I mentioned to @lesteve, BLIS only includes BLAS, so we would still have to use reference CLAPACK with all the existing hacks. While it we manage to have a generic build of OpenBLAS it would also have optimized LAPACK. Though it's in Fortran, so lfortran is probably our best bet there.

PR #2696, opened 2022-06-11
Drop Numpy use local blas lapack patch

hoodmane:
In either case, Numpy is build without blas/lapack support.
Perhaps we should consider loading blas/lapack for numpy?
It's a tradeoff between speed and load time I imagine.


rth (approved):
> Perhaps we should consider loading blas/lapack for numpy?

We could, but I'm not sure it would be worth. Performance wise it's going to be the same, until we start using optimized BLAS https://github.com/pyodide/pyodide/issues/227 while size wise CLAPACK is significantly worse,
```
less numpy-1.22.3-cp310-cp310-emscripten_3_1_13_wasm32.whl | grep lapack
 1961638  Defl:N   423621  78% 2022-06-11 22:54 d86386e5  numpy/linalg/lapack_lite.cpython-310-wasm32-emscripten.so
```
(size 0.4MB)
```
less CLAPACK-3.2.1.zip | cat
Archive:  CLAPACK-3.2.1.zip
 Length   Method    Size  Cmpr    Date    Time   CRC-32   Name
--------  ------  ------- ---- ---------- ----- --------  ----
       0  Stored        0   0% 2022-06-11 23:00 00000000  lib/
 3753715  Defl:N  1253413  67% 2022-06-11 23:00 025bf03f  lib/clapack_all.so
--------          -------  ---                            -------
 3753715          1253413  67%                            2 files
```
(size 12MB)

Actually, I get the numpy only need a subset of LAPACK  but still a more than 20x difference in size looks suspicous.

PR #2704, opened 2022-06-13
Add Rust tooling to Docker image

josephrocca:
Please feel free to close this if I've done it wrong! I was going to open an issue, but I figured it's simple enough to just make a pull request.

I'm assuming that it's a good idea to add the Rust tooling now that there's preliminary support for Rust (#1973), otherwise (unless I'm misunderstanding) every `meta.yaml` that requires Rust would need to do the install in `build.script`.

Again, please unceremoniously close this if my assumptions are incorrect here.

**Edit:** Oh, I guess we'd also want `rustup target add wasm32-unknown-emscripten`? (I've added a commit for this)


Dockerfile
```diff
@@ -8,9 +8,13 @@ RUN apt-get update \
         patch pkg-config swig unzip wget xz-utils \
         autoconf autotools-dev automake texinfo dejagnu \
         build-essential prelink autoconf libtool libltdl-dev \
-        gnupg2 libdbus-glib-1-2 sudo \
+        gnupg2 libdbus-glib-1-2 sudo curl \
   && rm -rf /var/lib/apt/lists/*
 
+RUN curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh -s -- -y
+RUN source $HOME/.cargo/env
+RUN rustup target add wasm32-unknown-emscripten
```
rth, line 16: So normally you should be able to install it with [`make rust`](https://github.com/pyodide/pyodide/blob/main/Makefile#L242). Also it's better to group commands to avoid creating too many Docker layers (and `make rust` would take care of it).

rth (approved):
Thanks @josephrocca !

hoodmane:
Thanks @josephrocca! Much appreciated. We should probably update the docker image in sync with this change...

Dockerfile
josephrocca: Thanks! Would this mean that I'd need to add `RUN git clone <pyodide repo url>` first so that `make` can access Pyodide's `Makefile`? And if so, where should the Pyodide repo be cloned to?

Oh unless you mean that we don't actually need this in the Dockerfile because users are expected to run `make` after running `./run_docker --pre-built`? ([related issue](https://github.com/pyodide/pyodide/issues/2705)) In that case, can the `rust` target be added as a default so that it's installed when people run `make`?

rth: Aww, you are right we don't have the source tree here. OK this is fine then.

rth, line 16: Let's merge these layers,

```suggestion
RUN curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh -s -- -y &&
    source $HOME/.cargo/env &&
    rustup target add wasm32-unknown-emscripten
```

and it would be good it you could build and push this image to your temporary account, then use it in `.circleci/config.yaml` and make sure that removing `make rust` there works. 
Then once CI passes, revert the changes to `.circleci/config.yaml`. Once merged we will release the new docker image via Github actions.

josephrocca: @rth Okay I've done this, but it looks like some of the tests have failed? Not sure if those ones usually pass.

hoodmane:
Worth noting that this is a 75% increase in the image size.

hoodmane:
Worth noting that this is a 75% increase in the image size.

ryanking13:
#3443 finished a remaining works in this PR. Thanks again @josephrocca!

Issue #2706, opened 2022-06-13
Passing custom flags to `cargo`


josephrocca:
I'm currently trying to get [`tokenizers`](https://github.com/huggingface/tokenizers/) working in Pyodide, and this requires using the `unstable_wasm` feature. If I understand correctly, this means that I need to pass `--features unstable_wasm` to cargo, which would require something like `built.cargoflags` in `meta.yaml`? Or is there some other way to achieve this that could suffice for now?

josephrocca:
Also, it's possible I'm making some sort of mistake here, but it seems like "default" flags are being passed to rustc during `build.script`? I was under the impression that `build.script` runs in a context that doesn't yet have all the environment/binary swapping? I need a place in `meta.yaml` where I can run some pre-building stuff that doesn't have any edits to its environment.

ryanking13:
> this means that I need to pass `--features unstable_wasm` to cargo, which would require something like `built.cargoflags` in `meta.yaml`

I think you can start with patching (adding `unstable_wasm` in `Cargo.toml`), which, I think, is cheaper for now.

> but it seems like "default" wasm-related flags are being passed to rustc during `build.script`

We have some rust-related flags defined in `Makefile.envs`. I don't know much about rust but I guess cargo uses them If they are defined in environment variables?

https://github.com/pyodide/pyodide/blob/26c3f4dbd4b0a9472a65a31fbd4b51b4c04e5362/Makefile.envs#L148-L152

> I was under the impression that `build.script` runs in a context that doesn't yet have all the environment/binary swapping? I need a place in `meta.yaml` where I can run some pre-building stuff that doesn't have any edits to its environment.

I'm not sure I understood your question correctly. Environment variables defined in `pyodide_build/buildpkg.py` are accessible in `build.script` (you can add one if needed).

https://github.com/pyodide/pyodide/blob/26c3f4dbd4b0a9472a65a31fbd4b51b4c04e5362/pyodide-build/pyodide_build/buildpkg.py#L126-L158

Also, if you use `export SOME_ENV=...` inside `build.script`. That variable will be accessible during that package build.

hoodmane:
See
https://github.com/PyO3/setuptools-rust/pull/243

hoodmane:
>  it seems like "default" wasm-related flags are being passed to rustc during build.script? 

You can avoid this via:

```yaml
build:
  script: |
    export RUSTFLAGS=""
```

josephrocca:
[...]

PR #2709, opened 2022-06-14
Mention that some RUSTFLAGS are set by default in built.script

josephrocca:
Per @hoodmane's comment here: https://github.com/pyodide/pyodide/issues/2706#issuecomment-1154714227


hoodmane:
Thanks @josephrocca!

PR #2712, opened 2022-06-15
Experiment: Default to -sSIDE_MODULE=2 in packages

hoodmane:


ryanking13:
We set `EXPORT_ALL` in our LDFLAGS, does setting `SIDE_MODULE=2` has effect in this situation?

https://github.com/pyodide/pyodide/blob/392908cf68d47247a21b815a8c23100aab0d8c50/Makefile.envs#L88

hoodmane:
Well `-sEXPORT_ALL` and `-sSIDE_MODULE=2` do not together behave the same as `-sSIDE_MODULE=1`. `-sSIDE_MODULE=1` hits a problem when compiling rust because Emscripten adds `--whole-archive` around everything when linking but the rust archives have the `lib.rmeta` file which can't be used because it isn't an object file.

But maybe we should also move `-sEXPORT_ALL` into the main module ldflags. We may need it for shared libraries but we definitely don't need it for packages.

hoodmane:
So we would need to tell the modules to export functions of the form `PyInit_*`.

ryanking13:
> Well `-sEXPORT_ALL` and `-sSIDE_MODULE=2` do not together behave the same as `-sSIDE_MODULE=1`. `-sSIDE_MODULE=1` hits a problem when compiling rust because Emscripten adds `--whole-archive` around everything when linking but the rust archives have the `lib.rmeta` file which can't be used because it isn't an object file.

Thanks for the explanation. I can't find [enough documentation](https://emscripten.org/docs/compiling/Dynamic-Linking.html?highlight=dlopen#code-size) about SIDE_MODULE=2 behavior... I'll have to look into emscripten codes.

> But maybe we should also move -sEXPORT_ALL into the main module ldflags. We may need it for shared libraries but I am pretty sure we don't need it for packages.

Yeah, I think it would be really great if we can make it work. I guess it will help us reduce code size a lot.

hoodmane:
This PR drops the size of the dist directory from 271mb to 226 mb which is a 17% size reduction.

ryanking13:
> This PR drops the size of the dist directory from 271mb to 226 mb which is a 17% size reduction.

Wow, thats very huge! Thanks @hoodmane. This is the file size list compared to main branch.

```
*** 5,7 ****
! 595K PIL-9.1.0-cp310-cp310-emscripten_3_1_13_wasm32.whl
! 949K PyWavelets-1.3.0-cp310-cp310-emscripten_3_1_13_wasm32.whl
! 114K PyYAML-6.0-cp310-cp310-emscripten_3_1_13_wasm32.whl
--- 5,7 ----
! 1.1M PIL-9.1.0-cp310-cp310-emscripten_3_1_13_wasm32.whl
! 1.1M PyWavelets-1.3.0-cp310-cp310-emscripten_3_1_13_wasm32.whl
! 119K PyYAML-6.0-cp310-cp310-emscripten_3_1_13_wasm32.whl
***************
*** 10 ****
! 11M SciPy-1.8.0-cp310-cp310-emscripten_3_1_13_wasm32.whl
--- 10 ----
! 12M SciPy-1.8.0-cp310-cp310-emscripten_3_1_13_wasm32.whl
***************
*** 27 ****
! 18K cffi_example-0.1-cp310-cp310-emscripten_3_1_13_wasm32.whl
--- 27 ----
! 288K cffi_example-0.1-cp310-cp310-emscripten_3_1_13_wasm32.whl
***************
*** 35 ****
! 1.3M cryptography-36.0.2-cp310-cp310-emscripten_3_1_13_wasm32.whl
--- 35 ----
! 1.4M cryptography-36.0.2-cp310-cp310-emscripten_3_1_13_wasm32.whl
***************
*** 48 ****
! 200K freesasa-2.1.0-cp310-cp310-emscripten_3_1_13_wasm32.whl
--- 48 ----
! 206K freesasa-2.1.0-cp310-cp310-emscripten_3_1_13_wasm32.whl
***************
*** 51 ****
! 709K galpy-1.7.2-cp310-cp310-emscripten_3_1_13_wasm32.whl
--- 51 ----
! 1.4M galpy-1.7.2-cp310-cp310-emscripten_3_1_13_wasm32.whl
***************
*** 53,54 ****
! 402K gmpy2-2.1.2-cp310-cp310-emscripten_3_1_13_wasm32.whl
! 1.7M gsw-3.4.0-cp310-cp310-emscripten_3_1_13_wasm32.whl
--- 53,54 ----
! 464K gmpy2-2.1.2-cp310-cp310-emscripten_3_1_13_wasm32.whl
! 1.8M gsw-3.4.0-cp310-cp310-emscripten_3_1_13_wasm32.whl
***************
*** 67 ****
! 1.5M lxml-4.8.0-cp310-cp310-emscripten_3_1_13_wasm32.whl
--- 67 ----
! 1.9M lxml-4.8.0-cp310-cp310-emscripten_3_1_13_wasm32.whl
***************
*** 79 ****
! 272K msprime-1.1.1-cp310-cp310-emscripten_3_1_13_wasm32.whl
--- 79 ----
! 979K msprime-1.1.1-cp310-cp310-emscripten_3_1_13_wasm32.whl
***************
*** 83 ****
! 169K nlopt-2.7.0-cp310-cp310-emscripten_3_1_13_wasm32.whl
--- 83 ----
! 180K nlopt-2.7.0-cp310-cp310-emscripten_3_1_13_wasm32.whl
***************
*** 87 ****
! 410K numcodecs-0.9.1-cp310-cp310-emscripten_3_1_13_wasm32.whl
--- 87 ----
! 552K numcodecs-0.9.1-cp310-cp310-emscripten_3_1_13_wasm32.whl
***************
*** 89 ****
! 3.0M numpy-1.22.3-cp310-cp310-emscripten_3_1_13_wasm32.whl
--- 89 ----
! 3.6M numpy-1.22.3-cp310-cp310-emscripten_3_1_13_wasm32.whl
***************
*** 91 ****
! 10M opencv_python-4.5.5.64-cp310-cp310-emscripten_3_1_13_wasm32.whl
--- 91 ----
! 11M opencv_python-4.5.5.64-cp310-cp310-emscripten_3_1_13_wasm32.whl
***************
*** 97 ****
! 4.3M pandas-1.4.2-cp310-cp310-emscripten_3_1_13_wasm32.whl
--- 97 ----
! 4.4M pandas-1.4.2-cp310-cp310-emscripten_3_1_13_wasm32.whl
***************
*** 107 ****
! 240K pyerfa-2.0.0.1-cp310-cp310-emscripten_3_1_13_wasm32.whl
--- 107 ----
! 239K pyerfa-2.0.0.1-cp310-cp310-emscripten_3_1_13_wasm32.whl
***************
*** 122,123 ****
! 469K python_sat-0.1.7.dev18-cp310-cp310-emscripten_3_1_13_wasm32.whl
! 146K python_solvespace-3.0.6-cp310-cp310-emscripten_3_1_13_wasm32.whl
--- 122,123 ----
! 529K python_sat-0.1.7.dev18-cp310-cp310-emscripten_3_1_13_wasm32.whl
! 174K python_solvespace-3.0.6-cp310-cp310-emscripten_3_1_13_wasm32.whl
***************
*** 126 ****
! 194K regex-2022.3.15-cp310-cp310-emscripten_3_1_13_wasm32.whl
--- 126 ----
! 195K regex-2022.3.15-cp310-cp310-emscripten_3_1_13_wasm32.whl
***************
*** 145 ****
! 3.7M statsmodels-0.13.2-cp310-cp310-emscripten_3_1_13_wasm32.whl
--- 145 ----
! 3.9M statsmodels-0.13.2-cp310-cp310-emscripten_3_1_13_wasm32.whl
***************
*** 149 ****
! 348K swiglpk-5.0.2-cp310-cp310-emscripten_3_1_13_wasm32.whl
--- 149 ----
! 370K swiglpk-5.0.2-cp310-cp310-emscripten_3_1_13_wasm32.whl
***************
*** 164 ****
! 285K tskit-0.4.1-cp310-cp310-emscripten_3_1_13_wasm32.whl
--- 164 ----
! 290K tskit-0.4.1-cp310-cp310-emscripten_3_1_13_wasm32.wh
```

pyodide-build/pyodide_build/pywasmcross.py
```diff
@@ -368,6 +371,23 @@ def replay_genargs_handle_argument(arg: str) -> str | None:
     return arg
 
 
+def calculate_exports_flag(line: list[str], export_all: bool) -> str:
+    objects = [arg for arg in line if arg.endswith(".a") or arg.endswith(".o")]
+    PYODIDE_ROOT = common.get_pyodide_root()
+    result = subprocess.run(
+        [PYODIDE_ROOT / "emsdk/emsdk/upstream/bin/llvm-nm", "-j", "--export-symbols"]
```
ryanking13, line 378: I think we can use `emnm`. It is just a wrapper of `llvm-nm`, and it is in our `PATH` like `emcc`, etc.

https://github.com/emscripten-core/emscripten/blob/main/tools/emnm.py

pyodide-build/pyodide_build/buildpkg.py
```diff
@@ -449,6 +449,7 @@ def compile(
             ldflags=build_metadata["ldflags"],
             target_install_dir=target_install_dir,
             replace_libs=replace_libs,
+            exports=build_metadata.get("exports", "init_functions"),
```
ryanking13, line 452: What is `init_functions`?

pyodide-build/pyodide_build/io.py
```diff
@@ -20,6 +20,7 @@
         "extras": list,  # List[Tuple[str, str]],
     },
     "build": {
+        "exports": str,
```
ryanking13, line 23: Please add to docs about this flag

ryanking13 (approved):
Thanks very much. I have some comments, otherwise LGTM.

hoodmane:
Thanks for the review @ryanking13! I will add some explanation soon.

As a followup, I think we could list explicitly the symbols xgboost uses via ctypes. With:
```sh
rg '_LIB\.[A-Za-z0-9_]*' -g'*.py' -o -I -N | sort -u | sed 's/^_LIB.\(.*\)$/    -  \1/g'
```
I find that it's

<details>
<summary>xgboost symbols</summary>

```yaml
    -  RabitAllreduce
    -  RabitBroadcast
    -  RabitFinalize
    -  RabitGetProcessorName
    -  RabitGetRank
    -  RabitGetWorldSize
    -  RabitInit
    -  RabitIsDistributed
    -  RabitTrackerPrint
    -  RabitVersionNumber
    -  XGBGetGlobalConfig
    -  XGBGetLastError
    -  XGBoosterBoostedRounds
    -  XGBoosterBoostOneIter
    -  XGBoosterCreate
    -  XGBoosterDumpModelEx
    -  XGBoosterEvalOneIter
    -  XGBoosterFeatureScore
    -  XGBoosterFree
    -  XGBoosterGetAttr
    -  XGBoosterGetAttrNames
    -  XGBoosterGetNumFeature
    -  XGBoosterGetStrFeatureInfo
    -  XGBoosterLoadJsonConfig
    -  XGBoosterLoadModel
    -  XGBoosterLoadModelFromBuffer
    -  XGBoosterPredictFromCSR
    -  XGBoosterPredictFromCudaArray
    -  XGBoosterPredictFromCudaColumnar
    -  XGBoosterPredictFromDense
    -  XGBoosterPredictFromDMatrix
    -  XGBoosterSaveJsonConfig
    -  XGBoosterSaveModel
    -  XGBoosterSaveModelToBuffer
    -  XGBoosterSerializeToBuffer
    -  XGBoosterSetAttr
    -  XGBoosterSetParam
    -  XGBoosterSetStrFeatureInfo
    -  XGBoosterSlice
    -  XGBoosterUnserializeFromBuffer
    -  XGBoosterUpdateOneIter
    -  XGBSetGlobalConfig
    -  XGBuildInfo
    -  XGDeviceQuantileDMatrixCreateFromCallback
    -  XGDMatrixCreateFromArrowCallback
    -  XGDMatrixCreateFromCallback
    -  XGDMatrixCreateFromCSCEx
    -  XGDMatrixCreateFromCSR
    -  XGDMatrixCreateFromCudaArrayInterface
    -  XGDMatrixCreateFromCudaColumnar
    -  XGDMatrixCreateFromDense
    -  XGDMatrixCreateFromDT
    -  XGDMatrixCreateFromFile
    -  XGDMatrixFree
    -  XGDMatrixGetFloatInfo
    -  XGDMatrixGetStrFeatureInfo
    -  XGDMatrixGetUIntInfo
    -  XGDMatrixNumCol
    -  XGDMatrixNumRow
    -  XGDMatrixSaveBinary
    -  XGDMatrixSetInfoFromInterface
    -  XGDMatrixSetStrFeatureInfo
    -  XGDMatrixSliceDMatrixEx
    -  XGImportArrowRecordBatch
    -  XGProxyDMatrixCreate
    -  XGProxyDMatrixSetDataCSR
    -  XGProxyDMatrixSetDataCudaArrayInterface
    -  XGProxyDMatrixSetDataCudaColumnar
    -  XGProxyDMatrixSetDataDense
```

</details>

pyodide-build/pyodide_build/pywasmcross.py
hoodmane: Good find, I hadn't noticed that one.

rth:
This looks like a very nice size improvement thanks! 

>  `pyinit`: The default. Only export Python module initialization symbols of
>  the form `PyInit_some_module`.

Hmm, so I don't fully understand what symbols are required to use Python .so files, but I wonder why people are not doing it on x86_64 to reduce the size of wheels. For instance, auditwheel has an option to strip symbols https://github.com/pypa/auditwheel/issues/63 and from what I understand numpy/scipy are using it. However if I do,
```
nm -gD site-packages/numpy/core/_multiarray_umath.cpython-39-x86_64-linux-gnu.so
```
on numpy installed from a manylinux wheel, there are still lots of `Py*` and `npy_*` symbols there, not just `PyInit__multiarray_umath`.

Maybe @ogrisel or @antocuni would have some perspective on this.

pyodide-build/pyodide_build/io.py
```diff
@@ -126,6 +127,13 @@ def _check_config_build(config: dict[str, Any]) -> Iterator[str]:
     build_metadata = config["build"]
     library = build_metadata.get("library", False)
     sharedlibrary = build_metadata.get("sharedlibrary", False)
+    exports = build_metadata.get("exports", "pyinit")
+    if isinstance(exports, str) and exports not in [
+        "pyinit",
+        "requested",
+        "whole_archive",
+    ]:
+        yield f"build/exports must be 'pyinit', 'explicit', 'all', or a list of strings not {build_metadata['exports']}"
```
ryanking13, line 136: This doesn't seem to match with docs?

hoodmane: Yeah I left the list of strings case undocumented since we aren't actually using it anywhere currently. We might want to take it out, or when we first use it we can document it. It seemed to work okay when I tried it but since the CI doesn't use it it's hard to know if it really works. I thought about explicitly listing the exports for xgboost but there are quite a lot.

hoodmane: Oh I see what you mean...

hoodmane: Oh I see what you mean, the error message is a bit wrong.

hoodmane: #2777 fixes it.

antocuni:
> This looks like a very nice size improvement thanks!
> 
> > `pyinit`: The default. Only export Python module initialization symbols of
> > the form `PyInit_some_module`.
> 
> Hmm, so I don't fully understand what symbols are required to use Python .so files, 

`PyInit_xxx` is the only symbol which is needed by CPython itself to load a `.so` module.

However, there are some packages which rely on exporting *more* symbols, but this is a per-package requirement, not a CPython one. I suppose that the most common use case is to export symbols and then load/call them using `ctypes` or `cffi`: this is possible because on linux when you load a `.so` all the exported symbols becomes immediately available to the whole process.
If these symbols are exported-but-not-called, a naive dead-code elimination would remove the whole functions from the `.so`: then, the `.so` would be correctly loaded by CPython (because it only needs `PyInit_xxx` to do so), but later the package would fail at runtime when it tries to load the symbols with `ctypes` (because they have been eliminated).
(I think that on Windows it's a bit different, but I'm not very experienced with it).

if I understand correctly, this is what e.g. `xdgboost` is doing, and the options introduced by this PR look very reasonable to me.

> but I wonder why people are not doing it on x86_64 to reduce the size of wheels. For instance, auditwheel has an option to strip symbols [pypa/auditwheel#63](https://github.com/pypa/auditwheel/issues/63) and from what I understand numpy/scipy are using it. However if I do,

note that stripping symbols is a different thing: when you strip, you keep the code in the `.so` and you just remove the mnemonic function names associated to the assembly code. This can be done for all internal functions which are *not exported*, even if they are called by other parts of the code (because in assembly, they will be called by address, not by name). The main visible effect is that you don't see function names in gdb stacktraces, and the space saving is small since you just remove the names.

If I understand correctly, `-sSIDE_MODULE=2` does a dead code elimination, which is completely different: it removes the WHOLE functions, but this can obviously be done only if the function is not called anywhere. Potentially, the space saving is very big (if you include a big library and call only one function) or very small (if you include a library and call *all* its functions).

> ```
> nm -gD site-packages/numpy/core/_multiarray_umath.cpython-39-x86_64-linux-gnu.so
> ```
> 
> on numpy installed from a manylinux wheel, there are still lots of `Py*` and `npy_*` symbols there, not just `PyInit__multiarray_umath`.

Before continuing, note that numpy is a bit of a special case because it defines also a C API for other C modules to call, but the vast majority of C extensions don't do that.
That said:
```
$ nm -gD x/lib/python3.9/site-packages/numpy/core/_multiarray_umath.cpython-39-x86_64-linux-gnu.so | grep Py
                 U PyArg_ParseTuple
                 U PyArg_ParseTupleAndKeywords
...                 
00000000005336e0 D PyArray_API
0000000000177e40 T PyArrayInitDTypeMeta_FromSpec
00000000001ddc40 T PyInit__multiarray_umath
0000000000534080 D PyUFunc_API
...
```

the vast majority of the `Py*` symbols are *imported*, i.e. they must be provided by someone else in order to load that .so (and the "someone else" in this particular case is `libpython.so`): `nm` shows them with the `U` flag, which means `Undefined` (precisely because they are defined somewhere else).
The non-U symbols are the ones actually defined in the `.so`. In particular:
  - `PyInit__multiarray_umath` is of type `T` (i.e., code) and it's needed to load the module
  - the other symbols are of type `D` (i.e. data) and are basically global variables. AFAIK `PyArray_API` and `PyUFUNC_API` are two big arrays of function pointers which are used to export the numpy C API
  - I don't know why they also need to export `PyArrayInitDTypeMeta_FromSpec` but I suspect is for the numpy C API as well.

I don't know why they also need to export the `npy_*` symbols, but I guess they have some good reason to do that.

rth:
Thanks for the detailed feedback!

PR #2713, opened 2022-06-15
Use -sSIDE_MODULE=2 when compiling rust

hoodmane:
instead of passing `-Wl,no-whole-archive`. This should reduce the size of the generated package (?)
and also doesn't require the compiler wrapper to insert the argument.


ryanking13:
We set `EXPORT_ALL` in our LDFLAGS, does setting `SIDE_MODULE=2` has effect in this situation?

https://github.com/pyodide/pyodide/blob/392908cf68d47247a21b815a8c23100aab0d8c50/Makefile.envs#L88

PR #2714, opened 2022-06-15
Install exceptiongroup in GHA

ryanking13:
Tests in GHA are failing due to hypothesis update. The hypothesis team added a fix https://github.com/HypothesisWorks/hypothesis/commit/1c153d8688c4bf014ff5172739d2e529ed6d3ef3 but we'll have to wait until their next release. This is a quick fix.


ryanking13:
Okay, this fix is wrong, it looks like we'll need to install exceptiongroup inside Pyodide?

hoodmane:
Yeah the problem is caused by the "all available types" hypothesis test. I think the issue is that some dependency is importing the exception group and so hypothesis notices it and pickles it and sends it to Pyodide which can't unpickle it.

PR #2721, opened 2022-06-16
Change shared library installation directory

ryanking13:
Change shared library installation directory from `/lib/python3.10/site-packages/` to `/usr/lib`.

#2686 and this PR improve how we load shared libraries. The future goal of these PRs is to load shared libraries on demand, instead of pre-loading all shared libraries, which will reduce initial package load time.


hoodmane:
> The future goal of these PRs is to load shared libraries on demand, instead of pre-loading all shared libraries

I still don't believe this is possible unless you mean once stack switching becomes available.

ryanking13:
> I still don't believe this is possible unless you mean once stack switching becomes available.

I don't know much about the blocker on this topic.

What I am expecting is that, when we call `loadWebAssemblyModule("moduleA_that_requires_moduleB.so")` it will try to find `moduleB.so`. Currently, we pre-load `moduleB.so`. But after #2686 and this PR I think we can let `loadWebAssemblyModule` to load `moduleB.so` dynamically. Is stack switching required for this?

hoodmane:
> What I am expecting is that, when we call loadWebAssemblyModule("moduleA_that_requires_moduleB.so") it will try to find moduleB.so. Currently, we pre-load moduleB.so. But after https://github.com/pyodide/pyodide/pull/2686 and this PR I think we can let loadWebAssemblyModule to load moduleB.so dynamically. Is stack switching required for this?

The thing that stack switching is required for is as follows: in native Python .so files are loaded on demand when you try to import them, so if there is a file on the Python path named `some_module.so` that exports a function `PyInit__some_module` and you say `import some_module` it will load `some_module.so` and give you the C module. This can't happen for us without stack switching because the Wasm loader is async only for modules that aren't quite small. So if we want to be able to `import some_module` we need to preload `some_module.so`.

hoodmane (approved):

ryanking13:
Thanks for the detailed explanation! Yes, I understand that we can't load Python `.so` files dynamically during `import module`.
On the other hand, I think we can still apply this in loading shared libraries like `openssl`, `CLAPACK` that are required by Python `.so` files. e.g. `pyodide.loadPackage("scipy")` loading `clapack_all.so` dynamically.

I'm not sure this is very useful, but at least we can prevent loading "unused" libraries. I found that some shared libraries like `suitesparse` contain several `.so` files, and some of them might be not used by Python packages.

hoodmane:
Yeah. Ideally if no Python package uses the `.so` file we wouldn't include it to begin with (so we don't have to download it) and better yet we would avoid building it too. But if package A doesn't need a library but package B does need it, then when we load package A we can skip loading it and then if we later load package B we can load it.

ryanking13:
The remaining issue is that `ssl` wants to load `_ssl` from `sys.path` (in native Linux environments `_ssl.so` will be placed in `/lib/python3.10/lib-dynload`). I guess I will have to treat `ssl` specially for now...

hoodmane:
Thanks @ryanking13!

rth:
Thanks for doing this. I think this is OK as long as there aren't other .so than those installed by pyodide and we don't plan to support virtualenv. If any of these conditions change in the future, we would probably need to revert this path change, and keep .so files associated with the python install as they are conventionally kept with pip or conda.

ryanking13:
> Thanks for doing this. I think this is OK as long as there aren't other .so than those installed by pyodide and we don't plan to support virtualenv. If any of these conditions change in the future, we would probably need to revert this path change, and keep .so files associated with the python install as they are conventionally kept with pip or conda.

Yeah, I think any installation path would be okay as long as we install in a consistent way. The purpose of this PR was to ensure that we can find a `.so` file in the same path, so that `Module.loadWebAssemblyModule` can find `.so` files in that path (which was a blocker of #2725).

PR #2724, opened 2022-06-16
Use nightly rust and simplify rust_emcc_wrapper

hoodmane:
This PR:
https://github.com/rust-lang/rust/pull/97928/files
allows us to simplify `rust_emcc_wrapper`.
Hopefully I can make another fix to Rust that will fix the libc problem and allow us to get rid of the wrapper entirely.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -242,6 +242,7 @@ emsdk/emsdk/.complete:
 rust:
 	wget https://sh.rustup.rs -O /rustup.sh
 	sh /rustup.sh -y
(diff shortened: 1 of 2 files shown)


ryanking13 (approved):
Thanks, @hoodmane! If you think we will have to update rust often, probably we should pause https://github.com/pyodide/pyodide/pull/2704 for a while?

hoodmane:
> If you think we will have to update rust often, probably we should pause https://github.com/pyodide/pyodide/pull/2704 for a while?

Yeah, I think we'd better wait on that until I get a few more fixes in:
https://github.com/rust-lang/rust/pulls?q=author%3Ahoodmane

PR #2730, opened 2022-06-16
using javascript native response.json()

pandian912:
### Description

For large responses converting to string and then back to json adds a lot of overhead.
Instead the json can be directly fetched from the javascript fetch response object.
https://developer.mozilla.org/en-US/docs/Web/API/Response/json 

### Checklists

- [ ] Add / update tests

### DIFF
--- a/src/py/pyodide/http.py
+++ b/src/py/pyodide/http.py
@@ -130,8 +130,7 @@ class FetchResponse:
         return await self.js_response.text()
 
     async def json(self) -> Any:
-        """Return the response body as a Javascript JSON object.
-        """
+        """Return the response body as a Javascript JSON object."""
         self._raise_if_failed()
         return await self.js_response.json()
 
--- a/src/py/pyodide/http.py
+++ b/src/py/pyodide/http.py
@@ -1,4 +1,3 @@
-import json
 from io import StringIO
 from typing import Any, BinaryIO, TextIO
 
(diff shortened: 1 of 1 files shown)


hoodmane:
Thanks for making the PR. This completely changes the return value: instead of returning a Python mix of dicts and lists, it returns a JsProxy. Maybe we could add another method that does this. But the point of `pyfetch` is to return Python objects, if you want a JsProxy, you can do 
```py
from js import fetch
resp = await fetch(url)
json = await resp.json()
```
so I'm not completely sure that one is needed.

Perhaps another approach would be to do `self.js_response.json().to_py()` but I think converting the string is likely to be more efficient.

 @rth

ryanking13:
I think we need some benchmark for this, comparing the speed of:

1. `resp.json()`
2. `json.loads(resp.string())`
3. `resp.json().to_py()`

pandian912:
1) the type output in the console put me off a bit. I actually thought that pyfetch was returning a Proxy
`  
    js.console.log(response.status, type(resp_json), resp_json, resp_json.keys())
      js.console.log(response.status, to_js(type(resp_json)), to_js(resp_json), to_js(resp_json.keys()))
`
`200 Proxy {arguments: null, caller: null, prototype: undefined, $$: {…}} Proxy {$$: {…}} Proxy {$$: {…}}
pyodide.asm.js:14 
200 Proxy {arguments: null, caller: null, prototype: undefined, $$: {…}} Map(4) {'userId' => 1, 'id' => 1, 'title' => 'delectus aut autem', 'completed' => false} Proxy {$$: {…}}`


2) I ran basic perf test with tiny and huge json files, I dont see any noticeable changes in either of the 2 approaches

rth:
> I dont see any noticeable changes in either of the 2 approaches

In that case, should we close this PR? If performance is comparable,  I would also say that `json.loads(response.text())` should give easier to debug error messages when in fail, as compared to recursively converting js objects to python dict.

pandian912:
Yeah sounds good to me.

PR #2732, opened 2022-06-17
Use node's native fetch if it's available

hoodmane:


ryanking13 (approved):
Thanks, @hoodmane. Probably we should update the node version in our Docker image BTW.

PR #2734, opened 2022-06-17
ENH Add flag to run bash script in cross build environment

hoodmane:
This script will run with the target environment variables and
sysconfigdata and with the pywasmcross compiler symlinks.
Any changes to the environment will persist to the main build
step but will not be seen in the post step (or anything else
done outside of the cross build environment). The working
directory for this script is the source directory.

@josephrocca 

- [x] Add / update tests


josephrocca:
I think it's likely some packages would want `build.script` to run before `build.cross-script`. And others packages might need it to be the other way around. Is that supported with this approach? If not, it might be as simple as adding a `build.script-order` option (or something like that) to switch the order?

hoodmane:
I was planning to only support first `build.script` and then `build.cross-script`.

josephrocca:
Yeah I was going to say that should probably be the default. So maybe best to leave it at that and can add option later if it's requested.

josephrocca:
I just tested this with `jax`, and it seems to be working. I was expecting to get the [`linux/futex.h` error](https://github.com/pyodide/pyodide/discussions/2720) since that's what I was previously "up to" and hadn't yet fixed, but instead got:
```
...
In file included from external/com_google_absl/absl/strings/numbers.h:27:
In file included from /src/emsdk/emsdk/upstream/lib/clang/15.0.0/include/x86intrin.h:13:
/src/emsdk/emsdk/upstream/lib/clang/15.0.0/include/ia32intrin.h:200:10: error: use of undeclared identifier '__builtin_ia32_readeflags_u32'
  return __builtin_ia32_readeflags_u32();
         ^
```
but this still looks like it's not to do with environment variable stuff. Guessing it's just some sort of build feature I need to disable akin to threading for the futex error. (Perhaps I'm doing something slightly differently than I was previously, and that's leading to getting this error before the futex one.)

hoodmane:
https://github.com/emscripten-core/emscripten-fastcomp-clang/issues/29#issuecomment-510216934

hoodmane:
The problem is that `__SSE4_2__` is defined so it assumes x64. See this PR for a similar issue:
https://github.com/numpy/numpy/pull/21734

hoodmane:
So the problem is that `-msse2` is being passed or something like that.
https://github.com/emscripten-core/emscripten/blob/fe7cfaad32706a63b9ece1cbba41999d3ace7e4b/emcc.py#L801

packages/sharedlib-test-py/meta.yaml
```diff
@@ -8,6 +8,11 @@ requirements:
 source:
   path: src
 build:
+  cross-script: |
+    pwd
+    echo PYTHONPATH: $PYTHONPATH
+    echo _PYTHON_HOST_PLATFORM: $_PYTHON_HOST_PLATFORM
+    echo PYTHONPATH: $PYTHONPATH
```
ryanking13, line 15:
```suggestion
```

It seems it is called twice?

pyodide-build/pyodide_build/buildpkg.py
```diff
@@ -440,17 +440,33 @@ def compile(
...
-        )
+        if "cross-script" in build_metadata:
+            with BashRunnerWithSharedEnvironment(build_env) as runner:
+                runner.run(build_metadata["cross-script"])
+                build_env = runner.env
+
+        from .pypabuild import build
+
+        try:
+            build(build_env, backend_flags)
```
ryanking13, line 464: Yes, calling `build` seems better to be placed here.

ryanking13 (commented):
Thanks, @hoodmane! Please update docs about `cross-script` and add tests (unless it is not too complicated to write a test).

I think your recent enhancements to our build system like `backend-flags` and `cross-script` will make package build a lot more comfortable. Thanks!

ryanking13 (approved):

PR #2735, opened 2022-06-17
MAINT Use the patched version of setuptools-rust

hoodmane:
This allows us to remove a few workarounds for setuptools-rust bugs that are now fixed.


ryanking13 (approved):
Thanks, @hoodmane!

Issue #2736, opened 2022-06-17
KeyError: 'PYWASMCROSS_ARGS'


josephrocca:
This error originally came up [here](https://github.com/pyodide/pyodide/issues/2198#issuecomment-1153120013), but even after using the new (yet to be merged) `build.cross-script` property, I'm still seeing it. Here's an exact sequence of commands that replicates this error:
```bash
git clone --depth 1 --branch cross-build-script https://github.com/hoodmane/pyodide pyodide-cross-build-script
cd pyodide-cross-build-script
./run_docker --pre-built
make
pip install ./pyodide-build
git clone --branch patch-3 --depth 1 https://github.com/josephrocca/jax /src/jax
mkdir /src/packages/jax
cd /src/packages/jax
```
Now create `meta.yaml` with this:
```yaml
package:
  name: jax
  version: 0.3.13
source:
  path: /src/jax

build: 
  script: | 
    sudo apt-get update
    sudo apt-get install -y python3-dev
    pip install numpy wheel
    
  cross-script: |
    python build/build.py
  cflags: -s USE_ZLIB=1
  ldflags: -s USE_ZLIB=1

test:
  imports:
    - jax
about:
  home: https://github.com/google/jax
  PyPI: https://pypi.org/project/jax
  summary: Differentiate, compile, and transform Numpy code.
  license: Apache-2.0
```
And now run:
```
python -m pyodide_build buildpkg meta.yaml
```
This will result in this error (see note below RE flaky-ness if you don't get this error):
```
ERROR: /src/.docker_home/.cache/bazel/_bazel_joe/7e454c473e82025e727812785b12ccc7/external/llvm-project/mlir/BUILD.bazel:6563:10: Compiling mlir/tools/mlir-tblgen/OpPythonBindingGen.cpp failed: (Exit 1): cc failed: error executing command 
  (cd /src/.docker_home/.cache/bazel/_bazel_joe/7e454c473e82025e727812785b12ccc7/execroot/__main__ && \
  exec env - \
    PATH=/src/tools/symlinks:/src/emsdk/emsdk:/src/emsdk/emsdk/upstream/emscripten:/src/emsdk/emsdk/node/14.18.2_64bit/bin:/src/node_modules/.bin/:/usr/local/bin:/usr/local/sbin:/usr/sbin:/usr/bin:/sbin:/bin:/src/packages/.artifacts/bin/ \
    PWD=/proc/self/cwd \
[...]

PR #2738, opened 2022-06-18
ENH Pass pywasmcross environment more carefully

hoodmane:
This is on top of #2734. Other build environments cannot be trusted to pass our environment variables down #2736. Instead, store them into a Python file in a temporary directory with the symlinks and import them. This also has the advantage that it makes it possible to replay specific compiler invocations from the build script because the environment is stored to disk (that is, if we override the deletion of the temporary directory by manually adjusting the code).


ryanking13 (commented):
Thanks for the work @hoodmane. Generally LGTM, but I am slightly concerned about the additional time overhead of reading a json file every time the compiler wrapper is called. Maybe we can compare the package build time when the CI passes.

hoodmane:
Let's store it in both the environment variable and in the file and only read it from the file if the environment variable is missing.

Issue #2741, opened 2022-06-20
RFC Release `pyodide-test-runner` to PyPI


ryanking13:
## Proposal

Sometimes I see people who want to test their Pyodide-based applications. So I think it would be great to release `pyodide-test-runner` to PyPI as a standalone package so that people who build applications with Pyodide can easily test their codes.

## Required steps

I think the steps needed for this are:
- [x] 1. Make `pyodide-test-runner` an installable package.
- [ ] 2. Leave only general functions and fixtures (`selenium`, `run_in_pyodide`, ...) inside `pyodide-test-runner` and move other things that will only be used in the Pyodide main repository (`package_is_built`, etc...) into some other place.
- [ ] 3. Write short documentation of using `pyodide-test-runner`.
- [ ] 4. Transfer it to a new repository and release it to PyPI.

## Things need to be considered

One thing we have to consider before transferring `pyodide-test-runner` into a new repository is that Pyodide update should affect `pyodide-test-runner` as less as possible. For example, we have a [code that initializes global hiwire objects](https://github.com/pyodide/pyodide/blob/38fae6e9cbcdd8e932a53bf1c763279dc55729d5/pyodide-test-runner/pyodide_test_runner/browser.py#L174-L194), which is vulnerable to Pyodide API updates. We'll have to move those kinds of code out of `pyodide-test-runner`.

rth:
> Transfer it to a new repository.

 It's a bit orthogonal to releasing to PyPi. I would say we should release to PyPi earlier rather than later. 

About moving it it to a new repo, yes, I can see the appeal. Though, it would make it harder to test changes to this repo, because it would be difficult to run the Pyodide test suite there, but maybe it's indeed the way forward.

ryanking13:
> Though, it would make it harder to test changes to this repo, because it would be difficult to run the Pyodide test suite there

Yes, that is the possible problem. As you say, we can start releasing it to PyPI (when it's ready) without moving it to a new repository.

rth:
Coming back to this, I'm very much in favor of creating a separate package pyodide/pyodide-test-runner. So far the testing setup looks relatively stable and I think it would make the release/update simpler. Otherwise, we would have to add yet another CI job to our already overly complex CI in this repo.

Though we need to decide the naming first #2878

PR #2742, opened 2022-06-20
TST Make `pyodide-test-runner` installable

ryanking13:
### Description

This makes `pyodide-test-runner` a regular installable Python package, instead of hacking with `sys.path`.

Related issue: #2741


rth (commented):
Thanks!

.circleci/config.yml
```diff
@@ -434,7 +439,7 @@ workflows:
 
       - test-main:
           name: test-core-node
-          test-params: -k node src packages/micropip packages/fpcast-test packages/sharedlib-test-py/ packages/cpp-exceptions-test/ pyodide-test-runner
+          test-params: -k node src packages/micropip packages/fpcast-test packages/sharedlib-test-py/ packages/cpp-exceptions-test/ pyodide-test-runner/pyodide_test_runner/tests
```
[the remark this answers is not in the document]
ryanking13: Yes, that sounds reasonable, thanks!

I updated the test path to `pyodide-test-runner/pyodide_test_runner`. (`pytest` complains when I set the test path to `pyodide-test-runner` because in that case `pyodide_test_runner` import path is ambiguous)

pyodide-test-runner/pyodide_test_runner/hook.py
```diff
@@ -0,0 +1,21 @@
+from pathlib import Path
+
+import pytest
+
+
+@pytest.hookimpl(tryfirst=True)
+def pytest_addoption(parser):
```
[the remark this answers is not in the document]
ryanking13: During installation, `pyodide-test-runner` registers `pytest11` entry points. `pytest`  looks up that entry point to discover and load plugins. So I think the path where pytest is invoked doesn't matter.

.circleci/config.yml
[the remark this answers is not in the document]
ryanking13: > I updated the test path to `pyodide-test-runner/pyodide_test_runner`. (`pytest` complains when I set the test path to `pyodide-test-runner` because in that case `pyodide_test_runner` import path is ambiguous)

Oh, it's failing... perhaps I should write separate test job for pyodide-test-runner.

.circleci/config.yml
```diff
@@ -202,6 +206,8 @@ jobs:
           name: stack-size
           command: |
             make npm-link
+            pip install ./pyodide-test-runner
+            npm install -g node-fetch@2
```
rth, line 210: Why do we need to install node-fetch? It should already be installed I think?

pyodide-test-runner/pyodide_test_runner/fixture.py
```diff
@@ -80,7 +82,7 @@ def selenium_common(
             f"Unknown runner or browser: {runner_type} / {request.param}"
         )
 
-    dist_dir = request.config.getoption("--dist-dir")
+    dist_dir = Path(os.getcwd(), request.config.getoption("--dist-dir"))
```
rth, line 85: --dist-dir could also be an absolute path in which case this will not work.

pyodide-test-runner/pyodide_test_runner/node_test_driver.js
```diff
@@ -25,8 +24,6 @@ const context = {
   TextDecoder: util.TextDecoder,
   TextEncoder: util.TextEncoder,
   URL,
-  atob: base64.decode,
-  btoa: base64.encode,
```
rth, line 29: Can you comment why removing it?

pyodide-test-runner/pyodide_test_runner/tests/test_decorator.py
```diff
@@ -1,83 +1,53 @@
-import asyncio
```
rth, line 1: Could you please comment why we need to refactor this file? Maybe Hood should review this.

rth (commented):
Thanks ! A few more questions otherwise looks good. 

The one thing I don't understand is the reason for the changes in test_decorator.py

pyodide-test-runner/pyodide_test_runner/fixture.py
ryanking13: I think it works even if `--dist-dir` is an absolute path:

```
>>> Path("/abs1", "/abs2")
PosixPath('/abs2')
```

pyodide-test-runner/pyodide_test_runner/node_test_driver.js
ryanking13: Those functions were previously required in `run_in_pyodide` decorator, but was removed (#2510).

https://github.com/pyodide/pyodide/blob/7a9273dd6c893ab1273f8c5d22f62f60296f13d5/pyodide-build/pyodide_build/testing.py#L88

.circleci/config.yml
ryanking13: node-fetch is installed locally during the installation of pyodide JS package, which means it cannot be accessed unless the js script file (`node_test_driver.js`) is located in the subdirectory of Pyodide source tree. In this PR we are installing `pyodide-test-runner`  to e.g. site-packages, so node-fetch needs to be installed globally.

Maybe we can update the node version in our docker image in a subsequent PR that supports native fetch to remove those duplicated node-fetch installations.

pyodide-test-runner/pyodide_test_runner/tests/test_decorator.py
ryanking13: I refactored this to remove all `import pyodide` statements. I wanted to make `pyodide-test-runner` testable without accessing Pyodide source tree.

What I mostly have done is replacing the selenium mock class with a real selenium class which, I guess, does not harm test coverage.

rth (approved):
Thanks!

PR #2743, opened 2022-06-20
Fix micropip.list() error when there is a package loaded from a custom url

ryanking13:
### Description

There is a bug in `micropip.list()` when there is a package loaded from a custom url by `pyodide.loadPackage()`.

#### Steps to Reproduce

```python
import pyodide_js
await pyodide_js.loadPackage("https://files.pythonhosted.org/packages/7d/4b/cdf1113a0e88b641893b814e9c36f69a6fda28cd88b62c7f0d858cde3166/snowballstemmer-2.0.0-py2.py3-none-any.whl")
import micropip
micropip.list()
```

```
  File "/lib/python3.10/site-packages/micropip/_micropip.py", line 607, in _list
    version = BUILTIN_PACKAGES[name]["version"]
KeyError: 'snowballstemmer'
```

### Checklists

- [x] Add / update tests

### DIFF
--- a/packages/micropip/src/micropip/_micropip.py
+++ b/packages/micropip/src/micropip/_micropip.py
@@ -604,10 +604,15 @@ def _list():
         if name in packages:
             continue
 
-        version = BUILTIN_PACKAGES[name]["version"]
-        source_ = "pyodide"
-        if pkg_source != "default channel":
-            # Pyodide package loaded from a custom URL
+        if name in BUILTIN_PACKAGES:
+            version = BUILTIN_PACKAGES[name]["version"]
+            source_ = "pyodide"
+            if pkg_source != "default channel":
+                # Pyodide package loaded from a custom URL
+                source_ = pkg_source
+        else:
+            # TODO: calculate version from wheel metadata
+            version = "unknown"
             source_ = pkg_source
         packages[name] = PackageMetadata(name=name, version=version, source=source_)
     return packages
--- a/packages/micropip/test_micropip.py
+++ b/packages/micropip/test_micropip.py
@@ -595,6 +595,24 @@ async def test_list_wheel_name_mismatch(mock_fetch: mock_fetch_cls) -> None:
     assert pkg_list[dummy_pkg_name].source.lower() == dummy_url
 
 
+def test_list_load_package_from_url(selenium_standalone_micropip):
+    with spawn_web_server(Path(__file__).parent / "test") as server:
+        server_hostname, server_port, _ = server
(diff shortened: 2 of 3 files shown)


packages/micropip/src/micropip/_micropip.py
```diff
@@ -604,10 +604,15 @@ def _list():
         if name in packages:
             continue
 
-        version = BUILTIN_PACKAGES[name]["version"]
-        source_ = "pyodide"
-        if pkg_source != "default channel":
-            # Pyodide package loaded from a custom URL
+        if name in BUILTIN_PACKAGES:
```
rth, line 607: BTW, I don't understand how `BUILTIN_PACKAGES` gets updated. As far as I understand, it's a set [converted from `pyodide_js._api.packages`](https://github.com/pyodide/pyodide/blob/8e8ccffddc669aad0b982798d93f0f10f8d11a78/packages/micropip/src/micropip/_compat_in_pyodide.py#L9) but should't we do this conversion where it's used each time instead of at import time once? Or otherwise keep it as a JsProxy?

rth (commented):
Thanks! 

Since micropip.list was added in the last release, this would need a changelog entry. Otherwise LGTM.

packages/micropip/src/micropip/_micropip.py
rth: Aww, right it's the fixed `package.json`. We should probably use more explicit names. The link between `_api.packages` / `BUILTIN_PACKAGES`  and `packages.json` is not very explicit from the name.

packages/micropip/test_micropip.py
```diff
@@ -595,6 +595,24 @@ async def test_list_wheel_name_mismatch(mock_fetch: mock_fetch_cls) -> None:
...
+def test_list_load_package_from_url(selenium_standalone_micropip):
+    with spawn_web_server(Path(__file__).parent / "test") as server:
+        server_hostname, server_port, _ = server
+        base_url = f"http://{server_hostname}:{server_port}/"
+        url = base_url + "snowballstemmer-2.0.0-py2.py3-none-any.whl"
+
+        selenium = selenium_standalone_micropip
+        selenium.run_js(
+            f"""
+            await pyodide.loadPackage({url!r});
```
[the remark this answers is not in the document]
ryanking13: `!r` adds single quote around the string, we need that.

```
await pyodide.loadPackage('https://...');
# await pyodide.loadPackage(https://...) -> syntax error!
```

PR #2746, opened 2022-06-20
DOCS clean and rebuild before publishing to npm

hoodmane:


rth:
We should. Also, these instructions are not correct I think because one needs to build all packages, not just the core ones, so they end up in the packages.json we ship. Otherwise, lots of packages can't be installed in Node.js which is what we had for the last release.

All the more reason to do this in CI.

hoodmane:
I am making another docs update PR, will fix this.

PR #2753, opened 2022-06-21
Pass -Zlink-native-libraries=no when building Rust

hoodmane:
This lets us delete the linker wrapper. Finally.


ryanking13 (approved):
Thanks a lot @hoodmane!

ryanking13:
It looks like there is a Cloudflare outage that somehow killed npm API (CI failure in main) 🤔

PR #2759, opened 2022-06-21
Fix micropip package loading from file:// in node

hoodmane:
The node native fetch doesn't support `file://` yet but we want micropip to support it anyways.


rth (approved):
Thanks! It would be good to have a test.

PR #2758, opened 2022-06-21
Update packages

hoodmane:
Left out:
* opencv-python
* matplotlib
@ryanking13 maybe you can work on those?

PR #2760, opened 2022-06-22
Bump matplotlib version to 3.5.2

ryanking13:


packages/matplotlib/meta.yaml
```diff
@@ -26,8 +26,9 @@ build:
   ldflags: -s USE_FREETYPE=1 -s USE_LIBPNG=1 -s USE_ZLIB=1
   replace-libs:
     - png16=png
+  script: export SETUPTOOLS_SCM_PRETEND_VERSION=3.5.2
```
hoodmane, line 29: What does this mean?

ryanking13: It tells setuptools_scm that its version is `3.5.2`. 

https://github.com/matplotlib/matplotlib/blob/710fce3df95e22701bd68bf6af2c8adbc9d67a79/setup.py#L338-L344

It looks like matplotlib was calculating its version from git tag, which is not valid in our case, which I think that is the reason of CI failure in your PR (#2758)

```
# Build log from #2758
Unpacking to: ./matplotlib-0.0...OK 
bash: line 0: cd: build/matplotlib-3.5.1/dist/matplotlib-3.5.1/: No such file or directory
```

I'm not sure why it didn't fail previously, maybe setuptools_scm made some update recently.

hoodmane:
```suggestion
  script: export SETUPTOOLS_SCM_PRETEND_VERSION=$PKG_VERSION
```

packages/matplotlib/meta.yaml
```diff
@@ -26,8 +26,9 @@ build:
...
   post: |
-    cd build/matplotlib-3.5.1/dist/matplotlib-3.5.1/
+    cd build/matplotlib-3.5.2/dist/matplotlib-3.5.2/
```
hoodmane, line 31:
```suggestion
    cd build/matplotlib-$PKG_VERSION/dist/matplotlib-$PKG_VERSION/
```

PR #2761, opened 2022-06-22
Bump opencv-python version to 4.6.0.66

ryanking13:


hoodmane:
Thanks @ryanking13! Would you mind looking into `yt` too? I think that's the last out of date one.

ryanking13:
> Thanks @ryanking13! Would you mind looking into `yt` too? I think that's the last out of date one.

Yeah I opened a PR #2768

PR #2767, opened 2022-06-22
ENH Add micropip support for loading wheels from emfs

### LINKED ISSUE #2731: Micropip should be able to install wheels from the file system
## 🚀 Feature

Micropip should be able to install a wheel which is in the emscripten file system from a path.

### Motivation

In node (and some day maybe in chrome), we can mount the local file system into the Emscripten file system. It would be useful to install these, for instance when testing whether a wheel was built correctly.

hoodmane:
Resolves #2731.

### Checklists

- [x] Add / update tests


packages/micropip/test_micropip.py
```diff
@@ -748,3 +775,13 @@ async def test_freeze(mock_fetch: mock_fetch_cls) -> None:
     assert pkg_metadata["imports"] == toplevel[0]
     assert dep1_metadata["imports"] == toplevel[1]
     assert dep2_metadata["imports"] == toplevel[2]
+
+
+@run_in_pyodide(packages=["micropip"])
+async def test_emfs_micropip(selenium_standalone_micropip, regex_wheel_name):
```
ryanking13, line 781: Maybe use [snowballstemmer wheel](https://github.com/pyodide/pyodide/tree/main/packages/micropip/test) instead, to make the test little more self-contained?

Some other tests use it such as:

https://github.com/pyodide/pyodide/blob/8c8109f1dfa9f3716f6391351152bcef23ffc006/packages/micropip/test_micropip.py#L248-L251

ryanking13 (approved):
Thanks, @hoodmane. LGTM, but I haven't followed the discussion in #2731 thoroughly so probably @rth would need to review this too.

packages/micropip/src/micropip/_micropip.py
```diff
@@ -412,42 +422,53 @@ async def install(
...
-          The file must be a wheel named in compliance with the
-          `PEP 427 naming convention <https://www.python.org/dev/peps/pep-0427/#file-format>`_.
+        - If the requirement ends in ``.whl``, the part of the requirement after
+          the last ``/``  must be a valid wheel name in compliance with the `PEP
+          427 naming convention.
+
+        - If the requirement ends in ``.whl`` and starts with ``emfs:``, it will
+          be interpreted as a path in the Emscripten file system (Pyodide's file
+          system). E.g., `emfs:../relative/path/wheel.whl` or
+          `emfs:/absolute/path/wheel.whl`.
```
rth, line 432: Trying to simplify the logic a bit.

```suggestion
        - If the requirement starts with ``emfs:``, it will
          be interpreted as a path in the Emscripten file system (Pyodide's file
          system). E.g., `emfs:../relative/path/wheel.whl` or
          `emfs:/absolute/path/wheel.whl`. In this case, only .whl files are supported.
```

packages/micropip/src/micropip/_micropip.py
```diff
@@ -412,42 +422,53 @@ async def install(
...
+
+        - If the requirement ends in ``.whl`` and does not start with ``emfs:``
+          it will be interpreted as a URL. The file must be a wheel named in
+          compliance with the `PEP 427 naming convention
+          <https://www.python.org/dev/peps/pep-0427/#file-format>`_.
```
rth, line 437:
```suggestion
        - If the requirement starts with ``http:``, ``http:`` or ends with `.whl`
          it will be interpreted as a URL. 
```
We already say what are requirements for wheel URL above.

packages/micropip/src/micropip/_micropip.py
```diff
@@ -412,42 +422,53 @@ async def install(
         A requirement or list of requirements to install. Each requirement is a
         string, which should be either a package name or URL to a wheel:
 
-        - If the requirement ends in ``.whl`` it will be interpreted as a URL.
-          The file must be a wheel named in compliance with the
-          `PEP 427 naming convention <https://www.python.org/dev/peps/pep-0427/#file-format>`_.
+        - If the requirement ends in ``.whl``, the part of the requirement after
+          the last ``/``  must be a valid wheel name in compliance with the `PEP
+          427 naming convention.
```
rth, line 427:
```suggestion
        - If the requirement is a URL it must end in ``.whl`` and the part of the requirement after
          the last ``/``  must be a valid wheel name in compliance with the 
          `PEP 427 naming convention
          <https://www.python.org/dev/peps/pep-0427/#file-format>`_.
```

packages/micropip/src/micropip/_micropip.py
```diff
@@ -412,42 +422,53 @@ async def install(
...
+          it will be interpreted as a URL. The file must be a wheel named in
+          compliance with the `PEP 427 naming convention
+          <https://www.python.org/dev/peps/pep-0427/#file-format>`_.
 
-        - If the requirement does not end in ``.whl``, it will interpreted as the
-          name of a package. A package by this name must either be present in the
-          Pyodide repository at :any:`indexURL <globalThis.loadPyodide>` or on PyPI
+        - If the requirement does not end in ``.whl``, it will interpreted as
+          the name of a package. A package with this name must either be present
+          in the Pyodide lock file or on PyPI.
```
rth, line 441:
```suggestion
        - Otherwise, it will be interpreted as
          the name of a package. A package with this name must either be present
          in the Pyodide lock file or on PyPI.
```

packages/micropip/src/micropip/_micropip.py
```diff
@@ -86,23 +90,28 @@ def is_compatible(self):
...
         return False
 
-    async def download(self, fetch_kwargs):
+    async def _fetch_bytes(self, fetch_kwargs):
         try:
-            wheel_bytes = await fetch_bytes(self.url, fetch_kwargs)
+            return await fetch_bytes(self.url, fetch_kwargs)
         except OSError as e:
-            if self.url.startswith("https://files.pythonhosted.org/"):
+            if self.parsed_url.hostname == "files.pythonhosted.org":
```
rth, line 97: Maybe also,

```suggestion
            if self.parsed_url.hostname in ["files.pythonhosted.org", "cdn.jsdelivr.net"]:
```
as we have wheels there.

packages/micropip/src/micropip/_micropip.py
```diff
@@ -86,23 +90,28 @@ def is_compatible(self):
...
                     "Check if the server is sending the correct 'Access-Control-Allow-Origin' header."
                 ) from e
 
-        self.data = BytesIO(wheel_bytes)
-
-        with ZipFile(self.data) as zip_file:
+    async def download(self, fetch_kwargs):
+        data: BytesIO | BufferedReader
+        if self.parsed_url.scheme == "emfs":
+            data = open(self.parsed_url.path, "rb")
```
rth, line 110: I wonder if it wouldn't be better to directly add this to fetch_bytes (used in `_fetch_bytes`). Yes, it's potentially one extra copy but that function already handles `file://` protocol, so it would make sense to have a function that is able to open URLs with different protocols rather than have this logic through the code.

No strong opinion about this though.

packages/micropip/src/micropip/_micropip.py
```diff
@@ -116,9 +125,10 @@ def validate(self):
             # No checksums available, e.g. because installing
             # from a different location than PyPI.
             return
+        if not isinstance(self.data, BytesIO):
```
rth, line 128: Similarly, probably better to be explicit that we don't want to check the checksum if `self.parsed_url.scheme == "emfs":` rather than if it's not BytesIO which is an implementation detail.

Actually, is this necessary? It shouldn't have a digest with `emfs:`  URL and should return already above, no?

rth (commented):
A few comments comment trying to simplify the URL/package name logic in the docstring, as I feel currently it exposes too many implementation details. IMO the logic should be something like,
 - if has protocol: "http", "https", "emfs", or ends with ".whl" then it's a URL
 - otherwise it's a package name
and even if it's not exactly the way we are checking it, I think the docstring should present a simple explanation of how the distrinction is made. Then if the implementation doesn't match we should fix it.

A few other comments below, otherwise looks good.

packages/micropip/src/micropip/_micropip.py
hoodmane: It's needed by mypy.

packages/micropip/src/micropip/_micropip.py
hoodmane: Not strictly related to this PR, but sure it seems like an improvement. Maybe store the known-CORS domains into a global variable. Relatedly, we should ask pypi to add CORs to their 404 page...

packages/micropip/src/micropip/_micropip.py
hoodmane: This is even possible without an extra copy if we make a few improvements to the code.

packages/micropip/test_micropip.py
hoodmane: I'm not sure I understand. Isn't the test already self contained?

PR #2769, opened 2022-06-23
Add noop "bind" function to PyProxyCallableMethods

### LINKED ISSUE #2757: Add "bind" to function PyProxy
## 🚀 Feature

Some libraries like Vue are calling "bind" on functions, to attempt to bind a javascript `this` to the function. Currently, there is no corresponding function in the PyProxy of functions and it is raising an error.

### Motivation

Not having `bind` is causing incompatibility with existing libraries.

### Pitch

Add a `bind` stub, even if it doesn't have any functionality.

johnwason:
### Description

This pull request adds a noop "bind" function to `PyProxyCallableMethods` in pyproxy.ts. Some libraries like Vue will call "bind" on functions passed as event handlers. Adding a no-op bind will not change the functionality of the proxy, but will prevent this error.

Fixes #2757

### Checklists

- [ ] Add / update tests

### DIFF
--- a/src/tests/test_pyproxy.py
+++ b/src/tests/test_pyproxy.py
@@ -889,7 +889,7 @@ def test_pyproxy_call(selenium):
         selenium.run_js("f.callKwargs(76, {x : 6})")
 
     assert_call("f.bind({})()", [2, 3])
-    assert_call("f.bind({}).$$ == f.$$", True)
+    assert_call("f.bind({}).$$ === f.$$", True)
 
     selenium.run_js("f.destroy()")
 
--- a/src/tests/test_pyproxy.py
+++ b/src/tests/test_pyproxy.py
@@ -889,6 +889,7 @@ def test_pyproxy_call(selenium):
         selenium.run_js("f.callKwargs(76, {x : 6})")
 
     assert_call("f.bind({})()", [2, 3])
+    assert_call("f.bind({}).$$ == f.$$", True)
 
     selenium.run_js("f.destroy()")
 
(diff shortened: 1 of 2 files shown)


hoodmane:
Do you think this is better than removing line 149?

johnwason:
I think removing line 149 will probably completely break the proxy, but I haven't tested it. Adding `bind` this way is noninvasive and appears to solve the problem.

hoodmane:
I don't think it would break the proxy, but this PR is fine.

hoodmane (approved):

johnwason:
Don't merge this yet, I am seeing another problem.

hoodmane:
Yeah that looks important.

johnwason:
The `bind()` function needed to return `this` for it to work properly. I have made the change. Sorry about that.

hoodmane:
Could you add a test to `src/tests/test_pyproxy.py` that checks that `proxy.bind({}) === proxy`?

johnwason:
Tests have been added

src/tests/test_pyproxy.py
```diff
@@ -888,6 +888,9 @@ def assert_call(s, val):
     with pytest.raises(selenium.JavascriptException, match=msg):
         selenium.run_js("f.callKwargs(76, {x : 6})")
 
+    assert_call("f.bind({})()", [2, 3])
+    assert_call("f.bind({}).$$ == f.$$", True)
```
hoodmane, line 892: What about:
```suggestion
    assert_call("f.bind({}) === f", True)
```

johnwason: Directly comparing functions didn't seem to work.

hoodmane: Okay. Let's use `===` though.

johnwason: Done

hoodmane:
Thanks @johnwason!

PR #2773, opened 2022-06-23
Don't log every exception.

dlech:
We don't need to emit a console warning every time a Python exception is wrapped in a JavaScript exception. Users can catch and log the exception themselves if they need to log it.

### Description

Background: I'm trying to implement cancellation via `KeyboardInterrupt` and this exception gets logged every time something gets cancelled and it causes unnecessary noise in the logs since cancellation is frequent in this particular application. I tried catching the KeyboardInterrupt in Python code, but since I'm using a `SharedArrayBuffer` to trigger the interrupt, there is always race condition where the interrupt can be triggered and the exception raised before the `try:` block is entered or after the `try:` block has ended (even if there is no code to run before or after the try block, the interpreter can still check for pending exceptions).

### Checklists

- ~Add a [CHANGELOG](https://github.com/pyodide/pyodide/blob/main/docs/project/changelog.md) entry~
- ~Add / update tests~
- ~Add new / update outdated documentation~

### DIFF
--- a/src/core/error_handling.c
+++ b/src/core/error_handling.c
@@ -219,6 +219,7 @@ finally:
   return jserror;
 }
 
+#ifdef DEBUG_F
 EM_JS(void, log_python_error, (JsRef jserror), {
   // If a js error occurs in here, it's a weird edge case. This will probably
   // never happen, but for maximum paranoia let's double check.
@@ -229,6 +230,7 @@ EM_JS(void, log_python_error, (JsRef jserror), {
     API.fatal_error(e);
   }
 });
+#endif
 
 /**
  * Convert the current Python error to a javascript error and throw it.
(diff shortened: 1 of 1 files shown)


src/core/error_handling.c
```diff
@@ -219,24 +219,13 @@ wrap_exception()
   return jserror;
 }
 
-EM_JS(void, log_python_error, (JsRef jserror), {
-  // If a js error occurs in here, it's a weird edge case. This will probably
-  // never happen, but for maximum paranoia let's double check.
-  try {
-    let msg = Hiwire.get_value(jserror).message;
```
hoodmane, line 226: This function doesn't seem to do anything. Have you tested that this change actually does what you want?

hoodmane:
Let's keep this but put it behind `#ifdef DEBUG_F`.

dlech:
Sounds good. I have updated the PR.

hoodmane (approved):
Thanks @dlech!

PR #2776, opened 2022-06-23
Add continuous deployment to npm

hoodmane:


ryanking13:
Thanks, @hoodmane. I was also working on this. How do you like this one?

https://github.com/ryanking13/pyodide/blob/a9f0f4211005f90f6daa5686b1f4b5220143bcf8/.circleci/config.yml#L287-L301

tools/deploy_to_npm.py
```diff
@@ -0,0 +1,30 @@
...
+cmd = ["npm", "publish"]
+if IS_PRERELEASE:
+    cmd.extend(["--tag", "next"])
+result = subprocess.run(cmd, cwd=cwd)
+if result.returncode:
+    sys.exit(result.returncode)
+
+if not IS_PRERELEASE:
+    sys.exit(
+        subprocess.run(["npm", "dist-tag", "add", f"{PKG}@{VERSION}", "next"], cwd=cwd)
```
ryanking13, line 29: I guess we don't want to add `next` tag to a stable release?

I think this was for:

```sh
# https://pyodide.org/en/0.19.1/development/maintainers.html#making-an-alpha-release
If you accidentally publish the alpha release over the stable latest tag, you can fix it with:
npm dist-tag add pyodide@a.b.c latest where a.b.c should be the latest stable version.
Then use npm dist-tag add pyodide@a.b.c-alpha.d next to set the next tag to point to
the just-published alpha release.
```

hoodmane: There are two tags `latest` which is the latest stable release and `next` which is the latest release. So when we publish an alpha, we only update next and when we publish a stable release we update both.

hoodmane:
> Thanks, @hoodmane. I was also working on this. How do you like this one?

I think yours looks better. Can you open a PR for it?

hoodmane:
This is the last thing I want to do before making an alpha release.

tools/deploy_to_npm.py
ryanking13: Okay, I was misunderstanding the docs. Thanks!

hoodmane:
I added the npm API key to CircleCI as `NPM_TOKEN`.

tools/deploy_to_npm.sh
```diff
@@ -0,0 +1,17 @@
+#!/usr/bin/env bash
+
+set -e
+
+echo "//registry.npmjs.org/:_authToken=${NPM_PUBLISH_TOKEN}" > ~/.npmrc
```
ryanking13, line 5:
```suggestion
echo "//registry.npmjs.org/:_authToken=${NPM_TOKEN}" > ~/.npmrc
```

ryanking13 (approved):
Thanks, @hoodmane!

PR #2777, opened 2022-06-23
Fix exports error message

hoodmane:


ryanking13:
Probably also rename `explicit` in docs to `requested`?

https://pyodide.org/en/latest/development/meta-yaml.html#build-exports

hoodmane:
Yeah right. I wasn't really sure how to name these.

PR #2782, opened 2022-06-24
Trigger pypi deploy from circleci

hoodmane:
@ryanking13

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -21,7 +21,6 @@ jobs:
     resource_class: small
     steps:
       - checkout
-
       - run:
           name: Test docs
           command: |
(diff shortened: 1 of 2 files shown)


.circleci/config.yml
```diff
@@ -297,6 +296,17 @@ jobs:
           command: |
             ./tools/deploy_to_npm.sh
 
+      - run:
+          name: Deploy to PyPI
+          command: |
+            # Ask github actions to deploy for us
```
ryanking13, line 302: Probably we can improve this after the 0.21.0 release.

ryanking13 (approved):
Thanks, @hoodmane. Yes, we should publish to PyPI after all tests pass.

PR #2783, opened 2022-06-24
Micropip: Improve error message when attempting to load wheel with wrong platform

hoodmane:
If the wheel is an Emscripten wheel of the wrong version, give the expected version and the wheel version. Otherwise, complain about platform mismatch.

### DIFF
--- a/packages/micropip/test_micropip.py
+++ b/packages/micropip/test_micropip.py
@@ -21,6 +21,8 @@ except ImportError:
 
 from pyodide_build import common
 
+cpver = f"cp{sys.version_info.major}{sys.version_info.minor}"
+
 
 @pytest.fixture
 def mock_platform(monkeypatch):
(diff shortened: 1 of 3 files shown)


ryanking13 (commented):
Thanks, @hoodmane. It would be great to have some tests.

packages/micropip/test_micropip.py
```diff
@@ -810,3 +814,65 @@ async def run_test(selenium, url, wheel_name):
...
+            raiseValueError(
+                r"Emscripten v2.0.27 but Pyodide was built with Emscripten v3.1.14"
+            ),
+        ),
+        (
+            "cp35",
+            "cp35m",
+            "emscripten_3_1_14_wasm32",
+            raiseValueError(
+                "Wheel abi 'cp35m' .* Supported abis are 'abi3' and 'cp310'."
```
ryanking13, line 852: Maybe remove Emscripten and Python version values from the regex check so that we don't have to update these value when we update Emscripten or Python version?

ryanking13 (approved):
Thanks, @hoodmane. There is a test failure, otherwise LGTM.
It looks like test failure in `test_typeconversions` is not related to this PR. Probably there is another error in our type conversion code...

hoodmane:
Yeah the test failure in type conversions is because on this line:
https://github.com/pyodide/pyodide/blob/deb344cd37ad75666022e26a7d68f76a638dce66/src/core/hiwire.c#L327
the tests say that it should be `-Number.MAX_SAFE_INTEGER <= result && result <= Number.MAX_SAFE_INTEGER`

Issue #2689, opened 2022-06-11
API Better organisation of the `pyodide` Python package


rth:
The `pyodide` Python package namespace is getting rather cluttered. It includes a mix of a bit random things, for instance,
 - functions related to the FFI: `create_proxy`, `to_js`, `create_once_callable`, `JsProxy`, ...
 - functions related to running code: `CodeRunner`, `run_js`
 - wrappers of JS functions: add_event_listener, `set_interval`, `open_url`, `pyfetch`
 - various util functions:  `find_imports`, `repr_shorten`
 - exceptions: JsException

and the situation is only going to get worse with time.
 
While it makes sense to keep some of these objects at the top level, others should be IMO moved to submodules. Right now the feeling I get when opening [the API docs page](https://pyodide.org/en/latest/usage/api/python-api.html) is that it's a bit messy and without much logic as to what things one can expect to find in this package.

So I would propose to,
 - [ ] more carefully review functions / classes present in this package and decide what should be moved to submodules (I'll comment more about it later)
 - [ ] create a documentation page per submodule. For example, in CPython [`http`](https://docs.python.org/3/library/http.html) API docs starts by linking to submodules (http.client, etc) which have separate documentation pages, before listing the objects it contains. I think we should do something similar. Now we have everything listed on the main page (e.g. `Console`) and others have no easily accessible  documentation (e.g. `pyfetch`) 
 - [ ] move functions to their respective submodules. When necessary allow a top-level import but with a FutureWarning stating that this will stop working, say in 2 releases. There is a balance to find to avoid too many warnings as that would be annoying for users.

Particularly for newly added functions `set_interval`, etc it's important to do this before the next release.

hoodmane:
I propose an `eval` submodule for CodeRunner, eval_code, etc, an `ffi` submodule for the ffi related stuff, and a `browser` submodule for the setTimeout wrappers.

PR #2787, opened 2022-06-25
Move eval functions to pyodide.code submodule

hoodmane:
This raises a deprecation warning the first time such a function is used. Work towards #2689.


src/py/pyodide/__init__.py
```diff
@@ -64,20 +75,32 @@
...
+
+def __getattr__(name):
+    if name in DEPRECATED_LIST:
+        warn(
+            f"{name} has been moved to pyodide.{DEPRECATED_LIST[name]}. "
+            "Accessing it through the pyodide module is deprecated.",
+            DeprecationWarning,
+        )
+        globals()[name] = globals()[f"_deprecated_{name}"]
+        return globals()[name]
```
rth, line 105: Do you need to rename at all:  I would have thought that just returning `globals()[name]` with a warning would be enough?  Though yes, they seem to be doing it in https://peps.python.org/pep-0562/

rth (commented):
+1 for creating a separate module with those, but I think we might want to avoid overwriting builtin name. Otherwise,
```
from pyodide import eval
```
will lead to confusing errors for users.

Maybe `pyodide.code`?

src/py/pyodide/__init__.py
hoodmane: This prevents the warning from being raised more than once per input.

pyodide-test-runner/pyodide_test_runner/tests/test_decorator.py
```diff
@@ -250,7 +250,7 @@ def test_raises_jsexception(selenium):
 
     @run_in_pyodide
     def raise_jsexception(selenium):
-        from pyodide import run_js
```
rth, line 253: Also needs renaming (here and above).

rth (approved):
Thanks! LGTM, once CI passes.

PR #2788, opened 2022-06-25
Add new warnoptions loadPyodide parameter

hoodmane:
Use it in the test suite to make `DeprecationWarning` into an error.

### DIFF
--- a/pyodide-test-runner/pyodide_test_runner/browser.py
+++ b/pyodide-test-runner/pyodide_test_runner/browser.py
@@ -164,7 +164,7 @@ class BrowserWrapper:
     def load_pyodide(self):
         self.run_js(
             """
-            let pyodide = await loadPyodide({ fullStdLib: false, jsglobals : self });
+            let pyodide = await loadPyodide({ fullStdLib: false, jsglobals : self, warnoptions : "error::DeprecationWarning" });
             self.pyodide = pyodide;
             globalThis.pyodide = pyodide;
             pyodide._api.inTestHoist = true; // improve some error messages for tests
--- a/pyodide-test-runner/pyodide_test_runner/browser.py
+++ b/pyodide-test-runner/pyodide_test_runner/browser.py
@@ -164,7 +164,7 @@ class BrowserWrapper:
     def load_pyodide(self):
         self.run_js(
             """
-            let pyodide = await loadPyodide({ fullStdLib: false, jsglobals : self, warnoptions : "error::DeprecationWarning" });
+            let pyodide = await loadPyodide({ fullStdLib: false, jsglobals : self, env: {PYTHONWARNINGS : "error::DeprecationWarning"} });
             self.pyodide = pyodide;
             globalThis.pyodide = pyodide;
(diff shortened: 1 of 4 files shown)


pyodide-test-runner/pyodide_test_runner/browser.py
```diff
@@ -164,7 +164,7 @@ def javascript_setup(self):
     def load_pyodide(self):
         self.run_js(
             """
-            let pyodide = await loadPyodide({ fullStdLib: false, jsglobals : self });
+            let pyodide = await loadPyodide({ fullStdLib: false, jsglobals : self, warnoptions : "error::DeprecationWarning" });
             self.pyodide = pyodide;
             globalThis.pyodide = pyodide;
             pyodide._api.inTestHoist = true; // improve some error messages for tests
```
ryanking13, line 170: For example, how about adding warnoptions here such as:

```
pyodide._module['PYTHONWARNINGS'] = ...
```

ryanking13 (commented):
Thanks, @hoodmane! I think this is mostly for tests and not for Pyodide end users, so what do you think of adding it inside our test API instead of `loadPyodide`?

hoodmane:
What if other people want to adjust warnings in their tests?

ryanking13:
> What if other people want to adjust warnings in their tests?

I think people can use `warnings.filterwarnings` or set the PYTHONWARNINGS variable during runtime unless it is required to adjust warning filters during Pyodide load time.

WDYT @rth?

hoodmane:
The `PYTHONWARNINGS` variable is only used at Python initialization, subsequent updates to it will be ignored. People could use `warnings.filterwarnings`. You could also use that in native Python but it provides a `-W` flag to pass warning settings anyways. Of the flags in `python --help` this one looks the most useful to me.

hoodmane:
@ryanking13 ideally it would be good to set these warnings to error on our core tests and not on the package tests. You have any idea what the best way to handle this would be?

ryanking13:
> You could also use that in native Python but it provides a `-W` flag to pass warning settings anyways. Of the flags in `python --help` this one looks the most useful to me.

Okay, I was a bit reluctant to add too many options in `loadPyodide`, but it seems like it is quite reasonable to have `warnoptions`, corresponding to `-W` flags in Python.

> ideally it would be good to set these warnings to error on our core tests and not on the package tests.

Yes, I'm +1 for it.

rth:
So there is a [number env variables](https://docs.python.org/3/using/cmdline.html#environment-variables) that control Python behavior (with warnings probably being the most useful). I certainly see the use case for this, but what bothers me is that a) we can't keep up adding new parameters for other env variables, b) that the chosen name is non standard.

Maybe we could pass the `env` and set it to `env: {PYTHONWARNINGS`: ...}`?  It's a bit more verbose but more generic. I suppose these env variables need to be passed at init, right (and setting them later has no effect)?

hoodmane:
> I suppose these env variables need to be passed at init, right (and setting them later has no effect)?

That is correct.

PR #2789, opened 2022-06-25
Add Robot Raconteur package to Pyodide

johnwason:
### Description

This PR adds the Robot Raconteur Pyodide package, which is a modified version of Robot Raconteur designed to run in the Pyodide environment, using WebSockets for communication. See https://github.com/robotraconteur/robotraconteur for the full version of the package. Robot Raconteur is used to communicate with robots and other automation components, and has a large library of drivers and packages: https://github.com/robotraconteur/robotraconteur-directory . It is used as the bases for PyRI, a robotics programming environment: https://github.com/pyri-project/pyri-core/blob/master/README.md . Using a bridge, this package can also be used to communicate with ROS2 networks: https://github.com/robotraconteur-contrib/robotraconteur_ros2_bridge

This PR does several things:

* Add the `RobotRaconteur` package
* Add a `boost-cpp` package, which is required to build `RobotRaconteur`
* Add a patch to `emsdk` for `library_eventloop.js`, see https://github.com/emscripten-core/emscripten/pull/17264
* Add a new environmental variable compile option `PYODIDE_WASM_EXCEPTIONS`. When enabled, this option will use `-fwasm-exceptions` for exceptions, which provides a significant performance improvement for C++.

The `test_cpp_exceptions.py` tests are failing in the `cpp-exceptions-test` package with the error: `TypeError: Module.ExceptionInfo is not a constructor` when `-fwasm-exceptions` are used. I assume this is due to the different way that exceptions are handled when webassembly native exceptions are used.

### Checklists

- [ ] Add / update tests


johnwason:
This PR also builds swig from source in the docker image since swig >=4.02 is required.

johnwason:
Not sure why this variable substitution is failing in boost-histogram:

```
em++: error: ${SIDE_MODULE_EXCEPTION_CXXFLAGS}: No such file or directory
```

emsdk/patches/0002-library-eventloop-signatures.patch
```diff
@@ -0,0 +1,150 @@
...
+diff --git a/src/library_eventloop.js b/src/library_eventloop.js
+index 2f232900dcd..5cf835d9973 100644
+--- a/src/library_eventloop.js
++++ b/src/library_eventloop.js
+@@ -73,7 +73,7 @@ LibraryJSEventLoop = {
+     '$runtimeKeepalivePush', '$runtimeKeepalivePop',
+ #endif
+   ],
+-  emscripten_set_immediate__sig: 'iii',
++  emscripten_set_immediate__sig: 'ipp',
```
hoodmane, line 104: These changes replacing 'i' with 'p' surely don't matter for us: they would be a problem for the wasm64 target but we don't use it and it doesn't work yet.

hoodmane: Ah I see it was requested upstream. That's fine then.

hoodmane:
> why this variable substitution is failing

Two possible issues:
1. Only a subset of the environment is passed down:
https://github.com/pyodide/pyodide/blob/e4397994f1061ca8af25299b96212815feffef66/pyodide-build/pyodide_build/buildpkg.py#L130
2. Substitution in certain variables is implemented in Python and may not exactly match what you would expect from either bash or Makefile variable substitution:
https://github.com/pyodide/pyodide/blob/e4397994f1061ca8af25299b96212815feffef66/pyodide-build/pyodide_build/pywasmcross.py#L586

johnwason:
I added the environmental variables to `buildpkg.py` here:

https://github.com/pyodide/pyodide/pull/2789/files#diff-b96d844fd86cba16c9863eb73c64ff6a57ae1486f56618d69d3db597b3858aeaR150-R152

The other packages are working correctly.

hoodmane:
> The test_cpp_exceptions.py tests are failing in the cpp-exceptions-test package with the error: TypeError: Module.ExceptionInfo is not a constructor when -fwasm-exceptions are used. I assume this is due to the different way that exceptions are handled when webassembly native exceptions are used.

The issue is with the runtime support for formatting uncaught exceptions. I upstreamed a variant of Pyodide's system and since then @aheejin has improved it and added support for wasm exceptions too. So ideally we would switch to using that upstream mechanism before merging this. It looks like it made it into Emscripten 3.1.14 so it should be straightforward to use it.

hoodmane:
Looks like we need to pass `-sEXPORT_EXCEPTION_HANDLING_HELPERS` and then we can use `Module.getExceptionMessage(e)` to get a useful string representation from whatever object the runtime throws.

Another important point is that Rust currently only supports Emscripten exceptions so until we add upstream support to Rust for wasm exceptions, any Rust packages will break with `-fwasm-exceptions`. We can make a note about this in the docs.

johnwason:
Since the wasm exceptions are optional it would make sense to mark it use-at-your-own-risk. Only a few packages use exceptions anyway.

hoodmane:
Yeah. Of course, in another year or so we would like to switch to using `-fwasm-exceptions` but currently too many people are still using browsers that don't support it.

hoodmane:
Could you split adding the packages and emscripten patch into a separate PR? That will be easier to merge than the wasm-exceptions stuff.

johnwason:
See #2794. I will create another PR with wasm exception support once #2794 is merged.

PR #2790, opened 2022-06-26
Move more functions into submodules of pyodide

### LINKED ISSUE #2689: API Better organisation of the `pyodide` Python package

hoodmane:
This resolves #2689.


ryanking13 (approved):
Thanks for the maintenance work @hoodmane! Names of submodules look good to me.

Please

1. add `pyodide.browser` docs
2. update the changelog
3. and update the deprecation timeline (maybe drop them in 0.23.0 to ensure two minor releases?)

packages/matplotlib/src/browser_backend.py
```diff
@@ -3,7 +3,7 @@
 from matplotlib.backend_bases import FigureCanvasBase, NavigationToolbar2, TimerBase
 
 from js import document
-from pyodide import (
+from pyodide.browser import (
```
rth, line 6: Maybe call this `js_wrappers` ? It's not really browser-specific, and also exists in node for instance.

src/py/pyodide/webloop.py
```diff
@@ -405,4 +405,19 @@ def set_event_loop(self, loop: Any) -> None:
         self._default_loop = loop
 
 
+def _initialize():
```
rth, line 408:
```suggestion
def _initialize_event_loop():
```
to be more specific.

docs/usage/api/python-api.md
```diff
@@ -17,11 +17,24 @@ By default there are two JavaScript modules. More can be added with
...
+.. toctree::
+   :hidden:
 
-.. automodule:: pyodide
-   :members:
-   :autosummary:
-   :autosummary-no-nesting:
+   python-api-code.md
+   python-api-ffi.md
+   python-api-http.md
```
rth, line 39: It would be probably better to have a shared parent folder.
```suggestion
   python-api/code.md
   python-api/ffi.md
   python-api/http.md
```

docs/usage/api/python-api.md
```diff
@@ -17,11 +17,24 @@ By default there are two JavaScript modules. More can be added with
...
+
+```{eval-rst}
+.. list-table::
+
+   *  - :py:mod:`pyodide.code`
+      - Utilities for evaluating Python and JavaScript code.
+   *  - :py:mod:`pyodide.ffi`
+      - The :any:`JsProxy` class and utilities to help interact with JavaScript code.
+   *  - :py:mod:`pyodide.http`
+      - Defines :any:`pyfetch` and other functions for making network requests.
```
rth, line 30: Also need to add console, and webloop, but we don't need to do it in this PR.

rth (commented):
So it means we would have nothing left in the root namespace. I suppose that's OK.

A few comments below otherwise LGTM.

packages/matplotlib/src/browser_backend.py
hoodmane: Maybe this should be a submodule of `ffi` or something. `pyodide.ffi.wrappers`?

rth: Yes, sounds good.

docs/usage/api/python-api.md
```diff
@@ -17,11 +17,32 @@ By default there are two JavaScript modules. More can be added with
...
-   :members:
-   :autosummary:
-   :autosummary-no-nesting:
-```
+   python-api/code.md
+   python-api/console.md
+   python-api/ffi.md
+   python-api/http.md
+   python-api/webloop.md
+````
```
ryanking13, line 48: This toctree is not correctly parsed.

https://pyodide--2790.org.readthedocs.build/en/2790/usage/api/python-api.html

PR #2793, opened 2022-06-26
Add py-cpuinfo

hoodmane:
This package is a dependency of `pytest-benchmark`.  `py-cpuinfo` doesn't have a wheel on PyPI. There is an open PR to add wheels, but `py-cpuinfo` isn't maintained.
https://github.com/workhorsy/py-cpuinfo/pull/165
https://github.com/samuelcolvin/pydantic-core/pull/118


ryanking13:
Perhaps we should ask `pytest-benchmark` to make cpuinfo an optional dependency? It doesn't seem `py-cpuinfo` will work correctly in browser environments to me.

hoodmane:
Sounds like a better option.

PR #2794, opened 2022-06-26
Add Robot Raconteur package to Pyodide (without wasm-exceptions)

johnwason:
### Description

This PR adds the Robot Raconteur Pyodide package, which is a modified version of Robot Raconteur designed to run in the Pyodide environment, using WebSockets for communication. See https://github.com/robotraconteur/robotraconteur for the full version of the package. Robot Raconteur is used to communicate with robots and other automation components, and has a large library of drivers and packages: https://github.com/robotraconteur/robotraconteur-directory . It is used as the bases for PyRI, a robotics programming environment: https://github.com/pyri-project/pyri-core/blob/master/README.md . Using a bridge, this package can also be used to communicate with ROS2 networks: https://github.com/robotraconteur-contrib/robotraconteur_ros2_bridge

This PR does several things:

* Add the `RobotRaconteur` package
* Add a `boost-cpp` package, which is required to build `RobotRaconteur`
* Add a patch to `emsdk` for `library_eventloop.js`, see https://github.com/emscripten-core/emscripten/pull/17264
* Build swig 4.02 from source in docker. Robot Raconteur requires swig >=4.02

This is the same as #2789 but removes `-fwasm-exceptions` support.

### Checklists

- [ ] Add / update tests


johnwason:
```
-- Found SWIG: /usr/bin/swig3.0 (found version "3.0.12")  
CMake Error at /usr/local/lib/python3.10/site-packages/cmake/data/share/cmake-3.22/Modules/FindPackageHandleStandardArgs.cmake:230 (message):
  Could NOT find SWIG: Found unsuitable version "3.0.12", but required is at
  least "4.0.2" (found /usr/bin/swig3.0, )
'''

This error will be fixed with the docker updates in this PR.

**Comment by hoodmane** [comment]
How long does it take to build swig?

**Comment by johnwason** [comment]
> How long does it take to build swig?

On my workstation a few seconds. Not sure about on CI, but not long.

**Comment by hoodmane** [comment]
That seems acceptable. Adding it to the docker image means it won't be built in CI either. Though it might be easier to build swig as part of the build script of robot raconteur though? It's annoying having to update the docker image.

**Comment by johnwason** [comment]
For Debian packages I have been including the generated source in the package since the swig included is out of date. Let me see if I can do that here.

File: Makefile.envs [hunk 1]
```diff
@@ -115,6 +115,8 @@ export MAIN_MODULE_LDFLAGS= $(LDFLAGS_BASE) \
 	-lnodefs.js \
 	-lproxyfs.js \
 	-lworkerfs.js \
+	-lwebsocket.js \
+	-leventloop.js \
```
hoodmane, line 119: For the record, these seem to add 100kb to the output size.

johnwason: I don't thing it is possible to include those as part of a side module.

hoodmane: Definitely not. I think the 100kb is fine, but it's good to record it in the PR.

hoodmane: I suspect most of it is from the web sockets which at some point other people will likely want to use.

hoodmane (approved):

PR #2795, opened 2022-06-26
Add macros to extract common logic for calling into Python

hoodmane:
This makes new macros to extract the common logic for try, catch, finally blocks when calling into Python. The main reason to do this is so that we can easily adjust all of these entry points at the same time. In particular, I want to add a counter that tracks the entry depth so that we can have special handling for certain errors (for instance, `SystemExit`) when exiting the outermost Python layer.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -113,7 +113,7 @@ src/js/pyproxy.gen.ts : src/core/pyproxy.* src/core/*.h
 	echo "// Do not edit it directly!" >> $@
 	cat src/core/pyproxy.ts | \
 		sed '/^\/\/\s*pyodide-skip/,/^\/\/\s*end-pyodide-skip/d' | \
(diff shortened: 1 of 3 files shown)


rth:
Personally, I find that this makes the code significantly less readable, as it introduces a somewhat custom syntax for these code blocs. No other way we could avoid it?

hoodmane:
Figured you wouldn't like it. Not sure. Emscripten has custom syntax for their JavaScript for similar reasons.

One alternative is to add a whole lot of extra function calls. I'm concerned that it would cause a significant loss of performance in certain workflows but who knows. We already have a lot of layers of indirection so maybe adding a few more won't matter that much.

Another option is to copy paste code, however this construct occurs 22 times so adding a few extra lines to each one is annoying, easy to mess up, and adds a lot of repetition. Otoh it isn't great for code size to be able to add so much code without noticing.

We don't benchmark this, maybe doing it with lambdas and three extra function calls would

hoodmane:
Maybe I should just copy paste.

PR #2799, opened 2022-06-28
Add pytest-benchmark

hoodmane:


ryanking13 (approved):
Thanks @hoodmane. Could you add some tests and update the changelog?

Issue #366, opened 2019-03-26
Provide debug builds


mdboom:
The pre-build versions of Pyodide are all "release" and missing debugging symbols.

It might be handy to also provide debug builds built with the `--profiling` so that WebAssembly debugging will work better.  This will double our compile times, though, but maybe that's not too bad if it's only on tagged releases.

PR #2800, opened 2022-06-28
Deploy a debug version of Pyodide to the CDN

hoodmane:
At end of deploy stage, rebuild Pyodide with debug symbols and deploy it.

Closes https://github.com/pyodide/pyodide/issues/366


hoodmane:
@rth

ryanking13:
I think Roman wanted to drop `pyodide-cdn2.iodide.io`. Perhaps we can deploy it to jsdelivr?
BTW, I can't find a CI job which deploys to the jsdelivr.

hoodmane:
I think jsdelivr is currently set up to mirror `pyodide-cdn2.iodide.io`?

rth:
> I think jsdelivr is currently set up to mirror pyodide-cdn2.iodide.io?

No. What happens is that we have a S3 buckets named that way for historical reasons that we, unfortunately, can't easily rename. The bucket name doesn't really matter and it's distinct from a domain name. It is then,
 - exposed via this iodide.io domain name through Cloudflare however caching doesn't work as expected.
 - exposed via jsdelivr with proper caching.

So the bottom line is we should never use the pyodide-cdn2.iodide.io domain name. I need to ask [wlach](https://github.com/wlach)  to setup a redirect to jsdelivr.

rth (approved):
Thanks! Yes, having a debug build has been discussed for a while https://github.com/pyodide/pyodide/issues/366 though we mostly managed without so far.

hoodmane:
Oops we're in the wrong docker image. I guess I'll have to rearrange it a bit.

PR #2811, opened 2022-06-29
More work on create_xbuildenv to allow building numpy

hoodmane:
This is more WIP on create_xbuildenv. I am including some of the `WASM_LIB_DIR` but not all of it to keep size in check. The current cross build environment that we would upload/download is 20mb.

To use this, we need an extra CLI entrypoint which I am working on.


hoodmane:
Successfully built out of tree (in a slightly ad-hoc way):
* numpy
* cryptography
* astropy

pyodide-build/pyodide_build/create_xbuildenv.py
```diff
@@ -29,12 +29,51 @@ def copy_xbuild_files(xbuildenv_path):
...
+        sysconfig_dir,
+        Path("Makefile.envs"),
+        wasm_lib_dir / "CLAPACK",
+        wasm_lib_dir / "cmake",
+        Path("tools/pyo3_config.ini"),
+    ]
+    # Some ad-hoc stuff here to moderate size. We'd like to include all of
+    # wasm_lib_dir but there's 180mb of it. Better to leave out all the video
+    # codecs and stuff.
+    for pkg in ["ssl", "libcrypto", "zlib", "xml", "mpfr"]:
```
ryanking13, line 57: So this assumes that all of libraries are already built and we copy them into a isolated xbuild env?

hoodmane: If you're curious, an example of a working xbuildenv is here:
http://pyodide-cache.s3-website-us-east-1.amazonaws.com/xbuildenv/1.tar
If you install `pyodide-build` from my `out-of-tree` branch:
```
pip install git+https://github.com/hoodmane/pyodide.git@out-of-tree#subdirectory=pyodide-build
```
then you can build e.g., numpy with:
```
pywasmbuild
```

pyodide-build/pyodide_build/install_xbuildenv.py
```diff
@@ -37,5 +38,5 @@ def main(args):
         ]
     )
     shutil.copytree(
-        xbuildenv_path / "site-packages", host_site_packages, dirs_exist_ok=True
+        xbuildenv_path / "site-packages-extras", host_site_packages, dirs_exist_ok=True
```
rth, line 41: Could you add a comment about what's happening here, and why it needs to be != from site-packages.

rth (approved):
Thanks, generally LGTM.

pyodide-build/pyodide_build/install_xbuildenv.py
hoodmane: The name change isn't technically necessary, it's intended as a clarification. The real `site_packages` directory is at
```
pyodide-root/packages/.artifacts/lib/python3.10/site-packages/
```
and is generated by
```sh
pip install -r requirements.txt -t $HOSTSITEPACKAGES
cp site-packages-extra $HOSTSITEPACKAGES
```
hence the name `site-packages-extra`.

PR #2812, opened 2022-06-30
Enable libhdf5

ryanking13:
I guess we need to do this before v0.21.0 release?


ryanking13:
I have no idea why importing `h5py` is resulting in an infinite recursion...

ryanking13:
I am not sure I can fix this before we release 0.21.0. I have zero knowledge about libhdf5 so probably I need some help from @bmaranville.

Since `h5py` was not included in 0.20.0 so I think it is okay to release 0.21.0 without including `h5py`.

bmaranville:
The source of the errors is mysterious to me too.  Might need help from the HDF group to unwind this.

hoodmane:
I don't think this needs to be a release blocker. We can put it back in in v0.21.1.

ryanking13:
Okay, let's do this after 0.21.0 release.

ryanking13:
Okay, I think it works now.

I changed the libhdf5 into a shared library and the error is gone. I don't understand why, but it works. 🤷

I added some hacks to build a shared library version of libhdf5. I hope someone who knows libhdf5 build system better than me can remove these hacks someday.

docs/development/meta-yaml.md
```diff
@@ -150,6 +150,14 @@ Python package, the script section will be run before the build system runs
 `setup.py`. This script is run by `bash` in the directory where the tarball was
 extracted.
 
+There are special environment variables defined:
+
+- `$PKGDIR`: The directory in which the `meta.yaml` file resides.
```
ryanking13, line 155: I changed the `PKGDIR` available in build/script.

docs/development/meta-yaml.md
```diff
@@ -161,14 +169,7 @@ is the source directory.
 
 ### `build/post`
 
-Shell commands to run after building the library. These are run with `bash`, and
-there are two special environment variables defined:
-
-- `$SITEPACKAGES`: The `site-packages` directory into which the package has been
```
ryanking13, line 167: I guess this `SITEPACKAGES` variable does not exist anymore?

ryanking13:
@hoomane Can I ask for a review on this? This is a blocker of #3012.

docs/development/meta-yaml.md
hoodmane: While we're updating this, can we add info about which directory the post script is run in? Would also be helpful to mention here that when this script is run, the `dist` directory will contain the built wheel unpacked with `python -m wheel unpack` so it's possible to manually add, delete, change, move files etc. I probably should have updated the docs here when I made these changes...

packages/h5py/patches/0001-Fix-incompatible-pointer-type.patch
```diff
@@ -0,0 +1,57 @@
+From 30ba08deaa551b0920530f6b2edac59f2dd9e28e Mon Sep 17 00:00:00 2001
```
hoodmane, line 1: Did you upstream this?

hoodmane (approved):
I'm okay with merging this assuming that it works. Sorry for the slow review, I forgot about this.

packages/h5py/patches/0001-Fix-incompatible-pointer-type.patch
ryanking13: Yes, I did (not merged yet): https://github.com/h5py/h5py/pull/2142

docs/development/meta-yaml.md
ryanking13: Updated docs about `build/post` key. Maybe we can change the directory where the post script is run to make it consistent with `build/script`.

hoodmane: Thanks.

> Maybe we can change the directory where the post script is run to make it consistent with build/script.

I always get confused about the directories so this might be good. But if so, we should do it as a followup.

PR #2815, opened 2022-06-30
Add plotly package

afbarbaro:
### Description
Adds `plotly` to the provided packages. This package offers several additional types of charts and visualizations and complements `bokeh`, which is already included in `pyodide`. Makes it easier to start producing charts without having to install the package in code (because it can be loaded directly using `loadPackagesFromImports`).

### Checklists

- [x] Add / update tests

### DIFF
--- a/packages/plotly/test_plotly.py
+++ b/packages/plotly/test_plotly.py
@@ -1,0 +1,12 @@
+from pyodide_test_runner import run_in_pyodide
+
+
+@run_in_pyodide(packages=["plotly"])
+def test_plotly(selenium):
+    import plotly.express as px
+
+    df = px.data.gapminder().query("country=='Canada'")
+    fig = px.line(df, x="year", y="lifeExp", title='Life expectancy in Canada')
+    fig.show()
+
+    
--- a/packages/plotly/test_plotly.py
+++ b/packages/plotly/test_plotly.py
@@ -6,7 +6,5 @@ def test_plotly(selenium):
     import plotly.express as px
 
     df = px.data.gapminder().query("country=='Canada'")
-    fig = px.line(df, x="year", y="lifeExp", title='Life expectancy in Canada')
+    fig = px.line(df, x="year", y="lifeExp", title="Life expectancy in Canada")
     fig.show()
-
-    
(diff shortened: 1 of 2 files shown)


packages/plotly/test_plotly.py
```diff
@@ -0,0 +1,10 @@
+from pyodide_test_runner import run_in_pyodide
+
+
+@run_in_pyodide(packages=["plotly"])
```
hoodmane, line 4: you will probably need:
```suggestion
@pytest.mark.xfail_browsers(node="plotting doesn't work in node")
@run_in_pyodide(packages=["plotly"])
```

ryanking13:
> Makes it easier to start producing charts without having to install the package in code (because it can be loaded directly using `loadPackagesFromImports`).

If the package is installable through `micropip`, we prefer not to add it to the repository if possible. We recently added [`micropip.freeze()`](https://pyodide.org/en/latest/usage/api/micropip-api.html#module-micropip) API which, I think, is suitable for your use case. Could you try that instead?

afbarbaro:
> > Makes it easier to start producing charts without having to install the package in code (because it can be loaded directly using `loadPackagesFromImports`).
> 
> If the package is installable through `micropip`, we prefer not to add it to the repository unless it requires some patch. We recently added [`micropip.freeze()`](https://pyodide.org/en/latest/usage/api/micropip-api.html#module-micropip) API which, I think, is suitable for your use case. Could you try that instead?

Thank you! `micropip.freeze()` looks helpful. I will try it out.

afbarbaro:
> > > Makes it easier to start producing charts without having to install the package in code (because it can be loaded directly using `loadPackagesFromImports`).
> > 
> > 
> > If the package is installable through `micropip`, we prefer not to add it to the repository unless it requires some patch. We recently added [`micropip.freeze()`](https://pyodide.org/en/latest/usage/api/micropip-api.html#module-micropip) API which, I think, is suitable for your use case. Could you try that instead?
> 
> Thank you! `micropip.freeze()` looks helpful. I will try it out.

This worked well for my usecase, so I'm closing out this PR. Thank you @hoodmane !

PR #2820, opened 2022-07-01
Fix GHA timeout on firefox test

ryanking13:
For some reason, the recent playwright-python update (`v.1.23.0`) made this test slower on firefox, which results in a CI timeout.

### DIFF
--- a/src/tests/test_typeconversions.py
+++ b/src/tests/test_typeconversions.py
@@ -14,7 +14,7 @@ from pyodide_test_runner.hypothesis import (
 
 
 @given(s=text())
-@settings(deadline=15000)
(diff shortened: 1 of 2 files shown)


src/tests/test_typeconversions.py
```diff
@@ -14,7 +14,7 @@
 
 
 @given(s=text())
-@settings(deadline=4000)
+@settings(deadline=15000)
```
hoodmane, line 17: Is this change still needed?

ryanking13: Nop, it is not needed.

hoodmane (approved):

PR #2822, opened 2022-07-01
Remove replace_libs

hoodmane:
We don't use this feature anymore.


ryanking13:
It is still in matplotlib meta.yaml file. But it seems like it successfully builds?

https://github.com/pyodide/pyodide/blob/52287a179dfd9aa0db28dca4ac54051cbc86ba04/packages/matplotlib/meta.yaml#L27-L28

ryanking13 (approved):
Thanks, @hoodmane. Please also remove it from the meta.yaml specification docs.

PR #2821, opened 2022-07-01
Respect symbol visibility in export calculation

hoodmane:
We shouldn't export symbols with `__attribute__((visibility("hidden")))`. Numpy has a test to check that symbols with this attribute aren't exported. This fixes that test.

PR #2823, opened 2022-07-01
Finish out of tree build system (except xbuildenv deploy)

hoodmane:
This completes the out of tree build CLI. This PR is paired up with:
https://github.com/numpy/numpy/pull/21895
I have also successfully built scikit-learn, statsmodels, pandas, and astropy with this.
 
The last thing we need to do after this is set up deployment of the cross build environment. We can deploy one version to s3 for each tagged commit. I will do that in a separate PR after this is merged.

This is on top of #2821.

### Checklists

- [ ] Add / update tests


pyodide-build/pyodide_build/install_xbuildenv.py
```diff
@@ -15,12 +15,27 @@ def make_parser(parser: argparse.ArgumentParser) -> argparse.ArgumentParser:
...
-    xbuildenv_path = Path(args.xbuild_env[0])
+def download_xbuild_env(xbuildenv_path: Path) -> None:
+    from shutil import rmtree, unpack_archive
+    from tempfile import NamedTemporaryFile
+    from urllib.request import urlretrieve
+
+    rmtree(xbuildenv_path, ignore_errors=True)
+    with NamedTemporaryFile(suffix=".tar") as f:
+        urlretrieve(
+            "http://pyodide-cache.s3-website-us-east-1.amazonaws.com/xbuildenv/1.tar",
```
hoodmane, line 31: I'm intending to deploy these to e.g., `xbuildenv/0.21.0a3.tar". So in the future we will look up the version of `pyodide_build` and use that in the string here.

pyodide-build/pyodide_build/common.py
```diff
@@ -229,11 +229,14 @@ def init_environment() -> None:
         os.environ["PYODIDE_ROOT"] = str(search_pyodide_root(os.getcwd()))
 
     os.environ.update(get_make_environment_vars())
-    hostsitepackages = get_hostsitepackages()
-    pythonpath = [
-        hostsitepackages,
-    ]
-    os.environ["PYTHONPATH"] = ":".join(pythonpath)
+    try:
+        hostsitepackages = get_hostsitepackages()
```
hoodmane, line 233: This call will fail if we haven't downloaded the xbuildenv yet.

hoodmane, line 233: This call will fail if we haven't installed the xbuildenv yet.

hoodmane:
@ryanking13 would you mind looking into the failures here? I don't understand how they are related to the changes and cannot reproduce them locally.

ryanking13 (commented):
Thanks, @hoodmane. Probably Roman should review this too, but I think we are going in the proper direction. There are some hard-coded strings that would better be parameterized, but I think that can be done in a separate PR.

So if I understand correctly, this PR is to make the following available?

```sh
git clone https://github.com/a-package-that-can-be-built-with-pypa-build/pkg
cd pkg
pip install pyodide-build
pywasm
# pkg-1.0.0-wasm-cp310-cp310-emscripten_3_1_14_wasm32.whl
```

pyodide-build/pyodide_build/install_xbuildenv.py
```diff
@@ -15,12 +15,27 @@ def make_parser(parser: argparse.ArgumentParser) -> argparse.ArgumentParser:
...
-    xbuildenv_path = Path(args.xbuild_env[0])
+def download_xbuild_env(version: str, xbuildenv_path: Path) -> None:
+    from shutil import rmtree, unpack_archive
+    from tempfile import NamedTemporaryFile
+    from urllib.request import urlretrieve
+
+    rmtree(xbuildenv_path, ignore_errors=True)
+    with NamedTemporaryFile(suffix=".tar") as f:
+        urlretrieve(
+            f"http://pyodide-cache.s3-website-us-east-1.amazonaws.com/xbuildenv/{version}.tar",
```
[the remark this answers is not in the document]
hoodmane: Roman set it up. We could use it for other things too.

pyodide-build/pyodide_build/install_xbuildenv.py
```diff
@@ -42,3 +57,11 @@ def main(args: argparse.Namespace) -> None:
     shutil.copytree(
         xbuildenv_path / "site-packages-extras", host_site_packages, dirs_exist_ok=True
     )
+
+
+def main(args: argparse.Namespace) -> None:
+    xbuildenv_path = Path(args.xbuild_env[0])
+    version = "2"
```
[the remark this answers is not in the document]
hoodmane: One version per Pyodide release. After we merge this I am planning to add a CI job to deploy the cross build environment as part of deploy-release. It would also be nice to periodically deploy a "nightly" build at some point. Maybe once a month or once every other week or something.

hoodmane:
> So if I understand correctly, this PR is to make the following available?

Yup. See here for instance:
https://github.com/hoodmane/numpy/blob/12c656f228eb548a8f633876466345f15e39b095/.github/workflows/emscripten.yml#L49

ryanking13:
> > So if I understand correctly, this PR is to make the following available?
> 
> Yup. See here for instance: https://github.com/hoodmane/numpy/blob/12c656f228eb548a8f633876466345f15e39b095/.github/workflows/emscripten.yml#L49

Nice, it looks very cool :)

hoodmane:
@rth would you mind reviewing this?

pyodide-build/setup.py
```diff
@@ -5,6 +5,7 @@
     setuptools.setup(
         entry_points={
             "console_scripts": [
+                "pywasm = pyodide_build.out_of_tree.__main__:main",
```
rth, line 8: That's a bit confusing with https://github.com/mohanson/pywasm

So I guess it would be a bit difficult to use subcommands here. Sill maybe something like `pyodide-build-wasm` even if it's a bit more verbose? pywasmbuild would also be better. Ideally, it would be to have pyodide in the name but maybe it will be too complex of a name if we include it.

pyodide-build/pyodide_build/install_xbuildenv.py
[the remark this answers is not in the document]
hoodmane: It's not linux specific, it's an Emscripten static library. 

> `2.tar`

I was thinking about naming them by version e.g., `0.21.0.tar.gz`. But probably github releases is a good idea for the release versions.

pyodide-build/setup.py
rth: Although if you are calling it as `pywasm build` (I missed the subcommand part), build in the name would indeed be a bit redundant.

Maybe we should just call this command `pyodide` and make this the new CLI API?

hoodmane: The main distinction in my mind is between in-tree and out-of-tree commands. E.g., `mkpkg` should be a public in-tree command, whereas it makes no sense out of tree. This is an out of tree command. People are likely to only need one CLI as appropriate to whether they are doing an in tree or out of tree build. So I think it makes sense to keep these as separate CLIs.

hoodmane:
Is it okay if we fix the name of the CLI entrypoint and the `.tar.gz` extension in a subsequent PR? I think it would be easier to handle these things separately.

rth:
Yes, sure thanks! Very nice progress on this! 

In https://github.com/scikit-learn/scikit-learn/issues/23727 I was saying that it's maybe a bit early to run in CI, but it looks like it's closer than I thought thanks to all the work you did! Though having a robust runner for large test suites probably still needs some work, particularly those that can potentially segfault due to scipy.

PR #2827, opened 2022-07-02
Update Emscripten to 3.1.15

hoodmane:


hoodmane:
Looks like memory corruption in xgboost and opencv @ryanking13...

ryanking13:
Hmm, the error message is `indirect call signature mismatch`, so probably it is emitting an invalid function signature. I'll see what emscripten changed in 3.1.15 release.

[There is an error message `0 is not a valid OpCode`](https://app.circleci.com/pipelines/github/hoodmane/pyodide/3592/workflows/833bab36-5cab-4bd5-a746-7a01f02156e3/jobs/42510?invite=true#step-105-108)  when building `pyodide.asm.js`. I don't think this is related to the opencv-python and xgboost failure though.

hoodmane:
>  the error message is `indirect call signature mismatch`

I think the bad call is because of earlier memory corruption. In wasm there is no memory protection, you can dereference any number between 0 and the size of the memory (it seems to start out at 20 megabytes). So if stuff is going wrong, you don't see a memory fault. In many cases the first time that it traps is because it tries to call a corrupted function pointer and the signature doesn't match.

ryanking13:
> I think the bad call is because of earlier memory corruption. In wasm there is no memory protection,

Yeah, it seems like `__wasm_call_ctors` which is called during the instantiation of the library is corrupted in some way...

hoodmane:
Since these are both new packages, maybe we should disable them too until we figure out how to make them work with the newer Emscripten? I think it's important to keep the compiler up to date if possible.

ryanking13:
Since OpenCV is one of the most requested packages, I would like to include it in 0.21.0 if possible. Maybe we can release 0.21.0 with Emscripten 3.1.14 and then disable those packages in 0.22.0.dev (assuming that we can fix this issue until next next release...)?

hoodmane:
Sounds good.

tiran:
There have been several EMSDK releases since the PR was created. Could you try 3.1.18? It has 64bit time_t.

hoodmane:
My guess is that it will still break open-cv. But we may be planning to start only updating Python and Emscripten at the same time once a year until https://github.com/emscripten-core/emscripten/issues/15917 is resolved so maybe it would be worth disabling `open-cv` and `xgboost` and then reenabling in `0.21.1` as I suggested above. Or maybe we should wait one more Pyodide release before locking the Emscripten version to the Python version.

PR #2830, opened 2022-07-02
Use big list of naughty strings to test string conversions

hoodmane:


src/tests/test_typeconversions.py
```diff
@@ -64,6 +66,46 @@ def test_string_conversion2(selenium, s):
...
 
+def blns():
+    yield ""
+    import base64
+    import json
+    from urllib.request import urlopen
+
+    try:
+        with urlopen(
+            "https://raw.githubusercontent.com/minimaxir/big-list-of-naughty-strings/master/blns.base64.json"
```
ryanking13, line 77: The list of strings are shorter than I think, so perhaps we can embed it instead of downloading it?

ryanking13 (approved):
Thanks, @hoodmane!

src/tests/test_typeconversions.py
rth: +1 possibly under `pyodide/src/tests/` but maybe gzip-compressed, to avoid appearing in grep.

rth (approved):

hoodmane:
>  gzip-compressed

Well it's already base 64 encoded so I think it isn't a problem.

PR #2835, opened 2022-07-04
Upgrade galpy to v1.8.0, remove patches

jobovy:
### Description

I released ``galpy`` version 1.8.0 which upstreamed the patches that were necessary in #2676. So after this, maintaining ``galpy`` in ``pyodide`` should be easy. Haven't tested the build yet though, so hopefully everything will just work!


ryanking13 (approved):
Thanks you so much @jobovy!

PR #2837, opened 2022-07-04
Write built packages to a file in package build jobs

hoodmane:
This is partial work towards splitting up the package test jobs into several different pieces so that it isn't necessary to wait for scipy and open-cv to build to test packages that don't depend on numpy.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -75,7 +75,9 @@ jobs:
             source pyodide_env.sh
 
             ccache -z
-            PYODIDE_PACKAGES="core" make
+            PYODIDE_PACKAGES="core" \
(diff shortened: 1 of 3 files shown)


ryanking13:
Thanks, @hoodmane. I think this is not bad, but perhaps we can explore more about pytest cache mechanism and see what we can do with it?

Since we [skip tests for un-built packages](https://github.com/pyodide/pyodide/blob/d8ae3cd1e0d41fd95e5fe72ca7e9a0781673ba82/pyodide-test-runner/pyodide_test_runner/utils.py#L45), so maybe we can cache the skipped test lists and re-run only skipped test.

hoodmane:
Sure that sounds like an interesting idea. We can leave this open if you think there is a better way.

PR #2838, opened 2022-07-05
Add general cmake toolchain file for package build

ryanking13:
### Description

This PR adds a CMake toolchain file that can be used for building Pyodide Python packages.
I hope this would make adding a new Python package which uses CMake easier.

The content of toolchain file is mostly adapted from [the Emscripten CMake toolchain file](https://github.com/emscripten-core/emscripten/blob/main/cmake/Modules/Platform/Emscripten.cmake) with minor modifications to make it work with Pyodide build system such as:

1. Enable shared lib
2. Do not set compilers to `emcc`, `em++`, etc because we want to use `pywasmcross` compiler wrapper.


Makefile.envs
```diff
@@ -147,6 +147,8 @@ export MAIN_MODULE_CFLAGS= $(CFLAGS_BASE) \
 
 export STDLIB_MODULE_CFLAGS= $(SIDE_MODULE_CFLAGS) -I Include/ -I .
 
+export PYODIDE_CMAKE_TOOLCHAIN_FILE=$(PYODIDE_ROOT)/tools/cmake/Toolchain.cmake
```
ryanking13, line 150: This should not be named `CMAKE_TOOLCHAIN_FILE`, because we don't want to use this toolchain file for library builds like libyaml.

hoodmane (commented):
Looks good to me. Maybe @henryiii could look at it. Could you add the toolchain file to `to_copy` in `create_xbuildenv.py`?

rth:
Thanks for doing this. Agreed that it would be nice to have @henryiii 's opinion of this, but in any case if you are able to build opencv and xgboost with it, it sounds like an improvement with respect to what we had before and I would be +1 to merge it.

ryanking13:
> Thanks for doing this. Agreed that it would be nice to have @henryiii 's opinion of this, but in any case if you are able to build opencv and xgboost with it, it sounds like an improvement with respect to what we had before and I would be +1 to merge it.

I am working on a similar but slightly nicer approach for this.  I'll close this and make a separate PR.

PR #2844, opened 2022-07-06
Move invalidate_caches to the end of micropip.install

hoodmane:
I think this fixes the bug discussed in:
https://github.com/samuelcolvin/pydantic-core/pull/149

### DIFF
--- a/packages/micropip/src/micropip/_micropip.py
+++ b/packages/micropip/src/micropip/_micropip.py
@@ -522,7 +522,6 @@ async def install(
         A ``Future`` that resolves to ``None`` when all packages have been
         downloaded and installed.
     """
-    importlib.invalidate_caches()
     ctx = default_environment()
     if isinstance(requirements, str):
         requirements = [requirements]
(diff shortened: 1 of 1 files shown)


samuelcolvin:
🙏 

Thanks so much.

hoodmane:
Can you check if it works by adding:
```
import importlib;  importlib.invalidate_caches()
```
after the last `micropip.install`?

samuelcolvin:
Will do tomorrow.

ryanking13:
> > I had some slightly weird behaviour where pytest-speed wasn't being recognised as a plugin by pytest sometimes
> 
> This is a known issue with pytest plugins in Pyodide, I would really like to fix it but I haven't been able to understand the cause. The problem is that for some reason `importlib.metadata.distributions()` doesn't see the new packages sometimes. You can see this also with `micropip.list()`: on runs where it is working it will show the version of `pytest-speed` as `0.3.1`, but if it is not working it will show it as `UNKNOWN`.

I think I also had some experiments with `importlib.invalidate_caches()` from a couple of months ago and it was not very successful at that time.

The experiment was about changing `pyodide.loadedPackages` to load installed packages list from `importlib.metadata.distributions()`.

ryanking13 (approved):
Apart from my failure, invalidating caches after installation seems reasonable to me.

hoodmane:
Ideally it would be nice to have a test but I was having trouble creating one.

rth (approved):

PR #2848, opened 2022-07-07
BLD Ignore runtime lib flag on linking

ryanking13:
Setuptools adds `-R`  (runtime library) flag when `runtime_library_dirs` are specified, but `wasm-ld` does not accept that flag.


pyodide-build/pyodide_build/pywasmcross.py
```diff
@@ -301,11 +301,16 @@ def replay_genargs_handle_linker_opts(arg: str) -> str | None:
...
-        # ignore unsupported --sysroot compile argument used in conda
-        if opt.startswith("--sysroot="):
-            continue
-        if opt.startswith("--version-script="):
+
+        if opt.startswith(
+            (
+                "--sysroot=",  # ignore unsupported --sysroot compile argument used in conda
+                "--version-script=",
+                "-R",  # wasm-ld does not accept -R (runtime libraries)
```
rth, line 309: This would skip anything starting with R which can lead to hard to debug issues. Does it need to be `startswith` or can it be `opt == "-R"`? Otherwise, maybe some more specific regex would be better?

ryanking13: > This would skip anything starting with R which can lead to hard to debug issues.

Yes, that can be an issue. How about `startswith("-R/")`? Normally an absolute path is appended after `-R`. (e.g. [`-Wl,-R/root/repo/packages/.libs/lib`](https://app.circleci.com/pipelines/github/pyodide/pyodide/4442/workflows/566dde87-2e48-4e74-8774-44097091e198/jobs/50598?invite=true#step-105-182))

hoodmane:
I think we should try to start using code coverage tools on pyodide-build. It would be useful to know which packages and tests exercise certain of these branches. Probably we should comment more aggressively too since many lines of pywasmcross have no clear justification for why they exist and it's hard to guess.

ryanking13:
> I think we should try to start using code coverage tools on pyodide-build. It would be useful to know which packages and tests exercise certain of these branches. Probably we should comment more aggressively too since many lines of pywasmcross have no clear justification for why they exist and it's hard to guess.

Yes, right. We should also add some more logging to `pywasmcross` so that it can be easily debugged.

hoodmane (approved):

PR #2850, opened 2022-07-07
Add xbuildenv to github releases

hoodmane:
This adds a step that generates an xbuildenv to the circleci CI and deploys it to github releases when we make a release. I also updated the url in download_xbuildenv to point to github releases.


pyodide-build/pyodide_build/__init__.py
```diff
@@ -0,0 +1 @@
+__version__ = "0.21.0.dev0"
```
ryanking13, line 1: Could you update `setup.cfg` of pyodide-build to reference this version variable (and remove setup.cfg from `bump_version`)?

AFAIK something like this will work:

```py
# inside setup.cfg
version = attr: pyodide_build.__init__.__version__
```

ryanking13 (approved):
Thanks @hoodmane!

pyodide-build/pyodide_build/__init__.py
hoodmane: We could probably do something similar in `src/py/setup.cfg` using the version from `src/py/pyodide/__init__.py`.

rth: Lately, I have been using more and more,
```py
from importlib.metadata import version, PackageNotFoundError

try:
    __version__ = version("package-name")
except PackageNotFoundError:
    # package is not installed
    pass
```
in package `__init__`. https://github.com/pypa/setuptools_scm#retrieving-package-version-at-runtime  It means that one only gets the version if the package is installed, but people shouldn't use the package if it isn't installed anyway.

rth (approved):

pyodide-build/pyodide_build/__init__.py
hoodmane: This sounds like a good idea, maybe we can switch to using this in the future.

PR #2852, opened 2022-07-08
CI Split up package test jobs

ryanking13:
### Description

This PR splits package test in CI so that `no-numpy-dependents` packages can be tested earlier.

In detail, pytest will now save test results into its cache directory, and if `--skip-passed` option is given, it will skip previously successful tests.

@hoodmane


hoodmane (approved):
Thanks @ryanking13! This looks really helpful. Looks like this way you get package results for no-numpy-dependents after around 50 minutes rather than after 100 minutes.

ryanking13:

Issue #2853, opened 2022-07-08
Wrong folder specified in Building Custom Package instructions


viblo:
## 📚 Documentation

Was trying to build a custom package (pymunk) following the guide at https://pyodide.org/en/latest/development/new-packages.html

Under https://pyodide.org/en/latest/development/new-packages.html#building-the-package-and-investigating-issues the text says you should run `pyodide_build buildall` from the `packages/<package-name>` folder. However, when I did so I got an error about meta.yml not found. Instead I think it should be run from the src folder, just as the other commands. When I did so everything worked as expected.

hoodmane:
Presumably at some point it used to say to use `buildpkg` there instead of `buildall`.

viblo:
Ill make a PR, just wanted to verify that I understood it correctly first. Today is the first time Ive tried to do anything with Pyodide :)

PR #2854, opened 2022-07-08
Fix doc for building package with pyodide_build

viblo:
### Description

Fixes the documentation issue in #2853 
In the old instructions to build package it was written that you should run `pyodide_build buildall` from the `packages/<package-name>` folder. However, the command that comes next assumes you are in the root (like the previous commands, i.e. to create the meta.yml template). Fixed by removing the reference to the folder.

### Checklists

### DIFF
--- a/docs/development/new-packages.md
+++ b/docs/development/new-packages.md
@@ -119,7 +119,7 @@ build:
 ### 2. Building the package and investigating issues
 
 Once the `meta.yaml` file is ready, build the package with the following
-commands
+command
 
 ```sh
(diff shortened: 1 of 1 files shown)


docs/development/new-packages.md
```diff
@@ -119,7 +119,7 @@ build:
 ### 2. Building the package and investigating issues
 
 Once the `meta.yaml` file is ready, build the package with the following
-commands from inside the package directory `packages/<package-name>`
+commands
```
hoodmane, line 122:
```suggestion
command
```

PR #2858, opened 2022-07-10
Change out of tree build entrypoint name from pywasm to pyodide

hoodmane:
As recommended by @rth in a comment on #2823.


ryanking13:
Thanks, @hoodmane. I am working on #1977 and was using the name `pyodide` for it. Maybe we can think of a better name for #1977 when a minimal working outcome gets ready (I think I will be able to make a PR in a few weeks).

PR #2859, opened 2022-07-10
0.21.0a3

hoodmane:
Making another release so we can test the cross build stuff in Numpy CI against an actual release. Also, this is a good opportunity to test the changes to the deploy release job since 0.21.0a2.


rth:
Thanks also, this would include all the API deprecations in the Python package, so it's good to have an alpha with that.

hoodmane:
Yeah it's really nice that the process of making these is so simple now

ryanking13:
It seems like the deployment failed. Probably the paths here need to be limited to `./xbuildenv`:

https://github.com/pyodide/pyodide/blob/1fc2fc4256b7c12830a67e26c85f65bf9f54cdf5/.circleci/config.yml#L217-L220

PR #2862, opened 2022-07-11
CI Remove package source files after build

ryanking13:
### Description

Clean up package source files after build in order to reduce the size of data shared between CI jobs. 
This reduces the running time of CI jobs such as `test-packages-chrome`, around 3 minutes.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -165,6 +165,11 @@ jobs:
             tar --exclude="node_modules" -cjf pyodide-build.tar.gz dist
             tar cjf build-logs.tar.gz  packages/build-logs
 
+      - run:
+          name: Clean up package source files
+          command: |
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -165,6 +165,11 @@ jobs:
             tar --exclude="node_modules" -cjf pyodide-build.tar.gz dist
             tar cjf build-logs.tar.gz  packages/build-logs
 
+      - run:
+          name: Clean up package source files
+          command: |
+            cd packages && find **/build ! -name '.packaged' -type f -exec rm -f {} +
```
hoodmane, line 171: What does the ! do?

hoodmane (approved):

.circleci/config.yml
ryanking13: It means "not". It is there in order to exclude `.packaged` file from remove targets.

PR #2865, opened 2022-07-12
Astropy tests

jobovy:
### Description

I noticed that `astropy` doesn't have any tests currently, so this PR adds some. 

Because the `astropy` test suite is included in the package, I did this by running a small selection of the package's tests in pyodide. Locally using Chrome, these tests take ~90s. If that's too long, we could remove some of the slower ones. I tried to select a variety of tests that cover the sub-packages that I myself at least use a lot.

To run the package tests, it's necessary to install a few `pytest` plugins that `astropy` uses, so I installed those using `micropip` in the test (and, thus, couldn't use the `run_in_pyodide` decorator I think). Maybe there is a better way to do this? I also found that I have to install `pytest`, even though I had built it in `pyodide`, is that normal?

Not sure whether this requires a `CHANGELOG` entry?

### Checklists

- [x] Add / update tests


ryanking13:
Thanks, @jobovy!

> thus, couldn't use the `run_in_pyodide` decorator I think

Yo can use `run_in_pyodide` decorator:

```py
@pytest.mark.asyncio
@run_in_pyodide(packages=["astropy", "pytest", "micropip"])
def test_astropy(selenium, package, specific_test):
    import micropip
    import pytest
    import astropy
    await micropip.install("...")
    ...
```

> I also found that I have to install `pytest`, even though I had built it in `pyodide`, is that normal?

No, you don't need to install `pytest` with micropip. It can be loaded via `pyodide.loadPackage`. If you use `@run_in_pyodide(packages=["astropy", "pytest", "micropip"])` it will be loaded automatically.

jobovy:
Thanks for the feedback @ryanking13!

Okay, I implemented the tests in this way, which seems to work locally (except that the test function needs to be `async def test_...`). The tests now do seem noticeably slower (~140s vs. 90s before).

However, I think the CI will still fail, because these tests are not being skipped in the `no-numpy-dependents` test even though the package wasn't built (because it depends on `numpy`). I think I traced this down to an issue in the test runner, specifically here:

https://github.com/pyodide/pyodide/blob/04d18ca671fd186d5dea55e7dc9845d5e9306d81/pyodide-test-runner/pyodide_test_runner/utils.py#L57-L62

The regular expression on line 60 here doesn't catch these tests, because they are parameterized tests and thus have names like `...[chrome-constants-test_]` etc. The regular expression only catches tests like `...[chrome]`. I think this could be fixed by replacing line 60 with
```
rf"test_[\w\-]+\[({browsers})[^\]]*\]", item.name
```
that is, just catch anything that isn't a closing `]`. This works locally to skip the tests when `astropy` isn't built. 

I can open a separate PR for that change to the test runner if you think that's a good change. Otherwise we'd need another change to skip the tests in this PR when the package isn't built.

ryanking13:
> The regular expression on line 60 here doesn't catch these tests, because they are parameterized tests and thus have names like `...[chrome-constants-test_]` etc. The regular expression only catches tests like `...[chrome]`. I think this could be fixed by replacing line 60 with
> 
> ```
> rf"test_[\w\-]+\[({browsers})[^\]]*\]", item.name
> ```
> 
> that is, just catch anything that isn't a closing `]`. This works locally to skip the tests when `astropy` isn't built.
> 
> I can open a separate PR for that change to the test runner if you think that's a good change. Otherwise we'd need another change to skip the tests in this PR when the package isn't built.

Oh, yes. Thanks for catching that out! I think your fix is reasonable. Feel free to make a separate PR or you can add that fix in this PR.

jobovy:
> 
> Oh, yes. Thanks for catching that out! I think your fix is reasonable. Feel free to make a separate PR or you can add that fix in this PR.

Okay, I added it here so we can directly test that it works as intended.

jobovy:
Looks like everything passes now.

However, in a previous run, one of the core tests errored (don't know why, doesn't now) and wasn't re-run, instead there was this failure:

https://app.circleci.com/pipelines/github/pyodide/pyodide/4516/workflows/b7876b82-bff1-4442-96a8-9dab8fdc4f9b/jobs/51854?invite=true#step-102-1752

I think this is because the pytest cache directory is hardcoded here:
https://github.com/pyodide/pyodide/blob/2f1c19700765851ed208d4d0f2251003ca33c998/tools/pytest_wrapper.py#L24-L26
but in #2852 the pytest cache dir was changed for some CI runs, so this is the wrong directory to look for the `lastfailed` file.

I included a change here (https://github.com/pyodide/pyodide/pull/2865/commits/2220eda601e42454c6be38b7a154823fa6d7ba55) that uses the correct cache directory when it's different from the default. Unfortunately, I don't think any tests failed in the latest CI run, so this wasn't actually used and tested...

jobovy:
Also, comparing to a previous run, this seems to add ~1 to 2 minutes to the package tests CI run. That's not a lot compared to how long the full test suite runs, but it is a significant increase in just the `numpy-dependents` package tests (~10% change). Unfortunately, the tests that I added here that take the longest are also the most crucial to make sure `astropy` works as intended in some of its most common use cases. So overall, I think it's worth the extra time.

ryanking13:
> However, in a previous run, one of the core tests errored (don't know why, doesn't now)

It's okay. Selenium webdriver fails occasionally, it's not related to your PR.

> but in [#2852](https://github.com/pyodide/pyodide/pull/2852) the pytest cache dir was changed for some CI runs, so this is the wrong directory to look for the lastfailed file.

Right, I forgot to fix the cache directory. Thanks!

> Unfortunately, the tests that I added here that take the longest are also the most crucial to make sure astropy works as intended in some of its most common use cases. So overall, I think it's worth the extra time.

1-2 minutes of additional test time is not very small, but yes, it's worth the extra time. So I think it is Okay.

packages/astropy/test_astropy.py
```diff
@@ -0,0 +1,48 @@
...
+        ("coordinates", "test_representation.py and not test_unitphysics"),
+        ("coordinates", "test_sky_coord.py"),
+        ("coordinates", "test_sky_coord_velocities.py"),
+        ("coordinates", "test_transformations.py"),
+        ("cosmology", "test_base.py"),
+        ("cosmology", "test_lambdacdm.py"),
+        ("cosmology", "test_units.py"),
+    ],
+)
+@pytest.mark.asyncio
```
ryanking13, line 31:
```suggestion
```

I think I was mistaken. This marker is not needed.

ryanking13 (approved):
Thanks @jobovy! LGTM

packages/astropy/test_astropy.py
jobovy: Yes, it seems like this isn't necessary, I guess because of the way `run_in_pyodide` is written.

jobovy:
Exciting, one of the core tests failed in the last CI run and was then re-run correctly:

https://app.circleci.com/pipelines/github/pyodide/pyodide/4531/workflows/e3e24186-205d-4511-813d-ac82dad09ace/jobs/51982?invite=true#step-102-1972

So my change re:the cache directory seems to work.

ryanking13:
Thanks a lot @jobovy!

PR #2866, opened 2022-07-12
CI Fix xbuildenv deploy

ryanking13:


hoodmane:
Thanks @ryanking13!

hoodmane:
I guess after this we can make another 0.21.0a3 release. I think the best way is to move the tag and then force push to main. In order to do this you will have to temporarily remove the branch protection. Maybe make another commit called "v0.21.0a3 (for real this time)". And then we should revert the version numbers.

By the way, thanks again for fixing the release process! `bump_version.py` is really great!

rth:
Maybe let's not force push to main if you mean re-writing git history. Pushing another commit with this tag  -- sure.

hoodmane:
Tags have to be unique and moving the tag requires a force push. This doesn't otherwise change history. Maybe we should just skip to v0.21.0a4

PR #2868, opened 2022-07-12
Added REBOUND package.

hannorein:
### Description
This is a pull request to add the orbit integration package [REBOUND](https://github.com/hannorein/rebound) to pyodide.

### Checklists

- [x] Add / update tests

### DIFF
--- a/packages/rebound/meta.yaml
+++ b/packages/rebound/meta.yaml
@@ -1,0 +1,16 @@
+package:
+  name: rebound
+  version: 3.19.4
+build:
+  exports: requested
+source:
+  url: https://files.pythonhosted.org/packages/78/37/95ab84dfb72a37d78420ffa1ee087dfbb6f1098117435e2191f2deff9067/rebound-3.19.4.tar.gz
+  sha256: b705235c6619b4be5368faaa07a865e6d44f9856b9b31721d1032ebe0f2199ec
+test:
+  imports:
+    - rebound
+about:
+  home: https://github.com/hannorein/rebound/
+  PyPI: https://pypi.org/project/rebound
+  summary: An open-source multi-purpose N-body code
+  license: GPL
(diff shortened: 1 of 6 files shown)


packages/rebound/meta.yaml
```diff
@@ -0,0 +1,16 @@
+package:
+  name: rebound
+  version: 3.19.4
+build:
+  exports: requested
+source:
+  url: https://files.pythonhosted.org/packages/78/37/95ab84dfb72a37d78420ffa1ee087dfbb6f1098117435e2191f2deff9067/rebound-3.19.4.tar.gz
+  sha256: b705235c6619b4be5368faaa07a865e6d44f9856b9b31721d1032ebe0f2199ec
+test:
```
jobovy, line 9: Because you need `numpy` in your test, I think you technically need to add
```suggestion
requirements:
  run:
    - numpy
test:    
 ```
here (there's no way to specify that a requirement is only for testing at this time AFAIK). Otherwise, the test would not be able to run in a partial build that doesn't include `numpy`. But I think the CI tests will work either way, because they build `numpy`.

File: packages/rebound/test_rebound.py [hunk 2]
```diff
@@ -0,0 +1,20 @@
+from pyodide_test_runner import run_in_pyodide
+
+
+@run_in_pyodide(
+    packages=[
+        "rebound",
```
jobovy, line 6: You have to add `numpy` here as in
```suggestion
        "rebound","numpy",
```
```

jobovy (commented):
I don't know whether you've tested this locally, but with my suggested edits the test passes (it doesn't without the `run_in_pyodide` change).

hannorein:
Thanks! You're too fast for me!

ryanking13 (approved):
Thanks, @hannorein! LGTM, please update [the changelog](https://github.com/pyodide/pyodide/blob/main/docs/project/changelog.md?plain=1#L206-L211)

hannorein:
Working really well! Demo: http://rebound.hanno-rein.de

ryanking13:
Thanks @hannorein!

Issue #2280, opened 2022-03-17
CI Use Github Actions more


ryanking13:
We are currently using two CI services: CircleCI and Github Actions.

- CircleCI: docker-based build test and full test suites.
- Github Actions: for non-docker-based build test and linting.

Though CircleCI worked quite well for our purpose, we've also experienced some problems in using CircleCI. (e.g. limited job running time and resources, mysterious hangs, unstable build time)

The main blocker of adopting GHA was that it doesn't support partial re-runs for failed jobs, but recently (yesterday) Github announced that [they now support partial re-runs](https://github.blog/2022-03-16-save-time-partial-re-runs-github-actions/). So I think it's time to consider using GHA more for our project.

For a start, I think we can let CircleCI as is and run more jobs in GHA for a parallel with different build/test settings. For example, docker-based / non-docker-based build (same as we are doing now) and selenium / playwright testing. And if we find that one of the services is more stable and useful, we can migrate one into the other.

WDYT?

hoodmane:
One other thing I like about CircleCI is the test failure summary for example:
https://app.circleci.com/pipelines/github/hoodmane/pyodide?branch=multiple-load2&filter=all
Is there any way to give junitxml to gha and make it produce summaries like that?

ryanking13:
> One other thing I like about CircleCI is the test failure summary for example: https://app.circleci.com/pipelines/github/hoodmane/pyodide?branch=multiple-load2&filter=all Is there any way to give junitxml to gha and make it produce summaries like that?

It seems like there is third-party action which does [similar thing for GHA](https://github.com/marketplace/actions/publish-unit-test-results), but we will need some trial and errors.

hoodmane:
Sounds good. What about rerun with ssh? Is there anything like that?

ryanking13:
> Sounds good. What about rerun with ssh? Is there anything like that?

I guess [this action](https://github.com/marketplace/actions/debugging-with-tmate) will do the job?

hoodmane:
Cool =)

hoodmane:
Another setting that I like that we have with circleci is that it will trigger on my branches even if I haven't opened a PR. Presumably this is also possible to set up with github actions?

ryanking13:
> Another setting that I like that we have with circleci is that it will trigger on my branches even if I haven't opened a PR. Presumably this is also possible to set up with github actions?

[...]

PR #2869, opened 2022-07-13
CI Add test summary on GHA

ryanking13:
### Description

Adds test summary for GHA.

You can see here how it looks like: https://github.com/ryanking13/pyodide/actions/runs/2653516166

Partial work of #2280


hoodmane (approved):
Thanks @ryanking13, looks good to me.

PR #2871, opened 2022-07-14
Enable codecov

rth:
Enable codecov for better visualization of pytest-cov reports.

Codecov is normally already enable at the org level.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -35,6 +35,7 @@ jobs:
           mkdir test-results
           python3 -m pip install ./pyodide-test-runner
           python3 -m pip install -e ./pyodide-build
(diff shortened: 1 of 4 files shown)


.github/workflows/main.yml
```diff
@@ -20,6 +20,32 @@ jobs:
           python-version: 3.10.2
       - uses: pre-commit/action@v2.0.3
 
+  test-python:
+    runs-on: ubuntu-latest
+    container: pyodide/pyodide-env:20220629-py310-chrome102-firefox100
```
ryanking13, line 25: Come to think of it, does `test-python` need to be run in the docker image? We don't need browser drivers for this test.

ryanking13 (approved):
Thanks, @rth. LGTM.

Issue #2874, opened 2022-07-15
No such file or directory: .pytest_cache/v/cache/lastfailed


rth:
I have seen this error already a couple of times, if pytest fails during collection this paths doesn't exist, and we should just exit in https://github.com/pyodide/pyodide/blob/ed22c06e07d342723542961f7805951096684cda/tools/pytest_wrapper.py#L25

PR #2875, opened 2022-07-15
pytest_wrapper.py: don't rerun tests if collection fails

rth:
Closes https://github.com/pyodide/pyodide/issues/2874


ryanking13 (approved):
Thanks, @rth!

PR #2856, opened 2022-07-08
DOC update the instructions on how to use Node.js

antocuni:
Fixes #2849.

TODO:
  - try to actually build the docs (I'm committing this blindly since I didnt' find the instructions on how to build the docs locally)
  - fix `src/js/README.md`

PR #2883, opened 2022-07-18
Include README when deploying npm package

### LINKED ISSUE #2855: Upload js readme to npm
As mentioned in https://github.com/pyodide/pyodide/issues/2849#issuecomment-1178846101

> Currently we upload from dist so that README isn't ending up on npm anymore. If we want it included we should copy it into dist before the node deploy and also add it to the package.json files list:
https://github.com/pyodide/pyodide/blob/c05b76273b76be18f9ed5745e2ff06e9dfdc4135/src/js/package.json#L59

We should fix it as otherwise https://www.npmjs.com/package/pyodide/v/0.21.0-alpha.2 doesn't look very helpful compared to what we had before https://www.npmjs.com/package/pyodide/v/0.20.0
Though it does mean repeating the documentation a bit.

ryanking13:
Close #2855

### DIFF
--- a/src/js/README.md
+++ b/src/js/README.md
@@ -11,16 +11,69 @@ releases](https://github.com/pyodide/pyodide/releases)
 Then you can load Pyodide in Node.js as follows,
 
 ```js
-let pyodide_pkg = await import("pyodide/pyodide.js");
+// hello_python.js
+const { loadPyodide } = require("pyodide");
 
-let pyodide = await pyodide_pkg.loadPyodide({
-  indexURL: "<pyodide artifacts folder>",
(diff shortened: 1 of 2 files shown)


rth:
Thanks! The other issue is that https://github.com/pyodide/pyodide/tree/main/src/js readme probably needs updating to be consistent with what we put in https://github.com/pyodide/pyodide/pull/2856 as I'm not sure the current instructions work otherwise.

ryanking13:
I check that the current instruction still works. But it looks like the updated docs in #2856 seem more detailed, so I copied it to `src/js/README.md`.

PR #2887, opened 2022-07-18
Add gnu readline

hoodmane:
This adds gnu readline. This should in principle allow us to match the normal Python repl behavior (which relies on readline). We could dynamically link it instead if we want to add it to unvendored stdlib modules.

### DIFF
--- a/Makefile.envs
+++ b/Makefile.envs
@@ -111,6 +111,8 @@ export MAIN_MODULE_LDFLAGS= $(LDFLAGS_BASE) \
 	-lffi \
 	-lsqlite3 \
 	-lbz2 \
+	-lreadline \
+	-ltermcap \
 	-lstdc++ \
 	-lidbfs.js \
 	-lnodefs.js \
--- a/cpython/Makefile
+++ b/cpython/Makefile
@@ -28,7 +28,18 @@ FFIBUILD=$(ROOT)/build/libffi
 LIBFFIREPO=https://github.com/hoodmane/libffi-emscripten
 LIBFFI_TAG=2022-06-23
 
-all: $(INSTALL)/lib/$(LIB) $(INSTALL)/lib/libffi.a
+
+READLINE_NAME=readline-8.1
+READLINE_URL=https://ftp.gnu.org/gnu/readline/$(READLINE_NAME).tar.gz
+READLINE_TARBALL=$(ROOT)/downloads/$(READLINE_NAME).tar.gz
+READLINE_BUILD=$(ROOT)/build/$(READLINE_NAME)
+
+TERMCAP_NAME=termcap-1.3.1
+TERMCAP_URL=https://ftp.gnu.org/gnu/termcap/$(TERMCAP_NAME).tar.gz
+TERMCAP_TARBALL=$(ROOT)/downloads/$(TERMCAP_NAME).tar.gz
+TERMCAP_BUILD=$(ROOT)/build/$(TERMCAP_NAME)
+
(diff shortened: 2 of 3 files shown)


hoodmane:
It seems like the way that Python normally wants to use this is with select. It will require further investigation whether this is the right way to go.

tiran:
You need to build ``readline`` with ``-DNEED_EXTERN_PC``. Otherwise you'll get several duplicate symbol conflicts:

```
wasm-ld: error: duplicate symbol: PC
>>> defined in /emsdk/upstream/emscripten/cache/sysroot/lib/wasm32-emscripten/libreadline.a(terminal.o)
>>> defined in /emsdk/upstream/emscripten/cache/sysroot/lib/wasm32-emscripten/libtermcap.a(termcap.o)

wasm-ld: error: duplicate symbol: UP
>>> defined in /emsdk/upstream/emscripten/cache/sysroot/lib/wasm32-emscripten/libreadline.a(terminal.o)
>>> defined in /emsdk/upstream/emscripten/cache/sysroot/lib/wasm32-emscripten/libtermcap.a(tparam.o)

wasm-ld: error: duplicate symbol: BC
>>> defined in /emsdk/upstream/emscripten/cache/sysroot/lib/wasm32-emscripten/libreadline.a(terminal.o)
>>> defined in /emsdk/upstream/emscripten/cache/sysroot/lib/wasm32-emscripten/libtermcap.a(tparam.o)
```

tiran:
From my notes

## termcap

```
emconfigure ./configure --prefix=/opt/emsdk-extras
```

Create config.h to work around missing headers (``write`` needs ``unistd.h``)
```
cat > config.h << EOF
#define STDC_HEADERS 1
#include <unistd.h>
EOF
```

build
```
emmake make CFLAGS="-DHAVE_CONFIG_H=1 -fPIC -Wno-deprecated-non-prototype"
emmake make install
```

## readline

```
$ CPPFLAGS="-DHAVE_POSIX_SIGNALS=1 -DNEED_EXTERN_PC -I/opt/emsdk-extras/include" \
       CFLAGS="-fPIC" \
       LDFLAGS="-L/opt/emsdk-extras/lib" \
       emconfigure ./configure --disable-shared -C --prefix=/opt/emsdk-extras
$ emmake make
```

## Python

compile and link the readline module with ``-lreadline -ltermcap``

antocuni:
> This adds gnu readline. This should in principle allow us to match the normal Python repl behavior (which relies on readline). 

I'm confused: how can it possibly work?
If I understand correctly how CPython uses `libreadline`, where is basically a `while()` loop which calls `select()` to wait until there is input available on stdin, and then calls `rl_callback_read_char()` to actually read the input:
https://github.com/python/cpython/blob/286549b7cc8f3610014c3db204a693b4e219ef46/Modules/readline.c#L1325-L1370

But the call to `select()` is blocking, so if you call readline from the main thread of the browser, this will freeze the UI, won't it?

Am I missing anything?
Or maybe this is intended to be used only in node.js or webworkers?

hoodmane:
> how can it possibly work?

The short answer is it probably can't work without stack switching. With great difficulty one could set it up to run in a web worker, but not without a lot of custom junk.

hoodmane:
I did some investigation into stack switching here:
https://github.com/pyodide/pyodide/pull/3210

antocuni:
> > how can it possibly work?
> 
> The short answer is it probably can't work reasonably without stack switching.

ok, this also matches my intuition/understanding.

> With great difficulty one could set it up to run in a web worker, but not without a lot of custom junk.

what is the "lot of custom junk"?

You probably are already aware of it, but I recently discovered the [cowasm project](https://github.com/sagemathinc/cowasm/blob/main/docs/differences-from-pyodide.md#synchronous-input) which seems to do exactly that: Python+WASM+webworker+synchronous input.
But I didn't investigate the details of what they do.

PR #2889, opened 2022-07-19
Set initialization script for pyodide-test-runner

ryanking13:
Take out `initialize_global_hiwire_objects` from pytest-pyodide (pyodide-test-runner) to conftest.py.

The purpose of this PR is to loosen the dependency between pytest-pyodide and pyodide so we can update pyodide API without modifying pytest-pyodide.

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -20,22 +20,20 @@ from pyodide_test_runner.utils import parse_xfail_browsers
 # There are a bunch of global objects that occasionally enter the hiwire cache
 # but never leave. The refcount checks get angry about them if they aren't preloaded.
 # We need to go through and touch them all once to keep everything okay.
-os.environ[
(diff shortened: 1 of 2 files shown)


rth:
LGTM, thanks. Though now that the code is in a separate package, the steps would likely be first merge https://github.com/pyodide/pyodide/pull/2893, make this change in pytest-pyodide, release then apply the change in this PR.

ryanking13:
> LGTM, thanks. Though now that the code is in a separate package, the steps would likely be first merge #2893, make this change in pytest-pyodide, release then apply the change in this PR.

Yeah, I just wanted to check that it works properly :), I'll open this to pytest-pyodide too.

ryanking13:
Merging, so #2893 can use this.

PR #2892, opened 2022-07-19
add meta.yaml for z3solver to package directory

NikolajBjorner:
Signed-off-by: GitHub <noreply@github.com>

### Description

Please consider merging the following pull request.
It adds a package description for [z3-solver](https://github.com/z3prover/z3). 
I have built in for wasm in the latest pyodide docker image, so I hope it works.
We would like to use pyodide to run the python bindings to z3 in a browser to make it easier for users to 
learn and experiment with. You can find a preliminary place-holder [here](https://microsoft.github.io/z3guide/docs/Programming%20Z3/Using%20Z3%20from%20Python/Introduction)

### DIFF
--- a/packages/z3-solver/meta.yaml
+++ b/packages/z3-solver/meta.yaml
@@ -1,5 +1,5 @@
 package:
-  name: z3-solver
+  name: z3_solver
   version: 4.9.1.0
 source:
   url: https://files.pythonhosted.org/packages/91/be/d247b751fd779e2e6abb175347cf97893e8606a19deaeab4ed26581b81a0/z3-solver-4.9.1.0.tar.gz
--- a/packages/z3-solver/meta.yaml
+++ b/packages/z3-solver/meta.yaml
@@ -12,4 +12,4 @@ about:
   PyPI: https://pypi.org/project/z3-solver
   summary: an efficient SMT solver library
   license: MIT License
-  
+
--- a/packages/z3-solver/meta.yaml
+++ b/packages/z3-solver/meta.yaml
@@ -5,11 +5,10 @@ source:
   url: https://files.pythonhosted.org/packages/91/be/d247b751fd779e2e6abb175347cf97893e8606a19deaeab4ed26581b81a0/z3-solver-4.9.1.0.tar.gz
   sha256: 5ed51c9748a23b6e21af9a526fa21f7ff0a76fd696aff436676300f162e10ddc
 test:
-  imports:
+  imports:pre-commit run --all-files
     - z3-solver
 about:
   home: https://github.com/Z3Prover/z3
   PyPI: https://pypi.org/project/z3-solver
   summary: an efficient SMT solver library
   license: MIT License
-
--- a/packages/z3-solver/meta.yaml
+++ b/packages/z3-solver/meta.yaml
@@ -1,0 +1,15 @@
+package:
+  name: z3-solver
+  version: 4.9.1.0
+source:
+  url: https://files.pythonhosted.org/packages/91/be/d247b751fd779e2e6abb175347cf97893e8606a19deaeab4ed26581b81a0/z3-sol
+ver-4.9.1.0.tar.gz
+  sha256: 5ed51c9748a23b6e21af9a526fa21f7ff0a76fd696aff436676300f162e10ddc
+test:
+  imports:
+    - z3-solver
+about:
+  home: https://github.com/Z3Prover/z3
+  PyPI: https://pypi.org/project/z3-solver
+  summary: an efficient SMT solver library
+  license: MIT License
--- a/packages/z3-solver/meta.yaml
+++ b/packages/z3-solver/meta.yaml
@@ -5,7 +5,7 @@ source:
   url: https://files.pythonhosted.org/packages/91/be/d247b751fd779e2e6abb175347cf97893e8606a19deaeab4ed26581b81a0/z3-solver-4.9.1.0.tar.gz
   sha256: 5ed51c9748a23b6e21af9a526fa21f7ff0a76fd696aff436676300f162e10ddc
 test:
-  imports:pre-commit run --all-files
+  imports:
     - z3-solver
 about:
   home: https://github.com/Z3Prover/z3
--- a/packages/z3-solver/meta.yaml
+++ b/packages/z3-solver/meta.yaml
@@ -2,8 +2,7 @@ package:
   name: z3-solver
   version: 4.9.1.0
 source:
-  url: https://files.pythonhosted.org/packages/91/be/d247b751fd779e2e6abb175347cf97893e8606a19deaeab4ed26581b81a0/z3-sol
-ver-4.9.1.0.tar.gz
+  url: https://files.pythonhosted.org/packages/91/be/d247b751fd779e2e6abb175347cf97893e8606a19deaeab4ed26581b81a0/z3-solver-4.9.1.0.tar.gz
   sha256: 5ed51c9748a23b6e21af9a526fa21f7ff0a76fd696aff436676300f162e10ddc
 test:
   imports:
@@ -13,3 +12,4 @@ about:
   PyPI: https://pypi.org/project/z3-solver
   summary: an efficient SMT solver library
   license: MIT License
+  
--- a/packages/z3solver/meta.yaml
+++ b/packages/z3solver/meta.yaml
@@ -1,5 +1,5 @@
 package:
-  name: z3_solver
+  name: z3solver
   version: 4.9.1.0
 source:
   url: https://files.pythonhosted.org/packages/91/be/d247b751fd779e2e6abb175347cf97893e8606a19deaeab4ed26581b81a0/z3-solver-4.9.1.0.tar.gz
--- a/packages/z3solver/meta.yaml
+++ b/packages/z3solver/meta.yaml
@@ -4,8 +4,11 @@ package:
 source:
   url: https://files.pythonhosted.org/packages/91/be/d247b751fd779e2e6abb175347cf97893e8606a19deaeab4ed26581b81a0/z3-solver-4.9.1.0.tar.gz
   sha256: 5ed51c9748a23b6e21af9a526fa21f7ff0a76fd696aff436676300f162e10ddc
+requirements:
+  run:
+    - setuptools
 build:
-  ldflags: -s SIDE_MODULE=1
+  exports: whole_archive
 test:
   imports:
     - z3
--- a/packages/z3solver/meta.yaml
+++ b/packages/z3solver/meta.yaml
@@ -6,7 +6,7 @@ source:
   sha256: 5ed51c9748a23b6e21af9a526fa21f7ff0a76fd696aff436676300f162e10ddc
 test:
   imports:
-    - z3-solver
+    - z3solver
 about:
   home: https://github.com/Z3Prover/z3
   PyPI: https://pypi.org/project/z3-solver
--- a/packages/z3solver/meta.yaml
+++ b/packages/z3solver/meta.yaml
@@ -4,6 +4,8 @@ package:
 source:
   url: https://files.pythonhosted.org/packages/91/be/d247b751fd779e2e6abb175347cf97893e8606a19deaeab4ed26581b81a0/z3-solver-4.9.1.0.tar.gz
   sha256: 5ed51c9748a23b6e21af9a526fa21f7ff0a76fd696aff436676300f162e10ddc
+build:
+  ldflags: -s SIDE_MODULE=1
 test:
   imports:
     - z3
--- a/packages/z3solver/meta.yaml
+++ b/packages/z3solver/meta.yaml
@@ -6,7 +6,7 @@ source:
   sha256: 5ed51c9748a23b6e21af9a526fa21f7ff0a76fd696aff436676300f162e10ddc
 test:
   imports:
-    - z3solver
+    - z3
 about:
   home: https://github.com/Z3Prover/z3
   PyPI: https://pypi.org/project/z3-solver
--- a/packages/z3solver/test_z3solver.py
+++ b/packages/z3solver/test_z3solver.py
@@ -8,6 +8,7 @@ def test_z3_socrates(selenium):
     """test to see if z3 is alive"""
 
     import z3
+    
     obj = z3.DeclareSort("Object")
     human = z3.Function("Human", obj, z3.BoolSort())
     mortal = z3.Function("Mortal", obj, z3.BoolSort())
--- a/packages/z3solver/test_z3solver.py
+++ b/packages/z3solver/test_z3solver.py
@@ -5,30 +5,19 @@ from pyodide_test_runner import run_in_pyodide
     packages=["z3solver"],
 )
 def test_z3_socrates(selenium):
-    import io
-
-    from z3 import *
-
-    Object = DeclareSort('Object')
-    Human = Function('Human', Object, BoolSort())
-    Mortal = Function('Mortal', Object, BoolSort())
-
+    """ test to see if z3 is alive """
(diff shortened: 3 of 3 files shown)


NikolajBjorner:
From what I can tell the current build breaks appear to be external.

packages/z3-solver/meta.yaml
```diff
@@ -0,0 +1,14 @@
+package:
+  name: z3_solver
```
jobovy, line 2: I believe ``pyodide-build`` assumes that the directory name matches the package name and that's why the build package test fails [here](https://app.circleci.com/pipelines/github/pyodide/pyodide/4567/workflows/a43b1b5a-1ea3-48a3-8f5d-48b2eea27a82/jobs/52486?invite=true#step-105-56). So this should be ``z3-solver`` or the directory name should be ``z3_solver``.

I think the other test failure is indeed unrelated.

ryanking13, line 2:
```suggestion
  name: z3-solver
```

packages/z3-solver/meta.yaml
```diff
@@ -0,0 +1,14 @@
...
+  version: 4.9.1.0
+source:
+  url: https://files.pythonhosted.org/packages/91/be/d247b751fd779e2e6abb175347cf97893e8606a19deaeab4ed26581b81a0/z3-solver-4.9.1.0.tar.gz
+  sha256: 5ed51c9748a23b6e21af9a526fa21f7ff0a76fd696aff436676300f162e10ddc
+test:
+  imports:
+    - z3-solver
```
ryanking13, line 9:
```suggestion
    - z3_solver
```

packages/z3-solver/meta.yaml
ryanking13: > I believe `pyodide-build` assumes that the directory name matches the package name and that's why the build package test fails [here](https://app.circleci.com/pipelines/github/pyodide/pyodide/4567/workflows/a43b1b5a-1ea3-48a3-8f5d-48b2eea27a82/jobs/52486?invite=true#step-105-56). So this should be `z3-solver` or the directory name should be `z3_solver`.

Yes, that's right.

ryanking13:
Looks like it builds without error, which is quite promising :)

Please add some test so that we can verify it is working correctly. You can create a `test_z3solver.py`in the same directory where meta.yaml file is in. See [test_pillow.py](https://github.com/pyodide/pyodide/blob/main/packages/Pillow/test_pillow.py) for an example.

packages/z3solver/meta.yaml
```diff
@@ -0,0 +1,14 @@
+package:
+  name: z3solver
+  version: 4.9.1.0
+source:
+  url: https://files.pythonhosted.org/packages/91/be/d247b751fd779e2e6abb175347cf97893e8606a19deaeab4ed26581b81a0/z3-solver-4.9.1.0.tar.gz
+  sha256: 5ed51c9748a23b6e21af9a526fa21f7ff0a76fd696aff436676300f162e10ddc
+test:
+  imports:
+    - z3solver
```
jobovy, line 9: The previous build didn't actually pass the import test. This needs to be
```suggestion
    - z3
  ```

File: packages/z3solver/test_z3solver.py [hunk 4]
```diff
@@ -0,0 +1,34 @@
...
+
+    s = Solver()
+    s.add(axioms)
+
+    print(s.check()) # prints sat so axioms are coherent
+
+    # classical refutation
+    s.add(Not(Mortal(socrates)))
+
+    print(s.check()) # prints unsat so socrates is Mortal
```
jobovy, line 32: I think this test should have some ``assert`` statements to make sure the code is behaving as you expect. Currently it just tests whether it runs.

jobovy:
pre-commit isn't able to push lint changes to this PR because you are using an organization. So the easiest thing to do would be for you to install pre-commit and the hooks: https://pre-commit.com/ and then it will run before you push commits.

Current diff required is
```
diff --git a/packages/z3solver/test_z3solver.py b/packages/z3solver/test_z3solver.py
index a22c209d..139db7f1 100644
--- a/packages/z3solver/test_z3solver.py
+++ b/packages/z3solver/test_z3solver.py
@@ -8,6 +8,7 @@ def test_z3_socrates(selenium):
     """test to see if z3 is alive"""
 
     import z3
+
     obj = z3.DeclareSort("Object")
     human = z3.Function("Human", obj, z3.BoolSort())
     mortal = z3.Function("Mortal", obj, z3.BoolSort())
@@ -18,7 +19,7 @@ def test_z3_socrates(selenium):
     axioms = [z3.ForAll([x], z3.Implies(human(x), mortal(x))), human(socrates)]
     s = z3.Solver()
     s.add(axioms)
-    print(s.check())  
+    print(s.check())
     # prints sat so axioms are coherent
     assert z3.sat == s.check()
     # classical refutation
@@ -26,5 +27,5 @@ def test_z3_socrates(selenium):
     is_sat = s.check()
     print(is_sat)
     assert is_sat == z3.unsat
-    # prints unsat so socrates is Mortal    
+    # prints unsat so socrates is Mortal
     print(is_sat)
```

NikolajBjorner:
thanks for hint about pre-commit. I ran across it somewhere earlier today but didn't make the association (I am weak on using tools).

NikolajBjorner:
Note
```
 pyodide_test_runner.browser.JavascriptException: PythonError: Traceback (most recent call last):
E     File "<exec>", line 13, in __tmp
E     File "/root/repo/packages/z3solver/test_z3solver.py", line 10, in test_z3_socrates
E     File "/lib/python3.10/site-packages/z3/__init__.py", line 1, in <module>
E       from .z3 import *
E     File "/lib/python3.10/site-packages/z3/z3.py", line 46, in <module>
E       from . import z3core
E     File "/lib/python3.10/site-packages/z3/z3core.py", line 67, in <module>
E       raise Z3Exception("libz3.%s not found." % _ext)
E   z3.z3types.Z3Exception: libz3.so not found.
```

Looks like I need to patch z3 (my working guess, it uses a different suffix, such as "a" instead of "so", if anybody see I am misguided appreciate the heads up)

ryanking13:
Try adding this line inside meta.yaml.

```yaml
build:
  ldflags: -s SIDE_MODULE=1
```

NikolajBjorner:
appreciated, giving it a spin.
Also possible that setup.py isn't prepared correctly 
https://github.com/Z3Prover/z3/blob/2c8df54b70f5456efd0b2c3f4917d2b314098f19/src/api/python/setup.py#L53
Pardon the fumbling.

ryanking13:
> appreciated, giving it a spin. Also possible that setup.py isn't prepared correctly https://github.com/Z3Prover/z3/blob/2c8df54b70f5456efd0b2c3f4917d2b314098f19/src/api/python/setup.py#L53

I think that line is okay. When I looked inside the built wheel file, it contains `libz3.so` correctly, though its content is not a shared library that we want.

Apart from that, I think it would be helpful if that setup.py provide some ways to specify CMake args using env variable or commandline flags. It looks like the [CMake args are hardcoded inside setup.py](https://github.com/Z3Prover/z3/blob/2c8df54b70f5456efd0b2c3f4917d2b314098f19/src/api/python/setup.py#L105-L135) and I can't see a way to override it. Very often we need to modify some CMake args to make it work in browser environments (e.g. disable threading)

NikolajBjorner:
> disable threading

Seems indeed to be an astute observation. The local WASM / JS / TS build sets Z3_SINGLE_THREADED=True

NikolajBjorner:
It might take some overall tax on the build time, what seems to be about 5 threads.

```
[73/88] (thread 2) built ffmpeg in 616.90 s
[53/88] (thread 1) built z3solver in 1367.17 s

===================================================
built all packages in 1965.97 s
```

rth:
I wonder if we should build this out of tree instead. So that you would have the wheel in the z3 CI and get a wheel installable by micropip. We likely need to update the documentation for that use case anyway.
Because yes, 22min build time for a specialized package is reaching a bit the limit of what we can do with the current build system.

hoodmane:
Rather than setting `-sSIDE_MODULE=1` it would be better to use `exports: requested` or `exports: whole_archive`.

ryanking13:
> I wonder if we should build this out of tree instead. So that you would have the wheel in the z3 CI and get a wheel installable by micropip. We likely need to update the documentation for that use case anyway. Because yes, 22min build time for a specialized package is reaching a bit the limit of what we can do with the current build system.

I think there should be some room to optimize the build time of this library.

For example, it is building system libraries redundantly due to link time optimization.

```
6015:cache:INFO: generating system library: sysroot/lib/wasm32-emscripten/lto/libGL.a... (this will be cached in "/root/repo/emsdk/emsdk/upstream/emscripten/cache/sysroot/lib/wasm32-emscripten/lto/libGL.a" for subsequent builds)
6018:cache:INFO: generating system library: sysroot/lib/wasm32-emscripten/lto/libal.a... (this will be cached in "/root/repo/emsdk/emsdk/upstream/emscripten/cache/sysroot/lib/wasm32-emscripten/lto/libal.a" for subsequent builds)
6020:cache:INFO: generating system library: sysroot/lib/wasm32-emscripten/lto/libhtml5.a... (this will be cached in "/root/repo/emsdk/emsdk/upstream/emscripten/cache/sysroot/lib/wasm32-emscripten/lto/libhtml5.a" for subsequent builds)
6022:cache:INFO: generating system library: sysroot/lib/wasm32-emscripten/lto/libstubs.a... (this will be cached in "/root/repo/emsdk/emsdk/upstream/emscripten/cache/sysroot/lib/wasm32-emscripten/lto/libstubs.a" for subsequent builds)
6024:cache:INFO: generating system library: sysroot/lib/wasm32-emscripten/lto/libnoexit.a... (this will be cached in "/root/repo/emsdk/emsdk/upstream/emscripten/cache/sysroot/lib/wasm32-emscripten/lto/libnoexit.a" for subsequent builds)
6026:cache:INFO: generating system library: sysroot/lib/wasm32-emscripten/lto/libc.a... (this will be cached in "/root/repo/emsdk/emsdk/upstream/emscripten/cache/sysroot/lib/wasm32-emscripten/lto/libc.a" for subsequent builds)
6028:cache:INFO: generating system library: sysroot/lib/wasm32-emscripten/lto/libdlmalloc.a... (this will be cached in "/root/repo/emsdk/emsdk/upstream/emscripten/cache/sysroot/lib/wasm32-emscripten/lto/libdlmalloc.a" for subsequent builds)
6030:cache:INFO: generating system library: sysroot/lib/wasm32-emscripten/lto/libcompiler_rt.a... (this will be cached in "/root/repo/emsdk/emsdk/upstream/emscripten/cache/sysroot/lib/wasm32-emscripten/lto/libcompiler_rt.a" for subsequent builds)
6033:cache:INFO: generating system library: sysroot/lib/wasm32-emscripten/lto/libc++-noexcept.a... (this will be cached in "/root/repo/emsdk/emsdk/upstream/emscripten/cache/sysroot/lib/wasm32-emscripten/lto/libc++-noexcept.a" for subsequent builds)
6035:cache:INFO: generating system library: sysroot/lib/wasm32-emscripten/lto/libc++abi-noexcept.a... (this will be cached in "/root/repo/emsdk/emsdk/upstream/emscripten/cache/sysroot/lib/wasm32-emscripten/lto/libc++abi-noexcept.a" for subsequent builds)
6037:cache:INFO: generating system library: sysroot/lib/wasm32-emscripten/lto/libsockets.a... (this will be cached in "/root/repo/emsdk/emsdk/upstream/emscripten/cache/sysroot/lib/wasm32-emscripten/lto/libsockets.a" for subsequent builds)
```

Anyway yes, out of tree build (at least for large packages) should be our midterm goal.

hoodmane:
Well I think this package would also be completely straight forward to build out of tree so there isn't much reason not to do it.

hoodmane:
I guess we still need a deployment solution for out of tree builds though...

packages/z3solver/meta.yaml
```diff
@@ -0,0 +1,16 @@
+package:
+  name: z3solver
+  version: 4.9.1.0
+source:
+  url: https://files.pythonhosted.org/packages/91/be/d247b751fd779e2e6abb175347cf97893e8606a19deaeab4ed26581b81a0/z3-solver-4.9.1.0.tar.gz
+  sha256: 5ed51c9748a23b6e21af9a526fa21f7ff0a76fd696aff436676300f162e10ddc
+build:
+  ldflags: -s SIDE_MODULE=1
```
jobovy, line 8: Your build has bigger problems, but to avoid the error about ``pkg_resources`` being unavailable in the import test, you have to add
```
requirements:
  run:
    - setuptools
  ```
and any other packages that are required to import the package.

**Comment by ryanking13** [comment]
Okay, I looked into the build log and the problem seems that we strip `-shared` flag which makes `emcc` create an invalid object file. (created a PR that fixes it #2897)

> em++: warning: object file output extension (.so) used for non-object output.  If you meant to build an object file please use `-c,` `-r`, or `-shared` [-Wemcc]

**Comment by ryanking13** [comment]
> 3\. The docker with pyodide/pyodide doesn't have the pyodide-test-runner. OTOH, it builds z3solver according to the instructions.
>     When cloning the fork locally to get test-runner, the build stops working.
> 
> Getting diagnostics from the failed CI appears as the main approach for a novice (=me) who is prone to get small details wrong when following the online instructions.

Could you try [run_docker](https://github.com/pyodide/pyodide/blob/main/run_docker) script instead? If you are on Windows you'll need to use WSL.

**Comment by NikolajBjorner** [comment]
It is difficult to express how much I appreciate getting the hint "docker run -it -v%cd%:/src pyodide/pyodide-env:20220629-py310-chrome102-firefox100 /bin/bash". The fact that docker desktop starts with /bin/sh by default is not helpful and it doesn't exactly tell you how to change the shell.

Now still, I get the same error on both docker and with the git container build: 

ImportError while loading conftest '/src/conftest.py'.
conftest.py:15: in <module>
    from pyodide_test_runner.utils import maybe_skip_test
E   ModuleNotFoundError: No module named 'pyodide_test_runner'

So still updating the pull request online and having the CI go through it remains the main option to get past this.
The build on docker/windows also takes ages (as the documentation warns about for various non-native platforms).

**Comment by hoodmane** [comment]
I think you may just need to say `pip install pyodide-test-runner`.

**Comment by NikolajBjorner** [comment]
Pretty sure I even tried this before posting the latest nudnik message. 
While it sounds very reasonable, I don't think it works.


```
root@91dee40cc444:/src# pip install pyodide-test-runner
ERROR: Could not find a version that satisfies the requirement pyodide-test-runner (from versions: none)
ERROR: No matching distribution found for pyodide-test-runner
WARNING: You are using pip version 21.2.4; however, version 22.2 is available.
You should consider upgrading via the '/usr/local/bin/python -m pip install --upgrade pip' command.
root@91dee40cc444:/src# pip install pyodide_test_runner
ERROR: Could not find a version that satisfies the requirement pyodide_test_runner (from versions: none)
ERROR: No matching distribution found for pyodide_test_runner
WARNING: You are using pip version 21.2.4; however, version 22.2 is available.
You should consider upgrading via the '/usr/local/bin/python -m pip install --upgrade pip' command.
```

hoodmane:
What about `pip install ./pyodide-test-runner`?

NikolajBjorner:
Thanks for all the help. Since the original pull request, I tried different environments, the latest was using Ubuntu 20 cloud VM to reproduce build steps to avoid high overhead with Docker on Windows.

Each case requires more tinkering and insights into dependencies and build than I can undertake on my own.
If the process for integrating native code for pyodide evolves or if I am able to find someone willing and competent to pull through the integration a pyodide package would be fantastic and would have a path towards integration with online tutorials using WASM runtime for browser local experiences.

Issue #2878, opened 2022-07-17
pyodide-test-runner naming


rth:
Currently, we have the `pyodide-test-runner` package which exposes both the tools to run Pyodide in Node and browsers from host Python and pytest fixtures. Other  relevant use case is running Pyodide from Python are,
 -  sandboxed code execution
 -  CLI applications using Pyodide, possibly related to what Hood is experimenting with
 
we essentially already have the code to make it possible and I wonder if it would be better to have,
 a) a pyodide-test-runner package (current situation) and tell people to either use this or create their own if they want to run in WASM from Python
 b) have a separate 
    - `pyodide-runner` package to run code in Pyodide from Python + possible some CLI related tools
    - `pytest-pyodide` package (following the convention for pytest plugins) with the pytest related hooks and fixtures (which would depend on `pyodide-runner`)

No strong opinion. WDYT @hoodmane @ryanking13 ?

In any case, we should decide this before we upload the package to PyPi https://github.com/pyodide/pyodide/issues/2741

ryanking13:
I'm +1 for renaming `pyodide-test-runner` to `pytest-pyodide` as it is mostly a pytest plugin.

About `pyodide-runner`, I am not sure there are people who use Pyodide inside Python except for testing, maybe Hood knows better about this. Anyway, I'm +1 for having a separate package `pyodide-runner` if there are users who benefit from it.

rth:
OK, I created https://github.com/pyodide/pytest-pyodide I'll let you push the current contents of https://github.com/pyodide/pyodide/tree/main/pyodide-test-runner there (since you did most of the work on it) and then we can rename it, remove it from this repo and update the docs in follow up PRs.

rth:
About, `pyodide-runner` yeah let's way until there is a clear use case.

rth:
> I'll let you push the current contents

Actually, since I have time to work on this now, I can do it and put you as the commit author.

ryanking13:
> Actually, since I have time to work on this now,

Thanks for working on that! I think the top priority is to take off [`initialize_global_hiwire_objects`](https://github.com/pyodide/pyodide/blob/be1b3c2a6f3ae869300b5630f523dbfb15e3d2e3/pyodide-test-runner/pyodide_test_runner/browser.py#L174-L195) from the test package so that we can update Pyodide API without modifying the `pytest-pyodide`.

ryanking13:
[...]

PR #2893, opened 2022-07-19
MAINT Switch to pytest-pyodide

rth:
Use https://github.com/pyodide/pytest-pyodide instead of the repo included in this file.

The constraint this adds is that if some changes are necessary to that package one would need to make a PR to pytest-pyodide first.

Closes https://github.com/pyodide/pyodide/issues/2878 
Closes https://github.com/pyodide/pyodide/issues/2741


ryanking13:
```
INTERNALERROR>     item.config.option.runtime != "host" and "runtime" not in item.fixturenames
INTERNALERROR> AttributeError: 'DoctestItem' object has no attribute 'fixturenames'
```

Oops...

ryanking13:
It seems like the test `pyodide-build/pyodide_build/tests/test_pywasmcross.py::exports_node FAILED [ 96%]` was not running previously because its name contains "node" so it was filtered by -k "not node".

rth:
Yeah, those are tricky. Maybe let's xfail it for now and fix separately.

ryanking13:
> Yeah, those selectors are tricky. Maybe let's xfail it for now and fix it separately.

Yes, let's do it that way.

ryanking13:
Thanks, @rth!

hoodmane:
Do we need to update the instructions for how to test Pyodide/set up the development environment? People have been confused about this lately. Presumably the first time you use the docker environment you need `pip install pytest-pyodide-git-url`?

ryanking13:
> Do we need to update the instructions for how to test Pyodide/set up the development environment? People have been confused about this lately. Presumably the first time you use the docker environment you need `pip install pytest-pyodide-git-url`?

Yes, we should now mention in docs that `pytest-pyodide` is a test dependency.

Issue #2879, opened 2022-07-17
A single Pyodide CLI entry-point


rth:
So currently we have the `pyodide` CLI entry point (defined in pyodide-build)  and the `pyodide-build` entry point (also defined in pyodide-build). I think ideally we want a single `pyodide` CLI entry point with multiple subcommands (possibly nested). 

We don't need to fix this now (related to https://github.com/pyodide/pyodide/issues/2741), but IMO before the 0.21 release,
 - we need to have a clear vision of how this is going to evolve and which CLI we are going to keep in the end
 - where this main CLI entrypoint needs to be defined. The problem of defining the main `pyodide` entry point in `pyodide-build` means that we can't use it for other things (e.g. testing). So maybe it would be better to have a standalone `pyodide-cli` package defining the entry point (outside of the mono-repo, and pyodide-build would depend on it), and then other packages can register sub-commands there.

ryanking13:
> I think ideally we want a single `pyodide` CLI entry point with multiple subcommands (possibly nested).

Yes, that would be ideal. Actually, #1977 was about this topic I think.

> * So maybe it would be better to have a standalone `pyodide-cli` package defining the entry point

https://github.com/ryanking13/pyodide/pull/14 is a WIP job that I am working on for #1977. It adds a `pyodide-cli` package which can be used for a [single entry point for Pyodide](https://github.com/ryanking13/pyodide/runs/7243933919?check_suite_focus=true#step:7:93).

> which CLI we are going to keep in the end

Basically, I think `pyodide-cli` needs to be a replacement for `make` and `pyodide-build`. So in the end, `pyodide-build` CLI can be removed IMO.

rth:
Yeah, https://github.com/pyodide/pyodide/issues/1977 is the more general discussion about this (which we don't need to address straight await). This issue is for a small subset of that (i.e. naming) that I think we should do before the current `pyodide` entry point is released in 0.21 as it will be more confusing to change an already existing command later.

> https://github.com/ryanking13/pyodide/pull/14 is a WIP job that I am working on for https://github.com/pyodide/pyodide/issues/1977.

Great, should we create a separate repo `pyodide/pyodide-cli` for it? I was thinking we could,
 1. create pyodide/pyodide-cli with the minimal package setup (not implementing all the actual commands), the `pyodide` entrypoint, and release it to PyPi via GHA.
[...]

PR #2896, opened 2022-07-21
Rename out of tree CLI entrypoint

ryanking13:
Related: #2879

This PR renames the `pyodide` CLI entrypoint used for out-of-tree build to `pyodide-xbuild` (no strong opinion, please suggest a better name), so we can reserve the `pyodide` entrypoint for the future.

@hoodmane

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -63,7 +63,7 @@ autosummary_generate = True
 autodoc_default_flags = ["members", "inherited-members"]
 
 # Add modules to be mocked.
-mock_modules = ["ruamel.yaml", "tomli"]
+mock_modules = ["ruamel.yaml", "tomli", "build"]
 
 # Add any paths that contain templates here, relative to this directory.
 templates_path = ["_templates"]
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -63,7 +63,7 @@ autosummary_generate = True
 autodoc_default_flags = ["members", "inherited-members"]
 
 # Add modules to be mocked.
-mock_modules = ["ruamel.yaml", "tomli", "build"]
+mock_modules = ["ruamel.yaml", "tomli", "build", "build.__main__", "build.env"]
 
 # Add any paths that contain templates here, relative to this directory.
 templates_path = ["_templates"]
--- a/pyodide-build/pyodide_build/__main__.py
+++ b/pyodide-build/pyodide_build/__main__.py
@@ -4,12 +4,13 @@ import sys
 
 from . import buildall, buildpkg, create_xbuildenv, install_xbuildenv, mkpkg, serve
 from .common import init_environment
+from .out_of_tree import __main__ as out_of_tree
 
 
 def make_parser() -> argparse.ArgumentParser:
     """Create an argument parser with argparse"""
 
-    main_parser = argparse.ArgumentParser(prog="pyodide")
+    main_parser = argparse.ArgumentParser(prog="pyodide-build")
     main_parser.description = "A command line interface (CLI) for pyodide_build"
     subparsers = main_parser.add_subparsers(help="action")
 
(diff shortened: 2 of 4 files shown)


hoodmane:
Maybe keep it as `pyodide` but make an `xbuild` subcommand? I thought the plan was to have `pyodide` be the entrypoint for everything but then have various subcommands.

ryanking13:
Perhaps then use `pyodide-build xbuild`  for now? I think `pyodide` entry point will be defined in a new package `pyodide-cli` and `pyodide-build` will be plugged into it, so we can separate a package that defines CLI interface and a package that does the build job.

hoodmane:
Well I guess xbuild is also not a descriptive name since they are all cross builds. I guess we could call it `pyodide-build out-of-tree` and that would be more accurate.

ryanking13:
> Well I guess xbuild is also not a descriptive name since they are all cross builds. I guess we could call it `pyodide-build out-of-tree` and that would be more accurate.

Okay, let's go with `pyodide-build out-of-tree` (or `out_of_tree` for consistency with other commands).

rth:
Can we hold this for a day or two? Thinking about a better name and experimenting with a single entry point.

jobovy:
Just the two cents of an outsider, but it seems weird to me to overly focus on 'out-of-tree', because long-term presumably almost all packages will be built 'out of tree' and that will be the normal build process (as it is for regular Python packages). So I would assume that 'out-of-tree' will be the normal 'build-package' going forward and reserve a more specialized name for 'in-tree' builds, which would only happen in this repository.

hoodmane:
Well I think the goal here is to reserve the better Pyodide namespace for later. My thought with using the `pyodide` top level command for this is that `pyodide-build` can continue to be more or less private to the project and `pyodide` can be public. This is more or less holding over the public namespace for later. But until we fix the name it probably isn't wise to start adding this to the CI of many different packages because that will create a large amount of work if we decide to rename in the future.

jobovy:
> But until we fix the name it probably isn't wise to start adding this to the CI of many different packages because that will create a large amount of work if we decide to rename in the future.

Right. Once you release the 'out-of'tree' capability, people will probably start using this (I'm planning to), so you should probably figure out the public API before releasing that if you can.

hoodmane:
> you should probably figure out the public API before releasing that

Yeah I agree it needs to be a release blocker. Thanks for your feedback @jobovy!

ryanking13:
> Can we hold this for a day or two? Thinking about a better name and experimenting with a single entry point.

Sure, I opened a draft PR for a single entry point (#2904)

> Right. Once you release the 'out-of'tree' capability, people will probably start using this (I'm planning to), so you should probably figure out the public API before releasing that if you can.

Right, I think we should be careful about releasing this capability in public. @hoodmane do you think we should document `create_xbuildenv`, `install_xbuildenv` in release `0.21.0`? Otherwise, we can keep it in private until it becomes more mature.

hoodmane:
I intend for both `create_xbuildenv` and `install_xbuildenv` to remain private, at least until there is need for them to be public. I think any modification to the cross build environment can be done in tree because creating it involves multiple packages which depend on each other. The public out of tree build entry point can automatically download and install the cross build environment as needed.

ryanking13:
> I intend for both `create_xbuildenv` and `install_xbuildenv` to remain private, at least until there is need for them to be public. I think any modification to the cross build environment can be done in tree because creating it involves multiple packages which depend on each other. The public out of tree build entry point can automatically download and install the cross build environment as needed.

Sounds good to me. In that case, probably it would be better to hide them [from the docs](https://pyodide.org/en/latest/usage/api/pyodide-build-cli.html).

rth:
I proposed a few names in https://github.com/pyodide/pyodide/issues/2879#issuecomment-1195088519.

I think once we agree on the name, for this release it doesn't matter if the pyodide entrypoint is defined in pyodide-build (and not in a separate package which would take more work). As long as the name is forward compatible how it is implemented should be more or less transparent to users.

ryanking13:
> I think once we agree on the name, for this release it doesn't matter if the pyodide entrypoint is defined in pyodide-build (and not in a separate package which would take more work). As long as the name is forward compatible how it is implemented should be more or less transparent to users.

That sounds reasonable to me.

ryanking13:
Closing as this is not required for 0.21.0 release.

PR #2897, opened 2022-07-21
Do not strip -shared flag

ryanking13:
I think it is no more required and is malfunctioning in #2892

### DIFF
--- a/pyodide-build/pyodide_build/pywasmcross.py
+++ b/pyodide-build/pyodide_build/pywasmcross.py
@@ -342,9 +342,6 @@ def replay_genargs_handle_argument(arg: str) -> str | None:
 
     # fmt: off
     if arg in [
-        # don't use -shared, SIDE_MODULE is already used
-        # and -shared breaks it
-        "-shared",
         # threading is disabled for now
         "-pthread",
         # this only applies to compiling fortran code, but we already f2c'd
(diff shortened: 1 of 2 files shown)


hoodmane:
But if emscripten sees the shared flag all it does is strip it and print a warning that it is ignoring it right?

hoodmane:
Though if it doesn't cause extra warnings then this is an improvement since it reduces the amount of logic in pywasmcross.

ryanking13:
> But if emscripten sees the shared flag all it does is strip it and print a warning that it is ignoring it right?

It seems like the behavior of emcc [differs when there exists a `-shared` flag or not](https://github.com/emscripten-core/emscripten/blob/d947b506149902497628c3114dbb89e50ee12c0a/emcc.py#L1669-L1679), which is what I think is happening in #2892. `libz3.so` needs to be an object file, but it currently emits JavaScript code since we strip `-shared` flag.

```
- The default output format is now executable JavaScript.  Previously we would
  default to output objecting files unless, for example, the output name ended
  in `.js`.  This is contrary to behaviour of clang and gcc.  Now emscripten
  will always produce and executable unless the `-c`, `-r` or `-shared` flags
  are given.  This is true even when the name of the output file ends in `.o`.
  e.g, `emcc foo.c -o foo.o` will produce a JavaScript file called `foo.o`.
  This might surprise some users (although it matches the behavior of existing
  toolchains) so we now produce a warning in this case.
```

tiran:
Python 3.11 builds shared extensions with ``-shared -sSIDE_MODULE=1``. It works fine with Emscripten 3.1.x.

PR #2899, opened 2022-07-22
Remove pymalloc check

ryanking13:
This check was added 4 years ago. I think it was dealing with [Python built with `--with-pymalloc` case](https://peps.python.org/pep-3149/), which has been removed [since Python3.8](https://docs.python.org/3/whatsnew/3.8.html#build-and-c-api-changes).

I'm not sure what the comment is about.


hoodmane (approved):
I think the comment reflects a misunderstanding of what the m means (multithreaded instead of pymalloc). Thanks @ryanking13!

PR #2904, opened 2022-07-23
Add a new CLI entrypoint for Pyodide

ryanking13:
Related: #2879, #1977

### Description

This PR is not ready to be reviewed yet. But I'm opening this because I would like to start a discussion for the future design.

This PR mainly does two things:

1. Add a new package `pyodide-cli` which provides an extensible CLI interface (`pyodide`) for Pyodide. For now, it supports building Pyodide and packages using `pyodide-build` API.

__Example:__

[asciicast](https://asciinema.org/a/CufVOTLVOm7ExM7AOoeQh0zfE)

2. Migrate some parts of Makefile into `pyodide-build`.

The purpose of this change is to replace `make` with a high-level build tool that Python devs are easier to use. This is done by [doit](https://pydoit.org/). One advantage of using doit is that it has better dependency management than make so we can reduce rebuild time a lot.

This change means that `pyodide-build` is now used not only for building packages but also for building Pyodide core.

### Checklists

- [ ] Add / update tests


pyodide-build/pyodide_build/doit_tasks.py
```diff
@@ -0,0 +1,426 @@
...
+from pyodide_build.buildpkg import get_bash_runner
+from pyodide_build.common import get_make_flag, search_pyodide_root
+
+# from doit.reporter import ZeroReporter
+
+DOIT_CONFIG = {
+    "verbosity": 2,
+    "minversion": "0.36.0",
+    "action_string_formatting": "both",
+    # "reporter": ZeroReporter,
```
bollwyvl, line 16: A other subjective, but hard-won config insights:
- `"par_type": "thread",`
  - this is the only one that works correctly across all platforms
- `"backend": "sqlite3",
  - this is the least likely to get corrupted
  - it only creates one file

bollwyvl:
Started digging into the dirty of the doit tasks before reading the top-level saying this isn't ready to be reviewed :blush:  Very exciting stuff!

A few notes from doit at scale:
- `task_deps` is unfortunately not as good a means to manage long chains as one might hope
  - it seems to only respect whether a task _has been run_, not that _it is up-to-date_
- In a perfect world, every successful task would generate a file with contents that perfectly represent a function of its `file_deps`... 
    - practically, this isn't the case
    - however, just `touch`ing an empty file and declaring it as a `targets` and adding it to the `file_deps` of the downstream will at least fix the "run, but not up to date" issue
- `@create_after` is really powerful
- while `doit` makes it _easy_ to use shell string actions... 
  - it quickly falls apart on windows, or with utf-8 paths
- adding a `cwd` argument to your `run_in_env` wrapper will clean a lot `cd` stuff
  - kwargs are often more readable for python actions, e.g. `(task_func, [], {"cwd": "foo"})
- `uptodate` is awesome, especially when combined with `doit.tools.config_changed`
  - a primary case is digging into the data of a JSON/yaml/toml file
    - e.g. you only need to run `npm install` when the `*dependencies` keys of `package.json` change
  - however, parsing hundreds of files to even _generate_ tasks is really expensive

Some (sometimes elusive) goals of a good set of tasks:
- `doit -n8` works
  -  this sometimes means having to force serial behavior
- `doit && doit` eventually "stabilizes"
  - if there is a task that always runs, it needs a `file_deps` and a `targets`
- every single `def task_` has a good, unique docstring
- every single `yield {"name": ...}` also has a unique `"doc": "..."` member

ryanking13:
Thanks @bollwyvl, your advice is really helpful for me:) It was a bit hard to find a good reference for using doit at scale...

rth:
Thanks for working on this @ryanking13 ! I'll review it after we make the release. Though in general I feel a good fist step would be to create a separate `pyodide/pyodide-cli` repo with the main entry ploing + plugin mechanism (which should be completely orthogonal to how we reimplement Makefile functionality) and use it in this MR.

ryanking13:
> Though in general, I feel a good first step would be to create a separate `pyodide/pyodide-cli` repo with the main entry ploing + plugin mechanism (which should be completely orthogonal to how we reimplement Makefile functionality) and use it in this MR.

Yes, I was able to remove the dependency between `pyodide-cli` and `pyodide-build` (thanks to bollwyvl) so it will be okay to make a separate `pyodide/pyodide-cli` repo now.

rth:
I created https://github.com/pyodide/pyodide-cli could you make a PR with just that package there?

rth:
I synced with upstream/main and switched to using [pyodide-cli](https://github.com/pyodide/pyodide-cli) as a dependency so the diff here is a bit shorter. 

Overall this looks very nice. I think we should split this MR further command by command (e.g. starting with an easy one such new_package). Then for each command we add,  remove the legacy CLI entrypoint (we are not going to support multiple ways of doing the same thing CLI wise, and this will be a breaking change), update the docs accordingly and make sure there is some test for the CLI maybe running `pyodide <command> --help` at least.

ryanking13:
> I synced with upstream/main and switched to using [pyodide-cli](https://github.com/pyodide/pyodide-cli) as a dependency so the diff here is a bit shorter.

Thanks! I will start working on this again after I finish other ongoing work.

rth:
So a lot of this was done I think. One remaining thing is to build cpython & emsdk when running `pyodide build-recipes`.

Would it be better to close this and open a separate PR for it, or do you plan to resolve conflicts here @ryanking13 ?

ryanking13:
> Would it be better to close this and open a separate PR for it, or do you plan to resolve conflicts here @ryanking13 ?

Thanks for the ping @rth. I'll close this and open separate PRs to improve our build system step-by-step.

PR #2906, opened 2022-07-24
Make JsProxy subclass Mapping or Sequence when appropriate

hoodmane:
This ensures that `PySequence_Check` of a `JsProxy` of a `Array` returns `True`, that `isinstance(jsproxy_of_array, Sequence)` returns `True`, and that `isinstance(jsproxy_of_mapping, Mapping)` returns `True`.
This makes it possible to use `JsArray` and `JsMap` with the `match` keyword.
See https://github.com/pyodide/pyodide/discussions/2903

To make the Array a valid `MutableSequence`, we are supposed to implement `append`, `reverse`, `extend`, `pop`, `remove`, and `__iadd__`, `__reversed__`, `index`, and `count`. Not sure if this is really necessary but the Python docs say we should...

Edit: I added the `Mapping` and `Sequence` mixin methods in #2907 and #3275.

### Checklists

- [x] Add / update tests


ryanking13 (commented):
Thanks, @hoodmane. There are some test failures, but they don't seem relevant to this change. Probably we should increase the timeout for hypothesis tests as playwright-firefox seems to get timeout consistently...

rth:
> To make the Array a valid MutableSequence, we are supposed to implement append, reverse, extend, pop, remove, and __iadd__, __reversed__, index, and count. Not sure if this is really necessary but the Python docs say we should...

That's not true for numpy arrays, so I think it's fine not to do it,
```py
>>> import numpy as np
>>> x = np.ones(10)
>>> x.pop(1)
  File "<stdin>", line 1, in <module>
AttributeError: 'numpy.ndarray' object has no attribute 'pop'
```

rth:
Maybe @antocuni would have an interest in reviewing )

hoodmane:
Okay since I added the mixin methods I'm going to merge this now.

PR #2907, opened 2022-07-25
Implement slicing for arrays

hoodmane:
More work on features discussed in #2903.

### DIFF
--- a/src/core/jsproxy.c
+++ b/src/core/jsproxy.c
@@ -504,11 +504,52 @@ JsProxy_ass_subscript_array(PyObject* o, PyObject* item, PyObject* pyvalue)
   JsProxy* self = (JsProxy*)o;
   bool success = false;
   JsRef idvalue = NULL;
+  PyObject* seq = NULL;
   Py_ssize_t i;
   if (PySlice_Check(item)) {
-    PyErr_SetString(PyExc_NotImplementedError,
-                    "Slice subscripting isn't implemented");
-    return -1;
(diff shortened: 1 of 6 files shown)


src/core/jsproxy.c
```diff
@@ -467,9 +467,27 @@ JsProxy_subscript_array(PyObject* o, PyObject* item)
     return pyresult;
   }
   if (PySlice_Check(item)) {
-    PyErr_SetString(PyExc_NotImplementedError,
-                    "Slice subscripting isn't implemented");
-    return NULL;
+    Py_ssize_t start, stop, step, slicelength;
+    if (PySlice_Unpack(item, &start, &stop, &step) < 0) {
```
ryanking13, line 471: I think we can use `FAIL_IF_MINUS_ONE` here too, same as what you did in `JsArray_ass_subscript`.

ryanking13 (approved):
Thanks, @hoodmane. Looks very cool. Please update the changelog, otherwise LGTM.

src/core/jsproxy.c
hoodmane: Can't because there is no `finally` block in this function. Perhaps we should add one that just returns `NULL`.

hoodmane: I adulterated my typical C style here with some code that was nearly copied verbatim from `cpython/Objects/listobject.c`

rth:
Very nice!

PR #2909, opened 2022-07-25
Add REBOUNDx package

hannorein:
### Description

This PR is for adding support for the REBOUNDX package, a library for including additional forces in REBOUND (REBOUND is a package that has been recently added to pyodide). This is not working yet, so don't merge it yet. I'm just creating the PR to get CI to run the tests.

### Checklists

- [x] Add / update tests


hannorein:
I might need a bit of help here. Background: this is a package (REBOUNDx) that depends on another package (REBOUND). The build script of REBOUNDx exits with an error message, saying it can't find the .so file for REBOUND. I think the message comes from this import statement in the setup.py file [here](https://github.com/hannorein/reboundx/blob/master/setup.py#L32). But I'm unsure why because REBOUND itself builds and runs fine (#2868).

A few questions:

How does `python -m pyodide_build buildpkg` install the dependencies? Is it using the local yaml file for the dependencies or is it getting them from elsewhere? 

Getting the right path for the .so library has always been a bit of a hack in both REBOUND and REBOUNDx. We always had trouble getting this to work on different operating systems. Is [this](https://github.com/hannorein/rebound/blob/main/rebound/__init__.py#L23) the right way to load the .so library?

If I understand it correctly, `python -m pyodide_build buildpkg` installs a virtual environment. Is there a way to play around with it after the build fails? It would be great if I could check how the installation for REBOUND turned out. 

Thank you!

ryanking13:
Could you try adding `cross-build-env: true` inside the meta.yaml of `rebound` package? ([example](https://github.com/pyodide/pyodide/blob/7231cab3ffc83f6221fafb7458f9b223d2a7c759/packages/pycparser/meta.yaml#L7-L8))

It seems like the `reboundx` package searches for the local `rebound` package during the build, so we need to install the cross-built rebound package locally.

ryanking13:
> How does `python -m pyodide_build buildpkg` install the dependencies? Is it using the local yaml file for the dependencies or is it getting them from elsewhere?

It uses a local yaml file to install dependencies, note that `pyodide_build buildpkg` actually does not install dependencies. `pyodide_build buildall` does that.

> Getting the right path for the .so library has always been a bit of a hack in both REBOUND and REBOUNDx. We always had trouble getting [this](https://github.com/hannorein/rebound/blob/main/rebound/__init__.py#L23) to work on different operating systems. Is this the right way to load the .so library?

I think ctypes.cdll worked fine with Emscripten.

> If I understand it correctly, python -m pyodide_build buildpkg installs a virtual environment. Is there a way to play around with it after the build fails? It would be great if I could check how the installation for REBOUND turned out.

The virtual build environment will be cleaned after the build (even if it fails). Probably you can hack [pypabuild.py](https://github.com/pyodide/pyodide/blob/7231cab3ffc83f6221fafb7458f9b223d2a7c759/pyodide-build/pyodide_build/pypabuild.py#L64-L95) to keep the build environment.

hannorein:
Thanks a lot for the tips! I've tried adding `cross-build-env: true` but the error remained the same. I'll see if I can hack pypabuild.py to keep the virtual environment around after it fails. That would probably help a lot to find what's going on...

ryanking13:
If the build fails, could you also try adding:

```yaml
build:
  cross-build-files:
    - librebound.cpython-310-wasm32-emscripten.so
```

hannorein:
Also no luck with that change.

hoodmane:
To keep the virtual environment, uncomment line 74 of pypabuild.py

ryanking13:
> Also no luck with that change.

You need to add `cross-build-env` and `cross-build-files` to the `rebound` not `reboundx`.

hannorein:
Oh, ok. Thank! That did something. The error message changed to "invalid ELF header".

jobovy:
I looked into this a little and while I did get the code to compile, it still doesn't run. But I can describe what I did in the hope that it might help clarify the issues.

It seems to me that the first problem is that the cross-compilation calls `setup.py` first on the host and then on the target, but even on the host it expects the `.cpython-310-wasm32-emscripten.so` suffix, so it tries to find the same file twice but wants it to be the host version first and the target version second. So adding the `cross-build-env: true` stuff to `rebound`'s `meta.yaml` doesn't help, because that copies the _host_ library (with suffix `.cpython-310-x86_64-linux-gnu.so`). So what I did to get around this is:
- I mostly removed the part where `reboundx`'s `setup.py` imports `rebound` to set the linking arguments with [this patch](https://github.com/jobovy/pyodide/blob/74e1c47669e5eb68e3f09e5dee5bc394fc18c8ff/packages/reboundx/patches/0002-Skip-build_ext-in-pyodide.patch), because I don't think this is really necessary.
- I copied the `rebound` library to `WASM_LIBRARY_DIR` using
  ```
    post: |
      cp $WHEELDIR/librebound.cpython-$PYMAJOR$PYMINOR-*.so $WASM_LIBRARY_DIR/
  ```
  in its `build:` section and then added
  ```
    ldflags: |
      -L$(WASM_LIBRARY_DIR)/
  ```
  to `reboundx`'s `build:` section. Then the `wasm-ld` linker can link to the correct file. I couldn't get the environment variables for the suffix to work in `post:` so that's why I used `librebound.cpython-$PYMAJOR$PYMINOR-*.so`.

This leaves some small issues:
- One is that somehow `-R.` is included in the `wasm-ld` link command and that's not an allowed flag, so I added `"-R."` to the linker options to be removed here: https://github.com/jobovy/pyodide/blob/2d9c51f372e84475fc1898971f28c61b4c903762/pyodide-build/pyodide_build/pywasmcross.py#L305
- `reboundx` uses `rand_r`, which is an obsolete `stdlib` function and, I guess, not included in the `emscripten` version. So I changed those to `rand` with [this patch](https://github.com/jobovy/pyodide/blob/74e1c47669e5eb68e3f09e5dee5bc394fc18c8ff/packages/reboundx/patches/0001-rand_r-rand.patch). But that might not be the right way to change these calls.

Then `reboundx` builds fine and seems to produce a correct wheel (with the shared-object library in it). I've pushed all of this to https://github.com/jobovy/pyodide/tree/reboundx.


However, trying to run the code then gives various errors and weird behavior:
- Just doing `import rebound` for some reason tries to load `reboundx`, although it's not a dependency. This fails with an error like "GET http://.../dist/librebound.cpython-310-wasm32-emscripten.so 404 (Not Found)"
- Doing
  ```
  import micropip
  await micropip.install('rebound')
  import rebound
  ```
  does _not_ install `reboundx`. Using this to import `reboundx` also works without errors.
- However, in both cases, actually trying to run [the example from the reboundx docs](https://reboundx.readthedocs.io/en/latest/python_quickstart.html#python-quickstart) fails with
  ```
  TypeError: Cannot read properties of undefined (reading 'apply')
  ```
  and the browser console gives
Screen Shot 2022-07-26 at 10 17 19 AM

Anyway, there's probably a better way to deal with the build issues, but this is the only thing I could get to work and I've already spent too much time on this already...

jobovy:
Okay, figuring that the issue might be that symbols aren't correctly linked from the `rebound` shared-object library during the `reboundx` build, I got something to work by compiling `rebound`'s C extension to a regular archive in a very hacky way, by adding
```
  post: |
    emar rcs librebound.a $PKGDIR/build/rebound-3.19.4/build/temp.emscripten_3_1_14_wasm32-cpython-310/src/*.o
    cp librebound.a $WASM_LIBRARY_DIR/lib/
```
to `rebound`'s `meta.yaml` and then making sure to link to `-lrebound` in the `reboundx` compilation (e.g., using [this patch](https://github.com/jobovy/pyodide/blob/355b115a7bd823acebfe65ffa20c9848c0c5b850/packages/reboundx/patches/0003-Just-link-to-librebound.patch)). There must be a better way, but at least this seems to work (working version [here](http://astro.utoronto.ca/~bovy/pyodide/dist/console.html); full code [here)](https://github.com/jobovy/pyodide/commit/355b115a7bd823acebfe65ffa20c9848c0c5b850)

hoodmane:
> cross-compilation calls setup.py first on the host and then on the target, but even on the host it expects the .cpython-310-wasm32-emscripten.so suffix

Hmm that's not the intended behavior. I think it calls setup.py on the host purely to inquiry about build dependencies. Some packages really don't expect this and do nontrivial work even when asked for build dependencies. We could try to set up the cross build environment even when inquiring about dependencies or perhaps having an option to explicitly specify build dependencies in the meta.yaml file would be helpful.

Well the fatal error is a fairly traditional dynamic linking error: some symbol needed from the GOT was not available when you made the call. It's possible you need to add some modification to ensure that symbols from the rebound so file are loaded with global visibility... First though it would be good to work out the name of the missing symbol. I can provide more detailed instructions when I have time. I'm on vacation now so I don't have that much time to look into these things.

hoodmane:
I think we solved a similar symbol visibility issue in scipy here:
https://github.com/pyodide/pyodide/blob/d4e157c972fc19b9846ac4f161c034b5a1728d44/packages/scipy/patches/0002-loadDynamicLibrary-flapack.patch#L45

jobovy:
> I think we solved a similar symbol visibility issue in scipy here:
> 
> https://github.com/pyodide/pyodide/blob/d4e157c972fc19b9846ac4f161c034b5a1728d44/packages/scipy/patches/0002-loadDynamicLibrary-flapack.patch#L45

Yeah, that approach combined with my first approach seems to work: https://github.com/jobovy/pyodide/commit/841bf2e1d977ce5feea9237d8b2c509c815f757e. Still gives the "GET http://.../dist/librebound.cpython-310-wasm32-emscripten.so 404 (Not Found)" console error when loading `reboundx`, but the example now runs, so I guess it's harmless?
 
> Hmm that's not the intended behavior. I think it calls setup.py on the host purely to inquiry about build dependencies. Some packages really don't expect this and do nontrivial work even when asked for build dependencies. We could try to set up the cross build environment even when inquiring about dependencies or perhaps having an option to explicitly specify build dependencies in the meta.yaml file would be helpful.

I'm not sure that what I said is really going on, I was just speculating that this was the problem, because it seemed like giving the host (unix I guess) version of the shared-library first and then the emscripten one second resolved the issue. But I don't understand the cross-compiler well enough to really know what's going on!

hoodmane:
As an aside, `loadDynamicLibrary` is misleadingly named. It doesn't load a library at all, it makes the symbols from an already loaded library globally visible.

jobovy:
The CI build over at #2914 worked fine and I've put the artifacts here: http://astro.utoronto.ca/~bovy/pyodide-pr2914/dist/console.html. `reboundx` seems to work fine there. So maybe @hannorein could pull in the latest change there and add a test/changelog entry here so this could be reviewed?

hannorein:
Thanks so much for everyone's help! I'm at the airport at the moment -- I might be a day or two before I get to it....

ryanking13:
> It's possible you need to add some modification to ensure that symbols from the rebound so file are loaded with global visibility... First though it would be good to work out the name of the missing symbol. I can provide more detailed instructions when I have time. I'm on vacation now so I don't have that much time to look into these things.

That's very interesting. I'm not sure why symbol visibility problem happens. We load [all dynamic libraries globally when loading a package](https://github.com/pyodide/pyodide/blob/d4e157c972fc19b9846ac4f161c034b5a1728d44/src/js/load-package.ts#L310-L315).

I wish you have a good vacation by the way👍

packages/reboundx/test_reboundx.py
```diff
@@ -0,0 +1,28 @@
+from pyodide_test_runner import run_in_pyodide
+
+
+@run_in_pyodide(
+    packages=[
+        "rebound",
```
jobovy, line 6: You need:
```suggestion
        "rebound",
        "reboundx",        
```
Strangely enough the CI checks are okay even though this test failed...

ryanking13: > Strangely enough the CI checks are okay even though this test failed...

Yeah, that is a bug in our test (#2915)

hoodmane:
> I'm not sure why symbol visibility problem happens. We load [all dynamic libraries globally when loading a package](https://github.com/pyodide/pyodide/blob/d4e157c972fc19b9846ac4f161c034b5a1728d44/src/js/load-package.ts#L310-L315).

That code only executes if `sharedlibrary: true` is in the `meta.yaml` file. This makes sense because generally Python uses `dlopen` and `dlsym` to handle .so files and so it doesn't need the symbols to have entries in the GOT. We probably should add a finer-grained way to expose specific so files at load time that is less hacky.

ryanking13:
> That code only executes if `sharedlibrary: true` is in the `meta.yaml` file.

Hmm, AFAIK that code is executed every time regardless of whether we are installing a package or a shared library, which is because WASM loader is async only.

Didn't we discuss this previously [here](https://github.com/pyodide/pyodide/pull/2721#issuecomment-1157132083)? Or maybe I am misunderstanding something...

hoodmane:
Well it's behind `if(shared)` and if you chase it through, `shared` is controlled by the `sharedlibrary` flag.

ryanking13:
> Well it's behind `if(shared)` and if you chase it through, `shared` is controlled by the `sharedlibrary` flag.

Oh, you are right, thanks for the explanation. I won't bother you more, enjoy your vacation :)

hannorein:
@jobovy Sorry for the stupid question. What commands did you use to get the console working, especially, how did you compile all the packages? I can compile REBOUND and REBOUNDx now, but it takes ages trying to compile numpy. In docker, I'm running `make`, and then `python -m pyodide_build buildall --only "reboundx" packages dist` (which then tries to build numpy). I'm probably missing something... Thanks!

jobovy:
> @jobovy Sorry for the stupid question. What commands did you use to get the console working, especially, how did you compile all the packages? I can compile REBOUND and REBOUNDx now, but it takes ages trying to compile numpy. In docker, I'm running `make`, and then `python -m pyodide_build buildall --only "reboundx" packages dist` (which then tries to build numpy). I'm probably missing something... Thanks!

To get the console, I do
```
PYODIDE_PACKAGES="reboundx" make
```
That does take a while to run (certainly the first time), but normally `numpy` doesn't take that long to build (but it can take a while, then you just have to wait). Are you doing this on a Mac? The docker image runs take very long on a mac, I use a unix server for this.

hannorein:
You were right. It just took about an hour to compile numpy and matplotlib... Thanks!

hannorein:
@jobovy: Using `rand()` instead of `rand_r` is not great because it prevents reproducible results. And using `srand()` would also be problematic because it's not thread safe. But I'm not sure [0001-rand_r-rand.patch](https://github.com/pyodide/pyodide/pull/2909#diff-67d91961f6cb6d0b99214b9a1933a9cb0de43ea1739fe0daffd5d58916e4fa09) is actually necessary. REBOUND uses the same function (`rand_r`) and seems to compile fine. Can I remove the patch or am I missing something?

jobovy:
> @jobovy: Using `rand()` instead of `rand_r` is not great because it prevents reproducible results. And using `srand()` would also be problematic because it's not thread safe. But I'm not sure [0001-rand_r-rand.patch](https://github.com/pyodide/pyodide/pull/2909#diff-67d91961f6cb6d0b99214b9a1933a9cb0de43ea1739fe0daffd5d58916e4fa09) is actually necessary. REBOUND uses the same function (`rand_r`) and seems to compile fine. Can I remove the patch or am I missing something?

The code doesn't compile with the patch, but I figured out why. `rand_r` is [only defined in `emscripten`'s version of `stdlib` if `_GNU_SOURCE` or some other pre-compiler macros are defined](https://github.com/emscripten-core/emscripten/blob/e05e72d9c49fe15578e73934ce525a894d1b712a/system/lib/libc/musl/include/stdlib.h#L85-L100). In `rebound`, you [explicitly define this](https://github.com/hannorein/rebound/blob/399189ba37c604af7bde5c42ebceabb7d34cfc7a/setup.py#L58), but [not in `reboundx`](https://github.com/dtamayo/reboundx/blob/e8f715a1cdcd0e453ba1f21448a0c977bace14ad/setup.py#L71). So either you need to patch that line in `reboundx`'s `setup.py` or just make the change to `reboundx` (probably a good idea) and release a new version. Then you can remove the first patch (and the other patch is still correctly applied).

hannorein:
This is now working for me locally. I've updated REBOUND and REBOUNDx so that no patches are required. But for some reason the CI tests now fail with `No module named 'pyodide_test_runner'`.  Not sure what could have changed and not sure how to fix it.

jobovy:
> This is now working for me locally. I've updated REBOUND and REBOUNDx so that no patches are required. But for some reason the CI tests now fail with `No module named 'pyodide_test_runner'`. Not sure what could have changed and not sure how to fix it.

I think you just need to switch to the new `pytest_pyodide` package as is done here: https://github.com/pyodide/pyodide/commit/c85e00806b38b23efc2b2c656457fcd8117cfee3#diff-8fb10e08e53c67a74a4440eaaa957063b706bdada950466adf124abeaca8bb9c.

hoodmane (approved):
Generally looks good to me, thanks all for working on it. The CI doesn't seem to be working quite right @ryanking13 it doesn't look like `test_packages_common.py::import[chrome-reboundx]` ran either on `test-packages-chrome-no-numpy-dependents` or on `test-packages-chrome`. But it seems that `test_reboundx.py::modifymass` most likely is running and working so I think we can merge this PR.

hoodmane:
Ah looks like #2942 fixes the import test issue?

pyodide-build/pyodide_build/pywasmcross.py
```diff
@@ -307,6 +307,7 @@ def replay_genargs_handle_linker_opts(arg: str) -> str | None:
                 "--sysroot=",  # ignore unsupported --sysroot compile argument used in conda
                 "--version-script=",
                 "-R/",  # wasm-ld does not accept -R (runtime libraries)
+                "-R.",  # wasm-ld does not accept -R (runtime libraries)
```
jobovy, line 310: LGTM, but the only thing I was wondering about is whether this change I made to add `-R.` to the list of compile options to be removed in `pywasmcross.py` should instead go in this list slightly higher up:
https://github.com/pyodide/pyodide/blob/14708295cd1d2a38235e2dd3727ed7760e29eaa0/pyodide-build/pyodide_build/pywasmcross.py#L294-L303
so only `-R.` exactly gets removed rather than anything that starts with `-R.`. Not sure whether there could be valid options that start with `-R.`. OTOH, maybe leaving it as is is best so things like `-R./another_dir/` also get removed (also not sure whether that could ever happen though).

hoodmane: I think it's fine this way since I doubt we would ever want to keep a flag that starts with `-R.`

PR #2806, opened 2022-06-29
Move sqlite3 and libffi to emscripten ports

ryanking13:
### Description

Related: #2805

This PR adds a emscripten port script for `sqlite3` and `libffi` libraries that can be built via embuilder.
This PR also replaces `bzip2` and `zlib` that are used in CPython build with official emscripten ports of those libraries.

PR #2912, opened 2022-07-26
BLD Use emscripten ports of bzip and zlib when building cpython

ryanking13:
Use official emscripten ports for bzip and zlib when building cpython, so we don't need to rebuild these libraries when building other packages.

This was part of #2806.

After we update Emscripten version to 3.1.15 (#2827), we can also use sqlite3 port (thanks to tiran).

### DIFF
--- a/Makefile.envs
+++ b/Makefile.envs
@@ -100,6 +100,8 @@ export MAIN_MODULE_LDFLAGS= $(LDFLAGS_BASE) \
 	-s EXPORT_NAME="'_createPyodideModule'" \
 	-s EXCEPTION_CATCHING_ALLOWED=['we only want to allow exception handling in side modules'] \
 	-s DEMANGLE_SUPPORT=1 \
+	-s USE_ZLIB \
+	-s USE_BZIP2 \
 	-s FORCE_FILESYSTEM=1 \
 	-s TOTAL_MEMORY=20971520 \
 	-s ALLOW_MEMORY_GROWTH=1 \
@@ -109,8 +111,6 @@ export MAIN_MODULE_LDFLAGS= $(LDFLAGS_BASE) \
 	-lpython$(PYMAJOR).$(PYMINOR) \
 	-lffi \
 	-lsqlite3 \
-	-lbz2 \
-	-lz \
 	-lstdc++ \
 	-lidbfs.js \
 	-lnodefs.js \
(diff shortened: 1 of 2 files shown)


Makefile.envs
```diff
@@ -100,7 +100,6 @@ export MAIN_MODULE_LDFLAGS= $(LDFLAGS_BASE) \
 	-s EXPORT_NAME="'_createPyodideModule'" \
 	-s EXCEPTION_CATCHING_ALLOWED=['we only want to allow exception handling in side modules'] \
 	-s DEMANGLE_SUPPORT=1 \
-	-s USE_ZLIB=1 \
```
tiran, line 103: I recommend that you use ``-s USE_ZLIB`` and ``-s USE_BZIP2`` here, too. It is the correct way to link with libz and libbz2.

Makefile.envs
```diff
@@ -100,7 +100,6 @@ export MAIN_MODULE_LDFLAGS= $(LDFLAGS_BASE) \
 	-s EXPORT_NAME="'_createPyodideModule'" \
 	-s EXCEPTION_CATCHING_ALLOWED=['we only want to allow exception handling in side modules'] \
 	-s DEMANGLE_SUPPORT=1 \
```
tiran, line 102:
```suggestion
	-s DEMANGLE_SUPPORT=1 \
	-s USE_ZLIB \
	-s USE_BZIP2 \
```

Makefile.envs
```diff
@@ -111,6 +110,7 @@ export MAIN_MODULE_LDFLAGS= $(LDFLAGS_BASE) \
 	-lffi \
 	-lsqlite3 \
 	-lbz2 \
+	-lz \
```
tiran, line 113:
```suggestion
```

tiran (changes requested):

Makefile.envs
ryanking13: Personally, I prefer calling `embuilder build zlib` instead of using `-s USE_ZLIB` so I can explicitly understand what is going on behind, but yeah, for now `-s USE_ZLIB` seems more recommended...

tiran: Python 3.11 uses the flags for compiler and linker. The ``-sUSE_FOO`` variant adds correct compiler and linker flags for various build flavors: debug/release, PIC/non-PIC, and (in the case of sqlite) single/multi-threaded. Yes, that's eight different builds for sqlite.

rth (approved):
Thanks a lot @ryanking13 and thanks for the review @tiran !

PR #2913, opened 2022-07-26
DOC Use dev CDN url if building latest docs

### LINKED ISSUE #2911: https://pyodide.org/en/latest/console.html doesn't show latest `main` version
## 🐛 Bug

The https://pyodide.org/en/latest/console.html console is stuck at `v0.21.0a3`. I believe this is because the version wasn't bumped to the next 'dev' version after the `v0.21.0a3` release, so somebody probably needs to run
```
./tools/bump_version.py --new-version 0.21.0.dev0

```
Without `dev` in the version, the documentation's console uses the release:
https://github.com/pyodide/pyodide/blob/7231cab3ffc83f6221fafb7458f9b223d2a7c759/docs/conf.py#L20-L23

### To Reproduce

Go to https://pyodide.org/en/latest/console.html and load a package added since v0.21.0a3, e.g., `import rebound`

ryanking13:
Resolve #2911

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -17,7 +17,7 @@ project = "Pyodide"
 copyright = "2019-2022, Pyodide contributors and Mozilla"
 pyodide_version = "0.21.0a3"
 
-if ".dev" in pyodide_version:
+if ".dev" in pyodide_version or os.environ.get("READTHEDOCS_VERSION") == "latest":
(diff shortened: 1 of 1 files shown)


rth (approved):
Indeed, thanks!

Thanks also @jobovy for reporting it.

PR #2915, opened 2022-07-27
TST Fix test skip check

ryanking13:
`--skip-passed` flag sometimes skips some tests that were not actually passed. This fixes it.


ryanking13:
Merging since this bug is deceiving us that tests (that are actually failing) pass.

PR #2925, opened 2022-08-01
Add package `brotli`

MuTsunTsai:
### Description

This is related to issue #2921. Add compression package `brotli`.

### Checklists

- [x] Add a [CHANGELOG](https://github.com/pyodide/pyodide/blob/main/docs/project/changelog.md) entry
- [x] Add / update tests
- [x] Add new / update outdated documentation

PR #2926, opened 2022-08-01
TST Fix location of the package import test file

ryanking13:
~~This file prevents pytest from adding the directory to sys.path during the test.~~

_Edited_

This change  prevents pytest from adding the `packages` directory to sys.path during the test

See: https://github.com/pyodide/pyodide/pull/2925#pullrequestreview-1056792252


rth:
I'm not sure about this fix. Can't we control this via pytest settings rather than by making `packages` a Python module which IMO isn't really desirable?

ryanking13:
I wasn't able to find a way to disable pytest sys.path injection. One other solution can be simply moving `test_packages_common.py` into some other directory (e.g. `packages/_test`), which will prevent pytest from adding `packages/` to sys.path. WDYT?

rth:
Sounds good, thanks!

PR #2938, opened 2022-08-04
Be more careful about different array variants

hoodmane:
This fixes a few problems introduced in #2907. The slicing doesn't work correctly for `TypedArray` or `NodeList` and there are several possible routes to take with `TypedArray` so I think it's better to leave it not implemented for now. Also, `extend` and `+=` don't work for `TypedArray` and `NodeList`. This also gives a better error if you attempt `del a[0]` when `a` is a `TypedArray` or `NodeList`. The former have a fixed length, the latter are immutable.

### Checklists

- [x] Add / update tests

### DIFF
--- a/src/core/jsproxy.c
+++ b/src/core/jsproxy.c
@@ -41,17 +41,19 @@
 #include "structmember.h"
 
 // clang-format off
-#define IS_ITERABLE  (1<<0)
-#define IS_ITERATOR  (1<<1)
-#define HAS_LENGTH   (1<<2)
-#define HAS_GET      (1<<3)
-#define HAS_SET      (1<<4)
-#define HAS_HAS      (1<<5)
-#define HAS_INCLUDES (1<<6)
-#define IS_AWAITABLE (1<<7)
-#define IS_BUFFER    (1<<8)
-#define IS_CALLABLE  (1<<9)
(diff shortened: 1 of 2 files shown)


src/core/jsproxy.c
```diff
@@ -2227,6 +2289,32 @@ JsProxy_get_subtype(int flags)
   return (PyTypeObject*)type;
 }
 
+EM_JS_NUM(int, JsProxy_array_detect, (JsRef idobj), {
+  try {
+    let obj = Hiwire.get_value(idobj);
+    let result = 0;
```
ryanking13, line 2295:
```suggestion
```

This variable doesn't seem used?

src/core/jsproxy.c
```diff
@@ -587,6 +619,25 @@ JsArray_ass_subscript(PyObject* o, PyObject* item, PyObject* pyvalue)
   return success ? 0 : -1;
 }
 
+/**
+ * __setitem__ and __delitem__ for proxies of TypedArrays, controlled by
+ * IS_TYPED_ARRAY
```
ryanking13, line 624:
```suggestion
 * IS_TYPEDARRAY
```

ryanking13 (approved):
Thanks @hoodmane!

PR #2939, opened 2022-08-05
Add lzma

### LINKED ISSUE #2937: Add lzma
As mentioned by @hoodmane in https://github.com/pyodide/pyodide/discussions/2930#discussioncomment-3316181

> Is there an issue open about lzma? What is our position on it again? That we want it but there is no emscripten port and we haven't gotten to it?

I think the main concern was the size increase for everyone vs few people actually needing it. Depending on the size maybe we could make it an unvendored stdlib package (or include by default if the size is negligible).

ryanking13:
### Description

Close #2937

Add lzma,  but unvendors it from stdlib (like ssl).

```sh
$ du _lzma.so
204K    packages/lzma/dist/_lzma.so
```

### Checklists

- [x] Add / update tests


rth:
Sounds good, thanks! But maybe let's make a decision on https://github.com/pyodide/pyodide/issues/2940 before merging this.

ryanking13:
> Sounds good, thanks! But maybe let's make a decision on #2940 before merging this.

Yeah, I think the idea from encukou is worth investigating. Not sure how it can be implemented though.

> I think it would be enough to make imports fail with “module sqlite3 is not included in PyIodide base, install it separately by doing X, see Y for more details”

rth:
Yeah, I was thinking an extra import hook (after all the other ones are run) that raises a custom ImportError for a pre-defined list of packages.

rth (approved):
Actually in any case +1 to merge this. Having this module as optional is better than not having it at all.

packages/lzma/meta.yaml
```diff
@@ -0,0 +1,23 @@
+package:
+  name: lzma
+  version: 1.0.0 # Nonesense
+  _cpython_dynlib: true
+source:
+  path: empty
```
hoodmane, line 6: We really ought to have a workaround for this. Maybe we should skip the `source` section entirely if `_cpython_dynlib` is true.

hoodmane (approved):

packages/lzma/meta.yaml
ryanking13: Yes, we need a proper workaround before we unvendor more stdlibs...

PR #2941, opened 2022-08-05
Add python-magic

### LINKED ISSUE #2922: python-magic
## 🐍 Package Request

- Package Name and Version : python-magic 0.4.27
- Package URL : https://github.com/ahupp/python-magic
- Package Dependencies that needs to be resolved first: libmagic

It's more of a question than a package request. This package needs a library `libmagic.so` and a binary database file `magic.mgc`. I've managed to build both of them, but I don't know how to put `magic.mgc` into a proper location.
`libmagic.so` hard-codes search paths of `magic.mgc`, but it can be patched. So where to put `magic.mgc` is flexible. I hope it can be bound with `python-magic` so that it can be loaded on demand.

eagleoflqj:
### Description

Closes #2922

The lib has to be built twice because `magic.mgc` should be generated by an executable (explained [here](https://bugs.astron.com/view.php?id=198#c3487)).
The 2 patches are specific to pyodide design choice (how we handle shared library and data file), so I don't think they should be upstream-ed.
### Checklists

- [x] Add / update tests

### DIFF
--- a/packages/python-magic/meta.yaml
+++ b/packages/python-magic/meta.yaml
@@ -15,7 +15,7 @@ requirements:
 
 test:
   imports:
-    - python-magic
+    - magic
 
 about:
   home: http://github.com/ahupp/python-magic
--- a/packages/python-magic/meta.yaml
+++ b/packages/python-magic/meta.yaml
@@ -1,0 +1,24 @@
+package:
+  name: python-magic
+  version: 0.4.27
+
+source:
+  url: https://files.pythonhosted.org/packages/da/db/0b3e28ac047452d079d375ec6798bf76a036a08182dbb39ed38116a49130/python-magic-0.4.27.tar.gz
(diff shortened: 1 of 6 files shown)


packages/python-magic/meta.yaml
```diff
@@ -0,0 +1,24 @@
...
+  patches:
+    - patches/0001-Hardcode-libmagic.so-location.patch
+
+requirements:
+  run:
+    - libmagic
+
+test:
+  imports:
+    - python-magic
```
ryanking13, line 18:
```suggestion
    - magic
```

The import test is not running correctly now (will be fixed in #2942)

hoodmane (approved):
We could probably use a more flexible way to put files at particular paths in the Pyodide file system.

Thanks @eagleoflqj!

PR #2942, opened 2022-08-05
TST Fix package import test

ryanking13:
After #2926, `maybe_skip_test` was skipping `test_import_*` tests. (It was interpreting `_test` as a package name)

The new `maybe_skip_test` is taken from `pytest-pyodide` with a slight modification. After this PR, we can remove it from `pytest-pyodide`.

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -2,6 +2,7 @@
 Various common utilities for testing.
 """
 import pathlib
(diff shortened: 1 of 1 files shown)


ryanking13:
Merging to make tests run correctly.

PR #2943, opened 2022-08-05
Report checksum when checksum doesn't match in a package build

hoodmane:
This is a convenience modification. If I am updating a url I often want to update the checksum to force the new url to be accepted when the check fails. This makes it easy to copy the correct checksum from the error message.


ryanking13 (approved):
Thanks, @hoodmane! The `opencv-python` build failure seems to be related to the recent CMake release. I'll open a PR fixing it.

Issue #2940, opened 2022-08-05
[Discussion] Changing the default behavior of loading unvendored stdlib


ryanking13:
## Proposed refactoring or deprecation

- Related: #1542

We unvendors some stdlib packages in order to reduce the size of the main module.

- distutils
- tests
- ssl
- lzma (#2939)
- (maybe more in the future, e.g. sqlite3, unicodedata, ..., #1826)

The `pyodide.loadPyodide` API accepts `fullstdlib` parameter, which is currently true in default. What `fullstdlib` currently does is [load distutils package if true](https://github.com/pyodide/pyodide/blob/14708295cd1d2a38235e2dd3727ed7760e29eaa0/src/js/pyodide.ts#L319-L321). It does not affect loading other unvendored stdlibs like ssl, which I think is a bit confusing for users.

IMO not loading ssl and some other unvendored stdlibs are reasonable since the sizes of those packages are quite big.

```
960K    distutils.tar
840K    ssl-1.0.0.zip
```

So I think it would be great to
1. Change the default value of fullstdlib to “False”.
2. If fullstdlib is True, load all unvendored packages including distutils, ssl, lzma (except for `test` which is too big (~22MB) and mostly not used by users).
3. Add unvendored stdlibs to “always packages” when building.

WDYT?

rth:
We should discuss this at https://discuss.python.org/c/webassembly/28. I wonder how scalable our current approach is and whether it wouldn't break people's assumptions about what the stdlib should contain (particularly now that Wasm is an officially supported CPython target).

ryanking13:
> We should discuss this at https://discuss.python.org/c/webassembly/28. I wonder how scalable our current approach is and whether it wouldn't break people's assumptions about what the stdlib should contain (particularly now that Wasm is an officially supported CPython target).

Yes, I think you are right. Actually, I found that you have already opened a discussion last year about this: https://discuss.python.org/t/minifying-the-stdlib-in-pyodide/8414/6.

ryanking13:
Opened a discussion here: https://discuss.python.org/t/unvendoring-some-of-stdlib-modules/18076

ryanking13:
Back in this topic, my opinion is that we can start with not loading distutils by default (i.e. setting default value of fullstdlib to false), while unvendoring other packages can be discussed case by case later.

rth:
[...]

PR #2946, opened 2022-08-08
Unvendor sqlite3 from stdlib

ryanking13:
### Description

Unvendor the standard library sqlite3 to reduce the size of the main module. It reduces the size of `pyodide.asm.wasm` around ~1.4MB.

Related: #2940, #1542

__Before:__

```sh
du ../dist-main/pyodide*
4944    pyodide.asm.data
1516    pyodide.asm.js
8616    pyodide.asm.wasm
```

__After:__

```sh
$ du pyodide*
4944    pyodide.asm.data
1468    pyodide.asm.js
7264    pyodide.asm.wasm
$ du sqlite3*
468    sqlite3-1.0.0.zip
```

### Checklists

- [x] Add / update tests


tiran:
Emscripten 3.1.15 and newer come with sqlite3 in emports, `-sUSE_SQLITE3`.

ryanking13:
> Emscripten 3.1.15 and newer come with sqlite3 in emports, `-sUSE_SQLITE3`.

Yes, thanks for your porting work @tiran! For now, we are stuck in Emscripten 3.1.14 (#2827)

hoodmane:
Maybe we should backport the patch from emscripten 3.15.

rth:
BTW, `du` measure the size in the filesystem, not the file size and would not be accurate in a number of cases (https://askubuntu.com/a/1163848), so we should probably stop using it for this metric.

rth:
Overall +1 to unvendor it, if there are no major objections from Christian. Particularly now that you added more explicit exception errors for unvendored / missing stlib packages in https://github.com/pyodide/pyodide/pull/2973

There are probably not that many packages that need sqlite3 compared to the total number, and for the rest an extra 1.5MB (probably 0.5MB compressed) is a lot both in load and WASM compilation time.

tiran:
I found a problem with Emscripten ports, see https://github.com/emscripten-core/emscripten/issues/17675

To work around the problem, do ``embuilder build --pic sqlite3`` and add this to Modules/Setup.local

```
*shared*
_sqlite3 _sqlite/blob.c _sqlite/connection.c _sqlite/cursor.c _sqlite/microprotocols.c _sqlite/module.c _sqlite/prepare_protocol.c _sqlite/row.c _sqlite/statement.c _sqlite/util.c -lsqlite3
```

ryanking13:
> I found a problem with Emscripten ports, see [emscripten-core/emscripten#17675](https://github.com/emscripten-core/emscripten/issues/17675)

I was always wondering why `-sUSE_...` doesn't really works when linking packages, and that was the reason... thanks for the info Christian!

docs/project/changelog.md
```diff
@@ -24,6 +24,11 @@ substitutions:
   `append`, and `pop` for `JsProxy` of Javascript arrays.
   {pr}`2970`
 
+- {{ Breaking }} \_sqlite3 stdlib module is not unvendored by default.
+  Now it needs to be maually loaded by {any}`pyodide.loadPackage`
+  or {any}`micropip.install`.
```
hoodmane, line 29:
```suggestion
- {{ Breaking }} Unvendored the \_sqlite3 module from the standard library.
  Before `sqlite3` was included by default. Now it needs to be loaded with 
  {any}`pyodide.loadPackage` or {any}`micropip.install`.
```

hoodmane:
Do we also want to unvendor the `sqlite` python package? It is 143kb uncompressed.

ryanking13:
> Do we also want to unvendor the `sqlite` python package? It is 143kb uncompressed.

Right, probably we should do that for `ssl` and `lzma` too.

hoodmane (approved):
This looks very nice. Thanks @ryanking13!

ryanking13:
Thanks for the reviews!

PR #2950, opened 2022-08-08
Update Emscripten to 3.1.18

hoodmane:
As requested by @tiran. Maybe we'll get lucky and opencv-python and xgboost will build on this version...


hoodmane:
CF https://github.com/pyodide/pyodide/issues/2951

Makefile.envs
```diff
@@ -1,7 +1,7 @@
 export PYVERSION ?= 3.10.2
 # Note: when updating EMSCRIPTEN_VERSION make sure to update
 # the version number in the "uname" patch.
-export PYODIDE_EMSCRIPTEN_VERSION ?= 3.1.14
+export PYODIDE_EMSCRIPTEN_VERSION ?= 3.1.15
```
tiran, line 4:
```suggestion
export PYODIDE_EMSCRIPTEN_VERSION ?= 3.1.18
```

hoodmane: Oops...

ryanking13:
I still don't understand what [the log](https://github.com/pyodide/pyodide/runs/7725790795?check_suite_focus=true#step:9:108)

> 0 is not a valid OpCode

means, and how it is (possibly) releated to the test failures.

hoodmane:
Me neither. But at least xgboost and opencv build here...

hoodmane:
So when we tried to update to 3.1.15, it broke building of xgboost and opencv. Trying to update to 3.1.18 breaks building of RobotRaconteur and testing of pyproj. Though I actually have a theory for why the RobotRaconteur build is broken.

ryanking13:
> So when we tried to update to 3.1.15, it broke building of xgboost and opencv.

In 3.1.15, xgboost and opencv builds, but it was the test that failed.

ryanking13:
I did some binary search among commits in `emscripten-releases` which emsdk uses, and the error happend after this commit: https://chromium.googlesource.com/emscripten-releases/+/c3f55f5c179c073c569684a0640727edb7c5d10e.

However, that commit itself includes 24 commits of `llvm-project`... (sigh)

hoodmane:
Now failing with:
```
Error: bad export type for `gk_cur_jbufs`: undefined
```

ryanking13:
> Now `scipy.linalg` is failing with:
> 
> ```
> Error: bad export type for `gk_cur_jbufs`: undefined
> ```

 `suitesparse` seems failing.

ryanking13:
```
[LinkError: WebAssembly.instantiate(): Import #77 module="env" function="mktime" error: imported function does not match the expected type]
```
It seems like the error is related to the 64bit `time_t` update and hopefully correctly passing LDFLAGS would fix it.

ryanking13:
It looks like it is building now except for minor test errors.

ryanking13 (approved):
Thanks a lot @hoodmane! Please update the changelog.

Probably we can fix `RobotRaconteur` failure after merging this.

hoodmane:
Even on the main branch, `RobotRaconteur` build fails locally for me.

Issue #2949, opened 2022-08-08
DOC Document how to use out of tree builds


rth:
We should document this feature better (while mentioning that it's experimental).

PR #2953, opened 2022-08-09
DOC Document out-of-tree build

rth:
Closes https://github.com/pyodide/pyodide/issues/2949

cc @hoodmane


hoodmane:
Looks good to me thanks @rth. This can probably be refined more in the future (for instance it would be great to add info on testing) but for now I'm +1 to merge this and then make the release.

PR #2954, opened 2022-08-10
ENH Simplify shared library loading process

ryanking13:
Instead of calling `loadWebAssemblyModule` and save it to `preloadedWasm`, directly call `loadDynamicLibrary` with proper fs args to locate other shared libs that are dependent.

This PR also removes one partially related scipy patch (I think is was okay to remove that patch even before this PR)


hoodmane:
I don't think we should load all dynamic libraries globally. What if two of them define the same symbol? Does it not work to use the `shared` flag to control whether the library is loaded globally?

ryanking13:
> I don't think we should load all dynamic libraries globally. What if two of them define the same symbol? Does it not work to use the `shared` flag to control whether the library is loaded globally?

Good point! Fixed it.

ryanking13:
Opened a separate issue about the `reboundx` failure: #2960

hoodmane (approved):
Nice work @ryanking13!

Issue #1824, opened 2021-09-09
Re-use selenium browser cache for tests


rth:
Right now for each test that uses the `selenium_standalone` fixture, will create a new browser profile. In particular this means that the browser cache will not be shared between such tests.

I think creating one browser profile and sharing it among all tests would make those tests significantly faster, since packages and likely compiled wasm results will already be in the browser cache. Or at least indicate the same cache folder in each profile's config.

It looks like there [might be reasons](https://stackoverflow.com/a/40527704/1791279) for creating a new profile each time, but this would be still worth investigating I think.

PR #2957, opened 2022-08-11
TST Experiment: refresh page instead of loading a new web driver

ryanking13:
This is an experiment on optimizing selenium_standalone fixture.
Instead of loading a new webdriver every time, reuse the selenium instance but refresh a page.

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -233,3 +233,17 @@ def extra_checks_test_wrapper(browser, trace_hiwire_refs, trace_pyproxies):
 
 def package_is_built(package_name):
     return _package_is_built(package_name, pytest.pyodide_dist_dir)
+
+
+@pytest.fixture(scope="function")
+def selenium_standalone(selenium):
+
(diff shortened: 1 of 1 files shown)


rth:
Worth investigating indeed. By itself starting a new selenium session is really fast, the problem is that it doesn't then reuse the existing cache (including the WASM compilation cache) https://github.com/pyodide/pyodide/issues/1824

ryanking13:
> Worth investigating indeed. By itself starting a new selenium session is really fast, the problem is that it doesn't then reuse the existing cache (including the WASM compilation cache) #1824

Right, from my local experiments, I found that this approach is somehow useful for micropip tests (50sec -> 35sec), but not very useful for package import tests which were my primary goal. The bottleneck is still in WASM compilation.

ryanking13:
BTW, aside from test time, I think this can be useful for safari tests, which strictly require a single selenium session at a time. #2578

Issue #2951, opened 2022-08-08
RFC Only update Emscripten and Python version at the same time


hoodmane:
## Proposal

Starting with Pyodide v0.22.0 we should only update Emscripten at the same time as we update Python.

### Motivation

Until https://github.com/emscripten-core/emscripten/issues/15917 is resolved, every Emscripten update is assumed to be an ABI break. If we are going to move towards allowing people to build and deploy their own wheels, it would be annoying for them to have to rebuild every time

### Alternatives

Projects have build a new wheel whenever we update Emscripten. They can check the appropriate Emscripten version via `repodata.json:info.platform`. The correct Python version is also exposed via `repodata.json:info.python`. We should make a github action to build against a given Pyodide version that uses `repodata.json` to determine the appropriate version. If the projects use a continuous deployment system, then they can easily specify a range of Pyodide versions to build and deploy wheels for.

### Status

**Rejected.**

rth:
Sounds good. This only applies to major Python versions, not minor ones, right?

hoodmane:
Yes. We could update the minor Python version more frequently if we choose to. Particularly after we update to 3.11 it might be desirable since we won't have to migrate many patches.

ryanking13:
Sounds good to me too. Probably we should use a long living development branch to make sure the Emscripten update doesn't break anything.

> Until [emscripten-core/emscripten#15917](https://github.com/emscripten-core/emscripten/issues/15917) is resolved, every Emscripten update is assumed to be an ABI break. If we are going to move towards allowing people to build and deploy their own wheels, it would be annoying for them to have to rebuild every time

Interesting. I didn't know that they don't guarantee any ABI stability. I'm curious how emscripten-forge team is dealing with this. @wolfv

tiran:
> Sounds good. This only applies to major Python versions, not minor ones, right?
> Yes. We could update the minor Python version more frequently if we choose to.

Naming is hard :) We aren't consistent with our naming conventions for releases, too. Technically our major releases are minor releases, `sys.version_info(major=3, minor=10, micro=6, releaselevel='final', serial=0)`. Maybe call them *annual feature release* and *bugfix release*?

wolfv:
[...]

PR #2958, opened 2022-08-11
Bump emscripten to 3.1.15

ryanking13:
### Description

There are 3 packages (opencv-python, xgboost and pyproj) that were broken when we tried to update Emscripten to 3.1.15.
The error happens in `__wasm_call_ctors`, which is a constructor function called during the initialization of the WASM module.

Between emscripten 3.1.14 and 3.1.15, the behavior related to `__wasm_call_ctors` has been slightly changed.

Before 3.1.15, the data relocation happened "before" `__wasm_call_ctors`, while in 3.1.15, it happens "after" the `__wasm_call_ctors`. To be honest, I don't fully understand the rationale for this change.
(See: https://github.com/llvm/llvm-project/commit/53217ecb882c25bb69e6065512a0627828d6e870)

~So I added a patch that retries `__wasm_call_ctors` after data relocation if it fails. It works, but I'm not sure whether this is a proper fix though.~

_Edit:_

Used upstream commit: https://github.com/emscripten-core/emscripten/pull/17628

### Checklists


ryanking13:
Upstream issue about this error: https://github.com/emscripten-core/emscripten/issues/17601

hoodmane:
Great!

ryanking13:
@hoodmane can we merge this?

hoodmane:
I suppose so. I guess we'll switch to doing #2951 when we release 0.22.0?

ryanking13:
> I suppose so. I guess we'll switch to doing #2951 when we release 0.22.0?

Yes probably. Though I'm not 100% sure that we can follow it as Python releases annually while emscripten develops much faster than that.

> As a followup we can presumably update to 3.1.18 without further issue.

I hope that 3.1.18 update will be much smoother 😂

PR #2961, opened 2022-08-12
Add DuckDB package v0.4.0

IanCal:
### Description

This adds DuckDB based on the 0.4.0 version on pypi.
DuckDB is a fast, columnar, in process database engine and builds with no dependencies other than numpy (which already builds in pyodide). Think sqlite for OLAP.

One key issue here is that duckdb has an extension that can be used for loading files from remote systems. I have not got that extension building.

There are no patches or changes I needed to make to get this building.

### Checklists

* Not added to the changelog as changelog didn't seem to include new packages
* No tests added, didn't see these for just adding a package
* No docs added

Hopefully those are correct but I'm happy to update if needed.


ryanking13:
Thanks, @IanCal!

Please add a changelog entry ([see here](https://github.com/pyodide/pyodide/blob/b497ce26ed94774a8d695a3e61016a7a26e45e3b/docs/project/changelog.md?plain=1#L17)), and some basic tests. You need to add a `test_duckdb.py` file inside the duckdb directory. See [this for an example](https://github.com/pyodide/pyodide/blob/main/packages/xarray/test_xarray.py).

IanCal:
@ryanking13 ah I see the right bit now, and hadn't realised how to do the tests, thanks for the heads up :) will sort that and update thanks!

hoodmane:
This can be merged once it gets a changelog entry.

ryanking13:
> This can be merged once it gets a changelog entry.

And some tests.

ryanking13:
There are some build errors:

```
/tmp/build-env-kf7dtedq/lib/python3.10/site-packages/pybind11/include/pybind11/pybind11.h:15:10: fatal error: 'attr.h' file not found
#include "attr.h"

/tmp/build-env-kf7dtedq/lib/python3.10/site-packages/pybind11/include/pybind11/detail/class.h:13:10: fatal error: '../options.h' file not found
#include "../options.h"
```

I'm not sure which recent changes caused this build failure, maybe becuase of emscripten update?

henryiii:
Those are just relative includes, so they should work, assuming the files are also there.

Is this something breaking that was working, or a new addition breaking?

ryanking13:
Wow, thanks for the prompt response :)

> Those are just relative includes, so they should work, assuming the files are also there.
> 
> Is this something breaking that was working, or a new addition breaking?

Right, they are just header files... I'm not sure what is happening.

I think this PR worked initially (at least according to CI, which was 12 days ago), and now it breaks. Things we've changed related to the build system during that period are:

1. Pinned Cmake version to < 3.24
2. Pinned setuptools version to < 65
3. Updated Emscripten version to 3.1.18

so I guess one of the above is breaking the build.

henryiii:
Why the CMake and Setuptools pins?

henryiii:
Where should I be looking to see the failure?

packages/duckdb/test_duckdb.py
```diff
@@ -0,0 +1,17 @@
+from pytest_pyodide import run_in_pyodide
+
+
+@run_in_pyodide(packages=["duckdb", "pandas"])
+def test_duckdb(selenium):
+    import duckdb
+    import pandas as pd
+
+    example_df = pd.DataFrame({"col_1": [1, 2, 3, 4, 5], "col_2": [10, 20, 30, 40, 50]})
```
henryiii, line 9: This example_df is never being used according to flake8, due to the string usage (wow, that looks odd). Needs a noqa added.

ryanking13:
> Why the CMake and Setuptools pins?

We pinned CMake because CMake 3.24 is causing failure in building `opencv-python` (which is mostly my fault, I have to fix)
and Setuptools becuase it was not compatible with `numpy.distutils` (#2971)

> Where should I be looking to see the failure?

Full build logs can be downloaded here: https://circleci.com/api/v1.1/project/github/pyodide/pyodide/54600/output/105/0?file=true&allocation-id=6304dd480052047ced8ffe5b-0-build%2F746E1738. It is 4.6MB because of so many warnings...

henryiii:
This is pulling pybind11 via `setup_requires`, which has been deprecated for years. But don't see why it wouldn't work. Those files are in the SDist (I think setup_requires doesn't support wheels, only eggs & SDists). I don't see how enscripten could have broken relative includes and not be breaking in all sorts of other places too.

packages/duckdb/test_duckdb.py
IanCal: There's another approach that keeps the variable used and looks cleaner where the table is explicitly registered so I've changed over to that.

IanCal:
Oh - if the tests are passing but it's failing on a "no-numpy-dependents" is there an extra flag I should add to say it is dependent on that? At the very least the tests are as I'm using pandas which I think has numpy as a dep.

packages/duckdb/test_duckdb.py
henryiii: That looks much better!

packages/duckdb/test_duckdb.py
```diff
@@ -0,0 +1,19 @@
+from pytest_pyodide import run_in_pyodide
+
+
+@run_in_pyodide(packages=["duckdb", "pandas"])
```
hoodmane, line 4: You need to declare `pandas` as a requirement in the meta.yaml file. Look at the meta.yaml files for other packages to see how.
```suggestion
@run_in_pyodide(packages=["duckdb"])
```

ryanking13:
Interesting that the build CI now passes... maybe there is some unknown flakiness in our CI.

> This is pulling pybind11 via `setup_requires`, which has been deprecated for years. But don't see why it wouldn't work. Those files are in the SDist (I think setup_requires doesn't support wheels, only eggs & SDists). I don't see how enscripten could have broken relative includes and not be breaking in all sorts of other places too.

Thanks for the ideas Henry!

packages/duckdb/test_duckdb.py
ryanking13: Probably we need to add an option to skip the test if `run_in_pyodide` requires packages that are not built.

hoodmane: Perhaps, but in this case it's a mistake.

henryiii: Pandas is not a requirement, though, it's just needed for this test? I'd guess duckdb doesn't _require_ pandas, but just supports it?

IanCal: Pandas isn't a requirement looking at what happens when you install duckdb in an empty env. However numpy *is*. I'll update the meta.yaml and update the test to remove pandas as a requirement

argenisleon:
Hi,

Anything I can help with so we can fix and merge this? :)

ryanking13:
@argenisleon Thanks for your interest. Yes, this PR is blocked due to the flakiness in the build. I can't reproduce locally, but it happens in the CI. So if you can help us find a reason of the error, or at least make it reproducible, it will be a lot of help.

```
/root/repo/packages/duckdb/build/duckdb-0.4.0/src/include/duckdb_python/pybind_wrapper.hpp:13:10: fatal error: 'pybind11/stl.h' file not found
#include <pybind11/stl.h>
         ^~~~~~~~~~~~~~~~
```

The error is quite mysterious, which it says that it cannot find pybind11 header files, which should be available.

madhur-tandon:
Hi, Can this PR be updated to use `0.6.0` version now?

paultiq:
I took a look at this, but went down a different path of loading duckdb_wasm instead of building duckdb. 

Proof of concept here: https://github.com/iqmo-org/jupylite_duckdb/blob/main/notebooks/example.ipynb

I'm still trying to figure out if this (wasm) approach is sufficient or whether there's significant drawback vs the build approach in this thread. Would appreciate your thoughts, @IanCal : if the build is a stumbling block, would it make more sense to go to wasm route?

cpcloud:
Perhaps this can be closed in favor of #4684?

carlopi:
I guess this this can now be closed, given https://github.com/pyodide/pyodide/pull/4684 has landed.

Issue #2960, opened 2022-08-11
Importing reboundx fails occasionally


ryanking13:
## 🐛 Bug

Importing reboundx fails occasionally in the stable/latest branch.

The reason for the failure is that we download and install packages simultaneously.
(To be more precise, we install shared libraries first then install normal wheel packages.)

https://github.com/pyodide/pyodide/blob/b497ce26ed94774a8d695a3e61016a7a26e45e3b/src/js/load-package.ts#L436-L448

The problem is that `reboundx` is dependent on another non-shared library: `rebound`. When loading `reboundx`, it loads `libreboundx.so` which load `librebound.so` which is in `rebound` package. So if `reboundx` is downloaded and installed "before" `rebound`, it fails to load because there is no `librebound.so` in the virtual file system.

I think the `reboundx` needs to bundle `librebound.so` inside its wheel. However, we can try to fix this issue on Pyodide side.

The possible solution can be:

1. When `pyodide.loadPackage()` is called, load dependencies first, then load the requested package.
2. Separate out downloading and dynlib loading process, so downloading all packages happens first, then loading dynlibs happens next.

Personally, I think option 2 is better since if we use option 1, it might be tough to handle cases like `pyodide.loadPyodide(["reboundx", "rebound"])`. The downside of option 2 is that there is a potential slowdown in loading packages.

### To Reproduce

```
await loadPackage("reboundx")
```

![image](https://user-images.githubusercontent.com/24893111/184254798-379a2dce-1c3b-4068-98b4-3c984ee66794.png)

### Environment

- Pyodide Version: 0.21.0

cc: @hannorein

hoodmane:
Maybe a mixed method would work: try to load straight away, but if a dependency is missing throw an error and try again after all packages are downloaded?

hannorein:
Is this just an error message in the console or does the package not work? Is it possible the package itself works fine despite the error message?

Here's a short test to see if it works:

```
import rebound, reboundx
sim = rebound.Simulation()
sim.add(m=1.)
sim.add(m=3e-3,a=4)
sim.add(m=5e-3,a=10)
rebx = reboundx.Extras(sim)
mof = rebx.load_force(“modify_orbits_forces")
rebx.add_force(mof)
sim.particles[2].params["tau_a"] = -200000
sim.particles[2].params["tau_e"] = -400000
sim.integrate(1e5)
```

ryanking13:
[...]

PR #2963, opened 2022-08-12
Retry when loading a package is failed

ryanking13:
### Description

Fixes: #2960 

If loading a package fails, retry it after all dependencies are downloaded.

image

For example,

1. `reboundx` depends on a shared library `librebound.so` that is in `rebound` package.
2. In the first attempt, it fails to find `librebound.so` inside the virtual file system.
3. Retry loading `reboundx` after `rebound` is downloaded, now `librebound.so` is available.

Note that 404 error happens because Emscripten loadDynamicLibrary function [tries to search the file from network](https://github.com/emscripten-core/emscripten/blob/c881fde312988ed466f0cea2a4a4d1d8e5d7403f/src/library_dylink.js#L804-L808) when it is not available on the file system.

### Checklists


hoodmane:
Maybe we should throw an error in `findObject` or `findFile` as appropriate if the file isn't found to prevent the network request since we don't want it to make the request.

ryanking13:
> Maybe we should throw an error in `findObject` or `findFile` as appropriate if the file isn't found to prevent the network request since we don't want it to make the request.

Yeah, we can try that, I think it can also help us get some more informative error messages. I'll try it in #2954 after this PR.

ryanking13:
@hoodmane, @rth Can I ask for a review on this?

src/js/load-package.ts
```diff
@@ -441,21 +440,46 @@ export async function loadPackage(
...
-    if (Object.keys(failed).length > 0) {
-      const failedNames = Object.keys(failed).join(", ");
+
+    if (failed.size > 0) {
+      const failedNames = Array.from(failed.keys()).join(", ");
       messageCallback(`Failed to load ${failedNames}`);
-      for (const [name, err] of Object.entries(failed)) {
+      for (const [name, err] of failed) {
         console.warn(`The following error occurred while loading ${name}:`);
         console.error(err);
```
hoodmane, line 484:
```suggestion
        errorCallback(`The following error occurred while loading ${name}:`);
        errorCallback(err);
```

src/js/load-package.ts
```diff
@@ -441,21 +440,46 @@ export async function loadPackage(
...
+          failed.set(name, err);
         });
     }
     await Promise.all(Object.values(packageInstallPromises));
 
+    // Retry failed packages after all dependencies are downloaded.
+    const installPromisesRetry: { [name: string]: Promise<void> } = {};
+    const toLoadRetry = new Map(
+      [...toLoadShared, ...toLoad].filter(([name, channel]) => failed.has(name))
+    );
```
hoodmane, line 452: The other way here is a bit more efficient:
```suggestion
    const toLoadRetry = new Map(
      failed.keys().map(k => [k, toLoadShared[k] || toLoad[k]])
    );
```

src/js/load-package.ts
```diff
@@ -441,21 +440,46 @@ export async function loadPackage(
...
+    for (const [name, channel] of toLoadRetry) {
+      console.warn(
```
hoodmane, line 454:
```suggestion
      errorCallback(
```

src/js/load-package.ts
```diff
@@ -441,21 +440,46 @@ export async function loadPackage(
...
         })
         .catch((err) => {
-          console.warn(err);
-          failed[name] = err;
+          failed.set(name, err);
         });
     }
     await Promise.all(Object.values(packageInstallPromises));
 
+    // Retry failed packages after all dependencies are downloaded.
```
hoodmane, line 448:
```suggestion
    // Retry failed packages after all dependencies are downloaded: If an `.so` file 
    // from package A requires an `.so` file from package B at load time, it's possible
    // that package B is loaded before package A and fails. This is a hacky way of 
    // reordering the two dependencies (in addition to the toLoadShared, toLoad 
    // separation). Both approaches are only bandaids: What if there is a three layer
    // load time dependency?
```

hoodmane (approved):
I don't like this approach: this adds an extra hack (in addition to the `toLoadShared` hack) that makes the problem go away in a special case (one stage load time dependencies) without actually solving the general problem. Eventually we will see packages that still don't work (as soon as package A depends on package B depends on package C at load time this won't work).

In the meantime I suppose this is okay.

hoodmane:
I am looking into a better solution to this.

hoodmane:
I take it back this is a general solution if a bit hacky. I will add a year test. It also needs a better comment to explain how it works since I keep forgetting.

rth:
Wouldn't it be better to mark `rebound` as a shared library (well maybe in a different sense than what we have now), and do not call `loadDynamicLibrary` on .so in `reboundx` until `rebound` is fully loaded? Retrying stuff for something that should be handled by dependency resolution doesn't sound optimal indeed.

It sounds like this is a similar situation with CLAPACK & scipy, or am I missing something? I did also run into related issues when playing with pyodide-pack and trying to load all .so from a single archive.

ryanking13:
> as soon as package A depends on package B depends on package C at load time this won't work

I think this case will work, because retrying happens after downloading all .so files. But yes, this is really hacky way of solving the problem and I don't like it very much.

> Wouldn't it be better to mark rebound as a shared library (well maybe in a different sense than what we have now), and do not call loadDynamicLibrary on .so in reboundx until rebound is fully loaded? Retrying stuff for something that should be handled by dependency resolution doesn't sound optimal indeed.

We can try that, but it will add another hack: marking a python package as a shared library. I'll see if that works.

> It sounds like this is a similar situation with CLAPACK & scipy, or am I missing something? I did also run into related issues when playing with pyodide-pack and trying to load all .so from a single archive.

I don't know much about CLAPACK & scipy issue, could you explain more about it?

ryanking13:
Aside of marking `rebound` as a shared library. I think this problem can happen between shared libraries.

If a shared library A depends on another shared library B, I think we can face this problem again.
The general solution would be downloading all dependency set first, than loading them after all downloads are complete. But I'm not sure how much it will influence the load time.

hoodmane:
I think the approach here is sound. #3020 is a bit more satisfying but it is probably slower.

ryanking13:
> I think the approach here is sound. #3020 is a bit more satisfying but it is probably slower.

Let's check how much #3020 will affect the load time and decide which one we should use.

PR #2966, opened 2022-08-13
Add package `pycryptodomex`

Starz0r:
### Description

Add cryptographic package `pycryptodomex`

### Checklists

- [x] Add / update tests


hoodmane (approved):
Thanks @Starz0r!

ryanking13:
@Starz0r we found that pycryptodomex was not building because the metadata file was not named correctly: `meta.yml` (should be `meta.yaml`). We fixed the name issue, but the package has a build failure so we disabled it (#3006) for now.

mbledkowski:
> @Starz0r we found that pycryptodomex was not building because the metadata file was not named correctly: `meta.yml` (should be `meta.yaml`). We fixed the name issue, but the package has a build failure so we disabled it (#3006) for now.

Any updates?

ryanking13:
@mbledkowski

No, I don't have any knowledge about `pycryptodomex` so one who have interest and knowledge about pycryptodomex would need to drive the work.

mbledkowski:
@ryanking13 I am interested in packaging pycryptodomex myself, but I do not really understand how the development process should look like. I have meta.yaml and test file, I built it, and according to documentation, I am now supposed to test this package using pyodide REPL, but there is no detail information about how to do this. Maybe this is something obvious, but I am not that proficient in python syntax, and simple import do not work, help("modules") do not return the module i just built.
Can you redirect me to resource that would help me in this case, or tell me where can I get in contact with people that know more about pyodide.
Best regards, mble

ryanking13:
@mbledkowski

I am sorry if our documentation isn't detailed enough, and we're very happy that you're interested in building packages! Pyodide's package building system is actually quite complicated to explain because it's a mix of the WASM compilation toolchain and the complex Python packaging system, and because different Python packages have different build methods (which is an old problem with Python), so you often need to understand how the package you're building is built rather than how Pyodide itself is built.

> I have meta.yaml and test file, I built it, and according to documentation, I am now supposed to test this package using pyodide REPL, but there are no detailed information about how to do this.

Looking at the other issue you opened (#4159), it looks like the build failed, and it's actually not uncommon for packages with complex native dependencies to fail to cross-compile, so we'll need to analyze why the build failed. If you can't analyze the reason for the failure yourself, the best thing to do is to post a draft PR with an updated meta.yaml file so that maintainers can see the full build log. It's hard to tell what we're dealing with by looking at the fragmentary errors in the issue, so it would be nice to see the full build log so that we can analyze it together.

---

On a side note, if your reason for wanting to build pycryptodomex is to use `yt-dlp`, I should probably start by saying that it will be harder than you might think (at least, it will not just work by building pycryptodomex). You might need to handle the following issues as well:

- WebSocket: Python websockets is not available in the browser, you will probably write a JS websocket wrapper for Python, similar to how [pyodide-http](https://github.com/koenvo/pyodide-http) does. However, there are no guarantees that the JS websocket API will provide the enough functionality that required.

- FFMPEG: AFAIK, yt-dlp uses ffmpeg to process videos. Although we are building a WASM port of the ffmpeg library, ffmpeg is used by running yt-dlp by calling a subprocess, which is not supported by Pyodide. So if you run into problems in that area, you will have to modify a lot of code in yt-dlp yourself.

- CORS: youtube does not have CORS headers set, so probably you need a set a CORS proxy server to send a request to youtube. This is relatively easy (See: https://bartbroere.eu/2023/04/22/combining-cors-proxies-with-pyodide/)

spence:
fyi, ffmpeg is available as a wasm library: https://github.com/ffmpegwasm/ffmpeg.wasm (though there may be [limitations](https://ffmpegwasm.netlify.app/docs/faq/))

PR #2967, opened 2022-08-13
Add websockets

Starz0r:
### Description

Add package `websockets`

### Checklists

- [x] Add / update tests


Starz0r:
Perhaps my test case is a bit too ambitious? I'm not entirely sure how to get multiple functions to be visible inside the test runner, including my global variable that needs to be set to stop the test from running infinitely.

hoodmane:
I think this package is unlikely to be able to work correctly in Pyodide, at least not without major patching.

Starz0r:
I don't think it's possible to run a websocket server from the browser? Don't think it can't be done from Node however, but in my mind, that seems like a good compromise.

Maybe if I could split where each tests are run that would be good enough? Client tests running through browsers, and Node running both Server and Client tests. I'd be down to refactor the tests into that.

ryanking13:
> Maybe if I could split where each tests are run that would be good enough? Client tests running through browsers, and Node running both Server and Client tests. I'd be down to refactor the tests into that.

If you can make server tests work, it will be amazing. But if it doesn't work, I think we can accept this PR with client tests only.

hoodmane:
To be clear, I don't believe that the client will work either.

joemarshall:
This is the sort of thing that the following pull request for pytest-pyodide is designed for - i.e. to make it possible to run normal tests in pyodide without having to add all the pyodide shenanigans in there.

https://github.com/pyodide/pytest-pyodide/pull/62

joemarshall:
I do wonder whether this should really be part of pyodide-http by the way - that's how requests and things that do networking are handled (and probably aiohttp at some point should be in there); i.e. acknowledging that anything that supports this kind of thing in browser is likely to be some kind of polyfill rather than a full port.

ryanking13:

Issue #2971, opened 2022-08-15
`ModuleNotFoundError: No module named 'distutils.msvccompiler'` in package build CI


hoodmane:
## 🐛 Bug

Several different packages have failed with `ModuleNotFoundError: No module named 'distutils.msvccompiler'`. See for instance:
https://app.circleci.com/pipelines/github/hoodmane/pyodide/3667/workflows/dd548c6b-f1c8-40d3-88f5-e71573bf4934/jobs/43716

@ryanking13 any ideas? Maybe we need to pin something?

ryanking13:
I think we need to pin the setuptools version. Setuptools embeds distutils in its source code and it seems like they updated something recently: 

https://github.com/pypa/setuptools/pull/3505

aminvakil:
FYI I also faced this issue and the problem with pinning setuptools is that some packages try to upgrade setuptools in their build process anyway:
```
Step 9/10 : RUN pip install --ignore-installed implicit==0.4.8 -v
 ---> Running in 8ed28a7f9227
Using pip 22.2.2 from /usr/local/lib/python3.8/site-packages/pip (python 3.8)
Collecting implicit==0.4.8
  Downloading implicit-0.4.8.tar.gz (1.1 MB)
     ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 1.1/1.1 MB 1.4 MB/s eta 0:00:00
  Installing build dependencies: started
  Running command pip subprocess to install build dependencies
  Collecting setuptools>=40.8.0
    Downloading setuptools-65.0.0-py3-none-any.whl (1.2 MB)
       ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 1.2/1.2 MB 1.4 MB/s eta 0:00:00
  Collecting wheel
    Downloading wheel-0.37.1-py2.py3-none-any.whl (35 kB)
  Installing collected packages: wheel, setuptools
  Successfully installed setuptools-65.0.0 wheel-0.37.1
  WARNING: Running pip as the 'root' user can result in broken permissions and conflicting behaviour with the system package manager. It is recommended to use a virtual environment instead: https://pip.pypa.io/warnings/venv
  Installing build dependencies: finished with status 'done'
  Getting requirements to build wheel: started
  Running command Getting requirements to build wheel
  Traceback (most recent call last):
[...]

PR #2972, opened 2022-08-16
Pin setuptools version temporarily

ryanking13:
Releated: #2971


ryanking13:
Merging since this is breaking all other PRs.

henryiii:
This shouldn't be needed anymore, setuptools restored the module removal.

ryanking13:
Thanks for the info!

PR #2973, opened 2022-08-16
Add an import hook for unvendored and removed stdlibs

ryanking13:
### Description

This PR adds an extra import hook that handles cases when one tries to import unvendored or removed stdlib modules. It raises `ModuleNotFoundError` with more informative error messages instead of just showing "No module named ..." 

Related: #2940

### Checklists

- [x] Add / update tests

### DIFF
--- a/packages/_ssl/meta.yaml
+++ b/packages/_ssl/meta.yaml
@@ -1,5 +1,5 @@
 package:
-  name: ssl
+  name: _ssl
   version: 1.0.0 # Nonesense
   _cpython_dynlib: true
 source:
--- a/src/py/_pyodide/_importhook.py
+++ b/src/py/_pyodide/_importhook.py
@@ -133,7 +133,7 @@ def register_js_finder() -> None:
     sys.meta_path.append(jsfinder)
 
 
-class StdlibFinder(MetaPathFinder):
+class UnvendoredStdlibFinder(MetaPathFinder):
     def __init__(self) -> None:
         # `test`` is not a stdlib module, but we unvendors in anyway.
         self.stdlibs = sys.stdlib_module_names | {"test"}
@@ -154,18 +154,18 @@ class StdlibFinder(MetaPathFinder):
 
         if parent in self.unvendored_stdlibs:
             raise ModuleNotFoundError(
-                f"The module '{parent}' is unvendored from Pyodide stdlib, "
+                f"The module '{parent}' is unvendored from the Python standard library in the Pyodide distribution, "
                 f'you can install it by calling: await pyodide.loadPackage("{parent}"). '
                 "See https://pyodide.org/en/stable/usage/wasm-constraints.html for more details."
             )
         else:
             raise ModuleNotFoundError(
-                f"The module '{parent}' is removed from Pyodide stdlib "
+                f"The module '{parent}' is removed from the Python standard library in the Pyodide distribution "
                 "due to browser limitations. "
                 "See https://pyodide.org/en/stable/usage/wasm-constraints.html for more details."
             )
 
 
-def register_stdlib_finder() -> None:
+def register_unvendored_stdlib_finder() -> None:
     # Note: this finder must be placed in the end of meta paths.
-    sys.meta_path.append(StdlibFinder())
+    sys.meta_path.append(UnvendoredStdlibFinder())
--- a/src/py/_pyodide/_importhook.py
+++ b/src/py/_pyodide/_importhook.py
@@ -146,6 +146,10 @@ class StdlibFinder(MetaPathFinder):
         target: ModuleType | None = None,
     ) -> ModuleSpec | None:
         [parent, _, _] = fullname.rpartition(".")
+
+        if not parent:
+            return None
+
         if parent in self.unvendored_stdlibs:
             print(
                 f"The module '{parent}' is unvendored from Pyodide stdlib, "
--- a/src/py/_pyodide/_importhook.py
+++ b/src/py/_pyodide/_importhook.py
@@ -135,20 +135,11 @@ def register_js_finder() -> None:
 
 class StdlibFinder(MetaPathFinder):
     def __init__(self) -> None:
-        self.stdlibs = sys.stdlib_module_names
+        # `test`` is not a stdlib module, but we unvendors in anyway.
+        self.stdlibs = sys.stdlib_module_names | {"test"}
+
         # TODO: put list of unvendored stdlibs to somewhere else?
         self.unvendored_stdlibs = {"distutils", "test", "_ssl", "_lzma"} & self.stdlibs
-        self.removed_stdlibs = {
-            "curses",
-            "dbm",
-            "ensurepip",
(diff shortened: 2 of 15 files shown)


src/py/_pyodide/_importhook.py
```diff
@@ -131,3 +131,41 @@ def register_js_finder() -> None:
...
+        target: ModuleType | None = None,
+    ) -> ModuleSpec | None:
+        [parent, _, _] = fullname.partition(".")
+
+        if not parent or parent in sys.modules or parent not in self.stdlibs:
+            return None
+
+        if parent in self.unvendored_stdlibs:
+            raise ModuleNotFoundError(
+                f"The module '{parent}' is unvendored from Pyodide stdlib, "
```
ryanking13, line 157: I am not sure whether the wording "unvendored" and "removed" is easy to understand for new users.

src/py/_pyodide/_importhook.py
```diff
@@ -131,3 +131,41 @@ def register_js_finder() -> None:
     global JsProxy
     JsProxy = _pyodide_core.JsProxy
     sys.meta_path.append(jsfinder)
+
+
+class StdlibFinder(MetaPathFinder):
+    def __init__(self) -> None:
+        # `test`` is not a stdlib module, but we unvendors in anyway.
```
rth, line 138: Could just add a docstring with a short description of this class?

src/py/_pyodide/_importhook.py
```diff
@@ -131,3 +131,41 @@ def register_js_finder() -> None:
...
+    ) -> ModuleSpec | None:
+        [parent, _, _] = fullname.partition(".")
+
+        if not parent or parent in sys.modules or parent not in self.stdlibs:
+            return None
+
+        if parent in self.unvendored_stdlibs:
+            raise ModuleNotFoundError(
+                f"The module '{parent}' is unvendored from Pyodide stdlib, "
+                f'you can install it by calling: await pyodide.loadPackage("{parent}"). '
```
rth, line 158: Let's maybe rather use micropip (and assume it's either loaded or will be loaded via import), because loadPackage needs to be called from JS while most users will use this from Python and can only write Python.

src/py/_pyodide/_importhook.py
```diff
@@ -131,3 +131,41 @@ def register_js_finder() -> None:
...
+
+    def find_spec(
+        self,
+        fullname: str,
+        path: Sequence[bytes | str] | None,
+        target: ModuleType | None = None,
+    ) -> ModuleSpec | None:
+        [parent, _, _] = fullname.partition(".")
+
+        if not parent or parent in sys.modules or parent not in self.stdlibs:
```
rth, line 152: And so if it's in sys.modules another loader should handle it? Maybe worth a comment that generally when this branch is for continuing to default import resolution.

src/py/_pyodide/_importhook.py
```diff
@@ -131,3 +131,41 @@ def register_js_finder() -> None:
...
+        else:
+            raise ModuleNotFoundError(
+                f"The module '{parent}' is removed from Pyodide stdlib "
+                "due to browser limitations. "
+                "See https://pyodide.org/en/stable/usage/wasm-constraints.html for more details."
+            )
+
+
+def register_stdlib_finder() -> None:
+    # Note: this finder must be placed in the end of meta paths.
```
rth, line 170: Please add a docstring.

rth (commented):
Thanks @ryanking13 , this is very nice!

rth:
Also it wouldn't hurt to add an extra test for StdlibFinder more directly to make sure it works as expected when calling `StdlibFinder().find_spec(fullname, ...)` (including returning None)

src/py/_pyodide/_importhook.py
```diff
@@ -131,3 +131,41 @@ def register_js_finder() -> None:
...
+    sys.meta_path.append(StdlibFinder())
```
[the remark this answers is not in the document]
hoodmane: Well we don't do that in `register_js_finder` plus it's private and should only be called once on initialization, so I am not sure that it is worth it. 

It is worth pointing out that the setup with where we call `register_js_finder` is a bit peculiar because of the specifics of the bootstrap process so we could call `register_stdlib_finder` in a more natural place. But keeping them together does make a lot of sense.

src/py/_pyodide/_importhook.py
hoodmane: You could also say `pyodide_js.loadPackage` here. `from pyodide_js import loadPackage;  await loadPackage("some_package")` will work even if `micropip` isn't around.

packages/_ssl/meta.yaml
```diff
@@ -1,5 +1,5 @@
 package:
-  name: ssl
+  name: _ssl
```
hoodmane, line 2: This change will cause a bit of a regression with `loadPackageFromImports`. Currently `import ssl` in `console.html` works, but if you change the name to `_ssl` then it will fail.

src/py/_pyodide/_importhook.py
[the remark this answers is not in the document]
rth: Yeah, probably not necessary. It was just a thought.

packages/_ssl/meta.yaml
rth: Yes, good point. I misread it as being changed the other way around.

ryanking13: `import ssl` will still work in the console since `loadPackageFromImports`  uses `test/imports` key to determine the top-level import name and I didn't change it. Though we probably need to change its behavior later (https://github.com/pyodide/pyodide/pull/2581#discussion_r878958962)

src/py/_pyodide/_importhook.py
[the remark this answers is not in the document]
ryanking13: > Maybe check that's not already there before adding it, in case it's called twice for some reason?

Added a check.

> It is worth pointing out that the setup with where we call `register_js_finder` is a bit peculiar because of the specifics of the bootstrap process so we could call `register_stdlib_finder` in a more natural place. But keeping them together does make a lot of sense.

Yeah, it is not related to the bootstrap thing, but I just put it near the `regiest_js_finder`

src/py/_pyodide/_importhook.py
ryanking13: Changed to `micropip.install`. Maybe we can mention other ways in docs.

src/py/_pyodide/_importhook.py
ryanking13: > And so if it's in sys.modules another loader should handle it

That condition is needed when someone makes a mistake such as:

```
import os.submodule_that_not_exists
```

The default importer will fail (return None) and it will fall back to this importer.

hoodmane (approved):

rth:
Thanks again!

ryanking13:
Thanks for the reviews!

PR #2975, opened 2022-08-16
Add an Emscripten version check for out of tree builds

hoodmane:
The out of tree build system applies an abi tag based on `PYODIDE_EMSCRIPTEN_VERSION` in `Makefile.envs`. Prior to this PR, it will happily accept the version of emscripten (say 3.1.18) and produce a wheel tagged like it was produced from the expected Emscripten version (say 3.1.14). This causes confusing errors due to ABI mismatch see https://github.com/igraph/python-igraph/issues/560. This error message should help prevent such confusion.

### DIFF
--- a/pyodide-build/pyodide_build/tests/test_common.py
+++ b/pyodide-build/pyodide_build/tests/test_common.py
@@ -130,3 +130,58 @@ def test_search_pyodide_root(tmp_path):
     pyproject_file.unlink()
     with pytest.raises(FileNotFoundError):
         search_pyodide_root(tmp_path)
+
+
+def test_check_emscripten_version(monkeypatch):
+    from pyodide_build import common
+
+    s = None
+
+    def get_emscripten_version_info():
+        return s
+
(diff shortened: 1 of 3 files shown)


ryanking13 (approved):
LGTM, Thank!

pyodide-build/pyodide_build/tests/test_common.py
```diff
@@ -130,3 +130,58 @@ def test_search_pyodide_root(tmp_path):
...
+"""
+
+    with pytest.raises(
+        RuntimeError,
+        match=f"Incorrect Emscripten version 1.39.20. Need Emscripten version {needed_version}",
+    ):
+        common.check_emscripten_version()
+
+    s = """\
+emcc (Emscripten gcc/clang-like replacement + linker emulating GNU ld) 3.1.14 (4343cbec72b7db283ea3bda1adc6cb1811ae9a73)
```
ryanking13, line 171: We need to fix hard-coded version here

PR #2976, opened 2022-08-16
Add command to create Pyodide virtual environment

hoodmane:
This is work in progress. To use:

1. `pip install pyodide-build`
2. `pyodide venv .venv-pyodide`
3. `source .venv-pyodide/bin/active`
4. `pip install .[tests]` or `pip install -r test_requirements.txt` etc
5. `python -m pytest`


rth:
Very cool! We should probably pre-install micropip here by default, same way as venv pre-installs pip. 

It's very impressive that `python -m virtualenv`  works and it's important to have the python executable for a number of applications (e.g. IDE) but I wonder what should be the final user interface for this, 
 a) I have the impression virtualenv is being more or less deprecated in favor of stdlib venv https://stackoverflow.com/a/47559925/1791279
 b) users probably shouldn't have to download the repo to install it. Maybe in the future, we should put this either in some Python (e.g. pyodide-cli) or npm package?

hoodmane:
It seems to me that the standard library `venv` is missing a lot of features present in `virtualenv`. Most critically:

>  The venv module does not offer all features of this library, to name just a few more prominent:
> * ...
> * cannot create virtual environments for arbitrarily installed python versions (and automatically discover these),
> * ...

https://virtualenv.pypa.io/en/latest/

I think for this reason it is unusable for us. I have just added a script 
[create_pyodide_venv.py](https://github.com/pyodide/pyodide/pull/2976/files#diff-6b3271910244461f7119f2f8335f5f0e9b57031f1a4deb7630081f4ee30a9f00) that sets everything up rather nicely.

rth:
Not reviewing as from what I understood the goal is to split it in separate chunks, but otherwise very much looking forward to this functionality!

.circleci/config.yml
```diff
@@ -315,6 +315,43 @@ jobs:
...
+          command: |
+            set -x
+            export PYODIDE_ROOT=$(pwd)
+            echo $PYODIDE_ROOT
+            cd ~
+            mkdir test
+            cd test
+            git clone https://github.com/python-attrs/attrs --depth 1 --branch 22.1.0
+            git clone https://github.com/zopefoundation/zope.interface.git --depth 1 --branch 5.4.0
+            git clone https://github.com/nedbat/coveragepy.git --depth 1 --branch coverage-5.5
```
rth, line 335: Why do we need these 3 ?

rth:
> To use:
>  pip install pyodide-build

I wonder if pyodide-build is the best place to put the CLI. It might be better to put it directly in `pyodide-cli` since this is unrelated to the build and doesn't need all the build related stuff that  `pyodide-build` might need. Also so we can start using the new typer CLI for new entrypoints. This means we would need to switch to `pyodide build` to register this subcommand in `pyodide-cli` via the plugin mechanism (to avoid the name clash), but we should do it anyway at some point.

.circleci/config.yml
hoodmane: Dependencies installed by `attrs[tests]`. Would be great to replace the build front end used by pip with `pyodide build` then it could do all this automatically. But it's not clear that's possible without unstable modifications to pip.

hoodmane:
> It might be better to put it directly in [pyodide-cli](https://github.com/pyodide/pyodide-cli)

Sounds good to me. For this PR I don't want to change it, we can move it as a follow-up.

hoodmane:
This is ready for review. I think I'll add a few more tests eventually but I think it should be in good shape (at least aside from any dumb mistakes I just made refactoring venv.py).

src/py/pyodide/_package_loader.py
```diff
@@ -22,8 +23,14 @@
...
 SITE_PACKAGES = Path(getsitepackages()[0])
-STD_LIB = Path(sysconfig.get_path("stdlib"))
-TARGETS = {"site": SITE_PACKAGES, "lib": STD_LIB, "dynlib": Path("/usr/lib")}
+if sys.base_prefix != sys.prefix:  # Are we in a virtualenv?
+    STD_LIB = Path(sysconfig.get_path("stdlib"))
+    TARGETS = {"site": SITE_PACKAGES, "lib": STD_LIB, "dynlib": Path("/usr/lib")}
+else:
+    # Better not put stuff into /usr/lib or /lib/python3.10! For now let's stick
+    # everyone into SITE_PACKAGES in this case
+    TARGETS = {"site": SITE_PACKAGES, "lib": SITE_PACKAGES, "dynlib": SITE_PACKAGES}
```
ryanking13, line 32: Is this because we mount host `/usr` directory when using a command line runner?

pyodide-build/pyodide_build/out_of_tree/venv.py
```diff
@@ -0,0 +1,287 @@
...
+    # TODO: Also load all shared libs.
+    result = subprocess.run(
+        [
+            venv_bin / "python",
+            "-c",
+            dedent(
+                """
+                from _pyodide._importhook import UNVENDORED_STDLIBS_AND_TEST;
+                from pyodide_js import loadPackage;
+                loadPackage(UNVENDORED_STDLIBS_AND_TEST);
```
ryanking13, line 245: Do we need to load test?

pyodide-build/pyodide_build/out_of_tree/venv.py
```diff
@@ -0,0 +1,287 @@
...
+    # Python in the shebang. Use whichever Python was used to invoke
+    # pyodide venv.
+    host_python_path = venv_bin / "python3.10-host"
+    (venv_bin / "pip").write_text(
+        # Other than the shebang and the monkey patch, this is exactly what
+        # normal pip looks like.
+        dedent(
+            f"""
+            #!{host_python_path}
+            # -*- coding: utf-8 -*-
```
ryanking13, line 138: Is this still required? IIRC this was a Python 2 stuff.

ryanking13 (commented):
Thanks, @hoodmane! I was a little overwhelmed about the size of this PR initially, but after some separate PRs, now it seems like it has a reasonable size of changes.

I think it is generally in a good shape. I'll try to review this more thoroughly in a couple of days unless Roman does.

src/py/pyodide/_package_loader.py
hoodmane: Yeah. We want to load them permanently but not into global user directories. If we didn't mount `/usr` then we'd also have the problem that we'd install them into memory and throw them away. Maybe there is a more appropriate path for libraries inside the virtual environment.

hoodmane: Actually I think the emfs `/usr/lib` may shadow the native one.

pyodide-build/pyodide_build/out_of_tree/venv.py
hoodmane: I guess we don't need to but I don't see why not do so. It's around in the native Python.

pyodide-build/pyodide_build/out_of_tree/venv.py
hoodmane: Probably not, I just copied from pip.

hoodmane: Removed it:
https://stackoverflow.com/questions/14083111/should-i-use-encoding-declaration-in-python-3

src/py/pyodide/_package_loader.py
ryanking13: Okay, maybe we can think of installing all libraries into site-packages even if we are not in virtualenv, which is reverting the change that I made in #2721.

IIRC why I changed the installation directory was that shapely was trying to find geos from '/usr/lib'. But I think I can try to make a fix in #2954.

pyodide-build/pyodide_build/out_of_tree/venv.py
ryanking13: Oh, I was misunderstanding the code here. I thought that this needs to be called every time, but it looks like it only happens onetime at the virtualenv initialization step. Nice!

hoodmane:
@henryiii if I include `type: ignore[import]` it complains that it is not needed but if I don't include it it complains that it doesn't know what virtualenv is.
```py
    from virtualenv import session_via_cli  # type: ignore[import]
```

henryiii:
You should add it to pyproject.toml at https://github.com/pyodide/pyodide/blob/555f7828f4339b056f2d32b821e6a752784f4729/pyproject.toml#L25

henryiii:
(One of the runs is not strictly checked for imports, and it seems this file is being checked in both MyPy runs. You might want to exclude it from one of them.)

.circleci/config.yml
rth: Why do we need `attrs[tests]` though? Can we at least use `attrs[tests-no-zope]` without zope?

src/py/pyodide/_package_loader.py
```diff
@@ -22,8 +23,14 @@
 from ._core import IN_BROWSER, JsProxy, to_js
 
 SITE_PACKAGES = Path(getsitepackages()[0])
-STD_LIB = Path(sysconfig.get_path("stdlib"))
-TARGETS = {"site": SITE_PACKAGES, "lib": STD_LIB, "dynlib": Path("/usr/lib")}
+if sys.base_prefix != sys.prefix:  # Are we in a virtualenv?
```
rth, line 26: Maybe,
```suggestion
if sys.base_prefix != sys.prefix:  # We are in a virtualenv
```
because otherwise there is a doubt which branch corresponds to the virtualenv and it's not very obvious from the code.

Also here aren't the two branches inverted,  https://stackoverflow.com/a/1883251/1791279 ?

> The most reliable way to check for this is to check whether sys.prefix == sys.base_prefix. If they are equal, you are not in a virtual environment; if they are unequal, you are.

So they are unequal in the first branch, meaning we are in a venv, while we used this code before outside of it.

src/tests/test_cmdline_runner.py
```diff
@@ -104,3 +108,223 @@ def test_invalid_cmdline_option(selenium):
...
+        == dedent(
+            """
+            Collecting certifi>=2017.4.17
+            Collecting charset-normalizer<3,>=2
+            Collecting idna<4,>=2.5
+            Collecting requests
+            Collecting urllib3<1.27,>=1.21.1
+            Installing collected packages: urllib3, idna, charset-normalizer, certifi, requests
+            Looking in links: /src/dist
+            Successfully installed certifi-2022.6.15 charset-normalizer-2.1.1 idna-3.3 requests-2.28.1 urllib3-1.26.12
```
rth, line 263: This is probably bound to break often with version updates. Maybe we could replace any `([\d+]\.?)+` by `*` to make it more robust.

src/tests/test_cmdline_runner.py
```diff
@@ -104,3 +108,223 @@ def test_invalid_cmdline_option(selenium):
...
+    # since these don't reproduce.
+    cleaned_stdout = re.sub("^  .*?\n", "", cleaned_stdout, flags=re.MULTILINE)
+    assert (
+        cleaned_stdout
+        == dedent(
+            """
+            Looking in links: /src/dist
+            Processing ./dist/regex-2022.6.2-cp310-cp310-emscripten_3_1_20_wasm32.whl
+            Installing collected packages: regex
+            Successfully installed regex-2022.6.2
```
rth, line 306: Same as above

rth (commented):
Looking forward to this! A couple of comments otherwise LGTM.

src/py/pyodide/_package_loader.py
hoodmane: Yes thanks. This is the reason for a bunch of the test failures.

pyodide-build/pyodide_build/out_of_tree/venv.py
```diff
@@ -0,0 +1,282 @@
...
+def create_pip_conf(venv_root: Path) -> None:
+    """Create pip.conf file in venv root
+
+    This file adds a few options that will always be used by pip install.
+    """
+    if in_xbuild_env():
+        # In the xbuild_env, we don't have the packages locally. We will include
+        # in the xbuildenv a PEP 503 index for the vendored Pyodide packages
+        # https://peps.python.org/pep-0503/
+        repo = f'extra-index-url={get_pyodide_root()/"dist/pypi_index"}'
```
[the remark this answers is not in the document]
hoodmane: Sounds reasonable. I originally had an implementation of this as part of this PR but I split it into another local branch in order to reduce the diff here.

hoodmane:
@henryiii I figured out the issue: if `venv` is imported from a test then the test mypy will process it. But we can fix this by using `importlib.importmodule` to import it because then mypy can't understand what happened.

pyodide-build/pyodide_build/out_of_tree/venv.py
```diff
@@ -0,0 +1,258 @@
...
+            """
+        )
+    )
+    (venv_bin / "pip").chmod(0o777)
+
+    pyversion = f"{sys.version_info.major}.{sys.version_info.minor}"
+    other_pips = [
+        venv_bin / "pip3",
+        venv_bin / f"pip{pyversion}",
+        venv_bin / f"pip-{pyversion}",
```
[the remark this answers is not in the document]
hoodmane: Apparently so, since `pip.unlink()` works. It seems harmless but I can remove it if you prefer.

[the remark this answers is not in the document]
hoodmane: Apparently so, since `pip.unlink()` works. It seems harmless but I can remove it if you prefer.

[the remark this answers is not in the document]
rth: Well we can keep it, until someone complains (if ever) :)

pyodide-build/pyodide_build/out_of_tree/venv.py
```diff
@@ -0,0 +1,258 @@
...
+        capture_output=True,
+        encoding="utf8",
+    )
+    check_result(result, "ERROR: failed to install unvendored stdlib modules")
+
+
+def create_pyodide_venv(dest: Path) -> None:
+    """Create a Pyodide virtualenv and store it into dest"""
+    print("Creating Pyodide virtualenv at", str(dest))
+    from virtualenv import session_via_cli  # type: ignore[import]
```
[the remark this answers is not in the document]
hoodmane: Yeah this use case is out of scope for `venv` because it only creates virtual environments for the interpreter it is run in. So `python-exe -m venv` makes a virtual environment for `python-exe`. Since `venv` won't work correctly in Pyodide, it isn't really useful. virtualenv handles it without any trouble because it takes an interpreter parameter.

If it makes you feel any better, nox also uses virtualenv:
https://github.com/wntrblm/nox/blob/ffb4fca2b81a5905bf8dae674b33a8371da1c5f5/pyproject.toml#L30

hoodmane:
@rth @ryanking13 getting the error:
```
__main__.py: error: argument --rt/--runtime: invalid choice: 'src' (choose from 'firefox', 'chrome', 'safari', 'node', 'host', 'firefox-no-host', 'chrome-no-host', 'safari-no-host', 'node-no-host')
```

ryanking13:
> @rth @ryanking13 getting the error:
> 
> ```
> __main__.py: error: argument --rt/--runtime: invalid choice: 'src' (choose from 'firefox', 'chrome', 'safari', 'node', 'host', 'firefox-no-host', 'chrome-no-host', 'safari-no-host', 'node-no-host')
> ```

#3084 will fix it.

ryanking13:
Thanks a lot!

PR #2979, opened 2022-08-17
MAINT Build test and distutils stdlib modules with pyodide-build

ryanking13:
### Description

Move build process of test and distutils into packages directory so all unvendored stdlibs can be built in the same way by pyodide-build.

### Checklists

- [x] Add / update tests


ryanking13:
One minor side effect of this change is that the rebuilding Pyodide core becomes slightly faster as we don't rebuild test modules every time.

hoodmane (approved):
Generally looks good to me.

hoodmane:
> One minor side effect of this change is that the rebuilding Pyodide core becomes slightly faster as we don't rebuild test modules every time.

I've been doing `make dist/pyodide.asm.js` for a long time. I guess I'm not satisfied with how often the Makefile doesn't rebuild things.

PR #2980, opened 2022-08-17
loadPackage should say that packages downloaded from pypi come from pypi

hoodmane:
If micropip loads a package, we put it into `pyodide.loadedPackages` so that `loadPackage` will know about it. Currently, if one installs via `micropip` and then via `loadPackage`:
```py
from pyodide_js import loadPackage
import micropip
await micropip.install("pytest==7.1.1")
await loadPackage("pytest")
```
it will say `pytest already loaded from https://pythonhosted.org/longstuff`. This makes it say `pytest already loaded from pypi`.

### DIFF
--- a/packages/micropip/src/micropip/_micropip.py
+++ b/packages/micropip/src/micropip/_micropip.py
@@ -208,6 +208,9 @@ class WheelInfo:
             self.write_dist_info(
                 "PYODIDE_REQUIRES", json.dumps(sorted(x.name for x in self._requires))
             )
+        name = self.project_name
+        assert name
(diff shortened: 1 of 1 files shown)


ryanking13 (approved):
Thanks, @hoodmane. It LGTM, please update the changelog.

Maybe in the future, we need to change the `pyodide.loadedPackages` to use importlib to find installed packages. I think I once tried that a few months ago, but the performance of reading all files was the bottleneck.

hoodmane:
In #2976 I've added some extra code to populate `loadedPackages` from `importlib.metadata` on startup.

Issue #738, opened 2020-08-04
Native file system API


SimonBiggs:
The following two items were raised over at jyve:

* https://github.com/deathbeds/jyve/issues/47
* https://github.com/deathbeds/jyve/issues/46

Potentially it might be worth building these into pyodide for package installation and/or local file system access.

It looks like it'll be waiting upon emscripten to pick it up:

https://github.com/emscripten-core/emscripten/issues/10017

rth:
@hoodmane has a WIP solution for this using Chrome FileSystem API and  https://github.com/hoodmane/synclink in  https://github.com/hoodmane/worker-pyodide-console/tree/nativefs

ryanking13:
Probably https://github.com/emscripten-core/emscripten/pull/16804 will make things easier (at least for chrome users).

ryanking13:
This was resolved by #2987

westurner:
Thanks @ryanking13 ?

- [ ] DOC: how is this distinct from "Acessing local files"
https://jupyterlite.readthedocs.io/en/stable/howto/content/filesystem-access.html

Does this solvr for jupyterlab-git in WASM in JupyterLite?

And then also what about BusyBox ash, bash, or zsh filesystem access for
Terminals in JupyterLite?

PR #2987, opened 2022-08-18
ENH Add API for mounting native file system

ryanking13:
### Description

Related: https://github.com/pyodide/pyodide/issues/738

This PR adds an API `pyodide.mountNativeFS` which mounts [FileSystemDirectoryHandle](https://developer.mozilla.org/en-US/docs/Web/API/FileSystemDirectoryHandle) into a Pyodide Python file system.

Note that there are some limitations such as:

- File System Access API is only supported by Chromium-based browsers (Safari also implements a portion of it but they only support Origin Private File system so it is not very useful for common users).

- The file system is *asynchronous*, so one needs to call [`syncfs` to persist changes](https://emscripten.org/docs/api_reference/Filesystem-API.html#FS.syncfs).

Since it is asynchronous, it does not require any WebWorker, SharedArrayBuffer, or Atomics. But I think we can extend this to a synchronous version using those features. Since Hood already made a [reference implementation of synchronous native file system access](https://github.com/hoodmane/worker-pyodide-console) I believe it is possible. Maybe we can use either [synclink](https://github.com/hoodmane/synclink) by Hood or [comsync](https://github.com/alexmojaki/comsync) by @alexmojaki.

The native fs implementation is adapted from https://github.com/emscripten-core/emscripten/pull/16804 (thanks to @connorjclark), the upstream PR is blocked because Emscripten team is working on WasmFS which will possibly replace MEMFS, but I think we can use it until WasmFS becomes more stable.

### Checklists

- [x] Add / update tests


alexmojaki:
> Maybe we can use either [synclink](https://github.com/hoodmane/synclink) by Hood or [comsync](https://github.com/alexmojaki/comsync) by @alexmojaki.

comsync is probably not the right tool for this, but [sync-message](https://github.com/alexmojaki/sync-message) (which is the base of comsync) might be. Last I spoke to @hoodmane, it seemed like the plan was to integrate sync-message into synclink.

> The file system is asynchronous, so one needs to call [syncfs to persist changes](https://emscripten.org/docs/api_reference/Filesystem-API.html#FS.syncfs).

I see this code in the tests:

```python
        pathlib.Path("/nativefs/test_write").write_text("hello_write")
        pathlib.Path("/nativefs/test_write").read_text() == "hello_write"
```

Does that mean that `/nativefs/test_write` is being updated synchronously in memory, but this won't be seen on the native FS until `syncfs` is called? That still seems really great. If we can just call `syncfs` after every call to Python code, no one will notice anything wrong unless they check the files outside of Pyodide while the Python code is still running, which I'm guessing will be very rare.

ryanking13:
> comsync is probably not the right tool for this, but [sync-message](https://github.com/alexmojaki/sync-message) (which is the base of comsync) might be. Last I spoke to @hoodmane, it seemed like the plan was to integrate sync-message into synclink.

Thanks for the correction! I haven't seen the detail of sync-message and comsync yet, so I'll have to learn how it works.

> Does that mean that /nativefs/test_write is being updated synchronously in memory, but this won't be seen on the native FS until syncfs is called?

Exactly.

>  If we can just call syncfs after every call to Python code, no one will notice anything wrong unless they check the files outside of Pyodide while the Python code is still running,

Yes, that can be an option. I'm not sure how much overhead it costs for syncing files every time. Note that an asynchronous file system is also used in IDBFS, but I haven't seen the benchmark for `syncfs`.

ryanking13:
@rth @hoodmane Can I ask for a review on this?

src/tests/test_filesystem.py
```diff
@@ -105,3 +105,146 @@ def test_idbfs_persist_code(selenium_standalone):
...
+
+@pytest.mark.xfail_browsers(
+    node="Not available", firefox="Not available", safari="Not available"
+)
+def test_nativefs_dir(request, selenium_standalone):
+
+    if request.config.option.runner == "playwright":
+        pytest.xfail("Playwright doesn't support file system access APIs")
+
+    selenium = selenium_standalone
```
hoodmane, line 118: Should these tests check whether the file appears on the local file system?

hoodmane (approved):
Looks good to me. Sorry for being so slow to review. Thanks @ryanking13!

src/tests/test_filesystem.py
ryanking13: Yes it has tests about that: `dirHandleMount.entries()` will list files in the local file system and I added a check that files generated from Python(Pyodide) are visible in  `dirHandleMount.entries()`

hoodmane: Shouldn't you be able to see them in the actual file system outside of selenium too? Also how do these tests handle the directory picker?

ryanking13: That's a good point. Actually, the test in this PR uses [OPFS (Origin Private File System)](https://webkit.org/blog/12257/the-file-system-access-api-with-origin-private-file-system/) which uses IndexDB not a real local file system.

I wasn't able to add a test with *real* local file system, using `showDirectoryPicker()` API. That API requires user gesture and I wasn't able to find a way to test it using selenium.

hoodmane: Great could you add a comment saying this to the test? Then I'm +1 to merge it.

ryanking13:

PR #2995, opened 2022-08-20
Backports for v0.21.1

hoodmane:
Don't squash these.

@rth @ryanking13


ryanking13 (approved):
Thanks for the backporting work!

Issue #2402, opened 2022-04-18
RFC Provide a way to download packages for Node


ryanking13:
## 🚀 Feature

From the discussion (https://github.com/pyodide/pyodide/pull/2394#discussion_r851565711), it would be good to provide an official way to download Pyodide packages into the local filesystem for Node users, so that they don't need to manually download it from e.g. Github release.

Example: `pyodide install` (`npx pyodide install`)

For a reference, other npm packages that need to download external binaries do:

- [puppeteer](https://github.com/puppeteer/puppeteer): has a [install script](https://github.com/puppeteer/puppeteer/blob/2b45027d256f85f21a0c824183696b237e00ad33/package.json#L29) which downloads Chromium browser binary when doing `npm install puppeteer`.

- [playwright](https://github.com/microsoft/playwright): has a command `playwright install` to download browser binaries (and ffmpeg).

rth:
I previously opened a similar issue in https://github.com/pyodide/pyodide-cli/issues/3 as that is now a better repo for it. So I'm going to close this one, and copy-paste the above proposal there.

PR #2996, opened 2022-08-20
Rename released bundle from pyodide-build.tar.gz to pyodide.tar.gz

hoodmane:
It's confusing that we call the release `pyodide-build.tar.gz`. This renames it to `pyodide.tar.gz`.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -96,7 +96,7 @@ jobs:
       - run:
           name: Zip build directory
           command: |
-            tar cjf pyodide-build.tar.gz dist
+            tar cjf pyodide.tar.gz dist
 
       - persist_to_workspace:
           root: .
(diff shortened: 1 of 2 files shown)


hoodmane:
Actually might be useful to split into two zip files:
`pyodide.tar.gz`: `pyodide.js`, `pyodide.asm.js`, `pyodide.asm.wasm`, `pyodide_py.tar`, and `repodata.json`
`pyodide-packages.tar.gz`: the packages

this would make it a lot easier to download the development version of Pyodide for use in node.

ryanking13 (approved):
Thanks, @hoodmane!

> Actually might be useful to split into two zip files:
>pyodide.tar.gz: pyodide.js, pyodide.asm.js, pyodide.asm.wasm, pyodide_py.tar, and repodata.json
> pyodide-packages.tar.gz: the packages

> this would make it a lot easier to download the development version of Pyodide for use in node.

That sounds good for node, but it might be a little annoying for people who simply want to download a single archive and run a server (which I often do). Anyway, I agree that it would be nice if we have a better way to download packages from node (#2402)

hoodmane:
Maybe we can just add another release called `pyodide-minimal.tar.gz`?

ryanking13:
> Maybe we can just add another release called `pyodide-minimal.tar.gz`?

Sounds good to me.

PR #2997, opened 2022-08-21
Add then, catch, and finally_ methods to future

### LINKED ISSUE #2923: 'Future' object has no attribute 'then'
## 🐛 Bug

If I understand correctly, starting with version 0.17, Pyodide integrates the implementation of `await` for JsProxy. So when JS returns a `Promise`, it converts it to `Future` in Python, which allows us to use `await`.

Hoverer, the `Future` object has no attribute `then`, so it is no longer possible to build `then`/`catch` chains like in older versions.

### To Reproduce

```python
import js
js.fetch('http://karay.me/truepyxel/test.json').then(lambda resp: resp.json()).then(lambda data: data.msg).catch(lambda err: 'there were error: '+err.message)
```

### Environment

- Pyodide Version: 0.20
- Browser version: 103.0.5060.134

hoodmane:
Resolves #2923.

### Checklists

- [x] Add / update tests


ryanking13 (approved):
Thanks a lot, @hoodmane. It looks very cool.

It would be nice if there is a test coverage on real JS promises. For now, there are host tests only.

PR #2998, opened 2022-08-22
Do not load unvendored stdlibs by default

### LINKED ISSUE #2940: [Discussion] Changing the default behavior of loading unvendored stdlib

ryanking13:
Close #2940 

### Description

As discussed in #2940, this PR changes the default value of the `fullStdLib` flag in `loadPyodide` to false.
This is a breaking change because for now `distutils` is not loaded by default.

Also, when `fullStdLib` is set to true, it loads all unvendored stdlibs except for test.

### Checklists

- [x] Add / update tests


src/js/pyodide.ts
```diff
@@ -319,7 +319,10 @@ export async function loadPyodide(
     throw new Error("Lock file version doesn't match Pyodide version");
   }
   if (config.fullStdLib) {
-    await pyodide.loadPackage(["distutils"]);
+    let unvendoredStdlibs = Array.from(
+      API._pyodide._importhook.UNVENDORED_STDLIBS
+    );
+    await pyodide.loadPackage(unvendoredStdlibs);
```
hoodmane, line 325: `pyodide.loadPackage` accepts a `pyproxy`:
https://github.com/pyodide/pyodide/blob/main/src/js/load-package.ts#L356
```suggestion
    await pyodide.loadPackage(API._pyodide._importhook.UNVENDORED_STDLIBS);
```

src/py/_pyodide/_importhook.py
```diff
@@ -150,15 +154,7 @@ class UnvendoredStdlibFinder(MetaPathFinder):
...
-
-        # TODO: put list of unvendored stdlibs to somewhere else?
-        self.unvendored_stdlibs = {
-            "distutils",
-            "test",
-            "ssl",
-            "lzma",
-            "sqlite3",
-        } & self.stdlibs
+        self.unvendored_stdlibs = UNVENDORED_STDLIBS_AND_TEST & self.stdlibs
```
hoodmane, line 157: Could you add a comment to explain the point of this? Shouldn't everything in `UNVENDORED_STDLIBS` be in `self.stdlibs`?

hoodmane (approved):
Thanks @ryanking13!

src/py/_pyodide/_importhook.py
ryanking13: > Shouldn't everything in UNVENDORED_STDLIBS be in self.stdlibs?

Yes UNVENDORED_STDLIBS should be a subset of stdlibs. I just wanted to make sure that we don't add non-stdlibs by mistake.

src/js/pyodide.ts
ryanking13: Thanks!

src/py/_pyodide/_importhook.py
hoodmane: Could we make that into an assertion then?

ryanking13: Sounds good.

docs/project/changelog.md
```diff
@@ -33,6 +33,12 @@ substitutions:
   importing unvendored or removed stdlib modules fails.
   {pr}`2973`
 
+- {{ Breaking }} The default value of `fullStdLib` in {any}`loadPyodide` has been
+  changed to `false`. This means Pyodide now will not load distutils by default.
```
rth, line 37: distutils, sqlite3 and ssl. + better to link to the docs for more details.

docs/project/changelog.md
```diff
@@ -33,6 +33,12 @@ substitutions:
...
+  If `fullStdLib` is set to `true`, it will load all unvendored stdlib modules.
```
rth, line 38: Maybe add a note that it's preferable to load the required modules explicitly, instead of using `fullStdLib=true`.

docs/usage/wasm-constraints.md
```diff
@@ -10,13 +10,18 @@ or via [patches](https://github.com/pyodide/pyodide/tree/main/cpython/patches).
...
 ### Optional modules
 
-The following stdlib modules are included by default, however
-they can be excluded with `loadPyodide({ fullStdLib : false })`.
-Individual modules can then be loaded as necessary using
-{any}`pyodide.loadPackage`,
+The following stdlib modules are unvendored by default,
+in order to reduce initial download size of Python distribution.
+You can load all unvendored stdlib modules
+when initializing Pyodide with, `loadPyodide({ fullStdLib : true })`.
```
rth, line 16: "however this will have a significant implement on the download size.
Instead, it is better to load individual modules as needed using ... "

rth (approved):
Thanks!

PR #2999, opened 2022-08-22
Add extra minimal release, also add xbuildenv to artifacts

hoodmane:
See #2996.

### Checklists


rth (approved):
How about rather `pyodide-core` to match what we have in CI?  
Otherwise LGTM, thanks!

hoodmane:
Thanks for the review @rth!

PR #3004, opened 2022-08-22
DOC Switch to Plausible analytics in the docs

rth:
Due to the same reasons as in https://github.com/pyodide/pyodide-blog/pull/33

Also adds analytics to console.html (only when it's deployed as part of the docs).

The Google Analytics is  configured in readthedocs, so I'll disable it once this is merged.

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -154,7 +154,7 @@ if IN_READTHEDOCS:
             # insert the analytics script after the `pyodide.js` script
             console_html.insert(
                 idx,
-                '<script defer data-domain="pyodide.org" src="https://plausible.io/js/plausible.js"></script>"\n',
+                '<script defer data-domain="pyodide.org" src="https://plausible.io/js/plausible.js"></script>\n',
             )
             break
     else:
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -146,6 +146,21 @@ if IN_READTHEDOCS:
         encoding="utf-8",
     )
     print(res)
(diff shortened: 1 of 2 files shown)


docs/conf.py
```diff
@@ -146,6 +146,21 @@ def delete_attrs(cls):
...
     print(res)
+    # insert the Plausible analytics script to console.html
+    console_path = Path("_build/html/console.html")
+    console_html = console_path.read_text().splitlines(keepends=True)
+    for idx, line in enumerate(list(console_html)):
+        if 'pyodide.js">' in line:
+            # insert the analytics script after the `pyodide.js` script
+            console_html.insert(
+                idx,
+                '<script defer data-domain="pyodide.org" src="https://plausible.io/js/plausible.js"></script>\n',
```
ryanking13, line 157: Maybe we can think of having two copies of `console.html` in the future, instead of doing complicated "replace" stuff. It will increase the maintenance burden, but we can remove all this stuff including PYODIDE_BASE_URL.

ryanking13 (approved):
Thanks @rth!

docs/conf.py
rth: yes, this is not ideal. Not sure about two copies as there is some amount of code there and it would be annoying to ask contributors to make the change twice. I think a better way is to use jinja2 templating more in the future, which would be easier once the main makefile is in python (or otherwise we have to use `jinja2-cli`).

PR #3005, opened 2022-08-22
DOC Fix typo in format_exception example

patrick91:
The code was importing `format_exception` but then it wasn't using it 😊


ryanking13 (approved):
Thanks @patrick91!

Issue #3001, opened 2022-08-22
Bug: sqlite3 does not depend on nltk and regex


rth:
In 0.21.1 deployment,  when running `import sqlite3` in the REPL it would load  nltk and regex which is wrong.  

It looks like for sqlite3 `depends` and `imports` fields are switched, between the [repodata.json](https://cdn.jsdelivr.net/pyodide/v0.21.1/full/repodata.json) and the [meta.yaml](https://github.com/pyodide/pyodide/blob/main/packages/nltk/meta.yaml)
![image](https://user-images.githubusercontent.com/630936/185934857-c1c910ce-15a4-4426-982d-cb5de8582f87.png)

but then other packages seem to be fine.

This is serious enough to probably warrant a 0.21.2 release, as it adds an extra MB for apps that load sqlite3.

Version which reproduces: 0.20.0, 0.21.1 (and probably those in between)
Version which don't reproduce: dev

cc @ryanking13

rth:
No actually the meta.yaml was just wrong previously https://github.com/pyodide/pyodide/blob/0.21.1/packages/nltk/meta.yaml

hoodmane:
Looks like I added it in this commit
https://github.com/hoodmane/pyodide/commit/24e5f413c7a6125fbb3c770e1be3d76beb06120f
as part of #1637. Naturally I gave no additional explanation and as part of such a large pr it got no attention.

rth:
Less critical but sharedlib-test import is also wrong
![image](https://user-images.githubusercontent.com/630936/185938370-48ef16b3-0488-4c90-999d-5b56d7e46ee6.png)

it's not specified in meta.yaml, so probably something wrong with the code logic.

It might be good to add a validation to the dependency tree that,
 1. import fields are valid imports (e.g. cannot contain "-")
 2. that they are unique (e.g. `sqlite3` import cannot be present in two packages).

hoodmane:
Also mkpkg should substitute `-` for `_` in the imports field since this often is enough to give a correct result.

rth:
> Looks like I added it in this commit

and I missed it during review. Well it's not that critical in a sense it only happens if users `import sqlite3` in runPythonAsync, and they might not realize it's happening. But since it's a core package and it downloads 1MB extra in this case, it might still be nice to fix.

BTW, I run into this while testing sqlite3 unvendoring and being on the wrong version. So the outcome was rather unexpected :)

ryanking13:
[...]

PR #3006, opened 2022-08-23
Parse top-level import name from wheel file

ryanking13:
### Description

Changes how `imports` key in repodata.json is calculated. This key is used in `loadPackagesFromImport` to get the top-level import name of a package.

Previously we were using `test/imports` key in meta.yaml to calculate this key, without mentioning this in docs. Therefore it was very easy to make a mistake such as #3001.

This PR changes this by calculating top-level imports dynamically from the wheel file:

~~1) Try to parse the top-level import name from `top_level.txt` file inside the wheel file. This file exists if the package is built with setuptools, but may not exist if the package is built with flit, poetry, etc.~~
2) Iterate the top-level directories and files inside the wheel file, and add to top-level import if there are directories containing python file or if there is a top-level python file.
3) If both of the above fails, use the package name inside meta.yaml as a top-level import name (e.g. shared libraries)

~~FYI these packages don't have top_level.txt:~~

~~- packages that use flit: `pyparsing`, `more_itertools`, `tomli`, `tomli_w`, `threadpoolctl`, `typing_extensions`~~
~~- packages that use hatchling: `jsonschema`, `soupsieve`,~~
~~- packages that use poetry: `pkgconfig`
~~- packages that use distutils: `distlib` (changed to setuptools in latest release)~~

### Checklists

- [x] Add / update tests


pyodide-build/pyodide_build/buildall.py
```diff
@@ -477,6 +478,48 @@ def _generate_package_hash(full_path: Path) -> str:
...
+        # so even setuptools may remove this someday.
+        return top_level_file.read_text().strip().split("\n")
+
+    # If there is no top_level.txt file, we will find top level imports by
+    # 1) a python file on a top-level directory
+    # 2) a sub directory with __init__.py
+    # following: https://github.com/pypa/setuptools/blob/d680efc8b4cd9aa388d07d3e298b870d26e9e04b/setuptools/discovery.py#L122
+    top_level_imports = []
+    for subdir in whlzip.iterdir():
+        if subdir.is_file() and subdir.name.endswith(".py"):
```
rth, line 507: We should also skip if the subdir if it cannot be an import name (e.g. contains `-` or spaces) I think.

Otherwise currently it finds a folder with `-` as an import name.

pyodide-build/pyodide_build/buildall.py
```diff
@@ -477,6 +478,48 @@ def _generate_package_hash(full_path: Path) -> str:
...
 
 
+def parse_top_level_import_name(whlfile: Path) -> list[str] | None:
+    """
+    Parse the top-level import name from a package file.
+    """
+
+    assert whlfile.name.endswith(".whl"), "Not a package file"
+
+    pkg_name_ver = "-".join(whlfile.name.split("-")[:2])
```
rth, line 488: Better to use `from packaging.utils import parse_wheel_filename`

rth (commented):
Thanks! The nice thing is it also catches private imports (e.g. `_pytest`). It's an interesting approach, but note that are a number of problematic imports it finds below. Some are easy to fix, others are harder. Also this is hard to notice / review for users so I wonder whether the current system (but possibly with auto-generated list of imports by `mkpkg` using this code) so it could be reviewed would be more reliable.

image

test and tests should be explicitly excluded,
image

image

Not sure about those `lib*` folders, but maybe not critical,
image

image

image

I stopped at "l" letter in `repodata.json`. I imagine there are more of the same types of issues later.

ryanking13:
> but note that are a number of problematic imports it finds below. Some are easy to fix, others are harder.

Right... it seems like the `top_level.txt` is not very reliable so probably we should iterate directories even if `top_level.txt` exist. Thanks for pointing that out.

> so I wonder whether the current system (but possibly with auto-generated list of imports by `mkpkg` using this code) so it could be reviewed would be more reliable.

Yeah, that would be more flexible.

ryanking13:
The issue with auto-generating top level imports in mkpkg is that the final directory structure is not visible until we build a wheel.

ofek:
> Iterate the top-level directories and files inside the wheel file, and add to top-level import if there are directories containing __init__py or if there is a top-level python file.

Don't forget about implicit namespace packages https://packaging.python.org/en/latest/guides/packaging-namespace-packages/#native-namespace-packages

packages/attrs/meta.yaml
```diff
@@ -1,6 +1,9 @@
 package:
   name: attrs
   version: 21.4.0
+  top-level:
+    - attr
+    - attrs
```
ryanking13, line 6: Added `attrs`

packages/biopython/meta.yaml
```diff
@@ -1,6 +1,9 @@
 package:
   name: biopython
   version: "1.79"
+  top-level:
+    - Bio
+    - BioSQL
```
ryanking13, line 6: Added `BioSQL`.

packages/matplotlib/meta.yaml
```diff
@@ -1,7 +1,10 @@
 package:
   name: matplotlib
   version: 3.5.2
-
+  top-level:
+    - pylab
```
ryanking13, line 5: matplotlib has a top-level file `pylab.py`. Not sure about the usage of it.

packages/pycryptodomex/meta.yaml
```diff
@@ -1,12 +1,14 @@
 package:
   name: pycryptodomex
   version: 3.15.0
+  top-level:
+    - Cryptodome
```
ryanking13, line 5: Intestingly, this file was named wrong (`meta.yml`), so this package was not built.

packages/pyrsistent/meta.yaml
```diff
@@ -1,11 +1,15 @@
 package:
   name: pyrsistent
   version: 0.18.1
+  top-level:
+    - _pyrsistent_version
```
ryanking13, line 5: It has a top-level version file...

packages/pytest/meta.yaml
```diff
@@ -1,6 +1,9 @@
 package:
   name: pytest
   version: 7.1.2
+  top-level:
+    - _pytest
```
ryanking13, line 5: Added `_pytest`.

packages/pyyaml/meta.yaml
```diff
@@ -1,6 +1,9 @@
 package:
   name: pyyaml
   version: "6.0"
+  top-level:
+    - _yaml
```
ryanking13, line 5: Added `_yaml`.

packages/ruamel.yaml/meta.yaml
```diff
@@ -1,6 +1,8 @@
 package:
-  name: ruamel
+  name: ruamel.yaml
```
ryanking13, line 2: ruamel.yaml is a correct package name.

packages/scikit-image/meta.yaml
```diff
@@ -1,6 +1,9 @@
 package:
   name: scikit-image
   version: 0.19.3
+  top-level:
+    - skimage
+    # - doc  # scikit-image has top level `doc` directory... but it is too common name so we don't want to use it as a hint for top-level import.
```
ryanking13, line 6: scikit-image have top-level doc directory, but the name doc is too common so I think we don't want to load `scikit-image` is someone do `import doc`.

packages/setuptools/meta.yaml
```diff
@@ -1,6 +1,10 @@
 package:
   name: setuptools
   version: 62.6.0
+  top-level:
+    - _distutils_hack
```
ryanking13, line 5: setuptools has `_distutils_hack`.

packages/sympy/meta.yaml
```diff
@@ -1,6 +1,9 @@
 package:
   name: sympy
   version: 1.10.1
+  top-level:
+    - isympy
```
ryanking13, line 5: sympy has top-leve `isympy.py` file.

packages/toolz/meta.yaml
```diff
@@ -1,9 +1,18 @@
 package:
   name: toolz
   version: 0.11.2
+  top-level:
+    - tlz
```
ryanking13, line 5: Added `tlz`.

ryanking13 (commented):
Updated such as:

1. Add a new key `package/top-level` in meta.yaml which is used for top-level import names.
2. `package/top-level` is auto calculated in `mkpkg` command. This is done by downloading a wheel file of a native platform using its directory structure. (I hope the directory structure doesn't vary between platforms...)
3. If a wheel file doesn't exist, it will add a placeholder so one can manually add top-level import names.

To make a review easier, I added comments on packages that have changes.

packages/pycryptodomex/meta.yaml
ryanking13: It seems like the build is faling, disabled for now.

hoodmane:
Sorry for the delay on review for this @ryanking13. I'll try to look at it when I get a chance. Seems like it's in good shape.

packages/Jinja2/meta.yaml
```diff
@@ -10,3 +12,8 @@ source:
 test:
   imports:
     - jinja2
```
rth, line 14: I think when tested imports matches `top-level` (probably most cases), we should remove `test.imports` and fallback to `package.top-level` to avoid redundant information in the `meta.yaml`

packages/more-itertools/meta.yaml
```diff
@@ -1,9 +1,16 @@
...
   sha256: c5122bffc5f104d37c1626b8615b511f3427aa5389b94d61e5ef8236bfbc3ddb
   url: https://files.pythonhosted.org/packages/bd/3f/c4b3dbd315e248f84c388bd4a72b131a29f123ecacc37ffb2b3834546e42/more_itertools-8.13.0-py3-none-any.whl
 test:
   imports:
     - more_itertools
+about:
+  home: ""
+  PyPI: https://pypi.org/project/more-itertools
+  summary: More routines for operating on iterables, beyond itertools
+  license: ""
```
rth, line 16: Can remove both home and license if they are empty

packages/pytest/meta.yaml
```diff
@@ -16,4 +19,10 @@ requirements:
     - iniconfig
 test:
   imports:
+    - _pytest
```
rth, line 22: I don't think we need to test it, it's an internal feature of pytest. All we care that pytest import works (and it likely wouldn't if it can't import _pytest)

packages/pyyaml/meta.yaml
```diff
@@ -17,4 +20,10 @@ requirements:
     - libyaml
 test:
   imports:
+    - _yaml
```
rth, line 23: Similarly, not needed here IMO

packages/scikit-image/meta.yaml
rth: Reported to https://github.com/scikit-image/scikit-image/issues/6505

pyodide-build/pyodide_build/common.py
```diff
@@ -91,6 +93,49 @@ def find_matching_wheels(wheel_paths: Iterable[Path]) -> Iterator[Path]:
                 yield wheel_path
 
 
+def parse_top_level_import_name(whlfile: Path) -> list[str] | None:
+    """
+    Parse the top-level import name from a package file.
+    """
+
+    assert whlfile.name.endswith(".whl"), "Not a package file"
```
rth, line 101: Assert should not be used to check user input https://stackoverflow.com/a/17531188/1791279

Better to explicitly raise some exception if the conditions False.

pyodide-build/pyodide_build/common.py
```diff
@@ -91,6 +93,49 @@ def find_matching_wheels(wheel_paths: Iterable[Path]) -> Iterator[Path]:
                 yield wheel_path
 
 
+def parse_top_level_import_name(whlfile: Path) -> list[str] | None:
+    """
+    Parse the top-level import name from a package file.
```
rth, line 98:
```suggestion
    Parse the top-level import names from a wheel file.
```

pyodide-build/pyodide_build/mkpkg.py
```diff
@@ -63,12 +69,15 @@ def _find_sdist(pypi_metadata: MetadataDict) -> URLDict | None:
     return None
 
 
-def _find_wheel(pypi_metadata: MetadataDict) -> URLDict | None:
+def _find_wheel(pypi_metadata: MetadataDict, native: bool = False) -> URLDict | None:
     """Get wheel file path from the metadata"""
+    if native:
+        predicate = lambda filename: filename.endswith(".whl")
+    else:
+        predicate = lambda filename: filename.endswith("py3-none-any.whl")
```
rth, line 77: maybe,
```suggestion

    predicate = lambda filename: filename.endswith(".whl" if native else "py3-none-any.whl")
```

rth (approved):
A coupe of comments, otherwise LGTM. Thanks!

packages/Jinja2/meta.yaml
ryanking13: Sounds good to me

ryanking13:
Thanks for the thorough reviews!

PR #3009, opened 2022-08-24
Remove socketmodule from ssl

ryanking13:
### Description

I don't see why we are building socketmodule in ssl. We already have `_socket` vendored.

### DIFF
--- a/packages/ssl/meta.yaml
+++ b/packages/ssl/meta.yaml
@@ -10,13 +10,11 @@ build:
     mkdir dist
     export DISTDIR=$(pwd)/dist
     cd $CPYTHONBUILD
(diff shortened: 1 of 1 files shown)


hoodmane (approved):
Thanks @ryanking13.

ryanking13:
> Should we be vendoring socket? Does it provide any features that actually work?

I'm not sure. People can use some constants e.g. `socket.AF_INET` but most functionalities won't work. threading and multiprocessing are in the same situation, and I'm not very sure whether unvendoring them would break some packages or not.

rth:
> threading and multiprocessing are in the same situation, and I'm not very sure whether unvendoring them would break some packages or not.

Yes, it would: we had that experience previously.. A lot of packages import multiprocessing or threading somewhere in the critical path but then either use them in very localized way of for functionality that works (e.g. using locks)

PR #3010, opened 2022-08-24
Unpin setuptools version

ryanking13:
Close: #2971 

> FYI, 65.0.2 restored these removed modules.

Thanks to Henry for the information


rth (approved):
Thanks!

PR #3020, opened 2022-08-26
Load packages in topologically sorted order

### LINKED ISSUE #2960: Importing reboundx fails occasionally

ryanking13:
### Description

Trying to fix #2960 in a more fundamental way by loading packages in sorted order.

I'm not sure how it would affect the package load time... added a simple benchmark for testing package load time.

### Checklists


hoodmane:
Could we benchmark all package load times as part of the import tests? Maybe as a separate PR?

ryanking13:
> Could we benchmark all package load times as part of the import tests? Maybe as a separate PR?

Okay, I'll separate a PR. Not sure how much time it will take to run import tests for all packages though.

hoodmane:
Don't we already run import tests for all packages?

ryanking13:
Yes we do and it is the most time taking part in our test. I think if we want to benchmark the import time, we need to run imports multiple times, which will multiply the test time.

pyodide-build/pyodide_build/buildall.py
```diff
@@ -522,11 +522,6 @@ def generate_packagedata(
             }
             packages[name.lower() + "-tests"] = pkg_entry
 
-    # Workaround for circular dependency between soupsieve and beautifulsoup4
-    # TODO: FIXME!!
-    if "soupsieve" in packages:
-        packages["soupsieve"]["depends"].append("beautifulsoup4")
```
ryanking13, line 528: I removed this because circular dependency is causing trouble and soupsieve docs mention that one needs to install beautifulsoup4 manually.(https://github.com/facelessuser/soupsieve/commit/9a792216ea1ffce018c7968105540c08b82b96db)

src/js/load-package.ts
```diff
@@ -354,104 +434,75 @@ export async function loadPackage(
...
-          await installPackage(name, buffer, channel);
-          loaded.push(name);
-          loadedPackages[name] = channel;
-        })
-        .catch((err) => {
-          console.warn(err);
-          failed[name] = err;
-        });
+      // TODO: add support for prefetching modules by awaiting on a promise right
+      // here which resolves in loadPyodide when the bootstrap is done.
```
hoodmane, line 481: I don't understand this comment.

hoodmane (commented):
Nice work @ryanking13. I am definitely not sorry to see `toLoadShared` removed. If the tests pass and benchmarks look okay then I am +1 to merge this.

src/js/load-package.ts
ryanking13: Well that comment is originally written by you: https://github.com/pyodide/pyodide/blame/939b615df00b1dd46d463f22c5fcaf01bd483da2/src/js/load-package.ts#L412

I guess you wanted to add a feature to prefetch modules when loadPyodide is called?

hoodmane: Ah okay I understand what I meant now. That probably would be good to do someday.

packages/soupsieve/meta.yaml
```diff
@@ -4,6 +4,9 @@ package:
...
   url: https://files.pythonhosted.org/packages/16/e3/4ad79882b92617e3a4a0df1960d6bce08edfb637737ac5c3f3ba29022e25/soupsieve-2.3.2.post1-py3-none-any.whl
-test:
-  imports:
-    - soupsieve
+# Importing soupsieve without installing beautifulsoup4 will fail.
+# But we don't want to add beautifulsoup4 as a dependency, because
+# it creates a circular dependency.
+# test:
+#   imports:
+#     - soupsieve
```
rth, line 12: +1 but this probably needs a changelog entry.

src/js/load-package.ts
```diff
@@ -230,6 +266,50 @@ async function installPackage(
   }
 }
 
+/**
+ * Download then install the package.
+ * Downloads can be done in parallel, but installs must be done following the dependency.
```
rth, line 271: "for dependencies first" or something similar

rth (commented):
It looks like the benchmarks are rather similar to main. Though I'm not sure how trustworthy they are in CI, given things like,
```
test_import[firefox-scikit-learn]        118.7177 (31.75)    118.7177 (36.92)    118.7177 (26.48)     0.0000 (1.0)    
test_import[firefox-scipy]               195.0509 (52.17)    195.0509 (60.66)    195.0509 (43.51)     0.0000 (1.0)  
```
and scikit-learn which should take longer to load than scipy given that it's its dependency.

Overall +1 for this approach as well. It would be good to explain a bit more what is done and also change the PR title. It's not really sorted order but rather dependency resolution order as far as I understand.

hoodmane:
> It's not really sorted order but rather dependency resolution order as far as I understand.

"topologically sorted" is what I would call it

hoodmane:
I do think that except for shared libraries we would only need the dependencies to be downloaded not loaded.

ryanking13:
> and scikit-learn which should take longer to load than scipy given that it's its dependency.

Yes, that is because scipy has [multiple import tests](https://github.com/pyodide/pyodide/blob/0df5f790952d6cf413ef99493c27d1765a1ea8c0/packages/scipy/meta.yaml#L114-L152). I agree that the benchmarks are not very reliable anyway.

> It's not really sorted order but rather dependency resolution order as far as I understand.

> "topologically sorted" is what I would call it

Thanks for the suggestions. I always have trouble writing a brief but concise PR title 🙄

ryanking13:
> I do think that except for shared libraries we would only need the dependencies to be downloaded not loaded.

But we have some abnormal cases like rebound... I think there is some room to optimize this more, like dividing steps into download, extract, and load.

hoodmane:
> I always have trouble writing a brief but concise PR title 

I think naming things is one of the hardest things in programming and it has to help me a bit that English is my first language.

hoodmane:
Thanks for working on this @ryanking13! It's nice to see it improving but it's quite hard work.

PR #3021, opened 2022-08-27
Add argument handling and _run_main function

hoodmane:
The command line runner in #2976 finally works, but it is a large change set so I am planning to split it up. This is the first PR split off from there.

This PR adds a patch to Python to expose `pymain_run_python`. We cannot use the public API `Py_RunMain` for this purpose because it finalizes the Python interpreter with `Py_FinalizeEx` when it is done. If we start an async task with `Py_RunMain` then it will segfault. `pymain_run_python` does a large amount of work and reproducing it is undesirable.

I added an `args` parameter which accepts command line arguments. The private entry point `pyodide._main._run_main()` executes the main Python entrypoint without shutting down the interpreter and returns the return code.

### Checklists

- [x] Add / update tests


hoodmane:
@tiran

tiran:
Do you have a traceback for the segfault?

hoodmane:
I can produce it but it just looks like trying to call python apis when the interpreter is null (because of `Py_Finalize`).

rth (approved):
>  but it is a large change set so I am planning to split it up

Thanks!

Well to see with Christian, if this could be better resolved in CPython, but as a temporary solution, this sounds fine to me.

PR #3022, opened 2022-08-27
Don't install standard library into the site-packages directory

hoodmane:
Usually the Python standard library is not installed into site-packages. This is important because if we use a virtual environment, then `/lib/python3.10/site-packages` directory will not be on the import path but `/lib/python3.10` will be.

Split off from #2976.

### Checklists

- [x] Add / update tests


ryanking13 (approved):
Thanks!

rth:
It might be worth adding a changelog entry for this in some later PR in case people relied on the current paths.

hoodmane:
Okay, I will do that. I don't suspect anyone will notice or care though...

PR #3023, opened 2022-08-27
MAINT Improve handling of fatal startup errors

hoodmane:
This moves the fatal startup errors into a separate function and removes the `"oops"` error message.

### DIFF
--- a/src/core/error_handling.ts
+++ b/src/core/error_handling.ts
@@ -127,6 +127,31 @@ API.fatal_error = function (e: any) {
   throw e;
 };
 
+class FatalPyodideError extends Error {}
+Object.defineProperty(FatalPyodideError.prototype, "name", {
+  value: FatalPyodideError.name,
+});
+
+API.fatal_loading_error = function (...args: string[]) {
+  let message = args.join(" ");
+  if (Module._PyErr_Occurred()) {
+    const FS = Module.FS;
+    // Redirect stderr
+    FS.closeStream(2);
+    FS.unlink("/dev/stderr");
+    const tracebackcodes: number[] = [];
+    FS.createDevice("/dev", "stderr", null, (e: number) =>
+      tracebackcodes.push(e)
+    );
+    FS.open("/dev/stderr", 1);
+    // Print traceback to stderr, decode it, and add it to message
+    Module._PyErr_Print();
+    let traceback = new TextDecoder().decode(new Uint8Array(tracebackcodes));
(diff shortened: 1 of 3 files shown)


hoodmane:
The new exception looks like:
```py
FatalPyodideError: Failed to import _pyodide module
  File "/lib/python3.10/_pyodide/__init__.py", line 20, in <module>
    blah()
  File "/lib/python3.10/_pyodide/__init__.py", line 19, in blah
    f()
  File "/lib/python3.10/_pyodide/__init__.py", line 18, in f
    raise Exception("oops")
Exception: oops
```

src/core/error_handling.ts
```diff
@@ -127,6 +127,42 @@ API.fatal_error = function (e: any) {
...
+
+let stderr_chars: number[] = [];
+API.capture_stderr = function () {
+  stderr_chars = [];
+  const FS = Module.FS;
+  FS.createDevice("/dev", "capture_stderr", null, (e: number) =>
+    stderr_chars.push(e)
+  );
+  FS.closeStream(2);
+  FS.open("/dev/capture_stderr", 1);
```
ryanking13, line 143: Interesting. So we close stderr fd (2), then open another device so it gets the just closed stderr fd...

src/core/error_handling.ts
```diff
@@ -127,6 +127,42 @@ API.fatal_error = function (e: any) {
...
+};
+
+API.restore_stderr = function () {
+  const FS = Module.FS;
+  FS.closeStream(2);
+  FS.unlink("/dev/capture_stderr");
+  FS.open("/dev/stderr", 1);
```
ryanking13, line 150: I guess [`1` is `O_WRONLY`](https://github.com/emscripten-core/emscripten/blob/e05e72d9c49fe15578e73934ce525a894d1b712a/system/lib/libc/musl/include/fcntl.h#L49)? Could you add a comment on that or it will be hard to understand in a first glance.

ryanking13 (approved):
Thanks @hoodmane!

src/core/error_handling.ts
hoodmane: Yes. As far as I can tell there is no obvious better way to do this (`open` can't specify a file descriptor, `dup` isn't available from the JavaScript API and `FS.createStream` is a bit low level). Probably deserves a comment.

hoodmane:

PR #3024, opened 2022-08-27
Handle case when Python startup code calls exit

hoodmane:
If Python startup code calls `exit` (for instance because `--version` or `--help` was passed), then Emscripten calls `quit` which throws the error. The enclosing environment catches that error and burns the evidence. This modifies `quit` to record the fact that `exit` was called and rethrow the error (which will include the exit code). Otherwise, trying to run the command line runner with `python --version` will segfault.

### DIFF
--- a/src/js/module.ts
+++ b/src/js/module.ts
@@ -3,7 +3,7 @@ export interface Module {
   noImageDecoding: boolean;
   noAudioDecoding: boolean;
   noWasmDecoding: boolean;
-  quit: (a: any, b: any) => void;
+  quit: (status: number, toThrow: Error) => void;
   preRun: { (): void }[];
   print: (a: string) => void;
   printErr: (a: string) => void;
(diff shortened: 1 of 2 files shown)


src/js/module.ts
```diff
@@ -23,6 +24,10 @@ export function createModule(): any {
   Module.noWasmDecoding = false; // we preload wasm using the built in plugin now
   Module.preloadedWasm = {};
   Module.preRun = [];
+  Module.quit = (status: any, toThrow: any) => {
```
ryanking13, line 27:
```suggestion
  Module.quit = (status: number, toThrow: Error) => {
```

src/js/module.ts
```diff
@@ -3,6 +3,7 @@ export interface Module {
   noImageDecoding: boolean;
   noAudioDecoding: boolean;
   noWasmDecoding: boolean;
+  quit: (a: any, b: any) => void;
```
ryanking13, line 6:
```suggestion
  quit: (status: number, toThrow: Error) => void;
```

ryanking13 (approved):
Thanks @hoodmane!

hoodmane:
Well if something goes wrong initializing the interpreter, certain fatal error execution paths will call exit. This PR is an improvement to never-should-happen failure cases like #3023. Before this PR, the call to exit is swallowed but the Python interpreter is left uninitialized and the first attempt to call a Python api segfaults. So this PR should make it a bit easier to tell where the failure occurred.

PR #3027, opened 2022-08-27
Emscripten 3.1.20

hoodmane:
GSW patch:
https://github.com/TEOS-10/GSW-Python/pull/99

galpy patch:
https://github.com/jobovy/galpy/pull/499

pyerfa patch:
https://github.com/liberfa/pyerfa/pull/89

astropy patch:
https://github.com/astropy/astropy/pull/13586

scipy patch:
https://github.com/scipy/scipy/pull/16924


hoodmane:
@ryanking13 could you review this? I think it's good to merge.

packages/pyerfa/meta.yaml
```diff
@@ -4,6 +4,10 @@ package:
 source:
   url: https://files.pythonhosted.org/packages/78/fd/0148f0e54f0c6f48a141409df65d74a5f1dae2e139f23d50a43c58c16098/pyerfa-2.0.0.1.tar.gz
   sha256: 2fd4637ffe2c1e6ede7482c13f583ba7c73119d78bef90175448ce506a0ede30
+  patches:
+    - patches/0001-Fix-compilation-with-Wincompatible-function-pointer-.patch
+build:
+  script: python3.10 erfa_generator.py
```
ryanking13, line 10: What does this command do?

ryanking13 (approved):
Thanks, @hoodmane! It seems like every Emscripten update is a challenge.

packages/pyerfa/meta.yaml
hoodmane: See added comments.

hoodmane:
> It seems like every Emscripten update is a challenge.

Yeah it's always exciting. At least this was a relatively easy one.

rth:
Regarding that CPython version / Emscripten version policy, does it mean that we only enforce it from next version and not this one? Or do we aim to update CPython for the next release?

hoodmane:
I think we're starting from next release. Is that what we've decided @ryanking13?

I think it would be a good idea to work on some emscripten related automation. It would be nice to automatically try out each emscripten-releases commit...

PR #3028, opened 2022-08-28
0.21.2 backports

hoodmane:


ryanking13 (commented):
Thanks for the backporting work! Changes in `cpython/remove_modules.txt` of #2946 needs to be also backported.

rth:
Thanks for making this release! The changelog needs fixing though. I'll push a commit to stable with the final release date/version.

rth:
Also, we forgot to include a fix for https://github.com/pyodide/pyodide/issues/3001 which was actually rather critical and just a change to the meta.yaml file. It was in the 0.21.2 milestone. Sorry, I was traveling yesterday so didn't see it.

So I'm not sure what we should do now, between doing a 0.21.3 for it, or just leaving it as is.

ryanking13:
> Also, we forgot to include a fix for #3001 which was actually rather critical and just a change to the meta.yaml file. It was in the 0.21.2 milestone. Sorry, I was traveling yesterday so didn't see it.
>

Oops. Yes, it was fixed in #2946 but I forgot to tell we should include it.

hoodmane:
Might as well make an 0.21.3, it only takes a few minutes to make a release now (thanks again @ryanking13!)

PR #3030, opened 2022-08-29
TST Benchmark package load time

ryanking13:
benchmark test load time in import tests in `test_packages_common.py`. This PR aims to measure how #3020 affects the package load time.

Note that since package import tests take too long to run, I set the minimal number of repentance to 1.

### DIFF
--- a/packages/_tests/test_packages_common.py
+++ b/packages/_tests/test_packages_common.py
@@ -1,5 +1,6 @@
 import functools
 import os
+import time
 from typing import Any
 
 import pytest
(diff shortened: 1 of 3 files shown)


hoodmane (approved):
Thanks @ryanking13!

packages/_tests/test_packages_common.py
```diff
@@ -44,7 +46,14 @@ def test_parse_package(name: str) -> None:
 @pytest.mark.skip_refcount_check
 @pytest.mark.driver_timeout(60)
 @pytest.mark.parametrize("name", registered_packages())
-def test_import(name: str, selenium_standalone: _BrowserBaseRunner) -> None:
+@pytest.mark.benchmark(
+    max_time=3,
+    min_rounds=1,
+    timer=time.time,
```
rth, line 52: Better to use `time.perf_counter()` as  `time.time` is really imprecise (~16ms precision) on Windows though I guess we are not running there too much yet.

ryanking13: Thanks for suggestion, opened a PR.

PR #3039, opened 2022-08-30
Initialize loadedPackages from importlib distributions

hoodmane:
This initializes `pyodide.loadedPackages` from the file system `importlib.metadata.distributions`. This makes it possible to handle the case when the `site_packages` is persistent or packages are preinstalled there (say with `--preload-file`).

Split off from #2976.

### Checklists

- [x] Add / update tests

### DIFF
--- a/src/py/pyodide/_package_loader.py
+++ b/src/py/pyodide/_package_loader.py
@@ -6,6 +6,8 @@ import sysconfig
 import tarfile
 from collections.abc import Iterable
 from importlib.machinery import EXTENSION_SUFFIXES
+from importlib.metadata import Distribution
+from importlib.metadata import distributions as importlib_distributions
 from pathlib import Path
 from site import getsitepackages
 from tempfile import NamedTemporaryFile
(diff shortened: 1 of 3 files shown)


src/py/pyodide/_package_loader.py
```diff
@@ -299,6 +306,31 @@ def get_dynlibs(archive: IO[bytes], suffix: str, target_dir: Path) -> list[str]:
     ]
 
 
+def get_dist_source(dist: Distribution) -> str | None:
```
rth, line 309: Could you add a docstring with at least 1 sentence summary indicating what exactly is the source. I gather it's the package name (but that, requires reading the code). Also a short summary of the logic here would be helpful.

Also shouldn't this be at somewhat shared with micropip: I imagine the mechanism for parsing importlib distributions is similar.

rth (approved):
Thanks, one comment otherwise LGTM.

src/py/pyodide/_package_loader.py
hoodmane: It's used in `loadPackage` messages:
```js
messageCallback(`${pkg} already loaded from ${loaded}`);
```
https://github.com/pyodide/pyodide/blob/main/src/js/load-package.ts#L369
so it's primarily to explain to the user why we think the package has already been loaded. It's important that there be some entry in `loadedPackages` to prevent `loadPackage` from copying another version of the package over it.

PR #3040, opened 2022-08-30
Webloop in progress tracking

hoodmane:
More work split from #2976. This adds tracking for how many web loop handles are active. It also adds handlers for when `SystemExit` or `KeyboardInterrupt` are raised. The purpose of these is to allow us to keep the Python interpreter alive until all tasks are completed, unless `SystemExit` or `KeyboardInterrupt` is raised. 

See https://github.com/hoodmane/pyodide/blob/cmdline-runner/tools/python.js for their use in context.

### Checklists

- [x] Add / update tests

### DIFF
--- a/src/py/pyodide/webloop.py
+++ b/src/py/pyodide/webloop.py
@@ -140,6 +140,10 @@ class WebLoop(asyncio.AbstractEventLoop):
         asyncio._set_running_loop(self)
         self._exception_handler = None
         self._current_handle = None
+        self._in_progress = 0
+        self._no_in_progress_handler = None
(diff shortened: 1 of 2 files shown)


hoodmane:
Thanks for the review!

rth:
Thanks for working on this. 

> Also it's really annoying that the code coverage doesn't see test coverage from non host tests. 

Yeah, it would have been nice to try to extract the coverage information from within the WASM FS and append it to existing coverage information, though it's not easy, also because paths don't exactly match. Though maybe with what you are doing in the CLI runner and the node runner that would be more managable.

PR #3042, opened 2022-08-30
DOCS Add more info on Rust compilation

skelsec:
### Description
I was having a hard time getting the cryptography package compiled on the latest version in the docker container because of the special requirements to get Rust working.  
Added a small section to the documentation describing the steps I took to make it work.

### Checklists

### DIFF
--- a/docs/development/new-packages.md
+++ b/docs/development/new-packages.md
@@ -386,3 +386,27 @@ Currently it is necessary to run `source $CARGO_HOME/env` in the build script [a
 but other than that there may be no other issues if you are lucky.
 
 As mentioned [here](https://github.com/pyodide/pyodide/issues/2706#issuecomment-1154655224), by default certain wasm-related `RUSTFLAGS` are set during `build.script` and can be removed with `export RUSTFLAGS=""`.
+
+#### Setting up Rust in the docker container
+
+This part is for developers who use the docker image and wish to compile Python packages containing Rust code.
+If you clone the Pyodide repo from Github the docker container will not have `rust` installed. For this you'd need to install `rust` using the preferred method described [here](https://www.rust-lang.org/tools/install). 
+
+```
+apt update
+apt install curl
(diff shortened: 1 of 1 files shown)


hoodmane:
Thanks @skelsec. We should really get rid of `-Z link-native-libraries=no`. The problem is that Emscripten errors out if you try to compile with `-sSIDE_MODULE` and `-lc`. I think the behavior of normal compilers is to ignore `-lc` when linking dynamic libraries. So far I have been unable to convince @sbc100 to upstream a hack to handle this, but perhaps we should patch Emscripten to drop this and then we can use stable rust.

sbc100:
Is there some reason not to use `-Z link-native-libraries=no` ?   It seems like the right solution to me. 

When libc and libm are static libraries (as they are in emscripten) one does not link them into shared libraries.. otherwise one ends up with multiple copies of them.   I could be convinced otherwise I think, but it sounds like `-Z link-native-libraries=no` is the correct flag for building side modules.

hoodmane:
> Is there some reason not to use -Z link-native-libraries=no ?

It's nightly only and it's not clear that rust has any plans to stabilize it.

sbc100:
> > Is there some reason not to use -Z link-native-libraries=no ?
> 
> It's nightly only and it's not clear that rust has any plans to stabilize it.

So how does this work on platforms/SDKs that statically link libc and libm?   For example doesn't apline linux link libc statically by default?  How does rust work there?  Wouldn't linking with `-lc` -and `-lm` result ineach dso having its own copy of libc and libc?

hoodmane:
That is a good question, it would be good to look into it.

One big problem with `-Z link-native-libraries=no` is that to Rust "native" means not rust. So passing this option to get rid of libc will also foil attempts to statically link any C libraries. We just want to get rid of things like libc and libm. So I really think it's more of a hack and not something that works well in the long term.

PR #3043, opened 2022-08-30
Update scipy to 1.9.1 (no meson)

hoodmane:
This updates scipy to v1.9.1. This was mercifully easy:

* for now we disable meson
* we dropped `patches/0014-BUG-Fix-signature-of-D_IIR_forback-1-2.patch` since it was upstreamed
* we had to add a patch to put fitpack back into a shape that makes f2c happy
* we need one more `-Wincompatible-function-pointer-types` fix upstream PR: https://github.com/scipy/scipy/pull/16934

### Checklists


rgommers:
That all looks good. I'll have a look at your new upstreamed patch now.

PR #3044, opened 2022-08-30
Add command line runner main entry script

hoodmane:
This PR contains the changes from #3039 and #3040 which should be merged first. The purpose is to add the main command line script from #2976.

### Checklists

- [x] Add / update tests


hoodmane:
@rth @ryanking13 would appreciate review on this one next. Though perhaps I should add tests first.

ryanking13:
Thanks for your works! I won't be available for this weekend so I'll look into this in next monday or tuesday. I hope you have a nice weekend 🙏👍

hoodmane:
Thanks. You too!

tools/python.js
```diff
@@ -0,0 +1,242 @@
...
+    FS.open("/dev/stderr", 1);
+}
+
+/**
+ * Determine which native top level directories to mount into the Emscripten
+ * file system.
+ *
+ * This is a bit brittle, if the machine has a top level directory with certain
+ * names it is possible this could break. The most surprising one here is tmp, I
+ * am not sure why but if we link tmp then the process silently fails.
```
rth, line 130: Maybe because /tmp could be on a different filesystem, e.g on a tmpfs.

Why do we need to mount all the other dirs though? Wouldn't it be better / safer to mount only the current folder and its subdirs (or at least have an opt-in way to mount all of these).

A concern is that by now the users might be used that pyodide runs in a in memory FS,  including in node. Now if you give them the real FS, they might do things that break their system without realising they are not modifying the in memory FS.

tools/python.js
```diff
@@ -0,0 +1,242 @@
+#!/bin/sh
```
rth, line 1: Maybe calling it `python` would be better since it's not really a pure js file anyway?

rth (approved):
Thanks for working on this! A few comments, otherwise LGTM.

tools/python.js
```diff
@@ -0,0 +1,242 @@
...
+    // ARGS=-- helpfully is not empty but does nothing.
+    process.stdout.write("--");
+}
+EOF
+)")
+
+exec node "$ARGS" "$0" "$@"
+`;
+
+const { loadPyodide } = require("../dist/pyodide");
```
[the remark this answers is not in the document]
hoodmane: Possibly. My plan was to include this as part of the `xbuildenv` so then the structure would be:
```
tools/python.js
dist/pyodide{.asm.js, .asm.wasm, .js, _py.tar}
dist/repodata.json
cpython/installs/...
packages/.libs/...
```
and so this fixed path is good enough.

tools/python.js
hoodmane: Fine with me.

tools/python.js
hoodmane: The goal of the command line runner is to be as close as we can manage to a drop-in replacement for ordinary python. This works very poorly with the sandboxed file system. For example, in `python path/to/somefile.py`, Python is going to look up the file to execute in the Emscripten file system and if it isn't in there then it will fail. The same is true with any files in arguments. If we are going to have a virtual environment, the virtual environment needs to be included. `pytest` has to look up the test files and whatever other stuff it needs. In my tests these files are not typically in the current folder or its subdirectories. The details of this are a bit different between running locally, vs CircleCI or github actions.

I also don't think this is very useful except for testing that functionality works with Pyodide. I think aside from that, this is worse in every way than a native Python. 

Maybe we could set up an environment variable to specify which directories to mount though.

ryanking13 (approved):
Thanks @hoodmane! It looks very cool.

src/js/pyodide.ts
```diff
@@ -263,6 +265,8 @@ export async function loadPyodide(
     stderr?: (msg: string) => void;
     jsglobals?: object;
     args?: string[];
+    _node_mounts?: string[];
+    _working_directory?: string;
```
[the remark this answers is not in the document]
hoodmane: Good idea, I just didn't notice that setting.

hoodmane:
I'm going to merge this now. In the future it might be a good idea to add ways to control the import path of `pyodide.asm.js` or to change the isolation/unisolation of the emfs as @rth suggests.

PR #3046, opened 2022-08-31
Use time.perf_counter for benchmark

ryanking13:
As suggested from @rth in https://github.com/pyodide/pyodide/pull/3030#discussion_r958217305


hoodmane (approved):

hoodmane:
Maybe we should also add `benchmark-time.json` to the circleci artifacts.

PR #3047, opened 2022-08-31
Minor build optimizations for libraries

ryanking13:
Minor build optimizations for libraries by:

- Not building executables and test if possible
- Using `--disable-dependency-tracking` flags for libraries that use autoconf

### DIFF
--- a/packages/liblzma/meta.yaml
+++ b/packages/liblzma/meta.yaml
@@ -19,6 +19,7 @@ build:
         --disable-scripts \
         --disable-doc \
         --enable-shared=no \
+        --disable-dependency-tracking \
         --prefix=${WASM_LIBRARY_DIR}
 
     emmake make -j ${PYODIDE_JOBS:-3}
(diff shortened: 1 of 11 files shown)


hoodmane (approved):
Looks good to me.

packages/liblzma/meta.yaml
```diff
@@ -19,6 +19,7 @@ build:
         --disable-scripts \
         --disable-doc \
         --enable-shared=no \
+        --disable-dependency-tracking \
```
rth, line 22: I think it's disabled by default already? https://www.gnu.org/software/automake/manual/html_node/Dependency-Tracking.html

ryanking13: I think it is not disabled by default. I can observe the log in main:

```
checking dependency style of /root/repo/emsdk/emsdk/upstream/emscripten/emcc... gcc3
```

which is changed in this PR:

```
checking dependency style of /root/repo/emsdk/emsdk/upstream/emscripten/emcc... none
```

rth: OK, thanks for checking

PR #3048, opened 2022-08-31
Scipy: Use NPY_BLAS_LIBS and NPY_LAPACK_LIBS environment variables

hoodmane:
Use environment variables to tell numpy about BLAS and LAPACK. I also removed the `global` library loading, fingers crossed we might not need it anymore now.


rth:
The reason we were not doing it is that then CLAPACK would be an extra dependency of numpy, and so adding 1.2MB to download (+ increased load time). Currently,what we are doing is equivalent to statically linking a subset of netlib BLAS / LAPACK in numpy (since as far as I remember it has a lite BLAS version) which is not ideal, but in practice this way is more lightweight with comparable performance.

This might change though once we switch to a high-performant BLAS https://github.com/pyodide/pyodide/issues/227

rth (approved):
OK never mind my previous comment, you are doing it for scipy which is totally fine. 

> I also removed the global library loading, fingers crossed we might not need it anymore now.

Looks like this is still necessary, otherwise LGTM.

PR #3052, opened 2022-08-31
[Docs] Fix 'Loading packages' CDN url

prete:
### Description
There's a double shalsh in the CDN url on the [Loading packages](https://pyodide.org/en/stable/usage/loading-packages.html) page.

`PYODIDE_CDN_URL` already ends with a forward-slash, so when building the CDN URL a double slash is added (ie: `https://cdn.jsdelivr.net/pyodide/v0.21.2/full//pyodide.js`) and that is not a valid url. 

No slash is needed between the variable `PYODIDE_CDN_URL` and `pyodide.js`, like in the rest of the docs, so removing that extra shalsh after PYODIDE_CDN_URL creates the valid URL. 

### Checklists

- [ ] Add / update tests

### DIFF
--- a/docs/usage/loading-packages.md
+++ b/docs/usage/loading-packages.md
@@ -134,10 +134,7 @@ installs from PyPI. See [this stack overflow answer](https://stackoverflow.com/q
     <meta charset="utf-8" />
   </head>
   <body>
-    <script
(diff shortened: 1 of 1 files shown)


hoodmane:
Thanks!

PR #3055, opened 2022-09-01
Raise error when static libraries are specified as a runtime dependency

ryanking13:
Follow-up works for #2132.

### DIFF
--- a/packages/RobotRaconteur/meta.yaml
+++ b/packages/RobotRaconteur/meta.yaml
@@ -9,8 +9,9 @@ source:
 
 requirements:
   run:
-    - boost-cpp
     - numpy
+  host:
+    - boost-cpp
 
 build:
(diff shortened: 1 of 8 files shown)


rth (approved):
Nice! We should also update [the docs](https://pyodide.org/en/stable/development/meta-yaml.html#build-library) that `library` if True is then necessarily a static library.

Though I wonder if it wouldn't be better to have a single flag `library`: which can be `static`, `shared`, or False. Rather than two binary flags `library` and `sharedlibrary`. It's weird to have library=False for something that's a shared library.

hoodmane:
> I wonder if it wouldn't be better to have a single flag library: which can be static, shared, or False.

+1 for this that's a great idea

rth:
Opened https://github.com/pyodide/pyodide/issues/3066 as a follow-up improvement.

Otherwise merging as is, thanks!

PR #3059, opened 2022-09-02
Fix and xfail some flaky tests

ryanking13:
- `test_imp:test_unencodeable`: fails in firefox periodically.
- `astropy tests`: started to timeout periodically recently, not sure why. I guess it is because astropy tests install pytest plugins dynamically during tests.


rth:
Thanks!

Issue #3033, opened 2022-08-29
Relation between pyodide.loadPackage and micropip is not clear


viblo:
## 📚 Documentation

It is not clear in the documentation what is javascript and what is python. This becomes very confusing when reading about how to import a package at https://pyodide.org/en/stable/usage/loading-packages.html 

The page starts with 

> [pyodide.loadPackage](https://pyodide.org/en/stable/usage/api/js-api.html#pyodide.loadPackage) for packages built with Pyodide, or
> [micropip.install](https://pyodide.org/en/stable/usage/api/micropip-api.html#micropip.install) for pure Python packages with wheels available on PyPI or from other URLs.

Then below its a note:
> [micropip](https://pyodide.org/en/stable/usage/api/micropip-api.html#module-micropip) can also be used to load packages built in Pyodide (in which case it relies on [pyodide.loadPackage](https://pyodide.org/en/stable/usage/api/js-api.html#pyodide.loadPackage)).

This becomes very confusing. The pyodide referenced here is a javascript package, while micropip is used from within python. If you are already in python repl you should use micropip to install both pure packages and non-pure wheels (as long as they target the right env). It gets extra confusing as its possible to `import pyodide` from the repl, but the imported module does not contain loadPackage.

hoodmane:
Yep. For the record, you can import `loadPackage` via `from pyodide_js import loadPackage`. From JavaScript you can access the Python package as `pyodide.pyodide_py`. We've never figured out how to describe this situation particularly clearly in the docs.

viblo:
Thanks 👍

PR #3064, opened 2022-09-04
DOC Improve documentation on loading packages

rth:
Closes https://github.com/pyodide/pyodide/issues/3033

### DIFF
--- a/docs/usage/loading-packages.md
+++ b/docs/usage/loading-packages.md
@@ -45,7 +45,7 @@ appropriate,
 
 - to load micropip itself,
 - when you are optimizing for size, do not want to install the `micropip`
-  package, and do not need to install packages from PyPI.
+  package, and do not need to install packages from PyPI with dependency resolution.
 
 ## Micropip
 
@@ -57,10 +57,8 @@ Pyodide supports installing following types of packages with {mod}`micropip`,
 - pure Python and binary wasm32/emscripten wheels (also informally known as
   "Pyodide packages" or "packages built by Pyodide") from the JsDelivr CDN and
   custom URLs.
-  {func}`micropip.install` returns a Python
-  [Future](https://docs.python.org/3/library/asyncio-future.html) so you can
-  await the future or otherwise use the Python future API to do work once the
-  packages have finished loading:
+  {func}`micropip.install` is an async Python function which returns a
+  coroutine, so it need to be called with an `await` clause to run.
 
 ```pyodide
 await pyodide.loadPackage("micropip");
--- a/docs/usage/loading-packages.md
+++ b/docs/usage/loading-packages.md
@@ -156,8 +156,8 @@ await pyodide.loadPackage(
 The file name must be a valid wheel name.
 
 When you request a package from the official repository, all of the package's
-dependencies are also loaded. Dependency resolution is not yet implemented when
-loading packages from custom URLs.
+dependencies are also loaded. There is no dependency resolution when
+loading packages from custom URLs. If you want dependency resolution for custom URLs, use `micropip`.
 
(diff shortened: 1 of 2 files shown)


rth:
Regarding the issue you opened, let me know if there are still things unclear with this update. The rendered version is [here](https://pyodide--3064.org.readthedocs.build/en/3064/usage/loading-packages.html)

docs/usage/loading-packages.md
```diff
@@ -5,81 +5,62 @@
...
-  await pyodide.loadPackage("matplotlib");
-  // matplotlib is now available
-}
-main();
-```
-
-(micropip)=
+- to load micropip itself,
+- when you are optimizing for size, do not want to install the `micropip`
+  package, and do not need to install packages from PyPI.
```
hoodmane, line 48: This may not need to be said here, but you can use `loadPackage` to load wheels directly from PyPI as long as you also load their dependencies. Particularly when there are deep dependency trees this can be a lot faster because it requires 2n sequential requests to load a tree of depth n and particularly if early packages are big, downloading all of the wheels sequentially could take a while.

docs/usage/loading-packages.md
```diff
@@ -5,81 +5,62 @@
...
-the future or otherwise use the Python future API to do work once the packages
-have finished loading:
+- pure Python wheels from PyPI with {mod}`micropip`.
+- pure Python and binary wasm32/emscripten wheels (also informally known as
+  "Pyodide packages" or "packages built by Pyodide") from the JsDelivr CDN and
+  custom URLs.
+  {func}`micropip.install` returns a Python
+  [Future](https://docs.python.org/3/library/asyncio-future.html) so you can
+  await the future or otherwise use the Python future API to do work once the
+  packages have finished loading:
```
hoodmane, line 63: This is out of date, currently `micropip.install` is a normal `async` method so it returns a coroutine. So if you **don't** either await it or schedule it with `asyncio.ensure_future` it won't happen and you'll probably get a message like ` RuntimeWarning: coroutine 'f' was never awaited`. This was a breaking change in v0.20.0 but it seems that most people were `await`ing the result so nobody has complained.

docs/usage/loading-packages.md
```diff
@@ -153,6 +136,55 @@ installs from PyPI. See [this stack overflow answer](https://stackoverflow.com/q
...
+await pyodide.loadPackage(
+  "https://foo/bar/numpy-1.22.3-cp310-cp310-emscripten_3_1_13_wasm32.whl"
+);
+```
+
+The file name must be a valid wheel name.
+
+When you request a package from the official repository, all of the package's
+dependencies are also loaded. Dependency resolution is not yet implemented when
+loading packages from custom URLs.
```
hoodmane, line 160:
```suggestion
dependencies are also loaded. There is no dependency resolution when
loading packages from custom URLs. If you want dependency resolution for custom URLs, use `micropip`.
```

hoodmane: Since IMO we are never going to implement this.

docs/usage/loading-packages.md
rth: Yes, good point. But similarly, you could pass this list of URLs to micropip with `deps=False` and it would be faster. And it also makes for a simpler overall message. I adjusted this sentence a bit.

rth: I leave this out for now, because this is basically the use case of loading freezed packages, and we need to improve the explanations of how to do that anyway.

docs/usage/loading-packages.md
rth: OK thanks, I'll just change to say it's a async Python that returns a coroutine. People shouldn't be using ensure_future for this.

PR #3067, opened 2022-09-04
Add codecov configuration

rth:
Coverage and codecov is a bit brittle and can make the CI fail if the computed coverage fluctuates unrelated to code changes (e.g. in https://github.com/pyodide/pyodide/pull/3053). 

This adds a minimal threshold on what decrease in coverage should be considered a CI failure (adapted from scikit-learn)

### DIFF
--- a/.codecov.yml
+++ b/.codecov.yml
@@ -1,0 +1,25 @@
+comment: false
+
+coverage:
+  status:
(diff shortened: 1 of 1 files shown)


ryanking13 (approved):
Thanks!

Issue #2432, opened 2022-05-02
Ability to use Pyodide from a service worker


simonw:
## 🚀 Feature

I'd like to be able to use Pyodide in a service worker, not just in a web worker.

### Motivation

I want to be able to run Python server-side applications (including my [Datasette](https://datasette.io/) application) entirely within the browser.

To do this, I want to register a service worker which intercepts calls to e.g. `/fixtures/facetable` ([this page](https://latest.datasette.io/fixtures/facetable)) and runs my original server-side code using Pyodide.

I've been trying to get this working but then I ran into this error:

![image](https://user-images.githubusercontent.com/9599/166171678-2909fee5-e79d-4a11-b1ba-819059774639.png)

On further research, it looks like `XMLHttpRequest` is supported in web workers but not in service workers: https://stackoverflow.com/questions/37112425/xmlhttprequest-within-service-worker/37114241#37114241

So I think the fix here may be for Pyodide to either drop using `XMLHttpRequest` or to have a fallback for when it's not available. Though there may be plenty of other reasons it doesn't work in a service worker that I've not yet found yet!

### Pitch

I'd like to be able to run Python code in a service worker using Pyodide.

simonw:
It's possible that I don't need this after all - I'm new to service workers, and I'm now wondering if I can achieve what I want using a web worker instead and running additional code using the HTML5 history API whenever a page navigation event is attempted.

simonw:
... on further research I've confirmed that web workers are the way to go for this one. I'll leave this issue open just in case there are other use-cases for which service worker would be appropriate, but I have a strong hunch that this feature isn't actually useful and the issue should be closed.

hoodmane:
I think this is worth fixing. All we need to do is catch and throw away `ImportError` from the failure to load `XMLHttpRequest`. In this case, `open_url` will fail if you try to use it, but that is okay.

simonw:
Here's my write up of what I've built so far: https://simonwillison.net/2022/May/4/datasette-lite/

Aierie:
If someone wants to use pyodide from a background script in a Chrome Extension with manifest v3, they would encounter this problem (+others, since it is a service worker with additional restrictions).

rth:
> So I think the fix here may be for Pyodide to either drop using XMLHttpRequest or to have a fallback for when it's not available. 

[...]

PR #3070, opened 2022-09-05
Allow use with (module-type) service workers

Aierie:
### Description
Should solve https://github.com/pyodide/pyodide/issues/2432

Allows users to do a static import of `pyodide.asm.js` to set `globalThis._createPyodideModule`. With this change, users should be able to use Pyodide in module-type service workers by manually importing `pyodide.asm.js` + polyfilling `XMLHttpRequest`, without breaking existing dynamic import behaviour.

Considered removing the dynamic import in favour of a static import, but this might change behaviour of Pyodide in environments where only loading the big (1.6MB) `pyodide.asm.js` file when necessary is important.

I will update documentation with a new section guiding folks on how to use Pyodide with service workers, and I have drafted some content (Intercept a fetch call for some JSON and modify the data). There are 2 types of workers, with different requirements:

- classic - just polyfill `XMLHttpRequest`, setup is otherwise similar to a web worker
- module - import `pyodide.asm.js` + polyfill `XMLHttpRequest`

I'll write up material for both with explanations for the fixes.

### Checklists
- [x] Add / update tests

### DIFF
--- a/docs/project/changelog.md
+++ b/docs/project/changelog.md
@@ -13,6 +13,7 @@ substitutions:
 # Change Log
 
 ## Unreleased
+- {{ Enhancement }} Users can do a static import of `pyodide/pyodide.asm.js` to avoid issues with dynamic imports. This allows the use of Pyodide with module-type service workers. 
 
 - {{ Enhancement }} Emscripten was updated to Version 3.1.21
   {pr}`2958`, {pr}`2950`, {pr}`3027`, {pr}`3107`
--- a/docs/project/changelog.md
+++ b/docs/project/changelog.md
@@ -14,7 +14,10 @@ substitutions:
 
 ## Unreleased
 
-- {{ Enhancement }} Users can do a static import of `pyodide/pyodide.asm.js` to avoid issues with dynamic imports. This allows the use of Pyodide with module-type service workers.
+- {{ Enhancement }} Users can do a static import of `pyodide/pyodide.asm.js` to
+  avoid issues with dynamic imports. This allows the use of Pyodide with
+  module-type service workers. 
+  {pr}`3070`
 
 - {{ Enhancement }} Emscripten was updated to Version 3.1.21
   {pr}`2958`, {pr}`2950`, {pr}`3027`, {pr}`3107`
--- a/docs/project/changelog.md
+++ b/docs/project/changelog.md
@@ -13,7 +13,8 @@ substitutions:
 # Change Log
 
 ## Unreleased
-- {{ Enhancement }} Users can do a static import of `pyodide/pyodide.asm.js` to avoid issues with dynamic imports. This allows the use of Pyodide with module-type service workers. 
+
+- {{ Enhancement }} Users can do a static import of `pyodide/pyodide.asm.js` to avoid issues with dynamic imports. This allows the use of Pyodide with module-type service workers.
 
 - {{ Enhancement }} Emscripten was updated to Version 3.1.21
   {pr}`2958`, {pr}`2950`, {pr}`3027`, {pr}`3107`
--- a/docs/project/changelog.md
+++ b/docs/project/changelog.md
@@ -16,7 +16,7 @@ substitutions:
 
 - {{ Enhancement }} Users can do a static import of `pyodide/pyodide.asm.js` to
   avoid issues with dynamic imports. This allows the use of Pyodide with
-  module-type service workers. 
+  module-type service workers.
   {pr}`3070`
 
 - {{ Enhancement }} Emscripten was updated to Version 3.1.21
--- a/docs/usage/index.md
+++ b/docs/usage/index.md
@@ -73,9 +73,17 @@ officially supported.
 ## Web Workers
 
 By default, WebAssembly runs in the main browser thread, and it can make UI
-non-responsive for long-running computations.
+non-responsive for long-running computations. 
 
-To avoid this situation, one solution is to run {ref}`Pyodide in a WebWorker <using_from_webworker>`.
+To avoid this situation, one solution is to run {ref}`Pyodide in a WebWorker <using_from_webworker>`. 
+
+## Service Workers
+
+It's also possible to run {ref}`Pyodide in a Service Worker <using_from_service_worker>`. Service workers have some unique capabilities, one of them being running code in response to `fetch` calls on a webpage. 
+
+If you're not sure whether you need web workers or service workers, here's an [overview + comparison of the two][workers overview]. 
+
+[workers overview]: https://web.dev/workers-overview/
 
 ## Node.js
 
@@ -167,4 +175,5 @@ Python says that 1+1= 2
    webworker.md
    loading-custom-python-code.md
    file-system.md
+   service-worker.md
 ```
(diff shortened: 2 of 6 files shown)


docs/project/changelog.md
```diff
@@ -14,6 +14,8 @@ substitutions:
 
 ## Unreleased
 
+- {{ Enhancement }} Users can do a static import of `pyodide/pyodide.asm.js` to avoid issues with dynamic imports. This allows the use of Pyodide with module-type service workers.
```
hoodmane, line 17:
```suggestion
- {{ Enhancement }} Users can do a static import of `pyodide/pyodide.asm.js` to
  avoid issues with dynamic imports. This allows the use of Pyodide with
  module-type service workers. 
  {pr}`3070`
```

src/js/pyodide.ts
```diff
@@ -299,8 +299,14 @@ export async function loadPyodide(
   // locateFile tells Emscripten where to find the data files that initialize
   // the file system.
   Module.locateFile = (path: string) => config.indexURL + path;
-  const scriptSrc = `${config.indexURL}pyodide.asm.js`;
-  await loadScript(scriptSrc);
+
+  // If the pyodide.asm.js script has been imported, we can skip the dynamic import
+  // Users can then do a static import of the script in environments where
+  // dynamic importing is not allowed or not desirable, like module-type service workers
+  if (!("_createPyodideModule" in globalThis)) {
```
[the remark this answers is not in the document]
Aierie: Thanks!

hoodmane:
Generally looks reasonable to me. Could you add a test?

Aierie:
> Generally looks reasonable to me. Could you add a test?

Will add a test for manual import. Ideally I would like to add one for service workers as well, would I have to contribute to https://github.com/pyodide/pytest-pyodide for that, and should I do that first?

hoodmane:
Maybe start with a test that's not in a worker, we can figure out how to do the service worker later.

docs/usage/index.md
```diff
@@ -77,6 +77,14 @@ non-responsive for long-running computations.
...
 To avoid this situation, one solution is to run {ref}`Pyodide in a WebWorker <using_from_webworker>`.
 
+## Service Workers
+
+It's also possible to run {ref}`Pyodide in a Service Worker <using_from_service_worker>`. Service workers have some unique capabilities, one of them being running code in response to `fetch` calls on a webpage.
+
+If you're not sure whether you need web workers or service workers, here's an [overview + comparison of the two][workers overview].
+
+[workers overview]: https://web.dev/workers-overview/
+
```
hoodmane, line 87: I think we should merge this with the webworkers section:
```suggestion
It's also possible to run {ref}`Pyodide in a Service Worker <using_from_service_worker>`. 

If you're not sure whether you need web workers or service workers, here's an [overview and comparison of the two](https://web.dev/workers-overview/).
```

docs/usage/service-worker.md
```diff
@@ -0,0 +1,417 @@
+(using_from_service_worker)=
+
+# Using Pyodide in a service worker
+
+This document describes how to use Pyodide to execute Python scripts in a service worker. Compared to typical web workers, service workers are more related to acting as a network proxy, handling background tasks, and things like caching and offline. See [this article](https://web.dev/workers-overview/#use-cases) for more info.
```
hoodmane, line 5: Can you hard wrap this to 80ish characters? Some lines can be a bit longer but we don't want 100+ character long lines.

docs/usage/service-worker.md
```diff
@@ -0,0 +1,417 @@
...
+// With this, we won't need to reload the page before the service worker can intercept fetch requests
+// https://developer.mozilla.org/en-US/docs/Web/API/Clients/claim#examples
+self.addEventListener("activate", function (event) {
+  event.waitUntil(self.clients.claim());
+});
+```
+
+### XMLHttpRequest polyfill
+
+This script should be imported into your service worker, to assign `XMLHttpRequest` in the service worker's global scope.
```
hoodmane, line 193: Let's not put the polyfill in the docs. We can give people a url or something.

src/tests/test_pyodide.py
```diff
@@ -1425,3 +1425,26 @@ def test_csp(selenium_standalone_noload):
...
+        shutil.copy(
+            get_pyodide_root() / "src/templates/module_static_import_test.html",
+            target_path,
+        )
+        selenium.goto(f"{selenium.base_url}/module_static_import_test.html")
+        selenium.javascript_setup()
+        selenium.load_pyodide()
+        selenium.run_js(
+            """
+            pyodide.runPython(`
```
hoodmane, line 1444: This test doesn't fail failed prior to the change to `pyodide.ts`. We need to make sure that `${indexURL}/pyodide.asm.js` doesn't exist. Use `tmpdir`, copy the `dist` directory to the temporary directory but remove `pyodide.asm.js` (or move it to a subfolder) and then the static import can be from the correct url. This should fail prior to the change because they dynamic import is from the wrong path.

docs/usage/service-worker.md
Aierie: I feel like people might have trouble finding a good polyfill that works out of the box from npm. Most haven't been updated in years and the one I've vendored here has a bug where it assigns the wrong variable and breaks when used outside of Node.js. I'll try looking for one though.

docs/usage/service-worker.md
Aierie: done!

docs/usage/service-worker.md
hoodmane: In that case, I am okay with including this as a separate file `.js` in this folder and linking to it.

hoodmane:
Thanks for your work on this @Aierie! We appreciate it.

docs/usage/service-worker.md
Aierie: The author of xhr-shim responded quickly to a PR and now we can just recommend that package instead!

rth:
Very nice work @Aierie thanks for doing this! Generally sounds good to me but I haven't really worked with service workers, so I don't know much about it.

A question I have is how do we make sure this tutorial in the documentation doesn't get outdated and that we don't break something unidvertantly? Is `test_static_import` the only thing we could reasonably add to the test suite? Can't we actually have a service worker setup  in CI or would that be too complicated to maintain?

Aierie:
> Can't we actually have a service worker setup in CI or would that be too complicated to maintain?
I think it's possible to test service worker functionality - some advice [here](https://jeffy.info/2022/08/25/testing-a-service-worker.html) seems practical and I can try to introduce a simple test for say, network call interception. I'm unsure whether this should be included in this PR though.

> A question I have is how do we make sure this tutorial in the documentation doesn't get outdated and that we don't break something unidvertantly? 
re: breakage/getting outdated, I think if we can get service worker tests for both module-type and classic-type workers, that should cover most of the demo since the critical bits are around setup, another thing that I could do is to reduce the surface area of Pyodide's API that the current example relies on.

> Is test_static_import the only thing we could reasonably add to the test suite? 
I think we can add a node test that also covers the static import, I'll look into this.

hoodmane:
I'm +1 to merge this as is and make improvements in follow-up prs if that's okay with you @Aierie.

Aierie:
> I'm +1 to merge this as is and make improvements in follow-up prs if that's okay with you @Aierie.

Sure, that's fine by me!

rth:
Thanks you @Aierie and @hoodmane !

TomerBin:
Hey @Aierie ! Thanks for all of your great work here.
Does this feature means it's now possible to import pyodide from a v3 extension?

PR #3074, opened 2022-09-06
Add version variable for Pyodide JavaScript Package

### LINKED ISSUE #3068: Export version from pyodide JS module
## 🚀 Feature

In `pyodide.d.ts` I see `declare let version: string;` but it's not exported. It'd be great if it was exported so that I can do this:

```ts
import {version} from "pyodide";
```

### Motivation

I have [some code](https://github.com/alexmojaki/pyodide-worker-runner/blob/e7dd3d0ee1dff457bf9d6104944477840a83e5a7/lib/index.ts#L16) that roughly looks like this:

```ts
import {loadPyodide, PyodideInterface} from "pyodide";

const version = "0.21.1";
const indexURL = `https://cdn.jsdelivr.net/pyodide/v${version}/full/`;
const pyodide: PyodideInterface = await loadPyodide({indexURL});
if (pyodide.version !== version) {
  throw new Error(
    `loadPyodide loaded version ${pyodide.version} instead of ${version}`,
  );
}
```

I'd like to import `version` instead of setting it manually, so that it always automatically matches whatever version of pyodide is installed.

ryanking13:
### Description

Close #3068 

Adds a static version variable for Pyodide JavaScript package.

One can use it via

```js
import { version } from 'pyodide';
```

Previously, it was only accessible after calling `loadPyodide`:

```js
pyodide.version
```

I also added a check that compares the version of pyodide JS package and Python package when loading Pyodide. If they differ, it will throw an error. This will help preventing some common errors of setting different version of indexURL.

Note that this check happens after unpacking pyodide Python package, so the error that happens before initializing Pyodide is not caught. For example, the error: `TypeError: Module.API.capture_stderr is not a function.` reported by @sspickle (in Gitter) will still happen, as the newly added `API.capture_stderr()` API [is used in the initialization step](https://github.com/pyodide/pyodide/blob/555f7828f4339b056f2d32b821e6a752784f4729/src/js/pyodide.ts#L89).

cc: @alexmojaki 

### Checklists

- [x] Add / update tests


hoodmane:
Thanks @ryanking13!
If we add:
```
char[] version = "0.22.0";
```
to `main.c` or `API.version = "0.22.0"` to `pre.js` then we could perform a version check before doing anything else and prevent errors like `TypeError: Module.API.capture_stderr is not a function.`

ryanking13:
> ```
> char[] version = "0.22.0";
> ```
> 
> to `main.c` or `API.version = "0.22.0"` to `pre.js` then we could perform a version check before doing anything else and prevent errors like `TypeError: Module.API.capture_stderr is not a function.`

That sounds good to me.

src/js/pyodide.ts
```diff
@@ -144,11 +147,9 @@ function finalizeBootstrap(API: any, config: ConfigType) {
   API.pyodide_code = import_module("pyodide.code");
   API.pyodide_ffi = import_module("pyodide.ffi");
   API.package_loader = import_module("pyodide._package_loader");
-  API.version = API.pyodide_py.__version__;
 
   // copy some last constants onto public API.
   pyodide.pyodide_py = API.pyodide_py;
-  pyodide.version = API.version;
```
hoodmane, line 151: We still need to assign `pyodide.version` or else this is an API break.
```
pyodide.version = version
```

src/js/pyodide.ts
```diff
@@ -324,6 +325,15 @@ export async function loadPyodide(
     throw Module.exited.toThrow;
   }
 
+  if (Module.version !== version) {
```
hoodmane, line 328: As above:
```suggestion
  if (API.version !== version) {
```

src/js/pyodide.ts
hoodmane: Oh I guess this happens in `api.ts`.

src/js/pyodide.ts
hoodmane: and in a few other places...

hoodmane (approved):

src/core/pre.js
```diff
@@ -3,3 +3,5 @@ const Hiwire = {};
 const Tests = {};
 API.tests = Tests;
 Module.hiwire = Hiwire;
+
+Module.version = "0.22.0.dev0";
```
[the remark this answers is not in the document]
ryanking13: Sounds good to me

ryanking13:
Thank you for the reviews!

hoodmane:
Thanks @ryanking13!

PR #3075, opened 2022-09-07
Fix creation of PyProxy when CSP without unsafe-eval is used

hoodmane:
See discussion following https://github.com/pyodide/pyodide/issues/2432#issuecomment-1235646608.

* [x] Add test
* [x] Update changelog

### DIFF
--- a/docs/project/changelog.md
+++ b/docs/project/changelog.md
@@ -44,6 +44,11 @@ substitutions:
   `indexURL` (this was a regression in v0.21.2).
   {pr}`3077`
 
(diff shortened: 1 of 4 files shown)


rth (approved):
LGTM, once there is a test. Thanks!

hoodmane:
Great, now that we have Safari tests the test suite can remind us about all the features that are missing from Safari.

hoodmane:
@ryanking13 how can I tell which version of Safari we are testing against?

PR #3077, opened 2022-09-07
Fix browser_resolvePath when no base is provided

### LINKED ISSUE #3076: Calling `loadPyodide()` with `indexURL` results in error
## 🐛 Bug

When `loadPyodide()` is called with a value for `indexURL` that is not a complete URL (with a protocol), it results in an error.

I believe this was caused by #3013.

## To Reproduce

```js
await loadPyodide({ indexURL: "./" });
```

This produces this error:

```
pyodide.js:1 Uncaught (in promise) TypeError: Failed to construct 'URL': Invalid URL
    at resolvePath (pyodide.js:1:7795)
    at loadPyodide (pyodide.js:1:12107)
    at async main (console.html:76:30)
```

In my particular use case, I'm not using  `"./"`, but a relative path, like `"./foo"`.

It _does_ work if you use:

```
await loadPyodide({ indexURL: "https://cdn.jsdelivr.net/pyodide/dev/full/" });
```

### Expected behavior

The docs for `indexURL` say:

> options.indexURL (string()) – The URL from which Pyodide will load the main Pyodide runtime and packages. Defaults to the url that pyodide is loaded from with the file name (pyodide.js or pyodide.mjs) removed. It is recommended that you leave this undefined, providing an incorrect value can cause broken behavior.

Based on this documentation, it's not entirely clear to me whether paths like `"./foo"` or `"/foo"` _should_ work, but I do know that they used to work.

### Environment

- Pyodide Version: 0.21.2
- Browser version: 104.0.5112.101

hoodmane:
Fixes #3076. This needs test coverage.

### Checklists

- [x] Add / update tests

### DIFF
--- a/docs/project/changelog.md
+++ b/docs/project/changelog.md
@@ -38,6 +38,10 @@ substitutions:
   be passed as command line arguments to the Python interpreter at start up.
   {pr}`3021`
 
+- {{ Fix }} It works again to use `loadPyodide` with a relative URL as
+  `indexURL` (this was a regression in v0.21.2).
+  {pr}`3077`
+
 ### Build System / Package Loading
 
 - New packages: pycryptodomex {pr}`2966`, pycryptodome {pr}`2965`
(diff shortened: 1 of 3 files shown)


rth (approved):
LGTM, thanks.

> This needs test coverage.

Something like the following maybe?

```py

def test_loadPyodide_relative_url(selenium_standalone):
    """Check that loading Pyodide with a relative URL works"""
    selenium_standalone.run_js(
        """
        self.pyodide = await loadPyodide({ indexURL: "./" });
        """
    )

```

hoodmane:
Thanks @rth used your suggestion.

PR #3078, opened 2022-09-07
fix PATH and ERRNO_CODES failing puppeteer tests

Aierie:
These tests were failing when I ran `npm run test:browser`, because `globalThis.pyodide` was set in puppeteer's page context. If in node, `page.evaluate` will simply call the function which accesses the node `globalThis` that does have the pyodide property set.


rth:
Thanks! Not sure about the CircleCI error. Could you maybe enable CircleCI on your Pyodide fork and retrigger CI to see if it helps?

rth:
Thanks!

PR #3079, opened 2022-09-07
Switch to meta.yaml parsing with pydantic

rth:
This switches to meta.yaml parsing with [pydantic](https://pydantic-docs.helpmanual.io/).

The main advantage is that the code is simpler and more type aware. So mypy should be able to catch more issues earlier rather than at build time. The default values are now part of the Pydantic spec, rather than being scattered though the code.

Also one day we might be able to generate the docs from this spec rather than it evolving separately.

In breaking changes this replaces `parse_package_config` with `MetaConfig.from_yaml`. Not that we plan of changing the format, but it makes it a bit more obvious that the main config specification is done in this pydantic class, and meta.yaml is just a serialization of it.  It also makes it easier to have `MetaConfig.to_yaml` if need be.

- [x] Re-implement all the extra validations of the config we had (and write associated tests)
- [x] Add / update tests


docs/sphinx_pyodide/sphinx_pyodide/packages.py
```diff
@@ -51,12 +51,12 @@ def run(self):
         def parse_package_info(
             self, config: pathlib.Path
         ) -> tuple[str, str, bool, str]:
-            yaml_data = parse_package_config(config)
+            yaml_data = MetaConfig.from_yaml(config)
 
-            name = yaml_data["package"]["name"]
-            version = yaml_data["package"]["version"]
-            tag = yaml_data["package"].get("_tag", "")
```
rth, line 58: For some reason, pydantic errors on `_` prefixed fields, so aliased them to the names without prefixes. The meta.yaml is unchanged.

docs/sphinx_pyodide/sphinx_pyodide/packages.py
```diff
@@ -51,12 +51,12 @@ def run(self):
...
-            is_library = yaml_data.get("build", {}).get("library", False)
```
rth, line 59: During the review, one needs to be careful about these default values being consistent with the default values I put into the Pydantic spec.

pyodide-build/pyodide_build/buildpkg.py
```diff
@@ -191,26 +188,19 @@ def check_checksum(archive: Path, source_metadata: dict[str, Any]) -> None:
     source_metadata
         The source section from meta.yaml.
     """
-    checksum_keys = {"md5", "sha256"}.intersection(source_metadata)
```
rth, line 194: I dropped support for md5, since as far as I can tell we are no longer using it.

pyodide-build/pyodide_build/buildpkg.py
```diff
@@ -251,7 +241,9 @@ def download_and_extract(
     src_metadata
         The source section from meta.yaml.
     """
-    response = request.urlopen(src_metadata["url"])
+    # We only call this function when the URL is defined
+    url = cast(str, src_metadata.url)
```
rth, line 245: Mypy is not able to determine that this is only called with a non None URL and needs some help (this only affects types).

pyodide-build/pyodide_build/buildpkg.py
```diff
@@ -317,23 +309,23 @@ def prepare_source(
         shutil.rmtree(buildpath)
     os.makedirs(buildpath)
 
-    if "url" in src_metadata:
```
rth, line 320: Now `src_metadata.url` etc always exists, however it is None if not set

pyodide-build/pyodide_build/buildpkg.py
```diff
@@ -317,23 +309,23 @@ def prepare_source(
...
         download_and_extract(buildpath, srcpath, src_metadata)
         return
-    if "path" not in src_metadata:
+    if src_metadata.path is None:
         raise ValueError(
             "Incorrect source provided. Either a url or a path must be provided."
         )
 
-    srcdir = Path(src_metadata["path"]).resolve()
+    srcdir = src_metadata.path.resolve()
```
rth, line 320: It's already a Path in the spec

pyodide-build/pyodide_build/io.py
```diff
@@ -1,255 +1,224 @@
...
+
+    class Config:
+        extra = pydantic.Extra.forbid
+
+
+class _AboutSpec(BaseModel):
+    home: str | None = None
+    PyPI: str | None = None
+    summary: str | None = None
+    license: str | None = None
```
rth, line 145: There was previously some ambiguity here between an empty string and it being not set.

pyodide-build/pyodide_build/io.py
```diff
@@ -1,255 +1,224 @@
...
+    cxxflags: str = ""
+    ldflags: str = ""
+    library: bool = False
+    sharedlibrary: bool = False
+    cross_script: str | None = Field(None, alias="cross-script")
+    script: str | None = None
+    post: str | None = None
+    unvendor_tests: bool = Field(True, alias="unvendor-tests")
+    cross_build_env: bool = Field(False, alias="cross-build-env")
+    cross_build_files: list[str] = Field([], alias="cross-build-files")
```
rth, line 92: There is a bit of a question about whether we keep these aliases or rename the meta.yaml fields to use `_`. If we rename them it should probably be a separate PR anyway.

pyodide-build/pyodide_build/io.py
```diff
@@ -1,255 +1,224 @@
...
+            )
+
+        return values
+
+
+_BuildSpecExports = Literal["pyinit", "requested", "whole_archive"]
+
+
+class _BuildSpec(BaseModel):
+    exports: _BuildSpecExports | list[_BuildSpecExports] = "pyinit"
```
rth, line 80: Wouldn't it be easier to make it a list only  and disallow strings?

rth:
This should be ready to review.

pyodide-build/pyodide_build/pywasmcross.py
```diff
@@ -98,6 +98,10 @@ def make_command_wrapper_symlinks(
         env[var] = symlink
 
 
+# Also defined in pyodide_build.io (though aviding to import dependent modules here)
```
ryanking13, line 101:
```suggestion
# Also defined in pyodide_build.io (though avoiding to import dependent modules here)
```

pyodide-build/pyodide_build/buildpkg.py
ryanking13: I think adding something like `if not src_metadata.url: ...` before this line makes mypy happy.

pyodide-build/pyodide_build/io.py
ryanking13: I have no strong opinion. It seems like YAML has no strong convention for this.

ryanking13 (approved):
Thanks, @rth! It looks much nicer to me.

pyodide-build/pyodide_build/buildpkg.py
rth: Indeed, but then you need to raise an exception when it is None. Technically that exception is not tested and decreases code coverage (unless one writes a test for it)

This has the same end result but is shorter and without untested lines of code.

PR #3085, opened 2022-09-08
FIX for webpack: Export binary files in javascript package

dlech:
### Description

This adds the distributed binary files to the "exports" section of the package.json file. This allows the files to be referenced, e.g. by `require.resolve('pyodide/distutils.tar')` which is useful for tools like webpack.

### Checklists


dlech:
Another user who would benefit from this change: https://github.com/mneil/pyodide-webpack/blob/main/patch.js

hoodmane:
Well we can do this but we probably won't call it a breaking change if we rename, move, or delete these files.

hoodmane:
Oh I see they are here:
https://github.com/mneil/pyodide-webpack/blob/main/webpack.config.js#L39-L44
We should probably have our own webpack plugin that does this so that downstream doesn't need to know the exact list of files we use.

dlech:
> We should probably have our own webpack plugin

That would be great.

hoodmane:
> webpack plugin

I would appreciate it if someone else could help with this though. I've tried to write a webpack plugin several times but always run into the problem that I was unable to use webpack with Pyodide at all. Perhaps with @mneil's example I would have an easier time, but the webpack plugin API is also very confusing. If it's just a wrapper around `CopyPlugin` then that probably isn't too hard. 

In @mneil's example, he also patches `pyodide.js` to remove the global assignment to `loadPyodide`, but if you use `pyodide.mjs` instead of `pyodide.js` it doesn't assign `loadPyodide` to the global scope. We make `pyodide.js` out of `pyodide.umd.ts` which just adds the `globalThis` assignment to `pyodide.ts`:
https://github.com/pyodide/pyodide/blob/main/src/js/pyodide.umd.ts
Whereas `pyodide.mjs` is made from `pyodide.ts` directly. Though deleting that line is also an option if using webpack to load a module is not possible.

hoodmane:
Okay after some discussion @mneil and @dlech have convinced me that this is a good idea. Thanks guys!

hoodmane:
Since this PR was opened, we apparently changed the name of `distutils.tar` to `distutils-1.0.0.zip`.

Issue #1955, opened 2021-11-13
Add more docstring examples


rth:
I think it would be helpful to have more example in docstring on how to use various functions and methods exposed in Pyodide. The general approach could be to,
 - have a short description documenting the function/method/class and its parameters/return value
 - include an example, ideally that is run as a doctest (we need to figure out how to do this) otherwise those will get outdated in time
 - link to the user manual for more details, when it makes sense

This is related to #1498 

### Python API

(looking at the API documentation)

- [ ] pyodide.CodeRunner
- [ ] pyodide.JsProxy 
- [ ] console.Console
- [ ] console.ConsoleFuture
- [ ] console.repr_shorten
- [ ] pyodide.create_once_callable
- [ ] pyodide.create_proxy
- [ ] pyodide.destroy_proxies
- [ ] pyodide.eval_code
- [ ] pyodide.eval_code_async
- [x] pyodide.find_imports
- [ ] pyodide.open_url
- [ ] pyodide.register_js_module
- [x] pyodide.should_quiet
- [ ] pyodide.to_js
- [ ] pyodide.unregister_js_module
- [ ] class webloop.WebLoop

### Micropip API

- [ ] micropip.install

### JS API

It's still unclear to me how dostring examples should look like in JS and whether they could be run as doctests. Will open a separate issue.

- [ ] pyodide.isPyProxy
- [ ] pyodide.loadPackage
- [ ] pyodide.loadPackagesFromImports: the is a verbal example, but I think it would be easier to read it as code.
- [ ] pyodide.registerComlink
- [ ] pyodide.registerJsModule
- [ ] pyodide.runPython
- [x] pyodide.runPythonAsync
- [ ] pyodide.setStandardStreams also misses argument description
- [ ] pyodide.toPy
- [ ]  pyodide.PyBuffer: there is an example but I find it quite complex
- [ ] PyProxy (and possibly its methods)

dataxerik:
Hi @rth 

I was wondering if it was cool if I worked on this?

hoodmane:
Please do! We very much appreciate it.

oow214:
I would like to contribute to this work. Can I try too?

hoodmane:
Yes. Everyone is welcome to work on this. Thanks!

Abhi149209:
@hoodmane is this issue fixed , if not please assign me . 
/assign

rth:
Sure @carrycooldude work on this would be very much appreciated.

[...]

PR #3090, opened 2022-09-09
DOCS add docstring examples for eval_code

dataxerik:
### Description

As a part of #1955, I've created this PR to add examples for eval_code. I feel it's easier to do this in parts. I tried to be exhaustive with the examples. 

### Checklists
NOTE: I'm not sure if this warrants a changelog entry or not 

### DIFF
--- a/src/py/_pyodide/_base.py
+++ b/src/py/_pyodide/_base.py
@@ -425,7 +425,7 @@ def eval_code(
 
     Examples
     --------
-    >>> from pyodide import eval_code
+    >>> from pyodide.code import eval_code
     >>> source = '''
     1 + 1
     '''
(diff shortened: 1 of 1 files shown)


src/py/_pyodide/_base.py
```diff
@@ -422,6 +422,30 @@ def eval_code(
...
+    >>> eval_code(source, globals=my_globals, locals=my_locals)
+    200 100
+    >>> source = "test = 1 + 1"
+    >>> eval_code(source, return_mode="last_expr_or_assign")
+    2
+    >>> eval_code(source, return_mode="last_expr")
+    >>> eval_code(source, return_mode="none")
+    >>> source = "print('Hello from pyodide')" # Pretend this is open('example_of_filename.py', 'r').read()
+    >>> eval_code(source, filename="example_of_filename.py")
+    Hello from pyodide
```
hoodmane, line 448: The `filename` argument doesn't actually change the result here in any way. Maybe it would be good to raise an error and look at the traceback?

hoodmane (commented):
Thanks!

src/py/_pyodide/_base.py
dataxerik: I was thinking the same thing about the result. I've update the example. Let me know if that's what you were thinking as well

hoodmane:
Looks good to me. Thanks @dataxerik!

PR #3094, opened 2022-09-11
Fix webpack warnings and reenable webpack test

### LINKED ISSUE #3087: webpack complains about dynamic imports
## 🐛 Bug

When using Pyodide with webpack I am seeing the following warnings:

```
WARNING in ./node_modules/pyodide/pyodide.mjs 231:56-67
Critical dependency: the request of a dependency is an expression

WARNING in ./node_modules/pyodide/pyodide.mjs 236:10-21
Critical dependency: the request of a dependency is an expression

WARNING in ./node_modules/pyodide/pyodide.mjs 243:98-140
Critical dependency: the request of a dependency is an expression
```

### To Reproduce

I haven't figured out a minimal way to reproduce this yet. 

### Expected behavior

The `pyodide` package should not trigger any webpack warnings.

### Environment

- Pyodide Version: 0.21.2
- Browser version: N/A

### Additional context

The relevant source code is:

https://github.com/pyodide/pyodide/blob/e2ce091b4c9953f06deca7a6c4f1b52ea716d53b/src/js/compat.ts#L177
https://github.com/pyodide/pyodide/blob/e2ce091b4c9953f06deca7a6c4f1b52ea716d53b/src/js/compat.ts#L188
https://github.com/pyodide/pyodide/blob/e2ce091b4c9953f06deca7a6c4f1b52ea716d53b/src/js/compat.ts#L216

The first of these did have `/* webpackIgnore: true */` but that was removed in #2468. I'm not sure if the comment would persist to the minified code that is distributed or not.

hoodmane:
This uses sed to insert `/* webpackIgnore: true */` comments into `pyodide.js`. This resolves #3087. I also enabled a check that a simple webpack config builds without warnings. In the future it would be good to add a test that it also runs without error.

### Checklists

- [x] Add / update tests

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -74,7 +74,7 @@ node_modules/.installed : src/js/package.json src/js/package-lock.json
 
 dist/pyodide.js src/js/_pyodide.out.js: src/js/*.ts src/js/pyproxy.gen.ts src/js/error_handling.gen.ts node_modules/.installed
 	npx rollup -c src/js/rollup.config.js
-	sed -i -E 's!await import\(([a-z0-9._/"-]*?)\)!await import(\1 /* webpack ignore */)!g' pyodide.js
+	sed -i 's!await import(!await import(/* webpackIgnore: true */ !g' dist/pyodide.js
 
 dist/package.json : src/js/package.json
 	cp $< $@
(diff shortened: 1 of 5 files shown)


Makefile
```diff
@@ -74,6 +74,7 @@ node_modules/.installed : src/js/package.json src/js/package-lock.json
 
 dist/pyodide.js src/js/_pyodide.out.js: src/js/*.ts src/js/pyproxy.gen.ts src/js/error_handling.gen.ts node_modules/.installed
 	npx rollup -c src/js/rollup.config.js
+	sed -i 's!await import(!await import(/* webpackIgnore: true */ !g' dist/pyodide.js
```
rth, line 77: Can you add a comment about what it does, as it's not entirely obvious from the regex, particularly given that '!' is used as a separator from what I understand?

rth (approved):
Thanks @hoodmane.

Issue #2179, opened 2022-02-15
Add package build cache


rth:
We are currently using ccache to cache the compilation of individual objects, which certainly makes the build faster.

However, I wonder if it wouldn't be better to also cache at the package level. Essentially, hash the source tarball, the meta.yaml, patches, emsdk version and env variables passed to the build script. Then take an existing wheel if such hash already exists.

For instance, even with ccache, building scipy still takes very long.

This would allow us to continue having monorepo builds, but waste less resources on rebuilding packages that don't need to be rebuilt.

hoodmane:
Sounds like a good idea to me, particularly for numpy and scipy which are the big bottlenecks. We could start by always rebuilding everything on main and stable but not on other branches?

Is it possible to build things that depend on scipy without rebuilding scipy first? Clearly static libraries are needed at link time, but maybe other things don't actually need to be ordered.

rth:
> We could start by always rebuilding everything on main and stable but not on other branches?

PRs are indeed the main issue.

> Is it possible to build things that depend on scipy without rebuilding scipy first?

That's where it gets a bit more complicated, for packages that are needed at build time such as numpy or scipy and have [`build->skip_host=False`](https://pyodide.org/en/stable/development/meta-yaml.html?highlight=skip_host#build-skip-host) we could try caching both the produced wasm wheel and the installed folder for amd64.
For scipy very few actually need it at build, but for instance scikit-learn does due to the use of Cython BLAS API from scipy.

ryanking13:
Sounds nice to me too. Building packages is really a bottleneck in CI.

> Essentially, hash the source tarball, the meta.yaml, patches, emsdk version and env variables passed to the build script.

What would happen if the static libraries linked to the package has been updated? For example, `zlib` is linked to many packages, and if we add a patch to zlib, we should detect and rebuild the linked packages, if I'm understanding correctly.

hoodmane:
Also, what's the deal with zlib as a library? We already have a copy of zlib in the main module, why do we have to statically link it again into the side modules?

rth:
> What would happen if the static libraries linked to the package has been updated?

Yes, the fact that a dependency was built (as opposed to loaded from cache) should likely trigger the build of all downstream dependencies. We already have [some related logic](https://github.com/pyodide/pyodide/blob/55fbd32ef319c2a2ed5ce297cf4b863a171ac8c4/pyodide-build/pyodide_build/buildpkg.py#L559) so that would probably need updating.

[...]

PR #3095, opened 2022-09-13
TST Run package tests in Safari

ryanking13:
Follow up works on #2578 


ryanking13:
I'm not sure why micropip is not dealing with tags correctly. I have no Mac so it is not easy to debug...

rth (approved):
Great, thanks! Good to see that tests pass in Safari  almost without changes. It adds ~30min to CI runtime, so we do need change the way packages are built to avoid having this extra compute for each commit.

ryanking13:
Yes, the CI time is becoming worse. Probably we should find a way to do more caching (#2179). Actually, the package build is [not cached at all for now](https://app.circleci.com/pipelines/github/pyodide/pyodide/4795/workflows/a7835f8e-0d73-4782-bce3-b70e3ad59241/jobs/56456) so we need to fix that first.

PR #3098, opened 2022-09-14
Add support for loading .so files in command line runner

hoodmane:
When they are larger than the synchronous compilation threshold. 

Notes:
1. This only works if the package works with `global_dso: false` in #3054. Hopefully we can gradually move more to explicitly linking shared libraries they depend on.
2. `pip install scipy` makes the command line runner **extremely** slow. Without scipy installed, `python -c 'print(1)'` runs in about 1 second, but with it installed it takes more like 10 seconds (time to load `clapack_all.so` and 111 different .so files in scipy, totaling 20 megabytes). We have to load all of this despite the fact that we won't use any of it.

### Checklists


src/py/pyodide/_package_loader.py
```diff
@@ -26,12 +26,18 @@
...
-    # Better not put stuff into /usr/lib or /lib/python3.10! For now let's stick
-    # everyone into SITE_PACKAGES in this case
-    TARGETS = {"site": SITE_PACKAGES, "lib": SITE_PACKAGES, "dynlib": SITE_PACKAGES}
+    # Better not put stuff into /usr/lib or /lib/python3.10! Stick the stdlib
+    # into SITE_PACKAGES and dsos up two directories.
+    #
+    # e.g., SITE_PACKAGES = .venv/lib/python3.10/site_packages
+    # and   DSO_DIR       = .venv/lib/
+    STD_LIB = SITE_PACKAGES
+    DSO_DIR = SITE_PACKAGES.parents[1]
```
ryanking13, line 38: Could you check if `import shapely` works in cmdline runner? I think is always try to load `geos` from `/usr/lib`.

ryanking13 (commented):
Thanks @hoodmane!

> pip install scipy makes the command line runner extremely slow. Without scipy installed, python -c 'print(1)' runs in about 1 second, but with it installed it takes more like 10 seconds (time to load clapack_all.so and 111 different .so files in scipy, totaling 20 megabytes). We have to load all of this despite the fact that we won't use any of it.

Well that sounds horrible. Can't we pass `loadAsync` flags to false when calling `loadDynamicLibrary`? Since in cmdline runner we already have files in the file system so maybe we can compile libraries synchronously on demand?

src/py/pyodide/_package_loader.py
hoodmane: I suspect it doesn't work.

hoodmane:
> Can't we pass `loadAsync` flags to false when calling `loadDynamicLibrary`?

It's worth checking, but I think not. I think v8 has caps on the maximum size of a wasm module that can be synchronously loaded and node inherits those caps.

hoodmane:
This PR is needed in order for the command line runner to work in numpy CI.

tools/python
```diff
@@ -185,6 +185,27 @@ async function main() {
         process.exit(code);
     };
 
+    py.runPython(
+        `
+        from pyodide._package_loader import SITE_PACKAGES, should_load_dynlib
+        from pyodide.ffi import to_js
+        dynlibs_to_load = to_js([
+            str(path) for path in SITE_PACKAGES.glob("**/*.so")
```
ryanking13, line 193: Could you use a same regex used in https://github.com/pyodide/pyodide/pull/3154, just in case?

ryanking13 (approved):
Thanks for working on this @hoodmane. Overall I am +1 with merging this. I guess #3054 is not mandatory for this (though making #3054 work will make loading libraries faster)?

hoodmane:
Well without #3054 or some other changes this won't work for any package that uses a shared library. But there is still a benefit because it makes the other packages work

ryanking13:
> Well without #3054 or some other changes this won't work for any package that uses a shared library.

Why won't it work? Is loading a library globally causing some error?

hoodmane:
No the global libraries are just missing. I don't want to always load all shared libraries because loading them is really slow. But we also don't know which ones are needed.

ryanking13:
> No the global libraries are just missing. I don't want to always load all shared libraries because loading them is really slow. But we also don't know which ones are needed.

Okay, that makes sense. How much does loading the library globally slower compared to loading it locally? I mean, even if we manage to make libraries work locally, we will still have to load all of them at the initialization step unless synchronously compiling a WASM module is available.

hoodmane:
> How much does loading the library globally slower compared to loading it locally?

I think they are pretty much the same speed. The issue is in knowing whether or not we have to load them at all: we are using `pip` to install things but `pip` has no idea whether or not `numpy` or `scipy` require some external non-wheel `.so` file.

ryanking13:
In #3167 I opened an issue about analyzing which shared libs are used in the wheel and embedding them into the wheel. That issue is still WIP, but finding which libraries are required for each wheel or .so file is possible with the `auditwheel-emscripten` package that I suggested in that issue. Do you think it would help?

For example, you can:

```sh
$ pip install auditwheel-emscripten
$ pyodide audit show Shapely-1.8.2-cp310-cp310-emscripten_3_1_21_wasm32.whl
The following external shared libraries are required:
{
│   'shapely/speedups/_speedups.cpython-310-wasm32-emscripten.so': ['libgeos_c.so'],
│   'shapely/vectorized/_vectorized.cpython-310-wasm32-emscripten.so': ['libgeos_c.so']
}
```

```py
>>> import auditwheel_emscripten
>>> auditwheel_emscripten.show("Shapely-1.8.2-cp310-cp310-emscripten_3_1_21_wasm32.whl")
{'shapely/speedups/_speedups.cpython-310-wasm32-emscripten.so': ['libgeos_c.so'], 'shapely/vectorized/_vectorized.cpython-310-wasm32-emscripten.so': ['libgeos_c.so']}
```

hoodmane:
> Do you think it would help?

Yes, this does sound like it would likely be a solution. Thank you @ryanking13!

tools/python
```diff
@@ -185,6 +185,29 @@ async function main() {
...
 
+    py.runPython(
+        `
+        from pyodide._package_loader import SITE_PACKAGES, should_load_dynlib
+        from pyodide.ffi import to_js
+        import re
+        SHAREDLIB_REGEX = re.compile(r"\.so(.\d+)*$")
+        dynlibs_to_load = to_js([
+            str(path) for path in SITE_PACKAGES.glob("**/*.so*")
+            if SHAREDLIB_REGEX.search(path.name) and should_load_dynlib(path)
```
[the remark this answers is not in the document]
hoodmane: I think it's better to use the `glob` to filter out things that don't contain `.so` first is good. But I see that the `SHAREDLIB_REGEX` is redundant.

src/js/load-package.ts
```diff
@@ -323,6 +323,115 @@ async function downloadAndInstall(
   }
 }
 
+/**
+ * @returns A new asynchronous lock
+ * @private
+ */
+function createLock() {
```
ryanking13, line 330: I moved these functions (createLock, loadDynlib) into a separate file in #3187.

tools/python
```diff
@@ -185,6 +185,27 @@ async function main() {
...
     };
 
+    py.runPython(
+        `
+        from pyodide._package_loader import SITE_PACKAGES, should_load_dynlib
+        from pyodide.ffi import to_js
+        import re
+            str(path) for path in SITE_PACKAGES.glob("**/*.so*")
+            if should_load_dynlib(path)
+        ])
```
ryanking13, line 195: I guess there is a syntax error here? (no starting brackets)

src/py/pyodide/_package_loader.py
ryanking13: #3167 would (hopefully) help fix the `shapely` import issue.

ryanking13:
Thanks a lot @hoodmane!

rth:
Great, thanks @hoodmane and @ryanking13 for getting to the bottom of it. We should probably document this more. 

So what's the practical impact on say loading CLAPACK .so in scipy from a venv https://github.com/pyodide/pyodide/issues/3186? How do I specify the location of the .so to load if it's in a separate package or do we now expect to vendor all .so in the wheel?  cc @lesteve

PR #3102, opened 2022-09-14
DOCS Add faq entry on importlib.invalidate_caches

hoodmane:


rth (approved):
Thanks! Indeed, I also forgot about this )

PR #3104, opened 2022-09-15
CI Fix ccache in package build

ryanking13:
Package build jobs in our CI were not using ccache at all. I'm not sure it worked before, but it doesn't work now.
(`_EMCC_CACHE` is not set during pyodide-build buildpkg)

This PR fixes it, and it also introduces a more strict cache key calculation script.

The good news is that when cached, there is some noticeable effect on the build time:

In [main](https://app.circleci.com/pipelines/github/pyodide/pyodide/4807/workflows/b6b92362-b1da-4ccf-a096-6baefe205996):

- `build-packages-no-numpy-dependents`: 22min
- `build-packages-numpy-dependents`: 28min
- `build-packages-opencv-python`: 29min

In [this PR](https://app.circleci.com/pipelines/github/ryanking13/pyodide/986/workflows/27130da3-e74f-418c-93c4-4169efe35947):

- `build-packages-no-numpy-dependents`: 12min (cache hit rate 97.50 %)
- `build-packages-numpy-dependents`: 24min (cache hit rate 76.66 %) (Maybe the value of `PYODIDE_JOBS` is not very optimal?)
- `build-packages-opencv-python`: 12min (cache hit rate 46.59 %)


.circleci/config.yml
```diff
@@ -135,8 +141,7 @@ jobs:
 
       - restore_cache:
           keys:
-            - -pkg2-v20220303-{{ checksum "Makefile.envs" }}
-            - -pkg2-v20220303
+            - -pkg3-v20220915-{{ checksum "build_hash.txt" }}-<< parameters.packages >>
```
ryanking13, line 144: Note that I used seperate cache keys for each incremental build jobs.
I wasn't able to find a way to mix `persist_to_workspace` and `save_cache` to make a single cache work...

pyodide-build/pyodide_build/buildpkg.py
```diff
@@ -160,9 +160,7 @@ def get_bash_runner() -> Iterator[BashRunnerWithSharedEnvironment]:
         env["PKG_CONFIG_PATH"] += f":{os.environ['PKG_CONFIG_PATH']}"
 
     with BashRunnerWithSharedEnvironment(env=env) as b:
-        b.run(
-            f"source {PYODIDE_ROOT}/emsdk/emsdk/emsdk_env.sh", stderr=subprocess.DEVNULL
-        )
+        b.run(f"source {PYODIDE_ROOT}/pyodide_env.sh", stderr=subprocess.DEVNULL)
```
ryanking13, line 163: `pyodide_env.sh` calls `emsdk_env.sh` interally, and it sets `_EMCC_CACHE`.

tools/calculate_build_cache_key.py
```diff
@@ -0,0 +1,61 @@
...
+    return ignorefile.read_text().splitlines()
+
+
+def main():
+    root: Path = Path(__file__).parent.parent
+    targets: list[Path] = [
+        root / "Makefile",
+        root / "Makefile.envs",
+        root / "pyodide-build",
+        root / "packages",
```
ryanking13, line 31: This check is quite strict. Any changes in `pyodide-build` or `packages` will invalidate the cache. Maybe we can relax the condition in the future.

rth:
That's great @ryanking13 , thanks for looking into it!

> This PR fixes it, and it also introduces a more strict cache key calculation script.

I think it's indeed too string right now. Ccache already has some internal hash of the env for each file it processes to determine if it should used the cached version or recompile. So for instance, if we change a package version, we shouldn't invalidate all caches. At worst, it would be incorrect for a few .so in a given package, and cache would handle that. 

I'm also not sure what's the right granularity, but I think right now it might be too detailed. Maybe just on Makefile.env & Makefile + something that has the Pyodide version could be enough? Though I think it's a good idea to have a separate Python script for it as you have done.

 BTW, FYI information there was some related exploration in https://github.com/pyodide/pyodide/pull/1262  on determining at what point do we need to rebuild things.

ryanking13:
> I'm also not sure what's the right granularity, but I think right now it might be too detailed. Maybe just on Makefile.env & Makefile + something that has the Pyodide version could be enough? Though I think it's a good idea to have a separate Python script for it as you have done.

Thanks for the suggestions @rth. I relaxed the condition a lot. How does it look now?

To be honest, I am quite ignorant of how ccache hash things (though I studied a bit while making this PR), so I wasn't sure if ccache is smart enough to correctly deal with minor meta.yaml updates.

ryanking13:
@rth, would it be OK to land this?

rth:
Thanks a lot @ryanking13 !

> I wasn't sure if ccache is smart enough to correctly deal with minor meta.yaml updates.

Yeah, it's a bit annoying to debug what it hashes exactly on (and why for a given file it's a hash miss rather than a hash hit), but overall you can rely on it to not use stale cache objects. For instance, locally you never have to clean ccache cache and it works fine.

 It also doesn't help we are using not the official ccache, but [a fork](https://github.com/emscripten-core/emsdk/blob/3e7e0ab8c352786fa5b799845f16627eb35a9f89/emsdk_manifest.json#L644) hacked together to work with emscripten.

PR #3108, opened 2022-09-15
Fix pyodide venv install from local file system for pure Python packages

hoodmane:
Currently `pip` can't invoke the Pyodide build backend, but it should still be able to build pure Python packages from source. To allow this, we need to restore `_PYTHON_SYSCONFIGDATA_NAME` so that it isn't inherited by subprocesses. Otherwise, they will fail to import `sysconfig`. Annoyingly, when pip invokes the subprocess it seems to pass `_PYTHON_SYSCONFIGDATA_NAME` down but not `PYTHONPATH` so just setting the `PYTHONPATH` isn't good enough. Also, restoring `_PYTHON_HOST_PLATFORM` breaks installation of Emscripten wheels.

### Checklists

- [x] Add / update tests


ryanking13:
Thanks @hoodmane!

PR #3109, opened 2022-09-15
DOCS Add docs for out of tree builds

hoodmane:


docs/development/out-of-tree.md
```diff
@@ -0,0 +1,96 @@
...
+command line commands:
+
+```sh
+pip install pyodide-build`
+pyodide build`
+```
+
+Pyodide currently only supports Linux for out of tree builds, though there is a
+good change it will work in MacOS too. If you are using Windows, try Windows
+Subsystem Linux.
```
ryanking13, line 23:
```suggestion
Subsystem for Linux.
```

docs/development/out-of-tree.md
```diff
@@ -0,0 +1,96 @@
...
+environment with:
+
+```sh
+pyodide venv .venv-pyodide
+```
+
+Enter it just like a normal virtual environment:
+
+```sh
+source .venv-pyodide/bin/python
```
ryanking13, line 56: Maybe add a note that it requires Node?

ryanking13 (commented):
Thanks, @hoodmane! I only checked the minor typos, so probably Roman would want to review this.

docs/development/out-of-tree.md
hoodmane: What ever happened to the `node-fetch` requirement? Do people not need that anymore?

ryanking13: What do you mean? I think we are using it for node < 18 which doesn't have builtin fetch.

https://github.com/pyodide/pyodide/blob/0eb6036b8556800eb0d074e2f9c08db8a32f7a35/src/js/compat.ts#L40

hoodmane: But say you have node 14 installed on your system and you do `pip install pyodide-build` and then start using the command line runner. How does `node-fetch` get installed?

ryanking13: It is not installed by `pip install pyodide-build` (obviously). Probably you installed it globally before, or probably you did `make` before. During make we install `node-fetch` "locally" while installing Pyodide JS package, and then we copy `node_modules` into the Pyodide root (which is quite strange and I think we can remove it now :/), so `node-fetch` is available when you are in Pyodide root repository.

https://github.com/pyodide/pyodide/blob/0eb6036b8556800eb0d074e2f9c08db8a32f7a35/Makefile#L70-L73

hoodmane: Okay great. And the reason that it works in CI is that I am testing the out of tree build stuff in circleci where node-fetch is already installed. Perhaps we should try to install `node-fetch` as part of creating the virtualenv because other than node-fetch it's self contained.

ryanking13: Sounds great. It would be great if Node >= v18 becomes dominant so we can simply tell people to use Node v18 and drop all node-fetch stuff... (The active LTS of Node v18 starts next month)

hoodmane:

rth (approved):
Very nice @hoodmane ! Looking forward to playing with it :)

docs/development/out-of-tree.md
```diff
@@ -0,0 +1,97 @@
...
+## Building binary packages for Pyodide
+
+If your package is a pure Python package (i.e., if the wheel ends in
+`py3-none-any.whl`) then it can be built in whichever way you normally build it.
+Otherwise, the procedure is simple. In your package directory run the following
+command line commands:
+
+```sh
+pip install pyodide-build
+pyodide build
```
rth, line 18: Also, where does it specify which version of Pyodide it's going to use? Is there a mapping between pyodide-build and pyodide versions somewhere?

docs/development/out-of-tree.md
```diff
@@ -0,0 +1,97 @@
...
+As a warning, things are pretty weird inside of the Pyodide virtual environment
+because `python` points to the Pyodide Python runtime. Any program that uses
+Python and is sensitive to the current virtual environment will probably break.
+
+You can install whatever dependencies you need with pip. For a pure Python
+package with a `tests` extra, the following will work:
+
+```sh
+pip install -e .[tests]
+```
```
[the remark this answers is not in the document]
hoodmane: Maybe we should mention that it is an editable install.

docs/development/out-of-tree.md
```diff
@@ -0,0 +1,97 @@
...
+
+```sh
+python -m pytest
+```
+
+but for instance `numpy` uses a file called `runtests.py`; the following works:
+
+```sh
+python runtests.py
+```
```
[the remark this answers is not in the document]
hoodmane: My point here is that if you have some Python script you can run it. You can use `python -c "some code"`, you can use `python -m some.module` and you can use `python some_script.py`. Maybe people will surmise this because it's called "Python".

docs/development/out-of-tree.md
```diff
@@ -0,0 +1,97 @@
...
+
+and you can pass options to it just like normal. Currently `subprocess` doesn't
+work, so if you have a test runner that uses `subprocess` then it cannot be
```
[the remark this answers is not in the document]
hoodmane: The pip runs in the host Python!

docs/development/out-of-tree.md
```diff
@@ -0,0 +1,97 @@
...
+Otherwise, the procedure is simple. In your package directory run the following
+command line commands:
+
+```sh
+pip install pyodide-build
+pyodide build
+```
+
+Pyodide currently only supports Linux for out of tree builds, though there is a
+good change it will work in MacOS too. If you are using Windows, try Windows
```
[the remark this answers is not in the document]
hoodmane: Yeah. We should add a mac CI run for this.

docs/development/out-of-tree.md
hoodmane: By default it uses the same version as `pyodide-build`. We really ought to add a separate command to install the xbuildenv rather than doing it automatically.

docs/development/out-of-tree.md
```diff
@@ -0,0 +1,97 @@
+(out-of-tree)=
```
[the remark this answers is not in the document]
hoodmane: I don't have a strong opinion, if you think of something better I am happy to change it. By the way, is there a way to add a redirect to the docs? Then if we move it later we can just make the old URL redirect to the new one and it removes some of the worry about naming it right from the beginning.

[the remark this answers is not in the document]
kylebarron: Given that there's already a page `building-from-sources.md`, maybe `building-out-of-tree.md` would make sense?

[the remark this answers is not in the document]
hoodmane: Well `building-out-of-tree` it isn't shorter than `out-of-tree`. And this is about both building and testing. But I would be okay with that.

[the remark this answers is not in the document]
jobovy: How about ``build-and-test-packages.md``? That way it's future-proof once ``pyodide build`` becomes the standard way to create packages.

docs/development/out-of-tree.md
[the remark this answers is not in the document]
hoodmane: Also, `pytest` often doesn't work for `numpy` for me. I get the following error:
```
ImportError: Error importing numpy: you should not try to import numpy from
        its source directory; please exit the numpy source tree, and relaunch
        your python interpreter from there.
```

docs/development/out-of-tree.md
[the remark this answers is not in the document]
hoodmane: Okay went with @jobovy's suggestion. Thanks guys!

docs/development/out-of-tree.md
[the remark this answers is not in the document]
hoodmane: Opened #3123.

docs/development/out-of-tree.md
hoodmane: Wow node-fetch v3 has 8mb of `node_modules`, I guess because it requires web-streams-polyfill which is 200kb of code but it has 15 different variants and each variant has a source map. But node-fetch v2 does not require web-streams-polyfill and leads to only 512kb of `node_modules`.

hoodmane:
@henryiii again I don't understand the mypy error here:
```
    res = subprocess.run(
        ["npm", "i", "node-fetch@2"],
        cwd=xbuildenv_root,
        capture_stdout=True,
        encoding="utf8",
    )
```
looks fine to me...

hoodmane:
Oh okay, I just misnamed `capture_output`. The error is confusing because of the many overloads of `subprocess.run`.

docs/development/out-of-tree.md
hoodmane: Okay, I updated `create_xbuildenv` to vendor the appropriate `node-fetch` stuff into `node_modules`. This should mean that we don't need any comment about `node-fetch`. It's also good to have the `node_modules` folder in the xbuildenv root because that means that the execution environment won't be polluted by some random other `node_modules` folder in a parent directory.

henryiii:
Isn't it `"utf-8"`?

hoodmane:
It seems to also accept `"utf8"` without any trouble.

PR #3113, opened 2022-09-16
Fix pypi deploy for patch releases

hoodmane:
Apparently `repository_dispatch` only works on the main branch and what we need is `workflow_dispatch`. After we merge this, we can backport it to stable and I can do an ad-hoc test by trying to use it to deploy pyodide-build v0.21.3.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -478,12 +478,13 @@ jobs:
           name: Deploy to PyPI
           command: |
             # Ask github actions to deploy for us
+            export PAYLOAD={"ref":"${CIRCLE_TAG}","inputs":{}}
             curl \
               -X POST \
               -H "Accept: application/vnd.github.v3+json" \
-              -H "Authorization: token ${GITHUB_TOKEN}" \
-              https://api.github.com/repos/${CIRCLE_PROJECT_USERNAME}/${CIRCLE_PROJECT_REPONAME}/dispatches \
-              -d '{"event_type":"deploy-release","client_payload":{}}'
+              -H "Authorization: Bearer ${GITHUB_TOKEN}" \
+              https://api.github.com/repos/${CIRCLE_PROJECT_USERNAME}/${CIRCLE_PROJECT_REPONAME}/dispatches/deploy-release \
+              -d "$PAYLOAD"
 
       - run:
           name: Set PYODIDE_BASE_URL
--- a/.github/workflows/deploy_release.yml
+++ b/.github/workflows/deploy_release.yml
@@ -1,6 +1,6 @@
 name: deploy-release
 on:
-  repository_dispatch:
+  workflow_dispatch:
     types: [deploy-release]
 jobs:
   deploy-pyodide-build-pypi:


rth (approved):
Thanks!

hoodmane:
Well this still doesn't work.

ryanking13:
I think it will work when we release a new version with a tag, as `stable` branch have no tag so `CIRCLE_TAG` would be empty.

hoodmane:
I don't think the `curl` request works at all. I haven't figured out how to make it work with `WORKFLOW_DISPATCH`. I'm trying to follow the directions here:
https://docs.github.com/en/rest/actions/workflows#create-a-workflow-dispatch-event

but I get the response:
```
{
  "message": "Not Found",
  "documentation_url": "https://docs.github.com/rest/reference/actions#create-a-workflow-dispatch-event"
}
```

ryanking13:
https://github.com/pyodide/pyodide/actions/runs/3064780993 but I see it was triggered 3 days ago?

hoodmane:
That was a manual trigger from the web interface, not from `curl`. Since this PR was merged, I have been unable to trigger the job via curl.

.circleci/config.yml
```diff
@@ -478,12 +478,13 @@ jobs:
...
             # Ask github actions to deploy for us
+            export PAYLOAD={"ref":"${CIRCLE_TAG}","inputs":{}}
             curl \
               -X POST \
               -H "Accept: application/vnd.github.v3+json" \
-              -H "Authorization: token ${GITHUB_TOKEN}" \
-              https://api.github.com/repos/${CIRCLE_PROJECT_USERNAME}/${CIRCLE_PROJECT_REPONAME}/dispatches \
-              -d '{"event_type":"deploy-release","client_payload":{}}'
+              -H "Authorization: Bearer ${GITHUB_TOKEN}" \
+              https://api.github.com/repos/${CIRCLE_PROJECT_USERNAME}/${CIRCLE_PROJECT_REPONAME}/dispatches/deploy-release \
```
ryanking13, line 486: It seems like the [order is different](https://docs.github.com/en/rest/actions/workflows#create-a-workflow-dispatch-event)

```
/repos/{owner}/{repo}/actions/workflows/{workflow_id}/dispatches
```

```suggestion
https://api.github.com/repos/${CIRCLE_PROJECT_USERNAME}/${CIRCLE_PROJECT_REPONAME}/deploy-release/dispatches \
```

ryanking13 (commented):
> That was a manual trigger from the web interface, not from curl. Since this PR was merged, I have been unable to trigger the job via curl.

Oh, okay. I thought that it was triggered by curl.

.circleci/config.yml
hoodmane: Can create an API token and test this out? I tried a bunch of different variations like this and none of them were working for me. But maybe I was just being dumb.

ryanking13: I removed the `type: [deploy-release]` part in [deploy_release.yml](https://github.com/pyodide/pyodide/blob/b0dc19e4a169c9dcf97e21fe0ac2c635a0900f13/.github/workflows/deploy_release.yml#L4), and I was able to run a workflow with curl.

I think it was blocking a workflow run (it is [mentioned in `repository_dispatch` docs](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#repository_dispatch), but probably it was also affecting `workflow_dispatch`:

```
When you make a request to create a repository_dispatch event, you must specify an event_type to describe the activity type.
By default, all repository_dispatch activity types trigger a workflow to run.
You can use the types keyword to limit your workflow to run when a specific event_type value is sent in the repository_dispatch webhook payload.
```

Issue #3093, opened 2022-09-10
Move micropip to a separate repo in the org


rth:
Continuing with the effort of making the code base more approachable (https://github.com/pyodide/pyodide/issues/2014), I would propose to move micropip to a separate repo in the Pyodide org.

### Advantages 
 - mich faster iteration cycle to change anything in micropip. CI should take minutes not an hour as currently.
 - for new contributors, it's also much easier to start since it's a pure Python package they don't need to deal with the rest of the build system.
 - can update micropip for a given Pyodide release by loading the pure Python wheel from PyPi with `pyodide.loadPackage`. The release schedule of micropip can become independent from Pyodide.
 - there are several other projects in this space (e.g. emscripten-forge, https://github.com/ethanhs/python-wasm) and this would make it  easier to reuse and contribute back fixes if necessary. There is still the FFI question which is more difficult, but right now even if emscripten-forge wanted to support installing Pyodide packages (and assuming  binary compatibility) they have no other way short of vendoring it. Similarly, I'm hoping this could encourage some of the micropip related fixes from JupyterLite to be contributed upstream.

### Advantages 

- a bit more setup work, but we have done this for a lot of packages now, so it's mostly copy-pasting
- makes it harder to make large changes across different components of Pyodide, though I think micropip is pretty stable now and there are rarely PRs that need to change both the internals and micropip at the same time.

bollwyvl:
What `piplite`/`jupyterlite` hacks are wanted? Happy to shed some stuff!

Even if _all_ our hacks weren't desirable, it would be lovely to have a few more documented ways to customize micropip, rather than trying to solve some things Once And For All... presenting the two together:

- handling multiple package sources
  - we [check](https://github.com/jupyterlite/jupyterlite/blob/main/packages/pyolite-kernel/py/piplite/piplite/piplite.py#L65) for a custom Warehouse (or [warehouse-like](https://github.com/pypi/warehouse/blob/main/warehouse/legacy/api/json.py#L96) [file format](https://jupyterlite.readthedocs.io/en/latest/_static/pypi/all.json)) for custom/offline packages
  - notionally, `micropip.add_pypi_json_resolver(key: str, get_pypi_json: Callable[[str, dict], PackageReleases])`
- disabling package sources
  - for offline (but still dynamic) installations, it's nice to have some guardrails in place to avoid calling out to Big PyPI
  - notionally, `micropip.remove_pypi_json_resolver(key: str)`
[...]

PR #3122, opened 2022-09-18
Include micropip from a separate repository

rth:
Closes https://github.com/pyodide/pyodide/issues/3093

Includes micropip from https://github.com/pyodide/micropip

One thing I'm not sure about is whether we still want to run the micropip tests there or if we rely on them passing in https://github.com/pyodide/micropip. Also, it's not very clear to me right now how to run tests from that repo here, particularly since we don't want tests to be included in the wheel.

We should probably have a separate documentation site for micropip, in that repo, so they are updated independently from the main docs.


ryanking13:
> One thing I'm not sure about is whether we still want to run the micropip tests there or if we rely on them passing in https://github.com/pyodide/micropip.

I think we should run micropip tests in `pyodide/micropip`. It looks like `pyodide/micropip` vendors subset of `src/py` to run tests locally (outside of browser), but I think it is okay to drop outside of browser support and run full test suite inside Pyodide.

> We should probably have a separate documentation site for micropip, in that repo, so they are updated independently from the main docs.

Yes, indeed we need a seperate docs for micropip.

rth:
> outside of browser support and run full test suite inside Pyodide.

I'm not sure you can do that, because a number of tests need to start an HTTP server which you can't do from within the browser (e.g. [here](https://github.com/pyodide/micropip/blob/main/tests/test_micropip.py#L284)). It's also rather practical to have host tests when possible, as it for instance is correctly counted toward code coverage.

Also, it's a bit questionable whether it's useful to run those for each commit in this repo, as opposed to when updating micropip only (though yes, this would be harder to setup). 

The other solution is we could keep some of the tests in `test_micropip.py`, and have a more extensive test suite in the other repo.

ryanking13:
> The other solution is we could keep some of the tests in `test_micropip.py`, and have a more extensive test suite in the other repo.

That sounds reasonable to me.

rth:
CI is good, can anyone have a look so we can merge this? Thanks!

packages/micropip/meta.yaml
```diff
@@ -1,13 +1,16 @@
...
+
+source:
+  sha256: 693a7fd7ce92fd95eaf9d88fe243f45951ad5c694cccf89493cbcc3666bbdeab
+  url: https://files.pythonhosted.org/packages/14/19/e5d36b2cc156c66858493dc118956f6fed2b1408e62be12e4ffed392b7cb/micropip-0.1.0-py3-none-any.whl
+
 requirements:
   run:
     - pyparsing
     - packaging
-    - distutils # TODO: remove once there is a release with https://github.com/pypa/packaging/pull/396
```
ryanking13, line 11: It is good to see micropip having fewer dependencies!

ryanking13 (approved):
Thanks a lot @rth!

PR #3124, opened 2022-09-18
Use upstream chrono for cryptography

hoodmane:
Since Chrono made a release fixing the Emscripten target we don't need the Chrono patch anymore. Still need the `parking_lot` patch.


ryanking13 (approved):
Thanks, @hoodmane! LGTM if the build pass.

hoodmane:
Oh I should have added a changelog entry for updating the package version.

PR #3128, opened 2022-09-19
BLD simplify building CMake based packages

ryanking13:
### Description

This PR makes building CMake based packages easier.

It is better than #2838, as (ideally) users don't have know anything. Everything is done behind the scene.

- Introduce a custom toolchain file for Pyodide. It inherits most of settings from the [original Emscripten toolchain file](https://github.com/emscripten-core/emscripten/blob/main/cmake/Modules/Platform/Emscripten.cmake) with some modifications. Packages built by `pyodide-build` will automatically use that toolchain file.
- Proxy `cmake` in `pywasmcross.py`, in order to proxy other compiler toolchains.


emsdk/patches/0001-Don-t-set-toolchain-file-if-CMAKE_TOOLCHAIN_FILE-env.patch
```diff
@@ -0,0 +1,33 @@
+From a04edfb157cd8bd4f07abc9bafa9cb0e8c51e4d8 Mon Sep 17 00:00:00 2001
+From: Gyeongjae Choi <def6488@gmail.com>
+Date: Tue, 30 Aug 2022 05:57:51 +0000
+Subject: [PATCH] Don't set toolchain file if CMAKE_TOOLCHAIN_FILE env variable
```
ryanking13, line 4: This patch is upstreamed and merged (https://github.com/emscripten-core/emscripten/pull/17757)

pyodide-build/pyodide_build/pywasmcross.py
```diff
@@ -363,6 +374,33 @@ def replay_genargs_handle_argument(arg: str) -> str | None:
     return arg
 
 
+def get_cmake_compiler_flags() -> list[str]:
```
hoodmane, line 377: Could you add a test for this to make codecov happy? If you are feeling generous, you could also add tests for `remove_avoided_requirements` and `install_reqs`.

hoodmane (approved):
It generally good to me, but I don't know much about CMake. Maybe @henryiii could take a look at it if he has time.

ryanking13:
> It generally good to me, but I don't know much about CMake. Maybe @henryiii could take a look at it if he has time.

Sounds good to me. If Henry is busy, I'll wait for a few days and merge after adding some tests.

Actually, I read some parts of CMake books that he wrote while making this PR. Thanks for the amazing books and workshops :)

henryiii:
I'll try to look at it tomorrow. Remind me if I don't!

henryiii:
Looks pretty good, hard to tell what the implications are without trying it out - I'm hoping to get to a point where I can test scikit-build-core soonish. One thought - where are you getting the list of build requirements? One very important trick I think we will heavily be moving to is to declare cmake (and ninja) via PEP 517's `get_requires_for_build_wheel` instead of PEP 518's static list. This is required to work on systems that already provide cmake of a sufficient version and don't support wheels. I expect this will also be true in Pyodide, as if the mechanism is triggered, it will see the recent-enough CMake and will not inject cmake as a requirement. What version of CMake is this now?

ryanking13:
Thanks for the review @henryiii!

> What version of CMake is this now?

I think the cmake version installed in our Docker image is 3.22.6.

> One thought - where are you getting the list of build requirements?

We are relying on the pypa/build to detect build requirements.

https://github.com/pyodide/pyodide/blob/1a66912341e12f8538f82acbe5e4f06188236115/pyodide-build/pyodide_build/pypabuild.py#L86

ryanking13:
Merging so that this can be used in #3138. Thanks for the reviews @hoodmane and @henryiii! I think we can fix cmake dependency injection issue in a separate PR.

PR #3129, opened 2022-09-19
Add debug build option for js package

ryanking13:
This PR uses [terser's conditional compilation feature](https://github.com/terser/terser#conditional-compilation) to allow debug build for Pyodide JS packages.
One can use it with `PYODIDE_DEBUG_JS=1 make`.

For now, I am trying to use this feature to debug shared library loading procedure, but maybe we can add more debug logs in the future.

@hoodmane Do we have a general env variable for debug build? I see something like `PYODIDE_SYMBOLS`, but the name doesn't seem very general.


hoodmane:
Currently we have:
* PYODIDE_SYMBOLS
* EXTRA_CFLAGS="-D DEBUG_F"
* this PR adds `PYODIDE_DEBUG_JS`

would be good to have a flag that sets all of these.

src/js/rollup.config.js
```diff
@@ -33,7 +35,12 @@ function config({ input, output, name, format, minify }) {
       nodeResolve(),
       minify
         ? terser({
-            compress: true,
+            compress: {
+              dead_code: true,
+              global_defs: {
+                DEBUG,
+              },
+            },
```
hoodmane, line 43: Do we maybe not want to compress in debug mode? Or have an environment variable to disable compression?

ryanking13: Good point, I should also disable the compression

ryanking13:
> would be good to have a flag that sets all of these.

Okay, how about `PYODIDE_DEBUG`? (I am afraid just 'DEBUG' may affect other tools that we don't know)

hoodmane:
Why not keep `PYODIDE_DEBUG_JS` as a separate environment variable and then have `PYODIDE_DEBUG` set all three of them?

ryanking13:
> Why not keep `PYODIDE_DEBUG_JS` as a separate environment variable and then have `PYODIDE_DEBUG` set all three of them?

I thought that it would be nice to have a single variable instead of havine multiple ones. But I have no strong opinion. I'll go with your idea.

hoodmane:
Well I think it's nice to have one variable that sets everything, but sometimes you really only need one of the debug features enabled (and enabling all of them at once may create huge amounts of log noise).

hoodmane:
Another option which may be better is to have separate runtime controls for each noisy thing and `PYODIDE_DEBUG=1` could compile in all debug features but by default leave noisy ones disabled.

I have this for instance with `tracerefs` which requires `-D DEBUG_F` to be set to compile it in, but then it also is behind a runtime flag. This is because it logs so much information that if you keep it enabled all the time it's impossible to find anything relevant in the torrent of logging.

ryanking13:
> but sometimes you really only need one of the debug features enabled

Right, that make sense.

hoodmane (approved):

ryanking13:
Thank you for the review @hoodmane!

ryanking13:
Oh, I should've checked linter before merging...

PR #3130, opened 2022-09-19
Add descr_get handler to JsMethod

hoodmane:
This is the converse of #3103. It allows a JavaScript function to be used as a Python method. When it is called, `this` will be a `PyProxy` pointing to the Python object the method is called on.

### Checklists

- [x] Add / update tests


ryanking13 (approved):
It looks good to me. Thanks @hoodmane!

rth:
Thank you!

hoodmane:
It seems like this introduced a regression in the matplotlib canvas backend. So far it is nontrivial to reproduce.

ryanking13:
> It seems like this introduced a regression in the matplotlib canvas backend. So far it is nontrivial to reproduce.

Yes, It seems like the error doesn't always happen so it is not easy to analyze.

hoodmane:
Yeah I was unable to reproduce it locally either. When I run benchmarks they all work. Strange...

PR #3132, opened 2022-09-19
Keep webpackIgnore comments when building js package

ryanking13:
Fixes:  #3032

It adds terser config to not strip /* webpackIgnore */ comments. I think this is better than using sed after build.


hoodmane:
Woww nice! Thanks @ryanking13 --- apparently it was a good idea to read the terser docs.

PR #3136, opened 2022-09-20
DOCS Make docs wider

### LINKED ISSUE #3135: The content area in the docs is too narrow
## 📚 Documentation

In the documentation strings, rendered code examples only fit 63 characters of width. It would be nice if we could make the content area a bit larger so that code examples fit at least 80 characters. On my screen, the content area is exactly the middle third of the screen, with the left and right thirds devoted to menus.

hoodmane:
Resolves #3135. Thanks to @jobovy for telling me how to do this.

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -11,6 +11,8 @@ from pathlib import Path
 from typing import Any
 from unittest import mock
 
+panels_add_bootstrap_css = False
+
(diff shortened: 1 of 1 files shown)


ryanking13 (approved):
Thanks @hoodmane and @jobovy!

Issue #3116, opened 2022-09-16
LightGBM Support Attempt


yzhliu:
## 🐍 Package Request

- Package Name and Version: lightgbm 3.3.2
- Package URL:
  * https://github.com/microsoft/LightGBM/tree/v3.3.2
  * https://files.pythonhosted.org/packages/8f/51/d272fa065be3be615d279be915705fa3824b86155e36c974cfb8d3ceec1e/lightgbm-3.3.2.tar.gz
- Package Dependencies that needs to be resolved first: N/A

## Attempts

I actually have done some work try to get it compiled, and it did. See https://github.com/yzhliu/pyodide/tree/lightgbm/packages/lightgbm for details. And it also imports successfully in the browser.

However when I tried to run the following script from [LightGBM tutorial](https://lightgbm.readthedocs.io/en/v3.3.2/Python-Intro.html), it runs into several “*RuntimeError: null function or function signature mismatch*” error in Chrome.

```python
import numpy as np
import lightgbm as lgb
data = np.random.rand(500, 10)  # 500 entities, each contains 10 features
label = np.random.randint(2, size=500)  # binary target
train_data = lgb.Dataset(data, label=label)
param = {'num_leaves': 31, 'objective': 'binary'}
param['metric'] = 'auc'
num_round = 10
bst = lgb.train(param, train_data, num_round)
```

### 1. Problematic support for std::function<std::vector<double>(int)> return type

It first failed at https://github.com/microsoft/LightGBM/blob/v3.3.2/src/c_api.cpp#L1110. It seems to be complaining about `std::vector<std::function<std::vector<double>(int)>>` return type. In addition, [the usage of template](https://github.com/microsoft/LightGBM/blob/v3.3.2/src/c_api.cpp#L2380) in std::function also seems to be problematic. Therefore I changed it to a class and it went through. See [the patch](https://github.com/yzhliu/pyodide/blob/lightgbm/packages/lightgbm/patches/0001-RowFunctionFromDenseMatricCB.patch) for details.

### 2. Template class and virtual method?

[...]

PR #3138, opened 2022-09-20
Add lightgbm package

yzhliu:
### Description

add the lightgbm package. related issue #3116 .

Question though: I used patch to specify setup.py argument (--noomp), wondering if there's a better way to do so?

### Checklists

- [x] Add / update tests


hoodmane:
I think there's a key called `backend-args` where you should put `--noomp`.

packages/lightgbm/meta.yaml
```diff
@@ -0,0 +1,25 @@
+package:
+  name: lightgbm
+  version: 3.3.2
+source:
+  url: https://files.pythonhosted.org/packages/8f/51/d272fa065be3be615d279be915705fa3824b86155e36c974cfb8d3ceec1e/lightgbm-3.3.2.tar.gz
+  sha256: 5d25d16e77c844c297ece2044df57651139bc3c8ad8c4108916374267ac68b64
+  patches:
+    - patches/0001-noomp-True.patch
+build:
+  script: |
```
hoodmane, line 10:
```suggestion
  backend-flags: --noomp
  script: |
```

packages/lightgbm/meta.yaml
```diff
@@ -0,0 +1,25 @@
...
+    export CMAKE_TOOLCHAIN_FILE=$PYODIDE_ROOT/packages/lightgbm/cmake/Toolchain.cmake
+  exports: "whole_archive"
```
hoodmane, line 12: Does it not work with `exports: "requested"`?

hoodmane:
Also, maybe this can be simplified after we merge #3128 @ryanking13?

ryanking13:
> Also, maybe this can be simplified after we merge #3128 @ryanking13?

Yes, I think we won't need a toolchain file after #3128.

ryanking13:
Thanks for the porting work @yzhliu. We merged #3128 so I think you won't need to use a custom CMAKE_TOOLCHAIN_FILE now.

yzhliu:
@ryanking13 @hoodmane cool, I'll change accordingly.

packages/lightgbm/meta.yaml
yzhliu: It'll fail saying "OSError: [Errno 7] Argument list too long: 'em++'"

yzhliu:
@hoodmane @ryanking13 could you do another round of code review? and how shall I fix the test coverage check?

pyodide-build/pyodide_build/pypabuild.py
```diff
@@ -98,8 +98,17 @@ def _build_in_isolated_env(
         symlink_unisolated_packages(env)
         install_reqs(env, builder.build_system_requires)
         installed_requires_for_build = False
+        # "--build-option" contains custom options that only apply to current package,
+        # passing it to the dependencies may cause error.
+        # e.g., lightgbm accepts "--nomp" as one of its setup.py arguments,
+        # but "--nomp" isn't supported by its dependencies such as numpy, scipy, etc.
+        reqs_config_settings = {
+            k: v for k, v in config_settings.items() if k != "--build-option"
```
ryanking13, line 106: Is our `backend-flags` propagating to dependencies? That seems very weird.

ryanking13: Ah, okay we are passing backend_flags in `builder.get_requires_for_build`. I think we should only pass it to `builder.build` and not  `builder.get_requires_for_build` @hoodmane?

ryanking13:
Thanks @yzhliu! I think the test coverage is okay. I think the backend flags passed to the dependencies is a bug and otherwise it LGTM.

pyodide-build/pyodide_build/pypabuild.py
hoodmane: I certainly didn't mean for it to propagate to dependencies!

ryanking13: > I certainly didn't mean for it to propagate to dependencies!

Okay, thanks for the confirmation. @yzhliu Could you remove all backend-flags(config_settings) passed to `builder.get_requires_for_build`? Thanks for catching this out!

yzhliu: sure, I'll do a sweep over the related code.

yzhliu: @ryanking13 seems here are the only two usage of `builder.get_requires_for_build` in the code base.

ryanking13: Yes, there are two appearances of `builder.get_requires_for_build`.

yzhliu: I changed both of them, is the PR good to go? or did I miss anything else?

ryanking13: Is it pushed? I can't notice the change and it looks like the last push you made is 3 days ago before the above comments.

yzhliu: no the original commit already has the two changes,
https://github.com/pyodide/pyodide/pull/3138/files#diff-8a5df82b2fccc945f9e52b43277a75fa9e7a3f3ea1f5b03725b97d186f84225fR110
https://github.com/pyodide/pyodide/pull/3138/files#diff-8a5df82b2fccc945f9e52b43277a75fa9e7a3f3ea1f5b03725b97d186f84225fR122

ryanking13 (approved):
Thanks again @yzhliu!

PR #3139, opened 2022-09-21
CI Fix PyPI deploy

ryanking13:
Try to fix #3113 again


rth:
Why do we actually have to trigger it indirectly? I would say we can deploy to PyPi on tag unrelated to the other CI jobs. If the tag is made we can assume that the CI is fine (and even if it fails it's likely some temporary connection issue).

ryanking13:
> Why do we actually have to trigger it indirectly? I would say we can deploy to PyPi on tag unrelated to the other CI jobs. If the tag is made we can assume that the test suite is fine (and even if it fails it's likely some temporary connection issue).

Yes, invoking it indirectly is a pain and I am +1 with your idea. JavaScript package need to be deployed after the build as it uses some build artifacts, but `pyodide-build` is okay to be deployed unrelated to the other CI jobs.

ryanking13:
BTW, unrelated to this topic, I heard that `iodide.io` url is dead recently, is our s3 bucket still working?

https://github.com/pyodide/pyodide/blob/024dcbe8f06bb106045f1f66d32bb8b8ffdb0e2a/.circleci/config.yml#L494-L505

rth:
Yes, the s3 bucket is owned by me so we are fine.  The name is for historical reasons from where it was served via iodide.io

Though actually, it might be good to create an AWS Pyodide account at some point and host it there, particularly now that the bandwidth fees are much-reduced thanks to JsDelivr. Though it would still be some work to migrate, as one cannot just transfer a bucket but has to re-create a new one with exactly the same metadata, so maybe it's not worth it for now.

rth (approved):
Thanks!

rth:
> is our s3 bucket still working?

Well, we would see some concerned user reports if it wasn't  :)

hoodmane:
> Why do we actually have to trigger it indirectly?

We want one of two things to happen:
1. Deploy everything
2. Deploy nothing

What we don't want is a test fails and so some things get deployed and others don't. If a test failure causes no deploy to occur we can fix the test and get a working version. If there's a partial deploy, the release is likely to be irreparably screwed up.

hoodmane:
For instance in the v0.21.3 release there was a real test failure which I had to fix. This has happened several times. I think it's reasonable to assume that at some point there will be real test failures because we make mistakes and plan for it.

rth:
I get your point, but this is not happening currently. We have plenty of releases where some of the things failed while others deployed. Pre-built docker is probably still failing. So having this complicated setup only increases the likelihood that this pyodide-build will fail to deploy, and doesn't guarantee the all-or-nothing apporach.

hoodmane:
> it's also maintainers responsibility to check that the branch is green before tagging it.

Yeah. Perhaps we should add to the release checklist to wait until the CI runs to completion before adding a tag. My point is that I make a large number of mistakes in making releases and anything that makes it a bit less error prone is helpful.

Particularly in the case of 0.21.3 the only new change was the version numbers but I updated those incorrectly and caused the failure.

But you have some good points. \*shrug*

PR #3140, opened 2022-09-22
Use pyodide-cli entrypoint for out-of-tree CLI

ryanking13:
@hoodmane Could you check that this it is working as expected? (especially the behavior of `create_pyodide_script`)
Feel free to take this over.

```sh
$ pip install ./pyodide-build
$ pyodide
                                                                                                                                                             
 Usage: pyodide [OPTIONS] COMMAND [ARGS]...                                                                                                                  
                                                                                                                                                             
 A command line interface for Pyodide.                                                                                                                       
 Other CLI subcommands are registered via the plugin system by installing Pyodide compatible packages (e.g. pyodide-build).                                  
                                                                                                                                                             
╭─ Options ─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╮
│ --version                                                                                                                                                 │
│ --help             Show this message and exit.                                                                                                            │
╰───────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╯
╭─ Commands ────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╮
│ build                Use pypa/build to build a Python package                                                                                             │
│ venv                 Create a Pyodide virtual environment                                                                                                 │
╰───────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╯

### CODE REVIEW

**Review by rth — COMMENTED** [review]
Code wise looks great, thanks @ryanking13 . 

Let's let Hood test it indeed.

File: pyodide-build/setup.cfg [hunk 1]
```diff
@@ -29,11 +29,17 @@ install_requires =
     build==0.7.0
     virtualenv
     pydantic>=1.10.2
+    pyodide-cli>=0.2.0
+    typer
```
rth, line 33: You can probably remove the typer dependency. It's up to pyodide-cli to define the version of typer it need.

pyodide-build/pyodide_build/out_of_tree/venv.py
```diff
@@ -192,11 +181,15 @@ def create_pyodide_script(venv_bin: Path) -> None:
...
+    original_pyodide_cli = shutil.which("pyodide")
+    if original_pyodide_cli is None:
+        raise RuntimeError("ERROR: pyodide cli not found")
+
     pyodide_path.write_text(
         dedent(
             f"""
             #!/bin/sh
-            {environment} exec {sys.executable} -m pyodide_build.out_of_tree "$@"
+            {environment} exec {original_pyodide_cli} "$@"
```
rth, line 192: Actually can't you use the previous apporach that only relied on PYTHONPATH? That would be more robust as compared to hoping that that pyodide is both in PATH and PYTHONPATH.

ryanking13: I am not sure I am doing the correct thing here. I hope Hood can check and fix this...

hoodmane: Yeah I'll try to get to it this evening. Today is the first day of classes so I've been a bit busy getting everything set up.

ryanking13: Thanks!

hoodmane: Well what you had worked but we can do it much simpler. To be clear, this simpler approach would have worked from the beginning but for some reason I didn't think of it the first time.

hoodmane (approved):

hoodmane:
Okay I guess that doesn't quite work...

ryanking13:
> Okay I guess that doesn't quite work...

The error message is about the emscripten compiler so probably setting PATH should fix? I guess users need to do `./emsdk activate` by themselves before using `pyodide build` as we don't vendor emscripten compilers in xbuildenv?

hoodmane:
I think it is necessary and sufficient to set `PYODIDE_ROOT`.

hoodmane:
Hmm no you're right we need to set the PATH too because we don't vendor the emscripten compiler in xbuildenv.

hoodmane:
Okay I'm going to merge this.

ryanking13:
Thanks for finishing this @hoodmane!

PR #3148, opened 2022-09-25
Emscripten 3.1.23

hoodmane:
### Checklists


ryanking13:
I guess the build error is related to the change that Emscripten's default C++ standard is changed to `c++17` and `random_shuffle` [has been removed](https://en.cppreference.com/w/cpp/algorithm/random_shuffle) in c++17.

ryanking13:
It looks like we are adding `std=c++14` in [LDFLAGS](https://github.com/pyodide/pyodide/blob/9a3365a6d0325caec5c609967c11edc041cb6d99/Makefile.envs#L95) but not in CFLAGS now. I'm not sure why.

hoodmane:
I opened an issue about it here:
https://github.com/matplotlib/matplotlib/issues/24010

ryanking13:
Thanks, probably we need to add `std=c++14` (or `std=c++17` if it works) in our CFLAGS.

ryanking13:
Hmm it seems like the build.log is not created after the build.

CallumJHays:
The issue seems to be typer not parsing the "-std=c++14" correctly as the value of the argument "--cxxflags" .
This throws an error, stopping the build before build.log is ever created for that module.

I was able to get typer to parse it correctly by making the following changes to `pyodide_build/buildall.py`:

image

hoodmane:
Thanks for looking into this. I guess we should change all the arguments
like that...

ryanking13:
The CI failures are not from this PR and are fixed in main branch, I synced to the main and rerun CI. Fingers crossed 🤞

hoodmane:
Thanks @CallumJHays and @ryanking13!

PR #3150, opened 2022-09-26
DOC Remove warnings about pyodide-cdn2.iodide.io

ryanking13:
Since mozilla took down iodide.io domain recently, I think we don't need the warnings about `pyodide-cdn2.iodide.io` now.

### DIFF
--- a/docs/project/changelog.md
+++ b/docs/project/changelog.md
@@ -14,6 +14,9 @@ substitutions:
 
 ## Unreleased
 
+- `pyodide-cdn2.iodide.io` is not available anymore. Please use `https://cdn.jsdelivr.net/pyodide` instead.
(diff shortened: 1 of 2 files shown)


rth (approved):
Thanks, indeed.

hoodmane:
Is there a way to get `https://cdn.jsdelivr.net/pyodide/v0.21.2/debug/console.html` to not respond with 403 forbidden? I'm not sure how to use the debug console from a url anymore.

rth:
No, we can't serve HTML from JsDelivr (they explicitly don't allow it), so we would need to deploy it somewhere else. 

I'm thinking that maybe we should have a separate bucket for the dev deployment (or even deploy on github pages).

PR #3154, opened 2022-09-27
Handle shared libraries with version suffix

ryanking13:
### Description

Fix handling shared libraries with version suffix such as `libabc.so.11.22.0`. It fixes both building and loading packages.

### Checklists

- [x] Add / update tests


hoodmane (approved):
Looks good, thanks!

PR #3156, opened 2022-09-27
Fix cmdline runner when stdout doesn't have columns/rows attributes

hoodmane:
This fixes a bug that came up in https://github.com/numpy/numpy/pull/21895.


ryanking13 (approved):
Thanks @hoodmane, it looks good to me when CI pass. It seems like the network connection in the CircleCI server is little bit unstable now.

PR #3157, opened 2022-09-28
BLD Retry when package download fails

ryanking13:
We sometimes encounter `urllib.error.URLError` when downloading packages in CI, which is mostly caused by timeout.
I'm not sure this would fix the situation, but it is better then doing nothing.

### DIFF
--- a/pyodide-build/pyodide_build/buildpkg.py
+++ b/pyodide-build/pyodide_build/buildpkg.py
@@ -15,6 +15,7 @@ import subprocess
 import sys
 import sysconfig
 import textwrap
+import urllib
 from collections.abc import Generator, Iterator
 from contextlib import contextmanager
 from datetime import datetime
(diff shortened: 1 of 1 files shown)


pyodide-build/pyodide_build/buildpkg.py
```diff
@@ -234,7 +235,19 @@ def download_and_extract(
...
     # We only call this function when the URL is defined
     url = cast(str, src_metadata.url)
-    response = request.urlopen(url)
+    retry_cnt = 3
+    while True:
+        try:
+            response = request.urlopen(url)
+        except urllib.error.URLError as e:
+            if retry_cnt == 0:
+                raise RuntimeError(f"Failed to download {url}") from e
```
rth, line 244:
```suggestion
                raise RuntimeError(f"Failed to download {url} after 3 trials") from e
```

rth (approved):
There are existing tools for this (e.g. https://stackoverflow.com/a/35504626/1791279, https://urllib3.readthedocs.io/en/latest/reference/urllib3.util.html#module-urllib3.util.retry) but this is probably OK as is.

Maybe a bit simpler with a for loop over `range(retry_cnt)`? That would make it obvious that this is never going to be an infinite loop without reading the code logic inside it.

ryanking13:
> There are existing tools for this (e.g. https://stackoverflow.com/a/35504626/1791279, https://urllib3.readthedocs.io/en/latest/reference/urllib3.util.html#module-urllib3.util.retry) but this is probably OK as is.

Yeah, I didn't want to add a new dependency for this simple feature.

Issue #2785, opened 2022-06-24
Package Request: support for pyheif


teonbrooks:
Came across this initial post when looking to see if Pyodide supported pyheif. doesn't look like the original asker made a follow up issue. It would be great if we have pyheif supported within Pyodide. 

> 'heic' and 'webp' are modern web image format, is it possible to add 'pyheif' and 'webp' packages?

Yes, please open a separate issue about it.

_Originally posted by @rth in https://github.com/pyodide/pyodide/issues/1632#issuecomment-856520492_

PR #3161, opened 2022-09-30
PKG add libheif, pyheif, and pillow_heif (decoder only)

ryanking13:
### Description

Close: #2785 

This adds libheif and Python packages that use libheif to support HEIC image format.

Note that this is "decoder" only. libheif use libde265 for decoder and libx265 for encoder, and I only added libde265 in this PR.
I think decoding is more important in our use case, hence this PR.

For those who are interested in "encoding" heic format images (possibly me in the future...), I was able to build libx265 but Pyodide hangs when I tried to use it via pillow_heif. You can see [meta.yaml file of libx265 that I tried here](https://github.com/ryanking13/pyodide/blob/pkg/heif/packages/libx265/meta.yaml).

cc: @teonbrooks

### Checklists

- [x] Add / update tests


packages/pillow_heif/test_pillow_heif.py
```diff
@@ -0,0 +1,47 @@
...
+            assert heif_file.mode == "RGBA"
+            assert len(heif_file.data) == 278784
+            assert heif_file.stride == 1056
+        """
+    )
+
+
+def test_pillow(selenium):
+    selenium.load_package(["Pillow", "pillow_heif"])
+    selenium.run(
```
rth, line 31: `@run_in_pyodide(packages = ["Pillow", "pillow_heif"])` here and above ?

rth (approved):
Thanks @ryanking13 ! We could maybe use `run_in_pyodide` for tests, but otherwise LGTM.

Could you please restart the benchmark CI job to make sure that the failure is unrelated?

ryanking13:
> Thanks @ryanking13 ! We could maybe use `run_in_pyodide` for tests, but otherwise LGTM.
> 
> Could you please restart the benchmark CI job to make sure that the failure is unrelated?

Done. Thanks for the review!

teonbrooks:
thanks @ryanking13!

PR #3162, opened 2022-09-30
Prevent from searching meta.yaml recursively inside extracted package source directory

ryanking13:
### Description

This fixes the behavior of searching meta.yaml recursively inside the extracted package source directory.
It sometimes cause error on packages that contain local meta.yaml file due to conda support.


hoodmane (approved):
Makes sense. readthedocs failure seems unrelated?

ryanking13:
> Makes sense. readthedocs failure seems unrelated?

It doesn't seem related. Thanks for the review!

PR #3163, opened 2022-09-30
Add roundtrip parameter to create_proxy

hoodmane:
A typical annoying problem is as follows: if a PyProxy is to be assigned to a JavaScript object from Python:
```py
from js import o
o.f = create_proxy(f)
```
to destroy the proxy it is necessary to do something like:
```py
from pyodide.code import run_js
run_js("(o) => o.f.destroy()")(o)
```
This is because if we say `o.f` in Python we get the unwrapped Python object f rather than the wrapped `PyProxy`, so `o.f.destroy()` will be an `AttributeError`. This PR fixes the problem by setting an internal attribute on the `PyProxy` indicating that when converted to JavaScript it should be wrapped in a `JsProxy` rather than being unwrapped into a normal Python object. This makes it possible to destroy the proxy from Python with `o.f.destroy()`.

### Checklists

- [x] Add / update tests

### DIFF
--- a/docs/project/changelog.md
+++ b/docs/project/changelog.md
@@ -64,6 +64,12 @@ substitutions:
   receive `this` as the first argument if set to `True`
   {pr}`3103`, {pr}`3145`
 
+- {{ Enhancement }} `create_proxy` now takes an optional `roundtrip` parameter.
+  If this is set to `True`, then when the proxy is converted back to Python, it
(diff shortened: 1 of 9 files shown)


ryanking13 (approved):
Thanks, @hoodmane! The roundtrip parameter looks very convenient. Generally LGTM. Please update the docs of `create_proxy`.

PR #3168, opened 2022-10-05
Correctly specify files to be packed in a npm package

ryanking13:
I found that `pyodide_tblib-1.7.1-py3-none-any.whl` is vendored in pyodide npm package. This fixes it by specifying files to be packed one by one.

I also added dry run option for the npm release script. It will help us to check which files are vendored before releasing npm package.

### DIFF
--- a/src/js/package.json
+++ b/src/js/package.json
@@ -61,7 +61,12 @@
     "./repodata.json": "./repodata.json"
   },
   "files": [
-    "pyodide*",
+    "pyodide.asm.js",
+    "pyodide.asm.wasm",
+    "pyodide_py.tar",
+    "pyodide.mjs",
+    "pyodide.js.map",
+    "pyodide.mjs.map",
     "repodata.json",
     "console.html"
   ],
--- a/src/js/package.json
+++ b/src/js/package.json
@@ -63,10 +63,12 @@
   "files": [
     "pyodide.asm.js",
     "pyodide.asm.wasm",
+    "pyodide.asm.data",
     "pyodide_py.tar",
     "pyodide.mjs",
     "pyodide.js.map",
     "pyodide.mjs.map",
+    "pyodide.d.ts",
     "repodata.json",
     "console.html"
   ],
--- a/tools/deploy_to_npm.sh
+++ b/tools/deploy_to_npm.sh
@@ -4,14 +4,21 @@ set -e
 
 echo "//registry.npmjs.org/:_authToken=${NPM_TOKEN}" > ~/.npmrc
 
+SCRIPT_DIR=$( cd -- "$( dirname -- "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )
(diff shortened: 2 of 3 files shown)


src/js/package.json
```diff
@@ -61,7 +61,12 @@
     "./repodata.json": "./repodata.json"
   },
   "files": [
-    "pyodide*",
+    "pyodide.asm.js",
+    "pyodide.asm.wasm",
+    "pyodide_py.tar",
+    "pyodide.mjs",
+    "pyodide.js.map",
+    "pyodide.mjs.map",
```
rth, line 69: Could you confirm that these are only files included in the latest release (aside from those that were added by error as you mentioned)?

tools/deploy_to_npm.sh
```diff
@@ -4,14 +4,21 @@ set -e
...
 # FIXME: since we release from dist directory, README file needs to be copied before release
 cp src/js/README.md dist/
+cp src/js/package.json dist/
 
 cd dist/
 
 PACKAGE_NAME=$(node -p "require('./package.json').name")
 JS_VERSION=$(node -p "require('./package.json').version")
-if [[ ${JS_VERSION} =~ [alpha|beta|rc|dev] ]]; then
+if [[ -n "${DRY_RUN}" ]]; then
```
rth, line 18: So how would one use it? Does it need to be specified in CI somewhere?

ryanking13: I used this manually for this PR `DRY_RUN=1 ./tools/deploy_to_npm.sh`, but I think we should add this to CI so we can check it everytime.

ryanking13:
> In the wheels we distribute, it would also be useful to have some sanity checks. For instance, I think distutils used to ship large .exe files and we took a while to detect and remove those. Maybe some of these checks could be done as part of auditwheel-emscripten.

That sounds like a nice idea. Let's discuss it in a seperate issue.

src/js/package.json
ryanking13:
```
console.html
distutils.tar
package.json
pyodide.asm.data
pyodide.asm.js
pyodide.asm.wasm
pyodide.d.ts
pyodide.js
pyodide.js.map
pyodide.mjs
pyodide.mjs.map
pyodide_py.tar
pyodide_tblib-1.7.1-py3-none-any.whl
README.md
repodata.json
```

These files are included in the 0.21.3 npm distribution. distutils.tar is removed after we changed the default behavior of loading fullstdlib (https://github.com/pyodide/pyodide/pull/2998). But I think we can think of including stdlibs in npm distribution (and also micropip and its dependents as suggested in https://github.com/pyodide/pyodide-webpack-plugin/pull/5#issuecomment-1258892445).

tools/deploy_to_npm.sh
ryanking13: I added this to `test-js` CI job. How does it look now?

PR #3175, opened 2022-10-11
Add `pyodide skeleton` command

ryanking13:
### Description

This adds a `pyodide package` CLI command which replaces `pyodide-build mkpkg`. This is a split up from #2904.

It has two subcommands:

- pyodide package new
- pyodide package update

which corresponds to `mkpkg <pkg>` and `mkpkg <pkg> --update`.

_Update:_

`pyodide package` => `pyodide skeleton`.

- pyodide skeleton pypi <pkg>
- pyodide skeleton pypi <pkg> --update

### Checklists

- [x] Add / update tests
- ~~[ ] Add new / update outdated documentation~~


ryanking13:
I think we should update the docs after implementing all other build commands, in order not to have `pyodide ...` and `pyodide-build ...` mixed in the document.

rth:
Thanks @ryanking13 ! I think, naming wise `pyodide package` it's not very clear about what it does. Maybe `pyodide package new` would have been clear , but then again it's longer and "creating a package" can mean a lot of different things (create the recipe, create the wheel etc).

Maybe, `pyodide recipe init` (or `new`) / `pyodide recipe update` and have two separate subcommand instead of the `--update` option  ?

Alternatively with conda, [an identical command](https://docs.conda.io/projects/conda-build/en/latest/user-guide/tutorials/build-pkgs-skeleton.html#building-a-simple-package-with-conda-skeleton-pypi) is called, `conda skeleton pypi`. So maybe we can just copy that, to avoid having to come up with original names. For now, the only supported backend would be pypi, but it's not too implausible that we could copy the conda-forge or emscripten-forge recipe to create one for Pyodide one day.

pyodide-build/pyodide_build/tests/test_cli.py
```diff
@@ -0,0 +1,51 @@
...
+from typer.testing import CliRunner  # type: ignore[import]
+
+from pyodide_build.cli import new_package
+
+runner = CliRunner()
+
+
+@pytest.fixture
+def packages_temp_path(tmp_path):
+    yield tmp_path
```
rth, line 11: Can't we just use tmp_path then?

pyodide-build/pyodide_build/cli/new_package.py
```diff
@@ -0,0 +1,87 @@
+from pathlib import Path
```
rth, line 1: Same here the file name could be `new_recipe.py` or something similar. That's what conda call these .yaml files.

rth (commented):
> Maybe, pyodide recipe init (or new) / pyodide recipe update and have two separate subcommand instead of the --update option ?

OK after looking at the code, that's already what you implemented except with the `pyodide package` name. My previous comments about the name still apply.

Aside from the name, code wise this LGTM. Thanks!

ryanking13:
> Maybe, `pyodide recipe init` (or `new`) / `pyodide recipe update` and have two separate subcommand instead of the `--update` option ?
> 
> Alternatively with conda, [an identical command](https://docs.conda.io/projects/conda-build/en/latest/user-guide/tutorials/build-pkgs-skeleton.html#building-a-simple-package-with-conda-skeleton-pypi) is called, `conda skeleton pypi`. So maybe we can just copy that, to avoid having to come up with original names. For now, the only supported backend would be pypi, but it's not too implausible that we could copy the conda-forge or emscripten-forge recipe to create one for Pyodide one day.

`recipe` or `skeleton` sounds much better to me. Thanks for the suggestion! Maybe then `pyodide skeleton pypi <pkg>`? I'm not sure how we should name a subcommand for package updates then.

pyodide-build/pyodide_build/tests/test_cli.py
ryanking13: Good point. I think I added that to use the same tmp_path for multiple tests but I only write a single test so we don't need that fixture now.

rth:
> Maybe then pyodide skeleton pypi <pkg>

Hmm, can we put it back as an option `--update`?

ryanking13:
Okay, I changed the entrypoint to `pyodide skeleton`:

- pyodide skeleton pypi <pkg>
- pyodide skeleton pypi <pkg> --update
- pyodide skeleton pypi <pkg> --update-patched

rth:
Thanks @ryanking13 !

PR #3178, opened 2022-10-11
Fix repr_shorten and add shorten

oow214:
### Description
Added example for repr_shorten in issue #1955.
When I first saw that function, it was difficult to understand how to use it. I wrote what I could understand. It would be great if other people could help with this example as well.

Fixed bugs found when adding examples.
Added `shorten` function.

It's my first PR so I'm not sure if it's ok. Please tell me if I did anything wrong. I am ready to edit at any time. Based on this PR, I want to continue to make other contributions.

### Checklists
- [x] Fix `split` argument bug of repr_shorten and add shorten function

### DIFF
--- a/docs/project/changelog.md
+++ b/docs/project/changelog.md
@@ -95,6 +95,9 @@ substitutions:
 - {{ Fix }} Shared libraries with version suffix are now handled correctly.
   {pr}`3154`
 
+- {{ Fix }} Fixed bug in `split` argument of `repr_shorten`. Added `shorten` function.
+  {pr}`3178`
+
 ### Build System / Package Loading
 
 - New packages: pycryptodomex {pr}`2966`, pycryptodome {pr}`2965`,
--- a/docs/project/changelog.md
+++ b/docs/project/changelog.md
@@ -95,7 +95,7 @@ substitutions:
 - {{ Fix }} Shared libraries with version suffix are now handled correctly.
   {pr}`3154`
 
-- {{ Fix }} Fixed bug in `split` argument of `repr_shorten`. Added `shorten` function.
+- {{ Fix }} Fixed bug in `split` argument of {any}`repr_shorten`. Added {any}`shorten` function.
   {pr}`3178`
 
 ### Build System / Package Loading
--- a/src/py/pyodide/console.py
+++ b/src/py/pyodide/console.py
@@ -486,7 +486,7 @@ def shorten(
         split = limit // 2
     if split > len(text) / 2:
         split = int(len(text) / 2)
-    if len(text) - 2 > limit:
+    if len(text) > limit:
         text = f"{text[:split]}{separator}{text[-split:]}"
     return text
 
@@ -507,26 +507,24 @@ def repr_shorten(
     >>> from pyodide.console import repr_shorten
     >>> sep = "_"
     >>> for i in range(6):
-    ...     repr_shorten("1234", limit=i, separator=sep)
-    "_'1234'"
-    "_'1234'"
+    ...     repr_shorten("abcd", limit=i, separator=sep)
+    "_'abcd'"
+    "_'abcd'"
     "'_'"
     "'_'"
-    "'1_4'"
-    "'1234'"
+    "'a_d'"
+    "'a_d'"
     >>> for i in range(6):
-    ...     repr_shorten("1234", limit=i, split=i, separator=sep)
-    "_'1234'"
-    "'_'"
-    "'1_4'"
-    "'12_34'"
-    "'12_34'"
-    "'1234'"
-    >>> len("0123456789")
-    10
-    >>> repr_shorten("0123456789", limit=1, split=4, separator=sep)
-    "'012_789'"
-    >>> repr_shorten("0123456789", limit=15, split=4, separator=sep)
-    "'0123456789'"
+    ...     repr_shorten(1234, limit=i, split=i, separator=sep)
+    '_1234'
+    '1_4'
+    '12_34'
+    '12_34'
+    '1234'
+    '1234'
+    >>> repr_shorten(123456789, limit=1, split=4, separator=sep)
+    '1234_6789'
+    >>> repr_shorten(123456789, limit=15, split=4, separator=sep)
+    '123456789'
     """
     return shorten(repr(value), limit=limit, split=split, separator=separator)
--- a/src/py/pyodide/console.py
+++ b/src/py/pyodide/console.py
@@ -492,25 +492,22 @@ def repr_shorten(
     Examples
     --------
     >>> from pyodide.console import repr_shorten
-    >>> sep = ".."
+    >>> sep = "_"
     >>> for i in range(7):
     ...   repr_shorten("1234", limit=i, separator=sep)
-    "..'1234'"
-    "..'1234'"
-    "'..'"
-    "'..'"
-    "'1..4'"
-    "'1..4'"
+    "_'1234'"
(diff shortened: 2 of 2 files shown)


ryanking13:
Thanks for the PR @oow214! I think we should limit the maximum value of `split` argument to the half of original string so that things like this should not happen.

```py
>>> repr_shorten("5678", split=6, separator=sep)
5678..5678
```

oow214:
I didn't know it was a bug. I changed the example.

ryanking13:
> I didn't know it was a bug. I changed the example.

Thanks, are you interested in fixing the bug in this PR? (In that case, the title of the PR needs to be changed)

I think the fix would be limiting the value of `split` to less than or equal to `limit // 2` so that `repr_shorten` should always return a shortened string.

hoodmane:
Probably should just add `split = min(split, limit//2)`.

src/py/pyodide/console.py
```diff
@@ -488,10 +488,40 @@ def repr_shorten(
...
+    >>> repr_shorten("0123456789", limit=15, split=4, separator=sep)
+    "'0123456789'"
     """
     if split is None:
         split = limit // 2
     text = repr(value)
-    if len(text) > limit:
+    if split > len(text) / 2:
+        split = int(len(text) / 2)
+    if len(text) - 2 > limit:
```
oow214, line 525: > If it is longer than ``limit`` then return the firsts ``split``
> characters and the last ``split`` characters separated.

The correct repr_shorten output as I understand it is:
```python
>>> for i in range(10):
...     repr_shorten("12345678", limit=i, split=i, separator=sep)

"linit:0 | split:0 | _'12345678'"
"linit:1 | split:1 | '_'"
"linit:2 | split:2 | '1_8'"
"linit:3 | split:3 | '12_78'"
"linit:4 | split:4 | '123_678'"
"linit:5 | split:5 | '1234_5678'"
"linit:6 | split:5 | '1234_5678'"
"linit:7 | split:5 | '1234_5678'"
"linit:8 | split:5 | '12345678'"
"linit:9 | split:5 | '12345678'"
```

So I thought I needed to fix two things.
1. Shouldn't this be fixed because using `repr` includes `''` and adds 2 to the length of the string? `if len(text) - 2 > limit:`
2. Shouldn't `split' just compare if it's less than half the length of the original string and change the value if it's greater?

Can you please confirm what I'm misunderstanding?

hoodmane: 1. I think it is supposed to apply repr and then shorten the result. Applying repr to a string adds quotes. It would probably be a better API to require `value` to be a `str` and only do the shortening and then call it like `repr_shorten(repr(v))`. Currently that would be an API break. Maybe add a separate `shorten` function and redefine `repr_shorten` to do `shorten(repr(v), other args)`. That way if you already have a string that you want to shorten it won't gain quotes.
2. Yeah something like that sounds good.

Also, I think if limit is very small we should throw a value error (does it make any sense for limit to be 0 or 1?)
Also, regardless of what we do to limit the value of `split` we should make sure we don't see cases like you show where the separator is added but no characters are removed. Maybe something like:
```py
if len(shortened_string) >= len(original_string):
   return original_string
```
would be good.

src/py/pyodide/console.py
```diff
@@ -479,6 +479,20 @@ async def runcode(self, source: str, code: CodeRunner) -> ConsoleFuture:
...
 
 
+def shorten(
+    text: str, limit: int = 1000, split: int | None = None, separator: str = "..."
+) -> str:
+    if limit < 2:
+        raise ValueError("limit must be greater than or equal to 4.")
+    if split is None:
+        split = limit // 2
+    if split > len(text) / 2:
```
ryanking13, line 489: Maybe:

```suggestion
    split = min(split, len(text) // 2)
```

ryanking13 (approved):
Thanks for your work, @oow214! I have a minor comment, otherwise LGTM. Could you please update the [changelog](https://github.com/pyodide/pyodide/blob/main/docs/project/changelog.md)?

The CI failure is not related to this PR (It is fixed in #3185). So please merge upstream main branch to this PR.

src/py/pyodide/console.py
oow214: I fixed it and updated the changelog as well. Thank you so much for kindly letting me know! Is there anything else I should do?

oow214: I fixed it and updated the changelog as well. Thank you so much for kindly letting me know!

docs/project/changelog.md
```diff
@@ -95,6 +95,9 @@ substitutions:
 - {{ Fix }} Shared libraries with version suffix are now handled correctly.
   {pr}`3154`
 
+- {{ Fix }} Fixed bug in `split` argument of `repr_shorten`. Added `shorten` function.
```
hoodmane, line 98:
```suggestion
- {{ Fix }} Fixed bug in `split` argument of {any}`repr_shorten`. Added {any}`shorten` function.
```

src/py/pyodide/console.py
```diff
@@ -479,19 +479,66 @@ async def runcode(self, source: str, code: CodeRunner) -> ConsoleFuture:
...
 ) -> str:
-    """Compute the string representation of ``value`` and shorten it
-    if necessary.
+    """Shorten ``text`` it if necessary.
 
     If it is longer than ``limit`` then return the firsts ``split``
     characters and the last ``split`` characters separated by '...'.
     Default value for ``split`` is `limit // 2`.
+    If ``split`` is longer than the length of ``len(value) // 2``, ``split`` will be `len(value) // 2`.
+    limit must be greater than or equal to 2. If it is less than 2, an error occurs.
```
hoodmane, line 491: Some edits:
```suggestion
    """Shorten ``text`` if it is longer than ``limit``.

	If ``len(text) <= limit`` then return ``text`` unchanged.
    If ``text`` is longer than ``limit`` then return the firsts ``split``
    characters and the last ``split`` characters separated by ``separator``.
    The default value for ``split`` is `limit // 2`.
    Values of ``split`` larger than ``len(value) // 2`` will have the same effect as 
    when ``split`` is `len(value) // 2`.
    A value error is raised if ``limit`` is less than 2.
```
Perhaps we should add `Parameters` and `Returns` sections like the other functions in this file?

src/py/pyodide/console.py
```diff
@@ -479,19 +479,66 @@ async def runcode(self, source: str, code: CodeRunner) -> ConsoleFuture:
...
+) -> str:
+    """Compute the string representation of ``value`` and shorten it
+    if necessary.
+
+    If it is longer than ``limit`` then return the firsts ``split``
+    characters and the last ``split`` characters separated by '...'.
+    Default value for ``split`` is `limit // 2`.
+    If ``split`` is longer than the length of ``len(value) // 2``, ``split`` will be `len(value) // 2`.
+    limit must be greater than or equal to 4. If it is less than 4, an error occurs.
+
```
hoodmane, line 523:
```suggestion
	This is equivalent to `shorten(repr(value), limit, split, separator)`.
```

hoodmane (approved):
Looks good to me. Thanks for working on this @oow214! I made some suggestions for the documentation strings. I'm +1 to merge it with these changes.

src/py/pyodide/console.py
oow214: looks so good! I've fixed them all. Thank you. 😊

hoodmane:
Thanks again @oow214!

PR #3185, opened 2022-10-14
Bump zlib version to 1.2.13

ryanking13:
It looks like zlib yanked the v1.2.12 due to [CVE-2022-37434](https://nvd.nist.gov/vuln/detail/CVE-2022-37434).


ryanking13:
Merging since this is causing failure in other PRs.

PR #3187, opened 2022-10-17
MAINT Split out package loading and dynlib loading code

ryanking13:
### Description

This splits out `loadDynlib` function from `load-package.ts` into a new file `dynload.ts`. It just reorganizes code, no functional change. 

This is a preparation for #3183, which will (possibly) add some helper functions on sharedlib loading.


rth (approved):
Thanks LGTM, but we should probably fix CI before merging.

PR #3188, opened 2022-10-17
CI Ignore if codecov fails

ryanking13:
I wasn't able to find a way to fix codecov 404 error that happens occasionally... This PR just makes CI pass even if codecov fails.


rth (approved):
Thanks!

PR #3192, opened 2022-10-20
CI Build and test all packages with GHA

ryanking13:
### Description

This makes GHA build and test all packages on every push.

Since we use conda env in GHA, this will help us make sure that we can build all packages outside of the Docker image.

Note that GHA has advantages and disadvantages compared to CircleCI:

- GHA allows each job to run to maximum 6 hours. So we don't need to care about job timeout.
- GHA has 2-core / 7GB RAM, compared to 4-core / 8GB RAM of CircleCI large resource class. So building packages in GHA is much slower (though I think there are some rooms to optimize this by parallelizing build)
- Passing data between jobs is slower in GHA. It is because `upload_artifacts` action in GHA chunks data into 4MB while uploading.

So I think GHA is not a replacement for CircleCI, but we can still use it for the auxiliary CI framework for testing conda env-based builds (at least until it is free of charge :)).

Further plans that I haven't included in this PR:

- Making GHA run faster
  - caching conda env
  - parallelizing package build / test
- Running selenium tests
- MacOS build
- Things that are mentioned in #2280


rth:
Thanks for working on this!
My concern is that running build CI for all packages for all commits is already very wasteful in terms of compute resources. Doing that for 2 CI providers would be even worse. 
So making sure that each package can build outside of docker is certainly useful information (and we should definitely integrate your changes), but maybe we could keep a manual trigger on this CI job? Although then there is a risk it would get outdated if never run.

jobovy:
> Thanks for working on this! My concern is that running build CI for all packages for all commits is already very wasteful in terms of compute resources. Doing that for 2 CI providers would be even worse. So making sure that each package can build outside of docker is certainly useful information (and we should definitely integrate your changes), but maybe we could keep a manual trigger on this CI job? Although then there is a risk it would get outdated if never run.

One option could be to only build and test packages on GHA that have changes in their package directory and then just run everything using a cron job every week or so to regularly make sure that all packages can be built. I've been using https://github.com/jitterbit/get-changed-files to do things like this and it works well.

Also, to speed-up and cache the conda environment, you may be able to use `micromamba` through https://github.com/mamba-org/provision-with-micromamba.

One big advantage of using GHA is that people can then easily run the workflows on their own forks. So combined with only building/testing packages with changes, that would allow people to much more easily make sure that a new or updated package works before opening a PR that always runs all the builds/tests.

ryanking13:
> My concern is that running build CI for all packages for all commits is already very wasteful in terms of compute resources. Doing that for 2 CI providers would be even worse.

Yes, I agree that it is quite wasteful to run this every time. How about running this only in the main branch (+ manual run) after merge but not on pull requests?

> One option could be to only build and test packages on GHA that have changes in their package directory

Yes, that is our future goal (#1262 #2179), but the changes in other places (pyodide-build, emscripten, ...) might break the build so it is not very straightforward to implement it.

> Also, to speed-up and cache the conda environment, you may be able to use micromamba through https://github.com/mamba-org/provision-with-micromamba.

Thanks for the information!

ryanking13:
I changed it to run only in the main branch after merged or by manually. If you think it still runs too often, we can instead use a cron job as suggested by @jobovy.

WDYT?

rth:
Firefox tests seem to be very slow with playwright (up to 1h for core tests), would it make sense to disable it? I mean we are already testing various browsers in CircleCI this is only to check that the build in conda is OK. Keeping the parametrized matrix config, in case we want to change in the future is useful.

Otherwise, at an estimated 1 commit per day on master, and 3 hours per run, on a 2 core CPU that's 2k CPU hours per year (or roughly  54$ in cloud cost). Probably acceptable. We can always disable it later if it takes too long to run. Otherwise yes, a cron job could work too indeed.

ryanking13:
> Firefox tests seem to be very slow with playwright (up to 1h for core tests), would it make sense to disable it? I mean we are already testing various browsers in CircleCI this is only to check that the build in conda is OK. Keeping the parametrized matrix config, in case we want to change in the future is useful.

Right. Playwright Firefox tests take too long and let's disable it for now.

> Otherwise, at an estimated 1 commit per day on master, and 3 hours per run, on a 2 core CPU that's 2k CPU hours per year (or roughly 54$ in cloud cost). Probably acceptable. We can always disable it later if it takes too long to run. Otherwise yes, a cron job could work too indeed.

Yes, it would be easy to switch to cron jobs anytime.

rth:
Do we still want this @ryanking13 ? In the end I'm not sure there is a problem that needs solving by this, especially if means waiting more on PRs. I would say we should go in the direction of building less rather than building more at each commit. Or switch to cron already, but then in practice to we want to have more CI to fix than now ^^

ryanking13:
> I would say we should go in the direction of building less rather than building more at each commit.

Right, I agree. I think we can close this for now. I'll open a separate PR for issues that I've found while I was working on this PR.

PR #3193, opened 2022-10-20
Fix undefined symbol error when loading shared lib

ryanking13:
The `allowUndefined` parameter passed to `loadDynamicLibrary` was removed by mistake in #2954.

Note that `allowUndefined` needs to be set to true because some shared libs have undefined symbols that are defined in another shared library. For example, `openssl` has two sharedlibs: `libssl`, `libcrypto` and `libssl` requires some symbols defined in `libcrypto`.


ryanking13:
Merging since this is causing errors in other PRs: #3178, #3192

PR #3196, opened 2022-10-21
build packages from pypi direct

joemarshall:
This lets `pyodide build` create a package from pypi directly into a dist folder.

If there's a pure python wheel, it just downloads it. Otherwise it downloads the sourcedist and builds it using the pyodide build environment.

Use it like:

`pyodide build --package <PACKAGESPEC>`

where <PACKAGESPEC> is a package plus optional version stuff.

It doesn't do dependencies yet. Right now it downloads using unearth. It could also be modified to use micropip to do everything, but I think micropip needs some refactoring to actually work full in a cross-compilation environment.


pyodide-build/pyodide_build/cli/build_oot.py
```diff
@@ -10,13 +10,17 @@ def main(
...
+        help="pypi package-spec to download and build",
+    ),
     ctx: typer.Context = typer.Context,
 ) -> None:
     """Use pypa/build to build a Python package"""
     initialize_pyodide_root()
     common.check_emscripten_version()
     backend_flags = ctx.args
-    build.run(exports, backend_flags)
+    build.run(exports, package, backend_flags)
```
hoodmane, line 23: I think we should have a different subcommand for this. I would like to keep `pyodide build` as a subset of [pypa/build](https://pypa-build.readthedocs.io/en/latest/#python--m-build)'s API. In `pypa/build` the positional argument is a path to the local directory we want to build which I think is more natural. 

Also in favor of using a different time, this current command will just download a wheel from pypi if there is a pure Python wheel, so in that case nothing is built at all and the name is a misnomer.

hoodmane (commented):
Thanks @joemarshall, this is nice.

pyodide-build/pyodide_build/out_of_tree/build.py
```diff
@@ -1,15 +1,61 @@
...
+from pathlib import Path
+from tempfile import TemporaryDirectory
+
+from unearth.finder import PackageFinder, TargetPython
 
 from .. import common, pypabuild, pywasmcross
 
 
-def run(exports, args):
+def fetch_pypi_package(package_spec, destdir):
```
ryanking13, line 11: `unearth` looks great. In `mkpkg.py`, there is also a function that downloads from PyPI, so it would be nice if we can merge it to one. (No action required for this PR).

joemarshall:
Okay, I changed it so that it is `pyodide fetch_wheel` to get a pypi wheel.

joemarshall:
There's some tests failed on circleci - I'm not sure why; I think it may just need re-running the tests (which I can't do I think)

ryanking13:
> There's some tests failed on circleci - I'm not sure why; I think it may just need re-running the tests (which I can't do I think)

This has been fixed upstream, so please merge the upstream main branch.

> Okay, I changed it so that it is pyodide fetch_wheel to get a pypi wheel.

I think the name `fetch_wheel` isn't very clear about what it does, and I'm not sure we want it for a top-level command as what it does is mostly similar to `pyodide build` except for downloading packages from PyPI.

I am +1 with Hood's idea of " keep pyodide build as a subset of [pypa/build](https://pypa-build.readthedocs.io/en/latest/#python--m-build)'s API" though. So maybe we can have a different subcommand group for this. This should also be discussed at https://github.com/pyodide/pyodide/pull/3216 too.

joemarshall:
Hmm. How about pyodide package_build, with --meta for meta file build and --pypi for pypi build?

The alternative would be to add two very obviously pyodide specific arguments to pyodide build, which would break the idea that it was a pure subset of pypa build, but wouldn't break argument passing etc. 

So you'd have e.g. pyodide build args... which would pass everything through as normal, 
pyodide build --from-pypi name args, 
 pyodide build --from-pyodide-meta

joemarshall:
I've updated this so that it has

| command | result |
|-------------|-------|
| `pyodide package-build recipe` | build packages from recipe folder |
| `pyodide package-build pypi` | build or fetch a single package from pypi |
| `pyodide package-build source` | build the current source folder (same as pyodide build) |
| `pyodide package-build url` | build or fetch a package from a url either tgz, tar.gz zip or wheel |

joemarshall:
I think this is pretty close now. Any feedback welcome.

hoodmane:
I'll try to look at it later today. Thanks again for working on this @joemarshall!

hoodmane (approved):
Looks good to me. Probably @ryanking13 should review this too. Thanks @joemarshall!

run_docker
```diff
@@ -137,6 +137,7 @@ CONTAINER=$(\
         --groups sudo  \
         $USER_NAME \
       ; \
+      echo 'export PATH=\$PATH:$USER_HOME/.local/bin' >> /etc/profile; \
```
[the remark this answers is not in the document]
ryanking13: When we do `pip install ./pyodide-build` inside the docker image, executables are installed into `~/.local/bin`. But it is not in the PATH, so it cannot be used unless we modify the PATH manually. I just added this for my quality of life.

ryanking13:
Thanks for the porting work @joemarshall. I'll try to review this in a few days.

rth:
Thanks a lot for doing this work @joemarshall, and sorry for not reviewing earlier. 

Code wise I'll trust existing reviews, and it's good to have separate functions for these different use cases. I'm only wondering if we could group some of these endpoints together. For instance, `pypi`, `source`, `url` could very well be a single endpoint that determines from the passed argument which type it is, can't it?  

> I am +1 with Hood's idea of " keep pyodide build as a subset of [pypa/build](https://pypa-build.readthedocs.io/en/latest/#python--m-build)'s API"

I see your point, but for users, it's going to be very confusing to have multiple commands for a build that do the same thing (pyodiode build == pyodide package-build source). What would be the downside of keeping the pypa/build API but extending the `srcdir` to accept URL and PyPI package,
```
pyodide build <path>
pyodide build <url>
pyodide build pypi:package-name    # maybe with a prefix to make it easier to detect
```
?
The only extra thing they do externally is to fetch the source from somewhere, I don't imagine there would be many extra features needed
And then the separate endpoint for in-tree build
```
pyodide build-recipe <path>
```
since its API is rather different.

Sorry, I know if implemented this would revert some of the changes you did earlier, but I think it's important we all agree on the API since we can't then change it easily. My general argument is that the most popular CLI auto-detects the argument, rather than have separate endpoints for each type and it's better for usability (`pip install`, `pd.read_csv (path or url)`, etc). 

Or else re-evaluate why we need `pyodide build` if there is a `pyodide package-build source` (but of course, the former is a much nicer name).  I would say from a user perspective people don't care whether it follows or doesn't the `pypa/build` API, users just want one easy-to-remember command to build their wheel.

hoodmane:
My main concern is that we might use up API space that is in the future used in a different way by `pypa/build` and then run into compatibility problems. I think there's a huge benefit to having tools that are as close as possible to drop in replacements for the non wasm equivalent. Otherwise we might end up having to maintain extra workarounds for the places where the tools diverge. 

But as long as there's some compatible option I don't care that much what we name it. It's sensible to use the shortest name for the tool that's friendliest to human users. So we could take over the name `pyodide build` for this and reserve a different command for a subset of `pypa/build`.

rth:
> reserve a different command for a subset of pypa/build.

OK but in practice what's the benefit for users of exposing it, if we already expose a different command that does the same thing with a more extensive API?  My concern is that then some blogs/tutorials out there will use one version, others will use the other one and it will just create more confusion about how to use Pyodide.

> My main concern is that we might use up API space that is in the future used in a different way by pypa/build and then run into compatibility problems.
>  I think there's a huge benefit to having tools that are as close as possible to drop in replacements for the non wasm equivalent. 

Certainly about compatibility. So if we say we have a command that is a superset of `pypa/build` that does exactly the same thing as it, but with a few extra features wouldn't it be better than having two commands? If `pypa/build` changes its behavior we will change accordingly. It doesn't really stop us from adding one or two extra features on top, or even optional CLI arguments if need be. 

The reason for compatibility is that people familiar with pypa/build would be able to use `pyodide build` without changing anything, which will remain true even if we add extra features to it. Unless I'm missing something.

hoodmane:
>  people familiar with pypa/build would be able to use pyodide build without changing anything

Well I'm really more interested in tools that are built on top of `pypa/build`. Like for instance, it would be nice to have a tool like `pip` that automatically cross builds packages. If the existing tool uses `pypa/build` and we have a nearly drop-in replacement for it available, then that minimizes the number of changes that need to be made and confusing incompatibilities to debug. It also makes it easier for people writing a new tool that is intended to work both inside and outside of Pyodide. For human users / writing new Pyodide-only tools you could use the more convenient API.

> if we say we have a command that is a superset of `pypa/build` that does exactly the same thing as it, but with a few extra features wouldn't it be better than having two commands?

I guess my worry is that later `pypa/build` might add a feature that uses the same syntax space but to mean a different thing. Then to maintain "it's a superset of pypa/build" we have to break backwards compatibility.

You have good points and I am not completely sure about this.

rth:
> If the existing tool uses pypa/build and we have a nearly drop-in replacement for it available, then that minimizes the number of changes that need to be made and confusing incompatibilities to debug.
> [...]
> I guess my worry is that later pypa/build might add a feature that uses the same syntax space but to mean a different thing. Then to maintain "it's a superset of pypa/build" we have to break backward compatibility.

I see your concerns. OK how about we start with a single build command that is a superset of features, and if for any reason it doesn't work as drop-in replacement (which should be our goal) or they change their API significantly, we can decide then to either,
 - expose this second CLI with strict subset API
 - or break backward compatibility to make it work
 
So we could address this if (or when) the concern materializes while optimizing for an easy-to-remember and use API for now? Because it's also possible that it won't be an issue.

Sorry, I should have commented earlier about this :) Any preference from your side on this CLI naming subject @joemarshall @ryanking13 ?

joemarshall:
pip will work from pypi / url / source just fine won't it, I think the logic of that is:

1) If it is a url (or potentially a vcs link), fetch that.
2) If it is a name without version identifier etc. then check if it is a valid folder or wheel name.
3) Otherwise, assume it is a pip versionspec

In pip there is actually a detail which is that you can specify extras for case 1 and 2 by putting quote marks around the url or folder name and then adding [extranames] outside the quote marks. This doesn't matter here because we don't build deps (and extras only affect deps).

So it would be pretty straightforward to make it so that we have just:
a) `pyodide build <url | folder | versionspec>` which without an argument just builds the current folder as it should. 
b) `pyodide build-recipes <params>` which builds a full recipes folder.

I think incidentally this is probably another PR at some point, but we should add a way to build *just* a single meta file, which would work nicely here, because it is a pain not being able to build a single package based metafile without dependencies. 
e.g. `pyodide build <url | folder | versionspec | wheelfile | metafile>`

Also, another possible PR, it would be nice to have it work with a requirements.txt (obviously you can script that pretty straightforwardly, but they're a thing, would be worth supporting them).

joemarshall:
It's done, as above, `pyodide build [package_spec|url|source folder]` works now. `pyodide build-recipes` does the metafile recipe folder build.

Seems to work okay. I think tests need fixing still though.

docs/project/changelog.md
```diff
@@ -106,6 +106,15 @@ substitutions:
   `pyodide-build mkpkg` will be replaced by `pyodide sekeleton pypi`.
   {pr}`3175`
 
+- Added a new CLI command `pyodide package-build` which builds packages from various sources.
+  | command | result |
+  |-------------|-------|
+  | `pyodide package-build recipe` | build packages from recipe folder (replaces `pyodide-build buildall`) |
```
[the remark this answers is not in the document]
ryanking13: > Sorry to be annoying about naming: implementation we can always change, but not the command name, so we might as well spend some time on it to make sure it's optimal )

No worries, I think we all agree that naming is really important.

ryanking13:
> I think incidentally this is probably another PR at some point, but we should add a way to build _just_ a single meta file, which would work nicely here, because it is a pain not being able to build a single package based metafile without dependencies.
> e.g. `pyodide build <url | folder | versionspec | wheelfile | metafile>`

Thanks, @joemarshall. The reason I didn't add a command for building a single recipe is that I was thinking of revising the `buildall` and `buildpkg` API and merging them if possible (for now buildpkg + resolving dependencies + creating repodata.json is buildall) someday. Not sure what the API should look like though, for example, roman suggested `--no-deps` flag in #2879.

pyodide-build/pyodide_build/tests/test_cli.py
```diff
@@ -44,3 +50,59 @@ def test_skeleton_pypi(tmp_path):
     )
     assert result.exit_code != 0
     assert "already exists" in str(result.exception)
+
+
+def test_build_recipe(tmp_path, monkeypatch, request):
+    if request.config.getoption("runtime") == "host":
+        pytest.skip(reason="Can't build recipe without building pyodide first")
```
ryanking13, line 57: The reason this test fails is that there is no `_sysconfigdata__emscripten_wasm32-emscripten.py` (See https://github.com/pyodide/pyodide/actions/runs/3373970375/jobs/5599098432).

I think we can mock it during the test. But I think we can do that in a separate PR. So no action is required for this PR.

pyodide-build/pyodide_build/cli/build.py
```diff
@@ -0,0 +1,243 @@
...
+        help="Build source, can be source folder, pypi version specification, or url to a source dist archive or wheel file. If this is blank, it will build the current directory.",
+    ),
+    exports: str = typer.Option(
+        "requested",
+        help="Which symbols should be exported when linking .so files?",
+    ),
+    ctx: typer.Context = typer.Context,
+) -> None:
+    """Use pypa/build to build a Python package from source, pypi or url."""
+    if source_location is None or len(source_location) == 0:
```
ryanking13, line 225: Maybe simply: 
```suggestion
    if not source_location:
```

pyodide-build/pyodide_build/cli/build.py
```diff
@@ -0,0 +1,243 @@
...
+        # build the current folder
+        source(".", exports, ctx)
+    elif source_location.find("://") != -1:
+        url(source_location, exports, ctx)
+    elif Path(source_location).exists():
```
ryanking13, line 230: Maybe: 
```suggestion
    elif Path(source_location).is_dir():
```

ryanking13 (commented):
Thanks, @joemarshall. Currently, we lack some tests for out-of-tree entrypoint, but let's do that in a separate PR, because it might need some tweaks or changes in internal implementation, and I don't want to introduce a new topic in this PR.

> a) pyodide build <url | folder | versionspec> which without an argument just builds the current folder as it should.

I am +1 with this single API + auto-detection design.

joemarshall:
Just had a thought on this - in the long term, does it make sense to only support out of tree package building. 

I.e. so there's a pyodide-core package which builds python and pyodide js shenanigans and makes the build artifacts that are required to run out of tree build, then all the packages are built using out of tree build against those generated build artifacts.

And the repo splits into pyodide and pyodide packages somehow...

rth (commented):
@hoodmane Could you please confirm you are OK with the latest API here? 

A couple minor comments otherwise LGTM, thanks @joemarshall !

rth:
> Just had a thought on this - in the long term, does it make sense to only support out of tree package building.

From what I understand out of tree building is still for moderately straightforward wheels. Building packages with lots of C dependencies and without dependency resolution would be more difficult. 

> And the repo splits into pyodide and pyodide packages somehow...

+1 for that in any case.

joemarshall:
I think I've unbroken that test. It skips tests that won't work in CI because pyodide isn't built when test-python is called.

Currently the tests never get run on CI, because it never tests pyodide_build after pyodide itself is built. 

It probably makes sense to run pyodide_build tests in the test-core tests also. Or run the test-python with pyodide built. 

Actually, thinking about it, in pyodide build tests, we really need two runs of the test also, in and out of tree.

Is the plan to move pyodide_build outside the pyodide repo at some point? Might it be best to wait until then to properly fix these tests?

hoodmane:
You can do what I did for the command line runner tests and take and ignore a selenium fixture then skip if the browser isn't node... We should probably add a mark for post built host tests.

joemarshall:
> You can do what I did for the command line runner tests and take and ignore a selenium fixture then skip if the browser isn't node... We should probably add a mark for post built host tests.

That needs the pyodide_build tests folder to be added into the test-core-node call right? Is that going to break anything bad (other than a load of skipped tests in the log)

joemarshall:
I made the pyodide_build tests run in node now, so you can now make a test like this using the runtime fixture which won't get skipped:

```
def test_fetch_or_build_pypi_with_pyodide(tmp_path, runtime):
    if runtime != "node":
        pytest.xfail("node only")
```

joemarshall:
I think this is ready to go now, seems to pass tests, and I looked into the circleci output and they are being run which is nice.

joemarshall:
codecov doesn't count tests done in test-core-node by the way, that's why that is too low. 

Would it be possible just to move test-python to be after the pyodide build? Or even run two passes, one before and one after and merge the coverage results?

rth:
Normally that should already be the case: different jobs submit partial coverage and it would merge them. Though overall codecov is not very reliable.

hoodmane (approved):
Yeah seems good to me. Thanks for working on this @joemarshall this is a great contribution.

ryanking13:
Thanks @joemarshall!

PR #3197, opened 2022-10-21
NFC Disable codecov annotations in pull requests

hoodmane:
These make it really hard to read the diffs in pull requests.

### DIFF
--- a/codecov.yml
+++ b/codecov.yml
@@ -1,0 +1,2 @@
+github_checks:
+  annotations: false


hoodmane:
In the meantime, you can delete these annoying boxes with the following code in the browser console:
```js
$$(".check-annotation-warning").forEach(e => e.parentElement.remove())
```

ryanking13:
Thanks! I was disabling it manually when it is annoying :)

image

This change will make us less care about coverages... but I agree that it is quite annoying and there are things that are hard to test.

PR #3202, opened 2022-10-24
TST Fix jsexception test

ryanking13:
The host-side JsException class has been changed after: https://github.com/pyodide/pytest-pyodide/pull/44

This PR also pins pytest-pyodide version to prevent CI failure due to pytest-pyodide update.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -28,8 +28,7 @@ jobs:
           name: Test docs
           command: |
             mkdir test-results
-            pip install pytest-pyodide
-            pip install -e ./pyodide-build
+            pip install -r requirements.txt
             npm install -g node-fetch@2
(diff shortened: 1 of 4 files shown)


hoodmane:
How about changing pytest_pyodide to monkey patch JsException? I think this
change looks weird. If we're going to have a class with the same name both
inside and outside selenium, it should have the same path.

rth:
> How about changing pytest_pyodide to monkey patch JsException?

I opened https://github.com/pyodide/pytest-pyodide/issues/53 about it with a somewhat different suggestion.

Though still +1 to merge this as is (thanks @ryanking13), to fix CI then we can look into it with no urgency and revert as  necessary.

ryanking13:
Yes, for now let's merge this for a hotfix, then discuss for a more long term solution in  https://github.com/pyodide/pytest-pyodide/issues/53

PR #3205, opened 2022-10-25
Disable matplotlib benchmark temporarily

ryanking13:
Benchmark CI job is occasionally failing after #3130.

The error seems to happen in `html5_canvas_backend`, but it is hard to analyze. So until we manage to fix it, I think we better disable the matplotlib benchmark...


rth:
Thanks!

Issue #2664, opened 2022-06-05
Add greenlet


rth:
Recently @antocuni mentioned that it could be worth exploring the idea of [green threads](https://en.wikipedia.org/wiki/Green_threads) in relationship with https://github.com/pyodide/pyodide/issues/1503 and https://discuss.python.org/t/wasm-python-synchronous-io-working-group/15509.

The first step would likely be to add the [greenlet](https://greenlet.readthedocs.io/en/latest/). The build setup  is in https://github.com/pyodide/pyodide/commit/7bda00302363f3ea444a2e37a04d782bd95c6a34 however this currently fails with, 
```
src/greenlet/greenlet.c:374:6: error: "greenlet needs to be ported to this platform, 
or taught how to detect your compiler properly."
```
so it looks one should add a WASM platform implementation [here](https://github.com/python-greenlet/greenlet/tree/master/src/greenlet/platform). The implementations for other platforms are not very long, but quite low level.

I don't know much about this topic, but I found it interesting that greenlet is often used in combination with gevent which is an API on top of the [libev](http://software.schmorp.de/pkg/libev.html) (inspired by libevent) or [libuv](http://libuv.org/) event loops. While the last two are event loops used e.g. in Chromium and Node.js respectively.

hoodmane:
Related: [wasm stack switching proposal](https://github.com/WebAssembly/stack-switching/blob/main/proposals/stack-switching/Overview.md)

antocuni:
Some more thoughts about this. My initial idea is that once we have greenlets, it should be easy/possible to simulate synchronous calls from Python. E.g., I can imagine having an implementation of `requests.get` which under the hoods uses a greenlet to stop its execution and yield back the control to the browser. So, it would be sync from the perspective of the python user and async from the browser point of view.

In order to support greenlets, we need some sort of C level stack switching.
I'm new to the webassembly/emscripten world and I'm still trying to wrap my head around it, but from what I understand there are at least three different approaches:
- [asyncify](https://emscripten.org/docs/porting/asyncify.html): if I understand correctly, when using this emscripten rewrites the WASM in order to support stack-switching, and it adds some overhead. Moreover, I fear that it would need rewrite the *whole* CPython in order to work, and probably all the extensions would need to be asyncified as well. But again, I might have misunderstood
- [emscripten WASM stack switching](https://github.com/emscripten-core/emscripten/pull/16779): this looks more promising but it's also pretty new, we need to investigate how well it works
[...]

PR #3210, opened 2022-10-30
Stack switching

hoodmane:
Uses the JS Promise integration stack switching API to allow blocking for JavaScript promises and PyodideFuture objects.

- [x] Update tests
- [x] Merge https://github.com/pyodide/pytest-pyodide/pull/89
- [x] Release a version of pytest-pyodide
- [x] Update changelog


hoodmane:
We can also use this to insert a sleep into the Python main loop, so that the print statements from the Python loop and the Javascript loop are interleaved:
```js
for(let i = 1; i < 100; i++){
  sleep(100*i).then(() => console.log("jsi", i));
}
f = py.runPython(`
def f():
  for i in range(100_000):
    if i % 100 == 0:
      print("pyi", i)
  import pytest # segfaults if pytest hasn't been imported before =(
f
`);
await f.callSyncifying({});
```
So the stack switching is extremely promising. It also seems likely that we will be able to make a version of Pyodide that uses stack switching if it is available but gracefully falls back if it is not available.

ryanking13:
Thanks for the investigation @hoodmane! Indeed it looks quite promising.

> with node v18.x

Nice. Since node v18 is now in an LTS state, I think it is okay to use new features of v18.

hoodmane:
Apparently they completely rearranged this proposal though. It seems like it now requires wasm reference types to use, and they may have removed the `WebAssembly.Suspender` API. Wasm reference types have been available in Firefox since v79 July 2020, in Safari since v15.0 released in September 2021, and in Chrome since v96 released November 2021. So it would be nice not to use them. But also, there is currently no way to make C/C++ code that emits wasm reference types, the only documented way to use them is via assembly code.
https://github.com/emscripten-core/emscripten/issues/15878

Both the backwards compatibility problem and the how-do-we-even-emit-code problem could be solved by hand coding the wrapper like what I did for `createDyncallWrapper` but this will be a pain:
https://github.com/emscripten-core/emscripten/pull/17328/files#diff-8bb085f397f4174fe93a8cf0cda27a12a2647a4afe8ce7116ed740935f5d1188

hoodmane:
I will have to try rewriting this to use the newer approach and see if it fixes the segfaults...

hoodmane:
See:
https://github.com/WebAssembly/js-promise-integration/blob/main/proposals/js-promise-integration/Overview.md

hoodmane:
Okay, so this works great in node v19 with flags `--experimental-wasm-type-reflection` and `--experimental-wasm-stack-switching` or in chrome version 109.0.5384.0 (Official Build) dev with the `Experimental WebAssembly Stack Switching` flag set to enabled. I have two tests: one shows that we can allow background tasks to occur intermingled with synchronous Python. Another uses an import hook to download and install `pytest` when an import is attempted.

This feature magically transcends our most annoying limitations. 

I think the trickiest design question is how to deal with mixing syncifying calls with non syncifying calls. Syncifying calls always return a promise. Non syncifying calls trap if a promise is syncified. It should be possible to set a flag on entry to the vm indicating whether it was a syncifying entry or a nonsyncifying one. A more sophisticated approach to reentrancy might be needed as well.

src/core/async_call.js
```diff
@@ -0,0 +1,75 @@
...
+      return _orig_Py_CheckEmscriptenSignals_Helper();
+    };
+    const suspending__Py_CheckEmscriptenSignals_Helper =
+      new WebAssembly.Function(
+        { parameters: ["externref"], results: ["i32"] },
+        wrapper,
+        { suspending: "first" },
+      );
+    const bytes = [
+      0, 97, 115, 109, 1, 0, 0, 0, 1, 10, 2, 96, 1, 111, 1, 127, 96, 0, 1, 127,
```
ryanking13, line 22: What's this?

hoodmane: The wat currently looks like:
<details>
<summary>Syncify wat</summary>


```wat
(module
    (global $suspender (import "e" "s") (mut externref))
    ;; Status flag to tell us if suspender is usable
    (global $check (import "e" "c") (mut i32))
    ;; Wrapped syncify function. Expects suspender as a first argument and a
    ;; JsRef to promise as second argument. Returns a JsRef to the result.
    (import "e" "i" (func $syncify_promise_import (param externref i32) (result i32)))
    ;; Wrapped syncify_promise that handles suspender stuff automatically so
    ;; callee doesn't need to worry about it.
    (func $syncify_promise_export (export "o")
      (param $idpromise i32) (result i32)
      (local $suspender_copy externref)
      (global.get $check)
      (i32.eqz)
      if
        ;; If no valid suspender, return 0. Callee needs to check for this case
        ;; and set Python error flag (or otherwise handle it).
        i32.const 0
        return
      end
      ;; Store suspender into a local variable in order to manage reentrancy
      (global.get $suspender)
      (local.tee $suspender_copy)
      (local.get $idpromise)
      (call $syncify_promise_import) ;; onwards call args are (suspender, orig argument)
      (local.get $suspender_copy)
      (global.set $suspender) ;; restore global variable
    )
)
```



</details>
The check signals wat is similar:

<details>
<summary>check signals wat</summary>

```wat
(module
    (global $suspender (import "e" "s") (mut externref))
    (global $check (import "e" "c") (mut i32))
    (import "e" "i" (func $CheckSignals (result i32)))
    (import "e" "r" (func $await_sleep_0 (param externref)))
    (func $suspending_call_export (export "o") (result i32)
      (local $suspender_copy externref)
      (global.get $check)
      if
        ;; If we have a valid suspender, insert an await sleep 0 to unblock
        ;; event loop.
        (global.get $suspender)
        (local.tee $suspender_copy)
        (call $await_sleep_0)
        (local.get $suspender_copy)
        (global.set $suspender)
      end
      ;; In any case call original CheckSignals function
      (call $CheckSignals)
    )
)
```

</details>

The `__pyproxy_apply` wrapper receives a suspender as the first argument which it stores into the suspender global variable. Then it plugs the rest of the arguments into the wrapped function:
<details>
<summary>__pyproxy_apply wat</summary>

```wat
(module
    (global $suspender (import "e" "s") (mut externref))
    (import "e" "i" (func $apply_import (param i32 i32 i32 i32 i32) (result i32)))
    (func $apply_export (export "o")
      (param $susp externref) (param $a i32) (param $b i32) (param $c i32) (param $d i32) (param $e i32) (result i32)
      ;; Store first variable into suspender global
      (local.get $susp)
      (global.set $suspender)
      ;; Plug remaining args into onwards call
      (local.get $a)
      (local.get $b)
      (local.get $c)
      (local.get $d)
      (local.get $e)
      (call $apply_import)
    )
)
```
</details>

hoodmane:
We should probably look into using the wasm binary toolkit to convert wat text source files to wasm. Then we can convert the wasm to base64, create javascript files which define the module binaries, and inject these javascript files with `--pre-js`.

hoodmane:
This code isn't properly reentrant. In order to get it to work, we have to mess around with the Python `TState`. There is the idea of using greenlet #2664 but looking at the code I'm feeling a bit pessimistic about our chances of porting greenlet.

rth:
Very exciting work! Hopefully, the browser support will get stabilized in the near future.

hoodmane:
We seem to be hitting a bug in Emscripten that is triggered when `WebAssembly.Function` exists.

rth:
Thanks for working on this! It would certainly solve lots of issues! Do you have some information as to when stack switching could potentially be stabilized (or at least implemented) in top 3 browsers? I don't see it in https://webassembly.org/roadmap/ for now.

hoodmane:
It seems that it works in chrome v112 and v113 but not in chrome v114...

hoodmane:
All tests pass now! Still have to fix ctypes compatibility though.

hoodmane:
> Do you have some information as to when stack switching could potentially be stabilized (or at least implemented) in top 3 browsers?

No, I'm not sure. It's a stage 3 proposal "Implementation Phase":
https://github.com/WebAssembly/proposals
For comparison, Wasm EH is stage 3 and has been implemented by Chrome, Firefox, Safari, and Node not behind a flag. JS Type Reflection is also stage 3 and has been implemented by Chrome and Safari. JSPI is only implemented by Chrome and only behind a flag. But I think Chrome might remove the flag in the next ~six months.

hoodmane:
@rth @ryanking13 
This is ready in the sense that the only remaining test failures are tests that fail on the main branch. I know it's a fair amount to review. I split off #3954, and I can work on splitting this into a few PRs that provide intermediate features...

ryanking13:
Thanks for a tremendous work @hoodmane! I'll start reviewing split-off PRs soon. I was not following the whole activities in the organization lately because of some personal busy things, so feel free to ping me if there is a blocker that I missed.

lesteve:
FWIW I ran the scipy test suite against this PR (commit 017ae1c0eacc515a9e31d4c6e26a4f850e403f4) and there are some new (compared to `main`) failures as hinted in https://github.com/pyodide/pyodide/pull/3953#issuecomment-1605050751:

```
scipy.linalg.tests result expected in ['passed'], got 'fatal error or timeout' instead
scipy.sparse.linalg._dsolve.tests result expected in ['passed'], got 'fatal error or timeout' instead
scipy.sparse.linalg.tests result expected in ['passed'], got 'fatal error or timeout' instead
scipy.spatial.tests result expected in ['passed'], got 'fatal error or timeout' instead
scipy.stats.tests result expected in ['failed'], got 'fatal error or timeout' instead
```

hoodmane:
I'm confused, `scipy.linalg.tests` seems to fail in the same place on the main branch as far as I can tell...

hoodmane:
Oh I see you've skipped those.

hoodmane:
Let's move the scipy discussion to #3957.

hoodmane:
@rth @ryanking13 this is ready to review. By splitting off a bunch of stuff, I've gotten it down to 408 lines added for features and 326 lines for tests which is more than a 2x decrease in size from before. The current state of this PR leaves out three things:
* ctypes integration
* stack state management
* Python state management

I have dealt with all three of these issues on local branches and will make more followup PRs after this is merged to fix them.

src/core/stack_switching/suspenders.mjs
```diff
@@ -0,0 +1,166 @@
...
+  const instance = new WebAssembly.Instance(module, {
+    e: {
+      s: Module.suspenderGlobal,
+      i: suspending_f,
+      c: Module.validSuspender,
+    },
+  });
+  // Assign to the function pointer so that hiwire_syncify calls our wrapper
+  // function
+  HEAP32[_syncifyHandler / 4] = addFunction(instance.exports.o);
```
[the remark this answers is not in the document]
hoodmane: Right the other assignment to `Module.syncifyHandler` is out of date. I updated it, but I guess I need a test for this.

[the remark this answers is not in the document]
hoodmane: Oh it was dead code anyways.

packages/fpcast-test/fpcast-test/fpcast-test.c
```diff
@@ -130,6 +155,19 @@ static struct PyModuleDef module = {
   Test_Methods
 };
 
+typedef double
+DFunc(double);
+
+double
+iter_f(DFunc f, double x, int n)
```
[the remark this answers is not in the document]
hoodmane: Also dead code, I was originally using this to test ctypes but I found a simpler way. Thanks!

src/core/pyproxy.ts
```diff
@@ -18,6 +18,7 @@ declare var Tests: any;
 declare var Module: any;
 declare var Hiwire: any;
 declare var API: any;
+declare function sleep(ms: number): Promise<undefined>;
```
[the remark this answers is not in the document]
hoodmane: I put it in `pre.js` because that makes it visible everywhere (except in the loader logic): inside of `api.ts`, inside of `EM_JS`, etc. `sleep` is a kind of simple utility function, so I figure it could eventually be useful in other places. But it could be in `pyproxy.ts`, it's not all that important where it goes.

rth:
Thanks for doing this work! I'll review it tomorrow.

src/core/pyproxy.ts
```diff
@@ -450,6 +459,59 @@ Module.callPyObjectKwargs = function (
   return result;
 };
 
+async function callPyObjectKwargsSuspending(
```
rth, line 462: Can you add a docstring briefly summarizing what this function does?

src/core/stack_switching/esbuild.config.mjs
```diff
@@ -4,10 +4,38 @@
  */
 
 import { build } from "esbuild";
+import { readFileSync } from "node:fs";
+import loadWabt from "../../js/node_modules/wabt/index.js";
```
rth, line 8: Can we avoid the hardcoded assumption as to where node_modules are? It doesn't work with a non relative import?

src/core/stack_switching/esbuild.config.mjs
```diff
@@ -4,10 +4,38 @@
...
 import { dirname, join } from "node:path";
 
 const __dirname = dirname(new URL(import.meta.url).pathname);
 
+const { parseWat } = await loadWabt();
+
+/**
+ * An esbuild plugin to handle wat imports. It uses the wasm binary toolkit to
```
rth, line 16: There was https://github.com/mitschabaude/esbuild-plugin-wat but it doesn't look maintained.

src/core/stack_switching/suspenders.mjs
```diff
@@ -0,0 +1,178 @@
...
+}
+
+// for using wasm types as map keys
+function wasmTypeToString(ty) {
+  return `params:${ty.parameters};results:${ty.results}`;
+}
+
+/**
+ * This function stores the first argument into suspenderGlobal and then makes
+ * an onward call with one fewer argument.
```
rth, line 62: Maybe another sentence explaining for what purpose this is done wouldn't hurt.

src/core/stack_switching/suspenders.mjs
```diff
@@ -0,0 +1,178 @@
...
+ *   awaits a promise, then resumes execution and returns the promise result
+ *   (used by hiwire_syncify)
+ *
+ * If the creation of these fails because JSPI is missing, then we set it up so
+ * that callKwargsSyncifying and hiwire_syncify will always raise errors and
+ * everything else can work as normal.
+ */
+export function initSuspenders() {
+  // It would be nice to have a better way to feature detect wasm stack
+  // switching than this. For now I haven't come up with anything better.
```
rth, line 160: [This](https://github.com/GoogleChromeLabs/wasm-feature-detect/blob/main/src/detectors/jspi/index.js) doesn't work ?

conftest.py
```diff
@@ -16,6 +16,14 @@
 import pytest_pyodide.runner
 from pytest_pyodide.utils import package_is_built as _package_is_built
 
+pytest_pyodide.runner.CHROME_FLAGS.extend(
+    [
+        "--enable-features=WebAssemblyExperimentalJSPI",
+        "--enable-experimental-webassembly-features",
+    ]
+)
+pytest_pyodide.runner.NODE_FLAGS.extend(["--experimental-wasm-stack-switching"])
```
rth, line 25: So the idea is that we are still running Firefox tests without JSPI right? 

I wonder if it wouldn't be good to still run some tests for chrome (and possibly node) without this flag. Since that's the default configuration where most users are currently.

rth (commented):
That's certainly ambitious work that would solve a lot of Podide's problems! Thanks @hoodmane !

I'm not following all of what's happening on the lower level side, but some more comments are below and +1 from me for merging.

Maybe the major point is to make sure that there are no changes in behavior if JSPI is disabled, so maybe we should still run tests without JSPI (but rather adds CI runs where it's enabled). I'm not sure.

For the documentation, are you planning to add it separately (even if it's experimental). In particular, I would be interested in how this compares, broadly speaking, to ASYNCIFY=2 in Emscripten which also uses JSPI from what I understand.

src/core/stack_switching/suspenders.mjs
rth: BTW that whole library looks quite nice (and lightweight)  https://github.com/GoogleChromeLabs/wasm-feature-detect maybe we should be using it?

src/core/stack_switching/esbuild.config.mjs
hoodmane: In any case, for 17 lines of code I'd rather spin our own than bring in a dependency. Of course we're depending on the wasm binary toolkit here but I think that has a good chance of being maintained.

src/core/stack_switching/esbuild.config.mjs
hoodmane: Problem is that this directory isn't under `node_modules`. Most likely there is a better way, like moving this script into `src/js` or something. But I'm not sure what the best approach is.

src/core/stack_switching/suspenders.mjs
hoodmane: It does work. Though I'm a bit unclear on what the point of `WebAssembly.Suspender` is. I opened an issue about it on the proposal:
https://github.com/WebAssembly/js-promise-integration/issues/21

conftest.py
hoodmane: I think it would be good to run the fpcast tests, the C++ exception handling tests, the ctypes tests, the rust tests, and the recursion limit tests in the core test suite in both configurations. The remaining tests can be run in just one or the other.

src/core/stack_switching/esbuild.config.mjs
hoodmane: I would like to leave fixing this to a followup.

conftest.py
hoodmane: I'll fix this in a followup PR.

hoodmane:
@rth okay to merge this?

rth:
Sorry for the delay @hoodmane . Yes, no objection on my part. Thanks for doing all this work!

Issue #3092, opened 2022-09-10
More general unvendoring mechanism


rth:
Currently, we unvendor tests via the `build/unvendor-tests` flag. It could be useful to additionally have a more generic unvendoring mechanism to unvendor other sub-components. For instance, some packages have included datasets (sometimes rather large) that it could be useful to unvendor:  geopandas 1MB, scikit-learn 0.5MB, scikit-image: 7.5MB

Maybe it could be a flag of the form,
```yaml
build:
    unvendor:
         - package-name: <regrex or glob>
         - package-name2: <regex or glob 2>
```

Also,
- [ ] ideally, the list of unvendored packages for a given package should be easily discoverable in the  `repodata.json` (including the unvendored tests)
- [ ] we could add another `MetaPathFinder`  similar to [`UnvendoredStdlibFinder`](https://github.com/pyodide/pyodide/blob/b4433e1b4a28f4d904c85bd417f95ea667d329fc/src/py/_pyodide/_importhook.py#L145) which for a given ImporError, looks at the path, checks it's in a known `top_level` fields, and if it raises a meaningful error about how to install the missing component with micropip. This would improve a lot the usability of unvendored packages.  
- [ ] The previous step would require that the `top_level` entries are correctly auto-determined from the unvendored packages (which is likely not the case currently). Related https://github.com/pyodide/pyodide/pull/3006

ryanking13:
It sounds good to me. 

> * looks at the path, checks it's in a known `top_level` fields

I think unvendored components are mostly not top_level directories, so I don't think it is a good idea to use top_level key for that usage. Perhaps we need a separate key for that.

rth:
>  unvendored components are mostly not top_level directories, so I think we need a separate key for that usage.

Good point, that they are not top level. In the end though we don't really care much whether these paths are top level or not,  I think? All we care is that we can use it to have the import name -> package name mapping. And it would be better to have 1 field for this rather than two. So maybe having an `package>imports` field would be actually more general than `package>top_level`. Particularly since it's not part of a stable release yet.  WDYT?

ryanking13:
> > unvendored components are mostly not top_level directories, so I think we need a separate key for that usage.
> 
[...]

Issue #3091, opened 2022-09-09
add fiona


raybellwaves:
## 🐍 Package Request

- Package Name and Version : fiona >= 1.8 e.g. 1.8.12
- Package URL : https://github.com/Toblerity/Fiona
- Package Dependencies that needs to be resolved first: Not sure

dopplershift:
GDAL (C++) will be needed first.

rth:
Yes, GDAL would likely be difficult. Though fortunately, it's possible to use Geopandas without fiona at least for GeoJSON format https://github.com/pyodide/pyodide/issues/1569 and likely other file formats in the future https://github.com/geopandas/geopandas/issues/2548#issuecomment-1242702730

ryanking13:
fiona and geopandas is now added (#3213). If you have trouble using it, feel free to open a new issue.

MarcSkovMadsen:
Due to a request from Moving Pandas here https://discourse.holoviz.org/t/try-out-panel-sharing/4549/5 I just tried to install fiona and geopandas in the pyodide console via micropip. But it fails. Should I open a new issue @ryanking13 ?

![image](https://user-images.githubusercontent.com/42288570/201507244-72856526-1674-47f4-aa6a-6008f0284b00.png)

Also the test in #3213 fails https://ryanking13.github.io/geopandas-pyodide-test/

![image](https://user-images.githubusercontent.com/42288570/201507302-0cacb6d8-6d15-4b12-adad-05fe4bc00825.png)

FYI. @raybellwaves

ryanking13:
I guess it is answered in #3248? (I also fixed the URL in my example repo, thanks for the report!)

Issue #2580, opened 2022-05-19
Removing pure Python packages from the repository


ryanking13:
Currently, we have dozens of pure Python packages in the repository. Some packages require a patch, and some packages don't have a wheel, but many packages exist only because other packages with C-extension depend on them in runtime.

Note that there was some conversation about removing these pure Python packages from the repository in https://github.com/pyodide/pyodide/pull/2132#issuecomment-1059680883. I am opening this issue is to extend that discussion.

I think there is two things two discuss about this topic.

**1. How to specify dependencies: explicit or implicit**

1) We can specify pure python package dependencies in `meta.yaml` file following conda syntax as @henryiii suggested such as:

```yaml
requirements:
  run:
    - numpy
    - matplotlib
    - sympy
    - pip:
        - another_package
```

2) Or we could implicitly (please fix this wording if it sounds awkward) resolve dependencies from the wheel METADATA, which is how micropip is currently doing.

The latter would be optimal because we don't need to maintain the list of dependencies but it is little more complicated.

**2. Adding pure Python packages (and its dependencies) to the distribution or not**

1) We can download pure Python packages from PyPI in the build time with their dependencies and deploy them together with packages with C-extension.

2) Or we can download them in the runtime, i.e. mark URLs of each package in the `packages.json` and let `pyodide.loadPackage` to download the package from PyPI.

If we use former approach we can 1) deploy all packages in the same CDN and 2) can use parallel download for pure Python packages with its dependencies.

On the other hand if we use latter approach, we can 1) reduce the size of the distribution, while 2) dependencies of each pure Python package need to be resolved in a runtime.

---

There must be other things that I didn't consider. Please share some ideas.

ryanking13:
> Maybe we can include them in the distribution if there are tests to unvendor and otherwise not? Or maybe we should check whether the tests exceed some threshold in size. If there are a few small test files, it's not a big waste to download them. If there is 300kb of tests, probably it's worth unvendoring?

I never thought about the test file size. Good point.

ryanking13:
> Probably explicit is good, ideally we would update `mkpkg` so it can generate the list automatically.

A possible issue of that can would be how we deal with dependencies of dependencies.

hoodmane:
I confirmed that it is okay to put PyPI urls into `packages.json` in the `file_name` field. It required a one line change to `_package_loader.unpack_buffer` but that was it:

[...]

PR #3213, opened 2022-10-31
PKG Add geopandas, fiona, and GDAL

### LINKED ISSUE #1569: Adding geopandas?
Is it thinkable to include geopandas? Or is that too heavily depending on C/C++ packages?

ryanking13:
### Description

Close #1569, #3091

This PR adds geopandas and its dependencies: fiona, gdal.

- Some other pure Python libraries are also added (click, munch, cligj) because fiona depends on them. (Those can probably be removed when #2580 is accomplished)
 - geopandas is also a pure Python library, but I added it in order to run tests and maybe to unvendor some of its components later #3092.

Obviously, GDAL was challenging to build, as it has dozens of dependencies. I only enabled recommended libraries, while disabling most of the optional dependencies.

I am not very used to geopandas and geospatial data, so I am not sure I have included all commonly used libraries that GDAL uses. So people who are familiar with geospatial data would need to check if I have omitted dependencies that are critical to deal with geospatial data.

### Checklists
 - [x] Add / update tests


ryanking13:
Demo: https://ryanking13.github.io/geopandas-pyodide-test/

(It takes very long to load at the first time)

packages/gdal/meta.yaml
```diff
@@ -0,0 +1,102 @@
...
+    export EMSCRIPTEN_BIN=$EMSCRIPTEN_SYSROOT/bin
+    export EMSCRIPTEN_LIB=$EMSCRIPTEN_SYSROOT/lib/wasm32-emscripten/pic
+
+    mkdir -p dist
+    export DISTDIR=$(pwd)/dist
+
+    embuilder build zlib --pic
+    embuilder build libjpeg --pic
+    embuilder build libpng --pic
+    embuilder build sqlite3 --pic
```
hoodmane, line 31: Do we need a second copy of sqlite3 for gdal? Or does this share the original one?

packages/gdal/meta.yaml
```diff
@@ -0,0 +1,102 @@
...
+
+    mkdir -p build
+    cd build && emcmake cmake .. \
+      -DCMAKE_INSTALL_PREFIX=$WASM_LIBRARY_DIR \
```
hoodmane, line 35: So many flags...

packages/gdal/meta.yaml
```diff
@@ -0,0 +1,102 @@
...
+
+
+    # Note:
+    # CMake tries to link the same libraries multiple times at the final link time.
+    # This is probably because the same libraries are used in multiple submodules.
+    # This behavior is okay when those libraries are "shared" libraries,
+    # but we often build static libraries and linking static libraries
+    # multiple times results in a duplicated symbol error.
+    # I wasn't able to find a way to prevent CMake from emitting duplicated libraries.
+    # This is a hack which removes all duplicated appearances of static libs at the final link time.
```
hoodmane, line 91: I guess it's a good thing you are very good at cmake...

hoodmane (approved):
Thanks @ryanking13! How long do the tests take?

packages/gdal/meta.yaml
ryanking13: It shares the original one. `embuilder build sqlite3` will explicitly build sqlite3 while adding `-sUSE_SQLITE3` flag will implicitly build it. But both shares the same library (built into emsdk/emsdk/emscripten/cache/sysroot/...).

I found that `-sUSE_...` flags are sometimes ignored (maybe it is related to CMake or https://github.com/emscripten-core/emscripten/issues/17675, not sure though). So I used `embuilder` to explicitly build those libraries.

packages/gdal/meta.yaml
ryanking13: Every time I build a new CMake package, I learn new things about it 😂

ryanking13:
> How long do the tests take?

In CI:

- fiona (40s in chrome / 13s in firefox / 12s in node)
- geopandas (70s in chrome / 125s in firefox / 81s in node)

Maybe I need to skip some tests in geopandas.

ryanking13:
@raybellwaves @dopplershift @leouieda

Are you interested in testing this PR? It would be nice if some of you can check if it works well for common use cases, as I am not very used to dealing with geospatial data.

You can download build artifacts from: https://output.circle-artifacts.com/output/job/212b5abe-4de0-4287-a6a5-f29bda80551b/artifacts/0/root/repo/pyodide.tar.gz

packages/gdal/meta.yaml
ryanking13: Actually, after #3128, CMake can now search libraries that are in WASM_LIBRARY_DIR automatically. But in order to prevent some unexpected side effects (selecting a static library instead of a shared library, etc...), I just specified all library paths. I have some future plans to clean up all these verbose flags, but I'm in an "it works, so don't touch it" state 👀

rth:
> Maybe I need to skip some tests in geopandas.

Yes, reducing the test runtime would be preferable. Untill we find a way to properly run the test suite as part of our CI, without it taking too long (and re-running everything).

ryanking13:
> Yes, reducing the test runtime would be preferable. Untill we find a way to properly run the test suite as part of our CI, without it taking too long (and re-running everything).

Okay, I reduced the geopandas test duration:

- chrome: 23s
- firefox: 82s
- node: 30s
- safari: 27s

The firefox still takes very long to run test, but around ~60s is import time, so I guess there is no way to reduce this :/

ryanking13:
Thanks for the reviews! The test failure doesn't seem related to this PR. merging.

deeplook:
Awesome, thanks so much!! ❤️

PR #3216, opened 2022-11-01
Add `pyodide build recipe` command

ryanking13:
### Description

Related: #2879

This adds `pyodide build recipe` command, which will replace `pyodide-build buildall`.

I think the subcommand name "recipe" fits as it uses meta.yaml to build packages compared to out-of-tree build. But in #2879, @rth suggested the name "graph" or "tree" so I am open to other names.

- Usage:
  - `pyodide build recipe numpy scipy matplotlib`
  - `pyodide build recipe "*"`

After this PR, we will have two subcommands in `pyodide build`:

- `pyodide build recipe`
- `pyodide build wheel` (calling `pyodide build` without subcommand will invoke `pyodide build wheel`).

### Checklists

- [x] Add / update tests


run_docker
```diff
@@ -137,6 +137,7 @@ CONTAINER=$(\
         --groups sudo  \
         $USER_NAME \
       ; \
+      echo 'export PATH=\$PATH:$USER_HOME/.local/bin' >> /etc/profile; \
```
ryanking13, line 140: This is to make `pyodide` entrypoint (and other python applications) availble in the docker image.

pyodide-build/pyodide_build/cli/build.py
```diff
@@ -0,0 +1,111 @@
...
+    buildall.build_packages(recipe_dir_, output_dir, args)
+
+
+@app.command("wheel")  # type: ignore[misc]
+def build_wheel(
+    exports: str = typer.Option(
+        "requested",
+        help="Which symbols should be exported when linking .so files?",
+    ),
+    flags: list[str] = typer.Option(None, help="Build flags passed to pypa/build"),
```
ryanking13, line 106: Note that this is a breaking change. Previously, flags was passed via: `pyodide build flag1 flag2 flag3`. But now it has to be passed via `pyodide build --flags "flag1 flag2 flag3"`.

This is to prevent confusion between flag and subcommand.

ryanking13:
> I would like to keep pyodide build as a subset of [pypa/build](https://pypa-build.readthedocs.io/en/latest/#python--m-build)'s API. In pypa/build the positional argument is a path to the local directory we want to build which I think is more natural.

It seems like [@hoodmane wants to keep `pyodide build` as a subset of `pypa/build`](https://github.com/pyodide/pyodide/pull/3196#discussion_r1002032179) (I think it makes sense), so I guess we need to find a different name for this subcommand.

joemarshall:
I integrated this functionality into @3196 (and copied much of the code from this PR in the process), so this hopefully won't be needed.

PR #3220, opened 2022-11-02
CI Move test-docs job to GHA

ryanking13:
This PR moves the` test-docs` job to GHA, so we can check earlier if the requirements for a document are broken.


rth (approved):
Thanks!

PR #3222, opened 2022-11-03
DOC Add a FAQ entry for async filesystem usage

### LINKED ISSUE #3208: idbfs not really write data to idb
## 🐛 Bug

idbfs not really write data to idb

### To Reproduce
run pyodide in chrome: 

in js console:
pyodide.FS.mkdir("/mnt/")
pyodide.FS.mount(pyodide.FS.filesystem.IDBFS,{},"/mnt/")

run python code by pyodide,the py code as the follow:
f=open("/mnt/test.txt","w")
f.write("123456")
f.flush()
f.close()

and then,in chrome's idb data tab,you can't see any data for the file "test.txt".

and after i refresh the web page,the file "test.txt" disapeared.

### Environment

- Pyodide Version 0.21.2:
- Browser version chrome 106.0.5249.119:

ryanking13:
This adds a FAQ entry for #3208

Close #3208

### DIFF
--- a/docs/usage/faq.md
+++ b/docs/usage/faq.md
@@ -411,11 +411,11 @@ from mymodule import hello
 hello()
 ```
 
-## Why the changes I made in IndexDB doesn't persist?
+## Why changes made to IndexedDB don't persist?
 
-Unlike other filesystems, indexDB (pyodide.FS.filesystem.IDBFS) is an asynchronous filesystem.
-This is because browsers offer only asynchronous interfaces for indexDB.
-So in order to make changes persist, you have to call
+Unlike other filesystems, IndexedDB (pyodide.FS.filesystem.IDBFS) is an asynchronous filesystem.
+This is because browsers offer only asynchronous interfaces for IndexedDB.
+So in order to persist changes, you have to call
 [`pyodide.FS.syncfs()`](https://emscripten.org/docs/api_reference/Filesystem-API.html#FS.syncfs).
 See [Emscripten File System API](https://emscripten.org/docs/api_reference/Filesystem-API.html#persistent-data)
 for more details.
--- a/docs/usage/faq.md
+++ b/docs/usage/faq.md
@@ -51,7 +51,8 @@ For security reasons JavaScript in the browser is not allowed to load local data
 You will run into Network Errors, due to the [Same Origin Policy](https://en.wikipedia.org/wiki/Same-origin_policy).
 There is a
 [File System API](https://wicg.github.io/file-system-access/) supported in Chrome
-but not in Firefox or Safari.
+but not in Firefox or Safari. See {ref}`nativefs-api` for experimental local file system
+support.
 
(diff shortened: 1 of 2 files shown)


docs/usage/faq.md
```diff
@@ -409,3 +410,12 @@ importlib.invalidate_caches() # Make sure Python notices the new .py file
 from mymodule import hello
 hello()
 ```
+
+## Why the changes I made in IndexDB doesn't persist?
```
rth, line 414: Maybe
```suggestion
## Why changes made to IndexedDB don't persist?
```

docs/usage/faq.md
```diff
@@ -409,3 +410,12 @@ importlib.invalidate_caches() # Make sure Python notices the new .py file
...
+
+Unlike other filesystems, indexDB (pyodide.FS.filesystem.IDBFS) is an asynchronous filesystem.
+This is because browsers offer only asynchronous interfaces for indexDB.
```
rth, line 417:
```suggestion
This is because browsers offer only asynchronous interfaces for IndexedDB.
```

docs/usage/faq.md
```diff
@@ -409,3 +410,12 @@ importlib.invalidate_caches() # Make sure Python notices the new .py file
 from mymodule import hello
 hello()
 ```
+
+## Why the changes I made in IndexDB doesn't persist?
+
+Unlike other filesystems, indexDB (pyodide.FS.filesystem.IDBFS) is an asynchronous filesystem.
```
rth, line 416:
```suggestion
Unlike other filesystems, IndexedDB (pyodide.FS.filesystem.IDBFS) is an asynchronous filesystem.
```

docs/usage/faq.md
```diff
@@ -409,3 +410,12 @@ importlib.invalidate_caches() # Make sure Python notices the new .py file
...
+This is because browsers offer only asynchronous interfaces for indexDB.
+So in order to make changes persist, you have to call
```
rth, line 418:
```suggestion
So in order to persist changes, you have to call
```

rth (approved):
Thanks @ryanking13 ! A few minor suggestions.

docs/usage/faq.md
ryanking13: Wow I've been knowing the wrong name for years.

ryanking13:
Thanks for the review @rth!

hoodmane:
I think we should add an FAQ "how do I persist files/data between sessions" before this one that explains how to use the idbfs. Because it looks to me like we have a bigger problem which isn't just that people don't know how to make sure the data persists, but also many people are not likely to know that the feature exists at all.

hoodmane:
Though @ryanking13 has improved this quite a bit from "we don't mention it at all" to "we kinda mention it a little".

ryanking13:
Yeah, it would be nice if we have a FAQ entry about using various file system types.

PR #3224, opened 2022-11-03
Fix CLAPACK install path

ryanking13:
This is split up from #3167.

This change CLAPACK to be installed in the same directory where other libraries are saved.


rth (approved):
Thanks!

PR #3228, opened 2022-11-04
MAINT Separate FS creation function from loadDynlib

ryanking13:
### Description

This is another split up from #3167. It separates FS object creation part inside the loadDynlib function into a separate function.
It makes the loadDynblib function easy to read. No functional changes are included, except for some additional debug logging and types.


ryanking13:
Thanks for the review!

PR #3230, opened 2022-11-05
Add a tool to allow us to embed wasm modules with hand coded wat

hoodmane:
This allows us to embed extra wasm modules into the output. This is split off of #3210. We need this for #3210 since
1. We don't want to assume that externref is supported but we need to use it as part of the JS Promise integration proposal. To handle this we segment off the externrefs into separate wasm modules. If externref is not supported, we load most everything successfully but these small modules fail to load
2. Even supposing that we know that externref is supported, there is no way to write C code such that the compiled code uses externref. It's possible using the emscripten assembly format, but that is less well documented and harder to use for our purposes than wat.

This PR doesn't actually change anything by itself.


tools/assemble_wat.js
```diff
@@ -0,0 +1,17 @@
+const fs = require("fs");
+const { execFileSync } = require("child_process");
+
+process.argv[2].split();
+
+const path = process.argv[2].split("/");
+const filename = path.pop().split(".")[0];
+process.chdir(path.join("/"));
+
+execFileSync("wat2wasm", [filename + ".wat", "--enable-all"]);
```
ryanking13, line 10: Myabe check `wat2wasm` exists before executing it, and show proper error messages if it doesn't exist?

ryanking13 (approved):
Thanks @hoodmane!

PR #3232, opened 2022-11-06
Fixes to package loading from #3229

hoodmane:
Add package directory to library search path and set target to 'site' when we `loadPackage` from a wheel.
This fixes two minor bugs I found in #3229 that prevent that package from working correctly.

### DIFF
--- a/src/js/dynload.ts
+++ b/src/js/dynload.ts
@@ -29,10 +29,11 @@ export async function loadDynlib(lib: string, shared: boolean) {
     console.debug(`Loading a dynamic library ${lib}`);
   }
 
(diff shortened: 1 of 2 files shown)


ryanking13:
Interesting. I suspect this will have no effect because if a path is an absolute path, it will not look at other candidate directories. But if it actually fixes the error in #3229, there should be something that I am missing...

hoodmane:
The path to the python module in #3229 is absolute but the path to the helper shared library is not.

ryanking13 (approved):
Oh, okay. That makes sense. Thanks!

PR #3233, opened 2022-11-07
Continulet

hoodmane:
This implements continulet. Continulet is a pypy module on top of which they implement greenlet, so this is work on #2664. This is on top of #3210. It passes almost all of the tests from the pypy test suite.

I haven't implemented `permute` yet, which `greenlet` also uses. The pypy implementation of `permute` kind of freaks me out: they move around some frame data and it changes the frame out from under the function. See here:
https://foss.heptapod.net/pypy/pypy/-/blob/branch/default/pypy/module/_continuation/test/test_stacklet.py#L742
Anyways probably it's possible to make that work too.

Unfortunately this only works with v8 version >= 10.8. Node 19 has v8 10.7. But it works in chrome beta and chrome unstable.

The pypy implementation is here:
https://foss.heptapod.net/pypy/pypy/-/blob/branch/default/pypy/module/_continuation/interp_continuation.py
The pypy test suite is here:
https://foss.heptapod.net/pypy/pypy/-/blob/branch/default/pypy/module/_continuation/test/test_stacklet.py

### Checklists

- [ ] Add / update tests


rth:
Any chance this and https://github.com/pyodide/pyodide/pull/3210 could be implemented as a plugin, so that it wouldn't be included by default, but there would be a way to enable it with the default build? Until browser support gets better that is.

hoodmane:
> implemented as a plugin

Yeah that seems like a good idea, as opposed to keeping it as a long-lived branch. I think it's possible with some effort.

PR #3234, opened 2022-11-07
Add vendor-sharedlib key in meta.yaml spec

ryanking13:
### Description

This is anothor split from #3183. 

This adds `vendor-sharedlib` key. If set to True, pyodide-build will copy shared libs into the wheel file after build.
It is similiar to [`auditwheel repair`](https://github.com/pypa/auditwheel), but without modifying RPATH.

Since there is no such thing like RPATH in Emscripten WASM module, we need to manage it manually in JS side (which will be a last split up PR for #3183).

Note that `vendor-sharedlib` is default to False. So no packages are affected by this PR itself.

### Checklists

- [x] Add / update tests


hoodmane (approved):

rth:
Nice to see more use of auditwheel-emscripten, thanks @ryanking13 !

PR #3238, opened 2022-11-10
Merge library and sharedlibrary key to build/type

### LINKED ISSUE #3066: Merge sharedlibrary and library meta.yaml flags
As proposed in https://github.com/pyodide/pyodide/pull/3055#pullrequestreview-1094620159,

>  it would be better to have a single flag library: which can be static, shared, or False. Rather than two binary flags library and sharedlibrary. It's weird to have library=False for something that's a shared library.

ryanking13:
### Description

Resolve #3066

This adds a new key `build/type` in meta.yaml spec, which merges `build/library` and `build/sharedlibrary` key.

It has 4 values (we can add more if needed):
- package (default, normal Python package)
- static_library
- shared_library
- cpython_module (unvendored cpython modules. Previously, it was handled via a private key `package/_cpython_dynlib`)

### Checklists

- [x] Add / update tests
- [x] Check documentation package list rendering


hoodmane (approved):
Thanks @ryanking13!

rth:
Sounds good to me as well. Thanks @ryanking13 !

PR #3239, opened 2022-11-10
Emscripten 3.1.25

hoodmane:
Hard to keep up with these. We just got up to latest and then there's a new release. I am loving not having to migrate patches anymore!

### DIFF
--- a/Makefile.envs
+++ b/Makefile.envs
@@ -1,5 +1,5 @@
 export PYVERSION ?= 3.10.2
-export PYODIDE_EMSCRIPTEN_VERSION ?= 3.1.24
+export PYODIDE_EMSCRIPTEN_VERSION ?= 3.1.25
 
 export PLATFORM_TRIPLET=wasm32-emscripten
 export SYSCONFIG_NAME=_sysconfigdata__emscripten_$(PLATFORM_TRIPLET)
--- a/docs/project/changelog.md
+++ b/docs/project/changelog.md
@@ -28,8 +28,8 @@ substitutions:
   module-type service workers.
   {pr}`3070`
 
-- {{ Enhancement }} Emscripten was updated to Version 3.1.24
-  {pr}`2958`, {pr}`2950`, {pr}`3027`, {pr}`3107`, {pr}`3148`, {pr}`3236`
+- {{ Enhancement }} Emscripten was updated to Version 3.1.25
+  {pr}`2958`, {pr}`2950`, {pr}`3027`, {pr}`3107`, {pr}`3148`, {pr}`3236`, {pr}`3239`
 
 - {{ Enhancement }} Added a new API {any}`pyodide.mountNativeFS`
   which mounts [FileSystemDirectoryHandle](https://developer.mozilla.org/en-US/docs/Web/API/FileSystemDirectoryHandle)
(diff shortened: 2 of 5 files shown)


hoodmane:
C++ trouble. @ryanking13 I hope you'll fix it =)

ryanking13:
I'll see if I can fix it :^)

ryanking13:
```
./boost/container_hash/hash.hpp:132:33: error: no template named 'unary_function' in namespace 'std'; did you mean '__unary_function'?
        struct hash_base : std::unary_function<T, std::size_t> {};
                           ~~~~~^~~~~~~~~~~~~~
                                __unary_function
/root/repo/emsdk/emsdk/upstream/emscripten/cache/sysroot/include/c++/v1/__functional/unary_function.h:46:1: note: '__unary_function' declared here
using __unary_function = __unary_function_keep_layout_base<_Arg, _Result>;
```

The Emscripten 3.1.25 [updated LLVM version to 15](https://github.com/emscripten-core/emscripten/pull/18113), and it seems like it is an upstream issue of LLVM15 and boost incompatibility.

- https://github.com/fnalacceleratormodeling/synergia2/issues/80
- https://github.com/boostorg/container_hash/issues/22

ryanking13:
Oh, actually [boost already released 1.80.0 with a fix](https://www.boost.org/users/history/version_1_80_0.html). So we will just have to update boost to 1.80.0

hoodmane:
Hmm it's still not working. Maybe try updating boost as a separate pr?

I also am slightly worried whether rust will work. I guess we should disable boost and see if anything else breaks.

ryanking13:
Oh, right. The Boost team mentioned that it was "found too late to be fixed in the release", but they provided a patch file instead: https://www.boost.org/patches/1_80_0/0005-config-libcpp15.patch

hoodmane:
> built boost-cpp in 202.90

Great work @ryanking13 =D

hoodmane:
I think the crash in matplotlib is happening via Pillow via freetype. It's pretty hard to tell but my best guess is the error is happening here:
https://gitlab.freedesktop.org/freetype/freetype/-/blob/master/src/autofit/afloader.c#L128

The `call_indirect` instruction declares type `(param i32 i32 i32) -> void` but is defined with type `(param i32 i32 i32) -> i32`. The function being called in the source looks like it is defined to return void...

hoodmane:
I think the key thing here is to build freetype with debug symbols. Not sure how to do that though.

ryanking13:
It looks like the emscripten team [switched freetype repo](https://github.com/emscripten-core/emscripten/pull/18095/files) and they dropped a patch which modifies function signatures: https://github.com/emscripten-ports/FreeType/commit/40a760c963bc2b76575918ff91a515c8474db4e0. Maybe we can revert that change and see what happens?

hoodmane:
```patch
- typedef void
+ typedef FT_Error
  (*AF_WritingSystem_ApplyHintsFunc)( AF_GlyphHints    hints,
                                      FT_Outline*      outline,
                                      AF_StyleMetrics  metrics );
```
That's the one!

ryanking13:
@sbc100 I think we found that the patch in https://github.com/emscripten-ports/FreeType/commit/40a760c963bc2b76575918ff91a515c8474db4e0 was useful.

sbc100:
Oh, my mistake i didn't realize we how downstream freetype patches, we should revert that change.

hoodmane:
I really don't understand the current traceback.

ryanking13:
> I really don't understand the current traceback. The hash value is correct and there seems to be only one `sha512hash` argument.

Well, it seems like the signature of `fetch_project` has been changed too (https://github.com/emscripten-core/emscripten/pull/18148)

ryanking13 (approved):
Now it seems it is working well. Thanks, @hoodmane!

PR #3240, opened 2022-11-10
NFC Update gitignore

hoodmane:
Sort the .gitignore folders and add `pytest-pyodide` and virtual environments that start with `.venv`.


ryanking13 (approved):
Thanks!

PR #3242, opened 2022-11-10
Fix docs on loadPyodide

hoodmane:
You **can** load several versions of Pyodide since version 0.21.0. We should backport this.


ryanking13 (approved):
Thanks!

PR #3247, opened 2022-11-12
Remove "Python initialization complete" log line

rth:
A small change to logging when loading Pyodide, to also indicate the time it took for the initialization,

Currently, in the console, we print,
```
Python initialization complete
```
With this change,  it would log something like,
```
Python initialization complete in 1230 ms
```

This is helpful for measuring load time (e.g. related to https://github.com/pyodide/pyodide/pull/3166/) without writing extra code, for instance just by loading the console.html 

This likely doesn't need a changelog entry.

### DIFF
--- a/src/js/pyodide.ts
+++ b/src/js/pyodide.ts
@@ -267,6 +267,7 @@ export async function loadPyodide(
     _node_mounts?: string[];
   } = {},
 ): Promise<PyodideInterface> {
+  const t0_init = new Date();
   await initNodeModules();
   let indexURL = options.indexURL || calculateIndexURL();
   indexURL = resolvePath(indexURL); // A relative indexURL causes havoc.
@@ -371,6 +372,10 @@ If you updated the Pyodide version, make sure you also updated the 'indexURL' pa
   if (config.fullStdLib) {
     await pyodide.loadPackage(API._pyodide._importhook.UNVENDORED_STDLIBS);
   }
-  pyodide.runPython("print('Python initialization complete')");
+  const init_time = (new Date().getTime() - t0_init.getTime()).toFixed(0);
+
+  pyodide.runPython(
+    `print('Python initialization complete in ${init_time} ms')`,
+  );
   return pyodide;
 }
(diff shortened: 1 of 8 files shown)


hoodmane:
I've been thinking we should consider not printing this message anymore. Or at least have a `loadPyodide` flag that can turn it off. This would avoid the need to have a workaround to filter it out in the command line runner:
https://github.com/pyodide/pyodide/blob/main/tools/python#L147
Maybe we could also add info about loading time onto the `pyodide` api, like as `pyodide.performanceInfo.initializationTime` or something? That's convenient for tools.

src/js/pyodide.ts
```diff
@@ -267,6 +267,7 @@ export async function loadPyodide(
     _node_mounts?: string[];
   } = {},
 ): Promise<PyodideInterface> {
+  const t0_init = new Date();
```
bollwyvl, line 270: Note there is also [performance.now](https://developer.mozilla.org/en-US/docs/Web/API/Performance/now) which provides higher-resolution timing, if that's desired.

bollwyvl:
> pyodide.performanceInfo.initializationTime

>  consider not printing this message anymore

:+1: to offering maximally-quiet output, as well as more fine-grained performance info... perhaps by offering `messageCallback` and `errorCallback`s in more places. 

It would be lovely if `micropip` provided such output redirection (and timing information), as well, such that UIs could interact more directly messages and errors... progress bars and all that.

rth:
Sound good. +1 for a programmatically accessible performance object and for removing that log line.

> Note there is also [performance.now](https://developer.mozilla.org/en-US/docs/Web/API/Performance/now)

Thanks, that's indeed what I was looking for.

> perhaps by offering messageCallback and errorCallbacks in more places.

Indeed, there is more improvement to be done there. Would you mind opening an issue about it?

rth:
OK, for now, I just removed that log line that indeed bypassed the stdout handler. 
We can discuss programmatically accessible performance in https://github.com/pyodide/pyodide/issues/3260

ryanking13 (approved):
Thanks!

> This likely doesn't need a changelog entry.

I think it wouldn't hurt to add one. There is [Hyrum's law](https://www.hyrumslaw.com/) and there might be someone who is relying on that message.

PR #3251, opened 2022-11-13
Update scipy to v1.9.3

hoodmane:
I also tried dropping the prini patch since @WarrenWeckesser investigated and says it isn't needed / is wrong. Let's see what happens.


hoodmane:
Indeed the patch is not needed. Thanks again @WarrenWeckesser!

rth (approved):
Thanks!

Issue #3255, opened 2022-11-14
Add pyinstrument 4.4.x


bollwyvl:
## 🐍 Package Request

- Package Name and Version: `pyinstrument` / latest
- Package URL: https://github.com/joerick/pyinstrument
- Package Dependencies that needs to be resolved first: `[]`

## Notes

If even possible, it would be interesting to have `pyinstrument` available, as it is one of the few tools out there that does a good job of profiling `async` methods. Being able to profile, even with whatever overhead it adds, would still be meaningful for assessing the impact of other packages.

## References

- https://github.com/pyodide/micropip/issues/24#issuecomment-1314397063
- [conda-forge recipe](https://github.com/conda-forge/pyinstrument-feedstock/blob/main/recipe/meta.yaml)

hoodmane:
Seems like it builds fine.

bollwyvl:
:rofl: we need a `maintainerinstrument` for that absurd display of `async` development!

PR #3258, opened 2022-11-14
Package pyinstrument

rth:
Closes https://github.com/pyodide/pyodide/issues/3255


packages/pyinstrument/test_pyinstrument.py
```diff
@@ -0,0 +1,20 @@
...
+    """
+    import asyncio
+
+    from pyinstrument import Profiler
+
+    p = Profiler()
+    with p:
+        await asyncio.sleep(0.1)
+
+    p.print()
```
hoodmane, line 20: Maybe assert that p.output_text() matches some regex?

bollwyvl: There's also a JSON output available from `p.last_session.save('foo.pyisession')`, which might be more palatable to parse and less likely to change for purely cosmetic reasons... is there a way to play around with it on a CI output somewhere? I must confess I haven't gotten my local dev setup working in a while.

ryanking13: You can download the build artifacts from: https://output.circle-artifacts.com/output/job/7ffb988b-80ed-42ec-8d3f-debebcda4dc3/artifacts/0/root/repo/pyodide.tar.gz

bollwyvl: here's a thing... i feel like the JSON output is the artifact that would actually be saved as part of a benchmark, so having a handle on what can be known about it seems important, even if you'd generally reload the output. It does seem a bit odd that the JSON output doesn't break up the raw record data more (just strings with ansi codes) but meh.

```py
import asyncio
import json
from pathlib import Path

from pyinstrument.profiler import Profiler, Session

with Profiler() as p1:
    await asyncio.sleep(0.1)

p1.print(show_all=True, timeline=True)

session_file = Path("foo.pyisession")
s1 = p1.last_session
s1.save(session_file)

json_opts = dict(indent=2, sort_keys=True)
s1_data = json.load(session_file.open())
s1_json = json.dumps(s1_data, **json_opts)

# _all_ the keys as of 4.4.0, but presumably more could be added
expected_keys = {
    "cpu_time",
    "duration", 
    "frame_records", 
    "program", 
    "sample_count", 
    "start_call_stack", 
    "start_time", 
}
missing_keys = expected_keys - set(s1_data)

assert not missing_keys, f"session JSON missing: {missing_keys}"

s2 = Session.load(session_file)
s2_json = json.dumps(s2.to_json(), **json_opts)

assert s1_json == s2_json, "loaded JSON did not match"
```

rth: Thanks, @bollwyvl, added your extensive test.

ryanking13 (approved):
Thanks @rth!

bollwyvl:
Ha, just noticed this branch references `pyinstaller`. Now _that_ would be some crazy business, and I have no idea what it would actually _mean_ :exploding_head: 

I guess if you could freeze down an entire pyodide interpreter + a stack of deps with an entrypoint that could be stuffed into a single HTML containing the entire shooting match, that would sure be something. We've had a related request here: https://github.com/jupyterlite/jupyterlite/issues/497 ... I've been down that road with the old requirejs/browserify stack, and _maybe_ it would be possible with some extremely gnarly webpack and a really _ripping_ serviceworker... but i can't imagine what the performance profile would be like, even it was truly archival grade at that point.

bollwyvl:
> Well you could but I'm not sure it's really a good idea. 

Indeed, i didn't say it was a _good_ idea, merely that a "single file you get from ... somwhere... and click and it runs" has a lot going for it. A lot of Web APIs don't work with `file://` URLs, anyway... more's the pity, though the security profile is atrocious.

> serve those files

Even getting a working, _local_ HTTP server (much less good enough, on-demand internet to reach PyPI/CDN) is a fairly significant hurdle for many folks. I confess I've been tinkering with [fossil-scm](https://www2.fossil-scm.org/) as a (terrifyingly) fully-functional backend: a 3mb binary which will run on a toaster (but more importantly a RpiW) and a (X)XXmb sqlite file will fit on a throw-away USB stick is... compelling. But that's still not going to fly on a low-power Chromebook/library computer.

bollwyvl:
Anything else amiss here?

hoodmane:
I think this can be merged if the changelog merge conflict is resolved.

bollwyvl:
:heart: thanks all!

PR #3261, opened 2022-11-16
Remove pycryptodomex from the package list

ryanking13:
### Description

pycryptodomex was added in #2966, but it had an invalid recipe name `meta.yml` (not `meta.yaml`)  so our build system didn't build that package.

I found that in #3006, and I also found that it is not building well, so I disabled it then. So, in other words, pycryptodomex never worked in Pyodide.

I would like to remove it from the changelog and the repository for now, so that we don't add it to our next stable release accidentally. Perhaps someone interested can re-add this package.

cc: @Starz0r

### Checklists


hoodmane (approved):
Thanks @ryanking13 this has confused me several times.

rth:
> so our build system didn't build that package.

Why did we have the output wheel in the dist folder then in the commit before this one (cf build-packages CI),
```
1.1M	dist/pycryptodome-3.15.0-cp35-abi3-emscripten_3_1_25_wasm32.whl
1.3M	dist/pycryptodome-tests.tar
```
(I also had trouble with it because it's an abi3 wheel, and assumed that cp35 was OK in this case, but maybe I'm wrong).

rth:
OK, different name, makes sense )

hoodmane:
Yeah the fact that it's one letter different really confused me more than once too. This is kind of funny.

The abi3 rule is the cp35 has to be <= current interpreter (idea being that things are added to stable ABI but not removed or changed).

PR #3263, opened 2022-11-16
Use a more robust method to improve our ModuleNotFound errors

hoodmane:
@ryanking13 added these very nice error messages to the `ModuleNotFound` errors. However they introduce a few problems:
1. `find_spec` is supposed to return `None` or a spec and not to raise an error. If it raises errors, it can cause trouble in code that tries to check if a module is installed or not.
2. Other code that tries to add new import hooks has to know to insert them before these error-raising import hooks.

See the discussion in #3262.

This instead patches `importlib._bootstrap` to create a function called `_get_module_not_found_error`. We then can monkey patch this to modify the error messages that `importlib` raises.

### Checklists

- [x] Add / update tests


hoodmane:
I also made a thread about this on discuss.python.org:
https://discuss.python.org/t/a-hook-to-allow-adding-notes-to-modulenotfounderror/21151/2

ryanking13:
Thanks for the fix @hoodmane!

How about modifying `sys.excepthook`, something like this? (I have never tried, just an idea).

```
def pyodide_except_hook(type, value, traceback):
    orig_excepthook = sys.excepthook
    if isinstance(type, ModuleNotFoundError):
        ...
    else:
        return orig_excepthook(...)

sys.excepthook = pyodide_except_hook
```

I think this won't require patching CPython.

hoodmane:
We currently never call `sys.excepthook`. Instead of that, we turn errors into JavaScript exceptions when they reach the bottom of the stack.

ryanking13:
Oh, okay. You are right. I was able to check that `sys.excepthook` is never called.

rth:
No objection to this change, but it would be worth opening a feature request at Cpython to make it possible change this message  without patching. Or if they don't want to update the public API of importlib for this, maybe at least they would be OK with adding this patch upstream so that it can be modified via monkeypatching (instead of real patching).

hoodmane:
> the patch to Python/importlib.h in particular looks a bit annoying to maintain.

Well it's autogenerated, you just have to run `make regen-importlib`.

PR #3264, opened 2022-11-17
Support loading shared libraries inside a wheel file globally

### LINKED ISSUE #3167: Embed external shared libraries into the wheel

ryanking13:
### Description

Resolve #3167 
Close #3183 

This is a final split-up of #3183.

This PR mostly does two things:

1. Add a way to detect whether libraries inside the wheel should be loaded globally or not: If a library is needed by another library inside a wheel file, load it globally. 

2. Apply `vendor-sharedlib` (#3234) keys to some packages that depend on other shared libraries: fiona (gdal), pyheif (libheif), h5py (libhdf5), shapely (geos).

Note that @rth had some concerns about downloading the same library multiple times (#3167). For now I applied it to the shared libraries that are used in only one or two packages, but we can discuss more about this.

> For packages we built in tree, I'm less sure, as this would lead to potentially redundant downloads (e.g. BLAS downloaded multiple times for numpy and scipy once we switch to a better BLAS). Though it would be worth analysing how many dependent packages each shared library has in the current dependency graph, and whether indeed vendoring it inside the wheel would be that significant.

### Checklists


packages/shapely/patches/0001-Search-libs-from-auditwheel-repaired-directory.patch
```diff
@@ -0,0 +1,24 @@
...
+diff --git a/shapely/geos.py b/shapely/geos.py
+index 88c5f53..fef5cbd 100644
+--- a/shapely/geos.py
++++ b/shapely/geos.py
+@@ -62,7 +62,7 @@ def exists_conda_env():
+     """Does this module exist in a conda environment?"""
+     return os.path.exists(os.path.join(sys.prefix, 'conda-meta'))
+
+-if sys.platform.startswith('linux'):
++if sys.platform.startswith(('linux', 'emscripten')):
```
hoodmane, line 19: Could you try to upstream this?

src/js/dynload.ts
```diff
@@ -97,6 +114,19 @@ export async function loadDynlib(lib: string, global: boolean) {
...
+
+    // Emscripten saves the list of loaded libraries in LDSO.loadedLibsByName.
+    // However, since emscripten dylink metadata only contains the name of the
+    // library not the full path, we need to update it manually in order to
+    // prevent loading same library twice.
+    if (global && Module.PATH.isAbs(lib)) {
+      const libName: string = Module.PATH.basename(lib);
+      const dso: any = Module.LDSO.loadedLibsByName[libName];
+      if (!dso) {
+        Module.LDSO.loadedLibsByName[libName] =
```
hoodmane, line 126: Funny we used to have to do the same thing with `preloadedWasm`, then `preloadedWasm` got removed and now we have to add the same hack with the new `loadedLibsByName`.

src/js/dynload.ts
```diff
@@ -110,4 +140,84 @@ export async function loadDynlib(lib: string, global: boolean) {
...
+    const global = globalLibs.has(Module.PATH.basename(path));
+    return {
+      path: path,
+      global: global || !!pkg.shared_library,
+    };
+  });
+
+  // Sort libraries so that global libraries can be loaded first.
+  // TODO(ryanking13): It is not clear why global libraries should be loaded first.
+  //    But without this, we get a fatal error when loading the libraries.
```
hoodmane, line 185: I guess it's just lucky we understand what's going on as much as we do.

src/js/dynload.ts
```diff
@@ -110,4 +140,84 @@ export async function loadDynlib(lib: string, global: boolean) {
...
+  dynlibs.sort((lib: { global: boolean }) => (lib.global ? -1 : 1));
```
hoodmane, line 186: This is a bit wasteful I think, better would be:
```suggestion
type T = { global: boolean };
dynlibs.sort((lib1: T, lib2 : T) => (lib2.global - lib1.global));
```

src/js/dynload.ts
```diff
@@ -110,4 +140,84 @@ export async function loadDynlib(lib: string, global: boolean) {
...
+  }
+
+  const globalLibs = new Set<string>();
+
+  libs.forEach((lib: string) => {
+    const binary = readFile(lib);
+    const needed = Module.getDylinkMetadata(binary).neededDynlibs;
+    needed.forEach((lib: string) => {
+      globalLibs.add(lib);
+    });
```
hoodmane, line 216: Is this a conservative overestimate of the libs that need to be loaded globally? Or is this exactly the condition? Also, does this code also run for shared libraries that are loaded separately? It's fairly efficient to turn an originally local library into a global one when it turns out to be needed later. Do we handle this case?

In any case, it's really nice to see these improvements.

hoodmane (approved):
Great work @ryanking13!

src/js/dynload.ts
ryanking13: Yeah, while writing this PR, I noticed that `preloadedWasm` hack was doing the same thing... my bad.

packages/shapely/patches/0001-Search-libs-from-auditwheel-repaired-directory.patch
ryanking13: shapely 2.0 releases soon, and I think we won't require this patch anymore after we update to 2.0 (they no more loads `libgeos.so` directly)

src/js/dynload.ts
hoodmane: Well this is an improvement in a lot of other ways over using `preloadedWasm`, and at least you tried removing the hack and learned that it wouldn't go away.

src/js/dynload.ts
ryanking13: > Is this a conservative overestimate of the libs that need to be loaded globally? Or is this exactly the condition?

Since `neededDynlibs` contains all dynamic libraries linked during the compilation (via `-l` flag), I think this is not an overestimate unless we find a way to load symbols of external libraries loaded locally.

On the other hand, this can rather be an "underestimate" of libs, because users who write compile commands might omit to specify dynamic libraries that should be linked. For example: in #3234, I found and fixed that our `sharedlib-test-py` package [wasn't linked to `sharedlib-test.so` correctly](https://github.com/pyodide/pyodide/pull/3234/files#diff-ac3008498f6901cf8542ab2b87911b5477946de93a7babe0ded3279c912887f7).

> Also, does this code also run for shared libraries that are loaded separately?

Good point, I think we can skip this part for shared libraries that are loaded separately.

> It's fairly efficient to turn an originally local library into a global one when it turns out to be needed later. Do we handle this case?

No, I think we never turn a local library into a global one for now. Since `global` flag is propagated to dependencies when calling `loadDynamicLibrary`, a locally loaded dynamic library does not load its dependencies globally. 

(I'm not sure this behavior is a bug in Emscripten, probably I should open an issue in Emscripten).

ryanking13: > (I'm not sure this behavior is a bug in Emscripten, probably I should open an issue in Emscripten).

Opened an issue in Emscripten (https://github.com/emscripten-core/emscripten/issues/18264). If it is accepted, it might be possible for us to remove whole global checks.

ryanking13:
Thanks for the review!

PR #3270, opened 2022-11-19
DOCS: Get rid of those damn parentheses at the end of type names

hoodmane:
This has been bothering me for a long time. Fix discovered by trial and error. The parens are getting added here:
https://github.com/sphinx-doc/sphinx/blob/master/sphinx/roles.py#L87


hoodmane:
Seems to me like this may be a sphinx bug.

ryanking13 (approved):

Issue #3267, opened 2022-11-17
JsProxy: support element deletion


rth:
Currently from Python,
```py
>>> import pyodide_js
>>> del pyodide_js.loadedPackages.micropip
```
works, same as from JavaScript,
```js
delete pyodide.loadedPackages['micropip']
```

However the following fails,
```py
>>> import pyodide_js
>>> del pyodide_js.loadedPackages['micropip']
Traceback (most recent call last):
  File "<console>", line 1, in <module>
TypeError: 'pyodide.JsProxy' object does not support item deletion
```

It could be nice if del also worked on the second case.

hoodmane:
It used to work this way. I wanted to save `__delitem__` for calling the `delete` method since JavaScript containers typically use `o.delete(x)`. So this way `__delitem__` works correctly on `Set` and `Map` and similar objects. It also works in the expected way on arrays (it calls `splice`).

The issue is that there is no idiomatic JavaScript way of declaring an object as a container. If it has no length and is not iterable, then it doesn't seem like a container to me and hence implementing container methods can cause confusion when various Python internals do reflection and decide the object is a container.

I could see adding some sort of internal marker or adaptor to say that an object is being used as a map. Or maybe we should change `loadedPackages` into a JavaScript `Map`.

rth:
Thanks for the explanations! The motivation for the current situation sounds very reasonable, I'm not saying we need to do anything about this, I was mostly curious.

> The issue is that there is no idiomatic JavaScript way of declaring an object as a container.

Yes, I can see how that could be a problem. Many JS users would probably use objects instead of map for containers. 

Maybe we could add a note about it https://pyodide.org/en/stable/usage/type-conversions.html. Also I guess now, people searching for this exception will end up here )

hoodmane:
I think we could add an internal `JsProxy` flag to treat the object as a container and a method to create a new proxy with that flag set. If you really wanted to make it a property of the JavaScript object and not of the proxy, we could tag the object with a Symbol but I think that introduces some tricky questions. But the internal JsProxy flag would be a reasonable route. `PyProxy` already has the `PyProxyProps` flags for handling `bind`, `captureThis`, and `roundtrip`.

rth:
> I think we could add an internal JsProxy flag to treat the object as a container and a method to create a new proxy with that flag set.

OK, but how would determine if it's a container or not to set that flag?

hoodmane:
There would be a method that copies the JsProxy and sets the flag on the copy.

PR #3273, opened 2022-11-19
Add `as_object_map` method to JsProxy

hoodmane:
This adds a method which returns a new JsProxy that acts like a container of the JavaScript object's `ownProperties`.

See #3267

### Checklists

- [x] Add / update tests

### DIFF
--- a/src/core/jsproxy.c
+++ b/src/core/jsproxy.c
@@ -1610,10 +1610,10 @@ JsObjMap_ass_subscript_js,
 {
   let obj = Hiwire.get_value(idobj);
   let key = Hiwire.get_value(idkey);
-  if (!Object.prototype.hasOwnProperty.call(obj, key)) {
-    return -1;
-  }
   if(idvalue === 0) {
+    if (!Object.prototype.hasOwnProperty.call(obj, key)) {
+      return -1;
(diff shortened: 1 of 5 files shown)


src/core/jsproxy.c
```diff
@@ -1500,6 +1503,174 @@ static PyMethodDef JsProxy_finally_MethodDef = {
...
+  return result;
+}
+
+EM_JS_NUM(int, JsObjMap_length_js, (JsRef idobj), {
+  let jsobj = Hiwire.get_value(idobj);
+  let length = 0;
+  for (let _ of Module.iterObject(jsobj)) {
+    length++;
+  }
+  return length;
```
ryanking13, line 1547: Maybe we need to mention somewhere that it takes `O(n)` for `len(jsproxy_object_map)`?

ryanking13 (approved):
Thanks, @hoodmane! It looks really cool.

src/core/jsproxy.c
hoodmane: Yeah perhaps this is worth pointing out. This is a good reason to use a `Map` instead =)

hoodmane:

PR #3275, opened 2022-11-20
Implement Mapping / MutableMapping methods for JsProxy of a Map

hoodmane:
This adds methods `keys`, `items`, `values`, `get`, `pop`, `setdefault`, `popitem`, `update`, and `clear` to a `JsProxy` of a map-like object. Since both Sequence-like objects and Map-like objects have a `pop` method but they have different signatures, I had to split the `JsProxy` documentation class into several different classes. This is probably a good idea anyways. As a followup I could set it up so that `isinstance` and `issubclass` work correctly on these documentation classes.

### Checklists

- [x] Add / update tests


ryanking13 (approved):
Thanks @hoodmane!

hoodmane:

PR #3279, opened 2022-11-21
Split building and installing packages in buildall

ryanking13:
### Description

This PR splits 1) building packages and 2) copying built packages into dist directory and creating repodata.json. 

Note that `pyodide-build buildall` command is not affected, while `pyodide build-recipes` command now accepts a new flag `--install`.

The purpose of this PR is to merge `buildpkg` and `buildall` into a single api `pyodide build-recipes`, so we can use such as:

- `pyodide build-recipes scipy`
- `pyodide build-recipes scipy --no-deps` (as suggested by @rth in #2879)

### Checklists

- [x] Add / update tests


packages/Makefile
```diff
@@ -7,6 +7,8 @@ all: pyodide-build
 	mkdir -p $(HOSTINSTALLDIR) $(WASM_LIBRARY_DIR)
 	PYODIDE_ROOT=$(PYODIDE_ROOT) pyodide build-recipes \
 	"$(PYODIDE_PACKAGES)" \
+	--install \
+	--install-dir=../dist \
```
hoodmane, line 11: Would it make sense to combine these into one flag? Or is there a default `install-dir`?

hoodmane (approved):
Looks good to me.

packages/Makefile
ryanking13: >  Or is there a default install-dir?

I set default install dir to `$(pwd)/dist` (not PYODIDE_ROOT/dist because I want it to make it work out-of-tree someday).

> Would it make sense to combine these into one flag?

I didn't want to mix boolean and string flag (and I guess it is pretty hard to do so with typer).

ryanking13:

Issue #3211, opened 2022-10-30
Ability to run user's python code equivalent to using -OO option


nmstoker:
## 🚀 Feature

With the regular Python interpreter, you can call it with the flag -OO that optimises running the user's code by ignoring asserts and dropping docstring.

If this isn't wildly inappropriate for Pyodide (for some reason I've overlooked) could this be considered for implementation?

### Motivation

By intentionally skipping assetts code is saved from running and this can be both useful and would result in a speed up. Having an equivalent for Pyodide would make it somewhat consistent with the regular interpreter.

### Pitch

Seems like it's worth discussion at least 🙂

### Alternatives

Drop this suggestion if it just doesn't make sense or isn't much of a priority 

### Additional context

Documentation on the option is here:

https://docs.python.org/3/using/cmdline.html#cmdoption-OO

hoodmane:
I think this should be possible in dev Pyodide by passing
`loadPyodide({args:["-OO"]})`. All arguments in the `args` list are passed
unchanged to the Python interpreter which can do what it likes with them.

ryanking13:
Closing as resolved.

PR #3282, opened 2022-11-23
Fix docstring error when loadPyodide is called with -OO option

ryanking13:
### Description

Related: #3211

When `loadPyodide` is called with `-OO` ({args: ["-OO"]}), which strips out all docstrings, it fails with error:

image

### Checklists

- [x] Add / update tests


hoodmane (approved):
Thanks @ryanking13!

ryanking13:
Thanks for the review!

Issue #3203, opened 2022-10-24
Fatal error with snippet using `np.random` and `scipy.linalg`


lesteve:
## 🐛 Bug

I get a fatal error when trying to run a snippet using `numpy.random` and `scipy.linalg` inside node.

I get somewhat similar issues when trying to run the snippet in Chromium or Firefox.

Note the behaviour is not always the same on each run. In particular, the error does not happen on the same loop iteration. I have seen this kind of variations:
- Pyodide fatal error: `memory access out of bounds` or `table index out of bounds` or `null function or function signature mismatch`
- weird Python error that does not make sense (since the code works in previous loop iterations) e.g. `TypeError: my_make_regression() got multiple values for argument 'n_features'` or `SystemError: Type does not define the tp_name field.`
- in Chromium: the tab crashes with a "Aw snap" message
- what seems like a hang, i.e. the loop counter is not printed anymore

I tried to get more info by building Pyodide myself with `-g2` flags on commit https://github.com/pyodide/pyodide/commit/2326c05d83bb172d3a8b3d267d3440ed687f5c8d. Out of 20 runs, here is the kind of stack-trace I got (with some kind of grouping):

<details>
<summary>
memory access out of bounds dlfree (3 times)
</summary

```
Stack (most recent call first):
  File "<exec>", line 9 in my_make_regression
  File "<exec>", line 5 in <module>
  File "/lib/python3.10/_pyodide/_base.py", line 357 in run_async
  File "/lib/python3.10/_pyodide/_base.py", line 531 in eval_code_async
  File "/lib/python3.10/asyncio/tasks.py", line 232 in __step
  File "/lib/python3.10/asyncio/events.py", line 80 in _run
  File "/lib/python3.10/pyodide/webloop.py", line 267 in run_handle
/home/lesteve/dev/scikit-learn-tests-pyodide/node_modules/pyodide/pyodide.asm.js:452
      throw ex;
      ^

RuntimeError: memory access out of bounds
    at dlfree (wasm://wasm/01f433ea:wasm-function[8627]:0x3a2b27)
    at _PyMem_RawFree (wasm://wasm/01f433ea:wasm-function[2006]:0x15b64f)
    at PyObject_Free (wasm://wasm/01f433ea:wasm-function[2038]:0x15c899)
    at object_dealloc (wasm://wasm/01f433ea:wasm-function[2259]:0x16cad1)
[...]

PR #3284, opened 2022-11-23
Update numpy to v1.23.5

lesteve:
### Description

This PR updates numpy to its latest released version 1.23.5. Currently the numpy version is 1.22.4 and there are Pyodide-related fixes that made it to the 1.23 version. For a particular that I bumped into, see https://github.com/pyodide/pyodide/issues/3203#issuecomment-1290551624.

### Checklists

- [x] Add / update tests


lesteve:
Hmmm it seems like there is a failure in matplotlib tests only for browsers but not for node. If you have any hints on this, they would be more than welcome!

If not, I will try to investigate and reproduce the issue locally!

rth:
Thanks @lesteve ! There is a 66MB core dump we probably want to exclude in this PR ^^

Yes, the matplotlib failure is not very explicit, reproducing locally would certainly be useful!

lesteve:
I can reproduce locally with the following snippet:
```py
import matplotlib

matplotlib.use("module://matplotlib_pyodide.html5_canvas_backend")
import datetime

import numpy as np
from matplotlib.dates import (
    YEARLY,
    DateFormatter,
    RRuleLocator,
    drange,
    rrulewrapper,
)

# tick every 5th easter
np.random.seed(42)
rule = rrulewrapper(YEARLY, byeaster=1, interval=5)
loc = RRuleLocator(rule)
formatter = DateFormatter("%m/%d/%y")
date1 = datetime.date(1952, 1, 1)
date2 = datetime.date(2004, 4, 12)
delta = datetime.timedelta(days=100)

dates = drange(date1, date2, delta)
s = np.random.rand(len(dates))  # make up some random y values

fig, ax = plt.subplots()
plt.plot_date(dates, s)
ax.xaxis.set_major_locator(loc)
ax.xaxis.set_major_formatter(formatter)
labels = ax.get_xticklabels()
```

The error I get:
```
pyodide.asm.js:10 Uncaught RuntimeError: null function or function signature mismatch
    at PyArray_CastDescrToDType (009c0826:0x548eb)
    at PyArray_PromoteTypes (009c0826:0x55829)
    at PyUFunc_AdditionTypeResolver (009c0826:0x192b54)
    at resolve_descriptors (009c0826:0x171c5d)
    at ufunc_generic_fastcall (009c0826:0x16dd97)
    at ufunc_generic_vectorcall (009c0826:0x16cd36)
    at pyodide.asm.wasm:0x11bc14
    at pyodide.asm.wasm:0x11bd8a
    at array_add (009c0826:0x10eed3)
    at gentype_add (009c0826:0x11a7c8)
```

Now my problem is that when I click on the address to go to the code, it crashes Chromium :cry:

hoodmane:
@lesteve can you check if the reproduction I posted here works?
https://bugs.chromium.org/p/chromium/issues/detail?id=1393046
If so, would you mind adding a comment there saying that the issue reproduces for you and which chrome version you are on?

lesteve:
> This is a regression in recent chromiums. Try it in a version < 103 or so.

And now I am on a wild goose chase trying to figure out how to install an old version with snap on Ubuntu ...

hoodmane:
Try here:
https://dl.google.com/linux/chrome/deb/pool/main/g/google-chrome-stable/google-chrome-stable_99.0.4844.84-1_amd64.deb

hoodmane:
>  trying to figure out how to install an old Chromium version

it's surprisingly difficult...

lesteve:
Thanks, with an older Chromium version, I can debug further and the issue is in
the `datetime_to_timedelta_resolve_descriptors` numpy function:
The issue seems to be in the https://github.com/numpy/numpy/blob/22e683d84f2584a6f9a57b2c58d1def6743fcf83/numpy/core/src/multiarray/datetime.c#L3903

lesteve:
Not sure what is going on here, it seems like in numpy 1.22 the C function had only 4 parameters but has 5 parameters in numpy 1.23. Somehow the wasm still thinks the function has 4 parameters.

I tried `make clean` just in case but I still get the same error.

hoodmane:
We need to cherry-pick this patch:
https://github.com/numpy/numpy/pull/21894

lesteve:
Thanks somehow I did not search for it correctly in the numpy repository. I think the easiest thing to do is to have a patch in pyodide. I pushed commit and the CI is green :tada: 

Just curious:
- any hinsights into why was it working inside node and not inside browers?
- do you have a tool to generate the patch conveniently from a git commit (I did `git show <commit> > my-file.patch` but maybe there , I have seen that there is a footer like this at the end of most of the patches I have looked at but I am not sure where it is coming from (maybe the git version?):
  ```
  -- 
  2.25.1
  ``` 
- is there a convenient way to run pyodide on a source folder and still use `make`, or is using `pyodide build` the most convenient in this case

hoodmane:
> insights into why was it working inside node and not inside browers?

We don't run the matplotlib test in node at all since we don't have a working backend in node.

> do you have a tool to generate the patch conveniently from a git commit?

I like to use 
```sh
git checkout <COMMIT>
git format patch HEAD~1 ~/pyodide/packages/numpy/patches
```
or often I'll keep a branch named `v1.23.5-pyodide` and commit all the patches I want there, then I can update them all at once via:
```sh
rm ~/pyodide/packages/numpy/patches/*
git format-patch v1.23.5 ~/pyodide/packages/numpy/patches/*
```
though then you have to update the `meta.yaml` file for which I use:
```sh
cd ~/pyodide/packages/numpy
ls -1 patches
```
To update to a new version you can use:
```sh
git checkout -b v1.24.0-pyodide
git rebase v1.23.5 --onto v1.24.0
```

> is there a convenient way to run pyodide on a source folder and still use make, or is using pyodide build (and copying the .whl in dist) the most convenient in this case?

Using `pyodide build` is the thing that I have currently set up. There is a reasonable way to set it up to allow using `make`, if you have a use case for it I could consider adding a feature like that. In that case you would be responsible for ensuring that the finished wheel ends up in the `dist` folder since there is no way to know in general where `make` will put build artifacts.

hoodmane:
Thanks @lesteve!

ryanking13:
Thanks, @lesteve!

> * any hinsights into why was it working inside node and not inside browers?

The failing matplotlib test is skipped in node (html5_backend of matplotlib doesn't work in node AFAIK)

> do you have a tool to generate the patch conveniently from a git commit (I did git show <commit> > my-file.patch but maybe there , I have seen that there is a footer like this at the end of most of the patches I have looked at but I am not sure where it is coming from (maybe the git version?):

I use [`git format-patch ...`](https://git-scm.com/docs/git-format-patch). See also: [Generating patches](https://pyodide.org/en/stable/development/new-packages.html#generating-patches) (Hood wrote it).

> is there a convenient way to run pyodide on a source folder and still use make, or is using pyodide build (and copying the .whl in dist) the most convenient in this case?

If you need some trial and error to build and patch package, I normally just use `make` + [`pyodide-build buildpkg --continue`](https://pyodide.org/en/stable/development/new-packages.html#partial-rebuilds), but Hood might know a better way than me

lesteve:
Thanks a lot to both of you for the tips!

Great to see this merged too!

hoodmane:
> I normally use make + [pyodide-build buildpkg --continue](https://pyodide.org/en/stable/development/new-packages.html#partial-rebuilds), but Hood might know a better way than me

It is possible to leak the symlink directory by inserting an extra line here:
https://github.com/pyodide/pyodide/blob/main/pyodide-build/pyodide_build/pywasmcross.py#L141
setting `symlink_dir_str` to some fixed thing. Then if you run `buildpkg` once and it crashes, you can do `export PATH=SYMLINK_DIR:"$PATH"` and then `cd packages/numpy/build/numpy-1.23.5` and then you can copy commands out of the build log and they will behave exactly the same way they do inside of pywasmcross.

PR #3289, opened 2022-11-25
Add type field to PythonError

hoodmane:
This adds a new `error.type` field to `PythonError` exceptions which contains `type(e).__qualname__`.

### Checklists

- [x] Add a [CHANGELOG](https://github.com/pyodide/pyodide/blob/main/docs/project/changelog.md) entry
- [x] Add / update tests
- [x] Add new / update outdated documentation

PR #3290, opened 2022-11-25
Add support to PyProxy for using async python generators from js

hoodmane:
This is on top of #3289. It adds support for using async Python generators via PyProxy.

### Checklists

- [x] Add / update tests

### DIFF
--- a/src/core/pyproxy.ts
+++ b/src/core/pyproxy.ts
@@ -17,6 +17,7 @@
 declare var Module: any;
 declare var Hiwire: any;
 declare var API: any;
+declare var HEAPU32: Uint32Array;
 
 // pyodide-skip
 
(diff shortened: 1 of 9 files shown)


src/core/pyproxy.ts
```diff
@@ -938,6 +1026,43 @@ export class PyProxyIteratorMethods {
   }
 }
 
+export class PyProxyAsyncIteratorMethods {
+  async next(arg: any = undefined): Promise<IteratorResult<any, any>> {
```
ryanking13, line 1030: It seems like `PyProxy.next` is shown twice in docs (async and sync version), but I'm not sure what we should do. So I'm okay to merge.

ryanking13 (approved):
Thanks @hoodmane!

src/core/pyproxy.ts
hoodmane: I think we can add `/** @private */` to one.

hoodmane: In the long run it'd probably be good to break up the `PyProxy` docs like the way I did to the `JsProxy` docs in #3277.

ryanking13: > I think we can add `/** @private */` to one.

That makes sense. Thanks!

hoodmane: Of course the docs aren't strictly accurate since `next` might instead translate to Python `anext`. But it is okay.

hoodmane:

PR #3297, opened 2022-11-28
Add test for MutableMapping methods on object_maps and fix bug

hoodmane:
This adds tests of all the MutableMapping methods for objects returned by `as_object_map`. They all worked but I was accidentally adding only the non mutating methods. I also added `JsMap` to `pyodide.ffi` and fixed it so that `isinstance` works correctly and mypy agrees with the usage.

### Checklists


src/py/_pyodide/_core_docs.py
```diff
@@ -441,6 +447,38 @@ def reverse(self) -> None:
 
 
 class JsMap(JsProxy):
```
hoodmane, line 449: One question: perhaps we should have a `JsMap` and a `JsMutableMap`?

ryanking13: I think it is a good idea.

src/py/_pyodide/_core_docs.py
```diff
@@ -441,7 +448,35 @@ def reverse(self) -> None:
...
+
+    _js_type_flags = ["HAS_GET | HAS_LENGTH | IS_ITERABLE", "IS_OBJECT_MAP"]
+
+    def __getitem__(self, idx: Any) -> Any:
+        return None
+
+    def __setitem__(self, idx: Any, value: Any) -> None:
+        pass
+
+    def __delitem__(self, idx: Any) -> None:
```
ryanking13, line 468: Isn't `__setitem__` and `__delitem__` for `JsMutableMap`?

ryanking13 (approved):
Thanks, @hoodmane!

src/py/_pyodide/_core_docs.py
hoodmane: Yes good point.

PR #3298, opened 2022-11-29
Add a typeshed for the js module

hoodmane:
These are the changes I made in order to get the newest pre-commit #3254 to pass. This significantly improves the discernment of the type system.

The one pain point is that mypy does not seem to understand the behavior of `then`: if the `onfulfilled` function returns an awaitable, then it awaits the result. Thus, we shouldn't get `PyodideFuture[Coroutine[str]]` it should instead be collapsed into `PyodideFuture[str]`. I told it that the `onfulfilled` argument should have type `Callable[[T], S | Awaitable[S]]` but the type system seems to get confused about it. Though Microsoft's PyLance type checker understands it fine. If I tell it the type of an intermediate value it figures it out:
```js
# mypy isn't smart enough to typecheck this without assistance
b: PyodideFuture[JsProxy]
b = fetch(url).then(get_json)
name = await b.then(get_name)
reveal_type(name) # it's a str!
```
@henryiii is there a way to deal with this or is it a known mypy limitation?

### Checklists


henryiii:
This might be an issue with the lack of higher-kinded types. However, have you tried an overload?

hoodmane:
Is there a way to ask for S to be not awaitable?

henryiii:
```python
import asyncio
from collections.abc import Awaitable
from typing import TypeVar

T = TypeVar("T")

async def then(x: T | Awaitable[T]) -> T:
    if isinstance(x, Awaitable):
        return await x
    return x

async def g1() -> int:
    await asyncio.sleep(1)
    return 1

def g2() -> int:
    return 2

async def main() -> None:
    print(await then(g1()))
    print(await then(g2()))


if __name__ == "__main__":
    asyncio.run(main())
```

This fails mypy. If you bind to an `int`, for example, it works.

hoodmane:
Thanks for looking into it @henryiii!

henryiii:
Here's the fix (not sure how you want it, so just pasting it as a patch):

<details>
<summary>Patch:</summary>

```diff
diff --git a/src/py/pyodide/webloop.py b/src/py/pyodide/webloop.py
index 278dae7f..d163298c 100644
--- a/src/py/pyodide/webloop.py
+++ b/src/py/pyodide/webloop.py
@@ -6,7 +6,7 @@ import time
 import traceback
 from asyncio import Future
 from collections.abc import Awaitable, Callable
-from typing import Any, TypeVar
+from typing import Any, TypeVar, overload
 
 from ._core import IN_BROWSER, create_once_callable
 
@@ -22,6 +22,34 @@ class PyodideFuture(Future[T]):
     Javascript promise API.
     """
 
+    @overload
+    def then(
+        self,
+        onfulfilled: None,
+        onrejected: Callable[[BaseException], Awaitable[S]],
+    ) -> "PyodideFuture[S]": ...
+
+    @overload
+    def then(
+        self,
+        onfulfilled: None,
+        onrejected: Callable[[BaseException], S],
+    ) -> "PyodideFuture[S]": ...
+
+    @overload
+    def then(
+        self,
+        onfulfilled: Callable[[T], Awaitable[S]],
+        onrejected: Callable[[BaseException], Awaitable[S]] | None = None,
+    ) -> "PyodideFuture[S]": ...
+
+    @overload
+    def then(
+        self,
+        onfulfilled: Callable[[T], S],
+        onrejected: Callable[[BaseException], S] | None = None,
+    ) -> "PyodideFuture[S]": ...
+
     def then(
         self,
         onfulfilled: Callable[[T], S | Awaitable[S]] | None,
@@ -90,9 +118,21 @@ class PyodideFuture(Future[T]):
         self.add_done_callback(wrapper)
         return result
 
+    @overload
+    def catch(
+        self, onrejected: Callable[[BaseException], Awaitable[S]]
+    ) -> "PyodideFuture[S]":
+        ...
+
+    @overload
     def catch(
-        self, onrejected: Callable[[BaseException], S | Awaitable[S]] | None
+        self, onrejected: Callable[[BaseException], S]
     ) -> "PyodideFuture[S]":
+        ...
+
+    def catch(
+        self, onrejected: Callable[[BaseException], object]
+    ) -> "PyodideFuture[Any]":
         return self.then(None, onrejected)
 
     def finally_(self, onfinally: Callable[[], None]) -> "PyodideFuture[T]":
diff --git a/src/tests/test_webloop.py b/src/tests/test_webloop.py
index 0b017964..dbd31618 100644
--- a/src/tests/test_webloop.py
+++ b/src/tests/test_webloop.py
@@ -352,7 +352,6 @@ async def test_pyodide_future2(selenium):
 
     url = "https://pypi.org/pypi/pytest/json"
     # mypy isn't smart enough to typecheck this without assistance
-    b: PyodideFuture[JsProxy]
     b = fetch(url).then(get_json)
     name = await b.then(get_name)
     assert name == "pytest"
```

</details>

ryanking13 (approved):
Thanks a lot!

rth (approved):
Very nice indeed! Thanks!

PR #3300, opened 2022-11-29
Add `requirements/executable` key in meta.yaml spec

### LINKED ISSUE #3237: Specify non-python build requirements in meta.yaml
### Motivation

Some packages require extra executables during the build. For example:

- scipy: gfortran, f2c
- cryptography: cargo
- libproj: sqlite3
- nlopt: swig

We are listing these requirements in our [docs](https://pyodide.org/en/stable/development/building-from-sources.html), but (I guess) most people don't need to build complete package sets locally, so it is not very useful for new people who want to build Pyodide locally (and it might make people freak out: "So many requirements to build!").

Actually, to build Pyodide with core packages (in Linux), we only need some [essential build tools + Python and Node.js](https://github.com/pyodide/pyodide/blob/af206d159cc127920b191e1bbfd75dfbef5f981d/.github/workflows/main.yml#L107).
Note that for now `make` fails [if `gfortran` or `f2c` not exists](https://github.com/pyodide/pyodide/blob/af206d159cc127920b191e1bbfd75dfbef5f981d/tools/dependency-check.sh#L43-L46), even if we don't build scipy.

So I think it would be nice to have a way to specify and check per-package requirements, instead of listing all of them in docs.

### Feature

Add `requirements/executable` key (open to other names), which specifies extra executables required to build a package.

During the build, it can be easily checked via `shutil.which("...")` and raise an error if those executables are unavailable.

### Alternatives

If adding a new key is not preferable, we can just use `which <executable>` inside `build/script` instead.

ryanking13:
### Description

Close #3237 

Adds a new key in meta.yaml spec: `requirements/executable` which specifies the list of executables required to build a package.
Unlike conda, we don't build or install these executables, these key exists just to halt build earlier

### Checklists

- [x] Add / update tests


pyodide-build/pyodide_build/buildall.py
```diff
@@ -272,6 +275,20 @@ def generate_dependency_graph(
...
 
+    # Check executables required to build packages
+    missing_executables = defaultdict(list)
+    for name in requested:
+        pkg = pkg_map[name]
+        for exe in find_missing_executables(pkg.executables_required):
+            missing_executables[exe].append(name)
+
+    if missing_executables:
+        error_msg = "Following executables are missing in the host system:\n"
```
hoodmane, line 286:
```suggestion
        error_msg = "The following executables are missing in the host system:\n"
```

pyodide-build/pyodide_build/buildpkg.py
```diff
@@ -927,6 +927,11 @@ def main(args: argparse.Namespace) -> None:
     pkg.build.cxxflags += f" {args.cxxflags}"
     pkg.build.ldflags += f" {args.ldflags}"
 
+    missing_executables = find_missing_executables(pkg.requirements.executable)
+    if missing_executables:
+        error_msg = f"Following executables are required but missing in the host system: {', '.join(missing_executables)}"
```
hoodmane, line 932: To keep the line a bit shorter maybe:
```suggestion
        missing_string = ", ".join(missing_executables)
        error_msg = (
            f"The following executables are required but missing in the host system: "
            + missing_string
        )
```

hoodmane (approved):
Thanks @ryanking13!

ryanking13:

PR #3301, opened 2022-11-29
NFC Add getTypeTag helper function

hoodmane:
We use `Object.prototype.toString.call` a lot so I thought it would be good to add a helper function for this.


ryanking13 (approved):
Thanks!

Issue #2006, opened 2021-11-28
Issues building pyodide on Ubuntu 21.10


wlach:
## 🐛 Bug

Running `make` on Ubuntu 21.10, I get the following error:

```
gmake[4]: Leaving directory '/home/wlach/src/pyodide/emsdk/emsdk/ccache/git-emscripten_64bit/build_emscripten_64'
[ 85%] Built target ccache
[ 85%] Building CXX object unittest/CMakeFiles/unittest.dir/test_Config.cpp.o
[ 86%] Building CXX object unittest/CMakeFiles/unittest.dir/test_Counters.cpp.o
[ 87%] Building CXX object unittest/CMakeFiles/unittest.dir/test_Depfile.cpp.o
[ 89%] Building CXX object unittest/CMakeFiles/unittest.dir/test_FormatNonstdStringView.cpp.o
[ 89%] Building CXX object unittest/CMakeFiles/unittest.dir/test_Hash.cpp.o
[ 90%] Building CXX object unittest/CMakeFiles/unittest.dir/test_Lockfile.cpp.o
In file included from ../../src/unittest/main.cpp:27:
../../src/src/third_party/doctest.h:4189:47: error: size of array ‘altStackMem’ is not an integral constant-expression
 4189 |         static char             altStackMem[4 * SIGSTKSZ];
      |                                               ^
gmake[4]: *** [unittest/CMakeFiles/unittest.dir/build.make:95: unittest/CMakeFiles/unittest.dir/main.cpp.o] Error 1
```

It appears the problem has to do with the doctest.h file included in the version of ccache used to build emscripten (`./emsdk/emsdk/ccache/git-emscripten_64bit/src/src/third_party/doctest.h`). This issue seems to have been fixed upstream (https://github.com/onqtam/doctest/issues/473), though does not seem to have made its way into the version of ccache used by emscripten (https://github.com/juj/ccache/tree/emscripten)

wlach:
I think I've fixed this problem in a fork, but might be best to wait to see if upstream (which is also a fork) accepts my PR before doing anything here

rth:
Thanks for looking into it! The dependence on a non-maintained cache fork is not ideal: it's not strictly necessary, but it does make the build faster.

weiji14:
I'm getting the exact same `doctest.h:4189:47: error: size of array ‘altStackMem’` error on Ubuntu 22.04, is there a workaround for this at the moment?

Real-Gecko:
[...]

PR #3309, opened 2022-12-01
Unpin host Python patch versions in GHA

ryanking13:
Resolve: #2006

`ubuntu 21.10` or `ubuntu 22.04` which comes with `glibc >= 2.34` had an API change:

```
* Add _SC_MINSIGSTKSZ and _SC_SIGSTKSZ.  When _DYNAMIC_STACK_SIZE_SOURCE
  or _GNU_SOURCE are defined, MINSIGSTKSZ and SIGSTKSZ are no longer
  constant on Linux.  MINSIGSTKSZ is redefined to sysconf(_SC_MINSIGSTKSZ)
  and SIGSTKSZ is redefined to sysconf (_SC_SIGSTKSZ).  This supports
  dynamic sized register sets for modern architectural features like
  Arm SVE.
``` 

This causes trouble when installing ccache port (juj/ccache) for emscripten.
This issue is fixed in upstream ccache but it is not applied to the port.

@wlach made a PR (https://github.com/juj/ccache/pull/2), but it didn't get a response.

This PR overrides ccache URL to my fork (interestingly emsdk supports overriding urls). I applied @wlach's patch to it. We can remove this when upstream ccache port gets updated.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -17,7 +17,7 @@ jobs:
       - uses: actions/checkout@v2
       - uses: actions/setup-python@v3
         with:
-          python-version: 3.10
+          python-version: "3.10"
       - uses: pre-commit/action@v2.0.3
 
   test-docs:
@@ -26,7 +26,7 @@ jobs:
       - uses: actions/checkout@v2
       - uses: actions/setup-python@v3
         with:
-          python-version: 3.10
+          python-version: "3.10"
       - name: Install requirements
         shell: bash -l {0}
         run: |
@@ -49,7 +49,7 @@ jobs:
 
       - uses: actions/setup-python@v3
         with:
-          python-version: 3.10
+          python-version: "3.10"
 
       - name: Install requirements
         shell: bash -l {0}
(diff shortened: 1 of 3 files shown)


.github/workflows/main.yml
```diff
@@ -49,7 +49,7 @@ jobs:
 
       - uses: actions/setup-python@v3
         with:
-          python-version: 3.10.2
+          python-version: 3.10
```
ryanking13, line 52: `setup-python` doesn't have Python 3.10.2 for `ubuntu >= 22.04`. (See: https://github.com/pyodide/pytest-pyodide/pull/64). I think it is okay to use python 3.10.X for linting, testing docs, and running host tests of pyodide-build.

rth (approved):
Thanks @ryanking13 !

ryanking13:
Actually, the upstream PR is accepted. So we don't need to use a fork now. I'll remove that part, change the PR title, and merge.

Anyway, it was worth trying it, as I now know `--override-repository` flag which we can use if we need to patch tools anytime later :)

PR #3310, opened 2022-12-01
Get deps from pypi during build 

joemarshall:
### Description
Further extension to the work on pyodide build. This adds two new arguments:

`--build-dependencies`

Get dependency list from wheel and build (or download) wheels for all dependencies alongside the main wheel.

`--skip-dependency NAME` 
Skip a dependency - NAME can be either a package name, or the path to a json package list file in either repodata.json format (ie. Built in pyodide packages) or jupyterlite all.json format (i.e. jupyterlite wheel folder). 

It uses resolvelib to do dependency resolution because it can handle backtracking etc. If you call pyodide build to build something then chuck all the wheel files into jupyterlite pypi directory it should now be able to load without hitting pypi (as piplite uses the version provided locally by default). 

Needs tests but I thought it worth chucking up as soon as it was working for people to see

### Checklists

- [x] Add / update tests


pyodide-build/pyodide_build/out_of_tree/pypi.py
```diff
@@ -0,0 +1,377 @@
...
+from resolvelib.providers import AbstractProvider
+from unearth.evaluator import TargetPython
+from unearth.finder import PackageFinder
+
+from .. import common
+from . import build
+
+
+@contextmanager
+def stream_redirected(to=os.devnull, stream=None):
```
rth, line 32: Maybe you could use contextlib.redirect_stderr and redirect_stdout to avoid re-implementing it.

rth (commented):
Thanks for working on this @joemarshall !

I'm wondering if we actually want to manage the dependency resolution ourselves. The other alternative could be to use something like [pip-tools](https://github.com/jazzband/pip-tools) to generate a dependency tree and output the requirements.txt with pinned versions that can then be built.

I guess here we could do a more accurate resolution for the emscripten_wasm32 platform, but I'm a bit concerned about the feature creap. In particular,
a) it might be better to list the output of packages to build, so the user can double-check them before actually building them. So I wonder even if we implement this, if wouldn't be better to write a `requirements.txt` files with some new CLI command and then run `build` on that input.
b) there are a lot of ways in which dependency resolution could produce results different to what the user wanted. So we need to consider whether we want to deal with such issues in the future, or if we consider this out of scope, and rather rely on some existing tool to do some of this.

Overall I agree this is useful functionality, and I'm not opposed to this PR, but I just want to be sure we also consider existing alternatives (if any) that would allow solving part of this use case.

pyodide-build/pyodide_build/out_of_tree/pypi.py
joemarshall: contextlib doesn't redirect subprocess output - that's why this one is needed. Otherwise one would need to delve into anywhere subprocess is called (i.e. deep inside build) to redirect that.

joemarshall:
I was a bit torn about doing it like this with resolvelib (which is basically the resolution logic from pip made into a library) vs pip-tools (which uses more of pip internals, but is basically a similar thing). 

I think the pain with pip-tools is that there are a bunch of use cases that it doesn't easily support which are really super-useful for emscripten. In particular:

1) It doesn't support ignoring packages. Unless pyodide starts entirely using pypi packages and not using the pyodide package repository, this is a real pain; especially if the ignored packages have complex dependency trees, unwrapping them from the piptools output is a right pain. I also have the particular use case that there are various packages that use stuff to do cool multiprocessing stuff or processor specific stuff, which have dependencies on compiled packages that will just never compile on emscripten.

2) It doesn't really support cross-compiling - this isn't too bad right now as most packages don't deliberately support emscripten, but with things like jupyterlite, I think increasingly there will be packages that know about emscripten. I've got a couple of packages that drop deps on emscripten for example where they are used to run servers, or to do graphics.

3) It doesn't support building emscripten wheels for sdist packages. This is important because 
  a) a surprising number of packages are sdist only anyway, and many packages are also sdist only on emscripten
  b) you need to build the wheel before you can tell what the dependencies are, so you can't reasonably separate building wheels from resolving dependencies.

It's easy enough to make this logic output pinned requirements.txt files if required (and building from them is pretty trivial anyway given we support pypi builds already), but I do think having the core bit that calls resolvelib and makes wheels at the same time is important to support correctly resolving for emscripten. 

I have to moan about python packaging again btw, I get that in theory different builds could have different build-time generated dependencies, and that there are all sorts of build tools, but the fact that there is no standard metadata file included in an sdist which just specifies dependencies for each platform makes me sad. Having to compile wheels in order to find dependencies is a bit awful...

joemarshall:
> it might be better to list the output of packages to build, so the user can double-check them before actually building them. So I wonder even if we implement this, if wouldn't be better to write a `requirements.txt` files with some new CLI command and then run `build` on that input.

The problem I ran into is things that are sdist only, or don't have pyodide specific wheels. There's no reliable way to determine requirements for an sdist other than to compile it. So then my script to generate the requirements.txt had to build all the non-pure python packages anyway, before it could output a set of resolutions, and having it done in one go saves doubling time spent compiling.

Basically I think anything that makes the requirements.txt files will 
a) have to be integrated with and hence call into the pyodide build system to make wheels during the resolution process.
b) Need to use some kind of resolver library, either libresolve or pip-tools

It could in theory be a separate pyodide command that does this, although that would lose the benefit of being able to build everything for a project from scratch whilst only building wheels once. Dunno how important that is?

joemarshall:
@rth Any thoughts on the above:

I had a bit of a play with pip-tools again, but until there's a full cross-build environment for pyodide including support for using pyodide built-in packages in pip, it just isn't trivial to do this using pip-tools. I tried with pyodide pyenv but pip didn't really work for me there.

You end up hacking an isolated virtual environment and then hacking sys to change the values of sys_platform etc., then telling it a whole load of packages are unsafe, and the unsafe list has to include any dependencies of any packages that you already have, as it still traverses dependencies of unsafe packages. It freaks me out slightly as you end up relying on the internal behaviour of pip-tools, which in turn is tied to the internal behaviour of pip, both of which don't really support running as libraries as opposed to scripts. 

Personally I think that one script in pyodide build that does the downloading and building of dependencies (and of requirements.txt files) using libraries that are meant to be used for dependency resolution is probably less fragile and less work to maintain than a script to hack a virtual environment so that pip-tools works, which will end up needing maintenance as pip-tools upgrades things or changes behaviour. 

It would be possible with a load of work on cross-env and pyodide to get things to a state where one could just build entirely using pip, but I think that's some way off? I really want reliable package building for complex dependency trees to be there soon, as it just makes everything so much simpler when porting anything complicated that uses multiple libraries across.

hoodmane:
> It would be possible with a load of work on cross-env and pyodide to get things to a state where one could just build entirely using pip, but I think that's some way off? I really want reliable package building for complex dependency trees to be there soon, as it just makes everything so much simpler when porting anything complicated that uses multiple libraries across.

I looked into this a bit and I agree it looks quite complicated and fragile. If there is an alternative to writing some complicated hack to make pip's internals understand what we need then it is preferable. It would be cool to get editable installs working correctly though. Currently with my venvs we can do editable installs of pure Python packages, but not of binary packages.

pyodide-build/pyodide_build/out_of_tree/pypi.py
```diff
@@ -0,0 +1,386 @@
...
+# cache of built sdists by url source to wheel file path
+BUILD_CACHE: dict[str, dict[str, Any]] = {}
+
+
+def get_cached_built_wheel(url):
+    parsed_url = urlparse(url)
+    gz_name = Path(parsed_url.path).name
+    if url in BUILD_CACHE:
+        return BUILD_CACHE[url]["path"]
+    else:
```
hoodmane, line 60: Remove this `else` and deindent the rest of the function.

pyodide-build/pyodide_build/out_of_tree/pypi.py
```diff
@@ -0,0 +1,386 @@
...
+            stream.flush()
+            os.dup2(copied.fileno(), stream_fd)  # $ exec >&copied
+            to = None
+
+
+# cache of built sdists by url source to wheel file path
+BUILD_CACHE: dict[str, dict[str, Any]] = {}
+
+
+def get_cached_built_wheel(url):
```
hoodmane, line 55: maybe:
```suggestion
def build_wheel(url):
```
Also why not just use `functools.cache`?

pyodide-build/pyodide_build/out_of_tree/pypi.py
```diff
@@ -0,0 +1,386 @@
...
+            data = requests.get(url).content
+            f.write(data)
+            f.close()
+            shutil.unpack_archive(f.name, build_dir.name)
+            os.unlink(f.name)
+            files = list(os.listdir(build_dir.name))
+            if len(files) == 1:
+                os.chdir(os.path.join(build_dir.name, files[0]))
+            else:
+                os.chdir(build_dir.name)
```
hoodmane, line 75: Why not:
```suggestion        
        data = requests.get(url).content
        with tempfile.NamedTemporaryFile(suffix=".tar.gz") as f:
            f.write(data)
            f.close()
            shutil.unpack_archive(f.name, build_dir.name)
        files = list(os.listdir(build_dir.name))
        if len(files) == 1:
            os.chdir(os.path.join(build_dir.name, files[0]))
        else:
            os.chdir(build_dir.name)
```

pyodide-build/pyodide_build/out_of_tree/pypi.py
```diff
@@ -0,0 +1,386 @@
...
+        os.chdir(build_dir.name)
+        with tempfile.NamedTemporaryFile(suffix=".tar.gz", delete=False) as f:
+            data = requests.get(url).content
+            f.write(data)
+            f.close()
+            shutil.unpack_archive(f.name, build_dir.name)
+            os.unlink(f.name)
+            files = list(os.listdir(build_dir.name))
+            if len(files) == 1:
+                os.chdir(os.path.join(build_dir.name, files[0]))
```
hoodmane, line 73: Is it necessarily the case that if there is only one file, it is a directory?

pyodide-build/pyodide_build/out_of_tree/pypi.py
```diff
@@ -0,0 +1,386 @@
...
+                with stream_redirected(to=f, stream=sys.stdout):
+                    with stream_redirected(to=f, stream=sys.stderr):
+                        wheel_path = build.run(
+                            PyPIProvider.BUILD_EXPORTS, PyPIProvider.BUILD_FLAGS
+                        )
+            except BaseException as e:
+                print(" Failed\n Error is:")
+                f.seek(0)
+                sys.stdout.write(f.read())
+                raise e
```
hoodmane, line 88: How about combining the with blocks?
```suggestion
        with (
            tempfile.TemporaryFile(mode="w+") as f,
            stream_redirected(to=f, stream=sys.stdout),
            stream_redirected(to=f, stream=sys.stderr)
        ):
            try:
                wheel_path = build.run(
                    PyPIProvider.BUILD_EXPORTS, PyPIProvider.BUILD_FLAGS
                )
            except BaseException as e:
                print(" Failed\n Error is:")
                f.seek(0)
                sys.stdout.write(f.read())
                raise e
```

pyodide-build/pyodide_build/out_of_tree/pypi.py
```diff
@@ -0,0 +1,386 @@
...
+
+    @property
+    def dependencies(self):
+        if self._dependencies is None:
+            self._dependencies = list(self._get_dependencies())
+        return self._dependencies
+
+
+if TYPE_CHECKING:
+    APBase = AbstractProvider[Requirement, Candidate, str]
```
hoodmane, line 145: What goes wrong if you write `AbstractProvider[Requirement, Candidate, str]` when `not TYPE_CHECKING`?

pyodide-build/pyodide_build/out_of_tree/pypi.py
```diff
@@ -0,0 +1,386 @@
...
+        extras = self.extras if self.extras else [""]
+
+        for d in deps:
+            r = Requirement(d)
+            if r.marker is None:
+                yield r
+            else:
+                for e in extras:
+                    if r.marker.evaluate({"extra": e}):
+                        yield r
```
hoodmane, line 135: Probably this should be:
```suggestion
                        yield r
                        break
```
so that the requirement isn't yielded multiple times?

pyodide-build/pyodide_build/out_of_tree/pypi.py
```diff
@@ -0,0 +1,386 @@
...
+        project/version.
+        """
+        raise NotImplementedError
+
+    def identify(self, requirement_or_candidate):
+        base = super().identify(requirement_or_candidate)
+        return base
+
+    def get_dependencies(self, candidate):
+        deps = list(super().get_dependencies(candidate))
```
hoodmane, line 170: Since `AbstractProvider.get_dependencies` raises `NotImplementedError`
https://github.com/sarugaku/resolvelib/blob/master/src/resolvelib/providers.py#L109
doesn't this just unconditionally raise `NotImplementedError`? I think you should get rid of `ExtrasProvider` and combine the logic here with the logic in `PyPIProvider.get_dependencies`.

pyodide-build/pyodide_build/cli/build.py
```diff
@@ -121,27 +93,32 @@ def url(
...
     initialize_pyodide_root()
+    orig_dir = Path.cwd()
+    if source_location != ".":
+        # build in this folder
+        os.chdir(source_location)
     common.check_emscripten_version()
-    backend_flags = [source_location] + ctx.args
-    build.run(exports, backend_flags)
+    backend_flags = ctx.args
+    built_wheel = Path(build.run(exports, backend_flags, outdir=orig_dir / "dist"))
```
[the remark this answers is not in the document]
joemarshall: good spot - all do paths now

pyodide-build/pyodide_build/out_of_tree/pypi.py
joemarshall: because NamedTemporaryFile deletes on first close, not when the context manager exits. Stupid, but it is how it works annoyingly.

joemarshall:
@hoodmane I got round to doing the review changes for this. I also added thorough tests which serve a very basic fake pypi to test things like unresolvable requirements and complicated extra handling.

I think at the moment it feels like it is reliable enough to merge. I have a couple of todos that would be worth putting in which is to read and resolve requirements.txt files, and to write out a flat requirements.txt but ideally I'd like to put those in a separate pr?

hoodmane:
Sounds good to me. This pr is already quite large any additional features would be best left to follow ups. I'll try to look this over one more time and merge it in a bit.

hoodmane:
Or you could merge it if you like.

pyodide-build/pyodide_build/out_of_tree/pypi.py
ryanking13: Could you please write a comment about it for the people who read this code in the future?

pyodide-build/pyodide_build/tests/test_pypi.py
```diff
@@ -0,0 +1,321 @@
...
+    built_wheels = set(output_dir.glob("*.whl"))
+    assert len(built_wheels) == 2
+
+
+@pytest.mark.parametrize(
+    "fn", [test_fetch_or_build_pypi, test_fetch_or_build_pypi_with_deps_and_extras]
+)
+def test_host_in_node_test(runtime, fn, tmp_path):
+    if runtime.startswith("node"):
+        fn(tmp_path)
```
ryanking13, line 307: These tests are not actually running. I am trying to fix broken CLI in #3390. Probably you will need to fix tests in a same way in #3390.

ryanking13 (commented):
Thanks, @joemarshall. I have minor comments about some details. I wasn't following the full conversation in this PR, so I'll let others approve.

joemarshall:
> 

Hi @hoodmane , I've done the review changes above now, if you could merge that would be great, I don't have merge access to this repo.

ryanking13 (changes requested):
Thanks again for your work @joemarshall. Please update the changelog, and some remaining test issues need to be fixed.

joemarshall:
@ryanking13 Thanks, done that now.

joemarshall:
Damn, something broken in node tests, I'll sort that tomorrow

joemarshall:
@ryanking13 tests run in ci correctly now

ryanking13 (approved):
Thanks!

PR #3313, opened 2022-12-02
Make fs timestamps have millisecond resolution rather than second resolution

### LINKED ISSUE #3311: ModuleNotFoundError after writing multiple files with FS.writeFile
## 🐛 Bug

This might be a duplicate of #737, but I'm opening a new issue because I'm seeing something very weird.
I'm using Pyodide  0.21.3.

### To Reproduce

Consider the following snippet (I run it in node, but I can reproduce the same behavior also in the browser):
```js

function busyWait(ms) {
    let start = new Date().valueOf();
    while( (new Date()).valueOf() - start < ms)
        ;
}

async function main() {
        indexURL: "./node_modules/pyodide",
    });

    console.log("Pyodide version: ", pyodide.version);

    let success = 0;
    let errors = 0;
    for(let i=0; i<100; i++) {
        // uncommenting EITHER of the following lines, it works reliably
        //pyodide.runPython('import importlib; importlib.invalidate_caches()');
        //busyWait(1000);

        pyodide.FS.writeFile(`bar${i}.py`, `x = ${i}`);
        try {
            pyodide.runPython(`import bar${i} as bar; print('OK ', bar.x)`);
            success++;
        }
        catch(e) {
            //console.log(e);
            console.log('ERR', i);
            errors++;
        }
    }
    console.log("success:", success);
    console.log("errors :", errors);
    console.log("error rate: ", (errors/(success+errors))*100, "%");
}

main();
```

The expected result is obviously to see all `OK` and 0% error rate, but what happens is that very often I get `ERR`.
I observed the following:
- if I uncomment the call to `invalidate_caches()`, it passes reliably
- if I uncomment `busyWait(1000)`, (i.e., 1 second) it passes reliably

The interesting part is what happens with lower values of `busyWait`:
- with `busyWait(500)`, it fails ~50% of the times
- with `busyWait(100)` it fails ~90% of the times
- with `busyWait(10)` it fails ~99% of the times

The following is just a guess, but I think that considering the above it's a reasonable explanation:
- somewhere there is a timestamp which is stored with a resolution of 1s
- if the "new" files are written "too early", importlib think that it is of the same epoch as the "old files/directories" and thus it thinks the cache is still valid
- if you wait long enough, the 1s timestamp gets a new value and importlib correctly notices that there is a "new" file

I am not sure whether this is a bug/"feature" of emscripten. I thought it's a good idea to open it here so that we can maybe discuss possible pyodide-specific mitigations and/or workarounds

hoodmane:
Resolves #3311. We can probably remove all the calls to `invalidate_caches` too...

Upstream PR: https://github.com/emscripten-core/emscripten/pull/18301

### Checklists

- [ ] Add / update tests

### DIFF
--- a/emsdk/patches/0001-Actually-fill-in-tv_nsec-fields-of-stat-struct.patch
+++ b/emsdk/patches/0001-Actually-fill-in-tv_nsec-fields-of-stat-struct.patch
@@ -1,0 +1,31 @@
+From 50aeeda991f464e22ede556eb1647efea258adc4 Mon Sep 17 00:00:00 2001
+From: Hood Chatham <roberthoodchatham@gmail.com>
+Date: Thu, 1 Dec 2022 17:30:05 -0800
+Subject: [PATCH] Actually fill in tv_nsec fields of stat struct
+
+---
+ src/library_syscall.js | 6 +++---
+ 1 file changed, 3 insertions(+), 3 deletions(-)
+
+diff --git a/src/library_syscall.js b/src/library_syscall.js
+index 8f261ec92..400985062 100644
+--- a/src/library_syscall.js
++++ b/src/library_syscall.js
+@@ -62,11 +62,11 @@ var SyscallsLibrary = {
+       {{{ makeSetValue('buf', C_STRUCTS.stat.st_blksize, '4096', 'i32') }}};
+       {{{ makeSetValue('buf', C_STRUCTS.stat.st_blocks, 'stat.blocks', 'i32') }}};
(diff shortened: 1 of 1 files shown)


ryanking13:
Thanks, @hoodmane!

> We can probably remove all the calls to `invalidate_caches` too...

IMO, we can still keep `invalidate_caches` calls if their overhead is negligible (probably we can measure it), as it ensures that the cache is invalidated instead of relying on modification times.

ryanking13 (approved):
We would need some tests, otherwise LGTM. Thanks!

rth:
For tests, since it was accepted upstream, I think we can probably merge as is. Just needs a changelog entry.

PR #3317, opened 2022-12-03
Add generator argument lifetime handling

hoodmane:
This adds lifetime handling to the arguments of JavaScript sync and async generator functions and to the arguments of `send` and `asend`. The arguments and the `send` values are kept alive until the generator finishes. This is accomplished through a wrapper around the generator. The return value of the generator is not destroyed by the wrapper, but it will get destroyed in the C code for `asend`/`athrow`/`aclose` (unless the generator wrapper is sent back to JavaScript and called directly).

As a followup we could offer the option to wrap generators that were created in JavaScript. Currently if you import a generator from JavaScript then the lifetimes of the `send` arguments are not extended so you have to wrap them with `create_proxy`.

### Checklists

- [x] Add / update tests


src/tests/test_jsproxy.py
```diff
@@ -2100,3 +2100,59 @@ async def test_agen_aclose(selenium):
...
+    from pyodide.ffi import JsGenerator
+
+    f = run_js(
+        """
+        (function *(x) {
+            let l = [x];
+            l.push(yield);
+            l.push(yield);
+            l.push(yield);
+            return pyodide.toPy(l.map((x) => x.toString()));
```
ryanking13, line 2119: So, before this PR, proxies inside `l` is already destroyed so it will fail here I guess? Would you mind adding some comments about what these tests actually want to test? It was a bit hard for me to understand at first glance.

ryanking13 (approved):
LGTM, Thanks a lot @hoodmane!

src/tests/test_jsproxy.py
hoodmane: Yeah. I'll add a comment.

hoodmane: One funny thing here is that it creates a difference between generators that were created in JavaScript and then brought over to Python and generators that were accessed as an attribute from a `JsProxy`. There's a similar problem with promises. It creates an annoying inconsistency, not sure how best to deal with it.

I think also it would be good not to destroy proxies with `roundtrip=true` in a lot of cases...

hoodmane:
@ryanking13 Odd that the playwright tests fail but the Selenium tests don't... It's a real failure too not a timeout.

Issue #2545, opened 2022-05-11
Using the gensim module (NLP)


Amodio:
## 🐍 Package Request

- gensim / latest
- https://files.pythonhosted.org/packages/aa/1a/e3a87ee67ffa989438dcf4f113a710335aa95ce96c586c68cac0fa6af0ee/gensim-4.2.0-cp310-cp310-win_amd64.whl

Hello, I am unable to load the gensim module (for doing NLP in the browser) with micropip: `ValueError: 'gensim-4.2.0-cp310-cp310-win_amd64.whl' is not a pure Python 3 wheel`
Is there any solution to fix this or is someone working on including it to pyodide?

Loading another module by URL works on the same host:
```
      async function main() {
        let pyodide = await loadPyodide()
        await pyodide.loadPackage("micropip")
        await pyodide.runPythonAsync(`
        import micropip
        await micropip.install('https://files.pythonhosted.org/packages/ed/dc/c02e01294f7265e63a7315fe086dd1df7dacb9f840a804da846b96d01b96/snowballstemmer-2.2.0-py2.py3-none-any.whl')
        print(micropip.list()) 
      `)
```

Thanks

dopplershift:
The `gensim` wheel you mention is pre-compiled for x86_64 for Windows, while the other wheel (for `snowballstemmer`) only contains Python code. Using `gensim` in pyodide will require [adding it as a package to pyodide](https://pyodide.org/en/stable/development/new-packages.html) so that the C/C++ code in gensim is compiled to Wasm.

PR #3326, opened 2022-12-06
Add gensim package #2545

Amodio:
### Description

This PR adds the gensim (and smart_open) package(s).

### DIFF
--- a/packages/gensim/meta.yaml
+++ b/packages/gensim/meta.yaml
@@ -1,0 +1,31 @@
+package:
+  name: gensim
+  version: 4.2.0
+  top-level:
+    - gensim
+source:
+  url: https://files.pythonhosted.org/packages/24/97/2197f018ee9f8ce2f071b2d9c6711c76159aead710f8d24a2bf006082a28/gensim-4.2.0.tar.gz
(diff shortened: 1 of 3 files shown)


packages/gensim/meta.yaml
```diff
@@ -0,0 +1,31 @@
...
+    - smart_open
+test:
+  imports:
+    - gensim
+    - gensim.corpora
+    - gensim.models
+    - gensim.parsing
+    - gensim.scripts
+    - gensim.similarities
+    - gensim.test
```
rth, line 25: tests are unvendored by default, so should remove this line, as long as
[`build/unvendor-tests`](https://pyodide.org/en/stable/development/meta-yaml.html#build-unvendor-tests) option is enabled (it is by default).

This is also why CI is failing I think.

rth (commented):
Thanks! One comment below, also please add it to the list in https://github.com/pyodide/pyodide/blame/main/docs/project/changelog.md#L263

packages/gensim/meta.yaml
Amodio: Thank you, should be better :)

rth:
Thanks @Amodio

PR #3331, opened 2022-12-09
Package OpenBLAS and use OpenBLAS in scipy

lesteve:
### Description

Opening a draft PR on this.

This could potentially get rid of some of the low-level issues like https://github.com/pyodide/pyodide/issues/3203.

Also OpenBLAS is more maintained than CLAPACK.

Related to #227.

### Checklists


lesteve:
So I have the same error locally as in https://circleci.com/gh/pyodide/pyodide/62097
```
make: *** No rule to make target 'Makefile.'
```

Note the `.` at the end. For now this is what I have discovered:
In OpenBLAS, `Makefile.conf` is auto-generated in `Makefile.system` by this line:
```
# Generating Makefile.conf and config.h
DUMMY := $(shell $(MAKE) -C $(TOPDIR) -f Makefile.prebuild CC="$(CC)" FC="$(FC)" HOSTCC="$(HOSTCC)" HOST_CFLAGS="$(GETARCH_FLAGS)" CFLAGS="$(CFLAGS)" BINARY=$(BINARY) USE_OPENMP=$(USE_OPENMP) DYNAMIC_ARCH=$(DYNAMIC_ARCH) TARGET_CORE=$(TARGET_CORE) ONLY_CBLAS=$(ONLY_CBLAS) TARGET=$(TARGET) all)
```

And somehow inside the Pyodide docker, the generated `Makefile.conf` has `ARCH=` and `Makefile.system` includes `Makefile.$(ARCH)` so `Makefile.` (with the `.` at the end) which does not exist.

I managed to build OpenBLAS on my machine not sure what is different compared to the Pyodide docker image.

A mystery that needs to be investigated a bit further ...

lesteve:
For now I will use a recent commit on the OpenBLAS `develop` branch where the previous error does not show up.

packages/openblas/meta.yaml
```diff
@@ -0,0 +1,24 @@
...
+  script: |
+    # seems like .zip does not maintain executable flags, need to reset these
+    chmod u+x c_check
+    chmod u+x f_check
+    chmod u+x exports/gensymbol
+    make CC=emcc HOSTCC=gcc TARGET=RISCV64_GENERIC NOFORTRAN=1 USE_THREAD=0
+    mkdir -p dist
+    cp libopenblas.so dist/
+    mkdir -p ${WASM_LIBRARY_DIR}/{lib,include}
+    # TODO do I need include as in CLAPACK?
```
rth, line 22: I would have guessed yes, but you can also probably not include it and see if there are any related include errors as done now.

rth (commented):
Thanks for working on this! So it looks like OpenBLAS is building OK. Next step could be to remove CLAPACK (also to make sure it's not used) and link to OpenBLAS instead in scipy.

Maybe we should build it with `-Os` instead of `-O2` as currently, it produces a zip of 2.8MB while CLAPACK was 1.24MB

lesteve:
There are some errors when linking scipy with OpenBLAS:

```
wasm-ld: error: function signature mismatch: daxpy_
>>> defined as (i32, i32, i32, i32, i32, i32) -> i32 in build/temp.emscripten_3_1_27_wasm32-3.10/libodrpack.a(d_odr.o)
>>> defined as (i32, i32, i32, i32, i32, i32) -> void in /src/packages/.libs/lib/libopenblas.so

wasm-ld: error: function signature mismatch: dswap_
>>> defined as (i32, i32, i32, i32, i32) -> i32 in build/temp.emscripten_3_1_27_wasm32-3.10/libodrpack.a(d_lpk.o)
>>> defined as (i32, i32, i32, i32, i32) -> void in /src/packages/.libs/lib/libopenblas.so

wasm-ld: error: function signature mismatch: drot_
>>> defined as (i32, i32, i32, i32, i32, i32, i32) -> i32 in build/temp.emscripten_3_1_27_wasm32-3.10/libodrpack.a(d_odr.o)
>>> defined as (i32, i32, i32, i32, i32, i32, i32) -> void in /src/packages/.libs/lib/libopenblas.so

wasm-ld: error: function signature mismatch: dcopy_
>>> defined as (i32, i32, i32, i32, i32) -> i32 in build/temp.emscripten_3_1_27_wasm32-3.10/libodrpack.a(d_odr.o)
>>> defined as (i32, i32, i32, i32, i32) -> void in /src/packages/.libs/lib/libopenblas.so

wasm-ld: error: function signature mismatch: dscal_
>>> defined as (i32, i32, i32, i32) -> i32 in build/temp.emscripten_3_1_27_wasm32-3.10/libodrpack.a(d_lpk.o)
>>> defined as (i32, i32, i32, i32) -> void in /src/packages/.libs/lib/libopenblas.so

wasm-ld: error: function signature mismatch: drotg_
>>> defined as (i32, i32, i32, i32) -> i32 in build/temp.emscripten_3_1_27_wasm32-3.10/libodrpack.a(d_odr.o)
>>> defined as (i32, i32, i32, i32) -> void in /src/packages/.libs/lib/libopenblas.so
emcc: error: '/src/emsdk/emsdk/upstream/bin/wasm-ld -o build/lib.emscripten_3_1_27_wasm32-3.10/scipy/odr/__odrpack.cpython-310-wasm32-emscripten.so --fatal-warnings -L/src/packages/.artifacts/lib/python3.10/site-packages/numpy//core/lib/ -L/src/packages/.artifacts/lib/python3.10/site-packages/numpy//random/lib/ -L/src/cpython/installs/python-3.10.2/lib/ build/temp.emscripten_3_1_27_wasm32-3.10/scipy/odr/__odrpack.o -Lbuild/temp.emscripten_3_1_27_wasm32-3.10 build/temp.emscripten_3_1_27_wasm32-3.10/libodrpack.a /src/packages/.libs/lib/libopenblas.so -L/src/emsdk/emsdk/upstream/emscripten/cache/sysroot/lib/wasm32-emscripten/pic -mllvm -combiner-global-alias-analysis=false -mllvm -enable-emscripten-sjlj -mllvm -disable-lsr --import-undefined --import-memory --strip-debug --export-if-defined=PyInit___odrpack --export-if-defined=__start_em_asm --export-if-defined=__stop_em_asm --export-if-defined=__start_em_lib_deps --export-if-defined=__stop_em_lib_deps --export-if-defined=__start_em_js --export-if-defined=__stop_em_js --export-if-defined=__wasm_apply_data_relocs --export=__wasm_call_ctors --experimental-pic -shared --no-export-dynamic' failed (returned 1)
error: Command "/tmp/tmp_tdnjvie/gfortran -Wall -g -Wall -g -shared build/temp.emscripten_3_1_27_wasm32-3.10/scipy/odr/__odrpack.o -Lbuild/temp.emscripten_3_1_27_wasm32-3.10 -lodrpack -lgfortran -o build/lib.emscripten_3_1_27_wasm32-3.10/scipy/odr/__odrpack.cpython-310-wasm32-emscripten.so -I/src/packages/.libs/include /src/packages/.libs/lib/libopenblas.so" failed with exit status 1
```

There are already some warning when building OpenBLAS, so maybe this can be fixed in OpenBLAS:

hoodmane:
We've been patching all void return value functions to return i32. This has something to do with alternative returns.

lesteve:
OK I'll try something like this, thanks!

Out of curiosity, the convention to always return int rather than void is a Pyodide convention (always returning `void` could have worked too) or there is some reason behind it?

hoodmane:
It's because of a fortran feature called alternative returns.

hoodmane:
https://docs.oracle.com/cd/E19957-01/805-4940/6j4m1u7qv/index.html

lesteve:
So scipy builds now but I get errors like this locally for example when doing `import scipy.linalg` in the Pyodide console:
```
pytest_pyodide.runner.JavascriptException: Error: bad export type for `ztrsm_`: undefined
```

They happens in CircleCI as well e.g. this build https://app.circleci.com/pipelines/github/pyodide/pyodide/5095/workflows/d2b521c2-9683-4bc0-a994-ffd7520610e2/jobs/62303:
```
E   pytest_pyodide.runner.JavascriptException: Error: bad export type for `ztrsm_`: undefined

E       at Object.reportUndefinedSymbols (/root/repo/dist/pyodide.asm.js:10:346898)

E       at loadPackage (/root/repo/dist/pyodide.asm.js:10:142844)

E       at async Object.loadPackagesFromImports (/root/repo/dist/pyodide.asm.js:10:156427)

E       at async evalmachine.<anonymous>:5:1

E       at async evalmachine.<anonymous>:4:26

E       at async evalCode (/usr/local/lib/python3.10/site-packages/pytest_pyodide/node_test_driver.js:81:33)
```

ryanking13:
You need to add `${SIDE_MODULE_LDFLAGS}` (which adds `-s SIDE_MODULE` flag) to LDFLAGS when calling make. Without it, the built .so file actually becomes a static_library. You can find the warning message in the build log of openblas:

```
```

lesteve:
Thanks for the input, I tried to do follow your suggestion in my last commit, let's see what happens!

lesteve:
OK I get a build error now in openblas somehow the LDFLAGS has ends up being in a clang command to compile a `.c` file:


I am going to propose

hoodmane:
@lesteve Maybe add a second environment variable to Makefile.envs which is LDFLAGS minus `std=c++`?

ryanking13:
> @lesteve Maybe add a second environment variable to Makefile.envs which is LDFLAGS minus `std=c++`?

Or maybe we can just remove it from LDFLAGS? Do we actually need to be specified in LDFLAGS?

hoodmane:
Might as well try.

rth:
About the latest error at runtime,
```
pytest_pyodide.runner.JavascriptException: Error: bad export type for `f2pywrapzlange_`: undefined
    at Object.reportUndefinedSymbols (http://127.0.0.1:60433/pyodide.asm.js:10:346898)
    at loadPackage (http://127.0.0.1:60433/pyodide.asm.js:10:142844)
    at async Object.loadPackagesFromImports (http://127.0.0.1:60433/pyodide.asm.js:10:156427)
    at async run (eval at executeAsyncScript (:552:26), <anonymous>:6:1)
    at async eval (eval at executeAsyncScript (:552:26), <anonymous>:20:34)
```
so the `zlange_` symbols are correctly built,
```
$  pyodide auditwheel exports  dist/libopenblas.so  | grep zlange
│   │   Export(name='zlange_', kind=<ExternType.FUNC: 0>, index=3125),
│   │   Export(name='LAPACKE_zlange', kind=<ExternType.FUNC: 0>, index=5767),
│   │   Export(name='LAPACKE_zlange_work', kind=<ExternType.FUNC: 0>, index=5768),
```
but it looks like there is an issue with f2py wrappers

lesteve:
In my last commit, I think I managed to get back to get back to the previous state where scipy was building but there was an error at import time. Locally I do get `pyodide.JsException: Error: bad export type for `f2pywrapzlange_`: undefined` as https://github.com/pyodide/pyodide/pull/3331#issuecomment-1400193205.

lesteve:
So by trying to debug a bit more with Chromium I realised actually that there were errors like this:

```
LinkError: WebAssembly.instantiate(): Import #29 module="env" function="dcopy_" error: imported function does not match the expected type
```

Any insights into this more than welcome!

This kind of error is caught in `pyodide.asm.js` with this kind of code:
```js
try {
  const buffer = await (async function (
      name,
      channel,
      checkIntegrity = !0
    ) {
      let file_name, uri, file_sub_resource_hash;
      if ("default channel" === channel) {
        if (!(name in API.repodata_packages))
          throw new Error(
            `Internal error: no entry for package named ${name}`
          );
        (file_name = API.repodata_packages[name].file_name),
          (uri = resolvePath(file_name, API.config.indexURL)),
          (file_sub_resource_hash =
            API.package_loader.sub_resource_hash(
              API.repodata_packages[name].sha256
            ));
      } else (uri = channel), (file_sub_resource_hash = void 0);
      checkIntegrity || (file_sub_resource_hash = void 0);
      try {
        return await loadBinaryFile(uri, file_sub_resource_hash);
      } catch (e) {
        if (!IN_NODE || "default channel" !== channel) throw e;
      }
      console.log(
        `Didn't find package ${file_name} locally, attempting to load from ${cdnURL}`
      );
      let binary = await loadBinaryFile(cdnURL + file_name);
      return (
        console.log(
          `Package ${file_name} loaded from ${cdnURL}, caching the wheel in node_modules for future use.`
        ),
        await nodeFsPromisesMod.writeFile(uri, binary),
        binary
      );
    })(pkg.name, pkg.channel, checkIntegrity),
    installPromisDependencies = pkg.depends.map((dependency) =>
      toLoad.has(dependency)
        ? toLoad.get(dependency).done
        : Promise.resolve()
    );
  await Promise.all(installPromisDependencies),
    await (async function (name, buffer, channel) {
      let pkg = API.repodata_packages[name];
      pkg ||
        (pkg = {
          file_name: ".whl",
          shared_library: !1,
          depends: [],
          imports: [],
          install_dir: "site",
        });
      const filename = pkg.file_name,
        dynlibs = API.package_loader.unpack_buffer.callKwargs({
          buffer: buffer,
          filename: filename,
          target: pkg.install_dir,
          calculate_dynlibs: !0,
          installer: "pyodide.loadPackage",
          source: "default channel" === channel ? "pyodide" : channel,
        });
      await loadDynlibsFromPackage(pkg, dynlibs);
    })(pkg.name, buffer, pkg.channel),
    loaded.add(pkg.name),
    (loadedPackages[pkg.name] = pkg.channel);
} catch (err) {
  failed.set(name, err);
  console.log(name);
  console.log(err);
} finally {
  pkg.done.resolve();
}
```

ryanking13:
Could you try patching the return type of that function either in openblas or scipy?

I suspect this is another void / int type mismatch issue https://github.com/pyodide/pyodide/pull/3331#issuecomment-1345940239. in scipy, the [return type of dcopy_ is void](https://github.com/scipy/scipy/blob/8a64c938ddf1ae4c02a08d2c5e38daeb8d061d38/scipy/sparse/linalg/_dsolve/SuperLU/SRC/ilu_dcopy_to_ucol.c#L29) while openblas, it seems like the [return type  int](https://github.com/xianyi/OpenBLAS/blob/974acb39ff86121a5a94be4853f58bd728b56b81/relapack/src/dsytrf_rook_rec2.c#L54)

lesteve:
Good point, I actually tried many things and uncommented the `sed` to change the return type from void to int in `openblas/meta.yaml`. Now I get this with my added logging:

```
LinkError: WebAssembly.instantiate(): Import #77 module="env" function="scasum_" error: imported function does not match the expected type
```

Is there a way with `pyodide auditwheel` (and/or `wasm-objsump`?) to get the type of the function both in `libopenblas.so` and the scipy wheel to understand a bit better what is going on, rather than guessing?

ryanking13:
> Is there a way with pyodide auditwheel (and/or wasm-objsump?) to get the type of the function both in libopenblas.so and the scipy wheel to understand a bit better what is going on, rather than guessing?

I added [that feature](https://github.com/ryanking13/auditwheel-emscripten/pull/9) for you to auditwheel-emscripten :)

```sh
pip install auditwheel-emscripten>=0.0.11
```

Adding `--show-type` flag when calling `pyodide auditwheel exports` and `pyodide auditwheel imports` will show function types.
I didn't test much, so please let me know if something does not work well.

For `sasum`, the return type mismatches: f32 vs f64.

```
$ pyodide auditwheel exports libopenblas.so --show-type | grep sasum
FUNC      sasum_  (i32, i32, i32) -> (f32)
```

```
$ pyodide auditwheel imports scipy-1.9.3-cp311-cp311-emscripten_3_1_31_wasm32.whl --show-type | grep sasum
env      FUNC    sasum_  (i32, i32, i32) -> (f64)
```

lesteve:
> I added https://github.com/ryanking13/auditwheel-emscripten/pull/9 for you to auditwheel-emscripten :)

Wow very nice, thanks a lot, this looks super useful! I'll give it a go and let you know if I spot issues.

hoodmane:
[pyodide.asm.js.txt](https://github.com/pyodide/pyodide/files/10944611/pyodide.asm.js.txt)

lesteve:
So there is some progress on the OpenBLAS front! I have some hope that the CI will be green on my last commit :crossed_fingers:. 

I took plenty of shortcuts to get there so this will need some clean-up. 

Running the full scipy test suite locally on this branch it looks like there may be some regression introduced by these changes, so I'll try to take a closer look.

rth:
Wow, great job @lesteve !

lesteve:
CI is green as expected (OK codecov ruins the greenness but who cares :wink:). Now comes the remaining 80% of the work:
- run the full scipy tests to test this more thoroughly and make sure there are not too many regressions
- cleaning this up
- remember why I changed some of the things and whether I really needed to change it

rth:
> Running the full scipy test suite locally on this branch it looks like there may a few regressions introduced by these changes

Well at least now we would have people upstream in openblas to talk about it :)

lesteve:
> Well at least now we would have people upstream in openblas to talk about it :)

The thing is the regressions may have been introduced by this PR changes somehow ... for example right now it seems like the determinant of a complex matrix is always zero in this branch ...

```py
from scipy.linalg._flinalg_py import get_flinalg_funcs

a = np.array([[1, 2], [3, 4j]])

fdet, = get_flinalg_funcs(('det',), (a,), debug=1)
print(fdet.__doc__)
a_det = fdet(a)
print('scipy', a_det)
```

prints `(0j, 1)` (the determinant is the first element of the tuple). Trying to debug this right now ...

lesteve:
I have been one step further and managed to reproduce the issue with `zgetrf` with cython lapack scipy bindings


```cython
def test():
    cdef int n = 2
    cdef int ipv
    cdef int info
    arr = np.array([[1, 2], [3, 4j]])
    cdef cnp.complex128_t [:] arr_view = arr.ravel()
    print('input', arr)
    zgetrf(&n, &n, &arr_view[0], &n, &ipv, &info)
    result = np.asarray(<cnp.complex128_t[:n, :n]> &arr_view[0])
    print('result', result)
    print('ipv', ipv)
    print('info', info)
```

(if you really want to know I patch scipy/linalg/_generate_pyx.py to add this and then the cython lapack test to be able to run my code as a test ...)

which prints 
```
input [[1.+0.j 2.+0.j]
[3.+0.j 0.+4.j]]
result [[1.+0.j 2.+0.j]
[0.+4.j 3.-8.j]]
ipv 2
info 1
```

info = 1 means the matrix is singular see [this](https://netlib.org/lapack/explore-html/d3/d01/group__complex16_g_ecomputational_ga5b625680e6251feb29e386193914981c.html) for more details.

Not sure what the next step is, maybe trying to write directly some C code using OpenBLAS :scream:.

I tried to compile with source map following https://pyodide.org/en/stable/development/debugging.html#using-c-source-maps hoping I may be able to debug this with a browser debugger but so far I haven't managed to see the C code with the emscripten.

lesteve:
Turns out it was an OpenBLAS bug in 0.3.22 that has been fixed in 0.3.23 :face_with_head_bandage:, see https://github.com/xianyi/OpenBLAS/issues/3976 or https://github.com/scipy/scipy/pull/18225.

Looking at the bright side, I learned how to write C code using OpenBLAS, compile it with emscripten, and run it with node :wink:

rth:
From my point of view, there are only bright sides:
 a) some of the BLAS related issues would now be getting fixed upstream
 b) if we have any questions about hacking OpenBLAS for emscripten we could ask you questions :)

Seriously though, thanks so much for working on this @lesteve ! It's a difficult topic, but I guess there is no other way if we want to do more serious scientific computing in the browser.

hoodmane:
I had a draft of essentially the same comment as @rth =)

pyodide-build/pyodide_build/_f2c_fixes.py
```diff
@@ -229,6 +231,19 @@ def fix_f2c_output(f2c_output_path: str) -> str | None:
...
+    # return a real... but maybe not all?? Or maybe the others are affected also
+    # by this bug.
+    # pattern = "|".join(["sasum_", "scasum_", "scnrm2_", "sdot_", "sdsdot_", "snrm2_",
+    #                     "clanhs_", "slamch_", "slapy2_", "psnrm2_"])
+    # if "linalg" in f2c_output_path:
+    #     lines = [
+    #         re.sub(f"extern doublereal ({pattern})", r"extern float \1", line)
+    #         for line in lines
+    #     ]
+
```
hoodmane, line 246:
```suggestion
```

pyodide-build/pyodide_build/_f2c_fixes.py
```diff
@@ -478,8 +493,14 @@ def scipy_fix_cfile(path: str) -> None:
...
-    if path.endswith("_fblasmodule.c"):
-        text = text.replace(" float (*f2py_func)", " double (*f2py_func)")
+    if path.endswith("stats/statlib/spearman.c"):
+        # in scipy/stats/statlib/swilk.f ALNORM is called with a double, and in
+        # scipy/stats/statlib/spearman.f with a real this generates
+        # inconsistent signature. Let's use double in both, I don't think this
+        # code path will work (but at least it will compile) since it needs
+        # "ALNORM = algorithm AS66", which I don't think we have with the f2c
+        # route
+        text = text.replace("extern real alnorm_", "extern doublereal alnorm_")
```
hoodmane, line 503: Is there an upstream bug here?

packages/gensim/meta.yaml
```diff
@@ -12,6 +12,14 @@ requirements:
     - scipy
     - six
     - smart_open
+build:
+  script: |
+    sed -ri 's/(cdef dsdot_ptr.+)fblas\.sdot/\1fblas.ddot/' gensim/models/word2vec_inner.pyx
```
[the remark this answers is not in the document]
lesteve: I'll do a proper patch when I clean up. Essentially gensim is trying to decide whether `sdot` returns a double or a float, with some clever C casts of function but in emscripten that yields to a signature mismatch. This is the line that creates the signature mismatch (at the call site expect a f64 but the function called returns a f32):
https://github.com/RaRe-Technologies/gensim/blob/ccffa84918dc13a6fbb5c82d6ba0e5e741266a3b/gensim/models/word2vec_inner.pyx#L943

The commit is from 2013 and only says "work around float/double sdot issues in scipy blas". On use case I can think of if if `sdot` comes from a `f2c`ed Fortran file, in which case indeed `sdot` returns a double.

packages/scipy/meta.yaml
```diff
@@ -97,13 +113,26 @@ build:
...
     # sed -i 's/^void/int/g' scipy/linalg/cython_lapack_signatures.txt
 
+    # Avoid Boost error "Sorry, but this platform does not have sufficient long
+    # double support for the special functions to be reliably implemented"
+    # Maybe this means result are garbage but at least it will compile ... no
+    # idea why I need to do this when switching to OpenBLAS ... maybe the type
+    # has changed somewhere and I get a boost template instantiation that I was
+    # not getting before???
+
+    sed -i 's/#if (defined(__NetBSD__) || defined(__EMSCRIPTEN__)/#if (defined(__NetBSD__)/' scipy/_lib/boost/boost/math/tools/config.hpp
```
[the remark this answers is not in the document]
lesteve: yep probably. A similar fix was done recently for Cygwin: https://github.com/scipy/scipy/pull/16370

pyodide-build/pyodide_build/_f2c_fixes.py
lesteve: Good point indeed, probably worth a PR to scipy

packages/gensim/meta.yaml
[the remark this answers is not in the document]
hoodmane: Okay, we should talk to the gensim people about this and see if we can upstream a solution.

hoodmane:
Regarding all my upstreaming suggestions: we should merge this PR first and then look into it.

packages/scipy/info.md
```diff
@@ -93,3 +62,19 @@ int str_len = 14;
...
+TO_KEEP='linalg|sparse|_lib|_build_utils'
+# Update scipy/setup.py
+perl -pi -e "s@(config.add_subpackage\(')(?!$TO_KEEP)@# \1\2@" scipy/setup.py
+# delete unwanted folders to avoid unneeded cythonization
+folders_to_delete=$(find scipy -mindepth 1 -maxdepth 1 -type d | grep -vP "$TO_KEEP")
+rm -rf $folders_to_delete
+```
+
+Building only `scipy.(linalg|sparse|_lib|_build_utils)` takes ~4 minutes on my
+machine compared to ~10-15 minutes for a full scipy build.
```
hoodmane, line 80: Thanks for adding this!

packages/libf2c/meta.yaml
```diff
@@ -1,19 +1,16 @@
 package:
-  name: CLAPACK
-  version: 3.2.1
+  name: libf2c
```
hoodmane, line 2: Why is this called libf2c? Also, can you explain somewhere why we need this in addition to openblas?

packages/openblas/patches/0001-Add-Wno-implicit-function-declaration-flag.patch
```diff
@@ -0,0 +1,25 @@
...
+diff --git a/Makefile.rule b/Makefile.rule
+index 5f787a9c..6890046a 100644
+--- a/Makefile.rule
++++ b/Makefile.rule
+@@ -228,7 +228,7 @@ NO_AFFINITY = 1
+ # Common Optimization Flag;
+ # The default -O2 is enough.
+ # Flags for POWER8 are defined in Makefile.power. Don't modify COMMON_OPT
+-# COMMON_OPT = -O2
++COMMON_OPT = -O2 -Wno-implicit-function-declaration -Wno-return-type
```
hoodmane, line 19: Can you explain why we need this again? (Maybe in the commit message for the patch.) Is it something we can work with openblas to fix?

packages/scipy/patches/0009-sasum-returns-double-not-float.patch
hoodmane, line 1: I guess this was another thing fixed by `f2c -R` =) There's so many pitfalls with this stuff...

packages/scipy/patches/0011-When-forward-declaring-print_soln-give-it-the-correc.patch
```diff
@@ -0,0 +1,70 @@
+From bb4e0d92ee2ba91ac5defd36ca5781433de25688 Mon Sep 17 00:00:00 2001
+From: Hood Chatham <roberthoodchatham@gmail.com>
+Date: Fri, 10 Mar 2023 14:47:00 +0000
+Subject: [PATCH 11/11] When forward declaring print_soln, give it the correct
+ signature
+
```
hoodmane, line 6: Maybe add a comment saying it's fixed by https://github.com/scipy/scipy/pull/18124

hoodmane:
This is looking pretty good @lesteve! You still have it marked as a draft, is there any significant task you think needs to be done before merging or just cleanup and finishing touches?

lesteve:
> You still have it marked as a draft, is there any significant task you think needs to be done before merging or just cleanup and finishing touches?

There is still a bit of cleanup I want to do, and then I'll mark this PR as ready for review! One thing I'd like to do as well is to run the full scipy test suite before merging to make sure there are not too many regressions. I have already done this from time to time and everything looks good, but doing it once before merging would still be good because it is so easy to mess something without realising, e.g. introduce a signature mismatch somewhere.

packages/libf2c/meta.yaml
lesteve: We are only using the f2clib part of CLAPACK so I renamed these files, it would have been weird in my opinion to have CLAPACK and OpenBLAS.

The f2c part is needed for the f2ced Fortran files in scipy for example to define things like `c_abs`. Note that the f2clib.a symbols are added in libopenblas.so in the OpenBLAS meta.yaml.

Not sure where to add this kind of info, maybe in the meta.yaml? Let me know if you see a good place for this!

packages/scipy/patches/0009-sasum-returns-double-not-float.patch
lesteve: Yes exactly. I think doing `f2c -R` does fix a lot of the float vs double return issues.

packages/libf2c/meta.yaml
hoodmane: > in the meta.yaml?

Yeah as a comment at the top of `meta.yaml` works.

packages/openblas/patches/0001-Add-Wno-implicit-function-declaration-flag.patch
lesteve: I'll add the following comment in the patch:

This is needed because we are changing many signatures to return int instead of void with some regex expressions but we are not modifying the returned value which would potentially be a lot more tricky.

hoodmane: What about no implicit function declaration?

lesteve: I think I removed the `-Wno-implicit-function-declaration`, maybe you are looking at an old diff?

hoodmane: Ah great!

lesteve:
The failure seems to be because the HTTPS certificate has expired for https://mpfr.loria.fr/mpfr-4.1.0/mpfr-4.1.0.tar.xz, oh well ...

packages/scipy/patches/0011-When-forward-declaring-print_soln-give-it-the-correc.patch
rth: Unresolving as I think it still needs the above  comment in the patch description.

rth (commented):
That's great, thanks a lot @lesteve !

> For now I renamed CLAPACK to libf2c, but we still download the full CLAPACK, this was easier to do it this way since we already have the necessary patches. 

I think it's fine this way, considering all the other things we download a few extra MB is not going to matter.

Personally this LGTM, I'll let Hood do the final review/approval.

rth:
> About the CI failure and the problematic HTTPS certificate, 

That was fixed in https://github.com/pyodide/pyodide/pull/3749

packages/scipy/patches/0011-When-forward-declaring-print_soln-give-it-the-correc.patch
lesteve: I actually have added the info in the scipy meta.yaml instead https://github.com/pyodide/pyodide/pull/3331/files#diff-d6d5fb2d234d6c6bafa1cff3848a6f49e87a62111e50baab21b4e954ae2b85be

ryanking13:
> For completeness, the macOS test-core failure does not seem related to this PR. It seems to happen from time to time on main e.g. https://github.com/pyodide/pyodide/actions/runs/4663743467/jobs/8255711856.

Yes, macOS CI is a bit flaky

ryanking13:
Thanks for your hard work @lesteve! I haven't been following this conversation closely, so I'll also let Hood approve it.

My only concern is that CI-wise this PR adds a library that takes over 10 minutes to build when it's not cached. But we are trying to work on it (#3544, #3341), so hopefully we can find a way around it.

hoodmane (approved):

hoodmane:
Okay merging. Thanks again @lesteve! Really valuable to have someone contributing stuff like this.

lesteve:
Great to see this one merged!

Thanks a lot for the comments and the help, in particular the session in Paris helped a lot to see how you debug the kind of issues I was facing!

rth:
@lesteve Were you planning of doing some benchmarks for say matrix multiplication before and after this change? You can take for instance 0.23.0 with CLAPACK and compare it with the dev deployment with OpenBLAS. I would expect for it to be a bit better but not massively better, at least until we optionally allow SIMD. So a next step could be to go and ask OpenBLAS people what could be done to make it better ^

Although to evaluate matrix multiplication we would probably need to use some low level scipy functions, since numpy still currently uses the built-in reference BLAS.

rth:
That works. Opened https://github.com/pyodide/pyodide/issues/3763 about it. Maybe I'll have a look later.

PR #3333, opened 2022-12-09
NFC Use initialization function to load _pyodide_core

hoodmane:
This adds `_pyodide_core` to the init tab and initializes it using that. This is a small step toward unvendoring the Pyodide ffi.


rth (approved):
Looks nice, thanks @hoodmane !

There are some canceled jobs in CI, do we need to restart it?

hoodmane:
Thanks for the review @rth!

PR #3334, opened 2022-12-09
Emscripten 3.1.28

hoodmane:
Closes #3313.

- [ ] Update changelog


hoodmane:
The failure here is due to a missing symbol called
```c++
lru11::Cache<OSRProjTLSCache::EPSGCacheKey, std::__2::shared_ptr<PJconsts>, lru11::NullLock, std::__2::unordered_map<OSRProjTLSCache::EPSGCacheKey, std::__2::__list_iterator<lru11::KeyValuePair<OSRProjTLSCache::EPSGCacheKey, std::__2::shared_ptr<PJconsts> >, void*>, OSRProjTLSCache::EPSGCacheKeyHasher, std::__2::equal_to<OSRProjTLSCache::EPSGCacheKey>, std::__2::allocator<std::__2::pair<OSRProjTLSCache::EPSGCacheKey const, std::__2::__list_iterator<lru11::KeyValuePair<OSRProjTLSCache::EPSGCacheKey, std::__2::shared_ptr<PJconsts> >, void*> > > > >::Cache(unsigned long, unsigned long)
```

hoodmane:
This symbol is from the gdal package. I unfortunately didn't build gdal with debug symbols but with the fiona symbols we see:
```
pyodide.asm.js:15120 Uncaught TypeError: Cannot read property 'apply' of undefined
    at stubs.<computed> (pyodide.asm.js:15119)
    at <anonymous>:wasm-function[4279]:0x3c4a43
    at <anonymous>:wasm-function[4278]:0x3c4335
    at _ZN19OGRSpatialReference7Private5clearEv (<anonymous>:wasm-function[3662]:0x37abeb)
    at _ZN19OGRSpatialReference15importFromEPSGAEi (<anonymous>:wasm-function[3757]:0x37feb6)
    at _ZN19OGRSpatialReference16SetFromUserInputEPKcPKS1_ (<anonymous>:wasm-function[3763]:0x38165d)
    at OSRSetFromUserInput (<anonymous>:wasm-function[3774]:0x382d40)
    at __pyx_pw_5fiona_4_crs_1crs_to_wkt (<anonymous>:wasm-function[110]:0x5b2a)
    at method_call_trampoline (pyodide.asm.js:9231)
```
@ryanking13 probably this is another one for you =)

ryanking13:
I'll look into it. It seems like the error is from libproj.

ryanking13:
I'm still investigating this issue, here is what I've found until now.

After updating Emscripten to 3.1.28, two "undefined" symbols are added to libproj.a:

- `osgeo::proj::lru11::Cache<unsigned long long, std::__2::vector<unsigned char, std::__2::allocator<unsigned char> >, osgeo::proj::lru11::NullLock, std::__2::unordered_map<unsigned long long, std::__2::__list_iterator<osgeo::proj::lru11::KeyValuePair<unsigned long long, std::__2::vector<unsigned char, std::__2::allocator<unsigned char> > >, void*>, std::__2::hash<unsigned long long>, std::__2::equal_to<unsigned long long>, std::__2::allocator<std::__2::pair<unsigned long long const, std::__2::__list_iterator<osgeo::proj::lru11::KeyValuePair<unsigned long long, std::__2::vector<unsigned char, std::__2::allocator<unsigned char> > >, void*> > > > >::Cache(unsigned long, unsigned long)`

- `std::__2::__list_iterator<osgeo::proj::io::Step, void*>::__list_iterator[abi:v15006]()`

These symbols didn't exist in libproj compiled with Emscripten 3.1.27.

I did bisecting emscripten releases and the problem happened in this commit: https://chromium.googlesource.com/emscripten-releases/+/cf45103f8016a93ac77860d6aea3b0d6187d8549
which consists of 41 changes in LLVM. I think it is a bug in LLVM, but unfortunately, it was not easy for me to spot a change that causes the issue. So I guess I should build LLVM locally and start bisecting again when I have time...

kleisauke:
Thanks for bisecting this LLVM issue! It sounds like the same issue mentioned in PR https://github.com/emscripten-core/emscripten/pull/18388. Using the latest tip-of-tree build of emsdk probably fixes this.

sbc100:
> Thanks for bisecting this LLVM issue! It sounds like the same issue mentioned in PR [emscripten-core/emscripten#18388](https://github.com/emscripten-core/emscripten/pull/18388). Using the latest tip-of-tree build of emsdk probably fixes this.

Perhaps you could also see if this issue is fixed with emscripten tot?  If it is that is great news.  But I would still be curious to know what kind of llvm change could have causes this breakage and how it could have gone undetected by our CI systems on the emscripten side.

kleisauke:
> Perhaps you could also see if this issue is fixed with emscripten tot?

Sure, but let me do this after the weekend. Feel free to investigate this earlier (I'm also very curious now which LLVM commit caused this and how we might avoid this issue in the future).

ryanking13:
Thanks! I'll try using emscripten tot next week.

kleisauke:
I bisected this a bit earlier (out of curiosity), I can confirm that the above LLVM revision range caused the link failure I ran into.

After manually reviewing those 41 revisions, I think this was caused by commit https://github.com/llvm/llvm-project/commit/a96a6ed83230265f3eab09a94f0e9525d05f8a74, see for example the differential revision in Phabricator: https://reviews.llvm.org/D136554.

For example, a user reported a similar linker failure on that commit: https://reviews.llvm.org/D136554#3982161.

ryanking13:
> Using the latest tip-of-tree build of emsdk probably fixes this.

I can confirm that this issue is fixed when I use tot emsdk.

hoodmane:
In that case we can just wait until version 3.1.29 comes out...

hoodmane:
Thanks everyone for looking into this!

ryanking13:
The issue was gone in 3.1.29. Thanks everyone for the help!

PR #3335, opened 2022-12-09
Build libpyodide.a

hoodmane:
This is work towards unvendoring the Pyodide foreign function interface. This is on top of #3333.

### Description

Prior to this point, we included a large amount of critical functionality with `--pre-js`. So we could create an archive called `libpyodide.a` with the object files but to use it you would have to pass `--pre-js _pyodide.out.js` at link time. This embeds all of this stuff in an object file called `pyodide_pre.o` which goes in our archive so you get all the needed js runtime by linking it. 

Of course someone trying to use this still has to get the Python code onto the import path, either using `--preload-file`, using Python to unpack it as a zip archive as we now do, with zipimporter, or otherwise. They also will have to link `libpython.a` (is CPython going to start distributing an Emscripten libpython?) and probably various other things.

We have to use a hack to inject the JavaScript code into the object files. The normal `EM_JS` macro cannot handle arbitrary JavaScript code -- for example it fails with many regex. Instead we manually generate write a C source file that does what we need using `xxd`. The generated C code is similar to what `EM_JS` generates, but it uses an array initializer rather than a string initializer for the characters avoiding the C preprocessor / compiler's strange opinions about strings.

This will be significantly more useful once we upgrade to Python 3.11.

### Checklists

- [ ] Add / update tests


rth:
Great work on this! 

As discussed in https://github.com/pyodide/pyodide/issues/3340 postponing after the 0.22 release unless you object.

hoodmane:
Nope that's how it should go. This is really most useful combined with Python 3.11. Then we can grab an official emscripten libpython311.a and link them together and get that working.

hoodmane:
@rth @ryanking13 would appreciate review on this.

Makefile
```diff
@@ -27,24 +27,45 @@ all: check \
 dist/pyodide_py.tar: $(wildcard src/py/pyodide/*.py)  $(wildcard src/py/_pyodide/*.py)
 	cd src/py && tar --exclude '*__pycache__*' -cf ../../dist/pyodide_py.tar pyodide _pyodide
 
-dist/pyodide.asm.js: \
+src/core/pyodide_pre.o: src/js/_pyodide.out.js src/core/pre.js
```
rth, line 30: A high  level comments with some explanations of what's done here (some of what you put in the description), wouldn't hurt!

Makefile
```diff
@@ -27,24 +27,45 @@ all: check \
...
 	src/core/_pyodide_core.o \
 	src/core/js2python.o \
 	src/core/jsproxy.o \
-	src/core/main.o  \
 	src/core/pyproxy.o \
 	src/core/python2js_buffer.o \
 	src/core/python2js.o \
-	src/js/_pyodide.out.js \
+	src/core/pyodide_pre.o
+	emar rcs dist/libpyodide.a $(filter %.o,$^)
```
rth, line 58: We won't be able to server .a files from the CDN without whitelisting the extension (which requires a manual request). Does the extension matters much here: A libpyodide.wasm or libpyodide.data wouldn't be correct?

From a MIME type point of view,it would be [`application/octet-stream`](https://mimetype.io/application/octet-stream)  or `application/wasm` anyway as far as I understand.

rth (commented):
Thanks @hoodmane. A couple of comments, otherwise LGTM.

Makefile
hoodmane: I'm not sure: if you say `-lpyodide` and you have a file called `libpyodide.data` it will probably be linked? Note that `.a` files are archives though and not wasm files.

hoodmane: I think these would be distributed more from github releases than from the cdn though.

Makefile
```diff
@@ -27,24 +27,45 @@ all: check \
...
+	rm -f src/core/pyodide_pre.gen.c
+	echo '__attribute__((used)) __attribute__((section("em_js"), aligned(1)))' >> src/core/pyodide_pre.gen.c
+	echo 'char __em_js__pyodide_js_init[] = {'  >> src/core/pyodide_pre.gen.c
+	cat tmp.dat  | xxd -i - >> src/core/pyodide_pre.gen.c
+	echo ', 0};' >> src/core/pyodide_pre.gen.c
+
+	rm tmp.dat
+	emcc -c src/core/pyodide_pre.gen.c -o src/core/pyodide_pre.o
+
+dist/libpyodide.a: \
```
ryanking13, line 47: Do we want to include this in the dist archive?

ryanking13 (approved):
Great, thanks @hoodmane!

Makefile
hoodmane: Not clear. I'm not sure where else to put it. We want it:
* as a separate github release
* in the circleci artifacts

Let's put it in dist for now and we can move it later when we figure out where to put it.

ryanking13: > Let's put it in dist for now and we can move it later when we figure out where to put it.

Sounds good to me.

Issue #3340, opened 2022-12-12
Release 0.22


ryanking13:
Last week, Roman and I talked about releasing Pyodide 0.22 soon (maybe before the new year, but can be delayed as most of us might go winter vacation).

This issue is to summarize what needs to be done before releasing 0.22. I listed things that I can remember. Please add more if you think something is need to be done before 0.22.

__What needs to be done before 0.22__

- [x] Finish Hood's work on JsProxy/PyProxy iterator / generator
  - #3346
  - #3317
- [x] #3268
- [x] Release micropip 0.2.0 and use it in Pyodide  https://github.com/pyodide/pyodide/pull/3347
  - Pyodide docs should have a link to micropip docs.
  - Link external references using [intersphinx](https://www.sphinx-doc.org/en/master/usage/extensions/intersphinx.html) plugin.
- [x] #3263
- [x] #3308
- [x] ~~#3092 (probably we need to rename the key before release)~~
- [x] Clean up the changelog (regroup in subsections)  #3370 
- [x] Make sure all major features are documented (including new CLI, and venv creation).  
       - https://github.com/pyodide/pyodide/pull/3367
- [x]  #3349
- [x] Add release CI for pyodide-py https://github.com/pyodide/pyodide/issues/1839

__Things that are not going be included in 0.22__

- Update to Python 3.11: #3252
- Bundle the standard library in a zip file: #3166
- #3269
- libpyodide.a #3335

hoodmane:
Thanks @ryanking13!

rth:
> More general unvendoring mechanism #3092

IMO this could wait for 0.23.

The other things I added to the 0.22 TODO list is,
>  Clean up the changelog (regroup in subsections)
> Make sure all major features are documented (including new CLI, and venv creation).

hoodmane:
We should go ahead and make the release whenever someone gets to it.

PR #3347, opened 2022-12-12
Update to micropip v0.2.0

rth:
Also links to the new micropip docs.

Links to the micropip API reference should now be resolved to https://micropip.pyodide.org thanks to intersphinx. Though if we provide direct links, I'm not sure how to preserve a compatible version.

xref https://github.com/pyodide/pyodide/issues/3340

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -11,6 +11,8 @@ from pathlib import Path
 from typing import Any
 from unittest import mock
 
+import micropip
+
 panels_add_bootstrap_css = False
 
 # -- Project information -----------------------------------------------------
@@ -33,6 +35,7 @@ else:
 extensions = [
     "sphinx.ext.autodoc",
     "sphinx.ext.autosummary",
+    "sphinx.ext.intersphinx",
     "sphinxcontrib.napoleon",
     "myst_parser",
     "sphinx_js",
@@ -43,6 +46,7 @@ extensions = [
     "versionwarning.extension",
     "sphinx_issues",
 ]
+
 
 myst_enable_extensions = ["substitution"]
 
@@ -63,6 +67,10 @@ versionwarning_body_selector = "#main-content > div"
 
 autosummary_generate = True
 autodoc_default_flags = ["members", "inherited-members"]
+
+intersphinx_mapping = {
+    "micropip": (f"https://micropip.pyodide.org/en/v{micropip.__version__}/", None)
+}
 
 # Add modules to be mocked.
 mock_modules = ["ruamel.yaml", "tomli"]
--- a/docs/usage/api-reference.md
+++ b/docs/usage/api-reference.md
@@ -6,6 +6,6 @@
 
    api/js-api.md
    api/python-api.md
-   api/micropip-api.md
+   Micropip API <https://micropip.pyodide.org/en/stable/project/api.html>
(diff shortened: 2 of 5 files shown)


docs/conf.py
```diff
@@ -64,6 +68,10 @@
 autosummary_generate = True
 autodoc_default_flags = ["members", "inherited-members"]
 
+intersphinx_mapping = {
+    "micropip": (f"https://micropip.pyodide.org/en/v{micropip.__version__}/", None)
```
rth, line 72: New versions should also be released there on tags, but we will see.

docs/usage/api-reference.md
```diff
@@ -6,6 +6,6 @@
 
    api/js-api.md
    api/python-api.md
-   api/micropip-api.md
+   Micropip API <https://micropip.pyodide.org/en/stable/project/api.html>
```
rth, line 9: Or we just remove it, not sure

rth:
So there should be a way to redirect from https://pyodide.org/en/stable/usage/api/micropip-api.html to https://micropip.pyodide.org/en/stable/project/api.html [in readthedocs](https://docs.readthedocs.io/en/stable/user-defined-redirects.html) but so far I haven't managed to make it work. Maybe it only works for newly deployed versions. Not sure.
image

packages/micropip/meta.yaml
```diff
@@ -1,14 +1,14 @@
 package:
   name: micropip
   # This version needs to be consistent with docs/requirements-doc.txt
-  version: "0.1.0"
+  version: "0.2.0"
   _tag: pyodide.stdlib
```
ryanking13, line 5:
```suggestion
```

We don't need this tag anymore as it is now versioned.

docs/usage/api-reference.md
ryanking13: Maybe we should also link to specific version of micropip here? If not, when we release a new micropip version, old Pyodide versions will have old micropip but will be linked to latest micropip docs.

ryanking13 (approved):
Thanks @rth!

docs/usage/api-reference.md
ryanking13: Perhaps we can parse micropip version directly from meta.yaml file. It is quite hacky, but we are already doing something similar [here](https://github.com/pyodide/pyodide/blob/62dcf9119f406fe551390b899f0a0f0a5b3529fc/docs/sphinx_pyodide/sphinx_pyodide/packages.py#L25-L26)

rth: Thanks for the pointer. That's a possibility but I don't really want to create a sphinx plugin just for this. Ideally, users would never see this page once we figure out redirects. All the other links should redirect to the matching micropip version. Overall older versions of Pyodide should work with newer versions of micropip, so it's probably not that critical.

hoodmane: But if people never actually see this page for some reason then I agree it's not a big deal...

rth: Actually, just sphinx ref to `micropip` should link to the correct version, without any hacks, as we are doing in https://pyodide.org/en/latest/usage/loading-packages.html#installing-packages 

Will fix in another PR

PR #3349, opened 2022-12-13
Rename unvendored _hashlib to hashlib

ryanking13:
Rename _hashlib to hashlib so it can be loaded via micropip.

See: https://github.com/pyodide/micropip/pull/41


docs/usage/wasm-constraints.md
```diff
@@ -22,7 +22,7 @@ Instead, it is better to load individual modules as needed using
 - ssl
 - lzma
 - sqlite3
-- \_hashlib
+- \_hashlib (you need to load it via `micropip.install('hashlib')`)
```
rth, line 25: IMO that's a private module so it doesn't make much sense to mention it since it's not documented what it does. Instead  we should rather indicate how this affects the `hashlib` public module.

ryanking13:
> So what happens to hashlib if _hashlib is not installed: it's usable but with very little functionality.

Yes, the list of algorithms available will be limited without _hashlib (openssl):

without _hashlib:

```
>>> hashlib.algorithms_available
{'sha384', 'blake2b', 'md5', 'shake_128', 'blake2s', 'sha256', 'shake_256', 'sha3_224', 'sha3_256', 'sha3_512', 'sha224', 'sha1', 'sha3_384', 'sha512'}
```

with _hashlib:

```
>>> hashlib.algorithms_available
{'sm3', 'mdc2', 'sha512_256', 'whirlpool', 'sha1', 'sha224', 'sha3_512', 'sha3_256', 'md5', 'md4', 'shake_256', 'blake2s', 'sha512_224', 'shake_128', 'blake2b', 'md5-sha1', 'sha384', 'sh
a3_224', 'sha512', 'sha3_384', 'ripemd160', 'sha256'}
```

docs/usage/wasm-constraints.md
ryanking13: I removed the detail of the private module and added some information about `hashlib`.

ryanking13:
Yeah, hashlib is quite an annoying case. I am not sure there is an optimal way to deal with this unless we link OpenSSL statically to libpython.

rth:
Thank you!

Issue #3292, opened 2022-11-25
Package python-libarchive-c


bollwyvl:
## 🐍 Package Request

- Package Name and Version: `python-libarchive-c` / `latest`
- Package URL:
  - [github](https://github.com/Changaco/python-libarchive-c)
  - [pypi](https://pypi.org/project/libarchive-c/)
- Package Dependencies that needs to be resolved first:
  - `libarchive` (C lib) 
    - [npm](https://www.npmjs.com/package/libarchive-wasm) 
    - [conda-forge](https://github.com/conda-forge/libarchive-feedstock/blob/main/recipe/meta.yaml)

motivation: recently have need for looking at some `.zst` files: there are several bindings, but ol' libarchive has been around longer, is used in more places, and can do all the other things as well.

PR #3353, opened 2022-12-14
Patch ctypes.util.find_library to support Emscripten platform

ryanking13:
### Description

Several packages use `ctypes.util.find_library` to search shared libraries. But it does not support Emscripten platform. I opened a upstream PR about this (https://github.com/python/cpython/pull/99950), but it isn't accepted as CPython WASM build has no support for ctypes now.

This PR adds a patch that adds Emscripten platform support for ctypes.util.find_library: searching libraries from paths that ar in LD_LIBRARY_PATH env variable.

Note that [Emscripten uses LD_LIBRARY_PATH for `dlopen` internally](https://github.com/emscripten-core/emscripten/blob/efc18ae6e268373ff984d1622979422f99f3db0a/src/library_dylink.js#L965-L968), though I think it is not documented :/

This is also required for #3292 

### Checklists

- [x] Add / update tests


rth:
Looks very nice, thank ! I'll let Hood approve.

src/js/dynload.ts
```diff
@@ -32,7 +32,7 @@ function createDynlibFS(
   readFileFunc?: ReadFileType,
 ): LoadDynlibFS {
   const dirname = lib.substring(0, lib.lastIndexOf("/"));
-  const defaultSearchDirs = [API.dsodir, API.sitepackages, dirname];
+  const defaultSearchDirs = API.defaultLdLibraryPath + [dirname];
```
hoodmane, line 35: You can't add lists like this in Typescript can you? You are probably thinking of Python.

hoodmane (approved):
Generally looks good to me though I am confused about list addition in typescript (see comment).

src/js/dynload.ts
ryanking13: Good point, Thanks! It seems like this was the reason why I was getting mysterious errors 😂

PR #3354, opened 2022-12-14
Fix misconfigured meta.yaml files

ryanking13:
Fixes some misconfigured meta.yaml files that I found in #3192.

### DIFF
--- a/packages/scipy/meta.yaml
+++ b/packages/scipy/meta.yaml
@@ -111,6 +111,7 @@ requirements:
   host:
     - numpy
     - CLAPACK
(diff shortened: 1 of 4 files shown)


packages/scipy/meta.yaml
```diff
@@ -111,6 +111,7 @@ requirements:
   host:
     - numpy
     - CLAPACK
+    - boost-cpp
```
ryanking13, line 114: boost is not a hard dependency of scipy, but it seems like scipy uses it if it is found during build.

PR #3355, opened 2022-12-14
PKG Add pyodide-http

rth:
[pyodide-http](https://github.com/koenvo/pyodide-http) is a pure Python package by @koenvo. I feel it's useful for it to be part of the default distribution, so it could be loaded from imports without having to load micropip first.

Also added it to related projects.


koenvo:
Thanks a lot for including pyodide-http  🙏

PR #3358, opened 2022-12-15
Relocate files inside tools directory

ryanking13:
### Description

This PR relocates files inside `tools` directory so that `tools` directory doesn't have to be vendored when creating xbuildenv.

- cmake, pyo3 config ==> moved to `pyodide-build`.
- venv python script ==> moved to templates, then copied to dist during build.


hoodmane (approved):
Thanks @ryanking13!

PR #3362, opened 2022-12-18
NFC Minor python2js string conversion cleanup

hoodmane:
We remove `hiwire_string_ascii` in favor of `hiwire_string_utf8` (which I think is faster).
The `hiwire_string_ucs#` functions are only used in python2js for string conversion and
should only be used there I think so I moved them out of hiwire into that file.

I keep getting confused about the implementation of python2js string conversion
functions so I added a little FAQ for my future self. I also replaced the error in the
invalid string kind case with an `assert false` since CPython itself uses assertions to
check that a string has a valid kind.


ryanking13 (approved):
Nice FAQ. Thanks!

Issue #20, opened 2018-04-04
Modularization


mdboom:
This is a big one.
It would be nice to not have to build all libraries into one monolithic `.data` file, but to have individual libraries download on demand.  This might take the form of an "overlay filesystem" (a general emscripten solution), or a custom import hook (a Python-specific solution).
Also, modularizing the build system to make it easier to add new libraries would be nice.  Bonus points to build such a tool on top of Anaconda.

PR #3367, opened 2022-12-18
Add Pyodide CLI documentation

rth:
This adds an auto-generate section about the  Pyodide CLI. 

Requires https://github.com/pyodide/pyodide-cli/pull/17. There is no typer plugin for sphinx, so we have to use [sphinx-click](https://sphinx-click.readthedocs.io/en/latest/usage/) (since typer is a wrapper around click it's not an issue).

The question is what we should do about other packages that register subcommands in the Pyodide CLI (e.g. emscripten-auditwheel, pyodide-pack etc):
 - either we do nothing but they are a bit harder to discover.
 - or we add them to `docs/requirements-doc.txt` and the corresponding CLI docs are generated. But then,
   - a) it's a bit unclear how to manage versions of these packages
   - b) the corresponding subcommands don't explicitly indicate that they need this external package. We should probably fix this in any case.
 - or maybe something in the middle, where we additionally link to the home page of those packages and let them document this. Probably the best solution, I'll go with it for now.

About `pyodide venv` we need to document it in a separate PR, tag it as experimental, and explain the current limitations. Probably the command description needs updating, but maybe also add a section somewhere in  "How to use Pyodide".

Also opened a few issues in https://github.com/ryanking13/auditwheel-emscripten/issues related to this.


rth:
>  what we should do about other packages that register subcommands in the Pyodide CLI (e.g. emscripten-auditwheel)

Though actually since emscripten-auditwheel is now a dependency of pyodide-build it means it will be installed and auto-documented here. This still raises the question of how to manage its version and that we should be more explicit about where each command is defined in.

rth:
OK, to fix CI we would need to merge https://github.com/pyodide/pyodide-cli/pull/17 and make a `pyodide-cli` release. As for some reason, pip is not happy with installing from a git branch, as a temporary solution, and it's not worth investigating.

hoodmane:
There are some minimal docs on virtual environments here:
https://pyodide.org/en/latest/development/building-and-testing-packages.html
probably a more detailed explanation of the capabilities and limitations would be good...

rth:
RTD CI failure due to https://github.com/pyodide/pyodide/issues/3375 but otherwise this is ready. You can see the last working build in https://pyodide--3367.org.readthedocs.build/en/3367/ before I merged with main which introduced the failure.

> Hmm, probably we need to find a way to "not" auto-document those packages.

If the users see these commands when they install pyodide-build it's also a bit strange not to document them. I was thinking we could maybe just improve the CLI subcommand docs as to where a given subcommand is defined https://github.com/pyodide/pyodide-cli/issues/18 but this could also be done later.

rth:
Can anyone review, please? It's one one of the few remaining blockers for v0.22 https://github.com/pyodide/pyodide/issues/3340

pyodide-build/setup.cfg
```diff
@@ -29,7 +29,7 @@ install_requires =
     build==0.7.0
     virtualenv
     pydantic>=1.10.2
-    pyodide-cli>=0.2.0
+    pyodide-cli~=0.2.1
```
ryanking13, line 32: Probably let's release `pyodide-cli` 0.2.2 including #20. We can also update `auditwheel-emscripten` to 0.0.9 which renamed the entrypoint to `pyodide auditwheel`.

rth: Agreed. `~=0.2.1` is equivalent to `==0.2.*` so we can still merge this and it will use the latest versions once those releases are made.

docs/requirements-doc.txt
```diff
@@ -10,7 +10,11 @@ sphinx-issues==2.0.0
 sphinx-js==3.1
 sphinx-version-warning~=1.1.2
 sphinx-panels
+sphinx-click
 markupsafe<2.1.0
 pydantic
+pyodide-cli~=0.2.1
+# Packages that we want to document as part of the Pyodide CLI
+./pyodide-build/
```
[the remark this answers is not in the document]
rth: Good point. Removed. Thanks for the review!

PR #3369, opened 2022-12-18
Don't destroy roundtrip PyProxies automatically

hoodmane:
There's a bunch of places where we want to destroy a `PyProxy` if it is being
handed back into Python. This way, if the user wants to keep the proxy around
they can return a copy() of it (whereas otherwise they would have no way to
destroy it after the JavaScript execution is ended).

```js
function new_dict(){
  const dict = pyodide.globals.get("dict");
  const result = dict([[1,2],[3,4]]);
  return result;
  // Proxy is destroyed after it is returned!
}
```

If the `PyProxy` has roundtrip set, however, it is generally a bad idea to
destroy it in these cases. In many cases, this means that the object returned
to Python is actually unusable (because we get a destroyed double proxy).

One slightly annoying question is how to deal with the case where (1) we
want the return value NOT to be destroyed and (2) it may or may not be roundtrip

```
function f(px){
   // We don't want px to be destroyed so we have to return a copy (the copy
   // gets destroyed instead). But if px is roundtrip then the copy is also
   // roundtrip and neither gets destroyed so there is a leak. What to do???
   return px.copy();
}
```

### Checklists

- [x] Add / update tests

### DIFF
--- a/docs/project/changelog.md
+++ b/docs/project/changelog.md
@@ -83,7 +83,7 @@ substitutions:
   If this is set to `True`, then when the proxy is converted back to Python, it
   is converted back to the same double proxy. This allows the proxy to be
   destroyed from Python even if no reference is retained.
-  {pr}`3163`
+  {pr}`3163`, {pr}`3369`
 
 - {{ Enhancement }} Pyodide now shows more helpful error messages when
   importing packages that are included in Pyodide fails.
--- a/docs/project/deprecation-timeline.md
+++ b/docs/project/deprecation-timeline.md
@@ -12,13 +12,15 @@ deprecation warnings. More details about each item can often be found in the
 
 ## 0.24.0
 
-- The `messageCallback` and `errorCallback` argument to `loadPackage` and `loadPackagesFromImports` will be passed as a
-  named argument only.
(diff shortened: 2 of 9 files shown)


rth:
Do you want to include this in 0.22?

hoodmane:
I think it would be nice to include. It makes `create_proxy` work a bit more reasonably.

ryanking13 (approved):
There is a failing test, otherwise LGTM. Thanks @hoodmane!

hoodmane:
We should also add a deprecation message for use like `pyproxy.destroy("msg")` rather than `pyproxy.destroy({msg:"msg"})`. Also, maybe it should be spelled `message` instead of `msg`?

ryanking13:
> We should also add a deprecation message for use like `pyproxy.destroy("msg")` rather than `pyproxy.destroy({msg:"msg"})`

Sounds good to me. (In that case, please also update the [deprecation timeline](https://github.com/pyodide/pyodide/blob/main/docs/project/deprecation-timeline.md)!)

> Also, maybe it should be spelled `message` instead of `msg`?

I'm +1 to it.

PR #3371, opened 2022-12-19
Fix top level multiprocessing.connection import and update yt

hoodmane:
This updates yt. The new yt has a top level `multiprocessing.connection` import so we patch that so that it can be imported without raising an error.

- [x] Add / update tests

### DIFF
--- a/cpython/patches/0015-Allow-multiprocessing.connection-top-level-import.patch
+++ b/cpython/patches/0015-Allow-multiprocessing.connection-top-level-import.patch
@@ -1,0 +1,28 @@
+From becbcc43e3e09a9c065ea7856a2d3069061fd570 Mon Sep 17 00:00:00 2001
+From: Hood Chatham <roberthoodchatham@gmail.com>
+Date: Mon, 19 Dec 2022 09:09:14 -0800
+Subject: [PATCH 15/15] Allow multiprocessing.connection top level import
+
+---
+ Lib/multiprocessing/connection.py | 5 ++++-
+ 1 file changed, 4 insertions(+), 1 deletion(-)
+
+diff --git a/Lib/multiprocessing/connection.py b/Lib/multiprocessing/connection.py
+index 510e4b5aba..372f9ced37 100644
+--- a/Lib/multiprocessing/connection.py
++++ b/Lib/multiprocessing/connection.py
+@@ -18,7 +18,10 @@
+ import tempfile
+ import itertools
+ 
+-import _multiprocessing
++try:
++    import _multiprocessing
++except ModuleNotFoundError:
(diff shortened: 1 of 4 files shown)


rth:
Yeah, we have been going a bit back and forth on this. The problem is if we add such an empty file, then it's hard to direct at import time that Python is built without multiprocessing. This is occasionally done by CPython itself (e.g. [here](https://github.com/python/cpython/blob/c1d93b6411f975d67e43942f1a2745a22983c18c/Lib/test/test___all__.py#L10)). So the general logic for missing functionality so far is if it's not available, the  `_` prefixed module should fail to import.

Some packages have now come to rely on the current functionality (e.g. https://github.com/joblib/joblib/pull/1256 added a fix specifically for Pyodide), so we cannot easily change this behavior without breaking working code. We should probably have a test for the current behavior with a comment explaining it. Unless we have a solution we are confident would address all the use-cases, and I'm not convinced this is really it.

In this case, IMO it would be better to make changes to the multiprocessing module so it's importable but not usable without `_multiprocessing` but we would need to talk to CPyton devs about it, to make it more permanent. Though maybe the easiest is for astropy to handle the case where multiprocessing fails to import.

rth:
> I think yt is the issue here, not astropy

Yes, wrong astrophysics related package :)

> Arguably, depending on import _multiprocessing failing isn't a great way to detect that multi-processing is unavailable. Especially given that import multiprocessing works.

Agree that the situation is not ideal. IMO this kind of thing should be specified by CPython as to what should happen when a stdlib functionality is not available. I opened a [discuss.python.org thread](https://discuss.python.org/t/expected-behavior-for-unsupported-stdlib-modules-in-the-browser/15689/15) about it some time ago, but there was no clear conclusion, as far as I remember.

hoodmane:
Right. This also inconveniently breaks the tests. Let's just patch yt then.

hoodmane:
@rth how about this?

hoodmane:
> Arguably, depending on import _multiprocessing failing isn't a great way to detect that multi-processing is unavailable. Especially given that import multiprocessing works.

It would be nice to have a setting `multiprocessing.DISABLED` or something like that.

rth (approved):
Sounds good to me, thanks!

> It would be nice to have a setting multiprocessing.DISABLED or something like that.

Maybe, but we are not the only platform/build with these problems so we shouldn't have to invent new things either IMO. Similarly this kind of applies to many other optional stdlib modules, that may be excluded in the default Pyodide build. For instance, I just discovered there was a [PEP 534](https://peps.python.org/pep-0534/) (deferred since 2016)  which seems very relevant to this problem.

PR #3376, opened 2022-12-20
Add `pyodide config` command to show config variables used in Pyodide

ryanking13:
### Description

This adds `pyodide config` command which shows config variables, which is for now subset of env variables in `Makefile.envs` used in pyodide-build.

The common use case of this command is to find an emscripten version used in a specific version of Pyodide, so that one can do:

```bash
$ pip install pyodide-build
$ emsdk install $(pyodide config PYODIDE_EMSCRIPTEN_VERSION)
$ pyodide build
```

I would like to include this in 0.22 if possible.

### Checklists

- [x] Add / update tests


ryanking13:
Well it still fails with a different error :(

ryanking13:
Thanks for the ideas @rth! I 100% agree with your thoughts.

> 1. would we even need to set the config via this CLI and if yes how would the API look like ?

I'm +1 with `pip` style API.

> 2\. Do we need all those env variables exposed in practice, and if we do do we promise some backward compatibility on those ?

Good point. probably we should not directly expose env variable itself, but instead make a list of variables we would like to provide with config API. So one can use it with `pyodide config get emscripten_version`.

ryanking13:
Okay, I changed this to pip-style API, with two subcommands available. We can probably add a write option when we migrate Makefile to something else.

- pyodide config list
- pyodide config get

For now, I exposed two config variables: `emscripten_version` and `python_version`. So that people can use it when 1) installing emscripten 2) installing python with pyenv, etc. Though installing python with Pyodide API is kind of a chicken and egg problem as it needs Python to run the CLI 🤔.

rth (approved):
LGTM, thanks!

PR #3381, opened 2022-12-22
Fix some function signature mismatches in scipy PROPACK

### LINKED ISSUE #3380: scipy.sparse.linalg._svdp signature mismatch
### To Reproduce

```py
import numpy as np
from scipy.sparse.linalg._svdp import _svdp

np.random.seed(0)
n, k = 70, 10
A = np.random.random((n, n))
_svdp(A, k, kmax=5*k)
```

Stack-trace with debug symbols:
```
Uncaught RuntimeError: null function or function signature mismatch
    at dgetu0_ (00055cfe:0x9d7e)
    at dlanbpro_ (00055cfe:0xd2b2)
    at dlansvd_irl__ (00055cfe:0xa19e)
    at dynCall (pyodide.asm.js:16146:40)
    at pyodide.asm.js:16154:18
    at stubs.<computed> (pyodide.asm.js:16567:33)
    at f2py_rout__dpropack_dlansvd_irl (00055cfe:0x75a7)
    at fortran_call (0003e3ba:0x6642)
    at _PyObject_Call (pyodide.asm.wasm:0x131064)
    at PyObject_Call (pyodide.asm.wasm:0x13109d)
```

Debugging with Chromium, the function is called with 7 arguments with a i32 return type:
```
call_indirect (param i32 i32 i32 i32 i32 i32 i32) (result i32)
```
but wasm functions has 8 arguments and returns f32
```
(func $cb_daprod_in___user__routines (;111;) (param $var0 i32) (param $var1 i32) (param $var2 i32) (param $var3 i32) (param $var4 i32) (param $var5 i32) (param $var6 i32) (param $var7 i32) (result f32)
```

Not sure where this comes from ...

### Expected behavior

No error

### Environment

- Pyodide Version: both stable (0.21.3) and latest
- Browser version: not relevant

### Additional context

Bumped into this when looking at `scipy.sparse` test suite issues in https://github.com/lesteve/scipy-tests-pyodide/issues/3

hoodmane:
This is an attempt to resolve #3380. There seems to be three problems:
1. The usual business with fortran enums, string arguments and ftnlen for the `transa` parameter.
2. The `aprod` argument is declared to be a subroutine, which should have return type i32 but it in fact has return type f32, so we'd better say it's a function with return type real
3. For reasons I cannot comprehend, when I say the return type is real, f2c generates a declaration that says it has return type double. It feels like a bug in f2c but it is so major and obvious that it's hard to believe anyone could have missed it. Anyways I hack around it by adding a little bit more nonsense to f2cfixes.

This probably isn't quite right yet since running `_svdp` doesn't crash but it prints a very large number of the following warning:
```
Warning: call-back function cb_daprod_in___user__routines did not provide return value (index=0, type=float)
```

### Checklists

- [x] Add / update tests

### DIFF
--- a/packages/scipy/meta.yaml
+++ b/packages/scipy/meta.yaml
@@ -32,6 +32,7 @@ source:
     - patches/0011-Change-qh_eachvoronoi_all-FILE-type.patch
     - patches/0012-Remove-meson-requirements.patch
     - patches/0013-Fix-fitpack.patch
+    - patches/fix-daprod.patch
 
 build:
   cflags: |
@@ -39,6 +40,7 @@ build:
     -I$(HOSTSITEPACKAGES)/pythran/
     -Wno-return-type
     -DUNDERSCORE_G77
+    -DDEBUGCFUNCS
   ldflags: |
     -L$(NUMPY_LIB)/core/lib/
     -L$(NUMPY_LIB)/random/lib/
--- a/packages/scipy/meta.yaml
+++ b/packages/scipy/meta.yaml
@@ -40,7 +40,7 @@ build:
     -I$(HOSTSITEPACKAGES)/pythran/
     -Wno-return-type
     -DUNDERSCORE_G77
-    -DDEBUGCFUNCS
+  # -DDEBUGCFUNCS
   ldflags: |
     -L$(NUMPY_LIB)/core/lib/
     -L$(NUMPY_LIB)/random/lib/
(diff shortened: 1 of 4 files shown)


hoodmane:
It's certainly not functioning correctly  -- the results are very different from running it under native scipy.

rth:
Thanks for looking into it, still.

lesteve:
> For reasons I cannot comprehend, when I say the return type is real, f2c generates a declaration that says it has return type double. It feels like a bug in f2c but it is so major and obvious that it's hard to believe anyone could have missed it.

I bumped into https://gcc.gnu.org/onlinedocs/gfortran/Code-Gen-Options.html which may be relevant:

> The calling conventions used by g77 (originally implemented in f2c) require functions that return type default REAL to actually return the C type double, and functions that return type COMPLEX to return the values via an extra argument in the calling sequence that points to where to store the return value. Under the default GNU calling conventions, such functions simply return their results as they would in GNU C—default REAL functions return the C type float, and COMPLEX functions return the GNU C type complex. Additionally, this option implies the -fsecond-underscore option, unless -fno-second-underscore is explicitly requested.

PR #3383, opened 2022-12-23
Update auditwheel-emscripten version to 0.0.9

ryanking13:
I changed the subcommand name from `pyodide audit` to `pyodide auditwheel`, regarding the suggestion of @rth https://github.com/ryanking13/auditwheel-emscripten/issues/4

It affects the docs only (pyodide CLI).


pyodide-build/setup.cfg
```diff
@@ -35,7 +35,7 @@ install_requires =
     requests
     types-requests
     typer
-    auditwheel-emscripten==0.0.8
+    auditwheel-emscripten==0.0.9
```
rth, line 38: Not critical but we could also pin to a range of versions, e.g.
```
auditwheel-emscripten>=0.0.9,<0.1.0
```
so that if there are any bugs, they can be fixed without waiting for a new pyodide-build release. Which would be the advantage of having this as a separate package.

ryanking13: Yeah, sounds good to me.

PR #3384, opened 2022-12-23
Fix scipy exception handling

### LINKED ISSUE #3382: Code creates Pyodide fatal error rather than Python exception that can be caught
## 🐛 Bug

Some exceptions (I am guessing which are thrown from Cython or compiled code) creates Pyodide fatal errors rather than Python exception which can be caught.

### To Reproduce

```py
import numpy as np
from scipy.spatial.distance import cdist

out = np.ones((2, 2))
arr = np.array([[1, 2]])

try:
    cdist(arr, arr, out=out)
except ValueError:
    print('exception caught')
```

This creates a Pyodide fatal error. With debug symbols the stack-trace looks like this:
```
CppException std::invalid_argument: Output array has incorrect shape.
    at convertCppException (http://localhost:8000/pyodide.asm.js:5740:16)
    at API.fatal_error (http://localhost:8000/pyodide.asm.js:5758:17)
    at Module.callPyObjectKwargs (http://localhost:8000/pyodide.asm.js:4425:17)
    at Module.callPyObject (http://localhost:8000/pyodide.asm.js:4440:25)
    at wrapper (http://localhost:8000/pyodide.asm.js:9248:27)
```

Another similar one:
```py
from scipy.sparse._sparsetools import test_throw_error
try:
    test_throw_error()
except MemoryError:
    print('caught exception')
```

The code for `test_throw_error` is written in C https://github.com/scipy/scipy/blob/e2bb12727dad4c7c575abda22b459aeb35962ab8/scipy/sparse/sparsetools/csr.h#L1698-L1701:
```C
inline int test_throw_error() {
    throw std::bad_alloc();
    return 1;
}
```

Another one from Cython code:

```py
from scipy.signal import lombscargle

try:
    lombscargle(x=[1], y=[1, 2], freqs=[1, 2, 3])
except ValueError:
    print('caught exception')
```

Stack-trace with debug symbols:
```
CppException (anonymous namespace)::pythonic::types::ValueError: std::exception
    at convertCppException (http://localhost:8000/pyodide.asm.js:5740:16)
    at API.fatal_error (http://localhost:8000/pyodide.asm.js:5758:17)
    at Module.callPyObjectKwargs (http://localhost:8000/pyodide.asm.js:4425:17)
    at Module.callPyObject (http://localhost:8000/pyodide.asm.js:4440:25)
    at wrapper (http://localhost:8000/pyodide.asm.js:9248:27)
```

The exception is raised in Cython code https://github.com/scipy/scipy/blob/e2bb12727dad4c7c575abda22b459aeb35962ab8/scipy/signal/_spectral.pyx#L55-L56

### Expected behavior

This should catch the exception and print `exception caught`

### Environment

- Pyodide Version: both stable and latest
- Browser version: not relevant

### Additional context

I saw this kind of issues a few times in scipy test suite tests. Typically they arise in tests that check that code is raising an exception.

hoodmane:
Resolves #3382.

- [x] Add / update tests

### DIFF
--- a/docs/project/changelog.md
+++ b/docs/project/changelog.md
@@ -305,15 +305,12 @@ substitutions:
 - {{ Fix }} Scipy CSR data is now handled correctly in XGBoost.
   {pr}`3194`
 
-- {{ Update }} Upgraded SciPy to version 1.9.1.
-  {pr}`3043`
+- {{ Update }} Upgraded packages: SciPy 1.9.1 {pr}`3043`, pandas 1.5.0 {pr}`3134`,
+  numpy 1.23.3 {pr}`3284`, scikit-learn 1.1.3 {pr}`3324` as well as most of the other packages
+  {pr}`3348` {pr}`3365`. See {ref}`packages-in-pyodide` for more details.
 
-- {{ Update }} Upgraded pandas to version 1.5.0.
-  {pr}`3134`
-
-- {{ Update }} Upgraded packages: numpy (1.23.5), {pr}`3284`
-
-- {{ Update }} Upgraded packages: scikit-learn (1.1.3), {pr}`3324`
+- {{ Fix }} Fix scipy handling of exceptions that are raised from C++ code.
+  {pr}`3384`.
 
 ## Version 0.21.3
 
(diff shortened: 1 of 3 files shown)


ryanking13 (approved):
Thanks, @hoomane!

This change increases the size of the scipy wheel by ~0.2MB, which, I think, is okay.

```
# Before
11429241 scipy-1.9.3-cp310-cp310-emscripten_3_1_27_wasm32.whl
# After
11651526 scipy-1.9.3-cp310-cp310-emscripten_3_1_27_wasm32.whl
```

rth:
Thanks, Hood!

PR #3385, opened 2022-12-23
Improve mypy typing for pyodide package

hoodmane:
The way we were handling the "routing" in `_core.py` (where in browser things are imported from `_pyodide_core` and otherwise things from `_pyodide._core_docs`) made mypy type a bunch of things as `Any`. This changes it to prevent mypy from understanding the imports from `_pyodide_core` so it will correctly type all of the methods with their types from `_pyodide._core_docs`.

This created a bunch of new type errors, so I also made various fixes to make things typecheck again.

### Checklists

### DIFF
--- a/src/py/_pyodide/_core_docs.py
+++ b/src/py/_pyodide/_core_docs.py
@@ -8,11 +8,12 @@ from collections.abc import (
     KeysView,
     Mapping,
     MutableMapping,
+    Sequence,
     ValuesView,
 )
 from functools import reduce
 from types import TracebackType
-from typing import IO, Any, Awaitable
+from typing import IO, Any, Awaitable, overload
 
 # All docstrings for public `core` APIs should be extracted from here. We use
 # the utilities in `docstring.py` and `docstring.c` to format them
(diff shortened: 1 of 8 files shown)


src/py/_pyodide/_core_docs.py
```diff
@@ -909,19 +954,70 @@ def create_proxy(
 # from python2js
 
 
+@overload
+def to_js(
+    obj: list[Any] | tuple[Any],
```
rth, line 959: It's unfortunate we have to repeat the rest of the annotations, just for this argument and output type, but I also don't see a way around it.

rth (approved):
Thanks LGTM.

At some point we need to see if there is a way to better format types in the documentation, as it's becoming not very readable
image

hoodmane:
Thanks for the review @rth!

hoodmane:
Agreed about the types in api docs. I'll try to look into improving it soon. For instance, it would be a huge improvement if instead of `collections.abc.Iterable` it could just say `Iterable`. Similarly, if we could render `Optional[T]` as something like `T?` and `Union[S,T]` as `S | T` this would be more readable. Also, if the types are really long, it would be good to start each argument on its own line.

hoodmane:
It's also oddly inconsistent: See [throw](https://pyodide.org/en/latest/usage/api/python-api/ffi.html#pyodide.ffi.JsGenerator.throw) where the first argument uses `A | B` but the second argument uses `Union[A, B]`.

PR #3387, opened 2022-12-26
Add type parameters to JsProxy subtypes

hoodmane:
Added type parameters to `JsArray`, `JsMap`, `JsMutableMap`, etc. Based on [the typesheds for typing](https://github.com/python/typeshed/blob/main/stdlib/typing.pyi). I skipped `JsCallable` and a few other more complicated types, those can be handled later.

### Checklists

### DIFF
--- a/src/py/_pyodide/_core_docs.py
+++ b/src/py/_pyodide/_core_docs.py
@@ -580,7 +580,7 @@ class JsMutableMap(JsMap[_KT, _VT], Generic[_KT, _VT]):
         """
         raise NotImplementedError
 
-    def setdefault(self, key: _KT, default: _VT) -> _VT:
+    def setdefault(self, key: _KT, default: _VT | None = None) -> _VT:
         """If key in self, return self[key]. Otherwise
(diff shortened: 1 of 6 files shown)


src/py/_pyodide/_core_docs.py
```diff
@@ -772,7 +818,7 @@ async def json(self) -> JsProxy:
         raise NotImplementedError
 
 
-class JsAsyncGenerator(JsIterable):
+class JsAsyncGenerator(JsAsyncIterable[_T_co], Generic[_T_co, _T_contra, _V_co]):
```
hoodmane, line 821: This should have been `JsAsyncIterable` all along...

ryanking13 (approved):
Thanks, @hoodmane!

image

The type vars are not very readable in docs, so it would be nice if we can find a way to improve it.

hoodmane:
I think I made it a lot better:
Screenshot from 2022-12-27 17-35-30

ryanking13:
Cool, great!

PR #3390, opened 2022-12-27
Fix broken pyodide cli tests

ryanking13:
Some CLI tests were misconfigured and were skipped in our CI.

Note that these tests are for now not runnable before building Pyodide.

### DIFF
--- a/pyodide-build/pyodide_build/cli/build.py
+++ b/pyodide-build/pyodide_build/cli/build.py
@@ -14,8 +14,6 @@ from .. import common
 from ..out_of_tree import build
 from ..out_of_tree.utils import initialize_pyodide_root
 
-app = typer.Typer()
-
 
 def _fetch_pypi_package(package_spec, destdir):
     PYMAJOR = common.get_make_flag("PYMAJOR")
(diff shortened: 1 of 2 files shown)


pyodide-build/pyodide_build/cli/build.py
```diff
@@ -14,8 +14,6 @@
 from ..out_of_tree import build
 from ..out_of_tree.utils import initialize_pyodide_root
 
-app = typer.Typer()
```
hoodmane, line 17: What is the significance of including/removing this? Doesn't this introduce the `build` subcommand?

ryanking13: It has no effect. The `pyodide build` entrypoint directly uses [`main`](https://github.com/pyodide/pyodide/blob/4114f20986a3caa923d9bd512a3795a8b9044a49/pyodide-build/setup.cfg#L45) function. I think @joemarshall added this `app` variable in order to use it in tests. I changed it in a way that [typer docs](https://typer.tiangolo.com/tutorial/testing/#test-a-function) recommend, so it is not needed now.

PR #3391, opened 2022-12-27
DOC Fix various sphinx warnings

ryanking13:
Fix various sphinx build warnings.

Note that I removed **all cross-references** in the changelog document. It causes trouble whenever we make an API change.
For example, Hood did some great work of splitting JsProxy classes into subclasses in this release, but the changelog of older versions still contains cross-references to `JsProxy.<method>` which doesn't exist anymore. It doesn't make sense to change it to e.g. `JsBuffer.<method>` or `JsArray.<method>` as those classes are not available in that version, so I think the reasonable option would be not using cross-references.


docs/conf.py
```diff
@@ -70,7 +70,7 @@
 autodoc_default_flags = ["members", "inherited-members"]
 
 intersphinx_mapping = {
-    "micropip": (f"https://micropip.pyodide.org/en/v{micropip.__version__}/", None)
+    "micropip": (f"https://micropip.pyodide.org/en/v{micropip.__version__}/", None)  # type: ignore[attr-defined]
```
hoodmane, line 73: This creates a mypy-tests error, presumably it fixes an error in plain mypy... Maybe we could add an exclude pattern to mypy-tests to skip this file?

src/py/_pyodide/_core_docs.py
```diff
@@ -527,9 +527,9 @@ class JsMutableMap(JsMap):
     number (idiomatically it should be called ``.size``) and it must be
     iterable.
 
-    Instances of the JavaScript builtin ``Map`` class are ``JsMutableMap``s.
+    Instances of the JavaScript builtin ``Map`` class are ``JsMutableMap`` s.
```
hoodmane, line 530: I don't think we want these two spaces.

hoodmane (approved):

docs/conf.py
ryanking13: I guess my local micropip version was a problem.

src/py/_pyodide/_core_docs.py
ryanking13: It wasn't displayed correctly:

[before]

image

[after]

image

hoodmane:
Thanks @ryanking13!

PR #3398, opened 2022-12-30
Fix 'version' in api docs

hoodmane:
The documentation generation seems to get confused by the default export and
displays the `version` field as `default` instead.  This switches to a normal export.


ryanking13:
Thanks!

PR #3399, opened 2022-12-30
DOCS Fix the async prefix, improvement to summary table format

hoodmane:
The `async` prefix was supposed to appear before an async function both in the summary
table and in the documentation entry on the function, but it was only appearing in the
summary table. Looking at the sphinx source code, I'm not sure why what I was doing
worked at all actually. This fixes the problem.

This also makes summary table entries bold, which makes them slightly more consistent
with the documentation entries.


ryanking13 (approved):
Very nice. Thanks!

PR #3404, opened 2023-01-03
DOC Rearrange out-of-tree build doc

ryanking13:
The out-of-tree build doc was splitted into two documents, this PR merges it to one.


docs/development/building-and-testing-packages.md
```diff
@@ -1,27 +1,65 @@
...
+Subsystem for Linux.
+
 ## Building binary packages for Pyodide
 
 If your package is a pure Python package (i.e., if the wheel ends in
 `py3-none-any.whl`) then follow the official PyPA documentation on building
 [wheels](https://packaging.python.org/en/latest/tutorials/packaging-projects/#generating-distribution-archives)
-Otherwise, the procedure is simple. In your package directory run the following
-command line commands:
+Otherwise, the procedure is simple.
```
hoodmane, line 18: Maybe we shouldn't say it is "simple" since in some circumstances it will not be simple (and even when it is, not everyone will find it to be),

hoodmane (approved):

docs/development/building-and-testing-packages.md
ryanking13: Right, we should not judge readers. Thanks!

PR #3405, opened 2023-01-03
DOCS Use autodoc typehints and make various improvements to autodoc rendering

hoodmane:
Use `autodoc-typehints` to generate the types for the Python API docs. This is a major
improvement to the rendering of a lot of things, since the signature isn't so cluttered 
with types.

I also fixed the rendering of return value annotations which used to be messed up in
a few places. I added `async` prefixes to the entries in the auto summary table and
made the font upright and bold for the signatures in the autosummary table, just like
they are in the actual documentation. I also added cpython core to intersphinx.

Finally, I went through a bunch of the API docs and fixed them to add xrefs to the
cpython docs and to make them work correctly with autodoc type hints.

<details>
<summary>Before 1</summary>

Screenshot from 2023-01-02 21-14-24

</details>

<details>
<summary>After 1</summary>

Screenshot from 2023-01-02 21-13-53

</details>

<details>
<summary>Before 2</summary>

Screenshot from 2023-01-02 21-20-49

</details>

<details>
<summary>After 2</summary>

Screenshot from 2023-01-02 21-20-36

</details>

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -46,6 +46,7 @@ extensions = [
     "sphinx_argparse_cli",
     "versionwarning.extension",
     "sphinx_issues",
+    "sphinx_autodoc_typehints",
 ]
 
 
@@ -70,7 +71,8 @@ autosummary_generate = True
 autodoc_default_flags = ["members", "inherited-members"]
 
 intersphinx_mapping = {
-    "micropip": (f"https://micropip.pyodide.org/en/v{micropip.__version__}/", None)
+    "python": ("https://docs.python.org/3.10", None),
+    "micropip": (f"https://micropip.pyodide.org/en/v{micropip.__version__}/", None),
 }
 
 # Add modules to be mocked.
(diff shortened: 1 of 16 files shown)


docs/conf.py
```diff
@@ -258,6 +260,47 @@ def globalReplace(app, docname, source):
 
 global_replacements = {"{{PYODIDE_CDN_URL}}": CDN_URL}
 
+# Add extra CSS classes to some documentation fields. We use this in pyodide.css
+# to fix the rendering of autodoc return value annotations.
+from sphinx.util.docfields import Field
```
rth, line 265: Should we open an  issue about it at https://github.com/executablebooks/sphinx-book-theme ? It should ideally work with autodoc-typehints  without hacks.
Unless it's a problem with numpydoc that we are using

rth (commented):
Thanks for doing this. Will try to review more carefully later today, but overall using autodoc-typehints sound great! It sounds like it does exactly what I was hoping in https://github.com/numpy/numpydoc/issues/196

docs/conf.py
hoodmane: I think this is more Napoleon's fault or at least it is fixing a problem that can be reproduced without autodoc-typehints.

src/py/_pyodide/_base.py
```diff
@@ -316,7 +324,7 @@ async def run_async(
         self,
         globals: dict[str, Any] | None = None,
         locals: dict[str, Any] | None = None,
-    ) -> None:
+    ) -> Any:
```
hoodmane, line 327: This was an actual bug in our type annotations.

docs/conf.py
hoodmane: The issue specifically is that Napoleon (I think?) messes up the rendering if you don't provide an explicit return value type, either because you just don't want to, or because the return type annotation is filled in by some tool. It may also not be Napoleon's fault, there are so many different tools here that all only sort of work together that it is hard to tell where to place the blame when things don't work.

src/js/api.ts
```diff
@@ -15,7 +15,7 @@ import { setStdin, setStdout, setStderr } from "./streams";
 API.loadBinaryFile = loadBinaryFile;
 
 /**
- * An alias to the Python :py:mod:`pyodide` package.
+ * An alias to the Python `pyodide </usage/api/python-api.html>`_ package.
```
rth, line 18: Are you sure this works? It would be better to link to  a reference here. I think it wouldn't work when deployed under the /en/stable prefix in RTD.

Otherwise LGTM, thanks!

I'll fix this and merge, so we can use it in other PRs.

rth (approved):

hoodmane:
Thanks @rth!

hoodmane:
We should backport this to stable.

hoodmane:
We should backport this to stable.

PR #3411, opened 2023-01-04
Add a CLI command to create a zipfile of Python libraries

ryanking13:
### Description

This PR is a split-off from #3166.

This adds `pyodide create-zip` (open to other names) CLI command into pyodide-build,
which bundles a directory into a zipfile, so it can be imported.

Note that this is not part of our build system yet, I'll make a follow up PR after #3252 and #3253 are merged.

### Checklists

- [x] Add / update tests


pyodide-build/pyodide_build/pyzip.py
```diff
@@ -0,0 +1,128 @@
...
+#       So we should not pack them into the zip file in order to make e.g. import ctypes.test work.
+#       Note that all these tests are moved to the subdirectory of `test` module in upstream CPython 3.12.0a1.
+#       So we don't need this after we upgrade to 3.12.0
+NOT_ZIPPED_FILES = ("ctypes/", "unittest/")
+
+
+def default_filterfunc(
+    root: Path, verbose: bool = False
+) -> Callable[[str, list[str]], set[str]]:
+    def filterfunc(path: Path | str, names: list[str]) -> set[str]:
```
rth, line 44: Could you add a docstring describing what exactly we are filtering ?

pyodide-build/pyodide_build/pyzip.py
```diff
@@ -0,0 +1,128 @@
...
+    libdir: Path | str,
+    output: Path | str = "python",
+    pycompile: bool = False,
+    filterfunc: Callable[[str, list[str]], set[str]] | None = None,
+) -> None:
+    """
+    Bundle Python standard libraries into a zip file.
+
+    The basic idea of this function is similar to the standard library's
+    [PyZipFile](https://github.com/python/cpython/blob/59665d0280c2299ea87e9af45cedc90656cb6f55/Lib/zipfile/__init__.py#L2011).
```
rth, line 81:
```
{ref}`zipfile.PyZipFile`
``` should work I think once we enable stdlib sphinx inter-linking in https://github.com/pyodide/pyodide/pull/3405/files#diff-85933aa74a2d66c3e4dcdf7a9ad8397f5a7971080d34ef1108296a7c6b69e7e3R74

File: pyodide-build/pyodide_build/pyzip.py [hunk 3]
```diff
@@ -0,0 +1,128 @@
...
+    output = output.with_name(output.name.rstrip(".zip"))
+
+    if filterfunc is None:
+        filterfunc = default_filterfunc(libdir)
+
+    with TemporaryDirectory() as temp_dir_str:
+        temp_dir = Path(temp_dir_str)
+        shutil.copytree(libdir, temp_dir, ignore=filterfunc, dirs_exist_ok=True)
+
+        shutil.make_archive(str(output), "zip", temp_dir)
```
rth, line 128: Does it use compression by default?

pyodide-build/pyodide_build/tests/test_pyzip.py
```diff
@@ -0,0 +1,45 @@
...
+def test_import_from_zip(temp_python_lib, tmp_path):
+    output = tmp_path / "python.zip"
+
+    create_zip(temp_python_lib, output, pycompile=False, filterfunc=None)
+
+    assert output.exists()
+
+    import sys
+
+    sys.path.insert(0, str(output))
```
rth, line 40: Maybe monkeypatch it so we don't change sys.path globally

pyodide-build/pyodide_build/cli/create_zip.py
```diff
@@ -0,0 +1,23 @@
+from pathlib import Path
+
+import typer  # type: ignore[import]
```
rth, line 3: It's weird that mypy complains. Maybe it should be added to https://github.com/pyodide/pyodide/blob/main/pyproject.toml#L53?

rth (commented):
Thanks @ryanking13 ! A few minor comments otherwise LGTM.

For the name maybe `pyodide create-zipfile` since that's the name of the Python stdlib module? Do we generally expect users to use this? If not maybe we shouldn't make it part of the public CLI API?

Also needs a changelog.

pyodide-build/pyodide_build/pyzip.py
```diff
@@ -0,0 +1,128 @@
...
+    - We need an option to "not" compile the files in the zip file
+
+    hence this function.
+
+    Parameters
+    ----------
+    libdir : Path or str
+        Path to the directory containing the Python standard library.
+
+    output : Path or str, optional
```
rth, line 95: Also now that https://github.com/pyodide/pyodide/pull/3405 is merged, you should be able to remove the types in docstrings and have them auto-generated from type annotations.

ryanking13:
Thanks for the through review @rth!

> > For the name maybe `pyodide create-zipfile` since that's the name of the Python stdlib module? Do we generally expect users to use this? If not maybe we shouldn't make it part of the public CLI API?

I think users will not use this directly, but I am thinking of using this CLI in our makefile. Probably we can hide it from the help message and docs?

pyodide-build/pyodide_build/pyzip.py
ryanking13: Yeah it does, though I think it is not a very well documented behavior. (https://github.com/python/cpython/blob/31b639a99281bd9582e88137a3be2868c4154a88/Lib/shutil.py#L978)

pyodide-build/pyodide_build/cli/create_zip.py
ryanking13: Added typer to [[tools.mypy.overrides]].

pyodide-build/pyodide_build/tests/test_cli.py
```diff
@@ -127,3 +132,52 @@ def test_config_get(cfg_name, env_var):
     )
 
     assert result.stdout.strip() == common.get_make_flag(env_var)
+
+
+def test_fetch_or_build_pypi(selenium, tmp_path):
```
rth, line 137: There is already a test named the same way on main. I wonder if it's badly merging with `main` in CI and we end up with 2 identically named tests one of which would be ignored. This would explain why coverage says it's not run in CI.

Could you merge the main in, please to see if it would help?

rth (approved):
Thanks @ryanking13 . Minor comment otherwise LGTM.

pyodide-build/pyodide_build/tests/test_cli.py
rth: Merged main to resolve conflict. Otherwise, nevermind my comment, I did see the function test on main locally, but not after syncing with upstream. Still unclear why codecov doesn't see coverage here, but merging anyway.

ryanking13: Yeah, this test was moved to [test_pypi.py](https://github.com/pyodide/pyodide/blob/87ea5862c4a154bcd0277eae40cb566c3f0edbc3/pyodide-build/pyodide_build/tests/test_pypi.py#L212) in #3310. I think I made a mistake while merging main.

PR #3412, opened 2023-01-04
Fix pyodide-py package name

ryanking13:
I think this will fix the release CI failure (https://github.com/pyodide/pyodide/actions/runs/3829966769/jobs/6517241654).

Someone would need to make a manual release after merge.


rth:
Thanks! I uploaded the wheel manually

PR #3417, opened 2023-01-04
fix: pass only build args

### LINKED ISSUE #3415: `pyodide build` passes build location as backend argument
I ran into this in our CI: the source directory is being passed as an arg, which ultimately is interpreted as a `config-setting` instead of setting the cwd. This breaks our PEP517 build backend that doesn't expect this argument. I think we just want `os.chdir` here?

https://github.com/pyodide/pyodide/blob/e90154f3e22354a6104ec9c8a3e2ca1baaa2b085/pyodide-build/pyodide_build/cli/build.py#L140

agoose77:
### Checklists

- [ ] Add / update tests

Fixes #3415


ryanking13 (approved):
Thanks @agoose77! Could you please update the changelog?

Issue #3420, opened 2023-01-04
Deno support


rth:
Deno support was discussed in https://github.com/pyodide/pyodide/issues/1477#issuecomment-821942072 and I think it would be good to add it and have some minimal tests in CI. As discussed in the linked PR only a few minor fixes are needed. I'll open a PR shortly.

Personally, the use case I'm interested in that could benefit from running in Deno is server side sandboxing https://github.com/pyodide/pyodide/issues/869#issuecomment-1268006881

What's left is:

 - [ ] Support running in Deno
 - [ ] Add  support for deno runtime to `pytest-pyodide`
 - [ ] Add some tests with Deno in CI
 - [ ] Update documentation

rajsite:
@rth know if this issue is still moving forward? I'm really interested in the Deno server side sandboxing use case as well.

rth:
I haven't found much time to work on it lately, but it's still something I want to do.

So if you are interested in hacking on it and continue https://github.com/pyodide/pyodide/pull/3421 or https://github.com/pyodide/pyodide/pull/3428 that would certainly be appreciated :)

rajsite:
I experimented with a different direction and it looks really promising! Deno has been adding [npm package support](https://deno.com/manual@v1.32.5/node/npm_specifiers) so I tried loading pyodide in Deno as a CommonJS module.

I had to make a patch to `pyodide.js` and `pyodide.asm.js` as the `IN_NODE` heuristic was failing in Deno's npm mode:

<img width="245" alt="image" src="https://user-images.githubusercontent.com/1588923/234761274-26b51391-f9dc-4140-9fb2-404452d864dd.png">

Looks like the `process.relesase.name` property is not defined on the `process` object in Deno's npm compatibility mode and I filed an issue to track it: https://github.com/denoland/deno/issues/18870

But with that patch I can run pyodide in Deno 🎉

One can reproduce it with the current `pyodide@0.23.1` release by doing the following:

1. Create a `main.ts` with the contents:
   
    ```ts
    // Note directly importing to the path of the CommonJS module
    // as importing the bare `npm:pyodide` resolves to the esm module
    // which seems to have a lot more changes needed for Deno support
    import {loadPyodide} from "npm:pyodide/pyodide.js";
    
    const pyodide = await loadPyodide();
    await pyodide.loadPackage('micropip');
    const micropip = pyodide.pyimport("micropip");
    await micropip.install('snowballstemmer');
[...]

PR #3421, opened 2023-01-04
Draft: Improve Deno compatibility

rth:
Toward https://github.com/pyodide/pyodide/issues/3420

This adds a couple of patches for Deno suggested by @josephrocca in https://github.com/pyodide/pyodide/issues/1477#issuecomment-821942072. Other patches discussed there are no longer needed, which is good news.

A minimal example adapted from the above discussion,
1. Install [deno](https://deno.land/manual@v1.29.1/getting_started/installation)
2. `cd pyodide/dist/`
3. Create `main.js`,
    ```py
    import "http://127.0.0.1:3001/pyodide.js";

      import micropip

      await micropip.install('snowballstemmer')

      import snowballstemmer
      stemmer = snowballstemmer.stemmer('english')
      assert stemmer.stemWords("Testing".split()) == ['Test']
    `);
    ```
4. Serve files,
    ```bash
    python -m http.server 3001
    ```
5. Run Deno
    ```
    deno run --reload --allow-net --location=http://127.0.0.1:3001 main-save.js
    ```
    The `--reload` option is only needed for development, otherwise it caches downloaded files.
    
TODO: 
- [ ] Fix CI 
- [ ]  Add some tests with Deno in CI
- [ ] Update documentation


src/js/compat.ts
```diff
@@ -72,6 +78,24 @@ export async function initNodeModules() {
...
+ */
+export async function initDenoModules() {
+  if (!IN_DENO) {
+    return;
+  }
+  // Workaround for https://github.com/emscripten-core/emscripten/issues/6241
+
+  let XHRModule = await import(
+    // @ts-ignore
+    /* webpackIgnore: true */ "https://deno.land/x/xmlhttprequest_deno@v0.0.2/mod.js"
```
rth, line 94: This should never be used with webpack, since deno has its own bundler from what I understand.
That package was uploaded by josephrocca, see https://deno.land/x/xmlhttprequest_deno@v0.0.2

Of course, the ideal solution would be to fix it in https://github.com/emscripten-core/emscripten/issues/6241 but that would require quite a bit of refactoring. So this is temporary workaround.

rth:
So this uses `eval` to run  `pyodide.asm.js`, is simple but only works when Pyodide and other files are served via an HTTP server.

I have also looked into re-using more of the Node code in https://github.com/pyodide/pyodide/compare/main...rth:pyodide:deno-support-v2?expand=1, which should hopefully allow to load files without an HTTP server. It also uses `vm.runInThisContext` instead of `eval`.  However, that needs pre-installing all modules that are loaded via `require`  in emscripten generated `pyodide.asm.js` (same as we are doing in Node). Overall the code should be very similar to Node but with some minor differences. For now, I still haven't managed to solve all the errors. 

So I think the plan is to merge this simple version (once the TODO items above are addressed) and mark Deno support as very experimental. Then look into better integration with that other PR.

 Similarly, streams setup `src/js/streams.ts` for Deno would be better done in a separate PR, as it's slightly different from Node (e.g. `Deno.stdout` instead of `process.stdout`), though maybe the letter could [still be reused](https://deno.land/std@0.110.0/node/process.ts?s=stdout). And also `streams.ts` has a bunch of imports `require` which need to be adapted for Deno.

josephrocca:
@rth Thanks for moving this forward! I've lost a lot of context on all of this so I'm probably not going to be all that helpful here, but:

> For now, I still haven't managed to solve all the errors.

What are these errors, specifically?

> this uses eval to run pyodide.asm.js, is simple but only works when Pyodide and other files are served via an HTTP server.

If the `Deno` global is available, then it could just use `await Deno.readTextFile(url)` instead of `await fetch(url).then(r => r.text())`, right? Or am I misunderstanding (or would that not be any better)? If `Deno.readTextFile` would be more appropriate, then it might make sense to abstract it all out into a "readTextFile" function that handles the runtime differences internally?

rth:
For that other PR https://github.com/pyodide/pyodide/pull/3428  the latest error is,
<details>

```
Loading micropip, pyparsing, packaging
Failed to load micropip, pyparsing, packaging
Cannot read properties of undefined (reading 'resolve')
The following error occurred while loading pyparsing:
Cannot read properties of undefined (reading 'resolve')
The following error occurred while loading packaging:
Cannot read properties of undefined (reading 'resolve')
error: Uncaught PythonError: Traceback (most recent call last):
    raise self._exception
    result = coro.send(None)
  File "/lib/python3.10/_pyodide/_base.py", line 541, in eval_code_async
    await CodeRunner(
  File "/lib/python3.10/_pyodide/_base.py", line 363, in run_async
    await coroutine
  File "<exec>", line 2, in <module>
ModuleNotFoundError: The module 'micropip' is included in the Pyodide distribution, but it is not installed.
  await micropip.install("micropip") in Python, or
  await pyodide.loadPackage("micropip") in JavaScript
See https://pyodide.org/en/stable/usage/loading-packages.html for more details.

    at new_error (eval at runInThisContext (https://deno.land/std@0.170.0/node/vm.ts:10:21), <anonymous>:10:180676)
    at <anonymous> (http://127.0.0.1:3001/pyodide.asm.wasm:1:948539)
    at <anonymous> (http://127.0.0.1:3001/pyodide.asm.wasm:1:977271)
    at method_call_trampoline (eval at runInThisContext (https://deno.land/std@0.170.0/node/vm.ts:10:21), <anonymous>:10:233580)
```
</details>
 so I probably forgot to replace a `require` with an `import` somewhere, need to look into it. Also paths seemed to be resolved a bit differently to Node (and so code that worked for Node no longer worked for some reason). Anyway, nothing major.


> If the Deno global is available, then it could just use await Deno.readTextFile(url) instead of await fetch(url).then(r => r.text()), right? 

yes absolutely. In node, we are currently directly [importing the path for the local JS file](https://github.com/pyodide/pyodide/pull/3428/files#diff-0503c45203a13549426fcea74188746dd314feb61fe75c458485132d3de229b8R286), so that should also work if we want to reuse existing code.

Issue #2816, opened 2022-06-30
Add package `tokenizers`


josephrocca:
There has recently been some exciting progress on bringing Hugging Face's Tokenizers lib to wasm/browsers:

 * @Narsil: https://github.com/huggingface/tokenizers/pull/1009
 * @mbrunel: https://github.com/mithril-security/tokenizers-wasm
 * Discussion: 
   * https://github.com/huggingface/tokenizers/issues/935
   * https://github.com/huggingface/tokenizers/issues/63
   * https://github.com/huggingface/tokenizers/issues/1010

I think it'd be great if the full Python bindings were available in the browser (including JupyterLite) via Pyodide, and it doesn't seem like too much extra work.

@messense has been super helpful over on [this issue](https://github.com/huggingface/tokenizers/issues/1010) and the instructions below now produce a `.whl` that seems to *almost* work in the browser (problem is explained below).

### Build instructions:
Visit [this branch](https://github.com/josephrocca/tokenizers/tree/pyodide) and start a Github Codespace on it, then run these commands:
```bash
curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh -s -- -y
source $HOME/.cargo/env
rustup toolchain install nightly
rustup target add --toolchain nightly wasm32-unknown-emscripten
rustup component add rust-src --toolchain nightly-x86_64-unknown-linux-gnu
cargo install --git https://github.com/PyO3/maturin.git maturin # tested and working at commit hash 05169de
git clone https://github.com/emscripten-core/emsdk.git # tested and working at commit hash 517e02f
cd emsdk
./emsdk install latest
./emsdk activate latest
source ./emsdk_env.sh
cd ../bindings/python
RUSTUP_TOOLCHAIN=nightly maturin build --release -o dist --target wasm32-unknown-emscripten -i python3.10
```

### The problem
This produces a `.whl` file that seems to work, except for a threading issue as you can see [in this demo](https://josephrocca.github.io/tokenizers-pyodide/demo/) ([source repo](https://github.com/josephrocca/tokenizers-pyodide)). @ryanking13 mentioned in [this comment](https://github.com/pyodide/pyodide/discussions/2720#discussioncomment-3008755) that Pyodide has some problems with threads that come up at runtime, so perhaps that's what's happening here?

For quick reference, here's the error that you'll see in the browser console in the above-linked demo:

```
[...]

PR #3422, opened 2023-01-05
PKG Add tokenizers

ryanking13:
### Description

Close: #2816 

This adds huggingface/tokenizers. @josephrocca already did some excellent work in [his repository](https://github.com/josephrocca/tokenizers/tree/pyodide) so the porting was quite easy.

Note that `tokenizer.train("<file_path>")` fails with error:

```
PyThreadState_Get: the function must be called with the GIL held, but the GIL is released (the current Python thread state is NULL)
```

while `tokenizer.train_from_iterator(<iterator>)` works. I think something goes wrong while reading a file (releasing and holding GIL), but it is not easy to figure out what is happening (I would appreciate it if @josephrocca can help).

### Checklists

- [x] Add / update tests


hoodmane:
> the function must be called with the GIL held, but the GIL is released 

We could we just delete the gil? We don't need it since there's no threading...

ryanking13:
> We could just delete the gil? We don't need it since there's no threading...

Do you mean in CPython side? or in tokenizer (pyo3) side?

hoodmane:
I was thinking of making gil acquisition and release into noops by patching cpython. But there are options.

ryanking13:
> I was thinking of making gil acquisition and release into noops by patching cpython. But there are options.

That sounds very interesting. I think it is worth to try, but as long as we eventually want to support threads, it might not be a long term solution.

josephrocca:
Thanks for moving this forward, Gyeongjae! I probably won't be much help here, because I didn't get any further than [this demo](https://josephrocca.github.io/tokenizers-pyodide/demo/) - I didn't actually test training at all, so I'm pleasantly surprised that that works. One thought: IIRC, I added the patch below as a way to avoid having to run `os.environ["TOKENIZERS_PARALLELISM"] = "0"` at runtime, but maybe this patch doesn't properly do that (anymore?) - i.e. maybe the `TOKENIZERS_PARALLELISM` variable is used in other places that I missed. So it might be worth setting that environment variable before calling `tokenizer.train("<file_path>")` just to see if that fixes it. If it does work, then maybe it'd make sense to just add a patch that sets `TOKENIZERS_PARALLELISM` during init. (But to be clear this is just a random guess without having looked into this too deeply)

```diff
@@ -28,7 +28,7 @@ pub fn get_parallelism() -> bool {
             v.make_ascii_lowercase();
             !matches!(v.as_ref(), "" | "off" | "false" | "f" | "no" | "n" | "0")
         }
-        Err(_) => true, // If we couldn't get the variable, we use the default
+        Err(_) => false, // If we couldn't get the variable, we use the default
     }
 }
```

josephrocca:
Was just trying to build this an in-tree builds don't see to work anymore? Tried in the docker image (`./run_docker`) and in Codespaces - same error `/src/.docker_home/.cargo/env: No such file or directory`:

```
patching file src/utils/regex.rs
Using Plan A...
Hunk #1 succeeded at 1.
Hmm...  The next patch looks like a unified diff to me...
--------------------------
...
|diff --git a/tokenizers-lib/src/utils/parallelism.rs b/tokenizers-lib/src/utils/parallelism.rs
|index f2de5ae..720c0cf 100644
|--- a/tokenizers-lib/src/utils/parallelism.rs
|+++ b/tokenizers-lib/src/utils/parallelism.rs
--------------------------
patching file tokenizers-lib/src/utils/parallelism.rs
Using Plan A...
Hunk #1 succeeded at 28.
Hmm...  Ignoring the trailing garbage.
done
bash: /src/.docker_home/.cargo/env: No such file or directory
[2023-01-07 13:15:37] Building package tokenizers...
ERROR: bash command failed
    source $CARGO_HOME/env
    rm -f rust-toolchain  # rust-toolchain file overrides toolchain setting

[2023-01-07 13:15:37] Succeeded building package tokenizers in 0.2 seconds.
ERROR: cancelling buildall
```

ryanking13:
> /src/.docker_home/.cargo/env: No such file or directory

For now you will have to [run `make rust` first](https://pyodide.org/en/stable/development/building-from-sources.html?highlight=make%20rust#partial-builds) to install rust toolchain in a specific directory, though it might be changed when #3441 is merged.

Issue #3427, opened 2023-01-05
Add package: onnx


josephrocca:
## 🐍 Package Request

- Package Name and Version : onnx / v1.13.0 (latest)
- Package URL : https://github.com/onnx/onnx/tree/v1.13.0
- Package Dependencies that needs to be resolved first: [protobuf](https://github.com/pyodide/pyodide/issues/3424)

# Build instructions
```bash
git clone --single-branch v21.12 https://github.com/onnx/onnx
cd onnx
git submodule update --init --recursive

# pyodide setup:
pip install pyodide-build

# install emscripten
git clone https://github.com/emscripten-core/emsdk
cd emsdk
PYODIDE_EMSCRIPTEN_VERSION=$(pyodide config get emscripten_version)
./emsdk install ${PYODIDE_EMSCRIPTEN_VERSION}
./emsdk activate ${PYODIDE_EMSCRIPTEN_VERSION}
source emsdk_env.sh
cd ../

# install protobuf (the whole lib, not just the `protoc` binary)
git clone https://github.com/protocolbuffers/protobuf.git
cd protobuf
git checkout v3.20.2
git submodule update --init --recursive
mkdir build_source && cd build_source
cmake ../cmake -Dprotobuf_BUILD_SHARED_LIBS=OFF -DCMAKE_INSTALL_PREFIX=/usr -DCMAKE_INSTALL_SYSCONFDIR=/etc -DCMAKE_POSITION_INDEPENDENT_CODE=ON -Dprotobuf_BUILD_TESTS=OFF -DCMAKE_BUILD_TYPE=Release
make -j$(nproc)
sudo make install
cd ../../

pip install "pybind11[global]"
export PATH="$(pwd)/protobuf/build_source:$PATH"
export CMAKE_ARGS="-DONNX_USE_LITE_PROTO=ON -DProtobuf_INCLUDE_DIR=$(pwd)/protobuf/src -DProtobuf_LIBRARIES=$(pwd)/protobuf/build_source -DProtobuf_LITE_LIBRARY=$(pwd)/protobuf/build_source/libprotobuf-lite.a -Dpybind11_DIR=$(python -c 'import pybind11 as _; print(_.__path__[0])')/share/cmake/pybind11"
pyodide build
```

# Progress so far

The above build instruction seem to work fine, but the next issue that I'm working through is this runtime error:

```
[...]

PR #3434, opened 2023-01-07
Ignore --exclude-libs link flag

josephrocca:
### Description

Following @ryanking13's instructions here: https://github.com/pyodide/pyodide/issues/3427#issuecomment-1374422693

I'm assuming that `--exclude-libs` will always have arguments following it, so we should be testing `startswith("--exclude-libs=")` rather than the existence of `--exclude-libs`.


ryanking13 (approved):
Thanks, @josephrocca!

PR #3438, opened 2023-01-08
Handle response files in `calculate_exports`

### LINKED ISSUE #3436: Improve `pywasmcross` to handle response files
@ryanking13 mentioned [here](https://github.com/pyodide/pyodide/issues/3427#issuecomment-1374418990) that:

> pywasmcross calculates what symbols need to be exported by looking at objects files (.a, .o). But it doesn't handle response files (.rsp). So we need to add a logic to parse response files there.

I actually didn't even know what a response file was, so I just asked ChatGPT to do it for me, and here's what it says:

---

A response file is a file that contains a list of arguments that can be passed to a command line tool. For example, suppose we have a command line tool called "tool" that accepts a list of input files and processes them. Instead of typing out the long list of input files every time we want to run the tool, we can put the list of input files in a response file and pass the response file to the tool using the "@" symbol. For example:
```
tool @response_file.rsp
```
The contents of response_file.rsp might look something like this:
```
input_file1.txt
input_file2.txt
input_file3.txt
```
To handle response files in the `calculate_exports` function, we can modify the function to parse the response file and extract the list of arguments contained within it. Here is an example of how this could be done:

```py
def calculate_exports(line: list[str], export_all: bool) -> Iterable[str]:
    """
    Collect up all the object files and archive files being linked and list out
    symbols in them that are marked as public. If ``export_all`` is ``True``,
    then return all public symbols. If not, return only the public symbols that
    begin with `PyInit`.
    """
    # Parse response files and extract list of arguments
    objects = []
    for arg in line:
        if arg.endswith(".rsp"):
            with open(arg, "r") as f:
                objects.extend(f.read().splitlines())
        else:
            objects.append(arg)
    
    # Filter objects to only include .a and .o files
    objects = [obj for obj in objects if obj.endswith((".a", ".o"))]
    
    exports = None
    # Using emnm is simpler but it cannot handle bitcode. If we're only
    # exporting the PyInit symbols, save effort by using nm.
    if export_all:
        exports = calculate_object_exports_readobj(objects)
    if exports is None:
        # Either export_all is false or we are linking at least one bitcode
        # object. Fall back to a more conservative estimate of the symbols
        # exported. This can export things with `__visibility__("hidden")`
        exports = calculate_object_exports_nm(objects)
    if export_all:
        return exports
    return (x for x in exports if x.startswith("PyInit"))
```

---

Seems simple enough? It just replaced [this line](https://github.com/pyodide/pyodide/blob/aeb066bbcef310a419bf058347bbc4d23ca52ec1/pyodide-build/pyodide_build/pywasmcross.py#L501):

```py
   objects = [arg for arg in line if arg.endswith((".a", ".o"))]
```

with this:

```py
    # Parse response files and extract list of arguments
    objects = []
    for arg in line:
        if arg.endswith(".rsp"):
            with open(arg, "r") as f:
                objects.extend(f.read().splitlines())
        else:
            objects.append(arg)
    
    # Filter objects to only include .a and .o files
    objects = [obj for obj in objects if obj.endswith((".a", ".o"))]
```

josephrocca:
Closes #3436

### Description

This allows pywasmcross to handle response files in its `calculate_exports` parsing.

See here for info: https://github.com/pyodide/pyodide/issues/3427#issuecomment-1374418990

### Checklists

- [ ] Add / update tests

## Additional notes

I haven't written any tests for this. I've only tested it in a notebook:

image

### DIFF
--- a/pyodide-build/pyodide_build/pywasmcross.py
+++ b/pyodide-build/pyodide_build/pywasmcross.py
@@ -505,7 +505,7 @@ def calculate_exports(line: list[str], export_all: bool) -> Iterable[str]:
             all_args.extend(Path(arg).read_text().splitlines())
         else:
             all_args.append(arg)
-    
+
     objects = [arg for arg in all_args if arg.endswith((".a", ".o"))]
 
     exports = None
--- a/pyodide-build/pyodide_build/pywasmcross.py
+++ b/pyodide-build/pyodide_build/pywasmcross.py
@@ -498,15 +498,14 @@ def calculate_exports(line: list[str], export_all: bool) -> Iterable[str]:
     then return all public symbols. If not, return only the public symbols that
     begin with `PyInit`.
     """
-    # Extract list of arguments, including those from response files
+    # Get all arguments, including those from response files
     all_args = []
     for arg in line:
         if arg.endswith(".rsp"):
             all_args.extend(Path(arg).read_text().splitlines())
         else:
             all_args.append(arg)
-
-    # Filter objects to only include .a and .o files
+    
     objects = [arg for arg in all_args if arg.endswith((".a", ".o"))]
 
     exports = None
--- a/pyodide-build/pyodide_build/pywasmcross.py
+++ b/pyodide-build/pyodide_build/pywasmcross.py
@@ -498,7 +498,18 @@ def calculate_exports(line: list[str], export_all: bool) -> Iterable[str]:
     then return all public symbols. If not, return only the public symbols that
     begin with `PyInit`.
     """
-    objects = [arg for arg in line if arg.endswith((".a", ".o"))]
+    # Parse response files and extract list of arguments
+    objects = []
+    for arg in line:
+        if arg.endswith(".rsp"):
+            with open(arg, "r") as f:
+                objects.extend(f.read().splitlines())
+        else:
+            objects.append(arg)
+    
+    # Filter objects to only include .a and .o files
+    objects = [obj for obj in objects if obj.endswith((".a", ".o"))]
+    
     exports = None
     # Using emnm is simpler but it cannot handle bitcode. If we're only
     # exporting the PyInit symbols, save effort by using nm.
--- a/pyodide-build/pyodide_build/pywasmcross.py
+++ b/pyodide-build/pyodide_build/pywasmcross.py
@@ -498,16 +498,16 @@ def calculate_exports(line: list[str], export_all: bool) -> Iterable[str]:
     then return all public symbols. If not, return only the public symbols that
     begin with `PyInit`.
     """
-    # Parse response files and extract list of arguments
-    objects = []
+    # Extract list of arguments, including those from response files
+    all_args = []
     for arg in line:
         if arg.endswith(".rsp"):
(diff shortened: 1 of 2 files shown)


pyodide-build/pyodide_build/pywasmcross.py
```diff
@@ -498,7 +498,18 @@ def calculate_exports(line: list[str], export_all: bool) -> Iterable[str]:
     then return all public symbols. If not, return only the public symbols that
     begin with `PyInit`.
     """
-    objects = [arg for arg in line if arg.endswith((".a", ".o"))]
+    # Parse response files and extract list of arguments
+    objects = []
+    for arg in line:
+        if arg.endswith(".rsp"):
+            with open(arg) as f:
```
hoodmane, line 505: How about:
```py
objects.extend(Path(arg).read_text().split lines())
```

hoodmane (approved):
Thanks @josephrocca, makes sense to me. 

One question: if the arguments are really long, would we potentially need to use a response file? Or is it okay to unpack the response file into the arguments in all cases? I think there are limits to how long an argument list for a process is allowed to be? I guess if we run into trouble we could generate a new response file by filtering the old one...

josephrocca:
I'm semi-blindly following [this comment](https://github.com/pyodide/pyodide/issues/3427#issuecomment-1374418990), but I think I understand what you mean - these arguments are used as part of an execution [here](https://github.com/pyodide/pyodide/blob/b0b07efe94c75d5cd268e3dba659095dfda3b39c/pyodide-build/pyodide_build/pywasmcross.py#L705), and if there are too many, this could cause an error? I'm a little confused though because my understanding from @ryanking13's comment was that this change was to actually fix the "Argument list too long" error that I was getting [here](https://github.com/pyodide/pyodide/issues/3427#issuecomment-1374415900) :sweat_smile:

hoodmane:
Yeah this won't fix that error. You'd fix that error by checking how many arguments there are and writing the arguments to a response file if there are too many. But I think this change still fixes an edge case. 

We could use a test case with a very large number of arguments for this.

ryanking13:
Actually, it seems like the two tools that we use to calculate exports ([llvm-nm](https://llvm.org/docs/CommandGuide/llvm-nm.html#cmdoption-llvm-nm-arg-file) and [llvm-readobj](https://llvm.org/docs/CommandGuide/llvm-readobj.html#cmdoption-llvm-readobj-arg-file)) accept response files as-is. So I guess we don't need to parse it but we can throw them directly?

ryanking13:
> Actually, it seems like the two tools that we use to calculate exports ([llvm-nm](https://llvm.org/docs/CommandGuide/llvm-nm.html#cmdoption-llvm-nm-arg-file) and [llvm-readobj](https://llvm.org/docs/CommandGuide/llvm-readobj.html#cmdoption-llvm-readobj-arg-file)) accept response files as-is. So I guess we don't need to parse response files but we can pass them directly?

@josephrocca Are you interested in continuing this PR? If not, I think I can continue your work.

josephrocca:
Hey @ryanking13 please do take over, thank you! Feel free to close this and start a new pull request, of course.

Thanks!

Issue #3287, opened 2022-11-24
pyodide build option to install emscripten, rust etc. automatically


joemarshall:
## 🚀 Feature

pyodide build right now auto-installs pyodide distribution, but still needs emsdk to build any binary wheels. It also needs rustup / cargo installed correctly with the right settings to nicely build things that have rust involved. 

It would be nice if when emsdk wasn't found on path, or when it needs rust / cargo, it auto-installs it somehow.

### Motivation

For out-of-tree builds, it is a pain having to install the right emsdk version & correct rust things each time you setup a dev environment. I imagine as more things get supported (e.g. when fortran for wasm finally happens and scipy builds with it), yet more things will be required for out of tree build to work nicely.

### Pitch

1) On building a project, if emsdk isn't found, or emscripten is the wrong version, it should show a prompt asking you to install emsdk or point at existing emsdk, and let you choose location.

2) If a project has rust files, it should prompt you to install rustup/cargo.

3) There should be a command line parameter of some kind to set auto-install tools into a particular directory (or conversely to disable auto-install of tools)

### Additional context

I built pyrus_cramjam which includes rust as an out of tree build, and it was a bit of a pain working out what I needed to make it work.

ryanking13:
I have some work-in-progress PR that aims to move the whole Pyodide build process into `pyodide-build` (https://github.com/pyodide/pyodide/pull/2904). For now, it is only for the in-tree builds but I think we can try extending it to support out-of-tree builds.

PR #3441, opened 2023-01-08
Add rust toolchain to the pyodide-env docker image

ryanking13:
This continues work in #2704 of @josephrocca.

This adds rust toolchain into our docker image so we don't have to install it every time.

`make rust` command still exists, but I removed it from the docs. So now a user (who wants to build a package that requires rust) is responsible for installing rust toolchain in their build setup, which is reasonable I think.

Note that @joemarshall suggested an opposite way in #3287: add a way to install rust in `pyodide-build`. But I personally think it would add unnecessary maintenance burden. Of course, someone could implement it via pyodide-cli in a separate package. 

WDYT?


Dockerfile
```diff
@@ -12,10 +12,15 @@ RUN apt-get update \
...
 
-ADD docs/requirements-doc.txt requirements.txt /
+# Normally, it is a bad idea to install rustup and cargo in
+# system directories (it should not be shared between users),
+# but this docker image is only for building packages, so I hope it is ok.
+RUN wget -q -O - https://sh.rustup.rs | \
+    RUSTUP_HOME=/usr CARGO_HOME=/usr sh -s -- -y --profile minimal --no-modify-path
 
-RUN pip3 --no-cache-dir install -r /requirements.txt \
-  && pip3 --no-cache-dir install -r /requirements-doc.txt
```
ryanking13, line 18: Installing requirements-doc.txt fails because it now contains [a relative path to pyodide-build](https://github.com/pyodide/pyodide/blob/fd49100981dd09ec1c00238456baca58d513bf4c/docs/requirements-doc.txt#L18). I think we don't need build docs inside the docker image anyway.

hoodmane: > we don't need build docs inside the docker image anyway.

I think the docker image should be able to build the docs. Why not `cd src && pip install -r docs/requirements-docs`.

ryanking13: > I think the docker image should be able to build the docs

Okay, I reverted the change.

> Why not cd src && pip install -r docs/requirements-docs.

Actually, `pyodide-build` directory needs to be copied into the docker image.

ryanking13:
Note that this change increases the size of the docker image from 1.74GB to 2.29GB

Dockerfile
```diff
@@ -12,10 +12,19 @@ RUN apt-get update \
...
+ADD requirements.txt docs/requirements-doc.txt /
+ADD pyodide-build /pyodide-build
```
hoodmane, line 22: This seems problematic to me -- doesn't this mean that it won't see local changes to `pyodide-build` in the source directory because we've installed this copy of pyodide-build? So if we update `pyodide-build` we would also have to update the docker image?

hoodmane: I guess it is probably not different than what is already going on though. I'm a bit confused why it works.

ryanking13: > I'm a bit confused why it works.

Well, the readthedocs doesn't use the docker image, it [installs from requirements file](https://github.com/pyodide/pyodide/blob/fd49100981dd09ec1c00238456baca58d513bf4c/.readthedocs.yml#L15). So it will use tot pyodide-build.

When building Pyodide, we [re-installs the pyodide-build](https://github.com/pyodide/pyodide/blob/fd49100981dd09ec1c00238456baca58d513bf4c/packages/Makefile#L16-L17), so the old pyodide-build that are bundled in the docker image will not be used anyway.

ryanking13: So the only case you'll get a old documentation is when you build a documentation from a fresh docker image.

hoodmane (approved):

Issue #3341, opened 2022-12-12
Build libraries in a seperate repository


ryanking13:
Unlike Python packages, static and shared libraries included in Pyodide repository are built with naive emcc commands, not our pywasmcross wrapper. So it can be easily built from separate repositories.

So I think it would be nice if we can prebuild some of the libraries (which doesn't have complicated build scripts) in another repo and use the built libraries in our CI. Probably we can make a new organization like `pyodide-ports` and create repositories in it.

For example, in meta.yaml:

```yaml
# libxml2.zip containing `libxml2.so` and headers.
# Then, we can just uncompress it to `WASM_LIBRARY_DIR` during the build process
source:
    url: http://.../libxml2.zip
    prebuilt: true
```

__Advantages__

1. Can reduce CI time dramatically.
2. People can easily re-use them when building their own python package. For now, we only [vendor some of it in our xbuildenv](https://github.com/pyodide/pyodide/blob/72414962a871d4cb939fe5e4fe04e4627fea6c4a/pyodide-build/pyodide_build/create_xbuildenv.py#L70), so it is a bit hard to use these libraries externally.

__Disadvantages__

1. Adds maintenance burden
2. Updating the Emscripten version will become a bit more stressful. (Though maybe we can automate some of it...)
4. Updating flags for side modules will become a bit more stressful.

Partially related to @joemarshall's issue #3336

WDYT?

rth:
Well the ideal case IMO would be to build all packages in a separate repo. Libraries indeed don't need pywasmcross  but they are still tied to a specific emscripten version used in Pyodide and the associated flags. So in my opinion having a separate setup just for libraries, wouldn't really address our scalability issues, while making how packages are built less homogeneous.

Ideally the workflow I would like is to have,
 - a repo with only meta.yaml and corresponding tests that builds for a given Pyodide stable release.
 - for a new PR build only the changed / added packages
 - deploy them to the CDN under some path that includes the hash of the emscripten/pyodide version
 - have a way to create a distribution repodata.json from either these previously built packages. This would also then make it more realistic to optionally run the test suite of packages.
 - we could still build & test a subset of packages we want in the main repo.
 
It's still a bit blocked by https://github.com/pyodide/pyodide/issues/3049
[...]

PR #3444, opened 2023-01-12
Add tag key in meta.yaml spec

ryanking13:
### Description

I am preparing to make `pyodide build-recipes` CLI to work out-of-tree, so people can write and test recipes in their own repository. This PR is a first step to make it work.

This adds a new `tag` key in meta.yaml spec. A package can have multiples tags, and it is a metadata used to group packages.
For example, `PYODIDE_PACKAGES="tag-name" make` or `pyodide build-recipes tag-name` will build packages which contain that tag.

I removed the hard coded list of `ALWAYS_PACKAGES`, `CORE_PACKAGES`, `CORE_SCIPY_PACKAGES` from pyodide-build, and instead added a `always`, `core`, `min-scipy-stack` tag to corresponding packages.
(Note that we previously had a private key `_tag` used for documentation rendering.)

Partially related: #3341 

### Checklists

- [x] Add / update tests


rth:
> pyodide build-packages tag-name will build packages which contain that tag.

I find it a bit strange that the argument could be a tag or a package name. The reason we can put core in packages to build is that it's a meta-package so still something that relates to a package. Maybe it's just a matter of vocabulary in the documentation. The other alternative would be to build via,
```
pyodide build-packages --tag tag-name
```
but I agree that's probably not very useful.

Another thing, could you specify a list of valid tags in [the pydantic model](https://github.com/pyodide/pyodide/blob/cc7a734c498298bc59dd31f59eaf917ee7688acb/pyodide-build/pyodide_build/io.py#L12)  (via `typing.Literal`) so that if one makes a typo it would be detected?

ryanking13:
> ```
> pyodide build-packages --tag tag-name
> ```
> 
> 
> but I agree that's probably not very useful.

This could work, but it is not very compatible with our current build system that use `PYODIDE_PACKAGES="..." make`. Of course we can introduce another env variable for tags, but I'm a bit reluctant to it.

> Another thing, could you specify a list of valid tags in [the pydantic model](https://github.com/pyodide/pyodide/blob/cc7a734c498298bc59dd31f59eaf917ee7688acb/pyodide-build/pyodide_build/io.py#L12) (via typing.Literal) so that if one makes a typo it would be detected?

Well, there is no invalid tags. I was thinking that people can add any tags they want.

pyodide-build/pyodide_build/recipe.py
```diff
@@ -0,0 +1,95 @@
...
+        except Exception as e:
+            raise ValueError(f"Could not parse {recipe}.") from e
+
+    return recipes
+
+
+def load_recipes(
+    recipe_dir: Path,
+    names_or_tags: Iterable[str],
+    load_always: bool = True,
```
rth, line 29: Maybe we could make  it `load_always_tag` to be less confusing

docs/development/meta-yaml.md
```diff
@@ -45,6 +45,16 @@ Some packages may have multiple top-level import names.
 For instance, `setuptools` exposes `setuptools` and `pkg_resources`
 as a top-level import names.
 
+### `package/tag`
+
+The list of tag of the package. This is a meta information used to group
+the package. Normally this is not needed, but it can be useful to group
+packages by functionality. For example, the `min-scipy-stack` tag is used
+to group packages that are part of the minimal scipy stack.
```
rth, line 53: It could we be useful to list what are currently used tags

pyodide-build/pyodide_build/recipe.py
```diff
@@ -0,0 +1,95 @@
...
+            - "*" : all packages
+            - "no-numpy-dependents" : all packages except those that depend on numpy (including numpy itself)
+    load_always
+        Whether to load packages with the "always" tag/
+
+    Returns
+    -------
+    recipes
+        Dictionary of package name => config
+    flags
```
rth, line 51: I don't fully understand why we need to additionally introduce the notion of flags. We could return the list of tags here, and check if `no-numpy-dependents` is in it, can't we? 

Or at least give it some more explicit name, because it's easy to confuse flags and tags, especially if one has a subset of keys of the other.

rth (commented):
> Maybe we can add prefix for tags like, pyodide build-recipes pkg1 pkg2 tag:tag1 tag:tag2?

+1 for that.

> Well, there is no invalid tags. I was thinking that people can add any tags they want.

OK, but IMO we should still document currently used tags and link with the valid list in https://pyodide.org/en/stable/development/building-from-sources.html?highlight=min-scipy#partial-builds

Because a tag is currently only useful if `pyodide-build` has some internal logic that acts on the tag. Users can add arbitrary tags, but it will have no effect,  so what would be the use case there?  On the other hand, it's easy to mistype a tag, and then packages will not build as expected, which is harder to debug.  But yeah, we can leave the current behavior no strong objection.

pyodide-build/pyodide_build/recipe.py
ryanking13: I wanted to hide details about tags / meta-packages from buildall, but I think it was a bad choice. I removed flags and let `generate_dependency_graph` deal with no-numpy-dependents.

ryanking13:
Okay, I think I addressed most of the suggestions. Thanks for the detailed review @rth!

> Because a tag is currently only useful if pyodide-build has some internal logic that acts on the tag. Users can add arbitrary tags, but it will have no effect, so what would be the use case there? On the other hand, it's easy to mistype a tag, and then packages will not build as expected, which is harder to debug. But yeah, we can leave the current behavior no strong objection.

Yeah, I think most of the users will not use tags (unless someone makes a private set of packages) and this is basically for internal use. The primary goal for this PR is to reduce the dependency between pyodide-build and the pyodide main repository tree so that we are not bothered by ALWAYS and CORE packages when running pyodide-build out-of-tree.

pyodide-build/pyodide_build/recipe.py
```diff
@@ -0,0 +1,101 @@
...
+    recipes: dict[str, MetaConfig] = {}
+
+    for name_or_tag in names_or_tags:
+        # 1. package name
+        if name_or_tag in available_recipes:
+            recipes[name_or_tag] = available_recipes[name_or_tag].copy(deep=True)
+
+        # 2. tag
+        elif name_or_tag.startswith("tag:") and name_or_tag[4:] in tagged_recipes:
+            tag = name_or_tag[4:]
```
hoodmane, line 70:
```suggestion
        elif (
            name_or_tag.startswith("tag:")
            and (tag := name_or_tag.remove_prefix("tag:")) in tagged_recipes
        ):
```

hoodmane (approved):
Thanks @ryanking13! This is a nice change.

ryanking13:
Thanks for the reviews!

Issue #3425, opened 2023-01-05
If `pyodide build` fails, then next call to `pyodide build` should start fresh?


josephrocca:
## 🐛 Bug

It seems like `pyodide build` is kinda "stateful" in some way, or has some persistent side effects. I'd have thought it'd start completely fresh each time (i.e. delete `.pyodide-xbuildenv` and any other state), but maybe some stuff is cached for performance reasons, or something?

I haven't got a clean reproduction unfortunately, but sometimes I get an error like the one below that goes away if I run `pyodide build` again without changing anything or doing anything at all. Please note that this issue is not related to the specifics of the error messages presented here - i.e. I'm not trying to "get past" this error. This bug report is about the inconsistency of consecutive calls to `pyodide build`.

```
running bdist_wheel
running build
running build_py
running create_version
running cmake_build
Extra cmake args: ['-DONNX_USE_LITE_PROTO=ON', '-DProtobuf_INCLUDE_DIR=/workspaces/onnx/protobuf/src', '-DProtobuf_LIBRARIES=/workspaces/onnx/protobuf/build_source', '-DProtobuf_LITE_LIBRARY=/workspaces/onnx/protobuf/build_source/libprotobuf-lite.a', '-Dpybind11_DIR=/usr/local/python/3.10.4/lib/python3.10/site-packages/pybind11/share/cmake/pybind11']
[...]

PR #3445, opened 2023-01-12
FIX Append --fresh flag to let cmake ignore existing cache

ryanking13:
### Description

Resolve: #3425

### Checklists

- [x] Add / update tests


rth (approved):
Thanks for debugging this!

rth:
There is a failure to build nlops with cmake with this change though

PR #3448, opened 2023-01-13
Cleanup of documentation configuration

hoodmane:
Use the version of napoleon vendored in sphinx: `sphinx.ext.napolean` instead of `sphinxcontrib.napoleon`. Remove `language = None`, this causes a warning saying it expected a string. Ignore all dot files. I also removed the monkey patch for `get_object_members` from `autodoc.ModuleDocumenter`. It had no observable effect on the documentation.


rth (approved):
Thanks Hood!

PR #3457, opened 2023-01-16
NFC Add test_raise_js_error

hoodmane:
This adds a simple error test since it seems that all the other tests are more complicated.


ryanking13 (approved):
Thanks!

PR #3458, opened 2023-01-16
MAINT When importing _pyodide_core fails, throw a readable error message

hoodmane:
Helpful when developing the C code.


ryanking13 (approved):
Indeed, thanks!

PR #3459, opened 2023-01-16
MAINT Define `InternalError` and `ConversionError` in _pyodide_core

hoodmane:
Some minor cleanup: define `InternalError` and `ConversionError` in `_pyodide_core` and just import them into C.

### DIFF
--- a/src/py/_pyodide/_core_docs.py
+++ b/src/py/_pyodide/_core_docs.py
@@ -936,6 +936,8 @@ class ConversionError(Exception):
 
 
 class InternalError(Exception):
+    """Thrown when a recoverable assertion error occurs in internal Pyodide code"""
+
     pass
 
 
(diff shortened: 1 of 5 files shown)


src/py/_pyodide/_core_docs.py
```diff
@@ -933,6 +935,10 @@ class ConversionError(Exception):
     """An error thrown when conversion between JavaScript and Python fails."""
 
 
+class InternalError(Exception):
+    pass
```
ryanking13, line 939: Could you add a short docstring about this class?

ryanking13 (approved):
Thanks @hoodmane! It seems like there is a test failure (due to the difference between`pyodide.ffi.ConversionError` and `pyodide.ConversionError` I guess), otherwise LGTM.

PR #3460, opened 2023-01-16
NFC Add accessor macros for JsProxy fields

hoodmane:
This is maintenance split off from #3455.


ryanking13 (approved):
Thanks! It looks much better.

PR #3461, opened 2023-01-17
Update docs requirements

hoodmane:
With this we can get up to Sphinx 5.3.0 (from Sphinx 4.5.0). I dropped sphinx-panels since it has a version pin on Spinx < 5 (we don't seem to use it anyways). I moved to sphinx-book-theme 0.4.0rc1 since version 0.3.0 pins Sphinx < 5. 

This is relevant because new versions of sphinx-autodoc-typehints require Sphinx>=5.3 so we can't pull in my fixes to sphinx-autodoc-typehints.

In the future, I would like for version caps (==, <, <=, and =~) to have comments explaining them. I removed all the version caps since none of them seem to actively break. In particular, we can now use jinja2 version 3 so we don't have to pin docutils 0.16. There's quite a mess of version constraints here... I'm starting to understand why @henryiii is so against defensive version caps, particularly for less actively maintained packages.


docs/requirements-doc.txt
```diff
@@ -1,19 +1,18 @@
...
-sphinx-issues==2.0.0
-sphinx-js==3.2.1
-sphinx-version-warning~=1.1.2
-sphinx-panels
+sphinx>=5.3.0
+sphinx-argparse-cli>=1.6.0
+sphinx_book_theme>=0.4.0rc1
+sphinx-issues
+sphinx-js>=3.2.1
+sphinx-version-warning>=1.1.2
```
ryanking13, line 10: BTW, `sphinx-version-warning` is not maintained and it has a bug that requires [root permission to run](https://github.com/humitos/sphinx-version-warning/issues/22), so probably we should replace it someday.

ryanking13 (approved):
>  we can now use jinja2 version 3 so we don't have to pin docutils 0.16.

Wow, that is really a good news! :) Thanks a lot!

rth:
> I'm starting to understand why henryiii is so against defensive version caps, particularly for less actively maintained packages.

Yeah, I would advocate for using https://github.com/jazzband/pip-tools to pin doc requirements and avoid them breaking periodically while avoiding this upper version limit issue. Currently, it would be very hard to re-build docs from a few years ago with unpinned deps, if we ever need to.

hoodmane:
+1 for that.

PR #3465, opened 2023-01-17
DOCS: Help Spinx find ffi types

hoodmane:
Two fixes:

The `pyodide.ffi` stuff is defined in `_pyodide._core_docs`. We don't want `_pyodide._core_docs` to appear in the documentation because this isn't where you should import things from so we override the `__name__` of `_pyodide._core_docs` to be `pyodide.ffi`. But then Sphinx fails to locate the source for the stuff defined in `_pyodide._core_docs`. This patches `ModuleAnalyzer` to tell it to look for the source of things from `pyodide.ffi` in `_pyodide._core_docs`.

If a function has a typing.overload then the code path that sphinx-autodoc takes is not one that sphinx-autodoc-typehints can handle.
See https://github.com/tox-dev/sphinx-autodoc-typehints/issues/296

### DIFF
--- a/docs/sphinx_pyodide/sphinx_pyodide/__init__.py
+++ b/docs/sphinx_pyodide/sphinx_pyodide/__init__.py
@@ -41,8 +41,28 @@ def patch_templates():
     JsRenderer.rst = patched_rst_method
 
 
+def patch_documenter():
+    from sphinx.ext.autodoc import ModuleAnalyzer
+
+    orig_for_module = ModuleAnalyzer.for_module.__func__
+
+    @classmethod  # type: ignore[misc]
+    def for_module(cls: type, modname: str) -> ModuleAnalyzer:
+        if modname == "pyodide.ffi":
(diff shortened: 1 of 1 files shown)


ryanking13:
> We don't want _pyodide._core_docs to appear in the documentation

Where does this appear in docs? I can't find an example.

hoodmane:
It doesn't appear in the docs because of the hack here:
https://github.com/pyodide/pyodide/blob/main/src/py/_pyodide/_core_docs.py?plain=1#L38
But that hack confuses the sphinx analyzer.

I think this change actually makes no difference in the docs right now because there are no attributes in `_core_docs.py`. But depending on what you put in there, there are sometimes examples where the docs are rendered inconsistently with other files because of the `__name__` hack.

ryanking13 (approved):
Great, thanks for the explanation!

hoodmane:
Thanks for the review!

PR #3466, opened 2023-01-17
More updates to docstrings

hoodmane:
I added a new decorator called `@docs_argspec` to override the argument specification
of a function used in the docs by setting `func.__wraps__` to a fake function. This only
happens when building the docs, normally it is a no-op.

`@docs_argspec` is needed when using `@overload` because mypy requires the argspec
of the main function to be at least as general as the argspecs of all the overloads, which
causes suboptimal rendering in the api docs.


ryanking13 (approved):
Thanks, @hoodmane!

It seems like the linter is complaining about the type of map update, otherwise looks good to me.

PR #3469, opened 2023-01-17
Reproducible package builds (and requirements.txt support)

joemarshall:
Output pinned versions to dist/package-versions.txt when building with dependencies

Build from requirements.txt (with or without dependencies). If you build a pinned version package-versions.txt, it will reproduce the same deps tree.


rth:
Thanks for working on this!

> Build from requirements.txt (with or without dependencies)

Great. Does someone know why pip has  `requirements.txt` installation via the `-r` flag as opposed to without it? It may mean that there are cases where it's difficult to auto-detect requirements.txt file without that flag.

>  Output pinned versions to dist/package-versions.txt when building with dependencies

Could you explain the use case for this? My concern is that `python -m build` doesn't do that, so the question is why should we.

joemarshall:
> Thanks for working on this!
> 
> > Build from requirements.txt (with or without dependencies)
> 
> Great. Does someone know why pip has `requirements.txt` installation via the `-r` flag as opposed to without it? It may mean that there are cases where it's difficult to auto-detect requirements.txt file without that flag.

I wondered that. I'm not quite sure - I think with pip the logic appears to be that there's no reason it can't auto detect, but a requirements.txt is a different thing to a single requirement. It also allows for flexibility in terms of naming etc.

> 
> > Output pinned versions to dist/package-versions.txt when building with dependencies
> 
> Could you explain the use case for this? My concern is that `python -m build` doesn't do that, so the question is why should we.

The use case is that you want to create an exactly repeatable build of a package for CI / testing, you need to build with exactly the same versions across all machines.  python -m build does support basically this functionality if you use a different backend, e.g. poetry uses a `poetry.lock` file to achieve this, pipenv uses `pipenv.lock` to force it in a build virtual environment etc. 

It's also particularly useful for running things in jupyterlite / pyodide, because if you do it this way, you know exactly the set of packages you need to host yourself to avoid pypi downloads (and potentially which packages to auto-load with pyodide.loadPackage on startup so that all your desired imports work straight away), whereas if you use broad version specifiers for the build, you'll get latest versions of things whenever you build, along with potentially added or removed dependencies, which may or may not be built for pyodide and require fixing up.

You can obviously specify exact versions in your pyproject.toml / setup.cfg / setup.py, but the dependencies specified there also go into your wheel, where you don't want exact versions specified otherwise no one will be able to build it (and typically you don't want a flattened dependencies tree in your wheel either). You can also just build the wheel and generate the lock file from the filenames, but given it's a two line change here to do this, I think it's worth putting in.

joemarshall:
@rth Now has a test for a reproducible build, changelog etc.

docs/project/changelog.md
```diff
@@ -20,6 +20,8 @@ myst:
 - {{ Enhancement }} Improved logging in `pyodide-build` with rich.
   {pr}`3442`
 
+- {{ Enhancement}} Added `pyodide-build` support for building a list of packages from a requirements.txt file. Also auto-output a list of chosen dependencies in the same format when building a package and dependencies. This enables repeatable builds of packages. {pr}`3469`
```
rth, line 23: Maybe the second sentence could be a bit rephrased, as  `manually output` by opposition to `auto-output` doesn't make sense for software :)
For instance
```
This command also writes the list of dependencies detected when building packages to `requirements.txt` for better reproducibility.
```

Also repeatable ->  reproducible.

pyodide-build/pyodide_build/cli/build.py
```diff
@@ -151,6 +154,30 @@ def main(
...
         wheel = source(source_location, exports, ctx)
+    elif source_location.lower().endswith(".txt"):
+        # a requirements.txt - build it (and optionally deps)
+        if not Path(source_location).exists():
+            raise RuntimeError(
+                f"Couldn't find requirements text file {source_location}"
+            )
+        with open(source_location) as f:
+            reqs = [x.strip() for x in f.readlines()]
+            try:
```
rth, line 165: Can dedent all the following

pyodide-build/pyodide_build/out_of_tree/pypi.py
```diff
@@ -351,14 +340,80 @@ def build_dependencies_for_wheel(
...
+        for x in result.mapping.values():
+            download_or_build_wheel(x.url, target_folder)
+            if len(x.extras) > 0:
+                extratxt = "[" + ",".join(x.extras) + "]"
+            else:
+                extratxt = ""
+            version_file.write(f"{x.name}{extratxt}=={x.version}\n")
+
+
+def build_multiple_wheels_from_pypi(
```
rth, line 360: Maybe,
```
build_wheels_from_pypi
```
it's already plurial

pyodide-build/pyodide_build/out_of_tree/pypi.py
```diff
@@ -324,23 +317,19 @@ def build_dependencies_for_wheel(
...
-    if metadata is None:
-        raise RuntimeError(f"Can't find package metadata in {wheel}")
 
-    deps: list[str] = metadata.get_all("Requires-Dist", [])
+def _resolve_and_build(
+    deps: list[str],
+    target_folder: Path,
+    build_dependencies: bool,
+    extras: list[str],
+) -> None:
```
rth, line 326: A 1 line docstring with a short description wouldn't hurt.

rth (commented):
Thanks @joemarshall !

>  does support basically this functionality if you use a different backend, e.g. poetry uses a poetry.lock file to achieve this, pipenv uses pipenv.lock to force it in a build virtual environment etc.

I'm well aware of the usefulness of the lock files for reproducibility. My concern is that we started out as an equivalent tool to PyPA `build` and now are adding more features on top, with some hardcoded filename and location lockfiles that none of the other tools use. There was a recent [blog post](https://pradyunsg.me/blog/2023/01/21/thoughts-on-python-packaging/) about how having N Python packaging tools that each make slightly different (but reasonable) choices, including for lockfiles, results in the current rather confusing situation for users. I just want to make sure that we are not contributing to this problem.

In this case, maybe we could do add this functionality via an optional `--output-lockfile` argument (need to check if there is already some similar argument in existing tools, and we could use the same name). So that we are a bit less opinionated about the file name of this lock file or its location?

Also depends what @hoodmane @ryanking13 think about it.

Code wise LGTM aside from a few comments below, but a second code review also wouldn't hurt.

joemarshall:
That sounds sensible. I did wonder about whether to put a flag on it.

I guess a good aim with pyodide build would be to have it sitting outside packaging tools somehow. I guess it would have to create a venv which would a) override pypa build, and b) override platform settings, but without actually running python in pyodide. 

Basically, an environment where pip, pip-tools, pipenv, poetry or whatever work, and build pyodide wheels but without the overhead of firing up pyodide for each python process.

ryanking13 (commented):
Thanks, @joemarshall. To be honest, I am still not sure if pyodide-build should deal with the complicated dependency resolution of Python. But since there is no package index or package manager for WebAssembly wheels, I think it makes sense that pyodide-build should support those features for now.

In the future, I really hope that we can reuse some other tools for dependency resolution. Such as creating lock files using poetry or pip-tools... though I understand there are several requirements mentioned in #3196. I would appreciate it if you are interested in improving this situation.

The codewise looks good to me. About the lockfile, I'm +1 to Roman's idea of having an optional flag for it.

joemarshall:
Okay, I've done the review fixes suggested above. Hopefully ready to go.

pyodide-build/pyodide_build/cli/build.py
```diff
@@ -130,6 +134,10 @@ def main(
     build_dependencies: bool = typer.Option(
         False, help="Fetch non-pyodide dependencies from pypi and build them too."
     ),
+    output_lockfile: bool = typer.Option(
+        False,
+        help="Output list of resolved dependencies alongside the build in dist/package_versions.txt",
```
ryanking13, line 139: As @rth mentioned, I think it would be nice to make this flag to specify a path of a lockfile, instead of a boolean flag. So that we are a bit less opinionated about the file name of this lock file or its location.

pyodide-build/pyodide_build/tests/test_pypi.py
```diff
@@ -307,3 +307,67 @@ def test_fake_pypi_extras_build(selenium, tmp_path, fake_pypi_url):
...
+    # TODO: - make test run without pyodide
+    output_dir = tmp_path / "dist"
+    # build package that resolves right
+    app = typer.Typer()
+    app.command()(build.main)
+
+    # override a dependency version and build
+    # pkg-a
+    with open(tmp_path / "requirements.txt", "w") as req_file:
+        req_file.write(
```
ryanking13, line 322: Maybe let's use `writelines` for readability.

pyodide-build/pyodide_build/cli/build.py
```diff
@@ -151,6 +158,47 @@ def main(
     elif Path(source_location).is_dir():
         # a folder, build it
         wheel = source(source_location, exports, ctx)
+    elif source_location.lower().endswith(".txt"):
```
ryanking13, line 161: Since a package name can actually endswith ".txt" (though it is not very recommended...). I think we could use a `-r` flag for requirements files same as pip.

ryanking13 (approved):
Thanks @joemarshall and sorry for the late review. I have some minor comments otherwise looks good to me.

joemarshall:
@ryanking13 @rth 

Review changes above done now, so it is `pyodide build -r requirements.txt`, and `pyodide build --output-lockfile lockfile.txt` (and changelog updated to reflect them)

rth (approved):
Thanks for doing this work @joemarshall and for your patience to get this merged!

ryanking13:
Yes, thanks again @joemarshall!

PR #3473, opened 2023-01-17
Add sphinx-design, fix tabs

hoodmane:
In #3461 I dropped `sphinx-panels` but of course we *were* using it. It has a successor called
`sphinx-design` which works with sphinx 5.x (still not 6.x but we have several packages that cap 
sphinx <6). I also updated the use of the tabbed directive to the new sphinx-design api.


ryanking13 (approved):
Thanks!

PR #3477, opened 2023-01-19
DOCS Add sphinx cross referencing to MDN docs using intersphinx

hoodmane:
And updated docstrings and documentation to use 
```rst
:sphinx:xref:`style`
```
for most things rather than including full urls. Also added a lot more xrefs to MDN docs.


ryanking13 (approved):
Cool! Thanks, @hoodmane.

PR #3481, opened 2023-01-21
DOCS Improve JavaScript type rendering

hoodmane:
This improves the rendering of JavaScript types in the api docs.

## Before:
Screenshot from 2023-01-21 12-48-20

## After:

Screenshot from 2023-01-21 12-47-53


rth (approved):
Great to have cross-linking with JS types, thanks!

PR #3484, opened 2023-01-22
Fix `from pyodide.ffi import *`

### LINKED ISSUE #3432: import ffi * is broken for 22.0+
## 🐛 Bug

Can no longer call `from pyodide.ffi import *`
Traceback
```
  File "/lib/python3.10/_pyodide/_base.py", line 468, in eval_code
    .run(globals, locals)
  File "/lib/python3.10/_pyodide/_base.py", line 312, in run
    coroutine = eval(self.code, globals, locals)
  File "<exec>", line 2, in <module>
  File "/home/pyodide/launchapp.py", line 1, in <module>
    from pyodide.ffi import *
AttributeError: module 'pyodide.ffi' has no attribute 'IN_BROWSER
```

### To Reproduce

`from pyodide.ffi import *
`
### Environment

- Pyodide Version: v0.22.0 and dev 
- Browser version: Chrome

### Additional context
Not sure if IN_BROWSER is supposed to be available in that package or not. This is the [commit](https://github.com/pyodide/pyodide/pull/3298/files#diff-d66f085dfec63b0802ed32af9c70f0ac5dd1bcf599805afa45fce94599517d1e) that broke it. There probably needs to be some tests added to make sure the import * works for all the modules that are defining `__all__`. Not sure what the proper correction is to the current package structure though. _core doesnt define it in its all and I am unsure if that is the actual bug.

hoodmane:
Resolves #3432.


rth (approved):
This is indeed much better, thanks! Changelog needed.

PR #3487, opened 2023-01-22
Backports for 0.22.1

hoodmane:


rth:
Thanks for fixing a lot of those 0.22.0 issues! It looks like nothing is left in 0.21.1 milestone, do you plan to add more fixes before merging this?

hoodmane:
There's a few more that need to be cherry picked but then I think we can release 0.21.1.

hoodmane:
Maybe we can include #3496. We should also fix the console deploy...

ryanking13:
> We should also fix the console deploy...

It seems like readthedocs changed the output directory from `_build` to `_readthedocs`, which broke our console.html deploy.

- https://github.com/readthedocs/readthedocs.org/pull/9888

PR #3488, opened 2023-01-22
Fix setStdin to insert a null after each buffer if autoEOF parameter is true.

### LINKED ISSUE #3413: Infinite prompt by using an input in v0.22.0
## 🐛 Bug

When I'm using `input` with the new version, show an prompt in infinite mode.

### To Reproduce

`input('Type text')`

### Expected behavior

Display the alert only once. 

### Environment

- Pyodide Version : 0.22.0

### Additional context

https://www.loom.com/share/08ccfbf98afc43a984feb4eccfd53d3f

hoodmane:
Resolves #3413.

I added a new argument to `setStdin` called `autoEOF` which if `true` causes `setStdin` to automatically
insert an EOF after each buffer or string. It defaults to `true`, which is a breaking change. Set it to `false`
to get the behavior from before this PR.

I also added an option that allows the stdin function to directly return a character code.

- [ ] Add / update tests

### DIFF
--- a/docs/project/changelog.md
+++ b/docs/project/changelog.md
@@ -28,6 +28,10 @@ myst:
 - {{ Fix }} `pyodide config` won't print extra messages anymore.
   {pr}`3483`
 
+- {{ Breaking }} `setStdin` now accepts an extra `autoEOF` parameter. If `true`,
+  it will insert an EOF automatically after each string or buffer. Defaults to `true`.
+  {pr}`3488`
+
 ### Build System
 
 - {{ Enhancement }} Improved logging in `pyodide-build` with rich.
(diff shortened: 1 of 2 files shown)


ryanking13 (approved):

> It defaults to true, which is a breaking change

Is this also a breaking change for people who were using 0.21.X? If not, I think it is actually okay to add this to 0.22.1 as a bugfix. No strong opinion though.

hoodmane:
No, it fixes the behavior back to 0.21.X behavior. I am leaning toward adding this to 0.22.1.

rth:
Thanks a lot for addressing this @hoodmane !

PR #3490, opened 2023-01-22
fix: pass only build args

hoodmane:
This is #3417 backported to stable for 0.22.1. Thanks @agoose77!


rth (commented):
Thanks!

PR #3495, opened 2023-01-23
Pass same environment variables to out of tree builds as in tree ones

### LINKED ISSUE #3493: Detecting Pyodide, at build time
## 📚 Documentation

The current documentation reads:

To detect Pyodide, at build time use,

```
import os

if "PYODIDE" in os.environ:
    # building for Pyodide
```

This used to work on my computer using pyodide-build v0.21.

When I tried it using pyodide-build v0.22 however, the test `if "PYODIDE" in os.environ` returned false.

Inspecting my `os.environ`, I changed the test to `if ("PYODIDE_ROOT" in os.environ):`, and my setup.py detects when the code is being built with pyodide.

This lead me to believe the docs needed updating. However, searching the pyodide repo, I saw `if "PYODIDE" in os.environ` in the following patch file: https://github.com/pyodide/pyodide/blob/main/packages/Pillow/patches/0001-Enable-image-formats.patch

So I'm not really sure anymore ...

hoodmane:
Resolves #3493

### Description

- [ ] Add / update tests

### DIFF
--- a/pyodide-build/pyodide_build/common.py
+++ b/pyodide-build/pyodide_build/common.py
@@ -459,3 +459,21 @@ def chdir(new_dir: Path) -> Generator[None, None, None]:
         yield
     finally:
         os.chdir(orig_dir)
+
+
+def set_build_environment(env: dict[str, str]) -> None:
+    env.update({key: os.environ[key] for key in BUILD_VARS})
+    env["PYODIDE"] = "1"
+    if "PYODIDE_JOBS" in os.environ:
+        env["PYODIDE_JOBS"] = os.environ["PYODIDE_JOBS"]
+
+    env["PKG_CONFIG_PATH"] = env["WASM_PKG_CONFIG_PATH"]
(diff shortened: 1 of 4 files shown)


pyodide-build/pyodide_build/common.py
```diff
@@ -459,3 +459,21 @@ def chdir(new_dir: Path) -> Generator[None, None, None]:
         yield
     finally:
         os.chdir(orig_dir)
+
+
+def set_build_environment(env: dict[str, str]) -> None:
```
ryanking13, line 464: Could you please write a docstring for this function?

ryanking13 (approved):
Thanks, @hoodmane! Please update the changelog as this would be a breaking change for packages that use out-of-tree build.

About the CI failure, I think the problem is that the cargo is not meant to be used to build several packages in parallel. We will need to find a workaround for it (Not related to this PR).

PR #3496, opened 2023-01-24
Improve messages in case of system exit

### LINKED ISSUE #3486: Possible bug: Pyodide has suffered a fatal error with tag [object Object]
## 🐛 Bug

```
Pyodide has suffered a fatal error. Please report this to the Pyodide maintainers. pyodide.asm.js:10:154032
    fatal_error pyodide.asm.js:10
    callPyObjectKwargs pyodide.asm.js:10
    callPyObject pyodide.asm.js:10
    apply pyodide.asm.js:10
    apply pyodide.asm.js:10
    runPython pyodide.asm.js:10
The cause of the fatal error was: pyodide.asm.js:10:154132
Error: A value of type object with tag [object Object] was thrown as an error!
String interpolation of the thrown value gives """[object Object]""".
The thrown value's toString method returns """[object Object]""".
```

### To Reproduce (console for extra info)

```html
<!DOCTYPE html>
<html>
  <head>
    <script src="https://cdnjs.cloudflare.com/ajax/libs/pyodide/0.22.0/pyodide.min.js"></script>
  </head>

  <body>
    <textarea readonly id="output" rows=24 cols=80></textarea>
  </body>

  <script>
    async function main() {
      const output = document.getElementById("output");
      let pyodide = await loadPyodide({indexURL: "https://cdn.jsdelivr.net/pyodide/v0.22.0/full/"});
      console.log("pyodide loaded");

      console.log("micropip loaded");
      await micropip.install("mypy");
      console.log("mypy loaded");

      try {
        pyodide.runPython(`
          import io
          import sys
          sys.stdout = io.StringIO()
        `);
        console.log("setup loaded. Wait. Next part will take some time.");

        pyodide.runPython(`
          import mypy.main
          with open("test.py", "w") as f:
              f.write('def test() -> None: return 5')
          mypy.main.main(args=["test.py"])
        `);
        console.log("code loaded");

        output.value = pyodide.runPython(`
          sys.stdout.getvalue()
        `);
        console.log("done");
      } catch (err) {
        output.value = err;
        console.log("error");
      }

      return pyodide;
    }

    main();
  </script>
</html>
```

### Expected behavior

Expected Pyodide to not crash. Expected output is (from `mypy`):
```
test.py:1: error: No return value expected  [return-value]
Found 1 error in 1 file (checked 1 source file)
```

Code can be run local (`pip install mypy`):
```py
import mypy.main

with open("test.py", "w") as f:
    f.write('def test() -> None: return 5')

mypy.main.main(args=["test.py"])
```

### Environment

- Pyodide Version: 0.22.0 from CDN (see HTML above)
- Browser version: all
- Any other relevant information: none

### Additional context

Trying to run `mypy` on file in MEMFS (default). Workarounds accepted.

hoodmane:
This prints a better set of error messages in case someone calls `os._exit()` or in C code `exit()` is used.

Resolves #3486.

### DIFF
--- a/docs/project/changelog.md
+++ b/docs/project/changelog.md
@@ -36,6 +36,10 @@ myst:
   it will insert an EOF automatically after each string or buffer. Defaults to `true`.
   {pr}`3488`
 
+- {{ Enhancement }} Pyodide displays a better message when someone calls posix
+  `exit` or `os._exit`.
+  {pr}`3496`
+
 ### Build System
 
 - {{ Enhancement }} Improved logging in `pyodide-build` with rich.
--- a/src/core/error_handling.ts
+++ b/src/core/error_handling.ts
@@ -8,7 +8,7 @@ function ensureCaughtObjectIsError(e: any): Error {
   if (typeof e === "string") {
     // Sometimes emscripten throws a raw string...
     e = new Error(e);
-  } else if (typeof e === "object" && e.name === "ExitStatus") {
+  } else if (e && typeof e === "object" && e.name === "ExitStatus") {
     let status = e.status;
     e = new Exit(e.message);
     e.status = status;
--- a/src/core/error_handling.ts
+++ b/src/core/error_handling.ts
@@ -8,6 +8,10 @@ function ensureCaughtObjectIsError(e: any): Error {
   if (typeof e === "string") {
     // Sometimes emscripten throws a raw string...
     e = new Error(e);
+  } else if (typeof e === "object" && e.name === "ExitStatus") {
+    let status = e.status;
+    e = new Exit(e.message);
+    e.status = status;
   } else if (
     typeof e !== "object" ||
     e === null ||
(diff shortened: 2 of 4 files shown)


rth (approved):
Thanks! Do we need a test for this?

So there are no use cases where the users might want to call exit? Say if they create multiple pyodide instances, and try to exit one from Pyodide. It will quit the instance, but will have no impact on the used memory?

hoodmane:
> So there are no use cases where the users might want to call exit? Say if they create multiple pyodide instances, and try to exit one from Pyodide. It will quit the instance, but will have no impact on the used memory?

I am not sure. I don't mean to say that this is the correct behavior. Currently calling exit ends up in the fatal error handler, so I thought we should make that a bit smarter. But there might be something more sensible we could do.

rth:
OK, I understand. Well, it's an improvement over the current situation in any case. Let's merge it.

PR #3504, opened 2023-01-25
Add compiled mypy

### LINKED ISSUE #3497: Why is mypy so slow in Pyodide?
# Speeding up code

Do you know how to speed up the code? It takes less than a second to run on a standard Python installation without Pyodide, but over a minute with Pyodide:

```py
import os
os._exit = lambda arg: None

import mypy.main

with open("test.py", "w") as f:
    f.write('def test() -> None: return 5')

mypy.main.main(args=["test.py"])
```

Is Pyodide performance that bad?

# Timings

Running it twice in a row (using cache) still takes a while:
```py
# Pyodide: 60 seconds (creates cache)
# Local: < 1 second
mypy.main.main(args=["test.py"])

# Pyodide: 5 seconds (uses cache)
# Local: < 1 second
mypy.main.main(args=["test.py"])
```

Cache can be disabled, in which case, it takes the full time:
```py
# Pyodide: 60 seconds
# Local: < 1 second
mypy.main.main(args=["--cache-dir=/dev/null", "test.py"])

# Pyodide: 60 seconds
# Local: < 1 second
mypy.main.main(args=["--cache-dir=/dev/null", "test.py"])
```

_Originally posted by @ulmemxpoc in https://github.com/pyodide/pyodide/issues/3486#issuecomment-1399573992_

hoodmane:
Resolves #3497.

### DIFF
--- a/packages/mypy/meta.yaml
+++ b/packages/mypy/meta.yaml
@@ -8,7 +8,8 @@ source:
   url: https://files.pythonhosted.org/packages/0e/5c/fbe112ca73d4c6a9e65336f48099c60800514d8949b4129c093a84a28dc8/mypy-0.991.tar.gz
   sha256: 3c0165ba8f354a6d9881809ef29f1a9318a236a6d81c690094c5df32107bde06
 build:
-  script: export MYPY_USE_MYPYC=1
+  script: |
+    export MYPY_USE_MYPYC=1
     export MYPYC_OPT_LEVEL=3
 about:
   home: http://www.mypy-lang.org/
--- a/packages/mypy/meta.yaml
+++ b/packages/mypy/meta.yaml
@@ -1,0 +1,17 @@
+package:
+  name: mypy
+  version: "0.991"
+  top-level:
+    - mypyc
+    - mypy
+source:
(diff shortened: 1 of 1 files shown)


packages/mypy/meta.yaml
```diff
@@ -0,0 +1,17 @@
...
+  name: mypy
+  version: "0.991"
+  top-level:
+    - mypyc
+    - mypy
+source:
+  url: https://files.pythonhosted.org/packages/0e/5c/fbe112ca73d4c6a9e65336f48099c60800514d8949b4129c093a84a28dc8/mypy-0.991.tar.gz
+  sha256: 3c0165ba8f354a6d9881809ef29f1a9318a236a6d81c690094c5df32107bde06
+build:
+  script: export MYPY_USE_MYPYC=1
```
ryanking13, line 11:
```suggestion
  script: |
    export MYPY_USE_MYPYC=1
```

ryanking13 (approved):
Thanks, @hoodmane! This would need a changelog and test.

It is worth noting that it takes 9min to build mypy in CI.

PR #3505, opened 2023-01-25
Disable AUTO_JS_LIBRARIES and AUTO_NATIVE_LIBRARIES flags

ryanking13:
### Description

These two flags are enabled by default and add extra JS and native libraries to the main module:

See: 

1) https://github.com/emscripten-core/emscripten/blob/aef3f83323bb1e9eed5b6ca9356469419ac30a25/tools/system_libs.py#L2080-L2083

2) https://github.com/emscripten-core/emscripten/blob/d46b42480f0b76216a458ff24df62df7fb0186de/src/modules.js#L107-L120

I think we can disable them by default, and link each library explicitly when we need them. For example, we can link SDL-related libraries when we started to use SDL-based libraries (I will open a PR soon).

This change decreases the size of the main module (pyodide.asm.js) ~0.3MB:

Before:

```
stat --format="%s %n" pyodide.asm*
5145705 pyodide.asm.data
1549713 pyodide.asm.js
8926762 pyodide.asm.wasm
```

After:

```
5145705 pyodide.asm.data
1204075 pyodide.asm.js
8862216 pyodide.asm.wasm
```

### Checklists


rth (approved):
Thanks!

PR #3507, opened 2023-01-25
Apply upstream cpython commits to relocate test directories

ryanking13:
Fetch some CPython upstream commits that [relocate stdlib test files into `test` directory](https://github.com/python/cpython/issues/93839). This is a  preparation for #3166.
Previously `ctypes` and `unittest` modules had test directories in them so it was hard to bundle them in a zipfile. I hoped this to be included in Python 3.11, but sadly it was delayed to 3.12.


hoodmane (approved):
Thanks @ryanking13!

PR #3510, opened 2023-01-25
DOCS Fix console.html deploy

hoodmane:
Put `console.html` in `app.outdir`.

See readthedocs/readthedocs.org PR 9888


ryanking13:
It seems like the console in the latest docs is still broken. The script tag for pyodide.js has an invalid src value.

PR #3512, opened 2023-01-26
Refactor conf.py

hoodmane:
Instead of putting stuff behind `IN_SPHINX`, define functions and call them from the `setup` function. 
In these functions, if we want to expose variables as part of the config we have to assign to `app.config.some_var` which is more explicit.

We still have to make the path change at top level. To improve this, in the future we should:
1. rename the sphinx_pyodide folder to sphinx-pyodide
2. add a `pyproject.toml` and `setup.py` so we can `pip install -e` it
3. instead of modifying the path, source the virtual environment

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -138,17 +138,14 @@ extra_sys_path_dirs = [
 
 # Try not to cause side effects if we are imported incidentally.
 
-try:
-    import sphinx
-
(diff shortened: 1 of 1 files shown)


ryanking13 (approved):
Indeed it looks better, thanks!

PR #3514, opened 2023-01-27
Add BuildArgs to store common build arguments 

ryanking13:
### Description

This renames `ReplayArgs` in pywasmcross to `BuildArgs` and uses it in buildall and buildpkg. No functional changes are involved except for some modifications in function parameters and how logs are copied after the build.

This PR makes it easier to call methods in buildall and buildpkg from the new CLI.

### Checklists

- [x] Add / update tests


pyodide-build/pyodide_build/pywasmcross.py
```diff
@@ -51,25 +51,28 @@
...
-        "ldflags",
-        "target_install_dir",
-        "builddir",
-        "pythoninclude",
-        "exports",
-    ],
-)
+
+@dataclasses.dataclass(eq=False, order=False, kw_only=True)
+class BuildArgs:
```
rth, line 62: Probably not critical, but we could make this a Pydantic class (inherit from pydantic.BaseModel and remove the decorator), so that types would be enforced at runtime, and then if you specify the type of `target_install_dir` and `host_install_dir`  as `Path` it will auto-convert to that, so we won't have to do it later. The API would be identical otherwise.

rth (approved):
Thanks, this is better indeed!

pyodide-build/pyodide_build/pywasmcross.py
ryanking13: Yes, that sounds great. Though probably we need to check the runtime overhead. pywasmcross is invoked thousand of times during the compilation so I think we need to be careful when adding extra imports and some heavy logics.

PR #3518, opened 2023-01-28
MAINT Remove "use strict" pragmas from source code

hoodmane:
We insert these at the top of the file with sed. It's redundant to put them in individual functions.


ryanking13 (approved):
Thanks!

PR #3520, opened 2023-01-30
ENH Add --no-deps parameter to build-recipes CLI

ryanking13:
### Description

This adds `--no-deps` parameter to `pyodide build-recipes` CLI, which is a replacement for `pyodide_build buildpkg` entrypoint.

Ref: #2879

### Checklists

- [x] Add / update tests

### DIFF
--- a/docs/project/changelog.md
+++ b/docs/project/changelog.md
@@ -67,6 +67,10 @@ myst:
   This command is hidden by default since it is not intended for use by end users.
   {pr}`3411` {pr}`3463`
 
+- `pyodide build-recipes` now accepts `--no-deps` parameter,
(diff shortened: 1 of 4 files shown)


hoodmane:
Will this display the stdout/stderr on the console as it is executing? Because that is the main reason I use `buildpkg`. Repeatedly using `cat packages/pkgname/build.log | tail` is okay but it's nice to just see the output.

ryanking13:
> Will this display the stdout/stderr on the console as it is executing? Because that is the main reason I use buildpkg.

Yes, this will show the exactly same output to the console. The exception-handling part that I added will only happen when running tests.

hoodmane (approved):

PR #3522, opened 2023-01-30
Replace isort, flake8, pycln and pyupgrade by ruff

rth:
[Ruff](https://github.com/charliermarsh/ruff) is getting popular lately, and I really like the idea of having a single tool for Python lining instead of N which currently exist. Also written in Rust which is a good use case for this.

 It's supposed to have equivalent functionality, to the ones we are using. The fact that major projects are adopting it is a good sign (see list in their readme).  There are also a few extra options in their readme that look interesting and that we could enable (without having to install yet another plugin)

We won't gain much speed-wise: previously isort, flake8 (which itself is an abstraction over pycodestyle and pyflakes), pycln and pyupgrade  on all files was around 1s, and now it's 20ms. But with mypy taking 10s, it makes mostly no difference.
Still it makes the overall linting configuration easier to maintain.

If we go forward with this, I imagine a changelog is not needed?


src/py/pyodide/__init__.py
```diff
@@ -17,20 +17,24 @@
...
-from .ffi import destroy_proxies  # noqa: F401
-from .ffi import register_js_module  # noqa: F401
-from .ffi import to_js  # noqa: F401
-from .ffi import unregister_js_module  # noqa: F401
+from .code import (
+    CodeRunner,  # noqa: F401
+    eval_code,  # noqa: F401
+    eval_code_async,  # noqa: F401
+    find_imports,  # noqa: F401
+    should_quiet,  # noqa: F401
```
rth, line 25: Looks like its parser is able to handle comments unlike isort

rth:
It looks like it's failing to build on pre-commit.ci though, with no error message and it's not clear where I can report this on to pre-commit.ci. It looks like there is intentionally no way to report issues there unless I missed something :(

rth:
Well, it's building fine with pre-commit.ci in another simpler pre-commit setup [here](https://results.pre-commit.ci/run/github/116300991/1675117738.KdUyTIykRt2D15f4jYRJvw)  (for a different repo).

ryanking13:
Thanks, @rth! Ruff looks very cool :)

> No idea why.

Hmm, I can run this locally, and GHA lint job also works fine... so it is not clear what is happening in pre-commit.ci.

henryiii:
I've been using it elsewhere and it's been fine on pre-commit.ci, too. Maybe try rebuilding in case it's a fluke?

src/py/pyodide/__init__.py
henryiii: This really should have an `__all__`, which would make the comments unneeded.

pyproject.toml
```diff
@@ -40,21 +40,33 @@ module = [
...
-[tool.pycln]
-all = true
+[tool.ruff]
+# F: pyflakes
+# E, W: pycodestyles
+# where flake8 = pyflakes + pycodestyles
+# B: flake8 - bugbear
+# UP: pyupgrade
+# I: isort
+select = ['E', 'F', "W", 'B', 'UP', 'I']
```
henryiii, line 50: It might be better to follow normal bugbear usage and select `B0` + whatever extra B9 checks you use (`B904` and `B905` are the only ones there right now, IIRC). Unlike flake8, selecting "X" selects "X9*" too. Bugbear doesn't mind adding _very_ opinionated checks to the 9 series.

pyproject.toml
```diff
@@ -40,21 +40,33 @@ module = [
...
+# Remove E999 once pattern matching is supported
+# https://github.com/charliermarsh/ruff/issues/282
+ignore = ["E402", "E501","E731", "E741", "E999"]
```
henryiii, line 53:
```suggestion
ignore = ["E402", "E501", "E731", "E741", "E999"]
```

pyproject.toml
henryiii: Also, I recommend putting the comments inline, or they might get outdated. See https://github.com/scikit-build/scikit-build-core/blob/d64e68f7db91a1adee1d6d4ae281899a4f45b2a8/pyproject.toml#L192-L212 for example.

src/py/pyodide/__init__.py
hoodmane: We didn't want to put these in `__all__` because they are deprecated as exports.

pyodide-build/pyodide_build/tests/test_pywasmcross.py
```diff
@@ -2,13 +2,13 @@
...
 
-from pyodide_build.pywasmcross import handle_command_generate_args  # noqa: E402
-from pyodide_build.pywasmcross import replay_f2c  # noqa: E402
 from pyodide_build.pywasmcross import (
     BuildArgs,
     calculate_exports,
     get_cmake_compiler_flags,
     get_library_output,
+    handle_command_generate_args,  # noqa: E402
+    replay_f2c,  # noqa: E402
```
hoodmane, line 11: But probably these ones should be fixed by adding an `__all__` to pywasmcross.py?

src/py/pyodide/__init__.py
hoodmane: Actually according to our deprecation timeline these should be removed in 0.23.0, so we can take this stuff out.

pyodide-build/pyodide_build/tests/test_pywasmcross.py
ryanking13: I think that ignoring E402 is not needed anymore. It is about "module level import not at top of file" but now the import is done in top of file.

ryanking13 (approved):
Thanks!

src/py/pyodide/__init__.py
rth: Yes, good point. Will do it in a separate PR so it's more easily discoverable as users will likely hit it it they haven't paid attention to console warnings.

rth (commented):
Thank you all for your reviews and your comments!

cclauss:
https://github.com/charliermarsh/ruff/issues/282 is now closed.  What is required to remove this [E999](https://github.com/search?q=repo%3Apyodide%2Fpyodide+E999)?
```
# Remove E999 once pattern matching is supported
# https://github.com/charliermarsh/ruff/issues/282
ignore = ["E402", "E501", "E731", "E741", "E999"]
```

henryiii:
I'd guess only upgrading to the latest Ruff. It should support full 3.11 syntax now.

PR #3523, opened 2023-01-31
Add new `pyodide.ffi` JavaScript submodule with PyProxy subtypes

hoodmane:
This creates a new `pyodide.ffi` submodule and adds a bunch of new subclasses of `PyProxy` to it.

There are three stages in which we are concerned with the behavior of the objects we define:
1. at time of static typechecks
2. at execution time
3. when generating docs

Prior to this PR, the subtypes of PyProxy only work well for static type checks, they work acceptably at runtime (just don't let the user access them), and the docs don't look that great. This PR is primarily intended to improve the docs for PyProxy, but they also make execution time checks work better: you can now say `obj instanceof pyodide.ffi.PyCallable` instead of `obj.isCallable()` which I is easier to understand and to cross reference against the documentation. I am marking `isCallable` as deprecated.

I also made a bunch of edits and improvements to the docs.

I have deprecated `PyProxyCallable` in favor of `pyodide.ffi.PyCallable` and `PyProxy.isCallable` in favor of `obj instanceof pyodide.ffi.PyCallable`.
`PyBuffer` has been renamed to `pyodide.ffi.PyBufferView` and a new `PyBuffer` has been created which is a subtype of `PyProxy`.

### Checklists

- [x] Add a deprecation timeline entry
- [x] Add / update tests


ryanking13:
Thanks, @hoodmane! I'll review this soon. It seems like the CPython build has been broken since yesterday... not sure why.

> One annoyance here is the presence of PyBuffer. I would like to rename the current PyBuffer to something else (PyBufferView?) and rename the thing I have currently called PyProxyBuffer to PyBuffer. This may need to happen over a few releases to make time for deprecation warnings.

+1 for renaming PyBuffer. Deprecation warnings until 0.24.0 (or 0.25.0) would be nice I think.

> Another thought: perhaps we could rename JsProxy to JsObject and PyProxy to PyObject?

I am not sure about this. Those two classes are highly used by users (though not directly), so I think we should be careful about renaming those classes. We have been also using the term "proxy" for a long time, including methods like `create_proxy` and `destroy_proxies`.

ryanking13 (approved):
Thanks, @hoodmane. I have a comment about exposing PyProxy classes, otherwise looks good to me.

src/js/api.ts
```diff
@@ -504,6 +523,21 @@ export type PyodideInterface = {
...
+  PyDict: typeof PyDict;
+  PyIterable: typeof PyIterable;
+  PyIterator: typeof PyIterator;
+  PyAwaitable: typeof PyAwaitable;
+  PyCallable: typeof PyCallable;
+  PyProxyBuffer: typeof PyProxyBuffer;
+  PyAsyncIterable: typeof PyAsyncIterable;
+  PyAsyncIterator: typeof PyAsyncIterator;
+  PyGenerator: typeof PyGenerator;
+  PyAsyncGenerator: typeof PyAsyncGenerator;
```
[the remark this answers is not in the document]
hoodmane: Yeah that sounds good to me.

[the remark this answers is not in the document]
hoodmane: Okay I think I worked out how to "correctly" do submodules like this.

hoodmane:
Thanks for this detailed explanation @mneil!c

> IDK if that deprecation in the doc transfers over to the generated typings or not or how that works. I would have to play around with it.

It seems to me that the type generator we are using does include documentation comments in the generated docs so this should work again supposing we have cooperation from linters.

hoodmane:
@ryanking13 this changed a lot since you reviewed it so if you could review again I would appreciate it. I could split it into multiple PRs too since it's a large change set.

hoodmane:
@mneil `tsd` has `expectDeprecated` and `expectNotDeprecated` but they seem to be a bit spotty.

docs/sphinx_pyodide/sphinx_pyodide/packages.py
```diff
@@ -27,7 +27,10 @@ def run(self):
 
             packages = {}
             for package in packages_list:
-                name, version, is_package, tag = self.parse_package_info(package)
+                try:
+                    name, version, is_package, tag = self.parse_package_info(package)
+                except Exception:
+                    continue
```
ryanking13, line 33: Perhaps print some warning message if the parsing fails, so we can check?

Makefile
```diff
@@ -135,9 +135,14 @@ dist/package.json : src/js/package.json
...
 
-dist/pyodide.d.ts: src/js/*.ts src/js/pyproxy.gen.ts src/js/error_handling.gen.ts
-	npx dts-bundle-generator src/js/pyodide.ts --export-referenced-types false
-	mv src/js/pyodide.d.ts dist
+dist/pyodide.d.ts dist/pyodide/ffi.d.ts: src/js/*.ts src/js/pyproxy.gen.ts src/js/error_handling.gen.ts
+	npx dts-bundle-generator src/js/{pyodide,ffi}.ts --export-referenced-types false --project src/js/tsconfig.json
+	mv src/js/{pyodide,ffi}.d.ts dist
+	python3 tools/fixup-type-definitions.py dist/pyodide.d.ts
+	python3 tools/fixup-type-definitions.py dist/ffi.d.ts
+	sed -i "s/export declare const ffi/declare const ffi/" dist/ffi.d.ts
```
ryanking13, line 143:
```suggestion
```

I guess this is already handled in `fixup-type-definitions.py`?

src/js/util.ts
```diff
@@ -0,0 +1,49 @@
...
+      return cache[n];
+    } else {
+      let result = fn(n);
+      cache[n] = result;
+      return result;
+    }
+  };
+};
+
+export function makeWarnOnce(warning: string) {
```
ryanking13, line 20:
```suggestion
/** @private */
export function makeWarnOnce(warning: string) {
```

Makefile
```diff
@@ -135,9 +135,14 @@ dist/package.json : src/js/package.json
 npm-link: dist/package.json
 	cd src/test-js && npm ci && npm link ../../dist
 
-dist/pyodide.d.ts: src/js/*.ts src/js/pyproxy.gen.ts src/js/error_handling.gen.ts
-	npx dts-bundle-generator src/js/pyodide.ts --export-referenced-types false
-	mv src/js/pyodide.d.ts dist
+dist/pyodide.d.ts dist/pyodide/ffi.d.ts: src/js/*.ts src/js/pyproxy.gen.ts src/js/error_handling.gen.ts
```
ryanking13, line 138: It seems like ffi.d.ts is not bundled when creating npm package: https://app.circleci.com/pipelines/github/hoodmane/pyodide/4659/workflows/3f881873-23be-422c-9100-99711b652e16/jobs/62596/parallel-runs/0/steps/0-104

ryanking13 (approved):
Thanks again @hoodmane! I have some minor comments + it would be nice to modify the PR title to describe what it does.

Makefile
hoodmane: Good catch, thanks!

hoodmane:
Thanks for the second review @ryanking13!

PR #3525, opened 2023-01-31
Unvendor pydoc_data and _pydecimal

ryanking13:
### Description

Removes / unvendors some python modules:

- Remove `_aix_support.py`, which is for supporting IBM AIX OS.

- Unvendor `_pydecimal.py`.
  - _pydecimal is a pure Python implementation of `decimal` module. [Importing `decimal` fallbacks](https://github.com/python/cpython/blob/main/Lib/decimal.py) to `_pydecimal` if the C-implementation `_decimal` is not available. In our case, _decimal is available, so _pydecimal will not be normally used. 

- Unvendor `pydoc_data`.
  - pydoc_data contains [a large (~700KB) dictionary](https://github.com/python/cpython/blob/main/Lib/pydoc_data/topics.py) for explaining python builtins. This is mostly used when `help("...")` is called.

### Checklists

- [x] Add / update tests


src/py/_pyodide/_importhook.py
```diff
@@ -135,7 +135,15 @@ def register_js_finder() -> None:
...
-UNVENDORED_STDLIBS = ["distutils", "ssl", "lzma", "sqlite3", "hashlib"]
+UNVENDORED_STDLIBS = [
+    "distutils",
+    "ssl",
+    "lzma",
+    "sqlite3",
+    "hashlib",
+    "pydoc_data",
+    "pydecimal",
+]
```
hoodmane, line 146: In the future, maybe we could switch to putting this information in `repodata.json`.

hoodmane (approved):
Thanks @ryanking13!

src/py/_pyodide/_importhook.py
ryanking13: Indeed, this should not be hardcoded inside pyodide-py, and putting this inside repodata.json would be nice.

ryanking13:
Thanks for the review!

PR #3528, opened 2023-02-01
Add package_type field to repodata.json

ryanking13:
### Description

This PR adds `package_type` field to repodata.json and use it to create a list of unvendored standard libraries. After this we don't need to manage a hard-coded list of unvendored stdlib lists in pyodide-py.

### Checklists

- [x] Add / update tests


hoodmane (approved):
Thanks @ryanking13!

PR #3532, opened 2023-02-01
Fix console.html url in docs

ryanking13:
@hoodmane I guess calling `calculate_pyodide_version` is omitted during the refactor in #3512.


hoodmane:
I think this has to be called after `apply_patches` or the rendering of the Python API docs will be messed up.

hoodmane (approved):
Thanks @ryanking13 !

PR #3533, opened 2023-02-02
DOC Fix version warning CSS

### LINKED ISSUE #3513: CSS broken in latest docs
## 🐛 Bug

image

It seems like the versionwarning banner is not displayed correctly in latest docs.

ryanking13:
Close #3513

- Drop [sphinx-version-warning](https://github.com/humitos/sphinx-version-warning) extension. It hasn't been updated for 3 years and it has a c[ritical bug that it requires a write permission to /usr/local/lib](https://github.com/humitos/sphinx-version-warning/issues/22)
- Show a version warning message in a announcement banner, by using a feature in sphinx-book-theme. Actually, I wanted to add a [header](https://sphinx-book-theme.readthedocs.io/en/stable/customize/header.html), but it seems like the sphinx-book-theme 0.4.0 is having some refactorings and adding `sections/header.html` doesn't work...

How it looks like:

image

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -37,7 +37,6 @@ extensions = [
     "autodocsumm",
     "sphinx_pyodide",
     "sphinx_argparse_cli",
-    "versionwarning.extension",
     "sphinx_issues",
     "sphinx_autodoc_typehints",
     "sphinx_design",  # Used for tabs in building-from-sources.md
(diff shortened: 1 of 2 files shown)


hoodmane:
Thanks @ryanking13! What creates the "you are viewing the docs from a PR" warning? Does readthedocs do that? Of course the urls for the PR docs are weird so people wouldn't see them by accident so they are less important.

rth (approved):
Thanks @ryanking13 ! I really had trouble with this extension in the past so indeed better to do it ourselves, particularly since it doesn't look too complicated.

> What creates the "you are viewing the docs from a PR" warning? Does readthedocs do that? 

Yeah, RDT is injecting things as part of their CI apparently.

Issue #3524, opened 2023-01-31
[Discussion] Unvendor deep-frozen stdlibs?


ryanking13:
Thanks to the hard work of Hood, we now run Python 3.11.

One big, but not very well-known change in CPython 3.11 is that it started to deep-freeze some of the standard libraries, i.e. compile python source codes into C and statically link them to libpython. According to cpython docs, a ["significant performance improvements to startup"](https://docs.python.org/3/whatsnew/changelog.html#id123) is achieved by this deep-freeze.

However, the trade-off is that it increases the bundle size, which is a crucial factor for us.

```sh
# 0.22.1 (Python 3.10.2)
7.2M pyodide.asm.wasm

# 0.23.0a1 (Python 3.11.1)
8.6M pyodide.asm.wasm

# Python 3.11.1 without freezing stdlibs except for importlib, zipimport
# (https://github.com/ryanking13/pyodide/tree/unfreeze)
7.9M pyodide.asm.wasm
```

So I would like to discuss some possible options:

1) No change.
2) Disable deep-freezing modules. This will reduce the size of pyodide.asm.wasm, but we cannot get the advantage of bootstrap optimization. This will also make it hard to do #3166.
3) Unvendor or remove python source codes of deep-frozen modules. This will reduce the size of pyodide.asm.data, so we can reduce the total download size increase. Users will not be affected (I guess...?) unless they try to access the source code of the frozen modules.

I would like to suggest option 3. Any ideas are welcome.

WDYT?

hoodmane:
Thanks for opening this issue @ryanking13. Option 3 sounds good to me. I would prefer the unvendor option so we could allow some `loadPyodide` setting to download all of these at startup. It might be nice if we could come up with some way to adjust the behavior of the `inspect` module so that it can tell people how to get the missing source code, though not if this is too much work.

rth:
> a ["significant performance improvements to startup"](https://docs.python.org/3/whatsnew/changelog.html#id123) is achieved by this deep-freeze.

I would say it would be good to benchmark to see to what extent it's true for us. In their benchmarks, they didn't have to compile WASM code. So while in general, more C code tends to be faster, in our case it might very well make things slower because of increased compile time. Hard to say without measuring it.

For the size I wouldn't say it's that critical, what matters IMO is the compressed size of pyodide.asm.wasm which was 2.53 MB for 0.22.1 and is 2.87 MB  for `main` so it's pretty small difference.  

[...]

PR #3539, opened 2023-02-03
Deepfreeze encodings module

ryanking13:
### Description

Deep-freeze stdlib `encodings`. After this, the size of `pyodide.asm.wasm` **increases** ~1.1MB.

Deep-freezing `encodings` is, hopefully, a final step to achive initializing Python interpreter without any python codes, which is a requirement for #3166. This was not done in upstream because of following reason:

```python
# For now we do not freeze the encodings, due # to the noise all
# those extra modules add to the text printed during the build.
# (See https://github.com/python/cpython/pull/28398#pullrequestreview-756856469.)
```

- Related: #3524 

### Checklists

- [x] Add / update tests

### DIFF
--- a/cpython/Makefile
+++ b/cpython/Makefile
@@ -102,6 +102,7 @@ $(BUILD)/$(LIB): $(BUILD)/Makefile $(BUILD)/pyconfig.h $(BUILD)/Modules/Setup.lo
 	cp Setup.local $(BUILD)/Modules/
 	( \
 		cd $(BUILD); \
+		make regen-frozen; \
 		emmake make CROSS_COMPILE=yes $(LIB) -j $${PYODIDE_JOBS:-3} \
 	)
 	touch $(BUILD)/$(LIB)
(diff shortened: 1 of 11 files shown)


cpython/Makefile
```diff
@@ -102,6 +102,7 @@ $(BUILD)/$(LIB): $(BUILD)/Makefile $(BUILD)/pyconfig.h $(BUILD)/Modules/Setup.lo
 	cp Setup.local $(BUILD)/Modules/
 	( \
 		cd $(BUILD); \
+		make regen-frozen; \
```
hoodmane, line 105: I had trouble with this before because it builds and runs an executable but it has already been configured to cross compile. But if this just works, that's convenient.

hoodmane (approved):
Thanks @ryanking13! The encodings module is quite big so I've wondered whether we could unvendor less often used parts of it...

hoodmane:
@rth since this increases the module size perhaps you have some comments.

rth:
Commented in https://github.com/pyodide/pyodide/issues/3524 Thanks for the ping and sorry for the slow response. Struggling a bit with keeping up with GH messages lately.

ryanking13:
> sorry for the slow response. Struggling a bit with keeping up with GH messages lately.

@rth It's okay. By the way, I saw that a massive protest is happening in Paris, I hope you are safe and sound.

rth:
Labor strikes are fairly common in France, with larger ones every couple of years, so nothing unusual :)

PR #3542, opened 2023-02-03
Disable NODEJS_CATCH_EXIT

hoodmane:
`NODEJS_CATCH_EXIT` alters global state every time loadPyodide is called, causing leak warnings if loadPyodide is run repeatedly. Also, we always catch all exceptions ourself so it is not needed.

`NODEJS_CATCH_REJECTION` alters the behavior of promises in node v14. 

These settings are both intended for applications, not for libraries.

See https://github.com/pyodide/pyodide/discussions/3541 and https://github.com/emscripten-core/emscripten/issues/18659


rth (approved):
Thanks @hoodmane !

PR #3545, opened 2023-02-06
Fix isPyProxy so that it works correctly with multiple Pyodide interpreters

hoodmane:
This actually just removes code so that we use the default `instanceof` behavior. The added test fatally fails without this patch.

### Checklists

- [x] Add / update tests


ryanking13 (approved):
Thanks @hoodmane!

PR #3546, opened 2023-02-06
Docs: Minor improvements

hoodmane:
I text wrapped a bunch of things, added a few extra paragraphs of explanation here and there, fixed some broken xrefs and added some new xrefs.


rth (approved):
Thanks!

PR #3547, opened 2023-02-06
NFC Rearrange api.ts

hoodmane:
This modestly simplifies api.ts by moving the public definitions into the `PyodideInterface` definition.
A followup to #3523.


ryanking13 (approved):
Thanks! There is a one test failing.

PR #3556, opened 2023-02-08
Add `replace_nbsp` parameter to Console.push

### LINKED ISSUE #3337: REPL Copy-paste with spaces doesn't work
In Firefox if one writes anything with spaces then tries to copy-paste the input to a standard Python REPL, one gets,
```
SyntaxError: invalid non-printable character U+00A0
```
this is because spaces are replaced by this character in Firefox as least.

In Pyodide REPL,
```py
>>> 1 + 1
```
Copy-pasting quoted in the host Python REPL,
```py
>>> '1 + 1'
'1\xa0+\xa01'
```
Likely due to https://github.com/jcubic/jquery.terminal/blob/b23a628283e12bd8996a3a2d8676cff6d1587a7a/__tests__/terminal.spec.js#L264 but unclear if it's due to a issue on they side (can't find anything in their repo) or a misconfiguration on our side. So might need to investigate more before reporting it upstream.

I'm pretty sure it worked OK before, so it might also be due to browser update, as currently even old Pyodide releases have this behavior.

ryanking13:
### Description

Resolve #3337

Add `replace_nbsp` parameter to Console.push, and use it in console.html.
If the parameter is set to True, replace all nbsp letters to regular space.

### Checklists

- [x] Add / update tests

### DIFF
--- a/src/py/pyodide/console.py
+++ b/src/py/pyodide/console.py
@@ -333,6 +333,10 @@ class Console:
         finally:
             self._streams_redirected = False
 
+    def _replace_nbsp(self, string: str) -> str:
+        string_ = string.replace("\u00A0", " ")
+        return string_
+
     def runsource(self, source: str, filename: str = "<console>") -> ConsoleFuture:
         """Compile and run source code in the interpreter."""
         res: ConsoleFuture | None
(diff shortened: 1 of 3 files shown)


hoodmane:
I think it would be better to make it a keyword only argument.

ryanking13:
> I think it would be better to make it a keyword only argument.

Sounds good to me. Can we pass keyword only argument in JS side?

hoodmane:
You have to use `push.callKwargs(s, {replace_nbsp: true});`. But are we sure this is better than `push(s.replaceAll("nbsp", " "))`?

ryanking13:
> But are we sure this is better than push(s.replaceAll("nbsp", " "))?

I think your idea is better.

ryanking13:
Closing in favor of #3558, thanks for the suggesion Hood!

PR #3544, opened 2023-02-06
Out-of-tree recipe build support

ryanking13:
### Description

Add out-of-tree support of `pyodide build-recipes`. Thanks to the awesome xbuildenv work by Hood, the required code change was quite small.

I am testing this branch here: https://github.com/ryanking13/pyodide-recipes-mirror

### Checklists

- [x] Add a [CHANGELOG](https://github.com/pyodide/pyodide/blob/main/docs/project/changelog.md) entry

PR #3559, opened 2023-02-09
ENH Autodetect the number of cores when invoking `pyodide build-recipes`

ryanking13:
### Description

When calling `pyoddie build-recipes`, set the number of parallel jobs to number of cpus in the system by default.

Split off from #3544 with some modification.

### Checklists

- [x] Add / update tests

### DIFF
--- a/docs/project/changelog.md
+++ b/docs/project/changelog.md
@@ -100,8 +100,8 @@ myst:
   `PYODIDE_PACKAGES="tag:core" make`.
   {pr}`3444`
 
-- {{ Enhancement }} `pyodide build-recipes` now autodetect number of
-  CPU cores in the system and use them for parallel builds.
+- {{ Enhancement }} `pyodide build-recipes` now autodetects the number of
+  CPU cores in the system and uses them for parallel builds.
   {pr}`3559`
 
 ### Pyodide CLI
--- a/docs/project/changelog.md
+++ b/docs/project/changelog.md
@@ -100,6 +100,10 @@ myst:
   `PYODIDE_PACKAGES="tag:core" make`.
   {pr}`3444`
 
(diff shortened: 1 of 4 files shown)


docs/project/changelog.md
```diff
@@ -100,6 +100,10 @@ myst:
   `PYODIDE_PACKAGES="tag:core" make`.
   {pr}`3444`
 
+- {{ Enhancement }} `pyodide build-recipes` now autodetect number of
+  CPU cores in the system and use them for parallel builds.
```
hoodmane, line 104:
```suggestion
- {{ Enhancement }} `pyodide build-recipes` now autodetects the number of
  CPU cores in the system and uses them for parallel builds.
```

hoodmane (approved):

rth:
The issue is that `os.cpu_count` is not really reliable. For instance in CI one often runs on CPU that have 16 or more cores, while only a few can be used (e.g. 4 in CircleCI). Then when starting 16 parallel build processes on a machine that can only use 4 or 2, it's actually quite counterproductive and will make the build slower. See https://github.com/joblib/loky/issues/111 for more context.

A better CPU detection taking into account different edge cases is done for instance in [loky](https://github.com/joblib/loky/blob/master/loky/backend/context.py#L71).

So I don't know we could either keep things as is or try to either vendor or use that as a dependency. It's not very critical, mostly FYI.

ryanking13:
Thanks for the information @rth! I think we can use it as a dependency unless it is too heavy.

rth:
It's not very heavy, in dependencies [it has cloudpickle and psutils](https://github.com/conda-forge/loky-feedstock/blob/main/recipe/meta.yaml#L23).

PR #3566, opened 2023-02-10
BLD Remove _pywasmcross entrypoint

ryanking13:
### Description

Remove `_pywasmcross` entrypoint that sometimes cause an issue when combined with Python version management tools like `pyenv`. (See [the comment here](https://github.com/pyodide/pyodide/pull/3544#discussion_r1096869651))

Instead, copy `pywasmcross.py` itself into the symlink directory and execute it.

### DIFF
--- a/.gitpod.yml
+++ b/.gitpod.yml
@@ -9,9 +9,8 @@ github:
     addLabel: false
 tasks:
   - name: Setup the environment
-    command: |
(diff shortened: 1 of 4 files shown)


hoodmane (approved):

hoodmane:
Thanks @ryanking13!

PR #3577, opened 2023-02-13
Remove globally pinned build requirements

ryanking13:
I think we don't need these anymore.


hoodmane:
That's nice. Thanks @ryanking13.

PR #3582, opened 2023-02-14
Download standard libraries in a zipfile

ryanking13:
### Description

This is a re-work of #3166. After this PR, the standard libraries are not vendored with `pyodide.asm.data`. Instead, it is now downloaded separately in a zip file similar to how we download pyodide_py.tar.

~~The hard part of this PR is to bootstrap Python without Python source codes. This requires deep-freezing all modules that are required during the bootstrap.~~

~~As mentioned in #3539, `encodings` module is needed during the bootstrap stage, but it is a large module (> 1.0MB) containing a bunch of encodings in the world, so I tried to freeze a subset of the encodings module that is used in the bootstrap (thanks for the idea from Hood), so the size increase of pyodide.asm.wasm is only 70KB.~~

~~However, the problem of freezing only part of the module is that the current Python import system does not support it. Therefore, I added some hacks to make it work, which is explained in a comment in pyodide.ts.~~

### File size change after this PR:

Before

```
- pyodide.asm.data:
    Original size: 4555 KB
    Gzip compressed size (level 1): 3062 KB
    Gzip compressed size (level 6): 2911 KB
    Gzip compressed size (level 9): 2911 KB
    Brotli compressed size: 2522 KB
- pyodide.asm.js:
    Original size: 1186 KB
    Gzip compressed size (level 1): 246 KB
    Gzip compressed size (level 6): 208 KB
    Gzip compressed size (level 9): 208 KB
    Brotli compressed size: 169 KB
- pyodide.asm.wasm:
    Original size: 8715 KB
    Gzip compressed size (level 1): 3142 KB
    Gzip compressed size (level 6): 2791 KB
    Gzip compressed size (level 9): 2791 KB
    Brotli compressed size: 2096 KB
```

After

```
- pyodide.asm.data:
    Original size: 4 KB
    Gzip compressed size (level 1): 0 KB
    Gzip compressed size (level 6): 0 KB
    Gzip compressed size (level 9): 0 KB
    Brotli compressed size: 0 KB
- pyodide.asm.js:
    Original size: 1086 KB
    Gzip compressed size (level 1): 213 KB
    Gzip compressed size (level 6): 177 KB
    Gzip compressed size (level 9): 177 KB
    Brotli compressed size: 145 KB
- pyodide.asm.wasm:
    Original size: 8775 KB
    Gzip compressed size (level 1): 3162 KB
    Gzip compressed size (level 6): 2806 KB
    Gzip compressed size (level 9): 2806 KB
    Brotli compressed size: 2105 KB
- python_stdlib.zip:
    Original size: 2230 KB
    Gzip compressed size (level 1): 2196 KB
    Gzip compressed size (level 6): 2194 KB
    Gzip compressed size (level 9): 2194 KB
    Brotli compressed size: 2181 KB
```

### Remaining works:

- remove pyodide.asm.data: the only remaining parts inside pyodide.asm.data is our [webbrowser.py](https://github.com/pyodide/pyodide/blob/main/src/py/lib/webbrowser.py) stub, I think we can either remove it or move it into pyodide_py.
- maybe merge pyodide_py.tar into the standard libraries zipfile.
- add option to use pycompiled standard libraries.

### Checklists

- [x] Add / update tests


rth (commented):
Thanks for working on this @ryanking13 ! It would be great indeed. 

Looks like some tests still need work. Overall the changes look reasonable to me, ping me once CI is good and you need a final review.

.circleci/config.yml
```diff
@@ -100,7 +100,7 @@ jobs:
           name: Zip build directory
           command: |
             tar cjf pyodide.tar.gz dist
-            tar cjf pyodide-core.tar.gz dist/pyodide{.js,.mjs,.asm.js,.asm.wasm,.asm.data,_py.tar} dist/{package,repodata}.json
+            tar cjf pyodide-core.tar.gz dist/pyodide{.js,.mjs,.asm.js,.asm.wasm,_py.tar} dist/{package,repodata}.json dist/pyodide_stdlib.zip
```
hoodmane, line 103: Maybe python_stdlib?

ryanking13: Sounds good.

ryanking13:
@rth I think this is now ready to be reviewed now.

1. I reverted changes in #3585. I found that freezing encodings module (with a bunch of hacks) doesn't work well in some non-trivial cases like invoking subinterpreters, so I decided not to use it unless CPython adds support for partially freezing modules.

2. Instead, I changed to install standard libraries before bootstrapping Python interpreter, by adding it to Module.preRun. This makes bootstrapping a lot simpler because I was able to remove all sys.path and sys.modules hacks. Note that Emscripten installs .data files in the same way.

There are some things left to do, but I would like to do them in follow-up PRs, since this PR is already quite big enough.

hoodmane:
For a followup, we could deep freeze it under a different name "encodings_minimal" and patch the interpreter to use this other module on startup.

src/tests/test_stdlib_fixes.py
```diff
@@ -237,3 +237,60 @@ def test_encodings_deepfrozen(selenium):
     for enc in all_encodings:
         codecs.getencoder(enc)
         codecs.getdecoder(enc)
+
+
+@run_in_pyodide
+def test_zipimport_traceback(selenium):
+    import json.decoder
```
rth, line 244: Maybe we could add a short comment about why this test is needed. Took me some time to remember why we do this )

rth (approved):
A few minor comments otherwise LGTM. 

I'm personally very excited by this PR,  thanks again @ryanking13 !

.circleci/config.yml
```diff
@@ -354,14 +354,14 @@ jobs:
...
-            git checkout 164054a9c6fbd2176f386b6552ed8d079c6bcc04
-            ./build.sh
+      # - run:
+      #     name: check if Pyodide works with Webpack
+      #     command: |
+      #       git clone https://github.com/pyodide/pyodide-webpack-example.git
+      #       export DEV_PYODIDE_PATH=`realpath dist`
+      #       cd pyodide-webpack-example
+      #       git checkout 164054a9c6fbd2176f386b6552ed8d079c6bcc04
+      #       ./build.sh
```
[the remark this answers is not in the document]
ryanking13: Thanks for opening the issue, I added a comment.

ryanking13:
> For a followup, we could deep freeze it under a different name "encodings_minimal" and patch the interpreter to use this other module on startup.

src/js/module.ts
```diff
@@ -66,6 +77,19 @@ export function createModule(): any {
...
+/**
+ * @private
+ */
+export function saveStreamToFile(
+  Module: Module,
+  stream: Uint8Array,
+  fileName: string,
+) {
+  const f = Module.FS.open(fileName, "w");
+  Module.FS.write(f, stream, 0, stream.byteLength);
```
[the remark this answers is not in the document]
ryanking13: Good point, thanks!

PR #3584, opened 2023-02-15
BLD Add DISTDIR variable

ryanking13:
### Description

Adds `DISTDIR` env variable that points to the directory where libraries should be copied to when building shared libraries with recipe.

### Checklists


hoodmane (approved):
Thanks @ryanking13!

ryanking13:
Thanks for the review!

PR #3585, opened 2023-02-15
Deepfreeze essential parts of the encodings module

ryanking13:
### Description

This is a split off from #3582.

Deepfreeze essential parts of encodings module (encodings.\__init__, encodings.utf_8, and encodings.aliases), which is required during the Python bootstrap.


src/js/pyodide.ts
```diff
@@ -65,6 +65,33 @@ function wrapPythonGlobals(globals_dict: PyDict, builtins_dict: PyDict) {
...
+ * @private
+ */
+function postInitializePython(Module: any) {
+  const code = `
+import sys, _imp, encodings, codecs
+codecs.unregister(encodings.search_function)
+del sys.modules["encodings"]
+del sys.modules["encodings.utf_8"]
+del sys.modules["encodings.aliases"]
+_imp._override_frozen_modules_for_tests(-1)
```
hoodmane, line 81: Shouldn't you set this back to 0 afterwards?

ryanking13: I think it is not necessary. All frozen modules are already imported and saved in `sys.modules`, so we won't get extra performance advantages from setting the value back.

hoodmane (approved):

hoodmane:
Thanks so much for working on this @ryanking13, it's a really great feature.

ryanking13:
Thanks for the review!

PR #3586, opened 2023-02-15
Fix activating pyodide venv command docs

messense:
### Checklists


ryanking13 (approved):
Thanks @messense!

PR #3591, opened 2023-02-16
MAINT Initialize filesystem in a single point

ryanking13:
### Description

We do several filesystem operations before loading Python.

- Mount _node_mounts
- Create default directories
- Register NativeFS file system

This PR organizes all those operations into a single function.

This is an another split off from #3582.


hoodmane (approved):
Thanks!

PR #3592, opened 2023-02-17
Add url parameter to install_xbuildenv

### LINKED ISSUE #3567: Option to install arbitrary version of xbuildenv for debug purpose
When testing out-of-tree builds with tot pyodide-build, it is sometimes annoying to create a cross-build environment.

So I think it would be nice to have the option to install an arbitrary version of cross-build env, such as:

```
pyodide install-xbuildenv --version x.x.x
```

This should be a private entry point only for maintainers/contributors, since installing an incompatible version of cross-build env will definitely break the build. Those who use this entry point should know what they are doing.

ryanking13:
### Description

Resolve #3567 

Adds url paramter to `install_xbuildenv`, which helps installing xbuildenv from arbitrary URL. This is mostly for developers / maintainers who want to test tot xbuildenv.

This PR also adds a private CLI entrypoint `pyodide xbuildenv install`.

### Checklists

- [x] Add / update tests

### DIFF
--- a/pyodide-build/pyodide_build/cli/xbuildenv.py
+++ b/pyodide-build/pyodide_build/cli/xbuildenv.py
@@ -1,0 +1,32 @@
+from pathlib import Path
+
+import typer
+
+from ..install_xbuildenv import install
+
+app = typer.Typer(hidden=True)
+
+
+@app.callback()
+def callback():
+    """
+    Create or install cross build environment
+    """
+
+
+@app.command("install")  # type: ignore[misc]
+def _install(
+    path: Path = typer.Option(".pyodide-xbuildenv", help="path to xbuildenv directory"),
+    download: bool = typer.Option(False, help="download xbuildenv before installing"),
+    url: str = typer.Option(None, help="URL to download xbuildenv from"),
+) -> None:
+    """
+    Install xbuildenv.
+
+    The isntalled environment is the same as the one that would result from
+    `PYODIDE_PACKAGES='scipy' make` except that it is much faster.
+    The goal is to enable out-of-tree builds for binary packages that depend
+    on numpy or scipy.
+    Note: this is a private endpoint that should not be used outside of the Pyodide Makefile.
+    """
+    install(path, download=download, url=url)
--- a/pyodide-build/pyodide_build/install_xbuildenv.py
+++ b/pyodide-build/pyodide_build/install_xbuildenv.py
@@ -90,6 +90,25 @@ def install_xbuildenv(version: str, xbuildenv_path: Path) -> None:
 
 
 def install(path: Path, *, download: bool = False, url: str | None = None) -> None:
+    """
+    Install cross-build environment.
+
+    Parameters
+    ----------
+    path
+        A path to the cross-build environment.
+    download
+        Whether to download the cross-build environment before installing it.
+    url
+        URL to download the cross-build environment from. This is only used
+        if `download` is True. The URL should point to a tarball containing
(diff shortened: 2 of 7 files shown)


pyodide-build/pyodide_build/cli/xbuildenv.py
```diff
@@ -0,0 +1,32 @@
...
+
+@app.callback()
+def callback():
+    """
+    Create or install cross build environment
+    """
+
+
+@app.command("install")  # type: ignore[misc]
+def _install(
```
hoodmane, line 18: Is this hidden from `pyodide --help` or just the docs?

ryanking13: It is hidden from both of help message and docs.

pyodide-build/pyodide_build/tests/test_cli.py
```diff
@@ -328,3 +335,24 @@ def test_create_zipfile_compile(temp_python_lib, tmp_path):
...
+    envpath = Path(tmp_path) / ".xbuildenv"
+    result = runner.invoke(
+        xbuildenv.app,
+        [
+            "install",
+            "--path",
+            str(envpath),
+            "--download",
+            "--url",
+            "https://github.com/pyodide/pyodide/releases/download/0.22.1/xbuildenv-0.22.1.tar.bz2",
```
rth, line 350: It's a 10MB file, not sure about downloading it each time this test is run. At least maybe we could put it behind some pytest mark and not run it by default?

If we cannot create a similar archive for testing at runtime that is.

pyodide-build/pyodide_build/install_xbuildenv.py
```diff
@@ -79,11 +89,14 @@ def install_xbuildenv(version: str, xbuildenv_path: Path) -> None:
     create_pypa_index(repodata["packages"], xbuildenv_root, cdn_base)
 
 
-def main(args: argparse.Namespace) -> None:
+def install(path: Path, *, download: bool = False, url: str | None = None) -> None:
```
rth, line 92: Some doctest explaining what url is supposed to contain wouldn't hurt,

pyodide-build/pyodide_build/install_xbuildenv.py
```diff
@@ -20,32 +21,41 @@ def make_parser(parser: argparse.ArgumentParser) -> argparse.ArgumentParser:
         "Note: this is a private endpoint that should not be used outside of the Pyodide Makefile."
     )
     parser.add_argument("--download", action="store_true", help="Download xbuild env")
+    parser.add_argument("--url", help="URL to download xbuild env from", default=None)
     parser.add_argument("xbuildenv", type=str, nargs=1)
     return parser
```
rth, line 26: BTW, Do you know why we still have this argparse code and not using typer here? Do we still need it?

rth (approved):

pyodide-build/pyodide_build/tests/test_cli.py
ryanking13: > If we cannot create a similar archive for testing at runtime that is.

I'll try this approach.

pyodide-build/pyodide_build/install_xbuildenv.py
ryanking13: I am trying to keep `pyodide_build` entrypoint and `pyodide` entrypoint compatible until we completely drop `pyodide_build` entrypoint. Maybe we should decide when we will drop `pyodide_build` entrypoint.

PR #3598, opened 2023-02-21
Use loky.cpu_count() instead of os.cpu_count()

ryanking13:
### Description

As discussed in https://github.com/pyodide/pyodide/pull/3559#issuecomment-1435965170

### Checklists

- [x] Add / update tests


rth (approved):
One of the linting pre-commit jobs fails but it's weird that the other one passes.

Otherwise LGTM, thanks!

PR #3602, opened 2023-02-22
CI Add MacOS build

ryanking13:
### Description

Adds CI for MacOS, continuing @rth 's work in #1912.

I added a GNU sed / patch check, which kills the build if GNU sed / patch is not found. At first, I tried to make all sed/patch commands GNU / BSD compatible, (#1770) but I realized that it was pretty annoying than I thought.

Maybe in the future, we can try making all sed/patch commands in our build script GNU / BSD compatible...

Close #1912
Close #590


.github/workflows/main.yml
```diff
@@ -2,7 +2,6 @@ name: main
 
 on:
   push:
-    branches: [main]
```
cclauss, line 5: I believe that this will create double runs CI for contributors who have write access to this repository.
I usually use the following syntax to limit double runs.
```
on:
  push:
    branches: [main]
  pull_request:
    branches: [main]
```

ryanking13: @cclauss Thanks for the suggestion. Actually, I want to trigger the CI when I push to my fork in non-main branch so I can run CI before making a pull request. Do you have a suggestion for that setting?

ryanking13: @cclauss Thanks for the suggestion. Actually, I want to trigger the CI when I push to my fork in non-main branch so I can run CI before making a pull request. Do you have a suggestion for that setting?

cclauss: OK.  Now I understand the use case.  Your modification will probably work OK for that.

ryanking13:
@rth Could you review this when you have time? Mostly I just adapted changes in your draft PR #1912.

rth:
Superseded by https://github.com/pyodide/pyodide/pull/1912

tools/dependency-check.sh
```diff
@@ -40,8 +40,32 @@ check_cmake() {
   check_binary_present "cmake"
 }
 
+check_sed() {
+  check_binary_present "sed"
+  gnu_sed_found=$(sed --help | grep -q gnu.org)
+
+  if [ "${gnu_sed_found}" ]; then
+    echo >&2 "Pyodide requires GNU sed."
+    echo >&2 "If you are on macOS you can install it with 'brew install gnu-sed' and then add it to your PATH."
```
rth, line 49: Nice!

rth (approved):
Thanks @ryanking13 ! 

Sorry, I closed it by mistake. Merging since the CI was green before I retriggered it.

rth:
It would have been nice to have the MacOS arm64 build, but it doesn't looks like it would happen very soon for GHA https://github.com/github/roadmap/issues/528

PR #3606, opened 2023-02-24
Fix mypy issue in pre-commit

ryanking13:
Trying to make mypy happy

### DIFF
--- a/pyodide-build/pyodide_build/cli/build.py
+++ b/pyodide-build/pyodide_build/cli/build.py
@@ -67,37 +67,40 @@ def url(
     backend_flags = ctx.args
     curdir = Path.cwd()
     (curdir / "dist").mkdir(exist_ok=True)
-    with requests.get(package_url, stream=True) as response:
-        parsed_url = urlparse(response.url)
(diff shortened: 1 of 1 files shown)


pyodide-build/pyodide_build/cli/build.py
```diff
@@ -67,37 +67,40 @@ def url(
     backend_flags = ctx.args
     curdir = Path.cwd()
     (curdir / "dist").mkdir(exist_ok=True)
-    with requests.get(package_url, stream=True) as response:
```
cclauss, line 70:
```
        response = cast(requests.Response, response)  # placate mypy
```

rth: yes, I agree it's probably better and pre-commit passes in https://github.com/pyodide/pyodide/pull/3609

So maybe let's go with that, if no objections from @ryanking13 ?

ryanking13: Indeed, that's better. Thanks @cclauss!

PR #3607, opened 2023-02-24
FIX Change rust packages not to build concurrently

### LINKED ISSUE #3565: Issue in building rust-based packages in parallel
We are having some issue when building multiple rust-based packages in parallel.

For example, see: https://app.circleci.com/pipelines/github/pyodide/pyodide/5328/workflows/a0ced892-ac1a-4519-984c-d98a84fabe29/jobs/66900

It is expected behavior since `rustup` was not designed for concurrent use (https://github.com/rust-lang/rustup/issues/988).

Some possible workarounds that I can think of are:

1. Add some logic to not build rust-baed packages in parallel
2. Create a fresh local rust build environment everytime when building rust-based package. I found a Python package called [rustenv](https://github.com/chriskuehl/rustenv), which is designed for that purpose. It is a ~150LoC script, which basically installs rust every time from `https://sh.rustup.rs` into a local directory.

I slightly prefer option 1, as we don't have to be responsible for installing build requirements. But I have no strong opinion.

ryanking13:
### Description

Resolve #3565

### Checklists

- [x] Add / update tests


pyodide-build/pyodide_build/buildall.py
```diff
@@ -538,14 +545,25 @@ def build_from_graph(
...
-            pkg = build_queue.get()[1]
+            _, pkg = build_queue.get()
 
             with thread_lock:
+                if is_rust_package(pkg):
+                    # Don't build multiple rust packages at the same time
+                    # See: https://github.com/pyodide/pyodide/issues/3565
+                    if building_rust_pkg:
+                        build_queue.put((job_priority(pkg), pkg))
+                        sleep(0.01)
```
hoodmane, line 562: What's this sleep do? It needs a comment I think.

ryanking13: It releases the GIL so that main thread can queue new packages, following the same step done in [here](https://github.com/pyodide/pyodide/blob/205deef3ccc38b32b510e2854bf12f73e9b332c8/pyodide-build/pyodide_build/buildall.py#L572-L573), will add a comment.

hoodmane:
Also, if you just have two rust packages, this will like spin continuously with the second package? Seems a bit tricky to make it behave reasonably. But maybe it just doesn't matter?

ryanking13:
> Also, if you just have two rust packages, this will like spin continuously with the second package? Seems a bit tricky to make it behave reasonably. But maybe it just doesn't matter?

Yes, one thread will keep pushing and popping the second package from queue until the first one finishes. I hope it doesn't matter, but when we find it matters, I guess we can add an additional lock for rust packages or incrementally increase the sleep time.

hoodmane:
Maybe add a comment indicating that this is what happens when only rust classes are left and we don't exactly love it. Also let's increase the wait time to 50 or 100 ms.

ryanking13:
> Maybe add a comment indicating that this is what happens when only rust classes are left and we don't exactly love it. Also let's increase the wait time to 50 or 100 ms.

Sounds good. Done.

pyodide-build/pyodide_build/buildall.py
```diff
@@ -538,14 +545,30 @@ def build_from_graph(
...
         while True:
-            pkg = build_queue.get()[1]
+            _, pkg = build_queue.get()
 
             with thread_lock:
+                if is_rust_package(pkg):
+                    # Don't build multiple rust packages at the same time.
+                    # See: https://github.com/pyodide/pyodide/issues/3565
+                    # Note that if there are only rust packages left in the queue,
+                    # this will keep push and pop package until the current rust package
```
hoodmane, line 561:
```suggestion
                    # this will keep pushing and popping packages until the current rust package
```

hoodmane (approved):

ryanking13:
Thanks for the review!

pyodide-build/pyodide_build/buildall.py
```diff
@@ -538,14 +545,30 @@ def build_from_graph(
...
+                    # Don't build multiple rust packages at the same time.
+                    # See: https://github.com/pyodide/pyodide/issues/3565
+                    # Note that if there are only rust packages left in the queue,
+                    # this will keep pushing and popping packages until the current rust package
+                    # is built. This is not ideal but presumably the overhead is negligible.
+                    if building_rust_pkg:
+                        build_queue.put((job_priority(pkg), pkg))
+
+                        # Release the GIL so new packages get queued
+                        sleep(0.1)
```
hoodmane, line 567: Maybe we could try releasing the lock before performing this sleep? Since Python threads are somewhere between threads and coroutines and the only yield point is when the lock is held this may have something to do with the problem. Maybe also put one or more unconditional `sleep(0)` with the lock unacquired.

PR #3608, opened 2023-02-24
Upgrade GitHub Actions

cclauss:
### Description

The GitHub Actions read `$PYVERSION` from the file `Makefile.envs` so the tests run on the same version of Python.
* `PYVERSION=$(git grep 'export PYVERSION ?=' Makefile.envs | cut -d' ' -f4)`

### Checklists

- [x] Add / update tests

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -10,22 +10,36 @@ concurrency:
   cancel-in-progress: true
 
 jobs:
+  get_python_version:
+    runs-on: ubuntu-latest
+    outputs:
+      PYVERSION: ${{ steps.get_python_version.outputs.PYVERSION }}
+    steps:
+      - uses: actions/checkout@v3
+      - name: Read the Python version from Makefile.envs
+        id: get_python_version
+        run: |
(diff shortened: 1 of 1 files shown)


.github/workflows/main.yml
```diff
@@ -13,19 +13,19 @@ jobs:
...
   test-docs:
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v2
-      - uses: actions/setup-python@v3
+      - uses: actions/checkout@v3
+      - uses: actions/setup-python@v4
         with:
-          python-version: 3.11.1
+          python-version: 3.11.2
```
hoodmane, line 28: We should probably update this at the same time that we update the Pyodide version of Python to 3.11.2. Though it's not super important.

cclauss: It would be helpful if the pyodide build process used a `python-version` file which `actions/setup-Python` can also read.
* https://github.com/actions/setup-python/blob/main/docs/advanced-usage.md#using-the-python-version-file-input

ryanking13: I think we would like to keep all build configs in a single file.

cclauss: Read the Python version from Makefile.envs

ryanking13 (approved):
Thanks!

PR #3611, opened 2023-02-26
draft:remove pyo3_config

joemarshall:
I'm just pushing this PR to see if removing pyo3_config breaks the full build in CI


cclauss:
Please rebase to fix the linter error.

joemarshall:
Okay, so I worked out the problem - old versions of pyo3 don't build nicely. I think this is probably not worth doing until cryptography and bcrypt both use new pyo3 (which was non-trivial). I'll close this for now. 

Does mean that packages using maturin < 0.14.14 don't work, but it is trivial to patch them up.

Issue #3612, opened 2023-02-27
mypyc is failing to build in CI


ryanking13:
See: https://github.com/python/mypy/pull/14788

PR #3613, opened 2023-02-28
Pin types-setuptools version 

ryanking13:
Close: #3612

A temporary fix until mypy releases a new version.

### DIFF
--- a/pyodide-build/pyodide_build/pypabuild.py
+++ b/pyodide-build/pyodide_build/pypabuild.py
@@ -75,6 +75,16 @@ def install_reqs(env: IsolatedEnv, reqs: set[str]) -> None:
         )
     )
 
+    pinned_reqs = {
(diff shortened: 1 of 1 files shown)


rth:
Thanks!

rth:
Also, no idea why git clone started to fail in CircleCI on main. Possibly related to https://support.circleci.com/hc/en-us/articles/4410498350107-Build-Fails-due-to-SSH-Handshake so maybe we could try updating the docker image.

ryanking13:
> Also, no idea why git clone started to fail in CircleCI on main. Possibly related to https://support.circleci.com/hc/en-us/articles/4410498350107-Build-Fails-due-to-SSH-Handshake so maybe we could try updating the docker image.

Hmm, but it works on forks so I'm not sure if our Docker image is a problem. We can try updating the docker image anyways.

PR #3616, opened 2023-03-01
Add FAQ, Docs for working around JS Objects named like Python Keywords

### LINKED ISSUE #3615: `Array.from` in Python Causes Syntax Error
## 🐛 Bug

The [Array.from()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/from) static method (alternative constructor) throws a syntax error when used in Python.

### To Reproduce

The following code throws a SyntaxError: 

```python
<script>
  async function main() {
    pyodide.runPython(`
      from js import Array
      a = Array.from([1,2,3])
      print(a)
      `);
  };
  main();
</script>
```
```
...
 File "/lib/python3.10/_pyodide/_base.py", line 144, in _parse_and_compile_gen
 mod = compile(source, filename, mode, flags | ast.PyCF_ONLY_AST)
  File "<exec>", line 3
    a = Array.from([1,2,3])
              ^^^^
SyntaxError: invalid syntax
```

However, the following code works:
```python
<script>
  async function main() {
    const generator = pyodide.runPython(`
      from js import Array
      fromFunc = getattr(Array, 'from')
      b = fromFunc([4,5,6])
      print(b)
      `);
  };
  main();
</script>
```

### Expected behavior

This behavior is, in a way, consistent with Python, where directly setting to or accessing the `from` attribute of an object is also a syntax error, but you can create one with `setattr` and `getattr` (or similar):

```python
class A():
        ...

#SyntaxError
A.from = lambda: print("Whoops") 

#Works
setattr(A, 'from', lambda : print("from() was called!"))
getattr(A, 'from')()
```

But it's unfortunate that the simple Array.from() method is made more cumbersome to use by the Python keyword. As for the solution? I don't know - could be something like detecting properties on JS objects that are Python keywords, and creating an additional property for them? (`_from`?). Or maybe just highlighting the issue of JS objects with properties that are Python keywords is enough.

Forgive me if that is already documented somewhere, I couldn't find a specific reference to it.

### Environment

- Pyodide Version 0.22.1
- Browser version Chrome Version 110.0.5481.178 (Official Build) (64-bit)

JeffersGlass:
### Description

In response to #3615, this PR adds a new section to the FAQ and the Type Conversions page, with strategies for dealing with JavaScript objects whose names or properties are the same as Python keywords.  

For instance, to pull an example from this PR's FAQ.md:

```js
people = {global: "lots and lots"};

//raises SyntaxError, since 'global' is a Python reserved keyword
pyodide.runPython(`from js import people; print(people.global)`);

//Using setattr() and getattr() to access the 'global' attribute by name works:
pyodide.runPython(`
    from js import people
    setattr(people, 'global', 'even more')
    print(getattr(people, 'global'))
    `);
```

### Checklists

_None of the checklist items apply - this is purely additions to the documentation._

Closes #3615.

### DIFF
--- a/docs/usage/faq.md
+++ b/docs/usage/faq.md
@@ -423,3 +423,48 @@ So in order to persist changes, you have to call
 [`pyodide.FS.syncfs()`](https://emscripten.org/docs/api_reference/Filesystem-API.html#FS.syncfs).
 See [Emscripten File System API](https://emscripten.org/docs/api_reference/Filesystem-API.html#persistent-data)
 for more details.
+
+## How can I access JavaScript objects/attributes in Python if their names are Python keywords?
+
+Some JavaScript objects may have names or attributes which are also [Python Keywords](https://docs.python.org/3/reference/lexical_analysis.html#keywords), making them difficult to interact with when importing them into Python. For example, all three of the following uses of `runPython` will throw a SyntaxError:
+
+```pyodide
+//The built-in method Array.from() overlaps with Python's "from"
+pyodide.runPython(`from js import Array; print(Array.from([1,2,3]))`);
+
+//"global" is a valid attribute name in JS, but a reserved keyword in Python
+people = {global: "lots and lots"};
+pyodide.runPython(`from js import people; print(people.global)`);
+
+//"lambda" is a valid object name in JS, but a reserved keyword in Python
+lambda = (x) => {return x + 1};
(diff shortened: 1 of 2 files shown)


ryanking13 (approved):
Thanks, @JeffersGlass!

hoodmane:
Thanks! We really appreciate all updates to the faq.

PR #3617, opened 2023-03-01
Handling for reserved Python keywords as JsProxy attrs

### LINKED ISSUE #3615: `Array.from` in Python Causes Syntax Error

hoodmane:
Resolves #3615. Things like `from` and `finally` are reserved keywords in Python and so `a.finally` 
is a `SyntaxError`. This automatically reroutes `a.from_` on a `JsProxy` to refer to `a.from` so it
can be used reasonably conveniently from Python.

### Checklists

- [x] Add / update tests


rth (approved):
Thanks @hoodmane !

(Needs a changelog)

hoodmane:
Thanks for the review. Worth noting that one bad aspect of this is that:
```py
for x in dir(p):
  delattr(p,x)
```
will fail. I see not much reason to ever do that. To help with tab complete and to fix this very obscure detail, we may want to only include e.g. `from_` and not include `from` in `dir` so `o.fr<tab>` completes to `o.from_` which works instead of `o.from` which does not. This also keeps these attributes from getting double counted in `dir`.

src/core/hiwire.c
```diff
@@ -908,16 +908,43 @@ EM_JS_REF(JsRef, JsObject_New, (), {
...
   let jsobj = Hiwire.get_value(idobj);
   let jskey = UTF8ToString(ptrkey);
   let result = jsobj[jskey];
   // clang-format off
-  if (result === undefined && !(jskey in jsobj)) {
-    // clang-format on
+  if (result !== undefined || jskey in jsobj) {
+    return Hiwire.new_value(result);
+  }
+  if(!jskey.endsWith("_")) {
```
ryanking13, line 932: Some comments here about what we are doing wouldn't hurt.

ryanking13 (approved):
Thanks!

hoodmane:
Okay I changed the logic of this since it was last reviewed so I would appreciate one more review.

src/core/hiwire.c
```diff
@@ -908,33 +908,82 @@ EM_JS_REF(JsRef, JsObject_New, (), {
...
+ */
+EM_JS(int, normalizeReservedWords, (int action, int word), {
+  // clang-format off
+  // 1. if word is not a reserved word followed by 0 or more underscores, return
+  //    it unchanged.
+  const noTrailing_ = word.replace(/_*$/, "");
+  if (!isReservedWord(noTrailing_)) {
+    return word;
+  }
+  // 2. If there is at lease one trailing underscore, return the word with a
```
ryanking13, line 948:
```suggestion
  // 2. If there is at least one trailing underscore, return the word with a
```

ryanking13 (approved):
Great work, thanks @hoodmane!

PR #3621, opened 2023-03-02
Merge pyodide_py into python_stdlib.zip

ryanking13:
### Description

Merges pyodide_py.tar into python_stdlib.zip. This reduces the number of network requests during the initialization.

The original intention of splitting pyodide_py.tar from pyodide.asm.data was to give users an option to provide custom files (#2023), which is now available by modifying python_stdlib.zip directly. So I think there is no reason to have both python_stdlib.zip and pyodide_py.tar.

### Checklists

- [x] Add / update tests


hoodmane:
To be more specific, my main intention was to make the build step faster when I modify Python code.

pyodide-build/pyodide_build/pyzip.py
```diff
@@ -5,7 +5,7 @@
 
 from ._py_compile import _compile
 
-# This files are removed from the stdlib by default
+# These files are removed from the stdlib by default
```
hoodmane, line 8: I'd also remove "by default" here and in the second comment since it's unclear what it means.
```suggestion
# These files are removed from the stdlib
```

pyodide-build/pyodide_build/pyzip.py
```diff
@@ -22,10 +22,9 @@
     "tkinter/",
     "turtle.py",
     "turtledemo",
-    "webbrowser.py",
 )
 
-# This files are unvendored from the stdlib by default
+# These files are unvendored from the stdlib by default
```
hoodmane, line 27:
```suggestion
# These files are unvendored from the stdlib and can be loaded with `loadPackage`
```
Or what is the difference between unvendored and removed?

hoodmane (approved):

pyodide-build/pyodide_build/pyzip.py
```diff
@@ -49,22 +51,51 @@ def default_filterfunc(
...
+
+        if path.is_file() and name in (
+            "LICENSE",
+            "LICENSE.txt",
+            "setup.py",
+            ".gitignore",
+        ):
+            return True
+
+        if path.is_file() and name.endswith(("pyi", "toml", "cfg", "md", "rst")):
```
rth, line 72: This might be a bit too restrictive, but we will see if people complain.

rth (approved):
Thanks!

hoodmane:
Thanks again @ryanking13! This is great work.

PR #3624, opened 2023-03-02
Ruff is now compatible with Python pattern matching

cclauss:
### Description

As discussed at https://github.com/pyodide/pyodide/pull/3522#issuecomment-1450003295 Ruff is now compatible with Python pattern matching so we can add a few more Ruff tests.

### Checklists

- [x] Add / update tests


henryiii:
You have to bump the ruff version, it's too old here: https://github.com/pyodide/pyodide/blob/79aa563f2c199fb69f962d3d066abbdedfaeb9bc/.pre-commit-config.yaml#L20 (I think it was 0.250 or 0.251 that added pattern matching)

cclauss:
Yes, @henryiii thanks.  I got distracted on the other so now I have created  #3650 to take its place.

PR #3625, opened 2023-03-03
Add private APIs to get Python version before bootstrapping Python

ryanking13:
### Description

Add private APIs (module.\_py_version_{major,minor,micro}) to get the Python version before bootstrapping Python, and use it in Module.preRun to initialize filesystem.

### Checklists

- [x] Add / update tests


hoodmane (approved):
Nice work. This approach to dealing with the version never occurred to me before...

Issue #2045, opened 2021-12-16
RFC Plans for Pyodide package loading system


hoodmane:
# Plans for Pyodide Package Loading system

This supersedes #1470. #1940 is also related.

## Different ways to do dependency resolution
In #1470, I asked the question: how do we explain the difference between `pyodide.loadPackage` and `micropip.install`? Furthermore the point was raised in #1940 that added yet another way to install packages on top of the existing two would add to existing confusion about the two existing package loading methods.

I now have a good explanation for what the distinction should be:
1. `pyodide.unpackArchive` does no dependency resolution. It can be used to load a wheel which may just be missing its dependencies. This is good for when users want to do manual dependency resolution.
2. `pyodide.loadPackage` does static dependency resolution. It only can load packages that we thought we might want to load ahead of time. It already has precomputed which dependency tree it plans to load so there's no need to go mucking about with complicated wheel operations.
3. `micropip` is for dynamic dependency resolution. This should only be used to load packages we didn't know ahead of time we would need. It should not be used in apps that always use the same set of packages. Good for a REPL or any place where you are evaluating arbitrary user code.

### New `pyodide_build` entrypoints to update `packages.json`

We should be able to add packages to `packages.json` in some simple way. We could do dependency resolution for packages with suitable wheels available on PyPi and add entries to `packages.json` pointing at the appropriate pypi wheels. We should be able to generate these from a requirements file.

## Switching to wheels
In #2027 I have rewritten the Pyodide package management system to build and load wheels. This will increase compatibility between `pyodide.loadPackage` and `micropip.install`. In principle, `pyodide.loadPackage` can be used to load other archives in addition to wheels. In the future, I am hoping not to use the Emscripten file_packager format for anything except the Python standard library which we need before we can start using `shutil`. Once `shutil` works, the rest can be done with normal archive formats.

### Remove support for loading from urls
In order to get a clean distinction between the purposes of our different package loading methods, we should remove loading urls with `pyodide.loadPackage`. We have no way to do any dependency resolution on them. We will have `pyodide.unpackArchive` or a related API for loading packages without dependency resolution. Instead, `loadPackage` will only be able to load named packages that appear in `packages.json`.

### Add support for loading wheels from pypi or other locations

Since `pyodide.loadPackage` loads wheels now, there is no reason it can't load compatible wheels from PyPi. The appropriate URL should be precomputed when we create `packages.json`.

### Logging levels

[...]

PR #3628, opened 2023-03-03
Improve pyodide.loadPackage docs

rth:
Improve `pyodide.loadPackage` documentation which was a bit outdated.

Though in https://github.com/pyodide/pyodide/issues/2045 there is a proposal to,
> Remove support for loading from URLs in pyodide.loadPackage

while this goes in a bit different direction of documenting the current behavior.

### DIFF
--- a/src/js/load-package.ts
+++ b/src/js/load-package.ts
@@ -360,7 +360,7 @@ const cbDeprecationWarnOnce = makeWarnOnce(
 );
 
 /**
- * Load packages from the Pyodide distribution or by URL.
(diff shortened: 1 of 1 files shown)


hoodmane:
> Remove support for loading from URLs in pyodide.loadPackage

I am pretty sure we are not going to do this.

rth:
Agreed.

hoodmane:
Someone needs to filter through that and locate the ideas that we should still discuss or implement and make a new issue with just these action items. But it takes a while to get around to these things...

ryanking13 (approved):
Thanks @rth!

PR #3636, opened 2023-03-07
DOC Remove package loading from roadmap

hoodmane:
@rth and I agreed that the majority of this is done and it should be removed from our roadmap.


rth:
Just to provide more context @ryanking13 , we had an in-person discussion with Hood about some of the roadmap topics when he was in Paris recently. That's the outcome of that discussion, transcribed to PRs but of course, if you have any comments about it, please comment :)

ryanking13:
Cool, thanks for the information!

PR #3638, opened 2023-03-07
Add hereditary option to as_object_map

hoodmane:
This automatically apply as_object_map to the values of the map when the values are "plain old objects".

This behavior is slightly naive but I think it will satisfy most people who have requested this feature.

- [x] Add / update tests

### DIFF
--- a/src/core/jsproxy.h
+++ b/src/core/jsproxy.h
@@ -1,3 +1,7 @@
+/**
+ * A Python object that wraps a JavaScript object. Used to allow JavaScript
+ * objects to be passed into Python.
+ */
 #ifndef JSPROXY_H
 #define JSPROXY_H
 // clang-format off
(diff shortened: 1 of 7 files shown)


src/core/jsproxy.h
```diff
@@ -6,46 +10,53 @@
 // clang-format on
 #include "hiwire.h"
 
-/** A Python object that a JavaScript object inside. Used for any non-standard
- *  data types that are passed from JavaScript to Python.
+/**
+ *
```
ryanking13, line 14: I guess you were thinking of adding a comment here?

ryanking13 (commented):
Thanks, @hoodmane! What do you think of having a more general `depth` parameter, as what we do in `PyProxy.toPy` and in `JsProxy.to_js`?

PR #3644, opened 2023-03-08
DOC using conda for building from sources (Linux & Mac OS)

rth:
This tries to improve the building from the source documentation in particular by documenting how to install requirements with conda.

 Following https://github.com/pyodide/pyodide/pull/3602 we now test this both for Linux and MacOS (Intel CPU) in GHA CI. I also can confirm these instructions work with the Arm64 M2 CPU for Mac OS (since conda-forge provides the required binaries).


docs/development/building-from-sources.md
```diff
@@ -78,21 +84,52 @@ To build on Linux, you need:
...
+You would need a working native compiler toolchain, enough to build
+  [CPython](https://devguide.python.org/getting-started/setup-building/index.html#linux), for example,
+- `apt install build-essential` on Debian based systems.
+- Conda which can be installed from [MiniForge](https://github.com/conda-forge/miniforge)
+
+Then run,
+
+- `conda create -c conda-forge -n pyodide-env python=3.11`
+- `conda activate conda-forge`
+- `conda install -c conda-forge nodejs ccache f2c pkg-config swig make patch pkg-config texinfo autoconf automake libtool`
```
rth, line 101: If someone knows how to nest a non inline code block here let me know. The problem is that there are already twice nested
````
```
````
code blocks for creating tabs, and apparently one can't just recursively add more ` everywhere to make it work)

File: docs/development/building-from-sources.md [hunk 2]
```diff
@@ -78,21 +84,52 @@ To build on Linux, you need:
...
-- A working native compiler toolchain, enough to build
-  [CPython](https://devguide.python.org/getting-started/setup-building/index.html#macos-and-os-x).
+You would need a working native compiler toolchain, enough to build
+  [CPython](https://devguide.python.org/getting-started/setup-building/index.html#linux), for example,
+- `apt install build-essential` on Debian based systems.
+- Conda which can be installed from [MiniForge](https://github.com/conda-forge/miniforge)
+
+Then run,
+
+- `conda create -c conda-forge -n pyodide-env python=3.11`
```
ryanking13, line 99: Perhaps we can provide conda `environment.yaml` file and tell people to use it?

docs/development/building-from-sources.md
```diff
@@ -78,21 +84,52 @@ To build on Linux, you need:
...
-- System libraries in the root directory (
-  `sudo installer -pkg /Library/Developer/CommandLineTools/Packages/macOS_SDK_headers_for_macOS_10.14.pkg -target /`
-  should do it, see https://github.com/pyenv/pyenv/issues/1219#issuecomment-428305417)
-- coreutils for and other essential Unix utilities (`brew install coreutils`).
-- cmake (`brew install cmake`)
-- autoconf, automaker & libtool (`brew install autoconf automaker libtool`)
-- It is also recommended installing the GNU patch (`brew install gpatch`), and
-  GNU sed (`brew install gnu-sed`) and [re-defining them temporarily as `patch` and
+- `brew install coreutils cmake autoconf automaker libtool`
+- It is also recommended installing the GNU patch and
```
ryanking13, line 130: Actually, it is now mandatory to use gnu patch and sed.

ryanking13 (approved):
Thanks for the documentation work @rth!

docs/development/building-from-sources.md
jobovy, line 101: Just indenting by four spaces
```suggestion
    conda create -c conda-forge -n pyodide-env python=3.11`
    conda activate conda-forge`
    conda install -c conda-forge nodejs ccache f2c pkg-config swig make patch pkg-config texinfo autoconf automake libtool`
```
seems to do the trick.

jobovy: See my comment below: https://github.com/pyodide/pyodide/pull/3644#discussion_r1130278271

rth:
There was again some issue with SSH cert in CircleCI, but that was running on my fork and none of the CircleCI jobs matter were affected here, so merging as is. Thanks for the reviews!

docs/development/building-from-sources.md
rth: I forgot to address this comment.  I will try to actually follow these instructions with brew and see if any other things need updating.

bollwyvl:
Lovely stuff. :heart: 

Some other thoughts:
- for big envs, i'd _really_ recommend encouraging folk towards the `Mambaforge` variant of miniforge as it is faster at:
  - solving
  - downloading
  - installing (maybe)
- `micromamba` is _even smaller_, and doesn't create _another python_
- (almost?) all of the deps of this repo (i could find) are also available on conda-forge
  - if there's more out-of-band stuff, we can definitely add it
  - _always_ looking for more maintainers :heart_eyes: 
    - just [ask the bot](https://github.com/conda-forge/pyodide-build-feedstock/issues/new?assignees=&labels=&template=2-bot-commands.yml) :robot: 
- in addition to an `environment.yml`
  - maintaining (and testing in CI where possible) against `.conda.lock` files is a really good way to give folks something _exactly_ like CI...
    - these skip the expensive `Solving environment...` stage altogether
    - this could be done offline, or in a single CI job with [conda-lock](https://github.com/conda/conda-lock)

rth:
Thanks for your comments Nick!

> encouraging folk towards the Mambaforge

I know, but currently, this installs only a dozen of libraries mostly non Python, so the install time or dependency resolution is not really an issue. Also I'm pointing to the miniforge installer which does list  both mambaforge and miniforge installers. So I feel it's up to that repo's readme to explain to  users which one they should be installing and why.

> all of the deps of this repo (i could find) are also available on conda-forge 

 I personally find it a pain to maintain both an `environment.yml` and a `requirements.txt` where some files overlap and others don't.  That would make sense if we switch everything to `environment.yml` but we still need a working install outside of conda (for instance for Docker or people who don't want to use conda).
 
 We also have a [bunch of requirements](https://github.com/pyodide/pyodide/blob/main/pyodide-build/setup.cfg#L35) that we update quite often (micropip, pyodide-cli etc). Right now the workflow is create a tag, and use that package in some PR. Having to wait for a conda-forge forge PR even if made by a bot, would add extra complexity.

> maintaining (and testing in CI where possible) against .conda.lock

Thanks for the suggestion, would that work if there is a mixture of conda and pip installed packages?

PR #3645, opened 2023-03-09
Handle response files in pywasmcross

### LINKED ISSUE #3436: Improve `pywasmcross` to handle response files

ryanking13:
### Description

Superseeds #3438

Close #3438
Close #3436

### Checklists


hoodmane (approved):
Thanks @ryanking13!

ryanking13:
Thanks for the review!

PR #3649, opened 2023-03-10
WIP: build scipy with meson

rth:
Tries to continue @hoodmane 's work in  https://github.com/pyodide/pyodide/pull/3037


rth:
So there are a few issues that happen here,

1. 
```
Dependency lookup for OpenBLAS with method 'pkgconfig' failed: Pkg-config binary for machine 1 not  
found. Giving up.
```
as discussed in the parent issue. pkg-config is installed and setting `export PKG_CONFIG_LIBDIR=$WASM_PKG_CONFIG_PATH` should fix the path, so I don't know why it cannot find it.


2. `WARNING: Unknown CPU family wasm, please report this at https://github.com/mesonbuild/meson/issues/new`
     Maybe a false positive, the config looks correct.
     
3. ` ERROR: Compiler /usr/bin/gfortran can not compile programs.` depending on what one puts in the config of `tools/emscripten.meson.cross`. 

Initially, I felt the quickest might be to monkeypatch the [compiler checks in meson](https://github.com/mesonbuild/meson/blob/master/mesonbuild/compilers/fortran.py) but since it's run in a separate process by `meson-python` we can't easily modify that code at runtime, short of installing a modified version of meson.

So we would indeed likely need to make current checks pass with our compiler wrappers, as Hood was suggesting.

I'll likely not continue this PR in the near future. This was mostly so we have a separate PR on scipy with meson, independent from scipy. Feel free to take it over.

rth:
BTW, it looks like the last version of scikit-image also migrated to meson. So maybe that would be an easier package to start, since it only has Cython with a little C.

ryanking13:
There are too many build tools in the world 😂

rth:
Indeed :) But with distutils deprecations, many of Python scientific packages (or packages with C extensions) will move to meson. So that's one we likely can't avoid.

PR #3650, opened 2023-03-11
Ruff is now compatible with Python pattern matching

cclauss:
### Description

Closes #3624

Previous versions of `ruff` did not support Python 3.11 pattern matching and treated calls to it as SyntaxErrors.
* `pre-commit autoupdate`  # includes changes for psf/black 2023 stable style
* Fix codespell typos
* Fix mypy issues
* Fix ruff issues and add new ruff rules (pylint, McCabe code complexity) to the testing.

### Checklists

- [x] Add / update tests


pyproject.toml
```diff
@@ -27,46 +27,53 @@ module = [
...
   "PGH",   # pygrep-hooks
+  "PLC",   # pylint conventions
+  "PLE",   # pylint errors
+  "UP",    # pyupgrade
 ]
-# Remove E999 once pattern matching is supported
-# https://github.com/charliermarsh/ruff/issues/282
-ignore = ["E402", "E501", "E731", "E741", "E999"]
+ignore = ["E402", "E731", "E741"]
+line-length = 264
```
rth, line 61: Why is this needed? If there is a very long line somewhere I would have though it should be skipped explicitly. 
The line length we have should be still around 88 as black.

rth (approved):
Thanks @cclauss ! 
One comment otherwise LGTM.

pyproject.toml
cclauss: % `ruff --select=E501 --line-length=88 --statistics .`
```
260	E501	Line too long (104 > 88 characters)
```
Leaving `line-length` explicit encourages contributors not to add even longer lines.  Using `--ignore = ["E501"]` puts no such guardrails in place.

If there were a dozen long lines then I would have just shortened them.  It is kinda ugly to add `  # noqa: E501` in the codebase 260 times.  It would be a great `first-issue` to reduce long lines by 10 or 20 characters per line.

src/py/_pyodide/_base.py
```diff
@@ -71,7 +71,7 @@ def _last_assign_to_expr(mod: ast.Module) -> None:
         # In this case there can be multiple targets as in `a = b = 1`.
         # We just take the first one.
         target = last_node.targets[0]
-    elif isinstance(last_node, (ast.AugAssign, ast.AnnAssign)):
+    elif isinstance(last_node, ast.AugAssign | ast.AnnAssign):
```
cclauss, line 74: Is this autocorrect correct?!?  I have not used the `ast` module but this just looks wrong to me.
Repeated on line 117.

cclauss: Yes!  [PEP 604](https://peps.python.org/pep-0604) is implemented in Python >= 3.10 as discussed in https://github.com/charliermarsh/ruff/issues/2923

pyproject.toml
rth: OK, then I would have preferred to keep the E501 ignore. Most of the code is going to be reformatted by black or prettier anyway, as the value of 264 seems rather arbitrary and maybe a bit misleading with respect to the line length we do expect in the code.

cclauss: [The repo could use more good_first_issue](https://github.com/pyodide/pyodide/issues?q=is%3Aopen+is%3Aissue+label%3A%22good+first+issue%22) ;-)

cclauss: long-lines = ∞

src/tests/test_jsproxy.py
```diff
@@ -2249,8 +2250,8 @@ def test_python_reserved_keywords(selenium):
     assert o.async_ == 1
     assert o.async__ == 2
     assert o.async___ == 3
-    assert getattr(o, "async_") == 1
-    assert getattr(o, "async__") == 2
+    assert o.async_ == 1
+    assert o.async__ == 2
```
ryanking13, line 2254: These tests need to be kept as-is.

ryanking13 (changes requested):
Thanks @cclauss!

src/tests/test_jsproxy.py
hoodmane: Well the assertions here are a bit paranoid they could perhaps be removed. But they should either be removed or put back.

cclauss, line 2254:
```suggestion
    assert getattr(o, "async_") == 1
    assert getattr(o, "async__") == 2
```
% `ruff rule B009`
# get-attr-with-constant (B009)

Derived from the **flake8-bugbear** linter.

Autofix is always available.

Message formats:
* Do not call `getattr` with a constant attribute value. It is not any safer than normal property access.

hoodmane (approved):

hoodmane:
Thanks @cclauss!

PR #3656, opened 2023-03-14
Fix scikit-image optional dependencies

rth:
scikit-image currently installs both matplotlib and distutils at runtime, while I think they are not necessary.
 - matplotlib is an [optional dependency](https://github.com/scikit-image/scikit-image/blob/v0.19.3/requirements/optional.txt)
 - distutils is a build dependency

So this should reduce the download size when using scikit-image significantly (6.4 MB for matplotlib)

### DIFF
--- a/packages/scikit-image/meta.yaml
+++ b/packages/scikit-image/meta.yaml
@@ -13,12 +13,11 @@ source:
 requirements:
   host:
     - numpy
-  run:
     - distutils
+  run:
     - packaging
     - numpy
     - scipy
-    - matplotlib
     - networkx
     - Pillow
     - imageio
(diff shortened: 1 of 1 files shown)


packages/scikit-image/meta.yaml
```diff
@@ -13,12 +13,11 @@ source:
 requirements:
   host:
     - numpy
-  run:
     - distutils
```
ryanking13, line 16:
```suggestion
```

I guess setting `distutils` in a host dependency has no effect so I think we can simply remove that.

ryanking13 (approved):
Thanks @rth!

hoodmane:
Thanks @rth!

Issue #3332, opened 2022-12-09
Remove puppeteer from runtime JS dependencies 


rth:
In a call earlier and we discussed that it would be good to remove the puppeteer from the runtime JS dependencies to avoid re-downloading Chromium at each pyodide re-build.

@bollwyvl Had ideas on how to better do this, and I'll let him explain this :)

cc @ryanking13

bollwyvl:
Some options:
- move the heavyweight tests to separate folder, e.g https://github.com/jupyterlite/jupyterlite/tree/main/ui-tests
- use playwright, which is slightly smarter about managing its downloaded binaries
- use a known folder location with `PUPPETEER_CACHE_DIR`

ryanking13:
Closing as resolved. We've removed puppeteer from build dependencies.

PR #3660, opened 2023-03-15
Remove puppeteer from dev dependencies

hoodmane:
Every time we `npm i` or `npm ci` we are forced to install 150mb of chrome. I often do this many times a day when I'm working on Pyodide probably hundreds of times in total. I have also never used puppeteer since I use selenium for tests. This is a waste of time and bandwidth. We should only install puppeteer when needed.

I assume some CI steps will need to be updated to explicitly install puppeteer.

### DIFF
--- a/src/js/package-lock.json
+++ b/src/js/package-lock.json
@@ -33,7 +33,6 @@
         "npm-run-all": "^4.1.5",
         "nyc": "^15.1.0",
         "prettier": "^2.2.1",
-        "puppeteer": "^13.6.0",
         "rollup": "^2.48.0",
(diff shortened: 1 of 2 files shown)


ryanking13 (approved):
Thanks, @hoodmane! I agree that it is annoying that we download chromium whenever we build Pyodide, and there was an issue with this before #3332. I think moving to playwright would help as mentioned in that issue (it doesn't download browsers on installation), but I also normally write all tests in Python, so I'm not sure it will be actually useful...

Anyway, for now, we are not running those tests in our CI, so I'm +1 for removing puppeteer dev dependency.

Issue #3316, opened 2022-12-03
plugin/hook for Pyodide?


ryanking13:
Recently, we are getting some requests and ideas of accessing and modifying Pyodide internal objects before initialization:

- Accessing and modifying filesystem: #3274
- Caching compiled .py sources on runtime: https://github.com/pyodide/pyodide/issues/3269#issuecomment-1335252076

So I think it would be nice if we can provide some plugin/hook (I'm not sure this is a correct term for this) interface that users can easily add custom operations before initialization, without modifying or monkey patching Pyodide. Ideally, users can write useful plugins/hooks themselves, and maybe if we later found some of those hooks are critical for most use cases, we can think of porting them into Pyodide.

Below is my very rough idea of the plugin/hook interface. I haven't thought through this, so any ideas are welcome.

1. Add plugins parameter to loadPyodide

```typescript
type PyodidePlugin= {
  preInit?: ...
  preRun?: ...
  onFSInit?: ...
  onPyodideInit?: ...
  ...
}

function loadPyodide(options: {
   ...
   plugins?: PyodidePlugin[]
   ...
}
```

2. Run each hook inside plugins before/after specific steps during initialization

```typescript
function loadPyodide(...) {
   // after initializing module object
   plugins.forEach(hook=> { if (hook.hasOwnProperty("onModuleInit")) hook.onModuleInit(module) })
   ...
   // after initializing Pyodide
   plugins.forEach(hook=> { if (hook.hasOwnProperty("onPyodideInit")) hook.onPyodideInit() })
}
```

Maybe this is also partially related to how we can implement #3260.

cc: @joemarshall @madhur-tandon  @antocuni

joemarshall:
I'd like hook on package loading also, I.e. onloadpackage

joemarshall:
And onfsinit/onpyodideinit should be clearly named so it is obvious whether they get called before or after. E.g. onPyodideInited and onFSAvailable

ryanking13:
> I'd like hook on package loading also, I.e. onloadpackage

Yes, that sounds good to me. Probably we need to discuss whether we should implement it in micropip or in pyodide.loadPackage.

> And onfsinit/onpyodideinit should be clearly named so it is obvious whether they get called before or after. E.g. onPyodideInited and onFSAvailable

Yeah, the name of hooks that I wrote is just examples.

Issue #3274, opened 2022-11-20
Access to Emscripten's file system directly using Module.FS


madhur-tandon:
## Feature

Access the Emscripten Filesystem directly instead of waiting for `pyodide` to initialise and be ready first.

### Motivation

The Emscripten File system is provided by using the flag `-s 'EXPORTED_RUNTIME_METHODS=["FS"]'` when compiling C programs to WASM using `emcc`. This provides a direct access to the file system using `Module.FS`. 

In case of `pyodide` however, because of the way it is compiled, the file system is accessible through `pyodide.FS` (which is just a re-export of `Module.FS`) or perhaps through `pyodide._module.FS` as mentioned here:

https://pyodide.org/en/latest/usage/file-system.html#dealing-with-the-file-system

While there is no difference apart from the way to access it, the side effect is of waiting for `pyodide` to be ready first.

### Pitch

It would be nice to have access to `Module.FS` directly, which has the advantage of not waiting on the initialisation / startup time. One use-case could be fetching and setting up any additional non-pyodide specific files in parallel even before the runtime is ready.

Further, envisioning that more runtimes / interpreters (such as Micropython) will / are adopting WebAssembly, this provides a consistent way of accessing the File System across all such runtimes / interpreters.

### Alternatives

This is just meant to initiate a discussion and perhaps there are better ways to think about this. And there might be some caveats in this approach which are being overlooked.

### Additional context

Tagging some folks from PyScript here: @fpliger @antocuni @ntoll and from Pyodide as well: @hoodmane @rth

antocuni:
For more pyscript-specific context which might be helpful to understand why it would be useful for us: currently pyscript has a way to fetch files from external URLs and save the to the virtual FS. A very simple example is this:
```html
<py-config>
[[fetch]]
files = ["foo.py", "bar/__init__.py", "bar/baz.py"]
</py-config>
```

Currently we need to wait for `loadPyodide` to finish before we can start writing them to the FS.
I think that with enough workarounds we could probably find a way to "fetch-early/write-late", but having access to the FS in advance sounds cleaner (and might be generally useful anyway).

rth:
Thanks for the proposal! As far as I remember the initial motivation was to avoid creating global variables, and that having everything in the pyodide object is better from the namespacing point of view. In particular, defining global variables might be a problem when there are multiple Pyodide instances loaded, but Hood would probably have more thoughts about this.

> we would like to do (2) and (3) in parallel

[...]

Issue #3269, opened 2022-11-18
RFC Distribute bytecode compiled sources (.pyc)


rth:
To summarize the discussion in https://github.com/pyodide/pyodide/issues/3050 and the associated PRs there is currently a proposal to start distributing bytecode (.pyc) files instead of the original source files (.py),
 - for the standard library  https://github.com/pyodide/pyodide/pull/3166 (which will additionally put them in a zip file)
 - for package wheels https://github.com/pyodide/pyodide/pull/3253

The goal of this issue is to get user feedback on this idea and see if there are any concerns about it. So please don't hesitate to comment. You can also vote with emoji :+1: (choice 1) and :-1: (choice 2).

### Advantages

This has the potential to significantly speed up the initial load time as Python no longer needs to bytecode compile source code and can run the interpreter on it directly. See benchmarks below for more detail.

### Disadvantages

The are two main limitations,
 1. the source code is no longer readable, so for instance, if you get an exception, you will no longer be able to open an in-browser text editor to read the source code in dependencies and investigate it
 2. In the exception tracebacks,
     - we lose the quoted source code (including the improved [Python 3.11 error location indicated with arrows ('^')](https://docs.python.org/3/tutorial/errors.html#syntax-errors)) (see [this example](https://discuss.python.org/t/wasm-unvendoring-some-of-stdlib-modules/18076/8))
     - in traceback of packages, we can no longer reliably indicate the absolute path and instead will likely use a relative path (cf [this discussion](https://discuss.python.org/t/proper-file-paths-to-show-in-exception-for-py-compiled-files/21108))

### Possibilities

There are several possible outcomes,
 1. this proposal is accepted, we only distribute .pyc files without associated source code (.py), since distributing both would increase download size even more
 2. the proposal is rejected and we continue distributing .py files only
 3. we make separate builds: one for .py source and one for .pyc. This will significantly increase our maintenance burden and it's unclear if it would happen (so this may be the same as 2, or it may happen).

### Benchmarks 

See [source](https://discuss.python.org/t/proper-file-paths-to-show-in-exception-for-py-compiled-files/21108/6) for more details,
![image](https://user-images.githubusercontent.com/630936/202668007-3b3bba86-dffa-4b35-b903-79c817511d38.png)

casatir:
Hi,

[...]

PR #3664, opened 2023-03-17
Plugin / Hook support

ryanking13:
**This PR is not for merging.** This PR includes a simple hook that can override our file system bootstrap stage.

- Related: #3316

I am opening this to gather more opinions about supporting hooks which people can modify Pyodide's internal behavior.

Some possible use cases that I can think of are:

1. Modifying the file system before bootstrapping Pyodide.

For example, @madhur-tandon opened an issue (#3274) about having a separate file system that can prefetch files and can be shared between Pyodide instances.

On the other hand, @joemarshall mentioned about https://github.com/pyodide/pyodide/issues/3269#issuecomment-1335252076 caching the file system into indexedDB, and reuse it in the next visit (included PoC in this PR).

2. Supporting other platforms.

For example, there is a PR about adding support for greasemonkey (#3597). It is a good feature, however, AFAIK no Pyodide devs know deeply about greasemonkey, so it is hard for us to maintain it. But if we can support it as a hook/plugin, then someone else who knows about greasemonkey can maintain it.

---

If we decide to support this feature, then I think first we need to 1) standardize the Pyodide bootstrap stages and document it, 2) and also standardize some internal APIs (`loadScript`, `loadBinaryFile`) that people might want to modify.

WDYT?

cc: @madhur-tandon, @antocuni, @tubaman, @joemarshall

### DIFF
--- a/src/js/module.ts
+++ b/src/js/module.ts
@@ -57,6 +57,7 @@ export interface Module {
   FS: FS;
   addRunDependency: (id: string) => void;
   removeRunDependency: (id: string) => void;
+  fileSystemInitialized: boolean;
 }
 
 /**
@@ -130,6 +131,10 @@ function mountLocalDirectories(Module: Module, mounts: string[]) {
  * @param stdlibPromise A promise that resolves to the standard library.
  */
 function installStdlib(Module: Module, stdlibURL: string) {
+  if (Module.fileSystemInitialized === true) {
+    return;
+  }
+
   const stdlibPromise: Promise<Uint8Array> = loadBinaryFile(stdlibURL);
 
   Module.preRun.push(() => {
@@ -153,6 +158,7 @@ function installStdlib(Module: Module, stdlibURL: string) {
       })
       .finally(() => {
         Module.removeRunDependency("install-stdlib");
+        Module.fileSystemInitialized = true;
       });
   });
 }
@@ -162,6 +168,12 @@ function installStdlib(Module: Module, stdlibURL: string) {
  * @private
  */
 export function initializeFileSystem(Module: Module, config: ConfigType) {
+  config.hooks!.forEach((hook) => {
+    if (typeof hook.beforeFileSystemInitialized === "function") {
+      hook.beforeFileSystemInitialized(Module);
+    }
+  });
+
   installStdlib(Module, config.indexURL + "python_stdlib.zip");
   setHomeDirectory(Module, config.homedir);
   mountLocalDirectories(Module, config._node_mounts);
--- a/src/js/pyodide.ts
+++ b/src/js/pyodide.ts
@@ -4,7 +4,7 @@
 import ErrorStackParser from "error-stack-parser";
 import { loadScript, initNodeModules, pathSep, resolvePath } from "./compat";
 
-import { createModule, initializeFileSystem } from "./module";
+import { createModule, initializeFileSystem, Module } from "./module";
 import { version } from "./version";
 
 import type { PyodideInterface } from "./api.js";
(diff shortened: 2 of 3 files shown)


bollwyvl:
Sure, [jupyterlite-pyodide-kernel](https://github.com/jupyterlite/pyodide-kernel) would likely abuse the hell out of hooks, if available :smiling_imp: Tough to say _how_ exactly, before they exist, though. Maybe, knowing we're in a worker, if this gave the kernel the ability to work around https://github.com/jupyterlite/jupyterlite/issues/275, it would make sense to do it until a hack wasn't needed for the general cases.

hoodmane:
I'm surprised to hear that jupyterlite hasn't implementing blocking io yet. What's the holdup?

bollwyvl:
>  What's the holdup?

`input` isn't a big blocker for _my_ use cases, so haven't made time to do it... otherwise, the old story of: a lotta _"justs"_ and _"needs"_ for stuff in issues, but markedly fewer hands on keys on PRs. if there's good in-the-wild examples, we'd pick it up though.

hoodmane:
I think it would be reasonably quick to get something basic working with @alexmojaki's sync-message library.

bollwyvl:
> [`sync-message`](https://www.npmjs.com/package/sync-message)

yeah, it's all those caveats at the top. we haven't yet made the `ServiceWorker` full you-are-not-going-to-space-today required (though you lose synced contents), and it sounds like the alternative carries a fair amount of deployment risk...

rth:
What bothers me a bit with this approach is that it would be another abstraction on top of [emscripten hooks](https://emscripten.org/docs/api_reference/module.html?highlight=prerun#Module.preRun), while also relying on those hooks in the example. So in practice users would need to know both, and know when both Python `beforeFileSystemInitialized` is called and what `Module.preRun` does. And also that we would need to define a hook for each position, if one wants to go into details:  beforeCreateHomeDirectory, beforeSetEnvironment, beforeMountLocalDirectories etc.

I guess the reason we can't directly pass a list of functions to pass to preRun, preInit etc is that the function there doesn't take `Module` as an argument, and so it has to be passed via a closure. But how about instead of
```js
       indexedDBHook = {
            beforeFileSystemInitialized: (Module) => {
                Module.preRun.push(() => {
                    ....
                });
            }
        }
```
do 
```js
       indexedDBHook = {
            preRun: (Module) => {
                return (() => {
                    ....
                });
            }
        }
```
so we keep exactly the same names, and add it at the end of the `preRun` array. Though then the issue there is also all those `preRun` instructions we have and we may want to run some of them in a different order.

The other alternative is for instance,
```js
indexedDBHook : (Module) => {
                Module.preRun.push(() => {
                    ....
                });
           
}
```
where you don't necessarily need to append to `preRun` at the end, but could insert the function in arbitrary position. That would work but be more brittle to Pyodide version changes. Though maybe if we had some other array storing what we actually added to preRun e.g. `Module.preRunFuncNames = ["installStdlib", "createHomeDirectory", "setEnvironment", "mountLocalDirectories"]` one could then determine the index where to insert the hook function or even remplace one of those steps with a custom implementation.

WDYT? Not necessarily saying we should do that, just exploring possibilities. 

My use case, is I wanted to add some JS hooks to determine which exported symbols are accessed from the main module, probably via `Module.preInit` (or  before `Module.instantiateWasm` is called).

ryanking13:
@rth Thanks for the ideas! I agree that the abstraction in the current implementation is not very good.

I like the first option, hoping that the order of preRun does not matter... probably we can start with the first option, and switch to the second when we find some problems. 

```
       indexedDBHook = {
            preRun: (Module) => {
                return (() => {
                    ....
                });
            }
        }
```

PR #3665, opened 2023-03-17
Emscripten 3.1.33

hoodmane:


hoodmane:
Okay, the update breaks lightgbm and opencv, but I vote that we merge this and fix them in a followup.

hoodmane:
@ryanking13 this okay with you?

ryanking13:
Yes, it is okay with me. I'll have a look on build errors.

Probably we should print list of packages that are disabled when building packages by the way, so we don't forget disabled packages.

Issue #3655, opened 2023-03-14
Python wheels are not compressed by the CDN


rth:
It looks like the JsDelivr CDN does not compress Python wheels.

For instance, if you load numpy,
```
https://cdn.jsdelivr.net/pyodide/v0.22.1/full/numpy-1.23.5-cp310-cp310-emscripten_3_1_27_wasm32.whl

## Request Headers
Accept-Encoding: gzip, deflate, br

## Response Headers
content-length: 3126204
content-type: binary/octet-stream
There is no response encoding.
```
it will download 3.1MB of wheel. It's not terrible as it will be zip compressed, but brotli should achieve better compression.

However, if one takes the wheel (zip file) and apply brotli, the size is still around 3.0MB. So re-compressing on top doesn't help much (and it's disabled currently).

While if we were to take these files, put them in a tar (or a zip with 0 compression) and brotli compress it, the size would be 1.9MB in this example. But then technically it's no longer a standard wheel file.

Still, maybe we should switch to producing wheels with no compression, forcing the CDN to re-compress them (if possible) and let the browser do the decompression.

Anyway, also to keep this in mind when repacking more optimized bundles.

ryanking13:
> While if we were to take these files, put them in a tar (or a zip with 0 compression) and brotli compress it, the size would be 1.9MB in this example. But then technically it's no longer a standard wheel file.

That's very Interesting. We should check the size all wheel files + python_stdlib.zip. Probably we can also benefit from decompress speed if we use zip with 0 compression.

rth:
I need to check with JsDelivr folks if it's possible to compress binary files. It may also be that whether it's compressed is defined by MIME type, and that we should change the MIME type from `binary/octet-stream` to `application/wasm`.

MartinKolarik:
Yes, compression is based on mime type, and binaries are not compressed. Changing it to `application/wasm` would enable compression, but as you already found, double compression isn't particularly effective.

rth:
Thanks for the confirmation @MartinKolarik !

bollwyvl:
A big win for non-official-CDN would be compressing the `.data` file, if possible: `application/octet-stream` doesn't get compressed by default on a number of hosts (e.g. RTD), though it _does_ on `jsdelivr`.

rth:
Yes, this should be resolved in https://github.com/pyodide/pyodide/pull/3667. We were doing this previously for `.data` files but apparently forgot to adapt the mechanism after switching to `.whl` format.

PR #3667, opened 2023-03-17
Fix wheel compression with the JsDelivr CDN

rth:
Closes https://github.com/pyodide/pyodide/issues/3655

Following discussions in https://github.com/pyodide/pyodide/issues/3655 this sets the MIME type for wheels and zip files to `application/wasm` to enable CDN compression. Aslo optionally disables compression in zip files and wheels to make the brotli compression by the CDN more efficient.

We were doing this previously for `.data` files but didn't adapt the corresponding code when we switched to `.whl`

### Implementation details

Neither `pypa/build` (https://github.com/pypa/build/issues/258)  nor `wheel pack` (https://github.com/pypa/wheel/issues/355) currently reliably support switching the compression type or the compression level in the produced files. This means that we have to re-create new wheels/zip files after those tools are used. This involves extra file copies but I don't see a better alternative and it should be fast enough.

### DIFF
--- a/pyodide-build/pyodide_build/out_of_tree/pypi.py
+++ b/pyodide-build/pyodide_build/out_of_tree/pypi.py
@@ -27,7 +27,7 @@ from unearth.evaluator import TargetPython
 from unearth.finder import PackageFinder
 
 from .. import common
-from ..common import chdir
+from ..common import chdir, repack_zip_archive
 from ..logger import logger
 from . import build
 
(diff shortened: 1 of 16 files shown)


pyodide-build/pyodide_build/out_of_tree/pypi.py
```diff
@@ -194,15 +194,21 @@ def get_project_from_pypi(package_name, extras):
...
+    url: str, target_directory: Path, compression_level: int = 6
+) -> None:
     parsed_url = urlparse(url)
     if parsed_url.path.endswith("gz"):
         wheel_file = get_built_wheel(url)
         shutil.copy(wheel_file, target_directory)
+        wheel_path = target_directory / wheel_file.name
     elif parsed_url.path.endswith(".whl"):
+        wheel_path = target_directory / Path(parsed_url.path).name
         with open(target_directory / Path(parsed_url.path).name, "wb") as f:
```
ryanking13, line 207:
```suggestion
        with open(wheel_path, "wb") as f:
```

ryanking13 (approved):
Thanks @rth. This is awesome!

I check the size change of core packages, and it is very impressive.

Brotli compressed size:

(In main branch)

- python_stdlib.zip: 2214 KB
- astropy: 5790 KB
- numpy: 3017 KB
- opencv_python: 10334 KB
- pandas: 5071 KB
- scikit_image: 9717 KB
- scikit_learn: 4719 KB
- scipy: 11257 KB

(In this PR)

- python_stdlib.zip: 1423 KB
- astropy: 3965 KB
- numpy: 1966 KB
- opencv_python: 7737 KB
- pandas: 2881 KB
- scikit_image: 7666 KB
- scikit_learn: 2469 KB
- scipy: 7566 KB

rth:
OK, so the uploaded wheels are indeed uncompressed but so far the MIME type returned by the CDN is still octet-stream and the compression is not applied. I need to look into it.

PR #3668, opened 2023-03-17
Remove deprecated pyodide-build buildall and mkpkg  CLI endpoints

rth:
Removes deprecated  CLI entrypoints `pyodide-build buildall` and `pyodide-build mkpkg`, since the equivalent commands are now accessible via the main `pyodide` CLI.

We have implicitly marked these entrypoints as deprecated in the last release ([here](https://github.com/pyodide/pyodide/blame/main/docs/project/changelog.md#L499) and [here](https://github.com/pyodide/pyodide/blame/main/docs/project/changelog.md#L495)), so there is not point in keeping them around. Keeping both makes it just confusing when changes need to be made in the code base.


pyodide-build/pyodide_build/serve.py
```diff
@@ -12,9 +12,6 @@ def end_headers(self) -> None:
         super().end_headers()
 
 
-Handler.extensions_map[".wasm"] = "application/wasm"
```
rth, line 15: This was fixed a long time ago in upstream CPython somewhere around 3.7 or 3.8

ryanking13 (approved):
Thanks!

Issue #3632, opened 2023-03-04
Customizing python_stdlib.zip


rth:
Now that the stdlib is in python_stdlib.zip with https://github.com/pyodide/pyodide/pull/3582 (thanks again @ryanking13 !) the question is what could be the API to customize this file. 

For instance, in pydide-pack I would like to be able to customize what is put into the stdlib after repacking. There are several options,
 1. have a link to the stdlib in the `repodata.json`: it makes sense, but is problematic if one want to customize this without re-deploying a separate `repodata.json`
 2. have an option in `loadPyodide` with an optional URL to the stdlib. Either a) extending `fullStdLib: bool | str` to support URLs or b) adding yet another parameter there
 

WDYT?

ryanking13:
I like option 2, so people can provide their custom version of stdlibs. Some people might even want to put all third-party packages they use into a zipfile.

I was thinking of providing a compiled version of stdlib (e.g. python_stdlib_pyc.zip), and adding a parameter in `loadPyodide` to select which version of stdlib to use. But if we are going for option 2, then probably we won't need to add an extra parameter.

PR #3670, opened 2023-03-18
Add stdLibURL parameter to loadPyodide

rth:
Allows to customize the URL from which the Python standard library is loaded

Closes https://github.com/pyodide/pyodide/issues/3632


src/tests/test_pyodide.py
```diff
@@ -1792,3 +1792,26 @@ def test_python_version(selenium):
...
+
+    try:
+        selenium.run_js(
+            """
+            let pyodide = await loadPyodide({
+                fullStdLib: false,
+                stdLibURL: "./python_stdlib2.zip",
+            });
+            // Check that we can import stdlib library modules
+            let math = pyodide.pyimport('math');
```
ryanking13, line 1812: `math` is a built-in library so I think we should test a different one.

src/js/module.ts
```diff
@@ -162,7 +162,14 @@ function installStdlib(Module: Module, stdlibURL: string) {
  * @private
  */
 export function initializeFileSystem(Module: Module, config: ConfigType) {
-  installStdlib(Module, config.indexURL + "python_stdlib.zip");
+  let stdLibURL;
+  if (config.stdLibURL != undefined) {
+    stdLibURL = config.stdLibURL;
```
ryanking13, line 167: Question: what happens if we pass relative urls?

ryanking13 (approved):
Thanks, @rth! Probably we need to document how `stdLibURL` and `fullStdLib` will work together later. Though for that, I guess we should first write a stable specification for repodata.json...

src/js/module.ts
rth: It just passes them to fetch, so relative URLs also work. That's what is tested currently.

src/tests/test_pyodide.py
```diff
@@ -1792,3 +1792,26 @@ def test_python_version(selenium):
...
+    try:
+        selenium.run_js(
+            """
+            let pyodide = await loadPyodide({
+                fullStdLib: false,
+                stdLibURL: "./python_stdlib2.zip",
+            });
+            // Check that we can import stdlib library modules
+            let math = pyodide.pyimport('statistics');
+            assert(() => math.median([2, 3, 1]) === 2)
```
hoodmane, line 1813:
```suggestion
            let statistics = pyodide.pyimport('statistics');
            assert(() => statistics.median([2, 3, 1]) === 2)
```

WALID179 (approved):

PR #3676, opened 2023-03-20
Add experimental support for EM_JS code in side modules

hoodmane:
Based on my discussion with @szabolcsdombi in #3671, I thought it would be fun to experiment with enabling `EM_JS` in side modules. The `EM_JS` is put into a file called `blah.so.js` which contains code like:
```js
globalThis.__tmpMergeLibsFunction = function(Module) {
	function function_implemented_in_javascript(argument) { console.log('hello', argument); }
	
	Module.mergeLibSymbols(
		{function_implemented_in_javascript}
	);
}
```
we load it with `loadScript`, call `__tmpMergeLibsFunction` and then delete it. We need a lock around this code to ensure that another library doesn't write over the `__tmpMergeLibsFunction` global before the first is used (we need to use a global because of classic workers).

I tested it manually against @szabolcsdombi's example:
https://github.com/szabolcsdombi/pyodide-custom-module-imports/commit/3d51d923cb89941024ea31cc38b8df73a2a199ad
and it seems to work.

@sbc100

### Checklists

- [ ] Add / update tests

### DIFF
--- a/src/js/dynload.ts
+++ b/src/js/dynload.ts
@@ -7,6 +7,7 @@ declare var DEBUG: boolean;
 import { createLock } from "./lock";
 import { memoize } from "./util";
 import { PackageData } from "./load-package";
+import { loadScript } from "./compat";
 
 type ReadFileType = (path: string) => Uint8Array;
 
@@ -157,7 +158,7 @@ export async function loadDynlib(
  */
 export async function loadDynlibsFromPackage(
   pkg: PackageData,
-  dynlibPaths: string[],
+  libPaths: [string[], string[]],
 ) {
   // assume that shared libraries of a package are located in <package-name>.libs directory,
   // following the convention of auditwheel.
@@ -167,6 +168,15 @@ export async function loadDynlibsFromPackage(
 
   // This prevents from reading large libraries multiple times.
   const readFileMemoized: ReadFileType = memoize(Module.FS.readFile);
+  const [dynlibPaths, jslibPaths] = libPaths;
+  for (let jslib of jslibPaths) {
+    let libstr = Module.FS.readFile(jslib, { encoding: "utf8" });
+    libstr = "data:text/javascript," + libstr;
+    await loadScript(libstr);
+    let gt = globalThis as any;
+    gt.__tmpMergeLibsFunction(Module);
+    delete gt.__tmpMergeLibsFunction;
+  }
 
   const forceGlobal: boolean = !!pkg.shared_library;
 
--- a/emsdk/patches/0001-Allow-EM_JS-to-be-used-in-side-modules.patch
+++ b/emsdk/patches/0001-Allow-EM_JS-to-be-used-in-side-modules.patch
@@ -1,0 +1,48 @@
+From fb68a1cf6e2acb37cb91027747997279bcd034af Mon Sep 17 00:00:00 2001
+From: Hood Chatham <roberthoodchatham@gmail.com>
+Date: Mon, 20 Mar 2023 18:34:22 +0000
+Subject: [PATCH] Allow EM_JS to be used in side modules
+
+---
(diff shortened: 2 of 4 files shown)


src/js/dynload.ts
```diff
@@ -167,6 +168,15 @@ export async function loadDynlibsFromPackage(
...
   // This prevents from reading large libraries multiple times.
   const readFileMemoized: ReadFileType = memoize(Module.FS.readFile);
+  const [dynlibPaths, jslibPaths] = libPaths;
+  for (let jslib of jslibPaths) {
+    let libstr = Module.FS.readFile(jslib, { encoding: "utf8" });
+    libstr = "data:text/javascript," + libstr;
+    await loadScript(libstr);
+    let gt = globalThis as any;
+    gt.__tmpMergeLibsFunction(Module);
+    delete gt.__tmpMergeLibsFunction;
```
hoodmane, line 178: There should be an async lock around this code.

sbc100:
For EM_ASM this was fixed in https://github.com/emscripten-core/emscripten/pull/18228.   Here use `new Function` after extracting the JS string from linear memory.  Would that approach work for you?

hoodmane:
It would be nice to support pages with a content security policy that does not allow `unsafe-eval` (we need `wasm-unsafe-eval` . Though I'm not sure whether `import("data:text/javascript,blah")` or `importScripts("data:text/javascript,blah")` is allowed, since it is just `eval` in different clothes. There isn't any mention of it in the spec though:
> The following JavaScript execution sinks are gated on the "unsafe-eval" source expression:
>
>    * [eval()](https://tc39.github.io/ecma262#sec-eval-x)
>    * [Function()](https://tc39.github.io/ecma262#sec-function-objects)
>    * [setTimeout()](https://html.spec.whatwg.org/multipage/timers-and-user-prompts.html#dom-settimeout) with an initial argument which is not callable.
>    * [setInterval()](https://html.spec.whatwg.org/multipage/timers-and-user-prompts.html#dom-setinterval) with an initial argument which is not callable.

https://www.w3.org/TR/CSP3/#directive-script-src

sbc100:
> It would be nice to support pages with a content security policy that does not allow `unsafe-eval` (we obviously need `wasm-unsafe-eval`).
> 
> Though I'm not entirely sure that this is possible. I'm not sure whether `import("data:text/javascript,blah")` or ``("data:text/javascript,blah")` is allowed, since it is just `eval` in different clothes. There isn't any mention of it in the spec:
> 
> > The following JavaScript execution sinks are gated on the "unsafe-eval" source expression:
> > 
> > * [eval()](https://tc39.github.io/ecma262#sec-eval-x)
> > * [Function()](https://tc39.github.io/ecma262#sec-function-objects)
> > * [setTimeout()](https://html.spec.whatwg.org/multipage/timers-and-user-prompts.html#dom-settimeout) with an initial argument which is not callable.
> > * [setInterval()](https://html.spec.whatwg.org/multipage/timers-and-user-prompts.html#dom-setinterval) with an initial argument which is not callable.
> 
> https://www.w3.org/TR/CSP3/#directive-script-src

Upstream, we will probably provide "out-of-the-box" support for using `new Function` initially, and then we can look into `importScripts` as a followup if folks run into issues with it.

hoodmane:
I guess another feature that is missing from the approach I'm using here is the closure. You write:
 
> Its very common for EM_ASM code to assume it can call internal functions like out and err as well as any JS library functions.
https://github.com/emscripten-core/emscripten/pull/18228#issuecomment-1319366676

Which is certainly true. It is possible to say `Module.out` everywhere but it would be nice not to have to do it. Too bad `with` is not allowed in strict mode we could use `with(Module)` 😆.

ryanking13:
Cool, it would be really interesting if we can move some JS parts into packages so we can reduce the size of the main module.

PR #3677, opened 2023-03-21
Remove deprecated methods

ryanking13:
Remove deprecated methods before releasing 0.23 following the [deprecation timeline](https://github.com/pyodide/pyodide/blob/main/docs/project/deprecation-timeline.md)

### DIFF
--- a/src/tests/test_pyodide_http.py
+++ b/src/tests/test_pyodide_http.py
@@ -12,8 +12,8 @@ def test_open_url(selenium, httpserver):
     assert (
         selenium.run(
             f"""
-            import pyodide
-            pyodide.open_url('{request_url}').read()
+            from pyodide.http import open_url
+            open_url('{request_url}').read()
             """
         )
         == "HELLO"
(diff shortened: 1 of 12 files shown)


hoodmane:
Thanks @ryanking13! You need to keep
```py
from . import _state # noqa: F401
```
in `__init__.py`

ryanking13:
Good point, thanks!

src/tests/test_pyodide_http.py
```diff
@@ -12,8 +12,8 @@ def test_open_url(selenium, httpserver):
     assert (
         selenium.run(
             f"""
-            import pyodide
-            pyodide.open_url('{request_url}').read()
+            from pyodide.http import open_url
```
hoodmane, line 15: Have to add
```js
    pyodide.pyimport("pyodide.http").destroy();
```
to `INITIALIZE_SCRIPT`.

rth (approved):
Maybe we should still mention it in the changelog that deprecated methods were removed?
Otherwise LGTM.

PR #3684, opened 2023-03-22
MAINT Improvements to mkpkg error handling

### LINKED ISSUE #3682: pyodide-build shouldn't display tracebacks for "user error" / expected failures
## 🐛 Bug

pyodide-build has been updated to display rich tracebacks. This is useful only if the error was actually unexpected. In many places `pyodide-build` exits by raising an error which should just print the message and set a nonzero status code but not display some giant traceback message.

hoodmane:
Resolves #3682. Removes tracebacks from "user errors" and improves error messages.


ryanking13 (approved):
Thanks, @hoodmane! Maybe let's keep #3682 open until we fix error messages in other commands too.

PR #3685, opened 2023-03-22
Update all packages

hoodmane:


packages/gdal/meta.yaml
```diff
@@ -1,10 +1,10 @@
 package:
   name: gdal
-  version: 3.5.1
+  version: 3.6.3
 
 source:
-  url: https://github.com/OSGeo/gdal/releases/download/v3.5.2/gdal-3.5.2.tar.gz
-  sha256: fbd696e1b2a858fbd2eb3718db16b14ed9ba82521d3578770d480c74fe1146d2
+  url: https://files.pythonhosted.org/packages/1e/82/71ce6786f464da7591f77bfc550002c85bd2df683d4a4929d69f2a3f1808/GDAL-3.6.3.tar.gz
```
ryanking13, line 6: It seems like pyodide-build thought that gdal is a python package. Probably we need to change the name to the `libgdal` (and also update the logic in mkpkg to ignore libraries).

hoodmane: I fixed this in #3684. The problem was that mkpkg wasn't updated when the type field was introduced.

hoodmane:
@rth while you're at it could you revert the change to gdal?

rth:
Yes, will do later today.

rth:
I removed a few upstreamed patches, and reverted packages that would need more work to update (and could be done in a separate PR). In particular, xgboost and scikit-image now build via meson.

rth:
For Fiona, we apparently vendor the zip files with tests. On one side we may not want to download it at test time, but committing 150kB of binary data to the repo for each update of a single package is not a very sustainable solution either. For now, I reverted this update.

About changelog, I don't really see point of documenting package updates (unless it's some really important update). Pyodide is among other things a  package distribution, so each release will have package updates and that's why we keep [that table with package versions](https://pyodide.org/en/latest/usage/packages-in-pyodide.html) for each release. IMO linking to it should be enough.

hoodmane:
> About changelog,

Agreed. We've been completely inconsistent about this anyways...

hoodmane:
I wonder if we could add a marker to packages that we revert to indicate that they require manual intervention to upgrade. Since I think we keep reverting the same packages...

ryanking13:
Thanks, @rth!

> For Fiona, we apparently vendor the zip files with tests. On one side we may not want to download it at test time, but committing 150kB of binary data to the repo for each update of a single package is not a very sustainable solution either. For now, I reverted this update.

I was too lazy to write tests while I was working on GDAL+fiona 👀. Since we now support out-of-tree builds of recipes, probably we can build and test some packages out-of-tree. I'll look into this after the release.

Issue #3687, opened 2023-03-23
CDN wheel compression not applied


ryanking13:
> OK, so the uploaded wheels are indeed uncompressed but so far the MIME type returned by the CDN is still octet-stream and the compression is not applied. I need to look into it.

_Originally posted by @rth in https://github.com/pyodide/pyodide/issues/3667#issuecomment-1477760076_
            
I'm opening this to make sure we don't forget this before releasing 0.23.0

rth:
So indeed the content-type is not correctly applied,
```
aws s3api head-object --bucket pyodide[..] --key dev/full/numpy-1.24.2-cp311-cp311-emscripten_3_1_32_wasm32.whl
{
    "AcceptRanges": "bytes",
    "LastModified": "Fri, 24 Mar 2023 05:45:42 GMT",
    "ContentLength": 3012139,
    "ETag": "\"ee0cd032415368b5ee863b2bcb9166f2\"",
    "CacheControl": "max-age=3600, public",
    "ContentEncoding": "gzip",
    "ContentType": "binary/octet-stream",
    "ServerSideEncryption": "AES256",
    "Metadata": {}
}
```
Maybe there was I was misusing multiple includes/excludes in the AWS CLI command. Will try to fix it later today.

rth:
I confirm that it works as expected now.

PR #3693, opened 2023-03-24
Fix compression of the dev deployment

rth:
Closes https://github.com/pyodide/pyodide/issues/3687

I forgot to make the changes for the dev deployment in  https://github.com/pyodide/pyodide/pull/3667
Overall those are mostly identical lines repeated several times, however because the cost of error is pretty high I prefer not to factorize it for now. Instead, we should rethink this system, maybe moving dev deployment to a separate bucket and making release deployment immutable.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -517,7 +517,7 @@ jobs:
           command: |
             find dist/ -type f -print0 | xargs -0 -n1 -I@ bash -c "echo \"Compressing @\"; gzip @; mv @.gz @;"
             aws s3 rm --recursive "s3://pyodide-cdn2.iodide.io/dev/full/"
-            aws s3 sync dist/ "s3://pyodide-cdn2.iodide.io/dev/full/" --exclude '*.data' --cache-control 'max-age=3600, public' --content-encoding 'gzip'  # 1 hour cache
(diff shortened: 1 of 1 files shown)


ryanking13 (approved):
Thanks!

.circleci/config.yml
```diff
@@ -517,15 +517,15 @@ jobs:
           command: |
             find dist/ -type f -print0 | xargs -0 -n1 -I@ bash -c "echo \"Compressing @\"; gzip @; mv @.gz @;"
             aws s3 rm --recursive "s3://pyodide-cdn2.iodide.io/dev/full/"
-            aws s3 sync dist/ "s3://pyodide-cdn2.iodide.io/dev/full/" --exclude '*.data' --cache-control 'max-age=3600, public' --content-encoding 'gzip'  # 1 hour cache
+            aws s3 sync dist/ "s3://pyodide-cdn2.iodide.io/dev/full/" --exclude '*.zip' --exclude '*.whl' --exclude "*.tar" --cache-control 'max-age=3600, public' --content-encoding 'gzip'  # 1 hour cache
             aws s3 sync dist/ "s3://pyodide-cdn2.iodide.io/dev/full/" --exclude '*' --include '*.data' --cache-control 'max-age=3600, public'  --content-type 'application/wasm' --content-encoding 'gzip'  # 1 hour cache
```
ryanking13, line 521: Actually, I think we need to fix two more lines as we have two types of deployments (normal, debug).

PR #3695, opened 2023-03-24
Fix invalid yaml for github actions example in doc

viblo:
### Description
Fixes indentation of the example github action yaml for building.

### Checklists


hoodmane:
Is there something wrong with this patch @viblo? It looks correct to me.

viblo:
@hoodmane I replaced it with https://github.com/pyodide/pyodide/pull/3696 instead (which updated the emscripten build action version as well)

PR #3699, opened 2023-03-25
DOC Remove pyodide-build CLI from the documentation

rth:
The [remaining commands in pyodide-build  CLI](https://pyodide.org/en/latest/usage/api/pyodide-build-cli.html) are mostly private so there is no reason to keep them documented.

Closes https://github.com/pyodide/pyodide/issues/2879


ryanking13 (approved):
Thanks, @rth!

Issue #3407, opened 2023-01-03
Deploy both py and pyc sources


rth:
I think the outcome of https://github.com/pyodide/pyodide/issues/3269 is that while .pyc sources are a huge load time improvement, for some use-cases (e.g. teaching) full errors tracebacks is more critical than load speed. So the ideal case would be to distribute both versions.

As mentioned by @hoodmane  in  https://github.com/pyodide/pyodide/issues/3051#issuecomment-1370070607
> Ideally I think it ought to be a setting for loadPyodide whether to use the pyc files or the full ones. We should also probably look further into the possibilities for fixing up tracebacks, if it's reasonable we might offer an API for it.

For the topic of fixing the tracebacks with the pyc version, it still means that the source code of the py version needs to be deployed somewhere. I mean even if it turns out that the best way is some other format for source code (e.g. so a particular line could be fetched via an HTTP range request when the exception happens), in end I feel like aside from this traceback loading mechanism, there should be a way to look at actual source code we distribute if there is any doubt for debugging without having to rebuild anything.

There are several ways both versions could be deployed,
 1. Using a different suffix in the indexURL e.g. (`v0.22.0/full/` vs `v0.22.0/pyc/`).  There could be an option in loadPyodide to make switching easier.
 2. Using the same indexURL, where both versions are deployed. The problem here is that for packages, if wheels are in the same folder, it's difficult to differentiate pyc compiled wheel from the original wheel (e.g. for numpy it would be the same filename), also one would need two repodata.json. So in the end IMO option 1 would be easier.

For package wheels https://github.com/pyodide/pyodide/pull/3253 it would be trivial to build both at the same time. For CPython itself  https://github.com/pyodide/pyodide/pull/3166 I guess we could also do it in the build core. The question is though which version we test (testing both would be too slow) and how we make sure that the version we don't test is not broken.

ryanking13:
I also think option 1 is better.

> The question is though which version we test (testing both would be too slow) and how we make sure that the version we don't test is not broken.

If we would want to test only one version, I think testing .pyc compiled version is more reasonable. If it works on .pyc version, I think it will hopefully also work on .py version, but the opposite might not be true. For example, if a package has a hardcoded path to `.py` file inside the source code, it might not work in compiled versions.

rth:
Resolved in https://github.com/pyodide/pyodide/pull/3701

PR #3700, opened 2023-03-26
Support for py-compiling folders + repodata.json update

rth:
This extends `pyodide py-compile` to take as input zip files, and a folder with built packages. It would also now update file paths and checksums in `repodata.json` when run on the full `dist/ ` folder.

Here we also run it as part for the CI on all packages and check that the resulting py-compiled packages can be imported in  node (which is sufficient to detect the most obvious issues). See output of the `ci/circleci: build-test-pyc-packages` CI job.


This is necessary to deploy both py and pyc builds https://github.com/pyodide/pyodide/issues/3407 (the actual deployment is not implemented in this PR).

```
 Usage: pyodide py-compile [OPTIONS] PATH                                                                                                             
                                                                                                                                                      
 Compile .py files to .pyc in a wheel, a zip file, or a folder with wheels or zip files.                                                              
 If the provided folder contains the repodata.json file, it will be     rewritten with the updated wheel / zip file paths and sha256 checksums.       
 Registered by: pyodide-build                                                                                                                         
                                                                                                                                                      
╭─ Arguments ────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╮
│ *    path      PATH  Path to the input wheel or a folder with wheels or zip files.                                                                 │
│                      [default: None]                                                                                                               │
│                      [required]                                                                                                                    │
╰────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╯
╭─ Options ──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╮
│ --silent               --no-silent             Silent mode, do not print anything.                                                                 │
│                                                [default: no-silent]                                                                                │
│ --keep                 --no-keep               Keep the original wheel / zip file.                                                                 │
│                                                [default: no-keep]                                                                                  │
│ --compression-level                   INTEGER  Compression level to use for the created zip file                                                   │
│                                                [default: 6]                                                                                        │
│ --help                                         Show this message and exit.          
````

### CODE REVIEW

File: pyodide-build/pyodide_build/common.py [hunk 1]
```diff
@@ -464,3 +465,27 @@ def repack_zip_archive(archive_path: Path, compression_level: int = 6) -> None:
         ) as fh_zip_out:
             for name in fh_zip_in.namelist():
                 fh_zip_out.writestr(name, fh_zip_in.read(name))
+
+
+def _get_sha256_checksum(archive: Path) -> str:
```
rth, line 470: Extracted without changes from buildpkg.py where this code was before.

pyodide-build/pyodide_build/tests/test_common.py
```diff
@@ -218,7 +218,7 @@ def test_make_zip_archive(tmp_path, compression_level, expected_compression_type
     make_zip_archive(output_dir, input_dir, compression_level=compression_level)
 
     with zipfile.ZipFile(output_dir) as fh:
-        assert fh.namelist() == ["b.txt", "c/", "c/d"]
+        assert set(fh.namelist()) == {"b.txt", "c/", "c/d"}
```
rth, line 221: Unrelated but this test was failing for me locally due to a different file order.

pyodide-build/pyodide_build/_py_compile.py
```diff
@@ -190,20 +195,131 @@ def _py_compile_wheel(
...
     Returns
     -------
-    wheel_path_out
-        processed wheel with .pyc files.
+    path
+        path to processed archive with .pyc files. Or None if the file was not py-compiled
+        (e.g. if it's a zip with no .py files)
+    """
+    if input_path.suffix not in [".whl", ".zip"]:
+        raise ValueError(f"Error: only .whl files are supported, got {input_path.name}")
```
ryanking13, line 203:
```suggestion
        raise ValueError(f"Error: only .whl or .zip files are supported, got {input_path.name}")
```

pyodide-build/pyodide_build/_py_compile.py
```diff
@@ -190,20 +195,131 @@ def _py_compile_wheel(
...
+    Returns None if the file should not be py-compiled.
+
+    Examples
+    --------
+    >>> _get_py_compiled_archive_name(Path("snowballstemmer-2.2.0-py2.py3-none-any.whl"))
+    'snowballstemmer-2.2.0-cp311-none-any.whl'
+    >>> _get_py_compiled_archive_name(Path("test-1.0.0.zip"))
+    """
+    # TODO: fix py-compilation of the following packages
+    if path.name.startswith(("RobotRaconteur", "astropy-", "opencv_python-")):
```
ryanking13, line 230: What's the problem with these packages?

ryanking13 (approved):
Thanks!

pyodide-build/pyodide_build/_py_compile.py
rth: - RobotRaconteur has some file with invalid syntax
 - opencv-python explicitly requires the `config.py` file to exist, so it shouldn't be compiled but then it breaks our test assumtion of all or nothing for .pyc
 - astropy also py-compiles fine but fails with an import error, don't remember why now.
 
 Overall because this PR is already quite large I would rather fix those separately. Will open an issue.

PR #3701, opened 2023-03-26
Deploy py-compiled build

rth:
This enables deployment of a  py-compiled build under,
```
https://cdn.jsdelivr.net/pyodide/<version>/pyc/
```

Here both the packages and the stdlib would be py-compiled.

With the stdlib being unvendored (thanks @ryanking13 ) it's actually relatively straightforward to do this as a post-processing step after the build. 

This is built on top of https://github.com/pyodide/pyodide/pull/3700 so that PR should be merged first.

Closes https://github.com/pyodide/pyodide/issues/3269 Closes https://github.com/pyodide/pyodide/issues/3269

This also replaces the previous `aws cli` commands with a Python script, so that CI setup is less repetitive, and we can validate things better. 
```
$ python tools/deploy_s3.py --help
                                                                                                                                                      
 Usage: deploy_s3.py [OPTIONS] LOCAL_FOLDER REMOTE_PREFIX                                                                                             
                                                                                                                                                      
 Deploy a dist folder with Pyodide packages to AWS S3                                                                                                 
                                                                                                                                                      
╭─ Arguments ────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╮
│ *    local_folder       PATH  Path to the local folder [default: None] [required]                                                                  │
│ *    remote_prefix      PATH  Remote prefix [default: None] [required]                                                                             │
╰────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╯
╭─ Options ──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╮
│ *  --bucket                                         TEXT  bucket name [default: None] [required]                                                   │
│    --cache-control                                  TEXT  Cache control header to set [default: max-age=30758400, immutable, public]               │
│    --pretend               --no-pretend                   Don't actually upload anything [default: no-pretend]                                     │
│    --overwrite             --no-overwrite                 Overwrite existing files [default: no-overwrite]                                         │
│    --rm-remote-prefix      --no-rm-remote-prefix          Remove existing files under the remote prefix [default: no-rm-remote-prefix]             │
│    --install-completion                                   Install completion for the current shell.                                                │
│    --show-completion                                      Show completion for the current shell, to copy it or customize the installation.         │
│    --help                                                 Show this message and exit.                                                              │
╰────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╯

```

TODO:
 - [x]  switch to a docker image in CI used for deployment where we can install pyodide-build (needs Python 3.11). As the `cibuilds/github` image we use is no longer maintainer.
 - [x] manually check that the instructions in that docker image work.


hoodmane:
> This also replaces the previous aws cli commands with a Python script, so that CI setup is less repetitive, and we can validate things better.

Maybe this should be a separate pr?

rth:
Well most of this PR is re-writing that deployment script. Once we have it,  doing the pyc deployment is just a couple of lines in the CI setup since most of the work for it is done in https://github.com/pyodide/pyodide/issues/3407

.circleci/config.yml
```diff
@@ -500,39 +502,30 @@ jobs:
...
-            # to reduce storage size and outbound AWS traffic. The HTTP
-            # requests for these files made to S3 will result in a response
-            # with gzip compression. However because JsDelivr CDN proxies and caches these
-            # requests, this has no impact on the compression of zip files and
-            # wheels once the files are served to users via CDN.
-            find dist/ -type f -print0 | xargs -0 -n1 -I@ bash -c "echo \"Compressing @\"; gzip @; mv @.gz @;"
-            aws s3 sync dist/ "s3://pyodide-cdn2.iodide.io/v${CIRCLE_TAG}/full/" --exclude '*.zip' --exclude '*.whl' --exclude "*.tar" --cache-control 'max-age=30758400, immutable, public' --content-encoding 'gzip'    # 1 year cache
-            aws s3 sync dist/ "s3://pyodide-cdn2.iodide.io/v${CIRCLE_TAG}/full/" --exclude '*' --include '*.zip' --include "*.whl" --include "*.tar" --cache-control 'max-age=30758400, immutable, public'  --content-type 'application/wasm' --content-encoding 'gzip'  # 1 year
+            python3 tools/deploy_s3.py dist/ "v${CIRCLE_TAG}/full/" --bucket "pyodide-cdn2.iodide.io" --cache-control 'max-age=30758400, immutable, public'
```
hoodmane, line 513: Good to see this deployment logic factored out and tested. In the future I think it would be good to move this whole recipe into a Python script that we can test (other than install requirements of course).

docs/project/changelog.md
```diff
@@ -134,6 +134,12 @@ myst:
   the paths and checksums it contains will also be updated
   {pr}`3700`
 
+- {{ Enhancement }} A py-compiled build which has smaller and faster to load
```
hoodmane, line 137:
```suggestion
- {{ Enhancement }} A py-compiled build which has smaller and faster-to-load
```
nitpick

docs/project/changelog.md
```diff
@@ -134,6 +134,12 @@ myst:
...
+  packages is now deployed under
+  `https://cdn.jsdelivr.net/pyodide/v0.23.0/pyc/` (also for for following
+  versions). The exceptions obtained with this builds will not include code
```
hoodmane, line 140:
```suggestion
  `https://cdn.jsdelivr.net/pyodide/v0.23.0/pyc/` (also for future
  versions). The exceptions obtained with this builds will not include code
```

tools/deploy_s3.py
```diff
@@ -0,0 +1,124 @@
...
+    s3_client = boto3.client("s3")
+
+    typer.echo(f"Deploying {local_folder} to s3://{bucket}/{remote_prefix}")
+    typer.echo("Options: ")
+    typer.echo(f" - {cache_control=}")
+    typer.echo(f" - {pretend=}")
+    typer.echo(" - content-encoding: gzip")
+
+    if rm_remote_prefix:
+        _rm_s3_prefix(bucket, str(remote_prefix).lstrip("/"))
```
hoodmane, line 57: Maybe we want to validate the prefix in some way to reduce chance of accidentally deleting everything?

tools/deploy_s3.py
```diff
@@ -0,0 +1,124 @@
...
+            fh_compressed.seek(0)
+
+            content_type = None
+            if file_path.suffix in (".zip", ".whl", ".tar", ".a"):
+                content_type = "application/wasm"
+            elif file_path.suffix == ".ts":
+                # This will not be correctly detected by mimetypes.
+                # However, JsDelivr will currently not serve .ts file in the
+                # custom CDN configuration, so it does not really matter.
+                content_type = "text/x.typescript"
```
hoodmane, line 91: `.d.ts` files?

tools/deploy_s3.py
hoodmane: Shouldn't pretend also apply to deleting stuff?

rth: > Shouldn't pretend also apply to deleting stuff?

Yes, good that someone is reviewing this code ^^

> Maybe we want to validate the prefix in some way to reduce chance of accidentally deleting everything?

Indeed. Let's do more validation of both of the prefix in general and only allow deletion of prefixes ending in "/dev/".

tools/deploy_s3.py
rth: Well the MIME type for .ts and .d.ts files is [a bit unclear](https://stackoverflow.com/a/13214026/1791279) but it doesn't really matter, since both are currently not allowed to be served via JsDelivr

https://cdn.jsdelivr.net/pyodide/v0.22.1/full/pyodide.ts (returns 403 forbidden)

So people get it from npm anyway.

rth:
The `dev` deployment went through but the returned MIME types are not correct. Currently they are `text/plain` for both .js and .zip files which is wrong. Will look into it.

PR #3702, opened 2023-03-26
Customize build type in indexURL used by console.html

rth:
Allows customizing the build version used in the REPL via the `build` URL parameter.  For instance,
```
https://pyodide.org/en/latest/console.html?build=debug
```
would load the debug dev build (and set the index URL to `https://cdn.jsdelivr.net/pyodide/dev/debug/`).

@hoodmane you mentioned this for debug builds. And also something useful for py-compiled builds https://github.com/pyodide/pyodide/pull/3701

I have not seen any noticeable load time impact of changing the script URL dynamically in the REPL, but it's also a bit hard to measure reliably.

You can test it with the [current RTD deployment](https://pyodide--3702.org.readthedocs.build/en/3702/console.html?build=debug)

Maybe there is some better parameter name than "build".


src/templates/console.html
```diff
@@ -234,7 +234,22 @@
...
+      let indexURL = "{{ PYODIDE_BASE_URL }}";
+      const urlParams = new URLSearchParams(window.location.search);
+      if (urlParams.has("build")) {
+        indexURL = indexURL.replace(
+          "/full/",
+          "/" + urlParams.get("build") + "/",
+        );
+      }
+      script.src = indexURL + "pyodide.js";
+      document.head.appendChild(script);
```
hoodmane, line 252: Rather than this script tag stuff, you should use `await import(script);`. This can go in the top of main.

hoodmane (commented):
Thanks for working on this @rth! This is indeed a feature that I would appreciate.

src/templates/console.html
rth: Yes, good idea, done. Also added some validation for the URL param to be in ['full', 'debug', 'pyc'] for now.

Issue #3588, opened 2023-02-15
Package Request - openai


RighiAmaral:
🐍 Package Request

- Package Name and Version: openai / latest:
- Package URL: https://pypi.org/project/openai/ 
- Package Dependencies that needs to be resolved first: 
  - aiohttp
    - aiosignal
      - frozenlist 
    - async-timeout 
    - attrs 
    - charset-normalizer
    - frozenlist 
    - multidict 
    - yarl 
      - idna 
      - multidict 
  - requests 
    - certifi 
    - charset-normalizer 
    - idna 
    - urllib3 
  - tqdm 
    - colorama

rth:
yarl and idna will be added with https://github.com/pyodide/pyodide/pull/3703

Right now the installation would fail with,
```py
>>> await micropip.install('openai', keep_going=True)
Traceback (most recent call last):
  File "<console>", line 1, in <module>
  File "/lib/python3.11/site-packages/micropip/_micropip.py", line 580, in install
    raise ValueError(
ValueError: Can't find a pure Python 3 wheel for: 'multidict<5.0,>=4.5'
```
because we package multidict 6.0 and it's pulling this old version because it installs aiohttp 3.6.2 from 2019 as that's the last version that has a pure python wheel (and they switched to C extensions since).

So we would indeed also need to package `aiohttp` but then it's a bit unfortunate to also have to package all those pure Python packages it depends on. Maybe we need to rethink how we handle such cases.

ht55ght55:
bump

mgmorcos:
👍

maksiksking:
bump
OpenAI's API tools are incredibly relevant right now with AI becoming more and more prominent, it's currently only possible to use GPT-3 (quite bad old models) without this module.

This would make developing ChatGPT apps way easier and is definitely in high demand.

pritam-dey3:
bump. This will be helpful to build more applications that interact with chatGPT that runs entirely on client side. Since the applications mostly send requests back forth with OpenAI, it doesn't make sense to route the same requests through backend. True, it can be done with javascript, but we want to use Python.

atbe:
bumop

bbaranow:
bump

subbu963:
bump

avielfedida:
bump

hoodmane:
Closed by #4539.

PR #3703, opened 2023-03-27
PKG Add yarl and idna

rth:
This adds the [yarl](https://github.com/aio-libs/yarl) package which is a requirement (with some Cython code) for the openai client library (https://github.com/pyodide/pyodide/issues/3588).  At least 3 people asked for it on different channels...

[idna](https://github.com/kjd/idna) is a Pure python dependency of yarl.

In any case, both of these libraries are pretty small and seem useful for web-related stuff.

I added tests and a changelog. Unless there are comments, will merge on green CI later today as this is fairly low risk.

### DIFF
--- a/docs/project/changelog.md
+++ b/docs/project/changelog.md
@@ -199,9 +199,10 @@ myst:
 ### Packages
 
 - New packages: fastparquet {pr}`3590`, cramjam {pr}`3590`, pynacl {pr}`3500`,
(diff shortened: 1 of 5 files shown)


ryanking13 (approved):
> At least 3 people asked for it on different channels...

Seems like lots of people want to use ChatGPT in Pyodide ^^...;

PR #3707, opened 2023-03-28
DOC Re-organize changelog for 0.23.0

rth:
Mostly moving entries around and better organizing them in sections.

+ adding the list of contributors to this release

### DIFF
--- a/docs/project/changelog.md
+++ b/docs/project/changelog.md
@@ -15,33 +15,7 @@ myst:
 
 ## Unreleased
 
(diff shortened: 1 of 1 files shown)


rth:
Also it looks like the badges in the changelog got broken since 0.22.1 probably because the CSS definition changed in the theme we use.  I put the CSS  back in this PR.

image

[Rendered version](https://pyodide--3707.org.readthedocs.build/en/3707/project/changelog.html)

PR #3719, opened 2023-03-31
Detect when PyProxies are used with the GIL released and raise an error

### LINKED ISSUE #3717: Should detect when a PyProxy is used with GIL released
## 🐛 Bug

The following code attempts to use a PyProxy with the GIL released.
```js
pyodide = await loadPyodide({stdout: (msg) => pyodide.globals.get("x")});
pyodide.runPython("print(1)")
```
Chaos ensues:
Screenshot from 2023-03-30 17-10-59

### Expected behavior

Throw a reasonable no-gil error. Don't fatally fail.

hoodmane:
Resolves #3717.

Before:

```js
$ tstate = pyodide._module._PyEval_SaveThread()
$ pyodide.globals.get("x")
```

After:

```js
$ tstate = pyodide._module._PyEval_SaveThread()
$ pyodide.globals.get("x")
Uncaught NoGilError: Attempted to use PyProxy when Python GIL not held
```

- [x] Add / update tests


hoodmane:
We could also consider getting rid of the GIL entirely, but it seems like more effort.

ryanking13 (approved):
Thanks, @hoodmane!

> We could also consider getting rid of the GIL entirely, but it seems like more effort.

Probably. But, I'm a bit reluctant if it requires a patch for CPython that can't be upstreamed.

PR #3720, opened 2023-03-31
Using esbuild instead of rollup (take 2)

hoodmane:
This is @WebReflection's #3679 with an extra commit added to fix the failing test.


rth (approved):
Thanks!

PR #3723, opened 2023-03-31
export python_stdlib.zip in package.json

dlech:
Tools like webpack can't resolve files that are not exported, so we need to export `python_stdlib.zip` along with the other binary files.

### Description

It looks like v0.23.0 changed which binary files are needed (removed `pyodide_py.tar` and `pyodide.asm.data` and added `python_stdlib.zip`). The old files were removed from the exports but the new file was not added to the exports.

### Checklists

### DIFF
--- a/docs/project/changelog.md
+++ b/docs/project/changelog.md
@@ -13,6 +13,13 @@ myst:
 
 # Change Log
 
+## Unreleased
+
+### Deployment
+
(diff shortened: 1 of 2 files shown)


hoodmane (approved):

hoodmane:
Thanks!

ryanking13:
Thanks! I added it to `files` but forgot to update `exports`.

PR #3725, opened 2023-03-31
Update rebound and reboundx

hannorein:
### Description

This PR updates the REBOUND package in pyodide.

### DIFF
--- a/packages/rebound/meta.yaml
+++ b/packages/rebound/meta.yaml
@@ -1,6 +1,6 @@
 package:
   name: rebound
-  version: 3.19.8
+  version: 3.24.1
   top-level:
     - rebound
 build:
@@ -12,8 +12,8 @@ build:
     cp $WHEELDIR/librebound.cpython-*.so $WASM_LIBRARY_DIR/lib/
 
 source:
-  url: https://files.pythonhosted.org/packages/b9/e4/aae78477fc9462ccdf2017ae1fa38c7048cdac0c2af60cff57f69c019cc9/rebound-3.19.8.tar.gz
-  sha256: 827c4f83785f129cbcaba46fd57f62d3a98af231d0ec8e5af3aa9302c424ffa6
+  url: https://files.pythonhosted.org/packages/a4/7e/919165d478025fb8421b1d692a2aa9bfb8876acae2695276e930f4382606/rebound-3.24.1.tar.gz
+  sha256: aebda40639b3bfbbee783320fca9e9267350ab5e258e52a2c46d0e248fa5a5e1
 requirements:
   run:
     - numpy
--- a/packages/rebound/meta.yaml
+++ b/packages/rebound/meta.yaml
@@ -22,4 +22,3 @@ about:
   PyPI: https://pypi.org/project/rebound
   summary: An open-source multi-purpose N-body code
   license: GPL
-# Note: this package needs to be manually upgraded to the next version
--- a/packages/rebound/meta.yaml
+++ b/packages/rebound/meta.yaml
@@ -1,6 +1,6 @@
 package:
   name: rebound
-  version: 3.24.1
+  version: 3.24.2
   top-level:
     - rebound
 build:
@@ -12,8 +12,8 @@ build:
     cp $WHEELDIR/librebound.cpython-*.so $WASM_LIBRARY_DIR/lib/
 
 source:
-  url: https://files.pythonhosted.org/packages/a4/7e/919165d478025fb8421b1d692a2aa9bfb8876acae2695276e930f4382606/rebound-3.24.1.tar.gz
-  sha256: aebda40639b3bfbbee783320fca9e9267350ab5e258e52a2c46d0e248fa5a5e1
+  url: https://files.pythonhosted.org/packages/5c/66/7564ac591bb088d7f35a59d560b5c26426bb8ef5415522232819fe11c45a/rebound-3.24.2.tar.gz
+  sha256: a3c5d4c6a10b9c1538e1051edcd6341ef3d13142b698ee9c4a1f112d9684d803
 requirements:
   run:
     - numpy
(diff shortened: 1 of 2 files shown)


hannorein:
@dtamayo I'm trying to update REBOUND and REBOUNDx to their latest versions, but this simple [unit test](https://github.com/hannorein/pyodide/blob/rebound-3.24.1/packages/reboundx/test_reboundx.py) fails now. The error messages are not very helpful. Can you think of any recent change in REBOUND/REBOUNDx that causes this? It works if I run the code locally (without pyodide).

hoodmane:
Well the failure is on this line:
```
sim.particles[0].params["tau_mass"] = -1000
```
The failure is on this line:
```
        param_type = clibreboundx.rebx_get_type(self.rebx, c_char_p(key.encode('ascii')))
```
in this file:
https://github.com/dtamayo/reboundx/blob/master/reboundx/params.py

```
  File "/lib/python3.11/site-packages/reboundx/params.py", line 56 in __setitem__
  File "<exec>", line 13 in <module>
  File "/lib/python311.zip/_pyodide/_base.py", line 310 in run
  File "/lib/python311.zip/_pyodide/_base.py", line 468 in eval_code

RuntimeError: index out of bounds (in ffi_call)
```
So it's a bad function pointer. Could be due to memory corruption. Or something.
Here:
https://github.com/dtamayo/reboundx/blob/master/reboundx/params.py#L23 
the value of `parent._sim.contents.extras` is 17863 which looks like a function pointer and not a data pointer. I am not sure if it expects a function pointer or a data pointer here. The pointer points to `rebx_free_particle_ap` which is defined here:
https://github.com/dtamayo/reboundx/blob/master/src/core.c#L846
This is passed as the first argument to `rebx_get_type` which is expected to be `struct rebx_extras*` not a function pointer: 
https://github.com/dtamayo/reboundx/blob/master/src/core.c#L1040
hence the failure. Question is why `extras` has the wrong value.

hoodmane:
@hannorein I don't see a 3.24.1 tag on https://github.com/hannorein/rebound/. Where do I find the commit for 3.24.1?

hoodmane:
So the problem is in the behavior of `reboundx.Extras`:
```
    print("extras", sim.extras)
    rebx = reboundx.Extras(sim)
    print("extras", sim.extras)
```
prints:
```
extras None
extras 17863
```

hoodmane:
It must be a memory layout mismatch? When we try to read `sim->extras` we get `sim->free_particle_ap`?

hannorein:
@hoodmane Thanks. That helps a lot - but I also do not understand yet why there is a mismatch - and why it only appears here and not when using a normal python environment. (the latest version of REBOUND is now on GitHub - I forgot to push)

hoodmane:
I ran into the problem that the simulation struct is very large and the extras pointer is at the very end so it's hard to count the offset and comparing it to the offer ctypes expects and what's at that memory location.

Maybe try moving extras to the beginning of the structure and see if the error moves? Are there any external constraints on the memory layout?

hannorein:
You're right, the struct is very large. And I'm pretty sure moving `extras` to the beginning will solve the issue for this variable, but then some other variable will have the same problem. I can see two possible explanations:

1) For some reason REBOUNDx is using the wrong rebound.h file. That would make sense locally, if a previous version of the file exists in some cache. But I think that's unlikely to occur during the CI which starts from scratch. 
2) Both REBOUND and REBOUNDx use the same rebound.h file, but for some reason the alignment is different for the structure. A different compiler flag for the two libraries might do that. But I don't see anything obvious.

hoodmane:
Well you also use ctypes so it could also be a mismatch between the offset that ctypes expects and the actual offset. 
If you were really really unlucky you could be hitting one of the bugs in my libffi implementation -- there are some test failures in the libffi test suite, though none in the ctypes test suite (they've since been fixed but I haven't updated the implementation we use #3629). I think this is pretty unlikely though.

hannorein:
Thanks to your help, I found one bug in the ctypes description of the structure (a double was listed as a pointer to a double). Let's see if that fixes it...

hoodmane:
Ah that could do it: on 64 bit systems `double*` and `double` are the same size. On 32 bit systems they are not.

hannorein:
This indeed might have fixed the issue. The tests still fail, but it seems unrelated to the two packages unless I'm missing something.

hoodmane:
Safari tests are just flaky. Thanks @hannorein!

hoodmane:
Can you remove:

> # Note: this package needs to be manually upgraded to the next version

From the two meta.yaml files? Then I will merge.

hannorein:
Done. Thanks a lot for your help! 

I'll need to setup a way to run the unit tests on 32 bit systems... then I would have caught this issue before!

hoodmane (approved):

hoodmane:
> I'll need to setup a way to run the unit tests

I can provide help with this. I recommend starting with the numpy workflow. I just opened a pr to update it for Pyodide 0.23, you can see the worflow here:
https://github.com/hoodmane/numpy/blob/pyodide-0.23.0/.github/workflows/emscripten.yml

hoodmane:
Screenshot from 2023-04-01 18-00-24

hannorein:
Thank you. I'll look into it..

hannorein:
FYI, I'm going through all of these now. Most of them are just missing the target accuracy by a tiny bit. This is because the math libraries give slightly different results on the different architecture and I've probably been a bit too strict on the unit tests. So far no other memory issues.

hannorein:
Yes, cffi would have some nice features. But right now REBOUND has no dependencies (as ctypes is part of the standard library) -- I really like that ...

hannorein:
I just noticed that in the latest release from three days ago (0.23.1), REBOUND is still at version 3.19.8, despite this PR which pushes it to 3.24.1. Why has it not updated?

ryanking13:
Normally, we only include bug fixes in a patch release.

hannorein:
Oh, I didn't realize that. Thank you.

PR #3727, opened 2023-04-01
Fix passing compression-level in `pyodide py-compile` CLI

rth:
The `compression-level` wasn't passed through in the internal functions when the target is a single archive.

(Sorry  for git history,  rebased on the wrong branch)

### DIFF
--- a/docs/project/changelog.md
+++ b/docs/project/changelog.md
@@ -22,7 +22,7 @@ myst:
 
 ### CLI
 
-- {{ Fix }} Fix `pyodide py-compress` not to ignore the `--compression-level` option when applied on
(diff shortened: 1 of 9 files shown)


docs/project/changelog.md
```diff
@@ -20,6 +20,12 @@ myst:
 - {{ Fix }} Export `python_stdlib.zip` in `package.json`.
   {pr}`3723`
 
+### CLI
+
+- {{ Fix }} Fix `pyodide py-compress` not to ignore the `--compression-level` option when applied on
```
ryanking13, line 25:
```suggestion
- {{ Fix }} Fix `pyodide py-compile` not to ignore the `--compression-level` option when applied on
```

ryanking13 (approved):
Thanks!

Issue #3672, opened 2023-03-19
Create .pyi code stubs for pyodide, pyodide.ffi, pyodide.http, etc


simonegiacomelli:
## 🚀 Feature

Create stub for pyodide, pyodide.ffi, pyodide.http, etc

### Motivation

I often find myself wanting for ide completion and automatic assisted imports. 
Right now the ide underlines pyodide imports as errors.

### Pitch

I already did pyi both for pyodide and for the web-dom, so I have some experienge. 
I would be happy to contribute with some pyodide pyi infrastructure.

### Additional context

I'm building a library that facilitates the use of pyodide and the communication between python/pyodide and python/server so I often use pyodide.ffi and related imports.

rth:
We do have type annotations for those modules already, so shouldn't it be just a matter of exposing them to IDE: maybe we need add the `py.typed` file?

But yes, if you want to work on making the situation better that would be certainly very welcome!

Generating Python types for JS objects accessed via the API, would be a more ambitious project, the inverse of https://github.com/pyodide/pyodide/issues/2162

hoodmane:
I think all you need to do is `pip install pyodide_py` and you get type annotations. The main problem is that this feature is completely undocumented.

hoodmane:
I think all you need to do is `pip install pyodide_py` and you get type annotations. The main problem is that this feature is completely undocumented.

hoodmane:
Though also I'm not sure if `js.pyi` is included in `pyodide-py`. It ought to be...

hoodmane:
> I would be happy to contribute with some pyodide pyi infrastructure.

Thanks! We don't need pyi files for the pyodide package itself since we already have types for it. But if you are interested in adding more apis to `js.pyi` that would be fantastic.

simonegiacomelli:
> I think all you need to do is `pip install pyodide-py` and you get type annotations. The main problem is that this feature is completely undocumented.

@hoodmane Thanks! I didn't know the existance of `pyodide-py`.

I tried the package and I noticed that some imports, e.g., create_proxy refers to `_pyodide._core_docs`.
Looking through the files I saw only .py and no .pyi. 

Is this package needed for python execution somewhere?

hoodmane:
Yeah the file`js.pyi` should be in there but it isn't currently. You can find it here: https://github.com/pyodide/pyodide/blob/main/src/py/js.pyi

simonegiacomelli:
[...]

PR #3731, opened 2023-04-02
Include js typeshed in pyodide_py module

hoodmane:
This adds a the js typeshed to pyodide_py.

One question is what happens if people want to install their own more detailed typesheds #3672. We may prefer to upload the typesheds as a separate pypi package so that people can substitute their alternatives.


src/py/setup.py
```diff
@@ -1,5 +1,18 @@
...
 
-if __name__ == "__main__":
-    setuptools.setup()
+from setuptools import setup
+
+
+def find_stubs(package):
+    stubs = []
+    for root, _dirs, files in os.walk(package):
+        for file in files:
```
rth, line 10: Can probably use,
```
for file in Path(package).glob("**/*")
````

File: src/py/setup.py [hunk 2]
```diff
@@ -1,5 +1,18 @@
...
+            path = os.path.join(root, file).replace(package + os.sep, "", 1)
```
rth, line 11: Can you explain why we are doing the replace here?

hoodmane: I copied this from the typeshed pep.

rth: OK, but what does it do ? ^^

I mean we have a single `js-stubs/__init__.pyi` file so do we actually need this?

hoodmane: This could probably be simplified a lot yeah. I think just putting in a pathlib glob would be a good idea since we might want to add more typeshed files in the future and it would confuse people if they weren't included...

rth: Yes. Also  I'm a bit confused that this replace seems to be constructing a path of the form,
```
../py/__init__.pyi
```
which doesn't  exist. Or at least maybe it's a side effect of `os.walk`, if we go with pathlib glob, it might just ben,
```
package_data = {"js-stubs": Path(package).glob("**/*")}
```

rth:
Do we still need this with https://github.com/pyodide/webtypy ?

PR #3737, opened 2023-04-04
Simplify build environment initialization

### LINKED ISSUE #3722: Broken pyodide build /abs/path/to/module in 0.23.0
## 🐛 Bug

My build pipeline started failing after upgrading to 0.23.0.
It complains about the `PYODIDE_EMSCRIPTEN_VERSION` env variable if the cwd is different from the `[SOURCE_LOCATION]` argument.

I used to build my extensions with an absolute path to them:

```
pyodide build /abs/path/to/module
```

If the cwd is changed to `/abs/path/to/module` then the Key error for the `PYODIDE_EMSCRIPTEN_VERSION` is not longer present. And the very same command (with the abs path in the arguments) is executed, the build passes.

Here is the stack trace:

```
#0 0.664 │ /usr/local/lib/python3.11/site-packages/pyodide_build/common.py:199 in       │
#0 0.664 │ get_make_flag                                                                │
#0 0.664 │                                                                              │
#0 0.664 │   196 │   │   SIDE_MODULE_CFLAGS                                             │
#0 0.664 │   197 │   │   SIDE_MODULE_CXXFLAGS                                           │
#0 0.664 │   198 │   """                                                                │
#0 0.664 │ ❱ 199 │   return get_make_environment_vars()[name]                           │
#0 0.664 │   200                                                                        │
#0 0.664 │   201                                                                        │
#0 0.664 │   202 def get_pyversion() -> str:                                            │
#0 0.664 ╰──────────────────────────────────────────────────────────────────────────────╯
#0 0.664 KeyError: 'PYODIDE_EMSCRIPTEN_VERSION'
```

### To Reproduce

Uncomment the `WORKDIR /zengl/` to get the same build passing.

```Dockerfile
FROM python:3.11.2
RUN python3 -m pip install pyodide-build==0.23.0
RUN git clone https://github.com/emscripten-core/emsdk.git &&\
    /emsdk/emsdk install `pyodide config get emscripten_version` &&\
    /emsdk/emsdk activate `pyodide config get emscripten_version`

RUN wget -q https://f003.backblazeb2.com/file/zengl-data/examples/web-example-assets.tar.gz

RUN git clone https://github.com/szabolcsdombi/zengl /zengl
# WORKDIR /zengl/
RUN echo 'source /emsdk/emsdk_env.sh && pyodide build /zengl/' | bash
```

### Expected behavior

the CWD should not affect the lookup of the env variables.
the CWD should not affect the build when an absolute path is given.

### Environment

- Pyodide Version 0.23.0
- Browser version not relevant
- pyodide-build==0.23.0
- emscripten_version 3.1.32

Build environment: Docker official Python 3.11.2 image

ryanking13:
### Description

Resolve #3722

This PR simplifies build environment initialization.

Merges two private pyodide-build APIs `init_environment` and `initialize_pyodide_root`
into a single a single API: `init_environment`.
Regardless of whether we are doing in-tree or out-of-tree build, `init_environment` will do the correct thing.

### Checklists

- [ ] Add / update tests


pyodide-build/pyodide_build/tests/test_build_env.py
```diff
@@ -0,0 +1,106 @@
...
+@pytest.fixture(scope="function", autouse=True)
+def reset_env_vars():
+    # Will reset the environment variables to their original values after each test.
+
+    old_environ = dict(os.environ)
+
+    yield
+
+    os.environ.clear()
+    os.environ.update(old_environ)
```
hoodmane, line 18:
```suggestion
    try:
       yield
    finally:
       os.environ.clear()
       os.environ.update(old_environ)
```
I think is needed so that cleanup is run even on error

pyodide-build/pyodide_build/build_env.py
```diff
@@ -0,0 +1,230 @@
...
+    init_environment()
+    return Path(os.environ["PYODIDE_ROOT"])
+
+
+def in_xbuildenv() -> bool:
+    pyodide_root = get_pyodide_root()
+    return pyodide_root.name == "pyodide-root"
+
+
+def search_pyodide_root(curdir: str | Path, *, max_depth: int = 5) -> Path:
```
rth, line 76: (note to self): Copied without changes from `common.py`

pyodide-build/pyodide_build/build_env.py
```diff
@@ -0,0 +1,230 @@
...
+        if "tool" in configs and "pyodide" in configs["tool"]:
+            return base
+
+    raise FileNotFoundError(
+        "Could not find Pyodide root directory. If you are not in the Pyodide directory, set `PYODIDE_ROOT=<pyodide-root-directory>`."
+    )
+
+
+@functools.cache
+def get_build_environment_vars() -> dict[str, str]:
```
rth, line 107: Previously was: `common.set_build_environment`

pyodide-build/pyodide_build/build_env.py
```diff
@@ -0,0 +1,230 @@
...
+    if "PYODIDE_ROOT" in os.environ:
+        return
+
+    # 2) If we are doing an in-tree build,
+    #    set PYODIDE_ROOT to the root of the Pyodide repository
+    try:
+        os.environ["PYODIDE_ROOT"] = str(search_pyodide_root(Path.cwd()))
+        return
+    except FileNotFoundError:
+        pass
```
rth, line 178: This may not be a good idea. If we are doing in tree build, and for some reason that file is not found, it should be a hard failure mentioning that either the wrong command was used by the user  or that it was run from the wrong folder.  We shouldn't try to download `xbuildenv` zip in that case. This means that this command does need to know whether we are doing in tree or out-of-tree build.

pyodide-build/pyodide_build/build_env.py
```diff
@@ -0,0 +1,230 @@
...
+
+    This is not a public API, use get_build_environment_vars instead.
+    """
+
+    from .common import exit_with_stdio  # avoid circular import
+
+    PYODIDE_ROOT = get_pyodide_root()
+    environment = {}
+    result = subprocess.run(
+        ["make", "-f", str(PYODIDE_ROOT / "Makefile.envs"), ".output_vars"],
```
rth, line 211: Unrelated to this PR: WDYT if we move part of the configuration intended to be modifiable by the user from `Makefile.envs` to some toml file and have some script that sets those env variables if needed. Not in this PR. Just to try to move away from makefiles?

pyodide-build/pyodide_build/install_xbuildenv.py
```diff
@@ -112,8 +118,16 @@ def install(path: Path, *, download: bool = False, url: str | None = None) -> No
     from . import __version__
 
     version = __version__
+
     if download:
```
rth, line 122: A better default behavior might be to check if it's downloaded already maybe? To avoid re-downloading this at each `pyodide build` command.

pyodide-build/pyodide_build/build_env.py
rth: Could you please check that this includes changes from https://github.com/pyodide/pyodide/pull/3744 ?

rth (commented):
Thanks @ryanking13 . A couple of comments are below. Overall LGTM but I feel it's also a quite large refactoring to be included as a bug fix. 

Also I don't know what's our definition of public API for pyodide_buid but technically `pyodide_build.common.search_pyodide_root` is public API and moving it to `build_env` is a breaking change if anyone uses it. It's probably fine to do in a major release since in practice we make no guarantees about any of the Python API for pyodide_build. Bit it probably still not a good idea to put it in a bug fix release. 

To make it clearer, probably also doesn't hurt to make `pyodide_build/_build_env` private.


There is no smaller fix we could do for the parent issue (and merge this for the next major release)?

ryanking13:
> There is no smaller fix we could do for the parent issue (and merge this for the next major release)?

Sure I will split this into smaller ones. This PR actually got a lot bigger than I first thought :)

pyodide-build/pyodide_build/build_env.py
ryanking13: I think, for now, setting env variables e.g. `WASM_LIBRARY_DIR=... pyodide build recipes` will overwrite Makefile.envs.

But yes, I do want to move some of the configurations in Makefile.envs  that are not used in building the main module to somewhere else that users can easily modify.

pyodide-build/pyodide_build/build_env.py
ryanking13: Yes, probably we should discuss more about how to distinguish in-tree / out-of-tree build. Though,  personally, I think we should go in a direction to merge in-tree / out-of-tree build (which is one of my intention in this PR).

For now, this function is copied from [initialize_pyodide_root](https://github.com/pyodide/pyodide/blob/b80188535396353266ba5f2499450cc695336fbb/pyodide-build/pyodide_build/out_of_tree/utils.py#L29-L39), without changing the behavior.

pyodide-build/pyodide_build/tests/test_build_env.py
ryanking13: Good point, thanks!

PR #3744, opened 2023-04-06
Fix set_build_environment so that it works even if HOME is unset

hoodmane:
I'm running into a crash in cibuildwheel where some of its CI environments don't set HOME. This fixes `set_build_environment` to be a bit more resistant against crashes due to missing environment variables.

- [x] Add / update tests

### DIFF
--- a/pyodide-build/pyodide_build/common.py
+++ b/pyodide-build/pyodide_build/common.py
@@ -29,6 +29,7 @@ from .recipe import load_all_recipes
 BUILD_VARS: set[str] = {
     "PATH",
     "PYTHONPATH",
+    "PYODIDE_JOBS",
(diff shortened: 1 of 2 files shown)


ryanking13 (approved):
Thanks, @hoodmane!

> I'm running into a crash in cibuildwheel where some of its CI environments don't set HOME. 

Acutally, do you know why we are passing `HOME` to build environments? It seems like you added this when you were working on [Rust support](https://github.com/pyodide/pyodide/pull/2378). But I don't see why we need that variable.

Anyway, this PR LGTM.

hoodmane:
> do you know why we are passing HOME to build environments? 

Yes rust looks in `~/.cargo`.

PR #3746, opened 2023-04-06
Pass the source and build directory explicitly, don't use chdir

hoodmane:
This switches to passing the source and build directories as arguments. It adds an output-directory argument to pyodide build allowing us to indicate where the output wheel should go independent of the build directory. I also did some cleanup of the logic added in #3310

### Checklist

- [ ] Add / update tests

### DIFF
--- a/pyodide-build/pyodide_build/cli/build.py
+++ b/pyodide-build/pyodide_build/cli/build.py
@@ -1,4 +1,3 @@
-import os
 import re
 import shutil
 import tempfile
@@ -21,6 +20,7 @@ from ..out_of_tree.utils import initialize_pyodide_root
 
 def pypi(
     package: str,
+    output_directory: Path,
     exports: str = typer.Option(
         "requested",
         help="Which symbols should be exported when linking .so files?",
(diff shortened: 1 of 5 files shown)


pyodide-build/pyodide_build/cli/build.py
```diff
@@ -127,6 +112,10 @@ def main(
         "",
         help="Build source, can be source folder, pypi version specification, or url to a source dist archive or wheel file. If this is blank, it will build the current directory.",
     ),
+    output_directory: str = typer.Option(
```
ryanking13, line 115:
```suggestion
    output_directory: Path = typer.Option(
```

ryanking13 (approved):
Sorry that I missed this!

pyodide-build/pyodide_build/cli/build.py
```diff
@@ -31,30 +31,37 @@ def pypi(
...
-        os.chdir(tmpdir)
-        built_wheel = build.run(exports, backend_flags, outdir=curdir / "dist")
-        os.chdir(curdir)
+        built_wheel = build.run(srcdir, output_directory, exports, backend_flags)
         return built_wheel
 
 
+def download_url(url: str, output_directory: Path) -> str:
+    with requests.get(url, stream=True) as response:
+        urlpath = Path(urlparse(response.url).path)
```
rth, line 52: Technically a PurePosixPath if anyone tries to run this on windows.

rth (commented):
Thanks!

pyodide-build/pyodide_build/cli/build.py
```diff
@@ -31,30 +31,37 @@ def pypi(
...
+        if urlpath.suffix == ".gz":
+            urlpath = urlpath.with_suffix("")
+        file_name = urlpath.name
+        with open(output_directory / file_name, "wb") as f:
+            for chunk in response.iter_content(chunk_size=1 << 20):
```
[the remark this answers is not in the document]
hoodmane: Or maybe `megabyte = 1 << 20`. It's a great math fact that 10^3 is approximately 2^10.

pyodide-build/pyodide_build/cli/build.py
hoodmane: Okay, so it's relying on some decorator magic. Slightly concerning but okay.

hoodmane: So if you say:
```
a : A = typer.Option("blah")
```
does it try to set `a = A("blah")`? Or is it using a table of known conversions or something?

rth: Looking at https://github.com/tiangolo/typer/issues/77 looks like there is a pre-defined list of types. But Path and dates are the most useful.

ryanking13: Yes, there are some [pre-defined list of types](https://typer.tiangolo.com/tutorial/parameter-types/path/) that typer can handle. I mean, it is not critical so we can just use the type `str` and handle it manually if you feel concerned or if we don't want to rely too much on internal features of typer or click.

hoodmane: No I'm good with it thanks for the suggestion and for explaining!

PR #3748, opened 2023-04-07
ENH Add then, catch, and finally_ to the Tasks too

hoodmane:
I only just realized that we only added these methods to our Futures not to our tasks. `asyncio.Task` is a subclass of `asyncio.Future` so we should use a `PyodideTask` which is a subclass of `PyodideFuture`.

- [x] Add / update tests

### DIFF
--- a/src/py/pyodide/webloop.py
+++ b/src/py/pyodide/webloop.py
@@ -4,7 +4,7 @@ import inspect
 import sys
 import time
 import traceback
-from asyncio import Future
+from asyncio import Future, Task
 from collections.abc import Awaitable, Callable
 from typing import Any, TypeVar, overload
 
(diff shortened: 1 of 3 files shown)


src/py/pyodide/webloop.py
```diff
@@ -167,6 +167,10 @@ def wrapper(fut: Future[T]) -> None:
         return result
 
 
+class PyodideTask(Task[T], PyodideFuture[T]):
```
rth, line 170: I'm wondering if we should have called it just `Future` and  `Task` so you could do,
```
from pyodide.webloop import Future, Task
```
as a drop-in replacement of whatever asyncio user had. Probably too late now anyway (and I'm not 100 % that was a good idea in the first place).

rth: In the docstring it would be good to explain when would one want to use this instead of `asyncio.Task`  also shouldn't we expose this in the docs as well?

rth (approved):

src/py/pyodide/webloop.py
hoodmane: The user isn't really supposed to use it directly. The documentation for `Task` says:

> Manual instantiation of Tasks is discouraged.

Yeah it's a good idea to add a bit of documentation.

PR #3750, opened 2023-04-09
More improvements to JsProxy_compute_typeflags

hoodmane:
This is a followup to #3740 which rearranges the logic a bit more on further consideration.

I found out that if you create a revocable Proxy and then revoke it, a huge number of browser
internal operations raise TypeErrors. Such an object is thus an excellent stress test for our
handling of explosive objects. Overall we do quite well, `isPyProxy` and `compute_type_flags`
are actually the only spots where we run into trouble (and `PyProxy_Check` is the only spot
that is raising a fatal error).

This changes it so that an object with a `length` getter that throws an error or returns
not an integer will have a `__len__` in Python that always throws an error. It can't check
the type anymore because it avoids invoking getters so if the result is a return value from
a getter it won't know the type. The same is true for the `Error.name`, `Error.message`, and
`Error.stack` fields.

To check for methods, I still use `typeof x[method] === "function"`, invoking the getter.
Usually functions don't have getters anyways and I think it is weirder to have false positives
in this case. If there is a getter and it raises an error, we catch it and throw it away.

### Checklist

- [x] Add / update tests

### DIFF
--- a/src/core/jsproxy.c
+++ b/src/core/jsproxy.c
@@ -4006,10 +4006,10 @@ EM_JS_NUM(int, JsProxy_compute_typeflags, (JsRef idobj), {
   // we'll run into trouble. I think that for this to happen, someone would have
   // to pass in some weird and maliciously constructed object. Anyways for
   // maximum safety, we double check that only one of these is set.
-  SET_FLAG_IF(IS_CALLABLE, typeof obj === "function")
-  SET_FLAG_IF_HAS_METHOD(IS_AWAITABLE, "then")
-  SET_FLAG_IF_HAS_METHOD(IS_ITERABLE, Symbol.iterator)
-  SET_FLAG_IF_HAS_METHOD(IS_ASYNC_ITERABLE, Symbol.asyncIterator)
+  SET_FLAG_IF(IS_CALLABLE, typeof obj === "function");
+  SET_FLAG_IF_HAS_METHOD(IS_AWAITABLE, "then");
+  SET_FLAG_IF_HAS_METHOD(IS_ITERABLE, Symbol.iterator);
+  SET_FLAG_IF_HAS_METHOD(IS_ASYNC_ITERABLE, Symbol.asyncIterator);
   SET_FLAG_IF(IS_ITERATOR, hasMethod(obj, "next") && (hasMethod(obj, Symbol.iterator) || !hasMethod(obj, Symbol.asyncIterator)));
   SET_FLAG_IF(IS_ASYNC_ITERATOR, hasMethod(obj, "next") && (!hasMethod(obj, Symbol.iterator) || hasMethod(obj, Symbol.asyncIterator)));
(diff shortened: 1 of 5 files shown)


src/core/pre.js
```diff
@@ -6,3 +6,39 @@ API.version = "0.24.0.dev0";
...
+ *
+ * Observe whether a property exists or not without invoking getters.
+ * Should never throw as long as arguments have the correct types.
+ *
+ * obj: an object
+ * prop: a string or symbol
+ */
+function hasProperty(obj, prop) {
+  while (obj) {
+    if (Object.getOwnPropertyDescriptor(obj, prop)) {
```
rth, line 21: Why not use hasOwn here?

src/core/pre.js
```diff
@@ -6,3 +6,39 @@ API.version = "0.24.0.dev0";
...
+ * Safe property check
+ *
+ * Observe whether a property exists or not without invoking getters.
+ * Should never throw as long as arguments have the correct types.
+ *
+ * obj: an object
+ * prop: a string or symbol
+ */
+function hasProperty(obj, prop) {
+  while (obj) {
```
rth, line 20: It's a bit crazy there is no easier way to do this in JS.

src/core/jsproxy.c
```diff
@@ -3982,14 +3982,14 @@ JsProxy_get_subtype(int flags)
 }
 
 #define SET_FLAG_IF(flag, cond)                                                \
-  try {                                                                        \
-    if (cond) {                                                                \
-      type_flags |= flag;                                                      \
-    }                                                                          \
-  } catch (e) {                                                                \
-    IF_DEBUG(console.warn(`Error raised in SET_FLAG_IF(#flag, #cond)`));       \
```
rth, line 3990: Can you comment it's no longer a concern?

rth (commented):
Thanks! Overall LGTM. CI is failing though.

src/core/pre.js
hoodmane: Too recent:
https://caniuse.com/?search=hasOwn

src/core/pre.js
hoodmane: Yes...

src/core/jsproxy.c
hoodmane: I will write more tests that pass various horrible stuff through this code to make sure it never throws. I guess the comment should be "The condition must never raise an error!"
But in fact it's better for tests if this fails than if it gets swallowed as @ryanking13 points out.

hoodmane:
Oh I think I found the most horrible JavaScript object yet: create a revokable proxy and revoke it. If Pyodide touches it it raises a fatal error.

src/core/jsproxy.c
hoodmane: Ok `test_revoked_proxy` makes every single step here throw and catch an error.

PR #3751, opened 2023-04-10
Make sure PYODIDE_ROOT points to absolute path in out-of-tree build

ryanking13:
### Description

Resolve: #3722

This is a minimal fix for #3722 to include in 0.23.1

### Checklists

### DIFF
--- a/docs/project/changelog.md
+++ b/docs/project/changelog.md
@@ -26,6 +26,11 @@ myst:
   a single file.
   {pr}`3727`
 
+### Build System
+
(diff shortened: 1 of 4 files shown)


hoodmane:
What about my #3746? I think that would also fix it. Though these are probably both good: don't change the working directory and make the path absolute. You could fix it with either change but why not both?

ryanking13:
@hoodmane Oh, I think I didn't see your work in #3746 :) Both changes seem reasonable, so I think we can accept both.

hoodmane (approved):

PR #3752, opened 2023-04-10
Make executables installed in Pyodide venv run in Pyodide not in host

hoodmane:
- [x] Add a test


ryanking13:
What exactly does changing sys.executable do?

hoodmane:
The issue is that pip can't run inside of Pyodide because it needs to do stuff like make requests that don't work inside the virtual environment. So the virtual environment `bin` folder contains a `python-host` executable which is a symlink to the `sys.executable` that created the virtual environment. The shebang for pip uses `python-host`. But then when pip installs an executable it uses `sys.executable` to create the shebang for the installed executable. If that is `python-host` then the installed executable will run outside of Pyodide. For instance, `pytest` will run not in Pyodide but in the host python. This is not really what we want.

ryanking13 (approved):
Thanks for the explanation! Makes sense.

hoodmane:
This seems to have broken the test. I didn't realize it wasn't running because of the opencv failure.

PR #3755, opened 2023-04-11
Move mypy to the end of pre-commit

hoodmane:
Mypy is much slower than any of the other checks so it should go last.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -27,33 +27,6 @@ repos:
     hooks:
       - id: black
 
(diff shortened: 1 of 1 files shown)


cclauss:
There is lots of talk about whether mypy should even be run in pre-commit:
* https://github.com/pre-commit/mirrors-mypy
* https://github.com/pre-commit/pre-commit/issues/1580
* https://jaredkhan.com/blog/mypy-pre-commit

I have started to remove if from pre-commit in projects but would be interested in your perspective.

PR #3765, opened 2023-04-13
Make build variables accessible only through the API

ryanking13:
### Description

This is a split-off from #3737.

This PR makes build variables only accessible through the API, instead of polluting os.environ.

I also changed the API as below, as a preparation to move away from makefiles.

- `get_make_environment_vars` => `get_build_environment_vars`
- `get_make_flag` => `get_build_flag`

### Checklists

- [x] Add / update tests


pyodide-build/pyodide_build/common.py
```diff
@@ -335,11 +360,8 @@ def get_pyodide_root() -> Path:
 
 @functools.cache
 def get_unisolated_packages() -> list[str]:
-    import json
-
-    if "UNISOLATED_PACKAGES" in os.environ:
```
hoodmane, line 340: This will make us recompute `unisolated_packages` by scanning every `meta.yaml` file once for every compilation command? There's a reason for this caching logic... `load_all_recipes` is expensive and `functools.cache` doesn't work across subprocesses.

I think we should add a cross-process cache function for this and `get_build_environment_vars` that caches into an environment variable.

pyodide-build/pyodide_build/out_of_tree/pypi.py
```diff
@@ -336,7 +336,7 @@ def _resolve_and_build(
     requirements = []
 
     target_env = {
-        "python_version": f'{common.get_make_flag("PYMAJOR")}.{common.get_make_flag("PYMINOR")}',
+        "python_version": f'{common.get_build_flag("PYMAJOR")}.{common.get_build_flag("PYMINOR")}',
```
hoodmane, line 339: While you're at it why not add a function for this to common?

.circleci/config.yml
```diff
@@ -300,6 +300,7 @@ jobs:
               export CACHE_DIR=".test_cache/<< parameters.cache-dir >>"
             fi
             echo "pytest cache dir: $CACHE_DIR"
+            source pyodide_env.sh
```
hoodmane, line 303: Why do we need to source this now?

pyodide-build/pyodide_build/common.py
ryanking13: Good point. I didn't realize that this was for subprocesses. Will revert this change.

pyodide-build/pyodide_build/out_of_tree/pypi.py
ryanking13: Sounds good.

.circleci/config.yml
ryanking13: Since this PR makes os.environ no longer being updated, the PATH value will not be updated either, so we need to tell it where to find the emscripten compiler.

It is worth noting that previously pyodide-build tests relied on some implicit behavior of _get_pyodide_root()_ to update PATH:

1. _get_pyodide_root()_ calls _init_environment()_ which calls _get_make_environment_vars()_
1. _get_make_environment_vars()_ will call `make -f Makefile.envs .output_vars` which runs _pyodide_env.sh_
2. _init_environment()_ [updates os.environ](https://github.com/pyodide/pyodide/blob/4a0ca4ed48ff109679df4e246e8ed79901099f37/pyodide-build/pyodide_build/common.py#L317) (this PR removes it), so tests that require emscripten compilers can run successfully.

I think this is a very error-prone, so I changed it to explicitly source `pyodide_env.sh` before running tests that require emscripten compilers. In the future, I would like to make pyodide-build install emscripten itself (https://github.com/pyodide/pyodide/issues/3287)

pyodide-build/pyodide_build/common.py
ryanking13: > every compilation command?

Wait, but is this true? I think we are calling it [once per each package when installing requirements](https://github.com/pyodide/pyodide/blob/4a0ca4ed48ff109679df4e246e8ed79901099f37/pyodide-build/pyodide_build/pypabuild.py#L104).

hoodmane: Ah, yes. `pywasmcross.py` looks very clean actually: only thing it uses from the rest of the package is `_f2c_fixes.py`.

ryanking13:
@hoodmane Can we merge this?

hoodmane:
I think so. Sorry for being slow to review again... Lemme look it over again.

hoodmane (approved):

ryanking13:
Thanks for the review!

Issue #3759, opened 2023-04-12
Upload top-of-tree xbuildenv to S3


ryanking13:
When testing pyodide-build out-of-tree build feature, I often need to manually download xbuildenv.

```sh
pip install "git+https://github.com/pyodide/pyodide/@main#subdirectory=pyodide-build"
pyodide xbuildenv install --download --url "<xbuildenv_url>"
pyodide build
```

For now, we don't upload top-of-tree xbuildenv anywhere, so I use circle CI build artifacts. For example: [here](https://github.com/ryanking13/pyodide-recipes-mirror/blob/8c5bb491ca4c43059f9eb31b9fb92764c215cba8/.github/workflows/build.yml#L36-L37).

So I think it would be useful to upload top-of-tree xbuildenv archive, so we can download it from fixed URL.
Maybe into `https://cdn.jsdelivr.net/pyodide/dev/xbuildenv.zip` or `https://cdn.jsdelivr.net/pyodide/dev/xbuildenv/xbuildenv.zip`?

rth:
Sounds good to me. We did setup another"cache" S3 bucket initially, more for experimentation where we can also upload things. But the main bucket in the dev folder probably makes more sense in this case.

PR #3766, opened 2023-04-13
Deploy top-of-tree xbuildenv and pyodide-core to S3

ryanking13:
Resolve: #3759

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -568,7 +568,8 @@ jobs:
             tar cjf dist/pyodide-core.tar.bz2 pyodide/pyodide{.js,.mjs,.asm.js,.asm.wasm} pyodide/{package,repodata}.json pyodide/python_stdlib.zip
             tar cjf dist/xbuildenv.tar.bz2  xbuildenv/
 
-            python3 tools/deploy_s3.py dist/ "dev/xbuildenv/" --bucket "pyodide-cdn2.iodide.io" --cache-control 'max-age=3600, public' --rm-remote-prefix
+            python3 tools/deploy_s3.py dist/ "xbuildenv/dev" --bucket "pyodide-cache" --cache-control 'max-age=3600, public' --overwrite \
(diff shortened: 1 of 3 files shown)


.circleci/config.yml
```diff
@@ -556,6 +557,19 @@ jobs:
...
+      - run:
+          # Unlike the release version, we upload the dev version to S3 not to GitHub.
+          name: Deploy release files to S3
+          command: |
+            mkdir -p /tmp/ghr/dist
+            cp -r dist /tmp/ghr/pyodide
+            cp -r xbuildenv /tmp/ghr/xbuildenv
+            cd /tmp/ghr
+            tar cjf dist/pyodide-core.tar.bz2 pyodide/pyodide{.js,.mjs,.asm.js,.asm.wasm} pyodide/{package,repodata}.json pyodide/python_stdlib.zip
+            tar cjf dist/xbuildenv.tar.bz2  xbuildenv/
```
rth, line 569: Ahh, I think we haven't whitelisted this extension with jsdelivr so it won't work. 

Maybe we could put it to `pyodide-cache` bucket where it was originally,
```
python3 tools/deploy_s3.py dist/ "xbuildenv/" --bucket "pyodide-cache" --cache-control 'max-age=3600, public' --rm-remote-prefix
```
but then we need to pass it different credentials somehow. For instance maybe define `AWS_ACCESS_KEY_ID_CACHE`,  `AWS_SECRET_ACCESS_KEY_CACHE` (sent you by email) and then set them back to the variables without the `_CACHE` prefix just for the below command.

ryanking13: Thanks, I changed the bucket and added the corresponding keys.

rth (approved):
Thanks!

rth:
Just a couple of test failures. 
```
FAILED tools/tests/test_deploy_s3.py::deploy_to_s3_overwrite - TypeError: str expected, not OptionInfo
FAILED tools/tests/test_deploy_s3.py::deploy_to_s3_mime_type - TypeError: str expected, not OptionInfo
```
you need to explicitly pass all arguments in tests, as default typer arguments are then not valid.

PR #3778, opened 2023-04-18
DOC Hide disabled packages from docs

ryanking13:
Hides packages that are disabled (`_disabled: true`) from docs.

For now we have two such packages, which were disabled during Python 3.11 update: [freesasa](https://github.com/pyodide/pyodide/blob/25ba93485a781cb900ea9386b5935ad075633a39/packages/freesasa/meta.yaml#L4), [wordcloud](https://github.com/pyodide/pyodide/blob/25ba93485a781cb900ea9386b5935ad075633a39/packages/wordcloud/meta.yaml#L4).


rth (approved):
Thanks!

PR #3786, opened 2023-04-21
Fix bokeh package (add xyzservices, include pandas as runtime dependency)

bartbroere:
Currently, importing and using `bokeh` fails because of the missing `xyzservices` package. I realise it's a pure-Python package, but hopefully it's okay if it fixes one of the already included packages.

The quickest way to reproduce the error is:
```python
from bokeh.plotting import figure
```

I'll also try to add a regression test to detect when bokeh might break in the future for similar reasons.

### Checklists
- [x] Add / update tests


bartbroere:
By adding the test I found out `pandas` is a runtime dependency as well, so I included that change in the `meta.yaml` as well

+ https://app.circleci.com/pipelines/github/pyodide/pyodide/5720/workflows/9f36387b-dc94-47a9-9418-03ddcc4799ed/jobs/72066
+ https://docs.bokeh.org/en/latest/docs/first_steps/installation.html#installing-required-dependencies

docs/project/changelog.md
```diff
@@ -22,15 +22,15 @@ myst:
   present also on `Task`s as well as `Future`s.
   {pr}`3748`
 
-- {{ Enhancement }} Added `headers` property to `pyoidide.http.FetchResponse`.
+- {{ Enhancement }} Added `headers` property to `pyodide.http.FetchResponse`.
```
ryanking13, line 25: Nice catch, thanks!

packages/xyzservices/meta.yaml
```diff
@@ -0,0 +1,13 @@
+package:
+  name: xyzservices
```
ryanking13, line 2: Could you add some basic tests for xyzservices?

ryanking13 (approved):
Thanks @bartbroere

packages/xyzservices/meta.yaml
bartbroere: Yes, no problem. I'll add them to the PR.

packages/bokeh/test_bokeh.py
```diff
@@ -0,0 +1,14 @@
+def test_bokeh(selenium):
```
hoodmane, line 1:
```suggestion
from pytest_pyodide.decorator import run_in_pyodide

@run_in_pyodide(packages=["bokeh"])
def test_bokeh(selenium):
```
please look at what the other package tests do and imitate them.

packages/xyzservices/test_xyzservices.py
```diff
@@ -0,0 +1,15 @@
+def test_xyzservices(selenium):
```
hoodmane, line 1: Also needs to use @run_in_pyodide

ryanking13 (approved):
Thanks, @bartbroere!

PR #3790, opened 2023-04-23
Numpy patch release 1.24.3

bartbroere:
Let’s see if the patch release passes CI.


ryanking13 (approved):
Looks like there is no issue. Thanks @bartbroere!

PR #3794, opened 2023-04-24
Update scipy to 1.10.1

lesteve:
### Description

This updates scipy to the latest release, namely 1.10.1.

For now this is a draft, checking that this does not break the CI, this will need a bit of clean-up later.

### Checklists


hoodmane:
=) well however long we can get away without meson it's all good. I do think it would be nice to take advantage of the parallelism that I think meson allows. We'll get there eventually.

hoodmane (approved):
Thanks again for doing this @lesteve! I cannot tell you how happy I am to have someone else contributing to Scipy upkeep!

lesteve:
This one was relatively painless :wink:!

Running the scipy test suite locally it does not seem to have introduced any regression. The only thing I noticed is that there is a test that now takes 6 minutes instead of ~10 seconds. My hunch is that this is because of a scipy change that relies on floating point exception. For now I am going to skip this test and maybe investigate later in more details.

PR #3801, opened 2023-04-28
Add simplejson

bartbroere:
Add the `simplejson` package, so other packages can be more easily micropip installed if they happen to depend on this.

- [x] Add / update tests


lesteve:
simplejson is pure Python right? So you should be able to install it with micropip. 

> so other packages can be more easily micropip installed if they happen to depend on this.

I don't quite understand what you mean by this, but maybe I am missing something. In principle `await micropip.install('package-that-depends-on-simplejson')` should work too, do you have an example of package where that does not work? 



In which case, you should be able to install with `micropip`. What kind of package

bartbroere:
No, it has some (optional) C extensions for speedups, and doesn't ship a pure Python wheel, so it can't be installed with `micropip`.

If we include it in Pyodide, we enable installing some libraries which have `simplejson` as a dependency.

lesteve:
OK I was not aware of this. For some reason `await micropip.install('simplejson')` is working fine in the stable console (the wheel it is using is https://files.pythonhosted.org/packages/56/40/c58cd470a57af4affa87da639a9d9d339a0d5898d98faa608ac43f3e191e/simplejson-3.19.1-py3-none-any.whl), but  it is very likely not getting the optimised code written in C.

hoodmane:
Well it doesn't hurt to include a faster compiled version.

bartbroere:
> Well it doesn't hurt to include a faster compiled version.

Alright! The test that imports `simplejson._speedups` should ensure that we don't end up with a pure Python version without realising it. (Which is possible because the `setup.py` seems to do pure Python as a fallback).

Let's hope it passes all CI checks this time.

.pre-commit-config.yaml
```diff
@@ -76,6 +76,7 @@ repos:
           - types-docutils
           - types-pyyaml
           - types-setuptools
+          - types-simplejson
```
hoodmane, line 79: Do we really need this? If it's only for the one `test_simplejson` I'd prefer just putting `type: ignore` comments.

bartbroere: Alright, that sounds better to me as well, they would be needed only for the type checker indeed.

hoodmane (approved):

hoodmane:
Conveniently it doesn't depend on numpy so it isn't affected by #3816.

hoodmane:
Thanks @bartbroere!

PR #3802, opened 2023-04-30
Avoid compressing archives twice

ryanking13:
Follow up of #3766. This prevents compressing tar.gz / tar.bz2 archives with gzip again.

I accidentally pushed a couple of times to the upstream repo (not in the main branch) instead of my fork, sorry if I made your notifications noisy.

### DIFF
--- a/tools/deploy_s3.py
+++ b/tools/deploy_s3.py
@@ -111,15 +111,23 @@ def deploy_to_s3_main(
             raise typer.Exit()
 
         with open(file_path, "rb") as fh_in:
-            # Use gzip compression for storage. This only impacts storage on
-            # AWS and transfer between S3 and the CDN. It has no impact on the
-            # compression received by the end user (since the CDN re-compresses
-            # files).
-            fh_compressed = io.BytesIO()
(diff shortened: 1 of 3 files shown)


tools/deploy_s3.py
```diff
@@ -136,8 +144,8 @@ def deploy_to_s3_main(
 
             extra_args = {
                 "CacheControl": cache_control,
-                "ContentEncoding": "gzip",
                 "ContentType": content_type,
+                "ContentEncoding": "identity" if compressed else "gzip",
```
rth, line 148: Are you sure about this this "identity" option? I can't find any documentation about it.
Maybe just not setting that field would be easier?

rth (approved):
Thanks @ryanking13 !

tools/deploy_s3.py
ryanking13: I think you are right, [RFC](https://www.rfc-editor.org/rfc/rfc2616#section-14.11) says that it should be used only in `Accept-Encoding` header.

PR #3804, opened 2023-04-30
Don't remove build logs after build failure

### LINKED ISSUE #3770: No logs when erroring on build package
I'm trying to build a package with Cyhton sources with the "in tree" approach.

When running 

```
pyodide build-recipes sisl --install
```

I get

```
The following packages are already built: boost-cpp, distutils, hashlib, libf2c, liblzma, lzma, micropip, numpy, openblas, openssl, packaging, pydecimal, pydoc_data, pyparsing,  
scipy, setuptools, sqlite3, ssl, test, and tqdm                                                                                                                                   
Building the following packages: sisl                                                                                                                                             
[1/1] (thread 1) building sisl ⠙                                                            
Building packages... ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 0/1   0% Time elapsed: 0:00:09Error building sisl. Printing build logs.                                                                                                                                         
ERROR: No build log found.                                                                                                                                                        
ERROR: cancelling buildall                                                                                                                                                        
[1/1] (thread 1) building sisl ⠹                                                            
Building packages... ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 0/1   0% Time elapsed: 0:00:09
```

Therefore, I have no clue what's wrong. When does an error without build logs happen? Can I get some more verbose output in some way? Thanks!

I'm running this on the Docker container, after succesfully building all packages, even an old version of the package I'm trying to build.

ryanking13:
Fix #3770 (it is already closed, but anyway)

The reason why no logs are being printed when a build fails was quite interesting:

1. There is a code that deletes a build log when a build fails (This PR removes it).
2. That code was not working before #3746, since we changed the working directory into srcpath.
3. After #3746, that code started to "work", so the build log gets deleted when a build fails.

I think we definitely don't want to delete the build log. So this PR removes it.

### DIFF
--- a/pyodide-build/pyodide_build/buildpkg.py
+++ b/pyodide-build/pyodide_build/buildpkg.py
@@ -426,13 +426,7 @@ def compile(
         from .pypabuild import build
 
         outpath = srcpath / "dist"
-        try:
-            build(srcpath, outpath, build_env, backend_flags)
-        except BaseException:
-            build_log_path = Path("build.log")
-            if build_log_path.exists():
-                build_log_path.unlink()
-            raise
+        build(srcpath, outpath, build_env, backend_flags)
 
 
 def replace_so_abi_tags(wheel_dir: Path) -> None:


hoodmane (approved):

hoodmane:
Thanks!

hoodmane:
Backported to stable.

PR #3810, opened 2023-04-30
Deno smoke test

rajsite:
### Description

Adds a smoke test to the CI for pyodide Deno support.
Uses the deno documented workflow for [editing the contents of `node_modules` ](https://deno.com/manual@v1.33.1/node/npm_specifiers#--node-modules-dir-flag) to test against the local build of pyodide.

### Checklists

- Add a [CHANGELOG](https://github.com/pyodide/pyodide/blob/main/docs/project/changelog.md) entry. 
  - Should there be a Changelog entry before it is documented?
- [x] Add / update tests
- Add new / update outdated documentation.
  - Documentation will be a follow-up PR.


.github/workflows/main.yml
```diff
@@ -238,3 +238,30 @@ jobs:
...
         if: always()
+
+  test-deno:
+    runs-on: ${{ matrix.os }}
+
+    needs: [build-core]
+    strategy:
+      fail-fast: false
+      matrix:
+        os: [ubuntu-latest, macos-latest]
```
rth, line 249: Probably just linux would be OK.

rth (commented):
Thanks @rajsite ! 

I think it's fine if this PR only adds the CI job, even if it's very simple, it's better than nothing until we add deno support to pytest-pyodide. For the docs, yes if you want to write some that would be very welcome. Probably easier to do it in separate PR: you can also disable all CI build except from docs building by adding `[skip ci]` in the commit message.

Here the main issue I see is that it would be good to load `pyodide.js` from the dev deployment rather than npm. So we are actually testing the dev branch. If for instance we start a HTTP server in the backround serving `dist/` and loading it from
```
import pyodideModule from "https://localhost:<some-port>/pyodide.js";
```
that wouldn't work?

rth:
Unless you mean that by copying `dist/` to `node_modules` it's going to take local files, and  not query npm? Which then could indeed be enough for the purpose of this test.

rajsite:
> Unless you mean that by copying `dist/` to `node_modules` it's going to take local files, and not query npm? Which then could indeed be enough for the purpose of this test.

Yep! It's technically a little of both but I think it's appropriate. This chain of commands:
https://github.com/pyodide/pyodide/blob/0f5e9464fea9699c70699c3f9e242d956f063559/src/test-deno/deno.json#L3-L8

Will:
1. Populate the local node_modules from npm (with a flag telling deno to make a local node_modules directory)
2. Copy the contents of the build dist folder into node_modules/pyodide:

https://github.com/pyodide/pyodide/blob/0f5e9464fea9699c70699c3f9e242d956f063559/src/test-deno/deno.json#L7

Follows their docs recommendation for [editing a package before execution](https://deno.com/manual@v1.33.1/node/npm_specifiers#--node-modules-dir-flag):

> In the case where you want to modify the contents of the node_modules directory before execution, you can run deno cache with --node-modules-dir, modify the contents, then run the script.
>
> For example:
> 
> deno cache --node-modules-dir main.ts
> deno run --allow-read=. --allow-write=. scripts/your_script_to_modify_node_modules_dir.ts
> deno run --node-modules-dir main.ts

.github/workflows/main.yml
rajsite: done!

rajsite:
> Here the main issue I see is that it would be good to load `pyodide.js` from the dev deployment rather than npm. So we are actually testing the dev branch. If for instance we start a HTTP server in the backround serving `dist/` and loading it from
> 
> ```
> import pyodideModule from "https://localhost:<some-port>/pyodide.js";
> ```
> 
> that wouldn't work?

I don't believe that would work. I'm relying on the somewhat magical `npm:` import resolution behavior of Deno. So the syntax you show would be treated as an es6 module import, not a commonjs import.

I think for this PR it makes sense to use the workflow from Deno's docs of editing the node_modules contents, but let me know if you want to look at different directions.

hoodmane:
> error: Uncaught Error: Cannot determine runtime environment

Ideally we should fix this...

hoodmane:
Though I don't see that it should be a blocker on this PR.

src/test-deno/deno.lock
hoodmane, line 1: How does one regenerate this file if the requirements change?

hoodmane (approved):
I think this needs a readme explaining how to generate the lock file, but aside from that I'm good with merging this.

hoodmane:
Thanks for working on this @rajsite!

src/test-deno/deno.lock
rajsite: This ended up being a more interesting question than I would like.
Turns out that it's super easy, deno just always updates it when `deno cache` is run, which frankly seems like a flaw coming from my expectations of how npm package-lock.json files work: https://github.com/denoland/deno/issues/7075

So if a new version of pyodide is released, since I didn't put an exact specifier on the pyodide import, it'll just pull in the new pyodide version when `deno cache` is run on the CI and dirty the lock file (I verified that to be the case locally).

There is a workaround to [fail the build when the dependency tree changes](https://github.com/denoland/deno/issues/7075#issuecomment-1027758722). Using that workaround, if I put a version specifier on the pyodide import I think that would guard against pyodide updates causing the workaround check to fail. I'm not sure if transitive dependency updates would result in build failures though due to updates during `deno cache`. That might not be an issue since pyodide has so few updates though.

Proposed Outcome:
1. But a version specifier on the pyodide import (that gets thrown away and replaced with a local build anyway)
2. Add the assertion check from that workaround to know if the resolved dependencies change by erroring
3. Add README instructions for updating the lock file (run `deno cache` locally and check-in the update)

rajsite: Implemented the proposal with strict validation and added README docs 👍

rajsite:
Absolutely no rush or anything. Just an FYI that I think I addressed the existing comments. Not sure if y'all prefer resets or anything @rth @hoodmane.

Thanks for y'alls time and good questions helping me learn more about Deno on the fly! :P

hoodmane (approved):

PR #3811, opened 2023-05-01
Change --output-directory to --outdir

### LINKED ISSUE #3806: Allow specifying output directory in pyodide build with -o or --outdir
I added the capability to specify output directory in pyodide build with `--output-directory`. But here `pypa/build` accepts either `--outdir` or `-o`. We should accept these as ways to specify the output directory for consistency

hoodmane:
Resolves #3806.

### DIFF
--- a/docs/project/changelog.md
+++ b/docs/project/changelog.md
@@ -15,8 +15,9 @@ myst:
 
 ## Unreleased
 
-- {{ Breaking }} Changed the name of the `--output-directory` argument to
-  `pyodide build` to `--outdir` to match pypa/build.
+- {{ Enhancement }} Changed the name of the `--output-directory` argument to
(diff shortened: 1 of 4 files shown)


rth (commented):
This is a breaking change though. We can't put in a patch release. I though you were planning to to a deprecation cycle (i.e. define a second parameter `outdir`) and deprecate this one.

hoodmane:
Okay fixed it to deprecate instead of change.

rth (approved):
Thanks !

> We introduced this feature in a patch release two weeks ago. Surely we should be able to break it in a patch release? I highly doubt many people use it yet...

Then it's indeed more debatable, but probably still not great to have a breaking change in the changelog without really strong reasons. The current version would work, thanks!

Issue #3424, opened 2023-01-05
Add package: protobuf


josephrocca:
## 🐍 Package Request

- Package Name and Version : protobuf / latest
- Package URL : https://github.com/protocolbuffers/protobuf/tree/v21.12/python
- Package Dependencies that needs to be resolved first: None?

It seems like this already works "out of the box", but it's just that they haven't published a pypi package specifically for the pure python implementation? I think the `protobuf` pypi package is for their C++ implementation? I'm not sure why they wouldn't publish two separate packages, but maybe it's quite rare for people to use the pure python implementation, since it's significantly slower.

I built the whl and made a quick demo and it seems to work fine:

* built whl: https://github.com/josephrocca/protobuf-pyodide/raw/main/demo/protobuf-4.21.12-py2.py3-none-any.whl
* working demo: https://josephrocca.github.io/protobuf-pyodide/demo/
* repo: https://github.com/josephrocca/protobuf-pyodide

## Code I used to build it:

```bash
git clone --branch v21.12 https://github.com/protocolbuffers/protobuf
cd protobuf

# pyodide setup:
pip install pyodide-build
git clone https://github.com/emscripten-core/emsdk
cd emsdk
pyodide config get emscripten_version  # this line is needed due to: https://github.com/pyodide/pyodide/issues/3430
./emsdk install $(pyodide config get emscripten_version)
./emsdk activate $(pyodide config get emscripten_version)
source emsdk_env.sh
cd ../

# Install `protoc` binary - https://askubuntu.com/questions/1072683/how-can-i-install-protoc-on-ubuntu-16-04/1157801#1157801
# CAUTION: Must be same version as the protobuf repo being built!
wget https://github.com/protocolbuffers/protobuf/releases/download/v21.12/protoc-21.12-linux-x86_64.zip
sudo unzip -o protoc-21.12-linux-x86_64.zip -d /usr/local bin/protoc
sudo unzip -o protoc-21.12-linux-x86_64.zip -d /usr/local 'include/*'

# install bazel:
npm install -g @bazel/bazelisk

# build
cd python
pyodide build

# create test file:
wget https://gist.githubusercontent.com/josephrocca/f9184834444be2515b264aca2c9e7fde/raw/cc25a18b1e4d9a71bd99c8343bd31bfd43116e85/addressbook.proto
protoc -I=. --python_out=. addressbook.proto
# now you have `addressbook_pb2.py` in the the current working directory
```

## Code from demo:
```html
[...]

PR #3813, opened 2023-05-01
Add protobuf package

bartbroere:
Adds the Python `protobuf` package. I'll keep it marked as a draft until I'm sure it's okay and actually works as intended.

Having this package in place would open up a lot of possibilities for other packages. Backwards compatibility of protobufs seems okay, so I don't see any dangers for compatibility in having the newest release included in Pyodide. See related discussion in: https://github.com/pyodide/pyodide/issues/3424

- [x] Add / update tests

### DIFF
--- a/docs/project/changelog.md
+++ b/docs/project/changelog.md
@@ -36,7 +36,7 @@ myst:
   {pr}`3331`.
 - New packages: sourmash {pr}`3635`, screed {pr}`3635`, bitstring {pr}`3635`,
   deprecation {pr}`3635`, cachetools {pr}`3635`, xyzservices {pr}`3786`,
-  simplejson {pr}`3801`.
+  simplejson {pr}`3801`, protobuf {pr}`3813`.
 - Upgraded libmpfr to 4.2.0 {pr}`3756`.
 - Upgraded scipy to 1.10.1 {pr}`3794`
 
--- a/packages/protobuf/meta.yaml
+++ b/packages/protobuf/meta.yaml
@@ -1,0 +1,15 @@
+package:
+  name: protobuf
+  version: 4.22.3
+  top-level:
+    - google
+
+source:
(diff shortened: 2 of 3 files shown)


ryanking13:
Thanks for the porting work @bartbroere. I have some novice questions about protobuf:

1. What is the difference between, protobuf library / protobuf Python package / protoc binary? Does protobuf Python package compatible with other versions (e.g. protobuf <= 3.X) of protobuf / protoc?
2. What is the difference between protobuf pure Python package / C-extension package?

bartbroere:
First of all, I still have some of the same questions, that's why I'll keep this pull request a draft for now.

> 1. What is the difference between, protobuf library / protobuf Python package / protoc binary? Does protobuf Python package compatible with other versions (e.g. protobuf <= 3.X) of protobuf / protoc?

The `protoc` binary originally is a C++ executable I believe, but there also is a Cython executable that imports `libprotoc` and is probably functionally equivalent, with the added bonus that it can be called from Python code directly without `subprocess.call` trickery. This last `protoc` flavour is in `grpcio_tools`, a package that I hope to add after this.

> 2. What is the difference between protobuf pure Python package / C-extension package?

I'm not sure, but I see adding the `protobuf` package as the first step towards adding more useful packages, such as `onnx`

bartbroere:
`protobuf` isn't a pure Python wheel by the way, it seems, but it can be.

Like other packages it [has a fallback mechanism](https://github.com/protocolbuffers/protobuf/blob/main/python/google/protobuf/internal/api_implementation.py#L73-L79)
I added a test that checks whether the binary component is present.

And regarding `onnx`, this will be fixed in a release sometime soon I expect: https://github.com/onnx/onnx/pull/4956/files#diff-4d7c51b1efe9043e44439a949dfd92e5827321b34082903477fd04876edb7552

So, to summarise, I think it all looks okay to merge.

ryanking13 (approved):
Thanks, @bartbroere!

PR #3814, opened 2023-05-01
fix: drop some no-longer-needed bits from pydodide build

henryiii:
### Description

A few things here aren't really needed:

* `"wheel"` is injected via a PEP 517 hook when building wheels for all versions of setuptools that support PEP 517.
* `LICENSE` is already included, no need for a MANIFEST.in. You can compare SDist and Git via `pipx run check-sdist`.
* Recent-ish versions of Pip and Setuptools now support editable installs without setup.py

### DIFF
--- a/pyodide-build/MANIFEST.in
+++ b/pyodide-build/MANIFEST.in
@@ -1,1 +1,0 @@
-include LICENSE
--- a/pyodide-build/pyproject.toml
+++ b/pyodide-build/pyproject.toml
@@ -1,4 +1,3 @@
 [build-system]
-requires = ["setuptools>=42", "wheel"]
-
+requires = ["setuptools>=42"]
 build-backend = "setuptools.build_meta"
--- a/pyodide-build/setup.py
+++ b/pyodide-build/setup.py
@@ -1,5 +1,0 @@
-# Needed for editable install
-import setuptools
-
-if __name__ == "__main__":
-    setuptools.setup()


henryiii:
Why are namespace packages used? And where are they used?

henryiii:
It's a package without a `__init__.py`. Generally should only be used to combine multiple packages (like adding `a.c` to `a.b` from a different place on disk).

hoodmane:
And how do you tell that we're using one?

henryiii:
https://github.com/pyodide/pyodide/blob/4777f5920059ac97bc159bc6c5c725dbca2703a6/pyodide-build/setup.cfg#L23 is what I saw that I prefer to avoid if possible.

henryiii:
Yep, `tools` is missing a `__init__.py`. Namespace packages weren't created to avoid empty `__init__.py`'s, they were created to allow packages to be split; you get collisions with `__init__.py`'s without namespace packages. The better solution here would be to drop an `__init__.py` in these if you are using importlib.resources to access them. Or to treat them all as data.

hoodmane:
Hmm it's not clear to me why we need it. In the tests we add the tools directory to the path, but that should allow us to import modules from the directory. I don't see any instances of `import tools` or `from tools import x` or anything.

hoodmane:
Thanks @henryiii!

hoodmane:
Hopefully we can fix the namespace packages in a followup.

ryanking13:
> The better solution here would be to drop an __init__.py in these if you are using importlib.resources to access them.

Sounds good to me. We don't access files in `tools` directory with importlib, but I always had bad experiences dealing with namespaces packages... so adding an empty `__init__.py` then use `packages = find:` might be better.

hoodmane:
If we don't import from it isn't the correct thing to indicate that everything in the directory is data files?

ryanking13:
> If we don't import from it isn't the correct thing to indicate that everything in the directory is data files?

That would be ideal, but I never made [Subdirectory for data files](https://setuptools.pypa.io/en/latest/userguide/datafiles.html#subdirectory-for-data-files) + `packages = find:` combination successful.

PR #3817, opened 2023-05-01
ABI break: Use wasm exception handling

hoodmane:
This should have significant benefits for code size and speed, particularly for Rust.

Requires:
Firefox 100 released May 3, 2022 (previously Firefox 78 released June 30, 2020)
Safari 15.2 released December 14, 2021 (previously 14.2 released Jul 19, 2021)
Chrome 95 released Oct 19, 2021 (previously Chrome 85 released Aug. 25, 2020)

https://webassembly.org/roadmap/

I wonder if we have to modify this in some way...
https://github.com/rust-lang/rust/blob/master/compiler/rustc_codegen_ssa/src/back/link.rs#L2213-L2217

- [ ] Add / update tests


rth:
> This should have significant benefits for code size and speed,

Do you have some number for those? The  size impact on pyodide.asm.wasm seems not that large at least (after compression at least)
**main**
```
- pyodide.asm.wasm:
    Original size: 8739 KB
    Brotli compressed size: 2099 KB
```
**this PR**
```
- pyodide.asm.wasm:
    Original size: 8654 KB
    Brotli compressed size: 2086 KB
```

> Requires:

If we look at documentation visits, out of 17k visitors last month,
 - 10 users have <firefox 100
 - 12 users have safari < 15
 - no users have Chrome < 95 (at least in that table).
 
 So overall sounds good. What would happen if someone loads Pyodide in an unsupported browser would it crash at initial load or only crash if there is a wasm exception? 
 
 BTW emscripten has settings for [minimal versions of browsers](https://github.com/emscripten-core/emscripten/blob/main/src/settings.js#L1793) I don't know if it's doing a runtime check but maybe we should start setting those.

rth:
BTW, this also means one would need at least Nodejs 17.0 (release 2021-10) from what I understand. It should be OK but it can mean that people will no longer be able to use node installed via the system package manager if their linux system is not up to date.

hoodmane:
I have to look more carefully. The thing is that there are no exceptions in the main module so it wouldn't be affected in size very much. I think it should mostly affect the size and performance of Rust / C++ code.

hoodmane:
> What would happen if someone loads Pyodide in an unsupported browser would it crash at initial load or only crash if there is a wasm exception?

It would crash at load time. We could feature detect an supported browser and raise a better error if you like.

hoodmane:
Actually this isn't right, it probably won't crash until someone tries to load a Rust or C++ package.

hoodmane:
I think this is currently blocked on needing to adjust the Rust compiler.

hoodmane:
I opened an issue on the Rust compiler:
https://github.com/rust-lang/rust/issues/112195

hoodmane:
False alarm, I got it working with:
```
-C llvm-args=-enable-emscripten-cxx-exceptions=0
```

PR #3819, opened 2023-05-02
Update node in docker image from v14 to v20

hoodmane:
### Description

For #3817 we need to test against node >=17. For #3210 it is most convenient to test against node >= 20.
This updates the docker image to use node 20.


hoodmane:
So far I've failed to reproduce test_fileio-node failure locally.

hoodmane:
Interestingly it seems to be a v8 regression of some sort between the version used in node v14 and the more modern version used in node v20 and chrome 112.

ryanking13:
Thanks, @hoodmane. I always wanted to use Node.js version >= 18 (yay built-in fetch) so I am basically happy with this.

Since Node.js version 14's end-of-life was 2 days ago, and Node.js version 16's end-of-life is coming soon (in 5 months) regarding our release cycle, I am +1 with this PR.

One thing we should decide though is "What is going to be the Minimal supported Node.js version of Pyodide". After this PR and #3817 are merged, should we say users to use Node.js >= 0.17 if they have trouble in Node.js < 0.17?

hoodmane:
Let's say min supported node is 18.

ryanking13:
> Let's say min supported node is 18.

Sounds okay to me. We need to mention it in the docs then.
I think @rth was also okay [with this](https://github.com/pyodide/pyodide/pull/3817#issuecomment-1531188787), but let's wait for what he says.

rth:
Yeah, no objections from me.  Thanks for working on it. Doing a cycling trip so I'll be offline for a few more days.

ryanking13:
Have a nice trip Roman!

PR #3821, opened 2023-05-03
Refactor install from pypi test

hoodmane:
This should be more robust against changes to what pip prints to stdout.


ryanking13 (approved):
Thanks!

ryanking13:
Though the test didn't run in CI because of the timeout in package build.

PR #3824, opened 2023-05-03
Change the name of repodata.json to pyodide-lock.json

### LINKED ISSUE #3788: Change `repodata.json` name to something else?
I am told by conda people that repodata.json is supposed to be the input to a solver. Our repodata.json is a lock file which is the output of a solver. This is a misleading name clash. We will need to spend a lot of time explaining to conda people that this file isn't what they expect to be. Can we rename it?

hoodmane:
Resolves #3788


src/js/load-package.ts
```diff
@@ -17,28 +17,28 @@ import { makeWarnOnce } from "./util";
...
+  if (!lockfile.packages) {
     throw new Error(
-      "Loaded repodata.json does not contain the expected key 'packages'.",
+      "Loaded pyodide lock file does not contain the expected key 'packages'.",
     );
   }
-  API.repodata_info = repodata.info;
-  API.repodata_packages = repodata.packages;
+  API.repodata_info = lockfile.info;
+  API.repodata_packages = lockfile.packages;
```
ryanking13, line 41: Probably we should update the variable name in `micropip` and here together.

ryanking13 (approved):
Thanks, @hoodmane. I think the name `pyodide-lock.json` is good.

src/js/load-package.ts
hoodmane: For compatibility reasons we could store it into both `repodata_xxx` and `lockfile_xxx` for a while.

rth (approved):
LGTM as well.

Issue #3336, opened 2022-12-09
Way to include large packages without having to build them in the pyodide build


joemarshall:
I was looking at building the pyside qt bindings for pyodide. However to build it, you need to build:

1) host qt
2) qt for webassembly
3) pyside

There are pre-built versions of qt for webassembly, but they only work on a fixed version of emscripten, so I think I need to build everything from source to work with pyodide. 

The build process is close to an hour on my PC, so it's obviously a stupid idea to include it, but on the other hand, being able to use qt in pyodide would be super neat, and I can it being really useful for quite a few people (including someone who might be paying me to port some things at some point in the near future!).

Now I can obviously build it in a separate repo, and let people micropip.install with a release URL, but it would be really nice to be easily able to create a pyodide package definition that just pointed at an external wheel from e.g. a project release on github, so that people could see it existed easily.

Maybe this goes into a wider discussion of package repositories for pyodide though, or at least of splitting off the pyodide package repo from the pyodide core...

ryanking13:
> Now I can obviously build it in a separate repo, and let people micropip.install with a release URL, but it would be really nice to be easily able to create a pyodide package definition that just pointed at an external wheel from e.g. a project release on github, so that people could see it existed easily.

BTW, what happens if you add WASM wheel url to `meta.yaml`?

```yaml
source:
  url: https://some.url/pkg-1.0.0-cp310-cp310-emscripten_3_1_27_wasm32.whl
  sha256: ...
```

I think for now this is quite the only way to include a package that is too big to compile in Pyodide CI.

rth:
Also related to the discussion in https://github.com/pyodide/pyodide/issues/3341 but as a short term solution +1 to support including emscripten/wasm32 wheels under source/url from trusted sources

rth:
With the understanding that it would be disabled if no builds exist in the external repository, when we update to a new emscripten version, and the original maintainer would need to make a PR to re-enable it. Hmm, which also doesn't sound very scalable.

ryanking13:
Closing as we now support wasm wheels in the recipe.

PR #3827, opened 2023-05-04
POC: build recipes in seperate repository

ryanking13:
This is a proof-of-concept of building recipes in a separate repository, using tot xbuildenv. You can see how the packages are built in: https://github.com/ryanking13/pyodide-recipes-mirror

We are getting more and more packages, and therefore CI timeouts often, so we need a workaround. Probably this PR cannot be a long-term solution but can work as a workaround until PyPI supports uploading Pyodide wheels.

Of course, this PR will increase the maintenance burden: we will have to maintain two slightly different recipes. I think it would be possible to automate some parts of recipe updates... but it will definitely be less convenient than managing recipes in one place.

I don't think we should move out all recipes, but maybe we can think of moving some packages based on criteria. In this PR, I selected mypy. mypy 1) takes long time to build (~10min), 2) and it has no dependencies, so it is a good candidate I think. 

- Related: https://github.com/pyodide/pyodide/issues/3341
- Related: #3336

WDYT?

### DIFF
--- a/packages/mypy/meta.yaml
+++ b/packages/mypy/meta.yaml
@@ -5,12 +5,12 @@ package:
     - mypyc
     - mypy
 source:
-  url: https://files.pythonhosted.org/packages/62/54/be80f8d01f5cf72f774a77f9f750527a6fa733f09f78b1da30e8fa3914e6/mypy-1.1.1.tar.gz
-  sha256: ae9ceae0f5b9059f33dbc62dea087e942c0ccab4b7a003719cb70f9b8abfa32f
-build:
-  script: |
-    export MYPY_USE_MYPYC=1
-    export MYPYC_OPT_LEVEL=3
+  url: https://github.com/ryanking13/pyodide-recipes-mirror/releases/download/20230503/mypy-1.1.1-cp311-cp311-emscripten_3_1_32_wasm32.whl
+  sha256: f6103543a6612116b6fcbde63fd6de07e79f3d1c16e3bbd5585d77585bf23a9d
+# build:
+#   script: |
+#     export MYPY_USE_MYPYC=1
+#     export MYPYC_OPT_LEVEL=3
 about:
   home: http://www.mypy-lang.org/
   PyPI: https://pypi.org/project/mypy
--- a/pyodide-build/pyodide_build/pypabuild.py
+++ b/pyodide-build/pyodide_build/pypabuild.py
@@ -78,7 +78,9 @@ def install_reqs(env: IsolatedEnv, reqs: set[str]) -> None:
     pinned_reqs = {
         # Remove this when mypy releases a new version
         # https://github.com/python/mypy/pull/14788
-        "types-setuptools": "types-setuptools<67.4.0.2"
+        "types-setuptools": "types-setuptools<67.4.0.2",
+        # https://github.com/pyodide/pyodide/issues/3816
+        "pythran>=0.12.0,<0.13.0": "gast!=0.5.4",
     }
 
     for pkg, req in pinned_reqs.items():
--- a/pyodide-build/pyodide_build/pypabuild.py
+++ b/pyodide-build/pyodide_build/pypabuild.py
@@ -78,9 +78,7 @@ def install_reqs(env: IsolatedEnv, reqs: set[str]) -> None:
     pinned_reqs = {
         # Remove this when mypy releases a new version
         # https://github.com/python/mypy/pull/14788
-        "types-setuptools": "types-setuptools<67.4.0.2",
-        # https://github.com/pyodide/pyodide/issues/3816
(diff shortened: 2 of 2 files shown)


hoodmane:
Thanks for looking into this @ryanking13! I think it's a good idea. I would like to see fewer packages built on every commit. How do you see it working with the release process?

ryanking13:
I think we can

1) Trigger recipe build CI when a commit is merged to Pyodide main branch to make sure any updates in pyodide-build do not break the package build.

2) Release packages before releasing a new Pyodide version or when we update packages. This needs to be done manually by creating a new tag.

3) Rewriting recipe URLs in Pyodide main repository needs to be automated. I think I can write a simple script that does this.

hoodmane:
> Trigger recipe build CI when a commit is merged

Speaking of which we should think about making some automated tests against tip of tree python. Do you know a way to have a github action auto trigger when someone commits to someone else's repo?

ryanking13:
> Do you know a way to have a github action auto trigger when someone commits to someone else's repo? (This is a bit off topic I know...)

We can use [`repository_dispatch`](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#repository_dispatch) I think. We have an experience of using it before when we [trigger Github Actions](https://github.com/pyodide/pyodide/blob/0.21.0/.github/workflows/deploy_release.yml#L3-L4) in [CircleCI](https://github.com/pyodide/pyodide/blob/0.21.0/.circleci/config.yml#L356-L365) (later we moved all release jobs to GHA so these were removed.)

rth:
Thanks for investigating it!

> Of course, this PR will increase the maintenance burden: we will have to maintain two slightly different recipes

I was thinking of doing something as follows. Move all of `packages/` folder to a different repo, as is. Then,

 - clone that repo as part of the CI here and,
   - at each commit build some representative subset of packages (IMO we are building too much stuff currently)
   - (initially) have some way to build everything (e.g. via a commit message & git tags). So initially it wouldn't really change anything for our release progress.
 - in that repo, initially maybe just check yaml validity or maybe try to build with the latest pyodide version it doesn't really matter.

In the medium term,
 - in the package repo, for a given commit build only the package of a changed meta.yaml and its dependencies. Always upload the build packages to s3 under e.g. `wheels/<package-name>/<wheel-name>` if it doesn't already exist. This means that the wheel name needs contain a hash e.g of the emscripten version and compile flags or some other way to know that a given wheel is compatible or not with another build of the same wheel.
 - we expose a simple API for the build packages. So there can be several versions of a given package built with the same pyodide version there in theory.
 - the main pyodide repo no longer uploads packages. It can still build some subset of them for testing. 

For the release,
1. the main pyodide repo  uploads the build CPython to `cdn.jsdelivr.net/pyodide/<version>/full/`
2. All packages are being built from the packages repo for that pyodide version, and uploaded say under `cdn.jsdelivr.net/pyodide/wheels/<package-name>/<wheel-path>`. Ideally we should know if a package needs to be rebuilt , before building it, based on the hash of the emscripten version & compilation flags, but I'm not sure if it would be robust enough.
3. There is a last step that generates the repodata.json parsing latest compatible versions of wheels from the sample API and uploading it. At the same time we provide tooling some that people can take a list of requirements, point to our sample API + PyPI, and easily generate a repodata.json for the versions they need.


Let me know whether this is compatible with what you were planning.

ryanking13:
@rth Thanks for the detailed feedback!

I am basically +1 with your idea and I think it is mostly compatible with what I am planning to do. Then I think we need to talk about "what is the first step we need to do"?

Maybe I can start by moving all recipes to a separate repository (e.g. pyodide/pyodide-recipes), then make all packages built at that repository, while this repository builds some representative subset of packages for each commit.
Then one temporary issue is that, both repositories should upload artifacts to `cdn.jsdelivr.net/pyodide/dev/full/` (this repository will push pyodide.* and the package repository will push wheels and repodata.json)

ryanking13:
> in the package repo, for a given commit build only the package of a changed meta.yaml and its dependencies. Always upload the build packages to s3 under e.g. wheels/<package-name>/<wheel-name> if it doesn't already exist. This means that the wheel name needs contain a hash e.g of the emscripten version and compile flags or some other way to know that a given wheel is compatible or not with another build of the same wheel.

Compatibility issue always gives me (and all of us) a headache... I am not fully satisfied with hashing flags, but I think it is a reasonable approach.

rth:
> Maybe I can start by moving all recipes to a separate repository (e.g. pyodide/pyodide-recipes), then make all packages built at that repository, while this repository builds some representative subset of packages for each commit.

+1. I would say even if temporary deployment to dev is not ideal that's not critical.

ryanking13:
Closing as we now have packages like PyArrow that actually builds from other repositories, and #4987 will unvendor packages

PR #3830, opened 2023-05-05
use cmake instead of configure for zlib.

katietz:
### Description

the use of configure might lead on specific version of emscripten (seen on 3.1.36) to build failures due libtools inability to check for object file format.
Additionally the switch to CMake agrees more to the way zlib is built nowadays by distributors.


hoodmane (approved):
Thanks @katietz!

PR #3833, opened 2023-05-06
NFC Fix llvm-readobj path hack

hoodmane:
Requested by @katietz

### DIFF
--- a/pyodide-build/pyodide_build/pywasmcross.py
+++ b/pyodide-build/pyodide_build/pywasmcross.py
@@ -52,7 +52,6 @@ if IS_COMPILER_INVOCATION:
 
 
 import dataclasses
-import shutil
 import subprocess
 from collections.abc import Iterable, Iterator
 from typing import Literal, NoReturn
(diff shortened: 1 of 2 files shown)


pyodide-build/pyodide_build/pywasmcross.py
```diff
@@ -390,11 +389,8 @@ def _calculate_object_exports_readobj_parse(output: str) -> list[str]:
 
 
 def calculate_object_exports_readobj(objects: list[str]) -> list[str] | None:
-    which_emcc = shutil.which("emcc")
-    assert which_emcc
-    emcc = Path(which_emcc)
     args = [
-        str((emcc / "../../bin/llvm-readobj").resolve()),
+        "llvm-readobj",
```
ryanking13, line 393: Will this work for out-of-tree build where users install emscripten by themselves?

hoodmane: Ah! Good catch. We can put this back in as a fallback if it's not on the path.

PR #3835, opened 2023-05-07
NFC Fix sed patterns

katietz:
### Description

Use of sed code is actually pretty OS sepcific.  Therefore I replaced indicated changes by a patch file instead.
Underlying issue for this change was that all patterns in SED commands weren't replaced multiple times.  (see that final 'g' was missing in SED pattern.
That led to broken archives for emscripten vs native use of ranlib/ar tool


hoodmane:
Seems like the patch context doesn't match the file.

katietz:
yes, indeed ... this project uses ARCH for emar ... which is actually pretty unusual.  Fixed it

PR #3837, opened 2023-05-08
Use gfortran instead of gcc for preprocessing in replay_f2c

katietz:
### Description

This PR fixes some issues regarding OSX build.  As gcc is not necessarily available nativly on osx, we need to use instead gfortran frontend for preprocessing.  Later is also the thing we probe for, so it is even more direct
Additionally this PR removes use of linker-option -install_name, which might be introduces on darwin OS.  For now I just did so for rebound package.
openblas needs some adjustments for use of em<tools> instead of native ones.  Additional it adds a patch we makes sure it will make use of emar in cmake.

### DIFF
--- a/packages/openblas/patches/0003-fix-cmake-ar-use.patch
+++ b/packages/openblas/patches/0003-fix-cmake-ar-use.patch
@@ -1,0 +1,24 @@
+Index: openblas-0.3.23/CMakeLists.txt
+===================================================================
+--- openblas-0.3.23.orig/CMakeLists.txt
++++ openblas-0.3.23/CMakeLists.txt
+@@ -252,15 +252,15 @@ if (APPLE AND DYNAMIC_ARCH AND BUILD_SHA
+   if (NOT NOFORTRAN)
+   set (CMAKE_Fortran_USE_RESPONSE_FILE_FOR_OBJECTS 1)
+   set (CMAKE_Fortran_CREATE_SHARED_LIBRARY
+- "sh -c 'cat ${CMAKE_BINARY_DIR}/CMakeFiles/openblas_shared.dir/objects*.rsp | xargs -n 1024 ar -ru libopenblas.a && exit 0' "
+- "sh -c 'ar -ru libopenblas.a ${CMAKE_BINARY_DIR}/driver/others/CMakeFiles/driver_others.dir/xerbla.c.o && exit 0' "
++ "sh -c 'cat ${CMAKE_BINARY_DIR}/CMakeFiles/openblas_shared.dir/objects*.rsp | xargs -n 1024 emar -ru libopenblas.a && exit 0' "
++ "sh -c 'emar -ru libopenblas.a ${CMAKE_BINARY_DIR}/driver/others/CMakeFiles/driver_others.dir/xerbla.c.o && exit 0' "
+  "sh -c 'echo \"\" | ${CMAKE_Fortran_COMPILER} -o dummy.o -c -x f95-cpp-input - '"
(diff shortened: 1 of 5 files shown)


packages/openblas/patches/0003-fix-cmake-ar-use.patch
```diff
@@ -0,0 +1,24 @@
...
+===================================================================
+--- openblas-0.3.23.orig/CMakeLists.txt
++++ openblas-0.3.23/CMakeLists.txt
+@@ -252,15 +252,15 @@ if (APPLE AND DYNAMIC_ARCH AND BUILD_SHA
+   if (NOT NOFORTRAN)
+   set (CMAKE_Fortran_USE_RESPONSE_FILE_FOR_OBJECTS 1)
+   set (CMAKE_Fortran_CREATE_SHARED_LIBRARY
+- "sh -c 'cat ${CMAKE_BINARY_DIR}/CMakeFiles/openblas_shared.dir/objects*.rsp | xargs -n 1024 ar -ru libopenblas.a && exit 0' "
+- "sh -c 'ar -ru libopenblas.a ${CMAKE_BINARY_DIR}/driver/others/CMakeFiles/driver_others.dir/xerbla.c.o && exit 0' "
++ "sh -c 'cat ${CMAKE_BINARY_DIR}/CMakeFiles/openblas_shared.dir/objects*.rsp | xargs -n 1024 emar -ru libopenblas.a && exit 0' "
```
ryanking13, line 11: I guess we can use `${CMAKE_AR}` here. Maybe we should upstream this.

```suggestion
+ "sh -c 'cat ${CMAKE_BINARY_DIR}/CMakeFiles/openblas_shared.dir/objects*.rsp | xargs -n 1024 ${CMAKE_AR} -ru libopenblas.a && exit 0' "
```

ryanking13 (approved):
Thanks @katietz!

ryanking13 (approved):
Thanks @katietz!

packages/rebound/patches/0001-fix-install_name.patch
```diff
@@ -0,0 +1,15 @@
...
+--- rebound-3.24.2.orig/setup.py
++++ rebound-3.24.2/setup.py
+@@ -23,8 +23,8 @@ extra_link_args=[]
+ if sys.platform == 'darwin':
+     from distutils import sysconfig
+     vars = sysconfig.get_config_vars()
+-    vars['LDSHARED'] = vars['LDSHARED'].replace('-bundle', '-shared')
+-    extra_link_args=['-Wl,-install_name,@rpath/librebound'+suffix]
++    # vars['LDSHARED'] = vars['LDSHARED'].replace('-bundle', '-shared')
++    # extra_link_args=['-Wl,-install_name,@rpath/librebound'+suffix]
```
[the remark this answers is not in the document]
katietz: I noticed this while using a system python interpreter for host on osx-arm64.  With those lines wasm-ld is choking about unknown linker option.  See that this lines are just active is python reports 'darwin' as platform.

packages/openblas/patches/0003-fix-cmake-ar-use.patch
katietz: yeah, good point

ryanking13:
Thanks again @katietz.

It seems like linter is complaining about the complexity of a function. I am not sure we want that complexity check @hoodmane.Let's ignore the warning for now.

hoodmane:
I think ignoring it for now is fine, but it wouldn't be a bad idea to split that function into two or three parts.

katietz:
what is this deno-test related to this changes?  it seems to me that on one builder an new version gets pulled in by npm, which causes this

ryanking13:
> what is this deno-test related to this changes? it seems to me that on one builder an new version gets pulled in by npm, which causes this

It is not related to this PR.

ryanking13:
Could you please rename the title of your PR to be more meaningful? Then I think we can merge this.

rth:
Thanks @katietz !

PR #3840, opened 2023-05-08
Fix llvm-readobj when emscipten is installed with emsdk out of tree

hoodmane:
Thanks to @ryanking13 for pointing out that my previous patch breaks out of tree builds.

### DIFF
--- a/pyodide-build/pyodide_build/pywasmcross.py
+++ b/pyodide-build/pyodide_build/pywasmcross.py
@@ -53,6 +53,7 @@ if IS_COMPILER_INVOCATION:
 
 
 import dataclasses
+import shutil
 import subprocess
 from collections.abc import Iterable, Iterator
 from typing import Literal, NoReturn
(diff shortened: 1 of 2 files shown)


hoodmane:
@rajsite I don't like the way that the deno test is failing for reasons unrelated to the patch. Can we remove the check that the lockfile is up to date?

pyodide-build/pyodide_build/pywasmcross.py
```diff
@@ -390,8 +391,14 @@ def _calculate_object_exports_readobj_parse(output: str) -> list[str]:
 
 
 def calculate_object_exports_readobj(objects: list[str]) -> list[str] | None:
+    readobj_path = shutil.which("llvm-readobj")
+    if not readobj_path:
+        which_emcc = shutil.which("emcc")
+        assert which_emcc
+        emcc = Path(which_emcc)
+        readobj_path = str((emcc / "../../bin/llvm-readobj").resolve())
```
ryanking13, line 399: What do you think of appending this directory to PATH? (maybe in [here](https://github.com/pyodide/pyodide/blob/5754d211bec8cbbea49b0949b3e3e1e596cff9ee/pyodide-build/pyodide_build/pypabuild.py#L227)), so shutil.which() can automatically fallback to the emsdk llvm-readobj when it is not in host PATH.

ryanking13 (approved):
Thanks, @hoodmane!

ryanking13 (approved):
Thanks, @hoodmane!

pyodide-build/pyodide_build/pywasmcross.py
hoodmane: Well that would be a reasonable thing to do. But I think this is fine, and ensures that we don't put any unknown things on the path.

ryanking13: > But I think this is fine, and ensures that we don't put any unknown things on the path.

Yes, that makes sense. This is fine with me. Thanks!

rajsite:
> @rajsite I don't like the way that the deno test is failing for reasons unrelated to the patch. Can we remove the check that the lockfile is up to date?

Will do!

PR #3842, opened 2023-05-08
Add lmdb module

benjeffery:
Adding lmdb, as it is a dependency of tsinfer, which I need for a deployment of interactive educational material.

### DIFF
--- a/packages/lmdb/test_lmdb.py
+++ b/packages/lmdb/test_lmdb.py
@@ -1,0 +1,30 @@
+from pytest_pyodide import run_in_pyodide
+
+
+@run_in_pyodide(packages=["lmdb"])
+def test_lmdb(selenium):
+    from tempfile import TemporaryDirectory
+
+    import lmdb
+
+    with TemporaryDirectory() as tmpdir:
+        path = f"{tmpdir}/mytestdb"
+
+        # create a new lmdb Environment
+        env = lmdb.open(path, max_dbs=10)
+
+        # make sure the environment is open
+        assert env.open_db(b"mydb")
+
+        # try to create a new lmdb Transaction
+        with env.begin(write=True) as txn:
+            # add a key-value pair to the Transaction
+            txn.put(b"key", b"value")
+
+        # try to retrieve the value we just added
+        with env.begin() as txn:
+            value = txn.get(b"key")
+        assert value == b"value"
+
+        # cleanup
+        env.close()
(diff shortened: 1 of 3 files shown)


benjeffery:
> How does the connection to lmdb database works? Does it make HTTP requests or is it lower-level network communications? In the latter case, this would be problematic as opening a socket connection will not work in the browser.

The database is on-disk - similar to sqlite. So no network is needed.

ryanking13:
@benjeffery Could you add some basic tests, so we can make sure it works?

ryanking13 (approved):
Thanks, @benjeffery!

packages/lmdb/test_lmdb.py
```diff
@@ -0,0 +1,30 @@
+from tempfile import TemporaryDirectory
+
+from pytest_pyodide import run_in_pyodide
+
+
+@run_in_pyodide(packages=["lmdb"])
+def test_lmdb(selenium):
+    import lmdb
```
ryanking13, line 8: You need to use `from tempfile import TemporaryDirectory` inside this function.

hoodmane:
Uh oh there seems to be a symbol resolution problem.

ryanking13:
The error message says that it cannot resolve the symbol `pthread_mutexattr_setrobust`, which seems related to threading (Pyodide does not support thread).

I guess there should be a compile flag that disables threading in lmdb? (Maybe you can try flags used here: https://github.com/LMDB/lmdb/blob/3947014aed7ffe39a79991fa7fb5b234da47ad1a/libraries/liblmdb/mdb.c#L313-L339)

benjeffery:
> The error message says that it cannot resolve the symbol `pthread_mutexattr_setrobust`, which seems related to threading (Pyodide does not support thread).
> 
> I guess there should be a compile flag that disables threading in lmdb? Maybe you can try flags used here: https://github.com/LMDB/lmdb/blob/3947014aed7ffe39a79991fa7fb5b234da47ad1a/libraries/liblmdb/mdb.c#L313-L339

Thanks so much for looking into this! I'll try that out.

benjeffery:
> > The error message says that it cannot resolve the symbol `pthread_mutexattr_setrobust`, which seems related to threading (Pyodide does not support thread).
> > I guess there should be a compile flag that disables threading in lmdb? Maybe you can try flags used here: https://github.com/LMDB/lmdb/blob/3947014aed7ffe39a79991fa7fb5b234da47ad1a/libraries/liblmdb/mdb.c#L313-L339
> 
> Thanks so much for looking into this! I'll try that out.

Seems I'd have to make a substantial patch to get this in. Closing for now, thanks for the pointers!

PR #3845, opened 2023-05-09
Support Fortran-file preprocessing on case-insensitive file-system

katietz:
### Checklists

On case-insensitive file-systems - eg like MacOS standard Extended file-system (not APFS), FAT, etc - it is problematic if for files with .F extension the output is written as file with .f extension, as both actually are referring to the same file.
After the operation of f2c, the original file is restored to its original name.

### DIFF
--- a/pyodide-build/pyodide_build/pywasmcross.py
+++ b/pyodide-build/pyodide_build/pywasmcross.py
@@ -151,7 +151,9 @@ def replay_f2c(args: list[str], dryrun: bool = False) -> list[str] | None:
                     if os.path.isfile(filepath):
                         os.remove(filepath)
                     if os.path.isfile(filepath.with_suffix(".oldF")):
-                        shutil.move(filepath.with_suffix(".oldF"), filepath.with_suffix(".F"))
+                        shutil.move(
+                            filepath.with_suffix(".oldF"), filepath.with_suffix(".F")
+                        )
             new_args.append(arg[:-2] + ".c")
             found_source = True
         else:
--- a/pyodide-build/pyodide_build/pywasmcross.py
+++ b/pyodide-build/pyodide_build/pywasmcross.py
@@ -10,8 +10,6 @@ cross-compiling and then pass the command long to emscripten.
 """
 import json
 import os
-import re
-import shutil
 import sys
 from pathlib import Path
 
@@ -54,6 +52,8 @@ if IS_COMPILER_INVOCATION:
 
 
 import dataclasses
+import re
+import shutil
 import subprocess
 from collections.abc import Iterable, Iterator
 from typing import Literal, NoReturn
--- a/pyodide-build/pyodide_build/pywasmcross.py
+++ b/pyodide-build/pyodide_build/pywasmcross.py
@@ -126,7 +126,7 @@ def replay_f2c(args: list[str], dryrun: bool = False) -> list[str] | None:
                             filepathwith_suffix(".f77"),
                         ]
                     )
-                    filepath = filepath.with_suffix(".fi77")
+                    filepath = filepath.with_suffix(".f77")
                 # -R flag is important, it means that Fortran functions that
                 # return real e.g. sdot will be transformed into C functions
                 # that return float. For historic reasons, by default f2c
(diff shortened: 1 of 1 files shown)


hoodmane:
I don't like this idea because then I cannot easily rerun partial compilations debugging package builds.

katietz:
well, it is not so much a question if -liking- but more a question of operation at all.  I will see what I can do here to make it re-runable

katietz:
For case-insensitiv filesystem the approach is re-runable ... so we would need to probe the kind of file-system here, or better said if original approach would work, or not ...

rth:
Even case-insensitive file systems aside, it's weird that we have .F which is the input and .f which is the output. So I would be in favor of renaming the input to something else as done here, provided that this is indeed rerunnable for partial compilations.

hoodmane:
Can we rename the output file instead?

katietz:
well, I am not sure which extension would be sane here ... but a .f77 might be the right thing?

katietz:
> Can we rename the output file instead?

Rename which output?  that from the preprocessor, or that one from f2c?  I assume you mean the first one, but not sure

katietz:
hmm, this doesn't work so well ... f2c complains that the file doesn't and with either '.F' or '.f' ... so renaming is indeed the way to go

hoodmane:
> f2c complains that the file doesn't end with either '.F' or '.f' 

I need a crying react.

katietz:
As the output file name is generated implicit by the input-file name ... (why there should be a option to specifiy different output-names ... to much comfort for a tool) ... a rename is not necessary what we can use here either ... but ... there's still one arrow in the quiver... it can read from stdin, and then outputs to stdout ... let me see if we can use this

katietz:
found a solution by reading from standard-in and writing to standard-out

pyodide-build/pyodide_build/pywasmcross.py
```diff
@@ -112,8 +112,10 @@ def replay_f2c(args: list[str], dryrun: bool = False) -> list[str] | None:
                 if arg.endswith(".F"):
                     # .F files apparently expect to be run through the C
                     # preprocessor (they have #ifdef's in them)
-                    # Use gfortran frontend, as gcc frontend might not be
-                    # present ...
+                    # The file-system might be not case-sensitive,
+                    # so take care to handle this by renaming.
```
hoodmane, line 116: Don't we want to keep the comment about `gfortran`? Maybe also mention osx.
```suggestion
                    # Use gfortran frontend, as gcc frontend might not be
                    # present on osx
                    # The file-system might be not case-sensitive,
                    # so take care to handle this by renaming.
```

pyodide-build/pyodide_build/pywasmcross.py
```diff
@@ -11,6 +11,7 @@
 import json
 import os
 import re
+import shutil
```
hoodmane, line 14: Put `shutil` import back below the prelude where it was before. While you're at it could you move `import re` down there too?
```suggestion
```

pyodide-build/pyodide_build/pywasmcross.py
```diff
@@ -134,8 +136,16 @@ def replay_f2c(args: list[str], dryrun: bool = False) -> list[str] | None:
                 # Fortran files, see
                 # https://github.com/xianyi/OpenBLAS/pull/3539#issuecomment-1493897254
                 # for more details
-                subprocess.check_call(["f2c", "-R", filepath.name], cwd=filepath.parent)
+                with open(filepath) as input_pipe:
+                    with open(filepath.with_suffix(".c"), "w") as output_pipe:
```
hoodmane, line 140:
```suggestion
                with (
                    open(filepath) as input_pipe,
                    open(filepath.with_suffix(".c"), "w") as output_pipe,
                ):
```

pyodide-build/pyodide_build/pywasmcross.py
```diff
@@ -134,8 +136,16 @@ def replay_f2c(args: list[str], dryrun: bool = False) -> list[str] | None:
...
+                        subprocess.check_call(
+                            ["f2c", "-R"],
+                            stdin=input_pipe,
+                            stdout=output_pipe,
+                            cwd=filepath.parent,
+                        )
                 fix_f2c_output(arg[:-2] + ".c")
+                # if files where renamed, restore old naming
```
hoodmane, line 148:
```suggestion
```

hoodmane (approved):
Thanks @katietz, this generally looks good to me. I have a few nitpicks.

pyodide-build/pyodide_build/pywasmcross.py
katietz: done

Issue #3847, opened 2023-05-09
Deno test lockfile validation fails CI


rajsite:
## 🐛 Bug

Looks like the current pattern for deno lock file validation does not work well for npm based dependencies.

The comparison lockfile is created from the source files but unlike esm based modules from deno the source files do not tend to point to specific versions of libraries. This results in the comparison lock file showing the latest versions of npm packages while the intention is that the lockfile captures new versions of dependencies added to the project.

Possible solutions:
- Maybe import maps can be used to change what the specifiers resolve to?

### To Reproduce

See [example build](https://github.com/pyodide/pyodide/actions/runs/4918213538/jobs/8784916600) reproduced below:

```
Run cd src/test-deno && deno task test
Download https://registry.npmjs.org/base-6[4](https://github.com/pyodide/pyodide/actions/runs/4918213538/jobs/8784916600#step:5:5)
Download https://registry.npmjs.org/node-fetch
Download https://registry.npmjs.org/pyodide
Download https://registry.npmjs.org/tr46
Download https://registry.npmjs.org/webidl-conversions
Download https://registry.npmjs.org/whatwg-url
Download https://registry.npmjs.org/ws
Task test deno task setup && deno task smoke-test
Task setup deno task cache:validate && deno task node-modules:clear && deno task node-modules:setup && deno task node-modules:setup-local-pyodide
Task cache:validate deno cache --lock=/tmp/pyodide-deno.lock --lock-write smoke-test.ts && diff -u deno.lock /tmp/pyodide-deno.lock
Download https://registry.npmjs.org/encoding
Download https://registry.npmjs.org/bufferutil
Download https://registry.npmjs.org/utf-8-validate
Download https://registry.npmjs.org/ws/-/ws-8.13.0.tgz
Download https://registry.npmjs.org/tr46/-/tr46-0.0.3.tgz
Download https://registry.npmjs.org/pyodide/-/pyodide-0.23.1.tgz
Download https://registry.npmjs.org/base-64/-/base-64-1.0.0.tgz
Download https://registry.npmjs.org/webidl-conversions/-/webidl-conversions-3.0.1.tgz
Download https://registry.npmjs.org/node-fetch/-/node-fetch-2.6.10.tgz
Download https://registry.npmjs.org/whatwg-url/-/whatwg-url-[5](https://github.com/pyodide/pyodide/actions/runs/4918213538/jobs/8784916600#step:5:6).0.0.tgz
[...]

PR #3848, opened 2023-05-09
Skip deno lock validation

rajsite:
### Description

Deno lockfile validation is causing a build failures. See https://github.com/pyodide/pyodide/issues/3847

This PR disables the lock file validation temporarily to prevent CI errors.

### Checklists

- Add a [CHANGELOG](https://github.com/pyodide/pyodide/blob/main/docs/project/changelog.md) entry. reason: CI only change
- Add / update tests
- Add new / update outdated documentation

### DIFF
--- a/src/test-deno/deno.json
+++ b/src/test-deno/deno.json
@@ -1,7 +1,7 @@
 {
   "tasks": {
     "test": "deno task setup && deno task smoke-test",
-    "setup": "deno task cache:validate && deno task node-modules:clear && deno task node-modules:setup && deno task node-modules:setup-local-pyodide",
+    "setup": "deno task node-modules:clear && deno task node-modules:setup && deno task node-modules:setup-local-pyodide",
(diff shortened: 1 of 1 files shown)


hoodmane:
Ideally I would like the test to use locked versions of dependencies. So if there is a new version of some package, we don't use it we use the version in the lock file. Isn't this the point of lock files?

Anyways just disabling validation and using the newest version is okay in the meantime but eventually the CI will get broken by a package update unless we lock the dependencies correctly.

rajsite:
> Ideally I would like the test to use locked versions of dependencies. So if there is a new version of some package, we don't use it we use the version in the lock file. Isn't this the point of lock files?
> 
> Anyways just disabling validation and using the newest version is okay in the meantime but eventually the CI will get broken by a package update unless we lock the dependencies correctly.

FWIW it'll at most use the latest defined by pyodide's package.json version specifiers but yea Deno lock files seem weird. I'll keep trying to understand it in #3847

PR #3853, opened 2023-05-13
Add js array methods to pyproxy of list

### LINKED ISSUE #3849: Proxied JS method.apply(context, list) fail while method(*list) doesn't
## 🐛 Bug

While working on [this PyScript issue](https://github.com/pyscript/pyscript/pull/1459) I've noticed that `context.method.apply(context, list)` doesn't work while `context.method(*list)` does.

I don't mind using the latter as that's also more Pythonic but that might surprise JS developers using Pyodide proxies that mimic JS APIs.

### To Reproduce

```python
import js
classList = js.document.body.classList
classList.add.apply(classList, ["a", "b"])
```

### Expected behavior

The method should be invoked with *n* arguments as by JS specs.

### Environment

- Pyodide Version: latest
- Browser version: any
- Any other relevant information: nope

### Additional context

Happy to have it as won't fix but at least there's a related issue that explain the *gotcha*.

hoodmane:
The goal of this PR is to work towards making a pyproxy of a list a drop-in replacement for an array. 
This resolves #3849.

- [x] Separate flags for MutableSequence and for Sequence (and for list if needed)
- [x] Adjust the set handler too
- [x] Add / update tests


src/core/pyproxy.ts
```diff
@@ -1437,6 +1445,205 @@ export class PyAsyncGeneratorMethods {
...
+  }
+  if (astr < bstr) {
+    return -1;
+  }
+  return 1;
+}
+
+// Missing:
+// flatMap, flat,
+export class PySequenceMethods {
```
ryanking13, line 1472: Do we have tests for each methods here?

src/core/pyproxy.ts
```diff
@@ -1437,6 +1445,205 @@ export class PyAsyncGeneratorMethods {
...
+    const cmp_to_key = functools.cmp_to_key;
+    let cf: (a: any, b: any) => number;
+    if (compareFn) {
+      cf = compareFn;
+    } else {
+      cf = defaultCompareFunc;
+    }
+    // spec says arguments to compareFunc "Will never be undefined."
+    // and undefined values should get sorted to end of list.
+    // Make wrapper to ensure this
```
ryanking13, line 1593: Hah, didn't know this.

ryanking13 (approved):
Thanks, @hoodmane. It generally looks good to me.

src/core/pyproxy.ts
hoodmane: Not yet, it's a todo...

PR #3855, opened 2023-05-13
Docs: recommend rustup `stable` when building Rust packages

jtpio:
### Description

Apparently it might not be needed to set `rustup` to target `nightly` anymore.

cc @manonmarchand who was recently looking into this

### Checklists

- [ ] Add / update tests

### DIFF
--- a/docs/development/new-packages.md
+++ b/docs/development/new-packages.md
@@ -381,7 +381,7 @@ imports are synchronous so it is impossible to load `.so` files lazily.
 We currently build `cryptography` which is a Rust extension built with PyO3 and
 `setuptools-rust`. It should be reasonably easy to build other Rust extensions.
 If you want to build a package with Rust extension, you will need Rust >= 1.41,
-and you need to set the rustup toolchain to `nightly`, and the target to
+and you can set the rustup toolchain to `stable`, and the target to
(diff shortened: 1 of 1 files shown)


ManonMarchand:
Thanks! Let me check for the lastest pyodide version before merging cause my tests are from a few months ago. We could also pin the first compatible version to allow more flexibility than "lastest"?

hoodmane:
It is still needed to target nightly because of [`-Z link-native-libraries=no`](https://github.com/pyodide/pyodide/blob/main/Makefile.envs#L170). The problem is that if you use `emcc blah.o -o mylib.so -sSIDE_MODULE -lc` then the link fails. `-Z link-native-libraries=no` is an unstable-only option that prevents rust from passing `-lc` to the linker. It would be nice to get the Emscripten linker to not fail in this case. As I understand it linking using clang or gcc with `-shared -lc` does nothing at all, since `libc` is linked by default the linker just ignores `-lc`. 

@sbc100 would you be willing to accept a patch to fix this? It would indeed be nice to use stable rust.

ManonMarchand:
When trying again, actually even the beta version of the toolchain does not have the `link-native-libraries` in the official flags.

Issue #398, opened 2019-04-18
pyodide - requests module is not available


andresn:
Any chance you'll add this? I'm looking to bring in data from Google Sheets.

```
Error: Traceback (most recent call last):
  File "/lib/python3.7/site-packages/pyodide.py", line 42, in eval_code
    exec(compile(mod, '<exec>', mode='exec'), ns, ns)
  File "<exec>", line 3, in <module>
ModuleNotFoundError: No module named 'requests'
```

hamilton:
@andresn FYI I moved this over to the Pyodide repository ~

kwmiebach:
If the requests module is not available, is there any other way to fetch data over http in pyodide?

gijswobben:
How about something like this:

```
import json
from js import window

print(
    window.fetch("https://jsonplaceholder.typicode.com/todos/1")
    .then(lambda response: response.text())
    .then(lambda data: print(json.dumps(json.loads(data), indent=4)))
)
```

preetiverma525:
Is there any way to use the post api in pyodide

rth:
> is there any other way to fetch data over http in pyodide?

Yes, with [`pyodide.open_url`](https://pyodide.readthedocs.io/en/latest/python-api/pyodide.open_url.html#pyodide-open-url) or using Web APIs directly e.g. with `window.fetch` (as mentioned above) or XMLHttpRequest (see `open_url` [implementation](https://github.com/iodide-project/pyodide/blob/8c22c98a43e228c213f3f1dd7eae79ac88da4e90/src/pyodide.py#L14))

> Is there any way to use the post api in pyodide

Similarly you [can make POST API calls](https://stackoverflow.com/a/64789621/1791279) with `XMLHttpRequest`  (ref https://github.com/iodide-project/pyodide/issues/797)

Perzan:
Maybe a requests implementation could be made on top of these. Any thoughts?

hoodmane:
@Perzan the main issue is that the `requests` is synchronous whereas `fetch` is asynchronous. It is possible to make synchronous calls with `XMLHttpRequest`, but it can't support very many options and is not very efficient. 

[...]

PR #3858, opened 2023-05-15
Add scikit-bio v0.5.8

sminot:
### Description

I would like to add a recipe for scikit-bio (https://scikit-bio.org/), which provides a number of extremely useful functions for bioinformatics.
Because the dependencies for scikit-bio largely mirror those for scikit-learn (which has already been built for Pyodide), I was hoping that this would be relatively straightforward.

### Checklists

- [ ] Add / update tests

### DIFF
--- a/docs/project/changelog.md
+++ b/docs/project/changelog.md
@@ -37,7 +37,7 @@ myst:
   {pr}`3331`.
 - New packages: sourmash {pr}`3635`, screed {pr}`3635`, bitstring {pr}`3635`,
   deprecation {pr}`3635`, cachetools {pr}`3635`, xyzservices {pr}`3786`,
-  simplejson {pr}`3801`, protobuf {pr}`3813`.
+  simplejson {pr}`3801`, protobuf {pr}`3813`, scikit-bio {pr}3858.
 - Upgraded libmpfr to 4.2.0 {pr}`3756`.
 - Upgraded scipy to 1.10.1 {pr}`3794`
 
--- a/packages/scikit-bio/meta.yaml
+++ b/packages/scikit-bio/meta.yaml
@@ -20,7 +20,7 @@ requirements:
     - scipy
     - joblib
     - threadpoolctl
-    - requests
+    - pyodide_http
 
 test:
   imports:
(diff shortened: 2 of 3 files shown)


lesteve:
scikit-bio depends on requests which is not packaged inside Pyodide, see https://github.com/pyodide/pyodide/issues/398. I am not sure about the way forward, others may jump in with suggestions ...

Maybe (probably a bit naive) requests and pyodide-http could be packaged in Pyodide so that the requests package inside Pyodide kind of works?

For more context, in a Pyodide console you can do things like:
```py
import micropip
await micropip.install(['requests', 'pyodide-http'])

import requests
import pyodide_http
pyodide_http.patch_all()

url = 'https://cdn.jsdelivr.net/pyodide/v0.23.1/full/pyodide.js'
# this will work with some limitations e.g. CORS headers need to be set by the server
r = requests.get(url)
print(r.content[:1000])
```

rth:
Moving `import requests` into the two methods that need it should be enough. That way it wouldn't fail at import time if requests is not installed, and we could treat it as an optional dependency. For using that functionality in Pyodide one could then indeed micropip install `pyodide-http` and `requests`. 

This can be done initially via [a patch in this PR](https://pyodide.org/en/stable/development/new-packages.html#generating-patches), but ideally yes, should be done upstream. It wouldn't have any impact on users outside of Pyodide.

sminot:
Thank you for providing more helpful details, @rth, I really appreciate it.

I'm more than happy to patch upstream, but I'm afraid that I don't completely understand the change that I would make to the code so that the `requests` package is patched appropriately by the `pyodide_http.patch_all()` method. Is there an easy way to modify the code so that it will work equally within or outside of Pyodide?

rth:
You don't need to do anything about that. Normally once one installs `requests` and `pydide-http` it should just work  after one runs `pyodide_http.patch_all()`. Or at least that's what `pydide-http`  promises, worth double checking in this particular casr.

The reason for the necessary change is that we currently can't specify  `requests` as a mandatory requirement in Pyodide. So scikit-bio needs to be importable even if requests is not installed https://github.com/pyodide/pyodide/issues/3876

agriyakhetarpal:
Hi @sminot, is this PR still relevant to you? `requests` is now available with Pyodide – thinks should work smoothly, unless there are new dependencies that need to be added.

sminot:
Thanks for checking in. This is no longer a pressing issue, especially with `requests` available. I'll close this out.

PR #3859, opened 2023-05-17
Update pytest-pyodide to 0.52

hoodmane:


ryanking13:
https://github.com/pyodide/pyodide/blob/63b4af4da965413fc1cb4e97d768cc852c6a9de0/src/tests/test_package_loading.py#L68-L69

I guess we need to rename test.html there to something else, then explicitly use `selenium.goto` to get that page.

hoodmane:
Benchmark stack size flakes a lot these days...

PR #3863, opened 2023-05-17
Make Pyodide work a bit better with deno

hoodmane:
I'm not sure all of these changes are good. We'll see what the test suite thinks.

cc @rajsite

### DIFF
--- a/src/js/compat.ts
+++ b/src/js/compat.ts
@@ -1,12 +1,17 @@
 // Detect if we're in node
 declare var process: any;
+declare var Deno: any;
 
+export const IN_DENO = typeof Deno !== undefined;
+const node_prefix = IN_DENO ? "node:" : "";
+const npm_prefix = IN_DENO ? "npm:" : "";
 export const IN_NODE =
-  typeof process === "object" &&
-  typeof process.versions === "object" &&
-  typeof process.versions.node === "string" &&
-  typeof process.browser ===
-    "undefined"; /* This last condition checks if we run the browser shim of process */
+  IN_DENO ||
+  (typeof process === "object" &&
+    typeof process.versions === "object" &&
+    typeof process.versions.node === "string" &&
+    typeof process.browser ===
+      "undefined"); /* This last condition checks if we run the browser shim of process */
 
 let nodeUrlMod: any;
 let nodeFetch: any;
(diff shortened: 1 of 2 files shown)


src/js/streams.ts
hoodmane, line 1: Getting rid of `Buffer` makes `streams.ts` at least sort of work -- we can at least write to stdout if we polyfill `process`.

src/js/compat.ts
```diff
@@ -96,11 +116,7 @@ if (IN_NODE) {
  * In Windows, it's \.
  * @private
  */
-export let pathSep: string;
-
-if (!IN_NODE) {
-  pathSep = "/";
-}
+export let pathSep = "/";
```
hoodmane, line 119: This minor simplification doesn't change the behavior.

hoodmane:
I polyfilled process like:
```js
import { hrtime, stdin, stdout, stderr, env, exit, cwd, argv, platform, versions } from 'node:process';

globalThis.process = {
    release: {name: "node"},
    versions,
    exit, env, cwd, argv, platform,
    hrtime, stdin, stdout, stderr,
	binding: () => ({})
}
```
maybe some form of this could go into compat too?

src/js/compat.ts
```diff
@@ -31,48 +34,61 @@ export async function initNodeModules() {
...
-  const ws = await import("ws");
-  const child_process = await import("child_process");
+  const fs = await import("node:fs");
+  const crypto = await import("node:crypto");
+  // @ts-ignore
+  const tty = await import("node:tty");
+  // Stub these out in deno because otherwise I get:
+  // TypeError: Loading unprepared module: npm:ws
+  const ws = IN_DENO ? {} : await import("ws");
+  const child_process = IN_DENO ? {} : await import("child_process");
```
ryanking13, line 73: Interesting. Does it mean we are not using these two modules actually?

rajsite: Deno tries to be more like a browser context. It has globals for the WHATWG apis like WebSocket, URL, and fetch already in the globalThis. Shouldn't need the npm packages in those cases

hoodmane: Well they are pretty seldomly used. `ws` is used by `-lwebsocket.js`. `child_process` is used by `system` / `execl`. It should be possible to import them in deno, but I am having trouble getting them to work, and working minus some rare features is probably better than not working at all.

hoodmane: > Shouldn't need the npm packages / node imports in those cases

Well sure, but only if we are willing to write deno-specific code. In the short term, it would be nice to go through the node polyfills unless there is a pressing reason not to. Longer term maybe we can add deno-specific code, but as long as we are only building one version of Pyodide any deno-specific code also has to be downloaded but not used by browsers so there are reasons not to duplicate if there is a way to make it work.

PR #3864, opened 2023-05-17
Remove in-tree dependency when building unvendored python modules

ryanking13:
This PR is a preparation to move recipes to other repository (#3827).

This makes unvendored python module recipes use CPython archive from remote URL instead of using in-tree CPython source tree.


packages/distutils/meta.yaml
```diff
@@ -5,9 +5,10 @@ package:
     - always
   top-level:
     - distutils
+source:
+  sha256: 065040198fde6fd7229b8c0538fc8b848824e35930e3904f46c9e90e9f0f1a3f
+  url: https://github.com/python/cpython/archive/refs/tags/v$(PYVERSION).tar.gz
```
ryanking13, line 10: Unfortunately, sha256 hash needs to be updated manually when we update CPython version...

packages/hashlib/meta.yaml
```diff
@@ -5,11 +5,12 @@ package:
     - always
   top-level:
     - _hashlib
+source:
+  sha256: 065040198fde6fd7229b8c0538fc8b848824e35930e3904f46c9e90e9f0f1a3f
+  url: https://github.com/python/cpython/archive/refs/tags/v$(PYVERSION).tar.gz
```
rth, line 10: Are you sure it's safe to take these from the original archive and compile them without running the configure step? Isn't there a risk something would be badly configured then?

packages/test/patches/0005-gh-93839-Move-Lib-ctypes-test-to-Lib-test-test_ctype.patch
```diff
@@ -0,0 +1,753 @@
+From d82e0bfe8b98a122ca443b356d81998c804b686e Mon Sep 17 00:00:00 2001
+From: Victor Stinner <vstinner@python.org>
+Date: Tue, 21 Jun 2022 10:24:33 +0200
+Subject: [PATCH 5/9] gh-93839: Move Lib/ctypes/test/ to Lib/test/test_ctypes/
+ (#94041)
+
+* Move Lib/ctypes/test/ to Lib/test/test_ctypes/
+* Remove Lib/test/test_ctypes.py
```
rth, line 8: Could you explain why the above two are necessary? Can't we keep it in that folder?

packages/distutils/meta.yaml
rth: We could put it in an env variable otherwise? For 7 manual replacements, it's probably worth it...

rth (commented):
Thanks for working on this!

packages/distutils/meta.yaml
ryanking13: Sound good.

packages/hashlib/meta.yaml
ryanking13: That's a good point. I checked that the files are not changed before/after configure (with [this simple script](https://gist.github.com/ryanking13/2b8ab6af2b9c8423f58b0394e311b092)). And we are running full CPython test suites for unvendored packages for hopefully it is okay...

packages/test/patches/0005-gh-93839-Move-Lib-ctypes-test-to-Lib-test-test_ctype.patch
ryanking13: These patches are necessary to run ctypes, unittest tests. We apply these patches when building CPython, so it also needs to be applied here (https://github.com/pyodide/pyodide/pull/3507). This is a bit annoying, but we will be able to remove these patches when we update to Python 3.12.

Issue #3839, opened 2023-05-08
pyodide build breaks #include "errcode.h"


julian-smith-artifex-com:
## 🐛 Bug

`pyodide build` seems to add `emsdk/.pyodide-xbuildenv/xbuildenv/pyodide-root/cpython/installs/python-3.11.2/include/python3.11` to the include path, which results in `#include "errcode.h"` picking up the Python header, and not a local project (tesseract in this case)'s `errcode.h`.

### To Reproduce

Build a Python package with  a `.cpp` file that does `#include "errcode.h"` to include a local file `errcode.h`.

### Expected behavior

Package should build correctly, `#include`ing the local `errcode.h` file.

### Environment

Linux (Devuan).

- Pyodide Version:
- pyodide_build-0.23.2
- pyodide_cli-0.2.3

- emsdk commit b113f24842c6e "3.1.37 (#1212)"

- Browser version:
Not relevant.

- Any other relevant information:

### Additional context

I'm building Python package PyMuPDF. The problem occurs when PyMuPDF's `setup.py` itself builds MuPDF, a C PDF library, when it builds Tesseract.

A workaround i have used is to replace all occurrencies of `#include "errcode.h"` in MuPDF's Tesseract code with `#include "../ccutil/errcode.h"`, which i think bypasses the default include path set up by `pyodide build`.

ryanking13:
Thanks for the report @julian-smith-artifex-com.

> pyodide build seems to add emsdk/.pyodide-xbuildenv/xbuildenv/pyodide-root/cpython/installs/python-3.11.2/include/python3.11 to the include path

Yes we append Python header files to include path.

Could you please share us how MuPDF specifies the include path? As far as I know, llvm-cc (and gcc) searches header files left-to-right order and we append `-I<python_headers_directory>` at the right-end of the complier command, so I think it should search local header files first.

julian-smith-artifex-com:
Thanks for taking a look.

Here's the failing compile in the build output:

```
[...]

PR #3866, opened 2023-05-18
Append extra commandline args after original args

ryanking13:
### Description

Resolve: #3839 

Switch the order of commandline args in pywasmcross so that custom arguments that we add gets appended after the original arguments.

### Checklists

- [x] Add / update tests


hoodmane:
Nice reduction in the optflags / debugflags logic too.

PR #3868, opened 2023-05-18
Add special case handling for DomException

hoodmane:
DomException is a weird special case. According to WHATWG, there are two types of Exception objects, simple exceptions and DOMExceptions. The spec says:

> if an implementation gives native Error objects special powers or
> nonstandard properties (such as a stack property), it should also expose
> those on DOMException objects

Firefox respects this and has DomException.stack. But Safari and Chrome do not. Hence we need a special check for DomException.


src/tests/test_pyodide.py
```diff
@@ -583,6 +583,17 @@ def test_run_python_js_error(selenium):
     )
 
 
+@run_in_pyodide
```
ryanking13, line 586: Should be skipped in Node.

ryanking13 (approved):

rth (approved):
Thanks!

PR #3870, opened 2023-05-18
Add an argument to loadPyodide to pass environment variables

hoodmane:
This also updates the command line runner to pass in all ambient environment variables.

`homedir` is now deprecated. I added handling for `homedir` and `env.HOME`: if both are passed, raise an error, otherwise they mean the same thing. 

Resolves #3869.

- [x] Add / update tests


ryanking13 (approved):
Thanks, @hoodmane!

> I added handling for homedir and env.HOME: if both are passed, raise an error, otherwise they mean the same thing.

Maybe we can deprecate `homddir` argument? Setting HOME env variable seems more natural way to set the home directory.

docs/project/changelog.md
```diff
@@ -45,6 +45,11 @@ myst:
   new array with the result values from the iterable appended.
   {pr}`3904`
 
+- {{ Enhancement }} It is now possible to pass environment variables to
+  `loadPyodide` via the `env` argument. `HOMEDIR` is deprecated in favor of
```
ryanking13, line 49:
```suggestion
  `loadPyodide` via the `env` argument. `homedir` is deprecated in favor of
```

ryanking13 (approved):

PR #3871, opened 2023-05-18
Export emscripten specific sysconfigdata to host's python

katietz:
### Description

we need to make sure that host's python has the emscripten related sysconfigdata-file at its python-related location installed, so that later use of _PYTHON_SYSCONFIGDATA_NAME is actually working as expected.


katietz:
agree on that copying it to python's lib folder might be problematic.  Copying it instead into side-packages might be an alternative here.

katietz:
well, in context of other virtual environment (like conda), it needs to be copied to python's prefix too.  But I agree it is an issue about permissions issues.  But well, actually we might want to copy it first into python's site-packages, which should be actually always writable, if pyodide-build is instealled.  So we would have at least a fixed place we could copy it from

ryanking13:
I'm sorry, but I don't fully understand, can you explain a bit more? I guess it could be because I don't know conda very well.

By the way, if the sysconfigdata is used only when building packages, I think the copying should be done in this part of [pypabuild.py](https://github.com/pyodide/pyodide/blob/fd2959f7dcb7b000685cb7f3a2c8e72f518bdd74/pyodide-build/pyodide_build/pypabuild.py#L41-L59). Because those who build packages out-of-tree won't be using makefiles.

ryanking13:
Closing as no activity.

PR #3883, opened 2023-05-29
Merge initialization functions for in-tree and out-of-tree builds

ryanking13:
### Description

This is another split off of #3737.

Merges two private pyodide-build APIs `init_environment` and `initialize_pyodide_root` into a single a single API: `init_environment`.
Regardless of whether we are doing in-tree or out-of-tree build, `init_environment` will do the correct thing.

### Checklists

- [x] Add / update tests


pyodide-build/pyodide_build/common.py
```diff
@@ -232,15 +232,14 @@ def get_build_environment_vars() -> dict[str, str]:
     # TODO: Add modifiable configuration file instead.
     # (https://github.com/pyodide/pyodide/pull/3737/files#r1161247201)
     env.update({key: os.environ[key] for key in BUILD_VARS if key in os.environ})
-
     env["PYODIDE"] = "1"
 
-    if "PYODIDE_JOBS" in os.environ:
```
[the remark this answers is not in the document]
ryanking13: I think I removed it by mistake while resolving merge conflicts.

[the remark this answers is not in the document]
hoodmane: I'm not really sure what it does but if we wanted to remove it we'd need to discuss it more specifically.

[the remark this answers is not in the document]
ryanking13: When building libraries, `PYODIDE_JOBS` is passed as a commandline parameter of `make`. I'm not sure how much it affects the build performance, but yeah we discuss it later.

pyodide-build/pyodide_build/tests/test_build_env.py
```diff
@@ -0,0 +1,165 @@
...
+    def test_get_build_flag(self, reset_env_vars, reset_cache):
+        for key, val in common.get_build_environment_vars().items():
+            assert common.get_build_flag(key) == val
+
+        with pytest.raises(ValueError):
+            common.get_build_flag("UNKNOWN_VAR")
+
+
+class TestOutOfTree(TestInTree):
+    # TODO: selenium fixture is a hack to make these tests run only after building Pyodide.
```
hoodmane, line 109: At this point we really ought to make some dedicated fixtures for this since we do this a lot.

pyodide-build/pyodide_build/install_xbuildenv.py
```diff
@@ -96,6 +96,15 @@ def install(path: Path, *, download: bool = False, url: str | None = None) -> No
...
-        download_xbuildenv(version, path, url=url)
+        if path.exists():
+            logger.warning("xbuild environment already exists, skipping download")
+        else:
+            _download_xbuildenv(version, path, url=url)
+    else:
+        if not path.exists():
+            logger.error("xbuild environment not exists")
+            raise FileNotFoundError(path)
+
```
hoodmane, line 109: Nitpick maybe this is slightly neater:
```suggestion
    if not download and not path.exists():
        logger.error("xbuild environment not exists")
        raise FileNotFoundError(path)

    if download and path.exists():
        logger.warning("xbuild environment already exists, skipping download")
    elif download:
        _download_xbuildenv(version, path, url=url)
```

pyodide-build/pyodide_build/common.py
```diff
@@ -348,18 +350,68 @@ def search_pyodide_root(curdir: str | Path, *, max_depth: int = 5) -> Path:
...
+    xbuildenv_path = Path(".pyodide-xbuildenv").resolve()
+
+    if xbuildenv_path.exists():
+        os.environ["PYODIDE_ROOT"] = str(xbuildenv_path / "xbuildenv" / "pyodide-root")
+        return
+
+    with ExitStack() as stack:
+        if quiet:
+            # Prevent writes to stdout
+            stack.enter_context(redirect_stdout(StringIO()))
```
hoodmane, line 410: I know this logic was only moved by this PR, but I learned the following pattern recently which I think is a bit tidier:
```suggestion
    from contextlib import nullcontext
    if quiet:
       context = redirect_stdout(StringIO())
    else:
       context = nullcontext()
    with context:
```

pyodide-build/pyodide_build/common.py
```diff
@@ -348,18 +350,68 @@ def search_pyodide_root(curdir: str | Path, *, max_depth: int = 5) -> Path:
...
+    #    set PYODIDE_ROOT to the root of the Pyodide repository
+    try:
+        os.environ["PYODIDE_ROOT"] = str(search_pyodide_root(Path.cwd()))
+        return
+    except FileNotFoundError:
+        pass
+
+    # 3) If we are doing an out-of-tree build,
+    #    download and install the Pyodide build environment
+    xbuildenv_path = Path(".pyodide-xbuildenv").resolve()
```
hoodmane, line 401: Not directly connected to this PR, but it would be a good idea to include the Pyodide version here. It's confusing when an xbuildenv from the wrong Pyodide version is used.

pyodide-build/pyodide_build/common.py
```diff
@@ -348,18 +350,68 @@ def search_pyodide_root(curdir: str | Path, *, max_depth: int = 5) -> Path:
...
     if "sphinx" in sys.modules:
         os.environ["PYODIDE_ROOT"] = ""
+        return
 
+    # 1) If PYODIDE_ROOT is already set, do nothing
     if "PYODIDE_ROOT" in os.environ:
-        os.environ["PYODIDE_ROOT"] = str(Path(os.environ["PYODIDE_ROOT"]).resolve())
-    else:
-        os.environ["PYODIDE_ROOT"] = str(search_pyodide_root(os.getcwd()))
+        return
```
hoodmane, line 389: I think this guard should be the very first thing in the function body. If `_set_pyodide_root` is only called from `init_environment` this might be redundant to the guard on `init_environment`. Though perhaps it is still good for readability.

pyodide-build/pyodide_build/tests/test_common.py
```diff
@@ -249,29 +248,23 @@ def test_repack_zip_archive(
     assert input_path.stat().st_size == expected_size
 
 
-def test_set_build_environment(monkeypatch):
+def test_get_build_environment_vars_host_env(monkeypatch):
     import os
 
-    monkeypatch.delenv("PKG_CONFIG_PATH", raising=False)
-    monkeypatch.delenv("WASM_PKG_CONFIG_PATH", raising=False)
```
hoodmane, line 256: I think we should keep a test that there is no error when both of these are undefined.

pyodide-build/pyodide_build/common.py
```diff
@@ -348,18 +350,68 @@ def search_pyodide_root(curdir: str | Path, *, max_depth: int = 5) -> Path:
...
+    """
+    Initialize Pyodide build environment.
+    This function needs to be called before any other Pyodide build functions.
+    """
     if os.environ.get("__LOADED_PYODIDE_ENV"):
         return
+
     os.environ["__LOADED_PYODIDE_ENV"] = "1"
+
+    _set_pyodide_root(quiet=quiet)
```
hoodmane, line 363: Is `_set_pyodide_root` ever called from anywhere else? Or is the split purely for readability? Might be good to mention.

hoodmane (approved):
Thanks @ryanking13! I have some nitpicks but otherwise it looks great.

pyodide-build/pyodide_build/common.py
ryanking13: It's purely for readability and `_set_pyodide_root` is not supposed to be called in other places. I wanted to split the build environment initialization into several steps. Will add some comments.

pyodide-build/pyodide_build/common.py
ryanking13: I think the `__LOADED_PYODIDE_ENV` guard in `init_environment` is a bit redundant, but I didn't want to introduce an unexpected side effect. I'll check it in a subsequent PR.

pyodide-build/pyodide_build/tests/test_build_env.py
ryanking13: Yeah... or maybe we should find a way to run these tests without building Pyodide.

pyodide-build/pyodide_build/tests/test_common.py
ryanking13: Actually, I realized that we don't need to separate WASM_PKG_CONFIG_PATH and PKG_CONFIG_PATH anymore since `get_build_environment_vars` does not pollute host environment variables anymore, and we have a logic that the host environment variable takes precedence over variables in Makefile.envs.

pyodide-build/pyodide_build/common.py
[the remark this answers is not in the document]
ryanking13: Oh, now I remember that I removed this because it is redundant.

`env.update({key: os.environ[key] for key in BUILD_VARS if key in os.environ})` will overwrite every variable that is defined in the host environment so we don't need to do it again for PYODIDE_JOBS.

PR #3890, opened 2023-05-31
Emscripten 3.1.39

hoodmane:


hoodmane:
@ryanking13 you okay with adding this patch for now until we find a better way to deal with this?

ryanking13:
Yeah, I am okay with adding this patch. This only affects the main module so it should be okay.

PR #3904, opened 2023-06-05
Add sequence methods to JsProxy of arrays

hoodmane:
Prior to this PR, internal Python code paths that use `PySequence_*` methods directly would fail on JS Arrays. To fix this, we have to implement `sq_length`, `sq_item` and (for mutable sequences) `sq_ass_item`.

#### Sequence methods:

- [x] sq_item
- [x] sq_ass_item
- [x] sq_length
- [x] sq_concat
- [x] sq_repeat
- [x] sq_inplace_repeat
- [x] sq_contains

#### Checklist

- [x] Add / update tests


src/py/_pyodide/_core_docs.py
hoodmane, line 1: Moved the types `JsArray`, `JsMap`, etc below `JsIterable` so they can inherit from `JsIterable`.

ryanking13 (approved):
Cool, thanks!

PR #3908, opened 2023-06-06
NFC Factor out common setup for rust packages

hoodmane:
This moves the common setup for Rust packages into pyodide build. It may need to be rearranged on #3817 so it is helpful not to repeat it.


pyodide-build/pyodide_build/buildpkg.py
```diff
@@ -776,7 +774,9 @@ def _build_package_inner(
             patch(pkg_root, srcpath, source_metadata)
 
         src_dist_dir.mkdir(exist_ok=True, parents=True)
-        run_script(build_dir, srcpath, build_metadata, bash_runner)
+        if "rustup" in pkg.requirements.executable:
```
ryanking13, line 777: There is a [helper method for this](https://github.com/pyodide/pyodide/blob/dd9d71df2ca075a1bcf6ffbf936c79420ebe98f1/pyodide-build/pyodide_build/buildall.py#L434) in buildall. Maybe we need to move that method to buildpkg to prevent circular imports between buildpkg and buildall.

ryanking13 (approved):
Thanks!

PR #3012, opened 2022-08-24
Add package `pyart` and `netcdf4`

mgrover1:
### Description

Package for working with weather radar data. Repo is https://github.com/ARM-DOE/pyart

Also added netcdf4 (closes https://github.com/pyodide/pyodide/issues/1582)

### Checklists

- [x] Add a [CHANGELOG](https://github.com/pyodide/pyodide/blob/main/docs/project/changelog.md) entry
- [x] Add / update tests

Issue #1582, opened 2021-05-06
include netcdf4 and h5py


yxdragon:
hdf and netcdf is two important format in geography and meteorology, request include it in pyodide.

rth:
> yes, zarr is more modern.
but for geography and meteorology, history data is very important, and most of history data is in hdf and netcdf.

Happy to include those, but someone would need to do the work on packaging them.

> another question about hdf/netcdf/zarr. When we read it from native, the reader hold a pointer to the file. And read data we need just in time. But in pyodide, a web environment, is the "just in time read" possible? I worry about browser would load the whole file as memory stream when we upload the file. (hdf/nc/zarr file may be very large)

Well if you need only a small slices of the dataset you would need to,
 - on the client: make an HTTP request indicating in some way the slice you want
 - on the server: either dynamically serve the appropriate slice for a given request, or have some per-determined chunking on your dataset. Or correctly respond to [HTTP range requests](https://developer.mozilla.org/en-US/docs/Web/HTTP/Range_requests)
 
In any case some work is necessary on your part: if you were previously doing just in time loads from disk you would need to translate that with Web APIs.

ax3l:
For HDF5: #1799

bmaranville:
@ax3l and I worked together and h5py is introduced in PR #2411 which is passing all the tests and almost ready to go!

agriyakhetarpal:
Both `netcdf4` (#3910) and `h5py` (#2411, linked against a WASM build for `libhdf5`) have been available with Pyodide for a few releases by now, so I'll close this issue; please don't hesitate to let us know in case you face any troubles with Pyodide or with either of the packages in particular! :)

Issue #771, opened 2020-10-19
not able to install geopy package 


preetiverma525:
Hi 

i need to install geopy package as i want to find the latitude and longitude of location in pyodide  but unable to install 
error no  module find geopy 

i already mention geopy in html page as well
 pyodide.loadPackage(['matplotlib','pandas','micropip','geopy']).then(() => {
         pyodide.runPython(pythonText);
  
         
      		
         });

please help

rth:
`geopy` is not included pyodide directly. You need to use micropip to install it from PyPi https://pyodide.readthedocs.io/en/latest/pypi.html  (that should work as geopy [has a pure python wheel](https://pypi.org/project/geopy/2.0.0/#files), with a filename ending in `none-any.whl`). You also need to do this for all of it's dependencies.

Don't hesitate to open issues or preferably post stack overflow questions with the [pyodide tag](https://stackoverflow.com/search?q=pyodide) if you have other questions.

yxdragon:
many package is release as .gz (not whl), So I mus download the source, and rebuild it as whl, and upload to my cloud disk. could micropip support .gz directly?

yxdragon:
some request: hdf and netcdf is two important science format, could pyodide include h5py and netcdf4?

rth:
> could micropip support .gz directly?

No unfortunately it's technically very difficult / not possible cf https://github.com/pyodide/pyodide/issues/1501#issuecomment-823447188. See also https://github.com/pypa/packaging-problems/issues/496 discussion for a way to improve this situation in the ecosystem.

> hdf and netcdf is two important science format, could pyodide include h5py and netcdf4?

Please open a separate issue. Assuming it's not to hard to build it should be possible. You can also use the zarr format as a more modern alternative (which is already packaged)

yxdragon:
yes, zarr is more modern. 
but for geography and meteorology, history data is very important, and most of history data is in hdf and netcdf.

yxdragon:
@rth another question about hdf/netcdf/zarr. When we read it from native, the reader hold a pointer to the file. And read data we need just in time. But in pyodide, a web environment, is the "just in time read" possible? I worry about browser would load the whole file as memory stream when we upload the file. (hdf/nc/zarr file may be very large)

rth:
Replied in https://github.com/pyodide/pyodide/issues/1582

PR #3910, opened 2023-06-06
Add `netcdf4` package

juntyr:
NetCDF4 has often been requested to be added to Pyodide (see #771, #1569, #1582, and #3012). As I got a build working in my local Pyodide version, I wanted to put this PR up as a draft in the hopes that others can perhaps improve upon the small hacks in building libnetcdf to get it working :)

### Checklists

- [x] Add / update tests

### DIFF
--- a/packages/netcdf4/test_netcdf4.py
+++ b/packages/netcdf4/test_netcdf4.py
@@ -1,3 +1,4 @@
+import pytest
 from pytest_pyodide import run_in_pyodide
 
 
--- a/packages/netcdf4/test_netcdf4.py
+++ b/packages/netcdf4/test_netcdf4.py
@@ -1,7 +1,7 @@
 from pytest_pyodide import run_in_pyodide
 
 
-@matplotlib_test_decorator
+@pytest.mark.driver_timeout(60)
 @run_in_pyodide(packages=["netCDF4", "numpy"])
 def test_netCDF4_tutorial(selenium):
     """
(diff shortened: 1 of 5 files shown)


juntyr:
?r @ryanking13

ryanking13:
Oh, sorry I think I missed this PR since it was in a draft state before 🙏.

Could you please add some basic tests for the netcdf4 package and update the changelog?

packages/netcdf4/test_netcdf4.py
```diff
@@ -0,0 +1,443 @@
+import pytest
+from pytest_pyodide import run_in_pyodide
+
+
+@pytest.mark.driver_timeout(60)
+@run_in_pyodide(packages=["netCDF4", "numpy"])
+def test_netCDF4_tutorial(selenium):
+    """
+    https://github.com/Unidata/netcdf4-python/blob/master/examples/tutorial.py
```
ryanking13, line 9: Great, but could you replace the print statements in this test with `assert` and compare the values, so that we can check if this library works correctly?

ryanking13 (commented):
Thanks. There was a CI failure unrelated to this PR (network issue), so I rerun the CI.

packages/netcdf4/test_netcdf4.py
juntyr: That's a good idea - in the end, I didn't convert each print individually but wrote an `assert_print` helper function that checks the (normalised) prints against the expected outputs.

ryanking13 (approved):
Thanks again @juntyr!

juntyr:
> Thanks again @juntyr!

Thank you @ryanking13 for your guidance and feedback!

PR #3916, opened 2023-06-10
MAINT Use sigstore to validate Python source signature

hoodmane:
This should make Python updates a little easier and more secure.

See: https://www.python.org/download/sigstore/

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -6,7 +6,7 @@ defaults: &defaults
     # Note: when updating the docker image version,
     #       make sure there are no extra old versions lying around.
     #       (e.g. `rg -F --hidden <old_tag>`)
-    - image: pyodide/pyodide-env:20230506-chrome112-firefox112-py311
+    - image: hoodmane/pyodide-env:20230509-chrome114-firefox114-py311
   environment:
     - EMSDK_NUM_CORES: 3
       EMCC_CORES: 3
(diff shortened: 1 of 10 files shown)


ryanking13 (commented):
Thanks, @hoodmane. 

I am +1 with the idea of making a download more secure, but this means that users now need to install sigstore and its dependencies to build Pyodide, which is a bit of a hassle I think.

Maybe we can run this only in CI so that users who build Pyodide locally don't need to install sigstore manually?

hoodmane:
>  this means that users now need to install sigstore and its dependencies to build Pyodide, which is a bit of a hassle I think.

It's just a pip install. What about making a custom virtualenv in the Makefile and using that?

ryanking13:
> It's just a pip install.

Yeah, it's just a pip install, but I like the way things are now, where all you need to build Pyodide (with core packages) is basic build tools and the right version of Python.

> What about making a custom virtualenv in the Makefile and using that?

Sounds good to me.

hoodmane:
This does make me wonder a bit what the threat model is here. We pip install from https://raw.githubusercontent.com/sigstore/sigstore-python/v1.1.2/install/requirements.txt
so we have to trust the githubusercontent domain and also the people with release access to sigstore. I guess this system protects us against `python.org` being compromised. But not against github being compromised. But it's not possible to protect us against github being compromised since we use it for all of our hosting...

hoodmane:
@ryanking13 We still use `PYTHON_ARCHIVE_SHA256` for unvendored stdlib modules. I think we should add a special key e.g.,
```yaml
   source:
      python_stdlib: true
```
or something for these.

hoodmane:
Actually I'm going to revert this: some of the dependencies of sigstore fail to build against Python 3.12b2 so this will make it harder to test prereleases.

PR #3922, opened 2023-06-12
Don't set PYODIDE_ROOT inside xbuildenv installation function

ryanking13:
This is some cleanup after #3883.

- Make the `PYODIDE_ROOT` be modified only inside the `init_environment` function.
- Removes `__LOADED_PYODIDE_ENV`. I think it is redundant as we use `PYODIDE_ROOT` for a similar purpose.


pyodide-build/pyodide_build/common.py
```diff
@@ -375,62 +375,37 @@ def init_environment(*, quiet: bool = False) -> None:
...
+def _init_xbuild_env(*, quiet: bool = False) -> Path:
     """
-    Set PYODIDE_ROOT environment variable.
-
-    This function works both in-tree and out-of-tree builds:
-    - In-tree builds: Searches for the root of the Pyodide repository in parent directories
-    - Out-of-tree builds: Downloads and installs the Pyodide build environment into the current directory
-
-    Note: this function is supposed to be called only in init_environment(), and should not be called directly.
+    Initialize the build environment for out-of-tree builds.
```
hoodmane, line 393: Could you add documentation of the return value to the docstring?

pyodide-build/pyodide_build/common.py
```diff
@@ -375,62 +375,37 @@ def init_environment(*, quiet: bool = False) -> None:
...
     quiet
         If True, do not print any messages
     """
-
     from . import install_xbuildenv  # avoid circular import
 
-    # If we are building docs, we don't need to know the PYODIDE_ROOT
-    if "sphinx" in sys.modules:
-        os.environ["PYODIDE_ROOT"] = ""
-        return
```
hoodmane, line 407: Don't we want to keep some version of this? I guess `search_pyodide_root` just works? why did we add this case in the first place?

hoodmane (approved):
Looks very nice.

pyodide-build/pyodide_build/common.py
ryanking13: > why did we add this case in the first place?

If I remember correctly, we needed this because the `sphinx_argparse_cli` extension was required to run the pyodide-build source code to create the docs of pyodide-build CLI. But we removed pyodide-build CLI from docs in favor of pyodide CLI, it is not needed anymore.

hoodmane: Ok, thanks for the explanation.

PR #3923, opened 2023-06-12
Update tskit to 0.5.5

hyanwong:
### Description

Tskit release 0.5.5 is a minor bugfix release containing a fix for for 32bit (emscripten) binaries. See https://github.com/tskit-dev/tskit/commit/332d5b7461ac895eedb9fd95c5e4bc4d28a817cb

### DIFF
--- a/packages/tskit/meta.yaml
+++ b/packages/tskit/meta.yaml
@@ -1,11 +1,11 @@
 package:
   name: tskit
-  version: 0.5.4
+  version: 0.5.5
   top-level:
     - tskit
 source:
-  url: https://files.pythonhosted.org/packages/ee/b3/137716b4ee755dff17928217922fe15880212b9ce2f3391ab77335caebb0/tskit-0.5.4.tar.gz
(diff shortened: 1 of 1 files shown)


hyanwong:
Test failure here doesn't seem related to the updated package, so is some other pyodide change, I assume.

hoodmane:
Yeah that's been failing on main.

hoodmane:
Thanks @hyanwong.

hyanwong:
This doesn't seem to be in the 0.23.3 release that came out 2 days ago? E.g. in the REPL:

```python
>>> from pyodide_js import version
  print(version)
0.23.3
>>> tskit.__version__  # should be 0.5.5
'0.5.4'
```

Is there anything I need to do to get this into the next release?

ryanking13:
> Is there anything I need to do to get this into the next release?

We will include it in the next 0.24.0 release. We usually don't include package version updates in the patch release, but in this case, I think probably we should've included it... as it includes a bugfix.

PR #3925, opened 2023-06-12
Make PyProxy of a callable an instanceof Function

### LINKED ISSUE #3924: Python Callables are not instanceof Function
## 🐛 Bug

Using the `instanceof` operator on PyProxy's of Python Functions (or other Callables) returns False.

### To Reproduce

All of the following result in `false`:

```js
const pyFunc_0 = pyodide.runPython(`
    lambda: print("zero")
`);

const pyFunc_1 = pyodide.runPython(`
    from pyodide.ffi import create_proxy
    create_proxy(lambda: print("one"))
`);

const pyFunc_2 = pyodide.runPython(`
    def foo():
        print("two")
    foo
`)

const pyFunc_3 = pyodide.runPython(`
    class A():
        def a(self):
            print("three") # method from class
    A.a
`)

const pyFunc_4 = pyodide.runPython(`
    class A():
        def a(self):
            print("four") # method from instance
    A().a
`)

const pyCallable_5 = pyodide.runPython(`
    class B():
        def __call__(self):
            print("five (B as a callable instance)")

    b = B()
    b
`)

console.log(pyFunc_0 instanceof Function);
console.log(pyFunc_1 instanceof Function);
console.log(pyFunc_2 instanceof Function)
console.log(pyFunc_3 instanceof Function)
console.log(pyFunc_4 instanceof Function)
console.log(pyCallable_5 instanceof Function)

// Call all the things, to verify they really are callable
pyFunc_0()
pyFunc_1()
pyFunc_2()
pyFunc_3('a')
pyFunc_4()
pyCallable_5()
```

### Expected behavior

I'd expect Python callables (or at least functions/lambdas) make `instanceof Function` true.

### Environment

- This the case in Pyodide:
  -  0.23
  -  0.22
  -  0.21
  -  0.20 (*with the previous import of `create_proxy` from just the `pyodide` module)
  -  0.19 (*with the indexURL supplied to `loadPyodide`)
  - *That's as far as I got; in 0.18 and earlier, this code throws a fatal error: `TypeError: Module._run_python_simple_inner is not a function`

hoodmane:
Resolves #3924. Make a copy of the PyProxy class which is a subclass of Function. Since the difference between `PyProxy` and `PyProxyFunction` is an implementation detail, we use `[Symbol.hasInstance]` to make them recognize the same set of objects as their instances.

E.g.,
```js
const pyFunc = pyodide.runPython(`lambda: print("zero")`);
pyFunc instanceof Function
```

### Checklist

- [x] Add / update tests


JeffersGlass:
Awesome! All the tests here and in that issue covered... so here's one more. 😇 

With this PR, adding a `__call__` attribute in JavaScript after class creation makes `instanceof Function` still false:

```js
const pyA = pyodide.runPython(`
    def A():
        ...
    A
`)
pyA.__call__ = () => console.log("I AM A")     // Order of these two lines makes no difference
const pyInstance = pyodide.runPython('A()')  // ^^^^
console.log("Instance of A is callable:", pyInstance instanceof Function) //false
```

But doing it in Pure Python works:
```py
const pyB = pyodide.runPython(`
    class B():
        ...

    B.__call__ = lambda self: print("B got called")
    B()
`)
console.log("Instance of B is callable:", pyB instanceof Function) // true
```

hoodmane:
Do you mean to say `class A:`?

hoodmane:
If I do:
```
pyA = pyodide.runPython(`
    class A:
        ...
    A
`)
pyA.__call__ = () => console.log("I AM A");
pyodide.runPython('A()()'); // Logs "I AM A"
pyInstance = pyodide.runPython('A()');
console.log("pyInstance instanceof Function:", pyInstance instanceof Function); // True
```
If you create `pyInstance` before setting `pyA.__call__` you'll get `false`, but this can't change for performance reasons.

JeffersGlass:
Oops yes, meant `class A()`, good catch.

Interesting, and good to know - is that because the subclass determination (pyCallable, pyAwaitable etc) is done at instance creation-time?

hoodmane:
> is that because the subclass determination is done at instance creation-time?

Yes exactly. This is true in both directions from JS to Python and from Python to JS. Using prototypes to store common functionality takes less memory and is much better for JIT if you you process a bunch of stuff of the same shape. You could imagine a system where you inspect the object again every time it is accessed but the performance of it is poor. A better approach would be to use some sort of cache invalidation like what the JavaScript engine itself uses if you modify the prototype of a class with existing instances. But that gets complicated pretty fast, and it's not clear that it's all that useful. Much easier not to bother.

hoodmane:
Strangely this works in Chrome but not in Firefox or Node.

ryanking13 (approved):
Thanks!

> Strangely this works in Chrome but not in Firefox or Node.

Hmm, interestingly it works in my Firefox browser v112 locally...

src/core/pyproxy.ts
```diff
@@ -2035,7 +2050,11 @@ const PyProxyHandlers = {
     // Note: must report "prototype" in proxy when we are callable.
     // (We can return the wrong value from "get" handler though.)
     let objHasKey = Reflect.has(jsobj, jskey);
-    if (objHasKey) {
+    if (
+      objHasKey &&
+      jskey in jsobj &&
```
WebReflection, line 2055: isn't this exactly what `Reflect.has(jsobj, jskey)` returns?

src/core/pyproxy.ts
```diff
@@ -2053,7 +2072,12 @@ const PyProxyHandlers = {
     // 2. the result of Python getattr
 
     // python_getattr will crash if given a Symbol.
-    if (jskey in jsobj || typeof jskey === "symbol") {
+    if (
+      typeof jskey === "symbol" ||
+      (jskey in jsobj &&
+        (!(jsobj instanceof Function) ||
+          (jskey !== "name" && jskey !== "length")))
```
WebReflection, line 2079: repeated: I suggest this check is factored out as utility such as `isKeyAcceptable` or something

src/core/pyproxy.ts
hoodmane: Oops good point.

src/tests/test_pyproxy.py
```diff
@@ -2096,3 +2096,78 @@ def test_pyproxy_of_list_fill(selenium, func):
...
+    if selenium.browser in ["firefox", "node"]:
+        # A hack to make the test work: In node and firefox this test fails. But
+        # I can't reproduce the failure in a normal browser / outside of the
+        # test suite. The trouble seems to be that the value of
+        # `globalThis.Function` changes its identity from when we define
+        # `PyProxyFunction` to when we execute this test. So we store `Function`
+        # on `pyodide._api.tests` so we can retrieve the original value of it
+        # for the test. This is nonsense but because the failure only occurs in
+        # the test suite and not in real life I guess it's okay????
+        # Also, no clue how node and firefox are affected but not Chrome.
```
hoodmane, line 2112: @WebReflection does this make any sense to you why the Function prototype would change? If I store:
```js
pyodide._module.tests.Function = Function
```
and then later compare:
```js
console.log(pyodide._module.tests.Function === Function);
```
this logs `false`. This causes the test to fail because `pyFunc` is an instanceof `pyodide._module.tests.Function` but not of `Function`. But the tests also only fail in Firefox and node and not in chrome...

WebReflection: is the prototype that changes or the global reference? the only thing I can think about in *NodeJS* is that the `.call` doesn't default to `globalThis` apparently, as mentioned in MDN, and maybe that runs before Firefox and it patches somehow the function globally to simulate a browser via node? Or maybe the test suite uses JSDOM and it changes somehow the global function? I don't have other clues but it's weird indeed.

PR #3927, opened 2023-06-12
Run replace_so_abi_tags on out of tree builds

hoodmane:
As requested by @henryiii.

### Checklist

- [x] Add / update tests

### DIFF
--- a/pyodide-build/pyodide_build/common.py
+++ b/pyodide-build/pyodide_build/common.py
@@ -576,3 +576,57 @@ def _get_sha256_checksum(archive: Path) -> str:
             if len(chunk) < CHUNK_SIZE:
                 break
     return h.hexdigest()
+
+
+def unpack_wheel(path: Path) -> None:
+    result = subprocess.run(
+        [sys.executable, "-m", "wheel", "unpack", path.name],
+        check=False,
+        encoding="utf-8",
+        cwd=path.parent,
+    )
+    if result.returncode != 0:
+        logger.error(f"ERROR: Unpacking wheel {path.name} failed")
+        exit_with_stdio(result)
+
+
+def pack_wheel(path: Path) -> None:
+    result = subprocess.run(
+        [sys.executable, "-m", "wheel", "pack", path.name],
+        check=False,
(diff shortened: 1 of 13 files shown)


henryiii:
This sounds great! I'll try to work on the other patch tomorrow afternoon, and take a look here if you want me to.

pyodide-build/pyodide_build/common.py
```diff
@@ -576,3 +576,57 @@ def _get_sha256_checksum(archive: Path) -> str:
...
+        encoding="utf-8",
+        cwd=path.parent,
+    )
+    if result.returncode != 0:
+        logger.error(f"ERROR: Packing wheel {path} failed")
+        exit_with_stdio(result)
+
+
+@contextmanager
+def modify_wheel(wheel: Path) -> Iterator[Path]:
```
ryanking13, line 606: Nice!

ryanking13 (approved):
Thanks, @hoodmane!

pyodide-build/pyodide_build/common.py
ryanking13: Could you add a docstring for this method?

rth:
> `pyodide build` now replaces native `.so` slugs with Emscripten
  slugs. Usually `.so`s in the generated wheels are actually Emscripten `.so`s
  so this is good. If they are actually native `.so`s then there is a problem
  either way.
  
  Not very critical, but should we actually check that the .so are emscripten .so rather than native .so that ended up there by mistake? For instance, we could check for the [WASM magic number](https://openhome.cc/eGossip/WebAssembly/Module.html) in the first 4 bytes maybe? It's supposed to be, `0061 736d`. Though I get the same bytes but in a different order, when I try,

```
$ hexdump -n 8 numpy/core/_multiarray_umath.cpython-311-wasm32-emscripten.so
0000000 6100 6d73 0001 0000 
```
but maybe I'm using hexdump wrong )

hoodmane:
Yeah that's a good idea. If they are actually native, maybe we should throw. That would be a very helpful addition.

PR #3929, opened 2023-06-14
Additional Arguments for `add_event_listener`/`remove_event_listener`

### LINKED ISSUE #3879: Support Additional Options for `add_event_listener` & `remove_event_listener`
## 🚀 Feature

The functions `pyodide.ffi.wrappers.add_event_listener` and `remove_event_listener` are useful wrappers around the browser's native functionality. However, the APIs for [addEventListener](https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/addEventListener) and [removeEventListener](https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/removeEventListener) include an option third parameter (`options` or `useCapture`), which adds some useful functionality. These third parameters would be useful to have supported in Pyodide.

### Motivation

This is coming from some PyScript developers, who find the `capture` option (to handle the event while it's bubbling down the DOM tree) and the `once` option (to make the callable only callable once) useful.

### Pitch

Modify the implementations of `add_event_listener` and `remove_event_listener` to support these additional options; possibly in a more Python way than having a third positional argument.

### Alternatives

These options are available if the use does the proxy management on their own, i.e. 
```python
from pyodide.ffi import create_proxy, to_js
from js import document, Object

elt = document.getElementById("some-id")
hello_func = create_proxy(lambda _: print("Hello"))
elt.addEventListener('click', hello_func, to_js({'capture': True}, dict_converter = Object.fromEntries))
```

JeffersGlass:
### Description

Adds an additional `*args` argument to `add_event_listener` and `remove_event_listener`; any additional values passed to these wrappers (that is, beyond `elt`, `event`, and `listener`) are passed through to the corresponding JavaScript function.

Because modern browsers seem to mostly expect [an options object as the third parameter](https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/addEventListener#parameters), any `dict` arguments passed to `*args` are converted to a JS object using `to_js(..., dict_converter=Object.fromEntries)`. If this feels like doing too much for the user, happy to remove this part.

This is my first time interacting with the docs and test system in a significant way; I hope I have done it right, and happy to make adjustments in substance or style wherever is needed.

Resolves #3879 

### Checklists

- [x] Add / update tests

### DIFF
--- a/src/tests/test_browser_apis.py
+++ b/src/tests/test_browser_apis.py
@@ -271,8 +271,8 @@ async def test_additional_event_listener_arguments(selenium):
 
     add_event_listener(x, "click", foo, once=True, something_else="foo")
 
-    assert x.listeners.click[0].kwargs.once is True
-    assert x.listeners.click[0].kwargs.something_else == "foo"
+    assert x.listeners.click[0].kwargs["once"] is True
+    assert x.listeners.click[0].kwargs["something_else"] == "foo"
 
     remove_event_listener(x, "click", foo)
     assert len(x.listeners.click) == 0


JeffersGlass:
I see I have some mypy errors, will work to resolve those.

src/py/pyodide/ffi/wrappers.py
```diff
@@ -31,26 +38,67 @@ def destroy():
 
 
 def add_event_listener(
-    elt: JsDomElement, event: str, listener: Callable[[Any], None]
+    elt: JsDomElement, event: str, listener: Callable[[Any], None], *args: any
```
hoodmane, line 41: I think passing kwargs makes more sense here:
```suggestion
    elt: JsDomElement, event: str, listener: Callable[[Any], None], **kwargs: Any
```

src/py/pyodide/ffi/wrappers.py
```diff
@@ -31,26 +38,67 @@ def destroy():
...
+        :py:meth:`to_js(..., dict_converter=Object.fromEntries) <pyodide.ffi.to_js>`.
     """
     proxy = create_proxy(listener)
     EVENT_LISTENERS[(elt.js_id, event, listener)] = proxy
-    elt.addEventListener(event, cast(Callable[[Any], None], proxy))
+    converted_args = (
+        to_js(arg, dict_converter=Object.fromEntries) if type(arg) == dict else arg
+        for arg in args
+    )  # dicts are converted to objects
+    elt.addEventListener(event, cast(Callable[[Any], None], proxy), *converted_args)
```
hoodmane, line 70:
```suggestion
    elt.addEventListener(event, cast(Callable[[Any], None], proxy), **kwargs)
```

src/py/pyodide/ffi/wrappers.py
```diff
@@ -31,26 +38,67 @@ def destroy():
...
 
 
 def remove_event_listener(
-    elt: JsDomElement, event: str, listener: Callable[[Any], None]
+    elt: JsDomElement, event: str, listener: Callable[[Any], None], *args: any
```
hoodmane, line 74:
```suggestion
    elt: JsDomElement, event: str, listener: Callable[[Any], None], **kwargs: Any
```

src/py/pyodide/ffi/wrappers.py
```diff
@@ -31,26 +38,67 @@ def destroy():
...
+        convenience, ``dict`` arguments are converted to JavaScript Objects using
+        :py:meth:`to_js(..., dict_converter=Object.fromEntries) <pyodide.ffi.to_js>`.
     """
     proxy = EVENT_LISTENERS.pop((elt.js_id, event, listener))
-    elt.removeEventListener(event, cast(Callable[[Any], None], proxy))
+    converted_args = (
+        to_js(arg, dict_converter=Object.fromEntries) if type(arg) == dict else arg
+        for arg in args
+    )  # dicts are converted to objects
+    elt.removeEventListener(event, cast(Callable[[Any], None], proxy), converted_args)
```
hoodmane, line 101:
```suggestion
    elt.removeEventListener(event, cast(Callable[[Any], None], proxy), **kwargs)
```

src/tests/test_browser_apis.py
```diff
@@ -224,6 +224,63 @@ def foo(obj):
...
+async def test_additional_event_listener_arguments(selenium):
+    from pyodide.code import run_js
+
+    x = run_js(
+        """
+class MockObject {
+    constructor() {
+        this.listeners = {};
+    }
+    addEventListener(event, handler, arg1, arg2) {
```
hoodmane, line 237: Why does the mock have a nonstandard signature?

hoodmane (commented):
Thanks @JeffersGlass!

hoodmane:
In `_core_docs`, both  `addEventListener` and `removeEventListener` should be modified to take `**kwargs` too.

src/tests/test_browser_apis.py
JeffersGlass: Adjusted to just use `kwargs`. Thanks!

JeffersGlass:
@hoodmane thank you for the review and the pointers! I've updated both functions to us `**kwargs`, plus the documentation and `_core_docs` for the same.

src/py/_pyodide/_core_docs.py
```diff
@@ -1027,10 +1027,14 @@ def children(self) -> Sequence["JsDomElement"]:
     def appendChild(self, child: "JsDomElement") -> None:
         pass
 
-    def addEventListener(self, event: str, listener: Callable[[Any], None]) -> None:
+    def addEventListener(
+        self, event: str, listener: Callable[[Any], None], kwargs: Any
```
hoodmane, line 1031:
```suggestion
        self, event: str, listener: Callable[[Any], None], **kwargs: Any
```

src/py/_pyodide/_core_docs.py
```diff
@@ -1027,10 +1027,14 @@ def children(self) -> Sequence["JsDomElement"]:
...
+    ) -> None:
         pass
 
-    def removeEventListener(self, event: str, listener: Callable[[Any], None]) -> None:
+    def removeEventListener(
+        self, event: str, listener: Callable[[Any], None], kwargs: Any
```
hoodmane, line 1036:
```suggestion
        self, event: str, listener: Callable[[Any], None], **kwargs: Any
```

src/tests/test_browser_apis.py
```diff
@@ -224,6 +224,60 @@ def foo(obj):
     remove_event_listener(x, "click", foo)
 
 
+@run_in_pyodide
+async def test_additional_event_listener_arguments(selenium):
+    from pyodide.code import run_js
+
+    x = run_js(
+        """
+class MockObject {
```
hoodmane, line 233: Indent this so that the body and the closing `"""` match the opening `"""`.

src/py/_pyodide/_core_docs.py
JeffersGlass: I think this does actually want to be `kwargs` (or something like `options`) if we want to pass along a single object with all the kwargs passed as attributes, yes?

src/py/_pyodide/_core_docs.py
JeffersGlass: I think this does actually want to be `kwargs` (or something like `options`) if we want to pass along a single object with all the kwargs passed as attributes, yes? (Duplicate of above)

src/py/_pyodide/_core_docs.py
JeffersGlass: Actually that might not be what we want... apparently there are cases where you'd want to pass multiple separate arguments which might be objects or booleans (i.e. `options` and `wantsUntrusted`), which makes me thing we should use `*args`?

hoodmane: It's the same to say `{capture: true}` as to say `true` there, so I don't see the reason to use the unnamed version, it's strictly less readable. If you want to be 100% accurate I think you'd need something like the following:
```
@override
def addEventListener(
    self, event, handler: Callable[[Event], None], capture: bool = False, /
):
    ...


@override
def addEventListener(
    self,
    event: str,
    handler: Callable[[Event], None],
    /,
    *,
    capture: bool = False,
    once: bool = False,
    passive: bool = False,
    signal: JsProxy | None
):
    ...

def addEventListener(self, event, handler, /, *args, **kwargs):
    raise NotImplementedError
```

src/tests/test_browser_apis.py
```diff
@@ -224,6 +224,60 @@ def foo(obj):
...
+
+    from pyodide.ffi.wrappers import add_event_listener, remove_event_listener
+
+    def foo(obj):
+        pass
+
+    add_event_listener(x, "click", foo, once=True, something_else="foo")
+
+    assert x.listeners.click[0].kwargs.once is True
+    assert x.listeners.click[0].kwargs.something_else == "foo"
```
hoodmane, line 275: Probably needs to be:
```suggestion
    assert x.listeners.click[0].kwargs["once"] is True
    assert x.listeners.click[0].kwargs["something_else"] == "foo"
```

hoodmane:
Hmm so it was probably right originally. But the test was failing for some reason @JeffersGlass ...

PR #3934, opened 2023-06-17
MAINT Relocate build environment management related methods

ryanking13:
Relocates methods that are related to the build environment management (common.py ==> build_env.py (open to other names) ).

This is a final split of #3737. Close #3737


rth:
> codecov/project — 64.75% (-1.19%) compared to 5fe0d2e 

That's a bit suspicious for a refactoring moving functions around. Could you double check locally that the coverage is comparable (could be just some codecov issue)? Otherwise thanks for working on this! (Also feel free to ignore this comment).

ryanking13:
I'll check the coverage again before merging and add more tests if possible. Thanks for pointing that out!

ryanking13:
> Could you double check locally that the coverage is comparable

I changed a test `test_build1` in test_cli.py to be run after the Pyodide build, as I had some trouble running it locally without building Pyodide. So this test is now not covered in `test_python` but in `test_core`. I think this was the reason for the drop in coverage.

I'll merge this, but in the future, we should find a way to run tests that use xbuildenv without building Pyodide, probably by mocking some files that consist xbuildenv.

PR #3935, opened 2023-06-17
Add types for FS of PyodideAPI

jon-edward:
### Description

This adds types for `PyodideAPI.FS`

### Checklists

### DIFF
--- a/src/js/api.ts
+++ b/src/js/api.ts
@@ -14,6 +14,8 @@ import "./error_handling.gen.js";
 import { setStdin, setStdout, setStderr } from "./streams";
 import { makeWarnOnce } from "./util";
 
+import "emscripten";
+
 API.loadBinaryFile = loadBinaryFile;
 
 /**
@@ -112,7 +114,7 @@ export class PyodideAPI {
    * are available as members of ``FS.filesystems``:
    * ``IDBFS``, ``NODEFS``, ``PROXYFS``, ``WORKERFS``.
    */
-  static FS = {} as any;
+  static FS = {} as typeof FS;
   /**
    * An alias to the `Emscripten Path API
    * <https://github.com/emscripten-core/emscripten/blob/main/src/library_path.js>`_.
--- a/Makefile
+++ b/Makefile
@@ -135,7 +135,7 @@ npm-link: dist/package.json
 	cd src/test-js && npm ci && npm link ../../dist
 
 dist/pyodide.d.ts dist/pyodide/ffi.d.ts: src/js/*.ts src/js/pyproxy.gen.ts src/js/error_handling.gen.ts
-	npx dts-bundle-generator src/js/{pyodide,ffi}.ts --export-referenced-types false --project src/js/tsconfig.json
+	npx dts-bundle-generator src/js/{pyodide,ffi}.ts --export-referenced-types false --project src/js/tsconfig.json --external-types emscripten
 	mv src/js/{pyodide,ffi}.d.ts dist
 	python3 tools/fixup-type-definitions.py dist/pyodide.d.ts
 	python3 tools/fixup-type-definitions.py dist/ffi.d.ts
--- a/docs/project/changelog.md
+++ b/docs/project/changelog.md
@@ -1657,6 +1657,7 @@ _August 3rd, 2021_
   converting dictionaries. For instance, pass `Object.fromEntries` to convert to
   an object or `Array.from` to convert to an array of pairs.
   {pr}`1742`
+- {{ API }} `PyodideAPI.FS` now contains type information from `@types/emscripten`. 
 
 ### pyodide-build
 
--- a/src/js/esbuild.config.mjs
+++ b/src/js/esbuild.config.mjs
@@ -70,7 +70,7 @@ try {
     // simplify the umd dance for CommonJS by trying to set info on `exports`
     content = content.replace(
       "//# sourceMappingURL",
-      `try{Object.assign(exports,${name})}catch(_){}\nglobalThis.${name}=${name}.${name};\n//# sourceMappingURL`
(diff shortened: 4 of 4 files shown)


src/js/api.ts
```diff
@@ -14,6 +14,8 @@ import "./error_handling.gen.js";
 import { setStdin, setStdout, setStderr } from "./streams";
 import { makeWarnOnce } from "./util";
 
+import "emscripten";
```
hoodmane, line 17: where does the "emscripten" import come from? It presumably needs to be in `package.json`. I see:
https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/emscripten/index.d.ts
but it says they are type definitions for Emscripten 1.39.16 which is way out of date. I guess we could try to bring them up to date if there are in fact any discrepancies. I don't know that the FS API changes much.

hoodmane (commented):
Thanks for working on this @jon-edward!

src/js/api.ts
jon-edward: `@types/emscripten` was already required in [`package.json`](https://github.com/pyodide/pyodide/blob/1691d347d15a2c211cd49aebe6f15d42dfdf2369/src/js/package.json#L20) at the most current major version, so that didn't need to be changed. More current type declarations are definitely needed, admittedly

hoodmane: I see, it was introduced in #1692. @bollwyvl does that pr actually use `@types/emscripten`?

Anyways it would be helpful if we could add a bunch of ts tests that use the different functionality of `FS` so that we have an assertion that our type defs reasonably match the runtime behavior. Maybe we could work on that in a followup. what do you think @ryanking13?

Types could definitely use more focused attention but we seem to have trouble ever getting around to working on them with all the other maintenance work around here...

jon-edward: That's understandable. Specifically with the `FS` library I found myself constantly going back to the Emscripten API reference so I thought adding type declarations (outdated or not) would at least give me a starting place to work off of.

hoodmane: I do think shipping slightly incorrect type declarations is a good way to get bug reports and help gradually fix the problems.

jon-edward:
On a few tests ([here](https://github.com/pyodide/pyodide/actions/runs/5299405057/jobs/9592572016?pr=3935) for example) it errors on `Error: Dynamic require of "emscripten" is not supported`.

I believe this is a result of the `esbuild` step, which is addressed [here](https://github.com/evanw/esbuild/issues/1921). Can anyone give direction on ignoring `require`s? I'm thinking it might require usage of the [`esbuild-plugin-ignore`](https://www.npmjs.com/package/esbuild-plugin-ignore).

hoodmane:
I think you need to remove `import "emscripten"` and place 
```ts
/// <reference types="emscripten" />
```
at the top of the file.

jon-edward:
Seems like everything's working with the triple-slash directive. Should give us a simple base to build tests off of and improve the preexisting type declarations for `emscripten`.

hoodmane (approved):
Looks good to me. Let's wait and see what @ryanking13 thinks before merging.

src/js/api.ts
bollwyvl: Yep: wrong types are often _worse_ than no types, but _maybe_ semi-offically  (or at least _de facto_) wrong types are at least useful. There are a lot of `: any`s down in things at various levels. But having _sort of_ the right shape is a lot more powerful than _just_ any, which is why I opted to bring in those, old as they are. The gross ambient types are not my preference, hence the weird jsdoc construction around [`Module`](https://github.com/pyodide/pyodide/pull/1692/files#diff-29e0ed0a380c63720014afb4aa3b272caddb5e1913c28d0ace55149e2758ba16R4). Something similar can be achieved with _actual_ typescript via, e.g. `typeof Module.FS`.

On the whole, though, this PR seems good, and would be _ready_ to accept improved upstream types, if they existed.

jon-edward: I agree that ambient types without a direct import aren't ideal, I’ve been struggling with finding a way to import a specific name from a `dts` file like what's provided in `emscripten` but I think that's a current limitation of TS which is why I landed on `import "emscripten"` however that brings about issues with building. 

In the upcoming days I might try to sort through emscripten's current `FS` library (just at a quick glance I feel it's somewhere [here](https://github.com/emscripten-core/emscripten/blob/main/src/library_fs.js)) and determine if we even have an issue with significantly outdated types against what I can find in the old `@types/emscripten`

jon-edward: I didn't realize pyodide already had an interface for `FS`, why isn't that exposed to `PyodideAPI.FS`?

ryanking13:
Thanks! Using `@types/emscripten` would be really nice, but I think we need to add some tests before we merge this (or at least we should make some test templates so we can add more tests later).

I think updating `@types/emscripten` was discussed in Jupyterlite Paris workshop last year, not sure anybody is working on it though. Obviously, it is not very exciting work to do 😂.

hoodmane:
So I'm a bit confused about the status here. Maybe it would be better to expose our own somewhat tested types rather than exposing types from Emscripten 1.39? We could add docstrings later...

ryanking13:
> So I'm a bit confused about the status here. Maybe it would be better to expose our own somewhat tested types rather than exposing types from Emscripten 1.39? We could add docstrings later...

I think we need to 

1. Write some tests for FS type in @types/emscripten to see if it is compatible with the Emscripten version that we use.
2. If it is compatible, then we replace our custom FS type with `@types/emscripten`.
3. If it is not compatible, then probably we need to use our custom FS type and maybe upstream it to `@types/emscripten` later...

ryanking13:
Superceded by #5910

PR #3949, opened 2023-06-23
Use pyodide-lock for pyodide-lock.json parsing in Python

rth:
Uses https://github.com/pyodide/pyodide-lock for the Pydantic model of `pyodide-lock.json`. Nothing is yet fixed API wise, so if you think we should change Pydantic class names or method names defined there, let me know.

I guess in time if we generate a JSON schema for it (https://github.com/pyodide/pyodide-lock/issues/3),  we could also validate the parsing on the JS side. This might be useful particularly if users start creating custom lock files.


ryanking13 (approved):
Very nice, thanks!

> This might be useful particularly if users start creating custom lock files.

Maybe we need to start defining which fields are mandatory and which fields are optional. For instance, `package_type` and `unvendored_tests` fields highly depend on specific behavior of pyodide-build so it needs to be optional I think. (Unrelated to this PR)

rth:
> For instance, package_type and unvendored_tests fields highly depend on specific behavior of pyodide-build so it needs to be optional I think. 

I added default values for all non critical fields in https://github.com/pyodide/pyodide-lock/blob/main/pyodide_lock/spec.py#L26. If we make them optional, we would need to set all of the defaults to None, and then check everywhere in the code that it's not None before using them. Which is probably more work for the same result in the end.

PR #3953, opened 2023-06-23
Replace void return value with int in quadpack.h

hoodmane:
This is needed to build with wasm-exceptions or with the compatibility workaround for JSPI + JS exceptions.
From #3210.


rth (approved):

hoodmane:
To describe a bit more what's going on, in C++
```Cpp
try {
	some_func(arg1, arg2, ...);
} catch (SomeException e) {
	// ...
}
```
generates code like:
```
invoke_sig(some_func_ptr, arg1, arg2, ...);
```
where `invoke_sig` is a JS trampoline that does a JS try/catch. So in particular, if `some_func` is declared with the wrong type but it is only called in try blocks, then it will work anyways. But if instead of js exceptions you use wasm exceptions then it's back to "null function or function signature mismatch".

hoodmane:
@lesteve if you get a chance could you run the scipy test suite against #3210 and see if there are any other new function signature mismatches?

lesteve:
I'll try to do that and let you know!

PR #3954, opened 2023-06-23
move trampolines from cpython source into Pyodide source

hoodmane:
This is split off from #3210. 

We need to change the way the Python trampolines work for #3210. This does the first step of moving the definitions into our own code.

#3955 is on top of this and it changes the definition to the one we need for jspi.


hoodmane:
@ryanking13 @rth I think this one should be reviewed/merged first. Hopefully it isn't too complicated, it just moves some logic from the cpython source into Pyodide.

cpython/patches/0007-gh-93839-Use-load_package_tests-for-testmock-GH-9405.patch
ryanking13, line 1: Note to myself: this patch is applied to [unvendored test module](https://github.com/pyodide/pyodide/tree/main/packages/test/patches
).

src/core/trampoline.c
```diff
@@ -0,0 +1,57 @@
+#include <Python.h>
+#include <emscripten.h>
+
+// clang-format off
+EM_JS(
+PyObject*,
+py_emjs_trampoline_js,
+(PyCFunctionWithKeywords func, PyObject* self, PyObject* args, PyObject* kw),
+{
+    return wasmTableMirror[func](self, args, kw);
```
ryanking13, line 10: Can't we use [`getWasmTableEntry`](https://github.com/emscripten-core/emscripten/blob/818523af5aacd41b8154906832a3c2b26e43a1b4/src/library.js#L3272-L3277) here? It also contains a fallback to wasmTable when wasmTableMirror is not available.

cpython/patches/0007-Remove-trampoline-definitions.patch
```diff
@@ -0,0 +1,88 @@
+From dfb8c93d4e9754e56ce805088e33fed2668c2834 Mon Sep 17 00:00:00 2001
+From: Hood Chatham <roberthoodchatham@gmail.com>
+Date: Mon, 19 Jun 2023 12:32:05 -0700
+Subject: [PATCH 7/7] Remove trampoline definitions
+
+We need to make our own JSPI aware versions
```
ryanking13, line 6: Could you please add a comment on where we locate our own version of these methods, for those who see this file later?

src/core/trampoline.c
```diff
@@ -0,0 +1,57 @@
+#include <Python.h>
```
ryanking13, line 1: A top-level comment of what this file does would be nice.

ryanking13 (approved):
Thanks, @hoodmane, and sorry for the late review. I have a few comments, otherwise LGTM.

hoodmane:
Closed in favor of #3964

PR #3955, opened 2023-06-24
Use a JS Type Reflection based trampoline if possible

hoodmane:
This is split off from #3210 and on top of #3954.

When type reflection is available, this uses an alternate trampoline that
inspects how many arguments the function actually needs and then chooses the
appropriate call with a switch block. This is necessary to use JSPI because
it does not interact well with trampolines.

This code will be used on Chrome and Safari but not Firefox. As far as I can tell it's
performance neutral.


hoodmane:
Closed in favor of https://github.com/pyodide/pyodide/pull/3964

PR #3956, opened 2023-06-24
Fix failing hypothesis tests on the main branch

hoodmane:
Hopefully


packages/hypothesis/meta.yaml
ryanking13, line 1: Dependencies from METADATA:

```
Requires-Dist: attrs (>=19.2.0)
Requires-Dist: sortedcontainers (<3.0.0,>=2.1.0)
Requires-Dist: exceptiongroup (>=1.0.0) ; python_version < "3.11"
```

packages/hypothesis/meta.yaml
```diff
@@ -0,0 +1,17 @@
+package:
+  name: hypothesis
+  version: 6.79.2
+  tag:
+    - core
+  top-level:
+    - _hypothesis_ftz_detector
+    - _hypothesis_pytestplugin
```
ryanking13, line 8: I know that these are auto-generated from mkpkg, but I think we can remove these from top-level since users will not import these modules directly.

ryanking13 (approved):
Thanks!

rth:
Thanks a lot for fixing it!

PR #3957, opened 2023-06-24
Use Wasm Exceptions for invoke stubs if they exist

hoodmane:
This is another piece split off from #3210. If wasm exception handling is available, use dynamically generated modules with wasm eh for the invoke stubs instead of a JS trampoline. This is useful because JS trampolines interact poorly with JSPI.

- [x] explanation at top of `src/core/continuations.js`
- [x] docstrings
- [x] unit tests


ryanking13:
What's the status of this PR? It doesn't seem `continuation.js` is bundled to pyodide.asm.js in this PR. Is it intended?

hoodmane:
>  Is it intended?

Uhh... no I definitely meant to include it.

hoodmane:
I ran the scipy test suite against this PR (commit 7017ae1c0eacc515a9e31d4c6e26a4f850e403f4) and there are some new (compared to `main`) failures as hinted in https://github.com/pyodide/pyodide/pull/3953#issuecomment-1605050751:

```
scipy.linalg.tests result expected in ['passed'], got 'fatal error or timeout' instead
scipy.sparse.linalg._dsolve.tests result expected in ['passed'], got 'fatal error or timeout' instead
scipy.sparse.linalg.tests result expected in ['passed'], got 'fatal error or timeout' instead
scipy.spatial.tests result expected in ['passed'], got 'fatal error or timeout' instead
scipy.stats.tests result expected in ['failed'], got 'fatal error or timeout' instead
```

All the fatal errors seems related to `dgees` with a traceback like this:
```
RuntimeError: null function or function signature mismatch
    at wasm://wasm/58c477ee:wasm-function[3]:0xad
    at stubs.<computed> (/home/lesteve/dev/scipy-tests-pyodide/node_modules/pyodide/pyodide.asm.js:2604:33)
    at f2py_rout__flapack_dgees (wasm://wasm/005feb3a:wasm-function[234]:0x3bc97)
    at fortran_call (wasm://wasm/0003e50e:wasm-function[100]:0x659b)
    at py_emjs_trampoline_js (/home/lesteve/dev/scipy-tests-pyodide/node_modules/pyodide/pyodide.asm.js:1703:81)
    at _PyCFunctionWithKeywords_TrampolineCall (wasm://wasm/0258e5f6:wasm-function[483]:0x15bfaf)
    at _PyObject_MakeTpCall (wasm://wasm/0258e5f6:wasm-function[1063]:0x19d889)
    at PyObject_Vectorcall (wasm://wasm/0258e5f6:wasm-function[1071]:0x19de7a)
    at _PyEval_EvalFrameDefault (wasm://wasm/0258e5f6:wasm-function[3206]:0x264a5a)
    at _PyEval_Vector (wasm://wasm/0258e5f6:wasm-function[3207]:0x268c27) {
  pyodide_fatal_error: true
}
```

There are also `memory access out of bounds` errors in  `scipy.sparse.linalg._dsolve.tests` e.g. `test_linsolve.py::TestLinsolve::test_singular_gh_3312`
```
    at __get_exception_message (wasm://wasm/0258e5f6:wasm-function[12567]:0x4c8d42)
    at Module.___get_exception_message (/home/lesteve/dev/scipy-tests-pyodide/node_modules/pyodide/pyodide.asm.js:65827:117)
    at /home/lesteve/dev/scipy-tests-pyodide/node_modules/pyodide/pyodide.asm.js:2087:9
    at withStackSave (/home/lesteve/dev/scipy-tests-pyodide/node_modules/pyodide/pyodide.asm.js:1769:17)
    at getExceptionMessageCommon (/home/lesteve/dev/scipy-tests-pyodide/node_modules/pyodide/pyodide.asm.js:2084:14)
    at Object.getExceptionMessage (/home/lesteve/dev/scipy-tests-pyodide/node_modules/pyodide/pyodide.asm.js:2102:14)
    at convertCppException (/home/lesteve/dev/scipy-tests-pyodide/node_modules/pyodide/pyodide.asm.js:1135:597)
    at API.fatal_error (/home/lesteve/dev/scipy-tests-pyodide/node_modules/pyodide/pyodide.asm.js:1135:921)
    at Module.callPyObjectKwargs (/home/lesteve/dev/scipy-tests-pyodide/node_modules/pyodide/pyodide.asm.js:1131:705)
    at Module.callPyObject (/home/lesteve/dev/scipy-tests-pyodide/node_modules/pyodide/pyodide.asm.js:1131:1626)
```

_Originally posted by @lesteve in https://github.com/pyodide/pyodide/issues/3210#issuecomment-1607305515_

hoodmane:
The `test_singular_gh_3312` error was a bug that made `longjmp` not work at all, which I think I've fixed. I hope that the gees fatal error is fixed by https://github.com/pyodide/pyodide/pull/3957/files#diff-a7ab2b529bc659537dd832d0b0ec6c3d98193116aea7d1327cf677c187137f1f

hoodmane:
I can confirm that scipy-tests-pyodide now says:
```
Test results matched expected ones
```
Thanks again @lesteve for making scipy-tests-pyodide it's super helpful!

lesteve:
> Thanks again @lesteve for making scipy-tests-pyodide it's super helpful!

Great to hear that :heart:!

hoodmane:
@rth @ryanking13 would appreciate a review on this one if either of you has a chance.

rth (commented):
Thanks for working on this topic! It certainly looks challenging.

Could you add,
 -  a paragraph of comments in `src/core/continuations.js`describing broadly what this file does
 -  docstrings for at least more complex functions (e.g createInvokeModule)
 - because it's pretty low level and wasm binary opcode manipulations, maybe it would be good to write some unit tests for it. I don't know how we test js files in src/core, but I would feel more confident if we had some tests there. Otherwise the current situation is that we know that it works because it doesn't break the build and the overall test suite, but if it fails due to this one day it would hard to manage if there is no way to add non regression tests.
 - from what I understand we have to do this dynamically because WASM exception handling is still fairly recent. In the medium term, once it's been available for all browsers for a while, wouldn't it be better to do this statically as a post-processing step? Do you have any feeling if it has a measurable runtime cost currently?
 - a changelog entry

src/core/continuations.js
```diff
@@ -0,0 +1,328 @@
...
+
+function insertSectionPrefix(sectionCode, sectionBody) {
+  var section = [sectionCode];
+  uleb128Encode(sectionBody.length, section); // length of section in bytes
+  section.push(...sectionBody);
+  return section;
+}
+
+const typeCodes = {
+  i32: 0x7f,
```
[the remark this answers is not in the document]
hoodmane: oh wow I didn't know about that index that's so useful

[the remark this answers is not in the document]
hoodmane: Though for typecodes the right place is its neighbor:
https://webassembly.github.io/spec/core/appendix/index-types.html

hoodmane:
> from what I understand we have to do this dynamically because WASM exception handling is still fairly recent. In the medium term, once it's been available for all browsers for a while, wouldn't it be better to do this statically as a post-processing step? Do you have any feeling if it has a measurable runtime cost currently?

I don't think the runtime cost is very high. My feeling is that it must be comparable to or better than the cost of the JS invoke. I do think doing it statically would be a great idea and allow a fair amount of simplification. In fact, the post-process step would be a great way to deal with Rust's wasm-exceptions problems: I think we can have it produce the code for the js exceptions ABI and switch it as a postprocessing step.

src/js/test/unit/invokes/invoke_dd.wat
hoodmane, line 1: @sbc100 would it make sense to generate code like this if someone uses `-fexceptions -sASYNCIFY=2`? Or is the plan to require `-fwasm-exceptions` with `-sASYNCIFY=2`?

sbc100: `-fwasm-exceptions` is already shipping in chrome (and maybe firefox)?  So its pretty close the being ready to enable by default.  `-sASYNCIFY=2` is quite a long way from shipping and is not even complete in v8 at this point.   

So if I understand your question correctly, then no, I don't think it would be worth investing in solutions that rely on `-sASYNCIFY=2` but somehow don't have wasm exceptions available.  Based on current data, I don't imagine such a target will exist.

ryanking13 (commented):
Thanks! This is really a nice improvement.

This generally looks okay to me, but I think the JS tests in src/js are not running in CI right now, so we would probably need to enable js tests in CI.

src/js/test/unit/invokes/invoke_dd.wat
hoodmane: Is anyone working on fixing the Rust compiler so that it can emit code that works with `-fwasm-exceptions`? Because this is where I am stuck on using it.

hoodmane: I guess I've already opened a few issues about that, can probably continue the discussion there. Thanks @sbc100 that makes sense.

ryanking13 (approved):
Thanks @hoodmane!

hoodmane:
I'm going to leave the changelog entry for when we get user-facing features. This PR doesn't have any obvious effect on the user.

PR #3960, opened 2023-06-27
Scipy: Fix gees calls from wasm invokes

hoodmane:
This is the scipy patch from #3957. It fixes more function pointer mismatches which are masked by setjmp/longjmp.

### DIFF
--- a/packages/scipy/meta.yaml
+++ b/packages/scipy/meta.yaml
@@ -34,7 +34,7 @@ source:
     # been fixed upstream in https://github.com/scipy/scipy/pull/18124
     - patches/0010-When-forward-declaring-print_soln-give-it-the-correc.patch
     - patches/0011-Enable-long-double-support-in-Boost-library-for-emsc.patch
-    - patches/0001-Fix-gees-calls.patch
+    - patches/0012-Fix-gees-calls.patch
 
 build:
   cflags: |
--- a/packages/scipy/meta.yaml
+++ b/packages/scipy/meta.yaml
@@ -34,6 +34,7 @@ source:
     # been fixed upstream in https://github.com/scipy/scipy/pull/18124
     - patches/0010-When-forward-declaring-print_soln-give-it-the-correc.patch
     - patches/0011-Enable-long-double-support-in-Boost-library-for-emsc.patch
+    - patches/0001-Fix-gees-calls.patch
 
 build:
   cflags: |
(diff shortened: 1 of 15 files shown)


packages/scipy/meta.yaml
```diff
@@ -34,6 +34,7 @@ source:
     # been fixed upstream in https://github.com/scipy/scipy/pull/18124
     - patches/0010-When-forward-declaring-print_soln-give-it-the-correc.patch
     - patches/0011-Enable-long-double-support-in-Boost-library-for-emsc.patch
+    - patches/0001-Fix-gees-calls.patch
```
lesteve, line 37: You may want to use `0012-Fix-gees-calls.patch` to be consistent with the other patch filenames?

lesteve (approved):
This seems the right fix if it works. When I quickly looked at this issue, I noticed indeed that dgees was called with 2 supplementary arguments (called with 17 parameters instead of 15).

lesteve:
BTW since you asked a while ago, I recently realised that it is straightforward to debug a node script with Chromium:

In a terminal launch `node inspect`:
```
node inspect my-script.js
```

In Chromium open:
```
chrome://inspect
```

You should see something listed in the "Remote target":

I find it convenient to click on "Open dedicated DevTools for Node" that opens a separate window and connect automatically to the right remote target:
image

You can type `c` to continue and `r` to restart in the terminal where you launch `node inspect`.

hoodmane:
By the way I tried to fix the `dotu` calls by applying this patch:
```patch
-  callstatement (*f2py_func)(&<prefix2c>dotu,&n,x+offx,&incx,y+offy,&incy)
-  callprotoargument <ctype2c>*,F_INT*,<ctype2c>*,F_INT*,<ctype2c>*,F_INT*
+  callstatement (*f2py_func)(&<prefix2c>dotu,&n,x,&offx,&incx,y,&offy,&incy)
+  callprotoargument <ctype2c>*,F_INT*,<ctype2c>*,F_INT*,F_INT*,<ctype2c>*,F_INT*,F_INT*
```
but it creates weird load errors complaining about a missing symbol `wcdotu_`. I don't know why this patch would create a missing symbol error though.

ryanking13 (approved):
Thanks! The scipy build is failing in CI... I reran it, let's see if it was CI issue.

hoodmane:
I think it might be a typo in meta.yaml

lesteve:
> but it creates weird load errors complaining about a missing symbol wcdotu_. I don't know why this patch would create a missing symbol error though.

Not sure either. Full disclosure the scipy issues with complex numbers are not very high in my priority list, of course if someone manages to fix it, I would be more than happy!

hoodmane:
Yeah not very high in my priority list either just thought I'd give it a try since I was looking at scipy anyways.

PR #3963, opened 2023-06-27
Only check if key in obj if obj[key] returns undefined

hoodmane:
This is a minor optimization in `JsObject_GetString`. We only need to check if `key in obj` when `obj[key]` returns `undefined` to distinguish between missing and present but set to `undefined`.

@WebReflection

- [x] Add test

### DIFF
--- a/src/core/hiwire.c
+++ b/src/core/hiwire.c
@@ -966,12 +966,15 @@ EM_JS(int, normalizeReservedWords, (int word), {
 });
 
 EM_JS_REF(JsRef, JsObject_GetString, (JsRef idobj, const char* ptrkey), {
-  let jsobj = Hiwire.get_value(idobj);
-  let jskey = normalizeReservedWords(UTF8ToString(ptrkey));
-  if (jskey in jsobj) {
-    return Hiwire.new_value(jsobj[jskey]);
+  const jsobj = Hiwire.get_value(idobj);
+  const jskey = normalizeReservedWords(UTF8ToString(ptrkey));
+  const result = jsobj[jskey];
+  // clang-format off
+  if (result === undefined && !(jskey in jsobj)) {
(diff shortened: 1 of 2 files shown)


src/core/hiwire.c
```diff
@@ -966,12 +966,15 @@ EM_JS(int, normalizeReservedWords, (int word), {
...
-  let jsobj = Hiwire.get_value(idobj);
-  let jskey = normalizeReservedWords(UTF8ToString(ptrkey));
-  if (jskey in jsobj) {
-    return Hiwire.new_value(jsobj[jskey]);
+  const jsobj = Hiwire.get_value(idobj);
+  const jskey = normalizeReservedWords(UTF8ToString(ptrkey));
+  const result = jsobj[jskey];
+  // clang-format off
+  if (result === undefined && !(jskey in jsobj)) {
+    // clang-format on
```
antocuni, line 974: just curious, why do you need to turn off `clang-format` here?

WebReflection:
If anyone is interested more about this issue, *Proxies* in JS don't always expose a `has` trap, specially when **lazy accessors** are meant to be created on demand at runtime, where `has` means nothing there but also it cannot always return `true` because some stuff might not be there or work once accessed directly.

In our specific exchange, I am using cross realms proxies orchestration via Atomics and SharedArrayBuffer so that in a *Worker* there's nothing `has` can answer to unless it does a whole roundtrip to the *Main* thread, and vice-versa, and such roundtrip is done either on *get* or *set* trap (later on within the *set* trap) so that `has` is hard to reason about and non-nonsensical in practice.

src/core/hiwire.c
hoodmane: clang-format converts `===` ==> `== =` which is a syntax error.

PR #3964, opened 2023-06-28
Patch CPython to use a type reflection trampoline if possible

hoodmane:
This is a replacement for #3954 and #3955 structured as an upstream patch.

CF upstream PR:
https://github.com/python/cpython/pull/106219


cpython/patches/0008-Make-Emscripten-trampolines-work-with-JSPI.patch
```diff
@@ -0,0 +1,213 @@
...
++      case 0:
++        return ((zero_arg)func)();
++      case 1:
++        return ((one_arg)func)(self);
++      case 2:
++        return ((two_arg)func)(self, args);
++      case 3:
++        return ((three_arg)func)(self, args, kw);
++      default:
++        PyErr_SetString(PyExc_SystemError, "Handler takes too many arguments");
```
ryanking13, line 177: What happens if a function falls into here? Is this a "should not happen" condition?

ryanking13 (approved):
Thanks, @hoodmane, it looks generally good to me, Let's see what CPython devs respond in the upstream PR.

Personally, I am okay with merging this even if the upstream PR is rejected for some reason, so this does not block other JSPI related features.

cpython/patches/0008-Make-Emscripten-trampolines-work-with-JSPI.patch
hoodmane: Yes. Maybe we could fail more dramatically to make it more obvious. But I have never seen a function in the wild which takes too many arguments. The issue is always with functions that take too few arguments because they don't care about some of them.

hoodmane:
Looks like they are likely willing to merge it, I got an approval. I'm actually pretty surprised that Brett rejected your `ctypes.util.find_library` patch.

ryanking13:
> I'm actually pretty surprised that Brett rejected your ctypes.util.find_library patch.

Yeah, I think I'll try it again when they include ctypes module in their WASM build.

hoodmane:
Okay I'm going to merge this since I think the upstream PR is ready. Hopefully upstream will merge this soon too.

PR #3972, opened 2023-07-01
Backports for 0.23.4

hoodmane:
Merge don't squash.

- [x] Add environment variable for `--exports`
- [x] Update changelog on main branch


rth:
Thanks for managing this patch release!

src/js/pyodide.ts
```diff
@@ -268,6 +281,7 @@ export async function loadPyodide(
     lockFileURL: indexURL! + "repodata.json",
     args: [],
     _node_mounts: [],
+    packageCacheDir: indexURL,
```
netroy, line 284: @hoodmane thanks for the release 🙏🏽. but looks like [these lines from the original change](https://github.com/pyodide/pyodide/pull/3967/files#diff-1969cdb37fcf1a4e57d2ea723b79a2fe3ddcc87dc89a3df05e4b3f714dcc9a67R213-R222) were not cherry-picked, so the typescript complains when I try to pass in `packageCacheDir`.

PR #3973, opened 2023-07-01
ENH Add environment variable to control 'exports' cli argument

hoodmane:
If "--exports" is not passed as a CLI argument, check for an environment variable called PYODIDE_EXPORTS. If this is also missing, default to "requested".

cc @henryiii 

### Checklists

- [x] Add / update tests

### DIFF
--- a/pyodide-build/pyodide_build/cli/build.py
+++ b/pyodide-build/pyodide_build/cli/build.py
@@ -21,9 +21,8 @@ from ..out_of_tree.pypi import (
 def pypi(
     package: str,
     output_directory: Path,
-    exports: str
-    | None = typer.Option(
-        None,
+    exports: str = typer.Option(
+        "",
         help="Which symbols should be exported when linking .so files?",
     ),
     ctx: typer.Context = typer.Context,  # type: ignore[assignment]
@@ -61,9 +60,8 @@ def download_url(url: str, output_directory: Path) -> str:
 def url(
     package_url: str,
     output_directory: Path,
-    exports: str
-    | None = typer.Option(
-        None,
+    exports: str = typer.Option(
+        "",
         help="Which symbols should be exported when linking .so files?",
     ),
     ctx: typer.Context = typer.Context,  # type: ignore[assignment]
@@ -90,9 +88,8 @@ def url(
 def source(
     source_location: Path,
     output_directory: Path,
-    exports: str
-    | None = typer.Option(
-        None,
+    exports: str = typer.Option(
+        "",
         help="Which symbols should be exported when linking .so files?",
     ),
     ctx: typer.Context = typer.Context,  # type: ignore[assignment]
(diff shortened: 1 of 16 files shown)


pyodide-build/pyodide_build/io.py
```diff
@@ -75,7 +76,19 @@ def _check_patches_extra(cls, values: dict[str, Any]) -> dict[str, Any]:
...
+    requested = "requested"
+    whole_archive = "whole_archive"
+
+
 _BuildSpecExports = Literal["pyinit", "requested", "whole_archive"]
+
+
+def _convert_exports(e: _EnumExports | None) -> _BuildSpecExports | None:
+    return cast(_BuildSpecExports | None, e)
+
```
hoodmane, line 90: This is awkward but we need it until typer merges something like this: https://github.com/tiangolo/typer/pull/429

henryiii:
Not at computer, but Typer (if that’s what you are using) has integrated support for envvars: https://typer.tiangolo.com/tutorial/arguments/envvar/ (most CLI parsers do).

hoodmane:
Ah great thanks for pointing that out @henryiii.

rth:
Thanks! Could you also document it under https://pyodide.org/en/stable/development/building-from-sources.html#environment-variables ? Maybe it would make sense to have something more specific? Like `PYODIDE_BUILD_EXPORTS` to indicate that it's related to the `pyodide build` command. There are a lot of things that could be exports in different contexts. Otherwise LGTM.

hoodmane:
@rth @ryanking13 could one of you give this a quick review? I want to backport it to 0.23.4 and it's the last thing I want to add to that release.

pyodide-build/pyodide_build/pypabuild.py
```diff
@@ -186,7 +186,7 @@ def get_build_env(
     cxxflags: str,
     ldflags: str,
     target_install_dir: str,
-    exports: _BuildSpecExports | list[_BuildSpecExports],
+    exports: _BuildSpecExports | list[str],
```
ryanking13, line 189: Maybe we can define a type for `_BuildSpecExports | list[str]`.  It is used everywhere in this PR.

pyodide-build/pyodide_build/cli/build.py
```diff
@@ -18,11 +20,30 @@
...
 
+def convert_exports(exports: str) -> _BuildSpecExports | list[str]:
+    if "," in exports:
+        return [x.strip() for x in exports.split(",") if x.strip()]
+    if exports == "pyinit":
+        return "pyinit"
+    if exports == "requested":
+        return "requested"
+    if exports == "whole_archive":
+        return "whole_archive"
```
ryanking13, line 31: Maybe:

```py
from typing import get_args
possible_exports = get_args(_BuildSpecExports)
if exports in possible_exports:
    return exports
```

ryanking13 (approved):
Thanks, @hoodmane! and sorry for the late review. I have some minor comments otherwise LGTM. Please also update the changelog.

pyodide-build/pyodide_build/cli/build.py
hoodmane: I think this pattern doesn't convince mypy? You could say `return cast(_BuildSpecExports, exports)`.

hoodmane: Thanks for the input @henryiii and @ryanking13!

rth:
Nice! LGTM as well and CI is green.

PR #3975, opened 2023-07-03
Add a exports/types field to package.json

khill-fbmc:
Forgive me if this isn't right, but don't we need to export the types as well?

### Description

I was having trouble in VSCode with no type hints

### Solution

Add `"types": "./pyodide.d.ts"` to the exports definition.

### DIFF
--- a/src/js/package.json
+++ b/src/js/package.json
@@ -42,7 +42,8 @@
   "exports": {
     ".": {
       "require": "./pyodide.js",
-      "import": "./pyodide.mjs"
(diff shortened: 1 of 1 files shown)


ryanking13 (approved):
Thanks for the repro @khill-fbmc! After reading the [typescript docs](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-4-7.html#packagejson-exports-imports-and-self-referencing), it looks like this is the right change.

hoodmane:
Thanks @khill-fbmc!

Issue #3979, opened 2023-07-05
A Simple JSON API for Pyodide packages


rth:
It could be useful to expose a Simple JSON API for packages built by Pyodide, so they can be easier interacted with by existing Python tooling (e.g. https://github.com/pyodide/pyodide-lock/issues/10).  As @ryanking13 mentioned we have [`pyodide_build/create_pypa_index.py`](https://github.com/pyodide/pyodide/blob/main/pyodide-build/pyodide_build/create_pypa_index.py) which generates the Simple HTML API, but json would be better.

Another thing is maybe it would be better to include all available builds rather than having an index per release. 

So I did a quick experimentation with a [script](https://github.com/rth/pyodide/blob/simple-json-api-script/tools/update-simple-json-api.py) that,
1. Parses tags from the pyodide github repo
2. Feches the corresponding `pyodide-lock.json`
3. Write a Simple JSON API with the information it finds

The result can experimentally be seen at,
```
https://pyodide-pypi-api.s3.amazonaws.com/simple/
```
and e.g. 
```
https://pyodide-pypi-api.s3.amazonaws.com/simple/numpy/
```
for a given package.

While I was aware of it, implementing this made me realize that there is really no dependency information in this API. That's fine if there is only a couple of packages, but for a large dependency tree it's pretty bad as one has to download wheels to discover dependencies. I think there may have been another PEP for including dependency information, I need to look into it.

hoodmane:
There is another pep for dependency information.

rth:
Ahh yes that one, thanks!

> the repository MUST serve the distribution’s Core Metadata file alongside the distribution with a .metadata appended to the distribution’s file name. For example, the Core Metadata of a distribution served at /files/distribution-1.0-py3.none.any.whl would be located at /files/distribution-1.0-py3.none.any.whl.metadata. 

A bit unfortunate it has to be under the same base path.

rth:
@freakboy3742 let me know if the above API works for you. This exposes all Pyodide releases but normally if you query for a given plaform tag it should work. For instance, https://cdn.jsdelivr.net/pyodide/v0.23.4/full/repodata.json has `emscripten_3_1_32` platform tag so the wheels for Pyodide v0.23.4 should be of the form `*-cp311-cp311-emscripten_3_1_32_wasm32.whl`. There might be multiple compatible files, due to patch release in which case the last one should be taken.

freakboy3742:
[...]

PR #3981, opened 2023-07-06
Implement PEP 658: .whl.metadata files

### LINKED ISSUE #3980: Implement PEP 658: .whl.metadata files
To make Simple JSON API more useful(https://github.com/pyodide/pyodide/issues/3979, ), it would be good to deploy the .whl.metadata files per [PEP 658](https://peps.python.org/pep-0658/), so that Python packages manager could more easily identify dependencies. 

This can be done by for instance,
 1. adding a binary  `--metadata-files` option to [`pyodide build-recipes`](https://github.com/pyodide/pyodide/blob/main/pyodide-build/pyodide_build/cli/build_recipes.py#L11) (default to False for backward compatibility)
 2. set to True in the [packages/Makefile](https://github.com/pyodide/pyodide/blob/main/packages/Makefile#L16)
 3. somewhere around [here](https://github.com/pyodide/pyodide/blob/main/pyodide-build/pyodide_build/buildpkg.py#L862) once the wheel is created,  as described in the PEP, extract the METADATA file, and write it in a file named identically to the wheel but with a `.metadata` suffix (see PEP).
 4. Check that such files will be deployed the to CDN with the correct content type by adding another extension check [here](https://github.com/pyodide/pyodide/blob/main/tools/deploy_s3.py#L135) and setting it to the appropriate content type (probably `text/plain` ?)
 5. Ask for this new file extension to be whitelisted in the CDN (I can do it).

@juntyr this could be a good first issue making small changes to different parts of the build system if you are interested.

juntyr:
### Description

To make Simple JSON API more useful (https://github.com/pyodide/micropip/issues/62, https://github.com/pyodide/pyodide/issues/3979), we now deploy the .whl.metadata files per [PEP 658](https://peps.python.org/pep-0658/), so that Python packages manager could more easily identify dependencies.

A `--metadata-files` boolean option is added to the `pyodide build-recipes` command, which defaults to `False` for backward compatibility. When the option is passed alongside the `--install` option and without the `--no_deps` option, the METADATA file of each wheel file `FILENAME.whl` is now installed alongside in `FILENAME.whl.metadata`.

Fixes #3980

### Checklists

- [x] Add / update tests

### DIFF
--- a/pyodide-build/pyodide_build/buildall.py
+++ b/pyodide-build/pyodide_build/buildall.py
@@ -806,7 +806,10 @@ def install_packages(
 
     logger.info(f"Copying built packages to {output_dir}")
     copy_packages_to_dist_dir(
-        pkg_map.values(), output_dir, compression_level=compression_level, metadata_files=metadata_files,
+        pkg_map.values(),
+        output_dir,
+        compression_level=compression_level,
+        metadata_files=metadata_files,
     )
 
     lockfile_path = output_dir / "pyodide-lock.json"
--- a/pyodide-build/pyodide_build/buildall.py
+++ b/pyodide-build/pyodide_build/buildall.py
@@ -78,7 +78,7 @@ class BasePackage:
     def needs_rebuild(self) -> bool:
         return needs_rebuild(self.pkgdir, self.pkgdir / "build", self.meta.source)
 
-    def build(self, build_args: BuildArgs) -> None:
+    def build(self, build_args: BuildArgs, extract_metadata_file: bool = False) -> None:
         raise NotImplementedError()
 
     def dist_artifact_path(self) -> Path:
@@ -131,7 +131,7 @@ class Package(BasePackage):
             return tests[0]
         return None
 
-    def build(self, build_args: BuildArgs) -> None:
+    def build(self, build_args: BuildArgs, extract_metadata_file: bool = False) -> None:
         p = subprocess.run(
             [
                 sys.executable,
@@ -149,7 +149,7 @@ class Package(BasePackage):
                 # been updated and should be rebuilt even though its own
                 # files haven't been updated.
                 "--force-rebuild",
-            ],
+            ] + (["--metadata-file"] if extract_metadata_file else []),
             check=False,
             stdout=subprocess.DEVNULL,
             stderr=subprocess.DEVNULL,
@@ -488,6 +488,7 @@ def generate_needs_build_set(pkg_map: dict[str, BasePackage]) -> set[str]:
 def build_from_graph(
     pkg_map: dict[str, BasePackage],
     build_args: BuildArgs,
+    extract_metadata_files: bool = False,
     n_jobs: int = 1,
     force_rebuild: bool = False,
 ) -> None:
@@ -585,7 +586,7 @@ def build_from_graph(
 
             success = True
             try:
-                pkg.build(build_args)
+                pkg.build(build_args, extract_metadata_file=extract_metadata_files)
             except Exception as e:
                 built_queue.put(e)
                 success = False
@@ -706,7 +707,7 @@ def generate_lockfile(
 
 
 def copy_packages_to_dist_dir(
-    packages: Iterable[BasePackage], output_dir: Path, compression_level: int = 6
+    packages: Iterable[BasePackage], output_dir: Path, compression_level: int = 6,
 ) -> None:
     for pkg in packages:
         if pkg.package_type == "static_library":
@@ -728,6 +729,7 @@ def build_packages(
     packages_dir: Path,
     targets: str,
     build_args: BuildArgs,
+    extract_metadata_files: bool = False,
     n_jobs: int = 1,
     force_rebuild: bool = False,
 ) -> dict[str, BasePackage]:
@@ -737,7 +739,7 @@ def build_packages(
         packages_dir, set(requested_packages.keys()), disabled
     )
 
-    build_from_graph(pkg_map, build_args, n_jobs, force_rebuild)
+    build_from_graph(pkg_map, build_args, extract_metadata_files, n_jobs, force_rebuild)
     for pkg in pkg_map.values():
         assert isinstance(pkg, Package)
 
@@ -773,7 +775,7 @@ def copy_logs(pkg_map: dict[str, BasePackage], log_dir: Path) -> None:
 
 
 def install_packages(
-    pkg_map: dict[str, BasePackage], output_dir: Path, compression_level: int = 6
+    pkg_map: dict[str, BasePackage], output_dir: Path, compression_level: int = 6, metadata_files: bool = False
 ) -> None:
     """
     Install packages into the output directory.
--- a/pyodide-build/pyodide_build/buildall.py
+++ b/pyodide-build/pyodide_build/buildall.py
@@ -18,6 +18,7 @@ from queue import PriorityQueue, Queue
 from threading import Lock, Thread
 from time import perf_counter, sleep
 from typing import Any
+from zipfile import ZipFile
 
 from pyodide_lock import PyodideLockSpec
 from pyodide_lock.spec import PackageSpec as PackageLockSpec
@@ -78,7 +79,7 @@ class BasePackage:
     def needs_rebuild(self) -> bool:
         return needs_rebuild(self.pkgdir, self.pkgdir / "build", self.meta.source)
 
-    def build(self, build_args: BuildArgs, extract_metadata_file: bool = False) -> None:
+    def build(self, build_args: BuildArgs) -> None:
         raise NotImplementedError()
 
     def dist_artifact_path(self) -> Path:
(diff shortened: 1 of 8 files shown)


juntyr:
@juntyr TODO: handle the build all case to propagate the metadata file extraction option

pyodide-build/pyodide_build/buildpkg.py
```diff
@@ -547,6 +550,16 @@ def package_wheel(
         finally:
             shutil.rmtree(test_dir, ignore_errors=True)
 
+    if extract_metadata_file:
+        with TemporaryDirectory() as temp_dir:
+            unpack_wheel(wheel, Path(temp_dir))
```
rth, line 555: There is no need to unzip the whole thing. You can use [zipfile](https://docs.python.org/3/library/zipfile.html) to only load the one relevant file. A wheel is  zip file.

tools/deploy_s3.py
```diff
@@ -137,6 +137,8 @@ def deploy_to_s3_main(
                 # However, JsDelivr will currently not serve .ts file in the
                 # custom CDN configuration, so it does not really matter.
                 content_type = "text/x.typescript"
+            elif file_path.suffix == ".metadata":
```
rth, line 140: Maybe,
```suggestion
            elif file_path.name.endswith('.whl.metadata'):
```
to be a bit more specific.

pyodide-build/pyodide_build/buildpkg.py
```diff
@@ -473,6 +475,7 @@ def package_wheel(
     build_metadata: _BuildSpec,
     bash_runner: BashRunnerWithSharedEnvironment,
     host_install_dir: str,
+    extract_metadata_file: bool,
```
rth, line 478: Can probably give it a default value,
```suggestion
    extract_metadata_file: bool = False
```

juntyr:
@rth Thank you for your feedback! I've prototyped two possible implementations.

The first one, based on your suggestions can be seen at https://github.com/pyodide/pyodide/compare/main...juntyr:pyodide:54c482b849846ff01e323824c4af84e2116916ba (note though that it's only 95% complete). It pushes the metadata extraction down to building the package. However, this requires quite invasive code changes. Furthermore, the interaction with installing is a bit weird. What if the package was built with extraction but is installed without metadata? What if it's the other way around?

At the moment, I prefer the second option, which you can see in the current diff. Metadata extraction is now pushed to the installation stage, which means the diff is very clean.

rth (commented):
Thanks @juntyr ! No the current version is fine, that was just an approximate suggestion as to where to put it in the code.

Overall LGTM assuming tests pass, but since I was the one who proposed this feature it would be good to have a final review/apporval done by another maintainer.

pyodide-build/pyodide_build/cli/build_recipes.py
```diff
@@ -91,9 +97,9 @@ def recipe(
     build_args = buildall.set_default_build_args(build_args)
 
     if no_deps:
-        if install or log_dir_:
+        if install or log_dir_ or metadata_files:
             logger.warning(
-                "WARNING: when --no-deps is set, --install and --log-dir parameters are ignored",
+                "WARNING: when --no-deps is set, the --install, --log-dir, and --metadata-files parameters are ignored",
```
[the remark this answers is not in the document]
juntyr: In that case, no installation happens, so the extraction is not run. We could still run it and also extract the metadata into the per-package dist-files, but I again prefer not infecting the building code as deeply as option 1 requires.

pyodide-build/pyodide_build/cli/build_recipes.py
```diff
@@ -118,5 +124,12 @@ def recipe(
...
             buildall.install_packages(
-                pkg_map, install_dir_, compression_level=compression_level
+                pkg_map,
+                install_dir_,
+                compression_level=compression_level,
+                metadata_files=metadata_files,
+            )
+        else:
+            logger.warning(
+                "WARNING: when --install is not set, the --metadata-files parameter is ignored",
```
ryanking13, line 134: I think we should only show this warning when the metadata-files option is set.

pyodide-build/pyodide_build/buildall.py
```diff
@@ -719,6 +723,15 @@ def copy_packages_to_dist_dir(
             output_dir / dist_artifact_path.name, compression_level=compression_level
         )
 
+        if metadata_files and dist_artifact_path.suffix == ".whl":
```
ryanking13, line 726: Could you move this (extracting metadata file from a wheel) into a separate function, and add some tests for the function? I think you can put it in `common.py`.

ryanking13 (changes requested):
Thanks @juntyr for working on this!

pyodide-build/pyodide_build/buildall.py
```diff
@@ -719,6 +723,15 @@ def copy_packages_to_dist_dir(
...
+            with ZipFile(dist_artifact_path, mode="r") as wheel:
+                name, ver, _ = dist_artifact_path.name.split("-", 2)
+                metadata_path = str(Path(f"{name}-{ver}.dist-info") / "METADATA")
```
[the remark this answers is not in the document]
juntyr: FYI, your proposed approach could be implemented as follows:
```python
metadata_infos = [
    info for info in wheel.infolist() if Path(info).name == "METADATA" and len(Path(info).parts) == 2
]

if len(metadata_infos) > 1:
    raise Exception(
        f"Unexpected number of METADATA files when building {pkg.name}: {len(metadata_infos)}"
    )
elif len(metadata_infos) == 1:
    metadata_info = metadata_infos[0]
    metadata_path = metadata_info.filename
    metadata_info.filename = f"{dist_artifact_path.name}.metadata"
    wheel.extract(metadata_path, output_dir)
```

pyodide-build/pyodide_build/buildall.py
juntyr: Thank you for the suggestions - I've extracted a helper function and added two tests for it.

pyodide-build/pyodide_build/buildall.py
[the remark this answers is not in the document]
ryanking13: I think it is okay to raise an error and quit if some packages does not follow the PEP. I believe `pip install` will also fail on those cases.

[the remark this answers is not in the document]
ryanking13: You can probably adapt how pip finds a .dist_info dir: https://github.com/pypa/pip/blob/ea727e4d6ab598f34f97c50a22350febc1214a97/src/pip/_internal/utils/wheel.py#L38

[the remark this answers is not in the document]
juntyr: > Probably it would be useful to check how pip finds a .dist_info dir: https://github.com/pypa/pip/blob/ea727e4d6ab598f34f97c50a22350febc1214a97/src/pip/_internal/utils/wheel.py#L38

I'm not sure I'm happy with my latest changes, as they increase the complexity a bit. However, we would now use the same functionality as pip.

pyodide-build/pyodide_build/tests/test_common.py
```diff
@@ -123,3 +124,35 @@ def test_repack_zip_archive(
...
+
+    input_path_empty = tmp_path / "pkg-0.2-abc.whl"
+
+    with zipfile.ZipFile(input_path_empty, "w") as fh:
+        pass
+
+    output_path_empty = tmp_path / f"{input_path_empty.name}.metadata"
+
+    try:
+        extract_wheel_metadata_file(input_path_empty, output_path_empty)
```
ryanking13, line 154: nitpick: you can use `pytests.raises` here:

```python
with pytest.raises(SomeException):
    extract_wheel_metadata_file(...)
```

pyodide-build/pyodide_build/common.py
```diff
@@ -312,3 +314,76 @@ def modify_wheel(wheel: Path) -> Iterator[Path]:
...
+            )
+        )
+
+    return info_dir
+
+
+_canonicalize_regex = re.compile(r"[-_.]+")
+
+
+def _canonicalize_package_name(name: str) -> str:
```
ryanking13, line 383: pyodide-build depends on `packaging`, so you can import this function from it:

```py
from packaging.utils import canonicalize_name
```

ryanking13 (approved):
Thanks, @juntyr!

pyodide-build/pyodide_build/tests/test_common.py
juntyr: Thanks for the suggestion

pyodide-build/pyodide_build/common.py
juntyr: Ah, good catch :)

juntyr:
Are the [CLI docs](https://pyodide.org/en/stable/usage/api/pyodide-cli.html) autogenerated? If so, I probably don't need to update any further documentation.

juntyr:
@ryanking13 @rth Thank you so much for your feedback! This PR should now be ready to me squashed and merged.

ryanking13:
Thanks!

> Are the [CLI docs](https://pyodide.org/en/stable/usage/api/pyodide-cli.html) autogenerated? If so, I probably don't need to update any further documentation.

Yes, it is auto-generated.

rth:
Thanks a lot @juntyr and @ryanking13 for the review!

PR #3987, opened 2023-07-10
Move stack switching logic into a new stack switching folder.

hoodmane:
This is a minor rearrangment.

I renamed "continuations" stuff to "stack_switching" and made a new folder for stack switching logic. I also added logic in esbuild.config.mjs to calculate which functions exported from stack_switching.mjs and automatically copy them to Module and into the Emscripten namespace. This makes unit testing a little neater since there is no modification of the Module object.


Makefile
```diff
@@ -225,9 +225,10 @@ benchmark: all
 
 clean:
 	rm -fr dist/*
-	rm -fr src/*/*.o
-	rm -fr src/*/*.gen.*
 	rm -fr node_modules
+	shopt -s globstar && rm -fr src/**/*.o
```
rth, line 229: On my Ubtunu  system,
```
zsh: command not found: shopt
```
so it's bash specific. Not sure if it's worth having bash-specific commands here.

Similarly MacOS has zsh so it will not work there.

rth (approved):
One comment otherwise LGTM.

Makefile
hoodmane: I switched to using `find -delete` which should hopefully work more broadly.

hoodmane:
Thanks for the review @rth!

Issue #3989, opened 2023-07-10
Add `igraph` package


ntamas:
## 🐍 Package Request

- Package Name and Version : igraph / latest
- Package URL : https://github.com/igraph/python-igraph, https://pypi.org/project/igraph/
- Package Dependencies that needs to be resolved first: none

## Progress so far

`python-igraph` already supports WASM wheels, its `setup.py` is fully prepared to handle WASM builds, and WASM wheels are regularly built in CI (see [here](https://github.com/igraph/python-igraph/blob/main/.github/workflows/build.yml#L169)). I believe it should be easy to add a recipe for `pyodide` and I'm willing to experiment with that; I just need feedback from the maintainers to see if such a contribution would be accepted before getting my hands dirty :)

Related issue in igraph's issue tracker: https://github.com/igraph/python-igraph/issues/560

ryanking13:
Sure, we welcome PRs that add new packages!

szhorvat:
Any chance to make this happen by Thursday evening CET, for a bit of extra advertising opportunity?

szhorvat:
@ryanking13 If the PR can me merged today, I would be able to demo it tomorrow at a presentation and give pyodide some extra publicity.

ryanking13:
@szhorvat The PR is merged and you'll be able to use it in the [latest console](https://pyodide.org/en/latest/console.html) soon.

PR #3991, opened 2023-07-11
Add `igraph` package

ntamas:
### Description

This PR adds support for the `igraph` package in Pyodide, as discussed in #3989

Currently igraph has a dependency on the `texttable` package, which is pure Python so it could be used as-is with Pyodide, but it is not in the `packages` folder yet. The `texttable` package is needed only for a very minor subset of igraph's functionality (pretty-printing vertex and edge attribute tables).

### Checklists

- [x] Add / update tests


ntamas:
Please let me know how to proceed with the inclusion of the `texttable` package (if needed); see the PR description for more details.

ntamas:
Seems like the package builds correctly now; one of the CI failures is due to a timeout and the other one is not related to the newly added package.

ryanking13:
Thanks, @ntamas! Please update the changelog. 

> Please let me know how to proceed with the inclusion of the texttable package (if needed); see the PR description for more details.

If `texttable` is an optional dependency for `igraph`, then it is up to you whether or not to add it.

You can run `pyodide skeleton pypi texttable` to add a recipe for texttable, then you need to add the following in the `igraph` recipe.

```
requirements:
  run:
    - texttable
```

ntamas:
Converting to draft because a new release of `igraph` (0.10.6) is about to happen in a few hours and it would be great to get this PR merged with that version instead.

ntamas:
`igraph` updated to 0.10.6 and it builds fine locally on my machine, so let's see how this goes in CI.

packages/igraph/meta.yaml
```diff
@@ -0,0 +1,27 @@
+package:
+  name: igraph
+  version: 0.10.5
```
ryanking13, line 3:
```suggestion
  version: 0.10.6
```

ryanking13 (approved):
Thanks a lot, @ntamas!

PR #3992, opened 2023-07-12
Stack Switching: Save the Python state when suspending and restore it when done

hoodmane:
This allows bidirectional stack switching and other arbitrary control flow, whereas
without this patch arbitrary control flow terribly messes up the Python interpreter
state.

By far the most subtle thing here is the manipulation of the cframe state. This
required careful reading of the greenlet source code and a large amount of trial
and error to get right.

Unfortunately the file pystate.c is tightly coupled to the Python version. There are
apparently plans to try to upstream some of this code from greenlet into Python itself.
In the meantime, it should be possible to continue updating it by imitating greenlet.


ryanking13 (approved):
Thanks for working on this @hoodmane! I tried similar stuff in #3705 using only public APIs but failed and gave up, so I am very glad you made this work :)

I didn't read the code in detail TBH, as I don't know about the internal and private frame state of CPython, but it generally looks okay to me (thanks for the detailed comment).

ryanking13:
Thanks again @hoodmane!

PR #3994, opened 2023-07-12
Only schedule Python coroutine if it's the result of calling an async function

hoodmane:
When we call a Python function from JavaScript, if the result is a coroutine we have logic to automatically schedule it. This is so that a coroutine function can be used as an event handler or otherwise as a drop in replacement for a JS async function: when a JS async function is called, its execution is automatically scheduled, whereas in Python an async function must be awaited or scheduled with `create_task`. Contexts that expect JS async functions may not do any of these things but still want the function to execute.

A confusing edge case that arises is when a coroutine is returned from a function which is not a Python async function. This most often happens by accident, for example
 with `eval_code`. It's perhaps slightly unexpected that we schedule it in this case.
This PR adjusts it so that if a non-async python is called and it returns a coroutine, we no longer schedule it.

One edge case that's still here is that if an async function returns a coroutine, that coroutine will get automatically scheduled:

```js
pyodide.runPython(`
async def f():
   return g()

async def g():
   print("This is executed")
`);
pyodide.globals.get("g")()
```

This is because we attach the coroutine to a JS promise and if a JS promise resolves to a thenable, the thenable is automatically thened. If we were sufficiently upset about this, we could use a custom thenable that doesn't have this flattening behavior.

@antocuni 

- [x] Add / update tests

### DIFF
--- a/src/core/pyproxy.c
+++ b/src/core/pyproxy.c
@@ -700,6 +700,8 @@ finally:
 bool
 _iscoroutinefunction(PyObject* f)
 {
+  _Py_IDENTIFIER(_is_coroutine_marker);
+
   // Some fast paths for common cases to avoid calling into Python
   if (PyMethod_Check(f)) {
     f = PyMethod_GET_FUNCTION(f);
@@ -709,7 +711,7 @@ _iscoroutinefunction(PyObject* f)
   // to make sure we don't accidentally return false negatives when we update to
   // 3.12.
   if (PyFunction_Check(f) &&
-      !PyObject_HasAttrString(f, "_is_coroutine_marker")) {
+      !PyObject_HasAttr(f, _PyUnicode_FromId(&PyId__is_coroutine_marker))) {
     PyFunctionObject* func = (PyFunctionObject*)f;
     PyCodeObject* code = (PyCodeObject*)PyFunction_GET_CODE(func);
     return (code->co_flags) & CO_COROUTINE;
--- a/src/core/pyproxy.c
+++ b/src/core/pyproxy.c
@@ -700,11 +700,29 @@ finally:
 bool
 _iscoroutinefunction(PyObject* f)
 {
+  // Some fast paths for common cases to avoid calling into Python
+  if (PyMethod_Check(f)) {
+    f = PyMethod_GET_FUNCTION(f);
+  }
+
+  // _is_coroutine_marker is added to Python stdlib in 3.12. Check for it here
+  // to make sure we don't accidentally return false negatives when we update to
+  // 3.12.
+  if (PyFunction_Check(f) &&
+      !PyObject_HasAttrString(f, "_is_coroutine_marker")) {
+    PyFunctionObject* func = (PyFunctionObject*)f;
+    PyCodeObject* code = (PyCodeObject*)PyFunction_GET_CODE(func);
+    return (code->co_flags) & CO_COROUTINE;
+  }
+
+  // Wasn't a basic callable, call into inspect.iscoroutinefunction
   PyObject* result = PyObject_CallOneArg(iscoroutinefunction, f);
   if (!result) {
     PyErr_Clear();
   }
-  return Py_IsTrue(result);
+  bool ret = Py_IsTrue(result);
+  Py_CLEAR(result);
+  return ret;
 }
 
 JsRef
--- a/src/core/pyproxy.c
+++ b/src/core/pyproxy.c
@@ -26,10 +26,11 @@ check_gil()
   }
 }
 
-PyObject* Generator;
-PyObject* AsyncGenerator;
-PyObject* Sequence;
-PyObject* MutableSequence;
+static PyObject* Generator;
+static PyObject* AsyncGenerator;
+static PyObject* Sequence;
+static PyObject* MutableSequence;
+static PyObject* iscoroutinefunction;
 
 _Py_IDENTIFIER(result);
 _Py_IDENTIFIER(pop);
@@ -694,6 +695,16 @@ finally:
   Py_CLEAR(pyresult);
   Py_CLEAR(pykwnames);
   return idresult;
+}
+
+bool
+_iscoroutinefunction(PyObject* f)
+{
+  PyObject* result = PyObject_CallOneArg(iscoroutinefunction, f);
+  if (!result) {
+    PyErr_Clear();
+  }
+  return Py_IsTrue(result);
 }
 
 JsRef
@@ -1401,6 +1412,7 @@ pyproxy_init(PyObject* core)
 
   PyObject* collections_abc = NULL;
   PyObject* docstring_source = NULL;
+  PyObject* inspect = NULL;
 
   collections_abc = PyImport_ImportModule("collections.abc");
   FAIL_IF_NULL(collections_abc);
@@ -1421,9 +1433,15 @@ pyproxy_init(PyObject* core)
   FAIL_IF_NULL(asyncio);
   FAIL_IF_MINUS_ONE(PyType_Ready(&FutureDoneCallbackType));
 
+  inspect = PyImport_ImportModule("inspect");
+  FAIL_IF_NULL(inspect);
+  iscoroutinefunction = PyObject_GetAttrString(inspect, "iscoroutinefunction");
+  FAIL_IF_NULL(iscoroutinefunction);
+
   success = true;
 finally:
   Py_CLEAR(docstring_source);
   Py_CLEAR(collections_abc);
+  Py_CLEAR(inspect);
   return success ? 0 : -1;
 }
--- a/src/core/pyproxy.ts
+++ b/src/core/pyproxy.ts
@@ -445,7 +445,12 @@ Module.callPyObjectKwargs = function (
   let result = Hiwire.pop_value(idresult);
   // Automatically schedule coroutines
   if (result && result.type === "coroutine" && result._ensure_future) {
-    result._ensure_future();
+    Py_ENTER();
+    let is_coroutine = Module.__iscoroutinefunction(ptrobj);
+    Py_EXIT();
+    if (is_coroutine) {
+      result._ensure_future();
+    }
   }
   return result;
 };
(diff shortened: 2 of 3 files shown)


src/core/pyproxy.c
```diff
@@ -696,6 +697,16 @@ _pyproxy_apply(PyObject* callable,
...
+bool
+_iscoroutinefunction(PyObject* f)
+{
+  PyObject* result = PyObject_CallOneArg(iscoroutinefunction, f);
+  if (!result) {
+    PyErr_Clear();
+  }
+  return Py_IsTrue(result);
+}
+
```
antocuni, line 709: so, if I understand correctly, now we are calling the app-level `inspect.iscoroutinefunction()` every time we call a PyProxy, is it correct?
I don't know what is the actual performance penalty, but it might be non-negligible.

`inspect.iscoroutinefunction` seems very trivial, although `_has_code_flag()` is not completely trivial:
https://github.com/antocuni/cpython/blob/deaf509e8fc6e0363bd6f26d52ad42f976ec42f2/Lib/inspect.py#L391-L414

maybe it is worth considering adding a fast path for common cases like "plan function" and "plain bound method", and check the flags directly in C. Something along this line (mostly pseudocode):
```c
bool iscoroutinefunciton(PyObject *obj) {
    if (PyFunction_Check(obj)) {
        PyFunctionObject *func = (PyFunctionObject*)obj;
        return func->func_code->co_flags & CO_COROUTINE;
    }
    else if (PyMethod_Check(obj)) {
        ...
    }
    else {
        // fallback to call inspect.iscoroutinefunction as we do now
    }
}
```

antocuni (approved):
LGTM.
If I understand correctly, this changes the behavior of also this case:
```python
import js

async def foo():
    print('foo')

def bar():
    return foo()

js.bar = bar

// ... from javascript
window.bar();
```

With the old logic, `foo()` was automatically awaited, but with the new one it won't be. 
I *think* that the equivalent code in JS behaves differently, because there is no difference between async-functions and normal-functions-which-happens-to-return-a-promise.

Basically, I think that we are in a situation in which it is impossible to win, because any decision leads to some "unexpected behavior", depending on the expectations of whoever writes/reads the code -- and this happens because the expectations are very different between Python and JS.

I think that given the current state of affairs, the behavior introduced by this PR is the least bad: we keep supporting the common case of "calling a python async function from an event handler", but preserving semi-sane behavior for other cases.

My ideal solution would be to just avoid the auto-await in all cases, and provide some easy-to-use syntax for people who wants to call-a-pyproxy-and-automatically-schedule-it, e.g. something like this:
```js
let foo = pyodide.globals.get('foo');
foo(); // doesn't auto-await
foo.schedule(); // DOES auto await
// or other possible names:
foo.await_eventually();
foo.ensure_future();
foo.create_task();
...
```

But I suppose it's too late to introduce this change without breaking all the existing pyodide code around

hoodmane:
> My ideal solution would be to just avoid the auto-await in all cases, and provide some easy-to-use syntax for people who wants to call-a-pyproxy-and-automatically-schedule-it

It worked like this for one major release of Pyodide and people were not that happy about it. A handful of people did something like:
```
pyodide.runPython(`
async def handler(event):
   # async code
`);
emitter.addEventHandler("event", pyodide.globals.get("handler"));
```
Of course I told these people to do:
```js
const handler = pyodide.globals.get("handler");
emitter.addEventHandler("event", (...args) => Promise.resolve(handler(...args)));
```
but this was surprising to people.

On the other hand, a lot of people don't love "this borrowed proxy was destroyed automatically at the end of a function call" but they have to deal with it because we decided that it was worth it to make them unhappy. Though that is also important for compatibility since otherwise you run into weird problems like someone's utility routine:
```py
def my_utility(callback):
   o = ClassWithDestructor()
   callback(o)
   del o # I am a bad person and my code assumes o's destructor is called here
```
and then this won't work when `callback` is a JS function.

src/core/pyproxy.c
hoodmane: Looks like this Python 3.12 added [`markcoroutinefunction`](https://docs.python.org/3.12/library/inspect.html#inspect.markcoroutinefunction) which adds a possibility to get false negatives. But we can check if `hasattr(o, "_is_coroutine_marker")` and fall back to `inspect.iscoroutinefunction` in that case.

This does add a bit of extra maintenance burden when updating Python versions...

antocuni: it's debatable whether they are false negatives.
I'll write my thoughts in the main discussion

antocuni:
Ok, the more I think about it, the more it's a mess :man_facepalming:

With the advent of `markcoroutinefunction`, we have basically four cases:
1. `async def` functions
2. `def` functions which happens to return a coroutine
3. `def` functions which happens to return a coroutine and are `@markcoroutinefunction`
4. `def` functions which are `@makrcoroutinefunction` but they don't actually return a coroutine (I'm sure they will exist in the wild, eventually)

I fear that any attempt to try to be "too smart" will end up in broken corner cases.

The question is: what is the behavior that we want?
- the `python way`: coroutines are never auto awaited
- the `JS way`: any coroutine/task/future/promise which is returned by some function is automatically scheduled, independently on whether the function itself is `async` or not. I think this is the current behavior.
- the `autoawait-async-def way`: I think this is what this PR is trying to do: if you call an `async def` python function from JS, it is automatically awaited
- some other option which I can't think of?

I think that the `autoawait-async-def way` opens a can of worms, specifically because the distinction between "async-def functions" and "def functions returning a coroutine" is a bit blurry, and the fact that they are introducing `@markcoroutinefunction` makes me thinking that in the long run we will have a mess of "coroutine-functions-which-are-not-really-but-people-expects-to-be".

If we are fine with breaking compatibility, I still think that `the python way` is the cleanest solution and the one which will cause the least amount of weird surprises in the long run.

If the `python way` is not an option, then we can stick to the `JS way` and fix the original problem with a different hack: we could mark `eval_code` with a special decorator which says "don't auto-await the returned value".

(Unrelated but: we should also question whether using `eval_code` is the right thing to do for `runPython`: I believe that `let's try to magically turn python into a functional language and return the "last result"` is a fundamentally broken game, which is basically good only for REPLs: so `runPython` should do the "sane" thing by default (exec), and do "eval" only if explicitly asked. Else, you open the door for a can of worm like this one :))

ryanking13 (approved):
Thanks, @hoodmane! My opinion is that it is okay as long as we are consistent.

> emitter.addEventHandler("event", pyodide.globals.get("handler"));

I also like this being auto-scheduled.

> def functions which are @makrcoroutinefunction but they don't actually return a coroutine (I'm sure they will exist in the wild, eventually)

🤦‍♀️

hoodmane:
The logic I currently have is pretty decent:
1. `async def` is scheduled
2. `@markcoroutinefunction` will get scheduled in 3.12 if it returns a coroutine otherwise nothing bad will happen
3. everything else is not scheduled even if it returns a coroutine

I first check that the returned object is a coroutine and only in that case do I check whether the called object is a coroutine function.

antocuni:
> The logic I currently have is pretty decent:
[cut]

ok, your call :)

I still think that it leads to very weird corner cases which are unpredictable and unexplainable unless you know the details of the logic (so, YOU know of course, but the random pyodide user won't), and potentially the decorator could be applied by someone out of your control; e.g., imagine the following:
```python

from mylib import logging

@logging
async def onclick(...):
    ...

maybe the next version of `@logging` introduces/removes the decorator and the behavior of your app changes suddenly.
And note that for the rest of the python world this change is basically a no-op, so it will probably go unnoticed. But for pyodide becomes an incredibly big change in behavior.

Anyway, it's a small detail. The vast majority of users will never notice such a thing, and for the 0.001% which will: *hello future reader, I tried my best to save you 😅*

:)

hoodmane:
Well there are two other alternatives as I see it:
1. don't do this PR and leave the behavior as is
2. apply some special magic just in `runPython` to specifically prevent coroutines returned from `runPython` from being scheduled

It's also worth noting that it isn't just Pyodide that relies on `inspect.iscoroutinefunction`, there is other asyncio code in the wild using it. This is why the people added the `@markcoroutinefunction` decorator: because they were running into problems where their function was being treated differently depending on whether or not `iscoroutinefunction` accepted it. Decorators that don't use `functools.wraps` mess up in a variety of subtle ways. This logic is complicated but not in a completely unique way.

One thing that is annoying about this logic is compared to asyncio frameworks it's in a more exotic codebase. JS developers are less likely to understand the subtleties of Python and it will be harder for Python developers to understand the Python/C/JS codebase. But perhaps if we clearly document this behavior it will suffice...

antocuni:
> Well there are two other alternatives as I see it:
> 
> 1. don't do this PR and leave the behavior as is
> 2. apply some special magic just in `runPython` to specifically prevent coroutines returned from `runPython` from being scheduled

yes, this is what I meant here:

> fix the original problem with a different hack: we could mark `eval_code` with a special decorator which says "don't auto-await the returned value".


> It's also worth noting that it isn't just Pyodide that relies on `inspect.iscoroutinefunction`, there is other asyncio code in the wild using it. This is why the people added the `@markcoroutinefunction` decorator: because they were running into problems where their function was being treated differently depending on whether or not `iscoroutinefunction` accepted it. Decorators that don't use `functools.wraps` mess up in a variety of subtle ways. This logic is complicated but not in a completely unique way.

ah ok, I didn't know about this. Then yes, the logic proposed in this PR is not too bad (in the sense that it's still a mess, but it's caused by asyncio, not by pyodide :)).

> One thing that is annoying about this logic is compared to asyncio frameworks it's in a more exotic codebase. JS developers are less likely to understand the subtleties of Python and it will be harder for Python developers to understand the Python/C/JS codebase. 

100% agree. Basically, as I said above, I think it's a game that we cannot win: the JS and Python "expectations" are so different that *any* solution will cause some "unexpected behavior" for someone.

To be clear, I think that the solution proposed by this PR is "good enough" for most use cases, so +1 for merging this.

> But perhaps if we clearly document this behavior it will suffice...

you are very optimistic :) 😂

PR #4000, opened 2023-07-18
Disable pyproj

hoodmane:
There seems to be some cython related build failure. Maybe we need to cap cython?


hoodmane:
Ah yes, Cython 3.0.0 was released yesterday:
https://docs.cython.org/en/latest/src/changes.html#id8

ryanking13:
@hoodmane does this PR still needed?

hoodmane:
Nope.

PR #4001, opened 2023-07-18
Cap cython<3

hoodmane:
I added tools/constraints.txt and set it into the PIP_CONSTRAINT environment variable.


rth (approved):
Thanks!

ryanking13:
Thanks! Just curious, where does `PIP_CONSTRAINT` consumed? in pypa/build? we don't use `pip` so it is interesting that `PIP_*` works.

I think you need to add `*.txt` into the following config to make it vendored with the wheel. But I am not sure we want to vendor it in pyodide-build wheel, as we don't want to cap the version forever.

https://github.com/pyodide/pyodide/blob/f17b185f50636d448c58650b8ddeeefdbc4ba9f7/pyodide-build/pyproject.toml#L83

So maybe we can pass `PIP_CONSTRAINT` env variable in the Makefile without modifying pyodide-build?

hoodmane:
In a followup it might be interesting to allow recipes to have extra requirements or constraints for the packages to be downloaded from pip in the build environment. Because this sort of global constraint doesn't really let us relax it one package at a time. In fact, if a single package depended on cython3 we'd be in trouble.

PR #4005, opened 2023-07-21
Fix package name compatibility with wheel filename

rth:
Currently, package names in the Pyodide distribution may differ from the package name in the wheel filename. This prevents them from being installable with pip via the Simple JSON API (https://github.com/pyodide/pyodide/issues/3979#issuecomment-1644618784). It think it would also be an issue when installed in a Pyodide virtualenv.

In practice, this means that the package name for packages with wheels will now match the name on PyPI.

This fixes it, by adding an additional check to pyodide-lock.

TODO:
 - [x] merge and release https://github.com/pyodide/pyodide-lock/pull/11 (could someone review it please?)
 - [x] fix pillow https://github.com/pyodide/pyodide/issues/3979#issuecomment-1644618784 (I think it's probably due to us overwriting the `setup.cfg`?)
 - [x] fix tblib which currently fails.  with
       ```
        'tblib': ["Package name in wheel filename 'pyodide-tblib' does not match 'tblib'", 
                     "Version in the wheel filename '1.7.1' does not match package version '1.7'"]
       ```
       @hoodmane as far as I understand it's a fork you maintain. Any objection if we rename this package in Pyodide to `pyodide-tblib` as is the case on PyPI?


hoodmane:
Maybe let's hold this off to the next release since it isn't ready?

ryanking13:
> Maybe let's hold this off to the next release since it isn't ready?

Yes, I remove this PR from the 0.24.0 milestone.

ryanking13:
I resolved the remaining incompatible package names. Merging so we can include this in version 0.25.0. Thanks @rth!

PR #4007, opened 2023-07-22
Add docstrings for open_url, CodeRunner, JsProxy

maksbialas:
Add docstring examples (doctests) for:
* http.open_url
* code.CodeRunner
* pyodide.JsProxy's `object_entries`, `object_keys` and `object_values`

Partially addresses: https://github.com/pyodide/pyodide/issues/1955


src/py/_pyodide/_base.py
```diff
@@ -211,6 +211,24 @@ class CodeRunner:
...
+    --------
+    >>> from pyodide.code import CodeRunner
+    >>> source = "1 + 1"
+    >>> code_runner = CodeRunner(source)
+    >>> code_runner.compile()
+    <_pyodide._base.CodeRunner object at 0x113de58>
+    >>> code_runner.run()
+    2
+    >>> my_globals = { "y": 20 }
+    >>> my_locals = { "y": 5 }
```
rth, line 225: Can you run black on it somehow? Probably no spaces after { no?

Also it would be better to define a differently named variable in globals and locals says x and y. Then un the code just do x + y and it should work as far as I understand.

src/py/_pyodide/_core_docs.py
```diff
@@ -125,15 +125,46 @@ def typeof(self) -> str:
...
-        "The JavaScript API ``Object.entries(object)``"
+        """
+        The JavaScript API ``Object.entries(object)``
+
+        Examples
+        --------
+        >>> from pyodide.code import run_js
+        >>> js_obj = run_js("({first: 'aa', second: 22})")
+        >>> entries = js_obj.object_entries()
+        >>> [(key, val) for key, val in entries]
```
rth, line 136: Wouldn't  `list(entries)` produce the same thing?

src/py/_pyodide/_core_docs.py
```diff
@@ -125,15 +125,46 @@ def typeof(self) -> str:
...
-        "The JavaScript API ``Object.keys(object)``"
+        """
+        The JavaScript API ``Object.keys(object)``
+
+        Examples
+        --------
+        >>> from pyodide.code import run_js
+        >>> js_obj = run_js("({first: 1, second: 2, third: 3})")
+        >>> keys = js_obj.object_keys()
+        >>> [key for key in keys]
```
rth, line 151: Same with a `list`

src/py/_pyodide/_core_docs.py
```diff
@@ -125,15 +125,46 @@ def typeof(self) -> str:
...
-        "The JavaScript API ``Object.values(object)``"
+        """
+        The JavaScript API ``Object.values(object)``
+
+        Examples
+        --------
+        >>> from pyodide.code import run_js
+        >>> js_obj = run_js("({first: 1, second: 2, third: 3})")
+        >>> values = js_obj.object_values()
+        >>> [val for val in values]
```
rth, line 165: Same

rth (approved):
Thanks a few minor comments but otherwise LGTM once we skip these docstests.

src/py/_pyodide/_core_docs.py
hoodmane: I think the inner level will still be arrays then.

rth:
Thanks again @maksbialas!

PR #4009, opened 2023-07-22
Add support for the @example JSDoc directive

tomasr8:
### Description

Adds support for the [`@example`](https://jsdoc.app/tags-example.html) directive in JSDoc. This makes it possible to add code examples to JS functions.

This is what the code example looks like:
image


rth (approved):
Thanks @tomasr8 !

PR #4011, opened 2023-07-23
Add examples for runPython()

tomasr8:
### Description

#1955

Adds some examples for the `pyodide.runPython()` JS function

image


rth (approved):
Thanks a lot @tomasr8 !

PR #4018, opened 2023-07-26
Add .so WASM Compatibility Check

### LINKED ISSUE #4017: Add check for WASM magic number in .so files
> `pyodide build` now replaces native `.so` slugs with Emscripten
> slugs. Usually `.so`s in the generated wheels are actually Emscripten `.so`s
> so this is good. If they are actually native `.so`s then there is a problem
> either way.
  
  Not very critical, but should we actually check that the .so are emscripten .so rather than native .so that ended up there by mistake? For instance, we could check for the [WASM magic number](https://openhome.cc/eGossip/WebAssembly/Module.html) in the first 4 bytes maybe? It's supposed to be, `0061 736d`. Though I get the same bytes but in a different order, when I try,

```
$ hexdump -n 8 numpy/core/_multiarray_umath.cpython-311-wasm32-emscripten.so
0000000 6100 6d73 0001 0000 
```
but maybe I'm using hexdump wrong (**Edit:** yes, with the `-C` option it's better)

_Originally posted by @rth in https://github.com/pyodide/pyodide/issues/3927#issuecomment-1599511454_

Martolivna:
### Description

This adds `check_wasm_magic_number` function to verify .so files for WASM compatibility. The function checks the first 4 bytes for the WASM binary magic number, `b"\0asm"`.
Includes tests for validation.

Feedback appreciated. Thanks!

Resolves #4017

### Checklists

- [x] Add / update tests

### DIFF
--- a/pyodide-build/pyodide_build/common.py
+++ b/pyodide-build/pyodide_build/common.py
@@ -373,8 +373,7 @@ def get_wheel_dist_info_dir(wheel: ZipFile, pkg_name: str) -> str:
     return info_dir
 
 
-def check_wacm_magic_number(file_path: Path):
+def check_wacm_magic_number(file_path: Path) -> bool:
     WASM_BINARY_MAGIC = b"\0asm"
     with file_path.open(mode="rb") as file:
         return file.read(4) == WASM_BINARY_MAGIC
-    
--- a/pyodide-build/pyodide_build/common.py
+++ b/pyodide-build/pyodide_build/common.py
@@ -373,7 +373,7 @@ def get_wheel_dist_info_dir(wheel: ZipFile, pkg_name: str) -> str:
     return info_dir
 
 
-def check_wacm_magic_number(file_path: Path) -> bool:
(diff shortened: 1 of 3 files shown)


pyodide-build/pyodide_build/common.py
```diff
@@ -371,3 +371,9 @@ def get_wheel_dist_info_dir(wheel: ZipFile, pkg_name: str) -> str:
         )
 
     return info_dir
+
+
+def check_wacm_magic_number(file_path: Path) -> bool:
```
hoodmane, line 376: Typo: wacm --> wasm

hoodmane (approved):
Thanks @Martolivna. 

In the future, one question is whether we should also check the version number. Currently a wasm binary starts with the 8 bytes:
```
\x00asm\x00\x00\x00\x01
```
where after the magic number comes a 32 bit version number which so far has to be 1 (I think). Probably it's fine not to check it.

PR #4021, opened 2023-07-27
Fix typo in docs

tomasr8:
I assume this was meant to say _but not both_ (looking at the function body seems to confirm that)


hoodmane:
Thanks!

PR #4024, opened 2023-07-28
MAINT Hoist out wrapPythonGlobals proxy handlers

hoodmane:
More minor cleanup advised by @WebReflection.


hoodmane:
Ok it's not obvious if there's a good way to make this work.

PR #4027, opened 2023-07-31
PERF Only render extra destroyed information when debug mode is on

hoodmane:
Rendering the destroyed error messages for PyProxies is pretty inefficient. This adds a setting to turn on debug mode. When debug mode is off, a cheaper destroyed message is used instead.

- [x] Add / update tests

### DIFF
--- a/src/tests/test_jsproxy.py
+++ b/src/tests/test_jsproxy.py
@@ -348,6 +348,7 @@ def test_call_pyproxy_destroy_args(selenium):
     selenium.run_js(
         r"""
         let y;
+        pyodide.setDebug(true);
         self.f = function(x){ y = x; }
         pyodide.runPython(`
             from js import f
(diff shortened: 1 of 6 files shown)


src/tests/test_jsproxy.py
```diff
@@ -361,6 +362,23 @@ def test_call_pyproxy_destroy_args(selenium):
...
+        pyodide.setDebug(false);
+        self.f = function(x){ y = x; }
+        pyodide.runPython(`
+            from js import f
+            f({})
+            f([])
+        `);
+        assertThrows(() => y.length, "Error",
+            "This borrowed proxy was automatically destroyed at the end of a function call.*\n" +
+            'For more information about the cause of this error, use `pyodide.setDebug.true.`'
```
ryanking13, line 377:
```suggestion
            'For more information about the cause of this error, use `pyodide.setDebug(true)`'
```

ryanking13 (approved):
LGTM, thanks!

src/tests/test_jsproxy.py
hoodmane: The regex won't match then since it makes a regex group. I think you need `\\(`. But it's annoying to debug these regex...

rth:
I missed most of the discussions in the last 2-3 weeks, and trying to catch up now, but nice work on performance improvement!

Issue #3774, opened 2023-04-15
ModuleNotFoundError: No module named 'Cython' when using pyodide build 


gabrielfougeron:
## 🐛 Bug

When trying out the new pyodide 0.23.1, I get a "ModuleNotFoundError: No module named 'Cython' " error when building.

This is especially surprizing (to me) because cython is a dependency of my project.

### Environment

pyodide_build 0.23.1
emsdk 3.1.32
python 3.11.3

ryanking13:
Could you please show us the pyproject.toml file of your project? Does it specify Cython as a build dependency?

We hard-coded [Cython as a build depencency in <0.23.0, but not anymore](https://github.com/pyodide/pyodide/blob/5ef5622414f0fe8ab8c54694e04ab1b13939b7d4/pyodide-build/pyodide_build/pypabuild.py#L71-L81).

gabrielfougeron:
Worked perfectly, thank you very much!

PR #4028, opened 2023-07-31
Add CoolProp v6.4.3 to packages

nascheinkman:
### Description

This adds the [CoolProp](https://github.com/CoolProp/CoolProp) library to pyodide packages. CoolProp is an open-source thermodynamic and transport properties database. Some development notes I had:

 - Uses sourceforge for source code because it was the only host I found that also packaged all submodules
 - The `pyproject.toml` file I wrote was necessary to get Cython to be recognized. I followed guidance from here: #3774 to figure it out.
 - CoolProp is mainly a C++ project, and the Python wrapper is nested away in a `wrappers/Python` folder. This means the `setup.py` file for the project is not in the root directory of the source as expected. The `setup.py` file I wrote gets put in the root directory and just calls the correct `setup.py` file of the project, changing directories and `sys.path` to meet its filepath expectations.
 - Currently, anything that tries to use REFPROP will cause a CppException that seems to crash Pyodide.

I'm not sure how you would like to handle the CppException. I don't believe there's a way for pyodide to package REFPROP (it's paid [licensed software](https://www.nist.gov/srd/refprop)). On my system install of python, I also get an error, but it doesn't kick me from the REPL prompt. I'm personally ok with leaving it as is, because it doesn't cause issues as long as REFPROP isn't used, but I thought I'd bring it up.

A minimum code sample to bring up the CppException once CoolProp is built is:
```python
import pyodide_js
await pyodide_js.loadPackage('micropip')
import micropip
await micropip.install('coolprop')

from CoolProp.CoolProp import PropsSI
PropsSI('D', 'T', 300, 'P', 101325, 'REFPROP::R32[0.697615]&R125[0.3032385]')
```

Besides for that, everything else seems to be working, but I haven't done rigorous testing.

### Checklists

This is my first PR, so I don't know how much of this stuff which came in the PR template would need to be done. 

- [ ] Add / update tests

### DIFF
--- a/docs/project/changelog.md
+++ b/docs/project/changelog.md
@@ -71,7 +71,8 @@ myst:
 - New packages: sourmash {pr}`3635`, screed {pr}`3635`, bitstring {pr}`3635`,
   deprecation {pr}`3635`, cachetools {pr}`3635`, xyzservices {pr}`3786`,
   simplejson {pr}`3801`, protobuf {pr}`3813`, peewee {pr}`3897`,
-  Cartopy {pr}`3909`, pyshp {pr}`3909`, netCDF4 {pr}`3910`, igraph {pr}`3991`.
+  Cartopy {pr}`3909`, pyshp {pr}`3909`, netCDF4 {pr}`3910`, igraph {pr}`3991`,
+  CoolProp {pr}`4028`.
 - Upgraded libmpfr to 4.2.0 {pr}`3756`.
 - Upgraded scipy to 1.11.1 {pr}`3794`, {pr}`3996`
 - Upgraded scikit-image to 0.21 {pr}`3874`
--- a/packages/coolprop/meta.yaml
+++ b/packages/coolprop/meta.yaml
@@ -23,7 +23,6 @@ requirements:
 test:
   imports:
     - CoolProp
-    - CoolProp.tests
     - CoolProp.GUI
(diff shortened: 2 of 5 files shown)


ryanking13 (commented):
Thanks @nascheinkman!

Could you add tests to make sure that some core functionalities of CoolProp are working? You can add `test_coolprop.py` in the same directory. Check [other packages](https://github.com/pyodide/pyodide/blob/main/packages/sourmash/test_sourmash.py) to find out how you can write a test.

> Uses sourceforge for source code because it was the only host I found that also packaged all submodules

Hmm, I think Sourceforge is not a very stable place to download a file, but if it is the only host we can find, then I think it is okay. If there is an issue later, then we can try hosting it from somewhere else.

ryanking13 (approved):

dvd101x:
Hi, this is super valuable to me! I'm excited this effort is being made to include CoolProp!

ibell:
I think it would be wise to use the GitHub release zip: https://github.com/CoolProp/CoolProp/releases/tag/v6.5.0

ibell:
Or this one from pypi is even better: https://pypi.org/project/CoolProp/6.5.0/#files

rth:
@ibell feel free to open a PR to change the URL. Agreed that those sound better.

PR #4033, opened 2023-08-02
ENH Improve PyProxy handling of read only properties

hoodmane:
A descriptor is *writable* if either writable is true or it has a setter. A descriptor is *deletable* if it is configurable. Also, the normal JS behavior here is to return false, not to throw.

- [x] Add / update tests


hoodmane:
we should probably run our entire test suite in strict mode...

PR #4035, opened 2023-08-03
ENH Rework streams handling

hoodmane:
This fixes a number problems with the old stream handling:

1. Not possible to set a custom errno (necessary for proper interrupt handling and possibly for other things)

2. Inefficient: in a lot of cases we have data in one buffer and we need it placed into a different buffer, but we have to implement a function that gets one byte out of the source buffer and then call it repeatedly to move one byte at a time to the target buffer.

3. Ease of implementation: in many cases we already have perfectly good buffer manipulation APIs, so if we have direct access to the true source or target buffer we can just use these. See: the node IO code, which got much simpler.

This is backwards compatible, so you can still use the old input mechanism or use buffered or raw output. But it adds a new method of directly implementing read/write. For simplicity, we insure that the source/destination buffers are always Uint8Array views that point to exactly the region that is meant to be read/written. Because TypedArrays have support for views, I see no reason APIs should pass around a larger buffer with auxilary offset/length parameters.

If you do use the old mechanisms, they are still faster and can correctly support keyboard interrupts. Other than that I think the original behavior is unchanged. I added a lot more test coverage to help ensure that, since the streams tests I wrote before had pretty anemic coverage.

I think the read/write APIs are mostly pretty simple to use, with the exception that someone might forget to return the number of bytes read. JavaScript's ordinary behavior coerces the `undefined` to a 0, which leads to an infinite loop where the filesystem repeatedly asks to read/write the same data since it sees no progress. I added a check that writes an error message to the console and sets EIO when undefined is returned so the infinite loop is prevented and the problem is explained.

- [x] Add / update tests


Makefile
```diff
@@ -134,13 +134,19 @@ dist/package.json : src/js/package.json
...
+	# See https://github.com/timocov/dts-bundle-generator/issues/255
+	echo "export declare const defines: {[k:string] : number};" > src/js/generated_struct_info32.gen.json.d.ts
+	cd src/js && npx dts-bundle-generator {pyodide,ffi}.ts --export-referenced-types false --project tsconfig.json
 	mv src/js/{pyodide,ffi}.d.ts dist
 	python3 tools/fixup-type-definitions.py dist/pyodide.d.ts
 	python3 tools/fixup-type-definitions.py dist/ffi.d.ts
 
 
+src/js/generated_struct_info32.gen.json : emsdk/emsdk/.complete
+	cp emsdk/emsdk/upstream/emscripten/src/generated_struct_info32.json $@
```
ryanking13, line 148: How about embedding this JSON file (or maybe a TS-converted version of it) into `/src/js` and updating it when we update emscripten version (we will definitely need to automate it).

Looking at this PR, I think things getting a bit complex just to handle this file. I am even okay with hardcoding constants that we use if it doesn't change often.

hoodmane: Fine with me. I'd rather use this file, since if they do change it would be pretty hard to figure out why things were breaking.

hoodmane: Okay, I checked in this file and added a check to `emsdk/Makefile` to check that the files match after installing emscripten.

docs/usage/streams.md
```diff
@@ -7,36 +7,351 @@ Pyodide has three functions {js:func}`pyodide.setStdin`,
...
+different ways to do this depending on the options passed to `setStdin`.
+
+### Always raise IO Error
+
+If we pass `{error: true}`, any read from stdin raises an I/O error.
+
+```js
+pyodide.setStdin({ error: true });
+pyodide.runPython(`
+    with pytest.raises(OsError, match="I/O error"):
```
ryanking13, line 24:
```suggestion
    with pytest.raises(OSError, match="I/O error"):
```

docs/usage/streams.md
```diff
@@ -7,36 +7,351 @@ Pyodide has three functions {js:func}`pyodide.setStdin`,
...
+
+As with {js:func}`pyodide.setStdin`, there are quite a few different ways to set
+the standard output handlers.
+
+### Set the default behavior
+
+As with stdin, `pyodide.setStdout()` sets the default behavior. In node, this is
+to write directly to `process.stdout`. In the browser, the default is as if you
+wrote
+`setStdout({batched: (str) => console.log(str)}})`
```
ryanking13, line 265:
```suggestion
`setStdout({batched: (str) => console.log(str)})`
```

docs/usage/streams.md
ryanking13, line 1: Thanks for writing detailed docs 👍

src/js/streams.ts
```diff
@@ -295,201 +477,244 @@ export function setStdout(
...
- *   pyodide.setStderr({ raw: (byte) => console.warn(byte) });
- *   pyodide.runPython("import sys; print('ABC', file=sys.stderr)");
- *   // 65
- *   // 66
- *   // 67
- *   // 10 (the ascii values for 'ABC' including a new line character)
- * }
- * main();
+ * Sets the standard error handler. See the documentation for
+ * {js:func}`pyodide.setStdout`.
```
ryanking13, line 489:
```suggestion
 * :js:func:`pyodide.setStdout`.
```

src/js/streams.ts
```diff
@@ -23,78 +47,165 @@ export type InFuncType = () =>
...
-
 /**
  * We call refreshStreams at the end of every update method, but refreshStreams
  * won't work until initializeStreams is called. So when INITIALIZED is false,
  * refreshStreams is a no-op.
  * @private
  */
 let INITIALIZED = false;
+const DEVOPS: { [k: number]: Reader & Writer } = {};
+const DEVS = {} as {
```
ryanking13, line 58: Could you add a comment explaining where the values for `DEVS` is set?

src/js/streams.ts
```diff
@@ -23,78 +47,165 @@ export type InFuncType = () =>
...
+}
+
+function isErrnoError(e: any) {
+  return e && typeof e === "object" && "errno" in e;
+}
+
+const waitBuffer = new Int32Array(
+  new WebAssembly.Memory({ shared: true, initial: 1, maximum: 1 }).buffer,
+);
+function syncSleep(ms: number): boolean {
```
ryanking13, line 84: Maybe let's call the parameter name `timeout`.

src/js/streams.ts
```diff
@@ -295,201 +477,244 @@ export function setStdout(
   options: {
     batched?: (a: string) => void;
     raw?: (a: number) => void;
+    write?: (a: Uint8Array) => number;
```
ryanking13, line 480: These are rendered in the docs, so maybe use other than `a`?

src/tests/test_streams.py
```diff
@@ -0,0 +1,579 @@
...
+        "This has a \x00 null byte in the middle...",
+        "several\nlines\noftext",
+        "pyodidé",
+        "碘化物",
+        "🐍",
+    ]
+    outstrings: list[str] = sum(
+        ((s.removesuffix("\n") + "\n").splitlines(keepends=True) for s in strings), []
+    )
+    print(outstrings)
```
ryanking13, line 23:
```suggestion
```

If this was for debugging.

ryanking13 (approved):
Thanks, @hoodmane and sorry for the late review.

I have a few minor comments, otherwise LGTM.

hoodmane:
Thanks for the detailed review @ryanking13!

Makefile
```diff
@@ -135,13 +135,13 @@ npm-link: dist/package.json
 	cd src/test-js && npm ci && npm link ../../dist
 
 dist/pyodide.d.ts dist/pyodide/ffi.d.ts: src/js/*.ts src/js/pyproxy.gen.ts src/js/error_handling.gen.ts
-	npx dts-bundle-generator src/js/{pyodide,ffi}.ts --export-referenced-types false --project src/js/tsconfig.json
+	node src/js/esbuild.config.mjs
+	# See https://github.com/timocov/dts-bundle-generator/issues/255
+	cd src/js && npx dts-bundle-generator {pyodide,ffi}.ts --export-referenced-types false --project tsconfig.json
```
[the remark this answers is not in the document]
hoodmane: No we can revert this whole file.

docs/conf.py
```diff
@@ -259,6 +259,13 @@ def ensure_typedoc_on_path():
     )
 
 
+def get_emscripten_version():
```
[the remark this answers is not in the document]
hoodmane: Nope.

docs/usage/streams.md
```diff
@@ -7,36 +7,351 @@ Pyodide has three functions {js:func}`pyodide.setStdin`,
...
+        input()
+`);
+```
+
+### Set the default behavior
+
+You can set the default behavior by calling `pyodide.setStdin()` with no
+arguments. In Node the default behavior is to read directly from Node's standard
+input. In the browser, the default is the same as
+`pyodide.setStdin({ error: true })`.
```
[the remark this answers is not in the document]
hoodmane: Good point, I should change that.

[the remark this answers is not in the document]
hoodmane: To be clear this is a change in behavior, but it's probably a fix:
```js
const pyodide = loadPyodide();
pyodide.runPython(`
input("hi") # opens a prompt
`);
pyodide.setStdin()
pyodide.runPython(`
input("hi") # raises IOError
`);
```

Issue #1282, opened 2021-02-24
Build orjson for pyodide


jmsmdy:
The project [orjson](https://github.com/ijl/orjson) offers significantly faster JSON encoding of objects than other Python packages (40x faster than the json package in many instances), with direct encoding of numpy arrays built in. 

It would be nice if orjson could be built for pyodide. It is written in Rust, so that may be an issue.

dalcde:
The Rust-C interop story when compiling to webassembly isn't very good.
I've looked into this a few months ago and it go too far; iirc they use
a different C ABI.

spicy-sauce:
What's the status with this one? Not sure if it was handled or not as I'm getting:
ValueError: Can't find a pure Python 3 wheel for 'orjson<4.0.0,>=3.8.7'.

Thanks

hoodmane:
It's probably straight forward. I'll look into it.

What's the difference between orjson and ujson? As I understand they are both rust json libs?

spicy-sauce:
Yes, both orjson and ujson are Rust libraries for working with JSON data, but they have some differences in terms of their features, performance, and use cases.

**Features** 
orjson is a fast JSON serialization library that focuses on providing the best performance possible, but it does not provide all the features that some other JSON libraries offer, such as support for custom serialization or deserialization behavior. On the other hand, ujson is a more feature-rich JSON library that provides options for customizing serialization and deserialization behavior, including support for custom data types, but it may not be as optimized for performance as orjson.

**Performance**
orjson is known for its extremely fast serialization and deserialization performance, making it one of the fastest JSON libraries available in Rust. It uses a combination of optimized algorithms and low-level Rust optimizations to achieve its performance. ujson also offers good performance, but it may not be as fast as orjson in some cases.

**Use cases**
orjson is primarily designed for use cases where high-performance JSON serialization and deserialization are critical, such as in performance-sensitive applications or systems that require processing large amounts of JSON data with minimal overhead. ujson, on the other hand, may be more suitable for use cases that require a balance between performance and features, and where customization of serialization and deserialization behavior is desired.

Also see:
https://pythonspeed.com/articles/faster-json-library/

hoodmane:
Well it builds first try no problem.

spicy-sauce:
Does it mean that you should pack it and place under packages?

spicy-sauce:
Can you please share the current status of this one? Thanks!

spicy-sauce:
[...]

PR #4036, opened 2023-08-03
orjson package

spicy-sauce:
### Description

This PR adds `orjson` package support, see more details in #1282.

### Checklists

This is my first PR, so if I missed something or did it wrong, please let me know and I'll update accordingly. 

Let me know if I should add the orjson and where, wasn't sure if it should be listed under `Packages` section at this point. 
- [x] Add / update tests - per @hoodmane's request, I added `test_orjson.py`.
Let me know if this should be updated and where.


hoodmane:
Uh oh it fails to load with "cannot resolve symbol yyjson_alc_pool_init"...

spicy-sauce:
Seems like an issue build issue, right? Will you delve into it? Thanks.

ryanking13:
I looked into this a bit, and it seems that the problem is caused by the `-Z link-native-libraries=no` option.

orjson needs to link a native library (libyyjson.a), which is written in c, but because of the `-Z link-native-libraries=no` option, it won't link to it.

The Pyodide part is solved by resetting RUSTFLAGS with `export RUSTFLAGS=""`, but maturin sets the same option, so it should be overridden too.

@messense Hi, looking at the maturin code, it seems that [a user can overwrite the link-native-libraries option](https://github.com/PyO3/maturin/blob/66a1511f9c22e3549fb1caa580ae213864560225/src/compile.rs#L241-L244), but I wasn't able to do it. I tried 1) setting the RUSTFLAGS env variable and 2) adding the following config to .cargo/config.toml

```
[target.wasm32-unknown-emscripten]
rustflags = [
    "-Zlink-native-librarie=no",
]
```

but both didn't work. What is the correct way of setting the flag?

ryanking13:
@hoodmane by the way, maybe we don't need to set RUSTFLAGS anymore if maturin sets the same flags?

messense:
@ryanking13 I think that check in maturin is a bit wrong: https://github.com/PyO3/maturin/pull/1724

ryanking13:
@messense Thanks for the prompt fix! I'll try this again when a new version of maturin is released.

spicy-sauce:
> @messense Thanks for the prompt fix! I'll try this again when a new version of maturin is released.

They released a new version 3 weeks ago, is it relevant?
https://github.com/PyO3/maturin/releases/tag/v1.2.3

spicy-sauce:
> @spicy-sauce
> 
> Yes, thanks for the ping.
> 
> It seems like adding the following in `orjson` meta.yaml works now.
> 
> ```yaml
> build:
>   script:
>     export RUSTFLAGS="-Z link-native-libraries=yes"
> requirements:
>   executable:
>     - rustup
> ```
> 
> Could you add this to the meta.yaml + sync upstream/main branch of Pyodide in this PR?

done

hoodmane:
Thanks @spicy-sauce, @ryanking13, and @messense!

spicy-sauce:
Any plans to release a new version with this anytime soon? 
Thanks.

rth:
This PR was included in the 0.24.0 release.

PR #4044, opened 2023-08-04
Adding pymongo package for MongoDB database manipulation

SorinAlexB:
### Description

Adding pymongo module for MongoDB database.


ryanking13:
Thanks, @SorinAlexB.

Pyodide does not support sockets because of browser limitations, so I'm pessimistic that pymongo and other DB client packages will work. How do you plan to use pymongo in Pyodide?

SorinAlexB:
I wanted to make an browser application using a mongo database for registration and login system. I wanted to use mongoDB because I find it easier to use than mysql.

ryanking13:
mysql and pymongo both use socket to access to remote database, so unless you are thinking of using an in memory database, I think it won't be very easy. See https://github.com/pyodide/pyodide/issues/3818 for Pyodide limitations about sockets.

SorinAlexB:
Thanks!

PR #4052, opened 2023-08-08
ENH Add `response.text` to `pyodide.http.FetchResponse`

hoodmane:
We've received feedback from users that use other requests APIs that they expect the method to be called `response.text` instead of `response.string`. Indeed both the Fetch response API and the Python requests library use this convention:

https://developer.mozilla.org/en-US/docs/Web/API/Response/text https://requests.readthedocs.io/en/latest/api/#requests.Response.text

This adds `response.text` to `FetchResponse`. It is a synonym for `response.string`. 

We could also deprecate `response.string` but I don't see any immediate need to do so.

cc @pelson who suggested this change.

- [ ] Add / update tests

### DIFF
--- a/src/py/pyodide/http.py
+++ b/src/py/pyodide/http.py
@@ -160,7 +160,18 @@ class FetchResponse:
         return await self.js_response.arrayBuffer()
 
     async def string(self) -> str:
-        """Return the response body as a string"""
+        """Return the response body as a string
+
+        Does the same thing as :py:meth:`Response.text`.
+        """
+        self._raise_if_failed()
+        return await self.js_response.text()
+
+    async def text(self) -> str:
+        """Return the response body as a string
+
+        Does the same thing as :py:meth:`Response.string`.
+        """
         self._raise_if_failed()
         return await self.js_response.text()
 


src/py/pyodide/http.py
```diff
@@ -160,7 +160,18 @@ async def buffer(self) -> JsBuffer:
         return await self.js_response.arrayBuffer()
 
     async def string(self) -> str:
-        """Return the response body as a string"""
+        """Return the response body as a string
+
+        Does the same thing as :py:meth:`Response.text`.
+        """
+        self._raise_if_failed()
+        return await self.js_response.text()
```
ryanking13, line 168: Maybe let's just do `return self.text()`?

ryanking13 (approved):

> We could also deprecate response.string but I don't see any immediate need to do so.

Yes, I think we can just keep it for backward compatibility.

rth:
> We could also deprecate response.string but I don't see any immediate need to do so.

Can we at least mention in the docstring which one should be used, and mention that the other one is deprecated in the docstring? Having two methods that do exactly the same thing is not very pythonic. But yeah I agree that there is no need to actually raise a deprecation warning.

ryanking13:
> 1. we tell people not to use it

I think option 1 with mentioning not to use it in the docstring is okay for this case, as the extra maintenance burden is negligible.

PR #4059, opened 2023-08-10
PERF Don't repeat conversion of destroyed messages

hoodmane:
This converts `destroy_proxy` and `destroy_proxies` to take static js strings so that we don't repeatedly convert them.


ryanking13 (approved):
Thanks for your efforts to improve performance!

PR #4066, opened 2023-08-14
TST Use oneshot handlers for pytest-httpserver

ryanking13:
This is an attempt to fix occasional GHA - `test_pyfetch_cors_error` failures on macos.

This PR fixes two potential problems that I can think of:

1) use one-shot handler so that `pytest-httpserver` don't reuse the handler.
2) use a unique endpoint for each test to remove the potential response caching in playwright.


hoodmane (approved):

hoodmane:
It looks reasonable to me, but tests seem to failing.

ryanking13:
Ok CI is now green (haven't seen that for a long time 😂)

PR #4072, opened 2023-08-17
PERF Add a hiwire stack and use it in python2js

hoodmane:
Prior to this PR, an even JSRef was "immortal" and odd JsRefs were normal. This adds a third type of JsRef called "stack references". These are references with `id & 3 == 2`, if `id & 3 == 0`, then it is an immortal reference and all odd references are ordinary "heap references".

In addition to `Hiwire.new_value` we have `Hiwire.new_stack` which allocates a stack reference. The APIs for dealing with stack and heap references are otherwise unchanged. These APIs have been updated as follows:

* `Hiwire.new_stack`: push argument onto stack and return `(index << 2) | 2`.
* `Hiwire.incref`: on a stack reference, make a new heap reference with same value and return it.
* `Hiwire.decref`: on a stack reference, check that it points to top of stack, if not raise fatal error. If so, pop it from stack.
* `Hiwire.get_value`: If it's a stack reference, then look it up on the stack.

I also moved immortal values to an immortals array for faster lookup. This improves the speed of the following profile by a factor of 3:

```py
jsf = run_js("() => {}")
timeit("jsf(1, 1, 1, 1, 1, 1)", number=n, globals={"jsf": jsf}) / n * 1e6
```
It also seems to take about 15% off the execution time of `to_js` for a large struct.

### DIFF
--- a/src/core/hiwire.c
+++ b/src/core/hiwire.c
@@ -32,13 +32,22 @@ HIWIRE_INIT_CONSTS();
   _hiwire.obj_to_key.set(js_value, Hiwire.hiwire_attr);
 
 // clang-format off
+
 // JsRefs are:
 // * ordinary if they are odd,
 // * immortal if they are divisible by 4
 // * stack references if they are congruent to 2 mod 4
+//
 // Note that "NULL" is immortal which is important.
+//
+// Both immortal and stack indexes are converted to id by bitshifting right by
+// two to remove the lower order bits which indicate the reference type.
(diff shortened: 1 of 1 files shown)


ryanking13 (approved):
Thanks, @hoodmane! It generally looks okay to me.

Would you mind defining bit checks and operations as a macro?

- (idval & 3) === 0
- (idval & 3) === 2
- (idval >> 2)
- ...

It will be very hard to understand these codes without reading the description in this PR.

PR #4078, opened 2023-08-19
Docs: add reference to removed std lib pkgs

nmstoker:
### Description

Several Python Standard Library packages that were also removed for common sense reasons are not available but were missing from the documents.

This PR only impacts documentation.

### Checklists


hoodmane:
Thanks!

PR #4079, opened 2023-08-20
Add comment that WSL 2 is needed on Windows

Vuizur:
### Description
I think the tutorial under https://pyodide.org/en/stable/development/new-packages.html currently is not compatible with Windows (The build-recipes command calls make and has a Makefile.envs that contains Linux-specific code like `export HOSTPYTHONROOT=$(shell python${PYMAJOR}.${PYMINOR} -c "import sys; print(sys.prefix)"))`, so we should mention it in the docs. Please correct me if I am wrong.

### DIFF
--- a/docs/development/new-packages.md
+++ b/docs/development/new-packages.md
@@ -33,7 +33,9 @@ is a Python package without C extensions):
   the wheel.
 
 If however the package has C extensions or its code requires patching, then
-continue to the next steps.
+continue to the next steps. 
+
+If you are on Windows, you will need to use WSL 2.
 
 ```{note}
 To determine if a package has C extensions, check if its `setup.py` contains


Vuizur:
I think using the docker image might be obligatory as well. When not using it, I get:
```
Creating meta.yaml package for spacy                                                                                                                                                                       
'\\wsl.localhost\Ubuntu-22.04\home\pux\programs\pyodide'
CMD.EXE was started with the above path as the current directory.
UNC paths are not supported.  Defaulting to Windows directory.
```
I might be doing something wrong though, so no idea.

ryanking13:
Thanks, @Vuizur!

We use `npx` to prettify the meta.yaml file after creating it, and it looks like the error happens when the npx (node, npm) is installed in the Windows file system.

PR #4088, opened 2023-08-23
Update packages for 0.24.0

ryanking13:


hoodmane:
It seems like this is almost ready, but for some reason there are still geopandas failures, even though it seems like you updated this?

ryanking13:
> It seems like this is almost ready, but for some reason there are still geopandas failures, even though it seems like you updated this?

I reverted pytest version update and it seems like the error is gone. I am not sure why 🤔 Anyway, I think this is ready.

hoodmane:
> I reverted pytest version update and it seems like the error is gone. I am not sure why 🤔

Spooky...

PR #4091, opened 2023-08-24
PERF Preload wasm to improve load time

hoodmane:
This uses `Module.instantiateWasm` to fetch `pyodide.asm.wasm` at the same time as `pyodide.asm.js`.

### DIFF
--- a/src/js/module.ts
+++ b/src/js/module.ts
@@ -96,6 +96,13 @@ export interface Module {
   removeRunDependency: (id: string) => void;
   reportUndefinedSymbols: () => void;
   ERRNO_CODES: { [k: string]: number };
+  instantiateWasm?: (
+    imports: { [key: string]: any },
+    successCallback: (
+      instance: WebAssembly.Instance,
+      module: WebAssembly.Module,
+    ) => void,
+  ) => void;
 }
 
 /**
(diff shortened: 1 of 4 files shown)


rth (approved):
Thanks, nice optimisation! So indeed starting to load `pyodide.wasm` at the same time as `pyodide.asm.js` saves almost 300ms when loading the artifacts from Circle CI. Btw I rediscovered that clicking on [`console.html`](https://output.circle-artifacts.com/output/job/d98bbf4a-891e-4adc-a2dd-919ba9db677f/artifacts/0/root/repo/dist/console.html) in CircleCI artifacts just works now.

Here is the profiling in Chrome on main,
image

and this PR,
image

ryanking13:
> Btw I rediscovered that clicking on [console.html](https://output.circle-artifacts.com/output/job/d98bbf4a-891e-4adc-a2dd-919ba9db677f/artifacts/0/root/repo/dist/console.html) in CircleCI artifacts just works now.

Wow I didn't know that :)

rth:
Yeah, I'm pretty sure it wasn't working before either due to CORS issue or something else but now it does and it's really nice :)

src/js/module.ts
```diff
@@ -219,3 +226,49 @@ export function initializeFileSystem(Module: Module, config: ConfigType) {
...
+  if (SOURCEMAP) {
+    // According to the docs:
+    //
+    // "Sanitizers or source map is currently not supported if overriding
+    // WebAssembly instantiation with Module.instantiateWasm."
+    // https://emscripten.org/docs/api_reference/module.html?highlight=instantiatewasm#Module.instantiateWasm
+    return;
+  }
+  const { binary, response } = getBinaryResponse(
+    indexURL! + "pyodide.asm.wasm",
```
ryanking13, line 240: nitpick: is `!` needed? `indexURL` is always a string.

ryanking13 (approved):

src/js/module.ts
hoodmane: Looks like you're right. It is used like this in a couple of places.

PR #4093, opened 2023-08-26
Update Pillow version to 10.0.0 and fix its wheel filename

ryanking13:
### Description

This PR upgrades Pillow's version to 10.0.0 and removes the behavior that previously overwrote setup.cfg. This fixes an issue that  a wheel file of Pillow to be named incorrectly (`PIL-...`)

- Related: #4005


hoodmane:
Thanks!

PR #4094, opened 2023-08-26
Emscripten 3.1.45

hoodmane:
- [x] Update changelog


hoodmane:
Building numcodecs failed:
```
clang: error: unsupported option '-mno-sse2' for target 'wasm32-unknown-emscripten'
clang: error: unsupported option '-mno-avx2' for target 'wasm32-unknown-emscripten'
```

ryanking13:
It looks like previously llvm ignored those SIMD flags, but now it throws an error. In the build log of the main branch:

```
clang: warning: argument unused during compilation: '-mno-sse2' [-Wunused-command-line-argument]
clang: warning: argument unused during compilation: '-mno-avx2' [-Wunused-command-line-argument]
```

Those flags are added by [`DISABLE_NUMCODECS_SSE2`, `DISABLE_NUMCODECS_AVX2`](https://github.com/pyodide/pyodide/blob/main/packages/numcodecs/meta.yaml#L14-L15) env variables. So let's remove them and see what happens.

hoodmane:
That fails too:
```
emcc: error: Passing any of -msse, -msse2, -msse3, -mssse3, -msse4.1, -msse4.2, 
-msse4, -mavx, -mfpu=neon flags also requires passing -msimd128 (or             
-mrelaxed-simd)! 
```
How about filtering `-mno-sse2` in pywasmcross?

ryanking13:
Yes, it looks like we need to filter out SIMD flags in pywasmcross, until we support SIMD.

> How about filtering -mno-sse2 in pywasmcross?

`-mno-sse2` and `-mno-avx` are removed, but other flags remain, so I guess we should filter out `-msse, -msse2, -msse3, -mssse3, -msse4.1, -msse4.2, -msse4, -mavx`.

ryanking13:
Yep... I'll start working on #3827 again after the release so we don't have to build whole packages on every commit.

PR #4096, opened 2023-08-27
PERF Avoid instance checks in pyproxy_getflags when possible

hoodmane:
Profiling shows that `PyObject_IsInstance` is pretty expensive. This skips it when conditions we've already measured imply that they will return false anyways.

This gives a 5x performance improvement on the following:
```js
pyodide.runPython(`
from pyodide.code import run_js
from timeit import timeit
f = run_js("() => {}")

d = {}

n = 200000
timeit("f(d)", number=n, globals={"f": f, "d": d}) / n * 1e6
`);
```

If `d` is a function, it gives a 1.7x speedup, if it's a list iterator, a 2x speedup. So it depends on what the object is but in most cases it's a lot faster than before.


src/core/pyproxy.c
```diff
@@ -209,18 +204,24 @@ pyproxy_getflags(PyObject* pyobj)
   // YIELD_FROM. GET_AWAITABLE uses _PyCoro_GetAwaitableIter defined in
   // genobject.c. This tests whether _PyCoro_GetAwaitableIter is likely to
   // succeed.
-  SET_FLAG_IF(IS_AWAITABLE, async_proto->am_await || gen_is_coroutine(pyobj));
+  //  || gen_is_coroutine(pyobj)
```
ryanking13, line 207: What's this?

ryanking13 (approved):
Very nice. Thanks!

src/core/pyproxy.c
hoodmane: Oh I left it here as a note to stick `gen_is_coroutine` outside of this function -- this function now only works on the type, but `gen_is_coroutine` needs to look at `generator->gen_code->co_flags` to see whether `CO_COROUTINE` is set, so it depends on the object itself not just the type of the object. I moved this down to then end of `pyproxy_getflags`.

ryanking13: Thanks for the explanation. It would be good to add a comment about it.

PR #4098, opened 2023-08-28
DOCS minor changes to quickstart

hoodmane:
* Removed numpy from `Accessing Python scope from JavaScript` example since it's irrelevant
* added forward reference from top of page to complete html example to help prevent people from writing their own before discovering it


ryanking13 (approved):
Thanks!

PR #4101, opened 2023-08-28
feat: add awkward-cpp v22

agoose77:
### Description

Add the [awkward-cpp](https://github.com/scikit-hep/awkward/tree/main/awkward-cpp) package.

### Checklists

- [x] Add / update tests


packages/awkward-cpp/meta.yaml
```diff
@@ -0,0 +1,27 @@
+package:
+  name: awkward-cpp
+  version: 22
+  top-level:
+    - awkward_cpp
+
+source:
+  url: https://pypi.io/packages/source/a/awkward-cpp/awkward-cpp-22.tar.gz
```
ryanking13, line 8:
```suggestion
  url: https://files.pythonhosted.org/packages/8b/f8/97efcc2d52e2f9967f9beed86ed6736fc6eed91eb2d974a1bdc8c438146d/awkward-cpp-22.tar.gz
```

Let's use `https://files.pythonhosted.org` url to prevent redundant redirections.

packages/awkward-cpp/meta.yaml
```diff
@@ -0,0 +1,27 @@
...
+  sha256: 21679636fb21cfe3715f88a32326a579199384db2da4a62995c310502d7fe85f
+
+build:
+  script: |
+    export CMAKE_ARGS="${CMAKE_ARGS} -DEMSCRIPTEN=1"
+  exports: whole_archive
```
ryanking13, line 14: What happens if you don't set `whole_archive`?

ryanking13 (commented):
Thanks @agoose77! Could you add some basic tests for awkward-cpp?

packages/awkward-cpp/meta.yaml
agoose77: It used to be the case that some symbols were missing. Right now I can't seem to test this locally; I'm getting 
```
OSError: [Errno 7] Argument list too long: 'em++'
```

I don't recall this happening before.

hoodmane: I think it's okay for now then then, we have a bug in our symbol logic that causes this `Argument list too long`  error when there are a lot of exported symbols.

hoodmane (approved):

PR #4104, opened 2023-08-29
Update geopandas to 0.13.2

hoodmane:
Geopandas 0.13 unblocks pandas 2.0, cf https://github.com/pyodide/pyodide/issues/3734#issuecomment-1695941255


ryanking13:
I think I failed to update geopandas in #4088, but I didn't look at the error messages in detail.

hoodmane:
I think shapely 2.0 is a known troublesome upgrade.

ryanking13:
> I think shapely 2.0 is a known troublesome upgrade.

Right. I recently tried and failed because of some symbol issues in geos.

rth:
But so CI is passing here (aside from MacOS which fails for unrelated reasons) unless I'm missing something? Let me sync with upstream to see if it's still OK.

rth:
The only failed package test is bokeh on firefox. There are also safari CI failures, but those sound unrelated (also broken on main). So it should be OK to merge this, no?

ryanking13:
Oh, indeed, it is passing CI. Based on my previous comment, it looks like I was getting an error... but I don't remember. Then let's merge and see what happens.

PR #4106, opened 2023-08-30
CI Split package build steps as a temporary solution for CI timeout

ryanking13:
Adds `build-libraries` job which runs before `build-no-numpy-dependents` which builds libraries first. This is obviously not a long term solution, but works for now.


hoodmane:
Thanks @ryanking13. Indeed this isn't ideal but at least it solves the short term problem.

PR #4125, opened 2023-09-09
Update numpy to 1.25.2

rgommers:
### Description

This is the last version of `numpy` that defaults to a distutils/setup.py-based build. I am also looking into moving the build to Meson (xref https://github.com/numpy/numpy/pull/24603), but I thought it'd be good to make this update in case anything changed in the 1.25.x release series.

### Checklists

- [ ] Add / update tests


hoodmane:
Thanks @rgommers! It seems to be mad about `-std=c++17`? I'll try building it locally.

rgommers:
Argh, I was hoping it would be smooth sailing. I tried building locally in a conda env following the contributor guide, but that had some hiccup. I will try with the recommended Docker setup instead.

And yes indeed, more C++ in this release. This snippet from the release notes may also be relevant:
```
C++ standard library usage
--------------------------
NumPy builds now depend on the C++ standard library, because
the ``numpy.core._multiarray_umath`` extension is linked with
the C++ linker.
```

C++17 is no problem in principle I assume, right?

hoodmane:
Are there any simple tricks to leave tmp files in the file system after compilation fails? I always have this problem where the failing command takes arguments that were in `/tmp` and now they are gone...

hoodmane:
Ah but this test is hard coded into `setup.py` so it's easy to create my own.

rgommers:
Yes, that is usually a pain. If you build via `pip install`, then `--no-clean` is helpful. Or otherwise yes, peek in the source of `test.c` and it's typically only a one-liner.

hoodmane:
Okay found the problem.

hoodmane:
It built really fast, `1m 31s`, tests pass. Did you change something to improve the build speed?

rgommers:
Not that I know of. Locally 1m 31s isn't unusual for a local parallel build though. I just tried on Linux:

```
$ time python setup.py build_ext -i -j4
real    1m21,634s
user    3m36,304s
sys     0m13,231s
```

It'll get a lot faster still from 1.26.0 onwards, due to the Meson rewrite:
```
$ time python -m build -wnx
real    0m40,042s
user    7m1,368s
sys     0m22,224s
```

rgommers:
The macOS job failure looks unrelated. Something that picked up the system Python 2.7 in the final `check-size` step.

hoodmane:
Yeah the github macos job flakes a lot so I tend not to pay much attention to it especially if the circleci safari tests pass.

rgommers:
I can reproduce the initial failure and the completed build after your fix locally now. Thanks for the help!

hoodmane:
Thanks again @rgommers!

PR #4128, opened 2023-09-09
Use hoodmane/hiwire library instead of our own implementation

hoodmane:
This change switches to my external implementation of hiwire. This is the minimal change set to do this, it uses some hacks to avoid changing any files outside of `hiwire.{c,h,js}`. In followups, I will gradually switch to using the new APIs rather than compatibility shims.


hoodmane:
@rth @ryanking13 can somebody review this?

docs/project/changelog.md
```diff
@@ -22,6 +22,9 @@ myst:
 - {{ Fix }} Fixed `LONG_BIT definition appears wrong for platform` error happened in out-of-tree build.
   {pr}`4136`
 
+- {{ Enhancement }} Unvendored hiwire
```
ryanking13, line 25: Maybe this is too vague. It is not a user-facing change so maybe we need to

1) remove it from the change log at all
2) or add more detailed information what happend

cpython/Makefile
```diff
@@ -78,6 +82,20 @@ $(INSTALL)/lib/libffi.a :
...
+	(\
+		cd $(HIWIREBUILD) \
+		&& git init \
+		&& git fetch --depth 1 $(HIWIREREPO) $(HIWIRE_COMMIT) \
+		&& git checkout FETCH_HEAD \
+		&& . $(PYODIDE_ROOT)/emsdk/emsdk/emsdk_env.sh \
+		&& CC=emcc EMSCRIPTEN_DEDUPLICATE=1 EXTERN_FAIL=1 make \
+	)
+	cp -r $(HIWIREBUILD)/dist/lib $(INSTALL)/
+	cp -r $(HIWIREBUILD)/dist/include $(INSTALL)/include/hiwire
```
ryanking13, line 97: nitpick: It would be nice if libhiwire have a `make install` subcommand (No actions required for this PR).

Makefile.envs
```diff
@@ -165,7 +166,9 @@ export MAIN_MODULE_CFLAGS= $(CFLAGS_BASE) \
 	-Werror=int-conversion \
 	-Werror=incompatible-pointer-types \
 	-Werror=unused-result \
+	-mreference-types \
 	-I$(PYTHONINCLUDE) \
+	-I$(PYTHONINCLUDE)/.. \
```
ryanking13, line 171: Maybe we need to find a better place to install header files later... (No action required for this PR)

src/core/hiwire.c
```diff
@@ -6,144 +6,147 @@
...
-#define _NEW_INFO(version, field_and_flag) \
-  (((version) & VERSION_MASK) | (field_and_flag))
-
-// make a new reference with the same version as info and the given index.
-#define HEAP_NEW_REF(index, info) _NEW_INFO(info, ((index) << 1) | 1)
-// new occupied info: same version as argument info, NEW_INFO_FLAGS says occupied with refcount 1
-#define HEAP_NEW_OCCUPIED_INFO(info) _NEW_INFO(info, NEW_INFO_FLAGS)
-// new unoccupied info, increment version and nextfree in the field
-#define FREE_LIST_INFO(info, nextfree) _NEW_INFO(_NEXT_VERSION(info), (nextfree) << 1)
+EM_JS(void, hiwire_invalid_id, (int type, JsRef ref), {
```
ryanking13, line 147: What is this function for? Debugging?

ryanking13 (approved):
Thanks for the huge work @hoodmane! It looks good to me with trust.

docs/project/changelog.md
hoodmane: Agreed, maybe just remove it... maybe we should have a change log tag like `architecture` for non user facing stuff like this but I'm okay leaving it out.

emsdk/patches/0001-Support-externref-in-EM_JS-functions-with-dynamic-li.patch
hoodmane, line 1: This was been merged into emscripten yesterday.

src/core/hiwire.c
hoodmane: Yeah it's called if someone dereferences an id that is malformed, is freed, or hasn't been assigned yet. As a to-do I should add back in the much more detailed internal error handling we had before...

cpython/Makefile
hoodmane: I didn't do that because I don't really know how to Makefile but it probably isn't very hard.

Makefile.envs
hoodmane: I think we can remove this once we get rid of `src/core/hiwire.h` which shadows the real `hiwire.h`. For this pr I tried not to change files outside of `hiwire.*`.

docs/project/changelog.md
hoodmane: Removing this for now, we can revisit if these changes deserve a changelog entry later.

src/core/hiwire.c
hoodmane: Okay I added extra info here and a test.

rth:
Thanks for working on this @hoodmane ! Sound good to me as well.

hoodmane:
It seems to have broken safari though so we may have to try to revert and reland this...

PR #4130, opened 2023-09-10
PERF Store js string in the Python string after we convert it

hoodmane:
This patches the Python interpreter to add an extra field to strings. We use this field to store JavaScript strings so that we can avoid converting each Python string more than once.

### DIFF
--- a/src/core/python2js.c
+++ b/src/core/python2js.c
@@ -164,22 +164,40 @@ EM_JS_REF(JsRef, _python2js_ucs4, (const char* ptr, int len), {
   return Hiwire.new_value(jsstr);
 });
 
+void
+PyUnicode_SetExtra(PyObject* unicode, JsRef extra);
+
+JsRef
+PyUnicode_GetExtra(PyObject* unicode);
+
 static JsRef
 _python2js_unicode(PyObject* x)
 {
+  JsRef result = PyUnicode_GetExtra(x);
+  if (result != NULL) {
+    hiwire_incref(result);
+    return result;
+  }
+
   int kind = PyUnicode_KIND(x);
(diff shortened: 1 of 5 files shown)


hoodmane:
Maybe we can disable this feature for some of the leak tests...

src/core/python2js.c
```diff
@@ -164,22 +164,40 @@ EM_JS_REF(JsRef, _python2js_ucs4, (const char* ptr, int len), {
   return Hiwire.new_value(jsstr);
 });
 
+void
+PyUnicode_SetExtra(PyObject* unicode, JsRef extra);
+
+JsRef
+PyUnicode_GetExtra(PyObject* unicode);
```
ryanking13, line 171: Please add some comments so people can find the patch that implements these functions later.

ryanking13 (approved):
Thanks @hoodmane! I'm a little reluctant to add a patch that can't be permanently removed, but I think this pattern will be very common, so if we can get enough performance gains, I'm +1 for it.

Makefile.envs
```diff
@@ -249,7 +249,8 @@ export PYO3_CROSS_INCLUDE_DIR=${PYTHONINCLUDE}
 export RUSTFLAGS= \
 	-C link-arg=-sSIDE_MODULE=2 \
 	-C link-arg=-sWASM_BIGINT \
-	-Z link-native-libraries=no
+	-Z link-native-libraries=no \
+	--cfg pyodide_js_string
```
ryanking13, line 253: is it safe to add this config even if the build backend does not accept this config?

hoodmane: Yeah if the rust code doesn't check for it, it will just do nothing.

PR #4131, opened 2023-09-10
Fix using Python 2 in MacOS CI

ryanking13:
Hopefully fixes an error: https://github.com/pyodide/pyodide/actions/runs/6132234123/job/16642912432


rth (approved):
Thanks! Yes, I also keep forgetting about this.

PR #4138, opened 2023-09-13
PERF Only `finalizationRegistry.unregister` proxy if it is registered

hoodmane:
This call is slow, don't do it if we don't need to.


ryanking13 (approved):
Thanks!

PR #4143, opened 2023-09-16
Make loadPyodide global in console.html

rth:
So it's easier to test something about `loadPyodide` in a `console.html` page which is already deployed.


ryanking13 (approved):
Thanks!

hoodmane:
I've been repeatedly annoyed about this too.

PR #4145, opened 2023-09-16
Rename setup.py in coolprop

ryanking13:
When I run `pytest -v packages`, pytest runs `packages/coolprop/setup.py` and throws error which is very annoying.

This PR Renames `packages/coolprop/setup.py` to `packages/coolprop/setup.py.wrapper` so that pytest does not collect it.


rth:
This works, but maybe better to do it via some [ignore pattern in `pyproject.toml`](https://github.com/pyodide/pyodide/blob/e492392f1b7716484a4681b707468fd191818d11/pyproject.toml#L112) to be more general?

rth (approved):

ryanking13:
Thanks for the suggestion! I added an global pytest option to ignore `packages/**/extras` and moved all extra files of packages to `extras` directory.

packages/matplotlib/meta.yaml
```diff
@@ -36,8 +36,8 @@ build:
     rm -rf matplotlib/backends/qt_editor
     rm -rf matplotlib/backends/web_backend
     rm -rf sphinxext
-    cp $PKGDIR/src/fontlist.json matplotlib
-    cp $PKGDIR/src/Humor-Sans-1.0.ttf matplotlib/mpl-data/fonts/ttf/Humor-Sans.ttf
+    cp $PKGDIR/extras/fontlist.json matplotlib
+    cp $PKGDIR/extras/Humor-Sans-1.0.ttf matplotlib/mpl-data/fonts/ttf/Humor-Sans.ttf
```
rth, line 40: Maybe we could have just kept the src/ folder for this then? No strong option. Anyway, now that you made all the changes, probably no point in reverting.

rth (approved):

packages/matplotlib/meta.yaml
ryanking13: Yeah, matplotlib is an exceptional case where the files should be copied after the package is built, but for now I'm going to put it in the same folder because it seems to be semantically similar to extras.

Issue #4144, opened 2023-09-16
Improve importlib.metadata import time


rth:
During a performance investigation with @hoodmane  a few months ago, we saw that at Pyodide initialization,  `importlib.metadata` took a while to import, in particular, because it loads the `email` module which is large.

Here is the output of `await loadPyodide({args: ['-X', 'importtime']});` showing import time of each module (it should be read from the bottom to the top as far as I understand). 

<details>

```
mport time: self [us] | cumulative | imported package
import time:      1000 |       1000 |   _io
import time:         0 |          0 |   marshal
import time:      1000 |       1000 |   posix 
import time:      3000 |       5000 | _frozen_importlib_external 
import time:         0 |          0 |   time 
import time:      1000 |       1000 | zipimport 
import time:         0 |          0 |     _codecs 
import time:      2000 |       2000 |   codecs 
import time:     18000 |      18000 |   encodings.aliases 
import time:     43000 |      63000 | encodings 
import time:      3000 |       3000 | encodings.utf_8 
import time:         0 |          0 | _signal 
import time:         0 |          0 |     _abc 
import time:      1000 |       1000 |   abc 
import time:      2000 |       3000 | io 
import time:         0 |          0 |       _stat 
import time:      2000 |       2000 |     stat 
import time:      3000 |       3000 |     _collections_abc 
import time:         0 |          0 |       genericpath 
import time:      1000 |       1000 |     posixpath 
import time:      2000 |       8000 |   os 
import time:      1000 |       1000 |   _sitebuiltins 
import time:      2000 |       2000 |   sitecustomize 
import time:      1000 |      12000 | site 
import time:         0 |          0 |   gc 
import time:      3000 |       3000 |   pyodide.ffi 
import time:      6000 |       9000 | pyodide._state 
import time:      1000 |       1000 |   js 
import time:     34000 |      35000 | pyodide.webloop 
import time:     59000 |      59000 | sysconfig 
import time:      1000 |       1000 |     _struct 
import time:      2000 |       3000 |   struct 
import time:         0 |          0 |   pwd 
import time:      1000 |       1000 |   grp 
import time:    179000 |     183000 | tarfile 
import time:      1000 |       1000 |     _csv 
import time:     30000 |      31000 |   csv 
import time:      3000 |       3000 |   email 
import time:         0 |          0 |     binascii 
import time:      1000 |       1000 |     lzma 
import time:    183000 |     184000 |   zipfile 
[...]

PR #4147, opened 2023-09-16
Don't use importlib.metadata during initialisation

rth:
Closes https://github.com/pyodide/pyodide/issues/4144

So we manually identify the packages installed based on their `.dist-info/*` files instead of relying on `importlib.metadata`. This also probably does much less work that `importlib.metadata` e.g. no need to parse all the metadata fields, we only need the package name.

Code wise it's not that much complex, the question is how robust this is. In particular, this assumes that `*.dist-info/METADATA` files do exist to get the non-normalized package name. [It should](https://packaging.python.org/en/latest/specifications/recording-installed-packages/#the-dist-info-directory)  with installers that respect the spec, but I don't know if in practice there are some edge cases that we could be missing.

### DIFF
--- a/src/py/pyodide/_package_loader.py
+++ b/src/py/pyodide/_package_loader.py
@@ -2,7 +2,6 @@ import re
 import shutil
 import sys
 import sysconfig
-import tarfile
 from collections.abc import Iterable
 from importlib.machinery import EXTENSION_SUFFIXES
 from pathlib import Path
@@ -309,6 +308,8 @@ def get_dynlibs(archive: IO[bytes], suffix: str, target_dir: Path) -> list[str]:
         The list of paths to dynamic libraries ('.so' files) that were in the archive,
         but adjusted to point to their unpacked locations.
     """
+    import tarfile
+    
     dynlib_paths_iter: Iterable[str]
     if suffix in ZIP_TYPES:
         dynlib_paths_iter = ZipFile(archive).namelist()
--- a/src/py/pyodide/_package_loader.py
+++ b/src/py/pyodide/_package_loader.py
@@ -324,29 +324,37 @@ def get_dynlibs(archive: IO[bytes], suffix: str, target_dir: Path) -> list[str]:
     ]
 
 
-def get_dist_source(dist_path: Path) -> str:
-    """Get a description of the source of a package.
+def get_dist_source(dist_path: Path) -> tuple[str, str]:
+    """Get the package name and a description of the source of a package.
 
     This is used in loadPackage to explain where the package came from. Purely
     for informative purposes.
     """
+    with (dist_path / "METADATA").open() as f:
+        for line in f:
+            if line.startswith("Name:"):
+                dist_name = line[5:].strip()
+                break
+        else:
+            raise ValueError(f"Package name not found in {dist_path.name} METADATA")
+
     source_path = dist_path / "PYODIDE_SOURCE"
     if source_path.exists():
         source = source_path.read_text().strip()
         if source == "pyodide":
-            return "default channel"
+            return dist_name, "default channel"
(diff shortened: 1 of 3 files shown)


rth:
So this does reduce significantly the number of stdlib modules loaded at `loadPyodide` initialization ([REPL for this PR](https://output.circle-artifacts.com/output/job/78d0d336-2d6b-4884-803f-1add9032604a/artifacts/0/root/repo/dist/console.html)), as compared to the original list in the parent issue,

<details>

```
import time: self [us] | cumulative | imported package 
import time:      1000 |       1000 |   _io 
import time:         0 |          0 |   marshal 
import time:         0 |          0 |   posix 
import time:      3000 |       4000 | _frozen_importlib_external 
import time:         0 |          0 |   time 
import time:      2000 |       2000 | zipimport 
import time:         0 |          0 |     _codecs 
import time:      2000 |       2000 |   codecs 
import time:      6000 |       6000 |   encodings.aliases 
import time:     14000 |      22000 | encodings 
import time:      2000 |       2000 | encodings.utf_8 
import time:         0 |          0 | _signal 
import time:         0 |          0 |     _abc 
import time:      2000 |       2000 |   abc 
import time:      1000 |       3000 | io 
import time:         0 |          0 |       _stat 
import time:      1000 |       1000 |     stat 
import time:      2000 |       2000 |     _collections_abc 
import time:         0 |          0 |       genericpath 
import time:      1000 |       1000 |     posixpath 
import time:      3000 |       7000 |   os 
import time:         0 |          0 |   _sitebuiltins 
import time:      2000 |       2000 |   sitecustomize 
import time:      4000 |      13000 | site 
import time:         0 |          0 |   gc 
import time:      3000 |       3000 |   pyodide.ffi 
import time:      6000 |       9000 | pyodide._state 
import time:         0 |          0 |   js 
import time:     17000 |      17000 | pyodide.webloop 
import time:     23000 |      23000 | sysconfig 
import time:         0 |          0 |     _struct 
import time:      4000 |       4000 |   struct 
import time:         0 |          0 |   pwd 
import time:         0 |          0 |   grp 
import time:     73000 |      77000 | tarfile 
import time:         0 |          0 |   binascii 
import time:         0 |          0 |   lzma 
import time:     66000 |      66000 | zipfile 
import time:         0 |          0 | pyodide_js 
import time:     18000 |      18000 | _sysconfigdata__emscripten_wasm32-emscripten
```

</details>

However timing `loadPyodide` has large fluctuations, so I don't really see a noticeable impact there.

src/py/pyodide/_package_loader.py
```diff
@@ -326,27 +324,37 @@ def get_dynlibs(archive: IO[bytes], suffix: str, target_dir: Path) -> list[str]:
...
     """
-    source = dist.read_text("PYODIDE_SOURCE")
-    if source == "pyodide":
-        return "default channel"
-    if source:
-        return source
-    direct_url = dist.read_text("direct_url.json")
-    if direct_url:
+    with (dist_path / "METADATA").open() as f:
+        for line in f:
```
ryanking13, line 334: Wow, I didn't know that it is possible to iterate an file descriptor for years

ryanking13 (approved):
Thanks for working on this @rth! It looks good to me.

ryanking13:
> import time:     73000 |      77000 | tarfile 

Probably we can also remove tarfile from initialization.

It is used here:

https://github.com/pyodide/pyodide/blob/d644a405b48a2d36de6f1766b5f753d10ad377bb/src/py/pyodide/_package_loader.py#L318

but I think we can move the import statement into the function to delay the import

rth:
> Probably we can also remove tarfile from initialization.

Good point, thanks. Done.

PR #4163, opened 2023-09-21
FIX patch a bug in Emscripten that broke matplotlib

### LINKED ISSUE #4153: matplotlib breaks in v0.24.0
## 🐛 Bug

In an attempt to update [polyscript](https://github.com/pyscript/polyscript/pull/38) to this latest version of Pyodide, I've smoke-tested our long working [matplotlib example](https://github.com/pyscript/polyscript/blob/main/test/matplot.html) and saw this on screen:

Screenshot from 2023-09-18 14-27-31

### To Reproduce

  * grab [polyscript](https://github.com/pyscript/polyscript)
  * enter the folder then checkout `pyodide-update` branch
  * `npm i` followed by `npm run build` and finally `npm run server`
  * visit the localhost url and try any *matplot* example you like (they are all the same)
  * see the error ... fee free to change the version via absolute URL or local file [in here](https://github.com/pyscript/polyscript/blob/main/test/matplot.html#L12) as `version="http://localhost..."` then refresh for better debugging
  * feel free to use `NO_MIN npm run build` if you think something is wrong to the library (although the same example works by using previous version)

### Expected behavior

A page with an image once everything is done ... I just see a lot of errors instead with 0.24

### Environment

- Pyodide Version: 0.24.0
- Browser version: latest Chrome
- Any other relevant information: I can create a plain example if needed ... just let me know

hoodmane:
See upstream PR https://github.com/emscripten-core/emscripten/pull/20309.
Resolves #4153.

- [x] Add / update tests


ryanking13 (approved):

PR #4164, opened 2023-09-22
FIX Don't crash if checkInterrupt() is called with gil released and no ib

hoodmane:
`Py_EmscriptenSignalBuffer` could be undefined

- [x] Add / update tests


ryanking13 (approved):
Thanks!

PR #4170, opened 2023-09-24
Shorten error handling wrapper

hoodmane:
This shortens pyodide.asm.js by about 5kb by reducing the space taken up by this 
unreachable macro-generated code


rth:
Nice, thanks! I wondering whether we should enable running the closure compiler via the emscripten option. It's annoying to also require Java, and it might also add some compilation time, but maybe we could make it optional via some flag.

rth (approved):

PR #4179, opened 2023-09-26
CI Change macos resource class to gen2

ryanking13:
Following the deprecation notice from Circie CI: https://discuss.circleci.com/t/macos-intel-support-deprecation-in-january-2024/48718

Let's see if Pyodide runs well with m1...

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -321,7 +321,7 @@ jobs:
         description: pytest-cache-dir.
         type: string
         default: ""
-    resource_class: medium
+    resource_class: macos.m1.medium.gen1
     macos:
       xcode: 14.3.1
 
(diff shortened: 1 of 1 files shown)


ryanking13:
Hmm it doesn't seem to work. Maybe it is not in the free plan? (https://circleci.com/product/features/resource-classes/#macos)

rth:
Yes, sounds like it. Well, we'll probably have only run MacOS on GH Actions..

ryanking13:
> Yes, sounds like it. Well, we'll probably have only run MacOS on GH Actions..

For now, I changed the resource to gen2 which is supported for 3 more months from now. Let's remove MacOS CI if CircleCI still does not support MacOS for free plan users 3 months later.

rth:
Thanks!

cclauss:
M1 is now Generally Available on GitHub Actions…
https://github.blog/changelog/2024-04-01-macos-14-sonoma-is-generally-available-and-the-latest-macos-runner-image

PR #4185, opened 2023-09-27
Add cffi to pynacl host dependency

### LINKED ISSUE #4184: PyNaCl build is flaky
## 🐛 Bug

Sometimes we observe the following error occurring in CI.

```
FAILED _tests/test_packages_common.py::import[chrome-pynacl] - pytest_pyodide.runner.JavascriptException: PythonError: Traceback (most recent call last):
  File "/lib/python311.zip/_pyodide/_base.py", line 571, in eval_code_async
    await CodeRunner(
  File "/lib/python311.zip/_pyodide/_base.py", line 394, in run_async
    coroutine = eval(self.code, globals, locals)
                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
ModuleNotFoundError: The module 'pynacl' is included in the Pyodide distribution, but it is not installed.
  await micropip.install("pynacl") in Python, or
  await pyodide.loadPackage("pynacl") in JavaScript
See https://pyodide.org/en/stable/usage/loading-packages.html for more details.
```

When I looked at the `pynacl` wheel of the failed PR, it contains the following directories:

```
 .
 |-PyNaCl-1.5.0.data
 |---purelib
 |-----nacl
 |-------bindings
 |-------pwhash
 |-PyNaCl-1.5.0.dist-info
```

while the "correct" `pynacl` wheel should contain:

```
 .
 |-PyNaCl-1.5.0.dist-info
 |-nacl
 |---bindings
 |---pwhash
```

So it seems like the build system is putting source codes of pynacl to the .data/purelib directory for some reason.

I can guess that the setuptools is related, but I have no more idea. Regarding that this behavior is not happening all the time, maybe building multiple packages together is causing this...

ryanking13:
Close #4184

### DIFF
--- a/packages/pynacl/meta.yaml
+++ b/packages/pynacl/meta.yaml
@@ -11,7 +11,7 @@ source:
 requirements:
   run:
     - cffi
-  build:
+  host:
     - cffi
 about:
   home: https://github.com/pyca/pynacl/
--- a/packages/pynacl/meta.yaml
+++ b/packages/pynacl/meta.yaml
@@ -11,6 +11,8 @@ source:
 requirements:
   run:
     - cffi
+  build:
+    - cffi
 about:
   home: https://github.com/pyca/pynacl/
   PyPI: https://pypi.org/project/pynacl


rth (approved):
Not sure how it was working even occasionally without this.

ryanking13:
> Not sure how it was working even occasionally without this.

Maybe we were lucky that cffi was built before pynacl in many cases :)

PR #4186, opened 2023-09-27
use emscripten llvm-readobj by default

joemarshall:
`pywasmcross` uses llvm-readobj. Currently it first checks if there's one on the path, and falls back to emscripten's version.

This means `pyodide build` fails if you have a system with an old llvm installed on the path.

This change makes it use the correct one from emsdk for preference.

### DIFF
--- a/pyodide-build/pyodide_build/pywasmcross.py
+++ b/pyodide-build/pyodide_build/pywasmcross.py
@@ -413,12 +413,21 @@ def _calculate_object_exports_readobj_parse(output: str) -> list[str]:
 
 
 def calculate_object_exports_readobj(objects: list[str]) -> list[str] | None:
-    readobj_path = shutil.which("llvm-readobj")
-    if not readobj_path:
-        which_emcc = shutil.which("emcc")
-        assert which_emcc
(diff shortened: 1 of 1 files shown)


pyodide-build/pyodide_build/pywasmcross.py
```diff
@@ -413,12 +413,21 @@ def _calculate_object_exports_readobj_parse(output: str) -> list[str]:
...
+    if readobj.exists():
+        readobj_path: str | None = str(readobj)
+    else:
+        readobj_path = shutil.which("llvm-readobj")
+    assert readobj_path
+    args = [
+        readobj_path,
+        "--section-details",
+        "-st",
+    ] + objects
```
rth, line 429: Duplicate block from below, I think.

rth (approved):
Thanks!

ryanking13:
`pyodide-build` is now moved to a separate repository. Could you please re open the PR if it is still necessary?

agriyakhetarpal:
Oh, I forgot about that, yes. I think this would still be useful; let me migrate it.

PR #4187, opened 2023-09-27
Vendor private pypa/build APIs to pyodide-build

ryanking13:
### Description

This vendors private pypa/build APIs into pyodide-build to make pyodide-build robust to pypa/build update.


rth (approved):
Thanks @ryanking13 ! Indeed it makes sense vendor those few logging functions.

PR #4189, opened 2023-09-27
DOCS Some minor tuneups

hoodmane:
1. Future versions of typedoc don't allow tags to contain _ so change `@doc_type` ==> `@doctype`
2. `Sequence` and `MutableSequence` are miscategorized because we forgot to export them from `ffi.ts`. TODO: Make it so we don't need to do this??
3. For the canvas getters/setters I moved the documentation into the top level. This means that the APIs get a summary in the summary table.
4. Also, `private` stops working for certain things in future typedoc versions, so replace it with `@hidden`


rth (approved):
Thanks!

PR #4193, opened 2023-09-29
Switch to using sphinx-js fork

hoodmane:
This allows us to remove a lot of code from sphinx_pyodide. It will take more effort to upstream the automodule documenter features, but all the type formatting features have been moved upstream and I added hooks to render xrefs into the types and to avoid the converter monkeypatch.

The new version also has some neat improvements like automatically fixing up `code between single upticks` into proper code instead of rendering it in italics, better "examples" rendering, etc. Mostly tiny things. Also we moved to typedoc 0.25.0 which itself does a lot of things a bit better.


hoodmane:
@rth @ryanking13 Can someone review this when they get a chance?

src/js/tsdoc.json
```diff
@@ -0,0 +1,23 @@
+{
+  "$schema": "https://developer.microsoft.com/en-us/json-schemas/tsdoc/v0/tsdoc.schema.json",
+  "extends": ["typedoc/tsdoc.json"],
+  "noStandardTags": false,
+  "tagDefinitions": [
```
rth, line 5: Can you add a few comments about what these are used for,  for later reference?

rth (approved):
Thanks, @hoodmane, it's a nice simplification of the code sphinx_pyodide! I'll let you double-check that the rendering is working as expected )

src/js/tsdoc.json
hoodmane: Well we can't add comments in json files...

hoodmane:
> I'll let you double-check that the rendering is working as expected )

Yeah I spent a **lot** of time comparing the rendering to get all the details correct.

src/js/tsdoc.json
rth: It shows that I haven't edited json for a while )

rth:
Thanks a lot, Hood for your patience in doing this!

src/js/tsdoc.json
hoodmane: I'll add a comment in about this file in `jsdoc.py`, because it is indeed easy to forget about it.

Issue #1576, opened 2021-05-05
Document preloading options


joemarshall:
I'm not sure where this needs to be in the documentation, but I think it would be worth having examples of how to preload everything in javascript (and making the example console.html etc. do this right). Then people like me who are not javascript ninjas might know how important it is for pyodide load time.

I'm using a funky CDN and my own custom build of pyodide right now, otherwise I would do it. 

My preload at the top of my HTML file looks like this: It took a bit of fiddling with crossorigin things and the correct as values, but this reduces pyodide ready time from 12 seconds to 4 seconds without any network cache.

It finishes loading all network resources at 1.29 seconds here on my okay network connection (home connection, 79 Mbps). From cache it is usually <800ms.

The other 3 seconds are spent in emscripten / python startup shenanigans. 

Here is what my preloads relating to pyodide look like:
```
<link rel="preload" href="https://gitcdn.link/repo/joemarshall/websensors/main/pyodide/pyodide.js" as="script" >
<link rel="preload" href="https://gitcdn.link/repo/joemarshall/websensors/main/pyodide/pyodide.asm.data" as="fetch" crossorigin="anonymous">
<link rel="preload" href="https://gitcdn.link/repo/joemarshall/websensors/main/pyodide/pyodide.asm.wasm" as="fetch" crossorigin="anonymous">
<link rel="preload" href="https://gitcdn.link/repo/joemarshall/websensors/main/pyodide/pyodide.asm.js" as="script">
<link rel="preload" href="https://gitcdn.link/repo/joemarshall/websensors/main/pyodide/packages.json" as="fetch" crossorigin="anonymous">
```

hoodmane:
A good place to start would be adding this to `console.html`. Probably a lot of people start out by copying the setup there.

But we should have a section of the docs for this. It'd also be helpful to remind people to use a cdn or configure their servers to compress the files since in practice this seems to be a step many of us amateur web developers forget.

hoodmane:
According to: https://web.dev/module-workers/

> `importScripts()` loads scripts using the `script` resource type, just like `<script>`. 

so this same stuff should work if Pyodide is loaded via `importScripts` in a webworker.

stefnotch:
I just tried this out and couldn't quite get it to behave as expected. Maybe this is indeed an issue:
https://stackoverflow.com/questions/63761046/what-happen-when-a-script-is-loaded-as-preload-modulepreload

rth:
[...]

PR #4199, opened 2023-10-03
Rearrange the way that we load the lock file

hoodmane:
This fetches the lock file a bit earlier


src/js/load-package.ts
```diff
@@ -64,7 +57,9 @@ async function initializePackageIndex(lockFileURL: string) {
   await loadPackage(API.config.packages, { messageCallback() {} });
 }
 
-API.packageIndexReady = initializePackageIndex(API.config.lockFileURL);
+if (API.lockFilePromise) {
```
rth, line 60: Is there a case where this is not true though and shouldn't it be an error? or is it just for typing? Maybe worth adding a comment.

rth (approved):
Thanks @hoodmane ! Nice improvement indeed,

before

image

after

image

(it's circle CI artifacts so there is an extra redirect for pyodide-lock.json).

Generally, maybe we should just document pre-loading, though the experimentation in https://github.com/pyodide/pyodide/issues/1576 wasn't very conclusive. This is kind of implementing that manually.

hoodmane:
It seems to break a bit on Node for some reason...

src/js/load-package.ts
hoodmane: Yeah well I feel bad about this not being a function call from `pyodide.ts` but it can't be for performance reasons. But I'm also interested in allowing libpyodide to be linked into other applications that possibly don't use the loader logic. Particularly I've been investigating things like linking libpyodide into a Python without dynamic linking, memory snapshotting etc. Mostly what we have is flexible enough that we can just replace `pyodide.ts` and link libpyodide.a but this line causes trouble.

PR #4202, opened 2023-10-04
MAINT Add some more types to various things

hoodmane:


src/core/error_handling.ts
```diff
@@ -74,14 +75,16 @@ let fatal_error_occurred = false;
  * @argument e {Error} The cause of the fatal error.
  * @private
  */
-API.fatal_error = function (e: any) {
+API.fatal_error = function (e: any): never {
   if (e && e.pyodide_fatal_error) {
+    // @ts-ignore
```
ryanking13, line 80: Why this function has `never` return type?

src/js/lib.ts
ryanking13, line 1: I am not sure if `lib` is the best name for this file. As it mostly defines types, maybe something like `types.ts` would be better?

hoodmane: > mostly

I would say entirely

src/core/error_handling.ts
hoodmane: I want it to type narrow:
```ts
let x;
try {
  x = get number();
} catch(e) { 
    fatal_error(e);
}
// now I want x to have type number not type number | undefined
```

hoodmane: The exceptions to it not returning are in various cases that we're already in the process of fatally error teardown and we don't want to recursively fail. I could care less if there are more logic errors on this already dying code path and I don't want typescript trying to warn me about it.

ryanking13: Okay, I think it makes sense. Thanks for the explanation.

ryanking13 (approved):
Thanks, @hoodmane!

hoodmane:
Thanks for the review @ryanking131

PR #4204, opened 2023-10-06
BLD Make meson use pywasmcross wrappers

ryanking13:
### Description

Makes `meson` build system use pywasmcross compiler wrappers instead of calling `emcc` directly. For now, only scikit-image is affected, but this is a preparation to support other packages like scipy, numpy.

This PR includes the following changes:

- Adds a custom runner for pypa/build which prepends pywasmcross symlink directory to the PATH whenever it invokes a subprocess.
  - By doing this, calling `meson` will use pywasmcross wrapper, not a real meson binary.
- meson cross-file is dynamically created in the symlink directory


slackline:
Would be great to see this merged `scikit-image` and `numpy` at least are affected, the later of which is a key package for many.

slackline:
Fair-dos and thanks for working on this, very much appreciated.

ryanking13:
Superseded by #4240

PR #4205, opened 2023-10-06
BLD Unpin types-setuptools

ryanking13:
I think this is not needed anymore (https://github.com/pyodide/pyodide/pull/3613)


rth (approved):
Thanks!

PR #4211, opened 2023-10-10
CI Use xcode 15.0.0 for Safari tests

hoodmane:
It seems that there is a bug in table instructions in Safari 16.3 which is used in the safari tests but that Safari 16.4 has fixed it. xcode 15.0.0 uses Safari 16.6 which does not have the bug.


ryanking13:
Thanks!

ryanking13:
It looks like safari tests get very unstable after this update. Maybe we should turn it off temporarily?

PR #4212, opened 2023-10-11
Improve handling of generated typescript files

hoodmane:
This makes IDEs, documentation generation tools, and linters handle our generated files better. I made a new folder called `src/js/generated` to place the generated files into and added a ts resolution order so that we prefer files in `src/js/generated` to the same file in src/core. This way we don't have to copy `src/core/error_handling.ts` at all since we are not modifying it. Also, it makes it so that we don't need to move files around to generate the docs and it simplifies the makefile a little bit.


ryanking13 (approved):
LGTM, thanks!

hoodmane:

PR #4218, opened 2023-10-11
Use externrefs for some ffi code

hoodmane:
This begins migrating some of the ffi code towards using externrefs,
which should make it faster and shorter.

I'm planning on migrating from the existing `hiwire.h` functions to new `jslib.h` functions which will operate on `externref`s, but I think it will be easiest to do this in several PRs since it's going to be a fair amount of code that changes (in mostly boilerplate ways though).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -23,7 +23,7 @@ ccache
 cpython/downloads
 cpython/installs
 cpython/build.log
-dist/
+dist*/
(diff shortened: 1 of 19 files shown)


ryanking13 (approved):
Thanks for working on this @hoodmane! LGTM.

PR #4221, opened 2023-10-12
Change wheel platform (take 2)

hoodmane:
Supersedes #3823.

- [ ] Add / update tests


Makefile.envs
```diff
@@ -1,5 +1,6 @@
 export PYVERSION ?= 3.11.3
 export PYODIDE_EMSCRIPTEN_VERSION ?= 3.1.45
+export PYODIDE_ABI ?= 202308_0
```
rth, line 3: Sounds good.

pyodide-build/pyodide_build/build_env.py
```diff
@@ -285,10 +293,12 @@ def pyodide_tags() -> Iterator[Tag]:
     """
     PYMAJOR = get_pyversion_major()
     PYMINOR = get_pyversion_minor()
-    PLATFORM = platform()
+    # Accept both "pyodide_abiver_wasm32" and "emscripten_ver_wasm32" (rust packages)
+    platforms = [platform(), emscripten_platform()]
```
rth, line 297: This should probably be documented somewhere on the page for loading packages. Maybe we could do a separate subsection on wheel compatibility?

pyodide-build/pyodide_build/common.py
```diff
@@ -269,6 +269,41 @@ def _get_sha256_checksum(archive: Path) -> str:
     return h.hexdigest()
 
 
+def ensure_wheel_platform(path: Path) -> Path:
```
rth, line 272: Are you sure this leaves universal wheels unchanged? Can you add a test for it? I think if  in a test, you give it an empty zipfile with a wheel-looking name it [should work](https://github.com/pypa/wheel/blob/e76040b0d970a0fe6211c31888fc929078ba02d9/src/wheel/cli/tags.py#L118)

src/core/pyversion.c
```diff
@@ -19,3 +23,30 @@ py_version_micro()
...
   return PY_MICRO_VERSION;
 }
+
+#include <sys/utsname.h>
+
+#define STRINGIFY(s) #s
+#define STR(s) STRINGIFY(s)
+
+int
+__syscall_uname(intptr_t buf)
```
rth, line 33: Can you share the uname before and after this change? Is that the thing that gets displayed when you start the interpreter or something else? It might still be useful to display the emscripten version there no?

For instance (assuming I'm not mistaken about what this does), on Linux the interpreter startup string is,
```
Python 3.10.4 | packaged by conda-forge | (main, Mar 24 2022, 17:39:04) [GCC 10.3.0] on linux
```
so it includes the compiler version. In our currently we have,
```
Python 3.11.3 (main, Sep 25 2023 20:45:01) on WebAssembly/Emscripten
```
and IMO it could be useful to include both Emscripten and Pyodide versions there.

Maybe,
```
Python 3.11.3 (main, Sep 25 2023 20:45:01) packaged by Pyodide (ABI 20230801) [Emscripten 1.38.1] on wasm32
```
though it's a bit too long, hmm.

rth (commented):
Thanks @hoodmane ! A few comments otherwise LGTM.

rth:
Also, do we plan for some programmatic way for getting the Emscripten and Pyodide versions from within the browser? One of them would be in the platform tag, but maybe it could be still useful to expose them as part of the Pyodide python or JS package?

In particular, from micropip how do we determine the compatible emscripten & pyodide platform wheels? BTW, with this ABI versioning `202308_01` and `202308_02` would not be compatible, right?

pyodide-build/pyodide_build/common.py
hoodmane: Universal wheels should be protected by the guard at the beginning:
```py
    if "emscripten" not in path.name:
        return path
```

src/core/pyversion.c
hoodmane: It still says:
```
Python 3.11.3 (main, Oct 13 2023 21:46:59) on WebAssembly/Emscripten
```
there. I'm not actually sure where all the different things come from...

hoodmane:
The plan would be for v202308_01 wheels to work on v202308_02 interpreters but not the other way around. The idea being that if we link an extra static library in we can bump the minor version number. But `packaging` and `pip` and other tools would need to know about this... Ideally package manipulation tools would also accept both pyodide_abi_wasm32 and emscripten_ver_wasm32 wheels, but pypi will reject emscripten_ver_wasm32 wheels. If this logic were updated, it probably wouldn't be sensitive to the `uname`.

ccing some pypa people: @henryiii @pradyunsg @di

src/core/pyversion.c
ryanking13: __Before__

```
>>> import platform
>>> platform.uname()
uname_result(system='Emscripten', node='emscripten', release='3.1.45', version='#1', machine='wasm32')
```

__After__

```
>>> import platform
>>> platform.uname()
uname_result(system='Pyodide', node='pyodide', release='202308_0', version='#1', machine='wasm32')
```

We can probably update our console banner string:

https://github.com/pyodide/pyodide/blob/5a1826355afe38a448ea06e029c3836ba00308d1/src/py/pyodide/console.py#L26-L29

ryanking13: Interestingly conda-forge patches platform.py to display the conda-forge banner: https://github.com/conda-forge/python-feedstock/blob/main/recipe/brand_python.py

rth: OK, makes sense

PR #4222, opened 2023-10-12
Put pyodide version into xbuildenv name

### LINKED ISSUE #4517: The .pyodide-xbuildenv cache is not version dependent, make build fail
## 🐛 Bug

From what I can tell doesnt the pyodide build version the cache folder `.pyodide-xbuildenv`. So if I install pyodide 0.25, run a pyodide build command and then upgrade pyodide the cache folder will contain data that doesnt match and will break the build, require me to manually clear the folder.

### To Reproduce

```
> pip install pyodide-build==0.25.0
> pyodide config get emscripten_version
3.1.46
> pip install pyodide-build==0.26.0a1 
> pyodide config get emscripten_version
xbuild environment already exists, skipping download
3.1.46
```

### Expected behavior

The second call to pyodide, with the updated version, should return emscripten version 3.1.52. (I think there are more think breaking as well, but this is an obvious part at least).

I think that it could either have caches for multiple versions within the folder, or simply clear it and fetch again. Since you wont switch between versions very often I think its perfectly fine to just clear the folder and fetch again. 

### Environment

- Pyodide Version 0.26.0a1

hoodmane:
This should reduce the likelihood of accidentally using an xbuildenv for the wrong Pyodide version.

Resolves #4517.


ryanking13 (approved):
Thanks @hoodmane. It looks like there is a test failure.

PR #4223, opened 2023-10-14
Fix Emscripten.cmake not vendored in pyodide-build wheel

### LINKED ISSUE #4216: Emscripten.cmake missing from pyodide-build wheel
## 🐛 Bug

It seems like `Emscripten.cmake` is not bundled with `pyodide-build` wheels any more (since version 0.24.0) and I wonder whether this is a bug or an intended change. (I haven't found anything in the changelog). We were relying on the presence of the toolchain file over there in https://github.com/igraph/python-igraph/ to build a wasm wheel in CI and it is not working any more.

### To Reproduce

See, e.g., https://github.com/igraph/python-igraph/actions/runs/6459502630/job/17535393423

### Expected behavior

`pyodide-build` wheels should keep on bundling `Emscripten.cmake` so we can simply `pip install pyodide-build` in CI and then build `python-igraph`'s wasm wheels as before.

### Environment

- Pyodide Version: 0.24.1
- Browser version: not applicable

ryanking13:
### Description

Resolve #4216

This PR fixes a bug where Emscripten.cmake is not vendored in pyodide-build wheel since 0.24.0.

This was caused by https://github.com/pyodide/pyodide/pull/3947, where `ini2toml` changed all keys to lowercase but the setuptools is case sensitive...

This PR also changes the `CMAKE_TOOLCHAIN_FILE` and `PYO3_CONFIG_FILE` environment variables to allow them to be overwritten by the user.

### Checklists

### DIFF
--- a/docs/project/changelog.md
+++ b/docs/project/changelog.md
@@ -28,6 +28,16 @@ myst:
   which reduces the time to load Pyodide.
   {pr}`4147`
 
+### Build system
+
+- {{ Fix }} Fixed `Emscripten.cmake` not vendored in pyodide-build since 0.24.0.
+  {pr}`4223`
+
+- {{ Fix }} pyodide-build now does not override `CMAKE_CONFIG_FILE` and `PYO3_CONFIG_FILE`
+  env variables if provided by user.
+  {pr}`4223`
+
+
 ## Version 0.24.1
 
 _September 25, 2023_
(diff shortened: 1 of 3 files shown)


hoodmane (approved):

henryiii:
All downstream CMake builds are still broken due to this. @abravalheri, are you aware of this ini2toml bug?

abravalheri:
> All downstream CMake builds are still broken due to this. @abravalheri, are you aware of this ini2toml bug?

I don't remember seeing this one before... Definitely something to be fixed (although I don't know how to do that yet 😅).

PR #4225, opened 2023-10-14
Drop Safari 14.x polyfill

hoodmane:
Since we added externref, we won't support Safari 14.x anymore so we don't need the polyfill.


ryanking13 (approved):
Thanks!

Maybe we need to updated the minimum supported safari version in docs then?

https://github.com/pyodide/pyodide/blob/af7544cbbb034131a4d7b842439d41bff38d1159/docs/usage/index.md?plain=1#L67

Makefile.envs
```diff
@@ -129,8 +129,6 @@ export MAIN_MODULE_LDFLAGS= $(LDFLAGS_BASE) \
 	-s TOTAL_MEMORY=20971520 \
 	-s ALLOW_MEMORY_GROWTH=1 \
 	-s EXPORT_ALL=1 \
-	-s POLYFILL \
-	-s MIN_SAFARI_VERSION=140000 \
```
rth, line 133: Also, we still should specify the min Safari version for emscripten, shouldn't we?

hoodmane: Well it is now going to default to `150000` which is the max version that Emscripten ever checks against. But I think we are in effect requiring Safari 16.4 because although the Safari 15.0 changelog reports adding support for externref, the Safari implementation is broken until version 16.4. Of course there is changelog entry to reporting the bugfix...

hoodmane:
The rest of this table is pretty out of date:

| Browser | Minimal supported version | Release date    |
| ------- | ------------------------- | --------------- |
| Firefox | 70.0                      | 22 October 2019 |
| Chrome  | 71.0                      | 4 December 2018 |


For several releases we've required:

| Browser | Minimal supported version | Release date    |
| ------- | ------------------------- | --------------- |
| Firefox | 78.0                      | 30 June 2020 |
| Chrome  | 85.0                      | 25 August 2020 |
| Safari  | 14.1                      | 4 May 2021 |

Now we'll be requiring:


| Browser | Minimal supported version | Release date    |
| ------- | ------------------------- | --------------- |
| Firefox | 79.0                      | 28 July 2020 |
| Chrome  | 96.0                      | 30 August 2022 |
| Safari  | 16.4                      | 27 March 2023 |

If only Safari 15.0 (September 20, 2021) had working externref support like it claims...
Chrome 96 and Safari 16.4 support SIMD so to require SIMD we only have to bump firefox MSV to 91 released August 10, 2021. Firefox really smokes the other browsers for webassembly features.

ryanking13:
Maybe we should replace the "minimum supported version" with something like "tested version"...?

PR #4236, opened 2023-10-21
Use JsVal instead of JsRef in more places (nr. 3)

hoodmane:
This also fixes `JsArray_pop` which leaked memory and returned a singleton list with the element removed rather than the element itself.

- [x] Add / update tests


src/tests/test_jsproxy.py
```diff
@@ -105,16 +105,16 @@ def test_jsproxy_document(selenium):
 @pytest.mark.parametrize(
     "js,result",
     [
-        ("{}", False),
+        ("{}", True),
```
ryanking13, line 108: Were we treating an empty JS object as a falsy value before?
I got a different result when I tried this.

```js
a = {}
pyodide.runPython(`
from js import a
print(bool(a))
`)
```
```
True
```

ryanking13 (approved):
Thanks! LGTM with a question.

> This also fixes JsArray_pop which leaked memory and returned a singleton list with the element removed rather than the element itself.

Nice!

src/tests/test_jsproxy.py
hoodmane: Empty object was always truthy, but the test was written incorrectly. If you do `eval("{}")`, `{}` is parsed as an empty block not as an object. Thus, it returns `undefined` which is `Falsey`. I adjusted the test so it calls `run_js(f"({js})")` which calls `eval("({})")` which returns empty object with is truthy.

I think empty object is most comparable to:
```py
class A:
   pass
bool(A())
```
We could do something like:
```js
function isTruthy(obj) {
  if (obj.constructor !== Object && obj.constructor !== null) {
    return true;
  }
  for (const prop in obj) {
    if (Object.hasOwn(obj, prop)) {
      return true;
    }
  }
  // object with default or null constructor and no ownProperties
  return false;
}
```

hoodmane: Anyways I am open to switching to this in the future.

hoodmane:
I added a lot of methods in the same PR with `JsArray_pop`, so it stands to reason that I forgot to test one.

src/tests/test_jsproxy.py
ryanking13: I see. Thanks for the explanation!

PR #4240, opened 2023-10-22
Bump numpy version to 1.26.1 and use meson backend

ryanking13:
### Description

supersedes #4204

This PR improves meson support in pyodide-build, and apply it to numpy and scikit-image.

1. Make meson use pywasmcross compiler wrappers, instead of calling `emcc` directly.
2. Apply emscripten.meson.cross meson cross file automatically even when a user does not specify it.
  a. Sadly, this doesn't work for numpy as it uses [vendored meson](https://github.com/numpy/numpy/tree/main/vendored-meson). So for numpy cross file is applied manually by adding a backend-flags.


pyodide-build/pyodide_build/tools/emscripten.meson.cross
```diff
@@ -0,0 +1,14 @@
+# compilar paths are omitted intentionally so we can override the compiler using environment variables
```
hoodmane, line 1: Typo in comment =P
```suggestion
# compiler paths are omitted intentionally so we can override the compiler using environment variables
```

pyodide-build/pyodide_build/pypabuild.py
```diff
@@ -33,10 +34,49 @@
...
+        env = os.environ.copy()
+        if extra_environ:
+            env.update(extra_environ)
+
+        # Some build dependencies like cmake, meson installs binaries to this directory
+        # and we should add it to the PATH so that they can be found.
+        env["BUILD_ENV_SCRIPTS_DIR"] = isolated_build_env._scripts_dir
+        env["PATH"] = f"{cross_build_env['COMPILER_WRAPPER_DIR']}:{env['PATH']}"
+        # For debugging: Uncomment the following line to print the build command
+        # print("Build backend call:", cmd, file=sys.stderr)
```
hoodmane, line 74:
```suggestion
        # print("Build backend call:", " ".join(str(x) for x in cmd), file=sys.stderr)
```

pyodide-build/pyodide_build/pywasmcross.py
```diff
@@ -573,6 +580,24 @@ def handle_command_generate_args(  # noqa: C901
...
+                "--cross-file",
+                os.environ["MESON_CROSS_FILE"],
+            ]
+
+        return line
+    elif cmd in ("install_name_tool", "otool"):
+        # In MacOS, meson tries to run install_name_tool to fix the rpath of the shared library
+        # assuming that it is a ELF file. We need to skip this step.
+        # See: https://github.com/mesonbuild/meson/issues/8027
+        return ["echo", *line]
```
hoodmane, line 600: I think if we don't want to run the code we can just return `None` here. Is there a reason you want to echo it?

hoodmane (approved):

rth:
Great, thanks for doing this work!

rth:
Also cc @rgommers related to https://github.com/numpy/numpy/pull/24603

pyodide-build/pyodide_build/pywasmcross.py
ryanking13: It was to make those commands be shown in the build.log, which makes debugging easier.

pyodide-build/pyodide_build/tools/emscripten.meson.cross
ryanking13: Thanks!

packages/numpy/meta.yaml
```diff
@@ -1,16 +1,24 @@
...
+  sha256: c8c6c72d4a9f831f328efb1312642a1cafafaa88981d9ab76368d50d07d93cbe
 
 build:
-  backend-flags: --disable-optimization
+  # numpy uses vendored meson, so we need to pass the cross file manually
+  backend-flags: |
+    setup-args=-Dallow-noblas=true
+    setup-args=--cross-file=${MESON_CROSS_FILE}
+  # numpy creates numpy/distutils/__pycache__ directory during the build.
+  # It breaks our test because there is a .pyc in the directory.
```
rgommers, line 18: This is caused by a direct import of a utility with:
```python
    spec = importlib.util.spec_from_file_location('conv_template', genapi_py)
    mod = importlib.util.module_from_spec(spec)
    spec.loader.exec_module(mod)
```

I can rework that if it's a problem, and add a test to prevent a regression. What is the constraint here, no `.pyc` files under `numpy/`? Is it okay for there to be some under `tools/` instead (so outside the package source tree)?

rgommers:
This looks great, thanks @ryanking13!

> Sadly, this doesn't work for numpy as it uses [vendored meson](https://github.com/numpy/numpy/tree/main/vendored-meson). So for numpy cross file is applied manually by adding a backend-flags.

That'll take a little while to untangle unfortunately, as we need to finish upstreaming both BLAS and SIMD support. I suspect we won't make that in time for 2.0. It seems like that's only a minor hiccup here though, fortunately.

packages/numpy/meta.yaml
hoodmane: I don't think this is a big enough deal that you should think about changing numpy. It's pretty normal for there to be build time imports, we just have to deal with it. We could just always delete `__pycache__` folders for all packages.

ryanking13: Right, it isn't a big issue.

ryanking13:
> That'll take a little while to untangle unfortunately, as we need to finish upstreaming both BLAS and SIMD support. I suspect we won't make that in time for 2.0. It seems like that's only a minor hiccup here though, fortunately.

Thanks for the response! This isn't a big issue either, so I believe it's okay. In fact, Pyodide has been working around package-specific build issues in various hacky ways for a long time :)

hoodmane:
Thanks for doing this @ryanking13!

ryanking13:
Thanks for the reviews!

slackline:
Thanks @ryanking13 really appreciate you sorting this out so quickly :+1:

PR #4241, opened 2023-10-22
Add more keys in pyodide config CLI command

ryanking13:
### Description

This PR adds more keys to `pyodide config` command, so the build flags used in in-tree build can be also used in out-of-tree build easily.

1. rust related flags: `rustflags`, `rust_toolchain` (https://github.com/PyO3/pyo3/pull/3468)
2. config file paths for cmake, pyo3, and meson (#4240)
3. compiler flags (`cflags`, `cxxflags` and `ldflags`).

### Checklists


hoodmane (approved):
Thanks @ryanking13!

PR #4247, opened 2023-10-22
Switch PyProxy.toString from repr to str

### LINKED ISSUE #4040: Question: Should `toString` use `__repr__` or `__str__`?
In https://github.com/pyodide/pyodide/issues/4037#issuecomment-1665679197 @Duncanr-glitch said:

> `the toString` method using the ```__repr__``` method just seems incorrect since there's a distinction between ```__repr__``` and ```__str__``` that python devs expect and sometimes rely on. Using ```__repr__``` when doing a bare console.log makes since, but not when a dev is trying to string the object.

Looking at the history, `toString` was added here: https://github.com/pyodide/pyodide/commit/9579aa99bcfdecf65eb99311e6bb25fe8af831a9
I am not aware that we've ever discussed what this should do. I personally don't have a strong opinion. Interestingly, `toString` is missing from our proxy table, we should probably update the table in any case.
https://pyodide.org/en/stable/usage/type-conversions.html#proxying-from-python-into-javascript

@rth @ryanking13 @antocuni what do you all think?

hoodmane:
Resolves #4040.

- [x] Add / update tests

### DIFF
--- a/src/js/pyodide.ts
+++ b/src/js/pyodide.ts
@@ -402,7 +402,7 @@ export async function loadPyodide(
     throw Module.exited.toThrow;
   }
   if (options.pyproxyToStringRepr) {
-    Module.HEAP8[Module._compat_to_string_repr] = 1;
+    API.setPyProxyToStringMethod("repr");
   }
 
   if (API.version !== version) {
--- a/src/js/pyodide.ts
+++ b/src/js/pyodide.ts
@@ -315,6 +315,12 @@ export async function loadPyodide(
      */
     packages?: string[];
     /**
+     * Opt into the old behavior where PyProxy.toString calls `repr` and not
+     * `str`.
+     * @deprecated
+     */
+    pyproxyToStringRepr?: boolean;
+    /**
      * @ignore
      */
     _node_mounts?: string[];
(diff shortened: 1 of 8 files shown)


hoodmane:
Astonishingly, it looks like we didn't have a single unit test that cares about this??

hoodmane:
I added more string related entries to the `type-conversions.md`. There are a lot of things missing from `type-conversions.md`, I will hopefully go through and update it when I have a moment.

src/js/pyodide.ts
```diff
@@ -395,6 +401,9 @@ export async function loadPyodide(
   if (Module.exited) {
     throw Module.exited.toThrow;
   }
+  if (options.pyproxyToStringRepr) {
+    API.setPyProxyToStringMethod("repr");
```
ryanking13, line 405: Why are we entering a string value rather than a boolean value here, and converting it inside the function?

ryanking13 (approved):
Thanks!

src/js/pyodide.ts
hoodmane: Meant to say `true` here.

hoodmane: I'm surprised typescript is okay with this...

PR #4251, opened 2023-10-23
Use JsVal instead of JsRef in more places (nr. 5)

hoodmane:
Converts a lot more JsProxy methods to use JsVals.

### DIFF
--- a/src/core/jsproxy.c
+++ b/src/core/jsproxy.c
@@ -1224,21 +1224,26 @@ static PyMethodDef JsProxy_object_values_MethodDef = {
   METH_NOARGS,
 };
 
+#define ERR_NO_LENGTH -2
+#define ERR_NEGATIVE_LENGTH -3
+#define ERR_LENGTH_TOO_BIG -4
+
 EM_JS_NUM(int, get_length_helper, (JsVal val), {
   // clang-format off
   let result;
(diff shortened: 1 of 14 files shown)


src/core/jsproxy.c
```diff
@@ -1239,15 +1224,90 @@ static PyMethodDef JsProxy_object_values_MethodDef = {
...
+  if (result >= 0) {
+    return result;
+  }
+  // Something went wrong. Case work:
+  // * -1: Either `val.size` or `val.length` was a getter which managed to raise
+  //    an error. Rude. (Also we don't defend against this in hiwire_has_length)
+  // * -2: Doesn't have a length or size, or they aren't of type "number".
+  //   But `hiwire_has_length` returned true? So it must have changed somehow.
+  // * -3: Length was >= 2^{31}
+  // * -4: Length was negative
```
ryanking13, line 1273: I think the comment of `-3` and `-4` is reversed.

ryanking13 (approved):
Thanks, @hoodmane!

src/core/jsproxy.c
hoodmane: This is probably a good place to use `#define`s.

hoodmane:

PR #4252, opened 2023-10-23
Use JsVal instead of JsRef (nr. 6)

hoodmane:
This gets rid of a lot more uses of hiwire from JavaScript. It also finished eliminating the need for
a `js2python` conversion function that takes a `JsRef`, so I renamed the temporary `js2python_val`
back to `js2python`.


ryanking13 (approved):
Thanks @hoodmane! You're probably the only one who fully understands this low-level stuff, so I really appreciate your work.

hoodmane:
Thanks for the reviews!

Issue #4255, opened 2023-10-25
import type { PyProxy} from "pyodide/ffi" doesn't work with NodeNext


quassnoi:
## 🐛 Bug

```typescript
import type { PyProxy} from "pyodide/ffi"
```

doesn't work in TypeScript with `NodeNext` or `bundler` code generation (the default in Vite)

### To Reproduce

https://www.typescriptlang.org/play?target=99&moduleResolution=99&module=199#code/JYWwDg9gTgLgBDAnmApnA3nACorUIAeicAvnAGb4hwBEYiEAJsIygPTnnA0DcAUHwDGEAHYBnCABsUAOkkQA5gAp0ZAIZjsufEQCU-IA

Make sure to add pyodide to Plugins / Custom npm modules on the right hand side

### Expected behavior

No errors

### Environment

- Pyodide Version: 0.24.1
- Browser version:
- Any other relevant information:

### Additional context

Latest versions of TypeScript do honor the exports section of package.json and ffi is not listed there.

Add this to the module's `package.json/exports`:

```
"./ffi": {
      "types": "./ffi.d.ts"
    }
```

hoodmane:
Thanks for the report @quassnoi would you be willing to make a PR?

quassnoi:
I haven't tried to build it all the way, but when I copy my changes to the local node_modules folder, it works

hoodmane:
I am having trouble reproducing this locally, can you give me a local reproduction? If I do:
```bash
echo "import type { PyProxy } from "pyodide/ffi";" > a.ts
npm init -y
npm i pyodide
npx tsc a.ts
```
I don't get any errors.

quassnoi:
> I am having trouble reproducing this locally, can you give me a local reproduction? If I do:
> 
> ```shell
> echo "import type { PyProxy } from "pyodide/ffi";" > a.ts
> npm init -y
> npm i pyodide
> npx tsc a.ts
> ```
> 
> I don't get any errors.

Create a tsconfig.json with this content:

```
{
    "compilerOptions": {
        "module": "NodeNext"
    }
}
```

quassnoi:
Just for the reference, [this](https://www.typescriptlang.org/docs/handbook/modules/reference.html#packagejson-exports) is what it is all about:

> Note that the presence of "exports" prevents any subpaths not explicitly listed or matched by a pattern in "exports" from being resolved.

hoodmane:
Great that causes our test to fail.

[...]

PR #4256, opened 2023-10-26
import type { PyProxy } from "pyodide/ffi" doesn't work with NodeNext…

quassnoi:
… (#4255)

### Description

```
import type { PyProxy } from "pyodide/ffi"
```

 doesn't work with NodeNext, this PR fixes it.

### Checklists

- [x] Add / update tests


hoodmane:
Thanks @quassnoi. we attempted to cover this with the tests here:
https://github.com/pyodide/pyodide/blob/main/src/test-js/index.test-d.ts

I'll try to look into why that test failed to catch it and see if we can fix that as part of this PR.

quassnoi:
> Thanks @quassnoi. we attempted to cover this with the tests here: https://github.com/pyodide/pyodide/blob/main/src/test-js/index.test-d.ts
> 
> I'll try to look into why that test failed to catch it and see if we can fix that as part of this PR.

If there's an easy way to run just the tests on the JS part, let me know how to do that and I can take look at it tonight.

I was trying to run `make`, but after a half an hour's battle with automake, I gave up and automake won.

hoodmane:
Okay, this needs the merge changelog conflict resolved, but other than that it should hopefully be good to go.

docs/project/changelog.md
```diff
@@ -28,6 +28,9 @@ myst:
   `loadPyodide`, but this may be removed in the future.
   {pr}`4247`
 
+- {{ Fix }} `import type { PyProxy } from "pyodide/ffi"` doesn't work with NodeNext.
```
hoodmane, line 31: Maybe:
```suggestion
- {{ Fix }} `import type { PyProxy } from "pyodide/ffi"` now works with the `NodeNext` typescript target.
```

hoodmane:
Thanks @quassnoi!

PR #4262, opened 2023-10-28
Use JsVal instead of JsRef (nr. 8)

hoodmane:
This is close to finishing the refactor. I removed the last few uses of the `Hiwire` JS APIs from the PyProxy Buffer apis. I reworked all of the `JsProxy_create` family to use `JsVal` everywhere. I removed most of the remaining logic from `hiwire.c`. The only remaining places where `JsRef` is used are in struct fields.

After this, the last steps are to move the remaining constant initialization code from `hiwire.c` to `jslib.c`, and remove then remove our `hiwire.h` completely so we can use the `libhiwire.h` headers directly.


ryanking13 (approved):
Thanks!

PR #4263, opened 2023-10-28
FIX In python2js pass conversion context by pointer instead of by value

### LINKED ISSUE #4260: `dict_converter=Object.fromEntries` does not work with the default converter
## 🐛 Bug

This screenshot in the pyodide terminal (https://pyodide.org/en/stable/console.html) explains the bug:

Screenshot from 2023-10-27 11-29-33

Here is the full script to reproduce the error:

```python
from pyodide.ffi import to_js
from js import Object

# The default converter just returns a dictionary
def default_converter(obj, c, _): return c({"a": 2})

# Create a class that is not convertable to js
class A: pass

# Convert to js, converting dicts to maps
to_js(A, default_converter=default_converter) # This is ok

# Convert to js, converting dicts to objects
to_js(A, dict_converter=Object.fromEntries, default_converter=default_converter) # This is not ok
```

### Expected behavior

I would expect that it worked with `Object.fromEntries` as well.

### Environment

- Pyodide Version: Python 3.11.3 (main, Sep 25 2023 20:45:01) on WebAssembly/Emscripten (also on pyodide `0.24.1`).
- Browser version: Brave [Version 1.31.91 Chromium: 95.0.4638.69] and Chrome [Version 118.0.5993.88]

hoodmane:
This fixes a bug in `python2js`. `python2js_custom__create_jscontext` was violating the wasm ABI for struct value arguments. A struct argument is passed as a stack pointer, but it violates ABI to store that pointer or modify the memory. This caused undefined behavior.

Fixes #4260

- [x] Add / update tests

### DIFF
--- a/docs/project/changelog.md
+++ b/docs/project/changelog.md
@@ -33,6 +33,10 @@ myst:
   `AttributeError`.
   {pr}`4254`
 
+- {{ Fix }} Fixed a bug that occurs when using `toJs` with both `dictConverter`
+  and `defaultConverter` arguments.
+  {pr}`4263`
+
 ### Pyodide CLI
 
 - {{ Enhancement }} `pyodide config` command now show additional config variables:
--- a/src/core/python2js.c
+++ b/src/core/python2js.c
@@ -30,18 +30,18 @@ typedef struct ConversionContext_s
   int depth;
   JsRef proxies;
   JsRef jscontext;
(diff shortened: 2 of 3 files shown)


ryanking13 (approved):
LGTM, Thanks for fixing it!

hoodmane:
I think this still isn't really right but it is less wrong. We should always pass the context as a pointer and reset the depth manually when a recursive call returns.

PR #4266, opened 2023-10-29
Add docstrings for to_js

MatthewAndreTaylor:
Add docstring example usage for to_js.

js.Array.from from is an invalid keyword in Python instead js.Array.new to define a new Array

Small Python syntax edits in docstring.

Partially addresses: https://github.com/pyodide/pyodide/issues/1955

### DIFF
--- a/src/py/_pyodide/_core_docs.py
+++ b/src/py/_pyodide/_core_docs.py
@@ -1260,21 +1260,20 @@ def to_js(
     >>> js_object.hasOwnProperty("height")
     False
     
-    >>> js_object = pyodide.ffi.to_js({'age': 20, 'name': 'john'}, dict_converter=Map.new)
+    >>> js_object = to_js({'age': 20, 'name': 'john'}, dict_converter=Map.new)
     >>> js_object.keys(), js_object.values()
     KeysView([object Map]) ValuesView([object Map])
     >>> [(k, v) for k, v in zip(js_object.keys(), js_object.values())]
     [('age', 20), ('name', 'john')]
     
-    
-    >>> js_object = pyodide.ffi.to_js({'age': 20, 'name': 'john'}, dict_converter=Array.new)
+    >>> js_object = to_js({'age': 20, 'name': 'john'}, dict_converter=Array.new)
     >>> [item for item in js_object]
     [age,20, name,john]
     >>> js_object.toString()
     age,20,name,john
 
     >>> class Bird: pass
-    >>> converter = lambda value, cache, _: Object.new(size=1, color='red') if isinstance(value, Bird) else None
+    >>> converter = lambda value, convert, cache: Object.new(size=1, color='red') if isinstance(value, Bird) else None
     >>> js_nest = to_js([Bird(), Bird()], default_converter=converter)
     >>> [bird for bird in js_nest]
     [[object Object], [object Object]]
--- a/src/py/_pyodide/_core_docs.py
+++ b/src/py/_pyodide/_core_docs.py
@@ -1226,7 +1226,7 @@ def to_js(
         result of the dict conversion. Some suggested values for this argument:
 
           * ``js.Map.new`` -- similar to the default behavior
-          * ``js.Array.new`` -- convert to an array of entries
+          * ``js.Array.from`` -- convert to an array of entries
           * ``js.Object.fromEntries`` -- convert to a JavaScript object
 
     default_converter:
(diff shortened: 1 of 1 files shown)


src/py/_pyodide/_core_docs.py
```diff
@@ -1226,7 +1226,7 @@ def to_js(
         result of the dict conversion. Some suggested values for this argument:
 
           * ``js.Map.new`` -- similar to the default behavior
-          * ``js.Array.from`` -- convert to an array of entries
+          * ``js.Array.new`` -- convert to an array of entries
```
hoodmane, line 1229: we want `Array.from_` here. `Array.new([1,2,3])` returns `[[1,2,3]]`, whereas `Array.from` returns `[1,2,3]`. You can access `Array.from` as `from_` in Python.

hoodmane:
Thanks for helping out with this @MatthewAndreTaylor! I made some initial comments.

src/py/_pyodide/_core_docs.py
```diff
@@ -1238,6 +1238,39 @@ def to_js(
 
     Examples
     --------
+    >>> from js import Object, Map, Array
+    >>> js_object = pyodide.ffi.to_js({'age': 20, 'name': 'john'}, dict_converter=Object.fromEntries)
```
hoodmane, line 1242: Maybe do `from pyodide.ffi import to_js` and then use unqualified `to_js`?

src/py/_pyodide/_core_docs.py
```diff
@@ -1238,6 +1238,39 @@ def to_js(
...
+    >>> js_object.hasOwnProperty("age")
+    True
+    >>> js_object.hasOwnProperty("height")
+    False
+
+    >>> js_object = pyodide.ffi.to_js({'age': 20, 'name': 'john'}, dict_converter=Map.new)
+    >>> js_object.keys(), js_object.values()
+    KeysView([object Map]) ValuesView([object Map])
+    >>> [(k, v) for k, v in zip(js_object.keys(), js_object.values())]
+    [('age', 20), ('name', 'john')]
```
hoodmane, line 1258: Since this is the default behavior, maybe put this example first and leave off `dict_converter`? Also, what about having a `>>> js_object` after we assign to it?

src/py/_pyodide/_core_docs.py
```diff
@@ -1238,6 +1238,39 @@ def to_js(
 
     Examples
     --------
+    >>> from js import Object, Map, Array
+    >>> js_object = pyodide.ffi.to_js({'age': 20, 'name': 'john'}, dict_converter=Object.fromEntries)
+    >>> js_object.age == 20
+    True
+    >>> js_object.name == 'john'
+    True
+    >>> js_object.toString()
```
hoodmane, line 1247:
```suggestion
    >>> js_object
```

src/py/_pyodide/_core_docs.py
```diff
@@ -1272,7 +1305,7 @@ def default_converter(_value, _ignored1, _ignored2):
 
         class Pair:
             def __init__(self, first, second):
-                self.first = first self.second = second
+                self.first, self.second = first, second
```
hoodmane, line 1308: Thanks for fixing this! I think it should read:
```suggestion
                self.first = first
                self.second = second
```

src/py/_pyodide/_core_docs.py
MatthewAndreTaylor: Yes I played around with the behavior and you are right the result is a nested Array in Javascript

src/py/_pyodide/_core_docs.py
MatthewAndreTaylor: Removed the param dict_converter=Map.new as it is the default behavior

src/py/_pyodide/_core_docs.py
hoodmane: Could you split the assignment back onto two lines?

hoodmane (approved):
Looks good, thanks @MatthewAndreTaylor. Please split `self.first, self.second = first, second` into two separate lines and then the changes look good.

@ryanking13 do you know how to prevent pytest from running the examples?

MatthewAndreTaylor:
Yep made the change 👍

ryanking13:
> @ryanking13 do you know how to prevent pytest from running the examples?

I think the simplest way is to put `# doctest: +SKIP` on each line.

https://github.com/pyodide/pyodide/blob/299c7bec1c610ab4625bbed90a979dc267403dce/src/py/_pyodide/_core_docs.py#L144-L154

MatthewAndreTaylor:
> > @ryanking13 do you know how to prevent pytest from running the examples?
> 
> I think the simplest way is to put `# doctest: +SKIP` on each line. Example:
> 
> https://github.com/pyodide/pyodide/blob/299c7bec1c610ab4625bbed90a979dc267403dce/src/py/_pyodide/_core_docs.py#L144-L154

I have added `# doctest: +SKIP` on each line as you showed

src/py/_pyodide/_core_docs.py
```diff
@@ -1238,6 +1238,41 @@ def to_js(
 
     Examples
     --------
+    >>> from js import Object, Map, Array
+    >>> from pyodide.ffi import to_js
```
ryanking13, line 1242: I think you need to put `# doctest: +SKIP` in the imports too.

ryanking13 (approved):
Thanks @MatthewAndreTaylor!

PR #4269, opened 2023-10-30
Drop node < 18 support

### LINKED ISSUE #4148: [Discussion] Drop support for Node versions that passed endoflife
## Proposal

Drop support for Node < 18 from the next Pyodide major release (0.25.0)

## Background

The endoflife of [Node.js 14 and 16 passed a few days ago](https://nodejs.dev/en/about/releases/). We just finished a major release, so I think this is a good time to talk about Node version support.

image

Here are the things that are currently associated with Node.js versions in Pyodide.

- The documentation mentions [how to run Pyodide in Node.js < 0.18](https://pyodide.org/en/stable/usage/index.html#node-js-versions-0-17)
- We have some compat codes for older Node versions
  - https://github.com/pyodide/pyodide/pull/4100#discussion_r1308449699
  -[compat.ts](https://github.com/pyodide/pyodide/blob/e2c2884e7a9afd90b23849220ee578bd2d523458/src/js/compat.ts#L40)
- JSPI requires Node >= 20
- The Docker image used for Pyodide CI uses Node 20

### Benefits from dropping Node < 18 support

- No more `node-fetch`.
- No more `--experimental-wasm-bigint` flags
- Better `MessageChannel` support (https://github.com/pyodide/pyodide/issues/4006)
- Some useful methods like [`AbortController`](https://developer.mozilla.org/en-US/docs/Web/API/AbortController)

### Drawbacks

- People who were using Node.js < 18 with Pyodide won't like it.

## Migration plan

- Mention minimal Node.js version support in docs
- Remove compat codes for old  Node.js versions
  - pytest-pyodide?

WDYT?

ryanking13:
### Description

Close #4148 

### Checklists

- [x] Add / update tests


hoodmane:
There's also `src/templates/python`

ryanking13:
> There's also src/templates/python

Thanks! I updated that file too. I don't fully understand that script, but according to your comment, maybe we don't need that complex polyglot script anymore? Anyway, I think we don't need to handle it in this PR.

src/templates/python
```diff
@@ -2,6 +2,8 @@
 ":" /* << "EOF"
 This file is a bash/node polyglot. This is needed for a few reasons:
 
+TODO: We don't support node < 18 anymore, so maybe we can remove this?
```
hoodmane, line 5: I think that we'll want to use this same mechanism for the `--experimental-wasm-stack-switching` flag soon, so I think it's fine to leave it.

rth (approved):
Thanks!

PR #4271, opened 2023-10-31
NFC hiwire cleanup

hoodmane:
Move remaining hiwire code to jslib and clean up headers


ryanking13 (approved):
Thanks @hoodmane! It looks like `build-core-nodylink` job is failing otherwise LGTM.

Issue #4051, opened 2023-08-08
Cannot .toJs() a numpy.ndarray containing BigInt values - Error: Unrecognized format character 'O'


paulcalcraft:
## 🐛 Bug

Running `.toJs()` on a PyProxy of a numpy.ndarray containing BigInt throws an exception: "Error: Unrecognized format character 'O'"

### To Reproduce

```html
<!doctype html>
<html>
  <head>
      <script src="https://cdn.jsdelivr.net/pyodide/v0.23.4/full/pyodide.js"></script>
  </head>
  <body>
    NumPy BigInt toJs() error<br>
    Open your browser console to see Pyodide output
    <script type="text/javascript">
      async function main(){
        let pyodide = await loadPyodide();
        console.log(pyodide.runPython(`
            import sys
            sys.version
        `));
        await pyodide.loadPackage("numpy");
        let ndarray = await pyodide.runPython(`import numpy as np

# Creating a numpy array with Python's native integers
bigints = [10**100, 10**200, 10**300]
arr = np.array(bigints, dtype=object)

# Output the array
arr`);
        console.log(ndarray.type); // numpy.ndarray
        ndarray.toJs(); // PythonError: pyodide.ffi.JsException: Error: Unrecognized format character 'O'.
      }
      main();
    </script>
  </body>
</html>
```

```
Uncaught (in promise) PythonError: pyodide.ffi.JsException: Error: Unrecognized format character 'O'.

The above exception was the direct cause of the following exception:

pyodide.ffi.ConversionError: Conversion from python to javascript failed

    at new_error (pyodide.asm.js:9:14992)
    at pyodide.asm.wasm:0x152d67
    at pyodide.asm.wasm:0x152e6c
    at Module._pythonexc2js (pyodide.asm.js:9:656029)
    at Proxy.toJs (pyodide.asm.js:9:77512)
    at main (bigint_pyodide_error_repro.html:24:7)
new_error @ pyodide.asm.js:9
$wrap_exception @ pyodide.asm.wasm:0x152d67
$pythonexc2js @ pyodide.asm.wasm:0x152e6c
Module._pythonexc2js @ pyodide.asm.js:9
toJs @ pyodide.asm.js:9
main @ bigint_pyodide_error_repro.html:24
await in main (async)
(anonymous) @ bigint_pyodide_error_repro.html:26
```

### Expected behavior

[...]

PR #4274, opened 2023-11-01
Changed to sequence check

MatthewAndreTaylor:
Changed _python2js_deep

Sequence types for example numpy arrays of objects are treated as buffers.
Sequences in Python are Iterable and have len (sq_length) and get (sq_item).
By checking that the PyObject* passed in is not a dictionary first only valid sequences will be converted as expected.

`// One might consider trying to convert things that satisfy PyMapping_Check to
// maps and things that satisfy PySequence_Check to lists. However
// PyMapping_Check "returns 1 for Python classes with a __getitem__() method"
// and PySequence_Check returns 1 for classes with a __getitem__ method that
// don't subclass dict. For this reason, I think we should stick to subclasses.`

However, by only checking PySequence_Check and checking PyDict_Check before ensures this is not the case.

Addresses issue: https://github.com/pyodide/pyodide/issues/4051

### DIFF
--- a/src/core/python2js.c
+++ b/src/core/python2js.c
@@ -379,14 +379,14 @@ _python2js_deep(ConversionContext* context, PyObject* x)
 {
   RETURN_IF_HAS_VALUE(_python2js_immutable(x));
   RETURN_IF_HAS_VALUE(_python2js_proxy(x));
-  if (PyList_Check(x) || PyTuple_Check(x)) {
-    return _python2js_sequence(context, x);
-  }
   if (PyDict_Check(x)) {
     return _python2js_dict(context, x);
   }
   if (PySet_Check(x)) {
     return _python2js_set(context, x);
+  }
+  if (PySequence_Check(x)) {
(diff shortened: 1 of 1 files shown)


hoodmane:
This is an interesting suggestion. I wonder why I originally put in the stricter check, I'll try and look at the git blame. We'll need a test and a change log entry.

hoodmane:
Thanks @MatthewAndreTaylor!

src/core/python2js.c
```diff
@@ -379,15 +379,15 @@ _python2js_deep(ConversionContext* context, PyObject* x)
...
-  }
   if (PyDict_Check(x)) {
     return _python2js_dict(context, x);
   }
   if (PySet_Check(x)) {
     return _python2js_set(context, x);
   }
+  if (PySequence_Check(x)) {
+    return _python2js_sequence(context, x);
+  }
```
hoodmane, line 390: You'll have to move this after the `PyObject_CheckBuffer` to fix the test failures.

Issue #4249, opened 2023-10-23
New fatal errors in scipy tests on `main` branch


lesteve:
## 🐛 Bug

There are new tests with fatal errors in the scipy tests from https://github.com/lesteve/scipy-tests-pyodide.

Last working commit (2 days ago): 408bffb7d22f0fef261b64007bdcde665f2b1e6e
First broken commit (yesterday): 429dcd4b1f3676e71f58c06770b7c98186572981

```
❯ git l 408bffb7..429dcd4b 
* 429dcd4b - Use externrefs for some ffi code (2) (#4226) (2 days ago) <Hood Chatham>
* 2d55a3e7 - xfail syncify_not_supported in node (2 days ago) <Hood Chatham>
* 7a7691c2 - Make jspi work correctly with nodylink builds (2 days ago) <Hood Chatham>
* 4bafdafc - Update JSPI patch for Emscripten 3.1.45 (2 days ago) <Hood Chatham>
* 6117d7c9 - Stack switching (#3210) (2 days ago) <Hood Chatham>
* d97eaa58 - Use branch in macro to choose trampoline (#3990) (2 days ago) <Hood Chatham>
* 31c1345a - Move stack switching logic into a new stack switching folder. (#3987) (2 days ago) <Hood Chatham>
* c931b06a - Patch CPython to use a type reflection trampoline if possible (#3964) (2 days ago) <Hood Chatham>
* 3085cb9d - Use Wasm Exceptions for invoke stubs if they exist (#3957) (2 days ago) <Hood Chatham>
```

Full log is [here](https://github.com/lesteve/scipy-tests-pyodide/actions/runs/6601590374/job/17932770264). Error seem similar to the ones in https://github.com/pyodide/pyodide/issues/4140 when the JSPI changes were reverted.

Here are the submodules with issue (below are the stacktraces from the log):
```
--------------------------------------------------------------------------------
Unexpected test results
--------------------------------------------------------------------------------
scipy.fft.tests result expected in ['passed'], got 'fatal error or timeout' instead
scipy.optimize.tests result expected in ['failed'], got 'fatal error or timeout' instead
scipy.sparse.linalg._dsolve.tests result expected in ['passed'], got 'fatal error or timeout' instead
scipy.special.tests result expected in ['failed'], got 'fatal error or timeout' instead
scipy.stats.tests result expected in ['failed'], got 'fatal error or timeout' instead
```

### scipy.fft.tests

```
The cause of the fatal error was:
[...]

PR #4276, opened 2023-11-01
Use meson build backend for scipy

ryanking13:
### Description

Trying to use meson build backend when building scipy.

Close #3649


hoodmane:
Thanks so much for your work on meson @ryanking13!

ryanking13:
Okay, I think now CI passes.

ryanking13:
cc: @lesteve

I hope not, but this may break something. Please let me know if something goes wrong after merging this PR.

lesteve:
Sounds good, currently Scipy is a bit broken on `main` (or at least a bit more broken than previously) due to https://github.com/pyodide/pyodide/issues/4249 but I'll let you know in case I notice additional issues.

lesteve:
Side-comment, it would probably be worth doing a PR on conda-forge/pkg-config-feedstock to bump the build number and have a package available with the fix from https://github.com/conda-forge/pkg-config-feedstock/pull/31

lesteve:
> Please let me know if something goes wrong after merging this PR.

FYI I ran the scipy tests locally on this PR and they pass, i.e. this PR doesn't break anything else than what is already broken in `main`. I had a look at the changes and they seem fine.

All in all, I think this can be merged!

ryanking13:
Oh, thanks so much for the check @lesteve!

pyodide-build/pyodide_build/_f2c_fixes.py
```diff
@@ -487,7 +487,14 @@ def scipy_fix_cfile(path: str) -> None:
 
     for lib in ["lapack", "blas"]:
         if path.endswith(f"cython_{lib}.c"):
-            header_path = Path(path).with_name(f"_{lib}_subroutines.h")
+            header_name = f"_{lib}_subroutines.h"
```
hoodmane, line 490: Maybe put this in a separate `find_header` function?

pyodide-build/pyodide_build/pywasmcross.py
```diff
@@ -672,12 +678,9 @@ def handle_command(
...
 
     if line[0] == "gfortran":
-        if "-dumpversion" in line:
-            sys.exit(subprocess.run(line).returncode)
         tmp = replay_f2c(line)
-        if tmp is None:
-            sys.exit(0)
-        line = tmp
+        if tmp is not None:
+            line = tmp
```
hoodmane, line 683: Can you explain this change? In particular we used to quit out here if `replay_f2c` returns `None` but now we don't? Why? When does this happen? What is the subprocess call below going to do in this case?

Also, we should switch `handle_command` to return an `int` since in `compile_main` it says `sys.exit(handle_command(args, build_args))` we can just `return 0` or `return returncode` in here.

pyodide-build/pyodide_build/pywasmcross.py
```diff
@@ -165,7 +165,6 @@ def replay_f2c(args: list[str], dryrun: bool = False) -> list[str] | None:
         found_source = True
 
     if not found_source:
-        print(f"f2c: source not found, skipping: {new_args_str}")
```
hoodmane, line 168: Does this happen? Is there a reason we don't want it to log this?

hoodmane (approved):
Thanks @ryanking13!

pyodide-build/pyodide_build/_f2c_fixes.py
hoodmane: Or maybe this is fine as is...

pyodide-build/pyodide_build/pywasmcross.py
ryanking13: During the setup step, meson calls `gfortran` (and other compilers as well) multiple times to test the compiler, like checking versions, compiling sample codes, etc. So yes, this can happen.

I removed the log because it was preventing meson from parsing the value from the result of gfortran's execution.

pyodide-build/pyodide_build/pywasmcross.py
ryanking13: As mentioned in another comment, meson runs gfortran multiple times during the compiler check phase, so sometimes replay_f2c doesn't do anything. So I changed it to run gfortran normally in that case.

pyodide-build/pyodide_build/_f2c_fixes.py
ryanking13: I moved it to a separate function. Thanks for the suggestion!

pyodide-build/pyodide_build/pywasmcross.py
ryanking13: > Also, we should switch handle_command to return an int since in compile_main it says sys.exit(handle_command(args, build_args)) we can just return 0 or return returncode in here.

Right, thanks for the pointer.

pyodide-build/pyodide_build/tests/test_build_env.py
```diff
@@ -60,9 +60,7 @@ def test_in_xbuildenv(self, reset_env_vars, reset_cache):
 
     def test_get_build_environment_vars(self, reset_env_vars, reset_cache):
         build_vars = build_env.get_build_environment_vars()
-        extra_vars = set(
-            ["PYODIDE", "PKG_CONFIG_PATH", "CMAKE_TOOLCHAIN_FILE", "PYO3_CONFIG_FILE"]
-        )
+        extra_vars = set(["PYODIDE", "CMAKE_TOOLCHAIN_FILE", "PYO3_CONFIG_FILE"])
```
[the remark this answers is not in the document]
ryanking13: These variables are for checking extra variables that does not come from Makefile.envs but are added [from the code](https://github.com/pyodide/pyodide/blob/ca6d63388c5365fda8f92b3d2f16672d57b9beef/pyodide-build/pyodide_build/build_env.py#L171-L184) . So PKG_CONFIG_PATH or PKG_CONFIG_LIBDIR does not need to belong there.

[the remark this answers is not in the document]
hoodmane: Could you add a comment like:
```py
# extra variables that does not come from Makefile.envs but are added by build_env.py
```

pyodide-build/pyodide_build/pywasmcross.py
hoodmane: Would the following work:
```
tmp = replay_f2c(line)
if tmp is None:
   # No source file, it's a query for information about the compiler. Pretend we're
   # gfortran by letting gfortran handle it
   return subprocess.run(line).returncode
line = tmp
```

Or is it important that we are calling `handle_command_generate_args` on it? I think it would help to have some comments here since this is a really confusing execution path.

ryanking13: I think your code will work. Thanks for the suggestion. 

> since this is a really confusing execution path.

I agree. Actually `pywasmcross` is becoming more complicated as we add more symlinks... I'll try do to some refactorings when I have bandwidth.

PR #4280, opened 2023-11-03
MAINT Make doctests run; run some doctests in pyodide 

hoodmane:
This uses https://github.com/pyodide/pytest-pyodide/pull/117 to run doctests in
Pyodide. I also turned on and fixed various doctests that were not working for
unrelated reasons


src/py/_pyodide/_base.py
```diff
@@ -468,6 +468,8 @@ def eval_code(
     >>> source = "1 + 1"
     >>> eval_code(source)
     2
+    >>> from js import Object; Object
+    None
```
hoodmane, line 472: This is intentionally failing...

rth:
Maybe this is a feature that should rather go to `pytest-pyodide`? I would say the most logical would be to run a doctest in pyodide if it is tagged as such (same as a regular test with pytest-pyodide). Maybe,
```py
>>> from js import fetch     #  doctest: +RUN_IN_PYODIDE
```
where `RUN_IN_PYODIDE` is a tag similar to `# doctest: +NORMALIZE_WHITESPACE`. From purely a user perspective, but I don't know what it would take to implement this, what parses those tags, and whether it's easy to add custom tags there.

Also maybe for pyodide, it would not be the most practical if all doctests should run there. In which case allowing for this 
option to be specified in pyproject as,
```toml
[tool.pytest]
doctest_optionflags = [
    "NORMALIZE_WHITESPACE",
    "RUN_IN_PYODIDE"
]
```
would help but again not sure what it would take.

So a PR with smaller scope as you did here is also fine.

hoodmane:
> Maybe this is a feature that should rather go to pytest-pyodide? 

Yes, that would make sense.

> I would say the most logical would be to run a doctest in pyodide if it is tagged as such (same as a regular test with pytest-pyodide). Maybe,
> ```
> >>> from js import fetch     #  doctest: +RUN_IN_PYODIDE
> ```
> where `RUN_IN_PYODIDE` is a tag similar to `# doctest: +NORMALIZE_WHITESPACE`.

I think it would be pretty easy for us to add our own tags since they are visible as inputs to the `DocTest.run` function that I've monkeypatched. One question though: I would like this to apply to the entire test, but I think the meaning of `doctest:` directives is ordinarily scoped to a single command. But I think we can just give it different semantics if we prefer.

ryanking13:
Thanks for working on this! I agree that this should go to `pytest-pyodide`.

> One question though: I would like this to apply to the entire test, but I think the meaning of doctest: directives is ordinarily scoped to a single command. But I think we can just give it different semantics if we prefer.

I've never actually used the doctest module myself, so I don't know how complicated it is, but +1 to give it different semantics.

PR #4281, opened 2023-11-03
NFC Move some logic from pyodide.ts to api.ts

hoodmane:
I think this is a bit tidier.

### DIFF
--- a/src/js/types.ts
+++ b/src/js/types.ts
@@ -244,7 +244,7 @@ export interface Module {
   API: API;
   postRun: ((a: Module) => void) | ((a: Module) => void)[];
   locateFile: (file: string) => string;
-  exited?: {toThrow: any};
+  exited?: { toThrow: any };
   ENV: { [key: string]: string };
   PATH: any;
   TTY: any;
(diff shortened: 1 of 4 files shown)


ryanking13:
Hmm, it is strange that the core test fails in GHA but not in Circle CI.

src/js/types.ts
```diff
@@ -238,8 +238,13 @@ export interface Module {
   noWasmDecoding: boolean;
   quit: (status: number, toThrow: Error) => void;
   preRun: { (): void }[];
-  print: (a: string) => void;
-  printErr: (a: string) => void;
+  print?: (a: string) => void;
+  printErr?: (a: string) => void;
+  arguments: string[];
+  API: API;
```
ryanking13, line 244: Maybe we need to separate methods that comes from emscripten and methods that are added by Pyodide ( no action required for this PR ).

ryanking13 (approved):
Thanks!

src/js/types.ts
hoodmane: Also, we should move away from Pyodide adding things as much as possible.

Issue #4232, opened 2023-10-17
Default indexURL is wrong in node?


hoodmane:
## 🐛 Bug

Install Pyodide into node and do:
```js
const {loadPyodide} = await import("pyodide/pyodide.mjs")
const py = await loadPyodide();
```
raises:

Error [ERR_MODULE_NOT_FOUND]: Cannot find module '/path/to/current/folder/file:/path/to/current/folder/node_modules/pyodide/pyodide.asm.js'

rth:
Yes, I think we broke something, likely in https://github.com/pyodide/pyodide/pull/3967. We might need to re-think again what's a reasonable behavior there in node vs browser.

Also possibly same issue in https://github.com/pyodide/pyodide/issues/4174

PR #4288, opened 2023-11-09
Fix default index url calculation in node ES module

ryanking13:
### Description

Resolve https://github.com/pyodide/pyodide/issues/4232

This fixes the default index URL calculation bug in the node environment when using ES modules.

The bug happens because [`__dirname` is not available in ESM module](https://blog.logrocket.com/alternatives-dirname-node-js-es-modules/), hence `calculateIndexURL` returns the incorrect path.

So I used `import.meta.url` to calculate the dirname when the environment is ESM. It looks like `import.meta.dirname` is coming to node (https://github.com/nodejs/node/pull/48740) soon, but we can't use it yet.

### Checklists

- [x] Add / update tests

### DIFF
--- a/src/js/compat.ts
+++ b/src/js/compat.ts
@@ -283,13 +283,6 @@ export async function calculateDirname(): Promise<string> {
     return __dirname;
   }
 
-  if (IN_NODE_ESM) {
-    const nodePath = await import("path");
-    const nodeUrl = await import("url");
-
-    return nodeUrl.fileURLToPath(nodePath.dirname(import.meta.url));
-  }
-
   let err: Error;
   try {
     throw new Error();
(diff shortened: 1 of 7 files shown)


src/js/compat.ts
```diff
@@ -274,3 +273,35 @@ export async function loadLockFile(lockFileURL: string): Promise<any> {
...
+export async function calculateDirname(): Promise<string> {
+  if (IN_NODE_COMMONJS) {
+    return __dirname;
+  }
+
+  if (IN_NODE_ESM) {
+    const nodePath = await import("path");
+    const nodeUrl = await import("url");
+
+    return nodeUrl.fileURLToPath(nodePath.dirname(import.meta.url));
```
ryanking13, line 290: This part is the main change.

src/js/environments.ts
hoodmane, line 1: Good idea moving this stuff into a different file.

src/js/compat.ts
```diff
@@ -274,3 +273,35 @@ export async function loadLockFile(lockFileURL: string): Promise<any> {
...
+ * This is used to guess the indexURL when it is not provided.
+ */
+export async function calculateDirname(): Promise<string> {
+  if (IN_NODE_COMMONJS) {
+    return __dirname;
+  }
+
+  if (IN_NODE_ESM) {
+    const nodePath = await import("path");
+    const nodeUrl = await import("url");
```
hoodmane, line 288: Do we need webpack ignores here or other nonsense like that?

hoodmane (approved):
Thanks @ryanking13!

src/js/compat.ts
ryanking13: Good point. I don't think it is required to use `/* webpackIgnore */` for builtin modules.

ryanking13:
> So I used import.meta.url to calculate the dirname when the environment is ESM. It looks like import.meta.dirname is coming to node (https://github.com/nodejs/node/pull/48740) soon, but we can't use it yet.

After struggling with `mocha + typescript + ESM` for a few hours, I gave up and replaced `import.meta` with the result of `ErrorStackParser`.

Issue #4004, opened 2023-07-20
WebpackIgnore comments disappear when using esbuild instead of rollup


Kol96:
### To Reproduce

webpackIgnore comments disappear  from the development JavaScript bundle
bad case: https://cdn.jsdelivr.net/pyodide/dev/full/pyodide.js
good case: https://cdn.jsdelivr.net/npm/pyodide@0.23.4/pyodide.js

### Expected behavior

keep webpackIgnore comments
`await import(/* webpackIgnore: true */url)`

### Environment

- Pyodide Version: latest
- Browser version:
- Any other relevant information:

rth:
Are you using [`pyodide-webpack-plugin`](https://github.com/pyodide/pyodide-webpack-plugin)? That's the recommended way of using webpack with Pyodide cc @mneil 

But otherwise yes, sounds like a bug, thanks for the report.

Kol96:
> 



> Are you using [`pyodide-webpack-plugin`](https://github.com/pyodide/pyodide-webpack-plugin)? That's the recommended way of using webpack with Pyodide cc @mneil
> 
> But otherwise yes, sounds like a bug, thanks for the report.

It seems that using string-replace-loader in webpack can solve my problem, I ultimately opted for directly importing pyodide.js using cdn.
`import(/* webpackIgnore: true */ "https://cdn.jsdelivr.net/pyodide/dev/full/pyodide.js");`

By the way, esbuild may not be able to solve this problem according to this [issue](https://github.com/evanw/esbuild/issues/221)

hoodmane:
We can probably fix this with the legal comments option in esbuild by setting legalComments "inline", writing `/*! webpackIgnore: true */` and then using sed to remove the exclamation point.

Kol96:
> We can probably fix this with the legal comments option in esbuild by setting legalComments "inline", writing `/*! webpackIgnore: true */` and then using sed to remove the exclamation point.

It didn't work, legal comments only works on the top of the file. Maybe just match `import(` and replace it with `import(/* webpackIgnore: true */ ` after esbuild done.

From esbuild doc:
Note that "statement-level" for JS and "rule-level" for CSS means the comment must appear in a context where multiple statements or rules are allowed such as in the top-level scope or in a statement or rule block. So comments inside expressions or at the declaration level are not considered legal comments.

hoodmane:
> Or leave it to webpack plugin, casue there are more than one bundlers, pyodide cannot be adapted for each tools.

[...]

PR #4294, opened 2023-11-14
Keep webpackIgnore comments in pyodide.js

### LINKED ISSUE #4174: Cannot find module 'https://cdn.jsdelivr.net/pyodide/dev/full/pyodide.asm.js'
Hi all. I have the following problem: I'm trying to use pyodide in a react. I installed the package using npm install, and imported the loadPyodide function from it. I have this code:
```
             indexURL: "https://cdn.jsdelivr.net/pyodide/v0.24.0/full/",
             fullStdLib: true,
             stdin: window.prompt,
             stdout: (text) => {
                 if (text !== "Python initialization complete") {
                     setResult(prevState => prevState + text + "\n")
                 }
             },
             stderr: (text) => {
                 setResult(prevState => prevState + text + "\n")
             }
         });
```
However I get an error:
```
Cannot find module 'https://cdn.jsdelivr.net/pyodide/dev/full/pyodide.asm.js'
     at http://localhost:3000/static/js/bundle.js:277779:11
     at async loadScript (http://localhost:3000/static/js/bundle.js:317327:43)
     at async loadPyodide (http://localhost:3000/static/js/bundle.js:317657:5)
     at async loadPython (http://localhost:3000/static/js/bundle.js:41685:21)
     at async run (http://localhost:3000/static/js/bundle.js:41761:24)
```
How can you solve it? I would appreciate your answer.

ryanking13:
### Description

Resolve #4174

After we changed our build system to esbuild (https://github.com/pyodide/pyodide/pull/3679), `/* webpackIgnore */` comments were stripped out after the build.

Since we dynamically import pyodide.asm.js from pyodide.[m]js, this comment is required for webpack bundler (unless we find a better way to deal with this... like merging pyodide.js and pyodide.mjs? I don't know.)

This PR fixes it by putting webpackIgnore comments for every `import()` statement after building pyodide.js.
A user can do this by adding a webpack plugin, as described in [pyodide-webpack-plugin](https://github.com/pyodide/pyodide-webpack-plugin/blob/9112d85cfe45325d43b73301a10c1202f1570f78/README.md?plain=1#L100-L119), but I think it is better if we can handle it.

Related:

- https://github.com/pyodide/pyodide/issues/4004
- https://github.com/pyodide/pyodide/issues/4174
- https://github.com/evanw/esbuild/issues/2721: esbuild added support for preserving magic comments, but it is not compatible with some `minify` flags.

### Checklists

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -17,6 +17,7 @@ __pycache__
 .pytest_cache/
 .vscode
 .venv*
+tsconfig.tsbuildinfo
 
 build
 ccache
--- a/Makefile
+++ b/Makefile
@@ -133,7 +133,7 @@ node_modules/.installed : src/js/package.json src/js/package-lock.json
 	touch node_modules/.installed
 
 dist/pyodide.js src/js/generated/_pyodide.out.js: src/js/*.ts src/js/generated/pyproxy.ts node_modules/.installed
-	cd src/js && npm run tsc && node esbuild.config.mjs && cd -
+	cd src/js && npm run build && cd -
 
 src/core/stack_switching/stack_switching.out.js : src/core/stack_switching/*.mjs
 	node src/core/stack_switching/esbuild.config.mjs
--- a/docs/project/changelog.md
+++ b/docs/project/changelog.md
@@ -23,6 +23,9 @@ myst:
 - {{ Fix }} Fixed default indexURL calculation in Node.js environment.
   {pr}`4288`
 
+- {{ Fix }} Fixed a bug that webpack messes up dynamic import of `pyodide.asm.js`.
+  {pr}`4294`
+
 - {{ Enhancement }} Added experimental support for stack switching.
   {pr}`3957`, {pr}`3964`, {pr}`3987`, {pr}`3990`, {pr}`3210`
 
--- a/src/js/esbuild.config.mjs
+++ b/src/js/esbuild.config.mjs
@@ -96,7 +96,14 @@ try {
     );
 
     // inject webpackIgnore comments
+<<<<<<< Updated upstream
     content = content.replace("import(", "import(/* webpackIgnore */");
+=======
+    content = content.replaceAll(
+      "import(",
+      "import(/* webpackIgnore */",
+    );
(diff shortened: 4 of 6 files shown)


hoodmane:
@mneil What's your thought on `webpackIgnore` comments? Do you think we should just drop them and encourage people to use your plugin?

mneil:
My opinion is that the library should not make any assumptions about how it will be consumed. Users choosing to bundle with webpack need to understand the implications of choosing that tool and how to work with it. Rollup, Turbopack, esbuild, Webpack, and other tools exist to do the same thing. I'm not sure pyodide should attempt to do anything specifically for one or more of these tools. 

The plugin now has a custom loader that skips parsing by webpack so that it doesn't import anything. It's close to adding the ignore comments on the imports. It also handles copying the necessary dependencies like pyodide.asm.js and other files. Obviously I'm biased in favor of the plugin :)

hoodmane:
> close to adding the ignore comments on the imports

Meaning it doesn't do this yet? So maybe we should leave them in for now and then drop all of these once the plugin adds support for this?

mneil:
No, I'm sorry for the confusion. What I meant by "close to adding" is that the functionality is similar, but it is accomplished in a different way. It does not need to add those comments because it doesn't go through webpack's javascript parser anymore. 

The comments are not necessary for the plugin to work correctly.

hoodmane:
Ok then I vote to drop these because they are a wart. If we want to provide webpack-specific support, we should focus on improving the discoverability of the plugin.

ryanking13:
Thanks for the info! I agree that the handling of bundler-specific things should be handled in the plugin. Maybe let's include this to 0.25.0 for now, and find a way to improve the discoverability of the plugin + test the plugin in the Pyodide repository.

hoodmane:
> include this to 0.25.0 for now, and find a way to improve the discoverability of the plugin + test the plugin in the Pyodide repository

Should we open one or more followup issues for these tasks?

ryanking13:
I opened a followup issue (https://github.com/pyodide/pyodide/issues/4361)

PR #4296, opened 2023-11-15
Fix static_import test to work in safari

ryanking13:
### Description

The `test_static_import` test [was failing in Safari](https://app.circleci.com/pipelines/github/pyodide/pyodide/6470/workflows/a3dbffc5-90bd-4ffe-9aed-065502cfe14f/jobs/84459) because it runs another Safari session while Safari does not allow multiple sessions running simultaneously.

We already have a [workaround for it](https://github.com/pyodide/pytest-pyodide/blob/2a9b65d9afe5126861e7f6badeee7a464934210c/pytest_pyodide/hook.py#L206-L221), but it was not working for this test, because it was not using selenium fixture directly. This PR refactors the test, so our workaround can be applied.

The safari tests are still very flaky... but this PR at least fix one.

### DIFF
--- a/src/tests/test_pyodide.py
+++ b/src/tests/test_pyodide.py
@@ -7,7 +7,6 @@ from typing import Any
 
 import pytest
 from pytest_pyodide import run_in_pyodide
-from pytest_pyodide.fixture import selenium_standalone_noload_common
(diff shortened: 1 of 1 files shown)


hoodmane (approved):
Thanks @ryanking13!

PR #4300, opened 2023-11-16
JSPI Improve support for runtimes that don't allow dynamic eval

hoodmane:
This fixes it so that if `new WebAssembly.Module` fails, we don't try to use it in JSPI. This doesn't 100% fix no-dynamic-eval but it makes it work a lot better.

- [x] Add / update tests

### DIFF
--- a/src/core/_pyodide_core.c
+++ b/src/core/_pyodide_core.c
@@ -8,7 +8,11 @@
 
 #define FATAL_ERROR(args...)                                                   \
   do {                                                                         \
-    PyErr_Format(PyExc_ImportError, args);                                     \
+    if (PyErr_Occurred()) {                                                    \
+      _PyErr_FormatFromCause(PyExc_ImportError, args);                         \
+    } else {                                                                   \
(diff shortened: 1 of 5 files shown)


ryanking13 (commented):
Thanks @hoodmane! However, the test you added didn't seem to pass the CI?

hoodmane:
It passes:
```
test_syncify.py::syncify_not_supported2[chrome] PASSED
test_syncify.py::syncify_not_supported2[firefox] PASSED
```
There was a flake on Safari but it's unrelated.

ryanking13:
> There was a flake on Safari but it's unrelated.

Could you please double-check that this is a flake on Safari? It seems to be failing consistently from what I can see, with the following error message:

```
wasm instantiation failed!
```

hoodmane:
Oh, interesting. I don't have a way to test things on Safari so we can just xfail it there.

hoodmane:
Ah I added a comment here:
```js
// Note: only will work with newer runtimes that have WebAssembly.Function
```
Probably the trouble is that our safari doesn't have `WebAssembly.Function`.

Issue #3843, opened 2023-05-09
Make pyodide.loadPackage return list of packages that are loaded


ryanking13:
I think it would be nice if pyodide.loadPackage returns list of packages that are loaded, so that downstream packages that use it (micropip) can know what packages are loaded.

- Related: https://github.com/pyodide/micropip/pull/60

ryanking13:
@RulerOfCakes are you interested in implementing this issue?

RulerOfCakes:
Thank you for the detailed explanation! I've opened #4306 based on your description.

PR #4306, opened 2023-11-18
Add package data returned from pyodide.loadPackage

RulerOfCakes:
### Description

Added some lockfile package types from [pyodide-lock](https://github.com/pyodide/pyodide-lock/blob/d8251675d0c6a93ab02a91a5d73c3d3d6d88a0ad/pyodide_lock/spec.py#L19-L28), to now return loaded package data from `pyodide.loadPackage`. This PR addresses #3843.

I'm not sure how to handle the case of external packages that aren't sourced from `pyodide-lock.json`, so I've left it empty in that case for now.

### Checklists

- [x] Add / update tests


src/js/types.ts
```diff
@@ -310,8 +317,8 @@ export interface API {
   lockFilePromise: Promise<any>;
   lockfile_unvendored_stdlibs: string[];
   lockfile_unvendored_stdlibs_and_test: string[];
-  lockfile_info: any;
-  lockfile_packages: any;
+  lockfile_info: LockfileInfo;
+  lockfile_packages: Record<string, PackageData>;
   repodata_packages: any;
   repodata_info: any;
```
hoodmane, line 323: `repodata_packages` and `repodata_info` are synonyms for `lockfile_packages` and `lockfile_info` so could you give them the same type?

src/js/types.ts
```diff
@@ -310,8 +317,8 @@ export interface API {
   lockFilePromise: Promise<any>;
```
hoodmane, line 317: While you're at it, maybe fill in the type of the lockfile which I think is:
```ts
{info: LockfileInfo, packages: Record<string, PackageData>}
```

src/js/load-package.ts
```diff
@@ -351,7 +369,7 @@ async function downloadAndInstall(
     await Promise.all(installPromiseDependencies);
 
     await installPackage(pkg.name, buffer, pkg.channel);
-    loaded.add(pkg.name);
+    if (pkg.packageData !== undefined) loaded.add(pkg.packageData);
```
hoodmane, line 372:
```suggestion
    if (pkg.packageData) {
       loaded.add(pkg.packageData);
    }
```

src/js/load-package.ts
```diff
@@ -483,8 +502,10 @@ export async function loadPackage(
     );
 
     Module.reportUndefinedSymbols();
-    if (loaded.size > 0) {
-      const successNames = Array.from(loaded).join(", ");
+    if (loadedPackageData.size > 0) {
+      const successNames = Array.from(loadedPackageData)
+        .map((pkg) => pkg.name)
+        .join(", ");
```
hoodmane, line 508: `Array.from(x).map(f)` should be written as `Array.from(x, f)`. And maybe we should sort these? 
```suggestion
      const successNames = Array.from(loadedPackageData, (pkg) => pkg.name).sort().join(", ");
```

src/js/load-package.ts
```diff
@@ -94,15 +94,27 @@ type PackageLoadMetadata = {
...
 };
 
+export type PackageType =
+  | "package"
+  | "cpython_module"
+  | "shared_library"
+  | "static_library";
+
 // Package data inside pyodide-lock.json
 export type PackageData = {
```
hoodmane, line 107: @ryanking13 Do we want to expose all this stuff? What are the stability guarantees we want to make? Maybe we should put a stability tag in the documentation to say that the `PackageData` return type is subject to change? Or mark specific fields as unstable?

A bunch of extra tsdoc tags work now for things like stability e.g., 
https://tsdoc.org/pages/tags/experimental/
or something like that. I'm not sure what the best way to say "this is more likely to change than most of our API surface" is.

src/js/load-package.ts
```diff
@@ -94,15 +94,27 @@ type PackageLoadMetadata = {
...
-  shared_library: boolean;
-  depends: string[];
-  imports: string[];
   install_dir: string;
+  sha256: string;
+  package_type: PackageType;
+  imports: string[];
+  depends: string[];
+  unvendored_tests: boolean;
+  shared_library: boolean; // This field is deprecated
```
hoodmane, line 117: Can we just get rid of `shared_library`? It isn't used internally and it's better to clean things up before exposing them.

hoodmane (commented):
Thanks for contributing @RulerOfCakes! Generally looks good to me. 

We should have a test checking that the added types are exported in `dist/pyodide.d.ts`. In particular, we need to update the tests checking the return value of `loadPackage` here:
https://github.com/pyodide/pyodide/blob/main/src/test-js/index.test-d.ts#L74
It's good to see that these tests are currently failing.

Also, `loadPackagesFromImports` should be updated to return the result from `loadPackage`.

src/js/load-package.ts
hoodmane: Maybe we can leave this to a followup...

src/js/load-package.ts
ryanking13: I think we would like only to expose

- name
- version
- file_name
- package_type

and add experiemntal_tag in `package_type`. Other keys don't need to be exposed.

src/js/load-package.ts
ryanking13: I think we can get rid of it.

src/js/test/integration/pyodide.test.js
```diff
@@ -11,9 +11,16 @@ describe("Pyodide", () => {
   describe("micropip", () => {
     before(async () => {
       const factory = async () => {
-        await pyodide.loadPackage(["micropip"]);
+        return pyodide.loadPackage(["micropip"]);
```
ryanking13, line 14: IIRC we are not running the integration tests in the CI right now. Could you confirm that this test passes?

src/js/load-package.ts
```diff
@@ -286,11 +299,16 @@ async function installPackage(
...
     pkg = {
+      name: "",
+      version: "",
       file_name: ".whl",
-      shared_library: false,
-      depends: [],
-      imports: [] as string[],
       install_dir: "site",
+      sha256: "",
+      package_type: "cpython_module",
```
ryanking13, line 307: This condition is for installing a package from a URL, not installing a cpython_module. (Sorry, the existing code is a bit messy). So I think you need to set the `package_type` to "package".

ryanking13 (commented):
Thanks @RulerOfCakes! I have some minor comments but it generally looks good to me.

Could you please write some tests in [test_package_loading.py](https://github.com/pyodide/pyodide/blob/main/src/tests/test_package_loading.py)?

src/js/test/integration/pyodide.test.js
RulerOfCakes: Sure, I have checked that the test passes but I had to install an additional package that wasn't in `package.json`, `puppeteer`, and run `conftest.js` manually before the tests. Is there documentation on how to run these tests that im missing?

ryanking13: > puppeteer, and run conftest.js manually before the tests. Is there documentation on how to run these tests that im missing? Or should these steps be added in package.json?

I apologize for the difficulty in testing. We haven't used JS tests much so far, so we don't have a lot of documentation on them. As for puppeteer, We've removed it from dependencies (https://github.com/pyodide/pyodide/issues/3332) for now because installing that package affects build times. In the long run, we plan to replace it with something like playwright.

src/tests/test_package_loading.py
```diff
@@ -136,7 +136,7 @@ def test_invalid_package_name(selenium):
 )
 def test_load_packages_multiple(selenium_standalone, packages):
     selenium = selenium_standalone
-    selenium.load_package(packages)
+    loaded_packages = selenium.load_package(packages)
```
ryanking13, line 139: [selnium.load_package](https://github.com/pyodide/pytest-pyodide/blob/045c569f8583f2bb456045a43ac172c7e74889ce/pytest_pyodide/runner.py#L328-L329) is a wrapper function for `pyodide.loadPackage`, but it does not return the return value.

Maybe we can update `pytest-pyodide`, but it would need a few steps, so for now let's call pyodide.loadPackage directly. You can do:

```
selenium.run_js("return await pyodide.loadPackage('...')")
```

src/js/load-package.ts
hoodmane: And ideally also filter the values at runtime...

src/tests/test_package_loading.py
RulerOfCakes: I have fixed the tests with your suggestion, but it looks like it is failing only on safari. Could you perhaps take a look and tell what may be wrong here? @ryanking13

src/js/types.ts
```diff
@@ -307,13 +319,13 @@ export interface API {
...
+  lockFilePromise: Promise<Lockfile>;
   lockfile_unvendored_stdlibs: string[];
   lockfile_unvendored_stdlibs_and_test: string[];
-  lockfile_info: any;
-  lockfile_packages: any;
-  repodata_packages: any;
-  repodata_info: any;
+  lockfile_info: LockfileInfo;
+  lockfile_packages: Record<string, PackageData>;
+  repodata_packages: Record<string, PackageData>;
```
hoodmane, line 327: Better to use `InternalPackageData` here since `API` is private anyways.

src/js/api.ts
```diff
@@ -192,8 +192,9 @@ export class PyodideAPI {
       }
     }
     if (packages.size) {
-      await loadPackage(Array.from(packages), options);
+      return loadPackage(Array.from(packages), options);
```
hoodmane, line 195: I think we would usually put an await here. It doesn't really change the behavior but perhaps it's a bit clearer.
```suggestion
      return await loadPackage(Array.from(packages), options);
```

src/js/load-package.ts
```diff
@@ -40,7 +40,7 @@ async function initializePackageIndex(lockFilePromise: Promise<any>) {
   // compute the inverted index for imports to package names
   API._import_name_to_package_name = new Map();
   for (let name of Object.keys(API.lockfile_packages)) {
-    const pkg = API.lockfile_packages[name];
+    const pkg = API.lockfile_packages[name] as InternalPackageData;
```
hoodmane, line 43: If you give `lockfile_packages` keys type `InternalPackageData` we shouldn't need these casts.

src/js/load-package.ts
```diff
@@ -502,6 +531,7 @@ export async function loadPackage(
     API.importlib.invalidate_caches();
   } finally {
     releaseLock();
+    return Array.from(loadedPackageData);
```
hoodmane, line 534: Don't put returns inside finally blocks! It will discard exceptions! e.g., try out the following:
```js
function f() {
   try {
      throw new Error("hi");
   } finally {
      return 0;
   }
}
f()
```
This should go at the bottom of the try block.

hoodmane: Let's also filter for the keys we want to expose:
```js
function filterPackageData({ name, version, file_name, package_type}: PrivatePackageData) : PackageData {
   return { name, version, file_name, package_type };
}
```
and then `return Array.from(loadedPackageData, filterPackageData);`.

hoodmane: @ryanking13 maybe we should also change `file_name` and `package_type` to camelCase in the public packagedata? WDYT?

RulerOfCakes: > Don't put returns inside finally blocks! It will discard exceptions! e.g., try out the following:
> 
> ```js
> function f() {
>    try {
>       throw new Error("hi");
>    } finally {
>       return 0;
>    }
> }
> f()
> ```
> 
> The return should go at the bottom of the try block.

wow, I wasn't aware of such behavior. Thanks for letting me know 😮

ryanking13: > @ryanking13 maybe we should also change file_name and package_type to camelCase in the public packagedata? WDYT?

Yeah, we haven't been very consistent about variable names, but it's probably a good idea to use camelCase for publicly exposed stuff.

`{ name, version, fileName, packageType }`

ryanking13 (approved):
Thanks @RulerOfCakes! Could you update the changelog and [address the review comment](https://github.com/pyodide/pyodide/pull/4306#discussion_r1404073698)? Otherwise it looks good to me. Thanks for your contribution!

RulerOfCakes:
> Thanks @RulerOfCakes! Could you update the changelog and [address the review comment](https://github.com/pyodide/pyodide/pull/4306#discussion_r1404073698)? Otherwise it looks good to me. Thanks for your contribution!

Thank you! I've updated the changelog as requested. Could you merge the PR?

ryanking13:
I updated the PR number in the changelog. Thanks again!

hoodmane:
Thanks again @RulerOfCakes!

PR #4319, opened 2023-12-02
Normalize package names in pyodide-lock.json 

### LINKED ISSUE #4182: API Normalize package names in pyodide-lock.json
In https://github.com/pyodide/pyodide-lock/pull/20#issuecomment-1735331697 we discussed that it would be good to normalize the package names. This was also supported by @bollwyvl's comment in https://github.com/pyodide/pyodide-lock/pull/18#discussion_r1336556106 and also what the [PEP 691](https://peps.python.org/pep-0691/) for Simple JSON API recommends. 

In the case of pyodide, I imagine that would mean doing,

```py
packages: {
    <normalized_name>: {
              name : <normalized_name>
              depends: [<normalized_names>]
     }
}
```
so we would be repeating the normalized name twice once as the dict key, and the second time as the inner name field. 

This however means that if say we do `micropip.install('scikit-learn')` (or if it's pulled as a dependency) in the logs we would see `installing scikit_learn` which I personally find strange. Unless we de-normalize the name somehow when displaying (at least replacing `_` with `-`).

For the input `meta.yaml` this souldn't change anything, and we would only normalize the output name.

cc @ryanking13 @joemarshall

ryanking13:
### Description

Close #4182

Normalizes package names stored in pyodide-lock.json, and update corresponding package loading codes.

### Checklists

- [x] Add / update tests
- [x] Update changelog


src/js/packaging-utils.ts
hoodmane, line 1: These need to be marked `@private` to keep them out of the docs.

hoodmane (approved):
Since the changelog mentions micropip maybe it would be good to add a micropip test for this?

src/js/packaging-utils.ts
ryanking13: Good point. Thanks!

PR #4322, opened 2023-12-03
TYPING More accurate type for `JsPromise`

hoodmane:
We gave `PyodideFuture` a very accurate type definition for `then` and `catch`. This copies the same types to `JsPromise`. Unfortunately I was unable to work out a way to share the types between the two of them due to limitations of Python's type system.


ryanking13 (approved):
Thanks for the typing work!

PR #4327, opened 2023-12-05
Fix stdin reader to avoid treating empty strings as EOF

### LINKED ISSUE #4258: Issue with stdin and empty strings using Node.js
## 🐛 Bug

This bug only occurs in the NPM version. I can't replicate this in a browser.
When using a custom stdin callback, if an empty string is returned, an EOFError is thrown. An empty string seems to be acting the same as null or undefined. I tried changing autoEOF to false but that didn't have an effect.

### To Reproduce

Make a new node project and put the following code in the the index.js
```javascript

const testStdin = async (arr) => {
    const py = await loadPyodide();
    let idx = 0;
    py.setStdin({
        stdin: () => {
            return arr[idx++];
        }
    });
    py.runPython(`
        a = input()
        b = input()
        c = input()
        print(a)
        print(b)
        print(c)
    `);
}

testStdin(['asdf', 'qwer', '']).then(console.log).catch(console.error)
```
- An EOFError is thrown

This code works correctly if the last item in the array is not empty

### Expected behavior

Expected to see the value of c be assigned ""

### Environment

- Pyodide Version: 0.24.0
- Node version: 20.5.0

ryanking13:
### Description

Close #4258

### Checklists

- [x] Add / update tests

### DIFF
--- a/docs/project/changelog.md
+++ b/docs/project/changelog.md
@@ -61,6 +61,9 @@ myst:
 - {{ Enhancement }} Added `JsArray.remove` and `JsArray.insert` methods.
   {pr}`4326`
 
+- {{ Fix }} `pyodide.setStdin` now does not consider an empty string as EOF.
+  {pr}`4327`
+
 ### Pyodide CLI
 
 - {{ Enhancement }} `pyodide config` command now show additional config variables:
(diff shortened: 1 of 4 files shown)


hoodmane:
Thanks @ryanking13. I tried this a while back and I thought there was a problem with it but probably I was over complicating things.

hoodmane (approved):

ryanking13:
Thanks for the review!

PR #4329, opened 2023-12-05
Add WebLoop.close method

hoodmane:
Which is a no-op, so we don't crash if people try to close the loop.


ryanking13 (approved):
Make sense. Thanks!

PR #4332, opened 2023-12-06
added requests and (direct from git) urllib3

joemarshall:
urllib3 in git now builds and tests with pyodide support. This means requests also should work on pyodide (and does at least for the very basic test case used for requests in urllib3 tests).

Maybe worth waiting until an actual urllib3 release before merging this PR, but I thought people might be interested in seeing this, as it makes tons of network stuff function nicely without any `pyodide-http` style monkey patching.


joemarshall:
Update - I put an absolutely basic sanity test that both packages function into pyodide. There is extensive testing in urllib3 CI, so I'm hoping that this is sufficient this end to essentially just check that any updates to pyodide haven't broken urllib3.

hoodmane:
Sorry for being slow to review @joemarshall I'll try to look at this later today.

ryanking13:
Thank you so much for working in this @joemarshall! This is a really important feature and one that many people have been asking for.

Question: I guess there are features of urllib3 or requests that are not supported due to the limitations of the Web API. Is this mentioned in the documentation of urllib3 or requests? If so, we'll need to [mention that in our FAQ](https://pyodide.org/en/stable/usage/faq.html#how-can-i-load-external-files-in-pyodide).

hoodmane (approved):
Thanks @joemarshall! Tests look good enough for now, we can add more if people report problems. As @ryanking13 suggested, maybe we should add a remark in the wasm limitations section of the docs. We can be a bit vague for now e.g., "some features of synchronous requests libraries may not work as expected". Also, if they rely on sync XHR, it's worth mentioning that some JS runtimes don't have XHR so they'll need a polyfill for sync XHR or they can't work.

rth:
Exciting to see that the upstream patch to urllib3 was merged. So for people who currently use pyodide-http, will that continue to work, if they keep their code unchanged after this update? I guess it should, as patches where done in requests there, right?

rth:
Also sorry I'm arriving a bit late to ask questions :), but https://github.com/urllib3/urllib3/pull/3195 would it work for binary files? Although technically not supported in sync request XHR, if I remember correctly pyodide-http did some hack around it.

joemarshall:
Yeah, the hack from pyodide http is used here which means that binary files work even on main browser thread.

The main thing that is weird is that in most cases, streaming http doesn't stream and falls back to downloading the whole thing. Oh, and you need to be aware that the browser handles all network traffic, so restrictions relating to cross origin requests etc. are all enforced, and features such as custom certificates, custom proxies etc. are ignored.

If you serve as cross origin isolated, and it has time to get the webworker running it can stream. There's details in the urllib3 documentation of how to do this, or will be once the docs pull request is merged in.

As an aside, it's annoying how slowly the webassembly JavaScript promise integration proposal is moving, to me that has so much potential for web python and webassembly - suddenly things like streaming http become trivial, and so much stuff where threading and web workers or service workers have been needed suddenly disappears. Oh and emscripten filesystem support could play nicely with synchronous c and python.

rth:
Thanks for your detailed response @joemarshall !

rth:
Aside from the documentation improvements suggested by Hood above, it looks like this is ready to be merged. 

cc @koenvo FYI. The scope of pyodide-http is likely larger from what I understood than just requests / urllib3.

ryanking13:
The PR looks good to be to. Should we include this in the next release? If we do, I think it's worth mentioning in the release note blog post (including its limitations) because people will be interested in it.

joemarshall:
@rth @ryanking13 Put that bit of docs in it and added to changelog now.

rth:
Thanks a lot @joemarshall ! Pushed a small for a conflict in the changelog. Will merge once CI passes.

PR #4351, opened 2023-12-23
DOC Replace minimal supported browser versions with tested browser versions

### LINKED ISSUE #4293: DOCS: Update minimum supported browser versions
They are pretty far out of date. See discussion in #4225.

ryanking13:
### Description

Close #4293

This PR replaces outdated minimum-supported browser versions with tested browser versions in docs.

According to Caniuse, users update their browser versions very quickly, so I think it's better to specify the version we're testing rather than mentioning a browser version that's 3-4 years old in the documentation.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -352,6 +352,11 @@ jobs:
             python -m pip install -r requirements.txt
             pip install -e ./pyodide-build
       - run:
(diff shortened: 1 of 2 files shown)


rth (approved):
Thanks!

PR #4353, opened 2023-12-24
Remove version suffix from shared libraries

### LINKED ISSUE #4335: Remove version suffix from shared libraries
Some libraries produce version-suffixed output depending on the build options, which can result in the same file being included multiple times in the zip archive.

In a native Linux environment, this is not a problem because only one of them is a real file and the rest are symlinks, but in our case, copying these files into a zip archive creates multiple identical files, increasing the size of the archive.

For example, in `geos`  library, the zip file contains:

```
Archive:  geos-3.10.3.zip
  Length      Date    Time    Name
---------  ---------- -----   ----
  1659324  2023-11-17 03:11   libgeos.so
  1659324  2023-11-17 03:11   libgeos.so.3.10.3
   107215  2023-11-17 03:11   libgeos_c.so
   107215  2023-11-17 03:11   libgeos_c.so.1
   107215  2023-11-17 03:11   libgeos_c.so.1.16.1
```

So it would be nice if we can strip out version suffix and only build a single library.

Related articles:

- Cmake: https://cmake.org/cmake/help/latest/prop_tgt/SOVERSION.html
- libtool: https://stackoverflow.com/questions/24395746/disabling-version-suffix-extension-in-autoconf-or-makefile

ryanking13:
Resolve #4335

Add cmake, autotools flags to prevent versioned shared libraries. There is no user facing change so it does not require changelog.


rth (approved):
Makes sense, thanks!

PR #4354, opened 2023-12-26
DOCS: Clarify `registerJsModule` and submodules

hoodmane:
This adds some clarification to the intended behavior of `registerJsModule`
regarding submodules. See discussion in #4346.

cc @WebReflection


ryanking13 (approved):
Thanks!

Issue #4070, opened 2023-08-16
package multidict


Cadair:
## 🐍 Package Request

- Package Name and Version : multidict
- Package URL : https://github.com/aio-libs/multidict
- Package Dependencies that needs to be resolved first: none

Hello, I am trying to get [sunpy](https://github.com/sunpy/sunpy) working in pyiodide and we depend on aiohttp which depends on multidict. Despite it being possible to build multidict in pure-python mode, they do not ship a pure wheel. I wonder if it's worth including in the pyiodide distribution to get the compiled speedups? Alternatively, should I approach the multidict project about publishing a pure wheel?

jobovy:
I think you'll need to add ``aiohttp`` itself as well (and perhaps other packages as well), because it also does not have pure-python wheels. ``micropip.install('aiohttp')`` does work, but it installs ``aiohttp`` version 3.6.2, which is from 2019 and which happened to have a pure Python wheel available.

I actually think this is a problem for ``pip`` in general and ``micropip`` specifically, that they will happily install a very old version if that's the only available one for your system/Python version/... without warning that you are installing a very old version. For example, when a package switches to having binary extensions, ``micropip`` would keep installing the last version with a pure Python wheel; I've run into this multiple times in the past. People checking to see whether a package needs to be compiled to be used in pyodide might not notice then that it should be compiled. Because this is perhaps a bigger issue for ``micropip``, perhaps it could warn that not the latest version is being installed if that's the case.

jobovy:
Some related issues/PRs:
* https://github.com/pyodide/pyodide/pull/3581: ``multidict`` is already part of pyodide, but the version is incompatible with ``aiohttp``
* https://github.com/pyodide/pyodide/issues/3588: requires ``aiohttp`` for the ``openai`` package. Dependencies of ``aiohttp`` listed there:
  * aiosignal
    * frozenlist
  * async-timeout
  * attrs
  * charset-normalizer
  * frozenlist
  * multidict
  * yarl
    * idna
    * multidict

As pointed out by @rth in https://github.com/pyodide/pyodide/issues/3588#issuecomment-1485828355, many of these (all?) are pure Python dependencies that would also have to be packaged within ``pyodide`` (and then there is still the issue of the ``multidict`` version required by ``aiohttp`` being incompatible with the latest one which is included in pyodide.

To get ``sunpy`` to work in pyodide for now, it might be easiest to not have ``aiohttp`` as a hard dependency.

[...]

Issue #574, opened 2019-11-30
Can't use WebSocket


filips123:
It is not possible to use WebSockets in Python. Although JavaScript's `WebSocket` object can be imported, it can't be constructed and used.

---

First code that doesn't work is:

```py
from js import WebSocket
ws = WebSocket('wss://echo.websocket.org')
```

After running it, there is an error:

```py
TypeError: Constructor WebSocket requires 'new'
_hiwire_call (https://pyodide.cdn.iodide.io/pyodide.asm.js:8:30963)
https://pyodide.cdn.iodide.io/pyodide.asm.wasm:wasm-function[314]:0xfab93::
https://pyodide.cdn.iodide.io/pyodide.asm.wasm:wasm-function[14868]:0x96fe7e::
https://pyodide.cdn.iodide.io/pyodide.asm.wasm:wasm-function[837]:0x1337f2::
https://pyodide.cdn.iodide.io/pyodide.asm.wasm:wasm-function[2749]:0x3287e0::
https://pyodide.cdn.iodide.io/pyodide.asm.wasm:wasm-function[2744]:0x3210b8::
https://pyodide.cdn.iodide.io/pyodide.asm.wasm:wasm-function[20574]:0x9834c7::
https://pyodide.cdn.iodide.io/pyodide.asm.wasm:wasm-function[2739]:0x3156d2::
https://pyodide.cdn.iodide.io/pyodide.asm.wasm:wasm-function[2738]:0x31460e::
https://pyodide.cdn.iodide.io/pyodide.asm.wasm:wasm-function[2686]:0x30fc38::
https://pyodide.cdn.iodide.io/pyodide.asm.wasm:wasm-function[15955]:0x973528::
https://pyodide.cdn.iodide.io/pyodide.asm.wasm:wasm-function[839]:0x133dc0::
https://pyodide.cdn.iodide.io/pyodide.asm.wasm:wasm-function[840]:0x1343c5::
https://pyodide.cdn.iodide.io/pyodide.asm.wasm:wasm-function[2749]:0x32869e::
https://pyodide.cdn.iodide.io/pyodide.asm.wasm:wasm-function[2744]:0x3210b8::
https://pyodide.cdn.iodide.io/pyodide.asm.wasm:wasm-function[20574]:0x9834c7::
https://pyodide.cdn.iodide.io/pyodide.asm.wasm:wasm-function[835]:0x1330e8::
[...]

PR #4355, opened 2023-12-27
new packages: `web3` and 'pyunormalize'

gorondan:
### Checklists

- [ x] Add a [CHANGELOG](https://github.com/pyodide/pyodide/blob/main/docs/project/changelog.md) entry

### DIFF
--- a/packages/pyunormalize/meta.yaml
+++ b/packages/pyunormalize/meta.yaml
@@ -1,0 +1,15 @@
+package:
+  name: pyunormalize
+  version: 15.1.0
+  top-level:
+    - pyunormalize
+source:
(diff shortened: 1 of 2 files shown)


gorondan:
I strongly believe that web3.py module should be available in pyodide and that it will be very useful, as there are tons of sdk's in python and interacting with smart contracts straight from front-end will be a blessing for many
Please support this proposal, as 'web3' has dependencies that need to be tested and added in /packages

ryanking13:
Thanks for the contribution @gorondan.

`web3.py` seems to have quite a few dependencies, most of which can be installed with micropip since it is a pure python package. Therefore, not all packages need to be included in the Pyodide distribution, and I would appreciate it if you could organize the dependencies well enough to indicate which binary extension packages should be included.

Another concern is that web3.py seems to rely on various networking libraries such as requests, aiohttp, and websockets, which are likely to not work well with Pyodide due to browser limitations. (See: #574, #4070). Therefore, someone needs to verify the behavior of this package.

ryanking13:

PR #4362, opened 2024-01-05
Emscripten 3.1.47

ryanking13:


ryanking13:
There seems like a regression in the llvm parser.

Rleated issue in upstream: https://github.com/llvm/llvm-project/issues/75335

ryanking13:
Some more related upstream issue:

- https://github.com/llvm/llvm-project/issues/68524
- https://github.com/llvm/llvm-project/issues/67979

ryanking13:
The bug seems to be in the parsing of comments by the LLVM parser. So for now I'm just stripping the comment out of the file that causes the problem ¯\\_(ツ)_/¯

ryanking13:
Seems like there is another issue... which is probably fixed in the latest LLVM: https://github.com/llvm/llvm-project/issues/76358

ryanking13:
Superseded by https://github.com/pyodide/pyodide/pull/4399

PR #4363, opened 2024-01-05
DOCS Add documentation for JSPI related methods

hoodmane:
Explains that they are experimental, require JSPI. wip, some xrefs are still broken.
It would still be nice to make `@experimental` tag autorender as an experimental admonition.

### DIFF
--- a/src/js/api.ts
+++ b/src/js/api.ts
@@ -285,6 +285,30 @@ export class PyodideAPI {
     return await API.pyodide_code.eval_code_async.callKwargs(code, options);
   }
 
+  /**
+   * Runs a Python code string like :js:func:`pyodide.runPython` but with stack
+   * switching enabled. Code executed in this way can use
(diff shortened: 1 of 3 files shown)


src/js/api.ts
```diff
@@ -285,6 +285,30 @@ export class PyodideAPI {
...
+   *
+   *    This feature is not yet stable.
+   *
+   * @experimental
+   * @param options
+   * @param options.globals An optional Python dictionary to use as the globals.
+   * Defaults to :js:attr:`pyodide.globals`.
+   * @param options.locals An optional Python dictionary to use as the locals.
+   *        Defaults to the same as ``globals``.
+   * @param options.filename An optional string to use as the filename. Defaults
```
ryanking13, line 306: It seems like `filename` key is missing in the options parameter.

ryanking13 (approved):
Thanks!

src/js/api.ts
hoodmane: Probably I should add it to the function for consistency.

PR #4370, opened 2024-01-09
adding dse packages

ddl-aj-rossman:
### Checklists

- [ ] Add / update tests

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -941,6 +941,6 @@ workflows:
             - build-test-pyc-packages
           filters:
             branches:
-              only: jupyterlite-dse
+              only: main
           context:
             - s3-deployment
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -941,6 +941,6 @@ workflows:
             - build-test-pyc-packages
           filters:
             branches:
-              only: main
+              only: jupyterlite-dse
           context:
             - s3-deployment
--- a/packages/beautifulsoup4/meta.yaml
+++ b/packages/beautifulsoup4/meta.yaml
@@ -3,6 +3,7 @@ package:
   version: 4.12.2
   top-level:
     - bs4
+  tag: core
 requirements:
   run:
     - soupsieve
--- a/packages/beautifulsoup4/meta.yaml
+++ b/packages/beautifulsoup4/meta.yaml
@@ -3,7 +3,8 @@ package:
   version: 4.12.2
   top-level:
     - bs4
-  tag: core
+  tag:
+    - core
 requirements:
   run:
     - soupsieve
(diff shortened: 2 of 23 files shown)


ryanking13:
Hi, @ddl-aj-rossman! All the packages you added in this PR seem like a pure Python wheel that is [installable with the micropip](https://pyodide.org/en/stable/development/new-packages.html#determining-if-creating-a-pyodide-package-is-necessary). So I think you don't need to add them in Pyodide distribution.

ryanking13:
> We're trying to cut down on the time required to get up and running with the notebook.

I see. You don't need to build the Pyodide from scratch for your use case.

I can suggest two ways to do it: modifying the lockfile, or bundling all packages into the zipfile.

1) modifying the lockfile

Assuming that all packages you want to use are pure Python packages,

first install all packages with micropip, and create a lockfile with `micropip.freeze`

```js
await pyodide.loadPackage("micropip")
const micropip = pyodide.pyimport("micropip")
await micropip.install(["packages", "you", "need"])
micropip.freeze()
```

micropip.freeze() will print the JSON lockfile which includes all packages that are installed. You can save it into a file, host it somewhere, and when loading Pyodide next time, [set the `lockFileURL` parameter](https://pyodide.org/en/stable/usage/api/js-api.html#globalThis.loadPyodide) to point that file:

```js
const pyodide = loadPyodide({
  lockFileURL: "<the lockfile created by micropip.freeze>"
  packages: ["packages", "you", "need"]
})
```

Using a lockfile allows you to skip the dependency resolution process that micropip does, which speeds up package installation and setting the `packages` parameter will download packages in parallel so it will speed up the initialization time.
However, if you need more optimization, try the second option.

2) bundling all packages into the zipfile.

This one is more close to "bake all of the packages we need into the kernel."

When Pyodide initializes, it downloads a compressed file called `python_stdlib.zip` and loads stdlib from it. You can put the content of other packages into this zipfile. This is semantically similar to saying that you will treat those packages as standard libraries.

```bash
$ unzip -l python_stdlib.zip
Archive:  python_stdlib.zip
  Length      Date    Time    Name
---------  ---------- -----   ----
        0  2023-12-11 20:21   asyncio/
        0  2023-12-11 20:21   collections/
        0  2023-12-11 20:21   concurrent/
        0  2023-12-11 20:21   ctypes/
        0  2023-12-11 20:21   email/
...
```

For example, say you want to include the package `seaborn`. Then you can do that by copying the content into the zipfile.

```bash
$ unzip python_stdlib.zip -d new_stdlib
$ unzip seaborn-0.13.1-py3-none-any.whl -d new_stdlib
$ zip -r new_stdlib.zip new_stdlib/*
```

Then host this `new_stdlib.zip` somewhere, and when loading Pyodide, set `stdLibURL` to point that file:

```js
const pyodide = loadPyodide({
  stdLibURL: "<the zipfile URL>"
})
pyodide.runPython("import seaborn")
```

ryanking13:

PR #4384, opened 2024-01-16
Update JSPI code to handle change in type reflection spec

### LINKED ISSUE #4378: Can't use alpha version: TypeError: WebAssembly.Function.type is not a function
## 🐛 Bug

I'm trying to use latest (alpha) versions of Pyodide (0.25.0a1 or above)
I was curious about stack switching with syncify() API
And when package is loaded and trying to install wasm code, the next error is raised:
```
wasm instantiation failed!
TypeError: WebAssembly.Function.type is not a function
```
Screenshot from 2024-01-14 23-49-00

### To Reproduce

Just visit [console](https://pyodide.org/en/latest/console.html) on dev build

### Expected behavior

It should work, isn't it? 🙃

### Environment

- Pyodide Version: 0.25.0a1, 0.25.0a2 or dev
- Browser version: Version 120.0.6099.199 (Official Build) (64-bit)

### Additional context

I found some discussion about similar code [here](https://github.com/WebAssembly/js-promise-integration/issues/25)
And the solution is to make next change:
```js
                const type = WebAssembly.Function.type(wasm_func);
```
to
```js
                const type = wasm_func.type();
```

On screenshot above I tried this in debugger and it seems to work :)

hoodmane:
Resolves #4378. Tested manually.

### DIFF
--- a/cpython/patches/0008-Make-Emscripten-trampolines-work-with-JSPI.patch
+++ b/cpython/patches/0008-Make-Emscripten-trampolines-work-with-JSPI.patch
@@ -270,7 +270,7 @@ index 0000000000..d2d24a6e0a
 +  if (n !== undefined) {
 +    return n;
 +  }
-+  n = WebAssembly.Function.type(wasmTable.get(func)).parameters.length;
++  n = wasmFunctionType(wasmTable.get(func)).parameters.length;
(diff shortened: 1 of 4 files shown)


ryanking13 (commented):
Thanks for handling this!

PR #4391, opened 2024-01-19
MAINT Improve error message when there is a lockfile version mismatch

hoodmane:
This makes it a bit easier to figure out why we get this error message around releases


ryanking13 (approved):
Thanks!

PR #4392, opened 2024-01-19
Add apis to discard extra arguments when calling Python functions

hoodmane:
This is intended to be useful for implementing handlers. The handler might not care about all the arguments. It's convenient to have some easy way to discard unneeded ones.

cc @antocuni 

For example, this code here:
https://github.com/pyscript/pyscript/blob/main/pyscript.core/src/stdlib/pyscript/event_handling.py#L31-L42
```py
        sig = inspect.signature(func)
        # Function doesn't receive events
        if not sig.parameters:

            def wrapper(*args, **kwargs):
                func()

            for el in elements:
                add_event_listener(el, event_type, wrapper)
        else:
            for el in elements:
                add_event_listener(el, event_type, func)
```
could be replaced with:
```py
        func = relaxed_wrap(func)
        for el in elements:
            add_event_listener(el, event_type, func)
```

- [x] Add / update tests


src/py/pyodide/code.py
```diff
@@ -25,11 +26,80 @@ def run_js(code: str, /) -> Any:
...
+        if param.kind == Parameter.KEYWORD_ONLY:
+            break
+    else:
+        new_params.append(Parameter("__var_keyword", Parameter.VAR_KEYWORD))
+    new_sig = sig.replace(parameters=new_params)
+    return new_sig
+
+
+@lru_cache
+def _relaxed_call_sig_cached(func):
```
ryanking13, line 59: Question: Why not put cache decorator directly to `_relaxed_call_sig`?

src/py/pyodide/code.py
```diff
@@ -25,11 +26,80 @@ def run_js(code: str, /) -> Any:
     return eval_(code)
 
 
+def _relaxed_call_sig(func):
```
ryanking13, line 29: Could you add some type hints if possible?

src/py/pyodide/code.py
```diff
@@ -25,11 +26,80 @@ def run_js(code: str, /) -> Any:
...
+    new_sig = sig.replace(parameters=new_params)
+    return new_sig
+
+
+@lru_cache
+def _relaxed_call_sig_cached(func):
+    return _relaxed_call_sig(func)
+
+
+def relaxed_wrap(func):
```
ryanking13, line 63: It looks like the implementation of `relaxed_wrap` and `relaxed_call` are quite similar. Maybe we can define one using another, or extract common parts?

ryanking13 (commented):
Thanks @hoodmane! Indeed it looks quite useful in some cases.

src/py/pyodide/code.py
hoodmane: Oh I meant to call the uncached `_relaxed_call_sig` from `relaxed_wrap`. I imagine `relaxed_call` will be called repeatedly with the same function argument but `relaxed_wrap` only once.

src/py/pyodide/code.py
hoodmane: Yeah having a helper like:
```py
def _do_call(func, sig, args, kwargs):
        bound = sig.bind(*args, **kwargs)
        bound.arguments.pop("__var_positional", None)
        bound.arguments.pop("__var_keyword", None)
        return func(*bound.args, **bound.kwargs)
```
would make sense.

src/py/pyodide/code.py
```diff
@@ -25,11 +28,96 @@ def run_js(code: str, /) -> Any:
...
+    """Decorator which creates a function that ignores extra arguments
+
+    If extra positional or keyword arguments are provided they will be
+    discarded.
+    """
+    sig = _relaxed_call_sig(func)
+    if sig is None:
+        raise TypeError("Cannot wrap function")
+    else:
+        sig2 = sig
```
ryanking13, line 87: Why is this needed?

src/py/pyodide/code.py
```diff
@@ -25,11 +28,96 @@ def run_js(code: str, /) -> Any:
...
+    if sig is None:
+        return func(*args, **kwargs)
+    return _do_call(func, sig, args, kwargs)
+
+
+def f(a, *args, b=7):
+    return [a, args, b]
+
+
+relaxed_call(f, 1, b=7, c=9) == [1, (), 7]
```
ryanking13, line 112: Maybe this was accidentally copy/pasted from the test code?

ryanking13 (approved):
Thanks!

src/py/pyodide/code.py
hoodmane: Right clearly.

src/py/pyodide/code.py
hoodmane: For some reason mypy doesn't type narrow this correctly otherwise.

PR #4395, opened 2024-01-20
Make pyimport able to return module attributes

hoodmane:
Before this PR, `pyimport` can be used like: `pyimport("package")` or `pyimport("package.module")` but `pyimport("package.attribute")` fails. This updates `pyimport` to also work to get package attributes.

- [x] Add / update tests


hoodmane:
@ryanking13 another possibility is to add an optional second `fromlist` argument:
```js
pyodide.pyimport(path, fromlist);
```
So that you could do:
```js
const { PyodideConsole, repr_shorten, BANNER } = pyodide.pyimport(
  "pyodide.console",
  ["PyodideConsole", "repr_shorten", "BANNER"]
);
```

hoodmane:
I guess I'll add that here and we can discuss.

hoodmane:
> `pyimport("package.attribute")` seems slightly weird as it is not a supported behavior in Python.

Well we are already deviating from `import a.b` since if you do `__import__("pkg.module")` it returns `pkg` whereas we are returning `pkg.module`. So we behave more like `from pkg import module` than `import pkg.module` already. I think supporting `pyimport("pkg.attribute")` as well meaning `from pkg import attribute` actually makes the mental model more consistent. It's also convenient. WDYT?

It is a little bit weird having:
```py
pyimport("pkg.module")            ==> from pkg import module
pyimport("pkg.module", fromlist)  ==> from pkg.module import <fromlist>
```
but I don't think it's completely crazy.

Also, when we pass a `fromlist` do you think we should return an Array or an Object?

hoodmane:
> we can already do a similar thing in JavaScript using [destructuring assignment syntax](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Destructuring_assignment).

Oh this is a great observation! The only very minor concern is that you create and leak a temporary `PyProxy` this way, but I think it shouldn't matter:
1. packages are already globally stored in `sys.modules`
2. you won't be creating a reference loop so this will eventually get cleaned out by the JS GC
3. this is happening in run-once setup code not in loops or anything.

ryanking13 (approved):
Thanks!

PR #4396, opened 2024-01-20
Update openssl

hoodmane:
Openssl seems to have yanked the previous version 1.1.1n


ryanking13:
It looks like there is a regression in scipy after this PR.

```
pytest_pyodide.runner.JavascriptException: Error: bad export type for 'sf_error': undefined
```

I am not sure how it is related though.

PR #4415, opened 2024-01-24
Add pyodide version to console banner

hoodmane:
I added the pyodide version number to the banner and did some simple tidying.


ryanking13 (approved):
Thanks! Thanks seems a trivial test failure. Otherwise LGTM.

Issue #4411, opened 2024-01-23
'npm install pyodide' results in Error (Can't resolve 'node-fetch' ...)


Hephaistos7:
## 🐛 Bug

The following 5 Warnings and 1 Error occur when installing and running pyodide through 'npm install pyodide':

./node_modules/pyodide/pyodide.js:19:39-46 - Warning: Critical dependency: require function is used in a way in which dependencies cannot be statically extracted
./node_modules/pyodide/pyodide.js:22:37-44 - Warning: Critical dependency: require function is used in a way in which dependencies cannot be statically extracted
./node_modules/pyodide/pyodide.js:381:19-28 - Warning: Critical dependency: the request of a dependency is an expression
./node_modules/pyodide/pyodide.js:392:42-51 - Warning: Critical dependency: the request of a dependency is an expression
./node_modules/pyodide/pyodide.js:407:125-156 - Warning: Critical dependency: the request of a dependency is an expression
./node_modules/pyodide/pyodide.js:299:127-147 - Error: Module not found: Error: Can't resolve 'node-fetch' in '...\node_modules\pyodide'

Upon removing the line in the code marked with a '#1', the following 3 Warnings and 1 Error still appear (notice that now they come from pyodide.mjs instead of pyodide.js):

./node_modules/pyodide/pyodide.mjs:362:17-26 - Warning: Critical dependency: the request of a dependency is an expression
./node_modules/pyodide/pyodide.mjs:373:40-49 - Warning: Critical dependency: the request of a dependency is an expression
./node_modules/pyodide/pyodide.mjs:388:123-154 - Warning: Critical dependency: the request of a dependency is an expression
./node_modules/pyodide/pyodide.mjs:280:125-145 - Error: Module not found: Error: Can't resolve 'node-fetch' in '...\node_modules\pyodide'

I do not really know how to interpret the line numbers, (.../pyodide.js:19:39-46) since pyodide.js, when opened, has only a total of 14 lines of code.

Also, it was my understanding that the module 'node-fetch' is not required to run Pyodide. That being said, upon getting it through (npm install node-fetch), a whole lot of other errors show up (which I'd be happy to share as well if need be). Those errors starting with '<e> [webpack-dev-middleware] HookWebpackError: ENOENT: no such file or directory, scandir
<e>     at makeWebpackError (...\node_modules\@angular-devkit\build-angular\node_modules\webpack\lib\HookWebpackError.js:48:9)ne@4.0.0 dev'

### To Reproduce

[...]

PR #4417, opened 2024-01-24
Remove node-fetch fallback

ryanking13:
### Description

Drops node-fetch fallback and add `engines` key to package.json.

Related: #4411 

It is worth noting that the `engines` key is just for [advisory only](https://docs.npmjs.com/cli/v10/configuring-npm/package-json#engines), so it will not prevent people from installing Pyodide in older Node versions... but at least we can tell users that we don't support old node versions anymore.

### Checklists


docs/project/changelog.md
```diff
@@ -26,6 +26,10 @@ myst:
   functions.
   {pr}`4392`
 
+- {{ Breaking }} Pyodide will not fallback to `node-fetch` anymore when `fetch`
+  is not available in the Node.js environment.
```
hoodmane, line 30: Is this really breaking since we already said we don't support Node 16? Maybe mention Node 16 in the message too?

hoodmane (approved):
Thanks @ryanking13!

docs/project/changelog.md
ryanking13: Thanks. I updated the changelog that it affects Node.js < 18.

PR #4423, opened 2024-01-25
`tools/dependency-check.sh`: Avoid issuing confusing error message

mkoeppe:
### Description

```
$ ./tools/dependency-check.sh 
Must compile with python ..
```

The trivial improvement of this PR is to change this to:
```
$ ./tools/dependency-check.sh 
./tools/dependency-check.sh: This script expects that environment variables PYMAJOR and PYMINOR are set; skipping check_python_version
```

### DIFF
--- a/tools/dependency-check.sh
+++ b/tools/dependency-check.sh
@@ -6,12 +6,21 @@ failure_exit() {
 }
 
 check_python_version() {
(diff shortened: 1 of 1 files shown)


hoodmane:
Thanks @mkoeppe!

PR #4426, opened 2024-01-26
Do not set constraint Cython<3; update `libproj`, `pyproj`, `h5py`

mkoeppe:
### Description

Since the release of Cython 3 in July 2023, many packages have been updated to either set an upper bound "Cython < 3", to support both Cython 0.29.x and 3.0, or to require Cython 3. 

Here we remove the global constraint Cython < 3 added in #4001 and update individual packages if necessary.

### Checklists


hoodmane (approved):
Thanks @mkoeppe!

PR #4431, opened 2024-01-27
Update docker image to bookworm

hoodmane:
Buster isn't available for anymore for Python 312 so we have to update.

Bookworm doesn't seem to have the prelink package unfortunately

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -6,7 +6,7 @@ defaults: &defaults
     # Note: when updating the docker image version,
     #       make sure there are no extra old versions lying around.
     #       (e.g. `rg -F --hidden <old_tag>`)
-    - image: pyodide/pyodide-env:20230506-chrome112-firefox112-py311
+    - image: pyodide/pyodide-env:20240127-chrome114-firefox122-py311
   environment:
     - EMSDK_NUM_CORES: 3
       EMCC_CORES: 3
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -1,7 +1,7 @@
 {
   "name": "Docker",
   // keep in sync with "run_docker"
-  "image": "pyodide/pyodide-env:20230506-chrome112-firefox112-py311",
+  "image": "pyodide/pyodide-env:20240127-chrome114-firefox122-py311",
   "remoteUser": "root",
   "onCreateCommand": ".devcontainer/onCreate-docker.sh"
 }
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,5 +1,5 @@
-FROM node:20.1.0-buster-slim AS node-image
-FROM python:3.11.3-slim-buster
+FROM node:20.11-bookworm-slim AS node-image
+FROM python:3.11.7-slim-bookworm
 
 # Requirements for building packages
 RUN apt-get update \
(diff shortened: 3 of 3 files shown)


.devcontainer/devcontainer.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "Docker",
   // keep in sync with "run_docker"
-  "image": "pyodide/pyodide-env:20230506-chrome112-firefox112-py311",
+  "image": "pyodide/pyodide-env:20240127-chrome114-firefox122-py311",
```
ryanking13, line 4: Was there some issue when you used more recent Chrome versions?

.circleci/config.yml
```diff
@@ -6,7 +6,7 @@ defaults: &defaults
     # Note: when updating the docker image version,
     #       make sure there are no extra old versions lying around.
     #       (e.g. `rg -F --hidden <old_tag>`)
-    - image: pyodide/pyodide-env:20230506-chrome112-firefox112-py311
+    - image: pyodide/pyodide-env:20240127-chrome114-firefox122-py311
   environment:
     - EMSDK_NUM_CORES: 3
```
ryanking13, line 11: We need to update `run_docker` as well.

ryanking13 (commented):
Thanks for updating the docker image! We haven't been updating it for a while.

.devcontainer/devcontainer.json
ryanking13: Oh, it seems like the url for chrome driver has been changed after version 114. I guess we need to update the dockerfile.

image

https://chromedriver.chromium.org/downloads

Dockerfile
```diff
@@ -1,13 +1,13 @@
...
+FROM node:20.11-bookworm-slim AS node-image
+FROM python:3.11.7-slim-bookworm
 
 # Requirements for building packages
 RUN apt-get update \
   && apt-get install -y --no-install-recommends \
         bzip2 ccache f2c g++ gfortran git make \
         patch pkg-config swig unzip wget xz-utils \
         autoconf autotools-dev automake texinfo dejagnu \
-        build-essential prelink autoconf libtool libltdl-dev \
```
hoodmane, line 10: @ryanking13 @rth does anyone know what prelink does for us or why it went away in bookworm?

rth: You added it in https://github.com/pyodide/pyodide/pull/1656 so I don't know exactly why we need it :) 

Looking at https://lwn.net/Articles/341244/ and https://forums.gentoo.org/viewtopic-t-1074192-start-0.html sounds like it was some debatable optimization, so maybe they just removed it.

hoodmane: It seems likely that I copied it out of docker files for building Python since I do find it there. Thanks for investigating @rth!

packages/test/meta.yaml
```diff
@@ -8,14 +8,10 @@ package:
...
   url: $(PYTHON_ARCHIVE_URL)
-  patches:
-    - patches/0005-gh-93839-Move-Lib-ctypes-test-to-Lib-test-test_ctype.patch
-    - patches/0006-gh-93839-Move-Lib-unttest-test-to-Lib-test-test_unit.patch
-    - patches/0007-gh-93839-Use-load_package_tests-for-testmock-GH-9405.patch
-    - patches/0008-Move-test-directories.patch
 build:
   type: cpython_module
   script: |
+    cat $(PYODIDE_ROOT)/cpython/patches/* | patch -p1
```
ryanking13, line 14: I don't want recipes to depend on files that do not exist under the recipe directory, as it will not work out-of-tree, though cpython modules are a bit special case.

Anyway, I think these patches can be removed after we update Python to 3.12, so it is probably okay.

ryanking13 (approved):
Thanks @hoodmane!

packages/test/meta.yaml
hoodmane: I don't like having to manually update these, and I don't think it's important that the cpython tests build out of tree.

PR #4435, opened 2024-01-27
Upgrade to Python 3.12.1

### LINKED ISSUE #4420: Python 3.12 version
## 🚀 Feature

Hi, I tried [REPL](https://pyodide.org/en/stable/console.html), maybe it uses the latest 0.25.0, and I noticed that the python is 3.11.3.

Python 3.12 has released for a few months with a lot of new features. Since there is no issue track the progress. So, I created this one.

### Motivation

N.A.

### Pitch

N.A.

### Alternatives

N.A.

### Additional context

N.A.

hoodmane:
Resolves #4420.

- [x] Add / update tests
- [x] Fix stack switching
- [x] sort through failing core Python tests

### Packages with build failures
   - [x] aiohttp
   - [x] awkward-cpp
   - [x] Cartopy
   - [x] cbor-diag
   - [x] cffi
   - [x] coolprop
   - [x] cramjam
   - [x] cryptography
   - [x] freesasa
   - [x] gdal
   - [x] gensim
   - [x] igraph
   - [x] kiwisolver
   - [x] libmagic
   - [x] libxml
   - [x] multidict
   - [x] opencv-python
   - [x] orjson
   - [x] pysam
   - [x] pywavelets
   - [x] sisl
   - [x] swiglpk
   - [x] wordcloud
   - [ ] xgboost

### Packages with test failures
   - [x] bcrypt
   - [x] fiona


hoodmane:
@ryanking13 when you get a chance I'd appreciate it if you could look into why the github actions build-core fails here.

ryanking13:
> @ryanking13 when you get a chance I'd appreciate it if you could look into why the github actions build-core fails here.

I think we need to update host python version to 3.12 in environment.yml

hoodmane:
@ryanking13 can you also look into fixing the xgboost distutils import failure?

Also, maybe we should add a special "Removed in py312" error for distutils?

```
  File "/lib/python3.12/site-packages/xgboost/__init__.py", line 14, in <module>
    from . import dask
  File "/lib/python3.12/site-packages/xgboost/dask.py", line 60, in <module>
    from .data import FeatureNames
  File "/lib/python3.12/site-packages/xgboost/data.py", line 5, in <module>
    from distutils import version
  File "/lib/python312.zip/_pyodide/_importhook.py", line 162, in get_module_not_found_error
    return orig_get_module_not_found_error(import_name)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

henryiii:
Is there somewhere I can see the lead up to that error? It's supposed to be passing the exact path to Python. And do you want to be using the `cmake` Python package, or a special wrapper around it? If the latter, `CMAKE_EXECUTABLE` will force the location.

hoodmane:
> Is there somewhere I can see the lead up to that error?

Uh well I tried building again and... now it works. It seems that setting `installed_base` in sysconfig fixed it:
```patch
--- a/cpython/adjust_sysconfig.py
+++ b/cpython/adjust_sysconfig.py
@@ -26,6 +26,8 @@ def adjust_sysconfig(config_vars: dict[str, str]):
         BLDSHARED="cc",
         CXX="c++",
         LDCXXSHARED="c++",
+        installed_base=config_vars["prefix"],
+        platbase=config_vars["prefix"],
     )
 ```

https://github.com/pyodide/pyodide/pull/4435/commits/494ddb1fcf3e1624c622bd3622f6eceabacbf8fb

**Comment by hoodmane** [comment]
@ryanking13 would you mind doing a separate PR to update opencv-python? I think the current version is incompatible with python 312 but there are newer releases that probably are so we need to upgrade and rebase the patches.

**Comment by hoodmane** [comment]
I think that setuptools still brings distutils though they also print a deprecation warning. So the specific issue of distutils may just require declaring setuptools as a build requirement.

**Comment by henryiii** [comment]
That's correct, though it's the maintained copy (which was generally the one injected most of the time anyway, but some libraries like NumPy and anyone using `numpy.distutils` couldn't use it).

**Comment by ryanking13** [comment]
> Hmm. This is going to be a bit of a headache. xgboost dropped 32bit support in a recent version, so we have a problem upgrading the version. So if they rely on distutils, it might require some complicated patch.

I got some positive responses from the xgboost team (https://github.com/dmlc/xgboost/issues/10015), so we I guess we will be possible to keep update xgboost :)

For this PR, I think we need to apply this patch to xgboost: https://github.com/dmlc/xgboost/pull/7770/files, which removed distutils usage.

**Comment by hoodmane** [comment]
@ryanking13 it looks like our xgboost is pretty old -- we have 1.6.1 released May 9th, 2022, the distutils patch went in on March 31 2022, latest release is 2.0.3 from December 19th 2023. So I suspect that if we upgrade it will fix the problem. Would you mind updating xgboost?

**Comment by ryanking13** [comment]
> @ryanking13 it looks like our xgboost is pretty old -- we have 1.6.1 released May 9th, 2022, the distutils patch went in on March 31 2022, latest release is 2.0.3 from December 19th 2023. So I suspect that if we upgrade it will fix the problem. Would you mind updating xgboost?

Yeah xgboost dropped 32bit support in 1.7.0 so I failed to update it further. Let me see if I can update it.

**Comment by hoodmane** [comment]
This PR is almost ready so I think we can just leave fixing xgboost to a followup.

File: Makefile.envs [hunk 1]
```diff
@@ -148,6 +149,7 @@ export MAIN_MODULE_LDFLAGS= $(LDFLAGS_BASE) \
 	-lwebsocket.js \
 	-leventloop.js \
 	-lhiwire \
+	-lHacl_Hash_SHA2 \
```
ryanking13, line 152: Maybe we can update [`Setup.local`](https://github.com/pyodide/pyodide/blob/main/cpython/Setup.local) such as:

```
_sha2 sha2module.c -I_hacl/include Modules/_hacl/libHacl_Hash_SHA2.a
```

instead of linking it here manually?

packages/test/meta.yaml
```diff
@@ -31,11 +67,18 @@ build:
...
+    emcc  ${SIDE_MODULE_LDFLAGS} ${TEST_CAPI_OBJECTS[@]//.c/.o} -o ${DISTDIR}/_testcapi.so
+
     for testname in ${TEST_EXTENSIONS}; do \
       emcc Modules/${testname/.so/.o} -o ${DISTDIR}/$testname ${SIDE_MODULE_LDFLAGS}
     done
 
     cd Lib && \
       tar --exclude=__pycache__ -cf - \
-      test distutils/tests sqlite3/test \
+      test test/testsqlite3/ \
```
ryanking13, line 83:
```suggestion
      test test/test_sqlite3/ \
```

I guess?

ryanking13 (approved):
Very nice! Thanks for your work. I have some minor comments, otherwise LGTM.

Makefile.envs
hoodmane: I'll fix this in a followup PR with a bit more `cpython/Makefile` tidying.

packages/test/meta.yaml
hoodmane: I don't really understand why this didn't cause any failures.

PR #4437, opened 2024-01-28
Pin pytest < 8.0 and make s3 test compatible to moto 5.0 update

ryanking13:
pytest 8.0 and moto 5.0 were released a few hours ago, and they seem to be breaking tests. This is a (temporary) fix.


hoodmane:
Thanks @ryanking13!

hoodmane (approved):

PR #4443, opened 2024-01-28
Split build-recipes and build-recipes-no-deps into separate commands

hoodmane:
There are a lot of options that are only implemented for one or the other of these. We warn about some combinations of options that are ignored but skip a lot of them. Seems best to turn them into separate commands so that the two commands each only accept arguments that change their behavior

- [x] Add / update tests

### DIFF
--- a/docs/project/changelog.md
+++ b/docs/project/changelog.md
@@ -36,7 +36,7 @@ myst:
 - {{ Enhancement }} Updated `pyimport` to support `pyimport("module.attribute")`.
   {pr}`4395`
 
-- {{ Breaking }} Tje `--no-deps` option to `pyodide build-recipes` has been
+- {{ Breaking }} The `--no-deps` option to `pyodide build-recipes` has been
   replaced with a separate subcommand `pyodide build-recipes-no-deps`.
   {pr}`3520`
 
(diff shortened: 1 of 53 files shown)


docs/project/changelog.md
```diff
@@ -36,6 +36,10 @@ myst:
 - {{ Enhancement }} Updated `pyimport` to support `pyimport("module.attribute")`.
   {pr}`4395`
 
+- {{ Breaking }} The `--no-deps` option to `pyodide build-recipes` has been
+  replaced with a separate subcommand `pyodide build-recipes-no-deps`.
+  {pr}`3520`
```
ryanking13, line 41: The pr number is not correct.

ryanking13 (approved):
Thanks! I agree that the `--no-deps` option makes `build-recipes` command behave quite differently, so I think it's okay to split it into two commands. I also felt a little uncomfortable that build-recipes was calling build-recipes internally again, so this change seems reasonable.

`build-recipes-no-deps` seems a bit long, but just `build-recipe` will be too confusing, and we already have `build` for out-of-tree build. So maybe there is no other choice.

hoodmane:
If we come up with a better name we can alias it later. It _would_ be really nice to have tab completion though.

PR #4448, opened 2024-01-30
Update conftest.py for pluggy 1.1

hoodmane:
pluggy 1.1 makes really annoying warnings if the original wrapper hooks raise errors telling us to use the new style wrapper. Unfortunately, the new style wrapper is a bit tricky to use correctly.

I think the old style wrapper is much better for people who care about careful error handling.


ryanking13:
It looks like pluggy 1.1 is yanked. Does this PR applies to versions > 1.1 as well?

image

ryanking13 (approved):
I see. Thanks for handling this.

conftest.py
```diff
@@ -232,18 +232,24 @@ def extra_checks_test_wrapper(browser, trace_hiwire_refs, trace_pyproxies):
...
+            # we override the error message with the better message returned by
+            # a.result() in the finally block.
+            browser.disable_pyproxy_tracing()
+            browser.restore_state()
+        finally:
+            try:
+                if err:
+                    raise err
+            finally:
+                del err
```
hoodmane, line 252: I guess this would be an improvement here:
```suggestion
        # If the test really screwed things up, these cleanup functions might fail.
        # Suppress any errors in this case so we keep the orignal error
        cm = contextlib.suppress(Exception) if err else contextlib.nullcontext()
        with cm:
            browser.disable_pyproxy_tracing()
            browser.restore_state()
```

PR #4450, opened 2024-01-30
build-recipes: Call search_pyodide_root to determine root

hoodmane:
This makes it work all over the source tree. I also increased the default depth for search_pyodide_root to 10 because I think 5 is too small.


ryanking13 (approved):
Thanks!

PR #4257, opened 2023-10-26
Add scipy tests failing on main probably due to stack switching

lesteve:
### Description

This adds tests that create fatal errors, see #4249 for more details. They should  create fatal errors in the CI too.

These tests can be used to in the PR that fixes the issue with stack switching.

Some of them may actually be test failures (e.g. the fft multithreading one is expected to fail since this is fine) but you don't seem to check the return value of pytest.main, so I think it is fine ...

PR #4455, opened 2024-01-30
Fixed Fatal error with CoolProp

### LINKED ISSUE #4451: an error in CoolProp generates a fatal error in Pyodide
## 🐛 Bug

Getting an error or warning in CoolProp crashes pyodide with a fatal error

### To Reproduce
Run the following code in Pyodide REPL
```py
from CoolProp.CoolProp import PropsSI
PropsSI('T','P',101325,'Q',0,'Water')    # yields 373.1242958476844
PropsSI('T','P',101325,'Q',0,'Walter')   # should yield an error
```
The CoolProp error causes a fatal error in Pyodide
```text
CppException CoolProp::CoolPropError<(CoolProp::CoolPropBaseError::ErrCode)4>: key [Walter] was not found in string_to_index_map in JSONFluidLibrary
```
### Expected behavior

Running the example code should generate an error in CoolProp as it does in Python (not crash Pyodide with a fatal error)

```
ValueError: Initialize failed for backend: "?", fluid: "Walter" fractions "[ 1.0000000000 ]"; error: key [Walter] was not found in string_to_index_map in JSONFluidLibrary : PropsSI("T","P",101325,"Q",0,"Walter")
```

### Environment

- Pyodide Version: 0.25.0
- Browser version: Firefox 122.0

### Additional context

First discussed here:
- https://github.com/CoolProp/CoolProp/discussions/2302#discussioncomment-7333820
- https://github.com/mgreminger/EngineeringPaper.xyz/discussions/204#discussioncomment-8308871

dvd101x:
### Description

This fixes #4451 so that CoolProp doesn't throw a Fatal Error to Pyodide

The solution was found and proposed by hoodmane at #4451

Included tests and added a test for Humid Air Properties. (it's my first time writing tests in python so I expect some issues along the way, sorry in advance for the inconvenience)

### Checklists

- [x] Add / update tests


hoodmane (approved):
Looks good to me. Thanks @dvd101x!

hoodmane:
We might need to increase the timeouts for these tests.

ryanking13:
It looks like it is still getting the fatal error in Chrome?

dvd101x:
Yes, it seems like the fatal error in chrome is causing the timeout.

> FAILED coolprop/test_coolprop.py::simple_phasesi[chrome] - pytest_pyodide.runner.JavascriptException: Error: Pyodide already fatally failed and can no longer be used.

> FAILED coolprop/test_coolprop.py::simple_HAPropsSI[chrome] - pytest_pyodide.runner.JavascriptException: Error: Pyodide already fatally failed and can no longer be used.
    at Object.get [as runPython]

hoodmane:
Try adding `@pytest.mark.driver_timeout(60)` to the tests which would raise the timeout to 60s.

hoodmane:
Hmm well that's not a timeout:
```
CppException CoolProp::CoolPropError<(CoolProp::CoolPropBaseError::ErrCode)4>: Initialize failed for backend: "?", fluid: "Walter" fractions "[ 1.0000000000 ]"; error: key [Walter] was not found in string_to_index_map in JSONFluidLibrary
CppException CoolProp::CoolPropError<(CoolProp::CoolPropBaseError::ErrCode)4>: Initialize failed for backend: "?", fluid: "Walter" fractions "[ 1.0000000000 ]"; error: key [Walter] was not found in string_to_index_map in JSONFluidLibrary
```

dvd101x:
That's right, the intention for them is to throw an error without crashing pyodide, as 'Walter' is not a valid fluid name.

ryanking13:
Hmm, its strange that it only crashes in Chrome...

dvd101x:
I agree, it's odd. maybe something from #4257 could aid here?

I see failing of Node and Chrome, I guess it's because they both use the V8 JavaScript Engine.

You might have more experience with these. ¿Any recommendations?

ryanking13:
> I see failing of Node and Chrome, I guess it's because they both use the V8 JavaScript Engine.

Yes probably. If it is a V8 engine issue. There is not much we can do.

One thing I noticed is that when I test this in my local chrome (v122), it worked as expected. We are using a slightly outdated chrome in our CI (v114) so I think we can try updating the chrome first and see what happens.

dvd101x:
Thanks!

I will update from main once in a while.

ryanking13:
I updated the chrome version in the main branch to 122, could you merge main and see if it works?

dvd101x:
Awesome!

Thanks, Gyeongjae!

I did merge

ryanking13:
Okay, I was able to reproduce this locally by enabling JSPI flag in my local chrome.

@hoodmane do you have any idea why enabling JSPI would interpret cpp exception a fatal error, while otherwise it is converted to normal error correctly?

hoodmane:
> any idea why enabling JSPI would interpret cpp exception a fatal error

Yes, it makes sense why a bug in the JSPI code could cause a problem like this. Thanks for noticing this @ryanking13, that's a good catch. I'll try to look into it soon. I'm really hoping to have a polished stack switching implementation for the next release =)

hoodmane:
Thanks @dvd101x!

dvd101x:
Thanks Hood and Gyeongjae!

I'm very grateful for all your help and follow up during this process!

PR #4458, opened 2024-01-31
MAINT Minor pywasmcross optimizations

ryanking13:
### Description

A few minor pywasmcross micro-optimizations. I am not sure how much it would affect the total build time, but it is worth investigating as it is invoked thousands of times for each package.

- replay_f2c: Not used most of the time except for Fortran compilers, so I moved it to `_f2c_fixes`.
- delayed imports + replaced dataclass with namedtuple which is much cheaper to import.
- reordered if-else statements so the common one comes first.
- minor refactorings

#### `python -X importtime` result

__Before__

image

__After__

image

### Checklists

- [ ] Add / update tests


pyodide-build/pyodide_build/pywasmcross.py
```diff
@@ -1,9 +1,5 @@
 #!/usr/bin/env python3
-"""Helper for cross-compiling distutils-based Python extensions.
-
-distutils has never had a proper cross-compilation story. This is a hack, which
-miraculously works, to get around that.
-
+"""
```
hoodmane, line 2: These are still accurate remarks if we just replace "distutils" with "Python":
```suggestion
"""Helper for cross-compiling Python binary extensions.

Python has never had a proper cross-compilation story. This is a hack, which
miraculously works, to get around that.

```

pyodide-build/pyodide_build/pywasmcross.py
```diff
@@ -57,16 +53,12 @@
     __name__ = PYWASMCROSS_ARGS.pop("orig__name__")
 
 
-import dataclasses
-import re
-import shutil
 import subprocess
 from collections.abc import Iterable, Iterator
-from typing import Literal
+from typing import Literal, NamedTuple
```
hoodmane, line 58: It always strikes me as pretty weird that `NamedTuple` is a concrete class that you subclass and it changes your behavior but it comes from the `typing` module.

pyodide-build/pyodide_build/pywasmcross.py
```diff
@@ -610,6 +513,9 @@ def handle_command_generate_args(  # noqa: C901
     elif cmd == "strip":
         line[0] = "emstrip"
         return line
+    elif cmd == "gfortran":
+        tmp = replay_f2c(line)
```
hoodmane, line 517: Don't you have to import `replay_f2c` somewhere? also what happened to:
```py
        if tmp is None:
            # No source file, it's a query for information about the compiler. Pretend we're
            # gfortran by letting gfortran handle it
            return subprocess.run(line).returncode
        line = tmp
```

pyodide-build/pyodide_build/pywasmcross.py
```diff
@@ -534,42 +444,35 @@ def handle_command_generate_args(  # noqa: C901
...
     >>> args = Args(cflags='', cxxflags='', ldflags='', target_install_dir='')
-    >>> handle_command_generate_args(['gcc', 'test.c'], args, False)
+    >>> handle_command_generate_args(['gcc', 'test.c'], args)
     ['emcc', 'test.c', '-Werror=implicit-function-declaration', '-Werror=mismatched-parameter-types', '-Werror=return-type']
     """
     if "-print-multiarch" in line:
         return ["echo", "wasm32-emscripten"]
-    for arg in line:
-        if arg.startswith("-print-file-name"):
-            return line
```
hoodmane, line 557: Why don't we need this anymore?

pyodide-build/pyodide_build/pywasmcross.py
ryanking13: Me too. `NamedTuple` and `TypedDict` surprises always surprises me.

pyodide-build/pyodide_build/pywasmcross.py
ryanking13: emcc now handles this flag, so I think we don't need this anymore:

https://github.com/emscripten-core/emscripten/blob/ec2ce4ce1d1596e3c45bdcd0a8e7cdeb0fe4ab3f/emcc.py#L609

ryanking13:
I think this is ready to be reviewed again.

I additionally renamed the `BuildArgs` in pywasmcross to `CrossCompileArgs` and moved unused parameters from pywasmcross.py to build_env.py to separate them.

pyodide-build/pyodide_build/pywasmcross.py
```diff
@@ -644,7 +550,8 @@ def handle_command_generate_args(  # noqa: C901
 
     # set linker and C flags to error on anything to do with function declarations being wrong.
     # Better to fail at compile or link time.
-    if is_link_command:
+    is_link_cmd = get_library_output(line) is not None
+    if is_link_cmd:
```
hoodmane, line 554: Maybe replace `get_library_output` with `is_link_cmd` since it seems like we don't do anything with the return value of `get_library_output` anymore:
```suggestion
    if is_link_cmd(line):
```

pyodide-build/pyodide_build/pywasmcross.py
```diff
@@ -698,7 +605,15 @@ def handle_command(
...
-    build_args = BuildArgs(**PYWASMCROSS_ARGS)
+    build_args = CrossCompileArgs(
+        pkgname=PYWASMCROSS_ARGS.pop("pkgname"),
+        cflags=PYWASMCROSS_ARGS.pop("cflags"),
+        cxxflags=PYWASMCROSS_ARGS.pop("cxxflags"),
+        ldflags=PYWASMCROSS_ARGS.pop("ldflags"),
+        target_install_dir=PYWASMCROSS_ARGS.pop("target_install_dir"),
+        pythoninclude=PYWASMCROSS_ARGS.pop("pythoninclude"),
+        exports=PYWASMCROSS_ARGS.pop("exports"),
+    )
```
hoodmane, line 616: Why do you use `pop` here?

hoodmane (approved):
Generally looks good to me.

pyodide-build/pyodide_build/pywasmcross.py
ryanking13: Makes sense. Thanks!

pyodide-build/pyodide_build/pywasmcross.py
ryanking13: No special reason. I just saw that `pop` is used elsewhere in pywasmcross, but yeah just accessing the keys will be fine.

ryanking13:
Thanks for the review!

PR #4460, opened 2024-01-31
Update bcrypt to 4.1.2

hoodmane:

### DIFF
--- a/packages/bcrypt/meta.yaml
+++ b/packages/bcrypt/meta.yaml
@@ -1,17 +1,11 @@
 package:
   name: bcrypt
-  version: 4.0.1
+  version: 4.1.2
   top-level:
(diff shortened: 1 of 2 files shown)


hoodmane:
@alex any ideas why this upgrade might be causing test failures? I might have to git bisect it.

alex:
This looks vaguely like running a new bcrypt against an old test file.

hoodmane:
Oh fantastic! Should have checked that. I assumed that the output of functions would be stable between minor versions even though I don't know anything about bcrypts stability guarantees or versioning scheme. I was worried it would be some strange miscompilation.

alex:
The old tests rely on the ability to monkey patch things, and thus are
sensitive to implementation details

PR #4461, opened 2024-02-01
pyodide_build.buildpkg: Do not copy dist/*.whl files from source trees

mkoeppe:
### Description

When a meta.yaml uses `source/path` to build from some source tree, `pyodide build-recipes` copies the source tree to the build directory before building. When the source directory has some `dist/*.whl` files, they are copied as well, which later causes "Exception: Unexpected number of wheels".
Here we filter them out.


hoodmane:
Wow thanks, this bug is really annoying.

hoodmane (approved):

PR #4462, opened 2024-02-01
Update to fiona 1.9.5

hoodmane:


ryanking13 (commented):
Thanks! I'll update the test later so we don't need the test file to run the test.

hoodmane:
Yeah this was a bit high effort for my taste but at least it is less specific to the particular fiona version.

ryanking13:
> Yeah this was a bit high effort for my taste

Sorry, I was lazy when I was first adding fiona. I felt that GDAL (dependency of fiona) was so complex that it needed to be fully tested, but analyzing and porting each test was too cumbersome.

hoodmane:
Yeah these packages are hard work either way. Thanks for porting them! I did notice that there are some C++ exceptions that should probably be wrapped into Python exceptions instead of turning into fatals, like with #4455.

PR #4463, opened 2024-02-01
TST Move run_docker test outside of pyodide-build

ryanking13:
It is weird that `test_run_docker` is inside pyodide-build. This PR moves the run_docker test file to `tools` directory. Maybe it is not the best place to put this test file, but it is at least better than pyodide-build.


hoodmane (approved):

hoodmane:
Thanks!

PR #4465, opened 2024-02-01
Update the `boost-cpp` package to build common libraries

gudzpoz:
### Description

The `boost-cpp` package was originally introduced as a dependency of the Robot Raconteur package, and it has only built `date_time, filesystem, system, regex, chrono, random, program_options` out of a ton of other libraries offered by Boost.

This PR adds options to the build commands in the `meta.yaml` file of `boost-cpp` to build more Boost libraries. The list is extracted by running `./b2 --show-libraries` with certain unsupported libraries excluded (see the comments in `meta.yaml`).

### Checklists

- [not necessary] Add / update tests
- [not necessary] Add new / update outdated documentation

### DIFF
--- a/docs/project/changelog.md
+++ b/docs/project/changelog.md
@@ -53,6 +53,11 @@ myst:
 - Upgraded `libproj` to 9.3.1, `pyproj` to 3.6.1, `h5py` to 3.10.0 {pr}`4426`,
   `packaging` to 23.2, `typing-extensions` to 4.9 {pr}`4428`
 
+- Updated `boost-cpp` config to build Boost.Atomic, Boost.Cobalt, Boost.Exception,
+  Boost.Graph, Boost.GraphParallel, Boost.Headers, Boost.IOStreams, Boost.Locale,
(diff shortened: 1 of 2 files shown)


hoodmane:
This doesn't seem to me to be a good thing by itself. Since this is a static library, it only gets used when it's linked into other binary extensions. If they don't use these features then at best it wastes compile time to build features they don't need. At worst the extra stuff could actually get linked in and bloat binaries with unused code.

gudzpoz:
Thanks! I was actually trying to compile [ledger](https://github.com/ledger/ledger) into a Wasm wheel, which requires extra boost libraries, and having `boost-cpp` build more libraries really helps (in that I don't need to submit a `boost-cpp` patch in my ledger PR).
But, yes, building more libraries incurs build time & maintenance costs. Moreover, the unused ones will then be left untested, which can lead to problems as well. I will close this PR then.

hoodmane:
Well if you have a pr adding a Python module that uses these modules then we can probably accept that. It would be fine to turn these on in the same pr.

PR #4467, opened 2024-02-01
MAINT sysconfigdata updates split from #4435

hoodmane:
Generate sysconfig data as a separate .PHONY build step (it's fast and useful to be able to retrigger so why not). Update sysconfigdata to give better values at build time.


ryanking13 (approved):
Thanks!

PR #4468, opened 2024-02-02
Updates to packages from #4435

hoodmane:
Mostly this is just package upgrades for Python 3.12 compatible versions. We also drop the distutils and nose requirements from every package. Any python 3.12 compatible package won't require them anyways so these are outdated.


ryanking13 (approved):
Thanks for updating packages!

PR #4469, opened 2024-02-02
MAINT Some changes to pyodide-build split from #4435

hoodmane:
Mostly making tests version neutral. I also deleted a few mentions of distutils from tests and documentation.


ryanking13 (approved):
Thanks!

PR #4473, opened 2024-02-02
Update cramjam to 2.8.1

hoodmane:
Split off from #4468 / #4435 for visibility because it's a bit troublesome.

- [ ] Add a [CHANGELOG](https://github.com/pyodide/pyodide/blob/main/docs/project/changelog.md) entry

PR #4475, opened 2024-02-03
Remove cramjam 2.8.1 monkeypatch and fix build issue

ryanking13:
See: https://github.com/pyodide/pyodide/pull/4473#issuecomment-1925029609

There was a upstream fix: https://github.com/gyscos/zstd-rs/pull/209, but was not applied to the latest zstd-rs release.


hoodmane:
Thanks for figuring it out and fixing it @ryanking13!

PR #4478, opened 2024-02-04
make scripts more posix compliant

e-nikolov:
### Description

Not all distros have `/bin/bash` or `/bin/sh`, this PR replaces some of those direct references with `/usr/bin/env bash`


hoodmane:
Thanks!

PR #4480, opened 2024-02-05
Fix WASM_LIBRARY_DIR variable to be compatible for out-of-tree build

ryanking13:
### Description

Move `WASM_LIBRARY_DIR` and `PKG_CONFIG_LIBDIR` variables from Makefile.envs to pyodide-build, so they can be dynamically set according to the build_dir.

When building recipes out-of-tree, there was a bug that WASM_LIBRARY_DIR is pointing the directory under the xbuildenv, this PR fixes it.

### Checklists

- [x] Add / update tests


pyodide-build/pyodide_build/create_xbuildenv.py
```diff
@@ -47,8 +47,9 @@ def get_relative_path(pyodide_root: Path, flag: str) -> Path:
         return Path(get_build_flag(flag)).relative_to(pyodide_root)
 
     pythoninclude = get_relative_path(pyodide_root, "PYTHONINCLUDE")
-    wasm_lib_dir = get_relative_path(pyodide_root, "WASM_LIBRARY_DIR")
     sysconfig_dir = get_relative_path(pyodide_root, "SYSCONFIGDATA_DIR")
+    # TODO: remove libs from the xbuildenv
+    wasm_lib_dir = pyodide_root / "packages" / ".libs"
```
ryanking13, line 52: In the future, I think we should remove library artifacts from `xbuildenv`. I think it's strange that some selected libraries are included here.

WDYT @hoodmane? Are these libraries in the xbuildenv used in numpy or scipy CI?

hoodmane: Not sure. I know that some people are using these, people have requested that we put more things in. I thought they were more for building packages that used numpy/scipy at build time.

ryanking13: > we put more things in

In the long run, We'll probably need to find a separate place to host the built libraries. Of course, this is the same issue as hosting packages.

pyodide-build/pyodide_build/buildpkg.py
```diff
@@ -92,6 +91,10 @@ def __init__(
         self.fullname = f"{self.name}-{self.version}"
         self.build_args = build_args
 
+        self.library_install_prefix = (
+            Path(build_dir).resolve() if build_dir else self.pkg_root.parent
```
hoodmane, line 95: Does `build_dir` refer to `PYODIDE_ROOT/packages`? That is surprising to me, maybe we should rename it?

hoodmane (approved):
Thanks @ryanking13!

pyodide-build/pyodide_build/buildpkg.py
ryanking13: Good point. This should be fixed.

PR #4481, opened 2024-02-05
BLD Run build/post script inside the unpacked wheel directory

ryanking13:
### Description

Changes `build/post` script to run inside the wheel directory, so we don't need to pass an extra `WHEELDIR` env variable.

### Checklists


hoodmane (approved):
Thanks!

PR #4482, opened 2024-02-05
Fix linking of libhacl and tidy up cpython makefile

hoodmane:
I also added a `PY` prefix to the Python build variables to make it a bit clearer.


ryanking13 (approved):
Thanks!

PR #4491, opened 2024-02-08
build-recipes: Add --keep-going option

hoodmane:
This adds a --keep-going option to build-recipes that will continue building packages even if some builds fail.
I think this will make updating the Python version or a bunch of packages less annoying.

As a followup, we should figure out how to stick the failed list somewhere we can easily use to:
 - revert updates to packages that failed to build
 - disable packages that failed to build
 - pin packages that failed to the current version so that auto update ignores them

### Checklist

- [ ] Add / update tests

### DIFF
--- a/pyodide-build/pyodide_build/buildall.py
+++ b/pyodide-build/pyodide_build/buildall.py
@@ -18,9 +18,8 @@ from pathlib import Path
 from queue import PriorityQueue, Queue
 from threading import Lock, Thread
 from time import perf_counter, sleep
-from typing import Any
+from typing import Any, NoReturn
 
-from packaging.utils import canonicalize_name
 from pyodide_lock import PyodideLockSpec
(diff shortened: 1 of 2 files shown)


pyodide-build/pyodide_build/buildall.py
```diff
@@ -200,10 +211,8 @@ def finish(self, success: bool, elapsed_time: float) -> None:
 
         self.finished = True
 
-        if success:
-            logger.success(done_message)
-        else:
-            logger.error(done_message)
+        color = SUCCESS_BLUE if success else FAIL_RED
+        print(color + done_message + END_COLOR)
```
hoodmane, line 215: Using `logger.error` here just leaves an empty line on the console. I couldn't figure out how to get it to print correctly. I think this is a reasonable alternative.

ryanking13 (commented):
Thanks! Indeed this will be useful when we upgrade packages.

PR #4493, opened 2024-02-08
Update bokeh to 3.3.4

ianthomas23:
Update bokeh to 3.3.4. I've checked it builds and run OK locally.


hoodmane:
Thanks @ianthomas23!

Issue #4498, opened 2024-02-09
Numpy v1.26.4 build fails: Module "features" does not exist


hoodmane:
@ryanking13 @mattip @seberg

v1.26.3 build succeeds

Failure for v1.26.4:

<details><summary>Details</summary>
<p>

```
The Meson build system
Version: 1.3.1
Source dir: /home/hood/Documents/programming/pyodide/packages/numpy/build/numpy-1.26.4
Build dir: /home/hood/Documents/programming/pyodide/packages/numpy/build/numpy-1.26.4/.mesonpy-c6ekdglk
Build type: cross build
Project name: NumPy
Project version: 1.26.4
Cross compiler sanity tests disabled via the cross file.
C compiler for the host machine: /tmp/tmpqfmd_k4r/cc (emscripten 3.1.52 "emcc (Emscripten gcc/clang-like replacement + linker emulating GNU ld) 3.1.52 (fa478400df3351f7153c0279bc638784d3d90334)")
C linker for the host machine: /tmp/tmpqfmd_k4r/cc ld.wasm 18.0.0
C++ compiler for the host machine: /tmp/tmpqfmd_k4r/c++ (emscripten 3.1.52 "emcc (Emscripten gcc/clang-like replacement + linker emulating GNU ld) 3.1.52 (fa478400df3351f7153c0279bc638784d3d90334)")
C++ linker for the host machine: /tmp/tmpqfmd_k4r/c++ ld.wasm 18.0.0
Cython compiler for the host machine: cython (cython 3.0.8)
C compiler for the build machine: ccache cc (gcc 9.4.0 "cc (Ubuntu 9.4.0-1ubuntu1~20.04.2) 9.4.0")
C linker for the build machine: cc ld.bfd 2.34
C++ compiler for the build machine: ccache c++ (gcc 9.4.0 "c++ (Ubuntu 9.4.0-1ubuntu1~20.04.2) 9.4.0")
C++ linker for the build machine: c++ ld.bfd 2.34
Cython compiler for the build machine: cython (cython 3.0.8)
Build machine cpu family: x86_64
Build machine cpu: x86_64
Host machine cpu family: wasm32
Host machine cpu: wasm
Target machine cpu family: wasm32
Target machine cpu: wasm
Program python3 found: YES (/tmp/build-env-mr2vjazb/bin/python)
Found pkg-config: YES (/usr/bin/pkg-config) 0.29.1
Run-time dependency python found: YES 3.12
[...]

PR #4500, opened 2024-02-09
Update most packages

hoodmane:
Except:

* geopandas -- we're running too many tests after the update so it times out, have to change the geopandas tests to do less.
* lightgbm
* matplotlib #4510
* numcodecs
* numpy: the release from this week fails to build #4498
* pillow_heif
* protocol_buffer
* pyinstrument
* python-sat
* rebound
* reboundx
* river
* scipy (#4499)
* shapely (#4503)
* sparseqr
* zengl (#4509)


packages/shapely/meta.yaml
```diff
@@ -1,13 +1,11 @@
...
   name: shapely
-  version: 1.8.2
+  version: 2.0.2
   top-level:
     - shapely
 source:
-  url: https://files.pythonhosted.org/packages/93/3c/cda77e57a08c49569de5bd90376c547bcb981420100adcb0f3770ed681b1/Shapely-1.8.2.tar.gz
-  sha256: 572af9d5006fd5e3213e37ee548912b0341fb26724d6dc8a4e3950c10197ebb6
-  patches:
-    - patches/0001-Search-libs-from-auditwheel-repaired-directory.patch
```
hoodmane, line 10: @ryanking13 Maybe dropping this patch is going to be a problem?

packages/pydantic/meta.yaml
```diff
@@ -1,11 +1,11 @@
 package:
   name: pydantic
-  version: 1.10.7
+  version: 2.6.1
   top-level:
     - pydantic
 source:
-  sha256: cfc83c0678b6ba51b0532bea66860617c4cd4251ecf76e9846fa5a9f3454e97e
-  url: https://files.pythonhosted.org/packages/43/5f/e53a850fd32dddefc998b6bfcbda843d4ff5b0dcac02a92e414ba6c97d46/pydantic-1.10.7.tar.gz
+  sha256: 4fd5c182a2488dc63e6d32737ff19937888001e2a6d86e94b3f233104a5d1fa9
```
hoodmane, line 7: This might not work, huh? We are missing `pydantic-core` which I think is now a rust package?

ryanking13: Yes, need to build `pydantic-core` to make pydantic 2.x work.

packages/shapely/meta.yaml
ryanking13: That patch needs to be dropped. Let me see if I can update shapely.

ianthomas23:
If you are updating matplotlib to 3.8.2 then it should also have contourpy as a new runtime dependency (see https://github.com/matplotlib/matplotlib/blob/4591c5b8c5a7b8b721f8173ba5d7ed7751fe6922/setup.py#L332). If you don't want to do it in this PR I will submit a separate one for it afterwards.

hoodmane:
Thanks for pointing that out @ianthomas23, I added it.

ryanking13:
Thanks for updating the packages!

PR #4508, opened 2024-02-10
Add numpy as host requirement for several packages

bartbroere:
I ran the build commands in the Docker container found in the Circle CI configuration, but the `gsw` package seems to be needing `numpy` installed in the host when building.

Update: same for the `tskit` package it seems

Update: `msprime` as well. More might follow.


hoodmane:
Thank!

PR #4510, opened 2024-02-10
Update matplotlib to 3.8.4

hoodmane:
Requires an update to matplotlib-pyodide before it will work: https://github.com/pyodide/matplotlib-pyodide/pull/52


hoodmane:
Most of the test failures seem to be `cannot import name 'maxdict' from 'matplotlib.cbook` which I think was fixed by https://github.com/pyodide/matplotlib-pyodide/pull/52. But there are also:

```
FAILED matplotlib/test_matplotlib.py::plot[chrome] - AttributeError: 'NoneType' object has no attribute 'get_element'
FAILED matplotlib/test_matplotlib.py::svg[chrome] - assert 15016 == 14998
```
which need to be investigated.

ryanking13:
> FAILED matplotlib/test_matplotlib.py::svg[chrome] - assert 15016 == 14998

The size of svg can change based on the implementation details, so I think we can just update the value or remove that assertion.

agriyakhetarpal:
Yes, I'm not too concerned about that test either. I patched `matplotlib-pyodide` to fix the missing `maxdict` attribute, so this is what is left for the rest of the tests:

```bash
___________________ test_draw_image_affine_transform[chrome] ___________________
packages/matplotlib/test_matplotlib.py:305: in test_draw_image_affine_transform
    run(selenium, handle, ref)
packages/matplotlib/test_matplotlib.py:296: in run
    plt.show()
/lib/python3.12/site-packages/matplotlib/pyplot.py:527: in show
    ???
/lib/python3.12/site-packages/matplotlib_pyodide/html5_canvas_backend.py:453: in show
    ???
/lib/python3.12/site-packages/matplotlib_pyodide/browser_backend.py:183: in show
    ???
E   AttributeError: 'NoneType' object has no attribute 'get_element'
```

ryanking13:
> FAILED matplotlib/test_matplotlib.py::plot[chrome] - AttributeError: 'NoneType' object has no attribute 'get_element'

This error happens [here](https://github.com/pyodide/matplotlib-pyodide/blob/18cd638fb0b70078c0e6bedc4a98331ac7652f33/matplotlib_pyodide/browser_backend.py#L183C24-L183C31), and it is [set to None](https://github.com/matplotlib/matplotlib/blob/8fb842c78f1b57b13a3d0fe63e83f3d31694f887/lib/matplotlib/backend_bases.py#L1733) when initialized.

Maybe we can just add a `if` block to check the toolbar and skip the logic if it is None.

agriyakhetarpal:
Let's see if these commits fix the tests. I have to rely on the CI because it's difficult to work with the browser tests locally – they fail consistently with timeouts for me (it's difficult to find versions even just slightly old off of Homebrew)

agriyakhetarpal:
Okay, we got down to 8 failures out of 13 tests... now we have `ValueError: 'bitmap' is not a valid value for output; supported values are 'path', 'agg', and 'macosx'`. I'll keep looking into it.

agriyakhetarpal:
I've hacked something together based on https://github.com/ipython/ipython/pull/12889 and an attempt at a path-based MathTextParser because "bitmap" was deprecated since version 3.4, so it has been removed by 3.8 (or maybe in an earlier version). My Copilot's code completions seem to help a bit (the code looks wrong at so many places, until it's coerced to do what I want it to do :P). Is there a reliable way to go through the code -> test -> debug -> code loop, and faster? For such a change, testing in CI is going to be quite slow.

ryanking13:
> Is there a reliable way to go through the code -> test -> debug -> code loop, and faster? For such a change, testing in CI is going to be quite slow.

I mostly use the docker image when running tests, which has the same environment (OS, browser versions, etc). If running tests locally doesn't fit for you, an (not that optimal) alternative would be updating the CI script to build only related packages.

agriyakhetarpal:
I added a job to build Matplotlib in the same way we build SciPy using commit messages. I think this should work for now, I'll remove it when we are done with this PR and it is ready to merge.

Note to self, tasks left:
- [ ] Fix Matplotlib tests with patches
- [ ] Get a 0.2.3 release out for `matplotlib-pyodide` so that we can drop the patches
- [ ] Drop Matplotlib commit-message-based builds

agriyakhetarpal:
Okay, looks like this is working because all the tests are passing with Chrome ([workflow run](https://github.com/pyodide/pyodide/actions/runs/12166983368/job/33937382370?pr=4510#step:14:166)) 🎉 Also, they work with Chrome 133 as well, we might be able to unpin Chrome 125 added in pyodide/matplotlib-pyodide#62 once we have 0.27 stable) and we can now remove an xfail (I've already removed that one in pyodide/matplotlib-pyodide#62) because it xpasses now. I will still wait for CircleCI so that we can look at the results from Firefox as well.

@ryanking13, I think I need a bit of help here again – do you know of a way that I can do some visual testing for this upgrade? How do you usually do so? As long as the rendering is alright, we should be good to go. I'd like to validate that the `MathTextParser` changes. As you recently told me, I can't use a local JupyterLite deployment for this because it uses a different backend.

Once we've verified this, I can merge pyodide/matplotlib-pyodide#64 and create a new release.

ryanking13:
Thanks for your work!

> @ryanking13, I think I need a bit of help here again – do you know of a way that I can do some visual testing for this upgrade? How do you usually do so? As long as the rendering is alright, we should be good to go. I'd like to validate that the MathTextParser changes. As you recently told me, I can't use a local JupyterLite deployment for this because it uses a different backend.

For visual testing, I think I would download the built distribution in this PR, write a simple HTML file that renders matplotlib plot and see if it work (maybe using multiple browsers)

agriyakhetarpal:
I'm not sure how the tests are passing, because building locally and trying to test it in my browser gives me a function signature mismatch:

```wasm
pyodide.asm.js:8225 Stack (most recent call first):
pyodide.asm.js:8225   File "/lib/python3.12/site-packages/matplotlib_pyodide/browser_backend.py", line 65 in _add_matplotlib_styles
pyodide.asm.js:8225   File "/lib/python3.12/site-packages/matplotlib_pyodide/browser_backend.py", line 30 in __init__
pyodide.asm.js:8225   File "/lib/python3.12/site-packages/matplotlib_pyodide/wasm_backend.py", line 31 in __init__
pyodide.asm.js:8225   File "/lib/python3.12/site-packages/matplotlib/backend_bases.py", line 2690 in create_with_canvas
pyodide.asm.js:8225   File "/lib/python3.12/site-packages/matplotlib/backend_bases.py", line 1785 in new_manager
pyodide.asm.js:8225   File "/lib/python3.12/site-packages/matplotlib/backend_bases.py", line 3416 in new_figure_manager_given_figure
pyodide.asm.js:8225   File "/lib/python3.12/site-packages/matplotlib/backend_bases.py", line 3411 in new_figure_manager
pyodide.asm.js:8225   File "/lib/python3.12/site-packages/matplotlib/pyplot.py", line 465 in new_figure_manager
pyodide.asm.js:8225   File "/lib/python3.12/site-packages/matplotlib/pyplot.py", line 934 in figure
pyodide.asm.js:8225   File "<exec>", line 8 in <module>
pyodide.asm.js:8225   File "/lib/python312.zip/_pyodide/_base.py", line 411 in run_async
pyodide.asm.js:8225   File "/lib/python312.zip/_pyodide/_base.py", line 597 in eval_code_async
pyodide.asm.js:8225   File "/lib/python312.zip/asyncio/events.py", line 88 in _run
pyodide.asm.js:8225   File "/lib/python312.zip/pyodide/webloop.py", line 332 in run_handle
pyodide.asm.js:4916 Uncaught (in promise) RuntimeError: null function or function signature mismatch
    at pyodide.asm.wasm._Py_Dealloc (pyodide.asm.wasm:0x21a6c7)
    at pyodide.asm.wasm.JsProxy_GetAttr_helper (pyodide.asm.wasm:0x16fef7)
    at pyodide.asm.wasm.JsProxy_GetMethod (pyodide.asm.wasm:0x16fa17)
    at pyodide.asm.wasm._PyObject_GetMethod (pyodide.asm.wasm:0x2191f3)
    at pyodide.asm.wasm._PyEval_EvalFrameDefault (pyodide.asm.wasm:0x2ade9e)
    at pyodide.asm.wasm._PyEval_Vector (pyodide.asm.wasm:0x2af75b)
    at pyodide.asm.wasm._PyFunction_Vectorcall (pyodide.asm.wasm:0x1cd34a)
    at pyodide.asm.wasm._PyObject_FastCallDictTstate (pyodide.asm.wasm:0x1cc874)
    at pyodide.asm.wasm._PyObject_Call_Prepend (pyodide.asm.wasm:0x1cd56b)
    at pyodide.asm.wasm.slot_tp_init (pyodide.asm.wasm:0x23331f)
```

and here's the HTML for what I'm using for testing (I've commented out the rest of the Matplotlib code, it seems to fail right at `plt.figure()` itself:

<details><summary>Tap to expand</summary>
<p>

```html
<!DOCTYPE html>
<html>
  <head>
    <title>Agriya debug</title>
    <style>
      body {
        font-family: system-ui, -apple-system, sans-serif;
        max-width: 1200px;
        margin: 0 auto;
        padding: 20px;
      }
      .output {
        margin: 20px 0;
        padding: 10px;
        border: 1px solid #ccc;
        border-radius: 4px;
      }
      .status {
        color: #666;
        font-style: italic;
      }
    </style>
  </head>
  <body>
    <h1>Some matplotlib MathText testing</h1>
    <div id="status" class="status">Loading Pyodide...</div>
    <div id="output" class="output"></div>

    <script>
      const indexURL = "./";

      async function main() {
        const status = document.getElementById("status");
        try {
          const pyodide = await loadPyodide({ indexURL });
          status.textContent = "Loading packages...";

          await pyodide.loadPackage([
            "micropip",
            "numpy",
            "matplotlib",
            "matplotlib_pyodide",
          ]);

          status.textContent = "Running tests...";

          const output = await pyodide.runPythonAsync(`
                    import matplotlib
                    import numpy as np

                    matplotlib.use("module://matplotlib_pyodide.html5_canvas_backend")

                    plt.figure(figsize=(8, 6))
                    # plt.text(0.5, 0.9, r'$\\alpha^2 + \\beta^2 = \\gamma^2$', 
                    #        fontsize=20, ha='center')
                    # plt.text(0.5, 0.7, r'$\\sum_{n=1}^{\\infty} \\frac{1}{n^2} = \\frac{\\pi^2}{6}$',
                    #         fontsize=20, ha='center')
                    # plt.text(0.5, 0.5, r'$\\int_0^{\\infty} e^{-x^2} dx = \\frac{\\sqrt{\\pi}}{2}$',
                    #         fontsize=20, ha='center')
                    # plt.text(0.5, 0.3, r'$\\frac{d}{dx}[x^n] = nx^{n-1}$',
                    #         fontsize=20, ha='center')
                    # plt.axis([0, 1, 0, 1])
                    # plt.title('Basic Mathematical Expressions')
                    # plt.grid(True)
                    # plt.show()

                    # x = np.linspace(-2*np.pi, 2*np.pi, 100)
                    # y = np.sin(x)

                    # plt.figure(figsize=(8, 6))
                    # plt.plot(x, y)
                    # plt.title(r'$f(x) = \\sin(x)$')
                    # plt.xlabel(r'$x$ (radians)')
                    # plt.ylabel(r'$\\sin(x)$')
                    # plt.grid(True)
                    # plt.show()

                    # plt.figure(figsize=(8, 6))
                    # plt.text(0.5, 0.8, r'$\\left(\\frac{\\partial^2}{\\partial x^2} + \\frac{\\partial^2}{\\partial y^2}\\right)\\Psi = 0$',
                    #         fontsize=18, ha='center')
                    # plt.text(0.5, 0.6, r'$\\oint_C \\vec{B} \\cdot d\\vec{l} = \\mu_0 \\left(I_{enc} + \\epsilon_0 \\frac{d\\Phi_E}{dt}\\right)$',
                    #         fontsize=16, ha='center')
                    # plt.text(0.5, 0.4, r'$\\hat{H} \\Psi = E \\Psi$',
                    #         fontsize=18, ha='center')
                    # plt.text(0.5, 0.2, r'$\\Delta x \\Delta p \\geq \\frac{\\hbar}{2}$',
                    #         fontsize=18, ha='center')
                    # plt.axis([0, 1, 0, 1])
                    # plt.title('Complex Mathematical Expressions')
                    # plt.grid(True)
                    # plt.show()
                `);

          status.textContent = "Tests completed successfully!";
        } catch (error) {
          status.textContent = "Error: " + error.message;
          console.error(error);
        }
      }

      const script = document.createElement("script");
      script.src = "./pyodide.js";
      script.onload = main;
      document.head.appendChild(script);
    </script>
  </body>
</html>
```

</p>
</details> 

and I'm not sure how to debug this – it's not obvious to me at the time of writing. The `_add_matplotlib_styles` function seems to be needed only in the JS code; I couldn't find a reference for it and how it's working. What I understand from `JsProxy_GetAttr_helper` being in the trace is that this isn't inside Python; the mismatch is between the proxy layer where Python/Pyodide interacts with the document. One way to fix this could be to defer the JS interaction to create styles only when the document is ready, but I tried that and didn't get anywhere... The error happens with both the WASM and the HTML5 backends, but did not show up in the tests.

agriyakhetarpal:
Goes without saying, any help is appreciated :)

ryanking13:
Thanks, Agriya. I tried reproducing the error with your example, but it works fine in my local environment. That's weird.

I only get some event-handler-related issues that are not very critical, I guess

image

ryanking13:
Using `wasm_backend` works fine in my local environment too.

image

agriyakhetarpal:
That's certainly something, but I'm glad that it's working on your end. Let me take a look in a moment with other browsers. It might also be the case that I'm using [Arc](https://arc.net/), which I feel is a nice browser to use but it is a custom skin over Chromium, after all. I'll also take a look at the `<...>_notify_event` missing attributes, but I'll have to reproduce them first :D

agriyakhetarpal:
Nope, the same signature mismatch exists for me on Chrome, Firefox, Arc, and Safari. Though, on Safari, the error is phrased slightly differently, but it just conveys the same error because the stack trace is identical:

```js
[Error] RuntimeError: call_indirect to a null table entry (evaluating '__pyproxy_apply(
              ptrobj,
              jsargs,
              num_pos_args,
              kwargs_names,
              num_kwargs
            )')
	(anonymous function) (pyodide.asm.js:3733)
	(anonymous function) (pyodide.asm.js:3772)
	callPyObjectKwargs (pyodide.asm.js:4842)
	(anonymous function) (pyodide.asm.js:4912)
	wrapper (pyodide.asm.js:1941)
	(anonymous function) (pyodide.asm.js:8448)
[Log] Stack (most recent call first): (pyodide.asm.js, line 8398)
[Log]   File "/lib/python3.12/site-packages/matplotlib_pyodide/browser_backend.py", line 63 in _add_matplotlib_styles (pyodide.asm.js, line 8398)
[Log]   File "/lib/python3.12/site-packages/matplotlib_pyodide/browser_backend.py", line 30 in __init__ (pyodide.asm.js, line 8398)
[Log]   File "/lib/python3.12/site-packages/matplotlib_pyodide/html5_canvas_backend.py", line 48 in __init__ (pyodide.asm.js, line 8398)
[Log]   File "/lib/python3.12/site-packages/matplotlib/backend_bases.py", line 2690 in create_with_canvas (pyodide.asm.js, line 8398)
[Log]   File "/lib/python3.12/site-packages/matplotlib/backend_bases.py", line 1785 in new_manager (pyodide.asm.js, line 8398)
[Log]   File "/lib/python3.12/site-packages/matplotlib/backend_bases.py", line 3416 in new_figure_manager_given_figure (pyodide.asm.js, line 8398)
[Log]   File "/lib/python3.12/site-packages/matplotlib/backend_bases.py", line 3411 in new_figure_manager (pyodide.asm.js, line 8398)
[Log]   File "/lib/python3.12/site-packages/matplotlib/pyplot.py", line 465 in new_figure_manager (pyodide.asm.js, line 8398)
[Log]   File "/lib/python3.12/site-packages/matplotlib/pyplot.py", line 934 in figure (pyodide.asm.js, line 8398)
[Log]   File "<exec>", line 9 in <module> (pyodide.asm.js, line 8398)
[Log]   File "/lib/python312.zip/_pyodide/_base.py", line 411 in run_async (pyodide.asm.js, line 8398)
[Log]   File "/lib/python312.zip/_pyodide/_base.py", line 597 in eval_code_async (pyodide.asm.js, line 8398)
[Log]   File "/lib/python312.zip/asyncio/events.py", line 88 in _run (pyodide.asm.js, line 8398)
[Log]   File "/lib/python312.zip/pyodide/webloop.py", line 332 in run_handle (pyodide.asm.js, line 8398)
```

Could I ask operating system + browser are you using, @ryanking13? Also, @hoodmane, could you please help test it out on your end, too? I even tried with two different live server extensions – [Live Server](https://marketplace.visualstudio.com/items?itemName=ritwickdey.LiveServer) and [Five Server](https://marketplace.visualstudio.com/items?itemName=yandeu.five-server) and with a simple `python -m http.server` and I get the same problem.

agriyakhetarpal:
(Also, at the risk of stating the obvious, I wanted to confirm that you've indeed been testing with this branch, i.e., with Matplotlib 3.8.4, in case that was missed by chance?)

ryanking13:
> (Also, at the risk of stating the obvious, I wanted to confirm that you've indeed been testing with this branch, i.e., with Matplotlib 3.8.4, in case that was missed by chance?)

Yes, I downloaded the built artifact from the `build-packages` CI output.

One question is, I see the stack trace `line 65 in _add_matplotlib_styles`, but it is an empty block in the file that I downloaded (from https://github.com/pyodide/pyodide/pull/4510/commits/9fc863b689fd071738109a322d2411bfc527b764)

ryanking13:
> Could I ask about the operating system + browser you are using, @ryanking13?

I am testing it with Windows 10 + [Chrome 131, 132 (beta), and Firefox 133]

agriyakhetarpal:
Downloading the built artifacts from the `build-packages` job (https://app.circleci.com/pipelines/github/hoodmane/pyodide/6606/workflows/8a0d066e-42c9-438b-b8a7-0292205e00ed/jobs/100146/artifacts) and using them for testing makes it "almost" work. This is most certainly a problem with the Pyodide distribution I compiled locally on my own machine; we should be good to proceed. It would be nice if we could find out why; unfortunately, I can't see any special differences here (both the CI job and I are using `PYODIDE_DEBUG`).

As for why I say "almost" here – the WASM backend works and displays the equations as you were able to see in https://github.com/pyodide/pyodide/pull/4510#issuecomment-2524844639, but the HTML5 backend doesn't display them...

ryanking13:
> Since the HTML5 backend isn't fully working: to not delay the 0.27 release any further, would it be a good idea to disable it completely and just let users rely on the WASM backend, and consider re-enabling this backend after the release? One regression would be that we can't rely on user-supplied fonts anymore, but I'm not aware of any others because of which the HTML5 backend was brought up.

Yep, I am fine with it, as I even proposed to remove the html5 backend at all.

> I'd love to receive additional help with this, because I'm not particularly acquainted with the matplotlib-pyodide codebase (and none of us are, either), and I am slightly constrained with being able to test it iteratively because of the reasons above. :)

Yes, after 0.27 release, let's think of how we can maintain matplotlib-pyodide better without putting too much effort

agriyakhetarpal:
Thanks, noted – I'll spend the rest of the day today trying to get this working, and if it doesn't, we can disable the backend and proceed. I think that the notify event failures might be masking what is needed to display the text. Though, my guess is that the backend is working (since the tests pass), it's just not able to display the fonts on the canvas for some reason. Maybe I've missed something. I pushed the changes to https://github.com/pyodide/matplotlib-pyodide/pull/64 as well in case you wish to review (and it's easier than reading through the changes from patch files here, I assume).

agriyakhetarpal:
Thanks, noted – I'll spend the rest of the day today trying to get this working, and if it doesn't, we can disable the backend and proceed. I think that the notify event failures might be masking what is needed to display the text. Though, my guess is that the backend is working (since the tests pass), it's just not able to display the fonts on the canvas for some reason. Maybe I've missed something that you can catch. I also pushed the changes to https://github.com/pyodide/matplotlib-pyodide/pull/64 as well in case you wish to review (and it's easier than reading through the changes from patch files here, I assume).

agriyakhetarpal:
A [code search](https://github.com/search?q=matplotlib.use(%22module://matplotlib_pyodide.html5_canvas_backend%22)&type=code) shows lots of uses of this backend in the wild, though!

ryanking13:
> A [code search](https://github.com/search?q=matplotlib.use(%22module://matplotlib_pyodide.html5_canvas_backend%22)&type=code) shows lots of uses of this backend in the wild, though!

Hmm, yeah indeed.

Maybe we can redirect html5_canvas_backend to wasm_backend to minimize the impact? Some parts of the backends would be incompatible, but most of them would be compatible... at least I hope.

agriyakhetarpal:
> Maybe we can redirect html5_canvas_backend to wasm_backend to minimize the impact? Some parts of the backends would be incompatible, but most of them would be compatible... at least I hope.

I think so. For this, do you mean that we should refactor the HTML5 backend classes so that they inherit from the WASM backend, or that we should do it externally? Could you please clarify?

agriyakhetarpal:
With the (temporary) additional debugging I added, it is revealed that the parser is indeed working as we expect it to; it's just that the fonts are missing on the canvas:

Additional debugging notes from debugging the MathTextParser outputs for matplotlib-pyodide

So we probably need to look at why they could be missing...

agriyakhetarpal:
Some additional logging + CI tests reveal:

```
Starting math text for: '$W^{3\beta}_{\delta_1 \rho_1 \sigma2} = U^{3\beta}{\delta_1 \rho_1} + \frac{1}{8 \pi 2} \int^{\alpha2}{\alpha_2} d \alpha^\prime2 \left[\frac{ U^{2\beta}{\delta_1 \rho_1} - \alpha^\prime2U^{1\beta}{\rho_1 \sigma2} }{U^{0\beta}{\rho_1 \sigma_2}}\right]$'
Initial position: (411.0, -1180.075)
Parsed text dimensions: 1048.0x129.0 (depth: 46.0)
Number of glyphs: 58
Number of rectangles: 2
DPI: 200.0, Canvas height: 1400
Transformed y-coordinate: 2580.075
Path rendering failed, falling back to image: currentTransform
```

so the text rendering starts successfully, and it seems to happen correctly. However, it then fails when trying to access the `currentTransform` property of the canvas context (which I used for debugging improperly), but I guess there's a greater issue here.

ryanking13:
> do you mean that we should refactor the HTML5 backend classes so that they inherit from the WASM backend, or that we should do it externally? Could you please clarify?

I was thinking of replacing the [following properties](https://github.com/pyodide/matplotlib-pyodide/blob/90638fad1feebd2d68b9a5d6e8064512adbd3941/matplotlib_pyodide/wasm_backend.py#L109) that defines the backend, and remove all other "unused" parts.

```diff
@_Backend.export
class _BackendHTMLCanvas(_Backend):
-    FigureCanvas = FigureCanvasHTMLCanvas
-    FigureManager = FigureManagerHTMLCanvas
+    FigureCanvas = FigureCanvasAggWasm
+    FigureManager = FigureManagerAggWasm

    @staticmethod
    def show(*args, **kwargs):
        from matplotlib import pyplot as plt

        plt.gcf().canvas.show(*args, **kwargs)

    @staticmethod
    def destroy(*args, **kwargs):
        from matplotlib import pyplot as plt

        plt.gcf().canvas.destroy(*args, **kwargs)
```

agriyakhetarpal:
Thanks for the pointer – redirected the backend, and both the HTML5 and WASM backends now use the WASM backend underneath. I've confirmed the changes with https://github.com/pyodide/pyodide/actions/runs/12354843167/artifacts/2326627934 as well. Let's move ahead with this PR in a moment!

agriyakhetarpal:
https://github.com/pyodide/matplotlib-pyodide/issues/60 has been completed, so I've updated `matplotlib-pyodide` to version 0.2.3 in b20bc94257f1bf4639002063923032fb8211305e. I would have liked to bump right up to Matplotlib 3.9.2, but I am not too keen on it – we could have a 0.27.1 release for that and a bunch of other package updates, perhaps...

packages/matplotlib/meta.yaml
```diff
@@ -48,6 +46,7 @@ requirements:
   host:
     - numpy
   run:
+    - contourpy
```
agriyakhetarpal, line 49:
```suggestion
```

I think I added this in error when testing something 😅

.github/workflows/main.yml
```diff
@@ -340,7 +410,7 @@ jobs:
       - name: Download build artifact
         uses: actions/download-artifact@v4
         with:
-          name: core-build-${{ runner.os }}
+          pattern: core-build-${{ runner.os }}_tag_core_numpy
```
agriyakhetarpal, line 413:
```suggestion
          name: core-build-${{ runner.os }}
```

(Reverting my changes)

.github/workflows/main.yml
```diff
@@ -310,7 +380,7 @@ jobs:
       - name: Download build artifact
         uses: actions/download-artifact@v4
         with:
-          name: core-build-${{ runner.os }}
+          name: core-build-${{ runner.os }}_tag_core_numpy
```
agriyakhetarpal, line 383:
```suggestion
          name: core-build-${{ runner.os }}
```

(Reverting my changes)

.github/workflows/main.yml
```diff
@@ -295,6 +338,33 @@ jobs:
...
+
+          COMMIT_MSG=$(git log --no-merges -1 --oneline)
+
+          # The matplotlib tests will be triggered on push or on pull_request when the commit
+          # message contains "[matplotlib]"
+          if [[ "$GITHUB_EVENT_NAME" == push ||
+                "$COMMIT_MSG" =~ \[matplotlib\] ]]; then
+              echo "trigger=true" >> "$GITHUB_OUTPUT"
+          fi
+
```
agriyakhetarpal, line 367:
```suggestion
```

(Reverting my changes)

.github/workflows/main.yml
```diff
@@ -277,7 +320,7 @@ jobs:
       - name: Download build artifact
         uses: actions/download-artifact@v4
         with:
-          name: core-build-${{ runner.os }}
+          pattern: core-build-${{ runner.os }}_*scipy*
```
agriyakhetarpal, line 323:
```suggestion
          name: core-build-${{ runner.os }}
```

.github/workflows/main.yml
```diff
@@ -178,7 +220,8 @@ jobs:
       - name: Download build artifact
         uses: actions/download-artifact@v4
         with:
-          name: core-build-${{ runner.os }}
+          # Only one build artifact is expected for macOS
+          pattern: core-build-${{ runner.os }}
```
agriyakhetarpal, line 224:
```suggestion
          name: core-build-${{ runner.os }}
```

.github/workflows/main.yml
```diff
@@ -150,13 +176,29 @@ jobs:
...
+        run: tar cjf pyodide-with-matplotlib.tar.gz dist/
+
+      # Temporary for this PR: remove this when merging
+      - name: Upload Pyodide distribution with Matplotlib (temporary)
+        if: contains(matrix.pyodide_packages, 'matplotlib')
+        uses: actions/upload-artifact@v4
+        with:
+          name: pyodide-with-matplotlib
+          path: pyodide-with-matplotlib.tar.gz
+
```
agriyakhetarpal, line 201:
```suggestion
```

(Reverting my changes)

.github/workflows/main.yml
```diff
@@ -150,13 +176,29 @@ jobs:
...
           ./tools/check_compressed_size.py dist/pyodide.asm.* dist/python_stdlib*
 
+      # Temporary for this PR: remove this when merging
+      - run: echo "PYODIDE_ARTIFACT_SUFFIX=$(echo '${{ matrix.pyodide_packages }}' | tr ':,' '_')" >> "$GITHUB_ENV"
+
       - name: Store artifacts build
         uses: actions/upload-artifact@v4
         with:
-          name: core-build-${{ runner.os }}
+          name: core-build-${{ runner.os }}_${{ env.PYODIDE_ARTIFACT_SUFFIX }}
```
agriyakhetarpal, line 185:
```suggestion
          name: core-build-${{ runner.os }}
```

.github/workflows/main.yml
```diff
@@ -150,13 +176,29 @@ jobs:
           pip install brotli
           ./tools/check_compressed_size.py dist/pyodide.asm.* dist/python_stdlib*
 
+      # Temporary for this PR: remove this when merging
+      - run: echo "PYODIDE_ARTIFACT_SUFFIX=$(echo '${{ matrix.pyodide_packages }}' | tr ':,' '_')" >> "$GITHUB_ENV"
+
```
agriyakhetarpal, line 181:
```suggestion
```

(Reverting my changes)

.github/workflows/main.yml
```diff
@@ -142,6 +145,29 @@ jobs:
...
+          runner: selenium
+          browser: chrome
+
+      - name: Test Matplotlib
+        if: contains(matrix.pyodide_packages, 'matplotlib')
+        shell: bash -l {0}
+        run: |
+          which pytest
+          pytest -svra packages/matplotlib/test_matplotlib.py --rt chrome
+
```
agriyakhetarpal, line 170:
```suggestion
```

(Reverting my changes)

.github/workflows/main.yml
```diff
@@ -58,10 +58,13 @@ jobs:
...
           - os: ubuntu-latest
             pyodide_packages: "tag:core,numpy${{ needs.test-scipy-trigger.outputs.test-scipy == 'true' && ',scipy' || '' }}"
+            # Temporary for this PR: remove this when merging
+          - os: ubuntu-latest
+            pyodide_packages: "tag:core,numpy${{ needs.test-matplotlib-trigger.outputs.test-matplotlib == 'true' && ',matplotlib' || '' }}"
           - os: macos-14
             pyodide_packages: "tag:core,numpy"
     runs-on: ${{ matrix.os }}
-    needs: [test-scipy-trigger]
+    needs: [test-scipy-trigger, test-matplotlib-trigger]
```
agriyakhetarpal, line 67:
```suggestion
    needs: [test-scipy-trigger]
```

(Reverting my changes)

.github/workflows/main.yml
```diff
@@ -58,10 +58,13 @@ jobs:
         include:
           - os: ubuntu-latest
             pyodide_packages: "tag:core,numpy${{ needs.test-scipy-trigger.outputs.test-scipy == 'true' && ',scipy' || '' }}"
+            # Temporary for this PR: remove this when merging
+          - os: ubuntu-latest
+            pyodide_packages: "tag:core,numpy${{ needs.test-matplotlib-trigger.outputs.test-matplotlib == 'true' && ',matplotlib' || '' }}"
```
agriyakhetarpal, line 63:
```suggestion
```

(Reverting my changes)

ianthomas23:
You definitely need `contourpy` as a dependency:
https://github.com/matplotlib/matplotlib/blob/8ac84ecb06ffa735f3da578f25aee1ee6f213cac/pyproject.toml#L33-L43

agriyakhetarpal:
> You definitely need `contourpy` as a dependency: [matplotlib/matplotlib@`8ac84ec`/pyproject.toml#L33-L43](https://github.com/matplotlib/matplotlib/blob/8ac84ecb06ffa735f3da578f25aee1ee6f213cac/pyproject.toml#L33-L43)

Yeah, noticed it from a local build just now, I guess that's why I added it back then. Added again in 53786f3166bc62c56a9b123b481c0f61f3e1c0e6.

agriyakhetarpal:
We should be good to merge this once CI passes. @hoodmane, it is your PR so I can't seek a review from you from the interface directly.

packages/matplotlib/extras/fontlist.json
hoodmane, line 1: I haven't read the discussion on the PR so apologies if this was answered somewhere. But why do we need this file now?

packages/matplotlib/meta.yaml
hoodmane: Did you mean to remove this? Or should it actually be here?

packages/matplotlib/test_matplotlib.py
```diff
@@ -113,7 +113,7 @@ def test_svg(selenium):
     plt.savefig(fd, format="svg")
 
     content = fd.getvalue().decode("utf8")
-    assert len(content) == 14998
+    assert len(content) == 15016
```
hoodmane, line 116: This is not a good test. But we didn't make it any worse in this PR.

packages/matplotlib/meta.yaml
hoodmane: Okay I see @ianthomas23's comment now.

packages/matplotlib/extras/fontlist.json
agriyakhetarpal: Do you mean it in the sense that we can't use custom fonts with the WASM backend? I haven't really thought about why the file exists in the first place, though – I just updated it to the one that Matplotlib 3.8.4 had. If it doesn't help Pyodide, I would be okay with removing it.

packages/matplotlib/test_matplotlib.py
agriyakhetarpal: Agreed, should I remove it altogether?

hoodmane:
Looks good to me @agriyakhetarpal. Please feel free to merge when packages tests pass.

agriyakhetarpal:
Argh, it is LightGBM yet again. I haven't been able to figure out what other package messes with it when building, because it never causes problems for me locally. Should we isolate it into its own CI job like we have for OpenCV?

agriyakhetarpal:
Argh, it is LightGBM yet again. I haven't been able to figure out what other package messes with it when building (which is what happens, according to Gyeongjae), because it never causes problems for me locally. Should we isolate it into its own CI job like we have for OpenCV?

ryanking13:
> because it never causes problems for me locally. Should we isolate it into its own CI job like we have for OpenCV?

After 0.27.0 release, I'll proceed to unvendoring packages, so I guess we'll be able to split package build jobs.

packages/matplotlib-pyodide/meta.yaml
```diff
@@ -1,11 +1,11 @@
 package:
   name: matplotlib-pyodide
-  version: 0.2.2
+  version: 0.2.3
   top-level:
     - matplotlib_pyodide
 source:
-  url: https://files.pythonhosted.org/packages/36/8e/f28cdc02d2eb5d9a711876c39be72701b09ec80ce9ea3a6b67a7f08d8212/matplotlib_pyodide-0.2.2-py3-none-any.whl
-  sha256: 712a1bd8a06875393f25b0dc7a35cff1ecde2df36446d1290aaa846fd225a5df
+  url: https://files.pythonhosted.org/packages/1a/f9/b5f36d3221e8e211f366544ccc817c61e8f5c2bcc912fe1e1bf9a6044909/matplotlib_pyodide-0.2.3.tar.gz
```
ryanking13, line 7: Could you replace the url with the wheel file before merge?

ryanking13 (approved):
Thanks!

packages/matplotlib/test_matplotlib.py
ryanking13: Sorry, I introduced that bad test. Let's remove/update the tests altogether in the separate Pr.

hoodmane:
Thanks @agriyakhetarpal and @ryanking13!

packages/matplotlib-pyodide/meta.yaml
hoodmane: Oops didn't notice this. Guess we can do it in a followup...

ianthomas23:
> I would have liked to bump right up to Matplotlib 3.9.2, but I am not too keen on it – we could have a 0.27.1 release for that and a bunch of other package updates, perhaps...

There are changes in Matplotlib 3.9 and 3.10 affecting backends which might need patching here. I suggest that 3.8.4 is recent enough for the moment and we hold off doing any more updates until I've had a chance to look into the modified WebAgg backend idea, if that is OK.

agriyakhetarpal:
> > I would have liked to bump right up to Matplotlib 3.9.2, but I am not too keen on it – we could have a 0.27.1 release for that and a bunch of other package updates, perhaps...
> 
> There are changes in Matplotlib 3.9 and 3.10 affecting backends which might need patching here. I suggest that 3.8.4 is recent enough for the moment and we hold off doing any more updates until I've had a chance to look into the modified WebAgg backend idea, if that is OK.

Makes sense, thanks, @ianthomas23, for noting and thanks again for taking out the time to collaborate with us on this (and for helping review this PR, of course!).

PR #4512, opened 2024-02-11
DOC Update Github Actions example to latest versions

viblo:
### Description

This PR updates the github actions example to build a library in the docs to use latest version of the actions used. (I tested this on my own library, here: https://github.com/viblo/pymunk/actions/runs/7860843245/job/21448495731 )

### Checklists

### DIFF
--- a/docs/development/building-and-testing-packages.md
+++ b/docs/development/building-and-testing-packages.md
@@ -150,14 +150,14 @@ wheel out of tree:
 ```yaml
 runs-on: ubuntu-22.04
   steps:
-  - uses: actions/checkout@v3
-  - uses: actions/setup-python@v4
+  - uses: actions/checkout@v4
+  - uses: actions/setup-python@v5
     with:
        python-version: 3.11.2
   - run: |
       pip install pyodide-build>=0.23.0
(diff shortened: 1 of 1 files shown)


hoodmane:
Thanks!

hoodmane:
@ryanking13 it might be good to have a reminder somewhere to update this, perhaps as part of release notes...

ryanking13:
> @ryanking13 it might be good to have a reminder somewhere to update this, perhaps as part of release notes...

Yes, probably. But I think it's not really critical to keep the action version up-to-date unless the old actions are broken. So I think it's positive that new contributors sometimes find things like this and raise PRs to fix the documentation.

ryanking13 (approved):

hoodmane:
But for instance the Python version here is still 3.11.

viblo:
Ah, I didnt update the python version since I only tried with the stable release of pyodide and not the upcoming version, and didnt reflect on the version of documentation or that the new version supports 3.12..

hoodmane:
Reasonably enough. I think this is good because as-is I can cherry-pick this commit to update the existing docs.

ryanking13:
> But for instance the Python version here is still 3.11.

Oh yeah, the Python version needs to be updated...

PR #4515, opened 2024-02-12
Reduce the amount of stuff that the makefile always redoes

hoodmane:
libgl is marked .PHONY and is a dependency for the link step which forces us to always redo the link step even though it's completely not necessary. I moved `embuilder build libgl` into the link step since that's the only place where it's needed.

`dist/python_stdlib.zip` and `dist/pyodide-lock.json` both declare a dependency on `pyodide_build` but there is never a file of that name so they always have to run. Instead, run `make pyodide_build` so that they can reuse the logic but they don't always run if their other deps are satisfied.

Now that `dist/python_stdlib.zip` doesn't always run it needs to rely on all files in `src/py`.

Make `pyodide_build` only invoke `pip install` if `pyodide_build` is not already installed. Pip generates a lot of noise and takes a while so skipping it when we don't need it helps a lot.

Make `dist/webworker.js` and friends not phony since they stopped needing to be when `PYODIDE_BASE_URL` was removed from them.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -23,7 +23,7 @@ all: check \
 	dist/webworker.js \
 	dist/webworker_dev.js \
 	dist/module_webworker_dev.js
-	echo -e "\nSUCCESS!"
+	@echo -e "\nSUCCESS!"
 
 src/core/pyodide_pre.o: src/js/generated/_pyodide.out.js src/core/pre.js src/core/stack_switching/stack_switching.out.js
 # Our goal here is to inject src/js/generated/_pyodide.out.js into an archive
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -178,11 +180,13 @@ src/js/generated/pyproxy.ts : src/core/pyproxy.* src/core/*.h
 		sed 's/^#pragma clang.*//g' \
 		>> $@
 
-pyodide_build: ./pyodide-build/pyodide_build/**
-	$(HOSTPYTHON) -m pip install -e ./pyodide-build
-	which pyodide >/dev/null
+pyodide_build:
+	@echo "Ensuring editable pyodide-build is installed"
+	./tools/check_editable_pyodide_build.py || $(HOSTPYTHON) -m pip install -e ./pyodide-build
```
ryanking13, line 185: > ./tools/check_editable_pyodide_build.py

where does this file come from?

ryanking13 (approved):
Thanks!

> Make pyodide_build only invoke pip install if pyodide_build is not already installed. Pip generates a lot of noise and takes a while so skipping it when we don't need it helps a lot.

I love that I don't have to look at the pip logs every time anymore.

Makefile
hoodmane: Forgot to add it, thanks.

PR #4518, opened 2024-02-14
Make install_xbuildenv work even if user doesn't source the venv

hoodmane:
This was written assuming that the user sourced the virtual environment explicitly, but it fails if they do `.venv/bin/pyodide venv .venv-pyodide`. There's probably some other places with similar problems.

@henryiii does this look reasonable to you?

### DIFF
--- a/pyodide-build/pyodide_build/build_env.py
+++ b/pyodide-build/pyodide_build/build_env.py
@@ -6,15 +6,10 @@ import os
 import re
 import subprocess
 import sys
-from collections.abc import (
-    Iterator,
-    Mapping,
-)
+from collections.abc import Iterator
 from contextlib import nullcontext, redirect_stdout
 from io import StringIO
-from os import PathLike
 from pathlib import Path
-from typing import TypeAlias
 
 if sys.version_info < (3, 11, 0):  # noqa: UP036
     import tomli as tomllib
@@ -26,10 +21,6 @@ from packaging.tags import Tag, compatible_tags, cpython_tags
 from .common import exit_with_stdio
 from .logger import logger
 from .recipe import load_all_recipes
-
-StrOrBytesPath: TypeAlias = str | bytes | PathLike[str] | PathLike[bytes]
-_ENV: TypeAlias = Mapping[bytes, StrOrBytesPath] | Mapping[str, StrOrBytesPath]
-
 
 RUST_BUILD_PRELUDE = """
 rustup toolchain install ${RUST_TOOLCHAIN} && rustup default ${RUST_TOOLCHAIN}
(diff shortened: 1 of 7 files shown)


henryiii:
The standard way to detect a venv is to check for `pyvenv.cfg` at the base of the environment. If it exists, it's a venv (and the contents tell you various details about the venv).

hoodmane:
> The standard way to detect a venv is to check for pyvenv.cfg at the base of the environment.

Well the situation is that I want someone to be able to say `.venv/bin/some-pyodide-cmd` and have the implementation of `some-pyodide-cmd` invoke `.venv/bin/pip`. If the user ran `source .venv/bin/activate` then that placed `.venv/bin/` at the beginning of their path and so `pip` will resolve to `.venv/bin/pip`. Otherwise we need to add it.

I thought `sys.prefix != sys.base_prefix` was a good enough way to test for a virtual environment. But you're saying I should do something more like `(Path(sys.prefix) / "pyvenv.cfg").exists()`?

pyodide-build/pyodide_build/build_env.py
```diff
@@ -363,3 +372,27 @@ def check_emscripten_version() -> None:
         raise RuntimeError(
             f"Incorrect Emscripten version {installed_version}. Need Emscripten version {needed_version}"
         )
+
+
+def calculate_venv_environment(env: _ENV | None = None) -> dict[str, str]:
```
ryanking13, line 377: It would be nice to have some tests for this.

ryanking13 (approved):
Generally looks good to me.

henryiii:
Here's a recent example: https://github.com/astral-sh/uv/pull/1504/files.

henryiii:
Do you need to compute `bin/pip`? Couldn't you use `{sys.executable} -m pip`?

hoodmane:
I guess we could also do 
```py
{sys.executable} -m pyodide_cli ...
```
instead of `pyodide ...` and get rid of this environment calculation entirely.

hoodmane:
@henryiii I think I found the most to-the-point version of this that I can come up with:
```py
def _calculate_venv_path() -> str:
    """Make sure that the current executable is the first one on the path."""
    path = os.environ["PATH"]
    if shutil.which(Path(sys.executable).name) == sys.executable:
        return path
    # We aren't the first entry on the path. Add current executable directory to
    # the front of path to correct this.
    bin_dir = str(Path(sys.executable).parent)
    return f"{bin_dir}:{path}"
```
Though perhaps an even simpler option would be to unconditionally prepend `bin_dir` to the path, since it's not like doubling it up harms anyone much.

hoodmane:
Okay, just using `sys.executable -m some_module` everywhere is definitely the simplest thing we can do so I'll stick with that. This was surprisingly confusing...

hoodmane:
Hmm well the last iteration seems particularly broken...

PR #4520, opened 2024-02-14
MAINT Make search_pyodide_root return None instead of raising

hoodmane:
Every spot that we call it we have a try block to catch this exception, so returning None makes the code more succinct


ryanking13 (approved):
Yes, I think it makes sense. Thanks!

PR #4528, opened 2024-02-17
Remove cython and tomli from pyodide-build dependency

ryanking13:
Resolve #4527

Removes `cython` and `tomli` from pyodide-build dependency and updates requires-python to 3.12


packages/logbook/meta.yaml
```diff
@@ -12,11 +12,6 @@ requirements:
     - setuptools
   run:
     - ssl
-
-build:
-  script: |
-    python -c 'from Cython.Build.Cythonize import main; main(["./logbook/_speedups.pyx"])'
```
ryanking13, line 18: logbook now does this in their setup.py, so this line is not required anymore.

https://github.com/getlogbook/logbook/blob/1.7.0.post0/setup.py

hoodmane:
cc @mkoeppe

ryanking13:
Thanks for the review!

Issue #4401, opened 2024-01-21
Callbacks with `syncify()` raises an error


kor0p:
## 🐛 Bug

When I define callback (e.g. event listener) in Pyodide, I want to use `syncify()` there as well
But it seems, that it doesn't work

### To Reproduce

Open [browser with last version of Pyodide](https://pyodide.org/en/stable/console.html) and run the next code:
```python
await pyodide.runPythonSyncifying(`
    from pyodide.ffi.wrappers import add_event_listener
    import asyncio
    import js

    async def async_handler():
        await asyncio.sleep(0.1)
        print('Hi there')

    def handler(event):
        asyncio.create_task(async_handler()).syncify()
        print('Hi here')

    add_event_listener(js.document.body, 'click', handler)
`);
```
Then, click on the body (or any other element, you specified)

You will see the error in the console:
![image](https://github.com/pyodide/pyodide/assets/32600554/149a504d-ec1c-42cd-b148-25ce3aa0a8d2)
Also, you will see the text "Hi there", that is inside `async_handler()`, but you won't see the text "Hi here", that goes after `syncify()`
So, you can create async handlers (as you could earlier), but not with using `syncify()` anywhere

### Expected behavior

Expected to work this way, or some other way (there are no examples for it, or I didn't find them)

### Environment

- Pyodide Version: 0.25.0
- Browser version: Version 120.0.6099.199 (Official Build) (64-bit)
- Any other relevant information:
- - Enabled JSPI flag in Chrome: chrome://flags/#enable-experimental-webassembly-stack-switching

hoodmane:
Thanks for the bug report @kor0p. It would definitely be helpful to improve the situation here. With handlers, the main question is whether the callee of the handler will want a sync context or an async one. I don't think there is any reasonable way to introspect this. It would be a good idea to add support for something like this in `add_event_listener`. 

There are also more significant issues about the interaction between stack switching and and async contexts which will have to be ironed out. In particular, our event loop should ideally enter execution contexts with stack switching enabled if the runtime supports it. Since our event loop is calling into the function, we know the context is async. Though this could potentially lead to funny things like this not working:
```py
def handler(event):
    function_that_uses_syncify()
    print('Hi here')

addListener(handler)
```
but being fixed if we make it into `async def handler`.

For now a possible workaround is:
```py
from pyodide.ffi import create_proxy
import asyncio
import js

async def async_handler():
    await asyncio.sleep(0.1)
    print('Hi there')

[...]

Issue #4398, opened 2024-01-20
Using `asyncio.gather()` with `syncify()` raises error


kor0p:
## 🐛 Bug
Using `asyncio.gather()` with `syncify()` isn't possible out of the box

### To Reproduce
Open [browser with last version of Pyodide](https://pyodide.org/en/latest/console.html) and run the next code:
```python
await pyodide.runPythonSyncifying(`
    import asyncio

    async def test(x):
        await asyncio.sleep(0.1 * x)
        return x + 1

    res = asyncio.create_task(asyncio.gather(test(5), test(3), test(10))).syncify()
    print(*res)
`);
```
This code will raise the error:
![image](https://github.com/pyodide/pyodide/assets/32600554/54d3e8bb-23c2-4ada-a051-c858d74e5471)

The next workaround will work as expected:
```python
await pyodide.runPythonSyncifying(`
    import asyncio

    async def test(x):
        await asyncio.sleep(0.1 * x)
        return x + 1

    async def better_gather(*coros):
        return await asyncio.gather(*coros)

    res = asyncio.create_task(better_gather(test(5), test(3), test(10))).syncify()
    print(*res)
`);
```

### Expected behavior
Expected the first code snippet to work :)

### Environment

- Pyodide Version: 0.25.0
- Browser version: Version 120.0.6099.199 (Official Build) (64-bit)
- Any other relevant information:
- - Enabled JSPI flag in Chrome: chrome://flags/#enable-experimental-webassembly-stack-switching

PR #4532, opened 2024-02-18
Stack switching: make pystate properly save and restore asyncio state

hoodmane:
This updates `pystate.c` to save and restore the current event loop task. This allows us to stack switch out of one async task and then block for the completion of another async task.

We also add `callPyObjectMaybeSuspending` which uses `promisingApply` if stack switching is available and otherwise makes a normal call. We add a private keyword argument `_may_syncify` to `create_once_callable` and have the event loop use this so async tasks can stack switch. We also make the promise handles use `callPyObjectMaybeSuspending` so that `promise.then`, `promise.finally_`, etc can use stack switching.

This creates the funny situation described in #4401 where a handler can stack switch only if the entrypoint was an async function. I think I can live with this.
<details><summary>Example</summary>
<p>

```py
from js import document, sleep
from pyodide.ffi import create_proxy

def f(e, entry="f"):
   print(f"Does this work? {entry}")
   try:
      sleep(100).syncify()
   except:
      print(f"no {entry}")  
   else:
      print(f"yes {entry}")  

async def g(e):
  f(e, entry="g")

document.addEventListener("click", create_proxy(f))
document.addEventListener("click", create_proxy(g))
```
prints:
```
Does this work? f
no f
Does this work? g
yes g
```

</p>
</details> 

A few API considerations:

1. Should we eliminate `pyodide.runPythonSyncifying`? After this PR it's sufficient to use `pyodide.runPythonAsync`.
2. Should we eliminate `PyProxy.callSyncifying`? It suffices to make it an `async` function. If you have a non async function you can make an async wrapper around it (see example).
3. Should we eliminate `syncify` and just tell people to use `asyncio.run_until_complete` instead? `syncify` has problems like #4398 which seem hard to fix.

cc @rth @ryanking13 @kor0p @dom96 @garrettgu10

### Checklist

- [x] Add / update tests


hoodmane:
One problem with `asyncio.run_until_complete` is that it already does not block when JSPI is not available, so it can't raise an exception if blocking is impossible. We should probably add a new API for "block or raise an exception if you can't". But it probably should take the Future/Promise as an argument since adding a method to Futures and Promises, though tempting, has some limitations e.g., #4398. Though if we do want to keep a method, we could patch the base `Future` rather than making a subclass.

Another TODO is to add a good way to query whether or not `syncify()` can work. This involves annoying bookkeeping since currently JSPI provides no way to ask the runtime (https://github.com/WebAssembly/js-promise-integration/issues/29). I think we can add appropriate bookkeeping in our EM_JS wrappers.

ryanking13:
Thanks for keep improving JSPI related stuff!

> Should we eliminate syncify and just tell people to use asyncio.run_until_complete instead? syncify has problems like https://github.com/pyodide/pyodide/issues/4398 which seem hard to fix.

Yes, I am +1 for using `asyncio.run_until_complete` if possible, so we can abstract JSPI if possible.

hoodmane:
What about the fact that it might just do nothing if jspi is not available? Surely we should offer an alternative API that raises if it can't block? Or should we make run_until_complete do that?

ryanking13:
I think `asyncio.run_until_complete` should do that (raising an error, or at least showing an warning), which will make existing codebases more portable.

PR #4538, opened 2024-02-21
CI Disable webworker tests in gha

ryanking13:
GHA still hangs, and sadly, #4525 didn't work.

Looking at the log, the test always seems to hang at webworker tests. So maybe we need to investigate this in `pytest-pyodide` side. This PR disables webworker tests in GHA until we find a fix.


hoodmane:
Thanks!

Issue #3704, opened 2023-03-28
OpenAI pkg


frtvi:
## 🐍 Package Request

- Package Name and Version : openai
- Package URL :https://github.com/openai/openai-python
- Package Dependencies that needs to be resolved first:

ryanking13:
Closing as a duplicate: https://github.com/pyodide/pyodide/issues/3588, but please don't hesitate to comment there.

PR #4543, opened 2024-02-22
Add package httpx

hoodmane:
Like aiohttp, it requires a patch to function. Ideally we should add support upstream. Currently only async requests are supported but we can support sync ones too with jspi.

Split off from #3704.

### Checklists

- [x] Add / update tests


ryanking13:
We already added httpx #5302 . Closing

PR #4546, opened 2024-02-22
Fix leak in pyproxy iteration

hoodmane:
There is a good change there are similar leaks in async iterators and other generator methods but I haven't checked.

- [x] Add / update tests


ryanking13 (approved):
Thanks!

PR #4547, opened 2024-02-23
Remove pyodide.runPythonSyncifying

hoodmane:
It is now a worse version of `pyodide.runPythonAsync` since you can stack switch from
there but also you can use top level await.

### DIFF
--- a/src/tests/test_syncify.py
+++ b/src/tests/test_syncify.py
@@ -26,7 +26,7 @@ def test_syncify_not_supported1(selenium_standalone_noload):
         delete WebAssembly.Suspender;
         let pyodide = await loadPyodide({});
         await assertThrowsAsync(
-          async () => await pyodide.runPythonAsync("1+1"),
+          async () => await pyodide._api.pyodide_code.eval_code.callSyncifying("1+1"),
           "Error",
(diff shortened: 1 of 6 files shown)


src/tests/test_syncify.py
```diff
@@ -184,36 +186,34 @@ def wrapper():
...
-            test = run_js(
-                '''
-                (async function test() {
-                    await sleep(1000);
-                    return 7;
-                })
-                '''
-            )
+@run_in_pyodide
+def test_syncify_ctypes(selenium):
```
ryanking13, line 193: It looks like this change is making the playwright hang. I'm not sure why exactly, but it's worth investigating (maybe it's stuck on getting the result of the run_in_pyodide?)

Would you mind skipping the test (instead of xfail) until we find the reason?

ryanking13 (approved):
Thanks!

src/tests/test_syncify.py
hoodmane: > Would you mind skipping the test (instead of xfail) 

Not at all. For what it's worth, I was also having issues with it hanging when running locally. I wasn't really aware that xfail and skip were different to be honest.

PR #4548, opened 2024-02-23
JSPI replace syncify() method with run_sync function

### LINKED ISSUE #4398: Using `asyncio.gather()` with `syncify()` raises error

hoodmane:
This resolves #4398 which happened because not all Futures get the syncify 
method properly attached. I think the name `block_for_awaitable` is a bit more 
explicit about what it does, but I am open to syncify or other names.


ryanking13 (approved):
Thanks @hoodmane! Codewise LGTM. I think this needs a changelog.

About the naming, I agree that `block_for_awaitable` is more straightforward than `syncify`. So I am okay with the current naming.

rth:
Thanks for doing this work! Some thoughts on the naming, following the ping, as someone who hasn't followed earlier discussions in detail .

1. `syncify`: I generally like the name. You have something async, make it sync. Very simple. Recently saw something similar in fsspec in another context, where they [called it `sync_wrapper`](https://filesystem-spec.readthedocs.io/en/latest/async.html?highlight=sync_wrapper#synchronous-api)
2. `block_for_awaitable`: I get what it does, and it might indeed be more specific but
   - it raises the question what's an awaitable.  I think the average Python user, me included, may be a bit fuzzy as to what's the technical difference between awaitable, coroutine, task, a future. And so now I have to make sure the argument I pass it is an awaitable. Yes, it's defined in https://docs.python.org/3/library/asyncio-task.html#awaitables but people will probably have to search for it.
   - block: raises the question if there is any other way to run an awaitable  from a non async function. If not then is there a need to specify it, all other sync functions also block.
3. In other choices, maybe `run_sync` (to make the link with `asyncio.run`). Tornado also apparently [has a `IOLoop.run_sync`](https://www.tornadoweb.org/en/stable/ioloop.html#tornado.ioloop.IOLoop.run_sync) for a somewhat similar operation, by I claim the precedence for coming up with this name haha.

Overall in terms of name preference I'm happy with `syncify`, I like `run_sync` but maybe I'm biased. Not very excited about `block_for_awaitable` but in the end it's just a name :)

hoodmane:
I personally think `syncify` has a lot more character than `block_for_awaitable`. I guess let's keep the name then.

rth:
Slight preference toward `run_sync` as simpler to write. For non native speakers, the `ify` termination may not be that obvious but OK with both.

hoodmane:
Ok, let's go with `run_sync`.

hoodmane:
Thanks @rth and @ryanking13!

PR #4550, opened 2024-02-23
Replace `pkgconfig` with `pkg-config` for Meson configuration file

### LINKED ISSUE #4540: Meson deprecation warning in `pyodide-build`: `pkgconfig` entry is deprecated and should be replaced by `pkg-config`
## 🐛 Bug

While trying to build and test the PyWavelets package in CI, I'm currently experiencing this Deprecation warning from the Meson build system: https://github.com/agriyakhetarpal/pywt/actions/runs/8003758318/job/21859733890#step:7:21

### To reproduce

It is difficult to produce a very _minimal_ MWE, but the workflow file for the above run is available. This seems to be coming from here:

https://github.com/pyodide/pyodide/blob/1bbebf32929ee73684b084797ec600c35a42165e/pyodide-build/pyodide_build/tools/emscripten.meson.cross#L1-L4

### Expected behavior

Should be `pkg-config = 'pkg-config'`, instead of the current `pkgconfig = 'pkg-config'`

### Environment

- `pyodide-build` version: 0.25.0
- `meson-python` version: 0.15.0
- Browser version: N/A

### Additional context

I can try to fix this in a small PR, would appreciate pointers on how to verify if the change actually removes the warning or if I'm looking at the wrong place!

agriyakhetarpal:
### Description

This PR is a small fix to silence a Meson deprecation warning, where the `pkgconfig` entry in the Meson cross-configuration file has to be replaced by the term `pkg-config`. This is referenced in the following lines:

https://github.com/mesonbuild/meson/blob/39e1bf19fb653ac4f040e48b9a33aaa59dea1196/mesonbuild/envconfig.py#L399-L404

Closes #4540 

### Checklists

- [ ] Add / update tests

### DIFF
--- a/pyodide-build/pyodide_build/tools/emscripten.meson.cross
+++ b/pyodide-build/pyodide_build/tools/emscripten.meson.cross
@@ -1,7 +1,7 @@
 # compiler paths are omitted intentionally so we can override the compiler using environment variables
 [binaries]
 exe_wrapper = 'node'
-pkgconfig = 'pkg-config'
+pkg-config = 'pkg-config'
 
 [properties]
 needs_exe_wrapper = true


agriyakhetarpal:
Hi @hoodmane, I think this should work – please let me know if this requires a CHANGELOG entry or if this is too minor for doing so. I don't see anything about this in the documentation either, but it is possible I may have missed something.

hoodmane (approved):
Perfect thanks!

PR #4553, opened 2024-02-23
Bump protobuf version to 4.24.4

bartbroere:
This PR bumps protobuf to ~4.25.3~ 4.24.4.
The 4.25.x release needs a bit more work.

#4558 contains more extensive testing for this package.


hoodmane (approved):
Thanks @bartbroere. Nice to increase test coverage a bit.

hoodmane:
By the way, I have been thinking about what we can do to improve the ergonomics of the test system. Let me know if you have opinions.

ryanking13:
It looks like there is a signature mismatch. Could you take a look if possible? Probably this document could help you: https://pyodide.org/en/stable/development/debugging.html

bartbroere:
> It looks like there is a signature mismatch. Could you take a look if possible? Probably this document could help you: https://pyodide.org/en/stable/development/debugging.html

Thanks for that resource. I'll be looking into this.

packages/protobuf/patches/make-implicit-pointer-explicit.patch
hoodmane, line 1: I'm a bit confused about how this compiles successfully both before and after this patch. Is `upb_MessageDef_MiniTable` a macro that casts the argument?

hoodmane (approved):
Patch is confusing but it does make the added tests pass. Thanks @bartbroere. Do you understand why:
1. it compiles successfully without the patch
2. it works correctly in native Python (I assume) but not in Pyodide?

I'd appreciate any explanation you can give.

packages/protobuf/patches/make-implicit-pointer-explicit.patch
bartbroere: On my end and in the CI it didn't build before this patch. (Note that I backed off a bit and first bumped to 4.24.x before even starting on 4.25.x support).

https://app.circleci.com/pipelines/github/pyodide/pyodide/6887/workflows/01b9a7e4-a06d-4b17-a990-bd8bd57c28e2/jobs/93955/parallel-runs/0/steps/0-104?invite=true#step-104-32873_80

Before the patch I got this error on the three lines addressed in the patch: `error: incompatible integer to pointer conversion`. This led me to think that the argument was being converted to a pointer implicitly, and decided to make it explicit.

I'll add a test to see if I broke the deepcopy functionality of protobuf messages by just turning the argument into a pointer.

bartbroere:
> Patch is confusing but it does make the added tests pass. Thanks @bartbroere. Do you understand why:
> 
> 1. it compiles successfully without the patch
> 2. it works correctly in native Python (I assume) but not in Pyodide?
> 
> I'd appreciate any explanation you can give.

My initial guess was that there's a difference between how typical C compilers and the Emscripten C compile handle implicit conversion to a pointer, but your theory that a macro might be responsible for this could also be true. I'll look into both theories.

Just to be sure I also added a test to see if this breaks something in a segmentation fault kind of way.

packages/protobuf/patches/make-implicit-pointer-explicit.patch
bartbroere: Thanks, that seems to do the trick. The newly added tests indeed showed the wrong memory being accessed with the patch.

hoodmane:
+1 for turn off flag and add a test that uses these functions.

hoodmane:
@bartbroere should I merge this?

bartbroere:
> @bartbroere should I merge this?

Yes, please. And thanks for the help!

hoodmane:
Thanks @bartbroere!

ryanking13:
Thanks!

PR #4554, opened 2024-02-23
Fix dup in nodefs

### LINKED ISSUE #4541: `OSError: [Errno 9] Bad file descriptor` when trying to load `.npy` files, works with `.npz` file format
## 🐛 Bug

NumPy is unable to load arrays from `.npy` binaries, but it can read from compressed `.npz` archives.

### To Reproduce

I noticed this error when compiling PyWavelets (`pywt`) from source via the Emscripten toolchain.

In an activated virtual environment created by Pyodide, run the following:

```bash
git clone https://github.com/PyWavelets/pywt.git
pip install .
```

and then

```python
import pywt
import numpy as np

aero = pywt.data.aero()

ref = np.array([[178, 178, 179],
                    [170, 173, 171],
                    [185, 174, 171]])

np.testing.assert_allclose(aero[:3, :3], ref)
```

should fail. However, [after converting](https://github.com/PyWavelets/pywt/pull/701/files#diff-86b5b5c7cbe8cc8368f6991c914b7263019507351ce567543cbf2b627b91aa57) these `.npy` files to `.npz`, NumPy can safely load the arrays from the files as requested.

Here is an example of conversion from `.npy` to `.npz`:

```python
import numpy as np
ecg = np.load("pywt/data/ecg.npy")
np.savez(file="ecg.npz", data=ecg)
```

after which `ecg.npz` can be loaded as follows:

```python
import numpy as np
loaded = np.load("ecg.npz")
print(loaded["data"])
```

### Expected behavior

The Pyodide environment should be able to load the `.npy` file format stored in a directory, but [fails with multiple `OSError`s](https://github.com/agriyakhetarpal/pywt/actions/runs/7993252511/job/21828629911), possibly due to the lack of a server for filesystem access as the Pyodide documentation mentions – but this doesn't explain why `.npz` files work?

The expected behaviour should be that all file formats work.

### Environment

- Pyodide Version: `pyodide-build` version 0.25.0
- Browser version:  N/A

### Additional context

xref: https://github.com/PyWavelets/pywt/pull/701

hoodmane:
Backport of emscripten-core/emscripten#21399. Resolves #4541

- [x] Add / update tests

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -54,6 +54,9 @@ pytest_pyodide.runner.INITIALIZE_SCRIPT = """
 only_node = pytest.mark.xfail_browsers(
     chrome="node only", firefox="node only", safari="node only"
 )
+only_chrome = pytest.mark.xfail_browsers(
+    node="chrome only", firefox="chrome only", safari="chrome only"
+)
 
 
 def pytest_addoption(parser):
--- a/docs/project/changelog.md
+++ b/docs/project/changelog.md
@@ -45,6 +45,10 @@ myst:
 
 - {{ Enhancement }} The `build/post` script now runs under the directory
   where the built wheel is unpacked.
+  {pr}`4481`
+
+- {{ Fix }} `dup` now works correctly in the Node filesystem.
+  {pr}`4554`
 
 ### Packages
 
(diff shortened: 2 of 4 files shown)


ryanking13 (approved):
Very nice. Thanks for fixing this!

hoodmane:
I'll add a test before merging.

PR #4558, opened 2024-02-24
Extend protobuf package testing

bartbroere:
When bumping the protobuf package, it turned out that it wasn't trivial.

I therefore decided to split adding the tests and bumping the version in two PR's. That way these tests could be merged first, and act as a nice regression test.


hoodmane:
Thanks!

PR #4565, opened 2024-02-25
Cleanup: Remove importlib.invalidate_caches calls

hoodmane:
It hasn't been needed since we brought in the patch that fixed the time functions on Emscripten:
https://github.com/emscripten-core/emscripten/pull/18301

### DIFF
--- a/src/tests/test_filesystem.py
+++ b/src/tests/test_filesystem.py
@@ -36,8 +36,6 @@ def test_idbfs_persist_code(selenium_standalone):
             p = pathlib.Path('{mount_dir}/test_idbfs/__init__.py')
             p.parent.mkdir(exist_ok=True, parents=True)
             p.write_text("def test(): return 7")
-            from importlib import invalidate_caches
-            invalidate_caches()
             import sys
             sys.path.append('{mount_dir}')
             from test_idbfs import test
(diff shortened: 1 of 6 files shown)


src/tests/test_filesystem.py
```diff
@@ -65,9 +63,7 @@ def test_idbfs_persist_code(selenium_standalone):
     selenium.run_js(
         f"""
         pyodide.runPython(`
-            from importlib import invalidate_caches
             import sys
-            invalidate_caches()
```
hoodmane, line 70: Hmm test fails without `invalidate_caches`. I suspect this is due to a bug in the IDBFS...

ryanking13:
IIRC, there was a browser timer precision issue about this (https://github.com/pyodide/pyodide/issues/3311). But I guess it is worth to experiment again.

hoodmane:
Well we fixed that with #3313. I think it would be good to remove this because if there are still time resolution problems, we would find out about them faster. Otherwise it's a bit confusing when users directly writing files can't import them but our package resolution system won't ever see problems like that.

ryanking13:
But according to Roman's comment here: https://github.com/pyodide/pyodide/issues/3311#issuecomment-1335546808, if the browser's timer precision is the problem, then emscripten-side fix wouldn't work anyway as emscripten relies on browser APIs (`Date.now()`).

ryanking13:
> I just don't see that it's very likely to be a problem.

Yeah, actually, I was just thinking it may be a browser issue, not that I think it's a real cause. In any case, if the import works well without invalidate_caches(), this is definitely a big improvement, so thanks for working on this.

PR #4567, opened 2024-02-25
MAINT Update error handling code to use new Python 3.12 APIs

hoodmane:
The only behavior change here should be that we are setting `sys.last_exc` (added in Python3.12). Since we haven't released with Python 3.12 yet, this doesn't need a changelog.

Python 3.12 added a bunch of modern APIs that handle a single exception object rather than the (type, val, tb) triples. They also 
repaired various quirks of the old error handling APIs. They are much easier to work with. Also, now that we have `capture_stderr` and `restore_stderr` we can use `PyErr_Print()` to format the traceback. This allows us to cut out a fair amount of code in error_handling.c.

This now uses `sys.excepthook` to format exceptions. We set `sys.excepthook` to `traceback.print_exception` because the default excepthook does not respect the linecache which we use to implement `pyodide.runPython("...", {file: "some_file.py"});`

### DIFF
--- a/src/py/_pyodide/__init__.py
+++ b/src/py/_pyodide/__init__.py
@@ -10,3 +10,21 @@
 from . import _base, _importhook
 
 __all__ = ["_base", "_importhook"]
+
+
+def set_excepthook():
+    import sys
+    import traceback
+
+    # We call sys.excepthook via PyErr_Print() in wrap_exception().
+    # traceback.print_exception in most ways behaves the same as the default
+    # sys.excepthook **except** traceback.print_exception uses the linecache
+    # whereas sys.excepthook only uses the file system. If the user calls
+    # `runPython` with a `file` argument, then we put this into the linecache
+    # but not into the filesystem. This means that the default excepthook won't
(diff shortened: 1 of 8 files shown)


ryanking13 (approved):
Thanks! Please update the failing tests about the formatting otherwise looks good to me.

src/py/_pyodide/__init__.py
```diff
@@ -10,3 +10,21 @@
...
+    # sys.excepthook **except** traceback.print_exception uses the linecache
+    # whereas sys.excepthook only uses the file system. If the user calls
+    # `runPython` with a `file` argument, then we put this into the linecache
+    # but not into the filesystem. This means that the default excepthook won't
+    # print the lines, but `traceback.print_exception` will. I think this is the
+    # only difference in their behavior.
+    #
+    # Python 3.13 seems to have switched to using `traceback.print_exception` as
+    # the default excepthook.
+    sys.excepthook = traceback.print_exception
```
ryanking13, line 30: Interesting. Thanks for the investigation!

PR #4568, opened 2024-02-26
Make `asyncio.sleep(0)` run faster

### LINKED ISSUE #4006: Better performing event loop
## Proposed refactoring or deprecation

The current implementation in [webloop.py](https://github.com/pyodide/pyodide/blob/3caa249177c09624b4e2039d4a68cfcf2fb014c4/src/py/pyodide/webloop.py#L342) relies on javascript's setTimeout function. While this implementation works, it introduces an issue for delays < 4ms, because per standard in all browsers `setTimeout` supports a minimum delay value of 4ms.    
On occasions developers introduce `await asyncio.sleep(0)` in their code to relinquish execution for a brief period in time in parts of asynchronous functions which would be called repetitively thus hogging the event loop.  
Since pyodide uses setTimeout, each call to `await asyncio.sleep(0)` will introduce a 4ms delay which in turn makes the code appear slow.

### Motivation
Aardwolf RDP client implementation is running slower than expected on pyodide because of this issue.

### Pitch
I'd like to have a replacement implemented in pyodide's `webloop.py` for `setTimeout` which supports faster callbacks.

### Additional context

I did some tests which did speed up my code by replacing calls for setTimeout with a MessageChannel-based implementation if `delay==0`. It worked on the live tests, here are the results:

- default behaviour
```
import time
import asyncio


>>> async def measure_execution_time():
...  num_measurements = 1000
...  total_time = 0
...  for _ in range(num_measurements):
...   start_time = time.perf_counter()
...   await asyncio.sleep(0)
...   end_time = time.perf_counter()
...   total_time += (end_time - start_time)
...  average_time = total_time / num_measurements
...  print(f"On average, the function took {average_time} seconds to execute")
... 

>>> await measure_execution_time()
On average, the function took 0.00456210000899992 seconds to execute
>>> await measure_execution_time()
On average, the function took 0.004544299998000042 seconds to execute
```
- modifying `webloop.py` to use `zeroTimeout`:
```
function zeroTimeout(callback, delay) {
    if(delay == 0){
        const channel = new MessageChannel();
        channel.port1.onmessage = callback;
        channel.port2.postMessage('');
    }
    else{
        setTimeout(callback, delay);
    }  
}
```
```
Python 3.11.2 (main, Jul 21 2023 11:59:42) on WebAssembly/Emscripten
>>> import asyncio
>>> import time
>>> async def measure_execution_time():
...  num_measurements = 1000
...  total_time = 0
...  for _ in range(num_measurements):
...   start_time = time.perf_counter()
...   await asyncio.sleep(0)
...   end_time = time.perf_counter()
...   total_time += (end_time - start_time)
...  average_time = total_time / num_measurements
...  print(f"On average, the function took {average_time} seconds to execute")
... 
>>> await measure_execution_time()
On average, the function took 0.00016000000100000023 seconds to execute
>>> await measure_execution_time()
On average, the function took 0.00016209999599999493 seconds to execute
>>> await measure_execution_time()
On average, the function took 0.00014569999600001893 seconds to execute
```
The `webloop.py` script was modified in the following way:
```
if IN_BROWSER:
    from js import setTimeout as originalSetTimeout

    try:
        from js import zeroTimeout as setTimeout
    except ImportError:
        setTimeout = originalSetTimeout
        print('Failed to import zeroTimeout')
```
```
 def run_handle(*args, **kwargs):
```
The latter modification was needed because the `MessageChannel` version must dispatch some data to `run_handle`.  
  
*WARNING*   
I have no idea if this messes up something on the long run!!!!

ryanking13:
### Description

Resolve #4006

This PR replaces `setTimeout` in webloop with a wrapper function `scheduleCallback` which supports zero timeout, as suggested in #4006.

When tested locally with chrome by running `asyncio.sleep(0)` 1000 times. I saw a speedup from 4.3s to about 0.2s.

### Checklists

- [x] Add / update tests


ryanking13:
> Thanks for picking up this backlog

I encountered this issue when working on pygame (which uses asyncio.sleep a lot) :)

src/js/api.ts
```diff
@@ -80,6 +81,8 @@ export class PyodideAPI {
   static setStdout = setStdout;
   /** @hidden */
   static setStderr = setStderr;
+  /** @hidden */
+  static scheduleCallback = scheduleCallback;
```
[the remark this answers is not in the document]
ryanking13: Thanks. I forgot that we can access it under `_api` namespace.

src/js/api.ts
```diff
@@ -51,6 +52,10 @@ API.saveState = () => API.pyodide_py._state.save_state();
 /** @private */
 API.restoreState = (state: any) => API.pyodide_py._state.restore_state(state);
 
+// Used in webloop
+/** @private */
+API.scheduleCallback = scheduleCallback;
```
hoodmane, line 57: You can just attach it to `API` in `scheduler.ts`.

ryanking13: It didn't work because then `scheduler.ts` is not imported from anywhere. Let me merge this as-is and the revise the code structure later. Thanks for the suggestion!

src/js/scheduler.ts
```diff
@@ -0,0 +1,14 @@
+/**
+ * Schedule a callback. Supports both immediate and delayed callbacks.
+ * @param callback The callback to be scheduled
+ * @param timeout The delay in milliseconds before the callback is called
+ */
+export function scheduleCallback(callback: () => void, timeout: number = 0) {
+  if (timeout <= 0) {
+    const channel = new MessageChannel();
```
e-nikolov, line 13: I was also working on a faster loop for my own project and at first I was also creating a new MessageChannel for each call, but that ran into garbage collection issues on both Chrome and Firefox when I had too many concurrent coroutines. The garbage collector could not keep up cleaning up the channels and the page kept crashing after the memory usage reached ~20GB RAM.

This is my current implementation, but I still need to clean it up a bit: https://github.com/e-nikolov/mpyc-web/blob/master/mpyc-web-py/lib/weblooperV10.py

It uses a single channel to avoid the memory issues.

ryanking13: I see. Thanks for letting me know and sharing your implementation!

When I did some simple checks in the browser devtools, it seemed like the MessageChannel and MessagePorts are garbage collected well. But I'll keep investigating.

ryanking13:
It seems like this breaks safari (and deno I guess). It is slow in safari, and breaks scheduling for some reason. Let me revert this and find if I can find another way to do this.

```
FAILED test_pyproxy.py::automatic_coroutine_scheduling[safari] - assert [3] == [3, 4, 6]
  Right contains 2 more items, first extra item: 4
  Full diff:
  - [3, 4, 6]
  + [3]
FAILED test_webloop.py::zero_timeout[safari] - AssertionError: elapsed: 16.26799988746643s
assert 16.26799988746643 < 4
```

PR #4574, opened 2024-02-28
Don't throw when calling str on a proxy without a toString method

hoodmane:
Resolves #4569. It still doesn't make sure `str(jsproxy)` never throws. It will throw if:
1. accessing `obj.toString` succeeds and returns a function, but calling the function throws
2. accessing `obj.toString` fails or returns not a function, and `Object.prototype.toString.call` fails.

See added tests. Sound reasonable @ryanking13? cc @CNSeniorious000.

- [x] Add / update tests

### DIFF
--- a/docs/project/changelog.md
+++ b/docs/project/changelog.md
@@ -16,6 +16,13 @@ myst:
 
 ## Unreleased
 
+- {{ Enhancement }} `str(jsproxy)` has been adjusted to not raise an error if
+  `jsproxy.toString` is undefined. Instead, it will use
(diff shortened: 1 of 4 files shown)


ryanking13 (approved):
Thanks for the quick fix @hoodmane! It seems like toString() results slightly differs by browsers, otherwise LGTM.

> accessing obj.toString succeeds and returns a function, but calling the function throws
accessing obj.toString fails or returns not a function, and Object.prototype.toString.call fails.

I think this is reasonable enough.

Issue #4579, opened 2024-02-28
Package request: Altair


joelostblom:
## 🐍 Package Request

- Package Name and Version: altair / latest (5.2)
- Package URL: https://pypi.org/project/altair/
- Package Dependencies that needs to be resolved first: None, [all the required dependencies](https://github.com/altair-viz/altair/blob/e58e8cbac613bf7babbd51d1798ec5da4b7d4fbb/pyproject.toml#L17-L27) are already listed at https://pyodide.org/en/stable/usage/packages-in-pyodide.html

Altair is a package for creating interactive data visualizations via a declarative syntax. Although the package is pure python and installable via micropip, it would be convenient to have it included in the base pyodide image (the wheel size is about 1MB). I'm a co-maintainer and happy to support with anything that is needed from the altair side of things to add it to the base pyodide image.

hoodmane:
PR welcome!

joelostblom:
Thanks for being open to a PR @hoodmane! I had a go using [the docs instructions](https://pyodide.org/en/stable/development/new-packages.html) and the package seems to build fine (I used the docker container), but I'm getting stuck at the step where I'm asked to open `localhost:<port>/console.html`. It seems like that file does not exist within the `dist/` directory. I tried serving the dist directory both from within docker and outside, and also to copy over the file manually. When I copy the file manually I get stuck at the spinning wheel that loads the terminal, and the browser console (both firefox and chromium) show the following error:

```
Uncaught (in promise) TypeError: The specifier “{{ PYODIDE_BASE_URL }}pyodide.mjs” was a bare specifier, but was not remapped to anything. Relative module specifiers must start with “./”, “../” or “/”.
```

Any idea of what is going on here and how I can run the console locally to test if the package built properly?

---

I can load https://pyodide.org/en/stable/console.html fine and it seems like the package built without issues:

![image](https://github.com/pyodide/pyodide/assets/4560057/8e939fb0-3852-4421-a2ed-cfadbd9401a6)

hoodmane:
> It seems like that file does not exist

Try `make dist/console.html`

joelostblom:
Thanks! It seems like I had to run `make all` in the end; I couldn't get it to work with just `all-but-packages` even when building `dist/console.html` separately. I added a note about this in the docs in https://github.com/pyodide/pyodide/pull/4581 and a PR for the altair package in https://github.com/pyodide/pyodide/pull/4580

PR #4581, opened 2024-02-29
Note that `make all` is needed to test in the console

joelostblom:
### Description

As mentioned in https://github.com/pyodide/pyodide/issues/4579#issuecomment-1970183159, but we also need to build the `dist/console.html` rule which is under `all`.

### Checklists

### DIFF
--- a/docs/development/new-packages.md
+++ b/docs/development/new-packages.md
@@ -150,12 +150,13 @@ and see if there are any errors.
 
 If the build succeeds you can try to load the package:
 
-1. Serve the dist directory with `python -m http.server --directory ./dist`.
+1. Build pyodide and all associated packages via `make all`.
+2. Serve the dist directory with `python -m http.server --directory ./dist`.
    If you use docker, you can execute this either outside of the docker container or
    make sure to forward a port by setting the environment variable
    PYODIDE_SYSTEM_PORT or starting docker with `./run_docker -p <port>`.
-2. Open `localhost:<port>/console.html` and try to import the package.
-3. You can test the package in the repl.
+3. Open `localhost:<port>/console.html` and try to import the package.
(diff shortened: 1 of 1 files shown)


docs/development/new-packages.md
```diff
@@ -150,12 +150,13 @@ and see if there are any errors.
 
 If the build succeeds you can try to load the package:
 
-1. Serve the dist directory with `python -m http.server --directory ./dist`.
+1. Build pyodide and all associated packages via `make all`.
```
hoodmane, line 153: `make all` is the same as just `make`:
```suggestion
1. Build pyodide and all associated packages via `make`.
```

docs/development/new-packages.md
```diff
@@ -150,12 +150,13 @@ and see if there are any errors.
...
+2. Serve the dist directory with `python -m http.server --directory ./dist`.
    If you use docker, you can execute this either outside of the docker container or
    make sure to forward a port by setting the environment variable
    PYODIDE_SYSTEM_PORT or starting docker with `./run_docker -p <port>`.
-2. Open `localhost:<port>/console.html` and try to import the package.
-3. You can test the package in the repl.
+3. Open `localhost:<port>/console.html` and try to import the package.
```
hoodmane, line 158: If you used `python -m http.server` it will have selected port 8000, so maybe we should just put that explicitly here? Not sure.

hoodmane (approved):
Thanks for helping with the docs @joelostblom.

docs/development/new-packages.md
joelostblom: Yes, I think that makes sense. Anyone who intentionally modifies the port would know to change it here too

```suggestion
3. Open `localhost:8000/console.html` and try to import the package.
```

Issue #4578, opened 2024-02-28
Package request: CVXPY + Clarabel


phschiele:
## 🐍 Package Request

- Package Name and Version: cvxpy/latest, clarabel/latest
- Package URL: https://github.com/cvxpy/cvxpy, https://github.com/oxfordcontrol/Clarabel.rs
- Package Dependencies that needs to be resolved first: All dependencies already available (NumPy, SciPy)

The CVXPY team would love to bring convex optimization to Pyodide (a long standing [user request](https://github.com/pyodide/pyodide/issues/3083))!
This would require bringing over the modeling language (CVXPY) as well as at least one solver (Clarabel).
Below is a short description of what would be required for both.

CVXPY:
- We've recently made it possible to run CVXPY without the custom C++ backend (https://github.com/cvxpy/cvxpy/pull/2363)
- Building pure Python wheels still requires a few lines of changes, which we would maintain in a diff here.

Clarabel:
- Clarabel is written in Rust with a PyO3 interface. The below Dockerfile was used to create a working test wheel
- For one problem type Clarabel relies on BLAS, and the currently generated wheel does not seem to work yet, so this would be the one open issue at the moment

We'd be happy to create a PR to get this started!

```dockerfile
FROM python:3.11.3
ENV EMSDK=/opt/emsdk EMSDK_NODE=/opt/emsdk/node/16.20.0_64bit/bin/node \
    PATH=/opt/emsdk:/opt/emsdk/upstream/emscripten:/opt/emsdk/node/16.20.0_64bit/bin:$PATH
RUN git clone https://github.com/emscripten-core/emsdk.git $EMSDK &&\
    emsdk install 3.1.45 && emsdk activate 3.1.45

# Install rust
RUN curl https://sh.rustup.rs -sSf | sh -s -- --default-toolchain nightly -y 

ENV PATH="/root/.cargo/bin:${PATH}"
RUN rustup target add wasm32-unknown-emscripten

# Install pipx
RUN python -m pip install pipx

# Install maturin
RUN pipx install maturin

# Add maturin to PATH
ENV PATH="/root/.local/bin:${PATH}"

# Build Clarabel.rs for wasm
RUN git clone https://github.com/oxfordcontrol/Clarabel.rs &&\
    cd Clarabel.rs &&\
    maturin build --release --target wasm32-unknown-emscripten
```

hoodmane:
> We'd be happy to create a PR to get this started!

PR welcome, thanks @phschiele.

[...]

PR #4587, opened 2024-03-01
Add packages cvxpy, clarabel

phschiele:
### Description
Adds `cvxpy` and `clarabel` as discussed in #4578 

### Checklists
- [x] Add / update tests


packages/cvxpy-base/meta.yaml
```diff
@@ -0,0 +1,26 @@
+package:
+  name: cvxpy-base
```
ryanking13, line 2: Question: What is the difference between cvxpy and cvxpy-base?

packages/cvxpy-base/patches/0001-Do-not-build-C-extensions.patch
```diff
@@ -0,0 +1,47 @@
...
+-import setup.extensions as setup_extensions
+ import setup.versioning as setup_versioning
+ 
+ # BEFORE importing setuptools, remove MANIFEST. Otherwise it may not be
+@@ -56,7 +55,7 @@ VERSION = setup_versioning.VERSION
+ setup(
+     version=str(VERSION),
+     cmdclass={'build_ext': build_ext_cvxpy},
+-    ext_modules=[setup_extensions.cvxcore, setup_extensions.sparsecholesky],
++    ext_modules=[],
```
ryanking13, line 41: Rather than keeping this patch in the Pyodide repository, it would be nice if you could add it to the cvxpy repository itself.

You may use `"PYODIDE" in os.environ` to detect if you are in Pyodide or building for Pyodide.

ryanking13 (commented):
Thanks @phschiele! I have some minor comments about building and testing these packages.

packages/cvxpy-base/meta.yaml
PTNobel: Mostly the dependencies. CVXPY requires the set of solvers necessary to be able to solve problems out of the box, e.g. ECOS, OSQP, SCS and Clarabel. cvxpy-base doesn't require any of the solvers, but still has the interfaces needed to communicate with any solver that's installed.

In other words, if you see CVXPY as a compiler from user written models to solver models, the CVXPY package depends on the solvers so you can compile and solve the model. CVXPY-base only has the compiler.

docs/project/changelog.md
```diff
@@ -68,7 +68,8 @@ myst:
 - New Packages: `cysignals`, `ppl`, `pplpy` {pr}`4407`, `flint`, `python-flint` {pr}`4410`,
   `memory_allocator` {pr}`4393`, `primesieve`, `primecount`, `primecountpy` {pr}`4477`,
   `pyxirr` {pr}`4513`, `ipython`, `asttokens`, `executing`, `prompt_toolkit`,
-  `pure_eval`, `stack_data`, `traitlets`, `wcwidth` {pr}`4452`, `altair` {pr}`4580`
+  `pure_eval`, `stack_data`, `traitlets`, `wcwidth` {pr}`4452`, `altair` {pr}`4580`,
+  `cvxpy` {pr}`#4587`, `clarabel` {pr}`4587`
```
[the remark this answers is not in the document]
phschiele: Fixed in [7759257](https://github.com/pyodide/pyodide/pull/4587/commits/7759257def84979eb76957f7acba7b74f0f1ba5f)

packages/clarabel/meta.yaml
```diff
@@ -0,0 +1,20 @@
...
+  sha256: b9da56f522806a847f3ece08f5f21fab7506eec3148b22446190de5206dd1e6f
+about:
+  home: https://github.com/oxfordcontrol/Clarabel.rs
+  PyPI: https://pypi.org/project/clarabel
+  summary: Clarabel Conic Interior Point Solver for Rust / Python
+  license: Apache-2.0
+build:
+  script: |
+    rustup default nightly
+    rustup target add wasm32-unknown-emscripten
```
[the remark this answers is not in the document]
phschiele: Fixed in [7759257](https://github.com/pyodide/pyodide/pull/4587/commits/7759257def84979eb76957f7acba7b74f0f1ba5f)

packages/clarabel/test_clarabel.py
```diff
@@ -0,0 +1,35 @@
+from pytest_pyodide import run_in_pyodide
+
+
+@run_in_pyodide(packages=["clarabel", "scipy", "numpy"])
```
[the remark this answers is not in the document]
phschiele: Added this in [7759257](https://github.com/pyodide/pyodide/pull/4587/commits/7759257def84979eb76957f7acba7b74f0f1ba5f).
I'm trying to get the test to run without numpy and scipy altogether by mocking the sparse matrix.

ryanking13:
```
pyo3_runtime.PanicException: Failed to load SciPy BLAS bindings.: PyErr { type: <class 'ModuleNotFoundError'>, value: ModuleNotFoundError("No module named 'scipy'"), traceback: None }
```

It looks like `clarabel` depends on scipy and numpy. Then, you need to add 

```yaml
requirements
  run:
    - numpy
    - scipy
```

to clarabel's meta.yaml as well.

packages/cvxpy-base/patches/0001-Do-not-build-C-extensions.patch
phschiele: Thanks a lot!

Opened a PR:
https://github.com/cvxpy/cvxpy/pull/2382/files

The other patched change is not part of the installation process, so it would require the `os.environ` lookup whenever the module is loaded. Is there another way to make such changes as part of the Pyodide build process?

Is it possible to point to a certain commit in our main branch instead of the latest release? The changes that are required to build for Pyodide are not released yet.

packages/cvxpy-base/test_cvxpy.py
```diff
@@ -0,0 +1,47 @@
+from pytest_pyodide import run_in_pyodide
+
+
+@run_in_pyodide(packages=["cvxpy"])
```
ryanking13, line 4: The package name here should follow the name in the meta.yaml. So this needs to be `cvxpy-base`.

packages/cvxpy-base/patches/0001-Do-not-build-C-extensions.patch
ryanking13: > it would require the os.environ lookup whenever the module is loaded. Is there another way to make such changes as part of the Pyodide build process?

I am not sure if there is a way to do that in the build process. But I don't think it's too heavy to look up a dictionary once when the module is loaded.

Note that `"PYODIDE" in os.environ` works in the build time but not in runtime. In runtime, you need to do `sys.platform == "emscripten"` or `"pyodide" in sys.modules` instead (see: https://pyodide.org/en/stable/usage/faq.html#how-to-detect-that-code-is-run-with-pyodide)

ryanking13: > Is it possible to point to a certain commit in our main branch instead of the latest release? The changes that are required to build for Pyodide are not released yet.

GitHub supports downloading an archive from a specific commit: https://stackoverflow.com/questions/13636559/how-to-download-zip-from-github-for-a-particular-commit-sha

packages/cvxpy-base/test_cvxpy.py
phschiele: Done in [c91b392](https://github.com/pyodide/pyodide/pull/4587/commits/c91b392bd7bd7dff0b5dab929d49a6ccc1480e68)

packages/cvxpy-base/patches/0001-Do-not-build-C-extensions.patch
phschiele: The CVXPY PR has been merged with your proposed changes.
Downloading from a commit that includes these changes for now: [e57847b](https://github.com/pyodide/pyodide/pull/4587/commits/e57847b44a90d708303b59c333880997eedee6ef)
We can switch to the PyPI version as soon as the next CVXPY release is out, which is scheduled for the end of March.
CC: @PTNobel

phschiele:
@ryanking13 I have addressed your comments. Looks like it works now for all browsers except firefox; any insights on what might be the cause here?

hoodmane:
You could try raising the timeout or just xfail in firefox.

ryanking13:
Yep, firefox is often slower than other browsers. You can put the decorator like

```python
@pytest.mark.driver_timeout(60)
```

or

```python
@pytest.mark.xfail_browsers(firefox="slow")
```

phschiele:
@hoodmane @ryanking13 Thanks for the feedback, added in [87771f9](https://github.com/pyodide/pyodide/pull/4587/commits/87771f97be4435ab92464164024b85d295f3c836)

CI is failing in `build-core`, which appears to be unrelated to my commit.

hoodmane:
I reran it.

ryanking13 (approved):
Thanks @phschiele! The GHA failure is not related to this PR.

PR #4590, opened 2024-03-05
Make `asyncio.sleep(0)` run faster (take 2)

ryanking13:
### Description

Trying again #4568 with a different approach. Mostly adapted from scheduler in [core-js](https://github.com/zloirock/core-js/blob/master/packages/core-js/internals/task.js) and [setimmediate.js](https://github.com/YuzuJS/setImmediate)

1. If Node, use `setImmidiate`. This API is deprecated in most browsers but exists in Node.
2. If in the browser main thread. Use `postMessage` + event listener
3. If in the web worker (except for safari), use `MessageChannel`.
4. Otherwise, fallback to `settimeout(0)`.

Unfortunetly, safari + web worker will fallback to `settimeout(0)` which is slow, but otherwise it works much faster.

### Checklists

- [x] Add / update tests

### DIFF
--- a/src/js/environments.ts
+++ b/src/js/environments.ts
@@ -31,10 +31,27 @@ export const IN_BROWSER_MAIN_THREAD =
   typeof window !== "undefined" &&
   typeof document !== "undefined" &&
   typeof document.createElement !== "undefined" &&
-  typeof sessionStorage !== "undefined";
+  typeof sessionStorage !== "undefined" &&
+  typeof importScripts === "undefined";
 
 /** @private */
 export const IN_BROWSER_WEB_WORKER =
   IN_BROWSER &&
   typeof importScripts !== "undefined" &&
   typeof self !== "undefined";
+
+/**
+ * Detects the current environment and returns a record with the results.
+ * This function is useful for debugging and testing purposes.
+ */
+export function detectEnvironment(): Record<string, boolean> {
+  return {
+    IN_NODE: IN_NODE,
+    IN_NODE_COMMONJS: IN_NODE_COMMONJS,
+    IN_NODE_ESM: IN_NODE_ESM,
+    IN_DENO: IN_DENO,
+    IN_BROWSER: IN_BROWSER,
+    IN_BROWSER_MAIN_THREAD: IN_BROWSER_MAIN_THREAD,
+    IN_BROWSER_WEB_WORKER: IN_BROWSER_WEB_WORKER,
+  };
+}
--- a/src/js/environments.ts
+++ b/src/js/environments.ts
@@ -41,7 +41,11 @@ export const IN_BROWSER_WEB_WORKER =
   typeof self !== "undefined";
 
 /** @private */
-export const IN_SAFARI = navigator && navigator.userAgent.indexOf("Chrome") == -1 && navigator.userAgent.indexOf("Safari") > -1
+export const IN_SAFARI =
+  typeof navigator !== "undefined" &&
+  typeof navigator.userAgent !== "undefined" &&
+  navigator.userAgent.indexOf("Chrome") == -1 &&
+  navigator.userAgent.indexOf("Safari") > -1;
 
 /**
  * Detects the current environment and returns a record with the results.
--- a/src/js/environments.ts
+++ b/src/js/environments.ts
@@ -40,6 +40,9 @@ export const IN_BROWSER_WEB_WORKER =
   typeof importScripts !== "undefined" &&
   typeof self !== "undefined";
 
+/** @private */
+export const IN_SAFARI = navigator && navigator.userAgent.indexOf("Chrome") == -1 && navigator.userAgent.indexOf("Safari") > -1
+
 /**
  * Detects the current environment and returns a record with the results.
  * This function is useful for debugging and testing purposes.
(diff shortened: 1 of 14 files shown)


src/js/scheduler.ts
```diff
@@ -0,0 +1,93 @@
...
+  const onGlobalMessage = (event: MessageEvent) => {
+    if (
+      typeof event.data === "string" &&
+      event.data.indexOf(scheduleCallbackImmediateMessagePrefix) === 0
+    ) {
+      const handle = +event.data.slice(
+        scheduleCallbackImmediateMessagePrefix.length,
+      );
+      const task = tasks[handle];
+      if (task) {
```
hoodmane, line 30: nit:
```js
if (!task) {
   return;
}
```

src/js/environments.ts
```diff
@@ -31,10 +31,35 @@ export const IN_BROWSER_MAIN_THREAD =
...
+export function detectEnvironment(): Record<string, boolean> {
+  return {
+    IN_NODE: IN_NODE,
+    IN_NODE_COMMONJS: IN_NODE_COMMONJS,
+    IN_NODE_ESM: IN_NODE_ESM,
+    IN_DENO: IN_DENO,
+    IN_BROWSER: IN_BROWSER,
+    IN_BROWSER_MAIN_THREAD: IN_BROWSER_MAIN_THREAD,
+    IN_BROWSER_WEB_WORKER: IN_BROWSER_WEB_WORKER,
+    IN_SAFARI: IN_SAFARI,
```
hoodmane, line 63:
```suggestion
    IN_NODE,
    IN_NODE_COMMONJS,
    IN_NODE_ESM,
    IN_DENO,
    IN_BROWSER,
    IN_BROWSER_MAIN_THREAD,
    IN_BROWSER_WEB_WORKER,
    IN_SAFARI,
```

hoodmane (approved):
Thanks for working this out @ryanking13. It looks quite tricky.

hoodmane:
See also:
https://github.com/WICG/scheduling-apis/blob/main/explainers/yield-and-continuation.md
https://developer.chrome.com/blog/introducing-scheduler-yield-origin-trial/

hoodmane:
Perhaps we should check for `scheduler.yield()` and use that if it exists.

src/js/scheduler.ts
```diff
@@ -0,0 +1,93 @@
...
+    tasks[nextTaskHandle] = callback;
+    globalThis.postMessage(
+      scheduleCallbackImmediateMessagePrefix + nextTaskHandle,
+      "*",
+    );
+    nextTaskHandle++;
+  } else if (IN_BROWSER_WEB_WORKER && !IN_SAFARI) {
+    const channel = new MessageChannel();
+    channel.port1.onmessage = () => callback();
+    channel.port2.postMessage("");
```
hoodmane, line 74: In the core-js implementation they only make one MessageChannel and share it for all messages. Is it really a good idea to allocate a separate MessageChannel every time we call this?

src/js/environments.ts
```diff
@@ -31,10 +31,35 @@ export const IN_BROWSER_MAIN_THREAD =
...
   typeof document !== "undefined" &&
   typeof document.createElement !== "undefined" &&
-  typeof sessionStorage !== "undefined";
+  typeof sessionStorage !== "undefined" &&
+  typeof importScripts === "undefined";
 
 /** @private */
 export const IN_BROWSER_WEB_WORKER =
   IN_BROWSER &&
   typeof importScripts !== "undefined" &&
```
hoodmane, line 40: Can we change this to `typeof importScripts === "function"`

src/js/scheduler.ts
```diff
@@ -0,0 +1,93 @@
...
+    // node has setImmediate, let's use it
+    setImmediate(callback);
+  } else if (IN_BROWSER_MAIN_THREAD) {
+    tasks[nextTaskHandle] = callback;
+    globalThis.postMessage(
+      scheduleCallbackImmediateMessagePrefix + nextTaskHandle,
+      "*",
+    );
+    nextTaskHandle++;
+  } else if (IN_BROWSER_WEB_WORKER && !IN_SAFARI) {
```
hoodmane, line 71: Can we also check that `typeof MessageChannel === "function"` here?

src/js/scheduler.ts
ryanking13: I tried that in [this commit](https://github.com/pyodide/pyodide/pull/4590/commits/b72cf110b8d2fc76d1b59ce60b0c51ade1a3609a), but somehow there was a failure in Node.js. I don't understand why.

src/tests/test_webloop.py
```diff
@@ -497,3 +497,62 @@ async def temp():
         loop._no_in_progress_handler = None
         loop._keyboard_interrupt_handler = None
         loop._system_exit_handler = None
+
+
+@run_in_pyodide
+async def test_zero_timeout(selenium):
```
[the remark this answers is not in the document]
ryanking13: No, safari + main thread seems to work fine with postMessage approach.

ryanking13:
> Perhaps we should check for scheduler.yield() and use that if it exists.

Thanks for the info! I quickly checked if this is available in my local chrome, but it seems like `scheduler.yield` is under the experimental flag. While it's fine to add experimental flag for testing (and we already do for JSPI), let's revisit this when the feature becomes stable, as we have other options for now.

PR #4591, opened 2024-03-05
Add docs for usage with Vite

samestep:
### Description

Currently it is fairly difficult to use the Pyodide npm package with a modern web build tool like Vite. For instance, Googling turns up a few GitHub discussion threads with some tips on how to do this, but they all seem to be some combination of incomplete or out of date. As a first step to addressing that difficulty, this PR adds a doc section on how to use Pyodide from npm in Vite.

Ideally, this will be followed up by improvements to the Pyodide npm package itself so that fewer of these hacky steps are needed.

### Checklists

- [ ] Add / update tests

I've left the test box unchecked, along with the changelog box because from a quick search I don't see any doc-only changes in the current changelog.

### DIFF
--- a/docs/usage/index.md
+++ b/docs/usage/index.md
@@ -161,7 +161,7 @@ export default defineConfig({ optimizeDeps: { exclude: ["pyodide"] } });
 You can test your setup with this `index.html` file:
 
 ```html
-<!DOCTYPE html>
+<!doctype html>
 <html>
   <head>
     <script type="module" src="/src/main.js"></script>
--- a/docs/usage/index.md
+++ b/docs/usage/index.md
@@ -143,7 +143,7 @@ The following instructions have been tested with Vite 5.1.4.
 ```
 
 If you have installed Pyodide via npm as described above, you can use it in Vite
-dev mode as follows. First, install the [`isomorphic-fetch`][] package:
+as follows. First, install the [`isomorphic-fetch`][] package:
 
 ```
 $ npm install --save isomorphic-fetch@^3
--- a/docs/usage/index.md
+++ b/docs/usage/index.md
@@ -135,3 +135,76 @@ undefined
    webworker.md
    service-worker.md
 ```
+
+## Vite
+
+```{note}
+The following instructions have been tested with Vite 5.1.4.
+```
+
+If you have installed Pyodide via npm as described above, you can use it in Vite
+dev mode as follows. First, install the [`isomorphic-fetch`][] package:
+
+```
+$ npm install --save isomorphic-fetch@^3
+```
+
+Then, exclude Pyodide from [Vite's dependency pre-bundling][optimizedeps] by
+setting `optimizeDeps.exclude` in your `vite.config.js` file:
+
+```js
+import { defineConfig } from "vite";
+
+export default defineConfig({ optimizeDeps: { exclude: ["pyodide"] } });
+```
+
+You can test your setup with this `index.html` file:
+
+```html
+<!DOCTYPE html>
+<html>
+  <head>
+    <script type="module" src="/src/main.js"></script>
+  </head>
+</html>
+```
+
+And this `src/main.js` file:
+
+```js
+import { loadPyodide } from "pyodide";
+
+async function hello_python() {
+  let pyodide = await loadPyodide();
+  return pyodide.runPythonAsync("1+1");
+}
+
+hello_python().then((result) => {
+  console.log("Python says that 1+1 =", result);
+});
+```
+
+This should be sufficient for Vite dev mode:
+
+```
+$ npx vite
+```
+
+For a production build, you must also manually make sure that all Pyodide files
+will be available in `dist/assets`, by first copying them to `public/assets`
+before building:
+
+```
+$ mkdir -p public/assets/
+$ cp node_modules/pyodide/* public/assets/
+$ npx vite build
+```
+
+Then you can view this production build to verify that it works:
+
+```
+$ npx vite preview
+```
+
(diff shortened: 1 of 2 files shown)


ryanking13:
Thanks for the contribution @samestep!

Are you interested in making a Vite plugin (or something similar) to manage this?

Pyodide has similar issues with different bundlers, but we can't be experts on every bundler, and we try to avoid making specific bundler-specific changes to Pyodide itself whenever possible. It's also easy for these documents to become outdated, and the next time we have an issue with Vite, the first thing we'll (probably) do is ping you, the original contributor of this documentation.

Therefore, we like to separate bundler-specific issues into other packages that can be managed by the bundler's experts and rely on the community to manage them. For webpack case, we have a [pyodide-webpack-plugin](https://github.com/pyodide/pyodide-webpack-plugin), and we would be happy to make a similar one for vite if you are interested.

samestep:
Thanks for the response @ryanking13! That all makes a lot of sense, I wouldn't want to impose an unreasonable maintenance burden on the Pyodide team.

However, there is a difference between what I'm hoping to achieve and what is achieved by that Pyodide Webpack plugin. If I understand correctly, that plugin loads Pyodide from a CDN. I can already use Pyodide in Vite from a CDN just fine, no special configuration required. What I really want is to be able to consume Pyodide as an npm package like I do with all my other dependencies, and I don't see how making a Vite plugin would achieve this; instead, the only ways I currently see to achieve my goal would be either a hacky solution (which I've documented here) or improvements to the Pyodide npm package itself (this is the one I would prefer).

Part of the hacky solution I've documented here is to work around Pyodide's usage of the `node-fetch` package, so it seems like that's already somewhat alleviated by #4417. The `optimizeDeps.exclude` config part is, in my opinion, an issue with Vite and not with Pyodide, so I'm happy to just go talk to the Vite folks about that rather than dealing with it here.

Then the only thing left would be the problem that currently all the Pyodide files need to be manually made available in `dist/assets`. I haven't investigated this, but my current guess is that the issue is that Pyodide's `loadScript` and `getBinaryResponse` functions in `src/js/compat.ts` are too confusing for Vite's support for [`await import`](https://vitejs.dev/guide/features#dynamic-import) and [`new URL`](https://vitejs.dev/guide/assets#new-url-url-import-meta-url).

I know that it is possible to build an npm package including Wasm code which nicely supports both Node.js and bundlers for browsers, because I've built some before using Rust and [`wasm-bindgen`](https://github.com/rustwasm/wasm-bindgen). In [my particular case](https://github.com/rose-lang/rose/blob/v0.4.7/packages/wasm/package.json#L7-L16), I achieved this by making use of the [`"exports"`](https://nodejs.org/api/packages.html#package-entry-points) entry in `package.json`. I see that you already make use of `"exports"` in `src/js/package.json`, so perhaps it's just a matter of splitting out a `"browser"` condition branch?

Anyways, please let me know if you'd be interested in accepting contributions to improve your npm package in this way. In summary, I do believe that it's possible to serve all these use cases just by using standard JavaScript features like dynamic `import`, `new URL`, and `"exports"`, while continuing to "avoid making specific bundler-specific changes to Pyodide itself" as you mentioned.

hoodmane:
> if you'd be interested in accepting contributions to improve your npm package in this way

Based on our issue tracker activity, these problems create a lot of ongoing headaches for the downstream. I am enthusiastic about anything we can do to fix them. I agree that it is important to do it in a way that doesn't impose too high of an ongoing maintenance burden since we are stretched a little thin here.

docs/usage/index.md
```diff
@@ -135,3 +135,76 @@ undefined
    webworker.md
    service-worker.md
 ```
+
+## Vite
+
+```{note}
+The following instructions have been tested with Vite 5.1.4.
+```
```
hoodmane, line 143: Maybe we should also put the Pyodide version here in case it gets out of date.

docs/usage/index.md
hoodmane, line 1: I'd prefer to see this in a separate file rather than on the index page.

ryanking13: +1

Let's make a separate page e.g. "Working with Bundlers" under this page. So we can put information about other bundlers as well.

samestep: Good idea, thanks! I just pushed a commit attempting to do that, but I don't see it in the sidebar on the [docs build](https://pyodide--4591.org.readthedocs.build/en/4591/usage/index.html). How can I fix the sidebar?

docs/usage/index.md
```diff
@@ -129,6 +129,7 @@ undefined
 .. toctree::
    :hidden:
 
+   working-with-bundlers.md
```
ryanking13, line 132: Let's put this in the bottom of this tree.

image

docs/usage/working-with-bundlers.md
```diff
@@ -0,0 +1,82 @@
...
+project.
+
+## Vite
+
+```{note}
+The following instructions have been tested with Pyodide 0.25.0 and Vite 5.1.4.
+```
+
+If you have installed Pyodide via npm, you can use it in Vite as follows. First,
+install the [`isomorphic-fetch`][] package:
```
ryanking13, line 17: Could you add some explanation why `isomorphic-fetch` is required?

ryanking13 (commented):
Thanks!

samestep:
Thanks for the updated review @ryanking13! I still can't figure out how to get the "Working with Bundlers" page link to show up in the sidebar though; how can I fix this? https://pyodide--4591.org.readthedocs.build/en/4591/usage/index.html

ryanking13 (approved):
Thanks @samestep!

PR #4607, opened 2024-03-12
Always build tblib

hoodmane:
This automatically makes sure everyone gets good tracebacks with `run_in_pyodide` which should make things easier for occasional contributors.

### DIFF
--- a/packages/tblib/meta.yaml
+++ b/packages/tblib/meta.yaml
@@ -4,6 +4,7 @@ package:
   top-level:
     - tblib
   tag:
+    - always
     - core
     - min-scipy-stack
 source:


ryanking13 (approved):
Thanks. Do you think we need to build `pytest` always as well?

hoodmane:
Probably we should. But pytest has a lot of dependencies whereas tblib has none.

hoodmane:
Incidentally the fact that our test suite uses these external packages makes it work a lot better but also we can't test against tip of tree Python unless we can build pytest and tblib there. They are not that likely to break on unstable Python versions but they certainly can. Probably worth it...

Issue #4537, opened 2024-02-21
Specify a time limit for a computation and raise a `TimeOutError`


coatless:
## 🚀 Feature

Allow people to specify a time limit for computation and raise a `TimeOutError` if it is exceeded. The feature can be either set by a global timeout limit or as part of the call to the `runPython*()` functions. 

### Motivation

The goal of this feature is to avoid having the browser tab crash when confronted with something like an infinite loop. 

### Pitch

Without a feature like this, we're more likely to run into issues when trying to recover from problematic programming.

### Alternatives

JSPI support would allow for an easier interrupt sequence. 

### Additional context

Discussed over on [gitter.im](https://matrix.to/#/!tzLDlLJSRpMoEoMBGG:gitter.im/$AfpqUh3TXwpHmfS6FauGa2vYz4-frxTAGgneLuYoEzE?via=gitter.im&via=matrix.org&via=ru-matrix.org)

hoodmane:
I am in favor of implementing this, but it will probably have to sit around as a backlog task for a while. I think we need to add a parametrized call method of some form to Python functions then this could look like:
```js
pyfunc.callWithOptions({ timeoutMs: 2000, kwargs: true}, posarg1, posarg2, {kwarg1, kwarg2})
```

coatless:
@hoodmane if that's the case, maybe consider creating a shelter module. So, you could have something like: 

```js
Shelter.capturePython(code, {
  captureStreams: true,
  captureGraphics: true,
  env: pyodide.globals,
  timeOutMs: 2000
})
```

This would be similar to webR's `Shelter.captureR()` ([Object overview](https://docs.r-wasm.org/webr/latest/objects.html#shelter), [Evaluating Code](https://docs.r-wasm.org/webr/latest/evaluating.html#evaluating-r-code-and-capturing-output-with-capturer), and [API Docs](https://docs.r-wasm.org/webr/latest/api/js/classes/WebR.Shelter.html#capturer)).

hoodmane:
Ah that's a cool api, thanks for the reference @coatless. I really like the look of `myShelter.purge();`, since collecting individual references and destroying them is a pain. Especially in our tests since we fail the test if there are dangling objects. I'm not sure we need something as complicated as that to set a timeout on a call into Python though.

PR #4608, opened 2024-03-12
Rename callSyncifying to callPromising, add callWithOptions

hoodmane:
Since we renamed `syncify` the method name `callSyncifying` doesn't make much sense anymore. It is implemented in terms of a so-called `promising` wasm call, so this lines up with that. OTOH:

* Nobody is likely to know what that refers to
* most people won't need this anyways
* it does return a promise so the name is somewhat descriptive
* the main point of it is to call a sync function and enable stack switching
* so maybe something like `suspendableCall` would be more appropriate?

We now have three boolean parameters for a Js-to-Python call:

1. kwargs
2. promising
3. relaxed

So we'd need 8 functions to cover all combinations of these. Currently we only have 6 of these. Rather than adding the two remaining combinations which will have annoying names, I added `callOptions` which takes an options argument as the first argument. Despite the fact that options usually go as the last argument, I think it makes sense to use the first argument for this so that all remaining args are passed on to Python. There's also a suggestion of adding a timeout parameter #4537 which this would reasonably accomodate.

Is `callOptions` a good name? @rth @ryanking13 anyone have thoughts on names (or anything else)?

### Checklist

- [x] Add / update tests


ryanking13 (commented):
Thanks @hoodmane!

Naming is hard and it's especially hard to come up with an intuitive name for something as technically complex as this.

My feeling for `callPromising` and `callOptions` is:

- callPromising: The relationship to run_sync doesn't seem to be obvious, at least from the name. If what we expect is to use callPromising with run_sync, wouldn't it be nice to have a little more similarity in the name?
- callOptions: callOptions sounds like an object that contains the parameters for the function call, rather than a method name. Maybe `callWithOptions`?

PR #4615, opened 2024-03-13
Fix fatal error when except blocks are interleaved by stack switching

hoodmane:
Before we moved the exception state into the saved ExceptionState as part of saving the Python state but we didn't clear the exception state. This logic was copied from greenlet; I can't find where they clear the exception state either. However, in our case it is definitely wrong. If we stack switch inside an except block and then enter and stack switch inside a second except block, after exiting from everything we end up witht the exception state set instead of cleared (as if we were still in an except block) and worse the exception state is set to an already freed exception. I'm not actually sure why the bug manifests in this particular way, but this change fixes it.

### Checklist

- [x] Add / update tests


ryanking13:
Thanks!

PR #4622, opened 2024-03-17
Keep GitHub Actions up to date with GitHub's Dependabot

cclauss:
Auto-generates pull requests like
* #4157

Which will fix warnings like at the bottom right of https://github.com/pyodide/pyodide/actions/runs/8306232175

* https://docs.github.com/en/code-security/dependabot/working-with-dependabot/keeping-your-actions-up-to-date-with-dependabot
* https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#package-ecosystem

### Checklists

- [ ] Add / update tests

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,13 @@
+# Keep GitHub Actions up to date with GitHub's Dependabot...
+# https://docs.github.com/en/code-security/dependabot/working-with-dependabot/keeping-your-actions-up-to-date-with-dependabot
+# https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#package-ecosystem
+version: 2
+updates:
+  - package-ecosystem: github-actions
+    directory: /
+    groups:
+      github-actions:
+        patterns:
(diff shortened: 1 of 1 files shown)


.github/dependabot.yml
```diff
@@ -0,0 +1,13 @@
...
+version: 2
+updates:
+  - package-ecosystem: github-actions
+    directory: /
+    groups:
+      github-actions:
+        patterns:
+          - "*" # Group all Actions updates into a single larger pull request
+    schedule:
+      interval: weekly
```
ryanking13, line 13: Maybe let's set the interval to monthly. Weekly PRs that update GHA are sometimes cumbersome, and for that reason, we've moved to pre-commit-config updates from weekly to quarterly as well.

ryanking13 (approved):
Thanks @cclauss! This is going to be very convenient.

.github/dependabot.yml
ryanking13: I see. That makes sense. Thanks @cclauss!

cclauss:
Generated:
* #4630

Issue #4603, opened 2024-03-11
Missing dependency in ipython package (0.26.0a3)


mhochsteger:
## 🐛 Bug

Ipython 8.22 depends on "matplotlib-inline", this dependency is missing in the meta.yaml (and also in pyodide-lock.json).

### To Reproduce

Following code is not running in jupyterlite using the pyodide kernel:
```python
import matplotlib.pyplot as plt
plt.plot ([1,2,3])
```
Error message:
```
ModuleNotFoundError: No module named 'matplotlib_inline'
```
You can try my enviroment here: https://mhochsteger.github.io/jupyterlite_test/lab/index.html?path=matplotlib_bug.ipynb
### Expected behavior

Well, no error message and a plot instead :)

### Environment

I built the jupyterlite environment using their demo project and updated the requirements to the latest alpha release.
```
# requirements.txt
ipyevents>=2.0.1
ipympl>=0.8.2
ipywidgets>=8.1.0,<9
jupyterlab-night
jupyterlab-open-url-parameter
jupyterlab~=4.1.3
jupyterlite-core==0.3.0b0
jupyterlite-pyodide-kernel==0.3.0a0
widgetsnbextension==4.0.7
```

```
# build command
jupyter lite build --contents content --output-dir dist --pyodide \
https://github.com/pyodide/pyodide/releases/download/0.26.0a3/pyodide-0.26.0a3.tar.bz2
```

Amazing project, keep up the good work!!!

rth:
Can you please report it to https://github.com/jupyterlite/pyodide-kernel ? It's a packaging issue there, I would say. Because it's also possible to use pyodide in the browser with https://github.com/pyodide/matplotlib-pyodide without using matplotlib-inline but that's not what is used in JupyterLite.

mhochsteger:
Thanks for the quick answer.
I posted it here since you are packaging ipython. Are you suggesting you omit this dependency on purpose and `matplotlib-inline` shoud be a dependency of the pyodide kernel instead?

bollwyvl:
Whether or not `pyodide` uses it directly, as the default backend, upstream explicitly depends on it:

- https://github.com/ipython/ipython/blob/83ff718a24701ad0f4ed2571c3a701d9c3c96226/pyproject.toml#L29

Perhaps a `micropip` equivalent to `pip check` would catch these sort of things.

rth:
[...]

PR #4626, opened 2024-03-19
Ipython dep

mhochsteger:
Description
Fixes https://github.com/pyodide/pyodide/issues/4603

Duplicate/follow up to https://github.com/pyodide/pyodide/pull/4605, since @mkoeppe didn't react and I would like to see matplotlib working in the next pyodide/jupyterlite release.
I set the `maintainer` (myself) and `requirements` fields,


mkoeppe:
Thanks for taking over! LGTM.

hoodmane:
CI is mad that ipython and matplotlib-inline have a circular dependency.

mhochsteger:
I added Ipython and matplotlib as dependencies because the import test won't work without it (see the original  PR). The upstream pip package only depends on traitlets (but makes no sense on it's own).
Is there a way to skip import tests for this package?

mhochsteger:
The import tests are now running locally (matplotlib-inline import tests is not testing anything because `package.top-level` is not set).

It seems to me that the pipeline failures have nothing to do with the changes in this branch.

mhochsteger:
@hoodmane I think it's now just unrelated tests that are failing. The circular dependency is fixed, and the import test for matplotlib-inline is skipped (as it's not working on its own). Is there anything else I can prepare/do?

PR #4631, opened 2024-03-21
Remove unused base-64 dependency and base64 Node compat

ryanking13:
### Description

1. Removes `base-64` dependency in pyodide npm package. This was only used in pytest-pyodide node driver in very old days but was removed in 2 years ago: https://github.com/pyodide/pyodide/pull/2742

2. Removes `base16ToBase64` compat for node. From Node.js >= 16.0, `btoa` is supported in node so we don't need it.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -76,11 +76,15 @@ jobs:
           environment-file: environment.yml
           channels: conda-forge
 
-      - name: Check Python versions
+      - name: Check Versions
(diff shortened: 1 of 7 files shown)


ryanking13:
Hmm, it looks like miniconda installs Node.js 12 and npm 6.9 which is quite old. Not sure if there have been some changes lately...

ryanking13:
Merging to fix CI failure in other PRs.

PR #4637, opened 2024-03-25
Fix toJs on dict subclasses

### LINKED ISSUE #4636: `.toJs()` of a `dict` gets empty map/object
## 🐛 Bug

- Converting a python `dict` to a js map/object is ok.
- Convertinig an instance of a subclass of `dict` to a js map/object is ok too.
- But Converting an instance of a subclass of both `dict` and `ChainMap` to a js map/object gets an empty map/object

### To Reproduce

First, make a JavaScript function to show what it gets:

```py
>>> from pyodide.code import run_js
>>> j = run_js("""(d) => JSON.stringify(d.toJs({ dict_converter: Object.fromEntries }))""")
>>> j({"a": "b"})
'{"a": "b"}'
```

Then we make a subclass `D` to extend `dict`, its converting works fine:

```py
>>> class D(dict):
...     pass
...
>>> j(D({"a": "b"}))
'{"a":"b"}'
```

But this fails:

```py
>>> from collections import ChainMap
>>> class D(ChainMap, dict):
...     pass
...
>>> j(D({"a": "b"}))
'{}'
```

Plus, I found that the mro matters:

```py
>>> class D(dict, ChainMap):  # here we change the order of base classes
...     pass
...
>>> j(D({"a": "b"}))
'{"a":"b"}'
```

But it should not be like this. I noticed the `dict_converter` receives an iterator of `[key, value]`, isn't it comes from the `.items()` method?

The `D` instances above do have a right `.items()` method but why the result is empty? So I believe this is a bug. I haven't find related information in the document yet.

### Expected behavior

```py
>>> from collections import ChainMap
>>> class D(ChainMap, dict):
...     pass
...
>>> j(D({"a": "b"}))  # this shouldn't returns '{}'
'{"a":"b"}'
```

### Environment

- Pyodide Version `0.25.0`
- Browser version `Edge 122.0.2365.92`

hoodmane:
Resolves #4636. `PyDict_Next` behaves inconsistently on dict subclasses so get items and iterate those instead.

- [x] Add / update tests

### DIFF
--- a/docs/project/changelog.md
+++ b/docs/project/changelog.md
@@ -78,6 +78,9 @@ myst:
   access unless they match a function on the `Map` prototype.
   {pr}`4576`
 
+- {{ Fix }} `toJs` now works as expected on subclasses of `dict`.
+  {pr}`4637`
+
 ### Packages
 
 - New Packages: `cysignals`, `ppl`, `pplpy` {pr}`4407`, `flint`, `python-flint` {pr}`4410`,
--- a/src/core/python2js.c
+++ b/src/core/python2js.c
@@ -234,14 +234,31 @@ static JsVal
 _python2js_dict(ConversionContext* context, PyObject* x)
 {
   bool success = false;
+  PyObject* items = NULL;
+  PyObject* iter = NULL;
+  PyObject* item = NULL;
 
+  _Py_IDENTIFIER(items);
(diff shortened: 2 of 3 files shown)


ryanking13 (approved):

PR #4639, opened 2024-03-26
Fix stack switching leaks

hoodmane:
Stack switching leaks datastack chunks. It is very confusing, but the leaked chunks are allocated in slabs of size 2^14 so it doesn't take long to leak a substantial amount of memory.

- [x] Add / update tests


ryanking13:
Thanks for the fix!

Issue #4557, opened 2024-02-24
RFC Decoupling the version of pyodide-build and Pyodide


ryanking13:
## Proposal Overview

This issue proposes to decouple the versioning of the pyodide-build and the Pyodide distribution, so that we can release new versions of the pyodide-build without having to release a new version of the Pyodide distribution.

1. This will allow us to release new versions of the pyodide-build more frequently, and to make it easier to maintain the pyodide-build.
2. This aligns with the goal of the out-of-tree build, as it is becoming more and more impossible to maintain the full packages sets in the Pyodide distribution, and we need to delegate the maintenance of the packages and recipes to the community.

As this has been discussed a few times in the past since we started supporting out-of-tree builds, I believe we have consensus on this proposal. So this issue mostly focuses on a way to implement this.

## Current Challenges

Currently, the version of the pyodide-build is highly coupled with the version of the Pyodide distribution.

- all build envs are defined in the Makefile.envs file, which is part of the Pyodide distribution.
- pyodide-build’s source code resides in the same repository as Pyodide.
- certain test dependencies for Pyodide are defined within pyodide-build.
- etc

### Support for multiple xbuildenv versions in pyodide-build

Currently, the pyodide-build only supports to install a corresponding version of the xbuildenv. We need to make it possible to install multiple versions of the Pyodide xbuildenv in a single pyodide-build version.

I think the expected form is for pyodide-build to behave like a _version manager, like emsdk, pyenv, or nvm_. This involves introducing new subcommands for listing, installing, and using various xbuildenv versions:

- `pyodide xbuildenv list`: List all possible xbuildenv versions.
- `pyodide xbuildenv install <version>`: Install a specific version of the xbuildenv.
  - This command already exists but for now it only supports to install a corresponding version of the xbuildenv.
- `pyodide xbuildenv use <version>`: Use a specific version of the xbuildenv.

To support this, we will need to maintain a metadata file that contains the information of the xbuildenv versions. This is similar to how [emsdk manages the versions of the emscripten SDK](https://github.com/emscripten-core/emsdk/blob/main/emscripten-releases-tags.json). This metadata file should be hosted on the web separately from pyodide-build so that we can update it when a new version of Pyodide (xbuildenv) is released.

```json
{
    "versions": [
        {
            "version": "0.26.0a1",
            "url": "<xbuildenv tarball url>",
            "sha256": "<sha256 of the tarball>",
            "python_version": "3.12.1",
            "emscripten_version": "...",
[...]

PR #4640, opened 2024-03-26
Add CrossBuildEnvManager class for managing xbuildenv

ryanking13:
### Description

Related: https://github.com/pyodide/pyodide/issues/4557

This adds `CrossBuildEnvManager` class. It is basically a class for managing xbuildenvs. It can install, uninstall xbuildenv versions, switch between installed versions, and list installed versions.

I also added corresponding subcommands for those methods (command names are inspired from pyenv):

- `pyodide xbuildenv versions`: lists all installed xbuildenv versions
- `pyodide xbuildenv version`: show current activated version
- `pyodide xbuildenv use`: select xbuildenv version to use
- `pyodide xbuildenv uninstall`: uninstall installed xbuildenv.

Note that xbuildenv subcommands are private (not displayed in `pyodide --help` or in our docs) for now, as it is not very useful for end users. I'll open more follow up PRs soon.

### Checklists

- [x] Add / update tests


cclauss:
Is https://pypi.org/project/pytest_httpserver missing?

ryanking13:
> Is https://pypi.org/project/pytest_httpserver missing?

Yeah, looks like I forgot to add it to the test dependency. Thanks!

pyodide-build/pyodide_build/create_pypi_index.py
```diff
@@ -4,15 +4,15 @@
 from pyodide_lock.spec import PackageSpec
 
 
-def create_pypa_index(
+def create_pypi_index(
```
hoodmane, line 7: I think this was meant to say "python package authority" = pypa not "python package index" = pypi. pypi is a specific package index. Maybe rename it to just `create_index` and call it `index` everywhere?

src/tests/test_cmdline_runner.py
```diff
@@ -476,15 +476,16 @@ def test_pypa_index(tmp_path):
...
         encoding="utf8",
     )
-    print("\n\nstdout:")
-    print(result.stdout)
-    print("\n\nstderr:")
-    print(result.stderr)
+    # print("\n\nstdout:")
+    # print(result.stdout)
+    # print("\n\nstderr:")
+    # print(result.stderr)
```
hoodmane, line 482: Probably should either keep these not commented out or remove them.

pyodide-build/pyodide_build/xbuildenv.py
```diff
@@ -0,0 +1,332 @@
...
+
+    def __init__(self, env_dir: str | Path) -> None:
+        self.env_dir = Path(env_dir).resolve()
+
+        try:
+            self.env_dir.mkdir(parents=True, exist_ok=True)
+        except Exception as e:
+            raise ValueError(
+                f"Failed to create cross-build environment at {self.env_dir}"
+            ) from e
```
hoodmane, line 31: Note that if you raise an exception in an `except` block, it is automatically chained to the original exception so it would be the same to do:
```py
except Exception:
  raise ValueError(...)
```

pyodide-build/pyodide_build/xbuildenv.py
```diff
@@ -0,0 +1,332 @@
...
+
+    def list_versions(self) -> list[str]:
+        """
+        List the downloaded xbuildenv versions.
+
+        TODO: add a parameter to list only compatible versions
+        """
+        versions = []
+        for version_dir in self.env_dir.glob("*"):
+            if not version_dir.is_dir() or str(version_dir) == str(self.symlink_dir):
```
hoodmane, line 62: You can just compare the paths directly:
```suggestion
            if not version_dir.is_dir() or version_dir == self.symlink_dir:
```

pyodide-build/pyodide_build/xbuildenv.py
```diff
@@ -0,0 +1,332 @@
...
+            If True, skip installing the cross-build packages. This is mostly for testing purposes.
+
+        Returns
+        -------
+        Path to the root directory for the cross-build environment.
+        """
+
+        if url:
+            if version:
+                raise ValueError("Cannot specify both version and url")
```
hoodmane, line 138: Nit: I like to separate the error case into its own block:
```suggestion
        if url and version:
            raise ValueError("Cannot specify both version and url")
        if url:
```

hoodmane (approved):
Thanks, looks like this should be a lot more user friendly.

pyodide-build/pyodide_build/create_pypi_index.py
ryanking13: Makes sense. Let's call it `package index` then.

ryanking13:
Thanks for the review!

PR #4641, opened 2024-03-27
Updated python-sat to version 1.8.dev6.

alexeyignatiev:
This PR updates `python-sat` to the most recent version (`1.8.dev6`). There are a large number of changes and fixes since the last version supported in Pyodide, including additional SAT solvers, arbitrary Boolean formula manipulation and on-the-fly clausification as well as external user-defined engines to be attached to solvers supporting the IPASIR-UP interface.

### DIFF
--- a/packages/python-sat/meta.yaml
+++ b/packages/python-sat/meta.yaml
@@ -1,15 +1,16 @@
 package:
   name: python-sat
-  version: 0.1.7.dev26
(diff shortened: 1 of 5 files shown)


ryanking13 (approved):
Thanks @alexeyignatiev!

alexeyignatiev:
Thanks, @ryanking13!

PR #4651, opened 2024-03-29
MAINT Remove include_js_file

hoodmane:
We can import these from src/js/ffi.ts now and I think it's a better way to go.


Makefile
```diff
@@ -190,15 +190,26 @@ src/js/generated/pyproxy.ts : src/core/pyproxy.* src/core/*.h
...
-	rm -f $@
 	echo "// This file is generated by applying the C preprocessor to core/pyproxy.ts" >> $@
 	echo "// It uses the macros defined in core/pyproxy.c" >> $@
 	echo "// Do not edit it directly!" >> $@
 	cat src/core/pyproxy.ts | \
 		sed '/^\/\/\s*pyodide-skip/,/^\/\/\s*end-pyodide-skip/d' | \
 		$(CC) -E -C -P -imacros src/core/pyproxy.c $(MAIN_MODULE_CFLAGS) - | \
 		sed 's/^#pragma clang.*//g' \
-		>> $@
+		> $@
```
ryanking13, line 200: Doesn't changing `>>` to `>` will remove the comments above?

src/js/ffi.ts
```diff
@@ -22,6 +22,8 @@ export type {
 export type { TypedArray } from "./types";
 
 export type { PythonError } from "../core/error_handling";
+import "generated/js2python";
+import "generated/python2js_buffer";
```
ryanking13, line 26: Let's adjust the import order so that imports from `generated`  be located together.

ryanking13 (approved):
Thanks! I have some minor comments.

src/js/ffi.ts
hoodmane: I think `generated/js2python` has to be after `../core/error_handling` because js2python uses `Module._PropagatePythonError` and error_handling assigns to it. But I should switch to using imports for this now that we can.

PR #4652, opened 2024-03-29
Memory snapshots

hoodmane:
Add support for memory snapshots


hoodmane:
Test suite ran in 979.65s on the main branch vs 898.47s on this branch, so about 10% faster.

hoodmane:
test-packages-node ran in 746.37s here vs 883.83s in node.

hoodmane:
### Without snapshot:

image

### With snapshot:

image

rth:
Very impressive! So this is re-loading Pyodide, right, where the snapshot was creating the first time it is loaded? Or is it possible to distribute snapshots directly?

hoodmane:
The snapshot is specific to the Pyodide build (it seems that emscripten's linker is not 100% deterministic). But it can be created just once and distributed as part of the core distribution. Two concerns about randomness:
1. This disables hash randomization. People who are concerned can either disable snapshots or recreate the snapshot occasionally with a new hash seed.
2. We need to reseed the random number generator before executing user code or else they will all get the same random numbers. This is as simple as `from random import seed; seed()`.

ryanking13:
Very interesting indeed. Do you think this can be used to snapshot the running filesystem (MEMFS)?

ryanking13:
Oh, I just read the Cloudflare blog post you wrote. It's so cool that Pyodide is integrated with Cloudflare workers! I think it will definitely push Pyodide forward in dealing with the complex technical hurdles that come with integrating Pyodide with other services. I was quite impressed with the memory snapshots that improved initialization speeds in that post.

pfebrer:
I can test this on my application if you have some instructions on how to use it to save and load an environment 🙂

hoodmane:
@pfebrer you can make a snapshot with 
```
node dist/snapshot.bin
``` 
Then you should be able to do
```js
const snapshotPromise = fetch("snapshot.bin").then(resp => resp.arrayBuffer());
const pyodide = loadPyodide({ _loadSnapshot: snapshotPromise});
```
and that'll load from the snapshot. You can't currently collect anything other than basic Pyodide initialization into the snapshot though.

pfebrer:
Thanks!
And I guess it is not as simple as moving the

```javascript
if (options._makeSnapshot) {
    // @ts-ignore
    pyodide._snapshot = Module.HEAP8.slice();
  }
```

part to the end of `loadPyodide` to include the loaded packages in `loadPyodide({packages: [...this packages]})`, right? :sweat_smile:

hoodmane:
The main issue is that you'll want to take the snapshot not just after the packages are loaded which won't save hardly any time but also after they are imported because most of the import time work happens then. Ideally you'd also want to snapshot whatever application-specific initialization work you need to do. So:
1. We need to set up the JS FFI in a way that allows us to serialize the FFI state.
2. We need to ensure that when dynamic libraries are loaded their shared library metadata is stored into the same place

For 1, I think we can use for the JS globals a special wrapper:

<details><summary>Details</summary>
<p>

```js
const accessorListSymbol = Symbol("accessorListSymbol");
function makeProxy(obj, accessorList) {
  return new Proxy(obj, {
    get(target, prop, receiver) {
      if (prop === accessorListSymbol) {
        return accessorList;
      }
      const orig = Reflect.get(...arguments);
      if (!["object", "function"].includes(typeof orig)) {
        return orig;
      }
      return makeProxy(orig, [...accessorList, prop]);
    },
    getPrototypeOf() {
      return makeProxy(Reflect.getPrototypeOf(...arguments), [
        ...accessorList,
        "[getProtoTypeOf]",
      ]);
    },
    // has, ownKeys, isExtensible left alone
    apply() {
      throw new Error("Apply" + accessorList.join("."));
    },
    // ... most other proxy handlers throw errors
  });
}
const jsglobals = makeProxy(globalThis, []);
const py = await loadPyodide({ jsglobals });
```

</p>
</details> 

We can use this to serialize the FFI externref tables. For 2, we need a patch to the Emscripten dynamic loading code that allocates dynamic library metadata.

pfebrer:
Ok, I see. I can't help with that but whenever you want someone to test it I'm in :)

Also how does the compatibility of these snapshots work across different browser versions, different browsers or different OS? (keeping the same pyodide version) Would you need to have a snapshot for each case?

hoodmane:
> how does the compatibility of these snapshots work across different browser versions, different browsers or different OS? (keeping the same pyodide version) Would you need to have a snapshot for each case?

The same snapshot should work in all of these cases. But it will depend on the Pyodide version.

pfebrer:
Awesome! Can't wait to try it then :smile:

.circleci/config.yml
```diff
@@ -263,6 +263,8 @@ jobs:
           name: Py-compile packages
           command: |
             pyodide py-compile --compression-level 0 dist/
+            rm dist/snapshot.bin
```
ryanking13, line 266: nit:

```suggestion
            rm -f dist/snapshot.bin
```

For safety. (don't raise error even if snapshot.bin does not exist).

Makefile
```diff
@@ -231,6 +232,13 @@ dist/python_stdlib.zip: $(call rwildcard,src/py/*) $(CPYTHONLIB)
 dist/test.html: src/templates/test.html
 	cp $< $@
 
+dist/makesnap.mjs: src/templates/makesnap.mjs
+	cp $< $@
+
+dist/snapshot.bin: all-but-packages dist/pyodide-lock.json dist/makesnap.mjs
+	cd dist && node --experimental-wasm-stack-switching makesnap.mjs
```
ryanking13, line 239: Question: Is `--experimental-wasm-stack-switching` always needed making a snapshot?

conftest.py
```diff
@@ -297,3 +297,23 @@ def strip_assertions_stderr(messages: Sequence[str]) -> list[str]:
...
+        snap = new Uint8Array(snap.buffer);
+        let pyodide = await loadPyodide({ fullStdLib: false, jsglobals : self, _loadSnapshot: snap });
+        self.pyodide = pyodide;
+        globalThis.pyodide = pyodide;
+        pyodide._api.inTestHoist = true; // improve some error messages for tests
+        """
+    )
+
+
+NodeRunner.load_pyodide = patched_load_pyodide
```
ryanking13, line 319: This patch is not very satisfactory. But I am fine if this is temporary.

src/js/pyodide.ts
```diff
@@ -214,6 +222,18 @@ export async function loadPyodide(
     await loadScript(scriptSrc);
   }
 
+  let snapshot;
+  if (options._loadSnapshot) {
+    snapshot = await options._loadSnapshot;
```
ryanking13, line 227: awaiting Uint8Array? Why?

src/js/pyodide.ts
```diff
@@ -214,6 +222,18 @@ export async function loadPyodide(
...
+    if (snapshot?.constructor?.name === "ArrayBuffer") {
```
ryanking13, line 228: Maybe the type for `_snapshot` needs to be updated?

src/js/api.ts
```diff
@@ -701,14 +701,59 @@ API.bootstrapFinalizedPromise = new Promise<void>(
...
+    get: () =>
+      API.public_api.toPy(
+        Object.getOwnPropertyNames(o).filter((name) => name !== "__all__"),
+      ),
+    enumerable: false,
+    configurable: true,
+  });
+}
+
+function syncUpSnapshotLoad1() {
```
ryanking13, line 718: Let's add some comments for these functions. Otherwise it would be very difficult to understand for others.

ryanking13 (commented):

Makefile
hoodmane: No, but the variable for whether or not it is turned on is stored in the snapshot. Since we run `test_stack_switching.py` against node and assume it will be on, we need this for the tests to pass.

conftest.py
hoodmane: Yeah ideally we should upstream into pytest pyodide maybe...

src/js/pyodide.ts
hoodmane: Changed to `Uint8Array | ArrayBuffer | PromiseLike<Uint8Array | ArrayBuffer>`.

pfebrer:
What has been merged is only the snapshots of basic pyodide initialization, right? Are there plans for the rest as of now?

PR #4653, opened 2024-03-30
Fix build error on pypa/build >= 1.2.0

ryanking13:
We started getting build errors after pypa/build release 1.2.0.

The private attribute we were using `isolated_env._scripts_dir` was removed in the new pypa/build release as they started to support multiple installers. This PR fixes it + pins pypa/build minor version.


hoodmane (approved):
Thanks for the quick fix @ryanking13!

ryanking13:
Also updated lightgbm version to make it compatible to build>=1.1.0 (https://github.com/pypa/build/pull/716)

PR #4656, opened 2024-03-31
Updated python-sat to version 1.8.dev7.

alexeyignatiev:
Updated python-sat to the vanilla head, fixing a bug in formula key flattening.


ryanking13 (approved):
Thanks!

PR #4662, opened 2024-04-02
Opt into JSPI origin trial

hoodmane:
This opts console.html into the JSPI origin trial

### DIFF
--- a/src/templates/console.html
+++ b/src/templates/console.html
@@ -2,6 +2,10 @@
 <html>
   <head>
     <meta charset="UTF-8" />
+    <meta
+      http-equiv="origin-trial"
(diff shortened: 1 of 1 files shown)


ryanking13:
Thanks! Indeed it would be nice to test JSPI features without enabling experiment features manually. I have some random questions and suggestions.

1. When decoded, that token has the origin: `pyodide.org.` So does it still work if we use it locally, in non pyodide.org?
2. The token seems to expire on Oct 29, maybe we should add a comment there to prevent it from expiring?
  a. The users will get different results if they access the console before / after Oct 29, are we okay with this?

hoodmane:
Good questions. Maybe we should only put it in the `/latest` console and not into the version we include with the release.

hoodmane:
This one should work for `localhost:8000`. I'll add it.

Ai8IXb0XqedlM/Q2guWXFfBkKiYY9uaPZpdjHqc8y0ZvpAfK9SKzp/dIuFH+txG/HEKxt59uIkk39hhWrhNgbw4AAABieyJvcmlnaW4iOiJodHRwOi8vbG9jYWxob3N0OjgwMDAiLCJmZWF0dXJlIjoiV2ViQXNzZW1ibHlKU1Byb21pc2VJbnRlZ3JhdGlvbiIsImV4cGlyeSI6MTczMDI0NjM5OX0=

ryanking13:
> Maybe we should only put it in the /latest console and not into the version we include with the release.

Sounds good to me.

PR #4671, opened 2024-04-06
Updated meta.yaml [python-sat package].

alexeyignatiev:
Updated the version of PySAT to 1.8.dev8, which adds support for CaDiCaL 1.9.5 with external reasoning engines to multiple (example) problem-solving scripts.


ryanking13 (approved):
Thanks!

PR #4694, opened 2024-04-13
BLD Prevent redundant xbuildenv installation if there is already a version in use

ryanking13:
### Description

After #4640, we can install multiple xbuildenvs simultaneously. As a follow up, this PR fixes `init_environment()` to use the currently specified version of xbuildenv instead of always using the xbuildenv that matches the pyodide-build version.

### Checklists

- [x] Add / update tests

### DIFF
--- a/pyodide-build/pyodide_build/tests/fixture.py
+++ b/pyodide-build/pyodide_build/tests/fixture.py
@@ -69,8 +69,6 @@ def reset_cache():
 def xbuildenv(selenium, tmp_path, reset_env_vars, reset_cache):
     import subprocess as sp
 
-    from pyodide_build import __version__
-
     assert "PYODIDE_ROOT" not in os.environ
 
     envpath = Path(tmp_path) / xbuildenv_dirname()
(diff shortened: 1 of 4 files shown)


pyodide-build/pyodide_build/tests/fixture.py
```diff
@@ -88,7 +86,7 @@ def xbuildenv(selenium, tmp_path, reset_env_vars, reset_cache):
 
     assert result.returncode == 0
 
-    version_dir = envpath / __version__
+    version_dir = envpath / "temp_version"
```
hoodmane, line 89: Where does `temp_version` come from?

ryanking13: It is a random name to test modifications to this PR. Before this PR, if there is no folder with a name correspond to the `__version__` (e.g. 0.26.0), it assumed that xbuildenv was not installed and downloaded that version remotely.

So to test this, I renamed the xbuildenv folder to something random other than __version__ to see if it work.

Issue #2933, opened 2022-08-02
pyarrow


clstaudt:
## 🐍 Package Request

- Package Name and Version: pyarrow / latest
- Package URL: https://github.com/apache/arrow
- Package Dependencies that needs to be resolved first: ?

clstaudt:
I'd be happy to contribute but this does not seem like an entry-level migration.

hoodmane:
To start, you should try `python -m pyodide_build mkpkg pyarrow` and then `PYODIDE_PACKAGES=pyarrow make -C packages`. If that fails you can post the errors you get here and we will try to help fix them.

hoodmane:
I guess it does sound a bit complicated:
https://github.com/apache/arrow/blob/master/docs/source/developers/python.rst

hoodmane:
You might have to hope that @ryanking13 takes an interest.

clstaudt:
pyarrow is one option for .parquet file support, the other being fastparquet. Neither are in pyodide so far. The file format is really common though.

hoodmane:
Fastparquet looks to me a bit simpler because the dependencies are also Python packages.

Mause:
DuckDB is a possibility as well, as it has parquet support, and is already compilable to WASM -> https://github.com/duckdb/duckdb-wasm

(Full disclosure, I work on it 😅)

hoodmane:
Does duckb have python bindings? It isn't obvious to me from the repos.

Mause:
> Does duckb have python bindings? It isn't obvious to me from the repos.

Yep! https://pypi.org/project/duckdb/

IanCal:
Duckdb builds without other issues, although I've not tried building the extensions. https://github.com/iancal/pyodide

Pyarrow would solve a lot of other issues though, including loading remote parquet files, and play nicely with duckdb so hopefully both can be included.

changhiskhan:
Just want to toss in my two cents : would absolutely LOVE to see BOTH pyarrow and duckdb supported. pyarrow makes it super easy to integrate with other python tools in the ecosystem. And also it would give js tools a fast way to work with large amounts of remote data via arrowjs

rth:
To give a status update on this, emscripten-forge does have [arrow packaged](https://github.com/emscripten-forge/recipes/blob/main/recipes/recipes_emscripten/arrow/recipe.yaml) by @DerThorsten so it does work with wasm. For pyodide someone would just need to do the work.

[...]

PR #4698, opened 2024-04-16
fix for exceptions in dynamic linked int64 code

joemarshall:
This is a backport of https://github.com/emscripten-core/emscripten/pull/21759 

It fixes a crash which happens when 
1) An exception is thrown inside a function returning int64
2) The function is in a dynamic library and has been called through dynamic invocation.

I could really do with this getting into the next release of pyodide as it is a blocker for #2933 , so I can't send people wheels to test  yet.

### DIFF
--- a/docs/project/changelog.md
+++ b/docs/project/changelog.md
@@ -16,6 +16,8 @@ myst:
 
 ## Unreleased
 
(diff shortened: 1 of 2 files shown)


hoodmane (approved):
Thanks!

PR #4699, opened 2024-04-16
Add JsProxy bindings system

hoodmane:
We've gotten a lot of complaints about how hard it is to call certain JS functions, but it's hard to do much to improve the situation without knowing anything about the specific function being called.

However, if we know about the API, we can choose how to convert to/from JS as appropriate and make things a lot easier. This adds a method to JsProxy called `bind_sig` which adjusts how `__getattr__` and `__call__` work on the proxy so we can choose our converters. I added a class called JsSignature which guides what we do when we call the function. I completely rewrote `JsProxy_Vectorcall` to use this. I added handling for Promise results so we can automatically bind a signature to the result of the Promise.

There's a lot leftover to do:
* `Generator` and `AsyncGenerator` types
* Alternatives `A | B` (try A converter and if it raises a TypeError fall back to B)
* `TypedDict` converters
* Make calling certain objects call `Reflect.construct`
* Cleaner `__getattr__` handling, `__setattr__` handling
* A way to raise AttributeError if the attribute is missing from the signature even if it's present in the JS object (important for backwards compatibility)
* Signatures for JS stdlib functions

My idea is that the signatures should simultaneously function as mypy typehints as much as this is practical.

@ryanking13 I would appreciate if you could just review the tests here to see if you agree with the behavior asserted there. I'm planning to split the actual implementation up into much smaller PRs.


src/tests/test_typeconversions.py
```diff
@@ -1883,3 +1883,307 @@ def test_html_array(selenium):
...
+    ) -> Annotated[list[int], Deep]:
+        raise NotImplementedError
+
+    f = run_js(
+        """
+        (function f(x, y, z) {
+            return [x.get("a"), y.b, z.c]
+        })
+        """
+    ).bind_sig(f)
```
ryanking13, line 1932: To many `f` here, which is confusing 😂

ryanking13 (commented):

It's a very interesting idea. I mean, it's not a very satisfying solution in the sense that we have to define a duplicate schema of the JS object in Python, but it's a reasonable way IMO.

I don't think this is going to be a very commonly used feature, but I think it is possible that some people writing libraries using Pyodide could use it to create a better bridge between JS<->Python (like in `webtypy`?)

So overall, I think the proposal in this PR is fine.

PR #4700, opened 2024-04-16
MAINT Improve dist/python_stdlib.zip rule

hoodmane:
Before this change it only checked for files in `pyodide` or `_pyodide` directory, this makes it look at all files inside the `src/py` directory.


ryanking13:
Good catch. Thanks!

PR #4702, opened 2024-04-18
Bumped python-sat from 1.8.dev10 to 1.8.dev12

alexeyignatiev:
This brings a few fixes of the issues related to arbitrary formula simplification and on-the-fly clausification. Namely, the same formula can have two sets of clauses associated with it depending on whether it is an outermost formula or a sub-formula. Also, this modifies the way Boolean constants are stored and handled.


ryanking13 (approved):
Thanks!

PR #4705, opened 2024-04-19
fix: support cmake command mode

### LINKED ISSUE #4704: cmake -E capabilities is broken
## 🐛 Bug

The most recent version of scikit-build-core switched to using `cmake -E capabilities` instead of `cmake --version`, as that's the recommended way to get information about CMake. https://github.com/scikit-build/scikit-build-core/pull/675 However, https://github.com/pyodide/pyodide/blob/975565bc3f14daf2d22545275ea9680f49a979f3/pyodide-build/pyodide_build/pywasmcross.py#L493 is adding `--fresh` to all cmake commands, which breaks this with "`-E capabilities accepts no additional arguments`". `cmake --version` doesn't care about the extra flag. (I assume this also may break all other `cmake -E` commands? Also `-P` (script) commands?)

Edit: -P is already checked, is -E needs to be added.

I think if this is going to be forcefully injected, it needs to be done more carefully based on the command being passed.

henryiii:
### Description

Fix #4704.

### Checklists

- [ ] Add / update tests

### DIFF
--- a/docs/project/changelog.md
+++ b/docs/project/changelog.md
@@ -16,6 +16,8 @@ myst:
 
 ## Unreleased
 
+- {{ Fix }} Pass through `-E` (command mode) arguments in CMake wrapper {pr}`4705`.
+
 - {{ Fix }} Fix exception handling in dynamic linking of int64 functions {pr}`4698`.
 
 - {{ Enhancement }} `str(jsproxy)` has been adjusted to not raise an error if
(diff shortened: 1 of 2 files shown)


ryanking13 (approved):
Thanks, @henryiii! Sounds reasonable to me. Could you update [the changelog](https://github.com/pyodide/pyodide/blob/main/docs/project/changelog.md)?

hoodmane:
Thanks!

PR #4711, opened 2024-04-22
More rearrangement to reduce diff in #4699

hoodmane:
More nonfunctional changes to reduce diff in #4699. Reduces the diff there by about 400 lines.


ryanking13 (approved):
Thanks for the rearrangements!

PR #4715, opened 2024-04-24
Emscripten 3.1.58

hoodmane:


hoodmane:
No it's this one:

> In -sMODULARIZE mode, the argument passed into the module constructor is no longer mutated in place. The expectation is that the module instance will be available via the constructor return value. Attempting to access methods on the object passed into the constructor will now abort

Hanae1963 (commented):
Y

ryanking13 (approved):
Thanks!

Issue #4674, opened 2024-04-06
pywasmcross exports list can be *way* too long and break builds


joemarshall:
In `pywasmcross.get_export_flags`, all exports are listed, and then added to a `-SEXPORTED_FUNCTIONS` command line. If you do this on something with a large number of exports, the command line becomes too long and the call to emcc fails.

I found this in building pyarrow, unless I set exports to 'whole_archive' the build fails with an 8 megabyte command line. Exporting everything is an okay temporary workaround, but not really a long term solution.

It should probably use a response file here given emcc supports response files for exported function lists.

ryanking13:
What build tool does pyarrow use? Why isn't the response file used currently? I think using response files is something that should be handled by the build tool, not by pywasmcross or emcc. For example, in CMake, we are [telling CMake](https://github.com/pyodide/pyodide/blob/fe9038e68b7dbddf6673f05084fdc024e075762b/pyodide-build/pyodide_build/tools/cmake/Modules/Platform/Emscripten.cmake#L81-L86) to use response files to prevent getting the same error, and the rest of process it handled by CMake.

joemarshall:
@ryanking13 It uses a disturbing combination of setuptools calling through to cmake for library build.

Response files work for everything else (libraries, source files etc.), the limitation here is that when linking with `pywasmcross`, it goes through all the object files to list exports, and then adds those to a `-SEXPORTED_FUNCTIONS` command line, which is passed on the command line. If you do this with a library that has approx 10,000 exports then bad things happen. This is the line where the command line is generated:

https://github.com/pyodide/pyodide/blob/fe9038e68b7dbddf6673f05084fdc024e075762b/pyodide-build/pyodide_build/pywasmcross.py#L431

If instead it generated a response file for the export names that would work in this case because emcc supports response files in this case. Because the list of exports is generated by pywasmcross we don't have access to it in cmake I think (although if you can work out a workaround that would be great)

hoodmane:
we've had this a TODO for years.

joemarshall:
Thinking about it, right now for things that have sub-libraries (we have libarrow-python, which is then used by all the submodules of pyarrow), the export argument to pyodide build is somewhat limited. 

Here I think ideally I need to export everything in libarrow-python, and then export only python entry points in the submodules. Is there a way to do this in cmake options or in setup.py when I'm running inside pyodide build / setuptools, or are the build options set somewhere outside of the control of those bits?

ryanking13:
[...]

PR #4717, opened 2024-04-24
FIX Use response file for EXPORTED_FUNCTIONS

ryanking13:
### Description

Resolve https://github.com/pyodide/pyodide/issues/4674.

Fixes "argument too long" issue when list of exports are too long. It was a fairly easy fix compared to the length of time this issue was left unattended.

Adding tests for this function is a bit tricky since it requires `emcc` to test it, so I didn't add any tests in this PR. For now, we can verify the behavior by checking that the recipes we were having trouble with build fine.

I'll work on improving the test pipeline so that we can test the part of pyodide-build that uses emcc.

### Checklists

### DIFF
--- a/docs/project/changelog.md
+++ b/docs/project/changelog.md
@@ -16,6 +16,9 @@ myst:
 
 ## Unreleased
 
+- {{ Fix }} pyodide-build now use response file when passing list of exported symbols to `emcc`.
+  This Fixes "Argument list too long" error.
+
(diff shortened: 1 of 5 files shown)


hoodmane (approved):
Looks like CI failed, maybe rerun it?

hoodmane:
Cherry-picked #4721 to this.

PR #4718, opened 2024-04-26
Distinguish between emscripten settings and instantiated emscripten module

hoodmane:
In Emscripten v3.1.58, `createPyodideModule` returns a distinct object from its argument so if we confuse `EmscriptenSettings` with the instantiated module, we'll get problems. This fixes these problems.

I also added some more type declarations.

Split from #4715.


ryanking13 (approved):
It's a nice change! I always thought there was too much information crammed into the module.

PR #4712, opened 2024-04-22
Run `make -C packages update-all`

hoodmane:
Packages that I reverted:
* bitstring
* fiona
* gdal
* matplotlib
* numcodecs
* packaging
* pandas
* Pillow
* pillow_heif
* protobuf
* pyinstrument
* rebound
* reboundx
* scipy
* shapely
* spareqr
* river
* xgboost
* yt
* zarr

PR #4719, opened 2024-04-26
SciPy v1.13.0

agriyakhetarpal:
### Description

This PR bumps the in-tree SciPy version in the Pyodide distribution to v1.13.0. This is an in-progress change, as I look to fix some of the older patches locally, and therefore this PR is being opened for purposes of visibility only at this time.

cc: @rgommers, and the package maintainers @lesteve, @steppi as listed in the recipe.

#### To be done 

- [x] Fix patches and updates files that have been upstreamed
- [x] Look at changes required `scipy/_build_utils/src/wrap_g77_abi.c` – changed from `scipy/_build_utils/src/wrap_g77_abi.f` in SciPy v1.12.0
- [x] Adjust build scripts and compiler flags as necessary
- [x] Scout for changes to the public API between these two versions and update `imports:`
- [ ] Ensure all tests and imports pass

### Checklists

- [x] Add / update tests


agriyakhetarpal:
I see that SciPy is being reverted from v1.13.0 to v1.12.0 in #4712 – I'm not sure what that means, so I'll be happy to wait until further progress or discussion on that PR

hoodmane:
>  I'm not sure what that means, so I'll be happy to wait until further progress or discussion on that PR (my understanding is that whichever packages still pass the tests with newer versions can stay and will not be reverted?)

Yeah basically we do one big PR for all the easy packages and then have to handle the ones that cause trouble one at a time as we have time. Please go ahead with working on this PR. If there is a conflict with #4712 for some reason I can help resolve it.

lesteve:
@agriyakhetarpal just curious, what is the status of this? Since you mentioned "as as I look to fix some of the older patches locally" I am guessing that you are still working on updating the patches?

For some weird reasons, it looks like the full CI was not run (random guess maybe because this is a Draft PR, although seems unlikely), can you try pushing an empty commit to trigger the CI again? I was confused and thought this PR was green untill I unfolded the "All checks" and found out that there is only readthedocs and pre-commit ...

agriyakhetarpal:
> @agriyakhetarpal just curious, what is the status of this? Since you mentioned "as as I look to fix some of the older patches locally" I am guessing that you are still working on updating the patches?

Hi, @lesteve, thank you for the ping. I have just returned to this PR late last week, and I was going to post an update today about how this is going, so, yes, doing this and updating the patches here is well on my radar for this week. I'll be sure to request help wherever I don't understand something fully!

> For some weird reasons, it looks like the full CI was not run (random guess maybe because this is a Draft PR, although seems unlikely), can you try pushing an empty commit to trigger the CI again? I was confused and thought this PR was green untill I unfolded the "All checks" and found out that there are only readthedocs and pre-commit ...

Ah, I guess the full CI was not run because I added the `[skip ci]` label in the commit message in order to save resources for others. :) I shall trigger CI in subsequent commits or when I rebase.

lesteve:
Ah makes sense I missed the `[skip ci]` thing and thought "OK the CI seems green, so this should be good to merge"

hoodmane:
It's an unfortunate thing about Skip Ci. Maybe we should add a "ci ran" check. Of course we can merge changes that skip CI, but the Ci is flaky enough that we merge failing CI a lot.

agriyakhetarpal:
I rebased on top of the latest changes from the `main` branch and fixed up the patches (all that was to fix, _most likely_) – I will hope that CI passes (🤞), and I shall update the to-do list in the PR description accordingly

agriyakhetarpal:
I rebased on top of the latest changes from the `main` branch and fixed up the patches (all that was to fix, _most likely_) – I will hope that CI passes (🤞), and I shall update the to-do list in the PR description accordingly.

agriyakhetarpal:
I assume that the failures under `ci/circleci: test-core-safari` are unrelated, I will wait until the `ci/circleci: build-libraries` job passes.

agriyakhetarpal:
Oops, it looks like `scipy` is under the "The following packages are disabled" list: https://app.circleci.com/pipelines/github/pyodide/pyodide/7181/workflows/88fa2666-a007-427a-826d-cd207b3a4485/jobs/100486?invite=true#step-104-9703_80, so it's not being built. I'll look if there is a special commit message to build SciPy or the SciPy stack in general based on https://pyodide.org/en/stable/development/building-from-sources.html#partial-builds.

agriyakhetarpal:
Build logs: https://app.circleci.com/pipelines/github/pyodide/pyodide/7182/workflows/36dda168-fe41-4dcf-859d-278178308521/jobs/100512/parallel-runs/0/steps/0-104

Note to self: most of the errors look to be coming from Fortran code because the f2c sub-command failed; I have to look at syntax errors in a few files, remove warnings about a few compiler flags (probably replace -Wno-maybe-unintialized with -Wno-unintialized and look at how Emscripten sets runtime search paths because -rpath was reported to be unknown, etc.), and correctly identify the file leading to the cause of the error (I am currently browsing through the logs on a mobile device and that does not help) – it's most likely inside FITPACK.

agriyakhetarpal:
> It's an unfortunate thing about Skip Ci. Maybe we should add a "ci ran" check. Of course we can merge changes that skip CI, but the Ci is flaky enough that we merge failing CI a lot.

Btw, @hoodmane, if you could elaborate on this slightly, I would be happy to add this "ci ran" check in some shape or form (if the change would be non-trivial, that is, of course).

agriyakhetarpal:
The errors were in `scipy/interpolate/fitpack/fpchec.f` (sub-module 1076 out of 1464) and were coming from lines like these:

```f90
if(t(i).le.t(i-1))then; ier=30; go to 80; endif
...
if(x(1).lt.t(k1) .or. x(m).gt.t(nk2))then; ier=40; go to 80;
endif
...
if(i.ge.m)then; ier=50; go to 80; endif
```

and so on. I checked the file versus SciPy v1.12 and [scipy/scipy@`a6284f6`.patch](https://github.com/scipy/scipy/commit/a6284f6cf7ac40865dc9810ea840f7b0164052e8.patch) is where this comes from, so I switched the if-else constructs to multiple lines in 6917e3348ab06ca2007b74b8f1a9ff5038b87fc4, and that makes them work for some reason.

and now SciPy v1.13.0 builds successfully (🎉): [workflow logs](https://app.circleci.com/pipelines/github/pyodide/pyodide/7184/workflows/ac105ea0-9db3-46ae-813e-0be839c8d48b/jobs/100552?invite=true#step-104-14978_80). I'll wait for the build to finish, following which I can add a CHANGELOG entry when this will be merged (please let me know if I should skip CI for that one) and complete the rest of the checklists.

agriyakhetarpal:
Never mind, there are 4 failures out of 5 tests on SciPy: `ImportError: dynamic module does not define module export function (PyInit_cython_lapack)`... I am diving further into this.

agriyakhetarpal:
I don't entirely know how to debug this and I am afraid that my attempts to look further did not lead to much insight. As far as I understand, not being experienced with Fortran codes, there could be a missing `PyInit_*` statement in the `scipy.linalg` side of things, or something was missed during the compilation of `cython_lapack` inside `scipy/linalg/meson.build` – I notice that 1.12.0 used `link_with: fwrappers` and 1.13.0 uses `link_with: [g77_abi_wrappers]`, and I tried to follow the discussions around that and I think the modified patch 5/8 should fix that in `scipy/_build_utils/src/wrap_g77_abi.c`. I inspected the auto-generated `cython_lapack.c` file diffs between v1.12.0 and v1.13.0 and I didn't catch anything relevant, except for the `F_FUNC` macros having been replaced by `BLAS_FUNC` everywhere, likely a result of the ABI wrappers.

Happy to receive suggestions on how I can debug this further and work towards a solution, @ryanking13 and @hoodmane – your comments will be highly appreciated!

hoodmane:
I'll take a look when I have a minute. At pycon right now.

rgommers:
If it helps, https://github.com/scipy/scipy/pull/20232 should be the relevant SciPy PR I believe. It replaced some of the problematic Fortran wrappers for ABI differences with C wrappers.

ryanking13:
Thanks for working on this! I will take a look after the PyCon (probably after May 23)

agriyakhetarpal:
@rgommers, I had a question that is extrinsic to these changes but probably still relevant – is there a policy on upstreaming particular non-intrusive patches to SciPy? I haven't followed the discussion in https://github.com/scipy/scipy/issues/15290 till the end but based on a quick look, there is a history of some having been upstreamed – in this case, patch 9/9 is a very small one and I would like to propose fixing it upstream after this PR here is complete/merged; it would (should) not break any existing changes and the syntax is probably more compatible across these different variants for Fortran compilers.

rgommers:
> is there a policy on upstreaming particular non-intrusive patches to SciPy

That all depends on whether the changes improve the SciPy code, or whether it's a workaround that's very specific to Pyodide. In this case, the original Fortran code seems correct and easier to read than the patched version - and this looks like an `f2c` issue, so I'd suggest not to upstream.

hoodmane:
The problem seems to be that `scipy/linalg/cython_lapack.cpython-312-wasm32-emscripten.so` fails to load (which causes other libs to fail to load which eventually leads to the missing `PyInit_` error). The loader error for cython_lapack is:
```
WebAssembly.instantiate(): Import #67 module="env" function="cbbcsd_": imported function does not match the expected type
```
which is frustratingly only the first of potentially many problems and also pretty darn vague. Checking the type of the symbol that it is getting (which presumably hasn't changed):
```js
pyodide._module.resolveGlobalSymbol("cbbcsd_").sym.type()
```
we see that it takes 29 i32 params and returns one i32 result.
```
$ wasm-objdump linalg/cython_lapack.cpython-312-wasm32-emscripten.so -j Import -x | grep cbbcsd
 - func[67] sig=30 <env.cbbcsd_> <- env.cbbcsd_
```
shows that `sig=30`. Checking sig 30 shows:
```
$ wasm-objdump linalg/cython_lapack.cpython-312-wasm32-emscripten.so -j Type -x | grep 30
 - type[30] (i32, i32, i32, i32, i32, i32, i32, i32, i32, i32, i32, i32, i32, i32, i32, i32, i32, i32, i32, i32, i32, i32, i32, i32, i32, i32, i32, i32, i32) -> nil
```
so the problem is the void return type. Of course.

hoodmane:
Looks like `_lapack_subroutines.h` has 1416 different functions that are declared as returning void. Those will all need to be changed to int. The only place I find `cbbcsd` in the source is in `cython_lapack_signatures.txt` so that's probably the source of truth here. When this was going through fortran it was getting fixed by `scipy_fix_cfile` but now that it isn't we ideally should patch the sources.

hoodmane:
Okay progress, now problem is `chla_transtype_`. It is defined with type `(i32, i32, i32) -> nil` but declared with type `(i32) -> i32`. This has a comment in f2c_fixes:
```
    chla_transtype is a special case. The OpenBLAS version of chla_transtype takes
    a return argument, whereas f2c thinks it should return the value.
```

hoodmane:
Where did all the build artifacts go? Meson seems to hide them better than the old build system did.

agriyakhetarpal:
I am not sure what happens under cross-compilation, but the build artifacts typically show up inside the a `build-install` directory in the source tree.

agriyakhetarpal:
I am not sure what happens under cross-compilation, but the build artifacts typically show up inside a `build-install` directory in the source tree.

hoodmane:
Hmm don't see any `build-install` directory. Could it be deleting them when its done or something?

rgommers:
> Where did all the build artifacts go? Meson seems to hide them better than the old build system did.

By default they go into a temporary directory if you use a plain `python -m build` (or `pip`), but you can add `-Cbuild-dir=build` and then all build artifacts are going under `./build/`. More detailed logs are in `build/meson-log/` and build artifacts are in corresponding subdirs of the source dirs (e.g., `build/scipy/linalg/`).

hoodmane:
We should just add `-Cbuild-dir=build` always.

agriyakhetarpal:
> We should just add -Cbuild-dir=build always.

@hoodmane, where would you recommend I add this? I tried to follow along similar recipes but couldn't find any, and then I looked at how this is done for the Pyodide CLI in `cli/build.py` which seemed simple enough – `pyodide build -C key=value`. However, how do I add this for SciPy where CircleCI builds it out of its recipe (and add it just for SciPy, of course)?

hoodmane:
Add it to `backend_flags` in the `build` section of the `meta.yaml` file.

agriyakhetarpal:
Okay, it failed to build SciPy but this is fixable, based on https://mesonbuild.com/meson-python/how-to-guides/config-settings.html#how-to-guides-config-settings. Let me push a change to fix this and analyse which other modules fail to import after the changes to the `cython_lapack_signatures.txt` failed and `chla_transtype_` is to be fixed.

agriyakhetarpal:
I think I have tried all combinations in `backend-flags:` (I had previously tried `backend_flags:` but Pydantic reported an error because it could not validate that schema) under `build:` in `scipy/meta.yaml`:

- `-Cbuild-dir=build`
- `--config-setting build-dir=build`
- `-C build-dir=build`
-` -Csetup-args="build-dir=build"`

and none of them seems to work with `meson-python` since it reports an unknown option in all of them (the fourth one wouldn't have worked, of course). Am I doing something wrong here?

rgommers:
The correct invocation should be:
```yaml
 backend-flags: build-dir=build
```

Searching other build recipes in this repo for examples is usually helpful. The numpy package has a usage of `backend-flags` that I based the above off of.

agriyakhetarpal:
Ah, there we go, thanks, @rgommers! I did do a search for this throughout the recipes before, and the reason I did not find anything was because I searched for `backend_flags` and not `backend-flags`, and I was looking for the `-C` option in specific (while I could have looked for `build-dir` as well, oops).

I feel this could be documented somewhere for use by others. There is a mention in the CHANGELOG, but it's not highlighted for package authors and maintainers on how to use `backend-flags` – maybe it's because not a lot of packages need it, especially not those that use `setuptools`, because `setuptools` doesn't support config settings properly at the time of writing (and it doesn't document those well either where they work).

hoodmane:
Documentation PRs are always very welcome. Our docs are very incomplete, and it would be nice for them to be less incomplete.

lesteve:
Just curious, was there any blocker for this? Is it more that it fell down the list of priorities?

I guess it would be nice to have scipy 1.13 (maybe 1.14) for Pyodide 0.26.2?
Not really sure what the timeline is for Pyodide 0.26.2 ...

agriyakhetarpal:
> Just curious, was there any blocker for this? Is it more that it fell down the list of priorities?

Thanks, @lesteve – it did fall down in my list of priorities in the past couple of weeks, and also because the compilation passes but failures on import are making me require more time to look into this – no blockers, though. Let me revisit this either within the next few days or as soon as July starts. In the meantime, any help that you can provide here is appreciated, of course.

lesteve:
> I was hoping to release 0.26.2 in a few days, it seems like there's quite a few commits going into it already.

Just for explicitness, I don't think this should hold up Pyodide 0.26.2 release. Personally, I am unlikely to have too much time to take a closer look in the short-term future.

I guess one of the first thing to do would be to fix conflicts?

It also seems the kind of issue that is quite time-consuming to fix, i.e. you need multiple trial and errors to get to the bottom of this (int vs void returns, scipy code generation with f2py, manipulating patches and/or getting this into `_f2c_fixes`)

agriyakhetarpal:
I always seem to have some trouble with the Docker images – first, faced some obscure JavaScript errors such as `Expected mapIndex to be 5, got 6` with no trace (fixed later with `esbuild`), and then I had issues getting `pytest` to be recognised when running the test suite. I couldn't trigger `scipy/test_scipy.py::cmdline_runner` even though I had Emscripten installed, so I'll push changes here for now and see what CI has to say.

The linalg tests that were earlier vexing us are now passing after https://github.com/scipy/scipy/pull/20558 and https://github.com/pyodide/pyodide/pull/4920, though, at least locally, and SciPy doesn't fail on import anymore:

```
scipy/test_scipy.py::scipy_linalg[node] PASSED                                        [ 11%]
scipy/test_scipy.py::brentq[node] PASSED                                              [ 22%]
scipy/test_scipy.py::dlamch[node] PASSED                                              [ 33%]
scipy/test_scipy.py::binom_ppf[node] PASSED                                           [ 44%]
```

agriyakhetarpal:
I always seem to have some trouble with the Docker images – first, faced some obscure JavaScript errors such as `Expected mapIndex to be 5, got 6` with no trace (fixed later with `esbuild`), and then I had issues getting `pytest` to be recognised when running the test suite. I couldn't trigger `scipy/test_scipy.py::cmdline_runner` even though I had Emscripten installed, so I'll push changes here for now and see what CI has to say.

The linalg tests that were earlier vexing us are now passing after https://github.com/scipy/scipy/pull/20558 and https://github.com/pyodide/pyodide/pull/4920, though, at least locally, and SciPy doesn't fail on import anymore:

```
scipy/test_scipy.py::scipy_linalg[node] PASSED                                        [ 11%]
scipy/test_scipy.py::brentq[node] PASSED                                              [ 22%]
scipy/test_scipy.py::dlamch[node] PASSED                                              [ 33%]
scipy/test_scipy.py::binom_ppf[node] PASSED                                           [ 44%]
```

Also, I added myself as a maintainer in/for the recipe, I hope that's alright!

agriyakhetarpal:
We don't have an error on import now but a function signature mismatch (or two) – so, progress: https://github.com/pyodide/pyodide/actions/runs/10170916568/job/28131809090. I guess I'll try to follow the instructions we have in the [Debugging tips](https://pyodide.org/en/stable/development/debugging.html#debugging-runtimeerror-function-signature-mismatch ) page and see if there's something I can make sense of. I have been putting some time into this recently and intend to continue to do so until completion, any help here is appreciated.

hoodmane:
It's interesting, I get a loader error if I don't remove `chla_transtype`:
```
--- a/scipy/linalg/cython_lapack_signatures.txt
+++ b/scipy/linalg/cython_lapack_signatures.txt
@@ -108,7 +108,6 @@ void chetrs(char *uplo, int *n, int *nrhs, c *a, int *lda, int *ipiv, c *b, int
 void chetrs2(char *uplo, int *n, int *nrhs, c *a, int *lda, int *ipiv, c *b, int *ldb, c *work, int *info)
 void chfrk(char *transr, char *uplo, char *trans, int *n, int *k, s *alpha, c *a, int *lda, s *beta, c *c)
 void chgeqz(char *job, char *compq, char *compz, int *n, int *ilo, int *ihi, c *h, int *ldh, c *t, int *ldt, c *alpha, c *beta, c *q, int *ldq, c *z, int *ldz, c *work, int *lwork, s *rwork, int *info)
-char chla_transtype(int *trans)
 void chpcon(char *uplo, int *n, c *ap, int *ipiv, s *anorm, s *rcond, c *work, int *info)
 void chpev(char *jobz, char *uplo, int *n, c *ap, s *w, c *z, int *ldz, c *work, s *rwork, int *info)
-- 
```

I think the correct signature is `int chla_transtype(char*, int*);

hoodmane:
@agriyakhetarpal how do you build scipy with symbols? There's another fatal error somewhere in arpack somewhere but the functions on the traceback are unnamed.

lesteve:
About building scipy in debug mode, I think something like this is enough:
```
EXTRA_CFLAGS=-g2 EXTRA_LDFLAGS=-g2 PYODIDE_PACKAGES="scipy" make
```

agriyakhetarpal:
Yes, and with the config settings, `-Csetup-args=-Dbuildtype=debug` should also work.

agriyakhetarpal:
I don't think the config settings helped, since the output from the logs is the same, just with different memory addresses. The function signature mismatch is in PROPACK. I did find https://github.com/pyodide/pyodide/pull/3381, which is probably not related. The command `EXTRA_CFLAGS=-g2 EXTRA_LDFLAGS=-g2 PYODIDE_PACKAGES="scipy" make` fails with unrelated errors due to the `dist/pyodide.d.ts` failing, but SciPy does compile.

agriyakhetarpal:
Another error with `_dop`: `ImportError: dynamic module does not define module export function (PyInit_
_dop)`

agriyakhetarpal:
and the test from `cmdline_test_file.py`, i.e.,

```python
import numpy as np
from scipy.sparse.linalg import svds

rng = np.random.default_rng(0)
A = rng.random((10, 10))

res = svds(A, k=3, which="LM", random_state=0)
print("res", res)
```

gives

```
Error: Dynamic linking error: cannot resolve symbol s_cmp
```

which has been a familiar symbol visibility problem

ryanking13:
> Error: Dynamic linking error: cannot resolve symbol s_cmp
which has been a familiar symbol visibility problem

This seems to be the same roblem I had here (https://github.com/pyodide/pyodide/pull/4876#issuecomment-2254403007).

`s_cmp` symbol is used in the following modules:

```
scipy/sparse/linalg/_eigen/arpack/_arpack.cpython-312-wasm32-emscripten.so: s_cmp
scipy/optimize/_minpack2.cpython-312-wasm32-emscripten.so: s_cmp
scipy/optimize/_lbfgsb.cpython-312-wasm32-emscripten.so: s_cmp
```

and among them, `_minpack2` is not linked to openblas.so

```
$ pyodide auditwheel show scipy-1.13.0-cp312-cp312-pyodide_2024_0_wasm32.whl | grep -e _arpack -e _minpack2 -e _lbfgsb
│   'scipy/sparse/linalg/_eigen/arpack/_arpack.cpython-312-wasm32-emscripten.so': ['libopenblas.so'],
│   'scipy/optimize/_minpack2.cpython-312-wasm32-emscripten.so': [],
│   'scipy/optimize/_lbfgsb.cpython-312-wasm32-emscripten.so': ['libopenblas.so'],
```

So I think you can update [this patch](https://github.com/pyodide/pyodide/blob/main/packages/scipy/patches/0011-Link-openblas-with-modules-that-require-f2c.patch) to link openblas to [_minpack2](https://github.com/scipy/scipy/blob/7dcd8c59933524986923cde8e9126f5fc2e6b30b/scipy/optimize/meson.build#L153)

agriyakhetarpal:
The `scipy.sparse.linalg.svds` error is indeed fixed by indicating dependence on OpenBLAS for `_minpack2`.

hoodmane:
> scipy.sparse.linalg.svds error is indeed fixed 

It looks to me like it's still failing?

agriyakhetarpal:
Oh, no – the error on import is fixed, not the failing FSM in the test-scipy job. I'm looking into how I can compile with symbol information. I think one of the patches is broken due to multiple `c_args:`, which I shall fix (strangely it doesn't appear until I add `-g2` to `cflags:`).

agriyakhetarpal:
I see PROPACK has some problems with the C++ standard when either `-g2/-g3` are set in `cflags:`/`cxxflags:`/`ldflags:`. Weirdly, it doesn't occur if the flags are not set.

```
../scipy/sparse/linalg/_propack/meson.build:67:29: ERROR: C++ Compiler does not support -std=c++17
```

agriyakhetarpal:
`EXTRA_CFLAGS=-g3 EXTRA_LDFLAGS=-g3` makes `lapack.dlamch` fail with another function signature mismatch, and the value in the stack is quite large for the WASM table: 9875559, which means it isn't easy to get what we need:

```
VM14370:1 Uncaught 
RangeError: WebAssembly.Table.get(): invalid index 9875559 into funcref table of size 38028
    at Object.eval (eval at <anonymous> (unknown source), <anonymous>:1:27)
    at 00603e66:0xc0029
    at 0004b7ea:0xd0c4
    at _PyEM_TrampolineCall_JS (pyodide.asm.js:10:124568)
    at pyodide.asm.wasm:0x1c2f5e
    at pyodide.asm.wasm:0x1c376f
    at pyodide.asm.wasm:0x2a0769
    at pyodide.asm.wasm:0x296247
    at pyodide.asm.wasm:0x293a9f
    at pyodide.asm.wasm:0x29f613
```

and the C source maps didn't help, since I'm in a Docker image.

hoodmane:
> 9875559

This is memory corruption, if we were using a more traditional target it probably would have segfaulted quite a bit ahead of this.

agriyakhetarpal:
I see. Not sure if it helps, but I found the `dlmach` failure through another test, not directly – that test couldn't load the `dlmach` symbol which prompted me to dig further. We do have a direct test for this behaviour, but that passes with `pytest`. I'll also need to take a look at the CircleCI test failures – it says `triu` can't be imported from `scipy.linalg`.

agriyakhetarpal:
Never mind – `triu` was removed from `scipy.linalg` in SciPy v1.13.0, which explains the error. The failure is from `gensim`, which I can look to update after the issues on SciPy are resolved.

hoodmane:
> Cython<3

You would think after running into this issue one time per year every year people would stop including cythonized code in their sdist. Anyone able to compile or run the code is going to have no trouble getting cython, so it doesn't make much sense.

agriyakhetarpal:
> > Cython<3
> 
> You would think after running into this issue one time per year every year people would stop including cythonized code in their sdist. Anyone able to compile or run the code is going to have no trouble getting cython, so it doesn't make much sense.

Yes, it doesn't make sense. However, they do mention that the Cython<3 pin is kept for performance reasons because Cython 3 makes their code an order of magnitude slower. Should we patch it and use Cython 3, since speed might not be of the essence as much in WASM?

hoodmane:
I think what we have is fine.

rgommers:
For PROPACK we fixed an ABI issue in `main` recently which caused MKL to crash as well. PROPACK is a bit of a mess to be honest. Since it was hairy to fix in SciPy itself and is fixed now, I doubt it's worth spending a lot more time on it here for 1.13.0. How about adjusting the test invocation or add a patch to skip all PROPACK tests? If everything else works, I'd merge this PR.

ryanking13:
That's okay with me. I think we need to skip some tests [here](https://github.com/pyodide/pyodide/blob/38d7b62fcc74af1979b4b13686f2ea241909b97e/packages/scipy/scipy-conftest.py#L18).

The CI failed in following file:

> sparse/linalg/_eigen/tests/test_svds.py::Test_SVDS_PROPACK

agriyakhetarpal:
The logs look like they are a bit too long to parse in the GitHub Actions UI, so I'm looking at them locally – the only failure I am able to find in the workflow run is `linalg/tests/test_decomp.py::test_svd_gesdd_nofegfault FAILED            [ 19%]`, and some other tests that fail because `multiprocessing` isn't available (or probably that those are the already xfailed tests, but that the logs are way too large). However, the test suite now runs until completion, so that's great. I just pushed 0504d338a1590a2d95b487d19287ff718cf8da37 to temporarily turn down the verbosity meter a bit.

agriyakhetarpal:
and that test looks like it has been fixed with https://github.com/scipy/scipy/pull/20349, so I should just need to patch it.

agriyakhetarpal:
Reading through the ILP64/LP64/LAPACK integer overflow discussions in the issue (and other related issues), I notice that the problem with that test seems to be that it is just out of memory, because the array is quite large to handle.

`numpy.core._exceptions._ArrayMemoryError: Unable to allocate 1.90 GiB for an array with shape (4799, 53130) and data type float64`

and this shouldn't fail with a memory error in a larger CI runner or a regular machine, perhaps. @rgommers, should we skip this test – and perhaps introduce a new message variable named `todo_out_of_memory_msg` (or similar) in `scipy-conftest.py`?

agriyakhetarpal:
Not that `sys.getsizeof()` is the best way to get true sizes for Python objects (or NumPy arrays even), but a float64 array with 4799x53130 dimensions does return something on the lines of 2039767088 even locally, which somewhat matches the error description here, so I guess it isn't a problem specific to WASM.

ryanking13:
> Edit: we are down to one failure now, as noted below!

Yay! Thanks a lot for your huge work Agriya!

> and this shouldn't fail with a memory error in a larger CI runner or a regular machine, perhaps.

I guess it can be an issue in WASM. The maximum memory size we can use is 4GB (as we are in 32bit space), so there were a few issues with allocating large array.

rgommers:
Great news! Re the last test, I suggest just skipping it upstream with a `skip(IS_WASM` and including that change as a patch here. That test isn't useful to Pyodide, and the code is trying to find the exact limits of when things outside of SciPy (in any other BLAS library) will overflow - that is inherently not robust.

rgommers:
Great news! Re the last test, I suggest just skipping it upstream with a `skip(IS_WASM` and including that change as a patch here. That test isn't useful to Pyodide, and the code is trying to find the exact limits of when things outside of SciPy (in any other BLAS library) will overflow - that is inherently not robust.

rgommers:
> @rgommers, since we don't have an out-of-tree build for SciPy yet, would you suggest going upstream or skipping it here? I'm of the impression that we could fix things upstream with more changes than fixing just one failing test: open to suggestions.

Every patch that makes sense to upstream (which should include most/all test skips) is welcome to be upstreamed for SciPy, so after adding the patch here to make this PR green it makes sense to me up upstream it so that when SciPy 1.15.0 comes around the patch here can be dropped.

It doesn't really matter that SciPy doesn't yet have an out-of-tree build, we carry lots of fixes for things like niche CPU architectures that Debian supports or for AIX, none of which we test in CI.

packages/scipy/meta.yaml
```diff
@@ -46,7 +50,9 @@ build:
     -L$(NUMPY_LIB)/core/lib/
     -L$(NUMPY_LIB)/random/lib/
     -fexceptions
-  backend-flags: build-dir=build
+  backend-flags: |
+    build-dir=build
+    setup-args=-Dbuildtype=debug
```
agriyakhetarpal, line 55: Commenting so that I remember to remove this

packages/scipy/patches/0005-Remove-test-modules-that-fails-to-build.patch
```diff
@@ -1,32 +1,34 @@
...
 From: Gyeongjae Choi <def6488@gmail.com>
 Date: Mon, 30 Oct 2023 14:35:04 +0000
-Subject: [PATCH 5/9] Remove test modules that fails to build
+Subject: [PATCH 5/15] Remove test modules that fails to build
 
 These are tests and they have both void vs int return value problems and implicit
 function argument cast problems. Not worth fixing for tests.
+
+Co-Developed-by: Gyeongjae Choi <def6488@gmail.com>
+Co-Authored-By: Gyeongjae Choi <def6488@gmail.com>
```
agriyakhetarpal, line 10:
```suggestion
```

packages/scipy/meta.yaml
agriyakhetarpal:
```suggestion
```

agriyakhetarpal:
All tests passing after relevant ones were skipped or removed: https://github.com/pyodide/pyodide/actions/runs/10320686508/job/28573393091

agriyakhetarpal:
The CircleCI test failures aren't related, I think, because they show that they timed out. I don't have permissions to re-run the CircleCI runners. Could someone trigger a re-run for them?

ryanking13:
> The CircleCI test failures aren't related, I think, because they show that they timed out. I don't have permissions to re-run the CircleCI runners. Could someone trigger a re-run for them?

Hmm, that's strange, CircleCI roles and permission settings are inherited from GitHub, so you should be able to run it. I sent you an circle CI invite link just in case.

ryanking13 (approved):
Thanks a lot @agriyakhetarpal! Let me go ahead and merge this. If the test timeout keep occurs, please increase the timeout.

agriyakhetarpal:
I see that the test_SVDS_PROPACK test that we skip for GHA `test-scipy` job is running on CircleCI, which causes the function signature mismatch

agriyakhetarpal:
(maybe we merged too early)

PR #4721, opened 2024-04-26
Remove possible query string from tarballname

hoodmane:
For CoolProp, we are getting `tarballname` as `'CoolProp_sources.zip?viasf=1'` which then crashes when we give it to `shutil.unpack_archive` with `Unknown archive format`.


ryanking13:
Thanks for fixing this!

PR #4722, opened 2024-04-27
Fix url filename parsing again

ryanking13:
### Description

Trying to fix #4721 again, using the URL inside meta.yaml instead of using the response URL (which may contain query strings).

Additionally, removes `cgi` module usage, which will be removed in python3.13

### DIFF
--- a/packages/ppl/meta.yaml
+++ b/packages/ppl/meta.yaml
@@ -1,11 +1,11 @@
 package:
   name: ppl
-  version: 1.2
+  version: "1.2"
   tag:
     - library
 source:
-  url: https://www.bugseng.com/products/ppl/download/ftp/releases/1.2/ppl-1.2.tar.xz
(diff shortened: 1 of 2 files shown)


packages/ppl/meta.yaml
```diff
@@ -1,11 +1,11 @@
 package:
   name: ppl
-  version: 1.2
+  version: "1.2"
   tag:
     - library
 source:
-  url: https://www.bugseng.com/products/ppl/download/ftp/releases/1.2/ppl-1.2.tar.xz
-  sha256: 691f0d5a4fb0e206f4e132fc9132c71d6e33cdda168470d40ac3cf62340e9a60
+  url: https://mirrors.mit.edu/sage/spkg/upstream/ppl/ppl-1.2.tar.bz2
```
ryanking13, line 7: ppl seems to deny user-agent of requests, hence replacing it with MIT mirror.

hoodmane (approved):
Thanks!

ryanking13:
Thanks for the review!

PR #4723, opened 2024-04-27
Update `pyodide create-zipfile` command to accept list of modules to exclude

ryanking13:
### Description

This changes `pyodide create-zipfile` command which is used to generate `python_stdlib.zip` to accept list of modules to exclude from commandline argument. Previously, they were hard-coded inside Python code.

This change loosens the dependency between the pyodide-build and the pyodide main module.


hoodmane (approved):
Thanks @ryanking13! Really appreciate your work on pyodide-build.

PR #4726, opened 2024-04-30
NFC: Clean up emscripten settings a bit

hoodmane:
I renamed module.ts to emscriptenSettings.ts which is more descriptive. Rather than setting a bunch of the settings fields from pyodide.ts, set them in createSettings. I also marked all these fields as readonly so typescript can help us ensure that they are set up correctly in createSettings and not edited from pyodide.ts.

I changed most of the functions in `emscriptenSettings.ts` to return a `PreRunFunc` rather than taking the settings as an argument and mutating it.

I also removed the `moduleLoaded` `postRun` hook, I believe that it hasn't been needed for a long time.


src/js/api.ts
```diff
@@ -1,4 +1,4 @@
-import "./module";
+import "./emscriptenSettings";
```
hoodmane, line 1: What's the point of this import??

src/js/emscriptenSettings.ts
ryanking13, line 1: I think we are using kebab case for js filenames, so maybe lets rename it to `emscripten-settings.ts`?

ryanking13 (approved):
Thanks!

PR #4732, opened 2024-05-01
Use Pyodide version instead of pyodide-build version when creating lockfile

ryanking13:
### Description

Updates lockfile creation part to use Pyodide version instead of pyodide-build version. For now, they are the same but we are planning to separate it.

To make Pyodide version available in pyodide-build, I put the Pyodide version into the makefile.envs.

### Checklists

- [x] Add / update tests


ryanking13:
Thanks for the review!

PR #4735, opened 2024-05-02
Add missing library dependencies to packages

ryanking13:
### Description

Minor recipe updates that I had issues with when building these packages out-of-tree. These packages rely on emscripten library ports, but it was not specified explicitly.


hoodmane (approved):
Thanks!

PR #4741, opened 2024-05-05
TST Improve xbuildenv and emscripten mocks

ryanking13:
### Description

The goal of this PR is to reduce the dependency on pyodide for testing pyodide-build. The main changes are:

1. Add a new fixture: `mock_emscripten` for mocking emscripten compiler toolchain. This fixture will create a dummy emcc (+ other toolchains) to the PATH. When running tests, all emcc calls will be intercepted by this command.

2. Improve `xbuildenv` fixture. Previously, this fixture actually created an xbuildenv from the Pyodide artifacts. So, we had to have Pyodide already built before running the tests. In this PR, this fixture now uses a pre-built minimum xbuildenv.

As a result, now all tests in `pyodide-build` run without building Pyodide. The only exception is `test_xbuildenv_create` which test `pyodide xbuildenv create` command. In a follow-up, I am planning to migrate the `provide venv` tests, which are in `test_cmdline_runner.py.`

### Checklists

- [x] Add / update tests


hoodmane (approved):
Thanks @ryanking13!

PR #4742, opened 2024-05-06
Update docs to use latest sphinx-js version

hoodmane:
This pulls in the tip of the master branch of sphinx-js and shortens the
jsdoc.py patches by about 60 lines.

I also fixed some broken links and various typedoc and sphinx warnings.
I turned on nitpicky mode and silenced intentionally broken links so in the
future we can catch broken xrefs better. These warnings also help turn up
cases where private fields are accidentally included into the docs.


ryanking13 (approved):
Thank you so much for your efforts on improving docs!

PR #4744, opened 2024-05-07
Update sphinx-js again and more edits

hoodmane:
This pulls in the newest commits of sphinx-js that can render type aliases and interfaces into the docs. I used it to add documentation for `PackageType` and `PackageData`. It also adds type params in the signature line of js functions, which I'm actually not sure is that great since the signature line otherwise has no types.

image

and 

image

### DIFF
--- a/src/js/api.ts
+++ b/src/js/api.ts
@@ -40,7 +40,7 @@ API.setPyProxyToStringMethod = function (useRepr: boolean): void {
   Module.HEAP8[Module._compat_to_string_repr] = +useRepr;
 };
 
-/** @private */
+/** @hidden */
 export type NativeFS = {
   syncfs: () => Promise<void>;
 };
@@ -647,7 +647,7 @@ export class PyodideAPI {
   }
 }
 
-/** @hidetype */
+/** @hidden */
 export type PyodideInterface = typeof PyodideAPI;
 
 /** @private */
--- a/docs/requirements-doc.txt
+++ b/docs/requirements-doc.txt
@@ -18,4 +18,4 @@ pydantic
 micropip==0.2.2
 jinja2>=3.0
 ruamel.yaml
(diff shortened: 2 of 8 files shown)


src/js/api.ts
```diff
@@ -40,7 +40,7 @@ API.setPyProxyToStringMethod = function (useRepr: boolean): void {
   Module.HEAP8[Module._compat_to_string_repr] = +useRepr;
 };
 
-/** @private */
+/** @hidden */
```
ryanking13, line 43: Are we replacing `@private` decorator to `@hidden` at all? Or are there some difference between the two?

ryanking13 (approved):
Thanks!

src/js/api.ts
hoodmane: `@hidden` makes typedoc drop the object from the ir:
https://github.com/TypeStrong/typedoc/blob/master/src/lib/converter/plugins/CommentPlugin.ts#L568

`@private` gets stored into `relection.flags.isPrivate` and gets excluded if the `--excludePrivate` argument is passed to typedoc:
https://github.com/TypeStrong/typedoc/blob/master/src/lib/converter/plugins/CommentPlugin.ts#L500-L505

We don't pass `--excludePrivate`. So instead we decide whether it should be private here by looking at the Python `is_private` field on the TopLevel:
https://github.com/pyodide/pyodide/blob/main/docs/sphinx_pyodide/sphinx_pyodide/jsdoc.py#L87-L90

On TopLevels that can also be "class members" we copy the `isPrivate` flag into the `is_private` ir field. We don't distinguish between a TopLevel function and a class Method, nor a TopLevel variable export and a class attribute. So variables, attributes, functions, and methods are all hidden by `@private` but `TypeAlias`, `Interface`, and `Class` are not.
https://github.com/pyodide/sphinx-js-fork/blob/master/sphinx_js/ir.py?plain=1#L143

Needless to say this requires some cleanup.

hoodmane: I'll try to fix this in a followup.

PR #4752, opened 2024-05-10
Remove deploy dependencies from pyodide-build's pyproject.toml

ryanking13:
### Description

This removes `[deploy]` dependencies from pyodide-build, and puts them under the `requirements-deploy.txt` in the root directory.

If these packages are defined in pyodide-build, they create a dependency that requires pyodide-build to be present in order to deploy Pyodide, so this PR removes them from pyodide-build.


hoodmane:
> If these packages are defined in pyodide-build, they create a dependency that requires pyodide-build to be present in order to deploy Pyodide, so this PR removes them from pyodide-build.

I'm not sure what this means, can you explain a bit further?

requirements.txt
```diff
@@ -1,6 +1,4 @@
-# build
-cmake
-ruamel.yaml
```
hoodmane, line 3: Did these move somewhere else?

hoodmane (approved):
Thanks @ryanking13!

hoodmane:
```
FAILED test_console.py::interactive_console - ModuleNotFoundError: No module named 'pytz'
```

ryanking13:
> I'm not sure what this means, can you explain a bit further?

boto3 and moto is used to deploy build artifacts to s3, including pyodide.js, pyodide.asm.js, etc. I don't think these need to be a dependency of pyodide-build, as they have nothing to do with "building packages". So I think it is strange to have these packages declared in pyodide-build.

Issue #4753, opened 2024-05-10
`make` in docker container `pyodide/pyodide-env` fails with `could not create work tree dir 'emsdk'`


goulashsoup:
## 🐛 Bug

`make` not working in docker container `pyodide/pyodide-env`:

```
@e8b479cd0979:/src$ make
[2024-05-10 20:55:07] Building emsdk...
make -C emsdk
make[1]: Entering directory '/src/emsdk'
if [ -d emsdk ]; then rm -rf emsdk; fi
git clone --depth 1 https://github.com/emscripten-core/emsdk.git
fatal: could not create work tree dir 'emsdk': Permission denied
make[1]: *** [Makefile:8: emsdk/.complete] Error 128
make[1]: Leaving directory '/src/emsdk'
make: *** [Makefile:313: emsdk/emsdk/.complete] Error 2
```

### To Reproduce

Probably prune docker images first, run `./run_docker` and then when the container is in interactive terminal, run `make` inside the container.

### Expected behavior

Build successful.

### Environment

- Pyodide Version: 0.25.1
- Browser version: -

- Commit hash of Pyodide git repository: 3d127ea24931d18b3a4710bcf43f81db2efebe67 and afe7215c06ec9cb5a4d145c4e28dd974038852bc
- Build environment: Ubuntu 23.10 LTS, Docker Desktop on Linux 4.30.0

hoodmane:
Maybe try `./run_docker --root`?

goulashsoup:
@hoodmane Thanks, that worked, but the docs don't state that I need root: https://pyodide.org/en/stable/development/building-from-sources.html

goulashsoup:
Added a PR for `--root` option in the docs.

goulashsoup:
@hoodmane If I should not use `--root` what permissions should I set before executing `./run_docker`.

goulashsoup:
> there is something specific to the way permissions are working on your machine that is breaking it without it.

It's a "normal" Ubuntu, no custom changes regarding permissions made actually...

I think it more likely that, when you checked in on a Linux, that you changed some permission once, and forgot about them writing the documentation.

hoodmane:
Well I develop primarily on linux (Ubuntu 22.04.3) and I think most of the main contributors use linux as well. Though I stopped using the docker container a while ago. I think linux is actually the only OS that is at least vaguely well supported. Windows works via WSL and Mac development sometimes is more broken and other times less broken.

[...]

PR #4754, opened 2024-05-10
Add required --root` to `./run_docker` build from source documentation

goulashsoup:
### Description

It seems `--root` option is required when running `make` in the docker container, see #4753

### Checklists

### DIFF
--- a/docs/development/building-from-sources.md
+++ b/docs/development/building-from-sources.md
@@ -43,6 +43,10 @@ These Docker images are also available from the Github packages at
 3. Run `make` to build.
 
 ```{note}
+If `make` is failing with `Permission denied`, you can run `./run_docker` with root permissions: `./run_docker --root`
+```
+
+```{note}
 You can control the resources allocated to the build by setting the env
 vars `EMSDK_NUM_CORE`, `EMCC_CORES` and `PYODIDE_JOBS` (the default for each is
 4).


hoodmane:
Well it's not required in general, there is something specific to the way permissions are working on your machine that is breaking it without it. I wouldn't really recommend this either because it has the annoying consequence that all the build artifacts are owned by root and you have to use sudo to remove them.

hoodmane:
But thanks for the PR @goulashsoup. How about instead adding the comment that if you see permissions problems when building, passing `--root` to `run_docker` can help?

goulashsoup:
> But thanks for the PR @goulashsoup. How about instead adding the comment that if you see permissions problems when building, passing --root to run_docker can help?

Can do that as well...

goulashsoup:
@hoodmane Changed it.

docs/development/building-from-sources.md
```diff
@@ -36,6 +36,10 @@ These Docker images are also available from the Github packages at
 [`github.com/orgs/pyodide/packages`](https://github.com/orgs/pyodide/packages).
 ```
 
+```{note}
+If `make` is failing with `Permission denied`, you can run `./run_docker` with root permissions: `./run_docker --root`
+```
+
```
hoodmane, line 42: Sounds good, can you move it down to line 48 after the steps?

goulashsoup: Done

hoodmane:
Thanks!

PR #4755, opened 2024-05-10
Rearrange changelog for 0.26.0

hoodmane:


docs/project/changelog.md
```diff
@@ -95,14 +86,38 @@ myst:
...
+  This fixes "Argument list too long" error.
+  {pr}`4717``
+
+- {{ Fix }} Pass through `-E` (command mode) arguments in CMake wrapper
+  {pr}`4705`.
+
+- {{ Fix }} Fix exception handling in dynamic linking of int64 functions
+  {pr}`4698`.
+
+- {{ Breaking }} `pyodide-build` entrypoint is removed in favor of `pyodide`.
```
henryiii, line 109: Should the package be renamed `pyodide`? Now you `pip install pyodide-build` to get `pyodide`, and the command I would use is `pipx run --spec pyodide-build pyodide ...`

henryiii: Also, is there a `pyodide` entry point to replace it? I don't see it in pyproject.toml?

hoodmane: I think we were concerned that people will think that `pip install pyodide` is getting the main Pyodide runtime, hence the warning about not being able to install Pyodide via pip. But we should at least tell people that they might want `pyodide-build` there.

PR #4756, opened 2024-05-10
JsProxy bindings: Call via Reflect.construct when signature is a class

hoodmane:
### Checklists

- [x] Add / update tests


ryanking13 (approved):
Thanks!

PR #4763, opened 2024-05-14
packages: Add lakers-python

chrysn:
### Description

This adds Lakers (`lakers-python` on PyPI), the Python wrappers for the Rust implementation of EDHOC (RFC9528). EDHOC is a compact key exchange protocol, primarily designed for embedded systems.

As a compiled package it can not use the pure Python wheels published through crates.io. It is useful in pyodide because aiocoap has [good support for pyodide and Jupyter](https://aiocoap.readthedocs.io/en/latest/pyodide.html) for end-to-end encryption between a browser console (or application) and an embedded device (connected through untrusted proxies on the Internet or through local Bluetooth connections). I expect to add EDHOC support to aiocoap through lakers.

### Checklists

- [x] Add / update tests
  Marked as checked because I don't think that adding a package needs any added documentation outside the change log.

### DIFF
--- a/docs/project/changelog.md
+++ b/docs/project/changelog.md
@@ -110,7 +110,8 @@ myst:
   `pyxirr` {pr}`4513`, `ipython`, `asttokens`, `executing`, `prompt_toolkit`,
   `pure_eval`, `stack_data`, `traitlets`, `wcwidth` {pr}`4452`, `altair` {pr}`4580`,
   `cvxpy` {pr}`4587`, `clarabel` {pr}`4587`, `matplotlib-inline` {pr}`4626`,
-  `pygame-ce` {pr}`4602`, `libcst` {pr}`4665`, `mmh3`, `pyiceberg` {pr}`4648`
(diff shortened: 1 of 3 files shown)


chrysn:
Downgraded to draft again due to https://github.com/openwsn-berkeley/lakers/issues/268 (that showed up in a CI step I lumped together with the unrelated build failures before); will upgrade and reopen once that is addressed.

chrysn:
By removing the needless `cargo update` on proc-macro2, the effect of the upstream issue can be avoided. While the published Python package still has an underlying oddity (the Cargo.toml referencing a workspace member that is not shipped), builds appear to pass now. I'll keep tracking the issue with upstream, but as things are, I'm restoring this to "ready for review".

hoodmane:
Thanks @chrysn. I think the versions of proc-macro2 from 1.0.61 to 1.0.68 don't work, but the newer versions are okay so you indeed shouldn't need to pin it at 1.0.60 anymore. Build failure looks unrelated and chrysn doesn't depend on numpy so it got built and tested in the no-numpy-dependents jobs. I guess I'll merge.

PR #4773, opened 2024-05-19
Add docs for generating a custom lock file

hoodmane:
cc @Alex-Monahan

### DIFF
--- a/docs/usage/lock-files.md
+++ b/docs/usage/lock-files.md
@@ -12,9 +12,12 @@ pyodide venv .venv-pyodide
 .venv-pyodide/bin/pip install fastapi
 .venv-pyodide/bin/python -c 'import micropip; print(micropip.freeze(), file=open("pyodide-lock.json","w"))'
 ```
+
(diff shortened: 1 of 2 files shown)


docs/usage/lock-files.md
```diff
@@ -0,0 +1,23 @@
+# Generating a custom lock file
+
+For applications that have static dependencies that are not part of the Pyodide
```
hoodmane, line 3:
```suggestion
For applications that have static dependencies that are not part of the default Pyodide
```

hoodmane:
Advice doesn't work right now due to: https://github.com/pyodide/micropip/issues/107 but hopefully soon.

ryanking13 (approved):
Thanks! I am not sure if we want to promote the use of `pyodide venv` in this way, but we can certainly improve docs when we find a better way.

hoodmane:
I think this should be a supported use case of `pyodide venv`: we provide no other way to get decent solves for the installations. But we need to actually make `micopip.freeze()` handle it somehow.

PR #4792, opened 2024-05-24
Add `zstandard`

### LINKED ISSUE #4788: `zstandard`
## 🐍 Package Request

- Package Name and Version : `zstandard`, version 0.22.0
- Package URL : https://pypi.org/project/zstandard/, https://github.com/indygreg/python-zstandard/
- Package Dependencies that needs to be resolved first: [none](https://github.com/indygreg/python-zstandard/blob/main/pyproject.toml) – this is a package written in C with Python bindings.

## 📝 Additional context

xref: https://github.com/zarr-developers/zarr-python/pull/1903, i.e., this is a required dependency of the upcoming version 3 for Zarr at the time of writing.

agriyakhetarpal:
### Description

This PR adds the [`zstandard`](https://pypi.org/project/zstandard/) Python package with the C and CFFI backends (the Rust backend is not compiled as of now, but can be a future improvement). `zstandard` is a required dependency in the upcoming version 3 release for Zarr. 

Closes #4788

### Checklists

- [x] Add / update tests

### Additional context

xref: https://github.com/zarr-developers/zarr-python/pull/1903

### DIFF
--- a/packages/zstandard/test_zstandard.py
+++ b/packages/zstandard/test_zstandard.py
@@ -7,9 +7,9 @@ from pytest_pyodide import run_in_pyodide
 def env(selenium):
     selenium.run_python(
         """
-      import os
-      os.environ["PYTHON_ZSTANDARD_IMPORT_POLICY"] = "default"
-      """
+        import os
+        os.environ["PYTHON_ZSTANDARD_IMPORT_POLICY"] = "default"
+        """
     )
     yield
 
--- a/packages/zstandard/test_zstandard.py
+++ b/packages/zstandard/test_zstandard.py
@@ -46,7 +46,7 @@ def test_compress_empty(selenium):
     params = zstd.get_frame_parameters(result)
     assert params.window_size == 1024
     assert params.dict_id == 0
-    assert params.has_checksum != 0
+    assert params.has_checksum is False
 
     cctx = zstd.ZstdCompressor()
     result = cctx.compress(b"")
@@ -73,10 +73,10 @@ def test_compress_large(selenium):
     # This matches the test for read_to_iter() below.
     cctx = zstd.ZstdCompressor(level=1, write_content_size=False)
     result = cctx.compress(b"f" * zstd.COMPRESSION_RECOMMENDED_INPUT_SIZE + b"o")
-    assert result == (
-        b"\x28\xb5\x2f\xfd\x00\x40\x54\x00\x00"
+    assert (
+        result == b"\x28\xb5\x2f\xfd\x00\x40\x54\x00\x00"
         b"\x10\x66\x66\x01\x00\xfb\xff\x39\xc0"
-        b"\x02\x09\x00\x00\x6f",
+        b"\x02\x09\x00\x00\x6f"
     )
 
 
--- a/packages/zstandard/test_zstandard.py
+++ b/packages/zstandard/test_zstandard.py
@@ -11,7 +11,7 @@ def set_zstd_backend(selenium, zstd_backend):
 
 
 @run_in_pyodide(packages=["zstandard"])
-@pytest.mark.fixture(params=["cext", "cffi"], autouse=True)
+@pytest.fixture(params=["cext", "cffi"], autouse=True)
 def zstd_backend(selenium, request):
     set_zstd_backend(selenium, request.param)
 
--- a/packages/zstandard/test_zstandard.py
+++ b/packages/zstandard/test_zstandard.py
@@ -1,9 +1,18 @@
-import os
+import pytest
+from pytest_pyodide import run_in_pyodide
+
 
 # Ensure C extension is used, and then CFFI as fallback, error if neither
-os.environ["PYTHON_ZSTANDARD_IMPORT_POLICY"] = "default"
+@pytest.fixture(autouse=True)
+def env(selenium):
+    selenium.run_python(
+        """
+      import os
+      os.environ["PYTHON_ZSTANDARD_IMPORT_POLICY"] = "default"
+      """
+    )
+    yield
 
-from pytest_pyodide import run_in_pyodide
 
 # ------- Some compression tests ----------------------------------------------
 
--- a/packages/zstandard/test_zstandard.py
+++ b/packages/zstandard/test_zstandard.py
@@ -1,0 +1,116 @@
+import os
+
+# Ensure C extension is used, and then CFFI as fallback, error if neither
+os.environ["PYTHON_ZSTANDARD_IMPORT_POLICY"] = "default"
+
+import zstandard as zstd
+from pytest_pyodide import run_in_pyodide
+
+# ------- Some compression tests ----------------------------------------------
+
+
+@run_in_pyodide(packages=["zstandard"])
+def test_zstandard_compression_and_decompression(selenium):
+    data = b"foo"
+    compress = zstd.ZstdCompressor(
+        write_checksum=True, write_content_size=True
+    ).compress
+    decompress = zstd.ZstdDecompressor().decompress
+
+    assert decompress(compress(data)) == data
+
+
+@run_in_pyodide(packages=["zstandard"])
+def test_zstandard_compression_and_decompression_with_level(selenium):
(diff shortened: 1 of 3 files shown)


agriyakhetarpal:
Looks like the build timed out before it could build `zstandard`. @ryanking13 or @hoodmane, could you please re-trigger the relevant CircleCI runs? Thanks! `zstandard` builds in ~16 seconds locally, FYI.

hoodmane:
`context deadline exceeded` means that too much time passed without anyone writing to stdout. We could add to `pyodide build-recipes` in CI a pulse that occasionally just writes a keepalive message so that this won't happen.

agriyakhetarpal:
There were two failing tests in 71c7e72874a5044046e38eb372c3ca77ac1a0e39 – I fixed them. Please let me know if I should add any more tests. Otherwise, I think this is ready to merge.

hoodmane:
I sent you an email, did you receive it? (No reply needed, just want confirmation that I got the right email for you.)

agriyakhetarpal:
Yes, received it, @hoodmane (I was going to reply later in the day today!).

packages/zstandard/test_zstandard.py
```diff
@@ -0,0 +1,129 @@
+import os
+
+# Ensure C extension is used, and then CFFI as fallback, error if neither
+os.environ["PYTHON_ZSTANDARD_IMPORT_POLICY"] = "default"
```
hoodmane, line 4: This looks wrong to me b/c you want the environment variable set inside Pyodide. You'll want:
```py
import pytest


@pytest.fixture
def env(selenium):
   selenium.run_python(
      """
      import os
      os.environ["PYTHON_ZSTANDARD_IMPORT_POLICY"] = "default"
      """
   )
   yield
```
and then pass env as a fixture to all the tests.

hoodmane (approved):
Thanks @agriyakhetarpal. Tests look good. One comment.

agriyakhetarpal:
Thanks for the feedback! Addressed in a7487a82b54c0173513f9f6c26cab9af00115106 by adding a `pytest` fixture and I additionally used `autouse=True` so that I do not need to apply it across all the functions.

packages/zstandard/test_zstandard.py
```diff
@@ -0,0 +1,138 @@
...
+
+
+# Ensure C extension is used, and then CFFI as fallback, error if neither
+@pytest.fixture(autouse=True)
+def env(selenium):
+    selenium.run_python(
+        """
+      import os
+      os.environ["PYTHON_ZSTANDARD_IMPORT_POLICY"] = "default"
+      """
```
hoodmane, line 12: Fix indentation

packages/zstandard/test_zstandard.py
```diff
@@ -0,0 +1,138 @@
...
+# Ensure C extension is used, and then CFFI as fallback, error if neither
+@pytest.fixture(autouse=True)
+def env(selenium):
+    selenium.run_python(
+        """
+        import os
+        os.environ["PYTHON_ZSTANDARD_IMPORT_POLICY"] = "default"
+        """
+    )
+    yield
```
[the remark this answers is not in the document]
agriyakhetarpal: Maybe we can remove this altogether? I noticed that this is the default policy regardless of whether we set an environment variable or not: https://python-zstandard.readthedocs.io/en/stable/api_usage.html#api-usage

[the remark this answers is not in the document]
hoodmane: Well we could run all the tests with this set to cext and cffi and see if they both work...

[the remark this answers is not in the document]
agriyakhetarpal: That's quite reasonable as well. Let me think of a solution on how to do this – something with the above fixture + a combination of `@pytest.parametrize(["cext", "cffi"])`, perhaps.

packages/zstandard/test_zstandard.py
```diff
@@ -2,22 +2,24 @@
 from pytest_pyodide import run_in_pyodide
 
 
-# Ensure C extension is used, and then CFFI as fallback, error if neither
+# Ensure that tests are executed with both the C extension and the CFFI backend
+@run_in_pyodide(packages=["zstandard"])
+def pytest_generate_tests(selenium, metafunc):
+    if "zstd_backend" in metafunc.fixturenames:
+        metafunc.parametrize("zstd_backend", ["cext", "cffi"])
```
hoodmane, line 9: I'm not sure what's going on here now. You should be able to use a single parametrized fixture for this. You shouldn't need to have a `pytest_generate_tests` or parametrize each test func individually.

packages/zstandard/test_zstandard.py
```diff
@@ -2,22 +2,24 @@
...
-def env(selenium):
-    selenium.run_python(
-        """
-        import os
-        os.environ["PYTHON_ZSTANDARD_IMPORT_POLICY"] = "default"
-        """
-    )
-    yield
+def set_zstd_backend(selenium, monkeypatch, zstd_backend):
+    monkeypatch.setenv("PYTHON_ZSTANDARD_IMPORT_POLICY", zstd_backend)
```
hoodmane, line 15: monkeypatch won't work with run_in_pyodide. Try:
```py
@run_in_pyodide
def set_zstd_backend(selenium, zstd_backend):
    import os
    os.environ["PYTHON_ZSTANDARD_IMPORT_POLICY"] = zstd_backend

@pytest.mark.fixture(params=["a", "b"])
def zstd_backend(selenium, request):
    set_zstd_backend(selenium, request.param)
```

packages/zstandard/test_zstandard.py
agriyakhetarpal: Ah, I was trying to use this guide for parametrization: https://pytest-with-eric.com/introduction/pytest-generate-tests/

Let me try with the code snippet you added below.

agriyakhetarpal:
In 8ce33140883b3fb2da9132fc9d9878b38eaa5fcc, I added `zstd.backend` so that we can avoid using the environment variable and instead set the backend programmatically. This should run the tests with both backends.

agriyakhetarpal:
In c66aaafc1c008bd709f14451be6c740a38e89ce3, I succumbed to parameterising each function directly... this shouldn't fail on anything and I thought it would be okay since there are not a lot of tests.

agriyakhetarpal:
Builds with both `cext` and `cffi` are passing on CircleCI, @hoodmane. Please let me know if anything else is needed to be done. Ready for review/merging, again.

hoodmane:
Thanks @agriyakhetarpal!

packages/zstandard/test_zstandard.py
hoodmane, line 1: Yeah this commit definitely wasn't going to work. The `request` fixture is not pickleable. You'd still need the pair of functions so that you can access `request.param` in the host runtime and then pickle just that string.

hoodmane: Oh, but I see it was wrong before this commit too because you put `@run_in_pyodide` on the fixture which is also wrong. If you want to avoid manually parametrizing each test, the correct code is:
```py
@run_in_pyodide(packages=["zstandard"])
def set_zstd_backend(selenium, zstd_backend):
    import zstandard as zstd

    zstd.backend = request.param


@pytest.fixture(params=["cext", "cffi"], autouse=True)
def zstd_backend(selenium, request):
    # Runs in host, request not pickleable so we can't send it to Pyodide.
   # Look up `request.param` which is a string hence pickleable and send that to Pyodide
    set_zstd_backend(selenium, request.param)
```

agriyakhetarpal: Oh, I didn't realise that – I thought the reason why it didn't work before was because I didn't add a `yield` statement for `set_zstd_backend`. Do we need to revert this change? I can add the fix you suggested.

hoodmane: Nah you can make a follow-up that does that. Label it nfc for nonfunctional change.

Thanks!

PR #4794, opened 2024-05-25
Retry on HTTPException when downloading packages

ryanking13:
### Description

This fix is mainly to handle the `IncompleteRead` error that occasionally happens when downloading packages.


hoodmane:
Thanks!

PR #4798, opened 2024-05-25
BLD Extract out xbuildenv create command from pyodide-build

ryanking13:
### Description

Related: #4557

This removes `pyodide xbuildenv create` command from pyodide-build, and copies it as a separate script into the `tools` directory.

What files are included in xbuildenv is dependent on versions of pyodide. Therefore, given that we plan to decouple the versions of pyodide-build and pyodide, I don't think this command should be part of pyodide-build, hence this PR.

### Checklists

- [x] Add / update tests


tools/create_xbuildenv.py
```diff
@@ -120,8 +121,42 @@ def create(
         encoding="utf8",
     )
     if res.returncode != 0:
-        logger.error("Failed to run pip freeze:")
-        exit_with_stdio(res)
```
hoodmane, line 124: Why not import `exit_with_stdio` from `pyodide_build`?

hoodmane (approved):
Thanks @ryanking13! Great work on unvendoring.

tools/create_xbuildenv.py
ryanking13: I tried to remove dependencies on pyodide-build as much as possible, so that changes in pyodide-build would not break this script.

PR #4797, opened 2024-05-25
Fix deprecated warnings in pyodide-build

ryanking13:
This fixes several warnings that are shown when running pyodide-build tests:

```
pyodide-build/pyodide_build/xbuildenv_releases.py:21
  /home/runner/work/pyodide/pyodide/pyodide-build/pyodide_build/xbuildenv_releases.py:21: PydanticDeprecatedSince20: `pydantic.config.Extra` is deprecated, use literal values instead (e.g. `extra='allow'`). Deprecated in Pydantic V2.0 to be removed in V3.0. See Pydantic V2 Migration Guide at https://errors.pydantic.dev/2.7/migration/
    extra=pydantic.Extra.forbid, title="CrossBuildEnvReleasesSpec"

pyodide-build/pyodide_build/tests/test_xbuildenv.py: 5 warnings
  /opt/hostedtoolcache/Python/3.12.1/x64/lib/python3.12/tarfile.py:2220: DeprecationWarning: Python 3.14 will, by default, filter extracted tar archives and reject files or modify their metadata. Use the filter argument to control this behavior.
    warnings.warn(

  /home/runner/work/pyodide/pyodide/pyodide-build/pyodide_build/buildall.py:191: DeprecationWarning: datetime.datetime.utcfromtimestamp() is deprecated and scheduled for removal in a future version. Use timezone-aware objects to represent datetimes in UTC: datetime.datetime.fromtimestamp(timestamp, datetime.UTC).
    time = datetime.utcfromtimestamp(elapsed_time)
```

PR #4801, opened 2024-05-26
NFC: better parametrize tests against C extension and CFFI backend (`zstandard`)

agriyakhetarpal:
### Description

This is a follow-up for #4792 as requested in https://github.com/pyodide/pyodide/pull/4792#discussion_r1615289414.

### Checklists

- [x] Add / update tests


hoodmane:
Just to be clear @agriyakhetarpal, I wouldn't normally quibble over non functional stuff like this, we usually just try to get the job done by whatever way possible. The main goal here is to help you understand pytest-pyodide a bit better, the code before this patch was totally fine.

I try to call out "less important" suggestions clearly. I generally review people more carefully if they are a committed community member since they learn and become more useful and since they are more likely to have the energy to actually make the requested changes. But feel free to push back on my suggestions if you find that it is making it difficult for you to get to what you're actually trying to do.

Also, though, it looks like we are seeing an influx of development effort on Pyodide in the last few months. If this trend continues, perhaps we can apply higher standards of review more generally.

hoodmane:
Thanks again for your hard work and enthusiasm!

hoodmane:
@ryanking13 failure is `_unpack_zipfile() got an unexpected keyword argument 'filter'` looks due to #4797?

agriyakhetarpal:
Thanks for the comment, @hoodmane. The project is fantastic at not what only what it does but also because of the maintainers who are trying to strive for as clean and quality code as possible. I have no qualms with receiving a higher standard of reviews on further PRs, that helps me serve the project better and also learn more about its internals. I plan to stay committed towards making further contributions.

hoodmane:
> who are trying to strive for as clean and quality code as possible

When you have bodies buried in your basement it's best to keep your lawn neat ;-)

agriyakhetarpal:
Thanks, @ryanking13 – just rebased!

agriyakhetarpal:
Fixing the CI failure now

hoodmane:
I'd suggest testing locally since especially for packages the time it takes for CI to get to it is quite long.

agriyakhetarpal:
> I'd suggest testing locally since especially for packages the time it takes for CI to get to it is quite long.

Oops, just read your comment as I pushed – I'll make note of this for future runs. Is testing locally just the same as running `pytest packages/mypackage/test_mypackage.py`? The reason I ask is because I didn't find anything specific about this in the documentation, except https://pyodide.org/en/stable/development/new-packages.html#writing-tests-for-your-package, and running `pytest` on this file inside a Docker container just returns `Error: Cannot find module '/src/dist/pyodide'` even though I did build both emsdk and CPython.

packages/zstandard/test_zstandard.py
```diff
@@ -22,10 +22,10 @@ def zstd_backend(selenium, request):
 
 
 @run_in_pyodide(packages=["zstandard"])
-def test_zstandard_compression_and_decompression(selenium, backend):
```
hoodmane, line 9: You pass `autouse=True` so you shouldn't need to do anything in the tests. Just remove the backend param and remove the assignment `zstd.backend = zstd_backend`. It is all handled by the fixture.

agriyakhetarpal: Thanks, that makes sense. Addressed in b7d336f57499183037df7eb76cbd8ba8e2920c74.

PR #4803, opened 2024-05-26
pyodide build: Only retag platformed wheels

hoodmane:
This matches pyodide build's logic to the logic in buildpkg

- [ ] Add / update tests

### DIFF
--- a/pyodide-build/pyodide_build/out_of_tree/build.py
+++ b/pyodide-build/pyodide_build/out_of_tree/build.py
@@ -41,7 +41,9 @@ def run(
         built_wheel = pypabuild.build(srcdir, outdir, env, config_settings)
 
     wheel_path = Path(built_wheel)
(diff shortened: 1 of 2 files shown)


hoodmane:
@ryanking13 can you help me test this? Seems like one of the fixtures you've just added should help...

ryanking13:
If what you want is to make xbuildenv installed, you can use `dummy_xbuildenv` fixture:

```python
# flake8: noqa

from .fixture import (
    dummy_xbuildenv,
    dummy_xbuildenv_url,
    reset_env_vars,
    reset_cache,
)

def test_non_platformed_build(dummy_xbuildenv, tmp_path):
    ...

```

Note that you need to import all those fixtures, to make pytest aware of those fixtures. It is not ideal and I think I'll be able to fix it when we unvendor pyodide-build.

hoodmane:
That did the trick.

PR #4806, opened 2024-05-26
FIX Show help when `-h` is passed to cli commands

### LINKED ISSUE #4783: `pyodide build -h` should print help text
## 🐛 Bug

`pyodide build -h` treats `-h` as a package name rather than as a request for help text.

hoodmane:
Resolves #4783.

- [ ] Add / update tests


ryanking13 (approved):
Thanks! Personally, I am okay with not adding tests for this if it is tricky to add one.

hoodmane:
Ok. It shouldn't be that hard all you have to do is run `pyodide build` and check the output. It's tested downstream in cibuildwheel...

PR #4808, opened 2024-05-27
CI Fix version parsing in update_cross_build_releases workflow

ryanking13:
This workflow was not running correctly (log: https://github.com/pyodide/pyodide/actions/runs/9184058597/job/25255773845)


hoodmane:
I think this broke the job:
https://github.com/pyodide/pyodide/actions/runs/9258385099/workflow

ryanking13:
Hmm,,, probably I should use `tag_name` instead of `name`... not sure why it is null value.

PR #4809, opened 2024-05-27
TST Try to make safari tests fail a little less often

ryanking13:
### Description

`test_webworker_zero_timeout` tests were always failing at the first run in Safari. This is because safari does not allow multiple sessions to run simultaneously. We are handling it in [pytest-pyodide](https://github.com/pyodide/pytest-pyodide/blob/f957dcd510eb62af286df608ed9a1861adce1b6d/pytest_pyodide/hook.py#L265-L279), but it was not working properly.

We can address this by improving the behavior of pytest-pyodide, but for now, this PR separates webworker tests for a workaround.


hoodmane (approved):
Thanks!

PR #4811, opened 2024-05-28
Fix incorrect logging module usage

### LINKED ISSUE #4810: Pyodide config recipe broken
## 🐛 Bug

The following recipe in the docs:

```bash
echo EMSCRIPTEN_VERSION=$(pyodide config get emscripten_version) >> $GITHUB_ENV
```

Now produces:

```yaml
EMSCRIPTEN_VERSION: Starting new HTTPS connection (1): github.com:443 [https://github.com:443](https://github.com/) "GET /pyodide/pyodide/releases/download/0.26.0/xbuildenv-0.26.0.tar.bz2 HTTP/1.1" 302 0 Starting new HTTPS connection (1): objects.githubusercontent.com:443 [https://objects.githubusercontent.com:443](https://objects.githubusercontent.com/) "GET /github-production-release-asset-2e65be/122663163/de270f41-72a5-4998-828e-bb88404ce29e?X-Amz-Algorithm=AWS4-HMAC-SHA256&X-Amz-Credential=releaseassetproduction%2F20240528%2Fus-east-1%2Fs3%2Faws4_request&X-Amz-Date=20240528T041437Z&X-Amz-Expires=300&X-Amz-Signature=e7e945b98795915f19c148130500dddb77cc6520852a3077883dc7e53da6edd1&X-Amz-SignedHeaders=host&actor_id=0&key_id=0&repo_id=122663163&response-content-disposition=attachment%3B%20filename%3Dxbuildenv-0.26.0.tar.bz2&response-content-type=application%2Foctet-stream HTTP/1.1" 200 19429390 3.1.58
```

Which does not work when passed to the next step!

### To Reproduce

(Above, or in https://github.com/scikit-hep/boost-histogram/pull/938)

### Expected behavior

If it really needs to print stuff like that, it should go to stderr.

### Environment

- Pyodide Version: 0.26.0
- Browser version:

ryanking13:
### Description

Resolve #4810 

According to the [documentation](https://docs.python.org/3/library/logging.html#logger-objects), we should never instantiate the logger directly, and instantiating a logger was resulting in a side effect that the log from urllib3 is propagated to pyodide-build.

> Loggers have the following attributes and methods. Note that Loggers should NEVER be instantiated directly

\+ We are kind of abusing the logging module to CLI outputs, which need to be addressed later.

### Checklists

### DIFF
--- a/pyodide-build/pyodide_build/logger.py
+++ b/pyodide-build/pyodide_build/logger.py
@@ -77,7 +77,7 @@ console_stderr = CIAwareConsole(stderr=True, theme=COLOR_THEME)
 logging.setLoggerClass(_Logger)
 
 
-def _get_logger(log_level: int) -> logging.Logger:
+def _get_logger(log_level: int) -> _Logger:
     logger = logging.getLogger(__name__)
     logger = cast(_Logger, logger)
     logger.setLevel(log_level)


pyodide-build/pyodide_build/logger.py
```diff
@@ -73,10 +73,11 @@ def format(self, records: logging.LogRecord) -> str:
 
 console_stdout = CIAwareConsole(theme=COLOR_THEME)
 console_stderr = CIAwareConsole(stderr=True, theme=COLOR_THEME)
+logging.setLoggerClass(_Logger)
 
 
 def _get_logger(log_level: int) -> _Logger:
```
hoodmane, line 79:
```suggestion
def _get_logger(log_level: int) -> Logger:
```

hoodmane:
Can we add a test?

PR #4812, opened 2024-05-28
FIX platform.system should be "Emscripten" not "emscripten"

### LINKED ISSUE #4813: Add support for environment markers in dependency specifiers
## Description

`micropip` at the time of writing apparently does not work with dependency specifiers – specifically, [environment markers](https://packaging.python.org/en/latest/specifications/dependency-specifiers/#environment-markers).

A reproducer is as follows:

##### `pyproject.toml`
```toml
[project]
name = "mypackage"
dependencies = [
    'asciitree',
    'numpy>=1.26.2',  # this works
    'fasteners; sys_platform != "emscripten"',  # or 'fasteners; platform_system != "Emscripten"'
    'scipy',
    # ...
]
license = { text = "BSD-3-Clause" }
# and so on
```

which ignores the restriction I put in place on `fasteners`, and proceeds to install it anyway ([workflow logs](https://github.com/agriyakhetarpal/zarr-python/actions/runs/9210973305/job/25339222585#step:8:48)). Though, I understand that this example might be a stub – another edge case with some additional context plus more logs is available below.

## Expected behaviour

The expected behaviour would be to support the parsing of environment-marker-based dependency specifiers when installing a package and therefore conditionally include or exclude dependencies from PEP-621-style dependency metadata in `pyproject.toml` through the supported grammar mentioned in the Python Packaging guide.

In this case, it should exclude `fasteners`.

## Additional context

I stumbled across this when I was trying to add out-of-tree Pyodide builds for the Zarr package (xref: zarr-developers/zarr-python#1903), where `zarr` requires `numcodecs` as a dependency, but it does not include the `[msgpack]` optional dependency for `numcodecs`. This makes the tests suite fail to run because of import errors across the package (not sure about the reason) – which is why I was trying to add something like this:

```toml
dependencies = [
    'numcodecs>=0.10.0',
    'numcodecs[msgpack]>=0.10.0; platform_system == "Emscripten"',
```

to conditionally include the `msgpack` dependency when running in a Pyodide virtual environment, but this does not currently work and `micropip` does not install `msgpack` (the `[msgpack]` optional dependency is defined to install just the `msgpack` package as of now).

However, if I remove `platform_system == "Emscripten"` from the above TOML table, and keep just `numcodecs[msgpack]>=0.10.0`, it _does_ install `msgpack` as expected ([workflow logs](https://github.com/agriyakhetarpal/zarr-python/actions/runs/9211012984/job/25339362619#step:8:51)).

While I can add `msgpack` as a dependency directly at this time because there is no pin on its version, it could cause issues in situations where there would be constraints on optional dependencies (perhaps pyodide/micropip#93 would be helpful there) – say, if `numcodecs[msgpack]` were to require `msgpack>=0.12.0`, I would have to install `msgpack>=0.12.0` in a further and separate `pip install ...` command invocation, which is probably okay for one dependency or two, but for cases where there can be several optional dependencies, there would not be much of a choice but to either carefully (and manually) sync in multiple locations the versions of the packages being installed, or prepare a lockfile beforehand and keep updating it.

hoodmane:
Resolves #4813
cc @agriyakhetarpal


ryanking13 (commented):
It seems like some tests fail?

Issue #4799, opened 2024-05-26
pyodide build should read settings from pyproject.toml


hoodmane:
> Could pyodide build read from tool.pyodide.build in pyproject.toml, by the way, perhaps? At least the exports setting might be useful there!

Said @henryiii https://github.com/pypa/cibuildwheel/pull/1456#issuecomment-2132064435

This is a great idea cc @ryanking13.

ryanking13:
Yes, that's what I was planning to put here:

https://github.com/pyodide/pyodide/blob/81e4d200dbdf74fdaffa7d59c544c8e5e56c96d2/pyodide-build/pyodide_build/config.py#L92-L94

We should probably use `tool.pyodide_build`, as we are using `tool.pyodide` to detect whether we are in an in-tree build (or change the logic of detecting in-tree build).

@RulerOfCakes Are you interested in working on this?

hoodmane:
Especially now that the public command is called `pyodide` I think that should be the public key and we should use something else internally.

hoodmane:
@henryiii was complaining that `pip install pyodide` doesn't install the Pyodide cli and `pip install pyodide-cli` is also useless. We need to do some work on the ux of out of tree builds, particularly the install step.

ryanking13:
> We need to do some work on the ux of out of tree builds, particularly the install step.

Yeah, I agree that there are a lot of pyodide-* tools, which can be confusing.

But even if most people are interested in pyodide-build, it's not the only pyodide tool, so I'm not sure how best to name it.

Maybe we can add all `pyodide-*` packages as a dependency of `pyodide` so that if users install `pip install pyodide`, then all pyodide tools become available.

ryanking13:
@RulerOfCakes thanks! Here are some information that will help you do this task:

0.

What we want is to let people specify some pyodide-build configs in `pyproject.toml` file. For instance,

```toml
[tool.pyodide.build]
meson_cross_file = "path/to/meson.cross"
```

1.

We have a fixed set of variables that are used in pyodide-build:

https://github.com/pyodide/pyodide/blob/81e4d200dbdf74fdaffa7d59c544c8e5e56c96d2/pyodide-build/pyodide_build/config.py#L109-L111

[...]

PR #4831, opened 2024-06-02
Read build settings from configuration file

RulerOfCakes:
### Description

Added implementation to read user-defined build settings from `$(PYODIDE_ROOT)/pyproject.toml`. As there is no clear distinction yet regarding which configuration keys should be overridable by the user, I have added a list of sample keys as suggested in #4799 .

It will read the configuration under the following entry:
```toml
[tool.pyodide.build]
rust_toolchain = "nightly"
# ...
```

### Checklists

- [x] Add / update tests

### DIFF
--- a/pyodide-build/pyodide_build/config.py
+++ b/pyodide-build/pyodide_build/config.py
@@ -173,7 +173,7 @@ BUILD_KEY_TO_VAR: dict[str, str] = {
 
 BUILD_VAR_TO_KEY = {v: k for k, v in BUILD_KEY_TO_VAR.items()}
 
-# Configuration keys that can be overriden by the user.
+# Configuration keys that can be overridden by the user.
 # TODO: distinguish between variables that are overridable by the user and those that are not.
 OVERRIDABLE_BUILD_KEYS = {
     "side_module_cflags",
--- a/pyodide-build/pyodide_build/config.py
+++ b/pyodide-build/pyodide_build/config.py
@@ -94,23 +94,12 @@ class ConfigManager:
         self, curdir: Path, env: Mapping[str, str]
     ) -> Mapping[str, str]:
         # avoid circular import
-        from pyodide_build.build_env import search_pyodide_root
+        from pyodide_build.build_env import search_pyproject_toml
 
-        pyproject_dir = search_pyodide_root(curdir)
+        pyproject_path, configs = search_pyproject_toml(curdir)
 
-        if pyproject_dir is None:
+        if pyproject_path is None:
             return {}
-
-        pyproject_file = pyproject_dir / "pyproject.toml"
-
-        if not pyproject_file.is_file():
-            return {}
-
-        try:
-            with pyproject_file.open("rb") as f:
-                configs = tomllib.load(f)
-        except tomllib.TOMLDecodeError as e:
-            raise ValueError(f"Could not parse {pyproject_file}.") from e
 
         if (
             "tool" in configs
--- a/pyodide-build/pyodide_build/config.py
+++ b/pyodide-build/pyodide_build/config.py
@@ -1,6 +1,5 @@
 import os
 import subprocess
-import tomllib
 from collections.abc import Mapping
 from pathlib import Path
 from types import MappingProxyType
--- a/pyodide-build/pyodide_build/config.py
+++ b/pyodide-build/pyodide_build/config.py
@@ -28,7 +28,7 @@ class ConfigManager:
         self._config = {
             **self._load_default_config(),
             **self._load_makefile_envs(),
-            **self._load_config_file(pyodide_root / "pyproject.toml", os.environ),
+            **self._load_config_file(Path.cwd(), os.environ),
             **self._load_config_from_env(os.environ),
         }
 
@@ -91,8 +91,18 @@ class ConfigManager:
         }
 
     def _load_config_file(
-        self, pyproject_file: Path, env: Mapping[str, str]
+        self, curdir: Path, env: Mapping[str, str]
     ) -> Mapping[str, str]:
+        # avoid circular import
+        from pyodide_build.build_env import search_pyodide_root
+
+        pyproject_dir = search_pyodide_root(curdir)
+
+        if pyproject_dir is None:
+            return {}
+
+        pyproject_file = pyproject_dir / "pyproject.toml"
+
         if not pyproject_file.is_file():
             return {}
 
--- a/pyodide-build/pyodide_build/config.py
+++ b/pyodide-build/pyodide_build/config.py
@@ -2,6 +2,7 @@ import os
 import subprocess
 from collections.abc import Mapping
 from pathlib import Path
+import tomllib
 from types import MappingProxyType
 
 from .common import _environment_substitute_str, exit_with_stdio
@@ -17,7 +18,7 @@ class ConfigManager:
 
         1. Command line arguments (TODO)
         2. Environment variables
-        3. Configuration file (TODO)
+        3. Configuration file
         4. Makefile.envs
         5. Default values
     """
@@ -27,7 +28,7 @@ class ConfigManager:
         self._config = {
             **self._load_default_config(),
             **self._load_makefile_envs(),
-            **self._load_config_file(),
+            **self._load_config_file(pyodide_root / "pyproject.toml", os.environ),
             **self._load_config_from_env(os.environ),
         }
 
@@ -89,9 +90,23 @@ class ConfigManager:
             BUILD_VAR_TO_KEY[key]: env[key] for key in env if key in BUILD_VAR_TO_KEY
         }
 
-    def _load_config_file(self) -> Mapping[str, str]:
-        # TODO: Implement this
-        return {}
+    def _load_config_file(self, pyproject_file: Path, env: Mapping[str,str]) -> Mapping[str, str]:
+        if not pyproject_file.is_file():
+            return {}
+
+        try:
+            with pyproject_file.open("rb") as f:
+                configs = tomllib.load(f)
+        except tomllib.TOMLDecodeError as e:
+            raise ValueError(f"Could not parse {pyproject_file}.") from e
+
+        if "tool" in configs and "pyodide" in configs["tool"] and "build" in configs["tool"]["pyodide"]:
+            return {
+                key: _environment_substitute_str(v, env)
+                for key, v in configs["tool"]["pyodide"]["build"].items() if key in OVERRIDABLE_BUILD_KEYS
+            }
+        else:
+            return {}
 
     @property
     def config(self) -> Mapping[str, str]:
@@ -150,6 +165,15 @@ BUILD_KEY_TO_VAR: dict[str, str] = {
 }
 
 BUILD_VAR_TO_KEY = {v: k for k, v in BUILD_KEY_TO_VAR.items()}
+
+# Configuration keys that can be overriden by the user.
+# TODO: distinguish between variables that are overridable by the user and those that are not.
+OVERRIDABLE_BUILD_KEYS = {
+    "side_module_cflags",
+    "side_module_ldflags",
+    "rust_toolchain",
+    "meson_cross_file",
+}
 
 # Default configuration values.
 TOOLS_DIR = Path(__file__).parent / "tools"
--- a/pyodide-build/pyodide_build/config.py
+++ b/pyodide-build/pyodide_build/config.py
@@ -191,8 +191,9 @@ BUILD_VAR_TO_KEY = {v: k for k, v in BUILD_KEY_TO_VAR.items()}
 # Configuration keys that can be overridden by the user.
 # TODO: distinguish between variables that are overridable by the user and those that are not.
 OVERRIDABLE_BUILD_KEYS = {
-    "side_module_cflags",
-    "side_module_ldflags",
+    "cflags",
+    "cxxflags",
+    "ldflags",
     "rust_toolchain",
     "meson_cross_file",
 }
--- a/pyodide-build/pyodide_build/config.py
+++ b/pyodide-build/pyodide_build/config.py
@@ -1,8 +1,8 @@
 import os
 import subprocess
+import tomllib
 from collections.abc import Mapping
 from pathlib import Path
-import tomllib
 from types import MappingProxyType
 
 from .common import _environment_substitute_str, exit_with_stdio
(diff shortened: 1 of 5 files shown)


pyodide-build/pyodide_build/config.py
```diff
@@ -27,7 +28,7 @@ def __init__(self, pyodide_root: Path):
         self._config = {
             **self._load_default_config(),
             **self._load_makefile_envs(),
-            **self._load_config_file(),
+            **self._load_config_file(pyodide_root / "pyproject.toml", os.environ),
```
ryanking13, line 31: Let's not read the configuration file from pyodide_root. In out-of-tree build, normally users will not access that directory.

Instead, let's pass the cwd as a parameter, and search the pyproject.toml file by looking for the parent directories (just like how we do in [search_pyodide_root](https://github.com/pyodide/pyodide/blob/ff3942e942cda1e538a6f83247ebfb0f93cecbd5/pyodide-build/pyodide_build/build_env.py#L94C5-L94C24))

pyodide-build/pyodide_build/tests/test_config.py
```diff
@@ -62,9 +62,31 @@ def test_load_config_from_env(self):
...
+    def test_load_config_from_file(self, tmp_path, reset_env_vars, reset_cache):
+        pyproject_file = tmp_path / "pyproject.toml"
+
+        env = {
+            "MESON_CROSS_FILE": "/path/to/crossfile",
+            "CFLAGS_BASE": "-O2",
+        }
+
+        pyproject_file.write_text("""[tool.pyodide.build]
+                                  invalid_flags = "this_should_not_be_parsed"
```
ryanking13, line 74: Maybe it would be better to show some warning if an invalid flag is used. Otherwise, it will be hard to detect typo.

ryanking13 (commented):
Thanks @RulerOfCakes!

pyodide-build/pyodide_build/config.py
RulerOfCakes: might be a bit too much of a shortcut, but I've changed it to directly use `search_pyodide_root` here instead to achieve the same thing. let me know if it feels wrong 🤔

ryanking13: You should not use `search_pyodide_root,` but I think we can extract some common parts in `search_pyodide_root` and use them here.

When it comes to building packages, pyodide-build has two modes: in-tree and out-of-tree. the in-tree build is when building packages inside `pyodide/pyodide` repository, out-of-tree build is when building packages somewhere else.

`search_pyodide_root` is used to detect if we are in in-tree build, while this PR targets both in-tree and out-of-tree builds. So, using search_pyodide_root here is not what we want.

Instead, I think we can extract a few parts of `search_pyodide_root` and make it a separate function: searching `pyproject.toml` by iterating the parent directories, and returning the content of pyproject.toml. Then we can use that function in both search_pyodide_root and here.

pyodide-build/pyodide_build/config.py
```diff
@@ -151,6 +188,15 @@ def to_env(self) -> dict[str, str]:
 
 BUILD_VAR_TO_KEY = {v: k for k, v in BUILD_KEY_TO_VAR.items()}
 
+# Configuration keys that can be overridden by the user.
+# TODO: distinguish between variables that are overridable by the user and those that are not.
+OVERRIDABLE_BUILD_KEYS = {
+    "side_module_cflags",
```
henryiii, line 194:
```suggestion
    "side_module_cflags",
    "side_module_cxxflags",
```

henryiii:
I'd also keep in mind, even if it's not in this PR, that there will be some settings not in this list, specifically `tool.pyodide.build.exports` for the `exports` setting.

pyodide-build/pyodide_build/config.py
ryanking13: Let's also remove "side_module" prefix as discussed in https://github.com/pyodide/pyodide/issues/4799#issuecomment-2151643233

pyodide-build/pyodide_build/config.py
```diff
@@ -151,6 +176,16 @@ def to_env(self) -> dict[str, str]:
...
 
+# Configuration keys that can be overridden by the user.
+# TODO: distinguish between variables that are overridable by the user and those that are not.
+OVERRIDABLE_BUILD_KEYS = {
+    "cflags",
+    "cxxflags",
+    "ldflags",
+    "rust_toolchain",
+    "meson_cross_file",
+}
```
ryanking13, line 187: You'll need to change the name of the variable

here:

https://github.com/pyodide/pyodide/blob/43cce6e5fffa4cfcb0ae3709a9c40844e720dabb/pyodide-build/pyodide_build/config.py#L134-L136

and here

https://github.com/pyodide/pyodide/blob/43cce6e5fffa4cfcb0ae3709a9c40844e720dabb/pyodide-build/pyodide_build/config.py#L174-L176

as well, so that the `cflags` key can affect the `SIDE_MODULE_CFLAGS` env variable.

(We may change the env variable name too, but it will be a breaking change so let's to that in a separate PR).

pyodide-build/pyodide_build/config.py
```diff
@@ -89,9 +89,34 @@ def _load_config_from_env(self, env: Mapping[str, str]) -> Mapping[str, str]:
...
         }
 
-    def _load_config_file(self) -> Mapping[str, str]:
-        # TODO: Implement this
-        return {}
+    def _load_config_file(
+        self, curdir: Path, env: Mapping[str, str]
+    ) -> Mapping[str, str]:
+        # avoid circular import
+        from pyodide_build.build_env import search_pyproject_toml
```
ryanking13, line 96: Let's move `search_pyproject_toml` function to `common.py`, so we don't need to care about the circular import.

ryanking13 (approved):
Thanks again @RulerOfCakes! I have some minor comments, otherwise it generally looks good to me.

pyodide-build/pyodide_build/config.py
RulerOfCakes: oh right, forgot that these were all somewhat coupled to each other.

hoodmane:
Seems like 0.26.1 is going to be a big release for a patch version =)

henryiii:
That sounds great. :)

cibuildwheel is likely to release soon, we are getting macOS free-threading builds and need to get those out.

henryiii:
I'd still like to see #4836 and #4844 at least. The cibuildwheel release probably should be this weekend, though I could adjust it a bit, I'd really like it out by Monday.

hoodmane:
#4844 is done (I don't really know how to test it though). 

#4836 is complicated to fix correctly but there's an easy quick fix and then I can come back to a correct fix. I am now thinking that what needs to be done is that the main wasm binary needs to reexport all imported JavaScript symbols. Then the normal `updateGOT` mechanism will put these exports into the GOT. The problem is that symbols end up in the GOT when either:
1. a wasm library that they are exported from is loaded, or
2. if they are a JS symbol, they aren't exported from any wasm library so they only get inserted when `reportUndefinedSymbols` is called.

I'd also like to change the behavior of `reportUndefinedSymbols` so that it won't mutate the GOT anymore.

hoodmane:
So maybe I'll merge these two today and cut 0.26.1.

henryiii:
If you use a list, this happens: 

```
╭───────────────────── Traceback (most recent call last) ──────────────────────╮
│ /opt/hostedtoolcache/Python/3.12.3/x64/lib/python3.12/site-packages/pyodide_ │
│ build/cli/config.py:57 in get_config                                         │
│                                                                              │
│   54 │   """                                                                 │
│   55 │   Get a value of a single config variable used in pyodide             │
│   56 │   """                                                                 │
│ ❱ 57 │   configs = _get_configs()                                            │
│   58 │                                                                       │
│   59 │   if config_var not in configs:                                       │
│   60 │   │   typer.echo(f"Config variable {config_var} not found.")          │
│                                                                              │
│ /opt/hostedtoolcache/Python/3.12.3/x64/lib/python3.12/site-packages/pyodide_ │
│ build/cli/config.py:29 in _get_configs                                       │
│                                                                              │
│   26                                                                         │
│   27                                                                         │
│   28 def _get_configs() -> dict[str, str]:                                   │
│ ❱ 29 │   init_environment(quiet=True)                                        │
│   30 │                                                                       │
│   31 │   configs: dict[str, str] = get_build_environment_vars(get_pyodide_ro │
│   32                                                                         │
│                                                                              │
│ /opt/hostedtoolcache/Python/3.12.3/x64/lib/python3.12/site-packages/pyodide_ │
│ build/build_env.py:57 in init_environment                                    │
│                                                                              │
│    54 │                                                                      │
│    55 │   root = search_pyodide_root(Path.cwd())                             │
│    56 │   if not root:  # Not in Pyodide tree                                │
│ ❱  57 │   │   root = _init_xbuild_env(quiet=quiet)                           │
│    58 │                                                                      │
│    59 │   os.environ["PYODIDE_ROOT"] = str(root)                             │
│    60                                                                        │
│                                                                              │
│ /opt/hostedtoolcache/Python/3.12.3/x64/lib/python3.12/site-packages/pyodide_ │
│ build/build_env.py:82 in _init_xbuild_env                                    │
│                                                                              │
│    79 │   with context:                                                      │
│    80 │   │   manager = CrossBuildEnvManager(xbuildenv_path)                 │
│    81 │   │   if manager.current_version is None:                            │
│ ❱  82 │   │   │   manager.install()                                          │
│    83 │   │                                                                  │
│    84 │   │   return manager.pyodide_root                                    │
│    85                                                                        │
│                                                                              │
│ /opt/hostedtoolcache/Python/3.12.3/x64/lib/python3.12/site-packages/pyodide_ │
│ build/xbuildenv.py:186 in install                                            │
│                                                                              │
│   183 │   │   except Exception as e:                                         │
│   184 │   │   │   # if the installation failed, remove the downloaded direct │
│   185 │   │   │   shutil.rmtree(download_path)                               │
│ ❱ 186 │   │   │   raise e                                                    │
│   187 │   │                                                                  │
│   188 │   │   return xbuildenv_pyodide_root                                  │
│   189                                                                        │
│                                                                              │
│ /opt/hostedtoolcache/Python/3.12.3/x64/lib/python3.12/site-packages/pyodide_ │
│ build/xbuildenv.py:173 in install                                            │
│                                                                              │
│   170 │   │   │   │   logger.info("Installing Pyodide cross-build environmen │
│   171 │   │   │   │                                                          │
│   172 │   │   │   │   if not skip_install_cross_build_packages:              │
│ ❱ 173 │   │   │   │   │   self._install_cross_build_packages(                │
│   174 │   │   │   │   │   │   xbuildenv_root, xbuildenv_pyodide_root         │
│   175 │   │   │   │   │   )                                                  │
│   176                                                                        │
│                                                                              │
│ /opt/hostedtoolcache/Python/3.12.3/x64/lib/python3.12/site-packages/pyodide_ │
│ build/xbuildenv.py:249 in _install_cross_build_packages                      │
│                                                                              │
│   246 │   │   xbuildenv_pyodide_root                                         │
│   247 │   │   │   Path to the pyodide-root directory inside the xbuildenv di │
│   248 │   │   """                                                            │
│ ❱ 249 │   │   host_site_packages = self._host_site_packages_dir(xbuildenv_py │
│   250 │   │   host_site_packages.mkdir(exist_ok=True, parents=True)          │
│   251 │   │   result = subprocess.run(                                       │
│   252 │   │   │   [                                                          │
│                                                                              │
│ /opt/hostedtoolcache/Python/3.12.3/x64/lib/python3.12/site-packages/pyodide_ │
│ build/xbuildenv.py:296 in _host_site_packages_dir                            │
│                                                                              │
│   293 │   │   │   xbuildenv_pyodide_root = self.pyodide_root                 │
│   294 │   │                                                                  │
│   295 │   │   return Path(                                                   │
│ ❱ 296 │   │   │   build_env.get_build_environment_vars(pyodide_root=xbuilden │
│   297 │   │   │   │   "HOSTSITEPACKAGES"                                     │
│   298 │   │   │   ]                                                          │
│   299 │   │   )                                                              │
│                                                                              │
│ /opt/hostedtoolcache/Python/3.12.3/x64/lib/python3.12/site-packages/pyodide_ │
│ build/build_env.py:120 in get_build_environment_vars                         │
│                                                                              │
│   117 │   """                                                                │
│   118 │   Get common environment variables for the in-tree and out-of-tree b │
│   119 │   """                                                                │
│ ❱ 120 │   config_manager = ConfigManager(pyodide_root)                       │
│   121 │   env = config_manager.to_env()                                      │
│   122 │                                                                      │
│   123 │   env.update(                                                        │
│                                                                              │
│ /opt/hostedtoolcache/Python/3.12.3/x64/lib/python3.12/site-packages/pyodide_ │
│ build/config.py:30 in __init__                                               │
│                                                                              │
│    27 │   │   self._config = {                                               │
│    28 │   │   │   **self._load_default_config(),                             │
│    29 │   │   │   **self._load_makefile_envs(),                              │
│ ❱  30 │   │   │   **self._load_config_file(Path.cwd(), os.environ),          │
│    31 │   │   │   **self._load_config_from_env(os.environ),                  │
│    32 │   │   }                                                              │
│    33                                                                        │
│                                                                              │
│ /opt/hostedtoolcache/Python/3.12.3/x64/lib/python3.12/site-packages/pyodide_ │
│ build/config.py:112 in _load_config_file                                     │
│                                                                              │
│   109 │   │   │   │   │   │   f"WARNING: The provided build key {key} is eit │
│   110 │   │   │   │   │   )                                                  │
│   111 │   │   │   │   │   continue                                           │
│ ❱ 112 │   │   │   │   build_config[key] = _environment_substitute_str(v, env │
│   113 │   │   │                                                              │
│   114 │   │   │   return build_config                                        │
│   115 │   │   else:                                                          │
│                                                                              │
│ /opt/hostedtoolcache/Python/3.12.3/x64/lib/python3.12/site-packages/pyodide_ │
│ build/common.py:133 in _environment_substitute_str                           │
│                                                                              │
│   130 │   │   env = dict(os.environ)                                         │
│   131 │                                                                      │
│   132 │   for e_name, e_value in env.items():                                │
│ ❱ 133 │   │   string = string.replace(f"$({e_name})", e_value)               │
│   134 │                                                                      │
│   135 │   return string                                                      │
│   136                                                                        │
╰──────────────────────────────────────────────────────────────────────────────╯
AttributeError: 'list' object has no attribute 'replace'
```

ryanking13:
Thanks for the report. I think handling non-string types needs to be improved.

PR #4844, opened 2024-06-07
Set platlibdir in venv pip

hoodmane:
Previously we'd read it from the host environment, which could cause trouble in environments like Fedora where it would have an unusual value. Resolves https://github.com/pypa/cibuildwheel/issues/1853

- [ ] Add / update tests

### DIFF
--- a/docs/project/changelog.md
+++ b/docs/project/changelog.md
@@ -32,6 +32,10 @@ myst:
   `InvalidStateError`.
   {pr}`4837`
 
+- {{ Fix }} Pyodide virtual environments now work correctly in Fedora and other
+  platforms with platlibdir not equal to "lib".
+  {pr}`4844`
+
 ### Packages
 
 - New Packages: `pytest-asyncio` {pr}`4819`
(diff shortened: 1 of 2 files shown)


hoodmane:
You can test this on a system with `PLATLIBDIR` equal to `lib`  by doing:
```bash
sudo ln -s /lib/python3.12/ /lib64/
```
and then running this test:
```py
@only_node
@needs_emscripten
def test_platlibdir(selenium, monkeypatch):
    venv_path = Path(".venv-pyodide-tmp-test")
    with venv_ctxmgr(venv_path) as venv:
        monkeypatch.setenv("PYTHONPLATLIBDIR", "lib64")
        result = install_pkg(venv, "regex")
        assert result.returncode == 0
        monkeypatch.setenv("PYTHONPLATLIBDIR", "")
        result = subprocess.run(
            [
                venv_path/ "bin/python",
                "-c",
                "import regex; print(regex.__path__)"
            ], capture_output=True, encoding="utf8"
        )
        print(result.stdout)
        print(result.stderr)
        assert result.returncode == 0
```
Unfortunately without the symlink, Python crashes on startup in the presence of the PLATLIBDIR environment variable so I'm not sure how to do a test that runs correctly when we're not allowed to write to `/lib64`.

ryanking13 (approved):
Thanks!

PR #4846, opened 2024-06-07
Pyfetch abort on cancel

CNSeniorious000:
Related to pyodide/micropip#111

#### New Behaviors:

1. If caught `CancelledError` in `pyfetch`, then send `abort` to `fetch`
2. Added a `abort()` to `FetchResponse` for use cases like reading the headers before abortion (maybe to avoid responses larger than expect)
3. Convert AbortError (which is a JsException) to `OSError` inside `_raise_if_failed`

There is still many corner cases to fix, I just create this draft for opinions.

#### TODOs:

1. abort on cancellation in `buffer()`, `json()`, ... (maybe I need to implement a `@abort_on_cancel` decorator on every async method of `FetchResponse`?
2. About the error handling, do you think it is necessary to convert every JsException into OSError? If so, will it be clearer to subclass `OSError` into maybe `StatusError`, `BodyUsedError` and `AbortError`?
3. Currently the tests seems reasonable but fails. I can't find why, can somebody help me with this?

### Checklists

- [x] Add / update tests


src/py/pyodide/http.py
```diff
@@ -138,9 +144,11 @@ def url(self) -> str:
         """
         return self.js_response.url
 
-    def _raise_if_failed(self) -> None:
+    def _raise_if_failed_or_aborted(self) -> None:
```
hoodmane, line 147: Maybe just leave the name alone? I feel like it's clear enough and with `_or_aborted` this is a bit of a mouthful.

hoodmane:
It seems that we are truly leaking the AbortSignal. I am currently trying to figure out why.

hoodmane:
Seems like it's a bug in `to_js`.

src/py/pyodide/http.py
```diff
@@ -270,6 +278,10 @@ async def unpack_archive(
         filename = self._url.rsplit("/", -1)[-1]
         unpack_buffer(buf, filename=filename, format=format, extract_dir=extract_dir)
 
+    def abort(self, reason: Any = None) -> None:
+        assert self.abort_controller is not None, "abort_controller is not set"
```
ryanking13, line 282: If we are going to allow setting None for `abort_controller`, I think it is better to either raise exception or ignore silently instead of using assert here.

src/py/pyodide/http.py
```diff
@@ -301,9 +313,19 @@ async def pyfetch(url: str, **kwargs: Any) -> FetchResponse:
     {'info': {'arch': 'wasm32', 'platform': 'emscripten_3_1_32',
     'version': '0.23.4', 'python': '3.11.2'}, ... # long output truncated
     """
+    controller = AbortController.new()
+    if "signal" in kwargs:
+        kwargs["signal"] = AbortSignal.any([kwargs["signal"], controller.signal])
```
ryanking13, line 318: [`AbortSignal.any` seems quite new](https://developer.mozilla.org/en-US/docs/Web/API/AbortSignal/any_static), Chrome started to support it in v116 (Released in 2023-08-15), Firefox v124 (2024-03-19), and Node.js v20.3.

So it would be nice if we can achieve this without using any. Personally, I would like to support Node.js v18 until its end-of-life (April 2025).

ryanking13 (commented):
Thanks @CNSeniorious000!

> About the error handling, do you think it is necessary to convert every JsException into OSError? If so, will it be clearer to subclass OSError into maybe StatusError, BodyUsedError and AbortError?

This sounds good to me.

src/py/pyodide/http.py
CNSeniorious000: Thanks. I am really bad at naming things😂

src/py/pyodide/http.py
CNSeniorious000: Thanks for suggestion. Let me try polyfill it in python then

ryanking13: I found a few posts about userland solutions, which may be helpful to you.

- https://github.com/shaseley/abort-signal-any/blob/main/README.md#userland-solutions
- https://samthor.au/2022/abortcontroller-is-your-friend/ (Helpers that may or may not exist section)

CNSeniorious000: Thank you very much🌹

hoodmane: You can do copy paste a js polyfill i'd recommend.

CNSeniorious000: Yeah I will do this tomorrow

CNSeniorious000: Sadly maybe not that simple. [The userland solution](https://github.com/shaseley/abort-signal-any/blob/main/README.md#userland-solutions) will leak memory since it don't remove the event listeners.

However, since we don't need to implement a perfect polyfill for `AbortSignal`, we just need to use it in our `pyfetch`, whose lifespan is under our control.

I plan to remove these event listeners in a `__del__`, when all JsFetchResponse are deleted.

CNSeniorious000: I had a look of [this document](https://docs.google.com/document/d/1LvmsBLV85p-PhSGvTH-YwgD6onuhh1VXLg8jPlH32H4) and noticed that it is actually polyfillable if only the browser supports weak references. But I failed to find a polyfill that implements this (possibly because I lack experience in searching for polyfills, in fact I simply did a quick search on npm).

hoodmane: We assume weak references and finalization registry so a polyfill should be possible.

CNSeniorious000: [Here](https://github.com/pyodide/pyodide/pull/4846/commits/52387f009a0f0edf23eda76a3f055adb401408fd#diff-1fb6404e8153dfd2db40ec8a5d07af6de9a4a5d596710907980397c79204d3aa) it is ~

src/py/pyodide/utils/abort/__init__.py
```diff
@@ -0,0 +1,35 @@
...
+    from js import AbortSignal
+
+    global _abort_signal_any
+
+    if AbortSignal.any is None:
+        from ...code import run_js  # type:ignore[unreachable]
+
+        path = Path(__file__).parent / "polyfill.js"
+        source = path.read_text() + ";polyfillAbortSignal()"
+        _abort_signal_any = run_js(source)
```
hoodmane, line 25: We don't want to use `run_js` here because some people need to run Pyodide in an environment where `eval` is disabled (by content security protocol or otherwise). See how we handled it in `scheduler.ts`:
https://github.com/pyodide/pyodide/blob/main/src/js/scheduler.ts
https://github.com/pyodide/pyodide/blob/main/src/js/api.ts?plain=1#L62
https://github.com/pyodide/pyodide/blob/main/src/py/pyodide/webloop.py?plain=1#L14

I think the reason that `scheduler.ts` exports and it's attached to `API` in `api.ts` is for typescript unit testing. If we don't need typescript unit tests we can attach it directly to API in `abortSignalAny.ts`.

src/py/pyodide/utils/abort/polyfill.js
```diff
@@ -0,0 +1,52 @@
...
+    const controller = new AbortController();
+    const controllerRef = new WeakRef(controller);
+    /** @type {[WeakRef<AbortSignal>, (() => void)][]} */
+    const eventListenerPairs = [];
+    let followingCount = signals.length;
+
+    /** @type {FinalizationRegistry<(callback: () => any) => void>} */
+    const registry = (globalThis.__abortSignalCleanups =
+      globalThis.__abortSignalCleanups ??
+      new FinalizationRegistry((callback) => void callback()));
```
hoodmane, line 16: Let's move this to the top of the file as:
```js
const registry = new FinalizationRegistry((callback) => void callback());
```

hoodmane: I guess we should also do the check at top level:
```js
export let abortSignalAny;
if (AbortSignal.any) {
    abortSignalAny = AbortSignal.any;
} else {
    const registry = new FinalizationRegistry((callback) => void callback());
    // ... define abortSignalAny ponyfill
}
```

src/py/pyodide/utils/abort/__init__.py
hoodmane: But we might as well keep the same pattern where we import in `api.ts` and attach it to API there.

src/py/pyodide/utils/abort/polyfill.js
CNSeniorious000: Should we export it?

src/py/pyodide/utils/abort/__init__.py
CNSeniorious000: I attached it in `abortSignalAny.ts` and imported this script in `api.ts`. I can use the other way of course

src/py/pyodide/utils/abort/polyfill.js
hoodmane: Yeah, then import from `api.ts` and set `API.abortSignalAny = abortSignalAny`. Then from Python you can do 
```py
from pyodide_js._api import abortSignalAny
```
and then you can use it.

CNSeniorious000: I attached it to API and it should work now. I mean, now I did't `export let` it because the only usage is through `pyodide_js._api`. Do you think we need to add some test for it?

CNSeniorious000: Even the core tests surprisingly passed 😂

hoodmane:
@CNSeniorious000 can you merge main? I just merged #4853 so if you bring that in here, the leak checks should pass.

src/js/abortSignalAny.ts
hoodmane, line 1: sphinx-js gets mad that there are no exports in this file. Requires this fix:
https://github.com/pyodide/sphinx-js-fork/pull/164

src/py/pyodide/utils/abort/__init__.py
hoodmane: Thanks!

CNSeniorious000:
Why the related tests in CircleCI failed for cannnot importing `AbortController`?

<details><summary>stack trace example</summary>
<p>

```py
  File "/lib/python3.12/site-packages/micropip/package_index.py", line 278, in query_package
    metadata, headers = await fetch_string_and_headers(url, _fetch_kwargs)
                        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/lib/python3.12/site-packages/micropip/_compat_in_pyodide.py", line 38, in fetch_string_and_headers
    response = await pyfetch(url, **kwargs)
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/lib/python312.zip/pyodide/http.py", line 355, in pyfetch
    controller = AbortController.new()
                 ^^^^^^^^^^^^^^^
NameError: name 'AbortController' is not defined
```

</p>
</details>

ryanking13:
> Why the related tests in CircleCI failed for cannnot importing AbortController?

This looks like a bug in `pytest-pyodide`. I guess we need to pass AbortController when [making a VM context](https://github.com/pyodide/pytest-pyodide/blob/f957dcd510eb62af286df608ed9a1861adce1b6d/pytest_pyodide/node_test_driver.js#L18-L35).

CNSeniorious000:
Found a new strange behavior.

#### Reproduction:

```py
from js import *
from pyodide.http import *

controller = AbortController.new()
controller.abort("reason")

try:
    f = await fetch("/", signal=controller.signal)
except BaseException as err:
    e = err
```

#### Result:

The error can't be caught and you can see this traceback in the console

```py
  File "/lib/python312.zip/_pyodide/_future_helper.py", line 10, in set_exception
    fut.set_exception(val)
TypeError: invalid exception object
```

<details><summary>And these in the devtools</summary>
<p>

```js
Uncaught (in promise) PythonError
    at new_error (pyodide.asm.js:10:9965)
    at pyodide.asm.wasm:0x16dbfd
    at pyodide.asm.wasm:0x16dcc3
    at Module._pythonexc2js (pyodide.asm.js:10:701544)
    at callPyObjectKwargs (pyodide.asm.js:10:64179)
    at Module.callPyObjectMaybePromising (pyodide.asm.js:10:65316)
    at onRejected (pyodide.asm.js:10:28303)
```

</p>
</details>

This can be reproduced in 0.25.* and 0.26.* too, with a little differences in the message

CNSeniorious000:
Sorry, it seems this is because controller.abort() can only receive a `reason` of `Error` type.

But MDN says it is `Any`. I am still investigating this. I think This is what make the test stucked.

hoodmane:
Can we xfail the test in node for now and enable it after we fix pytest-pyodide? Also I keep saying this but it's really annoying that we don't pass the whole global scope in node, ideally we should figure out what's going wrong when we do and fix it.

ryanking13:
> Can we xfail the test in node for now and enable it after we fix pytest-pyodide?

Sure, it's fine with me.

> I keep saying this but it's really annoying that we don't pass the whole global scope in node, ideally we should figure out what's going wrong when we do and fix it.

If you don't mind, I would xfail the flaky tests and revisit it when we have bandwidth.

CNSeniorious000:
> Theoretically, JS can throw anything, but I guess we are assuming that it is an Error object, resulting in an error.

Yes. `AbortController.prototype.abort(reason)`'s `reason` seems to be marked as `Any` but default to a `DOMException`. In [this commit](https://github.com/pyodide/pyodide/pull/4846/commits/0f1bccb86482f54981157449e0d5a0edb325a153) I set it to a `JsException` and the problem is solved.

CNSeniorious000:
> It is failing other tests as well (e.g. micropip)

Yes, it seems lots of tests failed due to this PR so I tend to wait until `pytest-pyodide` got fixed too, instead of affecting too much files.

src/py/pyodide/http.py
```diff
@@ -59,6 +80,35 @@ def open_url(url: str) -> StringIO:
...
+                    _construct_abort_reason(
+                        "\n".join(map(str, e.args)) if e.args else None
+                    )
+                )
+            raise
+
+    return wrapper
+
+
+def _construct_abort_reason(reason: Any) -> JsException | None:  # type:ignore[return]
```
ryanking13, line 107: I guess adding empty return will make mypy happy?

ryanking13 (approved):
Thanks @CNSeniorious000! It looks genrally good to me.

I released pytest-pyodide 0.58.0 that fixes AbortController issue in node, so could you please bump-up pytest-pyodide version, in [requirements.txt](https://github.com/pyodide/pyodide/blob/main/requirements.txt) and in [pyproject.toml](https://github.com/pyodide/pyodide/blob/main/pyodide-build/pyproject.toml)?

src/py/pyodide/http.py
CNSeniorious000: Yes. Every time at such moments, I always miss Pyright😂

CNSeniorious000:
What's the nodejs version of `ci/circleci: test-core-node` and `ci/circleci: test-core-node-nodylink`? I don't get it why tests failed on those environments.

ryanking13:
> What's the nodejs version of ci/circleci: test-core-node? I don't get it why tests failed on those environments.

It's [Node 20.11](https://github.com/pyodide/pyodide/blob/eb1d208638064bafa03e298cfbe358a6b5cc2832/Dockerfile#L1).

src/py/pyodide/http.py
```diff
@@ -270,6 +328,16 @@ async def unpack_archive(
         filename = self._url.rsplit("/", -1)[-1]
         unpack_buffer(buf, filename=filename, format=format, extract_dir=extract_dir)
 
+    def abort(self, reason: Any = None) -> None:
```
ryanking13, line 331: I just realized that we can only call this `abort` method after the fetch is done. Is it a useful behavior? Can it be a reason for why node tests are failing (sending abort signal after the fetch is finished)?

ryanking13: Oh, never mind. This raises AbortError in both Chrome and Node, so sending abort signal after the fetch seems fine.

```js
let controller = new AbortController();

let response = await fetch('http://example.com', {
    signal: controller.signal
});

controller.abort("reason")
await response.text()
```

ryanking13:
> What's the nodejs version of ci/circleci: test-core-node? I don't get it why tests failed on those environments.

Maybe this is a bug in pytest-pyodide NodeRunner. This code runs without error in pytest-pyodide, while in native node.js environments, it raises error.

```python
def test_node_fetch_abort(selenium):
    ret = selenium.run_js(
        """
        let controller = new AbortController();

        let response = await fetch('https://raw.githubusercontent.com/pyodide/pyodide/main/.editorconfig', {
            signal: controller.signal
        });

        controller.abort("reason")
        return await response.text()
        """
    )
    print(ret)
```

src/py/pyodide/http.py
CNSeniorious000: tldr: It may be the reason why `test_pyfetch_manually_abort` failed, but I think it is why `test_pyfetch_custom_abort_signal` failed.

> we can only call this `abort` method after the fetch is done. Is it useful?

IIUC, a fetch in js need to be await twice, first in `await fetch(...`, the second in `await resp.text()`. In python, cancelling a future, or using a custom `AbortController` passed in `kwargs` can abort in both `await`, and they can use `FetchResponse.abort` in the second stage.

The only way to stop a fetch (or a .text()/.json() method of a fetch's response) is to call the `abort` method on the `AbortController` instance. And I added this `FetchResponse.abort()` method just for convenience. Because I personally think python developers may not want to create a `AbortController` and abort it themselves, so I expose a abort method for them to stop the request.

But to be honest, it is only useful in the second stage (await resp.text()), so it may not be very useful and may cause confusion. If you think the drawbacks outweight the benefits I agree to remove that.

In `test_pyfetch_manually_abort`, I abort the resposne after the first stage. Maybe the request is already ended at that time? In other words, maybe the second stage is already ended before I called `resp.abort`.

But in `test_pyfetch_custom_abort_signal`, I abort the signal before calling `pyfetch`, so the request should never start. I can't think of a reason why this happens. I'm trying to reproduce this though.

ryanking13:
Okay, I found a bug in [pytest-pyodide](https://github.com/pyodide/pytest-pyodide/blob/4511f573ed6d07140d32403c0b5a3aa9574a1f0d/pytest_pyodide/node_test_driver.js#L13-L16).

src/py/pyodide/http.py
ryanking13: > But to be honest, it is only useful in the second stage, so it may not be very useful and may cause inconsistency and confusion. If you think the drawbacks outweight the benefits I agree to remove that.

No, I think it is fine to keep the `abort` method in `FetchResponse`. Even if there are not many real-world use cases, someone may use it, and using it for testing isn't that bad.

We may add another Python helper class or function to abort the "first" stage of fetch too as a follow-up PR if you think it might be useful for people.

ryanking13:
I released pytest-pyodide 0.58.1 including node fetch fix, could you please try it?

ryanking13:
Hmm, it seems like we are getting another strange error:

```
TypeError: BodyUsedError.__init__() takes 1 positional argument but 2 were given
```

I suspect this has something to do with how pytest-pyodide pickles/unpickles exceptions, but it would take some time to investigate. Meanwhile, could you update the constructor of the `BodyUsedError` class so that it takes extra `*args` and discards it? I think it would fix the error.

CNSeniorious000:
Good idea

CNSeniorious000:
Found inconsistency between nodejs and web browsers:

The following code always returns `false` in browsers, but `true` in nodejs.

```js
c = new AbortController(); r = await fetch("https://httpbin.org/get", { signal: c.signal }); c.abort(); r.bodyUsed
```

ryanking13:
> The following code always returns false in browsers, but true in nodejs

That's interesting. Maybe it's undefined behavior and behaves differently at different runtimes? Anyway, thanks for testing various scenarios!

The PR looks good for me, so I'll go ahead and merge. Do you want us to make a minor release (0.26.2) including this PR so you can use it in micropip?

CNSeniorious000:
No hurry, micropip should have expected behavior even in older versions of Pyodide, see https://github.com/pyodide/micropip/pull/111#discussion_r1625341622, so it shouldn't depend on this.

Of course I can't think of any reason not to include this PR in a minor version, as it doesn't introduce any backward incompatibilities. I just don't see the need to release 0.26.2 ahead of schedule for this.

hoodmane (commented):
Thanks @CNSeniorious000 this is really great work.

PR #4836, opened 2024-06-04
FIX Prevent segfaults in CLI runner

hoodmane:
After many hours of debugging, I minimized the problem down to this: https://github.com/emscripten-core/emscripten/issues/22052

Thanks to @henryiii for reporting. See thread here for my comments while I was debugging:
https://github.com/scikit-hep/boost-histogram/pull/938

Resolves #2964.

- [x] Add a [CHANGELOG](https://github.com/pyodide/pyodide/blob/main/docs/project/changelog.md) entry
- [x] Add / update tests

PR #4848, opened 2024-06-07
Backports for 0.26.1

hoodmane:
Merge don't squash!

TODO: add #4836.

Read build settings from configuration file
FIX platform.system should be "Emscripten" not "emscripten"
Add pytest-asyncio
ENH Add option to make run_until_complete stack switch
FIX Check if future is done before trying to set a result on it
Fix pyodide-build verbose log by replacing requests with urllib
Fix keys related behaviors on PyProxyJsonAdaptorDictHandlers
Update subtable name for checking in-tree build
Set platlibdir in venv pip
FIX Prevent segfaults in CLI runner


ryanking13:
Thanks for the backporting work!

PR #4850, opened 2024-06-07
Fix path to changelog.md

zqianem:
Just a small documentation fix


hoodmane (approved):
Thanks! This indeed looks like an improvement. The language about non developers being the target audience isn't really accurate since Pyodide is really a project for developers of other software. But I guess it means it's for downstream years.

PR #4855, opened 2024-06-08
CI Fill out more information to pyodide-cross-build-environments.json automatically

ryanking13:
Makes `update_cross_build_releases.py` script a little more smarter so it fills out `python_version`, `emscripten_version`, and `min_pyodide_build_version` fields automatically. So maintainers would not need to find the corresponding versions by themselves.

### DIFF
--- a/tools/update_cross_build_releases.py
+++ b/tools/update_cross_build_releases.py
@@ -78,7 +78,8 @@ def add_version(
         python_version=python_version or "FIXME",
         emscripten_version=emscripten_version or "FIXME",
         min_pyodide_build_version=min_pyodide_build_version or "FIXME",
-        # max_pyodide_build_version=max_pyodide_build_version or "FIXME",
+        # Max version is optional, and maintainers should update it when needed.
+        max_pyodide_build_version=max_pyodide_build_version or None,
     )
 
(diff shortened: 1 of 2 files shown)


tools/update_cross_build_releases.py
```diff
@@ -15,6 +19,10 @@
 
 BASE_URL = "https://github.com/pyodide/pyodide/releases/download/{version}/xbuildenv-{version}.tar.bz2"
 
+# Pyodide build version that is compatible with the latest cross-build environment
+# Note for maintainers: update this value when there is a breaking changes in the cross-build environment
+MIN_COMPATIBLE_PYODIDE_BUILD_VERSION = "0.26.0"
```
hoodmane, line 24: Could we move this to `Makefile.envs` in a followup? I don't think people will be looking at this file very much.

hoodmane (approved):
Thanks!

tools/update_cross_build_releases.py
ryanking13: Sure, my goal is not to change that value as much as possible, but if that happens, let's move it to the Makefile.

ryanking13:
Thank you for the review!

PR #4866, opened 2024-06-17
Vendor error-stack-parser

### LINKED ISSUE #4577: loadPyodide with indexURL tries to load stackframe.js from current url
## 🐛 Bug

Hello 👋 

I'm using pyodide (v0.25.0) in my NextJS project and have the following setup

- Copied the necessary files into the public folder
image

- Used the [pyodide-webpack-plugin](https://github.com/pyodide/pyodide-webpack-plugin) to take care of the unsupported loader issues.
image

- loaded pyodide from the public folder (used a custom hook for it)
```javascript
export const usePyodide = () => {
  const [pyodide, setPyodide] = useState<any>(null);

  useEffect(() => {
    const initializePyodide = async () => {
        indexURL: `/pyodide`,
      });
      const micropip = pyodide.pyimport('micropip');
      await micropip.install('black');
      setPyodide(pyodide);
    };

    initializePyodide();
  }, []);

  return pyodide;
};
```

On run I'm getting this stacktrace.js error
image

which is caused because it's trying to find it under the current full path which is `/web/build`. It cannot find it there ofc, so it's getting back the 404 page which it cannot parse, resulting in the exception.

image

The behaviour is completely clear to me but I'm not sure what I can do about it 🤔 I was under the impression that by setting the `indexURL` in the `loadPyodide` function I can tell pyodide where it can find all required files. And weirdly enough that also seems to be the case since the functionality itself is not broken? I can run python code just fine. In production this error is just 
"ignored" (printed to console but not breaking) but in dev mode it's more disruptive (see the error message screenshot).

### Expected behavior

Pyodide loading without the stacktrace.js error

### Environment

- Pyodide Version: 0.25.0
- Browser version: any browser
- Any other relevant information: NextJS

Any help would be appreciated.

Thank you and kind regards

### LINKED ISSUE #4863: Pyodide fails to load after require.js
## 🐛 Bug

When loading require.js before Pyodide, calling `loadPyodide` results in an error: `Uncaught (in promise) TypeError: z.default.parse is not a function`

This corresponds to https://github.com/pyodide/pyodide/blob/f1acd57a54aaeaf92cbb4e3a5d8768c4e25ccc83/src/js/compat.ts#L263

No error occurs if Pyodide is loaded first (at least not apparently).

### To Reproduce

It's pretty much the [first example in the docs](https://pyodide.org/en/stable/usage/quickstart.html#complete-example) but with `requirejs` loaded ahead of pyodide. I'm loading pyodide in a module as that helps track the source but the error is the same even without modules.

```html
<!doctype html>
<html>

<head>
  <script src="https://cdn.jsdelivr.net/npm/requirejs@2.3.6/require.min.js"></script>

  <script type="module">
    import { loadPyodide } from "https://cdn.jsdelivr.net/pyodide/v0.26.1/full/pyodide.mjs"

    async function main() {
      console.log(pyodide.runPython(`
              import sys
              sys.version
          `));
      let output = pyodide.runPython("1 + 2");
      document.getElementById("py-out").innerText = output;
    }
    main().catch((e) => {
      document.getElementById("py-out").innerText = "Error";
    });

    // export main so it can be tested in the dev console
    globalThis.main = main;
  </script>
</head>

<body>
  <div id="py-out">Running...</div>
</body>

</html>
```

### Expected behavior

Pyodide should load 

### Environment

- Pyodide Version 0.26.1
- Browser version Firefox 127.0/Linux and Chromium 126.0.6478.55/Linux

Platform: Arch Linux

### Additional context

I tested this with each version of Pyodide from v0.21.0 onward. The latest version of Pyodide which is compatible seems to be 0.23.4

ryanking13:
### Description

Resolve #4863
Resolve #4577

Vendors patched `error-stack-parser`, and its dependency `stackframe` into pyodide JS module, which will hopefully solve CommonJS related described in https://github.com/pyodide/pyodide/issues/4863#issuecomment-2169040888.

These libs are quite small and haven't been updated for more than 2 years, so I believe the maintenance burden of vendoring these libs would not be that big.

### Checklists

- [x] Add / update tests


ryanking13:
This also slightly (~2KB) decreases the size of pyodide.js (though it is not very important), as I removed some legacy codes (opera <= 11 support).

before 

```
16544 pyodide.js
16012 pyodide.mjs
```

after

```
14613 pyodide.js
13631 pyodide.mjs
```

ryanking13:
Thanks for the review!

PR #4867, opened 2024-06-17
Fix handling of SystemExit

### LINKED ISSUE #4864: Regression in 0.26.* - exit() crashes pyodide (and generates a different exception)
## 🐛 Bug

In 0.26.* - exit() - instead of raising a PythonError of type SystemExit as before - seems to raise a different unexpected ExitStatus object.
Furthermore (and more importantly), pyodide crashes shortly afterwards.
("Pyodide has suffered a fatal error. Please report this to the Pyodide maintainers.")

### To Reproduce

You can see the crash by just executing "exit()" in https://pyodide.org/en/stable/console.html

(And if you catch the exception in javascript and console.log it - you will see it's no longer a PythonError as in 0.25.0)

### Expected behavior

exit() should raise a PythonError of type SystemExit, which should not prevent pyodide from running again in the future.

### Environment

- Pyodide Version: 0.26.0 & 1:
- Browser version: Multiple :

hoodmane:
Resolves #4864. This fixes a regression that I introduced in #4567 which is that if a `SystemExit` crosses the FFI into JavaScript, we shut down Pyodide since then. I'm actually not sure how the SystemExit was working in the CLI runner before #4567, but the relevant `test_sys_exit` test covers the behavior so I guess it worked fine...

- [x] Add / update tests


src/js/pyodide.ts
```diff
@@ -202,13 +218,17 @@ export async function loadPyodide(
...
+    _pythonInspect: 1,
     packages: [],
     enableRunUntilComplete: false,
   };
   const config = Object.assign(default_config, options) as ConfigType;
   if (!config.env.HOME) {
     config.env.HOME = "/home/pyodide";
   }
+  if (config._pythonInspect) {
+    config.env.PYTHONINSPECT = "1";
```
ryanking13, line 230: How about using `PYTHONINSPECT` env variable directly instead of using a separate `_pythonInspect` parameter?

For instance:

```
config.env.PYTHONINSPECT = config.env.PYTHONINSPECT ?? "1"
```

ryanking13 (approved):
Thanks for fixing this! I have a minor comment, otherwise looks good to me.

src/js/pyodide.ts
hoodmane: Sounds good.

hoodmane: This will make the CLI runner handle the user setting `PYTHONINSPECT` too.

PR #4869, opened 2024-06-17
Fix #4861

### LINKED ISSUE #4861: PyProxy seems to have strange behavior in `0.26.1`
## 🐛 Bug

The description is a bit complicated, I recommend jump into the [stackblitz reproduction](https://stackblitz.com/edit/pyodide-issue-demo?file=src%2FMain.svelte).

I am using pyodide in a few complicated code-gen like usage, which need to exec dynamic python code from JavaScript. These code works well from pyodide `0.25.0` to `0.26.0`, but broke in `0.26.1`. The error message is strange and I can't find why. I looked into the diff between tag 0.26.0 and 0.26.1 but failed to tell which change caused this.

After simplifying, here is the minimal reproduceable example.

First, I execute code in a `ChainMap` for not polluting the global namespace, sounds reasonable, right? Since `eval`'s `globals` must be a `dict`, I mixined `ChainMap` with `dict`:

```py
class ChainMap(ChainMap, dict): ...
```

And I define a pair of function like this:

```py
async def f():
    return a  # not defined, but will be provided directly through the gloabls

async def g(context: dict[str, Any]):  # accept a readonly globals namespace
    return await eval(f.__code__, ChainMap({}, context))
```

Apparently, running `g({"a": 123})` should returns `123`. And I wraps this in `runPythonAsync` to get the PyProxy of this function. Then I create another function using the same approach.

<details><summary>complete code (can be directly paste into pyodide demo web console)</summary>
<p>

```ts
pyodide.setDebug(false); // modify this to get a different error message

pyodide.runPython(`
  import collections

  class ChainMap(collections.ChainMap, dict):
      """ to support layered eval context """
`);

async function child(context) {
  const func = await pyodide.runPythonAsync(`

      async def f():
          return a  # shoud be in global

      async def g(context: dict):
          assert isinstance(context, dict)
          return await eval(f.__code__, ChainMap({}, context))

      g
  `);
  return func(pyodide.toPy(context));
}

async function parent(context) {
  const func = await pyodide.runPythonAsync(`

      async def f():
          return await child(locals())

      async def g(context: dict):
          assert isinstance(context, dict)
          return await eval(f.__code__, ChainMap({}, context))

      g
  `);
  return await func(pyodide.toPy(context));
}

const context = { parent, child, a: 123 };

await parent(context) // the result should be `123`, but it raised a strange error
```

</p>
</details>

### To Reproduce

Since the use case is a little bit complicated, a minimal reproduction may still be a little bit long. So I made a dedicated [stackblitz reproduction](https://stackblitz.com/edit/pyodide-issue-demo?file=src%2FMain.svelte), which is a responsive application(when you change the code, the result will update too).

Open the link and you may see a red traceback (the loading process should be fast, normally in 1 minute). But if Ctrl+C kill the process and run:

```sh
pnpm i pyodide@0.26.0 && pnpm dev
```

The result should became a green `123`. This indicates that at least the same code is runnable in pyodide `0.26.0`

### Expected behavior

A green `123` instead of red traceback.

### Environment

- Pyodide Version: `0.26.1`
- Browser version: `Edge 125.0.2535.92`

### Additional context

Besides using `0.26.0`, I found that simply change all the async function to sync will solve this. (displaying a green 123) I am really confused. Already investing on this for a few days.

hoodmane:
and save some work in `wrap_promise`. This was how I should have written it to begin with. But I have no idea why this fixes #4861.

- [x] Add / update tests


ryanking13 (approved):
So... this just rearranges when `future_helper_mod` is initialized? I have no idea why it fixes the bug, but it seems to works  🤷

src/core/jsproxy.c
```diff
@@ -2650,11 +2650,9 @@ wrap_promise(JsVal promise, JsVal done_callback, PyObject* js2py_converter)
   result = _PyObject_CallMethodIdNoArgs(loop, &PyId_create_future);
   FAIL_IF_NULL(result);
 
-  helpers_mod = PyImport_ImportModule("_pyodide._future_helper");
-  FAIL_IF_NULL(helpers_mod);
```
CNSeniorious000, line 2654: Looks like this import always fails in our test case. Maybe some proxy got early destroyed so that we fail to import here?

hoodmane: It's not about a destroyed proxy. You can tell because if you change it to pass `__builtins__` it works:
```
    def g(x):
        import builtins
        return eval("x()", ChainMap({}, {"x": x, "__builtins__" : builtins}))
```

CNSeniorious000: I found the reason. In `PyImport_ImportModule`, it calls `__builtins__.__import__` internally.

When you are `await`ing a promis, you called `PyImport_ImportModule`, but the context don't have a `__builtins__`. This is reason.

hoodmane: But why does `__builtins__` not get assigned to the globals? Whenever you eval with a dict as globals, Python will assign a `__builtins__` key if it's missing.

CNSeniorious000: Really interesting. In https://github.com/python/cpython/blob/3.12/Python/pythonrun.c#L1712-L1716 they makes a mistake similar as https://github.com/pyodide/pyodide/issues/4636#issuecomment-2017707331

PR #4870, opened 2024-06-18
NFC minor code cleanup in dynload.ts

### LINKED ISSUE #4847: memoize in loadDynlibsFromPackage has no effect
Hi,

I was looking at this line:
 * https://github.com/pyodide/pyodide/blame/807250e8d22aa92213f9ad3e20648f109d079582/src/js/dynload.ts#L161

And it seems to have no effect, because the memorize creates a closure every time the function gets called. I guess it should move to the global namespace. I don't know if this is relevant for performance.

hoodmane:
We check the state of `pkg.shared_library` when it already must be false because of previous conditional. I also tidied up a bit and added some comments.

Resolves #4847.

### DIFF
--- a/src/js/dynload.ts
+++ b/src/js/dynload.ts
@@ -157,41 +157,41 @@ export async function loadDynlibsFromPackage(
     pkg.file_name.split("-")[0]
   }.libs`;
 
-  // This prevents from reading large libraries multiple times.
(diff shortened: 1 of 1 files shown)


ryanking13:
Thanks! I really need to investigate this global/local symbol someday.

PR #4871, opened 2024-06-18
 Fix dynamic library loading in cli runner

### LINKED ISSUE #3865: Issue when running some scipy.sparse.linalg code inside Pyodide venv
## 🐛 Bug

Running Scipy code inside a Pyodide venv yields to somme `Error: Dynamic linking error: cannot resolve symbol s_cmp`. Running inside a browser works fine and I am pretty sure running a .js with node explicitly does work as well since I have been doing it for a while in https://github.com/lesteve/scipy-tests-pyodide. Maybe the issue is with the Pyodide venv python nodejs wrapper?

### To Reproduce

```
pyodide venv .pyodide-venv
source .pyodide-venv/bin/activate
pip install scipy
python test.py
```

Where `test.py` has the following content:
```py
import numpy as np

from scipy.sparse.linalg import svds

rng = np.random.default_rng(0)
A = rng.random((10, 10))

res = svds(A, k=3, which='LM', random_state=0)
```

```
  File "/home/lesteve/test/.pyodide-venv/lib/python3.12/site-packages/scipy/sparse/linalg/_eigen/arpack/arpack.py", line 537 in iterate
  File "/home/lesteve/test/.pyodide-venv/lib/python3.12/site-packages/scipy/sparse/linalg/_eigen/arpack/arpack.py", line 1700 in eigsh
  File "/home/lesteve/test/.pyodide-venv/lib/python3.12/site-packages/scipy/sparse/linalg/_eigen/_svds.py", line 525 in svds
  File "/home/lesteve/test/test.py", line 8 in <module>
Error: Dynamic linking error: cannot resolve symbol s_cmp
    at stubs.<computed> (/home/lesteve/test/.pyodide-xbuildenv-0.26.0/0.26.0/xbuildenv/pyodide-root/dist/pyodide.asm.js:10:148862)
    at wasm://wasm/000fb3c2:wasm-function[207]:0x173c0
    at wasm://wasm/000fb3c2:wasm-function[180]:0x4472
    at wasm://wasm/0003e6d6:wasm-function[107]:0xbfa0
    at _PyEM_TrampolineCall_JS (/home/lesteve/test/.pyodide-xbuildenv-0.26.0/0.26.0/xbuildenv/pyodide-root/dist/pyodide.asm.js:10:125866)
    at wasm://wasm/0267d6e2:wasm-function[1134]:0x1c2db7
    at wasm://wasm/0267d6e2:wasm-function[1143]:0x1c35c8
    at wasm://wasm/0267d6e2:wasm-function[3408]:0x2a05c2
    at wasm://wasm/0267d6e2:wasm-function[3406]:0x2960a0
    at wasm://wasm/0267d6e2:wasm-function[4211]:0x30a9c5
```

Edit: for compleness as I was mentioning, the same code run with node directly `node test.js` does work:

```js
const { opendir } = require('node:fs/promises');

async function main() {
  
  await pyodide.loadPackage(["scipy"]);
  import numpy as np
  
  from scipy.sparse.linalg import svds
  
  rng = np.random.default_rng(0)
  A = rng.random((10, 10))
  
  res = svds(A, k=3, which='LM', random_state=0)
  `);

}

main();
```

### Expected behavior

No error

### Environment

- Pyodide Version: 0.26.0

hoodmane:
All dependent libs should be loaded as globals. We should try  to use this to simplify some of the logic in dynload.ts. Most likely what needs to happen is we need to inspect the wheel at build time to determine which libraries are "leaves" that are not dependencies of anything else and add this list to the package metadata.

Resolves #3865.

- [x] Update changelog
- [x] Add a test

### DIFF
--- a/docs/project/changelog.md
+++ b/docs/project/changelog.md
@@ -18,6 +18,11 @@ myst:
 
 - {{ Fix }} Don't leak the values in a dictionary when applying `to_js` to it.
   {pr}`4853`
+
+- {{ Fix }} Loading of dynamic libraries now works slightly better in the cli
+  runner. Resolved {issue}`3865`.
+  {pr}`4871`
+
 - {{ Enhancement }} Added implementation to abort `pyfetch` and `FetchResponse`
   manually or automatically.
   {pr}`4846`
--- a/emsdk/patches/0001-Add-back-fs.findObject-and-fs.readFile-in-loadLibDat.patch
+++ b/emsdk/patches/0001-Add-back-fs.findObject-and-fs.readFile-in-loadLibDat.patch
@@ -1,7 +1,7 @@
 From 137dae4dbdf9a192551582cdae827b085510956f Mon Sep 17 00:00:00 2001
 From: Hood Chatham <roberthoodchatham@gmail.com>
 Date: Fri, 2 Jun 2023 11:59:32 -0700
-Subject: [PATCH 1/5] Add back fs.findObject and fs.readFile in loadLibData
+Subject: [PATCH 1/6] Add back fs.findObject and fs.readFile in loadLibData
 
 See upstream PR:
 https://github.com/emscripten-core/emscripten/pull/19513
--- a/emsdk/patches/0002-Add-useful-error-when-symbol-resolution-fails.patch
+++ b/emsdk/patches/0002-Add-useful-error-when-symbol-resolution-fails.patch
@@ -1,7 +1,7 @@
 From e0cb884277200310eba263dcce5a7b1c4567bae6 Mon Sep 17 00:00:00 2001
 From: Hood Chatham <roberthoodchatham@gmail.com>
 Date: Fri, 19 May 2023 12:19:00 -0700
-Subject: [PATCH 2/5] Add useful error when symbol resolution fails
+Subject: [PATCH 2/6] Add useful error when symbol resolution fails
 
 Currently if symbol resolution fails, we get:
 ```js
--- a/emsdk/patches/0003-Changes-for-JSPI.patch
+++ b/emsdk/patches/0003-Changes-for-JSPI.patch
@@ -1,7 +1,7 @@
 From d4d5d35072b6d28c5f98b77a0d578e01bd25ef7e Mon Sep 17 00:00:00 2001
 From: Hood Chatham <roberthoodchatham@gmail.com>
 Date: Thu, 22 Jun 2023 18:53:22 -0700
-Subject: [PATCH 3/5] Changes for JSPI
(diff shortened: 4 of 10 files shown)


agriyakhetarpal:
Thanks, @hoodmane! I tried this with `statsmodels` here to see if this resolves the issue there – it did not, sadly: https://github.com/agriyakhetarpal/statsmodels/actions/runs/9615408856/job/26522486623#step:3:6991. The missing symbol is `pow_dd` (though I earlier faced an error for `pow_di`). It's great that this works for `s_cmp`, though...

hoodmane:
Perhaps there are several bugs =( could you give a minimal reproduction of the stats models problem? I'll look into it next.

agriyakhetarpal:
Yes, I shall open a new issue for this – I'll have to investigate how to provide a minimal enough reproducer, please hold on :)

hoodmane:
It doesn't have to be that small, if you can give me a single file that does it that would be good enough for me.

hoodmane:
Also @agriyakhetarpal can you check it against #4876? Since that changes this same code again.

ryanking13:
`pow_dd` symbol comes from openblas.so. @agriyakhetarpal Could you share us wheels that are used when testing statsmodels (including scipy, numpy)? Probably some of them are not linked correctly.

agriyakhetarpal:
Sorry for the late response over here, @ryanking13 and @hoodmane, but I did spend some time on this a while ago – and the problem has now subsided with either #5011 or #5031 having been merged. I have no issues when running the `statsmodels` tests using the Pyodide xbuildenv. Once we have 0.27 stable out and `cibuildwheel` updates to it internally, we can look at testing WASM wheels for `statsmodels` without the Node.js runner/script. The `pow_dd` unresolved symbol came from SciPy 1.12.0 (from a `scipy.special` shared object that I can't recall the name of while typing this comment), which was breaking the `import statsmodels` statement itself. I'll update the recipe for `statsmodels` to include a patch based on statsmodels/statsmodels#9270, since it also fixes a few import errors a bit deeper into some modules, besides fixing the tests. I will take a glance at `scikit-image`, next. :)

Side note: we should really have an option to use any Pyodide version (>0.22, perhaps) in `cibuildwheel` using an environment variable or Pyodide-specific configuration for it in `pyproject.toml` in `[tool.cibuildwheel.pyodide]`.

agriyakhetarpal:
And `scikit-image` works, too! PR linked above. In this one, unlike for `statsmodels`, I haven't checked which shared object in SciPy was the cause of the trouble, but I can do so if someone wants me to.

PR #4877, opened 2024-06-20
Fix biopython build

hoodmane:
@jaraco looks like pypa/setuptools#4369 may have caused a regression here?

Upstream PR: https://github.com/biopython/biopython/pull/4749

### DIFF
--- a/packages/biopython/extras/pyproject.toml
+++ b/packages/biopython/extras/pyproject.toml
@@ -1,3 +1,0 @@
-[build-system]
-requires = ["setuptools != 70.1"]
-build-backend = "setuptools.build_meta"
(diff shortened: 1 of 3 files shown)


ryanking13:
Thanks for the fix!

jaraco:
> @jaraco looks like [pypa/setuptools#4369](https://github.com/pypa/setuptools/pull/4369) may have caused a regression here?

Yes, that change did change the behavior, but in my estimation, biopython was depending on incidental details of the deployment/ecosystem, details that needed to change. Sorry for the inconvenience. Do let me know if you think there's something more that Setuptools should have done in making the transition.

PR #4876, opened 2024-06-20
Load shared libraries locally

ryanking13:
### Description

Fixes issues with symbol resolution and changes the behavior of loading shared libraries globally to local.

I found several libraries misconfigured while working on this: not having correct dependencies, so this PR fixes them as well.

### Checklists

- [x] Add a [CHANGELOG](https://github.com/pyodide/pyodide/blob/main/docs/project/changelog.md) entry

PR #4878, opened 2024-06-21
Fix recipes that do not link to their dependencies correctly

ryanking13:
### Description

Splitted off from #4876.

What this PR does is basically fixing incorrect shared library dependencies.

For instance, `cryptography` package depends on `openssl`, so the python extension should link to `libssl.so`.

However, before this PR, it was not linked correctly.

```sh
$  pyodide auditwheel show cryptography-42.0.5-cp312-cp312-pyodide_2024_0_wasm32.whl
{'cryptography\\hazmat\\bindings\\_rust.cpython-312-wasm32-emscripten.so': []}
```

After this PR, it is linked correctly.

```bash
$ pyodide auditwheel show cryptography-42.0.5-cp312-cp312-pyodide_2024_0_wasm32.whl
{
│   'cryptography\\hazmat\\bindings\\_rust.cpython-312-wasm32-emscripten.so': [
│   │   'libssl.so',
│   │   'libcrypto.so'
│   ]
}
```

__Why do we need this change?__

Because it is how shared libraries should work.

__Why was this working before?__

Because we loaded dependencies globally. `openssl` is loaded globally, so its symbols were visible to `cryptography` even if they are not linked. However, #4876 will change all library loading mechanism to local, so all shared libraries should now have correct dependencies to work.

### DIFF
--- a/packages/openssl/meta.yaml
+++ b/packages/openssl/meta.yaml
@@ -1,6 +1,6 @@
 package:
   name: openssl
-  version: 1.1.1n
+  version: 1.1.1w
   tag:
     - library
 source:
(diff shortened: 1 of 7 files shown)


packages/openssl/meta.yaml
```diff
@@ -1,6 +1,6 @@
 package:
   name: openssl
-  version: 1.1.1n
+  version: 1.1.1w
```
ryanking13, line 3: This is not version update, the version number was incorrect.

hoodmane (approved):
Thanks!

PR #4882, opened 2024-06-23
Unvendor pyodide-build

ryanking13:
### Description

pyodide-build now lives in [pyodide/pyodide-build](https://github.com/pyodide/pyodide-build).

After this change,

1) pyodide-build will have separate version numbers from Pyodide runtime. Actually, I just released pyodide-build 0.27, so it is already separated
2) If some change needs to be applied both in pyodide runtime and pyodide-build, it will be a little more uncomfortable. One should bump up pyodide-build first, and then update pyodide runtime. I've worked to loosen the connection between the two as much as possible to prevent this from happening, but as always, it can happen.


ryanking13:
This is ready to be reviewed now.

.circleci/config.yml
```diff
@@ -254,7 +254,7 @@ jobs:
           name: Install requirements
           command: |
             pip install -r requirements.txt
-            pip install -e ./pyodide-build
+            pip install pyodide-build --upgrade
```
hoodmane, line 257: What happens if we want to test against a fork of `pyodide-build`? Can we add an empty constraints file called `pyodide-build-version.txt` and pass `-c pyodide-build-version.txt` in CI?

hoodmane: Perhaps even add `pyodide-build` to `requirements.txt` and do:
```
pip install -r requirements.txt -c pyodide-build-version.txt
```

.devcontainer/onCreate-conda.sh
```diff
@@ -28,9 +28,6 @@ ln -sf "$PYODIDE_RECIPE_BUILD_DIR" packages/.build || echo "Note: Could not crea
 
 conda run -n pyodide-env --live-stream pip install -r requirements.txt
 
-# https://pyodide.org/en/stable/development/new-packages.html#prerequisites
-conda run -n pyodide-env --live-stream pip install -e ./pyodide-build
```
hoodmane, line 32: Presumably we need to add `pyodide-build` to `requirements.txt` for this to make sense...

docs/development/new-packages.md
```diff
@@ -75,12 +75,6 @@ This will mount the current working directory as `/src` within the container so
 if you build the package within the container the files created will persist in
 the directory after you exit the container.
 
-You should install `pyodide-build`:
-
-```bash
-pip install -e ./pyodide-build
```
hoodmane, line 81: Isn't this still the case?

Makefile.envs
```diff
@@ -2,6 +2,7 @@ export PYVERSION ?= 3.12.1
 export PYODIDE_EMSCRIPTEN_VERSION ?= 3.1.58
 export PYODIDE_VERSION ?= 0.27.0.dev0
 export PYODIDE_ABI_VERSION ?= 2024_0
+export PYODIDE_BUILD_VERSION ?= 0.27.3
```
hoodmane, line 5: I wonder if we should add optional `PYODIDE_BUILD_REPO` and `PYODIDE_BUILD_COMMIT` -- if these are defined, we can check in the `pyodide_build` rule that `direct_url.json` exists and has appropriate contents. Then perhaps we should install `pyodide_build` everywhere with `make pyodide_build`...?

.circleci/config.yml
ryanking13: That sounds reasonable to me.

.github/workflows/main.yml
```diff
@@ -9,45 +9,6 @@ concurrency:
...
-          python3 -m pip install -r requirements.txt -r requirements-deploy.txt
-      - name: Run tests
-        shell: bash -l {0}
-        run: |
-          PYODIDE_ROOT=. pytest \
-              --junitxml=test-results/junit.xml \
-              --verbose \
-              --runtime=host \
-              --cov=pyodide_build --cov=pyodide \
-              src pyodide-build packages/micropip/ packages/_tests tools/
```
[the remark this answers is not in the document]
ryanking13: Oh, I thought this job is testing only pyodide-build. Wil revert the change.

Makefile.envs
ryanking13: Sounds reasonable to me. Let me see if I can make if work.

.circleci/config.yml
ryanking13: Currently, the requirements.txt file is used to specify test dependencies, so it doesn't seem like a good idea to put build dependencies in there. Let me think of a better way.

.devcontainer/onCreate-conda.sh
ryanking13: Changed to `make pyodide_build`.

docs/development/maintainers.md
```diff
@@ -191,6 +191,11 @@ to update all packages and make a pull request with these changes. There will be
 build/test failures, revert the packages that fail the build or tests and make a
 note to update them independently.
 
+## Updating pyodide-build
+
+You can update the pyodide-build vession (or test with a different version / branch of pyodide-build)
+by changing the variables in `tools/check_and_instal_pyodide_build.py`.
```
hoodmane, line 197:
```suggestion
by changing the variables in `tools/check_and_install_pyodide_build.py`.
```

tools/check_and_install_pyodide_build.py
```diff
@@ -0,0 +1,56 @@
...
+import sys
+
+# Note for maintainers:
+#   Update the following variables when you want to update the version of the
+#   pyodide-build version used in the build process.
+#   If you want to make breaking changes in pyodide-build, it is also useful to
+#   change the repository URL to your fork to test the changes are working as
+#   expected.
+PYODIDE_BUILD_REPO: str = "https://github.com/pyodide/pyodide-build"
+PYODIDE_BUILD_COMMIT: str = "fac0109aa2acf14469320b049d710dd42639bf94"  # v0.27.3
```
hoodmane, line 12: I'd prefer if we could put this in `Makefile.envs` or failing that into `pyproject.toml`. I don't want to add more locations where configuration lives if we can help it.

ryanking13: Makes sense. I moved it to Makefile.envs.

.github/workflows/main.yml
```diff
@@ -43,7 +41,7 @@ jobs:
               --verbose \
               --runtime=host \
               --cov=pyodide_build --cov=pyodide \
-              src pyodide-build packages/micropip/ packages/_tests tools/
```
hoodmane, line 46: Is there a reason you deleted `packages/micropip` and `packages/_tests` here?

hoodmane (approved):
Thanks!

.github/workflows/main.yml
ryanking13: I put back `packages/_tests`. `packages/micropip` is not needed as there is no host test in micropip anyway.

PR #4889, opened 2024-06-26
added unix timezone package

joemarshall:
### Description
This is a package which installs unix style timezone data. This means things that handle timezones in C/C++ code should then work. It is needed for pyarrow because annoyingly Arrow's datetime library doesn't support the format of timezone data which is used by the tzdata python package.

This makes a package `unix-timezone` which you run by doing `import unix_timezone` to install the data.

An alternative would be if we were to add support for a package type that was just files to install to the file system, but I wasn't confident about going deep into the package installer to make it happen (the files need to end up at `/usr/share/zoneinfo` so I couldn't just use a shared_library package for this.

### Checklists

- [X] Add / update tests


ryanking13:
Thanks @joemarshall.

As I commented in #4892, I would prefer supporting [`data_files`](https://setuptools.pypa.io/en/latest/deprecated/distutils/setupscript.html#installing-additional-files) instead of having a package to install the file themselves.

Also, could you upload the package to PyPI instead of embedding the source code in Pyodide repository? Personally, I don't want to add more `source/path` type packages that requires us to manage those packages ourselves unless it is inevitable.

joemarshall:
I think the only advantage to having a separate package to this (as opposed to it being in data_files of arrow) is that it is possible other packages will also want unix style timezone data? 

Will stick this package on pypi btw.

ryanking13:
> I think the only advantage to having a separate package to this (as opposed to it being in data_files of arrow) is that it is possible other packages will also want unix style timezone data?

Oh, I mean, I am +1 with having a package `unix-timezones`, as Arrow folks would not want to bundle timezone data into Arrow. What I meant to say was that I think it would be better to put the timezone files in the data_files of the unix_timezones package and let the pyodide package installer install it.

ryanking13 (approved):
Thanks @joemarshall! I didn't look inside the `pyodide_unix_timezones` wheel, but I believe you.

PR #4907, opened 2024-06-30
Added `checkAPIVersion` option to `loadPyodide`

### LINKED ISSUE #4906: Is it possible to bypass the version check?
## 🚀 Feature

Can we provide a `ignoreVersionCheck` option to bypass this?

https://github.com/pyodide/pyodide/blob/cf41eafe1665ecc4e0682ab43aa271a32b719127/src/js/pyodide.ts#L260-L267

### Motivation

I am making a stackblitz reproduction template for #4895, if I can bootstrap another version of pyodide with the same version of pyodide(npm), it could be easier.

In another way, there are use cases that we need to have a preview of nightly build of pyodide-core, which don't have a compatible npm package. So I think we can provide a way to manually ignore the version check here.

Using options field is an approach. I think using environment variables is reasonable too.

### Pitch

If `ignoreVersionCheck === true`, don't throw this error, and `console.warn` it perhaps.

### Alternatives

If `env.IGNORE_VERSION_CHECK`, don't throw this error, and `console.warn` it perhaps.

### Additional context

I think this check is because different version of pyodide may have different bootstrap patterns. So using a wrong pyodide npm package to bootstrap another version of pyodide core will have unexpected behaviors. But since we don't have a nightly release for every commit, this proposal may be an easiest approach.

CNSeniorious000:
### Description

Resolves #4906

### Checklists

### DIFF
--- a/src/js/pyodide.ts
+++ b/src/js/pyodide.ts
@@ -169,6 +169,11 @@ export async function loadPyodide(
      */
     enableRunUntilComplete?: boolean;
     /**
+     * If false (default), throw an error if the version of Pyodide core does not
+     * match the version of the Pyodide js package. Otherwise, just log a warning.
+     */
+    allowVersionMismatch?: boolean;
+    /**
      * Used by the cli runner. If we want to detect a virtual environment from
      * the host file system, it needs to be visible from when `main()` is
      * called. The directories in this list will be mounted at the same address
@@ -258,12 +263,15 @@ export async function loadPyodide(
   }
 
   if (API.version !== version) {
-    throw new Error(
-      `\
+    const message = `\
 Pyodide version does not match: '${version}' <==> '${API.version}'. \
 If you updated the Pyodide version, make sure you also updated the 'indexURL' parameter passed to loadPyodide.\
-`,
-    );
+`;
+    if (options.allowVersionMismatch) {
+      console.warn(message);
+    } else {
+      throw new Error(message);
+    }
   }
   // Disable further loading of Emscripten file_packager stuff.
   Module.locateFile = (path: string) => {
--- a/src/js/pyodide.ts
+++ b/src/js/pyodide.ts
@@ -172,7 +172,7 @@ export async function loadPyodide(
      * If false (default), throw an error if the version of Pyodide core does not
      * match the version of the Pyodide js package.
      */
-    skipVersionCheck?: boolean;
+    checkAPIVersion?: boolean;
     /**
      * Used by the cli runner. If we want to detect a virtual environment from
      * the host file system, it needs to be visible from when `main()` is
@@ -262,7 +262,7 @@ export async function loadPyodide(
     API.setPyProxyToStringMethod(true);
   }
 
-  if (API.version !== version && !options.skipVersionCheck) {
+  if (API.version !== version && !options.checkAPIVersion) {
     throw new Error(`\
 Pyodide version does not match: '${version}' <==> '${API.version}'. \
 If you updated the Pyodide version, make sure you also updated the 'indexURL' parameter passed to loadPyodide.\
--- a/src/js/pyodide.ts
+++ b/src/js/pyodide.ts
@@ -264,7 +264,7 @@ export async function loadPyodide(
     API.setPyProxyToStringMethod(true);
   }
 
-  if (API.version !== version && !config.checkAPIVersion) {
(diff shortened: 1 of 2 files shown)


src/js/pyodide.ts
```diff
@@ -258,12 +263,15 @@ export async function loadPyodide(
...
-    throw new Error(
-      `\
+    const message = `\
 Pyodide version does not match: '${version}' <==> '${API.version}'. \
 If you updated the Pyodide version, make sure you also updated the 'indexURL' parameter passed to loadPyodide.\
-`,
-    );
+`;
+    if (options.allowVersionMismatch) {
+      console.warn(message);
```
ryanking13, line 271: I think it is okay to not show warning in this case, as this is probably what the users intended.

ryanking13 (approved):
Thanks @CNSeniorious000!

src/js/pyodide.ts
CNSeniorious000: Ok I will remove this. And another thing, which name do you think is better. I've thought of several name and wonder which one do you think is the most appreciate one

- `allowVersionMismatch`
- `skipVersionCheck`
- `ignoreVersionMismatch`
- `suppressVersionMismatch`
– ...

Not having the knowledge of these words' difference.

Since this is an exposed API and may be better not to change so I'd like to be careful.

hoodmane: Maybe `skipVersionCheck`?

ryanking13: In `loadPackage`, we have [`checkIntegrity` option](https://github.com/pyodide/pyodide/blob/5fb36a458be37270efecdc9638202a13b62d28f8/src/js/load-package.ts#L283) with default value `true`. So for consistency, maybe we can use a positive term like `checkAPIVersion` and setting it to true by default.

CNSeniorious000: I agree with the naming. But we don't have a default value for `options` here:

https://github.com/pyodide/pyodide/blob/9428c6f69e997c59a094538c7638d3e8ebb66284/src/js/pyodide.ts#L191

CNSeniorious000: Sorry I just found that here is the default options:

https://github.com/pyodide/pyodide/blob/9428c6f69e997c59a094538c7638d3e8ebb66284/src/js/pyodide.ts#L200-L213

hoodmane: The missing ones default to `undefined` / `false`

src/js/pyodide.ts
```diff
@@ -168,6 +169,11 @@ export async function loadPyodide(
      * Make loop.run_until_complete() function correctly using stack switching
      */
     enableRunUntilComplete?: boolean;
+    /**
+     * If false (default), throw an error if the version of Pyodide core does not
+     * match the version of the Pyodide js package.
```
ryanking13, line 174: Could you update the default value and the comment correspondingly?

CNSeniorious000: I am not sure how to express it. Maybe could you suggest a change here?

ryanking13: I think you can add `checkApiVersion: true` to default_config.

```js
  const default_config = {
    fullStdLib: false,
    jsglobals: globalThis,
    stdin: globalThis.prompt ? globalThis.prompt : undefined,
    lockFileURL: indexURL + "pyodide-lock.json",
    args: [],
    _node_mounts: [],
    env: {},
    packageCacheDir: indexURL,
    packages: [],
    enableRunUntilComplete: false,
    checkAPIVersion: true,
  };
```

src/js/pyodide.ts
```diff
@@ -168,6 +169,11 @@ export async function loadPyodide(
      * Make loop.run_until_complete() function correctly using stack switching
      */
     enableRunUntilComplete?: boolean;
+    /**
+     * If false (default), throw an error if the version of Pyodide core does not
```
ryanking13, line 173:
```suggestion
     * If true (default), throw an error if the version of Pyodide core does not
```

and change the default value of the comment

src/js/pyodide.ts
CNSeniorious000: Oh I get it. I forgot that we have reversed the meaning of this options. My fault 🙏

ryanking13:
Thanks @CNSeniorious000!

PR #4916, opened 2024-07-06
Support `bun`

### LINKED ISSUE #4915: `IN_NODE` flag is `false` in `bun` runtime
## 🐛 Bug

[Bun](https://bun.sh/) is a nodejs-compatible JavaScript runtime, but pyodide don't treats it as nodejs.

### To Reproduce

When I try to import pyodide in `bun`, it raises an error.

```sh
npx bun repl
```

This will start a JavaScript console, then type:

```js
await import("pyodide")
```

### Expected behavior

```sh
> await import("pyodide")
4 | `)>-1&&r.message.split(`
5 | `).length>r.stacktrace.split(`
6 | `).length?this.parseOpera9(r):r.stack?this.parseOpera11(r):this.parseOpera10(r)},"ErrorStackParser$$parseOpera"),parseOpera9:f(function(r){for(var u=/Line (\d+).*script (?:in )?(\S+)/i,i=r.message.split(`
7 | `),c=[],s=2,m=i.length;s<m;s+=2){var p=u.exec(i[s]);p&&c.push(new e({fileName:p[2],lineNumber:p[1],source:i[s]}))}return c},"ErrorStackParser$$parseOpera9"),parseOpera10:f(function(r){for(var u=/Line (\d+).*script (?:in )?(\S+)(?:: In function (\S+))?$/i,i=r.stacktrace.split(`
8 | `),c=[],s=0,m=i.length;s<m;s+=2){var p=u.exec(i[s]);p&&c.push(new e({functionName:p[3]||void 0,fileName:p[2],lineNumber:p[1],source:i[s]}))}return c},"ErrorStackParser$$parseOpera10"),parseOpera11:f(function(r){var u=r.stack.split(`
9 | `).filter(function(i){return!!i.match(o)&&!i.match(/^Error created at/)},this);return u.map(function(i){var c=i.split("@"),s=this.extractLocation(c.pop()),m=c.shift()||"",p=m.replace(/<anonymous function(: (\w+))?>/,"$2").replace(/\([^)]*\)/g,"")||void 0,d;m.match(/\(([^)]*)\)/)&&(d=m.replace(/^[^(]+\(([^)]*)\)$/,"$1"));var y=d===void 0||d==="[arguments not available]"?void 0:d.split(",");return new e({functionName:p,args:y,fileName:s[0],lineNumber:s[1],columnNumber:s[2],source:i})},this)},"ErrorStackParser$$parseOpera11")}},"ErrorStackParser"))});var z=le($());var g=typeof process=="object"&&typeof process.versions=="object"&&typeof process.versions.node=="string"&&typeof process.browser>"u",F=g&&typeof module<"u"&&typeof module.exports<"u"&&typeof w<"u"&&typeof __dirname<"u",C=g&&!F,de=typeof Deno<"u",W=!g&&!de,j=W&&typeof window=="object"&&typeof document=="object"&&typeof document.createElement=="function"&&typeof sessionStorage=="object"&&typeof importScripts!="function",B=W&&typeof importScripts=="funct | ... truncated 

error: Cannot determine runtime environment
      at /workspace/reasonify/node_modules/pyodide/pyodide.mjs:9:2633
```

### Environment

- Pyodide Version: `0.26.1`

### Additional context

https://github.com/pyodide/pyodide/blob/8ba8f4497cf25d6ec876fb2703dc77afa5f93bae/src/js/environments.ts#L8-L9

But in `bun`, process.browser is `false` so the last condition become `false`.

I think we can treat `bun` as the same as nodejs.

CNSeniorious000:
At least `loadPyodide` works now. The next question is, do we need to run other tests on `bun` too? Could someone direct me to the location for modifications?

### Description

Resolves #4915 

### Checklists

- [ ] Add / update tests


src/js/compat.ts
```diff
@@ -268,6 +269,9 @@ export async function calculateDirname(): Promise<string> {
 
     // FIXME: We would like to use import.meta.url here,
     // but mocha seems to mess with compiling typescript files to ES6.
+    if (IN_BUN) {
+      return nodeUrl.fileURLToPath(`file://${nodePath.dirname(fileName)}`);
```
hoodmane, line 273: Does this not work in node? If we could always prepend with `file://` we could drop `IN_BUN`.

CNSeniorious000: Looks like the nodejs and bun are not the same here.

```
node --experimental-wasm-stack-switching makesnap.mjs
node:internal/url:796
    this.#updateContext(bindingUrl.parse(input, base));
                                   ^

TypeError: Invalid URL
    at new URL (node:internal/url:796:36)
    at Module.fileURLToPath (node:internal/url:1457:12)
    at calculateDirname (file:///workspaces/pyodide/dist/pyodide.mjs:3:3203)
    at async loadPyodide (file:///workspaces/pyodide/dist/pyodide.mjs:3:8740)
    at async file:///workspaces/pyodide/dist/makesnap.mjs:7:12 {
  code: 'ERR_INVALID_URL',
  input: 'file://file:///workspaces/pyodide/dist'
}

Node.js v20.12.2
make: *** [Makefile:241: dist/snapshot.bin] Error 1
```

src/js/environments.ts
```diff
@@ -5,8 +5,10 @@ export const IN_NODE =
   typeof process === "object" &&
   typeof process.versions === "object" &&
   typeof process.versions.node === "string" &&
-  typeof process.browser ===
-    "undefined"; /* This last condition checks if we run the browser shim of process */
+  !process.browser; /* This last condition checks if we run the browser shim of process */
+
+/** @private */
+export const IN_BUN = IN_NODE && process.title === "bun";
```
ryanking13, line 11: Could you add it to [`detectEnvironment`](https://github.com/pyodide/pyodide/blob/main/src/js/environments.ts#L53) as well for testing?

ryanking13 (commented):
Thanks @CNSeniorious000!

> At least loadPyodide works now. The next question is, do we need to run other tests on Bun too? Could someone direct me to the location for modifications? 

I think you can add a smoke test like we do for [Deno](https://github.com/pyodide/pyodide/tree/main/src/test-deno).

You may add a similar directory like deno for Bun, and add a [new GitHub Action workflow](https://github.com/pyodide/pyodide/blob/62e6694080eced2e3e3eed3a128505ebca63059a/.github/workflows/main.yml#L228-L254) that runs it.

I think it is okay not to run complete test suites against Bun unless we decide to officially support it (which is out of our current bandwidth). So a basic smoke test would be fine with me.

src/js/compat.ts
ryanking13: Hmm, this is strange. Node also throws an error if there is no `file://` prefix:

```node
> url.fileURLToPath('/asdf/asdf')
Uncaught TypeError [ERR_INVALID_URL]: Invalid URL
    at __node_internal_captureLargerStackTrace (node:internal/errors:484:5)
    at new NodeError (node:internal/errors:393:5)
    at URL.onParseError (node:internal/url:564:9)
    at new URL (node:internal/url:644:5)
    at Object.fileURLToPath (node:internal/url:1484:12) {
  input: '/asdf/asdf',
  code: 'ERR_INVALID_URL'
}
```

So, I wonder what the difference is between the two. Could you please check the values of

```
let fileName = ErrorStackParser.parse(err)[0].fileName!;
```

and
```
nodePath.dirname(fileName)
```

?

I wonder if `ErrorStackParser` is not working correctly with Bun's stack trace.

CNSeniorious000: I added a `console.log` here:

```js
    console.log({ fileName, dirname: nodePath.dirname(fileName) })
```

During `make` it prints:

```
...
[2024-07-11 11:15:07] done building packages...
cp src/templates/console.html dist/console.html
sed -i -e 's#{{ PYODIDE_BASE_URL }}#./#g' dist/console.html
cd dist && node --experimental-wasm-stack-switching makesnap.mjs
{
  fileName: 'file:///workspaces/pyodide/dist/pyodide.mjs',
  dirname: 'file:///workspaces/pyodide/dist'
}
```

CNSeniorious000:
@ryanking13 I removed usage of `IN_BUN` and ensure `fileName` is prefixed with `file://` through this way:

```js
  if (!fileName.startsWith("file://")) {
    fileName = `file://${fileName}`; // Error stack filenames are not starting with `file://` in `Bun`
  }
```

Do you think it is OK? If so, do you think we still need `IN_BUN` in `environment.ts`?

If not, do you think this is a bug of `ErrorStackParser`?

src/js/environments.ts
```diff
@@ -19,6 +18,9 @@ export const IN_NODE_COMMONJS =
 /** @private */
 export const IN_NODE_ESM = IN_NODE && !IN_NODE_COMMONJS;
 
+/** @private */
+export const IN_BUN = IN_NODE && process.title === "bun";
```
Jarred-Sumner, line 22:
```suggestion
export const IN_BUN = typeof globalThis.Bun !== "undefined";
```

CNSeniorious000: Thanks @Jarred-Sumner!

CNSeniorious000:
Looks like [this](https://github.com/pyodide/pyodide/actions/runs/9899930560/job/27350054786?pr=4916) is just an accidental failure because [the same workflow in my forked repo ran successful](https://github.com/CNSeniorious000/pyodide/actions/runs/9899930280/job/27350050084).

```
Run bun smoke-test.js
  bun smoke-test.js
  shell: /usr/bin/bash -e {0}
RuntimeError: Out of bounds memory access (evaluating 'wasmTable.get(func)()')
      at WASM  ([wasm code])
      at WASM  ([wasm code])
      at WASM  ([wasm code])
      at WASM  ([wasm code])
      at WASM  ([wasm code])
      at WASM  ([wasm code])
      at WASM  ([wasm code])
      at WASM  ([wasm code])
      at WASM  ([wasm code])
      at WASM  ([wasm code])

Bun v1.1.18 (Linux x6[4](https://github.com/pyodide/pyodide/actions/runs/9899930560/job/27350054786?pr=4916#step:6:5) baseline)
Error: Process completed with exit code 1.
```

ryanking13 (approved):
Thanks @CNSeniorious000!

> Looks like [this](https://github.com/pyodide/pyodide/actions/runs/9899930560/job/27350054786?pr=4916) is just an accidental failure because [the same workflow in my forked repo ran successful](https://github.com/CNSeniorious000/pyodide/actions/runs/9899930280/job/27350050084).

The error message might be worth looking into, but for now it looks good as the CI passes.

PR #4917, opened 2024-07-06
`np.NaN` was deprecated in the NumPy 2.0 release. Use `np.nan` instead.

CNSeniorious000:
Numpy 2.0 deprecated many constants including `np.NaN`

### Description

This error is giving every PR check a red mark 🙄.

### Checklists


hoodmane:
Thanks!

Issue #4914, opened 2024-07-05
A single, public, Python API for installing package


ryanking13:
### Proposal

I would like to make a single, public, Python, API for installing a package. Here, "install" does not involve loading the package from an external URL or doing dependency resolution. Instead, it means unpacking the package into the proper directory and doing some post installation steps.

This API will be used inside `pyodide.loadPackage,` `micropip.install`, and can be used in other downstream projects as well.

Currently, we are using [`pyodide._package_loader.unpack_buffer`](https://github.com/pyodide/pyodide/blob/8ba8f4497cf25d6ec876fb2703dc77afa5f93bae/src/py/pyodide/_package_loader.py#L143) for a similar purpose, but it has following limitations.

1) It is a private API
2) It does not cover loading shared libraries
3) It's name isn't that straightforward.

So I would like to make a new API that would have signature such as:

```python
async install_package(
  buffer: bytes | JsBuffer,
  format: ...
  extract_dir: ...
  load_dynlibs: ...
  load_datafiles: ...
  <something else>
):
  # handles installing a package
  # A package can be a
  # 1. Python wheel
  # 2. Other pyodide-supported types (shared library in a .zip format)
  # 3. tar or zip archive?

  # This function should also handle
  # - loading shared libraries
  # - installing data_files
  # - etc
```

### Motivation

I would like to make `pyodide.loadPackage` and `micropip.install` use the same API when installing a package. For now, they are not shared very well. Therefore, for instance, if I want to add `data_files` support in both `pyodide.loadPackage` and `micropip.install`, I need to implement it separately in `pyodide._package_loader` and in `micropip.WheelInfo`, which is quite annoying.

hoodmane:
Sounds good to me!

hoodmane:
For stack switching, it would be nice if we could implement this in JavaScript. I think it's possible with `decompressionStream` at this point.

ryanking13:
> For stack switching, it would be nice if we could implement this in JavaScript. I think it's possible with decompressionStream at this point.

I see. Changing the top-level entry point to JavaScript seems fine to me. It'll probably use information from both Python and JS internally anyway

ShaheedHaque:
Worth considering #4596?

hoodmane:
> It'll probably use information from both Python and JS internally anyway

Ideally for memory snapshots I think it should be JS only. Though perhaps this means that I should be snapshoting the entire file system so that the package loader is not needed at all to initialize packages.

[...]

PR #4902, opened 2024-06-29
Package unvendored CPython stdlibs in wheel format

ryanking13:
### Description

Related: https://github.com/pyodide/pyodide/issues/4748

Packages all unvendored cpython modules in a wheel format, instead of using a custom zipfile.  Unvendored stdlibs are considered as a normal python package and installed in site-packages directory.

The package_type == 'cpython_module' still remains to distinguish it from normal packages (we are using it for custom import hook and in `fullstdlib` flag), but except for that, it works just like a normal python package.

__Possible Drawbacks__

1. Unvendored stdlibs now contain wheel metadata files. So packages like `importlib` wheel grap those metadata and consider it like a normal python package.
2. The installation directory for stdlibs is changed, previously `/lib/python3.12/`, now `/lib/python3.12/site-packages`.

### Checklists

- [x] Add a [CHANGELOG](https://github.com/pyodide/pyodide/blob/main/docs/project/changelog.md) entry
- [x] Add new / update outdated documentation
- [x] fix tests

PR #4922, opened 2024-07-08
MAINT cleanup `unpack_buffer()`

ryanking13:
### Description

Cleans up pyodide._package_loader.unpack_buffer a bit, NFC. This is a preparation for #4914.

1. Remove `stdlib` target, which no longer exists (#4902 removed it).
2. Remove the `target` parameter from unpack_buffer and separate it into a separate function `get_install_dir`.


hoodmane (approved):
Thanks!

PR #4923, opened 2024-07-09
MAINT Move common JS methods into a subdirectory

ryanking13:
### Description

Relocates some common JS methods that do not require Emscripten access into a sub directory, and adds unit tests for them. NFC.

### Checklists

- [x] Add / update tests


src/core/pyproxy.ts
```diff
@@ -18,7 +18,6 @@ declare var Tests: any;
 declare var Module: any;
 
 import { TypedArray } from "types";
-import { warnOnce } from "pyodide-util";
```
ryanking13, line 21: This function was not used.

src/js/package.json
```diff
@@ -31,6 +31,7 @@
     "npm-run-all": "^4.1.5",
     "nyc": "^15.1.0",
     "prettier": "^2.2.1",
+    "sinon": "^18.0.0",
```
hoodmane, line 34: What's sinon?

hoodmane (approved):
Thanks @ryanking13!

src/js/package.json
ryanking13: > What's sinon?

@hoodmane It's a test helper library for mocking / spying function call. 

Personally, I would like to switch our JS testing library from mocha to Jest or Vitest. Mocha requires `chai` for assertion and `sinon` for mocking, while Jest and Vitest have those features themselves, so we don't need to care about them.

ryanking13:
Thanks for the review!

src/js/package.json
hoodmane: If you want to switch to any JS testing library you prefer, feel free.

PR #4924, opened 2024-07-10
BLD Install libraries to local directory first

ryanking13:
### Description

This improves the installation of `static_library` and `shared_library` type recipes.

Currently, we conventionally install the final build artifacts of libraries into `WASM_LIBRARY_DIR` by manually specifying it in the `build/script`. The downside of this approach is that it's hard to distinguish which library each file belongs to as we build multiple libraries.

This PR changes it to the following:

1. It is now recommended to set the install prefix to `DISTDIR` not `WASM_LIBRARY_DIR`. All files installed in DISTDIR will be copied to WASM_LIBRARY_DIR by pyodide-build.
  a. It is worth noting that this is a backward-compatible change. Recipes that installs directly to WASM_LIBRARY_DIR will still work.
2. For shared libraries, users not must explicitly specify which files should be included in the final archive instead of including everything in the DISTDIR.

The rationale behind this change is to save the final build output of each library separately in its own directory. This makes it easy for us to investigate the output of our builds, and also allows us to distribute each build separately in the long run.

### Checklists

- [] Add a [CHANGELOG](https://github.com/pyodide/pyodide/blob/main/docs/project/changelog.md) entry
- [x] Add / update tests


ryanking13:
Seems like this isn't that good idea, regarding that often libraries have metadata about their installed directories. Closing.

Issue #4918, opened 2024-07-06
RFC Plans for unvendoring package recipes


ryanking13:
# Plans for unvendoring package recipes

Unvendoring pyodide-build is almost ready (#4882), and the next step is to unvendor package recipes.

This issue is to discuss the plan for unvendoring package recipes. This is a roadmap that I propose, which is a mixture of my thoughts and @rth's idea in https://github.com/pyodide/pyodide/pull/3827#issuecomment-1539824379. Please feel free to comment on this issue if you have any thoughts, concerns, or ideas.

## Goals

1. Unvendor package recipes from `pyodide/pyodide` repository, so that building pyodide runtime and building package recipes are separated.
2. After the unvendoring, pyodide repository should contain only the core set of packages that are necessary to test the pyodide runtime, which includes:
    - unvendored CPython modules
    - micropip and its dependencies
    - packages used to make the cross-build environment (e.g., numpy, scipy, etc.)
    - packages used for testing (e.g., pytest, fpcast-test, shardlib-test, etc.)
    
3. (optional) Remove test-only packages like fpcast-test and sharedlib-test from the distribution.

## Current status

As of 2024/07/06 there are 267 recipes in pyodide repository in total. Building all packages requires more than 2 hours in CI when not cached. Building and testing all packages exhausts the CI resource, and it becomes a bottleneck for the development of pyodide. It is also not scalable to build all packages in the pyodide repository, as the number of packages increases.

Building all packages in the pyodide repository was a good idea when Pyodide was in the early stage of development, where everything was changing rapidly. However, now that Pyodide is changing less frequently, and CPython is more stable, it is time to separate the package recipes from the pyodide repository.

## Short-term plan: `pyodide-recipes` repository

For a short-term plan, I propose to make a new repository called `pyodide-recipes` that contains all package recipes. We will put recipes and their tests in the `pyodide-recipes` repository (= the content of the `packages` directory in the pyodide/pyodide repository).

In `pyodide-recipes` repository, packages will be built and tested using:

1. Latest release of `pyodide-build`.
2. Tip-of-tree cross-build environment
3. Tip-of-tree Pyodide runtime (pyodide-core)

### Managing the recipes healthy

To keep the recipes healthy, `pyodide-recipes` repository will work as follows.

- When there is a PR that updates the recipe, packages that are affected by the PR will be built and tested.
- After the PR is merged, all packages will be built and tested in the main branch.
- All packages will be built and tested occasionally by a cron job.

[...]

PR #4925, opened 2024-07-10
NumPy v2

agriyakhetarpal:
### Description

This PR bumps the NumPy version in-tree to [version 2.0.0](https://github.com/numpy/numpy/issues/24300), [released on June 16, 2024](https://github.com/numpy/numpy/releases/tag/v2.0.0). A lot of packages here are expected to fail with their builds because there has been an ABI change in NumPy, and many packages are being tracked in https://github.com/numpy/numpy/issues/26191 – let us hope there won't be too many of them here. I would suggest that this goes into v0.27 rather than 0.26.2, not that there is a lot of time in 0.26.2 anyway...  I shall try my best to update as many failing packages as I can in this PR, or open separate PRs for each of them if that would be better. I should be able to get an estimate once the CI runs.

### Checklists

- [ ] Add / update tests

### DIFF
--- a/packages/numpy/meta.yaml
+++ b/packages/numpy/meta.yaml
@@ -1,13 +1,13 @@
 package:
   name: numpy
-  version: 1.26.4
+  version: 2.0.0
   tag:
     - min-scipy-stack
   top-level:
     - numpy
 source:
-  url: https://files.pythonhosted.org/packages/65/6e/09db70a523a96d25e115e71cc56a6f9031e7b8cd166c1ac8438307c14058/numpy-1.26.4.tar.gz
-  sha256: 2a02aba9ed12e4ac4eb3ea9421c420301a0c6460d9830d74a9df87efa4912010
+  url: https://files.pythonhosted.org/packages/05/35/fb1ada118002df3fe91b5c3b28bc0d90f879b881a5d8f68b1f9b79c44bfe/numpy-2.0.0.tar.gz
+  sha256: cf5d1c9e6837f8af9f92b6bd3e86d513cdc11f60fd62185cc49ec7d1aba34864
 
 build:
   # numpy uses vendored meson, so we need to pass the cross file manually


agriyakhetarpal:
First failure is with building NumPy itself 
```
FileNotFoundError: [Errno 2] No such file or directory:                         
'/tmp/tmpn7pa0ha3/numpy-2.0.0/numpy/core/include/numpy/numpyconfig.h'
```

which is just because it's included in `_core/` and not in `core/`, and similarly `_numpyconfig.h` will be created there

agriyakhetarpal:
The new error shows that the static libs are placed in newer paths as well.

From https://github.com/numpy/numpy/blob/2176571e1c7e5040df7a3cdfd9f7d859f4feb014/numpy/_core/npymath.ini.in#L18 I see that it moved to `_core/` as well, but `libnpyrandom` is at the same location: https://github.com/numpy/numpy/blob/2176571e1c7e5040df7a3cdfd9f7d859f4feb014/numpy/random/meson.build#L16

agriyakhetarpal:
I got NumPy to build with not a lot of effort, which is great. Next, I'm going to the NumPy dependents. SciPy v1.13.0 is the first version to support NumPy v2 (#4719), which I need to return to after #4920 gets merged. I'll explore the rest of the dependents right now.

ryanking13:
> I got NumPy to build with not a lot of effort, which is great. Next, I'm going to the NumPy dependents. SciPy v1.13.0 is the first version to support NumPy v2 (https://github.com/pyodide/pyodide/pull/4719), which I need to return to after https://github.com/pyodide/pyodide/pull/4920 gets merged. I'll explore the rest of the dependents right now.

Cooool! That's awesome. I agree that this should go to 0.27. While we try our best to update packages as much as possible to support Numpy 2.0, I think it is also an option to disable some of them temporarily if they do not support Numpy 2.0.

agriyakhetarpal:
> Cooool! That's awesome. I agree that this should go to 0.27. While we try our best to update packages as much as possible to support Numpy 2.0, I think it is also an option to disable some of them temporarily if they do not support Numpy 2.0.

@ryanking13, thanks for your input, I can look into disabling packages once I have further progress. Would you know of a way way to reliably run the Selenium code in `test_numpy.py` inside the provided Docker container? I seem to run into errors with `pyodide_dist_dir` everytime even when I spin up a fresh container. Running locally could help me debug faster.

ryanking13:
Could you share with me the error you get? Some of the package versions in our docker image (e.g. pytest) is outdated so it may cause some error, but I am not sure.

agriyakhetarpal:
Ah, I see it's a very standard error:

```bash
$ python packages/numpy/test_numpy.py
```

which gets me

```
  File "/src/packages/numpy/test_numpy.py", line 363, in <module>
    @run_in_pyodide(packages=["numpy"])
     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/src/.docker_home/.local/lib/python3.12/site-packages/pytest_pyodide/decorator.py", line 408, in __init__
    pytest_assert_rewrites and package_is_built("pytest")
                               ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/src/.docker_home/.local/lib/python3.12/site-packages/pytest_pyodide/decorator.py", line 22, in package_is_built
    return _package_is_built(package_name, pytest.pyodide_dist_dir)
                                           ^^^^^^^^^^^^^^^^^^^^^^^
AttributeError: module 'pytest' has no attribute 'pyodide_dist_dir'
```

I upgraded `pytest` and `pytest-pyodide` but didn't get anywhere with that, I'll spend some time debugging the JS interface error I just received

agriyakhetarpal:
NumPy test suite passing (not sure if this is the best way to handle the buffer), but I'll move on to the rest now.

agriyakhetarpal:
[`scikit-learn` v1.4.2](https://github.com/scikit-learn/scikit-learn/releases/tag/1.4.2) was the first one to include support for NumPy v2.0.0, but we already have v1.5.0 in the `main` branch. Should I bump it to the latest release here, i.e., v1.5.1?

agriyakhetarpal:
Here, I have updated OpenCV (`opencv-python`) and `yt`. As mentioned above, `scikit-learn` is already at a compatible version, so it seems that bumping SciPy to v1.13.0 (#4719) is the only issue holding up this PR. I assume that the versions of the rest of the packages we have right now will have been built against NumPy v2, but if they haven't been, I'll take care of that.

ryanking13:
> python packages/numpy/test_numpy.py

I think you need to call pytest not python.

```
pytest packages/numpy/test_numpy.py
```

FYI: I generally prefer verbose mode + chrome browser when testing locally.

```
pytest -v --rt=chrome packages/numpy/test_numpy.py
```

ryanking13:
> [scikit-learn v1.4.2](https://github.com/scikit-learn/scikit-learn/releases/tag/1.4.2) was the first one to include support for NumPy v2.0.0, but we already have v1.5.0 in the main branch. Should I bump it to the latest release here, i.e., v1.5.1?

Updating the package version is not required, but updates are always welcome. If you're having trouble updating, you can leave it for later.

> it seems that bumping SciPy to v1.13.0 (https://github.com/pyodide/pyodide/pull/4719) is the only issue holding up this PR

Awesome! Thank you for updating these packages!

agriyakhetarpal:
Updated `scikit-learn` to version 1.5.1 in 4ebc115ffd813b2da8b2e6fee8bf6e3782d06cd4 – tagging the recipe and package maintainer @lesteve for visibility

agriyakhetarpal:
Hi @johnwason, I found some of your previous PRs working on the recipe for RobotRaconteur. I've had some trouble with it, so I upgraded it to `v1.2.2` here (which I see was released eleven hours ago, at the time of writing). Given your alignment and previous work here, I was wondering whether you would like to enlist yourself as a maintainer for its recipe and help with upgrades and general upkeep? This way, other developers and users of Pyodide will know who to tag in case they seem to have problems. Thanks!

johnwason:
@agriyakhetarpal yes add me as the maintainer. The only change required for NumPy v2 in the upstream was to change the include search path. What errors are you seeing?

agriyakhetarpal:
Thanks for your confirmation, @johnwason – I think I managed to fix them all with the last few commits :)

agriyakhetarpal:
@hoodmane, the [CircleCI logs](https://app.circleci.com/pipelines/github/pyodide/pyodide/7685/workflows/e87fba4a-fe59-4fa3-a8f2-919ff3f6ce21/jobs/110292?invite=true#step-104-479276_80) display that SciPy v1.13.0 fails to compile because of a missing semicolon: would you suggest I patch `f2c` to fix this, or is there something bigger at play here? The strange thing is that I don't see why that should be a syntax error; even my text editor stops its syntax highlighting when I add a semicolon where it is asking me to.

rgommers:
@agriyakhetarpal re the `complex` collision: that's a common problem, since it seems everyone and their dog have the bad habit to do `#define complex` or something similar. In SciPy's `main` branch I see we have a patch for this:

```diff
--- a/scipy/sparse/linalg/_dsolve/SuperLU/SRC/slu_scomplex.h
+++ b/scipy/sparse/linalg/_dsolve/SuperLU/SRC/slu_scomplex.h
@@ -30,7 +30,6 @@ at the top-level directory.

 typedef struct { float r, i; } singlecomplex;

-#define complex singlecomplex  // backward compatibility
```

I don't know how the `f2c` header gets pulled in here (perhaps because BLAS is linked?), but you can probably patch SuperLU like the above to avoid the problem.

agriyakhetarpal:
I checked, and the patch doesn't apply, @rgommers :( since that macro did not exist in SciPy v1.13.0/v1.13.1/v1.14.0 for that file in SuperLU, neither in the source tree nor the PyPI sdist – below the `typedef struct { float r, i; } singlecomplex;` line. I think we'll need to try out something else. I could upgrade to v1.13.1 here, and see if that resolves the problem.

agriyakhetarpal:
While I have just pushed a few commits to build/upgrade to SciPy v1.13.1, it doesn't really build either, failing with the same error. I could put the SciPy upgrade in its own PR, though.

agriyakhetarpal:
Actually, let me indeed cherry-pick the SciPy v1.13.1 update to a new branch and PR, since I can then work on v1.14.0 further separately: https://github.com/pyodide/pyodide/pull/5011

agriyakhetarpal:
I'm pretty convinced that the issue with SciPy is due to the `f2c` header file. NumPy v2.0.0 changed all of the occurrences of the `complex` macro to `singlecomplex` in numpy/numpy#24364 to fix a SuperLU issue, and this change was not present in NumPy v1.26.4/v1.26.5. Some more packages amongst the NumPy dependents will fail to build, as with the commit I pushed above. Still, since we don't have a `--keep-going` option for `pyodide build-recipes`, some more failures keep turning up occasionally that I need to fix at times (which is expected and reasonable, of course) since a failure stops other builds as well. I will temporarily disable SciPy, fix the remaining packages in the `numpy-dependents` group, and re-enable SciPy again when I will have looked into this in more detail.

P.S. I opened a new issue about adding a `--keep-going` option (linked above) so that we can continue building recipes until they are completed.

agriyakhetarpal:
> I will temporarily disable SciPy, fix the remaining packages in the `numpy-dependents` group, and re-enable SciPy again when I will have looked into this in more detail.

I built all packages (minus the shared/static libraries, which I found to be particularly time-consuming), and there are no other build failures. I have two further questions:

1. Should we run `make -C packages update-all` with this PR to be on the safer side, or would that be too much work and better pursued in a separate PR? My opinion is that it should be a separate PR.
2. Is there a way to use a custom f2c header file to work around the problem? Patching NumPy to fix a SciPy build issue doesn't seem like the best way to do it (and I don't know whether it's possible to patch NumPy and replace the header there through the patches in SciPy's recipe, though we could just copy the header file and pass it to the compiler/build system).

ryanking13:
> Should we run make -C packages update-all with this PR to be on the safer side, or would that be too much work and better pursued in a separate PR? My opinion is that it should be a separate PR.

Yes I think we can do that in a separate PR.

> Is there a way to use a custom f2c header file to work around the problem? If f2c is an issue

I am not sure if I understand the f2c issue correctly. But if you want to patch f2c, as we are already using Hood's f2c fork to build scipy. You can try patching it directly.

https://github.com/pyodide/pyodide/blob/b4d6989fbd97f360c76a593b971ed7bbdae2efb0/packages/scipy/meta.yaml#L62-L64

juntyr:
What's the status of this PR?

ryanking13:
Agriya is leading this PR but he was a bit busy these days. I'll start looking at the build issue too maybe from next week.

ryanking13:
I opened a PR in your branch to fix the type definition issue (https://github.com/agriyakhetarpal/pyodide/pull/1). I haven't run the full scipy test suite, but I was able to build scipy locally.

agriyakhetarpal:
Bumping to NumPy v2.1.3 from the previous 2.0.0 meant that a bunch of more errors across packages crept in which weren't present earlier. I'll bump their versions. Is there a way in our CI to force running all tests for all packages instead of not running tests whose results are cached?

agriyakhetarpal:
I guess we don't need to upgrade PyArrow because the version that supports NumPy v2 is 16.0.0, and we have a later version: 17.0.0. I guess we'll have to investigate the `numpy.core.multiarray failed to import` error.

agriyakhetarpal:
I guess we don't need to upgrade PyArrow because the version that supports NumPy v2 is 16.0.0, and we have a later version: 17.0.0. We'll have to investigate the `numpy.core.multiarray failed to import` error.

agriyakhetarpal:
https://github.com/h5py/h5py/blob/3.12.1/setup_build.py#L145-L150 clears my confusion on the `libhdf5` version required for `h5py` version 3.12.1 – any version above 1.10.6 should work, and we have 1.12.1. I've updated `h5py` to version 3.12.1 in 4e61dcd20676e6b839c38d55b00100c30c03a542.

agriyakhetarpal:
PyArrow's case is an interesting one. Since PyArrow builds against a specific "released" Pyodide version in a separate repository, it can't build against NumPy v2 because this PR isn't merged and hasn't been a part of a release yet. We'll need to update the build workflow in https://github.com/joemarshall/pyarrow-pyodide for it to be able to PR branches such as this one and build Pyodide from them, too, and temporarily update the link to the wheel in the recipe here to that one in order to test it with NumPy v2. We'll need to think about this scenario in #4918. I'll look into it later in the day.

agriyakhetarpal:
I've bumped _most_ (if not all) packages that have NumPy as either a host or a runtime dependency, and in some cases, bumped to a slightly higher version. Matplotlib, PyArrow, [River](https://github.com/online-ml/river), and the duo of [`rebound`](https://github.com/dtamayo/reboundx) & [`reboundx`](https://github.com/dtamayo/reboundx) remain.

I don't think there should be a problem with `rebound`, but when I bumped it, I faced this Emscripten unresolved symbol which is in an area where I have no pointers on how to fix such a thing: `___em_js__reb_remove_last_line`

The good thing is that we do have recipe maintainers for these slightly tricky packages: @hannorein and @MaxHalford, you're listed for `rebound`/`reboundx` and `river` respectively, may I ask you both if you could be of help with us somehow? :)

As far as I understand, `rebound`/`reboundx` do not pin NumPy, but I'm unsure of tackling the packaging infrastructure. As for `river`, there is an in-progress update that seems to have been stalled: https://github.com/conda-forge/river-feedstock/pull/28 (sourced from https://github.com/online-ml/river/issues/1624) and it still supports Cython<3, which, IIRC, can't build wheels with NumPy>=2.0.0rc1 as a build-time dependency.

As discussed above, we can also disable these three packages temporarily to let this PR proceed in the time we work on getting Matplotlib and PyArrow to build.

hannorein:
Thanks for the heads up. The `reb_remove_last_line` function is defined in the same file using the `EM_JS` macro and it works when I compile it with emcc locally. So I'm a bit confused about why this error shows up. Is there a way to see the build log?

agriyakhetarpal:
Thanks for chiming in, @hannorein! We can't see the logs in CI very easily, but I'm happy to provide a trace from a local run in a Dockerfile for you:

<details>
<summary>Tap to expand</summary>

<br>

I ran the command:

```
PYODIDE_DEBUG=1 PYODIDE_PACKAGES="rebound" make
```

and here's what I got:

```
Error building rebound. Printing build logs.                                                                                                                         
Hmm...  Looks like a unified diff to me...                                                                                                                           
--------------------------                                                                                                                                           
|From c8c7f2588c555acc9cb1ecf9dbd5b56f1ccedeeb Mon Sep 17 00:00:00 2001                                                                                              
|From: Agriya Khetarpal <74401230+agriyakhetarpal@users.noreply.github.com>                                                                                          
|Date: Tue, 5 Nov 2024 14:03:28 +0530                                                                                                                                
|Subject: [PATCH] Fix install_name and drop other args                                                                                                               
|                                                                                                                                                                    
|---                                                                                                                                                                 
| setup.py | 2 --                                                                                                                                                    
| 1 file changed, 2 deletions(-)                                                                                                                                     
|                                                                                                                                                                    
|diff --git a/setup.py b/setup.py                                                                                                                                    
|index 6443cb27..3d55bc02 100644                                                                                                                                     
|--- a/setup.py                                                                                                                                                      
|+++ b/setup.py                                                                                                                                                      
--------------------------                                                                                                                                           
patching file setup.py                                                                                                                                               
Using Plan A...                                                                                                                                                      
Hunk #1 succeeded at 23.                                                                                                                                             
Hmm...  Ignoring the trailing garbage.                                                                                                                               
done                                                                                                                                                                 
Running source pyodide_env in /src/packages/rebound                                                                                                                  
/tmp/build-env-12a_kyk7/lib/python3.12/site-packages/setuptools/_distutils/dist.py:261: UserWarning: Unknown distribution option: 'tests_require'                    
  warnings.warn(msg)                                                                                                                                                 
/tmp/build-env-12a_kyk7/lib/python3.12/site-packages/setuptools/_distutils/dist.py:261: UserWarning: Unknown distribution option: 'test_suite'                       
  warnings.warn(msg)                                                                                                                                                 
running egg_info                                                                                                                                                     
writing rebound.egg-info/PKG-INFO                                                                                                                                    
writing dependency_links to rebound.egg-info/dependency_links.txt                                                                                                    
writing top-level names to rebound.egg-info/top_level.txt                                                                                                            
reading manifest file 'rebound.egg-info/SOURCES.txt'                                                                                                                 
reading manifest template 'MANIFEST.in'                                                                                                                              
adding license file 'LICENSE'                                                                                                                                        
writing manifest file 'rebound.egg-info/SOURCES.txt'                                                                                                                 
/tmp/build-env-12a_kyk7/lib/python3.12/site-packages/setuptools/_distutils/dist.py:261: UserWarning: Unknown distribution option: 'tests_require'                    
  warnings.warn(msg)                                                                                                                                                 
/tmp/build-env-12a_kyk7/lib/python3.12/site-packages/setuptools/_distutils/dist.py:261: UserWarning: Unknown distribution option: 'test_suite'                       
  warnings.warn(msg)                                                                                                                                                 
running bdist_wheel                                                                                                                                                  
running build                                                                                                                                                        
running build_py                                                                                                                                                     
creating build/lib.emscripten_3_1_58_wasm32-cpython-312/rebound                                                                                                      
copying rebound/simulation.py -> build/lib.emscripten_3_1_58_wasm32-cpython-312/rebound                                                                              
copying rebound/citations.py -> build/lib.emscripten_3_1_58_wasm32-cpython-312/rebound                                                                               
copying rebound/vectors.py -> build/lib.emscripten_3_1_58_wasm32-cpython-312/rebound                                                                                 
copying rebound/horizons.py -> build/lib.emscripten_3_1_58_wasm32-cpython-312/rebound                                                                                
copying rebound/widget.py -> build/lib.emscripten_3_1_58_wasm32-cpython-312/rebound                                                                                  
copying rebound/units.py -> build/lib.emscripten_3_1_58_wasm32-cpython-312/rebound                                                                                   
copying rebound/orbit.py -> build/lib.emscripten_3_1_58_wasm32-cpython-312/rebound                                                                                   
copying rebound/plotting.py -> build/lib.emscripten_3_1_58_wasm32-cpython-312/rebound                                                                                
copying rebound/tools.py -> build/lib.emscripten_3_1_58_wasm32-cpython-312/rebound                                                                                   
copying rebound/__init__.py -> build/lib.emscripten_3_1_58_wasm32-cpython-312/rebound                                                                                
copying rebound/simulationarchive.py -> build/lib.emscripten_3_1_58_wasm32-cpython-312/rebound                                                                       
copying rebound/particles.py -> build/lib.emscripten_3_1_58_wasm32-cpython-312/rebound                                                                               
copying rebound/hash.py -> build/lib.emscripten_3_1_58_wasm32-cpython-312/rebound                                                                                    
copying rebound/binary_field_descriptor.py -> build/lib.emscripten_3_1_58_wasm32-cpython-312/rebound                                                                 
copying rebound/particle.py -> build/lib.emscripten_3_1_58_wasm32-cpython-312/rebound                                                                                
copying rebound/variation.py -> build/lib.emscripten_3_1_58_wasm32-cpython-312/rebound                                                                               
copying rebound/rotation.py -> build/lib.emscripten_3_1_58_wasm32-cpython-312/rebound                                                                                
copying rebound/data.py -> build/lib.emscripten_3_1_58_wasm32-cpython-312/rebound                                                                                    
creating build/lib.emscripten_3_1_58_wasm32-cpython-312/rebound/integrators                                                                                          
copying rebound/integrators/saba.py -> build/lib.emscripten_3_1_58_wasm32-cpython-312/rebound/integrators                                                            
copying rebound/integrators/eos.py -> build/lib.emscripten_3_1_58_wasm32-cpython-312/rebound/integrators                                                             
copying rebound/integrators/trace.py -> build/lib.emscripten_3_1_58_wasm32-cpython-312/rebound/integrators                                                           
copying rebound/integrators/whfast512.py -> build/lib.emscripten_3_1_58_wasm32-cpython-312/rebound/integrators                                                       
copying rebound/integrators/janus.py -> build/lib.emscripten_3_1_58_wasm32-cpython-312/rebound/integrators                                                           
copying rebound/integrators/__init__.py -> build/lib.emscripten_3_1_58_wasm32-cpython-312/rebound/integrators                                                        
copying rebound/integrators/ias15.py -> build/lib.emscripten_3_1_58_wasm32-cpython-312/rebound/integrators                                                           
copying rebound/integrators/mercurius.py -> build/lib.emscripten_3_1_58_wasm32-cpython-312/rebound/integrators                                                       
copying rebound/integrators/bs.py -> build/lib.emscripten_3_1_58_wasm32-cpython-312/rebound/integrators                                                              
copying rebound/integrators/whfast.py -> build/lib.emscripten_3_1_58_wasm32-cpython-312/rebound/integrators                                                          
copying rebound/integrators/sei.py -> build/lib.emscripten_3_1_58_wasm32-cpython-312/rebound/integrators                                                             
copying rebound/rebound.h -> build/lib.emscripten_3_1_58_wasm32-cpython-312/rebound                                                                                  
running build_ext                                                                                                                                                    
building 'librebound' extension                                                                                                                                      
creating build/temp.emscripten_3_1_58_wasm32-cpython-312/src                                                                                                         
/tmp/tmpbfksie2y/cc -fno-strict-overflow -DNDEBUG -g -O3 -Wall -O2 -g0 -fPIC -DPY_CALL_TRAMPOLINE -DLIBREBOUND -Isrc -I/tmp/build-env-12a_kyk7/include               
-I/src/cpython/installs/python-3.12.1/include/python3.12 -c src/binarydiff.c -o build/temp.emscripten_3_1_58_wasm32-cpython-312/src/binarydiff.o -fstrict-aliasing   
-O3 -std=c99 -Wno-unknown-pragmas -DGITHASH=45546a163633ed95868d45013a24216e7339cdfd -DLIBREBOUND -D_GNU_SOURCE -DSERVER -fPIC                                       
/tmp/tmpbfksie2y/cc -fno-strict-overflow -DNDEBUG -g -O3 -Wall -O2 -g0 -fPIC -DPY_CALL_TRAMPOLINE -DLIBREBOUND -Isrc -I/tmp/build-env-12a_kyk7/include               
-I/src/cpython/installs/python-3.12.1/include/python3.12 -c src/boundary.c -o build/temp.emscripten_3_1_58_wasm32-cpython-312/src/boundary.o -fstrict-aliasing -O3   
-std=c99 -Wno-unknown-pragmas -DGITHASH=45546a163633ed95868d45013a24216e7339cdfd -DLIBREBOUND -D_GNU_SOURCE -DSERVER -fPIC                                           
/tmp/tmpbfksie2y/cc -fno-strict-overflow -DNDEBUG -g -O3 -Wall -O2 -g0 -fPIC -DPY_CALL_TRAMPOLINE -DLIBREBOUND -Isrc -I/tmp/build-env-12a_kyk7/include               
-I/src/cpython/installs/python-3.12.1/include/python3.12 -c src/collision.c -o build/temp.emscripten_3_1_58_wasm32-cpython-312/src/collision.o -fstrict-aliasing -O3 
-std=c99 -Wno-unknown-pragmas -DGITHASH=45546a163633ed95868d45013a24216e7339cdfd -DLIBREBOUND -D_GNU_SOURCE -DSERVER -fPIC                                           
/tmp/tmpbfksie2y/cc -fno-strict-overflow -DNDEBUG -g -O3 -Wall -O2 -g0 -fPIC -DPY_CALL_TRAMPOLINE -DLIBREBOUND -Isrc -I/tmp/build-env-12a_kyk7/include               
-I/src/cpython/installs/python-3.12.1/include/python3.12 -c src/derivatives.c -o build/temp.emscripten_3_1_58_wasm32-cpython-312/src/derivatives.o -fstrict-aliasing 
-O3 -std=c99 -Wno-unknown-pragmas -DGITHASH=45546a163633ed95868d45013a24216e7339cdfd -DLIBREBOUND -D_GNU_SOURCE -DSERVER -fPIC                                       
/tmp/tmpbfksie2y/cc -fno-strict-overflow -DNDEBUG -g -O3 -Wall -O2 -g0 -fPIC -DPY_CALL_TRAMPOLINE -DLIBREBOUND -Isrc -I/tmp/build-env-12a_kyk7/include               
-I/src/cpython/installs/python-3.12.1/include/python3.12 -c src/display.c -o build/temp.emscripten_3_1_58_wasm32-cpython-312/src/display.o -fstrict-aliasing -O3     
-std=c99 -Wno-unknown-pragmas -DGITHASH=45546a163633ed95868d45013a24216e7339cdfd -DLIBREBOUND -D_GNU_SOURCE -DSERVER -fPIC                                           
/tmp/tmpbfksie2y/cc -fno-strict-overflow -DNDEBUG -g -O3 -Wall -O2 -g0 -fPIC -DPY_CALL_TRAMPOLINE -DLIBREBOUND -Isrc -I/tmp/build-env-12a_kyk7/include               
-I/src/cpython/installs/python-3.12.1/include/python3.12 -c src/fmemopen.c -o build/temp.emscripten_3_1_58_wasm32-cpython-312/src/fmemopen.o -fstrict-aliasing -O3   
-std=c99 -Wno-unknown-pragmas -DGITHASH=45546a163633ed95868d45013a24216e7339cdfd -DLIBREBOUND -D_GNU_SOURCE -DSERVER -fPIC                                           
/tmp/tmpbfksie2y/cc -fno-strict-overflow -DNDEBUG -g -O3 -Wall -O2 -g0 -fPIC -DPY_CALL_TRAMPOLINE -DLIBREBOUND -Isrc -I/tmp/build-env-12a_kyk7/include               
-I/src/cpython/installs/python-3.12.1/include/python3.12 -c src/gravity.c -o build/temp.emscripten_3_1_58_wasm32-cpython-312/src/gravity.o -fstrict-aliasing -O3     
-std=c99 -Wno-unknown-pragmas -DGITHASH=45546a163633ed95868d45013a24216e7339cdfd -DLIBREBOUND -D_GNU_SOURCE -DSERVER -fPIC                                           
/tmp/tmpbfksie2y/cc -fno-strict-overflow -DNDEBUG -g -O3 -Wall -O2 -g0 -fPIC -DPY_CALL_TRAMPOLINE -DLIBREBOUND -Isrc -I/tmp/build-env-12a_kyk7/include               
-I/src/cpython/installs/python-3.12.1/include/python3.12 -c src/input.c -o build/temp.emscripten_3_1_58_wasm32-cpython-312/src/input.o -fstrict-aliasing -O3 -std=c99
-Wno-unknown-pragmas -DGITHASH=45546a163633ed95868d45013a24216e7339cdfd -DLIBREBOUND -D_GNU_SOURCE -DSERVER -fPIC                                                    
/tmp/tmpbfksie2y/cc -fno-strict-overflow -DNDEBUG -g -O3 -Wall -O2 -g0 -fPIC -DPY_CALL_TRAMPOLINE -DLIBREBOUND -Isrc -I/tmp/build-env-12a_kyk7/include               
-I/src/cpython/installs/python-3.12.1/include/python3.12 -c src/integrator.c -o build/temp.emscripten_3_1_58_wasm32-cpython-312/src/integrator.o -fstrict-aliasing   
-O3 -std=c99 -Wno-unknown-pragmas -DGITHASH=45546a163633ed95868d45013a24216e7339cdfd -DLIBREBOUND -D_GNU_SOURCE -DSERVER -fPIC                                       
/tmp/tmpbfksie2y/cc -fno-strict-overflow -DNDEBUG -g -O3 -Wall -O2 -g0 -fPIC -DPY_CALL_TRAMPOLINE -DLIBREBOUND -Isrc -I/tmp/build-env-12a_kyk7/include               
-I/src/cpython/installs/python-3.12.1/include/python3.12 -c src/integrator_bs.c -o build/temp.emscripten_3_1_58_wasm32-cpython-312/src/integrator_bs.o               
-fstrict-aliasing -O3 -std=c99 -Wno-unknown-pragmas -DGITHASH=45546a163633ed95868d45013a24216e7339cdfd -DLIBREBOUND -D_GNU_SOURCE -DSERVER -fPIC                     
/tmp/tmpbfksie2y/cc -fno-strict-overflow -DNDEBUG -g -O3 -Wall -O2 -g0 -fPIC -DPY_CALL_TRAMPOLINE -DLIBREBOUND -Isrc -I/tmp/build-env-12a_kyk7/include               
-I/src/cpython/installs/python-3.12.1/include/python3.12 -c src/integrator_eos.c -o build/temp.emscripten_3_1_58_wasm32-cpython-312/src/integrator_eos.o             
-fstrict-aliasing -O3 -std=c99 -Wno-unknown-pragmas -DGITHASH=45546a163633ed95868d45013a24216e7339cdfd -DLIBREBOUND -D_GNU_SOURCE -DSERVER -fPIC                     
/tmp/tmpbfksie2y/cc -fno-strict-overflow -DNDEBUG -g -O3 -Wall -O2 -g0 -fPIC -DPY_CALL_TRAMPOLINE -DLIBREBOUND -Isrc -I/tmp/build-env-12a_kyk7/include               
-I/src/cpython/installs/python-3.12.1/include/python3.12 -c src/integrator_ias15.c -o build/temp.emscripten_3_1_58_wasm32-cpython-312/src/integrator_ias15.o         
-fstrict-aliasing -O3 -std=c99 -Wno-unknown-pragmas -DGITHASH=45546a163633ed95868d45013a24216e7339cdfd -DLIBREBOUND -D_GNU_SOURCE -DSERVER -fPIC                     
/tmp/tmpbfksie2y/cc -fno-strict-overflow -DNDEBUG -g -O3 -Wall -O2 -g0 -fPIC -DPY_CALL_TRAMPOLINE -DLIBREBOUND -Isrc -I/tmp/build-env-12a_kyk7/include               
-I/src/cpython/installs/python-3.12.1/include/python3.12 -c src/integrator_janus.c -o build/temp.emscripten_3_1_58_wasm32-cpython-312/src/integrator_janus.o         
-fstrict-aliasing -O3 -std=c99 -Wno-unknown-pragmas -DGITHASH=45546a163633ed95868d45013a24216e7339cdfd -DLIBREBOUND -D_GNU_SOURCE -DSERVER -fPIC                     
/tmp/tmpbfksie2y/cc -fno-strict-overflow -DNDEBUG -g -O3 -Wall -O2 -g0 -fPIC -DPY_CALL_TRAMPOLINE -DLIBREBOUND -Isrc -I/tmp/build-env-12a_kyk7/include               
-I/src/cpython/installs/python-3.12.1/include/python3.12 -c src/integrator_leapfrog.c -o build/temp.emscripten_3_1_58_wasm32-cpython-312/src/integrator_leapfrog.o   
-fstrict-aliasing -O3 -std=c99 -Wno-unknown-pragmas -DGITHASH=45546a163633ed95868d45013a24216e7339cdfd -DLIBREBOUND -D_GNU_SOURCE -DSERVER -fPIC                     
/tmp/tmpbfksie2y/cc -fno-strict-overflow -DNDEBUG -g -O3 -Wall -O2 -g0 -fPIC -DPY_CALL_TRAMPOLINE -DLIBREBOUND -Isrc -I/tmp/build-env-12a_kyk7/include               
-I/src/cpython/installs/python-3.12.1/include/python3.12 -c src/integrator_mercurius.c -o build/temp.emscripten_3_1_58_wasm32-cpython-312/src/integrator_mercurius.o 
-fstrict-aliasing -O3 -std=c99 -Wno-unknown-pragmas -DGITHASH=45546a163633ed95868d45013a24216e7339cdfd -DLIBREBOUND -D_GNU_SOURCE -DSERVER -fPIC                     
/tmp/tmpbfksie2y/cc -fno-strict-overflow -DNDEBUG -g -O3 -Wall -O2 -g0 -fPIC -DPY_CALL_TRAMPOLINE -DLIBREBOUND -Isrc -I/tmp/build-env-12a_kyk7/include               
-I/src/cpython/installs/python-3.12.1/include/python3.12 -c src/integrator_saba.c -o build/temp.emscripten_3_1_58_wasm32-cpython-312/src/integrator_saba.o           
-fstrict-aliasing -O3 -std=c99 -Wno-unknown-pragmas -DGITHASH=45546a163633ed95868d45013a24216e7339cdfd -DLIBREBOUND -D_GNU_SOURCE -DSERVER -fPIC                     
/tmp/tmpbfksie2y/cc -fno-strict-overflow -DNDEBUG -g -O3 -Wall -O2 -g0 -fPIC -DPY_CALL_TRAMPOLINE -DLIBREBOUND -Isrc -I/tmp/build-env-12a_kyk7/include               
-I/src/cpython/installs/python-3.12.1/include/python3.12 -c src/integrator_sei.c -o build/temp.emscripten_3_1_58_wasm32-cpython-312/src/integrator_sei.o             
-fstrict-aliasing -O3 -std=c99 -Wno-unknown-pragmas -DGITHASH=45546a163633ed95868d45013a24216e7339cdfd -DLIBREBOUND -D_GNU_SOURCE -DSERVER -fPIC                     
/tmp/tmpbfksie2y/cc -fno-strict-overflow -DNDEBUG -g -O3 -Wall -O2 -g0 -fPIC -DPY_CALL_TRAMPOLINE -DLIBREBOUND -Isrc -I/tmp/build-env-12a_kyk7/include               
-I/src/cpython/installs/python-3.12.1/include/python3.12 -c src/integrator_trace.c -o build/temp.emscripten_3_1_58_wasm32-cpython-312/src/integrator_trace.o         
-fstrict-aliasing -O3 -std=c99 -Wno-unknown-pragmas -DGITHASH=45546a163633ed95868d45013a24216e7339cdfd -DLIBREBOUND -D_GNU_SOURCE -DSERVER -fPIC                     
/tmp/tmpbfksie2y/cc -fno-strict-overflow -DNDEBUG -g -O3 -Wall -O2 -g0 -fPIC -DPY_CALL_TRAMPOLINE -DLIBREBOUND -Isrc -I/tmp/build-env-12a_kyk7/include               
-I/src/cpython/installs/python-3.12.1/include/python3.12 -c src/integrator_whfast.c -o build/temp.emscripten_3_1_58_wasm32-cpython-312/src/integrator_whfast.o       
-fstrict-aliasing -O3 -std=c99 -Wno-unknown-pragmas -DGITHASH=45546a163633ed95868d45013a24216e7339cdfd -DLIBREBOUND -D_GNU_SOURCE -DSERVER -fPIC                     
/tmp/tmpbfksie2y/cc -fno-strict-overflow -DNDEBUG -g -O3 -Wall -O2 -g0 -fPIC -DPY_CALL_TRAMPOLINE -DLIBREBOUND -Isrc -I/tmp/build-env-12a_kyk7/include               
-I/src/cpython/installs/python-3.12.1/include/python3.12 -c src/integrator_whfast512.c -o build/temp.emscripten_3_1_58_wasm32-cpython-312/src/integrator_whfast512.o 
-fstrict-aliasing -O3 -std=c99 -Wno-unknown-pragmas -DGITHASH=45546a163633ed95868d45013a24216e7339cdfd -DLIBREBOUND -D_GNU_SOURCE -DSERVER -fPIC                     
/tmp/tmpbfksie2y/cc -fno-strict-overflow -DNDEBUG -g -O3 -Wall -O2 -g0 -fPIC -DPY_CALL_TRAMPOLINE -DLIBREBOUND -Isrc -I/tmp/build-env-12a_kyk7/include               
-I/src/cpython/installs/python-3.12.1/include/python3.12 -c src/output.c -o build/temp.emscripten_3_1_58_wasm32-cpython-312/src/output.o -fstrict-aliasing -O3       
-std=c99 -Wno-unknown-pragmas -DGITHASH=45546a163633ed95868d45013a24216e7339cdfd -DLIBREBOUND -D_GNU_SOURCE -DSERVER -fPIC                                           
/tmp/tmpbfksie2y/cc -fno-strict-overflow -DNDEBUG -g -O3 -Wall -O2 -g0 -fPIC -DPY_CALL_TRAMPOLINE -DLIBREBOUND -Isrc -I/tmp/build-env-12a_kyk7/include               
-I/src/cpython/installs/python-3.12.1/include/python3.12 -c src/particle.c -o build/temp.emscripten_3_1_58_wasm32-cpython-312/src/particle.o -fstrict-aliasing -O3   
-std=c99 -Wno-unknown-pragmas -DGITHASH=45546a163633ed95868d45013a24216e7339cdfd -DLIBREBOUND -D_GNU_SOURCE -DSERVER -fPIC                                           
/tmp/tmpbfksie2y/cc -fno-strict-overflow -DNDEBUG -g -O3 -Wall -O2 -g0 -fPIC -DPY_CALL_TRAMPOLINE -DLIBREBOUND -Isrc -I/tmp/build-env-12a_kyk7/include               
-I/src/cpython/installs/python-3.12.1/include/python3.12 -c src/rebound.c -o build/temp.emscripten_3_1_58_wasm32-cpython-312/src/rebound.o -fstrict-aliasing -O3     
-std=c99 -Wno-unknown-pragmas -DGITHASH=45546a163633ed95868d45013a24216e7339cdfd -DLIBREBOUND -D_GNU_SOURCE -DSERVER -fPIC                                           
/tmp/tmpbfksie2y/cc -fno-strict-overflow -DNDEBUG -g -O3 -Wall -O2 -g0 -fPIC -DPY_CALL_TRAMPOLINE -DLIBREBOUND -Isrc -I/tmp/build-env-12a_kyk7/include               
-I/src/cpython/installs/python-3.12.1/include/python3.12 -c src/rotations.c -o build/temp.emscripten_3_1_58_wasm32-cpython-312/src/rotations.o -fstrict-aliasing -O3 
-std=c99 -Wno-unknown-pragmas -DGITHASH=45546a163633ed95868d45013a24216e7339cdfd -DLIBREBOUND -D_GNU_SOURCE -DSERVER -fPIC                                           
/tmp/tmpbfksie2y/cc -fno-strict-overflow -DNDEBUG -g -O3 -Wall -O2 -g0 -fPIC -DPY_CALL_TRAMPOLINE -DLIBREBOUND -Isrc -I/tmp/build-env-12a_kyk7/include               
-I/src/cpython/installs/python-3.12.1/include/python3.12 -c src/server.c -o build/temp.emscripten_3_1_58_wasm32-cpython-312/src/server.o -fstrict-aliasing -O3       
-std=c99 -Wno-unknown-pragmas -DGITHASH=45546a163633ed95868d45013a24216e7339cdfd -DLIBREBOUND -D_GNU_SOURCE -DSERVER -fPIC                                           
/tmp/tmpbfksie2y/cc -fno-strict-overflow -DNDEBUG -g -O3 -Wall -O2 -g0 -fPIC -DPY_CALL_TRAMPOLINE -DLIBREBOUND -Isrc -I/tmp/build-env-12a_kyk7/include               
-I/src/cpython/installs/python-3.12.1/include/python3.12 -c src/simulationarchive.c -o build/temp.emscripten_3_1_58_wasm32-cpython-312/src/simulationarchive.o       
-fstrict-aliasing -O3 -std=c99 -Wno-unknown-pragmas -DGITHASH=45546a163633ed95868d45013a24216e7339cdfd -DLIBREBOUND -D_GNU_SOURCE -DSERVER -fPIC                     
/tmp/tmpbfksie2y/cc -fno-strict-overflow -DNDEBUG -g -O3 -Wall -O2 -g0 -fPIC -DPY_CALL_TRAMPOLINE -DLIBREBOUND -Isrc -I/tmp/build-env-12a_kyk7/include               
-I/src/cpython/installs/python-3.12.1/include/python3.12 -c src/tools.c -o build/temp.emscripten_3_1_58_wasm32-cpython-312/src/tools.o -fstrict-aliasing -O3 -std=c99
-Wno-unknown-pragmas -DGITHASH=45546a163633ed95868d45013a24216e7339cdfd -DLIBREBOUND -D_GNU_SOURCE -DSERVER -fPIC                                                    
/tmp/tmpbfksie2y/cc -fno-strict-overflow -DNDEBUG -g -O3 -Wall -O2 -g0 -fPIC -DPY_CALL_TRAMPOLINE -DLIBREBOUND -Isrc -I/tmp/build-env-12a_kyk7/include               
-I/src/cpython/installs/python-3.12.1/include/python3.12 -c src/transformations.c -o build/temp.emscripten_3_1_58_wasm32-cpython-312/src/transformations.o           
-fstrict-aliasing -O3 -std=c99 -Wno-unknown-pragmas -DGITHASH=45546a163633ed95868d45013a24216e7339cdfd -DLIBREBOUND -D_GNU_SOURCE -DSERVER -fPIC                     
/tmp/tmpbfksie2y/cc -fno-strict-overflow -DNDEBUG -g -O3 -Wall -O2 -g0 -fPIC -DPY_CALL_TRAMPOLINE -DLIBREBOUND -Isrc -I/tmp/build-env-12a_kyk7/include               
-I/src/cpython/installs/python-3.12.1/include/python3.12 -c src/tree.c -o build/temp.emscripten_3_1_58_wasm32-cpython-312/src/tree.o -fstrict-aliasing -O3 -std=c99  
-Wno-unknown-pragmas -DGITHASH=45546a163633ed95868d45013a24216e7339cdfd -DLIBREBOUND -D_GNU_SOURCE -DSERVER -fPIC                                                    
/tmp/tmpbfksie2y/cc build/temp.emscripten_3_1_58_wasm32-cpython-312/src/binarydiff.o build/temp.emscripten_3_1_58_wasm32-cpython-312/src/boundary.o                  
build/temp.emscripten_3_1_58_wasm32-cpython-312/src/collision.o build/temp.emscripten_3_1_58_wasm32-cpython-312/src/derivatives.o                                    
build/temp.emscripten_3_1_58_wasm32-cpython-312/src/display.o build/temp.emscripten_3_1_58_wasm32-cpython-312/src/fmemopen.o                                         
build/temp.emscripten_3_1_58_wasm32-cpython-312/src/gravity.o build/temp.emscripten_3_1_58_wasm32-cpython-312/src/input.o                                            
build/temp.emscripten_3_1_58_wasm32-cpython-312/src/integrator.o build/temp.emscripten_3_1_58_wasm32-cpython-312/src/integrator_bs.o                                 
build/temp.emscripten_3_1_58_wasm32-cpython-312/src/integrator_eos.o build/temp.emscripten_3_1_58_wasm32-cpython-312/src/integrator_ias15.o                          
build/temp.emscripten_3_1_58_wasm32-cpython-312/src/integrator_janus.o build/temp.emscripten_3_1_58_wasm32-cpython-312/src/integrator_leapfrog.o                     
build/temp.emscripten_3_1_58_wasm32-cpython-312/src/integrator_mercurius.o build/temp.emscripten_3_1_58_wasm32-cpython-312/src/integrator_saba.o                     
build/temp.emscripten_3_1_58_wasm32-cpython-312/src/integrator_sei.o build/temp.emscripten_3_1_58_wasm32-cpython-312/src/integrator_trace.o                          
build/temp.emscripten_3_1_58_wasm32-cpython-312/src/integrator_whfast.o build/temp.emscripten_3_1_58_wasm32-cpython-312/src/integrator_whfast512.o                   
build/temp.emscripten_3_1_58_wasm32-cpython-312/src/output.o build/temp.emscripten_3_1_58_wasm32-cpython-312/src/particle.o                                          
build/temp.emscripten_3_1_58_wasm32-cpython-312/src/rebound.o build/temp.emscripten_3_1_58_wasm32-cpython-312/src/rotations.o                                        
build/temp.emscripten_3_1_58_wasm32-cpython-312/src/server.o build/temp.emscripten_3_1_58_wasm32-cpython-312/src/simulationarchive.o                                 
build/temp.emscripten_3_1_58_wasm32-cpython-312/src/tools.o build/temp.emscripten_3_1_58_wasm32-cpython-312/src/transformations.o                                    
build/temp.emscripten_3_1_58_wasm32-cpython-312/src/tree.o -o build/lib.emscripten_3_1_58_wasm32-cpython-312/librebound.cpython-312-wasm32-emscripten.so             
emcc: error: undefined exported symbol: "___em_js__reb_remove_last_line" [-Wundefined] [-Werror]                                                                     
error: command '/tmp/tmpbfksie2y/cc' failed with exit code 1                                                                                                         
                                                                                                                                                                     
ERROR Backend subprocess exited when trying to invoke build_wheel                                                                                                    
[2024-11-05 15:17:22] Failed building package rebound in 10.7 seconds.                                                                                               
                                                                                                                                                                     
                                                                                                                                                                     
ERROR: cancelling buildall due to error building rebound
```

</details>

The patch applied in version 3.24.2 looks to be incorrect, since the `config_vars` and `extra_link_args` were being added for macOS, and not Emscripten. The logs are the same with or without the patch for version 4.4.3 (I haven't pushed that change because I had noticed the failure locally).

agriyakhetarpal:
> Thanks @agriyakhetarpal! That helps. I think I can see what's going on. If I just want to comment out this function call, what would be the easiest way to do that? Is there a precompiler flag that can tell me if it gets compiled for pyodide?

Shouldn't just guards around it with `#ifdef __EMSCRIPTEN__` work for this? I didn't get you, I'm afraid. Otherwise, if we want to check if things are being compiled for Pyodide, then checking for the `PYODIDE` environment variable's presence at build time is the way to do it.

In the meantime, I tried passing this to be exported via `EXPORTED_FUNCTIONS=["___em_js__reb_remove_last_line"]` to a list of `extra_link_args` next to `extra_compile_args` in `setup.py`, and tried to use `EMSCRIPTEN_KEEPALIVE` to not mangle this function, but that didn't help at all.

hannorein:
@hoodmane @agriyakhetarpal Yeah, it's a bit convoluted because of the different use cases. The c version of rebound can also be compiled with emcc (no python required). This is where I use the function in question. But `EM_JS` doesn't work if emcc just compiles it to a `.so` file (which is what is needed for the python package that pyodide creates). So again, I think it makes sense to just comment out that function call using the `PYODIDE` flag (but not the `__EMSCRIPTEN__` one). Let me know if I'm missing something. This is all still pretty new to me...

agriyakhetarpal:
> I think it makes sense to just comment out that function call using the PYODIDE flag (but not the __EMSCRIPTEN__ one)

Clarification: my experiment with the `PYODIDE` environment variable was inside `setup.py`, where `sys.platform == "emscripten"` won't return `True`, since we are cross-compiling from Linux/macOS. If we're compiling with Emscripten, then the `__EMSCRIPTEN__` macro would be inserted, of course. 

That said, I think @hoodmane would have a better idea on how to fix something like this, though. I've commented out the `reb_remove_last_line` function in relevant places and the compilation works as expected. However, the test in `test_rebound.py` is timing out quite consistently – there's a Pyodide fatal error that I am unable to access the full trace for.

`rebound` as a package via `/lib/python3.12/site-packages/librebound.cpython-312-wasm32-emscripten.so` does load, though.

agriyakhetarpal:
> I think it makes sense to just comment out that function call using the PYODIDE flag (but not the `__EMSCRIPTEN__` one)

Clarification: my experiment with the `PYODIDE` environment variable was inside `setup.py`, where `sys.platform == "emscripten"` won't return `True`, since we are cross-compiling from Linux/macOS. If we're compiling with Emscripten, then the `__EMSCRIPTEN__` macro would be inserted, of course. 

I think @hoodmane would have a better idea on how to fix something like this, though. I've commented out the `reb_remove_last_line` function in relevant places and the compilation works as expected. However, the test in `test_rebound.py` is timing out quite consistently – there's a Pyodide fatal error that I am unable to access the full trace for.

`rebound` as a package via `/lib/python3.12/site-packages/librebound.cpython-312-wasm32-emscripten.so` does load, though.

agriyakhetarpal:
Okay, I merged the `main` branch and brought in changes from #5163 – thanks for your assistance, @hannorein! That leaves us with river, Matplotlib, and PyArrow. I think some of the issues with Matplotlib in #4510 could be discovered and fixed from what we learn in matplotlib/matplotlib#29093. I'll be working on PyArrow in the meantime and will see if the in-tree tests for Matplotlib are easy fixes.

ryanking13:
Thanks for your work!

> PyArrow's case is an interesting one. Since PyArrow builds against a specific "released" Pyodide version in a separate repository, it can't build against NumPy v2 because this PR isn't merged and hasn't been a part of a release yet. We'll need to update the build workflow in https://github.com/joemarshall/pyarrow-pyodide for it to be able to PR branches such as this one and build Pyodide from them, too, and temporarily update the link to the wheel in the recipe here to that one in order to test it with NumPy v2. We'll need to think about this scenario in https://github.com/pyodide/pyodide/issues/4918. I'll look into it later in the day.

Yes, it is a bit tricky to handle those cases for now. Maybe after https://github.com/pyodide/pyodide-build/issues/43, it would become a bit easier to handle those cases, it's still going to take a while.

I would suggest 1) disable pyarrow in this PR and merge, 2) build pyarrow with latest xbuildenv (we publish it to http://pyodide-cache.s3-website-us-east-1.amazonaws.com/xbuildenv/dev/xbuildenv.tar.bz2). 3) update pyarrow in a separate PR.

agriyakhetarpal:
So, I've spent some time investigating the SciPy `logm` function signature mismatch that leads to the current timeout/fatal error, and I haven't gotten too far, so far. However, I did manage to isolate it to `scipy.linalg.schur`, which is calling the LAPACK function gees (`dgees`).

From a cursory look, it's not obvious to me where the signatures differ, and it's weird that this happens under NumPy 2 and not under NumPy 1.26 (the test passes on other PRs). I hope to look at it in more detail in some time (we've already had a patch to fix these functions in #3960, so maybe it's a matter of reworking it?).

agriyakhetarpal:
So, I've spent some time investigating the SciPy `logm` function signature mismatch that leads to the current timeout/fatal error, and I haven't gotten too far, so far. However, I did manage to isolate it to `scipy.linalg.schur`, which is calling the LAPACK function gees (`dgees`).

From a cursory look, it's not obvious to me where the signatures differ, and it's weird that this happens under NumPy 2 and not under NumPy 1.26 (the test passes on other PRs). I hope to look at it in more detail in some time (we've already had a patch to fix these functions in #3960, so _maybe_ it's just that we need to rework it?).

agriyakhetarpal:
The mismatch is exactly the same as the one reported earlier in #4310 which is why/where #4313 came in as a fix. The firstmost entry in the call stack is `$ndarray_from_pyobj (0003f1c2:0xb4ef)`, which suggests that something changed within `numpy.f2py` (maybe numpy/numpy#19388 – but that's too old?)

So, I'll try comparing the new SciPy behaviour versus the one for SciPy built against NumPy v1.26.4 later in the day.

agriyakhetarpal:
@ryanking13, may I ask how you inspected the output in https://github.com/pyodide/pyodide/issues/4310#issuecomment-1819040565? I've gotten down to the brink, and I have found that this is exactly the same error – I tried the experiment that you tried, and importing `scipy.integrate` before `scipy.linalg.logm` does not cause it to crash with the fatal error.

This is what I see in the WASM code where I set the breakpoint for:

```wasm
local.get $var0
global.get $env.__memory_base
i32.const 10812
i32.add
i32.load
i32.load offset=180
call_indirect (param i32) (result i32)
```

and the `null` reference with `pyodide._module.wasmTable.get(stack[1].value)` that was previously appearing does not appear after importing `scipy.integrate`. I thought that would provide more info on where to fix it, but instead, I see a really long internal WASM function defined as such: `(func $func498 (param $var0 i32) (result i32)`.

After having spent some time comparing the WASM instructions, I've come to find out that the `_npy_f2py_ARRAY_API` symbol defined in `f2py`'s [`fortranobject.h`](https://github.com/numpy/numpy/blob/7c0e2e4224c6feb04a2ac4aa851f49a2c2f6189f/numpy/f2py/src/fortranobject.h#L15) does not show up in NumPy version 2.1.3, but it does in version 1.26.4 – even with the `-fvisibility=default` compiler flag was added later in #4313. I am unsure how that flag helped in that PR, though – default visibility was the standard in Emscripten (I confirmed via the `emcc --cflags` command, which displays the default flags). I'm not sure how I should proceed further with making this symbol visible (what effects will public visibility have for Pyodide in specific?) but I'll keep looking for clues, still.

agriyakhetarpal:
> I don't remember exactly, but I think 1) I tried to look at the WASM code with debug build, 2) compared the imports/exports with `pyodide auditwheel imports/exports`, 3) searched some issues in scipy such as [scipy/scipy#15996](https://github.com/scipy/scipy/issues/15996), and experimented if making the symbol visible would help resolving the error.

Thanks, I tried similar things with 1) and 2), too, so I guess this is the right track to go.

> so I am not sure why the previous fix does not work here.

I've had a short discussion with @rgommers outside of this PR, and https://github.com/numpy/numpy/pull/26286 might be a potential cause.

agriyakhetarpal:
> I don't remember exactly, but I think 1) I tried to look at the WASM code with debug build, 2) compared the imports/exports with `pyodide auditwheel imports/exports`, 3) searched some issues in scipy such as [scipy/scipy#15996](https://github.com/scipy/scipy/issues/15996), and experimented if making the symbol visible would help resolving the error.

Thanks, I tried similar things with 1) and 2), too, so I guess this is the right track to go.

> so I am not sure why the previous fix does not work here.

I've had a short discussion with @rgommers outside of this PR, and https://github.com/numpy/numpy/pull/26286 might be a potential cause. I'll build against NumPy version 2.0.0rc1 and try to bisect further.

ryanking13:
This one looks suspicious too: https://github.com/numpy/numpy/pull/26103

ryanking13:
Adding the following flag to cflags seems to work.

```
-DNPY_API_SYMBOL_ATTRIBUTE=__attribute__((visibility("default")))
```

agriyakhetarpal:
Yes, that looks related as well. Since it landed in NumPy 2.1, 2.0 works (but I think it's better to update to the latest available).

rgommers:
Private symbols should be hidden, and `_npy_f2py_ARRAY_API` is definitely private. So that going missing is probably a symptom of a bug elsewhere in `numpy.f2py` or in `scipy`. See https://numpy.org/devdocs/reference/c-api/array.html#including-and-importing-the-c-api for what that define is for.

Or am I missing a Pyodide-specific reason here for why you want private symbols to be visible?

ryanking13:
> Or am I missing a Pyodide-specific reason here for why you want private symbols to be visible?

I have no idea too. For now, private symbols don't work for some reason I don't understand. Usually when I have a problem with Fortran, I blame it all on f2c and blindly believe that one day a Fortran compiler will fix that 😂

MaxHalford:
Hey! I just noticed River was mentionned. I'm one of its maintainers. Do you need me to do anything?

rgommers:
> For now, private symbols don't work for some reason I don't understand

If defining `-DNPY_API_SYMBOL_ATTRIBUTE=__attribute__((visibility("default")))` fixes the failing tests, it's specific to this single symbol and not all symbols. Which implies that it's a NumPy bug I think. If the Pyodide build is explicitly passing `-fvisibility=default`, that should be overriding the default behavior of trying to hide anything but a single `PyInit_<modulename>` symbol.

A separate bug report may be useful, with the actual failure included (I can't easily find it here). https://github.com/numpy/numpy/pull/26103 indeed seems like the most likely candidate that caused the problem, so another thing to do is to patch NumPy 2.1 in this branch, removing the `NPY_VISIBILITY_HIDDEN` usages that were added in that PR. If that fixes it, then we're certain that that change was the root cause.

agriyakhetarpal:
@MaxHalford, I've spent some time on this and I don't see any special failures when compiling on a macOS M-series device with NumPy v2.1.3 and Cython 3.0.11 beyond the solitary usage of a NumPy function `i1nd` that was deprecated in 2.0 (which was fixed easily via Ruff's NPY201 rule), and a bunch of failing doctests coming from NumPy 2.0's new data type promotion rules, i.e., something like what was an integer `85` previously does not get demoted and retains itself as `np.dtype64(85)`. Reading though the contributing guidelines for `river`, it's mentioned that that it's better to open a discussion first rather than a PR, so I've done that here: https://github.com/online-ml/river/discussions/1630. Please let me know your thoughts!

agriyakhetarpal:
I'm not sure what is up with the NumPy build; the `NPY_SIZEOF_COMPLEX_FLOAT` define seems to be in the right place indeed, so

```
Has header "complex.h" : YES                                                                                                                                          
Checking for type "complex float" : NO                                                                                                                                
                                                                                                                                                                      
../numpy/_core/meson.build:168:4: ERROR: Problem encountered: "complex.h" header does not include complex type complex float                                          
                                                                                                                                                                      
A full log can be found at /Users/agriyakhetarpal/Desktop/pyodide/packages/numpy/build/numpy-2.0.2/.mesonpy-6w4kva3w/meson-logs/meson-log.txt                         
                                                                                                                                                                      
ERROR Backend subprocess exited when trying to invoke build_wheel                                                                                                     
[2024-11-27 04:34:26] Failed building package numpy in 12.8 seconds.                                                                                                  
                                                                                                                                                                      
                                                                                                                                                                      
ERROR: cancelling buildall due to error building numpy  
```

It looks like something recent because commits from the week before have passed, but I haven't been able to isolate a cause yet.

packages/numpy/meta.yaml
```diff
@@ -18,16 +18,26 @@ build:
...
   post: |
     rm -rf numpy/distutils/__pycache__
+  # NumPy 2.1 disabled visibility for symbols outside of extension modules
+  # by default, so this breaks SciPy tests from modules that use f2py to
+  # build because they rely on the visibility of symbols in NumPy. This flag
+  # is currently used as a stop-gap measure. For more information, please see
+  # 1. https://github.com/numpy/numpy/pull/26286, and
+  # 2. https://github.com/numpy/numpy/pull/26103.
   cflags: |
+    -DNPY_API_SYMBOL_ATTRIBUTE=__attribute__((visibility("default")))
```
ryanking13, line 28: Oh, you need to set this flag on the scipy side, not on the numpy side. Sorry, I wasn't clear.

agriyakhetarpal: Thanks! For now, I'll downgrade to NumPy 2.0.2 and remove this completely, as you instructed in https://github.com/numpy/numpy/issues/27865#issuecomment-2503268541.

agriyakhetarpal:
There's a failure with building LightGBM in CI, which I'm unable to reproduce locally – LightGBM builds perfectly fine for me...

agriyakhetarpal:
I might need some help with fixing Matplotlib here, I disabled PyArrow earlier as requested. I was initially going to do a new release for `matplotlib-pyodide` yesterday and earlier in the day today, but on second thought(s), I don't think investing time into upgrading Matplotlib to 3.8.4 is helpful because there are a lot of versions, so what I've done is that I adapted all of the necessary changes in a patch to make Matplotlib 3.5.2 support NumPy v2 (@rgommers and I discussed this recently where it was noted that Matplotlib is most likely not doing something crazy with NumPy at build-time), and most of them are API-only (please see comments in the patch). I'm not sure if this is the right approach to go forward with.

rgommers:
> I don't think investing time into upgrading Matplotlib to 3.8.4 is helpful because there are a lot of versions, so what I've done is that I adapted all of the necessary changes in a patch to make Matplotlib 3.5.2 support NumPy v2 (@rgommers and I discussed this recently where it was noted that Matplotlib is most likely not doing something crazy with NumPy at build-time), and most of them are API-only (please see comments in the patch). I'm not sure if this is the right approach to go forward with.

I'm not sure about the details here, but I'd always ask "is this useful long term". Patching an old version of Matplotlib is in principle effort that only has a net negative effort on future work, especially if you're trying to patch package versions together that don't actually work together for regular installs from PyPI. So it seems wrong to me from first principles. I don't understand what the problem is with upgrading Matplotlib to latest, and then dealing with whatever is left (that's then Pyodide-specific issues)?

agriyakhetarpal:
> I'm not sure about the details here, but I'd always ask "is this useful long term". Patching an old version of Matplotlib is in principle effort that only has a net negative effort on future work, especially if you're trying to patch package versions together that don't actually work together for regular installs from PyPI. So it seems wrong to me from first principles. I don't understand what the problem is with upgrading Matplotlib to latest, and then dealing with whatever is left (that's then Pyodide-specific issues)?

Upgrading Matplotlib to version 3.8.4 or later does bring Pyodide-specific issues because of the gap between versions – that is, more function signature mismatches with `matplotlib-pyodide`'s WASM backend, in which no one has claimed expertise. I can try upgrading incrementally (to 3.6, 3.7.3, etc.) first, maybe that can help.

rgommers:
> Upgrading Matplotlib to version 3.8.4 or later does bring Pyodide-specific issues

It's very well possible that it's (too) hard right now, but this upgrade has to be done anyway sooner or later, so it's not wasted effort. I'd suggest at least trying, and otherwise do the maximum version bump you can get done and report blockers as new issues, before patching an old version.

agriyakhetarpal:
### In the last few commits

- 79a62e5e0de396fc72c3e1bbe97048ffab0e4af7 drops the Matplotlib patch because we upgraded Matplotlib to version 3.8.4 and `matplotlib-pyodide` to 0.2.3, in #4510
- 65571107f54bbeca3da124ad8cda1509bde9dc3e bumps `scikit-image` to https://github.com/scikit-image/scikit-image/releases/tag/v0.25.0 from a few days ago. Tagging the recipe maintainers @lagru and @mkcor for visibility around this. We can split it into a different PR at any time if needed, of course.
- 89bf1b86330d82e715685c0df4121721e1152216 triggered builds and test suite run for SciPy to ensure the `logm` null function situation is working again

### Next few action items

- [ ] Re-enabling packages that were disabled here in subsequent PRs
    - [ ] LightGBM has had consistent failures due to a concoction of builds for other packages interfering, and we haven't been able to figure out why – perhaps isolate it into its own job like OpenCV until we unvendor all the recipes via #4918, which will be worked on after 0.27 stable
    - [ ] PyArrow – already at a higher version than what is needed for NumPy 2.0 compatibility, so this is straightforward and we have no problems here – we just need to rebuild it in https://github.com/joemarshall/pyarrow-pyodide against NumPy 2.0. This can be done by temporarily adding a recipe for NumPy 2.0, merging it, issuing a new GitHub release with the wheel there, and updating the URL to it here. I'll start by opening a PR for this there.
- [ ] Review backlog for suitable PRs before the release
- [ ] 0.27.0 release  – #5064

hoodmane:
If lightgbm is the last blocker we can disable it and reenable in 0.27.1

juntyr:
I can confirm that pyarrow builds without issues (on my pyodide fork which is based on the latest main with today's matplotlib and moves the pyarrow recipe in-tree)

agriyakhetarpal:
I don't think it's a blocker in the sense that we need to fix anything to make it build; it's just that it builds correctly on around 2/10 occasions. But, isolating it at the cost of CI time is something we can do, so that it doesn't need to be removed from the distribution.

agriyakhetarpal:
> I can confirm that pyarrow builds without issues (on my pyodide fork which is based on the latest main with today's matplotlib and moves the pyarrow recipe in-tree)

Thanks, @juntyr! Noted, good to know :D

juntyr:
> > I can confirm that pyarrow builds without issues (on my pyodide fork which is based on the latest main with today's matplotlib and moves the pyarrow recipe in-tree)
> 
> Thanks, @juntyr! Noted, good to know :D

Here's the full recipe if you want a quick copy-paste

```yaml
package:
  name: pyarrow
  version: 18.1.0
  top-level:
    - pyarrow
source:
  url: https://github.com/apache/arrow/releases/download/apache-arrow-18.1.0/apache-arrow-18.1.0.tar.gz
  sha256: 2dc8da5f8796afe213ecc5e5aba85bb82d91520eff3cf315784a52d0fa61d7fc
  patches:
    - patches/autoload_timezones.patch
build:
  script: |
    # move things around so that the top level folder is the python folder
    # with the pyproject.toml, so pyodide build works
    cd ..
    # build libarrow as static library without the pyodide toolchain file
    # because that does bad things to the build
    BACKUP_TOOLCHAIN=${CMAKE_TOOLCHAIN_FILE}
    unset CMAKE_TOOLCHAIN_FILE
    if [ ! -d "${PKG_BUILD_DIR}_cpp" ]; then
      mv $PKG_BUILD_DIR ${PKG_BUILD_DIR}_cpp
      cp -r ${PKG_BUILD_DIR}_cpp/python $PKG_BUILD_DIR
      # build arrow cpp library and install to emscripten sysroot
      # n.b. the if statements below mean that you can call pyodide build-packages-no-deps
      # pyarrow --continue to debug build errors without having to do a full rebuild
      mkdir cppbuild
      cd cppbuild
      emcmake cmake --preset ninja-release-emscripten ${PKG_BUILD_DIR}_cpp/cpp
    else
      cd cppbuild
    fi
    cmake --build . --target=protobuf_ep_host -j 4
    cmake --build . --target=install -j 4
    cd ${PKG_BUILD_DIR}
    export CMAKE_TOOLCHAIN_FILE=${BACKUP_TOOLCHAIN}
    rm -f CMakeCache.txt
    rm -rf CMakeFiles
  cross-script: |
    cmake --version
about:
  home: https://arrow.apache.org/
  PyPI: https://pypi.org/project/pyarrow
  summary: Python library for Apache Arrow
  license: Apache License, Version 2.0
requirements:
  run:
    - numpy
    - pandas
    - pyodide-unix-timezones
  host:
    - numpy
extra:
  recipe-maintainers:
    - joemarshall
```

hoodmane:
> we can always build LightGBM's wheels and store them in a GitHub release

Sounds good to me.

agriyakhetarpal:
Looks like we are good to merge now. Surfacing @ryanking13's suggestion from above:

> I would suggest 1) disable pyarrow in this PR and merge, 2) build pyarrow with latest xbuildenv (we publish it to http://pyodide-cache.s3-website-us-east-1.amazonaws.com/xbuildenv/dev/xbuildenv.tar.bz2). 3) update pyarrow in a separate PR.

I'll be doing this in follow-up PRs for PyArrow and LightGBM, and we don't need to wait for them for this PR. Thanks!

agriyakhetarpal:
I confirmed that all tests for all packages are passing by temporarily removing the `--skip-passed` flag we have in `pytest_wrapper.py` so that we run them. I've reverted that commit now. If there are any reports of packages that don't correctly, we always have the provision for a 0.27.1 release.

Makefile.envs
```diff
@@ -32,7 +32,7 @@ export PYSTABLEVERSION=$(PYMAJOR).$(PYMINOR).$(PYMICRO)
 export HOSTPYTHONROOT=$(shell python${PYMAJOR}.${PYMINOR} -c "import sys; print(sys.prefix)")
 export HOSTPYTHON=$(HOSTPYTHONROOT)/bin/python$(PYMAJOR).$(PYMINOR)
 
-export PYTHON_ARCHIVE_URL=https://www.python.org/ftp/python/$(PYSTABLEVERSION)/Python-$(PYVERSION).tgz
+export PYTHON_ARCHIVE_URL=https://registry.npmmirror.com/-/binary/python/$(PYSTABLEVERSION)/Python-$(PYVERSION).tgz
```
hoodmane, line 35: What's this for?

packages/netcdf4/test_netcdf4.py
```diff
@@ -15,29 +15,29 @@ def test_netCDF4_tutorial(selenium):
 
     stdouts = [
         "NETCDF4",
-        "<class 'netCDF4._netCDF4.Dataset'>\nroot group (NETCDF4 data model, file format HDF5):\n    dimensions(sizes): \n    variables(dimensions): \n    groups: forecasts, analyses",
```
hoodmane, line 18: This seems like a pretty bad test. Ideally we shouldn't care about the exact repr of all this junk. No action needed here, but in the future would be great to remove it or replace it with a better test.

hoodmane (approved):
Thanks so much @agriyakhetarpal for all your hard work on this!

packages/netcdf4/test_netcdf4.py
agriyakhetarpal: I agree, it took me a while to see what was going on here!

.gitattributes
```diff
@@ -0,0 +1,2 @@
+# Configure LF line endings for text files
```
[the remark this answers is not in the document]
agriyakhetarpal: Ah, it's been a while since working on this – I might have messed it up from a different PR and added it here instead. I think it was for SciPy's BLAS signatures text file. It's nice to include to ensure we have consistent line endings across platforms, but I'll do this in a separate PR. Removed in ce7e6ccfaf29e591b720bc62d49b302f933a6e6f.

agriyakhetarpal:
Actually, PyArrow requires a small patch to build it with NumPy 2.0, which I am working on upstreaming to them.

agriyakhetarpal:
Actually, PyArrow requires a small patch to build it with NumPy 2.0, which I am working on upstreaming.

agriyakhetarpal:
I started https://github.com/pyodide/pyodide-numpy-2.0-rebuilds as a repository with my own account to build LightGBM and PyArrow separately, and upon transfer to the Pyodide organisation, GitHub decided to revoke my access to the repository's settings completely for some reason, just before I was going to finalise and release the built wheels (it's much better for the community for the wheels to point to a Pyodide repository rather than mine). We can archive the repository a bit later when we have a permalink to release artifacts.

We also have the option of using the "dev" xbuildenv, as @ryanking13 had mentioned in https://github.com/pyodide/pyodide/pull/4925#issuecomment-2466043612. The only reason why I didn't pursue this was because it, too, requires the CI job on the `main` branch to pass fully, and LightGBM is [currently failing](https://circleci.com/gh/pyodide/pyodide/121377) there again – so we cannot rely on it and can't keep re-triggering it until it does pass.

@hoodmane, could you please make me an administrator for that repository (or grant the "core" team administrator access to repositories, whichever is simpler)? I will never understand teams on GitHub 😅

ryanking13:
@agriyakhetarpal I gave you admin role for `pyodide-numpy-2.0-rebuilds` repository!

agriyakhetarpal:
> @agriyakhetarpal I gave you admin role for `pyodide-numpy-2.0-rebuilds` repository!

Thanks! I went ahead with it, and a release is now up: https://github.com/pyodide/pyodide-numpy-2.0-rebuilds/releases/tag/v1.0.0. Attestations to verify the build provenance of the wheels are also included. I guess we can go ahead with merging this PR because all tests are passing, and focus on adding these two wheels in follow-ups (#5265 and #5266).

agriyakhetarpal:
Thank you, all, for those involved with the code review and guidance – and especially the recipe maintainers, for their relentless assistance with package updates.

PR #4937, opened 2024-07-15
Fix OOT incompatible recipes

ryanking13:
### Description

Update recipes that had issues building or testing out-of-tree.

- sharedlib-test-py: Fix relative path
- pygame: Fix wheel path
- cpython_modules: hatchling ignores files specified in [`.gitignore` by default](https://hatch.pypa.io/1.12/config/build/#vcs). It resulted in not including the `.so` file in the wheel.

### DIFF
--- a/packages/hashlib/src/pyproject.toml
+++ b/packages/hashlib/src/pyproject.toml
@@ -10,7 +10,11 @@ version = "1.0.0"
 requires = ["hatchling"]
 build-backend = "hatchling.build"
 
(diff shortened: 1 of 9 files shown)


agriyakhetarpal (approved):
Thanks for the fixes, @ryanking13! I'm happy to help out with reviews where I can, and these changes look good to me.

ryanking13:
Thanks for the review Agriya!

PR #4935, opened 2024-07-15
Run scipy tests as part of the Github Action CI

lesteve:
### Description

It would make a lot of sense to have Pyodide developers be able to run Scipy tests rather than asking me to do it manually based on https://github.com/lesteve/scipy-tests-pyodide when needed. Last time this happened is for the f2c fork PR https://github.com/pyodide/pyodide/pull/4920#issuecomment-2217581640. You could imagine that the Scipy 1.13 PR https://github.com/pyodide/pyodide/pull/4719 would benefit from this as well as potentially many other PRs.

I am testing this on my fork, see [GH logs from my PR branch](https://github.com/lesteve/pyodide/actions/workflows/main.yml?query=branch%3Afull-scipy-tests) in particular this [GH log](https://github.com/lesteve/pyodide/actions/runs/9937311988) where the scipy tests ran without error.

This is a bit of a draft PR as there are a few questions:
1. when do we want to trigger the full Scipy tests? There was a Hood's [comment](https://github.com/pyodide/pyodide/pull/4499#issuecomment-1936361540) that suggested when some files changes in a few folders e.g. scipy, f2c, etc .... do we want to always run it on main? Do we want a commit-based trigger like `[full-scipy]` (in scikit-learn we use them for a few things, like forcing a full doc-build, running the Pyodide CI in a PR, etc ...)
2. this would simplify my life if it was OK to to always build scipy in the "core" build (I would estimate maybe 15-20 minutes added to build time) ... tests would only be run on some specific triggers. Right now, I separated scipy build from the core build but right now it is compiling everything again. I guess because I need to add more things in the cache although I don't know exactly which ones (emsdk, cpython, packages at the very least). I guess CircleCI does that with workspaces. 
3. test strategy is the same as the one I currently use in https://github.com/lesteve/scipy-tests-pyodide, i.e. run with node and a js wrapper script. One of the reason of still having this wrapper script is that I have some work-arounds to remove some Scipy test files whose import is problematic.

Side-comment: I am using `-m 'not slow'` because otherwise I saw some weird memory issue towards the end of Scipy tests, see this [build log](https://github.com/lesteve/pyodide/actions/runs/9934164081/job/27439073308) where some tests can not allocate enough memory for the numpy arrays. Locally I get a Pyodide fatal error at 97% of the test, not sure whether this is related.

<details>

<summary>One of the error</summary>

```
 _______________ TestGoodnessOfFit.test_against_anderson_gumbel_r _______________

[...]

PR #4938, opened 2024-07-16
Update codespell version in pre-commit

lesteve:
### Description

This updates codespell in precommit. This would help for https://github.com/pyodide/pyodide/pull/4935 because `precommit.ci` is unhappy and I don't manage to make it ignore "COO" for an unkown reason ...

I needed to add an ignore-list because somehow `atmost` is flagged by pre-commit 2.3.0 as a typo ...

codespell 2.3.0 detected real typos in other files that I fixed.

### Checklists

I ticked all of them since this does not seem applicable

- [X] Add / update tests


tools/codespell_ignore_words.txt
```diff
@@ -0,0 +1,2 @@
+atmost
+COO
```
hoodmane, line 2: We also have `"--ignore-words-list", "ags,aray,asend,ba,classs,crate,falsy,feld,inflight,lits,nd,slowy,te,oint,conveniant"` in the `.pre-commit-config` file, presumably we don't want to have both an inline ignore list and a separate file.

lesteve: Ah I missed it somehow, so many different ways to configure these things ...

I pushed a commit moving the config to `pyproject.toml`. I guess one small advantage is that you can run `codespell` directly if you want and it will use the config rather than having to run `precommit run codespell --files <file>`. I can definitely revert it if you like to have everything in `precommit`

pyproject.toml
```diff
@@ -128,4 +128,4 @@ markers = [
 
 [tool.codespell]
 ignore-words = 'tools/codespell_ignore_words.txt'
-skip = 'benchmark/benchmarks/pystone_benchmarks/pystone.py,src/js/package-lock.json'
+skip = 'benchmark/benchmarks/pystone_benchmarks/pystone.py,src/js/package-lock.json,tools/codespell_ignore_words.txt'
```
hoodmane, line 131: nice

tools/codespell_ignore_words.txt
hoodmane: > so many different ways to configure these things

Seriously.

> moving the config to pyproject.toml

Sounds good to me, I don't have a strong preference as long as we keep it consistent.

hoodmane (approved):
Thanks!

PR #4936, opened 2024-07-15
Add CasADi

agriyakhetarpal:
### Description

This PR aims to add [CasADi](https://web.casadi.org/), a C++ library with SWIG-generated bindings to Python. CasADi is a framework for Python and MATLAB (and Octave) which provides tools for nonlinear optimization and algorithmic differentiation, and offers interfaces to several solver suites, which are not required but can be turned on via CMake flags. Just the SUNDIALS solver suite is required to build, and the rest of the interfaces are optional (they are not enabled because of the high build times).

I figured that it would be easier to build off of the source distribution available on PyPI, since I earlier misunderstood Pyodide's build process when I was very new to it – I was trying to wrap my head around CasADi's incredibly complex build system that first generates a standards-compliant source distribution with a `setup.py` file. However, we can build from the source distribution, through which compiling a wheel is possible and is even tested in CI in the CasADi repository on GitHub.

Building this package is an incredibly complex task, and its wheels are large (~40 MB), at least on PyPI. Not having the aforementioned optional interfaces, such as those for MUMPS, IPOPT, SNOPT, BONMIN, other MILP/QP solvers, jitted OpenCL functions, etc., can reduce the impact of this constraint.

Closes #4551

### Checklists

- [x] Add a [CHANGELOG](https://github.com/pyodide/pyodide/blob/main/docs/project/changelog.md) entry
- [x] Add / update tests

Issue #4551, opened 2024-02-23
Package request: add the CasADi framework


agriyakhetarpal:
## 🐍 Package Request

- Package Name and Version : `casadi` (version 3.6.4 and later)
- Package URL : https://github.com/casadi/casadi, https://pypi.org/project/casadi/
- Package Dependencies that needs to be resolved first: CasADi is primarily written in C++ with bindings for Python and MATLAB, and AFAIK it does not mark dependencies on any Python packages, with the exception of NumPy. It does [provide interfaces](https://github.com/casadi/casadi/tree/main/casadi/interfaces) to a lot of C/C++ based solvers, however, such as BONMIN, IPOPT, NLOPT, SUNDIALS, Alpaqa, HiGHS, and more. The source build instructions for CasADi are available here: https://github.com/casadi/casadi/wiki/InstallationLinux

hoodmane:
If you work on this, we will answer your questions.

hoodmane:
What happens if you clone it and run `python -m build .` in the source directory? Presumably you're saying that doesn't work?

hoodmane:
Looks like the sdist at least includes a `setup.py`, so being optimistic, maybe they already did the extra weird steps before making the sdist and it won't be a problem?

agriyakhetarpal:
I just tried compiling it with the following command:

```bash
$ pip install casadi --no-binary casadi
```

<details><summary>Expand to view logs
</summary>
<p>

```bash
casadi-pyodide % 
Collecting casadi
  Downloading casadi-3.6.4.tar.gz (5.1 MB)
     ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 5.1/5.1 MB 2.5 MB/s eta 0:00:00
  Preparing metadata (setup.py) ... done
Collecting numpy (from casadi)
  Using cached numpy-1.26.4-cp311-cp311-macosx_11_0_arm64.whl.metadata (114 kB)
Using cached numpy-1.26.4-cp311-cp311-macosx_11_0_arm64.whl (14.0 MB)
Building wheels for collected packages: casadi
  Building wheel for casadi (setup.py) ... error
  error: subprocess-exited-with-error
  
  × python setup.py bdist_wheel did not run successfully.
  │ exit code: 1
  ╰─> [130 lines of output]
      running bdist_wheel
      running build
      running build_ext
[...]

PR #4947, opened 2024-07-17
Add `pythoninclude` to `pyodide config` CLI

agriyakhetarpal:
### Description

This PR is based on https://github.com/pyodide/pyodide/issues/4551#issuecomment-2229172277. I thought of adding `python_interp_include_dir`, but that name felt too long, so I switched to `python_include_dir`, and I then noticed that `pythoninclude` already exists as a variable that is getting computed through `Makefile.envs`. So, instead of renaming it from `pythoninclude` to `python_include_dir` and inducing an unneeded breaking change, I've just exposed it verbatim here, such that it is both accessible through the `pyodide config get pythoninclude` command for out-of-tree builds and visible in the list of default variables through the `pyodide config list` command.

It is to be noted that `emscripten_version` is the only variable that seems to be in active use for the `pyodide config` CLI based on a quick [GitHub code search](https://github.com/search?q=%22pyodide+config+get%22&type=code); perhaps this is because it is the only one that has been [documented](https://pyodide.org/en/stable/development/building-and-testing-packages.html#set-up-emscripten) before, so even if we were to rename and switch up a few variables, we should be fine (but I would prefer not to, personally).

The rationale behind this change is that some packages might need to read through the [Python headers](https://github.com/python/cpython/tree/main/Include) for some reason or the other, such as #4936. The Emscripten toolchain already has this directory passed to it through the use of the `-I$(PYTHONINCLUDE)` flag, and there is no need for any package to link against Python itself, so I'm unsure if this is an entirely beneficial patch – please let me know! At least, the change is simple and shouldn't cause any regressions, so those who wish to use this variable for their needs can do so through this PR. The use cases I've thought of so far are when compiled extensions that are defined in a package's `setup.py` rely more on the `setuptools` machinery to pass options to CMake, rather than having CMake do its thing with `FindPython`. However, we never know when or if more packages might benefit from using it, so IMO it feels okay to add.

### Checklists

### DIFF
--- a/docs/project/changelog.md
+++ b/docs/project/changelog.md
@@ -16,6 +16,12 @@ myst:
 
 ## Unreleased
 
+- {{ Enhancement }} The `pyodide config list` command now shows an additional
+  config variable `pythoninclude` that can be used to access the location of
+  the Pyodide Python interpreter headers. This variable can be accessed through
+  `pyodide config get pythoninclude` as well for use in out-of-tree builds.
+  {pr}`4947`
+
 - {{ Enhancement }} Add unix-timezones module, which installs Unix compatible
   timezone data in /usr/share/zoneinfo, for use with C/C++ libraries which do
   timezone handling.
--- a/pyodide-build/pyodide_build/cli/config.py
+++ b/pyodide-build/pyodide_build/cli/config.py
@@ -9,6 +9,7 @@ app = typer.Typer(help="Manage config variables used in pyodide")
 PYODIDE_CONFIGS = {
     "emscripten_version": "PYODIDE_EMSCRIPTEN_VERSION",
     "python_version": "PYVERSION",
+    "pythoninclude": "PYTHONINCLUDE",
     "rustflags": "RUSTFLAGS",
     "cmake_toolchain_file": "CMAKE_TOOLCHAIN_FILE",
     "pyo3_config_file": "PYO3_CONFIG_FILE",
--- a/pyodide-build/pyodide_build/config.py
+++ b/pyodide-build/pyodide_build/config.py
@@ -211,8 +211,9 @@ DEFAULT_CONFIG_COMPUTED: dict[str, str] = {
     "pyo3_cross_include_dir": "$(PYTHONINCLUDE)",
     # Misc
     "stdlib_module_cflags": "$(CFLAGS_BASE) -I$(PYTHONINCLUDE) -I Include/ -I. -IInclude/internal/",  # TODO: remove this
-    # Paths to build dependencies
+    # Paths to build dependencies and config files
     "host_install_dir": "$(PYODIDE_ROOT)/packages/.artifacts",
     "host_site_packages": "$(PYODIDE_ROOT)/packages/.artifacts/lib/python$(PYMAJOR).$(PYMINOR)/site-packages",
+    "pythoninclude": "$(PYTHONINCLUDE)",
     "numpy_lib": "$(PYODIDE_ROOT)/packages/.artifacts/lib/python$(PYMAJOR).$(PYMINOR)/site-packages/numpy/",
 }


hoodmane:
If packages need this in `setup.py` shouldn't they do:
```py
import sysconfig

print(sysconfig.get_config_var("INCLUDEPY"))
```

hoodmane:
Perhaps we should add a way to look up Pyodide's sysconfig variables from the cli?

agriyakhetarpal:
I think so too. Do you mean something like `pyodide config get sysconfig::variable`?

agriyakhetarpal:
> If packages need this in `setup.py` shouldn't they do:
> 
> ```python
> import sysconfig
> 
> print(sysconfig.get_config_var("INCLUDEPY"))
> ```

I think `sysconfig.get_path('include')` and `sysconfig.get_config_var('INCLUDEPY'))` are equivalent – both return the same thing.

hoodmane:
There is no variable called `include` or `INCLUDE` you'll just get `None`. But `sysconfig.get_config_var("INCLUDEDIR")` points up a level. Here are all of the related variables in a native Debian install of Python 3.12:
```
INCLUDEDIR ==> /usr/include
INCLUDEPY ==> /usr/include/python3.12
CONFINCLUDEPY ==> /usr/include/python3.12
INCLDIRSTOMAKE ==> /usr/include /usr/include /usr/include/python3.12 /usr/include/python3.12
```

agriyakhetarpal:
Oh, I think you (might) have misread – it's `sysconfig.get_path()`, it returns the values in `sysconfig.get_paths()`: https://docs.python.org/3/library/sysconfig.html#sysconfig.get_path, which are as follows:

```
{'data': '/opt/homebrew/Caskroom/miniforge/base',
 'include': '/opt/homebrew/Caskroom/miniforge/base/include/python3.10',
 'platinclude': '/opt/homebrew/Caskroom/miniforge/base/include/python3.10',
 'platlib': '/opt/homebrew/Caskroom/miniforge/base/lib/python3.10/site-packages',
 'platstdlib': '/opt/homebrew/Caskroom/miniforge/base/lib/python3.10',
 'purelib': '/opt/homebrew/Caskroom/miniforge/base/lib/python3.10/site-packages',
 'scripts': '/opt/homebrew/Caskroom/miniforge/base/bin',
 'stdlib': '/opt/homebrew/Caskroom/miniforge/base/lib/python3.10'}

```

ryanking13 (approved):
Thanks @agriyakhetarpal!

Could you open a same PR in [pyodide/pyodide-build](https://github.com/pyodide/pyodide-build) too, after merging this? Or if this patch isn't required right now, it is okay to close this and open it only in pyodide-build repository as I hope I'll be able to merge #4882 in a few days/weeks.

---

> The rationale behind this change is that some packages might need to read through the [Python headers](https://github.com/python/cpython/tree/main/Include) for some reason or the other, such as https://github.com/pyodide/pyodide/pull/4936

Yes, I agree that exposing this directory would be useful.

> If packages need this in setup.py shouldn't they do:

I think it is also common to get these variables in the CLI. For instance there is a `python3-config` CLI for that ([man page](https://helpmanual.io/man1/python3-config/)). So I think providing a similar way of getting those values for cross-compilation seems reasonable to me.

agriyakhetarpal:
Thanks for the review and the additional context, @ryanking13! I don't think the patch is required right now, and I wish to follow up with @hoodmane's suggestion on looking up Pyodide's sysconfig variables from the CLI, i.e., the ones in `sysconfig.get_paths()` as noted above in https://github.com/pyodide/pyodide/pull/4947#issuecomment-2235795955. Loading them dynamically makes sense to me. I can wait for you to merge #4882, and I'll push these changes to https://github.com/pyodide/pyodide-build after that.

Issue #4942, opened 2024-07-16
Contents of Release Files could be explained


Hephaistos7:
## 📚 Documentation

So I downloaded the latest release:  _pyodide-0.26.1.tar.bz2_ . But this Issue is relevant for all releases.

Inside of a release, I expect to find all the necessary files for running Pyodide, and all the wheels to run libraries like NumPy, Matplotlib and many more.

What I don't understand is:
- why are there html-files, like _test.html_ or _console.html_?
- why do some wheels have tests.tar-files, like _gsw-tests.tar_?
- why are some packages zipped, like _hashlib-1.0.0.zip_?
- why is there a font folder?
- why are there JS-files like _webworker.js_, that are more meant for testing than really using Pyodide?
- why is there a _package.json_, as this is not a npm-Package?
- and what do the .metadata files do?

I don't understand what most of these do and why they need to be in a release.

**My use case**:
So, I use Pyodide in my project and recently switched from loading from the CDN to including the release in my project. This allows me to build an offline-version that doesn't need to load e.g. NumPy from the internet.
In using  _pyodide-0.26.1.tar.bz2_ for this purpose, I feel like I include a lot of clutter into my project. So I manually have to delete all the files that don't do anything for my project.

Should there maybe be a release that does not include any test-files?

Greetings,
Jeff

PS: absolutely love the project!

hoodmane:
Would indeed make sense to explain the files. Have you looked at [pyodide-core-0.26.1.tar.bz2](https://github.com/pyodide/pyodide/releases/download/0.26.1/pyodide-core-0.26.1.tar.bz2)?

Hephaistos7:
So here is an overview of my understanding:

 **pyodide-core-0.26.1.tar.bz2** contains the following files:
 - pyodide.asm.js
 - pyodide.asm.wasm
 - pyodide.js
 - pyodide.mjs
 - python_stdlib.zip
 - package.json
 - pyodide-lock.json
 
pyodide-0.26.1.tar.bz2 contains these file tpyes on top of that:
- fonts folder
- snapshot.bin
- python file
- 3 html-files
- module_webworker_dev.js
- webworker.js
- webworker_dev.js
- makesnap.mjs
- pyodide.js.map
- pyodide.mjs.map
- 227 metadata files of python wheels
- 63 *-tests.tar files of various python packages
- ffi.d.ts
- pyodide.d.ts
- 227 python wheels for wasm
- 18 zip archives

[...]

PR #4390, opened 2024-01-19
DOC Add community examples section in the docs

ryanking13:
### Description

Adds a `Community Examples` section in the docs so users can contribute examples using Pyodide.
I added a template file that people can start with (probably we can improve it), and an very simple example using black formatter in Pyodide.

To make it easier for people to use the example, I've included a comment that we recommend that the example to be a single HTML file. Probably we can allow sppliting the HTML and .py file for readability, but I am not sure readthedocs supports hosting `.py` file.

### Checklists

- [x] Add new / update outdated documentation

PR #4379, opened 2024-01-15
Added Python Sandbox example

kor0p:
### Description

I want to help with #1498
There was a suggestion to create Python sandbox (like jsFiddle) to show Pyodide example without server
So, that's it :)
Example is deployed with GitHub Pages on https://kor0p.github.io/pyodide/examples/sandbox/

I didn't find "examples" directory, so created new one 😅
As mentioned in #1498, there should be tests for examples, so when idea will be approved, I'll create some tests for it, okay?

### Checklists

- [ ] Add tests

### Notes
An example is created using my little framework [BeePy](https://github.com/kor0p/BeePy), built on Pyodide. It has only main page in documentation and has no tests, but I'm working on it. I started this pet-project in Nov 2021, as a Junior, but I have big plans for it. Thanks for your work on Pyodide ♥️

PR #4951, opened 2024-07-19
added webworker version of console.html

joemarshall:
This adds another version of console.html which runs pyodide in a webworker. 

It can't be used for things that need to use js to talk to the window object, but for just python use, it is nice because you can e.g. run pytest tests and see the output in real time. 

No input or interrupt handling right now, although with jspi I guess it would be possible to do without too much hassle.


ryanking13:
Thanks @joemarshall. It's nice to have an example of using webworker, but I'm not sure if we should include it in the final distribution.

I'd like to reduce the files included in the distribution to only what is necessary (#4942), so IMO this kind of file would be better added to the example section of the documentation (#4379, #4390).

joemarshall:
Great - what's the state of examples? Are you going to merge #4390 as that seems like the obvious place for this. 

Is console.html going to move there also? Could maybe also add a `make examples` commandline that made examples as a subdirectory of dist (or in dist), so that anyone who wanted to test things could put all the examples into there automatically and ready to serve; it's nice how console.html etc. just turn up when you do a build, but I can see it doesn't need to be in a distribution by default any more.

I also have `run_pytest_in_pyodide.py` to contribute to examples also, which runs a full pytest test suite inside pyodide from command line (without requiring pytest collection to happen on the host machine like pytest-pyodide does). Used in arrow CI, and quite fiddly to write because of handling echoing of stdout / stderr correctly to host console in real-time requires a certain amount of fiddling in javascript + selenium (it's easy to make it output at the end of tests only, but then it just hangs while the pyodide call goes in, which is sub-optimal)

ryanking13:
> Are you going to merge https://github.com/pyodide/pyodide/pull/4390 as that seems like the obvious place for this.

@hoodmane @agriyakhetarpal WDYT? Are you okay with having examples section in our docs? Note that there was some discussion about it: https://github.com/pyodide/pyodide/pull/4379.

agriyakhetarpal:
Hi @ryanking13, thanks for the ping. I've gone through the previous discussions linked, and it's certainly useful to have an examples section in our own docs, +1 for merging that PR. There are some other examples with Pyodide that I know such as the one on https://learn.scientific-python.org/development/guides/repo-review/ by the Scientific Python cookiecutter template authors, which would be useful to link to. That said, having a collection of assorted/curated snippets that can perform regular tasks hosted on a separate website would be a great idea as well, similar to https://sphinx-extensions.readthedocs.io/en/latest/.

ryanking13:
Alright, let's then merge this, so we can put examples in the Pyodide docs.

ryanking13:
Alright, then, let's merge #4390 so that we can put examples like this in our docs.

joemarshall:
@ryanking13 I've moved the console webworker to examples now. It uses a separate .js file, so I slightly modified the instructions for examples to say a single .html file, or a single html plus a same name javascript file, just because workers without separate JS files are always a bit of a hack.

joemarshall:
Do examples need a changelog entry, or can we treat them as minor?

hoodmane:
I don't think examples need changelog entries.

Makefile
```diff
@@ -253,6 +253,14 @@ dist/console.html: src/templates/console.html
...
 	sed -i -e 's#{{ PYODIDE_BASE_URL }}#$(PYODIDE_BASE_URL)#g' $@
 
+dist/console_webworker.html: src/templates/console_webworker.html
+	cp $< $@
+	sed -i -e 's#{{ PYODIDE_BASE_URL }}#$(PYODIDE_BASE_URL)#g' $@
+
+dist/console_webworker.js: src/templates/console_webworker.js
+	cp $< $@
+	sed -i -e 's#{{ PYODIDE_BASE_URL }}#$(PYODIDE_BASE_URL)#g' $@
+
```
ryanking13, line 263: This can be removed now.

ryanking13 (commented):
Thanks @joemarshall!

PR #4959, opened 2024-07-24
Backports for 0.26.2

hoodmane:
Merge don't squash!


ryanking13 (approved):
Thank you for the version up work!

A minor concern for releasing 0.26.2 is that it will release pyodide-build 0.26.2 to PyPI which is slightly diverged from pyodide/pyodide-build. But there isn't user-facing breaking changes, so I guess it is okay.

PR #4961, opened 2024-07-24
PERF Don't format attribute error in JsProxy_GetAttr

hoodmane:
This initial `PyObject_GenericGetAttr` is expected to fail, but it formats an expensive AttributeError whenever it does. The internal method `_PyObject_GenericGetAttrWithDict` has a `suppress` argument that makes it not do this.

This saves ~36% of the time spent in `JsProxy_GetAttr`.

Tested on the profile:
```js
o = {f(){}}
g = pyodide.runPython(`
def g(x):
   for _ in range(1000000):
       x.f()

g
`);
g(o);
```

<details><summary>Before</summary>
<p>

image

</p>
</details> 

<details><summary>After</summary>
<p>

image

</p>
</details> 

cc @laffra

### Checklists

### DIFF
--- a/src/core/jsproxy.c
+++ b/src/core/jsproxy.c
@@ -371,11 +371,10 @@ EM_JS_VAL(JsVal, JsProxy_GetAttr_js, (JsVal jsobj, const char* ptrkey), {
 static PyObject*
 JsProxy_GetAttr(PyObject* self, PyObject* attr)
 {
-  PyObject* result = PyObject_GenericGetAttr(self, attr);
-  if (result != NULL || !PyErr_ExceptionMatches(PyExc_AttributeError)) {
+  PyObject* result = _PyObject_GenericGetAttrWithDict(self, attr, NULL, 1);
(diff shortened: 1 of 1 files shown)


ryanking13 (approved):
I don't know about that private API but the result looks good to me.

laffra:
This would have a significant impact on all FFI calls, correct?

hoodmane:
It should have a significant impact on all field accesses actually e.g., `anyJsObject.jsField`. If the result is a JsProxy, it reduces the time spent doing this by 43%. If the result is a number or boolean, it reduces the time of the access by 76%. If the result is a string, the time reduction will be somewhere in between.

hoodmane:
@ryanking13 at least `_PyObject_GenericGetAttrWithDict` still exists on the main branch. @vstinner would be nice to get a public version of this with the `suppress` option and a `status` argument so that we could tell the difference between success and failure without checking the error flag. But I'd imagine this is already on your radar?

PR #4968, opened 2024-07-28
TST Fix pygame and pytest tests

ryanking13:
Fixes tests of pytest and pygame-ce which I found issues while testing these packages out-of-tree.

- pytest: Remove numpy from the test dependency, and rewrite the test overall.
- pygame-ce: Prevent running tests if pygame-ce if not built.


hoodmane (approved):
Thanks!

hoodmane (approved):
Thanks!

ryanking13:
Thanks for the review!

PR #4969, opened 2024-07-28
Add pikepdf

SF73:
### Description

My attempt to add pikepdf to pyodide.

I already managed to build it out of the tree but, I don't know how to do it properly using meta.yaml files.

The main problem I'm facing is that pikepdf requires two environment variables :
- `QPDF_SOURCE_TREE`
- `QPDF_BUILD_LIBDIR`

and I don't know how to do it in a clean way.

(Right now I copy qpdf tree to WASM_LIBRARY_DIR/qpdf, the build is working but I cannot import it)

```py
>>> import pikepdf
  File "/lib/python3.12/site-packages/pikepdf/__init__.py", line 13, in <module>
    from pikepdf import _core
ImportError: dynamic module does not define module export function (PyInit__core)
The above exception was the direct cause of the following exception:
  File "/lib/python3.12/site-packages/pikepdf/__init__.py", line 16, in <module>
    raise ImportError(_msg) from _e
ImportError: pikepdf's extension library failed to import
```

### Checklists

- [ ] Add / update tests


packages/libqpdf/meta.yaml
```diff
@@ -0,0 +1,22 @@
...
+  type: static_library
+  script: |
+    embuilder build zlib
+    embuilder build libjpeg
+    emcmake cmake -S . -B build \
+           -DCMAKE_BUILD_TYPE=RelWithDebInfo \
+           -DUSE_IMPLICIT_CRYPTO=0 \
+           -DREQUIRE_CRYPTO_NATIVE=1
+    emmake cmake --build build --target libqpdf
+    cp -r . ${WASM_LIBRARY_DIR}/qpdf
```
ryanking13, line 20: Instead of copying it manually, I think you can do `make install` or `cmake --install` to let build system install libraries and header files properly in the desired directory. If the library follows the standard CMake configurations,

setting

```
-DCMAKE_INSTALL_PREFIX=$WASM_LIBRARY_DIR
```

on cmake configuration and calling

```
emmake make install
```

would work.

packages/libqpdf/meta.yaml
```diff
@@ -0,0 +1,22 @@
...
+  # How to do it in a cleaner way?
+  # better to use openssl instead of CRYPTO_NATIVE?
```
ryanking13, line 22: I don't know what is the CRYPTO_NATIVE is, but we support openssl so you can use it if you want.

packages/pikepdf/meta.yaml
```diff
@@ -0,0 +1,44 @@
...
+    -fexceptions
+  cxxflags: |
+    -fexceptions
+  ldflags: |
+    -fexceptions
+
+  # We need to pass the path to qpdf source as well as the path to libqpf
+  # Something like this ? (dirty)
+  script: |
+    export QPDF_SOURCE_TREE=${WASM_LIBRARY_DIR}/qpdf
```
ryanking13, line 31: This looks okay to me. Packages that use `CMake` or `scikit-build` would often locate these libraries more cleverly, but it seems like pikepdf doesn't have such ways.

ryanking13:
FYI, as you already successfully built this wheel out-of-tree, it is possible to just use the built wheel here, instead of rebuilding it. See the [duckdb](https://github.com/pyodide/pyodide/blob/73bcf49a3d4fb5a380e8aac7f59951ead5525017/packages/duckdb/meta.yaml#L7) example.

I would strongly recommend this way, as this would help us reduce the CI running time, regarding that `libpqdf` takes almost [20min](https://app.circleci.com/pipelines/github/pyodide/pyodide/7586/workflows/84d876bd-f6c8-44b4-a33a-2c871f92254b/jobs/108574?invite=true#step-104-11994_80) to build in our CI. The drawback of this approach is that you'll need to update the wheel URL manually when Pyodide updates Emscripten or Python version which may break the ABI.

SF73:
Ok, maybe I can ask directly if the maintainers of pikepdf are ok with adding a pyodide build step to their workflow.
Otherwise I can set it up correctly on my fork.
By the way, do you know why it took 20min?
[Out of tree builds](https://github.com/SF73/pikepdf/actions/workflows/build-wasm.yml) took ~3-4min for libqpdf and 1min for the wheel.

ryanking13:
> Ok, maybe I can ask directly if the maintainers of pikepdf are ok with adding a pyodide build step to their workflow.
Otherwise I can set it up correctly on my fork.

Both options are okay for us.

> By the way, do you know why it took 20min?

I think it is because we build multiple packages in parallel. So, not all of the 20 minutes was spent on libqpdf.

SF73:
On second thought, maybe it's easier to have the build step directly in Pyodide?

My main concern is that building it from the tree means different version depending on python and emscripten version combination. Also, currently emscripten version is not part of the wheel naming convention, this means we will have different releases with the same name, right?

I think it's still debatable, but in the meantime I fixed the recipe.

It works now, and the build time is consistent with what I got on my side (5-7min in codespace).

```
built libqpdf in 6m 36s 
```

I haven't tried building with openssl yet.
Also, I have a question about whether I should add a pure python package in the run requirements and meta.yaml file. 

I still need to add some tests and a line in the changelog if you're okay with having this as part of pyodide.

packages/pikepdf/meta.yaml
```diff
@@ -0,0 +1,43 @@
...
+    export QPDF_BUILD_LIBDIR=${WASM_LIBRARY_DIR}/lib
+
+requirements:
+  host:
+    - libqpdf
+  run:
+    - Pillow
+    - lxml
+    - packaging
+  #   - Deprecated # pure python package, should I add it as it can be pulled directly from pypi
```
SF73, line 35: Should I add this line as well as the generated meta file for this package?

ryanking13:
> On second thought, maybe it's easier to have the build step directly in Pyodide?

Building everything in Pyodide in-tree is easy, but we now have too many packages in Pyodide and we're already stretched to the limit in terms of CI resources. Also, in terms of managing individual packages, we want to make sure that they're managed by the people who know them best.

So, for now, I think we can accept this PR, but we plan to remove (unvendor) them from the Pyodide main repository (See: #4918) soon.

> My main concern is that building it from the tree means different version depending on python and emscripten version combination.

It's true, but I think it's quite normal thing. When a version of a programming language changes, it is natural for package maintainer to modify the package to keep up.

> Also, currently emscripten version is not part of the wheel naming convention, this means we will have different releases with the same name, right?

No, we do have ABI version in wheel name, either `emscripten_<version>` or `pyodide_<ABI_version>` is included in the wheel name. So if you build a wheel that targets different version of Pyodide, it will have different name.

> Also, I have a question about whether I should add a pure python package in the run requirements and meta.yaml file.

You should. We use that information to make a dependency graph. You also need to add a recipe for that pure python package.

ryanking13:

PR #4974, opened 2024-07-29
HttpStatusError: store status code

### LINKED ISSUE #4973: Store at least status code in HttpStatusError ?
## 🚀 Feature

Since #4846, HttpError are their own Exceptions (before it was a blank OSError). It would be nice to also store the status code in the exception, or as requests does it the full corresponding response.

### Motivation

In micropip, id like to raise_for_status deep in high level code, and if possible catch 404 on higher level, but not other error. Like not a 500 or a 403. This currently requires parsing the HttpError message.

### Pitch

```diff
class HttpStatusError(OSError):
    def __init__(self, status: int, status_text: str, url: str) -> None:
+       self.status = status
+       self.status_text = status_text
+       self.url = urls
        if 400 <= status < 500:
            super().__init__(f"{status} Client Error: {status_text} for url: {url}")
        elif 500 <= status < 600:
            super().__init__(f"{status} Server Error: {status_text} for url: {url}")
```

### Alternatives

Requests also store the full request:

```diff
class HttpStatusError(OSError):
    def __init__(self, status: int, status_text: str, url: str, *, req=None) -> None:
+       self.req = req
        if 400 <= status < 500:
            super().__init__(f"{status} Client Error: {status_text} for url: {url}")
        elif 500 <= status < 600:
            super().__init__(f"{status} Server Error: {status_text} for url: {url}")
```

Or, have a custom class:

```
class NotFoundError(HttpStatusError):
   ...
```

And a factory that returns the right exception, but that seem painful to code and use.

Carreau:
It can be convenient to access after the fact instead of passing the msg.

Closes #4973

### Checklists

- [x] Add / update tests
- ~~[ ] Add new / update outdated documentation~~

I don't think it needs docs, but a test for sure. Maybe a CHANGELOG entry. I'll do that later.


hoodmane:
Yes, please do add a test and a changelog entry.

hoodmane:
Thanks @Carreau!

Carreau:
I made one of the test more specific and added a line to the changelog.

src/py/pyodide/http.py
```diff
@@ -28,7 +28,14 @@
...
 
 class HttpStatusError(OSError):
+    status: int
+    status_text: str
+    url: str
+
     def __init__(self, status: int, status_text: str, url: str) -> None:
+        self.status = status
+        self.status_text = status_text
+        self.url = urls
```
hoodmane, line 38:
```suggestion
        self.url = url
```

hoodmane (approved):
Looks good to me once CI passes.

src/tests/test_pyodide_http.py
```diff
@@ -80,27 +80,38 @@ def raise_for_status_fixture(httpserver):
...
 
     resp = await pyfetch(raise_for_status_fixture["/status_404"])
     with pytest.raises(
-        OSError, match="404 Client Error: NOT FOUND for url: .*/status_404"
-    ):
+        HttpStatusError, match="404 Client Error: NOT FOUND for url: .*/status_404"
+    ) as error_404:
         resp.raise_for_status()
 
+    assert error_404.status == 404
```
hoodmane, line 100: Need to put `.value` here and in all of these places:
```suggestion
    assert error_404.value.status == 404
```

hoodmane:
Thanks again @Carreau!

PR #4979, opened 2024-07-31
MAINT Modularize `load-package.ts`

ryanking13:
### Description

The `load-package.ts` file has one public API `loadPackage()` and multiple private functions used internally by `loadPackage()`. I am planning to refactor these functions to make them clearer and easier to fix.

As a start, this PR modularizes functions in `load-package.ts` into a 'PackageManager' class, with a very few modifications. I also renamed the file to `package-manager.ts`.

I will do further refactorings in a follow-up PRs.


hoodmane:
Could you split this up into a PR that just renames the file and a second PR that moves stuff around in it? Currently the diff for this PR shows the entire file as changed so it is a bit difficult to review.

ryanking13:
Sure, I renamed a file. The diff is still not that pretty to review though 😅

src/js/load-package.ts
```diff
@@ -84,335 +90,437 @@ async function initializePackageIndex(lockFilePromise: Promise<Lockfile>) {
...
+    if (!pkg) {
+      pkg = {
+        name: "",
+        version: "",
+        file_name: ".whl",
         install_dir: "site",
         sha256: "",
         package_type: "package",
-        imports: [],
+        imports: [] as string[],
```
hoodmane, line 303: Is this needed? It knows the type of pkg...

hoodmane (approved):
Thanks @ryanking13!

src/js/load-package.ts
ryanking13: Right, typescript complained during the refactorings but it seems it is not needed anymore.

PR #4982, opened 2024-08-01
Update stack switching to support new JSPI API

### LINKED ISSUE #4883: The JSPI feature in Chrome has undergone changes. Are there any plans for subsequent compatibility updates to address this situation?
## 🚀 Feature

Support for stack switching using the WebAssembly.Suspender API.

### Motivation

The new API(WebAseembly.Suspending) will be live during the remainder of the origin trial — available with Chrome M126.
The previous API(WebAseembly.Suspender) will also be available during the origin trial; however, it is planned to be retired shortly after Chrome M128.

### Additional context

https://v8.dev/blog/jspi-newapi

hoodmane:
[The JSPI API changed](https://v8.dev/blog/jspi-newapi). Currently v8 implements both the old and the new version. We want to keep support for the old version for a while so we can use stack switching in node. This supports both.

I also simplified the implementation by quite a lot somewhat by accident in the process of trying to work out how to support the new API.

Resolves #4883.

We need to test against chrome >= 126 to get test coverage for the new version.

- [x] Update docker image to use Chrome 127
- [x] Add / update tests
- [ ] Fix memory leak tests


src/tests/test_stack_switching.py
```diff
@@ -627,12 +628,14 @@ def test_memory_leak(selenium, script):
         """
         """
         const t = pyodide.globals.get("test");
-        let p = [];
-        // warm up first to avoid edge problems
-        for (let i = 0; i < 200; i++) {
-            p.push(t.callPromising(1));
+        for (let i = 0; i < 1; i++) {
```
ryanking13, line 631: What's this one-time loop for? If you wanted to limit the scope, just wrapping them with `{ ... }` would work I guess,

src/core/stack_switching/esbuild.config.mjs
```diff
@@ -4,38 +4,10 @@
  */
 
 import { build } from "esbuild";
-import { readFileSync } from "node:fs";
-import loadWabt from "../../js/node_modules/wabt/index.js";
```
ryanking13, line 8: Do we still need `wabt` as a build dependency?

src/core/stack_switching/stack_switching.mjs
```diff
@@ -11,29 +11,38 @@ import {
...
+  suspenderGlobal,
+} from "./suspenders.mjs";
+export { StackState } from "./stack_state.mjs";
 export { jsWrapperTag };
 
-Module.jspiSupported = false;
-Module.validSuspender = { value: 0 };
+let canConstructWasm = true;
+try {
+  // Check that WebAssembly.Module constructor works -- if dynamic eval is
```
ryanking13, line 25: What are some cases where this doesn't work?

src/core/stack_switching/stack_switching.mjs
```diff
@@ -11,29 +11,38 @@ import {
...
-    // Check that WebAssembly.Module constructor works -- if dynamic eval is
-    // disabled it might raise.
-    // This is the smallest valid wasm module -- only the magic number and wasm
-    // version.
-    new WebAssembly.Module(new Uint8Array([0, 97, 115, 109, 1, 0, 0, 0]));
-    Module.jspiSupported = true;
-  } catch (e) {}
-}
+export const newJspiSupported = canConstructWasm && "Suspending" in WebAssembly;
+export const oldJspiSupported = canConstructWasm && "Suspender" in WebAssembly;
```
ryanking13, line 39: Let's mention the EOL (M128) of the old jspi, so we can know when to remove the old jspi.

ryanking13 (approved):
Thanks a lot for JSPI works!

src/core/stack_switching/stack_switching.mjs
hoodmane: It should be the end of life of node v22.

src/core/stack_switching/esbuild.config.mjs
hoodmane: No, I think it's only a test dependency now.

src/tests/test_stack_switching.py
hoodmane: Well this test is currently broken anyways. It looks like a possible v8 bug? It's probably going to be a lot of effort to track down...

src/core/stack_switching/stack_switching.mjs
ryanking13: I see. That's a long way off, then.

hoodmane: We can perhaps drop support sooner, say after node 24 comes out.

ryanking13: Sure, I'm okay with aggressively dropping backward compatibility if necessary.

Issue #2805, opened 2022-06-29
Install libraries following the protocol of emscripten ports.


ryanking13:
## Proposed refactoring or deprecation

We have two types of libraries that are used in the build:

1. Library ports that are officially provided by Emscripten (e.g. libjpeg, libpng, freetype, sdl)
2. Libraries that we build ourselves (e.g. libwebp, ffmpeg, openssl, libxml, sqlite3)

I think it would be great if we can install all libraries following the protocol of how emscripten installs its port. I think we can [write a port script](https://github.com/emscripten-core/emscripten/tree/main/tools/ports) and let emscripten ([embuilder](https://github.com/emscripten-core/emscripten/blob/main/embuilder.py)) deal with it.

### Motivation

If it works properly, we can take advantage of the features that emscripten provides. For example, emscripten automatically injects [cflags](https://github.com/emscripten-core/emscripten/blob/e91ea21531264f12e38a681485887212315ba08c/emcc.py#L784) and [pkg config path](https://github.com/emscripten-core/emscripten/blob/e91ea21531264f12e38a681485887212315ba08c/tools/building.py#L153-L154), so we don't have to use e.g. `WASM_LIBRARY_DIR`.

Also, I think it would be helpful for situations like #2555 in which some libraries (ssl, sqlite3) are required in both CPython and package build.

Maybe we can also have chances to upstream the library ports that we built.

rth:
> If it works properly,

What would be the workflow if  something doesn't work as expected: do we patch embuilder or there is enough flexibility to do this in the port script?

I see the appeal of this in the sense that it would allow us share packages setups with emscripten. So far though there aren't that many packages there, and it would basically make two build systems we would have to support.

So I agree that the current situation with cpython/Makefile is far from ideal. The way I saw it is that we would eventually move those to `meta.yaml` definitions, though that would also require moving cpython to be a package and I'm not quite sure how to achieve this. 

I'm not opposed to using ports more, but it would need to discuss more what would be implications of this, in terms of management of dependencies, how it would integrate into our build system, and how users would know whether use ports or meta.yaml format. 

Also ideally in the medium term, I think we should move package definitions to a separate repo, once the build system is more stable, so we can decouple package deployment from pyodide releases. There also I'm not sure what would be the implication of using ports.

ryanking13:
> > There also I'm not sure what would be the implication of using ports.
> 
> > but we would need to discuss more what would be implications of this,
> 
[...]

PR #4987, opened 2024-08-05
CI Use prebuilt packages

ryanking13:
This PR updates our CI job to use the prebuilt packages instead of building the whole packages on every commit.

Only the core packages will be built on each commit, while other packages are built and tested in `pyodide-recipes` repository. Then we run a simple import test on those packages in this repository.

This is an intermediate step for unvendoring recipes. After this PR is merged, I'll open a follow-up PR which removes non-core packages from this repository. Then the source of truth for the recipes will be changed from this repository to `pyodide-recipes`.


ryanking13:
Okay, the CI is passing. I'll need to update the prebuilt packages to sync with the main branch, other than that, I think this is good to go. I made a separate PR (#5314) for documentation update.

ryanking13:
@hoodmane I would appreciate it if you could review this PR.

Makefile
```diff
@@ -319,7 +319,6 @@ dist/pyodide-lock.json: FORCE
 	make -C packages
 	@date +"[%F %T] done building packages..."
 
-
```
hoodmane, line 322: Stray whitespace change, maybe revert this.

.circleci/config.yml
```diff
@@ -154,7 +161,7 @@ jobs:
             touch -m -d '1 Jan 2021 12:00' emsdk/emsdk/.emscripten
 
             ccache -z
-            PYODIDE_PACKAGES='<< parameters.packages >>' make dist/pyodide-lock.json
+            ENABLE_PREBUILT_PACKAGES='<< parameters.enable_prebuilt_packages >>' PYODIDE_PACKAGES='<< parameters.packages >>' make dist/pyodide-lock.json
```
hoodmane, line 164: I don't see an implementation for `ENABLE_PREBUILT_PACKAGES`...

hoodmane (commented):
The changes present look reasonable but I don't see anything to consume `enable_prebuilt_packages`. Can you explain how it's supposed to work?

Also, what will happen when we set `USE_PREBUILT_PACKAGES` to 0? Currently we need to split into several circleci steps in that case. What's the testing like?

Thanks for working on this @ryanking13!

.circleci/config.yml
ryanking13: It is used in the [packages/Makefile](https://github.com/pyodide/pyodide/blob/95dd5285866ec60007814cae7032ff31976dd4c2/packages/Makefile)

ryanking13 (commented):
Thanks for the review!

> The changes present look reasonable but I don't see anything to consume enable_prebuilt_packages. Can you explain how it's supposed to work?

It is used in the packages/Makefile. If enable_prebuilt_packages is set, it will download the packages from the URL, otherwise it will build packages in the `packages` directory as normal.

> Currently we need to split into several circleci steps in that case. 

The package build step in CI is simplified into two:

1. Building core packages

Builds packages in `packages` directory. In the packages directory, we remove all the non-necessary packages and leave "core" packages that we will build on every commit.

2. Make full distribution

Instead of building the whole package, download the prebuilt packages and make a Pyodide distribution (Pyodide core + pre-built packages)

(TODO) 3. Run basic test against the distribution

Not included in this PR, but I think we can run some basic test against the built distribution, like import test. 

> Also, what will happen when we set USE_PREBUILT_PACKAGES to 0? 

It will build packages in the packages directory, as before.

> What's the testing like?

Basically, this repository will no longer be responsible for building and testing individual packages, so I think we would not perform extensive package tests, but maybe run some basic tests.

The package tests will be performed in `pyodide/pyodide-recipes` or in separate repositories that builds packages out of tree.

.circleci/config.yml
hoodmane: Ah, right. Could you split the PR into a first one that makes the CI and Makefile changes and a second that only removes the recipes? Taking out the recipes adds a lot of noise to the diff.

hoodmane:
> It will build packages in the packages directory, as before.

But will it do this split into separate CI jobs? Or will it do it all as one step and time out?

ryanking13:
> But will it do this split into separate CI jobs? Or will it do it all as one step and time out?

It will do it in one step. I believe it can be done in 1 hour. Numpy and scipy are the only big ones.

.circleci/config.yml
ryanking13: Sure, let me try to split the PRs

hoodmane:
So numpy and scipy build in a separate job in that case? Can you explain after this change:

1. when USE_PREBUILT_PACKAGES is off, which packages are built in which CI jobs
2. when USE_PREBUILT_PACKAGES is on, which packages are built in which CI jobs

Also, it feels to me that even if `USE_PREBUILT_PACKAGES=1` is not currently failing, if we start adding a lot of more big packages to the recipes then it will start failing. I thought that one of the goals of the unvendoring was to support more packages with long build times. I assume that this is an intermediate step just to get the unvendoring done which doesn't yet solve the CI issue? What is your plan for solving it though?

ryanking13:
> So numpy and scipy build in a separate job in that case?

No, they are in the same job. 

The overall build steps are as follows:

image

1. build-core: as is, builds pyodide runtime and packages that are always built
2. build-core-packages: builds all packages in packages directory, including numpy and scipy.
3. build-full-packages: (the name is a bit misleading), do not actually build any packages, it just downloads the prebuilt packages and create a Pyodide distribution

ryanking13:
> Also, it feels to me that even if USE_PREBUILT_PACKAGES=1 is not currently failing, if we start adding a lot of more big packages to the recipes then it will start failing. I thought that one of the goals of the unvendoring was to support more packages with long build times. I assume that this is an intermediate step just to get the unvendoring done which doesn't yet solve the CI issue? What is your plan for solving it though?

Right, this PR is a short term plan described in this issue #4918. If we add more packages, pyodide/pyodide-recipes will start getting timeout as well.

So in the long term, I would like to adopt the conda-forge model.

1. Create a new organization e.g. `pyodide-recipes`.
2. In that organization, each repository will contain a single recipe (we will need to figure out how to deal with dependencies).
3. `pyodide/pyodide-recipes` will become a meta-repository: it does not actually build packages itself, but creates a lockfile using the recipes built in `pyodide-recipes` organization.

hoodmane:
So in that case, in the long term we can have a core set of recipes that we maintain and that we have CI infrastructure to build without timeouts and when we do migrations like #5320 we'll check them against this core set of recipes. All other recipes are maintained by someone else and it's their responsibility to fix the recipe after we migrate or ask for help?

ryanking13:
Yes, that's quite of my plan. Of course, if we introduce a breaking change like wasm-eh, we'll need to communicate more proactively and make it easier for package maintainers to update. In the process, the cost of communication will also increase. But I think that's a trade-off that comes with open-source projects growing and maturing.

hoodmane:
Well I think between unvendoring pyodide build and the nightly xbuildenv you've made most of the tools we need other than clear communication. I think having our abi breaks / toolchain updates happen once a year during a prespecified window will help with that. Soon we should work on the pyodide wheels on pypi PEP too...

hoodmane:
Anyways I think I approve of your plan.

ryanking13:
Thanks! In the end, I think all the problems seem to come down to the fact that Pyodide is a distribution. We need a way for packages to handle breaking changes in Pyodide before making a release whenever we make them, because the Pyodide release and the packages are a bundle...

hoodmane:
I thought we were planning to separate the package releases from the runtime releases? We could do something like have `pyodide-runtime` and `pyodide-packages` and release each of them separately then make a `pyodide` release just be a pair of a `pyodide-runtime` release and a `pyodide-packages` release.

ryanking13:
> We could do something like have pyodide-runtime and pyodide-packages and release each of them separately then make a pyodide release just be a pair of a pyodide-runtime release and a pyodide-packages release.

Yes, that's possible. We'll need to think of how we can optimize the cost of distributing the packages in that case.

ryanking13:
I'll wait for @hoodmane to finish the Emscripten updates, then rebase this PR.

hoodmane:
There's also #5320 which is getting very close.

hoodmane:
But if need be I can rebase on your change instead of you on mine.

ryanking13:
> There's also https://github.com/pyodide/pyodide/pull/5320 which is getting very close.

I see. Let's wait for that too.

hoodmane:
Thanks!

hoodmane:
Presumably this would work without deleting the packages directory? Can this pr just make the other changes and then we can have a second pr just to delete packages?

ryanking13:
> Presumably this would work without deleting the packages directory? Can this pr just make the other changes and then we can have a second pr just to delete packages?

Sounds good. I reverted all package deletions, and add `core` tag to the packages that we'll keep in this repository.

.circleci/config.yml
```diff
@@ -895,8 +793,8 @@ workflows:
       - deploy-release:
           requires:
             - check-release-version
+            - build-full-packages
             - test-core-firefox
-            - test-packages-firefox
```
hoodmane, line 918: Shouldn't we require `test-full-packages-firefox` here?

packages/atomicwrites/meta.yaml
```diff
@@ -1,6 +1,9 @@
 package:
   name: atomicwrites
   version: 1.4.1
+  tag:
+    - core
+    - pytest
```
hoodmane: What's the pytest tag mean?

hoodmane:
I think it would make sense here to split one more PR off with just the tag changes. It seems like those can land first?

packages/atomicwrites/meta.yaml
ryanking13: It is purely for informational purposes, so that I can distinguish packages in my follow up work. Packages that use the core tag (that will not be removed from this repository right now) can be categorized into following categories:

- pytest: packages that are used in pytest. We need these packages to make @run_in_pyodide work.
- cross-build: packages that are included in xbuildenv
- pyodide.test: packages that are used in our core test suite

It represents why these packages are remaining in this repository.

.circleci/config.yml
ryanking13: Right. Good point.

ryanking13:
I renamed the PR title as this PR does not remove recipes in this repository.

packages/atomicwrites/meta.yaml
ryanking13: Actually, it has another purpose for splitting the CI job.

Now the build steps are as follows

- build-core: Build pyodide runtime and packages that are used in testing core functionality (`tag:pytest, tag:pyodide.test`)
- build-core-packages: Build packages that are included in the xbuildenv (`tag:core, tag:cross-build`)
- build-all-packages: Builds all packages (or use prebuilt packages).

hoodmane:
Could you also update the description? Otherwise looks good to me.

hoodmane:
I have one remaining question: Won't we have problems with `build-packages-full` timing out when we set `USE_PREBUILT_PACKAGES` to 0?

agriyakhetarpal:
I'll also add in my review here over the weekend. Thanks!

ryanking13:
> Won't we have problems with build-packages-full timing out when we set USE_PREBUILT_PACKAGES to 0?

It will, but in the follow-up PR, I was planning to remove the packages in this repository except for the core ones, which is the original goal of this work.

ryanking13:
> I'll also add in my review here over the weekend. Thanks!

Sure, I'll wait for your review!

hoodmane:
So how do build changes work? For instance, how do we go about upgrading the compiler? We have to rebuild all the packages for that.

ryanking13:
> So how do build changes work? For instance, how do we go about upgrading the compiler? We have to rebuild all the packages for that.

My (brief) idea about that is explained in https://github.com/pyodide/pyodide/pull/5592.

And also in https://github.com/pyodide/pyodide-recipes/blob/main/docs/MAINTAINERS.md#supporting-stable-and-nightly-pyodide. I should unify the content of those two documentation though.

ryanking13:
Overall, I think it will become more like CPython's release model + conda-forge model.

We release a (pre-)alpha version with a frozen ABI. Then, the packages are updated accordingly using that ABI, and at some interval, we release a full distribution with the updated packages.

We aim to support all existing versions of packages where possible, but we don't guarantee it (we will gradually hand over the maintenance of package support to package maintainers).

hoodmane:
Makes sense, thanks for the explanation.

hoodmane (approved):

hoodmane:
I think we will need a way to test Emscripten changes against the full package set. Particularly, we wouldn't want to freeze a particular version of Emscripten on an llvm commit that has a clang regression that breaks a bunch of C++ packages. But there's no way to know about this without testing more packages when we update the compiler version.

ryanking13:
> I think we will need a way to test Emscripten changes against the full package set. Particularly, we wouldn't want to freeze a particular version of Emscripten on an llvm commit that has a clang regression that breaks a bunch of C++ packages. But there's no way to know about this without testing more packages when we update the compiler version.

That would be ideal, but we cannot test the whole ecosystem, so I think we need to make a compromise at some point.

At least we can try to keep building some core packages in this repository as an integration test. For instance, in pyodide-build, I added some [package sets](https://github.com/pyodide/pyodide-build/blob/main/integration_tests/recipes/README.md#list-of-recipes-and-their-purpose) to make sure we didn't break those packages. I think we can do some similar approach here as well.

hoodmane:
I think you can go ahead and merge this @ryanking13

agriyakhetarpal (approved):
Thanks, @ryanking13, looks good to me! Thanks for waiting for my review, and the slight delay from my end.

I had two questions, no blockers:
- longer term, would it make sense to host the packages on jsDelivr going forward as individual files, instead of a GitHub release artifact that contains all packages inside a tarball?
- I think we could also experiment with a new format for xbuildenvs. That is, call the current xbuildenv as xbuildenv-base, and also start publishing an xbuildenv-full that contains the shared libraries and the rest of the packages as well (more of a restructure as we already have such a structure in place). What do you think about this?

ryanking13:
> longer term, would it make sense to host the packages on jsDelivr going forward as individual files, instead of a GitHub release artifact that contains all packages inside a tarball?

In the really long term, my dream is not to build packages at all. All packages are built by package maintainers, and we just download the packages from PyPI.

In the mid-term, yes, probably we can host packages elsewhere. Note that we are using S3 (paid by Roman) to store packages, and use JSDelivr as CDN. I would like to avoid increasing any kind of storage cost as much as possible, but if we find a good place to host packages, yes I think we can use that.

> I think we could also experiment with a new format for xbuildenvs. That is, call the current xbuildenv as xbuildenv-base, and also start publishing an xbuildenv-full that contains the shared libraries and the rest of the packages as well (more of a restructure as we already have such a structure in place). What do you think about this?

For testing purpose, I think `pyodide venv` + `pip install` is doing the same thing. WDYT? I am also considering publishing build dependencies to some package index so they can be fetched during the package installation (https://github.com/pyodide/pyodide-build/issues/43)

agriyakhetarpal:
Thanks for the explanation!

> For testing purpose, I think `pyodide venv` + `pip install` is doing the same thing. WDYT? I am also considering publishing build dependencies to some package index so they can be fetched during the package installation ([pyodide/pyodide-build#43](https://github.com/pyodide/pyodide-build/issues/43))

Yes, however, the advantage of a full xbuildenv at build time would be to be able to rely on static/shared libraries for linking that are not included in the base xbuildenv, as libraries are only included with the packages that depend on them at installation time. Such an xbuildenv does come with the drawback of requiring users to rebuild the libraries themselves if they want to apply patches to those libraries. pyodide/pyodide-build#43 looks like a reasonable solution for the time being, but it also faces these restrictions.

ryanking13:
Yeah, dealing with static/shared libraries is quite annoying, especially in cross-compilation. I suggested something similar in (https://github.com/pyodide/pyodide/issues/3341, and https://github.com/pyodide/pyodide/issues/2805), but it was rejected.

I agree that we should provide some way for people to download and use the libraries, but I am not sure if xbuildenv is the best way to put that data. Overall, I think we need to move towards reducing the burden of managing packages and away from centralisation. If you have a good idea to share, don't hesitate to write an RFC and ask us a review.

pepijndevos:
I wonder if conda offers anything useful?

ryanking13:
> I wonder if conda offers anything useful?

We borrow some concepts of the conda recipe model, but there is no conda technology or infra used in Pyodide.

PR #4989, opened 2024-08-06
Update HttpStatusError Message if status not between 400 and 599, and update raise_for_status if error code is 600 or more.

Carreau:
Because I was creating dummy errors somewhere else with an error code of 123 and took me 30min to figure out why the repr of the errors were empty.

It also seem like raise_for_status should raise if the error code is 600 or more, the docstring should be updated as well.

### Checklists

~~- [ ] Add a [CHANGELOG](https://github.com/pyodide/pyodide/blob/main/docs/project/changelog.md) entry~~
 - [x] Add / update tests

### DIFF
--- a/src/py/pyodide/http.py
+++ b/src/py/pyodide/http.py
@@ -224,7 +224,7 @@ class FetchResponse:
 
     def raise_for_status(self) -> None:
         """Raise an :py:exc:`HttpStatusError` if the status of the response is an error (4xx or 5xx)"""
-        if 400 <= self.status <= 600:
+        if 400 <= self.status < 600:
             raise HttpStatusError(self.status, self.status_text, self.url)
 
     def clone(self) -> "FetchResponse":
--- a/src/py/pyodide/http.py
+++ b/src/py/pyodide/http.py
@@ -45,6 +45,13 @@ class HttpStatusError(OSError):
                 f"{status} Invalid error code not between 400 and 599: {status_text} for url: {url}"
             )
 
+    def __reduce__(self):
+        return (
+            self.__class__,
+            (self.status, self.status_text, self.url),
+            self.__dict__,
+        )
+
 
 class BodyUsedError(OSError):
     def __init__(self, *args: Any) -> None:
--- a/src/py/pyodide/http.py
+++ b/src/py/pyodide/http.py
@@ -40,9 +40,10 @@ class HttpStatusError(OSError):
             super().__init__(f"{status} Client Error: {status_text} for url: {url}")
         elif 500 <= status < 600:
             super().__init__(f"{status} Server Error: {status_text} for url: {url}")
-        raise ValueError(
-            f"Invalid error code not comprised between 400 and 599: {status}"
-        )
+        else:
+            super().__init__(
+                f"{status} Invalid error code not comprised between 400 and 599: {status_text} for url: {url}"
+            )
 
 
 class BodyUsedError(OSError):
--- a/src/py/pyodide/http.py
+++ b/src/py/pyodide/http.py
@@ -42,7 +42,7 @@ class HttpStatusError(OSError):
             super().__init__(f"{status} Server Error: {status_text} for url: {url}")
         else:
             super().__init__(
-                f"{status} Invalid error code not comprised between 400 and 599: {status_text} for url: {url}"
+                f"{status} Invalid error code not between 400 and 599: {status_text} for url: {url}"
             )
 
 
--- a/src/py/pyodide/http.py
+++ b/src/py/pyodide/http.py
@@ -216,8 +216,8 @@ class FetchResponse:
             raise BodyUsedError
 
     def raise_for_status(self) -> None:
-        """Raise an :py:exc:`OSError` if the status of the response is an error (4xx or 5xx)"""
-        if 400 <= self.status < 600:
+        """Raise an :py:class:`HttpStatusError` if the status of the response is an error (400 or more.)"""
+        if 400 <= self.status:
             raise HttpStatusError(self.status, self.status_text, self.url)
 
     def clone(self) -> "FetchResponse":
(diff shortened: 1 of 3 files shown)


src/py/pyodide/http.py
```diff
@@ -40,6 +40,9 @@ def __init__(self, status: int, status_text: str, url: str) -> None:
             super().__init__(f"{status} Client Error: {status_text} for url: {url}")
         elif 500 <= status < 600:
             super().__init__(f"{status} Server Error: {status_text} for url: {url}")
+        raise ValueError(
+            f"Invalid error code not comprised between 400 and 599: {status}"
+        )
```
hoodmane, line 45: I think it's a bad practice to raise an error from the constructor of an error if it can be avoided because it will overwrite the original error context. Instead, why not put a message that says that it's an unexpected status code?

Carreau: Sure. I dont' like being able to construct an invalid error, but that would works as well.

src/py/pyodide/http.py
```diff
@@ -40,6 +40,10 @@ def __init__(self, status: int, status_text: str, url: str) -> None:
             super().__init__(f"{status} Client Error: {status_text} for url: {url}")
         elif 500 <= status < 600:
             super().__init__(f"{status} Server Error: {status_text} for url: {url}")
+        else:
+            super().__init__(
+                f"{status} Invalid error code not comprised between 400 and 599: {status_text} for url: {url}"
```
hoodmane, line 45:
```suggestion
                f"{status} Invalid error code not between 400 and 599: {status_text} for url: {url}"
```

hoodmane (approved):
Looks good to me, thanks @Carreau. Please do add a test.

hoodmane (approved):
Looks good to me, thanks @Carreau. Please do add a test.

ryanking13 (changes requested):
I think you need to update this part too.

https://github.com/pyodide/pyodide/blob/461ddac819639bd1b1acbf6aedf1ccf94be19797/src/py/pyodide/http.py#L218-L221

hoodmane:
@ryanking13 isn't `raise_for_status` _supposed_ to do nothing if the status isn't an error? That way people can use it unconditionally to get rid of bad cases:
```py
request = make_request()
request.raise_for_status()
# code assuming request succeeded here
```
Or is there some other change you were thinking of?

Carreau:
I think what is meant is that it should raise for >600.

Carreau:
Plus the mention of OSError is incorrect.

Carreau:
note that raising for >600 is not what requests is doing – I assumed it should raise ? Maybe it should not. It does seem to me that > 600 are errors though ... I'm getting in a territory where I'm not comfortable, at the beginning I just wanted to make sure I could not create invalid error messages, because I was dumb in my testing. Maybe we should more carefully consider those changes ?

hoodmane:
I think I would be okay with leaving `raise_for_status` alone and just accepting the change to `HttpStatusError`.

ryanking13 (approved):
Oh, I see. Indeed [requests doesn't raise for > 600](https://github.com/psf/requests/blob/79b74ef704dc0d804937c0d015c5e9c3b02b79bd/src/requests/models.py#L997).

Then I think it would be better to follow how requests is doing: raising only for 400~599. Sorry for the confusing comment.

Carreau:
> Oh, I see. Indeed [requests doesn't raise for > 600](https://github.com/psf/requests/blob/79b74ef704dc0d804937c0d015c5e9c3b02b79bd/src/requests/models.py#L997).
> 
> Then I think it would be better to follow how requests is doing: raising only for 400~599. Sorry for the confusing comment.

No worries, I was pretty convinced requests did raise for >600, this feels extermely weird it doesn't.

src/tests/test_pyodide_http.py
```diff
@@ -112,6 +119,17 @@ async def test_pyfetch_raise_for_status_does_not_raise_200(
     assert error_504.value.status_text == "GATEWAY TIMEOUT"
     assert error_504.value.url.endswith("status_504")
 
+    resp = await pyfetch(raise_for_status_fixture["/status_900"])
+    with pytest.raises(
+        HttpStatusError,
+        match="900 Invalid error code not between 400 and 599: UNKNOWN for url: .*/status_900",
+    ) as error_900:
+        resp.raise_for_status()
```
ryanking13, line 127: This test will not raise then.

src/py/pyodide/http.py
```diff
@@ -212,8 +223,8 @@ def _raise_if_failed(self) -> None:
             raise BodyUsedError
 
     def raise_for_status(self) -> None:
-        """Raise an :py:exc:`OSError` if the status of the response is an error (4xx or 5xx)"""
-        if 400 <= self.status < 600:
+        """Raise an :py:exc:`HttpStatusError` if the status of the response is an error (4xx or 5xx)"""
+        if 400 <= self.status <= 600:
```
hoodmane, line 227: Shouldn't this be ` <` still?

Carreau:
```suggestion
        if 400 <= self.status < 600:
```

Yep

src/py/pyodide/http.py
```diff
@@ -40,6 +40,17 @@ def __init__(self, status: int, status_text: str, url: str) -> None:
...
+            super().__init__(
+                f"{status} Invalid error code not between 400 and 599: {status_text} for url: {url}"
+            )
+
+    def __reduce__(self):
+        return (
+            self.__class__,
+            (self.status, self.status_text, self.url),
+            self.__dict__,
+        )
```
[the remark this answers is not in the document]
Carreau: But without this the test added in test_pyodide_http.py that raises HttpError fails...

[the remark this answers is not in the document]
hoodmane: Only when the test was raising an uncaught HttpError, so would fail for other reasons. When the test is correct, it wouldn't happen.

But how about let's just add a separate test to this PR that does `pickle.load(pickle.dumps(HttpError(...)))`. It can even run in the host Python (no need for `@run_in_pyodide`).

hoodmane:
Thanks @Carreau!

Carreau:
Thanks !

PR #4993, opened 2024-08-07
Fix `pyimport` for some corner cases

### LINKED ISSUE #4992: `pyimport` fail to import a module from a module
## 🐛 Bug

After manually deleting the child module from `sys.modules` or deleting the attribute from the parent module, `pyimport` fail to import the child module, which is inconsistent with the behavior of `from ... import ...` statement.

### To Reproduce

```py
# setup files
from pathlib import Path
Path("a").mkdir()
Path("a/b.py").write_text("c = 1")

# import the first time and delete it
import a.b
del a.b

# from-import works fine:
from a import b
print(b.c)

# pyimport raises
from pyodide_js import pyimport
pyimport("a.b")

# but importing `a.b.c` is fine
pyimport("a.b.c")
```

### Expected behavior

```
  File "/lib/python312.zip/_pyodide/_base.py", line 655, in pyimport_impl
    res = getattr(res, fromlist[0])
          ^^^^^^^^^^^^^^^^^^^^^^^^^
AttributeError: module 'a' has no attribute 'b'
```

### Environment

- Pyodide Version: 0.26.2
- Browser version: Edge

### Additional context

Probably this is because of some caching mechanism of python.

Question: why don't we just implement `pyimport` by simply compile the path into `from-import` statement and exec it?

CNSeniorious000:
### Description

Fix #4992

### Checklists

- [x] Add / update tests

### DIFF
--- a/src/py/_pyodide/_base.py
+++ b/src/py/_pyodide/_base.py
@@ -11,6 +11,7 @@ import linecache
 import tokenize
 from collections.abc import Generator
 from copy import deepcopy
+from importlib import import_module
 from io import StringIO
 from textwrap import dedent
 from types import CodeType
@@ -655,5 +656,5 @@ def pyimport_impl(path: str) -> Any:
         try:
             res = getattr(res, fromlist[0])
         except AttributeError:
-            res = __import__(path)
+            res = import_module(path)
     return res
--- a/src/py/_pyodide/_base.py
+++ b/src/py/_pyodide/_base.py
@@ -11,6 +11,7 @@ import linecache
 import tokenize
 from collections.abc import Generator
 from copy import deepcopy
+from importlib import import_module
 from io import StringIO
 from textwrap import dedent
 from types import CodeType
@@ -649,8 +650,11 @@ def find_imports(source: str) -> list[str]:
 
 
 def pyimport_impl(path: str) -> Any:
-    [stem, *fromlist] = path.rsplit(".", 1)
-    res = __import__(stem, fromlist=fromlist)
-    if fromlist:
-        res = getattr(res, fromlist[0])
-    return res
+    if "." not in path:
+        return import_module(path)
+    stem, name = path.rsplit(".", 1)
+    mod = import_module(stem)
+    try:
+        return getattr(mod, name)
+    except AttributeError:
+        return import_module(path)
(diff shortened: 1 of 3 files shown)


CNSeniorious000:
How do you think of this? Or just replacing it with

```js
pyimport(path: string) {
  return pyodide.runPython(
    path.includes(".")
      ? `import ${path}; ${path}`
      : path.replace(/(.*)\.(.*)/, 'from $1 import $2; $2')
  )
}
```

may be another approach.

src/py/_pyodide/_base.py
```diff
@@ -649,8 +650,11 @@ def find_imports(source: str) -> list[str]:
...
-        res = getattr(res, fromlist[0])
-    return res
+    if "." not in path:
+        return import_module(path)
+    stem, name = path.rsplit(".", 1)
+    mod = import_module(stem)
+    try:
+        return getattr(mod, name)
+    except AttributeError:
+        return import_module(path)
```
hoodmane, line 660: How about this:
```suggestion
    # If multiple components, behave like `from a.b.c import last_part`
    # This compiles into 
    # IMPORT_NAME a.b.c
    # IMPORT_FROM last_part
    # for only one component, we want to behave like IMPORT_NAME
    [stem, *fromlist] = path.rsplit(".", 1)
    res = __import__(stem, fromlist=fromlist)
    if not fromlist:
       # Only one component, no IMPORT_FROM step so we can return now
       return res
    # Now imitate IMPORT_FROM
    try:
        return getattr(mod, name)
    except AttributeError:
        return import_module(path)
```

hoodmane:
Also, could you add the example from #4992 as a test?

src/py/_pyodide/_base.py
CNSeniorious000: In fact using `__import__` or `import_module` is just preference, I don't see differences in practices.
But `__import__` have some inconsistency (the same to `import_module`):

`__import__("a.b")` != `__import__("a", fromlist=["b"]).b`

So if you want minimal changes, this way would work:

```diff
    try:
        return getattr(mod, fromlist[0])
    except AttributeError:
+       return __import__(path)
```

I overthought it before.

hoodmane (approved):
Thanks!

src/py/_pyodide/_base.py
CNSeniorious000: Well I am wrong. Your code is just right. I know so little about `__import__`

ryanking13 (approved):
Thanks @CNSeniorious000!

PR #4996, opened 2024-08-08
Remove usage of `shared_library` key in pyodide lock file

ryanking13:
### Description

This key was deprecated and not used anymore, the final usage was removed in #4876. This PR removes it.

Node that even after removing this, we can retrieve the same value by checking `package_type` key in the lock file.


hoodmane (approved):
Thanks!

ryanking13:
Thanks for the review!

PR #5000, opened 2024-08-10
Add CHANGELOG entry for #4719, skip PROPACK test for `scipy.sparse.linalg._eigen`

agriyakhetarpal:
### Description

Missed adding this in #4719, adding it here. This fixes a timeout with the CircleCI checks coming from a failing SciPy test (which is skipped in GHA).

### Checklists

N/A

### DIFF
--- a/docs/project/changelog.md
+++ b/docs/project/changelog.md
@@ -53,6 +53,7 @@ myst:
 - Upgraded `pandas` to 2.2.2 {pr}`4893`
 - Upgraded `zengl` to 2.5.0 {pr}`4894`
 - Upgraded `sourmash` to 4.8.11 {pr}`4980`
+- Upgraded `scipy` to 1.13.0 {pr}`4719`
 - Added `casadi` 3.6.5 {pr}`4936`
 - Added `rasterio` 1.13.10, `affine` 2.4.0 {pr}`4983`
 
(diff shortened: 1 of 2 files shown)


ryanking13 (approved):
Thanks @agriyakhetarpal! Could you also update the test to skip the failing test?

> I see that the test_SVDS_PROPACK test that we skip for GHA test-scipy job is running on CircleCI, which causes the function signature mismatch. It causes other tests to time out.

agriyakhetarpal:
Yes, I was thinking of doing it here – I didn't find anything as to why CircleCI runs it when GHA skips it (at least, related to my PR). I'll take a look further.

agriyakhetarpal:
I guess when running the tests on the `scipy` trigger, the GHA job runs `mv scipy-conftest.py conftest.py` to rename it and hence, but the CircleCI jobs don't, so `--runtime=firefox-no-host packages/*/test*.py --skip-passed` won't pick up the configuration from that file. That doesn't explain how it worked before my PR, though, so my hunch is that checking for the commit message for `[scipy]` and renaming `scipy-conftest.py` should also be pursued in the CircleCI steps/jobs.

agriyakhetarpal:
Found it, it's failing because of this line:

https://github.com/pyodide/pyodide/blob/71a211860c1ac22ba7aeb02315fea377b279b18f/packages/scipy/test_scipy.py#L71

agriyakhetarpal:
Some of the tests might have been skipped because of the `pytest` caching, but I think this should be safe to merge now – I verified locally in an IPython shell (in SciPy, this is `python dev.py ipython`) that this runs the ARPACK and LOBPCG tests, and the PROPACK ones are skipped.

PR #5006, opened 2024-08-13
BLD Split src/js/esbuild.config.mjs into multiple files

hoodmane:
This separates the build step for _pyodide.out.js and pyodide.js. They happen at different steps so it makes more sense to keep them separate.


ryanking13 (approved):
Thanks @hoodmane. In the future, it would be nice if we can clearly distinguish what goes to pyodide.js and what goes to pyodide.asm.js. For now, it is quite entangled and it is hard to understand how js files are processed during the build process.

hoodmane:
Yeah I think we should split the js directory into pyodide.asm.js only / pyodide.js only / shared subdirectories.

PR #5007, opened 2024-08-13
 Automatic detection using podman

mokeyish:
### Description
If docker does not exist, podman is used instead.

### DIFF
--- a/run_docker
+++ b/run_docker
@@ -19,6 +19,14 @@ HEALTHCHECK_MESSAGE="Done initialization"
 
 set -eo pipefail
 
+# If docker does not exist, podman is used instead.
+if ! command -v docker &> /dev/null; then
+    USER_FLAG=()
+    docker() {
+      podman $@
+    }
+fi
+
 
 function usage() {
   cat > /dev/stdout <<EOF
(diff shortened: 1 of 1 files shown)


agriyakhetarpal:
The `awkward-cpp` build failure looks like it should be fixed by https://github.com/pybind/pybind11/pull/5301

ryanking13 (approved):
I never tried podman before TBH, but it looks okay to me.

agriyakhetarpal (commented):
This looks good to me as well, but if we are to provide this, could we make `podman` an alternative CLI option instead of a fallback? It won't add any extra complexity, and something like `./run_docker --engine docker,podman` would be great (with the default being Docker).

mokeyish:
> This looks good to me as well, but if we are to provide this, could we make `podman` an alternative CLI option instead of a fallback? It won't add any extra complexity, and something like `./run_docker --engine docker,podman` would be great (with the default being Docker).

Currently, docker is used by default, and podman will be tried only if docker does not exist. My personal preference is podman, because docker has permission vulnerabilities and is unsafe, so I basically won't install it.


I think automatic detection is good, adding an option to select manually is a bit troublesome.

agriyakhetarpal (approved):
That sounds good to me; thanks for the explanation, @mokeyish! Yes, I'm probably one of those who has both tools installed, but I guess there won't be many others. We can provide the choice of runtime if someone requests it.

PR #5009, opened 2024-08-14
Check that snapshot was created for the current Pyodide build

hoodmane:
Trying to load snapshots from other builds will crash.

- [x] Add / update tests


src/js/pyodide.ts
```diff
@@ -25,6 +25,8 @@ declare function _createPyodideModule(
   settings: EmscriptenSettings,
 ): Promise<Module>;
 
+declare const BUILD_ID: string;
```
ryanking13, line 28: Let's add some comment for what this variable is and where it is used.

ryanking13 (approved):
Thanks!

PR #5011, opened 2024-08-15
SciPy v1.13.1

agriyakhetarpal:
### Description

This PR updates SciPy to version `1.13.1`. Tagging recipe maintainers: @lesteve, @steppi; and @rgommers, for visibility. SciPy was previously updated in #4719 to version `1.13.0`.

### Checklists

- [x] Add a [CHANGELOG](https://github.com/pyodide/pyodide/blob/main/docs/project/changelog.md) entry

Issue #4160, opened 2023-09-21
pyodide-build: unable to locate numpy package while rebuilding the pandas 2.0.0 as pyodide


kishorekaruppusamy:
## 🐛 Bug

I am trying to build a `pyodide package` using the `pandas 2.0.0 tar.gz` version. pyodide build creates a isolated python environment and searching `numpy` .

### To Reproduce

- Build and Create a docker container using this [Image](https://github.com/docker-library/python/blob/81b6e5f0643965618d633cd6b811bf0879dee360/3.10/slim-bookworm/Dockerfile).
- Once created login to the container and follow [out of tree](https://pyodide.org/en/stable/development/building-and-testing-packages.html#building-and-testing-python-packages-out-of-tree) package building steps.
- Once the [Emscripten setup](https://pyodide.org/en/stable/development/building-and-testing-packages.html#set-up-emscripten) is completed. download the `pandas-2.0.0 tar.gz distribution` from [pypi](https://files.pythonhosted.org/packages/9f/12/0b6bdd627b99cb10816956c1047b0733ef33b61a84e3420faf4d3202df06/pandas-2.0.0.tar.gz) and extract that in some folder.
- Change directory into the package folder where the `setup.py or pyproject.toml file` is located and run pyodide build
### Expected behavior.

- Rebuild the `pandas 2.0.0` version as `pyodide package`.

### Error Log
```Downloading xbuild environment                                                                                                                                                                 
Installing xbuild environment                                                                                                                                                                  
* Creating virtualenv isolated environment...
* Installing packages in isolated environment... (Cython>=0.29.33,<3, setuptools>=61.0.0, versioneer[toml], wheel)
* Getting dependencies for wheel...
Traceback (most recent call last):
  File "/usr/local/lib/python3.10/site-packages/pep517/in_process/_in_process.py", line 351, in <module>
    main()
  File "/usr/local/lib/python3.10/site-packages/pep517/in_process/_in_process.py", line 333, in main
    json_out['return_val'] = hook(**hook_input['kwargs'])
  File "/usr/local/lib/python3.10/site-packages/pep517/in_process/_in_process.py", line 118, in get_requires_for_build_wheel
    return hook(config_settings)
  File "/tmp/build-env-f86aqev9/lib/python3.10/site-packages/setuptools/build_meta.py", line 355, in get_requires_for_build_wheel
    return self._get_build_requires(config_settings, requirements=['wheel'])
[...]

PR #5012, opened 2024-08-16
SciPy v1.14.0

agriyakhetarpal:
### Description

This PR updates SciPy to version 1.14.0. Tagging recipe maintainers: @lesteve, @steppi; and @rgommers, for visibility. SciPy was previously updated in https://github.com/pyodide/pyodide/pull/5011 to version 1.13.1.

### Checklists

- [x] Add / update tests


agriyakhetarpal:
`../scipy/meson.build:147:7: ERROR: Program 'f2py' not found or not executable`

I initially thought that there's now a hard requirement for NumPy v2 with SciPy v1.14.0, but that doesn't seem to be the case. I might need to provide hints to Meson/`meson-python` on where to find it (i.e., in NumPy's directories or in the site-packages directory).

rgommers:
You can build SciPy 1.14 against NumPy 1.26 just fine - only need to remember to then rebuild SciPy when NumPy 2.0 lands.

rgommers:
If `f2py` is going missing, the entrypoint may not be installed? It's defined in NumPy's `pyproject.toml`

agriyakhetarpal:
Yes. I realise that the error is because while NumPy is indeed built beforehand, it isn't installed into the environment, and therefore the `f2py` entry point does not exist in `PATH`. I'm not sure if there's a clean way of having NumPy installed rather than built when building SciPy, so we will have to work around this somehow.

agriyakhetarpal:
Okay, managed to fix this – I have had to revert all the uses of the f2py generator calls I could find, which were from your commit I mentioned above, @rgommers; it isn't an ideal solution since it conflicts with other patches and the diff is quite big (in terms of both tidiness and upstream-ability – or rather, the lack of them). Although, my hunch is that we don't need to do this at all, though, and simply restoring the `sys.executable -m` prefix before invoking `numpy.f2py` should fix it.


We are now able to start compiling SciPy v1.14.0. However, it fails with the following logs (truncated here):

<details>
<summary>Tap to view logs</summary>

```
Program _generate_pyx.py found: YES (/tmp/build-env-vb6iw_yf/bin/python /src/packages/scipy/build/scipy-1.14.0/scipy/linalg/_generate_pyx.py)                                
Program ../_build_utils/echo.py found: YES (/tmp/build-env-vb6iw_yf/bin/python /src/packages/scipy/build/scipy-1.14.0/scipy/linalg/../_build_utils/echo.py)                  
Program ../_generate_sparsetools.py found: YES (/tmp/build-env-vb6iw_yf/bin/python                                                                                           
/src/packages/scipy/build/scipy-1.14.0/scipy/sparse/sparsetools/../_generate_sparsetools.py)                                                                                 
Checking for size of "void*" : 4                                                                                                                                             
Compiler for Fortran supports arguments -w: YES                                                                                                                              
Build targets in project: 197                                                                                                                                                
                                                                                                                                                                             
scipy 1.14.0                                                                                                                                                                 
                                                                                                                                                                             
  User defined options                                                                                                                                                       
    Cross files : /usr/local/lib/python3.12/site-packages/pyodide_build/tools/emscripten.meson.cross                                                                         
    Native files: /src/packages/scipy/build/scipy-1.14.0/build/meson-python-native-file.ini                                                                                  
    buildtype   : release                                                                                                                                                    
    b_ndebug    : if-release                                                                                                                                                 
    b_vscrt     : md                                                                                                                                                         
                                                                                                                                                                             
Found ninja-1.11.1 at /usr/bin/ninja                                                                                                                                         
                                                                                                                                                                             
ERROR: ld.wasm does not support shared libraries.                                                                                                                            
                                                                                                                                                                             
A full log can be found at /src/packages/scipy/build/scipy-1.14.0/build/meson-logs/meson-log.txt                                                                             
                                                                                                                                                                             
ERROR Backend subprocess exited when trying to invoke build_wheel                                                                                                            
[2024-08-16 18:03:21] Failed building package scipy in 34.0 seconds.                                                                                                         
                                                                                                                                                                             
                                                                                                                                                                             
ERROR: cancelling buildall due to error building scipy
```

</details>

It suggests we are compiling a shared library somewhere, but I don't see anything in the log file; it's the same as the above trace. I'm now going to try to find out where a shared library has been defined in the target definitions (and where that changed between v1.13.0 and v1.14.0). I'm also going to try passing additional arguments to the `meson setup` and `meson install` commands to see if they can produce step-by-step output or bump the verbosity.

lucascolley:
@agriyakhetarpal perhaps https://github.com/scipy/scipy/pull/20321?

agriyakhetarpal:
Thank you, @lucascolley – will take a look!

agriyakhetarpal:
Never mind, patch 15 doesn't seem to work, based on the error raised in CI – I think we'll have to remove the f2py generators entirely unless I'm missing any other ways to fix this. I think I had an `f2py` executable installed in `src/.docker_home/local/bin`, which led me to believe that it did work like we wanted it to. @lucascolley, I've had a look at scipy/scipy#20321 and I feel that the functionality might not be worth fixing too much for Pyodide – we could either let that be a `static_library`, or remove the extension module altogether, what do you think?

Also, I now have an issue locally: the build procedure is trying to compile `test_fortran.pyf` despite the fact that I've disabled the compilation of test modules using install tags (i.e., `install-args=--tags=runtime,python-runtime,devel`). I don't see why that should happen?

<details>
<summary>Tap to view logs</summary>

```
730/1344] Compiling Fortran object scipy/io/_test_fortran.cpython-312-wasm32-emscripten.so.p/_test_fortran.f.o                                                              
FAILED: scipy/io/_test_fortran.cpython-312-wasm32-emscripten.so.p/_test_fortran.f.o                                                                                          
/tmp/tmpve9euhz9/gfortran -Iscipy/io/_test_fortran.cpython-312-wasm32-emscripten.so.p -Iscipy/io -I../scipy/io                                                               
-I../../../../../../tmp/build-env-l4q58zjw/lib/python3.12/site-packages/numpy/core/include                                                                                   
-I../../../../../../tmp/build-env-l4q58zjw/lib/python3.12/site-packages/numpy/f2py/src -Iscipy/lib_fortranobject.a.p -Iscipy/libdummy_g77_abi_wrappers.a.p                   
-I/src/packages/.libs/include -I/src/cpython/installs/python-3.12.1/include/python3.12 -I/usr/local/include/python3.12 -fvisibility=hidden -fdiagnostics-color=always        
-DNDEBUG -D_FILE_OFFSET_BITS=64 -Wall -std=legacy -O3 -Wno-conversion -fPIC -Wno-argument-mismatch -Wno-conversion -Wno-maybe-uninitialized -Wno-unused-dummy-argument       
-Wno-unused-label -Wno-unused-variable -Wno-tabs -Jscipy/io/_test_fortran.cpython-312-wasm32-emscripten.so.p -o                                                              
scipy/io/_test_fortran.cpython-312-wasm32-emscripten.so.p/_test_fortran.f.o -c ../scipy/io/_test_fortran.f                                                                   
   read_unformatted_double:                                                                                                                                                  
Error on line 3: syntax error                                                                                                                                                
Error on line 4: syntax error                                                                                                                                                
Error on line 5: syntax error                                                                                                                                                
Error processing entries before line 6: Declaration error for m: attempt to use undefined variable                                                                           
Error processing entries before line 6: Declaration error for n: attempt to use undefined variable                                                                           
Error processing entries before line 6: Declaration error for k: attempt to use undefined variable                                                                           
Error processing entries before line 6: Declaration error for a: attempt to use undefined variable                                                                           
Error processing entries before line 6: Declaration error for filename: attempt to use undefined variable                                                                    
Error on line 6: Declaration error for trim: attempt to use untyped function                                                                                                 
Error on line 6: bad file in open                                                                                                                                            
Error on line 6: non-character control clause                                                                                                                                
Warning on line 9: ignoring text after "end".                                                                                                                                
   read_unformatted_int:                                                                                                                                                     
Error on line 13: syntax error                                                                                                                                               
Error on line 14: syntax error                                                                                                                                               
Error on line 15: syntax error                                                                                                                                               
Error processing entries before line 16: Declaration error for m: attempt to use undefined variable                                                                          
Error processing entries before line 16: Declaration error for n: attempt to use undefined variable                                                                          
Error processing entries before line 16: Declaration error for k: attempt to use undefined variable                                                                          
Error processing entries before line 16: Declaration error for a: attempt to use undefined variable                                                                          
Error processing entries before line 16: Declaration error for filename: attempt to use undefined variable                                                                   
Error on line 16: Declaration error for trim: attempt to use untyped function                                                                                                
Error on line 16: bad file in open                                                                                                                                           
Error on line 16: non-character control clause                                                                                                                               
Warning on line 19: ignoring text after "end".                                                                                                                               
   read_unformatted_mixed:                                                                                                                                                   
Error on line 23: syntax error                                                                                                                                               
Error on line 24: syntax error                                                                                                                                               
Error on line 25: syntax error                                                                                                                                               
Error on line 26: syntax error                                                                                                                                               
Error processing entries before line 27: Declaration error for m: attempt to use undefined variable                                                                          
Error processing entries before line 27: Declaration error for n: attempt to use undefined variable                                                                          
Error processing entries before line 27: Declaration error for k: attempt to use undefined variable                                                                          
Error processing entries before line 27: Declaration error for a: attempt to use undefined variable                                                                          
Error processing entries before line 27: Declaration error for b: attempt to use undefined variable                                                                          
Error processing entries before line 27: Declaration error for filename: attempt to use undefined variable                                                                   
Error on line 27: Declaration error for trim: attempt to use untyped function                                                                                                
Error on line 27: bad file in open                                                                                                                                           
Error on line 27: non-character control clause                                                                                                                               
Warning on line 30: ignoring text after "end".                                                                                                                               
  File "/tmp/tmpve9euhz9/gfortran", line 632, in <module>                                                                                                                    
    compiler_main()                                                                                                                                                          
  File "/tmp/tmpve9euhz9/gfortran", line 628, in compiler_main                                                                                                               
    sys.exit(handle_command(args, build_args))                                                                                                                               
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^                                                                                                                                
  File "/tmp/tmpve9euhz9/gfortran", line 596, in handle_command                                                                                                              
    tmp = replay_f2c(line)                                                                                                                                                   
          ^^^^^^^^^^^^^^^^                                                                                                                                                   
  File "/tmp/tmpve9euhz9/_f2c_fixes.py", line 184, in replay_f2c                                                                                                             
    subprocess.check_call(                                                                                                                                                   
  File "/usr/local/lib/python3.12/subprocess.py", line 413, in check_call                                                                                                    
    raise CalledProcessError(retcode, cmd)                                                                                                                                   
subprocess.CalledProcessError: Command '['/src/packages/scipy/build/scipy-1.14.0/f2c/src/f2c', '-R']' returned non-zero exit status 1.
```

</details>

lucascolley:
>  I feel that the functionality might not be worth fixing too much for Pyodide – we could either let that be a static_library, or remove the extension module altogether, what do you think?

I'm not sure, probably best to defer to @steppi on that one

agriyakhetarpal:
I'm not facing the error in CI, i.e., `ModuleNotFoundError: No module named 'scipy.io.tests'`, when testing locally – I will continue to debug.

agriyakhetarpal:
All top-level module imports pass locally except for `scipy.stats` and `scipy.signal`; both fail due to the absence of `scipy.stats._page_trend_test`. This is a Python file, so it doesn't indicate a compilation failure – just that it wasn't copied into the wheel for some reason. I've just pushed another wheel build/test, this time with the install tags temporarily disabled so that we can allow the test suite to run fully – let's see how that goes.

agriyakhetarpal:
I have the test suite running up to 34% locally (and in CI too, at the time of writing), at which point the test 

`optimize/tests/test__shgo.py::TestShgoSobolTestFunctions::test_f3_sobol`

fails with an unresolved symbol `d_sign` – which also comes from OpenBLAS (see #4876). @ryanking13, could you teach me how you ascertain which symbols are exported by which module? Do I just need to run `wasm-objdump` over the relevant shared objects until I find the offending one?

agriyakhetarpal:
Progress! I have the test suite running up to 34% locally (and in CI too, at the time of writing), at which point the test 

`optimize/tests/test__shgo.py::TestShgoSobolTestFunctions::test_f3_sobol`

fails with an unresolved symbol `d_sign` – which also comes from OpenBLAS (see #4876). @ryanking13, could you teach me how you ascertain which symbols are exported by which module? Do I just need to run `wasm-objdump` over the relevant shared objects until I find the offending one?

hoodmane:
Another trick I use is `rg -uuu <symbol_name>` in the build directory which can tell you every time the symbol is mentioned in the build directory. If you want specifically object files, .so files, or archives you can pass for example `-g'*.so'` as well. ripgrep is very helpful in general. wasm-objdump and llvm-nm are also useful.

agriyakhetarpal:
Thanks, @hoodmane! In the time you commented, I did try something on similar lines by going through the 110 shared objects produced, recursively, in the unpacked wheel:

```bash
find scipy-1.14.0-cp312-cp312-pyodide_2024_0_wasm32/scipy/ -name "*.so" -type f -print0 | xargs -0 -I {} wasm-objdump {} --details
```

Took me a short while to conjure that up, but it seems to work well, and it gives me ten shared objects to look at. I tried `ripgrep`, through the command:

```bash
rg -uuu 'd_sign' scipy-1.14.0-cp312-cp312-pyodide_2024_0_wasm32/scipy/ --no-ignore --hidden --glob '*.so' 
```

which returns the same ten files, so we're good on that front.

The relevant extension modules are: `__odrpack`, `_slsqp`, `_odepack`, `_lsoda`, `_dop`, `_dfitpack`, `_arpack`, `_dpropack`, `_vode`, and `_zpropack`. Out of these, we currently patch `_dop`, and we patch the trio of `_cobyla`, `_minpack2`, and `_mvn` for other reasons/OpenBLAS symbols. I think we need to patch one or more from this list, and I'll try to find out which one(s) that would be.

packages/scipy/meta.yaml
```diff
@@ -17,25 +17,27 @@ package:
...
+    - patches/0008-Mark-mvndst-functions-recursive.patch
+    - patches/0009-Make-sreorth-recursive.patch
+    - patches/0010-Link-openblas-with-modules-that-require-f2c.patch
+    - patches/0011-Remove-fpchec-inline-if-else-constructs.patch
+    - patches/0012-Remove-chla_transtype.patch
+    - patches/0013-Set-wrapper-return-type-to-int.patch
+    - patches/0014-Skip-svd_gesdd-test.patch
+    - patches/0015-Remove-f2py-generator.patch
+    - patches/0016-Make-sf_error_state_lib-a-static-library.patch
+    - patches/0017-Remove-test-modules-that-fails-to-build.patch
```
agriyakhetarpal, line 40: Out of these, the only troublesome patch is the fifteenth one, see comments below. cc: @rgommers

packages/scipy/meta.yaml
```diff
@@ -50,8 +52,14 @@ build:
     -L$(NUMPY_LIB)/core/lib/
     -L$(NUMPY_LIB)/random/lib/
     -fexceptions
+
+  # Exclude tests via Meson's install tags functionality.
+  unvendor-tests: true
+  # install-args=--tags=runtime,python-runtime,devel
+  # Disable when running tests, enable when a PR is ready, i.e., building for distribution.
```
agriyakhetarpal, line 59: I haven't been able to get the[ install tags functionality](https://docs.scipy.org/doc/scipy/building/redistributable_binaries.html) working properly. Would you have any suggestions here, @rgommers? I would like that someone updating the package recipe can have a means of disabling the functionality momentarily to be able to trigger the tests, i.e., both the ones running on CircleCI and the ones running on GHA through the `[scipy]` commit marker established in #4935, and once the recipe is ready and passes all tests, enable it back so that the test-tagged modules can be not built when merging the recipe and we get a smaller wheel (and I can document that process better in a comment above this field).

However, it doesn't seem to work, and it still tries to build the modules that have been removed in patch `0017` (and fails in doing so). Could this be a bug in meson-python or am I using this functionality incorrectly? I can push another commit if needed with the install tags enabled to provide a failing job.

packages/scipy/meta.yaml
```diff
@@ -103,10 +111,6 @@ build:
     # Input error causes "duplicate symbol" linker errors. Empty out the file.
     echo "" > scipy/sparse/linalg/_dsolve/SuperLU/SRC/input_error.c
 
-  _retain-test-patterns:
-    - "*_page_trend_test.py"
-    - "*bws_test.py"
-
```
agriyakhetarpal, line 109: This was removed when testing the installation tags functionality (and I now see why these are required). I can bring this back once the conversations above have a resolution.

packages/scipy/patches/0010-Link-openblas-with-modules-that-require-f2c.patch
```diff
@@ -1,24 +1,26 @@
...
+Co-Developed-by: Agriya Khetarpal <74401230+agriyakhetarpal@users.noreply.github.com>
 ---
- scipy/integrate/meson.build | 2 +-
- scipy/optimize/meson.build  | 6 +++---
- scipy/stats/meson.build     | 2 +-
- 3 files changed, 5 insertions(+), 5 deletions(-)
+ scipy/integrate/meson.build   | 2 +-
+ scipy/optimize/meson.build    | 6 +++---
+ scipy/stats/meson.build       | 2 +-
+ 4 files changed, 5 insertions(+), 5 deletions(-)
```
agriyakhetarpal, line 16:
```suggestion
 scipy/integrate/meson.build | 2 +-
 scipy/optimize/meson.build  | 6 +++---
 scipy/stats/meson.build     | 2 +-
 3 files changed, 5 insertions(+), 5 deletions(-)
```

packages/scipy/patches/0015-Remove-f2py-generator.patch
agriyakhetarpal, line 1: This patch is a little problematic because I have had to remove the f2py generator(s) entirely and use the older method for specifiying targets, because `f2py` is required as an executable and it doesn't seem to be installed when NumPy's WASM wheel is built, because (I think) the Pyodide build procedure doesn't install NumPy after it has been built and when it is building SciPy. We could override this by adding `pip install numpy --force-reinstall` to install the Linux-amd64 NumPy wheel to access the executable from there, but I feel that is too messy. Given the nature of the patch, it has to exist either before or after the rest of the patches because it modifies and brings back a lot of the lines that collide with other patches.

agriyakhetarpal:
Although I can't request a review from you from the Reviewers tab, of course, I would appreciate it if you could also take a look at these changes, @rgommers – thanks!

packages/scipy/meta.yaml
agriyakhetarpal: Maybe it's related to https://github.com/scipy/scipy/pull/21225, but I don't see why it wouldn't respect `install_tag: tests` either way.

agriyakhetarpal:
The test failure for Firefox is unrelated – I increased the PyArrow timeout to six times the original value in 828a08f9c430aa4256ea90b9d1cda061fe966556. I'm assuming that a value of `120` means 120 seconds and not milliseconds. Probably overkill, but it makes the timeout statistically unlikely 🤷

ryanking13:
Cool, thanks @agriyakhetarpal!

> could you teach me how you ascertain which symbols are exported by which module (and its dependents)?

I think you've already figured it out, but FWIW, I use `auditwheel-emscripten` to list all symbols in the scipy wheel and search which .so files contains the problematic symbol.

```
pip install auditwheel-emscripten
pyodide auditwheel exports <scipy-imports>
```

ryanking13:
> because f2py is required as an executable and it doesn't seem to be present on PATH when NumPy's WASM wheel is built, because (I think) the Pyodide build procedure doesn't install NumPy after it has been built and when it is building SciPy.

I think we need to fix this somehow... we ignore numpy even if it is specified as a build requirement in the pyproject.toml ([code pointer](https://github.com/pyodide/pyodide-build/blob/main/pyodide_build/pypabuild.py#L126-L132), but it has been a problem for a few people (https://github.com/pyodide/pyodide/issues/4160).

ryanking13 (approved):

agriyakhetarpal:
> > If there are better alternatives to this, I'd be keen on exploring those.
> 
> Let me see if I can fix it in pyodide-build side.

Sure, thanks! I saw this comment a little late and was working on my script-based idea in the meantime, but ultimately anything is better than a script that we need to keep maintaining.

ilayn:
> Maybe I'll go into more depth the next time something gets rewritten in C/C++ in SciPy upstream 

It's L-BFGS-B and ARPACK on its way. Should it be SLSQP next? Otherwise will be PROPACK.

steppi:
> > I feel that the functionality might not be worth fixing too much for Pyodide – we could either let that be a static_library, or remove the extension module altogether, what do you think?
> 
> I'm not sure, probably best to defer to @steppi on that one

Sorry, I missed the ping while traveling. If it's too much trouble to get the shared library working, I'd prefer removing the special error handling functionality altogether rather than shipping a broken version that only works for half of the functions.

agriyakhetarpal:
> It's L-BFGS-B and ARPACK on its way. Should it be SLSQP next? Otherwise will be PROPACK.

Hi @ilayn – pinging @lesteve who would have more experience and can answer this better than I would, since he has historically maintained the SciPy test suite for Pyodide in a separate repository at https://github.com/lesteve/scipy-tests-pyodide (though it now runs with the `[scipy]` commit marker in Pyodide). I would suggest PROPACK, though, since we have some function signature mismatches with it that we currently skip.


> Sorry, I missed the ping while traveling. If it's too much trouble to get the shared library working, I'd prefer removing the special error handling functionality altogether rather than shipping a broken version that only works for half of the functions.

No worries, @steppi. I converted the shared library to a static one to allow compilation to proceed, and fortunately, all the tests passed, we're not seeing anything problematic.

lesteve:
> Hi @ilayn – pinging @lesteve who would have more experience and can answer this better than I would, since he has historically maintained the SciPy test suite for Pyodide in a separate repository at [lesteve/scipy-tests-pyodide](https://github.com/lesteve/scipy-tests-pyodide) (though it now runs with the [scipy] commit marker in Pyodide). I would suggest PROPACK, though, since we have some function signature mismatches with it that we currently skip.

No strong opinion on this, I would think that making Pyodide's life easier for Scipy support is only a small part of the story behind the Fortran rewrite.

ilayn:
> No strong opinion on this, I would think that making Pyodide's life easier for Scipy support is only a small part of the story behind the Fortran rewrite. 

Definitely and at the same time, we can make life easier by attacking the more problematic ones earlier for Pyodide. For me, the order does not matter, they are all excessively boring work,  but it might matter for you. 

But I guess we tackled the bigger problems so we are at the noise level in terms of preference which is good news for me. I'll see which one is a quicker win then.

hoodmane (approved):
Thanks @agriyakhetarpal!

agriyakhetarpal:
Thanks for the reviews! I initially thought I could wait until we can get the install tags functionality working, but I think that is not a blocker because we remove the tests away anyway using `unvendor-tests: true`; the install tags (if we get it to work) would just not build the test modules any time soon. If Ralf has no other comments on these changes, then we can merge this and I can get started with SciPy v1.14.1 that just came out – I've just been playing catch up over the past week 😁

rgommers:
This seems fine to me to merge. I'll comment on the f2py generator issue higher up - that patch isn't desirable, but let's not let that block a merge, hopefully we can figure that one out and remove the patch again later.

packages/scipy/patches/0015-Remove-f2py-generator.patch
rgommers: `f2py` is a code generator, just like `cython`. Hence it should be available and run on the build machine when cross-compiling. It's a little puzzling to me why it goes missing. In the old way, `numpy.f2py` was also used for code generation, so it was available somewhere. I _think_ what may be happening in that case is that it was run with the host (i.e. non-native on the build machine) interpreter. The environments set up by `pyodide-build` are probably not clean build/host envs, but a `crossenv`-like hack where the host interpreter runs fine on the build machine? That would explain why f2py codegen works but the `f2py` entrypoint isn't available.

packages/scipy/meta.yaml
rgommers: It's not a bug - all modules get built. Install tags are purely an install-time feature, selecting what targets get installed during the `meson install` step (which is separate from `meson compile`). If you don't want a target to be built at all, you have to patch it out still.

agriyakhetarpal:
Thanks, Ralf! Merged this, then. I'll put up a PR for the next version.

PR #5020, opened 2024-08-20
Add `arro3` to package list

kylebarron:
### Description

[arro3](https://github.com/kylebarron/arro3) is a new Python library for working with Apache Arrow data. See [documentation](https://kylebarron.dev/arro3/latest/).

arro3 is distributed with [namespace packaging](https://packaging.python.org/en/latest/guides/packaging-namespace-packages/). This means that users who only need to represent Arrow data (say as interop between Arrow-based Python packages) only need the core module. So this PR creates three new recipes: `arro3-core`, `arro3-io`, and `arro3-compute`.

cc @hoodmane who [suggested contributing this](https://www.linkedin.com/feed/update/urn:li:activity:7229497495783952384?commentUrn=urn%3Ali%3Acomment%3A%28activity%3A7229497495783952384%2C7229577417583931392%29&replyUrn=urn%3Ali%3Acomment%3A%28activity%3A7229497495783952384%2C7230137570712584192%29&dashCommentUrn=urn%3Ali%3Afsd_comment%3A%287229577417583931392%2Curn%3Ali%3Aactivity%3A7229497495783952384%29&dashReplyUrn=urn%3Ali%3Afsd_comment%3A%287230137570712584192%2Curn%3Ali%3Aactivity%3A7229497495783952384%29).

I'm expecting to make an arro3 0.3 release in the next few days and will bump these versions to 0.3 when that's released. (It can either be this or a follow up PR; I saw the [0.27.0a1](https://github.com/pyodide/pyodide/releases/tag/0.27.0a1) tag was published and figured I should make this PR sooner than later.)

### Checklists

- [x] Add / update tests


hoodmane:
It'll be a few more weeks at least before we release 0.27 I think so no huge rush.

hoodmane:
Thanks @kylebarron!

packages/arro3-core/test_arro3_core.py
```diff
@@ -0,0 +1,8 @@
+from pytest_pyodide import run_in_pyodide
+
+
+@run_in_pyodide(packages=["pyarrow", "numpy", "pandas"])
+def test_read_write_parquet(selenium):
+    from arro3.core import Array, DataType
+
+    arr = Array([1, 2, 3], DataType.int64())
```
kylebarron, line 8:
```suggestion
    Array([1, 2, 3], DataType.int64())
```

agriyakhetarpal:
Not a blocker, but it would be great if you could add licensing fields to the recipes too.

packages/arro3-core/test_arro3_core.py
```diff
@@ -0,0 +1,8 @@
+from pytest_pyodide import run_in_pyodide
+
+
+@run_in_pyodide(packages=["pyarrow", "numpy", "pandas"])
```
hoodmane, line 4: Don't you want `packages=["arro3"]` here? You don't directly import any of the others, so they shouldn't be in the list here.

kylebarron: oh yes, thank you. I copied and pasted from the pyarrow recipe

kylebarron:
I added license fields and updated to arro3 0.3.0.

packages/arro3-compute/meta.yaml
```diff
@@ -0,0 +1,20 @@
+package:
+  name: arro3-compute
+  version: 0.3.0
+  top-level:
+    - "arro3.compute"
```
hoodmane, line 5: CI is failing because top-level isn't supposed to contain a dot. Either these should be `top-level: [arro3]` or we need to fix `find_imports`. @ryanking13 perhaps it would make sense to change `find_imports` to support this:

```patch
+def _add_prefixes(s: set[str], mod: str) -> None:
+    [current, *rest] = mod.split(".")
+    s.add(current)
+    for part in rest:
+        current += f".{part}"
+        s.add(current)
+
+
 def find_imports(source: str) -> list[str]:
     """
     Finds the imports in a Python source code string
@@ -635,17 +643,17 @@ def find_imports(source: str) -> list[str]:
         mod = ast.parse(source)
     except SyntaxError:
         return []
-    imports = set()
+    imports: set[str] = set()
     for node in ast.walk(mod):
         if isinstance(node, ast.Import):
             for name in node.names:
                 node_name = name.name
-                imports.add(node_name.split(".")[0])
+                _add_prefixes(imports, node_name)
         elif isinstance(node, ast.ImportFrom):
             module_name = node.module
             if module_name is None:
                 continue
-            imports.add(module_name.split(".")[0])
+            _add_prefixes(imports, module_name)
     return list(sorted(imports))
```

kylebarron: I really like namespace packaging but it does tend to break tools 😅 

> Either these should be top-level: [arro3]

For a namespace module `arro3.core`, `import arro3` will do _nothing_ because there's no `__init__.py` file in the `arro3` directory. It's not until you reach `arro3/core/__init__.py` that there's anything to import.

ryanking13: Yes, I think it makes sense to change find_imports. Changing `top-level: [arro3]` would result in importing `arro3` loading all three arro3 subpackages, which might not be the intended behavior. I am not sure it is a good practice to locate three different packages in a same top-level directory though.

kylebarron: > Changing `top-level: [arro3]` would result in importing `arro3` loading all three arro3 subpackages

No, with [namespace packaging](https://packaging.python.org/en/latest/guides/packaging-namespace-packages/), calling `import arro3` does nothing. No code is loaded until importing one of the submodules:

```
In [1]: import arro3

In [2]: arro3.core.Array
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
Cell In[2], line 1
----> 1 arro3.core.Array

AttributeError: module 'arro3' has no attribute 'core'

In [3]: import arro3.core

In [4]: arro3.core.Array
Out[4]: arro3.core._core.Array
```

hoodmane:
Looks good @kylebarron, please add a changelog entry and then we will merge.

kylebarron:
Added to changelog

PR #5021, opened 2024-08-20
chore: bump PyArrow test timeout value

agriyakhetarpal:
### Description

`test_pyarrow.py::test_read_write_parquet` (#4950) has faced intermittent timeouts in both the Chrome and the Firefox tests (which I have generally seen are slower than the Node.js ones), as noticed over the past few days in unrelated test failures in other PRs. This PR increases the timeout value so that the timeout is statistically unlikely.

Also, tagging @joemarshall as the recipe maintainer for PyArrow, just in case this is a known problem or something that would merit a deeper look.

### Checklists

N/A

### DIFF
--- a/packages/pyarrow/test_pyarrow.py
+++ b/packages/pyarrow/test_pyarrow.py
@@ -1,6 +1,8 @@
+import pytest
 from pytest_pyodide import run_in_pyodide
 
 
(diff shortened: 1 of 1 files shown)


hoodmane:
Thanks!

PR #5025, opened 2024-08-22
Correct `PyodideAPI.FS` type declaration

CNSeniorious000:
### Description

typescript declaration for `pyodide.FS` is missing

### Checklists

- [ ] Add / update tests

### DIFF
--- a/src/js/api.ts
+++ b/src/js/api.ts
@@ -149,7 +149,7 @@ export class PyodideAPI {
    * are available as members of ``FS.filesystems``:
    * ``IDBFS``, ``NODEFS``, ``PROXYFS``, ``WORKERFS``.
    */
-  static FS = {} as any;
+  static FS = {} as typeof Module.FS;
   /**
    * An alias to the `Emscripten Path API
    * <https://github.com/emscripten-core/emscripten/blob/main/src/library_path.js>`_.


CNSeniorious000:
I also wonder if this type declaration is hand-written and maybe it is out of date:

https://github.com/pyodide/pyodide/blob/22b667019a2dce02a1f458f43af11c15270c59f6/src/js/types.ts#L213-L262

Useful methods like `FS.unmount` is not there.

hoodmane:
Yeah we only include methods that are used somewhere in our code.

CNSeniorious000:
Why not just use emscripten's type declaration itself from [@types/emscripten](https://www.npmjs.com/package/@types/emscripten)?

hoodmane:
Yeah we could use `@types/emscripten`.

hoodmane (approved):

ryanking13:
`@types/emscripten` was based on Emscripten 1.39 (though it is occasionally updated), and we need to check if those types still work for the latest Emscripten.

hoodmane:
I wonder if we're getting confused and the version number is just a coincidence? There's been commits in the last few months:
https://github.com/DefinitelyTyped/DefinitelyTyped/commits/master/types/emscripten/index.d.ts

But it isn't versioned with Emscripten so there are definitely going to be some incorrect definitions.

ryanking13:
Yep, maintaining `@types/emscripten` has been discussed for a long time. If I remember correctly, even at the Jupyterlite workshop in 2022, there was a discussion about maintaining an Emscripten types so that libraries like Pyodide, Jupyterlite, and PyScript could all utilize it. However, it is not that interesting task that most people want to do, so it got lost in the discussion.

So, @CNSeniorious000, if you are interested in upstreaming and maintaining Emscripten types, I highly appreciate it.

PR #5033, opened 2024-08-23
Snapshots: Add support for custom serializer

hoodmane:
The user may pass a serializer function to `makeSnapshot`. It will be called with JavaScript objects that we don't know how to snapshot. It must return an object that we can `JSON.stringify`. When restoring the snapshot, the user must pass a corresponding deserializer that will take the object returned by the serializer and create the original JSON object.

- [x] Add / update tests

### DIFF
--- a/src/js/api.ts
+++ b/src/js/api.ts
@@ -666,13 +666,17 @@ export class PyodideAPI {
     return orig;
   }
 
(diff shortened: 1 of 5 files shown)


ryanking13 (approved):
Thanks!

Issue #4892, opened 2024-06-26
Add datafiles package type


joemarshall:
## 🚀 Feature

A package type that just unzips to a given directory on the pyodide filesystem (optionally overwriting directories if they exist)

### Motivation

In #4889 I add a package that lets you install the unix timezone database to the right place on your pyodide runtime. You have to import the python package which then copies the timezone data to the right place. It would be nicer to have a way to add a package to pyodide that just installed the files directly to where I want them on the filesystem (`/usr/share/zoneinfo`)

### Alternatives

The alternative is to do it via a python module, as I do in #4889 but this means patching any modules that need it as a dependency to make sure the import the module which is a pain.

### Additional context

I hit this in building Apache Arrow - it requires the timezone database to be in the same format and in the place it lives on unix systems, and Arrow maintainers don't want to embed tz data into the Python wheel.

ryanking13:
I think supporting `data_files` (https://github.com/pyodide/micropip/issues/18) can do the trick instead of adding a separate package type. While `data_files` support in setuptools is deprecated, AFAIK there isn't a standard for installing arbitrary data files ([discussion](https://discuss.python.org/t/should-there-be-a-new-standard-for-installing-arbitrary-data-files/7853)), so probably supporting data_files is the most possible way for us.

ryanking13:
We've added support for `data_files`, so I don't think this is needed anymore.

PR #5034, opened 2024-08-24
Support installing data files with pyodide.loadPackage

ryanking13:
### Description

Support installing [data files](https://packaging.python.org/en/latest/specifications/binary-distribution-format/#the-data-directory) when loading packages with `pyodide.loadPackage()`.

Data files is supported in a few build systems including setuptools and hatch. While this is not a standard way of packaging files, a few packages including Jupyter notebook relies on it.

- Related issues
  - https://github.com/pyodide/micropip/issues/18
    - note that installing packages from URL using micropip.install will not support data files until we unify the installation mechasim of pyodide.loadPackage and micropip.install (#4914)
  - https://github.com/pyodide/pyodide/issues/4892

### Checklists

- [x] Add / update tests


hoodmane (approved):

PR #5039, opened 2024-08-28
Return all prefixes of the import path from `find_imports`

hoodmane:
Instead of just returning `["a"]` when we find `import a.b.c`, return `["a", "a.b", "a.b.c"]`. 
This fixes CI in #5020. As an alternative, we might want to forbid dots in the top-level field, but in that case we should check for it when loading `meta.yaml`.

- [x] Update changelog
- [x] Update tests
- [x] Update documentation


ryanking13 (approved):
Thanks! This would need a changelog.

PR #5040, opened 2024-08-29
MAINT Use dependency injection for JS PackageManager

ryanking13:
### Description

Updates PackageManager class to accept `API` and `Module` as a parameter instead of using a globally defined variable. This enables unit test for Package Manager class as we can inject mock object when testing.

### Checklists

- [x] Add / update tests


ryanking13:
@hoodmane 

I am not sure how should I interpret the snapshot error:

```
Error: Failed to get value at index 7
    at API.makeSnapshot (/root/repo/dist/pyodide.asm.js:10:103507)
    at Object.makeMemorySnapshot (/root/repo/dist/pyodide.asm.js:10:110080)
    at file:///root/repo/dist/makesnap.mjs:8:47
```

hoodmane:
I'll take a look soon.

ryanking13:
Thanks!

src/js/dynload.ts
```diff
@@ -201,5 +201,7 @@ export async function loadDynlibsFromPackage(
   }
 }
 
-API.loadDynlib = loadDynlib;
-API.loadDynlibsFromPackage = loadDynlibsFromPackage;
+if (typeof globalThis.API !== "undefined") {
```
hoodmane, line 204: You'd usually either want the condition to be `typeof API !== "undefined"` or `globalThis.API != undefined`. Or maybe just `if (globalThis.API)`. 

But why would `globalThis.API` be defined? Is this for testing? If so could you add a comment?

src/js/load-package.ts
```diff
@@ -165,7 +192,14 @@ class PackageManager {
     const loadedPackageData = new Set<InternalPackageData>();
     const messageCallback = options.messageCallback || console.log;
     const errorCallback = options.errorCallback || console.error;
-    if (names instanceof PyProxy) {
+
+    // originally, this condition was "names instanceof PyProxy",
+    // but it is changed to check names.toJs so that we can use type-only import for PyProxy and remove side effects.
+    // this change is required to run unit tests against this file, when global API or Module is not available.
+    // TODO: remove side effects from pyproxy.ts so that we can directly import PyProxy
```
hoodmane, line 199: Huh, I wonder how hard this will be.

hoodmane: But this replacement code isn't awful to be honest.

src/js/load-package.ts
```diff
@@ -165,7 +192,14 @@ class PackageManager {
...
+    // @ts-ignore
+    if (typeof names.toJs === "function") {
+      // @ts-ignore
       names = names.toJs();
     }
     if (!Array.isArray(names)) {
```
hoodmane, line 205: I wonder if we should check here if `typeof names  === "string"` and then we wouldn't need the type assertion.

hoodmane (approved):
Looks reasonable to me.

src/js/load-package.ts
ryanking13: I think we can fix it by cleaning up the pyproxy.ts code a bit, but it didn't work out easily the first few times I tried.

src/js/load-package.ts
ryanking13: Thanks for the suggestion. I tried that change but TS was complaining about the PyProxy type as the type is not narrowed enough.

src/js/dynload.ts
ryanking13: `typeof API !== "undefined"` worked. Thanks!

src/js/snapshot.ts
```diff
@@ -12,6 +12,7 @@ export function getExpectedKeys() {
     API,
     scheduleCallback,
     API,
+    {},
```
hoodmane, line 15: Confusing, this just undoes my change? You have any idea which of your subsequent changes made it go back? We probably need to see if we can come up with a better way to keep this list up to date...

ryanking13: > Confusing, this just undoes my change? You have any idea which of your subsequent changes made it go back? We probably need to see if we can come up with a better way to keep this list up to date...

I think changing `typeof globalThis.API` ==> `typeof API` made it go back for some reason... I have no idea way.

PR #5055, opened 2024-09-05
Update pydantic and pydantic_core

hoodmane:
`pydantic_core` is not compatible with Python 3.12.5 without this update. This should reduce the change of problems.

### DIFF
--- a/packages/pydantic/meta.yaml
+++ b/packages/pydantic/meta.yaml
@@ -1,11 +1,11 @@
 package:
   name: pydantic
-  version: 2.7.0
+  version: 2.8.2
   top-level:
     - pydantic
 source:
-  sha256: b5ecdd42262ca2462e2624793551e80911a1e989f462910bb81aef974b4bb383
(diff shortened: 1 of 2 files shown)


agriyakhetarpal (approved):
Thanks, @hoodmane!

agriyakhetarpal:
I can't re-run the failing builds for CircleCI here, but I can do so for other PRs – perhaps this is because you're the owner

vladsavelyev:
I wonder if it was actually updated in the registry? The webpage lists the older versions: https://pyodide.org/en/stable/usage/packages-in-pyodide.html

Screenshot 2024-09-13 at 11 04 48

And when I run `await micropip.install("pydantic==2.8.2");`, it errors with
```
Can't find a pure Python 3 wheel for 'pydantic-core==2.20.1'
```

`self.pyodide.loadPackage("pydantic")` also installs an older version.

vladsavelyev:
Gotcha, thank you so much for explaining 🙏

PR #5056, opened 2024-09-05
Improve std stream redirects in Console

### LINKED ISSUE #5053: `sys.stderr` has no attribute `writelines`
Not sure if this is a bug or feature request, depends how one views it :slightly_smiling_face: 

## 🐛 Bug

`sys.stderr` inside Pyodide REPL has no attribute `writelines`. As per Python docs, [`sys.stderr`](https://docs.python.org/3/library/sys.html#sys.stderr) is a file object with the interface defined in the [`io` module](https://docs.python.org/3/library/io.html#io.IOBase), which includes the `writelines()` method.

### To Reproduce

Open [Pyodide REPL](https://pyodide.org/en/stable/console.html) in the browser.
Evaluate the following code in the REPL:
```python
>>> import sys
>>> sys.stderr.write('Hello, World!\n')
Hello, World!
>>> sys.stderr.writelines(['Hello, ', 'World!\n'])
AttributeError: '_WriteStream' object has no attribute 'writelines'
```

### Expected behavior

Running the same in CPython 3.12 proper:
```python
>>> import sys
>>> sys.stderr.write('Hello, World!\n')
Hello, World!
14
>>> sys.stderr.writelines(['Hello, ', 'World!\n'])
Hello, World!
```

(NB there is another potentially problematic difference that the Pyodide flavour does not return the number of bytes written.)

### Environment

https://pyodide.org/en/stable/console.html on Mozilla Firefox 130.0 (64-bit):
```
Welcome to the Pyodide 0.26.2 terminal emulator 🐍
Python 3.12.1 (main, Jul 26 2024 14:03:47) on WebAssembly/Emscripten
```

### Additional context

`None`

hoodmane:
This hopefully resolves #5053 and makes the redirected console streams behave generally more like the default sys.stdout etc.

- [x] Add / update tests


src/py/pyodide/console.py
```diff
@@ -39,43 +40,102 @@ class redirect_stdin(_RedirectStream[Any]):
...
+        if 0 <= size < len(self._buffer):
+            result = self._buffer[:size]
+            self._buffer = self._buffer[size:]
+            return result
+        if size >= 0:
+            size -= len(self._buffer)
+        result = self._buffer
+        got = self._read_handler(size)
+        self._buffer = got[size:]
+        return result + got[:size]
```
ryanking13, line 125: If the intention is to read all buffer, shouldn't this be

```py
if size >= 0:
    size = len(self._buffer)
```

?

src/py/pyodide/console.py
```diff
@@ -39,43 +40,102 @@ class redirect_stdin(_RedirectStream[Any]):
...
+            # For some reason sys.stdin.read(None) works, but
+            # sys.stdin.readline(None) raises a TypeError
+            size = -1
+        if not isinstance(size, int):
+            raise TypeError(
+                f"argument should be integer or None, not '{type(size).__name__}'"
+            )
+        if 0 <= size < len(self._buffer):
+            result = self._buffer[:size]
+            self._buffer = self._buffer[size:]
```
ryanking13, line 118: (no action required for this PR)

If the buffer is large and one calls a read multiple times with a very short size, this can be a performance bottleneck, as slicing an array will take O(N). If we encounter that kind of issue. We can probably optimize this part.

ryanking13 (approved):
Thanks! I left some minor comments but it generally looks good to me.

src/py/pyodide/console.py
hoodmane: Yeah I suppose we should use a cyclic buffer ideally. Can leave that as a todo I guess...

src/py/pyodide/console.py
hoodmane: Well we need to call `_read_handler` with how much more we need. If we've been asked for a positive amount which is within the range of what we already have in the buffer, we just return a chunk of the buffer. If not, we've either been given a negative size (meaning read as much as possible) or a size larger than the buffer (meaning we have to make an onward call to read only the remainder).

PR #5058, opened 2024-09-05
Updates for `statsmodels`'s recipe and bump to v0.14.4

agriyakhetarpal:
### Description

This PR improves the recipe for the `statsmodels` package:
It backports https://github.com/statsmodels/statsmodels/pull/9270, which was introduced in the v0.14.2 release cycle, to the changes from the v0.14.2 sdist as a patch. It can be removed in v0.14.3. Additionally, as mentioned in https://github.com/pyodide/pyodide/pull/4871#issuecomment-2325393399, it's beneficial because it fixes some import errors for the statsmodels API that were not reported earlier by anyone else (to the best of my knowledge).
Some regression tests have been added. `test_zero_collinear` encountered an unresolved symbol `pow_dd` in SciPy v1.12.0 due to OpenBLAS, which has been fixed with other in-tree updates (as mentioned below).
- I have had to set `unvendor-tests: false` because the test setup for both aforementioned tests is convoluted and lengthy, and bringing it down to a single reproducer is a bit tricky. Please let me know if there are alternative ways to test `statsmodels` without having to unvendor the tests.
- I added myself as a recipe maintainer so that I can be pinged when `statsmodels` v0.14.3 comes out and the recipe here can be updated.
- Lastly, `patsy` and `pandas` are not build-time dependencies, so the dependencies section was cleaned up. This might help `statsmodels` build slightly quicker locally and in CI (even if it's marginal).

#### Additional context

- See also: https://github.com/statsmodels/statsmodels/pull/9343
- SciPy in-tree updates: #4719, #5012, #5031

### Checklists

- [x] Add / update tests


packages/statsmodels/patches/Add-WASM-support.patch
```diff
@@ -0,0 +1,813 @@
+diff --git a/.github/workflows/emscripten.yml b/.github/workflows/emscripten.yml
+new file mode 100644
+index 000000000..e5716b090
+--- /dev/null
++++ b/.github/workflows/emscripten.yml
```
ryanking13, line 5: Seems like this patch includes some non-necessary changes like GHA yaml file update. But if we can update the statmodels to 0.14.3 soon and remove this patch, I am okay to leave this as-is,

ryanking13 (approved):
Thanks!

agriyakhetarpal:
Thanks for taking a look, @ryanking13! I'll still try to produce a smaller reproducer, but if I can't, I'll merge this as is, since the tests are not too big. Let me mark this as a draft for now.

agriyakhetarpal:
https://github.com/statsmodels/statsmodels/releases/tag/v0.14.3 was released today, so I'll update the recipe to that version and re-work the patch.

agriyakhetarpal:
Actually, we can probably get v0.14.4 to come out soon, since I'm backporting the patch to the release branch (https://github.com/statsmodels/statsmodels/pull/9365). We should be able to drop the patch here, then.

agriyakhetarpal:
https://github.com/statsmodels/statsmodels/releases/tag/v0.14.4 is now up, so we can drop the patch. I'll wait for it to be up on PyPI, update this PR, and then merge it.

agriyakhetarpal:
The sdist is now up: https://pypi.org/project/statsmodels/0.14.4/#files and other wheels are trickling in

packages/statsmodels/patches/Add-WASM-support.patch
agriyakhetarpal: Resolving this conversation since the patch was dropped entirely.

agriyakhetarpal:
Tests are passing across the trio of Node.js, Firefox, and Chrome and statsmodels is now up to date with its PyPI counterpart. Thanks again for the review!

PR #5059, opened 2024-09-05
Fix LAPACK larfg function signature mismatch (SciPy)

agriyakhetarpal:
### Description

This PR closes #3379. It patches the function signature for the `larfg` subroutine in `scipy/linalg/flapack_other.pyf.src` to remove an `lx` additional argument that was being passed, resolving the original reproducer posted in the issue. However, I am not as acquainted with Fortran, so please let me know if there is a better way to do this and I can try to learn that.

@ilayn, I know that you had mentioned in https://github.com/pyodide/pyodide/issues/3379#issuecomment-2160587164 that you would do it when you would have got the chance; I hope it is okay with you for me to take it up because I felt that this was a reasonably easy fix to incorporate! I have granted you co-authorship in the patch file, thank you :)

### Checklists

- [x] Add / update tests

PR #5060, opened 2024-09-06
Fix SciPy LAPACK larf function signature mismatch

agriyakhetarpal:
### Description

From #5059 – I added the relevant test in `test_scipy.py`, but forgot to remove the test skip from `scipy-conftest.py`.

### Checklists

- [x] Add / update tests

### DIFF
--- a/packages/scipy/scipy-conftest.py
+++ b/packages/scipy/scipy-conftest.py
@@ -89,7 +89,6 @@ tests_to_mark = [
     # scipy/linalg tests
     ("test_blas.+test_complex_dotu", skip, todo_signature_mismatch_msg),
     ("test_cython_blas.+complex", skip, todo_signature_mismatch_msg),
-    ("test_lapack.py.+larfg_larf", skip, todo_signature_mismatch_msg),
     # scipy/ndimage/tests
     ("test_filters.py::TestThreading", xfail, thread_msg),
     # scipy/optimize/tests


hoodmane:
Do you think it was it a regression?

agriyakhetarpal:
The reproducer @lesteve had posted in the original issue (#3865) doesn't fail anymore (which is great), but `scipy/linalg/tests/test_lapack.py::test_larfg_larf` still fails – both code snippets are different. The fix in that PR was thus only partial – so, not a regression.

agriyakhetarpal:
Running the test step-by-step, we get:

```python
alpha, x, tau = larfg(a.shape[0]-1, a[1, 0], a[2:, 0]) # this works now
```

and further below:

```python
a[1:, :] = larf(v, tau.conjugate(), a[1:, :], np.zeros(a.shape[1]))
```

returns

`RuntimeError: call_indirect to a signature that does not match (evaluating '__pyproxy_apply(r,e,n,o,s)')`

so #5059 indeed fixed `larfg`, but not `larf`.

agriyakhetarpal:
I can further attempt to fix the signature for `larf` here and un-skip the test completely, but it looks a little more complicated to handle than `larfg`, let's see :) Otherwise, we definitely had an improvement, even though it was a partial one, and no regressions were involved – there is, hence, no need to revert #5059.

lesteve:
OK, thanks for trying, I got a bit too enthusiastic :sweat_smile:

hoodmane:
> The reproducer @lesteve had posted in the original issue (https://github.com/pyodide/pyodide/issues/3865) doesn't fail anymore (which is great)

So we definitely shouldn't revert I'd think.

agriyakhetarpal:
And I see https://github.com/scipy/scipy/pull/21290 with the problem :) @hoodmane, do we need to patch the entire file here to fix the line endings?

agriyakhetarpal:
We could fix this on the `pyodide-build` side by changing the patch command to `git am <...>`, but I fear breaking other patches in the wake of other files having improper endings, and `patch --binary` is the best option, anyway. I couldn't get the patch to apply so far, so I'm not sure what I am missing.

PR #5010, opened 2024-08-15
Add as_py_json method to appropriate JsProxies

hoodmane:
This is the Js to Python analogue of `asJsJson`. There's a fair amount of overlap in behavior with as_object_map here.

- [x] Add a [CHANGELOG](https://github.com/pyodide/pyodide/blob/main/docs/project/changelog.md) entry
- [x] Add / update tests
- [x] Add new / update outdated documentation

PR #5069, opened 2024-09-11
jsbind: Add support for Json as a JS to python annotation

hoodmane:
This is a quick followup to #5010 that makes `jsbind` use `as_py_json` in the expected way. No changelog needed since it's still not public API.

- [x] Add / update tests


ryanking13 (approved):
Thanks!

hoodmane:
Thanks for the review!

PR #5072, opened 2024-09-13
chore: bump iminuit to v2.29.1, and `scikit-learn` to v1.5.2

agriyakhetarpal:
### Description

This PR bumps two packages:
- `iminuit` to [version `2.29.1`](https://github.com/scikit-hep/iminuit/releases/tag/v2.29.1). The patch previously added in #4767 can now be deleted and dependence on `packaging` can be removed. Tagging the recipe maintainer @henryiii for visibility.
- `scikit-learn` to [version `1.5.2`](https://github.com/scikit-learn/scikit-learn/releases/tag/1.5.2). Tagging the recipe maintainer @lesteve for visibility.

I was previously involved with helping review the files associated with their recipes, so I thought it would be good to get updated versions of them before 0.27 stable lands.

### Checklists

- [x] Add a [CHANGELOG](https://github.com/pyodide/pyodide/blob/main/docs/project/changelog.md) entry
- [ ] Add / update tests

PR #5074, opened 2024-09-13
Remove `-fexceptions` from `pybind11`-based projects: `iminuit` and `boost-histogram`, bump `boost-histogram` to 1.5.0

agriyakhetarpal:
### Description

Post https://github.com/pybind/pybind11/pull/5298 and https://github.com/pybind/pybind11/pull/5301, we can remove `-fexceptions` from the compiler and the linker flags, since `pybind11` version 2.13.3 and later will set those for us. I removed them from `iminuit` over from #5072 and from `boost-histogram`.

Additionally, bumped `boost-histogram` to version 1.5.0. Are there any more packages that I might be missing?

cc: @henryiii

### Checklists


ryanking13 (approved):
Thanks!

PR #5086, opened 2024-09-25
Fix invalid package names

### LINKED ISSUE #4909: `stack-data`'s dependency map is wrong
## 🐛 Bug

`stack-data` is directly dependent on `pure-eval`, `executing` and `asttokens`. But the `requirements` field in [`meta.yaml`](https://github.com/pyodide/pyodide/blob/5fb36a458be37270efecdc9638202a13b62d28f8/packages/stack_data/meta.yaml) is empty.

### To Reproduce

```py
import stack_data
```

### Expected behavior

```
    import stack_data
  File "/lib/python3.12/site-packages/stack_data/__init__.py", line 1, in <module>
    from .core import Source, FrameInfo, markers_from_ranges, Options, LINE_GAP, Line, Variable, RangeInLine, \
  File "/lib/python3.12/site-packages/stack_data/core.py", line 15, in <module>
    import executing
ModuleNotFoundError: No module named 'executing'
```

### Environment

- Pyodide Version: 0.26.1
- Browser version: Microsoft Edge 126.0.2592.81

ryanking13:
Close #4909 

I guess we need to find a way to prevent this kind of error (probably in `pyodide-build`).


agriyakhetarpal (approved):
Thanks, @ryanking13! Does the regex in https://packaging.python.org/en/latest/specifications/name-normalization/#name-normalization work for our use case? If yes, either of us could integrate that in `pyodide-build`. I'm happy to step back if you wish to take it; otherwise, I could do it.

ryanking13:
Thank you for the review!

> Does the regex in https://packaging.python.org/en/latest/specifications/name-normalization/#name-normalization work for our use case?

Unfortunately no. Either normalized and not-normalized names can be a package name. I think the most accurate way would be

1) build the package first
2) check the package name in the METADATA (which is the source of truth of the package name)
3) compare it with the name in meta.yaml

But there might be a tricky case like scipy-test (https://github.com/pyodide/pyodide/pull/5031#issuecomment-2308531792), so I am not sure if we always want to make the two names the same.

agriyakhetarpal:
Sorry – I guess I meant to share https://packaging.python.org/en/latest/specifications/name-normalization/#name-format (the link to the heading above it), which is the regex we can use to validate the package name. Both normalised and non-normalised package names can exist, though, I assume that the most common build backends nowadays are adhering to standards and producing normalised package names already.

That said, https://packaging.python.org/en/latest/specifications/core-metadata/#name states that we should normalise the package name before we compare it with the one listed in the recipe file.

In the case of `scipy-tests`, I am not entirely sure if I follow the discussion in the right manner. The package name in the recipe file will be `scipy-tests`, and the wheel name would be `scipy_tests-1.15.0-cp312-cp312-pyodide_2024_0-wasm32.whl`, wouldn't it?

ryanking13:
> In the case of scipy-tests, I am not entirely sure if I follow the discussion in the right manner. The package name in the recipe file will be scipy-tests, and the wheel name would be scipy_tests-1.15.0-cp312-cp312-pyodide_2024_0-wasm32.whl, wouldn't it?

Oh, I must be wrong. I understood the Scipy package to have two versions: one with tests and one without. But is there will be another package "scipy-tests" then it should be okay.

PR #5087, opened 2024-09-25
Update pillow_heif to 0.18.0

teonbrooks:
### Description

Hi there. I would like to update `pillow_heif` to its latest version `0.18.0`. Updating it will align it with a more current version of PIllow, which Pyodide already supports.

### Checklists

- [x] Add / update tests


packages/pillow_heif/meta.yaml
```diff
@@ -1,9 +1,9 @@
 package:
   name: pillow_heif
```
agriyakhetarpal, line 2:
```suggestion
  name: pillow-heif
```

agriyakhetarpal (approved):
Thanks for the update, @teonbrooks! If you're interested, if would be great if you could add yourself as a recipe maintainer by including your name here. This would help us in the sense that we would know who to ping if we have any troubles with the recipe in the coming months and years. :)

teonbrooks:
hi @agriyakhetarpal, no prob! I think I will pass on being the recipe maintainer.  I don't have too deep of knowledge on how to patch it if things go awry; I'm just using the library in a project and noticed that it is pretty far behind in releases.

agriyakhetarpal:
No worries at all! Do you happen to know why the build for it is failing? I can take a look in detail later in the day, too.

packages/pillow_heif/meta.yaml
```diff
@@ -1,9 +1,9 @@
 package:
-  name: pillow_heif
-  version: 0.8.0
+  name: pillow-heif
```
ryanking13, line 2: Thanks for the catch. Could you also update the directory name? (pillow_heif -> pillow_heif)

agriyakhetarpal: >  Could you also update the directory name? (pillow_heif -> pillow_heif)

🤔 I think we had to rename the package and directory names to `pillow-heif` (maybe there was a typo here)?

ryanking13: Yes, that's what I meant, but I made a typo 😅

agriyakhetarpal:
The logs display correctly, now. I think we need to either drop the patch or rework it, since the latest version checks for the `libheif` include directories in a different manner in comparison to the previous version: [bigcat88/pillow_heif@`6b3e049`/setup.py#L176](https://github.com/bigcat88/pillow_heif/blob/6b3e0492ab7ef80a8929b457e50e05e630526a06/setup.py#L176).

agriyakhetarpal:
The logs display correctly, now (didn't realise the folder name was the issue). I think we need to either drop the patch or rework it, since the latest version checks for the `libheif` include directories in a different manner in comparison to the previous version: [bigcat88/pillow_heif@`6b3e049`/setup.py#L176](https://github.com/bigcat88/pillow_heif/blob/6b3e0492ab7ef80a8929b457e50e05e630526a06/setup.py#L176)

ryanking13:
Yes, I guess removing the patch and setting [`LIBHEIF_ROOT`](https://github.com/bigcat88/pillow_heif/blob/6b3e0492ab7ef80a8929b457e50e05e630526a06/setup.py#L83) to WASM_LIBRARY_DIR would work

teonbrooks:
@agriyakhetarpal just wondering, is there anything else I could do to help close this PR out?

agriyakhetarpal:
@teonbrooks – I can take a look at it; it's been a while, so I don't remember the original error. I'm not fully familiar with the build process for `pillow-heif`, though.

agriyakhetarpal:
I've been at it for a few hours now and was able to bump it further to version 0.20.0 locally from the currently updated-to version 0.18.0, and I made it respect the `LIBHEIF_ROOT` environment variable (plus moved a few things around) so that it now finds the correct HEIF include files. The issue is that the `libheif` that we ship ourselves is most likely not compatible with the one that `pillow-heif` provides the source and/or build scripts for (i.e., `libheif/linux_build_libs.py`) and subsequently tries to build from (the structure is different across both headers), so we need to either force it to build with its own `libheif` for this, or bump to compatible versions for it (read below, and those might or might not work since there are additional patches that are getting applied to `libheif`). I'm unable to get around an error: `wasm-ld: error: unable to find library -lheif` so far, because I can't find where the library gets copied to after it gets built. Some more searching led me to see that the package disables PEP 517-based build isolation, so we need to build it with `pyodide build --no-build-isolation`, which I am not sure whether it's allowed for us (and if there's a mechanism to add the flag) for the "in-tree" recipes in specific (@ryanking13 might know better about this).

I notice that `pillow-heif` uses `libde265` version 1.0.15 and `libheif` 1.18.2; the ones we have are versions 1.0.8 and 1.12.0 respectively – so we're not very far behind in terms of updating (but subsequent updates to `pillow-heif` would require version bumps to those two packages, too).

Another similar thing we could do is to implement a `libheif/emscripten_build_libs.py` file for this package in specific, and use it here. However, both this option and the option of providing `libheif-for-pillow-heif` and `libe265-for-pillow-heif` recipes come with the burden of additional maintenance until the time we unvendor recipes from the repository and _fully_ delegate the task for upgrading and maintaining recipes to the recipe maintainers (#4987) – if they're willing to, that is. The additional CI time increments wouldn't be too great, either.

Do you have thoughts on what we should proceed with, @ryanking13 and @teonbrooks? In general, picking up the `libheif` libs from `WASM_LIBRARY_DIR` would have worked, but `pillow-heif` has some tricky barriers to get through :D We also have an option of building these WASM wheels for `pillow-heif` out-of-tree with it managing its custom `libheif` and `libe265` Emscripten builds, and adding a link to them here, like we do for DuckDB and PyArrow.

ryanking13:
@agriyakhetarpal Thanks for the detailed information. Let me take a look, and yes, packages with libraries are quite tricky to build when they use no isolation option. If this is too complicated, I think it's beyond our scope and should be addressed by the pillow-heif maintainers.

ryanking13:
Opened a PR (#5178). Ideally, in the future we unvendor recipes, each package may use a different version of shared libraries.

agriyakhetarpal:
> Opened a PR (#5178). Ideally, in the future we unvendor recipes, each package may use a different version of shared libraries.

Thanks, that's what I had in mind! I'll be willing to help on #5178, in the meantime, I've pushed my changes here. Please feel free to take them forward if you wish to build up on those in your approach.

ryanking13:
Superseded by #5178. Thanks for investigating the issue Agriya!

PR #5089, opened 2024-09-28
DOC find pyodide module even if in parent directories

castedo:
### Description

The code in the doc of the main branch only works if `node_modules` is in the project/current/root directory. But node will also find packages in `node_modules` under parent directories, like `../node_modules` in my case, or `../../node_modules` etc...

With this fix/enhancement the pyodide package directory will be found in any of these parent sub directories by using `import.meta.resolve`.


ryanking13 (approved):
Thanks!

PR #5091, opened 2024-09-28
Update docker image to fix rust toolchains

ryanking13:
https://github.com/pyodide/pyodide/pull/4982 updated the build docker image, but for somer reason rust toolchains are not installed in the image, resulting in a build failure. This PR changes the docker image to fix it.


hoodmane:
Sorry about that!

ryanking13:
> Sorry about that! Not really sure what caused the problem, I just did docker build . like normal...

I'm guessing it was a network issue - the toolchain installation failed, but the build was still treated as successful.

PR #5092, opened 2024-10-01
Fix handling of conversion of 0d buffers

### LINKED ISSUE #5084: Wrong conversion from np.int64 to JavaScript
## 🐛 Bug

Calling toJs on a PyProxy with np.int64 returns weird results.

### To Reproduce

import numpy as np
from pyodide.ffi import to_js
to_js(np.int64(1))

### Expected behavior

np.int64(1) is converted to 1

### Environment

- Pyodide Version: Pyodide 0.26.2 and Pyodide dev (from https://cdn.jsdelivr.net/pyodide/dev/full/pyodide.js)
- Browser version: Latest Firefox and latest Chrome

hoodmane:
Also fixes conversion for 64 bit integer buffers. I added a buffer-test package so we can test 0d buffer conversion code without using numpy. In the future ideally we should add more test buffer types to buffer-test so that we can move more coverage of the buffer conversion code out of test_numpy and into the main test suite.

Resolves #5084.

- [x] Add / update tests


packages/buffer-test/buffer-test/setup.py
```diff
@@ -0,0 +1,18 @@
...
+    author="Hood Chatham",
+    author_email="roberthoodchatham@gmail.com",
+    description="Test Python buffers",
+    long_description_content_type="text/markdown",
+    classifiers=[
+        "Programming Language :: Python :: 3",
+        "License :: OSI Approved :: MIT License",
+        "Operating System :: OS Independent",
+    ],
+    # packages=["fpcast_test"],
```
ryanking13, line 15:
```suggestion
```

src/tests/test_typeconversions.py
ryanking13, line 1: How about putting this under buffet-test package directory?

ryanking13 (approved):
Thanks!

.circleci/config.yml
```diff
@@ -659,7 +659,7 @@ workflows:
 
       - test-main:
           name: test-core-chrome-nodylink
-          test-params: --runtime=chrome-no-host -m 'not requires_dynamic_linking' -k "not webworker" src packages/micropip  packages/fpcast-test packages/sharedlib-test-py/ packages/cpp-exceptions-test/
+          test-params: --runtime=chrome-no-host -m 'not requires_dynamic_linking' -k "not webworker" src packages/micropip  packages/fpcast-test packages/sharedlib-test-py/ packages/cpp-exceptions-test/ packages/buffer-test/
```
hoodmane, line 662: It would probably be helpful to extract the list of paths here into a separate variable instead of repeating it 6 times...

PR #5098, opened 2024-10-04
Upgrade sympy from 1.12 to 1.13.3

oscarbenjamin:
### Checklists

- [ ] Add / update tests

### DIFF
--- a/packages/sympy/test_sympy.py
+++ b/packages/sympy/test_sympy.py
@@ -1,8 +1,7 @@
+import pytest
 from pytest_pyodide import run_in_pyodide
 
 from conftest import package_is_built
-
-import pytest
 
 
 def skip_if_not_installed(packages):
--- a/packages/sympy/test_sympy.py
+++ b/packages/sympy/test_sympy.py
@@ -1,5 +1,6 @@
-from conftest import package_is_built
 from pytest_pyodide import run_in_pyodide
+
+from conftest import package_is_built
 
 
 def skip_if_not_installed(packages):
--- a/packages/sympy/test_sympy.py
+++ b/packages/sympy/test_sympy.py
@@ -1,7 +1,4 @@
-import pytest
 from pytest_pyodide import run_in_pyodide
-
-from conftest import package_is_built
 
 
 @run_in_pyodide(packages=["sympy"])
@@ -13,18 +10,6 @@ def test_sympy(selenium):
 
     assert c.subs({a: 3, b: 4}) == 5
 
-
-@run_in_pyodide(packages=["sympy", "python-flint"])
-def test_sympy_and_python_flint(selenium):
-    for package in ["sympy", "python-flint"]:
-        if not package_is_built(package):
-            pytest.skip(f"{package} not built")
-
-    import sympy
-    from sympy.external.gmpy import GROUND_TYPES
-
-    assert GROUND_TYPES == "flint"
-
-    # Use python-flint for factorisation:
+    # Uses python-flint if installed
     x = sympy.symbols("x")
     assert (x**2 - 1).factor() == (x + 1) * (x - 1)
--- a/packages/sympy/test_sympy.py
+++ b/packages/sympy/test_sympy.py
@@ -14,10 +14,10 @@ def test_sympy(selenium):
 @run_in_pyodide(packages=["sympy", "python-flint"])
 def test_sympy_and_python_flint(selenium):
     import sympy
-
     from sympy.external.gmpy import GROUND_TYPES
-    assert GROUND_TYPES == 'flint'
+
+    assert GROUND_TYPES == "flint"
 
     # Use python-flint for factorisation:
-    x = sympy.symbols('x')
+    x = sympy.symbols("x")
     assert (x**2 - 1).factor() == (x + 1) * (x - 1)
--- a/packages/sympy/test_sympy.py
+++ b/packages/sympy/test_sympy.py
@@ -9,3 +9,15 @@ def test_sympy(selenium):
     c = sympy.sqrt(a**2 + b**2)
 
     assert c.subs({a: 3, b: 4}) == 5
+
+
+@run_in_pyodide(packages=["sympy", "python-flint"])
+def test_sympy_and_python_flint(selenium):
+    import sympy
+
+    from sympy.external.gmpy import GROUND_TYPES
+    assert GROUND_TYPES == 'flint'
+
+    # Use python-flint for factorisation:
+    x = sympy.symbols('x')
+    assert (x**2 - 1).factor() == (x + 1) * (x - 1)
--- a/packages/sympy/test_sympy.py
+++ b/packages/sympy/test_sympy.py
@@ -16,7 +16,6 @@ def test_sympy(selenium):
 
 @run_in_pyodide(packages=["sympy", "python-flint"])
 def test_sympy_and_python_flint(selenium):
-
     for package in ["sympy", "python-flint"]:
         if not package_is_built(package):
             pytest.skip(f"{package} not built")
--- a/packages/sympy/test_sympy.py
+++ b/packages/sympy/test_sympy.py
@@ -20,6 +20,7 @@ def test_sympy(selenium):
     assert c.subs({a: 3, b: 4}) == 5
 
 
+@run_in_pyodide(packages=["sympy", "python-flint"])
 def test_sympy_and_python_flint(selenium):
     skip_if_not_installed(["sympy", "python-flint"])
 
--- a/packages/sympy/test_sympy.py
+++ b/packages/sympy/test_sympy.py
@@ -2,6 +2,8 @@ from pytest_pyodide import run_in_pyodide
 
 from conftest import package_is_built
 
+import pytest
+
 
 def skip_if_not_installed(packages):
     for package in packages:
--- a/packages/sympy/test_sympy.py
+++ b/packages/sympy/test_sympy.py
@@ -1,4 +1,11 @@
+from conftest import package_is_built
 from pytest_pyodide import run_in_pyodide
+
+
+def skip_if_not_installed(packages):
+    for package in packages:
+        if not package_is_built(package):
+            pytest.skip(f"{package} not built")
 
 
 @run_in_pyodide(packages=["sympy"])
@@ -11,13 +18,18 @@ def test_sympy(selenium):
     assert c.subs({a: 3, b: 4}) == 5
 
 
-@run_in_pyodide(packages=["sympy", "python-flint"])
 def test_sympy_and_python_flint(selenium):
-    import sympy
-    from sympy.external.gmpy import GROUND_TYPES
+    skip_if_not_installed(["sympy", "python-flint"])
 
-    assert GROUND_TYPES == "flint"
+    @run_in_pyodide(packages=["sympy", "python-flint"])
+    def run(selenium):
+        import sympy
+        from sympy.external.gmpy import GROUND_TYPES
 
-    # Use python-flint for factorisation:
-    x = sympy.symbols("x")
-    assert (x**2 - 1).factor() == (x + 1) * (x - 1)
+        assert GROUND_TYPES == "flint"
+
+        # Use python-flint for factorisation:
+        x = sympy.symbols("x")
+        assert (x**2 - 1).factor() == (x + 1) * (x - 1)
+
+    run()
--- a/packages/sympy/test_sympy.py
+++ b/packages/sympy/test_sympy.py
@@ -4,12 +4,6 @@ from pytest_pyodide import run_in_pyodide
 from conftest import package_is_built
 
 
-def skip_if_not_installed(packages):
-    for package in packages:
-        if not package_is_built(package):
-            pytest.skip(f"{package} not built")
-
-
 @run_in_pyodide(packages=["sympy"])
 def test_sympy(selenium):
     import sympy
@@ -22,17 +16,16 @@ def test_sympy(selenium):
 
 @run_in_pyodide(packages=["sympy", "python-flint"])
 def test_sympy_and_python_flint(selenium):
-    skip_if_not_installed(["sympy", "python-flint"])
 
-    @run_in_pyodide(packages=["sympy", "python-flint"])
-    def run(selenium):
-        import sympy
-        from sympy.external.gmpy import GROUND_TYPES
+    for package in ["sympy", "python-flint"]:
+        if not package_is_built(package):
+            pytest.skip(f"{package} not built")
 
-        assert GROUND_TYPES == "flint"
+    import sympy
+    from sympy.external.gmpy import GROUND_TYPES
 
-        # Use python-flint for factorisation:
-        x = sympy.symbols("x")
-        assert (x**2 - 1).factor() == (x + 1) * (x - 1)
+    assert GROUND_TYPES == "flint"
 
-    run()
+    # Use python-flint for factorisation:
+    x = sympy.symbols("x")
+    assert (x**2 - 1).factor() == (x + 1) * (x - 1)
--- a/docs/project/changelog.md
+++ b/docs/project/changelog.md
@@ -80,7 +80,7 @@ myst:
 - Upgraded `contourpy` to 1.3.0 {pr}`5048`
 - Upgraded `boost-histogram` to 1.5.0 {pr}`5074`
 - Upgraded `duckdb` to 1.1.0 {pr}`5078`
-- Upgraded `sympy` to 1.13.3 {pr}`XXX`
+- Upgraded `sympy` to 1.13.3 {pr}`5098`
 - Added `casadi` 3.6.6 {pr}`4936`, {pr}`5057`
 - Added `pyarrow` 17.0.0 {pr}`4950`
 - Added `rasterio` 1.13.10, `affine` 2.4.0 {pr}`4983`
--- a/docs/project/changelog.md
+++ b/docs/project/changelog.md
@@ -80,6 +80,7 @@ myst:
 - Upgraded `contourpy` to 1.3.0 {pr}`5048`
 - Upgraded `boost-histogram` to 1.5.0 {pr}`5074`
 - Upgraded `duckdb` to 1.1.0 {pr}`5078`
+- Upgraded `sympy` to 1.13.3 {pr}`XXX`
 - Added `casadi` 3.6.6 {pr}`4936`, {pr}`5057`
 - Added `pyarrow` 17.0.0 {pr}`4950`
 - Added `rasterio` 1.13.10, `affine` 2.4.0 {pr}`4983`
--- a/packages/python-flint/meta.yaml
+++ b/packages/python-flint/meta.yaml
@@ -25,3 +25,4 @@ about:
 extra:
   recipe-maintainers:
     - mkoeppe
+    - oscarbenjamin
--- a/packages/sympy/meta.yaml
+++ b/packages/sympy/meta.yaml
@@ -1,6 +1,6 @@
 package:
   name: sympy
-  version: "1.12"
+  version: "1.13.3"
   top-level:
     - isympy
     - sympy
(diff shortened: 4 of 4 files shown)


agriyakhetarpal (approved):

agriyakhetarpal:
Thanks for the update, @oscarbenjamin!

oscarbenjamin:
I notice the OP checklist suggests that maybe adding tests and docs are good. Is this okay for SymPy right now?

I'm happy to add tests/docs but I'm not sure what that might ideally look like. The full sympy test suite would take about an hour under pyodide in CI I think so probably not suitable for use here.

agriyakhetarpal:
Yes, @oscarbenjamin, tests are always welcome – we don't run the entire test suite for projects since it's usually overkill and our CI is too strained for that. However, a few tests for basic popular functionalities (symbolic operations, solvers, expression trees, et cetera) would be good to include. For how they look like, you could refer to the tests for any other project and add the relevant decorator(s) to wrap the tests: `@run_in_pyodide` should suffice. https://pyodide.org/en/stable/development/testing.html might help, please let us know if you need any additional information!

Also, since you've asked – would you like to maintain SymPy's recipe? Having maintainers for packages in the ecosystem listed here helps us in the sense that we'll know who to ping in case anything goes awry when updating them, and users know who to reach out to in case of uncaught bugs or problems. :)

As for running the entire test suite for SymPy within a WASM environment – doing so coupled with a CI job that performs an out-of-tree build is the best way to mark support for Pyodide. https://pyodide.org/en/stable/development/building-and-testing-packages.html has some instructions related to this, and I've been doing exactly this for a few packages in the Scientific Python ecosystem this year. Would you like me to implement this in the SymPy repository sometime?

oscarbenjamin:
I can maintain SymPy's recipe. Would that need to be listed in the metadata somehow?

We did previously have a CI job that ran the full SymPy test suite in pyodide. It was added by @eagleoflqj who also fixed related issues in the SymPy test suite and I think also some issues in pyodide to get it working. In the end we removed the CI job because it was flakey: often tests would pass but pyodide would hang afterwards until GitHub Actions timed out. 

Maybe that is fixed now and we could add the CI job again. All that is needed is a CI job that runs `pytest sympy` or `pytest --pyargs sympy`. Ideally this would use pytest-xdist for multiple cores but I don't know if that works with pyodide.

For testing in this repo I'm not sure that testing more than what is currently shown would give much benefit without testing the whole suite:
https://github.com/pyodide/pyodide/blob/67a4aa730797af6666d925a13c1fe8d631467f94/packages/sympy/test_sympy.py#L4-L11
SymPy is pure Python and I think most parts are unlikely to have any issue in pyodide. It would only be worth testing more if we had a clear idea of which kinds of things might be expected to fail in pyodide specifically.

What would be worth doing is testing SymPy 1.13 in combination with python-flint though because SymPy 1.13 makes use of python-flint internally when it is installed. It looks like python-flint is well tested here but it would be good to have at least a basic check that SymPy+python-flint does not crash.

oscarbenjamin:
The failed job looks unrelated to me but I might be misinterpreting the logs.

agriyakhetarpal:
> I can maintain SymPy's recipe. Would that need to be listed in the metadata somehow?

Superb, thanks for this! Yes, the metadata field you added is right. :)

> Maybe that is fixed now and we could add the CI job again. All that is needed is a CI job that runs `pytest sympy` or `pytest --pyargs sympy`. Ideally this would use pytest-xdist for multiple cores but I don't know if that works with pyodide.

> It would only be worth testing more if we had a clear idea of which kinds of things might be expected to fail in pyodide specifically.

I think it's worth trying again, since we might have gone through a few releases by then. @ryanking13 has previously mentioned that `pytest-xdist` should work, too, but no one has tried it yet for some time. Does the SymPy test suite have a `pytest -m "not slow"` option so that only a fraction of it can be tested so that the CI job doesn't take long? I would think even that is a reasonable improvement. SymPy is pure Python, yes, though any numerical functionalities are worth testing since they would be expected to fail or through weird errors with floating points (or any functionalities coming via dependencies). Other expected failures are the usual lack of threading and multiprocessing with Emscripten, and other limited modules from the stdlib, which I won't think SymPy would be using too much. I haven't been a big user of SymPy actually, but if we can get the test suite to pass and CI to not be flaky, it adds to the reasons towards keeping it. I haven't checked, but I'm assuming there is a decorator in the codebase for marking Pyodide-related test xfails already.

> The failed job looks unrelated to me but I might be misinterpreting the logs.

Yes, the job timed out – we have troubles with CircleCI from time to time due to this. I've just triggered a re-run for you.

packages/sympy/test_sympy.py
```diff
@@ -9,3 +9,15 @@ def test_sympy(selenium):
...
 
     assert c.subs({a: 3, b: 4}) == 5
+
+
+@run_in_pyodide(packages=["sympy", "python-flint"])
+def test_sympy_and_python_flint(selenium):
+    import sympy
+    from sympy.external.gmpy import GROUND_TYPES
+
+    assert GROUND_TYPES == "flint"
```
ryanking13, line 19: As python-flint is not a dependency of sympy, there is a chance that python-flint is not built when running this test (which will result in a failure). I would suggest updating the test code as follows:

```python
from conftest import package_is_built

def skip_if_not_installed(packages):
    for package in packages:
        if not package_is_built(package):
            pytest.skip(f"{package} not built")

def test_sympy_and_python_flint(selenium):
    skip_if_not_installed(["sympy", "python-flint"])
    
    @run_in_pyodide(packages=["sympy", "python-flint"])
    def run(selenium):
        import sympy
        from sympy.external.gmpy import GROUND_TYPES
    
        assert GROUND_TYPES == "flint"
    
        # Use python-flint for factorisation:
        x = sympy.symbols("x")
        assert (x**2 - 1).factor() == (x + 1) * (x - 1)

    run()
```

oscarbenjamin:
> Does the SymPy test suite have a `pytest -m "not slow"` option so that only a fraction of it can be tested so that the CI job doesn't take long?

Yes, the slow tests are already skipped by default:
https://github.com/sympy/sympy/blob/ce8940102bd778aa1fb53d5878c2d064e6a884dd/pyproject.toml#L1-L3
Even with that though it takes 20 minutes to run the not slow tests on 4 cores in CI. Under pyodide it would be slower so at least an hour I guess.

> SymPy is pure Python, yes, though any numerical functionalities are worth testing since they would be expected to fail or through weird errors with floating points (or any functionalities coming via dependencies).

Is floating point behaviour different in WASM compared to other platforms? Generally Python guarantees IEEE 754 floating point behaviour but libm can differ between e.g. MacOS vs Linux or Windows so e.g. `math.sin(0.1)` can give very slightly different values. You would not normally see this with SymPy though because it usually uses mpmath's platform independent implementation of these things.

> Other expected failures are the usual lack of threading and multiprocessing with Emscripten, and other limited modules from the stdlib, which I won't think SymPy would be using too much.

I think @eagleoflqj already worked through this in the SymPy test suite although it was some time ago so there might be new test failures. There already is a `skip_under_pyodide` decorator which is used for 4 tests (out of about 15000 total):
```console
$ git grep '@skip_under_pyodide'
sympy/core/tests/test_symbol.py:@skip_under_pyodide("Cannot create threads under pyodide.")
sympy/matrices/tests/test_matrices.py:@skip_under_pyodide("Cannot create threads under pyodide.")
sympy/matrices/tests/test_matrixbase.py:@skip_under_pyodide("Cannot create threads under pyodide.")
sympy/utilities/tests/test_misc.py:@skip_under_pyodide("Cannot create subprocess under pyodide.")
```

I wonder how pytest-xdist actually works in pyodide if there is no multiprocess support...

packages/sympy/test_sympy.py
oscarbenjamin: Good point. I've added that.

agriyakhetarpal:
> I wonder how pytest-xdist actually works in pyodide if there is no multiprocess support...

I read https://pytest-xdist.readthedocs.io/en/latest/how-it-works.html and I guess it doesn't use Python's `multiprocessing` module directly but instead relies on `execnet` – which does use `threading` a bit, though. I wonder if we could patch things, but others would know more how best to implement support.

oscarbenjamin:
I came up with this for a CI testing script:
```js
#!/usr/bin/env node

//
// pytest.js - Run tests with pytest under pyodide.
//
// Usage:
//
//   $ ./pytest.js [ARGS]
//
// Equivalent to:
//
//   $ pytest [ARGS]
//
// 1. Installs requirements-dev.txt and python-flint into node_modules/pyodide
// 2. Make the vcs checkout CWD in pyodide.
// 3. Runs pytest with the given arguments
//

const { argv } = require("node:process");
const { exit } = require("node:process");
var fs = require("fs")

requirements = fs.readFileSync("requirements-dev.txt", "utf-8")
    .split("\n")
    .filter(Boolean);
requirements.push("python-flint")

console.log("requirements:", requirements);

async function main(cmdargs) {

  await micropip.install(requirements);

  await pyodide.mountNodeFS("/sympy.git", ".");

  cmdargs_str = cmdargs.map((x) => `'${x}'`).join(", ");

  result = await pyodide.runPythonAsync(`

    import os
    import sys
    os.chdir('/sympy.git')
    sys.path.append('/sympy.git')

    import sympy
    import mpmath
    import pytest

    try:
        import flint
        flint_version = flint.__version__
    except ImportError:
        flint = None
        flint_version = 'N/A'

    from sympy.external.gmpy import GROUND_TYPES

    print(f"""

    sympy.__version__: {sympy.__version__}
    mpmath.__version__: {mpmath.__version__}
    flint.__version__: {flint_version}
    GROUND_TYPES: {GROUND_TYPES}

    Running pytest...

    """)

    # Actually run pytest:
    pytest.main([${cmdargs_str}])

`);
  return result;
}

main(argv.slice(2)).then((result) => {
    exit(result)
});
```
Feel free to suggest any improvements as I hardly know JS. This is basically just trying to recreate `python -m pytest [args]` with pyodide. I initially thought that I could probably do this with just one or two commands so maybe I'm missing something...

You can run it like:
```console
$ ./pytest.js --pyargs sympy.core --splits=10 --group=1
```
That uses pytest-split to divide the tests into groups so we can have several smaller CI jobs.

If you try pytest-xdist it fails due to subprocesses:
```console
$ ./pytest.js --pyargs sympy.core -n4
...
INTERNALERROR> OSError: [Errno 138] emscripten does not support processes.
```

agriyakhetarpal:
Nice, I didn't know about `pytest-split` – pretty neat! I don't know JS too well either, but this can be done in a better fashion nowadays using the following, which is, IMO, more "Pythonic" in many ways:

```bash

# building
pip install pyodide-build
pyodide build # python -m build works too, since the source is pure Python

# testing
pip install pyodide-build
pyodide venv .venv-pyodide
source .venv/bin/activate

# installation
pip install dist/*.whl
pip install pytest hypothesis # and any other dev dependencies/pytest plugins

# testing
pytest --pyargs sympy.core --splits=10 --group=1
```

and we can pass the tests into different runners with `${{ matrix.group }}` and similar methods of invocation. I'll be travelling for a bit over next week, so I'll have limited availability, but I'll try to spin something up and put together a PR to SymPy when I'm done testing.

oscarbenjamin:
Oh, that looks like a much nicer interface for using pyodide. I was thinking exactly that there should be something like that but didn't know about pyodide-build.

Unfortunately though it didn't work when I tried:
```console
$ pyodide venv .venv-pyodide
Creating Pyodide virtualenv at .venv-pyodide                                                         
╭──────────────────────────────── Traceback (most recent call last) ────────────────────────────────╮
│ /home/oscar/.pyenv/versions/3.12.0/envs/sympy-3.12.git/lib/python3.12/site-packages/pyodide_build │
│ /cli/venv.py:17 in main                                                                           │
│                                                                                                   │
│   14 ) -> None:                                                                                   │
│   15 │   """Create a Pyodide virtual environment"""                                               │
│   16 │   init_environment()                                                                       │
│ ❱ 17 │   venv.create_pyodide_venv(dest)                                                           │
│   18                                                                                              │
│                                                                                                   │
│ /home/oscar/.pyenv/versions/3.12.0/envs/sympy-3.12.git/lib/python3.12/site-packages/pyodide_build │
│ /out_of_tree/venv.py:265 in create_pyodide_venv                                                   │
│                                                                                                   │
│   262 │   │   sys.exit(1)                                                                         │
│   263 │                                                                                           │
│   264 │   interp_path = pyodide_dist_dir() / "python"                                             │
│ ❱ 265 │   session = session_via_cli(["--no-wheel", "-p", str(interp_path), str(dest)])            │
│   266 │   check_host_python_version(session)                                                      │
│   267 │                                                                                           │
│   268 │   try:                                                                                    │
│                                                                                                   │
│ /home/oscar/.pyenv/versions/3.12.0/envs/sympy-3.12.git/lib/python3.12/site-packages/virtualenv/ru │
│ n/__init__.py:49 in session_via_cli                                                               │
│                                                                                                   │
│    46 │   :return: the session object of the creation (its structure for now is experimental a    │
│    47 │   """  # noqa: D205                                                                       │
│    48 │   env = os.environ if env is None else env                                                │
│ ❱  49 │   parser, elements = build_parser(args, options, setup_logging, env)                      │
│    50 │   options = parser.parse_args(args)                                                       │
│    51 │   creator, seeder, activators = tuple(e.create(options) for e in elements)  # create t    │
│    52 │   return Session(                                                                         │
│                                                                                                   │
│ /home/oscar/.pyenv/versions/3.12.0/envs/sympy-3.12.git/lib/python3.12/site-packages/virtualenv/ru │
│ n/__init__.py:77 in build_parser                                                                  │
│                                                                                                   │
│    74 │   handle_extra_commands(options)                                                          │
│    75 │                                                                                           │
│    76 │   discover = get_discover(parser, args)                                                   │
│ ❱  77 │   parser._interpreter = interpreter = discover.interpreter  # noqa: SLF001                │
│    78 │   if interpreter is None:                                                                 │
│    79 │   │   msg = f"failed to find interpreter for {discover}"                                  │
│    80 │   │   raise RuntimeError(msg)                                                             │
│                                                                                                   │
│ /home/oscar/.pyenv/versions/3.12.0/envs/sympy-3.12.git/lib/python3.12/site-packages/virtualenv/di │
│ scovery/discover.py:41 in interpreter                                                             │
│                                                                                                   │
│   38 │   def interpreter(self):                                                                   │
│   39 │   │   """:return: the interpreter as returned by :meth:`run`, cached"""                    │
│   40 │   │   if self._has_run is False:                                                           │
│ ❱ 41 │   │   │   self._interpreter = self.run()                                                   │
│   42 │   │   │   self._has_run = True                                                             │
│   43 │   │   return self._interpreter                                                             │
│   44                                                                                              │
│                                                                                                   │
│ /home/oscar/.pyenv/versions/3.12.0/envs/sympy-3.12.git/lib/python3.12/site-packages/virtualenv/di │
│ scovery/builtin.py:46 in run                                                                      │
│                                                                                                   │
│    43 │                                                                                           │
│    44 │   def run(self):                                                                          │
│    45 │   │   for python_spec in self.python_spec:                                                │
│ ❱  46 │   │   │   result = get_interpreter(python_spec, self.try_first_with, self.app_data, se    │
│    47 │   │   │   if result is not None:                                                          │
│    48 │   │   │   │   return result                                                               │
│    49 │   │   return None                                                                         │
│                                                                                                   │
│ /home/oscar/.pyenv/versions/3.12.0/envs/sympy-3.12.git/lib/python3.12/site-packages/virtualenv/di │
│ scovery/builtin.py:61 in get_interpreter                                                          │
│                                                                                                   │
│    58 │   logging.info("find interpreter for spec %r", spec)                                      │
│    59 │   proposed_paths = set()                                                                  │
│    60 │   env = os.environ if env is None else env                                                │
│ ❱  61 │   for interpreter, impl_must_match in propose_interpreters(spec, try_first_with, app_d    │
│    62 │   │   key = interpreter.system_executable, impl_must_match                                │
│    63 │   │   if key in proposed_paths:                                                           │
│    64 │   │   │   continue                                                                        │
│                                                                                                   │
│ /home/oscar/.pyenv/versions/3.12.0/envs/sympy-3.12.git/lib/python3.12/site-packages/virtualenv/di │
│ scovery/builtin.py:93 in propose_interpreters                                                     │
│                                                                                                   │
│    90 │   │   │   if spec.is_abs:                                                                 │
│    91 │   │   │   │   raise                                                                       │
│    92 │   │   else:                                                                               │
│ ❱  93 │   │   │   yield PythonInfo.from_exe(os.path.abspath(spec.path), app_data, env=env), Tr    │
│    94 │   │   if spec.is_abs:                                                                     │
│    95 │   │   │   return                                                                          │
│    96 │   else:                                                                                   │
│                                                                                                   │
│ /home/oscar/.pyenv/versions/3.12.0/envs/sympy-3.12.git/lib/python3.12/site-packages/virtualenv/di │
│ scovery/py_info.py:385 in from_exe                                                                │
│                                                                                                   │
│   382 │   │   from virtualenv.discovery.cached_py_info import from_exe                            │
│   383 │   │                                                                                       │
│   384 │   │   env = os.environ if env is None else env                                            │
│ ❱ 385 │   │   proposed = from_exe(cls, app_data, exe, env=env, raise_on_error=raise_on_error,     │
│   386 │   │                                                                                       │
│   387 │   │   if isinstance(proposed, PythonInfo) and resolve_to_host:                            │
│   388 │   │   │   try:                                                                            │
│                                                                                                   │
│ /home/oscar/.pyenv/versions/3.12.0/envs/sympy-3.12.git/lib/python3.12/site-packages/virtualenv/di │
│ scovery/cached_py_info.py:33 in from_exe                                                          │
│                                                                                                   │
│    30 │   result = _get_from_cache(cls, app_data, exe, env, ignore_cache=ignore_cache)            │
│    31 │   if isinstance(result, Exception):                                                       │
│    32 │   │   if raise_on_error:                                                                  │
│ ❱  33 │   │   │   raise result                                                                    │
│    34 │   │   logging.info("%s", result)                                                          │
│    35 │   │   result = None                                                                       │
│    36 │   return result                                                                           │
╰───────────────────────────────────────────────────────────────────────────────────────────────────╯
RuntimeError: failed to query 
/media/oscar/EXTERNAL_SSD/current/active/sympy/.pyodide-xbuildenv-0.29.0/0.27.0a2/xbuildenv/pyodide-r
oot/dist/python with code 2 err: 
"/media/oscar/EXTERNAL_SSD/current/active/sympy/.pyodide-xbuildenv-0.29.0/0.27.0a2/xbuildenv/pyodide-
root/dist/python: can't open file 
'/home/oscar/.pyenv/versions/3.12.0/envs/sympy-3.12.git/lib/python3.12/site-packages/virtualenv/disco
very/py_info.py': [Errno 44] No such file or directory\n"
```

oscarbenjamin:
Is the current CI failure in this PR related to the changes here or is it something else (I find the CI here quite difficult to interpret):
```
=========================== short test summary info ============================
FAILED test_pyodide.py::module_not_found_note[chrome] - SystemError: error return without exception set
============================== 1 failed in 9.44s ===============================
```

hoodmane:
It's a known flake this looks good to merge. Thanks @oscarbenjamin!

oscarbenjamin:
Thanks!

oscarbenjamin:
How exactly do the package versions pull through to what is seen in the browser?

I see that pyodide 0.26.3 was released yesterday and if I go to the online [REPL](https://pyodide.org/en/stable/console.html) I get that version of pyodide but I don't see the SymPy update:
```python
Welcome to the Pyodide 0.26.3 terminal emulator 🐍
Python 3.12.1 (main, Oct 19 2024 08:33:26) on WebAssembly/Emscripten
>>> import sympy
>>> sympy.__version__
'1.12'
```
It looks like the change in this PR to update the SymPy version has not taken effect.

The reason I am asking about this is because I am trying to update the SymPy Live website to use SymPy 1.13.1 and python-flint 0.6.0:
https://live.sympy.org/
https://github.com/sympy/live

I have rebuilt the site in my fork but I am still getting SymPy 1.12:
https://oscarbenjamin.github.io/live/

agriyakhetarpal:
Hi @oscarbenjamin, Pyodide 0.26.3 is a patch release that has performed a few changes with the JSPI API, and it did not include this PR – we plan to incorporate most of the package upgrades with version 0.27. In the meantime, I'm working on some PRs to https://github.com/sympy/live as we speak (where it should then be possible to keep SymPy up-to-date by leaving in a custom SymPy wheel: https://jupyterlite.readthedocs.io/en/latest/howto/pyodide/wheels.html), and I've made progress on re-enabling the SymPy test suite running in Pyodide based on what we spoke earlier in this thread.

PR #5099, opened 2024-10-05
Add tree-sitter as a package

ericwb:
This change includes tree-sitter the packages that pyodide can install. The tree-sitter library has C extensions so it won't install with micropip from PyPI.

Tree sitter is one of a few requirements in order to run a SAST in the browser with PyScript.

https://pypi.org/project/tree-sitter
https://tree-sitter.github.io/tree-sitter
https://github.com/tree-sitter/py-tree-sitter

### Checklists

- [x] Add / update tests

### DIFF
--- a/docs/project/changelog.md
+++ b/docs/project/changelog.md
@@ -85,6 +85,7 @@ myst:
 - Added `rasterio` 1.13.10, `affine` 2.4.0 {pr}`4983`
 - Added `iminuit` 2.29.1 {pr}`4767`, {pr}`5072`
 - Added `arro3-core`, `arro3-io`, and `arro3-compute` 0.3.0, 0.4.0 {pr}`5020`, {pr}`5095`
+- Added `tree-sitter` 0.23.0 {pr}`5099`
 
 ## Version 0.26.2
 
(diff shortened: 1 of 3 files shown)


agriyakhetarpal (commented):
Thanks, @ericwb! Additionally, could you please add a few tests for the package in `packages/tree-sitter/test_tree-sitter.py` based on other projects' tests, and a CHANGELOG entry highlighting the new package?

ryanking13 (commented):
Thanks @ericwb. Could you please add some tests and add a changelog?

ericwb:
@agriyakhetarpal @ryanking13 I've added a test case, but you can't do much with tree-sitter on its own. You need to also have import a tree-sitter grammar for a particular language in order to be able to actual parsing. I was planning on adding some popular grammars as a follow on PR, such as https://pypi.org/project/tree-sitter-python/.

agriyakhetarpal (approved):
Thanks, @ericwb – following up with more packages and their tests sounds good. I guess `tree-sitter-python` would use `tree-sitter` as a dependency, so some tests there are nice to have.

PR #5102, opened 2024-10-07
Add tree-sitter-java, tree-sitter-go, tree-sitter-python

ericwb:
This is a follow-on to PR #5099 which added the dependency of tree-sitter to the packages.

This change adds tree-sitter-java, tree-sitter-go, tree-sitter-python packages and respective unit tests. Java, Go, and Python are popular languages, so its useful to add these grammars so a user can parse source code in those languages.

### Checklists

- [x] Add / update tests

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -41,3 +41,4 @@ pytest-pyodide
 tools/symlinks
 xbuildenv/
 .pyodide-xbuildenv*
+DS_Store
--- a/docs/project/changelog.md
+++ b/docs/project/changelog.md
@@ -87,6 +87,9 @@ myst:
 - Added `iminuit` 2.29.1 {pr}`4767`, {pr}`5072`
 - Added `arro3-core`, `arro3-io`, and `arro3-compute` 0.3.0, 0.4.0 {pr}`5020`, {pr}`5095`
 - Added `tree-sitter` 0.23.0 {pr}`5099`
+- Added `tree-sitter-go` 0.23.1 {pr}`5102`
+- Added `tree-sitter-java` 0.23.2 {pr}`5102`
+- Added `tree-sitter-python` 0.23.2 {pr}`5102`
 
 ## Version 0.26.2
 
--- a/packages/tree-sitter-go/meta.yaml
+++ b/packages/tree-sitter-go/meta.yaml
@@ -4,8 +4,10 @@ package:
   top-level:
     - tree_sitter_go
 source:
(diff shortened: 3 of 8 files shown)


hoodmane (approved):
Cool! Thanks @ericwb.

ericwb:
Looks like the build is failing due to not finding the dependent tree-sitter C header files. @hoodmane Any tips on how to resolve this. Sorry, I'm pretty new to pyodide's build environment.

hoodmane:
Doing a basic build only produces a wheel, it won't install any headers or libs. So if downstream packages need to compile/link against tree-sitter, there will need to be a build script to install those into `$WASM_LIBRARY_DIR`. I'll take a look into it tomorrow if you don't figure it out.

packages/tree-sitter-go/.DS_Store
agriyakhetarpal, line 1: Could you please remove this binary file, @ericwb?

packages/tree-sitter-java/.DS_Store
agriyakhetarpal, line 1: and this one

packages/tree-sitter-python/.DS_Store
agriyakhetarpal, line 1: and this one – it might be worth adding this to our `.gitignore` file, too.

agriyakhetarpal (changes requested):
Thanks, @ericwb!

hoodmane:
The problem is that the sdists don't contain the headers, so building from them won't work. I made some PRs to fix.
https://github.com/tree-sitter/tree-sitter-java/pull/188
https://github.com/tree-sitter/tree-sitter-go/pull/151
https://github.com/tree-sitter/tree-sitter-python/pull/285

For now, we should pull them in from github which does have the headers.

@henryiii tree-sitter does its builds with cibuildwheel, hypothetically it seems like it would be nice if cibuildwheel would build the wheel from the sdist so that if necessary files are missing from the sdist it would notice. But perhaps there's a good reason not to do that? Or it's supposed to work that way but it didn't function properly in this case?

agriyakhetarpal:
`cibuildwheel` can already build from an sdist just fine – one just needs to pass the path to the tarball in `package-dir:` if using the GitHub Action, or as a positional argument if using the CLI. I think more projects should adopt this approach. Henry's https://github.com/henryiii/check-sdist tool is quite useful, too.

PR #5112, opened 2024-10-11
Add test for rust stat64 abi mismatch

hoodmane:
This tests the fix for rust-lang/rust#131467 in pyodide/pyodide-build#40. The problem is that rust cannot correctly read file metadata because the ABI of fstat64 changed between the version of emscripten that the rust compiler built the standard library against and our version of the rust compiler.

By passing `-Zbuild-std` we rebuild the rust standard library with the correct abi. I also added a test that works only if we pass `-Zbuild-std`.


juntyr:
LGTM, thanks for adding the fix and the test so quickly!

hoodmane:
@ryanking13 any idea why it says `error: rustup is not installed at '/root/.cargo'` in CI?

ryanking13:
We don't install cargo under `/root`. We install it [under `/usr`](https://github.com/pyodide/pyodide/blob/3a6ca52d93153852ada67d3c260594b2f9132f4f/Dockerfile#L18) but I am not sure why it is complaining about the path. Let me investigate...

ryanking13:
When I set `CARGO_HOME=/usr`, that error message is gone, but buildingthe standard library still fails with the following error:

```
error: no matching package named `compiler_builtins` found                      
location searched: registry `crates-io`                                         
required by package `std v0.0.0                                                 
(/root/.rustup/toolchains/nightly-2024-01-29-x86_64-unknown-linux-gnu/lib/rustli
b/src/rust/library/std)`                                                        
💥 maturin failed                                                               
  Caused by: Failed to build a native library through cargo 
```

https://app.circleci.com/pipelines/github/ryanking13/pyodide/2531/workflows/c2488104-ed53-4db6-b8ee-e0b18d8ffaba/jobs/50690/parallel-runs/0/steps/0-104

hoodmane:
Probably related to this: https://github.com/rust-lang/compiler-builtins
I guess I'll try investigating more. Thanks @ryanking13!

juntyr:
Does the bug persist with `-Zbuild-std=std` or perhaps even `-Zbuild-std=std,compiler-builtins`

hoodmane:
Well it works fine when I run it locally in the docker image. I guess I should try "rerun with ssh" and try in the CI.

hoodmane:
Okay the problem is that `-Zbuild-std` depends on loading `compiler_builtins` from `crates.io` but `orjson` vendors its packages and has instructions not to look at crates.io:
```
$ cat orjson-3.10.1/.cargo/config.toml 
[source.crates-io]
replace-with = "vendored-sources"

[source.vendored-sources]
directory = "include/cargo"
```

hoodmane:
It's interesting that it lies in the error message though:
```console
$ cargo -Zbuild-std build --target wasm32-unknown-emscripten
error: no matching package named `compiler_builtins` found
location searched: registry `crates-io`
$ cargo search compiler_builtins
error: crates-io is replaced with non-remote-registry source dir /home/rchatham/Documents/programming/pyodide/packages/orjson/build/orjson-3.10.1/include/cargo;
include `--registry crates-io` to use crates.io
```

hoodmane:
I created a cargo bug report:
https://github.com/rust-lang/cargo/issues/14697

ryanking13:
> Okay the problem is that -Zbuild-std depends on loading compiler_builtins from crates.io but orjson vendors its packages and has instructions not to look at crates.io:

Huh, interesting. Maybe making it fallback to crates.io if the package does not exist would help resolving the issue but cargo [doesn't seem to support it](https://github.com/rust-lang/cargo/issues/3066)

hoodmane:
I guess `build-std` and `cargo vendor` absolutely don't work together.
https://github.com/rust-lang/wg-cargo-std-aware/issues/23
I think we'll need a way to drop `-Zbuild-std` for orjson.

PR #5124, opened 2024-10-18
Update altair to 5.4.1

joelostblom:
### Description

Update altair to 5.4.1, depends on https://github.com/pyodide/pyodide/pull/5121 since our dependencies changed in 5.4.

### Checklists

### DIFF
--- a/packages/altair/test_altair.py
+++ b/packages/altair/test_altair.py
@@ -9,11 +9,11 @@ def test_altair(selenium):
 
     data = alt.Data(
         values=[
-            {'a': 'A', 'b': 5},
-            {'a': 'B', 'b': 3},
-            {'a': 'C', 'b': 6},
-            {'a': 'D', 'b': 7},
-            {'a': 'E', 'b': 2}
+            {"a": "A", "b": 5},
+            {"a": "B", "b": 3},
+            {"a": "C", "b": 6},
+            {"a": "D", "b": 7},
+            {"a": "E", "b": 2},
         ]
     )
     c = alt.Chart(data).mark_bar().encode(x="a:N", y="b:Q").to_dict()


agriyakhetarpal (approved):
Thanks, @joelostblom!

packages/altair/meta.yaml
```diff
@@ -19,7 +19,5 @@ requirements:
     - typing-extensions
     - Jinja2
     - jsonschema
-    - numpy
-    - pandas
-    - toolz
     - packaging
+    - narwhals
```
ryanking13, line 23: I think this PR would need to wait until #5121 is merged.

mattijn: #5121 is merged. Can we re-run the failing CI?🙏

joelostblom: I retriggered by merging in the main branch after fixing a conflict, but it seems like the macos14 selenium test is still failing for what seems to be an unrelated reason:

image

hoodmane:
Don't worry about test-core-safari or test-core-node. But the actual test is failing, I think it's because the updated version of altair dropped the dependency on pandas?
```patch
--- a/packages/altair/test_altair.py
+++ b/packages/altair/test_altair.py
@@ -3,7 +3,7 @@ from pytest_pyodide import run_in_pyodide
 
 
 @pytest.mark.xfail_browsers(firefox="slow")
-@run_in_pyodide(packages=["altair"])
+@run_in_pyodide(packages=["altair", "pandas"])
 def test_altair(selenium):
     import altair as alt
     import pandas as pd
```

packages/altair/meta.yaml
ryanking13: It is an upstream selenium issue, so please ignore it.

joelostblom:
Thanks you @hoodmane , I included your suggested change in [628d64c](https://github.com/pyodide/pyodide/pull/5124/commits/628d64c73338b7904a32dac69bf39802431074e6)

packages/altair/test_altair.py
```diff
@@ -3,7 +3,7 @@
 
 
 @pytest.mark.xfail_browsers(firefox="slow")
-@run_in_pyodide(packages=["altair"])
+@run_in_pyodide(packages=["altair", "pandas"])
 def test_altair(selenium):
     import altair as alt
     import pandas as pd
```
agriyakhetarpal, line 9: This can fail in some cases (as it did in this instance) because `pandas` was removed as a runtime dependency above, so there is a chance `pandas` will not have been built by then (or will have been in the process of building in a different CI job). I am not sure if there is a reliable way to test both packages without one depending on another, do you have any thoughts, @ryanking13?

joelostblom: Thanks for that info! Technically, we don't need to use pandas and can create chart data via altair directly. I switched to using one of these approaches since it seems like an easier way to get this test to work.

agriyakhetarpal (approved):
Thanks, @joelostblom!

packages/altair/test_altair.py
ryanking13: You can use a similar approach [we use in duckdb test](https://github.com/pyodide/pyodide/blob/a5386c86b86135bda9c6f5d2d4f214726bacb7af/packages/duckdb/test_duckdb.py#L16); skipping the test until the dependency is built. But updating the test sounds better.

Issue #5120, opened 2024-10-17
Use of vite-plugin-static-copy NPM package with exclude patterns for Vite bundling example


castedo:
https://pyodide.org/en/latest/usage/working-with-bundlers.html
currently shows a Vite plugin that uses a hard-coded list of files that should be copied. Alternatively, the `vite-plugin-static-copy` can be used with a hard-coded list of patterns to exclude. Depending on how often the file list in the Pyodide distribute change, vs the stability of `vite-plugin-static-copy` this might or might not be a better approach.

An example of this alternative approach is here:
https://gitlab.com/castedo/pyodide-worker-example/-/blob/d317ef431c0e742339810b619c3626b726b275ec/vite-plugin-pyodide.js
to be used like this:
https://gitlab.com/castedo/pyodide-worker-example/-/blob/d317ef431c0e742339810b619c3626b726b275ec/vite.config.js

I'm happy to do a PR to the docs to show this use of `vite-plugin-static-copy` if you think it's a better approach. So far it's working well for me.

thangtuts:
how to load .whl files on your project?

castedo:
To load wheel files from a module worker, there are two approaches I can think of. One is to use the Pyodide CDN, the other is include the wheel in your own website.

## CDN approach

After calling loadPyodide without any parameters, you can then call [loadPackage](https://pyodide.org/en/stable/usage/api/js-api.html#pyodide.loadPackage) with absolute URLs to wheels anywhere, including the CDN that Pyodide suggests. For example, if you wanted to load the wheel for tomli you could use the URL
"https://cdn.jsdelivr.net/pyodide/v0.26.2/full/tomli_w-1.0.0-py3-none-any.whl"
as a name passed to loadPackage. 

## Include in website

For testing and development and possibily production, you can not use the CDN and instead include the wheels in the distribution for your own site. Here's an example where I have my own node package that stores some wheels in a directory named `pydist`:

https://gitlab.com/perm.pub/epijats-js/-/tree/main/pydist

Then in my Vue/Vite website application I have that node package as a dependency (I npm install it directly). I have viteStaticCopy copy all the wheels from the directory of that node package to my website distribution:

https://gitlab.com/perm.pub/snaptestsite/-/blob/a06685a4020a5ed86d6a6621a2e3f0f50bc9d1ca/vite.config.ts#L35

and then here is the code that runs in a web worker which grabs the wheels:

[...]

PR #5125, opened 2024-10-18
use vite-plugin-static-copy in doc example of Vite

castedo:
This is for enhancement #5120.

I've tested it by just putting the 3 files described on this doc page into one folder, following the instructions on the page, with the package versions mentioned on the page.

Assuming vite-plugin-static-copy is stable and works, this seems like a better approach.

@pauleveritt, you might have some feedback since you made an example from which this is inspired, thank you!

### DIFF
--- a/docs/usage/working-with-bundlers.md
+++ b/docs/usage/working-with-bundlers.md
@@ -31,7 +31,12 @@ import { viteStaticCopy } from "vite-plugin-static-copy";
 import { dirname, join } from "path";
 import { fileURLToPath } from "url";
 
-const PYODIDE_EXCLUDE = ["!**/package.json", "!**/*.{md,html}", "!**/*.d.ts"];
+const PYODIDE_EXCLUDE = [
+  "!**/*.{md,html}",
+  "!**/*.d.ts",
+  "!**/*.whl",
+  "!**/node_modules",
+];
 
 export function viteStaticCopyPyodide() {
   const pyodideDir = dirname(fileURLToPath(import.meta.resolve("pyodide")));
--- a/docs/usage/working-with-bundlers.md
+++ b/docs/usage/working-with-bundlers.md
@@ -10,50 +10,48 @@ project.
 ## Vite
 
 ```{note}
-The following instructions have been tested with Pyodide 0.26.0 and Vite 5.2.13.
+The following instructions have been tested with Pyodide 0.26.2, Vite 5.4.9, and
+vite-plugin-pyodide 2.0.0.
 ```
 
-First, install the Pyodide npm package:
+First, install the Pyodide and vite-plugin-pyodide npm packages:
 
 ```
-$ npm install pyodide
+$ npm install pyodide vite-plugin-static-copy
 ```
 
-Then, in your `vite.config.js` file, exclude Pyodide from [Vite's dependency
+Then, in your `vite.config.mjs` file, exclude Pyodide from [Vite's dependency
 pre-bundling][optimizedeps] by setting `optimizeDeps.exclude` and ensure that
 all Pyodide files will be available in `dist/assets` for production builds by
 using a Vite plugin:
 
 ```js
 import { defineConfig } from "vite";
-import { copyFile, mkdir } from "fs/promises";
+import { viteStaticCopy } from "vite-plugin-static-copy";
 import { dirname, join } from "path";
 import { fileURLToPath } from "url";
 
+const PYODIDE_EXCLUDE = [
+  "!**/package.json",
+  "!**/*.{md,html}",
+  "!**/*.d.ts",
+];
+
+export function viteStaticCopyPyodide() {
+  const pyodideDir = dirname(fileURLToPath(import.meta.resolve("pyodide")));
+  return viteStaticCopy({
+    targets: [
+      {
+        src: [join(pyodideDir, "*")].concat(PYODIDE_EXCLUDE),
+        dest: "assets",
+      },
+    ],
+  });
+}
+
 export default defineConfig({
   optimizeDeps: { exclude: ["pyodide"] },
-  plugins: [
-    {
-      name: "vite-plugin-pyodide",
-      generateBundle: async () => {
-        const assetsDir = "dist/assets";
-        await mkdir(assetsDir, { recursive: true });
-        const files = [
-          "pyodide-lock.json",
-          "pyodide.asm.js",
-          "pyodide.asm.wasm",
-          "python_stdlib.zip",
-        ];
-        const modulePath = fileURLToPath(import.meta.resolve("pyodide"));
-        for (const file of files) {
-          await copyFile(
-            join(dirname(modulePath), file),
-            join(assetsDir, file),
-          );
-        }
-      },
-    },
-  ],
+  plugins: [viteStaticCopyPyodide()],
 });
 ```
 
@@ -64,12 +62,12 @@ You can test your setup with this `index.html` file:
 <html lang="en">
   <head>
     <title>Vite + Pyodide</title>
-    <script type="module" src="/src/main.js"></script>
+    <script type="module" src="main.mjs"></script>
   </head>
 </html>
 ```
 
-And this `src/main.js` file:
+And this `src/main.mjs` file:
 
 ```js
 import { loadPyodide } from "pyodide";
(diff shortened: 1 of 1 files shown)


docs/usage/working-with-bundlers.md
```diff
@@ -10,50 +10,44 @@ project.
...
 
+const PYODIDE_EXCLUDE = ["!**/package.json", "!**/*.{md,html}", "!**/*.d.ts"];
+
+export function viteStaticCopyPyodide() {
+  const pyodideDir = dirname(fileURLToPath(import.meta.resolve("pyodide")));
+  return viteStaticCopy({
+    targets: [
+      {
+        src: [join(pyodideDir, "*")].concat(PYODIDE_EXCLUDE),
+        dest: "assets",
```
ryanking13, line 42: Will this include all the wheels in the directory? Doesn't it explode the final output size?

castedo: No, because the NPM package does not all the wheels.

`npm install pyodide` will install the NPM package and
```
pyodideDir = dirname(fileURLToPath(import.meta.resolve("pyodide")))
```
will find that NPM package.

Unless the user has done something really strange like make the full redistribution with all the wheel look like an NPM packege under some `node_modules` directory. I can't think of any other corner case off the top of my head.

castedo: The "NPM package" being the locally stored package downloaded from NPM.

castedo:
Oh, on another relative topic ... I should mention, I did NOT add the following pattern to `PYODIDE_EXCLUDE`:
```
"!**/pyodide.{js,mjs}*"
```
to this PR.

That means this PR code is letting the following files "slip through":
```
pyodide.js
pyodide.js.map
pyodide.mjs
pyodide.mjs.map
```
which is not what the current live example does. But I was thinking i might be useful for debugging to let those files get copied. But I'm not sure how much sense this makes. I fine adding exclusion `"!**/pyodide.{js,mjs}*"` to replicate the exact same result as the current live example.

docs/usage/working-with-bundlers.md
ryanking13: I see. That makes sense.

However, when you run Pyodide in local Node.js environment, it has a side effect of downloading packages into the `node-modules/pyodide` directory on demand as a side effect, which means that the content inside the node-modules can change. So I am a little worried that people might be confused about it.

castedo:
Ohhhhh, yeah, good point! Maybe I should add another exclusion pattern. Strange that I didn't suffer from this. Now that you point this out I'm surprised this hasn't happened to me already.

docs/usage/working-with-bundlers.md
castedo: Ohhhhh, yeah, good point! Maybe I should add another exclusion pattern. Strange that I didn't suffer from this. Now that you point this out I'm surprised this hasn't happened to me already.

castedo: I see, when I was testing this example code it was in a nice clean environment where nobody had started to party in the local `node-modules/pyodide` directory.

I also see now in a dirtier project that prettier has also been partying in this directory has created a `node_modules/pyodide/node_modules/.cache/prettier/` subdirectory.

Hmmm, maybe the inclusion list approach is safer than an exclusion list approach. :face_with_diagonal_mouth:

castedo: I'm thinking the trade-off depends on the respective risks of:

1) some unexpected and unpredictable files and folders appearing in the local `node-modules/pyodide` directory
VS
2) a Pyodide upgraded version changing the list of pyodide files that need to be redistributed in the build (and thus code copied from an older example longer working).

ryanking13: > a Pyodide upgraded version changing the list of pyodide files that need to be redistributed in the build (and thus code copied from an older example suddenly no longer working).

Yes, we do have this risk. So in [webpack-plugin](https://github.com/pyodide/pyodide-webpack-plugin/blob/main/lib/patterns.ts) we maintain a list of "core" files in the Pyodide distribution.

castedo: Oh, great data point to see what webpack-plugin is doing. It looks like it WAS taking the inclusion approach for version up until 0.24.0 and has since taken an exclusion approach.
Interestingly, webpack-plugin seems to always INCLUDE package.json whereas I was assuming it should be excluded.

castedo: I was starting to question the exclusion approach, but now that I see webpack-plugin ended up taking the exclusion approach, now I am back to thinking exclusion is the way to go. Let me know if you think we should discuss more whether exclusion approach is the right approach.

To wrap up this sub-thread here, we've realized two additional types of files that should be excluded:

1. all wheels "**/*.whl" and
2. this weird case I see locally of some files appearing under "node_modules" *within* the pyodide directory, so everything that might be there should be eliminated too.

I'll add another commit for these.

castedo:
@mneil Thank you for working on and sharing the webpack plugin for pyodide. I have been crafting something analogous for Vite, but just example code for documentation.

I'm not sure if I'm understanding the code correctly,  but it looks like you intentionally copy `package.json` from the node NPM package over to what gets redistributed on a website:

https://github.com/pyodide/pyodide-webpack-plugin/blob/6a32da3b4db68ba183b1e94c72fc96fc03465336/lib/patterns.ts#L40

I was intentionally EXCLUDING `package.json` for no particular reason other than I don't see what's the point. Is there a reason? Or in hindsight have you found no reason to redistribute it to a website. I'm new to NodeJS and have really only learned about ESM, so I'm assuming I'm not aware of some older CJS reason or something.

castedo:
> @castedo I haven't followed all the internal changes of pyodide lately but in the past, I believe, package.json was being fetched by the pyodide javascript. If it was not available over http pyodide would fail to load. That may not be the case anymore?

I have been seeing Pyodide load fine without `package.json`, but I also have not been targeting older browsers or CommonJS modules.

I went ahead and included it in my latest commit. My logic is there is very little downside to including small files. The client-side web application is not affected if it doesn't download it. The cost of redistributing small files is negligible. For debugging and diagnosis, some of these optional small files can be useful.

ryanking13:
>  I haven't followed all the internal changes of pyodide lately but in the past, I believe, package.json was being fetched by the pyodide javascript. If it was not available over http pyodide would fail to load. That may not be the case anymore?

I think we don't use package.json anymore.

castedo:
Any changes you want me to make @ryanking13 ? It looks good to me.

The only possible issues I can think of are:

1) Is a dependency on plugin `vite-plugin-static-copy` a bad idea for an example? The plugin seems well-adopted to me, so I'm inclined to think the dependency is fine.
2) Is an inclusion approach preferable to an exclusion approach? I have a few reasons that make me think exclusion approach is better.
3) Should *small* files that we suspect are not needed, but nonetheless might be useful for testing and debugging, be in the exclusion list so they are kept out of the copy? The current commit and my thoughts are it's better to let these small files pass through, but I don't have strong reasons for this. I'm happy to have them in the exclude list.

ryanking13:
@castedo 

Thanks. I think it is okay to merge as-is. We can improve documentation later when we encounter other issues :)

lien-dkseo:
not working on windows

```
error during build:
[vite-plugin-static-copy:build] No file was found to copy on D:\...\node_modules\pyodide\*,!**/*.{md,html},!**/*.d.ts,!**/*.whl,!**/node_modules src.
    at collectCopyTargets (file:///.../node_modules/vite-plugin-static-copy/dist/index.js:512:13)
```

lien-dkseo:
> not working on windows
> 
> ```
> error during build:
> [vite-plugin-static-copy:build] No file was found to copy on D:\...\node_modules\pyodide\*,!**/*.{md,html},!**/*.d.ts,!**/*.whl,!**/node_modules src.
>     at collectCopyTargets (file:///.../node_modules/vite-plugin-static-copy/dist/index.js:512:13)
> ```

According to https://www.npmjs.com/package/vite-plugin-static-copy,
In Windows, the code below is required.
```ts
import { normalizePath } from 'vite'

return viteStaticCopy({
    targets: [
      {
        src: [normalizePath(join(pyodideDir, "*"))].concat(PYODIDE_EXCLUDE),
        dest: "assets/pyodide",
      },
    ],
  });
```

and..

How about changing `dest: "assets"` to `dest: "assets/pyodide"`?

agriyakhetarpal:
Thanks for investigating @lien-dkseo, I think adding a short note for Windows users wluld be great – please feel free to open a PR!

PR #5129, opened 2024-10-22
Fix support for JSPI when there is no type reflection

hoodmane:
The new JSPI does not require wasm type reflection, so it's possible that JS runtimes will have JSPI support but no type reflection, for instance if they are enrolled in the JSPI origin trial or have the JSPI flag on but not the type reflection flag.

If type reflection is missing, we fall back to attempting to instantiate a dummy module that imports a function with a given signature. This will raise a LinkerError if the signature is wrong. By trying all four possibilities, we learn the actual signature of the function.

### Checklists

- [x] Add / update tests


src/core/stack_switching/calculate_wasm_func_nargs_fallback.mjs
```diff
@@ -0,0 +1,52 @@
...
+
+function getCheckerModule(n) {
+  if (checkerModules[n] === undefined) {
+    checkerModules[n] = makeCheckerModule(n);
+  }
+  return checkerModules[n];
+}
+
+export function calculateWasmFuncNargsFallback(functionPtr) {
+  for (let n = 0; n < 4; n++) {
```
ryanking13, line 43: How is the number 4 selected?

src/core/stack_switching/calculate_wasm_func_nargs_fallback.mjs
```diff
@@ -0,0 +1,52 @@
...
+    try {
+      new WebAssembly.Instance(getCheckerModule(n), {
+        e: { f: wasmTable.get(functionPtr) },
+      });
+      return n;
+    } catch (e) {}
```
ryanking13, line 49: What happens on error? Is it okay to ignore it?

hoodmane: We expect it's a `WebAssembly.LinkError`, but I should check this and rethrow other errors in case something surprising happens on this code path.

src/core/stack_switching/calculate_wasm_func_nargs_fallback.mjs
hoodmane: Yeah this also needs a comment.

It matches up with the cases here:
https://github.com/pyodide/pyodide/blob/main/cpython/patches/0004-Make-Emscripten-trampolines-work-with-JSPI.patch?plain=1#L290-L309

Basically, we're trying to adapt for all bad fpcast cases. The most common case where we need the trampoline is that the function is a `METH_NOARGS` function which either takes 2 arguments (the correct number) or 1 argument (dropping the `NULL` argument) or 0 arguments (dropping both `NULL` and the `module` argument). Or people could do a `METH_VARARGS | METH_KWDS` which should have signature `PyObject* mod, PyObject* args, PyObject* kwds` and ignore either the kwds, both the kwds and the args, or everything, so 3, 2, 1, or 0 arguments.

We're importing the function with various signatures `(n i32's) => i32` to find out which case it's in. If it's not one of 0, 1, 2, or 3 i32 arguments, then we're in some other weird case. For instance it might return a `f64` or something. The type reflection code wouldn't notice this and would attempt the call_indirect raising a fatal indirect call signature mismatch. This code would notice and raise a nonfatal `SystemError`.

src/core/stack_switching/calculate_wasm_func_nargs_fallback.mjs
hoodmane: Okay I moved everything but the actual call to `new WebAssembly.Instance` out of the try block and added a check that the error is a `LinkError` and rethrow other errors. This should make it less likely to swallow a mistake.

src/core/stack_switching/calculate_wasm_func_nargs_fallback.mjs
ryanking13: Got it. Thank for the explanation. Could you add the comment to the code as well?

ryanking13:
Seems like rustup is missing again in the docker image for some reason

hoodmane:
Huh I checked for rust when I built it and it seemed correctly installed.

PR #5132, opened 2024-10-24
Constrain meson < 1.6 to fix the scipy build

### LINKED ISSUE #5131: Regression in scipy build
## 🐛 Bug

It seems like scipy build is partially broken since Oct 19~21.

For some reason, `_unuran` module is built as a static library, breaking [scipy tests](https://app.circleci.com/pipelines/github/pyodide/pyodide/7949/workflows/28cb75b6-27b8-43e2-930d-ca6afec1ac5e/jobs/115204). This often happens when `SIDE_MODULE` flag is not correctly passed when linking the shared library, but it is mysterious why is happens.

```
[735/1329] Linking target scipy/stats/_unuran/unuran_wrapper.cpython-312-wasm32-emscripten.so
emcc: warning: ignoring unsupported linker flag: `--version-script=/root/repo/packages/scipy/build/scipy-1.14.1/scipy/_build_utils/link-version-pyinit.map` [-Wlinkflags]
```

Since there was no change in Pyodide side, I am guessing it has something to do with build dependencies of Scipy.

I've noticed that the latest successful scipy build used meson `1.5.2` while the one with error used `1.6.0` (released 2 days ago) so it might be related somehow.

juntyr:
Fixes #5131


ryanking13 (approved):
Thanks! It looks like the build is passing, but I reran the CI because of the CI timeout.

PR #5146, opened 2024-10-30
Add `rateslib` to packages

NickAltmann:
### Description

Adding meta for the rateslib package, a fixed income pricing library, see https://rateslib.readthedocs.io/en/latest/.

While the first release, 1.1, was pure python, subsequent releases have a rust extension, hence the need for a package here.

### Checklists

- [x] Add / update tests

### DIFF
--- a/packages/rateslib/meta.yaml
+++ b/packages/rateslib/meta.yaml
@@ -2,10 +2,9 @@ package:
   name: rateslib
   version: 1.5.0
   top-level:
-  - rateslib
+    - rateslib
 source:
-  url: 
-    https://files.pythonhosted.org/packages/02/d7/7695bb1b6a240ecfae77804d3fc242e30582fd83973f76dda03b23d921a6/rateslib-1.5.0.tar.gz
+  url: https://files.pythonhosted.org/packages/02/d7/7695bb1b6a240ecfae77804d3fc242e30582fd83973f76dda03b23d921a6/rateslib-1.5.0.tar.gz
   sha256: 7c7af2707f20974a70403689f84905757a4ed084e4c92229388d2fc30b331b04
 requirements:
   host:
@@ -19,11 +18,11 @@ requirements:
   executable:
     - rustup
 about:
-  home: https://github.com/attack68/rateslib 
+  home: https://github.com/attack68/rateslib
   PyPI: https://pypi.org/project/rateslib
   summary: A fixed income library for trading interest rates
   license: Creative Commons Attribution, Non-Commercial, No-Derivatives 4.0 International Licence
 extra:
   recipe-maintainers:
-  - attack68
-  - NickAltmann
+    - attack68
+    - NickAltmann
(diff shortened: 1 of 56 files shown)


packages/rateslib/meta.yaml
```diff
@@ -0,0 +1,28 @@
...
+  top-level:
+    - rateslib
+source:
+  url: https://files.pythonhosted.org/packages/02/d7/7695bb1b6a240ecfae77804d3fc242e30582fd83973f76dda03b23d921a6/rateslib-1.5.0.tar.gz
+  sha256: 7c7af2707f20974a70403689f84905757a4ed084e4c92229388d2fc30b331b04
+requirements:
+  host:
+    - numpy
+    - pandas
+    - matplotlib
```
ryanking13, line 13: You don't need to set these as a host dependency.

packages/rateslib/meta.yaml
```diff
@@ -0,0 +1,28 @@
...
+  executable:
+    - rustup
+about:
+  home: https://github.com/attack68/rateslib
+  PyPI: https://pypi.org/project/rateslib
+  summary: A fixed income library for trading interest rates
+  license: Creative Commons Attribution, Non-Commercial, No-Derivatives 4.0 International Licence
+extra:
+  recipe-maintainers:
+    - attack68
```
ryanking13, line 27: Could you please confirm that @attack68 would want to maintain the recipe?

ryanking13 (commented):
Thanks. Could you sync your PR with the main branch? We've just deployed a fix for CI timeout.

packages/rateslib/meta.yaml
NickAltmann: Ok, attack68 is the developer of the rateslib library itself, but for now I've removed him as maintainer for this recipe.  I'll provide him with instructions, and should he wish to perform an update later he can add himself at that point.

packages/rateslib/meta.yaml
```diff
@@ -0,0 +1,23 @@
...
+    - numpy
+    - pandas
+    - matplotlib
+  executable:
+    - rustup
+about:
+  home: https://github.com/attack68/rateslib
+  PyPI: https://pypi.org/project/rateslib
+  summary: A fixed income library for trading interest rates
+  license: Creative Commons Attribution, Non-Commercial, No-Derivatives 4.0 International Licence
```
agriyakhetarpal, line 20: Could we standardise this better by using the [SPDX short-form identifier](https://spdx.org/licenses/CC-BY-NC-ND-4.0.html) for this license?
```suggestion
  license: CC-BY-NC-ND-4.0
```

agriyakhetarpal (commented):
Thanks for the contribution, @NickAltmann!

agriyakhetarpal:
The failure can be adjusted either by shortening the test or by increasing the timeout value. I guess the test is already minimal, so the latter is better.

hoodmane:
The suggestion more specifically is to annotate the test with something like `@pytest.mark.driver_timeout(60)` which would raise the timeout to 60 seconds.

agriyakhetarpal (approved):
Looks like it's working now, thanks, @NickAltmann!

PR #5148, opened 2024-10-31
CI Split library build job

ryanking13:
Splits library build job as a workaround for the timeout. Builds static libraries first then build shared libraries.


hoodmane:
Thanks!

hoodmane (approved):

ryanking13:
Thanks for the review!

PR #5151, opened 2024-11-03
Upgrade `pydantic` and `pydantic_core`

riyavsinha:
### Description

- Standard upgrade for pydantic
- For `pydantic_core`, I upgraded the emscripten wheel in their repo (https://github.com/pydantic/pydantic-core/pull/1507), so rather than rebuilding it in pyodide, I think we can just use that directly? If I'm mistaken please let me know!

### Checklists

- [x] Add / update tests

### DIFF
--- a/docs/project/changelog.md
+++ b/docs/project/changelog.md
@@ -64,6 +64,8 @@ myst:
 
 ### Packages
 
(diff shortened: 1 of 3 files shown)


hoodmane:
Thanks!

PR #5162, opened 2024-11-06
Pin types-setuptools to fix mypy build

hoodmane:
Doesn't seem to work without:
https://github.com/pyodide/pyodide-build/pull/49

@agriyakhetarpal @ryanking13

### DIFF
--- a/Makefile.envs
+++ b/Makefile.envs
@@ -11,8 +11,8 @@ export PYTHON_ARCHIVE_SHA256=d01ec6a33bc10009b09c17da95cc2759af5a580a7316b3a446e
 # to your fork to test the changes are working as expected.
 
 # v0.27.3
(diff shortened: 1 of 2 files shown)


agriyakhetarpal:
Thanks, @hoodmane! I've proposed a fix earlier today upstream so that we can unpin this later: https://github.com/python/typeshed/pull/12958

agriyakhetarpal (approved):

hoodmane:
Can you also review the pyodide-build PR?

agriyakhetarpal:
Yes, did that already – I approved both that PR and this one

ryanking13:
A new version of types-setuptools is released so I think we don't need this PR

hoodmane:
Great

PR #5163, opened 2024-11-06
Bump rebound/reboundx to version 4.4.3/4.3.0

hannorein:
### Description

Bumps the REBOUND version from 3.24.2 to 4.4.3. Also bumps REBOUNDx to 4.3.0 Fixes the compile time error mentioned in #4925.

### Checklists

### DIFF
--- a/packages/rebound/patches/0001-fix-install_name.patch
+++ b/packages/rebound/patches/0001-fix-install_name.patch
@@ -1,15 +1,27 @@
-Index: rebound-3.24.2/setup.py
-===================================================================
---- rebound-3.24.2.orig/setup.py
-+++ rebound-3.24.2/setup.py
-@@ -23,8 +23,8 @@ extra_link_args=[]
+From 0a6cc79e6814d7e2dbc637541cad86cebc47f855 Mon Sep 17 00:00:00 2001
+From: Hanno Rein <hanno@hanno-rein.de>
+Date: Wed, 6 Nov 2024 17:37:22 -0500
+Subject: [PATCH 3/3] install_name
+
+---
+ setup.py | 4 ++--
+ 1 file changed, 2 insertions(+), 2 deletions(-)
+
+diff --git a/setup.py b/setup.py
+index 6443cb27..965b2fc9 100644
+--- a/setup.py
++++ b/setup.py
+@@ -23,8 +23,8 @@ except:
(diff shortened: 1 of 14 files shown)


packages/rebound/patches/0001-fix-install_name.patch
agriyakhetarpal, line 1: Could this patch be dropped – I noticed that the build worked fine without it?

packages/rebound/patches/0002-fix-output.patch
agriyakhetarpal, line 1: If the first patch is dropped, could you rename the file with "0001" instead?

agriyakhetarpal (approved):
Thanks, @hannorein! Just a minor nitpick-y suggestion.

hannorein:
It gets rid of a warning message so I thought I'll keep it for now. I'll merge these patches into the upstream in a future version of REBOUND so that we can get rid of all them.

agriyakhetarpal:
There's another error: `Error: Dynamic linking error: cannot resolve symbol emscripten_sleep` :(

hannorein:
I can fix that. But how come I didn't get those errors? And how can I make sure I fix them all?

agriyakhetarpal:
> I can fix that. But how come I didn't get those errors? And how can I make sure I fix them all?

You could run the tests locally with `pytest -svra packages/rebound/test_rebound.py` – but I usually find the CI logs to be better  sometimes, because from a local run I wasn't able to see that the timeout was due to a Pyodide fatal error (this one, in this case).

I'd recommend an out-of-tree build in your CI, too, if that helps and if you're interested: https://pyodide.org/en/stable/development/building-and-testing-packages.html

hannorein:
Thanks. I'm still a bit confused on how to do this properly. 

Right now the tests pass locally. If the package compiles and runs fine locally, then why does it not compile in the CI? 

Are the two failing CI tests `main / test-core (macos-14, selenium, safari)` and `ci/circleci: test-core-node-nodylink` failing because of something in the rebound/reboundx packages? Or is that unrelated?

hoodmane:
Those are unrelated, don't worry about them.

hoodmane:
CI looks good to me.

hoodmane:
Thanks @hannorein!

PR #5168, opened 2024-11-07
Add missing Pyodide v0.26.3 cross-build environment metadata

agriyakhetarpal:
### Description

https://github.com/pyodide/pyodide/actions/runs/11417310347 had failed, so I'm updating it with this PR.

I think we should make the workflow open an issue to remind us to do this manually if it fails (but also fix why it failed in the first place).

Closes pyodide/pyodide-build#55

### Checklists

N/A

### DIFF
--- a/pyodide-cross-build-environments.json
+++ b/pyodide-cross-build-environments.json
@@ -8,6 +8,14 @@
       "emscripten_version": "3.1.58",
       "min_pyodide_build_version": "0.26.0"
     },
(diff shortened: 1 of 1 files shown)


agriyakhetarpal:
This change doesn't have to do with anything related to the in-tree builds, so I'll merge this for now to use it for some downstream PRs. Additionally, I'll work on the task I mentioned in the issue description :D

Issue #4787, opened 2024-05-23
`crc32c`


agriyakhetarpal:
## 🐍 Package Request

- Package Name and Version : `crc32c`, version 2.4
- Package URL : https://pypi.org/project/crc32c/, https://github.com/ICRAR/crc32c
- Package Dependencies that needs to be resolved first: [none](https://github.com/ICRAR/crc32c/blob/master/setup.py) – this is a package written in C with Python bindings.

## 📝 Additional context

xref: https://github.com/zarr-developers/zarr-python/pull/1903, i.e., this is a required dependency of the upcoming version 3 for Zarr at the time of writing.

agriyakhetarpal:
I would be happy to try this out!

agriyakhetarpal:
I see that SSE2 instructions are used, they will probably need to be disabled in some way...

agriyakhetarpal:
With the `pyodide build-recipes crc32c --install` command:

`[8/13] (thread 8) built crc32c in 24s` 🎉

I will be adding a few tests from their test suite now to confirm that the basic functionality for the package works. Their license is the LGPL-2.1, though. Would that be all right or should I write a faction of the tests entirely by myself?

hoodmane:
I don't think it matters because we aren't mixing it into our logic or running it at runtime. But we don't need all the tests either.

PR #5169, opened 2024-11-07
Bump `crc32c` to version 2.7.1

agriyakhetarpal:
### Description

This PR bumps [`crc32c`](https://pypi.org/project/crc32c/) to the latest available version, 2.7.1. I'm updating this in accordance with https://github.com/zarr-developers/numcodecs/pull/529, where we previously added this package in-tree (#4787, #4789), but they are requiring a later version now. The package has no dependencies that need to be updated or otherwise.

### Checklists

- [ ] Add / update tests


agriyakhetarpal:
I also planned to add https://github.com/LLNL/zfp (`zfpy`) but they don't have a source distribution available on PyPI yet: https://pypi.org/project/zfpy/#files. It is a dependency for `numcodecs`, albeit not a required one (but still good enough to test for). Should I create a dummy sdist for it, or perhaps we can experiment with the GitHub tarball until I request that they publish one for the next release? Adding the package would also block #4925, though, because they don't support NumPy v2 yet (https://github.com/LLNL/zfp/issues/210), but there's a chance it might just work. I'll comment on that issue about both of these aspects. In the meantime, bumping just `crc32c` would here should suffice for `numcodecs`, since it seems to be required for the test suite (the tests that use it are not skipped), even though it's listed as an optional dependency.

agriyakhetarpal:
I also planned to add https://github.com/LLNL/zfp (`zfpy`) but they don't have a source distribution available on PyPI yet: https://pypi.org/project/zfpy/#files. It is a dependency for `numcodecs`, albeit not a required one (but still good enough to test for). Should I create a dummy sdist for it, or perhaps we can experiment with the GitHub tarball until I request that they publish one for the next release? Adding the package would also block #4925, though, because they don't support NumPy v2 yet (https://github.com/LLNL/zfp/issues/210), but there's a chance it might just work. I'll comment on that issue about both of these aspects. In the meantime, bumping just `crc32c` would here should suffice for `numcodecs`, since it seems to be required for the test suite (the tests that use it are not skipped if it isn't available), even though it's listed as an optional dependency.

ryanking13 (approved):
Thanks!

agriyakhetarpal:
Thanks for the review!

PR #5170, opened 2024-11-07
Catch KeyError in importhook

hoodmane:
- [x] Add / update tests


ryanking13 (commented):
Could you please give a little bit more context around this change?

hoodmane:
I got a `KeyError` error here when doing test collection for matplotlib. I'll try to make a repro when I have a moment.

hoodmane:
But it's not clear to me why parent *should* be in `sys.modules` here.

ryanking13 (approved):
> But it's not clear to me why parent should be in sys.modules here.

Hmm, I see. That code was added too long ago, so I am not sure about that, too. The change looks okay anyways.

hoodmane:
A reproduction for this looks like:
```py
import sys

def _import_module_using_spec(module_name):
    for meta_importer in sys.meta_path:
        spec = meta_importer.find_spec(module_name, path=[])

 _import_module_using_spec("a.b")
```
This is a fragment of what pytest does here:
https://github.com/pytest-dev/pytest/blob/main/src/_pytest/pathlib.py#L693-L698

These calls will never come from importlib, when processing `a.b` importlib will always look at `a` first and then `a.b`.

PR #5172, opened 2024-11-07
Add `zfp` and `zfpy`

### LINKED ISSUE #5171: `zfpy`
## 🐍 Package Request

- Package Name and Version : `zfpy`, version 1.0.1
- Package URL : https://github.com/LLNL/zfp/, https://computing.llnl.gov/projects/zfp
- Package Dependencies that needs to be resolved first: none

## Checklists

- [x] I have tried to install the package using `micropip.install(...)` and encountered the issue.
- [x] I have read the [documentation](https://pyodide.org/en/stable/development/building-and-testing-packages.html) and tried building the package myself.

agriyakhetarpal:
### Description

This PR adds two packages: the `zfp` compression library as a static library and its Python bindings via `zfpy`. The source can be accessed at https://github.com/LLNL/zfp. `zfpy` is currently an optional dependency for `numcodecs`, so, while it isn't strictly required, it shall help us test the functionality better in zarr-developers/numcodecs#529 in `numcodecs`'s test suite.

Closes #5171, and in continuation of #5169

### Checklists

- [x] Add / update tests


agriyakhetarpal:
I'm able to get it to compile, but it builds a Linux ELF binary instead of a WASM binary so I'm probably doing this incorrectly. The packaging infrastructue upstream is not very modern either, because the package is written in C/C++ with Fortran and Python interfaces. There's no sdist available on PyPI. It also tries to find Cython and fails, but other Cython-based projects don't need it as a dependency.

I'll most likely try splitting it into two recipes: `zfp`, and then `zfpy` based on that.

.pre-commit-config.yaml
```diff
@@ -82,7 +82,7 @@ repos:
         name: mypy-tests
         args: [--ignore-missing-imports]
         files: ^(packages/|docs|/conftest.py|src/tests)
-        exclude: (^packages/.*/setup.py|/src|^packages/aiohttp/aiohttp_patch.py$)
+        exclude: (^packages/.*/setup.py|/src|^packages/aiohttp/aiohttp_patch.py$|^packages/zfpy/test_zfpy.py$)
```
agriyakhetarpal, line 85: I've been compelled to add this because I adapted most of https://github.com/LLNL/zfp/blob/1.0.1/tests/python/test_numpy.py without much changes and the type annotations there are non-existent. As a result, it's quite messy out here without excluding the file.

packages/zfp/meta.yaml
```diff
@@ -0,0 +1,43 @@
...
+  sha256: 4984db6a55bc919831966dd17ba5e47ca7ac58668f4fd278ebd98cd2200da66f
+
+build:
+  # cflags: -fexceptions
+  # cxxflags: -fexceptions
+  # ldflags: -fexceptions
+  # Let's build a static target for now
+  type: static_library
+  # -DPYTHON_NUMPY_INCLUDE_DIRS=$NUMPY_INCLUDE_DIR \
+  # -DBIT_STREAM_WORD_TYPE=uint64               \
```
agriyakhetarpal, line 17:
```suggestion
```

packages/zfp/meta.yaml
```diff
@@ -0,0 +1,43 @@
...
+source:
+  # There's no sdist for zfpy on PyPI, so we get the tarball from the GitHub release instead
+  url: https://github.com/LLNL/zfp/archive/refs/tags/1.0.1.tar.gz
+  sha256: 4984db6a55bc919831966dd17ba5e47ca7ac58668f4fd278ebd98cd2200da66f
+
+build:
+  # cflags: -fexceptions
+  # cxxflags: -fexceptions
+  # ldflags: -fexceptions
+  # Let's build a static target for now
```
agriyakhetarpal, line 14:
```suggestion
```

packages/zfp/meta.yaml
```diff
@@ -0,0 +1,43 @@
...
+
+source:
+  # There's no sdist for zfpy on PyPI, so we get the tarball from the GitHub release instead
+  url: https://github.com/LLNL/zfp/archive/refs/tags/1.0.1.tar.gz
+  sha256: 4984db6a55bc919831966dd17ba5e47ca7ac58668f4fd278ebd98cd2200da66f
+
+build:
+  # cflags: -fexceptions
+  # cxxflags: -fexceptions
+  # ldflags: -fexceptions
```
agriyakhetarpal, line 13: Is there a way to add this flag for static libraries, otherwise?

packages/zfp/meta.yaml
```diff
@@ -0,0 +1,43 @@
...
+    emmake make -j ${PYODIDE_JOBS:-3} install
+
+about:
+  home: https://zfp.llnl.gov/
+  PyPI: https://pypi.org/project/zfpy/
+  summary: zfp compression in Python
+  license: BSD-3-Clause
+# extra:
+#   recipe-maintainers:
+#     - PUT_YOUR_GITHUB_USERNAME_HERE
```
agriyakhetarpal, line 43: I don't know if I can be a maintainer for this recipe, so I'll wait if someone has any interest upstream.
```suggestion

```

packages/zfpy/meta.yaml
```diff
@@ -0,0 +1,40 @@
...
+    -L$(WASM_LIBRARY_DIR)/lib
+
+about:
+  home: https://zfp.llnl.gov/
+  PyPI: https://pypi.org/project/zfpy/
+  summary: zfp compression in Python
+  license: BSD-3-Clause
+# extra:
+#   recipe-maintainers:
+#     - PUT_YOUR_GITHUB_USERNAME_HERE
```
agriyakhetarpal, line 40:
```suggestion
```

ryanking13:
> Another question: should I keep the name as zfp or rename it to libzfp?

packages/zfp/meta.yaml
```diff
@@ -0,0 +1,37 @@
+package:
+  name: zfp
+  version: 1.0.1
+
```
ryanking13, line 4: Let's add tags for our CI

```suggestion
  version: 1.0.1
  tag:
    - library
    - static_library
```

packages/zfp/meta.yaml
```diff
@@ -0,0 +1,37 @@
...
+build:
+  # cflags: -fexceptions
+  # cxxflags: -fexceptions
+  # ldflags: -fexceptions
+  type: static_library
+  exports: requested
+  script: |
+    mkdir -p pyodide_build
+    cd pyodide_build
```
ryanking13, line 18: The name pyodide_build is a bit confusing as it sounds like pyodide-build library. So how about just build.

```suggestion
    mkdir -p build
    cd build
```

packages/zfp/meta.yaml
```diff
@@ -0,0 +1,37 @@
...
+    emcmake cmake ${CMAKE_ARGS}                   \
+      -DBUILD_CFP=ON                              \
+      -DBUILD_UTILITIES=OFF                       \
+      -DBUILD_TESTING=ON                         \
```
ryanking13, line 22: Do we need to build test?

packages/zfpy/meta.yaml
```diff
@@ -0,0 +1,37 @@
+package:
+  name: zfpy
+  version: 1.0.1
+  top-level:
+    - zfpy
+
+requirements:
```
ryanking13, line 7: Let's rearrange the order of keys to package-source-build-requirements

packages/zfpy/patches/Modernise-packaging.patch
```diff
@@ -0,0 +1,69 @@
...
+ create mode 100644 pyproject.toml
+
+diff --git a/pyproject.toml b/pyproject.toml
+new file mode 100644
+index 0000000..88adcc8
+--- /dev/null
++++ b/pyproject.toml
+@@ -0,0 +1,13 @@
++[build-system]
++requires = ["setuptools", "cython", "numpy"]
```
ryanking13, line 27: 👍

ryanking13 (approved):
Thanks!

packages/zfp/meta.yaml
agriyakhetarpal: Thanks for the catch, no need to – I was just testing something :)

packages/zfp/meta.yaml
agriyakhetarpal: Answering my own question: there's most likely no need to, right now

packages/zfp/meta.yaml
agriyakhetarpal: Addressed in 53286ff13a4316e67a99e6c3a91303b89c34f3ad

packages/zfp/meta.yaml
agriyakhetarpal: Makes sense to me. Addressed in 998bf8f600f548e78dc38203213ac6150056a0a6

packages/zfp/meta.yaml
agriyakhetarpal: Addressed in 3a99b33e3a2bdb448540e70400a14dd0e53587f4.

packages/zfpy/meta.yaml
agriyakhetarpal: Addressed in 2253aa5202db3388aa0a5c46f0589ffe6d62f8b6.

agriyakhetarpal:
Thanks for the review!

PR #5173, opened 2024-11-08
Main

maartenbreddels:
This allows us to use sqlite extensions (such as https://github.com/asg017/sqlite-vec/issues/135 )

 * Upgrade sqlite to a version at least compatible with https://github.com/asg017/sqlite-vec/issues/116
 * build/compile sqlite3 outselves so we can control the flags, and enable
 * TODO: build Python with PY_SQLITE_ENABLE_LOAD_EXTENSION since those flags are used when compiling the Python extension (I manually changed pyconfig.h by hand, but I guess it's a configure flag in pyodide about how to build Python, that I could not find).

### DIFF
--- a/packages/sqlite3/meta.yaml
+++ b/packages/sqlite3/meta.yaml
@@ -27,7 +27,39 @@ build:
       "Modules/_sqlite/util.c"
     )
 
-    embuilder build sqlite3 --pic
+    export SQLITE_WASM_VERSION=3450300
+    export SQLITE_WASM_YEAR=2024
+    wget -O sqlite.zip https://www.sqlite.org/${SQLITE_WASM_YEAR}/sqlite-src-${SQLITE_WASM_VERSION}.zip
+    unzip -q -o sqlite.zip -d sqlite
+    (cd sqlite/sqlite-src-${SQLITE_WASM_VERSION}/ && ./configure --enable-all && make sqlite3.c)
+    export SQLITE_FLAGS="\
+      -DSTDC_HEADERS=1 \
+      -DHAVE_SYS_TYPES_H=1 \
+      -DHAVE_SYS_STAT_H=1 \
+      -DHAVE_STDLIB_H=1 \
+      -DHAVE_STRING_H=1 \
+      -DHAVE_MEMORY_H=1 \
+      -DHAVE_STRINGS_H=1 \
+      -DHAVE_INTTYPES_H=1 \
+      -DHAVE_STDINT_H=1 \
+      -DHAVE_UNISTD_H=1 \
+      -DHAVE_FDATASYNC=1 \
+      -DHAVE_USLEEP=1 \
+      -DHAVE_LOCALTIME_R=1 \
+      -DHAVE_GMTIME_R=1 \
+      -DHAVE_DECL_STRERROR_R=1 \
+      -DHAVE_STRERROR_R=1 \
+      -DHAVE_POSIX_FALLOCATE=1 \
+      -DSQLITE_ENABLE_MATH_FUNCTIONS=1 \
+      -DSQLITE_ENABLE_FTS4=1 \
+      -DSQLITE_ENABLE_FTS5=1 \
+      -DSQLITE_ENABLE_RTREE=1 \
+      -DSQLITE_ENABLE_GEOPOLY=1 \
+      -DSQLITE_OMIT_POPEN=1 \
+      -DSQLITE_THREADSAFE=0 \
+      -DSQLITE_CORE"
+
+    emcc -c sqlite/sqlite-src-${SQLITE_WASM_VERSION}/sqlite3.c -o sqlite3.c.o -g3 -fPIC
 
     for file in "${FILES[@]}"; do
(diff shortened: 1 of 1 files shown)


hoodmane:
Well it checks in `configure` by making a file that calls `sqlite3_load_extension` and then linking it with `-lsqlite3`. You can force it to return `true` by `ac_cv_lib_sqlite3_sqlite3_load_extension=true`. 

The correct way to do this would be:
1. Clone sqlite3 and build a static library for it in `cpython/Makefile`
2. Install this into `cpython/installs/python-3.12.1/`
3. Set up pkg-config correctly so that it sees your build of sqlite

Then the configure script will automatically pick it up and see that it has `sqlite3_load_extension`. But this is probably more maintenance effort than you want to do, especially since we're not doing it very correctly already.

maartenbreddels:
Another trick could be to append "#define PY_SQLITE_ENABLE_LOAD_EXTENSION=1" to the right `pyconfig.h` (not the one we unpack in the recipe, but the one that is being 'used'), on codespaces it was this file:
```
/workspaces/pyodide/cpython/build/Python-3.12.1/pyconfig.h
```

I have done this manually on the github codespaces machine, and that works. What do you think of that?

hoodmane:
I'd prefer to set `ac_cv_lib_sqlite3_sqlite3_load_extension=true` when invoking the configure script. `pyconfig.h` is part of the output of the configure script, as much as possible we want to produce it in the normal way since it minimizes surprise.

maartenbreddels:
I wouldn't know how to do this. Any tips?

maartenbreddels:
FYI, a proof of concept using an extension:
https://py.cafe/maartenbreddels/sqlite-vec-demo

hoodmane:
Up to choice of ordering (maybe put it at end instead) the diff would look something like:
```
diff --git a/cpython/Makefile b/cpython/Makefile
index 07b6b356..d9bf8a8f 100644
--- a/cpython/Makefile
+++ b/cpython/Makefile
@@ -138,6 +138,7 @@ $(PYBUILD)/Makefile: $(PYBUILD)/.patched
 		cd $(PYBUILD); \
 		CONFIG_SITE=./Tools/wasm/config.site-wasm32-emscripten READELF=true emconfigure \
 		  ./configure \
+			  ac_cv_lib_sqlite3_sqlite3_load_extension=true \
 			  CFLAGS="${PYTHON_CFLAGS}" \
 			  CPPFLAGS="-sUSE_BZIP2=1 -sUSE_ZLIB=1" \
 			  PLATFORM_TRIPLET="$(PLATFORM_TRIPLET)" \
```

maartenbreddels:
I applied your patch, run `make clean; make` in the cpython directory and saw:
```
...
checking for libnsl... no
checking for library containing yp_match... no
checking for sqlite3 >= 3.7.15... yes
checking for sqlite3.h... yes
checking for sqlite3_bind_double in -lsqlite3... yes
checking for sqlite3_column_decltype in -lsqlite3... yes
checking for sqlite3_column_double in -lsqlite3... yes
checking for sqlite3_complete in -lsqlite3... yes
checking for sqlite3_progress_handler in -lsqlite3... yes
checking for sqlite3_result_double in -lsqlite3... yes
checking for sqlite3_set_authorizer in -lsqlite3... yes
checking for sqlite3_trace_v2 in -lsqlite3... yes
checking for sqlite3_value_double in -lsqlite3... yes
checking for sqlite3_load_extension in -lsqlite3... (cached) true
checking for sqlite3_serialize in -lsqlite3... yes
checking for --enable-loadable-sqlite-extensions... no
checking for gdbm.h... no
...
```

But i still have
```
/* Define to 1 to build the sqlite module with loadable extensions support. */
/* #undef PY_SQLITE_ENABLE_LOAD_EXTENSION */
```

In /workspaces/pyodide/cpython/build/Python-3.12.1/pyconfig.h


Any idea how I can get rid of this cache?

hoodmane:
Did you also pass `--enable-loadable-sqlite-extensions` to configure?

maartenbreddels:
Thanks, together with chatgpt i found a working method :)

packages/sqlite3/meta.yaml
```diff
@@ -27,16 +27,48 @@ build:
       "Modules/_sqlite/util.c"
     )
 
-    embuilder build sqlite3 --pic
+    export SQLITE_WASM_VERSION=3450300
+    export SQLITE_WASM_YEAR=2024
+    wget -O sqlite.zip https://www.sqlite.org/${SQLITE_WASM_YEAR}/sqlite-src-${SQLITE_WASM_VERSION}.zip
```
ryanking13, line 32: Would you mind updating the [sqlite3 version in Emscripten](https://github.com/emscripten-core/emscripten/blob/e6ed9d3fc2821a72cc8ded4a02cac72b9f306f88/tools/ports/sqlite3.py#L8-L10) instead? It would be a two-step process, as Emscripten should first update the sqlite3 version, then Pyodide need to update the emscripten version, but it reduces our maintenance burden.

PR #5177, opened 2024-11-08
`make_test_list.py`: Parse the version out of Makefile.envs

hoodmane:
This fixes `make_test_list.py` so it can be run with any version of Python as opposed to having to match the Python version Pyodide uses.

### DIFF
--- a/src/tests/make_test_list.py
+++ b/src/tests/make_test_list.py
@@ -4,16 +4,39 @@ Generate a list of test modules in the CPython distribution.
 
 import os
 from pathlib import Path
-from sys import version
 from typing import Any
+import subprocess
 
 import ruamel.yaml
 
 yaml = ruamel.yaml.YAML()
 PYODIDE_ROOT = Path(__file__).parents[2]
-LIB_DIR = PYODIDE_ROOT / "cpython/build" f"/Python-{version.split(' ')[0]}" f"/Lib/"
-
 PYTHON_TESTS_YAML = Path(__file__).parent / "python_tests.yaml"
+
+
+def get_makefile_envs():
+    result = subprocess.run(
+        ["make", "-f", str(PYODIDE_ROOT / "Makefile.envs"), ".output_vars"],
+        capture_output=True,
+        text=True,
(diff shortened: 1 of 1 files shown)


ryanking13 (approved):
Thanks! I really want to get rid of Makefile.envs and use a more structured format which other tools can easily parse... but we can do it later.

cclauss:
My humble suggestion would be to store the version in `pyproject.toml` and then autoincrement and replicate it into other locations with a tool like https://pypi.org/project/bump-my-version which I have used with great success.  It would be nice to also take GitHub Action `main / get_python_version (pull_request) Successful in 3s` (above) out of the loop.

cclauss:
Is it the case that you do not run pre-commit locally because `mypy`'s 22sec processing time is just too long for a commit?  For large projects like this one, `mypy` probably does not belong in pre-commit.

ryanking13:
Moving build configs to pyproject.toml or somewhere else sounds okay to me. But as those configs are used in both Makefile and Python codes, we need to find a good format that can be handled in both side.

Bumping up the Python version in Pyodide is not a simple procedure, so I think it is okay to handle is manually. On the other hand, for bumping up the Pyodide version, we tried bumpversion tool before, but ended in using our custom script since we needed to deal with both Python and JS version which use different versioning.

ryanking13:
> Is it the case that you do not run pre-commit locally because mypy's ~20sec processing time is just too long for a commit? For large projects like this one, mypy probably does not belong in pre-commit.

Yeah, running mypy in precommit was a bit annoying to me too. I am open to removing it from precommit if we can still run mypy checks in the PR

hoodmane:
Yeah I try to run precommit manually. I've been dealing with some repos that configure precommit to only run on push which I think is a big improvement since a lot of the time I do `git commit -a -m wip` so that I can switch branches and I don't want to run precommit in that case.

cclauss:
If pre-commit is subsecond then it is not an issue but if it is several seconds, it is painful.

is there a way for pre-commit to run the subsecond stuff (codespell + ruff check + ruff format) on every commit and then mypy on push?

PR #5181, opened 2024-11-09
DRAFT: pre-commit: Only run the slow mypy on pre-push and manual

cclauss:
### Description

As discussed at https://github.com/pyodide/pyodide/pull/5177#issuecomment-2466105227 some contributors may be avoiding the use of pre-commit because of slow pre-commit jobs like `mypy`.  Disabling pre-commit puts extra stress on our continuous integration jobs on issues that could have been fixed before modifications were pushed.

The proposed changes will cause the slower `mypy` job to be executed only on pre-push or manual runs.  Other (faster) pre-commit jobs will be executed in all stages but the `mypy` job will be run at least once before changes are sent to our continuous integration.
* https://pre-commit.com/#confining-hooks-to-run-at-certain-stages

Q: `stages: [pre-commit, pre-merge-commit, pre-push, manual]` -- Should `pre-merge-commit` be added?

The manual `pre-commit run --all-files` can take about 20 seconds.
* https://results.pre-commit.ci/run/github/122663163/1731166378.P2xoksEbSaKXB2K3F96vzQ

### Checklists

- [ ] Add / update tests

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -59,6 +59,7 @@ repos:
     rev: "v1.11.0"
     hooks:
       - id: mypy
+        stages: [pre-push, manual]
         files: ^(packages/.*/src|src|pyodide-build/pyodide_build)
         exclude: (setup.py|.*test.*)
         args: []
--- a/requirements.txt
+++ b/requirements.txt
@@ -13,3 +13,6 @@ pytest-cov
 pytest-httpserver
 pytest-benchmark
 pytest-pyodide==0.58.3
+setuptools; python_version >= '3.12'
+# TODO(cclauss): Remove the Selenium line.
+selenium==4.25.0


ryanking13:
Thanks for improving pre-commit! For now, mypy seems to be skipped in PR running by pre-commit.ci, but I think mypy still need to run in when a user open a PR as users (including me) often forget to run mypy locally before making a push.

> Should pre-merge-commit be added?

I think we don't need this.

cclauss:
After further testing I now understand that locally to make this work the user would need to:
% `pre-commit install --install-hooks --hook-type pre-commit --hook-type pre-push`

On the user's local machine `mypy` would ___not___ run on `git commit` but would run on `git push`.  This was our objective.  Problems:
1. Users are quite used to the simple % `pre-commit install`
2. Stages mess up with `pre-commit run --all-files` and the job will not be run if `stages: [pre-push, manual]` or `stages: [pre-push]`.  This explains why `pre-commit.ci` is never running `mypy`.  I will open a `pre-commit` issue on this.

Thoughts?  I believe that we should probably close this pull request with unmerged commits.

hoodmane:
Thanks for looking into this @cclauss!

PR #4580, opened 2024-02-29
Add the altair package

joelostblom:
### Description

This adds the altair packages and closes https://github.com/pyodide/pyodide/issues/4579. I wasn't sure what tests would be appropriate, it seems like some package like mpl, pandas, etc tests very specific things whereas others such as shapely, bokeh, statsmodels, etc keep it more minimal. I started minimal for now but happy to add more if there are issues.

I couldn't test the actual graphical output via `console.html` and when I tried saving to a file via `chart.save('chart.html')` no errors were raised but I couldn't find the file anywhere (maybe it is saved to a temporary/virtual filesystem in the browser?). 

### Checklists

- [x] Add a [CHANGELOG](https://github.com/pyodide/pyodide/blob/main/docs/project/changelog.md) entry
- [x] Add / update tests
- [x] Add new / update outdated documentation

PR #5183, opened 2024-11-10
Add `vega-datasets` package

peter-gy:
### Description

Adds https://pypi.org/project/vega-datasets/ package to provide synergy with existing packages such as `altair` (#4580).

### Checklists

- [x] Add / update tests


packages/vega-datasets/meta.yaml
```diff
@@ -0,0 +1,19 @@
...
+  home: http://github.com/altair-viz/vega_datasets
+  PyPI: https://pypi.org/project/vega-datasets
+  summary: A Python package for offline access to Vega datasets
+  license: MIT
+extra:
+  recipe-maintainers:
+    - peter-gy
+requirements:
+  run:
+    - pandas
```
agriyakhetarpal, line 19:
```suggestion
requirements:
  run:
    - pandas
extra:
  recipe-maintainers:
    - peter-gy
```

Just shifting their positions :)

packages/vega-datasets/test_vega_datasets.py
```diff
@@ -0,0 +1,23 @@
+import pytest
+from pytest_pyodide import run_in_pyodide
+
+
+@pytest.mark.xfail_browsers(firefox="slow")
+@run_in_pyodide(packages=["vega_datasets"])
```
agriyakhetarpal, line 6: We can leave this be if it's way too slow, but if it's something that takes thirty or so seconds or can complete within a reasonably finite amount of time, it's probably fine to remove the xfail and increase the default timeout by decorating it with `@pytest.mark.driver_timeout(30)`

peter-gy:
Hey @agriyakhetarpal, many thanks for the quick review.
I addressed your points in 7ebdda063fe63c729c6d29f3d31a785bc039f925.

ryanking13:
Thanks, but it seems like `vega-datasets` is a pure Python package that is installable with micropip. Is there some reason you want to add a recipe for it?

agriyakhetarpal:
> the convenience of being able to copy-paste official examples from Altair [docs](https://altair-viz.github.io/user_guide/data.html) directly, without an additional `await micropip.install('vega-datasets')`.

Installing `vega-datasets` this way would still be required, actually – you'd need to add an additional `%pip install` command at the top of your notebooks: https://jupyterlite.readthedocs.io/en/stable/howto/pyodide/packages.html#installing-packages-at-runtime

peter-gy:
> > the convenience of being able to copy-paste official examples from Altair [docs](https://altair-viz.github.io/user_guide/data.html) directly, without an additional `await micropip.install('vega-datasets')`.
> 
> Installing `vega-datasets` this way would still be required, actually – you'd need to add an additional `%pip install` command at the top of your notebooks: https://jupyterlite.readthedocs.io/en/stable/howto/pyodide/packages.html#installing-packages-at-runtime

Maybe I am missing something. My understanding is that if we have a recipe for `vega-datasets` under `packages/` of this repo, when I do an in-tree partial build e.g. via

```bash
PYODIDE_PACKAGES="tag:core,altair,pandas,vega-datasets" make
```

then I use the produced custom distribution for JupyterLite, then I would not force users to `%pip install vega-datasets` as it would be part of the distro.

This is also the method JupyterLite docs outline as the "the most reasonable way to make additional packages available by default when starting the Python kernel" (https://jupyterlite.readthedocs.io/en/latest/howto/pyodide/packages.html#bundling-additional-packages-by-default).

agriyakhetarpal:
> Maybe I am missing something. My understanding is that if we have a recipe for `vega-datasets` under `packages/` of this repo, when I do an in-tree partial build e.g. via
> 
> ```shell
> PYODIDE_PACKAGES="tag:core,altair,pandas,vega-datasets" make
> ```
> 
> then I use the produced custom distribution for JupyterLite, then I would not force users to `%pip install vega-datasets` as it would be part of the distro.
> 
> This is also the method JupyterLite docs outline as the "the most reasonable way to make additional packages available by default when starting the Python kernel" ([jupyterlite.readthedocs.io/en/latest/howto/pyodide/packages.html#bundling-additional-packages-by-default](https://jupyterlite.readthedocs.io/en/latest/howto/pyodide/packages.html#bundling-additional-packages-by-default)).

Ah, I understand now. Yes, I didn't think about a custom Pyodide distribution; that's a valid option. I'd say in that case, you could check out the sources for vega-datasets in another GitHub Actions step in your workflow, build a wheel for it out-of-tree – which should be quick, and then copy it into the folder where your distribution is located? (side note: is there a way to update the lockfile after that?) I have no qualms with a pure Python package being added because we have many already, but I guess this is what @ryanking13 had in mind as well in https://github.com/pyodide/pyodide/pull/5183#issuecomment-2467716186.

peter-gy:
> > Maybe I am missing something. My understanding is that if we have a recipe for `vega-datasets` under `packages/` of this repo, when I do an in-tree partial build e.g. via
> > ```shell
> > PYODIDE_PACKAGES="tag:core,altair,pandas,vega-datasets" make
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > then I use the produced custom distribution for JupyterLite, then I would not force users to `%pip install vega-datasets` as it would be part of the distro.
> > This is also the method JupyterLite docs outline as the "the most reasonable way to make additional packages available by default when starting the Python kernel" ([jupyterlite.readthedocs.io/en/latest/howto/pyodide/packages.html#bundling-additional-packages-by-default](https://jupyterlite.readthedocs.io/en/latest/howto/pyodide/packages.html#bundling-additional-packages-by-default)).
> 
> Ah, I understand now. Yes, I didn't think about a custom Pyodide distribution; that's a valid option. I'd say in that case, you could check out the sources for vega-datasets in another GitHub Actions step in your workflow, build a wheel for it out-of-tree – which should be quick, and then copy it into the folder where your distribution is located? (side note: is there a way to update the lockfile after that?) I have no qualms with a pure Python package being added because we have many already, but I guess this is what @ryanking13 had in mind as well in [#5183 (comment)](https://github.com/pyodide/pyodide/pull/5183#issuecomment-2467716186).

Producing the custom distribution out of tree is exactly what I have been doing until now, and that's what I wanted to partially facilitate with this PR. Updating the lockfile is possible but tedious because the `pyodide lockfile add-wheels ...` CLI has a bug (https://github.com/pyodide/pyodide-lock/issues/28, pending PR https://github.com/pyodide/pyodide-lock/pull/29) which prevents the lockfile from being modified.

Here is a brief overview of my current approach:

1. Use a modified version of https://github.com/pyodide/pyodide/blob/0d2835684eefa12e081d0eab1cefa824cc7e3aee/.github/workflows/main.yml#L54 to build a distribution with packages for which there are officially maintained recipes
2. Pull the wheel of `vega-datasets` from PyPI
3. Copy the wheel under the custom distribution's folder built in step (1)
4. Use `pyodide-lock` from a custom Python script to properly mutate the lockfile (since the CLI from shell has no effect on the lockfile due to the above-mentioned bug)
5. Build the JupyterLite bundle with a custom Pyodide distro pointing to the patched custom build

If `vega-datasets` was part of the official packages, I could skip steps (2)-(4).
If you would rather not have this recipe as part of this repo, I can maintain a fork where I have it and checkout the forked repo in my GH workflow in step (1). But as I saw the `altair` recipe also being maintained here, it seemed OK to upstream recipe of `vega-datasets`.

agriyakhetarpal:
Thanks for offering the context and for linking the issues, @peter-gy! I guess we should get those fixed and released soon since that would make your workflow slightly easier, and I checked that the size of the wheel is not that big either. I'll request a review from @ryanking13 and delegate the decision to him, but in the meantime, adding `vega-datasets` to Pyodide in-tree sounds like a good option to me.

ryanking13:
Yeah, I think we can merge this PR. Thanks!

PR #5186, opened 2024-11-11
Fix sessionStorage-related crash when running in sandboxed iframe

josephrocca:
Fixes this problem when Pyodide is running within a [sandboxed](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/iframe#sandbox) iframe:

> Uncaught SecurityError: Failed to read the 'sessionStorage' property from 'Window': The document is sandboxed and lacks the 'allow-same-origin' flag.

image

The above screenshot is from Chrome. I haven't tested whether Firefox and Safari also throw and error for this.

The workaround in the meantime:
```js
delete window.sessionStorage
window.sessionStorage = {}
// now load Pyodide...
```

### DIFF
--- a/docs/project/changelog.md
+++ b/docs/project/changelog.md
@@ -62,6 +62,9 @@ myst:
   [Emscripten docs](https://emscripten.org/docs/api_reference/Filesystem-API.html#FS.trackingDelegate[callback%20name])
   for more information.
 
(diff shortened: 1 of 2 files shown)


hoodmane (approved):
Looks good to me, thanks!

hoodmane:
Please add a changelog entry.

Issue #5187, opened 2024-11-11
ModuleNotFoundError: No module named 'cmake_build_extension' when building package


stephanmg:
I tried to use `pyodide skeleton pypi amici` to create a package of the useful [AMICI](https://github.com/AMICI-dev/AMICI) package.

However I fail when using `pyodide build-recipes amici --install` after. 

AMICI seems to require `cmake-build-extension`.  I cloned the `pyodide` repository, started the docker container `./run_docker` and build `pyodide-build` according to the documentation.

Since `cmake-build-extension` seems to be required, I added it to the `meta.yaml` file under the subsection `host` in section `requirements`. While I can import the module from within the running Docker container, i.e. `python -c 'import cmake_build_extension` works, `pyodide build-recipes amici --install` fails, resulting in a `ModuleNotFoundError` for the former import statement.

My branch (forked from `main` branch of pyodide this morning) containing the `meta.yaml` is here:
https://github.com/stephanmg/pyodide/tree/package-amici/packages/amici

Any help / pointing to resources to help me create this package would be appreciated.

P.S.: Thank you @agriyakhetarpal for your help on Discord already.

## 🐍 Package Issue / Request

- AMICI, version latest
- [Package URL](https://pypi.org/project/amici/) 
- Package Dependencies that needs to be resolved first: A `setup.py` is provided and a `Project.toml` file.

## Checklists

- [x] I have tried to install the package using `micropip.install(...)` and encountered the issue that the package is not pure Python and uses extensively C extensions
- [x] I have read the [documentation](https://pyodide.org/en/stable/development/building-and-testing-packages.html) and tried building the package myself.

hoodmane:
Thanks for the report @stephanmg. The build occurs in an isolated virtual environment, so installing `cmake-build-extension` into the top level environment doesn't help. The problem is that it gets dropped here:
https://github.com/pyodide/pyodide-build/blob/main/pyodide_build/pypabuild.py#L122
because cmake is in the list of "avoided requirements":
https://github.com/pyodide/pyodide-build/blob/main/pyodide_build/pypabuild.py#L36-L39
[...]

PR #5188, opened 2024-11-11
Testing for pyodide/pyodide-build#60

hoodmane:
See upstream PR:
https://github.com/pyodide/pyodide-build/pull/60

Resolves pyodide/pyodide#5187. In that issue, the build requires a package called `cmake-build-extension` but because we have `cmake` in the list of requirements to avoid we drop it and it breaks the build.

This logic was introduced in pyodide/pyodide#2272 and did not come up in the review. Surprisingly this is the first problem it has caused. Let's see if this change breaks anything in Pyodide CI.


ryanking13:
It looks like skipping (host) numpy installation is not working after this change, resulting in scipy build error.

hoodmane:
Hmm I tried building scipy locally and it worked fine...

agriyakhetarpal:
I updated the submodule to point to https://github.com/pyodide/pyodide-build/pull/60 again in the previous commit, as we no longer include CMake as a part of avoided build-time requirements.

agriyakhetarpal:
Actually, if `patchelf` is the only avoided requirement now, do we even need pyodide/pyodide-build#60?

agriyakhetarpal:
Actually, if `patchelf` is the only avoided requirement now, do we even need pyodide/pyodide-build#60? I guess we could just merge the submodule and #5187 should work fully.

ryanking13:
I am having the statsmodels build issue in [pyodide-recipes PR](https://github.com/pyodide/pyodide-recipes/pull/72) as well. It uses tot pyodide-build, so I think it is not related to pyodide/pyodide-build#60. Maybe there was other regression.

PR #5189, opened 2024-11-11
MAINT Switch to a different mirror for Autoconf 2.71

agriyakhetarpal:
### Description

The current mirror has been failing for some time (intermittently) because of a "403: Forbidden" error. I've changed it to point to a different source in this PR.

### Checklists

N/A

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -19,7 +19,7 @@ RUN wget -q -O - https://sh.rustup.rs | \
     RUSTUP_HOME=/usr CARGO_HOME=/usr sh -s -- -y --profile minimal --no-modify-path
 
 # install autoconf 2.71, required by upstream libffi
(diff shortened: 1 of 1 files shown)


agriyakhetarpal:
> Python uses `https://ftp.gnu.org/gnu/autoconf/autoconf-${AUTOCONF_VERSION}.tar.gz` maybe we should too? [python/cpython-devcontainers@`main`/autoconf/Dockerfile#L24C10-L24C78](https://github.com/python/cpython-devcontainers/blob/main/autoconf/Dockerfile?rgh-link-date=2024-11-11T14%3A35%3A18Z#L24C10-L24C78)

Yes, I thought about it, but the Autoconf guidelines suggest that we use a mirror if possible: https://www.gnu.org/software/autoconf/#TOCdownloading. But I would be happy to make the switch if you insist.

hoodmane:
I do not insist.

PR #5192, opened 2024-11-11
MAINT Fix cross-build metadata update workflow, open issue on failures

agriyakhetarpal:
### Description

Continuing from https://github.com/pyodide/pyodide/pull/5168#issuecomment-2462387339, this PR updates the `update_cross_build_releases.yml` workflow to open an issue in the repository, since [the 0.26.3 update workflow run failed](https://github.com/pyodide/pyodide/actions/runs/11417310347) and we did not realise that was the case. Additionally, it fixes the cause of the failure by using `github.event.release.name` instead of `tag_name` which is sometimes inexplicably empty (and the tag name and the release name are the same in our case). I tested it on my fork with a workflow dispatch trigger: agriyakhetarpal/pyodide#2 and with a release: agriyakhetarpal/pyodide#4, and the update worked in both cases. I didn't test the issue creation job because I've kept issues disabled for my fork.

### Checklists

N/A

### DIFF
--- a/.github/workflows/update_cross_build_releases.yml
+++ b/.github/workflows/update_cross_build_releases.yml
@@ -6,7 +6,7 @@ on:
   workflow_dispatch:
     inputs:
       version:
-        description: "Version number"
+        description: "Version number to update to. This should not contain the 'v' prefix, since our releases on GitHub do not have it."
         required: true
 
 env:
--- a/.github/workflows/update_cross_build_releases.yml
+++ b/.github/workflows/update_cross_build_releases.yml
@@ -10,7 +10,7 @@ on:
         required: true
 
 env:
-  version: ${{ github.event_name == 'release' && github.event.release.name || github.event.inputs.version }}
+  version: ${{ github.event_name == 'release' && github.event.release.tag_name || github.event.inputs.version }}
 
 jobs:
   update_cross_build_releases:
--- a/.github/workflows/update_cross_build_releases.yml
+++ b/.github/workflows/update_cross_build_releases.yml
@@ -42,8 +42,9 @@ jobs:
           commit-message: Update cross-build metadata file for version ${{ env.version }} [skip ci]
           title: Update cross-build metadata file for version ${{ env.version }}
           body: |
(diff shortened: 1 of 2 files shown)


.github/workflows/update_cross_build_releases.yml
```diff
@@ -6,9 +6,12 @@ on:
   workflow_dispatch:
     inputs:
       version:
-        description: "Version number"
+        description: "Version number to update to. This should not contain the 'v' prefix, since our releases on GitHub do not have it."
         required: true
 
+env:
+  version: ${{ github.event_name == 'release' && github.event.release.name || github.event.inputs.version }}
```
ryanking13, line 13: I used `name` before, but it didn't work, so I replaced it with `tag_name` in https://github.com/pyodide/pyodide/pull/4820. 0.26.3 patch release does not include this commit so it probably used `name` and failed.

I am not exactly sure about the difference between the `name` and `tag_name` though. But could you please double check it?

ryanking13 (approved):
Thanks! I left a comment about the parameter name.

.github/workflows/update_cross_build_releases.yml
agriyakhetarpal: I'll double-check it – sure! I'm convinced that I used `release.name` yesterday and created a tag + a release from it on my fork, which worked (as seen from the PRs that were opened in my fork in the PR description). But since #4820 wasn't included in 0.26.3 and I can't find the code for what @pyodide-ci-bot is doing, let's go with `tag_name` as you suggest, since that will be a safer way.

Also, looking at https://github.com/pyodide/pyodide/actions/workflows/update_cross_build_releases.yml, it failed for all 0.26.X releases and all of the runs used `release.name`, it makes sense to switch.

agriyakhetarpal:
```suggestion
  version: ${{ github.event_name == 'release' && github.event.release.tag_name || github.event.inputs.version }}
```

agriyakhetarpal: Addressed in 3808d351668903d22fda116de69ee80574e58c1f. Thanks!

PR #5193, opened 2024-11-12
fix: npe from API.lockfile_info.version

mscolnick:
Fixes an NPE on `API.lockfile_info.version` since its not assigned from `lockfile.info.version`

cc @hoodmane


ryanking13 (approved):
Thanks!

PR #5208, opened 2024-11-19
Get PYVERSION from pyproject.toml project.requires-python

cclauss:
`pyproject.toml`:
```toml
[project]
requires-python = "==3.12.7"
```
As discussed at https://pyodide.org/en/stable/development/maintainers.html#steps we should try to hardcode the version of Python in fewer places.

Use [`pyproject.toml` `[project]` `requires-python`](https://packaging.python.org/en/latest/guides/writing-pyproject-toml/#python-requires) as the source of the `major.minor.micro` version of Python to build and use.

* [x] Use GitHub Action [`setup-python` `python-version-file`](https://github.com/actions/setup-python/blob/main/docs/advanced-usage.md#using-the-python-version-file-input) to enable GitHub Actions to read that value directly.
```
Run actions/setup-python@v5
  with:
    python-version-file: pyproject.toml
Extracted ==3.12.7 from pyproject.toml
Installed versions
  Successfully set up CPython (3.12.7)
```
Perhaps the proposed changes do not deliver sufficient improvement.

### Checklists

- [ ] Add / update tests

### DIFF
--- a/Makefile.envs
+++ b/Makefile.envs
@@ -1,4 +1,12 @@
-export PYVERSION ?= 3.12.7
+export PYVERSION ?= $(python3 <<EOF
+import sys, tomllib
+with open("pyproject.toml", "rb") as in_file:
+    requires_python = tomllib.load(in_file).get("project", {}).get("requires-python")
+assert requires_python, "requires-python not found in pyproject.toml"
+sys.stdout.write(f"{requires_python}\n")
+EOF
+)
+
 export PYODIDE_EMSCRIPTEN_VERSION ?= 3.1.58
 export PYODIDE_VERSION ?= 0.27.0.dev0
 export PYODIDE_ABI_VERSION ?= 2024_0
--- a/Makefile.envs
+++ b/Makefile.envs
@@ -1,12 +1,6 @@
-export PYVERSION ?= $(python3 <<EOF
-import sys, tomllib
-with open("pyproject.toml", "rb") as in_file:
-    requires_python = tomllib.load(in_file).get("project", {}).get("requires-python")
-assert requires_python, "requires-python not found in pyproject.toml"
-sys.stdout.write(f"{requires_python}\n")
-EOF
-)
+# Read project.requires-python from the pyproject.toml file.
 
+export PYVERSION ?= $(shell python3 tools/get_python_version.py)
 export PYODIDE_EMSCRIPTEN_VERSION ?= 3.1.58
 export PYODIDE_VERSION ?= 0.27.0.dev0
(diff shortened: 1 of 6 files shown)


Makefile.envs
```diff
@@ -1,4 +1,12 @@
-export PYVERSION ?= 3.12.7
+export PYVERSION ?= $(python3 <<EOF
+import sys, tomllib
+with open("pyproject.toml", "rb") as in_file:
+    requires_python = tomllib.load(in_file).get("project", {}).get("requires-python")
+assert requires_python, "requires-python not found in pyproject.toml"
+sys.stdout.write(f"{requires_python}\n")
+EOF
+)
```
hoodmane, line 8: I don't think heredoc works in defining makefiles. I think you would need to move this into `tools/get_py_version.py` and do:
```
export PYVERSION ?= $(shell python3.12 get_version.py)
```

hoodmane: But I'm concerned about the performance implications of this. Before this change, 
```sh
time make -f Makefile.envs .output_vars 
```
reports that it takes 5 ms. After this change it takes 300 ms so 60x longer. My guess is we source this file about 1000 times in our CI, so this could add as much as 5 minutes to our CI time. WDYT @ryanking13?

cclauss:
Closing.  Too slow and no real savings.

Makefile.envs
ryanking13: I don't think this is better than the current way. Let's think more about how we can improve our Makefile.

agriyakhetarpal:
We also have `ripgrep` as an option, right?

Here's what I noticed from a local run on a macOS machine:

#### with `PYVERSION` hardcoded in `Makefile.envs`

```bash
❯ hyperfine "make -f Makefile.envs .output_vars" --warmup 1
Benchmark 1: make -f Makefile.envs .output_vars
  Time (mean ± σ):      37.7 ms ±   9.4 ms    [User: 22.6 ms, System: 11.3 ms]
  Range (min … max):    33.2 ms …  99.5 ms    68 runs
```

#### Using `ripgrep` to parse the version from `pyproject.toml`

Command used: `rg -o -m 1 'requires-python = "3\.[0-9]+\.[0-9]+"' pyproject.toml | cut -d'"' -f2`

```bash
❯ hyperfine "make -f Makefile.envs .output_vars" --warmup 1
Benchmark 1: make -f Makefile.envs .output_vars
  Time (mean ± σ):       9.2 ms ±   1.7 ms    [User: 3.0 ms, System: 4.7 ms]
  Range (min … max):     7.9 ms …  24.7 ms    192 runs
```

so it's actually somewhat faster (not sure how, though).

PR #5209, opened 2024-11-19
Package sundials

stephanmg:
### Description

Adding a build-recipe for Sundials.

### DIFF
--- a/packages/sundials/meta.yaml
+++ b/packages/sundials/meta.yaml
@@ -1,0 +1,31 @@
+package:
+  name: sundials
+  version: 7.1.1
+
+source:
+  url: https://github.com/LLNL/sundials/releases/download/v7.1.1/sundials-7.1.1.tar.gz
+
+  sha256: ea7d6edfb52448ddfdc1ec48f89a721fe6c0a259c10f8ef56f60fcded87a94bb
+
+
+requirements:
+  host:
+    - openblas
+    - suitesparse
+
+build:
+  type: shared_library
+  script: |
+      mkdir build
+      cd build
+
+      emcmake cmake .. -DBUILD_ARKCODE=OFF -DBUILD_CVODE=OFF -DEXAMPLES_ENABLE_C=OFF -DENABLE_KLU=ON -DEXAMPLES_INSTALL=OFF
+
+      emmake make -j ${PYODIDE_JOBS:-3} install \
+      LDFLAGS="${SIDE_MODULE_LDFLAGS} -L${WASM_LIBRARY_DIR}/lib " \
+      BLAS="${WASM_LIBRARY_DIR}/lib/libopenblas.so" \
+      LAPACK="${WASM_LIBRARY_DIR}/lib/libopenblas.so" \
+      INSTALL=${WASM_LIBRARY_DIR} 
+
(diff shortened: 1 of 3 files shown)


packages/sundials/meta.yaml
```diff
@@ -0,0 +1,27 @@
...
+    - openblas
+    - suitesparse
+
+build:
+  type: shared_library
+  script: |
+    mkdir build
+    cd build
+
+    emcmake cmake .. -DBUILD_ARKCODE=OFF -DBUILD_CVODE=OFF -DEXAMPLES_ENABLE_C=OFF -DENABLE_KLU=ON -DEXAMPLES_INSTALL=OFF -DKLU_LIBRARY_DIR=${WASM_LIBRARY_DIR}/lib/ -DKLU_INCLUDE_DIR=${WASM_LIBRARY_DIR}/include
```
agriyakhetarpal, line 21:
```suggestion
    emcmake cmake .. -DBUILD_ARKODE=OFF -DBUILD_CVODE=OFF -DEXAMPLES_ENABLE_C=OFF -DENABLE_KLU=ON -DEXAMPLES_INSTALL=OFF -DKLU_LIBRARY_DIR=${WASM_LIBRARY_DIR}/lib/ -DKLU_INCLUDE_DIR=${WASM_LIBRARY_DIR}/include
```

stephanmg: Should also probably disable `KINSOL` and `IDA`.

agriyakhetarpal: Yes, if AMICI doesn't require them, please do

agriyakhetarpal:
I think COLAMD hasn't been built, so it needs to be enabled, too – or that the order of building is incorrect for some reason.

Last time I checked, it was: SuiteSparse_config, AMD, COLAMD, BTF, and then KLU (in that order, I guess).

stephanmg:
> Maybe we can consolidate #5210 here itself, since the build will be easier to validate in the same PR.

Do you want me to create another PR consolidating both and close #5210 and #5209 both?

agriyakhetarpal:
> > Maybe we can consolidate #5210 here itself, since the build will be easier to validate in the same PR.
> 
> Do you want me to create another PR consolidating both and close #5210 and #5209 both?

That would be great! Alternatively, since #5209 is this PR, so maybe we can bring the changes from #5210 here?

agriyakhetarpal:
I think the build failure is because the order is incorrect; KLU via the second patch is getting built before of COLAMD, but COLAMD is required as a prerequisite for KLU to build. Hence, it fails to find `libcolamd.so` because it hasn't been built yet.

stephanmg:
@agriyakhetarpal next try ;-)

stephanmg:
@agriyakhetarpal yes will do, will update here when ready.

agriyakhetarpal:
Marking this PR as a draft for the time being, thanks for your work!

stephanmg:
> Marking this PR as a draft for the time being, thanks for your work!

Thanks for marking that as a draft, currently I'm too swamped to finalize this, it should not much be left to do to get this working.

ryanking13:

PR #5230, opened 2024-11-28
Ruff rules for comprehensions and performance

cclauss:
### Description

Similar to:
* pyodide/pyodide-build#66

Upgrade [`ruff`](https://docs.astral.sh/ruff) to the current version.

Complete the compliance with Pylint rules.

Add various list, dict, set comprehension, and performance improvements.
* ~https://docs.astral.sh/ruff/rules/#flake8-async-async~
* https://docs.astral.sh/ruff/rules/#flake8-comprehensions-c4
* https://docs.astral.sh/ruff/rules/#perflint-perf

### Checklists

- [ ] Add / update tests

### DIFF
--- a/benchmark/benchmark.py
+++ b/benchmark/benchmark.py
@@ -20,8 +20,8 @@ SKIP = {"fft", "hyantes"}
 def print_entry(name, res):
     print(" - ", name)
     print(" " * 4, end="")
-    for name, dt in res.items():
-        print(f"{name}: {dt:.6f}  ", end="")
+    for res_name, dt in res.items():
+        print(f"{res_name}: {dt:.6f}  ", end="")
     print("")
 
 
--- a/conftest.py
+++ b/conftest.py
@@ -181,7 +181,7 @@ def pytest_configure(config):
     it to reduce the verbosity of the test names in the table.  This leaves
     enough room to see the information about the test failure in the summary.
     """
-    global CONFIG
+    global CONFIG  # noqa: PLW0602
 
     old_cwd_relative_nodeid = config.cwd_relative_nodeid
 
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -21,19 +21,19 @@ project = "Pyodide"
 copyright = "2019-2024, Pyodide contributors and Mozilla"
 
 nitpicky = True
-nitpick_ignore = []
+nitpick_ignore: list[tuple[str, str]] = []
 
 
 def ignore_typevars():
     """These are all intentionally broken. Disable the warnings about it."""
-    PY_TYPEVARS_TO_IGNORE = ["T", "T_co", "T_contra", "V_co", "KT", "VT", "VT_co"]
-    JS_TYPEVARS_TO_IGNORE = ["TResult", "TResult1", "TResult2", "U"]
+    PY_TYPEVARS_TO_IGNORE = ("T", "T_co", "T_contra", "V_co", "KT", "VT", "VT_co")
+    JS_TYPEVARS_TO_IGNORE = ("TResult", "TResult1", "TResult2", "U")
 
-    for typevar in PY_TYPEVARS_TO_IGNORE:
-        nitpick_ignore.append(("py:obj", f"_pyodide._core_docs.{typevar}"))
-
-    for typevar in JS_TYPEVARS_TO_IGNORE:
-        nitpick_ignore.append(("js:func", typevar))
+    nitpick_ignore.extend(
+        ("py:obj", f"_pyodide._core_docs.{typevar}")
+        for typevar in PY_TYPEVARS_TO_IGNORE
+    )
+    nitpick_ignore.extend(("js:func", typevar) for typevar in JS_TYPEVARS_TO_IGNORE)
 
 
 ignore_typevars()
@@ -292,7 +292,7 @@ def ensure_typedoc_on_path():
     if shutil.which("typedoc"):
         return
     if IN_READTHEDOCS:
-        subprocess.run(["npm", "ci"], cwd="../src/js")
+        subprocess.run(["npm", "ci"], cwd="../src/js", check=True)
         Path("../node_modules").symlink_to("../src/js/node_modules")
     if shutil.which("typedoc"):
         return
--- a/src/tests/test_pyproxy.py
+++ b/src/tests/test_pyproxy.py
@@ -387,7 +387,7 @@ def test_pyproxy_get_buffer_type_argument(selenium, array_type):
         assert check
         if fmt.lower() == "q":
             assert result == [hex(x).replace("0x", "") for x in list(mv.cast(fmt))]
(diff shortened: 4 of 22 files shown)


benchmark/benchmark.py
```diff
@@ -20,7 +20,7 @@
 def print_entry(name, res):
     print(" - ", name)
     print(" " * 4, end="")
-    for name, dt in res.items():
+    for name, dt in res.items():  # noqa: PLR1704
```
hoodmane, line 23: What's this rule?

conftest.py
```diff
@@ -181,7 +181,7 @@ def pytest_configure(config):
     it to reduce the verbosity of the test names in the table.  This leaves
     enough room to see the information about the test failure in the summary.
     """
-    global CONFIG
+    global CONFIG  # noqa: PLW0602
```
hoodmane, line 184: Can we blanket disable this rule against globals?

docs/conf.py
```diff
@@ -292,7 +292,7 @@ def ensure_typedoc_on_path():
     if shutil.which("typedoc"):
         return
     if IN_READTHEDOCS:
-        subprocess.run(["npm", "ci"], cwd="../src/js")
+        subprocess.run(["npm", "ci"], cwd="../src/js", check=False)
```
hoodmane, line 295: Good catch. Let's make this `check=True` here:
```suggestion
        subprocess.run(["npm", "ci"], cwd="../src/js", check=True)
```

benchmark/benchmark.py
cclauss: % [`ruff rule PLR1704`](https://docs.astral.sh/ruff/rules/redefined-argument-from-local)
# redefined-argument-from-local (PLR1704)

Derived from the **Pylint** linter.

## What it does
Checks for variables defined in `for`, `try`, `with` statements
that redefine function parameters.

## Why is this bad?
Redefined variables can cause unexpected behavior because of overridden function parameters.
If nested functions are declared, an inner function's body can override an outer function's parameters.

## Example
```python
def show(host_id=10.11):
    for host_id, host in [[12.13, "Venus"], [14.15, "Mars"]]:
        print(host_id, host)
```

Use instead:
```python
def show(host_id=10.11):
    for inner_host_id, host in [[12.13, "Venus"], [14.15, "Mars"]]:
        print(host_id, inner_host_id, host)
```

## Options
- `lint.dummy-variable-rgx`

## References
- [Pylint documentation](https://pylint.readthedocs.io/en/latest/user_guide/messages/refactor/redefined-argument-from-local.html)

conftest.py
cclauss: Doing it this way triggers the "Do we really need a global here?" discussion in code reviews and when refactoring but I will make the requested change.

hoodmane (commented):
Thanks @cclauss!

conftest.py
hoodmane: I think `global` by itself triggers a "Do we really need a global here?" discussion in review though. It is at least generally a short line.

hoodmane: > Now that I see that this rule is violated only once in the entire codebase, I am even more reluctant to change but... here we go.

Yeah maybe you can put it back on seeing as it only happens once.

hoodmane: Sorry.

src/tests/test_pyproxy.py
```diff
@@ -387,7 +387,7 @@ def test_pyproxy_get_buffer_type_argument(selenium, array_type):
         assert check
         if fmt.lower() == "q":
             assert result == [hex(x).replace("0x", "") for x in list(mv.cast(fmt))]
-        elif fmt == "f" or fmt == "d":
+        elif fmt in {"d", "f"}:
```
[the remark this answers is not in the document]
cclauss: `fmt = "df"` would be `True` using `str` lookup.  `set` lookup is marginally but consistently faster on my M1 Mac.
```
>>> from timeit import repeat
>>> use_set = "for fmt in 'abcdefg': fmt in {'d', 'f'}"
>>> use_str = "for fmt in 'abcdefg': fmt in 'df'"
>>> repeat(use_set, repeat=10)
>>> repeat(use_str, repeat=10)
```

tools/create_xbuildenv.py
```diff
@@ -13,7 +14,7 @@
     from pyodide_build.recipe import load_all_recipes
 except ImportError:
     print("Requires pyodide-build package to be installed")
-    exit(1)
+    sys.exit(1)
```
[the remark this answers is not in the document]
cclauss: % [`ruff rule PLR1722`](https://docs.astral.sh/ruff/rules/sys-exit-alias)

src/tests/test_pyproxy.py
```diff
@@ -446,36 +446,56 @@ class AwaitNext(Await, Next): pass
         return result;
         """
     )
-    assert result == dict(
```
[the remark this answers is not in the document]
cclauss: I will revert.  This rule merely saves on function call overhead.

benchmark/benchmark.py
hoodmane: Nice maybe we should rename the variable because that is bad.

cclauss:
Do we care about the sync-inside-async ASYNC230 issues or do we just leave the noqa in place?
* https://docs.astral.sh/ruff/rules/blocking-open-call-in-async-function

docs/conf.py
```diff
@@ -30,10 +30,10 @@ def ignore_typevars():
     JS_TYPEVARS_TO_IGNORE = ["TResult", "TResult1", "TResult2", "U"]
 
     for typevar in PY_TYPEVARS_TO_IGNORE:
-        nitpick_ignore.append(("py:obj", f"_pyodide._core_docs.{typevar}"))
+        nitpick_ignore.append(("py:obj", f"_pyodide._core_docs.{typevar}"))  # noqa: PERF401
```
hoodmane, line 33: Maybe let's go along with the lint and change this to:
```py
nitpick_ignore.extend(("py:obj", f"_pyodide._core_docs.{typevar}") for typevar in PY_TYPEVARS_TO_IGNORE)
nitpick_ignore.extend(("js:func", typevar) for typevar in JS_TYPEVARS_TO_IGNORE)
```

hoodmane (approved):
Thanks!

hoodmane:
> Do we care about the two sync-inside-async ASYNC230 issues or do we just leave the noqa in place?

I think it would be good to turn them off. open is very quick and there is no way we are going to pull in a dependency for an async-open. It's unclear if anyio even works. I guess it would be good to check. But it's a minor point.

agriyakhetarpal (approved):
Thanks, @cclauss! Are there any more rules to add here? I just wanted to ask before we merge this.

hoodmane:
Let's go ahead and merge this.

PR #5235, opened 2024-11-30
Update rateslib to version 1.6.0

attack68:
Update meta for new release.


agriyakhetarpal (approved):
Thanks, @attack68! I am not sure why it failed to find the Rustup executable, so I've re-triggered the CI job.

ryanking13:
I think we are hitting the same issue described in: https://github.com/pyodide/pyodide/pull/5112#issuecomment-2413779118

ryanking13:
```
lock file version 4 requires `-Znext-lockfile-bump`
```

This error looks suspicious.

Could you try putting the following yaml in the meta.yaml file

```
build:
  script: |
    export RUSTFLAGS="${RUSTFLAGS} -Znext-lockfile-bump"
```

or, downgrade the Cargo.lock file version?

packages/rateslib/meta.yaml
```diff
@@ -14,10 +14,11 @@ requirements:
   executable:
     - rustup
 about:
```
attack68, line 16:
```suggestion
    - rustup
build:
  script: |
    export RUSTFLAGS="${RUSTFLAGS} -Znext-lockfile-bump"    
about:
```

attack68:
Didnt seem to work:

```
  error: unknown unstable option: `next-lockfile-bump` 
```

This post seems to suggest that you need to upgrade the rust toolchain to >1.78.
https://users.rust-lang.org/t/lock-file-version-4-requires-znext-lockfile-bump/119754/2

Servers are running on 1.77
```
nightly-2024-01-29-x86_64-unknown-linux-gnu unchanged - rustc 1.77.0-nightly  
(635124704 2024-01-27)
```

ryanking13:
I see. Looks like it requires a quite new rust version. I'll try to update the rust version in our docker image.

Meanwhile, could you try to regenerate the cargo.lock file as the post says?

attack68:
The Python package is published and versioned, Pypi wont allow another upload under the same 1.6.0 version.
Maybe I can link it to another source but that potentially breaks file integrity and transparency of versions.
Otherwise might be best to wait for 1.6.1 or 1.7.0, by which time this issue might be solved anyway.

ryanking13:
#5237 will update the rust toolchain version used in Pyodide.

agriyakhetarpal:
#5249 is trying to update the Rust toolchain again, but maybe the build works with the one updated in #5237. Could you please try again, @attack68?

agriyakhetarpal:
I guess it's better to wait until 1.6.1 or 1.7.0 indeed...

attack68:
if i do a 1.6.1 tje requirement is still to have a lower cargo.lock version? Is that still relevant advice after the rust toolchain upgrade though or is it another issue?

agriyakhetarpal:
If a lower lock version can do the trick, we can go back to an older Rust version and be successful, I think that should work. It could also be a patch instead of a new 1.6.1 release if you wish it to be. I don't have any formal experience with writing or working with Rust :)

packages/rateslib/meta.yaml
```diff
@@ -1,23 +1,27 @@
...
 requirements:
   run:
     - numpy
     - pandas
     - matplotlib
   executable:
     - rustup
+build:
+  script: |
+    export RUSTFLAGS="${RUSTFLAGS} -Znext-lockfile-bump"
```
attack68, line 18:
```suggestion
```

agriyakhetarpal (approved):
Thanks, @attack68, I'm happy to approve now that the builds and tests are passing!

agriyakhetarpal (approved):
Thanks, @attack68, I'm happy to approve now that the builds and tests are passing! I pushed a CHANGELOG entry, which was missed in the previous PR where the package was added.

PR #5236, opened 2024-12-02
Update urllib3 and bring in Joe Marshall's JSPI support

hoodmane:
The JSPI PR https://github.com/urllib3/urllib3/pull/3427 is currently unreleased so I backported it onto 2.2.3

cc @joemarshall


packages/urllib3/patches/0001-Remove-emscripten-shim.patch
ryanking13, line 1: This one doesn't seem to be added in meta.yaml?

ryanking13 (approved):

PR #5244, opened 2024-12-05
Deduplicate pure_eval package

bartbroere:
I believe that Pyodide accidentally has `pure-eval` and `pure_eval` in the packages directory. Or am I missing something about the way this works?


bartbroere:
Looking at the diff of #5086 it seems that `pure-eval` was the only package that was not renamed in git, but added.

So I think this PR can be applied without issues, deduplicating the specification of the `pure-eval` package.

agriyakhetarpal (approved):
Thanks, @bartbroere, this is a good catch!

hoodmane:
Looks like some package declares a dependency on `pure_eval` and it needs to be updated to `pure-eval`? Seems to be in `packages/ipython/meta.yaml`.

bartbroere:
> Looks like some package declares a dependency on `pure_eval` and it needs to be updated to `pure-eval`? Seems to be in `packages/ipython/meta.yaml`.

Done! Building it again

PR #5251, opened 2024-12-10
Add apsw package

airen1986:
### Description

Add apsw package, which is another SQLITE wrapper, it uses latest SQLITE version 3.47.2

### Checklists

- [N] Add a [CHANGELOG](https://github.com/pyodide/pyodide/blob/main/docs/project/changelog.md) entry
- [Y] Add / update tests
- [N] Add new / update outdated documentation


packages/apsw/meta.yaml
```diff
@@ -0,0 +1,19 @@
...
+  url: https://files.pythonhosted.org/packages/e8/0b/de1376f0ffaec714eff65a203d1b86a3ff6aac3ef61ca70be63f0b516da2/apsw-3.47.2.0.tar.gz
+  sha256: 95d03c1236bb247d7b2241eda229bc9025fb6edb2b49257fc52a0ee64f5b36bc
+test:
+  imports:
+    - apsw
+about:
+  home: https://github.com/rogerbinns/apsw
+  PyPI: https://pypi.org/project/apsw
+  summary: Another Python SQLite Wrapper
+  license: OSI Approved
```
agriyakhetarpal, line 16: Could we switch to the SPDX license identifier? [`apsw`'s LICENSE file](https://github.com/rogerbinns/apsw/blob/master/LICENSE) says it is https://spdx.org/licenses/any-OSI.html.
```suggestion
  license: any-OSI
```

agriyakhetarpal (commented):
Thanks, @airen1986, could you please add a CHANGELOG entry as well?

agriyakhetarpal (approved):

akhilesh-bayer:
@ryanking13  Could it be available in 0.27 release?

agriyakhetarpal:
@akhilesh-bayer, IIRC, this package was not included in the recent release because of the build times. However, it can be included in a 0.27.1 or 0.27.2 release - let's see how long it takes to build in the current CI run. There is also an option for you to build the package out-of-tree (see https://github.com/joemarshall/pyarrow-pyodide for an example) and then include the URL to the wheel here. This would mean that you would be responsible for the package's compatibility and its updates to a higher extent than we would be, until we unvendor all recipes with Pyodide 0.28 (#4918).

ryanking13:
Let's merge this one. I'll continue unvendoring works.

PR #5252, opened 2024-12-11
Add highspy package

airen1986:
### Description

Highspy is python binding to HiGHS open source solver for large-scale LP/MIP/QP models

### Checklists

- [Y] Add a [CHANGELOG](https://github.com/pyodide/pyodide/blob/main/docs/project/changelog.md) entry
- [Y] Add / update tests
- [N] Add new / update outdated documentation


ryanking13:
`highspy` takes 14 mins to build in our CI without caching. Probably we need to postpone merging this PR until we unvendor packages to reduce the CI time.

akhilesh-bayer:
@agriyakhetarpal Could it be available in 0.27 release?

agriyakhetarpal:
@akhilesh-bayer, no, HiGHS takes a long time to build, as mentioned above. One way we could include it is if we build it separately, outside of Pyodide, and add a link to the wheel in the recipe file. Please see https://github.com/joemarshall/pyarrow-pyodide for an example.

airen1986:
@agriyakhetarpal OK, I can do that, but this will again break when 0.27 releases, as it will require packages to be built with 0.27 version? Or am I missing something?

agriyakhetarpal:
> @agriyakhetarpal OK, I can do that, but this will again break when 0.27 releases, as it will require packages to be built with 0.27 version? Or am I missing something?

Between 0.26 and 0.27, we have broken the NumPy ABI via #4925 but not the Emscripten ABI, since we did not upgrade to a version higher than 3.1.58. [Since `highspy` relies on NumPy at build time](https://github.com/ERGO-Code/HiGHS/blob/66f735e60ce4d4835edaa05a35f2bd4048969c54/pyproject.toml#L6), we would need to go through a pathway similar to this: https://github.com/pyodide/pyodide-numpy-2.0-rebuilds, and check out the latest `main` for the repository in the workflow instead of the `numpy-v2-mirror` branch so that NumPy 2.0 will be built.

ryanking13:

PR #5253, opened 2024-12-13
Unpin meson

ryanking13:
### Description

Applies https://github.com/pyodide/pyodide-build/pull/70 and see if it works.


ryanking13:
> Command 'emnm -j --export-symbols                                               
@scipy/stats/_unuran/unuran_wrapper.cpython-312-wasm32-emscripten.so.rsp'       
failed. Output to stderr was:                                                   
/root/repo/emsdk/emsdk/upstream/bin/llvm-nm: error: : unknown argument '-E'

Hmm, it seems like meson puts not only the object files but also the build flags into the response file, which is a valid behavi,or but it raisean s error for us.

What happens is, we pass the [response file to the `llvm-nm` to calculate the exports](https://github.com/pyodide/pyodide-build/blob/4d0b705e7d5cbeccffc77fe9a24bb9418ba5188e/pyodide_build/pywasmcross.py#L372), but inside the response file, it contains build flags such as `-sERROR_ON_UNDEFINED_SYMBOLS=0 -Wl,-O1 -shared -fPIC -lm -Wl,--version-script=/src/packages/scipy/build/scipy-1.14.1/scipy/_build_utils/link-version-pyinit.map` which llvm-nm does not consume.

I guess I should just block Meson from using the response file for Scipy...

ryanking13 (commented):
I'll go ahead and merge as CI passes.

A good news is that `meson` 1.6 added support for [LLVM flang](https://mesonbuild.com/Release-notes-for-1-6-0.html#support-for-llvmbased-flang-compiler). Last time when I tried LLVM flang + scipy had issues with meson, so I think I can try that again.

agriyakhetarpal:
> I'll go ahead and merge as CI passes.
> 
> A good news is that `meson` 1.6 added support for [LLVM flang](https://mesonbuild.com/Release-notes-for-1-6-0.html#support-for-llvmbased-flang-compiler). Last time when I tried LLVM flang + scipy, it had issues with meson, so I think I can try that again with new meson

This sounds great, thanks for sharing. After the 0.27 release, I shall try out building SciPy 1.15.0rc1 that was recently released with LLVM-Flang and this will hopefully reduce the number of patches and `seds` required, along with the module Fortran to Cython rewrites, enough for SciPy to have an out-of-tree build.

PR #5258, opened 2024-12-14
Upgrade zengl to 2.7.1 (#5258)

szabolcsdombi:
### Checklists

- [x] Add / update tests


ryanking13 (commented):
Could you please add `@pytest.mark.skip_pyproxy_check` decorator in the failing zengl test? The test failure means that there are some leaked proxies after running the test, but it is not a big deal. [See example here](https://github.com/pyodide/pyodide/blob/0065e612cfc8908b46d89c3320f24db4d7baeaff/packages/numpy/test_numpy.py#L116)

szabolcsdombi:
@ryanking13 before merging, I remember from last time the `if gl`: was necessary because running tests were not supporting a webgl2renderingcontext basically making the test only check if the import has any errors.

Should I dive deeper in running the tests?
Leaking a ref suggests me that we get a context and the leaked ref must be my `requestAnimationFrame` call or the stored ref to the webgl2renderingcontext.

btw, i have visual tests here: https://pyscript.com/@szabolcsdombi/collections/15a772b9-5c23-4eee-8018-aa465b31c83c

ryanking13:
> Should I dive deeper in running the tests?
Leaking a ref suggests me that we get a context and the leaked ref must be my requestAnimationFrame call or the stored ref to the webgl2renderingcontext.

It is up to you :) Normally, a leaked proxy won't be a big deal. But it would be worth investigating if you find any memory explosion when using the package.

ryanking13 (approved):
Thanks!

PR #5265, opened 2024-12-19
Use makefile instead of ninja for lightgbm

ryanking13:
### Description

I am trying to understand why the lightgbm build is frequently failing. This enables verbose logging and changes the ninja generator to makefile so it prints more logs.

Interestingly, after I changed the ninja to makefile, the build succeeded 4-5 times consecutively, but maybe I was just lucky...


agriyakhetarpal:
I think SciPy could be one of the issues. I cannot say with certainty, but in case SciPy is already built when LightGBM is triggered, it fails; if SciPy isn't, then then both packages succeed.

ryanking13:
> I think SciPy could be one of the issues. I cannot say with certainty, but in case SciPy is already built when LightGBM is triggered, it fails; if SciPy isn't, then then both packages succeed.

Interesting. The `*_tree_learner` modules that fail to build use `boost` internally, and Scipy does use `boost` too. So it may be related. But the usage of the `boost` library is hidden in the [`USE_GPU` flag](https://github.com/microsoft/LightGBM/blob/ee0131c554844084d6b4eb7fbdf76c069fc40ad2/src/treelearner/gpu_tree_learner.h#L27-L35), and we don't set that flag, so it will not be used...

If we think that scipy is a problem, maybe we can force lightgbm to be built before scipy...

hoodmane (approved):
Thanks @ryanking13!

agriyakhetarpal:
Ah, just noticed that the builds are passing here, but not in time – I opened #5266 to add the external LightGBM and PyArrow builds. Should we move ahead with your PR or mine? I am happy with either option.

agriyakhetarpal:
Updating this branch to see if the LightGBM build works with this generator with NumPy 2.0

agriyakhetarpal (approved):
It's passing without much trouble, thanks, @ryanking13! I'll drop LightGBM from #5266, and keep just the PyArrow wheel.

agriyakhetarpal:
Tested twice locally (and I tried my best to get a package that would also build SciPy in a different thread).

ryanking13:
Thanks @agriyakhetarpal! Let's re-use your wheel when this error happens again.

PR #5266, opened 2024-12-19
Add wheel builds for PyArrow, built with NumPy 2.0

agriyakhetarpal:
### Description

From https://github.com/pyodide/pyodide-numpy-2.0-rebuilds/releases/tag/v1.0.0 (temporary repository which will be archived later).

Tagging the recipe maintainers ~`@jameslamb` (LightGBM)~ and @joemarshall (PyArrow) for visibility.

>[!NOTE]
> Update to this PR: we can still make use of an external wheel for PyArrow, though, post #5265, I have dropped LightGBM's NumPy 2.0 build.

### Checklists


agriyakhetarpal:
Tests for LightGBM and PyArrow are passing here for Chrome and Node.js, as expected. However, I've removed LightGBM after its build was fixed with #5265, and added a CHANGELOG entry for PyArrow in 89144e87bc7f56b8e1de33bca5448001e5173f1a.

ryanking13 (approved):
Thanks!

agriyakhetarpal:
Thanks for the review!

jameslamb:
Thanks for the `@`, sorry it took a few days to respond. For the record, I would expect building `lightgbm` wheels with Ninja to generally work, but Unix Makefiles is totally fine.

I put up #5279 with some other recommended changes to the config for `lightgbm`.

agriyakhetarpal:
> Thanks for the `@`, sorry it took a few days to respond. For the record, I would expect building `lightgbm` wheels with Ninja to generally work, but Unix Makefiles is totally fine.
> 
> I put up #5279 with some other recommended changes to the config for `lightgbm`.

No worries, and thanks! Ninja works perfectly with building LightGBM standalone, except that we noticed that its build has some interference with either SciPy or some other uninvestigated package, and we decided to switch instead of spending time in an effort towards systematically investigating what the trouble is.

PR #5272, opened 2024-12-23
Fix a broken link in the docs

hoodmane:


agriyakhetarpal (approved):
Thanks! Would adding [Sphinx's linkcheck builder ](https://www.sphinx-doc.org/en/master/usage/configuration.html#options-for-the-linkcheck-builder) or a more general-purpose link checker (like [Lychee](https://lychee.cli.rs/)) make sense for our docs?

hoodmane:
Well this does make a warning at build time. I think we need to change the ci docs run so it fails if there are unexpected warnings. Running a general purpose link checker on the output is a good idea too but it wouldn't help in this case since here sphinx didn't emit a link.

hoodmane:
Well this does make a warning at build time. I think we need to change the ci docs run so it fails if there are unexpected warnings. Running a general purpose link checker on the output is a good idea too but it wouldn't help in this case since here sphinx didn't emit a link.

Issue #2577, opened 2022-05-18
Export additional Emscripten modules


kycutler:
## 🚀 Feature

Export Emscripten `Module` in a way that is reliable, or at least a few more of its exports such as `PATH` and `ERRNO_CODES`.

### Motivation

We use [BrowserFS](https://github.com/jvilk/BrowserFS) to help mount custom filesystems into Emscripten. However this requires the `PATH` and `ERRNO_CODES` exports from Emscripten in addition to `FS`.

Pyodide does not currently export these modules in a reliable manner. They can be accessed (currently) via `pyodide._module`, but there is no guarantee this will be reliable across releases.

### Pitch

Ideally, expose the entire Emscripten module in a reliable way so that all of its exports can be accessed. Alternatively, re-export `PATH` and `ERRNO_CODES`.

hoodmane:
> Ideally, expose the entire Emscripten module in a reliable way so that all of its exports can be accessed

This we won't do. The Module has tons of private stuff on it, some of it added by Emscripten, some of it added by us. None of the names on Module are guaranteed stable or intended as public API. Go ahead and use it if it works for your use case but we aren't ever offering stability guarantees about it.

> Re-export `PATH` and `ERRNO_CODES`.

Maybe. Can you give a code example showing how you use them? It's not clear to be from the BrowserFS docs that it is needed.

Due to a bug in Emscripten v2.0.28, the object `Module.ERRNO_CODES` is an empty object so even if we export it, it won't be helpful. We could add `Module.ERRNO_CODES = ERRNO_CODES` to `main.c` and that would probably fix it. I'm not sure if the bug has been fixed in more recent versions of Emscripten. But this illustrates why we don't want to make guarantees about this stuff.

Do you have a workaround to get access to the correct `Module.ERRNO_CODES`?

kycutler:
`PATH` is shown in the example in #613. Not sure why that example is missing the `ERRNO_CODES` but you can see it's used in the `EmscriptenFS` constructor [here](https://github.com/jvilk/BrowserFS/blob/a96aa2d417995dac7d376987839bc4e95e218e06/src/generic/emscripten_fs.ts#L384).

Interesting -- I hadn't noticed that ERRNO_CODES is empty. Basic errors still seem to work in my scenario so it's not clear exactly what functionality this breaks.

hoodmane:
In that case I am +1 for exposing `ERRNO_CODES` and `PATH`. 

For the record the bug looks like this:
```js
[...]

Issue #2147, opened 2022-01-30
pyodide_build buildpkg does not produce a .js file and does not report an error


paulftw:
## 🐛 Bug

I'm following official instructions on [creating a package](https://pyodide.org/en/stable/development/new-packages.html#building-the-package-and-investigating-issues) to build cadquery from github repo.

### To Reproduce

1. Build pyodide using official docker, install pyodide_build
2. Create `packages/cadquery/meta.yaml`:
```
package:
  name: cadquery
  version: "2.1"
source:
  url: https://github.com/CadQuery/cadquery/archive/refs/heads/master.zip

requirements:
  run:
    - ocp 7.5.3
    - hdf5 1.10.6 *1114
    - pyparsing >=2.1.9
    - ezdxf
    - ipython
    - typing_extensions
    - nptyping
    - nlopt
    - multimethod 1.6

test:
  imports:
    - cadquery

```

3. Run commands from documentation: 
```
export PYTHONPATH="$PYTHONPATH:/Users/myname/my-pyodide/pyodide-build/"
python -m pyodide_build buildpkg meta.yaml
cp build/*.data build/*.js ../../build/
```

Output of `cp`:
```
$ cp build/*.data build/*.js ../../build/
cp: build/*.data: No such file or directory
cp: build/*.js: No such file or directory
```

### Expected behavior

I expected to either get an error printed or a js file generated (or a mention in the documentation on why js isn't produced sometimes.)

Other details:

Output of `pyodide_build`:
```
$ python -m pyodide_build buildpkg meta.yaml
[2022-01-30 13:11:28] Building package cadquery...
running build
running build_py
creating build
creating build/lib
creating build/lib/cadquery
copying cadquery/cq_directive.py -> build/lib/cadquery
copying cadquery/units.py -> build/lib/cadquery
copying cadquery/cq.py -> build/lib/cadquery
copying cadquery/__init__.py -> build/lib/cadquery
copying cadquery/hull.py -> build/lib/cadquery
copying cadquery/selectors.py -> build/lib/cadquery
copying cadquery/types.py -> build/lib/cadquery
copying cadquery/cqgi.py -> build/lib/cadquery
copying cadquery/utils.py -> build/lib/cadquery
copying cadquery/assembly.py -> build/lib/cadquery
copying cadquery/sketch.py -> build/lib/cadquery
creating build/lib/cadquery/plugins
copying cadquery/plugins/__init__.py -> build/lib/cadquery/plugins
[...]

PR #5278, opened 2024-12-23
Update to typedoc 0.27

hoodmane:
And typescript 5.7. This gets us up to the latest version of typedoc and typescript. 

- [x] Manually check that there is no regression in documentation output


hoodmane:
Mostly the typedoc changes are not interesting but the following changelog notes look like they could potentially be used to replace stuff I've implemented in sphinx-js or sphinx-pyodide:

* Added support for JSDoc @hideconstructor tag. This tag should only be used to
  work around TypeScript#58653, prefer the more general @hidden/@ignore tag to
  hide members normally, #2577.
* Improved support for @param comments with nested object types, #2555.
* Improved support for @param comments which reference a type alias/interface. Important properties on the referenced type can now be highlighted with @param options.foo, which will result in the additional note being included under the documentation for that parameter, #2147. Note: This feature is limited to references. It is not supported on other types of types.
* Add support for an @expand tag which can be placed on type aliases and interfaces. When a type with @expand is referenced and TypeDoc has a place to include additional details about the type, the properties of the type will be included in the page where @expand is found. Note that use of this tag can significantly increase the size of your generated documentation if it is applied to commonly used types as it will result in inlining the comments for those types everywhere they are referenced, #2303.
* Add support for an @inline tag which can be placed on type aliases and interfaces. When a type with @inline is referenced, TypeDoc will resolve the referenced type and convert the type as if it was included directly within the referencing type. Note that use of this tag can significantly increase the size of your generated documentation if it is applied to commonly used types as it will result in inlining the comments for those types everywhere they are referenced, #2303.
* Introduced a new @useDeclaredType tag for type aliases which can sometimes improve their documentation, #2654.

hoodmane:
Well one unfortunate change here is that `Uint8Array` got replaced with `Uint8Array<ArrayBufferLike>`.

hoodmane:
It also shows `Iterable<T, any, any>` instead of `Iterable<T>`. I suppose generally the problem is that type variables with default values are not being omitted properly.

hoodmane:
And the `pyodide.canvas` apis lost their documentation strings.

hoodmane:
The `Uint8Array<ArrayBufferLike>` problem was fixed in 0.27.6
https://github.com/TypeStrong/typedoc/issues/2820

The canvas documentation comments went away in:
https://github.com/TypeStrong/typedoc/commit/d1be9569e556ceb3a162a653ab783d0729149f30

hoodmane:
Opened https://github.com/TypeStrong/typedoc/issues/2824 for the canvas issue, but we can work around it by changing canvas to use function declarations instead of `const` arrow functions.

PR #5279, opened 2024-12-24
expand lightgbm test, move more build settings into --config-settings

jameslamb:
### Description

Follow-up to #5265 and #5266

Proposes the following changes to `lightgbm` builds here:

* use `--config-settings` to set CMake verbosity instead of `sed`
* prefix config settings destined for `scikit-build-core` with `skbuild.`, to reduce the risk of conflicts with other build backends
  - ref: https://scikit-build-core.readthedocs.io/en/latest/configuration.html#configuration
* expand testing to cover more of the API:
  - opt out of overfitting protections that could result in a model of decision stubs (0-split trees) being trained
  - test model serialization (to custom text format and JSON) and deserialization (from custom text format)

### Checklists

- [x] Add / update tests

### Notes for Reviewers

This is my first contribution here, thanks in advance for your patience.

### DIFF
--- a/packages/lightgbm/meta.yaml
+++ b/packages/lightgbm/meta.yaml
@@ -5,10 +5,13 @@ source:
   url: https://files.pythonhosted.org/packages/4d/e6/41be1f8642257e21b4170e798c9a84e4268656ebfa3019586d82bfd281c9/lightgbm-4.5.0.tar.gz
   sha256: e1cd7baf0318d4e308a26575a63a4635f08df866ad3622a9d8e3d71d9637a1ba
 build:
-  backend-flags: cmake.define.USE_OPENMP=OFF
+  # skbuild.cmake.verbose should be switched to skbuild.build.verbose when
+  # upgrading to lightgbm>4.5.0 (the first release to set a minimum of scikit-build-core>=0.10)
+  backend-flags: |
+    skbuild.cmake.define.USE_OPENMP=OFF
+    skbuild.cmake.verbose=true
   exports: requested
   script: |
-    sed -i 's/build.verbose = false/build.verbose = true/g' pyproject.toml
     export CMAKE_GENERATOR="Unix Makefiles"
 requirements:
   run:
--- a/packages/lightgbm/test_lightgbm.py
+++ b/packages/lightgbm/test_lightgbm.py
@@ -9,13 +9,31 @@ def test_train_predict(selenium):
     import lightgbm as lgb
     import numpy as np
 
-    data = np.random.rand(50, 10)  # 50 entities, each contains 10 features
-    label = np.random.randint(2, size=50)  # binary target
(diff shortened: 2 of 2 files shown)


packages/lightgbm/meta.yaml
```diff
@@ -5,10 +5,13 @@ source:
   url: https://files.pythonhosted.org/packages/4d/e6/41be1f8642257e21b4170e798c9a84e4268656ebfa3019586d82bfd281c9/lightgbm-4.5.0.tar.gz
   sha256: e1cd7baf0318d4e308a26575a63a4635f08df866ad3622a9d8e3d71d9637a1ba
 build:
-  backend-flags: cmake.define.USE_OPENMP=OFF
+  # skbuild.cmake.verbose should be switched to skbuild.build.verbose when
+  # upgrading to lightgbm>4.5.0 (the first release to set a minimum of scikit-build-core>=0.10)
+  backend-flags: |
+    skbuild.cmake.define.USE_OPENMP=OFF
+    skbuild.cmake.verbose=true
```
agriyakhetarpal, line 12: Is it possible to set both options in LightGBM's `pyproject.toml` file instead? `pyodide-build` will not interfere with them when building. Also, `scikit-build-core` supports overrides for these options, so we can customise them on a per-platform basis: https://scikit-build-core.readthedocs.io/en/latest/overrides.html#env-string-or-bool

So, we could add something like

```toml
[tool.scikit-build]
cmake.define.USE_OPENMP = "ON"

[[tool.scikit-build.overrides]]
if.env.PYODIDE = true
cmake.verbose = true  # or build.verbose = true
inherit.cmake.define = "append"
cmake.define.USE_OPENMP = "OFF"
```

This could be something in consideration as a part of changes to resolve microsoft/LightGBM#5372, but would also be fine as a standalone build-system-only change so that we don't need to set these options defined ourselves. I'm happy to submit a PR, or step back if you might want to do this yourself. :)

agriyakhetarpal (approved):
Thanks for the contribution, @jameslamb! CI is passing as well. LGTM with one question:

packages/lightgbm/meta.yaml
jameslamb: Thanks for the suggestion.

I'd rather that `--config-settings` be used for this and that this configuration stay here in the `pyodide` repo for now. I don't support taking on a new part of the `scikit-build-core` interface (this override mechanism) just for this purpose.

I'm open to this possibility as a part of https://github.com/microsoft/LightGBM/issues/5372, like you mentioned, because then there would be a bigger benefit (not needing to keep the two files in sync, meaning that hopefully updates here could be just bumping a version number + SHA).

agriyakhetarpal: That makes sense; thanks for being open to it as a part of a greater change – that would be beneficial and would better explain the reason for adding this configuration.

PR #5281, opened 2024-12-27
Add package argon2-cffi

alarmfox:
### Description
Added `argon2-cffi` package including `argon2-cffi-bindings`. 

#### Support
Due to problems in threading the package works only when `parallelism=1` as shown below.
```py
import argon2

ph = argon2.PasswordHasher(parallelism=1)
hash = ph.hash("test")
assert ph.verify(hash, "test) is True
```

Currently I opened a [PR](https://github.com/hynek/argon2-cffi/pull/190) on the main repository which provides a check to validate parameters for `parallelism`

### Checklists

- [x] Add / update tests


packages/argon2-cffi-bindings/meta.yaml
```diff
@@ -0,0 +1,21 @@
...
+  url: https://files.pythonhosted.org/packages/b9/e9/184b8ccce6683b0aa2fbb7ba5683ea4b9c5763f1356347f1312c32e3c66e/argon2-cffi-bindings-21.2.0.tar.gz
+  sha256: bb89ceffa6c791807d1305ceb77dbfacc5aa499891d2c55661c6459651fc39e3
+about:
+  home: https://github.com/hynek/argon2-cffi-bindings
+  PyPI: https://pypi.org/project/argon2-cffi-bindings
+  summary: Low-level CFFI bindings for Argon2
+  license: MIT
+extra:
+  recipe-maintainers:
+    - alarmfox
```
agriyakhetarpal, line 16: Nitpicking, could you please move these fields to the end of the recipe (and similarly for other recipes)? We usually format the recipe with the order of "package metadata -> source -> requirements -> build -> about -> extra".

packages/argon2-cffi/meta.yaml
```diff
@@ -0,0 +1,22 @@
+package:
+  name: argon2-cffi
+  version: 23.1.0
+  top-level:
+    - argon2
+source:
+  url: https://files.pythonhosted.org/packages/a4/6a/e8a041599e78b6b3752da48000b14c8d1e8a04ded09c88c714ba047f34f5/argon2_cffi-23.1.0-py3-none-any.whl
+  sha256: c670642b78ba29641818ab2e68bd4e6a78ba53b7eff7b4c3815ae16abf91c7ea
```
agriyakhetarpal, line 8: Instead of including the wheel here, could we include the sdist and add your PR https://github.com/hynek/argon2-cffi/pull/190 as a patch with a short comment about it? Since `argon2-cffi` is pure Python, it shouldn't take more than ~2-3 seconds to build from the sdist, and we—or you, if you get to it first—will remember to drop the patch as soon as it gets merged and released.

agriyakhetarpal (commented):
Thanks for the contribution, @alarmfox! This looks good, I just have some general comments, plus I reviewed your PR upstream.

packages/argon2-cffi-bindings/meta.yaml
alarmfox: Done in the following commits

packages/argon2-cffi/meta.yaml
alarmfox: I moved to the sdist. I also added i run depenency for the binding one.
 
I also disabled a test because i couldn't (?) access the _pytest_ package to use the `pytest.raises` function. Any idea?

agriyakhetarpal: Ah, you may add the `import pytest` statement back there, and it should all work fine; I wonder if we need to add something about it to the docs, or whether this was something you faced locally?

packages/argon2-cffi/test_argon2-cffi.py
```diff
@@ -0,0 +1,30 @@
+import pytest
+from pytest_pyodide import run_in_pyodide
+
+
+@run_in_pyodide(packages=["argon2-cffi"])
+def argon2_cffi_helper(selenium):
```
hoodmane, line 6: If you're only going to call this function once, you may as well just inline it:
```suggestion
@run_in_pyodide(packages=["argon2-cffi"])
def test_argon2(selenium):
```

packages/argon2-cffi/test_argon2-cffi.py
```diff
@@ -0,0 +1,30 @@
...
+    import argon2
+
+    ph = argon2.PasswordHasher(parallelism=1)
+    hash = ph.hash("test")
+
+    assert ph.verify(hash, "test") is True
+
+    with pytest.raises(argon2.exceptions.UnsupportedParamsError) as exinfo:
```
hoodmane, line 14: You can't see the global scope from inside the `@run_in_pyodide` function so you'll need to import `pytest` inside the function rather than at top level.

packages/argon2-cffi/meta.yaml
alarmfox: I added back the test. For the docs, I added a message to the error providing context (i propagated the change also in the upstream PR). Now the user will have something like
```py
>>> import argon2
>>> ph = argon2.PasswordHasher(parallelism=2)
  File "/lib/python3.12/site-packages/argon2/_password_hasher.py", line 138, in __init__
    raise UnsupportedParamsError(msg)
argon2.exceptions.UnsupportedParamsError: within wasm/wasi environments `parallelism` must be set to 1
>>>
```
Maybe we should add something in the FAQ about the global problem something like

```
Q: Can i have multi-threading with pthread?
A: No, at the moment all libraries using plain multithreading will have to use 1 thread or add specific support for the platform
```

packages/argon2-cffi/test_argon2-cffi.py
alarmfox: Thanks i added this in the next commit

packages/argon2-cffi/meta.yaml
agriyakhetarpal: Thanks, the error message makes sense to me!

> Maybe we should add something in the FAQ about the global problem something like
> 
> ```
> Q: Can i have multi-threading with pthread?
> A: No, at the moment all libraries using plain multithreading will have to use 1 thread or add specific support for the platform
> ```

Yes, and we should add a small admonition to this section as well: https://pyodide.org/en/stable/development/new-packages.html#determining-if-creating-a-pyodide-package-is-necessary so that people check if the project they are trying to package will be eventually possible to support (in case it makes heavy use of `multiprocessing` or `threading`).

agriyakhetarpal:
Merged the latest changes from `main` to bring in #5285.

packages/argon2-cffi/meta.yaml
alarmfox: Thanks, I updated the documentation in the last commit. 
Marking this conversation as resolved.

docs/development/new-packages.md
```diff
@@ -17,6 +17,18 @@ Most pure Python packages can be installed directly from PyPI with
 {func}`micropip.install` if they have a pure Python wheel. Check if this is the
 case by trying `micropip.install("package-name")`.
 
+Due to [current limitations of webassemly](https://pyodide.org/en/stable/usage/wasm-constraints.html),
```
hoodmane, line 20: Could you separate the documentation update into a different PR?

alarmfox: Sure!

alarmfox: I reverted the last commit and added in https://github.com/pyodide/pyodide/pull/5287

agriyakhetarpal (approved):
Thanks, @alarmfox, looks great now! I agree with @hoodmane about splitting the docs changes in the new PR you opened.

agriyakhetarpal:
Tests are passing across Firefox, Chrome, and Node.js; let's merge

PR #5284, opened 2024-12-30
Fix and silence some Mypy failures

agriyakhetarpal:
### Description

This PR is trying to fix the failing pre-commit.ci CI job on recent PRs. I've fixed two comments and silenced some failures in `packages/netcdf4/test_netcdf4.py` by temporarily ignoring the file because there are many of them right now. I am not sure when or how they came up in the first place, since the file has been untouched for a while. dPlease let me know if you'd like me to fix them here, @hoodmane. Thanks!

### Checklists

N/A

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -82,7 +82,7 @@ repos:
         name: mypy-tests
         args: [--ignore-missing-imports]
         files: ^(packages/|docs|/conftest.py|src/tests)
-        exclude: (^packages/.*/setup.py|/src|^packages/aiohttp/aiohttp_patch.py$|^packages/zfpy/test_zfpy.py$)
+        exclude: (^packages/.*/setup.py|/src|^packages/aiohttp/aiohttp_patch.py$|^packages/zfpy/test_zfpy.py$|^packages/netcdf4/test_netcdf4.py$)
         additional_dependencies: *mypy-deps
 
   - repo: https://github.com/pre-commit/mirrors-prettier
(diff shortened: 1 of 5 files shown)


.pre-commit-config.yaml
```diff
@@ -82,7 +82,7 @@ repos:
         name: mypy-tests
         args: [--ignore-missing-imports]
         files: ^(packages/|docs|/conftest.py|src/tests)
-        exclude: (^packages/.*/setup.py|/src|^packages/aiohttp/aiohttp_patch.py$|^packages/zfpy/test_zfpy.py$)
+        exclude: (^packages/.*/setup.py|/src|^packages/aiohttp/aiohttp_patch.py$|^packages/zfpy/test_zfpy.py$|^packages/netcdf4/test_netcdf4.py$)
```
hoodmane, line 85: I wonder if it's possible to exclude files by putting a pragma at the top instead of listing them here.

hoodmane (approved):
Thanks @agriyakhetarpal !

.pre-commit-config.yaml
agriyakhetarpal: Ah, there is: https://mypy.readthedocs.io/en/stable/common_issues.html#ignoring-a-whole-file, will implement this change, except for the glob patterns.

hoodmane: Thanks!

agriyakhetarpal:
Thanks for the review!

PR #5287, opened 2024-12-30
Documentation: warning on multi-threading/processing

alarmfox:
### Documentation update on multi-threading/processing

This PR documents current limitation for multi-processing packages and shows  a workaround (paying a performance cost) as in https://github.com/pyodide/pyodide/pull/5281

### DIFF
--- a/docs/development/new-packages.md
+++ b/docs/development/new-packages.md
@@ -30,7 +30,7 @@ def _can_start_thread() -> bool:
 can_start_thread = _can_start_thread()
 
 if not can_start_thread:
-  # set n_threads = 1
+  n_threads = 1
 ```
 
 If there is no wheel on PyPI, but you believe there is nothing preventing it (it
--- a/docs/development/new-packages.md
+++ b/docs/development/new-packages.md
@@ -17,6 +17,18 @@ Most pure Python packages can be installed directly from PyPI with
 {func}`micropip.install` if they have a pure Python wheel. Check if this is the
 case by trying `micropip.install("package-name")`.
 
+Due to [current limitations of webassemly](https://pyodide.org/en/stable/usage/wasm-constraints.html),
+packages that make heavy use of multi-threading and multi-processing **may not** work
+without explicit support from maintainers. A workaround could be to avoid creating multiple
+threads when on wasm using a check like:
+
+```py
+is_wasm = sys.platform == "emscripten" or platform.machine() in ["wasm32", "wasm64"]
+
+if is_wasm:
+  # set n_threads = 1
+```
+
 If there is no wheel on PyPI, but you believe there is nothing preventing it (it
 is a Python package without C extensions):
 
--- a/docs/development/new-packages.md
+++ b/docs/development/new-packages.md
@@ -20,7 +20,7 @@ case by trying `micropip.install("package-name")`.
 Due to [current limitations of WebAssembly](https://pyodide.org/en/stable/usage/wasm-constraints.html),
 packages that make heavy use of multi-threading and multi-processing **may not** work
 without explicit support from maintainers. A workaround could be to avoid creating multiple
-threads when on wasm using a check like:
+threads when on WASM using a check like:
 
 ```py
 is_wasm = sys.platform == "emscripten" or platform.machine() in ["wasm32", "wasm64"]
--- a/docs/development/new-packages.md
+++ b/docs/development/new-packages.md
@@ -24,7 +24,7 @@ packages that use threading or multiprocessing will not work without a patch to 
 is_wasm = sys.platform == "emscripten" or platform.machine() in ["wasm32", "wasm64"]
 
 if is_wasm:
-  # set n_threads = 1
+  n_threads = 1
 ```
 
 If there is no wheel on PyPI, but you believe there is nothing preventing it (it
--- a/docs/development/new-packages.md
+++ b/docs/development/new-packages.md
@@ -17,10 +17,8 @@ Most pure Python packages can be installed directly from PyPI with
 {func}`micropip.install` if they have a pure Python wheel. Check if this is the
 case by trying `micropip.install("package-name")`.
 
-Due to [current limitations of WebAssembly](https://pyodide.org/en/stable/usage/wasm-constraints.html),
-packages that make heavy use of multi-threading and multi-processing **may not** work
-without explicit support from maintainers. A workaround could be to avoid creating multiple
-threads when on WASM using a check like:
+Because Pyodide [does not support threading or multiprocessing](https://pyodide.org/en/stable/usage/wasm-constraints.html),
+packages that use threading or multiprocessing will not work without a patch to disable it. For example,
 
 ```py
 is_wasm = sys.platform == "emscripten" or platform.machine() in ["wasm32", "wasm64"]
--- a/docs/development/new-packages.md
+++ b/docs/development/new-packages.md
@@ -19,12 +19,18 @@ case by trying `micropip.install("package-name")`.
 
 Because Pyodide [does not support threading or multiprocessing](https://pyodide.org/en/stable/usage/wasm-constraints.html),
 packages that use threading or multiprocessing will not work without a patch to disable it. For example,
+the following snippet will determine if the platform supports creating new threads.
 
 ```py
-is_wasm = sys.platform == "emscripten" or platform.machine() in ["wasm32", "wasm64"]
+def _can_start_thread() -> bool:
+    if sys.platform == "emscripten":
+        return sys._emscripten_info.pthreads
+    return platform.machine() not in ("wasm32", "wasm64")
 
-if is_wasm:
-  n_threads = 1
+can_start_thread = _can_start_thread()
+
+if not can_start_thread:
+  # set n_threads = 1
 ```
 
 If there is no wheel on PyPI, but you believe there is nothing preventing it (it
--- a/docs/development/new-packages.md
+++ b/docs/development/new-packages.md
@@ -17,7 +17,7 @@ Most pure Python packages can be installed directly from PyPI with
 {func}`micropip.install` if they have a pure Python wheel. Check if this is the
 case by trying `micropip.install("package-name")`.
 
-Due to [current limitations of webassemly](https://pyodide.org/en/stable/usage/wasm-constraints.html),
+Due to [current limitations of WebAssembly](https://pyodide.org/en/stable/usage/wasm-constraints.html),
 packages that make heavy use of multi-threading and multi-processing **may not** work
 without explicit support from maintainers. A workaround could be to avoid creating multiple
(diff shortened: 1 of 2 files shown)


docs/development/new-packages.md
```diff
@@ -17,6 +17,18 @@ Most pure Python packages can be installed directly from PyPI with
 {func}`micropip.install` if they have a pure Python wheel. Check if this is the
 case by trying `micropip.install("package-name")`.
 
+Due to [current limitations of webassemly](https://pyodide.org/en/stable/usage/wasm-constraints.html),
```
agriyakhetarpal, line 20:
```suggestion
Due to [current limitations of WebAssembly](https://pyodide.org/en/stable/usage/wasm-constraints.html),
```

docs/development/new-packages.md
```diff
@@ -17,6 +17,18 @@ Most pure Python packages can be installed directly from PyPI with
...
+packages that make heavy use of multi-threading and multi-processing **may not** work
+without explicit support from maintainers. A workaround could be to avoid creating multiple
+threads when on wasm using a check like:
```
agriyakhetarpal, line 23:
```suggestion
threads when on WASM using a check like:
```

docs/usage/faq.md
```diff
@@ -469,3 +469,19 @@ pyodide.runPython(`
     print(js_lambda, js_lambda_, js_lambda__) # 7, 7, 8
 `);
 ```
+
+## Can i use multi-threading/processing?
```
agriyakhetarpal, line 473:
```suggestion
## Can I use multi-threading/processing?
```

docs/usage/faq.md
```diff
@@ -469,3 +469,19 @@ pyodide.runPython(`
...
+
+Due to the limitation of current [wasm](https://pyodide.org/en/stable/usage/wasm-constraints.html)
+all packages that uses parallelism (eg. rely on _fork_ and _pthread_)
```
agriyakhetarpal, line 476:
```suggestion
all packages that use parallelism (eg. rely on _fork_ and _pthread_)
```

docs/development/new-packages.md
```diff
@@ -17,6 +17,18 @@ Most pure Python packages can be installed directly from PyPI with
 {func}`micropip.install` if they have a pure Python wheel. Check if this is the
 case by trying `micropip.install("package-name")`.
 
+Due to [current limitations of WebAssembly](https://pyodide.org/en/stable/usage/wasm-constraints.html),
+packages that make heavy use of multi-threading and multi-processing **may not** work
+without explicit support from maintainers. A workaround could be to avoid creating multiple
+threads when on WASM using a check like:
```
hoodmane, line 23: Maybe:
```suggestion
Because Pyodide [does not support threading or multiprocessing](https://pyodide.org/en/stable/usage/wasm-constraints.html),
packages that use threading or multiprocessing will not work without a patch to disable it. For example,
```

docs/development/new-packages.md
```diff
@@ -17,6 +17,18 @@ Most pure Python packages can be installed directly from PyPI with
...
+
+```py
+is_wasm = sys.platform == "emscripten" or platform.machine() in ["wasm32", "wasm64"]
```
hoodmane, line 26: It's always a bit annoying deciding how best to check this. `sys.platform == "emscripten"` implies `platform.machine() in ["wasm32", "wasm64"]` so it's a bit redundant. The Python test suite does:
```py
def _can_start_thread() -> bool:
    if sys.platform == "emscripten":
        return sys._emscripten_info.pthreads
    return platform.machine() not in ("wasm32", "wasm64")

can_start_thread = _can_start_thread()
```
it would be nice if threading/multiprocessing offered some API to query whether they work on the platform...

docs/development/new-packages.md
```diff
@@ -17,6 +17,18 @@ Most pure Python packages can be installed directly from PyPI with
...
+
+if is_wasm:
+  # set n_threads = 1
```
hoodmane, line 29: Since this is pseudocode anyways maybe we should just say:
```suggestion
  n_threads = 1
```

docs/usage/faq.md
```diff
@@ -469,3 +469,19 @@ pyodide.runPython(`
     print(js_lambda, js_lambda_, js_lambda__) # 7, 7, 8
 `);
 ```
+
+## Can I use multi-threading/processing?
```
hoodmane, line 473:
```suggestion
## Can I use threading/multiprocessing?
```

docs/usage/faq.md
```diff
@@ -469,3 +469,19 @@ pyodide.runPython(`
...
+
+Due to the limitation of current [wasm](https://pyodide.org/en/stable/usage/wasm-constraints.html)
+all packages that use parallelism (eg. rely on _fork_ and _pthread_)
+**may not work**. Accepting the performance degradation, a workaround
+consists of locking the number of threads down to `1` like shown below:
```
hoodmane, line 478:
```suggestion
No, fork and pthreads do not work in Pyodide. Attempts to use `threading`, `multiprocessing`, or `subprocess` will raise an `OSError`. You may be able to work around this by setting the number of threads to 1:
```

docs/usage/faq.md
```diff
@@ -469,3 +469,19 @@ pyodide.runPython(`
...
+
+```py
+is_wasm = sys.platform == "emscripten" or platform.machine() in ["wasm32", "wasm64"]
+
+if is_wasm:
+  # set n_threads = 1
+```
+
+Ultimately, developers will need to add specific support for the platform.
```
hoodmane, line 487: I don't think this sentence adds anything.
```suggestion
```

docs/usage/faq.md
agriyakhetarpal: Should we also mention that any attempts to use their APIs will cause fatal errors but simply `import threading`, etc. should not cause problems?

docs/development/new-packages.md
alarmfox: I agree. I don't know what `sys._emscripten_info.pthreads` returns as if i understand threads are supported but due to spectre they were disabled. If I try to compile `argon2-cffi` with `-lpthread` it gives no issue, but the error persists.

I like your function more than mine, but maybe we should just remove the "suggested workaround" to not make any wrong assumptions. What do you think?

docs/usage/faq.md
alarmfox: > Should we also mention that any attempts to use their APIs will cause fatal errors but simply `import threading`, etc. should not cause problems?

You're right. General info APIs and synchronous  work on the main thread, but I don't know how reliable the results are. For example this works:

```pycon
>>> current = threading.current_thread()
>>> current.name
'MainThread'
>>> current.is_alive()
True
>>> import threading
>>> def target (nums):
...     print(sum(nums))
... 
>>> t = threading.Thread(target=target, args = ([1,2,3],))
>>> t.run()
6
>>>
```

hoodmane: > any attempts to use their APIs will cause fatal errors

I don't believe that anything causes fatal errors. Just normal Python `OSError`s.

agriyakhetarpal: Yes

agriyakhetarpal: Maybe it's still better to guard the imports (or bring them into the functions where they are used) anyway.

alarmfox:
I tried to merge both suggestions keeping the  wasm-constraints link and changed the platform check

docs/usage/faq.md
```diff
@@ -469,3 +469,48 @@ pyodide.runPython(`
...
 ```
+
+## Can I use threading/multiprocessing/subprocess?
+
+No, fork and pthreads do not work in Pyodide (see more [here](https://pyodide.org/en/stable/usage/wasm-constraints.html)).
+Attempts to use `threading`, `multiprocessing`, or `subprocess` will raise a `RuntimeError`.
+You may be able to work around this by setting the number of threads to 1:
+
+```py
+def _can_start_thread() -> bool:
```
hoodmane, line 480: For some reason the `->` is marked as red here in the diff UI. I thought it meant there was a non ascii character but it seems to be normal...

docs/usage/faq.md
```diff
@@ -469,3 +469,48 @@ pyodide.runPython(`
...
+        return sys._emscripten_info.pthreads
+    return platform.machine() not in ("wasm32", "wasm64")
+
+can_start_thread = _can_start_thread()
+
+if not can_start_thread:
+  # set n_threads = 1
+```
+
+Bear in mind that you can still import the packages and use the general info API, but you cannot
```
hoodmane, line 491:
```suggestion
You can still import the packages and use the general info API, but you cannot
```

hoodmane (approved):
Thanks @alarmfox, I left a few more nits, and I still think the two code examples should just say `n_threads = 1`. Otherwise lgtm.

docs/usage/faq.md
```diff
@@ -469,3 +469,48 @@ pyodide.runPython(`
...
+>>> t = threading.Thread(target=target, args=([1, 2, 3], ))
+>>> t.run()
+6
+>>> t.start()
+Traceback (most recent call last):
+  File "<console>", line 1, in <module>
+  File "/lib/python312.zip/threading.py", line 994, in start
+    _start_new_thread(self._bootstrap, ())
+RuntimeError: can't start new thread
+>>>
```
hoodmane, line 515: Maybe delete this line:
```suggestion
```

docs/usage/faq.md
hoodmane: Do you see the red thing? Maybe it's a github bug? I think no action is needed.

alarmfox:
> Thanks @alarmfox, I left a few more nits, and I still think the two code examples should just say `n_threads = 1`. Otherwise lgtm.

Yes, I copied from a previous commit.

docs/development/new-packages.md
```diff
@@ -17,6 +17,22 @@ Most pure Python packages can be installed directly from PyPI with
...
+```py
+def _can_start_thread() -> bool:
+    if sys.platform == "emscripten":
+        return sys._emscripten_info.pthreads
+    return platform.machine() not in ("wasm32", "wasm64")
+
+can_start_thread = _can_start_thread()
+
+if not can_start_thread:
+  # set n_threads = 1
```
hoodmane, line 33:
```suggestion
  n_threads = 1
```

docs/usage/faq.md
alarmfox: Yes i can see the red arrow as well

hoodmane: _shrug_

hoodmane:
Thanks @alarmfox! We really appreciate it when people contribute documentation improvements.

hoodmane:
I backported this to the 0.27 branch so the changes should be visible on https://pyodide.org/en/stable/ in a few minutes.

PR #5289, opened 2024-12-31
Add the Online Laboratory for Climate Science and Meteorology to the list of related projects

juntyr:


docs/project/related-projects.md
```diff
@@ -31,6 +31,9 @@
 - [PyCafe](https://py.cafe) lets you host, edit, and share Python apps in your browser with a single click.
 - [quarto-live](https://github.com/r-wasm/quarto-live) uses Pyodide
   to create interactive Python code cells and exercises in [Quarto](https://quarto.org/) documents.
+- [Online Laboratory for Climate Science and Meteorology](https://github.com/climet-eu/lab) is a
+  JupyterLite-based environment that comes with additional packages from the weather and climate
+  community, and patches to support loading and working with large (remote) datasets.
```
agriyakhetarpal, line 36: I think we discussed some of your plans in https://github.com/climet-eu/lab/issues/10, but if there are more things that Pyodide could benefit from—such as on the "large (remote) datasets" aspect—at some point in time, they are more than welcome. :)

agriyakhetarpal (approved):
Thanks, @juntyr!

docs/project/related-projects.md
juntyr: I can give one example of a stack of quite hacky patches that bring a lot of user convenience but may have a better solution in the future.

At the moment one of the biggest enablers is a patch in fsspec that allows syncifying async filesystems (e.g. http). This normally requires a background thread, but it can be patched to use JSPI. At the moment, I cannot use JSPI yet, so instead every producer of "true" async (i.e. async in JS that we must await, or some asyncio combinators) is replaced with a sync variant that runs in a Python async method. Since those are just generators, I poll them once and if they resolve I can provide the syncified result of the async method (if they're still not resolved we have to give up and raise an exception).

Obviously, this will be much easier with JSPI, so once Pyodide feels confident in using it not just for optional improvements but for functionality, I will upstream my updated patch for fsspec. More generally, I'm really excited for JSPI and will try to help with making use of it inside Pyodide e.g. also for only lazy loading packages on imports (no more async install except for optional preloading) that will get Pyodide a lot closer to "just works as-is".

Anyways, happy New Year and thank you for all of your work in 2024 - it has been a joy to use Pyodide in my research project :)

agriyakhetarpal: That sounds good to me – thanks for the status update, and I wish you success in this new year, too!

PR #5178, opened 2024-11-09
Bump pillow-heif to 0.20.0 and pyheif 0.8.0

ryanking13:
supersedes #5087

PR #5292, opened 2025-01-01
Add CHANGELOG entry for #5178

teonbrooks:
Adding two missing upgrades to the changelog


agriyakhetarpal (approved):
Oops, thanks for the catch, @teonbrooks!

PR #5293, opened 2025-01-01
add: config.emscriptenSettings

warren-bank:
#### example of usage:

```javascript
  indexURL,
  emscriptenSettings: {
    wasmBinary,
    instantiateWasm: false
  }
})
```

### DIFF
--- a/src/js/emscripten-settings.ts
+++ b/src/js/emscripten-settings.ts
@@ -18,13 +18,14 @@ export interface EmscriptenSettings {
   readonly print?: (a: string) => void;
   readonly printErr?: (a: string) => void;
   readonly arguments: readonly string[];
-  readonly instantiateWasm?: (
+  readonly wasmBinary?: ArrayBuffer | Uint8Array;
+  readonly instantiateWasm?: false | ((
     imports: { [key: string]: any },
     successCallback: (
       instance: WebAssembly.Instance,
       module: WebAssembly.Module,
     ) => void,
-  ) => void;
+  ) => void);
   readonly API: API;
   readonly locateFile: (file: string) => string;
 
@@ -66,9 +67,11 @@ export function createSettings(config: ConfigType): EmscriptenSettings {
     // means dependency resolution has already failed and we want to throw an
     // error anyways.
     locateFile: (path: string) => config.indexURL + path,
-    instantiateWasm: getInstantiateWasmFunc(config.indexURL),
+    instantiateWasm: (config.emscriptenSettings?.wasmBinary ? false : getInstantiateWasmFunc(config.indexURL)),
   };
-  return settings;
+  return config.emscriptenSettings
+    ? Object.assign(settings, config.emscriptenSettings)
+    : settings;
 }
 
 /**
--- a/src/js/pyodide.ts
+++ b/src/js/pyodide.ts
@@ -54,6 +54,7 @@ export type ConfigType = {
   enableRunUntilComplete: boolean;
   checkAPIVersion: boolean;
   BUILD_ID: string;
+  emscriptenSettings?: EmscriptenSettings;
 };
 
 /**


hoodmane:
If we were to add such a setting I think it would be both permanently unstable and effectively marked as "if you use this and it breaks somehow it is likely to be marked won't fix" because there are so many ways to break everything with this. Maybe adding it as an undocumented setting with an underscore prefix would be helpful for people experimenting or for users who want a lot of control and don't mind being tightly coupled to upstream implementation details.

Ultimately, I think it would be nice instead to figure out what the actual use cases are and provide settings with less sharp edges to manage them. @ryanking13 has discussed the idea of adding startup hooks. I think we agree that this is a good idea but someone has to get around to working on it and I think we may not have a clear enough idea of what the downstream requirements.

warren-bank:
> If we were to add such a setting I think it would be both permanently unstable and effectively marked as "if you use this and it breaks..

imho, there's no such thing as idiot-proof..
and it's always better to offer too many levers than too few.
Having said that.. the devs will inevitably always get the blame. (c'est la vie)

Still, I probably should've made some attempt to protect the necessary attribute values.
Case in point, [`preRun`](https://github.com/pyodide/pyodide/blob/0.27.0/src/js/emscripten-settings.ts#L46) is pretty vital..
and being that it's an array of event listeners, any values provided by the user should be appended.

warren-bank:
off-topic observation.. what's up with these checks today?
Yesterday they all passed, and today a bunch are failing.
No changes to my code.. all I did was rebase.

hoodmane:
If you look in CI you'll see that there are some typescript errors:
```
emscripten-settings.ts:77:21 - error TS2339: Property 'push' does not exist on type 'readonly PreRunFunc[]'.

77     settings.preRun.push(...config.emscriptenSettings.preRun);
                       ~~~~

emscripten-settings.ts:78:12 - error TS2704: The operand of a 'delete' operator cannot be a read-only property.

78     delete config.emscriptenSettings.preRun;
              ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

warren-bank:
> If you look in CI you'll see that there are some typescript errors:

argh..
well, my earlier comment (about failed checks) was made before I added this 2nd commit..
what you referenced is obviously my bad; I hate Typescript so much..
any ideas for how to combine 2 objects with read-only attributes (elegantly)?
And by elegantly.. I mean without needing to create a 3rd object and iterate through keys.

hoodmane:
Well before we get into this, how about instead adding `wasmBinaryPromise` and `wasmBinaryResponse` options to the settings? It seems like that would be good enough for your use case?

hoodmane:
Also, how does lockfile loading work for you? Maybe we should also add `lockFilePromise` to the public settings? That would be more flexible than `lockFileUrl`.

hoodmane:
Hmm and also `stdLibPromise` instead of `stdLibUR`...

warren-bank:
well.. I just pushed what I think will fix the TS issue.
It creates a new object using spread operators.
Essentially, treating both objects as immutable.

Without diving back into the code to see what `wasmBinaryPromise` and `wasmBinaryResponse` are..
I don't remember seeing those..
I can tell you that all of the `*Url` attributes work great with `data:` Uris.
And passing a `WasmBinary` as an UInt8Array also works great.

warren-bank:
hmm.. I just checked `pyodide.asm.js` in the release..
a summary of the relevant code is:
```javascript
var _createPyodideModule = (() => {
  return function(moduleArg = {}) {
    var Module = Object.assign({}, moduleArg);

    var wasmBinary;
    if (Module["wasmBinary"]) wasmBinary = Module["wasmBinary"];
  }
})();
```

..where `moduleArg` is the [`EmscriptenSettings` type](https://github.com/pyodide/pyodide/blob/0.27.0/src/js/emscripten-settings.ts#L12),
the value of which.. this PR is updating

hoodmane:
What I mean is that here:
https://github.com/pyodide/pyodide/blob/main/src/js/emscripten-settings.ts?plain=1#L190
it expects to get either Promise for a `Response` or a Promise for a `Uint8Array`. I think we could add a new setting to `loadPyodide` called `wasmBinaryPromise` and then pass the value in here.

warren-bank:
oh.. it looks like the update fixed the Typescript;
`readthedocs` passed, which seems like a good indicator.

but again, so many of the checks are failing..
with error:
```text
ERROR: ./pyodide-build is not a valid editable requirement. It should either be a path to a local project or a VCS URL (beginning with bzr+http, bzr+https, bzr+ssh, bzr+sftp, bzr+ftp, bzr+lp, bzr+file, git+http, git+https, git+ssh, git+git, git+file, hg+file, hg+http, hg+https, hg+ssh, hg+static-http, svn+ssh, svn+http, svn+https, svn+svn, svn+file).
```

hoodmane:
Okay it was introduced by 00271f9, will fix.

hoodmane:
CI should be fixed now if you merge/rebase on main

warren-bank:
my concern.. and I'm not 100% sure how the pieces fit together on this.. I'd need to carefully trace the code,
but my concern is.. in "pyodide.asm.js":

```javascript
            var wasmBinaryFile;
            wasmBinaryFile = "pyodide.asm.wasm";
            if (!isDataURI(wasmBinaryFile)) {
                wasmBinaryFile = locateFile(wasmBinaryFile)
            }

            function getBinarySync(file) {
                if (file == wasmBinaryFile && wasmBinary) {
                    return new Uint8Array(wasmBinary)
                }
                if (readBinary) {
                    return readBinary(file)
                }
                throw "both async and sync fetching of the wasm failed"
            }
```

if `wasmBinary` is defined, then no network request.
otherwise, XHR for "pyodide.asm.wasm".

hoodmane:
The control flow is super confusing but that is never executed if `instantiateWasm` is passed as a setting. Try setting a breakpoint there.

warren-bank:
you raise an interesting question..

* you're using `instantiateWasm` to call `getInstantiateWasmFunc()` which eventually calls `WebAssembly.instantiate(buffer)`
* I'm overriding this param.. setting it to `false`.. so `WebAssembly.instantiate(buffer)` isn't called in this way
* I know anecdotally, because it works, that my test page must be calling it indirectly..
  - presumably when `_createPyodideModule(args)` is called,<br>but I'm not 100% sure.. I'll need trace that a bit to see exactly what's going on

warren-bank:
here is a pretty telling summarized trace through the code in "pyodide.asm.js":

```javascript
  function createWasm() {
      //...

      if (Module["instantiateWasm"]) {
          try {
              return Module["instantiateWasm"](info, receiveInstance)
          } catch (e) {
              err(`Module.instantiateWasm callback failed with error: ${e}`);
              readyPromiseReject(e)
          }
      }
      instantiateAsync(wasmBinary, wasmBinaryFile, info, receiveInstantiationResult).catch(readyPromiseReject);
      return {}
  }

  function instantiateAsync(binary, binaryFile, imports, callback) {
      if (!binary) {
          //...
      }
      return instantiateArrayBuffer(binaryFile, imports, callback)
  }

  function instantiateArrayBuffer(binaryFile, imports, receiver) {
      return getBinaryPromise(binaryFile).then(binary => WebAssembly.instantiate(binary, imports)).then(receiver, reason => {
          err(`failed to asynchronously prepare wasm: ${reason}`);
          abort(reason)
      })
  }

  function getBinaryPromise(binaryFile) {
      if (!wasmBinary) {
          //...
      }
      return Promise.resolve().then(() => getBinarySync(binaryFile))
  }

  function getBinarySync(file) {
      if (file == wasmBinaryFile && wasmBinary) {
          return new Uint8Array(wasmBinary)
      }
      if (readBinary) {
          return readBinary(file)
      }
      throw "both async and sync fetching of the wasm failed"
  }
```

notes:
* `createWasm()` is the only function that calls `instantiateWasm()`
  - this is where your hook gets called
  - my code skips this hook, and continues..
    * `getBinarySync` returns `wasmBinary`
    * `instantiateArrayBuffer` calls `WebAssembly.instantiate(wasmBinary)`
    * ???
    * profit!

hoodmane:
That approach is worse because you have to block for the binary before work starts in `loadPyodide`. `instantiateWasm` allows the work in `loadPyodide()` and in the Emscripten setup to proceed while waiting for the promise to resolve and only blocks for it when it actually needs the binary.

hoodmane:
This is the same disadvantage of using a `data:` url for `lockFileURL` rather than having a `lockFilePromise` setting: to make the `data:` url you have to first get the lock file, whereas you wouldn't have to do that if we had a `lockFilePromise` setting.

warren-bank:
so.. now that I have a handle on the control flow, I can respond to your idea..

* on the one hand..
  - if the `getInstantiateWasmFunc` is passed a `wasmBinary`..
    then the callback function that it returns would have this value in scope..
    and when the hook is called.. it could bypass the call to `getBinaryResponse`
* on the other hand..
  - I don't see any functionality being added by using this hook function, over using the default handler
  - the only thing that stands out is:
    ```javascript
        if (typeof WasmOffsetConverter !== "undefined") {
          // @ts-ignore
          wasmOffsetConverter = new WasmOffsetConverter(wasmBinary, module);
        }
    ```
  - this code confuses me:
    - the `wasmOffsetConverter` variable doesn't appear to be defined.. so unless I'm missing something
      * in a `strict` scope, this would throw an error
      * otherwise, its scope gets hoisted
    - the `wasmBinary` variable also appears to be undefined

warren-bank:
regardless of the name given to variables..
you're making the (valid) point that all of these values should accept either the corresponding data (ie: string or buffer, etc)..
_or_ a Promise that will resolve to it..
which gives added flexibility (and potentially improved performance) to the user..
and requires nearly no added code.. only an: `val = await val`
so.. yeah, I agree.. that either type should be accepted as parameter values.

hoodmane:
`WasmOffsetConverter` is an implementation detail for asan. In asan builds it's declared here:
https://github.com/emscripten-core/emscripten/blob/main/src/preamble.js?plain=1#L684

hoodmane:
Though the code behind there is wrong as you say because it should be `binary` not `wasmBinary` and we need to handle the `response` case.  We should either make a smoke test for asan builds so that we know that it works or get rid of this.

hoodmane:
> I don't see any functionality being added by using this hook function, over using the default handler

The advantage is that it starts the request sooner. Note that `getBinaryResponse()` starts the request in `getInstantiateWasmFunc()`, not in the callback. The default handler initiates the request itself, so by default we would do:

1. download `pyodide.mjs`, parse and run
2. `pyodide.mjs` downloads `pyodide.asm.js` then we parse and run it (it's a fair amount of JS so parsing takes a while)
3. `pyodide.asm.js` downloads `pyodide.asm.wasm` (also quite big).

With the callback,  instead we do:
1. download `pyodide.mjs` parse and run
2. simultaneously download `pyodide.asm.js` and `pyodide.asm.wasm`

This significantly improves startup performance.

warren-bank:
> 2. simultaneously download `pyodide.asm.js` and `pyodide.asm.wasm`
> 
> This significantly improves startup performance.

thanks for explaining.. now I totally get it.
[clever](https://github.com/pyodide/pyodide/blob/0.27.0/src/js/emscripten-settings.ts#L190-L191)..
* you start an async download and then immediately return a function..
* later, when this function gets called by the event hook.. the download promise is resolved

this optimization isn't relevant to what I was working on..
so I didn't even realize the timing of it..
but that's a very clever design!

ryanking13:
> If we were to add such a setting I think it would be both permanently unstable and effectively marked as "if you use this and it breaks somehow it is likely to be marked won't fix" because there are so many ways to break everything with this.

+1 . I don't want to expose this as a public API either, because we can't guarantee anything about it.

>  @ryanking13 has discussed the idea of adding startup hooks. 

> In order to prevent requiring the userscript to redownload everything (Pyodide js+wasm, .whl files, .zip files, etc) every time an extractor is triggered to run (to obtain the video URL only for the current web page.. to avoid CORS issues).. all of these files need to be stored by the userscript (in its metdata block) as either requirements (javascript) or resources (binary). The challenge was to initialize Pyodide and load a collection of packages from this storage cache.

Yes, this is exactly the thing that I wanted to solve with the _hook_ feature(https://github.com/pyodide/pyodide/pull/3664).

PR #5294, opened 2025-01-02
add: pyodide.loadBinaryPackage(name, buffer)

warren-bank:
sorry, that this branch also includes the commit belonging to #5293


src/js/load-package.ts
```diff
@@ -276,6 +276,33 @@ export class PackageManager {
...
+        file_name: name + '-1.0.0-py3-none-any.' + (format || 'whl'),
+        install_dir: 'site',
+        sha256: '',
+        package_type: 'package',
+        imports: [],
+        depends: []
+      }
+    };
+
+    await this.installPackage(metadata, new Uint8Array(buffer));
```
hoodmane, line 303: If `buffer` is a `Uint8Array` this copies. You want `API.typedArrayAsUint8Array` https://github.com/pyodide/pyodide/blob/main/src/core/pre.js?plain=1#L86

hoodmane:
It looks like the actual feature commit here would rebase cleanly onto the main branch. Please do `git rebase HEAD~1 --onto main` and then force push.

src/js/load-package.ts
warren-bank: great suggestion! ..thanks

in the process of removing an earlier commit belonging to a different PR..
I (effectively) rebased to the current HEAD of main.
You guys move quickly :)

hoodmane:
Here the failure is:
```
ERROR: ./pyodide-build is not a valid editable requirement. It should either be a path to a local project or a VCS URL (beginning with bzr+http, bzr+https, bzr+ssh, bzr+sftp, bzr+ftp, bzr+lp, bzr+file, git+http, git+https, git+ssh, git+git, git+file, hg+file, hg+http, hg+https, hg+ssh, hg+static-http, svn+ssh, svn+http, svn+https, svn+svn, svn+file).
```
Looks like it means we didn't do a recursive git checkout... I don't understand what the problem is on this PR.

hoodmane:
I think we can agree that something like this is useful. Let's see what @ryanking13 thinks of it.

ryanking13:
Could you please explain more about the use cases for this API? I think we need more discussion about this.

warren-bank:
My particular use case involves storing large files (ex: packages) into a cache, and then retrieving this data from cache as an ArrayBuffer. I made various attempts to load these packages using existing API methods, but nothing worked. This patch makes the loading process possible.. and simple.

I assume others (in the wild) will encounter their own use cases. I can't even guess what they could be.

For a concrete example, [here](https://warren-bank.github.io/single-page-apps/yt-dlp-web-console/index.html) are my attempts to run [yt-dlp](https://github.com/yt-dlp/yt-dlp) in Pyodide, where __all__ data is loaded from "cache"; note that the first link is always the most recent attempt. I put "cache" in quotes, because this is only a demo.. and in this demo the data is first loaded by `fetch()` into a data structure that hold everything in ArrayBuffers.. and __then__ proceeds with loading Pyodide from this data structure. When I get this demo working properly, then my intention is to migrate the code into a __userscript__.. where all of the data in this data structure is stored by the userscript manager into storage that is made available to the userscript at runtime through GM_* APIs.

In [this most recent attempt](https://warren-bank.github.io/single-page-apps/yt-dlp-web-console/index.pyodide-0.27.0-fork-2.0.0.html), Pyodide is loading everything perfectly. In doing so, my project has encountered a [new issue](https://github.com/python/cpython/issues/86228). But that's not pertinent to this discussion.

ryanking13:
We have an API for installing a wheel file: [`pyodide._api.install`](https://github.com/pyodide/pyodide/blob/c2a74e79a75626e1878bd62f85d3cb3ebc336e25/src/js/installer.ts#L23-L29), it is responsible for installing a wheel package into the file system, and do any required stuff. For now, this API is private, but I plan to make it public after the interface is stabilized.

Will this API be do the job for you? I think it mostly does the same job as `loadBinaryPackage` that you added in this PR. Note that `loadPackage internally uses [this API](https://github.com/pyodide/pyodide/blob/c2a74e79a75626e1878bd62f85d3cb3ebc336e25/src/js/load-package.ts#L454-L461).

hoodmane:
It seems to me that exposing `install` is probably one reasonable approach, though maybe when we do that we should think about including the call to `get_install_dir` in `install`? Or would users of the public API not want to do that?

But @ryanking13 I wonder if the best approach for this use case would be to add a setting that overrides `downloadPackage`. Then it would be possible to override that to look up the package buffer in a map and still use the normal dependency resolution logic.

ryanking13:
> It seems to me that exposing install is probably one reasonable approach, though maybe when we do that we should think about including the call to get_install_dir in install? Or would users of the public API not want to do that?

Yeah, I don't want to expose `get_install_dir`. Instead, users can put any directory to the `installDir`. If they want to install to the `site-packages`, they can do `site.getsitepackages()` and pass it to `install` function.

> But @ryanking13 I wonder if the best approach for this use case would be to add a setting that overrides downloadPackage. Then it would be possible to override that to look up the package buffer in a map and still use the normal dependency resolution logic.

Makes sense. But overriding downloadpackage sounds a bit hacky to me. Maybe the better way would be to expose our dependency resolution logic. There was a discussion about exposing access to the lock file (I don't remember the issue number...) to the users. So maybe we can combinelock file access + dependency resolution and expose it as a public API.

hoodmane:
If you look at `downloadPackage()` it already has a bunch of custom logic: sub-resource hashes has a toggle, in node we cache it in a directory. With a hook, people could implement their own different caching behavior (say they want to cache in the browser), they could turn off SRI for a specific package. Or whatever else. I think the fact that it has a simple API and the existing implementation is so complicated makes me think it could use a hook.

ryanking13:
Yes, providing a hook may help people reuse the functionalities that we've already implemented, such as caching and SRI. But I don't quite think it is too complex that people cannot implement themselves.

Rather, I think that providing downloadPackage as a hook makes it dependent on the interface we provided, such as how people should do caching or validation. I like to think of it as a separation of responsibilities. People are free to choose where and how to get external packages, how to cache them, and how to validate them. What they need to do is properly pass the ArrayBuffer to the `install` function that we provide.

hoodmane:
Well of course we can expose `install()`, maybe the user-facing version of it calls `reportUndefinedSymbols()` and `invalidate_caches()` on every call until we can remove those entirely. Then people are free to reimplement the whole thing. But if they do this, the standard `loadPackage()` and `loadPackagesFromImports()` won't work correctly, if they try to install something from `micropip` and it tries to call out to `loadPackage()` that also won't work correctly. The advantage of a hook is that the person setting up the runtime can change the part of the logic they care about while keeping all the standard APIs working.

I think it's a good idea to standardize and expose the pieces of this, like compare `importlib`. But most existing import systems like `importlib` and node's `module.registerHook` work by defining stages (searching for a module, then executing the module) and allowing the user to specify hooks to modify the stages. That way, whoever sets up the runtime can modify these hooks (pytest for example changes the execution stage with an ast rewrite) and it is mostly transparent to the code.

Of course, this isn't an import system. I would still argue that we're better off introducing hooks in logic that people want to swap out a lot rather than only exposing all the pieces and allowing people to build a replacement. No reason not to do both though as long as the pieces can have their sharp edges rounded off enough.

ryanking13:
Yeah, makes sense. If we can design an API/hook well enough and in a stable enough form, I think it's a good idea to make it easy for people to use. My concern was that if we were rushing to expose an interface when it wasn't yet stable enough, this would often result in breaking changes or slowing down our improvements.

So let's take the approach a little slower. I think there's a lot we can learn from the work on modifying micropip (https://github.com/pyodide/micropip/issues/112), and I believe we can refactor `loadPackage` as well while working on it.

hoodmane:
Agreed about not rushing. Though I'd also like to see us adopt an explicit stabilization policy so we can make a feature and document it as unstable, change for a couple of releases and then stabilize it. As opposed to always having to get everything right the first time.

PR #5295, opened 2025-01-03
Cleanup to Emscripten setup for asan builds

### LINKED ISSUE #5290: [observation] I could be wrong, but I think I spot an undefined variable..
* file: [emscripten-settings.ts](https://github.com/pyodide/pyodide/blob/0.26.4/src/js/emscripten-settings.ts#L212)
* function: _getInstantiateWasmFunc_
* variable: `wasmBinary`
  - this is the only mention of this variable name in the file
  - is it defined?
  - if not, should it instead be?: `await binary`

Sorry if I'm wrong. I didn't run any tests. This is only a quick drive-by observation, as I was reading some code.

hoodmane:
Previously, in asan builds we attempted to use an undefined variable in instantiateWasm so we would crash on startup. This fixes it.

Resolves #5290


ryanking13 (approved):
Thanks!

Issue #5268, opened 2024-12-20
llvm-flang + scipy progress tracker


ryanking13:
This issue is to track the progress in building Fortran packages using llvm-flang. The primary target is Scipy.

- llvm-flang
  - [x] included in the LLVM toolchain.
 
- WASM backend for llvm-flang
  - [ ] not officially supported (https://discourse.llvm.org/t/flang-web-assembly/61607).
  - [x] r-wasm team forked llvm and enabled WASM backend (https://github.com/r-wasm/flang-wasm)
  - [x] I used that fork to make a prebuilt binary using the same emscripten version in Pyodide (https://github.com/ryanking13/flang-llvm-emscripten)

- llvm-flang in Emscripten toolchain
  - [ ] flang is not integrated in the Emscripten toolchain, obviously, as no WASM backend supported yet.

- meson
  - [x] Added initial support for llvm-flang since v1.6.0 (https://github.com/mesonbuild/meson/pull/13323)
- scipy
  - [x] https://github.com/scipy/scipy/pull/21917

---

Error logs

__2024/12/20__

> LLVM ERROR: SmallVector unable to grow. Requested capacity (18446744072885119280) is larger than maximum value for size type (4294967295)

Hitting this error when compiling `PROPACK` codes. The each number in the log represents `2^64` and `2^32`, so it is certain that somehow it is related to the 32-bit, 64-bit variables (architectures), but it is unclear how to handle it.

<details><summary>Full error log</summary>
<p>

LLVM ERROR: SmallVector unable to grow. Requested capacity (18446744072885119280) is larger than maximum value for size type (4294967295)
PLEASE submit a bug report to https://github.com/llvm/llvm-project/issues/ and include the crash backtrace.
Stack dump:
[...]

PR #5297, opened 2025-01-03
SciPy 1.15.0

agriyakhetarpal:
### Description

This PR bumps the in-tree SciPy version in the Pyodide distribution to version 1.15.0, which was released a few hours ago: https://github.com/scipy/scipy/releases/tag/v1.15.0.

Here are my notes:
- A lot of troublesome FORTRAN77 code has been removed, notably: MINPACK, QUADPACK, LBFGSB, and id_dist. We had already applied a patch for the last package, which can now be dropped, including two other patches. The problem of not being able to use `f2py` directly via Meson generators remains, and I have to resort to using `python -m numpy.f2py` until pyodide/pyodide-build#21 is resolved.
- We also have a list of tasks to do for using LLVM-Flang for SciPy via #5268, similar to the effort by our friends at emscripten-forge/recipes#1288, but maybe it's easier to do this update first and then proceed with that one.
- I also have #5060 which was stalled at the time due to https://github.com/scipy/scipy#21290 not having made it to a release yet, which can be returned to after this update.

SciPy was previously updated in #5031 to version 1.14.1.

As always, cc @rgommers and the package maintainers @lesteve and @steppi as listed in the recipe.

### Checklists

- [ ] Add / update tests


agriyakhetarpal:
I've opened this as a quick PR, though I'll note that patch 10 (for `cython_lapack_signatures.txt`) still seems to cause me trouble (locally, at least). Let me see if it fails in CI as well.

agriyakhetarpal:
https://github.com/scipy/scipy/pull/21917 will probably need to be extended for WASM as well, but we use `gfortran` so I'm not quite sure what the fix would be... 

From the logs:

`Fortran compiler for the host machine: /tmp/tmpqvmfabb4/gfortran (gcc 12.2.0 "GNU Fortran (Debian 12.2.0-14) 12.2.0")`

rgommers:
> [scipy/scipy#21917](https://github.com/scipy/scipy/pull/21917) will probably need to be extended for WASM as well, but we use `gfortran` so I'm not quite sure what the fix would be...

It's just warnings, so this shouldn't be blocking I'd think. The compiler ID may be `emscripten` rather than `gcc`?

agriyakhetarpal:
So, the problem is coming in calls in ARPACK (which is now at version 3.9.1) – like:

```fortran
call ivout (logfil, 1, [nev], ndigit,
```

where array constructors like `[nev]` raise a syntax error, but scalar variables like `nev` itself are completely fine. This is most likely because the former syntax was introduced in Fortran 90 (or 2003?) and https://github.com/hoodmane/f2c is applicable only for FORTRAN 77? I don't think it is a good idea to patch ARPACK extensively, so maybe there's a flag we can use to fall back to the old(er) syntax?

https://github.com/ElmerCSC/elmerfem/issues/239 seems to have some insights here, which we could explore.

agriyakhetarpal:
> It's just warnings, so this shouldn't be blocking I'd think. The compiler ID may be `emscripten` rather than `gcc`?

Yes, that should be the case. I guess `ff.get_id() == 'gcc' or `ff.get_id() == 'emscripten'` should work.

rgommers:
Ah that makes sense. If the problem is `f2c` not supporting the syntax, than no GCC flag (`-std=legacy` or otherwise) is going to fix that I think - it either needs patching of the Fortran source, or a change to `f2c`. Maybe you can try locally to figure out how extensive the problem is @agriyakhetarpal? If it's only a few files then I'd say just patch it. If it's all over the place, then we may want to think twice (ARPACK is already in the process of being rewritten after all).

hoodmane:
f2c works on most fortran 90 code but not all of it. If the fortran libraries have code that requires newer features we should remove it or disable the library until we get flang working.

hoodmane:
Or I guess we could add more features to f2c I already added a handful of things so one more wouldn't hurt too much.

agriyakhetarpal:
Ralf and I discussed this on 06/01/2025: while the number of files to be patched is large (~40), the patches to them are quite minimal, so b1f2a063c21cfd0951082af2c702a5094126484f patches ARPACK. If this doesn't work with more significant changes, then we can consider modifying `f2c` and supporting the new syntax.

packages/scipy/patches/0016-Drop-unsupported-array-constructors-in-ARPACK.patch
```diff
@@ -0,0 +1,1255 @@
...
+
+diff --git a/scipy/sparse/linalg/_eigen/arpack/ARPACK/SRC/cgetv0.f b/scipy/sparse/linalg/_eigen/arpack/ARPACK/SRC/cgetv0.f
+index c231eadcb..fc4b24b92 100644
+--- a/scipy/sparse/linalg/_eigen/arpack/ARPACK/SRC/cgetv0.f
++++ b/scipy/sparse/linalg/_eigen/arpack/ARPACK/SRC/cgetv0.f
+@@ -361,9 +361,9 @@ c     | Check for further orthogonalization. |
+ c     %--------------------------------------%
+ c
+       if (msglvl .gt. 2) then
+-          call svout (logfil, 1, [rnorm0], ndigit,
```
hoodmane, line 65: Rather than applying a huge repetitive patch I'd prefer to fix this in `_f2c_fixes` or in f2c itself.

juntyr:
@agriyakhetarpal What's the status of this PR?

agriyakhetarpal:
Hi @juntyr, thanks for asking about it. The last time I worked on it, I stumbled on an issue related to the ARPACK update that wasn't fixable just by reverting to the older 3.3.0 syntax (the logs still exist here). I'll explore the `_f2c_fixes` approach @hoodmane has suggested in a short while.

agriyakhetarpal:
Hi all, this PR will be superseded by pyodide/pyodide-recipes#141, which bumps to SciPy 1.16.0. It passes the tests, and is the last rung in a series of PRs I've opened today. I'd like to request that we shift our further discussions there instead. Thank you!

PR #5301, opened 2025-01-06
Remove cffi from cross-build package

ryanking13:
I have a doubt about cffi being a cross build package. We set cffi as a cross-build package to make sure the build time cffi version is equal to the runtime cffi version, but is it a really necessary requirement?

In native environment, package built with cffi version A should be able to run with other cffi versions as ABI compatibility is guaranteed (if not, it is cffi's problem, not ours).

### DIFF
--- a/packages/cffi/meta.yaml
+++ b/packages/cffi/meta.yaml
@@ -9,8 +9,6 @@ requirements:
 source:
   url: https://files.pythonhosted.org/packages/fc/97/c783634659c2920c3fc70419e3af40972dbaf758daa229a7d6ea6135c90d/cffi-1.17.1.tar.gz
   sha256: 1c39c6016c32bc48dd54561950ebd6836e1670f2ae46128f67cf49e789c52824
-build:
-  cross-build-env: true
 about:
   home: http://cffi.readthedocs.org
   PyPI: https://pypi.org/project/cffi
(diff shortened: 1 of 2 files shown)


agriyakhetarpal (commented):
Thanks, @ryanking13!

> We set cffi as a cross-build package to make sure the build time cffi version is equal to the runtime cffi version, but is it a really necessary requirement?

I think it should be fine to remove in principle, but we should double-check before we hit the green button. There have been pre v1 cases where this was a problem, but I think it exists post v1 as well – there is at least one instance where the version is checked:

https://github.com/python-cffi/cffi/blob/25172b84cf51cf58241f074ccd1efd9c7648cb43/src/cffi/api.py#L54-L56

I tried, but I wasn't able to find any information online about the ABI compatibility of extensions generated using `cffi`. It does support the CPython limited/stable ABI, however, but that's for compatibility across Python versions, and not for `cffi` at build time/runtime.

ryanking13:
Thanks for the review!

> https://github.com/python-cffi/cffi/blob/25172b84cf51cf58241f074ccd1efd9c7648cb43/src/cffi/api.py#L54-L56

This part checks the `_cffi_backend` version, which should be vendored in the `cffi` package, so I think it is okay. I believe that problem existed when people didn't use wheels and installed python packages through e.g. `apt install package`.

Anyway, I agree that we should double-check if there exists any problem. I think @hoodmane originally added the cross-build-env constraints in `cffi`, so do you remember what happens when there was a version mismatch?

PR #5310, opened 2025-01-08
Use wasm-gc based call adaptor if available

hoodmane:
I figured out a way to remove the super inefficient `calculate_wasm_func_nargs_fallback` and replace it with wasm-gc code. This should perform much better. The cpython patch message says:

Part of the ongoing quest to support JSPI. The JSPI spec removed its dependence on JS type reflection, and now the plan is for runtimes to ship JSPI while keeping type reflection in stage 3. So we need an alternative way to count the number of parameters of a function. It is possible to count them by repeatedly trying to instantiate a webassembly module with the function as an import of a different type signature. But this is pretty inefficient.

Since WebAssembly gc is now stage 4, there is a new option. WebAssembly gc added the `ref.test` instruction which can ask if a funcref has a given type. It's a bit difficult to apply because even our usual assembler the wasm binary toolkit doesn't support this instruction yet. But all JS engines that support JSPI support it. We just have to do some manual work to produce the binary.

This code also has to be written carefully to interact properly with memory snapshots. Importantly, no JS initialization code can be called from the C initialization code. For this reason, we make a C function pointer to fill from JS and fill it in a preRun function.

Upstream PR:
https://github.com/python/cpython/pull/128628


cpython/patches/0005-Use-wasm-gc-based-call-adaptor-if-available.patch
```diff
@@ -0,0 +1,381 @@
...
++//     )
++// )
++addOnPreRun(() => {
++  // Try to initialize countArgsFunc
++  const code = new Uint8Array([
++    0x00, 0x61, 0x73, 0x6d, // \0asm magic number
++    0x01, 0x00, 0x00, 0x00, // version 1
++    0x01, 0x23, // Type section, body is 0x23 bytes
++      0x06, // 6 entries
++      0x60, 0x00, 0x01, 0x7f,                         // (type $type0 (func (param) (result i32)))
```
ryanking13, line 190: This part is quite scary 😅

ryanking13 (commented):
Thanks for working on this. I don't fully understand the patch the it generally looks okay to me. Let's see what CPython devs say about the patch.

cpython/patches/0005-Use-wasm-gc-based-call-adaptor-if-available.patch
hoodmane: Well you can see the wat above. The problem is that the wasm binary toolkit doesn't know how to handle `ref.test $type4`. Probably I should make a wasm binary toolkit PR to add support for this, but we'll see if I get around to it.

ryanking13:
Thanks!

PR #5314, opened 2025-01-09
DOC Update package build doc regarding recipe unvendoring

ryanking13:
Updates the package build doc, accounting for the package unvendoring. This PR needs to be merged after #4987.

I replaced “in-tree” and “out-of-tree” with “building packages using recipe” and “building from source”, as I think the distinction between the in-tree and out-of-tree has been blurred (we don't build packages "in-tree" anymore).


docs/development/building-packages-from-source.md
```diff
@@ -1,14 +1,14 @@
...
-Pyodide out of tree (for instance in your package's CI or for use with private
-packages).
+Pyodide, for instance in your package's CI or for use with private
+packages.
 
-Pyodide currently only supports Linux for out of tree builds, though there is a
-good change it will work in MacOS too. If you are using Windows, try Windows
-Subsystem for Linux.
+Pyodide currently only supports building packages in Linux environments officially,
+though there is a good change it will work in MacOS too.
```
tom-dudley, line 10:
```suggestion
though there is a good chance it will work in MacOS too.
```

Note: Pyodide builds and runs fine for me on a Mac M1 (all `make` targets + `build-recipes`). The only issues I've faced are:

- `sed` (Mac `sed` doesn't like the bare `-i` flag, but using `gnu-sed` works fine)
- `pkg-config` resolves SSL to system libs rather than the ones in `pyodide/packages/ssl` - this can be worked around by tweaking the `hashlib` and `ssl` `meta.yaml` files to hardcode the paths. (Relative paths may just work fine here, I've not properly tried.)

ryanking13: Thanks. Yes, we do build some packages in MacOS in our CI. But not all packages are built and tested, so we still have that note.

PR #5320, opened 2025-01-13
Use webassembly exception handling

hoodmane:
It's a bit hard to make the rust compiler play along, see the upstream PR for information about the rust difficulties https://github.com/pyodide/pyodide-build/pull/81

Based on / supersedes https://github.com/pyodide/pyodide/pull/3817

#### Build failures

- [x] python-flint
- [x] ppl
- [x] gdal

#### Import failures

These use setjmp/longjmp and have errors because the symbols for setjmp/longjmp are missing.

- [x] opencv
- [x] rasterio
- [x] optlang
- [x] fiona
- [x] swiglpk
- [x] pyiceberg
- [x] pygame-ce


hoodmane:
The only core test failure is `test_cpp_exceptions_and_syncify` which we'll need to deal with
https://github.com/emscripten-core/emscripten/issues/23395
in order to fix. This isn't exactly a regression either, I think the bug exists on the main branch but this change makes our test hit it.

hoodmane:
@mkoeppe I'm getting the following build failure from `primecount`:
```
/home/rchatham/Documents/programming/pyodide/packages/primecount/build/primecount-7.9/include/libdivide.h:2018:41: error: no member named 'denom' in 'divider<T, ALGO>'
 2018 |         return div.denom.magic == other.denom.magic &&
      |                                   ~~~~~ ^
/home/rchatham/Documents/programming/pyodide/packages/primecount/build/primecount-7.9/include/libdivide.h:2019:40: error: no member named 'denom' in 'divider<T, ALGO>'
 2019 |                div.denom.more == other.denom.more;
      |                                  ~~~~~ ^
```
I don't really understand why it would appear on this PR because it seems like a front end problem.

kimwalisch:
Your primecount error has already been fixed in the latest primecount-7.14 release.

hoodmane:
@ryanking13 we need a wasm-sjlj variant of `libSDL2_image_bmp-gif-jpg-png` for pygame, but it's not clear to me how to get one.

hoodmane:
I guess we need a patch to sdl2_image.py to make it wasm-sjlj aware like libpng...
https://github.com/emscripten-core/emscripten/blob/main/tools/ports/sdl2_image.py
https://github.com/emscripten-core/emscripten/blob/main/tools/ports/libpng.py

hoodmane:
@agriyakhetarpal @ryanking13 I think all tests are passing now, would appreciate if you could review this and https://github.com/pyodide/pyodide-build/pull/81.

ryanking13:
Thanks for working on this Hood. I'll review it soon, probably before this weekend.

Makefile.envs
```diff
@@ -5,10 +5,12 @@ export PYODIDE_ABI_VERSION ?= 2025_0
 
 export PYTHON_ARCHIVE_SHA256=73ac8fe780227bf371add8373c3079f42a0dc62deff8d612cd15a618082ab623
 
+export RUST_TOOLCHAIN ?= nightly-2025-01-15
+export RUST_EMSCRIPTEN_TARGET_URL ?= http://pyodide-cache.s3-website-us-east-1.amazonaws.com/rustc-emscripten-target/${RUST_TOOLCHAIN}.tar.bz2
```
ryanking13, line 9: Maybe let's store the archive in a more stable place, such as a GitHub release of a separate repository.

src/js/load-package.ts
```diff
@@ -44,7 +44,7 @@ export async function initializePackageIndex(
     );
   }
 
-  if (lockfile.info.version !== API.version) {
+  if (lockfile.info.version !== API.version && false) {
```
ryanking13, line 47: I guess this need to be reverted before merge?

hoodmane: I don't think I meant to commit this at all but yes.

agriyakhetarpal:
I remember that `pybind11`'s tests for Pyodide/WASM downstream were also passing `-fexceptions`. Should they also switch to `-fwasm-exceptions` once this PR makes it to a release?

hoodmane:
Yes everyone downstream needs to pass `-fwasm-exceptions` or they will get load time errors.

hoodmane:
But it'll be a bit annoying if they want to build for both Pyodide 0.27 and Pyodide 0.28 since they need different flags.

agriyakhetarpal:
Yes, that would be a breaking change for `cibuildwheel`. Could we document this better in the CHANGELOG? Maybe we can intercept this in `pywasmcross`, raising a warning and switching it out internally if set? I haven't contributed to that side of our codebase before, so if yes, I would be happy to try that out.

docs/project/changelog.md
```diff
@@ -18,6 +18,8 @@ myst:
 
 - ABI break: Upgraded Emscripten to 3.1.63 {pr}`5343` {pr}`5350` {pr}`5357`
   {pr}`5334` {pr}`5363`
+- ABI break: Switched to using WebAssembly exception handling for C++ errors,
+  Rust panics, and setjmp/longjmp. {pr}`5320`
```
agriyakhetarpal, line 22: Here's my attempt at documenting it better:

```suggestion
- ABI break: Switched to using WebAssembly exception handling for C++ errors,
  Rust panics, and setjmp/longjmp. Projects and build system helpers that have previously set the
`-fexceptions` compilation flag **must** switch to using `-fwasm-exceptions`. {pr}`5320`
```

hoodmane: Yeah we can intercept most things but not all things in pywasmcross.

Is it a breaking change for cibuildwheel? Does it automatically pick up new Pyodide versions and start building for them or is some intervention needed? In any case we can just fix it.

hoodmane:
I should clarify further about sjlj in the changelog.

docs/project/changelog.md
agriyakhetarpal: It's not a breaking change for `cibuildwheel`, but it'll be one for `pybind11` as it tests for WASM builds downstream (and Cython as well, should they choose to test WASM later on).

`cibuildwheel` doesn't actually pick up new Pyodide versions as they are released at the moment, yes. It's usually me who keeps updating it there. My PR for allowing arbitrary Pyodide versions is still open; I think I should revisit it.

hoodmane: It's pretty easy to hack on. You can probably just push your changes to my PR:
https://github.com/pyodide/pyodide-build/pull/81/files

hoodmane: Or maybe we should leave it to a follow-up pr.

agriyakhetarpal (approved):
Thanks, @hoodmane!

hoodmane:
Okay let's merge it =D

PR #5332, opened 2025-01-15
Emscripten 4.0.1

hoodmane:
Emscripten update 17 and 18/19

To fix:

- [ ] Pillow
- [ ] RobotRaconteur
- [ ] gdal
- [ ] libxslt
- [ ] pi-heif
- [ ] pillow-heif
- [ ] pygame-ce


hoodmane:
openssl fails to build here with:
```
wasm-ld: error: attempted static link of dynamic object ./libssl.so
wasm-ld: error: attempted static link of dynamic object ./libcrypto.so
```
It does not fail on 3.1.67. I guess we can bisect this too.

ryanking13:
From the error log, I'm guessing that something has changed to favor `libssl.so` when given `-lssl` and both `libssl.a` and `libssl.so` exist.

hoodmane:
Upstream issue:
https://github.com/emscripten-core/emscripten/issues/23572

hoodmane:
Next I'm getting:
```                                     
/home/rchatham/Documents/programming/pyodide/emsdk/emsdk/upstream/emscripten/cache/sysroot/include/c++/v1/__locale_dir/locale_base_api/musl.h:23:8: error: cannot add 'abi_tag' attribute in a             
redeclaration                                                                                                                                                                                              
   23 | inline _LIBCPP_HIDE_FROM_ABI long long strtoll_l(const char* __nptr, char** __endptr, int __base, locale_t) {                                                                                      
      |        ^                                                                                                                                                                                           
/home/rchatham/Documents/programming/pyodide/emsdk/emsdk/upstream/emscripten/cache/sysroot/include/c++/v1/__config:545:22: note: expanded from macro '_LIBCPP_HIDE_FROM_ABI'                               
  545 |       __attribute__((__abi_tag__(_LIBCPP_TOSTRING(_LIBCPP_ODR_SIGNATURE))))   
```

hoodmane:
Upstream issue:
https://github.com/emscripten-core/emscripten/issues/23573

hoodmane:
@ryanking13 could you look into the gdal build failure here?

ryanking13:
> @ryanking13 could you look into the gdal build failure here?

Which one? I see the `libxslt` build failure but not gdal.

ryanking13:
I see. I'll try to build gdal locally.

ryanking13:
I was able to build gdal locally without problem. Was the error related to your recent fix https://github.com/pyodide/pyodide/pull/5407?

hoodmane:
I also built it locally without problem. See CI failure here:
https://app.circleci.com/pipelines/github/hoodmane/pyodide/6867/workflows/95591817-04a7-4571-a11f-b505cfbf1cbb/jobs/104607

ryanking13:
Thanks

```
wasm-ld: error: attempted static link of dynamic object 
emcc: error: '/root/repo/emsdk/emsdk/upstream/bin/wasm-ld -o xsltproc.wasm      
xsltproc.o ../libxslt/.libs/libxslt.a -L/root/repo/packages/.libs/lib           
-L/root/repo/packages/.libs/lib/lib ../libexslt/.libs/libexslt.a                
/root/repo/packages/libxslt/build/libxslt-1.1.33/libxslt/.libs/libxslt.a        
/root/repo/packages/.libs/lib/libxml2.a /root/repo/packages/.libs/lib/libz.so
```

So looks like it is trying to link the `/root/repo/packages/.libs/lib/libz.so` shared library to the static WASM file. Probably we should find out a way to tell `wasm-ld` to prefer the static library.

hoodmane:
I'll try to get an MRO make an upstream report for `libx.a` vs `libx.so`.

hoodmane:
Well it doesn't seem like this specifically is the regression since the following fails in the same way on 3.1.74:
```console
$ ar -cq libx.a
$ touch a.c
$ emcc a.c -L. -lx
$ echo $?
0
$ touch x.c
$ emcc -sSIDE_MODULE x.c -o libx.so
$ emcc a.c -L. -lx
wasm-ld: error: attempted static link of dynamic object ./libx.so
```

hoodmane:
In any case, I opened https://github.com/emscripten-core/emscripten/issues/23591 and they fixed it so I backported the fix here.

ryanking13 (approved):
Thanks! Let's check the individual package build failures after the update.

Please update the PR title to 4.0.1

hoodmane:
@ryanking13 so to be clear you want to merge as is and then fix the broken packages in followup work? I am okay with that, though I think it would be good to merge #5320 first since it works correctly with all these packages. Also, we should maybe investigate fixing these a bit more before merging.

ryanking13:
> so to be clear you want to merge as is and then fix the broken packages in followup work?

Yes, I am personally okay with that. We can probably 1) https://github.com/pyodide/pyodide/pull/5320 and this and , 2) unvendor recipes, 3) investigate the package build issues in a separate repository.

hoodmane:
Okay sounds good to me, thanks!

ryanking13 (approved):
Thanks!

For the failing jobs, I think we can [remove matplotlib here](https://github.com/pyodide/pyodide/blob/764b220e586809645b242f34d6a850b4aef49161/benchmark/benchmark.py#L204) as we are not running matplotlib benchmark anyways. Aslo in pandas test, it seems like loading matplotlib is [not actually required](https://github.com/pyodide/pyodide/blob/764b220e586809645b242f34d6a850b4aef49161/packages/pandas/test_pandas.py#L52)

PR #5334, opened 2025-01-15
Emscripten 3.1.62

hoodmane:
+4 releases, a quarter of the way to 4.0.0. Opening at the same time as #5332 for faster bisecting.


Makefile.envs
```diff
@@ -5,6 +5,11 @@ export PYODIDE_ABI_VERSION ?= 2025_0
 
 export PYTHON_ARCHIVE_SHA256=73ac8fe780227bf371add8373c3079f42a0dc62deff8d612cd15a618082ab623
 
+# I'm not really sure what this explicit -Z link-native-libraries=yes is for,
+# `rustc -Z help` shows that `link-native-libraries=yes` is the default. But it
+# seems to be necessary...
+export RUSTFLAGS = -C link-arg=-sSIDE_MODULE=2 -Z link-native-libraries=yes
```
ryanking13: The default value [set by `pyodide-build` is `link-native-libraries=no`](https://github.com/pyodide/pyodide-build/blob/7ebf9a274de346477910ae9b6885a5aed0b98fff/pyodide_build/config.py#L207)

hoodmane: But that's part of the default value for `RUSTFLAGS`. Is `pyodide-build` appending the `Makefile.envs` `RUSTFLAGS` to the default `RUSTFLAGS`?

ryanking13: The value in the Makefile.envs will overwrite the default value specified in the pyodide-build.

But before this PR there was no RUSTFLAGS in makefile.envs. So the default value was used.

hoodmane: What I'm confused about is that when I said:
```
export RUSTFLAGS = -C link-arg=-sSIDE_MODULE=2
```
I got failures due to missing symbols from native libs. I thought this would override the default `RUSTFLAGS` and get rid of `-Z link-native-libraries=no` leaving it in the default setting which is `yes`. But in order to get it to work I had to explicitly put `-Z link-native-libraries=yes` here. I don't understand why.

ryanking13: Hmm, I see. That is very strange Indeed.

PR #5340, opened 2025-01-15
Rust: Remove -Zlink-native-libs=no and bigint flags

hoodmane:
We don't need these anymore. link_native_libs=no was fixed by:
https://github.com/rust-lang/libc/pull/4002
and wasm_bigint was fixed by:
https://github.com/rust-lang/rust/pull/131736


ryanking13 (commented):
Thanks! Probably we can add `rust_flags` to [`OVERRIDABLE_BUILD_KEYS`](https://github.com/pyodide/pyodide-build/blob/7ebf9a274de346477910ae9b6885a5aed0b98fff/pyodide_build/config.py#L188-L198), so we can set it in pyproject.toml not in Makefile.envs.

ryanking13:
> I wonder if it would make sense to keep Makefile.envs as the source of truth and then generate a json file from it and use that in other tools. Then we can have a CI job that checks that the generated json file matches the output of Makefile.envs. And that will let us use variables.

`pyodide-build` can read Makefile.envs and export the values using `pyodide config` CLI, so I think other tools can use that.

> I moved RUST_TOOLCHAIN back to Makefile.envs so I can substitute it into RUST_EMSCRIPTEN_TARGET_URL:

Hmm, maybe in this case, putting the values in Makefile.envs makes sense, as it should affect all packages that build against the new Pyodide versions, which will use wasm-eh.

Okay, then I am +1 with this PR.

ryanking13 (approved):

hoodmane:
Hitting `error: rustup is not installed at '/root/.cargo'` again.

PR #5341, opened 2025-01-16
Add patch to RobotRaconteur to add pyproject.toml for build system

johnwason:
### Description

This pull request adds a patch that adds a `pyproject.toml` file as for the build system as requested by https://github.com/robotraconteur/robotraconteur/issues/268.


johnwason:
I am not sure what I am looking at for these results. Where is my package `RobotRaconteur` being built? It depends on `boost`, `numpy`, and `openssl`.

hoodmane:
Since it depends on numpy it's built in `build-packages-numpy-dependents`. It takes over an hour for the ci to get there.

packages/RobotRaconteur/meta.yaml
```diff
@@ -5,6 +5,8 @@ package:
 source:
   url: https://github.com/robotraconteur/robotraconteur/releases/download/v1.2.2/RobotRaconteur-1.2.2-Source.tar.gz
   sha256: "837710102b5ee6ad81a22e14b313b1dd09995c3da3e7846355866553b3c52715"
+  patches:
+    - pyproject_toml.patch
```
hoodmane, line 9: If you're adding an entire file you can do:
```suggestion
  extras:
    - [pyproject.toml, pyproject.toml]
```
And then just add the file itself as `packages/RobotRaconteur/pyproject.toml`.

johnwason: Thanks I made the suggested change to use "extras" section

hoodmane (approved):

hoodmane:
Thanks!

johnwason:
The tests failed for pyproject.toml. I will take a look tomorrow.

agriyakhetarpal:
@johnwason, I had to do something similar with the `zfpy` package with #5172 recently, see here: https://github.com/pyodide/pyodide/blob/ead31b94075310df465213a25e01eeb5c54c1383/packages/zfpy/patches/Modernise-packaging.patch, which should help.

(This reminds me that I was requested to contribute this upstream as well, missed doing that at that time!)

packages/RobotRaconteur/pyproject.toml
```diff
@@ -0,0 +1,12 @@
...
+build-backend = "setuptools.build_meta"
+
+[project]
+name = "RobotRaconteur"
+version = "1.2.2"
+dependencies = ["numpy"]
+authors = [
+    {name = "John Wason", email = "wason@wasontech.com"}
+]
+license = "Apache-2.0"
```
hoodmane, line 12: Ci complains about a missing license field but sure looks like you have a license field here...

johnwason: I am not sure what it is complaining about either. The build for Robot Raconteur is pretty complicated, and the setup.py is generated during the native build so it may be confused. I will keep looking for the cause of the problem.

agriyakhetarpal: It is coming from the fact that the license is also specified in `setup.py`, and the entry in `pyproject.toml` is overridden by it: https://github.com/robotraconteur/robotraconteur/blob/35c50bbec7a66b5b05d4b585b4c63d536f902fd8/RobotRaconteurPython/setup.py.in#L39

so it would have to be removed from there.

Also, it would be better to store it here as `license = {text = "Apache License (2.0)"}`:

```suggestion
license = {text = "Apache License (2.0)"}
```

hoodmane: Could we just remove the license field from `pyproject.toml`?

agriyakhetarpal: Yes, that works, too

ryanking13:

PR #5346, opened 2025-01-16
CI Enable prebuilt packages

ryanking13:
### Description

Splitted off from #4987. Without removing the recipes, this PR adds a new CI job that downloads the prebuilt packages and creates a distribution.


hoodmane (approved):

PR #5215, opened 2024-11-20
Extract out JS API for package installation

ryanking13:
### Description

Related: #4914

This PR separates the logic regarding installing a package and exposes it as a single JS API. This API is accessible through `pyodide_js._api.install` in Python code, and it is expected to be used in micropip to unify the package installation logic of loadPackage and micropip.install.

### Checklists

- [x] Add / update tests

PR #5358, opened 2025-01-20
Update Installer.install API to allow writing custom metadata files

ryanking13:
### Description

Replaces `installer` and `source` parameter to `metadata`, which accepts a `Map<filename, content>`, in `install` API (added in https://github.com/pyodide/pyodide/pull/5215).

The goal of this PR is to:

1. Hide pyodide internals from the install API, I would like to make this API public in the future. ([related comment](https://github.com/pyodide/pyodide/pull/5294#issuecomment-2572998431))
2. micropip needs to store metadata files such as `PYODIDE_URL`, `PYODIDE_SHA256`, so we need a more general parameter so micropip can use it.


hoodmane:
I still think we should make a replacer for this function part of the `loadPyodide` arguments but that also requires the API to be public.

src/js/api.ts
```diff
@@ -520,7 +520,7 @@ export class PyodideAPI {
       buffer,
       format,
       extract_dir,
-      installer: "pyodide.unpackArchive",
+      metadata: new Map([["INSTALLER", "pyodide.unpackArchive"]]),
```
hoodmane, line 523: Rather than making a new map every time maybe we could make this a top level constant?

hoodmane: I guess this isn't used that much.

src/js/installer.ts
```diff
@@ -24,16 +24,14 @@ export class Installer {
     buffer: Uint8Array,
     filename: string,
     installDir: string,
-    installer: string,
-    source: string,
+    metadata?: Map<string, string>,
```
hoodmane, line 27:
```suggestion
    metadata?: readonly Map<string, string>,
```

hoodmane:
> I still think we should make a replacer for this function part of the loadPyodide arguments but that also requires the API to be public.

Oh I was confusing this with the `download` function never mind.

hoodmane (approved):
Thanks!

src/js/api.ts
ryanking13: Thanks. I think it is good to make it a constant.

PR #5361, opened 2025-01-21
TST Update import test to use lockfile

ryanking13:
### Description

Splitted off from #4987. Updates the package import test to use the lockfile instead of the meta.yaml file.

After we unvendor recipes, it is not possible to run import tests using the recipe file. So this PR changes the test to use the lockfile. 

Note that the `test.imports` section in the meta.yaml file is not exported to the lockflie, so the test coverage will drop. But we are running the import tests using meta.yaml file in [`pyodide-recipes` repository](https://github.com/pyodide/pyodide-recipes/blob/main/packages/test_packages_common.py) anyway, so it is not an issue IMO.


packages/_tests/test_packages_common.py
```diff
@@ -35,78 +23,69 @@ def registered_packages() -> list[str]:
...
-    pyc_count = sum(_get_file_count(f"/{key}/**/*.pyc") for key in import_names)
-    py_count = sum(_get_file_count(f"/{key}/**/*.py") for key in import_names)
-
-    assert not ((py_count > 0) and (pyc_count > 0))
-
-    # Make sure no exe files were loaded!
-    assert _get_file_count("/**/*.exe") == 0
+    for import_name in imports:
+        selenium_standalone.load_package(name)
+        selenium_standalone.run(f"import {import_name}")
```
hoodmane, line 91: I wonder if we could consider switching away from selenium_standalone for this. Or say doing several packages with the same selenium instance before rebooting it. Presumably loading all packages in one instance leads to OOM. But we used to be using selenium_standalone to validate that there are no missing deps. Now I think we are primarily concerned with validating the deps elsewhere and we're primarily concerned that the shared objects have the correct abi.

hoodmane: No action needed here of course.

packages/_tests/test_packages_common.py
```diff
@@ -35,78 +23,69 @@ def registered_packages() -> list[str]:
...
-            ))
-            """
-        )
-
-    import_names = meta.test.imports
-    if not import_names:
-        import_names = meta.package.top_level
-
-    if not import_names:
-        # Nothing to test
```
hoodmane, line 95: Might as well keep this comment I suppose.

packages/_tests/test_packages_common.py
ryanking13: Sounds reasonable to me. Thanks for the suggestion!

ryanking13:
> Did you move all the deleted checks to pyodide-recipes?

Yes, I copied the `test_packages_common` file to pyodide-recipes. I'll sync the package tests too.

ryanking13:
Okay, tests are passing now. I found and fixed several wrong dependencies in recipes.

Issue #5336, opened 2025-01-15
The versions of pydantic and pydantic-core do not match


frybox:
## 🐛 Bug

The current version of Pydantic is 2.9.2, and Pydantic-core is 2.25.1. At runtime, an error occurs because the former does not support invalid schemas.

### To Reproduce

```html
<script src="https://cdn.jsdelivr.net/pyodide/v0.27.1/full/pyodide.js"></script>
<script>
    async function main() {
        const pyodide = await loadPyodide();
        await pyodide.loadPackage('micropip');
        const micropip = pyodide.pyimport('micropip');
        await micropip.install('pydantic')
        pyodide.runPython(`
        from pydantic.json_schema import GenerateJsonSchema
        GenerateJsonSchema()
        `);
    }
    main();
</script>
```

run the former code in a browser, the described error occurs:

```
pyodide.asm.js:10 
 Uncaught (in promise) PythonError: Traceback (most recent call last):
  File "/lib/python3.12/site-packages/pydantic/json_schema.py", line 337, in build_schema_type_to_method
    mapping[key] = getattr(self, method_name)
                   ^^^^^^^^^^^^^^^^^^^^^^^^^^
AttributeError: 'GenerateJsonSchema' object has no attribute 'invalid_schema'

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "/lib/python312.zip/_pyodide/_base.py", line 523, in eval_code
    .run(globals, locals)
     ^^^^^^^^^^^^^^^^^^^^
  File "/lib/python312.zip/_pyodide/_base.py", line 357, in run
    coroutine = eval(self.code, globals, locals)
                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "<exec>", line 3, in <module>
  File "/lib/python3.12/site-packages/pydantic/json_schema.py", line 294, in __init__
    self._schema_type_to_method = self.build_schema_type_to_method()
                                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/lib/python3.12/site-packages/pydantic/json_schema.py", line 339, in build_schema_type_to_method
    raise TypeError(
TypeError: No method for generating JsonSchema for core_schema.type='invalid' (expected: GenerateJsonSchema.invalid_schema)
```

### Expected behavior

Upgrade Pydantic and Pydantic-core to the current compatible versions

### Environment

- Pyodide Version: 0.27.0, 0.27.1
- Browser version:
[...]

PR #5368, opened 2025-01-22
Upgrade pydantic to 2.10.5 and add tests

samuelcolvin:
### Description

This builds on top of #5367, but is a separate PR as I'm not sure I'm doing things correctly.

I'm trying to upgrade pydantic and pydantic-core since:
* `pydantic-core` was previously marked as disable
* issue reported in #5336
* I really really need pydantic 2.10 for a project I'm working on 🙏 

I have no idea if:
* this is the right way to upgrade a package - I manually calculated the hashes
* if the tests pass - I've tried for ages to get tests to run, but I can't avoid the test getting skipped with `(package 'pydantic' is not built.)`

I can confirm build seems to be working correctly.

### Checklists

- [x] Add / update tests

### DIFF
--- a/packages/pydantic/meta.yaml
+++ b/packages/pydantic/meta.yaml
@@ -15,43 +15,44 @@ requirements:
     - annotated-types
 test:
   imports:
-    - aliases
-    - alias_generators
-    - annotated_handlers
-    - class_validators
-    - color
-    - config
-    - dataclasses
-    - datetime_parse
-    - decorator
-    - deprecated
-    - env_settings
-    - errors
-    - error_wrappers
-    - experimental
-    - fields
-    - functional_serializers
-    - functional_validators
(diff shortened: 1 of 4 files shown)


packages/pydantic/test_pydantic.py
hoodmane, line 1: So all you need to do here is move the json and datetime imports inside the test. The global scope is not visible from inside the decorator.

packages/pydantic/meta.yaml
hoodmane, line 1: Let's also add a test.imports section with all the submodules of pydantic.

hoodmane:
Okay the tests pass for me locally now.

hoodmane (approved):

samuelcolvin:
Amazing, thank you so much.

PR #5379, opened 2025-01-28
Emscripten 3.1.65

hoodmane:
Emscrpten version 7/18


hoodmane:
Yeah the runner failure reproduces so I'll have to look into it. Making these updates is really a pain.

hoodmane:
Upstream issue: https://github.com/emscripten-core/emscripten/issues/23528
We'll need a patch for this.

PR #5391, opened 2025-01-30
Emscripten 3.1.69

hoodmane:
Emscripten update 11/19


hoodmane:
Now pygame has test failures...

ryanking13:
Probably, there is some change in the new SDL version. Only one test fails, so I think we can skip that test for now.

BTW, I thought I set to skip the pygame test suite in the CI (as it takes too long time to run), but I made a mistake to put this condition in the @run_in_pyodide decorator...

https://github.com/pyodide/pyodide/blob/be5e6de2cb83777e3c278922b07c3318e407afa4/packages/pygame-ce/test_pygame.py#L65-L66

ryanking13 (approved):

PR #5398, opened 2025-01-31
Matplotlib: use patched webagg backend

ianthomas23:
This PR makes the default matplotlib backend in pyodide a patched version of the `webagg` backend that is included in the matplotlib source code, as discussed at https://github.com/pyodide/matplotlib-pyodide/issues/65#issuecomment-2532463697. The idea is that this backend is more functionally complete that those in the `matplotlib-pyodide` repo and it is more maintainable long term, thus allowing support for `matplotlib-pyodide` to eventually be stopped (https://github.com/pyodide/matplotlib-pyodide/issues/65).

Video of it in action on my dev machine:

https://github.com/user-attachments/assets/63edeefd-dc2b-4348-80f9-36bb2f6d4263

which uses this example code (put it in the `dist` directory after building this PR):
```html
<!doctype html>
<html>
  <head>
    <script src="pyodide.js"></script>
  </head>
  <body>
      async function main(){
        await micropip.install('matplotlib');
        await pyodide.runPython(`
          import numpy as np
          fig, ax = plt.subplots()
          x, y = np.meshgrid(np.linspace(0, 2*np.pi, 20), np.linspace(0, 2*np.pi, 20))
          z = np.sin(x - y/4)*np.sin(y)
          cs = ax.contourf(x, y, z)
          fig.colorbar(cs, ax=ax)
          plt.show()`
        );
      }
      main();
    </script>
  </body>
</html>
```
Note it supports zooming and panning, undo/redo history, changing the size of the canvas, and export to PNG, SVG, PDF, etc file (not shown).

Summary of changes:
1. I have tried to make a minimal set of changes to matplotlib code (python, javascript and CSS files).
2. This is now the default backend in pyodide (outside of JupyterLite) so it is not necessary to use `matplotlib.use("webagg")`.
3. Use of `tornado` is removed as the python and javascript can directly call each other.
4. I've used mock python and javascript websockets to keep most of the websocket client code unchanged.
5. Messages sent via websocket from javascript to python are not JSON encoded and then decoded again as this is unnecessary.
6. JS code is injected from python using `run_js`.
7. CSS is inserted into the DOM at the end of the `document.head`.
8. Toolbar button images are obtained using a python callback that reads a PNG file from the wheel and returns it in binary to be wrapped as a `Blob`. I'm not sure this is the best approach but it works.
9. Download of an image is implemented via a websocket message rather than a specific `ondownload` callback.

- [x] Add / update tests


ryanking13:
Awesome! Thanks for sharing the progress. I haven't read the patch file in detail, but it looks like it already integrates into matplotlib better than matplotlib-pyodide, as Hood said.

I think we can discuss about managing the patched code when the PR is more complete, because the patch files are not small, and having them in .patch format would probably make the code less readable and less easy for others to modify.

agriyakhetarpal:
As a bonus, would it be possible to upstream this patch into Matplotlib someday?

hoodmane:
Yeah it would be nice if matplotlib could carry `pyodide-webagg`.

ianthomas23:
> I think we can discuss about managing the patched code when the PR is more complete, because the patch files are not small, and having them in .patch format would probably make the code less readable and less easy for others to modify.

The patch file is indeed a bad format for viewing the changes. I've created a never-to-be-merged PR at matplotlib/matplotlib#29568 so it is possible to view the side by side changes to confirm that most of the changes are removals, the additions are mostly the mock web sockets classes on both the Python and JavaScript sides, followed by the new `handle_save` function. In terms of review what I would most like is a `pyodide` expert to check on the use of `create_proxy`, `run_js`, `as_py_json` and `toJs`.

In terms of future upstreaming into Matplotlib, I think it will be possible to refactor the `webagg` code in Matplotlib so that the code removals here (`tornado` etc) are in a separate file so that the patch file is smaller and simpler. But upstreaming all of it is unlikely unless it is covered by CI, so that would depend on matplotlib/matplotlib#29093 being completed using this, which is subject to discussion and agreement between Matplotlib maintainers.

> Yeah it would be nice if matplotlib could carry pyodide-webagg.

I am being careful to call this the `webagg` backend as it is only slightly modified from the current Matplotlib `webagg` backend. If it has its own name that is different from `webagg` then after much discussion, etc, it will probably end up in a repo of its own, not very well maintained, similar to `matplotlib-pyodide`.

hoodmane:
Is this ready for review?

hoodmane:
> I am putting this up early to show the progress so far, and I am happy to take recommendations from pyodide experts about things that could be done better.

I guess I should read through it then. I'll try to get to it soon!

ianthomas23:
I am coming back to this now. Sorry it has taken so long!

New changes:

- Modified the patch (see matplotlib/matplotlib#29568 for easiest way to view the changes) to improve python/js proxy lifetimes.
- Added `TimerJs` class to patch, based on similar code in `matplotlib-pyodide`, to support Matplotlib animations.
- Added tests here, taken from the `matplotlib-pyodide` tests but using `webagg` backend instead.
- Modified changelog entry.

From my point of view this feels complete enough. There are no doubt some things that do not work perfectly, and probably some memory leaks, but it functionally works for most common Matplotlib use cases. I would like to get this merged soon so that I can concentrate on bringing the recipe up to date (from Matplotlib 3.8.4 to the latest 3.10.1) before improving it further and seeing if the patch can be simplified by upstreaming some of the changes to the Matplotlib repo.

ryanking13 (approved):
Thank you so much for this work @ianthomas23!

When I ran the code locally, it was far better than matplotlib-pyodide, especially that it supports zoom-in-out and saving figures natively. Love it.

chrome_6hEh0sLZLa

I left some minor comments, but otherwise it looks good to me. I'll wait for a few more days before merging to see if other maintainers would want to review too.

packages/matplotlib/patches/0004-WebAgg-backend.patch
```diff
@@ -0,0 +1,703 @@
...
+             self._force_full = True
+             self.draw_idle()
+ 
++    def handle_save(self, event):
++        figure_id = event['figure_id']
++        format = event['format']
++
++        mimetype = mimetypes.types_map.get(f".{format}")
++        if mimetype is None:
++            window.alert(f"Cannot download plot, unable to determine mimetype for '{format}'")
```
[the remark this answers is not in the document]
hoodmane: Or simply `from js import alert`

packages/matplotlib/patches/0004-WebAgg-backend.patch
```diff
@@ -0,0 +1,703 @@
+From 90a1bb0368189e7a7ce024032d5ff404367fd1fa Mon Sep 17 00:00:00 2001
+From: Ian Thomas <ianthomas23@gmail.com>
+Date: Fri, 31 Jan 2025 11:05:26 +0000
+Subject: [PATCH] Patch WebAgg backend
+
```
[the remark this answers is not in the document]
ianthomas23: Done

packages/matplotlib/meta.yaml
```diff
@@ -37,7 +38,7 @@ build:
   script: export SETUPTOOLS_SCM_PRETEND_VERSION=$PKG_VERSION
   post: |
     rm -rf matplotlib/backends/qt_editor
-    rm -rf matplotlib/backends/web_backend
+    find matplotlib/backends/web_backend -type f ! -name mpl.css ! -name mpl.js -delete
```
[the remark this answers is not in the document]
ianthomas23: Done

packages/matplotlib/patches/0004-WebAgg-backend.patch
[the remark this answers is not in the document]
ianthomas23: Done

packages/matplotlib/patches/0004-WebAgg-backend.patch
```diff
@@ -0,0 +1,721 @@
...
++            clear_interval(self._timer)
++            self._timer = None
+
+     def _timer_set_interval(self):
+         # Only stop and restart it if the timer has already been started
+@@ -117,44 +111,9 @@ class TimerTornado(backend_bases.TimerBase):
+             self._timer_start()
+
+
+-class TimerAsyncio(backend_bases.TimerBase):
```
hoodmane, line 470: We could also just use `TimerAsyncio` -- I think it would work fine and require a slightly smaller patch. Or is there something I'm misunderstanding?

packages/matplotlib/patches/0004-WebAgg-backend.patch
```diff
@@ -0,0 +1,721 @@
...
++            "href",
++            "data:{};base64,{}".format(
++                mimetype, base64.b64encode(data.getvalue()).decode("ascii")
++            ),
++        )
++        element.setAttribute("download", f"plot{figure_id}.{format}")
++        element.style.display = "none"
++        document.body.appendChild(element)
++        element.click()
++        document.body.removeChild(element)
```
hoodmane, line 564: I guess in 2025 there is still no better way to do this than this hack?

packages/matplotlib/patches/0004-WebAgg-backend.patch
```diff
@@ -0,0 +1,721 @@
...
+     def send_event(self, event_type, **kwargs):
+         if self.manager:
+             self.manager._send_event(event_type, **kwargs)
+@@ -498,6 +473,10 @@ class FigureManagerWebAgg(backend_bases.FigureManagerBase):
+         output.write("mpl.default_extension = {};".format(
+             json.dumps(FigureCanvasWebAggCore.get_default_filetype())))
+
++        output.write("mpl.toolbar_image_callback = null;\n")
++        output.write("mpl.set_toolbar_image_callback = function(c) {mpl.toolbar_image_callback=c;}\n")
++        output.write("mpl.set_toolbar_image_callback;\n")
```
hoodmane, line 575: What's the point of this third line?

packages/matplotlib/patches/0004-WebAgg-backend.patch
```diff
@@ -0,0 +1,721 @@
...
++    }
++
++    receive_json(content) {
++        this.onmessage({ data: content });
++    }
++
++    send(content) {
++        if (this.python_onmessage_callback != null) {
++            this.python_onmessage_callback(content);
++        }
```
hoodmane, line 644: Similar nit:
```suggestion
+        this?.python_onmessage_callback(content);
```

packages/matplotlib/patches/0004-WebAgg-backend.patch
```diff
@@ -0,0 +1,721 @@
...
++            this.onopen();
++        }
++    }
++
++    receive_binary(content, binary=true) {
++        var buffer = content.getBuffer();
++        content.destroy();
++        try {
++            blob = new Blob([buffer.data]);
++            this.onmessage({ data: blob });
```
hoodmane, line 631: It doesn't look like we declared blob? Also if we call it data we can shorten a bit.
```suggestion
+            const data = new Blob([buffer.data]);
+            this.onmessage({ data });
```

hoodmane (approved):
Thanks @ianthomas23. I left some nits, the most important seems to be that we don't declare `blob` in one spot? Is there a way to make sure that the JavaScript is executed in strict mode?

Also I think we could save some code by getting rid of `TimerJs` and using `TimerAsyncio` instead, but I'm not completely sure.

packages/matplotlib/patches/0004-WebAgg-backend.patch
ianthomas23: `TimerAsyncio` works, thanks.

packages/matplotlib/patches/0004-WebAgg-backend.patch
ianthomas23: It is not needed, I've removed it.

packages/matplotlib/patches/0004-WebAgg-backend.patch
```diff
@@ -0,0 +1,721 @@
...
++    set onopen(listener) {
++        this._onopen = listener;
++        this.readyState = 1;
+     }
++
++    open(python_onmessage_callback) {
++        this.python_onmessage_callback = python_onmessage_callback;
++        if (this.onopen != null) {
++            this.onopen();
++        }
```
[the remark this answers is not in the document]
ianthomas23: Done

packages/matplotlib/patches/0004-WebAgg-backend.patch
ianthomas23: Done

packages/matplotlib/patches/0004-WebAgg-backend.patch
ianthomas23: Done

packages/matplotlib/patches/0004-WebAgg-backend.patch
```diff
@@ -0,0 +1,721 @@
...
++        try {
++            blob = new Blob([buffer.data]);
++            this.onmessage({ data: blob });
++        } finally {
++            buffer.release(); // Release the memory when we're done
++        }
++    }
++
++    receive_json(content) {
++        this.onmessage({ data: content });
```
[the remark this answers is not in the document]
ianthomas23: Done

ianthomas23:
> Is there a way to make sure that the JavaScript is executed in strict mode?

Seemingly not. I've tried, and it identified a couple of missing declarations but ultimately it fails with some of the js/python conversions with something along the lines of "Cannot set property of object which has only a getter".

> Also I think we could save some code by getting rid of `TimerJs` and using `TimerAsyncio` instead, but I'm not completely sure.

Yes, you are right, I've done this.

hoodmane:
What's the change you made to make it strict mode? I can help debug that.

ianthomas23:
> What's the change you made to make it strict mode? I can help debug that.

I've pushed a commit with `"use strict"` added at the start of each of the 2 uses of `run_js`. I've taken this from https://github.com/pyodide/pyodide/blob/e10812dd80c57556cb17230fa71af994b87330ca/src/tests/test_pyproxy.py#L513 and similar.

hoodmane:
@ianthomas23 can you merge main? I think you need to bring in #5626 for CI to work again.

hoodmane:
Hmm tests seem not to be passing anymore.

ryanking13:
> Hmm tests seem not to be passing anymore.

Maybe there was a change in CSS and just re-creating the sample images would work?

vadimkantorov:
Regarding the toolbar images: if whl files  are actually zips and if toolbar images can be added as uncompressed entries, then they can be just read via an appropriate offset from whl-files (and the offset would just need to be calculated once, maybe at whl-building stage), if such optimization is wanted (although maybe it would be an overkill)

ianthomas23:
> Hmm tests seem not to be passing anymore.

The test failures were caused by my experiment with `use strict`. I've reverted that so that the tests pass.

hoodmane:
Okay let's merge it!

hoodmane:
@ianthomas23 can you update the pr description since it says it's not ready yet?

ianthomas23:
> @ianthomas23 can you update the pr description since it says it's not ready yet?

Done.

vadimkantorov:
Does this enable running matplotlib figure rendering in a separate WebWorket? This is required for heavy plots...

ryanking13:
> Does this enable running matplotlib figure rendering in a separate WebWorket? This is required for heavy plots...

I don't think so. webworker does not have access to the DOM which is required.

Of course, you can still build a pipeline to run matplotlib in webworker, store the figure, send a data to the main thread and render if from there.

vadimkantorov:
Yes, I'm meaning this flow, of rendering to an OffscreenCanvas

I guess for any practical usage of matplotlib for interactive analytics for large-ish datasets, there must be an example for running matplotlib/pyodide in a webworker...

ianthomas23:
The `webagg`-related backends of Matplotlib work by the python code creating a PNG of the image to display and sending this to the browser for display. Dragging, zooming, etc the image in the browser results in a round trip to the python code to recreate a new PNG for display. The Matplotlib core `webagg` backend uses `tornado` to handle the requests from the browser, the `pyodide` flavour of `webagg` has the comms replaced by direct function calls as the python and javascript code sit side by side.

The comms layer between the python and JS code could be replaced with something else if one wants a different architecture.

vadimkantorov:
For practical use, it seems that the comms between JS and Python should be webworker-based message-exchange?

PNG base64 or SVG text (or base64) should natively be okay for such interchange...

PR #5404, opened 2025-02-01
Update `micropip` to version 0.9.0

agriyakhetarpal:
### Description

https://github.com/pyodide/micropip/releases/tag/v0.9.0 was released earlier in the day; this pull request updates it here.

### Checklists

### DIFF
--- a/docs/project/changelog.md
+++ b/docs/project/changelog.md
@@ -24,6 +24,8 @@ myst:
 
 - Upgraded `narwhals` to 1.24.1 {pr}`5386`
 - Upgraded `rateslib` to 1.7.0 {pr}`5400`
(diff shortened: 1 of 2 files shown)


agriyakhetarpal:
Based on the failing tests, we either need to add `packaging` or remove it. Since we don't need it (only `micropip` depended on it), could we remove it as appropriate?

agriyakhetarpal:
The test is valuable to keep, but I'll check what to replace `packaging` by.

ryanking13:
Thanks @agriyakhetarpal! We should replace the tests that depends on third-party packages... probably after unvendoring packages, we can make some more test-only packages to use in our core test suite.

agriyakhetarpal:
Superseded by #5461

PR #5406, opened 2025-02-03
Use _Py_Version to compute Python version for library setup

hoodmane:
This requires this upstream fix which came out in Emscripten 3.1.73: https://github.com/emscripten-core/emscripten/pull/23020/

Applied upstream in the cpython repo here:
https://github.com/python/cpython/pull/127113/files#diff-31f61a127a59386957acba86e9acedf0a96bc6ebe1ac159a48e39b1a291111b5R73-R75

This fixes an error that occurs when building with assertions -- we called `_py_version_major` before the runtime was set up so it triggered an assertion failure.


src/js/emscripten-settings.ts
```diff
@@ -110,6 +110,15 @@ function mountLocalDirectories(mounts: string[]): PreRunFunc {
...
 
+function computeVersionTuple(Module: Module): [number, number, number] {
+  const versionHex = Module.HEAPU32[Module._Py_Version / 4].toString(16);
+  const res = versionHex
+    .padStart(8, "0")
+    .match(/.{1,2}/g)!
+    .map((x) => parseInt(x, 16));
+  return res as any;
+}
+
```
agriyakhetarpal, line 121:
```suggestion
function computeVersionTuple(Module: Module): [number, number, number] {
 const versionInt = Module.HEAPU32[Module._Py_Version >>> 2];
  const major = (versionInt >>> 24) & 0xFF;
  const minor = (versionInt >>> 16) & 0xFF;
  const micro = (versionInt >>> 8) & 0xFF;
  return [major, minor, micro];
}
```



Question: would an unsigned right shift be better? We are converting a number → hex string, then padding it, splitting it, and parsing it back to numbers (which might be unnecessary conversions). Please feel more than free to correct me if I am wrong somewhere :)

agriyakhetarpal (approved):
Looks good to me, thanks, @hoodmane!

src/js/emscripten-settings.ts
hoodmane: Your code makes sense to me, let's try it.

ryanking13 (commented):
Nice!

src/js/emscripten-settings.ts
hoodmane: @agriyakhetarpal do you want to send this change as a pr to python or should I do it?

agriyakhetarpal: I would love to, thanks!

Issue #5075, opened 2024-09-14
`pyodide.code.find_imports` suboptimal for submodule packages


juntyr:
## 🐛 Bug

`pyodide.code.find_imports("import earthkit.maps")` will return `["earthkit"]`, the correct output in general. However, the earthkit package family consists of several packages where e.g. `earthkit-maps` provides the `earthkit.maps` module (there is also a top-level package that re-exports all child packages under the expected submodules).

In my custom Pyodide build that includes the `earthkit` packages, `earthkit-map`'s `meta.yaml` file declares

```yaml
top-level:
    - earthkit.maps
```

i.e. Pyodide knows that this package provides this specific import.

Would it be possible for `pyodide.code.find_imports` to consult the index of pre-built packages, and if one provides a submodule, only import this package instead of the top-level one (which then imports all child packages even when only one is needed)?

### To Reproduce

Go to https://lab.climet.eu/lab/index.html, create a new notebook, and use
```python
# cell 1
import pyodide
print(pyodide.code.find_imports("import earthkit.maps"))

# cell 2
# The install works correct in only installing what is needed
%pip install earthkit.maps

# cell 3
# The import, for which JupyterLite consults pyodide.code.find_imports, now triggers extra installs
import earthkit.maps
```

### Environment

- Pyodide Version: a [patched](https://github.com/climet-eu/lab/blob/main/patches/pyodide.patch) version of https://github.com/pyodide/pyodide/tree/6621b6bca72ed2cc4e9e66ed24783cce0e8dd907

ryanking13:
Thanks for the report @juntyr. The behavior of `find_imports` has been changed in https://github.com/pyodide/pyodide/pull/5039, and it will now return all child import paths.

```
>>> pyodide.code.find_imports("import earthkit.maps")
['earthkit', 'earthkit.maps']
```

> Would it be possible for pyodide.code.find_imports to consult the index of pre-built packages, and if one provides a submodule, only import this package instead of the top-level one (which then imports all child packages even when only one is needed)?

`pyodide.loadPackagesFromImports`(https://github.com/pyodide/pyodide/blob/5a2a0c7d3e7aeee6855abe81f54d6ef6677f4862/src/js/api.ts#L205-L237) works in that way.

juntyr:
Thanks for your reply! 

> ```
> >>> pyodide.code.find_imports("import earthkit.maps")
> ['earthkit', 'earthkit.maps']
> ```

[...]

PR #5412, opened 2025-02-04
Add public APIs to expose the lockfile

ryanking13:
### Description

Adds two new public JS apis: `pyodide.getLockfileInfo` and `pyodide.getLockfilePackages`. This API exposes the content of the lock file to the users.

As discussed in https://github.com/pyodide/pyodide/issues/5075#issuecomment-2357464498, I would like to expose the lockfile so that users can rewrite functions such as `pyodide.core.find_imports`.

Also, micropip relies on private lockfile attributes, and I would like to make micropip use only public APIs for long-term stability.

### Checklists

- [x] Add / update tests


src/js/api.ts
```diff
@@ -683,6 +683,24 @@ export class PyodideAPI {
     }
     return API.makeSnapshot(serializer);
   }
+
+  /**
+   * Returns the `info` section of the pyodide lockfile used to load the current Pyodide instance.
+   * The format of the lockfile is defined in the `pyodide/pyodide-lock
+   * <https://github.com/pyodide/pyodide-lock`_ repository.
+   */
+  static getLockfileInfo() {
```
hoodmane, line 692: How about making these getters? 
```
   static get lockfileInfo() {
      return API.lockfile_info;
   }
```

hoodmane: And maybe this could be `lockfilePlatform`? That seems a bit clearer to me.

src/js/api.ts
```diff
@@ -683,6 +683,24 @@ export class PyodideAPI {
...
+  }
+
+  /**
+   * Returns the `packages` section of the pyodide lockfile used to load the current Pyodide instance.
+   * The format of the lockfile is defined in the `pyodide/pyodide-lock
+   * <https://github.com/pyodide/pyodide-lock`_ repository.
+   */
+  static getLockfilePackages() {
+    return API.lockfile_packages;
+  }
```
hoodmane, line 703: Perhaps there is a name that sounds more related to `loadPackages`, `loadedPackages`? These are the packages that can be loaded by `loadPackages`. But maybe `lockfilePackages` is fine.

hoodmane (approved):
As I said, I think these should be getters. And we should take a moment to bikeshed the names. But +1 for adding these to the public API.

src/js/api.ts
ryanking13: Sounds good to me. A minor downside is that the info section contains `platform` attribute. So it would be  `lockfilePlatform().platform` to access that attribute.

ryanking13:
@hoodmane 

Actually, I just unified the two APIs into one: `lockfile`, which I think has a better name.

Issue #4057, opened 2023-08-10
FS.syncfs throws error on Chrome for iPad (iOS Chrome generally?)


paulcalcraft:
## 🐛 Bug

After writing a file to OPFS and then running pyodide.FS.syncfs with populate=true, to make the file available in pyodide, pyodide throws an exception.
This only happens after the second time you run the page. Presumably because the error is thrown when comparing existing files.
It throws even if the new file you're trying to write has a different name, so it's not exclusive to handling overwrites.

The error is: `TypeError: undefinedi s not an object (evaluating 'e.timestamp.getTime')`

I believe I've tracked it down to this line: https://github.com/pyodide/pyodide/blob/f8f026a5de496d71d1c6427e44ddcff8272f5dd4/src/js/nativefs.ts#L180

`e["timestamp"].getTime() > e2["timestamp"].getTime())`

`e["timestamp"]` or `e2["timestamp"]` returns undefined so the .getTime() call throws: `TypeError: undefinedi s not an object (evaluating 'e.timestamp.getTime')`

It may be that Chrome on iPad is not following the specification when it comes to writing/reading from OPFS, and actually all objects should have a timestamp?

### To Reproduce

Available to test in this CodePen here: https://codepen.io/paulcalcraft/pen/abQrroB
No error on desktop or Safari iPad, but iPad chrome does error the second time you run it (page refresh).

Apologies for the long MWE, but I believe it's all necessary because iPad only supports writing with createSyncAccessHandle, which needs to be on a web worker.

```
<!doctype html>
<html>

<head>
    <script src="https://cdn.jsdelivr.net/pyodide/v0.23.4/full/pyodide.js"></script>
</head>

<body>
    FS.syncfs Pyodide Error on Chrome for iPad<br>
    Logs from Pyodide and worker are shown below.<br><br>
    <script type="text/javascript">
        console.log = function (message) {
            document.body.innerHTML += message + "<br>";
        }
        console.error = function (message) {
            document.body.innerHTML += "<span style='color:red'>" + message + "</span><br>";
        }
        async function main() {
            console.log = function (message) {
                document.body.innerHTML += message + "<br>";
            }
            let pyodide = await loadPyodide();
            console.log(pyodide.runPython(`
            import sys
            sys.version
        `)); // 3.11.2 (main, Jul  7 2023, 05:19:00) [Clang 17.0.0 (https://github.com/llvm/llvm-project df82394e7a2d06506718cafa347b

[...]

Issue #5425, opened 2025-02-14
Deprecated lastModifiedDate property of the File interface causes errors in FS.syncfs and incorrect FSNode timestamps in Firefox


mstrahov:
## 🐛 Bug

When mounting OPFS in Firefox, FSNodes received from file handles have "timestamp: NaN".  Chrome does not have this problem, shows correct timestamps for FSNodes.    

After mounting OPFS and changing files on Firefox `pyodide.FS.syncfs()` returns: 
TypeError: N.timestamp is undefined
    reconcile nativefs.ts:209
    reconcile nativefs.ts:199
    syncfs nativefs.ts:52
    syncfs pyodide.asm.js:10
    syncfs pyodide.asm.js:10

Since File API File.lastModifiedDate property is [deprecated by Firefox and Safari](https://developer.mozilla.org/en-US/docs/Web/API/File/lastModifiedDate#browser_compatibility), it is probably better to use  `new Date(file.lastModified)`  instead of  `file.lastModifiedDate`  in nativefs.ts:

https://github.com/pyodide/pyodide/blob/34fcd02172895d75db369994011409324f9e3cce/src/js/nativefs.ts#L82

https://github.com/pyodide/pyodide/blob/34fcd02172895d75db369994011409324f9e3cce/src/js/nativefs.ts#L138

### To Reproduce
mount OPFS:
```
let rootOPFS = await navigator.storage.getDirectory();
let mountpointOPFS = await pyodide.mountNativeFS('/opfs', rootOPFS);
```
if files exist in OPFS root, they do not have timestamps in properties of FSNodes in:
`pyodide.FS.analyzePath("/opfs").object.contents`

property is shown as `timestamp: NaN`

after changing files / adding new files:
`pyodide.FS.syncfs(true, (err)=>console.log(err));`   does not sync, gives error:
TypeError: N.timestamp is undefined
    reconcile nativefs.ts:209
    reconcile nativefs.ts:199
    syncfs nativefs.ts:52
    syncfs pyodide.asm.js:10
    syncfs pyodide.asm.js:10

### Expected behavior

As in Chrome:
existing OPFS files have timestamps after mounting

FS.syncfs() runs without TypeError: timestamp 

### Environment

- Pyodide Version: 0.27.2
- Browser version: Firefox 128.7.0esr (64-bit)  
- Any other relevant information:  works in Chromium 133.0.6943.53  

### Additional context

[...]

PR #5426, opened 2025-02-15
lastModifiedDate fix for issue/5425

mstrahov:
### Description
This is a fix for issue #5425  

File API File.lastModifiedDate property is [deprecated by Firefox and Safari](https://developer.mozilla.org/en-US/docs/Web/API/File/lastModifiedDate#browser_compatibility), it is better to use new Date(file.lastModified) instead of file.lastModifiedDate in nativefs.ts for Firefox and Safari compatibility

Most likely it'll also fix issue #4057, but unable to test on Safari to confirm.


ryanking13 (approved):
Thanks! Could you please update [the changelog](https://github.com/pyodide/pyodide/blob/main/docs/project/changelog.md)?

hoodmane:
Thanks @mstrahov!

PR #5283, opened 2024-12-30
Fix build regressions for `orjson`, `cramjam`, and PyXIRR due to Maturin 1.8

agriyakhetarpal:
### Description

Through https://github.com/PyO3/maturin/pull/2391 via the recently released [Maturin version 1.8.0](https://github.com/PyO3/maturin/releases/tag/v1.8.0), builds for PyXIRR (added in #4513) and `orjson` (added in #4036) have been failing.

Builds for other packages such as `cramjam` have been failing as well due to the same reason.

I've applied this PR for PyXIRR, which I submitted upstream: https://github.com/Anexen/pyxirr/pull/64, as a patch here, which can be dropped later. Tagging @Anexen as the listed recipe maintainer for visibility.

For `orjson`, bumping to the [newly released version 3.10.13](https://github.com/ijl/orjson/releases/tag/3.10.13) should work. Tagging @spicy-sauce as the listed recipe maintainer for visibility, no action needed.

### Checklists

- [x] Add a [CHANGELOG](https://github.com/pyodide/pyodide/blob/main/docs/project/changelog.md) entry

PR #5427, opened 2025-02-15
Update sha256 checksum for pyxel recipe

ryanking13:
Looks like GitHub updated the file content, resulting in checksum change.

### DIFF
--- a/docs/usage/examples/console_webworker.html
+++ b/docs/usage/examples/console_webworker.html
@@ -72,7 +72,8 @@
         padding: 4px;
       }
       #description-header h1 {
-        font-family: "Franklin Gothic Medium", "Arial Narrow", Arial, sans-serif;
+        font-family:
+          "Franklin Gothic Medium", "Arial Narrow", Arial, sans-serif;
       }
     </style>
   </head>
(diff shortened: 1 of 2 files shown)


agriyakhetarpal (commented):
Thanks, @ryanking13! I would suggest that we should bump the `pyxel` version here to the newest one available (version 2.3.6), as they have fixed the constraint with `maturin` version 1.8 that we encountered in #5283: https://github.com/kitao/pyxel/blob/502ef41c5e508dfa08ecd17165471355dbdf6d4c/CHANGELOG.md#228.

I have reached out to the `pyxel` team via https://github.com/kitao/pyxel/issues/601 about the prospect of reinstating source distributions on PyPI.

ryanking13:
Sounds good. I updated the version, let's see if it builds well

agriyakhetarpal (approved):

agriyakhetarpal (commented):
I think we need to move into the `python/` directory, as they used to have `pyproject.toml` at the root level in version 1.9.10, but moved it to a separate directory in https://github.com/kitao/pyxel/tree/v2.3.6/python.

ryanking13:
> I think we need to move into the python/ directory, as they used to have pyproject.toml at the root level in version 1.9.10, but moved it to a separate directory in https://github.com/kitao/pyxel/tree/v2.3.6/python.

Right, Thanks. Let's handle it in a separate PR. As this is blocking all other PRs.

PR #5429, opened 2025-02-17
Add pytest --collect-only job to ci

hoodmane:
There's a surprising number of things to fix here. We suggest running `pytest` without any modifications here:
https://pyodide.org/en/stable/development/testing.html#running-the-python-test-suite
So probably we should check that test collection doesn't fail when you do that.


ryanking13 (approved):
Thanks! Looks like `test_tarfile` is often getting timeout these days.

hoodmane:
It's one of the longest python tests, and it got longer since I enabled more tests that I fixed upstream in Emscripten. So we ought to increase the context deadline for it.

PR #5433, opened 2025-02-18
Increase timeout for tarfile test

ryanking13:
This test often timeouts these days (in chrome and node). I am not sure when it started.


agriyakhetarpal (approved):
Thanks, @ryanking13!

ryanking13:
Thanks for the review!

hoodmane:
I think this started #5417 because I removed skips for subtests that I fixed in Emscripten. But some of those subtests are slow, hence the new timeouts.

PR #5434, opened 2025-02-18
Don't ignore errors when initializing the nativeFS

ryanking13:
### Description

`nativefs.synfs` can fail when there is a large (>=4GB) file in the file system. We were silently ignoring the error in syncfs, which resulted in the error reported in our Discord channel:

> hello folks 👋 ... we (PyScript) recently discovered and landed the lovely mountNativeFS API and, of course, tested all it's abilities and mechanisms to provide a decent user experience (that works in our MicroPython offer too). I've noticed during one of my tests that a folder with a 4GB+ file would result, after a syncfs() into that folder being fully erased, presumably due the in memory ArrayBuffer ability to store up to an unsigned Int32 data.

This PR updates `montNativeFS` to propagate the error happened during the `syncfs`. After this PR, mounting a directory with a large file would show the following error message.

image

cc: @WebReflection 

### Checklists

- [ ] Add / update tests


WebReflection:
Tested these changes locally via our port for MicroPython and it works correctly now, the error is shown and the folder is not erased: https://github.com/pyscript/polyscript/pull/129

Thank you very much for fixing this so quickly 🙏

src/js/api.ts
```diff
@@ -557,12 +557,28 @@ export class PyodideAPI {
...
+          resolve();
+        }
+      });
+    });
 
     return {
       // sync browser ==> native
       syncfs: async () =>
-        new Promise((resolve, _) => Module.FS.syncfs(false, resolve)),
+        new Promise((resolve, reject) =>
```
WebReflection, line 573: not sure how strict you folks are around *TS* but `<void>` is missing in this sync `false` case. True that it wasn't there in the first place and *TS* can infer that so I let you decide if you should make it consistent here or before, removing the other `<void>`.

src/js/api.ts
```diff
@@ -557,12 +557,28 @@ export class PyodideAPI {
...
-    await new Promise((resolve, _) => Module.FS.syncfs(true, resolve));
+    await new Promise<void>((resolve, reject) => {
+      Module.FS.syncfs(true, (err: any) => {
+        if (err) {
+          reject(err);
+        } else {
+          resolve();
+        }
+      });
+    });
```
hoodmane, line 568: Can we make helper functions at top level like:
```js
function syncfs(direction: bool): Promise<void> {
   return new Promise((resolve, reject) => {
      Module.FS.syncfs(true, (err: any) => {
        if (err) {
          reject(err);
        } else {
          resolve();
        }
      });
    });
}

async function syncBrowserToNative() {
    return await syncfs(true);
}

async function syncNativeToBrowser() {
    return await syncfs(true);
}
```

hoodmane (approved):
Looks good to me other than one suggestions.

src/js/api.ts
WebReflection: FWIWI that was my idea for the port but I didn't want to create branched code ... also you forgot to use `direction` in that example, it's always `true` and I believe that's not what you meant to write.

WebReflection: never mind ... it was already addressed!

ryanking13:
Thanks for the reviews everyone!

PR #5442, opened 2025-02-20
Add `loadPackageSetStd[out|err]` functions to override the default output handler for package loading

juntyr:
### Description

Add `loadPackageSetStd[out|err]` functions to override the default output handler for package loading. The default, `console.[log|error]` is unchanged. These functions are useful when you want to redirect the output but don't control every call to `loadPackage`. Calls to `loadPackage` that want to change the redirection just for their call can still do so.

### Checklists

- [ ] Add / update tests

### DIFF
--- a/docs/project/changelog.md
+++ b/docs/project/changelog.md
@@ -20,6 +20,8 @@ myst:
   {pr}`5334` {pr}`5363`
 - Added `jiter` 0.8.2 {pr}`5388`
 - Added the `context` parameter to `WebLoop.create_task()` {pr}`5431`
+- Added the `loadPackageSetStdout` and `loadPackageSetStderr` functions to override the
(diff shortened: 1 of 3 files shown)


juntyr:
?r @hoodmane

ryanking13:
Thanks for opening the PR, but I think it's worth taking some time to think about a more generic API.

We already have APIs related to [streams](https://pyodide.org/en/stable/usage/streams.html), or we can make a more general logging-related interfaces. So rather than adding two public APIs to modify the behaviour of one function, I'd like to take a different approach.

juntyr:
Sure, a more general approach sounds good. I just wanted to put this up as a suggestion as it was the MVP patch I needed in my fork.

PR #5445, opened 2025-02-21
Skip wasm-gc on iOS Safari where it's borked

### LINKED ISSUE #5428: Pyodide 0.27.1+ not working on iOS (Safari)
## 🐛 Bug

This bug seems similar to #5280, but for iOS/webkit browsers and a different version lower-bound. Loading Pyodide 0.27.1+ on iOS fails and crashes the browser.

(I believe this issue is also making select pages on the marimo docs unreliable on IOS — pages that load pyodide either crash, or the iframe-d marimo notebook fails with "Maximum call stack size exceeded".)

### To Reproduce

On iOS, safari or Chrome —

* Go to https://pyodide.org/en/stable/console.html
* Try to open versions 0.27.1, 0.27.2
* Both fail — either the loading status indicator spins indefinitely, or the page crashes with "A problem repeatedly occurred ..." 
* Older versions of Pyodide work fine.

Image

Similarly, the marimo docs intermittently fail:

* https://docs.marimo.io/api/inputs/slider/#marimo.ui.slider

Image

### Expected behavior

Doesn't crash.

### Environment

- Pyodide Version 0.27.1+
- Browser version iOS 18.3; Chrome 133.0.6943.84

### Additional context

Perhaps "the maximum call stack size exceeded" we occasionally see in marimo docs is a clue?

ambv:
As of iOS 18.3.1, enabling wasm-gc is making the interpreter fail to load. Upstream pull request: python/cpython#130418.

macOS Safari 18.3 does not surface the issue.

Confirmed on device that disabling this restores interpreter function.

Resolves #5428.

### DIFF
--- a/cpython/patches/0009-Skip-wasm-gc-on-iOS-Safari-where-it-s-broken.patch
+++ b/cpython/patches/0009-Skip-wasm-gc-on-iOS-Safari-where-it-s-broken.patch
@@ -1,4 +1,4 @@
-From 40e14aaf37dd758c457a12754a89fa5a17bfe7bc Mon Sep 17 00:00:00 2001
+From c627686651a46118b2a653fa61673dfc158442d4 Mon Sep 17 00:00:00 2001
 From: =?UTF-8?q?=C5=81ukasz=20Langa?= <lukasz@langa.pl>
 Date: Fri, 21 Feb 2025 19:24:41 +0100
 Subject: [PATCH 9/9] Skip wasm-gc on iOS Safari where it's broken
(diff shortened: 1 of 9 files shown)


agriyakhetarpal (approved):
Approving this as @hoodmane approved the upstream PR, thanks, @ambv!

hoodmane:
We will need to update the patch here once @freakboy3742's upstream review is addressed.

hoodmane:
Okay merging since upstream pr was merged. Thanks @freakboy3742 and @ambv!

cpython/patches/0009-Skip-wasm-gc-on-iOS-Safari-where-it-s-broken.patch
```diff
@@ -0,0 +1,59 @@
...
+--- a/Python/emscripten_trampoline.c
++++ b/Python/emscripten_trampoline.c
+@@ -79,7 +79,13 @@ EM_JS(CountArgsFunc, _PyEM_GetCountArgsPtr, (), {
+ //         i32.const -1
+ //     )
+ // )
+-addOnPreRun(() => {
++
++function getPyEMCountArgsPtr() {
++    let isIOS = globalThis.navigator && /iPad|iPhone|iPod/.test(navigator.platform);
```
WebReflection, line 27: ouch ... was this *really* an *iOS* only thing? I understand the ugly workaround but I wonder how comes *iOS* would branch that much from *iPad* and I had a honest little laugh at reading *iPod* in there 😁 (nostalgic personal hysterical 
laugh I believe)

freakboy3742: So - it's been reported on iOS 18.3.1; iPads running it (or... in theory, iPods, but I don't know if there's any actually still supported that could run iPod) should also be affected, but I can't say I've tried personally. 

And I completely agree that UA sniffing isn't an ideal solution - but the perfect is the enemy of the good until we're able to determine a better underlying fix based on sniffing specific behavior. At this point, all we know is "iOS 18.3.1(+?) bad", and we have a fallback.

PR #5446, opened 2025-02-22
DOCS Reorder the docs/development category

hoodmane:
The old ordering doesn't make a whole lot of sense:
* building-and-testing-packages is for out of tree, so it doesn't belong in development at all
* order the development by complexity / order that people will want it:
  - adding packages is simple so that first
  - then build, test, debug, contribute


ryanking13:
Thanks. Please sync the main branch to include https://github.com/pyodide/pyodide/pull/5444 which fixes the RTD build issue.

> building-and-testing-packages is for out of tree, so it doesn't belong in development at all

FYI, I am planning to merge in-tree and out-of-tree docs after #4987, as the two concepts will become ambiguous.

ryanking13 (approved):

hoodmane:
> merge in-tree and out-of-tree docs

Well there's still a distinction between "build from recipes" and "build from pyproject.toml". I would vote that we give new names to the concepts that make sense rather than merging these two different build processes.

ryanking13:
> Well there's still a distinction between "build from recipes" and "build from pyproject.toml". I would vote that we give new names to the concepts that make sense rather than merging the docs these two different build processes.

Yeah right. I think we still need to distinguish between recipe-based and src-based builds. But we should call them differently than in-tree and out-of-tree.

PR #5447, opened 2025-02-22
Instead of ??? put <unknown JS frame/file> in JS frames for Python error

hoodmane:
This is much clearer.

### DIFF
--- a/src/core/error_handling.ts
+++ b/src/core/error_handling.ts
@@ -284,8 +284,8 @@ Module.handle_js_error = function (e: any) {
     if (isPyodideFrame(frame)) {
       break;
     }
-    const funcnameAddr = stringToNewUTF8(frame.functionName || "???");
-    const fileNameAddr = stringToNewUTF8(frame.fileName || "???.js");
+    const funcnameAddr = stringToNewUTF8(frame.functionName || "<unknown JS frame>");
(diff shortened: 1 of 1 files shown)


src/core/error_handling.ts
```diff
@@ -284,8 +284,12 @@ Module.handle_js_error = function (e: any) {
     if (isPyodideFrame(frame)) {
       break;
     }
-    const funcnameAddr = stringToNewUTF8(frame.functionName || "<unknown JS frame>");
-    const fileNameAddr = stringToNewUTF8(frame.fileName || "<unknown JS file>.js");
+    const funcnameAddr = stringToNewUTF8(
+      frame.functionName || "<unknown JS frame>",
+    );
+    const fileNameAddr = stringToNewUTF8(
+      frame.fileName || "<unknown JS file>.js",
```
hoodmane, line 291:
```suggestion
      frame.fileName || "<unknown JS file>",
```

ryanking13 (approved):
Thanks. I agree that it is much clearer than `???`.

hoodmane:
There are no test failures so we apparently don't have test coverage for this?

ryanking13:
> There are no test failures so we apparently don't have test coverage for this?

Yeah I guess so.

PR #5452, opened 2025-02-22
Allow calling registerJsModule with a non extensible JS object

hoodmane:
Currently we can't use `registerJsModule` with an ES6 module:
```js
registerJsModule("jsfs", await import("fs"))
```
because we try to add a `__all__` property to the JS object but the ES6 module is non extensible. Instead of adding a `__all__`, teach `from mod import *` about how to get the properties of a JS module.

- [x] Add / update tests


ryanking13 (approved):
Thanks @hoodmane!

> because we try to add a __all__ property to the JS object but the ES6 module is non extensible. Instead of adding a __all__, teach from mod import * about how to get the properties of a JS module.

Indeed, it is a tricky problem to solve. I am a little bit sad that we will have another CPython patch that cannot be upstreamed, but it looks like a reasonable solution to me.

PR #5456, opened 2025-02-22
Unpolyglot the pyodide python cli entrypoint

hoodmane:
This splits `src/templates/python` into a bash script and a js script.

### DIFF
--- a/src/js/pyodide.ts
+++ b/src/js/pyodide.ts
@@ -46,6 +46,7 @@ export type ConfigType = {
   stdout?: (msg: string) => void;
   stderr?: (msg: string) => void;
   jsglobals?: object;
+  sysExecutable?: string;
   args: string[];
   _node_mounts: string[];
   env: { [key: string]: string };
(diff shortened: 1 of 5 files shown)


src/js/pyodide.ts
```diff
@@ -46,6 +46,7 @@ export type ConfigType = {
   stdout?: (msg: string) => void;
   stderr?: (msg: string) => void;
   jsglobals?: object;
+  sysExecutable?: string;
```
ryanking13, line 49: How about making it private? For now, I don't think it will be used in any places except for the venv.

ryanking13 (approved):
Looks nice. Thanks!

src/js/pyodide.ts
hoodmane: Well we already have public `args`, this is allowing to specify `argsv[0]`. I don't think many people will need it but it's also likely to be very stable. On the other hand there's also not much harm in making it private.

PR #5462, opened 2025-02-24
Backports for 0.27.3

hoodmane:
- [x] Insert the actual date in the changelog
- [x] Merge don't squash

### DIFF
--- a/docs/project/changelog.md
+++ b/docs/project/changelog.md
@@ -18,6 +18,8 @@ myst:
 
 _February 26, 2025_
 
+- Added the `context` parameter to `WebLoop.create_task()` {pr}`5431`
+
 ### Packages
 
 - Upgraded `narwhals` to 1.24.1 {pr}`5386`
--- a/docs/project/changelog.md
+++ b/docs/project/changelog.md
@@ -22,6 +22,7 @@ _February 26, 2025_
 
 ### Packages
 
+- Added `h3` 4.2.1 {pr}`5436`
 - Upgraded `narwhals` to 1.24.1 {pr}`5386`
 
 ## Version 0.27.2
--- a/docs/project/changelog.md
+++ b/docs/project/changelog.md
@@ -20,6 +20,8 @@ _February 26, 2025_
 
 - Added the `context` parameter to `WebLoop.create_task()` {pr}`5431`
 
+- {{ Fix }} `mountNativeFS` API now correctly propagates the error. {pr}`5434`
+
 ### Packages
 
 - Added `h3` 4.2.1 {pr}`5436`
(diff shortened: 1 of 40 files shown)


docs/project/changelog.md
```diff
@@ -14,6 +14,24 @@ myst:
...
+- {{ Fix }} `mountNativeFS` API now correctly propagates the error. {pr}`5434`
+- {{ Fix }} Since 0.27.1, Pyodide has been broken in iOS because iOS ships
+  broken wasm-gc support. Pyodide feature detects whether the runtime supports
+  wasm-gc and uses it if it is present. Unfortunately, iOS passes the feature
+  detection but wasm-gc doesn't work as expected. {pr}`5445`
+
+### Packages
+
+- Upgraded `narwhals` to 1.24.1 {pr}`5386`
+- Upgraded Pyxel to 2.3.6 {pr}`5427`
```
agriyakhetarpal, line 31:
```suggestion
```

I think this one should be removed as #5427 only updated the checksum for the project, and the upgrade is being tried out in #5435.

docs/requirements-doc.txt
```diff
@@ -15,7 +16,7 @@ sphinx-autodoc-typehints>=1.21.7
 sphinx-design>=0.3.0
 pydantic
 # TODO: separate document for pyodide-build?
-pyodide-build>=0.27.3
+-e ./pyodide-build
```
agriyakhetarpal, line 19: I think this would break the Dockerfile as we don't copy the submodules there (I use that locally, at least)

docs/project/changelog.md
hoodmane: Thanks for the explanation, I misunderstood.

docs/requirements-doc.txt
hoodmane: We can fix the Dockerfile later.

hoodmane:
Now there are arro3 failures...

agriyakhetarpal:
I think it's a failure in `orjson` itself. From the logs, it says

```
Error: Dynamic linking error: cannot resolve symbol yyjson_alc_pool_init
```

defined in https://github.com/ijl/orjson/blob/268b6608d095e80470895d9ed4f13aa863fb8640/src/ffi/yyjson.rs#L62

hoodmane:
Okay, so we need:
```yaml
build:
  script: export RUSTFLAGS="-Z link-native-libraries=yes"
```

hoodmane:
Okay I think maybe I finally fixed everything...

ryanking13 (approved):
Thanks @hoodmane!

PR #5471, opened 2025-02-25
Add tools/backports.py to help with making patch releases

hoodmane:
The main purpose of this script is to automate the changelog transformations involved in backports. Many cherry-picks lead to changelog conflicts, and we need also to rearrange the changelog on the main branch.

To make the backport branch, first we query the set of PRs that are tagged with 'needs-backport'. Then we sort them in chronological order by date merged. (This was annoying to do manually -- the github interface lets you sort PRs by creation date or last modified date but not by merge date).

Then we cherry-pick each commit in order by merge date but also render the change log automatically. If the cherry-pick succeeds, we write out the new change log and amend the commit. If it fails, we write out the change log and add it. We also check if pyodide-build is in conflict and if so I take the new `pyodide-build` commit that. Then we try `git cherry-pick --continue`. If it still fails, we abort and ask the user to resolve conflicts manually, run `git cherry-pick --continue` and then rerun the script. For this to work, we need to set `rerere.enabled` and `rerere.autoupdate`.

I implement a parser for the changelog to perform the transformations needed. This is the most complicated part, since we want to maintain the ordering and the structure while moving entries from one section to another.

I also added utilities for adding or removing the "needs backport" label and for showing which backport PRs are missing changelog entries.

See #5462 and #5463 for the result of this script.

See https://github.com/pyodide/pyodide/pull/5474 for new release instructions using this script.

### DIFF
--- a/tools/backport.py
+++ b/tools/backport.py
@@ -208,7 +208,7 @@ class ChangelogSection:
     paragraphs:
         The list of paragraphs.
 
-    paragraph:
+    cur_paragraph:
         Parser state.
     """
 
--- a/tools/backport.py
+++ b/tools/backport.py
@@ -21,8 +21,8 @@ import subprocess
 import sys
 from collections import namedtuple
 from copy import deepcopy
-from datetime import datetime
 from dataclasses import dataclass, field
+from datetime import datetime
 from pathlib import Path
 from typing import Any, Self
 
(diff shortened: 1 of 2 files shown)


hoodmane:
See https://github.com/pyodide/pyodide/pull/5474 for how I expect this to be used.

ryanking13 (approved):
Thanks! I didn't read the code thoroughly to be honest, but I think we can fix it anytime when there is a bug.

hoodmane:
Thanks!

PR #5477, opened 2025-02-27
Python CLI: Mount the host tmp directory

hoodmane:
This used to cause crashy behavior but due to an Emscripten bug but it was fixed a long time ago.

- [x] Add a changelog entry
- [x] Update tests


src/templates/python_cli_entry.mjs
```diff
@@ -9,18 +9,12 @@ import { readdirSync } from "fs";
...
-  const skipDirs = ["dev", "lib", "proc", "tmp"];
+function dirsToMount() {
   return readdirSync("/")
-    .filter((dir) => !skipDirs.includes(dir))
+    .filter((dir) => !["dev", "lib", "proc"].includes(dir))
     .map((dir) => "/" + dir);
 }
 
-function dirsToMount() {
-  const extra_mounts = process.env["_PYODIDE_EXTRA_MOUNTS"] || "";
```
ryanking13, line 20: If this variable is not used anymore, could you please remove it from the test? https://github.com/pyodide/pyodide/blob/1c29a7ce2ed29e470cf972a25df1aa86fe97da1c/src/tests/test_cmdline_runner.py#L174

ryanking13 (approved):
Thanks!

src/templates/python_cli_entry.mjs
hoodmane: Good catch. And it's nice that we have coverage for this so we know it's not breaking anything.

hoodmane:
This change broke the `test-cmdline-runner` CI job. I checked that it will work again by Emscripten 4.0.5. I think it was specifically fixed by: https://github.com/emscripten-core/emscripten/pull/23480/

I think given that we are trying to update Emscripten more soon, I won't revert this. We can just wait to pull in the fix. But it clearly shouldn't be backported, unless we also want to backport the file system fix onto 3.1.74...

PR #5479, opened 2025-02-27
WebLoop: Call exception handler on unhandled task rejection

hoodmane:
This has been causing annoying problems by hiding bugs in asyncio code for a long time.

- [x] Add / update tests


src/tests/test_asyncio.py
```diff
@@ -1,10 +1,13 @@
 import asyncio
+import re
 import time
+from textwrap import dedent
 
 import pytest
 from pytest_pyodide import run_in_pyodide
 
 from pyodide.code import eval_code_async
+from pyodide_build.build_env import get_pyodide_root
```
ryanking13, line 10: You can use `from conftest import ROOT_PATH`. Let's avoid using pyodide-build in the tests if possible.

ryanking13 (approved):
Thanks! I left a minor comment on test

src/tests/test_asyncio.py
hoodmane: There's 4 other uses of `get_pyodide_root()` in the test suite. You want to send a PR to replace them?

hoodmane: Also, it would be nice if `ROOT_PATH` were called say... `PYODIDE_ROOT`. Currently, it's not very greppable for me.

ryanking13: Sure, let me open a PR to replace all of them.

PR #5481, opened 2025-02-27
MAINT Add context manager for TS and use it in loadPackage

ryanking13:
### Description

This PR implements a Python-like context manager in Typescript, and apply it in the `loadPackage`.

The high-level function `withContext` can be used to create a context and can be used such as:

```typescript
const result = withContext(
    doSetup,
    doCleanup,
    doRealwork,
);
```

This context manager is applied to `loadPackage` to set `messageCallback` and `errorCallback` as a context. Previously we needed drill this parameters into child functions, but not a top-level context manager handles it.

There's no functional change for most people, but this makes loadPackage a little more thread-unsafe. But we never guaranteed thread safety for loadPackage, so I think it is not a problem.

### Checklists

- [x] Add / update tests


src/js/load-package.ts
```diff
@@ -599,8 +626,6 @@ if (typeof API !== "undefined" && typeof Module !== "undefined") {
   loadedPackages = singletonPackageManager.loadedPackages;
 
   // TODO: Find a better way to register these functions
-  API.recursiveDependencies =
```
ryanking13, line 602: @hoodmane I don't remember why we've exposed this API (https://github.com/pyodide/pyodide/pull/4555). Are you using it somewhere downstream?

hoodmane: Doesn't look like it. I didn't explain what I was up to in #4555 and you didn't ask so I guess we'll never know...

src/js/common/contextManager.ts
```diff
@@ -0,0 +1,56 @@
...
+/**
+ * Creates a function wrapper that sets up a context before calling the function
+ * and cleans up afterwards. Works with both synchronous and asynchronous functions.
+ *
+ * @param setup Function to call before the wrapped function
+ * @param cleanup Function to call after the wrapped function
+ * @returns A function that wraps another function with the context
+ */
+export function createContextWrapper(setup: () => any, cleanup: () => void) {
+  return <T extends (...args: any[]) => any>(fn: T) => {
```
hoodmane, line 39: I think the following should work:
```suggestion
  return withContext.bind(null, setup, cleanup);
```

src/js/test/unit/common/contextManager.test.ts
```diff
@@ -0,0 +1,212 @@
...
+const assert = chai.assert;
+const expect = chai.expect;
+
+import {
+  withContext,
+  createContextWrapper,
+} from "../../../common/contextManager";
+
+describe("withContext", () => {
+  it("synchronous execution", () => {
```
hoodmane, line 14: Can you add tests for when setup and cleanup raise? Synchronously and asynchronously?

hoodmane (approved):
Looks good to me aside from a few comments.

src/js/common/contextManager.ts
ryanking13: Oh, right. Good catch. I wrote `createContextWrapper` first, and extracted the common parts into `withContext`, but I forgot to update the `createContextWrapper` to use `withContext` 😂

src/js/common/contextManager.ts
```diff
@@ -0,0 +1,56 @@
...
+    if (result instanceof Promise) {
+      return result.finally(() => cleanup());
+    } else {
+      cleanup();
+      return result;
+    }
+  } catch (e) {
+    cleanup();
+    throw e;
+  }
```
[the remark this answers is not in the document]
ryanking13: Looks better, Thanks!

src/js/common/contextManager.ts
```diff
@@ -0,0 +1,56 @@
...
+ * @param setup A function that will be called to setup the context
+ * @param cleanup A function that will be called to cleanup the context
+ * @param callback A function that will be called with the context
+ * @returns The result of the callback function (or a Promise if callback is async)
+ */
+export function withContext<T>(
+  setup: () => any,
+  cleanup: () => void,
+  callback: () => T | Promise<T>,
+): T | Promise<T> {
```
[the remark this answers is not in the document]
ryanking13: Thanks, I tried this but it seems like TS complains about the return type becuase of `result.finally(...)` block.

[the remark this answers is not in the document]
hoodmane: You can use `\\ @ts-ignore` or a cast. It's important to give the function a correct type signature for call sites, how to fix the implementation to remove errors is less important.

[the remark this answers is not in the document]
ryanking13: Sounds reasonable to me. I casted the type in the finally block.

src/js/common/contextManager.ts
```diff
@@ -0,0 +1,50 @@
...
+ * @param setup Function to call before the wrapped function
+ * @param cleanup Function to call after the wrapped function
+ * @returns A function that wraps another function with the context
+ */
+export function createContextWrapper(
+  setup: () => void,
+  cleanup: () => void,
+): <T extends (...args: any[]) => any>(
+  fn: T,
+) => (...args: Parameters<T>) => ReturnType<T> {
```
hoodmane, line 44: Is this different from:
```suggestion
) => T {
```

hoodmane (approved):
Let's maybe go though some more rounds of nits on the typescript types but generally looks good to me.

src/js/load-package.ts
```diff
@@ -182,12 +183,29 @@ export class PackageManager {
     } = {
       checkIntegrity: true,
     },
+  ): Promise<Array<PackageData>> {
```
hoodmane, line 186:
```suggestion
  ): Promise<PackageData[]> {
```

src/js/load-package.ts
```diff
@@ -182,12 +183,29 @@ export class PackageManager {
...
+    const wrappedLoadPackage = this.setCallbacks(
+      options.messageCallback,
+      options.errorCallback,
+    )(this.loadPackageInner);
+
+    return wrappedLoadPackage.call(this, names, options);
+  }
+
+  public async loadPackageInner(
+    names: string | PyProxy | Array<string>,
```
hoodmane, line 196:
```suggestion
    names: string | PyProxy | string[],
```

src/js/common/contextManager.ts
```diff
@@ -0,0 +1,50 @@
...
+  return <T extends (...args: any[]) => any>(fn: T) => {
```
hoodmane, line 45: I don't think we need to repeat the type since it should be inferred as the return value?
```suggestion
  return (fn) => {
```

ryanking13:
> Let's maybe go though some more rounds of nits on the typescript types but generally looks good to me.

Yeah, thanks for the comments on the types. It is always difficult for me to deal with complex typescript types. I think I have simplified the types enough now. (Claude was way better than me at dealing with types 😂)

src/js/common/contextManager.ts
ryanking13: T refers to the original function, while `(...args: Parameters<T>) => ReturnType<T>` refers to (any) new function that has the same signature to T. So I think they are different.

hoodmane:
Looks great now.

ryanking13:
Okay, I'll go ahead and merge. Thanks for the reviews! There is a failure in `cmdline_runner`, but it is happening in the main branch too. I gues it is related to the recent pytest 8.3.5 release, but we'll need to take a look.

PR #5485, opened 2025-02-28
Packaging platformdirs

FrancescAlted:
### Description

A first packaging intent for `platformdirs` package. 

### Checklists

This is necessary for `blosc2` packaging effort happening in https://github.com/pyodide/pyodide/pull/5482.


FrancescAlted:
As suggested, will try to package this dependency right in https://github.com/pyodide/pyodide/pull/5482.

PR #5486, opened 2025-02-28
Makefile: Only install pyodide-build once

hoodmane:


agriyakhetarpal (approved):
This seems reasonable to me, thanks!

hoodmane:
But it also doesn't work.

PR #5487, opened 2025-02-28
Ndindex

FrancescAlted:
### Description

This an attempt at packaging `ndindex` library.  This is a dependency for `blosc2` (see PR https://github.com/pyodide/pyodide/pull/5482).


FrancescAlted:
As suggested, will try to package this dependency right in https://github.com/pyodide/pyodide/pull/5482.

PR #5488, opened 2025-03-02
TST Remove pyodide_build usage in test if possible

ryanking13:
As discussed in https://github.com/pyodide/pyodide/pull/5479#discussion_r1975262224, remove pyodide_build from core test suites if possible.

I untouched `test_cmdline_runner.py` which we will probably move to pyodide-build later.

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -11,10 +11,10 @@ from collections.abc import Sequence
 
 import pytest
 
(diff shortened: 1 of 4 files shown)


hoodmane (approved):
Thanks @ryanking13!

PR #5496, opened 2025-03-11
MAINT Use #embed instead of xxd

hoodmane:
This is faster and removes one external tool.


ryanking13 (approved):
Indeed it is much shorter!

PR #5497, opened 2025-03-11
MAINT Don't try to install pyodide-build from local files in Docker image

hoodmane:
We can also work around this with:

```
ADD pyodide-build /pyodide-build
RUN SETUPTOOLS_SCM_PRETEND_VERSION=0 pip3 install -r requirements-doc.txt
```
but I think it's more sensible just to split up the requirements file


ryanking13:
Oh, right. I thought I removed pyodide-build installation from the docker image, but the requirements file had that.

ryanking13:
How about removing the installation of `requirements-doc.txt` in the dockerfile? I don't see why we need to install it in the dockerfile.

hoodmane:
> How about removing the installation of requirements-doc.txt in the dockerfile? I don't see why we need to install it in the docker.

No objection to this.

ryanking13:
> No objection to this.

Yeah, I am +1 for removing it.

hoodmane:
Okay I made that change.

ryanking13 (approved):
Thanks!

PR #5498, opened 2025-03-11
Python 3.13

hoodmane:
- [ ] Add / update tests

Packages to fix:

- [ ] Pillow
- [ ] fiona
- [ ] mypy
- [ ] pygame-ce https://github.com/pyodide/pyodide/pull/5522
- [ ] zarr https://github.com/pyodide/pyodide/pull/5524/


Dockerfile
```diff
@@ -8,7 +8,7 @@ RUN apt-get update \
         patch pkg-config swig unzip wget xz-utils \
         autoconf autotools-dev automake texinfo dejagnu \
         build-essential libtool libltdl-dev \
-        gnupg2 libdbus-glib-1-2 sudo sqlite3 \
+        gnupg2 libdbus-glib-1-2 sudo sqlite3 cmake \
```
ryanking13, line 11: pyodide-build depends on `cmake` PyPI package, so one should be available to use cmake after installing pyodide-build. I wonder why cmake is necessary here.

packages/test/meta.yaml
```diff
@@ -16,52 +16,58 @@ build:
...
+      _testinternalcapi.c \
+      _testinternalcapi/test_lock.c \
+      _testinternalcapi/pytime.c \
+      _testinternalcapi/set.c \
+      _testinternalcapi/test_critical_sections.c \
     )
 
 
-    export TEST_MODULE_CFLAGS="${SIDE_MODULE_CFLAGS} -I Include/ -I Include/internal/ -I ."
+    export TEST_MODULE_CFLAGS="${SIDE_MODULE_CFLAGS} -I Include/ -I Include/internal/ -I . -I../../../../../cpython/installs/python-3.13.2/include/"
```
ryanking13, line 70: Does the new include directory contain different headers from `-I Include/`?

packages/test/meta.yaml
```diff
@@ -74,10 +80,21 @@ build:
...
     export TEST_CAPI_OBJECTS=( "${TEST_CAPI_SRCS[@]/#/Modules/}" )
     emcc  ${SIDE_MODULE_LDFLAGS} ${TEST_CAPI_OBJECTS[@]//.c/.o} -o ${PKG_BUILD_DIR}/_testcapi.so
 
+
+    export TEST_INTERNALCAPI_OBJECTS=( "${TEST_INTERNALCAPI_SRCS[@]/#/Modules/}" )
+    echo emcc ${SIDE_MODULE_LDFLAGS} ${TEST_INTERNALCAPI_OBJECTS[@]//.c/.o} -o ${PKG_BUILD_DIR}/_testinternalcapi.so
+    emcc ${SIDE_MODULE_LDFLAGS} ${TEST_INTERNALCAPI_OBJECTS[@]//.c/.o} -o ${PKG_BUILD_DIR}/_testinternalcapi.so
+
+    wasm-objdump -x Modules/_testinternalcapi/test_lock.o | grep _PyTestInternalCapi_Init_Lock
+    wasm-objdump -x ${PKG_BUILD_DIR}/_testinternalcapi.so | grep _PyTestInternalCapi_Init_Lock
```
ryanking13, line 96: What is this for?

hoodmane: Debugging.

Dockerfile
hoodmane: I think I added it because of the error we were getting in https://github.com/pyodide/pyodide/pull/5510.

src/core/python_unexposed.h
```diff
@@ -0,0 +1,83 @@
...
+  return _PyObject_VectorcallMethodId(name, args, nargsf, _Py_NULL);
+}
+
+extern PyObject*
+_PyObject_CallMethodIdObjArgs(PyObject* obj, _Py_Identifier* name, ...);
+
+extern int
+_PyObject_SetAttrId(PyObject*, _Py_Identifier*, PyObject*);
+
+// Other API stuff
```
hoodmane, line 46: @vstinner do you have recommended replacements for any of these?

vstinner: Use `PyObject_SetAttr()`. You might continue using `_Py_Identifier` using `_PyUnicode_FromId()`.

hoodmane: Makes sense for the Id ones. What about the following ones:
```C
_PyArg_CheckPositional
_PyArg_ParseStack
_PyArg_ParseStackAndKeywords
_PyGen_SetStopIterationValue
_PyGen_FetchStopIterationValue
_Py_HashBytes
_PyObject_NextNotImplemented
_PySet_Update
_PyUnicode_EQ
```

vstinner: `_PyArg_CheckPositional()` is gone, you have to reimplement it using other functions. Or use the internal C API (pycore_xxx.h headers). Or re-export it manually (hack! less reliable):

```
extern int _PyArg_CheckPositional(const char *, Py_ssize_t, Py_ssize_t, Py_ssize_t);
```

Same for other functions.

For `_Py_HashBytes()` and `_PyUnicode_EQ()`, you can use `Py_HashBuffer()` and `PyUnicode_Equal()` from [pythoncapi-compat](https://pythoncapi-compat.readthedocs.io/): they work on Python 3.13 and older.

hoodmane: I think we can just wait until we update to Python 3.14 and then switch to using `PyUnicode_Equal()` and `Py_HashBuffer()` since we aren't trying to be source compatible between versions.

ryanking13:
Which one do you want to land first, by the way? This or #5332?

hoodmane:
I'm not sure yet, you have an opinion?

ryanking13:
It looks like both PRs are quite ready (except for the disabled packages, but I think we can handle them in a separate PR). So I have no strong opinion.

PR #5501, opened 2025-03-13
Disable pygame which was broken by setuptools 76

hoodmane:
Adding `requirements.constraint: - setuptools < 76` doesn't help.


hoodmane:
cc @ryanking13.

PR #5507, opened 2025-03-14
Added osqp v1.1.0b4

vineetbansal:
### Description

Added recipe for [OSQP](https://osqp.org/), the widely used QP solver.

### DIFF
--- a/packages/osqp/meta.yaml
+++ b/packages/osqp/meta.yaml
@@ -1,0 +1,19 @@
+package:
+  name: osqp
+  version: 1.0.0.b4
+  top-level:
+    - osqp
+requirements:
+  run:
+    - numpy
+source:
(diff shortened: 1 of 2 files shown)


hoodmane:
I think this won't work on the main branch because we've updated the Emscripten version.  Maybe it would make the most sense to rebase this PR onto stable and target it there. If it passes there, then we can merge it to stable and add a disabled copy on the main branch but disable it until our ABI is stable again.

hoodmane:
I guess let's first see what happens in the CI here.

vineetbansal:
@hoodmane - thanks. I've closed this PR and opened PR #5510 instead.

hoodmane:
I retriggered the ci here since I'd like to see it failing for the right reason.

PR #5513, opened 2025-03-16
Updated python-sat to version 1.8.dev16.

alexeyignatiev:
The PR updates `python-sat` to version `1.8.dev16`. This update fixes arbitrary formula classification when it contains constants. It also brings a way to customise and use randomness features of Glucose 4.2.1.


ryanking13 (approved):
Thanks!

PR #5518, opened 2025-03-17
Changelog for v0.27.4

hoodmane:
- [x] Insert the actual date in the changelog

### DIFF
--- a/docs/project/changelog.md
+++ b/docs/project/changelog.md
@@ -26,6 +26,26 @@ myst:
   compile time and link time.
   {pr}`5320`
 
+### `python` CLI entrypoint
+
+- {{ Fix }} The `python` CLI now mounts the `/tmp` directory. In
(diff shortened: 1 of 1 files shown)


docs/project/changelog.md
```diff
@@ -26,6 +26,26 @@ myst:
   compile time and link time.
   {pr}`5320`
 
+### `python` CLI entrypoint
```
ryanking13, line 29: I think many users who don't know about `pyodide venv` would be confused about what we're referring to when we say `python CLI entrypoint`. How about making the title a little clearer (here and in 0.27.4 section)

ryanking13 (approved):
Thanks!

docs/project/changelog.md
hoodmane: Do you have a specific suggestion?

PR #5522, opened 2025-03-17
Update pygame-ce to 2.5.3

hoodmane:
Required for Python 3.13.

@ryanking13 build fails with:

```
../meson.build:38:4: ERROR: Problem encountered: The meson buildconfig of pygame-ce does not support emscripten for now.  However it may be added in the future.
```
Could you take this one over?


ryanking13:
Yes, will do. Please disable it for now.

hoodmane:
We're accumulating quite a lot of disabled packages =(

PR #5523, opened 2025-03-17
Update pyo3 in rust-abi-test and rust-panic-test

hoodmane:
Necessary for Python 3.13 support. Split from
https://github.com/pyodide/pyodide/pull/5498


ryanking13 (approved):
Thanks!

PR #5524, opened 2025-03-17
Update to zar 3.0.5

hoodmane:
Needed for Python 3.13 support, but it seems to be more async than previous versions, which is troublesome while JSPI is still not available in most runtimes.


packages/zarr/patches/0001-Don-t-use-threads.patch
hoodmane, line 1: TODO: make this patch upstreaming friendly.

juntyr: fsspec has a similar issue, perhaps we could find a generalisable solution that can be upstreamed to any package wanting to syncify async and support Pyodide?

ryanking13:
Closing for now as I unvendored the recipes.

PR #5534, opened 2025-03-19
DOCS Improve the `Getting Stated` section of the readme

hoodmane:
We didn't suggest anything for package maintainers, for people who want to use bundlers, or for people who want to use it in node. This provides some more suggestions for first links for people to look at depending on their needs.

### DIFF
--- a/docs/usage/index.md
+++ b/docs/usage/index.md
@@ -35,8 +35,10 @@ application.
 
 ```{note}
 To avoid confusion, note that:
- - `cdn.jsdelivr.net/pyodide/` distributes Python packages built with Pyodide as well as `pyodide.js`
- - `cdn.jsdelivr.net/npm/pyodide@0.19.0/` is a mirror of the Pyodide NPM package, which includes none of the WASM files
+ - `cdn.jsdelivr.net/pyodide/` distributes Python packages built with Pyodide as well
+    as `pyodide.js`
+ - `cdn.jsdelivr.net/npm/pyodide@0.19.0/` is a mirror of the Pyodide NPM package which
+    includes only the Pyodide runtime, not any of the wheels.
 ```
 
 ### Supported browsers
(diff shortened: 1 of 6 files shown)


docs/usage/index.md
```diff
@@ -35,8 +35,10 @@ application.
 
 ```{note}
 To avoid confusion, note that:
- - `cdn.jsdelivr.net/pyodide/` distributes Python packages built with Pyodide as well as `pyodide.js`
- - `cdn.jsdelivr.net/npm/pyodide@0.19.0/` is a mirror of the Pyodide NPM package, which includes none of the WASM files
+ - `cdn.jsdelivr.net/pyodide/` distributes Python packages built with Pyodide as well
+    as `pyodide.js`
+ - `cdn.jsdelivr.net/npm/pyodide@0.19.0/` is a mirror of the Pyodide NPM package which
```
hoodmane, line 40: Maybe we should make the version here update with the current Pyodide version too.

hoodmane: > which includes none of the WASM files

Should have always been "includes none of the wheel files".

docs/usage/index.md
```diff
@@ -73,40 +77,30 @@ Starting with Pyodide 0.25.0, Node.js < 18 is no longer officially supported.
 Older versions of Node.js might still work, but they are not tested or guaranteed to work.
 ```
 
-```{note}
-The following instructions have been tested with Node.js 18.5.0. To use
-Pyodide with older versions of Node, you might need to use  additional command line
-arguments, see below.
-```
```
hoodmane, line 80: I think this note is superseded by the note above it saying that older versions aren't supported.

docs/usage/index.md
ryanking13: > Maybe we should make the version here update with the current Pyodide version too.

sounds good to me.

ryanking13 (approved):
Thanks!

PR #4539, opened 2024-02-22
Packages: Add openai

hoodmane:
Resolves #3588. Requires a patch to httpx to work:

<details><summary>httpx patch</summary>
<p>

```
from contextlib import contextmanager

from httpx._client import AsyncClient, BoundAsyncStream, logger
from httpx._models import Headers, Request, Response
from httpx._transports.default import AsyncResponseStream
from httpx._types import AsyncByteStream
from httpx._utils import Timer

from js import Headers as js_Headers
from js import fetch
from pyodide.ffi import create_proxy


@contextmanager
def acquire_buffer(content):
    if not content:
        yield None
        return
    body_px = create_proxy(content)
    body_buf = body_px.getBuffer("u8")
    try:
        yield body_buf.data
    finally:
        body_px.destroy()
        body_buf.release()

async def js_readable_stream_iter(js_readable_stream):
    reader = js_readable_stream.getReader()
    while True:
        res = await reader.read()
        if res.done:
            return
        yield res.value.to_bytes()


async def _send_single_request(self, request: Request) -> Response:
    """
    Sends a single request, without handling any redirections.
    """
    timer = Timer()
    await timer.async_start()

    if not isinstance(request.stream, AsyncByteStream):
        raise RuntimeError(
            "Attempted to send an sync request with an AsyncClient instance."
        )


    js_headers = js_Headers.new(request.headers.multi_items())
    with acquire_buffer(request.content) as body:
        js_resp = await fetch(str(request.url), method=request.method, headers=js_headers, body=body)


    response = Response(
        status_code=js_resp.status,
        headers=Headers(js_resp.headers),
        stream=AsyncResponseStream(js_readable_stream_iter(js_resp.body)),
    )


    assert isinstance(response.stream, AsyncByteStream)
    response.request = request
    response.stream = BoundAsyncStream(
        response.stream, response=response, timer=timer
    )
    self.cookies.extract_cookies(response)
    response.default_encoding = self._default_encoding

    logger.info(
        'HTTP Request: %s %s "%s %d %s"',
        request.method,
        request.url,
        response.http_version,
        response.status_code,
        response.reason_phrase,
    )

    return response

AsyncClient._send_single_request = _send_single_request
```

</p>
</details> 

### Checklists

- [ ] Add a [CHANGELOG](https://github.com/pyodide/pyodide/blob/main/docs/project/changelog.md) entry
- [ ] Add / update tests
- [ ] Add new / update outdated documentation
- [ ] Find someone to put in `recipe-maintainers`

PR #5536, opened 2025-03-21
Add openai

hoodmane:
Supersedes #4539. Now in better shape since we have httpx support thanks to @joemarshall

- [x] Add / update tests


packages/openai/test_openai.py
```diff
@@ -0,0 +1,18 @@
...
+
+
+def test_openai(selenium):
+    from pathlib import Path
+
+    test_openai_inner = (Path(__file__).parent / "helper_test_openai.py").read_text()
+    helper_test_openai(selenium, test_openai_inner)
+
+
+@run_in_pyodide(packages=["openai", "pytest_httpx", "pytest-asyncio"])
```
ryanking13, line 11: Since `pytest_httpx` and `pytest-asyncio` are not direct dependencies of openai, let's wrap the test to be skipped if they are not built, similar to [duckdb test that relies on pandas](https://github.com/pyodide/pyodide/blob/6c4760ffabff12e31ca34f77b6096064ac971b03/packages/duckdb/test_duckdb.py#L41-L43)

packages/openai/meta.yaml
```diff
@@ -0,0 +1,23 @@
+package:
+  name: openai
+  version: 1.12.0
+  top-level:
+    - openai
+source:
+  url: https://files.pythonhosted.org/packages/26/a1/75474477af2a1dae3a25f80b72bbaf20e8296191ece7fff2f67984206f33/openai-1.12.0-py3-none-any.whl
```
ryanking13, line 7: The latest release of `openai`  seems 1.68.2, while 1.12.0 was released a year ago. Could you please update the version. If there is a reason we cannot use the latest version, could you add a comment?

pyodide-build
ryanking13, line 1: Was there a reason to update the pyodide-build version in this PR? If not, I would suggest updating it in a separate PR.

ryanking13 (approved):
Thanks! I left some minor comments.

pyodide-build
hoodmane: Yeah it was an accident.

packages/openai/test_openai.py
hoodmane: Should `run_in_pyodide` default to behaving this way locally and insisting that the packages are all installed in CI?

ryanking13: Maybe, or maybe not. While it would be handy to have such a feature, I'm worried that we'll misjudge that the test passed without realizing it is actually skipped.

hoodmane: We should be able to make it skip them in `test-packages-*-no-numpy-dependents` but fail them in the final `test-packages-*` runs. I think that would be an improvement over our current adhoc approach?

ryanking13: > We should be able to make it skip them in test-packages-*-no-numpy-dependents but fail them in the final test-packages-* runs. I think that would be an improvement over our current adhoc approach?

That sounds reasonable to me.

PR #5537, opened 2025-03-22
Disable verbose loggings for a few packages + upgrade lightgbm

ryanking13:
As title, I disabled verbose logging of a few packages. I think it has some effect on reducing the build time (lightgbm takes very long to build), but not 100% sure.


packages/lightgbm/meta.yaml
```diff
@@ -1,17 +1,12 @@
...
-  url: https://files.pythonhosted.org/packages/4d/e6/41be1f8642257e21b4170e798c9a84e4268656ebfa3019586d82bfd281c9/lightgbm-4.5.0.tar.gz
-  sha256: e1cd7baf0318d4e308a26575a63a4635f08df866ad3622a9d8e3d71d9637a1ba
+  url: https://files.pythonhosted.org/packages/source/l/lightgbm/lightgbm-4.6.0.tar.gz
+  sha256: cb1c59720eb569389c0ba74d14f52351b573af489f230032a1c9f314f8bab7fe
 build:
-  # skbuild.cmake.verbose should be switched to skbuild.build.verbose when
-  # upgrading to lightgbm>4.5.0 (the first release to set a minimum of scikit-build-core>=0.10)
   backend-flags: |
     skbuild.cmake.define.USE_OPENMP=OFF
-    skbuild.cmake.verbose=true
```
hoodmane, line 14: Instead of deleting it entirely can we keep a comment on how to enable this?

packages/opencv-python/meta.yaml
```diff
@@ -46,8 +46,6 @@ build:
   # Build configs can be found at: https://docs.opencv.org/4.x/db/d05/tutorial_config_reference.html
 
   script: |
-    # export VERBOSE=1
```
hoodmane, line 49: Same, keep the comment please.

hoodmane (approved):
Thanks!

PR #5538, opened 2025-03-23
DOCS Fix vite-plugin-static-copy mentions in bundler docs

kasbah:
Seems `vite-plugin-pyodide` isn't used anywhere but is still being mentioned.


hoodmane:
Thanks!

PR #5539, opened 2025-03-24
Add fsInit hook

hoodmane:
When we made the cli runner, we added a special case hook for initializing the environment so that virtual environments, .pth hooks, sitecustomize files, and other things that are processed on interpreter startup can be added. This makes such a mechanism available to everyone and switches the node cli runner to use the public mechanism.

This has a similar goal to #3316 but introduces the simplest possible interface.

cc @freakboy3742

- [x] Add / update tests


ryanking13:
Is there a downstream project that wants to use this hook? If not, how about making it a private flag so we can migrate it to a more general hook implementation later? Sorry that I always talk about this, but never got started on the implementation.

freakboy3742:
@ryanking13 Yes, there is a downstream project: BeeWare. This came out of a conversation @hoodmane and I had on a private channel; the underlying use case/problem is beeware/briefcase#2204.

hoodmane:
I don't see that it will be too hard to maintain this feature even if we later add a more general hook system later.

src/js/types.ts
```diff
@@ -261,6 +261,7 @@ export interface FSType {
   syncfs(dir: boolean, oncomplete: (val: void) => void): void;
   findObject(a: string, dontResolveLastLink?: boolean): any;
   readFile(a: string): Uint8Array;
+  sitePackages: string;
```
ryanking13, line 264: Not sure about this. Do we really want to extend `FS` object? If we want to make it public to users, adding it directly to the API would be better for documentation and visibility.

hoodmane: Well I want to give the `fsInit` function access to it but the api itself isn't ready yet at that point.

hoodmane: The other option I considered was just passing it as an argument. I do think we should expose it somewhere.

ryanking13: > The other option I considered was just passing it as an argument.

This sounds better to me. I would like to keep the object that comes from the Emscripten untouched as much as possible (Module is kind of an exception though)

hoodmane: Okay I will switch it.

hoodmane: We could avoid changing `Module` a lot more than we do by favoring putting stuff on `API` or other objects we make.

ryanking13: Sounds good. One issue is that functions that come from the native code become attributes of the `Module`, which inevitably leads to a lot of cases where we need to access the Module. I guess we can distinguish the _Emscripten_ Module and _Pyodide_ Modules in a typing perspective, if we adopt DefinitelyTyped.

docs/project/changelog.md
```diff
@@ -26,6 +26,10 @@ myst:
   compile time and link time.
   {pr}`5320`
 - {{ Enhancement }} Upgrade to Python 3.13.1. {pr}`5498`
+- {{ Enhancement }} Added `fsInit` argument to `loadPyodide()` to schedule a
+  hook to run to setup the file system. {pr}`5539`
+- {{ Enhancement }} Added `pyodide.FS.sitePackages` which is the path to the
+  `sitePackages` directory. {pr}`5539`
```
ryanking13, line 32:
```suggestion
```

ryanking13 (approved):
Thanks!

PR #5544, opened 2025-03-26
Make stack switching work in cli entrypoint

hoodmane:
- [x] Add / update tests


src/py/pyodide/webloop.py
```diff
@@ -642,4 +643,21 @@ def _initialize_event_loop():
...
+    from pyodide_js._api import config
+
+    if loop_factory is None and config.enableRunUntilComplete:
+        loop = asyncio.events._get_running_loop()
+        if isinstance(loop, WebLoop):
+            return loop.run_until_complete(main)
+    return _orig_run(main, debug=debug, loop_factory=loop_factory)
+
+
+asyncio.run = _run
```
hoodmane, line 660: @ambv any opinions?

src/py/pyodide/webloop.py
```diff
@@ -627,6 +628,20 @@ def set_event_loop(self, loop: Any) -> None:
...
 
 
+_orig_run = asyncio.run
+
+
+@wraps(_orig_run)
+def _run(main, *, debug=None, loop_factory=None):
+    from pyodide_js._api import config
+
+    if loop_factory is None and config.enableRunUntilComplete:
```
ryanking13, line 638: Just curious if there is something we can do when `loop_factory` is passed.

ryanking13 (approved):
Thanks!

src/py/pyodide/webloop.py
hoodmane: I think our options are we could ignore it or we could do this.

hoodmane:
Thanks for the review!

PR #5550, opened 2025-03-28
Remove test for `pyodide venv` failure when venv already exists

hoodmane:
@agriyakhetarpal  tells me that this is a nonstandard behavior.


agriyakhetarpal (approved):
Thanks, @hoodmane – I think either of us in our team could address https://github.com/pyodide/pyodide-build/issues/126 so that these tests run against tip-of-tree `pyodide`.

PR #5552, opened 2025-03-28
Replace all python3.12 to python3.13

ryanking13:
Renames all remaining parts after #5498

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -1,7 +1,7 @@
 {
   "name": "Docker",
   // keep in sync with "run_docker"
-  "image": "pyodide/pyodide-env:20250116-chrome132-firefox134",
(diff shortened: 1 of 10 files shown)


hoodmane (approved):
Thanks!

.devcontainer/devcontainer.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "Docker",
   // keep in sync with "run_docker"
-  "image": "pyodide/pyodide-env:20250116-chrome132-firefox134",
+  "image": "pyodide/pyodide-env:20250311-chrome134-firefox136-py313",
```
agriyakhetarpal, line 4: We should update `run_docker` as well.

agriyakhetarpal (approved):

.devcontainer/devcontainer.json
hoodmane: It looks like it's already updated in #5498.

PR #5555, opened 2025-03-29
FFI: Make sure we can call js callables that don't inherit from function

hoodmane:
Another weird edgecase. Useful for cloudflare because its rpc objects are non function callables.

- [x] Add / update tests

### DIFF
--- a/src/core/jslib.c
+++ b/src/core/jslib.c
@@ -339,11 +339,11 @@ EM_JS_BOOL(bool, JsvFunction_Check, (JsVal obj), {
 });
 
 EM_JS_VAL(JsVal, JsvFunction_CallBound, (JsVal func, JsVal this_, JsVal args), {
-  return nullToUndefined(func.apply(this_, args));
+  return nullToUndefined(Function.prototype.apply.apply(func, [this_, args]));
 });
 
 EM_JS_VAL(JsVal, JsvFunction_Call_OneArg, (JsVal func, JsVal arg), {
-  return nullToUndefined(func.apply(null, [arg]));
+  return nullToUndefined(func(arg));
 });
 
 // clang-format off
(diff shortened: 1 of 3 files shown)


hoodmane:
@WebReflection would this have any effect on the pyscript rpc?

src/core/jslib.c
```diff
@@ -339,11 +339,11 @@ EM_JS_BOOL(bool, JsvFunction_Check, (JsVal obj), {
 });
 
 EM_JS_VAL(JsVal, JsvFunction_CallBound, (JsVal func, JsVal this_, JsVal args), {
-  return nullToUndefined(func.apply(this_, args));
+  return nullToUndefined(Function.prototype.apply.apply(func, [ this_, args ]));
```
ryanking13, line 342: apply.apply looks very funny 🤔

ryanking13 (approved):
Very weird case indeed. Thanks for handling it.

src/core/jslib.c
hoodmane: We could pretty reasonably say `Function.prototype.apply.call(func, this_, args)` but it would look less funny.

WebReflection: but it'd be nicer on the *JS* *GC* 😇

WebReflection:
@hoodmane thanks for the ping, unfortunately I haven't seen this in time ... without much background and just a test case to satisfy, I am not sure I understand the use case for a callable not callable ... in case it has to do with proxies here a cheatsheet on how to have the right proxy for the job https://gist.github.com/WebReflection/58ce94a0d2c4118a4f3f26934c9582f4#traps-vs-proxy-types

could you please elaborate a bit why/how those functions get created? In any case, I don't think this affects us but I'll be happy to double check and confirm once this gets released and adopted in PyScript.

hoodmane:
Thanks for the reference @WebReflection it is indeed interesting.

PR #5560, opened 2025-03-31
Emscripten 4.0.6

hoodmane:


packages/pyxel/meta.yaml
```diff
@@ -1,6 +1,7 @@
 package:
   name: pyxel
   version: 1.9.10
+  _disabled: true
```
ryanking13, line 4: What error do we get for pyxel?

ryanking13 (approved):

packages/pyxel/meta.yaml
hoodmane: Build succeeds, but get function sig mismatch or null in several tests.

PR #5564, opened 2025-04-02
Include emscripten patches in xbuildenv

hoodmane:
In case we want to apply them when building

### DIFF
--- a/tools/create_xbuildenv.py
+++ b/tools/create_xbuildenv.py
@@ -85,6 +85,10 @@ def _copy_wasm_libs(
             shutil.copy(pyodide_root / path, xbuildenv_root / path)
 
 
+def _copy_emcc_patches(pyodide_root: Path, xbuildenv_root: Path):
+    shutil.copytree(pyodide_root / "emsdk/patches", xbuildenv_root / "emsdk/patches")
+
+
 def create(
     path: str | Path,
     pyodide_root: Path,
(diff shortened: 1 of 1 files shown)


agriyakhetarpal (approved):
Thanks, @hoodmane! Side note: do you know what step in `pyodide-build` would be appropriate if we would like to expose an option to users if they want to apply these patches?

agriyakhetarpal:
> > Side note: do you know what step in pyodide-build would be appropriate if we would like to expose an option to users if they want to apply these patches?
> 
> I'm leaving that as a TODO. This at least makes the change we need on the runtime side.

Sure, leaving it for a later time makes sense to me, @hoodmane. Should we also include a note in the xbuildenv that these patches are currently not being applied? The existence of these patches might make users think that we have already used a patched version of Emscripten, which would be misleading. Even if it's not, we could also add a basic README conveying the xbuildenv's structure. What do you think?

agriyakhetarpal:
The README could have a table similar to the one we have here: https://pyodide.org/en/stable/usage/downloading-and-deploying.html#contents-of-pyodide-github-releases

hoodmane:
> Should we also include a note in the xbuildenv that these patches are currently not being applied?

Well we definitely shouldn't put the note in the xbuildenv since the xbuildenv has no idea whether or not the patches are being applied. When we make a `pyodide install-emscripten` subcommand it will presumably apply the patches, and then such a note will become false.

It wouldn't hurt to document the xbuildenv contents but nobody is asking us questions about it so I wouldn't treat it as a high priority.

PR #5565, opened 2025-04-02
Add a file describing the Pyodide ABI

hoodmane:
Needed for:
https://github.com/python/peps/pull/4328

### DIFF
--- a/docs/development/abi.md
+++ b/docs/development/abi.md
@@ -59,8 +59,9 @@ static libraries may be used by shared libraries.
 
 ### pyodide_2024_0
 
-The Emscripten version is 3.1.58. The Python version is 3.12. Python 3.12 must
-be used at build time. All shared libraries must be linked with `-sWASM_BIGINT`.
+This ABI is intended for use with Python 3.12. The Emscripten version is 3.1.58.
+
+All shared libraries must be linked with `-sWASM_BIGINT`.
 
 By default, C++ libraries are built with exceptions disabled, and `throw` is an
 abort. The same is true for `setjmp`/`longjmp`. To enable exceptions and
@@ -78,8 +79,7 @@ dynamic library dependencies should be placed in the wheel in a folder called
 This section reflects the aspirational ABI for `pyodide_2025_0`. This is all
 subject to change without notice.
 
-The Emscripten version is 4.0.6. The Python version is 3.13. Python 3.13 must be
-used at build time.
+This ABI is intended for use with Python 3.13. The Emscripten version is 4.0.6.
 
 By default, emcc will compile and link C++ libraries with exceptions disabled,
 so that `throw` is an abort. The same is true for `setjmp`/`longjmp`. To enable
--- a/docs/development/abi.md
+++ b/docs/development/abi.md
@@ -19,8 +19,8 @@ underscode. For the smallest result, it is recommended to link with:
 
 To force all symbols to be exported, link with `-sSIDE_MODULE=1`.
 
-If `-pthread` is used at compile or link time, the resulting libraries will not
-load.
+`-pthread` must not be used at compile or link time. If `-pthread` is used, the
+resulting libraries will not load.
 
 To compile Rust packages, the following flags must be passed to `rustc`:
 
--- a/docs/development/abi.md
+++ b/docs/development/abi.md
@@ -17,33 +17,32 @@ underscode. For the smallest result, it is recommended to link with:
 -sSIDE_MODULE=2 -sEXPORTED_FUNCTIONS=["_PyInit_MyCModule1", "_PyInit_MyCModule2]
 ```
 
-If you want to force all symbols to be exported, link with `-sSIDE_MODULE=1`.
+To force all symbols to be exported, link with `-sSIDE_MODULE=1`.
 
 If `-pthread` is used at compile or link time, the resulting libraries will not
 load.
 
-To compile Rust packages, you must pass
+To compile Rust packages, the following flags must be passed to `rustc`:
 
 ```
 -C link-arg=-sSIDE_MODULE=2 -C link-arg=-sWASM_BIGINT
 ```
 
-to `rustc`. If you are not not invoking `rustc` directly, set these into the
-`RUSTFLAGS` environment variable.
+When not invoking `rustc` directly, for instance when using cargo, set these
+into the `RUSTFLAGS` environment variable.
 
 Compiling with `-sSIDE_MODULE=1` will not work with Rust because Rust libraries
 contain a `lib.rmeta` file which is not an object file. Rust produces the
 correct list of exported symbols automatically so this should not be a problem
 in practice.
 
-If you get linker errors, you can try passing `-Z link-native-libraries=no` as a
-`RUSTFLAG`. This is only necessary on Rust libc versions older than 0.2.162.
+On Rust libc versions older than 0.2.162, it may be necessary to pass
+`-Z link-native-libraries=no` as a `RUSTFLAG`.
 
 ### pyodide_2024_0
 
-The Emscripten version is 3.1.58. The Python version is 3.12. You must use
-Python 3.12 at build time. All shared libraries must be linked with
-`-sWASM_BIGINT`.
+The Emscripten version is 3.1.58. The Python version is 3.12. Python 3.12 must
+be used at build time. All shared libraries must be linked with `-sWASM_BIGINT`.
 
 By default, C++ libraries are built with exceptions disabled, and `throw` is an
 abort. The same is true for `setjmp`/`longjmp`. To enable exceptions and
@@ -61,8 +60,8 @@ dynamic library dependencies should be placed in the wheel in a folder called
 This section reflects the aspirational ABI for `pyodide_2025_0`. This is all
 subject to change without notice.
 
-The Emscripten version is 4.0.6. The Python version is 3.13. You must use Python
-3.13 at build time.
+The Emscripten version is 4.0.6. The Python version is 3.13. Python 3.13 must be
+used at build time.
 
 By default, C++ libraries are built with exceptions disabled, and `throw` is an
 abort. The same is true for `setjmp`/`longjmp`. To enable exceptions and
@@ -72,14 +71,15 @@ There is full support for `RPATH`. The dynamic loader will only load
 dependencies that are properly specified on the `RPATH`, just being in
 `/lib/python3.13/site-packages/wheel_name.libs` is not sufficient.
 
-For building Rust, it is necessary to use a Rust nightly after January 15th, 2025. You must pass the flag `-Z emscripten-wasm-eh`. You will also need a
+For building Rust, it is necessary to use a Rust nightly after January 15th, 2025.
+The flag `-Z emscripten-wasm-eh` must be passed. It is also necessary to use a
 compatible emscripten sysroot that has been built with wasm exception handling.
 Such sysroots are produced and distributed by
 [pyodide/rust-emscripten-wasm-eh-sysroot](https://github.com/pyodide/rust-emscripten-wasm-eh-sysroot).
-This is only distributed for Rust nightly-2025-02-01. If you need a different
-version, you can clone the `pyodide/rust-emscripten-wasm-eh-sysroot` repository
-and follow the instructions to build it in the README. To install the emscripten
-sysroot:
+This is only distributed for Rust nightly-2025-02-01. To use a different Rust
+nighly, it is possible to clone the `pyodide/rust-emscripten-wasm-eh-sysroot`
+repository and follow the instructions in the README to build a compatible
+sysroot. To install the emscripten sysroot use:
 
 ```sh
 rustup toolchain install nightly-2025-02-01
@@ -89,10 +89,10 @@ wget https://github.com/pyodide/rust-emscripten-wasm-eh-sysroot/releases/downloa
 tar -xf emcc-4.0.6_nightly-2025-02-01.tar.bz2 --directory=$RUSTLIB
 ```
 
-Note that this is all necessary _even if_ the crate you are building is build
-with `-Cpanic=abort` because the Rust standard library is not built with
-`-Cpanic=abort` and so you will encounter linker errors due to pulling in
-symbols from the standard library that use the wrong unwinding ABI.
+Note that this is all necessary _even if_ the crate uses `-Cpanic=abort` because
+the Rust standard library is not built with `-Cpanic=abort` and so there will be
+linker errors due to pulling in symbols from the standard library that use the
+wrong unwinding ABI.
 
 ## ABI sensitive flags
 
--- a/docs/development/abi.md
+++ b/docs/development/abi.md
@@ -4,40 +4,56 @@
 
 ### General
 
-Shared libraries must be linked with `-sSIDE_MODULE=1` or
+#### Building for the Emscripten target
 
-```
--sSIDE_MODULE=2 -sEXPORTED_FUNCTIONS=<export list>
-```
+To build C/C++ projects, use emscripten compiler toolchain `emcc`.
 
-Note that the name of each symbol in the list needs to be prefixed with an
-underscode. For the smallest result, it is recommended to link with:
+To build Rust projects, use `rustc --target wasm32-unknown-emscripten` or
+`cargo build --target wasm32-unknown-emscripten`. When building, `emcc` must be
+on the path or linking will fail.
+
+#### Making a shared library
+
+Emscripten shared libraries use the [the WebAssembly binary
+format](https://webassembly.github.io/spec/core/binary/index.html) and have a
+[dynamic linking section](https://github.com/WebAssembly/tool-conventions/blob/main/DynamicLinking.md).
+
+`emcc` will ignore the `-shared` flag. To make a shared library with `emcc`, you
+must pass `-sSIDE_MODULE=1` or `-sSIDE_MODULE=2`.
+
+To make a shared library with `rustc`, pass `-C link-arg=-sSIDE_MODULE=2`. To
+build a shared library with `cargo`, put `-C link-arg=-sSIDE_MODULE=2` in the
+`RUSTFLAGS` environment variable.
+
+#### No pthreads support
+
+`-pthread` must not be used at compile or link time. If `-pthread` is used, the
+resulting libraries will not load.
+
+#### Controlling the Set of Exported Symbols
+
+All symbols that form part of a binary module's interface must be exported. It
+is desirable to produce the minimal list of exported symbols to keep download
+size and runtime to a minimum. Not exporting symbols also reduces chances of
+symbol collisions.
+
+Linking a shared libraries with `-sSIDE_MODULE=1` will pass `-whole-archive` to
+`wasm-ld` and so force inclusion of all object files and all symbols. Linking
+with `-sSIDE_MODULE=2` will only include symbols that are explicitly listed with
+`-sEXPORTED_FUNCTIONS=<export list>`. The name of each symbol in the list must
+be prefixed with an underscode. For the smallest result, it is recommended to
+link with:
 
 ```
 -sSIDE_MODULE=2 -sEXPORTED_FUNCTIONS=["_PyInit_MyCModule1", "_PyInit_MyCModule2]
 ```
 
-To force all symbols to be exported, link with `-sSIDE_MODULE=1`.
-
-`-pthread` must not be used at compile or link time. If `-pthread` is used, the
-resulting libraries will not load.
-
-To compile Rust packages, the following flags must be passed to `rustc`:
-
-```
--C link-arg=-sSIDE_MODULE=2 -C link-arg=-sWASM_BIGINT
-```
-
-When not invoking `rustc` directly, for instance when using cargo, set these
-into the `RUSTFLAGS` environment variable.
-
+To compile Rust packages, `-C link-arg=-sSIDE_MODULE=2` must be passed to rustc.
 Compiling with `-sSIDE_MODULE=1` will not work with Rust because Rust libraries
 contain a `lib.rmeta` file which is not an object file. Rust produces the
-correct list of exported symbols automatically so this should not be a problem
-in practice.
+correct list of exported symbols automatically so this should not be a problem.
 
-On Rust libc versions older than 0.2.162, it may be necessary to pass
-`-Z link-native-libraries=no` as a `RUSTFLAG`.
+#### Libraries Linked to the Interpreter
 
 Pyodide is statically linked with the following libraries:
 
@@ -64,11 +80,24 @@ By default, all builds of the Pyodide runtime with Python 3.12 will use the
 
 The Emscripten version is 3.1.58.
 
+#### WASM_BIGINT
+
 All shared libraries must be linked with `-sWASM_BIGINT`.
+
+Since Rust 1.84.0 or nightly 2024-11-05, Rust will automatically pass
+`-Clink-arg=-sWASM_BIGINT`. In earlier versions of Rust it was required to pass
+`-Clink-arg=-sWASM_BIGINT` explicitly. Passing it explicitly will work in all
+cases.
+
+#### Unwinding ABIs
 
 By default, C++ libraries are built with exceptions disabled, and `throw` is an
 abort. The same is true for `setjmp`/`longjmp`. To enable exceptions and
 `setjmp`/`longjmp`, `-fexceptions` must be passed at compile time and link time.
+
+Rust will automatically enable unwinding.
+
+#### Runtime Library Loading Path
 
 Specifying an RPATH is not supported and will cause link errors. The dynamic
 loader has been patched so that all dynamic libraries in a wheel named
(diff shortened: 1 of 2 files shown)


docs/development/abi.md
```diff
@@ -0,0 +1,208 @@
+# Pyodide Platform ABI
```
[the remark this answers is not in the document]
hoodmane: >  ABI before the flags section

This makes sense, the flags are what you have to do. The flags section is just background information to explain what the flags mean, why do they exist, and why we use the settings that we do.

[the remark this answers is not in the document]
hoodmane: Thanks for the extremely fast review @AA-Turner!

docs/development/abi.md
```diff
@@ -0,0 +1,219 @@
...
+- `zlib`
+
+All of these come from Emscripten ports and the versions of these libraries are
+determined by the version of Emscripten we build with. Any symbols from these
+static libraries may be used by shared libraries.
+
+### pyodide_2024_0
+
+The Emscripten version is 3.1.58. The Python version is 3.12. Python 3.12 must
+be used at build time. All shared libraries must be linked with `-sWASM_BIGINT`.
```
freakboy3742, line 63: Would it be worth flagging that the 3.12 pin is a result of being a "pre-Tier 3" specification?

docs/development/abi.md
```diff
@@ -0,0 +1,219 @@
...
+
+This section reflects the aspirational ABI for `pyodide_2025_0`. This is all
+subject to change without notice.
+
+The Emscripten version is 4.0.6. The Python version is 3.13. Python 3.13 must be
+used at build time.
+
+By default, C++ libraries are built with exceptions disabled, and `throw` is an
+abort. The same is true for `setjmp`/`longjmp`. To enable exceptions and
+`setjmp`/`longjmp`, `-fwasm-exceptions` must be passed at compile time and link time.
```
freakboy3742, line 86: I'm a little unclear how "by default" applies here. What is applying this default? Is there any practical use case for C++ code that *doesn't* support exceptions?

docs/development/abi.md
```diff
@@ -0,0 +1,219 @@
...
+
+There is full support for `RPATH`. The dynamic loader will only load
+dependencies that are properly specified on the `RPATH`, just being in
+`/lib/python3.13/site-packages/wheel_name.libs` is not sufficient.
+
+For building Rust, it is necessary to use a Rust nightly after January 15th, 2025.
```
freakboy3742, line 92: Presumably, it's a nightly, *or* a stable released after that date? A formal spec that references a "nightly" seems a little risky. How long will it be until there is a stable that has the necessary flag?

docs/development/abi.md
```diff
@@ -0,0 +1,219 @@
...
+The flag `-Z emscripten-wasm-eh` must be passed. It is also necessary to use a
+compatible emscripten sysroot that has been built with wasm exception handling.
```
freakboy3742, line 94:
```suggestion
compatible Emscripten sysroot that has been built with WASM exception handling.
```

docs/development/abi.md
```diff
@@ -0,0 +1,219 @@
...
+Such sysroots are produced and distributed by
+[pyodide/rust-emscripten-wasm-eh-sysroot](https://github.com/pyodide/rust-emscripten-wasm-eh-sysroot).
+This is only distributed for Rust nightly-2025-02-01. To use a different Rust
+nighly, it is possible to clone the `pyodide/rust-emscripten-wasm-eh-sysroot`
+repository and follow the instructions in the README to build a compatible
+sysroot. To install the emscripten sysroot use:
```
freakboy3742, line 100:
```suggestion
sysroot. To install the Emscripten sysroot use:
```

docs/development/abi.md
```diff
@@ -0,0 +1,219 @@
...
+- `libeventloop.js`
+- `libGL`
+- `libhtml5_webgl.js`
+- `libhtml5.js`
+- `liblz4`
+- `libsdl.js`
+- `libwebgl.js`
+- `libwebsocket.js`
+- `bzip2`
+- `zlib`
```
freakboy3742, line 54: Although it isn't version specific at present, Is this list likely to be version specific? It might be worth flagging that it's explicitly "consistent on all currently ABI tags".

docs/development/abi.md
```diff
@@ -0,0 +1,219 @@
+# Pyodide Platform ABI
+
+## ABIs
+
+### General
+
+Shared libraries must be linked with `-sSIDE_MODULE=1` or
```
freakboy3742, line 7: Would it be better to phrase this in terms of "All Python module symbols that form part of a binary module's interface must be exported; the Emscripten linker can be instructed to do this by passing `-sSIDE_MODULE=1` to export *all* symbols, or ..."

i.e., capturing the high level requirement, rather than the specific interpretation?

docs/development/abi.md
```diff
@@ -0,0 +1,219 @@
...
+underscode. For the smallest result, it is recommended to link with:
+
+```
+-sSIDE_MODULE=2 -sEXPORTED_FUNCTIONS=["_PyInit_MyCModule1", "_PyInit_MyCModule2]
+```
+
+To force all symbols to be exported, link with `-sSIDE_MODULE=1`.
+
+If `-pthread` is used at compile or link time, the resulting libraries will not
+load.
```
freakboy3742, line 23: In terms of a spec - should this use RFC2119 language? i.e., "`-pthread` MUST NOT be used in the compilation or linking of a module.

docs/development/abi.md
```diff
@@ -0,0 +1,219 @@
...
+-C link-arg=-sSIDE_MODULE=2 -C link-arg=-sWASM_BIGINT
+```
+
+When not invoking `rustc` directly, for instance when using cargo, set these
+into the `RUSTFLAGS` environment variable.
+
+Compiling with `-sSIDE_MODULE=1` will not work with Rust because Rust libraries
+contain a `lib.rmeta` file which is not an object file. Rust produces the
+correct list of exported symbols automatically so this should not be a problem
+in practice.
```
freakboy3742, line 37: Two notes here:

1. It seems a little weird that this section is split from the previous SIDE_MODULE discussion by the `-pthread` discussion
2. This seems like a "how to make Rust do the thing that is a high level requirement" (i.e., exporting symbols). In terms of framing this as a specification, it seems like there should be a broad structure with a heading describing the high level requirement, and subsections that describe specifics of how that requirement manifests on specific compilers/linkers.

docs/development/abi.md
freakboy3742: I'd also suggest that each high-level requirement should be a separate heading, so that from a compliance point of view, it's clear that there are N requirements must be adhered to.

freakboy3742 (changes requested):
Looks fairly comprehensive; a few typos and request for clarifications inline.

The one broader comment is whether it might be better to structure some of the details differently. 

There's essentially 3 types of information in this document:
1. high level functional requirements (e.g., module symbols must be exported)
2. specific details of how those requirements manifest (e.g, pass these arguments to emcc/rustc)
3. Explanatory notes and historical context for specific decisions (e.g., exception handling is odd because...)

All this information is useful and on-topic; but from a strict "specification" point of view, only (1) is needed; and if someone is trying to implement something that is specification compliant, they will likely be "language focused". To that end, I wonder if it might be better to pull the language/compiler explanations into a separate section. For example, the high level specification will state that modules need to export all symbols; then would be an emcc section that has a subsection that describes the difference between `-sSIDE_MODULE=1` and `=2`, and a rustc section that has a subsection desribing the rust-specific interpretations.

docs/development/abi.md
```diff
@@ -0,0 +1,219 @@
...
+### pyodide_2024_0
+
+The Emscripten version is 3.1.58. The Python version is 3.12. Python 3.12 must
+be used at build time. All shared libraries must be linked with `-sWASM_BIGINT`.
+
+By default, C++ libraries are built with exceptions disabled, and `throw` is an
+abort. The same is true for `setjmp`/`longjmp`. To enable exceptions and
+`setjmp`/`longjmp`, `-fexceptions` must be passed at compile time and link time.
+
+Specifying an RPATH is not supported and will cause link errors. The dynamic
```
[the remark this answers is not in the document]
hoodmane: Well `pywasmcross.py` drops it. But I'm describing the behavior of `emcc` here.

docs/development/abi.md
hoodmane: Right, it's expected that it will eventually become version-specific and then I'll duplicate this list into each of the specific ABIs.

docs/development/abi.md
hoodmane: The plan is at some point the behavior it selects becomes the default behavior. We added it as an opt-in behavior so we had an opportunity to iron out any problems before forcing everyone to use it. Though we also have to solicit stakeholders.
https://github.com/rust-lang/compiler-team/issues/801

Anyways, there's nothing all that wrong with using a nightly, we can always pick a nightly from a day that also had a stable release and then they'll be basically the same.

docs/development/abi.md
hoodmane: emcc is applying this default. There are many complex C++ code bases that don't use exceptions. For example, v8 doesn't use exceptions. But I agree it is more common to use exceptions. We probably will have `pyodide-build` pass `-fwasm-exceptions` by default at some point, I think there is no good reason not to do this. But emcc chooses defaults that make the code small, and then requires you to explicitly opt into functionality that makes the code size bigger.

docs/development/abi.md
hoodmane: Stable releases already have the flag, but there is a check that errors out if you pass a nightly-only flag to a stable build of the compiler.

docs/development/abi.md
hoodmane: I don't think it has anything to do with that?

docs/development/abi.md
hoodmane: I like the style where we explain the consequences of particular actions rather than dictating what people do. People are free to build libraries with `-pthread`. They just won't work. I'm trying to explain here what works.

But if you have a strong position in favor of RFC2119, I would consider wording stuff that way.

docs/development/abi.md
```diff
@@ -0,0 +1,219 @@
...
+dynamic library dependencies should be placed in the wheel in a folder called
+`wheel_name.libs`.
+
+### pyodide_2025_0 (under development)
+
+This section reflects the aspirational ABI for `pyodide_2025_0`. This is all
+subject to change without notice.
+
+The Emscripten version is 4.0.6. The Python version is 3.13. Python 3.13 must be
+used at build time.
```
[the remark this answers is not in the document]
hoodmane: Yes, it's also possible to produce cp312 builds with the `pyodide_2025_0` abi. But nobody is shipping an interpreter with this abi. I changed it to say 

> This ABI is intended for use with Python 3.13.

But maybe saying "All builds of the Pyodide runtime with Python 3.13 will use the `pyodide_2025_0` abi" would be clearer.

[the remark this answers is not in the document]
freakboy3742: Yeah - that language makes more sense (possibly even adding "by default", leaving the window open for someone to produce a `pyodide_2028_0` build of Python 3.13 at some point in the future.)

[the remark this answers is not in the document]
freakboy3742: Also - should that be 3.14? None of the Tier-3 work is being backported to 3.13.

docs/development/abi.md
freakboy3742: I was mostly thinking of the weird "Tier 3 then not" status of Emscripten in Python 3.12/13/14. I guess adopting the "Python3.12 builds use the 2024_0 ABI by default" language used for the 2025_0 tag would be just as accurate.

docs/development/abi.md
freakboy3742: No objection to clarifying the consequences (that modules won't load); but adhering to the generally accepted language for a formal specification can only be beneficial IMHO. I'd suggest:

"`-pthread` must not be used at compile or link time. If `-pthread` is used, the resulting libraries will not load.

docs/development/abi.md
freakboy3742: What I'm flagging is that this language suggests that you *must* use a nightly. If a stable is also an option, then we should clarify what stable (and any clarifications about changes related to be being stable). Speaking for myself, I try to avoid using nightlies unless (a) it's the thing I'm actually working on (e.g., CPython), or there's *literally* no option but to use a nightly.

docs/development/abi.md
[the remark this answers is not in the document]
hoodmane: No we'll use `pyodide_2026_0` for 3.14.

docs/development/abi.md
hoodmane: If you don't want to use a nightly, there is an alternative of changing one byte in the source code of a stable rust and building it from source. I think it's better to use a nightly from the official channel.

hoodmane: I'll start discussing turning this on by default with the rust folks. That would make life a lot simpler.

docs/development/abi.md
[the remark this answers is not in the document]
freakboy3742: All the changes we're making in CPython right now are landing in main, which will become 3.14 in October of this year. 3.13 doesn't *officially* have any Emscripten support. What am I missing? Is the intention to define `2026_0` as soon as 3.14 is finalised in October to bake in the ABI specification?

docs/development/abi.md
freakboy3742: Well... modifying and recompiling a stable build means you're not really using a stable build :-)

It feels like I'm missing something about how nightlies work with Rust, though. My expectation was that if a Monday's nightly build accepts an `-Z foo` flag, then a stable build released on Tuesday would accept `-Z foo`. Based on what you're saying here, it sounds like that might not be the case? Is that a feature of `-Z` flags and/or how Rust uses them?

Regardless - if the requirement is in fact that you *must* (i.e., RFC2119 MUST) use a nightly at present, then that's fine - but my expectation as an outside observer is that if I have stable release from after the provided date, I would be safe. If that's *not* the case, then this specification needs to be doubly clear about the requirement that you *must* use a nightly, and not a stable (or a clarification that "a stable that uses `-Z emscripten-wasm-eh` as the default behavior" would be OK, to leave the door open for future stable builds that would be compatible, assuming that option is ever made the default).

docs/development/abi.md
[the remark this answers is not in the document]
hoodmane: No we wait 6 months or so after a Python version is released to support it because it takes a long time for packages to support the new Python version. We just upgraded to python 3.13 and I still had to disable a couple of packages with no 3.13 support. We want to be able to test out the abi against a large number of packages to work out any problems so that hopefully we won't have to break the abi to fix problems after the release.

docs/development/abi.md
hoodmane: > assuming that option is ever made the default

To be clear, from the rust proposal where I added this flag the plan was that we would first add it as a nightly only flag and then make it the default and then later delete the flag and the old behavior. The point of the flag was to let us fix any problems while it was still optional. I think it is working great -- there's tons of problems but none of them are related to this flag. 

I'll start discussing with them what I should do to move to the default on step. But it's hard to get to all of this... I think I'll add to the flags background section a bunch of this discussion we're having.

docs/development/abi.md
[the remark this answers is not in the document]
freakboy3742: I guess I see where you're coming from, but from a standards perspective, that seems to be putting the cart before the horse. Until you've *got* an ABI, it's impossible to know if packages will support that ABI. Python3.14's ABI is defined at time of release of 3.14.0; the 6 month beta period leading up to that release is when problems with that ABI are shaken out. 

I'll grant that Pyodide is in an awkward situation (as is iOS and Android) where we are having to take responsibility for "testing the entire ecosystem" - so to that end I could understand maybe having a "provisional" ABI flag or something like that - but not defining the standard *at all*? That seems the wrong way around to me.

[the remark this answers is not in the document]
hoodmane: I have no objection to defining a provisional ABI starting even when there are alpha releases of the new Python as long as it's clear that the provisional ABI will be changed until ~ 6 months after the new Python is released.

Also, we're not defining the ABI of libpython, we're defining the ABI for system headers, stack unwinding, JS/C ffi, that sort of thing. By comparison, glibc is allowed to make releases when it wants to and manylinux is adapted to that.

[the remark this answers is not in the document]
hoodmane: We can start by saying `pyodide_2026_0` is provisionally equal to `pyodide_2025_0` which is provisionally specified here.

docs/development/abi.md
hoodmane: Since this is about packaging, I'm not sure that the CPython support tier is relevant? pypy uploads wheels to PyPI but cpython has no support policy for pypy at all.

docs/development/abi.md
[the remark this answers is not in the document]
hoodmane: I am open to trying to freeze our ABI earlier and seeing if it works. I guess my questions are:

### What benefits does it provide? 

Package maintainers can immediately build with newly released Python for Pyodide platform. If we give them an interpreter they can also test against it. I think "Package maintainers can immediately build and test with newly released Python for Pyodide platform" is an important goal which will make life better for maintainers. 

An alternative to an earlier platform freeze would be to build and release an interpreter with the new Python and the old platform. We could give it some special version to declare that it is primarily intended for testing and ship it with no additional packages, or a super barebones set. This would have the benefit that it would let them fix problems caused by the new Python version separate from the new platform.

I'm not sure which approach would be better.

### How would it work?

If we want to make ABI changes ever, we need to make them at some point. We could pretty reasonably test out the new ABI on the old Python version. It would mean that at every point in time we would be using an ABI that's six months older than what I proposed, but six months is very short so maybe moving around a six month window isn't a big deal.

### What does it look like for Pyodide's maintenance effort?

I think what it means is that we have to maintain two ABIs at once for about 9 months instead of for 3 months. I think this will be a big maintenance headache for us, it is hard enough to do for 3 months. Compared to this, my plan of maintaining two Python versions for 9 months instead of 3 months (my alternative approach) would be much easier. It would also require dramatically less CI resources.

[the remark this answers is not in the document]
hoodmane: @henryiii would appreciate your opinion on this.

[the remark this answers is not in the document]
hoodmane: I think all of this sort of stuff could go into a separate `policies.md` page that explains what Pyodide does and why.

docs/development/abi.md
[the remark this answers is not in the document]
ryanking13: No, pywasmcross.py does not drop rpath, emcc does it. I changed that behavior in https://github.com/emscripten-core/emscripten/pull/23805/files#diff-a9cba4e03450a831038c69937bac7910c60b2a341f1eef13b70aaf4e21cb7c5d.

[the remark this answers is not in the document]
hoodmane: Ah, it seems that we drop `-Wl,-R` but not `-Wl,-rpath`. 
https://github.com/pyodide/pyodide-build/blob/main/pyodide_build/pywasmcross.py?plain=1#L194-L203

The docs for `-R` say:
```
  -R FILE, --just-symbols FILE
                              Just link symbols (if directory, same as --rpath)
```
wasm-ld does not accept `-R`. We should consider rewriting it to `-rpath`.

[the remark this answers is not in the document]
ryanking13: Oh, I didn't know about `-R` flags.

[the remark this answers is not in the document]
hoodmane: Opened an Emscripten issue about `-R`:
https://github.com/emscripten-core/emscripten/issues/24049

docs/development/abi.md
```diff
@@ -0,0 +1,221 @@
...
+```sh
+rustup toolchain install nightly-2025-02-01
+TOOLCHAIN_ROOT=$(rustup which --toolchain nightly-2025-02-01 rustc)
+RUSTLIB=$TOOLCHAIN_ROOT/lib/rustlib
+wget https://github.com/pyodide/rust-emscripten-wasm-eh-sysroot/releases/download/emcc-4.0.6_nightly-2025-02-01/emcc-4.0.6_nightly-2025-02-01.tar.bz2
+tar -xf emcc-4.0.6_nightly-2025-02-01.tar.bz2 --directory=$RUSTLIB
+```
+
+Note that this is all necessary _even if_ the crate uses `-Cpanic=abort` because
+the Rust standard library is not built with `-Cpanic=abort` and so there will be
```
juntyr, line 112: You can rebuild std with `-Z build-std=std,panic_abort -Z build-std-features=panic_immediate_abort` to enable panic=abort inside std as well

hoodmane: I've had trouble with making `-Z build-std -C panic=abort` work at all.

hoodmane: Maybe `-Z build-std-features=panic_immediate_abort` is what you need to fix that. But my feeling `-Z build-std` is too buggy to be usable.

hoodmane: But I will mention the possibility of using `-Zbuild-std` instead.

docs/development/abi.md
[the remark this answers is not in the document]
henryiii: Once there's a PyPI tag, then maybe Pyodide can start reducing the number of packages it maintains if upstream can build it instead. If we could reduce the number of packages that are slow supporting a new Python version by helping them get on PyPI, that might help Pyodide not lag behind so badly. Many packages support Python 3.x before release; you are supposed to make sure you support 3.x in the beta/rc phase. NumPy and other foundational packages all support Python before release. As a package author, I'd like to build all my new wheels for Python 3.x during the RC phase, and ship them - it would be annoying if I had to then do a dedicated release just for Pyodide 6 months later.

Maybe pyodide could even move to having a "core" (no extra packages) release and a "full" release, then the "core" release could follow CPython? This would be exactly what CPython itself would do if supported moved up to Tier 2, from what I understand?

The platform doesn't have to be tied to the Python version; that's just a nice feature to help package authors only have to ship one wheel per Python version. The full wheel tag includes `cp314` in it. The idea was to keep users from having to make multiple platforms for a single wheel, and since you want to keep emscripten up to date, that means matching the CPython release cadence. But if a platform extended over two Python versions, that would be perfectly fine, it just would have an out of date emscripten.

(Does pyodide support the Stable ABI, by the way?)

Anyway, I think clearly describing why there's a delay would help.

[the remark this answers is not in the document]
hoodmane: Thanks @henryiii

> Maybe pyodide could even move to having a "core" (no extra packages) release and a "full" release, then the "core" release could follow CPython?

I think this is probably the way to go.

> (Does pyodide support the Stable ABI, by the way?)

Yes, but not in a useful way if we keep changing the system ABI. For instance, our cryptography wheel is tagged `cryptography-44.0.2-cp37-abi3-pyodide_2025_0_wasm32.whl`.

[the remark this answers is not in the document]
hoodmane: > a "core" (no extra packages) release and a "full" release

There would be something to work out here in terms of how we would version these. @ambv suggested that Pyodide should start putting the Python version as a component of our versions, so that they could look like `12.27.0` or `27.12.0` or maybe with `312` instead. Then we could put some extra decoration on the core releases? So we'd release `314.28.0core` at the same time as `313.28.0`? I don't really like this that much but it seems like it might be workable.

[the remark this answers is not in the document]
AA-Turner: I'd suggest avoiding `312`/`313` if possible, the major and minor should be joined with a separator (we are no longer in the single-digit version era!). Perhaps `3-13` if you want to avoid another dot?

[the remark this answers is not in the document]
ryanking13: I am also open to freezing the ABI beforehand, and provide a way for package maintainers to build their packages against that ABI. I do worry about the maintenance burden part that Hood pointed out though. Pyodide always lacks human resource, and managing two ABIs is quite a burden.

However, package management is a big part of managing Pyodide right now (almost two-thirds of new issues are request for supporting new packages). If we were less responsible for managing packages, I think the maintenance burden could be lightened considerably. I've been working on gradually delegating package management over the past few months, and I think we will soon see the outcome (also @agriyakhetarpal did a lot of work to build and test packages in their CI). Once we start uploading packages to PyPI, we can significantly reduce the number of packages we build ourselves.

> Maybe pyodide could even move to having a "core" (no extra packages) release and a "full" release, then the "core" release could follow CPython?

Yes, I'm +1 for this direction. If we move away from package management, a single version of Pyodide will no longer be tightly bound to a specific “set” of packages. I think we can make a "core" release without packages and encourage people to generate a lock file with the packages they want. Of course, we can also provide a "full" lock file separately so people can easily use those packages, but it can go separate from the Pyodide version.

freakboy3742 (commented):
The structural changes look good to me; I've flagged a couple of minor typos.

The only substantial issue that seems outstanding is how a "stable ABI" will be defined and rolled out (in terms of when it it is defined, when it's provisional, when it's stable etc), and how those labels relate to a package being "legal" from the perspective of the formal wheel tag specification that will be referencing this document. However, that seems to be as much an internal Pyodide issue as anything else; as long as there's a clearly defined ABI that that the wheel tag spec can reference, my perspective is that it's up to Pyodide how to apply that.

PR #5569, opened 2025-04-03
Fix the fontlist.json file in matplotlib

mdboom:
### Description

Corrects the content of the hardcoded `fontlist.json` file in the `matplotlib` package.

Having an invalid file causes the font cache to be regenerated every time at startup, which by measurement is almost a full second.

There are two problems for the current file:

- It is for a future version of matplotlib, so the version number doesn't match and it gets rejected and regenerated.
- It includes files installed on a desktop system that won't exist in the pyodide environment.

I have also added documentation about how to update this file in the future.

- [ ] Add / update tests


agriyakhetarpal (approved):
Thanks, @mdboom! Maybe I updated the file from the wrong version in #4510...

hoodmane:
We can put this into 0.27.5.

mdboom:
> Thanks, @mdboom! Maybe I updated the file from the wrong version in #4510...

No worries!

hoodmane:
@mdboom the added test is failing so I'm going to drop it from the backports:
https://app.circleci.com/pipelines/github/hoodmane/pyodide/7232/workflows/e28d6975-dcde-4b8b-bf73-7320dfca1d44/jobs/111881/tests#failed-test-0

agriyakhetarpal:
Thanks @hoodmane, we could simplify that script further to make PEP-723-compliant that can fetch `matplotlib` and `pyyaml` for us, which would make it simpler. I assume the font list will not differ across the Linux and WASM builds. Can we get the Linux one and use it for WASM?

agriyakhetarpal:
Thanks, @hoodmane. We could simplify that script to make it PEP-723-compliant – it can fetch `matplotlib` and `pyyaml` for us and make the changes. I assume the font list will not differ across the Linux and WASM builds. Can we get the Linux one and use it for WASM?

hoodmane:
From @mdboom's post and explanation I assume that it does matter:

> It includes files installed on a desktop system that won't exist in the pyodide environment.

PR #5587, opened 2025-04-09
Fix a few recipes

ryanking13:
### Description

Fixes a few recipes that I found issues with when building out of tree.

### DIFF
--- a/packages/opencv-python/meta.yaml
+++ b/packages/opencv-python/meta.yaml
@@ -27,8 +27,8 @@ requirements:
     - libtiff
     - setuptools
   constraint:
-    # ImportError: The 'wheel.bdist_wheel' module has been removed.                   
-    # Please update your setuptools to v70.1 or later.    
+    # ImportError: The 'wheel.bdist_wheel' module has been removed.
+    # Please update your setuptools to v70.1 or later.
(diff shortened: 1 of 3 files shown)


packages/opencv-python/meta.yaml
```diff
@@ -26,6 +26,10 @@ requirements:
     - libwebp
     - libtiff
     - setuptools
+  constraint:
+    # ImportError: The 'wheel.bdist_wheel' module has been removed.
+    # Please update your setuptools to v70.1 or later.
+    - wheel < 0.46.1
```
agriyakhetarpal, line 32: This is easy to patch, as we only need to replace all instances of `wheel.bdist_wheel` in `opencv-python`'s sources to `setuptools.command.bdist_wheel.bdist_wheel`.

Please feel free to leave it for this PR if you'd like!

agriyakhetarpal (approved):
Thanks, @ryanking13!

packages/opencv-python/meta.yaml
ryanking13: That's true, but I would like to avoid patching the source code of the package as much as possible. I think updating the opencv-python version would work, but that requires more work.

PR #5588, opened 2025-04-09
Fix `thisArg` when making js calls while taking snapshot

hoodmane:
If we pass `_makeSnapshot: true` to `loadPyodide`, it wraps `globalThis` in this `makeGlobalsProxy` function so we can serialize top level imports. This breaks functions like `fetch` which in Chrome raises:
```
Failed to execute 'fetch' on 'Window': Illegal invocation
```
In order to fix this, before calling a function we should unwrap the target.

### DIFF
--- a/src/js/snapshot.ts
+++ b/src/js/snapshot.ts
@@ -16,6 +16,7 @@ API.getExpectedKeys = function () {
 };
 
 const getAccessorList = Symbol("getAccessorList");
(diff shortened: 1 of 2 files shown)


dom96:
Gentle ping on this

ryanking13:
Added 0.29.0 milestone so we don't forget

PR #5592, opened 2025-04-12
DOC Update maintainer doc about ABI update regarding package unvendoring

ryanking13:
This updates the maintainer doc to explain how to deal with ABI upgrade after we unvendor recipes.

This should be merged along with #4987. Opening this PR separately ad #4987 has too many diffs and it is difficult to review.


hoodmane (approved):
Thanks @ryanking13 . The main thing I'd like to see in addition to this is some discussion about how to test against a wip abi. Also maybe we should have branches of pyodide-recipes for each abi? But we can work out details like that in follow-up. This looks good to me.

ryanking13:
> The main thing I'd like to see in addition to this is some discussion about how to test against a wip abi. Also maybe we should have branches of pyodide-recipes for each abi? But we can work out details like that in follow-up. This looks good to me.

Yeah, it's probably going to take a lot of follow-up. Let's take it one step at a time.

hoodmane:
@ryanking13 should we merge this now?

ryanking13:
Yes, let's merge!

PR #5595, opened 2025-04-16
Emscripten 4.0.7

hoodmane:
Released today.

- [ ] Update changelog


ryanking13 (approved):
Thanks. I am enabling packages in pyodide-recipes. Hopefully I think we will be able to enable most of the disabled packages before the next release.

Issue #5597, opened 2025-04-17
Memory Leak when running await asyncio.sleep(0) inside a WebWorker


joellindegger:
## 🐛 Bug

Running `await asyncio.sleep(0)` inside a WebWorker seems to cause a memory leak. The leak isn't big per call, but when calling asyncio.sleep(0) frequently, it will become Gigabytes in seconds.

### To Reproduce

```
# must be run inside of a WebWorker, and not in Safari
n = 1000000
for i in range(n):
    await asyncio.sleep(0)  # Memory leak occurs here
```

The following HTML demonstrates the issue as a single files. On my Mac on Firefox, after 1M repetitions, over 3GB of memory are used by the WebWorker.
[minimal_bug_demo.html.txt](https://github.com/user-attachments/files/19796891/minimal_bug_demo.html.txt)

### Expected behavior

Memory usage remains constant for the above example.

### Environment

- Pyodide Version 0.27.5:
- Browser version Firefox 137.0.1 (aarch64):
- Any other relevant information:

### Additional context

Based on my experimentation, the underlying issue is due to how scheduleCallbackImmediate() is implemented for WebWorkers: [Link](https://github.com/pyodide/pyodide/blob/4fbbbedc09496c6968086d69aadba75398718b13/src/js/scheduler.ts#L81)
A new MessageChannel is created for each call to asyncio.sleep(). Although the approach seems correct, it seems to me that the MessageChannel objects are not cleared up by GC.

The following example demonstrates the aparent issue with this pattern:
```
      let count = 0;
      while (count < 1000000) {
        const channel = new MessageChannel();
        channel.port1.onmessage = () => { count++; };
        channel.port2.postMessage("");
      }
```

ryanking13:
Thanks for the report. I guess cleaning up the `onmessage` field would resolve the issue, such as:

```ts
    const channel = new MessageChannel();
    const cleanup = () => {
      channel.port1.onmessage = null; // Remove the event listener
    };
    channel.port1.onmessage = () => {
      try {
        callback();
      } finally {
        cleanup();
      }
    };
    channel.port2.postMessage("");
```

Could you please open a PR for it?

hoodmane:
Closed by #5599.

PR #5599, opened 2025-04-18
clean up MessageChannel on scheduleCallbackImmediate

joellindegger:
Addresses the memory leak reported in #5597

### Checklist


hoodmane:
We could also consider reusing Channels, I don't know if there is any way to do that that would be more efficient than this.

hoodmane (approved):
Thanks @joellindegger! Please add a changelog entry.

src/js/scheduler.ts
```diff
@@ -79,7 +79,12 @@ function scheduleCallbackImmediate(callback: () => void) {
     typeof MessageChannel === "function"
   ) {
     const channel = new MessageChannel();
-    channel.port1.onmessage = () => callback();
+    channel.port1.onmessage = () => {
+      callback();
```
WebReflection, line 83: unrequested nit: the original idea was based on `try/finally` which seems more appropriate https://github.com/pyodide/pyodide/issues/5597#issuecomment-2814504475

regardless, I would move this callback after the cleanup to be sure such cleanup happens regardless of the outcome of `callback()` also because if `callback()` somehow involves another scheduling the cleanup after will break that scheduling.

ryanking13: Thanks for the comment! I am also +1 for your opinions.

PR #5601, opened 2025-04-22
add ngspice recipe

pepijndevos:
### Description

I would like to simulate electronic circuits on the web, to do that I will need

- PySpice, a pure python library using cffi (minor fixes needed)
- NgSpice, a pure C spice simulator

So I've started adding a recipe based on the Conda ngspice recipe:
https://github.com/conda-forge/ngspice-feedstock/blob/main/recipe/meta.yaml

Changes from Conda:
- removed Windows/OSX details
- removed OpenMP and readline dependency
- unclear what to do with ncurses
- changed prefix to `WASM_LIBRARY_DIR`  based on openblas recipe

But so far `ctypes.util.find_library("ngspice")` doesn't find anything. The `.so` seems to be copied to some `.libs` folder, but in `dist` the `ngspice` package is completely empty, so I'm not sure what the correct way to do this is.

### Checklists

- [x] Add / update tests


pepijndevos:
ftr here is a script that builds a wasm ngspice outside of pyodide https://github.com/danchitnis/ngspice/blob/master/emcc/run.sh

Meanwhile this now builds but fails to load:
```
The following error occurred while loading ngspice:
need the dylink section to be first
```

pepijndevos:
I found https://github.com/pyodide/pyodide/pull/2637#issuecomment-1140345077 but adding that to my recipe did not change anything so I'm kind of out of ideas

packages/ngspice/meta.yaml
```diff
@@ -0,0 +1,41 @@
+package:
+  name: ngspice
+  version: "44.2"
+
+source:
+  url: https://downloads.sourceforge.net/project/ngspice/ng-spice-rework/44.2/ngspice-44.2.tar.gz
+  sha256: e7dadfb7bd5474fd22409c1e5a67acdec19f77e597df68e17c5549bc1390d7fd
```
ryanking13, line 7: We don't prefer sourceforge URL's as it is often quite unstable. It would be nice if you can use some other URL's such as GitHub release.

packages/ngspice/meta.yaml
```diff
@@ -0,0 +1,41 @@
...
+
+build:
+  type: shared_library
+  script: |
+    ./autogen.sh
+
+    configure_args=(
```
ryanking13, line 14: You'll probably set CFLAGs and LDFLAGs for Pyodide to generate proper shared library.

See libmagic example.

https://github.com/pyodide/pyodide/blob/4fbbbedc09496c6968086d69aadba75398718b13/packages/libmagic/meta.yaml#L23-L27

packages/ngspice/meta.yaml
pepijndevos: Unfortunately this project is primarily developed on Sourceforge, and indeed they move old versions into a folder, breaking the URL. There are some unofficial github mirrors but they don't publish releases to my knowledge.

You can see the conda recipe also uses this URL, and indeed if you try to run it you'll get a 404.

Would you prefer some tag from an unofficial mirror? eg
https://github.com/imr/ngspice/archive/refs/tags/ngspice-44.2.zip

packages/ngspice/meta.yaml
pepijndevos: Yea I tried that, and just committed those changes, but I still get the same error. I'm not sure how to debug this.

pepijndevos: I've been fiddling more with the ngspice build system but no progress.

pepijndevos:
It seems like ngspice is using libtool which ignores linking flags. It appears there is dedicated emscripten support in libtool which passie the correct flags, but it seems it's not using that.

packages/ngspice/meta.yaml
ryanking13: I think you can download the binary from Sourceforge, host it somewhere in your personal GitHub repository, and download it from there.

ryanking13:
```
```

I can see that log message in the build log, and yes it means SIDE_MODULE flag is not passed to the compiler correctly.

I think you need to check the configure script of ngspice and check how you can pass custom CFLAGS into it.

pepijndevos:
The key seems to have been to pass --host to a sufficiently modern version of libtool to make it use the emscripten settings.

Current status:
```
>>> import micropip
>>> await micropip.install("ngspice")
>>> import ctypes.util
>>> ctypes.util.find_library("ngspice")
'/usr/lib/libngspice.so'
```

Further testing tbd, please let me know any other recipe changes necessary.

packages/ngspice/meta.yaml
```diff
@@ -0,0 +1,42 @@
...
+      --disable-openmp
+      --with-ngshared
+      --host=wasm32-unknown-emscripten
+    )
+
+    mkdir release-lib && cd release-lib
+    emconfigure ../configure "${configure_args[@]}"
+    emmake make -j ${PYODIDE_JOBS:-3}
+    emmake make install
+    cp ${WASM_LIBRARY_DIR}/lib/libngspice.so ${DISTDIR}
```
agriyakhetarpal, line 30:
```suggestion
```

So, I think the next step is adding a recipe for pyspice and some related tests, in the same PR preferably. The ngspice recipe that you've added here should work in theory as long as it's compiled correctly. We could then copy the shared library from `WASM_LIBRARY_DIR/lib/` to its build directory then and patch it to look for it, rather than doing so at this stage.

packages/ngspice/meta.yaml
```diff
@@ -0,0 +1,42 @@
...
+
+requirements:
+  host: []
+
+  run: []
+
```
agriyakhetarpal, line 36:
```suggestion
```

agriyakhetarpal (commented):
Thanks, @pepijndevos!

packages/ngspice/meta.yaml
pepijndevos: I'm not sure I follow.

PySpice, or rather its more active fork InSpice is a pure Python package, but you want to add a recipe for it that depends on this ngspice package? Sure.

I don't fully understand the current and proposed library locations. It seems that distdir is what ends up actually in the zip, but are you saying InSpice should instead look in a different location?

pepijndevos:
My excitement was a bit premature as expected. It does load, but I think what is happening is libtool uses SIDE_MODULE=2 but things aren't correctly exported so the library is actually empty:

```
$ wasm-objdump -d libngspice.so 

libngspice.so:	file format wasm 0x1

Code Disassembly:

000059 func[0] <__wasm_apply_data_relocs>:
 00005a: 0b                         | end
```

packages/ngspice/meta.yaml
agriyakhetarpal: > PySpice, or rather its more active fork InSpice is a pure Python package, but you want to add a recipe for it that depends on this ngspice package? Sure.

Yes, otherwise there's no way to test if ngspice works in our CI. It will constrain the installable version of PySpice to what is included in the Pyodide distribution, but this is going to be temporary until we finish moving away the recipes to https://github.com/pyodide/pyodide-recipes.

> I don't fully understand the current and proposed library locations. It seems that distdir is what ends up actually in the zip, but are you saying InSpice should instead look in a different location?

Yes. I assume PySpice needs to load `libngspice.so` from ctypes, so we need to make it look at either `LD_LIBRARY_PATH` set to the dist directory (which may not exist or may vary) or in the current working directory next to the `.py` file that tries to load it?

pepijndevos:
I have added `EMSCRIPTEN_KEEPALIVE` to the API functions with a patch, riding on existing export functionality for Windows DLLs and such. Now the library is far from empty but upon loading I get an error:

```
Error: bad export type for '_ZTINSt3__210__function6__baseIFN5duals4dualIdEES4_S4_S4_EEE': undefined
```

Some more info:
```
$ c++filt _ZTINSt3__210__function6__baseIFN5duals4dualIdEES4_S4_S4_EEE
typeinfo for std::__2::__function::__base<duals::dual<double> (duals::dual<double>, duals::dual<double>, duals::dual<double>)>
$ wasm-objdump -d libngspice.so | grep _ZTINSt3__210__function6__baseIFN5duals4dualIdEES4_S4_S4_EEE
 055c7b: 23 10                      | global.get 16 <GOT.mem._ZTINSt3__210__function6__baseIFN5duals4dualIdEES4_S4_S4_EEE>
 055e55: 23 10                      | global.get 16 <GOT.mem._ZTINSt3__210__function6__baseIFN5duals4dualIdEES4_S4_S4_EEE>
```

This appears to be due to https://github.com/imr/ngspice/tree/master/src/include/cppduals which appears to be used in hicum2 models.

I'm not exactly sure what's happening here. Is there some kind of indirection here that causes the symbol to be not included? Am I just missing a whole cppduals.so? Do I just need to go play whack-a-mole with manually exporting missing symbols?

pepijndevos:
Holger Vogt has supplied me with a [patch](https://sourceforge.net/p/ngspice/patches/99/?page=1#40f9) to disable the hicum2 model that is causing trouble, and with that I have a shared library that loads and appears to contain all the relevant symbols.

```
$ wasm-objdump -x packages/ngspice/build/ngspice-44.2/dist/libngspice.so | grep ngSpice_Init
 - func[174] sig=7 <ngSpice_Init_Sync>
 - func[175] sig=25 <ngSpice_Init>
 - func[174] <ngSpice_Init_Sync> -> "ngSpice_Init_Sync"
 - func[175] <ngSpice_Init> -> "ngSpice_Init"
 - func[174] size=132 <ngSpice_Init_Sync>
 - func[175] size=5882 <ngSpice_Init>
  - 0002a40: 6e20 6e67 5370 6963 655f 496e 6974 2066  n ngSpice_Init f
```

But actually trying to call them doesn't work yet.

```
>>> from InSpice.Spice.NgSpice.Shared import NgSpiceShared
Matplotlib is building the font cache; this may take a moment.
>>> lib = NgSpiceShared()
  File "/lib/python3.12/site-packages/InSpice/Spice/NgSpice/Shared.py", line 496, in __init__
    self._init_ngspice(send_data)
  File "/lib/python3.12/site-packages/InSpice/Spice/NgSpice/Shared.py", line 579, in _init_ngspice
    rc = self._ngspice_shared.ngSpice_Init(self._send_char_c,
         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/lib/python3.12/site-packages/cffi/api.py", line 914, in __getattr__
    make_accessor(name)
  File "/lib/python3.12/site-packages/cffi/api.py", line 910, in make_accessor
    accessors[name](name)
  File "/lib/python3.12/site-packages/cffi/api.py", line 840, in accessor_function
    value = backendlib.load_function(BType, name)
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
AttributeError: function/symbol 'ngSpice_Init' not found in library '<None>': Tried to lookup unknown symbol "ngSpice_Init" in dynamic lib: __main__. Did you
 mean: 'ngSpice_Circ'?
```

So tomorrow we debug what is going on in the relevant code: https://github.com/Innovoltive/InSpice/blob/main/InSpice/Spice/NgSpice/Shared.py

packages/ngspice/meta.yaml
```diff
@@ -0,0 +1,40 @@
+package:
+  name: ngspice
+  version: "44.2"
+
```
ryanking13, line 4:
```suggestion
package:
  name: ngspice
  version: "44.2"
  tag:
    - library
    - shared_library
```

Adding tags helps our build system to arrange the build order.

packages/ngspice/meta.yaml
ryanking13: > PySpice, or rather its more active fork InSpice is a pure Python package, but you want to add a recipe for it that depends on this ngspice package? Sure.

Yes please. Adding only ngspice is not that useful and it is difficult to test its behavior, unless there is no corresponding Python package.

ryanking13:
> ```
> $ wasm-objdump -x packages/ngspice/build/ngspice-44.2/dist/libngspice.so | grep ngSpice_Circ
>  - func[181] sig=5 <ngSpice_Circ>
>  - func[181] <ngSpice_Circ> -> "ngSpice_Circ"
>  - func[181] size=748 <ngSpice_Circ>
> ```
> 
> 
>     
>   
> 
> What is going on here?

It means it is still building the static library instead of shared library. I believe the CFLAGS and LDFLAGS are still not passed to the compiler for some reason.

pepijndevos:
Here you go:
[build.log](https://github.com/user-attachments/files/20081636/build.log)
[ngspice-44.2.zip](https://github.com/user-attachments/files/20081644/ngspice-44.2.zip)

ryanking13:
What commit or version of Pyodide are you using to build the package? The tip-of-tree Pyodide uses Python 3.13, but based on your log, it seems like you are in Python 3.12.

pepijndevos:
I used the stable branch because inspice depends on scipy packages that are disabled on main

ryanking13:
> I used the stable branch because inspice depends on scipy packages that are disabled on main

I see. Thanks. I looked at the library, and the binary seems okay. It is a correct Emscripten shared library.

However, I noticed that some functions have an invalid signature,

For example, the `time` function in Pyodide runtime returns 64bit int, but

```
      FUNC	time	(i32) -> (i64)
```

int libngspice, it is defined to return 32bit int.

```
       env      FUNC    time    (i32) -> (i32)
```

I think it happens when [`WASM_BIGINT`](https://github.com/pyodide/pyodide/blob/85e70cc8566538f1b31c8ae0ae0aa2c12332fb72/Makefile.envs#L144) is not passed to the LDFLAGS.

pepijndevos:
It seems to default to true though: https://emscripten.org/docs/tools_reference/settings_reference.html#wasm-bigint

hoodmane:
It does default to true on Emscripten 4.x but stable Pyodide uses version 3.1.58.

pepijndevos:
For some reason my build randomly started failing with `/usr/bin/ld: cannot find /usr/lib64/libpthread_nonshared.a: No such file or directory` so I haven't been able to test this and am about ready to defenestrate my computer. I will hopefully get back to this problem asap.

pepijndevos:
> stable Pyodide uses version 3.1.58.

Soooo I just rebuilt the recipe on main and it works?!

At least up to this point:
```
>>> import micropip
>>> import ctypes
>>> await micropip.install("ngspice")
>>> lib = ctypes.cdll.LoadLibrary("libngspice.so")
>>> lib.ngSpice_Circ
<_FuncPtr object at 0x1163ec0>
```
I ran it in a patched docker container for newer libtool, I'll attach it to this PR as well I guess.

But that raises the question, how to proceed? Of course I need to add InSpice, but it wont actually be runnable until all the required packages are... not disabled.

agriyakhetarpal:
If a new Docker image with a newer libtool is needed, we should be able to do that. Otherwise, one way would be to rely on the CI to test out the changes – but it can be very slow, of course.

pepijndevos:
I have already pushed an updated Dockerfile.

It also appears that matplotlib etc now just build on main, so I am able to proceed with testing.

It appears that somehow the CFFI mechanism in InSpice does something different than ctypes, the `<None>` and `__main__` are a bit suspicious to me:

```
>>> import micropip
>>> await micropip.install("ngspice")
>>> await micropip.install("ply")
>>> await micropip.install("http://127.0.0.1:8000/inspice-1.6.3.1-py3-none-any.whl")
>>> from InSpice.Spice.NgSpice.Shared import NgSpiceShared
>>> ng = NgSpiceShared()
  File "/lib/python3.13/site-packages/InSpice/Spice/NgSpice/Shared.py", line 496, in __init__
    self._init_ngspice(send_data)
    ~~~~~~~~~~~~~~~~~~^^^^^^^^^^^
  File "/lib/python3.13/site-packages/InSpice/Spice/NgSpice/Shared.py", line 579, in _init_ngsp
ice
    rc = self._ngspice_shared.ngSpice_Init(self._send_char_c,
         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/lib/python3.13/site-packages/cffi/api.py", line 914, in __getattr__
    make_accessor(name)
    ~~~~~~~~~~~~~^^^^^^
  File "/lib/python3.13/site-packages/cffi/api.py", line 910, in make_accessor
    accessors[name](name)
    ~~~~~~~~~~~~~~~^^^^^^
  File "/lib/python3.13/site-packages/cffi/api.py", line 840, in accessor_function
    value = backendlib.load_function(BType, name)
AttributeError: function/symbol 'ngSpice_Init' not found in library '<None>': Tried to lookup u
nknown symbol "ngSpice_Init" in dynamic lib: __main__. Did you mean: 'ngSpice_Circ'?
>>> import ctypes
>>> lib = ctypes.cdll.LoadLibrary("libngspice.so")
>>> lib.ngSpice_Init
<_FuncPtr object at 0x335ad78>
```

pepijndevos:
Progress!
After fixing the InSpice linking code I now get `Error: Dynamic linking error: cannot resolve symbol invoke_iii`

From some research it seems like this could be related to... longjmp??
https://stackoverflow.com/questions/45511465/purpose-of-invoke-functions-generated-by-emscripten#64113301
https://emscripten.org/docs/porting/setjmp-longjmp.html

https://github.com/pyodide/pyodide/blob/c778309e8fcb593ceee5025b9c091ab275dce002/Makefile.envs#L145-L146

So we're right back to LDFLAGS getting ignored it seems. Oh no.

pepijndevos:
Yoooooo it's working!
```
Welcome to the Pyodide 0.28.0.dev0 terminal emulator 🐍
Python 3.13.2 (main, May  7 2025 14:37:17) on WebAssembly/Emscripten
>>> import micropip
await micropip.install("ngspice")
await micropip.install("ply")
await micropip.install("http://127.0.0.1:8000/inspice-1.6.3.1-py3-none-any.whl")
from InSpice.Spice.NgSpice.Shared import NgSpiceShared
import logging
logging.basicConfig(level=logging.DEBUG)
ng = NgSpiceShared(verbose=True)
DEBUG:InSpice.Spice.NgSpice.Shared.NgSpiceShared:Search library "ngspice"
DEBUG:InSpice.Spice.NgSpice.Shared.NgSpiceShared:Load library /usr/lib/libngspice.so
Load library /usr/lib/libngspice.so
DEBUG:InSpice.Spice.NgSpice.Shared:b"stderr Warning: can't find the initialization file spinit."
WARNING:InSpice.Spice.NgSpice.Shared.NgSpiceShared:Warning: can't find the initialization file spinit.
DEBUG:InSpice.Spice.NgSpice.Shared:b'stdout ******'
DEBUG:InSpice.Spice.NgSpice.Shared:b'stdout ** ngspice-44.2 shared library'
DEBUG:InSpice.Spice.NgSpice.Shared:b'stdout ** Creation Date: Fri May  9 08:04:28 UTC 2025'
DEBUG:InSpice.Spice.NgSpice.Shared:b'stdout ******'
DEBUG:InSpice.Spice.NgSpice.Shared.NgSpiceShared:Execute command: version -f
DEBUG:InSpice.Spice.NgSpice.Shared:b'stdout ******'
DEBUG:InSpice.Spice.NgSpice.Shared:b'stdout ** ngspice-44.2 : Circuit level simulation program'
DEBUG:InSpice.Spice.NgSpice.Shared:b'stdout ** Compiled with KLU Direct Linear Solver'
DEBUG:InSpice.Spice.NgSpice.Shared:b'stdout ** The U. C. Berkeley CAD Group'
DEBUG:InSpice.Spice.NgSpice.Shared:b'stdout ** Copyright 1985-1994, Regents of the University of California.'
DEBUG:InSpice.Spice.NgSpice.Shared:b'stdout ** Copyright 2001-2024, The ngspice team.'
DEBUG:InSpice.Spice.NgSpice.Shared:b'stdout ** Please get your ngspice manual from https://ngspice.sourceforge.io/docs.html'
DEBUG:InSpice.Spice.NgSpice.Shared:b'stdout ** Please file your bug-reports at http://ngspice.sourceforge.net/bugrep.html'
DEBUG:InSpice.Spice.NgSpice.Shared:b'stdout ** Creation Date: Fri May  9 08:04:28 UTC 2025'
DEBUG:InSpice.Spice.NgSpice.Shared:b'stdout **'
DEBUG:InSpice.Spice.NgSpice.Shared:b'stdout ** CIDER 1.b1 (CODECS simulator) included'
DEBUG:InSpice.Spice.NgSpice.Shared:b"stdout ** Relevant compilation options (refer to user's manual):"
DEBUG:InSpice.Spice.NgSpice.Shared:b'stdout ** X11 interface not compiled into ngspice'
DEBUG:InSpice.Spice.NgSpice.Shared:b'stdout **'
DEBUG:InSpice.Spice.NgSpice.Shared:b'stdout ** ngspice shared library.'
DEBUG:InSpice.Spice.NgSpice.Shared:b'stdout ******'
DEBUG:InSpice.Spice.NgSpice.Shared.NgSpiceShared:Ngspice version 44 with extensions: CIDER
DEBUG:InSpice.Spice.NgSpice.Shared.NgSpiceShared:Execute command: set nomoremode
```

My patches to ImSpice haven't been accepted yet though so I can't make a working InSpice recipe yet.

pepijndevos:
Alright I have added an inspice recipe with the open PRs attached as patches, as well as a basic test.

This also forced me to add two other pypi package skeletons that I have no affiliation or knowledge about, so I'm hesitant to be recipe maintainer for those.

As far as I'm concerned this is now good to go, the only thing I don't understand yet is what you want me to do with the shared library location, because if I remove the line copying it to dist it's not found by inspice.

pepijndevos:
FYI I've uploaded a little demo to my website: http://wishfulcoding.nl/pyodide/demo.html

Also, the error I'm seeing in CI is consistent with libtool being too old, which my patch to the Dockerfile fixes, but I guess CI is just pulling from docker hub instead of building the image on the fly?

hoodmane:
It is pulling from docker hub. I can push a new image based on your branch.

pepijndevos:
I've rebased the PR. Would it be helpful if I split the Dockerfile into a separate PR?

agriyakhetarpal:
> I've rebased the PR. Would it be helpful if I split the Dockerfile into a separate PR?

That would be helpful, thanks!

ryanking13:
This is handled in pyodide-recipes and it will be included in Pyodide 0.28

PR #5606, opened 2025-04-23
UPD: update rateslib to 1.8.0

attack68:


ryanking13:
Could you sync your PR with the main branch? The package wasn't bulding because there were some disabled dependencies. We just enabled a few packages that was disabled before, so it will work now

ryanking13 (approved):

ryanking13:
Thanks!

PR #5607, opened 2025-04-24
Add support for running in d8

hoodmane:
I'd have to check how much adding the shell environment increases code size to see if we want to support this by default. If not, we should at least have it as an option.

### Checklist

- [x] Add / update tests


ryanking13:
What is d8?

agriyakhetarpal:
It's apparently the debug-friendly version of v8: https://v8.dev/docs/d8

hoodmane:
I asked @erikcorry to look into the V8 codegen for Python particularly for ceval.c and see if he can make it faster. He asked me to make Pyodide run in d8 so he doesn't have to rebuild V8 and node on every change. So in particular it's not necessary to support this in the released builds. 

This doesn't work without https://github.com/emscripten-core/emscripten/pull/24185

PR #5609, opened 2025-04-25
Enable several disabled packages

ryanking13:
I tested this in [pyodide-recipes](https://github.com/pyodide/pyodide-recipes/pulls?q=is%3Apr+is%3Aclosed).


packages/matplotlib/test_matplotlib.py
```diff
@@ -58,6 +58,7 @@ def test_pdf(selenium):
     plt.savefig(fd, format="pdf")
 
 
+@pytest.mark.xfail(reason="FIXME")
```
ryanking13, line 61: This test was broken after #5569.

packages/Pillow/patches/0001-Enable-image-formats.patch
hoodmane, line 1: What's the story about why you were able to delete this patch?

packages/fiona/meta.yaml
```diff
@@ -1,8 +1,6 @@
 package:
   name: fiona
   version: 1.9.5
-  # Fails with CppException std::invalid_argument: stoi: no conversion
```
hoodmane, line 4: This just works? I guess maybe there was a regression in one Emscripten update and then another update fixed it.

packages/pysam/patches/0001-use-emmake.patch
```diff
@@ -32,11 +32,11 @@ index 22f3cb3..39da16d 100644
  
  
  def run_make_print_config():
--    stdout = subprocess.check_output(["make", "-s", "print-config"])
-+    stdout = subprocess.check_output(["emmake", "make", "-s", "print-config"])
-     if IS_PYTHON3:
-         stdout = stdout.decode("ascii")
+-    stdout = subprocess.check_output(["make", "-s", "print-config"], encoding="ascii")
++    stdout = subprocess.check_output(["emmake", "make", "-s", "print-config"], encoding="ascii")
```
hoodmane, line 36: Should this be covered by pywasmcross?

hoodmane:
Okay merging this. Thanks @ryanking13!

packages/Pillow/patches/0001-Enable-image-formats.patch
ryanking13: This patch was used to teach Pillow that libraries such as libjpeg, libpng are available in the system. While working on the fix, I noticed that Pillow has a logic to locate those libraries by looking at the CFLAGS and LDFLAGS, so I updated meta.yaml to inject the path in those flags.

packages/fiona/meta.yaml
ryanking13: Yep, I think it was a regression in LLVM, and it was fixed in another update.

packages/pysam/patches/0001-use-emmake.patch
ryanking13: I am not sure why pywasmcross was not covering this case. Maybe we can check that later.

PR #5610, opened 2025-04-26
[DRAFT] replace `loadDynamicLibrary` with `emscripten_dlopen` 

ryanking13:
### Description

This PR replaces `loadDynamicLibrary` with `emscripten_dlopen`. We were using emscripten's private `loadDynamicLibrary` API to load `.so` files but now we can use `emscripten_dlopen` as rpath is landed in emscripten (it is not yet released but will be released in emscripten 4.0.8).

### Checklists

- [ ] Add / update tests


conftest.py
```diff
@@ -88,7 +88,8 @@ def set_configs():
         let pyodide = await loadPyodide({
             fullStdLib: false,
             jsglobals: self,
-            _loadSnapshot: snap,
+            // ENABLE ME!
+            // _loadSnapshot: snap,
```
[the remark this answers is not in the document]
hoodmane: It's mysterious and annoying...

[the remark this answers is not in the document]
ryanking13: What I have found so far is that when there are multiple `.so` files in the wheel. The first one always succeeds, but the second one constantly crashes.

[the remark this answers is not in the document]
ryanking13: Okay, it seems like `Module.addFunction` is not compatable with snapshot for some reason.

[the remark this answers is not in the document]
hoodmane: Ah yes, `Module.addFunction()` doesn't mix that will with snapshots. But I still don't understand why your reproducer would have caused a problem.

src/js/dynload.ts
```diff
@@ -129,42 +36,32 @@ export class DynlibLoader {
...
+      const libUTF8 = this.#module.stringToNewUTF8(lib);
+
+      try {
+        Module.pyodidePromiseLibraryLoading = createResolvable();
+        this.#module._emscripten_dlopen_wrapper(libUTF8, flags);
+        await Module.pyodidePromiseLibraryLoading;
+      } catch (e: any) {
+        console.error(`Failed to load dynamic library ${lib}:`, e);
+      } finally {
+        Module.pyodidePromiseLibraryLoading = undefined;
```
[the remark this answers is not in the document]
ryanking13: Thanks, but it looks like the snapshot is causing trouble again 🤔

ryanking13:
Okay, I think I found the root cause.

Emscripten has an internal counter that determines whether the runtime should shut down when no tasks are running, and this is primarily for pthread, if I understand correctly, but is enabled for the non-pthread build as well.

This runtime counter is [also updated when the dlopen is finished](https://github.com/emscripten-core/emscripten/blob/0e1522f95d651f8ca6f603a4f70bf77c09e3fbf4/src/lib/libdylink.js#L1272-L1278):

```
    function successCallback() {
      {{{ runtimeKeepalivePop() }}}
      callUserCallback(() => {{{ makeDynCall('vpp', 'onsuccess') }}}(handle, user_data));
    }

    {{{ runtimeKeepalivePush() }}}
    var promise = dlopenInternal(handle, { loadAsync: true });
```

When snapshot is disabled, this runtime counter is correctly set. However, when snapshot is enabled, it appears that the snapshot is overwriting the counter with a different, invalid value.

So, when loading the package with snapshot enabled, the runtime counter becomes 0, and Emscripten thinks that all the tasks are finished, and [kills the runtime](https://github.com/emscripten-core/emscripten/blob/0e1522f95d651f8ca6f603a4f70bf77c09e3fbf4/src/lib/libcore.js#L2089-L2100).

```
    if (!keepRuntimeAlive()) {
#if RUNTIME_DEBUG
      dbg(`maybeExit: calling exit() implicitly after user callback completed: ${EXITSTATUS}`);
#endif
      try {
#if PTHREADS
        if (ENVIRONMENT_IS_PTHREAD) __emscripten_thread_exit(EXITSTATUS);
        else
#endif
        _exit(EXITSTATUS);
      } catch (e) {
        handleException(e);
```

ryanking13:
I think I fixed most of the issues now. The remaining one is that I am getting the `memory access out of bounds` in some pyodide venv tests...

```
handleException: unwinding: EXITSTATUS=0
    at pyodide.asm.wasm.PyMapping_GetOptionalItem (wasm://wasm/pyodide.asm.wasm-0242f872:wasm-function[793]:0x14ee07)
    at pyodide.asm.wasm.import_add_module (wasm://wasm/pyodide.asm.wasm-0242f872:wasm-function[4014]:0x2a5d62)
    at pyodide.asm.wasm.PyImport_AddModuleRef (wasm://wasm/pyodide.asm.wasm-0242f872:wasm-function[4013]:0x2a5cf1)
    at pyodide.asm.wasm._PyRun_SimpleStringFlagsWithName (wasm://wasm/pyodide.asm.wasm-0242f872:wasm-function[4467]:0x2d570b)
    at pyodide.asm.wasm.PyRun_SimpleString (wasm://wasm/pyodide.asm.wasm-0242f872:wasm-function[4491]:0x2d6b18)
    at Object.rawRun (/src/dist/pyodide.asm.js:38060:25)
    at API.finalizeBootstrap (/src/dist/pyodide.asm.js:38667:44)
    at loadPyodide (file:///src/dist/pyodide.mjs:861:23)
    at async main (file:///src/dist/python_cli_entry.mjs:38:10)
```

hoodmane:
Well it's not ideal to use `addFunction()` because it dynamically compiles a webassembly module every time it's called unless webassembly type reflection is available:
https://github.com/emscripten-core/emscripten/blob/0e1522f95d651f8ca6f603a4f70bf77c09e3fbf4/src/lib/libaddfunction.js#L109

I like the approach I used in my commit better, we should see if we can make it work.

hoodmane:
Perhaps `addFunction()` should also be changed to cache the webassembly module, then it would be significantly more efficient.

ryanking13:
> I like the approach I used in my commit better, we should see if we can make it work.

I see. I got an strange error: `TypeError: Chaining cycle detected for promise` when using `emscripten_dloopen_promise`, so I reverted back to addFunction, but let me try another approach...

src/js/dynload.ts
```diff
@@ -129,42 +35,31 @@ export class DynlibLoader {
...
       console.debug(`Loading a dynamic library ${lib} (global: ${global})`);
 
-    const fs = this.createDynlibFS(lib, searchDirs);
-    const localScope = global ? null : {};
+    let flags = RTLD_NOW;
+    if (global) {
+      flags |= RTLD_GLOBAL;
+    } else {
+      flags |= RTLD_LOCAL;
+    }
```
hoodmane: This does nothing?
```suggestion
    }
```

ryanking13: Yes, but I think it is better to be explicit.

ryanking13:
> I see. I got an strange error: TypeError: Chaining cycle detected for promise when using emscripten_dloopen_promise, so I reverted back to addFunction, but let me try another approach...

I think this was a bug in Emscripten. Opened another PR in emscripten: https://github.com/emscripten-core/emscripten/pull/24461

src/js/dynload.ts
hoodmane: What about a comment? I think it's weird to have this arithmetic expression that's a no-op. Does the C header also define `RTLD_LOCAL` as 0?

ryanking13: > Does the C header also define RTLD_LOCAL as 0?

Yes it does. I copied the values from [dlfcn.h](https://github.com/emscripten-core/emscripten/blob/d45b3957321d6a43cd96f380ba603381ccc4db06/system/lib/libc/musl/include/dlfcn.h#L10-L15).

Anyway, I also noticed that we are not using the `global` flag anymore, so there isn't much reason to define those constants. I removed the no-op expression and added a comment.

src/tests/test_package_loading.py
```diff
@@ -368,18 +368,16 @@ def test_install_archive(selenium):
...
-        pyodide.FS.writeFile("/a.so", new Uint8Array(4))
-        await pyodide._api.loadDynlib("/a.so");
-        """
-    )
-    assert (
-        "Failed to load dynlib /a.so. We probably just tried to load a linux .so file or something."
-        in selenium.logs
-    )
+    # If we load a bad so file, it will raise with a message
+    try:
```
hoodmane: Should this be `with pytest.raises()`? Since we'd want the test to fail if no exception is raised.

Makefile.envs
```diff
@@ -271,6 +272,7 @@ ifeq ($(DISABLE_DYLINK), 1)
 		,HEAPF64
 	export MAIN_MODULE_LDFLAGS= $(LDFLAGS_BASE) \
 		-s MODULARIZE=1 \
+		-s RELOCATABLE=1 \
```
hoodmane: What's this for? Isn't it implied by `-s MAIN_MODULE=1`?

emsdk/patches/0003-dylink-Fix-rpath-calculation-in-nested-dependencies.patch
```diff
@@ -0,0 +1,76 @@
+From c371fb11a1c5b557557b2681a9b6be4cdaa78644 Mon Sep 17 00:00:00 2001
+From: ryanking13 <def6488@gmail.com>
+Date: Fri, 30 May 2025 15:54:17 +0900
+Subject: [PATCH 1/1] [dylink] Fix rpath calculation in nested dependencies
+
+Upstream PR:
+* https://github.com/emscripten-core/emscripten/pull/24234
+* https://github.com/emscripten-core/emscripten/pull/24461
```
hoodmane: Could we keep these two as separate patches since they are separate upstream PRs?

src/js/dynload.ts
```diff
@@ -20,151 +19,33 @@ export class DynlibLoader {
...
-  public async loadDynlib(lib: string, global: boolean, searchDirs?: string[]) {
+  public async loadDynlib(lib: string) {
     const releaseDynlibLock = await this._lock();
 
-    DEBUG &&
-      console.debug(`Loading a dynamic library ${lib} (global: ${global})`);
-
-    const fs = this.createDynlibFS(lib, searchDirs);
-    const localScope = global ? null : {};
+    DEBUG && console.debug(`Loading a dynamic library ${lib}`);
```
hoodmane:
```suggestion
    DEBUG && console.debug(`Loading dynamic library ${lib}`);
```

src/js/dynload.ts
```diff
@@ -20,151 +19,33 @@ export class DynlibLoader {
...
-        {
-          loadAsync: true,
-          nodelete: true,
-          allowUndefined: true,
-          global,
-          fs,
-        },
-        localScope,
-      );
+      const libUTF8 = this.#module.stringToNewUTF8(lib);
```
hoodmane: I don't see that we free this. How about:
```js
const stack = this.#module.stackSave();
const libUTF8 = this.#module.stringToUTF8OnStack(lib);
```
and then call `stackRestore()` in a finally block.

hoodmane: Or I guess we should call `stackRestore()` just after the call to `_emscripten_dlopen_promise`.

Makefile.envs
ryanking13: It was needed for `addFunction` (unless RELOCATABLE=1 is set, the `wasmTable` is null), but we don't use it anymore, so I think I can remove it. Thanks for the catch!

emsdk/patches/0003-dylink-Fix-rpath-calculation-in-nested-dependencies.patch
ryanking13: Separated patches.

src/js/dynload.ts
ryanking13: Updated to use `stackSave`. I wonder which one is cheaper by the way, using stackSave and restore, or freeing the pointer.

hoodmane: I think this must be cheaper since stack allocation is a very small amount of arithmetic whereas malloc and free have to do more elaborate bookkeeping.

ryanking13: I see. Makes sense. Thanks!

ryanking13:
Thanks for the review! I'll merge this for now, and update the patch when Emscripten implements microtask queue (https://github.com/emscripten-core/emscripten/pull/24481)

PR #5612, opened 2025-04-28
suppress errors generated by `which` test in pyodide_env.sh

Lukasdoe:
Just a very small change. I noticed a lot of long error messages generated by the `which` command that fails to find `emcc.py`:

The build logs would then contain a lot of these lines:
```bash
which: no emcc.py in (<full-system-path>)
```
where the <full-system-path> easily spanns > 1k chars.

With this change, the error messages are redirected to /dev/null and the correct output is not redirected.


hoodmane (approved):
Makes sense. Thanks @Lukasdoe!

PR #5613, opened 2025-04-28
Add `eager_converter` arg to `to_js` to override default conversions

hoodmane:
The `default_converter` argument is called after the builtin conversions so it cannot be used to modify the normal behavior. @dom96 needed a way to block converting tuples. This adds a way to do this.

One API question is whether we want to allow the default behavior by calling `convert()` or by returning a special value. I have implemented it via `convert()` which leads to some slightly messy special case handling. It would be possible to do a fallback when `None` is returned, but what if the actual intention was to convert to `undefined`? Other special values would be hard to come by either in Python or in JS or both. So I think this is fine.

### Checklists

- [x] Add / update tests

### DIFF
--- a/src/core/pyproxy.ts
+++ b/src/core/pyproxy.ts
@@ -779,6 +779,14 @@ export class PyProxy {
       convert: (obj: PyProxy) => any,
       cacheConversion: (obj: PyProxy, result: any) => void,
     ) => any;
+    /**
+     * Optional callback to convert objects which gets called after `string`,
+     * `number`, `boolean`, `undefined`/`null`, and `PyProxy` are converted but
+     * _before_ any default conversions are applied to standard data structures.
+     *
+     * Its arguments are the same as `dict_converter`.
+     * See the documentation of :meth:`~pyodide.ffi.to_js`.
+     */
     eager_converter?: (
       obj: PyProxy,
       convert: (obj: PyProxy) => any,
--- a/src/core/pyproxy.ts
+++ b/src/core/pyproxy.ts
@@ -742,6 +742,7 @@ export class PyProxy {
     create_pyproxies = true,
     dict_converter = undefined,
     default_converter = undefined,
+    eager_converter = undefined,
   }: {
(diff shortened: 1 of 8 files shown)


src/core/pyproxy.ts
```diff
@@ -778,6 +779,11 @@ export class PyProxy {
       convert: (obj: PyProxy) => any,
       cacheConversion: (obj: PyProxy, result: any) => void,
     ) => any;
+    eager_converter?: (
```
ryanking13, line 782: I think this needs a proper docstring.

ryanking13 (approved):
Thanks! I think this needs a changelog.

src/core/python2js.c
```diff
@@ -486,6 +492,9 @@ _python2js(ConversionContext *context, PyObject* x)
   } else {
     context->depth--;
     JsVal result = _python2js_deep(context, x);
+    if (context->proxies && pyproxy_Check(result)) {
+      JsvArray_Push(hiwire_get(context->proxies), result);
+    }
```
dom96, line 497: What's the impact if this is missed? Curious because it seems this is no longer done after every call to `_python2js_deep` (should it be done in `python2js_custom__create_jscontext` too?)

dom96 (approved):
Nice!

src/core/python2js.c
hoodmane: Well it's for management of the proxy lifetime -- the list of proxies can be destroyed all at once. The leak checks in the test suite should automatically catch bugs here, so hopefully that means this isn't wrong.

> this is no longer done after every call to `_python2js_deep`

I'm not sure what you mean by this?

> should it be done in python2js_custom__create_jscontext too?

This list is for PyProxies that are created and handed over to the caller for them to manage. The `jscontext` is an implementation detail of the conversion itself and the `python2js` code destroys it manually when done converting.

dom96: > I'm not sure what you mean by this?

I just meant I didn't see the proxy getting added to the context in `python2js_custom__create_jscontext`. But what you say makes sense.

PR #5616, opened 2025-05-01
Emscripten 4.0.8

hoodmane:


emsdk/patches/0003-Revert-Disallow-exports-that-are-not-valid-C-C-ident.patch
```diff
@@ -1,59 +0,0 @@
-From f5cdcfe8a967b8bb4dd8c344cdbbd97e02c0e856 Mon Sep 17 00:00:00 2001
-From: Hood Chatham <roberthoodchatham@gmail.com>
-Date: Mon, 31 Mar 2025 13:33:46 +0200
-Subject: [PATCH 3/3] Revert "Disallow exports that are not valid C/C++
```
ryanking13, line 4: Did upstream revert this change?

hoodmane: Pretty sure they did yes.

hoodmane: Looks like maybe they didn't: 
https://github.com/emscripten-core/emscripten/pull/23563

I guess we'll see in ci if it's still needed.

ryanking13: Thanks. I think we'll need it then. I faced an error when building Rust packages in pyodide-recipes without that patch.

hoodmane: Hopefully they'll fix it for the next release.

ryanking13 (approved):
Thanks!

It seems like test-python job has been failing for a few commits ago.

hoodmane:
I think the issue is that there are some ansi control sequences in the string.

PR #5618, opened 2025-05-05
Some minor clean up in core c code (NFC)

hoodmane:
This pulls in the latest hiwire commit which replaces `hiwire_incref_deduplicate` with `hiwire_new_deduplicate`.

We also added `(void)` to some 0-arg functions and did some tidying in `js2python.js`.


ryanking13 (approved):
Thanks. There seems like a failure in package test, not sure if it is related to this PR

hoodmane:
I don't understand how it could be... But it did happen multiple times.

PR #5621, opened 2025-05-07
Make loadPackage respect stdout/stderr streams

ryanking13:
### Description

This PR makes `pyodide.loadPackage` to respect the stdout/stderr streams set by the user when loading Pyodide, or using `setStdout` and `setStderr` APIs.

### Checklists

- [x] Add / update tests


src/core/print.c
```diff
@@ -0,0 +1,33 @@
+#include <emscripten.h>
+#include <stdio.h>
+
+// Note: We cannot use stdout / stderr const from stdio.h because we override
+// the stream when initializing the Pyodide runtime. Also users can override the
+// streams using `setStderr` and `setStdout`. Therefore, we always need to open
+// the stream on each call.
```
hoodmane, line 7: I'm confused about this -- don't we refresh `stdout` and `stderr` whenever we change the streams? They are basically just some extra state wrapping file descriptors 1 and 2 and we replace those file descriptors. If this comment is true, I would think it's a bug with our stream handling that should be fixed?

src/js/load-package.ts
```diff
@@ -128,15 +128,27 @@ export class PackageManager {
...
   constructor(api: PackageManagerAPI, pyodideModule: PackageManagerModule) {
     this.#api = api;
     this.#module = pyodideModule;
     this.#installer = new Installer(api, pyodideModule);
+
+    this.stdout = (msg: string) => {
+      const msgPtr = this.#module.stringToNewUTF8(msg);
+      this.#module._print_stdout(msgPtr);
+      this.#module._free(msgPtr);
+    };
```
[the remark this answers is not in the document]
hoodmane: Nevermind, seems like we can't use `withStackSave` without first resolving:
https://github.com/emscripten-core/emscripten/issues/24275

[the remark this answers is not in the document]
hoodmane: Though we could still use `stackSave` and `stackRestore` here which would be better than `malloc` and `free`.

src/core/print.c
ryanking13: But we are removing the `/dev/stdout` and creating it again right when we reset the stdout, isn't it?

the `stdout` macro in `stdio.h` is a [`FILE *` descriptor](https://github.com/openbsd/src/blob/91a4c8b7931217b6b2332ae9b98d4e414ca690dc/include/stdio.h#L198) of `/dev/stdout`, to if /dev/stdout changes it would break the pointer I think.

hoodmane: Have you tested it? I'd think the `stdout` `File *` contains a pointer to file descriptor 1. As long as we update file descriptor 1 to the right thing, it should work. `dup2(fd, 1)` can be used to redirect stdout and writes to stdout will follow.

If you can show that this is needed that would be a bug we should fix in our stream handling code rather than work around.

ryanking13: Last time when I checked, it failed, but let me check it again.

ryanking13: Using `stdout` and `stderr` works. Maybe I made a mistake when I tested before. Thank you for asking again!

conftest.py
```diff
@@ -197,6 +199,24 @@ def pytest_collection_modifyitems(config, items):
 
         maybe_skip_test(item, delayed=True)
 
+    # After https://github.com/pyodide/pyodide/pull/5621, all tests using loadPackage
+    # requires dynamic linking, and run_in_pyodide decorator always calls loadPackage
```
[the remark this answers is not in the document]
ryanking13: I think exporting the symbols works. Let me test that way.

src/core/print.c
```diff
@@ -0,0 +1,21 @@
...
```
hoodmane, line 7:
```suggestion
```

hoodmane:
Thanks @ryanking13!

samuelcolvin:
At the risk of sounding grumpy, this "enhancement" broke tests for pydantic-ai where we specifically don't want install messages send to stdout, and has taken me more hours than I would like to admin to track down.

@hoodmane what's the recommended way to avoid install logs going to stdout now we have this?

[Here's](https://github.com/pydantic/pydantic-ai/blob/78e006c8c7b1e087d57c50dacb00587b46fbdac5/mcp-run-python/src/runCode.ts#L34-L44) our current hacky workaround which this broke.

hoodmane:
What you have there makes sense to me.

samuelcolvin:
But it no longer works with this change - installation logs still make their way to stdout??

hoodmane:
This pr only changed the default behavior, if you pass explicit callbacks the behavior should be unchanged.

samuelcolvin:
No it definitely changes it.

samuelcolvin:
Checkout https://github.com/pydantic/pydantic-ai/pull/1944, then run:

```
# tests with 0.27.6 will pass
uv run --package mcp-run-python pytest mcp-run-python -k 'test_install_run_python_code'

# install 0.27.7
cd mcp-run-python
deno add npm:pyodide@0.27.7
cd ..

# test with 0.27.7 will fail because the "output" section (from stdout/sderr) will include all installation logs
uv run --package mcp-run-python pytest mcp-run-python -k 'test_install_run_python_code'
```

hoodmane:
The test just hangs for me either way.

samuelcolvin:
Maybe you don't have deno installed?

Either way, to demonstrate a problem here's a PR (https://github.com/pydantic/pydantic-ai/pull/1950) that just uprev's to `0.27.7`, and [here's](https://github.com/pydantic/pydantic-ai/actions/runs/15539912745/job/43748009616?pr=1950) the failures in CI because stdout has changed.

ryanking13:
> [Here's](https://github.com/pydantic/pydantic-ai/blob/78e006c8c7b1e087d57c50dacb00587b46fbdac5/mcp-run-python/src/runCode.ts#L34-L44) our current hacky workaround which this broke.

Hmm, this is weird. Let me take a look.

ryanking13:
I was able to reproduce the issue with this. Let me make a fix.

```js
// deno run --allow-read --allow-net --allow-write test.mjs
import { loadPyodide } from 'npm:pyodide';

async function main() {
    await pyodide.loadPackage("numpy", { messageCallback: (msg) => console.error('messageCallback:', msg) });
}

await main();
```

PR #5624, opened 2025-05-08
Make jsbind destinguish properly between class instances and classes

hoodmane:
I changed the signature of `bind_sig` to be `type[T] -> T` so that `proxy.bind_sig(SomeClass)` returns an object typed like an instance of `SomeClass`. I added `proxy.bind_class(SomeClass)` to return an object typed as the class `SomeClass`. At runtime, `bind_class` just calls `bind_sig(type(SomeClass))`, but in a way that mypy understands.

I also fixed the handling for non-static methods. They should have one fewer argument. We fix this by calling `__get__`.

- [x] Add / update tests

### DIFF
--- a/src/core/jsproxy.c
+++ b/src/core/jsproxy.c
@@ -332,6 +332,30 @@ static PyMethodDef JsProxy_bind_sig_MethodDef = {
   METH_O,
 };
 
+PyObject*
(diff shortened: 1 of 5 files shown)


ryanking13 (approved):
Thanks! There is a test failure, but codewise looks good to me.

PR #5626, opened 2025-05-10
Constraint cython versions for a few packages

ryanking13:
It looks like the latest cython release (3.1.0) is not compatible with some packages.


ryanking13:
Opened a separate PR #5628 to handle GHA failure.

I'll go ahead and merge to unblock other PRs

PR #5635, opened 2025-05-14
add newer libtool to docker image

pepijndevos:
### Description

This adds a newer libtool version that has support for building emscripten shared libraries which is needed for #5601 

### Checklists

- [ ] Add / update tests

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -7,9 +7,9 @@ RUN apt-get update \
         bzip2 ccache f2c g++ gfortran git make \
         patch pkg-config swig unzip wget xz-utils \
         autoconf autotools-dev automake texinfo dejagnu \
-        build-essential libtool libltdl-dev \
+        build-essential libltdl-dev \
         gnupg2 libdbus-glib-1-2 sudo sqlite3 \
-        ninja-build jq \
+        ninja-build jq cmake bison \
   && rm -rf /var/lib/apt/lists/*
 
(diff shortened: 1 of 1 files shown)


pepijndevos:
Does this require any of the checklist items? Maybe changelog but it's not something any user should care about.

agriyakhetarpal (approved):
Thanks, @pepijndevos! I guess leaving out the CHANGELOG would be fine. I'll leave it to @hoodmane to push the Docker images as I think it has been easy to mess up in the past :)

Dockerfile
```diff
@@ -21,6 +21,14 @@ RUN wget https://mirrors.ocf.berkeley.edu/gnu/autoconf/autoconf-2.71.tar.xz \
     && cp /usr/local/bin/autoconf /usr/bin/autoconf \
     && rm -rf autoconf-2.71
 
+# install libtool 2.5.4, required for emscripten support
```
ryanking13, line 24: Could you please update the comment mentioning that it is required for `ngspice` package specifically, so that we can remove it when we are not building ngspice in the repository? (#4918).

pepijndevos: Done

hoodmane:
I'm sorry for the delay @pepijndevos I'll try to deal with this tomorrow.

pepijndevos:
@hoodmane any news?

PR #5646, opened 2025-05-17
Update patch for Emscripten identifiers check to reflect upstream PR

hoodmane:
Upstream PR:
https://github.com/emscripten-core/emscripten/pull/24359


ryanking13 (approved):
Thanks!

jtbandes:
@hoodmane hi there, I see that this upstream PR has now been superseded by https://github.com/emscripten-core/emscripten/pull/24427 which is released in emscripten 4.0.12 and newer. I am trying to build a python+rust package for pyodide, but for compatibility with 0.28.2 I am forced to build with emscripten 4.0.9, and the build fails due to this same emcc error that has now been fixed upstream. My question is — is there any expected timeframe when pyodide will be updated to emscripten 4.0.12 or newer, so that building rust packages can work without me needing a custom patched build of emscripten 4.0.9? Anything I could do to help out? Thanks! 🙏

hoodmane:
Updating the emscripten version is an abi break so we do it at the same time as updating Python. So for Python 3.13 we're sticking with Emscripten 4.0.9. We can replace the patch with the one that eventually landed though.

jtbandes:
Thanks for the info. Unfortunately the only way I’ve been able to get the rust package to build has been to go back to the emscripten version before the bug was introduced, namely 3.x.

hoodmane:
You can't get it to work by applying the patch to emscripten 4.0.9? We are building rust packages successfully in our ci.

jtbandes:
Indeed I probably could :)  I just think it’s a lot to ask the average pyodide+rust user to also build a custom emcc in order to compile their package, so looking forward to the next release with 4.0.12+

hoodmane:
You don't actually have to build emcc, just install it and apply the patch to the installed copy and it will work immediately. I think cibuildwheel automates the entire process. @ryanking13 I guess ideally pyodide-build should get a subcommand to install emscripten.

PR #5649, opened 2025-05-18
Minor doc fixes

oscargus:
### Description

Main purpose was to add the missing `-` in the linker option as that will maybe trip someone up later.

But tried to find some more things so that it didn't become a single character PR...

### Checklist

- [N/A] Add a [CHANGELOG](https://github.com/pyodide/pyodide/blob/main/docs/project/changelog.md) entry
- [N/A] Add / update tests
- [N/A] Add new / update outdated documentation


hoodmane (approved):
Thanks!

hoodmane:
Backported to stable so the docs should be updated to reflect this change soon.

PR #5652, opened 2025-05-19
Use lockfileURL instead of IndexURL when resolving package URL

ryanking13:
### Description

This pull request changes how remote package URLs are resolved when a custom lockfile is used. Previously, we used indexURL as a base URL for resolving packages, while this PR changes it to use lockFileURL as a base URL. This PR is required from a package unvendoring perspective.

Consider the following scenario:

1. We release Pyodide 0.28 with a certain set of packages.
2. After 0.28 release, we add a few more packages to our distribution, and release it in `pyodide-recipes` repository.
3. There is a user who wants to use the newly added packages in Pyodide 0.28; he/she download the packages and the lockfile from the `pyodide-recipes` repository, host it somewhere, and use the `lockFileURL` parameter to use that package.

Then, he/she will load Pyodide such as:

```ts
await loadPyodide({
  indexURL: "https://cdn.jsdelivr.net/pyodide/v0.28.0/full",
  lockFileURL: "https://my.custom.website/pyodide-lock.json",
})
```

In this case, the new packages will be hosted in `https://my.custom.web.site`, not in our CDN, so it would be more natural to use the lockFileURL, as a base URL for the package installation.

### Checklist

- [x] Add / update tests

### DIFF
--- a/src/js/load-package.ts
+++ b/src/js/load-package.ts
@@ -146,14 +146,14 @@ export class PackageManager {
       this.#api.config.lockFileURL.substring(
         0,
         this.#api.config.lockFileURL.lastIndexOf("/") + 1,
-      ) || location.toString();
+      ) || globalThis.location?.toString();
 
     if (IN_NODE) {
       this.installBaseUrl = this.#api.config.packageCacheDir ?? lockfileBase;
     } else {
       this.installBaseUrl = lockfileBase;
     }
-    
+
     this.stdout = (msg: string) => {
       const sp = this.#module.stackSave();
       try {
--- a/src/js/load-package.ts
+++ b/src/js/load-package.ts
@@ -140,17 +140,18 @@ export class PackageManager {
     this.#module = pyodideModule;
     this.#installer = new Installer(api, pyodideModule);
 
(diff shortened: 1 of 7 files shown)


ryanking13:
> Perhaps micropip.freeze could be changed to always use absolute URLs?

That's a good point. Thanks. Yes I think micropip should emit absolute URLs.

src/js/load-package.ts
```diff
@@ -382,11 +398,11 @@ export class PackageManager {
 
   /**
    * Download a package. If `channel` is `DEFAULT_CHANNEL`, look up the wheel URL
-   * relative to packageCacheDir (when IN_NODE), or indexURL from `pyodide-lock.json`, otherwise use the URL specified by
+   * relative to packageCacheDir (when IN_NODE), or to lcokfileURL, otherwise use the URL specified by
```
hoodmane, line 401:
```suggestion
   * relative to packageCacheDir (when IN_NODE), or to lockfileURL, otherwise use the URL specified by
```

hoodmane (approved):
Makes sense to me, thanks!

ryanking13:
Thanks for the review!

PR #5653, opened 2025-05-19
Emscripten 4.0.9

hoodmane:


agriyakhetarpal:
I realised I didn't fix the `package_index` test in #5642, I'll do so in another PR right now.

agriyakhetarpal:
I think the failing package index test will be fixed after pyodide/pyodide-build#206 is reviewed/merged and we have a release with it (we can update the submodule as well). We'll be moving the tests to pyodide-build anyway with pyodide/pyodide-build#126, so it's fine to ignore it, too.

PR #5655, opened 2025-05-21
Implement ability to specify a custom fetcher to `pyfetch`.

dom96:
### Description

It's sometimes useful to be able to specify a custom fetcher. This PR enables pyfetch to receive a custom fetcher and use it.

### Checklist

- [x] Add / update tests

### DIFF
--- a/src/py/pyodide/http.py
+++ b/src/py/pyodide/http.py
@@ -396,7 +396,9 @@ class FetchResponse:
         self.abort_controller.abort(_construct_abort_reason(reason))
 
 
-async def pyfetch(url: str, /, *, signal: Any = None, fetcher: Any = None, **kwargs: Any) -> FetchResponse:
+async def pyfetch(
+    url: str, /, *, signal: Any = None, fetcher: Any = None, **kwargs: Any
+) -> FetchResponse:
     r"""Fetch the url and return the response.
 
     This functions provides a similar API to :js:func:`fetch` however it is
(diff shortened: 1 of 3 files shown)


dom96:
Yeah, so in Cloudflare Workers we have things like Durable Objects which you can call `fetch` on. Here is an example: https://developers.cloudflare.com/changelog/2025-05-14-python-worker-durable-object/. The `stub.fetch` line specifically.

I want pyfetch to be used for this so need it to support calling `stub.fetch` rather than the global fetch.

ryanking13:
I see. Makes sense. I am fine with this change, then, but I'll let @hoodmane approve as he might have more ideas.

Also, could you please update the testcase and the docsting accordingly?

src/py/pyodide/http.py
```diff
@@ -433,10 +433,11 @@ async def pyfetch(url: str, **kwargs: Any) -> FetchResponse:
     else:
         signal = controller.signal
     kwargs["signal"] = signal
+    fetcher = kwargs["custom_fetcher"] if "custom_fetcher" in kwargs else _jsfetch
```
hoodmane, line 436: Maybe let's just call the argument fetcher? And the `signal` handling could use a bit of tidying too. How about:
```py
async def pyfetch(url: str, /, *, signal: Any = None, fetcher = None, **kwargs: Any) -> FetchResponse:
    controller = AbortController.new()
    if signal:
        signal = abortSignalAny(Array.new(signal, controller.signal))
    else:
        signal = controller.signal
    kwargs["signal"] = signal
    fetcher = fetcher or _jsfetch
```
The docstring should also be updated to add this new parameter.

hoodmane:
@dom96 okay now you need to merge main and add `"Response"` here:
https://github.com/pyodide/pyodide/blob/main/conftest.py?plain=1#L94

hoodmane (approved):

Issue #5584, opened 2025-04-08
SDL2: SDL_CreateWindow and EGL calls fail in a side module


pthom:
## 🐛 Bug

When using SDL2, and trying to call `SDL_CreateWindow` with OpenGL (flag=`SDL_WINDOW_OPENGL`) I get an error that says "indirect call to null"

**Log extract:**

* Using safari
<img width="556" alt="Image" src="https://github.com/user-attachments/assets/13f48fa0-9710-48c5-9b04-217086bc6bd8" />

* Using firefox:
```
About to call SDL_CreateWindow [pyodide.asm.js:9:655386](http://localhost:8000/pyodide_dist/pyodide.asm.js)
Pyodide has suffered a fatal error. Please report this to the Pyodide maintainers. [pyodide.asm.js:9:604594](http://localhost:8000/pyodide_dist/pyodide.asm.js)
The cause of the fatal error was: [pyodide.asm.js:9:604694](http://localhost:8000/pyodide_dist/pyodide.asm.js)
RuntimeError: indirect call to null
...
```

### Environment

- Pyodide Version: using the main branch, commit 1539f304
- Browser version: this does not depend on the browser (will fail on firefox and safari)
- Build environment: Ubuntu 22.04, using conda

### Minimal reproducible example

Extract the example package I provide in the tgz file below (expand it into pyodide/packages)

[daft_lib.tgz](https://github.com/user-attachments/files/19648980/daft_lib.tgz)

**Content**
```
daft_lib/
├── meta.yaml           # contains a build step to trigger the build of fPIC libs (for SDL2 and html5)
├── daft_lib_src/
│   ├── CMakeLists.txt  # will manually link to ${EMSCRIPTEN_SYSROOT}/lib/wasm32-emscripten/pic/libSDL2.a
│   ├                   # (which is compiled thanks to the build step in meta.yaml)
│   ├
│   ├── daft_lib.cpp    # A minimal example (publishes dummy_sdl_call in python,
│   ├                   # which calls Sdl_CreateWindow)
│   ├                   # Can be compiled as a standalone emscripten app.
│   ├
│   ├── pyproject.toml  # uses nanobind
│   ├
│   ├── daft_lib.html   # Compilation of daft_lib.cpp as a standalone emscripten app
│   ├── daft_lib.js     # (here, the call to Sdl_CreateWindow does not fail)
│   ├── daft_lib.wasm*
[...]

PR #5656, opened 2025-05-22
Fix regression in symbol resolution and enable pyxel

ryanking13:
### Description

Related: #5584 (I'll close this issue manually after verifying that it works after the merge)

This fixes a symbol resolution issue that started to occur after we updated Emscripten version from 4.0.3 to 4.0.6.

#### What happened?

In `pre.js`, we set `Module.reportUndefinedSymbols` to no-op, and this was for no-dylink build.

Before Emscripten 4.0.6, the pre.js was applied before the Emscripten updates the attributes in the module. Therefore, in regular (dylink enabled) build, the `reportUndefinedSymbols` function was replaced with a normal one that Emscripten provides.

But this behavior was silently changed in Emscripten 4.0.6 (or 4.0.5 or 4.0.4, I am not sure), and the no-op implementation provided by `pre.js` is not overwritten by the Emscripten implementation.

So what happens is in the stable branch (0.27.6), the reportUndefinedSymbols is correctly set, 

```js
console.log(pyodide._module.reportUndefinedSymbols)

()=>{for(var[symName,entry]of Object.entries(GOT)){if(entry.value==0){var value=resolveGlobalSymbol(symName,true).sym;if(!value&&!entry.required){continue}if(typeof value=="function"){entry.value=add…
```

but in the master branch, it is set to no-op

```js
console.log(pyodide._module.reportUndefinedSymbols)

()=>{}
```

This caused an issue in the symbol resolution, resulting in a bug in resolving JS symbols from the side module (SDL).

### DIFF
--- a/src/core/pre.js
+++ b/src/core/pre.js
@@ -62,7 +62,7 @@ const errNoRet = () => {
 };
 
 // For no-dylink build, this is a no-op.
-Module.reportUndefinedSymbols ||= () => {};
+Module.reportUndefinedSymbols ??= () => {};
 
 const nullToUndefined = (x) => (x === null ? undefined : x);
 
--- a/src/core/pre.js
+++ b/src/core/pre.js
@@ -61,7 +61,8 @@ const errNoRet = () => {
   );
 };
 
-Module.reportUndefinedSymbols = () => {};
+// For no-dylink build, this is a no-op.
+Module.reportUndefinedSymbols ||= () => {};
 
 const nullToUndefined = (x) => (x === null ? undefined : x);
 
--- a/Makefile.envs
+++ b/Makefile.envs
@@ -100,6 +100,7 @@ export PYCOMPILE_EXCLUDE_FILES=\
 export DBGFLAGS_NODEBUG=-g0
 export DBGFLAGS_WASMDEBUG=-g2
 export DBGFLAGS_SOURCEMAPDEBUG=-g3
+export DBG_LDFLAGS_DYLINK=-sDYLINK_DEBUG=2
 export DBG_LDFLAGS_SOURCEMAPDEBUG=-gseparate-dwarf
 
 export DBGFLAGS=$(DBGFLAGS_NODEBUG)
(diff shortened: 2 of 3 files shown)


src/core/pre.js
```diff
@@ -61,7 +61,8 @@ const errNoRet = () => {
   );
 };
 
-Module.reportUndefinedSymbols = () => {};
+// For no-dylink build, this is a no-op.
+Module.reportUndefinedSymbols ||= () => {};
```
hoodmane, line 65:
```suggestion
Module.reportUndefinedSymbols ??= () => {};
```

hoodmane (approved):
Thanks for figuring it out @ryanking13 ! I'd really like to work with Emscripten to make `reportUndefinedSymbols` either unnecessary or rename it to make it more clear when it's required to call it.

ryanking13:
> I'd really like to work with Emscripten to make reportUndefinedSymbols either unnecessary or rename it to make it more clear when it's required to call it.

Yes, me too. It is confusing that the function name is "report", but what it actually does is resolve the symbols...

hoodmane:
Perhaps we can stop using it once we bring in the rpath changes. The point here is that our loader logic calls it after loading all the packages and in the builds with dynamic linking disabled we want that to not crash so the simplest thing is to make it into a no op. 

We've been using `reportUndefinedSymbols()` directly since the beginning of the Pyodide project seven years ago and this is the first I've heard about it not being public API.

hoodmane:
As a user of Emscripten there is no easy way to tell whether an API is public or not.

sbc100:
> Perhaps we can stop using it once we bring in the rpath changes. The point here is that our loader logic calls it after loading all the packages and in the builds with dynamic linking disabled we want that to not crash so the simplest thing is to make it into a no op.
> 
> We've been using `reportUndefinedSymbols()` directly since the beginning of the Pyodide project seven years ago and this is the first I've heard about it not being public API.

Sadly we don't have any good docs on internal vs external JS APIs or good enforcement mechanisms.  This means that folks end up depending on pretty much all the internal behaviour of the emscripten JS codebase.

I would say, in an ideal world, the whole of libdylink.js should probably be considered internal, and an implementation detail on the `dlopen` C function (which is the public API).   Of course, its hard to stop folks from building on these internals in a language like JS.

sbc100:
BTW. why do you need `reportUndefinedSymbols` to be callable in the statically linked build?  Can't the calling code be made to only conditionally call this?

e.g. `if (isDyanamicallyLinked) Module.reportUndefinedSymbols()`  or wrap the call in `#if MAIN_MODULE` if the code is being pre-processed?    (BTW did you know you can use the pre-processor in pre-js files now if you add `#preprocess` to the top of them?).  Monkey patching the generated code for a static build to make it appear as a dynamic build seem a little odd.

hoodmane:
Oops GitHub showed two copies of your comment @sbc100 so I deleted one and it deleted both...

hoodmane:
I agree with the accidentally deleted comment of @sbc100 that it would be less hacky to do 
```js
if (dynamicLinkingEnabled)
   reportUndefinedSymbols();
```

sbc100:
I think this was the deleted comment? :

BTW. why do you need reportUndefinedSymbols to be callable in the statically linked build? Can't the calling code be made to only conditionally call this?

e.g. if (isDyanamicallyLinked) Module.reportUndefinedSymbols() or wrap the call in #if MAIN_MODULE if the code is being pre-processed? (BTW did you know you can use the pre-processor in pre-js files now if you add #preprocess to the top of them?). Monkey patching the generated code for a static build to make it appear as a dynamic build seem a little odd.

PR #5660, opened 2025-05-26
DOC Update path to nightly Rust toolchain sysroot

agriyakhetarpal:
### Description

When installing our custom Rust sysroot stored from https://github.com/pyodide/rust-emscripten-wasm-eh-sysroot/releases/, I noticed that this snippet in the docs does not work because `rustup which --toolchain nightly-2025-02-01 rustc` command returns `/Users/<username>/.rustup/toolchains/nightly-2025-02-01-aarch64-apple-darwin/bin/rustc`, which is not a directory but rather the `rustc` executable. We want to install into `lib/rustlib/` one level back (outside of `bin/`) instead.

### Checklist

### DIFF
--- a/docs/development/abi.md
+++ b/docs/development/abi.md
@@ -166,7 +166,7 @@ sysroot. To install the Emscripten sysroot use:
 
 ```sh
 rustup toolchain install nightly-2025-02-01
-TOOLCHAIN_ROOT=$(rustup which --toolchain nightly-2025-02-01 rustc)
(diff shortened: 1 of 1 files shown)


ryanking13:
Actually, GitHub is having issues

image

agriyakhetarpal:
Argh, yes, I'm facing this on other PRs as well; it's sometimes a little funny when GitHub starts to rate limit itself :P

PR #5667, opened 2025-05-27
Better Python -> JS interoperability

WebReflection:
This MR goal is to show how to improve Python -> JS interoperability by doing the following things:

  * `has` and `get` traps for Python dictionaries translates into `key in dict` and `dict[key]` in *JS*, allowing the usage of PyProxy without needing to convert these for the entirety of the Web APIs that expects object literals and not "*maps*" out there: `document.addEventListener(type, fn, { "once": True })` as example would just work withou *ffi* or *to_js* and, most importantly, without surprises when native JS APIs are used
* user defined callbacks pass through the *FinalizationRegistry* out of the box too so that `create_proxy` is not needed because when we use 3rd party APIs we never know when or even if these will call passed callbacks ... 

We have seen people using `create_proxy` all over the place but also `to_js` and then many complained something "*is slow*" but the reality is that they don't care about destroying proxies, they can't control when callbacks are de-referenced so that "*guarding*" the memory easily backfires with tons of often unnecessary or redundant operations.

**Please Note** I have no idea if this would be *the best approach* to solve this issue but I've put a 1:1 built demo on npm to test this and so far results are super smooth and code way more clean + I have a global *Proxy* patch ready to land in case this approach or the idea of simplifying interoperability will be discarded from this team but patching globals to do what these few lines do feels like not the best we both can offer to our users.

Ideally I'd like to have this feature behind a `loadPyodide` flag so that I don't need to fork this project and maintain my own build or I don't need to patch globally the *Proxy* but I am overly-open to any discussion or conclusion or way forward, this MR is merely an example of what we'd love to have in *Pyodide* to stop explaining why things are broken or not working as expected, especially when 3rd party JS libraries are used out there.

Thanks in advance for any sort of suggestion, comment, hint or eventual discussion to help this topic forward 🙏

### Description

This MR allows the following code to work out of the box:

```python
from js import document

document.addEventListener("click", lambda e: print(e.type), { "once": True })
```

No `create_proxy` needed and no `to_js` needed, the listener is invoked only **once** and the internal *FinalizationRegistry* handles the callback automatically.

### Checklist

- [ ] Add / update tests

### DIFF
--- a/src/core/pyproxy.ts
+++ b/src/core/pyproxy.ts
@@ -2218,7 +2218,7 @@ const PyProxyHandlers = {
     // 2. The result of getattr
     let result = python_getattr(jsobj, jskey);
 
-    // allow `dict[key]` to work as falback
+    // allow `dict[key]` to work as fallback
     return result === undefined && jsobj instanceof PyDict && jsobj.has(jskey)
       ? jsobj.get(jskey)
       : result;


src/core/pyproxy.ts
```diff
@@ -297,7 +297,10 @@ function pyproxy_new(
   }
   const attrs = { shared, props };
   target[pyproxyAttrsSymbol] = attrs;
-  return proxy;
+  // implicit `create_proxy` non registered functions
```
juntyr, line 300: now?

ryanking13:
Thanks @WebReflection. I guess Hood would have more thoughts, but in general, I really like the idea to make

```
from js import document

document.addEventListener("click", lambda e: print(e.type), { "once": True })
```

just work.

I am +1 for adding `has` and `get` traps which will make Python dict work like a JS object, without `to_js`, but I am not very sure about calling `create_proxy` implicitly. As most people will not be aware that they are making dangling proxies everywhere. So if there is a better way to handle the lifetime of a proxy, it would be the best.

WebReflection:
> if there is a better way to handle the lifetime of a proxy, it would be the best.

there's not, because users don't know if they need a proxy to start with when they pass Python lambdas or functions to 3rd party JS API/libraries so the current state is that they create explicitly out of paranoia proxies all over.

Moreover, the *FianlizationRegistry* **is** the lifetime handler that works reliably already and if the argument is that one does not get to know when it kicks in, it's also the reason we all use Garbage Collectors behind our scripting scene (JS or Python).

At least our users never expect to handle manually lifetime of anything + they don't hold proxies anywhere and I need yet to see a single `proxy.destroy()` that is not handled internally by Pyodide, which is why I would like to propose a `loadPyodide` flag called `auto_proxy` so that whoever needs to handle proxies lifetime can use `ffi.create_proxy` and backward compatibility/memory "*safety*" is preserved but everyone else that doesn't care about it, doesn't handle it or doesn't understand it or cannot, practically, know when the lifetime of a proxy is good to be called/destroyed, can **opt-in** via such flag and have the suggested logic available to forget about it.

This is like revocable proxies in the JS world: "**nobody uses those**" because when you land a proxy somewhere you don't get to know or be notified when that proxy is not needed anymore or referenced, also because in telling you that proxy is not referenced anymore a reference to that proxy is held inevitably so that the *FinalizationRegistry* can't even work ... reaason *WeakRef* exist in the JS: if you hold a proxy to invoke `destroy()` after you are causing yourself issues interfering with the *FinalizationRegistry* ability to get rid of that proxy ... that ship has sealed long time ago though, the "*hold a proxy to desotry it*" concept, so here with that change the GC on the JS side will be a better job than manual, leaking-prone, user-land attempts to clear proxies once created due necessity.

> I am +1 for adding has and get traps which will make Python dict work like a JS object

great, that's also a no-brainer to me and both things work already exactly like that in MicroPython where writing code feels way more natural/Pythonic than Pyodide because requiring an FFI also makes code less portable between client/server/projects but again, if that's not desired, can we please have a flag there too so that `auto_literal` (as example) would make PyDict act like literals at least while reading properties, which is the mosto common/desired approach for anything that accepts an object literal to bootstrap any API configuration?

hoodmane:
The change to the get set and has handlers for dictionary proxies should be okay though the dictionary items need to come after the attributes on the dictionary object.

I don't think the proxy change makes sense because of the same two garbage collectors problem we've discussed before: if there's a Python to JavaScript to Python reference loop it gets leaked. Currently we deal with this by breaking references from JavaScript to Python manually and allow the Python garbage collector to automatically handle Python to JavaScript references. If we get @Thrameos 's two GCs proposal implemented we could make a change like this.

WebReflection:
@hoodmane thanks for the details, I'd be more than happy to split this topic in two parts:

  1. improve PyDict *readability* as it's done now, moving the check after
  2. remove the automatic create proxy

About point 1 though, I don't think it's something to worry about but `$` as prefix is a valid name in JS and maybe some library uses that as field.

The "*funny*" thing is that I personally use `{$: public, _:private}` pattern pretty frequently but it's true usually those are just wrappers passed around, not something expected at any API signature level, although that would result into `{['']: public, _:private}` if I move that check after ... are we saying that's OK? I think it cannot be "*worse than current state*" in general so it's OK, I just wanted to underline why I placed that check where it is now.

About point 2: the current state is that users just `create_proxy` ignoring the issue you are talking about and if they don't control/reference already created proxies they do that multiple times increasing the chances of the double reference problem as opposite of writing better code ... I think a boolean flag named `dangerously_create_proxy_with_possible_leaks` would both scare users and yet serve our purpose in helping them writing code easier to reason about so ... should I file a new MR proposing such flag as that double GC problem might take years to land in a production-ready state?

'cause whatever we do on our side, which is either intercepting all JS callbacks signatures with nested loops for arrays or dictionaries or patching globally the Proxy feels like moving mountains instead of poking a single leaf (which is that single line of code that "*solves it all*" for us and our users).

Thanks

WebReflection:
@hoodmane sorry I need more coffee but still about that double leak/problem ... if we use *WeakRef* anything landing on the JS side as proxy could fail if it's been freed on the Python side. As example of a *WeakProxy* here:

```js
class WeakProxy {
  #handler;
  constructor(target, handler) {
    this.#handler = handler;
    return new Proxy(new WeakRef(target), this);
  }
  // all Proxy traps or just those desired
  get(wr, ...rest) {
    const target = wr.deref();
    // check if target was collected at the Python side of affairs
    if (no_more_referenced(target))
      throw new Error('this proxy was collected and it is not usable anymore');
    return this.#handler.get(target, ...rest);
  }
}
```

Above idea is to never hold the Python reference, a *FinalizationRegistry* could be easily added to the returned Proxy or to target the `target` itself (to unregister it when convenient), double references should never leak, in the worst case scenario what leaks is a basic *WeakReference* which is a tiny wrapper (it's cheap and light) that might fail if ever used/invoked/accessed in the future but it can be collected at any time.

Add a *WeakMap* guard to return always the same proxy per `target` if already known and you have a 1:1 Python <-> JS proxy handler that is never leaked and it can be *revoked* (that's also another option, using `Proxy.reocable(target, handler)` instead) if the underlying Python reference is gone ... all these primitives should help avoiding that double GC issue but of course require 1. will to implement these ... and 2. agreement if that should be still behind a flag (which is OK at the beginning, it can be experimental or dangerous or whatever until we're confident everything is indeed better than before).

Happy to provide some concrete example that would survive the double GC gotcha in the wild.

ryanking13:
> thanks for the details, I'd be more than happy to split this topic in two parts:
> 1. improve PyDict readability as it's done now, moving the check after
> 2. remove the automatic create proxy

Yes, let's separate the tasks. For 1, I think we all agree with the direction. Some minor details can be changed, but at least I am happy with this change.

hoodmane:
@WebReflection if you don't mind I will write a similar change to this myself.

WebReflection:
@hoodmane I don't mind, your PR looks better/cleaner so, thanks, I'll close this one then.

PR #5668, opened 2025-05-28
Update docker image to 20250528-chrome136-firefox138-py313

ryanking13:
With autoconf 2.71


hoodmane:
Why do you want to downgrade the autoconf version?

PR #5669, opened 2025-05-28
Add gnucap-python recipe

pepijndevos:
### Description

I tried adding another python electronics simulator package similar to #5601 but I'm not sure I'm willing to see this one through. I'm mostly posting it in case there is a tenable solution that the maintainers have experience with.

I have managed to build the gnucap library itself with extensive hacks. This package does _not_ use autotools but has a bespoke configure script. I faced several issues with not being able to override variables from the command line that I had to patch out. I also had to manually split the build into native binaries used during compilation of the wasm library.

The more fundamental reason that makes me doubt a gnucap package is that advanced models are compiled to shared libraries. I think the logistics of that don't really work out in Pyodide since the end user doesn't really have access to a compiler. So for advance usage the user would have to install a matching pyodide build environment locally to compile models as out of tree packages or something like that. And at that point the entire point of a web based python environment is gone. Unless you can bootstrap a wasm emscripten to compile models on the fly?

The gnucap-python library does not currently compile at all. It is a SWIG package, but instead of a setup.py it uses autotools. It currently fails because it finds the host python binary and uses that to determine linking information and understandably fails to link a wasm library to native python.

I tried setting up a pyodide venv which opened up a whole new can of worms and is probably not the way to go. So you'd either do extensive patching and/or overriding of the autotools system, or make a setup.py file from scratch.

### Checklist

- [ ] Add / update tests


ryanking13:
@pepijndevos I moved the recipes to pyodide-recipes repository. So please reopen it there if you are still interested in this PR

pepijndevos:
I'm still interested but don't have a good solution to any of the problems outlined

PR #5671, opened 2025-05-29
Minor js perf improvements

WebReflection:
### Description

The goal of this MR is to improve some low hanging fruits around TS/JS perf, most notably:

  * avoid unnecessary creation of descriptors to actually find if a property is *own* + use the modern `Object.hasOwn` when available and fallback with `hasOwnProperty` instead of `getOwnPropertyDescriptor` as that creates each time a new object even with the same ref and key
  * avoid runtime creation and crawl of always the same array of strings and use a set to have predictable *O(1)* checks for something super frequent/common (critical path)
  * avoid retrieving N times char points

**edit** tried the following, it doesn't work - feel free to ignore

- - -

About latest point though, I wonder if Python would handle just as fine `ASSIGN_U16` so that the whole JS to Python string conversion could be actually simplified as much as this:

```js
function js2python_string(value) {
  let result = _PyUnicode_New(value.length, 0xffff);
  if (result === 0) {
    throw new PropagateError();
  }

  for (let ptr = _PyUnicode_Data(result), i = 0; i < value.length; i++) {
    ASSIGN_U16(ptr, 0, value.charCodeAt(i));
    ptr += 2;
  }
  return result;
}
```

If that works, that might be not the most space-efficient way but surely it'll be the easiest/most straight forward way to go?

Anyway, happy to change/improve/ignore anything you like.

### DIFF
--- a/src/core/js2python.js
+++ b/src/core/js2python.js
@@ -22,18 +22,15 @@ function js2python_string(value) {
   let ptr = _PyUnicode_Data(result);
   if (max_code_point > 0xffff) {
     for (let i = 0; i < num_code_points; i++) {
-      ASSIGN_U32(ptr, 0, code_points[i]);
-      ptr += 4;
+      ASSIGN_U32(ptr, i, code_points[i]);
     }
   } else if (max_code_point > 0xff) {
     for (let i = 0; i < num_code_points; i++) {
-      ASSIGN_U16(ptr, 0, code_points[i]);
-      ptr += 2;
+      ASSIGN_U16(ptr, i, code_points[i]);
     }
   } else {
     for (let i = 0; i < num_code_points; i++) {
-      ASSIGN_U8(ptr, 0, code_points[i]);
-      ptr += 1;
+      ASSIGN_U8(ptr, i, code_points[i]);
     }
   }
 
--- a/src/core/js2python.js
+++ b/src/core/js2python.js
@@ -6,13 +6,14 @@ function js2python_string(value) {
   // to determine if is needs to be a 1-, 2- or 4-byte string, since
   // Python handles all 3.
   let max_code_point = 0;
-  let num_code_points = 0;
+  const code_points = [];
   for (let c of value) {
-    num_code_points++;
     let code_point = c.codePointAt(0);
+    code_points.push(code_point);
     max_code_point = code_point > max_code_point ? code_point : max_code_point;
   }
 
+  let num_code_points = code_points.length;
   let result = _PyUnicode_New(num_code_points, max_code_point);
   if (result === 0) {
     throw new PropagateError();
@@ -20,18 +21,18 @@ function js2python_string(value) {
 
   let ptr = _PyUnicode_Data(result);
   if (max_code_point > 0xffff) {
-    for (let c of value) {
-      ASSIGN_U32(ptr, 0, c.codePointAt(0));
+    for (let i = 0; i < num_code_points; i++) {
+      ASSIGN_U32(ptr, 0, code_points[i]);
       ptr += 4;
     }
   } else if (max_code_point > 0xff) {
-    for (let c of value) {
-      ASSIGN_U16(ptr, 0, c.codePointAt(0));
+    for (let i = 0; i < num_code_points; i++) {
+      ASSIGN_U16(ptr, 0, code_points[i]);
       ptr += 2;
     }
   } else {
-    for (let c of value) {
-      ASSIGN_U8(ptr, 0, c.codePointAt(0));
+    for (let i = 0; i < num_code_points; i++) {
+      ASSIGN_U8(ptr, 0, code_points[i]);
       ptr += 1;
     }
   }
--- a/src/core/pre.js
+++ b/src/core/pre.js
@@ -14,6 +14,10 @@ function getTypeTag(x) {
 }
 API.getTypeTag = getTypeTag;
 
+const hasOwn =
+  Object.hasOwn ||
+  ((obj, prop) => Object.prototype.hasOwnProperty.call(obj, prop));
+
 /**
  * Safe property check
  *
@@ -26,7 +30,7 @@ API.getTypeTag = getTypeTag;
 function hasProperty(obj, prop) {
   try {
     while (obj) {
-      if (Object.getOwnPropertyDescriptor(obj, prop)) {
+      if (hasOwn(obj, prop)) {
         return true;
       }
       obj = Object.getPrototypeOf(obj);
@@ -70,7 +74,7 @@ const nullToUndefined = (x) => (x === null ? undefined : x);
 function isPromise(obj) {
   try {
     // clang-format off
-    return !!obj && typeof obj.then === "function";
+    return typeof obj?.then === "function";
     // clang-format on
   } catch (e) {
     return false;
@@ -95,7 +99,7 @@ API.typedArrayAsUint8Array = bufferAsUint8Array;
 
 Module.iterObject = function* (object) {
   for (let k in object) {
-    if (Object.prototype.hasOwnProperty.call(object, k)) {
+    if (hasOwn(object, k)) {
       yield k;
     }
   }
--- a/src/core/pre.js
+++ b/src/core/pre.js
@@ -14,15 +14,13 @@ function getTypeTag(x) {
 }
 API.getTypeTag = getTypeTag;
 
-const hasOwn =
-  Object.hasOwn ||
-  ((obj, prop) => Object.prototype.hasOwnProperty.call(obj, prop));
-
 /**
  * Safe property check
  *
  * Observe whether a property exists or not without invoking getters.
  * Should never throw as long as arguments have the correct types.
+ * This check is better than `prop in obj` because it works also with
+ * primitives where `"length" in "str"`, as example, would throw.
  *
  * obj: an object
  * prop: a string or symbol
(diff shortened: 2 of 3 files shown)


WebReflection:
@ryanking13 @hoodmane btw ... the `hasProperty` crawls the prototype too and check each one up to `null` and passing, I believe, to `Object.prototype` ... accordingly, that looks to me like an overly-engineered `prop in obj` operation, as that would basically do the same: check if a property is present down the prototypal chain ... is there any background/reason that check needs to be orchestrated as such? 'cause while `hasOwn` as helper makes sense to me, the `hasProperty` utility itself could be just removed entirely, imho, unless there is some weird case I am not considering, thanks.

ryanking13:
> About latest point though, I wonder if Python would handle just as fine ASSIGN_U16 so that the whole JS to Python string conversion could be actually simplified as much as this:

I am not 100% sure, as that code was [written 6 years ago](https://github.com/hoodmane/pyodide/commit/46f64b5c7cc6b318da9d4cb697eed5a6715b2540), but I think unicode is quite complex and that code if-else branch was required to support CJK strings and emojis. But I think you can try to see if it works. I believe we have some test cases for unicode.

WebReflection:
@ryanking13 

> I think you can try to see if it works

I tried, it fails at building ... so I guess it doesn't work for complex points ... all good though, it's already slightly simplified now.

src/core/js2python.js
```diff
@@ -6,32 +6,33 @@ function js2python_string(value) {
   // to determine if is needs to be a 1-, 2- or 4-byte string, since
   // Python handles all 3.
   let max_code_point = 0;
-  let num_code_points = 0;
+  const code_points = [];
```
hoodmane, line 9: This adds an extra allocation doesn't it? I'd think the original allocation-free version would be better.

src/core/pre.js
```diff
@@ -14,6 +14,10 @@ function getTypeTag(x) {
 }
 API.getTypeTag = getTypeTag;
 
+const hasOwn =
+  Object.hasOwn ||
```
hoodmane, line 18: I think we can assume that `Object.hasOwn()` is present.

hoodmane:
Maybe you could add a comment explaining why the `hasProperty()` code is right if it is right and you understand why it's right? This stuff can be a bit JavaScript voodoo so it'd be good to explain.

src/core/js2python.js
ryanking13: But it will be cleaned up after the function isn't it?

While we don't have a thorough benchmark, when I wrote a short benchmark script comparing the two method, it seems like caching is quite better than calling `codePointAt` multiple times:

<details><summary>Script</summary>
<p>

```js
function benchmark(label, fn, iterations = 100000) {
  const { performance } = require('perf_hooks');
  const start = performance.now();
  let result = 0;
  for (let i = 0; i < iterations; i++) {
    result += fn();
  }
  const end = performance.now();
  console.log(`${label}: ${(end - start).toFixed(2)} ms`);
  return result;
}

function createLongString(length) {
  return 'Hello 🌍🚀'.repeat(length); // Mixed characters
}

// Short and long test strings
const shortString = "Hello 🌍🚀";
const longString = createLongString(100);

// Functions

function noCache(str) {
  let sum = 0;
  for (let i = 0; i < 3; i++) {
    for (const char of str) {
      sum += char.codePointAt(0) || 0;
    }
  }
  return sum;
}

function withCache(str) {
  const cache = [];
  for (const char of str) {
    cache.push(char.codePointAt(0) || 0);
  }

  let sum = 0;
  for (let i = 0; i < 3; i++) {
    for (const val of cache) {
      sum += val;
    }
  }
  return sum;
}

// Run benchmarks

console.log("=== Short String Benchmark ===");
const resultShortNoCache = benchmark("No Cache (Short)", () => noCache(shortString));
const resultShortWithCache = benchmark("With Cache (Short)", () => withCache(shortString));
console.log("Short Match:", resultShortNoCache === resultShortWithCache);

console.log("\n=== Long String Benchmark ===");
const resultLongNoCache = benchmark("No Cache (Long)", () => noCache(longString));
const resultLongWithCache = benchmark("With Cache (Long)", () => withCache(longString));
console.log("Long Match:", resultLongNoCache === resultLongWithCache);

```

</p>
</details> 

```
=== Short String Benchmark ===
No Cache (Short): 31.93 ms
With Cache (Short): 19.77 ms
Short Match: true

=== Long String Benchmark ===
No Cache (Long): 2703.89 ms
With Cache (Long): 1379.77 ms
Long Match: true
```

WebReflection: calling prototype methods multiple times is never faster than caching and all *JS* engines are super fast at collecting one-off scoped references + the code as it is passes through *JIT* or turbofan right away in both `push` and each loop later on: it's fast path by all means, it can hardly go faster.

I could've used that int to increment the index instead of using *push* but there is a known v8 limitation around that patter: it throws with arrays bigger than 1.8GB while `push` will keep going up to the array limit which is 4GB.

As summary: calling and resolving codepoints is expensive and in there unnecessarily repeated twice per each same code while caching via a plain array is cheap and fast and it doesn't cause any memory issue because it's freed instantly after.

**However** if it's the allocation the concern that could have an outer scope, always same, pointer to an array that is then reset via `cache.length = 0` but it's a rather "*dirty hack*" that shouldn't be used these days and I am not sure that ends up jitted right away anymore so I agree with @ryanking13 the current implementation wins for perf and simplicity (at the engine optimization level).

WebReflection: btw ... we were allocating dozen object descriptors up to every prototype chain before via `Object.getOwnPropertyDescriptor` so that having just an *array* to do the right thing doesn't seem to be worse than what this code was doing before this PR in terms of allocations + it is faster by design/logic 😉

WebReflection:
@hoodmane done (removed `hasOwn` fallback + added a reason `hasProperty` is good as it is compared to `in` operator).

src/core/js2python.js
```diff
@@ -6,32 +6,33 @@ function js2python_string(value) {
...
   if (result === 0) {
     throw new PropagateError();
   }
 
   let ptr = _PyUnicode_Data(result);
   if (max_code_point > 0xffff) {
-    for (let c of value) {
-      ASSIGN_U32(ptr, 0, c.codePointAt(0));
+    for (let i = 0; i < num_code_points; i++) {
+      ASSIGN_U32(ptr, 0, code_points[i]);
```
hoodmane, line 25:
```suggestion
    for (let c of code_points) {
      ASSIGN_U32(ptr, 0, c);
```

hoodmane: Or what about:
```js
code_points.forEach((c, idx) => {
   ASSIGN_U32(ptr, idx, c);
});
```

src/core/pre.js
```diff
@@ -19,14 +19,16 @@ API.getTypeTag = getTypeTag;
  *
  * Observe whether a property exists or not without invoking getters.
  * Should never throw as long as arguments have the correct types.
+ * This check is better than `prop in obj` because it works also with
+ * primitives where `"length" in "str"`, as example, would throw.
```
hoodmane, line 23: Thanks!

src/core/pre.js
hoodmane: Yeah we generally aim for ~2 years of backwards compatibility with mainstream browsers. If there were a bunch of iOS users stranded on an ancient version or something we might make special allowance for that. But I'm pretty sure this won't be the newest feature we're requiring.

src/core/pre.js
```diff
@@ -19,14 +19,16 @@ API.getTypeTag = getTypeTag;
...
  *
  * obj: an object
  * prop: a string or symbol
  */
 function hasProperty(obj, prop) {
   try {
     while (obj) {
-      if (Object.getOwnPropertyDescriptor(obj, prop)) {
+      if (Object.hasOwn(obj, prop)) {
```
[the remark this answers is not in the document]
WebReflection: Yeah, creating a descriptor to use it just as truthy value is overkill and it allocates for no benefits whatsoever so this was more a GC/RAM/logic boost based on actual Boolean and zero waste of specs (the getOwnPropertyDescripor one vs just has own property)

src/core/js2python.js
WebReflection: you won't find a single benchmark out there where `forEach` is faster than a `for` loop but I am a bit confused ... you were worried about a single array allocation and now you want to create closures for no benefit/reason/performance gain? 🤔

for loops with a fixed length are inlined and JITted right away, closures and `forEach` are not ... before there was an iterator based loop which is also way slower than an index/length based one so to me that's a no?

WebReflection: it won't work as explained ... I eventually need to write this instead:

```js
for (let i = 0; i < num_code_points; i++) {
   ASSIGN_U32(ptr, i, code_points[i]);
}
for (let i = 0; i < num_code_points; i++) {
   ASSIGN_U32(ptr, i * 2, code_points[i]);
}
for (let i = 0; i < num_code_points; i++) {
   ASSIGN_U32(ptr, i * 4, code_points[i]);
}
```

and hope all tests pass again

hoodmane: See here:
https://github.com/pyodide/pyodide/blob/main/src/core/jsmemops.h?plain=1#L17

WebReflection: done, let's see if this works 🤞

PR #5677, opened 2025-06-02
Fix RobotRaconteur with Emscripten 4.0.9

johnwason:
### Description

Fix RobotRaconteur package build issues introduced by #5332

### DIFF
--- a/packages/RobotRaconteur/char16_fix.patch
+++ b/packages/RobotRaconteur/char16_fix.patch
@@ -5,11 +5,11 @@ index ca81358d..1f03f273 100644
 @@ -35,6 +35,8 @@
  #include <boost/range/numeric.hpp>
  #include <boost/tuple/tuple_comparison.hpp>
- 
+
 +#include <uchar.h>
 +
  using namespace boost::algorithm;
- 
+
  namespace RobotRaconteur
 @@ -2872,7 +2874,7 @@ static std::string ConstantDefinition_UnescapeString_Formatter(const boost::smat
      if (boost::starts_with(i, "\\u"))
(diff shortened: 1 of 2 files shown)


agriyakhetarpal:
Hi @johnwason, thanks a lot for taking this! Once the tests pass here, would you mind also creating a pull request with the same changes for the https://github.com/pyodide/pyodide-recipes repository? We have been tracking the packages to be fixed in https://github.com/pyodide/pyodide-recipes/issues/99, and that repository will be used more in the coming time for building and testing packages. All of the recipes will eventually move there.

johnwason:
@agriyakhetarpal no problem. Could you use a submodule during the transition so we don't have to make updates in two places?

agriyakhetarpal:
Yes, I think a submodule would make sense. I also raised the topic of the commit history being squashed/lost as a part of the packages being synced from that repository to here in https://github.com/pyodide/pyodide-recipes/issues/96. I'll add a comment there about this.

johnwason:
There are some commands like `git subtree split` that allow you to split a repository keeping the commit history. I don't remember exactly what command was used but we did something similar for another project.

agriyakhetarpal:
🤔 Thanks. Would it also help with the issue of the missing history between the syncs we've done so far? I imagine if we had a pair of forceps of sorts that could pick and match commits that have already been pushed months before, we could fix it. We kept the history for https://github.com/pyodide/pyodide-build, which likely used the command you suggested, but doing this was missed for the recipes repository.

johnwason:
Sorry the split was done by someone else so I don't know the details of the commands.

agriyakhetarpal:
No worries, thanks again for this PR!

PR #5681, opened 2025-06-03
Stack switching: Turn enableRunUntilComplete on by default

hoodmane:

### DIFF
--- a/docs/project/changelog.md
+++ b/docs/project/changelog.md
@@ -47,6 +47,12 @@ myst:
   streams that are passed to `loadPyodide()` or by `pyodide.setStdout()` and `pyodide.setStderr()`.
   {pr}`5621`
 
+- {{ Enhancement }} The `enableRunUntilComplete` option to `loadPyodide()` is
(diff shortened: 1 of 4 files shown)


docs/project/changelog.md
```diff
@@ -47,6 +47,12 @@ myst:
   streams that are passed to `loadPyodide()` or by `pyodide.setStdout()` and `pyodide.setStderr()`.
   {pr}`5621`
 
+- {{ Enhancement }} The `enableRunUntilComplete` option to `loadPyodide()` is
+  now on by default. This makes `run_until_complete` block using stack
+  switching, or crash if stack switching is disabled. If you need the old no-op
```
ryanking13, line 52: If it makes `run_until_complete` crash, then I think it should be tagged as a breaking change.

hoodmane: Fair point, I will update it.

hoodmane: I suspect that any code that was doing this was pretty broken to begin with though.

PR #5686, opened 2025-06-04
Use stack switching for time.sleep implementation if possible

hoodmane:


ryanking13:
The current `time.sleep()` implementation uses busy wait. What's the benefit if we replace it with stack switching?

hoodmane:
It allows other async work to be done during the sleep rather than blocking.

ryanking13:
Makes sense. Thanks!

ryanking13 (approved):

PR #5687, opened 2025-06-04
Add to_weakref method to JsProxy

hoodmane:
and use it to improve wrappers.py

- [ ] Add / update tests

### DIFF
--- a/src/py/pyodide/ffi/wrappers.py
+++ b/src/py/pyodide/ffi/wrappers.py
@@ -1,7 +1,16 @@
 from collections.abc import Callable
 from typing import Any, Protocol, cast
 
-from . import IN_BROWSER, JsDomElement, JsProxy, create_once_callable, create_proxy
+from _pyodide._core_docs import JsWeakRef
+
+from . import (
+    IN_BROWSER,
+    JsCallableDoubleProxy,
+    JsDomElement,
+    JsProxy,
+    create_once_callable,
+    create_proxy,
+)
 
 if IN_BROWSER:
     from js import clearInterval, clearTimeout, setInterval, setTimeout
@@ -27,7 +36,10 @@ class DUMMY_DESTROYABLE:
         pass
 
 
-EVENT_LISTENERS: dict[tuple[int, str, Callable[[Any], None]], Destroyable] = {}
+EVENT_LISTENERS: dict[
+    tuple[int, str, Callable[[Any], None]],
+    JsWeakRef[JsCallableDoubleProxy[[Any], None]],
+] = {}
 
 
 def add_event_listener(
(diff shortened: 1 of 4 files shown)


ryanking13 (approved):
Thanks!

src/py/pyodide/ffi/wrappers.py
```diff
@@ -38,7 +50,8 @@ def add_event_listener(
     JsProxy corresponding to the ``listener`` parameter.
     """
     proxy = create_proxy(listener)
-    EVENT_LISTENERS[(elt.js_id, event, listener)] = proxy
+    # Use weakref so the proxy can be freed if DOM element is removed.
```
WebReflection, line 52: FYI listeners attached to elements don't leak and are free to go once the element is GCd

hoodmane: Prior to this change, ones added via this `add_event_listener()` method would leak because they would stay in `EVENT_LISTENERS` map forever.

WebReflection: I see ... then this makes sense ... **but** ... if only there was a *FinalizationRegistry* at the *JS* side of affairs that would help avoiding entirely these issues with proxies ;-)

I hope one day proxies passed to JS will be observed at the GC level by default, can't wait for that WASM GC hook that helps with double links to land and be used in here too.

hoodmane: There is a finalization registry but if we hold a strong reference to a JavaScript object it will never be GCd and the finalization registry callback will never be called. What's needed to allow the finalization registry to work is that we hold a weak ref here.

WebReflection: yeah, [I suggested that](https://github.com/pyodide/pyodide/pull/5667#issuecomment-2921377125) as a way to go with proxies indeed so you never hold strongly their reference and you can free pointers on the other side, my comment was more about having this dance more spread by default but you mentioned that's not possible due double-linked edge cases: all good here, sorry for the noise.

WebReflection: FYI I've elaborated further my idea of a *WeakProxy* in here: https://github.com/WebReflection/weak-proxy

feel free to grab that code and change it if it helps making everything less `create_proxy` / `create_proxy_once` dependent 👋

WebReflection: if interested, I've added 100% code-coverage and whole examples on how to properly proxy list/tuple in a way that these survive `Array.isArray` and `proxy instanceof Array` checks and act exactly like arrays in the JS world, so that all checks around `/^[0-9]+$/` could actually go away for every single proxied reference that is not list or tuple.

There are 3 main things to implement correctly but it's possible and it works, as code coverage verifies in there.

`v0.1.0` it is then, I hope that helps or it helps changing that bit of logic in here (or in the future).

PR #5689, opened 2025-06-05
Add tags to the core packages

ryanking13:
Splitted out from #4987

This PR adds some informational tags to the core packages, which will remain in this repository after unvendoring recipes.

- core: core packages
  - pytest: packages that are used in pytest. We need these packages to make @run_in_pyodide work.
  - cross-build: packages that are included in xbuildenv
  - pyodide.test: packages that are used in our core test suite


hoodmane (approved):
Thanks!

PR #5692, opened 2025-06-10
Fix messageCallback and errorCallback not applied to some messages in loadPackage

ryanking13:
### Description

This fixes a bug that some of the messages printed by `pyodide.loadPackage` didn't respect messageCallback and errorCallback parameter after #5621.

It was reported by @samuelcolvin.

__Why?__

A Claude-generated answer... which is quite correct I think.

> The problem is that this.loadPackageInner is just passing a function reference, but it's not bound to this. When you call wrappedLoadPackage.call(this, names, options), you're correctly setting this for loadPackageInner, but when loadPackageInner calls other methods like downloadPackage, it's not preserving the wrapped context.
You need to bind this.loadPackageInner to this before passing it to setCallbacks

So the `logStdout` and `logStderr` calls in child functions were not using the wrapped context, becuase if was using different 'this'.

### Checklist

- [x] Add / update tests


ryanking13:
Thank you for the review!

samuelcolvin:
Nice! Thanks @ryanking13 for working on this.

PR #5695, opened 2025-06-11
Fix iPad detection in wasm-gc

### LINKED ISSUE #5670: iPad on OS 18 harakiri in 0.27.1 (not 0.27.0)
## 🐛 Bug

Just `await loadPyodide` on any modern iPad OS (iOS + iPad) combination and the page will be unreachable/unusable ... on Browserstack that's iPad Air 13 2025 with OS 18 and nothing works out there **since** version 0.27.1, it's OK in version 0.27.0.

This issue has been reported by a *PyScript* user but after investigation it turned out just explicitly using interpreter with version 0.27.0 nothing breaks anymore while every version from 0.27.1 to 0.27.6 doesn't even give us a chance to debug because `loadPyodide` throws and fails in a very peculiar way there (it seems to be FS related and I am not sure this is the root cause https://github.com/pyodide/pyodide/issues/4057)

### To Reproduce

Try pyodide on iPad with OS 18 and Safari ... **Note** on iPhone everything is fine but also on iPad with OS version 16 or 17 it's fine too !!!

### Expected behavior

The page is rendered/usable/not entirely broken

### Environment

- Pyodide Version: 0.27.1+
- Browser version: Safari on OS 18 within an iPad (iPad OS)
- Any other relevant information: older version works, iPhone latest works too

ryanking13:
### Description

Hopefully resolve #5670

- Upstream PR: https://github.com/python/cpython/pull/135388
### Checklist

### DIFF
--- a/cpython/patches/0009-Fix-iPad-detection-in-wasm-gc.patch
+++ b/cpython/patches/0009-Fix-iPad-detection-in-wasm-gc.patch
@@ -1,0 +1,33 @@
+From 738e25376b23611b6077c71f5ef55eefa66dd217 Mon Sep 17 00:00:00 2001
+From: ryanking13 <def6488@gmail.com>
+Date: Wed, 11 Jun 2025 08:54:37 +0000
+Subject: [PATCH 9/9] Fix iPad detection in wasm-gc
+
+Fixes iPad detection logic in wasm-gc, which was malfunctioning in recent iPadOS + safari.
+
+---
+ Python/emscripten_trampoline.c | 8 +++++++-
+ 1 file changed, 7 insertions(+), 1 deletion(-)
+
+diff --git a/Python/emscripten_trampoline.c b/Python/emscripten_trampoline.c
+index cc5047d6bd..4a3926caeb 100644
+--- a/Python/emscripten_trampoline.c
++++ b/Python/emscripten_trampoline.c
+@@ -71,7 +71,13 @@ EM_JS(CountArgsFunc, _PyEM_GetCountArgsPtr, (), {
+ // )
+ 
+ function getPyEMCountArgsPtr() {
+-    let isIOS = globalThis.navigator && /iPad|iPhone|iPod/.test(navigator.platform);
++    let isIOS = globalThis.navigator && (
++        /iPad|iPhone|iPod/.test(navigator.userAgent) ||
++        // Starting with iPadOS 13, iPads might send a platform string that looks like a desktop Mac.
++        // To differentiate, we check if the platform is 'MacIntel' (common for Macs and newer iPads)
++        // AND if the device has multi-touch capabilities (navigator.maxTouchPoints > 1)
++        (navigator.platform === 'MacIntel' && typeof navigator.maxTouchPoints !== 'undefined' && navigator.maxTouchPoints > 1)
++    )
+     if (isIOS) {
+         return 0;
+     }
+-- 
+2.49.0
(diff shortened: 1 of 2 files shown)


ryanking13:
@WebReflection 

Could you check if this patch works? I deployed it in the GH pages.

https://ryanking13.github.io/temp-pyodide-nightly-runtime/console.html

cpython/patches/0009-Fix-iPad-detection-in-wasm-gc.patch
```diff
@@ -0,0 +1,34 @@
...
+ // )
+ 
+ function getPyEMCountArgsPtr() {
+-    let isIOS = globalThis.navigator && /iPad|iPhone|iPod/.test(navigator.platform);
++    // Starting with iPadOS 13, iPads might send a platform string that looks like a desktop Mac.
++    // To differentiate, we check if the platform is 'MacIntel' (common for Macs and newer iPads)
++    // AND if the device has multi-touch capabilities (navigator.maxTouchPoints > 1)
++    let isIOS = globalThis.navigator && (
++        /iPad|iPhone|iPod/.test(navigator.userAgent) ||
++        (navigator.platform === 'MacIntel' && typeof navigator.maxTouchPoints !== 'undefined' && navigator.maxTouchPoints > 1)
```
WebReflection, line 26: FYI `undefined > 1` is always `false` so that a check for a *type* as *undefined* is bloat *but*, most importantly, a check that access the property *might* have side-effects ... it's a long story but if you really want to check the existence, you are better off with `'maxTouchPoints' in navigator` instead of `typeof navigator.maxTouchPoints` as `in` operator signals the presence of either accessors or values but it doesn't reach these (lazy accessors can backfire, I have learned this [long time ago](https://webreflection.blogspot.com/2011/08/please-stop-reassigning-for-no-reason.html))

WebReflection:
@ryanking13 *nit* a part, it works like a charm:

image

cpython/patches/0009-Fix-iPad-detection-in-wasm-gc.patch
hoodmane: Also, I think comparing `typeof something` to `"undefined"` should only be done for a top level name because `x === undefined` raises an error in strict mode. But `a.b === undefined` does not. They are distinct checks since if you set `a.b = undefined` then `"b" in a` is true.

But this check only happens once on startup so it's not all that important exactly how it happens as long as it filters out all the iDevices.

WebReflection: the fact `in` produces `true` but the value is `undefined` won't invalidate what I've said because `undefined > 1` is always false but, logically speaking, the check should be `typeof navigator.maxTouchPoints === 'number' && navigator.maxTouchPoints > 1` if we want to be sure that `typeof` has a reason to exist at all in there ... anyway, it's a *nit* but the current check is both redundant and not clean in its intent ... glad it shipped though 🥳

ryanking13: Yes, thanks for the heads up @WebReflection! Since upstream PR is merged as-is (that was fast 🙄), I would like to merge this as-is to prevent future conflict. This patch is a hack anyways, so let's drop it when Apple delivers a fix.

hoodmane:
Thanks @ryanking13 !

PR #1342, opened 2021-03-19
DOC Update project history and infrastructure support sections

rth:
I added the documentation sections about the project history and infractructure support, as part of https://github.com/iodide-project/pyodide/issues/1217

 @wlach @mdboom @bcolloran @hamilton @teonbrooks  If there are other things you would like to add there, for instance about the Iodide project there please don't hesitate to comment.

PR #5700, opened 2025-06-16
DOC Add special thanks to BrowserStack

ryanking13:
Adds special thanks section to BrowserStack for OSS testing.

### DIFF
--- a/README.md
+++ b/README.md
@@ -87,7 +87,6 @@ Pyodide also has a large number of small donors. If you’re interested in suppo
 
 ### Special thanks
 
-- [jsDelivr](https://www.jsdelivr.com/): Pyodide distribution is hosted on jsDelivr
 - [BrowserStack](https://www.browserstack.com/): This project is tested with BrowserStack
 
 ## License
--- a/README.md
+++ b/README.md
@@ -90,7 +90,6 @@ Pyodide also has a large number of small donors. If you’re interested in suppo
 - [jsDelivr](https://www.jsdelivr.com/): Pyodide distribution is hosted on jsDelivr
 - [BrowserStack](https://www.browserstack.com/): This project is tested with BrowserStack
 
-
 ## License
 
 Pyodide uses the [Mozilla Public License Version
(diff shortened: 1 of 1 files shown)


ryanking13:
@rth This PR is mostly for [BrowserStack Open Source Sponsorship](https://www.browserstack.com/open-source), but I added jsDelivr as well, as IIRC we sometimes ask jsDelivr to handle Pyodide-specific requests. I wonder if we are treated something special by jsDelivr.

agriyakhetarpal:
Thanks! I wonder if we should also thank CircleCI for the generous CI minutes. I'm not sure if we have any formal arrangement with them, but we could give credit where it is due?

ryanking13:
Yeah, obviously Circle CI is offering more credits than the free tier. I am not sure if they had something like open-source sponsorship program that we'd applied.

rth:
There is already a list in https://github.com/pyodide/pyodide/blob/02094ed74f50a9286c92cc94ac3b9c2b24c286a2/docs/project/about.md?plain=1#L102 so if we include some of those in the readme, we probably should still link to the full list.

> I wonder if we are treated something special by jsDelivr.

With Hood we met with them at EuroPython in Prague and occasionally exchanged by email (I can intro if you ever need). But it's also that the bandwidth is funded by their sponsors (Fastly, Cloudflare, and other CDNs) so indirectly we are also benefiting from that.

> [CircleCI] I am not sure if they had something like open-source sponsorship program that we'd applied.

I don't remember well, I think either we tried and didn't get it or we didn't try because we already were getting a lot of credits. I feel like either it's their default behavior for OSS to allow going above the limit, or some nice person at CircleCI just made an exception for our project (or at least didn't enforce the limit). 

Screenshot 2025-06-16 at 12 32 39

rth:
https://github.com/pyodide/pyodide/pull/1342/files#diff-d8c7802abb4d194b5f7280ec41c461eca69cb38aed169351021a518f8f99a079R29 mentions Mozilla and CircleCI for CI 4 years ago. But I'm not sure, I think Mozilla had a paid CircleCI account that we were using before Pyodide became a independent project. So our current setup should not be linked to that.

ryanking13:
I see. Thanks for the explanation!

PR #5708, opened 2025-06-20
Enable WebGL2 (-sMAX_WEBGL_VERSION=2)

pthom:
### Description

By default, emscripten does not enable WebGL 2 (only WebGL 1 is available).
This PR add support for WebGL2, by setting -sMAX_WEBGL_VERSION=2

(WebGL 2 was finalised and first appeared in stable browsers in January 2017)

**Impacts**
See [doc for MAX_WEBGL_VERSION](https://emscripten.org/docs/tools_reference/settings_reference.html#max-webgl-version)

*Setting MAX_WEBGL_VERSION=2 implies that*
- WebGL 2 is now available (it was not available without this flag)
- WebGL 1 is *still available*, but WebGL 2 will be the default if no version is specified
- WebGL 1 can be required explicitly; for example, by using OpenGL ES 2.0 like in the example below: 

```cpp
SDL_GL_SetAttribute(SDL_GL_CONTEXT_PROFILE_MASK, SDL_GL_CONTEXT_PROFILE_ES );
SDL_GL_SetAttribute(SDL_GL_CONTEXT_MAJOR_VERSION, 2);
SDL_GL_SetAttribute(SDL_GL_CONTEXT_MINOR_VERSION, 0);
```

### Related PR

This PR is related to [this pyodide-recipes PR](https://github.com/pyodide/pyodide-recipes/pull/116) which add the imgui_bundle package (and requires WebGL2 to work).

### Checklist

- [ ] Add / update tests

I don't think it will be required to add test and update the documentation. 

I'd be happy to discuss about this.

### DIFF
--- a/Makefile.envs
+++ b/Makefile.envs
@@ -191,6 +191,7 @@ export MAIN_MODULE_LDFLAGS= $(LDFLAGS_BASE) \
 	-lhtml5.js \
 	-lhtml5_webgl.js \
 	-lsdl.js \
+	-sMAX_WEBGL_VERSION=2  \
 	-sGL_WORKAROUND_SAFARI_GETCONTEXT_BUG=0
 
 EXPORTS=_main \
--- a/docs/project/changelog.md
+++ b/docs/project/changelog.md
@@ -62,6 +62,10 @@ myst:
 
 - {{ Fix }} Fixed iPad + Safari issue started to happen since 0.27.1. {pr}`5695`
 
+- {{ Enhancement }} Enable WebGL 2 (-sMAX_WEBGL_VERSION=2).
+  WebGL 1 is still available but must be required explicitly
+  (for example, by using OpenGL ES 2.0)
+
 ### `python` CLI entrypoint
 
 - {{ Fix }} The `python` CLI now mounts the `/tmp` directory. In


ryanking13 (commented):
@hoodmane, what do you think about this PR? We already released Pyodide 0.28.0a3 with frozen ABI, so adding a new flag to the main module should be done cautiously.

If this is backward compatible change, then I think it is okay to include this, but I wanted to make sure about it.

pthom:
I did a test to see if it is possible to successfully mix a "newer pyodide.js compiled *with this flag ( -sMAX_WEBGL_VERSION=2)*" with "older packages, compiled *before adding the flag*".

Here is what I did:
* Step 1: compile all pyodide-recipes and pyodide *without this flag*
* Step 2: compile pyodide.js *with this flag* (but do not compile the recipes)
* Step 3: run a test by loading a HTML page which load "a newer pyodide.js compiled *with this flag*", and "older packages, compiled *before adding the flag*"

The page did work perfectly, and displayed a full demo of ImGui Bundle where I could validate that the older packages numpy, typing_extensions, pydantic_core, pillow, opencv-python, imgui_bundle (compiled before adding the flag) are compatible with a new pyodide (after adding the flag). In the full demo, I could run successfully python code that calls all of these packages.

Below is its code: it shows that pyodide.js and the packages were loaded from different compilation sources.

```html
<!doctype html>
<html>
<head>
    <style>
        html, body { width: 100%; height: 100%; margin: 0; }
        #canvas { display: block; width: 100%; height: 100%;}
    </style>
    <script src="pyodide_dist/pyodide.js"></script>
</head>
<body>
<canvas id="canvas"></canvas>
    // ====================== Start of Python code ============================
    pythonCode = `
# Runs the full ImGui Bundle demo
from imgui_bundle import immapp
from imgui_bundle.demos_python import demo_imgui_bundle
runner_params, addons = demo_imgui_bundle.make_params()
immapp.run(runner_params=runner_params, add_ons_params=addons)
`
    // ====================== End of Python code ==============================


    async function main(){
        document.addEventListener('contextmenu', event => event.preventDefault());

        // Load Pyodide, from pyodide_dist/pyodide.js (compiled with the flag -sMAX_WEBGL_VERSION=2)

        // Load older packages, compiled *before adding the flag -sMAX_WEBGL_VERSION=2*
        await pyodide.loadPackage("pyodide_dist_nowebgl/numpy-2.2.5-cp313-cp313-pyodide_2025_0_wasm32.whl");
        await pyodide.loadPackage("pyodide_dist_nowebgl/typing_extensions-4.12.2-py3-none-any.whl");
        await pyodide.loadPackage("pyodide_dist_nowebgl/pydantic_core-2.27.2-cp313-cp313-pyodide_2025_0_wasm32.whl");
        await pyodide.loadPackage("pyodide_dist_nowebgl/pydantic-2.10.6-py3-none-any.whl");
        await pyodide.loadPackage("pyodide_dist_nowebgl/munch-4.0.0-py2.py3-none-any.whl");
        await pyodide.loadPackage("pyodide_dist_nowebgl/pillow-11.2.1-cp313-cp313-pyodide_2025_0_wasm32.whl");
        await pyodide.loadPackage("pyodide_dist_nowebgl/opencv_python-4.11.0.86-cp313-cp313-pyodide_2025_0_wasm32.whl");
        await pyodide.loadPackage("pyodide_dist_nowebgl/imgui_bundle-1.6.3-cp313-cp313-pyodide_2025_0_wasm32.whl");

        // Setup SDL
        let sdl2Canvas = document.getElementById("canvas");
        pyodide.canvas.setCanvas2D(sdl2Canvas);
        pyodide._api._skip_unwind_fatal_error = true;

        pyodide.runPython(pythonCode);
    }
    main();
</script>
</body>
</html>
```

In summary:
- `loadPyodide()` did work (and loaded a newer pyodide.js compiled with this flag)
- the calls to `loadPackage()` did work (and loaded olders packages compiled before this flag)
- `runPython(pythonCode)` did work 
- Bonus point: `runPython(pythonCode)` succeeded in creating a WebGL2 canvas. *This failed when using a pyodide.js  compiled without this flag.*

I know that this is just a surface test, but it might bring some interesting insight.

Thanks!

ryanking13:
Thanks for the testing packages. Could you also test `pyxel` with it? I think it also has something to do with WebGL. I have made a [playground](https://github.com/ryanking13/pyodide-pyxel-test) to test pyxel examples in Pyodide.

One thing I would also mention is that this PR increases the size of the pyodide.asm.js around 60KB, which is not that big but not that small as well.

before:

```
- pyodide.asm.js:
    Original size: 986 KB
    Gzip compressed size (level 1): 247 KB
    Gzip compressed size (level 6): 207 KB
    Gzip compressed size (level 9): 207 KB
    Brotli compressed size: 170 KB
```

after:

```
- pyodide.asm.js:
    Original size: 1045 KB
    Gzip compressed size (level 1): 257 KB
    Gzip compressed size (level 6): 214 KB
    Gzip compressed size (level 9): 214 KB
    Brotli compressed size: 176 KB
```

ryanking13:
@szabolcsdombi Could you also check if zengl works well with this change? I think zengl would be a major user of WebGL.

pthom:
> Could you also test `pyxel` with it? I think it also has something to do with WebGL. I have made a [playground](https://github.com/ryanking13/pyodide-pyxel-test) to test pyxel examples in Pyodide.

I cloned your pyxel playground and ran some tests with it. It worked perfectly. 

What I tested:

- using "a pyodide.js compiled with this flag", and "pyxel package, compiled before or after adding the flag" works. 
  Apps are launched correctly and the SDL canvas works. This means that *2D SDL apps built with pyxel are not impacted by the addition of WebGL2*.
- using "a pyodide.js compiled without this flag", and "pyxel package compiled after adding the flag" works.


Note: after cloning your playground I had to make a few changes so that I uses my compiled versions of pyodide and pyxel wheel.S

In practice I did change the pyxel.js  utility loader like so: 
```diff
diff --git a/pyxel.js b/pyxel.js
index c6e3224..883a781 100644
--- a/pyxel.js
+++ b/pyxel.js
@@ -1,7 +1,7 @@
 const NO_SLEEP_URL =
   "https://cdnjs.cloudflare.com/ajax/libs/nosleep/0.12.0/NoSleep.min.js";
 const PYODIDE_SDL2_URL =
-  "https://cdn.jsdelivr.net/pyodide/v0.23.0/full/pyodide.js";
+  "/pyodide_webgl/pyodide.js";  // use a webgl2 or nowebgl2 version here 
 const PYXEL_LOGO_PATH = "./docs/images/pyxel_logo_76x32.png";
 const TOUCH_TO_START_PATH = "./docs/images/touch_to_start_114x14.png";
 const CLICK_TO_START_PATH = "./docs/images/click_to_start_114x14.png";
@@ -152,7 +152,7 @@ async function _loadPyodideAndPyxel() {
   noSleep.enable();
   await _loadScript(PYODIDE_SDL2_URL);
   let pyodide = await loadPyodide();
-  await pyodide.loadPackage("pyxel");
+ //  use a webgl2 or nowebgl2 version here 
+  await pyodide.loadPackage("/pyodide_nowebgl/pyxel-1.9.10-cp37-abi3-pyodide_2025_0_wasm32.whl"); 
   let FS = pyodide.FS;
   FS.mkdir(PYXEL_WORKING_DIRECTORY);
```

---

> Does zengl works well with this change?

IMHO zengl might benefit from this as well, but I am not its author.

szabolcsdombi:
ZenGL should be unaffected by this change. It uses webgl2 already by a [custom adapter code](https://github.com/szabolcsdombi/zengl/blob/main/zengl.js#L46). These functions are not generic, they only work for the zengl use case, the correct code to mock a desktop gl api over webgl2 would be way more complex.

This change would help with the canvas. This will allow me to use an SDL window for the zengl examples. right now we [get or create a canvas](https://github.com/szabolcsdombi/zengl/blob/main/_zengl.py#L368-L387).

I will test ZenGL with this change either way, but please don't block on us, if there is anything we need to change I will update ZenGL.

Looking forward to this change!

ryanking13:
Got it. Thanks for the answers both! I'll wait for a few more days and merge this if there is no input from @hoodmane.

hoodmane:
Sounds fine to me.

ryanking13:
Thanks! @pthom, when the PR is ready, please mark it ready for review, then I'll merge.

PR #5711, opened 2025-06-24
TST Add dummy packages for testing and remove regex and Jinja2

ryanking13:
Removes `regex` and `Jinja2` recipes from this repository, and replace it with dummy packages only for testing.


hoodmane (approved):
Thanks!

PR #5715, opened 2025-06-26
Bump to NumPy 2.3.1, and split out `numpy-tests` (backport of pyodide/pyodide-recipes#86)

agriyakhetarpal:
### Description

This ports https://github.com/pyodide/pyodide-recipes/pull/86 to the Pyodide runtime repository.

### Checklist


ryanking13 (approved):
Thanks! It looks like the scipy version that we are using is slightly incompatible with the Numpy 2.3.1.

agriyakhetarpal:
Ah, right, let me try to get to #5297 first. I'll start debugging it in the pyodide-recipes repository!

juntyr:
numpy v2.4 was released yesterday and it would be great to support it, should this PR be merged first or updated to include the 2.4 changes?

ryanking13:
> numpy v2.4 was released yesterday and it would be great to support it, should this PR be merged first or updated to include the 2.4 changes?

In any case, we need to bump scipy first...

juntyr:
> Thanks! It looks like the scipy version that we are using is slightly incompatible with the Numpy 2.3.1.

What incompatibilities did arise when upgrading numpy before scipy? I've tried looking for an issue mentioning this but came up empty

juntyr:
> > Thanks! It looks like the scipy version that we are using is slightly incompatible with the Numpy 2.3.1.
> 
> What incompatibilities did arise when upgrading numpy before scipy? I've tried looking for an issue mentioning this but came up empty

I should have checked before, the current scipy bounds the version of numpy to <2.3

ryanking13:
we now have numpy 2.4.3 here so I think we can close this. also numpy-tests split is applied in pyodide-recipes.

Issue #5657, opened 2025-05-22
Don't call reportUndefinedSymbols in no dylink build


hoodmane:
BTW. why do you need reportUndefinedSymbols to be callable in the statically linked build? Can't the calling code be made to only conditionally call this?

e.g. if (isDyanamicallyLinked) Module.reportUndefinedSymbols() or wrap the call in #if MAIN_MODULE if the code is being pre-processed? (BTW did you know you can use the pre-processor in pre-js files now if you add #preprocess to the top of them?). Monkey patching the generated code for a static build to make it appear as a dynamic build seem a little odd.

_Originally posted by @sbc100 in https://github.com/pyodide/pyodide/pull/5656#issuecomment-2901947125_

ryanking13:
As you commented in the original PR (and as sbc100 suggested), maybe we can stop using reportUndefinedSymbols after landing RPATH changes.

Also, reportUndefinedSymbols was no op for 1 month after we upgraded to Emscripten 4.0.6, and there was no issue related to that (except for SDL packages), so maybe it is actually not needed anymore.

ryanking13:
Closing as resolved

PR #5716, opened 2025-06-26
Remove reportUndefinedSymbols

ryanking13:
As discussed in #5657. Let's see if it works.

### DIFF
--- a/src/core/pre.js
+++ b/src/core/pre.js
@@ -63,9 +63,6 @@ const errNoRet = () => {
   );
 };
 
-// For no-dylink build, this is a no-op.
-Module.reportUndefinedSymbols ??= () => {};
-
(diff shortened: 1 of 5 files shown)


hoodmane:
I'm pretty sure we can't leave this out.

ryanking13:
I removed `allowUndefined` flag in https://github.com/pyodide/pyodide/pull/5610, so I believe reportUndefinedSymbols is always called internally [here](https://github.com/emscripten-core/emscripten/blob/267448c1b3f90e3867a41e53e5bc2deadc6951af/src/lib/libdylink.js#L774C20-L776)

hoodmane:
Makes sense. Looks good to me then.

hoodmane (approved):

Issue #3968, opened 2023-06-29
Provide a way to convert None to null in to_js


philippjfr:
## 🚀 Feature

I propose some functionality be added to `pyodide.ffi.to_js` to allow Python `None` values to be translated into `null` values rather than `undefined` values. 

### Motivation

Currently `pyodide.ffi.to_js` converts all encountered `None` values to `undefined` on the JS side. In certain scenarios this is problematic because some downstream consumer of the converted value might expect `null` rather than `undefined`. One such case are Bokeh event messages. When converting the events generated by Bokeh in Python into JS objects that can be applied to a Bokeh Document they often contain serialized copies of the Model attribute values that have changed and Bokeh's property system makes a very clear distinction between a `null` value and an `undefined` value. Bokeh makes this distinction because `undefined` implies the value has not been set, while `null` is a perfectly valid value.

### Pitch

Ideally I'd simply provide some way to override mappings for specific types, e.g. `{None: pyodide.code.run_js('null')}`. Alternatively I'd also be happy for a custom `dict_converter` to be made available that allows overriding the translation.

hoodmane:
Does this work for you:
```py
from pyodide.ffi import to_js
from pyodide.code import run_js

dict_converter = run_js(
	"""
	((entries) => {
		for(let entry of entries) {
			if(entry[1] === undefined) {
				entry[1] = null;
			}
		}
		return Object.fromEntries(entries);
	})
	"""
)

to_js(some_object, dict_converter=dict_converter)
```

philippjfr:
It does indeed.

PR #5719, opened 2025-06-27
Add ffi support for JS null and switch to wasm-gc struct for error signaling

hoodmane:
We need JS null to represent actual null and not an error. Instead, we use wasm-gc except on safari iOS
where wasm-gc is broken or on old runtimes where wasm-gc support is missing. If we can't use wasm-gc,
we use a JS object and a JS trampoline. There is maybe a 3-4% performance hit associated with
this fallback compared to before this PR. Using wasm-gc to signal errors leads to no measured performance
change.

Resolves https://github.com/pyodide/pyodide/issues/3968.

- [x] Add / update tests

Left to a followup to keep diff down:
- [ ] Rename APIs that are no longer null related.


Makefile
```diff
@@ -68,6 +68,13 @@ src/core/pyodide_pre.gen.dat: src/js/generated/_pyodide.out.js src/core/pre.js s
 src/core/pyodide_pre.o: src/core/pyodide_pre.c src/core/pyodide_pre.gen.dat
 	unset _EMCC_CCACHE && emcc --std=c23 -c $< -o $@
 
+src/core/.installed_wasm_tools:
+	RUSTFLAGS= cargo install --locked wasm-tools
```
hoodmane: TODO: cargo binstall or grab from github releases?

Makefile.envs
```diff
@@ -173,6 +173,7 @@ export MAIN_MODULE_LDFLAGS= $(LDFLAGS_BASE) \
 	-s NODEJS_CATCH_REJECTION=0 \
 	-s EXPORTED_FUNCTIONS='$(EXPORTS)'\
 	-s EXIT_RUNTIME \
+	-sERROR_ON_UNDEFINED_SYMBOLS=0 \
```
ryanking13, line 178: why is this need to be allowed?

Makefile
```diff
@@ -130,6 +134,9 @@ dist/pyodide.asm.js: \
 	# Sed nonsense from https://stackoverflow.com/a/13383331
 	sed -i -n -e :a -e '1,7!{P;N;D;};N;ba' dist/pyodide.asm.js
 	echo "globalThis._createPyodideModule = _createPyodideModule;" >> dist/pyodide.asm.js
+
+	./emsdk/emsdk/upstream/bin/wasm-merge dist/pyodide.asm.wasm blah src/core/sentinel.wasm sentinel -o dist/pyodide.asm.wasm -all
```
ryanking13: What is `blah` for?

src/js/emscripten-settings.ts
```diff
@@ -5,6 +5,11 @@ import { initializeNativeFS } from "./nativefs";
 import { loadBinaryFile, getBinaryResponse } from "./compat";
 import { API, PreRunFunc, type Module } from "./types";
 
+// @ts-ignore
```
ryanking13: Could you please add the exact error that is ignored?

src/core/sentinel.wat
```diff
@@ -0,0 +1,28 @@
+(module
+  (type $empty_struct (struct))
+  (import "env" "is_safari" (global $is_safari i32))
+  (import "env" "safari_error" (global $safari_error externref))
+  (import "env" "safari_is_error" (func $safari_is_error (param externref) (result i32)))
```
ryanking13: Could you please cross-reference `emscripten-settings.ts` to know where these variables come from?

src/core/sentinel.wat
```diff
@@ -0,0 +1,28 @@
+(module
```
ryanking13, line 11: It would be nice to add a module comment to let the reader know why this module is needed, and when we can remove it.

src/js/emscripten-settings.ts
```diff
@@ -5,6 +5,11 @@ import { initializeNativeFS } from "./nativefs";
...
+import sentinelWasm from "./generated/sentinel.wasm";
+
+declare const sentinelWasm: Uint8Array;
```
ryanking13: I don't see any place where this variable is used. Is it for side effect?

ryanking13 (commented):
I am curious what's the benefit of using actual null instead of JS null (not opposed to this PR, just curious).

Makefile.envs
hoodmane: `create_sentinel` and `is_sentinel` are either defined by `sentinel.wat` or in JavaScript. So Emscripten doesn't know they are defined.

Makefile
hoodmane: This is outdated but `wasm-merge` provides some wasm module imports using another wasm module. `blah` is the name of the namespace that `pyodide.asm.wasm` is going to provide to other modules. Since we are only trying to resolve `sentinel` imports in `pyodide.asm.wasm` to functions from `sentinel.wasm` and not the other way around, `pyodide.asm.wasm` doesn't really need a name. OTOH it is important that `sentinel.wasm` is used to provide the `sentinel` namespace in `pyodide.asm.wasm`.

Unfortunately, we can't use `wasm-merge` because wasm-gc has not been supported long enough in JS engines for us to require it. So we need a separate wasm module and a fallback for it compilation fails.

src/js/emscripten-settings.ts
hoodmane: We `WebAssembly.compile()` it. I moved it to `sentinel.ts` which should be a bit easier to read hopefully.

hoodmane:
The point is to allow JS APIs that distinguish between `null` and `undefined`. For instance @ambv tells me that three.js treats them differently. Some Cloudflare APIs also treat null and undefined differently. At least three different people have complained to me about this in the last six months actualy. Related bug report:
https://github.com/pyodide/pyodide/issues/3968

hoodmane:
@ryanking13 I changed a bunch of stuff since your first review so if you could rereview and let me know what you think now. There still seem to be some crashes to iron out but I think they are minor find-and-replace style changes.

Makefile.envs
ryanking13: Could you mention that in Makefile too?

src/tests/test_typeconversions.py
```diff
@@ -799,14 +799,15 @@ def test_pythonexc2js(selenium):
...
-    assert run_js("null") is None
-    assert run_js("[null]")[0] is None
-    assert run_js("() => null")() is None
-    assert run_js("({a: null})").a is None
-    assert run_js("new Map([['a', null]])")["a"] is None
-    assert run_js("[null, null, null]").to_py() == [None, None, None]
-    assert run_js("new Map([['a', null]])").to_py() == {"a": None}
+    from pyodide.ffi import jsnull
+
+    assert run_js("null") is jsnull
```
ryanking13, line 804: This is going to be breaking change, so we should mention it in the changelog clearly.

src/core/sentinel.ts
```diff
@@ -0,0 +1,35 @@
+// @ts-ignore
+import sentinelWasm from "./sentinel.wasm";
+
+declare const sentinelWasm: Uint8Array;
+
+const sentinelInstancePromise: Promise<WebAssembly.Instance | undefined> =
+  (async function () {
+    const isIOS =
+      globalThis.navigator && /iPad|iPhone|iPod/.test(navigator.platform);
```
ryanking13: this check was not successful in iPad, so probably need to be updated:

https://github.com/pyodide/pyodide/blob/main/cpython/patches/0009-Fix-iPad-detection-in-wasm-gc.patch

src/core/sentinel.ts
```diff
@@ -0,0 +1,35 @@
...
+      if (e instanceof WebAssembly.CompileError) {
+        return undefined;
+      }
+      throw e;
+    }
+  })();
+
+export async function getSentinelImport() {
+  await sentinelInstancePromise;
+  const sentinelInstance = await sentinelInstancePromise;
```
ryanking13: Duplicate code

src/core/sentinel.ts
```diff
@@ -0,0 +1,35 @@
...
+      return await WebAssembly.instantiate(module);
+    } catch (e) {
+      if (e instanceof WebAssembly.CompileError) {
+        return undefined;
+      }
+      throw e;
+    }
+  })();
+
+export async function getSentinelImport() {
```
ryanking13: Could you add a type hint for the return type?

ryanking13 (approved):
Thanks. There is a build failure, but codewise looks fine to me.

src/core/sentinel.ts
hoodmane: Oops I copied it from the wrong branch of Python. Thanks!

hoodmane:
> 422 tests failed  out of 1439

So not everything is broken =)

Issue #5705, opened 2025-06-18
STATUS_STACK_BUFFER_OVERRUN crash and memory issue


caffeinepills:
## 🐛 Bug

When attempting to utilize scheduling with asyncio in python, I have been getting inconsistent crashes. 

The crash usually happens 6+ seconds after running, but generally the page (in Chrome) will state: 

![Image](https://github.com/user-attachments/assets/b7c65c59-14a9-4d4e-bafd-ad5fe9a1b18e)

Occasionally I may get something like this instead:
```
pyodide.asm.js:10 Pyodide has suffered a fatal error. Please report this to the Pyodide maintainers.
API.fatal_error @ pyodide.asm.js:10Understand this error
pyodide.asm.js:10 The cause of the fatal error was:
API.fatal_error @ pyodide.asm.js:10Understand this error
pyodide.asm.js:10 RangeError: WebAssembly.Table.get(): invalid address 10889992 in funcref table of size 13696
    at Object.calculateWasmFuncNargsFallback [as PyEM_CountArgs] (pyodide.asm.js:10:121024)
    at _PyEM_CountFuncParams (pyodide.asm.js:10:127268)
    at pyodide.asm.wasm:0x328712
    at pyodide.asm.wasm:0x1c441d
    at pyodide.asm.wasm:0x1c4c3c
    at pyodide.asm.wasm:0x2a2ba3
    at pyodide.asm.wasm:0x2a73f5
    at pyodide.asm.wasm:0x1c4fe5
    at pyodide.asm.wasm:0x1c6a8d
    at pyodide.asm.wasm:0x1c4b4f
```

### To Reproduce

After some troubleshooting down to a bare bones example. If I set the sleep interval using asyncio to a low number like 0 or 0.00001 then it will crash the tab after a little bit. 

```py
import asyncio

async def run_test(interval):
    while True:
        await asyncio.sleep(interval)

asyncio.create_task(run_test(0))
```

### Expected behavior

No crash.

### Environment

- Pyodide Version: 0.26.3 and up.
- Browser version: 137.0.7151.104 
- Any other relevant information:

### Additional context

I tested this and the behavior seems to start with 0.26.3, if I use 0.26.2, it doesn't crash, although the memory usage grows to an extreme size 500MB-1 GB according to the memory tab.

paultiq:
Dupe of #5702?

caffeinepills:
[...]

PR #5721, opened 2025-06-28
Try MessageChannel first in scheduleCallbackImmediate

ryanking13:
### Description

This partially resolves the issue in https://github.com/pyodide/pyodide/issues/5705.

While investigating the issue, I noticed that MessageChannel does not leak memory, but the postMessage does. However, I wasn't able to find the root cause... after struggling a few hours. And interestingly, the memory leak doesn't seem to happen in Safari (or at least, Safari GC seems to collect the memory before it gets blown up).

So, as a workaround, I changed the method order in `scheduleCallbackImmediate` to favor `MessageChannel` over `postMessage`.

### Checklist

- [ ] Add / update tests


hoodmane (approved):
Okay...

hoodmane:
Thanks!

ryanking13:
We can revisit this, maybe after releasing 0.28. I don't want to block the release because of this.

PR #5722, opened 2025-07-01
Rename JS_NULL to JS_ERROR, NFC

hoodmane:
I separated this from #5719 to keep the diff from getting too large.


ryanking13 (approved):
Thanks!

PR #5723, opened 2025-07-01
PERF Use sentinel for faster Jsv_novalue

hoodmane:
This will speed up python2js a bit.

### DIFF
--- a/src/core/jslib.c
+++ b/src/core/jslib.c
@@ -45,17 +45,21 @@ Jsv_GetNoValue(void)
   return create_sentinel(NO_VALUE_SENTINEL);
 }
 
-__attribute__((import_module("sentinel"),
-               import_name("sentinel_get_value"))) int sentinel_get_value(JsVal);
+__attribute__((
+  import_module("sentinel"),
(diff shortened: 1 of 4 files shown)


src/core/jslib.c
```diff
@@ -28,19 +28,42 @@ JS_INIT_CONSTS();
...
+
+int
+JsvNoValue_Check(JsVal x)
+{
+  return sentinel_get_value(x) == NO_VALUE_SENTINEL;
+}
 
 EM_JS_NUM(int, jslib_init_js, (void), {
   JS_INIT_CONSTS();
+  _hiwire_intern("end of constants");
```
ryanking13, line 66: Can we use something like ALL CAP STRING to more clearly mark this is a magic string for snapshot?

Also adding some comment would be nice.

ryanking13 (commented):
Looks like there is a test failure?

hoodmane:
I think the nodejs failure might be an incorrect JIT? Adding debug prints or even a no-op function call makes it go away. Marking as draft since this'll be hard to fix most likely.

PR #5732, opened 2025-07-04
Override lockfile version in the prebuilt packages

ryanking13:

PR #5733, opened 2025-07-04
Compare ABI version not Pyodide version

ryanking13:
### Description

Updates the lockfile version check to compare the ABI version, not the Pyodide version. The prebuilt packages may have a different Pyodide version if they use alpha or nightly xbuildenv.


hoodmane:
Nice. We should consider deprecating the pyodide_version key. I think we can merge this one instead of #5732?

ryanking13:
Yes, this can supersede #5732

Issue #5736, opened 2025-07-04
0.28.0: micropip lockFileURL failing when provided


WebReflection:
## 🐛 Bug

We've discussed this in Discord but I haven't seen any bug around it so here it is.

### To Reproduce

This page tries to store the `micropip.freeze()` result somewhere and reuse it for any other reload of the page.

As soon as that frozen cache is used as valid URL, an error about `TypeError: _emscripten_stack_get_current is not a function` is shown:

```html
<!DOCTYPE html>
<html lang="en">
<head>
    <meta charset="UTF-8">
    <meta name="viewport" content="width=device-width, initial-scale=1.0">
    <script type="module">
        // add ?clear to your test page to drop the entry
        if (location.search === '?clear') {
            localStorage.removeItem("test");
            location.href = location.href.slice(0, -6);
            throw new Error('reloading anyway');
        }

        import { loadPyodide } from "https://cdn.jsdelivr.net/pyodide/v0.28.0/full/pyodide.mjs";

        let timeout = 0;
        addEventListener('error', () => {
            clearTimeout(timeout);
            document.body.append('ERROR');
        });

        const json = localStorage.getItem("test");
        const options = {
            // not specific to the package, anyone fails
            packages: ["numpy"],
        };

        if (json) {
            options.lockFileURL = URL.createObjectURL(new Blob(
                [json],
                { type: 'application/json' },
            ));
        }

        const pyodide = await loadPyodide(options);

        if (json) {
            // irrelevant for the bug, just for correctness/example
            URL.revokeObjectURL(options.lockFileURL);
        }
        else {
            await pyodide.loadPackage('micropip');
            const micropip = pyodide.pyimport('micropip');
            await micropip.install(options.packages, { keep_going: true });
            const frozen = micropip.freeze();
            localStorage.setItem("test", frozen);
            micropip.destroy();
            // first time it's fine, next time refresh
            // and you'll see a weird error
        }

        timeout = setTimeout(() => {
            document.body.append('ALL GOOD - REFRESH');
        }, 1000);
    </script>
</head>
</html>
```

### Expected behavior

We use IndexedDB instead but the outcome is the same, what supposed to be a way to ensure faster bootstrap became "*harakiri*" because without any caching attempt things just work and things worked up to 0.28.0 release, but nothing in the BREAKING CHANGES suggests we're doing something really wrong and the error is a bit cryptic because it doesn't provide any follow-up action or detail about what's actually wrong in the process.

### Environment

[...]

PR #5737, opened 2025-07-05
FIX Buffer messages in PackageManager when the stream is not ready

ryanking13:
### Description

This fixes half of the bug in #5736. It does not fix the real error, but fixes the incorrect error message `TypeError: _emscripten_stack_get_current is not a function` being printed, and makes the stacktrace show the real error.

The problem was that, after #5621, the PackageManager uses the native APIs to print messages, which requires the bootstrap to finish to print messages. However, when the packages are preloaded by passing `packages` option in `loadPyodide`, the bootstrap doesn't finish until `bootstrapFinalizedPromise` is resolved.

This PR makes a temporary buffer that stores the streams, and flush it when the bootstrap is finalized. After this change, the reproducer in #5736 shows the real error message.

image

### Checklist

- [x] Add / update tests

### DIFF
--- a/src/js/load-package.ts
+++ b/src/js/load-package.ts
@@ -84,7 +84,7 @@ export async function initializePackageIndex(
   if (API.config.fullStdLib) {
     toLoad = [...toLoad, ...API.lockfile_unvendored_stdlibs];
   }
-  await loadPackage(toLoad);
+  await loadPackage(toLoad, { messageCallback() {} });
   // Have to wait for bootstrapFinalizedPromise before calling Python APIs
   await API.bootstrapFinalizedPromise;
   API.flushPackageManagerBuffers();
--- a/src/js/load-package.ts
+++ b/src/js/load-package.ts
@@ -84,9 +84,11 @@ export async function initializePackageIndex(
   if (API.config.fullStdLib) {
     toLoad = [...toLoad, ...API.lockfile_unvendored_stdlibs];
   }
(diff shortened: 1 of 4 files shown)


hoodmane (approved):
I would say you need a new test but by switching to not suppressing the info logs this is covered by existing tests. Still maybe we should have a test that if a package is not loaded correctly `loadPyodide()` gracefully exits.

Anyways it is more fancy than what I would have done but it makes sense. Thanks!

src/js/load-package.ts
```diff
@@ -84,9 +84,11 @@ export async function initializePackageIndex(
   if (API.config.fullStdLib) {
     toLoad = [...toLoad, ...API.lockfile_unvendored_stdlibs];
   }
-  await loadPackage(toLoad, { messageCallback() {} });
```
hoodmane, line 87: I think this line should be reverted.

ryanking13: Thanks. Reverted

ryanking13:
> Still maybe we should have a test that if a package is not loaded correctly loadPyodide() gracefully fails.

Yeah, I tried to test this change, but it is a little tricky to do so, as we print error (instead of raising an exception) in a very rare case, which is not easy to simulate. Let's go as-is for now, and let me find a better way to improve this behavior.

hoodmane:
Can't we force a 404 in ci? What about by deleting one of the test wheels temporarily?

Issue #5702, opened 2025-06-16
Pyodide Crashes with JSPI Enabled on Memory Leak Test


paultiq:
## Background

I have been seeing periodic crashes in Pyodide when JSPI is enabled that can't be reproduced when JSPI is disabled. 

The closest issue to this I found was #5597 which is listed as fixed in 0.27.7 via #5599. Running the test case from that issue shows that a crash occurs... albeit with different symptoms.

In the prior issue, high memory usage was observed (3GB+) in the webworker. In my test, memory appeared to remain constant with 0.27.7 but still crash. 

With Chrome's JSPI flag disabled, the crash does not occur. 

![Image](https://github.com/user-attachments/assets/c36418ba-3f36-4adf-9774-82b4870d02f4)

-----

The browser error displayed is `STATUS_STACK_BUFFER_OVERRUN`

Edit: Added Chrome error message to make this more searchable

-----

## 🐛 Bug

Running the test case from #5597 against 0.27.7 crashes after 300k iterations on Windows / Chrome with JSPI enabled. With JSPI disabled, test runs to completion. 

### To Reproduce

This is the same test case from #5597 modified to use 0.27.7. Crash also occurs against dev build.

[minimal_bug_demo_27_7.html.txt](https://github.com/user-attachments/files/20756615/minimal_bug_demo_27_7.html.txt)

### Expected behavior

Expect the test case to run to completion, which is does when JSPI is disabled. 

### Environment

- Pyodide Version:       importScripts('https://cdn.jsdelivr.net/pyodide/v0.27.7/full/pyodide.js');
- Browser version: Version 137.0.7151.104 (Official Build) (64-bit)
- Any other relevant information: Windows

### Additional context

```
pyodide.asm.js:10 RuntimeError: null function or function signature mismatch
    at pyodide.asm.wasm:0x2a2e9f
    at pyodide.asm.wasm:0x2a75df
    at pyodide.asm.wasm:0x1c53af
    at pyodide.asm.wasm:0x1c6e57
    at pyodide.asm.wasm:0x1c4f19
    at pyodide.asm.wasm:0x1c51f1
    at pyodide.asm.wasm:0x1c526f
    at pyodide.asm.wasm:0x2a1081
    at pyodide.asm.wasm:0x2a75df
    at pyodide.asm.wasm:0x1c53af
API.fatal_error	@	pyodide.asm.js:10
callPyObjectKwargsPromising	@	pyodide.asm.js:10
[...]

PR #5744, opened 2025-07-08
Tests cases to reproduce #5702

paultiq:
### Description

Additional test cases to reproduce #5702. They're based on the original html tests in the issue. 

Note that the same tests produce different error messages on different environments. 

For example, ` pytest -q -s tests/test_stack_switching.py::test_many_call_promisings_new_t` on my WSL system fails with `E   pytest_pyodide.runner.JavascriptException: RuntimeError: memory access out of bounds`


paultiq:
A few comments:
- Just a note, since it didn't seem to matter: I tested locally against 138, CI is using Chrome 134. 
- The CI doesn't include any Windows Chrome tests. The same tests experience different failures on the different OSs. On Windows, the tab crashes with STATUS_STACK_BUFFER_OVERRUN, and Selenium reports:  "selenium.common.exceptions.WebDriverException: Message: tab crashed". See https://github.com/paultiq/pyodide_testing/actions/runs/16143172501/job/45555218846 for example (showing tab crashed on Windows but timeout on Ubuntu and OSX)

paultiq:
Closing, resolved per #6260

PR #5745, opened 2025-07-08
chore: add missing defaults to `to_js` overloads

MarcoGorelli:
`to_js` was missing some defaults in the overloads, meaning that that overload wouldn't be matched if the user didn't pass the value explicitly

For example, in 
```python
from typing import overload, Literal, reveal_type

@overload
def func(a: str) -> Literal[2]: ...
@overload
def func(a: None) -> Literal[1]: ...

def func(a: str | None = None) -> Literal[1, 2]:
    return 1 if a is None else 2

reveal_type(func())
```
mypy gives
```
main.py:11: error: All overload variants of "func" require at least one argument  [call-overload]
```
but for
```python
from typing import overload, Literal, reveal_type

@overload
def func(a: str) -> Literal[2]: ...
@overload
def func(a: None = None) -> Literal[1]: ...

def func(a: str | None = None) -> Literal[1, 2]:
    return 1 if a is None else 2

reveal_type(func())
```
it gives
```
main.py:11: note: Revealed type is "Literal[1]"
```

### Checklist

- [ ] Add / update tests

### DIFF
--- a/src/py/_pyodide/_core_docs.py
+++ b/src/py/_pyodide/_core_docs.py
@@ -1449,9 +1449,9 @@ def to_js(
     /,
     *,
     depth: int = -1,
-    pyproxies: JsProxy | None,
+    pyproxies: JsProxy | None = None,
     create_pyproxies: bool,
-    dict_converter: None,
+    dict_converter: None = None,
     default_converter: ToJsConverter | None = None,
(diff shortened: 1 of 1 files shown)


hoodmane:
Thanks!

PR #5746, opened 2025-07-08
MAINT Separate loadPyodide into multiple stages

ryanking13:
### Description

This is a preparation for https://github.com/pyodide/pyodide/issues/3316. I am planning to start implementing the hooks for loadPyodide. For that, this PR separates the giant loadPyodide function into multiple small chunks.

The separation is not perfect: There are functions that have mixed responsibilities, but I'll work on refactoring them in subsequent PRs.

### DIFF
--- a/src/js/pyodide.ts
+++ b/src/js/pyodide.ts
@@ -37,25 +37,356 @@ declare const BUILD_ID: string;
  * @hidden
  */
 export type ConfigType = {
+  /**
+   * The URL from which Pyodide will load the main Pyodide runtime and
+   * packages. It is recommended that you leave this unchanged, providing an
+   * incorrect value can cause broken behavior.
+   *
+   * Default: The url that Pyodide is loaded from with the file name
+   * (``pyodide.js`` or ``pyodide.mjs``) removed.
+   */
   indexURL: string;
+
+  /**
+   * The file path where packages will be cached in node. If a package
+   * exists in ``packageCacheDir`` it is loaded from there, otherwise it is
+   * downloaded from the JsDelivr CDN and then cached into ``packageCacheDir``.
+   * Only applies when running in node; ignored in browsers.
+   *
+   * Default: same as indexURL
+   */
   packageCacheDir: string;
+
+  /**
+   * The URL from which Pyodide will load the Pyodide ``pyodide-lock.json`` lock
+   * file. You can produce custom lock files with :py:func:`micropip.freeze`.
+   * Default: ```${indexURL}/pyodide-lock.json```
+   */
   lockFileURL: string;
+
+  /**
+   * Load the full Python standard library. Setting this to false excludes
+   * unvendored modules from the standard library.
+   * Default: ``false``
+   */
   fullStdLib?: boolean;
+
+  /**
+   * The URL from which to load the standard library ``python_stdlib.zip``
+   * file. This URL includes the most of the Python standard library. Some
+   * stdlib modules were unvendored, and can be loaded separately
+   * with ``fullStdLib: true`` option or by their package name.
+   * Default: ```${indexURL}/python_stdlib.zip```
+   */
   stdLibURL?: string;
+
+  /**
+   * Override the standard input callback. Should ask the user for one line of
+   * input. The :js:func:`pyodide.setStdin` function is more flexible and
+   * should be preferred.
+   */
   stdin?: () => string;
+
+  /**
+   * Override the standard output callback. The :js:func:`pyodide.setStdout`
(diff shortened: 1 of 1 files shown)


ryanking13:
~~Oops... sphinx-js doesn't seem to support `Partial`.~~ fixed

src/js/pyodide.ts
```diff
@@ -33,7 +33,7 @@ declare function _createPyodideModule(
 declare const BUILD_ID: string;
 
 /**
- * See documentation for loadPyodide.
+ * See documentation for loadPyodide. We cannot use Partial<ConfigType> as sphinx-js cannot handle it.
```
hoodmane: Maybe we should teach sphinx-js to handle this.

src/js/pyodide.ts
```diff
@@ -48,15 +48,233 @@ export type ConfigType = {
...
+
+  return emscriptenSettings;
+}
+
+/**
+ * @private
+ */
+async function loadWasmScript(
+  config: ConfigType,
+  emscriptenSettings: EmscriptenSettings,
```
hoodmane: Do we use this argument?

src/js/pyodide.ts
```diff
@@ -48,15 +48,233 @@ export type ConfigType = {
...
+  }
+}
+
+/**
+ * @private
+ */
+async function prepareSnapshot(
+  config: ConfigType,
+  emscriptenSettings: EmscriptenSettings,
+): Promise<{ snapshot: Uint8Array | undefined }> {
```
hoodmane: Why not:
```suggestion
): Promise<Uint8Array | undefined> {
```

src/js/pyodide.ts
```diff
@@ -48,15 +48,233 @@ export type ConfigType = {
...
+  let snapshot: Uint8Array | undefined = undefined;
```
hoodmane:
```suggestion
```suggestion
  if (!config._loadSnapshot) {
    return undefined;
  }
  let snapshot: Uint8Array | undefined = undefined;
```

src/js/pyodide.ts
ryanking13: Yeah, I am not sure how difficult it would be though. It would be quite complex to support all TypeScript syntaxs 🤔

src/js/pyodide.ts
ryanking13: It is not needed. Thanks for the catch.

src/js/pyodide.ts
ryanking13: Thanks

src/js/pyodide.ts
ryanking13: Updated sphinx_pyodide to handle Partial.

ryanking13:
I guess I'll need to update sphinx-js first to make it render interface types correctly.

ryanking13:
Okay, the docs look okay now.

src/js/pyodide.ts
```diff
@@ -351,29 +377,99 @@ Pyodide version does not match: '${version}' <==> '${API.version}'. \
...
+ *   const pyodide = await loadPyodide({
+ *     fullStdLib: true,
+ *     stdout: (msg) => console.log(`Pyodide: ${msg}`),
+ *   });
+ *   console.log("Loaded Pyodide");
+ * }
+ * main();
+ */
+export async function loadPyodide(
+  options: Partial<PyodideConfig> = {},
```
hoodmane: I think the types/docs would be a bit cleaner if we put a `?` in front of every property and then used `type Config = Required<PyodideConfig>` internally.

src/js/pyodide.ts
```diff
@@ -51,215 +51,197 @@ declare function _createPyodideModule(
...
-  } = {},
-): Promise<PyodideAPI> {
+async function initializeConfiguration(
+  options: Partial<PyodideConfig> = {},
+): Promise<PyodideConfig> {
+  await initNodeModules();
+
   if (options.lockFileContents && options.lockFileURL) {
     throw new Error("Can't pass both lockFileContents and lockFileURL");
   }
```
hoodmane, line 253: We should the guard before we do any work.

hoodmane (approved):
LGTM other than one comment on types.

src/js/pyodide.ts
ryanking13: Yes, this is the first function that is called in loadPyodide, do you want to change something?

src/js/pyodide.ts
ryanking13: Sounds good. I updated the typing

PR #5749, opened 2025-07-11
CI Fix incorrect lockfile in xbuildenv

ryanking13:
The xbuildenv should contain the lockfile from the prebuilt packages set, not one that built from this repository. This fixes it.

- Related issue: https://github.com/duckdb/duckdb/pull/18173#pullrequestreview-2996403853


hoodmane (approved):
Thanks!

PR #5750, opened 2025-07-11
Fix installBaseUrl calculation in Node.js when relative URL is passed.

ryanking13:
### Description

This fixes a bug in the installBaseURL calculation in Node.js.

When a path such as

```js
loadPyodide({
  lockfileURL: "my-lockfile.json",
})
```

is passed, the installBaseURL is calculated to undefined in Node, which resulted in an error in package loading. This PR changes it to output cwd (`.`) instead.

### Checklist

- [x] Add / update tests

### DIFF
--- a/docs/project/changelog.md
+++ b/docs/project/changelog.md
@@ -21,6 +21,9 @@ myst:
   when the `lockfileURL` was set to a custom URL.
   {pr}`5737`
 
+- {{ Fix }} Fixed a bug in Node.js which providing a relative path to `lockFileURL` parameter of `loadPyodide()` did not work.
+  {pr}`5750`
+
 ## Version 0.28.0
 
 _July 4, 2025_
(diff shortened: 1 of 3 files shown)


hoodmane (approved):
Thanks!

alumni:
It seems that now `pyodide` tries to load the packages from the same folder as the lockfile instead of the cache folder.

```ts
    lockFileURL: '<project>/assets/pyodide-lock.json',
    packageCacheDir: '/tmp/cache/pyodide',
});
```

will cause: 

```
Loading micropip
Didn't find package micropip-0.10.1-py3-none-any.whl locally, attempting to load from <project>/assets/
Failed to load micropip
ENOENT: no such file or directory, open '<project>/assets/micropip-0.10.1-py3-none-any.whl'
  File "/lib/python313.zip/_pyodide/_base.py", line 666, in pyimport_impl
    res = __import__(stem, fromlist=fromlist)
ModuleNotFoundError: No module named 'micropip'
The module 'micropip' is included in the Pyodide distribution, but it is not installed.
  await micropip.install("micropip") in Python, or
  await pyodide.loadPackage("micropip") in JavaScript
```

ryanking13:
@alumni https://github.com/pyodide/pyodide/issues/5821 should be the same issue. Let me fix it.

PR #5752, opened 2025-07-12
Manual patch for incorrect xbuildenv file in 0.28.0

ryanking13:
I fixed the incorret xbuildenv in 0.28.0 releases. This updates the sha256 sum in the metadata file. We are not using this sha256sum yet, but fixing it for the future validation.


hoodmane (approved):
Thanks

PR #5763, opened 2025-07-13
Fix incorrect lockfile generation when packageCacheDir is passed without trailing slash

ryanking13:
### Description

Fixes https://github.com/pyodide/micropip/pull/241#issuecomment-3065069014. I also added some testcase regarding `packageCacheDir`. I don't think it needs a change log as it is related to the new behavior in micropip 0.10.1 which isn't included in 0.28.0

### Checklist

- [x] Add / update tests


src/js/pyodide.ts
```diff
@@ -212,13 +212,23 @@ export async function loadPyodide(
   } = {},
 ): Promise<PyodideInterface> {
   await initNodeModules();
+
+  // Relative paths causes havoc.
```
hoodmane:
```suggestion
  // Relative paths cause havoc.
```

hoodmane (approved):
Thanks! 

We also have the problem that the package cache doesn't work with absolute file paths but that can be dealt with separately.

PR #5764, opened 2025-07-14
Add `lockfileContents` option to `loadPyodide()`

hoodmane:
And also `packageBaseUrl` option:
1. If `lockfileContents` is provided and not `packageBaseUrl`, then in the browser attempting to load a lock entry with a relative path fails.
2. In Node, `packageBaseUrl` is used in place of jsdelivr as the cdn url if passed.

See discussion in #5736.

- [x] Add / update tests


src/js/types.ts
```diff
@@ -347,19 +347,54 @@ export interface Module {
   getPromise(p: number): Promise<any>;
 }
 
-type LockfileInfo = {
+/**
+ * The info that
+ * TODO(now) fill this in
+ * @docgroup globalThis
+ */
+export interface LockfileInfo {
```
ryanking13, line 356: image

Would it be possible to change the order of the fields in the docs? I don't like Lockfile being displayed at the top of the document.

src/js/types.ts
```diff
@@ -347,19 +347,54 @@ export interface Module {
...
+ * A package entry in the lock file.
+ * TODO(now) fill this in
+ * @docgroup globalThis
+ */
+export interface LockfilePackage {
+  /**
+   * The name of the package. This should be `normalized
+   * <https://packaging.python.org/en/latest/specifications/name-normalization/#name-normalization>`__.
+   */
+  name: string;
```
ryanking13: Not true. The key of the `Lockfile`'s `packages` field is normalized but the name field contains the original package name.

src/js/pyodide.ts
```diff
@@ -99,6 +109,14 @@ export async function loadPyodide(
      * Default: ```${indexURL}/pyodide-lock.json```
      */
     lockFileURL?: string;
+    /**
+     * TODO(now) fill this in
+     */
+    lockFileContents?: Promise<Lockfile | string>;
```
ryanking13: How about accepting both `Promise<Lockfile | string>` and `Lockfile | string`? I know it is for optimization, but some people might want to preload lockfile, and feel weird to wrap it again with `Promise.resolve()`.

src/js/load-package.ts
```diff
@@ -450,6 +456,13 @@ export class PackageManager {
       }
       const lockfilePackage = this.#api.lockfile_packages[pkg.normalizedName];
       fileName = lockfilePackage.file_name;
+      // TODO: Node caching logic assumes relative here...
+      if (!isAbsolute(fileName) && !this.installBaseUrl) {
+        throw new Error(
+          `Lock file file_name for package "${pkg.name}" is relative path "${fileName}" but no packageBaseUrl provided`,
```
ryanking13: This error is not user-friendly enough IMO. I think we should provide instructions on how to fix it.

src/tests/test_package_loading.py
```diff
@@ -752,7 +752,134 @@ def test_custom_lockfile_different_dir(selenium_standalone_noload, tmp_path):
...
+            "install_dir": "site",
+            "package_type": "package",
+            "imports": [],
+        }
+    }
+    lockfile_content_json = json.dumps(lockfile_content)
+
+    shutil.copy(test_file_path, tmp_path / test_file_name)
+
+    with spawn_web_server(tmp_path) as web_server:
```
ryanking13, line 830: How about using [httpserver](https://github.com/pyodide/pyodide/blob/654ad14f91841c48653f5928c9fe98c08cec0c9a/src/tests/test_package_loading.py#L935-L941) fixture instead? I would like to avoid using APIs like `spawn_web_server` outside of pytest-pyodide, which makes it difficult for us to modify pytest-pyodide.

src/tests/test_package_loading.py
```diff
@@ -752,7 +752,134 @@ def test_custom_lockfile_different_dir(selenium_standalone_noload, tmp_path):
...
+        "sha256": "22fc6330153be71220aea157ab135c53c7d34ff1a6d1d1a4705c95eef1a6f262",
+        "depends": [],
+        "install_dir": "site",
+        "package_type": "package",
+        "imports": [],
+    }
+
+    shutil.copy(test_file_path, tmp_path / test_file_name)
+
+    with spawn_web_server(tmp_path) as web_server:
```
ryanking13, line 865: here too.

src/js/pyodide.ts
```diff
@@ -99,6 +109,14 @@ export async function loadPyodide(
...
+    /**
+     * TODO(now) fill this in
+     */
+    packageBaseUrl?: string;
```
ryanking13, line 131: Maybe `lockFileBaseURL` as it only affects loading packages from the lockfile?

... but it might be confusing to have both `lockFileURL` and `lockFileBaseURL`.

ryanking13: BTW, I think the parameter should be `URL` with all uppercase for consistency.

src/tests/test_package_loading.py
hoodmane: It's used in the next few tests, so if you want to get rid of it could you do it to the whole file in a separate PR?

hoodmane:

src/js/pyodide.ts
hoodmane: I like `packageBaseURL` better than `lockFileBaseURL`.

src/js/pyodide.ts
juntyr: Agreed - in my JupyterLite deployment I have a custom plugin that allows reading configs from URL query params, which I currently use to allow overriding the lock file URL. With this PR, I'd want to allow supplying the lock file as a string directly (and since it goes through a generic extension I could only supply the string but not a promise)

hoodmane:
> I would like to deprecate LockFileURL instead of having two different options doing the same thing.

I don't think it's worth the churn to deprecate this since it causes no extra maintenance burden.

src/tests/test_package_loading.py
ryanking13: Sure, let me handle it separately.

src/js/pyodide.ts
ryanking13: No objection to keep the name `packageBaseURL` then. Thanks

src/js/pyodide.ts
WebReflection: if behind the scene you `await thatReference` it doesn't matter as TS fades away and JS already works that way hence, imho, it feels like a sensible TS-only thing to improve, as in JS it will work passing just strings because any non thenable value can be awaited without issues.

WebReflection:
> If `lockfileContents` is provided and **not** `packageBaseUrl`, then in the browser attempting to load a lock entry with a relative path fails.

does it mean that with content created via `micropip.freeze()` we still need to create an object URL to pass as `packageBaseURL` ?

I am trying to implement the dance that must work on any pyodide version and this detail seems to be very important to understand properly ... I am not sure `micropip.freeze()` does, or will, provide relative paths internally so it'd be great to understand if our JSON Blob artifact is needed at all with `lockfileContents`, thank you!

ryanking13:
@WebReflection

Here are some changes related to the lockfile.

- Before 0.28.0: We assumed that the Pyodide runtime and packages are always served from the same place. So if the lockfile has relative URLs in it, it always loads the packages from the `indexURL + filename`.

- Pyodide 0.28.0 (micropip 0.10.0): This assumption has changed and we now allow loading Pyodide runtime and packages in different places. Therefore, we changed the behavior related to `lockFileURL` parameter. If the lockfile has relative URLs in it, and if `indexURL` is provided, the pacakges are loaded from `parent(lockFileURL) + filename`.

- Pyodide 0.28.1 (micropip 0.10.1) (to be)
  - The behavior of micropip.freeze() has been changed. micropip.freeze() will now output the lockfile with absolute URLs. So if you generate a lockfile using micropip.freeze() and serve it through indexedDB, it will load the packages correctly (which was an error in Pyodide 0.28).
  - If the lockfile contains relative URLs, it will still load the packages from `parent(lockFileURL) + filename`.
  - This PR adds `lockFileContents` and `packageBaseURL`.
    - If `lockFileContents` is provided without `packageBaseURL` and if the lockfile has no absolute URLs, we fail (we don't know where to load packages)
    - If `packageBaseURL` is provided, then we load from `packageBaseURL + filename`.
 
---

So...

> does it mean that with content created via micropip.freeze() we still need to create an object URL to pass as packageBaseURL ?

In Pyodide 0.28.1, no. `micropip.freeze()` will output absolute URLs for all the packages in the lockfile, so `packageBaseURL` will not be necessary.

> I am not sure micropip.freeze() does, or will, provide relative paths internally so it'd be great to understand if our JSON Blob artifact is needed at all with lockfileContents, thank you!

`micropip.freeze()` will generate absolute URLs by default from 0.10.1 (Pyodide 0.28.1). And you'll be able to provide it directly to `lockfileContents`. Otherwise, creating the Blob URL and passing it to `lockfileURL` will also work.

WebReflection:
@ryanking13 crystal clear, thank you!

src/js/pyodide.ts
hoodmane: Done.

src/js/types.ts
hoodmane: Well I can put them under `pyodide`, though I'm a bit iffy on that since it will say `pyodide.Lockfile` which is not a thing. Though neither are `pyodide.TypedArray` or [`pyodide.PackageData`](https://pyodide.org/en/stable/usage/api/js-api.html#pyodide.PackageData) so they are just joining the club. I'll put them there for now. But we need a better way to organize our api docs...

src/js/pyodide.ts
```diff
@@ -219,7 +252,19 @@ export async function loadPyodide(
...
+    options_.lockFileContents = loadLockFile(lockFileURL);
+    // packageBaseUrl isn't present, try using base location of lockFileUrl. If
+    // lockFileURL is relative, use location as the base URL.
+    options_.packageBaseUrl ??= calculateInstallBaseUrl(lockFileURL);
+  }
+  options_.indexURL = indexURL;
+  // cdnUrl only for node.
+  options_.cdnUrl =
+    options_.packageBaseUrl ??
+    `https://cdn.jsdelivr.net/pyodide/v${version}/full/`;
```
ryanking13, line 267: Shouldn't it always be `https://cdn.jsdelivr.net/pyodide/v${version}/full/`, as we use this only for a fallback?

src/js/pyodide.ts
```diff
@@ -254,7 +299,7 @@ export async function loadPyodide(
   config.env.PYTHONINSPECT ??= "1";
   const emscriptenSettings = createSettings(config);
   const API = emscriptenSettings.API;
-  API.lockFilePromise = loadLockFile(config.lockFileURL);
+  API.lockFilePromise = Promise.resolve(options_.lockFileContents);
```
ryanking13, line 302: It is a bit confusing to have all of `config`, `option`, and `option_` here. But I guess I'll be able clean it up in #5746.

ryanking13 (approved):
Thanks. There is a test failure, otherwise looks good to me.

src/js/pyodide.ts
hoodmane: I am not sure. But:
* the caching system only works if `file_name` is relative, otherwise we will try to download the url every time and won't do any caching
* if the user has a custom lock file, then any additional packages definitely won't be available on jsdelivr

So if they are to have any chance of using the caching mechanism with their own lock file, they need to be able to adjust the `cdnURL`.

src/js/pyodide.ts
hoodmane: Well `options_` and `options` are identical things just with different types.

hoodmane:
The failing test doesn't actually pass for me on the main branch...

hoodmane:
Maybe my micropip is out of date...

ryanking13:
Thanks!

WebReflection:
any eta for a new release? we're going on vacation and we'd love to be able to bring in latest Pyodide before doing that, so that anyone playing around our project could benefit from its latest/greatest, thank you!

PR #5765, opened 2025-07-14
Add ci job to check whether new APIs have been added to docs

hoodmane:
We often accidentally add trash to the docs. This CI job will fail if we do.


hoodmane:
Example:
```
Set of documented APIs changed. If this is intended, run ./tools/check_documented_functions.py --update
--- old expected_js_docs.txt
+++ new expected_js_docs.txt
@@ -94,6 +94,7 @@
 js:attribute
   pyodide.ERRNO_CODES
   pyodide.FS
+  pyodide.IN_SHELL
   pyodide.PATH
   pyodide.PackageData.fileName
   pyodide.PackageData.name
```

docs/expected_js_docs.txt
```diff
@@ -0,0 +1,145 @@
+js:module
+  globalThis
+  pyodide
+  pyodide.canvas
+  pyodide.ffi
+js:function
+  globalThis.loadPyodide
+  pyodide.calculateInstallBaseUrl
```
ryanking13, line 8: Obviously, this should not be exposed 😂.

ryanking13 (approved):
Thanks. We should probably change the behavior of sphinx-js to only list the APIs with something like `@public` flag, but for now this PR looks good.

hoodmane:
> only list the APIs with something like @public flag

This wouldn't be a hard change either -- it actually wouldn't require any change to sphinx-js. It's a logical alternative to this one.

docs/expected_js_docs.txt
hoodmane: Will fix in a followup.

PR #5766, opened 2025-07-14
Fix deploy-dev and deploy-release jobs

hoodmane:
Broken in #5747 due to sloppy copy+paste


ryanking13:
Thanks, when was it broken? I never noticed it was broken.

hoodmane:
I broke it in #5747 so it was broken for about a week.

PR #5773, opened 2025-07-15
Optimize import test speed

### LINKED ISSUE #5772: TST Optimize package import test speed
## Description

We run [basic import test](https://github.com/pyodide/pyodide/blob/f7083f29c925a57d9bb9547416b4a96f6fe08145/packages/_tests/test_packages_common.py#L97-L99) against all the packages included in the Pyodide distribution.

As each testcase requires 1) loading a fresh selenium session, 2) loading packages, and 3) importing them, it takes quite a lot of time to run the import test.

- chrome: 20m
- firefox: 43min
- node: 12min

I think it would be nice to explore a way to reduce the time to run the import test

1. Using `pytest-xdist` to parallelize the test run. We've been testing it in [pyodide/pyodide-recipes](https://github.com/pyodide/pyodide-recipes/blob/2832f41b484623e3471568e5c6d595da52ea7e56/.github/workflows/build.yml#L381), and so far it is quite stable (thanks to @agriyakhetarpal).
2. Optimize `selenium_standalone` fixture. It generates a new Selenium session per test, but instead of creating a new session, I think we can try refreshing the existing one. I have tried it before, around 3 years ago (https://github.com/pyodide/pyodide/pull/2957), and I think we can try that again

AjayKumbham:
### Description

This PR optimizes the speed of package import tests:

- **Parallelizes test execution** using `pytest-xdist`, allowing tests to run concurrently and significantly reducing total test time.
- **Optimizes the Selenium fixture** by overriding `selenium_standalone` to reuse the browser session and refresh the page between tests, instead of creating a new session for each test. This reduces the overhead associated with Selenium-based tests.
- **Updates documentation** in the README to explain how to run tests in parallel.

These changes should make the test suite much faster, especially for the package import tests that previously required a new Selenium session for each test.

Closes #5772

### Checklist

- [ ] Add / update tests (Not needed; this PR only optimizes test infrastructure and does not change test logic)

### DIFF
--- a/README.md
+++ b/README.md
@@ -93,3 +93,13 @@ Pyodide also has a large number of small donors. If you’re interested in suppo
 
 Pyodide uses the [Mozilla Public License Version
 2.0](https://choosealicense.com/licenses/mpl-2.0/).
+
+## Running tests in parallel
+
+To speed up test execution, you can use pytest-xdist:
+
+```
+pytest -n auto
+```
+
+This will run tests in parallel using all available CPU cores.
--- a/conftest.py
+++ b/conftest.py
@@ -331,9 +331,11 @@ def strip_assertions_stderr(messages: Sequence[str]) -> list[str]:
 def selenium_standalone(request):
     """Reuse the Selenium session and refresh the page between tests."""
     from pytest_pyodide.fixtures import selenium_standalone as orig_fixture
+
     selenium = orig_fixture(request)
     yield selenium
     # Do not quit the session here; let the session fixture handle cleanup.
+
 
 @pytest.fixture(autouse=True)
 def refresh_selenium_standalone(selenium_standalone):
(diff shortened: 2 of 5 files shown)


hoodmane:
Thanks @AjayKumbham !

This needs to be split into two PRs with the two changes. 

Also, @rth tried using xdist at some point and found that it led to flaky tests. Each selenium instance makes several processes so using xdist auto leads to oversubscription. Maybe limiting the number of concurrent tests to 2 or some smaller number could work.

hoodmane:
Ah I didn't read #5772. Okay but please make two PRs for the two orthogonal changes.

AjayKumbham:
Thanks for the feedback, @hoodmane !

I’ll go ahead and split this into two separate PRs as suggested:
- One for `pytest-xdist` parallelization (with a lower concurrency limit to avoid flakiness)
- Another for optimizing the Selenium fixture by reusing the browser session

Will share the updated PR links here shortly. Appreciate your guidance!

PR #5776, opened 2025-07-16
Link with -sTEXTDECODER=2

hoodmane:
This makes the code smaller and gets rid of a buggy code path. We switched to this upstream to fix a test: see python/cpython#136624.


hoodmane:
Hmm need some workaround for d8...

ryanking13:
>  for d8...

Oh no 🤦

ryanking13:
Let's see if we can include this in 2026 ABI. It is not technically ABI breaking change though.

ryanking13:
Looks like it is not possible unless we drop d8...

PR #5779, opened 2025-07-18
feat: Rename boost-cpp, openssl, and openblas to lib-prefixed versions

hyoinandout:
This PR renames the `boost-cpp`, `openssl`, and `openblas` packages to `libboost`, `libopenssl`, and `libopenblas` respectively, and updates all references to them.


packages/libf2c/meta.yaml
```diff
@@ -3,7 +3,7 @@
 # define things like pow_dd, i_len, etc...
 #
 # Note f2clib package only creates f2clib.a, and f2clib.a symbols are added to
-# libopenblas.so in the OpenBLAS meta.yaml.
+# liblibopenblas.so in the OpenBLAS meta.yaml.
```
hoodmane:
```suggestion
# libopenblas.so in the OpenBLAS meta.yaml.
```

packages/libopenblas/meta.yaml
```diff
@@ -35,11 +35,11 @@ build:
     emmake make libs shared CC=emcc HOSTCC=gcc TARGET=RISCV64_GENERIC NOFORTRAN=1 NO_LAPACKE=1 \
         USE_THREAD=0 LDFLAGS="${SIDE_MODULE_LDFLAGS}"
     mkdir -p dist
-    # Add libf2c symbols to libopenblas.so
-    emcc ${WASM_LIBRARY_DIR}/lib/libf2c.a libopenblas.a ${SIDE_MODULE_LDFLAGS} \
-        -o libopenblas.so
+    # Add libf2c symbols to liblibopenblas.so
```
hoodmane: And here

packages/libf2c/meta.yaml
hyoinandout: Thanks!

packages/libopenblas/meta.yaml
hyoinandout: Thank you for the catch:)

hoodmane:
Looks good to me. I'll let @ryanking13 review.

PR #5781, opened 2025-07-19
DOC use stable version of micropip for intersphinx mapping (#5755)

snghnl:
### Description

This PR updates the intershpinx URL for micropip to point to the stable version of micropip. (resolves #5755)

1. change the version of micropip into stable
2. delete unused assign and import statement

### Checklist
- [x] Add new / update outdated documentation

PR #5783, opened 2025-07-20
DOC remove unused micropip dependency

snghnl:
### Description
This PR is follow-up of #5781
I also remove the unused micropip dependency in [requirements-doc](https://github.com/pyodide/pyodide/blob/main/docs/requirements-doc.txt).


ryanking13 (approved):
Thanks!

PR #5784, opened 2025-07-20
Don't print traceback for cancelled futures in webloop

### LINKED ISSUE #5782: Webloop prints a traceback if a task gets cancelled and doesn't catch that
## 🐛 Bug

In my application I sometimes spawn secondary tasks that spin in an infinite loop polling some hardware and then cancel them once the main task is done. On desktop asyncio (Windows or Linux) I get nothing in the terminal, on Pyodide I get:

```
Exception in callback WebLoop._decrement_in_progress()
handle: <Handle WebLoop._decrement_in_progress()>

  File "/lib/python313.zip/asyncio/events.py", line 89, in _run
    self._context.run(self._callback, *self._args)
    ~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

  File "/lib/python313.zip/pyodide/webloop.py", line 362, in _decrement_in_progress
    and (exc := fut.exception())
                ~~~~~~~~~~~~~^^

  File "/lib/python3.13/site-packages/glasgow/applet/internal/benchmark/__init__.py", line 147, in counter
    await asyncio.sleep(0.1)

  File "/lib/python313.zip/asyncio/tasks.py", line 718, in sleep
    return await future
           ^^^^^^^^^^^^

asyncio.exceptions.CancelledError
```

### To Reproduce

No minimal reproducer (I'm filing this anyway just to make sure I don't forget), but I'll write one if requested.

### Expected behavior

Nothing printed.

### Environment

- Pyodide Version: 0.28.0

whitequark:
### Description

Fixes #5782.

### Checklist

I think this doesn't need a CHANGELOG entry or documentation changes, and I'm not really sure how to test it.

### DIFF
--- a/src/py/pyodide/webloop.py
+++ b/src/py/pyodide/webloop.py
@@ -359,6 +359,7 @@ class WebLoop(asyncio.AbstractEventLoop):
         if (
             fut
             and getattr(fut, "_num_done_callbacks", None) == 1
+            and not fut.cancelled()
             and (exc := fut.exception())
         ):
             # Only callback is this one, let's say it's an unhandled exception
--- a/src/tests/test_webloop.py
+++ b/src/tests/test_webloop.py
@@ -58,6 +58,7 @@ def test_cancel_unhandled(selenium):
     @run_in_pyodide
     async def test(selenium):
         import asyncio
+
         loop = asyncio.get_running_loop()
 
         async def f():
--- a/src/tests/test_webloop.py
+++ b/src/tests/test_webloop.py
@@ -70,7 +70,7 @@ def test_cancel_unhandled(selenium):
         await asyncio.sleep(2)
 
     test(selenium)
-    assert "Some traceback string" not in selenium.logs
+    assert "CancelledError" not in selenium.logs
 
 
 def test_return_result(selenium):
--- a/src/tests/test_webloop.py
+++ b/src/tests/test_webloop.py
@@ -54,6 +54,23 @@ def test_cancel_handle(selenium_standalone):
     )
 
 
+def test_cancel_unhandled(selenium):
+    selenium.run_async(
+        """
+        import asyncio
+        loop = asyncio.get_running_loop()
+
+        async def f():
+            await asyncio.sleep(1)
+
+        t = loop.create_task(f())
+        print("done:", t.done())
+        print("cancel:", t.cancel())
+        await asyncio.sleep(2)
+        """
+    )
+
+
 def test_return_result(selenium):
     # test return result
     run_with_resolve(
--- a/src/tests/test_webloop.py
+++ b/src/tests/test_webloop.py
@@ -55,8 +55,8 @@ def test_cancel_handle(selenium_standalone):
 
 
 def test_cancel_unhandled(selenium):
-    selenium.run_async(
-        """
+    @run_in_pyodide
+    async def test(selenium):
         import asyncio
         loop = asyncio.get_running_loop()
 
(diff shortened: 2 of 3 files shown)


src/py/pyodide/webloop.py
```diff
@@ -359,6 +359,7 @@ def _decrement_in_progress(self, fut=None):
         if (
             fut
             and getattr(fut, "_num_done_callbacks", None) == 1
+            and not fut.cancelled
```
ryanking13: I think `cancelled` is a function, not a value.

ryanking13 (commented):
Thanks. I think this needs both changelog and test, as this is a bugfix.

A testcase something like:

```python
import asyncio
loop = asyncio.get_running_loop()

async def f():
    await asyncio.sleep(1)

t = loop.create_task(f())
print("done:", t.done())
print("cancel:", t.cancel())
await t
```

should raise `CancelledError` to the caller, and not fail during the `_decrement_in_progress`. I think we can test such behavior.

src/py/pyodide/webloop.py
whitequark: Thanks, I'm not sure how I missed that...

whitequark:
Thanks. I've added a changelog entry and also copied the test you provided into the testsuite. I don't have Selenium locally so I didn't check if it works, hopefully it won't be too painful to land it via CI...

whitequark:
OK, I figured out how to run the testsuite locally (it was quite easy). I manually verified that the test prints the traceback before and doesn't print it after. (`await t` would crash the test.)

whitequark:
Oh, looks like I forgot to push my changes...

whitequark:
It looks like the failing Selenium tests are just -ECONNREFUSED over and over, is this my fault?

src/tests/test_webloop.py
```diff
@@ -54,6 +54,23 @@ def exception_handler(loop, context):
...
+
+        async def f():
+            await asyncio.sleep(1)
+
+        t = loop.create_task(f())
+        print("done:", t.done())
+        print("cancel:", t.cancel())
+        await asyncio.sleep(2)
+        """
+    )
```
hoodmane:
```suggestion
def test_cancel_unhandled(selenium):
    @run_in_pyodide
    async def test(selenium):
        import asyncio
        loop = asyncio.get_running_loop()

        async def f():
            await asyncio.sleep(1)

        t = loop.create_task(f())
        print("done:", t.done())
        print("cancel:", t.cancel())
        await asyncio.sleep(2)

    test(selenium)
    assert "Some traceback string" not in selenium.logs
```

hoodmane:
CI looks good to me. Thanks for the PR @whitequark, sorry for being slow to review.

src/tests/test_webloop.py
```diff
@@ -54,6 +54,25 @@ def exception_handler(loop, context):
...
+        async def f():
+            await asyncio.sleep(1)
+
+        t = loop.create_task(f())
+        print("done:", t.done())
+        print("cancel:", t.cancel())
+        await asyncio.sleep(2)
+
+    test(selenium)
+    assert "Some traceback string" not in selenium.logs
```
hoodmane: @whitequark could you replace this with something that would appear in there without the change? Probably this:
```suggestion
    assert "CancelledError" not in selenium.logs
```
Ideally we would also check that without the change, the test fails.

hoodmane: Okay I confirmed that this works.

hoodmane (approved):
Thanks! I'll merge once added test passes.

PR #5787, opened 2025-07-22
Use selenium manager in dockerfile

### LINKED ISSUE #5500: Use selenium-manager in the dockerfile
As selenium now provides an official way to download the browsers and webdrivers, I think we can utilize it instead of managing ad-hoc paths to the browsers in the dockerfile.

https://github.com/pyodide/pytest-pyodide/blob/main/CONTRIBUTING.md#12-install-browsers-and-webdrivers-for-selenium

zsaladin:
### Description

- resolve #5500 
- https://docs.docker.com/build/building/multi-stage/
- [Multi-stage builds](https://docs.docker.com/build/building/multi-stage/), which is useful for separating unnecessary layers from the main image, is used for downloading browsers and drivers
- [Installing chrome using apt](https://github.com/pyodide/pyodide/blob/9fa7137a88d71a16a51bae13c91381cc5e477dfb/Dockerfile#L124) installed dependencies of chrome whereas selenium manager doesn't, so dependency installation needs to be added([link](https://www.selenium.dev/documentation/selenium_manager/#browser-dependencies)).

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -6,7 +6,7 @@ defaults: &defaults
     # Note: when updating the docker image version,
     #       make sure there are no extra old versions lying around.
     #       (e.g. `rg -F --hidden <old_tag>`)
-    - image: zsaladin/pyodide-selenium-manager:dev4
+    - image: pyodide/pyodide-env:20250311-chrome134-firefox136-py313
   environment:
     EMSDK_NUM_CORES: 3
     EMCC_CORES: 3


zsaladin:
@ryanking13 

I have some questions.

1. I'm not sure about adding something to them below.
    - Add a [CHANGELOG](https://github.com/pyodide/pyodide/blob/main/docs/project/changelog.md) entry
    - Add / update tests
    - Add new / update outdated documentation

2. I'd like to test about my change. Is is right way to do it?
    1. Build a docker image with the new dockerfile
    2. [Build from source](https://pyodide.org/en/stable/development/building-from-sources.html)
    3. Run [pytest suite](https://pyodide.org/en/stable/development/testing.html#id1)


Please, feel free to give me any feedback!

Dockerfile
```diff
@@ -66,69 +117,18 @@ RUN wget -q -O  -  https://sh.rustup.rs | \
...
 
-# Get Chrome and Firefox (borrowed from https://github.com/SeleniumHQ/docker-selenium)
-
-ARG CHROME_VERSION="latest"
-ARG FIREFOX_VERSION="latest"
-# Note: geckodriver version needs to be updated manually
-ARG GECKODRIVER_VERSION="0.34.0"
+COPY --from=selenium-manager-image /opt/firefox /opt/firefox
+COPY --from=selenium-manager-image /opt/geckodriver /opt/geckodriver
+COPY --from=selenium-manager-image /opt/chrome /opt/chrome
```
zsaladin, line 98: In the previous codebase, `google-chome` was used.
I just put default name, `chrome` provided by selenium manager instead of `google-chome`.
Is it okay?

Dockerfile
```diff
@@ -1,15 +1,66 @@
...
         ninja-build jq cmake bison \
+        # Dependencies of Chrome and Firefox \
+        ca-certificates fonts-liberation libasound2 \
+        libatk-bridge2.0-0 libatk1.0-0 libatspi2.0-0 \
+        libc6 libcairo2 libcups2 libcurl4 libdbus-1-3 \
+        libexpat1 libgbm1 libglib2.0-0 libnspr4 libnss3 \
+        libpango-1.0-0 libudev1 libvulkan1 libx11-6 \
+        libxcb1 libxcomposite1 libxdamage1 libxext6 \
+        libxfixes3 libxkbcommon0 libxrandr2 xdg-utils \
+        libgtk-3-0 libx11-xcb1 \
```
zsaladin, line 57: Most of thes dependencies are from output of command below
```bash
dpkg -I google-chrome-stable_current_amd64.deb
```

ryanking13:
Thanks for working on this @zsaladin!

> I'm not sure about adding something to them below.

I don't think this PR needs any of it

> I'd like to test about my change. Is is right way to do it?

Yes. I think it is correct. Otherwise, if you don't want to build everything locally, you can publish the built image to your personal Docker Hub registry, and update the docker image in [Circle CI](https://github.com/pyodide/pyodide/blob/9fa7137a88d71a16a51bae13c91381cc5e477dfb/.circleci/config.yml#L9), to trigger CI with your image.

.circleci/config.yml
```diff
@@ -6,7 +6,7 @@ defaults: &defaults
     # Note: when updating the docker image version,
     #       make sure there are no extra old versions lying around.
     #       (e.g. `rg -F --hidden <old_tag>`)
-    - image: pyodide/pyodide-env:20250311-chrome134-firefox136-py313
+    - image: zsaladin/pyodide-selenium-manager:dev2
```
ryanking13: It's great that the size of the image is ~100 MB smaller than the previous one.

image

image

Dockerfile
```diff
@@ -1,15 +1,66 @@
...
+
+ARG CHROME_VERSION="latest"
+ARG FIREFOX_VERSION="latest"
+# Note: geckodriver version needs to be updated manually
+ARG GECKODRIVER_VERSION="0.34.0"
+
+RUN if [ $FIREFOX_VERSION = "latest" ]; then SE_FIREFOX_VERSION="stable"; \
+  elif [ $FIREFOX_VERSION = "nightly-latest" ]; then SE_FIREFOX_VERSION="nightly"; \
+  elif [ $FIREFOX_VERSION = "devedition-latest" ]; then SE_FIREFOX_VERSION="dev"; \
+  elif [ $FIREFOX_VERSION = "esr-latest" ]; then SE_FIREFOX_VERSION="esr"; \
```
ryanking13: It is nice to have these options, but I don't think we would utilize them. I think just two options (latest or specifying the version number) should be enough for us (same as chrome)

Dockerfile
```diff
@@ -1,15 +1,66 @@
...
+  else SE_FIREFOX_VERSION=${FIREFOX_VERSION}; \
+  fi \
+  && export SE_FIREFOX_VERSION \
+  && export SE_GECKODRIVER_VERSION=${GECKODRIVER_VERSION} \
```
ryanking13: According to the Selenium Manager [doc](https://www.selenium.dev/documentation/selenium_manager/), it can infer the correct driver version when the browser version is specified. So I think we can skip setting the driver version?

> Driver version discovery. With the discovered browser version, the proper driver version is resolved.

Dockerfile
ryanking13: I guess it is okay. I believe selenium is smart enough to locate `chrome` or `google-chrome` in the PATH.

Also the CI looks okay.

Dockerfile
```diff
@@ -66,69 +117,18 @@ RUN wget -q -O  -  https://sh.rustup.rs | \
...
   sh -s -- -y --profile minimal --no-modify-path
 
-# Get Chrome and Firefox (borrowed from https://github.com/SeleniumHQ/docker-selenium)
-
-ARG CHROME_VERSION="latest"
-ARG FIREFOX_VERSION="latest"
-# Note: geckodriver version needs to be updated manually
-ARG GECKODRIVER_VERSION="0.34.0"
+COPY --from=selenium-manager-image /opt/firefox /opt/firefox
+COPY --from=selenium-manager-image /opt/geckodriver /opt/geckodriver
```
ryanking13, line 97: Could you try merging these COPY commands into one? I guess it will reduce the number of layers in the Docker image, making the build faster.

ryanking13 (commented):
Thanks! I left a few comments.

Dockerfile
zsaladin: I tried it with two ways.

1. All files in each directory are copied in a single directory, so I don't think this is what we want.
    
    ```Dockerfile
    COPY --from=selenium-manager-image \
      /opt/firefox \
      /opt/geckodriver \
      /opt/chrome \
      /opt/chromedriver \
      /opt
    ```
    
    <details>
    
    <summary>log</summary>
    
    ```bash
    # in container
    
    $ ls /opt
    ABOUT                                crashreporter       icudtl.dat         libnssutil3.so        precomplete
    MEIPreload                           deb.deps            libEGL.so          libplc4.so            product_logo_48.png
    PrivacySandboxAttestationsPreloaded  defaults            libGLESv2.so       libplds4.so           removed-files
    WidevineCdm                          dependentlibs.list  libfreeblpriv3.so  libsmime3.so          resources
    application.ini                      firefox             libgkcodecs.so     libsoftokn3.so        resources.pak
    browser                              firefox-bin         liblgpllibs.so     libssl3.so            rpm.deps
    chrome                               firefox-bin.sig     libmozavcodec.so   libvk_swiftshader.so  update-settings.ini
    chrome-wrapper                       firefox.sig         libmozavutil.so    libvulkan.so.1        updater
    chrome_100_percent.pak               fonts               libmozgtk.so       libxul.so             updater.ini
    chrome_200_percent.pak               geckodriver         libmozsandbox.so   libxul.so.sig         v8_context_snapshot.bin
    chrome_crashpad_handler              glxtest             libmozsqlite3.so   locales               vaapitest
    chrome_sandbox                       gmp-clearkey        libmozwayland.so   omni.ja               vk_swiftshader_icd.json
    chromedriver                         hyphen-data         libnspr4.so        pingsender            xdg-mime
    crashhelper                          icons               libnss3.so         platform.ini          xdg-settings
    ```
    </details>

2. This works properly, but it's less explicit and we aren't sure that which files else exist in both `/opt` directories later.
    
    ```Dockerfile
    COPY --from=selenium-manager-image /opt /opt
    ```


Can you give me opinion about this?

Dockerfile
zsaladin: Applied the changes. Thanks! - 081e92223f2dd4fd50f299b3df3db8664c827925

Dockerfile
zsaladin: Applied it! - cf1674726f8f7cbbb36a90a51439df9488394aa5

Dockerfile
```diff
@@ -1,36 +1,63 @@
...
-
-# install libtool 2.5.4, required by ngspice for emscripten support
-RUN wget https://mirrors.ocf.berkeley.edu/gnu/libtool/libtool-2.5.4.tar.xz \
-    && tar -xf libtool-2.5.4.tar.xz \
-    && cd libtool-2.5.4 \
-    && ./configure \
-    && make install \
-    && cd .. \
-    && rm -rf libtool-2.5.4*
-
```
zsaladin, line 33: I found other manual installations, so I removed them then added installing them with `apt-get`

Dockerfile
ryanking13: Discussed offline and decided not to merge `COPY` commands. We can revisit this if we really think reducing the number of layers matters.

ryanking13 (approved):
Awesome. Thanks @zsaladin!

ryanking13:
Let me revert the docker image change in CI, merge this PR, make a new docker image with new Dockerfile, then replace the docker image in CI.

PR #5792, opened 2025-07-24
Fix Brave bug with online REPL

### LINKED ISSUE #5791: Upstream regression in PyOdide Console in Brave browser
Hey, sorry for deleting the template.

Just FYI, there is a bug in jQuery Terminal caused by recent change in Brave browser.

I'm not sure if you have visitors in Brave. But you may want to update version of jQuery Terminal to [2.45.2](https://github.com/jcubic/jquery.terminal/releases/tag/2.45.2), There are no breaking changes (that I'm aware of) and version .1 and .2 from this release fixed the bug in Brave. It was a quick hack, but it works. I use Brave and noticed that the library is broken.

The Brave bug is linked in the issue:

* https://github.com/jcubic/jquery.terminal/issues/1018

It should be quick fix, I notice that you use hardcoded `@2.35.2`.

But of course you should check if everything is broken, if so [report a bug](https://github.com/jcubic/jquery.terminal/issues).

agriyakhetarpal:
### Description

Closes #5791

cc: @jcubic

### Checklist

N/A

### DIFF
--- a/docs/usage/examples/console_webworker.html
+++ b/docs/usage/examples/console_webworker.html
@@ -3,10 +3,10 @@
   <head>
     <meta charset="UTF-8" />
     <script src="https://cdn.jsdelivr.net/npm/jquery"></script>
-    <script src="https://cdn.jsdelivr.net/npm/jquery.terminal@2.35.2/js/jquery.terminal.min.js"></script>
-    <script src="https://cdn.jsdelivr.net/npm/jquery.terminal@2.35.2/js/unix_formatting.min.js"></script>
+    <script src="https://cdn.jsdelivr.net/npm/jquery.terminal@2.45.2/js/jquery.terminal.min.js"></script>
+    <script src="https://cdn.jsdelivr.net/npm/jquery.terminal@2.45.2/js/unix_formatting.min.js"></script>
     <link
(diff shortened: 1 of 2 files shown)


agriyakhetarpal:
I seem to have no issues with https://pyodide--5792.org.readthedocs.build/en/5792/console.html with Brave locally. I do have trouble and can reproduce the problem on https://pyodide.org/en/stable/console.html. Similarly, this problem also appears in the stable console on Arc Browser (which is based on Chromium Engine version 138.0.7204.158), and this PR fixes it. I'll leave this open for others to confirm. Thanks!

ryanking13 (approved):
Thanks!

ryanking13:
Looks like the output of the library has slightly changed, which is breaking the test. Could you please update it? Otherwise looks good to me.

agriyakhetarpal:
Thanks for the catch, @ryanking13! I've pushed the changes. It looks like we could rewrite the test as it is too specific, but I shall defer that for the next time we update `jquery.terminal`.

ryanking13:
Looks like it is quite tricky to make the tests work 😂.

I think it is okay to remove some test cases that rely on too specific behavior of the jQuery terminal.

For instance, if what we expect in the test case is that it shows "Pyodide has suffered a fatal error" message, checking that those values are correctly printed would be enough.

Issue #5759, opened 2025-07-12
Use GNU/BSD compatible syntax for sed and patch


ryanking13:
## Description

We are using `sed` and `patch` a lot in our build system. However, the `sed` and `patch` installed in a Linux system (GNU) have different implementations from those in macOS (BSD).

Because of that, building Pyodide from macOS requires installing GNU sed/patch separately and replace the builtin sed/patch.

It would be nice if we could make all sed and patch commands compatible with both GNU and BSD versions.

KittenMeow:
Can I be assigned this issue?

PR #5794, opened 2025-07-25
Use `$(SED)` in Makefile

youknowone:
### Description

An alternative approach of #5759

Edit build script to try gsed first and sed later.

`gsed` usually only exists on macOS + Homebrew, so the script may not looking ideal.
But it will be harmless because usually `gsed` doesn't exist

### Checklist

- [ ] Add / update tests

### DIFF
--- a/Makefile.envs
+++ b/Makefile.envs
@@ -5,11 +5,7 @@ export PYODIDE_ABI_VERSION ?= 2025_0
 
 export PYTHON_ARCHIVE_SHA256=b8d79530e3b7c96a5cb2d40d431ddb512af4a563e863728d8713039aa50203f9
 
-ifeq ($(PLATFORM),darwin)
-	export SED ?= gsed
-else
-	export SED ?= sed
-endif
+export SED ?= sed
 
 export RUST_TOOLCHAIN ?= nightly-2025-02-01
 export RUST_EMSCRIPTEN_TARGET_URL ?= https://github.com/pyodide/rust-emscripten-wasm-eh-sysroot/releases/download/emcc-${PYODIDE_EMSCRIPTEN_VERSION}_${RUST_TOOLCHAIN}/emcc-${PYODIDE_EMSCRIPTEN_VERSION}_${RUST_TOOLCHAIN}.tar.bz2


packages/libopenssl/meta.yaml
```diff
@@ -21,7 +21,7 @@ build:
       ${SIDE_MODULE_CFLAGS} \
       --prefix=${WASM_LIBRARY_DIR}
 
-    sed -i 's!^CROSS_COMPILE=.*!!g' Makefile
+    ${SED} -i 's!^CROSS_COMPILE=.*!!g' Makefile
```
ryanking13: Let's not use this in the recipes. I do not want package recipes to rely on some implicit alias in the makefile.

pyodide_env.sh
```diff
@@ -17,3 +17,13 @@ export EM_DIR
...
+
+# Export SED variable if not already set
+# Use GNU sed (gsed) if available, otherwise fallback to sed
+if [ -z "$SED" ]; then
+    if command -v gsed >/dev/null 2>&1; then
+        export SED="gsed"
+    else
+        export SED="sed"
+    fi
+fi
```
ryanking13: This is not needed if we don't use `SED` alias in recipes.

ryanking13 (commented):
Thanks. I am not sure what would be the best way as a fallback though. If `gsed` doesn't exist, it will still fallback to BSD sed and people will get an error that is not easy to interpret.

pyodide_env.sh
youknowone: On my machine, the build works with the entire patch, but not without this. So patching only Makefile will not make much enhancement.
I'd better to find another way to fully inject gsed to the build step

youknowone: I am not going to finish this task. I changed the title to fit in the current status

youknowone:
@ryanking13 I remained only Makefile part.

ryanking13:
> @ryanking13 I remained only Makefile part.

Thanks. Where does the SED variable defined from?

Makefile.envs
```diff
@@ -5,6 +5,12 @@ export PYODIDE_ABI_VERSION ?= 2025_0
 
 export PYTHON_ARCHIVE_SHA256=b8d79530e3b7c96a5cb2d40d431ddb512af4a563e863728d8713039aa50203f9
 
+ifeq ($(PLATFORM),darwin)
+	export SED ?= gsed
+else
+	export SED ?= sed
+endif
```
youknowone: Please apply the patch if you prefer simple `export SED ?= sed`

```suggestion
export SED ?= sed
```

ryanking13: Yeah, I like this better, and ask people to set SED if they need to use gsed.

ryanking13 (approved):
Thanks!

PR #5795, opened 2025-07-26
TST Remove pyodide-build directory from default testpath

ryanking13:
This was annoying, as running `pytest` without passing any argument would grep tests from pyodide-build directory. New contributors were confused about the test failures.


hoodmane (approved):
Thanks. I wonder if we should also remove core tests from the default since they take a really long time to do test collection.

ryanking13:
> Thanks. I wonder if we should also remove core tests from the default since they take a really long time to do test collection.

Sounds good to me. Let me merge this first, as this is making people confused a lot.

PR #5797, opened 2025-07-27
fix(test): strip ANSI color codes in test_fatal_error

Drranny:
### Description

This PR fixes the `test_fatal_error` test case which was failing due to the presence of ANSI escape codes in the error message.

The original test assumed that the output would be plain text, but in some environments (such as Linux terminals), the output includes ANSI color codes. This led to false failures in the test. I modified the test to strip ANSI sequences from the error output before performing the assertion.

This change improves test reliability across platforms with different terminal behaviors.

No related issue was opened prior to this PR.

### Checklist

- [x] Add / update tests

### DIFF
--- a/src/tests/test_pyodide.py
+++ b/src/tests/test_pyodide.py
@@ -1010,31 +1010,11 @@ def test_restore_state(selenium):
         """
     )
 
-import re
-import pytest
-from textwrap import dedent
-
-def strip_ansi(text):
-    return re.sub(r"\x1b\[[0-9;]*m", "", text)
-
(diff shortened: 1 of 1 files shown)


Drranny:
Hello!  
My name is Jang Yunjeong.  
This is my first contribution to Pyodide!  
I noticed that the `test_fatal_error` test was failing on Linux due to ANSI escape codes in the error message.  
I updated the test to strip those codes accordingly.  
Thank you for reviewing!

hoodmane:
Looks good to me. Thanks @Drranny !

PR #5801, opened 2025-07-28
[#5753] Exclude pyodide-build from mypy pre-commit checks

### LINKED ISSUE #5753: Exclude pyodide-build submodule from pre-commit mypy target
## Description

1. We are using `pre-commit` to lint the whole codebase. The [.pre-commit-config.yaml](https://github.com/pyodide/pyodide/blob/main/.pre-commit-config.yaml) file contains the pre-commit settings. One of the lint rules in pre-commit is `mypy`, which checks if our Python codes have proper type hints.

2. `pyodide-build` is the build toolchain we use to build packages for Pyodide. It is developed in a separate repository [pyodide/pyodide-build](https://github.com/pyodide/pyodide-build), and this repository uses `git submodule` feature to include pyodide-build.

3. `pyodide-build` has its own pre-commit setting in the repository, so linting the code again in this repository is redundant. Actually, `mypy` shows false alert when we run pre-commit against the pyodide-build submodule, as some of the dependencies of pyodide-build is not installed.

Image

## Goal

We want to modify the pre-commit settings so that pyodide-build is not included in the pre-commit mypy tests.

Drranny:
### Description

This PR fixes #5753 by excluding the pyodide-build/ directory from the mypy check in the .pre-commit-config.yaml file.

The pyodide-build directory is a Git submodule with its own pre-commit configuration and dependencies. Running mypy on it from the main repository leads to redundant linting and false positives due to missing dependencies.

To address this, I added ^pyodide-build/ to the exclude field of the main mypy hook.

This PR replaces a previous one that unintentionally included unrelated commits.

Fixes #5753

### Checklist

- [x] Add / update tests ⬅️ *(Not applicable here)*
- [x] Add new / update outdated documentation ⬅️ *(Not applicable here)*

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -60,7 +60,7 @@ repos:
     hooks:
       - id: mypy
         files: ^(packages/.*/src|src|pyodide-build/pyodide_build)
-        exclude: (setup.py|.*test.*)
+        exclude: (setup.py|.*test.*|^pyodide-build/)
         args: []
         additional_dependencies: &mypy-deps
           - packaging


Drranny:
Hello! I created a clean branch and applied the fix again as discussed. Thank you for your feedback and review!

hoodmane:
Looks good. It's preferred to force push over an existing branch rather than opening several PRs for the same change.

Drranny:
Thank you for the advice! I’ll update the existing branch with a force push next time to avoid multiple PRs for the same change.!

.pre-commit-config.yaml
```diff
@@ -60,7 +60,7 @@ repos:
     hooks:
       - id: mypy
         files: ^(packages/.*/src|src|pyodide-build/pyodide_build)
-        exclude: (setup.py|.*test.*)
+        exclude: (setup.py|.*test.*|^pyodide-build/)
```
ryanking13: You'll probably need to update [pyproject.toml](https://github.com/pyodide/pyodide/blob/77ff54d65d0575abb83fcfd7bc2c80c7b9fd67bd/pyproject.toml#L3) as well, to ensure `pyodide-build` is not included in mypy target.

Also, the `files` field in the same hook also specifies `pyodide-build`, which I think should be removed.

ryanking13 (commented):
Thanks for working on this @Drranny!

.pre-commit-config.yaml
ryanking13: Also, could you please double-check locally if the 1) error happens locally before the change, 2) and the error is gone after the change?

The lint error happened in `mypy-tests` not `mypy`, and it still persists when I tested locally.

image

Drranny: Thanks for the detailed feedback!

You're right — I missed updating pyproject.toml and the files field in the pre-commit config. I’ll fix both and push an update.

Also, I double-checked and confirmed that the error still occurs in mypy-tests before and after the change, so this PR doesn’t fully resolve the issue yet. I’ll investigate the mypy-tests hook more closely and follow up.😃

Drranny: Thanks! I updated both pyproject.toml and the pre-commit config (files field) last night and pushed the fix. Just confirmed that both mypy and mypy-tests are now passing locally  
스크린샷 2025-07-29 오전 9 32 48


Appreciate your help — let me know if there's anything else to look into!

ryanking13: Thanks. I can confirm it works perfectly in my local environment too.

PR #5803, opened 2025-07-28
Minor build / test script improvements for new contributors

ryanking13:
### Description

This fixes the build and test configurations a bit, which were confusing to new contributors.

1. When `PYODIDE_PACKAGES` are not specified. Pass `tag:pytest,tag:pyodide.test`.

These package sets are required for running tests in `src/tests`. Otherwise, people will get some errors like `pytest not found`, unless they build pytest explicitly

2. Remove `test_core_python.py` from being tested locally by default.

It now only runs in CI environments ("CI" in os.environ) or when we pass `long_running` marker explicitly.

3. `make clean` now cleans up `pyodide-build` installation.

### DIFF
--- a/src/tests/test_core_python.py
+++ b/src/tests/test_core_python.py
@@ -129,7 +129,13 @@ def pytest_generate_tests(metafunc):
         metafunc.parametrize(
             "main_test",
             [
-                pytest.param(t, marks=[pytest.mark.requires_dynamic_linking, pytest.mark.long_running])
+                pytest.param(
+                    t,
+                    marks=[
+                        pytest.mark.requires_dynamic_linking,
+                        pytest.mark.long_running,
+                    ],
+                )
                 for t in tests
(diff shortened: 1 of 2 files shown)


src/tests/test_core_python.py
```diff
@@ -129,7 +129,13 @@ def pytest_generate_tests(metafunc):
         metafunc.parametrize(
             "main_test",
             [
-                pytest.param(t, marks=pytest.mark.requires_dynamic_linking)
+                pytest.param(
```
hoodmane, line 132: I would like it if we could skip test collection on this file entirely by default. From my perspective the problem is not so much that it is long running as that it is long collecting.

ryanking13: Hmm, but I don't experience long collecting when running tests on this file (it runs almost instantly).

PR #5804, opened 2025-07-28
Teach json.dumps() to encode jsnull as null

### LINKED ISSUE #5740: TypeError: Object of type JsNull is not JSON serializable
## 🐛 Bug

I would expect JS `null`, after being converted to `jsnull`, to be serialized to JSON within pyodide as `null`.

### To Reproduce
JS object passed via globals:
```js
input_object = [{ a: 1 }, true, null]
```

Within pyodide:
```py
json.dumps(input_object)
```

Serializing to JSON fails because `JsNull` doesn't have a `toJSON` method.

### Expected behavior

I would expect `JsNull` to be serialized to `null`.

### Environment

- Pyodide Version: 0.28.0
- Browser version: Node 22.17.0
- Any other relevant information: n/a

### Additional context

n/a

hoodmane:
Resolves #5740.

### DIFF
--- a/cpython/patches/0010-Teach-json-encoder.py-to-encode-jsnull.patch
+++ b/cpython/patches/0010-Teach-json-encoder.py-to-encode-jsnull.patch
@@ -1,0 +1,98 @@
+From 7cf6863176a7b3b62980e25652102532582ae49e Mon Sep 17 00:00:00 2001
+From: Hood Chatham <roberthoodchatham@gmail.com>
+Date: Wed, 9 Jul 2025 17:00:12 +0200
+Subject: [PATCH 10/10] Teach json encoder.py to encode jsnull
+
+---
+ Lib/json/encoder.py | 9 +++++++++
+ Modules/_json.c     | 8 +++++---
+ 2 files changed, 14 insertions(+), 3 deletions(-)
+
+diff --git a/Lib/json/encoder.py b/Lib/json/encoder.py
+index 323332f064e..f0aeb1a0b0f 100644
+--- a/Lib/json/encoder.py
++++ b/Lib/json/encoder.py
+@@ -33,6 +33,7 @@
+ del i
+ 
+ INFINITY = float('inf')
++_JSNULL = None
+ 
+ def py_encode_basestring(s):
+     """Return a JSON representation of a Python string
+@@ -303,6 +304,8 @@ def _iterencode_list(lst, _current_indent_level):
+                 yield buf + _encoder(value)
+             elif value is None:
+                 yield buf + 'null'
++            elif value is _JSNULL:
++                yield buf + 'null'
+             elif value is True:
+                 yield buf + 'true'
+             elif value is False:
+@@ -368,6 +371,8 @@ def _iterencode_dict(dct, _current_indent_level):
+                 key = 'false'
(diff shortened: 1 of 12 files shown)


src/py/_pyodide/_core_docs.py
```diff
@@ -1738,6 +1738,9 @@ def __bool__(self):
 
 #: The Python representation of the JavaScript null object.
 jsnull: JsNull = object.__new__(JsNull)
+from json import encoder
+
+encoder._JSNULL = jsnull  # type:ignore[attr-defined]
```
ryanking13, line 1743: I don't quite understand this change. Does encoding fail if we don't have this here? If that's the case, why is it located in `_core_docs.py`, which is for documentation?

cpython/patches/0010-Teach-json-encoder.py-to-encode-jsnull.patch
```diff
@@ -0,0 +1,98 @@
...
+         elif o is False:
+diff --git a/Modules/_json.c b/Modules/_json.c
+index e33ef1f5eea..3152abd27f3 100644
+--- a/Modules/_json.c
++++ b/Modules/_json.c
+@@ -1302,11 +1302,13 @@ encoder_call(PyEncoderObject *self, PyObject *args, PyObject *kwds)
+     return result;
+ }
+ 
++extern PyObject* py_jsnull;
```
ryanking13, line 67: I wonder if there would be some chance we can upstream this.

ryanking13 (commented):
Thanks! It is a bit sad that we need another CPython patch for this.

cpython/patches/0010-Teach-json-encoder.py-to-encode-jsnull.patch
hoodmane: My plan is to make an Emscripten js ffi PEP. It will include stuff like this.

src/py/_pyodide/_core_docs.py
hoodmane: Well `_core_docs` is where we define `jsnull`. I could import it into another file and make the assignment there but doing it as close as possible to the location of definition of `jsnull` seemed the most sensible to me.

It's worth pointing out that `_core_docs.py` isn't just for docs, any value we need that can be defined in pure Python is defined here.

ryanking13: Oh, I see. I think we originally used this file for docs purpose only. Not sure when we started to put other things as well. Maybe we should find a better way to split files for docs and real object definition to avoid confusion? (no action required for this PR).

cpython/patches/0010-Teach-json-encoder.py-to-encode-jsnull.patch
ryanking13: That would be great

ryanking13 (approved):
Thanks!

src/py/_pyodide/_core_docs.py
hoodmane: Yeah it'd be nice to split it into two files, one for things that are replaced at runtime and a second for things that aren't.

hoodmane:

PR #5806, opened 2025-07-29
Add a check for cmake version

zsaladin:
### Description

Added a check for cmake version and give a detailed guidance.

### DIFF
--- a/tools/dependency-check.sh
+++ b/tools/dependency-check.sh
@@ -47,6 +47,15 @@ check_shasum() {
 
 check_cmake() {
   check_binary_present "cmake"
+
+  CMAKE_FULL_VERSION=$(cmake --version | head -n1 | cut -d " " -f 3)
+  CMAKE_MAJOR_VERSION=$(echo "${CMAKE_FULL_VERSION}" | cut -d "." -f 1)
+
+  if (( "${CMAKE_MAJOR_VERSION}" != "3" )); then
+    echo >&2 "Pyodide requires cmake 3."
+    echo >&2 "If you are on macOS brew would install higher version. Activating conda will fix it"
+    exit 1
+  fi
 }
 
 check_sed() {


tools/dependency-check.sh
```diff
@@ -47,6 +47,15 @@ check_shasum() {
 
 check_cmake() {
   check_binary_present "cmake"
+
+  CMAKE_FULL_VERSION=$(cmake --version | head -n1 | cut -d " " -f 3)
+  CMAKE_MAJOR_VERSION=$(echo "${CMAKE_FULL_VERSION}" | cut -d "." -f 1)
+
+  if (( "${CMAKE_MAJOR_VERSION}" != "3" )); then
+    echo >&2 "Pyodide requires cmake 3."
+    echo >&2 "If you are on macOS brew would install higher version. Activating conda will fix it"
```
zsaladin: In addition, this command below will fix the problem, and seems more mac-friendly.
However, it's too verbose, so I haven't added it.
Also it's related with #5793 

```bash
wget https://raw.githubusercontent.com/Homebrew/homebrew-core/b4e46db74e74a8c1650b38b1da222284ce1ec5ce/Formula/c/cmake.rb && brew install --build-from-source cmake.rb
```

hoodmane: Yeah it's not appropriate to attempt to install dependencies in a file called `dependency-check.sh`.

ryanking13: I do not want to recommend conda, unless there is no other easy way to install the old version of cmake in MacOS. I think the way that @youknowone added, installing the CMake Python package, which bundles the CMake binary in it, was a reasonable alternative, as we need Python installed anyway.

zsaladin: I was wondering if there's a particular reason you don't prefer using the existing [conda environment](https://github.com/pyodide/pyodide/blob/ab45a2025559b18dd2cbd896baccc00660b97ea5/environment.yml#L11).

Personally, I think installing packages with pip outside of an environment management tool can lead to potential issues. It might break existing dependency resolutions or be affected by future package installations. In this case, it interferes with the existing conda environment.

That’s why I believe it’s better to use a tool like conda, or to install packages in an isolated environment such as with [pipx](https://github.com/pypa/pipx). Since this project and `cmake` are already managed by conda, using conda here might be a better option than `pipx`.

Fortunately, cmake is a simple package with no additional dependencies, so it hasn’t caused any problems so far.

I'm happy to hear any feedback you might have 😀

zsaladin: Additionally, [this](https://github.com/pyodide/pyodide/blob/ab45a2025559b18dd2cbd896baccc00660b97ea5/environment.yml#L21-L22) seems to be intended for situations like this, so I thought it would make sense to leverage the existing managed environment.

zsaladin: @ryanking13 shared his opinion in another channel, and now I understand that pip is more approachable than conda for most people.

Applied this!
d7f4352a441cf429f7f2512ad1417676d3cd34ef

If you want to squash these commits, I will do it.

PR #5808, opened 2025-07-29
Resolve paths in python_cli_entry argv on Windows

dom96:
### Description

This is a change that gets us one step closer to being able to create a working Pyodide venv on Windows. 

### Checklist

- [ ] Add / update tests

### Test Plan

```
uv venv -p .\python.bat -v
DEBUG uv 0.8.3 (7e78f54e7 2025-07-24)
DEBUG Using Python request `.\python.bat` from explicit request
DEBUG Checking for Python interpreter at path `.\python.bat`
Using CPython 3.13.2 interpreter at: .
Creating virtual environment at: .venv
error: Failed to create virtual environment
  Caused by: cannot make an empty path absolute
```

### DIFF
--- a/src/templates/python_cli_entry.mjs
+++ b/src/templates/python_cli_entry.mjs
@@ -19,7 +19,23 @@ const thisProgramFlag = "--this-program=";
 const thisProgramIndex = process.argv.findIndex((x) =>
   x.startsWith(thisProgramFlag),
 );
-const args = process.argv.slice(thisProgramIndex + 1);
+const args = process.argv.slice(thisProgramIndex + 1).map(
+  arg => {
+    if (process.platform !== "win32") {
+      return arg;
+    }
+
+    // We replace Windows-style paths with Unix-style paths as that is how
+    // our Windows filesystem is mapped in emscripten.
+    if (arg.startsWith("\"C:\\")) {
+      return arg.slice(3).replaceAll("\\", "/");
+    } else if (arg.startsWith("C:\\")) {
+      return arg.slice(2).replaceAll("\\", "/");
+    } else {
+      return arg;
+    }
+  }
+);
 const _sysExecutable = process.argv[thisProgramIndex].slice(
   thisProgramFlag.length,
 );


src/templates/python_cli_entry.mjs
```diff
@@ -19,7 +19,21 @@ const thisProgramFlag = "--this-program=";
...
+  // We replace Windows-style paths with Unix-style paths as that is how
+  // our Windows filesystem is mapped in emscripten.
+  if (arg.startsWith('"C:\\')) {
+    return arg.slice(3).replaceAll("\\", "/");
+  } else if (arg.startsWith("C:\\")) {
+    return arg.slice(2).replaceAll("\\", "/");
+  } else {
+    return arg;
+  }
+});
```
hoodmane, line 36:
```suggestion
let args = process.argv.slice(thisProgramIndex + 1);
if (process.platform === "win32") {
  args = args.map((arg) => {
    // We replace Windows-style paths with Unix-style paths as that is how
    // our Windows filesystem is mapped in emscripten.
    if (arg.startsWith('"C:\\')) {
      return arg.slice(3).replaceAll("\\", "/");
    } else if (arg.startsWith("C:\\")) {
      return arg.slice(2).replaceAll("\\", "/");
    } else {
      return arg;
    }
  }
});
```

src/templates/python_cli_entry.mjs
```diff
@@ -19,7 +19,21 @@ const thisProgramFlag = "--this-program=";
...
 );
-const args = process.argv.slice(thisProgramIndex + 1);
+const args = process.argv.slice(thisProgramIndex + 1).map((arg) => {
+  if (process.platform !== "win32") {
+    return arg;
+  }
+
+  // We replace Windows-style paths with Unix-style paths as that is how
+  // our Windows filesystem is mapped in emscripten.
+  if (arg.startsWith('"C:\\')) {
```
ryanking13, line 29: Will there be a case that the user's drive is not C?

dom96: Good point, if the user runs this on a different partition then this will fail.

hoodmane: Maybe we should make a windows path replacement regex?

Issue #5756, opened 2025-07-12
Remove unnecessary setup.py for pyodide-py


ryanking13:
## Description

1. The setup.py under the [`src/py`](https://github.com/pyodide/pyodide/blob/main/src/py/setup.py) directory was needed when we first adopted the pyproject.toml file, and there was an issue with the editable install.

2. I believe the recent pip and setuptools versions don't need that file.

3. Let's remove it.

## Goal

Remove the unnecessary setup.py file in the src/py directory. The editable install should work without it.

kaifansariw:
@ryanking13 can you assign this issue to me.

agriyakhetarpal:
Hi @kaifansariw, how is this going? Please feel free to request for help with this, if you need any!

kaifansariw:
@agriyakhetarpal only this one file ??

kaifansariw:
@agriyakhetarpal i have made changes in code but don't know how to create pull request.
can you give me command.

ryanking13:
@kaifansariw 

You can follow the document in GitHub to create a pull request: https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/proposing-changes-to-your-work-with-pull-requests/creating-a-pull-request , if that is what you are asking.

It would be nice if you can verify that the change locally though. You can run the following command in `src/py`

```bash
# inside src/py
pip install -e .
```

and see if the `pyodide-py` is installed correctly in your local environment, by checking

```
pip freeze
```

kaifansariw:
<img width="890" height="184" alt="Image" src="https://github.com/user-attachments/assets/0dcfd290-8405-40a5-adfd-f3a1c11bf9e7" />
i dont know why pull request is not created yet

ryanking13:
You are in the branch `feature-1` but pushing the `main` branch.

Try

```
git push origin feature-1
```

ShauryaDusht:
Hi @kaifansariw 

Make a new branch
```
git checkout -b your-branch-name
```
Add and commit your changes
```
git add .
git commit -m "your message"
```
Push your branch to origin
```
git push origin your-branch-name
```

After that see [Pull requests](https://github.com/pyodide/pyodide/pulls) and explain things

hoodmane:
Closing as completed in #5812.

kaifansariw:
@hoodmane this issue was label as ossca 2025 , what does it mean is it an open source program and how can i participate in it

ryanking13:
[...]

PR #5812, opened 2025-08-01
Remove redundant setup.py file in src/py

kaifansariw:
### Description

  The setup.py under the [src/py](https://github.com/pyodide/pyodide/blob/main/src/py/setup.py) directory was needed when we first adopted the pyproject.toml file, and there was an issue with the editable install.<#5756> 
Deleted the setup.py under the [src/py].
### Checklist

- [ X] Add / update tests


kaifansariw:
@ryanking13  this issue was labled as ossca 2025 i would like to is it an open source program and how can i contribute to it.

PR #5813, opened 2025-08-01
Fix test_relaxed_call for Python 3.13 TypeError message format

Drranny:
### Description

This PR fixes the `test_relaxed_call` and `test_relaxed_wrap` tests, which were failing on Python 3.13 due to changes in the `TypeError` message format.

In Python 3.13, when a required positional-only argument is missing, the error message no longer includes the function name. However, the original tests used a regex that expected the full function name, which caused them to fail.

I updated the regex patterns in both tests to match the new, simplified message format. After the changes, both tests pass locally on Python 3.13.

- [V ] Add / update tests

image

image

### DIFF
--- a/src/tests/test_pyodide.py
+++ b/src/tests/test_pyodide.py
@@ -283,7 +283,6 @@ def test_relaxed_call():
     with pytest.raises(TypeError, match=r"missing .* argument: 'a'"):
         relaxed_call(f4, a=2, b=7)
 
-
     def f5(a, *args, b=7, **kwargs):
         return [a, args, b, kwargs]
 
--- a/src/tests/test_pyodide.py
+++ b/src/tests/test_pyodide.py
@@ -278,10 +278,7 @@ def test_relaxed_call():
     def f4(a, /, *args, b=7):
         return [a, args, b]
 
-    with pytest.raises(
-        TypeError,
-        match=r"missing .* argument: 'a'"
-    ):
+    with pytest.raises(TypeError, match=r"missing .* argument: 'a'"):
         relaxed_call(f4, a=2, b=7)
 
     def f5(a, *args, b=7, **kwargs):
@@ -341,12 +338,8 @@ def test_relaxed_wrap():
     def f4(a, /, *args, b=7):
         return [a, args, b]
 
-    with pytest.raises(
-        TypeError,
-        match=r"missing .* argument: 'a'"
-    ):
+    with pytest.raises(TypeError, match=r"missing .* argument: 'a'"):
         f4(a=2, b=7)
-
 
     @relaxed_wrap
     def f5(a, *args, b=7, **kwargs):
--- a/src/tests/test_pyodide.py
+++ b/src/tests/test_pyodide.py
@@ -278,8 +278,11 @@ def test_relaxed_call():
     def f4(a, /, *args, b=7):
         return [a, args, b]
 
+    # Since Python 3.13.3, the TypeError message for missing positional-only arguments has changed.
+    # See: https://github.com/python/cpython/pull/130192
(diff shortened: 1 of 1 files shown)


Drranny:
Hello! I noticed that both `test_relaxed_call` and `test_relaxed_wrap` were failing on Python 3.13 due to changes in the `TypeError` message format.

In Python 3.13, the error message for missing positional-only arguments no longer includes the function name. I updated the regex patterns in both tests to match the new format.

Both tests now pass locally on Python 3.13 — please let me know if there's anything else I should adjust.

Thanks for your time and review!

hoodmane:
They don't seem to be passing in CI.

Drranny:
> They don't seem to be passing in CI.

Thanks for the update!
I’ve fixed the test_relaxed_wrap failure on Python 3.13 and force-pushed the changes in [commit 2c647dd](https://github.com/Drranny/pyodide/commit/2c647dd).
After the update, the test now passes locally on both Python 3.12 and 3.13.

ryanking13:
That testcase is passing in our CI which uses Python 3.13 as well ([log](https://github.com/pyodide/pyodide/actions/runs/16646608603/job/47108736743#step:6:1108)). If you are having errors when you run the test locally, could you please attach what error do you get in your local environment?

Drranny:
> That testcase is passing in our CI which uses Python 3.13 as well ([log](https://github.com/pyodide/pyodide/actions/runs/16646608603/job/47108736743#step:6:1108)). If you are having errors when you run the test locally, could you please attach what error do you get in your local environment?

image

Thanks for the quick follow-up!
I checked the CI logs you shared and as you mentioned, the test does seem to pass on Python 3.13 in the CI environment.

That being said, I'm still encountering failures locally (also on Python 3.13), so I’m wondering if it could be something specific to my environment.

Here's the traceback I get:

```
TypeError: missing a required positional-only argument: 'a'
```

And the test fails with the following message:

```
AssertionError: Regex pattern did not match.
Regex: "test_relaxed_call\\.<locals>\\.f4\\(\\)\\ missing\\ 1\\ required\\ positional\\ argument:\\ 'a'"
Input: "missing a required positional-only argument: 'a'"
```

I’m happy to look into adjusting the regex or investigating further if that would be helpful — just let me know how you'd like to proceed!

ryanking13:
@Drranny 

Thanks for providing the error message! I checked if there was a change in CPython side, and indeed there was an update in the `Inspect` module 6 month ago: (https://github.com/python/cpython/pull/130192), and the patch is backported to Python3.13.X since Python 3.13.3.

So if you are using Python 3.13.3 or higher, it makes sense that it fails in your local environment (The CI environment uses CPython 3.13.2)

This PR is valid then. Thanks for fixing it.

src/tests/test_pyodide.py
```diff
@@ -278,12 +278,7 @@ def f3(a, *args, b=7):
     def f4(a, /, *args, b=7):
         return [a, args, b]
 
-    with pytest.raises(
-        TypeError,
-        match=re.escape(
-            "test_relaxed_call.<locals>.f4() missing 1 required positional argument: 'a'"
-        ),
-    ):
+    with pytest.raises(TypeError, match=r"missing .* argument: 'a'"):
```
ryanking13, line 283: Could you please add a comment with a link that the error message regarding the positional-only argument has been changed since CPython 3.13.3 (https://github.com/python/cpython/pull/130192)?

It would help future contributors who modify this test.

Drranny: I’ve added the comment and pushed the changes. Thank you for the helpful suggestion!

ryanking13:
Thanks @Drranny!

PR #5815, opened 2025-08-03
tests: fix uncaught_exception path duplication under Docker (/src)

leehanjeong:
### Description

* **Reasoning** – In the Docker test environment `PYODIDE_ROOT` is `/src`,
  while tracebacks start with `/src/src/...`.  The old test replaced
  `/src` with `$PYODIDE_ROOT`, which sometimes produced a duplicated
  `$PYODIDE_ROOT$PYODIDE_ROOT/...` path and caused sporadic failures.
* **Change** – Rewrote `test_asyncio.py::test_uncaught_exception` to use
  four straightforward `assert` statements:
    1. “Unhandled exception in event loop”
    2. “Traceback (most recent call last):”
    3. source file + function name (`src/tests/test_asyncio.py`, `in f`)
    4. “ZeroDivisionError: division by zero”
  Line numbers and root-path substitutions are no longer needed.

### DIFF
--- a/src/tests/test_asyncio.py
+++ b/src/tests/test_asyncio.py
@@ -1,12 +1,9 @@
 import asyncio
-import re
 import time
-from textwrap import dedent
 
 import pytest
 from pytest_pyodide import run_in_pyodide
 
-from conftest import PYODIDE_ROOT
 from pyodide.code import eval_code_async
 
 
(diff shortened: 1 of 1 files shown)


hoodmane:
What about just changing it to replace `"{PYODIDE_ROOT}` with `"`?

leehanjeong:
Thanks for the suggestion! Let me make sure I understand the implications correctly.

If we change `str(PYODIDE_ROOT)` to `""`, here's what would happen:

**In Docker environment (where `PYODIDE_ROOT` is `/src`):**
- Actual traceback: `/src/src/tests/test_asyncio.py` 
- After replacing all `/src` with `""`: `/tests/test_asyncio.py`
- If we update expected_message accordingly: `/tests/test_asyncio.py` ✅ (matches)

**In other environments (where `PYODIDE_ROOT` might be `/some/other/path`):**
- Actual traceback: `/some/other/path/src/tests/test_asyncio.py`
- After replacing all `/some/other/path` with `""`: `/src/tests/test_asyncio.py`
- Expected message: `/tests/test_asyncio.py` ❌ (doesn't match)

Am I understanding this correctly? The issue is that the solution would be environment-specific.

If we want to minimize code changes while fixing the duplication issue, what about using the `count` parameter to replace only the first occurrence?

```python
selenium.logs.replace(str(PYODIDE_ROOT), "$PYODIDE_ROOT", 1)  # Only replace first match
```

This would handle the Docker duplication case (`/src/src/...` → `$PYODIDE_ROOT/src/...`) while keeping the existing expected_message format. What do you think?

ryanking13:
Yeah, I suggested this way, and I think this is more robust than before. What we want to test in that test case is whether or not the exception inside the async function is correctly handled. We don't need to rely on a specific format of the error message.

ryanking13 (approved):

Issue #5754, opened 2025-07-12
Replace prettier in pre-commit with oxlint or other alternatives


ryanking13:
## Description

1. We are using pre-commit to lint the whole codebase. The [.pre-commit-config.yaml](https://github.com/pyodide/pyodide/blob/main/.pre-commit-config.yaml) file contains the pre-commit settings.

2. `prettier` is currently used to lint the TypeScript and yaml codes.

3. However, prettier is quite slow and the pre-commit.ci's prettier mirror is no longer maintained ([repo](https://github.com/pre-commit/mirrors-prettier)).

4. We would like to replace it with a more fast linter, such as [oxlint](https://oxc.rs/docs/guide/usage/linter).

5. yaml linter is no longer required, as the recipes in this repository has been mostly moved to [pyodide/pyodide-recipes](https://github.com/pyodide/pyodide-recipes).

## Goal

Replace prettier in the pre-commit config with a faster TypeScript linter.

rbm0524:
@ryanking13 Can you assign this issue to me?

ryanking13:
@rbm0524 assigned. Thanks!

hoodmane:
Prettier is formatting the typescript/JavaScript. We are not currently linting it.

hoodmane:
Precommit is formatting the typescript/JavaScript. We are not currently linting it.

rbm0524:
@ryanking13 Could you please close this issue? My PR was merged, but it didn't close automatically because I didn't write down the resolve keyword by mistake.

hoodmane:
Closed as completed in #5820.

PR #5818, opened 2025-08-05
style: Replace prettier with oxlint

rbm0524:
#### Related Issue
#5754 

I deleted prettier in the .pre-commit-config.yaml file and added oxlint. Is there a folder to exclude for oxlint application? Also, do I need to do more documentation? I'm not used to open source activities, so I'd appreciate it if you could let me know.

### Checklists
- [ ] Add / update tests

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -85,10 +85,12 @@ repos:
         exclude: (^packages/.*/setup.py|/src|pyodide-build)
         additional_dependencies: *mypy-deps
 
-  - repo: https://github.com/pre-commit/mirrors-prettier
-    rev: "v4.0.0-alpha.8"
+  - repo: https://github.com/oxc-project/mirrors-oxlint
+    rev: "v1.9.0"
     hooks:
-      - id: prettier
+      - id: oxlint
+        args: [--fix]
+        types_or: [javascript, jsx, ts, tsx]
 
 ci:
   autoupdate_schedule: "quarterly"
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -89,7 +89,6 @@ repos:
     rev: "v1.9.0"
     hooks:
       - id: oxlint
-        args: [--fix]
         types_or: [javascript, jsx, ts, tsx]
 
 ci:
--- a/docs/project/changelog.md
+++ b/docs/project/changelog.md
@@ -17,6 +17,10 @@ myst:
 
 ## Unreleased
 
+## Version 0.28.1
+
+_August 04, 2025_
+
 ### General
 
 - {{ Update }} Upgraded `micropip` to 0.10.1 {pr}`5739`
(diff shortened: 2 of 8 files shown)


.pre-commit-config.yaml
```diff
@@ -85,10 +85,11 @@ repos:
...
         additional_dependencies: *mypy-deps
 
-  - repo: https://github.com/pre-commit/mirrors-prettier
-    rev: "v4.0.0-alpha.8"
+  - repo: https://github.com/oxc-project/mirrors-oxlint
+    rev: "v1.9.0"
     hooks:
-      - id: prettier
+      - id: oxlint
+        types_or: [javascript, jsx, ts, tsx]
```
hoodmane, line 92: I'd leave out jsx and tsx since we'll never have them.

hoodmane:
Does oxlint pick up `.prettierignore`?

hoodmane:
It says here:

> The formatter is currently work in progress.

https://oxc.rs/docs/guide/usage/formatter.html

@ryanking13 maybe we shouldn't be removing prettier until oxc's formatter is considered ready? Seems like we can add oxc just as a linter for now and keep prettier to do the formatter.

.pre-commit-config.yaml
rbm0524: I'll reflect on that!

rbm0524:
> It says here:
> 
> > The formatter is currently work in progress.
> 
> https://oxc.rs/docs/guide/usage/formatter.html
> 
> @ryanking13 maybe we shouldn't be removing prettier until oxc's formatter is considered ready? Seems like we can add oxc just as a linter for now and keep prettier to do the formatter.

I agree not to get rid of prettier. oxlint has not been implemented with formatter yet, so it would be good to use it as a linter.

ryanking13:
Oops, I didn't know that oxc formatter is still in progress.

The problem with prettier now is that the [pre-commit plugin for prettier is archived](https://github.com/pre-commit/mirrors-prettier), and it is pretty slow.

How about trying [biome](https://biomejs.dev/) instead then? They have pre-commit hooks too, and they advertise themselves as a "Format, lint, and more in a fraction of a second" + "[97% compatibility with Prettier](https://console.algora.io/challenges/prettier),"

rbm0524:
> Oops, I didn't know that oxc formatter is still in progress.
> 
> The problem with prettier now is that the [pre-commit plugin for prettier is archived](https://github.com/pre-commit/mirrors-prettier), and it is pretty slow.
> 
> How about trying [biome](https://biomejs.dev/) instead then? They have pre-commit hooks too, and they advertise themselves as a "Format, lint, and more in a fraction of a second" + "[97% compatibility with Prettier](https://console.algora.io/challenges/prettier),"

Then I don't need linter, so I'll delete oxtlint and reflect biome.

hoodmane:
Thanks @rbm0524!

PR #5820, opened 2025-08-05
style: Replace prettier with biome

rbm0524:
#### Related issue
#5754 
We have applied the format of bio-format to js and ts. The function of bio-format is "Format the committed files". 
I tried to reflect '.pretiterignore' in the biome.json file, but
```
/code/src/js/biome.json configuration ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━

× Found a nested root configuration, but there's already a root configuration.


i The other configuration was found in /code.


i Use the migration command from the root of the project to update the configuration.


$ biome migrate --write

configuration ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━



× Biome exited because the configuration resulted in errors. Please fix them.
```

Is there a way to get through pre-commit.ci even if I use the biome.json file?
I also want to consider whether it is necessary to migrate the contents of `.pretterignore` because the format is applied only to `js` and `ts`.

When I apply the formatter, I get an error in the CI, is there a formatting script or recommended settings that I use in this project?

 Please let me know if there is anything else we need to revise or discuss. 

#5818 branch is closed. I reflected what we discussed in that PR.

### Checklist

- [ ] Add / update tests


hoodmane:
This is very noisy. Is there a way to configure it so that the settings are more similar to prettier? I'd prefer to see a few hundred lines of stuff max. We probably should also keep the formatting in a separate commit to main so we can add it to .git-blame-ignore-revs .

hoodmane:
> Biome provides a migrate command that reads your existing configuration files and attempts to port the settings into Biome:
>
> ```sh
> biome migrate prettier --write
> ```

hoodmane:
https://biomejs.dev/guides/migrate-eslint-prettier/#migrate-from-prettier

ryanking13:
> This is very noisy. Is there a way to configure it so that the settings are more similar to prettier? I'd prefer to see a few hundred lines of stuff max. We probably should also keep the formatting in a separate commit to main so we can add it to .git-blame-ignore-revs .

+1

It would be nice if we could start with a formatting rule that is mostly compatible to the current code in this PR, and gradually apply more formatting options

youknowone:
The most noisy part is switching from space to tab. Keeping space will make the diff reasonable size.

hoodmane:
`biome migrate prettier --write` apparently automatically migrates the config.

rbm0524:
> `biome migrate prettier --write` apparently automatically migrates the config.

biome supports migration from `.prettierrc.json` to `biome.json.` It does not reflect `.prettierignore`.
There's no `prettierrc.json`. So, I think don't need to migration.

https://biomejs.dev/guides/big-projects/#monorepo
There was a recommended rule in this document, so I reflected it. Any other configurations do you want?

rbm0524:
> This is very noisy. Is there a way to configure it so that the settings are more similar to prettier? I'd prefer to see a few hundred lines of stuff max. We probably should also keep the formatting in a separate commit to main so we can add it to .git-blame-ignore-revs .

Sorry for that. I was inexperienced. I added `.git-blame-ignore-revs` to hide specific commit. Is it applied correctly?

hoodmane:
We want to use the biome config that's as close as possible to the default prettier config. If it requires a `.prettierrc` file to do a migration just make an empty one.

rbm0524:
> We want to use the biome config that's as close as possible to the default prettier config. If it requires a `.prettierrc` file to do a migration just make an empty one.

Migration problem was solved. 
```
× Found a nested root configuration, but there's already a root configuration.


i The other configuration was found in /code.
```

https://biomejs.dev/guides/big-projects/#monorepo
In this documents, 

>Create nested configuration files, one in each package where it is needed. These nested configuration files must have the field "root" set to false.
For convenience, when you use the microsyntax extends: "//", you can omit "root": false, because it is already implied that this configuration isn’t a root configuration:

I solved this problem through `"extends": "//",`

hoodmane:
It doesn't seem solved -- it is changing all the files to be indented with tabs instead of spaces. That needs to be fixed.

rbm0524:
> It doesn't seem solved -- it is changing all the files to be indented with tabs instead of spaces. That needs to be fixed.

I created a `.prettierrc.json` and completed the migration. I think it's applied now, so please check.

ryanking13:
Hi @rbm0524, thanks for your work.

It seems like we need to teach biome to use space instead of tab, I think we need `biome.json` starting with some basic configuration.

As you can see in the commit log, pre-commit.ci is keep auto fixing space to tabs.

image

Here are some additional comments:

1. You don't need to add `.git-blame-ignore-revs` file in this PR, as we don't know the commit hash until it is merged. That file can be added in a follow-up PR if this PR generates a large diff.
2. Let's remove `.prettierrc.json` and add `biome.json` instead. I think we can start with some basic configuration, as outlined in the [biome docs](https://biomejs.dev/guides/configure-biome/#_top), and then add more rules.
3. You can run `pre-commit run -a` locally to apply the formatting to the entire codebase regardless of commit. This will help you understand how formatting will be applied to the codebase.

rbm0524:
> Hi @rbm0524, thanks for your work.
> 
> It seems like we need to teach biome to use space instead of tab, I think we need `biome.json` starting with some basic configuration.
> 
> As you can see in the commit log, pre-commit.ci is keep auto fixing space to tabs.
> 
> image
> Here are some additional comments:
> 
> 1. You don't need to add `.git-blame-ignore-revs` file in this PR, as we don't know the commit has until is is merged. That file can be added in a follow-up PR if this PR generates a large diff.
> 2. Let's remove `.prettierrc.json` and add `biome.json` instead. I think we can start with some basic configuration, as outlined in the [biome docs](https://biomejs.dev/guides/configure-biome/#_top), and then add more rules.
> 3. You can run `pre-commit run -a` locally to apply the formatting to the entire codebase regardless of commit. This will help you understand how formatting will be applied to the codebase.

Removed `.prettierrc.json`. `biome.json` existed in the previous commit, but nested `biome.json` still seems to cause errors. So I moved `biome.json` to root.

ryanking13:
Great. Thanks. Now I think we need to teach biome to use double quote instead of single quote, and append semicolons.

> but nested biome.json still seems to cause errors. So I moved biome.json to root.

Yes, I think it is okay, all the javascript files in this repository can follow the same rule.

biome.json
```diff
@@ -0,0 +1,54 @@
...
+		"attributePosition": "auto",
+		"bracketSameLine": false,
+		"bracketSpacing": true,
+		"expand": "auto",
+		"useEditorconfig": true
+	},
+	"linter": { "enabled": false, "rules": { "recommended": true } },
+	"javascript": {
+		"formatter": {
+			"jsxQuoteStyle": "double",
```
hoodmane:
```suggestion
```

biome.json
```diff
@@ -0,0 +1,54 @@
...
+			"quoteProperties": "asNeeded",
+			"trailingCommas": "all",
+			"semicolons": "asNeeded",
+			"arrowParentheses": "always",
+			"bracketSameLine": false,
+			"quoteStyle": "single",
```
hoodmane:
```suggestion
			"quoteStyle": "double",
```

biome.json
```diff
@@ -0,0 +1,54 @@
...
+		"expand": "auto",
+		"useEditorconfig": true
+	},
+	"linter": { "enabled": false, "rules": { "recommended": true } },
+	"javascript": {
+		"formatter": {
+			"jsxQuoteStyle": "double",
+			"quoteProperties": "asNeeded",
+			"trailingCommas": "all",
+			"semicolons": "asNeeded",
```
hoodmane: I am strongly against semicolons being as needed.

biome.json
rbm0524: I removed that

biome.json
rbm0524: I changed it

biome.json
rbm0524: I changed it to "always".
Now, semicolons are always added at the end of each statement.

hoodmane:
Looks great, thanks @rbm0524 . Now let's remove the ignore revs file and exclude pyodide-cross-build-environments.json and revert it and it'll be ready to merge.

rbm0524:
> Looks great, thanks @rbm0524 . Now let's remove the ignore revs file and exclude pyodide-cross-build-environments.json and revert it and it'll be ready to merge.

I removed ignore revs file and revert `pyodide-cross-build-environments.json`

hoodmane:
Thanks @rbm0524 !

rbm0524:
@hoodmane Thank you very much. I hope I can contribute more in the future!

PR #5824, opened 2025-08-07
Fix cdnURL in Node

### LINKED ISSUE #5821: Error loading micropip with pyodide version 0.28.1
## 🐛 Bug

When attempting to load `micropip` using `pyodide.loadPackage('micropip')`  in a node.js environment with Pyodide version 0.28.1, the operation fails with an ENOENT: no such file or directory error. This appears to be a regression from version 0.28.0, where the same code successfully fetches `micropip`.

### To Reproduce

```
const { loadPyodide } = require('pyodide');

async function main() {
};
main();
```

Result is:

```
Loading micropip
Didn't find package micropip-0.10.1-py3-none-any.whl locally, attempting to load from /tmp/pydtest/node_modules/pyodide/
Failed to load micropip
ENOENT: no such file or directory, open '/tmp/pydtest/node_modules/pyodide/micropip-0.10.1-py3-none-any.whl
```

### Expected behavior

```
Loading micropip
Didn't find package micropip-0.10.0-py3-none-any.whl locally, attempting to load from https://cdn.jsdelivr.net/pyodide/v0.28.0/full/
Package micropip-0.10.0-py3-none-any.whl loaded from https://cdn.jsdelivr.net/pyodide/v0.28.0/full/, caching the wheel in node_modules for future use.
Loaded micropip
```

### Environment

- Pyodide Version: **0.28.1**
- Node Version: **24.5.0**

ryanking13:
### Description

Resolve #5821

There was some regression introduced in #5764.

Previously, loading Pyodide in Node without passing any options would set

- indexURL: dist directory
- packageBaseURL: dist directory
- cdnURL: dist directory ==> incorrect

This ensures the CDN URL is always set to the same value.

### Checklist

- [x] Add / update tests

### DIFF
--- a/src/js/load-package.ts
+++ b/src/js/load-package.ts
@@ -116,7 +116,7 @@ export class PackageManager {
    *
    * exported for testing purposes.
    */
-  public cdnURL: string = "";
+  public cdnURL: string;
 
   /**
    * The set of loaded packages.
@@ -157,7 +157,7 @@ export class PackageManager {
       // back to cdnURL
       this.installBaseUrl =
         this.#api.config.packageCacheDir ?? this.#api.config.packageBaseUrl;
-      this.cdnURL = this.#api.config.cdnUrl;
+      this.cdnURL = `https://cdn.jsdelivr.net/pyodide/v${this.#api.version}/full/`;
     } else {
       // use packageBaseUrl as the base URL for the packages
       this.installBaseUrl = this.#api.config.packageBaseUrl;
--- a/src/js/load-package.ts
+++ b/src/js/load-package.ts
@@ -113,8 +113,10 @@ export class PackageManager {
    * Only used in Node. If we can't find a package in node_modules, we'll use this
    * to fetch the package from the cdn (and we'll store it into node_modules so
    * subsequent loads don't require a web request).
+   *
+   * exported for testing purposes.
    */
-  private cdnURL: string = "";
+  public cdnURL: string = "";
 
   /**
    * The set of loaded packages.
@@ -154,7 +156,7 @@ export class PackageManager {
       // In node, we'll try first to load from the packageCacheDir and then fall
       // back to cdnURL
       this.installBaseUrl =
-        this.#api.config.packageCacheDir ?? API.config.packageBaseUrl;
+        this.#api.config.packageCacheDir ?? this.#api.config.packageBaseUrl;
       this.cdnURL = this.#api.config.cdnUrl;
     } else {
       // use packageBaseUrl as the base URL for the packages
@@ -713,4 +715,6 @@ if (typeof API !== "undefined" && typeof Module !== "undefined") {
   if (API.lockFilePromise) {
     API.packageIndexReady = initializePackageIndex(API.lockFilePromise);
   }
+
+  API.packageManager = singletonPackageManager;
 }
--- a/src/js/pyodide.ts
+++ b/src/js/pyodide.ts
@@ -23,6 +23,7 @@ import type {
 } from "./types";
 import type { EmscriptenSettings } from "./emscripten-settings";
 import type { SnapshotConfig } from "./snapshot";
+import { withTrailingSlash } from "./common/path";
 export type { PyodideAPI, TypedArray };
 export type { LockfileInfo, LockfilePackage, Lockfile } from "./types";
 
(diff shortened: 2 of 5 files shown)


src/js/load-package.ts
```diff
@@ -154,8 +156,8 @@ export class PackageManager {
       // In node, we'll try first to load from the packageCacheDir and then fall
       // back to cdnURL
       this.installBaseUrl =
-        this.#api.config.packageCacheDir ?? API.config.packageBaseUrl;
-      this.cdnURL = this.#api.config.cdnUrl;
+        this.#api.config.packageCacheDir ?? this.#api.config.packageBaseUrl;
+      this.cdnURL = `https://cdn.jsdelivr.net/pyodide/v${this.#api.version}/full/`;
```
hoodmane: This doesn't make sense to me with a custom lock file: if the custom lock file includes any extra packages they won't be in the CDN. How is that supposed to work? I think the fix is to set `cdnURL` to `packageBaseUrl` only when one is explicitly provided. The bug as far as I'm concerned was that we prefer a computed `packageBaseUrl` to the default cdn url.

ryanking13: Hmm... okay. In that case, one should host all the packages in the Pyodide distribution together, right?

ryanking13: Okay, addressed. It always becomes tricky when it comes to custom lockfiles. Not sure what would be the best way to deal with it yet.

hoodmane: > It always becomes tricky when it comes to custom lockfiles. Not sure what would be the best way to deal with it yet.

Totally agree. It's really hard to understand our logic here and it's an intrinsically complex area. Someone should probably try to write an actual specification for this and think through all the different cases.

src/js/pyodide.ts
```diff
@@ -257,11 +258,16 @@ export async function loadPyodide(
...
-  indexURL = resolvePath(indexURL);
-  if (!indexURL.endsWith("/")) {
-    indexURL += "/";
-  }
+  indexURL = withTrailingSlash(resolvePath(indexURL));
   const options_ = options as ConfigType;
+
+  // cdnUrl only for node.
+  options_.packageBaseUrl = withTrailingSlash(options_.packageBaseUrl);
+  options_.cdnUrl = withTrailingSlash(
```
hoodmane, line 266: This second call to `withTrailingSlash()` is redundant since both of the two options already end in a trailing slash.

src/js/load-package.ts
ryanking13: Yeah, we should list some possible use cases and think of a way to improve the lockfile itself, too. Roman made some ideas before, but we haven't been made any progress after that (https://github.com/pyodide/pyodide-lock/issues).

src/js/pyodide.ts
ryanking13: Yes, but I wanted to be explicit as possible to prevent any human errors later, such as removing the slash from the CDN URL string later by mistake.

hoodmane: Maybe add a comment.

src/tests/test_package_loading.py
```diff
@@ -1189,3 +1189,78 @@ def test_micropip_freeze_with_package_cache_dir(selenium_standalone_noload, tmp_
     assert lockfile_content["packages"]["micropip"]["file_name"] == str(
         package_cache_dir / micropip_path.name
     )
+
+
+@only_node
```
hoodmane, line 1194: Does it need to be `@only_node`?

src/tests/test_package_loading.py
```diff
@@ -1189,3 +1189,78 @@ def test_micropip_freeze_with_package_cache_dir(selenium_standalone_noload, tmp_
...
+
+    def with_slash(path: str | Path) -> str:
+        return str(path).rstrip("/") + "/"
+
+    # no option
+    selenium.run_js(
+        f"""
+        pyodide = await loadPyodide();
+        version = pyodide._api.version;
+        assert(() => pyodide._api.packageManager.cdnURL === `https://cdn.jsdelivr.net/pyodide/v${{version}}/full/`);
```
hoodmane: Let's extract the jsdelivr url into a Python variable so we can do ` assert(() => pyodide._api.packageManager.cdnURL === "{JSDELIVR_URL}"` throughout.

hoodmane (approved):
Test looks great, thanks for adding it.

src/tests/test_package_loading.py
ryanking13: I added another testcase for browsers.

